PR #247, opened 2024-12-22
[dist] generate release-bin tar file

### LINKED ISSUE #246: [Feature] Generate fluss-xxx-bin.tar.gz package while building project.
### Search before asking

### Motivation

Currently, building project will not generate bin package, compress `build-target` package manually will make some differences between `Windows`、`OS X` and `Linux`, It's necessary to generator `fluss-xxx-bin.tar.gz` file for deploying fluss cluster conveniently.

### Solution

_No response_

### Anything else?

_No response_

### Willingness to contribute

Vipamp:
### Purpose

Linked issue: close #246 
This pr is aims to solve this feature.


fluss-dist/src/main/assemblies/assembly.xml
```diff
@@ -0,0 +1,157 @@
...
+        <format>dir</format>
+        <format>tar.gz</format>
+    </formats>
+
+    <includeBaseDirectory>true</includeBaseDirectory>
+    <baseDirectory>fluss-${project.version}-bin</baseDirectory>
+
+    <dependencySets>
+        <dependencySet>
+            <outputDirectory>lib</outputDirectory>
```
wuchong: What's the different between this `assembly.xml` and the existing `bin.xml`, `opt.xml`, `plugins.xml`? We have build `fluss-xxx-bin` directory. Can we just add a phase to package the directory into a `.tar.gz` file?

wuchong (commented):
Thanks, @Vipamp, I think this will be very convenient for users to build the Fluss binary locally.

fluss-dist/src/main/assemblies/assembly.xml
Vipamp: Hi,file `assembly.xml` is merged from `bin.xml`, `opt.xml` and `plugins.xml` with phase `tar.gz`, since adding the `tar.gz` phase to each file doesn't work, I add a new file and keep the original three files. If these three files are no longer required, these files will be deleted and keep `assembly.xml` instead,is it OK?

wuchong: We can't remove `bin.xml`, `opt.xml` and `plugins.xml`, because they are used to build `fluss-${project.version}-bin` directory, and the final release distribution tgz is built from this directory with additional licenses and notice. Currently, this works with the `create_binary_release.sh` script. 

To improve maintainability, we should eliminate the redundant assembly XML definitions, which complicate maintenance. Instead, we should focus on reusing the `fluss-dist/target/fluss-${project.version}-bin` directory for building the local tgz.


Here is how it works in my local, you can take this commit:
https://github.com/wuchong/fluss/commit/1cc0143a41755510f4557207c2703f70c060fe7b

Vipamp:
@wuchong @luoyuxia  PTAL，thinks

fluss-dist/src/main/assemblies/tgz.xml
```diff
@@ -0,0 +1,34 @@
...
+ distributed under the License is distributed on an "AS IS" BASIS,
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+-->
+<assembly
+        xmlns="http://maven.apache.org/plugins/maven-assembly-plugin/assembly/1.1.0"
+        xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+        xsi:schemaLocation="http://maven.apache.org/plugins/maven-assembly-plugin/assembly/1.1.0 http://maven.apache.org/xsd/assembly-1.1.0.xsd">
+    <id>tgt</id>
```
wuchong: tgz?

wuchong (approved):
LGTM

PR #248, opened 2024-12-23
[website] Add `install` stage to handle package changes

liyubin117:
### Purpose
Linked to https://github.com/alibaba/fluss/issues/252
once new packages introduced,  executed `deploy.sh` throws as follows:
```
 [cause]: Error: Docusaurus was unable to resolve the "@docusaurus/plugin-pwa" plugin. Make sure one of the following packages are installed:
  - @docusaurus/plugin-pwa
  - @docusaurus/docusaurus-plugin-plugin-pwa
```
we could add `install` stage to handle it.

### Tests
run deploy.sh

### DIFF
--- a/website/deploy.sh
+++ b/website/deploy.sh
@@ -17,5 +17,6 @@
 
 # docusaurus currently has bug with css minify, so we need to skip it
 npm run clear
+npm install
 npm run build -- --no-minify
 npm run deploy -- --skip-build


liyubin117:
@wuchong I found that the issue has stucked docs build for a while, PTAL, thanks :)

wuchong (approved):
LGTM

PR #276, opened 2024-12-27
[server] Revert the queue in RequestChannel from PriorityBlockingQueue back to ArrayBlockingQueue to fix config 'netty.server.max-queued-requests' error

swuferhong:
### Purpose

Linked issue: https://github.com/alibaba/fluss/issues/268

This pr is aims to fix the error that the server will receive unlimited requests even if we config the `netty.server.max-queued-requests` to a specify number. 
This issue arose because we introduced a `PriorityBlockingQueue` in the `RequestChannel` to replace the `ArrayBlockingQueue`. The `PriorityBlockingQueue`, unlike the `ArrayBlockingQueue`, is an unbounded queue. 

Therefore, in this PR, we will revert the queue in `RequestChannel` from `PriorityBlockingQueue` back to `ArrayBlockingQueue` to fix config 'netty.server.max-queued-requests' error. The reason we can roll back here is that the `PriorityBlockingQueue` has become unnecessary after introducing https://github.com/alibaba/fluss/pull/233 . In this pr, we significantly reduce the number of fetch requests during follower synchronization, which makes our server more stable even without the priority queue.

### DIFF
--- a/fluss-rpc/src/main/java/com/alibaba/fluss/rpc/netty/server/RequestChannel.java
+++ b/fluss-rpc/src/main/java/com/alibaba/fluss/rpc/netty/server/RequestChannel.java
@@ -21,8 +21,8 @@ import org.slf4j.LoggerFactory;
 
 import javax.annotation.concurrent.ThreadSafe;
 
+import java.util.concurrent.ArrayBlockingQueue;
 import java.util.concurrent.BlockingQueue;
-import java.util.concurrent.PriorityBlockingQueue;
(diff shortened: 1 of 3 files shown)


luoyuxia (approved):
LGTM! I'd like to verify in my env before merging it...

wuchong (approved):
LGTM.

wuchong:
@luoyuxia did you verify it in your env?

luoyuxia:
> @luoyuxia did you verify it in your env?

I think we can merge it first..

PR #277, opened 2024-12-27
[kv] Support version merge engine 

sunxiaojian:
### Purpose

Support version merge engine 

Linked issue: close https://github.com/alibaba/fluss/issues/213

### Tests

com.alibaba.fluss.connector.flink.sink.FlinkTableSinkITCase#testVersionMergeEngineWithTypeBigint
com.alibaba.fluss.connector.flink.sink.FlinkTableSinkITCase#testVersionMergeEngineWithTypeTimestamp
com.alibaba.fluss.client.table.FlussTableITCase#testMergeEngineWithVersion

### API and Format

No

### Documentation
 
No


sunxiaojian:
@wuchong CLA has been sent, PTAL

wuchong:
Thanks for the contribution @sunxiaojian ! Will take a look. cc @luoyuxia  as well.

sunxiaojian:
> Btw, the checkstyle is failed: https://github.com/alibaba/fluss/actions/runs/12515363835/job/34959455034?pr=277

fixed

fluss-common/src/main/java/com/alibaba/fluss/config/MergeEngine.java
```diff
@@ -0,0 +1,101 @@
...
+        switch (type) {
+            case FIRST_ROW:
+                return new MergeEngine(Type.FIRST_ROW);
+            case VERSION:
+                String column = options.get(ConfigOptions.TABLE_MERGE_ENGINE_VERSION_COLUMN);
+                if (column == null) {
+                    throw new IllegalArgumentException(
+                            "When the merge engine is set to version, the 'table.merge-engine.version.column' cannot be empty.");
+                }
+                return new MergeEngine(Type.VERSION, column);
```
luoyuxia: We'll need two checks for merge_engine_version_column:
1: the column is in the table
2: the data type of the column is supported as version column

fluss-server/src/main/java/com/alibaba/fluss/server/kv/mergeengine/VersionMergeEngineWrapper.java
```diff
@@ -0,0 +1,121 @@
...
+                            .getMillisecond()
+                    > oldRow.getTimestampNtz(fieldIndex, ((TimestampType) dataType).getPrecision())
+                            .getMillisecond();
+        } else if (dataType instanceof LocalZonedTimestampType) {
+            return newRow.getTimestampLtz(fieldIndex, ((TimestampType) dataType).getPrecision())
+                            .toEpochMicros()
+                    > oldRow.getTimestampLtz(fieldIndex, ((TimestampType) dataType).getPrecision())
+                            .toEpochMicros();
+        } else {
+            throw new FlussRuntimeException("Unsupported data type: " + dataType.asSummaryString());
```
luoyuxia: Validate the data type of version column while creating the table...

And I think `TINYINT`/`SMALLINT` should also be supported..
Maybe `DATE` type also can be supported?

fluss-server/src/main/java/com/alibaba/fluss/server/kv/mergeengine/MergeEngineWrapper.java
```diff
@@ -0,0 +1,45 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.alibaba.fluss.server.kv.mergeengine;
+
+import com.alibaba.fluss.record.KvRecord;
+
+/** Merge engine wrapper for table. */
+public interface MergeEngineWrapper {
```
luoyuxia: method writeRecord,getAppendedRecordCount, getLogOffset is an interface method, but only `AbstractMergeEngineWrapper`...And the overwrite method `update` by VersionMergeEngineWrapper is not an interface method.. Looks a little  strange to me...
And there's some duplicate code piece in method `update` between `AbstractMergeEngineWrapper` and `VersionMergeEngineWrapper`..

I'm wondering whether it'll be better to organize these classes as following:
1: 
Introduce an interface `RowMergeEngine`
```
public interface RowMergeEngine {

    /**
     * Upsert the old row with the new row. Return the row after upsert, null if the new row is ignored.
     *
     * @param oldRow the old row
     * @param newRow the new row
     */
    @Nullable
    BinaryRow upsertRow(BinaryRow oldRow, BinaryRow newRow);
}
```
2: And then implement it with a `VersionRowMergeEngine`:
```
public class VersionRowMergeEngine implements RowMergeEngine {
   @Nullable
    @Override
    public BinaryRow upsertRow(BinaryRow oldRow, BinaryRow newRow) {
        RowType rowType = schema.toRowType();

        if (checkNewRowVersion(mergeEngine, rowType, oldRow, newRow)) {
            return newRow;
        }
        return null;
    }
}
```

3: Then, we won't need to extract the put kv logic to `AbstractMergeEngineWrapper`, but hold a `RowMergeEngine` in KvTablet.. If `RowMergeEngine` is not null, we use 
RowMergeEngine to merge the row to a new row.. May be some thing like:
```
byte[] oldValue = getFromBufferOrKv(key);
// it's update
if (oldValue != null) {
  if (rowMergeEngine == null) {
    newRow = updateRow(
    oldRow, kvRecord.getRow(), partialUpdater);
  } else {
    newRow = rowMergeEngine.mergeRow(oldRow, kvRecord.getRow());
}

if (newRow != null) {
walBuilder.append(RowKind.UPDATE_BEFORE, oldRow);
walBuilder.append(RowKind.UPDATE_AFTER, newRow);
appendedRecordCount += 2;
....
}
```

fluss-server/src/main/java/com/alibaba/fluss/server/kv/mergeengine/VersionMergeEngineWrapper.java
```diff
@@ -0,0 +1,121 @@
...
+    // Check row version.
+    private boolean checkNewRowVersion(
+            MergeEngine mergeEngine, RowType rowType, BinaryRow oldRow, BinaryRow newRow) {
+        int fieldIndex = rowType.getFieldIndex(mergeEngine.getColumn());
+        if (fieldIndex == -1) {
+            return false;
+        }
+        DataType dataType = rowType.getTypeAt(fieldIndex);
+        if (dataType instanceof BigIntType) {
+            return newRow.getLong(fieldIndex) > oldRow.getLong(fieldIndex);
```
luoyuxia: What if the field value is null? We should use `InternalRow.FieldGetter` and handle null value...

fluss-server/src/test/java/com/alibaba/fluss/server/kv/KvTabletTest.java
```diff
@@ -124,7 +124,8 @@ void beforeEach() throws Exception {
                         conf,
```
luoyuxia: Add a ut for version merge engine in `KvTabletTest`...

fluss-connectors/fluss-connector-flink/src/main/java/com/alibaba/fluss/connector/flink/sink/FlinkTableSink.java
```diff
@@ -67,17 +68,21 @@ public class FlinkTableSink
     private boolean appliedUpdates = false;
     @Nullable private GenericRow deleteRow;
 
+    private final MergeEngine mergeEngine;
+
     public FlinkTableSink(
```
luoyuxia: Add validation to throw exception in class `FlinkTableSink` that update/delete/partial update is not supported in the version merge engine and a test to verify the exception should be thrown...

luoyuxia (commented):
@sunxiaojian Thanks for the pr... I left some comments... PTAL

sunxiaojian:
> @sunxiaojian Thanks for the pr... I left some comments... PTAL

@luoyuxia Thanks for the review. I will handle it as soon as possible

fluss-server/src/main/java/com/alibaba/fluss/server/kv/mergeengine/VersionMergeEngineWrapper.java
sunxiaojian: This should not be frequently used as a version column, but can be expanded later if necessary

fluss-connectors/fluss-connector-flink/src/main/java/com/alibaba/fluss/connector/flink/sink/FlinkTableSink.java
sunxiaojian: I think that it should support updates and partial updates if your update includes this version field

fluss-server/src/test/java/com/alibaba/fluss/server/kv/KvTabletTest.java
sunxiaojian: added

fluss-server/src/main/java/com/alibaba/fluss/server/kv/mergeengine/VersionMergeEngineWrapper.java
sunxiaojian: fixed

fluss-common/src/main/java/com/alibaba/fluss/config/MergeEngine.java
sunxiaojian: fixed

fluss-server/src/main/java/com/alibaba/fluss/server/kv/mergeengine/MergeEngineWrapper.java
sunxiaojian: thanks，Let me think about how to make better modifications first

fluss-connectors/fluss-connector-flink/src/main/java/com/alibaba/fluss/connector/flink/sink/FlinkTableSink.java
sunxiaojian: fixed

fluss-server/src/main/java/com/alibaba/fluss/server/kv/mergeengine/MergeEngineWrapper.java
sunxiaojian: @luoyuxia I have refactored the merge engine wrapper, and it should now have the most streamlined logic. If we abstract only part of the logic, it might become too fragmented.

sunxiaojian:
@luoyuxia PTAL again , thanks

fluss-common/src/main/java/com/alibaba/fluss/metadata/MergeEngine.java
```diff
@@ -18,22 +16,125 @@
...
+import com.alibaba.fluss.types.TimeType;
+import com.alibaba.fluss.types.TimestampType;
+
+import java.util.Map;
+import java.util.Objects;
+import java.util.Set;
 
-    private final String value;
+/** The merge engine for primary key table. */
+public class MergeEngine {
```
luoyuxia: Add `@since 0.6`

fluss-common/src/main/java/com/alibaba/fluss/metadata/MergeEngine.java
```diff
@@ -18,22 +16,125 @@
...
+
+    public static final Set<String> VERSION_SUPPORTED_DATA_TYPES =
+            Sets.newHashSet(
```
luoyuxia: nit: ImmutableSet.of(

fluss-common/src/main/java/com/alibaba/fluss/metadata/MergeEngine.java
```diff
@@ -18,22 +16,125 @@
...
+                    BigIntType.class.getName(),
+                    IntType.class.getName(),
+                    TimestampType.class.getName(),
+                    TimeType.class.getName(),
+                    LocalZonedTimestampType.class.getName());
+    private final Type type;
+    private final String column;
```
luoyuxia: add `@nullable` and to comment when the column will be null...

fluss-common/src/main/java/com/alibaba/fluss/metadata/MergeEngine.java
```diff
@@ -18,22 +16,125 @@
...
+        switch (type) {
+            case FIRST_ROW:
+                return new MergeEngine(Type.FIRST_ROW);
+            case VERSION:
+                String column = options.get(ConfigOptions.TABLE_MERGE_ENGINE_VERSION_COLUMN);
+                if (column == null) {
+                    throw new IllegalArgumentException(
+                            "When the merge engine is set to version, the 'table.merge-engine.version.column' cannot be empty.");
+                }
+                if (rowType != null) {
```
luoyuxia: we can do this kind of validation in `TableDescriptor` so that we can avoid passing `RowType` to just create a `MergeEngine` in here.  After remove the validation out, the `MergeEngine` now only has one method `public static MergeEngine create(Map<String, String> properties)` which is more clear...
And then the caller won't need to care about calling `public static @Nullable MergeEngine create(Configuration options, RowType rowType)` or `static MergeEngine create(Map<String, String> properties)`.

luoyuxia: Also, logically, create `MergeEngine` should not validate column, but attach it to a table, we should do the validation...

fluss-common/src/main/java/com/alibaba/fluss/metadata/MergeEngine.java
```diff
@@ -18,22 +16,125 @@
...
+    }
+
+    public static MergeEngine create(Map<String, String> properties, RowType rowType) {
+        return create(Configuration.fromMap(properties), rowType);
+    }
 
-    MergeEngine(String value) {
-        this.value = value;
+    public static MergeEngine create(Configuration options, RowType rowType) {
+        if (options == null) {
```
luoyuxia: When `options` is null? if never be null, don't need to check it..

fluss-common/src/main/java/com/alibaba/fluss/metadata/MergeEngine.java
```diff
@@ -18,22 +16,125 @@
...
+        return create(properties, null);
+    }
+
+    public static MergeEngine create(Map<String, String> properties, RowType rowType) {
+        return create(Configuration.fromMap(properties), rowType);
+    }
 
-    MergeEngine(String value) {
-        this.value = value;
+    public static MergeEngine create(Configuration options, RowType rowType) {
```
luoyuxia:
```suggestion
    public static @Nullable MergeEngine create(Configuration options, RowType rowType) {
```

fluss-common/src/main/java/com/alibaba/fluss/metadata/MergeEngine.java
```diff
@@ -18,22 +16,125 @@
...
+            return null;
+        }
+
+        switch (type) {
+            case FIRST_ROW:
+                return new MergeEngine(Type.FIRST_ROW);
+            case VERSION:
+                String column = options.get(ConfigOptions.TABLE_MERGE_ENGINE_VERSION_COLUMN);
+                if (column == null) {
+                    throw new IllegalArgumentException(
```
luoyuxia: nit:
```suggestion
                 throw new IllegalArgumentException(
                            String.format(
                                    "When the merge engine is set to version, the '%s' must be set.",
                                    ConfigOptions.TABLE_MERGE_ENGINE_VERSION_COLUMN.key()));
```

fluss-connectors/fluss-connector-flink/src/main/java/com/alibaba/fluss/connector/flink/sink/FlinkTableSink.java
```diff
@@ -75,7 +75,7 @@ public FlinkTableSink(
             RowType tableRowType,
             int[] primaryKeyIndexes,
             boolean streaming,
-            @Nullable MergeEngine mergeEngine) {
```
luoyuxia: Why remove this?

fluss-connectors/fluss-connector-flink/src/test/java/com/alibaba/fluss/connector/flink/sink/FlinkTableSinkITCase.java
```diff
@@ -787,6 +787,123 @@ void testUnsupportedStmtOnFirstRowMergeEngine() {
...
+        // test update a=4
+        tBatchEnv
+                .executeSql("UPDATE merge_engine_with_version SET b = 'v45', ts=1001 WHERE a = 4")
+                .await();
+        expectedRows = Arrays.asList("-U[4, v44, 1000]", "+U[4, v45, 1001]");
+        assertResultsIgnoreOrder(rowIter, expectedRows, false);
+
+        // test delete a=4
+        tBatchEnv.executeSql("delete from merge_engine_with_version WHERE a = 4").await();
+        expectedRows = Arrays.asList("-D[4, v45, 1001]");
```
luoyuxia: nit: 
```suggestion
        expectedRows = Collections.singletonList("-D[4, v45, 1001]");
```

fluss-client/src/test/java/com/alibaba/fluss/client/table/FlussTableITCase.java
```diff
@@ -910,6 +911,55 @@ void testFirstRowMergeEngine() throws Exception {
                                 .getRow();
                 assertThatRow(gotRow).withSchema(rowType).isEqualTo(expectedRows.get(id));
             }
+            // check scan change log
```
luoyuxia: I'd like to remove the piece of code since the change log behavior has been verify by `FlinkTableSinkITCase#testFirstRowMergeEngine`...Add it in here make the test code tediously

fluss-client/src/test/java/com/alibaba/fluss/client/table/FlussTableITCase.java
```diff
@@ -910,6 +911,55 @@ void testFirstRowMergeEngine() throws Exception {
...
+                TableDescriptor.builder()
+                        .schema(DATA3_SCHEMA_PK)
+                        .property(ConfigOptions.TABLE_MERGE_ENGINE, MergeEngine.Type.VERSION)
+                        .property(ConfigOptions.TABLE_MERGE_ENGINE_VERSION_COLUMN, "b")
+                        .build();
+        RowType rowType = DATA3_SCHEMA_PK.toRowType();
+        createTable(DATA3_TABLE_PATH_PK, tableDescriptor, false);
+
+        int rows = 3;
+        try (Table table = conn.getTable(DATA3_TABLE_PATH_PK)) {
```
luoyuxia, line 1073: I'd like to suggest to change to the following to simplify test code
```
try (Table table = conn.getTable(DATA3_TABLE_PATH_PK)) {
            // put rows.
            UpsertWriter upsertWriter = table.getUpsertWriter();
            List<ScanRecord> expectedScanRecords = new ArrayList<>(rows);
            // init rows.
            for (int row = 0; row < rows; row++) {
                upsertWriter.upsert(compactedRow(rowType, new Object[] {row, 1000L}));
                expectedScanRecords.add(
                        new ScanRecord(compactedRow(rowType, new Object[] {row, 1000L})));
            }
            // update row if id=0 and version < 1000L, will not update
            upsertWriter.upsert(compactedRow(rowType, new Object[] {0, 999L}));

            // update if version> 1000L
            upsertWriter.upsert(compactedRow(rowType, new Object[] {1, 1001L}));
            // update_before record, don't care about offset/timestamp
            expectedScanRecords.add(
                    new ScanRecord(
                            -1,
                            -1,
                            RowKind.UPDATE_BEFORE,
                            compactedRow(rowType, new Object[] {1, 1000L})));
            // update_after record
            expectedScanRecords.add(
                    new ScanRecord(
                            -1,
                            -1,
                            RowKind.UPDATE_AFTER,
                            compactedRow(rowType, new Object[] {1, 1001L})));
            rows = rows + 2;

            upsertWriter.flush();

            // check scan change log
            LogScanner logScanner = table.getLogScanner(new LogScan());
            logScanner.subscribeFromBeginning(0);

            List<ScanRecord> actualLogRecords = new ArrayList<>(rows);
            while (actualLogRecords.size() < rows) {
                ScanRecords scanRecords = logScanner.poll(Duration.ofSeconds(1));
                scanRecords.forEach(actualLogRecords::add);
            }

            assertThat(actualLogRecords).hasSize(rows);
            for (int i = 0; i < rows; i++) {
                ScanRecord actualScanRecord = actualLogRecords.get(i);
                ScanRecord expectedRecord = expectedScanRecords.get(i);
                assertThat(actualScanRecord.getRowKind()).isEqualTo(expectedRecord.getRowKind());
                assertThatRow(actualScanRecord.getRow())
                        .withSchema(rowType)
                        .isEqualTo(expectedRecord.getRow());
            }
        }
```

fluss-connectors/fluss-connector-flink/src/test/java/com/alibaba/fluss/connector/flink/sink/FlinkTableSinkITCase.java
```diff
@@ -787,6 +787,123 @@ void testUnsupportedStmtOnFirstRowMergeEngine() {
...
+
+        // test update a=4
+        tBatchEnv
+                .executeSql("UPDATE merge_engine_with_version SET b = 'v45', ts=1001 WHERE a = 4")
+                .await();
+        expectedRows = Arrays.asList("-U[4, v44, 1000]", "+U[4, v45, 1001]");
+        assertResultsIgnoreOrder(rowIter, expectedRows, false);
+
+        // test delete a=4
+        tBatchEnv.executeSql("delete from merge_engine_with_version WHERE a = 4").await();
```
luoyuxia: Won't delete throw unsupported exception?

fluss-server/src/main/java/com/alibaba/fluss/server/kv/mergeengine/MergeEngineWrapper.java
luoyuxia: TBH, I'd like to still keep my mind to not to wrap all behavior of writing a record to `RowMergeEngine`

But as we know, different developers have different code preferences, I'd like to share my thoughts about it.

- Logically, `RowMergeEngine` just cares about how to merge rows instead of decoding row, write to kv prewrite buffer, increase `logOffset`, etc.
- In design of merge engine in table, if no merge engine is specified,  we make it null. But in here, we use `DefaultRowMergeEngine` to handle the case of no merge engine.. That looks a little of weird to me.
- If we want to introduce a new merge engine, we **have to**  remember to call `super.update(key, oldRow, newRow);`, that's error prone to me..
- Extract all logic to `RowMergeEngine` make it's hard to track the flow of writing record . And it also makes the arguments passed to create `RowMergeEngine` hard to understand.. I'm not a fan of using too many arguments in class constructors.

fluss-server/src/main/java/com/alibaba/fluss/server/kv/mergeengine/VersionRowMergeEngine.java
```diff
@@ -0,0 +1,138 @@
...
+        int fieldIndex = rowType.getFieldIndex(mergeEngine.getColumn());
+        if (fieldIndex == -1) {
+            throw new IllegalArgumentException(
+                    String.format(
+                            "When the merge engine is set to version, the column %s does not exist.",
+                            mergeEngine.getColumn()));
+        }
+        InternalRow.FieldGetter fieldGetter = currentFieldGetters[fieldIndex];
+        Object oldValue = fieldGetter.getFieldOrNull(oldRow);
+        if (oldValue == null) {
```
luoyuxia: If we don't allow the field value to be null, the field's data type should also be restricted to `not null`. But I don't think we should throw exception directly if meet a null value..
Paimon's[ sequence field ](https://paimon.apache.org/docs/0.8/primary-key-table/sequence-rowkind/)also supports null value.

fluss-server/src/main/java/com/alibaba/fluss/server/kv/mergeengine/VersionRowMergeEngine.java
```diff
@@ -0,0 +1,138 @@
...
+                            mergeEngine.getColumn()));
+        }
+
+        DataType dataType = rowType.getTypeAt(fieldIndex);
+
+        if (dataType instanceof BigIntType) {
+            return (Long) newValue > (Long) oldValue;
+        } else if (dataType instanceof IntType) {
+            return (Integer) newValue > (Integer) oldValue;
+        } else if (dataType instanceof TimestampType || dataType instanceof TimeType) {
```
luoyuxia: If is `TimeType`, the value should be instance of Integer

fluss-connectors/fluss-connector-flink/src/main/java/com/alibaba/fluss/connector/flink/sink/FlinkTableSink.java
luoyuxia: For UPDATE, the behavior may be strange. The row may be changed to target row if the version field is not greater than the old version field.. DELETE statement can work, but I'm not sure whether to support it if update is not supported. 

Seems partial update can be supported if the column to be updated contains version column. But I'm not sure about supporting it since it will always require version column.

@wuchong What do you think of it? Should update/delete/partial update should be supported in version merge engine?

fluss-common/src/main/java/com/alibaba/fluss/metadata/MergeEngine.java
sunxiaojian: > Also, logically, create `MergeEngine` should not validate column, but attach it to a table, we should do the validation...

This seems to contradict the previous comment

fluss-connectors/fluss-connector-flink/src/main/java/com/alibaba/fluss/connector/flink/sink/FlinkTableSink.java
wuchong: Thanks @sunxiaojian and @luoyuxia , IMO, merge engine is a mechanism to merge **append-only messages** in some strategies. For example, LSM tree is a merge engine to merge logs where delete is an append message in the logs, so does [ClickHouse ReplacingMergeTree]( https://clickhouse.com/docs/en/engines/table-engines/mergetree-family/replacingmergetree). Therefore, merge engine is conflict with UPDATE/DELETE statement which directly works on the merged view. Maybe we can support that in the future, but let's keep it simple and semantics clean at the beginning and support it when necessary. 

That doesn't mean we don't support UPDATE/DELETE on merge engine tables, users still can express UPDATE/DELETE with `INSERT` statement (and I think that is enough to cover most of needs). For deletions on "version merge engine", we need to support defining `is_deleted` column just like how ClickHouse does.

fluss-server/src/main/java/com/alibaba/fluss/server/kv/mergeengine/VersionRowMergeEngine.java
sunxiaojian: If no exception is thrown, there are two scenarios:
1.If the old value is null, simply overwrite it with the new value;
2.If the new value is empty, ignore it directly;

If that's the case, I think you're right

fluss-connectors/fluss-connector-flink/src/main/java/com/alibaba/fluss/connector/flink/sink/FlinkTableSink.java
sunxiaojian: ok, I got, thanks

fluss-server/src/main/java/com/alibaba/fluss/server/kv/mergeengine/VersionRowMergeEngine.java
luoyuxia: Yes, that's what I'm thinking about.

sunxiaojian:
@luoyuxia PTAL again, thanks.

fluss-common/src/main/java/com/alibaba/fluss/metadata/MergeEngine.java
```diff
@@ -18,22 +16,95 @@
...
+        }
+        switch (type) {
+            case FIRST_ROW:
+                return new MergeEngine(Type.FIRST_ROW);
+            case VERSION:
+                String column = options.get(ConfigOptions.TABLE_MERGE_ENGINE_VERSION_COLUMN);
+                if (column == null) {
+                    throw new IllegalArgumentException(
+                            String.format(
+                                    "When the merge engine is set to version, the '%s' must be set.",
```
luoyuxia: Sorry for I missing the word `option`...
```suggestion
                                    "When the merge engine is set to version, the option '%s' must be set.",
```

fluss-server/src/main/java/com/alibaba/fluss/server/kv/KvTablet.java
```diff
@@ -267,7 +270,7 @@ public LogAppendInfo putAsLeader(
                                 KvRecordReadContext.createReadContext(kvFormat, fieldTypes);
                         ValueDecoder valueDecoder =
                                 new ValueDecoder(readContext.getRowDecoder(schemaId));
-
+                        RowMergeEngine rowMergeEngine = getRowMergeEngine(schema);
```
luoyuxia: Add a todo comment to reuse `RowMergeEngine` or reuse it in this pr...

fluss-server/src/main/java/com/alibaba/fluss/server/kv/KvTablet.java
```diff
@@ -281,12 +284,17 @@ public LogAppendInfo putAsLeader(
...
-                                        continue;
-                                    }
+
                                     BinaryRow oldRow = valueDecoder.decodeValue(oldValue).row;
                                     BinaryRow newRow = deleteRow(oldRow, partialUpdater);
+                                    if (rowMergeEngine.shouldSkipDeletion(newRow)) {
+                                        continue;
+                                    }
+                                    newRow = rowMergeEngine.merge(oldRow, newRow);
+                                    if (newRow == null) {
```
luoyuxia: warning from my ide `Duplicate condition 'newRow == null' `

fluss-server/src/main/java/com/alibaba/fluss/server/kv/KvTablet.java
```diff
@@ -281,12 +284,17 @@ public LogAppendInfo putAsLeader(
                                             "The specific key can't be found in kv tablet although the kv record is for deletion, "
                                                     + "ignore it directly as it doesn't exist in the kv tablet yet.");
                                 } else {
-                                    if (mergeEngine == MergeEngine.FIRST_ROW) {
-                                        // if the merge engine is first row, skip the deletion
-                                        continue;
-                                    }
+
```
luoyuxia: nit: remove the empty line...

fluss-server/src/main/java/com/alibaba/fluss/server/kv/mergeengine/VersionRowMergeEngine.java
```diff
@@ -0,0 +1,101 @@
...
+        }
+    }
+
+    @Override
+    public BinaryRow merge(BinaryRow oldRow, BinaryRow newRow) {
+        int fieldIndex = rowType.getFieldIndex(mergeEngine.getColumn());
+        if (fieldIndex == -1) {
+            throw new IllegalArgumentException(
+                    String.format(
+                            "When the merge engine is set to version, the column %s does not exist.",
```
luoyuxia: nit: 
```suggestion
                            "The merge engine is set to version, but the version column %s does not exist.",
```

fluss-server/src/main/java/com/alibaba/fluss/server/kv/mergeengine/VersionRowMergeEngine.java
```diff
@@ -0,0 +1,101 @@
...
+                            mergeEngine.getColumn()));
+        }
+        InternalRow.FieldGetter fieldGetter = currentFieldGetters[fieldIndex];
+        Object oldValue = fieldGetter.getFieldOrNull(oldRow);
+        Object newValue = fieldGetter.getFieldOrNull(newRow);
+        // If the old value is null, simply overwrite it with the new value.
+        if (oldValue == null) return newRow;
```
luoyuxia:
```suggestion
        if (oldValue == null) {
        return newRow;
        }
```

fluss-server/src/main/java/com/alibaba/fluss/server/kv/mergeengine/VersionRowMergeEngine.java
```diff
@@ -0,0 +1,101 @@
...
+        // If the new value is empty, ignore it directly.
+        if (newValue == null) return null;
```
luoyuxia: dito

fluss-server/src/main/java/com/alibaba/fluss/server/kv/mergeengine/VersionRowMergeEngine.java
```diff
@@ -0,0 +1,101 @@
...
+        if (dataType instanceof TimestampType) {
+            return (left, right) ->
+                    ((TimestampNtz) left).getMillisecond()
+                            > ((TimestampNtz) right).getMillisecond();
+        }
+        if (dataType instanceof LocalZonedTimestampType) {
+            return (left, right) ->
+                    ((TimestampLtz) left).toEpochMicros() > ((TimestampLtz) right).toEpochMicros();
+        }
+        throw new FlussRuntimeException("Unsupported data type: " + dataType.asSummaryString());
```
luoyuxia: nit:
```suggestion
        throw new UnsupportedOperationException("Unsupported data type: " + dataType);
```

fluss-server/src/main/java/com/alibaba/fluss/server/kv/mergeengine/VersionRowMergeEngine.java
```diff
@@ -0,0 +1,101 @@
...
+    @Override
+    public BinaryRow merge(BinaryRow oldRow, BinaryRow newRow) {
+        int fieldIndex = rowType.getFieldIndex(mergeEngine.getColumn());
+        if (fieldIndex == -1) {
+            throw new IllegalArgumentException(
+                    String.format(
+                            "When the merge engine is set to version, the column %s does not exist.",
+                            mergeEngine.getColumn()));
+        }
+        InternalRow.FieldGetter fieldGetter = currentFieldGetters[fieldIndex];
```
luoyuxia: nit: The `fieldGetter` for version column can be initialized in  constructing method `VersionRowMergeEngine(Schema schema, MergeEngine mergeEngine)`..

fluss-server/src/test/java/com/alibaba/fluss/server/kv/KvTabletTest.java
```diff
@@ -600,6 +603,80 @@ void testFirstRowMergeEngine(@TempDir File tempLogDir, @TempDir File tmpKvDir)
         checkEqual(actualLogRecords, expectedLogs);
     }
 
+    @Test
+    void testVersionRowMergeEngine(@TempDir File tempLogDir, @TempDir File tmpKvDir)
```
luoyuxia: Also verify null value can be handled properly in this ut..

fluss-common/src/main/java/com/alibaba/fluss/metadata/TableDescriptor.java
```diff
@@ -281,7 +281,7 @@ public boolean isDataLakeEnabled() {
     }
 
     public @Nullable MergeEngine getMergeEngine() {
-        return configuration().get(ConfigOptions.TABLE_MERGE_ENGINE);
+        return MergeEngine.create(properties);
```
luoyuxia: I mean in [here](https://github.com/alibaba/fluss/pull/277/files#diff-99c843b9b424c27326869c2bf3692a26f3a2421a95c8db75769d525ff5f6d8a1L138) to check the data type  of version column of version merge engine..

fluss-server/src/main/java/com/alibaba/fluss/server/kv/mergeengine/RowMergeEngine.java
```diff
@@ -0,0 +1,31 @@
...
+ * limitations under the License.
+ */
+package com.alibaba.fluss.server.kv.mergeengine;
+
+import com.alibaba.fluss.row.BinaryRow;
+
+/**
+ * The row merge engine for primary key table.
+ *
+ * @since 0.6
```
luoyuxia: nit: no need `@since 0.6` as it's not public api.

fluss-server/src/main/java/com/alibaba/fluss/server/kv/mergeengine/FirstRowMergeEngine.java
```diff
@@ -0,0 +1,35 @@
...
+ * @since 0.6
+ */
+public class FirstRowMergeEngine implements RowMergeEngine {
+    @Override
+    public BinaryRow merge(BinaryRow oldRow, BinaryRow newRow) {
+        return oldRow == null ? newRow : null;
+    }
+
+    @Override
+    public boolean shouldSkipDeletion(BinaryRow newRow) {
```
luoyuxia: FirstRowMergeEngine should always return true whenever the `newRow` is null or not..
And so that, this method won't need argument `BinaryRow newRow`..

fluss-server/src/main/java/com/alibaba/fluss/server/kv/KvTablet.java
```diff
@@ -281,12 +284,17 @@ public LogAppendInfo putAsLeader(
...
                                     BinaryRow oldRow = valueDecoder.decodeValue(oldValue).row;
                                     BinaryRow newRow = deleteRow(oldRow, partialUpdater);
+                                    if (rowMergeEngine.shouldSkipDeletion(newRow)) {
```
luoyuxia: Seems we can check `shouldSkipDeletion` before [here](https://github.com/alibaba/fluss/pull/277/files#diff-700b6858eb7857b9efe6ea761a22c9ded3863a072e12510d995f5f221f7fcd8bL277) to avoid fetch old row from rocksdb.

fluss-server/src/main/java/com/alibaba/fluss/server/kv/KvTablet.java
```diff
@@ -372,6 +380,23 @@ public LogAppendInfo putAsLeader(
                 });
     }
 
+    private RowMergeEngine getRowMergeEngine(Schema schema) {
```
luoyuxia: I append a commit about RowMergeEngine in my own branch https://github.com/luoyuxia/fluss/commit/67f7605b5d08dc5228271df43d6264a2080e3998
There're  two changes:
1. If no merge engine, I set the RowMergeEngine to null instead of creating a new RowMergeEngine.. Since using a RowMergeEngine for tablet without setting `merge engine` look strange to me.
2. if it's first row merge engine, don't to deserize the old value

You can have a look..Hope it can help

luoyuxia (commented):
@sunxiaojian Thanks for updating again... I left some comments. PTAL. Should be look good to me in next iterations..

sunxiaojian:
@luoyuxia About 1 ，Can refer to Paimon's default merge engine Deduplication (https://paimon.apache.org/docs/master/primary-key-table/merge-engine/overview/), which is similar to ours. I do not recommend returning null.

fluss-server/src/main/java/com/alibaba/fluss/server/kv/KvTablet.java
sunxiaojian: @luoyuxia About 1 ，Can refer to Paimon's default merge engine Deduplication (https://paimon.apache.org/docs/master/primary-key-table/merge-engine/overview/), which is similar to ours. I do not recommend returning null. By default, it can be understood as the last row merge engine. @wuchong What do you think of it?

sunxiaojian (commented):
@luoyuxia Thanks for  review ... I left some comments. PTAL.

fluss-common/src/main/java/com/alibaba/fluss/metadata/TableDescriptor.java
```diff
@@ -281,7 +283,27 @@ public boolean isDataLakeEnabled() {
     }
 
     public @Nullable MergeEngine getMergeEngine() {
-        return configuration().get(ConfigOptions.TABLE_MERGE_ENGINE);
+        MergeEngine mergeEngine = MergeEngine.create(properties);
+        if (mergeEngine != null) {
+            if (mergeEngine.getType() == MergeEngine.Type.VERSION) {
+                String column = mergeEngine.getColumn();
```
luoyuxia: Two comments about validate in this method
1. Actually, we only need to validate during creating the table. If validate in this method, every time we get a `TableDescriptor`  from server and try to get the merge engine of it, we have to do the validation.. Althoght it's not heavy, but unnesscassy..

2. If we may well support different data types for version in the future, I'd suggest to validate it in server side so that if support other data types, user won't need to bump it's client version. We also do some validation for partitioned table in server side. You can see it in method `sanityCheckPartitionedTable`

fluss-server/src/main/java/com/alibaba/fluss/server/kv/KvTablet.java
luoyuxia: TBH, althogth we haven't consider the default behavior of kv as `Deduplication` merge engine in our doc  like Paimon and code design,  I'm not too objection to condier the  default behaviorv as another special merge engine in here since now it's only for internal use.

But I think consider the default behavior as a specical merge engine may make the code logic complex..May be I'm wong.  

For example,  in update case, default behavior will need to consider partial update/delete, but the first_row, version merge engine don't need, so you will need to introduce method `ignoreDelete` and `ignorePartialUpdate` in `RowMergeEngine`. Logically, RowMergeEngine should only care how how to mere two row, the logical of `ignoreDelete` and `ignorePartialUpdate` should be considered by caller..


Also, If we decide  to consider the default as `Deduplication` ,  I'd like to suggest use a new class naed `DeduplicateRowMergeEngine` or `KeepLastRowMergeEngine`  or any thing else instead of using anonymous class

fluss-server/src/main/java/com/alibaba/fluss/server/kv/KvTablet.java
```diff
@@ -267,12 +270,31 @@ public LogAppendInfo putAsLeader(
...
                         ValueDecoder valueDecoder =
                                 new ValueDecoder(readContext.getRowDecoder(schemaId));
-
+                        RowMergeEngine rowMergeEngine = getRowMergeEngine(schema);
+                        /**
+                         * In theory, once configured, it can be ignored directly. In the future, we
+                         * can provide configurations to choose between ignoring or throwing
+                         * exceptions directly.
+                         */
+                        if (partialUpdater != null && rowMergeEngine.ignorePartialUpdate()) {
```
luoyuxia: Don't ignore partial update, if merge engine don't support partial update, throw exception directly instead of just ignoring it. It's not a expected behavior, ignoring it will hide the exception and make user mistake the rows are written to server..

luoyuxia (commented):
@sunxiaojian Thanks again.. Left minor comments..

fluss-server/src/main/java/com/alibaba/fluss/server/kv/KvTablet.java
sunxiaojian: @luoyuxia Thanks review again,    I have always believed that the merge engine should manage what to ignore, which RowKinds to support, and what field types the version column should accept within its own implementation class. If we leave these judgments to the caller, as the types of merge engines increase, the number of conditions will also grow, leading to decreased code readability. Therefore, I still think we should retain 'ignoreDelete' and 'ignorePartialUpdate', or move PartialUpdate validation to the merge engine implementation.

fluss-server/src/main/java/com/alibaba/fluss/server/coordinator/CoordinatorService.java
```diff
@@ -161,13 +164,39 @@ public CompletableFuture<CreateTableResponse> createTable(CreateTableRequest req
...
         return CompletableFuture.completedFuture(response);
     }
 
+    private void checkMergeEngine(MergeEngine mergeEngine, Schema schema) {
+        if (mergeEngine.getType() == MergeEngine.Type.VERSION) {
+            String column = mergeEngine.getColumn();
+            RowType rowType = schema.toRowType();
+            int fieldIndex = rowType.getFieldIndex(column);
+            if (fieldIndex == -1) {
+                throw new IllegalArgumentException(
```
luoyuxia: throw `InvalidTableException` just like what we do while validating partitioned table..
And add a test to verify invalidate version column will throw exception..

fluss-server/src/test/java/com/alibaba/fluss/server/kv/KvTabletTest.java
luoyuxia: don't forget my comments in here.

luoyuxia (commented):
@sunxiaojian Thanks.. I have no further comment exception for the argument in [here](https://github.com/alibaba/fluss/pull/277/files#r1909118078) about the abstract of `RowMergeEngine` in KvTablet. Btw, next time, please **append** a new commit instead of squash all into a commit so that reviewer can track what changes since last review instead of go through all files..
@wuchong Could you please have a final check?

fluss-server/src/main/java/com/alibaba/fluss/server/coordinator/CoordinatorService.java
sunxiaojian: fixed

fluss-server/src/test/java/com/alibaba/fluss/server/kv/KvTabletTest.java
sunxiaojian: > don't forget my comments in here.

fixed

wuchong:
Thanks for the awesome work @sunxiaojian @luoyuxia ! I will go through the code tomorrow to have a final check!

swuferhong:
Hi, @sunxiaojian For some comments that have already been resolved and are no longer in dispute, can you click "resolve conversion" on your side? There are too many comments, and I don't know which ones have been resolved finally.

sunxiaojian:
> Hi, @sunxiaojian For some comments that have already been resolved and are no longer in dispute, can you click "resolve conversion" on your side? There are too many comments, and I don't know which ones have been resolved finally.

ok

sunxiaojian:
> Hi, @sunxiaojian For some comments that have already been resolved and are no longer in dispute, can you click "resolve conversion" on your side? There are too many comments, and I don't know which ones have been resolved finally.

@swuferhong  Fixed, PTAL again, thanks.

fluss-connectors/fluss-connector-flink/src/test/java/com/alibaba/fluss/connector/flink/sink/FlinkTableSinkITCase.java
```diff
@@ -826,6 +826,111 @@ void testUnsupportedStmtOnFirstRowMergeEngine() {
                         tablePath);
     }
 
+    @Test
+    void testUnsupportedStmtOnVersionRowMergeEngine() {
```
wuchong:
```suggestion
    void testUnsupportedStmtOnVersionMergeEngine() {
```

fluss-connectors/fluss-connector-flink/src/test/java/com/alibaba/fluss/connector/flink/sink/FlinkTableSinkITCase.java
```diff
@@ -826,6 +826,111 @@ void testUnsupportedStmtOnFirstRowMergeEngine() {
...
+    @Test
+    void testVersionMergeEngineWithTypeBigint() throws Exception {
+        tEnv.executeSql(
+                "create table merge_engine_with_version (a int not null primary key not enforced,"
+                        + " b string, ts bigint) with('table.merge-engine' = 'version','table.merge-engine.version.column' = 'ts')");
+        tEnv.executeSql(
+                "create table log_sink (a int not null primary key not enforced, b string, ts bigint)");
+
+        JobClient insertJobClient =
+                tEnv.executeSql("insert into log_sink select * from merge_engine_with_version")
```
wuchong: why insert into another pk table? If we want to subscribe the changelogs of `merge_engine_with_version`, we can directly iterate `tEnv.executeSql("select * from merge_engine_with_version").collect()` in streaming mode.

fluss-connectors/fluss-connector-flink/src/test/java/com/alibaba/fluss/connector/flink/sink/FlinkTableSinkITCase.java
```diff
@@ -826,6 +826,111 @@ void testUnsupportedStmtOnFirstRowMergeEngine() {
...
+
+        // id=1 not update
+        List<String> expectedRows =
+                Arrays.asList("+I[1, v1, 1000]", "+I[2, v2, 1000]", "+I[3, v3, 1000]");
+
+        assertResultsIgnoreOrder(rowIter, expectedRows, false);
+
+        // insert again, update id=3
+        tEnv.executeSql(
+                        "insert into merge_engine_with_version (a, b, ts) VALUES (3, 'v33', 1001), (4, 'v44', 1000)")
```
wuchong: add a test data `2, v22, 1000` that has the same version to verify it should also be updated.

fluss-connectors/fluss-connector-flink/src/test/java/com/alibaba/fluss/connector/flink/sink/FlinkTableSinkITCase.java
```diff
@@ -826,6 +826,111 @@ void testUnsupportedStmtOnFirstRowMergeEngine() {
...
+                "create table merge_engine_with_version (a int not null primary key not enforced,"
+                        + " b string, ts TIMESTAMP(3)) with('table.merge-engine' = 'version','table.merge-engine.version.column' = 'ts')");
+
+        // insert once
+        tEnv.executeSql(
+                        "INSERT INTO merge_engine_with_version (a, b, ts) VALUES "
+                                + "(1, 'v1', TIMESTAMP '2024-12-27 12:00:00.123'), "
+                                + "(2, 'v2', TIMESTAMP '2024-12-27 12:00:00.123'), "
+                                + "(1, 'v11', TIMESTAMP '2024-12-27 11:59:59.123'), "
+                                + "(3, 'v3', TIMESTAMP '2024-12-27 12:00:00.123');")
```
wuchong: Add a test data that has the same timestamp to verify the later one should replace the previous one. 
```suggestion
                                + "(3, 'v33', TIMESTAMP '2024-12-27 12:00:00.123');")
                                + "(3, 'v3', TIMESTAMP '2024-12-27 12:00:00.123');")
```

fluss-server/src/main/java/com/alibaba/fluss/server/kv/mergeengine/VersionRowMergeEngine.java
```diff
@@ -0,0 +1,119 @@
...
+ */
+public class VersionRowMergeEngine implements RowMergeEngine {
+
+    public static final Set<String> VERSION_MERGE_ENGINE_SUPPORTED_DATA_TYPES =
+            ImmutableSet.of(
+                    BigIntType.class.getName(),
+                    IntType.class.getName(),
+                    TimestampType.class.getName(),
+                    TimeType.class.getName(),
+                    LocalZonedTimestampType.class.getName());
```
wuchong: Use `DataTypeRoot` instead.

fluss-connectors/fluss-connector-flink/src/main/java/com/alibaba/fluss/connector/flink/sink/FlinkTableSink.java
```diff
@@ -123,19 +123,22 @@ public SinkRuntimeProvider getSinkRuntimeProvider(Context context) {
...
-                } else if (mergeEngine == MergeEngine.FIRST_ROW) {
-                    throw new ValidationException(
-                            String.format(
-                                    "Table %s uses the '%s' merge engine which does not support partial updates. Please make sure the "
-                                            + "number of specified columns in INSERT INTO matches columns of the Fluss table.",
-                                    tablePath, MergeEngine.FIRST_ROW));
+                }
+                if (mergeEngine != null) {
+                    if (mergeEngine.getType() == MergeEngine.Type.FIRST_ROW
+                            || mergeEngine.getType() == MergeEngine.Type.VERSION) {
```
wuchong: All the merge engines don't support partial update, don't need to enumerate all the merge engines.

fluss-connectors/fluss-connector-flink/src/main/java/com/alibaba/fluss/connector/flink/sink/FlinkTableSink.java
```diff
@@ -311,12 +314,14 @@ private void validateUpdatableAndDeletable() {
...
         }
-
-        if (mergeEngine == MergeEngine.FIRST_ROW) {
-            throw new UnsupportedOperationException(
-                    String.format(
-                            "Table %s uses the '%s' merge engine which does not support DELETE or UPDATE statements.",
-                            tablePath, MergeEngine.FIRST_ROW));
+        if (mergeEngine != null) {
+            if (mergeEngine.getType() == MergeEngine.Type.FIRST_ROW
+                    || mergeEngine.getType() == MergeEngine.Type.VERSION) {
```
wuchong: All the merge engines don't support delete and update, don't need to enumerate all the merge engines.

fluss-common/src/main/java/com/alibaba/fluss/metadata/MergeEngine.java
```diff
@@ -18,22 +16,95 @@
...
+import java.util.Objects;
+
 /**
  * The merge engine for primary key table.
  *
  * @since 0.6
  */
-public enum MergeEngine {
-    FIRST_ROW("first_row");
+public class MergeEngine {
```
wuchong: It is unnecessary to combine merge engine type and version column into a structure. First of all, the `column` member is very confusing and doesn't apply to `first_row` and `aggregate` merge engines. Secondly, you have introduced a runtime `RowMergeEngine` which already combines the merge engine type and version column into one class during runtime, so this class is necessary. I would prefer to keep it as usual. 

Maybe we can rename it to `MergeEngineType` to be more specific. 


```java
/**
 * The merge engine for the primary key table.
 *
 * <p>A primary key table with a merge engine is a special kind of table, called "merge table". Fluss
 * provides 3 kinds of table: "primary key table", "log table", and "merge table". Merge table is a
 * primary key table that has a primary key definition but doesn't directly UPDATE and DELETE rows in
 * the table, and instead, it merges the append rows into a new data set according to the defined
 * {@link MergeEngineType}. Therefore, it doesn't support direct UPDATE (also partial-update) and
 * DELETE operations and only supports INSERT or APPEND operations.
 *
 * <p>Note: A primary key table doesn't have a merge engine by default.
 *
 * @since 0.6
 */
public enum MergeEngineType {
    FIRST_ROW("first_row"),
    VERSION("version");
    // TODO: Support AGGREGATE("aggregate") in the future (#212)

    private final String value;

    MergeEngineType(String value) {
        this.value = value;
    }

    @Override
    public String toString() {
        return value;
    }
}
```

fluss-server/src/main/java/com/alibaba/fluss/server/kv/mergeengine/RowMergeEngine.java
```diff
@@ -0,0 +1,32 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.alibaba.fluss.server.kv.mergeengine;
+
+import com.alibaba.fluss.row.BinaryRow;
+
+/** The row merge engine for primary key table. */
+public interface RowMergeEngine {
```
wuchong: I think having an abstraction to unify the merging behavior of deduplication, first row, version, and aggregate is a good idea. 

However, I would not name this "merge engine".  Because a table with merge engine should only **merge append rows**, and shouldn't support delete and update, so a "primary key table" doesn't have merge engine by default. A "primary key table" with "merge engine" will turn into "merge table". 

What we need for the unification of merging behavior is just how to merge new row with the existing row. We can call this interface `RowMerger` and it shouldn't contain `ignoreDelete()` and `ignorePartialUpdate()`. The interface can be:

```java
/** A merging interface defines how to merge a new row with existing row. */
public interface RowMerger {

    /**
     * Merge the old row with the new row.
     *
     * @param oldRow the old row
     * @param newRow the new row
     * @return the merged row, or null if the row is deleted. If the returned row is the same to the
     *     old row, then nothing happens to the row (no update, no delete).
     */
    @Nullable
    BinaryRow merge(BinaryRow oldRow, BinaryRow newRow);
}
```

And we can implement `FirstRowRowMerger`, `VersionRowMerger`, `DeduplicateRowMerger`, `PartialUpdateRowMerger`, `AggregateRowMerger`.

fluss-server/src/main/java/com/alibaba/fluss/server/kv/KvTablet.java
```diff
@@ -267,12 +271,27 @@ public LogAppendInfo putAsLeader(
                                 KvRecordReadContext.createReadContext(kvFormat, fieldTypes);
                         ValueDecoder valueDecoder =
                                 new ValueDecoder(readContext.getRowDecoder(schemaId));
-
+                        RowMergeEngine rowMergeEngine = getRowMergeEngine(schema);
```
wuchong: This is a heavy operation, the row merger should be initialized when kv tablet is setup. And a dynamic `PartialUpdateRowMerger` will be used if the `put` is a partial update.

fluss-server/src/main/java/com/alibaba/fluss/server/kv/KvTablet.java
```diff
@@ -267,12 +271,27 @@ public LogAppendInfo putAsLeader(
...
+                                            "The current merge engine %s does not support partial updates.",
+                                            mergeEngine.getType()));
+                        }
                         int appendedRecordCount = 0;
                         for (KvRecord kvRecord : kvRecords.records(readContext)) {
                             byte[] keyBytes = BytesUtils.toArray(kvRecord.getKey());
                             KvPreWriteBuffer.Key key = KvPreWriteBuffer.Key.of(keyBytes);
                             if (kvRecord.getRow() == null) {
+                                // currently, all supported merge engine will ignore delete row.
+                                if (rowMergeEngine.ignoreDelete()) {
```
wuchong: We still need the nullable `MergeEngineType` in the `KvTablet` to skip deleting if we introduce `RowMerger` without `ignoreDelete()`.

fluss-server/src/main/java/com/alibaba/fluss/server/kv/mergeengine/VersionRowMergeEngine.java
```diff
@@ -0,0 +1,119 @@
...
+        if (dataType instanceof BigIntType) {
+            return (left, right) -> (Long) left > (Long) right;
+        }
+        if (dataType instanceof IntType || dataType instanceof TimeType) {
+            return (left, right) -> (Integer) left > (Integer) right;
+        }
+        if (dataType instanceof TimestampType) {
+            return (left, right) ->
+                    ((TimestampNtz) left).getMillisecond()
+                            > ((TimestampNtz) right).getMillisecond();
```
wuchong: Should use `TimestampNtz#compareTo` to consider nano seconds. Add tests for `TIMESTAMP(9)`.

fluss-server/src/main/java/com/alibaba/fluss/server/kv/mergeengine/VersionRowMergeEngine.java
```diff
@@ -0,0 +1,119 @@
...
+        }
+        if (dataType instanceof LocalZonedTimestampType) {
+            return (left, right) ->
+                    ((TimestampLtz) left).toEpochMicros() > ((TimestampLtz) right).toEpochMicros();
```
wuchong: Should use `TimestampLtz#compareTo` to consider nano seconds. Add tests for `TIMESTAMP_LTZ(9)`.

fluss-server/src/main/java/com/alibaba/fluss/server/kv/mergeengine/VersionRowMergeEngine.java
```diff
@@ -0,0 +1,119 @@
...
+        }
+        throw new UnsupportedOperationException("Unsupported data type: " + dataType);
+    }
+
+    interface ValueComparator {
+        boolean isGreaterThan(Object left, Object right);
```
wuchong: Should return `int` or change to `isGreaterOrEqualThan` to handle equal case.

sunxiaojian:
> @sunxiaojian thanks for the contribution. This PR is already in a good shape. The comments are mainly around row merge engine. If you don't mind, I can help to refine this PR and merge it.

@wuchong Of course not, if you have time, I have already dealt with some of the comments.

fluss-connectors/fluss-connector-flink/src/test/java/com/alibaba/fluss/connector/flink/sink/FlinkTableSinkITCase.java
```diff
@@ -845,6 +845,164 @@ void testUnsupportedStmtOnFirstRowMergeEngine() {
...
+                        + " b string, ts TIMESTAMP(3)) with('table.merge-engine' = 'version','table.merge-engine.version.column' = 'ts')");
+
+        // insert once
+        tEnv.executeSql(
+                        "INSERT INTO merge_engine_with_version (a, b, ts) VALUES "
+                                + "(1, 'v1', TIMESTAMP '2024-12-27 12:00:00.123'), "
+                                + "(2, 'v2', TIMESTAMP '2024-12-27 12:00:00.123'), "
+                                + "(1, 'v11', TIMESTAMP '2024-12-27 11:59:59.123'), "
+                                + "(3, 'v3', TIMESTAMP '2024-12-27 12:00:00.123'),"
+                                + "(3, 'v33', TIMESTAMP '2024-12-27 12:00:00.123');")
```
wuchong, line 958: The last record of the same event_time/version should be retained. In practical, it's common to generate events with the same event_time, but we shouldn't drop the new values. So it's more acceptable to merge using last_row strategy instead of first_row strategy if version is the same. This is also [how Clickhouse does](https://clickhouse.com/docs/en/engines/table-engines/mergetree-family/replacingmergetree): 

> if ver specified. If ver is the same for several rows, then it will use "if ver is not specified" rule for them, i.e. the most recent inserted row will remain.

fluss-connectors/fluss-connector-flink/src/test/java/com/alibaba/fluss/connector/flink/sink/FlinkTableSinkITCase.java
```diff
@@ -845,6 +845,164 @@ void testUnsupportedStmtOnFirstRowMergeEngine() {
...
+                        + " b string, ts TIMESTAMP(9)) with('table.merge-engine' = 'version','table.merge-engine.version.column' = 'ts')");
+
+        // insert once
+        tEnv.executeSql(
+                        "INSERT INTO merge_engine_with_version (a, b, ts) VALUES "
+                                + "(1, 'v1', TIMESTAMP '2024-12-27 12:00:00.123456789'), "
+                                + "(2, 'v2', TIMESTAMP '2024-12-27 12:00:00.123456789'), "
+                                + "(1, 'v11', TIMESTAMP '2024-12-27 11:59:59.123456789'), "
+                                + "(3, 'v3', TIMESTAMP '2024-12-27 12:00:00.123456789'),"
+                                + "(3, 'v33', TIMESTAMP '2024-12-27 12:00:00.123456789');")
```
wuchong: To verify it works on nano seconds, the timestamp should only be different on nano unit.

wuchong (commented):
Thanks for the contribution @sunxiaojian, I will do a final refining and merge it.

wuchong:
@sunxiaojian , it seems you didn't correctly configure the `user.email` git config (unnecessary double quotes: `Xiaojian Sun <“sunxiaojian926@163.com”>`), and that's this commit https://github.com/alibaba/fluss/commit/ad2068379d7764c050283edff7892dce26a90e51 can't reflect your contribution.

sunxiaojian:
> @sunxiaojian , it seems you didn't correctly configure the `user.email` git config (unnecessary double quotes: `Xiaojian Sun <“sunxiaojian926@163.com”>`), and that's this commit [ad20683](https://github.com/alibaba/fluss/commit/ad2068379d7764c050283edff7892dce26a90e51) can't reflect your contribution.

ok, thanks.

PR #284, opened 2024-12-27
[connector] Fluss support scan.startup.timestamp is larger than max timestamp of bucket.

### LINKED ISSUE #274: [Feature] Flink Source throw the fetch timestamp %s is larger than the max timestamp %s.
### Search before asking

### Motivation

Currently, if the table bucket is not updated for two days, the max timestamp also will be earlier value.If I specified the source start from current, the exception will be thrown.

### Solution
just like what kafka source [1] and  paimon do.

kafka:
```java
// First get the current end offsets of the partitions. This is going to be used
        // in case we cannot find a suitable offsets based on the timestamp, i.e. the message
        // meeting the requirement of the timestamp have not been produced to Kafka yet, in
        // this case, we just use the latest offset.
        // We need to get the latest offsets before querying offsets by time to ensure that
        // no message is going to be missed.
```

paimon
```java
/**
     * Returns the latest snapshot earlier than the timestamp mills. A non-existent snapshot may be
     * returned if all snapshots are equal to or later than the timestamp mills.
     */
    public @Nullable Long earlierThanTimeMills(long timestampMills, boolean startFromChangelog) {
    }
```

[1] https://github.com/apache/flink-connector-kafka/blob/f6a077a9dd8d1d5e43fc545cc9baab227d8438a0/flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/source/enumerator/initializer/TimestampOffsetsInitializer.java#L51
[2] https://github.com/apache/paimon/blob/ba4576657edf1dde60ac321423d4fa6032169119/paimon-core/src/main/java/org/apache/paimon/utils/SnapshotManager.java#L306

loserwang1024:
### Purpose

Linked issue: close #274

### DIFF
--- a/fluss-connectors/fluss-connector-flink/src/test/java/com/alibaba/fluss/connector/flink/source/FlinkTableSourceITCase.java
+++ b/fluss-connectors/fluss-connector-flink/src/test/java/com/alibaba/fluss/connector/flink/source/FlinkTableSourceITCase.java
@@ -32,8 +32,6 @@ import org.apache.commons.lang3.RandomUtils;
 import org.apache.flink.api.common.typeinfo.TypeInformation;
 import org.apache.flink.api.common.typeinfo.Types;
 import org.apache.flink.api.java.typeutils.RowTypeInfo;
-import org.apache.flink.configuration.Configuration;
-import org.apache.flink.configuration.RestartStrategyOptions;
 import org.apache.flink.streaming.api.datastream.DataStream;
 import org.apache.flink.streaming.api.environment.StreamExecutionEnvironment;
 import org.apache.flink.table.api.DataTypes;
@@ -86,14 +84,7 @@ class FlinkTableSourceITCase extends FlinkTestBase {
         FlinkTestBase.beforeAll();
 
         String bootstrapServers = String.join(",", clientConf.get(ConfigOptions.BOOTSTRAP_SERVERS));
-        execEnv =
-                StreamExecutionEnvironment.getExecutionEnvironment(
-                        new Configuration()
-                                .set(
-                                        RestartStrategyOptions.RESTART_STRATEGY,
-                                        RestartStrategyOptions.RestartStrategyType
-                                                .NO_RESTART_STRATEGY
-                                                .getMainValue()));
+        execEnv = StreamExecutionEnvironment.getExecutionEnvironment();
         // create table environment
         tEnv = StreamTableEnvironment.create(execEnv, EnvironmentSettings.inStreamingMode());
         // crate catalog using sql
--- a/fluss-connectors/fluss-connector-flink/src/test/java/com/alibaba/fluss/connector/flink/source/FlinkTableSourceITCase.java
+++ b/fluss-connectors/fluss-connector-flink/src/test/java/com/alibaba/fluss/connector/flink/source/FlinkTableSourceITCase.java
@@ -32,6 +32,8 @@ import org.apache.commons.lang3.RandomUtils;
 import org.apache.flink.api.common.typeinfo.TypeInformation;
 import org.apache.flink.api.common.typeinfo.Types;
 import org.apache.flink.api.java.typeutils.RowTypeInfo;
+import org.apache.flink.configuration.Configuration;
+import org.apache.flink.configuration.RestartStrategyOptions;
 import org.apache.flink.streaming.api.datastream.DataStream;
 import org.apache.flink.streaming.api.environment.StreamExecutionEnvironment;
 import org.apache.flink.table.api.DataTypes;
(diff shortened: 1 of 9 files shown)


fluss-connectors/fluss-connector-flink/src/test/java/com/alibaba/fluss/connector/flink/source/FlinkTableSourceITCase.java
```diff
@@ -166,7 +166,7 @@ void testPkTableReadOnlySnapshot() throws Exception {
 
     @Test
     void testNonPkTableRead() throws Exception {
-        tEnv.executeSql("create table non_pk_table_test (a int, b varchar)");
+        tEnv.executeSql("create table f (a int, b varchar)");
```
wuchong: misupdate?

fluss-connectors/fluss-connector-flink/src/test/java/com/alibaba/fluss/connector/flink/source/FlinkTableSourceITCase.java
```diff
@@ -631,6 +631,49 @@ void testReadPrimaryKeyPartitionedTable() throws Exception {
         assertResultsIgnoreOrder(rowIter, expectedRowValues, true);
     }
 
+    @Test
+    void testReadTimestampOutOfBounds() throws Exception {
+
```
wuchong: nit: remove heading empty line

fluss-client/src/main/java/com/alibaba/fluss/client/admin/FlussAdmin.java
```diff
@@ -323,11 +330,21 @@ private static void sendListOffsetsRequest(
...
                                             if (resp.hasErrorCode()) {
-                                                bucketToOffsetMap
-                                                        .get(resp.getBucketId())
-                                                        .completeExceptionally(
-                                                                ApiError.fromErrorMessage(resp)
-                                                                        .exception());
+                                                ApiException exception =
+                                                        ApiError.fromErrorMessage(resp).exception();
+                                                if (exception
+                                                        instanceof InvalidTimestampException) {
```
wuchong: It's not encouraged to use exceptions as common information exchanging. I would recommend this can be handled in server-side to simplify client-side implementation (and future other clients).

fluss-connectors/fluss-connector-flink/src/main/java/com/alibaba/fluss/connector/flink/source/enumerator/initializer/BucketOffsetsRetrieverImpl.java
```diff
@@ -62,7 +62,26 @@ public Map<Integer, Long> earliestOffsets(
...
     public Map<Integer, Long> offsetsFromTimestamp(
             @Nullable String partitionName, Collection<Integer> buckets, long timestamp) {
-        return listOffsets(partitionName, buckets, new OffsetSpec.TimestampSpec(timestamp));
+
+        // First get the current end offsets of the buckets. This is going to be used in case we
+        // cannot find a suitable offsets based on the timestamp, i.e. the message meeting the
+        // requirement of the timestamp have not been produced to fluss yet. In this case, we just
+        // use the latest offset. We need to get the latest offsets before querying offsets by time
+        // to ensure that no message is going to be missed.
+        Map<Integer, Long> endOffsets = latestOffsets(partitionName, buckets);
```
wuchong: I recommend this fallback logic should be handled in server-side to have a data consistency and simplify client. 

From the perspective of `ListOffset(timestamp)` RPC method semantics, it should "return the offset of the first message whose timestamp is greater than or equals to the given timestamp. **If no such message is found, the log end offset is returned.**" We can still throw `InvalidTimestampException` if the timestamp is larger than current timestamp (a future time). 

And then, adds unit tests in `ReplicaManagerTest#testListOffsetsWithTimestamp`.

This is also how kafka implement it: https://github.com/apache/kafka/blob/b40a962a8ead04b10f7a8eba346fe79784de9e4e/core/src/main/scala/kafka/log/UnifiedLog.scala#L1252-L1256

wuchong (changes requested):

fluss-server/src/test/java/com/alibaba/fluss/server/replica/ReplicaManagerTest.java
```diff
@@ -880,10 +884,36 @@ void testListOffsetsWithTimestamp() throws Exception {
                             -1, ListOffsetsParam.TIMESTAMP_OFFSET_TYPE, commitTimestamp),
                     Collections.singleton(tb),
                     future1::complete);
+            List<ListOffsetsResultForBucket> listOffsetsResultForBuckets = future1.get();
+            for (ListOffsetsResultForBucket result : listOffsetsResultForBuckets) {
+                if (result.getErrorCode() != 0) {
+                    log.warn(result.getErrorMessage() + "," + result.getErrorCode());
```
wuchong: only used for debugging? can be removed now?

fluss-connectors/fluss-connector-flink/src/test/java/com/alibaba/fluss/connector/flink/source/FlinkTableSourceITCase.java
```diff
@@ -84,7 +86,14 @@ protected static void beforeAll() {
...
         String bootstrapServers = String.join(",", clientConf.get(ConfigOptions.BOOTSTRAP_SERVERS));
-        execEnv = StreamExecutionEnvironment.getExecutionEnvironment();
+        execEnv =
+                StreamExecutionEnvironment.getExecutionEnvironment(
+                        new Configuration()
+                                .set(
+                                        RestartStrategyOptions.RESTART_STRATEGY,
+                                        RestartStrategyOptions.RestartStrategyType
+                                                .NO_RESTART_STRATEGY
+                                                .getMainValue()));
```
wuchong: Why do we have to configure this?

wuchong:
@loserwang1024 I appended a commit to remove the unnecessary code. I will merge it if you have no concerns.

wuchong (approved):

PR #285, opened 2024-12-27
[flink] Fix read partitioned table throw LogScanner is not subscribed any bucket

### LINKED ISSUE #164: [Bug] Flink reads partitioned table throw exception ”LogScanner is not subscribed any bucket"
### Search before asking

### Fluss version

0.5.0

### Minimal reproduce step

Create a partitioned table, and using Flink to read the table

### What doesn't meet your expectations?

It should throw exeption

### Anything else?

1: currently, the logic of check partition changes is wrong, we should deduplicate the partitions from the splits assigned or pending
2: If the partitions should be deleted, the log scanner may has not buckets to substribe, it shouldn't throw exception when fetch

luoyuxia:
### Purpose

Linked issue: close #164

Fix read partitioned table throw LogScanner is not subscribed any bucket.

1. In FlinkSourceEnumerator, duplicate the partition of assigned partitions and pendingSplitAssignment to avoid handle the same partitions more than once which will cause one partition still in pending assignment to be removed.
2. In Flink split source reader, when no any subscribed bucket, not call `Scanner#poll` method, just return empty records directly

FlinkSourceEnumeratorTest#testDiscoverPartitionsPeriodically
FlinkSourceSplitReaderTest#testNoSubscribedBucket

### Documentation
N/A


fluss-connectors/fluss-connector-flink/src/main/java/com/alibaba/fluss/connector/flink/source/reader/FlinkSourceSplitReader.java
```diff
@@ -144,9 +144,15 @@ public RecordsWithSplitIds<RecordAndPos> fetch() throws IOException {
         } else {
             // may need to finish empty log splits
             if (!emptyLogSplits.isEmpty()) {
+                FlinkRecordsWithSplitIds records =
+                        new FlinkRecordsWithSplitIds(emptyLogSplits, flinkSourceReaderMetrics);
```
wuchong: should deep copy the `emptyLogSplits`? otherwise, it will be cleared before return. 

Add a unit test for it in `FlinkSourceSplitReaderTest`?

fluss-connectors/fluss-connector-flink/src/test/java/com/alibaba/fluss/connector/flink/source/reader/FlinkSourceSplitReaderTest.java
```diff
@@ -268,6 +268,26 @@ void testHandleMixSnapshotLogSplitChangesAndFetch() throws Exception {
...
+                        .build();
+        final TableDescriptor tableDescriptor =
+                TableDescriptor.builder().schema(schema).distributedBy(1).build();
+        createTable(tablePath, tableDescriptor);
+
+        try (FlinkSourceSplitReader splitReader =
+                createSplitReader(tablePath, schema.toRowType())) {
+            // fetch shouldn't throw exception
+            RecordsWithSplitIds<RecordAndPos> records = splitReader.fetch();
+            assertThat(records.nextSplit()).isNull();
```
wuchong, line 289: assert `records.nextRecordFromSplit()` is null as well?

fluss-server/src/main/java/com/alibaba/fluss/server/kv/KvTablet.java
```diff
@@ -345,7 +345,16 @@ public LogAppendInfo putAsLeader(
...
                             // now, we can build the full log.
-                            return logTablet.appendAsLeader(walBuilder.build());
+                            LogAppendInfo logAppendInfo =
+                                    logTablet.appendAsLeader(walBuilder.build());
+                            long logEndOffset = logAppendInfo.lastOffset();
+                            if (logEndOffset != logOffset) {
+                                LOG.warn(
+                                        "The log end offset {} is not equal to the expected log offset {}.",
+                                        logEndOffset,
+                                        logOffset);
```
wuchong: Do we still need this warning log?

fluss-server/src/main/java/com/alibaba/fluss/server/kv/KvTablet.java
```diff
@@ -357,6 +366,7 @@ public LogAppendInfo putAsLeader(
                                     false);
                         }
                     } catch (Throwable t) {
+                        LOG.warn("logEndOffsetOfPrevBatch: {}", logEndOffsetOfPrevBatch);
```
wuchong: Do we still need this warning log?

fluss-client/src/test/java/com/alibaba/fluss/client/table/FlussPartitionedTableITCase.java
```diff
@@ -58,6 +58,10 @@ void testPartitionedPrimaryKeyTable() throws Exception {
         Schema schema = createPartitionedTable(DATA1_TABLE_PATH_PK, true);
         Map<String, Long> partitionIdByNames =
                 FLUSS_CLUSTER_EXTENSION.waitUtilPartitionAllReady(DATA1_TABLE_PATH_PK);
+
+        TableDescriptor partitionTableDescriptor = TableDescriptor.builder().schema(schema).build();
+        createTable(TablePath.of("test_db_11", "tt"), partitionTableDescriptor, false);
```
wuchong: What's the purpose of this test code? It seems the created table is never used.

luoyuxia: Forget to remove these codelines for debug test failure...

fluss-server/src/main/java/com/alibaba/fluss/server/kv/KvTablet.java
luoyuxia: Forget to remove these code lines for debug...

fluss-connectors/fluss-connector-flink/src/test/java/com/alibaba/fluss/connector/flink/source/reader/FlinkSourceSplitReaderTest.java
luoyuxia: If assert `records.nextRecordFromSplit()` is null, it will throw `java.lang.NullPointerException: Make sure nextSplit() did not return null before iterate over the records split.  `FlinkRecordsWithSplitIds#nextRecordFromSplit will check the `currentRecordIterator`  is not null, but if no records in the `FlinkRecordsWithSplitIds`, the `currentRecordIterator` will be null.

I think check `nextSplit` is enough since if `nextSplit` is null, flink fetcher won't call method `records.nextRecordFromSplit()`

wuchong (approved):
LGTM.

PR #290, opened 2024-12-31
[Client] Fix the BufferExhaustedException thrown when write a high volume of data to log table with arrow format

swuferhong:
### Purpose

Linked issue: https://github.com/alibaba/fluss/issues/267

This pr is aims to fix the `BufferExhaustedException` thrown when write a high volume of data to log table with arrow format. Now, we will pre-allocated some memory for `ArrowLogWriteBatch` in writer thread instead of allocate memory in sender thread.


fluss-client/src/main/java/com/alibaba/fluss/client/write/RecordAccumulator.java
```diff
@@ -400,29 +399,37 @@ private WriteBatch.WriteBatchType getWriteBatchType(
...
+    private List<MemorySegment> allocateMemorySegments(
             WriteRecord writeRecord, WriteBatch.WriteBatchType writeBatchType)
-            throws InterruptedException {
+            throws InterruptedException, IOException {
         if (writeBatchType == WriteBatch.WriteBatchType.ARROW_LOG) {
-            return memorySegmentPool.nextSegment(true);
+            // pre-allocate a list of memory segments to hold by the arrow batch to avoid deadlock
+            // in sender thread.
+            int pageSize = memorySegmentPool.pageSize();
+            int memorySegmentSize = batchSize / pageSize;
```
loserwang1024:
```suggestion
          int memorySegmentSize = batchSize / pageSize + (batchSize%pageZise ==0 ? 0:1)
```

fluss-common/src/main/java/com/alibaba/fluss/memory/PreAllocatedManagedPagedOutputView.java
```diff
@@ -0,0 +1,59 @@
...
+        this.segmentPool = segmentPool;
+        currentSegmentIndex = 1;
+    }
+
+    @Override
+    protected MemorySegment nextSegment() throws IOException {
+        if (currentSegmentIndex < memorySegments.size()) {
+            return memorySegments.get(currentSegmentIndex++);
+        } else {
+            MemorySegment segment = segmentPool.nextSegment(waitingSegment);
```
loserwang1024: Are't the PreAllocatedManagedPagedOutputView be allocated all the memory in pre (which is memorySegments)? Why still need to get segment from segmentPool?

fluss-client/src/main/java/com/alibaba/fluss/client/write/RecordAccumulator.java
```diff
@@ -521,14 +528,15 @@ private RecordAppendResult appendNewBatch(
         // If the table is kv table we need to create a kv batch, otherwise we create a log batch.
         WriteBatch batch;
         if (writeBatchType == WriteBatch.WriteBatchType.KV) {
+            MemorySegment memorySegment = segments.get(0);
```
loserwang1024: Maybe add a check whether segments is only one element, in case that the `allocateMemorySegments` change later.

loserwang1024 (commented):
I have left some comments.

luoyuxia:
Curious about why pre-allocated some memory can solve this exception...

fluss-common/src/main/java/com/alibaba/fluss/memory/PreAllocatedManagedPagedOutputView.java
swuferhong: `PreAllocatedManagedPagedOutputView` only pre-allocate an approximate amount of memory in advance, which may not be enough to serialize the `ArrowWriteBatch`.

fluss-client/src/main/java/com/alibaba/fluss/client/write/RecordAccumulator.java
swuferhong: The WriterMemoryBuffer will be removed in the future, so there is no need to add a check here.(Always one memeorySegment).

fluss-client/src/main/java/com/alibaba/fluss/client/write/ArrowLogWriteBatch.java
```diff
@@ -124,9 +129,15 @@ public int sizeInBytes() {
...
-                .collect(Collectors.toList());
+        List<MemorySegment> usedMemorySegments =
+                outputView.getSegmentBytesViewList().stream()
+                        .map(MemorySegmentBytesView::getMemorySegment)
+                        .collect(Collectors.toList());
+        if (usedMemorySegments.size() > preAllocatedMemorySegments.size()) {
+            return usedMemorySegments;
+        } else {
+            return preAllocatedMemorySegments;
+        }
```
wuchong: Always return memory segments from outputView is fine?

fluss-client/src/main/java/com/alibaba/fluss/client/write/RecordAccumulator.java
```diff
@@ -400,29 +399,37 @@ private WriteBatch.WriteBatchType getWriteBatchType(
...
         } else {
             // get the new size.
             int size = Math.max(batchSize, writeRecord.getEstimatedSizeInBytes());
             // TODO check the remaining time to wait for allocating memory segment.
-            return writerMemoryBuffer.allocate(size, Long.MAX_VALUE);
+            return Collections.singletonList(writerMemoryBuffer.allocate(size, Long.MAX_VALUE));
         }
     }
 
     private void deallocateMemorySegment(
```
wuchong: `deallocateMemorySegments`

fluss-client/src/main/java/com/alibaba/fluss/client/write/RecordAccumulator.java
```diff
@@ -400,29 +399,37 @@ private WriteBatch.WriteBatchType getWriteBatchType(
...
+            throws InterruptedException, IOException {
         if (writeBatchType == WriteBatch.WriteBatchType.ARROW_LOG) {
-            return memorySegmentPool.nextSegment(true);
+            // pre-allocate a list of memory segments to hold by the arrow batch to avoid deadlock
+            // in sender thread.
+            int pageSize = memorySegmentPool.pageSize();
+            int memorySegmentSize = batchSize / pageSize;
+            List<MemorySegment> memorySegments = new ArrayList<>(memorySegmentSize);
+            for (int i = 0; i < memorySegmentSize; i++) {
+                memorySegments.add(memorySegmentPool.nextSegment(true));
```
wuchong: We should support allocating segments in an atomic way, otherwise, there may be starvation cases (e.g., there are only 10 pages and 10 people requesting 10 pages).

wuchong (approved):
I appended a commit to better unify the code logic and refactor the arrow writer logic a bit (removed `serialize()` methods). See the commit message for more details. https://github.com/alibaba/fluss/pull/290/commits/e579f0982084f2bec364019b69f8fb53dffd885d

swuferhong:
@wuchong  LGTM，can be merged.

PR #307, opened 2025-01-03
[flink](lookup) convert to Fluss GenericRow to avoid unnecessary deserialization

JNSimba:
### Purpose
convert to Fluss GenericRow to avoid unnecessary deserialization


luoyuxia:
Thanks for the pr, but seems this has been addressed in #371.. Let's close the pr

PR #312, opened 2025-01-06
[kv] Fix out of order exception after delete a not exist row

swuferhong:
### Purpose

Linked issue: https://github.com/alibaba/fluss/issues/230

This pr is aims to fix the `OutOfOrderSequenceException` after delete a not exist row or input keys with `MergeEngine` set to `FIRST_ROW`. The reason of this problem is that the `kvRecordBatch` send from client may not generated any CDC logs, 
for example, when client attempts to delete some non-existent keys or `MergeEngine` set to `FIRST_ROW`. In this case, the old implements will simply return but not update batchSequence,  as doing so would cause a  `OutOfOrderSequenceException` problem. Therefore, this pr will introduce a filed `lastOffsetDelta` in record batch used to calculate the lastOffset of the current batch as: [lastOffset = baseOffset + LastOffsetDelta] instead of [lastOffset = baseOffset + recordCount -1]. The reason for introducing this field is that there might be cases where the offset delta in batch does not match the recordCount. For example, when generating CDC logs for a kv table and sending a batch that only contains the deletion of non-existent kvs, no CDC logs would be generated. However, we need to increment the batchSequence for the corresponding writerId to make sure no {@link OutOfOrderSequenceException} will be thrown. In such a case, we would generate a logRecordBatch with a LastOffsetDelta of 1 but a recordCount of 0.

### DIFF
--- a/fluss-common/src/main/java/com/alibaba/fluss/record/DefaultLogRecordBatch.java
+++ b/fluss-common/src/main/java/com/alibaba/fluss/record/DefaultLogRecordBatch.java
@@ -18,6 +18,7 @@ package com.alibaba.fluss.record;
 
 import com.alibaba.fluss.annotation.PublicEvolving;
 import com.alibaba.fluss.exception.CorruptMessageException;
+import com.alibaba.fluss.exception.OutOfOrderSequenceException;
 import com.alibaba.fluss.memory.MemorySegment;
 import com.alibaba.fluss.metadata.LogFormat;
 import com.alibaba.fluss.row.arrow.ArrowReader;
@@ -50,6 +51,7 @@ import java.util.NoSuchElementException;
  *   <li>CRC => Uint32
  *   <li>SchemaId => Int16
  *   <li>Attributes => Int8
+ *   <li>LastOffsetDelta => Int32
  *   <li>WriterID => Int64
  *   <li>SequenceID => Int32
  *   <li>RecordCount => Int32
@@ -61,6 +63,15 @@ import java.util.NoSuchElementException;
  * magic byte before deciding how to interpret the bytes between the batch length and the magic
  * byte. The CRC-32C (Castagnoli) polynomial is used for the computation. CommitTimestamp is also
  * located before the CRC, because it is determined in server side.
+ *
+ * <p>The filed 'lastOffsetDelta is used to calculate the lastOffset of the current batch as:
+ * [lastOffset = baseOffset + LastOffsetDelta] instead of [lastOffset = baseOffset + recordCount -
+ * 1]. The reason for introducing this field is that there might be cases where the offset delta in
+ * batch does not match the recordCount. For example, when generating CDC logs for a kv table and
+ * sending a batch that only contains the deletion of non-existent kvs, no CDC logs would be
+ * generated. However, we need to increment the batchSequence for the corresponding writerId to make
+ * sure no {@link OutOfOrderSequenceException} will be thrown. In such a case, we would generate a
+ * logRecordBatch with a LastOffsetDelta of 1 but a recordCount of 0.
  *
  * <p>The current attributes are given below:
  *
@@ -82,6 +93,7 @@ public class DefaultLogRecordBatch implements LogRecordBatch {
     static final int CRC_LENGTH = 4;
     static final int SCHEMA_ID_LENGTH = 2;
     static final int ATTRIBUTE_LENGTH = 1;
+    static final int LAST_OFFSET_DELTA_LENGTH = 4;
     static final int WRITE_CLIENT_ID_LENGTH = 8;
     static final int BATCH_SEQUENCE_LENGTH = 4;
     static final int RECORDS_COUNT_LENGTH = 4;
@@ -93,7 +105,8 @@ public class DefaultLogRecordBatch implements LogRecordBatch {
     public static final int CRC_OFFSET = COMMIT_TIMESTAMP_OFFSET + COMMIT_TIMESTAMP_LENGTH;
     protected static final int SCHEMA_ID_OFFSET = CRC_OFFSET + CRC_LENGTH;
     static final int ATTRIBUTES_OFFSET = SCHEMA_ID_OFFSET + SCHEMA_ID_LENGTH;
-    static final int WRITE_CLIENT_ID_OFFSET = ATTRIBUTES_OFFSET + ATTRIBUTE_LENGTH;
+    static final int LAST_OFFSET_DELTA_OFFSET = ATTRIBUTES_OFFSET + ATTRIBUTE_LENGTH;
+    static final int WRITE_CLIENT_ID_OFFSET = LAST_OFFSET_DELTA_OFFSET + LAST_OFFSET_DELTA_LENGTH;
     static final int BATCH_SEQUENCE_OFFSET = WRITE_CLIENT_ID_OFFSET + WRITE_CLIENT_ID_LENGTH;
     public static final int RECORDS_COUNT_OFFSET = BATCH_SEQUENCE_OFFSET + BATCH_SEQUENCE_LENGTH;
     static final int RECORDS_OFFSET = RECORDS_COUNT_OFFSET + RECORDS_COUNT_LENGTH;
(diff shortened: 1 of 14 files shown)


swuferhong:
@wuchong @luoyuxia pr ready.

fluss-common/src/test/java/com/alibaba/fluss/record/DefaultLogRecordBatchTest.java
```diff
@@ -89,4 +92,75 @@ void testIndexedRowWriteAndReadBatch() throws Exception {
...
+    void testOverrideLastLogOffset() throws Exception {
+        List<IndexedRow> rows1 =
+                Arrays.asList(
+                        row(baseRowType, new Object[] {1, "1"}),
+                        row(baseRowType, new Object[] {2, "2"}),
+                        row(baseRowType, new Object[] {3, "3"}));
+
+        MemoryLogRecordsIndexedBuilder builder =
+                MemoryLogRecordsIndexedBuilder.builder(
+                        0L, schemaId, Integer.MAX_VALUE, magic, outputView);
```
wuchong: test with a non-zero base offset to verify the last offset is a `base_offset + delta_offset`

fluss-common/src/main/java/com/alibaba/fluss/record/MemoryLogRecordsArrowBuilder.java
```diff
@@ -227,6 +230,14 @@ public void resetWriterState(long writerId, int batchSequence) {
         this.batchSequence = batchSequence;
     }
 
+    public void overrideLastLogOffset(long lastLogOffset) {
```
wuchong: We don't need to provide this method. we only need to check `recordCount` is 0 or not and write corresponding offset_delta in `writeBatchHeader`. 

It's dangerous to provide this method, as it's only valid if the `lastLogOffset=0 and recordCount=0` or `lastLogOffset=baseOffset+recordCount-1`, but there is no any validation in this method.

fluss-common/src/main/java/com/alibaba/fluss/record/DefaultLogRecordBatch.java
```diff
@@ -62,6 +64,15 @@
  * byte. The CRC-32C (Castagnoli) polynomial is used for the computation. CommitTimestamp is also
  * located before the CRC, because it is determined in server side.
  *
+ * <p>The filed 'lastOffsetDelta is used to calculate the lastOffset of the current batch as:
```
wuchong:
```suggestion
 * <p>The field 'lastOffsetDelta' is used to calculate the lastOffset of the current batch as:
```

fluss-common/src/main/java/com/alibaba/fluss/record/DefaultLogRecordBatch.java
```diff
@@ -62,6 +64,15 @@
...
+ * [lastOffset = baseOffset + LastOffsetDelta] instead of [lastOffset = baseOffset + recordCount -
+ * 1]. The reason for introducing this field is that there might be cases where the offset delta in
+ * batch does not match the recordCount. For example, when generating CDC logs for a kv table and
+ * sending a batch that only contains the deletion of non-existent kvs, no CDC logs would be
+ * generated. However, we need to increment the batchSequence for the corresponding writerId to make
+ * sure no {@link OutOfOrderSequenceException} will be thrown. In such a case, we would generate a
+ * logRecordBatch with a LastOffsetDelta of 1 but a recordCount of 0.
```
wuchong:
```suggestion
 * logRecordBatch with a LastOffsetDelta of 0 but a recordCount of 0.
```

fluss-common/src/main/java/com/alibaba/fluss/record/MemoryLogRecordsIndexedBuilder.java
```diff
@@ -158,6 +160,14 @@ public void resetWriterState(long writerId, int batchSequence) {
         this.batchSequence = batchSequence;
     }
 
+    public void overrideLastLogOffset(long lastLogOffset) {
+        Preconditions.checkArgument(
```
wuchong: nit: Usually, we static import `Preconditions` to improve readability. that is directly use `checkArgument(...)` without prefix class name.

fluss-server/src/test/java/com/alibaba/fluss/server/replica/ReplicaManagerTest.java
```diff
@@ -562,6 +569,73 @@ void testPutKvWithOutOfBatchSequence() throws Exception {
                 DATA1_ROW_TYPE, records, Arrays.asList(expectedLogForData1, expectedLogForData2));
     }
 
+    @Test
+    void testPutKvWithDeleteNonExistsKey() throws Exception {
```
wuchong, line 573: This method is failed in my local with exception:

```
java.lang.IndexOutOfBoundsException
	at com.alibaba.fluss.memory.MemorySegment.getIntNativeEndian(MemorySegment.java:741)
	at com.alibaba.fluss.memory.MemorySegment.getInt(MemorySegment.java:714)
	at com.alibaba.fluss.record.DefaultLogRecordBatch.getRecordCount(DefaultLogRecordBatch.java:227)
	at com.alibaba.fluss.record.FileLogInputStream$FileChannelLogRecordBatch.getRecordCount(FileLogInputStream.java:161)
	at com.alibaba.fluss.server.replica.ReplicaManagerTest.testPutKvWithDeleteNonExistsKey(ReplicaManagerTest.java:620)
```

wuchong (commented):
If `recordCount` is zero, we also don't need to write any row_kind vector and arrow metadata and arrow buffer. That means the sizeInBytes of such log batch should be 48 (`RECORD_BATCH_HEADER_SIZE`). Could improve this and add a test for it?

fluss-common/src/main/java/com/alibaba/fluss/record/DefaultLogRecordBatch.java
swuferhong: Done!

fluss-common/src/main/java/com/alibaba/fluss/record/DefaultLogRecordBatch.java
swuferhong: Done!

fluss-common/src/main/java/com/alibaba/fluss/record/MemoryLogRecordsIndexedBuilder.java
swuferhong: Done! This method will be removed.

swuferhong:
> If `recordCount` is zero, we also don't need to write any row_kind vector and arrow metadata and arrow buffer. That means the sizeInBytes of such log batch should be 48 (`RECORD_BATCH_HEADER_SIZE`). Could improve this and add a test for it?

Currently, if the record count is zero in `MemoryLogRecordsArrowBuilder`, when call `serializeToOutputView`, it will return directly, so the row_kind vector and arrow metadata and arrow buffer are empty, so there is no need to add logic to cover this. But I will add a test to verify it.

fluss-common/src/main/java/com/alibaba/fluss/record/MemoryLogRecordsArrowBuilder.java
swuferhong: Done! This method has been removed.

fluss-common/src/test/java/com/alibaba/fluss/record/DefaultLogRecordBatchTest.java
swuferhong: Done!

fluss-server/src/test/java/com/alibaba/fluss/server/replica/ReplicaManagerTest.java
swuferhong: I ran many rounds locally without encountering this issue. Can you check again in your local env with the new commit?

swuferhong:
@wuchong  comments addressed.

wuchong (approved):
LGTM

fluss-server/src/test/java/com/alibaba/fluss/server/replica/ReplicaManagerTest.java
wuchong: After rebuilding the source code with the latest branch, the failure disappeared. I think local dirty classes cause this.

PR #322, opened 2025-01-13
[client] RecordAccumulator's ready method will be blocked if no batch is ready.

### LINKED ISSUE #306: [Bug] Fluss sender will be in busy loop when is in low traffic
### Search before asking

### Fluss version

main

### Minimal reproduce step

When in low traffic, the sender finds no records  ready to send, sleeps 1ms, and then  `runOnce` again to check whether records are ready which causes a busy loop to occupy more CPU..

### What doesn't meet your expectations?

N/A

### Anything else?

If I modify it to sleep 50ms, the CPU will be reduced from 120 to 80... But we will need a more sophisticated mechanism to solve it..

loserwang1024:
### Purpose

Linked issue: close #306 

### Tests

com.alibaba.fluss.client.write.RecordAccumulatorTest#testReadyWithTimeOut


loserwang1024:
@wuchong @swuferhong @luoyuxia , CC

loserwang1024:
rebase

fluss-client/src/test/java/com/alibaba/fluss/client/write/RecordAccumulatorTest.java
```diff
@@ -391,6 +392,41 @@ void testAwaitFlushComplete() throws Exception {
...
+        cluster = updateCluster(Arrays.asList(bucket1, bucket2, bucket3, bucket4));
+
+        // wait for ready timeout.
+        CompletableFuture<Void> future =
+                CompletableFuture.runAsync(
+                        () -> {
+                            accum.ready(cluster, 500L);
+                        });
+        assertThat(future).isNotCompleted();
+        Thread.sleep(1000L);
```
wuchong: Use `future.get()` or `future.join()` to wait async result. Never use sleep as it is not guaranteed to finish in the sleep time in different testing environment, and makes the tests unstable.

fluss-client/src/test/java/com/alibaba/fluss/client/write/RecordAccumulatorTest.java
```diff
@@ -391,6 +392,41 @@ void testAwaitFlushComplete() throws Exception {
         assertThatThrownBy(accum::awaitFlushCompletion).isInstanceOf(InterruptedException.class);
     }
 
+    @Test
+    void testReadyWithTimeOut() throws Exception {
```
wuchong: What's the purpose of this test? Do we need to check the `accumulator.ready` result? 

Maybe we can take Kafka's `RecordAccumulatorTest#testNextReadyCheckDelay` as an example.

fluss-client/src/main/java/com/alibaba/fluss/client/write/RecordAccumulator.java
```diff
@@ -232,6 +243,10 @@ public RecordAppendResult append(
      * </pre>
      */
     public ReadyCheckResult ready(Cluster cluster) {
+        return ready(cluster, batchTimeoutMs);
+    }
+
+    public ReadyCheckResult ready(Cluster cluster, long timeoutMs) {
```
wuchong: why do we need to provide this method? only testing purpose? If yes, why not use the `batchTimeoutMs` parameter in tests?

fluss-client/src/main/java/com/alibaba/fluss/client/write/RecordAccumulator.java
```diff
@@ -402,59 +420,59 @@ private void bucketReady(
...
             }
         }
     }
 
+    private boolean batchReady(Deque<WriteBatch> deque) {
+        final long waitedTimeMs;
+        final int dequeSize;
+        final boolean full;
+
+        boolean exhausted = writerBufferPool.queued() > 0;
```
wuchong: this should be moved out of batches foreach, as it requires to obtain a lock.

fluss-client/src/main/java/com/alibaba/fluss/client/write/RecordAccumulator.java
```diff
@@ -760,6 +779,30 @@ private void insertInSequenceOrder(Deque<WriteBatch> deque, WriteBatch batch) {
...
+                    }
+                });
+    }
+
+    private void waitBatchReady(long timeoutMs) {
+        try {
+            inLock(
+                    batchReadyLock,
+                    () -> {
+                        if (timeoutMs > 0) {
```
wuchong: should be checked before lock .

fluss-client/src/main/java/com/alibaba/fluss/client/write/RecordAccumulator.java
```diff
@@ -532,6 +550,7 @@ private RecordAppendResult tryAppend(
         WriteBatch last = deque.peekLast();
         if (last != null) {
             boolean success = last.tryAppend(writeRecord, callback);
+            notifyBatchReady(deque);
```
wuchong: why here need to notify batch ready?

wuchong (changes requested):

fluss-client/src/main/java/com/alibaba/fluss/client/write/RecordAccumulator.java
loserwang1024: @wuchong If not try to notify batch after each insert, the RecordAccumulator#ready will be blocked until timeout.  I use  use a producer-consumer-mode in RecordAccumulator, which block in consumer and signal by producer.

Maybe I do same thing as kafka,  accumulator.ready return the min wait time of each bucket, and then sender suspends with  `Thread.sleep(nextReadyCheckDelayMs)` and doesn't signal by consumer.

wuchong: I mean why notify for each record? Not when a batch is ready? It will result in busy loop that the sender doesn't wait at all.

loserwang1024:
@wuchong , I have refactored the design based on Kafka client

wuchong (approved):
LGTM

PR #324, opened 2025-01-14
[test] Harmonization of LazyMemorySegmentPoolTest and WriterMemoryBufferTest

### LINKED ISSUE #323: [Tests] Migrate tests of WriterMemoryBufferTest to LazyMemorySegmentPoolTest
### Search before asking

### Motivation

`WriterMemoryBuffer` has been deprecated and is replaced by `LazyMemorySegmentPool`, we should move tests of `WriterMemoryBufferTest` into `LazyMemorySegmentPoolTest` and then remove the `WriterMemoryBuffer`.

### Solution

_No response_

### Anything else?

_No response_

### Willingness to contribute

michaelkoepf:
### Purpose

Linked issue: close #323

- Moved `WriterMemoryBufferTest` tests to `LazyMemorySegmentPoolTest`
- Optimizations and bug fixes in `LazyMemorySegmentPool`
- Replaced constants in `LazyMemorySegmentPool` with configuration options
- Added additional test cases
- Removed associated deprecated classes
- Minor clean ups and refactoring for config options

### Tests

n/a

### API and Format

n/a

### Documentation

n/a


fluss-common/src/main/java/com/alibaba/fluss/config/ConfigOptions.java
```diff
@@ -252,6 +252,23 @@ public class ConfigOptions {
...
+    public static final ConfigOption<MemorySize> SERVER_BUFFER_PER_REQUEST_MEMORY_SIZE =
+            key("server.buffer.per-request-memory-size")
+                    .memoryType()
+                    .defaultValue(MemorySize.parse("0b"))
+                    .withDescription(
+                            "The minimum number of bytes that will be allocated by the writer rounded down to the closest multiple of "
+                                    + SERVER_BUFFER_PAGE_SIZE.key()
+                                    + " (but at least one page). This option allows to allocate memory in batches to have better CPU-cached friendliness due to contiguous segments.");
+
+    public static final ConfigOption<Long> SERVER_BUFFER_POOL_WAIT_TIMEOUT =
```
wuchong: Use `Duration` for a timeout to align with other timeout options? A reason for using Duration is that users can use readable string (e.g., `5s`) to configure the option.

fluss-common/src/main/java/com/alibaba/fluss/config/ConfigOptions.java
```diff
@@ -620,14 +637,22 @@ public class ConfigOptions {
                             "The writer or walBuilder will attempt to batch records together into one batch for"
                                     + " the same bucket. This helps performance on both the client and the server.");
 
-    @Deprecated
-    public static final ConfigOption<MemorySize> CLIENT_WRITER_LEGACY_BATCH_SIZE =
-            key("client.writer.legacy.batch-size")
+    // noDefaultValue() leads to NPE in some code parts, hence we default to 0b
+    public static final ConfigOption<MemorySize> CLIENT_WRITER_PER_REQUEST_MEMORY_SIZE =
+            key("client.writer.per-request-memory-size")
```
wuchong: `client.writer.buffer.per-request-memory-size` to have same prefix with other writer buffer options.

Move it after `CLIENT_WRITER_BUFFER_PAGE_SIZE` definition.

fluss-common/src/main/java/com/alibaba/fluss/config/ConfigOptions.java
```diff
@@ -620,14 +637,22 @@ public class ConfigOptions {
...
+                    .defaultValue(MemorySize.parse("0b"))
                     .withDescription(
-                            "The writer or walBuilder will attempt to batch records together into one batch for"
-                                    + " the same bucket. This helps performance on both the client and the server.");
+                            "The minimum number of bytes that will be allocated by the writer rounded down to the closes multiple of "
+                                    + CLIENT_WRITER_BUFFER_PAGE_SIZE.key()
+                                    + " (but at least one page). This option allows to allocate memory in batches to have better CPU-cached friendliness due to contiguous segments.");
+
+    public static final ConfigOption<Long> CLIENT_WRITER_WAIT_TIMEOUT =
+            key("client.writer.wait-timeout")
```
wuchong: `client.writer.buffer.wait-timeout` to have same prefix with other writer buffer options.

Use `Duration` as the option type.

fluss-common/src/main/java/com/alibaba/fluss/memory/LazyMemorySegmentPool.java
```diff
@@ -102,42 +103,50 @@ public static LazyMemorySegmentPool createWriterBufferPool(Configuration conf) {
...
     }
 
     public static LazyMemorySegmentPool createServerBufferPool(Configuration conf) {
         long totalBytes = conf.get(ConfigOptions.SERVER_BUFFER_MEMORY_SIZE).getBytes();
         int pageSize = (int) conf.get(ConfigOptions.SERVER_BUFFER_PAGE_SIZE).getBytes();
+        long perRequestMemorySize =
+                conf.getOptional(ConfigOptions.SERVER_BUFFER_PER_REQUEST_MEMORY_SIZE)
+                        .filter(s -> s.getBytes() > 0)
+                        .map(MemorySize::getBytes)
+                        .orElse((long) pageSize);
```
wuchong: Use `16mb` as default for the option? So we don't lose the CPU cache friendly after this change, and can simplify this to a single line `conf.get(ConfigOptions.CLIENT_WRITER_PER_REQUEST_MEMORY_SIZE).getBytes()`.

fluss-common/src/main/java/com/alibaba/fluss/memory/LazyMemorySegmentPool.java
```diff
@@ -102,42 +103,50 @@ public static LazyMemorySegmentPool createWriterBufferPool(Configuration conf) {
                         totalBytes,
                         ConfigOptions.CLIENT_WRITER_BATCH_SIZE.key(),
                         batchSize));
-
         int pageSize = (int) conf.get(ConfigOptions.CLIENT_WRITER_BUFFER_PAGE_SIZE).getBytes();
+        long perRequestMemorySize =
+                conf.getOptional(ConfigOptions.CLIENT_WRITER_PER_REQUEST_MEMORY_SIZE)
+                        .filter(s -> s.getBytes() > 0)
+                        .map(MemorySize::getBytes)
+                        .orElse((long) pageSize);
```
wuchong: Use `16mb` as default for the option?

fluss-common/src/main/java/com/alibaba/fluss/memory/LazyMemorySegmentPool.java
```diff
@@ -102,42 +103,50 @@ public static LazyMemorySegmentPool createWriterBufferPool(Configuration conf) {
...
+        return inLock(lock, () -> allocatePages(1).get(0));
     }
 
     @Override
     public List<MemorySegment> allocatePages(int requiredPages) throws IOException {
+        if (maxPages < requiredPages) { // immediately fail if the request is impossible to satisfy
+            throw new EOFException(
+                    String.format(
+                            "Allocation request cannot be satisfied because the number of maximum available pages is "
+                                    + "exceeded. Available pages: %d. Requested pages: %d",
```
wuchong:
```suggestion
                                    + "exceeded. Total pages: %d. Requested pages: %d",
```

`freePages` is "Available pages" .

fluss-common/src/main/java/com/alibaba/fluss/memory/LazyMemorySegmentPool.java
```diff
@@ -156,9 +165,11 @@ private List<MemorySegment> drain(int numPages) {
...
     }
 
-    private void lazilyAllocatePages(int required) {
-        if (cachePages.isEmpty()) {
-            int numPages = Math.min(required, perRequestPages);
+    @VisibleForTesting
+    protected void lazilyAllocatePages(int required) {
+        if (cachePages.size() < required) {
+            int numPages =
+                    Math.min(freePages(), Math.max(required - cachePages.size(), perRequestPages));
```
wuchong: I just found that it is not correct to use `freePages()` as an upper bound, we should use `un-allocated pages` as an upper bound.

For example, if we have 10 total pages, 3 in cache, 5 used, requesting 5 pages, perRequestPages is 5, then `free = total-used = 5`, it will allocate 5 pages, but the number of all the allocated pages (5 new pages, 3 cache pages, 5 used pages) exceeds total 10 pages. The correct pages to allocate is 2. 

A fix for this can be:
```java
int minAllocatePages = required - cachePages.size();
int maxAllocatePages = freePages() - cachePages.size();
// try to allocate more pages than minAllocatePages to have better CPU cache
int numPages = Math.min(maxAllocatePages, Math.max(minAllocatePages, perRequestPages));
```

Btw, could you add a unit test for this in `LazyMemorySegmentPoolTest.java`?

wuchong (commented):
Thanks @michaelkoepf for the great work! I left some comments.

wuchong:
@michaelkoepf no hurry, take it easy.

fluss-common/src/main/java/com/alibaba/fluss/config/ConfigOptions.java
michaelkoepf: which timeout duration do you prefer?

wuchong: I think we can use MAX (`Duration.ofMillis(Long.MAX_VALUE)`)as the default timeout duration.

michaelkoepf: ~found a nice constant value that i was not aware it exists in the java library. `ChronoUnit.FOREVER.getDuration()`~

nvm. cannot be used due to overflow...

fluss-common/src/test/java/com/alibaba/fluss/memory/LazyMemorySegmentPoolTest.java
```diff
@@ -132,6 +132,24 @@ void testPerRequestMemorySize() throws Exception {
...
+        assertThat(pool.freePages()).isEqualTo(2);
+        // should trigger an allocation request; should use the existing cached page and just
+        // allocate 1 additional one and not 2048/512=4
+        pool.allocatePages(2);
+        // check if page limit is respected
+        assertThat(pool.getAllCachePages().size()).isEqualTo(0);
+        // there should also be no further pages available
+        assertThat(pool.availableMemory()).isEqualTo(0);
+        assertThat(pool.freePages()).isEqualTo(0);
+    }
```
michaelkoepf, line 151: Requested test case

fluss-common/src/main/java/com/alibaba/fluss/memory/LazyMemorySegmentPool.java
michaelkoepf: See [here](https://github.com/alibaba/fluss/pull/324#discussion_r1975211879)

fluss-common/src/test/java/com/alibaba/fluss/memory/LazyMemorySegmentPoolTest.java
wuchong: thanks, this can reproduce the previous problem.

wuchong (approved):
Thanks @michaelkoepf for also fixing the Duration config inconsistency. I pushed a minor commit to fix the forgotten change of "Available pages" to "Total pages".

wuchong:
The unstable case in the build https://github.com/alibaba/fluss/actions/runs/13600540984/job/38025699943?pr=324 will be fixed in https://github.com/alibaba/fluss/pull/386/files#diff-c70b2b7f891f5da2da0a3414b098fbdc3814d89711264799f604159b7a900c5aR132

PR #330, opened 2025-01-21
[log] Arrow Compression support estimate compression ratio in client

swuferhong:
### Purpose

https://github.com/alibaba/fluss/issues/326

We have introduced arrow compression in https://github.com/alibaba/fluss/issues/187. However, Currently, the client calculates the batch size using the data before compress, which would cause a discrepancy between the actual batch size sent to the server and the batch size configured by the user, and in turn affects the compression efficiency. Therefore, in this pr we will introduce an estimator to continuously update the latest compression ratio in client, making the compressed batch size closer to the user's specified value.

### DIFF
--- a/fluss-client/src/test/java/com/alibaba/fluss/client/write/ArrowLogWriteBatchTest.java
+++ b/fluss-client/src/test/java/com/alibaba/fluss/client/write/ArrowLogWriteBatchTest.java
@@ -17,7 +17,6 @@
 package com.alibaba.fluss.client.write;
 
 import com.alibaba.fluss.compression.ArrowCompressionInfo;
-import com.alibaba.fluss.compression.ArrowCompressionRatioEstimator;
 import com.alibaba.fluss.compression.ArrowCompressionType;
 import com.alibaba.fluss.memory.MemorySegment;
 import com.alibaba.fluss.memory.PreAllocatedPagedOutputView;
@@ -29,12 +28,14 @@ import com.alibaba.fluss.record.LogRecordBatch;
 import com.alibaba.fluss.record.LogRecordReadContext;
 import com.alibaba.fluss.record.MemoryLogRecords;
 import com.alibaba.fluss.record.bytesview.BytesView;
+import com.alibaba.fluss.row.arrow.ArrowWriter;
 import com.alibaba.fluss.row.arrow.ArrowWriterPool;
 import com.alibaba.fluss.row.indexed.IndexedRow;
 import com.alibaba.fluss.shaded.arrow.org.apache.arrow.memory.BufferAllocator;
 import com.alibaba.fluss.shaded.arrow.org.apache.arrow.memory.RootAllocator;
 import com.alibaba.fluss.utils.CloseableIterator;
 
+import org.apache.commons.lang3.RandomStringUtils;
 import org.junit.jupiter.api.AfterEach;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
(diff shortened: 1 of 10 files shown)


fluss-client/src/test/java/com/alibaba/fluss/client/write/ArrowLogWriteBatchTest.java
```diff
@@ -166,6 +168,77 @@ void testAppendWithPreAllocatedMemorySegments() throws Exception {
...
+                            new PreAllocatedPagedOutputView(memorySegmentList));
+
+            int recordCount = 0;
+            while (arrowLogWriteBatch.tryAppend(
+                    createWriteRecord(
+                            row(
+                                    DATA1_ROW_TYPE,
+                                    new Object[] {
+                                        recordCount,
+                                        "a                                a                  a"
```
wuchong: Use guava `Strings.repeat("a", 50)` instead of counting characters manually, to improve code readability.

fluss-common/src/main/java/com/alibaba/fluss/row/arrow/ArrowWriter.java
```diff
@@ -61,6 +63,12 @@ public class ArrowWriter implements AutoCloseable {
      */
     public static final double BUFFER_USAGE_RATIO = 0.96;
 
+    /**
+     * The factor which is used to estimate the compression ratio of the serialized {@link
+     * ArrowRecordBatch}.
+     */
+    private static final float COMPRESSION_RATE_ESTIMATION_FACTOR = 1.05f;
```
wuchong: I believe that the purpose of this factor is to prevent memory shortages due to a decrease in compression rate during a particular compression, thereby adding a 5% margin for special case. However, we actually already have `BUFFER_USAGE_RATIO`, which can achieve the same goal, so there is no need to add another factor.

fluss-common/src/main/java/com/alibaba/fluss/row/arrow/ArrowWriter.java
```diff
@@ -138,7 +161,9 @@ public boolean isFull() {
             root.setRowCount(recordsCount);
             int metadataLength = getMetadataLength();
             int bodyLength = getBodyLength();
-            int currentSize = metadataLength + bodyLength;
+
+            int estimatedBodyLength = estimatedBytesWritten(bodyLength);
+            int currentSize = metadataLength + estimatedBodyLength;
```
wuchong, line 150: we also need to update `sizeInBytes()` to return the estimated size, otherwise, `RecordAccumulator#drain` can't respect the `maxSize` parameter.

fluss-common/src/main/java/com/alibaba/fluss/compression/ArrowCompressionRatioEstimator.java
```diff
@@ -0,0 +1,106 @@
...
+     * expected.
+     */
+    public static final float COMPRESSION_RATIO_IMPROVING_STEP = 0.005f;
+
+    /**
+     * The minimum speed to decrease compression ratio when a batch compresses worse than expected.
+     */
+    public static final float COMPRESSION_RATIO_DETERIORATE_STEP = 0.05f;
+
+    private final Map<Long, Map<String, Float>> compressionRatio;
```
wuchong: A table only have one compression type and compression level, we only need to maintain one compression for each table.

fluss-common/src/main/java/com/alibaba/fluss/row/arrow/ArrowWriter.java
```diff
@@ -253,6 +290,10 @@ public void recycle(long epoch) {
         if (this.epoch == epoch) {
             root.clear();
             provider.recycleWriter(this);
+
+            // reset the compression ratio.
+            this.estimatedCompressionRatio =
+                    compressionRatioEstimator.estimation(tableId, compressionInfo);
```
wuchong: This should be reset in `reset()` method.

PR #331, opened 2025-01-21
[kv] Use JDK9 Arrays#compare() to check bytes prefix for PrefixLookup

swuferhong:
### Purpose

Linked issue: https://github.com/alibaba/fluss/issues/271

Currently, `RocksDBKv#isPrefixEquals()` very inefficient to compare arrays byte by byte. Java9 provides a more efficient way [Arrays.compare(compare(byte[] a, int aFromIndex, int aToIndex, byte[] b, int bFromIndex, int bToIndex))](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/util/Arrays.html#compare(byte%5B%5D,int,int,byte%5B%5D,int,int)), underlying it leverages SIMD instructions.

As it is only supported since Java9, In this pr, we implement it  based on the running jdk env (the method can get the running JDK environment, if Java9 use the support Arrays.compare() method).

### DIFF
--- a/fluss-common/src/main/java/com/alibaba/fluss/utils/BytesUtils.java
+++ b/fluss-common/src/main/java/com/alibaba/fluss/utils/BytesUtils.java
@@ -17,8 +17,11 @@
 package com.alibaba.fluss.utils;
 
 import com.alibaba.fluss.annotation.Internal;
+import com.alibaba.fluss.utils.crc.Java;
 
+import java.lang.reflect.Method;
 import java.nio.ByteBuffer;
+import java.util.Arrays;
 
 /** Utils for bytes. */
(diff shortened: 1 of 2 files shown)


wuchong:
@swuferhong I adjusted the implementation a bit. 

1. extract the JDK9 compare logic from `RocksDBKv` into a common util so that it can be reused by others in the future. 
2. Use `Arrays.equals(byte[], int, int, byte[], int, int)` instead, because we only need to compare the prefix equality.
3. Avoid to new `PrefixComparer` for each `RocksDBKv` instance. In the final, it's just a util method.

wuchong (approved):
LGTM

PR #336, opened 2025-01-23
[docs] add some small improvements

polyzos:
### Purpose

Linked issue: close #xxx


website/docs/concepts/architecture.md
```diff
@@ -8,50 +8,50 @@ A Fluss cluster consists of two main processes: the **CoordinatorServer** and th
...
+- For **Primary Key Tables** which support updates, both the LogManager and KvManager are activated. The KvManager is used to support updates and kv lookups efficiently. The LogManager is used to store the changelogs of the table.
+- For **Log Tables** which only support `INSERT` operations, only the LogManager is activated, optimizing performance for write-heavy workloads.
 
-This architecture ensures the **TabletServer** delivers tailored data handling capabilities based on table types.
+This architecture ensures the TabletServer delivers tailored data handling capabilities based on table types.
 
 
-### LogStore
-The **LogStore** is designed to store log data, functioning similarly to a database binlog. 
+### LogManager
```
wuchong, line 28: I am still a bit hesitant because `LogStore` might make it easier for users to understand its role and responsibility (a component for storing log data), whereas `LogManager` may seem more abstract. So, I was thinking that maybe we could change the source code to replace `LogManager` with `LogStore` in order to align with the concept in the documentation.

From the perspective of concepts/architecture introduction, which word do you think is better?

website/docs/table-design/overview.md
```diff
@@ -6,24 +6,19 @@ sidebar_position: 2
...
-- **For PrimaryKey Tables**, the partition column must be a subset of the primary key to ensure uniqueness.
+- **Log Tables:** are optimized for append-only workloads and support only `INSERT` operations.
+- **Primary Key Tables:** are designed for managing and updating data in business applications. They support `INSERT`, `UPDATE`, and `DELETE` operations, based on a primary key.
 
-This design ensures efficient data organization, flexibility in handling different use cases, and adherence to data integrity constraints.
+Tables can also be partitioned (see [partitioned tables](../table-design/data-distribution/partitioning.md)) by defining a partition column, enabling efficient data organization. Partitioned tables store rows with the same partition value in a single partition. Partitioning applies to both Log Tables and Primary Key Tables, with specific considerations:
+- **Log Tables:** Partitioning is typically used for log data (e.g., date-based columns) to streamline data separation and retention processes.
+- **Primary Key Tables:** Partition columns must be a subset of the primary key to maintain data integrity and ensure uniqueness.
```
wuchong, line 19: Maybe we can add "The partitioning of a primary key table is primarily used for automatic data cleanup."

website/docs/concepts/architecture.md
```diff
@@ -8,50 +8,50 @@ A Fluss cluster consists of two main processes: the **CoordinatorServer** and th
...
-Both **LogStore** and **KvStore** adhere to the same bucket-splitting and tablet allocation policies. As a result, **LogTablets** and **KvTablets** with the same `tablet_id` are always allocated to the same **TabletServer** for efficient data management.
+Data for the LogManager and KvManager is stored within tablets. Each tablet consists of a LogTablet and, optionally, a KvTablet, depending on whether the table supports updates. 
+Both LogManager and KvManager adhere to the same bucket-splitting and tablet allocation policies. As a result, LogTablets and KvTablets with the same `tablet_id` are always allocated to the same TabletServer for efficient data management.
 
-The **LogTablet** supports multiple replicas based on the table's configured replication factor, ensuring high availability and fault tolerance. **Currently, replication is not supported for KvTablets**.
+The LogTablet supports multiple replicas based on the table's configured replication factor, ensuring high availability and fault tolerance. Currently, replication is not supported for KvTablets.
```
wuchong, line 42: It may confuse users that KvTablet doesn't support fault tolerance. How about this? "The LogTablet supports multiple replicas based on the table's configured replication factor, ensuring high availability and fault tolerance. In contrast, KvTablet maintains a single replica but achieves high availability and fault tolerance by periodically snapshotting data to remote storage. In the event of a failure, KvTablet can recover by restoring from the snapshot and replaying logs from the LogTablet."

We can add separate sections to introduce "Log Replication", "Kv Snapshot" under the "concepts". Like Flink: https://nightlies.apache.org/flink/flink-docs-release-1.20/docs/concepts/stateful-stream-processing/#state-persistence

website/src/css/custom.css
```diff
@@ -44,11 +44,11 @@
 
 
 .navbar__brand {
-  font-family: monaco;
+  font-family: sans-serif;
 }
 
 .hero__title {
-  font-family: monaco;
+  font-family: sans-serif;
```
wuchong, line 51: The original `monaco` is used to have a special look for our "Fluss" word in the nav bar, to use it as a temporary logo.

website/src/css/custom.css
```diff
@@ -30,7 +30,7 @@
   --ifm-color-primary-lighter: #0682ff;
   --ifm-color-primary-lightest: #2893ff;
   --ifm-code-font-size: 90%;
-  --ifm-font-family-base: "PingFang SC",Segoe UI,Roboto,Ubuntu,Cantarell,Helvetica,sans-serif,Apple Color Emoji,Segoe UI Emoji,Android Emoji,EmojiSymbols,EmojiOne Mozilla,Twemoji Mozilla,Segoe UI Symbol,Noto Color Emoji;
+  --ifm-font-family-base: "sans-serif", Helvetica, PingFang SC,Segoe UI,Roboto,Ubuntu,Cantarell,Helvetica,sans-serif,Apple Color Emoji,Segoe UI Emoji,Android Emoji,EmojiSymbols,EmojiOne Mozilla,Twemoji Mozilla,Segoe UI Symbol,Noto Color Emoji;
```
wuchong, line 33: Is there any purpose in changing the font family? I'm not an expert on UI, so just curious about any problem with the original fonts.

polyzos: @wuchong The main reason is that it just renders better, especially for blog posts and its more familiar to the eye

website/docs/concepts/architecture.md
polyzos: @wuchong I also like LogStore and KvStore more, i just thought it would be best to have consistency with the wording used in the source code to avoid confusion if users need to look into the code

wuchong: I created https://github.com/alibaba/fluss/issues/390 to do the renaming.

polyzos:
closing this for now

PR #351, opened 2025-02-06
[server] Support addPartition and dropPartition

swuferhong:
### Purpose

Linked issue: https://github.com/alibaba/fluss/issues/342

In this pr, we will introduce api addPartition and dropPartition in `Admin` to support add partition and drop partition for static  partition table. The static partition table is different from the auto partition table; it is not automatically created by the server based on rules. Instead, partitions can only be created by the user through an `addPartition` request.

To simplify the implementation, this PR only supports add/drop partitions for static partition tables, while auto partition tables do not support this feature.


swuferhong:
@wuchong  @luoyuxia  pr ready. Could you help review it?

fluss-client/src/main/java/com/alibaba/fluss/client/admin/Admin.java
```diff
@@ -328,4 +331,50 @@ ListOffsetsResult listOffsets(
...
     /** Describe the lake used for lakehouse storage. */
     CompletableFuture<LakeStorageInfo> describeLakeStorage();
+
+    /**
+     * Add a partition to a partitioned table.
+     *
+     * <p>The following exceptions can be anticipated when calling {@code get()} on returned future.
+     *
+     * <ul>
+     *   <li>{@link TableNotExistException} if the table does not exist and {@code ignoreIfExists}
```
SteNicholas: Why does `ignoreIfExists` take effect on non-existed table?

fluss-client/src/main/java/com/alibaba/fluss/client/admin/Admin.java
```diff
@@ -328,4 +331,50 @@ ListOffsetsResult listOffsets(
...
+    CompletableFuture<Void> addPartition(
+            TablePath tablePath, PartitionSpec partitionSpec, boolean ignoreIfExists);
+
+    /**
+     * Drop a partition from a partitioned table.
+     *
+     * <p>The following exceptions can be anticipated when calling {@code get()} on returned future.
+     *
+     * <ul>
+     *   <li>{@link TableNotExistException} if the table does not exist and {@code ignoreIfExists}
```
SteNicholas: Ditto.

fluss-client/src/test/java/com/alibaba/fluss/client/admin/FlussAdminITCase.java
```diff
@@ -563,6 +557,178 @@ void testGetServerNodes() throws Exception {
...
+
+        // drop one partition.
+        admin.dropPartition(
+                        tablePath, new PartitionSpec(Collections.singletonMap("age", "10")), false)
+                .get();
+        assertPartitionInfo(
+                admin.listPartitionInfos(tablePath).get(), Collections.singletonList("11"));
+
+        // test add partition already exists with ignoreIfNotExists = false.
+        assertThatThrownBy(
```
SteNicholas, line 644: Could this also verify other exceptions like `TableNotExistException`?

fluss-client/src/main/java/com/alibaba/fluss/client/admin/Admin.java
swuferhong: Sry, write one wrong comment. Done!

fluss-client/src/main/java/com/alibaba/fluss/client/admin/Admin.java
swuferhong: Done!

fluss-client/src/test/java/com/alibaba/fluss/client/admin/FlussAdminITCase.java
swuferhong: Done!

fluss-client/src/test/java/com/alibaba/fluss/client/admin/FlussAdminITCase.java
```diff
@@ -562,6 +556,206 @@ void testGetServerNodes() throws Exception {
         assertThat(serverNodes).containsExactlyInAnyOrderElementsOf(expectedNodes);
     }
 
+    @Test
```
SteNicholas, line 568: Add test cases for `SHOW PARTITIONS table_name` to verify the support for `FlinkCatalog#listPartitions`.

fluss-client/src/main/java/com/alibaba/fluss/client/admin/Admin.java
```diff
@@ -328,4 +331,48 @@ ListOffsetsResult listOffsets(
...
+     *   <li>{@link PartitionSpecInvalidException} if the input partition spec is invalid.
+     * </ul>
+     *
+     * @param tablePath The table path of the table.
+     * @param partitionSpec The partition spec to add.
+     * @param ignoreIfExists Flag to specify behavior when a partition with the given name already
+     *     exists: if set to false, throw a PartitionAlreadyExistsException, if set to true, do
+     *     nothing.
+     */
+    CompletableFuture<Void> addPartition(
```
luoyuxia: nit: i'd like suggest to name it to `createPartition` to align to `dropPartition`.

fluss-client/src/main/java/com/alibaba/fluss/client/admin/Admin.java
```diff
@@ -328,4 +331,48 @@ ListOffsetsResult listOffsets(
...
+     * Drop a partition from a partitioned table.
+     *
+     * <p>The following exceptions can be anticipated when calling {@code get()} on returned future.
+     *
+     * <ul>
+     *   <li>{@link TableNotExistException} if the table does not exist.
+     *   <li>{@link TableNotPartitionedException} if the table is not partitioned.
+     *   <li>{@link PartitionNotExistException} if the partition not exists and {@code
+     *       ignoreIfExists} is false.
+     *   <li>{@link PartitionSpecInvalidException} if the input partition spec is invalid.
```
luoyuxia: Do we really need`PartitionSpecInvalidException` exception? Any `PartitionSpecInvalidException` should also be `PartitionNotExistException`, and I think user may not care about partition spec is valid or not in dropPartition method.

fluss-client/src/test/java/com/alibaba/fluss/client/admin/FlussAdminITCase.java
```diff
@@ -562,6 +556,206 @@ void testGetServerNodes() throws Exception {
...
+                                        .column("age$", DataTypes.INT())
+                                        .build())
+                        .distributedBy(3, "id")
+                        .partitionedBy("age$")
+                        .build();
+        TablePath tablePath = TablePath.of(dbName, "test_create_illegal_partitioned_table_1");
+        assertThatThrownBy(() -> admin.createTable(tablePath, partitionedTable, true).get())
+                .cause()
+                .isInstanceOf(InvalidTableException.class)
+                .hasMessageContaining("Partition key should not contains separator: '$'");
```
luoyuxia: I'm wondering why Partition key should not contains separator: '$'.?

fluss-client/src/test/java/com/alibaba/fluss/client/admin/FlussAdminITCase.java
```diff
@@ -617,6 +811,21 @@ private void assertBucketSnapshot(
         assertThat(snapshotMetadata.getLogOffset()).isEqualTo(expectedSnapshot.getLogOffset());
     }
 
+    private void assertPartitionInfo(
+            List<PartitionInfo> partitionInfos, List<String> expectedPartitionNames) {
+        if (expectedPartitionNames.isEmpty()) {
```
luoyuxia: Can these piece of code simplify to 
```
assertThat(partitionInfos.size()).isEqualTo(expectedPartitionNames.size());
        assertThat(
                        partitionInfos.stream()
                                .map(PartitionInfo::getPartitionName)
                                .collect(Collectors.toList()))
                .containsExactlyInAnyOrderElementsOf(expectedPartitionNames);
```

fluss-client/src/test/java/com/alibaba/fluss/client/table/AutoPartitionedTableITCase.java
```diff
@@ -341,6 +346,56 @@ void testOperateNotExistPartitionShouldThrowException() throws Exception {
...
+                        tablePath,
+                        new PartitionSpec(
+                                Collections.singletonMap("c", String.valueOf(currentYear + 10))),
+                        false)
+                .get();
+        Map<String, Long> partitionIdByNames =
+                FLUSS_CLUSTER_EXTENSION.waitUtilPartitionAllReady(tablePath, 5);
+
+        List<PartitionInfo> partitionInfos = admin.listPartitionInfos(tablePath).get();
+        assertThat(partitionInfos.size()).isEqualTo(5);
```
luoyuxia: nit: simplify to 
```
List<PartitionInfo> partitionInfos = admin.listPartitionInfos(tablePath).get();
        List<String> expectedPartitions = new ArrayList<>(partitionIdByNames.keySet());
        expectedPartitions.add(String.valueOf(currentYear + 10));
        assertThat(
                        partitionInfos.stream()
                                .map(PartitionInfo::getPartitionName)
                                .collect(Collectors.toList()))
                .containsExactlyInAnyOrderElementsOf(expectedPartitions);
```
?

fluss-client/src/test/java/com/alibaba/fluss/client/table/AutoPartitionedTableITCase.java
```diff
@@ -341,6 +346,56 @@ void testOperateNotExistPartitionShouldThrowException() throws Exception {
...
+                                String.valueOf(currentYear + 1),
+                                String.valueOf(currentYear + 2),
+                                String.valueOf(currentYear + 3),
+                                String.valueOf(currentYear + 10)));
+
+        Table table = conn.getTable(tablePath);
+        AppendWriter appendWriter = table.newAppend().createWriter();
+        int recordsPerPartition = 5;
+        Map<Long, List<InternalRow>> expectPartitionAppendRows = new HashMap<>();
+        for (String partition : partitionIdByNames.keySet()) {
```
luoyuxia, line 360: Also append rows to added partition in `addPartition` api?

fluss-common/src/main/java/com/alibaba/fluss/exception/PartitionAlreadyExistsException.java
```diff
@@ -0,0 +1,35 @@
...
+ *
+ * @since 0.6
+ */
+@PublicEvolving
+public class PartitionAlreadyExistsException extends ApiException {
+    public PartitionAlreadyExistsException(String message) {
+        this(message, null);
+    }
+
+    public PartitionAlreadyExistsException(String message, Throwable cause) {
```
luoyuxia: nit: remove this method. I can't image when the cause is needed for `PartitionAlreadyExistsException`.

fluss-common/src/main/java/com/alibaba/fluss/exception/PartitionSpecInvalidException.java
```diff
@@ -0,0 +1,38 @@
...
+ * @since 0.6
+ */
+@PublicEvolving
+public class PartitionSpecInvalidException extends ApiException {
+
+    public PartitionSpecInvalidException(String message) {
+        this(message, null);
+    }
+
+    public PartitionSpecInvalidException(String message, Throwable cause) {
```
luoyuxia: dito

fluss-connectors/fluss-connector-flink/src/main/java/com/alibaba/fluss/connector/flink/catalog/FlinkCatalog.java
```diff
@@ -371,8 +379,42 @@ public void alterTable(ObjectPath objectPath, CatalogBaseTable catalogBaseTable,
...
+
+        try {
+            TablePath tablePath = toTablePath(objectPath);
+            TableInfo tableInfo = admin.getTableInfo(tablePath).get();
+            List<String> partitionKeys = tableInfo.getPartitionKeys();
+
+            List<PartitionInfo> partitionInfos = admin.listPartitionInfos(tablePath).get();
+            List<CatalogPartitionSpec> catalogPartitionSpecs = new ArrayList<>();
+            for (PartitionInfo partitionInfo : partitionInfos) {
+                String[] partitionSpec = partitionInfo.getPartitionName().split("\\$");
```
luoyuxia: nit: rename to `partitionValues`, partitionSpec always make me feel like it's key/value pair

fluss-connectors/fluss-connector-flink/src/main/java/com/alibaba/fluss/connector/flink/catalog/FlinkCatalog.java
```diff
@@ -412,14 +454,51 @@ public void createPartition(
...
+            admin.addPartition(toTablePath(objectPath), partitionSpec, b).get();
+        } catch (Exception e) {
+            Throwable t = ExceptionUtils.stripExecutionException(e);
+            if (isTableNotExist(t)) {
+                throw new TableNotExistException(getName(), objectPath);
+            } else if (isTableNotPartitioned(t)) {
+                throw new TableNotPartitionedException(getName(), objectPath);
+            } else if (isPartitionSpecInvalid(t)) {
+                throw new PartitionSpecInvalidException(
+                        getName(), null, objectPath, catalogPartitionSpec, e);
```
luoyuxia: why `partitionKeys` is passed as null?

fluss-connectors/fluss-connector-flink/src/main/java/com/alibaba/fluss/connector/flink/catalog/FlinkCatalog.java
```diff
@@ -412,14 +454,51 @@ public void createPartition(
...
+        PartitionSpec partitionSpec = new PartitionSpec(catalogPartitionSpec.getPartitionSpec());
+        try {
+            admin.dropPartition(toTablePath(objectPath), partitionSpec, b).get();
+        } catch (Exception e) {
+            Throwable t = ExceptionUtils.stripExecutionException(e);
+            if (isPartitionNotExist(t)) {
+                throw new PartitionNotExistException(
+                        getName(), objectPath, catalogPartitionSpec, e);
+            } else {
+                throw new CatalogException(
```
luoyuxia, line 515: What if `TableNotExistException`?

luoyuxia (commented):
Left some comments.

fluss-common/src/main/java/com/alibaba/fluss/exception/PartitionSpecInvalidException.java
```diff
@@ -0,0 +1,38 @@
...
+
+/**
+ * Exception for invalid PartitionSpec compared with partition key list of a partitioned Table. For
+ * example, it is thrown when the size of PartitionFiledSpec in request AddPartition exceeds the
+ * size of partition key list.
+ *
+ * @since 0.6
+ */
+@PublicEvolving
+public class PartitionSpecInvalidException extends ApiException {
```
wuchong: Can be renamed to `InvalidPartitionException` to align with `InvalidTableException` and `InvalidDatabaseException`. From the perspective of a user, users don't know what is a partition spec, partition spec is just a structure to describe a partition. So it's the partition is invalid underlying.

fluss-client/src/main/java/com/alibaba/fluss/client/admin/Admin.java
wuchong: +1 

And move the methods after `listPartitionInfos`.

fluss-common/src/main/java/com/alibaba/fluss/utils/PartitionUtils.java
```diff
@@ -0,0 +1,148 @@
...
+    public static String getPartitionName(
+            List<String> partitionKeys, List<String> partitionSpecs, boolean checkValid) {
+        checkArgument(
+                partitionKeys.size() == partitionSpecs.size(),
+                "The number of partition keys and partition specs should be the same.");
+
+        for (String value : partitionSpecs) {
+            if (checkValid && value.contains(PARTITION_SPEC_SEPARATOR)) {
+                throw new PartitionSpecInvalidException(
+                        "The value of partition key should not contains separator: '"
```
wuchong: nit:
```suggestion
                        "The partition value should not contain separator: '"
```

fluss-common/src/main/java/com/alibaba/fluss/utils/PartitionUtils.java
```diff
@@ -0,0 +1,148 @@
...
+    }
+
+    public static String getPartitionName(
+            List<String> partitionKeys, List<String> partitionSpecs, boolean checkValid) {
+        checkArgument(
+                partitionKeys.size() == partitionSpecs.size(),
+                "The number of partition keys and partition specs should be the same.");
+
+        for (String value : partitionSpecs) {
+            if (checkValid && value.contains(PARTITION_SPEC_SEPARATOR)) {
```
wuchong: This is a too simple check, we can use `TablePath#detectInvalidName` to check invalid characters.

fluss-client/src/main/java/com/alibaba/fluss/client/table/getter/PartitionGetter.java
```diff
@@ -18,15 +18,19 @@
...
+import java.util.Collections;
 import java.util.List;
 
+import static com.alibaba.fluss.utils.PartitionUtils.getPartitionName;
+
 /** A getter to get partition name from a row. */
 public class PartitionGetter {
 
+    // TODO currently, only support one partition key.
+    private final String partitionKeyName;
```
wuchong: Just name it `partitionKey` to avoid confusion with "partition name" concept. In the future, the member field is `List<String> partitionKeys`.

fluss-client/src/test/java/com/alibaba/fluss/client/admin/FlussAdminITCase.java
wuchong: Why partition key doesn't allow separator `$`?

fluss-client/src/test/java/com/alibaba/fluss/client/admin/FlussAdminITCase.java
wuchong: Even remove `assertThat(partitionInfos.size()).isEqualTo(expectedPartitionNames.size())` for better assertion error messages. The later assertion already asserts the size.

fluss-client/src/test/java/com/alibaba/fluss/client/admin/FlussAdminITCase.java
```diff
@@ -562,6 +556,206 @@ void testGetServerNodes() throws Exception {
...
+                        .distributedBy(3, "id")
+                        .partitionedBy("age")
+                        .build();
+        TablePath tablePath = TablePath.of(dbName, "test_add_and_drop_partitioned_table");
+        admin.createTable(tablePath, partitionedTable, true).get();
+        assertPartitionInfo(admin.listPartitionInfos(tablePath).get(), Collections.emptyList());
+
+        // add two partitions.
+        admin.addPartition(
+                        tablePath, new PartitionSpec(Collections.singletonMap("age", "10")), false)
```
wuchong: Introduce a `newPartitionSpec()` method to simplify the verbose code of creating partition spec.

```java
    private static PartitionSpec newPartitionSpec(String partitionKey, String partitionValue) {
        return new PartitionSpec(Collections.singletonMap(partitionKey, partitionValue));
    }
```

fluss-client/src/test/java/com/alibaba/fluss/client/admin/FlussAdminITCase.java
```diff
@@ -562,6 +556,206 @@ void testGetServerNodes() throws Exception {
...
+                        .build();
+        TablePath tablePath = TablePath.of(dbName, "test_create_illegal_partitioned_table_1");
+        assertThatThrownBy(() -> admin.createTable(tablePath, partitionedTable, true).get())
+                .cause()
+                .isInstanceOf(InvalidTableException.class)
+                .hasMessageContaining("Partition key should not contains separator: '$'");
+    }
+
+    @Test
+    void testAddAndDropPartitionsForStaticPartition() throws Exception {
```
wuchong: Add a test that add partition with wrong partition key, such as `PartitionSpec("name", "10")`

fluss-client/src/test/java/com/alibaba/fluss/client/table/AutoPartitionedTableITCase.java
wuchong: `assertThat(partitionInfos.size()).isEqualTo(5);` can be remove for better assertion error message if size is not equal.

fluss-client/src/test/java/com/alibaba/fluss/client/table/StaticPartitionedTableITCase.java
```diff
@@ -0,0 +1,229 @@
...
+import java.util.List;
+import java.util.Map;
+
+import static com.alibaba.fluss.testutils.DataTestUtils.keyRow;
+import static com.alibaba.fluss.testutils.DataTestUtils.row;
+import static com.alibaba.fluss.testutils.InternalRowAssert.assertThatRow;
+import static org.assertj.core.api.Assertions.assertThat;
+
+/** IT case for Fluss static partitioned table. */
+class StaticPartitionedTableITCase extends ClientToServerITCaseBase {
```
wuchong: Let's avoid using `static partitioned table`, because the opposites of `auto` is not `static`, and the opposites of `static` is `dynamic`. But Dynamic Partitioning can also works on non-auto-partitioned table ("static partitioned table" in your naming). That is very confused. 

So just name it `PartitionedTableITCase`.

fluss-client/src/test/java/com/alibaba/fluss/client/table/StaticPartitionedTableITCase.java
```diff
@@ -0,0 +1,229 @@
...
+        upsertWriter.flush();
+
+        Lookuper lookuper = table.newLookup().createLookuper();
+        // now, let's lookup the written data by look up.
+        for (PartitionInfo partitionInfo : partitionInfos) {
+            String partitionName = partitionInfo.getPartitionName();
+            for (int j = 0; j < recordsPerPartition; j++) {
+                InternalRow actualRow = row(j, "a" + j, partitionName);
+                InternalRow lookupRow =
+                        lookuper.lookup(keyRow(schema, new Object[] {j, null, partitionName}))
```
wuchong: Can use GenericRow here `row(j, partitionName)` which is much simpler.

fluss-common/src/main/java/com/alibaba/fluss/utils/PartitionUtils.java
```diff
@@ -0,0 +1,148 @@
...
+     * spec
+     * </pre>
+     *
+     * <p>For example, if the partition keys are [a], and the partition spec is [1], the partition
+     * name is "1".
+     *
+     * @param partitionKeys the partition keys
+     * @param partitionSpecs the partition specs
+     */
+    public static String getPartitionName(List<String> partitionKeys, List<String> partitionSpecs) {
```
wuchong: `partitionSpecs` -> `partitionValues`

Partition Sepc = partiton keys + partition values

Besides, personally, I think it's verbose to pass in partition keys, the partition name can be generated without partition keys.

fluss-common/src/main/java/com/alibaba/fluss/utils/PartitionUtils.java
```diff
@@ -0,0 +1,148 @@
...
+    /**
+     * Returns the partition name for a partition table of specify partition spec.
+     *
+     * <p>The partition name is in the following format:
+     *
+     * <pre>
+     * spec1$spec2$...$specN
+     * </pre>
+     *
+     * <p>For example, if the partition keys are [a, b, c], and the partition spec is [1, 2, 3], the
```
wuchong:
```suggestion
     * <p>For example, if the partition keys are [a, b, c], and the partition values are [1, 2, 3], the
```

fluss-common/src/main/java/com/alibaba/fluss/metadata/PartitionSpec.java
```diff
@@ -0,0 +1,69 @@
...
+        this.partitionSpec = Collections.unmodifiableMap(partitionSpec);
+    }
+
+    public Map<String, String> getPartitionSpec() {
+        return partitionSpec;
+    }
+
+    @Override
+    public String toString() {
+        return "PartitionSpec{" + "partitionSpec=" + partitionSpec + '}';
```
wuchong: nit: can simplify to `"PartitionSpec{" + partitionSpec + '}'`

fluss-common/src/main/java/com/alibaba/fluss/metadata/TablePath.java
```diff
@@ -180,6 +184,12 @@ private static boolean containsInvalidPattern(String identifier) {
                             || (c >= 'A' && c <= 'Z')
                             || c == '_'
                             || c == '-';
+
+            // partition name can contain '$' as separator.
+            if (isPartitionName && (c == '$')) {
```
wuchong: I think we can also allow `$` as database name and table name, so can simplify the logic here.

fluss-connectors/fluss-connector-flink/src/main/java/com/alibaba/fluss/connector/flink/catalog/FlinkCatalog.java
```diff
@@ -371,8 +379,42 @@ public void alterTable(ObjectPath objectPath, CatalogBaseTable catalogBaseTable,
     @Override
     public List<CatalogPartitionSpec> listPartitions(ObjectPath objectPath)
             throws TableNotExistException, TableNotPartitionedException, CatalogException {
-        // TODO: use admin.listPartitionInfos()
-        return Collections.emptyList();
+        // TODO lake table should support.
+        if (objectPath.toString().contains("$lake")) {
```
wuchong: `if (objectPath.getObjectName().contains(LAKE_TABLE_SPLITTER)) {`

fluss-connectors/fluss-connector-flink/src/main/java/com/alibaba/fluss/connector/flink/catalog/FlinkCatalog.java
wuchong: +1

fluss-connectors/fluss-connector-flink/src/test/java/com/alibaba/fluss/connector/flink/catalog/FlinkCatalogITCase.java
```diff
@@ -154,17 +159,17 @@ void testCreateTable() throws Exception {
...
                                         "create table test_table_unsupported"
-                                                + " (a int, b int) partitioned by (b)"))
+                                                + " (a int, $b int) partitioned by ($b)"))
                 .cause()
                 .isInstanceOf(CatalogException.class)
                 .cause()
-                .isInstanceOf(InvalidConfigException.class)
-                .hasMessageContaining("Currently, partitioned table must enable auto partition");
+                .isInstanceOf(InvalidTableException.class)
+                .hasMessageContaining("Partition key should not contains separator: '$'");
```
wuchong: Why this is not supported?

fluss-connectors/fluss-connector-flink/src/test/java/com/alibaba/fluss/connector/flink/catalog/FlinkCatalogITCase.java
```diff
@@ -359,4 +438,23 @@ private static void assertOptionsEqual(
         assertThat(actualOptions.size()).isEqualTo(expectedOptions.size());
         assertThat(actualOptions).isEqualTo(expectedOptions);
     }
+
+    private void assertCatalogPartitionSpecsEquals(
+            List<CatalogPartitionSpec> catalogPartitionSpecs,
+            List<String> partitionKeys,
+            List<String> expectedPartitionNames) {
```
wuchong: This assertion method is quite complex. An assertion should be so complex, otherwise, there may be bugs in the assertion. For example, this code forgot to check that the `CatalogPartitionSpec` keys may have more than `partitionKeys`. 

An assertion should be simplified by checking the actual values, rather than transforming on the actual values and then compare. You can just generate actual `CatalogPartitionSpec` using the `partitionKeys` and `partitionValues`. `partitionKeys` should be a nested List, but we can simplify to one level `List<String>` by only allowing single partition key (`String partitionkey, List<String> expectedPartitionValues`).

fluss-connectors/fluss-connector-flink/src/test/java/com/alibaba/fluss/connector/flink/sink/FlinkTableSinkITCase.java
```diff
@@ -484,28 +484,137 @@ void testIgnoreDelete(boolean isPrimaryKeyTable) throws Exception {
...
+                                "INSERT INTO %s PARTITION (c = 2000) values (22, 2222), (33, 3333)",
+                                tableName))
+                .await();
+        assertResultsIgnoreOrder(
+                rowIter, Arrays.asList("+I[22, 2222, 2000]", "+I[33, 3333, 2000]"), true);
+    }
+
+    @ParameterizedTest
+    @ValueSource(booleans = {true, false})
+    void testWritePartitionedPrimaryKeyTable(boolean isStaticPartitionedTable) throws Exception {
```
wuchong: There is no concept of "static partitioned table", use `isAutoPartition` flag.

wuchong: Seem there is no much difference with `testWritePartitionedLogTable`. Combine them into one test method?

fluss-rpc/src/main/proto/FlussApi.proto
```diff
@@ -382,6 +382,24 @@ message ListPartitionInfosResponse {
   repeated PbPartitionInfo partitions_info = 1;
 }
 
+// add partition request and response
+message AddPartitionRequest {
```
wuchong: `CreatePartitionRequest` to align with `CreateTableRequest`.

fluss-rpc/src/main/proto/FlussApi.proto
```diff
@@ -718,6 +736,11 @@ message PbPartitionInfo {
   required string partition_name = 2;
 }
 
+message PbPartitionSpec {
+  required string partition_key = 1;
+  required string value = 2;
```
wuchong: A partition spec is a Map of partition keys and partition values. So should be :

```
message PbPartitionSpec {
  repeated PbKeyValue partition_key_values = 1;
}
```

fluss-server/src/main/java/com/alibaba/fluss/server/coordinator/event/DropTableEvent.java
```diff
@@ -23,18 +23,25 @@ public class DropTableEvent implements CoordinatorEvent {
...
         this.tableId = tableId;
+        this.isPartitionTable = isPartitionTable;
         this.isAutoPartitionTable = isAutoPartitionTable;
     }
 
     public long getTableId() {
         return tableId;
     }
 
+    public boolean isPartitionTable() {
```
wuchong: `isPartitionedTable`

fluss-server/src/main/java/com/alibaba/fluss/server/coordinator/PartitionManager.java
```diff
@@ -0,0 +1,461 @@
...
+ *
+ * <p>For static partition tables, it's responsible for adding partitions to or drop partitions from
+ * static partition table.
+ *
+ * <p>For auto partition tables, it's responsible not only for adding partitions to or drop
+ * partitions from auto partition table, but also triggering auto partition for these auto partition
+ * tables in cluster periodically. It'll use a {@link ScheduledExecutorService} to schedule the auto
+ * partition which trigger auto partition for them.
+ */
+public class PartitionManager implements AutoCloseable {
```
wuchong: I don't think we need such big refactoring for the `AutoPartitionManager`. The `addPartition` and `dropPartition` are simple metadata operations on zookeeper (should be supported on `MetadataManager` just like `create/dropTable`). If it is a auto-partition table, `TableChangeWatcher` watches the changes of partition creation and deletion and pushes the events to   `AutoPartitionManager`, so no changes to `AutoPartitionManager`. If it is not a auto-partition table, then no changes to `AutoPartitionManager.

fluss-server/src/main/java/com/alibaba/fluss/server/coordinator/CoordinatorService.java
```diff
@@ -197,6 +203,49 @@ public CompletableFuture<DropTableResponse> dropTable(DropTableRequest request)
         return CompletableFuture.completedFuture(response);
     }
 
+    @Override
+    public CompletableFuture<AddPartitionResponse> addPartition(AddPartitionRequest request) {
+        AddPartitionResponse response = new AddPartitionResponse();
+        TablePath tablePath = toTablePath(request.getTablePath());
+        if (!metadataManager.tableExists(tablePath)) {
```
wuchong: we can remove this check, as the `metadataManager#getTable` will also check the existence.

fluss-server/src/main/java/com/alibaba/fluss/server/utils/TableDescriptorValidation.java
```diff
@@ -170,49 +173,68 @@ private static void checkTieredLog(Configuration tableConf) {
...
-
         if (isPartitioned) {
+            partitionKeys.forEach(
+                    partitionKey -> {
+                        if (partitionKey.contains(PARTITION_SPEC_SEPARATOR)) {
+                            throw new InvalidTableException(
+                                    "Partition key should not contains separator: '"
+                                            + PARTITION_SPEC_SEPARATOR
+                                            + "'");
+                        }
```
wuchong: shuold be remove, we don't have such limition.

fluss-server/src/main/java/com/alibaba/fluss/server/utils/TableDescriptorValidation.java
```diff
@@ -170,49 +173,68 @@ private static void checkTieredLog(Configuration tableConf) {
...
+    private static void checkPartition(
             Configuration tableConf, List<String> partitionKeys, RowType rowType) {
         boolean isPartitioned = !partitionKeys.isEmpty();
         AutoPartitionStrategy autoPartition = AutoPartitionStrategy.from(tableConf);
-
-        if (!isPartitioned && autoPartition.isAutoPartitionEnabled()) {
-            throw new InvalidConfigException(
-                    String.format(
-                            "Currently, auto partition is only supported for partitioned table, please set table property '%s' to false.",
-                            ConfigOptions.TABLE_AUTO_PARTITION_ENABLED.key()));
```
wuchong: I think this check should still be kept.

fluss-server/src/main/java/com/alibaba/fluss/server/coordinator/event/DropTableEvent.java
```diff
@@ -23,18 +23,25 @@ public class DropTableEvent implements CoordinatorEvent {
 
     private final long tableId;
 
+    private final boolean isPartitionTable;
+
     // true if the table is with auto partition enabled
     private final boolean isAutoPartitionTable;
 
-    public DropTableEvent(long tableId, boolean isAutoPartitionTable) {
+    public DropTableEvent(long tableId, boolean isPartitionTable, boolean isAutoPartitionTable) {
```
wuchong: Add a check on the parameters.

```java
        if (isAutoPartitionTable) {
            checkArgument(isPartitionTable, "Auto partition table must be partitioned table.");
        }
```

fluss-connectors/fluss-connector-flink/src/main/java/com/alibaba/fluss/connector/flink/catalog/FlinkCatalog.java
```diff
@@ -371,8 +379,42 @@ public void alterTable(ObjectPath objectPath, CatalogBaseTable catalogBaseTable,
...
+                throw new TableNotExistException(getName(), objectPath);
+            } else if (isTableNotPartitioned(t)) {
+                throw new TableNotPartitionedException(getName(), objectPath);
+            } else {
+                throw new CatalogException(
+                        String.format(
+                                "Failed to list partitions of table %s in %s",
+                                objectPath, getName()),
+                        t);
+            }
```
wuchong, line 418: Cover the exceptions in `FlinkCatalogTest`

fluss-client/src/test/java/com/alibaba/fluss/client/admin/FlussAdminITCase.java
```diff
@@ -562,6 +556,206 @@ void testGetServerNodes() throws Exception {
...
+                                admin.addPartition(
+                                                tablePath,
+                                                new PartitionSpec(
+                                                        Collections.singletonMap("age", "11")),
+                                                false)
+                                        .get())
+                .cause()
+                .isInstanceOf(PartitionAlreadyExistsException.class)
+                .hasMessageContaining(
+                        "Partition '11' already exists for table test_db.test_add_and_drop_partitioned_table");
```
wuchong: I suggest printing the fully qualified partition name (`"age=11"`) for errors and logs which is much more user-friendly. The partition name is just an internal identifier. 

We can introduce `getPartitionQualifiedName` util method in `PartitionUtils`:

```java

    /**
     * Returns the qualified partition name for a partition spec (partition keys and partition
     * values). The qualified partition name is usually used as the partition directory path, but is
     * used as a pretty display name of a partition. The qualified partition name is in the
     * following format:
     *
     * <pre>
     * key1=value1/key2=value2/.../keyN=valueN
     * </pre>
     */
    public static String getPartitionQualifiedName(
            List<String> partitionKeys, List<String> partitionValues) {
        checkArgument(
                partitionKeys.size() == partitionValues.size(),
                "The number of partition keys and partition values should be the same.");
        StringBuilder sb = new StringBuilder();
        for (int i = 0; i < partitionKeys.size(); i++) {
            sb.append(partitionKeys.get(i)).append("=").append(partitionValues.get(i));
            if (i != partitionKeys.size() - 1) {
                sb.append("/");
            }
        }
        return sb.toString();
    }
```

fluss-common/src/test/java/com/alibaba/fluss/utils/PartitionUtilsTest.java
```diff
@@ -17,47 +17,76 @@
...
+        assertThatThrownBy(
+                        () ->
+                                getPartitionName(
+                                        Arrays.asList("a", "b", "c"), Arrays.asList("1", "2")))
+                .isInstanceOf(IllegalArgumentException.class)
+                .hasMessageContaining(
+                        "The number of partition keys and partition specs should be the same.");
+
+        assertThatThrownBy(
+                        () -> getPartitionName(Arrays.asList("a", "b"), Arrays.asList("$1", "2")))
```
wuchong: Personally, I think the `getPartitionName` should just does getter/generate thing. The validation should be done in a separate method. This will make logic cleaner and make the `getPartiionName` better performance.

fluss-common/src/main/java/com/alibaba/fluss/utils/PartitionUtils.java
```diff
@@ -0,0 +1,148 @@
...
+public class PartitionUtils {
+    public static final String PARTITION_SPEC_SEPARATOR = "$";
+
+    public static final List<DataTypeRoot> STATIC_PARTITION_KEY_SUPPORTED_TYPES =
+            Arrays.asList(
+                    DataTypeRoot.STRING,
+                    DataTypeRoot.INTEGER,
+                    DataTypeRoot.BOOLEAN,
+                    DataTypeRoot.DATE,
+                    DataTypeRoot.TIME_WITHOUT_TIME_ZONE);
```
wuchong: Let's support other data types in other PRs. Currently, let's only support STRING type. The problem here is the partition value -> string conversion. There is no a standard how to convert a DATE or TIME into string, the Java `Data/Time#toString()` is not a common standard and usually has problems in partitions (e.g., contains special characters `:`).

wuchong (commented):
Thanks @swuferhong , I think there are 2 main problems of this PR.

1. Let's make the concepts clear of different partitoning concepts:
  - **Partitioned Table**: A table that has partitioned fields (aka. partition keys). 
  - **Auto Partitioned Table**: A Partitioned Table that enabled **auto partition**, the table will automatically generate new partitions and drop old partitions by current time. 
  - **Dynamic Partitioning**: Dynamically creates the partition when inserting records into the partition table. (i.e. INSERT INTO table without PARTITION clause)
  - **Static Partitioning**: Insert into pre-created partitions. (i.g. INSERT INTO PARTITION(...))

A partitioned table both supported Dynamic Partitioning and Static Partitioning, it depends on how to writing data into the partitioned table. So there is no concept of **Static Partitioned Table**. 

2. We don't need a dedicated manager for general partition creation or deletion. The `AutoPartitionManager` should just care about the scheduler of new partitions creation and old partition deletion.

fluss-client/src/test/java/com/alibaba/fluss/client/admin/FlussAdminITCase.java
wuchong: +1

fluss-client/src/main/java/com/alibaba/fluss/client/admin/Admin.java
swuferhong: Done!

fluss-common/src/main/java/com/alibaba/fluss/exception/PartitionSpecInvalidException.java
swuferhong: Done!

fluss-common/src/main/java/com/alibaba/fluss/exception/PartitionSpecInvalidException.java
swuferhong: Done!

fluss-common/src/main/java/com/alibaba/fluss/utils/PartitionUtils.java
swuferhong: Done!

fluss-client/src/main/java/com/alibaba/fluss/client/table/getter/PartitionGetter.java
swuferhong: Done!

fluss-client/src/test/java/com/alibaba/fluss/client/admin/FlussAdminITCase.java
swuferhong: Here I forgot to make the modification. Initially, when we discussed having multiple partition fields, the pattern for the partition name was set to be 'key1=value1$key2=value2$....'. This check was specifically designed for that format. However, with our current naming convention of 'value1$value2$....', this is no longer needed.

fluss-client/src/main/java/com/alibaba/fluss/client/admin/Admin.java
swuferhong: change `PartitionSpecInvalidException` to `InvalidPartitionException`

fluss-common/src/main/java/com/alibaba/fluss/exception/PartitionAlreadyExistsException.java
swuferhong: Done!

fluss-client/src/test/java/com/alibaba/fluss/client/admin/FlussAdminITCase.java
swuferhong: Done!

fluss-rpc/src/main/proto/FlussApi.proto
swuferhong: Done!

fluss-client/src/test/java/com/alibaba/fluss/client/admin/FlussAdminITCase.java
swuferhong: Done!

fluss-client/src/test/java/com/alibaba/fluss/client/table/AutoPartitionedTableITCase.java
swuferhong: Done!

fluss-client/src/test/java/com/alibaba/fluss/client/table/StaticPartitionedTableITCase.java
swuferhong: Done!

fluss-client/src/test/java/com/alibaba/fluss/client/table/StaticPartitionedTableITCase.java
swuferhong: Done!

fluss-common/src/main/java/com/alibaba/fluss/utils/PartitionUtils.java
swuferhong: Done!

fluss-common/src/main/java/com/alibaba/fluss/utils/PartitionUtils.java
swuferhong: Done!

fluss-common/src/main/java/com/alibaba/fluss/metadata/PartitionSpec.java
swuferhong: Done!

fluss-common/src/test/java/com/alibaba/fluss/utils/PartitionUtilsTest.java
swuferhong: Done!

fluss-server/src/main/java/com/alibaba/fluss/server/coordinator/event/DropTableEvent.java
swuferhong: Done!

fluss-server/src/main/java/com/alibaba/fluss/server/coordinator/event/DropTableEvent.java
swuferhong: Done!

fluss-rpc/src/main/proto/FlussApi.proto
swuferhong: Done!

fluss-server/src/main/java/com/alibaba/fluss/server/utils/TableDescriptorValidation.java
swuferhong: Done!  I misunderstood.

fluss-connectors/fluss-connector-flink/src/test/java/com/alibaba/fluss/connector/flink/sink/FlinkTableSinkITCase.java
swuferhong: Done!

fluss-common/src/main/java/com/alibaba/fluss/metadata/TablePath.java
swuferhong: Done!

fluss-connectors/fluss-connector-flink/src/main/java/com/alibaba/fluss/connector/flink/catalog/FlinkCatalog.java
swuferhong: Done!

fluss-connectors/fluss-connector-flink/src/test/java/com/alibaba/fluss/connector/flink/catalog/FlinkCatalogITCase.java
swuferhong: Here I forgot to make the modification. Initially, when we discussed having multiple partition fields, the pattern for the partition name was set to be 'key1=value1$key2=value2$....'. This check was specifically designed for that format. However, with our current naming convention of 'value1$value2$....', this is no longer needed.

I would remove this test case as we have support other partition table not only auto partitioned table.

fluss-client/src/test/java/com/alibaba/fluss/client/table/AutoPartitionedTableITCase.java
swuferhong: Done!

fluss-client/src/test/java/com/alibaba/fluss/client/admin/FlussAdminITCase.java
swuferhong: Done!

fluss-client/src/test/java/com/alibaba/fluss/client/table/AutoPartitionedTableITCase.java
swuferhong: Already added

fluss-connectors/fluss-connector-flink/src/main/java/com/alibaba/fluss/connector/flink/catalog/FlinkCatalog.java
swuferhong: Done!

fluss-connectors/fluss-connector-flink/src/main/java/com/alibaba/fluss/connector/flink/catalog/FlinkCatalog.java
swuferhong: Done!

fluss-connectors/fluss-connector-flink/src/main/java/com/alibaba/fluss/connector/flink/catalog/FlinkCatalog.java
swuferhong: Flink Catalog interface `Catalog#dropPartition(ObjectPath var1, CatalogPartitionSpec var2, boolean var3) throws PartitionNotExistException, CatalogException;`  does not provide a TableNotExistException error, so here reports the error as a CatalogException.

fluss-server/src/main/java/com/alibaba/fluss/server/coordinator/CoordinatorService.java
swuferhong: Done!

fluss-server/src/main/java/com/alibaba/fluss/server/utils/TableDescriptorValidation.java
swuferhong: Done!

fluss-common/src/main/java/com/alibaba/fluss/utils/PartitionUtils.java
swuferhong: Ok, I have created an issue: https://github.com/alibaba/fluss/issues/489 to trace this, also I added a `todo` comment in code.

fluss-connectors/fluss-connector-flink/src/main/java/com/alibaba/fluss/connector/flink/catalog/FlinkCatalog.java
swuferhong: Done!

fluss-client/src/test/java/com/alibaba/fluss/client/admin/FlussAdminITCase.java
swuferhong: Done!

fluss-connectors/fluss-connector-flink/src/test/java/com/alibaba/fluss/connector/flink/catalog/FlinkCatalogITCase.java
swuferhong: I change to use "show partition xxx" to collect actual results, and use  `assertResultsIgnoreOrder(org.apache.flink.util.CloseableIterator<Row> iterator, List<String> expected, boolean closeIterator)` to assert results.

fluss-common/src/main/java/com/alibaba/fluss/utils/PartitionUtils.java
swuferhong: Done!

swuferhong: Done!

fluss-client/src/test/java/com/alibaba/fluss/client/admin/FlussAdminITCase.java
swuferhong: Done!

fluss-server/src/main/java/com/alibaba/fluss/server/coordinator/PartitionManager.java
swuferhong: Done!

swuferhong:
@wuchong comments addressed

fluss-connectors/fluss-connector-flink/src/main/java/com/alibaba/fluss/connector/flink/catalog/FlinkCatalog.java
```diff
@@ -412,14 +456,61 @@ public void createPartition(
...
+                    TableInfo tableInfo = admin.getTableInfo(tablePath).get();
+                    throw new PartitionSpecInvalidException(
+                            getName(),
+                            tableInfo.getPartitionKeys(),
+                            objectPath,
+                            catalogPartitionSpec,
+                            e);
+                } catch (Exception ee) {
+                    // ignore.
+                }
```
wuchong: This makes the thrown `PartitionSpecInvalidException` ignored. Add test to verify the exception in `PartitionUtils#validatePartitionValues`

We can throw a general `CatalogExecption` if getting partition key failed.

```java
List<String> partitionKeys = null;
                try {
                    TableInfo tableInfo = admin.getTableInfo(tablePath).get();
                    partitionKeys = tableInfo.getPartitionKeys();
                } catch (Exception ee) {
                    // ignore.
                }
                if (partitionKeys != null) {
                    // throw specific partition exception if getting partition keys success.
                    throw new PartitionSpecInvalidException(
                            getName(), partitionKeys, objectPath, catalogPartitionSpec, e);
                } else {
                    // throw general exception if getting partition keys failed.
                    throw new CatalogException(
                            String.format(
                                    "PartitionSpec %s does not match partition keys of table %s in catalog %s.",
                                    partitionSpec, objectPath.getFullName(), catalogName),
                            e);
                }
```

fluss-server/src/main/java/com/alibaba/fluss/server/coordinator/MetadataManager.java
```diff
@@ -308,6 +316,110 @@ public long initWriterId() {
...
+        } catch (Exception e) {
+            throw new FlussRuntimeException(
+                    String.format(
+                            "Fail to get partition '%s' of table %s from zookeeper.",
+                            tablePath, getPartitionQualifiedName(partitionKeys, partitionSpec)),
+                    e);
+        }
+    }
+
+    public void registerPartitionToZk(
```
wuchong: `createPartition(TablePath tablePath, long tableId, PartitionAssignment partitionAssignment, String partitionName, boolean ignoreIfExists)` to align with `createTable`? All the operations in `MetadataManager` work on zookeeper, so I think don't need to mention `ToZK` in the method name.

fluss-server/src/main/java/com/alibaba/fluss/server/coordinator/MetadataManager.java
```diff
@@ -308,6 +316,110 @@ public long initWriterId() {
...
+                    "Partition '"
+                            + getPartitionQualifiedName(partitionKeys, partitionSpec)
+                            + "' does not exist for table "
+                            + tableInfo.getTablePath());
+        }
+
+        deletePartitionFromZk(tablePath, partitionName);
+    }
+
+    public void deletePartitionFromZk(TablePath tablePath, String partitionName) {
```
wuchong: `dropPartition(TablePath tablePath, String partitionName, boolean ignoreIfNotExists)` to align with `dropTable`. All the operations in `MetadataManager` work on zookeeper, so I think don't need to mention `ToZK` in the method name.

The difference between this method and `dropPartition(TableInfo tableInfo, PartitionSpec partitionSpec, boolean ignoreIfNotExists)` is the field type of `partitionSpec`.

fluss-server/src/main/java/com/alibaba/fluss/server/coordinator/MetadataManager.java
```diff
@@ -308,6 +316,110 @@ public long initWriterId() {
                 zookeeperClient::getWriterIdAndIncrement, "Fail to get writer id from zookeeper");
     }
 
+    public Set<String> getPartitions(TablePath tablePath) {
+        return uncheck(
+                () -> zookeeperClient.getPartitions(tablePath),
+                "Fail to get partitions from zookeeper for table " + tablePath);
+    }
+
+    public void addPartition(
```
wuchong: `createPartition`

fluss-server/src/main/java/com/alibaba/fluss/server/coordinator/AutoPartitionManager.java
```diff
@@ -180,44 +195,26 @@ protected void createPartitions(
...
+            long tableId = tableInfo.getTableId();
+            int replicaFactor = tableInfo.getTableConfig().getReplicationFactor();
+            int[] servers = metadataCache.getLiveServerIds();
+            Map<Integer, BucketAssignment> bucketAssignments =
+                    TableAssignmentUtils.generateAssignment(
+                                    tableInfo.getNumBuckets(), replicaFactor, servers)
+                            .getBucketAssignments();
+            PartitionAssignment partitionAssignment =
+                    new PartitionAssignment(tableInfo.getTableId(), bucketAssignments);
+            metadataManager.registerPartitionToZk(
```
wuchong: the partition maybe created by user manually concurrently, should catch `PartitionAlreadyExistsException` should log it, but still add the new partition name to `currentPartitions`.

fluss-server/src/main/java/com/alibaba/fluss/server/coordinator/AutoPartitionManager.java
```diff
@@ -265,21 +262,13 @@ private void dropPartitions(
...
-                        partitionName,
-                        tablePath);
-            } catch (Exception e) {
-                LOG.error(
-                        "Auto partitioning failed to delete partition {} for table [{}]",
-                        partitionName,
-                        tablePath,
-                        e);
-            }
+            metadataManager.deletePartitionFromZk(tablePath, partitionName);
```
wuchong: the partition may be dropped by user manually concurrently, should catch `PartitionNotExistException` and log it. But still need to call `partitionsToExpire.remove()`.

fluss-server/src/main/java/com/alibaba/fluss/server/coordinator/AutoPartitionManager.java
```diff
@@ -99,31 +99,47 @@ public AutoPartitionManager(
...
-        inLock(lock, () -> autoPartitionTables.put(tableId, tableInfo));
+        inLock(
+                lock,
+                () -> {
+                    autoPartitionTables.put(tableId, tableInfo);
+                    Set<String> partitionSet =
+                            partitionsByTable.computeIfAbsent(
+                                    tableInfo.getTableId(), k -> new TreeSet<>());
+                    checkNotNull(partitionSet, "Partition set is null.");
+                    partitionSet.addAll(metadataManager.getPartitions(tableInfo.getTablePath()));
```
wuchong: move the partition fetch out of the lock to reduce the lock time.

fluss-server/src/main/java/com/alibaba/fluss/server/coordinator/CoordinatorEventProcessor.java
```diff
@@ -331,12 +330,12 @@ private void initCoordinatorContext() throws Exception {
...
                             .getAutoPartitionStrategy()
                             .isAutoPartitionEnabled()) {
-                        autoPartitionTables.put(tableInfo.getTableId(), tableInfo);
+                        autoPartitionTables.add(tableInfo);
                     }
                 }
             }
         }
-        autoPartitionManager.initAutoPartitionTables(autoPartitionTables);
+        autoPartitionManager.initPartitionTables(autoPartitionTables);
```
wuchong: `initAutoPartitionTables`

fluss-server/src/main/java/com/alibaba/fluss/server/coordinator/CoordinatorEventProcessor.java
```diff
@@ -533,7 +532,7 @@ private void processDropTable(DropTableEvent dropTableEvent) {
         coordinatorContext.queueTableDeletion(Collections.singleton(dropTableEvent.getTableId()));
         tableManager.onDeleteTable(dropTableEvent.getTableId());
         if (dropTableEvent.isAutoPartitionTable()) {
-            autoPartitionManager.removeAutoPartitionTable(dropTableEvent.getTableId());
+            autoPartitionManager.removePartitionTable(dropTableEvent.getTableId());
```
wuchong: `removeAutoPartitionTable`

fluss-server/src/main/java/com/alibaba/fluss/server/coordinator/AutoPartitionManager.java
```diff
@@ -86,10 +86,10 @@ public class AutoPartitionManager implements AutoCloseable {
     private final Lock lock = new ReentrantLock();
 
     public AutoPartitionManager(
-            MetadataCache metadataCache, ZooKeeperClient zooKeeperClient, Configuration conf) {
+            MetadataCache metadataCache, MetadataManager metadataManager, Configuration conf) {
```
wuchong, line 93: We should also watch partition node creation and deletion events, and add the created partition or remove the dropped partition to the `AutoPartitionManager`, because the partition can be created by external users now. Besides, please add such a test for it.

fluss-common/src/main/java/com/alibaba/fluss/metadata/TablePath.java
```diff
@@ -179,7 +179,9 @@ private static boolean containsInvalidPattern(String identifier) {
                             || (c >= '0' && c <= '9')
                             || (c >= 'A' && c <= 'Z')
                             || c == '_'
-                            || c == '-';
+                            || c == '-'
+                            || c == '$';
```
wuchong: After reconsidering, we cannot allow the `$` character in table names because we use `$lake` as a suffix for lake tables in the Flink catalog. If a table name contains `$lake`, the catalog would be unable to distinguish whether it refers to a regular table or a lake table, leading to potential ambiguity.

We can revert the changes on `detectInvalidName`, but split the `partitionName` using `$`, and validate each partition value using `detectInvalidName`.

fluss-server/src/main/java/com/alibaba/fluss/server/coordinator/CoordinatorEventProcessor.java
swuferhong: Done!

fluss-server/src/main/java/com/alibaba/fluss/server/coordinator/CoordinatorEventProcessor.java
swuferhong: Done!

fluss-server/src/main/java/com/alibaba/fluss/server/coordinator/MetadataManager.java
swuferhong: Done!

fluss-server/src/main/java/com/alibaba/fluss/server/coordinator/AutoPartitionManager.java
swuferhong: Done!

fluss-server/src/main/java/com/alibaba/fluss/server/coordinator/MetadataManager.java
swuferhong: Done!

fluss-server/src/main/java/com/alibaba/fluss/server/coordinator/MetadataManager.java
swuferhong: Done!

fluss-server/src/main/java/com/alibaba/fluss/server/coordinator/AutoPartitionManager.java
swuferhong: Done!

fluss-server/src/main/java/com/alibaba/fluss/server/coordinator/AutoPartitionManager.java
swuferhong: Done!

fluss-server/src/main/java/com/alibaba/fluss/server/coordinator/AutoPartitionManager.java
swuferhong: Done!  add test in AutoPartitionManagerTest.

fluss-common/src/main/java/com/alibaba/fluss/metadata/TablePath.java
swuferhong: Done!

fluss-connectors/fluss-connector-flink/src/main/java/com/alibaba/fluss/connector/flink/catalog/FlinkCatalog.java
swuferhong: Done!

wuchong (approved):
LGTM

wuchong:
Finally, it passed...

PR #354, opened 2025-02-07
[client] Fix the possible data loss error while MemoryLogRecordsArrowBuilder#resetWriterState

swuferhong:
### Purpose

Linked issue: https://github.com/alibaba/fluss/issues/353

When MemoryLogRecordsArrowBuilder try to resetWriterState in case the produceLog request failed and the batch is re-enqueue to send with different write state, we will reset the bytesView to null, this is very dangerous. Because once we set bytesView to null, we won't be able to retrieve the data from arrowWriter while next build, which means data in this batch will be lost. The correct approach is to simply reset the batch header instead of reset the whole batch.

This pr is aims to fix this error by introducing a bool variable `resetBatchHeader` to reset batch header while next build call.

### DIFF
--- a/fluss-common/src/test/java/com/alibaba/fluss/record/MemoryLogRecordsArrowBuilderTest.java
+++ b/fluss-common/src/test/java/com/alibaba/fluss/record/MemoryLogRecordsArrowBuilderTest.java
@@ -326,6 +326,50 @@ public class MemoryLogRecordsArrowBuilderTest {
         }
     }
 
+    @Test
+    void testResetWriterState() throws Exception {
+        int maxSizeInBytes = 1024;
+        ArrowWriter writer =
+                provider.getOrCreateWriter(
+                        1L,
+                        DEFAULT_SCHEMA_ID,
+                        maxSizeInBytes,
+                        DATA1_ROW_TYPE,
+                        ArrowCompressionInfo.NO_COMPRESSION);
+        MemoryLogRecordsArrowBuilder builder =
+                createMemoryLogRecordsArrowBuilder(0, writer, 10, 1024);
+        List<RowKind> rowKinds =
+                DATA1.stream().map(row -> RowKind.APPEND_ONLY).collect(Collectors.toList());
+        List<InternalRow> rows =
+                DATA1.stream()
+                        .map(object -> row(DATA1_ROW_TYPE, object))
+                        .collect(Collectors.toList());
+        List<Object[]> expectedResult = new ArrayList<>();
+        while (!builder.isFull()) {
+            int rndIndex = RandomUtils.nextInt(0, DATA1.size());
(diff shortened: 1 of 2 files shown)


swuferhong:
@wuchong can you take a look at this bug fix?

fluss-common/src/test/java/com/alibaba/fluss/record/MemoryLogRecordsArrowBuilderTest.java
```diff
@@ -326,6 +326,44 @@ void testNoRecordAppend() throws Exception {
...
+        assertThat(builder.isFull()).isTrue();
+        builder.setWriterState(1L, 0);
+        builder.close();
+        assertThat(builder.isClosed()).isTrue();
+        MemoryLogRecords records = MemoryLogRecords.pointToBytesView(builder.build());
+        assertLogRecordsEquals(DATA1_ROW_TYPE, records, expectedResult);
+
+        // test reset writer state and build (This situation will happen when the produceLog request
+        // failed and the batch is re-enqueue to send with different write state).
+        builder.resetWriterState(1L, 1);
```
wuchong, line 365: assert the new writer id and sequence number for the generated batch as well!

fluss-common/src/test/java/com/alibaba/fluss/record/MemoryLogRecordsArrowBuilderTest.java
```diff
@@ -326,6 +326,44 @@ void testNoRecordAppend() throws Exception {
...
+                        .map(object -> row(DATA1_ROW_TYPE, object))
+                        .collect(Collectors.toList());
+        List<Object[]> expectedResult = new ArrayList<>();
+        while (!builder.isFull()) {
+            int rndIndex = RandomUtils.nextInt(0, DATA1.size());
+            builder.append(rowKinds.get(rndIndex), rows.get(rndIndex));
+            expectedResult.add(DATA1.get(rndIndex));
+        }
+        assertThat(builder.isFull()).isTrue();
+        builder.setWriterState(1L, 0);
```
wuchong, line 354: assert the writer id and sequence number!

fluss-common/src/test/java/com/alibaba/fluss/record/MemoryLogRecordsArrowBuilderTest.java
swuferhong: Done!

fluss-common/src/test/java/com/alibaba/fluss/record/MemoryLogRecordsArrowBuilderTest.java
swuferhong: Done!

PR #357, opened 2025-02-08
[kv] The kv pre-write buffer need to truncate while receive duplicated batches

swuferhong:
### Purpose

Linked issue: https://github.com/alibaba/fluss/issues/325

Currently, the `kv pre-write buffer` writes data into the buffer before submitting the CDC log. However, for duplicate KV batches, the duplicated CDC logs that we submit will not be written to disk by the logTablet, and the logEndOffset will not be increase. This results in `the MSN` in the `pre-write buffer` being greater than the `logEndOffset`, leading to an error. The purpose of this PR is to fix this issue. If we detect a duplicate KV batch, the data that has already been written into the `pre-write buffer` needs to be truncated.

Also, I add some metrics to avoid print so much logs when duplicated or error happend in `pre-write buffer`.

### DIFF
--- a/fluss-server/src/main/java/com/alibaba/fluss/server/kv/prewrite/KvPreWriteBuffer.java
+++ b/fluss-server/src/main/java/com/alibaba/fluss/server/kv/prewrite/KvPreWriteBuffer.java
@@ -18,11 +18,12 @@ package com.alibaba.fluss.server.kv.prewrite;
 
 import com.alibaba.fluss.annotation.VisibleForTesting;
 import com.alibaba.fluss.memory.MemorySegment;
+import com.alibaba.fluss.metrics.Counter;
+import com.alibaba.fluss.metrics.DescriptiveStatisticsHistogram;
+import com.alibaba.fluss.metrics.Histogram;
+import com.alibaba.fluss.metrics.SimpleCounter;
 import com.alibaba.fluss.server.kv.KvBatchWriter;
 import com.alibaba.fluss.utils.MurmurHashUtils;
-
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
 
 import javax.annotation.Nullable;
 import javax.annotation.concurrent.NotThreadSafe;
@@ -85,9 +86,6 @@ import static com.alibaba.fluss.utils.UnsafeUtils.BYTE_ARRAY_BASE_OFFSET;
  */
 @NotThreadSafe
 public class KvPreWriteBuffer implements AutoCloseable {
-
-    private static final Logger LOG = LoggerFactory.getLogger(KvPreWriteBuffer.class);
-
     private final KvBatchWriter kvBatchWriter;
 
     // a mapping from the key to the kv-entry
@@ -96,11 +94,23 @@ public class KvPreWriteBuffer implements AutoCloseable {
     // a linked list for all kv entries
     private final LinkedList<KvEntry> allKvEntries = new LinkedList<>();
 
+    // metrics related.
+    private final Counter flushCount;
+    private final Histogram flushLatencyHistogram;
+    private final Counter truncateAsDuplicatedCount;
+    private final Counter truncateAsErrorCount;
+
     // the max LSN in the buffer
     private long maxLogSequenceNumber = -1;
 
     public KvPreWriteBuffer(KvBatchWriter kvBatchWriter) {
         this.kvBatchWriter = kvBatchWriter;
+
+        flushCount = new SimpleCounter();
+        // consider won't flush frequently, we set a small window size
+        flushLatencyHistogram = new DescriptiveStatisticsHistogram(5);
+        truncateAsDuplicatedCount = new SimpleCounter();
+        truncateAsErrorCount = new SimpleCounter();
     }
 
     /**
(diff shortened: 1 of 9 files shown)


swuferhong:
@wucong @yuxia could you take a look at this bug fix, thx.

fluss-server/src/main/java/com/alibaba/fluss/server/kv/prewrite/KvPreWriteBuffer.java
```diff
@@ -426,4 +462,10 @@ public String toString() {
             return value == null ? "null" : "[" + Base64.getEncoder().encodeToString(value) + "]";
         }
     }
+
+    /** The reason why we truncate the kv pre-write buffer. */
+    public enum KvPreWriteBufferTruncateReason {
```
wuchong: too long, simplify it to `TruncateReason`.

fluss-server/src/test/java/com/alibaba/fluss/server/kv/prewrite/KvPreWriteBufferTest.java
```diff
@@ -165,7 +165,7 @@ void testTruncate() {
         bufferPut(buffer, "key1", "value1-1", elementCount++);
         assertThat(getValue(buffer, "key1")).isEqualTo("value1-1");
         assertThat(buffer.getMaxLSN()).isEqualTo(elementCount - 1);
-        buffer.truncateTo(5, "test");
+        buffer.truncateTo(5, KvPreWriteBuffer.KvPreWriteBufferTruncateReason.ERROR);
```
wuchong: nit: use on-demand or static import for the enum.

website/docs/maintenance/monitor-metrics.md
```diff
@@ -629,11 +629,31 @@ Some metrics might not be exposed when using other JVM implementations (e.g. IBM
...
     </tr>
     <tr>
-      <td rowspan="1">table_bucket_kv_snapshot</td>
+      <td rowspan="5">table_bucket_kv_snapshot</td>
       <td>latestSnapshotSize</td>
       <td>The latest kv snapshot size in bytes for this table bucket.</td>
       <td>Gauge</td>
     </tr>
+     <tr>
+      <td>preWriteBufferFlushPerSecond</td>
```
wuchong, line 633: They are belong to `table_bucket_kv`, not `table_bucket_kv_snapshot`.

fluss-server/src/main/java/com/alibaba/fluss/server/kv/prewrite/KvPreWriteBuffer.java
swuferhong: Done!

fluss-server/src/test/java/com/alibaba/fluss/server/kv/prewrite/KvPreWriteBufferTest.java
swuferhong: Done!

website/docs/maintenance/monitor-metrics.md
swuferhong: Done!

swuferhong:
@wuchong comments addressed.

fluss-server/src/main/java/com/alibaba/fluss/server/kv/prewrite/KvPreWriteBuffer.java
```diff
@@ -221,7 +235,12 @@ public void flush(long exclusiveUpToLogSequenceNumber) throws IOException {
             kvEntryMap.remove(entry.getKey(), entry);
         }
         // flush to underlying kv tablet
-        kvBatchWriter.flush();
+        if (flushedCount > 0) {
+            long start = System.nanoTime();
+            kvBatchWriter.flush();
+            flushCount.inc();
+            flushLatencyHistogram.update((System.nanoTime() - start) / 1_000_000);
```
luoyuxia, line 242: `kvBatchWriter` will hold a buffer, if buffer is full after `put(entry.getKey().key, value.value)` or `kvBatchWriter.delete(entry.getKey().key)`, it'll also flush to rocksdb..

If we calculate the cost of  `kvBatchWriter.flush()`, it'll means the latency of flush to rocksdb instead flushing kvPrewriterBuffer

wuchong (approved):
LGTM

fluss-server/src/main/java/com/alibaba/fluss/server/kv/prewrite/KvPreWriteBuffer.java
```diff
@@ -164,11 +174,12 @@ private void update(Key key, Value value, long lsn) {
...
-    public void truncateTo(long targetLogSequenceNumber, String truncateReason) {
-        LOG.info(
-                "Truncate the kv pre-write buffer to {} because we encounter error while building cdc logs: {}",
-                targetLogSequenceNumber,
-                truncateReason);
+    public void truncateTo(long targetLogSequenceNumber, TruncateReason truncateReason) {
+        if (truncateReason == TruncateReason.DUPLICATED) {
+            truncateAsDuplicatedCount.inc();
+        } else {
+            truncateAsErrorCount.inc();
```
luoyuxia, line 181: If the reason is not duplicated, which means is not unexpcted.. Can we print the error message? Otherwise, we have no way to know what exception happens in server.

website/docs/maintenance/monitor-metrics.md
```diff
@@ -628,6 +628,27 @@ Some metrics might not be exposed when using other JVM implementations (e.g. IBM
       <td>The number of bytes written per second to this table.</td>
       <td>Gauge</td>
     </tr>
+    <tr>
+      <td rowspan="4">table_bucket_kv</td>
+      <td>preWriteBufferFlushPerSecond</td>
+      <td>The kv pre-write buffer flush count per second.</td>
```
luoyuxia, line 634: nit:
```suggestion
      <td>Th count of the flush rows in kv pre-write buffer per second.</td>
```
?
The current description always make me feel like it's kv pre-wirte buffer flush count, not the count of flushed rows in kv pre-wirte buffer

website/docs/maintenance/monitor-metrics.md
```diff
@@ -628,6 +628,27 @@ Some metrics might not be exposed when using other JVM implementations (e.g. IBM
...
+      <td>Meter</td>
+    </tr>
+     <tr>
+      <td>preWriteBufferFlushLatencyMs</td>
+      <td>The kv pre-write buffer latency in ms.</td>
```
luoyuxia, line 639:
```suggestion
      <td>The kv pre-write buffer flush latency in ms.</td>
```

wuchong:
@swuferhong you can create another pull request to address the new comments.

PR #360, opened 2025-02-08
[client] Use TableInfo instead of TableDescriptor for created table

### LINKED ISSUE #363: Use TableInfo instead of TableDescriptor for created table
### Search before asking

### Motivation

Currently, many instances rely on `TableDescriptor` to retrieve table metadata, such as bucket keys, bucket numbers, and primary keys. However, because `TableDescriptor` is primarily designed for table creation, extracting metadata from it can be cumbersome, as numerous fields are optional. This often leads to lengthy method chains like `tableInfo.getTableDescriptor().getTableDistribution().get().getBucketCount().get()` just to obtain a bucket number. This approach not only complicates the code but also generates multiple IDE warnings due to forced retrieval from Optionals. It's important to note that once a table is created, it will always have a defined bucket number.

### Solution

This issue aims to streamline the process by exclusively using `TableInfo` to access the metadata of a created table. We will introduce several useful getters to significantly simplify the code, making it more efficient and readable.

### Anything else?

_No response_

### Willingness to contribute

wuchong:
### Purpose

Linked issue: close #363


fluss-server/src/main/java/com/alibaba/fluss/server/zk/data/TableRegistration.java
```diff
@@ -69,24 +80,39 @@ public TableDescriptor toTableDescriptor(Schema schema) {
                 TableDescriptor.builder()
```
luoyuxia: The method `toTableDescriptor` can be removed?

luoyuxia (approved):
Thanks for the pr.. Looks much better to me after refactoring...Only one minor comment..

fluss-server/src/main/java/com/alibaba/fluss/server/zk/data/TableRegistration.java
wuchong: good catch!

PR #362, opened 2025-02-08
all tables in lake storage configured cluster should use lake's bucke…

### LINKED ISSUE #343: [Feature] Tables in lake storeage confgiured cluster should use lake bucket assign strategy  to  support enable lake storage for table dynamically
### Search before asking

### Motivation

Currently, the table must be created with lake storage enabled to use lake storage. It's not convenient since in many cases users want to enable it  dynamically instead of enabling it while creating..We'd like to support enable lake storage for table dynamically.

### Solution

The reason why we design it currently is that the bucket assignment algorithm between Fluss and Lake format is different.. Some rows has been written into bucket according to Fluss  bucket assignment, then when enable it with lake storage, it'll fall into the bucket in lake according to Lake's bucket assignment which brings the inconsistency of bucket assignment of the same row between Fluss and Lake..
To solve it, we just allow enabling lake storage while creating the table to enforce all rows to use the lake's bucket assignment currently.
But if want to support  enable lake storage for table dynamically, we must:

- All rows of all tables should use the lake's bucket assignment whether it enables lake storage or not.

### Anything else?

_No response_

### Willingness to contribute

luoyuxia:
…t assigner

### Purpose

Linked issue: close #343

Make all tables in lake storage configured cluster use lake's bucket assign strategy..

- If a table is created in lakestorage configured cluster, set `table.datalake.format` to the corresponding datalake 
- If table is with  `table.datalake.format`, use lake's bucket assinger to assign bucket for rows
- Split current bucket assinger to `StaticBucketAssigner` and `DynamicBucketAssigner`.. For tables with bucket keys,
use `StaticBucketAssigner` to caclcuate bucket id during convert row into `WriteRecord`.

### Tests
LakeTableManagerITCase to verify  `table.datalake.format`  will be set in  lakestorage configured cluster
FlussLakeTableITCase to verfy write/lookup will use lake's bucket assinger t


fluss-client/src/main/java/com/alibaba/fluss/client/write/WriterClient.java
```diff
@@ -156,24 +154,37 @@ private void doSend(WriteRecord record, WriteCallback callback) {
...
+                            dynamicBucketAssigner != null
+                                    && dynamicBucketAssigner.abortIfBatchFull());
 
             if (result.abortRecordForNewBatch) {
                 int prevBucketId = bucketId;
-                bucketAssigner.onNewBatch(cluster, prevBucketId);
-                bucketId =
-                        bucketAssigner.assignBucket(
-                                record.getBucketKey(), record.getRow(), cluster);
+                if (dynamicBucketAssigner != null) {
```
wuchong: I still prefer to have only one bucket assigner interface, so we don't have different branches to do the bucketing here, and do bucketing twice (once in UpsertWriter, once in here). But I like the static/dynamic abstraction to make the implementation simpler. Maybe we can have a single root bucket assigner interface, but have 2 abstract classes to make concrete assigner easy to implement. 

```java
// root interface, WriterClient holds an instance of BucketAssigner for every table.
interface BucketAssigner {
  boolean abortIfBatchFull();
  void onNewBatch(Cluster cluster, int prevBucketId);
  int assignBucket(@Nullable byte[] bucketKey, Cluster cluster);
}

abstract class DynamicBucketAssigner implements BucketAssigner {

    @Override
    public int assignBucket(@Nullable byte[] bucketKey, Cluster cluster) {
        return assignBucket(cluster);
    }

    public abstract int assignBucket(Cluster cluster);
}

public abstract class StaticBucketAssigner implements BucketAssigner {

    abstract int assignBucket(byte[] bucketKey);

    @Override
    public int assignBucket(@Nullable byte[] bucketKey, Cluster cluster) {
        checkNotNull(bucketKey);
        return assignBucket(bucketKey);
    }

    @Override
    public boolean abortIfBatchFull() {
        return false;
    }

    @Override
    public void onNewBatch(Cluster cluster, int prevBucketId) {
        // do nothing
    }
}
```

fluss-client/src/main/java/com/alibaba/fluss/client/table/writer/UpsertWriterImpl.java
```diff
@@ -126,15 +143,19 @@ public CompletableFuture<UpsertResult> upsert(InternalRow row) {
...
-                        targetColumns))
+                        getPhysicalPath(row), WriteKind.DELETE, key, bucketId, null, targetColumns))
                 .thenApply(ignored -> DELETE_SUCCESS);
     }
+
+    private int getBucketId(byte[] keyBytes, InternalRow row) {
+        if (isDefaultBucketKey && bucketAssigner instanceof HashBucketAssigner) {
+            // is is default bucket key, and assigner is HashBucketAssAssigner,
+            // we can reuse the primary key bytes to calculate bucket id
+            return ((HashBucketAssigner) bucketAssigner).assignBucket(keyBytes);
```
wuchong: This is a super hack to do a hard cast, it means we didn't have a good abstraction here. 

As suggested above, if all the encoded (primary/bucket) key use the datalake KeyEncoder, I think the parameter of `BucketAssigner` can be changed to `byte[] key` from the `InternalRow`.

fluss-common/src/main/java/com/alibaba/fluss/config/TableConfig.java
```diff
@@ -75,6 +77,14 @@ public boolean isDataLakeEnabled() {
         return config.get(ConfigOptions.TABLE_DATALAKE_ENABLED);
     }
 
+    /**
+     * Return the data lake type of the table. It'll be the lake storage configured in Fluss whiling
+     * creating the table. Return null if no lake storage configured while creating.
+     */
+    public @Nullable String getDataLakeType() {
```
wuchong: `getDataLakeFormat()`

wuchong (changes requested):
Some high-level design suggestions:

1. The cluster datalake configuration should align with table configuration, how about:

```
datalake.format = paimon/iceberg
datalake.paimon.catalog = hive
datalake.paimon.metastore = thrift://localhost:9083
datalake.paimon.warehouse = hdfs://localhost:9000/user/hive/warehouse
```

2. Besides, the `table.datalake.format`, we can also add the catalog and metastore information into table config, so that the table can be self-contained, and connectors don't need to retrieve `describeLakeStorage` (`describeLakeStorage` can be removed then). If there is any auth/securtiy tokens to retrieve, we can add another RPC similar to `getFileSystemSecurityToken`, but that should be optional.

```
table.datalake.enabled = true/false
table.datalake.format = paimon/iceberg

// logical properties
table.datalake.paimon.catalog = hive
table.datalake.paimon.metastore = thrift://localhost:9083
table.datalake.paimon.warehouse = hdfs://localhost:9000/user/hive/warehouse
```

Considering the metastore location may be changed over time, it would be a logical property that is added during `getTableInfo` instead of a physical property stored in zookeeper. And thus metastore properties shouldn't be set or altered.

3. It's confusing that the key encoding is different for bucketing and storing. It's meaningless to keep our own key encoding when it is a datalake-enabled table. We also discussed that bucket-rescale can be supported even if we use Paimon/Iceberg key encoding/bucketing. So we can introduce two separate interfaces for encoding and bucketing. 

```java
interface KeyEncoder {
    byte[] encodeKey(InternalRow row);
}

interface BucketAssigner {
  int assignBucket(@Nullable byte[] bucketKey, Cluster cluster);
}
```

4. We should add tests to verify the encoding and bucketing is the same with Paimon

luoyuxia:
Comments addressed... Let's introduce Paimon's bucket & key encoding in #408 and remove describeLakeStorage in #411

fluss-lakehouse/fluss-lakehouse-paimon/src/main/java/com/alibaba/fluss/lakehouse/paimon/FlussLakehousePaimon.java
```diff
@@ -124,4 +124,14 @@ public static Map<String, String> extractConfigStartWith(
         }
         return extractedConfig;
     }
+
+    private static Map<String, String> normalizeToPaimonConfigs(
```
wuchong: Use `datalake.paimon.metastore` to replace `datalake.paimon.catalog`.

fluss-server/src/main/java/com/alibaba/fluss/server/utils/LakeStorageUtils.java
```diff
@@ -26,38 +27,47 @@
...
+                datalakeConfig.put(
+                        configKey.substring(DATALAKE_PAIMON_PREFIX.length()), configValue);
             }
         }
-        return new LakeStorageInfo(datalakeStorage, catalogConfig);
+        return new LakeStorageInfo(
+                datalakeFormat.toString(), normalizeToPaimonConfigs(datalakeConfig));
+    }
+
+    private static Map<String, String> normalizeToPaimonConfigs(
```
wuchong: Use `datalake.paimon.metastore` to replace `datalake.paimon.catalog`

fluss-server/src/main/java/com/alibaba/fluss/server/coordinator/CoordinatorService.java
```diff
@@ -186,6 +194,14 @@ private TableDescriptor applySystemDefaults(TableDescriptor tableDescriptor) {
             newDescriptor = newDescriptor.withReplicationFactor(defaultReplicationFactor);
         }
 
+        // if lake storage is not null, we need to add the datalake type
+        // to the property of the table
+        if (dataLakeFormat != null) {
```
wuchong: we should throw exception if the table is datalake enabled, but cluster not.

fluss-client/src/main/java/com/alibaba/fluss/client/lookup/PrefixKeyLookuper.java
```diff
@@ -80,15 +86,19 @@ public PrefixKeyLookuper(
...
+            this.bucketKeyEncoder =
+                    LakeKeyEncoderFactory.createKeyEncoder(
+                            dataLakeFormat, lookupRowType, tableInfo.getBucketKeys());
+            this.lakeBucketAssigner =
+                    LakeBucketAssignerFactory.createLakeBucketAssigner(dataLakeFormat, numBuckets);
         } else {
-            this.lakeTableBucketAssigner = null;
+            this.bucketKeyEncoder =
+                    CompactedKeyEncoder.createKeyEncoder(lookupRowType, tableInfo.getBucketKeys());
+            this.lakeBucketAssigner = null;
```
wuchong: Constructing the `bucketKeyEncoder` is still very complex. However, we can unify the default and lake assigners into one util method.

fluss-client/src/main/java/com/alibaba/fluss/client/utils/ClientUtils.java
```diff
@@ -136,16 +136,13 @@ public static Long getPartitionId(
...
-            return HashBucketAssigner.bucketForRowKey(keyBytes, numBuckets);
+            byte[] bucketKeyBytes,
+            @Nullable LakeBucketAssigner lakeBucketAssigner,
+            int numBuckets) {
+        if (lakeBucketAssigner == null) {
+            return HashBucketAssigner.bucketForRowKey(bucketKeyBytes, numBuckets);
         } else {
-            return lakeTableBucketAssigner.assignBucket(
-                    keyBytes, key, metadataUpdater.getCluster());
+            return lakeBucketAssigner.assignBucket(bucketKeyBytes);
```
wuchong: It is still not unified and complex to assign bucket. We can introduce a `BucketingFunction#bucketing(byte[] key, int numBuckets)` to unify the bucket assigned instead of using a util method and passing optional lake bucket assigner. We should hold only one not-null bucket assigner, that will be easier to use.

wuchong:
Please review the changes I just pushed @luoyuxia .

luoyuxia (commented):
@wuchong Thanks for the commit to unify encoding and bucket assigning.. Looks much better after unifying. It make simpler to use.. LGTM

wuchong (approved):

PR #367, opened 2025-02-10
[feat] [issue-345] feat: remove 'table.' and 'client.' validation for flink connector fa…

### LINKED ISSUE #345: [Feature] Flink connector factory shouldn't validate "table." and "client." options
### Search before asking

### Motivation

Currently, in the `FlinkTableFactory`, we validate the options table to ensure that it includes all entries from `FlinkTableFactory#requiredOptions` and contains no unexpected entries beyond those specified in `FlinkTableFactory#optionalOptions`. This validation helps catch mistakes when users enter incorrect option keys. However, it also prevents the connector from being compatible with newer versions of the Fluss table, which might include options unknown to the connector.

For instance, consider a scenario where a user is working with a connector version v0.5.0, while the Fluss cluster has been upgraded to v0.6.0. In version v0.6.0, the cluster may automatically add options like `table.lake.format` during table creation. Since this option is unrecognized by the older v0.5 connector, trying to create a Fluss table on a v0.6 cluster using this connector will include the unrecognized `table.lake.format` option, triggering an exception when the table is read by this connector. This situation forces users to upgrade to a v0.6 connector to interact with a v0.6 cluster, which is undesirable as it disrupts the smooth upgrade process of the Fluss cluster without affecting all clients.

To address this, we should exclude options prefixed with `table.*` and `client.*` from `FlinkTableFactory#optionalOptions`. Instead, we can use `FactoryHelper.validateExcept("table.", "client.")` for option validation, allowing flexibility for unknown options while maintaining compatibility.

https://github.com/alibaba/fluss/blob/d13758d12c030f18bd4faec8234b56cdc02ac331/fluss-connectors/fluss-connector-flink/src/main/java/com/alibaba/fluss/connector/flink/catalog/FlinkTableFactory.java#L74

https://github.com/alibaba/fluss/blob/d13758d12c030f18bd4faec8234b56cdc02ac331/fluss-connectors/fluss-connector-flink/src/main/java/com/alibaba/fluss/connector/flink/catalog/FlinkTableFactory.java#L138

### Solution

_No response_

### Anything else?

_No response_

### Willingness to contribute

gyang94:
### Purpose

Linked issue: close #345 

Remove 'table.' and 'client.' options validation in flink connector factory.

### Tests

N/A

### API and Format

N/A

### Documentation

N/A

### DIFF
--- a/fluss-connectors/fluss-connector-flink/src/main/java/com/alibaba/fluss/connector/flink/catalog/FlinkTableFactory.java
+++ b/fluss-connectors/fluss-connector-flink/src/main/java/com/alibaba/fluss/connector/flink/catalog/FlinkTableFactory.java
@@ -24,6 +24,7 @@ import com.alibaba.fluss.connector.flink.sink.FlinkTableSink;
 import com.alibaba.fluss.connector.flink.source.FlinkTableSource;
 import com.alibaba.fluss.connector.flink.utils.FlinkConnectorOptionsUtils;
 import com.alibaba.fluss.metadata.TablePath;
+
 import org.apache.flink.api.common.RuntimeExecutionMode;
 import org.apache.flink.configuration.ConfigOption;
 import org.apache.flink.configuration.CoreOptions;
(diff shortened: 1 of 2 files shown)


wuchong (commented):
Please add tests. You can add tests in `FlinkCatalogITCase` by creating tables with additional unknown `table.` and `client.` prefix options, and verify that these tables can be created and used successfully. Instead of running the query, use `tEnv.explainSql(..)` to verify that the execution plan is generated successfully. This approach reduces the test cost and is sufficient for the purpose of this test.

wuchong:
Btw, please run `mvn spotless:apply` to fix the checkstyle. You can configure the Save Action and spotless plugin in your IDEA to automatically fixing them when developing.

gyang94:
Hi @wuchong , I added tests case and fixed the checkstyle. Please check.

PR #370, opened 2025-02-10
[flink] Change the name of default value from initial to full for scan.startup.mode option

### LINKED ISSUE #346: [Feature] Change default value name from "initial" to "full" for scan.startup.mode option
### Search before asking

### Motivation

Currently, the default value of  `scan.startup.mode` is `initial`, which means "performs an **initial** snapshot first, and then continue read latest logs". However, the `initial` name may confuse users it only performs the "initial snapshot". 

### Solution

A better name can be "full" that reads the full content/data from the table. For streaming mode, it produces the latest snapshot on the table (no matter Log Table or Primary Key Table) upon first startup, and continue to read the latest logs. For batch mode, it just produce the latest snapshot but does not read new logs. The "full" word reflects it reads all data in batch mode, and read all data for current and future in streaming mode. 

[Paimon also uses "full"](https://paimon.apache.org/docs/master/maintenance/configurations/) for the same semantics. Paimon deprecated "full" and uses `compacted-full`, `latest-full`, `from-snapshot-full` instead. However, Fluss is not a data lake format (no compaction, no snapshots), and only has `latest-full`. Therefore, we can use the simplified `full` as the scan mode. 

### Anything else?

_No response_

### Willingness to contribute

SteNicholas:
### Purpose

Linked issue: close #346

The default value of `scan.startup.mode` is `initial` at present, which means `perform an initial snapshot first, and then continue read latest logs`. The `initial` name is confusing for users that it only performs the "initial snapshot". `full` is a better name that reads the full content/data from the table. For streaming mode, it produces the latest snapshot on the table (no matter Log Table or Primary Key Table) upon first startup, and continue to read the latest logs. For batch mode, it just produces the latest snapshot but does not read new logs. The `full` word reflects it reads all data in batch mode, and read all data for current and future in streaming mode. Paimon also uses [`full`](https://paimon.apache.org/docs/master/maintenance/configurations/) for the same semantics. Paimon deprecated `full` and uses `compacted-full`, `latest-full`, `from-snapshot-full` instead. However, Fluss is not a data lake format (no compaction, no snapshots), and only has latest-full. Therefore, we can use the simplified `full` as the scan mode.

### Tests

- `FlinkTableSourceITCase#testReadLogTableWithDifferentScanStartupMode`
- `FlinkTableSourceITCase#testReadKvTableWithScanStartupModeEqualsFull`

### API and Format

The name of default value for `scan.startup.mode` option changes from `initial` to `full`.

### Documentation

The name of default value for `scan.startup.mode` option changes from `initial` to `full` in `reads.md`.

### DIFF
--- a/fluss-connectors/fluss-connector-flink/src/main/java/com/alibaba/fluss/connector/flink/FlinkConnectorOptions.java
+++ b/fluss-connectors/fluss-connector-flink/src/main/java/com/alibaba/fluss/connector/flink/FlinkConnectorOptions.java
@@ -76,9 +76,11 @@ public class FlinkConnectorOptions {
     public static final ConfigOption<ScanStartupMode> SCAN_STARTUP_MODE =
             ConfigOptions.key("scan.startup.mode")
                     .enumType(ScanStartupMode.class)
-                    .defaultValue(ScanStartupMode.INITIAL)
+                    .defaultValue(ScanStartupMode.FULL)
                     .withDescription(
-                            "Optional startup mode for Fluss source. Default is 'initial'.");
+                            String.format(
+                                    "Optional startup mode for Fluss source. Default is '%s'.",
+                                    ScanStartupMode.FULL.value));
 
     public static final ConfigOption<String> SCAN_STARTUP_TIMESTAMP =
             ConfigOptions.key("scan.startup.timestamp")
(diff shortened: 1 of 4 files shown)


wuchong (commented):
LGTM>

wuchong (approved):

PR #382, opened 2025-02-11
[server] Support multi-field partition table

swuferhong:
### Purpose

Linked issue: https://github.com/alibaba/fluss/issues/195

This pr is aims to support multi-field partition table.


wuchong:
@swuferhong do we need to close this as it is very old, and we have separate the issue into sub issues.

swuferhong:
> @swuferhong do we need to close this as it is very old, and we have separate the issue into sub issues?

Sure, I have closed it.

PR #386, opened 2025-02-12
[server] Add the leader of bucket to metadata cache

luoyuxia:
### Purpose

Linked issue: close #386
Currnetly, to know the leader of a TableBucket, we always need to get from zk..When the buckets increase, it will cost too many time which will cause the job fail to startup just like the issue describes

In this pr, we only add the leader of a TableBucket to metadata for simplicity, maybe in the future, we can add more to metadata cache.

1. In `ServerMetadataCache`, cache the  mapping from `tableBucket ` to `leader`
2. Modify `UpdateMetadataRequest` in rpc proto to add `PbTableBucketMetadata`, `deleted_table_id`, `deleted_partition_id.
3. When the leader of any bucket changes, coordinator sends `UpdateMetadataRequest` with the updated leader in `PbTableBucketMetadata` to tablet servers, then tablet servers can update their own cache.
4. When  table/partition is dropped, coordinator sends `UpdateMetadataRequest` with  `deleted_table_id`/`deleted_partition_id` to tablet servers, then tablet servers can remove these tables from their own cache

### Tests
`ServerMetadataCacheImplTest` to verify the `ServerMetadataCache` works as expected.
`MetadataUpdateITCase` to verify when table created, bucket leader changed, server down/up, the cache is as expected.

### API and Format

N/A

### Documentation

N/A


luoyuxia:
@wuchong Add the leader of bucket to metadata cache to avoid get table metadata cost too much time.. Pr ready

fluss-connectors/fluss-connector-flink/src/test/java/com/alibaba/fluss/connector/flink/catalog/FlinkCatalogTest.java
```diff
@@ -130,7 +129,7 @@ void beforeEach() throws Exception {
         } catch (CatalogException e) {
             // the auto partitioned manager may create the db zk node
             // in an another thread, so if exception is NodeExistsException, just ignore
-            if (!ExceptionUtils.findThrowable(e, KeeperException.NodeExistsException.class)
+            if (!ExceptionUtils.findThrowableWithMessage(e, "KeeperException$NodeExistsException")
```
wuchong, line 132: Could you extract this into a separate hotfix pull request? This seems quite unstable in these days. https://github.com/alibaba/fluss/actions/runs/13626469089/job/38084662575

fluss-rpc/src/main/proto/FlussApi.proto
```diff
@@ -160,9 +160,10 @@ message MetadataResponse {
 
 // update metadata request and response, only send between server.
 message UpdateMetadataRequest {
-  optional PbServerNode coordinator_server = 1;
-  repeated PbServerNode tablet_servers = 2;
-  repeated PbTableMetadata table_metadata = 3;
+  optional PbClusterServerMetadata cluster_server_metadata = 1;
+  repeated PbTableBucketMetadata bucket_metadata = 2;
```
wuchong, line 164: We should keep compatible since this version. Could you update it?

wuchong (commented):
Could you rebase and resolve the conflicts?

PR #392, opened 2025-02-12
[hotfix] Revert build a Flink CatalogTable use CatalogTable#newBuilder() to backward compatible with lower Flink version

swuferhong:
### Purpose

Linked issue: close #xxx

In pr https://github.com/alibaba/fluss/pull/360, we change the way of building Flink `CatalogTable`, using `CatalogTable#newBuilder()` instead of `CatalogTable#of()` in `FlinkConversions#toFlinkTable()`. However, in lower version of Flink, like 'flink-1.17', `CatalogTable#newBuilder()` has not been introduced yet, so in order to maintain better compatibility with lower Flink versions, this PR will revert this change.


swuferhong:
Hi, @wuchong  could you take a lookup at this hotfix, thx.

wuchong (approved):
LGTM

PR #393, opened 2025-02-12
[hotfix] Modify getTableOrElseThrow method, add the table's database …

ljw-hit:
…name and table name to the exception message.

### Purpose

Linked issue: close #xxx

### DIFF
--- a/fluss-common/src/main/java/com/alibaba/fluss/cluster/Cluster.java
+++ b/fluss-common/src/main/java/com/alibaba/fluss/cluster/Cluster.java
@@ -244,9 +244,8 @@ public final class Cluster {
                         () ->
                                 new IllegalArgumentException(
                                         String.format(
-                                                "table: %s.%s not found in cluster",
-                                                tablePath.getDatabaseName(),
-                                                tablePath.getTableName())));
+                                                "table: %s not found in cluster",
+                                                tablePath.toString())));
     }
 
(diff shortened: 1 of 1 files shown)


fluss-common/src/main/java/com/alibaba/fluss/cluster/Cluster.java
```diff
@@ -240,7 +240,13 @@ public boolean contains(PhysicalTablePath physicalTablePath) {
...
     public TableInfo getTableOrElseThrow(TablePath tablePath) {
         return getTable(tablePath)
-                .orElseThrow(() -> new IllegalArgumentException("table not found in cluster"));
+                .orElseThrow(
+                        () ->
+                                new IllegalArgumentException(
+                                        String.format(
+                                                "table: %s.%s not found in cluster",
+                                                tablePath.getDatabaseName(),
+                                                tablePath.getTableName())));
```
swuferhong: There is no need to extract and print the database and table, as the TablePath has toString() method.

ljw-hit: resolved

swuferhong: Remove unnecessary 'toString()' while print tablePath, as String.format() can auto call Object.toString().

ljw-hit: resolved

PR #396, opened 2025-02-12
[hotfix] Don't initialize LakeTableBucketAssigner to avoid load unnecessary dependencies if dataLake disabled for target table

swuferhong:
### Purpose

Linked issue: close #xxx

Currently, In `PrimaryKeyLookuper` and `PrefixKeyLookuper`, we will always initialize the `LakeTableBucketAssigner`, even if the table doesn't enable dataLake. Since the current implementation of `LakeTableBucketAssigner` will always load the Paimon dependency, this can lead to situations where the dependency is not found after the actual packaging. 

Therefore, in this PR, we have modified it to only initialize the `LakeTableBucketAssigner` when dataLake is enabled.


swuferhong:
@luoyuxia  Could you take a look at this hotfix? thx

PR #407, opened 2025-02-17
[rpc] Support netty metrics

### LINKED ISSUE #135: [Feature] Add netty related metrics
<div type='discussions-op-text'>

<sup>Originally posted by **swuferhong** December  2, 2024</sup>
### Search before asking

### Motivation

Fluss uses `Netty` as the underlying network communication component, but it does not expose Netty-related metrics, such as the amount of network memory allocation and the number of requests etc. This makes it less convenient for optimizing network performance.

### Solution

_No response_

### Anything else?

_No response_

### Willingness to contribute

- [ ] I'm willing to submit a PR!</div>

SteNicholas:
### Purpose

Linked issue: close #135

Support netty metrics including:

- `netty`
  - `usedDirectMemory`
  - `numDirectArenas`
  - `numAllocationsPerSecond`
  - `numHugeAllocationsPerSecond`

### Tests

`NettyMetricsTest`

### API and Format

No.

### Documentation

`monitor-metrics.md` adds the `usedDirectMemory`, `numDirectArenas`, `numAllocationsPerSecond` and `numHugeAllocationsPerSecond` metrics.


SteNicholas:
Ping @wuchong.

fluss-rpc/src/main/java/com/alibaba/fluss/rpc/netty/NettyMemoryMetrics.java
```diff
@@ -0,0 +1,169 @@
...
+                        "numActiveNormalAllocations",
+                        "numActiveHugeAllocations",
+                        "numActiveBytes"));
+    }
+
+    public NettyMemoryMetrics(
+            MetricGroup metricGroup,
+            PooledByteBufAllocator pooledAllocator,
+            boolean verboseMetricsEnabled) {
+        this.metricGroup = metricGroup.addGroup(NETTY_MEMORY_METRIC_GROUP);
```
wuchong: Use "netty" metric group, so that we can add non-memory metrics into it in the future. And rename the class to `NettyMetrics` as well.

fluss-rpc/src/main/java/com/alibaba/fluss/rpc/netty/NettyMemoryMetrics.java
```diff
@@ -0,0 +1,169 @@
...
+                        "numHugeDeallocations",
+                        "numActiveAllocations",
+                        "numActiveTinyAllocations",
+                        "numActiveSmallAllocations",
+                        "numActiveNormalAllocations",
+                        "numActiveHugeAllocations",
+                        "numActiveBytes"));
+    }
+
+    public NettyMemoryMetrics(
```
wuchong: We don't need to instantiate the class, just expose a `registerNettyMetrics(..)` util method is enough.

fluss-rpc/src/main/java/com/alibaba/fluss/rpc/netty/NettyMemoryMetrics.java
```diff
@@ -0,0 +1,169 @@
...
+    public static final String NETTY_MEMORY_USED_DIRECT_MEMORY = "usedDirectMemory";
+    public static final String NETTY_MEMORY_USED_HEAP_ARENAS = "numHeapArenas";
+    public static final String NETTY_MEMORY_USED_DIRECT_ARENAS = "numDirectArenas";
+    public static final String NETTY_MEMORY_TINY_CACHE_SIZE = "tinyCacheSize";
+    public static final String NETTY_MEMORY_SMALL_CACHE_SIZE = "smallCacheSize";
+    public static final String NETTY_MEMORY_NORMAL_CACHE_SIZE = "normalCacheSize";
+    public static final String NETTY_MEMORY_NUM_THREAD_LOCAL_CACHES = "numThreadLocalCaches";
+    public static final String NETTY_MEMORY_CHUNK_SIZE = "chunkSize";
+    public static final String NETTY_MEMORY_HEAP_ARENA = "heapArena";
+    public static final String NETTY_MEMORY_DIRECT_ARENA = "directArena";
```
wuchong: Let's move all the metric names to `MetricNames` where we maintain all the metric names centralized.

wuchong (commented):
Thanks @SteNicholas for the contribution. 

The implementation is great. However, I think most of the registered metrics will never be used. We can reduce the metrics to simplify the implementation and costs, necessary metrics can be introduced when needed in the future. 

Some background:
1. Fluss never allocates heap buffers via Netty, so only direct buffers need to be monitored
2. Fluss may allocate huge buffers for high-throughput log producing scenarios, and huge allocation is time-consuming, so we should monitor huge allocations.

So the suggested metrics to monitor: 

- `usedDirectMemory`: Gauge, direct memory size used by netty
- `numDirectArenas`: Gauge, number of direct arenas in netty
- `allocationsPerSecond`: Meter, the number of allocations every second
- `hugeAllocationsPerSecond`: Meter, the number of huge allocations every second

We don't need to monitor detailed arenas-level metrics, this can simplify the code by removing Java reflection and reduce a lot of metrics. 

Btw, we also need to add documentation for the new metrics in the page: https://alibaba.github.io/fluss-docs/docs/maintenance/monitor-metrics/#threads . We can add a specific section for "Netty", the scope can be `
coordinator/tabletserver/client`.

SteNicholas:
@wuchong, thanks for detailed review and suggestions. I have addressed comments in https://github.com/alibaba/fluss/pull/407/commits/d54880d1adde170b46f8a9030c1dbdae78bb9536. PTAL.

fluss-rpc/src/main/java/com/alibaba/fluss/rpc/netty/NettyMetrics.java
```diff
@@ -0,0 +1,86 @@
...
+        nettyMetricGroup.<Long, Gauge<Long>>gauge(
+                MetricNames.NETTY_USED_DIRECT_MEMORY, pooledAllocatorMetric::usedDirectMemory);
+        nettyMetricGroup.<Integer, Gauge<Integer>>gauge(
+                MetricNames.NETTY_NUM_DIRECT_ARENAS, pooledAllocatorMetric::numDirectArenas);
+        nettyMetricGroup.meter(
+                MetricNames.NETTY_NUM_ALLOCATIONS_PER_SECONDS,
+                new MeterView(
+                        new AllocationsCounter(
+                                () ->
+                                        pooledAllocator.directArenas().stream()
```
wuchong: This is deprecated. Use `pooledAllocatorMetric.directArenas()` instead.

fluss-rpc/src/main/java/com/alibaba/fluss/rpc/netty/NettyMetrics.java
```diff
@@ -0,0 +1,86 @@
...
+        MetricGroup nettyMetricGroup = metricGroup.addGroup(NETTY_METRIC_GROUP);
+        PooledByteBufAllocatorMetric pooledAllocatorMetric = pooledAllocator.metric();
+        nettyMetricGroup.<Long, Gauge<Long>>gauge(
+                MetricNames.NETTY_USED_DIRECT_MEMORY, pooledAllocatorMetric::usedDirectMemory);
+        nettyMetricGroup.<Integer, Gauge<Integer>>gauge(
+                MetricNames.NETTY_NUM_DIRECT_ARENAS, pooledAllocatorMetric::numDirectArenas);
+        nettyMetricGroup.meter(
+                MetricNames.NETTY_NUM_ALLOCATIONS_PER_SECONDS,
+                new MeterView(
+                        new AllocationsCounter(
```
wuchong: We don't need to introduce `AllocationsCounter`, here can be 

```java
        nettyMetricGroup.meter(
                MetricNames.NETTY_NUM_ALLOCATIONS_PER_SECONDS,
                new MeterView(
                        () ->
                                pooledAllocatorMetric.directArenas().stream()
                                        .mapToLong(PoolArenaMetric::numAllocations)
                                        .sum()));
```

fluss-rpc/src/main/java/com/alibaba/fluss/rpc/netty/NettyMetrics.java
```diff
@@ -0,0 +1,86 @@
...
+                                        pooledAllocator.directArenas().stream()
+                                                .mapToLong(PoolArenaMetric::numAllocations)
+                                                .sum())));
+        nettyMetricGroup.meter(
+                MetricNames.NETTY_NUM_HUGE_ALLOCATIONS_PER_SECONDS,
+                new MeterView(
+                        new AllocationsCounter(
+                                () ->
+                                        pooledAllocator.directArenas().stream()
+                                                .mapToLong(PoolArenaMetric::numHugeAllocations)
```
wuchong: ditto.

wuchong (approved):
LGTM.

PR #410, opened 2025-02-17
[common] Declare serialVersionUID explicitly for serializable class

### LINKED ISSUE #409: Declare serialVersionUID explicitly for serializable class
### Search before asking

### Fluss version

0.5.0 (latest release)

### Please describe the bug 🐞

The serialization runtime associates with each serializable class a version number, called a `serialVersionUID`, which is used during deserialization to verify that the sender and receiver of a serialized object have loaded classes for that object that are compatible with respect to serialization. If the receiver has loaded a class for the object that has a different serialVersionUID than that of the corresponding sender's class, then deserialization will result in an `InvalidClassException`. A serializable class can declare its own `serialVersionUID` explicitly by declaring a field named serialVersionUID that must be static, final, and of type long

### Solution

Declare serialVersionUID explicitly for serializable class.

SteNicholas:
### Purpose

Linked issue: close #409

Declare `serialVersionUID` explicitly for serializable class `Password` and `Column`.

### Tests

No.

### API and Format

No.

### Documentation

No.

### DIFF
--- a/fluss-common/src/main/java/com/alibaba/fluss/config/Password.java
+++ b/fluss-common/src/main/java/com/alibaba/fluss/config/Password.java
@@ -20,6 +20,7 @@ import java.io.Serializable;
 
 /** A wrapper class for passwords to hide them while logging a config. */
 public class Password implements Serializable {
+    private static final long serialVersionUID = 1L;
     // the hidden content to be displayed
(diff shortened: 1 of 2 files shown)


wuchong (approved):
LGTM.

PR #417, opened 2025-02-17
[server] Fix the IndexOutOfBoundsException when reading cdc log from pk table with projection

swuferhong:
### Purpose

Linked issue: https://github.com/alibaba/fluss/issues/416

This pr is aims to fix the IndexOutOfBoundsException when reading cdc log from pk table with projection. This case will happen when we create a primary key table with `FIRST_ROW` or  `VERSION_BASED` merge engine enable. As `FIRST_ROW` for example, we will generate cdc log batch which contains no records, however, if we try to read this cdc log batch with projection, we will incorrectly read the arrow header in `FileLogProjection` even if it was empty, this pr is trying to fix it.


swuferhong:
Hi @wuchong  @luoyuxia  @loserwang1024, could you table a look at this bug fix? thx

loserwang1024 (approved):
LGTM

PR #422, opened 2025-02-18
[hotfix] ZstdCompressionCodec should use decompressedSize to get error name

libenchao:
This has been fixed in https://github.com/apache/arrow-java/pull/619

### Purpose

The error message is wrong when for zstd compression codec, I've fixed this in upstream project (https://github.com/apache/arrow-java/pull/619)


SteNicholas (approved):

luoyuxia (approved):
Thanks for the pr. LGTM!

PR #426, opened 2025-02-18
[client] Add retry when get one available tablet serverNode fails

zhaozijun109:
### Purpose

Linked issue: close #xxx


fluss-client/src/main/java/com/alibaba/fluss/client/FlussConnection.java
```diff
@@ -139,14 +139,14 @@ public RemoteFileDownloader getOrCreateRemoteFileDownloader() {
...
                     // create the admin read only gateway
-                    // todo: may add retry logic when no any available tablet server?
                     AdminReadOnlyGateway gateway =
                             GatewayClientProxy.createGatewayProxy(
                                     () ->
                                             getOneAvailableTabletServerNode(
                                                     metadataUpdater.getCluster()),
                                     rpcClient,
                                     AdminReadOnlyGateway.class);
+
```
SteNicholas: Unnecessary change. Pls revert this line.

fluss-client/src/main/java/com/alibaba/fluss/client/utils/MetadataUtils.java
```diff
@@ -52,6 +55,9 @@
 
 /** Utils for metadata for client. */
 public class MetadataUtils {
+    private static final Logger LOG = LoggerFactory.getLogger(MetadataUtils.class);
+
+    private static final int MAX_RETRY_TIMES = 5;
```
SteNicholas: Could this retry time be configurable?

polyzos:
@gkatzioura also added a retry mechanism [here](https://github.com/alibaba/fluss/issues/208#issuecomment-2664769613)
to fix some tests. I assume retry logic will be needed in other places as well, so it maybe makes sense to introduce a class to handle that or use a library. WDYT?

fluss-client/src/main/java/com/alibaba/fluss/client/FlussConnection.java
zhaozijun109: Ok, I will revert this line.

fluss-client/src/main/java/com/alibaba/fluss/client/utils/MetadataUtils.java
zhaozijun109: Yes, I will make it as system configuration

zhaozijun109:
@polyzos Good idea, I will try to make a common util method of retry logic. WDYT? @SteNicholas please left some comments, thank you

fluss-client/src/main/java/com/alibaba/fluss/client/utils/MetadataUtils.java
```diff
@@ -255,10 +261,30 @@ public NewTableMetadata(
     }
 
     public static ServerNode getOneAvailableTabletServerNode(Cluster cluster) {
-        List<ServerNode> aliveTabletServers = cluster.getAliveTabletServerList();
-        if (aliveTabletServers.isEmpty()) {
-            throw new FlussRuntimeException("no alive tablet server in cluster");
+        List<ServerNode> aliveTabletServers = null;
+        for (int retryTimes = 0; retryTimes <= MAX_RETRY_TIMES; retryTimes++) {
+            aliveTabletServers = cluster.getAliveTabletServerList();
+            if (aliveTabletServers.isEmpty()) {
```
luoyuxia, line 272: When `aliveTabletServers` is empty in `cluster`, the retry doesn't work since the cluster wil stil be empty for aliveTabletServers`, we need send metadata request to update `cluster`.

zhaozijun109: OK，I will adjust it，thank you.

SteNicholas:
@zhaozijun109, IMO, the retry mechanism maybe different from different behavior, therefore it may not need to introduce a common util method of retry logic.

zhaozijun109:
@SteNicholas Thank for your advice, I will re-adjust based on all the comments above

polyzos:
@SteNicholas Indeed, but there might be re-use of some implementations.. my thoughts were that it might be best to have them in a centralized place, to avoid introducing multiple retry implementations across the codebase, because I already saw 3 different cases popping up in different places. Whatever you think works best

zhaozijun109:
@luoyuxia @SteNicholas Could you please help me review again when you have free time? Thank you.

fluss-common/src/main/java/com/alibaba/fluss/config/ConfigOptions.java
```diff
@@ -876,6 +876,13 @@ public class ConfigOptions {
                             "Enable metrics for client. When metrics is enabled, the client "
                                     + "will collect metrics and report by the JMX metrics reporter.");
 
+    public static final ConfigOption<Integer> CLIENT_GET_TABLET_SERVER_NODE_MAX_RETRY_TIMES =
```
SteNicholas, line 879: Add this option in `configuration.md`.

SteNicholas:
@zhaozijun109, could you create new issue for this pull request? BTW, could you also add some description for this pull request?

zhaozijun109:
@SteNicholas Thank you for your review, I will create a new issue.

zhaozijun109:
@luoyuxia @SteNicholas The new issue is [#480](https://github.com/alibaba/fluss/issues/480), could you please help me to review again when you have free time, thank you.

wuchong:
Thank you for your contribution, @zhaozijun109! You're absolutely right—the current client metadata implementation has significant issues with retrying, timeout handling, and error management. While this PR effectively addresses the specific case of `getOneAvailableTabletServerNode`, there are many other methods in the client metadata module that could benefit from similar improvements.

To tackle this comprehensively, I've created [Issue #483](https://github.com/alibaba/fluss/issues/483) to propose a broader refactoring of the client metadata implementation. The goal is to address all recent bugs and provide a more robust, general solution.

zhaozijun109:
@wuchong Thank you, I will continue to study the principle of fluss and try my best to contribute it. Another, sounds good that it's very good  for refactoring of the client metadata implementation, many methods will benefit from this.

PR #428, opened 2025-02-19
[datalake/paimon] Introduce paimon encoding

### LINKED ISSUE #408: Introudce pamion's encoding & bucketing implementation to Fluss
### Search before asking

### Motivation

Now, to use pamion's encoding & bucketing, we must include Paimon lib.. But we can introudce paimon's implement to Fluss instead of include paimon lib..

### Solution

Copy the encoding & bucketing implementation from Paimon to fluss

### Anything else?

_No response_

### Willingness to contribute

luoyuxia:
### Purpose

Linked issue: close #408 

Introduce Paimon's encoding way to avoid include paimon bundle lib.

Introduce `PaimonBinaryRowWriter` in Fluss which follows Fluss' encoding row way.

### Tests
PaimonKeyEncoderTest to verify the introduced paimon encoding way is same with paimon's own implementation.
FlinkUnionReadPrimaryKeyTableITCase#testPrimaryKeyTable alreay verify write Fluss row with paimon encoding&bucket way implemented in Fluss, we can still lookup the row correctly by Paimon ap


luoyuxia:
@wuchong The pr is to Introduce paimon encoding to Fluss. Could you please help review it

fluss-common/src/main/java/com/alibaba/fluss/row/encode/paimon/PaimonBinaryRowWriter.java
```diff
@@ -0,0 +1,420 @@
...
+                break;
+            case TIMESTAMP_WITH_LOCAL_TIME_ZONE:
+                final int timestampLtzPrecision = getPrecision(fieldType);
+                fieldWriter =
+                        (writer, pos, value) ->
+                                writer.writeTimestampLtz(
+                                        pos, (TimestampLtz) value, timestampLtzPrecision);
+                break;
+            default:
+                throw new IllegalArgumentException("Unsupported type for IndexedRow: " + fieldType);
```
wuchong:
```suggestion
                throw new IllegalArgumentException("Unsupported type for Paimon BinaryRow writer: " + fieldType);
```

fluss-common/src/main/java/com/alibaba/fluss/row/encode/paimon/PaimonBinaryRowWriter.java
```diff
@@ -0,0 +1,420 @@
...
+        setNullBit(pos);
+        UnsafeUtils.putLong(buffer, getFieldOffset(pos), 0L);
+    }
+
+    private void setNullBit(int pos) {
+        UnsafeUtils.bitSet(buffer, 0, pos + HEADER_SIZE_IN_BITS);
+    }
+
+    public void writeRowKind(RowKind kind) {
+        // Fluss has APPEND_ONLY rowKind, so minus 1 to align with paimon
```
wuchong: Convert Fluss RowKind into Paimon RowKind. If the kind is `APPEND_ONLY`, there may be unpredictable problem to use `-1` kind byte here.

fluss-common/src/test/java/com/alibaba/fluss/row/encode/paimon/PaimonKeyEncoderTest.java
```diff
@@ -0,0 +1,125 @@
...
+        PaimonKeyEncoder paimonKeyEncoder = new PaimonKeyEncoder(allRowType, encodedKeys);
+
+        // encode with Fluss own implementation for Paimon
+        byte[] encodedKey = paimonKeyEncoder.encodeKey(indexedRow);
+
+        // encode with Paimon implementation
+        byte[] paimonEncodedKey = genPaimonRowForAllTypes(allRowType.getFieldCount()).toBytes();
+
+        // verify both the result should be same
+        assertThat(encodedKey).containsExactly(paimonEncodedKey);
```
wuchong: `assertThat(encodedKey).isEqualTo(paimonEncodedKey);` is fine here.

fluss-common/src/main/java/com/alibaba/fluss/row/encode/paimon/PaimonBinaryRowWriter.java
```diff
@@ -0,0 +1,420 @@
...
+    public void writeShort(int pos, short value) {
+        UnsafeUtils.putShort(buffer, getFieldOffset(pos), value);
+    }
+
+    public void writeInt(int pos, int value) {
+        UnsafeUtils.putInt(buffer, getFieldOffset(pos), value);
+    }
+
+    public void writeLong(int pos, long value) {
+        UnsafeUtils.putLong(buffer, getFieldOffset(pos), value);
```
wuchong: I recommend using Fluss's `MemorySegment#putLong` to write data. The issue with `Unsafe#putLong` is that it relies on the native byte order of the machine, which can lead to data corruption if the writer and reader operate on systems with different byte orders. Storage systems should always use a consistent, predefined byte order rather than depending on the machine's native order. I think this is a design flaw in Paimon. 

Since nearly all modern architectures (e.g., x86, ARM) use little-endian, we can safely assume Paimon uses little-endian as well. Using Fluss's `MemorySegment#putLong`, which guarantees little-endian encoding, ensures compatibility and avoids potential data corruption.

luoyuxia: Thanks..Agree with you.. At first, I also want to use `MemorySegment#putLong`, but find Paimon use  native byte order of the machine, so I use `Unsafe#putLong` to align with Paimon. Now, I change to use `MemorySegment#put` method in this class..

luoyuxia:
@wuchong Comments is addressed

wuchong (approved):
LGTM.

PR #450, opened 2025-02-20
[netty] Modify to debug level in NettyServerHandler#exceptionCaught method to avoid to many logs if NLB is mounted

### LINKED ISSUE #377: Connection reset is thrown by Fluss server when I mount a vip
### Search before asking

### Fluss version

0.5.0 (latest release)

### Please describe the bug 🐞

When I mount a vip, which will  perform telnet to servers periodically, but i found the server will throw the following message:

Image

  

### Solution

Now, I'm not sure my vip really does to servers because when I execute telnet to servers in my local env, I can't reproduce this error message..
Seems my vip will terminate connection  qucikly so that connection reset exception is thrown..

luoyuxia:
### Purpose

Linked issue: close #377

Modify to debug level to avoid too many logs if NLB(Network Load Balancer is mounted.


luoyuxia:

wuchong (approved):
LGTM.

PR #454, opened 2025-02-20
[Lake] Introduce fluss-lake module

polyzos:
Introduce fluss-lake module to address [434](https://github.com/alibaba/fluss/issues/434?reload=1)

### DIFF
--- a/fluss-lake/fluss-lake-common/pom.xml
+++ b/fluss-lake/fluss-lake-common/pom.xml
@@ -1,0 +1,16 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+
+    <parent>
+        <groupId>com.alibaba.fluss</groupId>
+        <artifactId>fluss-lake</artifactId>
+        <version>0.6-SNAPSHOT</version>
+    </parent>
+
+    <groupId>com.alibaba.fluss</groupId>
+    <name>Fluss : Lake : Common</name>
+
+</project>
--- a/fluss-lake/pom.xml
+++ b/fluss-lake/pom.xml
@@ -1,0 +1,19 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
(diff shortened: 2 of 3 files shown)


fluss-lake/pom.xml
```diff
@@ -0,0 +1,19 @@
...
+    <modelVersion>4.0.0</modelVersion>
+
+    <parent>
+        <groupId>com.alibaba.fluss</groupId>
+        <artifactId>fluss</artifactId>
+        <version>0.6-SNAPSHOT</version>
+    </parent>
+
+    <artifactId>fluss-lake</artifactId>
+    <name>Fluss : Lake</name>
```
leonardBang: <name>Fluss : Lake :</name>

pom.xml
```diff
@@ -72,6 +72,7 @@
         <module>fluss-connectors</module>
         <module>fluss-protogen</module>
         <module>fluss-benchmark</module>
+        <module>fluss-lake</module>
         <module>fluss-lakehouse</module>
```
leonardBang, line 77: Could you add a note under this line which likes something?
` <!-- fluss-lakehouse module is deprecated, will be removed once fluss-lake ready -->`

leonardBang (commented):
Thanks for your contribution, @polyzos , left two minor comments

polyzos:
@leonardBang i addressed them

luoyuxia:
Note the ci is failing

fluss-lake/fluss-lake-common/pom.xml
```diff
@@ -0,0 +1,31 @@
...
+  -->
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+
+    <parent>
+        <groupId>com.alibaba.fluss</groupId>
+        <artifactId>fluss-lake</artifactId>
+        <version>0.6-SNAPSHOT</version>
```
SteNicholas, line 25: Is this version better to use `0.7-SNAPSHOT` and merge this pull request after releasing cut? IMO, the 0.6 version doesn't include this.

polyzos: @SteNicholas correct this should be inlcuded in the 0.7 release. But how do you typically approach this?
what should the version be since we are still on 0.6-snapshot?

SteNicholas: @polyzos, will the 0.6 version release next week? Is it better to release cut 0.6 and bump version to `0.7-SNAPSHOT` firstly? Otherwise, 0.6 rc would include this module. cc @leonardBang, WDYT?

leonardBang: As we didn't code-freeze yet, I think newly-added module is okay. Even our release version(0.6) contains these modules, but they won't be used by users as we don't expose them via user docs.

leonardBang (approved):

SteNicholas (approved):

Issue #435, opened 2025-02-19
Introduces interfaces for Fluss Lakehouse Storage


leonardBang:
### Search before asking

- [x] I searched in the [issues](https://github.com/alibaba/fluss/issues) and found nothing similar.

### Description

Introduces interfaces for Fluss Lakehouse Storage

### Willingness to contribute

- [ ] I'm willing to submit a PR!

polyzos:
I will also pick this up after [434](https://github.com/alibaba/fluss/issues/434)

polyzos:
PTAL [#457](https://github.com/alibaba/fluss/pull/457)

PR #457, opened 2025-02-21
[Lake] Introduce lakehouse modules

polyzos:
Introduces interfaces for Fluss Lakehouse Storage [#435](https://github.com/alibaba/fluss/issues/435?reload=1)


fluss-lake/fluss-lake-common/src/main/java/com/alibaba/fluss/lake/common/assigner/BucketAssigner.java
```diff
@@ -0,0 +1,31 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.alibaba.fluss.lake.common.assigner;
+
+import com.alibaba.fluss.row.InternalRow;
+
+/** The BucketAssigner interface for assigning buckets to rows. */
+public interface BucketAssigner {
```
wuchong: We have introduced this thing in https://github.com/alibaba/fluss/pull/362. 
But renamed to `BucketingFunction` to avoid naming conflict with existing `BucketAssigner`. So we don't need to introduce `BucketAssigner` and `BucketAssignerFactory` anymore.

fluss-lake/pom.xml
```diff
@@ -31,4 +31,29 @@
...
+            <groupId>org.apache.flink</groupId>
+            <artifactId>flink-connector-files</artifactId>
+            <version>${flink.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.flink</groupId>
+            <artifactId>flink-core</artifactId>
+            <version>${flink.version}</version>
+        </dependency>
+    </dependencies>
```
wuchong: We don’t need to add dependencies to the parent POM since the `fluss-lake` module does not contain any source code. Instead, let’s introduce the necessary dependencies in the sub-modules

wuchong (commented):
It seems the `fluss-lake-common` only provides the interfaces, and no any concrete implementation from the [design doc](https://docs.google.com/document/d/1Ghw_Jb-yHztgGvO5OpRWgibmPClDivejp7UyLUgKxOc/edit?pli=1&tab=t.0). 

How about introducing these interfaces in `fluss-common` to make the modules and dependencies simpler. As every plugin has to dependent on `fluss-common` anyway. cc @luoyuxia @leonardBang ?

polyzos:
does the mean that the `lake-common` becomes redundant now? should it be removed? and have all common in `fluss-common/lakehouse`?

wuchong:
@polyzos yes, let's remove the `fluss-lake-common` from the design.

polyzos:
@wuchong `LakeWriter` though has a dependency on `ScanRecord` which is included in the `client` package and this will introduce a circular dependency if everything goes into common. Unless ScanRecord is also moved to common.. or is there a better approach?

wuchong:
@polyzos maybe we can rely on `LogRecord` which is totally the same.

leonardBang:
> It seems the `fluss-lake-common` only provides the interfaces, and no any concrete implementation from the [design doc](https://docs.google.com/document/d/1Ghw_Jb-yHztgGvO5OpRWgibmPClDivejp7UyLUgKxOc/edit?pli=1&tab=t.0).
> 
> How about introducing these interfaces in `fluss-common` to make the modules and dependencies simpler. As every plugin has to dependent on `fluss-common` anyway. cc @luoyuxia @leonardBang ?

+1

fluss-lake/fluss-lake-common/src/main/java/com/alibaba/fluss/lake/common/writer/LakeTieringFactory.java
```diff
@@ -0,0 +1,70 @@
...
+public interface LakeTieringFactory<WriteResult, CommitableT> {
+
+    /**
+     * Creates a lake writer to write Fluss's rows to Paimon/Iceberg rows.
+     *
+     * @param writerInitContext the context for initializing the writer
+     * @return the lake writer
+     * @throws IOException if an I/O error occurs
+     */
+    CompactWriter<WriteResult> createLakeWriter(WriterInitContext writerInitContext)
```
luoyuxia: CompactWriter -> LakeWriter

fluss-lake/pom.xml
```diff
@@ -31,4 +31,29 @@
...
+            <groupId>com.alibaba.fluss</groupId>
+            <artifactId>fluss-common</artifactId>
+            <version>${project.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>com.alibaba.fluss</groupId>
+            <artifactId>fluss-client</artifactId>
+            <version>${project.version}</version>
+        </dependency>
+        <!-- Flink Dependencies -->
```
luoyuxia: I think we don't need to introduce Flink dependency in fluss-lake module, we will have a `fluss-tiering-service-flink` module. We will put Flink Dependencies in `fluss-tiering-service-flink` module.

luoyuxia (commented):
@polyzos Thanks. Left some comments.

.idea/vcs.xml
```diff
@@ -15,6 +15,6 @@
     </option>
   </component>
   <component name="VcsDirectoryMappings">
-    <mapping directory="$PROJECT_DIR$" vcs="Git" />
+    <mapping directory="" vcs="Git" />
```
luoyuxia: why change this?

fluss-common/src/main/java/com/alibaba/fluss/lakehouse/writer/LakeTieringFactory.java
```diff
@@ -0,0 +1,69 @@
...
+    SimpleVersionedSerializer<WriteResult> getWriteResultSerializer();
+
+    /**
+     * Creates a lake committer to commit to Paimon/Iceberg.
+     *
+     * @param committerInitContext the context for initializing the committer
+     * @return the lake committer
+     * @throws IOException if an I/O error occurs
+     */
+    Committer<CommitableT> createLakeCommitter(CommitterInitContext committerInitContext)
```
wuchong: I think this should return `LakeCommitter`, thus we don't need to rely on Flink dependencies in `fluss-common`? cc @luoyuxia maybe a typo on the design doc?

luoyuxia: Yes, it should be  a typo on the design doc.

PR #466, opened 2025-02-21
[core] Add FlushConsolidationHandler to RPC Sever to optimize syscalls

dao-jun:
### Purpose

When we are writing/reading a lot of small entries, one of the main bottleneck in server becomes the CPU usage. A big part of it is caused by the syscall overhead when writing to the socket.

The reason is that we have many independent (and small) operations happening on the connection and each time we call writeAndFlush() on each of them, causing many write() calls on the socket.

### Tests

NONE

### DIFF
--- a/fluss-rpc/src/main/java/com/alibaba/fluss/rpc/netty/server/ServerChannelInitializer.java
+++ b/fluss-rpc/src/main/java/com/alibaba/fluss/rpc/netty/server/ServerChannelInitializer.java
@@ -20,6 +20,7 @@ import com.alibaba.fluss.rpc.netty.NettyLogger;
 import com.alibaba.fluss.shaded.netty4.io.netty.channel.ChannelInitializer;
 import com.alibaba.fluss.shaded.netty4.io.netty.channel.socket.SocketChannel;
 import com.alibaba.fluss.shaded.netty4.io.netty.handler.codec.LengthFieldBasedFrameDecoder;
+import com.alibaba.fluss.shaded.netty4.io.netty.handler.flush.FlushConsolidationHandler;
 import com.alibaba.fluss.shaded.netty4.io.netty.handler.timeout.IdleStateHandler;
 
 import static com.alibaba.fluss.utils.Preconditions.checkArgument;
@@ -47,6 +48,7 @@ final class ServerChannelInitializer extends ChannelInitializer<SocketChannel> {
         if (nettyLogger.getLoggingHandler() != null) {
             ch.pipeline().addLast("loggingHandler", nettyLogger.getLoggingHandler());
         }
+        ch.pipeline().addLast("consolidation", new FlushConsolidationHandler(1024, true));
         ch.pipeline()
                 .addLast(
                         "frameDecoder",


wuchong:
Great idea! We will estimate the performance effect of this change on our internal use case.

luoyuxia:
@wuchong I have verified this patch in our internal use case.. One flink job to write to fluss, and another job to consume it. 
Seems make no difference about the perfomance in my case.. And I also try to make batch size smaller, seems still no much difference. 
With zstd compression, the server cpu usage is almost same, about 0.5%
Without zstd compression, the server cpu usage is also almost same, about 2.2%
The cpu usage of Flink jobs are also almost same.

luoyuxia:
@wuchong I have verified this patch in our internal use case. One flink job to write to fluss, and another job to consume it. 
Seems make no difference about the perfomance in my case.  And I also try to make batch size smaller for fetch and write , seems still no much difference.. 
With zstd compression, the server cpu usage is almost same, about 0.5%
Without zstd compression, the server cpu usage is also almost same, about 2.2%
The cpu usage of Flink jobs are also almost same.

Seems our internal use case can't simulate the env that the pr try to resolve..

fluss-rpc/src/main/java/com/alibaba/fluss/rpc/netty/server/ServerChannelInitializer.java
```diff
@@ -47,6 +48,7 @@ protected void initChannel(SocketChannel ch) {
         if (nettyLogger.getLoggingHandler() != null) {
             ch.pipeline().addLast("loggingHandler", nettyLogger.getLoggingHandler());
         }
+        ch.pipeline().addLast("consolidation", new FlushConsolidationHandler(1024, true));
```
luoyuxia, line 51: Haven't dig into the source code. Curious about when the flush count is not reach `1024`, when will it been flushed? What's the strategy?

dao-jun:
> @wuchong I have verified this patch in our internal use case. One flink job to write to fluss, and another job to consume it. Seems make no difference about the perfomance in my case. And I also try to make batch size smaller for fetch and write , seems still no much difference.. With zstd compression, the server cpu usage is almost same, about 0.5% Without zstd compression, the server cpu usage is also almost same, about 2.2% The cpu usage of Flink jobs are also almost same.
> 
> Seems our internal use case can't simulate the env that the pr try to resolve..

https://github.com/jchambers/pushy/pull/657

fluss-rpc/src/main/java/com/alibaba/fluss/rpc/netty/server/ServerChannelInitializer.java
dao-jun: it will submit a flush task to netty io thread.  Run as soon as possible, but still yield to give a chance for additional writes to enqueue.

wuchong:
Based on the description of `FlushConsolidationHandler`, it appears to be highly beneficial for handling high-QPS small requests. However, this does not apply to Produce/Fetch Log operations. While Lookups can fall into this category, however, we already employ batching strategies for Lookups via `LookupQueue`. This likely explains why we do not observe any significant difference in our internal use case.

wuchong:
Closing it. We can reopen it if we find it has benefits for real use-cases.

PR #515, opened 2025-03-03
[common]Introduce Predicate to do filter and partition push down.

Alibaba-HZY:
[common]

Purpose
Because Flink Expression and Spark v1 Filter is not serializable, so fluss introduces Predicate to do filter and partition push down.

Tests
com.alibaba.fluss.predicate.PredicateBuilderTest
com.alibaba.fluss.predicate.PredicateTest

API and Format
no

Documentation
no


fluss-common/src/main/java/com/alibaba/fluss/predicate/And.java
```diff
@@ -0,0 +1,67 @@
...
+import java.util.Optional;
+
+/* This file is based on source code of Apache Paimon Project (https://paimon.apache.org/), licensed by the Apache
+ * Software Foundation (ASF) under the Apache License, Version 2.0. See the NOTICE file distributed with this work for
+ * additional information regarding copyright ownership. */
+
+/** A {@link CompoundPredicate.Function} to eval and. */
+public class And extends CompoundPredicate.Function {
+
+    private static final long serialVersionUID = 1L;
```
platinumhamburg: It is recommended to use the built-in serialver tool in the JDK to generate a unique UID, rather than uniformly using the constant value 1, as this helps detect serialization and deserialization compatibility issues.

fluss-common/src/main/java/com/alibaba/fluss/predicate/FieldRef.java
```diff
@@ -0,0 +1,79 @@
...
+import java.util.Objects;
+
+/* This file is based on source code of Apache Paimon Project (https://paimon.apache.org/), licensed by the Apache
+ * Software Foundation (ASF) under the Apache License, Version 2.0. See the NOTICE file distributed with this work for
+ * additional information regarding copyright ownership. */
+
+/** A reference to a field in an input. */
+public class FieldRef implements Serializable {
+
+    private static final long serialVersionUID = 1L;
```
platinumhamburg, line 32: As previously suggested, it is not advisable to uniformly use the constant value 1 for serialVersionUID.

fluss-common/src/main/java/com/alibaba/fluss/predicate/And.java
platinumhamburg: I've discussed with Jark, and we concluded that we should follow the Flink guide https://flink.apache.org/how-to-contribute/code-style-and-quality-java/#java-serialization, because serialver has strict compatibility checks that restrict developers' flexibility.

fluss-common/src/main/java/com/alibaba/fluss/predicate/CompareUtils.java
```diff
@@ -0,0 +1,60 @@
...
+        if (v1 instanceof Comparable) {
+            // because BinaryString can not serialize so v1 or v2 may be BinaryString convert to
+            // String for compare
+            if (v1 instanceof BinaryString) {
+                v1 = ((BinaryString) v1).toString();
+            }
+            if (v2 instanceof BinaryString) {
+                v2 = ((BinaryString) v2).toString();
+            }
+            return ((Comparable<Object>) v1).compareTo(v2);
```
wuchong: `BinaryString` supports comparable, we don't need to cast to String.

fluss-common/src/main/java/com/alibaba/fluss/predicate/Predicate.java
```diff
@@ -0,0 +1,50 @@
...
+ * Software Foundation (ASF) under the Apache License, Version 2.0. See the NOTICE file distributed with this work for
+ * additional information regarding copyright ownership. */
+
+/**
+ * Predicate which returns Boolean and provides testing by stats.
+ *
+ * @see PredicateBuilder
+ * @since 0.4.0
+ */
+public interface Predicate extends Serializable {
```
wuchong, line 37: Considering we have already implemented the statistic-based `test()` method, we should also expose it on the `Predicate`, in order to have complete test on the implementations.

fluss-common/src/main/java/com/alibaba/fluss/predicate/PredicateBuilder.java
```diff
@@ -0,0 +1,454 @@
...
+                predicate = predicateTemp;
+            } else {
+                predicate = PredicateBuilder.and(predicate, predicateTemp);
+            }
+        }
+        return predicate;
+    }
+
+    public static Predicate partitions(
+            List<Map<String, String>> partitions, RowType rowType, String defaultPartValue) {
```
wuchong: I think we can remove the `defaultPartValue` parameter, because Fluss doesn't support default values for partitions. Otherwise, it's hard to use this method in Fluss.

fluss-common/src/test/java/com/alibaba/fluss/predicate/PredicateTest.java
```diff
@@ -0,0 +1,424 @@
...
+
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+
+import static com.alibaba.fluss.row.BinaryString.fromString;
+import static org.assertj.core.api.Assertions.assertThat;
+
+/** Test for {@link Predicate}s. */
+public class PredicateTest {
```
wuchong, line 37: We should add statistic based tests on Predicate.

wuchong (commented):
I rebased and appended a commit to mention the mentioned comments and fix some license and code sytle problems.

fluss-common/src/main/java/com/alibaba/fluss/predicate/CompareUtils.java
HZY-Wade: because BinaryString  can not serializable ,in predicate use string instead, so in this method need cast value(BinaryString) to string

fluss-common/src/test/java/com/alibaba/fluss/predicate/PredicateTest.java
platinumhamburg: The statistic tests seems like based on PR of ArrayType support.

fluss-common/src/main/java/com/alibaba/fluss/predicate/Predicate.java
platinumhamburg: We can introduce the statistic-based test() method after ArrayType support PR merged.

wuchong: @platinumhamburg I added the statistic-based test() using `Long[]`. We can change it to `InternalArray` when it is introduced, but it's just an optimization.

fluss-common/src/main/java/com/alibaba/fluss/predicate/CompareUtils.java
wuchong: @Alibaba-HZY If `BinaryString` is not serializable, the predicate should use `String` literals, and the comparison parameters should be of type `String` rather than `BinaryString`. That means we still don't need to cast `BinaryString` to `String`.

wuchong:
@Alibaba-HZY @platinumhamburg If you are good for the changes in https://github.com/apache/fluss/pull/515/commits/938e26b0c0e8e3f3890c4fe04acab7d2a91bfaaf, I will merge this PR.

HZY-Wade:
> @Alibaba-HZY @platinumhamburg If you are good for the changes in [938e26b](https://github.com/apache/fluss/commit/938e26b0c0e8e3f3890c4fe04acab7d2a91bfaaf), I will merge this PR.

lgtm , left a comment

fluss-common/src/main/java/com/alibaba/fluss/predicate/CompareUtils.java
wuchong: @Alibaba-HZY I created https://github.com/apache/fluss/issues/1495 to fix `BinaryString` supports Java serialization. Then we can use `BinaryString` in Predicate and compare `BinaryString` directly.

wuchong:
Merging

Issue #524, opened 2025-03-04
Bootstrap server config as one of TabletServer cannot do admin operation like createDatabase


swuferhong:
### Search before asking

- [x] I searched in the [issues](https://github.com/alibaba/fluss/issues) and found nothing similar.

### Fluss version

main (development)

### Please describe the bug 🐞

Currently, if I config `bootstrap server` as one of `TabletServer`, I cannot do admin operation like createDatabase. 

Upon further debug, I found that `createDatabase` request was mistakenly sent to the TabletServer instead of the CoordinatorServer.

### Are you willing to submit a PR?

- [x] I'm willing to submit a PR!

PR #525, opened 2025-03-04
[server] Fix admin operation cannot be done error when bootstrap server config as one of TabletServer

swuferhong:
### Purpose

Linked issue: #524 

This pr is aims to fix the admin operation cannot be done error when `bootstrap server` config as one of `TabletServer`.  The reason for the issue is that the `NettyClient` has a connection cache, and the key for the cache is the `uid` of `ServerNode`, which is in parttern as `'cs' + id` or `'ts' + id`. In pr #456,  the UID of the `bootstrap server` was changed from `'ts--1'` to `'cs--1'`, which conflicted with our default `CoordinatorServer UID`, causing it to be identified as the same connection in the cache. The fix way is to change the default `CoordinatorServer UID` to `'cs-0'`.

### DIFF
--- a/fluss-client/src/test/java/com/alibaba/fluss/client/admin/FlussAdminITCase.java
+++ b/fluss-client/src/test/java/com/alibaba/fluss/client/admin/FlussAdminITCase.java
@@ -16,6 +16,8 @@
 
 package com.alibaba.fluss.client.admin;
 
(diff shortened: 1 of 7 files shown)


swuferhong:
@wuchong  @libenchao could you take a look at this bug fix? thx.

wuchong (approved):

PR #539, opened 2025-03-05
[docs] move 'Development Guide' to 'community'

### LINKED ISSUE #537: Move "Developer Guide" from Docs to Community
### Search before asking

### Description

The **"Developer Guide"**, which includes sections such as **"Building Fluss from Source"** and **"Import Fluss into an IDE"**, is currently located under the **`website/docs/*`** directory. However, these guides are not tied to any specific version of Fluss and are primarily intended for contributors rather than end users. As such, they are better suited for the **Community Documentation (`website/community/*`)** section, which focuses on resources for the contributor community.

### Willingness to contribute

gyang94:
### Purpose

Linked issue: close #537


.gitignore
```diff
@@ -36,3 +36,4 @@ website/package-lock.json
 website/versioned_docs
 website/versioned_sidebars
 website/versions.json
+website/pnpm-lock.yaml
```
wuchong, line 39: why we need to add this?

gyang94: If users build `websites` with `npm`, it will generate a lock file called `package-lock.json`, which has been already included in the git ignore file.
There are other package managers to build node projects, like `pnpm` and `yarn`. They will also generate lock file called `pnpm-lock.yaml` or `yarn.lock`. Generally those lock files should not be included in git, unless we want to "lock" the dependencies's version.
I build it with `pnpm` locally, thus I added it.

wuchong: thanks for the explanation.

Issue #311, opened 2025-01-06
[Bug] The time out for UpdateMetadata request too short to cause job failover


luoyuxia:
### Search before asking

- [X] I searched in the [issues](https://github.com/alibaba/fluss/issues) and found nothing similar.

### Fluss version

main

### Minimal reproduce step

Currently, `sendMetadataRequestAndRebuildCluster` will have a timeout of 3s. After 3s, the future will be completed without updating metadata. 

### What doesn't meet your expectations?

In my case, a partitioned table with 512 buckets, 512 parallelism for flink sink, it'll be timeout easily and then cause sink job fail..

First, write to a partition, it will try to update the metadata in method `checkAndUpdatePartitionMetadata` to fetch the partition's metadata. If it timeout, the metadata won't be updated, and it will then throw PartitionNotExists exception althogth the Partition does exist.

For my case, a time out of 60s works....

### Anything else?

I can see we need to introduce a request time out mechanism to avoid a request to hang out forever.. But for updating metadata request, it should throw Timeout exception instead of just log it to enable caller to decide retry or fail directly..

For example, when creating `FlussTable`, it'll try to fetch the metadata of the table in [metadataUpdater.checkAndUpdateTableMetadata(Collections.singleton(tablePath))](https://github.com/alibaba/fluss/blob/92c6d23bd25b9de61e05859e18c5b267c81f80db/fluss-client/src/main/java/com/alibaba/fluss/client/table/FlussTable.java#L145). If the metadata is timeout, the metadata can't be updated and cause it to throw `table not found in cluster` exception although the table does exist... At least, it should throw timeout exception instead of `table not found in cluster` exception which is really confused.

### Are you willing to submit a PR?

- [ ] I'm willing to submit a PR!

wuchong:
rafactor metadata updater in this issue.

PR #542, opened 2025-03-07
[hotfix] Increase timeout for update metadata request

luoyuxia:
### Purpose

If timeout is short, it will cause the werid exception as described in #311 , this pr is aimed to increase timeout to 30s.
30s should be fine for most case..If not, the next following pr to introduce global configurable timeout and cache to polish it.


luoyuxia:
I'd like to try to run serveral tests to make sure it won't make ci unstable

luoyuxia:
Remove time out will make ci unstable, so let's increase it to 30s. Ci is stable in 30s..

wuchong (approved):

PR #553, opened 2025-03-08
[Lake/Common] Introduces pluggable lakehouse interfaces

SML0127:
This close https://github.com/alibaba/fluss/issues/436


wuchong (commented):
Could you move the interfaces into package `com.alibaba.fluss.lakehouse.*` and add `@since 0.7` version annotations?

SML0127:
> Could you move the interfaces into package `com.alibaba.fluss.lakehouse.*` and add `@since 0.7` version annotations?

@wuchong applied your comments.

wuchong (approved):

PR #577, opened 2025-03-11
Introduce authenticate in Fluss.

### LINKED ISSUE #484: Fluss support authentication
### Search before asking

### Motivation

Fluss support authentication

### Solution

- [ ] Provide Plugin Authentication Interface.
- [ ] Provide an builtin Authenticator implement based on Sasl/PlainText.
- [ ] Add Test for Sasl/PlainText.

### Anything else?

_No response_

### Willingness to contribute

loserwang1024:
### Purpose

Linked issue: close #484 . This Pr is depended on #531 .

### Brief change log

 Introduce authenticate in Fluss.

### Tests
com.alibaba.fluss.rpc.netty.authenticate.AuthenticationTest

#### Config
* `server.authenticate.protocol.map`: A map defining the authentication protocol for each listener. The format is 'listenerName1:protocol1,listenerName2:protocol2', e.g., 'INTERNAL:PLAINTEXT,CLIENT:GSSAPI'. Each listener can be associated with a specific authentication protocol. Listeners not included in the map will use PLAINTEXT by default, which does not require authentication.
* client.authenticate.protocol:The authentication protocol used to authenticate the client.

The config of one protocol:
* `client.authenticate.${protocol}.xxx`
* `server.authenticate.${protocol}.xxx`

#### Interface
AuthenticationPlugin for client and server:
```java
/**
 * The AuthenticationPlugin interface defines a contract for authentication mechanisms in the Fluss
 * RPC system. Implementations of this interface provide specific authentication protocols (e.g.,
 * PLAINTEXT, AK-SK) and must be registered via the Service Provider Interface (SPI) mechanism to be
 * discoverable by the system.
 *
 * <p>Authentication plugins are used by {@link AuthenticatorLoader} to create client and server
 * authenticators based on configuration settings such as {@link
 * com.alibaba.fluss.config.ConfigOptions#CLIENT_AUTHENTICATE_PROTOCOL} and {@link
 * com.alibaba.fluss.config.ConfigOptions#SERVER_AUTHENTICATE_PROTOCOL_MAP}.
 *
 * <p>To implement this interface:
 *
 * <ol>
 *   <li>Define the supported authentication protocol via {@link #authProtocol()}.
 *   <li>Implement the required logic for client/server authentication (via subclasses like {@link
 *       ClientAuthenticationPlugin}).
 *   <li>Register the plugin implementation in {@code
 *       META-INF/services/com.alibaba.fluss.rpc.authenticate.AuthenticationPlugin}.
 * </ol>
 *
 * @since 0.1
 */
public interface AuthenticationPlugin extends Plugin {
    /**
     * Returns the authentication protocol identifier for this plugin (e.g., "PLAINTEXT", "AK-SK").
     * This identifier is used by the system to match plugins with configuration settings and select
     * the appropriate authentication mechanism.
     *
     * <p>It is typically configured via:
     *
     * <ul>
     *   <li>{@link com.alibaba.fluss.config.ConfigOptions#CLIENT_AUTHENTICATE_PROTOCOL} for
     *       client-side authentication.
     *   <li>{@link com.alibaba.fluss.config.ConfigOptions#SERVER_AUTHENTICATE_PROTOCOL_MAP} for
     *       server listener-specific authentication.
     * </ul>
     *
     * @return The protocol name (e.g., "PLAINTEXT").
     */
    String authProtocol();
}


public interface ClientAuthenticationPlugin extends AuthenticationPlugin {

    ClientAuthenticator createClientAuthenticator(Configuration configuration);
}

public interface ServerAuthenticationPlugin extends AuthenticationPlugin {

    ServerAuthenticator createServerAuthenticator(Configuration configuration);
}

```

Authenticator for client and server:
```java
public interface ServerAuthenticator {

    String protocol();

    byte[] evaluateResponse(byte[] token) throws AuthenticationException;

    /**
     * Create principal from authenticated token for later authorization.(this can only invoke if is
     * complete).
     */
    FlussPrincipal createPrincipal();

    boolean isComplete();
}
```
```java
public interface ClientAuthenticator {

    String protocol();

    byte[] authenticate(byte[] data);

    boolean isComplete();
}

```

Principal to identify user:
```java
/**
 * Represents a security principal in Fluss, defined by a {@code name} and {@code type}.
 *
 * <p>The principal type indicates the category of the principal (e.g., "User", "Group", "Role"),
 * while the name identifies the specific entity within that category. By default, the simple
 * authorizer uses "User" as the principal type, but custom authorizers can extend this to support
 * role-based or group-based access control lists (ACLs).
 *
 * <p>Example usage:
 *
 * <ul>
 *   <li>{@code new FlussPrincipal("admin", "User")} – A standard user principal.
 *   <li>{@code new FlussPrincipal("admins", "Group")} – A group-based principal for authorization.
 * </ul>
 *
 * @since 0.1
 */
public class FlussPrincipal implements Principal {
    public static final FlussPrincipal ANONYMOUS = new FlussPrincipal("ANONYMOUS", "User");

    private final String name;
    private final String type;

    public FlussPrincipal(String name, String type) {
        this.name = name;
        this.type = type;
    }

    @Override
    public String getName() {
        return name;
    }

    public String getType() {
        return type;
    }

    @Override
        if (o == null || getClass() != o.getClass()) {
            return false;
        }
        FlussPrincipal that = (FlussPrincipal) o;
        return Objects.equals(name, that.name) && Objects.equals(type, that.type);
    }

    @Override
    public int hashCode() {
        return Objects.hash(name, type);
    }
}

```

### DIFF
--- a/fluss-common/src/main/java/com/alibaba/fluss/security/auth/AuthenticationFactory.java
+++ b/fluss-common/src/main/java/com/alibaba/fluss/security/auth/AuthenticationFactory.java
@@ -1,0 +1,183 @@
+/*
+ *  Copyright (c) 2025 Alibaba Group Holding Ltd.
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+
+package com.alibaba.fluss.security.auth;
+
+import com.alibaba.fluss.config.ConfigOptions;
+import com.alibaba.fluss.config.Configuration;
+import com.alibaba.fluss.metadata.ValidationException;
+import com.alibaba.fluss.plugin.PluginManager;
+import com.alibaba.fluss.plugin.PluginUtils;
+
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.HashMap;
+import java.util.Iterator;
+import java.util.List;
+import java.util.Map;
+import java.util.ServiceLoader;
+import java.util.function.Supplier;
+import java.util.stream.Collectors;
+
+/**
+ * A manager responsible for loading and configuring client and server authenticators based on the
+ * provided configuration.
+ *
+ * <p>This class discovers authentication plugins via the classpath and configured plugins, and
+ * creates suppliers for authenticators using plugin-specific configurations.
+ *
+ * <p>Key functionalities include:
+ *
+ * <ul>
+ *   <li>Loading client authenticators based on {@link ConfigOptions#CLIENT_SECURITY_PROTOCOL}.
+ *   <li>Loading server authenticators for multiple endpoints, using listener-specific protocols
+ *       defined in {@link ConfigOptions#SERVER_SECURITY_PROTOCOL_MAP}.
+ *   <li>Discovering plugins through {@link ServiceLoader} and custom {@link PluginManager} for
+ *       extensibility.
+ * </ul>
+ *
+ * @since 0.7
+ */
+public class AuthenticationFactory {
+    private static final String CLIENT_AUTHENTICATOR_PREFIX = "client.security.";
+    private static final String SERVER_AUTHENTICATOR_PREFIX = "security.";
+
+    /**
+     * Loads a supplier for a client authenticator based on the configuration.
+     *
+     * @param configuration The configuration containing authentication settings and protocol
+     *     definitions.
+     * @return A supplier for creating the client authenticator.
+     */
+    public static Supplier<ClientAuthenticator> loadClientAuthenticatorSupplier(
+            Configuration configuration) {
+        String clientAuthenticateProtocol =
+                configuration.getString(ConfigOptions.CLIENT_SECURITY_PROTOCOL);
+        ClientAuthenticationPlugin authenticatorPlugin =
+                discoverPlugin(
+                        configuration,
+                        clientAuthenticateProtocol,
+                        ClientAuthenticationPlugin.class);
+
+        Map<String, String> allConfig = configuration.toMap();
+        Map<String, String> authConfig = new HashMap<>();
+        String prefix = CLIENT_AUTHENTICATOR_PREFIX + clientAuthenticateProtocol + ".";
+        allConfig
+                .keySet()
+                .forEach(
+                        key -> {
+                            if (key.startsWith(prefix)) {
+                                authConfig.put(key.substring(prefix.length()), allConfig.get(key));
+                            }
+                        });
+
+        return () ->
+                authenticatorPlugin.createClientAuthenticator(Configuration.fromMap(authConfig));
+    }
+
+    /**
+     * Loads suppliers for server authenticators for each endpoint, based on listener-specific
+     * protocols.
+     *
+     * @param configuration The configuration containing authentication settings and protocol
+     *     definitions.
+     * @return A map mapping listener names to suppliers for their corresponding server
+     *     authenticators.
+     */
+    public static Map<String, Supplier<ServerAuthenticator>> loadServerAuthenticatorSuppliers(
+            Configuration configuration) {
+
+        Map<String, Supplier<ServerAuthenticator>> serverAuthenticators = new HashMap<>();
+        Map<String, String> protocolMap =
+                configuration.getMap(ConfigOptions.SERVER_SECURITY_PROTOCOL_MAP);
+        Map<String, String> allConfigMap = configuration.toMap();
+
+        for (Map.Entry<String, String> protocolEntry : protocolMap.entrySet()) {
+
+            String serverAuthenticateProtocol = protocolEntry.getValue();
+            ServerAuthenticationPlugin serverAuthenticatorPlugin =
+                    discoverPlugin(
+                            configuration,
+                            serverAuthenticateProtocol,
+                            ServerAuthenticationPlugin.class);
+
+            // get the protocol config.
+            Map<String, String> authConfigMap = new HashMap<>();
+            String prefix = SERVER_AUTHENTICATOR_PREFIX + serverAuthenticateProtocol + ".";
+            allConfigMap
+                    .keySet()
+                    .forEach(
+                            key -> {
+                                if (key.startsWith(prefix)) {
+                                    authConfigMap.put(
+                                            key.substring(prefix.length()), allConfigMap.get(key));
+                                }
+                            });
+            Configuration authConfig = Configuration.fromMap(authConfigMap);
+            serverAuthenticators.put(
+                    protocolEntry.getKey(),
+                    () -> serverAuthenticatorPlugin.createServerAuthenticator(authConfig));
+        }
+        return serverAuthenticators;
+    }
+
+    /**
+     * Discovers an authentication plugin of the specified type and protocol from the classpath and
+     * configured plugins.
+     *
+     * @param configuration The configuration used to initialize the plugin manager.
+     * @param protocol The protocol name (e.g., "PLAINTEXT", "SASL_PLAIN") to match the plugin's
+     *     {@link AuthenticationPlugin#authProtocol()}.
+     * @return The discovered plugin instance.
+     * @throws ValidationException If no plugin or multiple plugins match the given protocol and
+     *     interface.
+     */
+    @SuppressWarnings("unchecked")
+    private static <T extends AuthenticationPlugin> T discoverPlugin(
+            Configuration configuration, String protocol, Class<T> pluginInterface) {
+        PluginManager pluginManager = PluginUtils.createPluginManagerFromRootFolder(configuration);
+        Collection<Supplier<Iterator<AuthenticationPlugin>>> pluginSuppliers = new ArrayList<>(2);
+        pluginSuppliers.add(() -> ServiceLoader.load(AuthenticationPlugin.class).iterator());
+        pluginSuppliers.add(() -> pluginManager.load(AuthenticationPlugin.class));
+
+        List<T> matchingPlugins = new ArrayList<>();
+        for (Supplier<Iterator<AuthenticationPlugin>> pluginIteratorsSupplier : pluginSuppliers) {
+            final Iterator<AuthenticationPlugin> foundPlugins = pluginIteratorsSupplier.get();
+            while (foundPlugins.hasNext()) {
+                AuthenticationPlugin plugin = foundPlugins.next();
+                if (plugin.authProtocol().equals(protocol)
+                        && pluginInterface.isAssignableFrom(plugin.getClass())) {
+                    matchingPlugins.add((T) plugin);
+                }
+            }
+        }
+        if (matchingPlugins.size() != 1) {
+            throw new ValidationException(
+                    String.format(
+                            "Could not find same authenticator plugin for protocol '%s' in the classpath.\n\n"
+                                    + "Available factory protocols are:\n\n"
+                                    + "%s",
+                            protocol,
+                            matchingPlugins.stream()
+                                    .map(f -> f.getClass().getName())
+                                    .distinct()
+                                    .sorted()
+                                    .collect(Collectors.joining("\n"))));
+        }
+
+        return matchingPlugins.get(0);
+    }
+}
--- a/fluss-common/src/main/java/com/alibaba/fluss/security/auth/AuthenticationFactory.java
+++ b/fluss-common/src/main/java/com/alibaba/fluss/security/auth/AuthenticationFactory.java
@@ -68,11 +68,7 @@ public class AuthenticationFactory {
         String clientAuthenticateProtocol =
                 configuration.getString(ConfigOptions.CLIENT_SECURITY_PROTOCOL);
         ClientAuthenticationPlugin authenticatorPlugin =
-                discoverPlugin(
-                        configuration,
-                        clientAuthenticateProtocol,
-                        ClientAuthenticationPlugin.class,
-                        null);
+                discoverPlugin(clientAuthenticateProtocol, ClientAuthenticationPlugin.class, null);
         return () -> authenticatorPlugin.createClientAuthenticator(configuration);
     }
 
@@ -87,17 +83,14 @@ public class AuthenticationFactory {
      */
     public static Map<String, Supplier<ServerAuthenticator>> loadServerAuthenticatorSuppliers(
             Configuration configuration) {
-
         PluginManager pluginManager = PluginUtils.createPluginManagerFromRootFolder(configuration);
         Map<String, Supplier<ServerAuthenticator>> serverAuthenticators = new HashMap<>();
         Map<String, String> protocolMap =
                 configuration.getMap(ConfigOptions.SERVER_SECURITY_PROTOCOL_MAP);
         for (Map.Entry<String, String> protocolEntry : protocolMap.entrySet()) {
-
             String serverAuthenticateProtocol = protocolEntry.getValue();
             ServerAuthenticationPlugin serverAuthenticatorPlugin =
                     discoverPlugin(
-                            configuration,
                             serverAuthenticateProtocol,
                             ServerAuthenticationPlugin.class,
                             pluginManager);
@@ -113,7 +106,6 @@ public class AuthenticationFactory {
      * Discovers an authentication plugin of the specified type and protocol from the classpath and
      * configured plugins.
      *
-     * @param configuration The configuration used to initialize the plugin manager.
      * @param protocol The protocol name (e.g., "PLAINTEXT", "SASL_PLAIN") to match the plugin's
      *     {@link AuthenticationPlugin#authProtocol()}.
      * @return The discovered plugin instance.
@@ -122,10 +114,7 @@ public class AuthenticationFactory {
      */
     @SuppressWarnings("unchecked")
     private static <T extends AuthenticationPlugin> T discoverPlugin(
-            Configuration configuration,
-            String protocol,
-            Class<T> pluginInterface,
-            @Nullable PluginManager pluginManager) {
+            String protocol, Class<T> pluginInterface, @Nullable PluginManager pluginManager) {
 
         Collection<Supplier<Iterator<AuthenticationPlugin>>> pluginSuppliers = new ArrayList<>(2);
         pluginSuppliers.add(() -> ServiceLoader.load(AuthenticationPlugin.class).iterator());
@@ -138,7 +127,7 @@ public class AuthenticationFactory {
             final Iterator<AuthenticationPlugin> foundPlugins = pluginIteratorsSupplier.get();
             while (foundPlugins.hasNext()) {
                 AuthenticationPlugin plugin = foundPlugins.next();
-                if (plugin.authProtocol().equals(protocol)
+                if (plugin.authProtocol().equalsIgnoreCase(protocol)
                         && pluginInterface.isAssignableFrom(plugin.getClass())) {
                     matchingPlugins.add((T) plugin);
                 }
--- a/fluss-common/src/main/java/com/alibaba/fluss/security/auth/AuthenticationFactory.java
+++ b/fluss-common/src/main/java/com/alibaba/fluss/security/auth/AuthenticationFactory.java
@@ -22,6 +22,8 @@ import com.alibaba.fluss.metadata.ValidationException;
 import com.alibaba.fluss.plugin.PluginManager;
 import com.alibaba.fluss.plugin.PluginUtils;
 
+import javax.annotation.Nullable;
+
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.HashMap;
@@ -52,7 +54,6 @@ import java.util.stream.Collectors;
  * @since 0.7
  */
 public class AuthenticationFactory {
-    private static final String CLIENT_AUTHENTICATOR_PREFIX = "client.security.";
     private static final String SERVER_AUTHENTICATOR_PREFIX = "security.";
 
     /**
@@ -70,22 +71,9 @@ public class AuthenticationFactory {
                 discoverPlugin(
                         configuration,
                         clientAuthenticateProtocol,
-                        ClientAuthenticationPlugin.class);
-
-        Map<String, String> allConfig = configuration.toMap();
-        Map<String, String> authConfig = new HashMap<>();
-        String prefix = CLIENT_AUTHENTICATOR_PREFIX + clientAuthenticateProtocol + ".";
-        allConfig
-                .keySet()
-                .forEach(
-                        key -> {
-                            if (key.startsWith(prefix)) {
-                                authConfig.put(key.substring(prefix.length()), allConfig.get(key));
-                            }
-                        });
-
-        return () ->
-                authenticatorPlugin.createClientAuthenticator(Configuration.fromMap(authConfig));
+                        ClientAuthenticationPlugin.class,
+                        null);
+        return () -> authenticatorPlugin.createClientAuthenticator(configuration);
     }
 
     /**
@@ -100,11 +88,10 @@ public class AuthenticationFactory {
     public static Map<String, Supplier<ServerAuthenticator>> loadServerAuthenticatorSuppliers(
             Configuration configuration) {
 
+        PluginManager pluginManager = PluginUtils.createPluginManagerFromRootFolder(configuration);
         Map<String, Supplier<ServerAuthenticator>> serverAuthenticators = new HashMap<>();
         Map<String, String> protocolMap =
                 configuration.getMap(ConfigOptions.SERVER_SECURITY_PROTOCOL_MAP);
-        Map<String, String> allConfigMap = configuration.toMap();
-
         for (Map.Entry<String, String> protocolEntry : protocolMap.entrySet()) {
 
             String serverAuthenticateProtocol = protocolEntry.getValue();
@@ -112,24 +99,12 @@ public class AuthenticationFactory {
                     discoverPlugin(
                             configuration,
                             serverAuthenticateProtocol,
-                            ServerAuthenticationPlugin.class);
+                            ServerAuthenticationPlugin.class,
+                            pluginManager);
 
-            // get the protocol config.
-            Map<String, String> authConfigMap = new HashMap<>();
-            String prefix = SERVER_AUTHENTICATOR_PREFIX + serverAuthenticateProtocol + ".";
-            allConfigMap
-                    .keySet()
-                    .forEach(
-                            key -> {
-                                if (key.startsWith(prefix)) {
-                                    authConfigMap.put(
-                                            key.substring(prefix.length()), allConfigMap.get(key));
-                                }
-                            });
-            Configuration authConfig = Configuration.fromMap(authConfigMap);
             serverAuthenticators.put(
                     protocolEntry.getKey(),
-                    () -> serverAuthenticatorPlugin.createServerAuthenticator(authConfig));
+                    () -> serverAuthenticatorPlugin.createServerAuthenticator(configuration));
         }
         return serverAuthenticators;
     }
@@ -147,11 +122,16 @@ public class AuthenticationFactory {
      */
     @SuppressWarnings("unchecked")
     private static <T extends AuthenticationPlugin> T discoverPlugin(
-            Configuration configuration, String protocol, Class<T> pluginInterface) {
-        PluginManager pluginManager = PluginUtils.createPluginManagerFromRootFolder(configuration);
+            Configuration configuration,
+            String protocol,
+            Class<T> pluginInterface,
+            @Nullable PluginManager pluginManager) {
+
         Collection<Supplier<Iterator<AuthenticationPlugin>>> pluginSuppliers = new ArrayList<>(2);
         pluginSuppliers.add(() -> ServiceLoader.load(AuthenticationPlugin.class).iterator());
-        pluginSuppliers.add(() -> pluginManager.load(AuthenticationPlugin.class));
+        if (pluginManager != null) {
+            pluginSuppliers.add(() -> pluginManager.load(AuthenticationPlugin.class));
+        }
 
         List<T> matchingPlugins = new ArrayList<>();
         for (Supplier<Iterator<AuthenticationPlugin>> pluginIteratorsSupplier : pluginSuppliers) {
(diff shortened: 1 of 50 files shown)


loserwang1024:
@wuchong , I have rebased this pr into main, would you like to help review it?

wuchong:
`Flink119CatalogITCase>FlinkCatalogITCase.testAuthentication:449` is failed.

wuchong:
I'm going to review this PR, could you rebase to the main branch?

loserwang1024:
> could you rebase to the main branch

@wuchong , Done it

fluss-common/src/main/java/com/alibaba/fluss/security/auth/AuthenticationFactory.java
```diff
@@ -0,0 +1,183 @@
...
+            final Iterator<AuthenticationPlugin> foundPlugins = pluginIteratorsSupplier.get();
+            while (foundPlugins.hasNext()) {
+                AuthenticationPlugin plugin = foundPlugins.next();
+                if (plugin.authProtocol().equals(protocol)
+                        && pluginInterface.isAssignableFrom(plugin.getClass())) {
+                    matchingPlugins.add((T) plugin);
+                }
+            }
+        }
+        if (matchingPlugins.size() != 1) {
```
wuchong: throw different exceptions for `>1` (multiple plugins for the same protocol)  and `=0` (not found).

fluss-common/src/main/java/com/alibaba/fluss/security/auth/AuthenticationFactory.java
```diff
@@ -0,0 +1,183 @@
...
+     * @param protocol The protocol name (e.g., "PLAINTEXT", "SASL_PLAIN") to match the plugin's
+     *     {@link AuthenticationPlugin#authProtocol()}.
+     * @return The discovered plugin instance.
+     * @throws ValidationException If no plugin or multiple plugins match the given protocol and
+     *     interface.
+     */
+    @SuppressWarnings("unchecked")
+    private static <T extends AuthenticationPlugin> T discoverPlugin(
+            Configuration configuration, String protocol, Class<T> pluginInterface) {
+        PluginManager pluginManager = PluginUtils.createPluginManagerFromRootFolder(configuration);
```
wuchong: The `PluginManager` should be optional, it should be null for client. Check how `com.alibaba.fluss.fs.FileSystem#initialize` implement it.

fluss-common/src/main/java/com/alibaba/fluss/security/auth/AuthenticationFactory.java
```diff
@@ -0,0 +1,183 @@
...
+                configuration.getString(ConfigOptions.CLIENT_SECURITY_PROTOCOL);
+        ClientAuthenticationPlugin authenticatorPlugin =
+                discoverPlugin(
+                        configuration,
+                        clientAuthenticateProtocol,
+                        ClientAuthenticationPlugin.class);
+
+        Map<String, String> allConfig = configuration.toMap();
+        Map<String, String> authConfig = new HashMap<>();
+        String prefix = CLIENT_AUTHENTICATOR_PREFIX + clientAuthenticateProtocol + ".";
```
wuchong: use lowercase `clientAuthenticateProtocol.toLowerCase()` to support lowercase as well? such as: `client.security.ram.ak`.

fluss-common/src/main/java/com/alibaba/fluss/security/auth/AuthenticationFactory.java
```diff
@@ -0,0 +1,183 @@
...
+        allConfig
+                .keySet()
+                .forEach(
+                        key -> {
+                            if (key.startsWith(prefix)) {
```
wuchong: Use `key.toLowerCase().startWith(prefix)` to support lowercase as well? such as `client.security.ram.ak`.

fluss-common/src/main/java/com/alibaba/fluss/security/auth/AuthenticationFactory.java
```diff
@@ -0,0 +1,183 @@
...
+ *   <li>Loading client authenticators based on {@link ConfigOptions#CLIENT_SECURITY_PROTOCOL}.
+ *   <li>Loading server authenticators for multiple endpoints, using listener-specific protocols
+ *       defined in {@link ConfigOptions#SERVER_SECURITY_PROTOCOL_MAP}.
+ *   <li>Discovering plugins through {@link ServiceLoader} and custom {@link PluginManager} for
+ *       extensibility.
+ * </ul>
+ *
+ * @since 0.7
+ */
+public class AuthenticationFactory {
```
wuchong, line 56: Add UT for this class.

fluss-common/src/main/java/com/alibaba/fluss/security/auth/AuthenticationFactory.java
```diff
@@ -0,0 +1,183 @@
...
+                            ServerAuthenticationPlugin.class);
+
+            // get the protocol config.
+            Map<String, String> authConfigMap = new HashMap<>();
+            String prefix = SERVER_AUTHENTICATOR_PREFIX + serverAuthenticateProtocol + ".";
+            allConfigMap
+                    .keySet()
+                    .forEach(
+                            key -> {
+                                if (key.startsWith(prefix)) {
```
wuchong: ditto

fluss-common/src/main/java/com/alibaba/fluss/security/auth/AuthenticationFactory.java
```diff
@@ -0,0 +1,183 @@
...
+                .keySet()
+                .forEach(
+                        key -> {
+                            if (key.startsWith(prefix)) {
+                                authConfig.put(key.substring(prefix.length()), allConfig.get(key));
+                            }
+                        });
+
+        return () ->
+                authenticatorPlugin.createClientAuthenticator(Configuration.fromMap(authConfig));
```
wuchong: Do we really need to strip the prefix for the configuration? It seems the prefix is static, not dynamic, and the plugin can use a complete key for the configuration, such as `client.security.ram.ak`.

fluss-common/src/main/java/com/alibaba/fluss/security/auth/PlainTextAuthenticationPlugin.java
```diff
@@ -0,0 +1,77 @@
...
+ *  limitations under the License.
+ */
+
+package com.alibaba.fluss.security.auth;
+
+import com.alibaba.fluss.config.Configuration;
+
+/** Authentication Plugin for PLAINTEXT which not need to do authentication. */
+public class PlainTextAuthenticationPlugin
+        implements ServerAuthenticationPlugin, ClientAuthenticationPlugin {
```
wuchong, line 24: I can't remember why we need to separate `ServerAuthenticationPlugin` and `ClientAuthenticationPlugin` instead of using one combined interface?

fluss-common/src/main/java/com/alibaba/fluss/security/auth/ClientAuthenticator.java
```diff
@@ -0,0 +1,27 @@
...
+
+/** Authenticator for client side. */
+public interface ClientAuthenticator {
+
+    String protocol();
+
+    byte[] authenticate(byte[] data);
+
+    boolean isComplete();
+}
```
wuchong: Add javadocs for the methods, currently, it's hard to understand what's the meaning of the parameters and return values, and how to implement them.

fluss-common/src/main/java/com/alibaba/fluss/security/auth/ServerAuthenticator.java
```diff
@@ -0,0 +1,29 @@
...
+import com.alibaba.fluss.exception.AuthenticationException;
+
+/** Authenticator for server side. */
+public interface ServerAuthenticator {
+
+    String protocol();
+
+    byte[] evaluateResponse(byte[] token) throws AuthenticationException;
+
+    boolean isComplete();
```
wuchong: ditto

fluss-rpc/src/main/java/com/alibaba/fluss/rpc/netty/client/ServerConnection.java
```diff
@@ -356,7 +366,43 @@ private void handleApiVersionsResponse(ApiMessage response, Throwable cause) {
         synchronized (lock) {
             serverApiVersions =
                     new ServerApiVersions(((ApiVersionsResponse) response).getApiVersionsList());
-            state = ConnectionState.READY;
+            LOG.info("Begin to authenticate with protocol {}", authenticator.protocol());
```
wuchong: change to DEBUG level? When connections increase, the log will be very verbose.

fluss-rpc/src/main/java/com/alibaba/fluss/rpc/netty/client/ServerConnection.java
```diff
@@ -356,7 +366,43 @@ private void handleApiVersionsResponse(ApiMessage response, Throwable cause) {
...
+            return;
+        }
+
+        synchronized (lock) {
+            sendAuthenticate(((AuthenticateResponse) response).getChallenge());
+        }
+    }
+
+    private void switchState(ConnectionState targetState) {
+        LOG.info("switch state form {} to {}", state, targetState);
```
wuchong: change to DEBUG level?

fluss-rpc/src/main/java/com/alibaba/fluss/rpc/netty/client/ServerConnection.java
```diff
@@ -356,7 +366,43 @@ private void handleApiVersionsResponse(ApiMessage response, Throwable cause) {
...
+            close(cause);
+            return;
+        }
+        if (!(response instanceof AuthenticateResponse)) {
+            close(new IllegalStateException("Unexpected response type " + response.getClass()));
+            return;
+        }
+
+        synchronized (lock) {
+            sendAuthenticate(((AuthenticateResponse) response).getChallenge());
```
wuchong: should check reponse `hasChallenge` before `getChallenge`

fluss-rpc/src/main/java/com/alibaba/fluss/rpc/netty/server/NettyServer.java
```diff
@@ -228,7 +220,11 @@ private Map<String, NetworkProtocolPlugin> getProtocolsByListenerName() {
         for (NetworkProtocolPlugin protocol : protocols) {
             for (String listenerName : protocol.listenerNames(conf)) {
                 checkState(
-                        !protocolsByListenerName.containsKey(listenerName),
+                        !protocolsByListenerName.containsKey(listenerName)
+                                || protocolsByListenerName
+                                        .get(listenerName)
+                                        .name()
+                                        .equals(FLUSS_PROTOCOL_NAME),
```
wuchong: don't need, remove it

fluss-rpc/src/main/java/com/alibaba/fluss/rpc/netty/server/NettyServer.java
```diff
@@ -237,13 +233,19 @@ private Map<String, NetworkProtocolPlugin> getProtocolsByListenerName() {
...
+        // Add the Fluss protocol plugin first to allow its listener names to be overridden by other
+        // protocols if necessary
+        protocolPlugins.add(new FlussProtocolPlugin(conf, serverType, listeners, requestsMetrics));
         if (conf.get(ConfigOptions.KAFKA_ENABLED)) {
-            return Collections.singletonList(
-                    loadProtocolPlugin(NetworkProtocolPlugin.KAFKA_PROTOCOL_NAME));
-        } else {
-            return Collections.emptyList();
+            protocolPlugins.add(loadProtocolPlugin(NetworkProtocolPlugin.KAFKA_PROTOCOL_NAME));
         }
```
wuchong:
```java
if (conf.get(ConfigOptions.KAFKA_ENABLED)) {
            NetworkProtocolPlugin kafkaPlugin =
                    loadProtocolPlugin(NetworkProtocolPlugin.KAFKA_PROTOCOL_NAME);
            listeners.removeAll(kafkaPlugin.listenerNames(conf));
            protocolPlugins.add(kafkaPlugin);
        }
        // Add the Fluss protocol plugin in the end to allow other protocol
        // pick their listener names first
        protocolPlugins.add(new FlussProtocolPlugin(conf, serverType, listeners, requestsMetrics));
```

fluss-rpc/src/main/java/com/alibaba/fluss/rpc/netty/server/NettyServer.java
```diff
@@ -305,4 +307,9 @@ public CompletableFuture<Void> closeAsync() {
                 channelShutdownFuture,
                 workerShutdownFuture);
     }
+
+    @VisibleForTesting
+    public ApiManager getApiManager() {
+        return ((FlussProtocolPlugin) protocols.get(0)).getApiManager();
```
wuchong: We can keep `apiManager` in the member fields of `NettyServer` to avoid hardcode get the first protocol in the `protocols` (the element index is very hack and there is no any guarantees for this)

fluss-rpc/src/main/java/com/alibaba/fluss/rpc/netty/server/NettyServer.java
```diff
@@ -90,11 +89,17 @@ public NettyServer(
...
-        this.apiManager = new ApiManager(service.providerType());
         this.endpoints = checkNotNull(endpoints, "endpoints");
-        this.protocols = loadProtocols(conf);
+        this.protocols =
+                loadProtocols(
+                        conf,
+                        service.providerType(),
+                        endpoints.stream()
+                                .map(Endpoint::getListenerName)
+                                .collect(Collectors.toList()),
```
wuchong: The `loadProtocols` can accept `Collection<Endpoint>` as a paramter, to move the verbose conversion into `loadProtocols`. This can make the code cleaner.

fluss-common/src/main/java/com/alibaba/fluss/security/auth/ClientAuthenticator.java
```diff
@@ -0,0 +1,27 @@
...
+package com.alibaba.fluss.security.auth;
+
+/** Authenticator for client side. */
+public interface ClientAuthenticator {
+
+    String protocol();
+
+    byte[] authenticate(byte[] data);
+
+    boolean isComplete();
```
wuchong: `isCompleted`

fluss-common/src/main/java/com/alibaba/fluss/security/auth/ServerAuthenticator.java
wuchong: `isCompleted`

fluss-rpc/src/main/java/com/alibaba/fluss/rpc/netty/server/NettyServerHandler.java
```diff
@@ -290,4 +308,62 @@ private Optional<RequestsMetrics.Metrics> getMetrics(FlussRequest request) {
...
+        }
+        future.complete(authenticateResponse);
+
+        if (authenticator.isComplete()) {
+            switchState(ConnectionState.READY);
+        }
+    }
+
+    private void switchState(ConnectionState targetState) {
+        LOG.info("switch state form {} to {}", state, targetState);
```
wuchong: change to DEBUG level?

fluss-rpc/src/main/java/com/alibaba/fluss/rpc/netty/server/NettyServerHandler.java
```diff
@@ -125,9 +134,13 @@ public void channelRead(ChannelHandlerContext ctx, Object msg) throws Exception
                     inflightResponseMap.computeIfAbsent(apiKey, k -> new ArrayDeque<>());
             inflightResponses.addLast(request);
             future.whenCompleteAsync((r, t) -> sendResponse(ctx, apiKey), ctx.executor());
-            requestChannel.putRequest(request);
+            if (state.isAuthenticating() && apiKey != ApiKeys.API_VERSIONS.id) {
```
wuchong: How to keep compatiblity for old version client that doesn't send authenticate request? Add such a test to verify the compatiblity?

fluss-rpc/src/main/java/com/alibaba/fluss/rpc/netty/server/ServerChannelInitializer.java
```diff
@@ -70,13 +80,20 @@ protected void initChannel(SocketChannel ch) {
                         // initialBytesToStrip=0 to include the frame size field after decoding
                         new LengthFieldBasedFrameDecoder(Integer.MAX_VALUE, 0, 4, 0, 0));
         ch.pipeline().addLast("idle", new IdleStateHandler(0, 0, maxIdleTimeSeconds));
+        ServerAuthenticator serverAuthenticator = authenticatorSupplier.get();
+        LOG.info(
```
wuchong: change to DEBUG log level.

fluss-rpc/src/main/java/com/alibaba/fluss/rpc/RpcGateway.java
```diff
@@ -29,4 +32,25 @@ public interface RpcGateway {
...
+     *
+     * @param request The authenticate request (not used in this method's implementation).
+     * @return Always returns {@code null} since the actual authentication handling occurs in {@link
+     *     com.alibaba.fluss.rpc.netty.server.NettyServerHandler}.
+     * @see com.alibaba.fluss.rpc.netty.server.NettyServerHandler#channelRead(ChannelHandlerContext,
+     *     Object) For authentication processing implementation.
+     */
+    @RPC(api = ApiKeys.AUTHENTICATE)
+    default CompletableFuture<AuthenticateResponse> authenticate(AuthenticateRequest request) {
+        return CompletableFuture.completedFuture(new AuthenticateResponse());
```
wuchong: throw an exception to indicate this method should be called directly?

fluss-rpc/src/main/proto/FlussApi.proto
```diff
@@ -742,4 +742,14 @@ message PbPartitionInfo {
...
+}
+
+message AuthenticateRequest {
+  required string protocol = 1;
+  required bytes token = 2;
+}
+
+message AuthenticateResponse {
+  optional bytes challenge = 1;
+}
```
wuchong: move near to XxxRequest and XxxResponse definition.

fluss-rpc/src/test/java/com/alibaba/fluss/rpc/netty/authenticate/AuthenticationTest.java
```diff
@@ -0,0 +1,212 @@
...
+                            metricGroup,
+                            RequestsMetrics.createCoordinatorServerRequestMetrics(metricGroup));
+            // override method of LIST_DATABASES for test request required authentication before,
+            nettyServer
+                    .getApiManager()
+                    .registerApiMethod(
+                            ApiKeys.LIST_DATABASES,
+                            TestingAuthenticateGatewayService.class.getMethod(
+                                    "listDatabases", ListDatabasesRequest.class),
+                            ServerType.COORDINATOR);
```
wuchong: This seems super hack. The `TestingAuthenticateGatewayService` can implements `TestCoordinatorGateway` with only implementing `listDatabases` method. In this way, we don't need to introduce `NettyServer#getApiManager` and `ApiManager#registerApiMethod` methods.

fluss-server/src/main/java/com/alibaba/fluss/server/coordinator/CoordinatorServer.java
```diff
@@ -126,6 +126,9 @@ public CoordinatorServer(Configuration conf) {
     }
 
     public static void main(String[] args) {
+        System.out.println("-----");
+        System.out.println(args);
+        System.out.println(args.length);
```
wuchong: Are they introduced by mistake? Shall we remove it?

fluss-flink/fluss-flink-common/src/main/java/com/alibaba/fluss/flink/catalog/FlinkCatalog.java
```diff
@@ -94,18 +94,20 @@ public class FlinkCatalog implements Catalog {
     protected final String bootstrapServers;
     protected Connection connection;
     protected Admin admin;
-
     private volatile @Nullable LakeCatalog lakeCatalog;
+    private final Map<String, String> securityConfigs;
```
wuchong: nit: move near to the final variables

fluss-flink/fluss-flink-common/src/main/java/com/alibaba/fluss/flink/catalog/FlinkCatalogFactory.java
```diff
@@ -49,12 +53,21 @@ public Set<ConfigOption<?>> optionalOptions() {
...
-        helper.validate();
+        helper.validateExcept(CLIENT_SECURITY_PREFIX);
+        Map<String, String> options = context.getOptions();
+        Map<String, String> securityConfigs = new HashMap<>();
+        options.forEach(
+                (key, value) -> {
+                    if (key.startsWith(CLIENT_SECURITY_PREFIX)) {
+                        securityConfigs.put(key, value);
+                    }
+                });
```
wuchong:
```suggestion
        Map<String, String> securityConfigs = extractPrefix(options, CLIENT_SECURITY_PREFIX);
```
with static import `import static com.alibaba.fluss.utils.PropertiesUtils.extractPrefix;`

fluss-flink/fluss-flink-common/src/main/java/com/alibaba/fluss/flink/catalog/FlinkTableFactory.java
```diff
@@ -200,17 +205,20 @@ public Set<ConfigOption<?>> optionalOptions() {
...
         // forward all client configs
-        for (ConfigOption<?> option : FlinkConnectorOptions.CLIENT_OPTIONS) {
-            if (tableOptions.get(option) != null) {
-                flussConfig.setString(option.key(), tableOptions.get(option).toString());
-            }
-        }
+        tableOptions.forEach(
+                (key, value) -> {
+                    if (FlussConfigUtils.CLIENT_OPTIONS.containsKey(key)
+                            || key.startsWith(CLIENT_SECURITY_PREFIX)) {
```
wuchong: I think we can simplify to `if (key.startsWith(CLIENT_PREFIX))`. Otherwise, `FlussConfigUtils.CLIENT_OPTIONS.containsKey(key)` may have problems that can't forward keys if using a old client version that doesn't contain new `client.*` configurations  registered in the table.

fluss-flink/fluss-flink-common/src/test/java/com/alibaba/fluss/flink/catalog/FlinkCatalogITCase.java
```diff
@@ -421,6 +424,69 @@ void testCreateTableWithUnknownOptions() {
...
+                    new FlinkCatalog(
+                            CATALOG_NAME,
+                            DEFAULT_DB,
+                            bootstrapServers,
+                            Thread.currentThread().getContextClassLoader(),
+                            Collections.emptyMap());
+            Catalog finalAuthenticateCatalog = authenticateCatalog;
+            assertThatThrownBy(finalAuthenticateCatalog::open)
+                    .cause()
+                    .hasMessageContaining("The connection has not completed authentication yet.");
```
wuchong: Missing to configure security configuration is very common, but the exception message is hard to guide users to configure security configurations. Can we improve the message to indicate this like missing to configure "client.security.xx.*" configs.

fluss-common/src/main/java/com/alibaba/fluss/security/auth/ClientAuthenticator.java
```diff
@@ -0,0 +1,27 @@
...
+ */
+
+package com.alibaba.fluss.security.auth;
+
+/** Authenticator for client side. */
+public interface ClientAuthenticator {
+
+    String protocol();
+
+    byte[] authenticate(byte[] data);
```
wuchong: Add `@Nullable` annotation as the return type can be `null`.

fluss-rpc/src/main/java/com/alibaba/fluss/rpc/netty/server/NettyServerHandler.java
```diff
@@ -290,4 +308,62 @@ private Optional<RequestsMetrics.Metrics> getMetrics(FlussRequest request) {
...
+            return;
+        }
+
+        AuthenticateRequest authenticateRequest = (AuthenticateRequest) requestMessage;
+        if (!authenticator.protocol().equals(authenticateRequest.getProtocol())) {
+            future.completeExceptionally(
+                    new AuthenticationException(
+                            String.format(
+                                    "Authenticate protocol not match: protocol of server is %s while protocol of client is %s",
+                                    authenticator.protocol(), authenticateRequest.getProtocol())));
```
wuchong: Add test in `AuthenticationTest` to verify this exception.

fluss-common/src/main/java/com/alibaba/fluss/security/auth/AuthenticationFactory.java
loserwang1024: Why strip the prefix for the configuration? Because I want to strict the prefix by code rather than convention(in case if custom plugin's key is conflict with later fluss configuration). 
Maybe we can remove it and add into documents to tell use use this prefix to develop their own authenticationPlugin.

fluss-common/src/main/java/com/alibaba/fluss/security/auth/AuthenticationFactory.java
loserwang1024: Remove this prefix check,  plugin itself define it.

fluss-common/src/main/java/com/alibaba/fluss/security/auth/AuthenticationFactory.java
loserwang1024: just remove it and will mention it in document later.

fluss-common/src/main/java/com/alibaba/fluss/security/auth/PlainTextAuthenticationPlugin.java
loserwang1024: There are some reason:
1.  ClientAuthenticationPlugin is always packaged in client jar, and ServerAuthenticationPlugin is included in server side.
2. sometime, user want to shade the detail of ServerAuthenticationPlugin(maybe involving some internal logical).

 I also found pulsar do same things[2], which is more clear to use.
https://pulsar.apache.org/docs/4.0.x/security-athenz/#enable-athenz-authentication-on-brokersproxies

fluss-rpc/src/main/java/com/alibaba/fluss/rpc/netty/server/NettyServerHandler.java
loserwang1024: If the server is not set any authenticator, then will use plaintext. Most tests aleady incudes it. If client is old version will server is not, just throw exception:  " Connection is still in the authentication process. Unable to handle API key"

fluss-rpc/src/test/java/com/alibaba/fluss/rpc/netty/authenticate/AuthenticationTest.java
loserwang1024: I have tried it before. However, TestCoordinatorGateway is in fluss-server, while this test is fluss-rpc. And fluss-server is depedends on fluss-rpc. If The TestingAuthenticateGatewayService can implements TestCoordinatorGateway, this will create a circular dependency.

It seems I have to move this test into fluss-server.

fluss-rpc/src/main/java/com/alibaba/fluss/rpc/netty/server/NettyServerHandler.java
```diff
@@ -290,4 +309,62 @@ private Optional<RequestsMetrics.Metrics> getMetrics(FlussRequest request) {
...
+            return;
+        }
+
+        AuthenticateResponse authenticateResponse = new AuthenticateResponse();
+        if (!authenticator.isCompleted()) {
+            byte[] token = authenticateRequest.getToken();
+            byte[] challenge;
+            if (!authenticator.isCompleted()
+                    && (challenge = authenticator.evaluateResponse(token)) != null) {
```
wuchong: should `evaluateResponse` first then check `authenticator.isCompleted()`?

fluss-common/src/main/java/com/alibaba/fluss/security/auth/AuthenticationFactory.java
```diff
@@ -0,0 +1,170 @@
...
+ *   <li>Loading server authenticators for multiple endpoints, using listener-specific protocols
+ *       defined in {@link ConfigOptions#SERVER_SECURITY_PROTOCOL_MAP}.
+ *   <li>Discovering plugins through {@link ServiceLoader} and custom {@link PluginManager} for
+ *       extensibility.
+ * </ul>
+ *
+ * @since 0.7
+ */
+public class AuthenticationFactory {
+    private static final String SERVER_AUTHENTICATOR_PREFIX = "security.";
```
wuchong, line 57: not used

fluss-common/src/main/java/com/alibaba/fluss/security/auth/AuthenticationFactory.java
```diff
@@ -0,0 +1,170 @@
...
+     * @param configuration The configuration used to initialize the plugin manager.
+     * @param protocol The protocol name (e.g., "PLAINTEXT", "SASL_PLAIN") to match the plugin's
+     *     {@link AuthenticationPlugin#authProtocol()}.
+     * @return The discovered plugin instance.
+     * @throws ValidationException If no plugin or multiple plugins match the given protocol and
+     *     interface.
+     */
+    @SuppressWarnings("unchecked")
+    private static <T extends AuthenticationPlugin> T discoverPlugin(
+            Configuration configuration,
```
wuchong: not used

fluss-common/src/main/java/com/alibaba/fluss/security/auth/AuthenticationFactory.java
```diff
@@ -0,0 +1,170 @@
...
+        if (pluginManager != null) {
+            pluginSuppliers.add(() -> pluginManager.load(AuthenticationPlugin.class));
+        }
+
+        List<T> matchingPlugins = new ArrayList<>();
+        for (Supplier<Iterator<AuthenticationPlugin>> pluginIteratorsSupplier : pluginSuppliers) {
+            final Iterator<AuthenticationPlugin> foundPlugins = pluginIteratorsSupplier.get();
+            while (foundPlugins.hasNext()) {
+                AuthenticationPlugin plugin = foundPlugins.next();
+                if (plugin.authProtocol().equals(protocol)
```
wuchong: We may use uppercase when developing the auth plugin(such as: `RAM`, `SSL`, `SASL`), but we can loose this to `equalsIgnoreCase` to allow users use lowercase in client side:

```
client.security.protocol: ram
client.security.ram.ak: xxxx
```

instead of having to use

```
client.security.protocol: RAM
client.security.RAM.ak: xxxx
```

fluss-rpc/src/main/java/com/alibaba/fluss/rpc/netty/server/NettyServerHandler.java
wuchong: OK. I checked that if server configured PLAINTEXT, the initial `ConnectionState` is ready, so old version client can still work without sending AUTH request.

wuchong (approved):
LGTM. I appended a commit to address the following minor comemtns.

wuchong:
@loserwang1024 please check the appended commit again.

Issue #570, opened 2025-03-10
Flink sink support hash by bucket key for PrimaryKey Table


swuferhong:
### Search before asking

- [x] I searched in the [issues](https://github.com/alibaba/fluss/issues) and found nothing similar.

### Motivation

Currently, while testing the performance of writing a `PrimaryKey` table in Flink, it has been observed that when the number of `buckets` for the `PrimaryKey` table is large, like 512, the write performance of the Flink sink is acceptable only when the parallelism is set to 512. If set to 256 or 108, there will be a significant performance bottleneck. Further debugging reveals that the client is stuck in memory allocation.

The reason for this situation is that Fluss client's batching data is based on `TableBucket`. However, due to the absence of a hash strategy, each client may process data from all `TableBuckets`, resulting in poor batching performance.

The solution to this problem is to add a re-hash strategy before the sink, so that data with the same Fluss bucket id is processed by the same concurrency, which can improve batching efficiency.

### Willingness to contribute

- [x] I'm willing to submit a PR!

PR #579, opened 2025-03-11
[flink] Flink sink support hash by bucket id for PrimaryKey Table

swuferhong:
### Purpose

Linked issue: #570 

This pr is aims to support hash by bucket key for `PrimaryKey` Table for Flink sink. The approach of this pr is to first hash the data by `bucket id` before writing to the Flink sink, sending data with the same `bucket id` to the same sink parallelism. This allows the Fluss writer to batch kv data more effectively, thus reducing the pressure on the client and improving throughput. From the testing results, this can greatly enhance the throughput of writing Fluss `PrimaryKey` tables and reduce memory usage, while also reducing pressure on the server side (fewer `putKv` requests and larger CDC log batches).


fluss-connectors/fluss-connector-flink/src/main/java/com/alibaba/fluss/connector/flink/FlinkConnectorOptions.java
```diff
@@ -106,6 +106,15 @@ public class FlinkConnectorOptions {
                     .defaultValue(false)
                     .withDescription("Whether to ignore retract（-U/-D) record.");
 
+    public static final ConfigOption<Boolean> SINK_PRE_HASH_BY_BUCKET_KEY =
+            ConfigOptions.key("sink.pre-hash-by-bucket-key")
+                    .booleanType()
+                    .defaultValue(false)
```
wuchong: default to true for better performance?

We should also add tests for enable and disable this option.

fluss-connectors/fluss-connector-flink/src/main/java/com/alibaba/fluss/connector/flink/FlinkConnectorOptions.java
```diff
@@ -106,6 +106,15 @@ public class FlinkConnectorOptions {
                     .defaultValue(false)
                     .withDescription("Whether to ignore retract（-U/-D) record.");
 
+    public static final ConfigOption<Boolean> SINK_PRE_HASH_BY_BUCKET_KEY =
+            ConfigOptions.key("sink.pre-hash-by-bucket-key")
```
wuchong: `sink.bucket-shuffle` for simplification. Besides, it's not "hash by bucket key", but "shuffle by bucket id". Please update the description of the option.

wuchong: Please update documentation in this PR as well.

fluss-connectors/fluss-connector-flink/src/main/java/com/alibaba/fluss/connector/flink/sink/FlinkSink.java
```diff
@@ -90,6 +104,11 @@ public AppendSinkWriterBuilder(
         public AppendSinkWriter createWriter() {
             return new AppendSinkWriter(tablePath, flussConfig, tableRowType, ignoreDelete);
         }
+
+        @Override
+        public DataStream<RowData> addPreWriteTopology(DataStream<RowData> input) {
+            return input;
```
wuchong: should also add bucket shuffle if it has bucket key?

fluss-connectors/fluss-connector-flink/src/main/java/com/alibaba/fluss/connector/flink/sink/FlinkSink.java
```diff
@@ -102,24 +121,42 @@ static class UpsertSinkWriterBuilder implements SinkWriterBuilder<UpsertSinkWrit
...
                     tablePath, flussConfig, tableRowType, targetColumnIndexes, ignoreDelete);
         }
+
+        @Override
+        public DataStream<RowData> addPreWriteTopology(DataStream<RowData> input) {
+            return sinkReHash
+                    ? input.partitionCustom(
+                            (bucketId, numPartitions) -> bucketId % numPartitions,
+                            new RowDataKeySelector(bucketKeyEncoder, numBucket))
+                    : input;
```
wuchong: It would be better to customize the the shuffle name on the topology graph on Web UI. Currently, it is "CUSTOM", would be better to name "BUCKET_SHUFFLE" to be more meaningful for users. 

This might need to call more internal methods on DataStream.

wuchong: Besides, this doesn't handle the case of partitioned table? I mean, data will be skew for partitioned table if `task parallelism` and `bucket number` are  not divisible.

fluss-connectors/fluss-connector-flink/src/main/java/com/alibaba/fluss/connector/flink/sink/RowDataKeySelector.java
```diff
@@ -0,0 +1,44 @@
...
+import com.alibaba.fluss.connector.flink.row.FlinkAsFlussRow;
+import com.alibaba.fluss.row.encode.KeyEncoder;
+
+import org.apache.flink.api.java.functions.KeySelector;
+import org.apache.flink.table.data.RowData;
+
+/** {@link KeySelector} to get bucket id. */
+public class RowDataKeySelector implements KeySelector<RowData, Integer> {
+
+    private final FlussBucketingFunction flussBucketingFunction;
```
wuchong: We shouldn't bind to  `FlussBucketingFunction`, because this table might enable lake storage and uses Paimon bucketing.

fluss-connectors/fluss-connector-flink/src/main/java/com/alibaba/fluss/connector/flink/sink/FlinkTableSink.java
```diff
@@ -168,7 +180,10 @@ public SinkRuntimeProvider getSinkRuntimeProvider(Context context) {
                                 flussConfig,
                                 tableRowType,
                                 targetColumnIndexes,
-                                ignoreDelete)
+                                ignoreDelete,
+                                numBucket,
+                                KeyEncoder.of(toFlussRowType(tableRowType), bucketKeys, null),
```
wuchong: lake format shouldn't be null.

fluss-common/src/main/java/com/alibaba/fluss/bucketing/BucketingFunction.java
```diff
@@ -20,8 +20,10 @@
 
 import javax.annotation.Nullable;
 
+import java.io.Serializable;
+
 /** An interface to assign a bucket according to the bucket key byte array. */
-public interface BucketingFunction {
+public interface BucketingFunction extends Serializable {
```
wuchong: We can construct `BucketingFunction` and `KeyEncoder` during runtime, to avoid to make so many classes `Serializable`.

wuchong (changes requested):

fluss-connectors/fluss-connector-flink/src/main/java/com/alibaba/fluss/connector/flink/FlinkConnectorOptions.java
swuferhong: Done!

fluss-connectors/fluss-connector-flink/src/main/java/com/alibaba/fluss/connector/flink/sink/RowDataKeySelector.java
swuferhong: Done!

fluss-connectors/fluss-connector-flink/src/main/java/com/alibaba/fluss/connector/flink/sink/FlinkSink.java
swuferhong: Done!

fluss-connectors/fluss-connector-flink/src/main/java/com/alibaba/fluss/connector/flink/sink/FlinkTableSink.java
swuferhong: Done!

fluss-common/src/main/java/com/alibaba/fluss/bucketing/BucketingFunction.java
swuferhong: Done!

fluss-connectors/fluss-connector-flink/src/main/java/com/alibaba/fluss/connector/flink/sink/FlinkSink.java
swuferhong: Done!

fluss-connectors/fluss-connector-flink/src/main/java/com/alibaba/fluss/connector/flink/FlinkConnectorOptions.java
swuferhong: Done!

swuferhong:
@wuchong comments addressed.

fluss-flink/fluss-flink-common/src/main/java/com/alibaba/fluss/flink/sink/FlinkRowDataChannelComputer.java
```diff
@@ -0,0 +1,103 @@
...
+        // not divisible, then we need to include the partition name as part of the shuffle key.
+        // This approach can help avoid the possible data skew. For example, if bucket number is 3
+        // and task parallelism is 2, it is highly possible that data shuffle becomes uneven. For
+        // instance, in task1, it might have 'partition0-bucket0', 'partition1-bucket0',
+        // 'partition0-bucket2', and 'partition1-bucket2', whereas in task2, it would only have
+        // 'partition0-bucket1' and 'partition1-bucket1'. As partition number increases, this
+        // situation becomes even more severe.
+        this.combineShuffleWithPartitionName =
+                partitionGetter != null
+                        && (numBucket % numChannels != 0 || numChannels % numBucket != 0);
```
wuchong: I think the condition should be ` partitionGetter != null && (numBucket % numChannels != 0)`, that means, if there are partition keys, and each channel doesn't have the same number of buckets, we should add partition name in the shuffle key.

For example, in the case of `numBucket = 8, numChannels = 4`, shouldn't add partition name shuffling. But currently, the result is true. 

Besides, please add unit tests around this flag for this class.

fluss-flink/fluss-flink-common/src/main/java/com/alibaba/fluss/flink/sink/FlinkRowDataChannelComputer.java
```diff
@@ -0,0 +1,103 @@
...
+import org.apache.flink.table.data.RowData;
+
+import javax.annotation.Nullable;
+
+import java.util.List;
+
+import static com.alibaba.fluss.utils.Preconditions.checkNotNull;
+
+/** {@link ChannelComputer} for flink {@link RowData}. */
+public class FlinkRowDataChannelComputer implements ChannelComputer<RowData> {
```
wuchong, line 36: Add `private static final long serialVersionUID = 1L;` for `Serializable` classes.

fluss-flink/fluss-flink-common/src/test/java/com/alibaba/fluss/flink/sink/FlinkTableSinkITCase.java
```diff
@@ -156,24 +157,41 @@ void testAppendLog(boolean compressed) throws Exception {
...
+    void testAppendLogWithBucketKeyWithSinkBucketShuffle() throws Exception {
+        testAppendLogWithBucketKey(true);
+    }
+
+    @Test
+    void testAppendLogWithBucketKeyWithoutSinkBucketShuffle() throws Exception {
+        testAppendLogWithBucketKey(false);
+    }
+
+    private void testAppendLogWithBucketKey(boolean sinkBucketShuffle) throws Exception {
```
wuchong, line 170: Did you try `@ParameterizedTest`? It would be better to use `@ParameterizedTest` if possible.

swuferhong: I tried.... But I encountered the same problem as when I developed #351 (add parameterized test in `FlinkTableSinkITCase`) : When using `@ParameterizedTest`, the `CI on GitHub` will hang to timeout, but in local env tests will passed. I'll create an issue to trace this problem and look into the root cause later:  https://github.com/alibaba/fluss/issues/659

fluss-flink/fluss-flink-common/src/main/java/com/alibaba/fluss/flink/sink/FlinkRowDataChannelComputer.java
swuferhong: Done!

fluss-flink/fluss-flink-common/src/main/java/com/alibaba/fluss/flink/sink/FlinkRowDataChannelComputer.java
swuferhong: Done!

fluss-flink/fluss-flink-common/src/test/java/com/alibaba/fluss/flink/sink/FlinkTableSinkITCase.java
wuchong: Thanks, I think we need to investigate the reason, otherwise, we don't know whether to add `@ParameterizedTest` or not.

wuchong (approved):

Issue #505, opened 2025-02-27
Support for Arrow Flight SQL Protocol and ADBC


wuchong:
### Search before asking

- [x] I searched in the [issues](https://github.com/alibaba/fluss/issues) and found nothing similar.

### Motivation

Fluss is a real-time streaming storage system designed for analytical workloads, offering sub-second latency for both reads and writes. To further enhance its interoperability and performance in modern data ecosystems, it is essential to integrate Fluss with widely adopted protocols like **Arrow Flight SQL** and **ADBC (Arrow Database Connectivity)**.

- **Arrow Flight SQL** is an RPC framework that enables efficient transfer of Apache Arrow data over the network, providing a high-performance alternative to traditional JDBC/ODBC for SQL-based interactions
- **ADBC** is a standardized API for connecting Arrow-native clients with databases, engines, and storage systems. It abstracts over different database implementations, making it easier for users to interact with Fluss using Arrow-compatible tools.

By supporting these protocols, Fluss can provide seamless integration with analytics tools, BI platforms, and other Arrow-native systems, such as **DuckDB**, while leveraging columnar data formats for faster data transfer and processing. Fluss uses Arrow format as the underlying log format, so it can achieve end-to-end zero-copy.

By supporting these protocols, Fluss enables seamless integration with analytics tools, BI platforms, and Arrow-native systems like DuckDB. Since Fluss uses the Apache Arrow format as its underlying log format, it can achieve end-to-end zero-copy data transfer and processing. 

### Solution

A design proposal should be submitted and discussed. For reference, see the Doris proposal for supporting Flight SQL: [Apache Doris Issue #25514 ](https://github.com/apache/doris/issues/25514).

## Reference

[1]: Arrow Flight SQL Documentation: [https://arrow.apache.org/docs/format/FlightSql.html ](https://arrow.apache.org/docs/format/FlightSql.html)2
[2]: ADBC Overview: [https://arrow.apache.org/adbc/ ](https://arrow.apache.org/adbc/)5
[3]: Arrow Flight SQL and ADBC Introduction: https://voltrondata.com/blog/simplifying-database-connectivity-with-arrow-flight-sql-and-adbc

### Willingness to contribute

- [ ] I'm willing to submit a PR!

PR #610, opened 2025-03-14
Start kafka netty server in TabletServer

dao-jun:
### Purpose

Linked issue: close https://github.com/alibaba/fluss/issues/609

When TabletServer is starting and kafka is enabled, start kafka server on configured port.

### Tests

NONE


wuchong:
Can we reuse the existing `NettyServer` and simply open a new port for Kafka, similar to how we support multiple listeners in [PR #531](https://github.com/alibaba/fluss/pull/531)? This way, we can reuse the Netty `EventLoopGroup` and `RequestProcessorPool` resources. In the future, we could even explore making Kafka compatible with reusing the existing Fluss port.

dao-jun:
No, they have different binary protocol, and we cannot use a same sharable NettyServerHandler and `FrameDecoder`.
And we have to process `InFlightRequests` in every single channel in fluss-kafka.

I think thread isolation is better for two systems, one is a messaging system, and another one is mostly for random access. 
Thread isolation can make them do their job better.

There is no need to forcefully reuse some existing modules, according to different needs to choose different implementation is better. Avoiding their mutual influence while also providing scalability for each other

wuchong:
@dao-jun While the binary protocols differ, they can still be made compatible because both Fluss and Kafka use the first 2 bytes as the API key ID, and their ID ranges are isolated. Ideally, this would allow us to reuse the same network port, simplifying deployment and usage. However, we can explore this optimization later and focus on the simpler approach of using separate ports for now.

On the other hand, I believe we should aim to reuse resources (workers and threads) even when using different ports. Consolidating resources into a single pool makes the system easier to operate and scale, as it allows us to scale one unified factor instead of balancing multiple independent ones. Fluss is already a hybrid system that supports both messaging and KV (random access, as you mentioned), all managed using a single resource pool. Adding Kafka support is essentially just another messaging protocol and does not significantly increase the workload complexity.

Looking ahead, we plan to support additional protocols such as Flight SQL protocol (#505), PostgreSQL, and [MCP](https://modelcontextprotocol.io/introduction). If each protocol requires its own dedicated resources, it will become increasingly challenging to tune and balance the system. By consolidating resources now, we can future-proof the design and ensure scalability as we expand our protocol support.

dao-jun:
@wuchong I'd considered that reuse `RequestProcessorPool#workerPool`, but it can't.
You can check `RequestProcessor#run()`, it runs in a loop to take `RpcRequest` from a blocking queue, and it will block the thread. 
﻿
Fluss's RPC model is unconventional, it implements an executor in it's own way, I can understand it wants to execute one channel's requests in a same thread, but it makes the threadpool can't be reused.
﻿
Maybe init a `SingleThreadExecutor[]` and choose an executor by hashing the task key is better.
﻿
In the other hand, even netty providers `new NioEventLoopGroup(int nThreads, Executor executor)` method to construct an EventLoopGroup, but the point is maybe one channel's requests can be executed in different threads so it may comes thread safety issue.
﻿
If fluss's requests executes in `worker eventLoopGroup`, we can reuse the `worker eventLoopGroup`.
﻿
So, at this stage, even though I want to reuse the thread resource, but it's hard to do.

wuchong:
Hi @dao-jun 

> It runs in a loop to take `RpcRequest` from a blocking queue, and it will block the thread.

Typically, request processing is very fast and does not block the thread. If it does become a bottleneck, this would be a common issue within Fluss, and we should address the thread model accordingly. However, this is an orthogonal problem to Kafka integration and should not block our decision-making when supporting Kafka.


> Fluss's RPC model is unconventional, it implements an executor in it's own way, I can understand it wants to execute one channel's requests in a same thread, but it makes the threadpool can't be reused.
> Maybe init a SingleThreadExecutor[] and choose an executor by hashing the task key is better.

Could you clarify what you mean by "it makes the threadpool can't be reused"? Currently, all requests are hashed by the channel id to utilize all threads in the thread pool. 


> In the other hand, even netty providers new NioEventLoopGroup(int nThreads, Executor executor) method to construct an EventLoopGroup, but the point is maybe one channel's requests can be executed in different threads so it may comes thread safety issue.

From my understanding, Netty's Reactor model guarantees that each `Channel` is bound to a fixed `EventLoop` for its entire lifecycle. All events for the same `Channel` are processed by that `EventLoop`, ensuring sequential execution within the `EventLoop`. This guarantees that events for the same `Channel` are processed in order within the `EventLoop`, which will not be executed in different threads.

dao-jun:
@wuchong I pushed a commit to reuse the netty acceptor and worker eventloopgroup, do you think it is OK?

wuchong:
@dao-jun Thanks for the updates. I’ve refactored the code based on your commits. The changes include:

1. [#531](https://github.com/alibaba/fluss/pull/531) was merged this week, which supports multiple endpoints/listeners. To align with this new listener mechanism, I introduced `kafka.listener.names` to replace `kafka.port`, enabling seamless support for multiple Kafka endpoints with different security protocols.

2. Introduced `NetworkProtocolPlugin` to support wire protocols (e.g., ADBC) in a pluggable way. This eliminates the need for a dedicated `KafkaNettyServer`, reusing Fluss's `NettyServer` and its thread/processor pools.

3. The new plugin mechanism decouples `fluss-server` from `fluss-kafka`, enabling self-contained unit tests (UT) and integration tests (IT) within `fluss-kafka`.

4. Added `KafkaRequestITCase` to test sending `ApiVersions` requests using Kafka's `NetworkClient` to Fluss's `NettyServer` with the Kafka endpoint enabled.

5. Changed the default `kafka.database` to `kafka` for better alignment with Fluss's default database (`fluss`). Moved database creation from TabletServer to Coordinator to reduce ZooKeeper operations.

Could you take a look the change if there is any problems? @dao-jun

wuchong:
merging this first to unblocked security protocol support: #577

@dao-jun we can continue to discuss the implementation here if there is any problems.

Issue #616, opened 2025-03-17
KvFileHandler and KvSnapshotHandle cost too much memory in CoordinatorServer


swuferhong:
### Search before asking

- [x] I searched in the [issues](https://github.com/alibaba/fluss/issues) and found nothing similar.

### Fluss version

main (development)

### Please describe the bug 🐞

When I creating a partitioned kv table with a large number of buckets like `512`. I found that the CoordinatorServe heap memory kept increasing, eventually reaching around 15 GB. Further inspection of the memory usage revealed that most of it was occupied by KvFileHandler and KvSnapshotHandle.

![Image](https://github.com/user-attachments/assets/bdf6bc4b-6951-4799-9cf7-53886e9a41b7)

### Are you willing to submit a PR?

- [x] I'm willing to submit a PR!

Issue #528, opened 2025-03-05
Reduce CoordinatorServer memory footprint for kv snapsgot


luoyuxia:
### Search before asking

- [x] I searched in the [issues](https://github.com/alibaba/fluss/issues) and found nothing similar.

### Description

Currently, CoordinatorServer will maintain the kv snapshots in memory to delete unused kv snapshot file via reference count which require to maintail all kv files(represent by `KvFileHandle`) in memory which is memory cost. We can reduce  memory footprint by following ways:

- [ ] Replace `FsPath` with `String` in `KvFileHandle`
- [ ] Store the relative path instead of absolute path in `KvFileHandle`

### Willingness to contribute

- [ ] I'm willing to submit a PR!

PR #617, opened 2025-03-17
[server] Change the default value of 'coordinator.io-pool.size' from 1 to 10

swuferhong:
### Purpose

Linked issue: #616 

When I creating a partitioned kv table with a large number of buckets like 512. I found that the `CoordinatorServe` heap memory kept increasing, eventually reaching around 15 GB. Further inspection of the memory usage revealed that most of it was occupied by `KvFileHandler` and `KvSnapshotHandle`.

The main reason for the accumulation of `KvFileHandler` and `KvSnapshotHandle` is that the speed of deleting `remote snapshot files` is very slow (the remote storage is OSS). Additionally, since the current concurrency level for deletion threads is only `1`, the backlog becomes increasingly severe, eventually leading to frequent full GCs and even OOM. The preliminary solution is to increase the `'coordinator.io-pool.size'` to `10`, allowing 10 concurrent threads to handle the deletion tasks. From the test results, CPU usage increases by only about `3%`, but the accumulation of `KvFileHandler` and `KvSnapshotHandle` no longer occurs.

In the long term, we need to further optimize the memory usage of `KvFileHandler` and `KvSnapshotHandle`. Issue #528  will continue to track this.


fluss-common/src/main/java/com/alibaba/fluss/config/ConfigOptions.java
```diff
@@ -169,12 +169,12 @@ public class ConfigOptions {
     public static final ConfigOption<Integer> COORDINATOR_IO_POOL_SIZE =
             key("coordinator.io-pool.size")
                     .intType()
-                    .defaultValue(1)
+                    .defaultValue(10)
```
luoyuxia, line 229: As dicuss before, we'd better to use cached thread pool `ThreadPoolExecutor` instead of just change the defaut value to 10. In some case of remote filesystem is fast to delete files like hdfs, or low pressure of kv snapshot, 10 may be too many.

wuchong: Can we use `server.background.threads` here? Kafka also has a `background.threads` with 10 default.

swuferhong: Ok, done!

swuferhong:
@wuchong @luoyuxia comments addressed, pr ready.

wuchong:
Upon further review, I realized that all `server.*` configurations pertain to the TabletServer. Therefore, using `server.background.threads` in this context is misleading. The original configuration, `coordinator.io-pool.size`, is more appropriate and self-explanatory.

Additionally, I found that Flink uses a similar configuration, `cluster.io-pool.size`, which defaults to `4 * CPU cores` for both the Master and TaskExecutor. Given this, I suggest simply increasing the `coordinator.io-pool.size` to `10`. This adjustment is straightforward and aligns with common practices.

Sorry for the back-and-forth on this matter. @swuferhong

swuferhong:
> Upon further review, I realized that all `server.*` configurations pertain to the TabletServer. Therefore, using `server.background.threads` for Coordinator is misleading. The original configuration, `coordinator.io-pool.size`, is more appropriate and self-explanatory.
> 
> Additionally, I found that Flink uses a similar configuration, `cluster.io-pool.size`, which defaults to `4 * CPU cores` for both the Master and TaskExecutor. Given this, I suggest simply increasing the `coordinator.io-pool.size` to `10`. This adjustment is straightforward and aligns with common practices.
> 
> Sorry for the back-and-forth on this matter. @swuferhong



> Upon further review, I realized that all `server.*` configurations pertain to the TabletServer. Therefore, using `server.background.threads` for Coordinator is misleading. The original configuration, `coordinator.io-pool.size`, is more appropriate and self-explanatory.
> 
> Additionally, I found that Flink uses a similar configuration, `cluster.io-pool.size`, which defaults to `4 * CPU cores` for both the Master and TaskExecutor. Given this, I suggest simply increasing the `coordinator.io-pool.size` to `10`. This adjustment is straightforward and aligns with common practices.
> 
> Sorry for the back-and-forth on this matter. @swuferhong

Ok, I will change it back, but I feel that the `ioExecutor` can still be moved to be managed within the `CoordinatorServer`. I will keep this part of the code.

swuferhong:
@wuchong comments addressed.

fluss-server/src/main/java/com/alibaba/fluss/server/tablet/TabletServer.java
```diff
@@ -400,6 +400,13 @@ private static void validateConfigs(Configuration conf) {
                             ConfigOptions.TABLET_SERVER_ID.key()));
         }
 
+        if (conf.get(ConfigOptions.BACKGROUND_THREADS) < 1) {
```
wuchong, line 403: `COORDINATOR_IO_POOL_SIZE` ?

swuferhong: `TabletServer` use `BACKGROUND_THREADS ` as io param.  Here I just noticed that there was no such check before, so I added it

PR #637, opened 2025-03-19
add deserialization schema and FlussRecord

polyzos:
This PR introduces:
- [x] a new FlussRecord to be used from the FlussSource
- [x] a FlussDeserializationSchema to be used from the FlussSource
- [x] some helper utils

### DIFF
--- a/fluss-flink/fluss-flink-common/src/main/java/com/alibaba/fluss/flink/row/FlinkRowToPojo.java
+++ b/fluss-flink/fluss-flink-common/src/main/java/com/alibaba/fluss/flink/row/FlinkRowToPojo.java
@@ -19,6 +19,8 @@ package com.alibaba.fluss.flink.row;
 import org.apache.flink.types.Row;
 
 import java.lang.reflect.Field;
+import java.lang.reflect.Modifier;
+import java.util.Arrays;
 
 /**
  * Helper class for converting a Flink Row to a POJO using reflection
@@ -30,7 +32,16 @@ public class FlinkRowToPojo {
     public static <T> T convert(Row row, Class<T> pojoClass) {
         try {
             T pojo = pojoClass.getDeclaredConstructor().newInstance();
-            Field[] fields = pojoClass.getDeclaredFields();
+            Field[] allFields = pojoClass.getDeclaredFields();
+
+            // Filter out synthetic fields (like JaCoCo's $jacocoData)
+            Field[] fields =
+                    Arrays.stream(allFields)
+                            .filter(
+                                    field ->
+                                            !field.isSynthetic()
+                                                    && !Modifier.isStatic(field.getModifiers()))
+                            .toArray(Field[]::new);
 
             if (row.getArity() != fields.length) {
                 throw new IllegalArgumentException(
--- a/fluss-flink/fluss-flink-common/src/main/java/com/alibaba/fluss/flink/row/FlinkRowToPojo.java
+++ b/fluss-flink/fluss-flink-common/src/main/java/com/alibaba/fluss/flink/row/FlinkRowToPojo.java
@@ -34,7 +34,7 @@ public class FlinkRowToPojo {
             T pojo = pojoClass.getDeclaredConstructor().newInstance();
             Field[] allFields = pojoClass.getDeclaredFields();
 
-            // Filter out synthetic fields (like JaCoCo's $jacocoData)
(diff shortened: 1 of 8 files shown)


fluss-flink/fluss-flink-common/src/main/java/com/alibaba/fluss/flink/row/FlinkRowToPojo.java
```diff
@@ -0,0 +1,91 @@
...
+                        value = ((Number) value).byteValue();
+                    }
+                }
+
+                field.set(pojo, value);
+            }
+
+            return pojo;
+        } catch (Exception e) {
+            throw new RuntimeException("Error converting Flink Row to POJO", e);
```
MehulBatra: Can we consider pointing to the exact field where the exception for conversion is happening
`throw new RuntimeException("Error converting Flink Row to POJO at field '" + field.getName() + "'", e);`

fluss-flink/fluss-flink-common/src/main/java/com/alibaba/fluss/flink/row/FlussRecord.java
```diff
@@ -0,0 +1,85 @@
...
+
+    public FlussRecord(long offset, long timestamp, ChangeType changeType, Row row) {
+        this.offset = offset;
+        this.timestamp = timestamp;
+        this.changeType = changeType;
+        this.row = row;
+    }
+
+    /** The position of this record in the corresponding fluss table bucket. */
+    public long logOffset() {
```
MehulBatra: Did we intentionally name it to logOffset instead of getLogOffset? if so, wouldn't it affect consistency as we used getChangeType & getRow

fluss-flink/fluss-flink-common/src/main/java/com/alibaba/fluss/flink/row/FlussRecord.java
```diff
@@ -0,0 +1,85 @@
...
+        return offset;
+    }
+
+    public long timestamp() {
```
MehulBatra: same as logOffset, getTimestamp

fluss-flink/fluss-flink-common/src/test/java/com/alibaba/fluss/flink/helper/Order.java
```diff
@@ -0,0 +1,80 @@
...
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+
+package com.alibaba.fluss.flink.helper;
+
+public class Order {
```
MehulBatra, line 19: Can we do it `public class Order implements Serializable`, as Flink is distributed, and adding this will make object transmission more robust?

polyzos: this is just for testing

fluss-flink/fluss-flink-common/src/test/java/com/alibaba/fluss/flink/row/FlussRecordTest.java
```diff
@@ -0,0 +1,88 @@
...
+
+        Row row2 = new Row(4);
+        row2.setField(0, 1L);
+        row2.setField(1, 100L);
+        row2.setField(2, 5);
+        row2.setField(3, "123 Test St");
+
+        // Create two identical FlussRecords
+        FlussRecord flussRecord1 =
+                new FlussRecord(0, System.currentTimeMillis(), ChangeType.INSERT, row1);
```
MehulBatra: System.currentTimeMillis() could vary, can we declare a fixed timestamp value instead, which we can use?

polyzos: i have addressed this in the latest commit

polyzos:
@wuchong After some more investigation I came to realize that maybe we are good with `ScanRecord` and `FlussRecord` is needed... seems like it introduces more complexity than actually solving a problem and since this PR is mainly around the FlussRercord and handling conversions with FlinkRows, I'm gonna close this.

PR #640, opened 2025-03-20
[Lake/Paimon] Create datalake enabled table should also create in lake

### LINKED ISSUE #430: Synchronously create Lake tables when Flusss lakehouse enabled during Fluss table creation
### Search before asking

### Description

Synchronously create Lake tables when Flusss lakehouse enabled during Fluss table creation, Supporting multiple Lake formats via interface abstraction

### Willingness to contribute

luoyuxia:
### Purpose

Linked issue: close #430

### Brief change log
1. Introduce `LakeStoragePluginSetUp` that load the `LakeStoragePlugin` by datalake format
2. Introduce `LakeCatalog` to create table in lake 
3. When create table with lake enabeld, create the table in lake via `LakeCatalog` 

### Tests

LakeEnabledTableCreateITCase


fluss-dist/src/main/assemblies/plugins.xml
```diff
@@ -25,6 +25,21 @@
...
+    <dependencySets>
+        <dependencySet>
+            <outputDirectory>plugins/paimon</outputDirectory>
+            <unpack>false</unpack>
+            <useProjectArtifact>false</useProjectArtifact>
+            <useProjectAttachments>false</useProjectAttachments>
+            <useTransitiveDependencies>true</useTransitiveDependencies>
+            <useTransitiveFiltering>true</useTransitiveFiltering>
+            <includes>
+                <include>org.apache.flink:flink-shaded-hadoop-2-uber</include>
```
luoyuxia, line 37: Paimon requires hadoop bundled, soe we include it in paimon plugin dir
https://paimon.apache.org/docs/master/flink/quick-start/

wuchong:
Is it ready to review? @luoyuxia

fluss-lake/fluss-lake-format-paimon/src/main/java/com/alibaba/fluss/lake/paimon/PaimonLakeCatalog.java
```diff
@@ -0,0 +1,145 @@
...
+        for (com.alibaba.fluss.metadata.Schema.Column column :
+                tableDescriptor.getSchema().getColumns()) {
+            schemaBuilder.column(
+                    column.getName(),
+                    column.getDataType().accept(FlussDataTypeToPaimonDataType.INSTANCE),
+                    column.getComment().orElse(null));
+        }
+
+        // set pk
+        if (tableDescriptor.getSchema().getPrimaryKey().isPresent()) {
```
luoyuxia: At first, I introduce additional `offset` and `timestamp` coumns, that's to enabled Fluss to subscribe the data in lake from a given  `offset`  and `timestamp` via Fluss client. But now, I feel like we can remove these two additional `offset` and `timestamp` columns at least for now.
1. Now, we mainly focus on Flink read the historical data in paimon and real-time data in Fluss. The `offset` and `timestamp` columns is not used. Introduce these columns may bring unnecassary complexity in early stage

2. subscribe via `offset`  and `timestamp` columns only works for log table and only works for paimon with bucket-num specified. But in paimon, it's recommend not to set bucket-num. So, `offset`  and `timestamp` columns become useless in most cases.

Still, we keep the possibility to support to subscribe the data in lake from a given  `offset`  and `timestamp` in the future. We can then introduce a option to enabled this feature for lake table.

luoyuxia:
> Is it ready to review? @luoyuxia

Yes, now, it's ready to review.

luoyuxia:
@wuchong @leonardBang Could you please help review?

fluss-lake/fluss-lake-format-paimon/src/main/java/com/alibaba/fluss/lake/paimon/PaimonLakeCatalog.java
luoyuxia: After discuss,  still keep the ability to subscribe via offset/timestamp, so, let's introduce another column `bucket` to help us to subscribe via bucket + offset.

fluss-common/src/main/java/com/alibaba/fluss/lakehouse/lakestorage/LakeCatalog.java
```diff
@@ -0,0 +1,40 @@
...
+ */
+
+package com.alibaba.fluss.lakehouse.lakestorage;
+
+import com.alibaba.fluss.exception.TableAlreadyExistException;
+import com.alibaba.fluss.metadata.TableDescriptor;
+import com.alibaba.fluss.metadata.TablePath;
+
+/** A catalog interface to modify metadata in external datalake. */
+public interface LakeCatalog extends AutoCloseable {
```
leonardBang, line 30: It deserves a `PublicEvolving` annotation as all datalake formats implementation will follow this interface?

fluss-lake/fluss-lake-format-paimon/src/main/java/com/alibaba/fluss/lake/paimon/PaimonLakeCatalog.java
```diff
@@ -0,0 +1,156 @@
...
+
+import java.util.List;
+
+/** A Paimon implementation of {@link LakeCatalog}. */
+public class PaimonLakeCatalog implements LakeCatalog {
+
+    public static final String OFFSET_COLUMN_NAME = "__offset";
+    public static final String TIMESTAMP_COLUMN_NAME = "__timestamp";
+    public static final String BUCKET_COLUMN_NAME = "__bucket";
+
```
leonardBang: For system metadata column, could we expose system metadata column configuration for users to avoid potential column conflict ?

fluss-lake/fluss-lake-format-paimon/src/main/java/com/alibaba/fluss/lake/paimon/PaimonLakeCatalog.java
```diff
@@ -0,0 +1,156 @@
...
+        // set custom properties to paimon schema
+        tableDescriptor.getCustomProperties().forEach(options::set);
+        schemaBuilder.options(options.toMap());
+        return schemaBuilder.build();
+    }
+
+    @Override
+    public void close() throws Exception {
+        paimonCatalog.close();
```
leonardBang: minor: check null before close in case  catalog instantiation fail

fluss-server/src/main/java/com/alibaba/fluss/server/coordinator/CoordinatorService.java
```diff
@@ -179,6 +189,22 @@ public CompletableFuture<CreateTableResponse> createTable(CreateTableRequest req
...
+        // before create table in fluss, we may create in lake
+        if (isDataLakeEnabled(tableDescriptor)) {
+            try {
+                getLakeCatalog().createTable(tablePath, tableDescriptor);
+            } catch (TableAlreadyExistException e) {
+                throw new TableAlreadyExistException(
+                        String.format(
+                                "The table %s already exists in %s catalog, please "
+                                        + "first drop the table in %s catalog.",
+                                tablePath, dataLakeFormat, dataLakeFormat));
```
leonardBang, line 250: Both `drop existed table` and `suggest a new table name` makes sense in this case, the later should be better?

leonardBang (approved):
Thanks @luoyuxia for the contribution, the PR looks generally to me, only left some minor comments

fluss-test-coverage/pom.xml
```diff
@@ -347,6 +347,7 @@
                                         <!-- end exclude for lakehouse-paimon -->
                                         <exclude>com.alibaba.fluss.lakehouse.cli.*</exclude>
                                         <exclude>com.alibaba.fluss.kafka.*</exclude>
+                                        <exclde>com.alibaba.fluss.lake.paimon.FlussDataTypeToPaimonDataType</exclde>
```
leonardBang, line 354: Adding a full types test in `LakeEnabledTableCreateITCase` is better?

fluss-lake/fluss-lake-format-paimon/src/main/java/com/alibaba/fluss/lake/paimon/PaimonLakeCatalog.java
luoyuxia: Currently, these system metadata columns are used for fluss client to subcribe from a given timetime/offset.  I hope it to be fixed now since most system's metadata column is fixed and fixed columns make it easy to understand.
I think we can make it congiurable in the future if we does found it help. It's a compatible change.

fluss-server/src/main/java/com/alibaba/fluss/server/coordinator/CoordinatorService.java
luoyuxia: I suggest both of them in the message.

fluss-test-coverage/pom.xml
luoyuxia: I add full types of fluss, but we have still to keep it in here since Fluss doesn't support array, map, row type. So the maxinum line coverage can only reach 65%.

leonardBang: makes sense to me

fluss-lake/fluss-lake-format-paimon/src/main/java/com/alibaba/fluss/lake/paimon/PaimonLakeCatalog.java
```diff
@@ -0,0 +1,158 @@
...
+        // set pk
+        if (tableDescriptor.getSchema().getPrimaryKey().isPresent()) {
+            schemaBuilder.primaryKey(
+                    tableDescriptor.getSchema().getPrimaryKey().get().getColumnNames());
+            options.set(
+                    CoreOptions.CHANGELOG_PRODUCER.key(),
+                    CoreOptions.ChangelogProducer.INPUT.toString());
+        } else {
+            // for log table, need to set bucket, offset and timestamp
+            schemaBuilder.column(BUCKET_COLUMN_NAME, DataTypes.INT());
```
leonardBang: I mean we need to  check the original schema contains same system column name like `__bucket` or not, to avoid conflict with users' original columns.

fluss-lake/fluss-lake-format-paimon/src/test/java/com/alibaba/fluss/lake/paimon/LakeEnabledTableCreateITCase.java
```diff
@@ -0,0 +1,406 @@
...
+        Options options = Options.fromMap(paimonTable.options());
+        assertThat(options.get(CoreOptions.BUCKET))
+                .isEqualTo(
+                        expectedBucketKey == null
+                                ? CoreOptions.BUCKET.defaultValue().intValue()
+                                : bucketNum);
+        assertThat(options.get(CoreOptions.BUCKET_KEY)).isEqualTo(expectedBucketKey);
+
+        // check custom properties
+        assertThat(paimonTable.options()).containsAllEntriesOf(expectedCustomProperties);
```
wuchong: check table properties are also included in paimon options.

fluss-lake/fluss-lake-format-paimon/src/main/java/com/alibaba/fluss/lake/paimon/PaimonLakeCatalog.java
```diff
@@ -0,0 +1,177 @@
...
+            options.set(
+                    CoreOptions.CHANGELOG_PRODUCER.key(),
+                    CoreOptions.ChangelogProducer.INPUT.toString());
+        }
+        // set partition keys
+        schemaBuilder.partitionKeys(tableDescriptor.getPartitionKeys());
+
+        // set custom properties to paimon schema
+        tableDescriptor.getCustomProperties().forEach(options::set);
+        schemaBuilder.options(options.toMap());
```
wuchong: 1. we should also add table properties into the paimon options, not only the custom properties.
2. we should add `fluss.` prefix to the properties, to avoid conflicts with paimon, and can better distinguish the fluss configs.

fluss-server/src/main/java/com/alibaba/fluss/server/coordinator/CoordinatorService.java
```diff
@@ -91,19 +96,24 @@ public final class CoordinatorService extends RpcServiceBase implements Coordina
 
     // null if the cluster hasn't configured datalake format
     private final @Nullable DataLakeFormat dataLakeFormat;
+    private final @Nullable LakeStorage lakeStorage;
+
+    private volatile @Nullable LakeCatalog lakeCatalog;
```
wuchong: There are too many `@Nullable`s. To simplify this, we can initialize `lakeCatalog` at the beginning since it is must to be used. Additionally, the `LakeStorage` object is unnecessary in this context because we do not create a tiering service in the coordinator. 

We should also ensure that `DataLakeFormat` and `LakeCatalog` are either both null or both non-null at the same time. Currently, the constructor does not enforce this consistency, which could lead to potential issues.

Besides, we should `close` the `LakeCatalog` in `shutdown()` to release potential resources.

fluss-lake/fluss-lake-format-paimon/src/test/java/com/alibaba/fluss/lake/paimon/LakeEnabledTableCreateITCase.java
```diff
@@ -0,0 +1,406 @@
...
+            TableDescriptor flussTable,
+            RowType expectedRowType,
+            @Nullable String expectedBucketKey,
+            int bucketNum,
+            Map<String, String> expectedCustomProperties) {
+        // check pk
+        if (!paimonTable.primaryKeys().isEmpty()) {
+            assertThat(paimonTable.primaryKeys())
+                    .isEqualTo(flussTable.getSchema().getPrimaryKey().get().getColumnNames());
+        }
```
wuchong:
```suggestion
        if (!flussTable.hasPrimaryKey()) {
            assertThat(paimonTable.primaryKeys()).isEmpty();
        } else {
            assertThat(paimonTable.primaryKeys())
                    .isEqualTo(flussTable.getSchema().getPrimaryKey().get().getColumnNames());
        }
```

Check no pk as well.

fluss-common/src/main/java/com/alibaba/fluss/lakehouse/lakestorage/LakeStoragePluginSetUp.java
```diff
@@ -0,0 +1,72 @@
...
+
+    private static Iterator<LakeStoragePlugin> getAllLakeStoragePlugins(
+            @Nullable PluginManager pluginManager) {
+        final Iterator<LakeStoragePlugin> pluginIteratorSPI =
+                ServiceLoader.load(LakeStoragePlugin.class).iterator();
+        final Iterator<LakeStoragePlugin> iteratorPlugins =
+                pluginManager == null
+                        ? pluginIteratorSPI
+                        : pluginManager.load(LakeStoragePlugin.class);
+        return Iterators.concat(iteratorPlugins, pluginIteratorSPI);
```
wuchong:
```suggestion
        if (pluginManager == null) {
            return pluginIteratorSPI;
        } else {
            return Iterators.concat(pluginIteratorSPI, pluginManager.load(LakeStoragePlugin.class));
        }
```

Currentlly, it returns `pluginIteratorSPI` twice if `pluginManager` is null.

fluss-lake/fluss-lake-format-paimon/src/main/java/com/alibaba/fluss/lake/paimon/PaimonLakeCatalog.java
```diff
@@ -0,0 +1,177 @@
...
+            // we use timestamp_ltz type
+            systemColumns.put(TIMESTAMP_COLUMN_NAME, DataTypes.TIMESTAMP_WITH_LOCAL_TIME_ZONE());
+        }
+
+        // set schema
+        for (com.alibaba.fluss.metadata.Schema.Column column :
+                tableDescriptor.getSchema().getColumns()) {
+            String columnName = column.getName();
+            if (systemColumns.containsKey(columnName)) {
+                throw new InvalidTableException(
```
wuchong: I created an issue https://github.com/alibaba/fluss/issues/810 to avoid creating table with system columns even if the table is not enabled lake.

wuchong (commented):
@luoyuxia the pull request looks good in general. I left some minor comments.

wuchong:
Besides, could you rename the module `fluss-lake-format-paimon` to `fluss-lake-paimon`? Because we will introduce the tiering service in `fluss-flink` instead of `fluss-lake/`. So `fluss-lake/` may only contain lake formats (e.g., `fluss-lake-iceberg`, `fluss-lake-delta`)

fluss-lake/fluss-lake-format-paimon/src/main/java/com/alibaba/fluss/lake/paimon/PaimonLakeCatalog.java
luoyuxia: Sometime we may want to add the options recognized by paimon into the paimon options, such as using parquet or orc format for the paimon table. Always add `fluss.`  prefix to the properties cause no way to set the options recognized by paimon. But we think in this pr, we can awalys add `fluss.`  prefix, and consider set options recognized by paimon in #836

luoyuxia:
@wuchong Comments addressed

fluss-lake/fluss-lake-format-paimon/src/main/java/com/alibaba/fluss/lake/paimon/PaimonLakeCatalog.java
wuchong: yes, we need a whitelist or a pattern.

wuchong (approved):

PR #643, opened 2025-03-20
[Flink] introduce flusssource for flink datastream api

polyzos:
- [x] Rename FlinkSource to FlussSource
- [x] Add deserialization schema
- [x] Introduce Builder Pattern
- [x] Update classes to use Generic and not be typed to RowData   
- [x] Introduce RowConverter Utils
- [x] Support Projection Pushdown

**Usage:**
```
   FlussSource<Order> flussSource =
                FlussSource.<Order>builder()
                        .setBootstrapServers(bootstrapServers)
                        .setDatabase(DEFAULT_DB)
                        .setTable(pkTableName)
                        .setStartingOffsets(OffsetsInitializer.earliest())
                        .setScanPartitionDiscoveryIntervalMs(1000L)
                        .setDeserializationSchema(new OrderDeserializationSchema())
                        .build();

        DataStreamSource<Order> stream =
                env.fromSource(flussSource, WatermarkStrategy.noWatermarks(), "Fluss Source");
```

Currently, it supports setting some key properties via the builder pattern, but to avoid overwhelming users they can set the rest with `Configuration` via the `setFlussConfig()`. We can always add more if required and after feedback, I just thought it's better like this for now.


polyzos:
Currently it supports reading from Fluss tables with projection pushdown.
If there are any missing features, like `limit` for batch scenarios, I would like to introduce them to a separate PR to avoid complicating things and have a working api.  @wuchong

fluss-flink/fluss-flink-common/src/main/java/com/alibaba/fluss/flink/source/FlussSource.java
```diff
@@ -0,0 +1,304 @@
...
+        if (builder.flussConfig == null) {
+            this.flussConf = new Configuration();
+        } else {
+            this.flussConf = builder.flussConfig;
+        }
+
+        tablePath = new TablePath(builder.database, builder.tableName);
+        flussConf.setString("bootstrap.servers", bootstrapServers);
+
+        Connection connection = ConnectionFactory.createConnection(flussConf);
```
MehulBatra: Can we Implement the acquiring resources, that can be released immediately after use?

```
TableInfo tableInfo;
try (
    Connection connection = ConnectionFactory.createConnection(flussConf);
    Admin admin = connection.getAdmin()
) {
    tableInfo = admin.getTableInfo(tablePath).get();
} 
```

fluss-flink/fluss-flink-common/src/main/java/com/alibaba/fluss/flink/source/FlussSource.java
```diff
@@ -0,0 +1,304 @@
...
+        Admin admin = connection.getAdmin();
+        TableInfo tableInfo;
+        try {
+            tableInfo = admin.getTableInfo(tablePath).get();
+        } catch (InterruptedException | ExecutionException e) {
```
MehulBatra: Also, can  we preserve the interrupted status when handling InterruptedException, 
Task cancellation might be dependent on interrupt status to trigger clean shutdown I am assuming.

```
catch (InterruptedException e) {
    Thread.currentThread().interrupt(); 
    throw new RuntimeException("Interrupted while getting table info for " + tablePath, e);
} catch (ExecutionException e) {
    throw new RuntimeException("Failed to get table info for " + tablePath, e);
}
```

fluss-flink/fluss-flink-common/src/main/java/com/alibaba/fluss/flink/source/FlussSource.java
```diff
@@ -0,0 +1,304 @@
...
+    private final Configuration flussConf;
+    private final TablePath tablePath;
+    private final boolean hasPrimaryKey;
+    private final boolean isPartitioned;
+    private final RowType sourceOutputType;
+    private final int[] projectedFields;
+    private final long scanPartitionDiscoveryIntervalMs;
+    private final boolean streaming;
+
+    private String bootstrapServers;
```
MehulBatra, line 74: Any specific reason we did not keep them final, I believe  it would be more robust to keep them final for thread safety.

polyzos: currently, there are two different constructors, one using the builder for the datastream that initializes bootstrap servers, while the other maintains compatibility with the current SQL interface and doesn't require that.. thus couldn't be set to final.

wuchong:
closing this as it will be broken into smaller PRs: https://github.com/alibaba/fluss/issues/569

PR #653, opened 2025-03-24
rename flink source/sink to fluss source/sink

polyzos:


wuchong:
How about renmaing `FlinkTableSink/Source` into `FlussTableSink/Source` as well? It's strange that a `FlinkTableSource` transforms into `FlussSource`.

polyzos:
@wuchong done

wuchong:
@polyzos It seems we still have many `FlinkXxx` classes that haven’t been renamed, and the renaming process has become a significant effort. How about introducing new public API classes, `FlussSource` and `FlussSink` (extends to the `FlinkSource` and `FlinkSink`), while keeping the original `FlinkSource` and `FlinkSink` as package-private? This approach avoids extensive refactoring and allows us to continue using the `Flink*` prefix for easier class identification when adding Spark/Trino connectors.

polyzos:
@wuchong sounds good . I will close this and open a new one.

PR #654, opened 2025-03-24
[server/kv] Fix KvTablet can't restore when a snapshot is discarded

### LINKED ISSUE #543: KvTablet may fail to restore when replica factor is 1
### Search before asking

### Fluss version

main (development)

### Please describe the bug 🐞

- Set replica factor to 1 for primary key table
- Kill one server, and start the server again
Then, it may throw fail to init KvTablet exception

### Solution

When kill one server, and start the server again, the coordinator will send notify leader to tablet server for twice(note twice is only happen when replica factor is 1) . Then, the following steps will be done:

1. In the first recieving notify leader request, it'll download snapshot files from remote. 
2. Then do snapshot and commit snapshot to coordinator server
3. but the leader epoch is smaller than the current leader epoch, after server failover, the leader epoch will be bumped, so coordinator server refuse this snapshot commit
4. then tablet server discard the snapshot, but it'll discard all the sst files in this snapshot including the sst files in previous snapshot..
5. In the second recieving notify leader request, it'll download snapshot files again, but the files have been deleted

The bug happens in Step 4, we shouldn't discard the sst files in previous snapshot.

luoyuxia:
### Purpose

Linked issue: close #543

To fix the issue that  KvTablet can't restore when a snapshot is discarded

### Brief change log

When the `KvFileHandler` is instance of `PlaceholderKvFileHandler`, shoudn't delete the file in method `discard`.  Flink Flink either doesn't delete for `PlaceholderState`.

### Tests

The added code in `RocksIncrementalSnapshotTest#testIncrementalSnapshot` will try to discard new snapshot with `PlaceholderKvFileHandler` and restore from previous snapshot.

### DIFF
--- a/fluss-common/src/main/java/com/alibaba/fluss/fs/utils/FileDownloadUtils.java
+++ b/fluss-common/src/main/java/com/alibaba/fluss/fs/utils/FileDownloadUtils.java
@@ -69,7 +69,7 @@ public class FileDownloadUtils {
                     transferDataToDirectoryAsync(fileDownloadSpecs, internalCloser, executorService)
                             .collect(Collectors.toList());
             // Wait until either all futures completed successfully or one failed exceptionally.
-            FutureUtils.completeAll(futures, new DownloadProgressAction(fileDownloadSpecs.size()))
+            FutureUtils.waitForAll(futures, new DownloadProgressAction(fileDownloadSpecs.size()))
                     .get();
         } catch (Exception e) {
             fileDownloadSpecs.stream()
--- a/fluss-common/src/main/java/com/alibaba/fluss/fs/utils/FileDownloadUtils.java
+++ b/fluss-common/src/main/java/com/alibaba/fluss/fs/utils/FileDownloadUtils.java
@@ -37,6 +37,7 @@ import java.io.IOException;
 import java.io.OutputStream;
 import java.nio.file.Files;
 import java.nio.file.Path;
+import java.util.ArrayList;
 import java.util.Collection;
 import java.util.List;
 import java.util.concurrent.CompletableFuture;
@@ -64,14 +65,16 @@ public class FileDownloadUtils {
         CloseableRegistry internalCloser = new CloseableRegistry();
         // Make sure we also react to external close signals.
         closeableRegistry.registerCloseable(internalCloser);
+        List<CompletableFuture<Long>> futures = new ArrayList<>();
         try {
-            List<CompletableFuture<Long>> futures =
+            futures =
                     transferDataToDirectoryAsync(fileDownloadSpecs, internalCloser, executorService)
                             .collect(Collectors.toList());
             // Wait until either all futures completed successfully or one failed exceptionally.
             FutureUtils.waitForAll(futures, new DownloadProgressAction(fileDownloadSpecs.size()))
                     .get();
         } catch (Exception e) {
+            futures.forEach(future -> future.cancel(true));
             fileDownloadSpecs.stream()
                     .map(FileDownloadSpec::getTargetDirectory)
                     .map(Path::toFile)
--- a/fluss-server/src/main/java/com/alibaba/fluss/server/kv/snapshot/KvTabletSnapshotTarget.java
+++ b/fluss-server/src/main/java/com/alibaba/fluss/server/kv/snapshot/KvTabletSnapshotTarget.java
@@ -18,7 +18,6 @@ package com.alibaba.fluss.server.kv.snapshot;
 
 import com.alibaba.fluss.annotation.VisibleForTesting;
 import com.alibaba.fluss.config.ConfigOptions;
-import com.alibaba.fluss.exception.FlussException;
 import com.alibaba.fluss.fs.FileSystem;
 import com.alibaba.fluss.fs.FsPath;
 import com.alibaba.fluss.metadata.TableBucket;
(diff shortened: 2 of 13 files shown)


fluss-common/src/main/java/com/alibaba/fluss/fs/utils/FileDownloadUtils.java
```diff
@@ -69,7 +69,7 @@ public static void transferAllDataToDirectory(
                     transferDataToDirectoryAsync(fileDownloadSpecs, internalCloser, executorService)
                             .collect(Collectors.toList());
             // Wait until either all futures completed successfully or one failed exceptionally.
-            FutureUtils.completeAll(futures, new DownloadProgressAction(fileDownloadSpecs.size()))
+            FutureUtils.waitForAll(futures, new DownloadProgressAction(fileDownloadSpecs.size()))
```
luoyuxia, line 74: I'd like to change to `waitForAll` which will return when any one exception is thrown.

fluss-server/src/main/java/com/alibaba/fluss/server/kv/snapshot/KvTabletSnapshotTarget.java
```diff
@@ -219,14 +219,9 @@ public void handleSnapshotResult(
...
-            handleSnapshotFailure(
-                    snapshotId,
-                    snapshotLocation,
-                    new FlussException(
-                            String.format(
-                                    "Fail to add the snapshot %s to completed snapshot store.",
-                                    completedSnapshot),
-                            t));
+            handleSnapshotFailure(snapshotId, snapshotLocation, t);
+            // throw the exception to make PeriodicSnapshotManager can catch the exception
```
luoyuxia, line 223: Throw the exception so that `PeriodicSnapshotManager` can warn `Fail to handle snapshot result during snapshot..`

luoyuxia:

fluss-common/src/main/java/com/alibaba/fluss/fs/utils/FileDownloadUtils.java
wuchong: Should we cancel all the futures if any exception thrown here? Otherwise, it may has conflict for deleting local file and downloading.

wuchong (commented):
we should also bump the leader epoch before recovery kv tablet to fix the mentioned problem in issue "but the leader epoch is smaller than the current leader epoch, after server failover, the leader epoch will be bumped, so coordinator server refuse this snapshot commit"

wuchong:
It seems the new fix leads to the failure of `StopReplicaITCase.testStopReplica`. Besides, could you add a test in `ReplicaTest` for the leader epoch change? Other than that, should we need to move `bucketEpoch = requestBucketEpoch` before the `onBecomeNewLeader()`?

luoyuxia:
> It seems the new fix leads to the failure of `StopReplicaITCase.testStopReplica`. Besides, could you add a test in `ReplicaTest` for the leader epoch change? Other than that, should we need to move `bucketEpoch = requestBucketEpoch` before the `onBecomeNewLeader()`?

I append a new commit to fix the test failure and add test to conver leader epoch change. I think we don't need to move `bucketEpoch = requestBucketEpoch` before the `onBecomeNewLeader()` currently since it won't involve `bucketEpoch` in `onBecomeNewLeader`.

fluss-server/src/test/java/com/alibaba/fluss/server/replica/ReplicaTest.java
```diff
@@ -101,7 +101,13 @@ void testMakeLeader() throws Exception {
...
         makeKvReplicaAsLeader(kvReplica);
         assertThat(kvReplica.getLogTablet()).isNotNull();
-        // TODO get kv is true.
+        assertThat(kvReplica.getKvTablet()).isNotNull();
+
+        // make as leader with leader epoch change
+        int expectedLeaderEpoch = 1;
+        makeKvReplicaAsLeader(kvReplica, expectedLeaderEpoch);
+        assertThat(kvReplica.getLeaderEpoch()).isEqualTo(expectedLeaderEpoch);
+        assertThat(kvReplica.getLogTablet()).isNotNull();
```
wuchong: Could you add a test that can reproduce the issue problem? Currently, it passes for old code.

fluss-server/src/test/java/com/alibaba/fluss/server/coordinator/StopReplicaITCase.java
```diff
@@ -156,9 +156,14 @@ private List<Path> assertReplicaExistAndGetTableOrPartitionDirs(
...
-                assertThat(replica.getKvTablet()).isNotNull();
-                assertThat(replica.getKvTablet().getKvTabletDir()).exists();
+                Replica kvReplica =
+                        FLUSS_CLUSTER_EXTENSION.waitAndGetLeaderReplica(replica.getTableBucket());
+                retry(
+                        Duration.ofMinutes(1),
+                        () -> {
+                            assertThat(kvReplica.getKvTablet()).isNotNull();
+                            assertThat(kvReplica.getKvTablet().getKvTabletDir()).exists();
+                        });
```
wuchong: This change looks strange to me. If the `kvReplica` is already leader, why the `kvtablet` is not ready? It means the leader flag in `Replica` is incorrect. 

Should we move `leaderReplicaIdOpt.set(localTabletServerId);` after `onBecomeNewLeader()`?

wuchong (approved):

Issue #644, opened 2025-03-21
Restart cluster may cause two leaders of a single bucket


luoyuxia:
### Search before asking

- [x] I searched in the [issues](https://github.com/alibaba/fluss/issues) and found nothing similar.

### Fluss version

0.6.0 (latest release)

### Please describe the bug 🐞

Restart the whole Fluss cluster, including coordiantor & tablet server, it may happens:
1. assume a `bucket{leader = 0, leaderEpoch=4, isr=[0, 1]}`
2. server 1 down, => won't adjust isr, then still is `{leader=0, leaderEpoch=4, isr=[0, 1]`
3. server 0 down =>  no any live isr, won't change leader, then {leader=0, leaderEpoch=4, isr=[0, 1], the bucket will be offline since the leader is offline
4. server 1 up
   4.1. Set the replica  of the bucket in server 1 to online, the server1 will try to become follower to fetch from server 0, but current, server 0 is not avaiable, so server 1 throws `Could not find leader 0 in server metadata by id for follower replica 1 while make follower for bucket `,
   4.2. Elect a new leader for the bucket, update to `LeaderAndIsr{leader=1, leaderEpoch=5, isr=[0, 1]}`, the bucket will be online
5. The coordinator recieve the exception thown by step 4.1, and the exception is from server1 which is current leader of the bucket, then bucket will be set to offline.. **But in step 4.2, server 1 has becomed the leader**
6. server 0 up => since the bucket is offline, coordinator will select 0 as leader, so updated to LeaderAndIsr{leader=0, leaderEpoch=6, isr=[0, 1]}, then notify server 0 it has becomed leader.. **So server 0 also become leader of the bucket**

Then, it'll have two leaders in the cluster.

### Solution

Not sure how kafka handles it..

### Are you willing to submit a PR?

- [ ] I'm willing to submit a PR!

wuchong:
4.1 and 4.2 is strange, in theory, `becomeLeaderOrFollower` is notified by the Coordinator. It seems Coordinator sends becomeFollower and becomeLeader events to server1 at the same time?

luoyuxia:
This issue should be fixed now after #662 , I'd like to add a ITcase to cover this case before close it.

libenchao:
@luoyuxia I'd like to take this issue, could you assign this to me?

wuchong:
@libenchao assigned to you

PR #656, opened 2025-03-24
[server] Use dynamic tablet server gateway for fetching log from leader 

### LINKED ISSUE #633: ReplicaFetcherManager#createFetcherThread create gateway should use MetadataCache#getTabletServer(int serverId) instead of using a static ServerNode
### Search before asking

### Fluss version

0.6.0 (latest release)

### Please describe the bug 🐞

`ReplicaFetcherManager#createFetcherThread` create `gateway` should use `MetadataCache#getTabletServer(int serverId)` instead of using a static `ServerNode`.  The reason is that when a `tabletServer` or `coordinatorServer` goes offline, the corresponding `IP address` and `port` may change. Using static `ServerNode` will prevent recovery in this situation.

### Solution

_No response_

luoyuxia:
### Purpose

Linked issue: close #633 

Now, the tablet server gateway to fetch log from leader is static, which will cause if the leader change/failover,  it will still fetch log from old ip address.
Assuming the case:
- server 0 is follower, server 1 is leader.  Coordinator server down.
- server 1 down
- server 1 up and ip is changed, but the server 0 will stil try to fetch from old ip address.

### Brief change log
`ReplicaFetcherThread` use dynamic tablet server gateway to fetch log from leader, which will got latest server node of the tablet server. 

### Tests

Exsting tests has covered it.

### DIFF
--- a/fluss-server/src/main/java/com/alibaba/fluss/server/replica/ReplicaManager.java
+++ b/fluss-server/src/main/java/com/alibaba/fluss/server/replica/ReplicaManager.java
@@ -17,7 +17,6 @@
 package com.alibaba.fluss.server.replica;
 
 import com.alibaba.fluss.annotation.VisibleForTesting;
-import com.alibaba.fluss.cluster.ServerNode;
 import com.alibaba.fluss.config.ConfigOptions;
 import com.alibaba.fluss.config.Configuration;
 import com.alibaba.fluss.exception.FencedLeaderEpochException;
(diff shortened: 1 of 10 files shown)


luoyuxia:
@wuchong @swuferhong Could you please help review it?

fluss-server/src/main/java/com/alibaba/fluss/server/replica/ReplicaManager.java
```diff
@@ -851,7 +857,7 @@ private void addFetcherForReplicas(
                     bucketAndStatus.put(
                             tb,
                             new InitialFetchStatus(
-                                    tb.getTableId(), leader.get(), logTablet.localLogEndOffset()));
+                                    tb.getTableId(), leaderId, logTablet.localLogEndOffset()));
```
wuchong: As we don't need to get the leader `ServerNode` here, is it still necessary to return a bucket level error to CoordinatorServer when the leader `ServerNode` is not in metadata?

fluss-server/src/main/java/com/alibaba/fluss/server/replica/fetcher/LeaderEndpoint.java
```diff
@@ -28,8 +27,8 @@
 /** Defines the interface to be used to access a tablet server that is a leader. */
 interface LeaderEndpoint {
 
-    /** The specific tablet server address we want to connect to. */
-    ServerNode leaderNode();
+    /** The tablet server id we want to connect to. */
+    int leaderNode();
```
wuchong: rename to `leaderServerId`

fluss-server/src/main/java/com/alibaba/fluss/server/replica/fetcher/RemoteLeaderEndpoint.java
```diff
@@ -41,7 +40,7 @@
 /** Facilitates fetches from a remote replica leader in one tablet server. */
 final class RemoteLeaderEndpoint implements LeaderEndpoint {
     private final int followerServerId;
-    private final ServerNode remoteNode;
+    private final int remoteNode;
```
wuchong: rename to `removeServerId`.

fluss-server/src/main/java/com/alibaba/fluss/server/replica/ReplicaManager.java
luoyuxia: Yes, I believe we don't need to return a bucket level error . I meant to remove bucket level error thrown in #644 since #644 is part of caused by the thrown error.  But I think we can remove bucket level error thrown in here, and along with #664,  #644 won't happen.

wuchong (approved):

PR #664, opened 2025-03-25
[coordinator] Kick off offline replica from bucket isr

luoyuxia:
### Purpose

Linked issue: close #xxx


fluss-client/src/test/java/com/alibaba/fluss/client/admin/ClientToServerITCaseBase.java
```diff
@@ -110,8 +110,6 @@ private static Configuration initConfig() {
         conf.setInt(ConfigOptions.DEFAULT_REPLICATION_FACTOR, 3);
         // set a shorter interval for testing purpose
         conf.set(ConfigOptions.KV_SNAPSHOT_INTERVAL, Duration.ofSeconds(1));
-        // set a shorter max lag time to make tests in FlussFailServerTableITCase faster
-        conf.set(ConfigOptions.LOG_REPLICA_MAX_LAG_TIME, Duration.ofSeconds(10));
```
luoyuxia, line 114: Now, we don't need to set to a shorter max lag time since offline replica will be removed from isr by coordinator after it knows the replica is offline.

luoyuxia:
@wuchong @swuferhong Could you please help review?

fluss-server/src/test/java/com/alibaba/fluss/server/replica/AdjustIsrITCase.java
```diff
@@ -63,7 +73,7 @@ void beforeEach() {
     }
 
     @Test
-    void testIsrShrinkAndExpend() throws Exception {
+    void testIsrShrinkAndExpand() throws Exception {
```
wuchong, line 80: Add some comment on this test method that this method tests Isr change of log lag.

fluss-server/src/main/java/com/alibaba/fluss/server/coordinator/statemachine/ReplicaStateMachine.java
```diff
@@ -206,19 +217,12 @@ private void doHandleStateChanges(
...
-                                        return;
-                                    }
-                                } else {
-                                    partitionName = null;
+                                try {
+                                    partitionName = getPartitionName(tableBucket);
+                                } catch (PartitionNotExistException e) {
+                                    LOG.error(e.getMessage());
+                                    logFailedSateChange(replica, currentState, targetState);
+                                    return;
```
wuchong, line 225: should be `continue`? Otherwise, some LeaderAndIsr changes may not be notified.

luoyuxia: Do you mean the LeaderAndIsr change for other `validReplicas`? If so, `return` will just skip the current replica, and the other valida replicas will still go through the loop.  
Also, user `continue` wiil make my IDE error the message `Continue outside of loop`

wuchong: Sorry, I mean this `return` https://github.com/alibaba/fluss/pull/664/files/33facf0e8b96a793ecb717a07affad2c855dcf4d#diff-83fa38b29066726ca71790bb02d2e2734d8ebfbe24c23806eb55a53d64f6c824R289

luoyuxia: Good catch! Make mistake in copy code..

wuchong (approved):

Issue #655, opened 2025-03-24
Sender maybe block forever if the number of inflightRequests for a certain tableBucket reaches maxInflightRequestsPerBucket and idempotence enabled


swuferhong:
### Search before asking

- [x] I searched in the [issues](https://github.com/alibaba/fluss/issues) and found nothing similar.

### Fluss version

0.6.0 (latest release)

### Please describe the bug 🐞

Currently, before sending `produceLog` request, the `sender `needs to first obtain batches that can be drained through the `RecordAccumulator#drain()` method. If there are no batches available to drain, it will skip this round of sending. Inside the drain method, there's a check `shouldStopDrainBatchesForBucket()` to determine whether the current draining process should be stopped. The core part of this check is `idempotenceManager.canSendMoreRequests(tableBucket)`. If the number of inflight requests of this bucket exceeds `maxInflightRequestsPerBucket`, the draining process is exited.

Consider the following scenario: when a `tabletServer` goes down, the number of inflight requests for a bucket can quickly reach `maxInflightRequestsPerBucket`. In this situation, the `sender` for this process will get block forever.

### Solution

Kafka also have a  option `MAX_IN_FLIGHT_REQUESTS_PER_CONNECTION` to control the max inflight request send to one server, the default value is `5` same as fluss `CLIENT_WRITER_MAX_INFLIGHT_REQUESTS_PER_BUCKET`.  However Kafka use this option as a network value. 

### Are you willing to submit a PR?

- [x] I'm willing to submit a PR!

PR #669, opened 2025-03-27
[client] Fix Sender maybe block forever error if inflightBatchSize in IdempotenceManager exceed maxInflightRequestsPerBucket

swuferhong:
### Purpose

Linked issue: #655 

This pr is aims to fix `Sender` maybe block forever error if `inflightBatchSize` in `IdempotenceManager` exceed `maxInflightRequestsPerBucket`.

In our current implementation, when the `inflightBatchSize` exceeds `maxInflightRequestsPerBucket`, no further requests can be sent. This is incorrect. We should allow the batch with the` smallest (first) batchSequence` to always be in a sendable state. This ensures that when the server can handle the retriable error, the client can recover.

### DIFF
--- a/fluss-client/src/main/java/com/alibaba/fluss/client/write/RecordAccumulator.java
+++ b/fluss-client/src/main/java/com/alibaba/fluss/client/write/RecordAccumulator.java
@@ -653,21 +653,33 @@ public final class RecordAccumulator {
                 return true;
             }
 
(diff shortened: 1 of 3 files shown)


swuferhong:
@wuchong @luoyuxia  pr ready, could you help review it? thx.

wuchong (approved):
Well done! LGTM.

Issue #671, opened 2025-03-27
LogManager load logs always assume it recovery from a clean shutdown even if the previous run was an abnormal exit


swuferhong:
### Search before asking

- [x] I searched in the [issues](https://github.com/alibaba/fluss/issues) and found nothing similar.

### Fluss version

main (development)

### Please describe the bug 🐞

Currently, `LogManager#loadLogs()` always assume it recovery from a clean shutdown even if the previous run was an abnormal exit. It will cause  `LogTablet.rebuildWriterState()`  recovers the incorrect `WriterState` because during a clean shutdown, `writerState` recovery always starts from the snapshots corresponding to the latest log. In the case of a unclean shutdown, it starts from scanning from the beginning of the log to recover the `WriterState`.

### Are you willing to submit a PR?

- [x] I'm willing to submit a PR!

wuchong:
Fixed in #672

PR #672, opened 2025-03-27
[server] Add cleanShutdown mark to indicate whether the LogManager's previous exit was a clean shutdown or an abnormal exit

swuferhong:
### Purpose

Linked issue: #671 

Add cleanShutdown mark to indicate whether the LogManager's previous exit was a clean shutdown or an abnormal exit. This pr is aims to fix the error show in #671


fluss-server/src/test/java/com/alibaba/fluss/server/log/LogManagerTest.java
```diff
@@ -209,6 +210,37 @@ void testRecoveryAfterLogManagerShutdown(String partitionName) throws Exception
         newLogManager.shutdown();
     }
 
+    @ParameterizedTest
+    @MethodSource("partitionProvider")
+    void testHasCleanShutdownMarkerAfterLogManagerShutdown(String partitionName) throws Exception {
```
wuchong, line 214: nit: I think we don't need to parameterized this test, as the partition doesn't affect the logic of shutdown flag.

swuferhong: Done!

wuchong (approved):

Issue #674, opened 2025-03-27
When a tabletServer recovers from an unclean shutdown, there may be issues with data loss


swuferhong:
### Search before asking

- [x] I searched in the [issues](https://github.com/alibaba/fluss/issues) and found nothing similar.

### Fluss version

main (development)

### Please describe the bug 🐞

Currently, the synchronization of fluss highWatermark involves first updating the leader and then updating the follower in the next round. Log recovery also depends on the highWatermark in the leader's local checkpoint cache to determine whether to truncate data. However, this approach poses a significant risk of data loss as the follower become leader, but the `HW` of follower lower than leader, it may cause a data truncate.

### Are you willing to submit a PR?

- [x] I'm willing to submit a PR!

Issue #673, opened 2025-03-27
Support leader epoch cache to avoid Inconsistent or lost data when the leader or follower goes offline abnormally


swuferhong:
### Search before asking

- [x] I searched in the [issues](https://github.com/alibaba/fluss/issues) and found nothing similar.

### Motivation

Currently, the synchronization of fluss `highWatermark` involves first updating the leader and then updating the follower in the next round. `Log recovery` also depends on the `highWatermark` in the leader's local checkpoint cache to determine whether to truncate data. However, this approach poses a significant risk of data loss. In issue xxx, we will modify the `highWatermark synchronization` way to update the follower first and then the leader in the next round. While this change can solve the data loss problem caused by server abnormal exits, it may introduce new problem of data duplication. Therefore, in this issue, we aim to adopt the approach from Kafka [KIP-101](https://cwiki.apache.org/confluence/display/KAFKA/KIP-101+-+Alter+Replication+Protocol+to+use+Leader+Epoch+rather+than+High+Watermark+for+Truncation) to resolve this problem through supporting leader epoch cache to avoid Inconsistent or lost data when the leader or follower goes offline abnormally.

### Willingness to contribute

- [ ] I'm willing to submit a PR!

PR #676, opened 2025-03-27
[server] Fix the potential data loss issues when a tabletServer recovers from an unclean shutdown

swuferhong:
### Purpose

Linked issue:  #674 

This pr is aims to fix the potential data loss issues when a `tabletServer` recovers from an unclean shutdown, the reason why there are a data loss issue is described in #674 .

Note that this is just an temporary solution, if we want to have a strong consistency guarantee, we should do as [KIP-101](https://cwiki.apache.org/confluence/display/KAFKA/KIP-101+-+Alter+Replication+Protocol+to+use+Leader+Epoch+rather+than+High+Watermark+for+Truncation) do, trace by:
#673


swuferhong:
@wuchong @luoyuxia  pr ready. Could you help review it? thx.

wuchong (approved):

Issue #685, opened 2025-03-28
Introduce batchDeliveryTimeout in client to expire batches that have been stuck for a long time


swuferhong:
### Search before asking

- [x] I searched in the [issues](https://github.com/alibaba/fluss/issues) and found nothing similar.

### Motivation

Introduce batchDeliveryTimeout in client to expire batches that have been stuck for a long time. The purpose of this approach is to enable our `Flink job` to recover the write operation by restarting the client via failover. Otherwise, the client may not be able to recover itself, especially when an unrecoverable out-of-order issue occurs due to the server going down unexpectedly.

### Willingness to contribute

- [x] I'm willing to submit a PR!

PR #686, opened 2025-03-28
[client] Introduce batchDeliveryTimeout in client to expire batches that have been stuck for a long time

swuferhong:
### Purpose

Linked issue: #685 

This pr is aims to introduce `batchDeliveryTimeout` in client to expire batches that have been stuck for a long time.  The purpose of this approach is to enable our Flink job can be recover the write operation by restarting the client via failover. Otherwise, the client may not be able to recover itself, especially when an unrecoverable out-of-order issue occurs due to the server going down unexpectedly.


fluss-common/src/main/java/com/alibaba/fluss/config/ConfigOptions.java
```diff
@@ -842,6 +842,23 @@ public class ConfigOptions {
                                     + "this timeout is the max time that delayed write try to complete. "
                                     + "The default setting is 30 seconds.");
 
+    public static final ConfigOption<Duration> CLIENT_WRITER_BATCH_DELIVERY_TIMEOUT =
+            key("client.writer.batch-delivery-timeout")
+                    .durationType()
+                    .defaultValue(Duration.ofMinutes(2))
+                    .withDescription(
+                            "An upper bound on the time to report success or failure after a call to "
+                                    + "<code>Sender#send()</code> returns. This limits the total time that a record "
```
wuchong, line 851: Sender is an internal class which is hard to understand for users here. In Kafka docs, the `send()` means `Producer#send(..)`. Therefore, we should use `AppendWriter#append(..)`, `UpsertWriter#upsert(..)` and `UpsertWriter#delete(..)`.

fluss-common/src/main/java/com/alibaba/fluss/config/ConfigOptions.java
```diff
@@ -842,6 +842,23 @@ public class ConfigOptions {
                                     + "this timeout is the max time that delayed write try to complete. "
                                     + "The default setting is 30 seconds.");
 
+    public static final ConfigOption<Duration> CLIENT_WRITER_BATCH_DELIVERY_TIMEOUT =
+            key("client.writer.batch-delivery-timeout")
```
wuchong, line 846: `client.writer.delivery-timeout`? From the perspective of users, the timeout is limiting for a single call of `append/upsert`. Batching is just an internal optimization.

fluss-common/src/main/java/com/alibaba/fluss/config/ConfigOptions.java
```diff
@@ -842,6 +842,23 @@ public class ConfigOptions {
...
+                                    + "<code>Sender#send()</code> returns. This limits the total time that a record "
+                                    + "will be delayed prior to sending, the time to await ack from the tabletServer"
+                                    + "(if expected), and the time allowed either an unrecoverable error is encountered, "
+                                    + "the retries have been exhausted, or the record is added to a batch which reached "
+                                    + "an earlier delivery expiration deadline. Thw vale of this config should be greater "
+                                    + "than or equal to the sum of '"
+                                    + CLIENT_REQUEST_TIMEOUT.key()
+                                    + "' and '"
+                                    + CLIENT_WRITER_BATCH_TIMEOUT.key()
+                                    + "'.");
```
wuchong, line 860:
```suggestion
                               "An upper bound on the time to report success or failure after a call to "
                                    + "<code>AppendWriter#append()</code>, <code>UpsertWriter#upsert()</code> and <code>UpsertWriter#delete()</code> returns. "
                                    + "This limits the total time that a record will be delayed prior to sending, the time to await ack from the TabletServer "
                                    + "(if expected), and the time allowed for retries send failures. The writer may report failure to send a record earlier "
                                    + "than this config if either an unrecoverable error is encountered, the retries have been exhausted, or the record is "
                                    + "added to a batch which reached an earlier delivery expiration deadline. The vale of this config should be greater "
                                    + "than or equal to the sum of '"
                                    + CLIENT_REQUEST_TIMEOUT.key()
                                    + "' and '"
                                    + CLIENT_WRITER_BATCH_TIMEOUT.key()
                                    + "'.");
```

Update the docs as well.

fluss-client/src/main/java/com/alibaba/fluss/client/write/Sender.java
```diff
@@ -280,6 +290,64 @@ private void maybeRemoveAndDeallocateBatch(WriteBatch batch) {
...
+                batchIt.hasNext(); ) {
+            Map.Entry<TableBucket, List<WriteBatch>> entry = batchIt.next();
+            List<WriteBatch> tbFlightBatches = entry.getValue();
+            if (tbFlightBatches != null) {
+                Iterator<WriteBatch> iter = tbFlightBatches.iterator();
+                while (iter.hasNext()) {
+                    WriteBatch batch = iter.next();
+                    if (batch.hasReachedDeliveryTimeout(accumulator.getDeliveryTimeoutMs(), now)) {
+                        iter.remove();
+                        if (!batch.isDone()) {
```
wuchong, line 329: Add comment to explain why the status of batch shouldn't be done here like Kafka?

fluss-client/src/main/java/com/alibaba/fluss/client/write/Sender.java
```diff
@@ -280,6 +290,64 @@ private void maybeRemoveAndDeallocateBatch(WriteBatch batch) {
...
+                                    now - batch.getCreatedMs())),
+                    false);
+        }
+    }
+
+    /** Get the in-flight batches that has reached batch delivery timeout. */
+    @VisibleForTesting
+    List<WriteBatch> getExpiredInflightBatches(long now) {
+        List<WriteBatch> expiredBatches = new ArrayList<>();
+        for (Iterator<Map.Entry<TableBucket, List<WriteBatch>>> batchIt =
```
wuchong, line 318: Need add a `synchronized (inFlightBatchesLock)` around the foreach? As it mutates the `inFlightBatches`

fluss-client/src/main/java/com/alibaba/fluss/client/write/WriterClient.java
```diff
@@ -249,7 +256,27 @@ private int configureRetries(boolean idempotenceEnabled) {
         return retries;
     }
 
-    private Sender newSender(short acks, int retries) {
+    private long configureBatchDeliveryTimeout(Configuration conf) {
```
wuchong, line 259: `validateAndGetDeliveryTimeout` ? As it doesn't "configure"

fluss-client/src/main/java/com/alibaba/fluss/client/write/Sender.java
```diff
@@ -216,7 +225,9 @@ private void sendWriteData() throws Exception {
         if (!batches.isEmpty()) {
             addToInflightBatches(batches);
 
-            // TODO add logic for batch expire.
+            // check and expire batches if they have reached batch delivery timeout. This can avoid
+            // the client to wait for a long time while the batch is forever retry.
+            checkAndExpireBatches(now);
```
wuchong, line 230: What happens if the batch is expired here but is still in the `sendWriteRequests(batches)`?  Will the expired batch still be sent to servers? What happens if the batches fail or success from the request?

wuchong (commented):
Could you add an IT case for this in `FlussFailServerTableITCase` (mock all servers are down and the request timeout)?

fluss-client/src/main/java/com/alibaba/fluss/client/write/Sender.java
swuferhong: > Need add a `synchronized (inFlightBatchesLock)` around the foreach? As it mutates the `inFlightBatches`

Cool, I encountered a concurrent modification error while testing today.

fluss-common/src/main/java/com/alibaba/fluss/config/ConfigOptions.java
swuferhong: Done!

fluss-common/src/main/java/com/alibaba/fluss/config/ConfigOptions.java
swuferhong: Done!

fluss-common/src/main/java/com/alibaba/fluss/config/ConfigOptions.java
swuferhong: Done!

fluss-client/src/main/java/com/alibaba/fluss/client/write/Sender.java
swuferhong: Done!

fluss-client/src/main/java/com/alibaba/fluss/client/write/WriterClient.java
swuferhong: Done!

PR #691, opened 2025-03-30
Make Flink Source Generic

polyzos:
This PR makes the FlinkSource Generic, so it can later be used and extended by the FlussSource that will be exposed for the Datastream API and can be typed to any Pojo.

For internal implementations and usage with the SQL/Table connector, we use `RowData` by default


fluss-flink/fluss-flink-common/src/main/java/com/alibaba/fluss/flink/source/reader/FlinkSourceReader.java
```diff
@@ -69,7 +69,11 @@ public FlinkSourceReader(
                                         projectedFields,
                                         flinkSourceReaderMetrics),
                         (ignore) -> {}),
-                new FlinkRecordEmitter(sourceOutputType),
+                new FlinkRecordEmitter(
+                        new RowDataDeserializationSchema(),
```
wuchong: We shouldn't hardcode `RowDataDeserializationSchema` here, otherwise, the `FlinkSourceReader` is bound to `RowData` instead of a generic `OUT` type. 

Here is how Kafka connector implement the generic deserialization schema: https://github.com/apache/flink-connector-kafka/blob/727327df7f0f2358c03e85022eae20079290a847/flink-connector-kafka/src/main/java/org/apache/flink/connector/kafka/source/KafkaSource.java#L162

1. The `FlinkSource` should take a generic deserialization schema.
2. The deserialization schema can be opened in `FlinkSource#createReader` which makes the code cleaner (doesn't need to catch exception)
3. initialize `FlinkRecordEmitter` with the opened deserialization schema in `FlinkSource#createReader` and pass to the constructor of `FlinkSourceReader`.

polyzos: Uh good catch, I totally missed that I added a concrete implementation, and I only tested the emitter with `RowData`. Since its generic now, I can add more

polyzos:
@wuchong Thank you for your comments.. I addressed them and also added tests for pojo in the `FlinkRecordEmitter`.
Let me know if there is anything else that requires attention.

fluss-flink/fluss-flink-common/src/main/java/com/alibaba/fluss/flink/source/emitter/FlinkRecordEmitter.java
```diff
@@ -41,21 +39,20 @@
  *
  * <p>when the record is from log data, it'll update the offset
  */
-public class FlinkRecordEmitter implements RecordEmitter<RecordAndPos, RowData, SourceSplitState> {
+public class FlinkRecordEmitter<OUT> implements RecordEmitter<RecordAndPos, OUT, SourceSplitState> {
     private static final Logger LOG = LoggerFactory.getLogger(FlinkRecordEmitter.class);
 
-    private final FlussRowToFlinkRowConverter converter;
+    private LakeRecordRecordEmitter<OUT> lakeRecordRecordEmitter;
+    private FlussDeserializationSchema<OUT> deserializationSchema;
```
wuchong: can be `final`.

wuchong (commented):
I pushed a commit to improve the code a bit to avoid duplicated anonymous `InitializationContext` definition. Will merge it once CI is passed.

PR #695, opened 2025-03-31
[flink] FlinkSink implements SupportsPreWriteTopology instead of WithPreWriteTopology

### LINKED ISSUE #622: FlinkSink implements SupportsPreWriteTopology instead of WithPreWriteTopology
### Search before asking

### Fluss version

main (development)

### Please describe the bug 🐞

In #579 we will let `FlinkSink` to implement  `WithPreWriteTopology`  to support shuffle by bucket id. But, in Flink-1.20, this interface will be deprecated and use `SupportsPreWriteTopology` instead.  So we should try to support different interface in different Flink version as we already support Flink multi version.

### Solution

_No response_

SteNicholas:
### Purpose

Linked issue: close #622.

### Brief change log

`FlinkSink` implements `SupportsPreWriteTopology` instead of `WithPreWriteTopology`.

### Tests

CI.

### API and Format

No.

### Documentation

No.

### DIFF
--- a/fluss-flink/fluss-flink-common/src/main/java/com/alibaba/fluss/flink/sink/FlinkSink.java
+++ b/fluss-flink/fluss-flink-common/src/main/java/com/alibaba/fluss/flink/sink/FlinkSink.java
@@ -28,6 +28,7 @@ import org.apache.flink.api.connector.sink2.Sink;
 import org.apache.flink.api.connector.sink2.SinkWriter;
 import org.apache.flink.api.connector.sink2.WriterInitContext;
 import org.apache.flink.runtime.metrics.groups.InternalSinkWriterMetricGroup;
+import org.apache.flink.streaming.api.connector.sink2.SupportsPreWriteTopology;
 import org.apache.flink.streaming.api.connector.sink2.WithPreWriteTopology;
 import org.apache.flink.streaming.api.datastream.DataStream;
 import org.apache.flink.table.data.RowData;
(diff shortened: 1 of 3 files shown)


fluss-flink/fluss-flink-common/src/main/java/com/alibaba/fluss/flink/sink/FlinkSink.java
```diff
@@ -42,13 +43,9 @@
...
-/**
- * Flink sink for Fluss.
- *
- * <p>TODO: WithPreWriteTopology need to be changed to supportsPreWriteTopology in Flink 1.20. Trace
- * by https://github.com/alibaba/fluss/issues/622.
- */
-class FlinkSink implements Sink<RowData>, WithPreWriteTopology<RowData> {
+/** Flink sink for Fluss. */
+class FlinkSink
+        implements Sink<RowData>, SupportsPreWriteTopology<RowData>, WithPreWriteTopology<RowData> {
```
wuchong: I don't want to be dependent on a deprecated API as this API is removed in Flink 2.0, and we are planning to support Flink 2.0 in this version. 

A solution can be just dependent on `SupportsPreWriteTopology` here, and make the new introduced `SupportsPreWriteTopology` to implements `WithPreWriteTopology` in `fluss-flink-1.18`.

wuchong (changes requested):

wuchong (approved):

PR #697, opened 2025-04-01
[client]Fix LogScannerImpl.poll always returns empty on the first request

### LINKED ISSUE #683: When consuming a table without a primary key from earliest, the first time fetch data return null
### Search before asking

### Fluss version

0.6.0 (latest release)

### Please describe the bug 🐞
When i run the test case com.alibaba.fluss.flink.source.FlinkTableSourceITCase#testNonPkTableRead, method com.alibaba.fluss.client.table.scanner.log.LogScannerImpl#poll
 returns an empty result first time,the poll method returns the correct result on the second request
 

### Solution

_No response_

Alibaba-HZY:
[client]Fix LogScannerImpl.poll always returns empty on the first request

### Purpose

Linked issue: close #683

### Tests
com.alibaba.fluss.client.table.FlussTableITCase#testPollOnce
com.alibaba.fluss.client.table.FlussTableITCase#testPollOnceEmpty

### DIFF
--- a/fluss-client/src/test/java/com/alibaba/fluss/client/table/FlussTableITCase.java
+++ b/fluss-client/src/test/java/com/alibaba/fluss/client/table/FlussTableITCase.java
@@ -64,6 +64,8 @@ import java.util.ArrayList;
 import java.util.Iterator;
 import java.util.List;
 import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.atomic.AtomicInteger;
 
 import static com.alibaba.fluss.client.table.scanner.batch.BatchScanUtils.collectRows;
 import static com.alibaba.fluss.record.TestData.DATA1_ROW_TYPE;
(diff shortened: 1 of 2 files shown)


fluss-client/src/test/java/com/alibaba/fluss/client/table/FlussTableITCase.java
```diff
@@ -156,6 +158,113 @@ void testAppendWithSmallBuffer(boolean indexedFormat) throws Exception {
         }
     }
 
+    @ParameterizedTest
+    @ValueSource(booleans = {true, false})
+    void testPollOnce(boolean indexedFormat) throws Exception {
```
wuchong: We don't need to parameterize the test as the polling behavior has nothing to do with log format. Avoid unnecessary parameterized test can help us reduce the test time.

fluss-client/src/test/java/com/alibaba/fluss/client/table/FlussTableITCase.java
```diff
@@ -156,6 +158,113 @@ void testAppendWithSmallBuffer(boolean indexedFormat) throws Exception {
...
+                    count++;
+                }
+            }
+            logScanner.close();
+        }
+    }
+
+    @ParameterizedTest
+    @ValueSource(booleans = {true, false})
+    void testPollOnceEmpty(boolean indexedFormat) throws Exception {
```
wuchong: I believe this test is unnecessary for the following reasons:

1. **Ambiguity in `poll` Behavior**:  
   The modified behavior of the `poll` method only guarantees that it will return results if data has been flushed. However, if the data has not been flushed, the behavior is non-deterministic—it may or may not return data. As a result, there is no clear expectation for the test to validate.

2. **Test Instability**:  
   The test is inherently unstable because it relies on timing-sensitive conditions. For instance, data might be flushed after a timeout, causing the first `poll` call to unexpectedly return data. This instability arises from the unpredictable nature of asynchronous processes and timeouts in testing environments. Specifically, it is impossible to reliably predict whether the timeout or the asynchronous flush will occur first, making the test unreliable.

wuchong (commented):
Thanks for the contribution @Alibaba-HZY , I left some comments.

fluss-client/src/test/java/com/alibaba/fluss/client/table/FlussTableITCase.java
HZY-Wade: You're right. I'm removing the test

wuchong (approved):

PR #701, opened 2025-04-01
[server] Use ChannelId#hashCode() in ServerChannelInitializer

### LINKED ISSUE #696: Use ChannelId#hashCode() in ServerChannelInitializer
### Search before asking

### Description

Currently, we use `ch.id().asLongText().hashCode()` to calculate a hash code for a Netty channel. However, we can use `ChannelId#hashCode()` directly which avoids re-calculate (see `io.netty.channel.DefaultChannelId#hashCode`).

### Willingness to contribute

caozhen1937:
### Purpose

Linked issue: close #696

### DIFF
--- a/fluss-rpc/src/main/java/com/alibaba/fluss/rpc/netty/server/ServerChannelInitializer.java
+++ b/fluss-rpc/src/main/java/com/alibaba/fluss/rpc/netty/server/ServerChannelInitializer.java
@@ -62,8 +62,7 @@ final class ServerChannelInitializer extends ChannelInitializer<SocketChannel> {
 
         // TODO: we can introduce a smarter and dynamic strategy to distribute requests to
         //  channels
-        int channelIndex =
-                MathUtils.murmurHash(ch.id().asLongText().hashCode()) % requestChannels.length;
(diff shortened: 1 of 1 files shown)


caozhen1937:
Hi @swuferhong , please help me understand what is the background of TODO here? What do we need to do?
https://github.com/alibaba/fluss/blob/063793b2677ceb1c3caea186aed31bf66862c5fa/fluss-rpc/src/main/java/com/alibaba/fluss/rpc/netty/server/ServerChannelInitializer.java#L63

wuchong (approved):

wuchong:
> Hi @swuferhong , please help me understand what is the background of TODO here? What do we need to do?
> 
> https://github.com/alibaba/fluss/blob/063793b2677ceb1c3caea186aed31bf66862c5fa/fluss-rpc/src/main/java/com/alibaba/fluss/rpc/netty/server/ServerChannelInitializer.java#L63-L66

We don't need to do anything for this TODO comment in this PR. The comment means the current hash strategy is a static strategy, which may lead to data skew on the request channels. A better solution (but of course, very complex) is to determine the request channel id dynamically according to the workload of every channel.

PR #708, opened 2025-04-02
[connector] Move META-INF/services/org.apache.flink.table.factories.Factory to each submodule

### LINKED ISSUE #707: Move META-INF/services/org.apache.flink.table.factories.Factory to each submodule
### Search before asking

### Description

Currently, fluss-flink-1.18 fluss-flink-1.19, and  fluss-flink-1.20 share same META-INF/services/org.apache.flink.table.factories.Factory  in fluss-flink-common.

If we extends FlinkCatalogFactory later in fluss-flink-1.20（for example Flink120CatalogFactory), then tests of fluss-flink-1.20 will still load parent FlinkCatalogFactory class from fluss-flink-common. Then exception will be thrown because multiple idendifier `fluss` which is not allowed in Flink Factory load.

### Willingness to contribute

loserwang1024:
…actory to each submodule

### Purpose

Linked issue: close #707


loserwang1024:
@wuchong , CC, WDYT?

wuchong (approved):

PR #717, opened 2025-04-06
Adapt `local-cluster.sh` and Dockerfile for `bind.listeners` config option

### LINKED ISSUE #618: refactor  server.yaml and shell to start after multiple endpoints
### Search before asking

### Motivation

After supporting multiple endpoints, we need:
1.  add new `bind.listener` to server.yaml 
2. coordinator-server.sh and tablet-server.sh should use different yaml config(otherwise, in local-cluster.sh), ports will conflict。

### Solution

_No response_

### Anything else?

_No response_

### Willingness to contribute

michaelkoepf:
### Purpose

Linked issue: close #618

### Brief change log

**Modules**

`fluss-dist`

Choose random port for `bind.listeners` in `local-cluster.sh`

**Docker**

Ensure backward compatability by removing `bind.listeners` from YAML config files.

### Tests

n/a

### API and Format

n/a

### Documentation

n/a

### DIFF
--- a/docker/docker-entrypoint.sh
+++ b/docker/docker-entrypoint.sh
@@ -20,6 +20,9 @@ COMMON_CONF_FILE="${FLUSS_HOME}/conf/common.yaml"
 prepare_configuration() {
     additional_conf_file="${FLUSS_HOME}/conf/$1"
 
+    # backward compatability: allow to use old [coordinator|tablet-server].host option in FLUSS_PROPERTIES
+    sed -i '/bind.listeners:/d' "${COMMON_CONF_FILE}" "$additional_conf_file"
+
     if [ -n "${FLUSS_PROPERTIES}" ]; then
         # copy over all configuration options of FLUSS_PROPERTIES to the common and the additional configuration file
         # since we cannot tell which configuration options are specific and which are not


fluss-dist/src/main/resources/common.yaml
```diff
@@ -14,14 +14,19 @@
 # limitations under the License.
 ################################################################################
 
+# Common default configuration file that is used by ALL start scripts, the Fluss Servers and the Lakehouse CLI.
+# In addition, there exist additional, dedicated default configuration files for some components that are also placed
+# in this folder. The dedicated configuration files are ONLY read by the respective Java applications but NOT by the
+# start scripts. In that case, configuration options in the dedicated files overwrite configuration options
+# in the common (i.e., this) file.
+
```
michaelkoepf, line 22: Note for reviewer: I suggest to rename `server.yaml` to `common.yaml` from 0.7.0 onward for the following reasons.

1. The file name `server` is a bit misleading, because the config file is not only read by the server but also by `config.sh` and the lakehouse module. And it's easier to change it now than at some later point.
2. It avoids confusion. With renaming it, there is a clear separation and it is clear which config files are read. It is either `server.yaml` or `common.yaml` + depending on the module, an additional file ([`coordinator-server|tablet-server].yaml`).

We can remove `server.yaml` when we deprecate the old `[coordinator|tablet-server]-host` config options.

Of course, I am open to discuss if renaming the file is a good approach or not.

wuchong: Sorry, @michaelkoepf, but from the user's perspective, the `server.yaml` file is intended for configuring servers, including both the coordinator and tablet servers. This is because it is used when running the `coordinator-server.sh` and `tablet-server.sh` scripts. On the other hand, `config.sh` is an internal shell script that is utilized by `coordinator-server.sh` and `tablet-server.sh` to manage configurations.

Additionally, in the new lakehouse architecture, the lakehouse-specific script will be removed, and the Flink script will be used to start the tiering service. At that point, the `server.yaml` file will be exclusively used for configuring Fluss servers.

michaelkoepf: @wuchong is just the renaming of `server.yaml` to `common.yaml` an issue? or also the additional `[coordinator-server|tablet-server].yaml` file?

wuchong: @michaelkoepf In my opinion, there is no need to separate the configuration files. Whether starting a Coordinator or a TabletServer, users can simply use the `server.yaml` file. If a user wants to run multiple servers on the same machine, they can create multiple YAML files and specify the appropriate YAML file as a parameter for the startup script. This approach is similar to how Kafka handles configurations with the command:  

```bash
bin/kafka-server-start.sh config/server.properties
```
This method offers greater flexibility compared to hardcoding 2 YAML files, as multiple TabletServers often require different node IDs and host configurations, making it impractical to share the same `yaml` file.  I believe supporting startup with a custom YAML file is a missing feature in our current implementation. We should consider creating a separate issue to address this functionality.

michaelkoepf: @wuchong 

> If a user wants to run multiple servers on the same machine, they can create multiple YAML files and specify the appropriate YAML file as a parameter for the startup script.

this means from 0.7.0 onwards, users must specify their own yaml file when using `local-cluster.sh` when they want to use the new `bind.listeners`, right? and `local-cluster.sh` must accept two additional arguments for the 2 yaml files (one for the tablet and one for the coordinator server)?

> I believe supporting startup with a custom YAML file is a missing feature in our current implementation.

I think the functionality in the Java code is already there, right? See [here](https://github.com/alibaba/fluss/blob/e214c314c4187c4333181c75c2e70b1b41443a36/fluss-common/src/main/java/com/alibaba/fluss/config/GlobalConfiguration.java#L44) and it should work with `coordinator-server.sh` and `tablet-server.sh`. We just need to find a solution for `local-cluster.sh` (see previous paragraph).

> We should consider creating a separate issue to address this functionality.

Separate issue or also separate PR? so i know what i should revert in this PR and what not.

wuchong: @michaelkoepf 
> this means from 0.7.0 onwards, users must specify their own yaml file when using local-cluster.sh when they want to use the new bind.listeners, right? and local-cluster.sh must accept two additional arguments for the 2 yaml files (one for the tablet and one for the coordinator server)?

We need to adapt `local-cluster.sh` to keep the original behavior. I think we can change the line `"${FLUSS_BIN_DIR}"/tablet-server.sh start` in `local-cluster.sh` to `"${FLUSS_BIN_DIR}"/tablet-server.sh start -Dbind.listeners=FLUSS://localhost:0`. So that, Coordinator use the `bind.listeners` configured in `server.yaml` (i.e., `FLUSS://localhost:9123`), and TabletServer use a random port on the same host. 

> I think the functionality in the Java code is already there, right? See [here](https://github.com/alibaba/fluss/blob/e214c314c4187c4333181c75c2e70b1b41443a36/fluss-common/src/main/java/com/alibaba/fluss/config/GlobalConfiguration.java#L44) and it should work with coordinator-server.sh and tablet-server.sh. We just need to find a solution for local-cluster.sh (see previous paragraph).

Cool. I think we need this feature to update the page https://alibaba.github.io/fluss-docs/docs/install-deploy/deploying-distributed-cluster/ that starts Coordinator and TabletServer on the same node. 

> Separate issue or also separate PR? so i know what i should revert in this PR and what not.

Let's separate into multiple issues and PRs, they are not related too much, this can also make reviewing easier.

PR #719, opened 2025-04-07
Fix an exception when create and drop a table with the same name multiple times

### LINKED ISSUE #712: The table id in table zknode is not same with the partition id in the children node of the table in zk
### Search before asking

### Fluss version

0.6.0 (latest release)

### Please describe the bug 🐞

Drop a table and create a table with the same table t1. Then it may happen  in zk that the `t1` zk node is with table id 12, but the children partition node is with table id 11. Then in client, the following exception will happen: 
```
Caused by: java.lang.IllegalArgumentException: table path not found for tableId 11 in cluster
	at com.alibaba.fluss.cluster.Cluster.lambda$getTablePathOrElseThrow$1(Cluster.java:177)
	at com.alibaba.fluss.cluster.Cluster.getTablePathOrElseThrow(Cluster.java:175)
	at com.alibaba.fluss.client.utils.MetadataUtils.lambda$getTableMetadataToUpdate$3(MetadataUtils.java:221)
	at java.lang.Iterable.forEach(Iterable.java:75)
	at com.alibaba.fluss.client.utils.MetadataUtils.getTableMetadataToUpdate(MetadataUtils.java:217)
	at com.alibaba.fluss.client.utils.MetadataUtils.lambda$sendMetadataRequestAndRebuildCluster$1(MetadataUtils.java:122)
```
It happens with following steps:
1. create tablet t1 with id 11
2. drop tablet t1
3. coordinator listen the t1 is created, do auto partition, create the partition in the zk path `tables/t1/partitions/p1` with table id 11
4. coordinator listen the t1 is dropped, just remove t1 id = 11 remove `AutoPartitionManager`.
5. create tablet t1 with id 12, upsert the zk path `tables/t1` with table id 12
So the inconsistent happens.

### Solution

Seems we should drop the partitions in step 4 after we listen the table `t1` is dropped.
If we don't drop the partitions, it may cause partition leak.

1. create tablet t1 with id 11
2. drop tablet t1
3. coordinator listen the t1 is created, do auto partition 
4. coordinator listen the t1 is dropped
Note, the partitions created in step 3  won't be dropped then.

zcoo:
### Purpose

Linked issue: close #712

### Brief change log
Drop table's all partitions when coordinator listens a partitioned table dropped.

### DIFF
--- a/fluss-server/src/main/java/com/alibaba/fluss/server/coordinator/AutoPartitionManager.java
+++ b/fluss-server/src/main/java/com/alibaba/fluss/server/coordinator/AutoPartitionManager.java
@@ -353,4 +353,9 @@ public class AutoPartitionManager implements AutoCloseable {
             periodicExecutor.shutdownNow();
         }
     }
+
+    @VisibleForTesting
+    public Map<Long, TableInfo> getAutoPartitionTables() {
+        return autoPartitionTables;
+    }
 }
--- a/fluss-server/src/main/java/com/alibaba/fluss/server/coordinator/CoordinatorEventProcessor.java
+++ b/fluss-server/src/main/java/com/alibaba/fluss/server/coordinator/CoordinatorEventProcessor.java
@@ -519,9 +519,6 @@ public class CoordinatorEventProcessor implements EventProcessor {
                 tableInfo.getTablePath(),
                 tableInfo.getTableId(),
                 createTableEvent.getTableAssignment());
-        if (createTableEvent.isAutoPartitionTable()) {
-            autoPartitionManager.addAutoPartitionTable(tableInfo);
-        }
     }
 
     private void processCreatePartition(CreatePartitionEvent createPartitionEvent) {
@@ -554,9 +551,6 @@ public class CoordinatorEventProcessor implements EventProcessor {
 
         coordinatorContext.queueTableDeletion(Collections.singleton(dropTableEvent.getTableId()));
         tableManager.onDeleteTable(dropTableEvent.getTableId());
-        if (dropTableEvent.isAutoPartitionTable()) {
-            autoPartitionManager.removeAutoPartitionTable(dropTableEvent.getTableId());
-        }
     }
 
     private void processDropPartition(DropPartitionEvent dropPartitionEvent) {
--- a/fluss-server/src/main/java/com/alibaba/fluss/server/coordinator/CoordinatorServer.java
+++ b/fluss-server/src/main/java/com/alibaba/fluss/server/coordinator/CoordinatorServer.java
@@ -153,6 +153,8 @@ public class CoordinatorServer extends ServerBase {
             this.metadataCache = new ServerMetadataCacheImpl();
 
             MetadataManager metadataManager = new MetadataManager(zkClient, conf);
+            this.autoPartitionManager =
+                    new AutoPartitionManager(metadataCache, metadataManager, conf);
             this.coordinatorService =
                     new CoordinatorService(
                             conf,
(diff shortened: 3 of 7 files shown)


wuchong:
I think there are issues with the current fix. 

When deleting `tables/t1`, its child partitions, such as `tables/t1/partitions/p1`, should also be deleted automatically. There should be no need to explicitly and additionally delete the partition directories.

In my opinion, the root cause here is that after ZooKeeper deletes `tables/t1`, the `AutoPartitionManager` is triggered and recreates the partitions (with the old table ID). This results in an exception when attempting to recreate the table.

The solution is straightforward: before dropping the table, the `AutoPartitionManager` should unregister the table id first.

With the current fix, which deletes partitions in the `processDropTableEvent`, may lead to the deletion of partitions belonging to the new table, potentially causing data loss.

cc @luoyuxia

zcoo:
@wuchong I think your solution may be inconsistent with current code implement.

Here is the original code logic:

Step 1. Create Table 1 (only created on zk node without adding to AutoPartitionManager)
Step 2. Delete Table 1 (only dropped on zk node)
Step 3. zk table creation is being monitored and the table creation request is added to the CoordinatorEventProcessor request queue
Step 4. The zk deletion table is being monitored and the deletion request is added to the CoordinatorEventProcessor request queue
Step 5. CoordinatorEventProcessor queue executes table creation request, AutoPartitionManager adds the table zk, creates table partition nodes
Step 6. CoordinatEventProcessor queue executes droping a table, but only unregister for table 1 from AutoPartitionManager without deleting the table partition node, resulting in the retention of the zk table partition node

"The solution is straightforward: before dropping the table, the AutoPartitionManager should unregister the table id first." 
So this solution doesn't work because before Step 2, the Table 1 haven't been register from AutoPartitionManager.

wuchong:
> Step 6. CoordinatEventProcessor queue executes droping a table, but only unregister for table 1 from AutoPartitionManager without deleting the table partition node, resulting in the retention of the zk table partition node

The partition node should has already been deleted in Step2, because we use `deletingChildrenIfNeeded()` for deleting the table node, see the code: 
https://github.com/alibaba/fluss/blob/main/fluss-server/src/main/java/com/alibaba/fluss/server/zk/ZooKeeperClient.java#L362

So there is no need for `AutoPartitionManager` to delete any partition nodes. If there is any dirty partition node exist, that's caused by unregitstering `AutoPartitionManager` for the table too late. 

> So this solution doesn't work because before Step 2, the Table 1 haven't been register from AutoPartitionManager.

Yes, I agree. So we need to register it into `AutoPartitionManager` in Step1 and unregister it in Step2. Besides, we need to remove the registering and unregsitering in CoordinatorEventProcessor.

luoyuxia:
+1 for register into AutoPartitionManager in  `CoordinatorService#createTable`.  No need to wait to get `CreateTableEvent` to put to AutoPartitionManager.

wuchong:
@zcoo @luoyuxia I reviewed the new fix and the failed cases. It seems the fix is not such easy as expected. We also need to handle the databse drop: read all the table_id of the tables in the database, which is very heavy operation. Besides, the partition node may still created under a re-created table node because a table-drop-recreate may happen in the middle of auto partition scheduling. It seems we have to tolerate the inconsistent case. 

Thus, I prefer a simple fix for this problem (see the PR https://github.com/alibaba/fluss/pull/766), and fix the rare inconsistent   case in a long-term issue https://github.com/alibaba/fluss/issues/767.

wuchong:
fixed by #766

PR #724, opened 2025-04-08
[server] IOException should shutdown TabletServer

### LINKED ISSUE #722: Write disk full may cause CorruptMessageException while fetch the records
### Search before asking

### Fluss version

main (development)

### Please describe the bug 🐞

one flink job write to fluss, one flink job read from fluss, then 

1. fluss disk is full, write throw exception which cause flink failover
2. the job read from fluss will throw CorruptMessageException

### Solution

1. When disk is full, the tablet server should self-kill directly just like kafka so that the write can still fall into another server instead of failover 
2. For `CorruptMessageException`, I get the log segment file and write code to read the log segment, haven't found any CorruptMessage. Also, when I restart the tablet server, no `CorruptMessageException` will be thrown. Seems the disk full cause unexpected state of tablet server, self-kill should fix the exception

luoyuxia:
### Purpose

Linked issue: close #722

When encounter `IOException`, shutdown the server directly to avoid `CorruptMessageException`

### Brief change log

When appendLog or putKv encounter`IOException`, shutdown the server directly

### Tests

TabletServerFailOverITCase#testIOExceptionShouldStopTabletServer

### DIFF
--- a/fluss-client/src/test/java/com/alibaba/fluss/client/admin/FlussAdminITCase.java
+++ b/fluss-client/src/test/java/com/alibaba/fluss/client/admin/FlussAdminITCase.java
@@ -54,8 +54,6 @@ import com.alibaba.fluss.metadata.TableBucket;
 import com.alibaba.fluss.metadata.TableDescriptor;
 import com.alibaba.fluss.metadata.TableInfo;
 import com.alibaba.fluss.metadata.TablePath;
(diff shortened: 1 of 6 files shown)


luoyuxia:
@wuchong Could you please help review this pr.

wuchong (approved):

PR #726, opened 2025-04-08
[Connector] Pojo To RowData Utility

MehulBatra:
Purpose
Linked issue: https://github.com/alibaba/fluss/issues/723

> Utility for PojoToRowData Converter for the end users to use inbuild for a quick startup.

Brief change log:

- Added utility to convert PojoToRowData
- Added tests for primitive types (boolean, numeric, floating point)
- Added tests for complex types (decimal, temporal, binary, char)
- Added edge case tests for null values and nullable fields
- Used Types.POJO(pojoClass) to get PojoTypeInfo
- Implemented the Converter Pattern, added FieldConverter interface, and Type-specific converter implementations for each data type.
- Made sure to not support for Nested Fields, as currently doesn't support one.

Tests:

- Added unit tests for all Fluss data types including boolean, numeric, decimal, date/time, and binary types, with appropriate value verification.
- Added a unit test to support that if a user passes a nested field, it would throw an error.

API and Format

Documentation


polyzos:
@MehulBatra as discussed there has already been some work around this [here](https://github.com/alibaba/fluss/pull/710/files#diff-ecd3113126114d33c87a5af74a2f0e3bd6fc00c650f6ca628bf1bb88668f9e2bR41).

So maybe after it's merged, maybe we can bring enhancements on it and suggest is as a helper utility for users that want to get the job done quickly.. and introduce some kind of caching so it doesn't have to recompute things repeatedly.

fluss-flink/fluss-flink-common/src/main/java/com/alibaba/fluss/flink/utils/PojoToRowDataConverter.java
```diff
@@ -0,0 +1,189 @@
...
+     *
+     * @param rowType The row schema to use for conversion
+     */
+    public PojoToRowDataConverter(Class<T> pojoClass, RowType rowType) {
+        this.pojoClass = pojoClass;
+        this.rowType = rowType;
+        initializeFieldMap();
+    }
+
+    private void initializeFieldMap() {
```
wuchong: Instead of analyze the pojo class using reflection by ourselves, I would suggest directly using Flink `Types.POJO(class)` to get the `PojoTypeInfo` and we can get the Java `Field` from it. If you look into the source code of `TypeExtractor.createTypeInfo`, you will find devil is in the details... 

Besides, we can't flatten all the nested fields, because there may be different fields with the same field name. Such nested POJO fields should be converted into nseted `InternalRow`. We don't need to support this case and can just throw nested POJO fields, as we currently doesn't support nested row type.

fluss-flink/fluss-flink-common/src/main/java/com/alibaba/fluss/flink/utils/PojoToRowDataConverter.java
```diff
@@ -0,0 +1,189 @@
...
+ *     new PojoToRowDataConverter<>(Order.class, rowType);
+ *
+ * // Convert a POJO to RowData
+ * Order order = new Order(1001L, 5001L, 10, "123 Mumbai");
+ * RowData rowData = converter.convert(order);
+ * }</pre>
+ *
+ * @param <T> The POJO type to convert
+ */
+public class PojoToRowDataConverter<T> {
```
wuchong: I think we can learn the implementation from `org.apache.flink.table.data.util.DataFormatConverters.PojoConverter` which 1. first checks the date type of `schema` and the Java class of the `Field`, and determine the concrete converter implementation (e.g., 3 different converters for TIMESTAMP type). 
2. use the converter to convert the object into a GenericRow.

wuchong (commented):
Thanks for the contribution @MehulBatra , I left some comments about the implementation, please let me know if you have any questions.

MehulBatra:
> Thanks for the contribution @MehulBatra , I left some comments about the implementation, please let me know if you have any questions.

Thanks for the feedback. I will address these comments over the coming weekend and get back to you in case I am stuck!

fluss-flink/fluss-flink-common/src/main/java/com/alibaba/fluss/flink/utils/PojoToRowDataConverter.java
MehulBatra: Sure, took the inspiration from that itself, and tried to go that route, Implemented the Converter Pattern using FieldConverter interface.

MehulBatra:
@wuchong Jark made changes as per the comments, please have a look and let me know if I underdid or overdid anything, I will accommodate that.

wuchong (approved):
@MehulBatra  

The pull request looks good to me overall, but I have one key concern: it currently implements a conversion from POJO to Flink's `RowData`. In my opinion, what we actually need is a utility for converting POJOs to Fluss' `InternalRow`. This utility would be essential for writing a `DataStream<POJO>` to the Fluss Sink, as all data types must eventually be converted into Fluss' `InternalRow`—this is the data type expected by `AppendWriter` and `UpsertWriter`.

On the other hand, the conversion from POJO to Flink's `RowData` is already implemented by Flink itself. This is used in scenarios where a `DataStream<T>` is converted into a Flink `Table` (e.g., via `org.apache.flink.table.api.bridge.java.StreamTableEnvironment#fromDataStream(org.apache.flink.streaming.api.datastream.DataStream<T>)`).

To address this, I have updated the pull request to focus on converting POJOs to Fluss' `InternalRow`. Additionally, I have added a pre-check for the field types of the POJO to ensure compatibility and prevent potential issues.

wuchong:
Merging...

MehulBatra:
> @MehulBatra
> 
> The pull request looks good to me overall, but I have one key concern: it currently implements a conversion from POJO to Flink's `RowData`. In my opinion, what we actually need is a utility for converting POJOs to Fluss' `InternalRow`. This utility would be essential for writing a `DataStream<POJO>` to the Fluss Sink, as all data types must eventually be converted into Fluss' `InternalRow`—this is the data type expected by `AppendWriter` and `UpsertWriter`.
> 
> On the other hand, the conversion from POJO to Flink's `RowData` is already implemented by Flink itself. This is used in scenarios where a `DataStream<T>` is converted into a Flink `Table` (e.g., via `org.apache.flink.table.api.bridge.java.StreamTableEnvironment#fromDataStream(org.apache.flink.streaming.api.datastream.DataStream<T>)`).
> 
> To address this, I have updated the pull request to focus on converting POJOs to Fluss' `InternalRow`. Additionally, I have added a pre-check for the field types of the POJO to ensure compatibility and prevent potential issues.

Thanks @wuchong  for identifying and fixing this critical issue! I initially misunderstood the relationship between Flink's RowData and Fluss' InternalRow.
I'll make sure to pay closer attention to these format distinctions in future PRs. Really appreciate your help and guidance!

Issue #703, opened 2025-04-02
When TabletServer uncleaned shutdown, replica may out of isr set because OutOfOrderSequenceException


swuferhong:
### Search before asking

- [x] I searched in the [issues](https://github.com/alibaba/fluss/issues) and found nothing similar.

### Fluss version

main (development)

### Please describe the bug 🐞

When I try to do an uncleaned shutdown, I found that some follower replicas were not added back to the `ISR` set. Upon analysis, it was discovered that the `ReplicaFetcherThread` encountered an error while writing the fetched replicaData to local storage:

`ERROR com.alibaba.fluss.server.replica.fetcher.ReplicaFetcherThread [] - Unexpected error occurred while processing data for bucket TableBucket{tableId=55, partitionId=150, bucket=79} at offset 1638983
com.alibaba.fluss.exception.OutOfOrderSequenceException: Out of order batch sequence for writer 9427 at offset 1638983 in table-bucket TableBucket{tableId=55, partitionId=150, bucket=79} : 69 (incoming batch seq.), -1 (current batch seq.)

![Image](https://github.com/user-attachments/assets/ff3d29b5-7c72-41fb-a4e5-81fa38a299de)

As this happen, `ReplicaFetcherThread` will remove this bucket which cause this follower didn't fetch any data from leader, and will be kill off from isr set.

The root cause of this error is that when one `tabletServer` takes a long time to recover from a shutdown, the follower of one replica may have a very large offset gap compared to the leader at this point, and multiple segments may have already been moved to remote storage. Currently, Fluss does not handle this situation properly, resulting in `WriterState` failing to recover correctly. This error can be reproduced locally as follows:
and an tests in  RemoteLogITCase, like:

```java
 @Test
    void testFollowerFetchMoveToRemoteLogWithWriterStates() throws Exception {
        long tableId =
                createTable(FLUSS_CLUSTER_EXTENSION, DATA1_TABLE_PATH, DATA1_TABLE_DESCRIPTOR);
        TableBucket tb = new TableBucket(tableId, 0);

        FLUSS_CLUSTER_EXTENSION.waitUtilAllReplicaReady(tb);
        int leader = FLUSS_CLUSTER_EXTENSION.waitAndGetLeader(tb);
        int follower;
        for (int i = 0; true; i++) {
            if (i != leader) {
                follower = i;
                break;
            }
        }
        // kill follower, and restart after some segments in leader has been copied to remote.
[...]

PR #728, opened 2025-04-09
[log] Follower fetch log already move to remote need to return RemoteLogFetchInfo to rebuild WriterState instead of throwing LogOffsetOutOfRangeException

swuferhong:
### Purpose

Linked issue: #703 

This pr is aims to change the way if follower fetch log already move to remote, in this situation, we need to return RemoteLogFetchInfo to rebuild WriterState instead of throwing LogOffsetOutOfRangeException. The reason why doing that see: #703


swuferhong:
> The test `LogTabletTest.testWriterStateTruncateFullyAndStartAt:344` is failed.
Done!

fluss-server/src/main/java/com/alibaba/fluss/server/replica/fetcher/ReplicaFetcherThread.java
```diff
@@ -500,40 +500,49 @@ private LogAppendInfo processFetchResultFromLocalStorage(
...
             LogTablet log = replica.getLogTablet();
-            File snapshotFile = FlussPaths.writerSnapshotFile(log.getLogDir(), nextFetchOffset);
-            buildWriterIdSnapshotFile(snapshotFile, remoteLogSegment, rlm);
+            log.writerStateManager().truncateFullyAndStartAt(0L);
 
-            // Reload writer id snapshot.
-            log.writerStateManager().truncateFullyAndReloadSnapshots();
+            File snapshotFile = FlussPaths.writerSnapshotFile(log.getLogDir(), nextFetchOffset);
+            buildWriterIdSnapshotFile(snapshotFile, remoteLogSegmentWithMaxStartOffset, rlm);
+            log.writerStateManager().reloadSnapshots();
```
wuchong, line 599: Why separate the original `truncateFullyAndReloadSnapshots` into `truncateFullyAndStartAt(0)` and `reloadSnapshots()`? It seems it is not necessary. If not necessary, we can avoid introducing the `WriterStateManager#reloadSnapshots()` method.

fluss-server/src/main/java/com/alibaba/fluss/server/replica/ReplicaManager.java
```diff
@@ -1030,9 +1030,7 @@ public Map<TableBucket, LogReadResult> readFromLog(
                 }
 
                 FetchLogResultForBucket result;
-                if (replica != null
-                        && e instanceof LogOffsetOutOfRangeException
-                        && !isFromFollower) {
```
wuchong, line 1035: Why remove it?

fluss-server/src/main/java/com/alibaba/fluss/server/log/LogTablet.java
```diff
@@ -1016,12 +1016,8 @@ private void rebuildWriterState(long lastOffset, WriterStateManager writerStateM
...
         synchronized (lock) {
             localLog.checkIfMemoryMappedBufferClosed();
-            rebuildWriterState(
-                    writerStateManager,
-                    localLog.getSegments(),
-                    localLog.getLocalLogStartOffset(),
-                    lastOffset,
-                    false);
+            // TODO, set to logStartOffset instead of 0 if we introduce logStartOffset.
+            rebuildWriterState(writerStateManager, localLog.getSegments(), 0, lastOffset, false);
```
wuchong, line 1055: Why the original `localLog.getLocalLogStartOffset()` is incorrect? And what's the differences between `localLog.getLocalLogStartOffset()` and the new `logStartOffset` ?

fluss-server/src/main/java/com/alibaba/fluss/server/replica/fetcher/RemoteLeaderEndpoint.java
```diff
@@ -86,7 +94,36 @@ public CompletableFuture<Map<TableBucket, FetchLogResultForBucket>> fetchLog(
...
+                                for (PbFetchLogRespForBucket bucketResp : bucketsRespList) {
+                                    TableBucket tableBucket =
+                                            new TableBucket(
+                                                    tableId,
+                                                    bucketResp.hasPartitionId()
+                                                            ? bucketResp.getPartitionId()
+                                                            : null,
+                                                    bucketResp.getBucketId());
+                                    Replica replica =
+                                            replicaManager.getReplicaOrException(tableBucket);
```
wuchong: According to the **"Principle of Least Knowledge"** (also known as the Law of Demeter), `replicaManager` represents an unnecessarily large dependency in this context, as we only need to retrieve the table path. A cleaner approach would be to add the `TablePath` to both `BucketFetchStatus` and `InitialFetchStatus`. 

In the `buildFetchLogRequest` method, we can construct a `Map<Long, TablePath>` from the existing `Map<TableBucket, BucketFetchStatus> replicas`. This map can then be passed to `LeaderEndpoint#fetchLog`, ensuring that the dependency remains minimal and the code adheres to the principle of least knowledge.

swuferhong: Done!

fluss-server/src/main/java/com/alibaba/fluss/server/replica/ReplicaManager.java
swuferhong: > Why remove it?

For the follower, it is not appropriate to simply throw a LogOffsetOutOfRangeException error here. Instead, remote information should be returned to allow the follower to rebuild the WriterSnapshot.

fluss-server/src/main/java/com/alibaba/fluss/server/replica/fetcher/ReplicaFetcherThread.java
swuferhong: Adding some comments to identify why do this.

fluss-server/src/main/java/com/alibaba/fluss/server/log/LogTablet.java
swuferhong: Here, I use `0` as the `logStartOffset` passed into `rebuildWriterState()`. The reason is that the current implementation of `logStartOffset` in Fluss is not yet fully refined, and there may be cases where `logStartOffset` is not updated. As a result, `logStartOffset` is not yet reliable. Once the issue with correctly updating `logStartOffset` is resolved in issue #744 , we can use logStartOffset here.

Additionally, using `0` versus using `logStartOffset` does not affect correctness—they both can restore the complete `WriterState`. The only difference is that using `logStartOffset` can potentially skip over more segments.

wuchong (approved):

PR #729, opened 2025-04-09
[hotfix] Change writerState flush from async to sync while rolling log to avoid conflicts with the writerState flush during logTablet close

swuferhong:
### Purpose

This pr is aims to change `writerState` flush from async to sync while rolling log to avoid conflicts with the `writerState` flush during logTablet close.  

In the `LogTablet#roll()` method, the current implementation `writes the writerState` while holding the lock, but the `file flush operation` is performed outside the lock. However, this could potentially conflict with the `writerState write operation `in the `LogTablet#close() method`. Considering that the `writerState flush operation` itself is not very resource-intensive and the `roll()` method isn't called frequently, we consider change it to a sync flush approach.

### DIFF
--- a/fluss-server/src/main/java/com/alibaba/fluss/server/log/LogTablet.java
+++ b/fluss-server/src/main/java/com/alibaba/fluss/server/log/LogTablet.java
@@ -782,16 +782,12 @@ public final class LogTablet {
             // current writer state end offset (which corresponds to the log end offset),
             // we manually override the state offset here prior to taking the snapshot.
             writerStateManager.updateMapEndOffset(segment.getBaseOffset());
-            // We avoid potentially-costly fsync call, since we acquire UnifiedLog#lock here which
-            // could block subsequent produces in the meantime.
-            // flush is done in the scheduler thread along with segment flushing below.
-            Optional<File> maybeSnapshot = writerStateManager.takeSnapshot(false);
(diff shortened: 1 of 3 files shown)


wuchong:
I'm going to merge this first, but leave a comment here:

It seems Kafka introduced the async on purpose to address a performance problem when flushing the writer state takes too long. 

https://issues.apache.org/jira/browse/KAFKA-15046
https://github.com/apache/kafka/commit/d71d0639d9d5ae39153d3e1f52f8ba8bd2ae8044

swuferhong:
> It seems Kafka introduced the async on purpose to address a performance problem when flushing the writer state takes too long. What exception/problem when using async?
> 
> https://issues.apache.org/jira/browse/KAFKA-15046 [apache/kafka@d71d063](https://github.com/apache/kafka/commit/d71d0639d9d5ae39153d3e1f52f8ba8bd2ae8044)

When testing the Fluss cluster update, the `Fluss tabletServer shutdown` also needs to call `writerStateManager.takeSnapshot()`. If this conflicts with the takeSnapshot in the `roll()` method, it may happen that the shutdown process acquires the lock before the roll() has flushed the snapshot and proceeds to write the snapshot, which will cause errors. Therefore, a better approach than `async flushing` should be used here to address the issue of blocking writes.

wuchong (approved):

PR #734, opened 2025-04-10
[docs] fluss admin client

polyzos:
Add documentation for Fluss admin client API

### DIFF
--- a/website/docs/apis/java-client.md
+++ b/website/docs/apis/java-client.md
@@ -1,0 +1,292 @@
+---
+sidebar_label: "Java Client"
+sidebar_position: 1
+---
+
+<!--
+ Copyright (c) 2025 Alibaba Group Holding Ltd.
+
+ Licensed under the Apache License, Version 2.0 (the "License");
+ you may not use this file except in compliance with the License.
+ You may obtain a copy of the License at
+
+      http://www.apache.org/licenses/LICENSE-2.0
+
+ Unless required by applicable law or agreed to in writing, software
+ distributed under the License is distributed on an "AS IS" BASIS,
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+-->
+
+# Fluss  Java Client
+## Overview
+`Fluss Admin` is a thread-safe client that supports asynchronous operations for managing and inspecting Fluss resources. It communicates with the Fluss cluster via RPC calls and provides methods for:
+
+
+* Managing databases (create, drop, list)
+* Managing tables (create, drop, list)
+* Managing partitions (create, drop, list)
+* Retrieving metadata (schemas, snapshots, server information)
+
+## Initialization
+
+```java
+Configuration conf = new Configuration(); 
+conf.setString("bootstrap.servers", "localhost:9123");
+
+
+Connection connection = ConnectionFactory.createConnection(conf);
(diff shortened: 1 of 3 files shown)


website/docs/apis/java-client.md
```diff
@@ -0,0 +1,292 @@
...
+
+ Unless required by applicable law or agreed to in writing, software
+ distributed under the License is distributed on an "AS IS" BASIS,
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+-->
+
+# Fluss  Java Client
+## Overview
```
wuchong, line 23: We need to introduce how to add the Java client maven dependency. And introduce what APIs provided. Overview, Fluss provides `Admin` API and `Table` API, we can first introduce what the API mainly used for.

website/docs/apis/java-client.md
```diff
@@ -0,0 +1,292 @@
...
+conf.setString("bootstrap.servers", "localhost:9123");
+
+
+Connection connection = ConnectionFactory.createConnection(conf);
+Admin admin = connection.getAdmin();
+
+admin.listDatabases().get().forEach(System.out::println);
+```
+
+## Working Operations
```
wuchong, line 71: I don't think we need to introduce or list every APIs of `Admin` and `Table`. That should be covered by javadoc (https://github.com/alibaba/fluss/issues/608). 

Here, I suggest only covering the core APIs like a quickstart. The suggested TOC:
```
## Initialization
Introduce the API entrance `Connection` and how to get `Admin` and `Table` from it. 
## Admin API
### How to create database and table
## Table API
### How to write data
### How to read data
### How to lookup data
```

website/docs/apis/java-client.md
```diff
@@ -0,0 +1,292 @@
...
+    row.setField(1, user.getAge());
+    row.setField(2, TimestampNtz.fromLocalDateTime(user.getCreatedAt()));
+    row.setField(3, user.isActive());
+    return row;
+}).collect(Collectors.toList());
+        
+System.out.println("Upserting rows to the table");
+rows.forEach(row -> {     
+    try {
+        table.newUpsert().createWriter().upsert(row).get();
```
wuchong: Usually, we should reuse the writer, and `flush()` after writing a batch of records instead of blocking `get()` for every `upsert()`.

website/docs/apis/java-client.md
```diff
@@ -0,0 +1,292 @@
...
+    .exceptionally(ex -> {
+        System.err.println("Failed to create database: " + ex.getMessage());
+        return null;
+    });
+```
+
+#### Example Output
+```shell
+Database created successfully
+{"version":1,"comment":"This is a test database","custom_properties":{"owner":"data-team"}}
```
wuchong: Where is this output from?

wuchong (commented):
1. Docusaurus doesn't enable Java hightlight by default. We need to manually add highlighting for the Java language:
https://docusaurus.io/docs/markdown-features/code-blocks#supported-languages

2. It seems this is not only for Admin API, but also for Table API.

website/docs/apis/java-client.md
polyzos: this was from a local example

polyzos:
@wuchong comments addressed.. let me know if there is anything else that requires attention

wuchong (approved):

PR #737, opened 2025-04-11
[client] fix: throw LogOffsetOutOfRangeException when scan offset ttl log

### LINKED ISSUE #521: Subscribe from the log offset that has been ttl don't throw OffsetOutOfRangeException
### Search before asking

### Fluss version

main (development)

### Please describe the bug 🐞

If  subscribe from the log offset that has been delete by remote log ttl,  I expect it to throw OffsetOutOfRangeException, but it doesn't

Can be reproduce  by the following IT:
```
public class FlussLogITCase {
    @RegisterExtension
    public static final FlussClusterExtension FLUSS_CLUSTER_EXTENSION =
            FlussClusterExtension.builder()
                    .setNumOfTabletServers(3)
                    .setClusterConf(initConfig())
                    .build();

    private static Configuration initConfig() {
        Configuration conf = new Configuration();
        conf.set(ConfigOptions.LOG_SEGMENT_FILE_SIZE, MemorySize.parse("1kb"));
        conf.set(ConfigOptions.REMOTE_LOG_TASK_INTERVAL_DURATION, Duration.ofSeconds(1));
        return conf;
    }

    protected Connection conn;
    protected Admin admin;
    protected Configuration clientConf;

    @BeforeEach
    protected void setup() throws Exception {
        clientConf = FLUSS_CLUSTER_EXTENSION.getClientConfig();
        conn = ConnectionFactory.createConnection(clientConf);
        admin = conn.getAdmin();
    }

    @Test
    void testSubscribeTTLLog() throws Exception {
        TablePath tablePath = TablePath.of("test_db", "testSubscribeTTLLog");
        TableDescriptor tableDescriptor =
                TableDescriptor.builder()
                        .schema(
                                Schema.newBuilder()
                                        .column("a", DataTypes.INT())
                                        .column("b", DataTypes.STRING())
                                        .build())
                        // ttl eagerly
                        .property(ConfigOptions.TABLE_LOG_TTL, Duration.ofSeconds(1))
                        .distributedBy(1)
                        .build();
        createTable(tablePath, tableDescriptor, false);
        try (Table table = conn.getTable(tablePath)) {
            AppendWriter appendWriter = table.newAppend().createWriter();
            for (int n = 0; n < 10; n++) {
                for (int i = 0; i < 10; i++) {
                    appendWriter.append(row(1, "a"));
                }
                appendWriter.flush();
            }

            // wait 5s for log is ttl
            Thread.sleep(5_000);

            try (LogScanner logScanner = table.newScan().createLogScanner()) {
                logScanner.subscribe(0, 2);
                while (true) {
                    ScanRecords scanRecords = logScanner.poll(Duration.ofSeconds(1));
                    scanRecords.iterator().forEachRemaining(System.out::println);
                }
            }
        }
    }

    protected long createTable(
            TablePath tablePath, TableDescriptor tableDescriptor, boolean ignoreIfExists)
            throws Exception {
        admin.createDatabase(tablePath.getDatabaseName(), DatabaseDescriptor.EMPTY, ignoreIfExists)
                .get();
        admin.createTable(tablePath, tableDescriptor, ignoreIfExists).get();
        return admin.getTableInfo(tablePath).get().getTableId();
    }
}
```

### Solution

_No response_

gyang94:
### Purpose

Linked issue: close #521 

### Brief change log

Based on comments in issue #521 , this PR did:
1. remove the empty records check in LogFetcher, to pass the CompletedFetch (with error) into logFetcherBuffer

> if (!MemoryLogRecords.EMPTY.equals(logRecords)) { 

p.s. Based on the comments from @loserwang1024 , this empty records check may be necessary for some cases. 

> In oder to not signal notEmptyCondition, add completed fetch to buffer until log records is not empty.

I guess it is going to avoid several null records CompletedFetch (without error) in the first several polls. 
But we can create a new issue ticket to resolve those cases. 

2. throw IllegalStateException when LogOffsetOutOfRangeException occurred in LogFetcherCollector, to show stack trace.

### Tests

Add `FlussLogITCase`. 
This test need to set a small ttl and wait for log clean up. It may not be good to reuse the cluster in existing IT cases. Thus add a new IT case for it.


fluss-client/src/test/java/com/alibaba/fluss/client/table/scanner/log/FlussLogITCase.java
```diff
@@ -0,0 +1,120 @@
...
+            AppendWriter appendWriter = table.newAppend().createWriter();
+            for (int n = 0; n < 10; n++) {
+                for (int i = 0; i < 10; i++) {
+                    appendWriter.append(row(1, "a"));
+                }
+                appendWriter.flush();
+            }
+
+            // wait 5s for log is ttl
+            Thread.sleep(5_000);
```
wuchong: We should try the best to avoid relying on sleep to wait some condition. This is very unstable when running tests on different machines, and may waste too many time to wait a stable condition.

The suggested way to add test:
1. we can subscribe from `Long.MIN_VALUE` to mock a `LogOffsetOutOfRangeException` instead of waiting for a true TTL. 
2. then we don't need a dedicated IT case class, just add the test in `LogScannerITCase` without changing the cluster default configuration and table configuration.

fluss-client/src/test/java/com/alibaba/fluss/client/table/scanner/log/FlussLogITCase.java
```diff
@@ -0,0 +1,120 @@
...
+                                        .column("b", DataTypes.STRING())
+                                        .build())
+                        // ttl eagerly
+                        .property(ConfigOptions.TABLE_LOG_TTL, Duration.ofSeconds(1))
+                        .distributedBy(1)
+                        .build();
+        createTable(tablePath, tableDescriptor, false);
+        try (Table table = conn.getTable(tablePath)) {
+            AppendWriter appendWriter = table.newAppend().createWriter();
+            for (int n = 0; n < 10; n++) {
```
wuchong: we don't need too many rounds, just one round is enough.

fluss-client/src/main/java/com/alibaba/fluss/client/table/scanner/log/LogFetcher.java
```diff
@@ -327,21 +326,22 @@ private synchronized void handleFetchLogResponse(
...
-                                                readContext,
-                                                logScannerStatus,
-                                                // skipping CRC check if projection push downed as
-                                                // the data is pruned
-                                                isCheckCrcs,
-                                                fetchOffset);
-                                logFetchBuffer.add(completedFetch);
-                            }
+                            //                            if
+                            // (!MemoryLogRecords.EMPTY.equals(logRecords)) {
```
wuchong: Remove the code and add a comment here:
"The records in fetchResultForBucket may be MemoryLogRecords.EMPTY. However, we should still add it to logFetchBuffer because we need to handle potential errors within it."

fluss-client/src/main/java/com/alibaba/fluss/client/table/scanner/log/LogFetcher.java
```diff
@@ -327,21 +326,22 @@ private synchronized void handleFetchLogResponse(
                                     fetchResultForBucket.getHighWatermark());
                         } else {
                             LogRecords logRecords = fetchResultForBucket.recordsOrEmpty();
```
wuchong: remove the useless code.

fluss-client/src/main/java/com/alibaba/fluss/client/table/scanner/log/LogFetchCollector.java
```diff
@@ -261,7 +261,7 @@ private void handleInitializeErrors(
             LOG.warn("Received unknown table or bucket error in fetch for bucket {}", tb);
             metadataUpdater.checkAndUpdateMetadata(tablePath, tb);
         } else if (error == Errors.LOG_OFFSET_OUT_OF_RANGE_EXCEPTION) {
-            throw new LogOffsetOutOfRangeException(errorMessage);
+            throw new IllegalStateException(error.exception());
```
wuchong: throw `FetchException` like other errors below. 
```java
throw new FetchException(
                    String.format("The fetching offset %s is out of range", fetchOffset),
                    error.exception());
```

wuchong:
Hi @gyang94  do you have time to update the PR? If not available in the recent days, I can help to update it and merge it.

gyang94:
@wuchong Thanks for review. I will update it this or next week.

gyang94:
@wuchong  updated. please check.

wuchong (approved):
https://github.com/alibaba/fluss/pull/588 has fixed this problem. So I rebased the branch and kept the tests to verify that the problem is fixed. Will merge it once CI passes.

PR #754, opened 2025-04-16
[filesystem] Optimize the filesystem plugin

### LINKED ISSUE #753: Optimize the filesystem plugin
### Search before asking

### Motivation

Optimize the FileSystem Plugin code

### Solution

1.Extract common logics (e.g. createHadoopConfiguration, setCredentialProvider methods) into HadoopUtils.
﻿
2.Directly use FileSystem.get() method to initialize FileSystem instances. The URI processing is handled internally by this method, eliminating the need for external explicit declarations.

### Anything else?

_No response_

### Willingness to contribute

sunxiaojian:
### Purpose

Linked issue: close #753

### Tests

N/A

### API and Format

N/A

### Documentation

N/A

### DIFF
--- a/fluss-filesystems/fluss-fs-hadoop/src/main/java/com/alibaba/fluss/fs/hdfs/HadoopFsPlugin.java
+++ b/fluss-filesystems/fluss-fs-hadoop/src/main/java/com/alibaba/fluss/fs/hdfs/HadoopFsPlugin.java
@@ -27,7 +27,6 @@ import org.slf4j.LoggerFactory;
 
 import java.io.IOException;
 import java.net.URI;
-import java.net.UnknownHostException;
 
 import static com.alibaba.fluss.utils.Preconditions.checkArgument;
 import static com.alibaba.fluss.utils.Preconditions.checkNotNull;
@@ -47,6 +46,9 @@ public class HadoopFsPlugin implements FileSystemPlugin {
 
     public static final String SCHEME = "hdfs";
 
+    /** The prefixes that Fluss adds to the Hadoop config. */
+    public static final String[] FLUSS_CONFIG_PREFIXES = {"fluss.hadoop."};
+
     private static final Logger LOG = LoggerFactory.getLogger(HadoopFsPlugin.class);
 
     @Override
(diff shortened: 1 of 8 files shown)


sunxiaojian:
@wuchong PTAL.

wuchong:
What's the benefit of this refactoring?

sunxiaojian:
> What's the benefit of this refactoring?
Reduce duplicate code. For example, URI handling is already included in Filesystem.get(), so we can directly call Filesystem.get() and remove the duplicate declarations in each plugin.

sunxiaojian:
> What's the benefit of this refactoring?

Reduce duplicate code. For example, URI handling is already included in Filesystem.get(), so we can directly call Filesystem.get() and remove the duplicate declarations in each plugin.

wuchong:
Thanks, @sunxiaojian. However, I still prefer not to change the existing code. The current implementation is running very stably (and Flink also uses the same code). The proposed refactoring does not provide significant benefits—it does not introduce new features, fix bugs, or improve performance. 

The only advantage of the refactoring is that it reduces code duplication between Fluss and Hadoop. However, I think it’s acceptable to keep the duplicate code for now. Since we cannot predict how the Hadoop code might evolve in the future, maintaining separate implementations ensures greater stability. More importantly, it is difficult to guarantee that the new code and logic will be as stable as the existing implementation and may have problems in some corner cases.

In a word, from my perspective, the refactoring does not bring substantial benefits, but it could potentially introduce new bugs. Therefore, I suggest we leave the existing code unchanged.

sunxiaojian:
@wuchong Ok, I just think that optimizing it this way would make it more standardized when extending a new filesystems like GCS. The implementation of Filesystem.get is also relatively simple, it introduces local caching. This is because initializing Filesystem is relatively slow due to configuration loading,  it shouldn’t introduce any new issues. But as you said, it’s not absolutely necessary since there hasn’t been a performance requirement yet.

wuchong:
@sunxiaojian we already have a cache for `FileSystem`, see `com.alibaba.fluss.fs.FileSystem#get`. The `com.alibaba.fluss.fs.FileSystemPlugin#create` should only be called once in a JVM.

wuchong:
@sunxiaojian yeah, let's keep it as is. we can refactor them when it is truly needed.

sunxiaojian:
> @sunxiaojian yeah, let's keep it as is. we can refactor them when it is truly needed.

@wuchong ok, closed

PR #766, opened 2025-04-19
[server] AutoPartitionManager shouldn't create partitions for dropped table

### LINKED ISSUE #712: The table id in table zknode is not same with the partition id in the children node of the table in zk

wuchong:
### Purpose

Linked issue: close #712

This is a simple fix for #712 that during auto-partition creating it should first check if the table is dropped or re-created before creating partitions. This still don't guarantes the table id consist between the table node and partition node, because the table maybe re-created after the checking... But this should happen very rarely and is tracked in https://github.com/alibaba/fluss/issues/767.

### DIFF
--- a/fluss-server/src/main/java/com/alibaba/fluss/server/zk/ZooKeeperClient.java
+++ b/fluss-server/src/main/java/com/alibaba/fluss/server/zk/ZooKeeperClient.java
@@ -426,11 +426,6 @@ public class ZooKeeperClient implements AutoCloseable {
         return stat.getNumChildren();
     }
 
-    public void createPartitionsPath(TablePath tablePath) throws Exception {
-        String path = PartitionsZNode.path(tablePath);
-        zkClient.create().withMode(CreateMode.PERSISTENT).forPath(path);
-    }
-
     /** Create a partition for a table in ZK. */
     public void registerPartition(
             TablePath tablePath, long tableId, String partitionName, long partitionId)
--- a/fluss-server/src/main/java/com/alibaba/fluss/server/zk/ZooKeeperClient.java
+++ b/fluss-server/src/main/java/com/alibaba/fluss/server/zk/ZooKeeperClient.java
@@ -426,13 +426,17 @@ public class ZooKeeperClient implements AutoCloseable {
         return stat.getNumChildren();
     }
 
+    public void createPartitionsPath(TablePath tablePath) throws Exception {
+        String path = PartitionsZNode.path(tablePath);
+        zkClient.create().withMode(CreateMode.PERSISTENT).forPath(path);
+    }
+
     /** Create a partition for a table in ZK. */
     public void registerPartition(
             TablePath tablePath, long tableId, String partitionName, long partitionId)
             throws Exception {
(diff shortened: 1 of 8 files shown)


fluss-server/src/main/java/com/alibaba/fluss/server/zk/ZooKeeperClient.java
```diff
@@ -418,6 +418,11 @@ public int getPartitionNumber(TablePath tablePath) throws Exception {
         return stat.getNumChildren();
     }
 
+    public void createPartitionsPath(TablePath tablePath) throws Exception {
```
luoyuxia: Seems this method is not used?

luoyuxia (commented):
@wuchong Just one minor comment. LGTM.
I believe that the table maybe re-created after the checking is very rarely and it should resolve all most all cases.

fluss-server/src/main/java/com/alibaba/fluss/server/zk/ZooKeeperClient.java
wuchong: Good catch!

wuchong:
@luoyuxia , I have addressed the comments and resolved the conflicts. Please take another look.

luoyuxia (approved):
LGTM！

PR #780, opened 2025-04-22
[lake] Introduce tiering assignment manager to CoordinatorServer

### LINKED ISSUE #433: Implement task allocation logic in Fluss Server to assign synchronization tasks to Tiering Service
### Search before asking

### Description

 Implement task allocation logic in Fluss Server to assign synchronization tasks to Tiering Service

### Willingness to contribute

luoyuxia:
### Purpose

Linked issue: close #433 

### Brief change log

- Introduce a table option `table.datalake.tiering.interval` to control the data fressness tiered from fluss in datalake.
- Introduce a `LakeTableTIeringManager` to coordinator server to manage the lifecycle of the lake table to be tiered, the state macheine looks as follows:
```
New -> Scheduled
Scheduled --> |after lake tiering interval| Pending
Pending --> |lake tiering service request table| Tiering
Tiering --> |once lake tiering finish| Scheduled
Tiering --> |lake tiering heartbeat timout| Scheduled
```

Then, LakeTiering can requet table from `LakeTableTIeringManager`:
1: After requet a table, it wil send heartbeat  periodically to tell coordiantor the table is still tiering. 
2: If the table hasn't been notified to be alive for a while, coordinator will assign the table to another tiering service
3: After tiering service finish one round of tiering for the table, notify the table has finished the tiering via heartbeat. Then `LakeTableTIeringManager` will wait `table.datalake.tiering.interval` to assign the table to other tiering service

### Tests

LakeTableTieringManagerTest.
Let's add IT after we introduce RPC in 432


fluss-server/src/test/java/com/alibaba/fluss/server/coordinator/LakeTableTieringManagerTest.java
```diff
@@ -0,0 +1,214 @@
...
+                    manualClock.advanceTime(Duration.ofSeconds(10));
+                    assertRequestTable(tableId2, tablePath2);
+                });
+
+        retry(
+                Duration.ofMinutes(1),
+                () -> {
+                    // advance a large time to mock tiering service heartbeat timeout
+                    // and check the request table, the table1 should be re-scheduled
+                    manualClock.advanceTime(Duration.ofMinutes(10));
```
wuchong: move it before `retry` to avoid the `advanceTime` to be called multiple times.

fluss-server/src/test/java/com/alibaba/fluss/server/coordinator/LakeTableTieringManagerTest.java
```diff
@@ -0,0 +1,214 @@
...
+
+        // advance time and mock tiering service heartbeat
+        manualClock.advanceTime(Duration.ofMillis(TIERING_SERVICE_TIMEOUT_MS - 1));
+        // tableid1 renew the livness, so that it won't be re-scheduled after heartbeat timeout
+        tableTieringManager.renewTieringLiveness(tableId1);
+        // should only get table2
+        retry(
+                Duration.ofSeconds(10),
+                () -> {
+                    manualClock.advanceTime(Duration.ofSeconds(10));
```
wuchong: move it before `retry` to avoid the `advanceTime` to be called multiple times.

fluss-server/src/main/java/com/alibaba/fluss/server/coordinator/LakeTableTieringManager.java
```diff
@@ -0,0 +1,252 @@
...
+        inLock(
+                lock,
+                () -> {
+                    tablePathById.remove(tableId);
+                    tieringIntervalByTableId.remove(tableId);
+                });
+    }
+
+    @Nullable
+    public Tuple2<Long, TablePath> requestTable() {
```
wuchong: This method is not thread-safe. `TimerTask` will modify the member fields in another thread. The implementation of `LakeTableTieringManager` should be thread-safe.

fluss-server/src/main/java/com/alibaba/fluss/server/coordinator/LakeTableTieringManager.java
```diff
@@ -0,0 +1,252 @@
...
+ * New -> Scheduled
+ * Scheduled --> |after lake tiering interval| Pending
+ * Pending --> |lake tiering service request table| Tiering
+ * Tiering --> |once lake tiering finish| Scheduled
+ * Tiering --> |lake tiering heartbeat timout| Scheduled
+ * }</pre>
+ */
+public class LakeTableTieringManager implements AutoCloseable {
+
+    protected static final long TIERING_SERVICE_TIMEOUT_MS = 3 * 60 * 1000; // 3 minutes
```
wuchong: The timeout should be larger than the default tiering interval? e.g., 10min?

fluss-server/src/main/java/com/alibaba/fluss/server/coordinator/LakeTableTieringManager.java
```diff
@@ -0,0 +1,252 @@
...
+        TablePath tablePath = tablePathById.get(tableId);
+        // the table has been dropped, request again
+        if (tablePath == null) {
+            return requestTable();
+        }
+        tieringLivenessProbe(tableId);
+        return new Tuple2<>(tableId, tablePath);
+    }
+
+    private void tieringLivenessProbe(long tableId) {
```
wuchong: This method is not easy to understand. IIUC, this method just registers a timeout timer for the tiering table. Then we can call it `registerTieringTimeout()`.

fluss-server/src/main/java/com/alibaba/fluss/server/coordinator/LakeTableTieringManager.java
```diff
@@ -0,0 +1,252 @@
...
+                new TimerTask(TIERING_SERVICE_TIMEOUT_MS) {
+                    @Override
+                    public void run() {
+                        inLock(
+                                lock,
+                                () -> {
+                                    // the table is not in tiering, we put it back to wait to be
+                                    // scheduled
+                                    if (!liveTieringTableIds.remove(tableId)) {
+                                        scheduleTableTiering(tableId);
```
wuchong: should add to queue again? instead of waiting another interval?

fluss-server/src/main/java/com/alibaba/fluss/server/coordinator/LakeTableTieringManager.java
```diff
@@ -0,0 +1,252 @@
...
+                                    } else {
+                                        // schedule liveness probe again
+                                        tieringLivenessProbe(tableId);
+                                    }
```
wuchong: This logic is hard to understand. What's the purpose of this method? Detect no-heartbeat tiering table or stop tiering timeout table? 
- If detect no-heartbeat tiering table, it would be better to add `lastHeartbeatTime` in the `liveTieringTableIds` for each table. So it's easy to judge for the timeout timer. 
- If stop tiering timeout table, I think we don't need to "schedule liveness probe again", just send STOP event to the flink job.

fluss-server/src/main/java/com/alibaba/fluss/server/coordinator/LakeTableTieringManager.java
```diff
@@ -0,0 +1,252 @@
...
+                            pendingTieringTables.add(tableId);
+                        }
+                    });
+        }
+    }
+
+    private class ExpiredOperationReaper extends ShutdownableThread {
+
+        public ExpiredOperationReaper() {
+            super("ExpiredOperationReaper", false);
```
wuchong: "LakeTieringExpiredOperationReaper" to distinguish with the name of `DelayedOperationManager.ExpiredOperationReaper`.

fluss-server/src/main/java/com/alibaba/fluss/server/coordinator/LakeTableTieringManager.java
```diff
@@ -0,0 +1,252 @@
...
+ * be tiered by tiering services.
+ *
+ * <p>The state machine of table to be tiered is as follows:
+ *
+ * <pre>{@code
+ * New -> Scheduled
+ * Scheduled --> |after lake tiering interval| Pending
+ * Pending --> |lake tiering service request table| Tiering
+ * Tiering --> |once lake tiering finish| Scheduled
+ * Tiering --> |lake tiering heartbeat timout| Scheduled
```
wuchong: Add description for each status. Besides, I didn't see any state machine for the tables. I think that's important to judge what's the situation of the current table.

fluss-server/src/main/java/com/alibaba/fluss/server/coordinator/LakeTableTieringManager.java
luoyuxia: Yes, it should be thread safe, forget to add lock.

fluss-server/src/main/java/com/alibaba/fluss/server/coordinator/LakeTableTieringManager.java
luoyuxia: I agree it'll be better to put to queue instead of waiting another interval to make tier happen asap

fluss-server/src/main/java/com/alibaba/fluss/server/coordinator/LakeTableTieringManager.java
luoyuxia: It's for detect no-heartbeat tiering table. It's good suggestion to add  `lastHeartbeatTime` to make it more clear.

fluss-server/src/main/java/com/alibaba/fluss/server/coordinator/LakeTableTieringManager.java
luoyuxia: Let's use 5 minutes. I'm afraid of 10min too long for timeout...

fluss-server/src/main/java/com/alibaba/fluss/server/coordinator/LakeTableTieringManager.java
luoyuxia: Now, I add the state machine to track state transmition and current state.

fluss-server/src/main/java/com/alibaba/fluss/server/coordinator/LakeTableTieringManager.java
wuchong: Now, it's a timeout for heatbeat, then it's better to be smaller. I think 1 minute or 2 minute is fine.

fluss-server/src/main/java/com/alibaba/fluss/server/coordinator/LakeTableTieringManager.java
```diff
@@ -0,0 +1,426 @@
...
+        this.pendingTieringTables = new ArrayDeque<>();
+        this.tieringStates = new HashMap<>();
+        this.liveTieringTableIds = new HashMap<>();
+        this.tablePathById = new HashMap<>();
+        this.tieringIntervalByTableId = new HashMap<>();
+        this.expirationReaper = new LakeTieringExpiredOperationReaper();
+        expirationReaper.start();
+    }
+
+    public void initWithLakeTables(List<TableInfo> lakeTables) {
```
wuchong: Add lock

fluss-server/src/main/java/com/alibaba/fluss/server/coordinator/LakeTableTieringManager.java
```diff
@@ -0,0 +1,426 @@
...
+        for (TableInfo tableInfo : lakeTables) {
+            doHandleStateChange(tableInfo.getTableId(), TieringState.ReInitialized);
+            addLakeTableToTier(tableInfo);
+        }
+    }
+
+    public void addNewLakeTable(TableInfo tableInfo) {
```
wuchong, line 179: Add lock

fluss-server/src/main/java/com/alibaba/fluss/server/coordinator/LakeTableTieringManager.java
```diff
@@ -0,0 +1,426 @@
...
+        scheduleTableTiering(tableId, null);
+    }
+
+    private void scheduleTableTiering(long tableId, @Nullable Long delayMs) {
+        Long tieringInterval = tieringIntervalByTableId.get(tableId);
+        if (tieringInterval == null) {
+            // the table has been dropped, return directly
+            return;
+        }
+        lakeTieringScheduleTimer.add(
```
wuchong: We should add it into the tiering queue directly if the last committed snapshot exceeds the tiering interval. Currently, if the coordinator restart, it adds a tiering interval to all the tables which is not necessary.

fluss-server/src/main/java/com/alibaba/fluss/server/coordinator/LakeTableTieringManager.java
```diff
@@ -0,0 +1,426 @@
...
+                    if (tablePath == null) {
+                        return requestTable();
+                    }
+                    registerTieringTimeout(tableId);
+                    doHandleStateChange(tableId, TieringState.Tiering);
+                    return new Tuple2<>(tableId, tablePath);
+                });
+    }
+
+    private void registerTieringTimeout(long tableId) {
```
wuchong: When the tiering process finishes, the timer task is not canceled—it feels like there might be an issue with that. Since `liveTieringTableIds` is cleared immediately when tiering completes, would it be simpler to have a separate thread periodically check the heartbeat of `liveTieringTableIds`?

fluss-server/src/main/java/com/alibaba/fluss/server/coordinator/LakeTableTieringManager.java
```diff
@@ -0,0 +1,426 @@
...
+                () ->
+                        // schedule it again if one round of tiering is finished
+                        doHandleStateChange(tableId, TieringState.Scheduled));
+    }
+
+    public void renewTieringHeartbeat(long tableId) {
+        inLock(
+                lock,
+                () -> {
+                    liveTieringTableIds.put(tableId, clock.milliseconds());
```
wuchong, line 310: We must check the tiering state, and the state must be `Tiering`, otherwise, it should return/throw exception messages to allow the tiering job to cancel the data movement.

fluss-server/src/main/java/com/alibaba/fluss/server/coordinator/LakeTableTieringManager.java
```diff
@@ -0,0 +1,426 @@
...
+        return lastHeartbeat == null
+                || clock.milliseconds() - lastHeartbeat >= TIERING_SERVICE_TIMEOUT_MS;
+    }
+
+    public void finishTableTiering(long tableId) {
+        inLock(
+                lock,
+                () ->
+                        // schedule it again if one round of tiering is finished
+                        doHandleStateChange(tableId, TieringState.Scheduled));
```
wuchong: It would be better to add a `Tiered` state, so the state transition becomes `Tiering → Tiered → Scheduled`.
Because transitioning directly from `New/Initialized → Scheduled` is invalid and this can happen in case of a zombie flink job comes back.

Similarity, we should add `Failed` state, to allow the job reports failing and let other jobs pick the table again.

fluss-server/src/main/java/com/alibaba/fluss/server/coordinator/LakeTableTieringManager.java
```diff
@@ -0,0 +1,426 @@
...
+                // do nothing
+                break;
+            case Scheduled:
+                if (currentState == TieringState.New) {
+                    scheduleTableTiering(tableId);
+                } else if (currentState == TieringState.ReInitialized) {
+                    // schedule the table, but we wait for a period of TIMEOUT
+                    // to avoid the existing tiering services are still tiering it since
+                    // it's a table reinitialized by coordinator server
+                    scheduleTableTiering(tableId, TIERING_SERVICE_TIMEOUT_MS);
```
wuchong: We don't need to wait another interval. Because zombie flink job comes back after the timeout, we have to handle the case of two flink job (one is the zombie flink job) both handling the same table. 

We can handle this using epoch mechanism. There are 2 epochs, one is the coordinator epoch which is already there. The other is the table epoch which is increased for every adding into tiering queue. 

The heartbeat report and finish report should check both epoch is equal, otherwise, should response errors.

fluss-server/src/main/java/com/alibaba/fluss/server/coordinator/LakeTableTieringManager.java
```diff
@@ -0,0 +1,426 @@
...
+                    // schedule it again
+                    liveTieringTableIds.remove(tableId);
+                    scheduleTableTiering(tableId);
+                }
+                break;
+            case Pending:
+                pendingTieringTables.add(tableId);
+                break;
+            case Tiering:
+                // do nothing
```
wuchong, line 395: As I commented on `registerTieringTimeout` method, we should add initial heartbeat (current system time) to `liveTieringTableIds`, and let the timer thread scan the `liveTieringTableIds` to check the timeout tables periodically.

fluss-server/src/main/java/com/alibaba/fluss/server/coordinator/LakeTableTieringManager.java
```diff
@@ -0,0 +1,426 @@
...
+            // when target state is new or reinitialized, it's valid when current state is null
+            return curState == null;
+        }
+        return targetState.validPreviousStates().contains(curState);
+    }
+
+    private void doStateChange(long tableId, TieringState fromState, TieringState toState) {
+        tieringStates.put(tableId, toState);
+        LOG.debug(
+                "Successfully changed state for table {} from {} to {}.",
```
wuchong:
```suggestion
                "Successfully changed tiering state for table {} from {} to {}.",
```

fluss-server/src/main/java/com/alibaba/fluss/server/coordinator/LakeTableTieringManager.java
luoyuxia: Yes, it was just to simplify the logic. But seems add the logic is not complex, now, I added the logic in the TieringManager but left it in next pr to record snapshot commit timestamp.

fluss-server/src/main/java/com/alibaba/fluss/server/coordinator/LakeTableTieringManager.java
luoyuxia: Good suggestion to have a separate thread periodically check

fluss-server/src/main/java/com/alibaba/fluss/server/coordinator/LakeTableTieringManager.java
luoyuxia: I was intending to not handle two flink jobs are writing paimon files, but do conflict detect while two flink jobs commiting snapshot to Fluss.

It sounds good to me to introduce a table tiering epoch to make tiering service fail fast, so that the tiering writing for zombie flink job can be refused to save resouces(flink resources, file resouces).

luoyuxia:
@wuchong Comments addressed.

wuchong:
@luoyuxia  please take a look the modified changes.

fluss-server/src/main/java/com/alibaba/fluss/server/coordinator/LakeTableTieringManager.java
```diff
@@ -79,16 +76,25 @@
...
- * Initialized --> |the interval from last lake snapshot is not less than tiering interval| Pending
- * Initialized --> |the interval from last lake snapshot is less than tiering interval| Scheduled
- * Scheduled --> |after waiting for a period of tiering interval| Pending
- * Pending --> |lake tiering service request table| Tiering
- * Tiering --> |once lake tiering finish| Tiered
- * Tiering --> |lake tiering heartbeat timout| Pending
- * Tiering --> |lake tiering report tiering| Failed
- * Tiered -> Scheduled
- * Failed -> Pending
+ * ┌─────┐ ┌──────┐
```
luoyuxia, line 79: This digram looks cool!

luoyuxia (commented):
@wuchong Thanks for the suggestion commits, LGTM!

PR #786, opened 2025-04-23
[server] Support generating rack aware bucket assignment when creating table

### LINKED ISSUE #785: Support generating rack aware bucket assignment when creating table
### Search before asking

### Motivation

Currently, Fluss only support generate bucket assignment info using round-robin strategy. However, in some cases, like deploy on K8s, we need to avoid multi replicas of one bucket was deployed to some host. So, we need to introduce one rack aware bucket assignment strategy.

### Solution

_No response_

### Anything else?

_No response_

### Willingness to contribute

swuferhong:
### Purpose

Linked issue: close #785 

Currently, Fluss only support generate bucket assignment info using round-robin strategy. However, in some cases, like deploy on K8s, we need to avoid multi replicas of one bucket was deployed to same host. 
So, in this pr I will introduce one rack aware bucket assignment strategy. For example:

If we have a tabletServer set as follow:
ts-id  -> machine rack
0 -> "rack1"
1 -> "rack3"
2 -> "rack3"
3 -> "rack2"
4 -> "rack2"
5 -> "rack1"

First we will  create a rack alternated tabletServers list:

[0, 3, 1, 5, 4, 2]

the list is in order rack1 -> rack2 -> rack3 -> rack1 -> rack2 -> rack3, and the preview tabletServerId lower than the next tabletServerId for the same rack.

Then an easy round-robin assignment can be applied. Assume 6 buckets with replica factor of 3, the assignment will be:

bucket0 -> 0,3,1
bucket1 -> 3,1,5
bucket2 -> 1,5,4
bucket3 -> 5,4,2
bucket4 -> 4,2,0
bucket5 -> 2,0,3

Once it has completed the first round-robin, if there are more buckets to assign, the algorithm will start shifting the followers. This is to ensure we will not always get the same set of sequences. In this case, if there is another bucket to assign (bucket6, bucket7), the assignment will be: 

bucket6 -> 0,4,2 (instead of repeating 0,3,1 as bucket0)
bucket7 -> 3,2,0  (instead of repeating 3,1,5 as bucket1)

### DIFF
--- a/fluss-server/src/test/java/com/alibaba/fluss/server/utils/TableAssignmentUtilsTest.java
+++ b/fluss-server/src/test/java/com/alibaba/fluss/server/utils/TableAssignmentUtilsTest.java
@@ -16,40 +16,84 @@
 
 package com.alibaba.fluss.server.utils;
 
+import com.alibaba.fluss.cluster.TabletServerInfo;
 import com.alibaba.fluss.exception.InvalidBucketsException;
 import com.alibaba.fluss.exception.InvalidReplicationFactorException;
+import com.alibaba.fluss.exception.InvalidServerRackInfoException;
 import com.alibaba.fluss.server.zk.data.BucketAssignment;
 import com.alibaba.fluss.server.zk.data.TableAssignment;
 
 import org.junit.jupiter.api.Test;
 
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.Comparator;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import java.util.stream.Collectors;
+
+import static com.alibaba.fluss.server.utils.TableAssignmentUtils.generateAssignment;
+import static com.alibaba.fluss.server.utils.TableAssignmentUtils.getRackAlternatedTabletServerList;
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.assertj.core.api.Assertions.assertThatThrownBy;
 
-/** Test for {@link com.alibaba.fluss.server.utils.TableAssignmentUtils} . */
+/** Test for {@link TableAssignmentUtils} for rack unaware mode and rack aware mode. */
 class TableAssignmentUtilsTest {
 
     @Test
-    void testTableAssignment() {
+    void testTableAssignmentWithRackUnAware() {
         // should throw exception since servers is empty
-        assertThatThrownBy(() -> TableAssignmentUtils.generateAssignment(5, -1, new int[0]))
+        assertThatThrownBy(
+                        () ->
+                                generateAssignment(
+                                        5,
+                                        -1,
+                                        toTabletServerInfo(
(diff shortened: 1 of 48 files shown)


swuferhong:
@wuchong @luoyuxia @loserwang1024 , the pr to support generating rack aware bucket assignment, PTAL, thx.

loserwang1024:
>. In this case, if there is another bucket to assign (bucket6, bucket7), the assignment will be:
> bucket6 -> 0,4,2 (instead of repeating 0,3,1 as bucket0)
> bucket7 -> 3,2,0 (instead of repeating 3,1,5 as bucket1)

 8 bucket( * 3) to 8 servers:
* server 1: 5 buckets
* server 2: 5 buckets
* server 3: 4 buckets
* server4:  4 buckets
* server 5: 3 buckets
* server 6: 3 bucket

It seems not even enough?

swuferhong:
> > . In this case, if there is another bucket to assign (bucket6, bucket7), the assignment will be:
> > bucket6 -> 0,4,2 (instead of repeating 0,3,1 as bucket0)
> > bucket7 -> 3,2,0 (instead of repeating 3,1,5 as bucket1)
> 
> 8 bucket( * 3) to 8 servers:
> 
> * server 0: 5 buckets
> * server 1: 5 buckets
> * server 2: 4 buckets
> * server 3:  4 buckets
> * server 4: 3 buckets
> * server 5: 3 bucket
> 
> 8 bucket( * 3) to 3 racks:
> 
> * rack 1 (server 0 & 5): 8 bucket
> * rack 2 (server 3 & 4): 7 bucket
> * rack 3( server 1 & 2): 9 bucket
> 
> It seems not even enough for servers and racks?

You must have made a calculation error, right? According to the above algorithm, when there are 8 buckets, the distribution result should be:
[0, 3, 1]
[3, 1, 5]
[1, 5, 4]
[5, 4, 2]
[4, 2, 0]
[2, 0, 3]
[0, 4, 2]
[3, 2, 0]

==>
server0: 5
server1:  3
server2:  5
server3:  4
server4:  4
server5:  3

=>
 rack 1 (server 0 & 5):  8
rack 2 (server 3 & 4):  8
rack 3( server 1 & 2):  8

swuferhong:
@wuchong Already rebase main branch. Pr ready.

fluss-server/src/test/java/com/alibaba/fluss/server/utils/TableAssignmentUtilsTest.java
```diff
@@ -87,4 +141,488 @@ void testTableAssignment() {
...
+                        0);
+        checkTableAssignment(
+                tableAssignment, rackMap, 6, nBuckets, replicationFactor, true, false, false);
+    }
+
+    @Test
+    void testAssignmentWithRackAwareWithUnevenRacks() {
+        Map<Integer, String> rackMap = new HashMap<>();
+        rackMap.put(0, "rack1");
+        rackMap.put(1, "rack2");
```
wuchong: This should be `rack1` to simulate the uneven racks.

Issue #750, opened 2025-04-16
Introduce OffsetsForLeaderEpochRequest/Response


swuferhong:
### Search before asking

- [x] I searched in the [issues](https://github.com/alibaba/fluss/issues) and found nothing similar.

### Motivation

To use LeaderEpoch rather than HighWatermark for truncation, we need to introduce `OffsetsForLeaderEpochRequest/Response` to fetch the latest `LeaderEpoch -> Offset` info from leader to truncate log.

### Willingness to contribute

- [x] I'm willing to submit a PR!

PR #788, opened 2025-04-24
[log] Introduce OffsetsForLeaderEpoch Request/Response

swuferhong:
### Purpose

Linked issue:  #750 

To use LeaderEpoch rather than HighWatermark for truncation, we need to introduce OffsetsForLeaderEpochRequest/Response to fetch the latest LeaderEpoch -> Offset info from leader to truncate log.


swuferhong:
@wuchong  pr ready, could you take a look at this pr? thx.

PR #794, opened 2025-04-24
[server] feat: eventQueueTime metrics in coordinator

### LINKED ISSUE #791: Introduce eventQueueTime in coordiantor
### Search before asking

### Motivation

in #711 , we introduce `eventQueueSize`, `eventProcessTimeMs`, but I think we should introduce `eventQueueSize` so that we can know the event process delay to minitor the event can be processed in time. 

### Solution

_No response_

### Anything else?

_No response_

### Willingness to contribute

gyang94:
### Purpose

Linked issue: close #791 

### Brief change log

1. add two fields `enqueueTimeMs` and `dequeueTimeMs` into `CoordinatorEvent`, to record the timestamp of enqueue and dequeue. To achieve this, I have to change the CoordinatorEvent interface to be an abstract class.

2. set the `enqueueTimeMs` when the event is put into the queue.  set the `dequeueTimeMs` when the event is taken out of the queue. At the same time, expose the eventQueue metrics by `dequeueTimeMs - enqueueTimeMs`.

### DIFF
--- a/fluss-server/src/main/java/com/alibaba/fluss/server/coordinator/event/CoordinatorEventManager.java
+++ b/fluss-server/src/main/java/com/alibaba/fluss/server/coordinator/event/CoordinatorEventManager.java
@@ -53,6 +53,7 @@ public final class CoordinatorEventManager implements EventManager {
 
     // metrics
     private Histogram eventProcessTime;
+    private Histogram eventQueueTime;
 
     private static final int WINDOW_SIZE = 1024;
 
@@ -69,6 +70,11 @@ public final class CoordinatorEventManager implements EventManager {
         eventProcessTime =
                 coordinatorMetricGroup.histogram(
                         MetricNames.EVENT_PROCESS_TIME_MS,
+                        new DescriptiveStatisticsHistogram(WINDOW_SIZE));
+
+        eventQueueTime =
+                coordinatorMetricGroup.histogram(
+                        MetricNames.EVENT_QUEUE_TIME_MS,
                         new DescriptiveStatisticsHistogram(WINDOW_SIZE));
     }
 
@@ -91,6 +97,7 @@ public final class CoordinatorEventManager implements EventManager {
                 putLock,
                 () -> {
                     try {
+                        event.setEnqueueTimeMs(System.currentTimeMillis());
                         queue.put(event);
                     } catch (InterruptedException e) {
                         LOG.error("Fail to put coordinator event {}.", event, e);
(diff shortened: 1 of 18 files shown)


fluss-server/src/main/java/com/alibaba/fluss/server/coordinator/event/CoordinatorEventManager.java
```diff
@@ -91,6 +97,7 @@ public void put(CoordinatorEvent event) {
                 putLock,
                 () -> {
                     try {
```
luoyuxia, line 99: There's no need to change CoordinatorEvent, I suggest to change like that
```
private class CoordinatorEventThread extends ShutdownableThread {

        public CoordinatorEventThread(String name) {
            super(name, false);
        }

        @Override
        public void doWork() throws Exception {
            QueuedEvent queuedEvent = queue.take();
            CoordinatorEvent coordinatorEvent = queuedEvent.event;
            long eventStartProcessTimeMs = System.currentTimeMillis();
            try {
                if (!(coordinatorEvent instanceof ShutdownEventThreadEvent)) {
                    eventQueueTime.update(System.currentTimeMillis() - queuedEvent.enqueueTimeMs);
                    eventProcessor.process(coordinatorEvent);
                }
            } catch (Throwable e) {
                log.error("Uncaught error processing event {}.", coordinatorEvent, e);
            } finally {
                long eventFinishProcessTimeMs = System.currentTimeMillis();
                eventProcessTime.update(eventFinishProcessTimeMs - eventStartProcessTimeMs);
            }
        }
    }

    private static class QueuedEvent {
        private final CoordinatorEvent event;
        private final long enqueueTimeMs;

        public QueuedEvent(CoordinatorEvent event, long enqueueTimeMs) {
            this.event = event;
            this.enqueueTimeMs = enqueueTimeMs;
        }
    }
```

fluss-server/src/main/java/com/alibaba/fluss/server/coordinator/event/CoordinatorEventManager.java
```diff
@@ -53,6 +53,7 @@ public final class CoordinatorEventManager implements EventManager {
 
     // metrics
     private Histogram eventProcessTime;
+    private Histogram eventQueueTime;
 
     private static final int WINDOW_SIZE = 1024;
```
luoyuxia: I'd like to change `WINDOW_SIZE` to 100 since the coordinator event don't happens very frequently.
```
// since coordinator event don't happens very frequently,
    // set to a smaller window size
    private static final int WINDOW_SIZE = 100;
```

luoyuxia (commented):
@gyang94 Thanks for the pr. Left some comments

gyang94:
@luoyuxia updated. Please check.

luoyuxia (approved):
LGTM!

PR #800, opened 2025-04-27
make flink sink generic

polyzos:


fluss-flink/fluss-flink-common/src/main/java/com/alibaba/fluss/flink/sink/FlinkSink.java
```diff
@@ -44,13 +44,13 @@
 import static com.alibaba.fluss.flink.utils.FlinkConversions.toFlussRowType;
 
 /** Flink sink for Fluss. */
-class FlinkSink implements Sink<RowData>, SupportsPreWriteTopology<RowData> {
+class FlinkSink<InputT> implements Sink<RowData>, SupportsPreWriteTopology<InputT> {
```
wuchong: The `FlinkSink` is still not generic, as it bounds to `Sink<RowData>` that only accepts a `DataStream<RowData>` as input.

wuchong:
Converted into Draft as we need to find a generic way to convert POJO into `Internal` with operation type.

fluss-flink/fluss-flink-common/src/main/java/com/alibaba/fluss/flink/sink/writer/FlinkSinkWriter.java
```diff
@@ -120,29 +159,40 @@ protected void initMetrics() {
...
+            RowWithOp rowWithOp = serializationSchema.serialize(inputValue);
+
+            InternalRow internalRow = rowWithOp.getInternalRow();
+            RowKind rowKind = rowWithOp.getRowKind();
 
-                        // Checking for exceptions from previous writes
-                        mailboxExecutor.execute(this::checkAsyncException, "Update error metric");
-                    }
-                });
+            if (ignoreDelete && (rowKind == RowKind.UPDATE_BEFORE || rowKind == RowKind.DELETE)) {
```
polyzos: @wuchong although i made some progress, seems like `UPSERTS` are not only handled on the server.. here we need to check if the RowKind is an `UPDATE_BEFORE` so we can handle the `ignoreDelete`

polyzos:
@wuchong PTAL

fluss-flink/fluss-flink-common/src/main/java/com/alibaba/fluss/flink/row/RowWithOp.java
```diff
@@ -0,0 +1,100 @@
...
+ *
+ * @see InternalRow
+ * @see OperationType
+ */
+public class RowWithOp {
+    /** The internal row data. */
+    private final InternalRow row;
+
+    /** The type of operation associated with this row (e.g., APPEND, UPSERT, DELETE). */
+    @Nullable private final OperationType opType;
```
wuchong: `opType` shouldn't be nullable now, because we provides `IGNORE` operation type to represent the `null` case.

fluss-flink/fluss-flink-common/src/main/java/com/alibaba/fluss/flink/sink/serializer/RowSerializationSchema.java
```diff
@@ -0,0 +1,122 @@
...
+import com.alibaba.fluss.flink.row.RowWithOp;
+import com.alibaba.fluss.row.InternalRow;
+
+import org.apache.flink.table.data.RowData;
+import org.apache.flink.types.RowKind;
+
+import javax.annotation.Nullable;
+
+/** Default implementation of RowDataConverter for RowData. */
+public class RowSerializationSchema implements FlussSerializationSchema<RowData> {
```
wuchong: `RowSerializationSchema` -> `RowDataSerializationSchema` to keep align with the source deserializer `RowDataDeserializationSchema`.

fluss-flink/fluss-flink-common/src/main/java/com/alibaba/fluss/flink/sink/writer/FlinkSinkWriter.java
```diff
@@ -120,35 +146,44 @@ protected void initMetrics() {
...
-                    }
-                });
-
-        numRecordsOutCounter.inc();
+            numRecordsOutCounter.inc();
+        } catch (Exception e) {
+            LOG.error(
+                    "Failed to serialize input value of type '{}': {}",
+                    inputValue.getClass().getName(),
+                    e.getMessage());
```
wuchong: We shouldn't swallow the exception, otherwise, users can't find problems when encountering dirty data or external system issues.

fluss-flink/fluss-flink-common/src/test/java/com/alibaba/fluss/flink/sink/FlinkRowDataChannelComputerTest.java
```diff
@@ -28,15 +36,41 @@
...
+        serializationSchema.open(
+                new FlussSerializationSchema.InitializationContext() {
+                    @Override
+                    public MetricGroup getMetricGroup() {
+                        return null;
+                    }
+
+                    @Override
+                    public UserCodeClassLoader getUserCodeClassLoader() {
+                        return null;
```
wuchong: If we can't provide userClassloader and metricGroup in some cases, we shouldn't expose them to users in `FlussSerializationSchema.InitializationContext`, otherwise, user can't use it because it will break in the case of null. 

Currently, it's fine not expose them to users as they are rarely used.

wuchong (approved):
fixed by appending commits, will merge them when CI is passed.

wuchong:
@polyzos I just found there is no binding issue with this PR. Please link an issue with pull requests in the future so that others can better track the motivation and background of the PR (usually issues has more information about discussion and designs). See the pull request guide: 

https://alibaba.github.io/fluss-docs/community/how-to-contribute/contribute-code/#pull-request-guide

polyzos:
@wuchong its linked [here](https://github.com/alibaba/fluss/issues/569)

wuchong:
@polyzos thank you for the pointer. In this case, you can create an sub-issue under the umbrella issue 569, and link the sub-issue with this PR. We don't link PR with umbrella issues.

wuchong:
The main idea is that each important pull request should be tracked by one corresponding issue, and this issue is only resolved by one PR.

Issue #582, opened 2025-03-11
Add options to config max limit of partition number and bucket number


swuferhong:
### Search before asking

- [x] I searched in the [issues](https://github.com/alibaba/fluss/issues) and found nothing similar.

### Motivation

To support `multi partitioned fields` or `dynamic partition` for fluss partitioned tables, we need to prevent a scenario where an excessive number of `partitions` could give a significant pressure to the server. Therefore, it is necessary to config limits on both the number of partitions and the number of buckets.

### Willingness to contribute

- [ ] I'm willing to submit a PR!

caozhen1937:
@swuferhong , is anyone working for this ticket? If no, please assign this ticket to me.

caozhen1937:
Hi @wuchong sorry for later reply, here is my thougnt:

caozhen1937:
Hi @wuchong sorry for later reply, here is my thougnt:

**For partition**
1. If dynamic partition is configured, partitions total number  is determined when creating the table. These parameters control the total number:
'table.auto-partition.time-unit' = 'YEAR'`
'table.auto-partition.num-precreate' = '5'
'table.auto-partition.num-retention' = '2'

2.  If multi partitioned fields are supported in the future, it can also determine partitions total number based on the number of partitioned fields when creating the table.

3. Implement  1)Add some validation logic in the `TableDescriptorValidation#checkPartition` function
 2)params: table.auto-partition.num-max


**For bucket**
Aame as partiton, add some validation logic in the `TableDescriptorValidation#checkDistribution` function.

wuchong:
@caozhen1937 I separate it into 2 issues. Let's first implement the #687. You can follow the suggested steps in the issue description. Leave comments if you have any problems when implement this feature.

caozhen1937:
@wuchong Thank you, get it, it is cluster level param.

PR #811, opened 2025-04-30
[Server] Introduce Default Max bucket number for a table

### LINKED ISSUE #688: Support max bucket number of a table
### Search before asking

### Description

See parent issue. 

### Willingness to contribute

MehulBatra:
### Purpose
Introduce a cluster-level config, which limits the maximum number of buckets that can be created for a partitioned table to avoid creating too many buckets. Default 128000

Linked issue: close #688, #582

### Brief change log

- New configuration parameter max.bucket.num with default value 128000 to limit the total number of buckets across all partitions of a table
- New exception class TooManyBucketsException extending ApiException to handle bucket limit violations
- Method getTotalBucketNumber(TablePath tablePath) in ZooKeeperClient to count buckets across partitions
- Check in TableDescriptorValidation.checkDistribution() to enforce limit while table creation 
- Bucket count validation in MetadataManager.createPartition() to enforce limits
- Exception handling in AutoPartitionManager to gracefully handle bucket limit exceptions
- MetadataManager constructor to store maxBucketNum as a final field for consistent application of limits
- AutoPartitionManager.createPartitions() to catch and log TooManyBucketsException for auto-partitioning

### Tests

- Added testAddTooManyBuckets() to FlussAdminITCase to verify the API behavior when bucket limits are reached for a partitioned table.
- Added testMaxBucketNum() to AutoPartitionManagerTest to verify that auto-partitioning respects bucket limits in AutoPartitionManagerTest
- Added testBucketLimitForNonPartitionedTable() to FlussAdminITCase to verify the API behavior when bucket limits are reached for a non-partitioned table

### Documentation

- Added JavaDoc comments explaining the bucket limitation behavior
- Added max bucket num limit in the maintenance/configuration.md


MehulBatra:
@wuchong, please help me review it. I have tried to do it the way we discussed in #688
Thank you!

fluss-server/src/main/java/com/alibaba/fluss/server/coordinator/MetadataManager.java
```diff
@@ -387,6 +390,26 @@ public void createPartition(
                     e);
         }
 
+        try {
+            int newPartitionBucketCount = partitionAssignment.getBucketAssignments().size();
+
+            int currentTotalBuckets = zookeeperClient.getTotalBucketNumber(tablePath);
```
wuchong: This is a quite heavy operation and we should avoid it. We can simply verify the current bucket count by `current_partiton_num * table_bucket_num`.

fluss-client/src/test/java/com/alibaba/fluss/client/admin/FlussAdminITCase.java
```diff
@@ -869,4 +870,43 @@ private List<FsPathAndFileName> toFsPathAndFileNames(KvSnapshotHandle kvSnapshot
                                         kvFileHandleAndLocalPath.getLocalPath()))
                 .collect(Collectors.toList());
     }
+
+    /**
+     * Test that creating a table with bucket count exceeding the maximum throws
+     * TooManyBucketsException.
+     */
+    @Test
+    public void testAddTooManyBuckets() throws Exception {
```
wuchong, line 879: Could you add one more test that verifies `TooManyBucketsException` when creating a non-partitioned table with too many buckets?

fluss-server/src/main/java/com/alibaba/fluss/server/coordinator/MetadataManager.java
MehulBatra: Definitely, I did overdo it to be on the safe side, but this seems like quite a light-weight approach.

fluss-client/src/test/java/com/alibaba/fluss/client/admin/FlussAdminITCase.java
MehulBatra: added a test case for non-partitioned table aswell.

MehulBatra:
@wuchong addressed the comments, Please have a look whenever you get a chance.

fluss-server/src/main/java/com/alibaba/fluss/server/zk/ZooKeeperClient.java
```diff
@@ -810,4 +810,28 @@ public void close() {
...
+        }
+
+        int bucketsPerPartition = tableReg.get().bucketCount;
+
+        if (!tableReg.get().isPartitioned()) {
+            // Non-partitioned table - return the bucket count directly
+            return bucketsPerPartition;
+        } else {
+            // Partitioned table - calculate based on partition count
+            int partitionCount = getPartitionNumber(tablePath);
```
wuchong: We already get the partition number in `MetadataManager#createPartition` before the bucket check and no need to get it from zookeeper again here.

fluss-server/src/main/java/com/alibaba/fluss/server/coordinator/MetadataManager.java
```diff
@@ -371,6 +372,7 @@ public void createPartition(
 
         try {
             int partitionNumber = zookeeperClient.getPartitionNumber(tablePath);
+            int maxPartitionNum = conf.get(ConfigOptions.MAX_PARTITION_NUM);
```
wuchong: It's better to remember the `maxPartitionNum` and `maxBucketNum` in member fields to reduce the memory footprint for unnecessary configs and avoid unnecessary redundant get config operations.

wuchong (approved):
LGTM. I pushed a commit to fix the minor comments.

fluss-server/src/main/java/com/alibaba/fluss/server/coordinator/MetadataManager.java
MehulBatra: Initially, I agreed, but later I realized we might need additional configurations. However, the memory footprint is an added cost, so we will maintain two separate variables.

fluss-server/src/main/java/com/alibaba/fluss/server/zk/ZooKeeperClient.java
MehulBatra: My bad, will fix it.

MehulBatra:
Thank you @wuchong !

Issue #837, opened 2025-05-07
Fluss s3/oss filesystem jar shouldn't bundle hadoop-hdfs & hadoop-hdfs-client lib


luoyuxia:
### Search before asking

- [x] I searched in the [issues](https://github.com/alibaba/fluss/issues) and found nothing similar.

### Description

Currently, module `fluss-fs-hadoop` will bundle  `hadoop-hdfs` & `hadoop-hdfs-client` lib, it's needed for accessing hdfs filesystem. But `fluss-fs-s3` and `fluss-fluss-oss` will also include `fluss-fs-hadoop` which cause to bundle `hadoop-hdfs` & `hadoop-hdfs-client`, but for oss and s3, `hadoop-hdfs` & `hadoop-hdfs-client`  is not needed.

We will need to split the moudle `fluss-fluss-hadoop` into two moudles then, one moudle maybe called `fluss-fs-hadoop-common` won't bundle `hadoop-hdfs` & `hadoop-hdfs-client`, and 
 `fluss-fs-s3` and `fluss-fluss-oss` will include `fluss-fs-hadoop-common`, another moudle, maybe called `fluss-fs-hadoop` will bundle `hadoop-hdfs` & `hadoop-hdfs-client`. 

### Willingness to contribute

- [ ] I'm willing to submit a PR!

luoyuxia:
@wuchong Do you have any comment or suggestion about the new moudle name?

wuchong:
@luoyuxia we can separate them into `fluss-fs-hadoop` (the shared common jar, don't need to deploy) and `fluss-fs-hdfs` (the bundled jar).

luoyuxia:
Resolved in #1522

PR #840, opened 2025-05-08
[license] Add NOTICE and LICENSE for published artifacts

### LINKED ISSUE #839: Add NOTICE and LICENSE for published artifaces
### Search before asking

### Description

As part of ensuring compliance with open-source licensing requirements and best practices, we need to include proper LICENSE and NOTICE files in all published artifacts (e.g., JARs).

We can learn from flink-ci-tools that automatically checks and reports the missing NOTICE files and license entries. 

### Willingness to contribute

wuchong:
### Purpose

Linked issue: close #839

### DIFF
--- a/.github/workflows/license-check.yml
+++ b/.github/workflows/license-check.yml
@@ -1,0 +1,54 @@
+# Licensed to the Apache Software Foundation (ASF) under one or more
+# contributor license agreements.  See the NOTICE file distributed with
+# this work for additional information regarding copyright ownership.
+# The ASF licenses this file to You under the Apache License, Version 2.0
+# (the "License"); you may not use this file except in compliance with
+# the License.  You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+name: Check License
+
+on: [push, pull_request]
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event_name }}-${{ github.event.number || github.run_id }}
+  cancel-in-progress: true
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    env:
+      MVN_COMMON_OPTIONS: -U -B --no-transfer-progress
+      MVN_BUILD_OUTPUT_FILE: "/tmp/mvn_build_output.out"
+      MVN_VALIDATION_DIR: "/tmp/fluss-validation-deployment"
+
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Set JDK
+        uses: actions/setup-java@v4
+        with:
+          java-version: 8
+          distribution: 'temurin'
+      - name: Build
+        run: |
+          set -o pipefail
+
+          ./mvnw clean deploy ${{ env.MVN_COMMON_OPTIONS }} -DskipTests \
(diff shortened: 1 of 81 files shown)


fluss-client/pom.xml
```diff
@@ -94,12 +94,13 @@
 
     <build>
         <plugins>
+            <!-- TODO: fluss-clients shouldn't publish bundled jar -->
```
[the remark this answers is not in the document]
wuchong: issue created: https://github.com/alibaba/fluss/issues/841

.github/workflows/license-check.yml
```diff
@@ -0,0 +1,54 @@
...
+        run: |
+          set -o pipefail
+
+          ./mvnw clean deploy ${{ env.MVN_COMMON_OPTIONS }} -DskipTests \
+            -DaltDeploymentRepository=validation_repository::default::file:${{ env.MVN_VALIDATION_DIR }} \
+            | tee ${{ env.MVN_BUILD_OUTPUT_FILE }}
+
+      - name: Check License
+        run: |
+          ./mvnw ${{ env.MVN_COMMON_OPTIONS }} exec:java@check-license -N \
```
SteNicholas, line 50: Why not use apache-rat-plugin to check license?

fluss-client/pom.xml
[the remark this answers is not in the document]
luoyuxia: If not publish `fluss-clients` bundled jar, does it mean that user can't use fluss-client jar directly to access fluss, they have to use flink?

fluss-filesystems/fluss-fs-hadoop/pom.xml
```diff
@@ -46,8 +46,9 @@
         <dependency>
             <groupId>org.apache.hadoop</groupId>
             <artifactId>hadoop-hdfs</artifactId>
-            <!-- to keep same to the haddop version of fluss-fs-hadoop-shaded -->
+            <!-- to keep same to the hadoop version of fluss-fs-hadoop-shaded -->
             <version>${fs.hadoopshaded.version}</version>
+            <scope>provided</scope>
```
luoyuxia, line 51: The `hadoop-hdfs` will still be needed for access hadoop filsystem via our `fluss-fs-hadoop` lib. I rememer at first, the  dependency is with `<optional>true</optional>`, but then we want to access hadoop filsystem via our `fluss-fs-hadoop` lib directly, so I remove the `<optional>true</optional>` to make it include in `hadoop-hdfs` lib.

Maybe we can left a todo to include it again in #837

luoyuxia (approved):
@wuchong Thanks for the greate pr. LGTM. Only two minor comments

fluss-filesystems/fluss-fs-hadoop/pom.xml
wuchong: OK. Let's add it back into `fluss-fs-hdfs` in #837.

fluss-client/pom.xml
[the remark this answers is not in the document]
wuchong: I mean, we shouldn't bundle dependencies in `fluss-client`, but it should still be published to maven central.

.github/workflows/license-check.yml
wuchong: They checks different things. RAT plugin checks license header of Fluss source code. The `check-license` tool checks the necessary NOTICE files (and bundled dependencies in the NOTICE files).

wuchong:
> @wuchong, I noticed that the LICENSE file in the root directory was missing some classes that were copied from other projects. Do we need to add these classes in the LICENSE file?

@SteNicholas I think you are right! Could you open an issue to address this?

wuchong:
Merging...

Issue #583, opened 2025-03-11
Support dynamic partitioning strategy to create partition while writing data if partition not exists


swuferhong:
### Search before asking

- [x] I searched in the [issues](https://github.com/alibaba/fluss/issues) and found nothing similar.

### Motivation

Currently, we only support `manual creating partition` or `auto-partitioning`.  However, in some scenarios, we may not know exactly how many partitions there are for this table, or the partitions cannot be enumerated by predefined rules. In such cases, the capability of `dynamic partitioning` becomes necessary, as it allows us to dynamically create partitions during the write data.

### Willingness to contribute

- [ ] I'm willing to submit a PR!

caozhen1937:
Hi @swuferhong @wuchong ,is this issue working on? if not, I'd like to take it. Can you assign to me?
Over the next few days, I'll be thinking about the implement plan, which I'll discuss here.

wuchong:
thanks @caozhen1937 , let's discuss the implementation desing here first.

caozhen1937:
There have three create partition concept:
1. `manual partitioning`. Create partitions by using `ALTER TABLE table1 ADD PARTITION` syntax.
2. `auto partitioning`. When data is written, a partition is automatically created if it does not exist.
3. `dynamic partitioning`. Automatically maintains (creates and deletes) partitions which are in a specified period of time. The time period is specified by the user when creating a table.

We can refer to Doris's definition: [manual-partitioning](https://doris.apache.org/docs/table-design/data-partitioning/manual-partitioning), [dynamic-partitioning](https://doris.apache.org/docs/table-design/data-partitioning/dynamic-partitioning), [auto-partitioning](https://doris.apache.org/docs/table-design/data-partitioning/auto-partitioning)

In Fluss, only support `manual partitioning` and `dynamic partitioning`(named auto partitioning in Fluss code). 
In this issue, we need support `auto partitioning`.

I think we can devide to two subtask.
1. change `auto partitioning`(already existed) to `dynamic partitioning`.
2. support `auto partitioning`, and be allowed to be used in conjunction with `dynamic partitioning`.
```
auto partitioning, code tracking path
AbstractTableWriter#getPhysicalPath 
    -- MetadataUpdater#checkAndUpdatePartitionMetadata  // client
    -- MetadataUpdater#updateMetadata  // client
        -- RpcServiceBase#getPartitionMetadata  // server
        -- RpcServiceBase#getAssignmentInfo // server
[...]

PR #843, opened 2025-05-09
[server] Support multi-level partition

### LINKED ISSUE #796: Support dynamic partition and support multiple partition columns
### Search before asking

### Motivation

Dynamic partition (i.e. automatically creating non-exist partitions when writing data) is a very useful feature to avoid manual partition management. 

Besides, multi level partitioning(i.e. more than a partition key) is also a common requirement. For example a table organized like ['dt', 'city'] partition can easily lookup data in different date of different city.

### Solution

Since now Fluss has supported automatic manage partitions(i.e. auto create and drop partition with a partition key of time), we'd like to combine these three features to jointly carry out partition management.

A possible practical scenario could be like this:
create a table with multible partition keys:
```
CREATE TABLE site_access(
  event_day STRING,
  site_id INT,
  city STRING,
  user_name STRING
  PRIMARY KEY(event_day, site_id) NOT ENFORCED 
) PARTITIONED BY (event_day, city);
```

With dynamic partition feature, we can directly insert a row without creating partition in advance:
```
INSERT INTO site_access 
    VALUES (20250424, 1, 'Shanghai', 'Eric');
```

And with some configuration of automatic partitions management, expired partition will be automatically dropped.

### Anything else?

_No response_

### Willingness to contribute

### LINKED ISSUE #847: Support multi-level partitions
### Search before asking

### Motivation

Support multi-level partition in server side, and client can create multi-level partition manually (e.g. `ALTER TABLE .. ADD PARTITION ...`)`. 

The support of automatically creation of multi-level partition during writing data is planned in https://github.com/alibaba/fluss/issues/583

### Solution

_No response_

### Anything else?

_No response_

### Willingness to contribute

zcoo:
### Purpose
This pr introduces "multi-level partitioning" and adapt it to automatic partitioning.

Linked issue: close #847


zcoo:
Test failed. I will check it

wuchong:
I changed the associated issue to #847 as this PR doesn't handle the dynamically creation of partition during writing data. I changed the original issue #796 into a parent issue with #847 and #583 as children issues.

fluss-server/src/main/java/com/alibaba/fluss/server/coordinator/AutoPartitionManager.java
```diff
@@ -307,43 +307,53 @@ private void dropPartitions(
...
-                        partitionKeys,
-                        currentZonedDateTime,
-                        -numToRetain,
-                        autoPartitionStrategy.timeUnit());
-
-        Iterator<String> partitionsToExpire =
-                currentPartitions.headSet(lastRetainPartition.getPartitionName(), false).iterator();
-
-        while (partitionsToExpire.hasNext()) {
-            String partitionName = partitionsToExpire.next();
```
wuchong, line 391: The original design is making `currentPartitions` as a sorted list, so it is very efficient to handle large partition set without iterating and decoding partition values. 

Could you create a separate sub-issue under https://github.com/alibaba/fluss/issues/796 to optimize this by refactoring  `currentPartitions` to sort by the partition time key?

fluss-flink/fluss-flink-common/src/test/java/com/alibaba/fluss/flink/catalog/FlinkCatalogITCase.java
```diff
@@ -190,35 +192,40 @@ void testPartitionedTable() throws Exception {
...
         // 2. add partitions.
-        tEnv.executeSql("alter table test_partitioned_table add partition (b = 1)");
-        tEnv.executeSql("alter table test_partitioned_table add partition (b = 2)");
-        tEnv.executeSql("alter table test_partitioned_table add partition (b = 3)");
-        List<String> expectedShowPartitionsResult = Arrays.asList("+I[b=1]", "+I[b=2]", "+I[b=3]");
+        tEnv.executeSql("alter table test_partitioned_table add partition (b = 1,dt = 1)");
+        tEnv.executeSql("alter table test_partitioned_table add partition (b = 2,dt = 1)");
+        tEnv.executeSql("alter table test_partitioned_table add partition (b = 3,dt = 1)");
+        List<String> expectedShowPartitionsResult =
+                Arrays.asList("+I[dt=1/b=1]", "+I[dt=1/b=2]", "+I[dt=1/b=3]");
```
wuchong: This looks like something is wrong. Why the `dt` is displayed before `b` partition value? Theoretically, the partition values should be in the order of partition field definition `b, dt`.

fluss-flink/fluss-flink-common/src/test/java/com/alibaba/fluss/flink/catalog/FlinkCatalogITCase.java
```diff
@@ -279,6 +286,124 @@ void testAutoPartitionedTable() throws Exception {
...
+                                        "create table test_auto_partitioned_table_with_multi_partition_keys (a int, b string, c string, dt string) partitioned by (b,c,dt) "
+                                                + "with ('table.auto-partition.enabled' = 'true',"
+                                                + " 'table.auto-partition.time-unit' = 'day',"
+                                                + " 'table.auto-partition.key' = 'dt',"
+                                                + " 'table.auto-partition.num-precreate' = '2'"
+                                                + ")"))
+                .cause()
+                .isInstanceOf(CatalogException.class)
+                .hasMessage(
+                        "Failed to create table fluss.test_auto_partitioned_table_with_multi_partition_keys in testcatalog");
```
wuchong: Please move all the validation test into a separate test method.

fluss-flink/fluss-flink-common/src/test/java/com/alibaba/fluss/flink/catalog/FlinkCatalogITCase.java
```diff
@@ -279,6 +286,124 @@ void testAutoPartitionedTable() throws Exception {
...
+                                        "create table test_auto_partitioned_table_with_multi_partition_keys (a int, b string, c string, dt string) partitioned by (b,c,dt) "
+                                                + "with ('table.auto-partition.enabled' = 'true',"
+                                                + " 'table.auto-partition.time-unit' = 'day',"
+                                                + " 'table.auto-partition.key' = 'a',"
+                                                + " 'table.auto-partition.num-precreate' = '0'"
+                                                + ")"))
+                .cause()
+                .isInstanceOf(CatalogException.class)
+                .hasMessage(
+                        "Failed to create table fluss.test_auto_partitioned_table_with_multi_partition_keys in testcatalog");
```
wuchong: Please assert the key exception message, e.g., `The specified key for auto partitioned table is not a partition key`.

Currently, all the 3 expected exception message are equal which may hide problems.

fluss-server/src/test/java/com/alibaba/fluss/server/testutils/FlussClusterExtension.java
```diff
@@ -668,6 +668,21 @@ public Map<String, Long> waitUntilPartitionsCreated(TablePath tablePath, int exp
                 "Fail to wait " + expectCount + " partitions created");
     }
 
+    public Map<String, Long> waitUntilPartitionsDrop(TablePath tablePath, int expectCount) {
```
wuchong: The implementation of `waitUntilPartitionsDrop` is identical to `waitUntilPartitionsCreated`. The current implementation is not stable, as the partitions are created first and then dropped. It has chances that meets an intermediate state matches the expected partition count but the excepted expired partitions are not dropped. 

Therefore, would be better to add a `List<String> droppedPartitions` parameter, and assert the remaining partitions don't contain the `droppedPartitions`.

fluss-common/src/main/java/com/alibaba/fluss/config/ConfigOptions.java
```diff
@@ -1073,7 +1082,9 @@ public class ConfigOptions {
                                     + "configured as 3, then partitions 20241111, 20241112, 20241113 will be pre-created. "
                                     + "If any one partition exists, it'll skip creating the partition. "
                                     + "The default value is 2, which means 2 partitions will be pre-created. "
-                                    + "If the 'table.auto-partition.time-unit' is 'DAY'(default), one precreated partition is for today and another one is for tomorrow.");
+                                    + "If the 'table.auto-partition.time-unit' is 'DAY'(default), one precreated partition is for today and another one is for tomorrow."
+                                    + "For a partition table with multiple partition keys, pre-create is unsupported and this value must be strictly less than or equal to 0."
```
wuchong: It is not friendly to force users to set another config. I think we can set it to 0 automatically by server. And add a description here, "For a partition table with multiple partition keys, pre-create is unsupported and will be set to 0 automatically when creating table if it is not explicitly specified."

Regarding implementation, I think we can add the logic in `CoordinatorService#applySystemDefaults` that set the config to 0 if it's multi-level partition table and this config is not explicitly specified.

zcoo:
@wuchong thanks for reviewing! I've fix these problem. PLAT

fluss-server/src/main/java/com/alibaba/fluss/server/coordinator/AutoPartitionManager.java
zcoo: ok. I create #864 to optimize this problem.

fluss-server/src/main/java/com/alibaba/fluss/server/coordinator/CoordinatorService.java
```diff
@@ -271,6 +272,14 @@ private TableDescriptor applySystemDefaults(TableDescriptor tableDescriptor) {
             newDescriptor = newDescriptor.withReplicationFactor(defaultReplicationFactor);
         }
 
+        // override set num-precreate for auto-partition table with multi-level partition keys
+        if ("true".equals(properties.get(ConfigOptions.TABLE_AUTO_PARTITION_ENABLED.key()))
+                && newDescriptor.getPartitionKeys().size() > 1) {
```
wuchong: In my opinion, this change simply introduces a different default value (`0`) for `table.auto-partition.num-precreate`, specifically for multi-level partitioned tables. Since it's just a default value, we only apply this value if the user hasn't explicitly configured `table.auto-partition.num-precreate`.

This approach ensures we don’t override any user-defined settings, while still allowing future support for configuring `table.auto-partition.num-precreate` on multi-level partitioned tables. Otherwise, user can't explicitly set it anymore.

wuchong (approved):
Thanks for the fixing! LGTM. I only left a comment about the overriding logic of `table.auto-partition.num-precreate` and pushed a commit to fix it. 

Besides, please do not use "git merge" to merge branches, otherwise the changes is hard to track and review. Please use "git rebase" to rebase branches instead. IntelliJ IDEA provide an easy tool to do git rebase, you can find the tool via `VCS -> Git -> Rebase`.

PR #850, opened 2025-05-11
[server] feat: authorize InitWriter RPC with table path parameters

### LINKED ISSUE #756: Authorize InitWriter RPC with table path parameters
### Search before asking

### Motivation

Currently, we have to introduce a dedicated IDEMPOTENT_WRITE operation type for authorization because IDEMPOTENT_WRITE is a cluster-level operation that doesn't have table path parameters. This makes the maintenance of IDEMPOTENT_WRITE very hard as we can't easily remove the permission of IDEMPOTENT_WRITE when user don't have any WRITE permission of tables. 

### Solution

We can make `InitWriter` carry table path informations (the tables it will write to), the server checks whether the user has WRITE permission on any of the given tables. If yes, then pass. If not, then error. This can help us to avoid introducing IDEMPOTENT_WRITE operation type, and make maintenance of IDEMPOTENT_WRITE  ACL easy. 

We can get the table path information from `MetadataUpdater` or `RecordAccumulator#writeBatches#keySet()`. Prefer the latter here. 

## Compatibility

For the old client that send `InitWriter` requests without parameters, 
 - if the server enables authorization: throw Auth exception. This doesn't break compatiblity, as old client doens't support auth and this exception is as expected. 
 - if the server doesn't enable authorization: pass

### Willingness to contribute

gyang94:
### Purpose

Linked issue: close #756 

### Brief change log

1. put `tablePath` parameters in `InitWriterRequest` when init writer id.
2. authorize the table paths in tabletServer

### Tests

The `client.writer.enable-idempotence` is defaultly true. Thus the current test cases in `FlinkAuthorizationITCase` will cover this authorization.

### DIFF
--- a/fluss-client/src/main/java/com/alibaba/fluss/client/write/IdempotenceManager.java
+++ b/fluss-client/src/main/java/com/alibaba/fluss/client/write/IdempotenceManager.java
@@ -277,11 +277,11 @@ public class IdempotenceManager {
         return false;
     }
 
-    void maybeWaitForWriterId() {
+    void maybeWaitForWriterId(InitWriterRequest initWriterRequest) {
         if (!isWriterIdValid()) {
             try {
                 tabletServerGateway
-                        .initWriter(new InitWriterRequest())
+                        .initWriter(initWriterRequest)
                         .thenAccept(response -> setWriterId(response.getWriterId()))
                         .exceptionally(
                                 e -> {
--- a/fluss-client/src/main/java/com/alibaba/fluss/client/write/IdempotenceManager.java
+++ b/fluss-client/src/main/java/com/alibaba/fluss/client/write/IdempotenceManager.java
@@ -20,6 +20,7 @@ import com.alibaba.fluss.annotation.Internal;
 import com.alibaba.fluss.annotation.VisibleForTesting;
 import com.alibaba.fluss.exception.OutOfOrderSequenceException;
 import com.alibaba.fluss.exception.UnknownWriterIdException;
+import com.alibaba.fluss.metadata.PhysicalTablePath;
 import com.alibaba.fluss.metadata.TableBucket;
 import com.alibaba.fluss.record.LogRecordBatch;
 import com.alibaba.fluss.rpc.gateway.TabletServerGateway;
@@ -32,6 +33,7 @@ import org.slf4j.LoggerFactory;
 import javax.annotation.concurrent.ThreadSafe;
 
 import java.util.Optional;
+import java.util.Set;
 
 import static com.alibaba.fluss.record.LogRecordBatch.NO_WRITER_ID;
 
@@ -277,11 +279,11 @@ public class IdempotenceManager {
         return false;
     }
 
-    void maybeWaitForWriterId(InitWriterRequest initWriterRequest) {
+    void maybeWaitForWriterId(Set<PhysicalTablePath> tablePaths) {
         if (!isWriterIdValid()) {
             try {
                 tabletServerGateway
-                        .initWriter(initWriterRequest)
+                        .initWriter(prepareInitWriterRequest(tablePaths))
                         .thenAccept(response -> setWriterId(response.getWriterId()))
                         .exceptionally(
                                 e -> {
@@ -295,6 +297,17 @@ public class IdempotenceManager {
                         e);
             }
         }
+    }
+
+    InitWriterRequest prepareInitWriterRequest(Set<PhysicalTablePath> tablePaths) {
+        InitWriterRequest initWriterRequest = new InitWriterRequest();
+        for (PhysicalTablePath tablePath : tablePaths) {
+            initWriterRequest
+                    .addTablePath()
+                    .setDatabaseName(tablePath.getDatabaseName())
+                    .setTableName(tablePath.getTableName());
+        }
+        return initWriterRequest;
     }
 
     private int maybeUpdateLastAckedSequence(TableBucket tableBucket, int sequence) {
--- a/fluss-client/src/main/java/com/alibaba/fluss/client/write/IdempotenceManager.java
+++ b/fluss-client/src/main/java/com/alibaba/fluss/client/write/IdempotenceManager.java
@@ -22,6 +22,7 @@ import com.alibaba.fluss.exception.OutOfOrderSequenceException;
 import com.alibaba.fluss.exception.UnknownWriterIdException;
 import com.alibaba.fluss.metadata.PhysicalTablePath;
 import com.alibaba.fluss.metadata.TableBucket;
+import com.alibaba.fluss.metadata.TablePath;
 import com.alibaba.fluss.record.LogRecordBatch;
 import com.alibaba.fluss.rpc.gateway.TabletServerGateway;
 import com.alibaba.fluss.rpc.messages.InitWriterRequest;
@@ -34,6 +35,7 @@ import javax.annotation.concurrent.ThreadSafe;
 
 import java.util.Optional;
 import java.util.Set;
+import java.util.stream.Collectors;
 
 import static com.alibaba.fluss.record.LogRecordBatch.NO_WRITER_ID;
 
@@ -299,9 +301,13 @@ public class IdempotenceManager {
         }
     }
 
-    InitWriterRequest prepareInitWriterRequest(Set<PhysicalTablePath> tablePaths) {
+    InitWriterRequest prepareInitWriterRequest(Set<PhysicalTablePath> physicalTables) {
         InitWriterRequest initWriterRequest = new InitWriterRequest();
-        for (PhysicalTablePath tablePath : tablePaths) {
+        Set<TablePath> tables =
+                physicalTables.stream()
+                        .map(PhysicalTablePath::getTablePath)
+                        .collect(Collectors.toSet());
+        for (TablePath tablePath : tables) {
             initWriterRequest
                     .addTablePath()
                     .setDatabaseName(tablePath.getDatabaseName())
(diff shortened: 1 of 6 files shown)


fluss-client/src/main/java/com/alibaba/fluss/client/write/Sender.java
```diff
@@ -172,13 +173,25 @@ public void run() {
     public void runOnce() throws Exception {
         if (idempotenceManager.idempotenceEnabled()) {
             // may be wait for writer id.
-            idempotenceManager.maybeWaitForWriterId();
+            idempotenceManager.maybeWaitForWriterId(prepareInitWriterRequest());
```
wuchong: We can move the construction of `InitWriterRequest` into `IdempotenceManager` when we need to send the request (`maybeWaitForWriterId` only accepts a `Set<PhysicalTablePath>` parameter). Currently, it's an overhead to construct the request for each run.

fluss-server/src/main/java/com/alibaba/fluss/server/tablet/TabletService.java
```diff
@@ -313,6 +315,14 @@ public CompletableFuture<ListOffsetsResponse> listOffsets(ListOffsetsRequest req
     @Override
     public CompletableFuture<InitWriterResponse> initWriter(InitWriterRequest request) {
         // todo: add authorization for table acl until https://github.com/alibaba/fluss/issues/756.
```
wuchong: can remove the TODO now.

fluss-server/src/main/java/com/alibaba/fluss/server/tablet/TabletService.java
```diff
@@ -350,11 +360,17 @@ private void authorizeTable(OperationType operationType, long tableId) {
...
                     String.format("This server does not host this table ID %s.", tableId));
         }
+        authorizeTable(operationType, tablePath.getTablePath());
+    }
+
+    private void authorizeTable(OperationType operationType, TablePath tablePath) {
+        if (tablePath == null) {
+            throw new UnknownTableOrBucketException(
+                    "The table path for authorization can not be null.");
+        }
```
wuchong: The `tablePath` shouldn't be null in this stage, the nullable check has been verified in `authorizeTable(OperationType operationType, long tableId)`. Otherwise, `initWriter` will throw `UnknownTableOrBucketException` if `tablePath` is null which is unexpected.

fluss-server/src/main/java/com/alibaba/fluss/server/tablet/TabletService.java
```diff
@@ -313,6 +315,14 @@ public CompletableFuture<ListOffsetsResponse> listOffsets(ListOffsetsRequest req
...
     public CompletableFuture<InitWriterResponse> initWriter(InitWriterRequest request) {
         // todo: add authorization for table acl until https://github.com/alibaba/fluss/issues/756.
+        List<PbTablePath> tablePathsList = request.getTablePathsList();
+        if (tablePathsList != null && !tablePathsList.isEmpty()) {
+            for (PbTablePath pbTablePath : tablePathsList) {
+                authorizeTable(
+                        WRITE,
+                        TablePath.of(pbTablePath.getDatabaseName(), pbTablePath.getTableName()));
+            }
+        }
```
wuchong: 1. if `authorizer` is not enabled, we ignore all the table paths. 
2. if `authorizer` is enabled, only throw `AuthorizationException` if the user has no WRITE permission on all the given tables. So we can't directly use `authorizeTable` method.

wuchong: Besides, we need to add unit test for the cases in `TabletServiceITCase` or a dedicated test class:
 (1) `authorizer` is enabled, but the table path in request is empty (exception) 
 (2) `authorizer` is enabled, and 1 table has write permission, the other doesn't have permission. (pass)
 (3) `authorizer` is enabled, and no table has write permission (exception)
 (4) `authorizer` is not enabled, and the table path in request is empty (pass, the backward compatibility)

fluss-client/src/main/java/com/alibaba/fluss/client/write/IdempotenceManager.java
```diff
@@ -297,6 +299,17 @@ void maybeWaitForWriterId() {
         }
     }
 
+    InitWriterRequest prepareInitWriterRequest(Set<PhysicalTablePath> tablePaths) {
+        InitWriterRequest initWriterRequest = new InitWriterRequest();
+        for (PhysicalTablePath tablePath : tablePaths) {
```
wuchong: we should deduplicate the `TablePath` from the `PhysicalTablePath` collection.

fluss-client/src/main/java/com/alibaba/fluss/client/write/Sender.java
```diff
@@ -172,7 +172,9 @@ public void run() {
     public void runOnce() throws Exception {
         if (idempotenceManager.idempotenceEnabled()) {
             // may be wait for writer id.
-            idempotenceManager.maybeWaitForWriterId();
+            Set<PhysicalTablePath> physicalTablePaths =
+                    accumulator.getPhysicalTablePathsInBatches();
+            idempotenceManager.maybeWaitForWriterId(physicalTablePaths);
```
wuchong: We should only request the writer_id when there are actual table paths in the batch. This helps avoid unnecessary requests when the writer/sender is running but no data has been appended.

fluss-client/src/test/java/com/alibaba/fluss/client/security/acl/FlussAuthorizationITCase.java
```diff
@@ -190,6 +197,14 @@ void testNoAuthorizer() throws Exception {
...
                                 })
                         .hasMessageContaining("No Authorizer is configured.");
+
+                // test initWriter without authorizer and empty table paths
+                FlussConnection flussConnection = (FlussConnection) connection;
+                TabletServerGateway tabletServerGateway =
+                        flussConnection.getMetadataUpdater().newTabletServerClientForNode(0);
+                InitWriterResponse response =
+                        tabletServerGateway.initWriter(new InitWriterRequest()).get();
+                assertThat(response.getWriterId()).isEqualTo(0);
```
wuchong: should use `isGreaterThanOrEqualTo`, because the zk id counter maybe increased by other tests.

fluss-client/src/test/java/com/alibaba/fluss/client/security/acl/FlussAuthorizationITCase.java
```diff
@@ -430,6 +445,75 @@ void testGetMetaInfo() throws Exception {
...
+        request.addTablePath()
+                .setTableName(writeAclTable.getTableName())
+                .setDatabaseName(writeAclTable.getDatabaseName());
+        request.addTablePath()
+                .setTableName(noWriteAclTable.getTableName())
+                .setDatabaseName(noWriteAclTable.getDatabaseName());
+
+        // test 2: one table has write permission, the other doesn't have permission
+        InitWriterResponse response = tabletServerGateway.initWriter(request).get();
+        assertThat(response.getWriterId()).isEqualTo(0);
```
wuchong: ditto

fluss-server/src/main/java/com/alibaba/fluss/server/tablet/TabletService.java
```diff
@@ -312,7 +314,38 @@ public CompletableFuture<ListOffsetsResponse> listOffsets(ListOffsetsRequest req
...
+
+            boolean hasAuthorizedPath = false;
+            for (PbTablePath pbTablePath : tablePathsList) {
+                if (authorizer.isAuthorized(
+                        currentSession(),
+                        WRITE,
+                        Resource.table(
+                                pbTablePath.getDatabaseName(), pbTablePath.getTableName()))) {
+                    hasAuthorizedPath = true;
+                    break;
```
wuchong: We can extract the authorization logic into a private method to simplify the logic.

fluss-server/src/main/java/com/alibaba/fluss/server/tablet/TabletService.java
```diff
@@ -312,7 +314,38 @@ public CompletableFuture<ListOffsetsResponse> listOffsets(ListOffsetsRequest req
 
     @Override
     public CompletableFuture<InitWriterResponse> initWriter(InitWriterRequest request) {
-        // todo: add authorization for table acl until https://github.com/alibaba/fluss/issues/756.
+        List<PbTablePath> tablePathsList = request.getTablePathsList();
+        if (authorizer != null) {
+            if (tablePathsList == null || tablePathsList.isEmpty()) {
+                throw new UnknownTableOrBucketException(
```
wuchong: should throw `AuthorizationException` here, because `UnknownTableOrBucketException` is a retriable exception which indicates clients to retry the request again.

wuchong (approved):
I left some comments and fixed them based on the latest changes.

gyang94:
@wuchong Thanks for your help on fixing.
I have another question: 
Will it happen in real cases which we will init writer with two different table paths? In my understanding, for each table, we will create writer client (with one writer id), then there will be only one table path in the request.

wuchong:
@gyang94 it will happend. The Sender/WriterClient is shared for the `FlussConnection`. But a `FlussConnection` can creates multiple `TableWriter` for different tables.

swuferhong:
> `FlinkAuthorizationITCase#testProduceAndConsumeLogTable` is keeping failing and produce a lot of error logs. This can be reproduced in local. The problem is we didn't raise the error in `IdempotenceManager` to record callbacks. And the flink client is keeping retrying. @swuferhong can help to create a separate pull request to fix this.

Hi @wuchong, @gyang94, I have pushed one commit to fix the retry error: https://github.com/alibaba/fluss/pull/932. pls take a look, thanks.

Issue #489, opened 2025-02-26
Partition key support more filed types


swuferhong:
### Search before asking

- [x] I searched in the [issues](https://github.com/alibaba/fluss/issues) and found nothing similar.

### Motivation

Currently, the partition key only support String type. In this issue we hope to support more filed types for partition key, such as Integer etc.

### Willingness to contribute

- [ ] I'm willing to submit a PR!

MaggieSullivan:
For DateTime we can use ISO 8601 format, which is an accepted international standard. However not every column is a good partition key. For example using a timestamp column without bucketing using something like `"DATE_TRUNC('MONTH', event_timestamp)"` will lead to many small buckets. Do we need to provide a mechanism to transform the partitioned field before using it as a partition? Or is that the responsibility of the user to make their own partition column?

wuchong:
We need to investigate how Hive/Paimon/Iceberg/DeltaLake handle different data types of partition keys.

MaggieSullivan:
Here's my preliminary investigation:

## Iceberg
- Iceberg manages the partition values as hidden partitions derived from real columns. The real columns can be transformed to hidden partition columns managed by Iceberg with partition transforms. Iceberg seems to be the only format with partition transform support.
- [Partition Transforms](https://iceberg.apache.org/spec/#partition-transforms)
- [Github Implementation](https://github.com/apache/iceberg/tree/main/api/src/main/java/org/apache/iceberg/transforms)

## Hive
- Hive style partitioning can be either static or dynamic(looked up by column value). But the column must be user created since no table partition transforms exist. 
- I found parts of the code that might be relevant:
  - https://github.com/apache/hive/tree/master/common/src/java/org/apache/hadoop/hive/common/type
  - https://github.com/apache/hive/blob/master/common/src/java/org/apache/hadoop/hive/common/type/Timestamp.java#L127
  - https://github.com/apache/hive/blob/master/standalone-metastore/metastore-common/src/main/java/org/apache/hadoop/hive/metastore/Warehouse.java#L599

## Paimon
- Same partitioning strategy as Apache Hive, dividing tables into sections based on specific column values. Tables, whether partitioned or not, are further divided into buckets. Buckets are determined by hashing one or more columns. 
- [Paimon partitions and buckets](https://paimon.apache.org/docs/master/learn-paimon/understand-files/#understand-partitions-and-buckets)
[...]

PR #852, opened 2025-05-12
[feature] Partition key support more data types WIP #489

MaggieSullivan:
### Purpose

Linked issue: [open issue #489](https://github.com/alibaba/fluss/issues/489)

### Brief change log

I added transform functionality to support using source columns of different data types as primary key.
This PR is a work in progress, but most of the main functions are there. Since this change involves redesigning PartitionSpec, I would like some early feedback on my approach.

### DIFF
--- a/fluss-common/src/main/java/com/alibaba/fluss/utils/transform/Days.java
+++ b/fluss-common/src/main/java/com/alibaba/fluss/utils/transform/Days.java
@@ -1,0 +1,50 @@
+/*
+ *  Copyright (c) 2024 Alibaba Group Holding Ltd.
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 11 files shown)


wuchong:
Thanks @MaggieSullivan for the draft. However, we should focus on support more data types for partition keys, and then support transformations. They are two separate issues. Support more data types means use `STRING`, `DATE`, `INT`, `BIGINT`, `TIMESTAMP` fields directly as the partition keys. And the problem is how to convert the data into a string representation and the string representation can be compatible as filesystem directory name, object storage path name, zookeeper path name.

MaggieSullivan:
Sounds good! I'll look into that this weekend and thanks for the feedback!

wuchong:
closing this PR for now, and looking forward to the new PR.

PR #862, opened 2025-05-13
[flink] FlinkCatalog supports create and drop materialized table

### LINKED ISSUE #861: FlinkCatalog support to create and drop materialized table
### Search before asking

### Description

FlinkCatalog support to create and drop materialized table

### Willingness to contribute

lsyldliu:
### Purpose

Linked issue: close #861 

### Brief change log

- `FlinkCatalog` supports create/get/drop materialized table
- Since materialized table is a concept introduced by Flink, we convert it into a common Fluss table, introduce relevant options, and save them in `customProperties` to identify it as a materialized table. This is also the current implementation method of Paimon
- Since Paimon already supports materialized tables and has introduced related options, the options introduced in Fluss are aligned with those in Paimon. Since Fluss supports tiering tables to Paimon, when a Fluss materialized table is sync to Paimon, it should also be a Paimon materialized table, so the option names of the two are kept consistent. Reference: https://github.com/apache/paimon/blob/master/paimon-common/src/main/java/org/apache/paimon/CoreOptions.java#L1664

### Tests

- Add unit test in FlinkConversionsTest
- Add unit test in FlinkCatalogTest
- Add integration test in MaterializedTableITCase

### API and Format

No

### Documentation

Yes, we will add the docs after finishing all pr.

### DIFF
--- a/fluss-flink/fluss-flink-1.18/src/main/java/org/apache/flink/table/catalog/CatalogBaseTable.java
+++ b/fluss-flink/fluss-flink-1.18/src/main/java/org/apache/flink/table/catalog/CatalogBaseTable.java
@@ -1,0 +1,115 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.flink.table.catalog;
+
+import org.apache.flink.annotation.PublicEvolving;
+import org.apache.flink.table.api.Schema;
+import org.apache.flink.table.api.TableSchema;
+import org.apache.flink.table.factories.DynamicTableFactory;
+
+import java.util.Map;
+import java.util.Optional;
+
+/**
+ * A common parent that describes the <i>unresolved</i> metadata of a table or view in a catalog.
+ *
+ * <p>Copy from Apache Flink, modified the TableKind enum to add MATERIALIZED_TABLE.
+ *
+ * @see CatalogTable
+ * @see CatalogView
+ */
+@PublicEvolving
+public interface CatalogBaseTable {
+
+    /** The kind of {@link CatalogBaseTable}. */
+    @PublicEvolving
+    enum TableKind {
+        TABLE,
+        MATERIALIZED_TABLE,
+        VIEW
+    }
+
+    /** The kind of table this {@link CatalogBaseTable} describes. */
+    TableKind getTableKind();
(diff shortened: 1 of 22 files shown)


lsyldliu:
cc @wuchong @loserwang1024 
@luoyuxia can you help review materialized table options because the materialized table also needs to tiering to Paimon

fluss-flink/fluss-flink-common/src/main/java/com/alibaba/fluss/flink/catalog/FlinkCatalog.java
```diff
@@ -282,12 +285,22 @@ public CatalogBaseTable getTable(ObjectPath objectPath)
             }
 
             // should be as a fluss table
-            CatalogTable catalogTable = FlinkConversions.toFlinkTable(tableInfo);
```
loserwang1024, line 304: It seems using a custom property to distinguish physical table and materialize table. I wonder:
1.  if user use a lower version of flink (or just use java client), can read and write a materialize table, is there any problem?
2. If fluss support update customer properties, then use drop the properties with prefix MATERIALIZED_TABLE, it seems that a physical table will be transformed as a materialize table. Is there any problem?

fluss-flink/fluss-flink-common/src/main/java/com/alibaba/fluss/flink/utils/FlinkConversions.java
```diff
@@ -185,13 +215,25 @@ public static TableDescriptor toFlussTable(ResolvedCatalogTable catalogTable) {
...
         CatalogPropertiesUtils.serializeComputedColumns(
                 customProperties, resolvedSchema.getColumns());
         CatalogPropertiesUtils.serializeWatermarkSpecs(
-                customProperties, catalogTable.getResolvedSchema().getWatermarkSpecs());
+                customProperties, catalogBaseTable.getResolvedSchema().getWatermarkSpecs());
 
-        String comment = catalogTable.getComment();
+        // Set materialized table flags to fluss table custom properties
+        if (CatalogBaseTable.TableKind.MATERIALIZED_TABLE == tableKind) {
+            serializeMaterializedTableToCustomProperties(
```
loserwang1024, line 240: What if user create a table with materialized customProperties? Should add a check here to prevent it.

loserwang1024 (commented):
I have left some message.

fluss-flink/fluss-flink-common/src/main/java/com/alibaba/fluss/flink/catalog/FlinkCatalog.java
lsyldliu: 1. Due to flink materialized table is a normal fluss table, so we can't control the user action outside flink. So user can use java api to read and write it.
2. We should not support drop the internal options related to materialized tables.

fluss-flink/fluss-flink-common/src/main/java/com/alibaba/fluss/flink/utils/FlinkConversions.java
lsyldliu: This is a good question, but it is unexpected behavior,  currently we can't control the user action,  and these options are internal, user shouldn't see it.

wuchong:
There are still some failed CI.

lsyldliu:
> There are still some failed CI.

Fixed

wuchong (approved):
Thank you @lsyldliu for the contribution, this pull request is in high quality with good test coverage. I don't have any review comments, I only move the materialized table options from `FlinkCatalogOptions` to `FlinkConnectorOptions` as it is a option for tables not for catalogs. 

Will merge it when CI is passed.

PR #863, opened 2025-05-13
Add HeartBeat RPC in Servers for Tiering Services

### LINKED ISSUE #432: [fluss-server] Add TieringService heartbeat gRPC interface for communication with Tiering Services
### Search before asking

### Description

Add TieringService heartbeat gRPC interface for communication with Tiering Services

### Willingness to contribute

luoyuxia:
### Purpose

Linked issue: close #432 

### Brief change log

Introduce RPC request for laketiering heartbeat request:
LakeTieringHeartbeatRequest:
```
message LakeTieringHeartbeatRequest {
  repeated PbHeartbeatReqForTable tiering_tables = 2;
  repeated PbHeartbeatReqForTable finished_tables = 3;
  repeated PbHeartbeatReqForTable failed_tables = 4;
  // whether to request a table
  optional bool request_table = 5;
}
message PbHeartbeatReqForTable {
  // the coordinator epoch when the table has been assigned to be tiering service
  required int64 table_id = 1;
  required int32 coordinator_epoch = 2;
  required int32 tiering_epoch = 3;
}
```

LakeTieringHeartbeatResponse:
```
message PbHeartbeatReqForTable {
  required int64 table_id = 1;
  // the coordinator epoch when the table is assigned to be tiering service
  required int32 coordinator_epoch = 2;
  // the tiering epoch when the table is assigned to be tiering service
  required int32 tiering_epoch = 3;
}

message PbHeartbeatRespForTable {
  required int64 table_id = 1;
  optional ErrorResponse error = 2;
}
```

The `LakeTieringHeartbeatRequest` contain a field `request_table` to tell coordinator server to assign a table to tier. It's designed to only request a table in one round of heartbeat. I'm not sure whether it's needed to request many tables in one round of heartbeat. So, keep it simiple currently. 

Then, the  lake tiering service can carry the tables are tiering, finished table,  failed_tables in the `request` to let coordinator server know the status of tiering.  
The `PbHeartbeatReqForTable` for each table(tiering, finished, failed) in request contains three fields:

- table_id: the table id for the table
- coordinator_epoch: the coordinator epoch when the table is assigned to be tiering service
When the `coordinator_epoch` in the request  is less than the `coordinator_epoch` of the coordinator, it means coordinator move happens but table is assigned by the stale coordinator. It'll throw `InvalidCoordinatorException` to refuse such request to avoid inconsistent.

When the `coordinator_epoch` in the request  is greater than the `coordinator_epoch` of the coordinator, it means it sends to a stale coordinator server, also throw `InvalidCoordinatorException` to refuse such request to avoid inconsistent. 

Note it may still happen that the tiering service send to the stale coordinator, like mark the table as finished and request the table again, but other tiering serevice are still tiering the table which is assgined by the active coordinator , which causes the inconsistent. But it should happen rarely.

- tiering_epoch: the tiering epoch when the table is assigned to be tiering service

When the `tiering_epoch` in request is not match the `tiering_epoch` in coordinator, it means the table has been assigned to other tiering service, throw `FencedTieringEpochException`

For each  `PbHeartbeatReqForTable` of each table in the request, there's one  `PbHeartbeatRespForTable`    for that.

If there is any exception(like `FencedTieringEpochException`) happen when mark tables as `tiering`, `finished`, `failed`, the cooresponding `PbHeartbeatRespForTable` will contains errors. 
The lake tiering service should check and act to the errors:
- If mark table as `tiering`:  if `FencedTieringEpochException`, `InvalidCoordinatorException`, abort the tiering directly since it means other tiering service may be tiering this table. Otherwise, retry to make as `tiering` in next heartbeats.
- if mark table as `finished`: if `FencedTieringEpochException`, `InvalidCoordinatorException`, remove from the finished table from lake tiering service, don't retry to mark  as `finished` in next heartbeats. Otherwise, retry to make as `tiering` in next heartbeats. 
- if mark table as `failed`: it should same as `finished`.

### Tests

`LakeTieringHeartBeatITCase`

### DIFF
--- a/fluss-rpc/src/main/proto/FlussApi.proto
+++ b/fluss-rpc/src/main/proto/FlussApi.proto
@@ -506,6 +506,25 @@ message DropAclsResponse{
   repeated PbDropAclsFilterResult filter_results = 1;
 }
 
+message LakeTieringHeartbeatRequest {
+  repeated PbHeartbeatReqForTable tiering_tables = 2;
+  repeated PbHeartbeatReqForTable finished_tables = 3;
+  repeated PbHeartbeatReqForTable failed_tables = 4;
+  // whether to request a table
+  optional bool request_table = 5;
+}
+
+message LakeTieringHeartbeatResponse {
+  // coordinator epoch
+  required int32 coordinator_epoch = 1;
+  // the returned table to tier, empty when no table is needed by the lake tiering service
+  optional PbLakeTieringTableInfo tiering_table = 2;
+  repeated PbHeartbeatRespForTable tiering_table_resp = 3;
+  repeated PbHeartbeatRespForTable finished_table_resp = 4;
+  repeated PbHeartbeatRespForTable failed_table_resp = 5;
+}
+
+
 // --------------- Inner classes ----------------
 message PbApiVersion {
   required int32 api_key = 1;
@@ -815,4 +834,23 @@ message PbDropAclsMatchingAcl {
   required PbAclInfo acl = 1;
   optional int32 error_code = 2;
   optional string error_message = 3;
+}
+
+message PbLakeTieringTableInfo {
+  required int64 table_id = 1;
+  required PbTablePath table_path = 2;
+  required int64 tiering_epoch = 3;
+}
+
+message PbHeartbeatReqForTable {
+  required int64 table_id = 1;
+  // the coordinator epoch when the table is assigned to be tiering service
+  required int32 coordinator_epoch = 2;
+  // the tiering epoch when the table is assigned to be tiering service
+  required int32 tiering_epoch = 3;
+}
+
+message PbHeartbeatRespForTable {
+  required int64 table_id = 1;
+  optional ErrorResponse error = 2;
 }
--- a/fluss-rpc/src/main/proto/FlussApi.proto
+++ b/fluss-rpc/src/main/proto/FlussApi.proto
@@ -507,11 +507,11 @@ message DropAclsResponse{
 }
 
 message LakeTieringHeartbeatRequest {
-  repeated PbHeartbeatReqForTable tiering_tables = 2;
-  repeated PbHeartbeatReqForTable finished_tables = 3;
-  repeated PbHeartbeatReqForTable failed_tables = 4;
+  repeated PbHeartbeatReqForTable tiering_tables = 1;
+  repeated PbHeartbeatReqForTable finished_tables = 2;
+  repeated PbHeartbeatReqForTable failed_tables = 3;
   // whether to request a table
-  optional bool request_table = 5;
+  optional bool request_table = 4;
 }
 
 message LakeTieringHeartbeatResponse {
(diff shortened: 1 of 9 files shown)


fluss-rpc/src/main/proto/FlussApi.proto
```diff
@@ -506,6 +506,25 @@ message DropAclsResponse{
   repeated PbDropAclsFilterResult filter_results = 1;
 }
 
+message LakeTieringHeartbeatRequest {
+  repeated PbHeartbeatReqForTable tiering_tables = 2;
```
leonardBang: minor: why we start from index 2 here?

fluss-server/src/main/java/com/alibaba/fluss/server/coordinator/CoordinatorService.java
```diff
@@ -133,9 +144,13 @@ public CoordinatorService(
                 authorizer);
         this.defaultBucketNumber = conf.getInt(ConfigOptions.DEFAULT_BUCKET_NUMBER);
         this.defaultReplicationFactor = conf.getInt(ConfigOptions.DEFAULT_REPLICATION_FACTOR);
-        this.eventManagerSupplier = eventManagerSupplier;
+        this.eventManagerSupplier =
```
leonardBang, line 147: I found we have two similar interfaces `EventManager` and `EventProcessor`, maybe we can refactor them or improve `EventManager` later

fluss-server/src/main/java/com/alibaba/fluss/server/coordinator/CoordinatorService.java
```diff
@@ -447,4 +462,74 @@ public CompletableFuture<CommitLakeTableSnapshotResponse> commitLakeTableSnapsho
...
+            }
+        }
+
+        // process failed tables
+        for (PbHeartbeatReqForTable failedTable : request.getFailedTablesList()) {
+            PbHeartbeatRespForTable pbHeartbeatRespForTable =
+                    heartbeatResponse.addFailedTableResp().setTableId(failedTable.getTableId());
+            try {
+                validateHeartbeatRequest(failedTable, currentCoordinatorEpoch);
+                lakeTableTieringManager.reportTieringFail(
```
leonardBang: We can process failed table firstly so that these table can change to pending state ASAP,  for finished and failed tables, we can process later.

fluss-rpc/src/main/proto/FlussApi.proto
luoyuxia: is by mistake

fluss-server/src/main/java/com/alibaba/fluss/server/coordinator/CoordinatorService.java
luoyuxia: `EventManager` is used to put `event` and poll `event`, `EventProcessor` is to process the `event`, more like Kafka's controller.

leonardBang (approved):
Thanks yuxia for the update, LGTM

leonardBang:
CI passed, merging...

PR #869, opened 2025-05-13
[Connector] Resource cleanup via AfterEach Extending AbstractStreamBase

### LINKED ISSUE #868: Unstable flink-core CI getting timed out randomly
### Search before asking

### Fluss version

0.6.0 (latest release)

### Please describe the bug 🐞

Flink-core CI is failing randomly due to resource cleanup from one test iteration starting but not completing before another test iteration accesses the same resources, causing the test case to run indefinitely, eventually causing the CI pipeline to timeout.

### Solution

_No response_

MehulBatra:
### Purpose

Linked issue: close #868

To prevent Flink-core CI from failing randomly due to resource cleanup from one test iteration starting but not completing before another test iteration accesses the same resources, causing the test case to run indefinitely, eventually causing the CI pipeline to timeout.

### Brief change log

- Proper cleanup of resources after each test using `AbstractTestBase` MiniClusterExtension using cleanupRunningJobs
- Explicitly wait for bucket snapshots to complete before collecting results in `testPrefixLookupWithCondition`
- Created a separate utility class `FlinkRowAssertionsUtils` for assertion method, moved away from `FlinkTestBase` to avoid duplicate code, & Improved error handling and closure for the same.
- Removed extensibility of `FlinkTestBase` from `FlinkMetricsITCase`, manually configured Fluss & Flink cluster extension and changes related to it.

 

### Tests

Targeting PrefixLookup type tests

### API and Format

None

### Documentation

None

### DIFF
--- a/fluss-flink/fluss-flink-common/src/test/java/com/alibaba/fluss/flink/source/FlinkTableSourceITCase.java
+++ b/fluss-flink/fluss-flink-common/src/test/java/com/alibaba/fluss/flink/source/FlinkTableSourceITCase.java
@@ -80,10 +80,6 @@ abstract class FlinkTableSourceITCase extends FlinkTestBase {
     @BeforeAll
     protected static void beforeAll() {
         FlinkTestBase.beforeAll();
-    }
-
-    @BeforeEach
-    void before() {
         // create database
         execEnv = StreamExecutionEnvironment.getExecutionEnvironment();
         // create table environment
@@ -97,6 +93,10 @@ abstract class FlinkTableSourceITCase extends FlinkTestBase {
         tEnv.executeSql("use catalog " + CATALOG_NAME);
 
         tEnv.getConfig().set(ExecutionConfigOptions.TABLE_EXEC_RESOURCE_DEFAULT_PARALLELISM, 4);
+    }
+
+    @BeforeEach
+    void before() {
         tEnv.executeSql("create database " + DEFAULT_DB);
         tEnv.useDatabase(DEFAULT_DB);
     }
--- a/fluss-flink/fluss-flink-common/src/test/java/com/alibaba/fluss/flink/source/FlinkTableSourceITCase.java
+++ b/fluss-flink/fluss-flink-common/src/test/java/com/alibaba/fluss/flink/source/FlinkTableSourceITCase.java
@@ -108,6 +108,30 @@ abstract class FlinkTableSourceITCase extends FlinkTestBase {
         tEnv.executeSql(String.format("drop database %s cascade", DEFAULT_DB));
     }
 
+    @AfterEach
+    void cleanupAfterTest() {
+        try {
+            // Cancel any running jobs in the environment
+            if (execEnv != null) {
+                try {
+                    execEnv.close();
+                } catch (Exception e) {
+                    // Log and continue
+                    System.err.println(
+                            "Error during execution environment cleanup: " + e.getMessage());
+                }
+            }
+
+            // Force garbage collection to help clean up lingering connections
+            System.gc();
+
+            // Add a small delay to allow resources to be released
+            Thread.sleep(500);
+        } catch (Exception e) {
+            System.err.println("Error during test cleanup: " + e.getMessage());
+        }
+    }
+
     @Test
     public void testCreateTableLike() throws Exception {
         tEnv.executeSql(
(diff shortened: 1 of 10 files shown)


fluss-flink/fluss-flink-common/src/test/java/com/alibaba/fluss/flink/source/FlinkTableSourceITCase.java
```diff
@@ -108,6 +108,30 @@ void after() {
         tEnv.executeSql(String.format("drop database %s cascade", DEFAULT_DB));
     }
 
+    @AfterEach
+    void cleanupAfterTest() {
+        try {
+            // Cancel any running jobs in the environment
+            if (execEnv != null) {
+                try {
+                    execEnv.close();
```
wuchong: 1. `execEnv` is initialized in `@BeforeAll`, if we close it in `@AfterEach`, it will has problems in next test method that the `execEnv` has been closed. 
2. The `org.apache.flink.streaming.api.environment.StreamExecutionEnvironment#close` doesn't do anything, thus no effect to the test.

MehulBatra: Resolved it by extending to AbstractTestBase using MiniclusterExtension afterEach via cleanUpRunningJobs

fluss-flink/fluss-flink-common/src/test/java/com/alibaba/fluss/flink/source/FlinkTableSourceITCase.java
```diff
@@ -911,6 +911,7 @@ void testPrefixLookupWithCondition(Caching caching, boolean async) throws Except
                                 + " ON src.b = h.name AND h.address = 'address5'",
                         dim);
 
+        waitUtilAllBucketFinishSnapshot(admin, TablePath.of(DEFAULT_DB, dim));
```
wuchong: Why do we need to add this? I think we can still remove this because lookup doesn't rely on snapshots.

wuchong (commented):
`FlinkAuthorizationITCase` is timeout, which also needs to extends `FlinkTestBase` and `AbstractTestBase`.

fluss-flink/fluss-flink-common/src/test/java/com/alibaba/fluss/flink/source/FlinkTableSourceITCase.java
MehulBatra: I added due to table lookup failure warning I saw in the maven logs, to avoid any future warning, but if it is not required I will remove.

MehulBatra:
> `FlinkAuthorizationITCase` is timeout, which also needs to extends `FlinkTestBase` and `AbstractTestBase`.

I will try to solve this today and ping you once done!

MehulBatra:
@wuchong can't extend `FlinkTestBase` directly as it has different generic Fluss cluster configurations, but `FlinkAuthorizationITCase` needs a specific security-enabled configuration; rather, I will directly extend `AbstractTestBase` to cater to Flink Minicluster needs

fluss-flink/fluss-flink-common/src/test/java/com/alibaba/fluss/flink/source/metrics/FlinkSourceReaderMetricsTest.java
```diff
@@ -32,7 +33,7 @@
 import static org.assertj.core.api.Assertions.assertThat;
 
 /** Unit test for {@link com.alibaba.fluss.flink.source.metrics.FlinkSourceReaderMetrics}. */
-class FlinkSourceReaderMetricsTest {
+class FlinkSourceReaderMetricsTest extends FlinkTestBase {
```
wuchong: `FlinkSourceReaderMetricsTest` is not an integration test (`ITCase`) that doesn't rely on a fluss cluster and flink cluster, therefore, it doesn't need to extend `FlinkTestBase`. Setting up clusters is very heavy which should be avoided as much as possible to reduce the test time.

wuchong (approved):
@MehulBatra thanks, this looks good to me except `FlinkSourceReaderMetricsTest` doesn't need to extend `FlinkTestBase`.

fluss-flink/fluss-flink-common/src/test/java/com/alibaba/fluss/flink/source/metrics/FlinkSourceReaderMetricsTest.java
MehulBatra: My bad, done mistakenly despite knowing, I will be more cautious.

MehulBatra:
> @MehulBatra thanks, this looks good to me except `FlinkSourceReaderMetricsTest` doesn't need to extend `FlinkTestBase`.

Thank you @wuchong for the review!

Issue #810, opened 2025-04-30
Avoid create tables with system column names


wuchong:
### Search before asking

- [x] I searched in the [issues](https://github.com/alibaba/fluss/issues) and found nothing similar.

### Motivation

https://github.com/alibaba/fluss/pull/640 introduced some system columns (`__offset`, `__timestamp`, `__bucket`) for lake enabled tables. We should throw exceptions during the creation of the table (even if lake is not enabled for them). This allows users to enable lake for the tables in the future without throwing exceptions.

### Willingness to contribute

- [ ] I'm willing to submit a PR!

zhaomin1423:
Could you assign it to me?

wuchong:
@MehulBatra the proposal sounds great to me! 

@zhaomin1423 are you interested in implementing this?

zhaomin1423:
> [@MehulBatra](https://github.com/MehulBatra) the proposal sounds great to me!
> 
> [@zhaomin1423](https://github.com/zhaomin1423) are you interested in implementing this?

Yes, thanks @MehulBatra

PR #878, opened 2025-05-14
Avoid create tables with system column names

zhaomin1423:
### Purpose

Linked issue: close [#810 ](https://github.com/alibaba/fluss/issues/810#issue-3030804465)

### Brief change log

Verify that the system columns are used.


fluss-server/src/main/java/com/alibaba/fluss/server/utils/TableDescriptorValidation.java
```diff
@@ -72,6 +78,23 @@ public static void validateTableDescriptor(TableDescriptor tableDescriptor, int
...
+                Arrays.asList(OFFSET_COLUMN_NAME, TIMESTAMP_COLUMN_NAME, BUCKET_COLUMN_NAME);
+        List<String> unsupportedColumns =
+                fieldNames.stream().filter(systemColumns::contains).collect(Collectors.toList());
+        if (!unsupportedColumns.isEmpty()) {
+            throw new InvalidTableException(
+                    String.format(
+                            "Unsupported columns: %s. %s are system columns, please don't use them.",
+                            String.join(", ", unsupportedColumns),
+                            String.join(", ", systemColumns)));
+        }
```
wuchong:
```suggestion
            throw new InvalidTableException(
                    String.format(
                            "%s cannot be used as column names, "
                                    + "because they are reserved system columns in Fluss. "
                                    + "Please use other names for these columns. "
                                    + "The reserved system columns are: %s",
                            String.join(", ", unsupportedColumns),
                            String.join(", ", systemColumns)));
```

I improved a bit for the exception message.

fluss-server/src/test/java/com/alibaba/fluss/server/utils/TableDescriptorValidationTest.java
```diff
@@ -0,0 +1,42 @@
...
+import com.alibaba.fluss.metadata.Schema;
+import com.alibaba.fluss.types.DataTypes;
+
+import org.junit.jupiter.api.Test;
+
+import static org.assertj.core.api.Assertions.assertThatNoException;
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
+
+/** Test for {@link TableDescriptorValidation} . */
+public class TableDescriptorValidationTest {
```
wuchong: Could you add an integration test `testCreateTableWithInvalidColumns` in `FlussAdminITCase`? If the IT case is added, I think we don't need this unit test anymore.

wuchong (commented):
Thanks for the contribution @zhaomin1423 , the changes looks good in general. I left some minor comments.

fluss-server/src/test/java/com/alibaba/fluss/server/utils/TableDescriptorValidationTest.java
zhaomin1423: Thanks, updated

fluss-server/src/main/java/com/alibaba/fluss/server/utils/TableDescriptorValidation.java
```diff
@@ -72,6 +77,25 @@ public static void validateTableDescriptor(TableDescriptor tableDescriptor, int
         checkMergeEngine(tableConf, hasPrimaryKey, schema);
         checkTieredLog(tableConf);
         checkPartition(tableConf, tableDescriptor.getPartitionKeys(), schema);
+        checkSystemColumns(schema);
+    }
+
+    private static void checkSystemColumns(RowType schema) {
+        List<String> fieldNames = schema.getFieldNames();
+        List<String> systemColumns =
```
MehulBatra: Can we create a set at the class level here? The list is going to grow, and it would help us to have a faster lookup via set and save us from recreating the same list on every function call.
@wuchong What do you think?

```
private static final Set<String> SYSTEM_COLUMNS = Collections.unmodifiableSet(
    new HashSet<>(Arrays.asList(OFFSET_COLUMN_NAME, TIMESTAMP_COLUMN_NAME, BUCKET_COLUMN_NAME)));
```

MehulBatra:
Thank you for your Contribution! left one comment remaining, looks good.

fluss-server/src/main/java/com/alibaba/fluss/server/utils/TableDescriptorValidation.java
wuchong: +1

wuchong: @zhaomin1423 could you update PR for this?

zhaomin1423: Updated, thanks @MehulBatra

wuchong (approved):

PR #893, opened 2025-05-15
[server] Support dynamic create partition when write data

### LINKED ISSUE #583: Support dynamic partitioning strategy to create partition while writing data if partition not exists

caozhen1937:
### Purpose

Linked issue: close #583

### DIFF
--- a/fluss-client/src/main/java/com/alibaba/fluss/client/table/writer/AbstractTableWriter.java
+++ b/fluss-client/src/main/java/com/alibaba/fluss/client/table/writer/AbstractTableWriter.java
@@ -16,20 +16,14 @@
 
 package com.alibaba.fluss.client.table.writer;
 
-import com.alibaba.fluss.client.admin.Admin;
-import com.alibaba.fluss.client.metadata.MetadataUpdater;
 import com.alibaba.fluss.client.table.getter.PartitionGetter;
 import com.alibaba.fluss.client.write.WriteRecord;
 import com.alibaba.fluss.client.write.WriterClient;
 import com.alibaba.fluss.config.ConfigOptions;
-import com.alibaba.fluss.exception.FlussRuntimeException;
-import com.alibaba.fluss.exception.PartitionNotExistException;
 import com.alibaba.fluss.metadata.PhysicalTablePath;
-import com.alibaba.fluss.metadata.ResolvedPartitionSpec;
 import com.alibaba.fluss.metadata.TableInfo;
 import com.alibaba.fluss.metadata.TablePath;
 import com.alibaba.fluss.row.InternalRow;
-import com.alibaba.fluss.utils.ExceptionUtils;
 
 import javax.annotation.Nullable;
 
@@ -43,13 +37,9 @@ public abstract class AbstractTableWriter implements TableWriter {
     protected final WriterClient writerClient;
     protected final int fieldCount;
     private final @Nullable PartitionGetter partitionFieldGetter;
-    private final MetadataUpdater metadataUpdater;
 
     protected AbstractTableWriter(
-            TablePath tablePath,
-            TableInfo tableInfo,
-            MetadataUpdater metadataUpdater,
-            WriterClient writerClient) {
+            TablePath tablePath, TableInfo tableInfo, WriterClient writerClient) {
         this.tablePath = tablePath;
         this.writerClient = writerClient;
         this.fieldCount = tableInfo.getRowType().getFieldCount();
@@ -57,7 +47,6 @@ public abstract class AbstractTableWriter implements TableWriter {
                 tableInfo.isPartitioned()
                         ? new PartitionGetter(tableInfo.getRowType(), tableInfo.getPartitionKeys())
                         : null;
-        this.metadataUpdater = metadataUpdater;
     }
 
     /**
@@ -85,43 +74,14 @@ public abstract class AbstractTableWriter implements TableWriter {
         return future;
     }
 
-    protected PhysicalTablePath getPhysicalPath(
-            InternalRow row, Admin admin, boolean isDynamicCreatePartition) {
+    protected PhysicalTablePath getPhysicalPath(InternalRow row) {
         // not partitioned table, return the original physical path
         if (partitionFieldGetter == null) {
             return PhysicalTablePath.of(tablePath);
         } else {
             // partitioned table, extract partition from the row
             String partition = partitionFieldGetter.getPartition(row);
-            PhysicalTablePath partitionPath = PhysicalTablePath.of(tablePath, partition);
-            // may update partition info
-            boolean isExists;
-            try {
-                isExists = metadataUpdater.checkAndUpdatePartitionMetadata(partitionPath);
-            } catch (Exception e) {
-                Throwable t = ExceptionUtils.stripExecutionException(e);
-                if (t.getCause() instanceof PartitionNotExistException) {
-                    if (!isDynamicCreatePartition) {
-                        throw new FlussRuntimeException(e);
-                    }
-                    isExists = false;
-                } else {
-                    throw new FlussRuntimeException(e);
-                }
-            }
-            if (!isExists && isDynamicCreatePartition) {
-                ResolvedPartitionSpec resolvedPartitionSpec =
-                        partitionFieldGetter.getResolvedPartitionSpec(row);
-                try {
-                    admin.createPartition(tablePath, resolvedPartitionSpec.toPartitionSpec(), true)
-                            .get();
-                    // after create partition, update partition metadata
-                    metadataUpdater.checkAndUpdatePartitionMetadata(partitionPath);
-                } catch (Exception e) {
-                    throw new FlussRuntimeException(e);
-                }
-            }
-            return partitionPath;
+            return PhysicalTablePath.of(tablePath, partition);
         }
     }
 
--- a/fluss-client/src/main/java/com/alibaba/fluss/client/table/writer/TableUpsert.java
+++ b/fluss-client/src/main/java/com/alibaba/fluss/client/table/writer/TableUpsert.java
@@ -16,8 +16,6 @@
 
 package com.alibaba.fluss.client.table.writer;
 
-import com.alibaba.fluss.client.admin.Admin;
-import com.alibaba.fluss.client.metadata.MetadataUpdater;
 import com.alibaba.fluss.client.write.WriterClient;
 import com.alibaba.fluss.metadata.TableInfo;
 import com.alibaba.fluss.metadata.TablePath;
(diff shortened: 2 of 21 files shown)


caozhen1937:
Hi @wuchong @swuferhong , PTAL. And there have some tests failed, I will track them.

swuferhong (commented):
Hi, @caozhen1937. Thank you for your contribution. I think we should not perform the `createPartition` operation within the `Metadata RPC` — doing so would blur the definition and purpose of the `Metadata RPC`, it only used for requesting the latest metadata. Moreover, the `Metadata RPC` is a common interface available on both the `TabletServer` and the `CoordinatorServer`, but `createPartition` is an operation that should only be submitted through the `CoordinatorServer`.

Therefore, I believe the correct flow should be: first request the metadata, and only if the partition does not exist, then invoke admin.createPartition.

caozhen1937:
> Hi, @caozhen1937. Thank you for your contribution. I think we should not perform the `createPartition` operation within the `Metadata RPC` — doing so would blur the definition and purpose of the `Metadata RPC`, it only used for requesting the latest metadata. Moreover, the `Metadata RPC` is a common interface available on both the `TabletServer` and the `CoordinatorServer`, but `createPartition` is an operation that should only be submitted through the `CoordinatorServer`.
> 
> Therefore, I believe the correct flow should be: first request the metadata, and only if the partition does not exist, then invoke admin.createPartition.

Ok, let me think and retry.

fluss-client/src/main/java/com/alibaba/fluss/client/table/writer/AbstractTableWriter.java
```diff
@@ -80,16 +82,22 @@ protected CompletableFuture<Void> send(WriteRecord record) {
         return future;
     }
 
-    protected PhysicalTablePath getPhysicalPath(InternalRow row) {
+    protected PhysicalTablePath getPhysicalPath(
+            InternalRow row, Admin admin, boolean isDynamicCreatePartition) {
```
swuferhong: isDynamicCreatePartition need set to be a configurable option.

fluss-client/src/main/java/com/alibaba/fluss/client/table/writer/TableUpsert.java
```diff
@@ -35,26 +36,31 @@ public class TableUpsert implements Upsert {
     private final MetadataUpdater metadataUpdater;
 
     private final @Nullable int[] targetColumns;
+    private final Admin admin;
 
+    // TODO 传入admin
```
swuferhong: remove it.

swuferhong (commented):
Please add more ITCase

fluss-client/src/main/java/com/alibaba/fluss/client/table/writer/AbstractTableWriter.java
```diff
@@ -80,16 +82,22 @@ protected CompletableFuture<Void> send(WriteRecord record) {
...
         // not partitioned table, return the original physical path
         if (partitionFieldGetter == null) {
             return PhysicalTablePath.of(tablePath);
         } else {
             // partitioned table, extract partition from the row
             String partition = partitionFieldGetter.getPartition(row);
+            ResolvedPartitionSpec resolvedPartitionSpec =
+                    partitionFieldGetter.getResolvedPartitionSpec(row);
```
swuferhong: Move this into `if (!isExists && isDynamicCreatePartition) {}` to avoid call every time even if partition exists

swuferhong:
Hi, @caozhen1937, I have validated your PR on a large-scale test dataset and found that the current implementation have significant performance bottlenecks. Additionally, because one partition is not created, it blocks the writes for all other partitions. Therefore, I modified a version based on your PR. If you don’t mind, I will push a new commit based on your PR to help complete this issue togethe?

caozhen1937:
> Hi, @caozhen1937, I have validated your PR on a large-scale test dataset and found that the current implementation have significant performance bottlenecks. Additionally, because one partition is not created, it blocks the writes for all other partitions. Therefore, I modified a version based on your PR. If you don’t mind, I will push a new commit based on your PR to help complete this issue togethe?

@swuferhong Thank you，you can push your change.

fluss-client/src/main/java/com/alibaba/fluss/client/write/DynamicPartitionCreator.java
```diff
@@ -0,0 +1,131 @@
...
+                    }
+                } else {
+                    throw new FlussRuntimeException(e);
+                }
+            }
+        }
+
+        synchronized (inflightPartitionsLock) {
+            if (isExists) {
+                if (dynamicPartitionEnabled) {
```
caozhen1937, line 94: I am a bit confused:

1.  Should `if (dynamicPartitionEnabled)` be placed outside `synchronized (inflightPartitionsLock)`?  Because I think create partition when `dynamicPartitionEnabled` is true.
2.  I understand `inflightPartitionsToCreate` is aim to save partitions that is creating. There have a case that a partition only write one data. In this code, `inflightPartitionsToCreate` will exist in some data cannot be deleted.  Can we remove in `createPartition(physicalTablePath)` when catch nothing expection?

fluss-client/src/main/java/com/alibaba/fluss/client/write/DynamicPartitionCreator.java
```diff
@@ -0,0 +1,131 @@
...
+        TablePath tablePath = physicalTablePath.getTablePath();
+        checkArgument(partitionName != null, "Partition name shouldn't be null.");
+        TableInfo tableInfo = metadataUpdater.getTableInfoOrElseThrow(tablePath);
+        List<String> partitionKeys = tableInfo.getPartitionKeys();
+        ResolvedPartitionSpec resolvedPartitionSpec =
+                ResolvedPartitionSpec.fromPartitionName(partitionKeys, partitionName);
+
+        admin.createPartition(tablePath, resolvedPartitionSpec.toPartitionSpec(), true)
+                .whenComplete(
+                        (partitionId, throwable) -> {
```
caozhen1937, line 117: Should we catch `InvalidPartitionException` when  run `validatePartitionSpec` method in `createPartition` and throw exception?

fluss-client/src/main/java/com/alibaba/fluss/client/write/WriteBatch.java
```diff
@@ -44,21 +44,26 @@ public abstract class WriteBatch {
...
     protected int recordCount;
     private long drainedMs;
+    private TableBucket tableBucket;
 
     public WriteBatch(
-            TableBucket tableBucket, PhysicalTablePath physicalTablePath, long createdMs) {
+            TableBucket tableBucket,
+            PhysicalTablePath physicalTablePath,
+            long createdMs,
+            boolean isPartitionedTable) {
```
wuchong, line 62: We don't need to introduce this flag, because it can be drived from `PhysicalTablePath` (check whether this is a not-null partition name).

fluss-common/src/main/java/com/alibaba/fluss/config/ConfigOptions.java
```diff
@@ -1105,6 +1105,14 @@ public class ConfigOptions {
                             "The format of the kv records in kv store. The default value is `compacted`. "
                                     + "The supported formats are `compacted` and `indexed`.");
 
+    public static final ConfigOption<Boolean> TABLE_DYNAMIC_PARTITION_ENABLED =
+            key("table.dynamic-partition.enabled")
```
wuchong, line 1109: Currently, all the `table.*` options are storage options which will affect the storage behavior. However, this option only changes the behavior of client. It would be better to put it in `client.writer.*`.

fluss-client/src/main/java/com/alibaba/fluss/client/table/getter/PartitionGetter.java
```diff
@@ -64,4 +64,16 @@ public String getPartition(InternalRow row) {
                 new ResolvedPartitionSpec(partitionKeys, partitionValues);
         return resolvedPartitionSpec.getPartitionName();
     }
+
+    public ResolvedPartitionSpec getResolvedPartitionSpec(InternalRow row) {
```
wuchong, line 68: never used, can be removed.

fluss-client/src/test/java/com/alibaba/fluss/client/table/PartitionedTableITCase.java
```diff
@@ -130,8 +138,96 @@ void testPartitionedLogTable() throws Exception {
...
+        Table table = conn.getTable(DATA1_TABLE_PATH_PK);
+        AppendWriter appendWriter = table.newAppend().createWriter();
+        int partitionSize = 5;
+
+        // first try to add records add wait partition created.
+        Map<String, List<InternalRow>> expectPartitionNameAndAppendRows = new HashMap<>();
+        for (int i = 0; i < partitionSize; i++) {
+            String partitionName = String.valueOf(i);
+            InternalRow row = row(i, "a" + i, partitionName);
+            appendWriter.append(row).get();
```
wuchong, line 169: we can remove the `get()` to reduce the test time? Because you have called the `flush()` after the loop.

fluss-client/src/test/java/com/alibaba/fluss/client/table/PartitionedTableITCase.java
```diff
@@ -130,8 +138,96 @@ void testPartitionedLogTable() throws Exception {
...
+            expectPartitionNameAndAppendRows
+                    .computeIfAbsent(partitionName, k -> new ArrayList<>())
+                    .add(row);
+        }
+        appendWriter.flush();
+
+        retry(
```
wuchong, line 176: can use `waitValue` to avoid another `admin.listPartitionInfos` call after the `retry`.

fluss-client/src/main/java/com/alibaba/fluss/client/write/WriteBatch.java
```diff
@@ -115,11 +120,20 @@ public boolean hasBatchSequence() {
         return batchSequence() != LogRecordBatch.NO_BATCH_SEQUENCE;
     }
 
+    public @Nullable Long getPartitionId() {
```
wuchong, line 123: Can we remove the `partitionId` and `tableBucket` in the `WriteBatch`?

fluss-client/src/main/java/com/alibaba/fluss/client/write/DynamicPartitionCreator.java
swuferhong: here we also want to catch `TooManyPartitionsException` and `TooManyBucketsException`. So I think this error can be handle together in one place.

swuferhong:
Hi, @caozhen1937. Currently, there is no test report available. An official performance report for testing dynamic partition + multi partition performs will be released with version 0.7.0.  As there are many other changes to WriteClient, I've created a new PR based on your original pull request: https://github.com/alibaba/fluss/pull/1002, PTAL. cc @wuchong  comments addressed in pr: https://github.com/alibaba/fluss/pull/1002

swuferhong:
I will close this pr soon, and the new pr link: https://github.com/alibaba/fluss/pull/1002

caozhen1937:
@swuferhong Get, Thank you!

PR #894, opened 2025-05-15
[hotfix] typo and auto-partition config description

caozhen1937:
### Purpose

hotfix

### Brief change log
The newest auto-partition config `table.auto-partition.key`.
1. description has typo
2. introduce in options.md


fluss-common/src/main/java/com/alibaba/fluss/config/ConfigOptions.java
```diff
@@ -1109,7 +1109,7 @@ public class ConfigOptions {
                     .stringType()
                     .noDefaultValue()
                     .withDescription(
-                            "The singe time partition key to execute auto partition on when the table has multiple partition keys."
+                            "The single time partition key which is used as the partition key for auto create partitions."
```
wuchong: This is still confusing... "The .. partition key is used as the partition key ..."

Maybe you can rearrange the description with the following content. 

"This configuration specifies the time-based partition key to be used for auto-partitioning when a table is partitioned using multiple partition keys.

Auto-partitioning relies on a time-based partition key to automatically manage partitions—such as creating new partitions and dropping outdated ones—by comparing the partition’s time value with the current system time. When a table uses multiple partition keys (e.g., a composite partitioning strategy), this property defines which of those keys should be used as the primary time dimension for auto-partitioning decisions."

wuchong (approved):

PR #908, opened 2025-05-19
627 introduce fluss sink ds

polyzos:


polyzos:
We will support target columns via column names with [this](https://github.com/alibaba/fluss/issues/925).

polyzos:
@wuchong Please help review to see if there is anything missing here

fluss-flink/fluss-flink-common/src/main/java/com/alibaba/fluss/flink/sink/FlussSinkBuilder.java
```diff
@@ -0,0 +1,235 @@
...
+        checkNotNull(tableRowType, "Table row type is required but not provided.");
+
+        checkNotNull(database, "Database is required but not provided.");
+        checkArgument(!database.isEmpty(), "Database cannot be empty.");
+
+        checkNotNull(tableName, "Table name is required but not provided.");
+        checkArgument(!tableName.isEmpty(), "Table name cannot be empty.");
+
+        checkArgument(
+                isUpsert && partialUpdateColumns == null,
```
MehulBatra: ` isUpsert || partialUpdateColumns == null,
` 
I believe the logic should state that partial updates aren't supported in append mode, but are allowed in upsert mode. Please correct me if I've missed anything.

fluss-flink/fluss-flink-common/src/main/java/com/alibaba/fluss/flink/sink/FlussSinkBuilder.java
```diff
@@ -0,0 +1,235 @@
...
+
+    /** Set target column indexes. */
+    public FlussSinkBuilder<InputT> setDataLakeFormat(DataLakeFormat lakeFormat) {
+        this.lakeFormat = lakeFormat;
+        return this;
+    }
+
+    /** Set shuffle by bucket id. */
+    public FlussSinkBuilder<InputT> setShuffleByBucketId(boolean shuffleByBucketId) {
+        if (!shuffleByBucketId) {
```
MehulBatra:
```
    this.shuffleByBucketId = shuffleByBucketId;
    return this;
```
The method only changes the value when the parameter is false, ignoring true values. This creates unpredictable behavior where once you set it to false, you can never set it back to true.

fluss-flink/fluss-flink-common/src/main/java/com/alibaba/fluss/flink/sink/FlussSinkBuilder.java
polyzos: @MehulBatra that's correct

fluss-flink/fluss-flink-common/src/main/java/com/alibaba/fluss/flink/sink/FlussSinkBuilder.java
polyzos: You are probably right, this check should be redundant

fluss-flink/fluss-flink-common/src/main/java/com/alibaba/fluss/flink/sink/FlussSinkBuilder.java
```diff
@@ -0,0 +1,235 @@
...
+ *
+ * <p>The builder supports automatic schema inference from POJO classes using reflection and
+ * provides options for customizing data conversion logic through custom converters.
+ *
+ * <p>Example usage:
+ *
+ * <pre>{@code
+ * FlinkSink<Order> sink = new FlussSinkBuilder<Order>()
+ *          .setBootstrapServers(bootstrapServers)
+ *          .setTable(tableName)
```
wuchong, line 55: add `.setDatabase` in the example which is required.

fluss-flink/fluss-flink-common/src/main/java/com/alibaba/fluss/flink/sink/FlussSinkBuilder.java
```diff
@@ -0,0 +1,235 @@
...
+    }
+
+    /** Set target column indexes. */
+    public FlussSinkBuilder<InputT> setPartialUpdateColumns(int[] partialUpdateColumns) {
+        this.partialUpdateColumns = partialUpdateColumns;
+        return this;
+    }
+
+    /** Set target column indexes. */
+    public FlussSinkBuilder<InputT> setDataLakeFormat(DataLakeFormat lakeFormat) {
```
wuchong: This can be obtained from table config, so users don't need to set it on sink manually.

fluss-flink/fluss-flink-common/src/main/java/com/alibaba/fluss/flink/sink/FlussSinkBuilder.java
wuchong: +1

fluss-flink/fluss-flink-common/src/main/java/com/alibaba/fluss/flink/sink/FlussSinkBuilder.java
```diff
@@ -0,0 +1,235 @@
...
+    public FlussSinkBuilder<InputT> useUpsert() {
+        this.isUpsert = true;
+        return this;
+    }
+
+    /** Configure this sink to use append-only semantics. */
+    public FlussSinkBuilder<InputT> useAppend() {
+        this.isUpsert = false;
+        return this;
+    }
```
wuchong: This is determined by the table itself, not by users?

fluss-flink/fluss-flink-common/src/main/java/com/alibaba/fluss/flink/sink/FlussSinkBuilder.java
```diff
@@ -0,0 +1,235 @@
...
+
+    /** Set a FlussSerializationSchema. */
+    public FlussSinkBuilder<InputT> setSerializationSchema(
+            FlussSerializationSchema<InputT> serializationSchema) {
+        this.serializationSchema = serializationSchema;
+        return this;
+    }
+
+    /** Build the FlussSink. */
+    public FlinkSink<InputT> build() {
```
wuchong: return the `FlussSink<InputT>`

fluss-flink/fluss-flink-common/src/main/java/com/alibaba/fluss/flink/sink/FlussSinkBuilder.java
```diff
@@ -0,0 +1,235 @@
...
+        validateConfiguration();
+
+        Configuration flussConfig = Configuration.fromMap(configOptions);
+
+        FlinkSink.SinkWriterBuilder<? extends FlinkSinkWriter<InputT>, InputT> writerBuilder;
+
+        TablePath tablePath = new TablePath(database, tableName);
+        flussConfig.setString("bootstrap.servers", bootstrapServers);
```
wuchong: nit: Use `ConfigOptions.BOOTSTRAP_SERVERS.key()` instead of hard code the option key.

fluss-flink/fluss-flink-common/src/test/java/com/alibaba/fluss/flink/sink/FlussSinkBuilderTest.java
```diff
@@ -0,0 +1,296 @@
...
+import java.util.Map;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
+
+/** Tests for {@link FlussSinkBuilder} configuration and argument handling. */
+class FlussSinkBuilderTest {
+    private String bootstrapServers = "localhost:9123";
+    private String databaseName = "testDb";
+    private String tableName = "testTable";
```
wuchong: nit: can be final

fluss-flink/fluss-flink-common/src/test/java/com/alibaba/fluss/flink/sink/FlussSinkITCase.java
```diff
@@ -0,0 +1,352 @@
...
+
+        RowData row9 = converter.toFlinkRowData(row(900L, 24L, 603, "addr4"));
+        row9.setRowKind(RowKind.UPDATE_BEFORE);
+
+        inputRows.add(row8);
+        inputRows.add(row9);
+
+        // Assert result size and elements match
+        Assertions.assertEquals(rows.size(), inputRows.size());
+        Assertions.assertTrue(rows.containsAll(inputRows));
```
wuchong: use assertj `assertThat` instead.

fluss-flink/fluss-flink-common/src/test/java/com/alibaba/fluss/flink/sink/FlussSinkITCase.java
```diff
@@ -0,0 +1,352 @@
...
+                    rows.add(order);
+                }
+            }
+        }
+
+        orders.add(new TestOrder(800, 23, 602, "addr3", RowKind.UPDATE_BEFORE));
+        orders.add(new TestOrder(900, 24, 603, "addr4", RowKind.UPDATE_BEFORE));
+
+        Assertions.assertEquals(rows.size(), orders.size());
+        Assertions.assertTrue(rows.containsAll(orders));
```
wuchong: ditto

fluss-flink/fluss-flink-common/src/test/java/com/alibaba/fluss/flink/sink/FlussSinkITCase.java
```diff
@@ -0,0 +1,352 @@
...
+        TablePath pkTablePath = TablePath.of(DEFAULT_DB, pkTableName);
+
+        createTable(pkTablePath, pkTableDescriptor);
+
+        this.ordersPKTablePath = new TablePath(DEFAULT_DB, pkTableName);
+
+        this.bootstrapServers = conn.getConfiguration().get(ConfigOptions.BOOTSTRAP_SERVERS).get(0);
+
+        env = StreamExecutionEnvironment.getExecutionEnvironment();
+        env.setParallelism(2);
```
wuchong: Could you rebase the branch to the latest `main`, and take `com.alibaba.fluss.flink.sink.FlinkTableSinkITCase` as an example about how to initialize `env`? We changed the way of env initialization in the recent days to make the tests more stable.

fluss-flink/fluss-flink-common/src/test/java/com/alibaba/fluss/flink/sink/FlussSinkITCase.java
```diff
@@ -0,0 +1,352 @@
...
+            row.setField(0, value.orderId);
+            row.setField(1, value.itemId);
+            row.setField(2, value.amount);
+            row.setField(3, BinaryString.fromString(value.address));
+
+            RowKind rowKind = value.rowKind;
+            switch (rowKind) {
+                case INSERT:
+                case UPDATE_AFTER:
+                case UPDATE_BEFORE:
```
wuchong: We should apply DELETE operation for `UPDATE_BEFORE` event.

fluss-flink/fluss-flink-common/src/test/java/com/alibaba/fluss/flink/sink/FlussSinkITCase.java
```diff
@@ -0,0 +1,352 @@
...
+        Assertions.assertEquals(rows.size(), orders.size());
+        Assertions.assertTrue(rows.containsAll(orders));
+    }
+
+    private static class TestOrder implements Serializable {
+        private long orderId;
+        private long itemId;
+        private int amount;
+        private String address;
+        private RowKind rowKind;
```
wuchong: nit: can mark them as `final`.

fluss-flink/fluss-flink-common/src/test/java/com/alibaba/fluss/flink/sink/FlussSinkITCase.java
```diff
@@ -0,0 +1,352 @@
...
+
+    private static StreamExecutionEnvironment env;
+
+    private static TableDescriptor pkTableDescriptor;
+
+    private static String bootstrapServers;
+
+    private static String pkTableName = "orders_test_pk";
+
+    private static TablePath ordersPKTablePath;
```
wuchong: never used, can be removed.

wuchong (commented):
Thanks @polyzos , I left some comments.

fluss-flink/fluss-flink-common/src/main/java/com/alibaba/fluss/flink/sink/FlussSinkBuilder.java
```diff
@@ -0,0 +1,235 @@
...
+    }
+
+    /** Set the table name for the sink. */
+    public FlussSinkBuilder<InputT> setTable(String table) {
+        this.tableName = table;
+        return this;
+    }
+
+    /** Set the row type for the sink. */
+    public FlussSinkBuilder<InputT> setRowType(RowType rowType) {
```
wuchong: The tableRowType can also be retrieved from `TableInfo`.

polyzos: @wuchong  TableInfo return a Fluss rowType, however the writer constructors require A Flink rowType.. do we have any build-in converters for the row types as well?

polyzos:
@wuchong Thank you for your comments. I left one comment and addressed the rest. Let me know if there is anything I missed

fluss-flink/fluss-flink-common/src/main/java/com/alibaba/fluss/flink/sink/FlussSinkBuilder.java
```diff
@@ -0,0 +1,221 @@
...
+    }
+
+    /** Set whether to ignore delete operations. */
+    public FlussSinkBuilder<InputT> setIgnoreDelete(boolean ignoreDelete) {
+        this.ignoreDelete = ignoreDelete;
+        return this;
+    }
+
+    /** Set target column indexes. */
+    public FlussSinkBuilder<InputT> setPartialUpdateColumns(int[] partialUpdateColumns) {
```
wuchong: I removed this method because the builder doesn't currently support partial updates. We can reintroduce it once partial update support to the DataStream API in the future. Since it's a public API, making changes after release will be difficult, so it's better to wait until the feature is ready. Additionally, I think using field names instead of field indexes would be more user-friendly and less error-prone.

wuchong (approved):
I added a commit to improve it a bit. Will merge it once CI is passed.

fluss-flink/fluss-flink-common/src/main/java/com/alibaba/fluss/flink/sink/FlussSinkBuilder.java
polyzos: @wuchong indeed the goal was to make it work with field names, but we had a separate ticket to track this, [here](https://github.com/alibaba/fluss/issues/925)

wuchong: @polyzos cool!

PR #924, opened 2025-05-19
[common] FlussScheduler should use scheduleWithFixedDelay instead of scheduleAtFixedRate

### LINKED ISSUE #923: FlussScheduler should use scheduleWithFixedDelay instead of scheduleAtFixedRate
### Search before asking

### Fluss version

0.6.0 (latest release)

### Please describe the bug 🐞

`scheduleAtFixedRate` means

> Creates and executes a periodic action that becomes enabled first after the given initial delay, and subsequently with the given period; that is executions will commence after initialDelay then initialDelay+period, then initialDelay + 2 * period, and so on.

`scheduleWithFixedDelay` means

> Creates and executes a periodic action that becomes enabled first after the given initial delay, and subsequently with the given delay between the termination of one execution and the commencement of the next.

Therefore we would experience task stacking in `DelayedWorkQueue` due to the long execution time of a single task for `scheduleAtFixedRate` case. Too much tasks remaining in `DelayedWorkQueue` will cost lot memory and CPU.

### Solution

`FlussScheduler` should use `scheduleWithFixedDelay` instead of `scheduleAtFixedRate`.

SteNicholas:
### Purpose

Linked issue: close #923.

`FlussScheduler` should use `scheduleWithFixedDelay` instead of `scheduleAtFixedRate`.

### Brief change log

`scheduleAtFixedRate` means

> Creates and executes a periodic action that becomes enabled first after the given initial delay, and subsequently with the given period; that is executions will commence after initialDelay then initialDelay+period, then initialDelay + 2 * period, and so on.

`scheduleWithFixedDelay` means

> Creates and executes a periodic action that becomes enabled first after the given initial delay, and subsequently with the given delay between the termination of one execution and the commencement of the next.

Therefore we would experience task stacking in `DelayedWorkQueue` due to the long execution time of a single task for `scheduleAtFixedRate` case. Too much tasks remaining in `DelayedWorkQueue` will cost lot memory and CPU.

Backport https://github.com/apache/celeborn/pull/1970.

### Tests

No.

### API and Format

No.

### Documentation

No.

### DIFF
--- a/fluss-common/src/main/java/com/alibaba/fluss/utils/concurrent/FlussScheduler.java
+++ b/fluss-common/src/main/java/com/alibaba/fluss/utils/concurrent/FlussScheduler.java
@@ -121,7 +121,7 @@ public class FlussScheduler implements Scheduler {
                             }
                         };
                 if (periodMs > 0) {
-                    return executor.scheduleAtFixedRate(
+                    return executor.scheduleWithFixedDelay(
                             runnable, delayMs, periodMs, TimeUnit.MILLISECONDS);
                 } else {
                     return executor.schedule(runnable, delayMs, TimeUnit.MILLISECONDS);


SteNicholas:
Ping @wuchong.

wuchong:
Thanks, @SteNicholas, for suggesting this change. However, I still prefer using `scheduleAtFixedRate` in this case because it provides a stronger guarantee on the **execution interval**.

For example, the `ReplicaManager#startHighWatermarkCheckPointThread` schedules a `checkpointHighWatermarks` task every 5 seconds. If we use `scheduleWithFixedDelay`, and the task takes 1 second to complete, the next execution will be delayed — resulting in checkpoints every 6 seconds. In contrast, `scheduleAtFixedRate` ensures that the checkpoint happens **every 5 seconds**, aligning better with user expectations.

In addition, Kafka has been using `scheduleAtFixedRate` for `KafkaScheduler` for many years without issues. I think we should stick with the current approach unless we encounter concrete problems in practice. Let’s revisit this if needed in the future, rather than making aggressive changes now.

PR #930, opened 2025-05-20
[server] Metadata RPC get table/partition metadata from serverCache instead of Zookeeper

### LINKED ISSUE #901: Metadata RPC should get table metadata from serverCache instead of Zookeeper
### Search before asking

### Motivation

`Metadata RPC` should get table metadata from serverCache instead of Zookeeper.  The purpose of doing this is that, when the cluster has a large amount of tables and buckets, fetching data from ZooKeeper once can take around 100ms. For 512 buckets, this would result in a total delay of about 50 seconds, which is unacceptable for `Metadata RPC` calls. Therefore, we want to retrieve the data from a cache instead. The completion of this work depends on issues #757 and #900.

### Solution

_No response_

### Anything else?

_No response_

### Willingness to contribute

swuferhong:
### Purpose

Linked issue: close #901 

Metadata RPC should get table/partition metadata from serverCache instead of Zookeeper. The purpose of doing this is that, when the cluster has a large amount of tables and buckets, fetching data from ZooKeeper once can take around 100ms. For 512 buckets, this would result in a total delay of about 50 seconds, which is unacceptable for Metadata RPC calls. Therefore, we want to retrieve the data from a cache instead.


fluss-server/src/main/java/com/alibaba/fluss/server/coordinator/CoordinatorService.java
```diff
@@ -121,6 +135,9 @@ public final class CoordinatorService extends RpcServiceBase implements Coordina
     private final int defaultReplicationFactor;
     private final Supplier<EventManager> eventManagerSupplier;
     private final Supplier<Integer> coordinatorEpochSupplier;
+    private final ServerMetadataCache metadataCache;
+    private final ZooKeeperClient zkClient;
```
wuchong: Use the parent `zkClient.`

fluss-server/src/main/java/com/alibaba/fluss/server/coordinator/CoordinatorService.java
```diff
@@ -121,6 +135,9 @@ public final class CoordinatorService extends RpcServiceBase implements Coordina
...
+    private final MetadataManager metadataManager;
```
wuchong: Use the parent `metadataManager`.

fluss-server/src/main/java/com/alibaba/fluss/server/coordinator/CoordinatorService.java
```diff
@@ -545,4 +581,72 @@ private void validateHeartbeatRequest(
                             heartbeatReqForTable.getTableId()));
         }
     }
+
+    private TableMetadata getTableMetadata(CoordinatorContext ctx, TablePath tablePath) {
+        // always get table info from zk.
+        TableInfo tableInfo = metadataManager.getTable(tablePath);
+        long tableId = ctx.getTableIdByPath(tablePath);
+        List<BucketMetadata> bucketMetadataList;
+        if (tableId == TableInfo.UNKNOWN_TABLE_ID) {
```
wuchong, line 586: Add TODO, we should be able to not response the meta if table id not exists.

fluss-server/src/main/java/com/alibaba/fluss/server/RpcServiceBase.java
```diff
@@ -509,42 +455,103 @@ public CompletableFuture<ListAclsResponse> listAcls(ListAclsRequest request) {
...
+                tablePaths.add(tablePath);
+            }
+        }
+
+        List<PartitionMetadata> partitionMetadata = new ArrayList<>();
+        List<PhysicalTablePath> partitions = new ArrayList<>();
+        request.getPartitionsPathsList()
+                .forEach(
+                        pbPhysicalTablePath ->
+                                partitions.add(toPhysicalTablePath(pbPhysicalTablePath)));
```
wuchong: Use stream API

```
        List<PhysicalTablePath> partitions =
                request.getPartitionsPathsList().stream()
                        .map(ServerRpcMessageUtils::toPhysicalTablePath)
```

fluss-server/src/test/java/com/alibaba/fluss/server/testutils/FlussClusterExtension.java
```diff
@@ -176,6 +176,11 @@ public void afterEach(ExtensionContext extensionContext) throws Exception {
                 metadataManager.dropTable(TablePath.of(defaultDb, table), true);
             }
         }
+
+        // remove all table metadata in tabletServer cache.
+        for (TabletServer tabletServer : tabletServers.values()) {
+            tabletServer.getMetadataCache().clearTableMetadata();
```
wuchong, line 183: Use `waitUtil` to wait all the table metadata cleared, instead of manually clear them.

wuchong:
track the failed case `TableChangeWatcherTest.testTableChanges` in https://github.com/alibaba/fluss/issues/1018
rerun the test...

PR #932, opened 2025-05-20
[client] Sender trigger to abort all writeBatches if InitWriter failed with AuthorizationException

### LINKED ISSUE #931: Sender need to abort all writeBatches if InitWriter failed with AuthorizationException
### Search before asking

### Description

Currently, if `Sender` try to send InitWriter request and failed with AuthorizationException, it will not raise the exception to record callbacks, which will cause the flink client keep retrying. In this case, we need to trigger abortBatch() to let these writeBatches in accumulator to abort and raise exception to record callbacks. See comments in #850 

### Willingness to contribute

swuferhong:
### Purpose

Linked issue: close #931  and #756 

Currently, if Sender try to send InitWriter request and failed with AuthorizationException, it will not raise the exception to record callbacks, which will cause the flink client keep retrying. In this case, we need to trigger abortBatch() to let these writeBatches in accumulator to abort and raise exception to record callbacks. See comments in https://github.com/alibaba/fluss/pull/850

### DIFF
--- a/fluss-client/src/main/java/com/alibaba/fluss/client/write/ArrowLogWriteBatch.java
+++ b/fluss-client/src/main/java/com/alibaba/fluss/client/write/ArrowLogWriteBatch.java
@@ -131,4 +131,9 @@ public class ArrowLogWriteBatch extends WriteBatch {
     public int batchSequence() {
         return recordsBuilder.batchSequence();
     }
+
+    @Override
+    public void abortRecordAppends() {
+        recordsBuilder.abort();
+    }
 }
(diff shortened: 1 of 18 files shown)


swuferhong:
@wuchong pr ready, pls table a look, thx.

gyang94:
Assume that in a connection the writer client sends records to two tables (one has write permission and another doesn't), is it possible that in one writeBatches that all the records are only from the "no write permission table"? 
If that happens, the intWriter will fail. It may blocks the writeBatches from the "has write permission table".

swuferhong:
> Assume that in a connection the writer client sends records to two tables (one has write permission and another doesn't), is it possible that in one writeBatches that all the records are only from the "no write permission table"? If that happens, the intWriter will fail. It may blocks the writeBatches from the "has write permission table".

Yes, the scenario you described can happen. However, the sender won't crash — instead, it will asynchronously return an error for the records trying to write to tables with "no write permission". For tables with proper write permissions ("write with permission table"), the write can still succeed, and the sender will be able to obtain the writerId and proceed with writing. After that, the writes to "no write permission tables" will still fail due to lack of write permissions instead of init writer failed.

wuchong (approved):
Thanks for the great work @swuferhong and @gyang94 ! LGTM.

PR #934, opened 2025-05-20
Fix Flink CI timeout

### LINKED ISSUE #933: Flink CI is still getting timeout
### Search before asking

### Fluss version

main (development)

### Please describe the bug 🐞

We have applied some patches to improve the timeout and stability of Flink CI in https://github.com/alibaba/fluss/issues/868. 

However, there are still 60min timeouts in the CI. For example: 
https://github.com/alibaba/fluss/actions/runs/15114700232/job/42482289578
https://github.com/alibaba/fluss/actions/runs/15109265578/job/42464824513

### Solution

_No response_

wuchong:
### Purpose

Linked issue: close #933

### Brief change log

1. The first commit fix a lot of `LogFetchCollector` log in maven output, because the `LogScanner#wakeup()` didn't interrupt `LogScanner#poll()`
2. The second commit follows the suggested way of [MiniClusterExtension](https://github.com/apache/flink/blob/release-1.20.1/flink-test-utils-parent/flink-test-utils/src/main/java/org/apache/flink/test/junit5/MiniClusterExtension.java#L59) to initialize `StreamExecutionEnvironment` in each test method instead of being shared in the test class.

### DIFF
--- a/fluss-flink/fluss-flink-common/src/test/java/com/alibaba/fluss/flink/sink/FlinkTableSinkITCase.java
+++ b/fluss-flink/fluss-flink-common/src/test/java/com/alibaba/fluss/flink/sink/FlinkTableSinkITCase.java
@@ -22,7 +22,6 @@ import com.alibaba.fluss.client.table.Table;
 import com.alibaba.fluss.client.table.scanner.ScanRecord;
 import com.alibaba.fluss.client.table.scanner.log.LogScanner;
 import com.alibaba.fluss.client.table.scanner.log.ScanRecords;
-import com.alibaba.fluss.config.ConfigOptions;
 import com.alibaba.fluss.config.Configuration;
 import com.alibaba.fluss.metadata.TableBucket;
 import com.alibaba.fluss.metadata.TablePath;
@@ -44,7 +43,6 @@ import org.apache.flink.types.Row;
 import org.apache.flink.types.RowKind;
 import org.apache.flink.util.CloseableIterator;
 import org.junit.jupiter.api.AfterEach;
-import org.junit.jupiter.api.BeforeAll;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.extension.RegisterExtension;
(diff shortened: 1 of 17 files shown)


fluss-flink/fluss-flink-common/src/test/java/com/alibaba/fluss/flink/sink/FlinkTableSinkITCase.java
```diff
@@ -120,11 +118,7 @@ static void beforeAll() {
         tBatchEnv.executeSql("use catalog " + CATALOG_NAME);
         tBatchEnv
                 .getConfig()
-                .set(ExecutionConfigOptions.TABLE_EXEC_RESOURCE_DEFAULT_PARALLELISM, 4);
-    }
-
-    @BeforeEach
-    void before() {
+                .set(ExecutionConfigOptions.TABLE_EXEC_RESOURCE_DEFAULT_PARALLELISM, 2);
```
MehulBatra, line 121: I am trying to understand. In other tests, it is set to 4, except here it is set to 2? Did we do it intentionally?

wuchong: The reason for reducing the default parallelism from `4` to `2` is that some tests in `FlinkTableSinkITCase`, such as `testFirstRowMergeEngine`, launch multiple Flink jobs concurrently. 

When each job is set to a parallelism of `4` (the default), it consumes all available slots in the Flink cluster (which has only 4 task slots). This leaves no resources for subsequent jobs, causing them to fail to start.

Lowering the parallelism to `2` allows multiple jobs to run concurrently without exhausting the cluster’s resources. I don’t think this change will impact test execution time, as these tests typically process only a small number of rows. The main goal of using parallelism greater than `1` is to verify correct behavior in distributed mode, rather than to optimize performance.

MehulBatra: sounds goods!

PR #937, opened 2025-05-21
[flink]Support partition pushdown(only equals) in Flink connector

Alibaba-HZY:
[flink]

Purpose
Support partition pushdown(only equals) in Flink connector
Linked issue: close https://github.com/alibaba/fluss/issues/196

Tests
com.alibaba.fluss.connector.flink.source.FlinkTableSourceITCase#testStreamingReadSinglePartitionPushDown
com.alibaba.fluss.connector.flink.source.FlinkTableSourceITCase#testStreamingReadMultiPartitionPushDown

API and Format
no

Documentation
no

### DIFF
--- a/fluss-flink/fluss-flink-common/src/main/java/com/alibaba/fluss/flink/source/FlinkSource.java
+++ b/fluss-flink/fluss-flink-common/src/main/java/com/alibaba/fluss/flink/source/FlinkSource.java
@@ -29,6 +29,7 @@ import com.alibaba.fluss.flink.source.split.SourceSplitBase;
 import com.alibaba.fluss.flink.source.split.SourceSplitSerializer;
 import com.alibaba.fluss.flink.source.state.FlussSourceEnumeratorStateSerializer;
 import com.alibaba.fluss.flink.source.state.SourceEnumeratorState;
+import com.alibaba.fluss.flink.utils.PushdownUtils;
 import com.alibaba.fluss.metadata.TablePath;
 import com.alibaba.fluss.types.RowType;
 
@@ -46,6 +47,8 @@ import org.apache.flink.core.io.SimpleVersionedSerializer;
 
 import javax.annotation.Nullable;
 
+import java.util.List;
+
 /** Flink source for Fluss. */
 public class FlinkSource<OUT>
         implements Source<OUT, SourceSplitBase, SourceEnumeratorState>, ResultTypeQueryable {
@@ -62,6 +65,8 @@ public class FlinkSource<OUT>
     private final boolean streaming;
     private final FlussDeserializationSchema<OUT> deserializationSchema;
 
+    @Nullable private final List<PushdownUtils.FieldEqual> partitionFilters;
+
     public FlinkSource(
             Configuration flussConf,
             TablePath tablePath,
@@ -73,6 +78,32 @@ public class FlinkSource<OUT>
             long scanPartitionDiscoveryIntervalMs,
             FlussDeserializationSchema<OUT> deserializationSchema,
             boolean streaming) {
+        this(
+                flussConf,
+                tablePath,
+                hasPrimaryKey,
+                isPartitioned,
+                sourceOutputType,
+                projectedFields,
+                offsetsInitializer,
+                scanPartitionDiscoveryIntervalMs,
+                deserializationSchema,
+                streaming,
+                null);
+    }
+
+    public FlinkSource(
+            Configuration flussConf,
+            TablePath tablePath,
+            boolean hasPrimaryKey,
+            boolean isPartitioned,
+            RowType sourceOutputType,
+            @Nullable int[] projectedFields,
+            OffsetsInitializer offsetsInitializer,
+            long scanPartitionDiscoveryIntervalMs,
+            FlussDeserializationSchema<OUT> deserializationSchema,
+            boolean streaming,
+            @Nullable List<PushdownUtils.FieldEqual> partitionFilters) {
         this.flussConf = flussConf;
         this.tablePath = tablePath;
         this.hasPrimaryKey = hasPrimaryKey;
@@ -83,6 +114,7 @@ public class FlinkSource<OUT>
         this.scanPartitionDiscoveryIntervalMs = scanPartitionDiscoveryIntervalMs;
         this.deserializationSchema = deserializationSchema;
         this.streaming = streaming;
+        this.partitionFilters = partitionFilters;
     }
 
     @Override
@@ -101,7 +133,8 @@ public class FlinkSource<OUT>
                 splitEnumeratorContext,
                 offsetsInitializer,
                 scanPartitionDiscoveryIntervalMs,
-                streaming);
+                streaming,
+                partitionFilters);
     }
 
     @Override
@@ -118,7 +151,8 @@ public class FlinkSource<OUT>
                 sourceEnumeratorState.getAssignedPartitions(),
                 offsetsInitializer,
                 scanPartitionDiscoveryIntervalMs,
-                streaming);
+                streaming,
+                partitionFilters);
     }
 
     @Override
--- a/fluss-flink/fluss-flink-common/src/main/java/com/alibaba/fluss/flink/source/FlinkSource.java
+++ b/fluss-flink/fluss-flink-common/src/main/java/com/alibaba/fluss/flink/source/FlinkSource.java
@@ -29,7 +29,7 @@ import com.alibaba.fluss.flink.source.split.SourceSplitBase;
 import com.alibaba.fluss.flink.source.split.SourceSplitSerializer;
 import com.alibaba.fluss.flink.source.state.FlussSourceEnumeratorStateSerializer;
 import com.alibaba.fluss.flink.source.state.SourceEnumeratorState;
-import com.alibaba.fluss.flink.utils.PushdownUtils;
+import com.alibaba.fluss.flink.utils.PushdownUtils.FieldEqual;
 import com.alibaba.fluss.metadata.TablePath;
 import com.alibaba.fluss.types.RowType;
 
@@ -65,32 +65,7 @@ public class FlinkSource<OUT>
     private final boolean streaming;
     private final FlussDeserializationSchema<OUT> deserializationSchema;
 
-    @Nullable private final List<PushdownUtils.FieldEqual> partitionFilters;
-
-    public FlinkSource(
-            Configuration flussConf,
-            TablePath tablePath,
-            boolean hasPrimaryKey,
-            boolean isPartitioned,
-            RowType sourceOutputType,
-            @Nullable int[] projectedFields,
-            OffsetsInitializer offsetsInitializer,
-            long scanPartitionDiscoveryIntervalMs,
-            FlussDeserializationSchema<OUT> deserializationSchema,
-            boolean streaming) {
-        this(
-                flussConf,
-                tablePath,
-                hasPrimaryKey,
-                isPartitioned,
-                sourceOutputType,
-                projectedFields,
-                offsetsInitializer,
-                scanPartitionDiscoveryIntervalMs,
-                deserializationSchema,
-                streaming,
-                null);
-    }
+    @Nullable private final List<FieldEqual> partitionFilters;
 
     public FlinkSource(
             Configuration flussConf,
@@ -103,7 +78,7 @@ public class FlinkSource<OUT>
             long scanPartitionDiscoveryIntervalMs,
             FlussDeserializationSchema<OUT> deserializationSchema,
             boolean streaming,
-            @Nullable List<PushdownUtils.FieldEqual> partitionFilters) {
+            List<FieldEqual> partitionFilters) {
         this.flussConf = flussConf;
         this.tablePath = tablePath;
         this.hasPrimaryKey = hasPrimaryKey;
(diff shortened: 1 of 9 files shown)


fluss-flink/fluss-flink-common/src/test/java/com/alibaba/fluss/flink/source/FlinkTableSourceITCase.java
```diff
@@ -928,6 +929,78 @@ void testLookupFullCacheThrowException() {
...
+                        + " (a int not null, b varchar, c string, primary key (a, c) NOT ENFORCED) partitioned by (c) "
+                        + "with ('table.auto-partition.enabled' = 'true', 'table.auto-partition.time-unit' = 'year')");
+        TablePath tablePath = TablePath.of(DEFAULT_DB, "partitioned_table");
+
+        // write data into partitions and wait snapshot is done
+        Map<Long, String> partitionNameById =
+                waitUntilPartitions(FLUSS_CLUSTER_EXTENSION.getZooKeeperClient(), tablePath);
+        List<String> expectedRowValues =
+                writeRowsToPartition(tablePath, partitionNameById.values()).stream()
+                        .filter(s -> s.contains("2025"))
```
swuferhong, line 943: Do not use specific partitions like "2025", "2026", etc. on an auto partitioned table, and avoid write static filter values "2025" directly into the SQL. This may lead to unexpected behavior when the code runs in 2027 or later. Instead of using a static partitioned table, and manually create a partition named "2025".

fluss-flink/fluss-flink-common/src/test/java/com/alibaba/fluss/flink/source/testutils/FlinkTestBase.java
```diff
@@ -265,6 +265,29 @@ protected List<String> writeRowsToPartition(TablePath tablePath, Collection<Stri
...
+            values[0] = keyValuePairs[0].split("=")[1];
+            values[1] = keyValuePairs[1].split("=")[1];
+
+            for (int i = 0; i < 10; i++) {
+                rows.add(row(i, "v1", values[0], values[1]));
+                expectedRowValues.add(String.format("+I[%d, v1, %s, %s]", i, values[0], values[1]));
+            }
+        }
+
+        // 写入记录
```
swuferhong: Chinese word.

swuferhong (commented):
Thanks for your contribution, I left some comments, for `FlinkTableSource`, we need to implement `SupportsPartitionPushDown` rather than treating partition push-down as a special type of filter push-down. Using the filter approach prevents us from leveraging the Flink Planner to determine which partitions can be pushed down.

fluss-flink/fluss-flink-common/src/test/java/com/alibaba/fluss/flink/source/FlinkTableSourceITCase.java
```diff
@@ -928,6 +929,75 @@ void testLookupFullCacheThrowException() {
                 .hasMessage("Full lookup caching is not supported yet.");
     }
 
+    // -------------------------------------------------------------------------------------
+    // Fluss partition push down tests
+    // -------------------------------------------------------------------------------------
```
swuferhong: no need to add this comment line?

fluss-flink/fluss-flink-common/src/test/java/com/alibaba/fluss/flink/source/testutils/FlinkTestBase.java
```diff
@@ -265,6 +265,28 @@ protected List<String> writeRowsToPartition(TablePath tablePath, Collection<Stri
         return expectedRowValues;
     }
 
+    protected List<String> writeRowsToTwoPartition(
+            TablePath tablePath, Collection<String> partitions) throws Exception {
```
swuferhong: In my opinion, this method is not a common util method because you used hardcoded constructs like 'String[] values = new String[2];', 'keyValuePairs[0]', and 'keyValuePairs[1]'. Others cannot easily reuse this method in their cases, so I suggest either refactoring the hardcoded logic in the method to make it more generic or moving it to the `FlinkTableSourceITCase` class for pushdown case only.

fluss-flink/fluss-flink-common/src/main/java/com/alibaba/fluss/flink/source/FlinkSource.java
```diff
@@ -73,6 +78,32 @@ public FlinkSource(
...
+            TablePath tablePath,
+            boolean hasPrimaryKey,
+            boolean isPartitioned,
+            RowType sourceOutputType,
+            @Nullable int[] projectedFields,
+            OffsetsInitializer offsetsInitializer,
+            long scanPartitionDiscoveryIntervalMs,
+            FlussDeserializationSchema<OUT> deserializationSchema,
+            boolean streaming,
+            @Nullable List<PushdownUtils.FieldEqual> partitionFilters) {
```
swuferhong: import the fully class name. Change `PushdownUtils.FieldEqual` to `FieldEqual`

fluss-flink/fluss-flink-common/src/main/java/com/alibaba/fluss/flink/source/FlinkSource.java
```diff
@@ -73,6 +78,32 @@ public FlinkSource(
...
+                tablePath,
+                hasPrimaryKey,
+                isPartitioned,
+                sourceOutputType,
+                projectedFields,
+                offsetsInitializer,
+                scanPartitionDiscoveryIntervalMs,
+                deserializationSchema,
+                streaming,
+                null);
```
swuferhong: Maybe we don’t need to add a separate constructor here, otherwise we’ll end up with too many constructors as the number of parameters increases.

fluss-flink/fluss-flink-common/src/main/java/com/alibaba/fluss/flink/source/FlinkTableSource.java
```diff
@@ -124,6 +125,8 @@ public class FlinkTableSource
 
     private long limit = -1;
 
+    @Nullable private List<PushdownUtils.FieldEqual> partitionFilters;
```
swuferhong: ditto, import the fully class name

fluss-flink/fluss-flink-common/src/main/java/com/alibaba/fluss/flink/source/FlinkSource.java
```diff
@@ -62,6 +65,8 @@ public class FlinkSource<OUT>
     private final boolean streaming;
     private final FlussDeserializationSchema<OUT> deserializationSchema;
 
+    @Nullable private final List<PushdownUtils.FieldEqual> partitionFilters;
```
swuferhong: There is no need to declare the list array as @Nullable, as this would simplify the conditional checks in `applyPartitionFilter()` for `partitionFilters != null && !partitionFilters.isEmpty()`.

fluss-flink/fluss-flink-common/src/main/java/com/alibaba/fluss/flink/source/enumerator/FlinkSourceEnumerator.java
```diff
@@ -263,6 +271,35 @@ private Set<PartitionInfo> listPartitions() {
...
+        if (partitionFilters != null && !partitionFilters.isEmpty()) {
+            return partitionInfos.stream()
+                    .filter(
+                            partitionInfo -> {
+                                Map<String, String> specMap =
+                                        partitionInfo.getPartitionSpec().getSpecMap();
+                                for (PushdownUtils.FieldEqual filter : partitionFilters) {
+                                    String partitionValue =
+                                            specMap.get(
+                                                    tableInfo
```
swuferhong: Move  `tableInfo.getRowType().getFieldNames()` out of for loop to reduce redundant call

fluss-flink/fluss-flink-common/src/main/java/com/alibaba/fluss/flink/source/FlinkTableSource.java
```diff
@@ -468,6 +487,30 @@ private Map<Integer, LogicalType> getPrimaryKeyTypes() {
...
+
+    private List<PushdownUtils.FieldEqual> serializeFieldEquals(
+            List<PushdownUtils.FieldEqual> fieldEquals) {
+        List<PushdownUtils.FieldEqual> serialize = new ArrayList<>();
+        for (PushdownUtils.FieldEqual fieldEqual : fieldEquals) {
+            if (fieldEqual.equalValue instanceof BinaryStringData) {
+                serialize.add(
+                        new PushdownUtils.FieldEqual(
+                                fieldEqual.fieldIndex,
+                                ((BinaryStringData) fieldEqual.equalValue).toString()));
```
swuferhong: remove redundant case `(BinaryStringData)`

swuferhong (commented):
Hi, @Alibaba-HZY. LGTM over all, but I left some comments.

fluss-flink/fluss-flink-common/src/main/java/com/alibaba/fluss/flink/source/FlinkSource.java
HZY-Wade: but in FlinkSource i declare  @Nullable ,do you mean that the null I transfer in was changed to Collections.emptyList()?

fluss-flink/fluss-flink-common/src/test/java/com/alibaba/fluss/flink/source/testutils/FlinkTestBase.java
HZY-Wade: it's a little difficult to make it more generic so i move  it to FlinkTableSourceITCase

fluss-flink/fluss-flink-common/src/main/java/com/alibaba/fluss/flink/source/FlinkTableSource.java
```diff
@@ -378,41 +384,55 @@ public void applyProjection(int[][] projectedFields, DataType producedDataType)
...
+        } else if (isPartitioned()) {
+            Map<Integer, LogicalType> partitionKeyTypes = getPartitionKeyTypes();
+            List<FieldEqual> fieldEquals =
+                    PushdownUtils.extractFieldEquals(
+                            filters,
+                            partitionKeyTypes,
+                            acceptedFilters,
+                            remainingFilters,
+                            ValueConversion.FLINK_INTERNAL_VALUE);
+            fieldEquals = serializeFieldEquals(fieldEquals);
```
luoyuxia: I' wondering why need to call `serializeFieldEquals`, I guess Flink's BinaryStringData doesn't implement `Serializable`, if so, we can use Fluss's `BinaryString`, which is Serializable.

fluss-flink/fluss-flink-common/src/main/java/com/alibaba/fluss/flink/source/FlinkTableSource.java
```diff
@@ -378,41 +384,55 @@ public void applyProjection(int[][] projectedFields, DataType producedDataType)
...
+            return Result.of(acceptedFilters, remainingFilters);
+        } else if (isPartitioned()) {
+            Map<Integer, LogicalType> partitionKeyTypes = getPartitionKeyTypes();
+            List<FieldEqual> fieldEquals =
+                    PushdownUtils.extractFieldEquals(
+                            filters,
+                            partitionKeyTypes,
+                            acceptedFilters,
+                            remainingFilters,
+                            ValueConversion.FLINK_INTERNAL_VALUE);
```
luoyuxia: Then, use `FLUSS_INTERNAL_VALUE`

fluss-flink/fluss-flink-common/src/test/java/com/alibaba/fluss/flink/source/FlinkTableSourceITCase.java
```diff
@@ -928,6 +929,93 @@ void testLookupFullCacheThrowException() {
...
+        tEnv.executeSql("alter table partitioned_table add partition (c=2025)");
+        tEnv.executeSql("alter table partitioned_table add partition (c=2026)");
+
+        List<String> expectedRowValues =
+                writeRowsToPartition(tablePath, Arrays.asList("2025", "2026")).stream()
+                        .filter(s -> s.contains("2025"))
+                        .collect(Collectors.toList());
+        waitUtilAllBucketFinishSnapshot(admin, tablePath, Arrays.asList("2025", "2026"));
+
+        org.apache.flink.util.CloseableIterator<Row> rowIter =
```
luoyuxia, line 954: also add test case to verify a query with partition filter and other column filter to check combine pushed down filters and non-pushed down filters can work correctly.

fluss-flink/fluss-flink-common/src/main/java/com/alibaba/fluss/flink/source/FlinkTableSource.java
HZY-Wade: Although BinaryString is Serializable，but it extends BinarySection, it's MemorySegment is No-Serializable
,so we must use String

luoyuxia: Haven't been aware it implement `Serializable` but not Serializable.

Alibaba-HZY:
> Thanks for the pr. Left minor comments. Otherwsie, LTM

Are there any other problems？

wuchong (approved):
Thanks @Alibaba-HZY , the PR looks good to me. I appended a minor commit to improve the code a bit.

PR #938, opened 2025-05-21
[flink] Allow to pass filesystem configurations to Fluss client

### LINKED ISSUE #935: Allow to pass filesystem configurations to Fluss client
### Search before asking

### Motivation

Fluss client will also need to access filesystem. But now, it will always use default configurations in fluss client since we never pass configuration to init the filesystem. We should allow to pass filesystem releated configuration to fluss client just like options `table.*`, `client.*`.

### Solution

_No response_

### Anything else?

_No response_

### Willingness to contribute

luoyuxia:
### Purpose

Linked issue: close #935 

To allow to pass filesystem configurations to Fluss client, so that Fluss client can use the `Filesystem` configured by user's configurations.

### Brief change log
- allow prefix `fs` as Flink's connector options, all our filesystem will accept configurations with `fs` prefix
- use `FlussConnection` configurations to init `Filesystem`

### Tests

- `testFileSystemRecognizeConnectionConf` to verify the configurations of `FlussConnection` can be passed to got `Filesystem`
- Add option in `testPkTableReadOnlySnapshot`  verify the option with `fs` prefix can be as Flink table's option


wuchong:
Sorry @luoyuxia , I think exposing all the `fs.*` is a dangerous behavior, because it seems user can overwrite all the filesystem configurations like `*.endpoint`, `*.accessKeySecret`, `*.roleArn`, `*.credentials.provider`. At least, it is not an expected to expose them, and it's hard to withdram them once exposed. Therefore, I recommend exposing only the necessary configuration by introducing `client.fs.*` options. What do you think?

luoyuxia:
> Sorry @luoyuxia , I think exposing all the `fs.*` is a dangerous behavior, because it seems user can overwrite all the filesystem configurations like `*.endpoint`, `*.accessKeySecret`, `*.roleArn`, `*.credentials.provider`. At least, it is not an expected to expose them, and it's hard to withdram them once exposed. Therefore, I recommend exposing only the necessary configuration by introducing `client.fs.*` options. What do you think?

Agree that it's dangerous to expose all.  So, in the pr, we just introduce the support to forward `client.fs.*` options in flink connector to filesystem. And the the future, add the real `client.fs.*` options, such as `client.fs.oss.multipart.download.threads` according to our needs. Is it right?

luoyuxia:
> > Sorry @luoyuxia , I think exposing all the `fs.*` is a dangerous behavior, because it seems user can overwrite all the filesystem configurations like `*.endpoint`, `*.accessKeySecret`, `*.roleArn`, `*.credentials.provider`. At least, it is not an expected to expose them, and it's hard to withdram them once exposed. Therefore, I recommend exposing only the necessary configuration by introducing `client.fs.*` options. What do you think?
> 
> Agree that it's dangerous to expose all. So, in the pr, we just introduce the support to forward `client.fs.*` options in flink connector to filesystem. And the the future, add the real `client.fs.*` options, such as `client.fs.oss.multipart.download.threads` according to our needs. Is it right?

Now, just forward `client.fs.*` prefix options to Filesystem to make the Filesystem pick the `client.fs.*`. @wuchong Pr updated not.

PR #939, opened 2025-05-21
[filesystem/oss] OssFileSystem should allow to configure credential provider

### LINKED ISSUE #936: OssFIlesystem should allow to configure credential provider
### Search before asking

### Description

Currently, we must need to configure ak/sk for oss filesystem in `server.yaml`, but it's not always safe to export it in `server.yaml`.  We need to allow user to configure their credential provider so that user can use `EnvironmentVariableCredentialsProvider` instead of export it in `server.yaml`.

### Willingness to contribute

luoyuxia:
…

### Purpose

Linked issue: close #936 

To allow OssFileSystem to configure credential provider

### Brief change log
if credential provider is set, not to always use `DynamicTemporaryOssCredentialsProvider`, use user's configured provider

### Tests

`OSSWithCredentialsProviderFileSystemBehaviorITCase`


luoyuxia:
@wuchong @loserwang1024 Allow to set credential provider to avoid expose ak/sk directly. Pr ready

wuchong (approved):
@luoyuxia thank you for the explanation.

PR #940, opened 2025-05-22
[doc] Fluss and Paimon data type mapping document adds binary and bytes data types

zhuangchong:
### Purpose

Fluss and Paimon data type mapping document adds binary and bytes data types


luoyuxia (approved):
Thanks.. LGTM

Issue #947, opened 2025-05-24
The initial capacity is inconsistent with the actual required capacity


zhaomin1423:
### Search before asking

- [x] I searched in the [issues](https://github.com/alibaba/fluss/issues) and found nothing similar.

### Fluss version

main (development)

### Please describe the bug 🐞

There are actually three elements, but the initial capacity is set to 2

### Are you willing to submit a PR?

- [x] I'm willing to submit a PR!

PR #948, opened 2025-05-24
The initial capacity is inconsistent with the actual required capacity

zhaomin1423:
### Purpose

Linked issue: close [#947 ](https://github.com/alibaba/fluss/issues/947#issue-3088609832)

### DIFF
--- a/fluss-server/src/main/java/com/alibaba/fluss/server/coordinator/CoordinatorServer.java
+++ b/fluss-server/src/main/java/com/alibaba/fluss/server/coordinator/CoordinatorServer.java
@@ -314,7 +314,7 @@ public class CoordinatorServer extends ServerBase {
                 exception = ExceptionUtils.firstOrSuppressed(t, exception);
             }
 
-            final Collection<CompletableFuture<Void>> terminationFutures = new ArrayList<>(2);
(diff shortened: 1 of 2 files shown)


fluss-server/src/main/java/com/alibaba/fluss/server/coordinator/CoordinatorServer.java
```diff
@@ -314,7 +314,7 @@ CompletableFuture<Void> stopServices() {
                 exception = ExceptionUtils.firstOrSuppressed(t, exception);
             }
 
-            final Collection<CompletableFuture<Void>> terminationFutures = new ArrayList<>(2);
+            final Collection<CompletableFuture<Void>> terminationFutures = new ArrayList<>(3);
```
luoyuxia, line 317: Thanks for the pr. But I don't think it does matter if  initial capacity is inconsistent, in the future, we may still add a new terminationFuture,  it may well come into inconsistent then since no any hard check for keep initial capacity consistent.

zhaomin1423: Thanks, I will close it.

PR #954, opened 2025-05-27
[Flink] Change logfetchercollector warning and timeout closableiterator

MehulBatra:
### Purpose

FLINK - CI Times out randomly due to Job failure/resource clean up issue.

Linked issue: close #934 

### Brief change log

> LogFetchCollector Changes:

Before: When we got an “unknown table or bucket” error, we would just log a warning and update metadata. Then the system would keep trying to fetch from that same non-existent bucket over and over again.

After: Now when we get an `UNKNOWN_TABLE_OR_BUCKET_EXCEPTION` error, we first check if the bucket is still assigned in our scanner status. If it is, we log the warning AND immediately remove that bucket from the scanner status so we stop trying to fetch from it. If the bucket was already unassigned, we just log a debug message instead of a warning to reduce noise.

>  AssertResultsIgnoreOrder utility method that reads results from Flink job iterators:

 
Before: The method would call `iterator.next()` in a simple for-loop, expecting to get exactly the number of records we wanted. But `iterator.next()` blocks forever if there’s no more data coming.

After: Now we check `iterator.hasNext()` before calling next(), and if there’s no data available, we wait up to 10 seconds. If we still don’t get data after the timeout, we assume the job finished normally and stop waiting.

### DIFF
--- a/fluss-client/src/main/java/com/alibaba/fluss/client/table/scanner/log/LogFetchCollector.java
+++ b/fluss-client/src/main/java/com/alibaba/fluss/client/table/scanner/log/LogFetchCollector.java
@@ -259,7 +259,21 @@ public class LogFetchCollector {
                     "Error in fetch for bucket {}: {}:{}", tb, error.exceptionName(), errorMessage);
             metadataUpdater.checkAndUpdateMetadata(tablePath, tb);
         } else if (error == Errors.UNKNOWN_TABLE_OR_BUCKET_EXCEPTION) {
-            LOG.warn("Received unknown table or bucket error in fetch for bucket {}", tb);
+            // ADD: Check if bucket should be removed from scanner status to stop repeated fetch
+            // attempts
+            Long currentOffset = logScannerStatus.getBucketOffset(tb);
+            if (currentOffset != null) {
+                LOG.warn(
+                        "Received unknown table or bucket error in fetch for bucket {}, removing from scanner status",
+                        tb);
+                // Remove the bucket from scanner status to prevent repeated fetch attempts
+                logScannerStatus.unassignScanBuckets(Collections.singletonList(tb));
+            } else {
+                // Bucket already unassigned, just log at debug level to reduce noise
+                LOG.debug(
+                        "Received unknown table or bucket error for already unassigned bucket {}",
+                        tb);
+            }
             metadataUpdater.checkAndUpdateMetadata(tablePath, tb);
         } else if (error == Errors.LOG_OFFSET_OUT_OF_RANGE_EXCEPTION) {
             throw new FetchException(
--- a/fluss-client/src/main/java/com/alibaba/fluss/client/table/scanner/log/LogFetchCollector.java
+++ b/fluss-client/src/main/java/com/alibaba/fluss/client/table/scanner/log/LogFetchCollector.java
@@ -249,7 +249,6 @@ public class LogFetchCollector {
             CompletedFetch completedFetch, Errors error, String errorMessage) {
         TableBucket tb = completedFetch.tableBucket;
         long fetchOffset = completedFetch.nextFetchOffset();
-
         if (error == Errors.NOT_LEADER_OR_FOLLOWER
                 || error == Errors.LOG_STORAGE_EXCEPTION
                 || error == Errors.KV_STORAGE_EXCEPTION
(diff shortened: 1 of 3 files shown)


fluss-flink/fluss-flink-common/src/test/java/com/alibaba/fluss/flink/source/testutils/FlinkRowAssertionsUtils.java
```diff
@@ -49,5 +82,51 @@ public static void assertResultsIgnoreOrder(
...
+            assertThat(actual).containsExactlyElementsOf(expected);
+
+        } catch (InterruptedException e) {
+            Thread.currentThread().interrupt();
+            throw new RuntimeException("Test interrupted", e);
+        } catch (Exception e) {
+            // Handle job completion gracefully
+            if (e.getCause() instanceof IllegalStateException
+                    && e.getMessage() != null
+                    && e.getMessage().contains("MiniCluster")) {
```
wuchong: Why it throws exception when the job completes gracefully?

fluss-flink/fluss-flink-common/src/test/java/com/alibaba/fluss/flink/source/testutils/FlinkRowAssertionsUtils.java
```diff
@@ -49,5 +82,51 @@ public static void assertResultsIgnoreOrder(
...
+                    Row r = rowIter.next();
+                    String row = r.toString();
+                    actual.add(row);
+                } else {
+                    // No more records available
+                    break;
+                }
+            }
+
+            assertThat(actual).containsExactlyElementsOf(expected);
```
wuchong: Could you add a fail message for the assertion that how long we have waited for the job result? This would be helpful to identify the case of flink job hang.

fluss-flink/fluss-flink-common/src/test/java/com/alibaba/fluss/flink/source/testutils/FlinkRowAssertionsUtils.java
```diff
@@ -49,5 +82,51 @@ public static void assertResultsIgnoreOrder(
         }
     }
 
-    // Add other shared utility methods here
+    // expects exact order
+    public static void assertQueryResultExactOrder(
```
wuchong: I think we can share most of the common logic with `assertResultsIgnoreOrder`? Because it seems the only difference is the compare order.

fluss-client/src/main/java/com/alibaba/fluss/client/table/scanner/log/LogFetchCollector.java
```diff
@@ -259,7 +259,21 @@ private void handleInitializeErrors(
...
-            LOG.warn("Received unknown table or bucket error in fetch for bucket {}", tb);
+            // Check if bucket should be removed from scanner status to stop repeated fetch
+            // attempts.
+            Long currentOffset = logScannerStatus.getBucketOffset(tb);
+            if (currentOffset != null) {
+                LOG.warn(
+                        "Received unknown table or bucket error in fetch for bucket {}, removing from scanner status",
+                        tb);
+                // Remove the bucket from scanner status to prevent repeated fetch attempts
+                logScannerStatus.unassignScanBuckets(Collections.singletonList(tb));
```
wuchong: I don't think we should unassign the bucket, because `UNKNOWN_TABLE_OR_BUCKET_EXCEPTION` is a common retriable exception that occurs when the replica leader is not yet ready on the given TabletServer. This typically happens during reads right after a table is created. If we unassign the bucket in this case, we risk losing the ability to read data from it.

MehulBatra: I will revert the unassigned change. I did it to avoid log spam for `UNKNOWN_TABLE_OR_BUCKET_EXCEPTION`, but yes, it could lead to another problem, as you stated, I will change that accordingly.

fluss-flink/fluss-flink-common/src/test/java/com/alibaba/fluss/flink/source/testutils/FlinkRowAssertionsUtils.java
MehulBatra: Job Completion Signal: When a Flink job processes all available data (finite dataset), it completes successfully
MiniCluster Shutdown: The MiniCluster automatically shuts down resources after job completion
Iterator Access: The test code continues trying to read from CloseableIterator<Row>
Resource Unavailable: The iterator tries to access already-closed MiniCluster resources
Exception Thrown: Instead of returning hasNext() = false, Flink throws an IllegalStateException

I guess Flink's iterator doesn't have a clean way to signal "job completed" for finite jobs. It discovers the job ended by getting an error when trying to communicate with the (now shut down) execution environment.

MehulBatra: Log: "MiniCluster is not yet running or has already been shut down"
It suggests the iterator tried to communicate with a cluster that's gone.

fluss-flink/fluss-flink-common/src/test/java/com/alibaba/fluss/flink/source/testutils/FlinkRowAssertionsUtils.java
MehulBatra: Sure

fluss-flink/fluss-flink-common/src/test/java/com/alibaba/fluss/flink/source/testutils/FlinkRowAssertionsUtils.java
MehulBatra: Sure

wuchong (approved):

MehulBatra:
Thank you for the review @wuchong !

PR #961, opened 2025-05-28
[lake/paimon] Add system columns for primary key table

luoyuxia:
### Purpose

Linked issue: close #xxx


fluss-lake/fluss-lake-paimon/src/main/java/com/alibaba/fluss/lake/paimon/PaimonLakeCatalog.java
```diff
@@ -45,6 +45,15 @@
 /** A Paimon implementation of {@link LakeCatalog}. */
 public class PaimonLakeCatalog implements LakeCatalog {
 
+    private static final LinkedHashMap<String, DataType> SYSTEM_COLUMNS =
+            new LinkedHashMap<String, DataType>() {
+                {
+                    put(BUCKET_COLUMN_NAME, DataTypes.INT());
```
luoyuxia: Although bucket system column is actual not needed for primary key table in our case. 
We just want to use bucket system column to help us filter given bucket files. But paimon must be bucket aware for primary key table. So, we can just use paimon's api to help use filter by bucket. 
But bucket system column is required for bucket-unware append-only table. 

So, in here, we always add bucket system column to keep the consistency

leonardBang: Could you add this ad java doc in code ?

And about the PR title，how about "[lake/paimon] Automatically append system columns for Fluss primary key table when create paimon table" ?

luoyuxia: @leonardBang Thanks for suggestion. Comments addressed.

PR #974, opened 2025-05-30
[bugfix] Fix a bug which makes memory allocation stuck

zcoo:
### Purpose
When `public void returnAll(List<MemorySegment> memory) {`
returns memory segment, it polls the first waiter and wakes it up in Deque to judge if freePages is less than requiredPages. If not, this waiter calls await() again and will no longer be waked up. This should cause memory allocation stuck and can't recover.


swuferhong (approved):
LGTM

Issue #757, opened 2025-04-17
UpdateMetadata RPC should update table path cache


wuchong:
### Search before asking

- [x] I searched in the [issues](https://github.com/alibaba/fluss/issues) and found nothing similar.

### Description

Currently, `UpdateMetadata` RPC only updates coordinator and tablet server informations in the cluster, but we should also update table path information in the cluster. This has already added in the RPC protocol, but didn't implement in Coordinator side and TabletServer side. 

### Willingness to contribute

- [ ] I'm willing to submit a PR!

wuchong:
@caozhen1937 Getting all `tablePath` by `MetadataManager` is a very heavy operation for zookeeper and should be avoided. The implementation is a little complex. We need to send `updateMetadata` for every table creation and deletion and also when the initialization of a coordiinator server. 

This task is in a high priority, thus I have been working it. Thanks for volunteering.

wuchong:
cc @swuferhong

PR #987, opened 2025-06-03
[hotfix] Remove residual tablePath in coordinatoContext when drop a partition table without any partition

### LINKED ISSUE #898: Drop a partitionTable without creating any partition will result in residual tablePath in coordinatoContext
### Search before asking

### Fluss version

main (development)

### Please describe the bug 🐞

Currently, if we try to drop a `partitionTable` without creating any partition will result in residual tablePath in coordinatoContext. The reason why is that for non-partitioned table, we will trigger `Table.resumeDeletions` while all buckets have been deleted. However, for partitioned table without creating any partition, there are no buckets to drop and `Table.resumeDeletions` will be not trigger.

### Solution

_No response_

zcoo:
…artitionTable without any partition

### Purpose
Remove residual tablePath in coordinatoContext when drop a partition table without any partition

Linked issue: close #898

### Tests

TableManagerITCase::testPartitionedTableManagement


swuferhong:
Hi, @zcoo after #757 has been merged, can you check  whether `MetadataUpdateITCase#testMetadataUpdateForTableCreateAndDrop` works after your bug fix.  Additionally, this is not a hotfix. The title and PR comments should specify the relevant module, such as [server] in this case.

zcoo:
@swuferhong I have resolved conflicts and this test case seems work well. But there are some other failed case. I will fix

fluss-server/src/main/java/com/alibaba/fluss/server/coordinator/CoordinatorEventProcessor.java
```diff
@@ -620,6 +620,10 @@ private void processDropTable(DropTableEvent dropTableEvent) {
             lakeTableTieringManager.removeLakeTable(tableId);
         }
 
+        // Try to resume in case of no replica was deleted but table was deleted
+        // for example a partitioned table without any partition.
+        tableManager.resumeDeletions();
```
luoyuxia, line 625: It'll always remove the tableid from context while some replicas for the partition of the table still exists.  It cause if the replica of partition delete fail and retry, it can't get the table path of the table from context, which casue the retry fail by missing table path in the context.  
```
if (coordinatorContext.areAllReplicasInState(
                    tableId, ReplicaState.ReplicaDeletionSuccessful)) {
                completeDeleteTable(tableId);
                LOG.info("Deletion of table with id {} successfully completed.", tableId);
            }
```  
There's no replica for the table id since we only record replicas for partition id of the table. 
So, it'll always consider it as delete successfully even though some  replicas for the partition of the table still exist.

PR #1002, opened 2025-06-05
[server] Support auto create partition when write

swuferhong:
### Purpose

Linked issue:  https://github.com/alibaba/fluss/issues/583

This is pr is based on #893


swuferhong:
@wuchong comments addressed.

fluss-common/src/main/java/com/alibaba/fluss/config/ConfigOptions.java
```diff
@@ -918,6 +918,15 @@ public class ConfigOptions {
                                     + "requests per bucket exceeds this setting, the writer will wait for the inflight "
                                     + "requests to complete before sending out new requests.");
 
+    public static final ConfigOption<Boolean> CLIENT_WRITER_DYNAMIC_PARTITION_ENABLED =
+            key("client.writer.dynamic-partition.enabled")
```
wuchong: -> `client.writer.dynamic-create-partition.enabled`
I found it is confusing of dynamic-partition and auto-partition, so let's add `create` keyword into the option key to make it clear.

fluss-client/src/main/java/com/alibaba/fluss/client/write/WriteBatch.java
```diff
@@ -133,16 +131,16 @@ public void abort(Exception exception) {
                     "Batch has already been completed in final stata " + finalState.get());
         }
 
-        LOG.trace("Abort batch for tableBucket {}", tableBucket, exception);
+        LOG.trace("Abort batch for table path {}", physicalTablePath, exception);
```
wuchong: Add the bucket id information to the log

fluss-client/src/test/java/com/alibaba/fluss/client/table/PartitionedTableITCase.java
```diff
@@ -130,6 +141,97 @@ void testPartitionedLogTable() throws Exception {
...
+                waitValue(
+                        () -> {
+                            List<PartitionInfo> partitionInfos =
+                                    admin.listPartitionInfos(DATA1_TABLE_PATH_PK).get();
+                            if (partitionInfos.size() == partitionSize) {
+                                return Optional.of(partitionInfos);
+                            }
+                            {
+                                return Optional.empty();
+                            }
```
wuchong:
```suggestion
                                return Optional.empty();
```

fluss-client/src/main/java/com/alibaba/fluss/client/write/RecordAccumulator.java
```diff
@@ -365,12 +384,21 @@ public void deallocate(WriteBatch batch) {
...
+    public Deque<WriteBatch> getDeque(
+            PhysicalTablePath path, int bucketId, boolean ignoreUnReadyPartitionTable) {
         BucketAndWriteBatches bucketAndWriteBatches = writeBatches.get(path);
         if (bucketAndWriteBatches == null) {
             return null;
         }
-        return bucketAndWriteBatches.batches.get(tableBucket.getBucket());
+
+        // for these partitioned tables, we need to check partitionId first.
+        if (!ignoreUnReadyPartitionTable
```
wuchong: Should reverse the condition? If we ignore unready table, we should check the partition id.

fluss-client/src/main/java/com/alibaba/fluss/client/write/RecordAccumulator.java
```diff
@@ -365,12 +384,21 @@ public void deallocate(WriteBatch batch) {
     }
 
     @VisibleForTesting
-    public Deque<WriteBatch> getDeque(PhysicalTablePath path, TableBucket tableBucket) {
+    public Deque<WriteBatch> getDeque(
+            PhysicalTablePath path, int bucketId, boolean ignoreUnReadyPartitionTable) {
```
wuchong: I'm not in favor of introducing flags to methods because they can make the code less self-explanatory. Instead, we can introduce two separate methods — `getReadyDeque` and `getDeque` — each serving a clear and distinct purpose.

swuferhong:
@wuchong comments addressed.

PR #1004, opened 2025-06-05
[server] Re-register server-node when ZkClient is session expired

### LINKED ISSUE #1000: Register server node into zookeeper again when zkclient reconnect
### Search before asking

### Motivation

When zkclient session timeout,  the temporary server node in zk will also be removed. 
Although it'll reconnect, but it'll never register the server node again into zk

### Solution

We should re-register server node into zookeeper when zkclient reconnect

### Anything else?

_No response_

### Willingness to contribute

luoyuxia:
### Purpose

Linked issue: close #1000 

### Brief change log
To re-register server node when zkClient is session expired

### DIFF
--- a/fluss-server/src/main/java/com/alibaba/fluss/server/coordinator/CoordinatorServer.java
+++ b/fluss-server/src/main/java/com/alibaba/fluss/server/coordinator/CoordinatorServer.java
@@ -42,6 +42,7 @@ import com.alibaba.fluss.server.metrics.group.CoordinatorMetricGroup;
 import com.alibaba.fluss.server.zk.ZooKeeperClient;
 import com.alibaba.fluss.server.zk.ZooKeeperUtils;
 import com.alibaba.fluss.server.zk.data.CoordinatorAddress;
+import com.alibaba.fluss.shaded.zookeeper3.org.apache.zookeeper.KeeperException;
 import com.alibaba.fluss.utils.ExceptionUtils;
 import com.alibaba.fluss.utils.ExecutorUtils;
 import com.alibaba.fluss.utils.concurrent.ExecutorThreadFactory;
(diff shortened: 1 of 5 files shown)


fluss-server/src/main/java/com/alibaba/fluss/server/coordinator/CoordinatorServer.java
```diff
@@ -282,6 +285,47 @@ private void registerCoordinatorLeader() throws Exception {
         zkClient.registerCoordinatorLeader(coordinatorAddress);
     }
 
+    private void registerZookeeperClientReconnectedListener() {
+        ZooKeeperUtils.registerZookeeperClientReInitSessionListener(
+                zkClient,
+                () -> {
+                    // we need to retry to register since although
```
luoyuxia, line 292: Also see [KAFKA-7165](https://issues.apache.org/jira/browse/KAFKA-7165), the ephemeral node may still exist. Kafka solve it by droping  existing ephemeral node if the node's session id is less than the current seesion id. 
I think it's kind of complex and hard to understand. So, in here, I just retry when the node exists. Retry 1 minute should work for almost all case.

fluss-server/src/main/java/com/alibaba/fluss/server/zk/ZooKeeperUtils.java
```diff
@@ -148,6 +151,53 @@ public static CuratorFrameworkWithUnhandledErrorListener startZookeeperClient(
...
+                ThrowingRunnable<Exception> reInitSessionCallback,
+                FatalErrorHandler fatalErrorHandler) {
+            this.reInitSessionCallback = reInitSessionCallback;
+            this.fatalErrorHandler = fatalErrorHandler;
+        }
+
+        public void stateChanged(
+                CuratorFramework curatorFramework, ConnectionState connectionState) {
+            switch (connectionState) {
+                case LOST:
```
luoyuxia, line 181: See https://curator.apache.org/apidocs/org/apache/curator/framework/state/ConnectionState.html

When lost, it means a session is expired.

PR #1047, opened 2025-06-09
[server] Optimize auto drop partition perfomance

### LINKED ISSUE #864: Optimize auto drop partition perfomance
### Search before asking

### Motivation

With multi-level partition keys, it is not efficient to handle large partition set because current implement should scan all partition names. 

### Solution
We may sort partition names by partition time key and thus directly drop partitions with time value older than expected.
_No response_

### Anything else?

_No response_

### Willingness to contribute

LB-Yu:
### Purpose

Linked issue: close #864

### DIFF
--- a/fluss-server/src/main/java/com/alibaba/fluss/server/coordinator/AutoPartitionManager.java
+++ b/fluss-server/src/main/java/com/alibaba/fluss/server/coordinator/AutoPartitionManager.java
@@ -49,11 +49,13 @@ import java.time.ZonedDateTime;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HashMap;
+import java.util.HashSet;
+import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
-import java.util.NavigableSet;
+import java.util.NavigableMap;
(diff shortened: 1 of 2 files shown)


wuchong:
cc @swuferhong , could you help to review this?

fluss-server/src/main/java/com/alibaba/fluss/server/coordinator/AutoPartitionManager.java
```diff
@@ -95,8 +97,10 @@ public class AutoPartitionManager implements AutoCloseable {
     // now, only consider day partition, todo: need to consider all partition unit
     private final Map<Long, Integer> autoCreateDayPartitionDelayMinutes = new HashMap<>();
 
+    // table id -> (auto partition value -> partition name set)
```
zcoo: table id -> (value of auto partition time key -> partition name set)

LiebingYu: Yes, this is more accurate, I have modified it.

PR #1049, opened 2025-06-09
[row] improve test on VectorizedColumnBatch

### LINKED ISSUE #1048: Improve test coverage on VectorizedColumnBatch
### Search before asking

### Description

Improve test coverage on VectorizedColumnBatch

### Willingness to contribute

gyang94:
### Purpose

Linked issue: close #1048


gyang94:
@wuchong A quick question:
It will throw an exception when get the value in ColumnVector if the data is null. In my test I have to added `if (!row.isNullAt(3))` to pre-check the null value. Is it the right way? suppose user should be able get the "null" value instead of an exception.
e.g. In the test case, the value of `IntColumnVector` is [3, null]. It throws an exception when calling `IntColumnVector.getInt()`.

fluss-common/src/test/java/com/alibaba/fluss/row/arrow/ArrowReaderWriterTest.java
```diff
@@ -160,7 +160,42 @@ void testReaderWriter() throws IOException {
...
+                assertThat(row.getShort(2)).isEqualTo(rowData.getShort(2));
+                if (!row.isNullAt(3)) {
+                    assertThat(row.getInt(3)).isEqualTo(rowData.getInt(3));
+                }
+                assertThat(row.getLong(4)).isEqualTo(rowData.getLong(4));
+                assertThat(row.getFloat(5)).isEqualTo(rowData.getFloat(5));
+                assertThat(row.getDouble(6)).isEqualTo(rowData.getDouble(6));
+                assertThat(row.getDecimal(7, 10, 3)).isEqualTo(rowData.getDecimal(7, 10, 3));
+                assertThat(row.getChar(8, 3)).isEqualTo(rowData.getChar(8, 3));
+                if (!row.isNullAt(9)) {
```
wuchong, line 176: should check it's null for this position in `rowData`, if it's null in `row`.

wuchong:
@gyang94 yes, it's by design. It's invalid to call `getXxx(n)` if it's null `isNullAt(n)`, because many `getXxx` returns primitive types for better performance, avoiding boxing.

fluss-common/src/test/java/com/alibaba/fluss/row/arrow/ArrowReaderWriterTest.java
gyang94: resolved

wuchong: Please fix other nullable checks.

gyang94: resolved

PR #1065, opened 2025-06-11
[lake]Migrate Lake Read/Write interfaces to fluss-lake-common

### LINKED ISSUE #442: [fluss-connectors/fluss-connector-flink] Migrate Lake Read/Write interfaces to fluss-lake-common
### Search before asking

### Description

 Migrate Lake Read/Write interfaces to fluss-lake-common

### Willingness to contribute

caozhen1937:
### Purpose

Linked issue: close #442


caozhen1937:
Hi @leonardBang, please take a review when you have time.

luoyuxia:
IIUC, it's by design that all lake common interface should be in `fluss-common` instead of `fluss-lake-common` .  Anyway, other modules will  require  `fluss-common`,  it we put all to `fluss-common`, the lake-format moudle won't care to import `fluss-lake-common` to avoid import a more module to simplify use.

wuchong:
> IIUC, it's by design that all lake common interface should be in `fluss-common` instead of `fluss-lake-common` . Anyway, other modules will require `fluss-common`, it we put all to `fluss-common`, the lake-format moudle won't care to import `fluss-lake-common` to avoid import a more module to simplify use.

+1

It's a bit different with the original design during the implementation. Let's keep the module arch simple for now. Closing this PR...

PR #1076, opened 2025-06-12
[hotfix] Adjust the writeBatch estimate logic to avoid buffer size always tending to decrease

swuferhong:
### Purpose

This is a hotfix to adjust the writeBatch estimate logic to avoid buffer size always tending to decrease


swuferhong:
From the test results, it can be seen that the size of different batches will be different and not too small：
image

PR #1081, opened 2025-06-13
[server] Accelerate deleting table or partition

### LINKED ISSUE #1054: Accelerate the speed of deleting tables and partitions
### Search before asking

### Motivation

As we already introduce #952 (Accelerate the speed of creating tables and partitions), and this pr greatly increases the speed of creating tables, so we want to introduce the acceleration of deleting tables and partitions

### Solution

_No response_

### Anything else?

_No response_

### Willingness to contribute

zcoo:
### Purpose
Accelerate deleting tables and partitions by merge zk update requests.

Linked issue: close #1054

### DIFF
--- a/fluss-server/src/main/java/org/apache/fluss/server/coordinator/CoordinatorEventProcessor.java
+++ b/fluss-server/src/main/java/org/apache/fluss/server/coordinator/CoordinatorEventProcessor.java
@@ -700,7 +700,7 @@ public class CoordinatorEventProcessor implements EventProcessor {
     private void processNewTabletServer(NewTabletServerEvent newTabletServerEvent) {
         // NOTE: we won't need to detect bounced tablet servers like Kafka as we won't
         // miss the event of tablet server un-register and register again since we can
-        // listener the children created and deleted in zk node.
+        // listen the children created and deleted in zk node.
 
         // Also, Kafka use broker epoch to make it can reject the LeaderAndIsrRequest,
         // UpdateMetadataRequest and StopReplicaRequest
--- a/fluss-server/src/main/java/org/apache/fluss/server/coordinator/statemachine/ReplicaStateMachine.java
+++ b/fluss-server/src/main/java/org/apache/fluss/server/coordinator/statemachine/ReplicaStateMachine.java
@@ -414,6 +414,7 @@ public class ReplicaStateMachine {
     private Map<TableBucketReplica, LeaderAndIsr> doRemoveReplicaFromIsr(
             Collection<TableBucketReplica> tableBucketReplicas) {
         Map<TableBucketReplica, LeaderAndIsr> adjustedLeaderAndIsr = new HashMap<>();
+        Map<TableBucket, LeaderAndIsr> toUpdateLeaderAndIsrList = new HashMap<>();
         for (TableBucketReplica tableBucketReplica : tableBucketReplicas) {
             TableBucket tableBucket = tableBucketReplica.getTableBucket();
             int replicaId = tableBucketReplica.getReplica();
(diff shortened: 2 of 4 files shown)


fluss-server/src/main/java/com/alibaba/fluss/server/zk/ZooKeeperClient.java
```diff
@@ -307,6 +307,37 @@ public void updateLeaderAndIsr(TableBucket tableBucket, LeaderAndIsr leaderAndIs
         LOG.info("Updated {} for bucket {} in Zookeeper.", leaderAndIsr, tableBucket);
     }
 
+    public void batchUpdateLeaderAndIsr(Map<TableBucket, LeaderAndIsr> batchUpdate)
```
swuferhong: Can we integrate this method into `ZookeeperClient`? It appears to be a generic method with substantial ZooKeeper-specific implementation details.

fluss-server/src/main/java/com/alibaba/fluss/server/coordinator/statemachine/ReplicaStateMachine.java
```diff
@@ -263,7 +265,9 @@ private void doHandleStateChanges(
 
                 // then, may remove the offline replica from isr
                 Map<TableBucketReplica, LeaderAndIsr> adjustedLeaderAndIsr =
-                        doRemoveReplicaFromIsr(validReplicas);
+                        checkIfDropTableOrPartitionRequest(validReplicas)
+                                ? doBatchRemoveReplicaFromIsr(validReplicas)
```
swuferhong: Ask a question: Why don’t we design this to be more generic?  Instead of identifying specific operations (e.g., create table/partition, drop table/partition), why we don't merging buckets by table/partition and perform batch registration sequentially. If the cluster experiences significant ISR adjustment (e.g., frequent pod shutdown/starting), would batch registration become infeasible?

wuchong:
As discussed offline with @zcoo , we can refator the implementation to make the current `doRemoveReplicaFromIsr` directly to use zookeeper batch update method. This makes the optimization more general that not limit to drop cases.

zcoo:
@wuchong Thanks. I refactor the implementation. PLAT~

fluss-server/src/main/java/com/alibaba/fluss/server/coordinator/statemachine/ReplicaStateMachine.java
```diff
@@ -438,25 +439,21 @@ private Map<TableBucketReplica, LeaderAndIsr> doRemoveReplicaFromIsr(
...
-                continue;
-            }
-            // update leader and isr
-            coordinatorContext.putBucketLeaderAndIsr(tableBucket, adjustLeaderAndIsr);
             adjustedLeaderAndIsr.put(tableBucketReplica, adjustLeaderAndIsr);
+            toUpdateLeaderAndIsrList.put(tableBucket, adjustLeaderAndIsr);
+        }
+        try {
+            zooKeeperClient.batchUpdateLeaderAndIsr(toUpdateLeaderAndIsrList);
+            toUpdateLeaderAndIsrList.forEach(coordinatorContext::putBucketLeaderAndIsr);
```
wuchong: As we moved the `coordinatorContext` updates at the end, this will change the logic of this method, because it always gets the current `leaderAndIsr` from `coordinatorContext`. 

For example, given we have `(leader=1, isr=[1,2,3])` in zk and `coordinatorContext`, and the replicas to offline are `[1,2]`. When we process offline replica `1`, it puts `(leader=-1, isr=[2,3])` to the temporary map. When we process offline replica `2`, the `leaderAndIsr` in `coordinatorContext` is still `(leader=1, isr=[1,2,3])`,  it puts `(leader=1, isr=[1,3])` to the temporary map and overrides the previous `(leader=-1, isr=[2,3])`. In the end, we get the wrong state in zk and `coordinatorContext`. 

We need to get the current `leaderAndIsr` from the `toUpdateLeaderAndIsrList` first, and then get from `coordinatorContext` if not found. 

Please also add such a unit test in `ReplicaStateMachineTest` to cover this case.

fluss-server/src/test/java/com/alibaba/fluss/server/zk/ZooKeeperClientTest.java
```diff
@@ -211,14 +242,44 @@ void testBatchCreateLeaderAndIsr() throws Exception {
...
         zookeeperClient.batchRegisterLeaderAndIsrForTablePartition(partitionTableBucket);
         for (int i = 0; i < 100; i++) {
+            // each should register successful
             Optional<LeaderAndIsr> optionalLeaderAndIsr =
                     zookeeperClient.getLeaderAndIsr(partitionTableBucket.get(i).getTableBucket());
             assertThat(optionalLeaderAndIsr.isPresent()).isTrue();
             assertThat(optionalLeaderAndIsr.get()).isIn(partitionleaderAndIsrList);
         }
+
+        Map<TableBucket, LeaderAndIsr> partitionUpdateMap =
```
wuchong: This method is quite long. Please consider to move the partitioned table test to a new method as there is nothing reused with the non-partitioned test. If you can unify the test code using `@ParameterizedTest`, that would be great.

fluss-server/src/main/java/com/alibaba/fluss/server/coordinator/statemachine/ReplicaStateMachine.java
zcoo: yes I think I miss this scene. Now it is resolved

fluss-server/src/test/java/com/alibaba/fluss/server/zk/ZooKeeperClientTest.java
zcoo: ok I use `@ParameterizedTest` to refactor

zcoo:
@wuchong  all resolved~ PLAT

wuchong (approved):

PR #1096, opened 2025-06-13
[minor][lake] Set default name for flink tiering service job 

leonardBang:
### Purpose

Linked issue: close #xxx

### Brief change log

 Set default name for flink tiering service job

### DIFF
--- a/fluss-flink/fluss-flink-common/src/main/java/com/alibaba/fluss/flink/tiering/LakeTieringJobBuilder.java
+++ b/fluss-flink/fluss-flink-common/src/main/java/com/alibaba/fluss/flink/tiering/LakeTieringJobBuilder.java
@@ -42,6 +42,8 @@ import static com.alibaba.fluss.utils.Preconditions.checkNotNull;
 /** The builder to build Flink lake tiering job. */
 public class LakeTieringJobBuilder {
 
+    private static final String DEFAULT_TIERING_SERVICE_JOB_NAME = "Fluss_Tiering_Service_Job";
+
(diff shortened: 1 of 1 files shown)


fluss-flink/fluss-flink-common/src/main/java/com/alibaba/fluss/flink/tiering/LakeTieringJobBuilder.java
```diff
@@ -42,6 +42,8 @@
 /** The builder to build Flink lake tiering job. */
 public class LakeTieringJobBuilder {
 
+    private static final String DEFAULT_TIERING_SERVICE_JOB_NAME = "Fluss_Tiering_Service_Job";
```
wuchong: We don't need underline? How about `Fluss Lake Tiering Service`?

fluss-flink/fluss-flink-common/src/main/java/com/alibaba/fluss/flink/tiering/LakeTieringJobBuilder.java
```diff
@@ -107,6 +109,6 @@ public JobClient build() throws Exception {
                 .setMaxParallelism(1)
                 .sinkTo(new DiscardingSink());
 
-        return env.executeAsync();
+        return env.executeAsync(DEFAULT_TIERING_SERVICE_JOB_NAME);
```
luoyuxia: What if user specifiy `pipeline.name` ? Will it override user's specified name?

PR #1114, opened 2025-06-15
[client-tool] provide basic fluss Command Line framework

simonsssu:
### Purpose

Linked issue: close #1081

### DIFF
--- a/fluss-client-tools/pom.xml
+++ b/fluss-client-tools/pom.xml
@@ -1,0 +1,143 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+ Copyright (c) 2024 Alibaba Group Holding Ltd.
+
+ Licensed under the Apache License, Version 2.0 (the "License");
+ you may not use this file except in compliance with the License.
+ You may obtain a copy of the License at
+
+      http://www.apache.org/licenses/LICENSE-2.0
+
+ Unless required by applicable law or agreed to in writing, software
+ distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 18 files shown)


SteNicholas:
@simonsssu, thanks for contribution. I suggest that the CLI framework could use `picocli` to support.

simonsssu:
> @simonsssu, thanks for contribution. I suggest that the CLI framework could use `picocli` to support.

We've tried picocli before and found that it's a little inconvenient to pass configs from parent cmd to child cmd. so we choose JCommander.

wuchong:
> > @simonsssu, thanks for contribution. I suggest that the CLI framework could use `picocli` to support.
> 
> We've tried picocli before and found that it's a little inconvenient to pass configs from parent cmd to child cmd. so we choose JCommander.

@SteNicholas Is it possible to do this in picocli?

simonsssu:
While Picocli's @ParentCommand annotation simplifies dependency injection between parent and child commands, I remember when we do auto registerion poc, it introduces inconvenience during automatic command registration, as each child command must explicitly declare this annotation to access parent configurations

simonsssu:
@wuchong @SteNicholas Do you think we can provide a picocli solution and we can evalute which one is more appropriate solution

SteNicholas:
@simonsssu, could you explain which inconvenience passing configs from parent to child? From my experience, the picocli is enough to build the framework of CLI.

simonsssu:
@SteNicholas I remember in auto registerion logics. let me try to find out the previous implementation.

simonsssu:
Picocli solution: https://github.com/apache/fluss/pull/1510
@SteNicholas I resolved the issue faced before. The code is concise now.
cc @wuchong

PR #1132, opened 2025-06-17
[flink] Move Factory tests which depends on different version of Flink frameworks to submodule

### LINKED ISSUE #1131: Remove Factory tests which depends on different version of Flink frameworks to submodule
### Search before asking

### Description

In https://github.com/alibaba/fluss/issues/707, we move META-INF/services/org.apache.flink.table.factories.Factory to each submodule for two reason:
1. Submodule ( which user apply to their own Flink version) can freely extend FlinkCatalogFactory.
2. Each tests (which depends on flink framework rather than just UT case ) should also verified in each module. Because currently fluss-flink-common only works on newest version, maybe not work on older version.

However, in https://github.com/alibaba/fluss/pull/989, it only test factory with framework's FactoryUtil.createCatalog in fluss-flink-common.

In this PR, I do hope it can also test for each submodule.

### Willingness to contribute

loserwang1024:
…ink frameworks to submodule

### Purpose

Linked issue: close #1131


loserwang1024:
@wuchong @leonardBang , CC

PR #1171, opened 2025-06-20
[test] TableChangeWatcherTest.testTableChanges is not stable

Rohan-flutterint:
### Purpose

Linked issue: close [https://github.com/alibaba/fluss/issues/1018](url)


wuchong:
Thanks @rohan-flutterint. However, I still find the fix a bit hard to understand. The test is very lightweight and should complete in a short amount of time — adding a timeout shouldn't actually resolve the issue.

PR #1172, opened 2025-06-21
[feature] Support alter database comment and custom properties

### LINKED ISSUE #1151: [Feature] Support alter database to update comment and custom properties
### Search before asking

### Motivation

Support alter database to update comment and custom properties.

### Solution

- Server side: Add `alterDatabase` for `AdminGateway`, and implement in `CoordinatorService`.
- Client side: Add `alterDatabase` for `Admin`, and implement in `FlussAdmin`.
- Flink side: Implement `alterDatabase` in `FlinkCatalog`.

### Anything else?

_No response_

### Willingness to contribute

LB-Yu:
### Purpose

Linked issue: close #1151

### DIFF
--- a/fluss-rpc/src/main/proto/FlussApi.proto
+++ b/fluss-rpc/src/main/proto/FlussApi.proto
@@ -63,6 +63,16 @@ message CreateDatabaseRequest {
 message CreateDatabaseResponse {
 }
 
+// alter database request and response
+message AlterDatabaseRequest {
+  required string database_name = 1;
+  required bool ignore_id_not_exists = 2;
+  required bytes database_json = 3;
+}
+
+message AlterDatabaseResponse {
+}
+
 // get table request and response
 message GetDatabaseInfoRequest {
   required string database_name = 1;
--- a/fluss-client/src/main/java/com/alibaba/fluss/client/admin/Admin.java
+++ b/fluss-client/src/main/java/com/alibaba/fluss/client/admin/Admin.java
@@ -162,6 +162,24 @@ public interface Admin extends AutoCloseable {
     CompletableFuture<List<String>> listDatabases();
 
     /**
+     * Alter a database asynchronously.
+     *
+     * <p>The following exceptions can be anticipated when calling {@code get()} on returned future.
+     *
+     * <ul>
+     *   <li>{@link DatabaseNotExistException} if the database not exists and {@code
+     *       ignoreIfNotExists} is false.
+     * </ul>
+     *
+     * @param databaseName The name of the database to alter.
+     * @param databaseDescriptor The descriptor of the database to alter.
+     * @param ignoreIfNotExists Flag to specify behavior when a database with the given name not
+     *     exists: if set to false, throw a DatabaseNotExistException, if set to true, do nothing.
+     */
+    CompletableFuture<Void> alterDatabase(
+            String databaseName, DatabaseDescriptor databaseDescriptor, boolean ignoreIfNotExists);
+
+    /**
      * Create a new table asynchronously.
      *
      * <p>The following exceptions can be anticipated when calling {@code get()} on returned future.
--- a/fluss-client/src/main/java/com/alibaba/fluss/client/admin/FlussAdmin.java
+++ b/fluss-client/src/main/java/com/alibaba/fluss/client/admin/FlussAdmin.java
@@ -39,6 +39,7 @@ import com.alibaba.fluss.rpc.GatewayClientProxy;
 import com.alibaba.fluss.rpc.RpcClient;
 import com.alibaba.fluss.rpc.gateway.AdminGateway;
 import com.alibaba.fluss.rpc.gateway.TabletServerGateway;
+import com.alibaba.fluss.rpc.messages.AlterDatabaseRequest;
 import com.alibaba.fluss.rpc.messages.CreateAclsRequest;
 import com.alibaba.fluss.rpc.messages.CreateDatabaseRequest;
 import com.alibaba.fluss.rpc.messages.CreateTableRequest;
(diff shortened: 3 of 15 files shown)


fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/catalog/FlinkCatalog.java
```diff
@@ -278,9 +279,34 @@ public void dropDatabase(String databaseName, boolean ignoreIfNotExists, boolean
...
+
+            // Compare and generate DatabaseChanges
+            List<DatabaseChange> databaseChanges =
+                    generateDatabaseChanges(currentDescriptor, newDescriptor);
+
+            admin.alterDatabase(databaseName, databaseChanges, ignoreIfNotExists).get();
+        } catch (Exception e) {
+            Throwable t = ExceptionUtils.stripExecutionException(e);
+            if (CatalogExceptionUtils.isDatabaseNotExist(t)) {
+                if (!ignoreIfNotExists) {
```
loserwang1024, line 322: `admin.alterDatabase(databaseName, databaseChanges, ignoreIfNotExists)` already handle ignoreIfNotExists. Thus, just throw exception here

fluss-rpc/src/main/proto/FlussApi.proto
```diff
@@ -64,6 +64,16 @@ message CreateDatabaseRequest {
 message CreateDatabaseResponse {
 }
 
+// alter database request and response
+message AlterDatabaseRequest {
```
loserwang1024, line 68: where is comment? I think comment should be seperated from  config_changes, @wuchong , WDYT?

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/catalog/FlinkCatalog.java
LiebingYu: `admin.alterDatabase` throws `org.apache.fluss.exception.DatabaseNotExistException` while `FlinkCatalog.alterDatabase` throws `org.apache.flink.table.catalog.exceptions.DatabaseNotExistException`. We need to wrap it here like `dropDatabase` does.

fluss-rpc/src/main/proto/FlussApi.proto
LiebingYu: I think we can treat comment as a specific config. Otherwise we should extract comment from `List<DatabaseChange> databaseChanges` which is a bit of a hassle.

LiebingYu:
CC @loserwang1024. Could you have a look again? I've resolve the comment issue.

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/catalog/FlinkCatalog.java
```diff
@@ -278,9 +280,59 @@ public void dropDatabase(String databaseName, boolean ignoreIfNotExists, boolean
...
+            // Check custom properties changes
+            Map<String, String> oldProps = currentDescriptor.getCustomProperties();
+            Map<String, String> newProps = catalogDatabase.getProperties();
+
+            newProps.forEach(
+                    (k, v) -> {
+                        // When execute `ALTER DATABASE db SET ('comment' = 'new comment');`,
+                        // the comment will be set in the properties of CatalogDatabase.
+                        if (k.equals(COMMENT_PROP)) {
+                            databaseChanges.add(DatabaseChange.updateComment(v));
```
wuchong: This branch currently interprets any property named `comment` as a request to update the database comment. While `FlinkConversions.toFlussDatabase()` preserves `CatalogDatabase.getProperties()` verbatim, this approach feels somewhat fragile. Specifically, it prevents the correct alteration of databases that legitimately utilize a custom property named `comment`.

Given that Flink SQL lacks native syntax for modifying database comments, integrating this specific operation into Flink SQL seems unnecessary for such an infrequent task. Instead, we have already exposed database comment modification via the `Admin` API. This allows tools like the web console and the future Fluss CLI to support this functionality directly without compromising the integrity of custom properties.

Therefore, I suggest to remove the comment updating support until Flink SQL officially support `MODIFY COMMENT` syntax.

fluss-client/src/test/java/org/apache/fluss/client/admin/FlussAdminITCase.java
```diff
@@ -178,6 +179,75 @@ void testGetDatabaseInfo() throws Exception {
...
+        // Verify initial state
+        assertThat(existingDescriptor.getComment().get()).isEqualTo("original comment");
+        assertThat(existingDescriptor.getCustomProperties()).containsEntry("key1", "value1");
+        assertThat(existingDescriptor.getCustomProperties()).containsEntry("key2", "value2");
+
+        // Alter database: add and modify custom properties
+        List<DatabaseChange> databaseChanges = new ArrayList<>();
+        databaseChanges.add(DatabaseChange.set("key3", "value3"));
+        databaseChanges.add(DatabaseChange.set("key1", "updated_value1"));
+        databaseChanges.add(DatabaseChange.updateComment("updated comment"));
```
wuchong, line 208: could you add a test to reset comment?

fluss-server/src/main/java/org/apache/fluss/server/coordinator/MetadataManager.java
```diff
@@ -136,8 +137,85 @@ public void createDatabase(
...
+                throw (DatabaseNotExistException) e;
+            } else if (e instanceof RuntimeException) {
+                throw (RuntimeException) e;
+            } else {
+                throw new FlussRuntimeException("Failed to alter database: " + databaseName, e);
+            }
+        }
+    }
+
+    private DatabaseDescriptor getUpdatedDatabaseDescriptor(
```
wuchong, line 185: nit: add `@Nullable` annotation to the return value.

wuchong (approved):
LGTM

PR #1177, opened 2025-06-23
[doc] Add tips for building fluss with Java 8

leonardBang:
### Purpose

 
Add tips for building fluss with Java 8


luoyuxia (approved):
+1 for that. I did meet the problem and spent some time to find solution.

PR #1187, opened 2025-06-24
[server] Fix zk partition residual when using dynamic partition

### LINKED ISSUE #1185: Dynamic partition create redundant partition assignment in Zookeeper
### Search before asking

### Fluss version

main (development)

### Please describe the bug 🐞

Reproduce steps:
1. Create a partitioned table
2. Insert into table with enough data and trigger dynamic partition using Flink.
3. In my case, these data are partitioned into 24 partitions.
4. In ZK path : /metadata/databases/[databaseName]/tables/[tableName]/partitions   **24 partitions** created.
However, in ZK path /tabletservers/partitions/[partitionId]  **32 partitions** created.
5. Drop table in Flink. 24 partitions are successfully deleted but 8 partitions are residual in ZK.

It seems a thread unsafe issue.

### Solution

Although we found it in dynamic partition scene, it is not only concern with dynamic partition. It happens when more than one client try to create partition with the same table id and the same partition name at the same time (which usually occurs in dynamic partition).

`com.alibaba.fluss.server.coordinator.MetadataManager#createPartition
`
This method is not thread safe, even though it checks if partition exists before registers to zk.
Thread 1: 
t1: check zk for partition name 「x」  -> t2: register to zk
Thread 2: 
t3: check zk for partition name 「x」-> t4: register to zk
t1 and t3 can happen in the same time.
t2 and t4 can happen in the same time.

```
try {
	long partitionId = zookeeperClient.getPartitionIdAndIncrement();
	// register partition assignments to zk first
	zookeeperClient.registerPartitionAssignment(
	        partitionId, partitionAssignment);
	// then register the partition metadata to zk
	zookeeperClient.registerPartition(
	        tablePath, tableId, partitionName, partitionId);
	LOG.info(
	        "Register partition {} to zookeeper for table [{}].",
	        partitionName,
	        tablePath);
}
```

Thus "registerPartitionAssignment" call twice because partition id is different 
but "zookeeperClient.registerPartition" call once because they share the zk node.
That is why 
In zk path : /metadata/databases/[databaseName]/tables/[tableName]/partitions   **24 partitions** created.
However, in zk path /tabletservers/partitions/[partitionId]  **32 partitions** created.

zcoo:
### Purpose

Linked issue: close #1185

### DIFF
--- a/fluss-server/src/main/java/com/alibaba/fluss/server/coordinator/MetadataManager.java
+++ b/fluss-server/src/main/java/com/alibaba/fluss/server/coordinator/MetadataManager.java
@@ -438,10 +438,9 @@ public class MetadataManager {
 
         try {
             long partitionId = zookeeperClient.getPartitionIdAndIncrement();
-            // register partition assignments to zk first
-            zookeeperClient.registerPartitionAssignment(partitionId, partitionAssignment);
-            // then register the partition metadata to zk
-            zookeeperClient.registerPartition(tablePath, tableId, partitionName, partitionId);
+            // register partition assignments and partition metadata to zk in transaction
+            zookeeperClient.registerPartitionAssignmentAndMetadata(
+                    partitionId, partitionAssignment, tablePath, tableId, partitionName);
             LOG.info(
                     "Register partition {} to zookeeper for table [{}].", partitionName, tablePath);
         } catch (KeeperException.NodeExistsException nodeExistsException) {
--- a/fluss-server/src/main/java/com/alibaba/fluss/server/coordinator/MetadataManager.java
+++ b/fluss-server/src/main/java/com/alibaba/fluss/server/coordinator/MetadataManager.java
@@ -440,7 +440,7 @@ public class MetadataManager {
             long partitionId = zookeeperClient.getPartitionIdAndIncrement();
             // register partition assignments and partition metadata to zk in transaction
             zookeeperClient.registerPartitionAssignmentAndMetadata(
-                    partitionId, partitionAssignment, tablePath, tableId, partitionName);
+                    partitionId, partitionName, partitionAssignment, tablePath, tableId);
             LOG.info(
                     "Register partition {} to zookeeper for table [{}].", partitionName, tablePath);
         } catch (KeeperException.NodeExistsException nodeExistsException) {
--- a/fluss-server/src/main/java/com/alibaba/fluss/server/zk/ZooKeeperClient.java
+++ b/fluss-server/src/main/java/com/alibaba/fluss/server/zk/ZooKeeperClient.java
@@ -190,16 +190,6 @@ public class ZooKeeperClient implements AutoCloseable {
         LOG.info("Registered table assignment {} for table id {}.", tableAssignment, tableId);
     }
 
-    /** Register partition assignment to ZK. */
-    public void registerPartitionAssignment(
-            long partitionId, PartitionAssignment partitionAssignment) throws Exception {
-        String path = PartitionIdZNode.path(partitionId);
-        zkClient.create()
-                .creatingParentsIfNeeded()
-                .withMode(CreateMode.PERSISTENT)
-                .forPath(path, PartitionIdZNode.encode(partitionAssignment));
-    }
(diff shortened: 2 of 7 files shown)


fluss-server/src/main/java/com/alibaba/fluss/server/coordinator/MetadataManager.java
```diff
@@ -385,78 +393,97 @@ public void createPartition(
...
+                    } catch (Exception e) {
+                        throw new FlussRuntimeException(
+                                String.format(
+                                        "Register partition to zookeeper failed to create partition %s for table [%s]",
+                                        partitionName, tablePath),
+                                e);
+                    }
+                });
+        if (!lock.isLocked() && lock.getQueueLength() == 0) {
+            createPartitionLockMap.remove(physicalTablePath, lock);
```
luoyuxia: If any exception happens in above `inLock(xxx` cde fragement, this lock won't be removed..

luoyuxia: Considering the following case:
- Thread A come to here, is going to remove the lock1
- Thread B acquired the lock1 in https://github.com/alibaba/fluss/pull/1187/files#diff-4623210f8554fe6ea208a8dfc413f997d6d10b8c64cfc3aa530f3620ddbd09ccR401
- Thread A remove the lock1 from the map
- Thread C find no lock for the partition, try to create lock2
There still race in Thread B & Thread C.

TBH, I'm not still thinking out a good way to solve it. Maybe a guava cache?

fluss-server/src/main/java/com/alibaba/fluss/server/coordinator/MetadataManager.java
```diff
@@ -385,78 +393,97 @@ public void createPartition(
...
+                                String.format(
+                                        "Failed to check total bucket count for table %s",
+                                        tablePath),
+                                e);
+                    }
+
+                    try {
+                        long partitionId = zookeeperClient.getPartitionIdAndIncrement();
+                        // register partition assignments to zk first
+                        zookeeperClient.registerPartitionAssignment(
```
luoyuxia: Can we wrap `registerPartitionAssignment` and `registerPartition` into a zk transation? We let zk to do the concurrent controll so that we won't need the lock. We will have clean code. 
The only bad is that we may fire mutiple zk transation for same partition. But it shouldn't happen frequently, so I think it's fine since the current implemenation also will fire mutiple zk operations for same partition.

luoyuxia (commented):
@zcoo Thanks for the pr. Left some comments. PTAL

fluss-server/src/main/java/com/alibaba/fluss/server/coordinator/MetadataManager.java
zcoo: very valuable suggestion. I have changed the solution.

zcoo:
@luoyuxia Thanks! I refactor my code to solve problem by using zk transaction instead of Lock Map! PLAT

fluss-server/src/main/java/com/alibaba/fluss/server/zk/ZooKeeperClient.java
```diff
@@ -519,6 +519,62 @@ public void deletePartition(TablePath tablePath, String partitionName) throws Ex
         zkClient.delete().forPath(path);
     }
 
+    /** Register partition assignment and metadata in transaction. */
+    public void registerPartitionAssignmentAndMetadata(
```
luoyuxia, line 502: with this method, we can remove methods `registerPartition` & `registerPartitionAssignment`

luoyuxia (commented):
@zcoo Thanks for update. Only one comment. Otherwise, LGTM!

fluss-server/src/main/java/com/alibaba/fluss/server/zk/ZooKeeperClient.java
zcoo: yes!

PR #1190, opened 2025-06-25
[doc]The DDL and DQL column names in the context do not match

ZmmBigdata:
### Purpose

Linked issue: close #xxx 

**What is the purpose of the change**:
Fix the issue of mismatched DDL and DQL column names in the context, ensuring that the query statement is consistent with the table structure definition.

### Brief change log

**Please describe the changes**:
https://alibaba.github.io/fluss-docs/docs/table-design/table-types/log-table/
The DDL and DQL column names in the context do not match:
1. Added a 'timestamp' column in the DDL definition to ensure that the table structure includes a timestamp field.
2. Modified the 'SELECT' statement to explicitly specify the column names for the query:
-Original statement: ` SELECT * FROM log_table WHERE timestamp>'2023-01-01'; `
-New statement: ` SELECT order_i, item_id FROM log_table WHERE timestamp>'2023-01-01'; `
3. Optimized the description of query columns, clearly stating that only the 'order_id', 'item_id', and 'timestamp' columns are accessed, and other columns (such as' address' and 'amount') will be automatically pruned by the database and will not participate in storage and reading.

### API and Format

**Does this change affect API or storage format**:
- [ ] 是  
- [x] 否  
本次修改仅调整了 SQL 查询语句和表结构定义，不影响对外 API 或数据存储格式。

### Documentation

**Does this change introduce a new feature**:
- [ ] 是  
- [x] 否  
本次修改是对现有功能的修复和优化，无需新增文档。

### DIFF
--- a/website/docs/table-design/table-types/log-table.md
+++ b/website/docs/table-design/table-types/log-table.md
@@ -34,7 +34,8 @@ CREATE TABLE log_table (
   order_id BIGINT,
   item_id BIGINT,
   amount INT,
-  address STRING
+  address STRING,
+  `timestamp` DATE
 )
 WITH ('bucket.num' = '3');
 ```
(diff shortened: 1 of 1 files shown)


website/docs/table-design/table-types/log-table.md
```diff
@@ -75,10 +76,10 @@ During query execution, query engines like Flink analyzes the query to identify
 For example the following streaming query:
 
 ```sql
-SELECT id, name FROM log_table WHERE timestamp > '2023-01-01';
+SELECT order_id, item_id FROM log_table WHERE timestamp > '2023-01-01';
```
wuchong: I think the filter condition also doesn't work, because the `'2023-01-01'` is not a timestamp type. Could you also replace this with other meaningful condition?

ZmmBigdata: > I think the filter condition also doesn't work, because the `'2023-01-01'` is not a timestamp type. Could you also replace this with other meaningful condition?

Wait a moment, look look

ZmmBigdata: The timestamp field name has been changed to dt

PR #1192, opened 2025-06-25
[connector] pass Fluss schema to lake writer

cwang9208:
### Purpose

Use the schema in Fluss as the single source of truth for the lake writers to avoid any inconsistency.

Moreover, we sometimes need the Fluss schema to pick the correct field writers when creating lake writer. For example, currently the LocalZonedTimestampType and TimestampType type in Fluss map to the same Arrow type: 
https://github.com/alibaba/fluss/blob/main/fluss-common/src/main/java/com/alibaba/fluss/utils/ArrowUtils.java#L476
https://github.com/alibaba/fluss/blob/main/fluss-common/src/main/java/com/alibaba/fluss/utils/ArrowUtils.java#L489

When we tier the table to a data lake with Arrow schema, we are not able to choose the appropriate type of filed writers.

### Brief change log

Add the schema in Fluss to WriterInitContext.

### Tests

n/a

### API and Format

n/a

### Documentation

n/a

### DIFF
--- a/fluss-common/src/main/java/com/alibaba/fluss/lake/writer/WriterInitContext.java
+++ b/fluss-common/src/main/java/com/alibaba/fluss/lake/writer/WriterInitContext.java
@@ -55,5 +55,10 @@ public interface WriterInitContext {
     @Nullable
     String partition();
 
+    /**
+     * Returns the table schema.
+     *
+     * @return the table schema
+     */
     Schema schema();
 }
--- a/fluss-common/src/main/java/com/alibaba/fluss/lake/writer/WriterInitContext.java
+++ b/fluss-common/src/main/java/com/alibaba/fluss/lake/writer/WriterInitContext.java
@@ -18,6 +18,7 @@
 package com.alibaba.fluss.lake.writer;
 
 import com.alibaba.fluss.annotation.PublicEvolving;
+import com.alibaba.fluss.metadata.Schema;
 import com.alibaba.fluss.metadata.TableBucket;
 import com.alibaba.fluss.metadata.TablePath;
 
@@ -53,4 +54,6 @@ public interface WriterInitContext {
      */
     @Nullable
     String partition();
+
+    Schema schema();
 }
--- a/fluss-lake/fluss-lake-paimon/src/test/java/com/alibaba/fluss/lake/paimon/tiering/PaimonTieringTest.java
+++ b/fluss-lake/fluss-lake-paimon/src/test/java/com/alibaba/fluss/lake/paimon/tiering/PaimonTieringTest.java
@@ -658,6 +658,11 @@ class PaimonTieringTest {
                     public String partition() {
                         return partition;
                     }
+
+                    @Override
+                    public com.alibaba.fluss.metadata.Schema schema() {
+                        return null;
+                    }
                 });
     }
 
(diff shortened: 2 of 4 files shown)


fluss-common/src/main/java/com/alibaba/fluss/lake/writer/WriterInitContext.java
```diff
@@ -53,4 +54,6 @@ public interface WriterInitContext {
      */
     @Nullable
     String partition();
+
+    Schema schema();
```
luoyuxia, line 63: nit:
add java doc for this method

fluss-lake/fluss-lake-paimon/src/test/java/com/alibaba/fluss/lake/paimon/tiering/PaimonTieringTest.java
```diff
@@ -658,6 +658,11 @@ public TableBucket tableBucket() {
                     public String partition() {
                         return partition;
                     }
+
+                    @Override
+                    public com.alibaba.fluss.metadata.Schema schema() {
+                        return null;
```
luoyuxia: nit: throw unsupportException.

fluss-common/src/main/java/com/alibaba/fluss/lake/writer/WriterInitContext.java
leonardBang: Do we need to consider compatibility for example use   `default  Optional<Schema> schema() {return Optional.empty()}`

luoyuxia: I prefer not to keep code clean since currently it still for inner use in some degree.

fluss-lake/fluss-lake-paimon/src/test/java/com/alibaba/fluss/lake/paimon/tiering/PaimonTieringTest.java
xx789633: Thanks for suggestion. Fixed.

fluss-common/src/main/java/com/alibaba/fluss/lake/writer/WriterInitContext.java
xx789633: Sorry I forgot it. Fixed.

xx789633: I don't think there will be any compatibility issues. The tiering reader will automatically fill the WriterInitContext struct with the Fluss schema and it depends on the connector to decide whether to use it or not. For now, Paimon doesn't even touch this field.

leonardBang: The compatibility issue actually exists, `WriterInitContext` is a public API with annotation `WriterInitContext`, add a method in this API is an compatibility-breaking change, imaging this case, user implement their own `xxLakeWriter` base on Fluss 0.7 API. And once they bump fluss version from 0.7 to 0.8, they need to adjust their  `xxLakeWriter` implementation and build `fluss-lake-xx.jar` for 0.8 instead of using existing  `fluss-lake-xx.jar`  of 0.7.
But as there should not have this user case at this moment and 0.8 will be a apache version, I agree to keep code clean and introduce this method here.

xx789633: Noted. Thanks for the clarification! @leonardBang

PR #1198, opened 2025-06-25
[flink] support partial updates via column names

### LINKED ISSUE #925: Support Column Names for Sink Pushdowns
### Search before asking

### Description

Currently the Fluss Sink for the datastream API supports projection pushdown, but columns as specified via indexes.
We want to update the sink, so it can use column names instead of indexes since its more user-friendly 

### Willingness to contribute

MehulBatra:
### Purpose

Linked issue: close #925

Support column names in Fluss DS Sink for partial-updates. 

### Brief change log

- Enhanced `FlussSinkBuilder` to support column names for partial updates via `setPartialUpdateColumns(String... columnNames)` method
- Added column name to index resolution logic in `FlussSinkBuilder.build()` method
- Added input validation for column names (null, empty, whitespace checks)
- Added defensive copying to prevent external array modifications
-  **Added private method `resolveColumnNamesToIndexes()` to convert column names to their corresponding schema indexes**
- **Added private method `getTableInfo()` to retrieve table metadata from Fluss cluster**
- Maintained backward compatibility with the existing index-based API

### Tests

**Unit Tests (FlussSinkBuilderTest):**
- `testSetPartialUpdateColumnsVarargs()` - Tests basic varargs functionality
- `testPartialUpdateColumnsDefensiveCopy()` - Tests defensive copying behavior
- `testPartialUpdateColumnsValidation()` - Tests input validation (null, empty, invalid column names)

**Integration Tests (FlussSinkITCase):**
- `testPartialUpdateWithColumnNames()` - End-to-end test with multiple inserts and partial updates using column names

### API and Format

**API Changes:**
- Added new public method: `FlussSinkBuilder.setPartialUpdateColumns(String... columnNames)`
- Added private helper methods for column resolution and table metadata retrieval
- No breaking changes - existing index-based API remains unchanged

### Documentation

Will make changes for the documentation in a Seperate PR

### DIFF
--- a/fluss-flink/fluss-flink-common/src/test/java/com/alibaba/fluss/flink/sink/FlussSinkITCase.java
+++ b/fluss-flink/fluss-flink-common/src/test/java/com/alibaba/fluss/flink/sink/FlussSinkITCase.java
@@ -375,7 +375,7 @@ public class FlussSinkITCase extends FlinkTestBase {
             row.setField(0, value.orderId);
             row.setField(1, value.itemId);
             row.setField(2, value.amount);
-            row.setField(3, BinaryString.fromString(value.address));
+            row.setField(3, value.address != null ? BinaryString.fromString(value.address) : null);
 
             RowKind rowKind = value.rowKind;
             switch (rowKind) {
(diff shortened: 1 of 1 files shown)


MehulBatra:
@wuchong @polyzos When you have some time, could you please take a look at this PR? Thank you!

polyzos:
@MehulBatra thanks for this.. I took a quick look and we might need to investigate further.
I added a test (i pushed it into the branch for visibility that uses two writers, but the partial-update fails (see the attached image) .. When I find some more time, I will try and take a closer look.

Moreover I think we should also add some tests with RowData.

Screenshot 2025-07-23 at 8 15 37 PM

MehulBatra:
> @MehulBatra thanks for this.. I took a quick look and we might need to investigate further. I added a test (i pushed it into the branch for visibility that uses two writers, but the partial-update fails (see the attached image) .. When I find some more time, I will try and take a closer look.
> 
> Moreover I think we should also add some tests with RowData.
> 
> Screenshot 2025-07-23 at 8 15 37 PM

Thanks for reporting this @polyzos will check and see what can be done.

PR #1216, opened 2025-06-26
[hotfix] Optimize the error msg when throwing LogOffsetOutOfRangeException

swuferhong:
### Purpose

Currently, When a `LogOffsetOutOfRangeException` is thrown in `LogFetchCollector`, the current error message only provides the default exception defined in `Errors`: "The log offset is out of range", which is insufficient for troubleshooting.

### DIFF
--- a/fluss-client/src/main/java/com/alibaba/fluss/client/table/scanner/log/LogFetchCollector.java
+++ b/fluss-client/src/main/java/com/alibaba/fluss/client/table/scanner/log/LogFetchCollector.java
@@ -263,8 +263,9 @@ public class LogFetchCollector {
             metadataUpdater.checkAndUpdateMetadata(tablePath, tb);
         } else if (error == Errors.LOG_OFFSET_OUT_OF_RANGE_EXCEPTION) {
             throw new FetchException(
-                    String.format("The fetching offset %s is out of range", fetchOffset),
-                    error.exception());
+                    String.format(
+                            "The fetching offset %s is out of range: %s",
+                            fetchOffset, errorMessage));
         } else if (error == Errors.AUTHORIZATION_EXCEPTION) {
             throw new AuthorizationException(errorMessage);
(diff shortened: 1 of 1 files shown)


fluss-client/src/main/java/com/alibaba/fluss/client/table/scanner/log/LogFetchCollector.java
```diff
@@ -263,8 +263,9 @@ private void handleInitializeErrors(
             metadataUpdater.checkAndUpdateMetadata(tablePath, tb);
         } else if (error == Errors.LOG_OFFSET_OUT_OF_RANGE_EXCEPTION) {
             throw new FetchException(
-                    String.format("The fetching offset %s is out of range", fetchOffset),
-                    error.exception());
+                    String.format(
+                            "The fetching offset %s is out of range: %s",
+                            fetchOffset, errorMessage));
```
wuchong: Sorry, I couldn't understand this change. It seems the new version have less error information (strip out the exception stack). The `error.exception()` should contain information in `errorMessage`.

swuferhong: For common error, The `error.exception()` indeed contain information in `errorMessage`. But in method `LogFetchCollector#initialize()`, it refactored `ApiError` into `Errors` and `ApiMessage.messageWithFallback()` to pass to the `handleInitializeErrors` method. As a result, the `Errors` class no longer contains detailed error messages, debug output is as follow:

image

wuchong: OK. Then I suggest to use `error.exception(errorMessage)` to carry the API exception in the error stack.

swuferhong:
@wuchong  comments addressed

loserwang1024 (approved):

PR #1222, opened 2025-06-27
[server] Introduce plugin.classloader.parent-first-patterns.additiona…

### LINKED ISSUE #1221: Introduce plugin.classloader.parent-first-patterns.additional configuration
### Search before asking

### Motivation

Currently, we provide hadoop-shaded class in lib plugin/paimon/. For most non-hadoop users, it's fine. But for some hadoop users, they want to use their own hadoop libs in theire hadoop env. I propose to a configuration `plugin.classloader.parent-first-patterns.additional ` to enable use to set `org.apache.hadoop.` prefix classs to use app classloader to load hadoop class.

### Solution

_No response_

### Anything else?

_No response_

### Willingness to contribute

luoyuxia:
…l configuration

### Purpose

Linked issue: close #1221 

### Brief change log

Introduce `plugin.classloader.parent-first-patterns.additional` just like Flink

### Tests

PluginConfigTest#testFromConfig

### DIFF
--- a/fluss-common/src/main/java/com/alibaba/fluss/plugin/PluginConfig.java
+++ b/fluss-common/src/main/java/com/alibaba/fluss/plugin/PluginConfig.java
@@ -17,15 +17,19 @@
 
 package com.alibaba.fluss.plugin;
 
-import com.alibaba.fluss.config.ConfigOptions;
 import com.alibaba.fluss.config.Configuration;
+import com.alibaba.fluss.shaded.guava32.com.google.common.collect.Iterables;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.io.File;
 import java.nio.file.Path;
+import java.util.List;
 import java.util.Optional;
+
+import static com.alibaba.fluss.config.ConfigOptions.PLUGIN_ALWAYS_PARENT_FIRST_LOADER_PATTERNS;
+import static com.alibaba.fluss.config.ConfigOptions.PLUGIN_ALWAYS_PARENT_FIRST_LOADER_PATTERNS_ADDITIONAL;
 
 /** Stores the configuration for plugins mechanism. */
 @SuppressWarnings("OptionalUsedAsFieldOrParameterType")
@@ -62,9 +66,13 @@ public class PluginConfig {
     public static PluginConfig fromConfiguration(Configuration configuration) {
         return new PluginConfig(
                 getPluginsDir().map(File::toPath),
-                configuration
-                        .get(ConfigOptions.PLUGIN_ALWAYS_PARENT_FIRST_LOADER_PATTERNS)
-                        .toArray(new String[0]));
+                getPluginParentFirstLoaderPatterns(configuration));
+    }
+
+    private static String[] getPluginParentFirstLoaderPatterns(Configuration config) {
+        List<String> base = config.get(PLUGIN_ALWAYS_PARENT_FIRST_LOADER_PATTERNS);
+        List<String> append = config.get(PLUGIN_ALWAYS_PARENT_FIRST_LOADER_PATTERNS_ADDITIONAL);
+        return Iterables.toArray(Iterables.concat(base, append), String.class);
     }
 
     public static Optional<File> getPluginsDir() {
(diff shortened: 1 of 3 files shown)


fluss-common/src/main/java/com/alibaba/fluss/plugin/PluginConfig.java
```diff
@@ -62,9 +66,13 @@ public String[] getAlwaysParentFirstPatterns() {
...
         return new PluginConfig(
                 getPluginsDir().map(File::toPath),
-                configuration
-                        .get(ConfigOptions.PLUGIN_ALWAYS_PARENT_FIRST_LOADER_PATTERNS)
-                        .toArray(new String[0]));
+                getPluginParentFirstLoaderPatterns(configuration));
+    }
+
+    private static String[] getPluginParentFirstLoaderPatterns(Configuration config) {
+        List<String> base = config.get(PLUGIN_ALWAYS_PARENT_FIRST_LOADER_PATTERNS);
```
loserwang1024:
```java
private static final class PluginClassLoader extends ComponentClassLoader {

        PluginClassLoader(
                URL[] pluginResourceURLs,
                ClassLoader flussClassLoader,
                String[] allowedFlussPackages) {
            super(
                    pluginResourceURLs,
                    flussClassLoader,
                    allowedFlussPackages,
                    new String[0],
                    Collections.emptyMap());
        }
    }
```
In currently fuss PluginClassLoader, componentFirstPackages is alway empty. That means only owner-First and  component-only. If we add PLUGIN_ALWAYS_PARENT_FIRST_LOADER_PATTERNS, this class will be loaded by component-only.  If user set PLUGIN_ALWAYS_PARENT_FIRST_LOADER_PATTERNS, then class is not provided by component, exception will be thrown.  Why not provided a way to set as component-first?

luoyuxia: I'm not sure I understand you correctly. But try to answer your question. 

If we add `PLUGIN_ALWAYS_PARENT_FIRST_LOADER_PATTERNS_ADDITIONAL`, I assume you mean `PLUGIN_ALWAYS_PARENT_FIRST_LOADER_PATTERNS_ADDITIONAL`, this class will be loaded by owner first.

If user set PLUGIN_ALWAYS_PARENT_FIRST_LOADER_PATTERNS, then class is not provided by component, exception will be thrown. I think exception is thrown.

We can provided a way to set as component-first, but that doesn't solve my problem.
I still can't find a case that we want to `component-first`.

wuchong:
The change looks good to me, I only have 2 questions:
1. Can this resolve the hadoop problem discussed in Slack: https://fluss-hq.slack.com/archives/C08QXHUH3M1/p1750331641353339?thread_ts=1750317156.642389&cid=C08QXHUH3M1 ? 
2. If this is used to allow users to set `org.apache.hadoop.` prefix classs to use app classloader to load hadoop class. Does this also works if we don't put `flink-shaded-hadoop-2-uber-2.8.3-10.0.jar` in the `plugins/paimon`? Why we put `flink-shaded-hadoop-2-uber-2.8.3-10.0.jar` under paimon plugin?

luoyuxia:
> The change looks good to me, I only have 2 questions:
> 
> 1. Can this resolve the hadoop problem discussed in Slack: https://fluss-hq.slack.com/archives/C08QXHUH3M1/p1750331641353339?thread_ts=1750317156.642389&cid=C08QXHUH3M1 ?
> 2. If this is used to allow users to set `org.apache.hadoop.` prefix classs to use app classloader to load hadoop class. Does this also works if we don't put `flink-shaded-hadoop-2-uber-2.8.3-10.0.jar` in the `plugins/paimon`? Why we put `flink-shaded-hadoop-2-uber-2.8.3-10.0.jar` under paimon plugin?

1: I'm still not sure why the hadoop(class not found) problem happens only in docker. But if hadoop env is already installed, it'll find the class via app classloader which should find the hadoop class.

2: Yes, it still works if  we don't put `flink-shaded-hadoop-2-uber-2.8.3-10.0.jar` in the `plugins/paimon`. The reason why we put `flink-shaded-hadoop-2-uber-2.8.3-10.0.jar` under paimon plugin is to simplify the use. No matter what paimon catalogs(hive, filesystem,jdbc) is used, the hadoop classes are required(althougth strange, but paimon does need them). Anyway, uses always need the hadoop  classes even thought they are using s3 filesystem catalog.  But there's  one case is that users may want to use their own hadoop lib in their own hadoop env, so introduce the pr to support that.

wuchong:
This looks good to me. @luoyuxia could you rebase the branch again. We can merge it once the CI is passed.

PR #1230, opened 2025-06-27
[examples] A module for fluss examples

drr00t:
Fluss newcomers onboarding :) 

### Purpose

Added a module to Fluss features usage examples, I did followed the Flink structure as proposal.

I am not found any issue related. If this make sense to you guys I can


fluss-examples/fluss-examples-flinkengine/pom.xml
```diff
@@ -0,0 +1,51 @@
...
+    <name>Fluss : Examples : FlinkEngine Usage :</name>
+    <packaging>pom</packaging>
+    <description>This an example of Fluss Flink Engine framework usage.</description>
+
+    <dependencies>
+        <!-- https://mvnrepository.com/artifact/com.alibaba.fluss/fluss-flink-1.20 -->
+        <dependency>
+            <groupId>com.alibaba.fluss</groupId>
+            <artifactId>fluss-flink-1.20</artifactId>
+            <version>0.8-SNAPSHOT</version>
```
wuchong: use `<version>${project.version}</version>`?

fluss-examples/fluss-examples-flinkengine/pom.xml
```diff
@@ -0,0 +1,51 @@
...
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+    <parent>
+        <groupId>com.alibaba.fluss</groupId>
+        <artifactId>fluss-examples</artifactId>
+        <version>0.8-SNAPSHOT</version>
+    </parent>
+
+    <artifactId>fluss-examples-flinkengine</artifactId>
```
wuchong: `<artifactId>fluss-examples-flink</artifactId>` for simplification?

fluss-examples/fluss-examples-flinkengine/src/main/java/com/alibaba/fluss/examples/Main.java
```diff
@@ -0,0 +1,53 @@
...
+
+import org.apache.flink.api.common.eventtime.WatermarkStrategy;
+import org.apache.flink.streaming.api.datastream.DataStreamSource;
+import org.apache.flink.streaming.api.environment.StreamExecutionEnvironment;
+
+/**
+ * Example of using FlussSource to read data from a Fluss database in a Flink application.
+ * https://alibaba.github.io/fluss-docs/docs/engine-flink/getting-started/
+ */
+public class Main {
```
wuchong: Rename to `DataStreamExample`? We should also add some instructions for how to submit the example jar to a Flink cluster. Maybe we can add a README under the `fluss-examples` directory.

fluss-examples/fluss-examples-javaclient/pom.xml
```diff
@@ -0,0 +1,43 @@
...
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+    <parent>
+        <groupId>com.alibaba.fluss</groupId>
+        <artifactId>fluss-examples</artifactId>
+        <version>0.8-SNAPSHOT</version>
+    </parent>
+
+    <artifactId>fluss-examples-javaclient</artifactId>
```
wuchong: `<artifactId>fluss-examples-java-client</artifactId>`?

fluss-examples/fluss-examples-javaclient/pom.xml
```diff
@@ -0,0 +1,43 @@
...
+    <name>Fluss : Examples : JavaClient Usage :</name>
+    <packaging>pom</packaging>
+    <description>This an example of Fluss Java client library apis usage.</description>
+
+    <dependencies>
+        <dependency>
+            <groupId>com.alibaba.fluss</groupId>
+            <artifactId>fluss-client</artifactId>
+            <version>0.8-SNAPSHOT</version>
```
wuchong: ditto

fluss-examples/fluss-examples-javaclient/src/main/java/com/alibaba/fluss/examples/MainFlussClientStreaming.java
```diff
@@ -0,0 +1,179 @@
...
+import java.time.LocalDateTime;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.concurrent.ExecutionException;
+import java.util.stream.Collectors;
+
+// TIP To <b>Run</b> code, press <shortcut actionId="Run"/> or
+// click the <icon src="AllIcons.Actions.Execute"/> icon in the gutter.
+/** Example Java Client usage for Fluss demonstrating streaming operations. */
+public class MainFlussClientStreaming {
```
wuchong: Rename to `JavaClientExample`?

fluss-examples/fluss-examples-flinkengine/src/main/java/com/alibaba/fluss/examples/Main.java
drr00t: I created a 3 docker compose files: **zookeeper**, **fluss** and **flink** to implemented this examples. I can add those + startup script to create the execution environment.

fluss-examples/fluss-examples-flinkengine/pom.xml
drr00t: Yes, would be nice to have this. But Fluss doesn't yet.... [fluss pom.xml file ](https://github.com/alibaba/fluss/blob/da6a2a473c70e1469123946045e99752a3e7375d/pom.xml#L32)

fluss-examples/fluss-examples-flinkengine/src/main/java/com/alibaba/fluss/examples/Main.java
drr00t: This I will improve documentation and include docker-compose for execution environment with script to generate some data, will be nice to have this.

fluss-examples/fluss-examples-flink/src/main/java/com/alibaba/fluss/examples/DataStreamExample.java
```diff
@@ -0,0 +1,53 @@
...
+ * https://alibaba.github.io/fluss-docs/docs/engine-flink/getting-started/
+ */
+public class DataStreamExample {
+
+    public static void main(String[] args) {
+        StreamExecutionEnvironment env = StreamExecutionEnvironment.getExecutionEnvironment();
+        // Create a FlussSource using the builder pattern
+        FlussSource<Order> flussSource =
+                FlussSource.<Order>builder()
+                        .setBootstrapServers("localhost:9123")
```
wuchong: we may also need to adapt the boostrap address to the dockerized environemnt?

fluss-examples/docker-compose.yml
```diff
@@ -0,0 +1,132 @@
...
+    environment:
+      FLUSS_PROPERTIES: |
+        zookeeper.address: zoo1:2181,zoo2:2181,zoo3:2181
+        bind.listeners: INTERNAL://coordinator-server:0, CLIENT://coordinator-server:9123
+        advertised.listeners: CLIENT://localhost:9123
+        internal.listener.name: INTERNAL
+    networks:
+      - fluss-net
+
+  tablet-server-0:
```
polyzos, line 62: I think it might be better to have 3-tablet servers instead of 3 zk server, to simulate a distributed environment, so its more realistic. WDYT? You can find one such file [here](https://github.com/ververica/ververica-fluss-examples/blob/main/fluss-java-client/docker-compose.yaml)

fluss-examples/fluss-examples-flink/pom.xml
```diff
@@ -0,0 +1,88 @@
...
+    <dependencies>
+        <!-- https://mvnrepository.com/artifact/com.alibaba.fluss/fluss-flink-1.20 -->
+        <dependency>
+            <groupId>com.alibaba.fluss</groupId>
+            <artifactId>fluss-flink-1.20</artifactId>
+            <version>${project.version}</version>
+        </dependency>
+
+        <dependency>
+            <groupId>org.apache.flink</groupId>
```
polyzos: for the datastream api, more depedencies should be needed.. so its better to add these as provided, and also add the `flink-version` under `properties`

fluss-examples/pom.xml
```diff
@@ -0,0 +1,45 @@
...
+    <modelVersion>4.0.0</modelVersion>
+    <parent>
+        <groupId>com.alibaba.fluss</groupId>
+        <artifactId>fluss</artifactId>
+        <version>0.8-SNAPSHOT</version>
+    </parent>
+
+    <artifactId>fluss-examples</artifactId>
+    <name>Fluss : Examples :</name>
+    <version>0.8-SNAPSHOT</version>
```
polyzos: add project version here

fluss-examples/pom.xml
```diff
@@ -0,0 +1,45 @@
...
+    <packaging>pom</packaging>
+
+    <modules>
+        <module>fluss-examples-java-client</module>
+        <module>fluss-examples-flink</module>
+    </modules>
+    <properties>
+        <project.version>0.8-SNAPSHOT</project.version>
+        <java.version>1.8</java.version>
```
polyzos: i think its better to go with java 11 here

fluss-examples/pom.xml
```diff
@@ -0,0 +1,45 @@
...
+    <name>Fluss : Examples :</name>
+    <version>0.8-SNAPSHOT</version>
+    <packaging>pom</packaging>
+
+    <modules>
+        <module>fluss-examples-java-client</module>
+        <module>fluss-examples-flink</module>
+    </modules>
+    <properties>
+        <project.version>0.8-SNAPSHOT</project.version>
```
polyzos: same.. use project version

polyzos:
@drr00t Thanks for your contribution.. Overall, it's a great idea, but lets see how to better structure this and have examples that can work standalone out of the box. 
Starting with faker as you suggested [here](https://github.com/alibaba/fluss/issues/1241) and then moving on creating some simple examples/use cases should be a good approach

fluss-examples/fluss-examples-flink/src/main/java/com/alibaba/fluss/examples/DataStreamExample.java
drr00t: Maybe with an application properties as I'm not using spring or nothing like maybe with can use form apache commons, but make sense to dockerize too.I'll take a look on this.

drr00t: @wuchong I made a simple change support external configuration via cli or env for future dockerization, but I'm working on a more complete example with data generation.

drr00t:
> @drr00t Thanks for your contribution.. Overall, it's a great idea, but lets see how to better structure this and have examples that can work standalone out of the box. Starting with faker as you suggested [here](https://github.com/alibaba/fluss/issues/1241) and then moving on creating some simple examples/use cases should be a good approach

I commited an example with datafaker as data generator

polyzos:
@drr00t Thanks for all the work.. I just took a look at this PR and because I see this grow I wanna share a few thoughts.
1. This module is about examples, but I feel like there is too much focus on data generators rather than examples
2. The focus should be on Fluss APIs, but there are too many classes and code spreading in the repo, including the use of webflux... I think keeping it focused and keeping it to the bare minimum in terms of the client should be our goal
3. Similar to Apache Flink examples (or other projects, like Pulsar - see [here](https://github.com/streamnative/examples/blob/master/clients/README.md)) the examples should implement something that covers something like a use case or a "pattern" for the user .. Similar to the docs it can be something simple like:
   - write/read using the Fluss client
   - how to perform lookups
   - how to scale the reader in terms of the number of buckets i.e start a thread per bucket available
   - column pruning examples (later with predicate pushdown) etc. 
 
The current approach although its great, feels more like something to be explained into a blog post that tries to build something, rather than examples, that can be used standalone and for users to just get up and running (mostly because they will have to navigate multiple classes, code etc.) .. Does it make sense? WDYT?

drr00t:
> @drr00t Thanks for all the work.. I just took a look at this PR and because I see this grow I wanna share a few thoughts.
> 
>     1. This module is about examples, but I feel like there is too much focus on data generators rather than examples
> 
>     2. The focus should be on Fluss APIs, but there are too many classes and code spreading in the repo, including the use of webflux... I think keeping it focused and keeping it to the bare minimum in terms of the client should be our goal
> 
>     3. Similar to Apache Flink examples (or other projects, like Pulsar - see [here](https://github.com/streamnative/examples/blob/master/clients/README.md)) the examples should implement something that covers something like a use case or a "pattern" for the user .. Similar to the docs it can be something simple like:
>        
>        * write/read using the Fluss client
>        * how to perform lookups
>        * how to scale the reader in terms of the number of buckets i.e start a thread per bucket available
>        * column pruning examples (later with predicate pushdown) etc.
> 
> 
> The current approach although its great, feels more like something to be explained into a blog post that tries to build something, rather than examples, that can be used standalone and for users to just get up and running (mostly because they will have to navigate multiple classes, code etc.) .. Does it make sense? WDYT?

@polyzos for the feedback it's make sense and I ask you to take look exactly to confirm this "impression" too complicated just to an example. This ecommerce module will going to need a blog post explaining what it does.

I will change approach, elaborate a simpler and focused examples to present the concepts referenced into our documentation.

polyzos:
and we can publish it easily on the Fluss website, because its nice work

drr00t:
hi @polyzos @wuchong,

I made a complete rewrite of PR and start to write few very basic examples exploring Fluss concepts, like: writing, scan, prefixlookup, append data to log table, upsert into primarykey table. All example can be executed via containers, Fluss server can be passed via, ENV,  CLI or default to localhost.

I did remove the previous realtime ecommerce analytics with data generation using datafaker and plan to write a blog post in another PR.

PR #1245, opened 2025-06-30
[file system] Configurable token delegation process to enable other authentication methods in S3 credential provider chain

### LINKED ISSUE #1246: [file system] Add configuration option to enable/disable token delegation process for S3 file system (allows support for MinIO and other S3-compatible object stores)
### Search before asking

### Motivation

Currently, the Fluss token delegation process cannot be disabled. This limits file systems to use Fluss-internal credential providers. By adding the configuration option, also other credential providers in the provider chain can be used.

### Solution

- Add configuration option to disable provider chain per file system (`fs.s3.enable-token-delegation`)
- Decide based on `fs.s3.enable-token-delegation` whether to obtain a token via STS or return a dummy token. The dummy token-approach is similar to the implementation of the HDFS file system and stops the token renewal process in `com.alibaba.fluss.client.token.DefaultSecurityTokenManager.java` after the initial iteration.
- Add `com.alibaba.fluss.fs.s3.token.DynamicTemporaryAWSCredentialsProvider.java` to provider chain by default and let the library figure out which credential provider in the chain is valid automatically.

### Anything else?

_No response_

### Willingness to contribute

michaelkoepf:
### Purpose

Linked issue: close #1246

### Brief change log

- Allow to enable/disable token delegation process in Fluss S3 filesystem via config option
- Pass filesystem config options from Fluss Flink connector to client module and whitelist certain config options with the client prefix in the S3 file system
- Determine if file system is initialized by server or client based on user config to correctly initialize the file system
- Update documentation

### Tests

**Unit Tests**

Added 

- `fluss-filesystems/fluss-fs-s3/src/test/java/com/alibaba/fluss/fs/s3/S3FileSystemPluginTest.java`
- `fluss-filesystems/fluss-fs-s3/src/test/java/com/alibaba/fluss/fs/s3/token/S3DelegationTokenReceiverTest.java`

**Example usage with MinIO**

1. Checkout PR and build using `./mvnw clean package -DskipTests`.

2. `docker` directory: Build fluss image with name `fluss:local-snapshot`. Execute the `prepare.sh` for the quickstart image, copy the local `fluss-fs-s3` from this PR into the `lib` folder and build the image with name `fluss-quickstart-flink:local-snapshot`.

3. `docker-compose.yml`

```yaml
services:
  #begin Fluss cluster
  coordinator-server:
    image: fluss:local-snapshot
    command: coordinatorServer
    depends_on:
      - zookeeper
      - minio-init
    environment:
      - ROOT_LOG_LEVEL=DEBUG
      - |
        FLUSS_PROPERTIES=
        zookeeper.address: zookeeper:2181
        bind.listeners: FLUSS://coordinator-server:9123
        remote.data.dir: s3://fluss/remote-data
        fs.s3.enable-token-delegation: false
        s3.access-key: fluss
        s3.secret-key: 12345678
        s3.path-style-access: true
        s3.region: unknown_region
        s3.endpoint: http://minio:9000
        datalake.format: paimon
        datalake.paimon.metastore: filesystem
        datalake.paimon.warehouse: /tmp/paimon
    volumes:
      - shared-tmpfs:/tmp/paimon
  tablet-server:
    image: fluss:local-snapshot
    command: tabletServer
    depends_on:
      - coordinator-server
    environment:
      - ROOT_LOG_LEVEL=DEBUG
      - |
        FLUSS_PROPERTIES=
        zookeeper.address: zookeeper:2181
        bind.listeners: FLUSS://tablet-server:9123
        data.dir: /tmp/fluss/data
        remote.data.dir: s3://fluss/remote-data
        remote.log.task-interval-duration: 15s
        # reduce for testing purposes
        log.segment.file-size: 4k
        fs.s3.enable-token-delegation: false
        s3.access-key: fluss
        s3.secret-key: 12345678
        s3.path-style-access: true
        s3.region: unknown_region
        s3.endpoint: http://minio:9000
        kv.snapshot.interval: 15s
        datalake.format: paimon
        datalake.paimon.metastore: filesystem
        datalake.paimon.warehouse: /tmp/paimon
    volumes:
      - shared-tmpfs:/tmp/paimon
  zookeeper:
    restart: always
    image: zookeeper:3.9.2
  #end
  #begin Flink cluster
  jobmanager:
    image: fluss-quickstart-flink:local-snapshot
    depends_on:
      - minio-init
    ports:
      - "8083:8081"
    command: jobmanager
    environment:
      - ROOT_LOG_LEVEL=DEBUG
      - |
        FLINK_PROPERTIES=
        jobmanager.rpc.address: jobmanager
        fluss.client.fs.s3.access-key: fluss
        fluss.client.fs.s3.secret-key: 12345678
  taskmanager:
    image: fluss-quickstart-flink:local-snapshot
    depends_on:
      - jobmanager
    command: taskmanager
    environment:
      - ROOT_LOG_LEVEL=DEBUG
      - |
        FLINK_PROPERTIES=
        jobmanager.rpc.address: jobmanager
        taskmanager.numberOfTaskSlots: 10
        taskmanager.memory.process.size: 2048m
        taskmanager.memory.framework.off-heap.size: 256m
  #end
  #begin minio
  minio:
    image: minio/minio:latest
    command: server /data --console-address ":9001"
    ports:
      - "9000:9000"
      - "9001:9001"
    environment:
      - MINIO_ROOT_USER=fluss
      - MINIO_ROOT_PASSWORD=12345678
    volumes:
      - minio-data:/data
  minio-init:
    image: minio/minio:latest
    depends_on:
      - minio
    entrypoint: >
      /bin/sh -c "
      sleep 10;
      /usr/bin/mc alias set minio http://minio:9000 fluss 12345678;
      /usr/bin/mc mb minio/fluss;
      /usr/bin/mc mb minio/paimon;
      exit 0;
      "
  #end

volumes:
  minio-data:
    driver: local
    driver_opts:
      type: "tmpfs"
      device: "tmpfs"
  shared-tmpfs:
    driver: local
    driver_opts:
      type: "tmpfs"
      device: "tmpfs"
```

Start with `docker compose up -d`. 

**Notes:** 

- Since we set `fs.s3.enable-token-delegation: false`, the token delegation process is _deactivated_, and there will be no exception thrown during obtaining the session token via STS.

- There is no need to set `region`, `endpoint` or `path-style-access` on the client side, because the information will be distributed from the server to the client even when token delegation is deactivated.

3. `docker compose exec jobmanager ./sql-client`

4. Based on quickstart guide, create catalog and table, and insert data.

```
CREATE CATALOG fluss_catalog WITH (
    'type' = 'fluss',
    'bootstrap.servers' = 'coordinator-server:9123'
);
```

```
USE CATALOG fluss_catalog;
```

```
CREATE TABLE fluss_order (
    `order_key` BIGINT,
    `cust_key` INT NOT NULL,
    `total_price` DECIMAL(15, 2),
    `order_date` DATE,
    `order_priority` STRING,
    `clerk` STRING,
    `ptime` AS PROCTIME(),
    PRIMARY KEY (`order_key`) NOT ENFORCED
);
```

```
EXECUTE STATEMENT SET
BEGIN
    INSERT INTO fluss_order SELECT * FROM `default_catalog`.`default_database`.source_order;
END;
```

5. Go to `http://localhost:9001`. Log in with user `fluss` and password `12345678`. 

6. Wait until `fluss/remote-data` contains the `log` folder, i.e., logs are tiered so we can be sure that we read from MinIO.

7. `SELECT * FROM fluss_order;` should successfully read from MinIO.

### API and Format

n/a

### Documentation

`fluss-docs/docs/maintenance/filesystems/s3`

### DIFF
--- a/fluss-common/src/main/java/org/apache/fluss/config/ConfigOptions.java
+++ b/fluss-common/src/main/java/org/apache/fluss/config/ConfigOptions.java
@@ -1883,46 +1883,46 @@ public class ConfigOptions {
         ZSTD
     }
 
-  // ------------------------------------------------------------------------
-  //  ConfigOptions Registry and Utilities
-  // ------------------------------------------------------------------------
+    // ------------------------------------------------------------------------
+    //  ConfigOptions Registry and Utilities
+    // ------------------------------------------------------------------------
 
-  /**
-   * Holder class for lazy initialization of ConfigOptions registry. This ensures that the
-   * registry is initialized only when first accessed, and guarantees that all ConfigOption fields
-   * are already initialized (since static initialization happens in declaration order).
-   */
-  private static class ConfigOptionsHolder {
-    private static final Map<String, ConfigOption<?>> CONFIG_OPTIONS_BY_KEY;
+    /**
+     * Holder class for lazy initialization of ConfigOptions registry. This ensures that the
+     * registry is initialized only when first accessed, and guarantees that all ConfigOption fields
+     * are already initialized (since static initialization happens in declaration order).
+     */
+    private static class ConfigOptionsHolder {
+        private static final Map<String, ConfigOption<?>> CONFIG_OPTIONS_BY_KEY;
 
-    static {
-      Map<String, ConfigOption<?>> options = new HashMap<>();
-      Field[] fields = ConfigOptions.class.getFields();
-      for (Field field : fields) {
-        if (!ConfigOption.class.isAssignableFrom(field.getType())) {
-          continue;
+        static {
+            Map<String, ConfigOption<?>> options = new HashMap<>();
+            Field[] fields = ConfigOptions.class.getFields();
+            for (Field field : fields) {
+                if (!ConfigOption.class.isAssignableFrom(field.getType())) {
+                    continue;
+                }
+                try {
+                    ConfigOption<?> configOption = (ConfigOption<?>) field.get(null);
+                    options.put(configOption.key(), configOption);
+                } catch (IllegalAccessException e) {
+                    // Ignore fields that cannot be accessed
+                }
+            }
+            CONFIG_OPTIONS_BY_KEY = Collections.unmodifiableMap(options);
         }
-        try {
-          ConfigOption<?> configOption = (ConfigOption<?>) field.get(null);
-          options.put(configOption.key(), configOption);
-        } catch (IllegalAccessException e) {
-          // Ignore fields that cannot be accessed
-        }
-      }
-      CONFIG_OPTIONS_BY_KEY = Collections.unmodifiableMap(options);
     }
-  }
 
-  /**
-   * Gets the ConfigOption for a given key.
-   *
-   * @param key the configuration key
-   * @return the ConfigOption if found, null otherwise
-   */
-  @Internal
-  public static ConfigOption<?> getConfigOption(String key) {
-    return ConfigOptionsHolder.CONFIG_OPTIONS_BY_KEY.get(key);
-  }
+    /**
+     * Gets the ConfigOption for a given key.
+     *
+     * @param key the configuration key
+     * @return the ConfigOption if found, null otherwise
+     */
+    @Internal
+    public static ConfigOption<?> getConfigOption(String key) {
+        return ConfigOptionsHolder.CONFIG_OPTIONS_BY_KEY.get(key);
+    }
 
     // ------------------------------------------------------------------------
     //  ConfigOptions for file systems (Fluss-internal, not recognized by Hadoop FS libraries)
--- a/fluss-filesystems/fluss-fs-s3/src/main/java/org/apache/fluss/fs/s3/S3FileSystemPlugin.java
+++ b/fluss-filesystems/fluss-fs-s3/src/main/java/org/apache/fluss/fs/s3/S3FileSystemPlugin.java
@@ -77,9 +77,6 @@ public class S3FileSystemPlugin implements FileSystemPlugin {
                             "aws.credentials.provider",
                             "assumed.role.credentials.provider"));
 
-    private static final List<String> TOKEN_DELEGATION_DEFAULT_CREDENTIAL_PROVIDER =
-            Collections.singletonList("org.apache.hadoop.fs.s3a.SimpleAWSCredentialsProvider");
-
     @Override
     public String getScheme() {
         return "s3";
@@ -94,11 +91,12 @@ public class S3FileSystemPlugin implements FileSystemPlugin {
         final boolean useTokenDelegation;
 
         if (isClient) {
-            // Only used on the server side, default to false
+            // Only relevant on the server side, default to false
             useTokenDelegation = false;
             // We do not know if token delegation on the server will be activated or deactivated.
             // Hence, we just add the Fluss credential provider for token delegation
-            // to the provider chain.
+            // to the provider chain and the file system will figure out a valid provider
+            // at runtime.
             setCredentialProviders(
                     hadoopConfig,
                     Collections.singletonList(DynamicTemporaryAWSCredentialsProvider.NAME));
@@ -143,10 +141,6 @@ public class S3FileSystemPlugin implements FileSystemPlugin {
     @VisibleForTesting
     org.apache.hadoop.conf.Configuration getHadoopConfiguration(Configuration flussConfig) {
         org.apache.hadoop.conf.Configuration conf = new org.apache.hadoop.conf.Configuration();
-
-        // Credential providers must be explicitly set by the user
-        conf.set(S3ConfigOptions.CREDENTIALS_PROVIDER_CONFIG_NAME, "");
-        conf.set(S3ConfigOptions.ASSUMED_ROLE_CREDENTIALS_PROVIDER_CONFIG_NAME, "");
 
         if (flussConfig == null) {
             return conf;
@@ -229,20 +223,15 @@ public class S3FileSystemPlugin implements FileSystemPlugin {
     }
 
     /**
-     * Determines whether the file system is initialized by a client or a server.
+     * Determines whether the file system is initialized on the client or server side.
      *
      * <p>The file system is considered client-initialized if:
      *
      * <ol>
      *   <li>All configuration options start with the prefix {@link
      *       S3FileSystemPlugin#CLIENT_PREFIX}, or
-     *   <li>No configuration options are provided.
+     *   <li>no configuration options are provided.
      * </ol>
-     *
-     * <p>The second case (no configuration options) is only valid when token delegation ({@link
-     * ConfigOptions#FILE_SYSTEM_S3_ENABLE_TOKEN_DELEGATION}) is <i>enabled</i> on the server. If
-     * token delegation is <i>disabled</i>, the client must provide at least one configuration
-     * option (e.g., a credential provider). Otherwise, the setup is invalid.
      *
      * <p>The file system is considered server-initialized if:
      *
@@ -251,16 +240,18 @@ public class S3FileSystemPlugin implements FileSystemPlugin {
      *   <li>None of the options start with the prefix {@link S3FileSystemPlugin#CLIENT_PREFIX}.
      * </ol>
      *
-     * <p>On the server side, there will always be at least one configuration option:
+     * <p>Any other scenario is invalid.
+     *
+     * <p>Note that on the server side, there will always be at least one configuration option:
      *
      * <ul>
-     *   <li>If {@link ConfigOptions#FILE_SYSTEM_S3_ENABLE_TOKEN_DELEGATION} is <code>true</code>
-     *       (default), then access key, secret key, endpoint, and region must be configured.
-     *   <li>If {@link ConfigOptions#FILE_SYSTEM_S3_ENABLE_TOKEN_DELEGATION} is <code>false</code>,
-     *       a credential provider must be configured.
+     *   <li>If token delegation is activated ({@link
+     *       ConfigOptions#FILE_SYSTEM_S3_ENABLE_TOKEN_DELEGATION} is <code>true</code>, which is
+     *       the default), then access key, secret key, endpoint, and region must be configured, or
+     *   <li>If token delegation is deactivated, {@link
+     *       ConfigOptions#FILE_SYSTEM_S3_ENABLE_TOKEN_DELEGATION} has to be explicitly set to
+     *       <code>false</code>.
      * </ul>
-     *
-     * <p>Any other scenario is invalid.
      *
      * @param config The configuration.
      * @return {@code true} if initialized by a client, {@code false} if initialized by a server.
@@ -285,7 +276,7 @@ public class S3FileSystemPlugin implements FileSystemPlugin {
         } else {
             LOG.error("Detected invalid configuration: {}", config);
             throw new InvalidConfigException(
-                    "Cannot initialize file system due to invalid configuration.");
+                    "Cannot initialize file system due to invalid configuration. If this exception occurs on the server side, make sure to provide access key, secret key, endpoint, and region when token delegation is activated.");
         }
     }
 
--- a/fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/catalog/FlinkTableFactory.java
+++ b/fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/catalog/FlinkTableFactory.java
@@ -247,16 +247,16 @@ public class FlinkTableFactory implements DynamicTableSourceFactory, DynamicTabl
                     }
                 });
 
-      // pass through all fluss options from flink config
-      try {
-        PropertiesUtils.extractAndRemovePrefix(flinkConfig.toMap(), FLUSS_PREFIX)
-          .forEach(flussConfig::setString);
-      } catch (NoSuchMethodError e) {
-        // Flink 1.18 does not have the toMap() method yet, see
-        // https://nightlies.apache.org/flink/flink-docs-release-1.18/api/java//org/apache/flink/configuration/ReadableConfig.html
-        LOG.warn(
-          "Passing config options with prefix 'fluss' via Flink config is only supported with Fluss Flink Connector 1.19 and higher. All config options with prefix 'fluss' will be ignored.");
-      }
+        // pass through all fluss options from flink config
+        try {
+            PropertiesUtils.extractAndRemovePrefix(flinkConfig.toMap(), FLUSS_PREFIX)
+                    .forEach(flussConfig::setString);
+        } catch (NoSuchMethodError e) {
+            // Flink 1.18 does not have the toMap() method yet, see
+            // https://nightlies.apache.org/flink/flink-docs-release-1.18/api/java//org/apache/flink/configuration/ReadableConfig.html
+            LOG.warn(
+                    "Passing config options with prefix 'fluss' via Flink config is only supported with Fluss Flink Connector 1.19 and higher. All config options with prefix 'fluss' will be ignored.");
+        }
 
         // Todo support LookupOptions.MAX_RETRIES. Currently, Fluss doesn't support connector level
         // retry. The option 'client.lookup.max-retries' is only for dealing with the
--- a/fluss-filesystems/fluss-fs-s3/src/main/java/org/apache/fluss/fs/s3/S3ConfigOptions.java
+++ b/fluss-filesystems/fluss-fs-s3/src/main/java/org/apache/fluss/fs/s3/S3ConfigOptions.java
@@ -20,6 +20,4 @@ package org.apache.fluss.fs.s3;
 /** S3 configuration options that are explicitly read by name. */
 public final class S3ConfigOptions {
     public static final String CREDENTIALS_PROVIDER_CONFIG_NAME = "fs.s3a.aws.credentials.provider";
-    public static final String ASSUMED_ROLE_CREDENTIALS_PROVIDER_CONFIG_NAME =
-            "fs.s3a.assumed.role.credentials.provider";
 }
--- a/fluss-filesystems/fluss-fs-s3/src/test/java/org/apache/fluss/fs/s3/S3FileSystemPluginTest.java
+++ b/fluss-filesystems/fluss-fs-s3/src/test/java/org/apache/fluss/fs/s3/S3FileSystemPluginTest.java
@@ -49,24 +49,28 @@ public class S3FileSystemPluginTest {
             Configuration configuration;
             boolean isClient;
             // Client
-            // Example: Activated token delegation on server side
+            // No options provided
             configuration = new Configuration();
             isClient = S3FileSystemPlugin.isClient(configuration);
             assertThat(isClient).isTrue();
 
-            // Example: S3-compatible object storage (e.g., MinIO); Token delegation must be
-            // deactivated on server side
+            // Options provided with "client.fs." prefix
             configuration = new Configuration();
             configuration.setString("client.fs.s3.access-key", "fluss");
             configuration.setString("client.fs.s3.secret-key", "12345678");
-            configuration.setString(
-                    "client.fs.s3.aws.credentials.provider",
-                    "org.apache.hadoop.fs.s3a.SimpleAWSCredentialsProvider");
             isClient = S3FileSystemPlugin.isClient(configuration);
             assertThat(isClient).isTrue();
 
             // Server
-            // Example: Activated token delegation
+            // Activated token delegation
+            configuration = new Configuration();
+            configuration.setString("s3.access-key", "fluss");
+            configuration.setString("s3.secret-key", "12345678");
+            configuration.setString("s3.endpoint", "s3://fluss-data/");
+            configuration.setString("s3.region", "us-east-1");
+            isClient = S3FileSystemPlugin.isClient(configuration);
+            assertThat(isClient).isFalse();
+
             configuration = new Configuration();
             configuration.setBoolean(ConfigOptions.FILE_SYSTEM_S3_ENABLE_TOKEN_DELEGATION, true);
             configuration.setString("s3.access-key", "fluss");
@@ -76,15 +80,18 @@ public class S3FileSystemPluginTest {
             isClient = S3FileSystemPlugin.isClient(configuration);
             assertThat(isClient).isFalse();
 
-            // Example: Deactivate token delegation with S3-compatible object storage (e.g., MinIO)
+            // Example: Deactivated token delegation
             configuration = new Configuration();
+            configuration.setBoolean(ConfigOptions.FILE_SYSTEM_S3_ENABLE_TOKEN_DELEGATION, false);
+            isClient = S3FileSystemPlugin.isClient(configuration);
+            assertThat(isClient).isFalse();
+
+            configuration = new Configuration();
+            configuration.setBoolean(ConfigOptions.FILE_SYSTEM_S3_ENABLE_TOKEN_DELEGATION, false);
             configuration.setString("s3.access-key", "fluss");
             configuration.setString("s3.secret-key", "12345678");
             configuration.setString("s3.endpoint", "http://minio:9000");
             configuration.setString("s3.path-style-access", "true");
-            configuration.setString(
-                    "s3.aws.credentials.provider",
-                    "org.apache.hadoop.fs.s3a.SimpleAWSCredentialsProvider");
             isClient = S3FileSystemPlugin.isClient(configuration);
             assertThat(isClient).isFalse();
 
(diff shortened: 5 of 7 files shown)


fluss-client/src/main/java/com/alibaba/fluss/client/FlussConnection.java
```diff
@@ -72,7 +72,7 @@ public final class FlussConnection implements Connection {
         // only pass options with 'client.fs.' prefix
         FileSystem.initialize(
                 Configuration.fromMap(
-                        extractPrefix(new HashMap<>(conf.toMap()), CLIENT_PREFIX + "fs.")),
+                        extractAndRemovePrefix(new HashMap<>(conf.toMap()), CLIENT_PREFIX + "fs.")),
```
michaelkoepf: Shouldn't the prefix be removed? Otherwise the filesystem plugin will not recognize them.

fluss-filesystems/fluss-fs-s3/src/main/java/org/apache/fluss/fs/s3/token/S3DelegationTokenReceiver.java
```diff
@@ -59,13 +63,8 @@ public static void updateHadoopConfig(org.apache.hadoop.conf.Configuration hadoo
             LOG.debug("Provider already exists");
         }
 
-        // then, set addition info
-        if (additionInfos == null) {
```
michaelkoepf, line 63: How do we actually ensure here that there is no race condition? i.e., that additionInfos is set before this method is called?

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/catalog/FlinkTableFactory.java
```diff
@@ -232,6 +238,17 @@ private static Configuration toFlussClientConfig(
                     }
                 });
 
+        // pass through all fluss options from flink config
+        try {
+            PropertiesUtils.extractAndRemovePrefix(flinkConfig.toMap(), FLUSS_PREFIX)
+                    .forEach(flussConfig::setString);
+        } catch (NoSuchMethodError e) {
+            // Flink 1.18 does not have the toMap() method yet, see
```
michaelkoepf, line 255: I added a note in the documentation for this.

fluss-client/src/main/java/com/alibaba/fluss/client/FlussConnection.java
luoyuxia: It's by design to avoid forward option directly to filesystem in here. The filesystem should decide to accept the options or not, and then remove the prefix..

This is to avoid client pass any option to filesystem to broke the filesystem's behavior. It's more like a white list that the corresponding filesystem can decide what option to accept.

fluss-filesystems/fluss-fs-s3/src/main/java/org/apache/fluss/fs/s3/token/S3DelegationTokenReceiver.java
luoyuxia: What do you mean race condition? additionInfos may be set before or after this method is called.

fluss-client/src/main/java/com/alibaba/fluss/client/FlussConnection.java
michaelkoepf: ok, will revert this change and whitelist options in the filesystem with the `CLIENT_PREFIX + "fs."` prefix.

i suggest to whitelist

`access-key`, `secret-key`, `path-style-access`, `region`, `endpoint` and `aws.credentials.provider` (and all of their equivalents, e.g., `access.key`).

wdyt?

fluss-filesystems/fluss-fs-s3/src/main/java/org/apache/fluss/fs/s3/token/S3DelegationTokenReceiver.java
michaelkoepf: sorry, not the entire code block i want to show is visible in the comment.

this code block is from `com.alibaba.fluss.fs.s3.token.S3DelegationTokenReceiver#updateHadoopConfig` and is _only called in file system creation_. it expects `additionInfos` to be set. if not, an exception is thrown.
```
if (additionInfos == null) {
// if addition info is null, it also means we have not received any token,
    // we throw InvalidCredentialsException
    throw new NoAwsCredentialsException(DynamicTemporaryAWSCredentialsProvider.COMPONENT);
} else {
    for (Map.Entry<String, String> entry : additionInfos.entrySet()) {
        hadoopConfig.set(entry.getKey(), entry.getValue());
    }
}
```

however, `additionInfos` is only written to in `com.alibaba.fluss.fs.s3.token.S3DelegationTokenReceiver#onNewTokensObtained()` which is only called by `SecurityTokenReceiverRepository`. 

1. how do we ensure that `com.alibaba.fluss.fs.s3.token.S3DelegationTokenReceiver#onNewTokensObtained()` was called, so we do not end up throwing an exception during file system creation?
2. couldn't we just move this code block

```
for (Map.Entry<String, String> entry : additionInfos.entrySet()) {
    hadoopConfig.set(entry.getKey(), entry.getValue());
```

into `com.alibaba.fluss.fs.s3.token.S3DelegationTokenReceiver#onNewTokensObtained()` and remove it in `com.alibaba.fluss.fs.s3.token.S3DelegationTokenReceiver#updateHadoopConfig`?

luoyuxia: 1: Yes, you are right. it may happen that it may be end up throwing an exception. But I think it happens rarely. If it happens, it can always retry, as a flink job. In the firstly version, I block it to wait until the token is received. But Jark think it's hard to debug if it hangs whiling obtaining token.  So, in here, I just throw the exception
2: I'm not sure how it can solve the problem. The `Filesystem` will call `updateHadoopConfig` to init a filesytem, even though we move the code block into `onNewTokensObtained`, but it may still happen that the token is not avaiable, and we still can't init the filsystem.

fluss-client/src/main/java/com/alibaba/fluss/client/FlussConnection.java
luoyuxia: Do you means each filesystem can decide their own whitelist? if so, it sounds good to me.

fluss-filesystems/fluss-fs-s3/src/main/java/com/alibaba/fluss/fs/s3/token/S3DelegationTokenReceiver.java
```diff
@@ -80,16 +79,21 @@ public String scheme() {
 
     @Override
     public void onNewTokensObtained(ObtainedSecurityToken token) {
-        LOG.info("Updating session credentials");
+        LOG.info("Trying to update session credentials");
 
         byte[] tokenBytes = token.getToken();
+        if (tokenBytes.length != 0) {
+            credentials = CredentialsJsonSerde.fromJson(tokenBytes);
```
michaelkoepf: shouldn't we lock on `credentials` and `additionalInfos` to be safe? both fields are accessed from separate threads (one reads, the other one writes) and assignment may not be a single machine level instruction.

ka-steve:
Do you have a planned timeline for the release date of this feature? In 0.8, S3 can only be used with baked-in permanent credentials, which, I assume, prevents most AWS customers using Fluss in production.

michaelkoepf:
In general, this PR is ready for review/testing. We need to figure out 2 things.

1. If AWS Credential Providers with Temporary Credentials (e.g., the IAM Credential Provider) finally work. @polyzos can provide AWS access. We need somebody who has experience in setting up different temporary authentication methods with S3 and EC2.
2. There is some anomaly that I currently cannot make sense of (see below).

The idea of the PR is to deactivate token delegation to be able to use other credential providers. Without token delegation, users should be forced to set the credential provider to avoid misconfiguration or use unsafe credential providers that use long-term credentials. A more detailed description is in the [update docs of the PR](https://github.com/apache/fluss/blob/2e4c770240817ee51b5e34747a0330d58b16ffec/website/docs/maintenance/filesystems/s3.md). To force users to set the credential provider when token delegation is deactivated, I set the credential provider config options to blank, see [here](https://github.com/apache/fluss/blob/2e4c770240817ee51b5e34747a0330d58b16ffec/fluss-filesystems/fluss-fs-s3/src/main/java/org/apache/fluss/fs/s3/S3FileSystemPlugin.java#L148).

If you check out the PR and follow the instructions under `Example usage with MinIO` (build Fluss from source, build Docker images with the specified names, etc.) you can see that `s3.aws.credentials.provider: org.apache.hadoop.fs.s3a.SimpleAWSCredentialsProvider` is set in the Coordinator and Tablet Server, and everything works. Data is written to MinIO.

Now the anomaly: Shut down the Docker Compose stack. Remove `s3.aws.credentials.provider: org.apache.hadoop.fs.s3a.SimpleAWSCredentialsProvider` from the Coordinator and Tablet server configuration **but leave the credentials (access key, secret) there**. Spin up the Docker Compose stack again. Follow the instructions in the example. For some reason, data is still written to MinIO. This is a contradiction to me, because I set the credential providers by default to blank.

polyzos:
@michaelkoepf  @luoyuxia I tested it out, and it works. But then I tried to enable both lake tiering.
Seems like im getting different behaviors between runs.. On the first case, although tiering worked fine, lake tiering wasn't adding any data.
On the second run i got this exception
```
Caused by: org.apache.fluss.exception.FlussRuntimeException: Failed to get snapshot metadata for table bucket TableBucket{tableId=1, bucket=1}, snapshot ID: 0, Table: ecommercedb.users
	at org.apache.fluss.client.table.scanner.TableScan.createBatchScanner(TableScan.java:138)
	at org.apache.fluss.flink.tiering.source.TieringSplitReader.checkSplitOrStartNext(TieringSplitReader.java:196)
	at org.apache.fluss.flink.tiering.source.TieringSplitReader.fetch(TieringSplitReader.java:123)
	at org.apache.flink.connector.base.source.reader.fetcher.FetchTask.run(FetchTask.java:58)
	at org.apache.flink.connector.base.source.reader.fetcher.SplitFetcher.runOnce(SplitFetcher.java:165)
	... 6 more
Caused by: java.util.concurrent.ExecutionException: org.apache.fluss.exception.KvSnapshotNotExistException: Failed to get kv snapshot metadata for table bucket TableBucket{tableId=1, bucket=1} and snapshot id 0. Error: Kv snapshot not found
	at java.base/java.util.concurrent.CompletableFuture.reportGet(Unknown Source)
	at java.base/java.util.concurrent.CompletableFuture.get(Unknown Source)
	at org.apache.fluss.client.table.scanner.TableScan.createBatchScanner(TableScan.java:135)
```

Can someone test if you can make it work with both? 
**PS: On AWS S3, both work fine**

leekeiabstraction:
> Remove s3.aws.credentials.provider: org.apache.hadoop.fs.s3a.SimpleAWSCredentialsProvider from the Coordinator and Tablet server configuration but leave the credentials (access key, secret) there. Spin up the Docker Compose stack again. Follow the instructions in the example. For some reason, data is still written to MinIO. This is a contradiction to me, because I set the credential providers by default to blank.

I might be missing something but the documented behaviour of unsetting `fs.s3a.assumed.role.credentials.provider` is to fallback to BasicAWSCredentialsProvider.
See here: [A list of providers can be set in fs.s3a.assumed.role.credentials.provider; if unset the standard BasicAWSCredentialsProvider credential provider is used, which uses fs.s3a.access.key and fs.s3a.secret.key.](https://hadoop.apache.org/docs/stable/hadoop-aws/tools/hadoop-aws/assumed_roles.html#IOException:_.E2.80.9CAssumedRoleCredentialProvider_cannot_be_in_fs.s3a.assumed.role.credentials.provider.E2.80.9D)

Were `fs.s3a.secret.key/access.key` in the properties during the second run?

> To force users to set the credential provider when token delegation is deactivated, I set the credential provider config options to blank, see [here](https://github.com/apache/fluss/blob/2e4c770240817ee51b5e34747a0330d58b16ffec/fluss-filesystems/fluss-fs-s3/src/main/java/org/apache/fluss/fs/s3/S3FileSystemPlugin.java#L148).

Additional question that might help my understanding, what do we expect the user experience to look like here when they disable token delegation and forgot to set credential provider?

polyzos:
There might be more, but for me it boils down to two things... Currently when configuring Fluss tiering to work with minio there is this exception:
```
org.apache.fluss.exception.SecurityTokenException: Failed to get file access security token: The security token included in the request is invalid. (Service: AWSSecurityTokenService; Status Code: 403; Error Code: InvalidClientTokenId; Request ID: c3cb1c0c-57fc-4386-87ac-436622b40f21; Proxy: null)
taskmanager-1         | 2025-11-27 07:08:05,066 WARN  org.apache.fluss.client.token.DefaultSecurityTokenManager    [] - Failed to update tokens, will try again in 3600000 ms
taskmanager-1         | org.apache.fluss.exception.FlussRuntimeException: org.apache.fluss.exception.SecurityTokenException: Failed to get file access security token: The security token included in the request is invalid. (Service: AWSSecurityTokenService; Status Code: 403; Error Code: InvalidClientTokenId; Request ID: c3cb1c0c-57fc-4386-87ac-436622b40f21; Proxy: null)
taskmanager-1      
```
So my understanding is that when you run against MinIO, you typically use static S3 credentials and a custom endpoint. That works fine with Hadoop S3A if you set the right configs. Fluss’s `DefaultSecurityTokenManager` still tries to obtain file-access tokens from the server. Those tokens are AWS STS–based. If your environment isn’t wired to AWS STS (as with local MinIO), calls fail.
In the S3 filesystem plugin (S3FileSystemPlugin#setCredentialProvider), if `fs.s3a.access.key` is present, Fluss does NOT inject the STS-based credential provider. That part is good. However, the client still creates and starts DefaultSecurityTokenManager when the remote file downloader is used `FlussConnection#getOrCreateRemoteFileDownloader`, which forces the token flow and can fail before you even touch S3..

So I think @michaelkoepf solution should work.. If someone has time to try out different scenarios and see the behavior with some stronger understanding than me, it would be amazing.

**The second** thing to test and figure out is this:  https://github.com/apache/fluss/discussions/1911#discussioncomment-15043102
Michael also added a comment there.

michaelkoepf:
@leekeiabstraction _very strong_ catch. 

> Were fs.s3a.secret.key/access.key in the properties during the second run?

yes, that was the manual test scenario: "but leave the credentials (access key, secret) there". verification check, kind of. but i did miss out on the `fs.s3a.assumed.role.credentials.provider` behavior, indeed.

> Additional question that might help my understanding, what do we expect the user experience to look like here when they disable token delegation and forgot to set credential provider?

we would have treated this as a configuration error, which causes s3 access to fail. this mechanism was introduced as a safety net. users should prefer short-term credentials, as we discourage embedding long-term credentials in the client. if they want to use long-term credentials in the client, they should be explicit about it by setting the respective credential provider. 

any ideas how we can achieve this despite the behavior of `fs.s3a.assumed.role.credentials.provider`? maybe we can set `fs.s3a.assumed.role.credentials.provider` to a credential provider that uses short-term credentials and remove the long-term credential providers from `s3.aws.credentials.provider`'s default chain (instead of setting it to blank)?

leekeiabstraction:
> any ideas how we can achieve this despite the behavior of fs.s3a.assumed.role.credentials.provider? maybe we can set fs.s3a.assumed.role.credentials.provider to a credential provider that uses short-term credentials and remove the long-term credential providers from s3.aws.credentials.provider's default chain (instead of setting it to blank)?

I may need to sync to understand the user experience that we want in terms of happy cases as well. 

But if you intend to support this specific scenario ie user should run into error when they disable token delegation without specifying 's3.aws.credentials.provider', my suggestion is that we should check the configurations and fail-fast by throwing an exception at the configuration stage, rather than passing it on and relying on dependency's behaviour. 

Also, modifying credentials provider behind the scene might catch user by surprise.

michaelkoepf:
> fail-fast by throwing an exception at the configuration stage, rather than passing it on and relying on dependency's behaviour

i think this was considered in the beginning. the problem there was that deactivating the token delegation (`fs.s3.enable-token-delegation`) is introduces as a server side config. in the client, we always prepend the token delegation credential provider to the chain and at runtime, a valid provider should be figured out automatically. on initialization of the file system, the client is not aware whether token delegation will be used or not. in general, you could also introduce the config on the client side, but keeping a boolean flag (`fs.s3.enable-token-delegation: `{true|false}`) in sync across clients and server was considered even less user friendly than the solution with setting the credential provider explicitly, iirc.

bottom line: i think the easiest way is to just leave default providers in place, if  possible. this would need to be further investigated, because config options are also used to figure out whether we are on the server or client side (see [here](https://github.com/apache/fluss/blob/2e4c770240817ee51b5e34747a0330d58b16ffec/fluss-filesystems/fluss-fs-s3/src/main/java/org/apache/fluss/fs/s3/S3FileSystemPlugin.java#L270)). but actually, i think checking if `fs.s3.enable-token-delegation` is present (no matter if deactivated or not) is sufficient for checking if we are on the server or client side, because it's not whitelisted for clients.

so i suggest, if it is possible to leave the credential providers in place, we put a warning in the docs that short-term credentials should be preferred if possible. wdyt based on your experience with hadoop aws in other projects? @leekeiabstraction

leekeiabstraction:
Hey @michaelkoepf 

> so i suggest, if it is possible to leave the credential providers in place, we put a warning in the docs that short-term credentials should be preferred if possible. wdyt based on your experience with hadoop aws in other projects?

I've not used TD in the past so taking my time to understand the scenarios here. 

Warning in logs and also documentation seems sensible to me but before recommending that, I'd like to understand the fuller picture as there are a few components that I do not have context on: MinIO and when/where configurations are defined. 

Can you elaborate the following? Happy to sync on huddle

S3

||Coordinator Server|Tablet Server|Client|Current client experience|Intended client experience after this PR|
|-|-|-|-|-|-|
|Token Delegation Enabled?|true|true|true|Session delegation token requested by client from STS?|Session delegation token requested by client from STS?|
|Token Delegation Enabled?|true|true|false|?|We're discussing this|
|Token Delegation Enabled?|false|false|false|N/A token delegation cannot be disabled|?|
|Token Delegation Enabled?|false|false|true|N/A token delegation cannot be disabled|?|

MinIO

||Coordinator Server|Tablet Server|Client|Current client experience|Intended client experience after this PR|
|-|-|-|-|-|-|
|Token Delegation Enabled?|true|true|true|?|?|
|Token Delegation Enabled?|true|true|false|?|We're discussing this|
|Token Delegation Enabled?|false|false|false|N/A token delegation cannot be disabled|?|
|Token Delegation Enabled?|false|false|true|N/A token delegation cannot be disabled|?|

fluss-common/src/main/java/org/apache/fluss/config/ConfigOptions.java
```diff
@@ -1923,4 +1923,19 @@ private static class ConfigOptionsHolder {
...
         return ConfigOptionsHolder.CONFIG_OPTIONS_BY_KEY.get(key);
     }
+
+    // ------------------------------------------------------------------------
+    //  ConfigOptions for file systems (Fluss-internal, not recognized by Hadoop FS libraries)
+    // ------------------------------------------------------------------------
+    public static final ConfigOption<Boolean> FILE_SYSTEM_S3_ENABLE_TOKEN_DELEGATION =
+            key("fs.s3.enable-token-delegation")
+                    .booleanType()
+                    .defaultValue(true)
```
leekeiabstraction: What is the motivation behind enabling token delegation by default? Is it primarily because it is currently enabled by default?

Given the nuances behind token delegation mechanism, and that it does not work for minIO, I believe it can be argued that disabling by default is a better user experience than say, user being unaware of token delegation and mistakenly believe that revoking client side access can be done by the revoking access of client side credentials provider that they have configured (won't work since token delegation creds provider takes priority over all other creds provider)

fluss-filesystems/fluss-fs-s3/src/main/java/org/apache/fluss/fs/s3/S3FileSystemPlugin.java
```diff
@@ -61,40 +87,115 @@ public FileSystem create(URI fsUri, Configuration flussConfig) throws IOExceptio
         org.apache.hadoop.conf.Configuration hadoopConfig =
                 mirrorCertainHadoopConfig(getHadoopConfiguration(flussConfig));
 
-        // set credential provider
-        setCredentialProvider(hadoopConfig);
+        final boolean isClient = isClient(flussConfig);
+        final boolean useTokenDelegation;
```
leekeiabstraction, line 91: Rename from `useTokenDelegation` to `shouldServeToken` or `shouldDelegateToken`  as technically the client side still uses token delegation if server is configured to perform token delegation.

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/catalog/FlinkTableFactory.java
```diff
@@ -241,6 +247,17 @@ private static Configuration toFlussClientConfig(
                     }
                 });
 
+        // pass through all fluss options from flink config
+        try {
+            PropertiesUtils.extractAndRemovePrefix(flinkConfig.toMap(), FLUSS_PREFIX)
```
leekeiabstraction, line 252: Line 243 to 248 already forward table configs with prefix `client` to fluss config. 

This will remove any table specific config if the name clashes. In such cases, should table specific config take precedence over flink config?

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/catalog/FlinkTableFactory.java
```diff
@@ -241,6 +247,17 @@ private static Configuration toFlussClientConfig(
...
+                    .forEach(flussConfig::setString);
```
leekeiabstraction, line 253: Should we enforce/prepend`client.` prefix here? Technically, all config in FlinkTableFactory is Fluss client side config. As is, we are leaking config non client config through if user missed out the `client` part.. I believe that leads to the best effort handling logic that we have within `isClient(configuration)`

fluss-filesystems/fluss-fs-s3/src/main/java/org/apache/fluss/fs/s3/S3FileSystemPlugin.java
```diff
@@ -61,40 +87,115 @@ public FileSystem create(URI fsUri, Configuration flussConfig) throws IOExceptio
...
 
                     LOG.debug(
-                            "Adding Fluss config entry for {} as {} to Hadoop config", key, newKey);
+                            "Adding Fluss config entry for {} as {} to Hadoop config",
+                            flussKey,
+                            hadoopConfigKey);
+                }
+
+                String flussKeyClientPrefix = CLIENT_PREFIX + flussPrefix;
+                if (flussKey.startsWith(flussKeyClientPrefix)) {
```
leekeiabstraction, line 167: Suggest the following to avoid unnecessary alloc and comparison.

```java
else if (flussKey.startsWith(CLIENT_PREFIX + flussPrefix)) {
```

fluss-filesystems/fluss-fs-s3/src/main/java/org/apache/fluss/fs/s3/token/S3DelegationTokenReceiver.java
```diff
@@ -35,42 +40,54 @@
...
+    public static void updateHadoopConfigCredentialProviders(
+            org.apache.hadoop.conf.Configuration hadoopConfig, List<String> credentialProvider) {
+        LOG.info("Updating credential providers in Hadoop configuration");
+
+        String providers = hadoopConfig.get(S3ConfigOptions.CREDENTIALS_PROVIDER_CONFIG_NAME, "");
+        List<String> credentialProviderPrependOrder = new ArrayList<>(credentialProvider);
+        Collections.reverse(credentialProviderPrependOrder);
+
+        for (String credentialProviderName : credentialProviderPrependOrder) {
+            if (!providers.contains(credentialProviderName)) {
```
leekeiabstraction, line 57: We should use exact match instead of contains.

leekeiabstraction: Also, if I'm reading correctly, we currently only add a single credential provider `Collections.singletonList(DynamicTemporaryAWSCredentialsProvider.NAME)` from S3FileSystemPlugin.java:102. Do we need to handle updating and adding a list of credentialProviders anywhere? If not, it might be worth simplifying the code and remove the updating/adding of list of credentialProviders

binary-signal:
> Now the anomaly: Shut down the Docker Compose stack. Remove `s3.aws.credentials.provider: org.apache.hadoop.fs.s3a.SimpleAWSCredentialsProvider` from the Coordinator and Tablet server configuration **but leave the credentials (access key, secret) there**. Spin up the Docker Compose stack again. Follow the instructions in the example. For some reason, data is still written to MinIO. This is a contradiction to me, because I set the credential providers by default to blank.

Unless you force the containers to recreate, they will keep using the old configuration. That's why you can still write to MinIO.

michaelkoepf:
subsumed by #2662

PR #1254, opened 2025-07-02
Add exception in throwable in lake writer & committer

polyzos:
Small improvement to add a exception root cause when trying to debug why the lakewriter or committer fail.


luoyuxia (approved):
@polyzos Thanks. LGTM

PR #1257, opened 2025-07-03
bugfix: fix exception type during reading from removed partitions

### LINKED ISSUE #1256: Client receives unclear exception when accessing removed partitions
### Search before asking

### Fluss version

0.7.0 (latest release)

### Please describe the bug 🐞

While we have implemented several latch mechanisms (e.g., Java locks) to protect atomicity in short critical sections, we still lack a proper database lock system to ensure ACID properties (specifically atomicity, consistency, and isolation) for DDL transactions. This allows concurrent operations to access intermediate states of DDL 'transactions', violating meta integrity. 

For example:
- RemoteLogDownloader may throw "No such file or directory" exceptions during partition removal
- ReplicaManager may throw LogOffsetOutOfRangeException during concurrent reads on deleted partitions

Though full ACID support for DDL transactions may be considered in future, we should initially:
1. Fix the exception types thrown in these scenarios
2. Make user code clearly identify and handle such errors

### Solution

_No response_

platinumhamburg:
### Purpose

Linked issue: close #1256
Fix the exception types thrown when concurrent partition drop happened during read.

### DIFF
--- a/fluss-client/src/main/java/org/apache/fluss/client/table/scanner/log/LogFetcher.java
+++ b/fluss-client/src/main/java/org/apache/fluss/client/table/scanner/log/LogFetcher.java
@@ -143,6 +143,7 @@ public class LogFetcher implements Closeable {
         this.remoteLogDownloader =
                 new RemoteLogDownloader(
                         tablePath, conf, remoteFileDownloader, scannerMetricGroup, metadataUpdater);
+        this.remoteLogDownloader.start();
     }
 
     /**


fluss-server/src/main/java/com/alibaba/fluss/server/replica/ReplicaManager.java
```diff
@@ -1081,6 +1083,28 @@ else if (canFetchFromRemoteLog(replica, fetchOffset)) {
...
+                // to protect atomicity in several short critical sections, we lack a true database
+                // lock system to ensure ACID properties (specifically atomicity, consistency,
+                // isolation) for DDL transactions. This limitation warrants further discussion in
+                // the future.
+                if (tb.getPartitionId() != null) {
+                    PartitionMetadata partitionMetadata =
+                            metadataCache.getPartitionMetadata(replica.getPhysicalTablePath());
+                    if (partitionMetadata == null
+                            || partitionMetadata.getPartitionId()
+                                    == PartitionMetadata.DELETED_PARTITION_ID) {
```
swuferhong: Why this parameter `PartitionMetadata.DELETED_PARTITION_ID` will be exposed to the client? This seems to define a protocol for the `Coordinator` to send `UpdateMetadata` RPC to `TabletServers`, but it shouldn’t represent an actual valid value accessible to clients.

Issue #1206, opened 2025-06-26
Flink sink checkpoint may continue to fail during cluster upgrading


swuferhong:
### Search before asking

- [x] I searched in the [issues](https://github.com/alibaba/fluss/issues) and found nothing similar.

### Fluss version

0.6.0 (latest release)

### Please describe the bug 🐞

When I try to upgrade server. I found that the Flink sink checkpoint may continue to fail. 

### Are you willing to submit a PR?

- [ ] I'm willing to submit a PR!

wuchong:
fixed in https://github.com/apache/fluss/pull/1258

PR #1258, opened 2025-07-03
[client] Fix the bug where the loss of writeBatch was caused by the origin batchLocation forget to update ServerNode info

swuferhong:
### Purpose

Linked issue: #1206


fluss-client/src/main/java/com/alibaba/fluss/client/utils/MetadataUtils.java
```diff
@@ -153,6 +152,19 @@ public static Cluster sendMetadataRequestAndRebuildCluster(
...
+                            // update the leader id to check whether the leader is alive or not.
+                            newBucketLocations.forEach(
+                                    (physicalTablePath, bucketLocations) ->
+                                            bucketLocations.forEach(
+                                                    bucketLocation -> {
+                                                        Integer leader = bucketLocation.getLeader();
+                                                        if (leader != null
+                                                                && newAliveTabletServers.get(leader)
+                                                                        == null) {
+                                                            bucketLocation.setLeader(null);
```
wuchong: This changes `leader` into a mutable variable which is hack and may have concurrency problems in the future.  From my understanding, it's fine to have `leader` non-null when the corresponding TabletServer ServerNode is null. Could you change and verify it?

fluss-client/src/main/java/com/alibaba/fluss/client/write/RecordAccumulator.java
```diff
@@ -283,16 +283,19 @@ public ReadyCheckResult ready(Cluster cluster) {
...
         Map<Integer, List<ReadyWriteBatch>> batches = new HashMap<>();
-        for (ServerNode node : nodes) {
+        for (Integer node : nodes) {
             List<ReadyWriteBatch> ready = drainBatchesForOneNode(cluster, node, maxSize);
             if (!ready.isEmpty()) {
-                batches.put(node.id(), ready);
+                List<ReadyWriteBatch> previous = batches.put(node, ready);
+                if (previous != null) {
+                    throw new IllegalStateException("Duplicate node in nodes when drain: " + node);
+                }
```
wuchong: nit: the checking is redundant because `nodes` is a set, and there shouldn't be duplicates in it.

fluss-client/src/main/java/com/alibaba/fluss/client/write/RecordAccumulator.java
```diff
@@ -531,10 +534,10 @@ private long bucketReady(
...
     private long batchReady(
             boolean exhausted,
-            ServerNode leader,
+            Integer leader,
             long waitedTimeMs,
             boolean full,
-            Set<ServerNode> readyNodes,
+            Set<Integer> readyNodes,
             long nextReadyCheckDelayMs) {
         if (!readyNodes.contains(leader)) {
```
wuchong, line 539: Add `leader != null` check before the `contains` check, or change the `leader` parameter into `int` type.

swuferhong: Change to `int`, as the `!= null` checking already done for `leader` while call this method.

fluss-client/src/main/java/com/alibaba/fluss/client/utils/MetadataUtils.java
swuferhong: I didn't encounter errors after removing it during local testing and simulated upgrade/downgrade job tests. We can proceed with removing it for now.

swuferhong:
@wuchong comments addressed.

PR #1267, opened 2025-07-04
[client] Fluss client shouldn't load plugin by thread context classloader.

### LINKED ISSUE #1262: Fluss client shouldn't load plugin by thread contextclassloader.
### Search before asking

### Fluss version

0.7.0 (latest release)

### Please describe the bug 🐞

Currently, when we put fluss connector into flink lib, the job preforms well. When I run with "--jar" or "add jar", the sql-client will fail with :
```java
ERROR] Could not execute SQL statement. Reason:
com.alibaba.fluss.metadata.ValidationException: No plugin for the protocol 'sasl' is found in the classpath
```

It's because when sql gateway run(in flink 1.20), the thread classloader is not the user classloader. Thus, though user classloader can loaded flink-fluss-1.20.jar, but the thread classloader cannot do it.

### Solution

_No response_

loserwang1024:
### Purpose

Linked issue: close #1262

### DIFF
--- a/fluss-flink/fluss-flink-common/src/test/java/com/alibaba/fluss/flink/catalog/FlinkCatalogITCase.java
+++ b/fluss-flink/fluss-flink-common/src/test/java/com/alibaba/fluss/flink/catalog/FlinkCatalogITCase.java
@@ -23,8 +23,6 @@ import com.alibaba.fluss.config.Configuration;
 import com.alibaba.fluss.exception.InvalidTableException;
 import com.alibaba.fluss.metadata.TablePath;
 import com.alibaba.fluss.server.testutils.FlussClusterExtension;
-import com.alibaba.fluss.utils.ParentResourceBlockingClassLoader;
-import com.alibaba.fluss.utils.TemporaryClassLoaderContext;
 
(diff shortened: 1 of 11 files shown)


loserwang1024:
@wuchong @leonardBang , CC

fluss-flink/fluss-flink-common/src/test/java/com/alibaba/fluss/flink/catalog/FlinkCatalogITCase.java
```diff
@@ -657,6 +660,28 @@ void testAuthentication() throws Exception {
         }
     }
 
+    // this test is to mock `--jar` which is not loaded by the app classloader.
+    @Test
+    void testNotIncludedInThreadContextClassloader() throws Exception {
```
wuchong: Move to `FlinkAuthorizationITCase`, and refactored into assert the success of catalog registration.

loserwang1024: modifed it.

PR #1303, opened 2025-07-10
Fluss admin use tablet server for read only operation to reduce the cost of coordinator.

### LINKED ISSUE #1302: Fluss admin use  tablet server for read only operation to reduce the cost of coordinator.
### Search before asking

### Motivation

Currently, all the operation of admin is depended on coordinator, this will cause two problems:
1. the traffic of coordinator is too much.
2. If coordinator is unreachable, the flink job can not read and write.

### Solution

_No response_

### Anything else?

_No response_

### Willingness to contribute

loserwang1024:
### Purpose

Linked issue: close #1302

### DIFF
--- a/fluss-client/src/test/java/com/alibaba/fluss/client/admin/FlussAdminITCase.java
+++ b/fluss-client/src/test/java/com/alibaba/fluss/client/admin/FlussAdminITCase.java
@@ -379,7 +379,6 @@ class FlussAdminITCase extends ClientToServerITCaseBase {
 
         // assert the cluster should have tablet server number to be 3
         FLUSS_CLUSTER_EXTENSION.assertHasTabletServerNumber(3);
-        FLUSS_CLUSTER_EXTENSION.waitUtilAllGatewayHasSameMetadata();
 
         // we can create the table now
         admin.createTable(tablePath, DEFAULT_TABLE_DESCRIPTOR, false).get();
--- a/fluss-client/src/test/java/com/alibaba/fluss/client/admin/FlussAdminITCase.java
+++ b/fluss-client/src/test/java/com/alibaba/fluss/client/admin/FlussAdminITCase.java
@@ -130,13 +130,13 @@ class FlussAdminITCase extends ClientToServerITCaseBase {
     void testGetDatabaseInfo() throws Exception {
         long timestampBeforeCreate = System.currentTimeMillis();
         admin.createDatabase(
-                "test_db_2",
-                DatabaseDescriptor.builder()
-                        .comment("test comment")
-                        .customProperty("key1", "value1")
-                        .build(),
-                false);
-        FLUSS_CLUSTER_EXTENSION.waitUtilAllGatewayHasSameMetadata();
+                        "test_db_2",
+                        DatabaseDescriptor.builder()
+                                .comment("test comment")
+                                .customProperty("key1", "value1")
+                                .build(),
+                        false)
+                .get();
         DatabaseInfo databaseInfo = admin.getDatabaseInfo("test_db_2").get();
         long timestampAfterCreate = System.currentTimeMillis();
         assertThat(databaseInfo.getCreatedTime()).isEqualTo(databaseInfo.getModifiedTime());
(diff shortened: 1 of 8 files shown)


loserwang1024:
@swuferhong , @wuchong , CC

fluss-server/src/test/java/com/alibaba/fluss/server/testutils/FlussClusterExtension.java
```diff
@@ -490,12 +492,20 @@ private TabletServerGateway newTabletServerClientForNode(ServerNode serverNode)
...
+        List<PbTablePath> allPbTables =
+                allTables.stream()
+                        .map(ServerRpcMessageUtils::fromTablePath)
+                        .collect(Collectors.toList());
         for (AdminReadOnlyGateway gateway : collectAllRpcGateways()) {
             retry(
                     Duration.ofMinutes(1),
                     () -> {
-                        MetadataResponse response = gateway.metadata(new MetadataRequest()).get();
+                        MetadataResponse response =
```
swuferhong: Why there need to check this? Read only admin operations all need to acquire zookeeper instead of metadata.

swuferhong (commented):
I left a minor comment.

mengnankkkk:
建议梳理只读 admin 操作的依赖，确保无不必要的 coordinator 依赖，测试代码中对 metadata 的获取是否可以移除或优化？
建议补充 tablet server 异常场景的测试用例，提升健壮性。
关键代码处建议增加注释，并在文档中说明只读路径的变化及影响。
明确 tablet server 不可用时的降级策略，避免只读操作不可用。
检查代码风格与项目保持一致。
如有性能数据，建议补充说明优化效果。

fluss-flink/fluss-flink-common/src/test/java/com/alibaba/fluss/flink/procedure/FlinkProcedureITCase.java
```diff
@@ -163,6 +170,16 @@ void testAclProcedureIgnoreCase(boolean upperCase) throws Exception {
...
+        List<AclBinding> aclBindings =
+                Collections.singletonList(
+                        new AclBinding(
+                                Resource.cluster(),
+                                new AccessControlEntry(
+                                        new FlussPrincipal("Alice", "User"),
+                                        "127.0.0.1",
+                                        OperationType.READ,
+                                        PermissionType.ALLOW)));
+        FLUSS_CLUSTER_EXTENSION.waitUtilAuthenticationSync(aclBindings, true);
```
wuchong: There shouldn't be any waiting required here. All write operations in `Admin` should provide **read-your-writes** consistency, meaning that once a write succeeds, all subsequent reads should return the updated data successfully. This is similar to how we ensure consistency for other write operations like `createTable` and `createDatabase`. Without this guarantee, it becomes very difficult for users to use the APIs reliably.

To avoid relying on eventual consistency or waiting mechanisms, the implementations of `com.alibaba.fluss.rpc.gateway.AdminReadOnlyGateway#listAcls` and `com.alibaba.fluss.server.authorizer.Authorizer#listAcls` should directly query ZooKeeper instead of using a cached value.

fluss-client/src/test/java/com/alibaba/fluss/client/admin/FlussAdminITCase.java
```diff
@@ -382,12 +383,16 @@ void testCreateTableWithInvalidReplicationFactor() throws Exception {
...
         admin.createTable(tablePath, DEFAULT_TABLE_DESCRIPTOR, false).get();
-        TableInfo tableInfo = admin.getTableInfo(DEFAULT_TABLE_PATH).get();
-        assertThat(tableInfo.toTableDescriptor())
-                .isEqualTo(
-                        DEFAULT_TABLE_DESCRIPTOR
-                                .withReplicationFactor(3)
-                                .withDataLakeFormat(DataLakeFormat.PAIMON));
+        // recreate the connection because the metadata of tablet server has changed
+        try (Connection conn = ConnectionFactory.createConnection(clientConf);
+                Admin admin = conn.getAdmin()) {
```
wuchong, line 387: Why does this require to create a new Admin for retrieving the correct table info? I think this also breaks the data consistency guarantee of the Admin API.

fluss-flink/fluss-flink-common/src/test/java/com/alibaba/fluss/flink/procedure/FlinkProcedureITCase.java
loserwang1024: com.alibaba.fluss.rpc.gateway.AdminReadOnlyGateway#listAcls and com.alibaba.fluss.rpc.gateway.AdminReadOnlyGateway#dropAcls is not deleted by a resource but AclBindingFilter, which means that 
1. we have to get all the resource's acl bindings for each operation without cache. It's too heavy.
2. If a use is not right to list or drop acl, still need to get all the resource's acl bindings before determining it.

fluss-client/src/test/java/com/alibaba/fluss/client/admin/FlussAdminITCase.java
loserwang1024: Because the tableserlet have been closed, if reuse connection, exception will be thrown. Maybe we need redesign the Admin to recreate new connection if the currently connection is not valid.

PR #1309, opened 2025-07-11
[website] Fix blog post author titles

michaelkoepf:
For podlings (= projects in the Apache Incubator), there is technically no "PMC", only Incabutor PMC (IPMC) and Podling PMC (PPMC). "PMC" is for graduated (top-level) projects. Since the title is shown on our _official_ blog post, IMHO we should be precise here.

> Each Podling Project Management Committee (PPMC) helps its Podling learn how to govern itself. It works **_like a PMC but_** reports to the Incubator PMC instead of to the ASF Board. Initially, it is composed of the Podling’s mentors and the initial committers. The PPMC is directly responsible for the oversight of the podling, and it also decides who to add as a PPMC member.

-- Source: https://incubator.apache.org/guides/ppmc.html

### DIFF
--- a/website/blog/authors.yml
+++ b/website/blog/authors.yml
@@ -18,19 +18,19 @@
 
 jark:
   name: Jark Wu
-  title: PMC member of Apache Fluss
+  title: PPMC member of Apache Fluss (Incubating)
   url: https://github.com/wuchong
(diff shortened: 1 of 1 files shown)


wuchong (approved):
You are right. Thanks for fixing this.

Issue #1341, opened 2025-07-15
Introduce key encoding for iceberg


luoyuxia:
### Search before asking

- [x] I searched in the [issues](https://github.com/alibaba/fluss/issues) and found nothing similar.

### Description

Introduce key encoding for iceberg 

### Willingness to contribute

- [ ] I'm willing to submit a PR!

luoyuxia:
See #1327

MehulBatra:
I would like to take it @luoyuxia 
Thank you!

Issue #1195, opened 2025-06-25
Upgrade JDK compile version as 11


loserwang1024:
### Search before asking

- [x] I searched in the [issues](https://github.com/alibaba/fluss/issues) and found nothing similar.

## Why need to upgrade jdk 11?
Currently, though we can compile fluss with jdk 8 and 11 by `mvn` command. But the target version is still java 8.

There are still some problems when use jdk11 to compile.

* the generated code is still target version 8 rather than 11, this means:

  1. We cannot use some improvement of java 11.
  2. When adding flink 2.x, if not compile and package, no problem. However, If we want to run tests  , we must compile this as java 11.

-  When using develop tools(such as idea), if we want to use JDK 11 and set language level as 11, the problem will occurs: Type `sun.misc.Unsafe` was not found`.

For better proformance and flink 2.x support, it's better to compile to java 11.

### 1. Upgrade JDK compile version as 11 

If user want to compile as java 8, just add  `-Pjava8-target` to mvn command.
For example, If user want to use

```java
mvn clean install -DskipTests -Pjava8-target

```

If user want to test it in idea with JDK 8,  just  choose the profile and then rebuild.

![Image](https://github.com/user-attachments/assets/e0be595e-4f09-44c5-b3c2-e8ab814a452c)

### 2. ci.yaml 

   1. add a new job to compile and run in jdk 11. In next version, only java 11 is provided.
   2. still reserve compile test in jdk8(but not run anymore) to guarantee that if user need a java 8 jar, they can compile by themselves.

### Willingness to contribute

- [x] I'm willing to submit a PR!

wuchong:
@loserwang1024, please search existing issues before creating a new one to avoid duplicates. Usually, the original issue already has many people watching, so it's best to follow up there if possible.

PR #1350, opened 2025-07-16
[lake/iceberg] Iceberg encoding strategy

MehulBatra:
### Purpose
Implement Iceberg encoding strategy for Fluss.

Linked Issue: #1341

### Brief change log
- Add `DataLakeFormat.ICEBERG` enum value
- Implementing `IcebergKeyEncoder` using the binary writer for key encoding
- Implemented 'IcebergBinaryRowWriter' to spport encoding

### Tests
Wrote a Unit tests to test the encoding matches between fluss and iceberg

### API and Format
None

### Documentation
- New feature: Iceberg lake format support for key encoding and bucketing


fluss-common/src/main/java/com/alibaba/fluss/row/encode/iceberg/IcebergBinaryRowWriter.java
```diff
@@ -0,0 +1,390 @@
...
+    }
+
+    public void setNullAt(int pos) {
+        // Iceberg handles nulls with a null marker byte
+        ensureCapacity(1);
+        segment.put(cursor, (byte) 0x00); // Null marker
+        cursor += 1;
+    }
+
+    public void writeChangeType(ChangeType kind) {
```
luoyuxia: We don't need to write change type, we only use the key encoder to encode columns not change type.

fluss-common/src/main/java/com/alibaba/fluss/row/encode/iceberg/IcebergBinaryRowWriter.java
```diff
@@ -0,0 +1,390 @@
...
+import com.alibaba.fluss.row.TimestampNtz;
+import com.alibaba.fluss.types.DataType;
+
+import java.io.Serializable;
+import java.nio.charset.StandardCharsets;
+import java.util.Arrays;
+
+import static com.alibaba.fluss.types.DataTypeChecks.getPrecision;
+
+/* This file is based on source code of Apache Iceberg Project (https://iceberg.apache.org/), licensed by the Apache
```
luoyuxia: Does it based on apache iceberg code? If not, 
We can remove this since this file is not based on `Apache Iceberg Project`

fluss-common/src/main/java/com/alibaba/fluss/row/encode/KeyEncoder.java
```diff
@@ -46,6 +47,8 @@ static KeyEncoder of(
             return CompactedKeyEncoder.createKeyEncoder(rowType, keyFields);
         } else if (lakeFormat == DataLakeFormat.PAIMON) {
             return new PaimonKeyEncoder(rowType, keyFields);
+        } else if (lakeFormat == DataLakeFormat.ICEBERG) {
+            return new IcebergKeyEncoder(rowType, keyFields);
```
luoyuxia, line 51: As the FIP said, we only support one key field currently. 
```
checkArgument(
                    keyFields.size() == 1,
                    "key fields must have exactly one field for iceberg format.");
```

fluss-common/src/main/java/com/alibaba/fluss/row/encode/iceberg/IcebergBinaryRowWriter.java
```diff
@@ -0,0 +1,390 @@
...
+ * Software Foundation (ASF) under the Apache License, Version 2.0. See the NOTICE file distributed with this work for
+ * additional information regarding copyright ownership. */
+
+/**
+ * A writer to encode Fluss's {@link InternalRow} using Iceberg's binary encoding way.
```
luoyuxia: Add a link to refer to the code that iceberg encodes in iceberg repo for reference.

fluss-common/src/main/java/com/alibaba/fluss/row/encode/iceberg/IcebergKeyEncoder.java
```diff
@@ -0,0 +1,64 @@
...
+        }
+
+        icebergBinaryRowWriter = new IcebergBinaryRowWriter(keys.size());
+    }
+
+    @Override
+    public byte[] encodeKey(InternalRow row) {
+        icebergBinaryRowWriter.reset();
+
+        // TODO: Discuss with Yuxio Iceberg doesn't store ChangeType in binary format like Paimon
```
luoyuxia: we don't need store ChangeType, paimon store it just because we need to align to paimon's encoding since paimon always ecnoding change type. But icberg don't require it.

fluss-common/src/main/java/com/alibaba/fluss/row/encode/iceberg/IcebergBinaryRowWriter.java
```diff
@@ -0,0 +1,390 @@
...
+     * Creates an accessor for writing the elements of an iceberg binary row writer during runtime.
+     *
+     * @param fieldType the field type of the indexed row
+     */
+    public static FieldWriter createFieldWriter(DataType fieldType) {
+        final FieldWriter fieldWriter;
+        switch (fieldType.getTypeRoot()) {
+            case CHAR:
+            case STRING:
+                fieldWriter = (writer, pos, value) -> writer.writeString(pos, (BinaryString) value);
```
luoyuxia: nit: simplify to 
```
writer.writeString((BinaryString) value);
```
The other writers can also be simplified.

fluss-common/src/main/java/com/alibaba/fluss/row/encode/iceberg/IcebergBinaryRowWriter.java
```diff
@@ -0,0 +1,390 @@
...
+                writer.setNullAt(pos);
+            } else {
+                fieldWriter.writeField(writer, pos, value);
+            }
+        };
+    }
+
+    /** Accessor for writing the elements of an iceberg binary row writer during runtime. */
+    interface FieldWriter extends Serializable {
+        void writeField(IcebergBinaryRowWriter writer, int pos, Object value);
```
luoyuxia: nit:
```suggestion
        void writeField(IcebergBinaryRowWriter writer, Object value);
```

fluss-common/src/main/java/com/alibaba/fluss/row/encode/iceberg/IcebergBinaryRowWriter.java
```diff
@@ -0,0 +1,390 @@
...
+    }
+
+    public void writeDouble(int pos, double value) {
+        ensureCapacity(8);
+        segment.putDouble(cursor, value);
+        cursor += 8;
+    }
+
+    public void writeString(int pos, BinaryString input) {
+        if (input == null) {
```
luoyuxia: don't need to check is null or not since we already check it in
```
 if (value == null) {
     writer.setNullAt(pos);
  }
```

fluss-common/src/main/java/com/alibaba/fluss/row/encode/iceberg/IcebergBinaryRowWriter.java
```diff
@@ -0,0 +1,390 @@
...
+class IcebergBinaryRowWriter {
+
+    private final int arity;
+    private byte[] buffer;
+    private MemorySegment segment;
+    private int cursor;
+
+    public IcebergBinaryRowWriter(int arity) {
+        // Verify MemorySegment uses little-endian for Iceberg compatibility
+        if (!MemorySegment.LITTLE_ENDIAN) {
```
luoyuxia: You can always use UnsafeUtils to put int, long, etc, since it will always be little endian.
```
UnsafeUtils.putLong(buffer, cursor, value);
```

fluss-common/src/main/java/com/alibaba/fluss/row/encode/iceberg/IcebergBinaryRowWriter.java
```diff
@@ -0,0 +1,390 @@
...
+
+            remaining -= toCopy;
+            currentTargetOffset += toCopy;
+            currentOffset = 0;
+        }
+    }
+
+    private static int calculateInitialSize(int arity) {
+        // Conservative estimate for Iceberg format
+        return 64 + (arity * 16); // Allow for variable length fields
```
luoyuxia: Why use `64 + (arity * 16);`? I'd suggest to use `8 + (arity * 8);` for space save. No need to allocate a larger byte array.

fluss-common/src/main/java/com/alibaba/fluss/row/encode/iceberg/IcebergBinaryRowWriter.java
luoyuxia: For write string, let's reuse what we do in `CompactedKeyWriter` intead of introducing new piece of code:
```
public void writeString(int pos, BinaryString input) {
        if (input.getSegments() == null) {
            byte[] bytes = BinaryString.encodeUTF8(input.toString());
            writeBytes(pos, bytes);
        } else {
            writeSegments(input.getSegments(), input.getOffset(), input.getSizeInBytes());
        }
    }

private void writeSegments(MemorySegment[] segments, int off, int len) {
        writeInt(len);
        if (len + off <= segments[0].size()) {
            write(segments[0], off, len);
        } else {
            write(segments, off, len);
        }
    }

    private void write(MemorySegment segment, int off, int len) {
        ensureCapacity(len);
        segment.get(off, this.buffer, this.cursor, len);
        this.cursor += len;
    }

    private void write(MemorySegment[] segments, int off, int len) {
        ensureCapacity(len);
        int toWrite = len;
        int fromOffset = off;
        int toOffset = this.cursor;
        for (MemorySegment sourceSegment : segments) {
            int remain = sourceSegment.size() - fromOffset;
            if (remain > 0) {
                int localToWrite = Math.min(remain, toWrite);
                sourceSegment.get(fromOffset, buffer, toOffset, localToWrite);
                toWrite -= localToWrite;
                toOffset += localToWrite;
                fromOffset = 0;
            } else {
                fromOffset -= sourceSegment.size();
            }
        }
        this.cursor += len;
    }
```

fluss-common/src/main/java/com/alibaba/fluss/row/encode/iceberg/IcebergBinaryRowWriter.java
```diff
@@ -0,0 +1,390 @@
...
+        if (value == null) {
+            setNullAt(pos);
+            return;
+        }
+
+        // Iceberg stores decimals as big-endian byte arrays
+        byte[] unscaledBytes = value.toUnscaledBytes();
+
+        // Calculate required bytes based on precision (following Iceberg's rules)
+        int requiredBytes = getIcebergDecimalBytes(precision);
```
luoyuxia: Where can i get the logic in iceberg repo?

fluss-common/src/test/java/com/alibaba/fluss/row/encode/iceberg/IcebergKeyEncoderTest.java
```diff
@@ -0,0 +1,144 @@
...
+        DataType[] allDataTypes = allRowType.getChildren().toArray(new DataType[0]);
+
+        IndexedRow indexedRow = genFlussRowForAllTypes(allDataTypes);
+        List<String> encodedKeys = allRowType.getFieldNames();
+        IcebergKeyEncoder icebergKeyEncoder = new IcebergKeyEncoder(allRowType, encodedKeys);
+
+        // encode with Fluss own implementation for Iceberg
+        byte[] encodedKey = icebergKeyEncoder.encodeKey(indexedRow);
+
+        // encode with Iceberg implementation using native Iceberg classes
```
luoyuxia: This is still not using native Iceberg classes to encode the keys. It still use our own IcebergBinaryRowWriter, so that we can't ensure our encoding is align with iceberg.

luoyuxia (commented):
@MehulBatra Thanks for the pr. I left minor comment. PTAL

fluss-common/src/main/java/com/alibaba/fluss/row/encode/iceberg/IcebergBinaryRowWriter.java
MehulBatra: sure, will remove it.

fluss-common/src/main/java/com/alibaba/fluss/row/encode/iceberg/IcebergBinaryRowWriter.java
MehulBatra: No, I kept it deliberaty just in case later if i might to copy any iceberg logic, but will remove it.

fluss-common/src/test/java/com/alibaba/fluss/row/encode/iceberg/IcebergKeyEncoderTest.java
MehulBatra: Sure I will see how can I encode the rows using iceberg library itself.

fluss-common/src/main/java/com/alibaba/fluss/row/encode/iceberg/IcebergBinaryRowWriter.java
MehulBatra: Sure!

MehulBatra:
Hi @luoyuxia I tried to take inspiration from
https://github.com/apache/iceberg/blob/main/api/src/main/java/org/apache/iceberg/types/Conversions.java
for encoding 
https://github.com/apache/iceberg/blob/main/api/src/main/java/org/apache/iceberg/types/Types.java
for types

fluss-common/src/main/java/com/alibaba/fluss/row/encode/iceberg/IcebergBinaryRowWriter.java
MehulBatra: https://github.com/apache/iceberg/blob/main/api/src/main/java/org/apache/iceberg/types/Conversions.java

MehulBatra:
Currently our CI runs for java8, when I am using iceberg 1.9.1 in my code as it only supports java 11 , it is failing the CI check, trying to downgrade the version to 1.4.3 to support java 8, my assumption it would pass post that.
Stacktrace: 
`Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.8.0:testCompile (default-testCompile) on project fluss-common: Compilation failure
Error:  /home/fluss/actions-runner/_work/fluss/fluss/fluss-common/src/test/java/com/alibaba/fluss/row/encode/iceberg/IcebergKeyEncoderTest.java:[30,32] cannot access org.apache.iceberg.types.Conversions
Error:    bad class file: /home/fluss/.m2/repository/org/apache/iceberg/iceberg-api/1.9.1/iceberg-api-1.9.1.jar(org/apache/iceberg/types/Conversions.class)
Error:      class file has wrong version 55.0, should be 52.0
Error:      Please remove or make sure it appears in the correct subdirectory of the classpath.`

luoyuxia:
> Hi @luoyuxia I tried to take inspiration from https://github.com/apache/iceberg/blob/main/api/src/main/java/org/apache/iceberg/types/Conversions.java for encoding https://github.com/apache/iceberg/blob/main/api/src/main/java/org/apache/iceberg/types/Types.java for types

Thanks for the explaination.  I think it make senses to be inspired from iceberg's encoding. But keep in mind we use it for different purpose:
- Iceebrg use it to store the binary values of field statistic, right?
- We use it for distribute records, in Fluss's implementation, it'll first encoding the bucket columns, and then use the encoded value to do bucket partition transfrom which should align with iceberg. But iceberg use literal value to do [bucket partition transfrom](https://github.com/apache/iceberg/blob/6bbbbf6f6f9ba6dc8cb273f8dcd00be1b5dfc399/api/src/main/java/org/apache/iceberg/transforms/Bucket.java#L98). So it means we don't really care about how to encode the bucket columns into binary value. We only need to get the literal value from the binary value(this will be called decoding binary value).  That means we can use any encoding way. But It's fine for me to following how iceberg encode the values.

fluss-common/src/main/java/com/alibaba/fluss/row/encode/iceberg/IcebergBinaryRowWriter.java
```diff
@@ -0,0 +1,326 @@
...
+import java.nio.ByteBuffer;
+import java.nio.ByteOrder;
+import java.util.Arrays;
+
+import static com.alibaba.fluss.types.DataTypeChecks.getPrecision;
+
+/**
+ * A writer to encode Fluss's {@link InternalRow} using Iceberg's binary encoding format.
+ *
+ * <p>This implementation follows Iceberg's binary encoding specification for partition and bucket
```
luoyuxia: IIUC, this is not for binary encoding specification for partition We can just remove this line of comment.

fluss-common/src/main/java/com/alibaba/fluss/row/encode/iceberg/IcebergBinaryRowWriter.java
```diff
@@ -0,0 +1,326 @@
...
+ * keys. Reference: https://iceberg.apache.org/spec/#partition-transforms
+ *
+ * <p>The encoding logic is based on Iceberg's Conversions.toByteBuffer() implementation:
+ * https://github.com/apache/iceberg/blob/main/api/src/main/java/org/apache/iceberg/types/Conversions.java
+ *
+ * <p>Key encoding principles from Iceberg's Conversions class:
+ *
+ * <ul>
+ *   <li>All numeric types (int, long, float, double, timestamps) use LITTLE-ENDIAN byte order
+ *   <li>Decimal and UUID types use BIG-ENDIAN byte order
```
luoyuxia: We have no UUID, so remove this `UUID types` from the comments.

fluss-common/src/main/java/com/alibaba/fluss/row/encode/iceberg/IcebergBinaryRowWriter.java
```diff
@@ -0,0 +1,326 @@
...
+ *   <li>NO length prefix for any type - the buffer size determines the length
```
luoyuxia: I think we will still need length prefix for varlength type, such as string. That's different from iceberg. If no length prefix, we can't figure out the bound of string values if we store two or more string values.

fluss-common/src/main/java/com/alibaba/fluss/row/encode/iceberg/IcebergBinaryRowWriter.java
```diff
@@ -0,0 +1,326 @@
...
+ *   <li>Strings are encoded as UTF-8 bytes without length prefix
```
luoyuxia: String should encoded with length prefix

fluss-common/src/main/java/com/alibaba/fluss/row/encode/iceberg/IcebergBinaryRowWriter.java
```diff
@@ -0,0 +1,326 @@
...
+        ensureCapacity(8);
+        UnsafeUtils.putDouble(buffer, cursor, value);
+        cursor += 8;
+    }
+
+    public void writeString(BinaryString value) {
+        // Based on Iceberg's Conversions.toByteBuffer() for STRING type:
+        // Strings are encoded as UTF-8 bytes without length prefix
+        // Reference:https://github.com/apache/iceberg/blob/main/api/src/main/java/org/apache/iceberg/types/Conversions.java
+        byte[] bytes = BinaryString.encodeUTF8(value.toString());
```
luoyuxia, line 131: We need to put the length of the bytes.

fluss-common/src/main/java/com/alibaba/fluss/row/encode/iceberg/IcebergBinaryRowWriter.java
```diff
@@ -0,0 +1,326 @@
...
+        ensureCapacity(bytes.length);
+        segment.put(cursor, bytes, 0, bytes.length);
+        cursor += bytes.length;
+    }
+
+    void writeBytes(byte[] bytes) {
+        // Based on Iceberg's Conversions.toByteBuffer() for BINARY type:
+        // Binary data is stored directly without length prefix
+        ensureCapacity(bytes.length);
```
luoyuxia: We also need to put the length of the bytes.

luoyuxia:
> Currently our CI runs for java8, when I am using iceberg 1.9.1 in my code as it only supports java 11 , it is failing the CI check, trying to downgrade the version to 1.4.3 to support java 8, my assumption it would pass post that. Stacktrace: `Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.8.0:testCompile (default-testCompile) on project fluss-common: Compilation failure Error: /home/fluss/actions-runner/_work/fluss/fluss/fluss-common/src/test/java/com/alibaba/fluss/row/encode/iceberg/IcebergKeyEncoderTest.java:[30,32] cannot access org.apache.iceberg.types.Conversions Error: bad class file: /home/fluss/.m2/repository/org/apache/iceberg/iceberg-api/1.9.1/iceberg-api-1.9.1.jar(org/apache/iceberg/types/Conversions.class) Error: class file has wrong version 55.0, should be 52.0 Error: Please remove or make sure it appears in the correct subdirectory of the classpath.`

If that's the case, we can downgrade the version to 1.4.3 firstly, and left todo when #1195 is resolved

fluss-common/src/main/java/com/alibaba/fluss/row/encode/iceberg/IcebergKeyEncoder.java
```diff
@@ -0,0 +1,84 @@
...
+                    icebergBinaryRowWriter, fieldGetters[i].getFieldOrNull(row));
+        }
+        return icebergBinaryRowWriter.toBytes();
+    }
+
+    /**
+     * Returns the encoded key as a ByteBuffer for Iceberg-compatible operations. This is useful
+     * when interfacing with Iceberg's hash functions which expect ByteBuffer.
+     */
+    public ByteBuffer encodeKeyAsBuffer(InternalRow row) {
```
luoyuxia: We don't need this method.

fluss-common/src/main/java/com/alibaba/fluss/row/encode/iceberg/IcebergBinaryRowWriter.java
```diff
@@ -0,0 +1,326 @@
...
+        if (precision <= 9) {
+            return 4; // Can fit in 4 bytes
+        } else if (precision <= 18) {
+            return 8; // Can fit in 8 bytes
+        } else {
+            return 16; // Requires 16 bytes
+        }
+    }
+
+    public ByteBuffer toByteBuffer() {
```
luoyuxia: we don't need this method

fluss-common/src/main/java/com/alibaba/fluss/row/encode/iceberg/IcebergBinaryRowWriter.java
```diff
@@ -0,0 +1,326 @@
...
+        // Iceberg stores timestamps as microseconds since epoch
+        // Reference:
+        // https://github.com/apache/iceberg/blob/main/api/src/main/java/org/apache/iceberg/types/Conversions.java
+        long micros = value.getMillisecond() * 1000L + (value.getNanoOfMillisecond() / 1000L);
+        writeLong(micros);
+    }
+
+    public void writeTimestampLtz(TimestampLtz value, int precision) {
+        // Iceberg stores timestamptz as microseconds since epoch in UTC
+        // Reference:
```
luoyuxia: We don't need the Reference  comment for every method since we have commented it in header

fluss-common/src/main/java/com/alibaba/fluss/row/encode/iceberg/IcebergBinaryRowWriter.java
```diff
@@ -0,0 +1,326 @@
...
+        // Reference:https://github.com/apache/iceberg/blob/main/api/src/main/java/org/apache/iceberg/types/Conversions.java
+        byte[] unscaledBytes = value.toUnscaledBytes();
+
+        // For Iceberg, decimals are stored with fixed byte lengths based on precision
+        int requiredBytes = getIcebergDecimalBytes(precision);
+
+        byte[] icebergBytes = new byte[requiredBytes];
+
+        // Convert to big-endian format with proper padding
+        if (unscaledBytes.length <= requiredBytes) {
```
luoyuxia: Where can i find the complex logic in iceberg. IIUC, iceberg just use `ByteBuffer.wrap(((BigDecimal) value).unscaledValue().toByteArray());`
Can the peice of code simplified?

fluss-common/src/main/java/com/alibaba/fluss/row/encode/iceberg/IcebergKeyEncoder.java
```diff
@@ -0,0 +1,84 @@
...
+
+    private final IcebergBinaryRowWriter icebergBinaryRowWriter;
+
+    public IcebergKeyEncoder(RowType rowType, List<String> keys) {
+
+        // Validate single key field requirement as per FIP
+        checkArgument(
+                keys.size() == 1,
+                "Key fields must have exactly one field for iceberg format, but got: %s",
+                keys.size());
```
luoyuxia:
```suggestion
                keys);
```
it will be better to print the keys

fluss-common/src/main/java/com/alibaba/fluss/row/encode/iceberg/IcebergBinaryRowWriter.java
```diff
@@ -0,0 +1,326 @@
...
+ *   <li>Decimal and UUID types use BIG-ENDIAN byte order
+ *   <li>NO length prefix for any type - the buffer size determines the length
+ *   <li>Strings are encoded as UTF-8 bytes without length prefix
+ *   <li>Timestamps are stored as long values (microseconds since epoch)
+ * </ul>
+ *
+ * <p>Note: This implementation uses Fluss's MemorySegment instead of ByteBuffer for performance,
+ * but maintains byte-level compatibility with Iceberg's encoding.
+ */
+class IcebergBinaryRowWriter {
```
luoyuxia, line 51: Here is iceberg code for bucket transform, so we only need to support encode the what icberge supports for bucket, for other datatypes, we can just throw unsupport exception to make the code simple
```
switch (type.typeId()) {
      case DATE:
      case INTEGER:
        return (B) new BucketInteger(numBuckets);
      case TIME:
      case TIMESTAMP:
      case LONG:
        return (B) new BucketLong(numBuckets);
      case DECIMAL:
        return (B) new BucketDecimal(numBuckets);
      case STRING:
        return (B) new BucketString(numBuckets);
      case FIXED:
      case BINARY:
        return (B) new BucketByteBuffer(numBuckets);
      case TIMESTAMP_NANO:
        return (B) new BucketTimestampNano(numBuckets);
      case UUID:
        return (B) new BucketUUID(numBuckets);
      default:
        throw new IllegalArgumentException("Cannot bucket by type: " + type);
    }
```

fluss-common/src/test/java/com/alibaba/fluss/row/encode/iceberg/IcebergKeyEncoderTest.java
```diff
@@ -0,0 +1,431 @@
...
+    private IndexedRow createRowWithTime(int millis) {
+        DataType[] dataTypes = {DataTypes.TIME()};
+        IndexedRowWriter writer = new IndexedRowWriter(dataTypes);
+        writer.writeInt(millis); // Fluss stores TIME as int (milliseconds)
+        IndexedRow row = new IndexedRow(dataTypes);
+        row.pointTo(writer.segment(), 0, writer.position());
+        return row;
+    }
+
+    private IndexedRow createRowWithBytes(byte[] value) {
```
luoyuxia: nit:
```
private IndexedRow createRowWithBytes(byte[] value) throws Exception {
        DataType[] dataTypes = {DataTypes.BYTES()};
        try (IndexedRowWriter writer = new IndexedRowWriter(dataTypes)) {
            writer.writeBytes(value);
            IndexedRow row = new IndexedRow(dataTypes);
            row.pointTo(writer.segment(), 0, writer.position());
            return row;
        }
    }
```
Close `IndexedRowWriter` to resolve the warning of IDE

pom.xml
```diff
@@ -121,13 +122,13 @@
         <spotless.license.header>
             /*
             * Licensed to the Apache Software Foundation (ASF) under one or more
-            * contributor license agreements.  See the NOTICE file distributed with
+            * contributor license agreements. See the NOTICE file distributed with
             * this work for additional information regarding copyright ownership.
             * The ASF licenses this file to You under the Apache License, Version 2.0
             * (the "License"); you may not use this file except in compliance with
-            * the License.  You may obtain a copy of the License at
+            * the License. You may obtain a copy of the License at
```
luoyuxia: Why change this? Other files are still using two blanks?

pom.xml
```diff
@@ -121,13 +122,13 @@
...
             *
-            *    http://www.apache.org/licenses/LICENSE-2.0
+            * http://www.apache.org/licenses/LICENSE-2.0
```
luoyuxia: dito

pom.xml
```diff
@@ -550,7 +551,7 @@
                             under the License.
                         -->
                         <license implementation="org.apache.rat.analysis.license.SimplePatternBasedLicense">
-                            <licenseFamilyCategory>AL2 </licenseFamilyCategory>
+                            <licenseFamilyCategory>AL2</licenseFamilyCategory>
```
luoyuxia, line 555: dito

pom.xml
```diff
@@ -832,10 +833,10 @@
                     <transformers combine.children="append">
                         <!-- The service transformer is needed to merge META-INF/services files -->
                         <transformer
-                            implementation="org.apache.maven.plugins.shade.resource.ServicesResourceTransformer"/>
+                                implementation="org.apache.maven.plugins.shade.resource.ServicesResourceTransformer"/>
```
luoyuxia: dito

pom.xml
```diff
@@ -832,10 +833,10 @@
...
                         <!-- The ApacheNoticeResourceTransformer collects and aggregates NOTICE files -->
                         <transformer
-                            implementation="org.apache.maven.plugins.shade.resource.ApacheNoticeResourceTransformer">
+                                implementation="org.apache.maven.plugins.shade.resource.ApacheNoticeResourceTransformer">
```
luoyuxia: dito

pom.xml
MehulBatra: my ide formatter doing something unsual will fix it.

fluss-common/src/main/java/com/alibaba/fluss/row/encode/iceberg/IcebergBinaryRowWriter.java
MehulBatra: While reffering to doc, it was little complicated, but where as in the conversion code it's easy, I have translated it to a simpler way now.

fluss-common/src/main/java/com/alibaba/fluss/row/encode/iceberg/IcebergBinaryRowWriter.java
MehulBatra: removed

fluss-common/src/main/java/com/alibaba/fluss/row/encode/iceberg/IcebergKeyEncoder.java
MehulBatra: removed it, i just wrote it  as an optional method to match the iceberg encode return type, but we don't need it

fluss-common/src/main/java/com/alibaba/fluss/row/encode/iceberg/IcebergBinaryRowWriter.java
MehulBatra: Bit confusing in support types comment for iceberg buckettransform you mentioned to provide support for UUID, here it's said remove it, so to be on the safer side i am removing UUID as we don't have support for that datatype in fluss

fluss-common/src/main/java/com/alibaba/fluss/row/encode/iceberg/IcebergBinaryRowWriter.java
MehulBatra: Whitelisted, except the UUID as we don't have that datatype in fluss and also iceberg 1.4.3 doesn't support nanoseconds, as it came in table v3, which was released in later version. will add a note for nanoseconds to support post #1195

fluss-common/src/main/java/com/alibaba/fluss/row/encode/iceberg/IcebergBinaryRowWriter.java
MehulBatra: make sense

fluss-common/src/main/java/com/alibaba/fluss/row/encode/iceberg/IcebergBinaryRowWriter.java
MehulBatra: ok, i will try to encode the lenght alongside

MehulBatra: can you explain a bit why we need length? why it is so important?

fluss-common/src/main/java/com/alibaba/fluss/row/encode/iceberg/IcebergBinaryRowWriter.java
MehulBatra: added length, in this also, but this is adding lenght prefix in writedecimal also.

MehulBatra: added comment in the writeDecimal part:
`
  public void writeDecimal(Decimal value, int precision) {
    byte[] unscaled = value.toUnscaledBytes();
    writeBytes(unscaled); // Adds 4-byte length prefix before the actual bytes
  }
`

pom.xml
MehulBatra: Somehow the formatter is not letting me change it back, it is getting autofixed

MehulBatra:
> > Hi @luoyuxia I tried to take inspiration from https://github.com/apache/iceberg/blob/main/api/src/main/java/org/apache/iceberg/types/Conversions.java for encoding https://github.com/apache/iceberg/blob/main/api/src/main/java/org/apache/iceberg/types/Types.java for types
> 
> Thanks for the explaination. I think it make senses to be inspired from iceberg's encoding. But keep in mind we use it for different purpose:
> 
> * Iceebrg use it to store the binary values of field statistic, right?
> * We use it for distribute records, in Fluss's implementation, it'll first encoding the bucket columns, and then use the encoded value to do bucket partition transfrom which should align with iceberg. But iceberg use literal value to do [bucket partition transfrom](https://github.com/apache/iceberg/blob/6bbbbf6f6f9ba6dc8cb273f8dcd00be1b5dfc399/api/src/main/java/org/apache/iceberg/transforms/Bucket.java#L98). So it means we don't really care about how to encode the bucket columns into binary value. We only need to get the literal value from the binary value(this will be called decoding binary value).  That means we can use any encoding way. But It's fine for me to following how iceberg encode the values.

-> Agreed Iceberg does use these to store partition stats and min/max for filtering.
-> The end goal is to encode the values in iceberg compatibal format, can you add more on this?
We only need to get the literal value from the binary value(this will be called decoding binary value).  That means we can use any encoding way. But It's fine for me to following how iceberg encode the values.

fluss-common/src/main/java/com/alibaba/fluss/row/encode/iceberg/IcebergBinaryRowWriter.java
luoyuxia: If we have two string columns, if not encode length of each string column, how can we know the bound of the bytes of the string column?

luoyuxia:
> -> The end goal is to encode the values in iceberg compatibal format, can you add more on this?

We don't need to encode the values in iceberg compatibal format. We just need to make sure the bucket partition transform align with iceberg. Iceberg just bucket via literal value, without the encoding process which is different from paimon. Fluss need to encode for we need to align the process that fluss write rows, which will 
- step1: first encoding row into bytes
- step2: hash(distribute) the **bytes** into a bucket. 

So, for iceberg, although we encode it, we still need to decode the bytes into the literal value(a int value, long value, etc), and use the iceberg bucket strategy. See `BucketInteger`, `BucketLong` in iceberg repo

fluss-common/src/main/java/com/alibaba/fluss/row/encode/iceberg/IcebergBinaryRowWriter.java
```diff
@@ -0,0 +1,224 @@
...
+    }
+
+    public void setNullAt(int pos) {
+        // For Iceberg key encoding, null values should not occur
+        // This is validated at the encoder level
+        throw new UnsupportedOperationException(
+                "Null values are not supported in Iceberg key encoding");
+    }
+
+    // todo: Revisit to add support as per Iceberg 1.9.1 post #1195 merge for Java 11 support
```
luoyuxia: Why put the todo in here? I suggest to add the todo in https://github.com/apache/fluss/pull/1350/files#diff-6051cc48685fc3e562c38ddbede5594cadce48e06db5b359c57e9c37aefb6ca6L36

fluss-common/src/test/java/com/alibaba/fluss/row/encode/iceberg/IcebergKeyEncoderTest.java
```diff
@@ -0,0 +1,395 @@
...
+
+/**
+ * Unit tests for {@link IcebergKeyEncoder} to verify the encoding matches Iceberg's format.
+ *
+ * <p>This test uses Iceberg's actual Conversions class to ensure our encoding is byte-for-byte
+ * compatible with Iceberg's implementation.
+ */
+class IcebergKeyEncoderTest {
+
+    // todo: Revisit to add support as per Iceberg 1.9.1 post #1195 merge for Java 11 support
```
luoyuxia: dito: remove the todo in here.

pom.xml
```diff
@@ -98,6 +98,7 @@
         <curator.version>5.4.0</curator.version>
         <netty.version>4.1.104</netty.version>
         <arrow.version>15.0.0</arrow.version>
+        <iceberg.version>1.4.3</iceberg.version>
```
luoyuxia: remove this, we dont need it in root `pom.xml`

fluss-common/src/main/java/com/alibaba/fluss/row/encode/iceberg/IcebergBinaryRowWriter.java
luoyuxia: Also, please create a issue to track it.

MehulBatra: https://github.com/apache/fluss/issues/1379

fluss-common/src/main/java/com/alibaba/fluss/row/encode/iceberg/IcebergBinaryRowWriter.java
MehulBatra: I missed it, I could imagine now it make sense if we have two columns and want to bucket them using them as a encoded string, it would look like this [0,0,0,3, C,A,T, 0,0,0,3, D,O,G] |len=3| |CAT|(utf-8 data) |len=3| |DOG|utf-8 data)

pom.xml
MehulBatra:
```
 <!-- paimon bundle, only for test purpose -->
    <dependency>
        <groupId>org.apache.paimon</groupId>
        <artifactId>paimon-bundle</artifactId>
        <version>${paimon.version}</version>
        <scope>test</scope>
    </dependency>
    <dependency>
        <groupId>org.apache.iceberg</groupId>
        <artifactId>iceberg-api</artifactId>
        <version>${iceberg.version}</version>
        <scope>test</scope>
    </dependency>
</dependencies>
```

Paimon and other packages also uses root pom.xml for version, I tried to match and did the same for iceberg-api in fluss-common pom.xml

luoyuxia: Sorry for missing that test need it. I'd like to retract this comment

MehulBatra:
Thank you @luoyuxia for the review & feedbacks!

PR #1355, opened 2025-07-17
[kv] support export key metrics for rocksdb

### LINKED ISSUE #1320: [KV] Introduce full metrics export for RocksDB in KV Tablet
### Search before asking

### Motivation

Currently, we only have logs but lack metrics to monitor the performance of RocksDB instances, which makes it difficult to tune the performance of KV load.

### Solution

Export the core metrics of RocksDB.

### Anything else?

_No response_

### Willingness to contribute

platinumhamburg:
### Purpose

Linked issue: close #1320


fluss-common/src/main/java/org/apache/fluss/config/ConfigOptions.java
```diff
@@ -1410,6 +1410,12 @@ public class ConfigOptions {
                             "The max size of the consumed memory for RocksDB batch write, "
                                     + "will flush just based on item count if this config set to 0.");
 
+    public static final ConfigOption<Boolean> KV_STATISTICS_ENABLED =
+            key("kv.rocksdb.statistics.enabled")
```
wuchong: Rename to `kv.rocksdb.metrics.enabled`? I think it also includes some metrics beyond statistics, like cache metrics.

fluss-server/src/main/java/com/alibaba/fluss/server/kv/rocksdb/RocksDBMetricsCollector.java
```diff
@@ -0,0 +1,527 @@
...
+        this.registered = true;
+
+        LOG.info(
+                "RocksDB metrics collector started for table {} bucket {}",
+                tablePath,
+                tableBucket.getBucket());
+    }
+
+    private void registerMetrics() {
+        // Cache metrics - 统一使用缓存变量
```
wuchong: Use English.

fluss-server/src/main/java/com/alibaba/fluss/server/kv/rocksdb/RocksDBMetricsCollector.java
```diff
@@ -0,0 +1,527 @@
...
+                MetricNames.ROCKSDB_COMPACTION_BYTES_WRITTEN,
+                (Gauge<Long>) () -> compactionBytesWritten);
+        bucketMetricGroup.gauge(
+                MetricNames.ROCKSDB_COMPACTION_CPU_TIME_MICROS,
+                (Gauge<Long>) () -> compactionCpuTimeMicros);
+        bucketMetricGroup.gauge(MetricNames.ROCKSDB_FLUSH_COUNT, (Gauge<Long>) () -> flushCount);
+        bucketMetricGroup.gauge(
+                MetricNames.ROCKSDB_FLUSH_BYTES_WRITTEN, (Gauge<Long>) () -> flushBytesWritten);
+
+        // Memory metrics - 统一使用缓存变量
```
wuchong: Use English.

fluss-server/src/main/java/com/alibaba/fluss/server/kv/rocksdb/RocksDBMetricsCollector.java
```diff
@@ -0,0 +1,527 @@
...
+                (Gauge<Long>) () -> tableReadersMemoryUsage);
+        bucketMetricGroup.gauge(
+                MetricNames.ROCKSDB_TOTAL_MEMORY_USAGE, (Gauge<Long>) () -> totalMemoryUsage);
+        bucketMetricGroup.gauge(
+                MetricNames.ROCKSDB_ACTIVE_MEMTABLE_SIZE, (Gauge<Long>) () -> activeMemtableSize);
+        bucketMetricGroup.gauge(
+                MetricNames.ROCKSDB_UNFLUSHED_MEMTABLE_SIZE,
+                (Gauge<Long>) () -> unflushedMemtableSize);
+
+        // BlockCache usage指标 - 统一使用缓存变量
```
wuchong: Use English.

fluss-server/src/main/java/org/apache/fluss/server/kv/KvTablet.java
```diff
@@ -259,6 +263,36 @@ public void registerMetrics(BucketMetricGroup bucketMetricGroup) {
         metricGroup.meter(
                 MetricNames.KV_PRE_WRITE_BUFFER_TRUNCATE_AS_ERROR_RATE,
                 new MeterView(kvPreWriteBuffer.getTruncateAsErrorCount()));
+
+        // Initialize RocksDB metrics reporter
+        inWriteLock(
+                kvLock,
```
wuchong: I think we don't need to initialize `rocksDBMetricsCollector` in lock, because there is no concurrency problem here. Besides, I think it's better to move the `rocksDBMetricsCollector` initialization to the constructor to simplify the logic and also make the `rocksDBMetricsCollector` to be final.

fluss-server/src/main/java/com/alibaba/fluss/server/kv/rocksdb/RocksDBMetricsCollector.java
```diff
@@ -0,0 +1,527 @@
...
+            filterBlockCacheHitCount = getTickerValue(TickerType.BLOCK_CACHE_FILTER_HIT);
+            filterBlockCacheMissCount = getTickerValue(TickerType.BLOCK_CACHE_FILTER_MISS);
+            dataBlockCacheHitCount = getTickerValue(TickerType.BLOCK_CACHE_DATA_HIT);
+            dataBlockCacheMissCount = getTickerValue(TickerType.BLOCK_CACHE_DATA_MISS);
+
+            // Update compaction metrics
+            compactionCount =
+                    getTickerValue(TickerType.COMPACT_READ_BYTES) > 0
+                            ? 1
+                            : 0; // Simplified since exact count may not be available
```
wuchong: It's confusing to report a count that only has values 0 and 1. Is it necessary to report this metric? Can `compactionBytesRead` be satisfied?

fluss-server/src/main/java/com/alibaba/fluss/server/kv/rocksdb/RocksDBMetricsCollector.java
```diff
@@ -0,0 +1,527 @@
...
+            compactionBytesRead = getTickerValue(TickerType.COMPACT_READ_BYTES);
+            compactionBytesWritten = getTickerValue(TickerType.COMPACT_WRITE_BYTES);
+            compactionCpuTimeMicros = getHistogramValue(HistogramType.COMPACTION_TIME);
+            flushCount = getTickerValue(TickerType.FLUSH_WRITE_BYTES) > 0 ? 1 : 0;
```
wuchong: ditto. Can `flushBytesWritten` be satisfied?

fluss-server/src/test/java/com/alibaba/fluss/server/kv/rocksdb/RocksDBMetricsCollectorTest.java
```diff
@@ -0,0 +1,248 @@
...
+/** Tests for {@link RocksDBMetricsCollector}. */
+class RocksDBMetricsCollectorTest {
+
+    @TempDir Path tempDir;
+    private RocksDBResourceContainer resourceContainer;
+    private RocksDBMetricsCollector collector;
+    private Statistics mockStatistics;
+    private AutoCloseable mocksCloseable;
+    private BucketMetricGroup bucketMetricGroup;
+    private AutoCloseable mockMemUtilAutocloseable;
```
wuchong: never used, can remove

fluss-server/src/main/java/com/alibaba/fluss/server/kv/rocksdb/RocksDBMetricsManager.java
```diff
@@ -0,0 +1,223 @@
...
+
+    /** Singleton instance. */
+    private static volatile RocksDBMetricsManager instance;
+
+    /** Update interval for metrics collection (in seconds). */
+    private static final long METRICS_UPDATE_INTERVAL_SECONDS = 5;
+
+    /** Number of threads for metrics collection. */
+    private static final int METRICS_COLLECTION_THREADS =
+            Math.max(1, Runtime.getRuntime().availableProcessors() / 4);
```
wuchong: IMO, single thread for rocksdb metrics updating is enough. A metric reporter shouldn't consume so many threads, otherwise, there must be too many metrics.

fluss-server/src/main/java/com/alibaba/fluss/server/kv/rocksdb/RocksDBMetricsManager.java
```diff
@@ -0,0 +1,223 @@
...
+                        createCollectorKey(collector),
+                        e);
+            }
+        }
+    }
+
+    /** Create a unique key for a collector. */
+    private String createCollectorKey(RocksDBMetricsCollector collector) {
+        return String.format(
+                "%s-%s-%s-%s",
```
wuchong: `-` is a valid character in database name and table name, therefore, it's better to use `/` as separator, to avoid the key conflicts.

fluss-server/src/test/java/com/alibaba/fluss/server/kv/rocksdb/RocksDBMetricsManagerTest.java
```diff
@@ -0,0 +1,423 @@
...
+                        partition,
+                        resourceContainer));
+    }
+
+    private void resetSingleton() {
+        try {
+            Field instanceField = RocksDBMetricsManager.class.getDeclaredField("instance");
+            instanceField.setAccessible(true);
+            instanceField.set(null, null);
+        } catch (Exception e) {
```
wuchong: Usually, reflection is highly not recommended, unless it is absolutely necessary. Here we can declare the `instance` variable as `protected` to make it can be accessed in the `RocksDBMetricsManagerTest` but not exposed publicly.

fluss-server/src/test/java/com/alibaba/fluss/server/kv/rocksdb/RocksDBMetricsManagerTest.java
```diff
@@ -0,0 +1,423 @@
...
+        for (int i = 0; i < collectorCount; i++) {
+            RocksDBMetricsCollector collector = createMockCollector("test_db", "test_table", i);
+            RocksDBMetricsCollector spyCollector = spy(collector);
+            manager.registerCollector(spyCollector);
+            spyCollectors.add(spyCollector);
+            collectors.add(collector);
+        }
+
+        // Wait for metrics collection to occur
+        Thread.sleep(6000); // Wait for at least one collection cycle
```
wuchong: Do not use `Thread.sleep` for testing. This kills time and is unstable when the CPU load is high on testing machines. 

The suggested way for testing scheduler is using `ManuallyTriggeredScheduledExecutorService` in tests. See `MetricViewUpdater` and `MetricViewUpdaterTest` as an example. The key idea is adding a protected constructor for `RocksDBMetricsManager` with `ScheduledExecutorService metricsCollectionExecutor` as parameter, and create the `RocksDBMetricsManager` with the `ManuallyTriggeredScheduledExecutorService` in the test method. And manually trigger the scheduler instead of sleeping.

fluss-server/src/test/java/com/alibaba/fluss/server/kv/rocksdb/RocksDBMetricsManagerTest.java
```diff
@@ -0,0 +1,423 @@
...
+
+        manager.registerCollector(faultyCollector);
+        manager.registerCollector(spyNormalCollector);
+
+        collectors.add(faultyCollector);
+        collectors.add(normalCollector);
+
+        // Wait for metrics collection
+        try {
+            Thread.sleep(6000);
```
wuchong: ditto

fluss-server/src/test/java/com/alibaba/fluss/server/kv/rocksdb/RocksDBMetricsManagerTest.java
```diff
@@ -0,0 +1,423 @@
...
+
+    @Test
+    void testEmptyCollectorsDoesNotCauseIssues() throws Exception {
+        RocksDBMetricsManager manager = RocksDBMetricsManager.getInstance();
+
+        // Start with no collectors
+        assertThat(manager.getCollectorCount()).isEqualTo(0);
+
+        // Wait a bit to ensure no exceptions are thrown
+        Thread.sleep(1000);
```
wuchong: ditto

fluss-server/src/main/java/com/alibaba/fluss/server/kv/rocksdb/RocksDBMetricsManager.java
```diff
@@ -0,0 +1,223 @@
...
+    private void gatherMetrics() {
+        if (collectors.isEmpty()) {
+            return;
+        }
+
+        List<RocksDBMetricsCollector> collectorList = new ArrayList<>(collectors.values());
+
+        // Process collectors in batches to avoid overwhelming the system
+        for (int i = 0; i < collectorList.size(); i += BATCH_SIZE) {
+            int endIndex = Math.min(i + BATCH_SIZE, collectorList.size());
```
wuchong: If the schduler is single-thread, then there is no need to submit batching tasks.

fluss-server/src/test/java/com/alibaba/fluss/server/kv/rocksdb/RocksDBMetricsManagerTest.java
```diff
@@ -0,0 +1,423 @@
...
+                            latch.countDown();
+                        }
+                    });
+        }
+
+        latch.await(10, TimeUnit.SECONDS);
+        executor.shutdown();
+
+        // Should handle concurrent access gracefully
+        assertThat(manager.getCollectorCount()).isGreaterThanOrEqualTo(0);
```
wuchong: I think we can assert the count value `50` here? It should be deterministic, right?

fluss-server/src/test/java/com/alibaba/fluss/server/kv/rocksdb/RocksDBMetricsManagerTest.java
```diff
@@ -0,0 +1,423 @@
...
+        RocksDBMetricsCollector collector2 = createMockCollector("db1", "table1", 1);
+        RocksDBMetricsCollector collector3 = createMockCollector("db2", "table1", 0);
+
+        // Different buckets should have different keys
+        assertThat(collector1.getTableBucket().getBucket())
+                .isNotEqualTo(collector2.getTableBucket().getBucket());
+
+        // Different databases should have different keys
+        assertThat(collector1.getTablePath().getDatabaseName())
+                .isNotEqualTo(collector3.getTablePath().getDatabaseName());
```
wuchong: I don't see it asserts anything. If we want to assert the reporter key is different, I think we need to assert `manager.getCollectorCount() = 3` here?

fluss-server/src/test/java/com/alibaba/fluss/server/kv/rocksdb/RocksDBMetricsManagerTest.java
```diff
@@ -0,0 +1,423 @@
...
+        collectors.add(collector2);
+        collectors.add(collector3);
+    }
+
+    @Test
+    void testPartitionedTableCollector() {
+        RocksDBMetricsCollector collector =
+                createMockCollectorWithPartition("test_db", "test_table", 0, "partition1");
+
+        assertThat(collector.getPartitionName()).isEqualTo("partition1");
```
wuchong: I don't see it asserts anything. If we want to assert the reporter key is different, I think we need to create another collector with different partition name and  assert `manager.getCollectorCount() = 2` here?

fluss-server/src/test/java/com/alibaba/fluss/server/kv/rocksdb/RocksDBMetricsManagerTest.java
```diff
@@ -0,0 +1,423 @@
...
+        // Register a collector to start the manager
+        RocksDBMetricsCollector collector = createMockCollector("test_db", "test_table", 0);
+        manager.registerCollector(collector);
+        collectors.add(collector);
+
+        assertThat(manager.isRunning()).isTrue();
+
+        // Get the executor service via reflection to verify it's cleaned up
+        Field executorField =
+                RocksDBMetricsManager.class.getDeclaredField("metricsCollectionExecutor");
```
wuchong: ditto. use `protected` variables instead of Java reflection.

fluss-server/src/test/java/org/apache/fluss/server/kv/rocksdb/RocksDBResourceContainerTest.java
```diff
@@ -187,8 +193,237 @@ void testConfigurationOptionsFromConfig() throws Exception {
...
+        }
+
+        // All statistics should be closed
+        for (Statistics stats : statisticsList) {
+            assertThat(stats.isOwningHandle()).isFalse();
+        }
+    }
+
+    @Test
+    void testStatisticsWithNullInstancePath() throws Exception {
```
wuchong: I don't see any values of this test, what's the use case with a null instance path?

fluss-server/src/test/java/org/apache/fluss/server/kv/rocksdb/RocksDBResourceContainerTest.java
```diff
@@ -187,8 +193,237 @@ void testConfigurationOptionsFromConfig() throws Exception {
...
+
+        try (RocksDBResourceContainer container = new RocksDBResourceContainer(config, null)) {
+            Statistics statistics = container.getStatistics();
+            assertThat(statistics).isNotNull();
+            assertThat(statistics.isOwningHandle()).isTrue();
+        }
+    }
+
+    @Test
+    void testStatisticsWithValidInstancePath(@TempDir Path tempFolder) throws Exception {
```
wuchong: I don't see any values of this test, is it necessary?

fluss-server/src/test/java/org/apache/fluss/server/kv/rocksdb/RocksDBResourceContainerTest.java
```diff
@@ -187,8 +193,237 @@ void testConfigurationOptionsFromConfig() throws Exception {
...
+        try (RocksDBResourceContainer container =
+                new RocksDBResourceContainer(config, instancePath)) {
+            Statistics statistics = container.getStatistics();
+            assertThat(statistics).isNotNull();
+            assertThat(statistics.isOwningHandle()).isTrue();
+        }
+    }
+
+    @Test
+    void testStatisticsConfigurationConsistency() throws Exception {
```
wuchong: I think this has been tested above.

fluss-server/src/test/java/org/apache/fluss/server/kv/rocksdb/RocksDBResourceContainerTest.java
```diff
@@ -187,8 +193,237 @@ void testConfigurationOptionsFromConfig() throws Exception {
...
+            assertThat(container.getStatistics()).isNotNull();
+        }
+    }
+
+    @Test
+    void testStatisticsMemoryManagement() throws Exception {
+        Configuration config = new Configuration();
+        config.setBoolean(ConfigOptions.KV_STATISTICS_ENABLED.key(), true);
+
+        // Create and close many containers to test memory management
```
wuchong: I don't see any tests or assersions for memory memory.

fluss-server/src/test/java/org/apache/fluss/server/kv/rocksdb/RocksDBResourceContainerTest.java
```diff
@@ -187,8 +193,237 @@ void testConfigurationOptionsFromConfig() throws Exception {
...
+            // Test concurrent access to statistics
+            int threadCount = 10;
+            CountDownLatch latch = new CountDownLatch(threadCount);
+            List<Exception> exceptions = Collections.synchronizedList(new ArrayList<>());
+
+            for (int i = 0; i < threadCount; i++) {
+                new Thread(
+                                () -> {
+                                    try {
+                                        // Access statistics concurrently
```
wuchong: I don't see any tests for the concurrency accessing. I think we don't need to add concurrency test just for `RocksDBResourceContainer#getStatistics`.

fluss-server/src/test/java/org/apache/fluss/server/kv/rocksdb/RocksDBResourceContainerTest.java
```diff
@@ -187,8 +193,237 @@ void testConfigurationOptionsFromConfig() throws Exception {
...
+                        .start();
+            }
+
+            latch.await(5, TimeUnit.SECONDS);
+            assertThat(exceptions).isEmpty();
+        }
+    }
+
+    @Test
+    void testStatisticsWithDifferentConfigurations() throws Exception {
```
wuchong: this has been tested above.

fluss-server/src/main/java/org/apache/fluss/server/kv/rocksdb/RocksDBResourceContainer.java
```diff
@@ -74,28 +77,25 @@ public class RocksDBResourceContainer implements AutoCloseable {
     /** The handles to be closed when the container is closed. */
     private final ArrayList<AutoCloseable> handlesToClose;
 
+    @Nullable private Statistics statistics;
+    @Nullable private Cache blockCache;
```
wuchong: They are initialized in the constructor, make them `final`, so the getters of them are thread-safe.

wuchong (commented):
Thanks @platinumhamburg , I left some comments about this changes. 

In general, the biggest concern of this PR is that it introduce too many metrics (>70) for a bucket-level RocksDB which will give a large pressure to the cluster and metric system. Flink also provides RocksDB native metrics, but are all disabled by default, and have to enable manually per metric: https://nightlies.apache.org/flink/flink-docs-master/docs/deployment/config/#rocksdb-native-metrics

wuchong (commented):
Thanks @platinumhamburg , I left some comments about this changes. 

In general, the biggest concern of this PR is that it introduce too many metrics (>70) for a bucket-level RocksDB which will give a large pressure to the cluster and metric system. Flink also provides RocksDB native metrics, but are all disabled by default, and have to enable manually per metric: https://nightlies.apache.org/flink/flink-docs-master/docs/deployment/config/#rocksdb-native-metrics

platinumhamburg:
This PR is deprecated, will submit a new one.

PR #1356, opened 2025-07-17
[hotfix] The DiscardingSink operator after LakeCommitter should always be with parallelism 1

luoyuxia:
### Purpose
Always to set `DiscardingSink` operator  after LakeCommitter to parallelism 1 to make it can be chained with `LakeCommitter`  since it's just a discarded sink 

Linked issue: close #xxx


fluss-flink/fluss-flink-common/src/main/java/com/alibaba/fluss/flink/tiering/LakeTieringJobBuilder.java
```diff
@@ -102,7 +102,9 @@ public JobClient build() throws Exception {
                         new TieringCommitOperatorFactory(flussConfig, lakeTieringFactory))
                 .setParallelism(1)
                 .setMaxParallelism(1)
-                .sinkTo(new DiscardingSink());
+                .sinkTo(new DiscardingSink())
+                .name("end")
```
luoyuxia, line 106: paimon also use name `end` for `DiscardingSink`, so let's also use it.

luoyuxia:
cc @leonardBang

PR #1386, opened 2025-07-23
[log] Fix the recovery log failed situation because of OutOfOrderSequenceException cause by writerId expire

### LINKED ISSUE #1383: Failed to recovery log because of OutOfOrderSequenceException cause by writerId expire
### Search before asking

### Fluss version

0.7.0 (latest release)

### Please describe the bug 🐞

When Fluss tabletServer restart and try to recovery log, an OutOfOrderSequenceException will be thrown as follow:

`Casued by: com.alibaba.fluss.exception.OutOfOrderSequenceException: Out of order batch sequence for writer 34055 at offset 2156767 in table-bucket Table-bucket TableBucket(tableId=13, partitionId=29, bucket=28) : 17871 (incoming batch seq.), -1 (current batch seq.)`

Image

This may cause by writerId expire (as we set 'server.writer-id.expiration-time' = '12h' )

### Solution

_No response_

swuferhong:
### Purpose

Linked issue: close #1383 

Fix the recovery log failed situation because of OutOfOrderSequenceException cause by writerId expire


wuchong:
IIRC, we want to refactor the fixing by implementing a special logic for rebuilding writer state from records that ignore OutOfOrder exception for records older than the writer-id expired time? So that we don't need to introduce the temporary&permanent TTL?

swuferhong:
> IIRC, we want to refactor the fixing by implementing a special logic for rebuilding writer state from records that ignore OutOfOrder exception for records older than the writer-id expired time? So that we don't need to introduce the temporary&permanent TTL?

Ok, I have changed the implements as you suggested.

swuferhong:
@wuchong comments addressed.

fluss-server/src/test/java/com/alibaba/fluss/server/log/WriterStateManagerTest.java
```diff
@@ -179,36 +182,94 @@ void testFetchSnapshotEmptySnapshot() {
...
+        clock.advanceTime(5000L, TimeUnit.MILLISECONDS);
+        // new one new WriterStateManager to mock tabletServer restart and reload from snapshot.
+        WriterStateManager stateManager2 =
+                new WriterStateManager(
+                        tableBucket,
+                        logDir,
+                        (int) conf.get(ConfigOptions.WRITER_ID_EXPIRATION_TIME).toMillis());
+        stateManager2.truncateAndReload(0L, 2L);
+        // all writers are retained.
+        assertThat(stateManager2.activeWriters().size()).isEqualTo(2);
```
wuchong: Why retain expired writers after reloading snapshot? IIUC, the retained expired writers will soon be cleaned up by the `PeriodicWriterIdExpirationCheck` thread. I think the new implementation can tolerate expired writers now.

fluss-server/src/test/java/com/alibaba/fluss/server/log/WriterStateManagerTest.java
```diff
@@ -179,36 +182,94 @@ void testFetchSnapshotEmptySnapshot() {
...
+        conf.set(ConfigOptions.WRITER_ID_EXPIRATION_TIME, Duration.ofSeconds(2));
+        WriterStateManager stateManager1 =
+                new WriterStateManager(
+                        tableBucket,
+                        logDir,
+                        (int) conf.get(ConfigOptions.WRITER_ID_EXPIRATION_TIME).toMillis());
+
+        // If we try to append an expired batch with none zero batch sequence, the
+        // OutOfOrderSequenceException will not been throw.
+        append(stateManager1, 1L, 10, 10L, true, clock.milliseconds());
```
wuchong, line 228: It will be better to add a check after this that the activeWriters in stateManager is 1, and the last sequence is 10 (use the new batch sequence).

fluss-server/src/main/java/com/alibaba/fluss/server/log/WriterAppendInfo.java
```diff
@@ -83,8 +88,9 @@ public WriterStateEntry toEntry() {
         return updatedEntry;
     }
 
-    private boolean inSequence(int lastBatchSeq, int nextBatchSeq) {
-        return nextBatchSeq == lastBatchSeq + 1L
+    private boolean inSequence(int lastBatchSeq, int nextBatchSeq, boolean isBatchExpired) {
+        return (lastBatchSeq == NO_BATCH_SEQUENCE && isBatchExpired)
```
wuchong: Add a comment to explain this statement, this is the core change of the logic.

fluss-server/src/main/java/com/alibaba/fluss/server/log/LogTablet.java
```diff
@@ -1115,7 +1115,8 @@ private static void updateWriterAppendInfo(
         // update writers.
         WriterAppendInfo appendInfo =
                 writers.computeIfAbsent(writerId, id -> writerStateManager.prepareUpdate(writerId));
-        appendInfo.append(batch);
+        appendInfo.append(
+                batch, writerStateManager.isBatchExpired(System.currentTimeMillis(), batch));
```
wuchong: Can we move the calculation of `writerStateManager.isBatchExpired` to `loadWritersFromRecords`? This can avoid too many calls of `System.currentTimeMillis()` for each record batch and only checks `isBatchExpired` for restoring state from records.

fluss-server/src/main/java/com/alibaba/fluss/server/log/WriterStateManager.java
```diff
@@ -419,6 +414,10 @@ private boolean isWriterExpired(long currentTimeMs, WriterStateEntry writerState
         return currentTimeMs - writerStateEntry.lastBatchTimestamp() > writerExpirationMs;
     }
 
+    public boolean isBatchExpired(long currentTimeMs, LogRecordBatch recordBatch) {
```
wuchong: How about renaming to `isWriterInBatchExpired` which I think is clearer?  `isBatchExpired` sounds like the batch is expired, but we don't have this concept.

fluss-server/src/main/java/com/alibaba/fluss/server/log/LogTablet.java
swuferhong: It seems we need to check both for restoring state from records and appending log as follower.

wuchong (approved):

PR #1401, opened 2025-07-24
[server] Coordinator Server Supports High-Available

### LINKED ISSUE #188: [Feature] CoordinatorServer Supports High-Available
### Search before asking

### Motivation

The current implementation of the CoordinatorServer in Fluss lacks high availability support, which poses a risk to system stability and reliability. Achieving high availability is crucial for ensuring that the CoordinatorServer can handle failures gracefully and continue to provide uninterrupted service to clients.

### Solution

We can deploy a standby CoordinatorServer alongside the primary one. In the event of a failure in the primary CoordinatorServer, the standby server will take over the control through a leader election process orchestrated by Zookeeper. Since the CoordinatorServer is stateless, this transition is seamless and straightforward.

### Anything else?

_No response_

### Willingness to contribute

zcoo:
### Purpose
Coordinator Server Supports High-Available.

Linked issue: close #188 

### Tests
see : com.alibaba.fluss.server.coordinator.CoordinatorServerElectionTest

### Documentation
Currently I make it by using an active coordinator server and several standby coordinator server. When several coordinator servers startup at the same time, only one can successfully preempt the ZK node and win the election and become active coordinator server leader. Once it failovers, the standby servers will take over it with an leader election again.


zcoo:
not ready for review

michaelkoepf:
@zcoo if it is not ready for review, you can click on `Convert to draft` in the right side bar, and, optionally, add `[WIP]` as prefix in the title of the PR.

zcoo:
@michaelkoepf i get it, thank you

zcoo:
Ready for review now! @wuchong @swuferhong

LiebingYu:
I have a question here. Do we need to properly handle `coordinatorEpoch` and `coordinatorEpochZkVersion` in this PR, just like Kafka does? In my opinion:
+ `coordinatorEpoch` prevents TabletServers from processing requests from zombie CoordinatorServers in the event of a split-brain scenario among CoordinatorServers.
+ `coordinatorEpochZkVersion`([KAFKA-6082](https://issues.apache.org/jira/browse/KAFKA-6082)) prevents zombie CoordinatorServers from modifying the ZK state during a split-brain scenario.

Should we take these two concerns into consideration in the same way as Kafka?

zcoo:
> I have a question here. Do we need to properly handle `coordinatorEpoch` and `coordinatorEpochZkVersion` in this PR, just like Kafka does? In my opinion:
> 
> * `coordinatorEpoch` prevents TabletServers from processing requests from zombie CoordinatorServers in the event of a split-brain scenario among CoordinatorServers.
> * `coordinatorEpochZkVersion`([KAFKA-6082](https://issues.apache.org/jira/browse/KAFKA-6082)) prevents zombie CoordinatorServers from modifying the ZK state during a split-brain scenario.
> 
> Should we take these two concerns into consideration in the same way as Kafka?

Thanks for reminder. I considered `coordinatorEpoch` and `coordinatorEpochZkVersion `, but ultimately did not implement them in this PR to simplify its complexity. I think we can handle it later.  

Do you have any suggestions?
@LB-Yu  @wuchong @swuferhong

LiebingYu:
> > I have a question here. Do we need to properly handle `coordinatorEpoch` and `coordinatorEpochZkVersion` in this PR, just like Kafka does? In my opinion:
> > 
> > * `coordinatorEpoch` prevents TabletServers from processing requests from zombie CoordinatorServers in the event of a split-brain scenario among CoordinatorServers.
> > * `coordinatorEpochZkVersion`([KAFKA-6082](https://issues.apache.org/jira/browse/KAFKA-6082)) prevents zombie CoordinatorServers from modifying the ZK state during a split-brain scenario.
> > 
> > Should we take these two concerns into consideration in the same way as Kafka?
> 
> Thanks for reminder. I considered `coordinatorEpoch` and `coordinatorEpochZkVersion `, but ultimately did not implement them in this PR to simplify its complexity. I think we can handle it later.
> 
> Do you have any suggestions? @LB-Yu @wuchong @swuferhong

IMO, we can split them into different PRs, but they should be merged together? Otherwise, we might introduce other metadata inconsistency issues when introducing HA. What's your opinion?

fluss-server/src/main/java/com/alibaba/fluss/server/coordinator/CoordinatorLeaderElection.java
```diff
@@ -0,0 +1,97 @@
...
+                                        serverId);
+                                initLeaderServices.run();
+
+                                // Do not return, otherwise the leader will be released immediately.
+                                while (true) {
+                                    try {
+                                        Thread.sleep(1000);
+                                    } catch (InterruptedException e) {
+                                    }
+                                }
```
LiebingYu: Why we don't use `LeaderLatch` here if we need to hold the leadership?

zcoo: Very good suggestion to me! It seems `LeaderLatch ` is more suitable in this scene and I will try to use it instead

fluss-server/src/main/java/com/alibaba/fluss/server/coordinator/CoordinatorServer.java
```diff
@@ -157,10 +158,45 @@ public static void main(String[] args) {
...
+                    new CoordinatorLeaderElection(zkClient.getCuratorClient(), serverId);
+            coordinatorLeaderElection.startElectLeader(
+                    () -> {
+                        try {
+                            startCoordinatorLeaderService();
+                        } catch (Exception e) {
+                            throw new RuntimeException(e);
+                        }
+                    });
+        }
```
LiebingYu: This logic looks strange to me. Why do we explicitly create the path for the first election, but use `LeaderLatch` for the standby Coordinator? Since we’re already using the framework, perhaps we could standardize on `LeaderLatch`?

Also, this logic seems problematic. If cs-1 becomes the leader on the first startup and later loses leadership due to a network issue, it appears it will never participate in the election again, because no `LeaderLatch` was registered for the node that became leader during the first startup.

zcoo: yes I think we don't need to use special logic in first round election. I have changed it.

LiebingYu:
Hi @zcoo. Overall, I have some concerns about several issues:

1. The mechanisms for leader election, startup, and resignation are not sufficiently robust. For reference, Kafka Controller provides a comprehensive implementation for controller lifecycle management:
```
Startup:
  Register ControllerChangeHandler and check whether the /controller path exists
  elect()
    Get activeControllerId from ZK
    If activeControllerId != -1, it means another broker has already become the controller, so just return
    Attempt to register as controller in ZK: registerControllerAndIncrementControllerEpoch
    If successful: call onControllerFailover()
    If failed:
      If it’s ControllerMovedException, call maybeResign (election failed)
      Otherwise, call triggerControllerMove (startup process failed)

When /controller node is created or its data changes:
  Call maybeResign()

When /controller node is deleted:
  processReelect
    maybeResign
    elect
```

2. There is no management of epochs. Combined with the lack of a leader resignation mechanism, this becomes extremely risky in the case of split-brain scenarios. There is the possibility that two leaders could be working simultaneously, leading to metadata inconsistencies.

3. After introducing multiple coordinators, should we re-examine potential race conditions in metadata management and ZK access to avoid the risks associated with split-brain situations?

zcoo:
@LiebingYu Sure. I think we need to these things and I will introduce epoc mechanism in this PR soon

swuferhong:
> > > I have a question here. Do we need to properly handle `coordinatorEpoch` and `coordinatorEpochZkVersion` in this PR, just like Kafka does? In my opinion:
> > > 
> > > * `coordinatorEpoch` prevents TabletServers from processing requests from zombie CoordinatorServers in the event of a split-brain scenario among CoordinatorServers.
> > > * `coordinatorEpochZkVersion`([KAFKA-6082](https://issues.apache.org/jira/browse/KAFKA-6082)) prevents zombie CoordinatorServers from modifying the ZK state during a split-brain scenario.
> > > 
> > > Should we take these two concerns into consideration in the same way as Kafka?
> > 
> > 
> > Thanks for reminder. I considered `coordinatorEpoch` and `coordinatorEpochZkVersion `, but ultimately did not implement them in this PR to simplify its complexity. I think we can handle it later.
> > Do you have any suggestions? @LB-Yu @wuchong @swuferhong
> 
> IMO, we can split them into different PRs, but they should be merged together? Otherwise, we might introduce other metadata inconsistency issues when introducing HA. What's your opinion?

+1  @zcoo Could you pls create these two issues? I think we can split the task and work together.

fluss-server/src/main/java/org/apache/fluss/server/zk/data/CoordinatorAddressJsonSerde.java
```diff
@@ -51,7 +51,7 @@ public void serialize(CoordinatorAddress coordinatorAddress, JsonGenerator gener
             throws IOException {
         generator.writeStartObject();
         writeVersion(generator);
-        generator.writeStringField(ID, coordinatorAddress.getId());
+        generator.writeNumberField(ID, coordinatorAddress.getId());
```
swuferhong, line 54: Don't modify it directly, backward compatibility must be preserved. In older versions, the ID is stored as `"id":"0"  (string)`, whereas in the new version it's stored as `"id":0 (integer)`. I believe the current code `int id = node.get(ID).asInt()` would throw an exception when encountering `"id":"0"` from older versions, since it's a JSON string, not a number.

The correct approach is to bump the schema version to `3`, and handle deserialization conditionally based on the version:

For `version 2`: read the id field as a string, then parse it to an integer.
For `version 3`: read the "id" field directly as an integer.
Additionally, compatibility tests must be added to verify that.

fluss-server/src/main/java/org/apache/fluss/server/coordinator/CoordinatorContext.java
```diff
@@ -54,6 +54,7 @@ public class CoordinatorContext {
     private static final Logger LOG = LoggerFactory.getLogger(CoordinatorContext.class);
 
     public static final int INITIAL_COORDINATOR_EPOCH = 0;
+    public static final int INITIAL_COORDINATOR_EPOCH_ZKVERSION = 0;
```
swuferhong: `INITIAL_COORDINATOR_EPOCH_ZKVERSION` -> `INITIAL_COORDINATOR_EPOCH_ZK_VERSION`

fluss-server/src/main/java/org/apache/fluss/server/coordinator/event/CoordinatorEventManager.java
```diff
@@ -62,6 +62,7 @@ public final class CoordinatorEventManager implements EventManager {
     private Histogram eventQueueTime;
 
     // Coordinator metrics moved from CoordinatorEventProcessor
+    private volatile int aliveCoordinatorServerCount;
```
swuferhong, line 65: `aliveCoordinatorServerCount` is never assigned value now?

fluss-server/src/main/java/org/apache/fluss/server/coordinator/CoordinatorServer.java
```diff
@@ -213,8 +239,8 @@ protected void startServices() throws Exception {
 
             registerCoordinatorLeader();
             // when init session, register coordinator server again
-            ZooKeeperUtils.registerZookeeperClientReInitSessionListener(
-                    zkClient, this::registerCoordinatorLeader, this);
+            //            ZooKeeperUtils.registerZookeeperClientReInitSessionListener(
+            //                    zkClient, this::registerCoordinatorLeader, this);
```
swuferhong: Why remove?

fluss-server/src/main/java/org/apache/fluss/server/coordinator/CoordinatorContext.java
```diff
@@ -104,13 +106,41 @@ public class CoordinatorContext {
 
     private ServerInfo coordinatorServerInfo = null;
     private int coordinatorEpoch = INITIAL_COORDINATOR_EPOCH;
+    private int coordinatorEpochZkVersion = INITIAL_COORDINATOR_EPOCH_ZKVERSION;
```
swuferhong: This value `coordinatorEpochZkVersion` don't need to be persisted in ZooKeeper? If it does need to be persisted, shouldn't we design the ZkData structure for it in this PR upfront?

fluss-server/src/main/java/org/apache/fluss/server/coordinator/CoordinatorServer.java
```diff
@@ -272,6 +298,41 @@ protected CompletableFuture<Result> closeAsync(Result result) {
         return terminationFuture;
     }
 
+    private void registerCoordinatorServer() throws Exception {
+        long startTime = System.currentTimeMillis();
+
+        // we need to retry to register since although
```
swuferhong, line 317: Format it.

fluss-server/src/main/java/org/apache/fluss/server/zk/ZooKeeperClient.java
```diff
@@ -161,20 +162,104 @@ public Optional<byte[]> getOrEmpty(String path) throws Exception {
...
+                                ZkData.CoordinatorEpochZNode.encode(newEpoch));
+
+                return Optional.of(newEpoch);
+            } catch (KeeperException.BadVersionException e) {
+                // Other coordinator leader has updated epoch.
+                // If this happens, it means our fence is in effect.
+                LOG.info("Coordinator leader {} failed to update epoch.", coordinatorId);
+            }
+        } catch (KeeperException.NodeExistsException e) {
+        }
```
swuferhong, line 228: Why catch this error but do nothing?

fluss-server/src/main/java/org/apache/fluss/server/coordinator/CoordinatorContext.java
```diff
@@ -104,13 +106,41 @@ public class CoordinatorContext {
...
+    public void setCoordinatorEpochAndZkVersion(int newEpoch, int newZkVersion) {
+        this.coordinatorEpoch = newEpoch;
+        this.coordinatorEpochZkVersion = newZkVersion;
+    }
+
+    public Set<Integer> getLiveCoordinatorServers() {
+        return liveCoordinatorServers;
+    }
+
+    @VisibleForTesting
```
swuferhong: No need? No test call this method.

fluss-server/src/test/java/org/apache/fluss/server/testutils/FlussClusterExtension.java
```diff
@@ -832,6 +834,15 @@ public CoordinatorServer getCoordinatorServer() {
         return coordinatorServer;
     }
 
+    public void waitUtilCoordinatorServerElected() {
+        waitUntil(
+                () -> {
+                    return zooKeeperClient.getCoordinatorLeaderAddress().isPresent();
```
swuferhong: Replace with expression lambda.

fluss-server/src/test/java/org/apache/fluss/server/coordinator/CoordinatorServerTest.java
```diff
@@ -63,10 +67,20 @@ protected ServerBase getStartFailServer() {
...
         assertThat(optCoordinatorAddr).isNotEmpty();
         verifyEndpoint(
                 optCoordinatorAddr.get().getEndpoints(),
                 coordinatorServer.getRpcServer().getBindEndpoints());
     }
+
+    public void waitUtilCoordinatorServerElected() {
+        waitUntil(
+                () -> {
+                    return zookeeperClient.getCoordinatorLeaderAddress().isPresent();
```
swuferhong: ditto，replace with expression lambda

fluss-server/src/main/java/org/apache/fluss/server/coordinator/event/watcher/CoordinatorChangeWatcher.java
```diff
@@ -0,0 +1,110 @@
...
+import org.apache.fluss.shaded.curator5.org.apache.curator.framework.recipes.cache.CuratorCache;
+import org.apache.fluss.shaded.curator5.org.apache.curator.framework.recipes.cache.CuratorCacheListener;
+import org.apache.fluss.shaded.curator5.org.apache.curator.utils.ZKPaths;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/** A watcher to watch the coordinator server changes(new/delete) in zookeeper. */
+public class CoordinatorServerChangeWatcher {
+
```
swuferhong, line 33: Rename to `CoordinatorChangeWatcher`?  I found classes related to the `coordinatorServer` are all named `CoordinatorXxxx`.

fluss-server/src/main/java/org/apache/fluss/server/coordinator/event/watcher/CoordinatorServerChangeWatcher.java
```diff
@@ -0,0 +1,110 @@
...
+import org.apache.fluss.shaded.curator5.org.apache.curator.framework.recipes.cache.CuratorCacheListener;
+import org.apache.fluss.shaded.curator5.org.apache.curator.utils.ZKPaths;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/** A watcher to watch the coordinator server changes(new/delete) in zookeeper. */
+public class CoordinatorServerChangeWatcher {
+
+    private static final Logger LOG = LoggerFactory.getLogger(CoordinatorServerChangeWatcher.class);
```
swuferhong: Do we need to introduce the class `CoordinatorServerChangeWatcher`? Its code is almost identical to that of `TabletServerChangeWatcher`, differing only in the input `zkPath`. Could we instead abstract `TabletServerChangeWatcher` into a common `ServerChangeWatcher`?

fluss-server/src/main/java/org/apache/fluss/server/coordinator/CoordinatorEventProcessor.java
```diff
@@ -248,7 +257,7 @@ public void shutdown() {
     private ServerInfo getCoordinatorServerInfo() {
         try {
             return zooKeeperClient
-                    .getCoordinatorAddress()
+                    .getCoordinatorLeaderAddress()
                     .map(
                             coordinatorAddress ->
                                     // TODO we set id to 0 as that CoordinatorServer don't support
```
swuferhong: Shouldn't this logic be updated so that we no longer pass coordinatorServer ID as `0`?

fluss-server/src/main/java/org/apache/fluss/server/zk/ZooKeeperClient.java
```diff
@@ -161,20 +162,104 @@ public Optional<byte[]> getOrEmpty(String path) throws Exception {
...
 
     /** Get the leader address registered in ZK. */
-    public Optional<CoordinatorAddress> getCoordinatorAddress() throws Exception {
-        Optional<byte[]> bytes = getOrEmpty(CoordinatorZNode.path());
-        return bytes.map(CoordinatorZNode::decode);
+    public Optional<CoordinatorAddress> getCoordinatorLeaderAddress() throws Exception {
+        Optional<byte[]> bytes = getOrEmpty(ZkData.CoordinatorLeaderZNode.path());
+        return bytes.map(
+                data ->
+                        // maybe a empty node when a leader is elected but not registered
```
swuferhong: `maybe a` => `maybe an`

fluss-server/src/main/java/org/apache/fluss/server/zk/data/ZkData.java
```diff
@@ -270,6 +301,24 @@ public static CoordinatorAddress decode(byte[] json) {
...
+        }
+
+        public static byte[] encode(int epoch) {
+            return String.valueOf(epoch).getBytes();
+        }
+
+        public static int decode(byte[] bytes) {
+            return Integer.parseInt(new String(bytes));
+        }
+    }
```
swuferhong, line 375: I feel the design here is a bit too complicated—we probably don't need separate `leader` and `epoch` to different node:
1. For `CoordinatorIdZNode`, it should store static node information, such as the data currently in `CoordinatorAddress`.
2. For `/coordinators/leader` and `/coordinators/epoch`, they should be merged into a single node, e.g., leader, which holds data like `["id"=xxx, "epoch"=xxx]`. Each update should read the previous epoch and then increment/update it accordingly.
3. Need to consider compatibility with the original version, as `CoordinatorAddress` stored in /coordinators/active.

fluss-server/src/main/java/org/apache/fluss/server/coordinator/CoordinatorContext.java
```diff
@@ -104,13 +106,41 @@ public class CoordinatorContext {
...
     public int getCoordinatorEpoch() {
         return coordinatorEpoch;
     }
 
+    public int getCoordinatorEpochZkVersion() {
+        return coordinatorEpochZkVersion;
+    }
+
+    public void setCoordinatorEpochAndZkVersion(int newEpoch, int newZkVersion) {
+        this.coordinatorEpoch = newEpoch;
```
swuferhong, line 126: It seems the `newZkVersion` will not be updated according to the value get from `zookeeper`?

keashem:
hi,@zcoo, What's the current progress of this pr...Can I work with you to complete this PR? I have already implemented a version to support coordinator high-available，we can together make works more quickly

zcoo:
> hi,@zcoo, What's the current progress of this pr...Can I work with you to complete this PR? I have already implemented a version to support coordinator high-available，we can together make works more quickly

Currently it is waiting for review now, I contacted @swuferhong to do code review offline. As you also implemented a version, can you also help make a review or merge our code to get a better version?

zcoo:
This pr split into 2 parts : see #2780 #2781

PR #1404, opened 2025-07-24
[Client] Add PojoConverterUtils

polyzos:


fluss-client/src/main/java/com/alibaba/fluss/client/utils/ConverterUtils.java
```diff
@@ -0,0 +1,608 @@
...
+    private final Constructor<T> defaultConstructor;
+
+    /**
+     * Creates a new converter for the specified POJO class and row type.
+     *
+     * @param pojoClass The class of POJOs to convert
+     * @param rowType The row schema to use for conversion
+     */
+    @SuppressWarnings("unchecked")
+    public ConverterUtils(Class<T> pojoClass, RowType rowType) {
```
michaelkoepf: Is there any reason for users to _not_ use the caching mechanism? If there is no reason to disable caching, we could make the constructor private, so users only have one way to obtain an instance of this class via the `getConverter` method?

fluss-client/src/main/java/com/alibaba/fluss/client/utils/ConverterUtils.java
```diff
@@ -0,0 +1,608 @@
...
+import java.time.ZoneOffset;
+import java.util.Arrays;
+import java.util.HashMap;
+import java.util.LinkedHashSet;
+import java.util.Map;
+import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
+
+/**
+ * Helper class for converting Java POJOs to Fluss's {@link InternalRow} format and vice versa.
```
michaelkoepf: I think the precondition introduced here ("Java POJO") is not restrictive enough, because a [POJO](https://www.baeldung.com/java-pojo-class#what-is-a-pojo) may not necessarily have a default constructor, but the code does assume one. 

Maybe it is better to use the more standardized concept of a [Java Bean](https://www.baeldung.com/java-pojo-class#what-is-a-javabean.)? Then, Java users immediately know what there classes must look like in order to use them with the converter. While Java Beans come with boilerplate (field access via getters), Java users won't bother too much because Java beans are everywhere, so they are used to it. An additional benefit is that by saying we expect a Java bean, there _must be_ a public constructor and public setters, so we do not need to tinker around with the `setAccessible` reflection calls. We can just fail with an exception.

fluss-client/src/main/java/com/alibaba/fluss/client/utils/ConverterUtils.java
```diff
@@ -0,0 +1,608 @@
...
+                }
+
+                // Create the appropriate converter for this field
+                converters[i] = createFieldToRowConverter(fieldType, field);
+            } else {
+                LOG.warn(
+                        "Field '{}' not found in POJO class {}. Will return null for this field.",
+                        fieldName,
+                        pojoClass.getName());
+                converters[i] = obj -> null;
```
michaelkoepf: Are there any valid cases where the field cannot be found in the POJO class? If not, maybe it is better to throw an exception because this would, imo, indicate a schema mismatch (i.e., wrong user input).

fluss-client/src/main/java/com/alibaba/fluss/client/utils/ConverterUtils.java
```diff
@@ -0,0 +1,608 @@
...
+                                field.getName());
+                        return null;
+                    }
+                };
+            default:
+                LOG.warn(
+                        "Unsupported type {} for field {}. Will use null for it.",
+                        fieldType.getTypeRoot(),
+                        field.getName());
+                return obj -> null;
```
michaelkoepf: Maybe better throw an exception if the user attempts to convert something with an unsupported data type?

polyzos:
@michaelkoepf, thank you for your comments; they made sense and were addressed.

fluss-client/src/main/java/com/alibaba/fluss/client/utils/ConverterUtils.java
```diff
@@ -0,0 +1,610 @@
...
+ * <p>Note: Nested POJO fields are not supported in the current implementation.
+ *
+ * @param <T> The POJO type to convert
+ */
+public class ConverterUtils<T> {
+    private static final Logger LOG = LoggerFactory.getLogger(ConverterUtils.class);
+
+    /** Cache for converters to avoid repeated reflection operations. */
+    private static final ConcurrentHashMap<CacheKey, ConverterUtils<?>> CONVERTER_CACHE =
+            MapUtils.newConcurrentHashMap();
```
MehulBatra: This cache grows indefinitely. In production systems with dynamic POJO types or frequent class loading/unloading, this will cause memory leaks in future, what do you think?

```
 if (CONVERTER_CACHE.size() >= MAX_CACHE_SIZE) {
            CONVERTER_CACHE.clear();
```

fluss-client/src/main/java/com/alibaba/fluss/client/utils/ConverterUtils.java
```diff
@@ -0,0 +1,610 @@
...
+            try {
+                value = fieldToRowConverters[i].convert(pojo);
+            } catch (IllegalAccessException e) {
+                LOG.warn(
+                        "Failed to access field {} in POJO class {}.",
+                        rowType.getFieldNames().get(i),
+                        pojoClass.getName(),
+                        e);
+            }
+            row.setField(i, value);
```
MehulBatra: When field access fails, the code sets null without throwing an exception. This corrupts data silently. Shouldn't we fail it?

fluss-client/src/main/java/com/alibaba/fluss/client/utils/ConverterUtils.java
polyzos: @MehulBatra, I don't think there is any fear for this growing indefinitely; in most cases, I even expect just one converter tbh.

It's also on the client, so there is no fear of OOM anything really.. This functionality here is also aimed more as a helper for cases where Pojos are quite big, and it's somewhat cumbersome to try and generate all the fields by hand.
WDYT?

fluss-client/src/main/java/com/alibaba/fluss/client/utils/ConverterUtils.java
polyzos: this is somewhat an edge case, im not sure how it can happen exactly, which is covered in tests, but in any case, even if set to null it will fail downstream

wuchong: +1, we should throw exceptions to let users know exactly the conversion fails instead of logging silently.

wuchong:
Thanks @polyzos for the pull request. In general this util looks good and useful. But I think we shouldn’t expose it directly to users. Ideally, users should call `writer.upsert(T record)` directly, without handling the conversion. The conversion from POJO to `InternalRow` should happen automatically under the API, because we already know the schema of the table. 

This magic is very similar to Flink Table API  `StreamTableEnvironment#fromDataStream(org.apache.flink.streaming.api.datastream.DataStream<T>)`  which automatically converts a `DataStream<T>` into `Table`. Underlying, it generates a `InputConversionOperator` to convert the POJO into `RowData`. It has a `DataStructureConverter` abstraction to do the conversion, and the POJO conversion magic is in `StructuredObjectConverter` (it uses codegen for conversion, but we can Java refelction in the first version).

We can introduce a `SpecificUpsertWriter<T> extends UpsertWriter` which provides additional POJO-type-based APIs `upsert(T record)` and `delete(T record)`. The `SpecificUpsertWriter<T>` can be got from `Upsert#createWriter(Class<T> specifcPojo)`. When the framework creates a `SpecificUpsertWriter`, it first checks the POJO type (whether it's a valid POJO type, and whether it can convert from/to InternalRow). 

We can introduce similar APIs for `Append` and `Scan`. 

Back to this pull request, I think we can first polish this PR a bit (I will leave some review comments later), and have another design/FIP to do the final API expose.

website/docs/apis/java-client.md
```diff
@@ -3,6 +3,24 @@ title: "Java Client"
 sidebar_position: 1
 ---
 
+<!--
```
wuchong: We want to avoid adding license comments at the top of doc markdown files, because Docusaurus uses the first line as the page’s meta description, and this "<!--" looks very strange as a description.

website/docs/apis/java-client.md
```diff
@@ -170,8 +188,8 @@ List<User> users = List.of(
 );
 ```
 
-**Note:** Currently data in Fluss is written in the form of `rows`, so we need to convert our POJO to `GenericRow`, while the Fluss community is working to provide
-a more user-friendly API for writing data.
+**Note:** Currently data in Fluss is written in the form of `rows`, so we need to convert our POJO to `GenericRow`. You can use the `ConverterUtils` helper class (see below) or manually convert your objects as shown here:
```
wuchong: Could you remove the doc changes as we don't want to expose the util directly?

fluss-client/src/test/java/com/alibaba/fluss/client/utils/ConverterUtilsTest.java
```diff
@@ -0,0 +1,485 @@
...
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+import java.util.Objects;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
+
+/** Tests for {@link ConverterUtils}. */
+public class ConverterUtilsTest extends ClientToServerITCaseBase {
```
wuchong: This doesn't need to extend `ClientToServerITCaseBase`, to avoid starting a cluster. This can be a very simple unit test to just test `ConverterUtils` itself. 

We should also add negative tests that uses illegal POJOs for conversion.

fluss-client/src/test/java/com/alibaba/fluss/client/utils/ConverterUtilsTest.java
```diff
@@ -0,0 +1,485 @@
...
+        assertThat(convertedPojo.floatField).isEqualTo(originalPojo.floatField);
+        assertThat(convertedPojo.doubleField).isEqualTo(originalPojo.doubleField);
+        assertThat(convertedPojo.stringField).isEqualTo(originalPojo.stringField);
+        assertThat(convertedPojo.bytesField).isEqualTo(originalPojo.bytesField);
+        assertThat(convertedPojo.decimalField).isEqualTo(originalPojo.decimalField);
+        assertThat(convertedPojo.dateField).isEqualTo(originalPojo.dateField);
+        assertThat(convertedPojo.timeField).isEqualTo(originalPojo.timeField);
+        assertThat(convertedPojo.timestampField).isEqualTo(originalPojo.timestampField);
+        assertThat(convertedPojo.timestampLtzField).isEqualTo(originalPojo.timestampLtzField);
+        assertThat(convertedPojo.offsetDateTimeField).isEqualTo(originalPojo.offsetDateTimeField);
```
wuchong: Remove the redundant assersions, beause the `assertThat(convertedPojo).isEqualTo(originalPojo);` already asserts each fields.

fluss-client/src/test/java/com/alibaba/fluss/client/utils/ConverterUtilsTest.java
```diff
@@ -0,0 +1,485 @@
...
+
+        // Verify the POJO
+        assertThat(pojo).isNotNull();
+        assertThat(pojo.booleanField).isEqualTo(false); // Default value for boolean is false
+        assertThat(pojo.byteField).isEqualTo((byte) 0); // Default value for byte is 0
+        assertThat(pojo.shortField).isEqualTo((short) 0); // Default value for short is 0
+        assertThat(pojo.intField).isEqualTo(0); // Default value for int is 0
+        assertThat(pojo.longField).isEqualTo(0L); // Default value for long is 0
+        assertThat(pojo.floatField).isEqualTo(0.0f); // Default value for float is 0.0
+        assertThat(pojo.doubleField).isEqualTo(0.0); // Default value for double is 0.0
```
wuchong: I think this is incorrect, if the original column value in `InternalRow` is null, the converted fields in POJO should also be null, otherwise, how users identify whether the value in Fluss table is null or is the 0 zero value? 

In this case, I think we should fail at the check phase when creating the converter, beacuse the fields are primitive values, not object values, which can't as a convert type for a nullable field. 

We should also add tests for object values `Double`, `Short` and so on, to verify they are null in this case.

fluss-client/src/test/java/com/alibaba/fluss/client/utils/ConverterUtilsTest.java
```diff
@@ -0,0 +1,485 @@
...
+        assertThat(convertedPojo.longField).isEqualTo(0L);
+        assertThat(convertedPojo.floatField).isEqualTo(0.0f);
+        assertThat(convertedPojo.doubleField).isEqualTo(0.0);
+        assertThat(convertedPojo.bytesField).isNull();
+        assertThat(convertedPojo.decimalField).isNull();
+        assertThat(convertedPojo.dateField).isNull();
+        assertThat(convertedPojo.timeField).isNull();
+        assertThat(convertedPojo.timestampField).isNull();
+        assertThat(convertedPojo.timestampLtzField).isNull();
+        assertThat(convertedPojo.offsetDateTimeField).isNull();
```
wuchong: Same with above, I think we shouldn't allow primitives for unmapped fields, and they should be null after conversion.

fluss-client/src/test/java/com/alibaba/fluss/client/utils/ConverterUtilsTest.java
```diff
@@ -0,0 +1,485 @@
...
+    }
+
+    @Test
+    public void testWriteAndReadPojos() throws Exception {
```
wuchong: Add POJO type test to FlussTableITCase, but I think in this pull request, we don't need IT case yet, because it's just a conversion util. We can introduce IT case, when support auto conversion.

fluss-client/src/main/java/com/alibaba/fluss/client/utils/ConverterUtils.java
wuchong: I think we can remove the cache. The `getConverter` will only be called once when an application is setup, so cache won't help much optimization here, but will introduce maintain overhead as @MehulBatra mentioned.

fluss-client/src/main/java/com/alibaba/fluss/client/utils/ConverterUtils.java
```diff
@@ -0,0 +1,610 @@
...
+            String fieldName = rowType.getFieldNames().get(i);
+            DataType fieldType = rowType.getTypeAt(i);
+
+            // Find field in POJO class
+            Field field = findField(pojoClass, fieldName);
+            if (field != null) {
+                pojoFields[i] = field;
+
+                // Check if the field type is supported
+                if (!SUPPORTED_TYPES.containsKey(fieldType.getTypeRoot())) {
```
wuchong: We should also check the field type is in the supported classes.

fluss-client/src/main/java/com/alibaba/fluss/client/utils/ConverterUtils.java
```diff
@@ -0,0 +1,610 @@
...
+    /**
+     * Utility method to create an ordered {@link LinkedHashSet} containing the specified Java type
+     * classes.
+     *
+     * <p>The returned set maintains the insertion order of the provided classes.
+     *
+     * @param javaTypes The Java type classes to include in the set. May be one or more classes.
+     * @return A new {@link LinkedHashSet} containing the given classes, preserving their order.
+     */
+    private static LinkedHashSet<Class<?>> orderedSet(Class<?>... javaTypes) {
```
wuchong: Do we need the orderness here? It seems we didn't use the orderness in the code.

fluss-client/src/main/java/com/alibaba/fluss/client/utils/ConverterUtils.java
```diff
@@ -0,0 +1,610 @@
...
+                return obj -> {
+                    Object value = field.get(obj);
+                    if (value == null) {
+                        return null;
+                    }
+                    if (value instanceof LocalDate) {
+                        return (int) ((LocalDate) value).toEpochDay();
+                    } else {
+                        LOG.warn(
+                                "Field {} is not a LocalDate. Cannot convert to int days.",
```
wuchong: throw exception instead of logging, this should be an error that user shoulld be aware of.

fluss-client/src/main/java/com/alibaba/fluss/client/utils/ConverterUtils.java
```diff
@@ -0,0 +1,610 @@
...
+                    Object value = field.get(obj);
+                    if (value == null) {
+                        return null;
+                    }
+                    if (value instanceof LocalTime) {
+                        LocalTime localTime = (LocalTime) value;
+                        return (int) (localTime.toNanoOfDay() / 1_000_000);
+                    } else {
+                        LOG.warn(
+                                "Field {} is not a LocalTime. Cannot convert to int millis.",
```
wuchong: ditto

fluss-client/src/main/java/com/alibaba/fluss/client/utils/ConverterUtils.java
```diff
@@ -0,0 +1,610 @@
...
+                return obj -> {
+                    Object value = field.get(obj);
+                    if (value == null) {
+                        return null;
+                    }
+                    if (value instanceof LocalDateTime) {
+                        return TimestampNtz.fromLocalDateTime((LocalDateTime) value);
+                    } else {
+                        LOG.warn(
+                                "Field {} is not a LocalDateTime. Cannot convert to TimestampData.",
```
wuchong: ditto

fluss-client/src/main/java/com/alibaba/fluss/client/utils/ConverterUtils.java
```diff
@@ -0,0 +1,610 @@
...
+                        return null;
+                    }
+                    if (value instanceof Instant) {
+                        return TimestampLtz.fromInstant((Instant) value);
+                    } else if (value instanceof OffsetDateTime) {
+                        OffsetDateTime offsetDateTime = (OffsetDateTime) value;
+                        return TimestampLtz.fromInstant(offsetDateTime.toInstant());
+                    } else {
+                        LOG.warn(
+                                "Field {} is not an Instant or OffsetDateTime. Cannot convert to TimestampData.",
```
wuchong: ditto

fluss-client/src/main/java/com/alibaba/fluss/client/utils/ConverterUtils.java
```diff
@@ -0,0 +1,610 @@
...
+        SUPPORTED_TYPES.put(
+                DataTypeRoot.TIMESTAMP_WITH_LOCAL_TIME_ZONE,
+                orderedSet(Instant.class, OffsetDateTime.class));
+
+        // TODO: Add more types when https://github.com/apache/fluss/issues/816 is merged
+    }
+
+    /** Interface for field conversion from POJO field to Fluss InternalRow field. */
+    private interface FieldToRowConverter {
+        Object convert(Object obj) throws IllegalAccessException;
```
wuchong: `throws Exception` throws any exception to make the wrapper catch all the exception.

fluss-client/src/main/java/com/alibaba/fluss/client/utils/ConverterUtils.java
```diff
@@ -0,0 +1,610 @@
...
+    }
+
+    /** Interface for field conversion from Fluss InternalRow field to POJO field. */
+    private interface RowToFieldConverter {
+        Object convert(InternalRow row, int pos) throws IllegalAccessException;
```
wuchong: `throws Exception` throws any exception to make the wrapper catch all the exception.

fluss-client/src/main/java/com/alibaba/fluss/client/utils/ConverterUtils.java
```diff
@@ -0,0 +1,610 @@
...
+            return null;
+        }
+
+        GenericRow row = new GenericRow(rowType.getFieldCount());
+
+        for (int i = 0; i < fieldToRowConverters.length; i++) {
+            Object value = null;
+            try {
+                value = fieldToRowConverters[i].convert(pojo);
+            } catch (IllegalAccessException e) {
```
wuchong: catch all the exceptions and re-throw with additional information.

fluss-client/src/main/java/com/alibaba/fluss/client/utils/ConverterUtils.java
```diff
@@ -0,0 +1,610 @@
...
+                    }
+                    int millis = row.getInt(pos);
+                    return LocalTime.ofNanoOfDay(millis * 1_000_000L);
+                };
+            case TIMESTAMP_WITHOUT_TIME_ZONE:
+                return (row, pos) -> {
+                    if (row.isNullAt(pos)) {
+                        return null;
+                    }
+                    TimestampNtz timestampNtz = row.getTimestampNtz(pos, getPrecision(fieldType));
```
wuchong: move the `getPrecision`  out of the lambda, to avoid repeated calls

fluss-client/src/main/java/com/alibaba/fluss/client/utils/ConverterUtils.java
```diff
@@ -0,0 +1,610 @@
...
+        }
+    }
+
+    /**
+     * Gets the precision of a data type.
+     *
+     * @param dataType The data type
+     * @return The precision
+     */
+    private static int getPrecision(DataType dataType) {
```
wuchong: Use `com.alibaba.fluss.types.DataTypeChecks.getPrecision` directly.

fluss-client/src/main/java/com/alibaba/fluss/client/utils/ConverterUtils.java
```diff
@@ -0,0 +1,610 @@
...
+ *
+ * // Convert a GenericRow back to POJO
+ * Order convertedOrder = converter.fromRow(row);
+ * }</pre>
+ *
+ * <p>Note: Nested POJO fields are not supported in the current implementation.
+ *
+ * @param <T> The POJO type to convert
+ */
+public class ConverterUtils<T> {
```
wuchong: Better to call this class `PojoConverterUtils`, conveter is too generic.

fluss-client/src/main/java/com/alibaba/fluss/client/utils/ConverterUtils.java
```diff
@@ -0,0 +1,610 @@
...
+    private final Constructor<T> defaultConstructor;
+
+    /**
+     * Creates a new converter for the specified POJO class and row type.
+     *
+     * @param pojoClass The class of POJOs to convert
+     * @param rowType The row schema to use for conversion
+     */
+    @SuppressWarnings("unchecked")
+    private ConverterUtils(Class<T> pojoClass, RowType rowType) {
```
wuchong: I think we need to separate the converter util into 2 class, one is for writer (convert from POJO to InternalRow), the other is for scanner (convert from InternalRow to POJO), because they have different validations (especially for nullable fields). 

For example, if this is for scanner, considering projection pushdown, 

If this converter is for scanner, considering projection pushdown case, there might be some fields are not seleced and value should be null. If this converter is for writer, considering partial updates case, there might be some fields are not used for writes and also will be null. 

So I suggest some requirements and pre-checks on the POJO types:  
1. ALL fields should be nullable, so no primitive types are allowed. This makes the usage of the POJO class feasible in any case. 
2. All fields should be exactly matched to the table schema (no less or more fields)
3. This should be a POJO class that (see [Flink docs](https://nightlies.apache.org/flink/flink-docs-master/docs/dev/datastream/fault-tolerance/serialization/types_serialization/#pojos), and [implementation](https://github.com/apache/flink/blob/master/flink-core/src/main/java/org/apache/flink/api/java/typeutils/TypeExtractor.java#L2121)): 
   - class must be public
   - must have a public constructor without arguments (default constructor).
   - All fields are either public or must be accessible through getter and setter functions. For a field called foo the getter and setter methods must be named getFoo() and setFoo().
4. We should introduce a internal PojoType class to collect the basic type and field information, this can easier later checks (compare the PojoType and table schema).

fluss-client/src/main/java/com/alibaba/fluss/client/utils/ConverterUtils.java
```diff
@@ -0,0 +1,610 @@
...
+            case DOUBLE:
+                return (row, pos) -> row.isNullAt(pos) ? null : row.getDouble(pos);
+            case CHAR:
+            case STRING:
+                return (row, pos) -> {
+                    if (row.isNullAt(pos)) {
+                        return null;
+                    }
+                    BinaryString binaryString = row.getString(pos);
+                    return binaryString.toString();
```
wuchong: should check the type class, we may need to conver it to `Character` or `char`.

fluss-client/src/main/java/com/alibaba/fluss/client/utils/ConverterUtils.java
```diff
@@ -0,0 +1,610 @@
...
+                                field.getName());
+                        return null;
+                    }
+                };
+            case TIMESTAMP_WITH_LOCAL_TIME_ZONE:
+                return obj -> {
+                    Object value = field.get(obj);
+                    if (value == null) {
+                        return null;
+                    }
```
wuchong: We should reuse the logic of get field value and null checks, and makes the conversion focus on not_null_object -> not_null_object conversion. You can do this by add a nullalbe wrapper on the top of a not-null wrapper.

fluss-client/src/main/java/com/alibaba/fluss/client/utils/ConverterUtils.java
```diff
@@ -0,0 +1,610 @@
...
+     * @param field The Java reflection field
+     * @return A converter for this field
+     */
+    private RowToFieldConverter createRowToFieldConverter(DataType fieldType, Field field) {
+        field.setAccessible(true);
+        Class<?> fieldClass = field.getType();
+
+        switch (fieldType.getTypeRoot()) {
+            case BOOLEAN:
+                return (row, pos) -> row.isNullAt(pos) ? null : row.getBoolean(pos);
```
wuchong: We should reuse the logic of null checks, and makes the conversion focus on not_null_object -> not_null_object conversion. You can do this by add a nullalbe wrapper on the top of a not-null wrapper.

polyzos:
@MehulBatra @wuchong thank you both for your comments! I will iterate ASAP.
I opened #1731  for tracking future improvements

polyzos:
@wuchong Thank you for being so mindful of the project. Something that started as just a helper utility might end up having a wider long-term impact thanks to your well-crafted feedback and guidance. 🙏
I addressed your comments, so please take a look and let me know if there is anything I have missed.

fluss-client/src/main/java/org/apache/fluss/client/converter/RowToPojoConverter.java
```diff
@@ -0,0 +1,255 @@
...
+        if (row == null) {
+            return null;
+        }
+        try {
+            T pojo = pojoType.getDefaultConstructor().newInstance();
+            for (int i = 0; i < rowReaders.length; i++) {
+                if (!row.isNullAt(i)) {
+                    Object v = rowReaders[i].convert(row, i);
+                    PojoType.Property prop =
+                            pojoType.getProperty(projection.getFieldNames().get(i));
```
wuchong: Currently, the `RowType#getFieldNames()` is a heavy operation. Could you add a member variable for the `projection.getFieldNames()`?

fluss-client/src/main/java/org/apache/fluss/client/converter/RowToPojoConverter.java
```diff
@@ -0,0 +1,255 @@
...
+     * @throws IllegalArgumentException if the target type is unsupported or constraints are
+     *     violated
+     */
+    private static Object convertTextValue(
+            DataType fieldType, PojoType.Property prop, BinaryString s) {
+        String v = s.toString();
+        String formattedMessage =
+                String.format(
+                        "Field %s expects exactly one character for CHAR type, got length %d.",
+                        prop.name, v.length());
```
wuchong: Formatting message for each record is a big overhead, I think it's fine to put them into the exception branch to have better performance. I think it's fine to not reuse the code for just a single line if the performance matters here.

fluss-client/src/main/java/org/apache/fluss/client/converter/RowToPojoConverter.java
```diff
@@ -0,0 +1,255 @@
...
+    }
+
+    /** Converts a TIME_WITHOUT_TIME_ZONE value stored as int millis of day to a LocalTime. */
+    private static LocalTime convertTimeValue(InternalRow row, int pos) {
+        return LocalTime.ofNanoOfDay(row.getInt(pos) * 1_000_000L);
+    }
+
+    /** Converts a TIMESTAMP_WITHOUT_TIME_ZONE value to a LocalDateTime honoring precision. */
+    private static Object convertTimestampNtzValue(DataType fieldType, InternalRow row, int pos) {
+        final int precision = DataTypeChecks.getPrecision(fieldType);
```
wuchong: We can get the precision for the type before the `convertTimestampNtzValue` method, and pass in the precision as a parameter, so it won't to get the precision for every record. This can improve the performance a lot, because the visitor is costing.

fluss-client/src/main/java/org/apache/fluss/client/converter/RowToPojoConverter.java
```diff
@@ -0,0 +1,255 @@
...
+        return t.toLocalDateTime();
+    }
+
+    /**
+     * Converts a TIMESTAMP_WITH_LOCAL_TIME_ZONE value to either Instant or OffsetDateTime in UTC,
+     * depending on the target POJO property type.
+     */
+    private static Object convertTimestampLtzValue(
+            DataType fieldType, PojoType.Property prop, InternalRow row, int pos) {
+        final int precision = DataTypeChecks.getPrecision(fieldType);
```
wuchong: ditto

wuchong (commented):
Thanks @polyzos for the updates. The new version looks really nice and the code in a high quality with full test coverage. I only left some minor comments. I think it's already in a good shape and ready to be merged.

polyzos:
@wuchong comments addressed.. PTAL!

wuchong (approved):
LGTM.

Rebased & squashed & minor improvements

PR #1410, opened 2025-07-28
Change the implementation of Admin#getServerNodes() to support carrying ServerTag

### LINKED ISSUE #1389: [rebalance] Change the implementation of Admin#getServerNodes() as introducing a new API to get cluster information
### Search before asking

### Description

Change the implementation of Admin#getServerNodes() as introducing a new API to get cluster information. In this way, we can realizing get ServerTag from this API instead of introducing a new API like `showServerTags`

### Willingness to contribute

LiebingYu:
### Purpose

Linked issue: close #1389

### DIFF
--- a/fluss-server/src/main/java/org/apache/fluss/server/RpcServiceBase.java
+++ b/fluss-server/src/main/java/org/apache/fluss/server/RpcServiceBase.java
@@ -20,6 +20,7 @@ package org.apache.fluss.server;
 import org.apache.fluss.annotation.VisibleForTesting;
 import org.apache.fluss.cluster.ServerNode;
 import org.apache.fluss.cluster.ServerType;
+import org.apache.fluss.cluster.rebalance.ServerTag;
 import org.apache.fluss.config.cluster.ConfigEntry;
 import org.apache.fluss.exception.FlussRuntimeException;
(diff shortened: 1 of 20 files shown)


fluss-server/src/main/java/org/apache/fluss/server/RpcServiceBase.java
```diff
@@ -712,4 +730,40 @@ private boolean isPartitionAssignmentMissingFromZk(long partitionId) {
...
+
+    protected DescribeClusterResponse makeDescribeClusterResponse(
+            String listenerName, ServerMetadataCache metadataCache) {
+        ServerNode coordinatorServer = metadataCache.getCoordinatorServer(listenerName);
+
+        Collection<ServerNode> aliveTabletServers =
+                metadataCache.getAllAliveTabletServers(listenerName).values();
+        Set<ServerNode> aliveTabletServersWithTag = new HashSet<>(aliveTabletServers.size());
+        try {
+            Optional<ServerTags> serverTagsOp = zkClient.getServerTags();
```
affo, line 742: Why not getting this out of the existing `CoordinatorMetadataCache`?
It would avoid a blocking call and fetch directly from memory.

affo (commented):
Just a doubt about adding a blocking call to ZK

PR #1422, opened 2025-07-30
[server] FETCH_LOG should return 1 remote segments file rather than all the segments by default.

### LINKED ISSUE #1421: RemoteLogDownloader should download remote files in priority of file timestamp
### Search before asking

### Description

Currently, FETCH_LOG returns all segments of a table bucket. This means that if we start fetching from two days ago and each table bucket contains 20 remote segments, the system will return 20 × (number of table buckets) files.

During the download process, the current behavior is to download all segments of table bucket 1 first, followed by all segments of table bucket 2, and so on. This leads to an unstable delay pattern: the latency drops from two days down to minutes, then suddenly spikes back up to nearly two days, and repeats this cycle continuously.

If instead, only one segment per table bucket is returned in each fetch, the delay metric would become much more stable and predictable, resulting in a smoother and more user-friendly experience.

### Solution

**Updated at 8.23**

As discussed offline with @wuchong and @loserwang1024 , we decided to use a simple way to fix this problem by download remote files in the order of `max_timestamp` of log segments. This allows us still fetch all the segment file metadata once to avoid potential performance degression and guarantee a stable increasing fetch delay metric. 

### Willingness to contribute

loserwang1024:
### Purpose

Linked issue: close #1421

### DIFF
--- a/fluss-client/src/main/java/com/alibaba/fluss/client/table/scanner/log/LogFetcher.java
+++ b/fluss-client/src/main/java/com/alibaba/fluss/client/table/scanner/log/LogFetcher.java
@@ -94,6 +94,7 @@ public class LogFetcher implements Closeable {
     private final RpcClient rpcClient;
     private final int maxFetchBytes;
     private final int maxBucketFetchBytes;
+    private final long maxBucketRemoteFetchBytes;
     private final int minFetchBytes;
     private final int maxFetchWaitMs;
     private final boolean isCheckCrcs;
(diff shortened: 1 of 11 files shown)


wuchong:
@loserwang1024 As we discussed offline, let's refactor it with a priority queue in the download thread.

wuchong:
I opened a new PR for this issue https://github.com/apache/fluss/pull/1579, please help to review

PR #1426, opened 2025-07-31
[docs] Fix warehouse path of paimon inconsistent in lakehouse-storage page

luoyuxia:
### Purpose

Linked issue: close #xxx
Although it's just a demo, but still fix it to avoid misleading.


beryllw:
LGTM.

FYI, the following parameters:
```plaintext
--datalake.format paimon
--datalake.paimon.metastore filesystem
--datalake.paimon.warehouse /tmp/paimon
```
are already configured in server.yml. The Flink datalake tiering service can fetch these configurations directly from the fluss cluster, eliminating the need for users to set them. Alternatively, we could add validation to ensure user configurations match the server's. I'll create a new issue to address this.

PR #1446, opened 2025-08-01
[server] Use batch operations of ZooKeeper to optimize updateLeaderAndIsr

### LINKED ISSUE #1445: Use batch operations of ZooKeeper to optimize the performance of updating LeaderAndIsr in tryProcessAdjustIsr.
### Search before asking

### Description

Currently, the operation to update LeaderAndIsr in tryProcessAdjustIsr() is executed synchronously and sequentially. In large clusters, if ISR flapping occurs, it may result in a large number of AdjustIsrReceivedEvent events, which can in turn cause the Coordinator event processing thread to become blocked. We need to optimize the performance of tryProcessAdjustIsr() so that it can return more quickly.

I tried both the batch and async API of Curator, and I foud that the batch API performs better.
```java
    public void batchUpdateLeaderAndIsr(Map<TableBucket, LeaderAndIsr> leaderANdIsrList)
            throws Exception {
        if (leaderANdIsrList.isEmpty()) {
            return;
        }

        List<CuratorOp> ops = new ArrayList<>(leaderANdIsrList.size());
        for (Map.Entry<TableBucket, LeaderAndIsr> entry : leaderANdIsrList.entrySet()) {
            TableBucket tableBucket = entry.getKey();
            LeaderAndIsr leaderAndIsr = entry.getValue();

            String path = LeaderAndIsrZNode.path(tableBucket);
            byte[] data = LeaderAndIsrZNode.encode(leaderAndIsr);
            CuratorOp updateOp = zkClient.transactionOp().setData().forPath(path, data);
            ops.add(updateOp);
            if (ops.size() == MAX_BATCH_SIZE) {
                zkClient.transaction().forOperations(ops);
                ops.clear();
            }
        }
        if (!ops.isEmpty()) {
            zkClient.transaction().forOperations(ops);
        }
    }

    public void asyncUpdateLeaderAndIsr(Map<TableBucket, LeaderAndIsr> leaderANdIsrList)
            throws Exception {
        if (leaderANdIsrList.isEmpty()) {
            return;
        }

        CountDownLatch latch = new CountDownLatch(leaderANdIsrList.size());
        for (Map.Entry<TableBucket, LeaderAndIsr> entry : leaderANdIsrList.entrySet()) {
            TableBucket tableBucket = entry.getKey();
            LeaderAndIsr leaderAndIsr = entry.getValue();

            String path = LeaderAndIsrZNode.path(tableBucket);
            byte[] data = LeaderAndIsrZNode.encode(leaderAndIsr);
            zkClient.setData()
                    .inBackground(
                            (client, event) -> {
                                latch.countDown();
                            })
                    .forPath(path, data);
        }
        latch.await();
    }

    @Test
    void testBatchUpdateLeaderAndIsr() throws Exception {
        int totalCount = 100000;

        Map<TableBucket, LeaderAndIsr> leaderAndIsrList = new HashMap<>();
        for (int i = 0; i < totalCount; i++) {
            TableBucket tableBucket = new TableBucket(1, i);
            LeaderAndIsr leaderAndIsr =
                    new LeaderAndIsr(i, 10, Arrays.asList(i + 1, i + 2, i + 3), 100, 1000);
            leaderAndIsrList.put(tableBucket, leaderAndIsr);
            zookeeperClient.registerLeaderAndIsr(tableBucket, leaderAndIsr);
        }

        long startTime = System.currentTimeMillis();
        zookeeperClient.batchUpdateLeaderAndIsr(leaderAndIsrList);
        long endTime = System.currentTimeMillis();
        System.out.println("batchUpdateLeaderAndIsr cost: " + (endTime - startTime) + "ms");

        startTime = System.currentTimeMillis();
        zookeeperClient.asyncUpdateLeaderAndIsr(leaderAndIsrList);
        endTime = System.currentTimeMillis();
        System.out.println("asyncUpdateLeaderAndIsr cost: " + (endTime - startTime) + "ms");

        startTime = System.currentTimeMillis();
        for (Map.Entry<TableBucket, LeaderAndIsr> entry : leaderAndIsrList.entrySet()) {
            zookeeperClient.updateLeaderAndIsr(entry.getKey(), entry.getValue());
        }
        endTime = System.currentTimeMillis();
        System.out.println("updateLeaderAndIsr cost: " + (endTime - startTime) + "ms");
    }
```

In my local test, I get the result like below. So I will use batch operation to optimize it.
```
batchUpdateLeaderAndIsr cost: 777ms
asyncUpdateLeaderAndIsr cost: 3528ms
updateLeaderAndIsr cost: 12294ms
```

### Willingness to contribute

LiebingYu:
…e of updating LeaderAndIsr in tryProcessAdjustIsr

### Purpose

Linked issue: close #1445

### DIFF
--- a/fluss-server/src/test/java/com/alibaba/fluss/server/coordinator/CoordinatorEventProcessorTest.java
+++ b/fluss-server/src/test/java/com/alibaba/fluss/server/coordinator/CoordinatorEventProcessorTest.java
@@ -31,16 +31,19 @@ import com.alibaba.fluss.metadata.TableBucketReplica;
 import com.alibaba.fluss.metadata.TableDescriptor;
 import com.alibaba.fluss.metadata.TablePartition;
 import com.alibaba.fluss.metadata.TablePath;
+import com.alibaba.fluss.rpc.messages.AdjustIsrResponse;
 import com.alibaba.fluss.rpc.messages.CommitKvSnapshotResponse;
 import com.alibaba.fluss.rpc.messages.CommitRemoteLogManifestResponse;
 import com.alibaba.fluss.rpc.messages.NotifyKvSnapshotOffsetRequest;
 import com.alibaba.fluss.rpc.messages.NotifyRemoteLogOffsetsRequest;
 import com.alibaba.fluss.server.coordinator.event.AccessContextEvent;
+import com.alibaba.fluss.server.coordinator.event.AdjustIsrReceivedEvent;
 import com.alibaba.fluss.server.coordinator.event.CommitKvSnapshotEvent;
 import com.alibaba.fluss.server.coordinator.event.CommitRemoteLogManifestEvent;
 import com.alibaba.fluss.server.coordinator.event.CoordinatorEventManager;
 import com.alibaba.fluss.server.coordinator.statemachine.BucketState;
 import com.alibaba.fluss.server.coordinator.statemachine.ReplicaState;
+import com.alibaba.fluss.server.entity.AdjustIsrResultForBucket;
 import com.alibaba.fluss.server.entity.CommitKvSnapshotData;
 import com.alibaba.fluss.server.entity.CommitRemoteLogManifestData;
 import com.alibaba.fluss.server.kv.snapshot.CompletedSnapshot;
@@ -78,6 +81,7 @@ import java.nio.file.Path;
 import java.time.Duration;
 import java.util.Arrays;
 import java.util.Collections;
+import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
@@ -99,6 +103,7 @@ import static com.alibaba.fluss.server.coordinator.statemachine.BucketState.Onli
 import static com.alibaba.fluss.server.coordinator.statemachine.ReplicaState.OfflineReplica;
 import static com.alibaba.fluss.server.coordinator.statemachine.ReplicaState.OnlineReplica;
 import static com.alibaba.fluss.server.testutils.KvTestUtils.mockCompletedSnapshot;
+import static com.alibaba.fluss.server.utils.ServerRpcMessageUtils.getAdjustIsrResponseData;
 import static com.alibaba.fluss.server.utils.TableAssignmentUtils.generateAssignment;
 import static com.alibaba.fluss.testutils.common.CommonTestUtils.retry;
 import static com.alibaba.fluss.testutils.common.CommonTestUtils.waitValue;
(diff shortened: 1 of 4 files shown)


fluss-server/src/test/java/com/alibaba/fluss/server/zk/ZooKeeperClientTest.java
```diff
@@ -220,6 +220,31 @@ void testBatchCreateLeaderAndIsr() throws Exception {
...
+
+        // try to register bucket leaderAndIsr
+        Map<TableBucket, LeaderAndIsr> leaderAndIsrList = new HashMap<>();
+        for (int i = 0; i < totalCount; i++) {
+            TableBucket tableBucket = new TableBucket(1, i);
+            LeaderAndIsr leaderAndIsr =
+                    new LeaderAndIsr(i, 10, Arrays.asList(i + 1, i + 2, i + 3), 100, 1000);
+            leaderAndIsrList.put(tableBucket, leaderAndIsr);
+            zookeeperClient.registerLeaderAndIsr(tableBucket, leaderAndIsr);
+        }
```
wuchong, line 236: Is it necessary to call `register leaderAndIsr`? I'm a bit confused about this code because this makes even if `batchUpdateLeaderAndIsr` fails, the subsequent assertion on `leaderAndIsr` could still pass. For example, when I remove line 238 (`zookeeperClient.batchUpdateLeaderAndIsr(leaderAndIsrList)`), the test still passes. Maybe we need to remove this part code.

fluss-server/src/test/java/com/alibaba/fluss/server/coordinator/CoordinatorEventProcessorTest.java
```diff
@@ -761,6 +766,59 @@ void testNotifyOffsetsWithShrinkISR(@TempDir Path tempDir) throws Exception {
         verifyReceiveRequestExceptFor(3, leader, NotifyKvSnapshotOffsetRequest.class);
     }
 
+    @Test
+    void testProcessAdjustIsr() throws Exception {
```
wuchong, line 770: This test seems doing a void testing. It gets the current `LeaderAndIsr`s and update `LeaderAndIsr`s without any changes, and then verify the `LeaderAndIsr` again. If the `processAdjustIsr` has something wrong, this test can't find anything. 

I suggest to do some modification on the retrieved `LeaderAndIsr` from ctx, like changing the leader to the second element in isr, and then send the new `LeaderAndIsr` within `AdjustIsrReceivedEvent` to coordinator, finally, asserting the `LeaderAndIsr` in ctx and response is updated to the new leader.

wuchong (commented):
Thanks @LB-Yu for the contribution, the changes look good to me. I only left some comments around tests.

fluss-server/src/test/java/com/alibaba/fluss/server/coordinator/CoordinatorEventProcessorTest.java
LiebingYu: I don't think this test is invalid. When handling the `AdjustIsrReceivedEvent` event (in the `CoordinatorEventProcessor#tryProcessAdjustIsr` method), the `bucketEpoch` of LeaderAndIsr is incremented (for example, from 100 to 101). This test verifies from the context that the `bucketEpoch` has indeed increased.

The main purpose of this test is to verify whether the handling of the `AdjustIsrReceivedEvent` event is correct. I think we shouldn't modify `LeaderAndIsr` manually, but rather check whether `LeaderAndIsr` reaches the expected state after `AdjustIsrReceivedEvent` is properly handled. What do you think?

fluss-server/src/test/java/com/alibaba/fluss/server/zk/ZooKeeperClientTest.java
LiebingYu: Thank you for pointing out the issue here. However, we do need to call `register leaderAndIsr` first, otherwise there will be a "node does not exist" problem when updating. In the previous implementation, I didn't use the new `LeaderAndIsr` to test `batchUpdateLeaderAndIsr`, which resulted in the test passing even after deletion (`zookeeperClient.batchUpdateLeaderAndIsr(updateLeaderAndIsrList);`). Now I am using a new `LeaderAndIsr` to verify that `batchUpdateLeaderAndIsr` is indeed updated successfully.

fluss-server/src/test/java/com/alibaba/fluss/server/coordinator/CoordinatorEventProcessorTest.java
wuchong: Sorry, I didn't notice the `bucketEpoch + 1` in the `newLeaderAndIsr()`.

wuchong (approved):
LGTM.

wuchong:
Reruning the failed CI...

PR #1463, opened 2025-08-04
[server] server process metadata request in async mode && metadata request support cacheOnly flag

### LINKED ISSUE #1454: [server] improve performance of metadata request in tablet server
### Search before asking

### Motivation

Currently, the tablet server handles metadata() requests in synchronous mode, which creates a performance bottleneck in RPC request processing. Additionally, when the requested metadata is not available in the local cache, the server queries ZooKeeper but does not update the cache with the retrieved results.
In large clusters, when thousands of Flink coordinator clients request metadata from the server, it is easy to fall into a timeout and retry cycle, and server RPC resources will be exhausted with no relief in sight.

### Solution

- make metadata() process async in tablet server
- When metadata cache misses occur, the tablet server updates the local cache with the results retrieved from ZooKeeper.
- metadata() supports a cache-only flag that allows client requests to fail fast when cache misses occur, making it easy for clients to retry from another server.

### Anything else?

_No response_

### Willingness to contribute

platinumhamburg:
### Purpose

Linked issue: close #1454


fluss-server/src/main/java/org/apache/fluss/server/metadata/CoordinatorMetadataFunctionProvider.java
```diff
@@ -0,0 +1,139 @@
...
+import javax.annotation.Nullable;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+import java.util.concurrent.CompletableFuture;
+
+/** The coordinator metadata function provider. */
+public class CoordinatorMetadataFunctionProvider implements MetadataFunctionProvider {
```
wuchong: We can simplify the interface to `MetadataProvider` and `CoordinatorMetadataProvider`, `TabletServerMetadataProvider`.

fluss-server/src/main/java/org/apache/fluss/server/metadata/CoordinatorMetadataFunctionProvider.java
```diff
@@ -0,0 +1,139 @@
...
+
+    private final ZooKeeperClient zkClient;
+
+    private final CoordinatorMetadataCache metadataCache;
```
wuchong: The `CoordinatorMetadataCache` is not used here.

fluss-server/src/main/java/org/apache/fluss/server/metadata/CoordinatorMetadataFunctionProvider.java
```diff
@@ -0,0 +1,139 @@
...
+            MetadataManager metadataManager) {
+        this.zkClient = zkClient;
+        this.metadataCache = metadataCache;
+        this.ctx = ctx;
+        this.metadataManager = metadataManager;
+    }
+
+    @Override
+    public Optional<TableMetadata> getTableMetadataFromCache(TablePath tablePath) {
+        TableInfo tableInfo = metadataManager.getTable(tablePath);
```
wuchong: 1. we can move this after the `if` condition, to reduce the cost of zk operation. 
2. this is still reading from zk instead of from cache, I think we can get TableInfo from `ctx` as well

fluss-server/src/main/java/org/apache/fluss/server/metadata/CoordinatorMetadataFunctionProvider.java
```diff
@@ -0,0 +1,139 @@
...
+        }
+        bucketMetadataList =
+                getBucketMetadataFromContext(ctx, tableId, null, ctx.getTableAssignment(tableId));
+        return Optional.of(new TableMetadata(tableInfo, bucketMetadataList));
+    }
+
+    @Override
+    public CompletableFuture<TableMetadata> getTableMetadataFromZk(TablePath tablePath) {
+        TableInfo tableInfo = metadataManager.getTable(tablePath);
+        return RpcServiceBase.getTableMetadataFromZkAsync(
```
wuchong: `RpcServiceBase.getTableMetadataFromZkAsync` is a purely zk operation, and is not related to `RpcServiceBase`, we should put it in `ZookeeperClient`. The same to the `getTableMetadataFromZkAsync` and `batchGetPartitionMetadataFromZkAsync`.

fluss-server/src/main/java/org/apache/fluss/server/metadata/CoordinatorMetadataFunctionProvider.java
```diff
@@ -0,0 +1,139 @@
...
+    @Override
+    public Optional<PhysicalTablePath> getPhysicalTablePathFromCache(long partitionId) {
+        return ctx.getPhysicalTablePath(partitionId);
+    }
+
+    @Override
+    public Optional<PartitionMetadata> getPartitionMetadataFromCache(
+            PhysicalTablePath partitionPath) {
+        TablePath tablePath =
+                new TablePath(partitionPath.getDatabaseName(), partitionPath.getTableName());
```
wuchong:
```suggestion
        TablePath tablePath = partitionPath.getTablePath();
```

fluss-server/src/main/java/org/apache/fluss/server/metadata/TabletServerMetadataCache.java
```diff
@@ -122,18 +120,12 @@ public TableMetadata getTableMetadata(TablePath tablePath) {
...
-            bucketMetadataList =
-                    getTableMetadataFromZk(
-                            zkClient, tablePath, tableInfo.getTableId(), tableInfo.isPartitioned());
-        } else {
-            // get table assignment from cache.
-            bucketMetadataList =
-                    new ArrayList<>(
-                            snapshot.getBucketMetadataForTable(tableIdOpt.getAsLong()).values());
+
+            return null;
```
wuchong: Add `@Nullable` annotation to the return type of the method, or return `Optional` directly (it seems it is used as `Optional` later).

fluss-server/src/main/java/org/apache/fluss/server/metadata/TabletServerMetadataCache.java
```diff
@@ -155,9 +147,8 @@ public PartitionMetadata getPartitionMetadata(PhysicalTablePath partitionPath) {
                     partitionId,
                     new ArrayList<>(snapshot.getBucketMetadataForPartition(partitionId).values()));
         } else {
-            // TODO no need to get assignment from zk if refactor client metadata cache. Trace by
-            // https://github.com/apache/fluss/issues/483
-            return getPartitionMetadataFromZk(partitionPath, zkClient);
+
+            return null;
```
wuchong: Add `@Nullable` annotation to the return type of the method, or return `Optional` directly (it seems it is used as `Optional` later).

fluss-server/src/test/java/org/apache/fluss/server/RpcServiceBaseTest.java
```diff
@@ -0,0 +1,758 @@
...
+    }
+
+    // A test implementation of RpcServiceBase to access the protected processMetadataRequest method
+    private static class TestRpcServiceBase extends RpcServiceBase {
+        public TestRpcServiceBase() {
+            super(null, null, null, null, null);
+        }
+
+        @Override
+        public void processMetadataRequest(
```
wuchong: Why override this method? It seems we didn't do anything except calling parent method?

fluss-server/src/test/java/org/apache/fluss/server/RpcServiceBaseTest.java
```diff
@@ -0,0 +1,758 @@
...
+    @Test
+    void testGetTableMetadataFromZkAsync_Success() throws Exception {
+        // Setup test data
+        Map<Integer, BucketAssignment> bucketAssignments = new HashMap<>();
+        bucketAssignments.put(0, BucketAssignment.of(1, 2, 3));
+        bucketAssignments.put(1, BucketAssignment.of(2, 3, 4));
+        TableAssignment tableAssignment = new TableAssignment(bucketAssignments);
+
+        // Mock ZooKeeper responses
+        when(mockZkClient.getTableAssignment(testTableId)).thenReturn(Optional.of(tableAssignment));
```
wuchong: Please do not use Mockito, it increases maintenance overhead in the long term, which should be avoided. In this case, we can simply register metadata into zk without mocking the zk client.

fluss-server/src/test/java/org/apache/fluss/server/RpcServiceBaseTest.java
```diff
@@ -0,0 +1,758 @@
...
+        public void shutdown() {}
+
+        @Override
+        public CompletableFuture<MetadataResponse> metadata(MetadataRequest request) {
+            return null;
+        }
+    }
+
+    @Test
+    void testGetTableMetadataFromZkAsync_Success() throws Exception {
```
wuchong: If we decide to move `RpcServiceBase.getTableMetadataFromZkAsync` and related methods to `ZookeeperClient`, we should also move all the related test methods to `ZookeeperClientTest` where we can reuse the embedded zk cluster for testing.

fluss-server/src/test/java/org/apache/fluss/server/RpcServiceBaseTest.java
```diff
@@ -0,0 +1,758 @@
...
+                        tablePath,
+                        testTableId,
+                        1,
+                        DATA1_PARTITIONED_TABLE_DESCRIPTOR,
+                        System.currentTimeMillis(),
+                        System.currentTimeMillis());
+        TableMetadata tableMetadata =
+                new TableMetadata(partitionTableInfo, Collections.emptyList());
+        when(functionProvider.getTableMetadataFromCache(eq(tablePath)))
+                .thenReturn(Optional.of(tableMetadata));
```
wuchong: Please do no use Mockito, it increases maintenance overhead in the long term, which should be avoided. In this case, we can simply pass ` authorizer` as `null`, and create a testing implementation for `MetadataFunctionProvider` and `ServerMetadataCache`.

fluss-server/src/main/java/org/apache/fluss/server/RpcServiceBase.java
```diff
@@ -143,6 +155,16 @@ public abstract class RpcServiceBase extends RpcGatewayService implements AdminR
...
+    private static final ExecutorService ZK_META_UPDATE_EXECUTOR =
+            Executors.newFixedThreadPool(8, new ExecutorThreadFactory("zk-metadata-update-"));
+
+    private static final ConcurrentHashMap<
+                    TableMetadataKey, CompletableFuture<List<BucketMetadata>>>
+            PENDING_TABLE_METADATA_FROM_ZK_FUTURES = MapUtils.newConcurrentHashMap();
+
+    private static final ConcurrentHashMap<PhysicalTablePath, CompletableFuture<PartitionMetadata>>
+            PENDING_PARTITION_METADATA_FROM_ZK_FUTURES = MapUtils.newConcurrentHashMap();
+
```
wuchong: We should avoid introducing global resources which is hard to maintain (resource leak in production) and mock (override in tests). In order to support async zk operations, we can learn from Kafka that:

1. Use the default executor service in zk client (`curatorFramework.inBackground(...)` in our case)
2. Extend the `ZookeeperClient` to support **async batch** operations. We have already introduced a basic implementation in https://github.com/apache/fluss/commit/8b9d2937c35f47dce0d034785ee684d49ba04bc3, which is easy to extend for more methods. 
3. Use the **async batch** API of `ZookeeperClient` for fetching metadatas of table/partition lists.

wuchong:
@platinumhamburg why do we need to change the protogen maven plugin configuration?

platinumhamburg:
> @platinumhamburg why do we need to change the protogen maven plugin configuration?

During local development, I occasionally encounter build failures caused by fluss-rpc not generating proto files with the error message "Failed to parse plugin descriptor for org.apache.fluss:fluss-protogen-maven-plugin". It seems that the Maven build dependency graph is not very stable or explicit. So I tried to modify the Maven plugin configuration to standardize management of the protobuf plugin in my local environment, and it seems to work. However, it's better to submit changes unrelated to this PR separately, so I will remove this modification.

platinumhamburg:
> @platinumhamburg why do we need to change the protogen maven plugin configuration?

I have reverted all changes related to the Maven plugin configuration. Thank you for reviewing again when you have time.

wuchong:
I squashed the commits for preparing the merge. There is a detailed description in the second commit message to describe the commit. @platinumhamburg please take a look at the second commit.

wuchong:
rebased and resolve the conflicts.

platinumhamburg:
> I squashed the commits for preparing the merge. There is a detailed description in the second commit message to describe the commit. @platinumhamburg please take a look at the second commit.

I think your refinement looks very clear, and perhaps the handleRequestInBackground() method could also be implemented using FutureUtils::asyncToSync for consistency.

wuchong:
Thanks @platinumhamburg , I will merge this PR to avoid reruning the CI. Could you please open another pull request for the `handleRequestInBackground()` changes?

platinumhamburg:
> Thanks @platinumhamburg , I will merge this PR to avoid reruning the CI. Could you please open another pull request for the `handleRequestInBackground()` changes?

Got it.

PR #1466, opened 2025-08-04
[coordinator] coordinator report event enqueue rate and event process time labeled by event_type

### LINKED ISSUE #1455: [coordinator] Label the eventProcessTimeMs metric by event type
### Search before asking

### Motivation

Currently, the eventProcessTimeMs metric in the coordinator server has no event_type label, which makes it difficult to evaluate the event processing performance of different event types.

### Solution

Add a label to the eventProcessTimeMs metric. Additionally, add a new metric eventEnqueuePerSecond to track the enqueue rate of different events.

### Anything else?

_No response_

### Willingness to contribute

platinumhamburg:
### Purpose

Linked issue: close #1455

### DIFF
--- a/website/docs/maintenance/observability/monitor-metrics.md
+++ b/website/docs/maintenance/observability/monitor-metrics.md
@@ -327,7 +327,7 @@ Some metrics might not be exposed when using other JVM implementations (e.g. IBM
     </tr>
     <tr>
       <td>eventQueueSize</td>
-      <td>The number of events waiting to be processed in the queue.</td>
+      <td>The number of events currently waiting to be processed in the coordinator event queue. This metric is labeled with <code>event_type</code> to distinguish between different types of coordinator events.</td>
       <td>Gauge</td>
     </tr>
     <tr>
@@ -337,7 +337,7 @@ Some metrics might not be exposed when using other JVM implementations (e.g. IBM
     </tr>
     <tr>
       <td>eventProcessingTimeMs</td>
-      <td>The time that an event took to be processed.</td>
+      <td>The time that an event took to be processed by the coordinator event processor. This metric is labeled with <code>event_type</code> to distinguish between different types of coordinator events.</td>
       <td>Histogram</td>
     </tr>
   </tbody>
--- a/website/docs/maintenance/observability/monitor-metrics.md
+++ b/website/docs/maintenance/observability/monitor-metrics.md
@@ -295,7 +295,7 @@ Some metrics might not be exposed when using other JVM implementations (e.g. IBM
   <tbody>
     <tr>
       <th rowspan="9"><strong>coordinator</strong></th>
-      <td style={{textAlign: 'center', verticalAlign: 'middle' }} rowspan="9">-</td>
+      <td style={{textAlign: 'center', verticalAlign: 'middle' }} rowspan="7">-</td>
       <td>activeCoordinatorCount</td>
       <td>The number of active CoordinatorServer in this cluster.</td>
       <td>Gauge</td>
@@ -330,27 +330,8 @@ Some metrics might not be exposed when using other JVM implementations (e.g. IBM
       <td>The time that an event spent waiting in the queue to be processed.</td>
       <td>Histogram</td>
     </tr>
-  </tbody>
-</table>
-
-### Coordinator Event
-
-<table class="table table-bordered">
-  <thead>
     <tr>
-      <th class="text-left" style={{width: '30pt'}}>Scope</th>
-      <th class="text-left" style={{width: '150pt'}}>Infix</th>
-      <th class="text-left" style={{width: '80pt'}}>Metrics</th>
-      <th class="text-left" style={{width: '300pt'}}>Description</th>
-      <th class="text-left" style={{width: '40pt'}}>Type</th>
-    </tr>
-  </thead>
-  <tbody>
-    <tr>
-      <th rowspan="9"><strong>coordinator</strong></th>
-      <td style={{textAlign: 'center', verticalAlign: 'middle' }} rowspan="9">event</td>
-    </tr>
-    <tr>
+      <td style={{textAlign: 'center', verticalAlign: 'middle' }} rowspan="2">event</td>
       <td>eventQueueSize</td>
       <td>The number of events currently waiting to be processed in the coordinator event queue. This metric is labeled with <code>event_type</code> to distinguish between different types of coordinator events.</td>
       <td>Gauge</td>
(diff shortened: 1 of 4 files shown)


fluss-server/src/main/java/com/alibaba/fluss/server/coordinator/event/CoordinatorEventManager.java
```diff
@@ -53,7 +58,10 @@ public final class CoordinatorEventManager implements EventManager {
     private final Lock putLock = new ReentrantLock();
 
     // metrics
-    private Histogram eventProcessingTime;
+    private final Map<Class<? extends CoordinatorEvent>, Histogram> eventProcessTimeByType =
+            MapUtils.newConcurrentHashMap();
+    private final Map<Class<? extends CoordinatorEvent>, Meter> eventEnqueueRateByType =
+            MapUtils.newConcurrentHashMap();
```
wuchong: I think both of them can be declared as `HashMap`, because `eventProcessTimeByType` is accessed in the single-thread `CoordinatorEventThread`, and `eventEnqueueRateByType` is accessed in the `putLock` protection.

fluss-server/src/main/java/com/alibaba/fluss/server/coordinator/event/CoordinatorEventManager.java
```diff
@@ -121,6 +126,33 @@ public void clearAndPut(CoordinatorEvent event) {
...
     }
 
+    private Histogram getOrCreateEventProcessTimeHistogram(
+            Class<? extends CoordinatorEvent> eventType) {
+        return eventProcessTimeByType.computeIfAbsent(
+                eventType,
+                type -> {
+                    String eventTypeName = type.getSimpleName();
+                    MetricGroup eventTypeGroup =
+                            coordinatorMetricGroup.addGroup("event_type", eventTypeName);
```
wuchong: Instead of using `addGroup`, I'm thinking that could we use `variables` here? Then we can use variables to filter or combine metrics for different event types.

wuchong (commented):
Please also update documentation about the new introduced metrics, I noticed that all the metrics in `CoordinatorEventManager` is not documented.

fluss-server/src/main/java/com/alibaba/fluss/server/coordinator/event/CoordinatorEventManager.java
platinumhamburg: Thanks for your suggestion. I'll replace the `ConcurrentHashMap` with a `HashMap`.

fluss-server/src/main/java/com/alibaba/fluss/server/coordinator/event/CoordinatorEventManager.java
platinumhamburg: Currently, in our metric system implementation, all metrics within the same MetricGroup share a common variables map. This means that reported metrics cannot distinguish between different event_type values within the same MetricGroup.

The addGroup() method creates (if absent) a child MetricGroup, which inherits the variable map from its parent group while adding additional labels (such as event_type) for it. This design is intentional to support our implementation requirements.

Furthermore, with the current implementation, metrics can be filtered and aggregated as expected.

> Instead of using `addGroup`, I'm thinking that could we use `variables` here? Then we can use variables to filter or combine metrics for different event types.

platinumhamburg:
@wuchong Thanks for your review comments. I have resolved all the comments and updated the documentation for the newly added metrics.

fluss-server/src/main/java/com/alibaba/fluss/server/coordinator/event/CoordinatorEventManager.java
wuchong: I mean we can create different `MetricGroup` instances with event type in the variable maps. Such as a `class CoordinatorEventMetricGroup(EventType)` with event type as constructor parameter, and the parameter will be used in the variable maps. So that all the event reports the same metric name, but with different variables (labels). 

IIUC, the current implementation can't aggregate different event types in metric system, as the are different metric names.

wuchong: It seems you have refactored in the way I mentioned above 👍

website/docs/maintenance/observability/monitor-metrics.md
```diff
@@ -337,7 +337,7 @@ Some metrics might not be exposed when using other JVM implementations (e.g. IBM
     </tr>
     <tr>
       <td>eventProcessingTimeMs</td>
-      <td>The time that an event took to be processed.</td>
+      <td>The time that an event took to be processed by the coordinator event processor. This metric is labeled with <code>event_type</code> to distinguish between different types of coordinator events.</td>
```
wuchong, line 341: I think currently the `eventProcessingTimeMs` is not directly under the `coordinator` group, but has an infix of `event`. Could you verify that and update the table?

website/docs/maintenance/observability/monitor-metrics.md
```diff
@@ -327,7 +327,7 @@ Some metrics might not be exposed when using other JVM implementations (e.g. IBM
     </tr>
     <tr>
       <td>eventQueueSize</td>
-      <td>The number of events waiting to be processed in the queue.</td>
+      <td>The number of events currently waiting to be processed in the coordinator event queue. This metric is labeled with <code>event_type</code> to distinguish between different types of coordinator events.</td>
```
wuchong: We should also update the `eventQueueSize`. Currently there are 2 metrics for `eventQueueSize`, one is directly under `coordinator` which reports the total queue size without `event_type` label. The other reports under the `event` infix with the `event_type` label.

website/docs/maintenance/observability/monitor-metrics.md
platinumhamburg: Thanks for your review comments, I reorganized the modified metrics in coordinator's  event subgroup.

website/docs/maintenance/observability/monitor-metrics.md
platinumhamburg: > We should also update the `eventQueueSize`. Currently there are 2 metrics for `eventQueueSize`, one is directly under `coordinator` which reports the total queue size without `event_type` label. The other reports under the `event` infix with the `event_type` label.

I fixed the infix issue of modified metrics. Besides, currently only have one metric for eventQueueSize that labeled by <code>event_type</code> now, because the total queue size can aggregate sum with the labeled metric.

wuchong:
I pushed a commit to update the metrics documentation. Currently, the coordinator metrics spans in 2 sections, I merged them into one. 

**Before:**
image

**After:**
image

wuchong (approved):

Issue #1304, opened 2025-07-11
TabletServer may discard already commited kv snapshot


luoyuxia:
### Search before asking

- [x] I searched in the [issues](https://github.com/alibaba/fluss/issues) and found nothing similar.

### Fluss version

main (development)

### Please describe the bug 🐞

1. tablet server try to commit kv snapshot to coordinator server
2. coordinator server commit the kv snapshot to zk, then failover
3. tablet server will got exception from coordinator server, but mistake it as a fail commit although coordinator server has committed to zk, then discard the commited kv snapshot

### Solution

We need to follow how tablet server commit log manifest, making coordinator server return error/success/unknow status explictly in response, so that tablet server can discard the snapshot if error instead of always discard the snapshot while the snapshot has been commited to coordinator..

### Are you willing to submit a PR?

- [ ] I'm willing to submit a PR!

wuchong:
Even if the coordinator server return error/success/unknow status explictly in response, when the coordinator fails, the tablet server can't receive any response from the coodinator, then what should the tablet server do? Should it discard the snapshot?

luoyuxia:
> Even if the coordinator server return error/success/unknow status explictly in response, when the coordinator fails, the tablet server can't receive any response from the coodinator, then what should the tablet server do? Should it discard the snapshot?

No, it shouldn't. But If the snapshot is 4, but if `RocksIncrementalSnapshot` do next snapshot 5, it should consider this as an unsuccessful snapshot and include ssts in snapshot 4 in snapshot 5.

PR #1470, opened 2025-08-04
[kv] coordinator must handle inconsistency or broken snapshot scenarios

### LINKED ISSUE #1469: [KV] Coordinator Permanently Rejects KV Snapshot Commit Upon Encountering Corrupted History KV Snapshot
### Search before asking

### Fluss version

0.7.0 (latest release)

### Please describe the bug 🐞

Current Situation: The bug described in issue https://github.com/apache/fluss/issues/1304 may cause inconsistent snapshot metadata between remote storage and ZooKeeper. This inconsistency prevents the affected table buckets from committing new KV snapshots.

### Solution

The Coordinator must handle inconsistency or broken snapshot scenarios.

platinumhamburg:
### Purpose

Linked issue: close #1469

### DIFF
--- a/fluss-server/src/main/java/org/apache/fluss/server/coordinator/CompletedSnapshotStoreManager.java
+++ b/fluss-server/src/main/java/org/apache/fluss/server/coordinator/CompletedSnapshotStoreManager.java
@@ -17,6 +17,7 @@
 
 package org.apache.fluss.server.coordinator;
 
+import org.apache.fluss.annotation.VisibleForTesting;
 import org.apache.fluss.metadata.TableBucket;
 import org.apache.fluss.server.kv.snapshot.CompletedSnapshot;
 import org.apache.fluss.server.kv.snapshot.CompletedSnapshotHandle;
@@ -37,6 +38,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.Executor;
+import java.util.function.Function;
 
 import static org.apache.fluss.utils.Preconditions.checkArgument;
 import static org.apache.fluss.utils.Preconditions.checkNotNull;
@@ -50,11 +52,12 @@ import static org.apache.fluss.utils.Preconditions.checkNotNull;
 public class CompletedSnapshotStoreManager {
 
     private static final Logger LOG = LoggerFactory.getLogger(CompletedSnapshotStoreManager.class);
-
     private final int maxNumberOfSnapshotsToRetain;
     private final ZooKeeperClient zooKeeperClient;
     private final Map<TableBucket, CompletedSnapshotStore> bucketCompletedSnapshotStores;
     private final Executor ioExecutor;
+    private final Function<ZooKeeperClient, CompletedSnapshotHandleStore>
+            makeZookeeperCompletedSnapshotHandleStore;
 
     public CompletedSnapshotStoreManager(
             int maxNumberOfSnapshotsToRetain,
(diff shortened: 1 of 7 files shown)


rafaelsmf:
Thanks for the fixing! I was testing Fluss locally and after the zookeeper container restart, I faced the issue related to this PR.

fluss-server/src/main/java/com/alibaba/fluss/server/coordinator/CompletedSnapshotStoreManager.java
```diff
@@ -120,13 +121,46 @@ private CompletedSnapshotStore createCompletedSnapshotStore(
...
 
         for (CompletedSnapshotHandle snapshotStateHandle : initialSnapshots) {
-            retrievedSnapshots.add(checkNotNull(snapshotStateHandle.retrieveCompleteSnapshot()));
+            try {
+                retrievedSnapshots.add(
+                        checkNotNull(snapshotStateHandle.retrieveCompleteSnapshot()));
+            } catch (Exception e) {
+                if (e.getMessage().contains(SNAPSHOT_META_DATA_NOT_EXISTS_ERROR_MESSAGE)) {
+                    LOG.error(
+                            "metadata not found for snapshot {} of table bucket {}, maybe snapshot already removed or broken, detail errorMsg: {}",
```
wuchong: minor: Let's capitalize the first letter of log messages to maintain consistency with the project's logging style.

fluss-server/src/main/java/com/alibaba/fluss/server/coordinator/CompletedSnapshotStoreManager.java
```diff
@@ -120,13 +121,46 @@ private CompletedSnapshotStore createCompletedSnapshotStore(
...
+                            snapshotStateHandle.getSnapshotId(),
+                            tableBucket,
+                            e.getMessage());
```
wuchong: Could we also print the full exception stack trace? I'm concerned that the root cause might be hidden within nested exceptions.

```suggestion
                            e.getMessage(), 
                            e);
```

fluss-server/src/main/java/com/alibaba/fluss/server/kv/snapshot/CompletedSnapshotHandle.java
```diff
@@ -40,11 +40,13 @@
  */
 public class CompletedSnapshotHandle {
 
+    private long snapshotId;
```
wuchong: use `final` modifier to mark this field is immutable.

wuchong (commented):
Could you add a test to reproduce this problem and verify the bug is fixed? Maybe a unit test in `CompletedSnapshotStoreManagerTest`.

wuchong:
I saw you mentioned the https://github.com/apache/fluss/issues/1304 in the issue description, can this pull request fix the problem of #1304?

platinumhamburg:
> I saw you mentioned the #1304 in the issue description, can this pull request fix the problem of #1304?

This PR cannot be used to close issue #1304 at this time. Issue #1304 may introduce broken snapshots or inconsistent snapshot metadata, while this PR can only handle already existing inconsistencies. I will submit another PR to fix #1304 later.

platinumhamburg:
> Could you add a test to reproduce this problem and verify the bug is fixed? Maybe a unit test in `CompletedSnapshotStoreManagerTest`.

Thank you for your feedback. I have addressed all of your suggestions and supplemented the test cases to ensure correctness.

fluss-server/src/test/java/com/alibaba/fluss/server/coordinator/CompletedSnapshotStoreManagerTest.java
```diff
@@ -172,6 +181,50 @@ void testRemoveCompletedSnapshotStoreFromManager() throws Exception {
...
+        when(validSnapshotHandle.getMetadataFilePath())
+                .thenReturn(validSnapshot.getMetadataFilePath());
+        when(validSnapshotHandle.retrieveCompleteSnapshot()).thenReturn(validSnapshot);
+
+        CompletedSnapshot invalidSnapshot =
+                KvTestUtils.mockCompletedSnapshot(tempDir, tableBucket, 2L);
+        CompletedSnapshotHandle invalidSnapshotHandle = mock(CompletedSnapshotHandle.class);
+        when(invalidSnapshotHandle.getSnapshotId()).thenReturn(invalidSnapshot.getSnapshotID());
+        when(invalidSnapshotHandle.getMetadataFilePath())
+                .thenReturn(invalidSnapshot.getMetadataFilePath());
```
wuchong: Please avoid use Mockito, use reusable test implementations instead. (see [the code quality guide and reason here](https://flink.apache.org/how-to-contribute/code-style-and-quality-common/#avoid-mockito---use-reusable-test-implementations)). 

Here, we can easily construct the valid and invalid `CompletedSnapshotHandle` using its constructor, one has valid path and the other has invlid path. This makes the code easy to maintain for the future, and cover more code path.

wuchong (commented):
Thanks @platinumhamburg for the fixing. 

Just a quick note: could you please avoid squashing commits while responding to review comments? Keeping the changes in separate commits makes it much easier to track what has been updated since the last review.

fluss-server/src/test/java/com/alibaba/fluss/server/coordinator/CompletedSnapshotStoreManagerTest.java
platinumhamburg: Thanks for the suggestion. I've refactored this case to avoid using Mockito.

wuchong (approved):

PR #1477, opened 2025-08-05
Use background mode for delete partition assignment in zk

### LINKED ISSUE #1476: Optimize the handling of DeleteReplicaResponseReceivedEvent
### Search before asking

### Description

Currently, We found that the processing of `DeleteReplicaResponseReceivedEvent` can be blocked for a long time in some cases, especially when handling partitions of KV tables.
Image

After some debug, I found the problem is at `ZooKeeperClient#deletePartitionAssignment`. This method will recursively delete all the children for the partition assignment ZNode which may spend a lot of time. In my local tests, deleting a parent node with three levels of child nodes (containing 1,024 first-level child nodes, each of which has 4 second-level child nodes, and each second-level child node has 5 third-level child nodes) takes more than 5 seconds.
```java
    @Test
    void test() throws Exception {
        String basePath = "/perfTest";
        int firstLevel = 1024;
        int secondLevel = 4;
        int thirdLevel = 5;

        CuratorFramework client = zookeeperClient.getCuratorClient();

        for (int i = 0; i < firstLevel; i++) {
            String level1 = basePath + "/n1_" + i;
            client.create().creatingParentsIfNeeded().forPath(level1);
            for (int j = 0; j < secondLevel; j++) {
                String level2 = level1 + "/n2_" + j;
                client.create().creatingParentsIfNeeded().forPath(level2);
                for (int k = 0; k < thirdLevel; k++) {
                    String level3 = level2 + "/n3_" + k;
                    client.create().creatingParentsIfNeeded().forPath(level3, ("dummy").getBytes());
                }
            }
        }

        long start = System.currentTimeMillis();
        client.delete()
                .deletingChildrenIfNeeded()
                .forPath(basePath);
        long end = System.currentTimeMillis();
        System.out.println("Delete finished. Cost: " + (end - start) + " ms");
    }
```

For KV tables, there are many snapshot nodes, resulting in more child nodes compared to Log tables. Therefore, the deletion event takes longer, and this issue is more pronounced for KV tables.
Image

### Willingness to contribute

LiebingYu:
### Purpose

Linked issue: close #1476

### DIFF
--- a/fluss-server/src/main/java/org/apache/fluss/server/coordinator/CoordinatorEventProcessor.java
+++ b/fluss-server/src/main/java/org/apache/fluss/server/coordinator/CoordinatorEventProcessor.java
@@ -176,7 +176,8 @@ public class CoordinatorEventProcessor implements EventProcessor {
                         coordinatorContext,
                         replicaStateMachine,
                         tableBucketStateMachine,
-                        new RemoteStorageCleaner(conf, ioExecutor));
+                        new RemoteStorageCleaner(conf, ioExecutor),
+                        ioExecutor);
         this.tableChangeWatcher = new TableChangeWatcher(zooKeeperClient, coordinatorEventManager);
         this.tabletServerChangeWatcher =
                 new TabletServerChangeWatcher(zooKeeperClient, coordinatorEventManager);
--- a/fluss-server/src/main/java/org/apache/fluss/server/coordinator/CoordinatorEventProcessor.java
+++ b/fluss-server/src/main/java/org/apache/fluss/server/coordinator/CoordinatorEventProcessor.java
@@ -168,7 +168,7 @@ public class CoordinatorEventProcessor implements EventProcessor {
                                 coordinatorEventManager,
                                 coordinatorContext),
                         zooKeeperClient);
-        this.metadataManager = new MetadataManager(zooKeeperClient, conf, ioExecutor);
+        this.metadataManager = new MetadataManager(zooKeeperClient, conf);
 
         this.tableManager =
                 new TableManager(
--- a/fluss-server/src/main/java/org/apache/fluss/server/coordinator/CoordinatorEventProcessor.java
+++ b/fluss-server/src/main/java/org/apache/fluss/server/coordinator/CoordinatorEventProcessor.java
@@ -168,7 +168,7 @@ public class CoordinatorEventProcessor implements EventProcessor {
                                 coordinatorEventManager,
                                 coordinatorContext),
                         zooKeeperClient);
-        this.metadataManager = new MetadataManager(zooKeeperClient, conf);
+        this.metadataManager = new MetadataManager(zooKeeperClient, conf, ioExecutor);
 
         this.tableManager =
                 new TableManager(
--- a/fluss-server/src/main/java/org/apache/fluss/server/coordinator/CoordinatorServer.java
+++ b/fluss-server/src/main/java/org/apache/fluss/server/coordinator/CoordinatorServer.java
@@ -183,7 +183,7 @@ public class CoordinatorServer extends ServerBase {
 
             this.lakeTableTieringManager = new LakeTableTieringManager();
 
-            MetadataManager metadataManager = new MetadataManager(zkClient, conf, ioExecutor);
+            MetadataManager metadataManager = new MetadataManager(zkClient, conf);
             this.coordinatorService =
                     new CoordinatorService(
                             conf,
@@ -321,7 +321,7 @@ public class CoordinatorServer extends ServerBase {
     }
 
     private void createDefaultDatabase() {
-        MetadataManager metadataManager = new MetadataManager(zkClient, conf, ioExecutor);
+        MetadataManager metadataManager = new MetadataManager(zkClient, conf);
         List<String> databases = metadataManager.listDatabases();
         if (databases.isEmpty()) {
             metadataManager.createDatabase(DEFAULT_DATABASE, DatabaseDescriptor.EMPTY, true);
--- a/fluss-server/src/main/java/org/apache/fluss/server/coordinator/CoordinatorServer.java
+++ b/fluss-server/src/main/java/org/apache/fluss/server/coordinator/CoordinatorServer.java
@@ -183,7 +183,7 @@ public class CoordinatorServer extends ServerBase {
 
             this.lakeTableTieringManager = new LakeTableTieringManager();
 
-            MetadataManager metadataManager = new MetadataManager(zkClient, conf);
+            MetadataManager metadataManager = new MetadataManager(zkClient, conf, ioExecutor);
             this.coordinatorService =
                     new CoordinatorService(
                             conf,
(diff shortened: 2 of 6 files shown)


fluss-server/src/main/java/com/alibaba/fluss/server/zk/ZooKeeperClient.java
```diff
@@ -222,8 +222,27 @@ public void deleteTableAssignment(long tableId) throws Exception {
 
     public void deletePartitionAssignment(long partitionId) throws Exception {
         String path = PartitionIdZNode.path(partitionId);
-        zkClient.delete().deletingChildrenIfNeeded().forPath(path);
-        LOG.info("Deleted table assignment for partition id {}.", partitionId);
+        // delete partition assignment ZNode will recursively delete all the children which may
```
platinumhamburg: Could this introduce inconsistent behavior during coordinator failover? Should we consider using ZooKeeper TransactionOp to solve this problem? I think dividing the ZK delete operation into asynchronous background sub-tasks may compromise the integrity of the DROP semantic. @wuchong What are your thoughts on this?

LiebingYu: > Could this introduce inconsistent behavior during coordinator failover? Should we consider using ZooKeeper TransactionOp to solve this problem? I think dividing the ZK delete operation into asynchronous background sub-tasks may compromise the integrity of the DROP semantic. @wuchong What are your thoughts on this?

Recursively delete children nodes cannot execute in TransactionOp. To ensure that PartitionAssignment can be completely removed during CoordinatorServer failover, I only remove the partition to be deleted from the CoordinatorContext after receiving a successful ZooKeeper deletion callback. In this way, if a failure occurs during the deletion process, it can be retried after a restart.

platinumhamburg: @LB-Yu Thank you for your detailed explanation. The design approach you've outlined  appears to be well-reasoned. I believe this PR is in good shape and I will proceed with testing it shortly.

platinumhamburg:
@LB-Yu I've tested this PR and found that it significantly improves the processing performance of Replica DeleteReplicaResponseReceivedEvent.

fluss-server/src/main/java/com/alibaba/fluss/server/zk/ZooKeeperClient.java
LiebingYu: > > Could this introduce inconsistent behavior during coordinator failover? Should we consider using ZooKeeper TransactionOp to solve this problem? I think dividing the ZK delete operation into asynchronous background sub-tasks may compromise the integrity of the DROP semantic. @wuchong What are your thoughts on this?
> 
> Recursively delete children nodes cannot execute in TransactionOp. To ensure that PartitionAssignment can be completely removed during CoordinatorServer failover, I only remove the partition to be deleted from the CoordinatorContext after receiving a successful ZooKeeper deletion callback. In this way, if a failure occurs during the deletion process, it can be retried after a restart.

After some testing, I abandoned the "only remove the partition to be deleted from the CoordinatorContext after receiving a successful ZooKeeper deletion callback" solution. It causes onDeletePartition() to be called multiple times.

wuchong:
Sorry, I do have some concerns about this pull request. I left the comments below:

1. Why moving it to background can improve the ZK performance? Doesn't the background execution also require recursively deleting all the children for the partition assignment? Could you share the benchmark result before and after using the background? 

2. I think this changes the semantic of `deletePartitionAssignment` which means all the assignment has been deleted after the method returns. This may introduce some unexpected behavior when using this method. For example: `TableManager#completeDeletePartition` removes the partition from `coordinatorContext` only when the assignement has been deleted in zk. But this method becomes async, and it will remove the partition from `coordinatorContext` before all the assignment is deleted from zk. 

3. We have a lot of `deletingChildrenIfNeeded` in `ZooKeeperClient`, should we also update other methods as well? Otherwise, it introduces in-consistent in the code. This makes future development confusing that which way is suggested to use.

LiebingYu:
> Sorry, I do have some concerns about this pull request. I left the comments below:
> 
> 1. Why moving it to background can improve the ZK performance? Doesn't the background execution also require recursively deleting all the children for the partition assignment? Could you share the benchmark result before and after using the background?
> 2. I think this changes the semantic of `deletePartitionAssignment` which means all the assignment has been deleted after the method returns. This may introduce some unexpected behavior when using this method. For example: `TableManager#completeDeletePartition` removes the partition from `coordinatorContext` only when the assignement has been deleted in zk. But this method becomes async, and it will remove the partition from `coordinatorContext` before all the assignment is deleted from zk.
> 3. We have a lot of `deletingChildrenIfNeeded` in `ZooKeeperClient`, should we also update other methods as well? Otherwise, it introduces in-consistent in the code. This makes future development confusing that which way is suggested to use.

1. Use backgroud cannot improve the ZK performance. But it avoid to block `processDeleteReplicaResponseReceived()`, which may take several seconds to complete. For KV tables with many snapshot nodes, the deletion event takes longer.
2. It does introduce a certain degree of inconsistency, but eventual consistency should be guaranteed. I'm not sure whether this eventual consistency could cause any issues.
3. In `deleteTableAssignment()` and `deletePartitionAssignment()` we better use background mode. Because the parent node may have a lot of children. However, deleting a table is not a common operation, so we might be able to ignore it for now.

wuchong:
Thank you for the response, @LiebingYu. I understand the context of this issue and would like to suggest the following improvements:

- It seems the default executor service of curator `inBackground()` uses a single-threaded executor, which could become a performance bottleneck under a high volume of partition deletion requests. To address this, we can leverage the coordinator’s `ioExecutor` to handle these operations asynchronously, which is better suited for I/O-bound tasks.
  
- We already perform the deletion of remote data directories asynchronously in `TableManager#completeDeletePartition`. We should extend this pattern by also making the deletion of ZK assignment asynchronous, either in `TableManager#completeDeletePartition` or `MetadataManager#completeDeletePartition`.

- Currently, even if the ZK assignment deletion fails, we proceed to remove the partition from `CoordinatorContext`. Given this behavior, making the ZK deletion asynchronous should not introduce additional risk, as the system already assumes best-effort cleanup.

- For consistency, we should apply the same async pattern to `MetadataManager#completeDeleteTable`.

- To implement this, we can pass the `ioExecutor` into `MetadataManager` for executing the async operations, without requiring any changes to the `ZookeeperClient` itself.

Let me know what you think.

LiebingYu:
@wuchong Thaks for the advice. I agree to keep the same async pattern to `MetadataManager#completeDeleteTable` and explicitly use `ioExecutor` instead of the default executor service of curator `inBackground()`.

wuchong (approved):
@LiebingYu thanks for the updates. 

I found that adding `ioExecutor` to `MetadataManager` involves too many changes. I updated the PR to add `ioExecutor` to the `TableManager`, and also updated the logic of table deletion. This looks simpler to maintain.

LiebingYu:
@wuchong fix UT for async deletion.

wuchong:
thanks for the fixing @LiebingYu

PR #1482, opened 2025-08-06
[kv] TabletSever must handle inconsistency or broken snapshot scenarios.

### LINKED ISSUE #1481: [KV] Tablet Server KvTablet Initialization Failure Due to Corrupted History KV Snapshot
### Search before asking

### Fluss version

0.7.0 (latest release)

### Please describe the bug 🐞

https://github.com/apache/fluss/issues/1304 may cause broken snapshot data even when the CompletedSnapshot has been committed to ZooKeeper. This can prevent the replica on the tablet server from recovering and switching to leader indefinitely due to KV snapshot download failures.

### Solution

The TabletSever also must handle inconsistency or broken snapshot scenarios.

platinumhamburg:
### Purpose

Linked issue: close #1481

### DIFF
--- a/fluss-server/src/test/java/org/apache/fluss/server/replica/ReplicaTest.java
+++ b/fluss-server/src/test/java/org/apache/fluss/server/replica/ReplicaTest.java
@@ -31,7 +31,6 @@ import org.apache.fluss.record.MemoryLogRecords;
 import org.apache.fluss.server.entity.NotifyLeaderAndIsrData;
 import org.apache.fluss.server.kv.KvTablet;
 import org.apache.fluss.server.kv.snapshot.CompletedSnapshot;
-import org.apache.fluss.server.kv.snapshot.KvFileHandleAndLocalPath;
 import org.apache.fluss.server.kv.snapshot.TestingCompletedKvSnapshotCommitter;
 import org.apache.fluss.server.log.FetchParams;
 import org.apache.fluss.server.log.LogAppendInfo;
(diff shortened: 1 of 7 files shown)


fluss-server/src/test/java/org/apache/fluss/server/replica/ReplicaTest.java
```diff
@@ -604,6 +704,39 @@ private void verifyGetKeyValues(
...
 
+    /**
+     * Simulate a broken snapshot by deleting its local metadata file while keeping ZK metadata.
+     * This simulates the scenario where snapshot files are corrupted or accidentally deleted but
+     * ZooKeeper still has the snapshot metadata.
+     */
+    private void simulateBrokenSnapshot(CompletedSnapshot snapshot) throws Exception {
+        // Delete the snapshot metadata file (_METADATA) to simulate file corruption
+        File metadataFile = new File(snapshot.getMetadataFilePath().getPath());
+        if (metadataFile.exists()) {
```
wuchong: The test cases can't reproduce the exception mentioned in the issue. 

That's because this method doesn't correctly delete metadata and SST files, because the file name of snapshot metadata&SST files uses UUID in remote storage. `_METADATA` is only used when the files are in local. So all the condition of `exists()` returns false. 

We can simplify use `snapshot.getKvSnapshotHandle().discard();` to simulate the broken snapshot. It will delete all the metadata files and unshared SST files of this snapshot.

wuchong (approved):
I added a commit to resolve the mentioned problem in the test. Will merge it once CI is passed.

Issue #1161, opened 2025-06-19
[Umbrella] Add backward and forward compatibility E2E tests


wuchong:
### Search before asking

- [x] I searched in the [issues](https://github.com/alibaba/fluss/issues) and found nothing similar.

### Motivation

We should guarantee the backward and forward compatibility for both client and server. That means:
- 0.7 client can work with 0.6- and 0.8+ servers.
- 0.7 server (TabletServer) can work with 0.6- and 0.8+ servers (CoordinatorServer) which is the rolling upgrade case

### Willingness to contribute

- [ ] I'm willing to submit a PR!

wuchong:
cc @swuferhong

swuferhong:
Hi, @wuchong. Could you assign this issue to me? I already have a basic PR ready. Considering that version 0.8 modified the package path, we may need to drop compatibility support for versions 0.6 and 0.7 to keep the framework implementation clean and simple. Since the 0.8 image has been released, I'll handle compatibility verification for 0.8 going forward.

PR #1485, opened 2025-08-06
[lake] Store partition name

luoyuxia:
### Purpose

Linked issue: close #xxx


luoyuxia:
@xx789633 Could you please help review this pr since it's a continue work of yours.

fluss-flink/fluss-flink-common/src/main/java/com/alibaba/fluss/flink/tiering/committer/TieringCommitOperator.java
```diff
@@ -203,32 +203,87 @@ private Committable commitWriteResults(
...
+                    && tableBucketWriteResult.partitionName() != null) {
+                partitionNameById.put(
+                        tableBucketWriteResult.tableBucket().getPartitionId(),
+                        tableBucketWriteResult.partitionName());
+            }
+        }
+
+        List<String> partitionKeys = new ArrayList<>();
+        if (!partitionNameById.isEmpty()) {
+            partitionKeys = admin.getTableInfo(tablePath).get().getPartitionKeys();
```
xx789633, line 254: In `checkFlussNotMissingLakeSnapshot`, we also call `admin.getTableInfo(tablePath)`. Can we combine these rpc calls to reduce cost?

fluss-flink/fluss-flink-common/src/main/java/com/alibaba/fluss/flink/tiering/committer/TieringCommitOperator.java
```diff
@@ -203,32 +203,87 @@ private Committable commitWriteResults(
...
+                String partitionQualifiedName = null;
+                if (partitionId != null) {
+                    // the partitionName is 2025$12$03, we need to convert to
+                    // qualified name year=2025/month=12/day=03
+                    partitionQualifiedName =
+                            ResolvedPartitionSpec.fromPartitionName(
+                                            partitionKeys, partitionNameById.get(partitionId))
+                                    .getPartitionQualifiedName();
+                }
+                partitionNameById.get(entry.getKey().getPartitionId());
```
xx789633: This is useless?

fluss-flink/fluss-flink-common/src/main/java/com/alibaba/fluss/flink/tiering/committer/FlussTableLakeSnapshotCommitter.java
```diff
@@ -114,15 +114,19 @@ private CommitLakeTableSnapshotRequest toCommitLakeTableSnapshotRequest(
...
             PbLakeTableOffsetForBucket pbLakeTableOffsetForBucket =
                     pbLakeTableSnapshotInfo.addBucketsReq();
-            TableBucket tableBucket = bucketEndOffsetEntry.getKey();
+            TableBucket tableBucket = bucketEndOffsetEntry.getKey().f0;
+            String partitionName = bucketEndOffsetEntry.getKey().f1;
             long endOffset = bucketEndOffsetEntry.getValue();
             if (tableBucket.getPartitionId() != null) {
                 pbLakeTableOffsetForBucket.setPartitionId(tableBucket.getPartitionId());
             }
+            if (partitionName != null) {
```
xx789633, line 127: Is it possible that partitionId is null while partitionName is not null?

fluss-common/src/main/java/com/alibaba/fluss/metadata/ResolvedPartitionSpec.java
```diff
@@ -125,6 +125,25 @@ public String getPartitionQualifiedName() {
         return sb.toString();
     }
 
+    public static ResolvedPartitionSpec fromPartitionQualifiedName(String qualifiedPartitionName) {
```
xx789633, line 128: Can this function handle the partition name string that contains special character "/"?

fluss-client/src/main/java/com/alibaba/fluss/client/metadata/LakeSnapshot.java
```diff
@@ -36,9 +36,17 @@ public class LakeSnapshot {
     // the specific log offset of the snapshot
     private final Map<TableBucket, Long> tableBucketsOffset;
 
-    public LakeSnapshot(long snapshotId, Map<TableBucket, Long> tableBucketsOffset) {
+    // the partition name by partition id of this lake snapshot if
+    // is a partitioned table, empty if not a partitioned table
+    private final Map<Long, String> partitionNameById;
```
xx789633, line 42: Can we wrap partition name in `tableBucketsOffset`'s key? Since partition id is duplicated in `tableBucketsOffset` and `partitionNameById`.

fluss-flink/fluss-flink-common/src/main/java/com/alibaba/fluss/flink/tiering/committer/TieringCommitOperator.java
luoyuxia: We can combine these rpc calls, but I feel it makes code more complex. And considering it's a corner case that ` TableInfo tableInfo = admin.getTableInfo(tablePath).get();` in method `checkFlussNotMissingLakeSnapshot` is called since it's only happen when some exception happens. So, I think it's acceptable to not merge in here.

fluss-flink/fluss-flink-common/src/main/java/com/alibaba/fluss/flink/tiering/committer/TieringCommitOperator.java
luoyuxia: Good catch. Removed now.

fluss-flink/fluss-flink-common/src/main/java/com/alibaba/fluss/flink/tiering/committer/FlussTableLakeSnapshotCommitter.java
luoyuxia: No, it's impossible. In any case it happens, it should be a bug.

xx789633: Since that, we can move the logic into the upper branch `if (tableBucket.getPartitionId() != null)`

fluss-common/src/main/java/com/alibaba/fluss/metadata/ResolvedPartitionSpec.java
luoyuxia: No, it can't. But as for a I know, partition name mostly likely won't contain special character `/` since it'll conflict with the file path separator making impossible to figure out what's the real partition.
I don't think any lake will allow this since lake will store partition by file path. At least, paimon will escape `/` in partition name.

fluss-flink/fluss-flink-common/src/main/java/com/alibaba/fluss/flink/tiering/committer/FlussTableLakeSnapshotCommitter.java
luoyuxia: But it may still happen that partition is not null, but partition name is null in a old version of tiering service..

fluss-client/src/main/java/com/alibaba/fluss/client/metadata/LakeSnapshot.java
luoyuxia: If wrap `partition name` in `tableBucketsOffset`, the `tableBucketsOffset` may look like `Map<Tuple2<TableBucket, String>, Long> tableBucketsOffset`, then in method 
`Map<TableBucket, Long> getTableBucketsOffset` and `Map<Long, String> getPartitionNameById` we will need to convert it into `Map<TableBucket, Long> `, `Map<Long, String> ` which I think a little of heavy. 
So, I tend to keep the duplication, and I think the duplication is acceptable, won't bring much memory consuming.

Maybe in the future, we can consider another refactor to make `TableBucket` contains the partition name to simpify the caller logic.

luoyuxia:
@xx789633 Thanks for your review and comments. PTAL again.

fluss-flink/fluss-flink-common/src/main/java/com/alibaba/fluss/flink/tiering/committer/FlussTableLakeSnapshotCommitter.java
xx789633: Makes sense.

xx789633: Just wondering whether we have any version compatibility test for now?

luoyuxia: Not yet. Actually, version compatibility test is lack for all the features of Fluss. There's a issue #1161 to track it.

fluss-common/src/main/java/com/alibaba/fluss/metadata/ResolvedPartitionSpec.java
xx789633: In Fluss, there is `InvalidPartitionException`. If a user attempts to create a partition `a/b/c/`, such an exception with the error message "contains one or more characters other than ASCII alphanumerics" will be thrown.

fluss-flink/fluss-flink-common/src/main/java/com/alibaba/fluss/flink/tiering/source/TableBucketWriteResultSerializer.java
```diff
@@ -65,6 +65,7 @@ public byte[] serialize(TableBucketWriteResult<WriteResult> tableBucketWriteResu
         if (tableBucket.getPartitionId() != null) {
             out.writeBoolean(true);
             out.writeLong(tableBucket.getPartitionId());
+            out.writeUTF(tableBucketWriteResult.partitionName());
```
xx789633, line 68: Do we need to consider compatibility here? The partition id is not null, while partition name is null.

luoyuxia: No, we don't need. The whole tiering service is stateless, and  the table bucket write result is only for network transmission from tiering writer operator to tiering commit operator which must be all old version or all new version.

xx789633:
Looks good to me. I think in the future we can move partitionName into TableBucket to make our data structures more compacted.

PR #1487, opened 2025-08-06
[server] handle uncaught exceptions in TabletServer during log recovery from residual data corresponding to already dropped tables

### LINKED ISSUE #1486: [server] TabletServer could not start because it attempted to recover tablet logs from residual data of already dropped tables.
### Search before asking

### Fluss version

0.7.0 (latest release)

### Please describe the bug 🐞

When a table is dropped on the CoordinatorServer, the table metadata is removed from ZooKeeper. TabletServers then perform replica cleanup actions asynchronously. However, if any TabletServer restarts before its local bucket data cleanup is completed, it will attempt to recover table logs from residual data in local disk of the already dropped table, which no longer has schema data in ZooKeeper. Consequently, the TabletServer will throw a fatal exception and be unable to restart.

The error messages are as follows:

2025-08-06 16:31:31,185 ERROR com.alibaba.fluss.server.ServerBase                          [] - Could not start TabletServer.
com.alibaba.fluss.exception.FlussException: Failed to start the TabletServer.
        at com.alibaba.fluss.server.ServerBase.start(ServerBase.java:136) ~[fluss-server-0.8-SNAPSHOT.jar:0.8-SNAPSHOT]
        at com.alibaba.fluss.server.ServerBase.startServer(ServerBase.java:93) [fluss-server-0.8-SNAPSHOT.jar:0.8-SNAPSHOT]
        at com.alibaba.fluss.server.tablet.TabletServer.main(TabletServer.java:172) [fluss-server-0.8-SNAPSHOT.jar:0.8-SNAPSHOT]
Caused by: com.alibaba.fluss.exception.FlussRuntimeException: Failed to recovery log
        at com.alibaba.fluss.server.log.LogManager.loadLogs(LogManager.java:213) ~[fluss-server-0.8-SNAPSHOT.jar:0.8-SNAPSHOT]
        at com.alibaba.fluss.server.log.LogManager.startup(LogManager.java:129) ~[fluss-server-0.8-SNAPSHOT.jar:0.8-SNAPSHOT]
        at com.alibaba.fluss.server.tablet.TabletServer.startServices(TabletServer.java:200) ~[fluss-server-0.8-SNAPSHOT.jar:0.8-SNAPSHOT]
        at com.alibaba.fluss.server.ServerBase.start(ServerBase.java:123) ~[fluss-server-0.8-SNAPSHOT.jar:0.8-SNAPSHOT]
        ... 2 more

### Solution

When TabletServers attempt to recover table logs from data whose corresponding table has no schema in ZooKeeper, they can confirm that the table has already been dropped. In this case, they can skip the recovery process, and the residual data can be safely removed.

platinumhamburg:
### Purpose

Linked issue: close #1486 1486

### DIFF
--- a/fluss-server/src/test/java/org/apache/fluss/server/log/DroppedTableRecoveryTest.java
+++ b/fluss-server/src/test/java/org/apache/fluss/server/log/DroppedTableRecoveryTest.java
@@ -120,46 +120,6 @@ final class DroppedTableRecoveryTest extends LogTestBase {
     }
 
     @Test
-    void testLogTabletResidualDataCleanupWhenSchemaNotFound() throws Exception {
-        // Create a log first
-        LogTablet log =
-                logManager.getOrCreateLog(
-                        PhysicalTablePath.of(tablePath), tableBucket, LogFormat.ARROW, 1, false);
-
-        // Write some data to the log
-        MemoryLogRecords records = genMemoryLogRecordsByObject(DATA1);
-        log.appendAsLeader(records);
-        log.flush(false);
-
-        // Get the log directory path before removing the table metadata
-        String logDir = log.getLogDir().getAbsolutePath();
-
-        // Shutdown log manager first
-        logManager.shutdown();
-
-        // Simulate table drop: remove metadata from ZooKeeper but keep log directory on disk
-        zkClient.deleteTable(tablePath);
-
-        // Create a new LogManager and start it
-        LogManager newLogManager =
-                LogManager.create(
-                        conf,
-                        zkClient,
-                        new FlussScheduler(1),
-                        SystemClock.getInstance(),
-                        TestingMetricGroups.TABLET_SERVER_METRICS);
-
-        // LogManager startup should clean up residual data when encountering
-        // SchemaNotExistException
-        newLogManager.startup();
-
-        // Verify that the residual data directory was cleaned up
-        assertThat(new File(logDir)).doesNotExist();
-
-        newLogManager.shutdown();
-    }
-
-    @Test
     void testMultipleLogTabletResidualDataDirectoriesCleanup() throws Exception {
         // Create multiple logs for the same table
         TableBucket tableBucket1 = new TableBucket(DATA1_TABLE_ID, 1);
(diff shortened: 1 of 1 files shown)


fluss-server/src/main/java/org/apache/fluss/server/log/LogManager.java
```diff
@@ -194,6 +195,21 @@ private void loadLogs() {
...
+                                                            if (e
+                                                                    instanceof
+                                                                    SchemaNotExistException) {
+                                                                LOG.error(
+                                                                        "schema not exist, table for {} has already been dropped, the residual deta will be removed.",
+                                                                        tabletDir,
+                                                                        e);
+                                                                FileUtils.deleteDirectoryQuietly(
+                                                                        tabletDir);
+                                                                return;
```
wuchong: This code is too unreadable, could you refactor this code to not use lambda?

fluss-server/src/main/java/org/apache/fluss/server/log/LogManager.java
```diff
@@ -194,6 +195,21 @@ private void loadLogs() {
...
                                                         } catch (Exception e) {
+                                                            LOG.error(
+                                                                    "Fail to loadLog from {}",
+                                                                    tabletDir,
+                                                                    e);
+                                                            if (e
+                                                                    instanceof
+                                                                    SchemaNotExistException) {
+                                                                LOG.error(
+                                                                        "schema not exist, table for {} has already been dropped, the residual deta will be removed.",
```
wuchong:
```suggestion
                                                                        "schema not exist, table for {} has already been dropped, the residual data will be removed.",
```

fluss-server/src/test/java/org/apache/fluss/server/log/DroppedTableRecoveryTest.java
```diff
@@ -0,0 +1,258 @@
...
+
+import static org.apache.fluss.record.TestData.DATA1;
+import static org.apache.fluss.record.TestData.DATA1_SCHEMA;
+import static org.apache.fluss.record.TestData.DATA1_TABLE_DESCRIPTOR;
+import static org.apache.fluss.record.TestData.DATA1_TABLE_ID;
+import static org.apache.fluss.testutils.DataTestUtils.genMemoryLogRecordsByObject;
+import static org.assertj.core.api.Assertions.assertThat;
+
+/** Test for handling recovery from residual data of already dropped tables. */
+final class DroppedTableRecoveryTest extends LogTestBase {
```
wuchong, line 59: Could you add a test for `KvManager` and kv tables? I'm not sure whether KvManager also has this problem, and it seems we haven't cleaned up the kv tablet dir.

fluss-server/src/main/java/org/apache/fluss/server/log/LogManager.java
platinumhamburg: Thanks for the suggestion! I'll refactor the code above.

fluss-server/src/test/java/org/apache/fluss/server/log/DroppedTableRecoveryTest.java
platinumhamburg: Got it，I'll add tests for those cases.

platinumhamburg:
@wuchong I have addressed all the comments above. Thank you for reviewing again.

wuchong (approved):

PR #1503, opened 2025-08-08
Add configuration options to control table creation restrictions

### LINKED ISSUE #1502: Implement table type creation restrictions via configuration options
### Search before asking

### Motivation

In certain operational scenarios, cluster service providers require the ability to disable or enable the creation of specific table types within the cluster. This represents a key operational and management requirement that would enhance service flexibility and control.

### Solution

Add configuration options to control table creation restrictions。

### Anything else?

_No response_

### Willingness to contribute

platinumhamburg:
### Purpose

Linked issue: close #1502

### DIFF
--- a/fluss-server/src/test/java/org/apache/fluss/server/coordinator/TableManagerITCase.java
+++ b/fluss-server/src/test/java/org/apache/fluss/server/coordinator/TableManagerITCase.java
@@ -126,55 +126,9 @@ class TableManagerITCase {
                     .setClusterConf(initConf())
                     .build();
 
-    // Extension for testing log table creation restriction
-    @RegisterExtension
-    public static final FlussClusterExtension FLUSS_CLUSTER_LOG_RESTRICTED_EXTENSION =
-            FlussClusterExtension.builder()
-                    .setNumOfTabletServers(3)
-                    .setCoordinatorServerListeners(
-                            String.format(
-                                    "%s://localhost:0, %s://localhost:0",
-                                    DEFAULT_LISTENER_NAME, CLIENT_LISTENER))
-                    .setTabletServerListeners(
-                            String.format(
-                                    "%s://localhost:0, %s://localhost:0",
-                                    DEFAULT_LISTENER_NAME, CLIENT_LISTENER))
-                    .setClusterConf(initLogRestrictedConf())
-                    .build();
-
-    // Extension for testing kv table creation restriction
-    @RegisterExtension
-    public static final FlussClusterExtension FLUSS_CLUSTER_KV_RESTRICTED_EXTENSION =
-            FlussClusterExtension.builder()
-                    .setNumOfTabletServers(3)
-                    .setCoordinatorServerListeners(
-                            String.format(
-                                    "%s://localhost:0, %s://localhost:0",
-                                    DEFAULT_LISTENER_NAME, CLIENT_LISTENER))
-                    .setTabletServerListeners(
-                            String.format(
-                                    "%s://localhost:0, %s://localhost:0",
-                                    DEFAULT_LISTENER_NAME, CLIENT_LISTENER))
-                    .setClusterConf(initKvRestrictedConf())
-                    .build();
-
     private static Configuration initConf() {
         Configuration conf = new Configuration();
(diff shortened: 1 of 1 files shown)


fluss-common/src/main/java/com/alibaba/fluss/config/ConfigOptions.java
```diff
@@ -1309,6 +1309,24 @@ public class ConfigOptions {
                             "The column name of the version column for the `versioned` merge engine. "
                                     + "If the merge engine is set to `versioned`, the version column must be set.");
 
+    public static final ConfigOption<Boolean> LOG_TABLE_ALLOW_CREATION =
+            key("log.table.allow-creation")
```
wuchong: How about `allow.create.log.tables` and `allow.create.kv.tables`? This makes sure both of them be able to be together on docs when we support auto generate docs in the future. 

And please put the ConfigOptions after `AUTO_PARTITION_CHECK_INTERVAL`? Because `table.*` are table-level configs, not server-side configs.

fluss-server/src/main/java/com/alibaba/fluss/server/coordinator/CoordinatorService.java
```diff
@@ -650,4 +657,36 @@ private static List<BucketMetadata> getBucketMetadataFromContext(
...
+        boolean hasPrimaryKey = tableDescriptor.hasPrimaryKey();
+
+        if (hasPrimaryKey) {
+            // This is a KV table (Primary Key Table)
+            if (!kvTableAllowCreation) {
+                throw new InvalidTableException(
+                        String.format(
+                                "Creation of KV tables (primary key tables) is not allowed. "
+                                        + "Table: %s. Please set '%s' to true to enable KV table creation.",
+                                tablePath, ConfigOptions.KV_TABLE_ALLOW_CREATION.key()));
```
wuchong:
```suggestion
                        "Creation of Primary Key Tables is disallowed in the cluster.");
```

1. Use `Primary Key Table` for user-facing messages, we don't have the concept of KV table actually. 
2. remove the cluster config from the exception message, because user shouldn't be aware of the cluster config and usually don't have the permission to access cluster configs (so it's not necessary to expose them).

fluss-server/src/main/java/com/alibaba/fluss/server/coordinator/CoordinatorService.java
```diff
@@ -650,4 +657,36 @@ private static List<BucketMetadata> getBucketMetadataFromContext(
...
+            }
+        } else {
+            // This is a Log table
+            if (!logTableAllowCreation) {
+                throw new InvalidTableException(
+                        String.format(
+                                "Creation of Log tables is not allowed. "
+                                        + "Table: %s. Please set '%s' to true to enable Log table creation.",
+                                tablePath, ConfigOptions.LOG_TABLE_ALLOW_CREATION.key()));
```
wuchong: ditto 

```suggestion
               throw new InvalidTableException(
                        "Creation of Log Tables is disallowed in the cluster.");
```

wuchong (commented):
Thanks @platinumhamburg for contribution. 

Please add tests for this feature (e.g., in `FlussAdminITCase` or `TableManagerITCase`). Besides, please add documentation for the new configs.

platinumhamburg:
@wuchong Thanks for your comments, all comments resolved, and I supplemented some cases。

fluss-server/src/test/java/org/apache/fluss/server/coordinator/TableManagerITCase.java
```diff
@@ -126,12 +126,58 @@ class TableManagerITCase {
                     .setClusterConf(initConf())
                     .build();
 
+    // Extension for testing log table creation restriction
+    @RegisterExtension
+    public static final FlussClusterExtension FLUSS_CLUSTER_LOG_RESTRICTED_EXTENSION =
```
wuchong: The `@RegisterExtension` is a global resource that shares between testing methods. It's create the log/kv restricted cluster in dedicated testing method to reduce the resource cost.

wuchong (approved):
I added a commit to improve the test a bit. Others looks good.

PR #1507, opened 2025-08-09
[server] Push partition filtering to the server-side to minimize network bandwidth usage.

xiaochen-zhou:
### Purpose

Push partition filtering to the server-side to minimize network bandwidth usage.

### DIFF
--- a/fluss-flink/fluss-flink-common/src/main/java/com/alibaba/fluss/flink/source/enumerator/FlinkSourceEnumerator.java
+++ b/fluss-flink/fluss-flink-common/src/main/java/com/alibaba/fluss/flink/source/enumerator/FlinkSourceEnumerator.java
@@ -37,6 +37,7 @@ import com.alibaba.fluss.flink.source.split.SourceSplitBase;
 import com.alibaba.fluss.flink.source.state.SourceEnumeratorState;
 import com.alibaba.fluss.flink.utils.PushdownUtils.FieldEqual;
 import com.alibaba.fluss.metadata.PartitionInfo;
+import com.alibaba.fluss.metadata.PartitionSpec;
 import com.alibaba.fluss.metadata.TableBucket;
 import com.alibaba.fluss.metadata.TableInfo;
 import com.alibaba.fluss.metadata.TablePath;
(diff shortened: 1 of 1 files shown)


fluss-flink/fluss-flink-common/src/main/java/com/alibaba/fluss/flink/source/enumerator/FlinkSourceEnumerator.java
```diff
@@ -263,8 +264,13 @@ private List<SourceSplitBase> initNonPartitionedSplits() {
 
     private Set<PartitionInfo> listPartitions() {
         try {
-            List<PartitionInfo> partitionInfos = flussAdmin.listPartitionInfos(tablePath).get();
-            partitionInfos = applyPartitionFilter(partitionInfos);
+            PartitionSpec partitionSpec = convertFiltersToPartitionSpec(partitionFilters);
```
wuchong: We can convert the filter to `partialPartitionSpec` during the construction of `FlinkSourceEnumerator` to avoid the repeated conversion for `listPartitions`.

wuchong (commented):
Thanks @xiaochen-zhou for the contribution. I have some comments for the PR.

Please create an issue and reach consensus on the approach before submitting a pull request. 

While this PR does enhance the code, I'm not sure it will significantly reduce network load in practice. The listPartitions call is made every 10 seconds, and typically there are fewer than 100 partitions in total — so the payload size is relatively small.

More importantly, the main bottleneck for this RPC lies in ZooKeeper performance, and since the partition filtering isn't pushed down to the ZooKeeper layer, the server still retrieves all partition data before applying the filter. As a result, we don’t get the expected reduction in backend load.

xiaochen-zhou:
> Thanks @xiaochen-zhou for the contribution. I have some comments for the PR.
> 
> Please create an issue and reach consensus on the approach before submitting a pull request.
> 
> While this PR does enhance the code, I'm not sure it will significantly reduce network load in practice. The listPartitions call is made every 10 seconds, and typically there are fewer than 100 partitions in total — so the payload size is relatively small.
> 
> More importantly, the main bottleneck for this RPC lies in ZooKeeper performance, and since the partition filtering isn't pushed down to the ZooKeeper layer, the server still retrieves all partition data before applying the filter. As a result, we don’t get the expected reduction in backend load.

Thanks  for your review. updated the code. I will create an issue.

xiaochen-zhou:
> Thanks @xiaochen-zhou for the contribution. I have some comments for the PR.
> 
> Please create an issue and reach consensus on the approach before submitting a pull request.
> 
> While this PR does enhance the code, I'm not sure it will significantly reduce network load in practice. The listPartitions call is made every 10 seconds, and typically there are fewer than 100 partitions in total — so the payload size is relatively small.
> 
> More importantly, the main bottleneck for this RPC lies in ZooKeeper performance, and since the partition filtering isn't pushed down to the ZooKeeper layer, the server still retrieves all partition data before applying the filter. As a result, we don’t get the expected reduction in backend load.

I think we can optimize on the ZK side: when the ResolvedPartitionSpec is a complete partition spec, we can directly retrieve the partition instead of scanning all partitions. This can satisfy some scenarios.

Issue #431, opened 2025-02-19
Automatically create corresponding Lake tables when 'table.datalake.enabled' enabled


leonardBang:
### Search before asking

- [x] I searched in the [issues](https://github.com/alibaba/fluss/issues) and found nothing similar.

### Description

Automatically create corresponding Lake tables when 'table.datalake.enabled' option is set to true

### Willingness to contribute

- [ ] I'm willing to submit a PR!

gyang94:
@luoyuxia Thank you. Could you assign this ticket to me? based on this [comment](https://github.com/alibaba/fluss/issues/430#issuecomment-2671347118)

leonardBang:
Hey, @gyang94 what's the progress of this ticket?

gyang94:
@leonardBang Sorry for late response. I didn't notice the comments here.
In the previous discussion, this ticket depends on the #430 . yuxia will do some initial work on `fluss-servers` module, and there will some refactoring about lakehouse related interface. This ticket will start after that.
Thus I haven't started this ticket yet. Is it ready to start now? If yes, I will start my work.

Refer to:
https://github.com/alibaba/fluss/issues/430#issuecomment-2671019627

gyang94:
Hi @luoyuxia , I found that it looks that you have implemented this feature during the lakehouse refactoring and commit in #640 . Is there anything still need to do on this ticket? 

<img width="1060" alt="Image" src="https://github.com/user-attachments/assets/36cdb219-de3f-44e0-baa3-3fe0a9f7321f" />

I just noticed that #430 was resolved these days, then I am going to start my work on this ticket. But I found the code is already there 😂  Could you confirm if this is done? If yes, maybe we can close this ticket. cc: @leonardBang

luoyuxia:
> Hi [@luoyuxia](https://github.com/luoyuxia) , I found that it looks that you have implemented this feature during the lakehouse refactoring and commit in [#640](https://github.com/alibaba/fluss/pull/640) . Is there anything still need to do on this ticket?
> 
[...]

PR #1509, opened 2025-08-10
[server] feat: support alter table properties

### LINKED ISSUE #100: [Feature] Support alter table to update properties and custom properties
### Search before asking

### Motivation

Support alter table to update properties and custom properties.

### Solution

_No response_

### Anything else?

_No response_

### Willingness to contribute

gyang94:
### Purpose

Linked issue: close #100 

Support alter table properties and custom properties


fluss-client/src/main/java/com/alibaba/fluss/client/admin/Admin.java
```diff
@@ -234,6 +234,23 @@ CompletableFuture<Void> createTable(
      */
     CompletableFuture<List<String>> listTables(String databaseName);
 
+    /**
+     * Alter a table.
+     *
+     * <p>The following exceptions can be anticipated when calling {@code get()} on returned future.
+     *
+     * <ul>
+     *   <li>{@link DatabaseNotExistException} if the database does not exist.
```
LiebingYu: And `TableNotExistException` if `ignoreIfNotExists` is false?

fluss-client/src/main/java/com/alibaba/fluss/client/admin/Admin.java
```diff
@@ -234,6 +234,23 @@ CompletableFuture<Void> createTable(
...
+     * </ul>
+     *
+     * @param tablePath The table path of the table.
+     * @param tableDescriptor The table descriptor.
+     * @param ignoreIfNotExists if it is true, do nothing if table does not exist. If false, throw a
+     *     TableNotExistException.
+     */
+    CompletableFuture<Void> alterTable(
```
LiebingYu, line 253: We should add some IT case for this.

fluss-server/src/main/java/com/alibaba/fluss/server/coordinator/MetadataManager.java
```diff
@@ -300,6 +303,65 @@ public long createTable(
                 "Fail to create table " + tablePath);
     }
 
+    public void alterTable(
+            TablePath tablePath, TableDescriptor tableDescriptor, boolean ignoreIfNotExists) {
+        // validate table properties before creating table
```
LiebingYu: // validate table properties before altering table

fluss-server/src/main/java/com/alibaba/fluss/server/coordinator/MetadataManager.java
```diff
@@ -300,6 +303,65 @@ public long createTable(
...
+        validateTableDescriptor(tableDescriptor, maxBucketNum);
```
LiebingYu: Currently we only support alter table properties. Should I fill the `Schema` here?

fluss-flink/fluss-flink-common/src/main/java/com/alibaba/fluss/flink/catalog/FlinkCatalog.java
```diff
@@ -389,9 +389,25 @@ public void createTable(ObjectPath objectPath, CatalogBaseTable table, boolean i
...
-        throw new UnsupportedOperationException();
+        TablePath tablePath = toTablePath(objectPath);
+        TableDescriptor tableDescriptor =
+                FlinkConversions.toFlussTable((ResolvedCatalogTable) newTable);
+        try {
+            admin.alterTable(tablePath, tableDescriptor, ignoreIfNotExist).get();
+        } catch (Exception e) {
+            Throwable t = ExceptionUtils.stripExecutionException(e);
+            if (CatalogExceptionUtils.isTableAlreadyExist(t)) {
+                throw new TableNotExistException(getName(), objectPath);
```
LiebingYu: I'm confused here. Why throw `TableNotExistException` when table already exists?

fluss-rpc/src/main/proto/FlussApi.proto
```diff
@@ -108,6 +108,17 @@ message CreateTableRequest {
...
 
+// alter table request and response
+message AlterTableRequest {
+  required PbTablePath table_path = 1;
+  required bytes table_json = 2;
+  required bool ignore_if_not_exists = 3;
+}
+
+message AlterTableResponse {
+
```
LiebingYu: Redundant blank line.

fluss-client/src/main/java/com/alibaba/fluss/client/admin/Admin.java
gyang94: added in `FlussAdminITCase.testAlterTable`

fluss-server/src/main/java/com/alibaba/fluss/server/coordinator/MetadataManager.java
gyang94: only validate properties now.

LiebingYu:
LGTM. Left a comment.

fluss-server/src/main/java/com/alibaba/fluss/server/coordinator/MetadataManager.java
```diff
@@ -300,6 +303,65 @@ public long createTable(
...
+            }
+        }
+    }
+
+    private TableRegistration getUpdatedTableRegistration(
+            TablePath tablePath, TableDescriptor updateTableDescriptor) {
+        TableRegistration existTableReg = getTableRegistration(tablePath);
+        Map<String, String> updateProperties = updateTableDescriptor.getProperties();
+        Map<String, String> updateCustomProperties = updateTableDescriptor.getCustomProperties();
+        validateAlterTableProperties(updateTableDescriptor);
```
LiebingYu: Is this a duplicated validation?

fluss-server/src/main/java/com/alibaba/fluss/server/coordinator/MetadataManager.java
```diff
@@ -300,6 +303,65 @@ public long createTable(
...
+                    return;
+                }
+                throw new TableNotExistException("Table " + tablePath + " does not exists.");
+            } else {
+                throw new FlussRuntimeException("Failed to alter table: " + tablePath, e);
+            }
+        }
+    }
+
+    private TableRegistration getUpdatedTableRegistration(
```
LiebingYu, line 342: I think we should check whether the property to be updated is the same as the existing one, and if so, no changes are required.

wuchong:
Hi @gyang94 , 

I suggest we first focus on supporting the RPC and API interfaces for altering **client-side table properties**, such as those under `client.*`, watermark configurations, and computed column settings. These are properties that can be safely updated online without requiring coordination with storage-layer components.

In contrast, properties under `table.*` are typically **storage-side configurations** that often require dynamic updates across TabletServers or changes to the underlying storage state. For example, flipping `table.datalake.enabled` from `false` to `true` would require not only updating the table config but also creating the corresponding external data lake table upfront.

gyang94:
> Hi @gyang94 ,
> 
> I suggest we first focus on supporting the RPC and API interfaces for altering **client-side table properties**, such as those under `client.*`, watermark configurations, and computed column settings. These are properties that can be safely updated online without requiring coordination with storage-layer components.
> 
> In contrast, properties under `table.*` are typically **storage-side configurations** that often require dynamic updates across TabletServers or changes to the underlying storage state. For example, flipping `table.datalake.enabled` from `false` to `true` would require not only updating the table config but also creating the corresponding external data lake table upfront.

Hi @wuchong , thanks for the comments. I agree. 
In this PR, I think it is better to focus on the altering API and process, not for altering any specific configs. 
To keep the update safely, I use a "whitelist"`FlussConfigUtils.ALTERABLE_CONFIG_OPTIONS` to check which property can be updated. This `table.datalake.enabled` is the one that I put into this list for... just enable altering something. It looks not a good choice here. I can modify it to use other safely updated configs.
I understand that we need do extra work when changing this config(#431 ). For this kind of configs, I expect they will be added into the "whitelist" one by one after handle extra work when altering them.

gyang94:
@LB-Yu @wuchong , updated based on comments, please check.
now I use 'client.connect-timeout' config as an example to alter a config.

luoyuxia:
I feel that the alter table interface is not easy to use. For alter table, users have to construct a new `TableDescriptor` even for just add new property. Can we follow the design of https://cwiki.apache.org/confluence/display/FLINK/FLIP-273%3A+Improve+the+Catalog+API+to+Support+ALTER+TABLE+syntax to make users only specified the new property to add instead of a whole `TableDescriptor`?

luoyuxia:
Flink also provide the new method to only pass the changed element instead of the whole catalog table. 
void alterTable(ObjectPath tablePath, CatalogBaseTable newTable, List<TableChange> tableChanges, boolean ignoreIfNotExists) throws TableNotExistException, CatalogException {
        this.alterTable(tablePath, newTable, ignoreIfNotExists);
    }

Pass the whole catalog table is real hard to use.

LiebingYu:
I aggree with @luoyuxia. Furthermore, I think we need to logically split the implementation of alterTable. alterTable involves various types of changes, such as comments, properties, buckets, columns, etc. We can separate each type of change into its own method, because some changes are not as simple as property changes, which only require modifications in ZK. 

For now we can only have alterProperties in alterTable implementation.

gyang94:
@luoyuxia @LB-Yu , thanks for your comments. I agreed.

Then we need to re-design it a little bit, this is what I think:
1. we need a fluss version table changes class, (e.g.`FlussTableChanges`) to define the Fluss table changes.
2. In `alterTable` function, to avoid pass whole `TableDescriptor`, we can only pass list of `FlussTableChanges` list to server. And then the server do altering based on different table change types. 
`alterTable(TablePath tablePath, List<FlussTableChanges> tableChanges, boolean ignoreIfNotExists)`. In this ticket, we only support `SetOptions` type.

Is that good? If yes, I will modify the implementation.

luoyuxia:
> @luoyuxia @LB-Yu , thanks for your comments. I agreed.
> 
> Then we need to re-design it a little bit, this is what I think:
> 
> 1. we need a fluss version table changes class, (e.g.`FlussTableChanges`) to define the Fluss table changes.
> 2. In `alterTable` function, to avoid pass whole `TableDescriptor`, we can only pass list of `FlussTableChanges` list to server. And then the server do altering based on different table change types.
>    `alterTable(TablePath tablePath, List<FlussTableChanges> tableChanges, boolean ignoreIfNotExists)`. In this ticket, we only support `SetOptions` type.
> 
> Is that good? If yes, I will modify the implementation.

I'm not sure. I think we may need a FIP for that. cc @wuchong

LiebingYu:
> @luoyuxia @LB-Yu , thanks for your comments. I agreed.
> 
> Then we need to re-design it a little bit, this is what I think:
> 
> 1. we need a fluss version table changes class, (e.g.`FlussTableChanges`) to define the Fluss table changes.
> 2. In `alterTable` function, to avoid pass whole `TableDescriptor`, we can only pass list of `FlussTableChanges` list to server. And then the server do altering based on different table change types.
>    `alterTable(TablePath tablePath, List<FlussTableChanges> tableChanges, boolean ignoreIfNotExists)`. In this ticket, we only support `SetOptions` type.
> 
> Is that good? If yes, I will modify the implementation.

@gyang94 would you like to fire a FIP?

gyang94:
@luoyuxia @LB-Yu Sure. I will take a look and fire a FIP about it.

luoyuxia:
Continue on #1625

Issue #1547, opened 2025-08-15
Adding Server-Level Storage Aggregation Metrics


swuferhong:
### Search before asking

- [x] I searched in the [issues](https://github.com/alibaba/fluss/issues) and found nothing similar.

### Motivation

Adding Server-Level Storage Aggregation Metrics

### Willingness to contribute

- [ ] I'm willing to submit a PR!

PR #1548, opened 2025-08-15
[metrics] Adding Server-Level Storage Aggregation Metrics

swuferhong:
### Purpose

Linked issue: #1547 

This pr is aims to introduce some Server-Level Storage Aggregation Metrics.

### DIFF
--- a/fluss-server/src/main/java/org/apache/fluss/server/coordinator/CompletedSnapshotStoreManager.java
+++ b/fluss-server/src/main/java/org/apache/fluss/server/coordinator/CompletedSnapshotStoreManager.java
@@ -28,7 +28,6 @@ import org.apache.fluss.server.kv.snapshot.CompletedSnapshotHandleStore;
 import org.apache.fluss.server.kv.snapshot.CompletedSnapshotStore;
 import org.apache.fluss.server.kv.snapshot.SharedKvFileRegistry;
 import org.apache.fluss.server.kv.snapshot.ZooKeeperCompletedSnapshotHandleStore;
-import org.apache.fluss.server.metrics.group.BucketMetricGroup;
 import org.apache.fluss.server.metrics.group.CoordinatorMetricGroup;
 import org.apache.fluss.server.zk.ZooKeeperClient;
 import org.apache.fluss.utils.MapUtils;
@@ -137,7 +136,7 @@ public class CompletedSnapshotStoreManager {
                                 bucket,
                                 end - start);
 
-                        BucketMetricGroup bucketMetricGroup =
+                        MetricGroup bucketMetricGroup =
                                 coordinatorMetricGroup.getTableBucketMetricGroup(
                                         tablePath, tableBucket);
                         if (bucketMetricGroup != null) {
--- a/fluss-server/src/main/java/org/apache/fluss/server/coordinator/CompletedSnapshotStoreManager.java
+++ b/fluss-server/src/main/java/org/apache/fluss/server/coordinator/CompletedSnapshotStoreManager.java
@@ -141,12 +141,16 @@ public class CompletedSnapshotStoreManager {
                                 coordinatorMetricGroup.getTableBucketMetricGroup(
                                         tablePath, tableBucket);
                         if (bucketMetricGroup != null) {
-                            LOG.info("ADDED bucketMetricGroup for tableBucket {}.", bucket);
+                            LOG.info("Add bucketMetricGroup for tableBucket {}.", bucket);
                             bucketMetricGroup.gauge(
                                     MetricNames.KV_NUM_SNAPSHOTS, () -> getNumSnapshots(bucket));
                             bucketMetricGroup.gauge(
                                     MetricNames.KV_ALL_SNAPSHOT_SIZE,
                                     () -> getAllSnapshotSize(bucket));
+                        } else {
+                            LOG.warn(
+                                    "Failed to add bucketMetricGroup for tableBucket {} when creating completed snapshot.",
+                                    bucket);
                         }
                         return snapshotStore;
                     } catch (Exception e) {
--- a/fluss-server/src/main/java/org/apache/fluss/server/coordinator/CompletedSnapshotStoreManager.java
+++ b/fluss-server/src/main/java/org/apache/fluss/server/coordinator/CompletedSnapshotStoreManager.java
@@ -19,12 +19,17 @@ package org.apache.fluss.server.coordinator;
 
 import org.apache.fluss.annotation.VisibleForTesting;
 import org.apache.fluss.metadata.TableBucket;
+import org.apache.fluss.metadata.TablePath;
+import org.apache.fluss.metrics.MetricNames;
+import org.apache.fluss.metrics.groups.MetricGroup;
 import org.apache.fluss.server.kv.snapshot.CompletedSnapshot;
 import org.apache.fluss.server.kv.snapshot.CompletedSnapshotHandle;
 import org.apache.fluss.server.kv.snapshot.CompletedSnapshotHandleStore;
 import org.apache.fluss.server.kv.snapshot.CompletedSnapshotStore;
 import org.apache.fluss.server.kv.snapshot.SharedKvFileRegistry;
 import org.apache.fluss.server.kv.snapshot.ZooKeeperCompletedSnapshotHandleStore;
+import org.apache.fluss.server.metrics.group.BucketMetricGroup;
+import org.apache.fluss.server.metrics.group.CoordinatorMetricGroup;
 import org.apache.fluss.server.zk.ZooKeeperClient;
 import org.apache.fluss.utils.MapUtils;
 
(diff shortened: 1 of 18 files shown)


wuchong:
@swuferhong could you rebase and resolve the conflicts?

swuferhong:
@wuchong conflicts resolved. PR ready.

fluss-server/src/main/java/org/apache/fluss/server/coordinator/CoordinatorEventProcessor.java
```diff
@@ -959,7 +984,8 @@ private void tryProcessCommitKvSnapshot(
                                 event.getAddCompletedSnapshotData().getCompletedSnapshot();
                         // add completed snapshot
                         CompletedSnapshotStore completedSnapshotStore =
-                                completedSnapshotStoreManager.getOrCreateCompletedSnapshotStore(tb);
+                                completedSnapshotStoreManager.getOrCreateCompletedSnapshotStore(
+                                        coordinatorContext.getTablePathById(tb.getTableId()), tb);
```
wuchong: move the access of `coordinatorContext`  out of the `ioExecutor`, otherwise, it's not thread-safe.

fluss-server/src/main/java/org/apache/fluss/server/log/LogTablet.java
```diff
@@ -339,7 +342,22 @@ public void registerMetrics(BucketMetricGroup bucketMetricGroup) {
...
+
+    public long logSize() {
+        return localLog.getSegments().sizeInBytes();
+    }
+
+    public long logicalStorageSize() {
+        if (remoteLogEndOffset <= 0L) {
+            return localLog.getSegments().sizeInBytes();
+        } else {
+            AtomicLong logicalStorageSize = new AtomicLong(remoteLogSize);
```
wuchong: Why use `AtomicLong` here? There should be multi-thread problems here. Using `.reduce(remoteLogSize, Long::sum)` should also work?

fluss-server/src/main/java/org/apache/fluss/server/replica/ReplicaManager.java
```diff
@@ -338,6 +353,33 @@ private int writerIdCount() {
...
+    private long logicalStorageLogSize() {
+        return onlineReplicas().map(Replica::logicalStorageLogSize).reduce(0L, Long::sum);
+    }
+
+    private long logicalStorageKvSize() {
+        return onlineReplicas().map(Replica::logicalStorageKvSize).reduce(0L, Long::sum);
+    }
+
+    private long physicalStorageLocalSize() {
+        AtomicLong localSize = new AtomicLong(0L);
```
wuchong: Why use `AtomicLong` not `.reduce` for this metric?

fluss-server/src/main/java/org/apache/fluss/server/metrics/group/CoordinatorMetricGroup.java
```diff
@@ -51,4 +66,128 @@ protected final void putVariables(Map<String, String> variables) {
...
+            return null;
+        }
+        return tableMetricGroup.buckets.get(tableBucket);
+    }
+
+    public void addTableBucketMetricGroup(
+            PhysicalTablePath physicalTablePath,
+            long tableId,
+            @Nullable Long partitionId,
+            Set<Integer> assignments) {
```
wuchong: `assignments` -> `buckets`?

fluss-server/src/main/java/org/apache/fluss/server/coordinator/CoordinatorEventProcessor.java
```diff
@@ -512,6 +513,14 @@ private void processCreateTable(CreateTableEvent createTableEvent) {
                     null,
                     null,
                     tableBuckets);
+
+            // register table metrics.
+            coordinatorMetricGroup.addTableBucketMetricGroup(
```
wuchong, line 571: If Coordinator restarts, how the existing table register the table and bucket metrics?

fluss-server/src/main/java/org/apache/fluss/server/coordinator/CompletedSnapshotStoreManager.java
```diff
@@ -104,6 +136,18 @@ public CompletedSnapshotStore getOrCreateCompletedSnapshotStore(TableBucket tabl
                                 "Created snapshot store for table bucket {} in {} ms.",
                                 bucket,
                                 end - start);
+
+                        BucketMetricGroup bucketMetricGroup =
+                                coordinatorMetricGroup.getTableBucketMetricGroup(
+                                        tablePath, tableBucket);
+                        if (bucketMetricGroup != null) {
```
wuchong, line 142: Will we miss reporting some snapshot results if the `bucketMetricGroup` not registered yet? Would be better to add warning logs for this case

fluss-server/src/main/java/org/apache/fluss/server/kv/snapshot/SharedKvFileRegistry.java
```diff
@@ -66,6 +66,22 @@ public SharedKvFileRegistry(Executor asyncDisposalExecutor) {
...
+        synchronized (registeredKvEntries) {
+            long size = 0;
+            for (SharedKvEntry entry : registeredKvEntries.values()) {
+                size += entry.kvFileHandle.getSize();
+            }
+            return size;
+        }
+    }
+
+    public long getNumSnapshots() {
```
wuchong: this is not used, can be removed.

fluss-server/src/main/java/org/apache/fluss/server/kv/snapshot/SharedKvFileRegistry.java
```diff
@@ -66,6 +66,22 @@ public SharedKvFileRegistry(Executor asyncDisposalExecutor) {
         this.open = true;
     }
 
+    public long getFileSize() {
+        synchronized (registeredKvEntries) {
```
wuchong: It's super heavy to access locks just for metrics. We can maintain a volatile varibles for file size and numSnapshots, and update them once the `registeredKvEntries` is updated. This is an incremental computation and only happens when a new snapshot is created. So it should be more lightweight.

fluss-server/src/main/java/org/apache/fluss/server/metrics/group/CoordinatorMetricGroup.java
```diff
@@ -51,4 +66,128 @@ protected final void putVariables(Map<String, String> variables) {
...
+    public void removeTablePartitionMetricsGroup(
+            TablePath tablePath, long tableId, long partitionId) {
+        SimpleTableMetricGroup tableMetricGroup = metricGroupByTable.get(tablePath);
+        if (tableMetricGroup != null) {
+            tableMetricGroup.removeBucketMetricsGroupForPartition(tableId, partitionId);
+        }
+    }
+
+    /** The metric group for table. */
+    public static class SimpleTableMetricGroup extends AbstractMetricGroup {
```
wuchong: can be `private`.

fluss-server/src/main/java/org/apache/fluss/server/metrics/group/BucketMetricGroup.java
```diff
@@ -37,7 +37,7 @@ public BucketMetricGroup(
             MetricRegistry registry,
             @Nullable String partitionName,
             int bucket,
-            TableMetricGroup parent) {
+            AbstractMetricGroup parent) {
```
wuchong: This is problematic. Because `BucketMetricGroup#getTableMetricGroup` returns the parent as `TableMetricGroup`.

swuferhong: I change the method from `BucketMetricGroup#getTableMetricGroup()` to `BucketMetricGroup#getParentMetricGroup()`, and change the result from `TableMetricGroup` to `AbstracrMetricGroup`.

fluss-server/src/main/java/org/apache/fluss/server/coordinator/CoordinatorEventProcessor.java
```diff
@@ -456,6 +457,20 @@ private void loadAssignment(
                 coordinatorContext.putBucketLeaderAndIsr(tableBucket, leaderAndIsr);
             }
         }
+
+        // register table/bucket metrics when initialing context.a
```
wuchong:
```suggestion
        // register table/bucket metrics when initialing context
```

fluss-server/src/main/java/org/apache/fluss/server/replica/Replica.java
```diff
@@ -246,10 +249,36 @@ public Replica(
     private void registerMetrics() {
         // get root metric in the reference: bucket -> table -> tabletServer
         TabletServerMetricGroup serverMetrics =
-                bucketMetricGroup.getTableMetricGroup().getServerMetricGroup();
+                ((TableMetricGroup) bucketMetricGroup.getParentMetricGroup())
```
wuchong: This is still very hack. We can introduce a private `SimpleBucketMetricGroup` in `CoordinatorMetricGroup` like `SimpleTableMetricGroup`. And expose it as `MetricGroup` for method `getTableBucketMetricGroup`.

swuferhong:
@wuchong comments addressed.

PR #1598, opened 2025-08-27
[flink] Union read in stream mode should support read from given timestamp

### LINKED ISSUE #1537: subscribe from timestamp should also consider data in lake in flink connector for log table
### Search before asking

### Description

Subscribe from timestamp should also consider data in lake in flink connector for log table

### Willingness to contribute

luoyuxia:
### Purpose

Linked issue: close #1537

### DIFF
--- a/fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/source/FlinkTableSource.java
+++ b/fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/source/FlinkTableSource.java
@@ -32,8 +32,12 @@ import org.apache.fluss.lake.source.LakeSource;
 import org.apache.fluss.lake.source.LakeSplit;
 import org.apache.fluss.metadata.MergeEngineType;
 import org.apache.fluss.metadata.TablePath;
+import org.apache.fluss.predicate.GreaterOrEqual;
+import org.apache.fluss.predicate.LeafPredicate;
 import org.apache.fluss.predicate.Predicate;
 import org.apache.fluss.predicate.PredicateBuilder;
+import org.apache.fluss.row.TimestampLtz;
+import org.apache.fluss.types.DataTypes;
 import org.apache.fluss.types.RowType;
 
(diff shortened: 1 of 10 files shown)


luoyuxia:
@caozhen1937 Please help me review this pr

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/source/FlinkTableSource.java
```diff
@@ -321,6 +342,36 @@ public boolean isBounded() {
...
+                        .acceptedPredicates();
+        if (acceptedPredicates.isEmpty()) {
+            LOG.warn(
+                    "The lake source doesn't accept the filter {}, won't read data from lake.",
+                    timestampFilter);
+            return false;
+        }
+        checkState(
+                acceptedPredicates.size() == 1
+                        && acceptedPredicates.get(0).equals(timestampFilter));
```
caozhen1937, line 371: Why is an exception being thrown here? If simply check if the `timestampFilter` exists, and if it doesn't, return false and log an error, is ok?

caozhen1937 (commented):
@luoyuxia LGTM. And a question in comment.

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/source/FlinkTableSource.java
luoyuxia: It means, we pass a filter to lake, and lake accept the filter, but the filter is not the passed filter. It should never happen, if it happens,   it must be that there is a bug in lake source implementation. 
So, I think we should throw exception directly to fast fail.

caozhen1937: Ok, I understand, it make sense.

luoyuxia:
Merging...

PR #1605, opened 2025-08-28
Support Compacted LogFormat

polyzos:


polyzos:
I'm closing this as #85 addresses that.. Gonna check though if there are any possible improvements that can benefit the current implementation and create a separate PR if required.

polyzos:
@swuferhong, before leaving this, since as you already addressed this, I wanted to check with you if you think there is anything we can reuse in case I missed something.
I only see this check https://github.com/apache/fluss/pull/1605/files#diff-74b13bc1eb78378f01a154188cdb7942d9b67dff4bbe2c1354a92617ae829349R148
And maybe there is any test we can reuse? Otherwise seems to me you have already covered everything, and my implementation doesn't add anything more, really.

swuferhong:
> @swuferhong, before leaving this, since as you already addressed this, I wanted to check with you if you think there is anything we can reuse in case I missed something. I only see this check https://github.com/apache/fluss/pull/1605/files#diff-74b13bc1eb78378f01a154188cdb7942d9b67dff4bbe2c1354a92617ae829349R148 And maybe there is any test we can reuse? Otherwise seems to me you have already covered everything, and my implementation doesn't add anything more, really.

Hi, @polyzos. I have wrongly closed #85  while organizing the PR. The work for "introduce compacted log row" hasn't been done yet. You can proceed with it in your current PR, but it's probably not high priority since there's currently no clear use case requiring this format.

polyzos:
@swuferhong So what happens with all the work that got merged? because I see there is lot of similarity in both PRs
Otherwise, maybe you can take a look at my PR?

swuferhong:
> @swuferhong So what happens with all the work that got merged? because I see there is lot of similarity in both PRs Otherwise, maybe you can take a look at my PR?

Hi, @polyzos  I don't quite understand this 'So what happens with all the work that got merged?'  as this work wasn't merged yet? 

I will try to review it. However, I'm not quite sure if it's necessary to introduce this format.

xx789633:
This feature is useful when projection is not required. it may not be a high priority.

luoyuxia:
@polyzos Hi, I'm thinking to push forward this pr. We have a user case that big embeding vector write to fluss , and flusss tier to lance when column prune is not required. Then use compacted change log format can much reduce the cost of server rebuilding arrow change log. Could you please rebase main branh?

polyzos:
@luoyuxia done 🫡

luoyuxia:
@polyzos Hi, thanks for your greate work. I verify it in my local, it works. 
I see you pr support:
-  write compacted row to log table
- generate change log with compacted row

But for me, the real highest priority  is **generate change log with compacted row**, so could you please split this pr into two, one is to support write compacted row to log table, another one is support change log with compacted row? I'll review the pr for **support change log with compacted row**?

wuchong:
Hi @polyzos , do you have time to resume the work of compacted format for log table? I think we only have a little things need to do for it (like the AppendWriter, the end-to-end tests for flink). I hope we can deliver a complete story for the compacted log format and deprecate the indexed log format in this version.

polyzos:
@wuchong I rebased. Let me know if you think there is anything else missing

fluss-client/src/main/java/org/apache/fluss/client/write/WriteRecord.java
```diff
@@ -127,6 +127,25 @@ public static WriteRecord forArrowAppend(
...
     }
 
+    /** Creates a write record for append operation for Compacted format. */
+    public static WriteRecord forCompactedAppend(
+            TableInfo tableInfo,
+            PhysicalTablePath tablePath,
+            CompactedRow row,
+            @Nullable byte[] bucketKey) {
+        checkNotNull(row);
+        int estimatedSizeInBytes = CompactedLogRecord.sizeOf(row) + RECORD_BATCH_HEADER_SIZE;
```
wuchong: This `RECORD_BATCH_HEADER_SIZE` is for kv record batch, we should use `LogRecordBatchFormat.recordBatchHeaderSize(CURRENT_LOG_MAGIC_VALUE)`

wuchong (approved):
Thanks @polyzos , this looks good to me. I only improved some code, like merge tests to reduce test time.

PR #1620, opened 2025-08-29
Introduce AlignedRow for row data encoding

### LINKED ISSUE #1615: Introduce BinaryRowData for row data encoding
### Search before asking

### Motivation

Currently, we support two types of BinaryRow: IndexedRow and CompactedRow.

The CompactedRow offers good space utilization but does not support random access and has heavy deserialization overhead. The IndexedRow provides a better tradeoff between storage efficiency and data access performance.

However, in performance-critical scenarios, we need a more efficient row encoding format. The BinaryRowData from Apache Flink and Apache Paimon could provide us with another viable option.

### Solution

Introduce BinaryRowData from Apache Flink and Apache Paimon.

### Anything else?

_No response_

### Willingness to contribute

platinumhamburg:
### Purpose

Linked issue: close #1615

### DIFF
--- a/fluss-common/src/test/java/org/apache/fluss/row/BinarySegmentUtilsTest.java
+++ b/fluss-common/src/test/java/org/apache/fluss/row/BinarySegmentUtilsTest.java
@@ -19,6 +19,8 @@ package org.apache.fluss.row;
 
 import org.apache.fluss.memory.MemorySegment;
 import org.apache.fluss.memory.MemorySegmentOutputView;
+import org.apache.fluss.row.aligned.AlignedRow;
+import org.apache.fluss.row.aligned.AlignedRowWriter;
 
 import org.junit.jupiter.api.Test;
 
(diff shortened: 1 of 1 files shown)


fluss-common/src/test/java/org/apache/fluss/row/BinaryRowDataTest.java
```diff
@@ -0,0 +1,728 @@
...
+        MemorySegment segment = MemorySegment.wrap(new byte[100]);
+        BinaryRowData row = new BinaryRowData(2);
+        row.pointTo(segment, 10, 48);
+        assertThat(segment).isSameAs(row.getSegments()[0]);
+        row.setInt(0, 5);
+        row.setDouble(1, 5.8D);
+    }
+
+    @Test
+    public void testSetAndGet() throws IOException, ClassNotFoundException {
```
wuchong: the exception signature can be removed

fluss-common/src/test/java/org/apache/fluss/row/BinaryRowDataTest.java
```diff
@@ -0,0 +1,728 @@
...
+        row.setNullAt(0);
+        row.setInt(1, 11);
+        row.setLong(2, 22);
+        row.setDouble(3, 33);
+        row.setBoolean(4, true);
+        row.setShort(5, (short) 55);
+        row.setByte(6, (byte) 66);
+        row.setFloat(7, 77f);
+
+        Consumer<BinaryRow> assertConsumer =
```
wuchong: Why use Consumer instead of assert directly on the row?

fluss-common/src/test/java/org/apache/fluss/row/BinaryRowDataTest.java
```diff
@@ -0,0 +1,728 @@
...
+        writer.reset();
+        writer.writeString(0, BinaryString.fromString("wahahah"));
+        writer.complete();
+        int hash2 = row.hashCode();
+
+        assertThat(hash2).isEqualTo(hash1);
+    }
+
+    @Test
+    public void testTimestampData() {
```
wuchong: `testTimestamp`

fluss-common/src/test/java/org/apache/fluss/row/BinaryRowDataTest.java
```diff
@@ -0,0 +1,728 @@
...
+            assertThat(row.isNullAt(1)).isTrue();
+            row.setTimestampNtz(0, TimestampNtz.fromMillis(-123L), precision);
+            assertThat(row.getTimestampNtz(0, 3).toString()).isEqualTo("1969-12-31T23:59:59.877");
+        }
+
+        // 2. not compact
+        {
+            final int precision = 9;
+            TimestampLtz timestamp1 =
+                    TimestampLtz.fromLocalDateTime(
```
wuchong: Create `TimestampLtz` from Java `Instant`.

fluss-common/src/main/java/org/apache/fluss/row/BinaryRowData.java
```diff
@@ -0,0 +1,504 @@
...
+ * <p>Fixed-length part will certainly fall into a MemorySegment, which will speed up the read and
+ * write of field. During the write phase, if the target memory segment has less space than fixed
+ * length part size, we will skip the space. So the number of fields in a single Row cannot exceed
+ * the capacity of a single MemorySegment, if there are too many fields, we suggest that user set a
+ * bigger pageSize of MemorySegment.
+ *
+ * <p>Variable-length part may fall into multiple MemorySegments.
+ */
+@Internal
+public final class BinaryRowData extends BinarySection
```
wuchong: `BinaryRowData` is a name in Flink, because it is `RowData` implementation in Flink. But it is strange if we use the same name in Fluss. Because `BinaryRowData` extends `BinaryRow` and `InternalRow`. 

Considering `CompactedRow` and `IndexedRow` are both `BinaryRow` which maps `InternalRow` on a `MemorySegment`, so I suggest to use another name for `BinaryRowData`. 

`BinaryRowData` aligns buffers with 8-bytes, maybe we can name it `AlignedRow`?  We should also rename `BinaryRowWriter` and `BinaryRowDataTest` as well.

fluss-common/src/main/java/org/apache/fluss/row/BinaryRowWriter.java
```diff
@@ -0,0 +1,411 @@
...
+ * <p>If want to reuse this writer, please invoke {@link #reset()} first.
+ */
+public final class BinaryRowWriter {
+
+    private final int nullBitsSizeInBytes;
+    private final BinaryRowData row;
+    private final int fixedSize;
+
+    protected MemorySegment segment;
+    protected int cursor;
```
wuchong: can be private.

fluss-common/src/main/java/org/apache/fluss/row/DataSetters.java
```diff
@@ -0,0 +1,68 @@
...
+ * limitations under the License.
+ */
+
+package org.apache.fluss.row;
+
+/**
+ * Provide type specialized setters to reduce if/else and eliminate box and unbox. This is mainly
+ * used on the binary format such as {@link BinaryRow}.
+ */
+public interface DataSetters {
```
wuchong: Better to name `TypedSetters`

fluss-common/src/main/java/org/apache/fluss/row/TimestampLtz.java
```diff
@@ -55,6 +57,21 @@ private TimestampLtz(long millisecond, int nanoOfMillisecond) {
         this.nanoOfMillisecond = nanoOfMillisecond;
     }
 
+    /**
+     * Creates an instance of {@link TimestampLtz} from an instance of {@link LocalDateTime}.
+     *
+     * @param dateTime an instance of {@link LocalDateTime}
+     */
+    public static TimestampLtz fromLocalDateTime(LocalDateTime dateTime) {
```
wuchong: It's on purpose we didn't introduce this static method on `TimestampLtz`, because `LocalDateTime` should be mapped to `TimestampNtz` instead of `TimestampLtz`. If converting `LocalDateTime` to `TimestampLtz`, users must to provides a local timezone parameter to convert the `LocalDateTime` into a epoch millisecond.

wuchong:
Please add the following comment on the header of `BinaryRowData`, `BinaryRowWriter`, and add the path to the `LICENSE` in the root dir. 

```
/* This file is based on source code of Apache Flink Project (https://flink.apache.org/), licensed by the Apache
 * Software Foundation (ASF) under the Apache License, Version 2.0. See the NOTICE file distributed with this work for
 * additional information regarding copyright ownership. */
```

platinumhamburg:
@wuchong Thank you for your advice. I have addressed all the issues and comments mentioned above. Please review the changes again when you have time.

wuchong (approved):

PR #1622, opened 2025-08-29
[test/flink] Collect flink rows should always be with timeout

### LINKED ISSUE #1621: Collect flink rows should always be with timeout
### Search before asking

### Motivation

Although #954 already introduced timeout closableiterator to fast fail flink test if it get more times to collect records before it reach to the ci 60min timeout.  But we still see the test stuck util ci 60min timeout. 
Here is just one example:
https://github.com/apache/fluss/actions/runs/16648794360

### Solution

- Correct the timeout closableiterator. Currently, it won't be timeout, since flink closableiterator#hasNext() will always block utils records comes in streaming mode
-  Make all directly use of flink closableiterator.next to use the timeout collect method

### Anything else?

_No response_

### Willingness to contribute

luoyuxia:
### Purpose

Linked issue: close #1621

### Brief change log
- Correct the timeout closableiterator. Currently, it won't be timeout, since flink closableiterator#hasNext() will always block utils records comes in streaming mode
- Make all directly use of flink closableiterator.next to use the timeout collect method. I try to find these code piece via `cusor`, but to be honest,   I'm not sure it convers all cases.
Also, I still can't figure out a good way to avoid such use in the future  

### Tests
existing UT & IT
Add a new UT to make sure it'll be really timeout.

### DIFF
--- a/fluss-flink/fluss-flink-common/src/test/java/org/apache/fluss/flink/source/testutils/FlinkRowAssertionsUtils.java
+++ b/fluss-flink/fluss-flink-common/src/test/java/org/apache/fluss/flink/source/testutils/FlinkRowAssertionsUtils.java
@@ -73,6 +73,13 @@ public class FlinkRowAssertionsUtils {
 
     public static List<String> collectRowsWithTimeout(
             CloseableIterator<Row> iterator, int expectedCount, boolean closeIterator) {
+        if (expectedCount < 0) {
+            throw new IllegalArgumentException(
+                    "Expected count must be non-negative: " + expectedCount);
+        }
+        if (iterator == null) {
+            throw new IllegalArgumentException("Iterator cannot be null");
+        }
         return collectRowsWithTimeout(
                 iterator,
                 expectedCount,


luoyuxia:
@MehulBatra Could you please help me review this pr since you're the author of the origin issue #954

fluss-flink/fluss-flink-common/src/test/java/org/apache/fluss/flink/source/testutils/FlinkRowAssertionsUtils.java
```diff
@@ -62,42 +66,45 @@ public static void assertQueryResultExactOrder(
...
-        int maxWaitTime = 60000; // 60 seconds
+        return collectRowsWithTimeout(
+                iterator,
+                expectedCount,
+                closeIterator,
+                // max wait 1 minute
+                Duration.ofMinutes(1));
+    }
 
+    protected static List<String> collectRowsWithTimeout(
```
MehulBatra, line 91: Would be beneficial for fail-fast validation, prevents confusing downstream errors, what do you think?
```
if (expectedCount < 0) {
    throw new IllegalArgumentException("Expected count must be non-negative: " + expectedCount);
}
if (iterator == null) {
    throw new IllegalArgumentException("Iterator cannot be null");
}
```

MehulBatra:
@luoyuxia Left one minor comment, PTAL, everything else looks good to me! Thanks for improving it, Hopefully CI should work well now.

PR #1655, opened 2025-09-08
[metric] Add cluster_id tag for client metric group

### LINKED ISSUE #1652: Add cluster_id tag for TabletServer client metrics
### Search before asking

### Description

Currently, the netty client metrics of TabletServer have no cluster_id tag. This makes it difficult to distinguish which cluster sent the request.

### Willingness to contribute

LiebingYu:
### Purpose

Linked issue: close #1652

### DIFF
--- a/fluss-server/src/main/java/org/apache/fluss/server/tablet/TabletServer.java
+++ b/fluss-server/src/main/java/org/apache/fluss/server/tablet/TabletServer.java
@@ -202,7 +202,10 @@ public class TabletServer extends ServerBase {
             // rpc client to sent request to the tablet server where the leader replica is located
             // to fetch log.
             this.clientMetricGroup =
-                    new ClientMetricGroup(metricRegistry, SERVER_NAME + "-" + serverId);
+                    new ClientMetricGroup(
+                            metricRegistry,
+                            ServerMetricUtils.validateAndGetClusterId(conf),
+                            SERVER_NAME + "-" + serverId);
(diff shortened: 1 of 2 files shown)


fluss-server/src/main/java/org/apache/fluss/server/tablet/TabletServer.java
```diff
@@ -202,7 +202,10 @@ protected void startServices() throws Exception {
             // rpc client to sent request to the tablet server where the leader replica is located
             // to fetch log.
             this.clientMetricGroup =
-                    new ClientMetricGroup(metricRegistry, SERVER_NAME + "-" + serverId);
+                    new ClientMetricGroup(
+                            metricRegistry,
+                            ServerMetricUtils.validateAndGetClusterId(conf),
+                            SERVER_NAME + "-" + serverId);
```
loserwang1024, line 208: It seems more like `server_id` rather than `cluster_id`? Different server in same cluster will collect different viriable.

LiebingYu: > It seems more like `server_id` rather than `cluster_id`? Different server in same cluster will collect different viriable.

`server_id` has already been added. Here I want to add `cluster_id` for client metric. The `cluster_id` is necessary when troubleshooting issues, for example, when examining metrics related to requests sent by the `ReplicaFetcherThread`. Currently, since the client metrics do not have a `cluster_id` label, we are unable to distinguish metrics from different clusters.

PR #1661, opened 2025-09-09
[hotfix] Paimon should always write bucket0 for bucket-unwared log table

luoyuxia:
### Purpose

Linked issue: close #xxx

### Brief change log

 Paimon should always write bucket0 for bucket-unwared log table

### Tests

Existing tests


fluss-lake/fluss-lake-paimon/src/main/java/org/apache/fluss/lake/paimon/tiering/append/AppendOnlyWriter.java
```diff
@@ -56,6 +60,11 @@ public void write(LogRecord record) throws Exception {
         // hacky, call internal method tableWrite.getWrite() to support
         // to write to given partition, otherwise, it'll always extract a partition from Paimon row
         // which may be costly
-        tableWrite.getWrite().write(partition, bucket, flussRecordAsPaimonRow);
+        int writtenBucket = bucket;
+        // if bucket-unaware mode, we have to use bucket = 0 to write to follow paimon best practice
```
xx789633, line 64: Is there a reference link for this best practice?

luoyuxia: No. Actually it's not best practice. It's a rule that engine should follows. If not, maybe bring unexpected behavior.

fluss-lake/fluss-lake-paimon/src/test/java/org/apache/fluss/lake/paimon/tiering/PaimonTieringTest.java
```diff
@@ -624,12 +622,15 @@ private CloseableIterator<InternalRow> getPaimonRows(
             // for log table, we can't filter by bucket directly, filter file by __bucket column
```
xx789633, line 622: Since the written bucket id of bucket-unaware table is always 0, can we assert that bucket id == 0 when reading?

luoyuxia:
@xx789633 Comments addressed again. PTAL

xx789633:
This looks good to me.

PR #1666, opened 2025-09-10
[client] Fix the MetadataRequest sent by LogFetcher timeout exception when upgrading cluster

### LINKED ISSUE #1665: MetadataRequest send by LogFetcher will be timeout if we upgrading cluster
### Search before asking

### Fluss version

main (development)

### Please describe the bug 🐞

MetadataRequest send by LogFetcher will be timeout if we upgrading cluster. This phenomenon has been mitigated after increasing 'client.netty.client.num-network-threads' in #1641 , but some flink subtasks still consistently experience metadata request timeouts. Therefore, we need to find the root cause to fix this issue

### Solution

_No response_

swuferhong:
### Purpose

Linked issue: close #1665 

This pr is aims to Fix the TimeoutException of MetadataRequest sent by LogFetcher when upgrading cluster. The root cause of this issue is that both `LogFetcher#sendFetches()` and `LogFetcher#handleFetchLogResponse()` hold the synchronized lock simultaneously. Specifically: 
1. `sendFetches()` initiates a metadata operation that sends a `MetadataRequest` while holding the lock.
2. `handleFetchLogResponse()` is triggered in the `whenComplete()` method of `FetchLogRequest`. The caller of `whenComplete()` – `ServerConnection.ResponseCallback#onRequestResult(int requestId, ApiMessage response)` – blocks the Netty client worker thread.

Deadlock Scenario:

- When `LogFetcher` triggers a `MetadataRequest`, it holds the synchronized lock of `LogFetcher`.
- Simultaneously, multiple `FetchLogResponses` return to `Netty client workers` (whose count is controlled by `client.netty.client.num-network-threads`).
- These responses invoke `ResponseCallback#onRequestResult`, which occupies the `Netty worker threads` and waits for the same `LogFetcher` lock.
- Result: Deadlock occurs because:
    a) The `MetadataResponse` cannot acquire a `Netty worker thread` (blocked by `FetchLogResponse` processing)
    b) `FetchLogResponse` processing waits indefinitely for the `LogFetcher` lock (held by `MetadataRequest`).
- Eventually, the `MetadataRequest` times out. After timeout, `ResponseCallback#onRequestResult` finally acquires the lock and recovers.


wuchong (approved):

fluss-client/src/main/java/org/apache/fluss/client/table/scanner/log/LogFetcher.java
```diff
@@ -243,18 +273,21 @@ private TableOrPartitions getTableOrPartitionsInFetchRequest(FetchLogRequest fet
...
+        TableOrPartitions(
                 @Nullable Set<Long> tableIds, @Nullable Set<TablePartition> tablePartitions) {
             this.tableIds = tableIds;
             this.tablePartitions = tablePartitions;
         }
     }
 
-    private void invalidTableOrPartitions(TableOrPartitions tableOrPartitions) {
+    @VisibleForTesting
+    public void invalidTableOrPartitions(TableOrPartitions tableOrPartitions) {
```
wuchong: Change them to package visible, it is only accessed in the same test class.

fluss-client/src/main/java/org/apache/fluss/client/table/scanner/log/LogFetcher.java
```diff
@@ -243,18 +273,21 @@ private TableOrPartitions getTableOrPartitionsInFetchRequest(FetchLogRequest fet
         return new TableOrPartitions(tableIdsInFetchRequest, tablePartitionsInFetchRequest);
     }
 
-    private static class TableOrPartitions {
+    /** A helper class to hold table ids or table partitions. */
+    @VisibleForTesting
+    public static class TableOrPartitions {
```
wuchong: Change them to package visible, it is only accessed in the same test class.

fluss-client/src/test/java/org/apache/fluss/client/table/scanner/log/LogFetcherTest.java
```diff
@@ -183,6 +188,37 @@ void testFetchWhenDestinationIsNullInMetadata() throws Exception {
...
+                            // the same reason as https://github.com/apache/fluss/pull/1666
+                            for (int i = 0; i < 1000; i++) {
+                                logFetcher.sendFetches();
+                                logFetcher.invalidTableOrPartitions(
+                                        new LogFetcher.TableOrPartitions(
+                                                Collections.singleton(tableId), null));
+                            }
+                        });
+
+        future.get(2, TimeUnit.MINUTES);
```
wuchong: assert future is done in the end, and close the executor

fluss-client/src/main/java/org/apache/fluss/client/utils/MetadataUtils.java
```diff
@@ -63,9 +63,10 @@ public class MetadataUtils {
      * add those table into cluster.
      */
     public static Cluster sendMetadataRequestAndRebuildCluster(
-            AdminReadOnlyGateway gateway, Set<TablePath> tablePaths)
+            AdminReadOnlyGateway gateway, Set<TablePath> tablePaths, long metadataRequestTimeoutMs)
```
wuchong: Please remove the timeout parameter, and use the default 30s in the test.

Issue #1542, opened 2025-08-14
Change default path to store rocksdb related logs


luoyuxia:
### Search before asking

- [x] I searched in the [issues](https://github.com/alibaba/fluss/issues) and found nothing similar.

### Motivation

As `kv.rocksdb.log.dir` said, if not set, Fluss will store rocksdb related logs into FLUSS_HOME/log/, but we also put important server log into  FLUSS_HOME/log/. The  rocksdb related logs will make it hard to find the important  server logs. So, we need to change the default path to FLUSS_HOME/log/rocksdb

### Willingness to contribute

- [ ] I'm willing to submit a PR!

beratli:
Hi! I would like to work on this issue. Can I be assigned to this task?

luoyuxia:
@beratli Hi, is there any progress in this issue, the date for code freeze for next release is at the end of this month, i hope it can be merged into next release version

rionmonster:
Hi @luoyuxia!

I came across this issue and was interested in contributing to Fluss (hopefully more in the future). I thought this was pretty low hanging fruit, but happy to make adjustments as/where needed.

luoyuxia:
@beratli Hi, is any progress on this issue. If you don't have time to finish this issue. I'd like to assign to @rionmonster since @rionmonster has raised a pull request.

luoyuxia:
@rionmonster HI, assigned to you~

PR #1675, opened 2025-09-10
[FLUSS-1542] Updated Default RocksDB Log Directory

rionmonster:
### Purpose

Linked issue: #1542 

Per Issue #1542, this pull request updates the default directory for RocksDB-specific logs to `FLUSS_HOME/log/rocksdb` (previously had no default directory set).

### Brief change log

Updated the `kv.rocksdb.log.dir` to now default to `FLUSS_HOME/log/rocksdb` and related configuration documentation and tests to reflect the change.

### Tests

Updated the existing `RocksDBResourceContainerTest.testDefaultDbLogDir` to now use the updated default value (changed from previous default log directory). No other tests within the `ConfigOptionsTests` were changed as all of those implementations appeared to be generic and no other property-specific tests were present.

### Documentation

No (related configuration documentation was updated to reflect the change)


fluss-common/src/main/java/org/apache/fluss/config/ConfigOptions.java
```diff
@@ -1401,7 +1401,7 @@ public class ConfigOptions {
     public static final ConfigOption<String> KV_LOG_DIR =
             key("kv.rocksdb.log.dir")
                     .stringType()
-                    .noDefaultValue()
+                    .defaultValue("FLUSS_HOME/log/rocksdb")
```
luoyuxia: Actually, FLUSS_HOME means the directory that you put Fluss, not a really a path name `FLUSS_HOME`.

luoyuxia: https://github.com/apache/fluss/blob/05ab5e86d1df69a96d56beed9ce258c603a478f3/fluss-server/src/main/java/org/apache/fluss/server/kv/rocksdb/RocksDBResourceContainer.java#L286

The parent of `logFilePath` will be `FLUSS_HOME/log`, so you can create a directory named `FLUSS_HOME/log/rocksdb` and `dbOptions.setDbLogDir` to the dir

luoyuxia (commented):
@rionmonster Thanks for the pr. Left some comments.

fluss-common/src/main/java/org/apache/fluss/config/ConfigOptions.java
rionmonster: Thank @luoyuxia! 100% makes sense. I'll go ahead and make these changes and update the PR.

rionmonster:
@luoyuxia 

I've gone ahead and made the necessary updates. I used the `FileUtils.createDirectory(...)` call to handle the creation of the logging directory as it seemed the most appropriate (and propagating the `throws IOException` to all of the related callers which I didn't love). I'm not sure if there's a preferred way to handle that, but wanted to mention it.

Additionally, the documentation update still mentions `FLUSS_HOME/log/rocksdb` which I think is clear, but may be improved (potentially by "$FLUSS_HOME/log/rocksdb" to indicate that the home directory is more dynamic). 

Open to any thoughts or better approaches around both of these items!

website/docs/maintenance/configuration.md
```diff
@@ -135,7 +135,7 @@ during the Fluss cluster working.
...
...ill then create the log under its own database folder as before.                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                               |
```
luoyuxia: Will it better to use `${FLUSS_HOME}/log/rocksdb` instead of FLUSS_HOME/log/rocksdb?
```suggestion
| kv.rocksdb.log.dir                                | String     | `${FLUSS_HOME}/log/rocksdb`   | The directory for RocksDB's information logging files. If empty (Fluss default setting), log files will be in the same directory as the Fluss log. If non-empty, this directory will be used and the data directory's absolute path will be used as the prefix of the log file name. If setting this option as a non-existing location, e.g `/dev/null`, RocksDB will then create the log under its own database folder as before.                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                               |
```

rionmonster: @luoyuxia 

This change has been made. I went ahead and squashed the commits as well to keep things clean as well (unsure if this is the preferred practice). Thanks for all of the feedback!

website/docs/maintenance/configuration.md
```diff
@@ -135,7 +135,7 @@ during the Fluss cluster working.
...
```
luoyuxia: Hi, you should should `${FLUSS_HOME}/log/rocksdb` instead of ${FLUSS_HOME}/log/rocksdb. Must with ``. Otherwise, the website build will fail

website/docs/maintenance/configuration.md
luoyuxia: @rionmonster We always **append** commit so that the reviewer can track what changes since last review.

rionmonster: @luoyuxia 

Understood, I’ll keep that in mind for any future commits/pull-requests. I know it can vary from project-to-project and that _some_ larger ones prefer just a single commit for a body of work.

Thanks again!

website/docs/maintenance/configuration.md
luoyuxia: Sorry, I mean 
```
`${FLUSS_HOME}/log/rocksdb`
```

rionmonster: @luoyuxia 

Done! Perfect opportunity to throw in one more commit (appended this time), which I _clearly_ manifested with my previous comment.

luoyuxia (approved):
@rionmonster Thanks for quick fix. Will merge after ci pass.

PR #1702, opened 2025-09-15
[KV] Implement Snapshot Query on KV Table

polyzos:
Here we introduce 2 new methods on the client:
**snapshotAll()** and **snapshotAllPartition(partitionname)** for a partitioned table. This provides flexibility to the users to query data for the latest partition which is where the realtime data lives for dashboards but also give them the ability to fetch earlier partition versions.

We **should add a note in the docs** that this functionality is for realtime use cases, that the table are not too large. 
With this functionality tables can be periodically queried to power dashboards and implement use cases like live location tracking for fleet management or topN - **where you typically have a small enough keyspace**. (maybe we can find more :) )


polyzos:
@platinumhamburg when you have time please help review this

platinumhamburg:
> @platinumhamburg when you have time please help review this

@polyzos No problem, I'm happy to work together on this PR.

platinumhamburg:
@polyzos Hi, Giannis, I have thoroughly read through this PR, and I feel that this PR is advancing some excellent work. However, I have some thoughts about its implementation, and perhaps we can have some discussion about it.

**1. Regarding Client-Side Interface Design**
Considering that the semantics of Lookuper is to provide point queries, I think defining scan queries in LookUper might not be a good idea. Perhaps adding the new interfaces to Scan would have more consistent semantics.
Given that: `FullScan == Scan without limit`, Therefore, a more natural client-side interface semantics would be:
`
table.newScan().createBatchScanner(...)
`
or
`
table.newScan().createLogScanner(...)
`

If we consider partition pruning requirements, we can define a filter interface for the Scan type:
`
table.newScan().filter(${partition_filter}).createBatchScanner(...)
`
We can temporarily constrain Predicate columns to only include partition columns, which can avoid committing to row-level filtering in this PR.

We further consider how the client side can refine the interface implementation, which involves two types: **TableScan** and **BatchScanner**. TableScan determines the Scanner type to create based on construction parameters.
The current BatchScanner implementation is KvSnapshotBatchScanner, which doesn't meet our needs. We can create a new type that inherits the BatchScanner interface called DefaultBatchScanner (or RemoteBatchScanner), and the scan logic newly added in PrimaryKeyLookuper in the current PR can be migrated into it.

2. Regarding ProtoBuf Protocol Definition

Perhaps we need to consider more points in protocol design:
- Maybe we should explicitly specify the Bucket list. Having TabletServer's ReplicaManager provide fuzzy filtering may not provide clear consistency semantics (considering scenarios like upgrades/downgrades, scaling, Coordinator primary-secondary switching, etc., the distribution state of Replicas may not be globally consistent)
- For a table with a large number of rows, the behavior of scanning all data in a single operation is terrible (we cannot restrict which tables users can perform fullScan on), so we must consider cases where multiple fetches are needed to complete a full scan. And once interfaces are made public, they must be stable, so even if we don't actually implement multiple scans in this PR, we can still define future-oriented protocols at the interface level. From this perspective, we might need a CreateSnapshot interface to define a temporary resource in a transient state on the server side and return a SnapshotId; thereafter we can use this SnapshotId multiple times to execute Scan requests to scan partial data; finally destroy the snapshot through a CloseSnapshot.

I look forward to hearing your thoughts, and welcome discussion at any time.

wuchong:
+1 for the proposal of @platinumhamburg , thanks for the detailed thoughts!

I think we need to discuss the public client API, RPC API and how to implement streaming read of kv snapshots. This deserves a FIP.

polyzos:
@platinumhamburg Thank you so much for all the detailed review and all this great feedback you have provided, which is more than helpfuk.

[Some context here:](https://github.com/apache/fluss/pull/1702#issuecomment-3305439761) the goal was to use this feature for realtime dashboard use cases - realtime user tracking, fleet management etc. where the key space is typically smaller due to visualization needs - i.e from really a few rows up to a few thousands with something ~1-5k as a threshold; **just poll the table (return a snapshot)**. At first i was thinking to make the threshold configurable to throw an exception and not allow the user to use this feature (to avoid OoO for large tables). Eventually I  thought of adding just a note in the docs, but seems like its better to add this.

Similarly for the API design, I was really reluctant in terms of the naming and where the api should live. I thought the scanner at first, but on my mind its more about unbounded / continuous data; this is just for a periodic db snapshot. The lookuper is about lookups on the kv table, and i thought of treating it as a "whole database" lookups.

Moreover because the intention was to return a full snapshot of the table as is - i.e its an aggregated table with the required columns, column pruning and predicate pushdown where not really taken into account. And same for the partition pruning because on the latest partition should have the latest desired values.

Seems like a made quite a few assumption in terms of correct usage! Let me know your thoughts on the above and I will close this PR and proceed with creating a design proposal first. 🙏 Again thank you so much for all this great feedback 🙇‍♂️ 

cc @wuchong

xx789633:
Hi @platinumhamburg @polyzos , thank you for sharing your detailed insights.

- I agree that the full scan should inherit the batch scanner API.
- Regarding the ProtoBuf protocol, I believe implementing [server-side streaming RPC](https://grpc.io/docs/languages/java/basics/#server-side-streaming-rpc) at the tablet server level would be beneficial for full table scan. This would reduce network roundtrips and simplify error handling—for example, the tablet server could close the opened snapshot immediately if any error occurs while sending back records. To mitigate potential performance impact, RocksDB snapshots should be released as soon as possible.

polyzos:
@platinumhamburg @xx789633 One thing I want to make clear is that the implementation here was intended for a **periodic poll/query of small** key/value tables (like 10-20 rows up to 1-2k -- maybe 10k for some extreme scenarios, but not larger, with also a small column cardinality - i.e 3-4 columns tops, thus small payload) , not for streaming use cases.

platinumhamburg:
> Hi @platinumhamburg @polyzos , thank you for sharing your detailed insights.
> 
> * I agree that the full scan should inherit the batch scanner API.
> * Regarding the ProtoBuf protocol, I believe implementing [server-side streaming RPC](https://grpc.io/docs/languages/java/basics/#server-side-streaming-rpc) at the tablet server level would be beneficial for full table scan. This would reduce network roundtrips and simplify error handling—for example, the tablet server could close the opened snapshot immediately if any error occurs while sending back records.

Hi @xx789633 , thank for advice, streaming RPC indeed provides us with more comfortable streaming read semantics and offers more natural exception handling and cleanup capabilities on Snapshot Pin. And I think Batch RPC also has its advantages. With Batch RPC, we can perform channel multiplexing within a single connection, allowing a single RPC request to serve multiple Table scanners simultaneously, just as we implemented in FetchLog requests. This can significantly reduce connection overhead. Perhaps we can further discuss and evaluate the pros and cons of both approaches.

polyzos:
For a periodic poll of a small table (every 1-5 seconds and 5 columns tops), what approach do you think makes?
I'm more inclined towards batch RPC, and I thought this approach with periodic polls makes sense since most of users find websockets and SSE way more intimidating.

However, if we want to consider large tables as well - although I think we should also identify the use cases first - a streaming approach indeed makes sense there.

platinumhamburg:
> @platinumhamburg @xx789633 One thing I want to make clear is that the implementation here was intended for a **periodic poll/query of small** key/value tables (like 10-20 rows up to 1-2k -- maybe 10k for some extreme scenarios, but not larger, with also a small column cardinality - i.e 3-4 columns tops, thus small payload) , not for streaming use cases.

Thanks @polyzos , I have understood the background requirements and your design trade-offs. Although we haven't reached complete consensus on all design issues in this PR yet, it has very thoroughly presented our background, objectives, and the concerns of all parties, and we were able to engage in detailed consideration of specific technical choices and details. I still feel this is excellent work and attempt, and I'm very happy to continue working together with everyone to push this work forward to achieve our goals, including the business requirements you mentioned.

platinumhamburg:
> For a periodic poll of a small table (every 1-5 seconds and 5 columns tops), what approach do you think makes? I'm more inclined towards batch RPC, and I thought this approach with periodic polls makes sense since most of users find websockets and SSE way more intimidating.
> 
> However, if we want to consider large tables as well - although I think we should also identify the use cases first - a streaming approach indeed makes sense there.

In this scenario, I tend to favor implementing the server-side protocol based on Batch RPC. We would create a Snapshot and retrieve the snapshot data through one or multiple pollBatch requests. The streaming read semantics can be provided to users by the client side in the form of an Iterator, but we don't necessarily have to implement it at the RPC layer. Under this pattern, full scans for both large and small tables are unified - for small tables, a single scan can return all data, while for large tables, multiple scans can be performed as needed. Therefore, the small table scenario naturally becomes a special case of the large table scenario.

polyzos:
@platinumhamburg Correct.. I want to check the best way to approach the initial design. Based on the convo here,
I will craft an initial design that:
1. extends the current scanner API (instead of having these on the Lookuper)
2. takes buckets into account for API
3. keeps things simple by using the batch RPC (at least in the initial proposal version)

However, since the goal was for small tables:
1. Do you think we should also consider multiple fetches? (this complicates thing a bit and im wondering if its indeed necessary)
2. Based on the above, if yes - I guess streaming RPC might make more sense in this scenario as well.

Do you think it makes sense to differentiate between this two use cases and start with the first approach? and if there is demand and use case also work on the second for larger tables and streaming? Or do you think it will complicate the design/implementation so its better to consider both scenarios now?

platinumhamburg:
> @platinumhamburg Correct.. I want to check the best way to approach the initial design. Based on the convo here, I will craft an initial design that:
> 
> 1. extends the current scanner API (instead of having these on the Lookuper)
> 2. takes buckets into account for API
> 3. keeps things simple by using the batch RPC (at least in the initial proposal version)
> 
> However, since the goal was for small tables:
> 
> 1. Do you think we should also consider multiple fetches? (this complicates thing a bit and im wondering if its indeed necessary)
> 2. Based on the above, if yes - I guess streaming RPC might make more sense in this scenario as well.
> 
> Do you think it makes sense to differentiate between this two use cases and start with the first approach? and if there is demand and use case also work on the second for larger tables and streaming? Or do you think it will complicate the design/implementation so its better to consider both scenarios now?

Based on my current thinking, I believe we might be better off not considering streaming semantics at the RPC layer, mainly for two reasons:

1. Our existing Batch RPC framework is a custom batch-semantics-based framework that has been heavily optimized for zero-copy operations. Developing an additional streaming RPC framework at this point would not provide equivalent benefits relative to the costs involved.
2. Streaming RPC only offers some convenience in usage, but it also comes with considerable performance limitations. For example, multiple Scanners cannot share connections (you can refer to the connection reuse mechanism currently implemented in the FetchLog system, where a single FetchLog request can simultaneously serve log fetching for multiple tables).

Regarding whether to support both small table and large table scanning scenarios in a single PR, I think it's feasible to achieve the goal through multiple steps - "implementation" is always flexible in task planning. However, what's truly challenging is the design of public RPC interfaces. Public APIs need to have certain stability and evolvability. We need to avoid scenarios like:

- We hastily expose an API that hasn't been thoroughly thought through, only to remove it several versions later.
- We expose an API that lacks evolvability, forcing subsequent feature development to create semantically fragmented API families.

Therefore, even if we plan to achieve our goals step by step at the API level, we should still have comprehensive design and planning for its future evolution and development from the very beginning.

PR #1706, opened 2025-09-16
[server] Forbid any table/partition creation that begins with `__`

### LINKED ISSUE #1705: Forbid any table/partition creation that begins with `__`
### Search before asking

### Description

See https://github.com/apache/fluss/issues/981#issuecomment-3294626197

### Willingness to contribute

caozhen1937:
### Purpose

Linked issue: close #1705


fluss-common/src/main/java/org/apache/fluss/metadata/TablePath.java
```diff
@@ -54,6 +54,8 @@ public class TablePath implements Serializable {
     // RecordAccumulator.ready)
     private Integer hash;
```
swuferhong, line 56: Can u add some comments for this variable. Like explain why we introduce this prefix.

fluss-common/src/main/java/org/apache/fluss/metadata/TablePath.java
```diff
@@ -127,20 +129,35 @@ public String toString() {
...
+                    "Table name "
+                            + tableName
+                            + " is invalid: "
+                            + (tableError != null ? tableError : tableInternalNameError));
         }
     }
 
+    public static String validatePrefix(String identifier) throws InvalidTableException {
+        if (identifier != null && identifier.startsWith(INTERNAL_NAME_PREFIX)) {
+            return "'" + INTERNAL_NAME_PREFIX + "' is not allowed as prefix";
```
swuferhong: The error message appears too simple. Could we include an explanation of why '__' cannot be used, since '__' is reserved for internal tables/internal partitions in Fluss server?

fluss-common/src/test/java/org/apache/fluss/utils/PartitionUtilsTest.java
```diff
@@ -52,18 +52,23 @@ class PartitionUtilsTest {
...
                 .isInstanceOf(InvalidPartitionException.class)
                 .hasMessageContaining(
                         "The partition value ?1 is invalid: '?1' contains one or more "
                                 + "characters other than ASCII alphanumerics, '_' and '-'");
 
+        assertThatThrownBy(() -> validatePartitionValues(Arrays.asList("__p1", "2"), true))
+                .isInstanceOf(InvalidPartitionException.class)
+                .hasMessageContaining(
+                        "The partition value __p1 is invalid: '__' is not allowed as prefix");
+
```
swuferhong, line 76: add ut when `isCreate` equals `false`.

swuferhong (commented):
Thanks @caozhen1937, LGTM. One more comment: Could you please add a test in the client's `FlussAdminITCase` that covers creating a database, creating a table, and creating partitions with a prefix?

swuferhong (approved):
LGTM

PR #1719, opened 2025-09-18
[common] Change Cluster#getRandomTabletServer() to really random retrieve tabletServer

### LINKED ISSUE #1718: Cluster#getRandomTabletServer() does not truly randomly retrieve a TabletServer
### Search before asking

### Fluss version

main (development)

### Please describe the bug 🐞

`Cluster#getRandomTabletServer()` does not truly randomly retrieve a `TabletServer`.  The current implementation of `getRandomTabletServer` always retrieves the first element from the map iterator, which results in consistently selecting `TabletServer-0`. 

`
public ServerNode getRandomTabletServer() {
        // TODO this method need to get one tablet server according to the load.
        List<ServerNode> serverNodes = new ArrayList<>(aliveTabletServersById.values());
        return !serverNodes.isEmpty() ? serverNodes.get(0) : null;
    }
`

PR #1303 changes most `read-only admin operations` through the `getRandomTabletServer` method. Since the selection is not actually random, this places excessive pressure on `TabletServer-0`—particularly for `Admin#listPartitionInfos(TablePath tablePath)` requests, which often take 5–7 seconds. This issue does not involve optimizing `Admin#listPartitionInfos(TablePath tablePath)`; it only aims to change the pseudo-random selection into a truly random one, thereby reducing the load on `TabletServer-0`.

### Solution

_No response_

swuferhong:
### Purpose

Linked issue: close #1718 

`Cluster#getRandomTabletServer()` does not truly randomly retrieve a `TabletServer`. The current implementation of `getRandomTabletServer` always retrieves the first element from the map iterator, which results in consistently selecting `TabletServer-0`.

```java
public ServerNode getRandomTabletServer() {
        // TODO this method need to get one tablet server according to the load.
        List<ServerNode> serverNodes = new ArrayList<>(aliveTabletServersById.values());
        return !serverNodes.isEmpty() ? serverNodes.get(0) : null;
    }
```

This pr is aims to change `Cluster#getRandomTabletServer()` to really random retrieve `tabletServer`.

### DIFF
--- a/fluss-common/src/main/java/org/apache/fluss/cluster/Cluster.java
+++ b/fluss-common/src/main/java/org/apache/fluss/cluster/Cluster.java
@@ -34,6 +34,8 @@ import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
+import java.util.concurrent.ThreadLocalRandom;
+import java.util.concurrent.atomic.AtomicLong;
 
 /**
  * An immutable representation of a subset of the server nodes, tables, and buckets and schemas in
@@ -44,6 +46,9 @@ import java.util.Set;
  */
 @Internal
 public final class Cluster {
+
+    private static final AtomicLong RANDOM_SEED = new AtomicLong(System.nanoTime());
+
     @Nullable private final ServerNode coordinatorServer;
     private final Map<PhysicalTablePath, List<BucketLocation>> availableLocationsByPath;
     private final Map<TableBucket, BucketLocation> availableLocationByBucket;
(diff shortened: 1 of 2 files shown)


fluss-common/src/main/java/org/apache/fluss/cluster/Cluster.java
```diff
@@ -209,7 +214,13 @@ public ServerNode getTabletServer(int id) {
     public ServerNode getRandomTabletServer() {
         // TODO this method need to get one tablet server according to the load.
         List<ServerNode> serverNodes = new ArrayList<>(aliveTabletServersById.values());
-        return !serverNodes.isEmpty() ? serverNodes.get(0) : null;
+        if (serverNodes.isEmpty()) {
+            return null;
+        }
+
+        long seedMix = RANDOM_SEED.get() ^ ThreadLocalRandom.current().nextLong();
```
wuchong: Is there any reason to use in this way? Why not use `ThreadLocalRandom.current().nextLong()` as the random value directly?

fluss-common/src/test/java/org/apache/fluss/cluster/ClusterTest.java
```diff
@@ -130,7 +130,35 @@ void testInvalidMetaAndUpdate() {
...
+    void testGetRandomTabletServer() {
+        Map<Integer, ServerNode> aliveTabletServersById = new HashMap<>();
+        for (int i = 0; i < 1000; i++) {
+            aliveTabletServersById.put(
+                    i, new ServerNode(i, "localhost", 99 + i, ServerType.TABLET_SERVER));
+        }
+        Cluster cluster = createCluster(aliveTabletServersById);
+
+        Map<ServerNode, Integer> selectionCount = new HashMap<>();
+        int selectionNumber = 100000;
```
wuchong: I think this test is over-designed. We don't need to test the random distribution, because we indeed uses Java Random util. We just need to test the result is not always the same node. E.g., 

Given 5 nodes, and get 5 nodes,  not all the nodes are the same.

PR #1724, opened 2025-09-18
[client/flink] generate javadoc for public modules

### LINKED ISSUE #1712: Setup Javadoc for fluss API discoverability
### Search before asking

### Description

We don't have the Javadoc yet, but doing so would be a major step forward for our developer ecosystem. It would provide a single, authoritative source of truth for our APIs, making it easier for developers to explore available classes, methods, and parameters without digging into the codebase. Beyond improving API discoverability, well-documented Javadocs speed up onboarding for new developers, reduce the learning curve, and lower the number of support queries.

### Willingness to contribute

MehulBatra:
### Purpose
Linked issue: close #1712

Implement  javadoc generation

### Brief change log

- Added javadoc-aggregate Maven profile in root pom.xml to generate aggregate javadoc for API modules
- Updated refresh_javadoc.sh script to use Maven profile
- Configured profile to include only API-relevant modules (fluss-client, fluss-flink) while excluding modules (fluss-protogen, fluss-jmh, fluss-test-*, fluss-metrics, common, server etc.)
- Enhanced javadoc plugin configuration with proper error handling, package exclusions, and Java 11 module exports
- Added version-aware output directory structure (website/static/javadoc/VERSION/)
- Created automatic redirect page for latest version navigation

### Tests

Manual testing of javadoc generation script with different scenarios
Verified generated HTML output structure and content
Validated redirect functionality and version-specific URL routing

### API and Format
No API changes - This is purely a build/documentation improvement that doesn't affect runtime APIs.

### Documentation
 The generated javadoc itself serves as API reference documentation.


wuchong:
Thanks for the great work @MehulBatra !  

Here are my thoughts about the javadoc design:

1. I don't want to maintain the generated javadoc in the main repo and branches, this will quickly increase our main repo size. which is very heavy. 
2. We can maintain each generated javadocs at the `apache/fluss-website` in the branches `javadoc-<version>`. 
3. We need to maintain javadocs for each version, including the released (not include the `com.alibaba.fluss` versions) and unreleased (`main` branch). 

Here are suggestions about how to implement: 

1. Create a new script `build_javadocs.sh` under `website/` along with `build_versioned_docs.sh`
 - The script just generates javadocs for the current branch, the generated javadocs can be put in the folder `website/genearted_javadocs/` (add this directory into `.gitignore`)
2. Add a `build_javadocs.yaml` in apache/fluss-website  https://github.com/apache/fluss-website/tree/main/.github/workflows
 - it's a daily scheduled github action
 - it checkout apache/fluss repo and for each `release-*` (exclude <= `release-0.7` branches) and `main` branch
 - for each branch, it runs the `build_javadocs.sh` script and commit the files in `website/genearted_javadocs/` to the branch `javadoc-<version>` of apache/fluss-website
 - maybe we can skip branches if the branch has no git log changes in 7 days, to reduce the build cost. 
 - I think we can get the SHORT_VERSION from the branch name easily, so no need to use mvn command to parse
 3. Update `website-deploy.yaml` in apache/fluss-website, add a step `Copy Javadocs` before the `Deploy website`. 
   - it copies javadocs files from branches `javadoc-<version>` to the directory `./build/javadoc/<version>`
   - you can learn the code from `build_versioned_docs.sh` in apache/fluss which has similar code. 
 4. Add javadoc links in our official docs.

MehulBatra:
Thanks for the feedback @wuchong , I agree on all the points will implement on these lines!

MehulBatra:
Github Action PR: https://github.com/apache/fluss-website/pull/1

MehulBatra:
image
image

website/build_javadocs.sh
```diff
@@ -0,0 +1,181 @@
...
+echo "Output directory: $OUTPUT_DIR"
+
+# Create output directory
+rm -rf "$OUTPUT_DIR"
+mkdir -p "$OUTPUT_DIR"
+
+# Create temporary POM for targeted javadoc generation
+TEMP_POM="temp-javadoc-pom.xml"
+cat > "$TEMP_POM" << EOF
+<?xml version="1.0" encoding="UTF-8"?>
```
wuchong, line 72: I remembered in the first version, you made the javadoc plugin changes in the root `pom.xml`. Why using a temp pom xml here? I think it's better to keep all the pom changes in the same place, so it's easier for future developers to modify both places.

website/build_javadocs.sh
```diff
@@ -0,0 +1,181 @@
...
+
+    <build>
+        <plugins>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-javadoc-plugin</artifactId>
+                <version>3.6.3</version>
+                <configuration>
+                    <windowTitle>Apache Fluss \${project.version} API</windowTitle>
+                    <doctitle>Apache Fluss \${project.version} API</doctitle>
```
wuchong, line 102: If we move the changes to root pom, we can use `build-helper-maven-plugin` to parse project version into without `-SNAPSHOT` suffix (e.g., the following `base.version` property). 


```xml
<build>
  <plugins>
    <plugin>
      <groupId>org.codehaus.mojo</groupId>
      <artifactId>build-helper-maven-plugin</artifactId>
      <version>3.5.0</version>
      <executions>
        <execution>
          <id>parse-version</id>
          <goals>
            <goal>parse-version</goal>
          </goals>
        </execution>
      </executions>
    </plugin>
  </plugins>
</build>
```

```xml
<properties>
  <base.version>${parsedVersion.majorVersion}.${parsedVersion.minorVersion}</base.version>
</properties>
```

website/build_javadocs.sh
```diff
@@ -0,0 +1,181 @@
...
+
+    <properties>
+        <maven.compiler.source>11</maven.compiler.source>
+        <maven.compiler.target>11</maven.compiler.target>
+        <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+    </properties>
+
+    <modules>
+        <module>fluss-client</module>
+        <module>fluss-flink/fluss-flink-common</module>
```
wuchong, line 91: I think we need to include `fluss-common`, `fluss-rpc` as well, most client side API are in `fluss-common`.

website/build_javadocs.sh
```diff
@@ -0,0 +1,181 @@
...
+<head>
+    <meta charset="UTF-8">
+    <meta http-equiv="refresh" content="0; url=${JAVADOC_VERSION}/">
+    <title>Apache Fluss API Documentation</title>
+</head>
+<body>
+    <p>Redirecting to <a href="${JAVADOC_VERSION}/">Apache Fluss ${JAVADOC_VERSION} API</a></p>
+</body>
+</html>
+EOF
```
wuchong, line 165: Do we really need this `index.html`?  As it is under `javadoc/index.html`, so multiple javadoc version will override this file.

wuchong:
Thank you @MehulBatra for the updates. The new architecture looks very good. I think there is only some minor improvements to do. Besides, I think we don't need a navbar for Javadoc, because Javadoc is also versioned, so it should be going into `Docs`, maybe under `Docs -> Fluss APIs -> Javadocs`

PR #1729, opened 2025-09-20
[lake] Introduce lake lag metrics for datalake enabled table

### LINKED ISSUE #1599: Introduce lake lag metrics for datalake enabled table
### Search before asking

### Description

It'll be better to introduce a lag metrics to mark the lag between the log offset in datalake and the log end offset.

### Willingness to contribute

caozhen1937:
### Purpose

Linked issue: close #1599

### DIFF
--- a/fluss-common/src/main/java/org/apache/fluss/metrics/MetricNames.java
+++ b/fluss-common/src/main/java/org/apache/fluss/metrics/MetricNames.java
@@ -132,8 +132,8 @@ public class MetricNames {
     public static final String LOG_NUM_SEGMENTS = "numSegments";
     public static final String LOG_END_OFFSET = "endOffset";
     public static final String REMOTE_LOG_SIZE = "size";
-    public static final String LOG_LAKE_LAG_RECORDS = "lakeLagRecords";
-    public static final String LOG_LAKE_LAG_TIME_MS = "lakeLagTimestampMs";
+    public static final String LOG_LAKE_PENDING_RECORDS = "pendingRecords";
+    public static final String LOG_LAKE_TIMESTAMP_LAG = "timestampLag";
 
     // for logic storage
     public static final String LOCAL_STORAGE_LOG_SIZE = "logSize";
--- a/fluss-common/src/main/java/org/apache/fluss/metrics/MetricNames.java
+++ b/fluss-common/src/main/java/org/apache/fluss/metrics/MetricNames.java
@@ -132,6 +132,8 @@ public class MetricNames {
     public static final String LOG_NUM_SEGMENTS = "numSegments";
     public static final String LOG_END_OFFSET = "endOffset";
     public static final String REMOTE_LOG_SIZE = "size";
+    public static final String LOG_LAKE_LAG_RECORDS = "lakeLagRecords";
+    public static final String LOG_LAKE_LAG_TIME_MS = "lakeLagTimestampMs";
 
     // for logic storage
     public static final String LOCAL_STORAGE_LOG_SIZE = "logSize";
--- a/fluss-server/src/main/java/org/apache/fluss/server/log/LogTablet.java
+++ b/fluss-server/src/main/java/org/apache/fluss/server/log/LogTablet.java
@@ -346,6 +346,18 @@ public final class LogTablet {
         metricGroup.gauge(
                 MetricNames.LOG_NUM_SEGMENTS, () -> localLog.getSegments().numberOfSegments());
         metricGroup.gauge(MetricNames.LOG_END_OFFSET, localLog::getLocalLogEndOffset);
+        metricGroup.gauge(
+                MetricNames.LOG_LAKE_LAG_RECORDS,
+                () ->
+                        getLakeLogEndOffset() < 0L
+                                ? localLogEndOffset()
+                                : localLogEndOffset() - getLakeLogEndOffset());
+        metricGroup.gauge(
+                MetricNames.LOG_LAKE_LAG_TIME_MS,
+                () ->
+                        getLakeMaxTimestamp() < 0L
+                                ? localMaxTimestamp()
+                                : localMaxTimestamp() - getLakeMaxTimestamp());
     }
 
     public long logSize() {
--- a/fluss-server/src/main/java/org/apache/fluss/server/log/LogTablet.java
+++ b/fluss-server/src/main/java/org/apache/fluss/server/log/LogTablet.java
@@ -346,18 +346,6 @@ public final class LogTablet {
         metricGroup.gauge(
                 MetricNames.LOG_NUM_SEGMENTS, () -> localLog.getSegments().numberOfSegments());
         metricGroup.gauge(MetricNames.LOG_END_OFFSET, localLog::getLocalLogEndOffset);
-        metricGroup.gauge(
-                MetricNames.LOG_LAKE_LAG_RECORDS,
-                () ->
-                        getLakeLogEndOffset() < 0L
-                                ? localLogEndOffset()
-                                : localLogEndOffset() - getLakeLogEndOffset());
-        metricGroup.gauge(
-                MetricNames.LOG_LAKE_LAG_TIME_MS,
-                () ->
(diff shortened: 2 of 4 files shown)


fluss-server/src/main/java/org/apache/fluss/server/log/LogTablet.java
```diff
@@ -346,6 +346,18 @@ public void registerMetrics(BucketMetricGroup bucketMetricGroup) {
         metricGroup.gauge(
                 MetricNames.LOG_NUM_SEGMENTS, () -> localLog.getSegments().numberOfSegments());
         metricGroup.gauge(MetricNames.LOG_END_OFFSET, localLog::getLocalLogEndOffset);
+        metricGroup.gauge(
```
luoyuxia: Not to register this metric in this tablet-level group. One bucket will have 3 log tablet, register in here will cause we have 3 lag metric. But we just need one, only the leader should report it.

luoyuxia: Also only datalake enabled table should report this metrics, other tables shouldn't report this metrics.

luoyuxia: Suggest to do like that:
```
private void onBecomeNewLeader() {
        updateLeaderEndOffsetSnapshot();

        if (isDataLakeEnabled()) {
            registerLakeTieringMetrics();
        }

        if (isKvTable()) {
            // if it's become new leader, we must
            // first destroy the old kv tablet
            // if exist. Otherwise, it'll use still the old kv tablet which will cause data loss
            dropKv();
            // now, we can create a new kv tablet
            createKv();
        }
    }

    private void registerLakeTieringMetrics() {
        lakeTieringMetricGroup = bucketMetricGroup.addGroup("lakeTiering");
        lakeTieringMetricGroup.gauge(
                MetricNames.LOG_LAKE_LAG_RECORDS,
                () ->
                        getLakeLogEndOffset() < 0L
                                ? getLogHighWatermark()
                                : getLogHighWatermark() - getLakeLogEndOffset());
        lakeTieringMetricGroup.gauge(
                MetricNames.LOG_LAKE_LAG_TIME_MS,
                () ->
                        logTablet.getLakeMaxTimestamp() < 0L
                                ? logTablet.getLakeMaxTimestamp()
                                : logTablet.localMaxTimestamp() - logTablet.getLakeMaxTimestamp());
    }

    private void onBecomeNewFollower() {
        if (isKvTable()) {
            // it should be from leader to follower, we need to destroy the kv tablet
            dropKv();
        }
        if (lakeTieringMetricGroup != null) {
            lakeTieringMetricGroup.close();
        }
    }
```

fluss-common/src/main/java/org/apache/fluss/metrics/MetricNames.java
```diff
@@ -132,6 +132,8 @@ public class MetricNames {
     public static final String LOG_NUM_SEGMENTS = "numSegments";
     public static final String LOG_END_OFFSET = "endOffset";
     public static final String REMOTE_LOG_SIZE = "size";
+    public static final String LOG_LAKE_LAG_RECORDS = "lakeLagRecords";
```
luoyuxia: I don't have a good naming in my mind, maybe

```suggestion
// for lake tiering
    public static final String LOG_LAKE_PENDING_RECORDS = "pendingRecords";
```

Inspired from Flink https://cwiki.apache.org/confluence/display/FLINK/FLIP-33%3A+Standardize+Connector+Metrics

fluss-common/src/main/java/org/apache/fluss/metrics/MetricNames.java
```diff
@@ -132,6 +132,8 @@ public class MetricNames {
...
+    public static final String LOG_LAKE_LAG_TIME_MS = "lakeLagTimestampMs";
```
luoyuxia:
```suggestion
    public static final String LOG_LAKE_TIMESTAMP_LAG = "timestampLag";
```

website/docs/maintenance/observability/monitor-metrics.md
```diff
@@ -757,6 +757,16 @@ Some metrics might not be exposed when using other JVM implementations (e.g. IBM
       <td>endOffset</td>
       <td>The end offset in local storage for this table bucket.</td>
       <td>Gauge</td>
+    </tr>
+     <tr>
+      <td>lakeLagRecords</td>
```
luoyuxia: If modify accoding to my suggestion, then the `infix` should be `table_bucket_lakeTiering`

luoyuxia (commented):
@caozhen1937 Thanks for the pr. Left some comments. PTAL.

fluss-common/src/main/java/org/apache/fluss/metrics/MetricNames.java
```diff
@@ -132,6 +132,8 @@ public class MetricNames {
     public static final String LOG_NUM_SEGMENTS = "numSegments";
     public static final String LOG_END_OFFSET = "endOffset";
     public static final String REMOTE_LOG_SIZE = "size";
+    public static final String LOG_LAKE_PENDING_RECORDS = "pendingRecords";
```
luoyuxia, line 135: Then, the metrics for lake tiering name will be :
- for the timestamp lag(the lag between latest timestamp of lake and fluss): `table_bucket_lakeTiering_timestampLag`
- for the records lag(how many records are not tiered into lake): `table_bucket_lakeTiering_pendingRecords`

@wuchong What do you think of it. Any suggestion on the naming ?

fluss-server/src/main/java/org/apache/fluss/server/replica/Replica.java
```diff
@@ -540,11 +546,30 @@ private void onBecomeNewLeader() {
...
+                MetricNames.LOG_LAKE_PENDING_RECORDS,
+                () ->
+                        getLakeLogEndOffset() < 0L
+                                ? getLogHighWatermark()
+                                : getLogHighWatermark() - getLakeLogEndOffset());
+        lakeTieringMetricGroup.gauge(
+                MetricNames.LOG_LAKE_TIMESTAMP_LAG,
+                () ->
+                        logTablet.getLakeMaxTimestamp() < 0L
+                                ? logTablet.getLakeMaxTimestamp()
```
luoyuxia:
```suggestion
                                ? -1
```
How about using -1 to avoid `logTablet.getLakeMaxTimestamp()` being something weird?

caozhen1937: @luoyuxia it make sense for me, and should `LOG_LAKE_LAG_RECORDS`  follow it?

caozhen1937: @luoyuxia I have another idea. The `pendingRecords` and the `timestampLag` are the two indicators representing the delay situation. Then, the `LakeMaxTimestamp` and `LakeLogEndOffset` can only be negative when they are initialized, right? So in this case, can we return localMaxTimestamp and LocalLogEndOffset?

luoyuxia: > @luoyuxia I have another idea. The `pendingRecords` and the `timestampLag` are the two indicators representing the delay situation. Then, the `LakeMaxTimestamp` and `LakeLogEndOffset` can only be negative when they are initialized, right? So in this case, can we return localMaxTimestamp and LocalLogEndOffset?

I suggest to return -1 explicitly

luoyuxia: > @luoyuxia it make sense for me, and should `LOG_LAKE_LAG_RECORDS` follow it?

Yes, I think `LOG_LAKE_LAG_RECORDS` follow it, return high watermark is not correct since log will ttl, high watermark  can't represent pending records.

caozhen1937: ok, let me change them to `-1`.

wuchong (approved):

PR #1732, opened 2025-09-21
[lake] The bucket of PaimonSplit should be -1 for bucket unwared paim…

### LINKED ISSUE #1678: The bucket of PaimonSplit should be -1 for bucket unwared paimon table
### Search before asking

### Description

As the java doc of `LakeSplit` said, when we don't care about bucket of paimon splits, we should return -1 in `bucket` method. Flink source will recongnize if bucket is ` -1` and then assign to any reader subtask. Otherwise,  the bucket of the data split will all be 0 for bucket unwared table, which cause it always assign to single one task.

### Willingness to contribute

caozhen1937:
…on table

### Purpose

Linked issue: close #1678

### DIFF
--- a/fluss-lake/fluss-lake-paimon/src/main/java/org/apache/fluss/lake/paimon/source/PaimonSplitPlanner.java
+++ b/fluss-lake/fluss-lake-paimon/src/main/java/org/apache/fluss/lake/paimon/source/PaimonSplitPlanner.java
@@ -68,14 +68,14 @@ public class PaimonSplitPlanner implements Planner<PaimonSplit> {
             try (Catalog catalog = getCatalog()) {
                 FileStoreTable fileStoreTable = getTable(catalog, tablePath, snapshotId);
                 InnerTableScan tableScan = fileStoreTable.newScan();
-                boolean isBucketAware = fileStoreTable.bucketMode() == BucketMode.BUCKET_UNAWARE;
+                boolean isBucketUnAware = fileStoreTable.bucketMode() == BucketMode.BUCKET_UNAWARE;
 
                 if (predicate != null) {
                     tableScan = tableScan.withFilter(predicate);
                 }
                 for (Split split : tableScan.plan().splits()) {
                     DataSplit dataSplit = (DataSplit) split;
-                    splits.add(new PaimonSplit(dataSplit, isBucketAware));
+                    splits.add(new PaimonSplit(dataSplit, isBucketUnAware));
                 }
             }
             return splits;
--- a/fluss-lake/fluss-lake-paimon/src/main/java/org/apache/fluss/lake/paimon/source/PaimonSplitSerializer.java
+++ b/fluss-lake/fluss-lake-paimon/src/main/java/org/apache/fluss/lake/paimon/source/PaimonSplitSerializer.java
@@ -45,7 +45,7 @@ public class PaimonSplitSerializer implements SimpleVersionedSerializer<PaimonSp
         DataOutputViewStreamWrapper view = new DataOutputViewStreamWrapper(out);
         DataSplit dataSplit = paimonSplit.dataSplit();
         InstantiationUtil.serializeObject(view, dataSplit);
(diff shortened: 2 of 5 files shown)


fluss-lake/fluss-lake-paimon/src/test/java/org/apache/fluss/lake/paimon/source/PaimonSplitPlannerTest.java
```diff
@@ -78,4 +79,37 @@ void testPlan() throws Exception {
                                 .map(PaimonSplit::dataSplit)
                                 .collect(Collectors.toList()));
     }
+
+    @Test
+    void testPlannerCreatesCorrectSplitsForLogTableWithoutBucketKey() throws Exception {
+        // 测试没有bucket key的log表
```
beryllw: Translate into English.

fluss-lake/fluss-lake-paimon/src/test/java/org/apache/fluss/lake/paimon/source/PaimonSplitPlannerTest.java
```diff
@@ -78,4 +79,37 @@ void testPlan() throws Exception {
...
+                        GenericRow.of(1, BinaryString.fromString("msg1")),
+                        GenericRow.of(2, BinaryString.fromString("msg2")));
+        writeRecord(tablePath, records);
+
+        Snapshot snapshot = table.latestSnapshot().get();
+
+        LakeSource<PaimonSplit> lakeSource = lakeStorage.createLakeSource(tablePath);
+        List<PaimonSplit> paimonSplits = lakeSource.createPlanner(snapshot::id).plan();
+
+        List<Split> actualSplits = ((FileStoreTable) table).newScan().plan().splits();
```
beryllw: Why do we need the actualSplits? Isn't paimonSplits sufficient?

caozhen1937: Yes, you're right. This is a clerical error.

fluss-lake/fluss-lake-paimon/src/main/java/org/apache/fluss/lake/paimon/source/PaimonSplitPlanner.java
```diff
@@ -67,12 +67,14 @@ public List<PaimonSplit> plan() {
             try (Catalog catalog = getCatalog()) {
                 FileStoreTable fileStoreTable = getTable(catalog, tablePath, snapshotId);
                 InnerTableScan tableScan = fileStoreTable.newScan();
+                boolean isBucketAware = isBucketAware(fileStoreTable);
```
luoyuxia:
```suggestion
                boolean isBucketAware = fileStoreTable.bucketMode() == BucketMode.BUCKET_UNAWARE;
```

fluss-lake/fluss-lake-paimon/src/main/java/org/apache/fluss/lake/paimon/source/PaimonSplitSerializer.java
```diff
@@ -26,14 +26,20 @@
...
 import java.io.ByteArrayOutputStream;
+import java.io.DataInputStream;
 import java.io.IOException;
 
 /** Serializer for paimon split. */
 public class PaimonSplitSerializer implements SimpleVersionedSerializer<PaimonSplit> {
 
+    // Version 1: Original version with only DataSplit
+    // Version 2: Added isBucketAware field for bucket-unaware table support
+    private static final int VERSION_1 = 1;
```
luoyuxia, line 35: we can just keep the version as still 1 since 0.8 is the only  one version that introduces `PaimonSplitSerializer`, we do change any thing during 0.8. 
But once 0.8 is released, any change in `PaimonSplitSerializer` should keep compatibility. But now, we don't need  compatibility.

Issue #1733, opened 2025-09-21
TabletServerShutdownITCase.testControlledShutdownRetriesFailover is unstable


wuchong:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/fluss/issues) and found nothing similar.

### Fluss version

main (development)

### Please describe the bug 🐞

https://github.com/apache/fluss/actions/runs/17894407292/job/50879072671

```
Error:  Tests run: 4, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 80.616 s <<< FAILURE! - in org.apache.fluss.server.tablet.TabletServerShutdownITCase
Error:  org.apache.fluss.server.tablet.TabletServerShutdownITCase.testControlledShutdownRetriesFailover  Time elapsed: 69.444 s  <<< FAILURE!
org.opentest4j.AssertionFailedError: 

expected: LeaderAndIsr{leader=1, leaderEpoch=0, isr=[1], coordinatorEpoch=0, bucketEpoch=1}
 but was: LeaderAndIsr{leader=1, leaderEpoch=0, isr=[1, 2], coordinatorEpoch=0, bucketEpoch=2}
	at jdk.internal.reflect.GeneratedConstructorAccessor76.newInstance(Unknown Source)
	at java.base/jdk.internal.reflect.DelegatingConstructorAccessorImpl.newInstance(DelegatingConstructorAccessorImpl.java:45)
	at org.apache.fluss.server.tablet.TabletServerShutdownITCase.lambda$testControlledShutdownRetriesFailover$2(TabletServerShutdownITCase.java:180)
	at org.apache.fluss.testutils.common.CommonTestUtils.retry(CommonTestUtils.java:142)
	at org.apache.fluss.server.tablet.TabletServerShutdownITCase.testControlledShutdownRetriesFailover(TabletServerShutdownITCase.java:176)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:566)
	at org.junit.platform.commons.util.ReflectionUtils.invokeMethod(ReflectionUtils.java:727)
[...]

PR #1734, opened 2025-09-22
[server] Coordinator reject the adjustIsr request if the expected adjusted isr contains shutdown tabletServers

swuferhong:
### Purpose

Linked issue: #1733 

The  `TabletServerShutdownITCase.testControlledShutdownRetriesFailover()` is unstable because` coordinator` would receive and accept the `excepted isr set` in ·AdjustIsr· request even if the expected isr set contains the shutdown `tabletServers`. `Coordinator` reject the `adjustIsr` request if the expected adjusted isr contains shutdown `tabletServers`


swuferhong:
@wuchong could you take a look. Fix the unstable ITCase.

PR #1736, opened 2025-09-22
[client] Set the default value of 'scan.partition.discovery.interval' to 1min to reduce the server load

swuferhong:
### Purpose

This pr is aims to set the default value of `scan.partition.discovery.interval` to `1min` to reduce the server load.  Currently, since Fluss `Admin#listPartitions(TablePath tablePath)` requires a large number of requests to ZooKeeper in server, for table with lots of partitions, this option cannot be set too small, as a small value would cause frequent requests and increase server load. In the future, once list partitions is optimized, the default value of this parameter can be reduced.


wuchong:
@swuferhong it seems the Flink test is failed because of this change. Could you take a look?

swuferhong:
@wuchong The tests passed.

wuchong (approved):

PR #1738, opened 2025-09-22
[KV] Avoid TabletServer to discard already commited kv snapshot

### LINKED ISSUE #1304: TabletServer may discard already commited kv snapshot

platinumhamburg:
### Purpose

Linked issue: close #1304

### DIFF
--- a/fluss-server/src/main/java/org/apache/fluss/server/kv/snapshot/KvTabletSnapshotTarget.java
+++ b/fluss-server/src/main/java/org/apache/fluss/server/kv/snapshot/KvTabletSnapshotTarget.java
@@ -58,7 +58,7 @@ public class KvTabletSnapshotTarget implements PeriodicSnapshotManager.SnapshotT
 
     private final CompletedKvSnapshotCommitter completedKvSnapshotCommitter;
 
-    @Nonnull private final ZooKeeperClient zooKeeperClient;
+    private final ZooKeeperClient zooKeeperClient;
 
     private final RocksIncrementalSnapshot rocksIncrementalSnapshot;
     private final FsPath remoteKvTabletDir;
@@ -87,7 +87,7 @@ public class KvTabletSnapshotTarget implements PeriodicSnapshotManager.SnapshotT
     KvTabletSnapshotTarget(
             TableBucket tableBucket,
             CompletedKvSnapshotCommitter completedKvSnapshotCommitter,
-            @Nonnull ZooKeeperClient zooKeeperClient,
+            ZooKeeperClient zooKeeperClient,
             RocksIncrementalSnapshot rocksIncrementalSnapshot,
             FsPath remoteKvTabletDir,
             Executor ioExecutor,
@@ -289,7 +289,6 @@ public class KvTabletSnapshotTarget implements PeriodicSnapshotManager.SnapshotT
         //
         // Idempotent check: Double check ZK to verify if the snapshot actually exists before
         // cleanup
-        boolean shouldCleanSnapshot = true;
         try {
             Optional<BucketSnapshot> zkSnapshot =
                     zooKeeperClient.getTableBucketSnapshot(tableBucket, snapshotId);
@@ -305,23 +304,30 @@ public class KvTabletSnapshotTarget implements PeriodicSnapshotManager.SnapshotT
 
                 // Update local state as if the commit was successful
                 updateStateOnCommitSuccess(snapshotId, snapshotResult);
-
-                shouldCleanSnapshot = false;
                 return; // Snapshot commit succeeded, return directly
+            } else {
+                // Snapshot does not exist in ZK, indicating the commit truly failed
+                LOG.warn(
+                        "Snapshot {} for TableBucket {} does not exist in ZK. "
+                                + "The commit truly failed, proceeding with cleanup.",
+                        snapshotId,
+                        tableBucket);
+                snapshotsCleaner.cleanSnapshot(completedSnapshot, () -> {}, ioExecutor);
+                handleSnapshotFailure(snapshotId, snapshotLocation, t);
             }
         } catch (Exception zkException) {
             LOG.warn(
                     "Failed to query ZK for snapshot {} of TableBucket {}. "
-                            + "Assuming commit failed and proceeding with cleanup.",
+                            + "Cannot determine actual snapshot status, keeping snapshot in current state "
+                            + "to avoid potential data loss.",
                     snapshotId,
                     tableBucket,
                     zkException);
-        }
-
-        // Only execute cleanup when the snapshot truly does not exist in ZK
-        if (shouldCleanSnapshot) {
-            snapshotsCleaner.cleanSnapshot(completedSnapshot, () -> {}, ioExecutor);
-            handleSnapshotFailure(snapshotId, snapshotLocation, t);
+            // When ZK query fails, we cannot determine the actual status.
+            // The snapshot might have succeeded or failed on the ZK side.
+            // Therefore, we must not clean up the snapshot files and not update local state.
+            // This avoids the risk of discarding a successfully committed snapshot that
+            // connectors may already be reading, which would cause data loss or job failure.
         }
 
         // throw the exception to make PeriodicSnapshotManager can catch the exception
--- a/fluss-server/src/test/java/org/apache/fluss/server/kv/snapshot/KvTabletSnapshotTargetTest.java
+++ b/fluss-server/src/test/java/org/apache/fluss/server/kv/snapshot/KvTabletSnapshotTargetTest.java
@@ -400,7 +400,7 @@ class KvTabletSnapshotTargetTest {
 
     @Test
     void testIdempotentCheckWhenZKQueryFails(@TempDir Path kvTabletDir) throws Exception {
-        // Test case: ZK query fails - should fall back to cleanup (conservative approach)
+        // Test case: ZK query fails - should keep snapshot in current state to avoid data loss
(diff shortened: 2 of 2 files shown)


fluss-server/src/main/java/org/apache/fluss/server/kv/snapshot/KvTabletSnapshotTarget.java
```diff
@@ -219,8 +227,56 @@ public void handleSnapshotResult(
...
+            // coordinator server commit the kv snapshot to zk, then failover.
+            // Tablet server will got exception from coordinator server, but mistake it as a fail
+            // commit although coordinator server has committed to zk, then discard the commited kv
+            // snapshot.
+            //
+            // Idempotent check: Double check ZK to verify if the snapshot actually exists before
+            // cleanup
+            boolean shouldCleanSnapshot = true;
+            try {
+                if (zooKeeperClient != null) {
```
wuchong: We don't need to check the nullability of `zooKeeperClient`, it's not-null, otherwise, it should be marked as `@Nullable`.

fluss-server/src/main/java/org/apache/fluss/server/kv/snapshot/KvTabletSnapshotTarget.java
```diff
@@ -219,8 +227,56 @@ public void handleSnapshotResult(
...
+                                        + "The commit was successful but response was lost due to coordinator failover. "
+                                        + "Skipping cleanup and treating as successful.",
+                                snapshotId,
+                                tableBucket);
+
+                        // Update local state as if the commit was successful
+                        rocksIncrementalSnapshot.notifySnapshotComplete(snapshotId);
+                        logOffsetOfLatestSnapshot = snapshotResult.getLogOffset();
+                        snapshotSize = snapshotResult.getSnapshotSize();
+                        updateMinRetainOffset.accept(snapshotResult.getLogOffset());
```
wuchong: extract this into a private method `handleSuccessfulSnapshot(SnapshotResult)`, and reuse with commit-kv successful code path.

fluss-server/src/main/java/org/apache/fluss/server/kv/snapshot/KvTabletSnapshotTarget.java
```diff
@@ -219,8 +227,56 @@ public void handleSnapshotResult(
             updateMinRetainOffset.accept(snapshotResult.getLogOffset());
         } catch (Exception e) {
             Throwable t = ExceptionUtils.stripExecutionException(e);
-            snapshotsCleaner.cleanSnapshot(completedSnapshot, () -> {}, ioExecutor);
-            handleSnapshotFailure(snapshotId, snapshotLocation, t);
+
+            // Fix for issue: https://github.com/apache/fluss/issues/1304
```
wuchong: This is a very long method, we can extract the check&cleanup logic into a private method to improve code readability, e.g. `handleSnapshotCommitFailed()`,

fluss-server/src/main/java/org/apache/fluss/server/kv/snapshot/KvTabletSnapshotTarget.java
```diff
@@ -219,8 +227,56 @@ public void handleSnapshotResult(
...
+                        updateMinRetainOffset.accept(snapshotResult.getLogOffset());
+
+                        shouldCleanSnapshot = false;
+                        return; // Snapshot commit succeeded, return directly
+                    }
+                }
+            } catch (Exception zkException) {
+                LOG.warn(
+                        "Failed to query ZK for snapshot {} of TableBucket {}. "
+                                + "Assuming commit failed and proceeding with cleanup.",
```
wuchong: I still think we shouldn't cleanup the snapshot if zk request failed, because it may exist in zk. We should only cleanup the snapshot when zk truly doesn't have this snapshot id. 

It's fine if we don't cleanup some unkown snapshots, the orphan files in remote storage should be cleanup by a dedicated service. It's also fine to not update local state for the snapshot, as it will eventually be updated by the next successful snapshot, and this doesn't affect data correctness.

fluss-server/src/test/java/org/apache/fluss/server/kv/snapshot/KvTabletSnapshotTargetTest.java
```diff
@@ -297,6 +297,139 @@ void testAddToSnapshotToStoreFail(@TempDir Path kvTabletDir) throws Exception {
...
+
+        TestRocksIncrementalSnapshot rocksIncrementalSnapshot =
+                (TestRocksIncrementalSnapshot) kvTabletSnapshotTarget.getRocksIncrementalSnapshot();
+
+        // The snapshot should be treated as successful due to idempotent check
+        // Even though commit threw exception, idempotent check should find it in ZK
+        retry(
+                Duration.ofMinutes(1),
+                () ->
+                        assertThat(rocksIncrementalSnapshot.completedSnapshots)
```
wuchong, line 354: `rocksIncrementalSnapshot.completedSnapshots` and `rocksIncrementalSnapshot.abortedSnapshots` is not thread-safe, would be better to use `ConcurrentHashMap.newKeySet()`.

fluss-server/src/main/java/org/apache/fluss/server/kv/snapshot/KvTabletSnapshotTarget.java
platinumhamburg: Should we treat this as a successful snapshot? If so, we would need to update certain local variables, which could introduce side effects and potentially lead to inconsistency exceptions. On the other hand, if we consider it a failed snapshot, no inconsistencies would occur, no orphaned files would be left behind, and the inconsistency would be automatically corrected by the mechanism introduced in https://github.com/apache/fluss/pull/1482 and https://github.com/apache/fluss/pull/1470 at a later time.

platinumhamburg: I tend to think that unless we are definitely sure the snapshot is successful, we should handle it as failed, as this will always ensure a correct result; otherwise, it may not.

platinumhamburg:
@wuchong Thank you for your review. I have addressed all the comments above, except for the handling of undetermined snapshots, which I still treat as failures. I've explained the reasoning behind this decision above. Please review again when you have time.

fluss-server/src/main/java/org/apache/fluss/server/kv/snapshot/KvTabletSnapshotTarget.java
wuchong: No, we should not treat this as either a successful or a failed snapshot.

When the ZooKeeper request fails, we cannot determine the actual status, the snapshot might have succeeded or failed on the ZK side.

Therefore, we must:

- Not clean up the snapshot files, and
- Not update the local state.

Leaving the snapshot in its current state avoids the risk of discarding a successfully committed snapshot that the connector may already be reading, which would cause data loss or job failure.

fluss-server/src/main/java/org/apache/fluss/server/kv/snapshot/KvTabletSnapshotTarget.java
```diff
@@ -55,6 +58,8 @@ public class KvTabletSnapshotTarget implements PeriodicSnapshotManager.SnapshotT
 
     private final CompletedKvSnapshotCommitter completedKvSnapshotCommitter;
 
+    @Nonnull private final ZooKeeperClient zooKeeperClient;
```
wuchong: minor: We don't need to add `@Nonnull` annotation for non-null fields, otherwise, we have so many fields need to add. We only need to and must to add `@Nullable` annotation for the nullable fields, so we know whether to check null when using it.

wuchong:
Hi @platinumhamburg , I left a comment about the snapshot cleanup, and a minor comment in the review. Please let me know your thoughts.

fluss-server/src/main/java/org/apache/fluss/server/kv/snapshot/KvTabletSnapshotTarget.java
platinumhamburg: > No, we should not treat this as either a successful or a failed snapshot.
> 
> When the ZooKeeper request fails, we cannot determine the actual status, the snapshot might have succeeded or failed on the ZK side.
> 
> Therefore, we must:
> 
> * Not clean up the snapshot files, and
> * Not update the local state.
> 
> Leaving the snapshot in its current state avoids the risk of discarding a successfully committed snapshot that the connector may already be reading, which would cause data loss or job failure.

This is indeed a better way to handle the exception. I'll update the implementation to reflect this approach.

platinumhamburg:
> Hi @platinumhamburg , I left a comment about the snapshot cleanup, and a minor comment in the review. Please let me know your thoughts.

Hi @wuchong , I agree that your suggestion is a better approach to handle the ZooKeeper query exception. I've modified my implementation accordingly and fixed another minor issue. Thanks for reviewing this again.

wuchong:
@platinumhamburg thanks for the updates. LGTM.

PR #1748, opened 2025-09-24
[flink] Improved Handling of Custom Properties During Fluss Table Creation

### LINKED ISSUE #1745: Flink create table will make all table options be custom properties
### Search before asking

### Fluss version

0.7.0 (latest release)

### Please describe the bug 🐞

In the definition, custom properties are not understood by Fluss. When using Flink to create fluss table,  the option understood by Fluss will be Fluss property. That's right. But the option understood by Fluss will also be set as custom properties.

Can be reproduce as following code:
```
void testTableConversionWithOptions() {
        Map<String, String> options = new HashMap<>();
        // forward table option & enum type
        options.put(ConfigOptions.TABLE_LOG_FORMAT.key(), "indexed");
        // forward client memory option
        options.put(ConfigOptions.CLIENT_WRITER_BUFFER_MEMORY_SIZE.key(), "64mb");
        // forward client duration option
        options.put(ConfigOptions.CLIENT_WRITER_BATCH_TIMEOUT.key(), "32s");

        ResolvedSchema schema =
                new ResolvedSchema(
                        Collections.singletonList(
                                Column.physical(
                                        "order_id",
                                        org.apache.flink.table.api.DataTypes.STRING().notNull())),
                        Collections.emptyList(),
                        null);
        CatalogTable flinkTable =
                CatalogTable.of(
                        Schema.newBuilder().fromResolvedSchema(schema).build(),
                        "test comment",
                        Collections.emptyList(),
                        options);

        TableDescriptor flussTable =
                FlinkConversions.toFlussTable(new ResolvedCatalogTable(flinkTable, schema));

        assertThat(flussTable.getProperties())
                .containsEntry(ConfigOptions.TABLE_LOG_FORMAT.key(), "indexed");

        ------ modify as following 
       HashMap<String, String> customProperties = new HashMap<>();
        customProperties.put(ConfigOptions.CLIENT_WRITER_BUFFER_MEMORY_SIZE.key(), "64mb");
        customProperties.put(ConfigOptions.CLIENT_WRITER_BATCH_TIMEOUT.key(), "32s");
        assertThat(flussTable.getCustomProperties()).containsExactlyEntriesOf(customProperties);
    }
```

The test will fail, 
```
Actual and expected should have same size but actual size is:
  3
while expected size is:
  2
Actual was:
  {"client.writer.batch-timeout"="32s", "client.writer.buffer.memory-size"="64mb", "table.log.format"="indexed"}
Expected was:
  ["client.writer.buffer.memory-size"="64mb", "client.writer.batch-timeout"="32s"]
```

"table.log.format"="indexed" is also in custom property.

@wuchong Is an issue needed to be resolved in v0.8

### Solution

_No response_

rionmonster:
### Purpose
Linked Issue: Closes #1745 

Per Issue https://github.com/apache/fluss/issues/1745, this pull request updates the handling of custom properties during Flink-based table creation, which previously would include _all_ properties as custom instead non-Fluss related properties.

### Brief change log

Updated the `FlinkConversions.toFlussTable(...)` function to identify any properties found within the configuration that were _not_ explicitly related to Fluss tables to properly distinguish between Fluss and non-Fluss (i.e. custom) properties.

### Tests

Added a new `FlinkConversionsTest.testTableConversionForCustomProperties` test case from the original issue that reproduced the issue and verified successful identification of both Fluss and custom properties (thanks to @luoyuxia for the original repro/test case).

### API and Format

N/A

### Documentation

N/A

### DIFF
--- a/fluss-flink/fluss-flink-common/src/test/java/org/apache/fluss/flink/catalog/FlinkCatalogITCase.java
+++ b/fluss-flink/fluss-flink-common/src/test/java/org/apache/fluss/flink/catalog/FlinkCatalogITCase.java
@@ -627,7 +627,7 @@ abstract class FlinkCatalogITCase {
             Map<String, String> expectedTableProperties = new HashMap<>();
             expectedTableProperties.put("table.datalake.format", "paimon");
             expectedTableProperties.put("table.replication.factor", "1");
-            assertThat(tableInfo.getTableConfig()).isEqualTo(expectedTableProperties);
+            assertThat(tableInfo.getProperties().toMap()).isEqualTo(expectedTableProperties);
 
             Map<String, String> expectedCustomProperties = new HashMap<>();
             expectedCustomProperties.put("k1", "v1");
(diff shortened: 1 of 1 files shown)


rionmonster:
**Alternative Consideration**

It's probably worth mentioning that I considered a few options prior to issuing this PR but the primary alternative felt a bit heavy-handed, but I'm open to changing things. The alternative consisted of introducing a private static class that would be responsible for performing a single pass over the configuration keys and identify each category during that iteration.

Something like the following along with a corresponding static function responsible for the extraction of those categories (e.g. `extractPropertyCategories(Configuration options)`) that would then expose each specific category:

```
private static class PropertyCategories {
    // allProperties probably not necessary here as the table only consumes table and custom properties
    final Map<String, String> allProperties;
    final Map<String, String> tableProperties;
    final Map<String, String> customProperties;
   
    ...
}
```

This would remove the key-wise comparison between all of the current properties with the table properties (which should be fairly inexpensive on its own).

PR #1749, opened 2025-09-24
[server] Recover log and index file for unclean shutdown

### LINKED ISSUE #1716: ReplicaFetcherThread keeps throwing UnknownServerException because of corrupt index file
### Search before asking

### Fluss version

main (development)

### Please describe the bug 🐞

When a `fetchLog` request needs to read from remote logs and happens to encounter a corrupt index file, the TabletServer is currently unable to handle this exception and the client will receive an `UnknownServerException`. In our production environment, the follower keeps receiving the following error, causing replica synchronization to stall.

```java
2025-09-17 19:36:00,476 WARN  org.apache.fluss.server.replica.fetcher.ReplicaFetcherThread [] - Error in response for fetch log request org.apache.fluss.rpc.messages.FetchLogRequest@160f9b0b
java.util.concurrent.ExecutionException: org.apache.fluss.exception.UnknownServerException: org.apache.fluss.server.exception.CorruptIndexException: Corrupt time index found, time index file (/mnt/fluss/data/0/remote-log-index-cache/60976417_41a99e37-f1a0-46ab-8745-b623ac4c2d37.timeindex) has non-zero size but the last timestamp is 0 which is less than the first timestamp 1758092497499
	at org.apache.fluss.server.log.TimeIndex.lambda$sanityCheck$0(TimeIndex.java:105)
	at org.apache.fluss.utils.concurrent.LockUtils.inLock(LockUtils.java:32)
	at org.apache.fluss.utils.concurrent.LockUtils.inReadLock(LockUtils.java:50)
	at org.apache.fluss.server.log.TimeIndex.sanityCheck(TimeIndex.java:94)
	at org.apache.fluss.server.log.remote.RemoteLogIndexCache.lambda$createCacheEntry$7(RemoteLogIndexCache.java:270)
	at org.apache.fluss.server.log.remote.RemoteLogIndexCache.loadIndexFile(RemoteLogIndexCache.java:496)
	at org.apache.fluss.server.log.remote.RemoteLogIndexCache.createCacheEntry(RemoteLogIndexCache.java:254)
	at org.apache.fluss.server.log.remote.RemoteLogIndexCache.lambda$getIndexEntry$2(RemoteLogIndexCache.java:221)
	at com.github.benmanes.caffeine.cache.BoundedLocalCache.lambda$doComputeIfAbsent$14(BoundedLocalCache.java:2406)
	at java.base/java.util.concurrent.ConcurrentHashMap.compute(ConcurrentHashMap.java:1916)
	at com.github.benmanes.caffeine.cache.BoundedLocalCache.doComputeIfAbsent(BoundedLocalCache.java:2404)
	at com.github.benmanes.caffeine.cache.BoundedLocalCache.computeIfAbsent(BoundedLocalCache.java:2387)
	at com.github.benmanes.caffeine.cache.LocalCache.computeIfAbsent(LocalCache.java:108)
	at com.github.benmanes.caffeine.cache.LocalManualCache.get(LocalManualCache.java:62)
	at org.apache.fluss.server.log.remote.RemoteLogIndexCache.lambda$getIndexEntry$3(RemoteLogIndexCache.java:219)
	at org.apache.fluss.utils.concurrent.LockUtils.inLock(LockUtils.java:42)
	at org.apache.fluss.utils.concurrent.LockUtils.inReadLock(LockUtils.java:55)
	at org.apache.fluss.server.log.remote.RemoteLogIndexCache.getIndexEntry(RemoteLogIndexCache.java:208)
	at org.apache.fluss.server.log.remote.RemoteLogIndexCach
	at java.util.concurrent.CompletableFuture.reportGet(CompletableFuture.java:396) ~[?:?]
	at java.util.concurrent.CompletableFuture.get(CompletableFuture.java:2096) ~[?:?]
	at org.apache.fluss.server.replica.fetcher.ReplicaFetcherThread.processFetchLogRequest(ReplicaFetcherThread.java:227) ~[fluss-server-0.8-SNAPSHOT.jar:0.8-SNAPSHOT]
	at java.util.Optional.ifPresent(Optional.java:178) [?:?]
	at org.apache.fluss.server.replica.fetcher.ReplicaFetcherThread.maybeFetch(ReplicaFetcherThread.java:153) [fluss-server-0.8-SNAPSHOT.jar:0.8-SNAPSHOT]
	at org.apache.fluss.server.replica.fetcher.ReplicaFetcherThread.doWork(ReplicaFetcherThread.java:124) [fluss-server-0.8-SNAPSHOT.jar:0.8-SNAPSHOT]
	at org.apache.fluss.utils.concurrent.ShutdownableThread.run(ShutdownableThread.java:96) [fluss-server-0.8-SNAPSHOT.jar:0.8-SNAPSHOT]
Caused by: org.apache.fluss.exception.UnknownServerException: org.apache.fluss.server.exception.CorruptIndexException: Corrupt time index found, time index file (/mnt/fluss/data/0/remote-log-index-cache/60976417_41a99e37-f1a0-46ab-8745-b623ac4c2d37.timeindex) has non-zero size but the last timestamp is 0 which is less than the first timestamp 1758092497499
	at org.apache.fluss.server.log.TimeIndex.lambda$sanityCheck$0(TimeIndex.java:105)
	at org.apache.fluss.utils.concurrent.LockUtils.inLock(LockUtils.java:32)
	at org.apache.fluss.utils.concurrent.LockUtils.inReadLock(LockUtils.java:50)
	at org.apache.fluss.server.log.TimeIndex.sanityCheck(TimeIndex.java:94)
	at org.apache.fluss.server.log.remote.RemoteLogIndexCache.lambda$createCacheEntry$7(RemoteLogIndexCache.java:270)
	at org.apache.fluss.server.log.remote.RemoteLogIndexCache.loadIndexFile(RemoteLogIndexCache.java:496)
	at org.apache.fluss.server.log.remote.RemoteLogIndexCache.createCacheEntry(RemoteLogIndexCache.java:254)
	at org.apache.fluss.server.log.remote.RemoteLogIndexCache.lambda$getIndexEntry$2(RemoteLogIndexCache.java:221)
	at com.github.benmanes.caffeine.cache.BoundedLocalCache.lambda$doComputeIfAbsent$14(BoundedLocalCache.java:2406)
	at java.base/java.util.concurrent.ConcurrentHashMap.compute(ConcurrentHashMap.java:1916)
	at com.github.benmanes.caffeine.cache.BoundedLocalCache.doComputeIfAbsent(BoundedLocalCache.java:2404)
	at com.github.benmanes.caffeine.cache.BoundedLocalCache.computeIfAbsent(BoundedLocalCache.java:2387)
	at com.github.benmanes.caffeine.cache.LocalCache.computeIfAbsent(LocalCache.java:108)
	at com.github.benmanes.caffeine.cache.LocalManualCache.get(LocalManualCache.java:62)
	at org.apache.fluss.server.log.remote.RemoteLogIndexCache.lambda$getIndexEntry$3(RemoteLogIndexCache.java:219)
	at org.apache.fluss.utils.concurrent.LockUtils.inLock(LockUtils.java:42)
	at org.apache.fluss.utils.concurrent.LockUtils.inReadLock(LockUtils.java:55)
	at org.apache.fluss.server.log.remote.RemoteLogIndexCache.getIndexEntry(RemoteLogIndexCache.java:208)
	at org.apache.fluss.server.log.remote.RemoteLogIndexCach
```

This issue will affect both clients and followers; further discussion is needed on how to fix client-side reading. This issue aims to first address the problem on the follower side.

### Solution

The reason for the `UnknownServerException` is that a `CorruptIndexException` was thrown while retrieving `firstStartPos` due to a corrupt index file.
```java
public class ReplicaManager {
    ...

    private @Nullable RemoteLogFetchInfo fetchLogFromRemote(Replica replica, long fetchOffset) {
        List<RemoteLogSegment> remoteLogSegmentList =
                remoteLogManager.relevantRemoteLogSegments(replica.getTableBucket(), fetchOffset);
        if (!remoteLogSegmentList.isEmpty()) {
            // will throw CorruptIndexException if index file corrupt
            int firstStartPos =
                    remoteLogManager.lookupPositionForOffset(
                            remoteLogSegmentList.get(0), fetchOffset);
                    ...
        } else {
            return null;
        }
    }
```

For the follower, firstStartPos is actually not needed, so we can simply skip this step for followers. This not only avoids the issue of the follower continuously receiving errors when encountering a corrupt index file, but also serves as a small performance optimization.

LiebingYu:
### Purpose

Linked issue: close #1716


LiebingYu:
Ready for CR @wuchong @swuferhong

fluss-server/src/main/java/org/apache/fluss/server/log/LogLoader.java
```diff
@@ -117,6 +122,37 @@ public LoadedLogOffsets load() throws IOException {
...
+     * @return The number of bytes truncated from the segment
+     * @throws LogSegmentOffsetOverflowException if the segment contains messages that cause index
+     *     offset overflow
+     */
+    private int recoverSegment(LogSegment segment) throws Exception {
+        WriterStateManager writerStateManager =
+                new WriterStateManager(
+                        logSegments.getTableBucket(),
+                        logTabletDir,
+                        this.writerStateManager.writerExpirationMs());
```
swuferhong, line 139: Why we need to new a  WriterStateManager? Maybe we can add a clear() method?

fluss-server/src/test/java/org/apache/fluss/server/log/LogLoaderTest.java
```diff
@@ -0,0 +1,206 @@
...
+                    FileChannel.open(indexFile.toPath(), StandardOpenOption.APPEND)) {
+                fileChannel.write(ByteBuffer.wrap(new byte[] {0}));
+            }
+        }
+
+        // test reopen the log without recovery, sanity check of index files should throw exception
+        logTablet = createLogTablet(true);
+        for (LogSegment segment : logTablet.logSegments()) {
+            if (segment.getBaseOffset() != logTablet.activeLogSegment().getBaseOffset()) {
+                assertThatThrownBy(segment.offsetIndex()::sanityCheck)
```
swuferhong, line 110: add exception msg.

fluss-server/src/test/java/org/apache/fluss/server/log/LogLoaderTest.java
```diff
@@ -0,0 +1,206 @@
...
+                        .isInstanceOf(CorruptIndexException.class);
+                assertThatThrownBy(segment.timeIndex()::sanityCheck)
```
swuferhong, line 118: ditto

fluss-server/src/test/java/org/apache/fluss/server/log/LogLoaderTest.java
```diff
@@ -0,0 +1,206 @@
...
+                            changeTypes,
+                            objects,
+                            LogFormat.ARROW,
+                            ArrowCompressionInfo.DEFAULT_COMPRESSION);
+            logTablet.appendAsFollower(memoryLogRecords);
+            baseOffset++;
+            batchSequence++;
+        }
+    }
+
```
swuferhong, line 316: Can you add a test to verify the `recoverySegment()` only recovery the `logSegments` generated after `recoveryPointCheckpoint`.  It appears that the logic related to `recoveryPointCheckpoint` has not been covered.

fluss-server/src/main/java/org/apache/fluss/server/log/LogLoader.java
```diff
@@ -129,16 +165,106 @@ public LoadedLogOffsets load() throws IOException {
...
+            while (unflushedIter.hasNext() && !truncated) {
+                LogSegment segment = unflushedIter.next();
+                LOG.info(
+                        "Recovering unflushed segment {}. {}/{} recovered for bucket {}",
+                        segment.getBaseOffset(),
+                        numFlushed,
+                        numUnflushed,
+                        logSegments.getTableBucket());
+
+                int truncatedBytes = -1;
```
swuferhong, line 186: Do we need to be this too aggressive here? Deleting all `subsequent logSegments` just because one cannot be repaired — I feel this might pose a risk of data loss. Also, we don't have test coverage for this logic.

fluss-server/src/main/java/org/apache/fluss/server/log/LogLoader.java
LiebingYu: This logic follow Kafka, and ceate a new WriterStateManager is a lightweight operation. I think it's ok.

fluss-server/src/main/java/org/apache/fluss/server/log/LogLoader.java
LiebingYu: This logic also follow Kafka. From my perspective, data loss is unlikely because the data is stored in multiple replicas. Once the file is truncated to the correct position, it can synchronize the latest data from the leader. If truncation is not carried out, the file appears to be unrecoverable, and if the host machine becomes the leader afterward, unforeseen problems might occur.

LiebingYu: And I add test to cover this.

fluss-server/src/main/java/org/apache/fluss/server/log/LogLoader.java
```diff
@@ -117,6 +122,37 @@ public LoadedLogOffsets load() throws IOException {
...
+        // that the current implementation of logStartOffset in Fluss is not yet fully refined, and
+        // there may be cases where logStartOffset is not updated. As a result, logStartOffset is
+        // not yet reliable. Once the issue with correctly updating logStartOffset is resolved in
+        // issue https://github.com/apache/fluss/issues/744, we can use logStartOffset here.
+        // Additionally, using 0 versus using logStartOffset does not affect correctness—they both
+        // can restore the complete WriterState. The only difference is that using logStartOffset
+        // can potentially skip over more segments.
+        LogTablet.rebuildWriterState(
+                writerStateManager, logSegments, 0, segment.getBaseOffset(), false);
+        int bytesTruncated = segment.recover();
```
wuchong, line 150: Do we need to pass `writerStateManager` into `segment.recover()` to update writer state like how kafka does? cc @swuferhong

fluss-server/src/main/java/org/apache/fluss/server/log/LogLoader.java
```diff
@@ -129,16 +165,106 @@ public LoadedLogOffsets load() throws IOException {
...
+                } catch (Exception e) {
+                    if (e instanceof InvalidOffsetException) {
+                        long startOffset = segment.getBaseOffset();
+                        LOG.warn(
+                                "Found invalid offset during recovery for bucket {}. Deleting the corrupt segment "
+                                        + "and creating an empty one with starting offset {}",
+                                logSegments.getTableBucket(),
+                                startOffset);
+                        truncatedBytes = segment.truncateTo(startOffset);
+                    }
```
wuchong, line 200: We should rethrow the exception in the else branch. Otherwise, the exception is swallowed.

fluss-server/src/main/java/org/apache/fluss/server/log/LogLoader.java
```diff
@@ -155,8 +281,26 @@ private void loadSegmentFiles() throws IOException {
...
+                                if (isCleanShutdown
+                                        || segment.getBaseOffset() < recoveryPointCheckpoint) {
+                                    LOG.error(
+                                            "Could not find offset index file corresponding to log file {} "
+                                                    + "for bucket {}, recovering segment and rebuilding index files...",
+                                            logSegments.getTableBucket(),
+                                            segment.getFileLogRecords().file().getAbsoluteFile());
+                                }
+                                recoverSegment(segment);
+                            }
```
wuchong, line 307: Should we catch `CorruptIndexException` and recover segment for this case like how kafka does? And we should rethrow other exceptions in else branch.

fluss-server/src/test/java/org/apache/fluss/server/log/LogLoaderTest.java
```diff
@@ -0,0 +1,325 @@
...
+        // test reopen the log with recovery, sanity check of index files should no corruption
+        logTablet = createLogTablet(false);
+        for (LogSegment segment : logTablet.logSegments()) {
+            segment.offsetIndex().sanityCheck();
+            segment.timeIndex().sanityCheck();
+        }
+        assertThat(numRecords).isEqualTo(logTablet.localLogEndOffset());
+        for (int i = 0; i < numRecords; i++) {
+            assertThat(i)
+                    .isEqualTo(logTablet.lookupOffsetForTimestamp(clock.milliseconds() + i * 10));
```
wuchong:
```
            assertThat(logTablet.lookupOffsetForTimestamp(clock.milliseconds() + i * 10))
                    .isEqualTo(i);
```

The actual value should in `assertThat(..)`

fluss-server/src/test/java/org/apache/fluss/server/log/LogLoaderTest.java
```diff
@@ -0,0 +1,325 @@
...
+
+        // reopen the log
+        logTablet = createLogTablet(false);
+        assertThat(logTablet.localLogEndOffset()).isEqualTo(numRecords);
+        // the index files should be rebuilt
+        assertThat(logTablet.logSegments().get(0).offsetIndex().entries() > 0).isTrue();
+        assertThat(logTablet.logSegments().get(0).timeIndex().entries() > 0).isTrue();
+        for (int i = 0; i < numRecords; i++) {
+            assertThat(i)
+                    .isEqualTo(logTablet.lookupOffsetForTimestamp(clock.milliseconds() + i * 10));
```
wuchong: ditto

fluss-server/src/main/java/org/apache/fluss/server/log/LogSegment.java
```diff
@@ -172,6 +172,23 @@ public void resizeIndexes(int size) throws IOException {
...
+        if (lazyOffsetIndex.file().exists()) {
+            // Resize the time index file to 0 if it is newly created.
+            if (timeIndexFileNewlyCreated) {
+                timeIndex().resize(0);
+            }
+            // Sanity checks for time index and offset index are skipped because
+            // we will recover the segments above the recovery point in recoverLog()
+            // in any case so sanity checking them here is redundant.
+        } else {
+            throw new NoSuchFieldException(
```
wuchong: should throw `NoSuchFileException`

fluss-server/src/main/java/org/apache/fluss/server/log/LogLoader.java
```diff
@@ -129,16 +165,106 @@ public LoadedLogOffsets load() throws IOException {
...
+                        toDelete,
+                        logSegments.getTableBucket(),
+                        e);
+            }
+        }
+    }
+
     /** Loads segments from disk into the provided segments. */
-    private void loadSegmentFiles() throws IOException {
+    private void loadSegmentFiles() throws Exception {
```
wuchong: We should keep throws `IOException` instead of generic Exception because this is an IO operation.

wuchong:
I added a commit to fix the `IOException` signature problem.

fluss-server/src/main/java/org/apache/fluss/server/log/LogLoader.java
LiebingYu: 1. Kafka throw `CorruptIndexException` because Kafka will do sanity check for transaction file which may throw `CorruptIndexException`. But here in fluss we don't do sanity check of index files in `segment.sanityCheck`. So I think we don't need to catch `CorruptIndexException` here.

fluss-server/src/main/java/org/apache/fluss/server/log/LogLoader.java
LiebingYu: Kafka use `writerStateManager` in `segment.recover()` mainly for txn operations. I think Fluss doen't need here. What's your opinion @swuferhong?

swuferhong: After discussing offline with @LiebingYu , we confirmed that it‘s necessary to rebuild the `writerState1. Although it is not used during recovery, this code path ensures that when `writerStateManager.takeSnapshot()` is called, a correct snapshot up to the current segment is written to disk. This way, during recovery, each segment can be restored based on a fully accurate writerState snapshot, without requiring a full global reconstruction.

PR #1750, opened 2025-09-24
[hotfix] Add bucket info for bucket related logs

LiebingYu:
### Purpose

Currently, a large number of Fluss logs do not include specific bucket information, which makes troubleshooting in production clusters difficult. This PR aims to uniformly add bucket information to these logs to avoid having to add it individually after issues arise.


LiebingYu:
CC @swuferhong

PR #1752, opened 2025-09-24
[client] Throw Exception when retry several times when downloading log.

### LINKED ISSUE #1751: RemoteLogDownloader deadlock when met exception.
### Search before asking

### Fluss version

0.7.0 (latest release)

### Please describe the bug 🐞

In the original code, the download lock (prefetchSemaphore) is released only in two cases:

1. After a RemoteLogSegment has been successfully read (drained), the lock is released via recycleRemoteLog.

Image

2. When the download of a file fails, the lock is released.

Image

Let us simplify the model: suppose a bucket contains three segment files — A, B, and C — and client.scanner.remote-log.prefetch-num = 1.
1. File A fails to download, so the lock is released. File A is then added back to the end of the queue.
2. File B downloads successfully, but the lock is not immediately released because it hasn't been drained.
Since file A has an earlier offset, it remains at the front of the queue and must be processed first. However, file B holds the prefetch lock, and file A cannot be reattempted until the lock is acquired again. But because B will never be drained (as A blocks its processing), the lock is never released — resulting in a deadlock.
3.  file C will not be downloaded, and file A will never be retried. The entire job becomes stuck.

### Solution

When RemoteLogDownloader failed to download a file, no longer release the semaphore but retied to download for several times. If still failed to download, but thrown the exception out of client, let the flink job fails and restarts.

loserwang1024:
### Purpose

Linked issue: close #1751

### DIFF
--- a/fluss-client/src/main/java/org/apache/fluss/client/table/scanner/log/RemoteLogDownloader.java
+++ b/fluss-client/src/main/java/org/apache/fluss/client/table/scanner/log/RemoteLogDownloader.java
@@ -63,6 +63,7 @@ public class RemoteLogDownloader implements Closeable {
     private static final Logger LOG = LoggerFactory.getLogger(RemoteLogDownloader.class);
 
     private static final long POLL_TIMEOUT = 5000L;
+    private static final int MAX_RETRY_COUNT = 5;
 
     private final Path localLogDir;
 
@@ -159,7 +160,11 @@ public class RemoteLogDownloader implements Closeable {
             return;
         }
 
-        TableBucket tableBucket = request.getTableBucket();
+        downloadRemoteLog(request, MAX_RETRY_COUNT, System.currentTimeMillis());
+    }
+
+    private void downloadRemoteLog(
+            RemoteLogDownloadRequest request, int retryCount, long startTime) {
         try {
             // 1. cleanup the finished logs first to free up disk space
             cleanupRemoteLogs();
@@ -168,7 +173,6 @@ public class RemoteLogDownloader implements Closeable {
             FsPathAndFileName fsPathAndFileName = request.getFsPathAndFileName();
             scannerMetricGroup.remoteFetchRequestCount().inc();
 
-            long startTime = System.currentTimeMillis();
             // download the remote file to local
             CompletableFuture<Long> completableFuture =
                     remoteFileDownloader.downloadFileAsync(fsPathAndFileName, localLogDir);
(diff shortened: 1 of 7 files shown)


loserwang1024:
@wuchong , CC

loserwang1024:
@wuchong , fix this test

loserwang1024:
@wuchong , CC, it's important for me.

loserwang1024:
@swuferhong , I have rebased this PR, CC

loserwang1024:
@swuferhong , I have rebased this PR again

fluss-client/src/main/java/org/apache/fluss/client/table/scanner/log/RemoteLogDownloader.java
```diff
@@ -236,16 +233,49 @@ void fetchOnce() throws Exception {
...
+            return;
+        }
+
+        LOG.error(
+                "Failed to download remote log segment file {} for table bucket {}.",
+                request.getFsPathAndFileName().getFileName(),
+                request.getTableBucket(),
+                ExceptionUtils.stripExecutionException(throwable));
+        scannerMetricGroup.remoteFetchErrorCount().inc();
+        if (retryCount >= 1) {
```
swuferhong, line 291: The `5` retries in `handleFetchException` are fired back-to-back with no delay/backoff between attempts. For a permanent error (e.g. the remote file does not exist), this just burns 6 immediate attempts before failing — acceptable, 

since it fails fast. But for transient errors (network blips, object-store throttling), 6 retries within a few milliseconds are likely to all fail, so we give up and trigger a failover before the transient condition has a chance to recover, which weakens the self-healing that retrying is supposed to provide. Could we add a (possibly exponential) backoff between retries?

fluss-client/src/main/java/org/apache/fluss/client/table/scanner/log/RemoteLogDownloader.java
```diff
@@ -62,6 +62,7 @@ public class RemoteLogDownloader implements Closeable {
     private static final Logger LOG = LoggerFactory.getLogger(RemoteLogDownloader.class);
 
     private static final long POLL_TIMEOUT = 5000L;
+    private static final int MAX_RETRY_COUNT = 5;
```
swuferhong: would be better to expose this as a ConfigOption?

swuferhong (commented):
Thanks, @loserwang1024, I have left some comments:

fluss-client/src/main/java/org/apache/fluss/client/table/scanner/log/RemoteLogDownloader.java
loserwang1024: done it.

swuferhong (approved):
LGTM +1

PR #1753, opened 2025-09-24
[server] feat: support alter table.datalake.enabled

### LINKED ISSUE #431: Automatically create corresponding Lake tables when 'table.datalake.enabled' enabled

gyang94:
### Purpose

Linked issue: close #431


fluss-client/src/test/java/org/apache/fluss/client/admin/FlussAdminITCase.java
```diff
@@ -261,7 +260,7 @@ void testAlterTable() throws Exception {
                                                 false)
                                         .get())
                 .cause()
-                .isInstanceOf(DatabaseNotExistException.class);
+                .isInstanceOf(TableNotExistException.class);
```
luoyuxia, line 267: Why change to `TableNotExistException`?  Shouldn't it be `DatabaseNotExistException`?

fluss-server/src/main/java/org/apache/fluss/server/coordinator/CoordinatorService.java
```diff
@@ -366,8 +373,36 @@ private void handleFlussTableChanges(
             }
         }
 
+        TableInfo existTableInfo = metadataManager.getTable(tablePath);
+        TableDescriptor existTableDescriptor = existTableInfo.toTableDescriptor();
+
         metadataManager.alterTableProperties(
```
luoyuxia, line 304: I'd suggest `alterTableProperties` return old tabledescriptor and new table descriptor since we already will know  old and new descriptor in this method. So that, we can avoid  calling
```
TableInfo existTableInfo = metadataManager.getTable(tablePath);
        TableDescriptor existTableDescriptor = existTableInfo.toTableDescriptor();
```

fluss-server/src/main/java/org/apache/fluss/server/coordinator/CoordinatorService.java
```diff
@@ -366,8 +373,36 @@ private void handleFlussTableChanges(
...
+        TableInfo alteredTableInfo = metadataManager.getTable(tablePath);
+        TableDescriptor alteredTableDescriptor = alteredTableInfo.toTableDescriptor();
+        // enable lake table
+        if (!isDataLakeEnabled(existTableDescriptor) && isDataLakeEnabled(alteredTableDescriptor)) {
+            // TODO: should tolerate if the lake exist but matches our schema. This ensures
+            // eventually
+            //  consistent by idempotently creating the table multiple times. See #846
+            // before create table in fluss, we may create in lake
+            try {
+                checkNotNull(lakeCatalog).createTable(tablePath, alteredTableDescriptor);
```
luoyuxia: actaully, I'd suggest first create the lake table and then alter table properites. 
The reason is that `create the lake table` is more like to fail, and if we alreay alter table properties, and then create lake table fail. It will then case although the table is datalake enabeld, but the lake table is not created.

fluss-client/src/test/java/org/apache/fluss/client/admin/FlussAdminITCase.java
gyang94: because now we have a `TableInfo existTableInfo = metadataManager.getTable(tablePath);` before alterTableConfigs. It only throw TableNotExistException
`

fluss-lake/fluss-lake-paimon/src/test/java/org/apache/fluss/lake/paimon/LakeEnabledTableCreateITCase.java
```diff
@@ -364,6 +366,218 @@ void testCreateLakeEnabledTableWithAllTypes() throws Exception {
...
+                        .build();
+        TablePath logTablePath = TablePath.of(DATABASE, "log_table_alter");
+        admin.createTable(logTablePath, logTable, false).get();
+
+        assertThatThrownBy(
+                        () ->
+                                paimonCatalog.getTable(
+                                        Identifier.create(DATABASE, logTablePath.getTableName())))
+                .isInstanceOf(Catalog.TableNotExistException.class)
+                .hasMessageContaining(
```
luoyuxia: nit: we don't need to check the error message of paimon

fluss-lake/fluss-lake-paimon/src/test/java/org/apache/fluss/lake/paimon/LakeEnabledTableCreateITCase.java
```diff
@@ -364,6 +366,218 @@ void testCreateLakeEnabledTableWithAllTypes() throws Exception {
...
+                        String.format(
+                                "Table %s.%s does not exist.",
+                                DATABASE, logTablePath.getTableName()));
+
+        // enable lake
+        TableChange.SetOption enableLake =
+                TableChange.set(ConfigOptions.TABLE_DATALAKE_ENABLED.key(), "true");
+        List<TableChange> changes = Arrays.asList(enableLake);
```
luoyuxia: nit:
```suggestion
        List<TableChange> changes = Collections.singletonList(enableLake);
```

fluss-lake/fluss-lake-paimon/src/test/java/org/apache/fluss/lake/paimon/LakeEnabledTableCreateITCase.java
```diff
@@ -364,6 +366,218 @@ void testCreateLakeEnabledTableWithAllTypes() throws Exception {
...
+                            BUCKET_COLUMN_NAME,
+                            OFFSET_COLUMN_NAME,
+                            TIMESTAMP_COLUMN_NAME
+                        }),
+                "log_c1,log_c2",
+                BUCKET_NUM);
+    }
+
+    @Test
+    void testAlterPkLakeEnabledTable() throws Exception {
```
luoyuxia: I think just verify for log table is enough since alter table to be lake enabeld shouldn't related to table types. 
So I think we can remove  `testAlterPkLakeEnabledTable`, `testAlterPartitionedLakeEnabledTable` to make testing simple

fluss-lake/fluss-lake-paimon/src/test/java/org/apache/fluss/lake/paimon/tiering/PaimonTieringITCase.java
```diff
@@ -170,6 +171,147 @@ void testTiering() throws Exception {
...
+            jobClient.cancel().get();
+        }
+    }
+
+    @Test
+    void testTieringForAlterTable() throws Exception {
+        TablePath t1 = TablePath.of(DEFAULT_DB, "pkTableAlter");
+        Map<String, String> tableProperties = new HashMap<>();
+        tableProperties.put(ConfigOptions.TABLE_DATALAKE_ENABLED.key(), "false");
+        tableProperties.put(ConfigOptions.TABLE_DATALAKE_FRESHNESS.key(), "500 ms");
```
luoyuxia: we can remove `tableProperties.put(ConfigOptions.TABLE_DATALAKE_FRESHNESS.key(), "500 ms");` since `createPkTable` method will add 
```
property(ConfigOptions.TABLE_DATALAKE_FRESHNESS, Duration.ofMillis(500))
```

fluss-lake/fluss-lake-paimon/src/test/java/org/apache/fluss/lake/paimon/tiering/PaimonTieringITCase.java
```diff
@@ -170,6 +171,147 @@ void testTiering() throws Exception {
                         0);
             }
 
+            String raw =
```
luoyuxia: Could you please also change https://github.com/apache/fluss/pull/1753/files#diff-1a54a1b286bd0515fc522305c66771939634b26c57711137461aba7ceaaa55bdL179 to be like this? 
Maybe you can extract a common method to make all test case can reuse this logic.

fluss-server/src/main/java/org/apache/fluss/server/coordinator/CoordinatorService.java
```diff
@@ -297,14 +302,92 @@ public CompletableFuture<AlterTablePropertiesResponse> alterTableProperties(
...
+                metadataManager.alterTableProperties(
+                        tablePath, tablePropertyChanges, request.isIgnoreIfNotExists());
+                postAlterTableProperties(tablePath, tableDescriptor);
+            }
+        } catch (Exception e) {
+            if (e instanceof TableNotExistException) {
+                if (!request.isIgnoreIfNotExists()) {
+                    throw (TableNotExistException) e;
+                }
+            } else if (e instanceof RuntimeException) {
```
luoyuxia: got warn from my IDE
```
Condition 'e instanceof RuntimeException' is always 'true' 
```

fluss-server/src/main/java/org/apache/fluss/server/coordinator/MetadataManager.java
```diff
@@ -309,22 +310,11 @@ public void alterTableProperties(
...
-
-            TableDescriptor tableDescriptor = tableInfo.toTableDescriptor();
-            TableDescriptor newDescriptor =
-                    getUpdatedTableDescriptor(tableDescriptor, tablePropertyChanges);
+            Tuple2<TableDescriptor, TableDescriptor> tuple =
+                    validateAndGetUpdatedTableDescriptor(tablePath, tablePropertyChanges);
+            TableDescriptor tableDescriptor = tuple.f0;
+            TableDescriptor newDescriptor = tuple.f1;
 
             if (newDescriptor != null) {
```
luoyuxia, line 336: From the following code:
```
// reuse the same validate logic with the createTable() method
                validateTableDescriptor(tableDescriptor, maxBucketNum);
```
Seems it should be validate with the new `tableDescriptor`
```
validateTableDescriptor(newDescriptor, maxBucketNum);
```
?
cc @wuchong

fluss-server/src/main/java/org/apache/fluss/server/coordinator/CoordinatorService.java
```diff
@@ -297,14 +302,92 @@ public CompletableFuture<AlterTablePropertiesResponse> alterTableProperties(
...
-        metadataManager.alterTableProperties(
-                tablePath,
-                toTablePropertyChanges(request.getConfigChangesList()),
-                request.isIgnoreIfNotExists());
+        TablePropertyChanges tablePropertyChanges =
+                toTablePropertyChanges(request.getConfigChangesList());
 
+        try {
+            Tuple2<TableDescriptor, TableDescriptor> tuple =
+                    metadataManager.validateAndGetUpdatedTableDescriptor(
```
luoyuxia: introduce a new method `validateAndGetUpdatedTableDescriptor` make it harder to understand. Could we just add the preAlterTableProperties logic to  
https://github.com/apache/fluss/pull/1753/files#diff-4b4f64d28bd52ed8858e9fe2adc5e50c2f9daf2c8d7c012a070d290ccf185181L329
?

fluss-server/src/main/java/org/apache/fluss/server/coordinator/CoordinatorService.java
```diff
@@ -297,14 +302,92 @@ public CompletableFuture<AlterTablePropertiesResponse> alterTableProperties(
...
+    private void postAlterTableProperties(TablePath tablePath, TableDescriptor tableDescriptor) {
+        mayAddRemoveTiering(tablePath, tableDescriptor);
+        // more post-alter actions can be added here
+    }
+
+    private void mayEnableDataLake(
+            TablePath tablePath, TableDescriptor tableDescriptor, TableDescriptor newDescriptor) {
+        boolean toEnableDataLake =
+                !isDataLakeEnabled(tableDescriptor) && isDataLakeEnabled(newDescriptor);
+        boolean toDisableDataLake =
```
luoyuxia: Got warning from my ide 
```
Variable 'toDisableDataLake' is never used 
```

fluss-server/src/main/java/org/apache/fluss/server/coordinator/CoordinatorService.java
```diff
@@ -297,14 +302,92 @@ public CompletableFuture<AlterTablePropertiesResponse> alterTableProperties(
...
+        }
         return CompletableFuture.completedFuture(new AlterTablePropertiesResponse());
     }
 
+    private void preAlterTableProperties(
+            TablePath tablePath,
+            TableDescriptor existingTableDescriptor,
+            TableDescriptor updatedDescriptor) {
+
+        mayEnableDataLake(tablePath, existingTableDescriptor, updatedDescriptor);
```
luoyuxia: can we just extract the code of `mayEnableDataLake` to here, removing the method `mayEnableDataLake`. I think the logic of  `mayEnableDataLake` is not complex, put all code into here make it easy to track

fluss-server/src/main/java/org/apache/fluss/server/coordinator/CoordinatorService.java
```diff
@@ -297,14 +302,92 @@ public CompletableFuture<AlterTablePropertiesResponse> alterTableProperties(
...
+        // more pre-alter actions can be added here
+    }
+
+    private void postAlterTableProperties(TablePath tablePath, TableDescriptor tableDescriptor) {
+        mayAddRemoveTiering(tablePath, tableDescriptor);
```
luoyuxia: dito

fluss-server/src/main/java/org/apache/fluss/server/coordinator/CoordinatorService.java
```diff
@@ -297,14 +302,92 @@ public CompletableFuture<AlterTablePropertiesResponse> alterTableProperties(
...
+                        String.format(
+                                "The table %s already exists in %s catalog, please "
+                                        + "first drop the table in %s catalog or use a new table name.",
+                                tablePath, dataLakeFormat, dataLakeFormat));
+            }
+        }
+    }
+
+    private void mayAddRemoveTiering(TablePath tablePath, TableDescriptor tableDescriptor) {
+        TableInfo alteredTableInfo = metadataManager.getTable(tablePath);
```
luoyuxia: 1: we don't  always, need to call `metadataManager.getTable` to get table info. The current logic will cause always get table info even for just custom properties change

2: Also, we can just use `tableReg.toTableInfo(tablePath, schemaInfo)` to get table info without calling method `metadataManager.getTable(`.

luoyuxia (commented):
@gyang94 Thanks for the pr. Left some comments.

fluss-server/src/main/java/org/apache/fluss/server/coordinator/MetadataManager.java
gyang94: I think it should be `newDescriptor` too. Updated.

fluss-server/src/main/java/org/apache/fluss/server/coordinator/CoordinatorService.java
gyang94: One uncomfortable thing is that the `preAlterTableProperties`(which will to call lakeCatalog to create a table) can only be running in `CoordinatorService`. Thus we need to `getUpdatedTableDescriptor()` in CoordinatorService before calling `MetadataManager.alterTableProperties()`.
Thus there will be two same pieces of code in both `CoordinatorService.alterTableProperties()` and `MetadataManager.alterTableProperties`, that is `validateAndGetUpdatedTableDescriptor()`. I removed this function now and keep codes there.

fluss-server/src/main/java/org/apache/fluss/server/coordinator/MetadataManager.java
```diff
@@ -352,14 +372,73 @@ public void alterTableProperties(
...
+
     /**
      * Get a new TableDescriptor with updated properties.
      *
      * @param tableDescriptor the current table descriptor.
      * @param tablePropertyChanges the changes for the table properties
      * @return the updated TableDescriptor, or null if no properties updated.
      */
-    private @Nullable TableDescriptor getUpdatedTableDescriptor(
+    public @Nullable TableDescriptor getUpdatedTableDescriptor(
```
luoyuxia: nit: change to private

fluss-server/src/main/java/org/apache/fluss/server/coordinator/MetadataManager.java
```diff
@@ -328,12 +335,25 @@ public void alterTableProperties(
 
             if (newDescriptor != null) {
                 // reuse the same validate logic with the createTable() method
-                validateTableDescriptor(tableDescriptor, maxBucketNum);
+                validateTableDescriptor(newDescriptor, maxBucketNum);
+
+                // pre alter table properties, e.g. create lake table in lake storage
```
luoyuxia:
```suggestion
                // pre alter table properties, e.g. create lake table in lake storage if it's to enable datalake for the table
```

fluss-server/src/main/java/org/apache/fluss/server/coordinator/MetadataManager.java
```diff
@@ -352,14 +372,73 @@ public void alterTableProperties(
...
+                !isDataLakeEnabled(tableDescriptor) && isDataLakeEnabled(newDescriptor);
+
+        // enable lake table
+        if (toEnableDataLake) {
+            // TODO: should tolerate if the lake exist but matches our schema. This ensures
+            // eventually
+            //  consistent by idempotently creating the table multiple times. See #846
+            // before create table in fluss, we may create in lake
+            try {
+                checkNotNull(lakeCatalog).createTable(tablePath, newDescriptor);
```
luoyuxia: if the lakeCatalog is null, should throw excepton 
```
if (lakeCatalog == null) {
                throw new InvalidAlterTableException(
                        "Cannot alter table "
                                + tablePath
                                + " to enable data lake, because the Fluss cluster doesn't enable datalake tables.");
            } else {
                try {
                    lakeCatalog.createTable(tablePath, newDescriptor);
                } catch (TableAlreadyExistException e) {
                    throw new LakeTableAlreadyExistException(
                            String.format(
                                    "The table %s already exists in %s catalog, please "
                                            + "first drop the table in %s catalog or use a new table name.",
                                    tablePath, dataLakeFormat, dataLakeFormat));
                }
            }
```

fluss-server/src/main/java/org/apache/fluss/server/coordinator/MetadataManager.java
```diff
@@ -328,12 +335,25 @@ public void alterTableProperties(
...
+                // pre alter table properties, e.g. create lake table in lake storage
+                preAlterTableProperties(
+                        tablePath, tableDescriptor, newDescriptor, lakeCatalog, dataLakeFormat);
                 // update the table to zk
                 TableRegistration updatedTableRegistration =
                         tableReg.newProperties(
                                 newDescriptor.getProperties(), newDescriptor.getCustomProperties());
                 zookeeperClient.updateTable(tablePath, updatedTableRegistration);
+
+                // post alter table properties, e.g. update lake table tiering manager
```
luoyuxia:
```suggestion
                // post alter table properties, e.g. add the table to lake table tiering manager if it's to enable datalake for the table
```

fluss-server/src/main/java/org/apache/fluss/server/coordinator/MetadataManager.java
```diff
@@ -352,14 +372,73 @@ public void alterTableProperties(
...
+                        String.format(
+                                "The table %s already exists in %s catalog, please "
+                                        + "first drop the table in %s catalog or use a new table name.",
+                                tablePath, dataLakeFormat, dataLakeFormat));
+            }
+        }
+        // more pre-alter actions can be added here
+    }
+
+    private void postAlterTableProperties(
```
luoyuxia, line 412: Feel a litlle for complex:
can be simplified to 
```
private void postAlterTableProperties(
            TablePath tablePath,
            SchemaInfo schemaInfo,
            TableDescriptor oldTableDescriptor,
            TableRegistration newTableRegistration,
            LakeTableTieringManager lakeTableTieringManager) {
        boolean toEnableDataLake =
                !isDataLakeEnabled(oldTableDescriptor)
                        && isDataLakeEnabled(newTableRegistration.properties);
        boolean toDisableDataLake =
                isDataLakeEnabled(oldTableDescriptor)
                        && !isDataLakeEnabled(newTableRegistration.properties);

        if (toEnableDataLake) {
            TableInfo newTableInfo = newTableRegistration.toTableInfo(tablePath, schemaInfo);
            // if the table is lake table, we need to add it to lake table tiering manager
            lakeTableTieringManager.addNewLakeTable(newTableInfo);
        } else if (toDisableDataLake) {
            lakeTableTieringManager.removeLakeTable(newTableRegistration.tableId);
        }
        // more post-alter actions can be added here
    }
```

```
private boolean isDataLakeEnabled(Map<String, String> properties) {
        String dataLakeEnabledValue = properties.get(ConfigOptions.TABLE_DATALAKE_ENABLED.key());
        return Boolean.parseBoolean(dataLakeEnabledValue);
    }
```

luoyuxia (approved):
LGTM! @wuchong Could you please have a final review?

PR #1754, opened 2025-09-25
[lake] Paimon lake table support alter table properties

### LINKED ISSUE #1682: Paimon support updating table properties
### Search before asking

### Description

`PaimonLakeCatalog` should implement `alterTable` to support updating table properties.

### Willingness to contribute

LiebingYu:
### Purpose

Linked issue: close #1682


fluss-server/src/main/java/org/apache/fluss/server/coordinator/CoordinatorService.java
```diff
@@ -368,6 +370,20 @@ private void handleFlussTableChanges(
 
         metadataManager.alterTableProperties(
                 tablePath, propertyChangesBuilder.build(), ignoreIfNotExists);
+
+        TableRegistration tableRegistration = metadataManager.getTableRegistration(tablePath);
```
luoyuxia: if `ignoreIfNotExists` is set to true, but the table not exist will still be thrown in here.

fluss-server/src/test/java/org/apache/fluss/server/lakehouse/TestingPaimonStoragePlugin.java
```diff
@@ -79,6 +82,10 @@ public void createTable(TablePath tablePath, TableDescriptor tableDescriptor)
             tableByPath.put(tablePath, tableDescriptor);
         }
 
+        @Override
+        public void alterTable(TablePath tablePath, List<TableChange> tableChanges)
+                throws TableNotExistException {}
```
luoyuxia: nit
```suggestion
                throws TableNotExistException {
                // do nothing
               }
```

fluss-server/src/main/java/org/apache/fluss/server/coordinator/CoordinatorService.java
```diff
@@ -368,6 +370,20 @@ private void handleFlussTableChanges(
...
+        // Update the lake table accordingly if lake table enabled
+        if (tableRegistration.getTableConfig().isDataLakeEnabled()) {
+            try {
+                checkNotNull(lakeCatalog).alterTable(tablePath, tableChanges);
+            } catch (TableNotExistException e) {
+                throw new LakeTableNotExistException(
```
luoyuxia: I'm hesitate to introduce this exception since the exception should never happen, right? 
If a table is datalake enabled, the lake table must exist excep for the some unexpected exception happen such as users drop the lake table munutally. 
And the exception message `first create the table in %s catalog` is not recommended to do for users since we don't recommend users to create the lake table by themself.

fluss-server/src/main/java/org/apache/fluss/server/coordinator/CoordinatorService.java
```diff
@@ -368,6 +370,20 @@ private void handleFlussTableChanges(
 
         metadataManager.alterTableProperties(
                 tablePath, propertyChangesBuilder.build(), ignoreIfNotExists);
+
+        TableRegistration tableRegistration = metadataManager.getTableRegistration(tablePath);
+        // Update the lake table accordingly if lake table enabled
+        if (tableRegistration.getTableConfig().isDataLakeEnabled()) {
+            try {
+                checkNotNull(lakeCatalog).alterTable(tablePath, tableChanges);
```
luoyuxia: I think we should first alter the table in lake and then alter the table in fluss to follow the logic of create lake enabled table. 
It will fail that alter lake table, but the table in fluss has already altered. 
If alter successfull, we expect all the preconditions change are met including alter the lake table.

fluss-lake/fluss-lake-paimon/src/test/java/org/apache/fluss/lake/paimon/PaimonLakeCatalogTest.java
```diff
@@ -0,0 +1,111 @@
...
+import org.junit.jupiter.api.io.TempDir;
+
+import java.io.File;
+import java.util.Arrays;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
+
+/** Unit test for {@link PaimonLakeCatalog}. */
+class PaimonLakeCatalogTest {
```
luoyuxia, line 39: Can you also add IT for it? `LakeEnabledTableCreateITCase` is a good place to add the it case.

fluss-lake/fluss-lake-paimon/src/main/java/org/apache/fluss/lake/paimon/utils/FlussToPaimonTableChangeConverter.java
```diff
@@ -0,0 +1,71 @@
...
+
+    public static List<SchemaChange> convert(
+            List<TableChange> tableChanges, Function<String, String> keyFun) {
+        List<SchemaChange> schemaChanges = new ArrayList<>(tableChanges.size());
+
+        for (TableChange tableChange : tableChanges) {
+            if (tableChange instanceof TableChange.SetOption) {
+                TableChange.SetOption setOption = (TableChange.SetOption) tableChange;
+                String key = setOption.getKey();
+                // Here we only prohibit direct modification of Paimon properties;
```
luoyuxia: why prohibit modification of paimon properties?

LiebingYu: IIUC, we want Fluss to manage the underlying lake table. Therefore, any property modifications to the Paimon table should be made by updating the Fluss table, rather than directly modifying the properties of the underlying Paimon table, as doing so could lead to unexpected issues.

For example, if we want to modify the bucket number of a table, we should not modify the `bucket` property of the Paimon table directly. Instead, we should update the `bucket.num` property of the Fluss table. Directly changing the `bucket` property of the Paimon table may result in unforeseen problems.

LiebingYu: After https://github.com/apache/fluss/pull/1760 we can safely remove it here.

LiebingYu:
CC @luoyuxia. I have rebase main and fix the comment, pls take a look when you have time.

fluss-lake/fluss-lake-paimon/src/main/java/org/apache/fluss/lake/paimon/utils/FlussToPaimonTableChangeConverter.java
```diff
@@ -0,0 +1,52 @@
...
+import org.apache.fluss.metadata.TableChange;
+
+import org.apache.paimon.schema.SchemaChange;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.function.Function;
+
+/** Converter for {@link TableChange} to {@link SchemaChange}. */
+public class FlussToPaimonTableChangeConverter {
```
luoyuxia: Move to `PaimonConversions`

fluss-lake/fluss-lake-paimon/src/main/java/org/apache/fluss/lake/paimon/utils/FlussToPaimonTableChangeConverter.java
```diff
@@ -0,0 +1,52 @@
...
+
+    public static List<SchemaChange> convert(
```
luoyuxia: nit:
keyFun -> optionKeyTransformer

luoyuxia (commented):
@LiebingYu Thanks for the pr. Left minor comments.

PR #1625, opened 2025-08-30
[server] feat: support alter table properties

gyang94:
### Purpose

Linked issue: close #100 
Support alter table properties and custom properties

PR #1760, opened 2025-09-25
[server] Improve the implementation of ALTER TABLE

wuchong:
### Purpose

Improve the implementation of #1625

Linked issue: close #xxx

### Brief change log

1. Rename the proto messages to better reflect the meaning (we need to update FIP as well)
    - `AlterTableConfigsRequest` -> `AlterTablePropertiesRequest`, to avoid confusion with `TableConfig` which referes to the table storage only configs. 
    - `PbAlterConfigsRequestInfo` -> `PbAlterConfig`, for simplification and use singular. This message will also be used by AlterCluster RPC in the near future. 
2. Validate the `bucket.num`, `bucket.key`, `bootstrap.servers` in the connector side, because we don't support yet. 
3. Throw exception if changing the `paimon.*` properties because we don't support yet.
4. Improve some code

### DIFF
--- a/fluss-flink/fluss-flink-common/src/test/java/org/apache/fluss/flink/catalog/FlinkCatalogITCase.java
+++ b/fluss-flink/fluss-flink-common/src/test/java/org/apache/fluss/flink/catalog/FlinkCatalogITCase.java
@@ -297,6 +297,7 @@ abstract class FlinkCatalogITCase {
         assertThat(table.getUnresolvedSchema()).isEqualTo(expectedSchema);
         Map<String, String> expectedOptions = new HashMap<>();
         expectedOptions.put("bucket.num", "10");
+        expectedOptions.put("table.datalake.format", "paimon");
         assertOptionsEqual(table.getOptions(), expectedOptions);
     }
 
@@ -564,6 +565,7 @@ abstract class FlinkCatalogITCase {
         expectedOptions.put("k1", "v1");
         expectedOptions.put(BUCKET_KEY.key(), "user");
         expectedOptions.put(BUCKET_NUMBER.key(), "1");
+        expectedOptions.put("table.datalake.format", "paimon");
         assertOptionsEqual(table.getOptions(), expectedOptions);
     }
 
--- a/fluss-server/src/main/java/org/apache/fluss/server/coordinator/CoordinatorService.java
+++ b/fluss-server/src/main/java/org/apache/fluss/server/coordinator/CoordinatorService.java
@@ -340,20 +340,19 @@ public final class CoordinatorService extends RpcServiceBase implements Coordina
             newDescriptor = newDescriptor.withProperties(newProperties);
         }
 
+        // override the datalake format if the table hasn't set it and the cluster configured
+        if (dataLakeFormat != null
+                && !properties.containsKey(ConfigOptions.TABLE_DATALAKE_FORMAT.key())) {
+            newDescriptor = newDescriptor.withDataLakeFormat(dataLakeFormat);
+        }
+
         // lake table can only be enabled when the cluster configures datalake format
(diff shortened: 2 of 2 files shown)


wuchong:
cc @gyang94

fluss-server/src/main/java/org/apache/fluss/server/coordinator/MetadataManager.java
```diff
@@ -308,52 +308,62 @@ public void alterTableProperties(
...
-
         try {
-            TableRegistration updatedTableRegistration =
-                    getUpdatedTableRegistration(tablePath, tablePropertyChanges);
-            if (updatedTableRegistration != null) {
+            // it throws TableNotExistException if the table or database not exists
+            TableRegistration tableReg = getTableRegistration(tablePath);
+            SchemaInfo schemaInfo = getLatestSchema(tablePath);
+            // we can't use MetadataManager#getTable here, because it will add the default
+            // lake options to the table properties, which may cause the validation failure
```
gyang94, line 316: A small question: now it looks those default properties won't affect the `validateAlterTableProperties` validation below. 
When will those default values make validation failure? Default values should be exactly what takes effect to a table.

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/FlinkConnectorOptions.java
```diff
@@ -132,6 +133,13 @@ public class FlinkConnectorOptions {
     public static final List<ConfigOption<?>> CLIENT_OPTIONS =
             FlinkConversions.toFlinkOptions(FlussConfigUtils.CLIENT_OPTIONS.values());
 
+    // --------------------------------------------------------------------------------------------
+    // modification disallowed connector options
+    // --------------------------------------------------------------------------------------------
+
+    public static final List<String> ALTER_DISALLOW_OPTIONS =
+            Arrays.asList(BUCKET_NUMBER.key(), BUCKET_KEY.key(), BOOTSTRAP_SERVERS.key());
```
gyang94, line 141: Should fluss server side also keeps a disallow list like this?

gyang94:
Looks good to me! Left two small general questions, it won't block the PR.
cc @wuchong

luoyuxia (approved):
@wuchong Thanks for the great improvement, which makes the alter table solid. LGTM!

fluss-server/src/main/java/org/apache/fluss/server/coordinator/MetadataManager.java
wuchong: The default datalake properties like `table.datalake.paimon.warehouse=...`. This is not included in `org.apache.fluss.config.FlussConfigUtils#TABLE_OPTIONS` and will fail when running `org.apache.fluss.server.utils.TableDescriptorValidation#validateTableDescriptor` with this config. So we exclude the default lake properties for validation, just like how a table is validated when table is created.

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/FlinkConnectorOptions.java
wuchong: These properties are flink specific options which will be translated into Fluss `TableDescriptor.TableDistribution` structure. So when supporting altering these properties, we need to convert to a `TableDistribution` change. Because Fluss only understand `TableDistribution`, not the `bucket.number` and `bucket.key` options. These options are custom properties to Fluss server. So we only validate them in flink connector side, not in server side.  Server side should only handle the `TableDistribution`.

wuchong:
Merging...

wuchong:
I updated https://cwiki.apache.org/confluence/display/FLUSS/FIP-15%3A+Alter+Table+Interface for the API changes.

PR #1766, opened 2025-09-28
[server] feat: support multiple table change types

### LINKED ISSUE #1762: Support fire multiple table changes via a single alter table rpc request
### Search before asking

### Description

We already support alter table property. We need to send a `AlterTablePropertiesRequest`.
But if we want to fire multiple table changes,  such like alter table property, alter add column. We need to fire multiple requests. There are two main drawbacks:
1: the alter table changes can't be atomic since we have to send multiple requests
2: the more important is that some  table changes may work together, such as add a new column & set the property about the colum.   It must be in single one request. Otherwise, the alter will fail.

It's easy to extend the current design  to support it. 
Change `AlterTablePropertiesRequest` to `AlterTableRequest` should be fine. If in the future, we support alter add column, it may be look like 
```
AlterTablePropertiesRequest {
   required PbTablePath table_path = 1;
  required bool ignore_if_not_exists = 2;
  repeated PbAlterConfig config_changes = 3;
  repeated PbAlterColumn column_changes = 4;
}

```

### Willingness to contribute

gyang94:
### Purpose

Linked issue: close #1762

### DIFF
--- a/fluss-client/src/main/java/org/apache/fluss/client/admin/FlussAdmin.java
+++ b/fluss-client/src/main/java/org/apache/fluss/client/admin/FlussAdmin.java
@@ -42,7 +42,7 @@ import org.apache.fluss.rpc.RpcClient;
 import org.apache.fluss.rpc.gateway.AdminGateway;
 import org.apache.fluss.rpc.gateway.AdminReadOnlyGateway;
 import org.apache.fluss.rpc.gateway.TabletServerGateway;
-import org.apache.fluss.rpc.messages.AlterTablePropertiesRequest;
+import org.apache.fluss.rpc.messages.AlterTableRequest;
 import org.apache.fluss.rpc.messages.CreateAclsRequest;
 import org.apache.fluss.rpc.messages.CreateDatabaseRequest;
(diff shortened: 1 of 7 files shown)


fluss-client/src/main/java/org/apache/fluss/client/admin/FlussAdmin.java
```diff
@@ -255,7 +255,7 @@ public CompletableFuture<Void> alterTable(
                 .setTablePath()
                 .setDatabaseName(tablePath.getDatabaseName())
                 .setTableName(tablePath.getTableName());
-        return gateway.alterTableProperties(request).thenApply(r -> null);
+        return gateway.alterTable(request).thenApply(r -> null);
```
luoyuxia, line 258: update java doc of method `AdminGateway#alterTable`
Currently, it's `the request to alter the configs of a table.`

luoyuxia (commented):
@gyang94 Thanks for the pr. Only left one comment. PTAL

wuchong (approved):

PR #1771, opened 2025-09-29
[server] Forbid resetting table.datalake.enabled

### LINKED ISSUE #1769: [server] Forbid resetting `table.datalake.enabled`
### Search before asking

### Fluss version

main (development)

### Please describe the bug 🐞

https://github.com/apache/fluss/pull/1753 support alter `table.datalake.enabled`. It allows users to reset(delete) `table.datalake.enabled` as well, which may case unexpected things to happen.

### Solution

Since `table.datalake.enabled` is a mandatory parameter for Fluss tables, in my opinion, we should directly prohibit resetting this parameter and always require users to explicitly set it to either `true` or `false`.

LiebingYu:
### Purpose

Linked issue: close #1769


gyang94:
Hi @LiebingYu, to disable reset `table.datalake.enabled`, should we also consider to disable to set `table.datalake.enabled` from true to false? They are equivalent.
Only allow to alter this parameter from false to true, not from true to false.

PR #1783, opened 2025-10-08
Support table.delete.behavior config to disable or ignore deletion on…

### LINKED ISSUE #1778: Support `table.delete.behavior` config to disable or ignore deletion on the table
### Search before asking

### Motivation

Currently, Fluss primary-key tables allow delete operations by default. However, certain use cases, such as delta joins or log deduplication (expected append-only changelogs) may want to prevent deletions. 

While Fluss provides a sink connector option, `sink.ignore.delete`, which silently drops delete records at the sink, this approach is not sufficient for system-level guarantees. Connector options can be overridden dynamically via SQL hints, making them unsuitable for enforcing strict table-level semantics.

To ensure strong, immutable guarantees that no deletions occur in the changelog stream, we propose a table-level configuration (e.g., `table.delete.behavior`) that is enforced by the system itself, either by rejecting deletes with an error or silently ignoring them, regardless of connector settings or client-side overrides.

This would allow users to safely declare a table as append-only, upsert-only enabling reliable downstream processing for use cases like delta joins. 

### Solution

We propose to introduce a unified, enum-based configuration to explicitly control this behavior on server side:

```
table.delete.behavior = [allow | disable | ignore]
```

- `allow` (default): Allow normal delete operations.
- `ignore`: Silently skip delete requests without error (may expose a metric for this case in the future). 
- `disable`: Reject delete requests with a clear error message (e.g., for strict immutability).

We should also update `org.apache.fluss.flink.source.FlinkTableSource#getChangelogMode` to reflect this behavior that no deletions in the generated changelogs.  

### Anything else?

We should add checks when creating tables with this option set: 

1. Only pk tables can set this option explicitly, log tables with this option should throw exception. 
2. Currently, the `FIRST_ROW` and `VERSIONED` merge engines **do not support delete operations**.  
Therefore, tables using these engines should **automatically default** to `table.delete.behavior=ignore`.  
Users may explicitly set `table.delete.behavior=disable` to reject deletions,  
but **setting `table.delete.behavior=allow` must throw an exception**, as deletes are fundamentally unsupported by these merge engines.

### Willingness to contribute

caozhen1937:
… the table

### Purpose

Linked issue: close #1778

### DIFF
--- a/fluss-server/src/main/java/org/apache/fluss/server/coordinator/CoordinatorService.java
+++ b/fluss-server/src/main/java/org/apache/fluss/server/coordinator/CoordinatorService.java
@@ -36,6 +36,7 @@ import org.apache.fluss.fs.FileSystem;
 import org.apache.fluss.metadata.DataLakeFormat;
 import org.apache.fluss.metadata.DatabaseDescriptor;
 import org.apache.fluss.metadata.DeleteBehavior;
+import org.apache.fluss.metadata.MergeEngineType;
 import org.apache.fluss.metadata.PartitionSpec;
 import org.apache.fluss.metadata.ResolvedPartitionSpec;
 import org.apache.fluss.metadata.TableChange;
@@ -408,13 +409,17 @@ public final class CoordinatorService extends RpcServiceBase implements Coordina
         // For tables with first_row or versioned merge engines, automatically set to IGNORE if
         // delete behavior is not set
         Configuration tableConf = Configuration.fromMap(tableDescriptor.getProperties());
-        if (!tableConf.getOptional(ConfigOptions.TABLE_DELETE_BEHAVIOR).isPresent()) {
-            Map<String, String> newProperties = new HashMap<>(newDescriptor.getProperties());
-            newProperties.put(
-                    ConfigOptions.TABLE_DELETE_BEHAVIOR.key(), DeleteBehavior.IGNORE.name());
-            newDescriptor = newDescriptor.withProperties(newProperties);
+        MergeEngineType mergeEngine =
+                tableConf.getOptional(ConfigOptions.TABLE_MERGE_ENGINE).orElse(null);
+        if (mergeEngine == MergeEngineType.FIRST_ROW || mergeEngine == MergeEngineType.VERSIONED) {
+            if (tableDescriptor.hasPrimaryKey()
+                    && !tableConf.getOptional(ConfigOptions.TABLE_DELETE_BEHAVIOR).isPresent()) {
+                Map<String, String> newProperties = new HashMap<>(newDescriptor.getProperties());
+                newProperties.put(
+                        ConfigOptions.TABLE_DELETE_BEHAVIOR.key(), DeleteBehavior.IGNORE.name());
+                newDescriptor = newDescriptor.withProperties(newProperties);
+            }
         }
-
         return newDescriptor;
     }
 
--- a/fluss-server/src/main/java/org/apache/fluss/server/utils/TableDescriptorValidation.java
+++ b/fluss-server/src/main/java/org/apache/fluss/server/utils/TableDescriptorValidation.java
@@ -323,7 +323,7 @@ public class TableDescriptorValidation {
                 tableConf.getOptional(ConfigOptions.TABLE_DELETE_BEHAVIOR);
         if (!hasPrimaryKey && deleteBehaviorOptional.isPresent()) {
             throw new InvalidConfigException(
-                    "Delete behavior configuration is only supported for primary key tables.");
+                    "The 'table.delete.behavior' configuration is only supported for primary key tables.");
         }
 
         // For tables with merge engines, automatically set appropriate delete behavior
@@ -336,7 +336,7 @@ public class TableDescriptorValidation {
                 throw new InvalidConfigException(
                         String.format(
                                 "Table with '%s' merge engine does not support delete operations. "
-                                        + "The delete behavior must be set to 'ignore' or 'disable', but got 'allow'.",
+                                        + "The 'table.delete.behavior' config must be set to 'ignore' or 'disable', but got 'allow'.",
                                 mergeEngine));
             }
         }
--- a/fluss-client/src/test/java/org/apache/fluss/client/admin/FlussAdminITCase.java
+++ b/fluss-client/src/test/java/org/apache/fluss/client/admin/FlussAdminITCase.java
@@ -363,7 +363,7 @@ class FlussAdminITCase extends ClientToServerITCaseBase {
                 .hasRootCauseInstanceOf(InvalidConfigException.class)
                 .hasMessageContaining(
                         "Table with 'FIRST_ROW' merge engine does not support delete operations. "
-                                + "The delete behavior must be set to 'ignore' or 'disable', but got 'allow'.");
+                                + "The 'table.delete.behavior' config must be set to 'ignore' or 'disable', but got 'allow'.");
 
         // Test 4: VERSIONED merge engine with delete behavior explicitly set to ALLOW
         TablePath tablePath4 = TablePath.of("fluss", "test_allow_delete_for_versioned");
@@ -381,7 +381,7 @@ class FlussAdminITCase extends ClientToServerITCaseBase {
                 .hasRootCauseInstanceOf(InvalidConfigException.class)
                 .hasMessageContaining(
                         "Table with 'VERSIONED' merge engine does not support delete operations. "
-                                + "The delete behavior must be set to 'ignore' or 'disable', but got 'allow'.");
+                                + "The 'table.delete.behavior' config must be set to 'ignore' or 'disable', but got 'allow'.");
 
         // Test 5: Log table - not allow to set delete behavior
         TablePath tablePath5 = TablePath.of("fluss", "test_set_delete_behavior_for_log_table");
@@ -396,7 +396,7 @@ class FlussAdminITCase extends ClientToServerITCaseBase {
         assertThatThrownBy(() -> admin.createTable(tablePath5, tableDescriptor5, false).join())
                 .hasRootCauseInstanceOf(InvalidConfigException.class)
                 .hasMessageContaining(
-                        "Delete behavior configuration is only supported for primary key tables.");
+                        "The 'table.delete.behavior' configuration is only supported for primary key tables.");
     }
 
     @Test


fluss-flink/fluss-flink-common/src/test/java/org/apache/fluss/flink/sink/FlinkTableSinkITCase.java
```diff
@@ -1095,4 +1095,107 @@ public InsertAndExpectValues(List<String> insertValues, List<String> expectedRow
...
+        TablePath tablePath = TablePath.of(DEFAULT_DB, tableName);
+        assertThatThrownBy(
+                        () ->
+                                tBatchEnv
+                                        .executeSql("DELETE FROM " + tableName + " WHERE a = 1")
+                                        .await())
+                .isInstanceOf(UnsupportedOperationException.class)
+                .hasMessage(
+                        String.format(
+                                "Table %s has delete behavior set to 'disable' which does not support DELETE statements.",
```
wuchong, line 1122: Add a `INSERT INTO` pipeline with `-D` data in it, to verify the behavior?

fluss-server/src/main/java/org/apache/fluss/server/kv/rowmerger/DefaultRowMerger.java
```diff
@@ -52,13 +59,26 @@ public BinaryRow merge(BinaryRow oldRow, BinaryRow newRow) {
...
+                        "Delete operations are disabled for this table. "
+                                + "The table.delete.behavior is set to 'disable'.");
+            default:
+                throw new IllegalArgumentException(
+                        "Unsupported delete behavior: " + deleteBehavior);
+        }
     }
 
     @Override
     public boolean supportsDelete() {
```
wuchong: We can refactor the `supportsDelete` interface to `deleteBehavior` to return the `DeleteBehavior` enum type. So `KvTablet` can directly decide ignore or throw exception or call `RowMerger#delete()` method for all RowMergers.

fluss-server/src/main/java/org/apache/fluss/server/kv/rowmerger/DefaultRowMerger.java
```diff
@@ -52,13 +59,26 @@ public BinaryRow merge(BinaryRow oldRow, BinaryRow newRow) {
...
     public BinaryRow delete(BinaryRow oldRow) {
-        // returns null to indicate the row is deleted
-        return null;
+        switch (deleteBehavior) {
+            case ALLOW:
+                // returns null to indicate the row is deleted
+                return null;
+            case IGNORE:
+                // returns the old row unchanged to ignore the delete operation
+                return oldRow;
```
wuchong: The `delete()` method currently is designed to return a merged row and generate `-U` and `+U` for it. So just returning `oldRow` doesn't match the behaivor of ignore, because it generates `-U`, `+U`. Can we rafactor the `supportsDelete` interface a bit, as mentioned below.

fluss-server/src/main/java/org/apache/fluss/server/utils/TableDescriptorValidation.java
```diff
@@ -316,6 +318,30 @@ private static void checkPartition(
         }
     }
 
+    private static void checkDeleteBehavior(Configuration tableConf, boolean hasPrimaryKey) {
+        DeleteBehavior deleteBehavior = tableConf.get(ConfigOptions.TABLE_DELETE_BEHAVIOR);
+        if (deleteBehavior != DeleteBehavior.ALLOW) {
```
wuchong: We don't support to ignore deletes if the delete behavior is set to IGNORE, so I suggest to simply no deleteBehavior set on the log table. You can use `tableConf.getOptional` to get whether the config is set or not.

fluss-server/src/main/java/org/apache/fluss/server/utils/TableDescriptorValidation.java
```diff
@@ -316,6 +318,30 @@ private static void checkPartition(
...
+            if (!hasPrimaryKey) {
+                throw new InvalidConfigException(
+                        "Delete behavior configuration is only supported for primary key tables.");
+            }
+        }
+
+        // For tables with merge engines, automatically set appropriate delete behavior
+        MergeEngineType mergeEngine = tableConf.get(ConfigOptions.TABLE_MERGE_ENGINE);
+        if (mergeEngine != null) {
```
wuchong: better to explicity check the merge engine type here, we will soon introduce new merge engine types which may not satisify this.

fluss-server/src/main/java/org/apache/fluss/server/utils/TableDescriptorValidation.java
```diff
@@ -316,6 +318,30 @@ private static void checkPartition(
         }
     }
 
+    private static void checkDeleteBehavior(Configuration tableConf, boolean hasPrimaryKey) {
```
wuchong, line 321: We need add create table tests to verify the exceptions.

fluss-server/src/main/java/org/apache/fluss/server/coordinator/CoordinatorService.java
```diff
@@ -389,6 +391,20 @@ private TableDescriptor applySystemDefaults(TableDescriptor tableDescriptor) {
                             ConfigOptions.TABLE_DATALAKE_ENABLED.key()));
         }
 
+        // For tables with first_row or versioned merge engines, automatically set to IGNORE if
```
wuchong, line 409: we need to add create table tests to verify the automantically set configs.

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/catalog/FlinkTableFactory.java
```diff
@@ -225,10 +226,10 @@ private static Configuration toFlussClientConfig(
                 ConfigOptions.BOOTSTRAP_SERVERS.key(),
                 tableOptions.get(FlinkConnectorOptions.BOOTSTRAP_SERVERS.key()));
 
-        // forward all client configs
+        // forward all client configs and table configs
         tableOptions.forEach(
                 (key, value) -> {
-                    if (key.startsWith(CLIENT_PREFIX)) {
+                    if (key.startsWith(CLIENT_PREFIX) || key.startsWith(TABLE_PREFIX)) {
```
wuchong: Why forward table storage configs? The returned flinkConfigs is mainly used to set environment or connection for the flink job.

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/source/FlinkTableSource.java
```diff
@@ -203,10 +206,29 @@ public ChangelogMode getChangelogMode() {
         } else {
             if (hasPrimaryKey()) {
                 // pk table
-                if (mergeEngineType == MergeEngineType.FIRST_ROW) {
+                if (mergeEngineType == MergeEngineType.FIRST_ROW
+                        || mergeEngineType == MergeEngineType.VERSIONED) {
```
wuchong: VERSIONED merge engine can generate updates, so we shouldn't add VERSIONED here.

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/source/FlinkTableSource.java
```diff
@@ -203,10 +206,29 @@ public ChangelogMode getChangelogMode() {
...
+                            tableConf.get(ConfigOptions.TABLE_DELETE_BEHAVIOR);
+                    if (deleteBehavior == DeleteBehavior.IGNORE) {
+                        // If delete operations are ignored, only insert and update are relevant
+                        return ChangelogMode.newBuilder()
+                                .addContainedKind(RowKind.INSERT)
+                                .addContainedKind(RowKind.UPDATE_BEFORE)
+                                .addContainedKind(RowKind.UPDATE_AFTER)
+                                .build();
+                    } else if (deleteBehavior == DeleteBehavior.DISABLE) {
+                        throw new UnsupportedOperationException(
```
wuchong: We shouldn't throw exception here, otherwise, all scan to a delete disabled table will fail at compile phase. Delete disabled table is the same to delete ignored table that doesn't generate delete events for the changelog mode.

website/docs/engine-flink/options.md
```diff
@@ -83,6 +83,7 @@ ALTER TABLE log_table SET ('table.log.ttl' = '7d');
...
+| table.delete.behavior                    | Enum     | ALLOW                               | Controls the behavior of delete operations on primary key tables. Three modes are supported: `ALLOW` (default) - allows normal delete operations; `IGNORE` - silently ignores delete requests without errors; `DISABLE` - rejects delete requests and throws explicit errors. This configuration provides system-level guarantees for specific use cases like delta joins or log deduplication where deletions should never occur. For tables with `first_row` or `versioned` merge engines, this option is automatically set to `IGNORE` and cannot be overridden. Only applicable to primary key tables; log tables are not affected by this setting.                                                                                                                                                                                                                                                                          |
```
wuchong: "This configuration provides system-level guarantees for specific use cases like delta joins or log deduplication where deletions should never occur."

==>

This configuration provides system-level guarantees for some downstream pipelines (e.g., Flink Delta Join) that must not receive any delete events in the changelog of the table.

wuchong: Remove "log tables are not affected by this setting" as we throw exception when defining this config on log table.

fluss-server/src/main/java/org/apache/fluss/server/kv/rowmerger/DefaultRowMerger.java
```diff
@@ -52,13 +59,26 @@ public BinaryRow merge(BinaryRow oldRow, BinaryRow newRow) {
...
+            case DISABLE:
+                throw new UnsupportedOperationException(
```
wuchong: We should introduce an `ApiException`, e.g., `DeletionDisabledException`, to allow this exception can be returned to the client side.

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/catalog/FlinkTableFactory.java
caozhen1937: Originally, it was intended to transmit the "delete behavior" param in order to control the "Delete" semantics of the sink. I might have misunderstood before. The "delete behavier" param only controls the changelog.

fluss-server/src/main/java/org/apache/fluss/server/utils/TableDescriptorValidation.java
```diff
@@ -319,20 +319,20 @@ private static void checkPartition(
...
         // For tables with merge engines, automatically set appropriate delete behavior
         MergeEngineType mergeEngine = tableConf.get(ConfigOptions.TABLE_MERGE_ENGINE);
-        if (mergeEngine != null) {
+        if (mergeEngine == MergeEngineType.FIRST_ROW || mergeEngine == MergeEngineType.VERSIONED) {
             // For FIRST_ROW and VERSIONED merge engines, delete operations are not supported
             // If user explicitly sets delete behavior to ALLOW, throw an exception
-            if (deleteBehavior == DeleteBehavior.ALLOW) {
+            if (deleteBehaviorOptional.isPresent()
+                    && deleteBehaviorOptional.get() == DeleteBehavior.ALLOW) {
                 throw new InvalidConfigException(
```
caozhen1937, line 336: @wuchong We have set "ignore" for FIRST_ROW and VERSIONED by default. In reality, this check logic will not be triggered.

wuchong: We should only set `ignore` for FIRST_ROW and VERSIONED if delete behavior is not set by users. If user explicitly defines delete behavior, we should honor it and validate the properties. Otherwise, it will be a hard overwrite which is very confusing to users.

wuchong (approved):
Thanks for the updates @caozhen1937 , I have reviewed again and added a comment to resolve the remaining issues. 

1. I moved the tests in `TableManagerITCase` to `FlussAdminITCase`, this can cover the create table behavior from a user perspective. I also added tests to verify creating merge engine tables without defining delete behavior (should automantically add ignore delete behavior property). 
2. Throw exception is explicitly execute `DELETE FROM` statement on the table with delete behavior of `disable`. 
3. Add tests to verify the exception if writing delete events to the disabled delete behavior tables. 
4. Add compatibility for RowMergers, because tables in v0.7 don't have delete behavior property, and use ALLOW by default will cause problems when initializing these tables.

PR #1784, opened 2025-10-09
Fluss sink supports dynamic shuffle.

### LINKED ISSUE #1789: Fluss sink supports dynamic shuffle
### Search before asking

### Motivation

The current Flink sink supports bucket-based shuffle by bucket ID. It's aim to improve batching efficiency: Data in the same bucket is always written by a single subtask, ensuring better batching performance.

However, still some problems:

1. Bottleneck from single subtask writing a bucket. Scaling the parallelism will not resolve bottlenecks if a single subtask becomes the performance constraint for a bucket.
2. Uneven traffic distribution across sink subtasks:

- Even with balanced input traffic, the number of partitions assigned to each sink subtask may vary significantly. For example, some subtasks may handle no partitions, while others manage multiple (see Table 1).
- If traffic across partitions is inherently imbalanced, this issue is exacerbated. A subtask processing partitions with high traffic may become a hotspot bottleneck.
3. The current design is ineffective for log tables without a bucket key .

### Solution

Referring what iceberg  [range mode](https://iceberg.apache.org/docs/latest/flink-writes/#range-distribution-experimental) does: Statistics are collected by every shuffle operator subtask and aggregated by the coordinator for every checkpoint cycle. 

### Anything else?

_No response_

### Willingness to contribute

loserwang1024:
### Purpose

Linked issue: close #1789

### DIFF
--- a/fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/sink/FlinkSink.java
+++ b/fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/sink/FlinkSink.java
@@ -38,9 +38,12 @@ import org.apache.flink.api.common.typeinfo.TypeInformation;
 import org.apache.flink.api.connector.sink2.SinkWriter;
 import org.apache.flink.metrics.groups.SinkWriterMetricGroup;
 import org.apache.flink.runtime.metrics.groups.InternalSinkWriterMetricGroup;
+import org.apache.flink.streaming.api.connector.sink2.SupportsPreWriteTopology;
 import org.apache.flink.streaming.api.datastream.DataStream;
 import org.apache.flink.streaming.api.datastream.DataStreamSink;
 import org.apache.flink.streaming.api.datastream.SingleOutputStreamOperator;
+import org.apache.flink.table.connector.sink.DataStreamSinkProvider;
+import org.apache.flink.table.connector.sink.SinkV2Provider;
 import org.apache.flink.table.types.logical.RowType;
 
 import javax.annotation.Nullable;
@@ -56,7 +59,7 @@ class FlinkSink<InputT> extends SinkAdapter<InputT> {
 
     private static final long serialVersionUID = 1L;
 
-    private final SinkWriterBuilder<? extends FlinkSinkWriter, InputT> builder;
+    protected final SinkWriterBuilder<? extends FlinkSinkWriter, InputT> builder;
     private final TablePath tablePath;
 
     FlinkSink(SinkWriterBuilder<? extends FlinkSinkWriter, InputT> builder, TablePath tablePath) {
@@ -72,6 +75,17 @@ class FlinkSink<InputT> extends SinkAdapter<InputT> {
         return flinkSinkWriter;
     }
 
+    /**
+     * {@link FlinkSink} serves as the SQL connector. Since we uses {@link DataStreamSinkProvider}
+     * (rather than {@link SinkV2Provider}), it does not automatically recognize or invoke the
+     * {@link SupportsPreWriteTopology} interface. Therefore, the pre-write topology must be added
+     * manually here.
+     *
+     * <p>In contrast, {@link FlussSink} is used directly as a DataStream connector, Flink’s runtime
+     * explicitly checks for the {@link SupportsPreWriteTopology} interface and automatically
+     * incorporates the pre-write topology if present. To support this path, the {@link
+     * SupportsPreWriteTopology} implementation resides in {@link FlussSink}.
+     */
     public DataStreamSink<InputT> apply(DataStream<InputT> input) {
         return builder.addPreWriteTopology(input)
                 .sinkTo(this)
--- a/fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/sink/FlussSink.java
+++ b/fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/sink/FlussSink.java
@@ -21,6 +21,9 @@ import org.apache.fluss.annotation.PublicEvolving;
 import org.apache.fluss.flink.sink.writer.FlinkSinkWriter;
 import org.apache.fluss.metadata.TablePath;
 
+import org.apache.flink.streaming.api.connector.sink2.SupportsPreWriteTopology;
+import org.apache.flink.streaming.api.datastream.DataStream;
+
 /**
  * FlussSink is a specialized Flink sink for writing data to Fluss.
  *
@@ -32,7 +35,8 @@ import org.apache.fluss.metadata.TablePath;
  * @since 0.7
  */
 @PublicEvolving
-public class FlussSink<InputT> extends FlinkSink<InputT> {
+public class FlussSink<InputT> extends FlinkSink<InputT>
+        implements SupportsPreWriteTopology<InputT> {
     private static final long serialVersionUID = 1L;
 
     /**
@@ -44,6 +48,11 @@ public class FlussSink<InputT> extends FlinkSink<InputT> {
             SinkWriterBuilder<? extends FlinkSinkWriter<InputT>, InputT> builder,
             TablePath tablePath) {
         super(builder, tablePath);
+    }
+
+    @Override
+    public DataStream<InputT> addPreWriteTopology(DataStream<InputT> input) {
+        return builder.addPreWriteTopology(input);
     }
 
     /**
--- a/fluss-flink/fluss-flink-common/src/test/java/org/apache/fluss/flink/sink/FlussSinkITCase.java
+++ b/fluss-flink/fluss-flink-common/src/test/java/org/apache/fluss/flink/sink/FlussSinkITCase.java
@@ -144,7 +144,7 @@ public class FlussSinkITCase extends FlinkTestBase {
                         .setSerializationSchema(serializationSchema)
                         .build();
 
-        flussSink.apply(stream).name("Fluss Sink");
+        stream.sinkTo(flussSink).name("Fluss Sink");
 
         env.executeAsync("Test RowData Fluss Sink");
 
@@ -272,7 +272,7 @@ public class FlussSinkITCase extends FlinkTestBase {
                         .setSerializationSchema(new TestOrderSerializationSchema())
                         .build();
 
-        flussSink.apply(stream).name("Fluss Sink");
+        stream.sinkTo(flussSink).name("Fluss Sink");
         env.executeAsync("Test Order Fluss Sink");
 
         Table table = conn.getTable(new TablePath(DEFAULT_DB, pkTableName));


loserwang1024:
@wuchong @leonardBang , CC

loserwang1024:
@wuchong , CC, this pr is needed.

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/sink/shuffle/StatisticsOrRecord.java
```diff
@@ -0,0 +1,117 @@
...
+
+import static org.apache.fluss.utils.Preconditions.checkArgument;
+
+/**
+ * Either a record or a statistics.
+ *
+ * @param <InputT>
+ */
+@Internal
+public class StatisticsOrRecord<InputT> {
```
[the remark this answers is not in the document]
loserwang1024: image

loserwang1024:
@wuchong , I have rebase this pr, CC

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/sink/shuffle/DataStatisticsCoordinator.java
```diff
@@ -0,0 +1,403 @@
...
+    }
+
+    private static class CoordinatorExecutorThreadFactory
+            implements ThreadFactory, Thread.UncaughtExceptionHandler {
+
+        private final String coordinatorThreadName;
+        private final ClassLoader classLoader;
+        private final Thread.UncaughtExceptionHandler errorHandler;
+
+        @javax.annotation.Nullable private Thread thread;
```
wuchong: remove unnecessary qualification

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/sink/shuffle/DataStatisticsCoordinator.java
```diff
@@ -0,0 +1,403 @@
...
+
+        CoordinatorExecutorThreadFactory(
+                final String coordinatorThreadName, final ClassLoader contextClassLoader) {
+            this(coordinatorThreadName, contextClassLoader, FatalExitExceptionHandler.INSTANCE);
+        }
+
+        @org.apache.flink.annotation.VisibleForTesting
```
wuchong: Use Fluss `@VisibleForTesting`.

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/sink/shuffle/DataStatisticsOperator.java
```diff
@@ -0,0 +1,169 @@
...
+        StatisticsEvent statisticsEvent = (StatisticsEvent) event;
+        LOG.info(
+                "Operator {} subtask {} received global data event from coordinator checkpoint {}",
+                operatorName,
+                subtaskIndex,
+                statisticsEvent.checkpointId());
+        DataStatistics globalStatistics =
+                StatisticsUtil.deserializeDataStatistics(
+                        statisticsEvent.statisticsBytes(), statisticsSerializer);
+        LOG.info("Global data statistics: {}", globalStatistics);
```
wuchong: `globalStatistics` is very huge, should use `debug` log level here.

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/sink/shuffle/StatisticsOrRecordChannelComputer.java
```diff
@@ -0,0 +1,314 @@
...
+                                : new WeightedRandomAssignment(
+                                        assignedSubtasks, subtaskWeights, random);
+                assignmentMap.put(currentKey, keyAssignment);
+                assignedSubtasks = new ArrayList<>();
+                subtaskWeights = new ArrayList<>();
+                currentKey = null;
+            }
+        }
+
+        LOG.info("Assignment map: {}", assignmentMap);
```
wuchong: Use `debug` log, otherwise, too many logs.

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/sink/shuffle/StatisticsOrRecordChannelComputer.java
```diff
@@ -0,0 +1,314 @@
...
+                RowWithOp rowWithOp = serializationSchema.serialize(wrapper.record());
+                InternalRow row = rowWithOp.getRow();
+                String partitionName = partitionGetter.getPartition(row);
+                return delegatePartitioner.select(partitionName, row, downstreamNumChannels);
+            }
+        } catch (Exception e) {
+            throw new FlussRuntimeException(
+                    String.format(
+                            "Failed to serialize static or record of type '%s' in FlinkRowDataChannelComputer: %s",
+                            wrapper != null ? wrapper.getClass().getName() : "null",
```
wuchong: wrapper is always not null here

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/sink/shuffle/StatisticsOrRecord.java
```diff
@@ -0,0 +1,117 @@
...
+            StatisticsOrRecord<InputT> reuse, TypeSerializer<DataStatistics> statisticsSerializer) {
+        if (reuse.hasStatistics()) {
+            return reuse;
+        } else {
+            // not reusable
+            return StatisticsOrRecord.fromStatistics(statisticsSerializer.createInstance());
+        }
+    }
+
+    boolean hasStatistics() {
```
wuchong: nit: rename to `isStatistics()`

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/sink/shuffle/StatisticsOrRecord.java
```diff
@@ -0,0 +1,117 @@
...
+        return statistics != null;
+    }
+
+    public boolean hasRecord() {
```
wuchong: nit: rename to `isRecord`.

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/sink/shuffle/StatisticsOrRecord.java
```diff
@@ -0,0 +1,117 @@
...
+        return record != null;
+    }
+
+    public DataStatistics statistics() {
+        return statistics;
+    }
+
+    public void statistics(DataStatistics newStatistics) {
```
wuchong: rename to `setStatistics`

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/sink/shuffle/StatisticsOrRecord.java
```diff
@@ -0,0 +1,117 @@
...
+        this.statistics = newStatistics;
+    }
+
+    public InputT record() {
+        return record;
+    }
+
+    public void record(InputT newRecord) {
```
wuchong: rename to `setRecord`.

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/sink/shuffle/WeightedRandomAssignment.java
```diff
@@ -0,0 +1,151 @@
...
+            int index = Arrays.binarySearch(cumulativeWeights, randomNumber);
+            // choose the subtask where randomNumber < cumulativeWeights[pos].
+            // this works regardless whether index is negative or not.
+            int position = Math.abs(index + 1);
+            checkState(
+                    position < assignedSubtasks.size(),
+                    "Invalid selected position: out of range. key weight = %s, random number = %s, cumulative weights array = %s",
+                    keyWeight,
+                    randomNumber,
+                    Arrays.toString(cumulativeWeights));
```
wuchong: Use `if` condition, otherwise, the `Arrays.toString(cumulativeWeights)` is always executed.

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/sink/FlinkSink.java
```diff
@@ -123,20 +134,76 @@ public AppendSinkWriter<InputT> createWriter(MailboxExecutor mailboxExecutor) {
...
+                                        toFlussRowType(tableRowType),
+                                        bucketKeys,
+                                        partitionKeys,
+                                        lakeFormat,
+                                        numBucket,
+                                        flussSerializationSchema),
+                                input.getParallelism());
+                    }
+                    return input;
+                case NONE:
```
wuchong, line 153: Move `NONE` to first

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/sink/FlinkSink.java
```diff
@@ -123,20 +134,76 @@ public AppendSinkWriter<InputT> createWriter(MailboxExecutor mailboxExecutor) {
...
+                            new StatisticsOrRecordTypeInformation<>(rowTypeInformation);
+                    SingleOutputStreamOperator<StatisticsOrRecord<InputT>> shuffleStream =
+                            input.transform(
+                                            "Dynamic shuffle data statistics",
+                                            statisticsOrRecordTypeInformation,
+                                            new DataStatisticsOperatorFactory<>(
+                                                    toFlussRowType(tableRowType),
+                                                    partitionKeys,
+                                                    flussSerializationSchema))
+                                    .uid("Dynamic shuffle data statistics" + tablePath)
```
wuchong:
```suggestion
                                    .uid("Dynamic shuffle data statistics " + tablePath)
```

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/sink/FlinkSink.java
```diff
@@ -123,20 +134,76 @@ public AppendSinkWriter<InputT> createWriter(MailboxExecutor mailboxExecutor) {
...
+                                            flussSerializationSchema),
+                                    input.getParallelism())
+                            .flatMap(
+                                    (FlatMapFunction<StatisticsOrRecord<InputT>, InputT>)
+                                            (statisticsOrRecord, out) -> {
+                                                if (statisticsOrRecord.hasRecord()) {
+                                                    out.collect(statisticsOrRecord.record());
+                                                }
+                                            })
+                            .uid("flat map" + tablePath)
```
wuchong:
```suggestion
                            .uid("flat map " + tablePath)
```

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/sink/FlinkSink.java
```diff
@@ -204,8 +272,13 @@ public DataStream<InputT> addPreWriteTopology(DataStream<InputT> input) {
                                     lakeFormat,
                                     numBucket,
                                     flussSerializationSchema),
-                            input.getParallelism())
-                    : input;
+                            input.getParallelism());
+                case NONE:
```
wuchong: move `NONE` to first

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/sink/FlinkTableSink.java
```diff
@@ -188,6 +194,22 @@ public SinkRuntimeProvider getSinkRuntimeProvider(Context context) {
...
+                @Override
+                public DataStreamSink<?> consumeDataStream(
+                        ProviderContext providerContext, DataStream<RowData> dataStream) {
+                    String sinkName =
+                            "sink"
+                                    + "-"
+                                    + tablePath.getTableName()
+                                    + "-"
+                                    + tablePath.getDatabaseName();
+                    return dataStream.sinkTo(flinkSink).uid(sinkName).name(sinkName);
```
wuchong: Do we really need to assign uid for the sink operator? We don't have state in sink for now.

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/sink/FlinkTableSink.java
```diff
@@ -79,7 +85,7 @@ public class FlinkTableSink
     private final DeleteBehavior tableDeleteBehavior;
     private final int numBucket;
     private final List<String> bucketKeys;
-    private final boolean shuffleByBucketId;
+    private DistributionMode distributionMode;
```
wuchong: `final`

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/sink/FlinkSink.java
```diff
@@ -123,20 +134,76 @@ public AppendSinkWriter<InputT> createWriter(MailboxExecutor mailboxExecutor) {
...
+                        throw new UnsupportedOperationException(
+                                "DYNAMIC_SHUFFLE is only supported for partition tables");
+                    }
+
+                    if (rowTypeInformation == null) {
+                        throw new UnsupportedOperationException(
+                                "RowTypeInformation is required for DYNAMIC_SHUFFLE mode.");
+                    }
+                    TypeInformation<StatisticsOrRecord<InputT>> statisticsOrRecordTypeInformation =
+                            new StatisticsOrRecordTypeInformation<>(rowTypeInformation);
```
wuchong: Use `new StatisticsOrRecordTypeInformation<>(input.getType());` so we don't need to introduce `rowTypeInformation`? And this will be safer.

fluss-flink/fluss-flink-common/src/test/java/org/apache/fluss/flink/sink/shuffle/DataStatisticsCoordinatorTest.java
```diff
@@ -0,0 +1,150 @@
...
+            Map<String, Long> partitionFrequencies = new HashMap<>();
+            partitionFrequencies.put("partition 1", 10000L);
+            partitionFrequencies.put("partition 2", 20000L);
+            coordinator.handleEventFromOperator(
+                    0,
+                    1,
+                    createStatisticsEvent(
+                            1, new DataStatistics(partitionFrequencies), dataStatisticsSerializer));
+
+            Thread.sleep(1000);
```
wuchong: Do we need to sleep here? It seems the following `coordinator.callInCoordinatorThread` has already waited for the pending events been processed.

fluss-flink/fluss-flink-common/src/test/java/org/apache/fluss/flink/sink/shuffle/DataStatisticsCoordinatorTest.java
```diff
@@ -0,0 +1,150 @@
...
+            Map<String, Long> expectedPartitionFrequencies = new HashMap<>();
+            expectedPartitionFrequencies.put("partition 1", 30000L);
+            expectedPartitionFrequencies.put("partition 2", 20000L);
+            StatisticsEvent expectedStatisticsEvent =
+                    createStatisticsEvent(
+                            1,
+                            new DataStatistics(expectedPartitionFrequencies),
+                            dataStatisticsSerializer);
+
+            coordinator.callInCoordinatorThread(
```
wuchong: This is a little hack that the introduced `callInCoordinatorThread` is just for testing, and the semantic is different from `runInCoordinatorThread` (`callInCoordinatorThread` will block until the callable is finished), but by name, they look similar. 

I suggest not introduce `callInCoordinatorThread`, and just use `runInCoordinatorThread` to execute a `CompletableFuture.complte(null)`, and wait the `CompletableFuture.get()` in test.

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/FlinkConnectorOptions.java
```diff
@@ -125,6 +126,23 @@ public class FlinkConnectorOptions {
                                     + BUCKET_KEY.key()
                                     + "' is defined. For Primary Key table, it is enabled by default.");
 
+    public static final ConfigOption<DistributionMode> SINK_DISTRIBUTION_MODE =
+            ConfigOptions.key("sink.distribution-mode")
+                    .enumType(DistributionMode.class)
+                    .defaultValue(DistributionMode.BUCKET_SHUFFLE)
```
wuchong: `sink.distribution-mode`

- Default: AUTO

- AUTO: Automatically chooses the best mode based on the table type. Uses BUCKET mode for Primary Key Tables and NONE for Log Tables to maximize throughput.

- NONE: Direct write without reshuffling.

- BUCKET: ..

- PARTITION_DYNAMIC: ...

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/FlinkConnectorOptions.java
```diff
@@ -125,6 +126,23 @@ public class FlinkConnectorOptions {
...
+                    .withDescription(
+                            "Defines the distribution mode for writing data to the sink. Available options are: \n"
+                                    + "- NONE: No specific distribution strategy. Data is forwarded as is.\n"
+                                    + "- BUCKET_SHUFFLE: Shuffle data by bucket ID before writing to sink. "
+                                    + "Shuffling the data with the same bucket ID to be processed by the same task "
+                                    + "can improve the efficiency of client processing and reduce resource consumption. "
```
wuchong: Mention this can improve throughput for primary key tables.

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/FlinkConnectorOptions.java
```diff
@@ -125,6 +126,23 @@ public class FlinkConnectorOptions {
...
+                                    + "For Log Table, bucket shuffle will only take effect when the '"
+                                    + BUCKET_KEY.key()
+                                    + "' is defined. For Primary Key table, it is enabled by default.\n"
+                                    + "- DYNAMIC_SHUFFLE: Dynamically adjust shuffle strategy based on partition key traffic patterns. "
+                                    + "This mode monitors data distribution and adjusts the shuffle behavior to balance the load. "
```
wuchong, line 163: Mention when DYNAMIC should be used, and what is the cost.

fluss-flink/fluss-flink-common/src/test/java/org/apache/fluss/flink/sink/FlinkTableSinkITCase.java
```diff
@@ -100,7 +102,10 @@ void before() {
         // open a catalog so that we can get table from the catalog
         String bootstrapServers = FLUSS_CLUSTER_EXTENSION.getBootstrapServers();
         // create table environment
-        env = StreamExecutionEnvironment.getExecutionEnvironment();
+        org.apache.flink.configuration.Configuration config =
+                new org.apache.flink.configuration.Configuration();
+        config.setString("taskmanager.numberOfTaskSlots", "3");
```
wuchong: Is this change necessary? Because the default slots is 4 defined in `AbstractTestBase`.

fluss-flink/fluss-flink-common/src/test/java/org/apache/fluss/flink/sink/FlinkTableSinkITCase.java
```diff
@@ -351,6 +363,59 @@ void testAppendLogWithMultiBatch() throws Exception {
...
+                        + "(8, 3508, 'stave'), "
+                        + "(9, 3509, 'Tim'), "
+                        + "(10, 3510, 'coco'), "
+                        + "(11, 3511, 'stave'), "
+                        + "(12, 3512, 'Tim')";
+        String insertPlan = tEnv.explainSql(insertSql, ExplainDetail.JSON_EXECUTION_PLAN);
+        if (distributionMode == DistributionMode.DYNAMIC_SHUFFLE) {
+            assertThat(insertPlan)
+                    .contains(String.format("\"ship_strategy\" : \"%s\"", distributionMode.name()));
+        } else {
```
wuchong, line 401: We should throw exception when it's BUCKET shuffle, as the table doesn't have bucket key. (after introduce AUTO default mode)

fluss-flink/fluss-flink-common/src/test/java/org/apache/fluss/flink/sink/FlinkTableSinkITCase.java
```diff
@@ -351,6 +363,59 @@ void testAppendLogWithMultiBatch() throws Exception {
...
+                        Arrays.asList("+I[8, 3508, stave]", "+I[11, 3511, stave]"),
+                        Arrays.asList("+I[10, 3510, coco]"));
+
+        List<String> expectedRows =
+                expectedGroups.stream().flatMap(List::stream).collect(Collectors.toList());
+
+        List<String> actual = collectRowsWithTimeout(rowIter, expectedRows.size());
+        assertThat(actual).containsExactlyInAnyOrderElementsOf(expectedRows);
+    }
+
```
wuchong, line 429: Please update the `testPut(boolean sinkBucketShuffle)` to use distributed mode.

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/sink/shuffle/DataStatistics.java
```diff
@@ -0,0 +1,75 @@
...
+    public boolean isEmpty() {
+        return partitionFrequency.isEmpty();
+    }
+
+    public void add(String partition, long value) {
+        if (partitionFrequency.containsKey(partition)) {
+            partitionFrequency.merge(partition, value, Long::sum);
+        } else {
+            // clone the sort key before adding to map because input sortKey object can be reused
+            partitionFrequency.put(partition, value);
```
wuchong: Can just `partitionFrequency.merge(partition, value, Long::sum);` for simplification and better performance.

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/sink/shuffle/DistributionMode.java
```diff
@@ -0,0 +1,49 @@
...
+    BUCKET_SHUFFLE("BUCKET_SHUFFLE"),
+    DYNAMIC_SHUFFLE("DYNAMIC_SHUFFLE");
+
+    private final String modeName;
+
+    DistributionMode(String modeName) {
+        this.modeName = modeName;
+    }
+
+    public String modeName() {
```
wuchong: we don't need the `modeName`, use `name()` directly.

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/sink/shuffle/DistributionMode.java
```diff
@@ -0,0 +1,49 @@
...
+        return modeName;
+    }
+
+    public static DistributionMode fromName(String modeName) {
+        checkArgument(null != modeName, "Invalid distribution mode: null");
+        try {
+            return DistributionMode.valueOf(modeName.toUpperCase(Locale.ENGLISH));
```
wuchong: nit: use `.toUpperCase()` is enough? I mean all the other places just use `.toUpperCase()`.

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/FlinkConnectorOptions.java
```diff
@@ -125,6 +126,23 @@ public class FlinkConnectorOptions {
                                     + BUCKET_KEY.key()
                                     + "' is defined. For Primary Key table, it is enabled by default.");
```
wuchong: Add deprecation annotation and comments to the config option

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/sink/serializer/FlussSerializationSchema.java
```diff
@@ -51,6 +51,10 @@ public interface FlussSerializationSchema<T> extends Serializable {
      */
     RowWithOp serialize(T value) throws Exception;
 
+    default long size(T value, RowType rowType) {
```
wuchong: This method is a little hack to me, and it's very confusing what's the relationship and order between this method and the `serialize` method. 

I suggest to add a `getEstimatedSizeInBytes` method to `RowWithOp` class, and add an additional construct that accepts `estimatedSizeInBytes` parameter. So we can combine the size estimation into the `serialize` method.

```
@Nuulable
public Long getEstimatedSizeInBytes() {
    ...
}
```

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/sink/serializer/RowDataSerializationSchema.java
```diff
@@ -159,6 +168,66 @@ public RowWithOp serialize(RowData value) throws Exception {
...
     }
 
+    @Override
+    public long size(RowData value, RowType rowType) {
+        if (value instanceof BinaryFormat) {
+            return ((BinaryFormat) value).getSizeInBytes();
+        }
+
+        long size = 0;
+        for (int i = 0; i < rowType.getFieldCount(); i++) {
```
wuchong: We should pre-compute fixed-length part size for better performance, and only need to compute var-length parts during runtime.

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/sink/serializer/RowDataSerializationSchema.java
```diff
@@ -159,6 +168,66 @@ public RowWithOp serialize(RowData value) throws Exception {
...
+            switch (typeRoot) {
+                case CHAR:
+                    size += ((CharType) (field.getType())).getLength();
+                    break;
+                case STRING:
+                    StringData stringData = value.getString(i);
+                    if (stringData instanceof BinaryStringData) {
+                        size += ((BinaryStringData) stringData).getSizeInBytes();
+                    } else {
+                        size += converter.getString(i).getSizeInBytes();
```
wuchong: Better to use `size += stringData.toBytes().length`. It's confusing what's the `converter` here, and what's the underlying row of `converter`.

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/sink/serializer/RowDataSerializationSchema.java
```diff
@@ -159,6 +168,66 @@ public RowWithOp serialize(RowData value) throws Exception {
...
+                    }
+                    break;
+                case BINARY:
+                    size += ((BinaryType) (field.getType())).getLength();
+                    break;
+                case BYTES:
+                    size += converter.getBytes(i).length;
```
wuchong: ditto

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/sink/shuffle/WeightedRandomAssignment.java
```diff
@@ -0,0 +1,151 @@
...
+                + ", subtaskWeights="
+                + subtaskWeights
+                + ", keyWeight="
+                + keyWeight
+                + ", cumulativeWeights="
+                + Arrays.toString(cumulativeWeights)
+                + '}';
+    }
+
+    protected double nextDouble(double origin, double bound) {
```
wuchong, line 149: Add comment to explain why we don't directly use `ThreadLocalThread.nextDouble(...)` here.

loserwang1024:
@wuchong  I have modified this PR ,CC

fluss-flink/fluss-flink-1.18/src/test/java/org/apache/fluss/flink/sink/Flink118TableSinkITCase.java
```diff
@@ -17,5 +17,25 @@
...
+        super.before();
+    }
+
+    @AfterEach
+    @Override
+    void after() throws Exception {
+        super.after();
+        // invoke here because the AbstractTestBase in 1.18 is junit 4.
+        AbstractTestBase.MINI_CLUSTER_RESOURCE.after();
+    }
```
wuchong: I don't understand this. Why we don't need this in other IT Cases? like `Flink118ComplexTypeITCase`, `Flink118TableSourceITCase`, etc. They all extends `AbstractTestBase`.

What's the problem if we don't add this?

wuchong: Is it because of the PARTITION_DYNAMIC jobs that require more task slots?

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/sink/serializer/RowDataSerializationSchema.java
```diff
@@ -155,8 +174,81 @@ public RowWithOp serialize(RowData value) throws Exception {
             row = outputProjection.replaceRow(row);
         }
         OperationType opType = toOperationType(value.getRowKind());
+        long estimatedSizeInBytes = calculateSize(value);
+        return new RowWithOp(row, opType, estimatedSizeInBytes);
```
wuchong: Currently, the `RowDataSerializationSchema` forces to calculate bytes in size for each record. However, only `DataStatisticsOperator` needs this. So this introduces perfromance regression.

I added a `isStatisticEnabled` flag on `FlussSerializationSchema#InitializationContext` to allow the `RowDataSerializationSchema` skip collect statistics for most cases.

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/sink/shuffle/DataStatisticsCoordinator.java
```diff
@@ -0,0 +1,359 @@
...
+                "Broadcast latest global statistics from checkpoint {} to all subtasks",
+                checkpointId);
+        // applyImmediately is set to false so that operator subtasks can
+        // apply the change at checkpoint boundary
+        StatisticsEvent statisticsEvent =
+                StatisticsEvent.createStatisticsEvent(
+                        checkpointId, statistics, statisticsSerializer);
+        for (int i = 0; i < context.currentParallelism(); ++i) {
+            try {
+                subtaskGateways.getSubtaskGateway(i).sendEvent(statisticsEvent).get();
```
wuchong: This is an blocking IO and will introduce many problems (e.g., blocks checkpoint) when task number very large. Why not just call `sendEvent` and leave it as is? We don't need to make sure the RPC success as you mentioned in the following exception catch warning.

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/sink/FlinkSink.java
```diff
@@ -62,9 +70,8 @@ protected SinkWriter<InputT> createWriter(
         return flinkSinkWriter;
     }
 
-    @Override
-    public DataStream<InputT> addPreWriteTopology(DataStream<InputT> input) {
-        return builder.addPreWriteTopology(input);
+    public DataStreamSink<InputT> addPreWriteTopology(DataStream<InputT> input) {
+        return builder.addPreWriteTopology(input).sinkTo(this);
```
wuchong: add `setParallelism` and use input parallelism.

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/sink/FlinkSink.java
```diff
@@ -123,22 +130,80 @@ public AppendSinkWriter<InputT> createWriter(MailboxExecutor mailboxExecutor) {
...
+                                            partitionKeys,
+                                            numBucket,
+                                            lakeFormat,
+                                            flussSerializationSchema),
+                                    input.getParallelism())
+                            .flatMap(
+                                    (FlatMapFunction<StatisticsOrRecord<InputT>, InputT>)
+                                            (statisticsOrRecord, out) -> {
+                                                if (statisticsOrRecord.isRecord()) {
+                                                    out.collect(statisticsOrRecord.record());
```
wuchong, line 198: Add operator name and add `setParallelism` and use input parallelism.

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/sink/FlinkSink.java
```diff
@@ -123,22 +130,80 @@ public AppendSinkWriter<InputT> createWriter(MailboxExecutor mailboxExecutor) {
...
+                                                }
+                                            })
+                            // To promote operator chaining with the downstream writer operator,
+                            // setting slot sharing group and the parallelism as default, {@link
+                            // SinkTransformationTranslator} will set the parallelism same as sink
+                            // transformation.
+                            .slotSharingGroup("shuffle-partition-custom-group")
```
wuchong: I’m not sure why `slotSharingGroup` is being set here. It doesn’t affect operator chaining.

In fact, I suspect this `slotSharingGroup` configuration may be the root cause of scheduling failures in PARTITION_DYNAMIC jobs due to insufficient resources. When all operators share the same `slotSharingGroup`, Flink can (and will) pack them into the same TaskManager slot. However, if different operators are assigned different slot sharing groups, they must be placed in separate slots, thereby increasing total slot demand.

When I removed this configuration, the PARTITION_DYNAMIC tests can be passed without `AbstractTestBase.MINI_CLUSTER_RESOURCE.before()` changes.

fluss-flink/fluss-flink-common/src/test/java/org/apache/fluss/flink/metrics/FlinkMetricsITCase.java
```diff
@@ -165,6 +166,7 @@ void testMetricsReport() throws Exception {
         createTable(tablePath, tableDescriptor);
 
         // test write
+        tEnv.getConfig().set(ExecutionConfigOptions.TABLE_EXEC_RESOURCE_DEFAULT_PARALLELISM, 1);
```
wuchong: Why setting paralleslims to `1`? This change doesn’t appear to be related to the main goal of the PR.

To help reviewers understand the intent and avoid unnecessary back-and-forth, it would be great to add a brief comment explaining the reason, or consider removing it if it’s not essential. This will help speed up the review and merge process.

fluss-flink/fluss-flink-common/src/test/java/org/apache/fluss/flink/sink/FlinkTableSinkITCase.java
```diff
@@ -261,22 +272,25 @@ void testAppendLogWithBucketKey(boolean sinkBucketShuffle) throws Exception {
...
+                    assertThat(index).isGreaterThan(prevIndex);
+                    prevIndex = index;
+                }
             }
         }
     }
 
     @Test
     void testAppendLogWithRoundRobin() throws Exception {
+        tEnv.getConfig().set(ExecutionConfigOptions.TABLE_EXEC_RESOURCE_DEFAULT_PARALLELISM, 1);
```
wuchong: Why setting paralleslims to 1?

fluss-flink/fluss-flink-common/src/test/java/org/apache/fluss/flink/sink/FlinkTableSinkITCase.java
```diff
@@ -544,12 +625,13 @@ void testPartialUpsertDuringAddColumn() throws Exception {
 
     @Test
     void testFirstRowMergeEngine() throws Exception {
+        tEnv.getConfig().set(ExecutionConfigOptions.TABLE_EXEC_RESOURCE_DEFAULT_PARALLELISM, 1);
```
wuchong: Why setting paralleslims to 1?

wuchong (approved):
@loserwang1024 thanks for the updating. I left some comments and pushed a commit to improve the implementation according to the comments. 

Please take a look at the changes and whether there is any concerns.

fluss-flink/fluss-flink-1.18/src/test/java/org/apache/fluss/flink/sink/Flink118TableSinkITCase.java
loserwang1024: @wuchong yes, it seems at least three slots are required. However, if not run  AbstractTestBase.MINI_CLUSTER_RESOURCE.before(), won't have 4 slots for each tm.

loserwang1024:
@wuchong , thanks for your improvement, LGTM

wuchong:
Merging...

PR #1802, opened 2025-10-13
[hotfix] Fix docker file build exception

luoyuxia:
### Purpose

Fix the exception when build with docker image:
```
Step 3/12 : COPY --chown=fluss:fluss build-target/ /opt/fluss/
unable to convert uid/gid chown string to host mapping: can't find uid for user fluss: no such user: fluss

```

**NOTE: Seems It only happens with `colima`, the offical `docker` still work**

### Brief change log

Create the user firstly.


morazow:
@luoyuxia this seems good to merge, is there anything else to do?

PR #1826, opened 2025-10-17
[lake/paimon] Prohibit users from setting Paimon properties that Fluss depends on

### LINKED ISSUE #1824: [lake] The default configuration of the Lake tables that Fluss depends on can be tampered with
### Search before asking

### Fluss version

main (development)

### Please describe the bug 🐞

When enable a lake (such as Paimon) table for a Fluss table, Fluss will set some default configs for the corresponding Paimon table, such as `bucket = -1` for log table without bucket keys. These configs suppose to be set by Fluss while not by users. If the user sets these properties, it will cause problems with the lakehouse synchronization.

For examle:
```java
    @Test
    void test() throws Exception {
        Map<String, String> customProperties = new HashMap<>();
        customProperties.put("k1", "v1");
        customProperties.put("paimon.file.format", "parquet");
        customProperties.put("paimon.bucket", "112");
        customProperties.put("paimon.bucket-key", "log_c1");

        TableDescriptor logTable =
                TableDescriptor.builder()
                        .schema(
                                Schema.newBuilder()
                                        .column("log_c1", DataTypes.INT())
                                        .column("log_c2", DataTypes.STRING())
                                        .build())
                        .property(ConfigOptions.TABLE_DATALAKE_ENABLED, true)
                        .customProperties(customProperties)
                        .distributedBy(3, "log_c1", "log_c2")
                        .build();
        TablePath logTablePath = TablePath.of(DATABASE, "log_table");
        admin.createTable(logTablePath, logTable, false).get();
        Table paimonLogTable =
                paimonCatalog.getTable(Identifier.create(DATABASE, logTablePath.getTableName()));

        System.out.println(paimonLogTable.options());
    }
```

We expect to get a Paimon table with the same bucket count as Fluss table. But we actually got a Paimon table with 112 buckets.
```
{bucket=112, path=..., fluss.table.datalake.enabled=true, fluss.table.datalake.format=paimon, partition.legacy-name=false, bucket-key=log_c1, file.format=parquet, fluss.k1=v1}

```

### Solution

We need to prevent users from setting these configs. 

LiebingYu:
### Purpose

Linked issue: close #1824

### DIFF
--- a/fluss-lake/fluss-lake-paimon/src/main/java/org/apache/fluss/lake/paimon/utils/PaimonConversions.java
+++ b/fluss-lake/fluss-lake-paimon/src/main/java/org/apache/fluss/lake/paimon/utils/PaimonConversions.java
@@ -57,16 +57,10 @@ public class PaimonConversions {
     /** Paimon config options set by Fluss should not be set by users. */
     @VisibleForTesting public static final Set<String> PAIMON_UNSETTABLE_OPTIONS = new HashSet<>();
 
-    @VisibleForTesting public static final Options PAIMON_DEFAULT_OPTIONS = new Options();
-
     static {
         PAIMON_UNSETTABLE_OPTIONS.add(CoreOptions.BUCKET.key());
         PAIMON_UNSETTABLE_OPTIONS.add(CoreOptions.BUCKET_KEY.key());
-        PAIMON_UNSETTABLE_OPTIONS.add(CoreOptions.CHANGELOG_PRODUCER.key());
-
-        // set partition.legacy-name to false, otherwise paimon will use toString for all types,
-        // which will cause inconsistent partition value for a same binary value
-        PAIMON_DEFAULT_OPTIONS.set(CoreOptions.PARTITION_GENERATE_LEGCY_NAME, false);
+        PAIMON_UNSETTABLE_OPTIONS.add(CoreOptions.PARTITION_GENERATE_LEGCY_NAME.key());
     }
 
     public static RowKind toRowKind(ChangeType changeType) {
@@ -120,13 +114,13 @@ public class PaimonConversions {
                 TableChange.SetOption setOption = (TableChange.SetOption) tableChange;
                 schemaChanges.add(
                         SchemaChange.setOption(
-                                getFlussPropertyKeyToPaimon(setOption.getKey()),
+                                convertFlussPropertyKeyToPaimon(setOption.getKey()),
                                 setOption.getValue()));
             } else if (tableChange instanceof TableChange.ResetOption) {
                 TableChange.ResetOption resetOption = (TableChange.ResetOption) tableChange;
                 schemaChanges.add(
                         SchemaChange.removeOption(
-                                getFlussPropertyKeyToPaimon(resetOption.getKey())));
+                                convertFlussPropertyKeyToPaimon(resetOption.getKey())));
             } else {
                 throw new UnsupportedOperationException(
                         "Unsupported table change: " + tableChange.getClass());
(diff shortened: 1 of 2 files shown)


fluss-lake/fluss-lake-paimon/src/main/java/org/apache/fluss/lake/paimon/utils/PaimonConversions.java
```diff
@@ -17,26 +17,58 @@
...
+
+    /** Paimon config options set by Fluss should not be set by users. */
+    @VisibleForTesting public static final Set<String> PAIMON_UNSETTABLE_OPTIONS = new HashSet<>();
+
+    @VisibleForTesting public static final Options PAIMON_DEFAULT_OPTIONS = new Options();
+
+    static {
+        PAIMON_UNSETTABLE_OPTIONS.add(CoreOptions.BUCKET.key());
+        PAIMON_UNSETTABLE_OPTIONS.add(CoreOptions.BUCKET_KEY.key());
+        PAIMON_UNSETTABLE_OPTIONS.add(CoreOptions.CHANGELOG_PRODUCER.key());
```
luoyuxia: I think we can make `CHANGELOG_PRODUCER` to be changed by users in case some users may don't want fluss generate change log? WDYT?

fluss-lake/fluss-lake-paimon/src/main/java/org/apache/fluss/lake/paimon/utils/PaimonConversions.java
```diff
@@ -17,26 +17,58 @@
...
 
+    // for fluss config
+    private static final String FLUSS_CONF_PREFIX = "fluss.";
+    // for paimon config
+    private static final String PAIMON_CONF_PREFIX = "paimon.";
+
+    /** Paimon config options set by Fluss should not be set by users. */
+    @VisibleForTesting public static final Set<String> PAIMON_UNSETTABLE_OPTIONS = new HashSet<>();
+
+    @VisibleForTesting public static final Options PAIMON_DEFAULT_OPTIONS = new Options();
```
luoyuxia: I suggest to remove PAIMON_DEFAULT_OPTIONS, put `PARTITION_GENERATE_LEGCY_NAME` into  `PAIMON_UNSETTABLE_OPTIONS`. 
And hard code 
```
options.set(CoreOptions.PARTITION_GENERATE_LEGCY_NAME, false);
```
direclty. Introduce two options make me hard to figure out what that mean.

fluss-lake/fluss-lake-paimon/src/main/java/org/apache/fluss/lake/paimon/utils/PaimonConversions.java
```diff
@@ -104,4 +135,111 @@ public static List<SchemaChange> toPaimonSchemaChanges(
...
+                        throw new InvalidConfigException(
+                                String.format(
+                                        "The Paimon option %s will be set automatically by Fluss "
+                                                + "and should not set manually.",
+                                        k));
+                    }
+                });
+    }
+
+    private static void setPaimonDefaultProperties(Options options) {
```
luoyuxia, line 219: nit:
`convertFlussPropertyKeyToPaimon`

luoyuxia (commented):
@LiebingYu Thanks for the pr. Left minor comments. PTAL

fluss-lake/fluss-lake-paimon/src/main/java/org/apache/fluss/lake/paimon/utils/PaimonConversions.java
LiebingYu: When do users need to change the `CHANGELOG_PRODUCER `? Fluss will generate the CDC logs and I think we should always use `INPUT` for Fluss lake table?

PR #1823, opened 2025-10-17
[hotfix] Make lake catalog as protected

luoyuxia:
### Purpose

Linked issue: close #xxx
Make lake catalog as projected in case someone would like to extend it

PR #1829, opened 2025-10-17
[hotfix][flink] Set lakeFlinkCatalog of FlinkCatalog as protected.

loserwang1024:
### Purpose

Set lakeFlinkCatalog of FlinkCatalog as protected so that user can extend it(and FlinkTableFactory).


loserwang1024:
close because https://github.com/apache/fluss/pull/1823

PR #1836, opened 2025-10-19
[build] Add support for jdk17 and jdk21 and add these jdks to CI

snuyanzin:
~~NOTE: 
**This PR is based on PR [[hotfix][tests] Fix unstable test IcebergRecordReaderTest#testReadLogTable](https://github.com/apache/fluss/pull/1837)**~~

~~I can rebase it once the first one is merged~~ 

### Purpose
Add support for jdk17, 21 and also add them to ci in order to be sure it is able to build and pass tests 

### Brief change log

1. it would make sense to bump both to the latest versions where jdk11 is still supported
1.25.0 - first googleJavaFormat where jdk11 was dropped (for more details https://github.com/google/google-java-format/releases?page=1)
2.xx.x - jdk11 is still supported 
3.0.0 is the first where jdk11 was dropped (for more details https://github.com/diffplug/spotless/releases?page=1)
2. Move spotless plugin to java11+ profile. The reason for that is that newer spotless requires java11 runtime, which means property like `skip.on.java8` will not help (not sure, may be completely remove this property?)
3. Adjust extra `add-exports`
4. Add jdks to ci

### Tests

existing + with newer jdks

### API and Format

no

### Documentation

no

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -55,3 +55,15 @@ jobs:
     uses: ./.github/workflows/ci-template.yaml
     with:
         java-version: "11"
+
+  build-on-jdk17:
+    name: "Java 17"
+    uses: ./.github/workflows/ci-template.yaml
+    with:
+      java-version: "17"
+
+  build-on-jdk21:
+    name: "Java 21"
+    uses: ./.github/workflows/ci-template.yaml
+    with:
+      java-version: "21"


.github/workflows/ci.yaml
```diff
@@ -54,4 +54,16 @@ jobs:
...
+    name: "Java 17"
+    uses: ./.github/workflows/ci-template.yaml
+    with:
+      java-version: "17"
+
+  build-on-jdk21:
+    name: "Java 21"
+    uses: ./.github/workflows/ci-template.yaml
+    with:
+      java-version: "21"
```
snuyanzin, line 69: or test only on one jdk, and only build for others?
Tests for every jdk may be move to nightly?

snuyanzin:
fyi: @wuchong

PR #1877, opened 2025-10-28
Numeric type widening converters

### LINKED ISSUE #1844: Relax conditions for converters
### Search before asking

### Motivation

Right now if POJO has type short or int and in schema was defined LONG it is failed.
However in most of SQL engines it could be handled with implicit CAST.
How about having same for some types if accuracy is able to be kept?

### Solution

_No response_

### Anything else?

_No response_

### Willingness to contribute

hemanthsavasere:
### Purpose

Linked issue: close #1844 

Perform automatic type widening between numeric types when converting POJOs to internal rows. Change from strict and exact type validation to widening conversion from lower to higher data types in POJO field's Java type and the schema's Fluss type.

Implementation Sample Scenarios :
POJO with int field + Schema with BIGINT → SUCCEEDS ✅
POJO with short field + Schema with BIGINT → SUCCEEDS ✅
POJO with int field + Schema with FLOAT → SUCCEEDS ✅
POJO with Long field + Schema with SMALLINT → FAILS (narrowing not supported) ❌

Widening Primitive Conversions :
byte (TINYINT) → short, int, long, float, double
short (SMALLINT) → int, long, float, double
int (INTEGER) → long, float, double
long (BIGINT) → float, double
float (FLOAT) → double

fluss-client

1. Expanded SUPPORTED_TYPES map in ConverterCommons.java to include all widening-compatible source types for each numeric target type
2. Added casting logic in PojoToRowConverter.java with dedicated conversion methods for each numeric type (convertToTinyInt, convertToSmallInt, convertToInteger, convertToBigInt, convertToFloat, convertToDouble)

fluss-flink-common

1. Updated SUPPORTED_TYPES map in PojoToRowConverter.java to include widening-compatible types (both primitives and boxed)
2. Implemented casting logic in createConverterForField() method with corresponding conversion helper methods

### Tests

 Added comprehensive tests in PojoToRowConverterTest.java for all widening combinations in fluss-client
 Added comprehensive tests in PojoToRowConverterTest.java covering all widening scenarios in fluss-flink-common


hemanthsavasere:
@polyzos Can you please review the pull request. Thanks

website/docs/engine-flink/datastream.mdx
```diff
@@ -164,6 +164,83 @@ public class OrderDeserializationSchema implements FlussDeserializationSchema<Or
...
+##### Numeric Type Widening for POJO Converters
+
+When using `PojoToRowConverter` for serialization, the converter supports automatic numeric type widening following Java's safe widening rules. This allows you to use smaller numeric types in your POJOs even when the table schema specifies larger types.
+
+**Supported Widenings**:
+- `byte` → `short`, `int`, `long`, `float`, `double`
+- `short` → `int`, `long`, `float`, `double`
+- `int` → `long`, `float`, `double`
+- `long` → `float`, `double`
+- `float` → `double`
```
snuyanzin: What about `Float.POSITIVE_INFINITY`, `Float.NEGATIVE_INFINITY`, `Float.NaN`
same for double
also would be great to have test for such corner cases

snuyanzin: Also not clear what about BigInteger/BigDecimal?

fluss-client/src/main/java/org/apache/fluss/client/converter/ConverterCommons.java
```diff
@@ -45,12 +45,39 @@ final class ConverterCommons {
...
+                DataTypeRoot.DOUBLE,
+                setOf(
+                        Double.class,
+                        Float.class,
+                        Long.class,
+                        Integer.class,
+                        Short.class,
+                        Byte.class)); // all numerics can widen to double
+
+        // Non-numeric types remain unchanged
```
snuyanzin: can you elaborate why 
```java
map.put(DataTypeRoot.DECIMAL, setOf(BigDecimal.class));
```
is considered as non-numeric?

fluss-client/src/main/java/org/apache/fluss/client/converter/PojoToRowConverter.java
```diff
@@ -234,6 +242,135 @@ private static FieldToRow createFieldConverter(PojoType.Property prop, DataType
...
+        }
+        if (v instanceof Integer) {
+            return ((Integer) v).doubleValue(); // int -> double widening
+        }
+        if (v instanceof Short) {
+            return ((Short) v).doubleValue(); // short -> double widening
+        }
+        if (v instanceof Byte) {
+            return ((Byte) v).doubleValue(); // byte -> double widening
+        }
```
snuyanzin: I wonder this can be replaced with something like
```java
if (v instanceof Number) {
  return ((Number) v).doubleValue();
}
```
similar for other places

fluss-client/src/test/java/org/apache/fluss/client/converter/PojoToRowConverterTest.java
```diff
@@ -168,4 +168,257 @@ public void testUnsupportedSchemaFieldTypeThrows() {
                 .hasMessageContaining("MAP")
                 .hasMessageContaining("mapField");
     }
+
+    // ----------------------- Numeric Type Widening Tests -----------------------
+
+    @Test
+    public void testByteToSmallIntWidening() {
```
snuyanzin:
```suggestion
   void testByteToSmallIntWidening() {
```
nit: here and in other places no need for `public` in junit5

fluss-client/src/test/java/org/apache/fluss/client/converter/PojoToRowConverterTest.java
```diff
@@ -168,4 +168,257 @@ public void testUnsupportedSchemaFieldTypeThrows() {
...
+
+        PojoToRowConverter<ShortFieldPojo> converter =
+                PojoToRowConverter.of(ShortFieldPojo.class, table, projection);
+
+        ShortFieldPojo pojo = new ShortFieldPojo();
+        pojo.quantity = (short) 999;
+        GenericRow row = converter.toRow(pojo);
+
+        assertThat(row.getLong(0)).isEqualTo(999L);
+    }
```
snuyanzin, line 377: all these and other tests in this class are pretty similar
can we parameterize them?

polyzos:
@hemanthsavasere can you remove the documentation? This functionality is meant for internal use.
The goal is to update the existing Java client to use this functionality under the hood so it's transparent to the users.
So we don't need to document it, as it won't be exposed to users directly

polyzos:
@snuyanzin thank you for reviewing this and your thoughtful feedback

website/docs/engine-flink/datastream.mdx
hemanthsavasere: @snuyanzin 
No changes in BigDecimal handling, but current implementation has

BigDecimal Handling: 
Native Support: Fluss has full native support for DECIMAL(p, s)type (precision 1-38, scale 0-p)

Optimized Internal Storage:
Compact mode (precision ≤ 18): Stored as long for efficiency
Full mode (precision > 18): Stored as BigDecimal) with byte array representation
No Kryo or generic serialization overhead
POJO Conversion: Only java.math.BigDecimal is accepted for DECIMAL columns

No automatic widening from primitive numeric types (byte, short, int, long) to DECIMAL
Must explicitly use BigDecimal in your POJO classes

Utility Methods:  
DecimalUtils provides casting operations from numeric types, strings, and between decimal precisions

fluss-client/src/main/java/org/apache/fluss/client/converter/ConverterCommons.java
hemanthsavasere: @snuyanzin  sorry, It's a mistake from my side. It's basically for types that are unchanged and does not suport widening

website/docs/engine-flink/datastream.mdx
hemanthsavasere: BigInteger Handling:
No Native Support: BigInteger is NOT a first-class type in Fluss

Internal Use Only: BigInteger is used internally for:
Unscaled values in non-compact Decimal representation
Byte array to Decimal conversions

fluss-client/src/main/java/org/apache/fluss/client/converter/PojoToRowConverter.java
hemanthsavasere: Agreed, I have done the changes.

website/docs/engine-flink/datastream.mdx
hemanthsavasere: Agreed, I have added more unit tests for Float.POSITIVE_INFINITY, Float.NEGATIVE_INFINITY, Float.NaN and the 
same for double.

hemanthsavasere:
@snuyanzin and @polyzos I have done the changes, can you please review PR.

hemanthsavasere:
@snuyanzin @polyzos I have completed the requested changes. When you have the opportunity, could you please review the PR?

wuchong:
Thanks for the contribution, @hemanthsavasere!

I appreciate the intent behind this change, but I have some concerns about introducing type widening in POJO ↔ Row conversion.

While it’s true that widening conversions (e.g., int → long) are safe in one direction—since they don’t lose precision—the challenge here is that POJO-Row conversion is bidirectional: it’s used both on the sink side (POJO → Row) and the source side (Row → POJO).

For example, suppose a table column is defined as `BIGINT`. If we allow a POJO to use `int` (relying on widening during write), that same POJO would fail or silently lose data when used for reading—because converting the stored `long` value back to `int` is a narrowing conversion that can cause overflow or precision loss.

To ensure correctness and consistency across both source and sink, I think we have to strict type alignment between the POJO fields and the table schema. This is also consistent with how Avro SpecificRecord works—it mandates exact type matching with the Avro schema to avoid ambiguity and runtime errors.

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/utils/PojoToRowConverter.java
```diff
@@ -77,12 +77,62 @@ public class PojoToRowConverter<T> {
...
+                        Float.class,
+                        float.class,
+                        Long.class,
+                        long.class,
+                        Integer.class,
+                        int.class,
+                        Short.class,
+                        short.class,
+                        Byte.class,
+                        byte.class));
```
gyang94, line 133: There are many duplicated code and definition for this kind of “type hierarchy”. We can simplify it with "incremental composition" or other ways. Here is a possible sample I am thinking of:
```
//  1. basic class type definination
Set<Class<?>> byteClasses = orderedSet(Byte.class, byte.class);
Set<Class<?>> shortClasses = orderedSet(Short.class, short.class);
Set<Class<?>> intClasses = orderedSet(Integer.class, int.class);
Set<Class<?>> longClasses = orderedSet(Long.class, long.class);
Set<Class<?>> floatClasses = orderedSet(Float.class, float.class);
Set<Class<?>> doubleClasses = orderedSet(Double.class, double.class);

// 2. combine them hierarchically.
SUPPORTED_TYPES.put(DataTypeRoot.SMALLINT, 
    combine(shortClasses, byteClasses));

SUPPORTED_TYPES.put(DataTypeRoot.INTEGER, 
    combine(intClasses, SUPPORTED_TYPES.get(DataTypeRoot.SMALLINT)));

SUPPORTED_TYPES.put(DataTypeRoot.BIGINT, 
    combine(longClasses, SUPPORTED_TYPES.get(DataTypeRoot.INTEGER)));

SUPPORTED_TYPES.put(DataTypeRoot.FLOAT, 
    combine(floatClasses, SUPPORTED_TYPES.get(DataTypeRoot.BIGINT)));

SUPPORTED_TYPES.put(DataTypeRoot.DOUBLE, 
    combine(doubleClasses, SUPPORTED_TYPES.get(DataTypeRoot.FLOAT)));

// 3. combine function
private Set<Class<?>> combine(Set<Class<?>> primary, Set<Class<?>> inherited) {
    Set<Class<?>> result = new LinkedHashSet<>(primary);
    result.addAll(inherited);
    return Collections.unmodifiableSet(result);
}
```

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/utils/PojoToRowConverter.java
```diff
@@ -335,6 +389,109 @@ public GenericRow convert(T pojo) {
...
+            return ((Short) value).longValue();
+        }
+        if (value instanceof Byte) {
+            return ((Byte) value).longValue();
+        }
+        throw new IllegalArgumentException(
+                "Field " + field.getName() + " cannot be converted to BIGINT");
+    }
+
+    private static Object convertToFloat(Field field, Object value)
```
gyang94, line 459: Similar duplication for `if-else` in these series of `convertToXXX` functions. We can make a general `safeConvert` function.
```
private static <T> T safeConvert(DataTypeRoot targetRoot, Field field, Object value, Function<Number, T> converter) {
    if (value == null) {
        return null;
    }
    
    // 1. validation (with SUPPORT_TYPES)
    Set<Class<?>> supported = SUPPORTED_TYPES.getOrDefault(targetRoot, Collections.emptySet());
    if (!supported.contains(value.getClass())) {
            String.format("Field %s (type: %s) cannot be converted to %s", 
                field.getName(), value.getClass().getSimpleName(), targetRoot));
    }

    // 2. convert
    if (value instanceof Number) {
        return converter.apply((Number) value);
    }

    throw new IllegalArgumentException("Unexpected value type for field " + field.getName());
}
```
Then in those `convertToXX` functions
```
private static Object convertToTinyInt(Field field, Object value) {
    return safeConvert(DataTypeRoot.TINYINT, field, value, Number::byteValue);
}

private static Object convertToSmallInt(Field field, Object value) {
    return safeConvert(DataTypeRoot.SMALLINT, field, value, Number::shortValue);
}

...
```

what do you think?

PR #1880, opened 2025-10-28
[hotfix] Use literal directly  instead of PARTITION_GENERATE_LEGCY_NAME to avoid NoSuchField error

luoyuxia:
### Purpose

Linked issue: close #xxx
 Use literal directly  instead of PARTITION_GENERATE_LEGCY_NAME to avoid NoSuchField error when mix with a paimon with higher version which has changed `PARTITION_GENERATE_LEGCY_NAME` to `PARTITION_GENERATE_LEGACY_NAME`

### DIFF
--- a/fluss-lake/fluss-lake-paimon/src/main/java/org/apache/fluss/lake/paimon/utils/PaimonConversions.java
+++ b/fluss-lake/fluss-lake-paimon/src/main/java/org/apache/fluss/lake/paimon/utils/PaimonConversions.java
@@ -49,6 +49,13 @@ import static org.apache.fluss.lake.paimon.PaimonLakeCatalog.SYSTEM_COLUMNS;
 /** Utils for conversion between Paimon and Fluss. */
 public class PaimonConversions {
 
+    // use literal directly, a future paimon version will rename
+    // the variable PARTITION_GENERATE_LEGCY_NAME to PARTITION_GENERATE_LEGACY_NAME, use literal
+    // can help avoid NoSuchField error
+    // todo: after upgrade paimon version, we call fall back to use PARTITION_GENERATE_LEGACY_NAME
+    // again
+    private static final String PARTITION_GENERATE_LEGACY_NAME_OPTION_KEY = "partition.legacy-name";
+
     // for fluss config
     private static final String FLUSS_CONF_PREFIX = "fluss.";
     // for paimon config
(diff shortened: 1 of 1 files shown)


LiebingYu:
+1

fluss-lake/fluss-lake-paimon/src/main/java/org/apache/fluss/lake/paimon/utils/PaimonConversions.java
```diff
@@ -60,7 +60,10 @@ public class PaimonConversions {
     static {
         PAIMON_UNSETTABLE_OPTIONS.add(CoreOptions.BUCKET.key());
         PAIMON_UNSETTABLE_OPTIONS.add(CoreOptions.BUCKET_KEY.key());
-        PAIMON_UNSETTABLE_OPTIONS.add(CoreOptions.PARTITION_GENERATE_LEGCY_NAME.key());
+        // use literal directly, a future paimon version will rename
+        // the variable PARTITION_GENERATE_LEGCY_NAME to PARTITION_GENERATE_LEGACY_NAME, use literal
+        // can help avoid NoSuchField error
+        PAIMON_UNSETTABLE_OPTIONS.add("partition.legacy-name");
```
loserwang1024: also modified `PARTITION_GENERATE_LEGCY_NAME`  in `setPaimonDefaultProperties`

leonardBang: +1， looks like paimon has a typo  LEGCY vs LEGACY

luoyuxia: I still don't want to modified. Keep `PARTITION_GENERATE_LEGCY_NAME` can help us find problem quickly(compile will fail) if Paimon remove `PARTITION_GENERATE_LEGCY_NAME` since we does rely on paimon's behavior of `PARTITION_GENERATE_LEGCY_NAME`. 

Since it's a tempory hacky fix since it's a typo of paimon, after we ungrade paimon to higer version, we can fall back  to use `PARTITION_GENERATE_LEGACY_NAME ` again, which I believe won't be changed in paimon

leonardBang: LOL, I just pointed out a typo in paimon and we should fix it in paimon project in a compatible way instead of using the internal key here.

luoyuxia: Let's change all to literal to make ervery thing consistent. If paimon remove this optioin, the existing test should discover it.

leonardBang (approved):

PR #1882, opened 2025-10-28
[hotfix][client] Use more efficient `EnumMap` in `ConverterCommons`

snuyanzin:
### Purpose

Since there is a map with enum elements as key then it would be more efficient to use `EnumMap`, the PR does that
### Brief change log

HashMap -> EnumMap for one particular case

### Tests

existing


snuyanzin:
cc @polyzos

PR #1897, opened 2025-10-30
[hotfix][client] Fix potential metadata inconsistency by deduplicating getBucketLocation call

swuferhong:
### Purpose

This pr is aims to fix the `NoSuchElementException` throw at client when upgrading cluster:

image

### DIFF
--- a/fluss-client/src/main/java/org/apache/fluss/client/table/scanner/log/LogFetcher.java
+++ b/fluss-client/src/main/java/org/apache/fluss/client/table/scanner/log/LogFetcher.java
@@ -65,6 +65,7 @@ import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
+import java.util.Optional;
 import java.util.Set;
 import java.util.stream.Collectors;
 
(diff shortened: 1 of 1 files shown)


fluss-client/src/main/java/org/apache/fluss/client/table/scanner/log/LogFetcher.java
```diff
@@ -502,8 +503,9 @@ private List<TableBucket> fetchableBuckets() {
     }
 
     private Integer getTableBucketLeader(TableBucket tableBucket) {
-        if (metadataUpdater.getBucketLocation(tableBucket).isPresent()) {
-            BucketLocation bucketLocation = metadataUpdater.getBucketLocation(tableBucket).get();
+        Optional<BucketLocation> bucketLocationOpt = metadataUpdater.getBucketLocation(tableBucket);
```
leonardBang, line 506: IIUC, here we want to call `metadataUpdater.getBucketLocation` once to make sure the status consistence. Could you change the commit message to `[hotfix][client] Fix potential metadata inconsistency by deduplicating getBucketLocation call` ?

Btw, we can also search similar places in client.

swuferhong: Done! I have checked the callers of MetadataUpdater and confirmed that there are no other inconsistent invocations.

leonardBang (approved):
Thanks @swuferhong for the update and check, LGTM

Issue #1907, opened 2025-10-30
Currently it's impossible to use inner classes for converters, also usually instead of get `is`/`has` could be used for booleans


snuyanzin:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/fluss/issues) and found nothing similar.

### Motivation

Put in in task name

### Willingness to contribute

- [x] I'm willing to submit a PR!

snuyanzin:
closed as implemented

PR #1908, opened 2025-10-30
[client] Add support for inner class and is/has methods for boolean

snuyanzin:
### Purpose

Implements #1907

### Brief change log

PojoType

### Tests

PojoTypeTest

### DIFF
--- a/fluss-client/src/main/java/org/apache/fluss/client/converter/PojoType.java
+++ b/fluss-client/src/main/java/org/apache/fluss/client/converter/PojoType.java
@@ -85,13 +85,6 @@ final class PojoType<T> {
             if (!publicField) {
                 // When not a public field, require both getter and setter
                 if (getter == null || setter == null) {
-                    if (field.getName().startsWith("this$")) {
-                        final Class type = field.getType();
-                        if ((type.getName() + "$" + pojoClass.getSimpleName())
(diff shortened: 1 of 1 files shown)


snuyanzin:
fyi: @polyzos

fluss-client/src/main/java/org/apache/fluss/client/converter/PojoType.java
```diff
@@ -143,32 +159,68 @@ private static Map<String, Field> discoverAllInstanceFields(Class<?> clazz) {
...
+    private static Map<String, Method> discoverGetters(
+            Class<?> clazz, Map<String, Field> fieldMap) {
+        final Map<String, Method> getters = new HashMap<>();
         for (Method m : clazz.getMethods()) { // public methods incl. inherited
-            if (m.getParameterCount() == 0
-                    && m.getName().startsWith("get")
-                    && !m.getReturnType().equals(void.class)) {
-                String prop = decapitalize(m.getName().substring(3));
+            final String prop = getGetterProp(m);
+            if (fieldMap.containsKey(prop)) {
```
snuyanzin, line 167: no need to store in map any data not matching field names.

E.g. if for some reason a POJO has  a number of methods like `getSomethingBlaBlaBla` which is not related to fields

PR #1915, opened 2025-10-31
Fix thread-safety problem of PrimaryKeyLoookuper and PrefixKeyLookuper

### LINKED ISSUE #1914: [Client] lookup() function in PrimaryKeyLookuper and PrefixKeyLookuper is not thread safety
### Search before asking

### Fluss version

0.7.0 (latest release)

### Please describe the bug 🐞

Currently, both PrimaryKeyLookuper and PrefixKeyLookuper utilize KeyEncoder to encode primary keys and bucket keys. However, there is no guarantee that KeyEncoder is thread-safe, which makes the lookup() function itself not thread-safe, even though the lookup() function returns a CompletableFuture, indicating that it is an asynchronous function by its signature.

When user code invokes the lookup() function concurrently, some unexpected errors may occur as follows: 
`
java.lang.ArrayIndexOutOfBoundsException: arraycopy: last destination index 3 out of bounds for byte[2]
	at java.base/java.lang.System.arraycopy(Native Method)
	at org.apache.fluss.row.compacted.CompactedRowWriter.toBytes(CompactedRowWriter.java:108)
	at org.apache.fluss.row.encode.CompactedKeyEncoder.encodeKey(CompactedKeyEncoder.java:84)
	at org.apache.fluss.client.lookup.PrimaryKeyLookuper.lookup(PrimaryKeyLookuper.java:107)
	at org.apache.fluss.flink.source.lookup.FlinkAsyncLookupFunction.fetchResult(FlinkAsyncLookupFunction.java:143)
	at org.apache.fluss.flink.source.lookup.FlinkAsyncLookupFunction.asyncLookup(FlinkAsyncLookupFunction.java:126)
	at org.apache.flink.table.runtime.functions.table.lookup.CachingAsyncLookupFunction.asyncLookup(CachingAsyncLookupFunction.java:92)
	at org.apache.flink.table.functions.AsyncLookupFunction.eval(AsyncLookupFunction.java:52)
`

### Solution

Add synchronized block for PrimaryKeyLookuper and PrefixKeyLookuper's lookup() to ensure thread-safety

platinumhamburg:
### Purpose

Linked issue: close #1914


fluss-client/src/main/java/org/apache/fluss/client/lookup/PrefixKeyLookuper.java
```diff
@@ -146,7 +146,7 @@ private void validatePrefixLookup(TableInfo tableInfo, List<String> lookupColumn
     }
 
     @Override
-    public CompletableFuture<LookupResult> lookup(InternalRow prefixKey) {
+    public synchronized CompletableFuture<LookupResult> lookup(InternalRow prefixKey) {
```
beryllw: Using `synchronized` on the entire `lookup` method might affect client-side `lookup` efficiency. May be we could narrow the lock scope to only cover thread-unsafe components instead?

xx789633: Hi @beryllw I'll try to optimize that.

luoyuxia: @xx789633 What's your solution to  optimize it?  IIUC, Flink lookup join operator should call lookup synchronously.. The mutiple thread calling should happen when retry happen which is call in a callback.

xx789633: Hi @luoyuxia @platinumhamburg @beryllw some members in lookuper are stateful, so I agree that there should be some lock protection for those members.

But what I'm more curious about is: in what scenarios would Flink perform concurrent lookups using the same lookuper instance? Are there any more detailed logs available for that particular bug?

xx789633: Hi @luoyuxia , I don't think the retry logic would cause concurrent lookups because the result future defined in `asyncLookup` will only be completed when the retry fails (with `resultFuture.completeExceptionally`) or successes (`resultFuture.complete`). In either case, Flink will wait for the result future to finish before its own retry logic.

swuferhong: > Hi @luoyuxia @platinumhamburg @beryllw some members in lookuper are stateful, so I agree that there should be some lock protection for those members.
> 
> But what I'm more curious about is: in what scenarios would Flink perform concurrent lookups using the same lookuper instance? Are there any more detailed logs available for that particular bug?

Hi, @xx789633. This case occurs during the `tabletServer` restart process (cluster upgrade). Currently, what I have observed is that it can be stably reproduced when restarting the `Flink lookup job` during the upgrade. I have tried to reproduce it in other scenarios, but without success.

wuchong (approved):
Thanks @platinumhamburg , this pull request is in high code quality with strong test coverage. I only improved some minor problems. I will merge it once CI is passed.

PR #1920, opened 2025-10-31
[client][hotfix] Invoke `RowType#getFieldNames` only once in `ConverterCommons`

snuyanzin:
### Purpose
It is a small optimization to avoid calling `RowType#getFieldNames` inside a loop since each time it should through fields and create a new collection of names

### DIFF
--- a/fluss-client/src/main/java/org/apache/fluss/client/converter/ConverterCommons.java
+++ b/fluss-client/src/main/java/org/apache/fluss/client/converter/ConverterCommons.java
@@ -26,6 +26,7 @@ import java.math.BigDecimal;
 import java.util.Arrays;
 import java.util.EnumMap;
 import java.util.HashSet;
+import java.util.List;
 import java.util.Map;
 import java.util.Set;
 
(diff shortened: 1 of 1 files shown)


fluss-client/src/main/java/org/apache/fluss/client/converter/ConverterCommons.java
```diff
@@ -67,15 +68,15 @@ private static Map<DataTypeRoot, Set<Class<?>>> createSupportedTypes() {
...
+        List<String> fieldNames = tableSchema.getFieldNames();
+        if (!pojoNames.containsAll(fieldNames)) {
             throw new IllegalArgumentException(
                     String.format(
                             "POJO fields %s must exactly match table schema fields %s.",
-                            pojoNames, tableNames));
+                            pojoNames, fieldNames));
         }
         for (int i = 0; i < tableSchema.getFieldCount(); i++) {
-            String name = tableSchema.getFieldNames().get(i);
```
snuyanzin, line 78: Avoid calling it inside a loop since each call generates a new collection
https://github.com/apache/fluss/blob/84b06ac6ef18ea629106e77755bee4b1963edd55/fluss-common/src/main/java/org/apache/fluss/types/RowType.java#L66-L68

snuyanzin:
fyi @polyzos

PR #1922, opened 2025-11-02
[release] Fix missing LICENSE and DISCLAIMER in binary distribution

wuchong:
### Purpose

This fixes the release issues mentioned by JB in the vote thread: https://lists.apache.org/thread/pyccbmxoz9wnf6m3blcrcvj8j9gzyow8

> +1 (binding)
> 
> NB: the binary distribution should be fixed (see later).
> 
> I checked:
> - Source distribution
> -- incubating is in the name/version
> -- signature and checksum are good
> -- DISCLAIMER is there
> -- LICENSE and NOTICE are good (NB: it would be good to mention Maven
> in LICENSE/NOTICE as we bundle mvnw in the source distribution)
> -- No binary file found in the source distribution
> -- ASF header is present in expected files and good (just missing in a
> couple of README.md, but not blocker/required)
> - Binary distribution
> -- incubating is in the name/version
> -- signature and checksum are good
> -- DISCLAIMER is missing in the source distribution
> -- LICENSE is missing in the root directory (for Fluss itself)
> -- LICENSE files in the licenses directory and NOTICE file are good
> -- ASF header is present in configuration file (in the conf folder)
> 
> Thanks,
> Regards
> JB

Linked issue: close #xxx

### Brief change log

- fix the `create_binary_release.sh` to copy the DISCLAIMER and LICENSE from the project root directory.
- Add ASF headers to README.md files
- Add mvnw files entries in LICENSE and NOTICE

### DIFF
--- a/tools/releasing/create_binary_release.sh
+++ b/tools/releasing/create_binary_release.sh
@@ -65,7 +65,7 @@ make_binary_release() {
   dir_name="fluss-$RELEASE_VERSION-bin"
 
   # enable release profile here (to check for the maven version)
-  # $MVN clean package -Prelease -am -Dgpg.skip -Dcheckstyle.skip=true -DskipTests
+  $MVN clean package -Prelease -am -Dgpg.skip -Dcheckstyle.skip=true -DskipTests
 
   cd fluss-dist/target/fluss-${RELEASE_VERSION}-bin
   cp ${FLUSS_DIR}/LICENSE ./fluss-${RELEASE_VERSION}/LICENSE


tools/releasing/create_binary_release.sh
```diff
@@ -68,7 +68,11 @@ make_binary_release() {
   $MVN clean package -Prelease -am -Dgpg.skip -Dcheckstyle.skip=true -DskipTests
 
   cd fluss-dist/target/fluss-${RELEASE_VERSION}-bin
+  cp ${FLUSS_DIR}/LICENSE ./fluss-${RELEASE_VERSION}/LICENSE
```
jbonofre, line 71: That's the `LICENSE` from the source distribution right ? 

I guess we need a specific one for the binary distribution (as we bundle dependencies).

Maybe we can do another PR specifically for `LICENSE` and `NOTICE` in the binary distribution.

wuchong: Yes, this copies `LICENSE` file from source code to the binary distribution to fix the missing `LICENSE` in binary distribution. I think the `LICENSE` file in source distribution is correct, right? 

The `NOTICE` file already exists in binary distribution, this PR just append the `NOTICE` content in source code into the existing binary `NOTICE` file.

wuchong:
Hi @jbonofre,

I’ve tested the PR locally, and it correctly generates the LICENSE and DISCLAIMER files in the binary distribution.

I’ll go ahead and merge this for now. If anything is missing or needs refinement, I’ll follow up with additional PRs. 

image

jbonofre:
Yeah I will create another PR because the LICENSE is correct for source distribution but not binary distribution

PR #1928, opened 2025-11-04
[Client] Fix LookupSender doesn't update bucket metadata when receive NotLeaderOrFollowerException in bucket level

### LINKED ISSUE #1927: LookupSender doesn't update bucket metadata when receive NotLeaderOrFollowerException in bucket level
### Search before asking

### Fluss version

main (development)

### Please describe the bug 🐞

LookupSender doesn't update bucket metadata when receive NotLeaderOrFollowerException in bucket level

### Solution

_No response_

swuferhong:
### Purpose

Linked issue: close #1927 

This pr is aims to fix `LookupSender` doesn't update bucket metadata when receive NotLeaderOrFollowerException in bucket level

### DIFF
--- a/fluss-client/src/main/java/org/apache/fluss/client/lookup/LookupSender.java
+++ b/fluss-client/src/main/java/org/apache/fluss/client/lookup/LookupSender.java
@@ -18,10 +18,15 @@
 package org.apache.fluss.client.lookup;
 
 import org.apache.fluss.annotation.Internal;
+import org.apache.fluss.annotation.VisibleForTesting;
 import org.apache.fluss.client.metadata.MetadataUpdater;
+import org.apache.fluss.exception.ApiException;
 import org.apache.fluss.exception.FlussRuntimeException;
+import org.apache.fluss.exception.InvalidMetadataException;
 import org.apache.fluss.exception.LeaderNotAvailableException;
+import org.apache.fluss.metadata.PhysicalTablePath;
 import org.apache.fluss.metadata.TableBucket;
+import org.apache.fluss.metadata.TablePartition;
 import org.apache.fluss.rpc.gateway.TabletServerGateway;
 import org.apache.fluss.rpc.messages.LookupRequest;
 import org.apache.fluss.rpc.messages.LookupResponse;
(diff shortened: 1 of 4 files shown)


fluss-client/src/main/java/org/apache/fluss/client/lookup/LookupSender.java
```diff
@@ -382,4 +395,47 @@ void initiateClose() {
...
+        private final @Nullable Set<Long> tableIds;
+        private final @Nullable Set<TablePartition> tablePartitions;
+
+        TableOrPartitions(
+                @Nullable Set<Long> tableIds, @Nullable Set<TablePartition> tablePartitions) {
+            this.tableIds = tableIds;
+            this.tablePartitions = tablePartitions;
+        }
+    }
+
```
loserwang1024, line 436: The code is same as LogFetcher, can we extract this same code in protected class? If some problem occurs, we only need to change once.

swuferhong: > The code is same as LogFetcher, can we extract this same code in protected class? If some problem occurs, we only need to change once.

@wuchong will refactor the `client metadata` code implement, so there is no need to uniform. It's sufficient to have test coverage.

luoyuxia (commented):
Seems "// todo: try to get value when get is re-triable in FLUSS-56857409" can be finished with this pr?

swuferhong:
> Seems "// todo: try to get value when get is re-triable in FLUSS-56857409" can be finished with this pr?

This PR may not be able to fix the issue. Only when the `LookupSender` introduces a retry mechanism can the lookup produce deterministic results. In the current implementation, the lookup will still throw errors, and retries must be handled at the upper layer (flink).

swuferhong:
> Seems "// todo: try to get value when get is re-triable in FLUSS-56857409" can be finished with this pr?

Hi, @platinumhamburg. Your PR #1915  has already addressed this issue. Could you please remove this `todo comment` in your PR and improve the tests accordingly?

platinumhamburg:
> > Seems "// todo: try to get value when get is re-triable in FLUSS-56857409" can be finished with this pr?
> 
> Hi, @platinumhamburg. Your PR #1915 has already addressed this issue. Could you please remove this `todo comment` in your PR and improve the tests accordingly?

I thought about it, and PR #1915 currently does not implement infinite retries nor support delayed backoff retries. In fact, even with limited retries, a single lookup does not guarantee the "deterministic" result expected here, so this TODO item should not be removed in PR #1915.

PR #1934, opened 2025-11-05
[flink] Always return all filters to flink when trying to apply filters in FlinkTableSource

### LINKED ISSUE #1933: When Flink attempts to push down filters to Fluss dim source, the Fluss dim source accepts some of the filters, but they do not take effect.
### Search before asking

### Fluss version

main (development)

### Please describe the bug 🐞

Currently, the Fluss dim table cannot resolve filters when being looked up, requiring Flink to handle filtering. However, when Flink attempts to push down filters to the Fluss dim source, it accepts some of the filters and returns the remaining back to Flink. As a result, some filters do not take effect.

### Solution

We cannot determine whether this source will ultimately be used as a scan source or a lookup source. Since fluss lookup sources cannot accept filters yet, to be safe, we'd better return all filters to the Flink planner.

xuyangzhong:
### Purpose

Linked issue: close #1933

### Brief change log

We cannot determine whether this source will ultimately be used as a scan source or a lookup source. Since fluss lookup sources cannot accept filters yet, to be safe, we must return all filters to the Flink planner.

### Tests

New tests are added both streaming and batch to verify this pr.

### API and Format

None.

### Documentation

None.

### DIFF
--- a/fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/source/FlinkTableSource.java
+++ b/fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/source/FlinkTableSource.java
@@ -521,9 +521,9 @@ public class FlinkTableSource
             }
             singleRowFilter = lookupRow;
 
-            // We cannot determine whether this source will ultimately be used as a scan source or
-            // a lookup source. Since fluss lookup sources cannot accept filters yet, to be safe, we
-            // return all filters to the Flink planner.
+            // FLINK-38635 We cannot determine whether this source will ultimately be used as a scan
+            // source or a lookup source. Since fluss lookup sources cannot accept filters yet, to
+            // be safe, we return all filters to the Flink planner.
             return Result.of(acceptedFilters, filters);
         } else if (isPartitioned()) {
             // apply partition filter pushdown
@@ -577,9 +577,9 @@ public class FlinkTableSource
                 }
             }
 
-            // We cannot determine whether this source will ultimately be used as a scan source or
-            // a lookup source. Since fluss lookup sources cannot accept filters yet, to be safe, we
-            // return all filters to the Flink planner.
+            // FLINK-38635 We cannot determine whether this source will ultimately be used as a scan
+            // source or a lookup source. Since fluss lookup sources cannot accept filters yet, to
+            // be safe, we return all filters to the Flink planner.
             return Result.of(acceptedFilters, filters);
         }
 


fluss-flink/fluss-flink-common/src/test/java/org/apache/fluss/flink/source/FlinkTableSourceBatchITCase.java
```diff
@@ -159,6 +159,84 @@ void testScanSingleRowFilterOnPartitionedTable() throws Exception {
         assertResultsIgnoreOrder(collected, expected, true);
     }
 
+    @Test
+    void testFilterOnLookupSource() throws Exception {
```
xuyangzhong, line 163: This is a defensive test, as currently, Flink batch does not push down filters to dim tables (an issue has already been created for this: https://issues.apache.org/jira/browse/FLINK-38634).

loserwang1024 (approved):
LGTM, please fix test first.

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/source/FlinkTableSource.java
```diff
@@ -572,7 +576,11 @@ && hasPrimaryKey()
                     }
                 }
             }
-            return Result.of(acceptedFilters, remainingFilters);
+
+            // We cannot determine whether this source will ultimately be used as a scan source or
+            // a lookup source. Since fluss lookup sources cannot accept filters yet, to be safe, we
+            // return all filters to the Flink planner.
+            return Result.of(acceptedFilters, filters);
```
luoyuxia, line 583: IIUC, it'll cause the plan still have the accepted filters?

xuyangzhong: Yes, you're right.

luoyuxia: Just wondering is there any way to avoid it. The source scan operator will always evaluate the filters even though the filters are already pushdown.
Is it possible to limit such behavior on lookup source only?

xuyangzhong: Currently, there is no way to determine whether the source during filter pushdown is a scan source or a lookup source in Flink. Given the current situation, in terms of performance and correctness, I believe we should always prioritize correctness.
Another approach, which might not be user-friendly, is to explicitly raise an error in `getLookupRuntimeProvider` when we detect that `singleRowFilter` or `partitionFilters` is not null (indicating that some filters have been applied). This error would inform users that lookup tables do not support filter pushdown and suggest adding a specific table options (e.g., enable filter push down = false) via a hint on the lookup table. However, this approach of raising errors could potentially harm the user experience. WDYT?

xuyangzhong: I have created a jira in Flink to follow this issue. https://issues.apache.org/jira/browse/FLINK-38635

loserwang1024 (approved):

PR #1940, opened 2025-11-06
[server] Prevent remote log TTL expiration until tiering completes

### LINKED ISSUE #1919: Prevent local data deletion until tiering completes
### Search before asking

### Description

Currently, if a table fails to dump data to Paimon within 2 minutes due to various reasons (network issues, Paimon service exceptions, etc.), the data on local disk will still be deleted when TTL expires, which may cause data loss.

## Current Behavior

- Local disk data is deleted strictly according to TTL expiration time
- Data will be deleted even if tiering (dumping to Paimon) has not completed
- This may lead to data loss risk

## Expected Behavior

- Local disk data should only be deleted according to TTL after successful tiering (dumping to Paimon)
- If data has not completed tiering, it should be retained even after TTL expiration until tiering is complete

### Willingness to contribute

LiebingYu:
### Purpose

Linked issue: close #1919

### DIFF
--- a/fluss-server/src/main/java/org/apache/fluss/server/log/remote/RemoteLogTablet.java
+++ b/fluss-server/src/main/java/org/apache/fluss/server/log/remote/RemoteLogTablet.java
@@ -155,13 +155,12 @@ public class RemoteLogTablet {
      * lake yet.
      *
      * @param currentTimeMs the current time in milliseconds
-     * @param isDataLakeEnabled whether data lake is enabled
-     * @param lakeLogEndOffset the log end offset that has been synced to lake, -1 if no lake sync
-     *     has occurred
+     * @param lakeLogEndOffset the log end offset that has been synced to lake, null if data lake is
+     *     disabled
      * @return list of expired segments that can be safely deleted
      */
     public List<RemoteLogSegment> expiredRemoteLogSegments(
-            long currentTimeMs, boolean isDataLakeEnabled, long lakeLogEndOffset) {
+            long currentTimeMs, Long lakeLogEndOffset) {
         if (!logExpireEnable()) {
             return Collections.emptyList();
         }
@@ -175,8 +174,9 @@ public class RemoteLogTablet {
                         if (currentTimeMs - ts > ttlMs) {
                             for (UUID uuid : entry.getValue()) {
                                 RemoteLogSegment segment = idToRemoteLogSegment.get(uuid);
-                                if (isDataLakeEnabled) {
-                                    // only include segments that have been tiered to lake.
+                                if (lakeLogEndOffset != null) {
+                                    // if datalake is enabled, only include segments that have been
+                                    // tiered to lake.
                                     if (segment.remoteLogEndOffset() <= lakeLogEndOffset) {
                                         expiredSegments.add(segment);
                                     }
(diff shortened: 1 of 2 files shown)


LiebingYu:
There's another problem for remote log TTL expiration when data lake enable. Let's say:
1. User create a Fluss table with data lake enable.
2. User disable the data lake.

The problem in step 2 is: We don't notify the Replica when disable the data lake. So remote log for this table will never been expired if user don't enable data lake again.

I suggest:
1. Temporarily prohibit users from disabling the lake before we resolve 2.
2. We need to discuss how data TTL should be handled if the user disables the data lake. One option is to do nothing—in this case, if the user never re-enables the data lake, all historical data will remain in Fluss, which would obviously put significant pressure on Fluss’s storage. Another option is to notify the replica when the data lake is disabled, so that afterwards the data's TTL will no longer consider the progress of tiering to the lake; the user would need to accept the risk of losing lake data if they have set an aggressive TTL. I prefer the latter approach. What do you think @luoyuxia ?

luoyuxia:
@LiebingYu The latter approach (ignoring tiering progress after disabling) might lead to silent data loss. For users, this is often an "unexpected behavior." If data correctness issues occur because the data was deleted before being successfully tiered, it becomes extremely difficult for us to troubleshoot and recover. We should require an explicit "force" flag or a specific confirmation if they truly want to bypass the tiering safety check.

I still prefer to not to do ttl when user disable the data lake. Realistically, if a user has already enabled the data lake, the scenarios where they would want to permanently disable it may be rare(I'm not sure the real case).

fluss-server/src/main/java/org/apache/fluss/server/log/LogTablet.java
```diff
@@ -516,9 +522,15 @@ public void updateMinRetainOffset(long minRetainOffset) {
...
 
+    public void updateIsDataLakeEnabled(boolean isDataLakeEnabled) {
+        this.isDataLakeEnabled = isDataLakeEnabled;
+    }
+
     public void updateLakeTableSnapshotId(long snapshotId) {
         if (snapshotId > this.lakeTableSnapshotId) {
             this.lakeTableSnapshotId = snapshotId;
+            // it means the data lake is enabled if we have got the snapshot id
+            this.isDataLakeEnabled = true;
```
luoyuxia, line 533: Currently, we rely this to update `isDataLakeEnabled` dynamically, in the other pr, we should rely on `updateMetataRequest` .

fluss-server/src/main/java/org/apache/fluss/server/log/remote/RemoteLogTablet.java
```diff
@@ -147,8 +147,21 @@ public List<RemoteLogSegment> allRemoteLogSegments() {
...
+     * lake yet.
+     *
+     * @param currentTimeMs the current time in milliseconds
+     * @param isDataLakeEnabled whether data lake is enabled
+     * @param lakeLogEndOffset the log end offset that has been synced to lake, -1 if no lake sync
+     *     has occurred
+     * @return list of expired segments that can be safely deleted
+     */
+    public List<RemoteLogSegment> expiredRemoteLogSegments(
+            long currentTimeMs, boolean isDataLakeEnabled, long lakeLogEndOffset) {
```
luoyuxia:
```suggestion
            long currentTimeMs, Long lakeLogEndOffset) {
```
null represent lake table?

fluss-server/src/main/java/org/apache/fluss/server/log/LogTablet.java
LiebingYu: Yes, we will handle this in https://github.com/apache/fluss/issues/2375.

wuchong (approved):

Issue #993, opened 2025-06-04
Implement obtainSecurityToken for google filesystem


luoyuxia:
### Search before asking

- [x] I searched in the [issues](https://github.com/alibaba/fluss/issues) and found nothing similar.

### Description

The newly introduced google filesystem missing `obtainSecurityToken` method. 

### Willingness to contribute

- [ ] I'm willing to submit a PR!

luoyuxia:
@gkatzioura Would you like to take it?

gkatzioura:
Going for it

wuchong:
@luoyuxia is it a blocker for users to use google filesystem?

luoyuxia:
> [@luoyuxia](https://github.com/luoyuxia) is it a blocker for users to use google filesystem?

Yes, it's. Without this, client can't read remote data in gfs.

wuchong:
@luoyuxia thanks, got it.

wuchong:
Hi @gkatzioura , Just checking in to see how things are going with this issue. No pressure at all — I'm just wondering if you're still able to work on it. If you're busy or need help, we'd be happy to assist or reassign it. Thanks again for taking it on!

gkatzioura:
@wuchong resuming on this one

PR #1941, opened 2025-11-06
Azure Blob Storage Support 1311

### LINKED ISSUE #1311: [file system] Add Azure Blob Storage
### Search before asking

### Motivation

Integration with [Microsoft Azure](https://azure.microsoft.com) ecosystem.

### Solution

Based on [Hadoop Azure](https://hadoop.apache.org/docs/r3.4.1/hadoop-azure/index.html)

### Anything else?

**Note:** Currently low priority, unless some users request this feature.

### Willingness to contribute

gkatzioura:
### Purpose

Linked issue: close #1311

Add File System support for Azure Blob Storage.

### DIFF
--- a/fluss-filesystems/fluss-fs-azure/src/main/resources/META-INF/NOTICE
+++ b/fluss-filesystems/fluss-fs-azure/src/main/resources/META-INF/NOTICE
@@ -57,6 +57,11 @@ This project bundles the following dependencies under the Go License (https://go
 
 - com.google.re2j:re2j:1.1
 
+This project bundles the following dependencies under the CDDL 1.1 license.
+See bundled license files for details.
+
+- javax.xml.bind:jaxb-api:2.3.1
+
 This project bundles the following dependencies under the Eclipse Distribution License (EDL) 1.0 (https://www.eclipse.org/org/documents/edl-v10.php). See bundled license files for details.
 
 - jakarta.activation:jakarta.activation-api:1.2.1
--- a/fluss-filesystems/fluss-fs-azure/src/main/resources/META-INF/licenses/LICENSE.jaxb
+++ b/fluss-filesystems/fluss-fs-azure/src/main/resources/META-INF/licenses/LICENSE.jaxb
@@ -1,0 +1,135 @@
+COMMON DEVELOPMENT AND DISTRIBUTION LICENSE (CDDL)Version 1.1
+
+1. Definitions.
+
+     1.1. "Contributor" means each individual or entity that creates or contributes to the creation of Modifications.
+
+     1.2. "Contributor Version" means the combination of the Original Software, prior Modifications used by a Contributor (if any), and the Modifications made by that particular Contributor.
+
+     1.3. "Covered Software" means (a) the Original Software, or (b) Modifications, or (c) the combination of files containing Original Software with files containing Modifications, in each case including portions thereof.
+
+     1.4. "Executable" means the Covered Software in any form other than Source Code.
+
+     1.5. "Initial Developer" means the individual or entity that first makes Original Software available under this License.
+
+     1.6. "Larger Work" means a work which combines Covered Software or portions thereof with code not governed by the terms of this License.
+
+     1.7. "License" means this document.
+
+     1.8. "Licensable" means having the right to grant, to the maximum extent possible, whether at the time of the initial grant or subsequently acquired, any and all of the rights conveyed herein.
+
+     1.9. "Modifications" means the Source Code and Executable form of any of the following:
+
+     A. Any file that results from an addition to, deletion from or modification of the contents of a file containing Original Software or previous Modifications;
+
+     B. Any new file that contains any part of the Original Software or previous Modification; or
+
+     C. Any new file that is contributed or otherwise made available under the terms of this License.
+
+     1.10. "Original Software" means the Source Code and Executable form of computer software code that is originally released under this License.
+
+     1.11. "Patent Claims" means any patent claim(s), now owned or hereafter acquired, including without limitation, method, process, and apparatus claims, in any patent Licensable by grantor.
+
+     1.12. "Source Code" means (a) the common form of computer software code in which modifications are made and (b) associated documentation included in or with such code.
+
+     1.13. "You" (or "Your") means an individual or a legal entity exercising rights under, and complying with all of the terms of, this License. For legal entities, "You" includes any entity which controls, is controlled by, or is under common control with You. For purposes of this definition, "control" means (a) the power, direct or indirect, to cause the direction or management of such entity, whether by contract or otherwise, or (b) ownership of more than fifty percent (50%) of the outstanding shares or beneficial ownership of such entity.
+
+2. License Grants.
+
+     2.1. The Initial Developer Grant.
+
+     Conditioned upon Your compliance with Section 3.1 below and subject to third party intellectual property claims, the Initial Developer hereby grants You a world-wide, royalty-free, non-exclusive license:
+
+     (a) under intellectual property rights (other than patent or trademark) Licensable by Initial Developer, to use, reproduce, modify, display, perform, sublicense and distribute the Original Software (or portions thereof), with or without Modifications, and/or as part of a Larger Work; and
+
+     (b) under Patent Claims infringed by the making, using or selling of Original Software, to make, have made, use, practice, sell, and offer for sale, and/or otherwise dispose of the Original Software (or portions thereof).
+
+     (c) The licenses granted in Sections 2.1(a) and (b) are effective on the date Initial Developer first distributes or otherwise makes the Original Software available to a third party under the terms of this License.
+
+     (d) Notwithstanding Section 2.1(b) above, no patent license is granted: (1) for code that You delete from the Original Software, or (2) for infringements caused by: (i) the modification of the Original Software, or (ii) the combination of the Original Software with other software or devices.
+
+     2.2. Contributor Grant.
+
+     Conditioned upon Your compliance with Section 3.1 below and subject to third party intellectual property claims, each Contributor hereby grants You a world-wide, royalty-free, non-exclusive license:
+
+     (a) under intellectual property rights (other than patent or trademark) Licensable by Contributor to use, reproduce, modify, display, perform, sublicense and distribute the Modifications created by such Contributor (or portions thereof), either on an unmodified basis, with other Modifications, as Covered Software and/or as part of a Larger Work; and
+
+     (b) under Patent Claims infringed by the making, using, or selling of Modifications made by that Contributor either alone and/or in combination with its Contributor Version (or portions of such combination), to make, use, sell, offer for sale, have made, and/or otherwise dispose of: (1) Modifications made by that Contributor (or portions thereof); and (2) the combination of Modifications made by that Contributor with its Contributor Version (or portions of such combination).
+
+     (c) The licenses granted in Sections 2.2(a) and 2.2(b) are effective on the date Contributor first distributes or otherwise makes the Modifications available to a third party.
+
+     (d) Notwithstanding Section 2.2(b) above, no patent license is granted: (1) for any code that Contributor has deleted from the Contributor Version; (2) for infringements caused by: (i) third party modifications of Contributor Version, or (ii) the combination of Modifications made by that Contributor with other software (except as part of the Contributor Version) or other devices; or (3) under Patent Claims infringed by Covered Software in the absence of Modifications made by that Contributor.
+
+3. Distribution Obligations.
+
+     3.1. Availability of Source Code.
+
+     Any Covered Software that You distribute or otherwise make available in Executable form must also be made available in Source Code form and that Source Code form must be distributed only under the terms of this License. You must include a copy of this License with every copy of the Source Code form of the Covered Software You distribute or otherwise make available. You must inform recipients of any such Covered Software in Executable form as to how they can obtain such Covered Software in Source Code form in a reasonable manner on or through a medium customarily used for software exchange.
+
+     3.2. Modifications.
+
+     The Modifications that You create or to which You contribute are governed by the terms of this License. You represent that You believe Your Modifications are Your original creation(s) and/or You have sufficient rights to grant the rights conveyed by this License.
+
+     3.3. Required Notices.
+
+     You must include a notice in each of Your Modifications that identifies You as the Contributor of the Modification. You may not remove or alter any copyright, patent or trademark notices contained within the Covered Software, or any notices of licensing or any descriptive text giving attribution to any Contributor or the Initial Developer.
+
+     3.4. Application of Additional Terms.
+
+     You may not offer or impose any terms on any Covered Software in Source Code form that alters or restricts the applicable version of this License or the recipients' rights hereunder. You may choose to offer, and to charge a fee for, warranty, support, indemnity or liability obligations to one or more recipients of Covered Software. However, you may do so only on Your own behalf, and not on behalf of the Initial Developer or any Contributor. You must make it absolutely clear that any such warranty, support, indemnity or liability obligation is offered by You alone, and You hereby agree to indemnify the Initial Developer and every Contributor for any liability incurred by the Initial Developer or such Contributor as a result of warranty, support, indemnity or liability terms You offer.
+
+     3.5. Distribution of Executable Versions.
+
+     You may distribute the Executable form of the Covered Software under the terms of this License or under the terms of a license of Your choice, which may contain terms different from this License, provided that You are in compliance with the terms of this License and that the license for the Executable form does not attempt to limit or alter the recipient's rights in the Source Code form from the rights set forth in this License. If You distribute the Covered Software in Executable form under a different license, You must make it absolutely clear that any terms which differ from this License are offered by You alone, not by the Initial Developer or Contributor. You hereby agree to indemnify the Initial Developer and every Contributor for any liability incurred by the Initial Developer or such Contributor as a result of any such terms You offer.
+
+     3.6. Larger Works.
+
+     You may create a Larger Work by combining Covered Software with other code not governed by the terms of this License and distribute the Larger Work as a single product. In such a case, You must make sure the requirements of this License are fulfilled for the Covered Software.
+
+4. Versions of the License.
+
+     4.1. New Versions.
+
+     Oracle is the initial license steward and may publish revised and/or new versions of this License from time to time. Each version will be given a distinguishing version number. Except as provided in Section 4.3, no one other than the license steward has the right to modify this License.
+
+     4.2. Effect of New Versions.
+
+     You may always continue to use, distribute or otherwise make the Covered Software available under the terms of the version of the License under which You originally received the Covered Software. If the Initial Developer includes a notice in the Original Software prohibiting it from being distributed or otherwise made available under any subsequent version of the License, You must distribute and make the Covered Software available under the terms of the version of the License under which You originally received the Covered Software. Otherwise, You may also choose to use, distribute or otherwise make the Covered Software available under the terms of any subsequent version of the License published by the license steward.
+
+     4.3. Modified Versions.
+
+     When You are an Initial Developer and You want to create a new license for Your Original Software, You may create and use a modified version of this License if You: (a) rename the license and remove any references to the name of the license steward (except to note that the license differs from this License); and (b) otherwise make it clear that the license contains terms which differ from this License.
+
+5. DISCLAIMER OF WARRANTY.
+
+     COVERED SOFTWARE IS PROVIDED UNDER THIS LICENSE ON AN "AS IS" BASIS, WITHOUT WARRANTY OF ANY KIND, EITHER EXPRESSED OR IMPLIED, INCLUDING, WITHOUT LIMITATION, WARRANTIES THAT THE COVERED SOFTWARE IS FREE OF DEFECTS, MERCHANTABLE, FIT FOR A PARTICULAR PURPOSE OR NON-INFRINGING. THE ENTIRE RISK AS TO THE QUALITY AND PERFORMANCE OF THE COVERED SOFTWARE IS WITH YOU. SHOULD ANY COVERED SOFTWARE PROVE DEFECTIVE IN ANY RESPECT, YOU (NOT THE INITIAL DEVELOPER OR ANY OTHER CONTRIBUTOR) ASSUME THE COST OF ANY NECESSARY SERVICING, REPAIR OR CORRECTION. THIS DISCLAIMER OF WARRANTY CONSTITUTES AN ESSENTIAL PART OF THIS LICENSE. NO USE OF ANY COVERED SOFTWARE IS AUTHORIZED HEREUNDER EXCEPT UNDER THIS DISCLAIMER.
+
+6. TERMINATION.
+
+     6.1. This License and the rights granted hereunder will terminate automatically if You fail to comply with terms herein and fail to cure such breach within 30 days of becoming aware of the breach. Provisions which, by their nature, must remain in effect beyond the termination of this License shall survive.
+
+     6.2. If You assert a patent infringement claim (excluding declaratory judgment actions) against Initial Developer or a Contributor (the Initial Developer or Contributor against whom You assert such claim is referred to as "Participant") alleging that the Participant Software (meaning the Contributor Version where the Participant is a Contributor or the Original Software where the Participant is the Initial Developer) directly or indirectly infringes any patent, then any and all rights granted directly or indirectly to You by such Participant, the Initial Developer (if the Initial Developer is not the Participant) and all Contributors under Sections 2.1 and/or 2.2 of this License shall, upon 60 days notice from Participant terminate prospectively and automatically at the expiration of such 60 day notice period, unless if within such 60 day period You withdraw Your claim with respect to the Participant Software against such Participant either unilaterally or pursuant to a written agreement with Participant.
+
+     6.3. If You assert a patent infringement claim against Participant alleging that the Participant Software directly or indirectly infringes any patent where such claim is resolved (such as by license or settlement) prior to the initiation of patent infringement litigation, then the reasonable value of the licenses granted by such Participant under Sections 2.1 or 2.2 shall be taken into account in determining the amount or value of any payment or license.
+
+     6.4. In the event of termination under Sections 6.1 or 6.2 above, all end user licenses that have been validly granted by You or any distributor hereunder prior to termination (excluding licenses granted to You by any distributor) shall survive termination.
+
+7. LIMITATION OF LIABILITY.
+
+     UNDER NO CIRCUMSTANCES AND UNDER NO LEGAL THEORY, WHETHER TORT (INCLUDING NEGLIGENCE), CONTRACT, OR OTHERWISE, SHALL YOU, THE INITIAL DEVELOPER, ANY OTHER CONTRIBUTOR, OR ANY DISTRIBUTOR OF COVERED SOFTWARE, OR ANY SUPPLIER OF ANY OF SUCH PARTIES, BE LIABLE TO ANY PERSON FOR ANY INDIRECT, SPECIAL, INCIDENTAL, OR CONSEQUENTIAL DAMAGES OF ANY CHARACTER INCLUDING, WITHOUT LIMITATION, DAMAGES FOR LOSS OF GOODWILL, WORK STOPPAGE, COMPUTER FAILURE OR MALFUNCTION, OR ANY AND ALL OTHER COMMERCIAL DAMAGES OR LOSSES, EVEN IF SUCH PARTY SHALL HAVE BEEN INFORMED OF THE POSSIBILITY OF SUCH DAMAGES. THIS LIMITATION OF LIABILITY SHALL NOT APPLY TO LIABILITY FOR DEATH OR PERSONAL INJURY RESULTING FROM SUCH PARTY'S NEGLIGENCE TO THE EXTENT APPLICABLE LAW PROHIBITS SUCH LIMITATION. SOME JURISDICTIONS DO NOT ALLOW THE EXCLUSION OR LIMITATION OF INCIDENTAL OR CONSEQUENTIAL DAMAGES, SO THIS EXCLUSION AND LIMITATION MAY NOT APPLY TO YOU.
+
+8. U.S. GOVERNMENT END USERS.
+
+     The Covered Software is a "commercial item," as that term is defined in 48 C.F.R. 2.101 (Oct. 1995), consisting of "commercial computer software" (as that term is defined at 48 C.F.R. ? 252.227-7014(a)(1)) and "commercial computer software documentation" as such terms are used in 48 C.F.R. 12.212 (Sept. 1995). Consistent with 48 C.F.R. 12.212 and 48 C.F.R. 227.7202-1 through 227.7202-4 (June 1995), all U.S. Government End Users acquire Covered Software with only those rights set forth herein. This U.S. Government Rights clause is in lieu of, and supersedes, any other FAR, DFAR, or other clause or provision that addresses Government rights in computer software under this License.
+
+9. MISCELLANEOUS.
+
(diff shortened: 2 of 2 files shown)


polyzos:
@gkatzioura, you mean -> Add File System support for Azure Blob Storage?

gkatzioura:
> @gkatzioura, you mean -> Add File System support for Azure Blob Storage?

Fixed

luoyuxia:
Just as https://github.com/apache/fluss/issues/993 , obtainSecurityToken method is required for client to access Azure Blob Storage when read from remote

gkatzioura:
> Just as #993 , obtainSecurityToken method is required for client to access Azure Blob Storage when read from remote

checking the options

polyzos:
@gkatzioura, we should make sure it works for all prefixes, like
https://nightlies.apache.org/flink/flink-docs-master/docs/deployment/filesystems/azure/#azure-blob-storage

polyzos:
**abfs://** works fine during my tests.

Screenshot 2025-12-03 at 11 28 12 AM

I pushed one commit to disable tests, because we can't meet the test coverage here, and I'm figuring out the issue with the license.

gkatzioura:
> **abfs://** works fine during my tests.
> 
> Screenshot 2025-12-03 at 11 28 12 AM
> I pushed one commit to disable tests, because we can't meet the test coverage here, and I'm figuring out the issue with the license.

Yes forgot to mention that I added the support for all formats

polyzos:
LicenseChecker fixed, and I just finished testing with all schemes. Everything works great 💪 🚀 
@luoyuxia @michaelkoepf Let me know if you have any comments on this PR, before merging

gkatzioura:
Hi @luoyuxia
some observations on the obtainSecurityToken functionality. 

I do receive the tokens when I read fluss data as a client

```dockerfile
try (LogScanner logScanner = table.newScan().createLogScanner()) {
            logScanner.subscribeFromBeginning(1);
            var record = logScanner.poll(Duration.ofSeconds(10));
            record.forEach(r -> _ );
}
```
Once log scanner is started the tokens start to be sent.
However I could not see the tokens being shared above fluss components (tablet-server coordinator server).
Similar to what happens with Flink JobManager tokens delegated to the TaskManager.
My assumption was that the token is shared among the fluss components then it is used with a 
hadoop custom provider type (Depends on the hadoop driver).

luoyuxia:
> Hi @luoyuxia some observations on the obtainSecurityToken functionality.
> 
> I do receive the tokens when I read fluss data as a client
> 
> ```dockerfile
> try (LogScanner logScanner = table.newScan().createLogScanner()) {
>             logScanner.subscribeFromBeginning(1);
>             var record = logScanner.poll(Duration.ofSeconds(10));
>             record.forEach(r -> _ );
> }
> ```
> 
> Once log scanner is started the tokens start to be sent. However I could not see the tokens being shared above fluss components (tablet-server coordinator server). Similar to what happens with Flink JobManager tokens delegated to the TaskManager. My assumption was that the token is shared among the fluss components then it is used with a hadoop custom provider type (Depends on the hadoop driver).

Hi, the security token is not shared above fluss components. Currently, it's only used by fluss-client, fluss-client require a token to read from remote directly.

fluss-filesystems/fluss-fs-abfs/src/main/java/org/apache/fluss/fs/abfs/AzureFileSystem.java
```diff
@@ -0,0 +1,66 @@
...
+ * Implementation of the Fluss {@link FileSystem} interface for Azure Blob Storage. This class
+ * implements the common behavior implemented directly by Fluss and delegates common calls to an
+ * implementation of Hadoop's filesystem abstraction.
+ */
+public class AzureFileSystem extends HadoopFileSystem {
+
+    private final String scheme;
+    private final Configuration conf;
+
+    private AzureDelegationTokenProvider delegationTokenProvider;
```
luoyuxia:
```suggestion
    private volatile AzureDelegationTokenProvider delegationTokenProvider;
```

fluss-filesystems/fluss-fs-abfs/src/main/java/org/apache/fluss/fs/abfs/token/AzureDelegationTokenProvider.java
```diff
@@ -0,0 +1,93 @@
...
+
+    public AzureDelegationTokenProvider(String scheme, Configuration conf) {
+        this.scheme = scheme;
+
+        this.clientId = conf.get(CLIENT_ID);
+        this.clientSecret = conf.get(CLIENT_SECRET);
+        this.authEndpoint = conf.get(ENDPOINT_KEY);
+        this.additionInfos = new HashMap<>();
+
+        for (String key : Arrays.asList(ENDPOINT_KEY)) {
```
luoyuxia:
```suggestion
        for (String key : Collections.singleton(ENDPOINT_KEY))
```

fluss-filesystems/fluss-fs-abfs/src/main/java/org/apache/fluss/fs/abfs/token/DynamicTemporaryAzureCredentialsProvider.java
```diff
@@ -0,0 +1,82 @@
...
+ * creation because that will break a chain of credential providers.
+ */
+public class DynamicTemporaryAzureCredentialsProvider extends AccessTokenProvider
+        implements CustomTokenProviderAdaptee {
+
+    public static final String NAME = DynamicTemporaryAzureCredentialsProvider.class.getName();
+
+    public static final String COMPONENT = "Dynamic session credentials for Fluss";
+
+    private volatile AzureADToken azureADToken;
```
luoyuxia: Seems azureADToken is not used?

fluss-filesystems/fluss-fs-azure/src/main/java/org/apache/fluss/fs/azure/token/AzureDelegationTokenReceiver.java
```diff
@@ -0,0 +1,93 @@
...
+/** Security token receiver for the abfs filesystem. */
+public class AzureDelegationTokenReceiver implements SecurityTokenReceiver {
+    public static final String PROVIDER_CONFIG_NAME = "fs.azure.account.oauth.provider.type";
+
+    private static final Logger LOG = LoggerFactory.getLogger(AzureDelegationTokenReceiver.class);
+
+    static volatile Credentials credentials;
+    static volatile Map<String, String> additionInfos;
+
+    public static void updateHadoopConfig(org.apache.hadoop.conf.Configuration hadoopConfig) {
```
luoyuxia, line 41: updateHadoopConfig is never called.

luoyuxia, line 41: You can refer `S3FileSystemPlugin` for where to call this method.
In `S3FileSystemPlugin`, it will check `ACCESS_KEY_ID` set or not.
- if set, do no thing. will use provide ak/sk
- if not set, it means the FileSystemPlugin is initialized in client side, it will need to get security token from Fluss to access s3. Then it'll call `S3DelegationTokenReceiver#updateHadoopConfig` to set the credential provider, whose token is from Fluss via method getSecurityToken

fluss-filesystems/fluss-fs-abfs/src/main/java/org/apache/fluss/fs/abfs/token/DynamicTemporaryAzureCredentialsProvider.java
gkatzioura: Removed, it is being handled behind the scenes by `org.apache.hadoop.fs.azurebfs.oauth2.AccessTokenProvider`

polyzos:
@gkatzioura is there anything else pending here?

gkatzioura:
@polyzos & @luoyuxia @wuchong fyi this is ready for the release

fluss-filesystems/fluss-fs-azure/src/main/java/org/apache/fluss/fs/azure/token/AzureDelegationTokenReceiver.java
```diff
@@ -0,0 +1,93 @@
...
+            hadoopConfig.set(PROVIDER_CONFIG_NAME, providers);
+        } else {
+            LOG.debug("Provider already exists");
+        }
+
+        // then, set addition info
+        if (additionInfos == null) {
+            // if addition info is null, it also means we have not received any token,
+            // we throw InvalidCredentialsException
+            throw new IllegalStateException(DynamicTemporaryAzureCredentialsProvider.COMPONENT);
```
leekeiabstraction, line 63: Name of exception type doesn't match between actual and comment

fluss-filesystems/fluss-fs-abfs/src/main/java/org/apache/fluss/fs/abfs/token/AzureDelegationTokenReceiver.java
```diff
@@ -0,0 +1,93 @@
...
+        LOG.info("Updating session credentials");
+
+        byte[] tokenBytes = token.getToken();
+
+        credentials = CredentialsJsonSerde.fromJson(tokenBytes);
+        additionInfos = token.getAdditionInfos();
+
+        LOG.info(
+                "Session credentials updated successfully with access key: {}.",
+                credentials.getAccessKeyId());
```
leekeiabstraction: This will always show up as null as AzureDelegationTokenProvider instantiates Credentials object with null `accessKeyId`, can we provide a more useful log message?

fluss-filesystems/fluss-fs-azure/src/test/java/org/apache/fluss/fs/azure/token/AuthServerHandler.java
```diff
@@ -0,0 +1,135 @@
...
+import static org.apache.fluss.shaded.guava32.com.google.common.net.HttpHeaders.CONTENT_ENCODING;
+import static org.apache.fluss.shaded.guava32.com.google.common.net.HttpHeaders.CONTENT_LENGTH;
+import static org.apache.fluss.shaded.netty4.io.netty.handler.codec.http.HttpHeaderNames.CONNECTION;
+import static org.apache.fluss.shaded.netty4.io.netty.handler.codec.http.HttpHeaderNames.CONTENT_TYPE;
+import static org.apache.fluss.shaded.netty4.io.netty.handler.codec.http.HttpHeaderValues.APPLICATION_JSON;
+import static org.apache.fluss.shaded.netty4.io.netty.handler.codec.http.HttpHeaderValues.CLOSE;
+import static org.apache.fluss.shaded.netty4.io.netty.handler.codec.http.HttpResponseStatus.NOT_FOUND;
+import static org.apache.fluss.shaded.netty4.io.netty.handler.codec.http.HttpResponseStatus.OK;
+
+/** Netty Handler for facilitating the Google auth token generation. */
```
leekeiabstraction: Should this say Azure instead of Google?

fluss-filesystems/fluss-fs-abfs/src/test/java/org/apache/fluss/fs/abfs/token/AuthServerHandler.java
```diff
@@ -0,0 +1,135 @@
...
+            HttpRequest req,
+            byte[] bytes,
+            HttpResponseStatus status,
+            AsciiString contentType) {
+        FullHttpResponse response =
+                new DefaultFullHttpResponse(
+                        req.protocolVersion(), status, Unpooled.wrappedBuffer(bytes));
+        response.headers()
+                .set(CONTENT_TYPE, contentType)
+                .set("Location", "http://localhost:8080/resumbable-upload")
```
leekeiabstraction: Question for my understanding: Why do we need to have this endpoint in test's AuthServerHandler?

fluss-filesystems/fluss-fs-abfs/src/test/java/org/apache/fluss/fs/abfs/token/AzureDelegationTokenProviderTest.java
```diff
@@ -0,0 +1,66 @@
...
+/** Tests for {@link AzureDelegationTokenProvider}. */
+public class AzureDelegationTokenProviderTest {
+
+    private static final String CONFIG_PREFIX = "fs.azure.account.oauth2.client";
+    private static final String CLIENT_ID = "testClientId";
+    private static final String CLIENT_SECRET = "testClientSecret";
+
+    private static final String ENDPOINT_KEY = "http://localhost:8080";
+
+    private static MockAuthServer mockGSServer;
```
leekeiabstraction: Should this say abfs instead?

fluss-filesystems/fluss-fs-azure/src/test/java/org/apache/fluss/fs/azure/token/AzureDelegationTokenProviderTest.java
```diff
@@ -0,0 +1,66 @@
...
+        configuration.set(CONFIG_PREFIX + ".id", CLIENT_ID);
+        configuration.set(CONFIG_PREFIX + ".secret", CLIENT_SECRET);
+        configuration.set(CONFIG_PREFIX + ".endpoint", ENDPOINT_KEY);
+        AzureDelegationTokenProvider azureDelegationTokenProvider =
+                new AzureDelegationTokenProvider("abfs", configuration);
+        ObtainedSecurityToken obtainedSecurityToken =
+                azureDelegationTokenProvider.obtainSecurityToken();
+        byte[] token = obtainedSecurityToken.getToken();
+        Credentials credentials = CredentialsJsonSerde.fromJson(token);
+        assertThat(credentials.getSecurityToken()).isEqualTo("token");
```
leekeiabstraction, line 63: We should also assert on `accessKeyId` and `secretAccessKey`, even if they are expected to be null.

fluss-filesystems/fluss-fs-abfs/src/test/java/org/apache/fluss/fs/abfs/token/AzureDelegationTokenReceiverTest.java
```diff
@@ -0,0 +1,89 @@
...
+
+import static org.apache.fluss.fs.abfs.token.AzureDelegationTokenReceiver.PROVIDER_CONFIG_NAME;
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
+
+/** Tests for {@link AzureDelegationTokenReceiver}. */
+class AzureDelegationTokenReceiverTest {
+
+    private static final String PROVIDER_CLASS_NAME = "TestProvider";
+    private static final String REGION = "testRegion";
```
leekeiabstraction: Let's remove REGION if it is unused.

fluss-filesystems/fluss-fs-azure/src/test/java/org/apache/fluss/fs/azure/token/DynamicTemporaryAzureCredentialsProviderTest.java
```diff
@@ -0,0 +1,67 @@
...
+        AzureDelegationTokenReceiver receiver = new AzureDelegationTokenReceiver();
+
+        byte[] json = CredentialsJsonSerde.toJson(credentials);
+
+        ObtainedSecurityToken obtainedSecurityToken =
+                new ObtainedSecurityToken("abfs", json, 1L, new HashMap<>());
+        receiver.onNewTokensObtained(obtainedSecurityToken);
+
+        AzureADToken azureADToken = provider.getToken();
+        assertThat(azureADToken.getAccessToken()).isEqualTo(credentials.getSecurityToken());
```
leekeiabstraction, line 74: Let's also assert on `accessKeyId` and `secretAccessKey`

fluss-filesystems/fluss-fs-abfs/src/test/java/org/apache/fluss/fs/abfs/token/DynamicTemporaryAzureCredentialsProviderTest.java
```diff
@@ -0,0 +1,67 @@
...
+import java.util.HashMap;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
+
+/** Tests for {@link DynamicTemporaryAzureCredentialsProvider}. */
+class DynamicTemporaryAzureCredentialsProviderTest {
+
+    private static final String CLIENT_ID = "testClientId";
+    private static final String CLIENT_SECRET = "testClientSecret";
```
leekeiabstraction: Let's use null values for these to be consistent with the behaviour of Provider.

fluss-filesystems/fluss-fs-abfs/src/test/java/org/apache/fluss/fs/abfs/AbfsFileSystemBehaviorITCase.java
```diff
@@ -0,0 +1,82 @@
...
+
+    private static final String CONFIG_PREFIX = "fs.azure.account";
+    private static final String CLIENT_ID = "testClientId";
+    private static final String CLIENT_SECRET = "testClientSecret";
+
+    private static final String AZURE_ACCOUNT_KEY = "ZmFrZS1rZXkK";
+    private static final String ENDPOINT_KEY = "http://localhost:8080";
+    public static final String ABFS_FS_PATH = "abfs://flus@test.dfs.core.windows.net/test";
+
+    private static MockAuthServer mockGSServer;
```
leekeiabstraction: Should this be named `mockAbfsServer` instead?

fluss-filesystems/fluss-fs-abfs/src/main/java/org/apache/fluss/fs/abfs/token/AzureDelegationTokenReceiver.java
gkatzioura: I changed the message and switched it to debug since only securityToken is sent, thus better not to leak it.

fluss-filesystems/fluss-fs-azure/src/test/java/org/apache/fluss/fs/azure/token/AuthServerHandler.java
gkatzioura: Changed to Azure

fluss-filesystems/fluss-fs-abfs/src/test/java/org/apache/fluss/fs/abfs/token/AuthServerHandler.java
gkatzioura: non needed removed

fluss-filesystems/fluss-fs-abfs/src/test/java/org/apache/fluss/fs/abfs/token/AzureDelegationTokenProviderTest.java
gkatzioura: Changed to mockAuthServer

fluss-filesystems/fluss-fs-azure/src/test/java/org/apache/fluss/fs/azure/token/AzureDelegationTokenProviderTest.java
gkatzioura: Added, seems like CredentialsJsonSerde maps null to "null", likely something to check on another ticket.

fluss-filesystems/fluss-fs-azure/src/main/java/org/apache/fluss/fs/azure/token/AzureDelegationTokenReceiver.java
gkatzioura: changed

fluss-filesystems/fluss-fs-azure/pom.xml
```diff
@@ -0,0 +1,362 @@
...
+            <groupId>org.apache.fluss</groupId>
+            <artifactId>fluss-fs-hadoop</artifactId>
+            <version>${project.version}</version>
+        </dependency>
+
+        <!-- Hadoop's file system abstraction (bundled) -->
+        <dependency>
+            <groupId>org.apache.hadoop</groupId>
+            <artifactId>hadoop-common</artifactId>
+            <version>${fs.hadoopshaded.version}</version>
```
wuchong: We should replace the current dependency with the `fluss-fs-hadoop-shaded` artifact, consistent with how other filesystems are handled. This avoids the need to manually declare a direct dependency on `hadoop-common`, which is difficult to maintain under centralized dependency management and often pulls in unnecessary transitive dependencies.

Moreover, depending directly on `hadoop-common` introduces several third-party libraries licensed under Category B (as defined in https://www.apache.org/legal/resolved.html). Including such dependencies requires us to bundle corresponding license notices in the resource folder, significantly increasing maintenance overhead.

Instead, using the shaded artifact cleanly encapsulates these dependencies. The correct declaration is:

```xml
<dependency>
    <groupId>org.apache.fluss</groupId>
    <artifactId>fluss-fs-hadoop-shaded</artifactId>
    <version>${project.version}</version>
</dependency>
```

fluss-filesystems/fluss-fs-azure/src/main/resources/META-INF/NOTICE
```diff
@@ -0,0 +1,86 @@
...
+
+- javax.xml.bind:jaxb-api:2.2.2
+- javax.xml.bind:jaxb-api:2.3.1
+- javax.activation:activation:1.1
+- com.sun.xml.bind:jaxb-impl:2.2.3-1
+- com.github.pjfanning:jersey-json:1.20
+
+This project bundles the following dependencies under the Eclipse Distribution License (EDL) 1.0 (https://www.eclipse.org/org/documents/edl-v10.php). See bundled license files for details.
+
+- jakarta.activation:jakarta.activation-api:1.2.1
```
wuchong, line 67: This change introduces many Category B dependencies. According to Apache Software Foundation policy, all Category B dependencies require a corresponding `LICENSE-xxx` notice in the resource folder. These dependencies are likely pulled in because of a direct dependency on `hadoop-common`.

To avoid this, please try replacing the direct `hadoop-common` dependency with `fluss-fs-hadoop-shaded`. This shaded artifact should encapsulate those transitive dependencies and may eliminate the need to declare them explicitly in the `NOTICE` file.

fluss-filesystems/fluss-fs-azure/src/main/java/org/apache/fluss/fs/azure/AbfsFileSystemPlugin.java
```diff
@@ -0,0 +1,123 @@
...
+import static org.apache.fluss.fs.abfs.token.AzureDelegationTokenProvider.ACCOUNT_KEY;
+import static org.apache.fluss.fs.abfs.token.AzureDelegationTokenProvider.CLIENT_ID;
+import static org.apache.fluss.fs.abfs.token.AzureDelegationTokenReceiver.PROVIDER_CONFIG_NAME;
+
+/** Simple factory for the Azure File System. */
+public class AbfsFileSystemPlugin implements FileSystemPlugin {
+
+    private static final Logger LOG = LoggerFactory.getLogger(AbfsFileSystemPlugin.class);
+
+    private static final String[] FLUSS_CONFIG_PREFIXES = {"azure.", "fs.azure.", "fs.wasb."};
```
wuchong: IIUC, these are the configuration options Fluss exposes to users for configuring Azure File System. Since this is the first time we’re making this API public, can we restrict the config prefix to `fs.azure.` only? This aligns with other filesystem implementations and avoids confusion about which prefix users should use.

fluss-test-coverage/pom.xml
```diff
@@ -362,6 +362,7 @@
                                         <exclude>org.apache.fluss.fs.hdfs.HdfsSecurityTokenReceiver</exclude>
                                         <exclude>org.apache.fluss.fs.oss.*</exclude>
                                         <exclude>org.apache.fluss.fs.s3.*</exclude>
+                                        <exclude>org.apache.fluss.fs.abfs.*</exclude>
```
wuchong: Have you tried to remove this exclusion? It seems you have added many tests that will always run, that should have some test coverages.

fluss-filesystems/fluss-fs-gs/src/test/java/org/apache/fluss/fs/gs/AuthServerHandler.java
```diff
@@ -43,7 +43,7 @@
 import static org.apache.fluss.shaded.netty4.io.netty.handler.codec.http.HttpResponseStatus.NOT_FOUND;
 import static org.apache.fluss.shaded.netty4.io.netty.handler.codec.http.HttpResponseStatus.OK;
 
-/** Netty Handler for facilitating the Google auth token generation. */
+/** Netty Handler for facilitating the Azure auth token generation. */
 public class AuthServerHandler extends SimpleChannelInboundHandler<HttpObject> {
```
wuchong, line 47: revert. This is under google fs package

fluss-filesystems/fluss-fs-azure/src/test/java/org/apache/fluss/fs/azure/token/AuthServerHandler.java
```diff
@@ -0,0 +1,134 @@
...
```
wuchong:
```suggestion
/** Netty Handler for facilitating the Azure auth token generation. */
```

fluss-filesystems/fluss-fs-azure/src/main/java/org/apache/fluss/fs/azure/WasbFileSystemPlugin.java
```diff
@@ -0,0 +1,27 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.fluss.fs.abfs;
+
+/** Simple factory for the Abfs file system, registered for the <tt>wasb://</tt> scheme. */
+public class WasbFileSystemPlugin extends AbfsFileSystemPlugin {
```
wuchong: The current javadocs are too brief and don't explain what these schemes are or their use cases. Suggested javadocs:

```java
/**
 * FileSystem plugin for Azure Blob Storage using the ABFS (Azure Blob File System) driver.
 * Registered for the {@code abfs://} scheme.
 *
 * <p>ABFS is the recommended driver for accessing Azure Data Lake Storage Gen2. 
 * Use this scheme for non-SSL connections to ADLS Gen2 storage accounts.
 *
 * <p>URI format: {@code abfs://<container>@<storage-account>.dfs.core.windows.net/<path>}
 */
public class AbfsFileSystemPlugin extends AzureFileSystemPlugin
```

```java
/**
 * FileSystem plugin for Azure Blob Storage using the ABFS driver with SSL/TLS encryption.
 * Registered for the {@code abfss://} scheme.
 *
 * <p>This is the secure (SSL-enabled) variant of ABFS, recommended for production use 
 * with Azure Data Lake Storage Gen2.
 *
 * <p>URI format: {@code abfss://<container>@<storage-account>.dfs.core.windows.net/<path>}
 */
public class AbfssFileSystemPlugin extends AzureFileSystemPlugin
```

```java
/**
 * FileSystem plugin for Azure Blob Storage using the WASB (Windows Azure Storage Blob) driver.
 * Registered for the {@code wasb://} scheme.
 *
 * <p>WASB is the legacy driver for accessing Azure Blob Storage. Consider using ABFS 
 * for new deployments as it provides better performance and security features.
 *
 * <p>URI format: {@code wasb://<container>@<storage-account>.blob.core.windows.net/<path>}
 */
public class WasbFileSystemPlugin extends AzureFileSystemPlugin
```

```java
/**
 * FileSystem plugin for Azure Blob Storage using the WASB driver with SSL/TLS encryption.
 * Registered for the {@code wasbs://} scheme.
 *
 * <p>This is the secure (SSL-enabled) variant of WASB for legacy Azure Blob Storage access.
 * For new deployments, consider using {@code abfss://} with Azure Data Lake Storage Gen2.
 *
 * <p>URI format: {@code wasbs://<container>@<storage-account>.blob.core.windows.net/<path>}
 */
public class WasbsFileSystemPlugin extends AzureFileSystemPlugin
```

This helps users understand:
1. What each abbreviation stands for
2. When to use each scheme
3. The URI format for configuration
4. The relationship between legacy (WASB) and modern (ABFS) drivers

fluss-filesystems/fluss-fs-azure/src/main/java/org/apache/fluss/fs/azure/token/AzureDelegationTokenReceiver.java
```diff
@@ -0,0 +1,93 @@
...
+                hadoopConfig.set(entry.getKey(), entry.getValue());
+            }
+        }
+
+        LOG.info("Updated Hadoop configuration successfully");
+    }
+
+    @Override
+    public String scheme() {
+        return "abfs";
```
wuchong: Is this `AzureDelegationTokenReceiver` can only be used for `abfs` or can be shared for all azure filesystems? If it can be shared, we should create `AzureDelegationTokenReceiver` for each azure schemes and register it to SPI.

fluss-filesystems/fluss-fs-azure/src/main/java/org/apache/fluss/fs/azure/AbfsFileSystemPlugin.java
```diff
@@ -0,0 +1,123 @@
...
+        return "abfs";
+    }
+
+    @Override
+    public FileSystem create(URI fsUri, Configuration flussConfig) throws IOException {
+        org.apache.hadoop.conf.Configuration hadoopConfig = getHadoopConfiguration(flussConfig);
+
+        setCredentialProvider(hadoopConfig);
+
+        // create the Google Hadoop FileSystem
```
wuchong:
```suggestion
        // create the Azure Hadoop FileSystem
```

fluss-filesystems/fluss-fs-azure/src/main/java/org/apache/fluss/fs/azure/AzureFileSystem.java
```diff
@@ -0,0 +1,66 @@
...
+ */
+public class AzureFileSystem extends HadoopFileSystem {
+
+    private final String scheme;
+    private final Configuration conf;
+
+    private volatile AzureDelegationTokenProvider delegationTokenProvider;
+
+    /**
+     * Wraps the given Hadoop File System object as a Flink File System object. The given Hadoop
```
wuchong:
```suggestion
     * Wraps the given Hadoop File System object as a Fluss File System object. The given Hadoop
```

wuchong:
Also, since the branch is far behind `main`, could you please rebase it onto the latest `main` branch? This will make the CI results more reliable.

fluss-filesystems/fluss-fs-azure/src/main/java/org/apache/fluss/fs/azure/token/AzureDelegationTokenReceiver.java
gkatzioura: Followed the same strategy as with the filesystem plugins, made an abstract class

fluss-filesystems/fluss-fs-azure/pom.xml
gkatzioura: Unfortunately due to missing class issues I had to revert, can follow up on another branch

polyzos: @wuchong it gives missing class exceptions
```
coordinator-server-1  | java.lang.NoClassDefFoundError: org/apache/hadoop/fs/impl/BackReference
coordinator-server-1  |         at org.apache.hadoop.fs.azurebfs.AzureBlobFileSystem.initialize(AzureBlobFileSystem.java:200) ~[?:?]
coordinator-server-1  |         at org.apache.fluss.fs.azure.AzureFileSystemPlugin.create(AzureFileSystemPlugin.java:60) ~[?:?]
coordinator-server-1  |         at org.apache.fluss.fs.azure.AbfsFileSystemPlugin.create(AbfsFileSystemPlugin.java:29) ~[?:?]
```

And thus the NOTICE file can't be updated either, I guess... maybe open a follow-up ticket as an improvement and investigate in the future? Unless you have any thoughts on this issue

PR #1942, opened 2025-11-07
[metric] Update metric group infix to meet document

zcoo:
### Purpose

Linked issue: close #xxx

### Brief change log

Remove an infix "log_" so that log part and kv part keep the same format.


swuferhong (approved):
LGTM. The current implementation is inconsistent with both the design document and the metrics documentation for version 0.8. This PR addresses and fixes the issue.

PR #1943, opened 2025-11-07
Fix Javadoc link to `Table`.

geoand:
Apparently, Javadoc doesn't like it when the class after the `@link` is in another line

For example, in IntelliJ the rendering looks like this before the change:

before

and like this after the change:

after


polyzos:
FYI @MehulBatra

PR #1946, opened 2025-11-08
Support Aggregation Merge Engine for Primary Key Tables

### LINKED ISSUE #212: [Feature] Introduce Merge Engines for Primary-Key Table
### Search before asking

### Motivation

We propose the introduction of merge engines for primary-key tables, enhancing data management and query capabilities. This feature will encompass the following merge engines, each with distinct functionalities:

- Version-Based Merge: This merge engine will allow data consolidation based on version numbers. It is very similar to ClickHouse [ReplacingMergeTree](https://clickhouse.com/docs/en/guides/replacing-merge-tree) that keeps the row with the highest version number. This is also very similar to HBase [checkAndPut](https://hbase.apache.org/1.2/apidocs/org/apache/hadoop/hbase/client/Table.html#checkAndPut(byte[],%20byte[],%20byte[],%20byte[],%20org.apache.hadoop.hbase.client.Put)) that updates records only when the condition is true. In this way, the out-of-order data can be guaranteed to be ultimately consistent with the upstream. 

- First-Row Based Merge: Designed to retain the first occurrence of a row with the same primary key, this engine boosts performance by minimizing writes to the kv store and generates only insert-only changelogs (some downstream jobs only accept insert-only changelogs). 

- Aggregate Table Merge: This engine will facilitate the merging of data by aggregating records, which is particularly useful for summarizing information and analytics purposes. It will support operations like `sum`, `max`, `min`, `count`, `avg`, among others, providing robust analytic capabilities.

### Solution

Introduce table properties (configured via Flink DDL with options):
```properties
'table.merge-engine' = 'first_row|high_version|aggregate'
'table.merge-engine.version.column' = '<version_column>'
'table.merge-engine.aggregate.<agg_column>' =  'sum | max | min | count | avg | first_value | last_value' 
```

The `merge-engine` properties are table storage properties, and therefore can't be changed after table creation. 

### Version-Based Merge

```sql
CREATE TABLE fluss_table (
  id BIGINT,
  version BIGINT,
  data STRING,
  PRIMARY KEY (id) NOT ENFORCED
) WITH (
  'connector' = 'fluss',
  'bootstrap.address' = '...',
  'table.merge-engine' = 'high_version',
  'table.merge-engine.version.column' = 'version'  -- updates when new version >= old version  
);
```

### First-Row-Based Merge

```sql
CREATE TABLE fluss_table (
  id BIGINT,
  version BIGINT,
  data STRING,
  PRIMARY KEY (id) NOT ENFORCED
) WITH (
  'connector' = 'fluss',
  'bootstrap.address' = '...',
  'table.merge-engine' = 'first_row' -- only the first row of the primary key will be retained.
);
```

### Aggregate-Based Merge

```sql
CREATE TABLE fluss_table (
    id BIGINT,
    price DOUBLE,
    sales BIGINT,
    PRIMARY KEY (id) NOT ENFORCED
) WITH (
    'table.merge-engine' = 'aggregate',
    'table.merge-engine.aggregate.price' = 'max',
    'table.merge-engine.aggregate.sales' = 'sum'
);
```

### Anything else?

Subtasks: 

- [ ] Introduce first-row merge engine for primary key table
- [ ] Introduce high-version merge engine for primary key table
- [ ] Introduce aggregate merge engine for primary key table

### Willingness to contribute

platinumhamburg:
### Purpose

Linked issue: close #212

### DIFF
--- a/fluss-common/src/main/java/org/apache/fluss/config/ConfigOptions.java
+++ b/fluss-common/src/main/java/org/apache/fluss/config/ConfigOptions.java
@@ -1387,9 +1387,10 @@ public class ConfigOptions {
                     .noDefaultValue()
                     .withDescription(
                             "Defines the merge engine for the primary key table. By default, primary key table doesn't have merge engine. "
-                                    + "The supported merge engines are `first_row` and `versioned`. "
+                                    + "The supported merge engines are `first_row`, `versioned`, and `aggregate`. "
                                     + "The `first_row` merge engine will keep the first row of the same primary key. "
-                                    + "The `versioned` merge engine will keep the row with the largest version of the same primary key.");
+                                    + "The `versioned` merge engine will keep the row with the largest version of the same primary key. "
+                                    + "The `aggregate` merge engine will aggregate rows with the same primary key using field-level aggregate functions.");
 
     public static final ConfigOption<String> TABLE_MERGE_ENGINE_VERSION_COLUMN =
             // we may need to introduce "del-column" in the future to support delete operation
@@ -1399,6 +1400,31 @@ public class ConfigOptions {
                     .withDescription(
                             "The column name of the version column for the `versioned` merge engine. "
                                     + "If the merge engine is set to `versioned`, the version column must be set.");
+
+    /**
+     * Prefix for aggregate merge engine configuration options.
+     *
+     * <p>Aggregate-specific options use the pattern: {@code
+     * table.merge-engine.aggregate.<field-name>}
+     */
+    public static final String AGGREGATE_MERGE_ENGINE_PREFIX = "table.merge-engine.aggregate";
+
+    public static final ConfigOption<String> TABLE_AGGREGATE_DEFAULT_FUNCTION =
+            key("table.merge-engine.aggregate.default-function")
+                    .stringType()
+                    .noDefaultValue()
+                    .withDescription(
+                            "Default aggregate function for all non-primary-key fields when using aggregate merge engine. "
+                                    + "If not set, 'last_non_null_value' will be used as default. "
+                                    + "Field-specific aggregate functions can be configured using 'table.merge-engine.aggregate.<field-name>'.");
+
+    public static final ConfigOption<Boolean> TABLE_AGG_REMOVE_RECORD_ON_DELETE =
+            key("table.aggregation.remove-record-on-delete")
+                    .booleanType()
+                    .defaultValue(false)
+                    .withDescription(
+                            "Whether to remove the entire record when receiving a DELETE operation in aggregate merge engine. "
+                                    + "By default, DELETE operations are not supported in aggregate merge engine.");
 
     public static final ConfigOption<DeleteBehavior> TABLE_DELETE_BEHAVIOR =
             key("table.delete.behavior")
--- a/fluss-common/src/main/java/org/apache/fluss/config/TableConfig.java
+++ b/fluss-common/src/main/java/org/apache/fluss/config/TableConfig.java
@@ -26,6 +26,8 @@ import org.apache.fluss.metadata.LogFormat;
 import org.apache.fluss.metadata.MergeEngineType;
 import org.apache.fluss.utils.AutoPartitionStrategy;
 
+import javax.annotation.Nullable;
+
 import java.time.Duration;
 import java.util.Optional;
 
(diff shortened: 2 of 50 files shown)


polyzos:
@platinumhamburg This is 🔥
I’m curious though, doesn’t the aggregation merge engine, require some kind of transactions mechanism to be introduced first?
If a job fails and restarts, does it ensure deterministic results?

platinumhamburg:
> @platinumhamburg This is 🔥 I’m curious though, doesn’t the aggregation merge engine, require some kind of transactions mechanism to be introduced first? If a job fails and restarts, does it ensure deterministic results?

Thanks @polyzos, your concern is absolutely correct. Currently, the KV engine does not have distributed transaction capabilities. The KV transaction commit relies solely on the local High Watermark, and even its idempotency mechanism is only based on the LogTablet's idempotency manager. Therefore, the KV engine currently only provides AT LEAST ONCE semantics and cannot provide Exactly Once semantics. This is not an issue introduced by the Aggregation Merge Engine itself, but aggregation operations are particularly sensitive to this limitation.

After careful evaluation, I agree that the reliability of the Aggregation Merge Engine does depend on distributed transactions. However, I still believe it can be treated as a relatively independent functional unit. The distributed transaction support for the KV engine is a broader topic that probably deserves a separate discussion or issue to address comprehensively.

polyzos:
@platinumhamburg i share similar thoughts.
What I’m curious is the best way to approach this until transactions are introduced.

I’m thinking that maybe it makes sense to combine it with a versioned merge engine table upstream to handle any potential duplication 
that might be introduced.

WDYT?

platinumhamburg:
@polyzos In my current consideration, I believe the consistency issue should be addressed separately in two domains:

- For Streaming Computation Domain: I think we should expose WriterState checkpoint/restore interfaces to the upper-layer compute engines. Engines like Flink can leverage these interfaces to implement state persistence, allowing Fluss to utilize the existing idempotence manager to ensure idempotent data writes.

- For Batch Processing Domain: The problem becomes significantly more complex, as each batch processing task execution will create independent Writers, and the idempotence manager cannot guarantee computational consistency. Imagine a batch processing task that fails over midway through writing data - the aggregation state of all related primary keys will be completely corrupted. To ensure consistency in the batch processing domain, we will need:

	- Explicit BeginTransaction (PrepareTransaction is optional) / CommitTransaction / AbortTransaction interfaces exposed to compute engines

	- True transactional visibility control for single Buckets (not through transactional buffers like PrewriteBuffer, but complete transaction control based on MvccValueEncoder/MvccValueDecoder + TransactionStateManager)

	- Compute engines coordinating calls to Bucket local transaction managers through a global transaction manager to complete transactional writes in an orderly manner.

In summary, for streaming computation consistency, we should be able to achieve the required consistency guarantees with relatively minor modifications based on our current infrastructure. However, for batch processing, we will need to introduce a major architectural extension.

platinumhamburg:
> @polyzos In my current consideration, I believe the consistency issue should be addressed separately in two domains:
> 
> * For Streaming Computation Domain: I think we should expose WriterState checkpoint/restore interfaces to the upper-layer compute engines. Engines like Flink can leverage these interfaces to implement state persistence, allowing Fluss to utilize the existing idempotence manager to ensure idempotent data writes.
> * For Batch Processing Domain: The problem becomes significantly more complex, as each batch processing task execution will create independent Writers, and the idempotence manager cannot guarantee computational consistency. Imagine a batch processing task that fails over midway through writing data - the aggregation state of all related primary keys will be completely corrupted. To ensure consistency in the batch processing domain, we will need:
>   
>   * Explicit BeginTransaction (PrepareTransaction is optional) / CommitTransaction / AbortTransaction interfaces exposed to compute engines
>   * True transactional visibility control for single Buckets (not through transactional buffers like PrewriteBuffer, but complete transaction control based on MvccValueEncoder/MvccValueDecoder + TransactionStateManager)
>   * Compute engines coordinating calls to Bucket local transaction managers through a global transaction manager to complete transactional writes in an orderly manner.
> 
> In summary, for streaming computation consistency, we should be able to achieve the required consistency guarantees with relatively minor modifications based on our current infrastructure. However, for batch processing, we will need to introduce a major architectural extension.

I actually overestimated the functionality of WriterState. Currently, the idempotency manager of WriterIdState cannot actually ensure streaming data consistency either, as its state recording is currently quite limited. Therefore, regardless of streaming or batch processing, the related mechanisms need to be reimplemented uniformly.

platinumhamburg:
This PR will be closed and split into several smaller PRs.

PR #1947, opened 2025-11-09
[blog] feat: blogs postgres to fluss cdc pipeline

gyang94:
### Purpose

A blog to introduce postgres to fluss cdc pipeline

Linked issue: close #xxx

### DIFF
--- a/website/blog/2025-11-09-cdc-postgres-to-fluss.md
+++ b/website/blog/2025-11-09-cdc-postgres-to-fluss.md
@@ -18,92 +18,91 @@ Ready to see it in action? The diagram below shows our target setup. Let's dive 
 
 ## Prerequisite
 
-### Postgres
+### Postgres and Fluss
 
-You need a running PostgreSQL server with `logical` replication enabled. You can install it via Docker, Homebrew, or by downloading it directly from [postgresql.org](http://postgresql.org/). We will use Homebrew for this demonstration:
-
-```bash
-# Install postgresql
-brew install postgresql
-
-# Start postgres server
-brew services start postgresql
-```
-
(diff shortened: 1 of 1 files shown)


wuchong (commented):
Thank you @gyang94 for the great blog!

Can we offer a docker-compose file to orchestrate all the components: PostgreSQL, Fluss, and Flink? A one-command setup would greatly simplify onboarding for all users. The brew command only  works on Macbook, however, not everyone uses Macbook.

gyang94:
@wuchong updated to docker for PostgresSQL and Fluss. 
For Flink, since it is required for Flink CDC to get the `FLINK_HOME` directory to load configs or libs in CLI, it may not good to put Flink into Docker as well. Thus keep to run Flink locally. Please check.

wuchong:
Hi @gyang94,  

Thank you for the contribution—this blog post is very helpful!  

However, I’d like to suggest **holding off on publishing it for now**. We’ve just merged **schema evolution support** into the `main` branch, and the upcoming Flink CDC release will also add **schema evolution support in pipeline mode for Fluss**. Once both are available, this blog will be even more compelling, complete, and valuable to users.  

In the meantime, would you be open to contributing a **new top-level section ("Data Ingestion")** to the official documentation? We could include **Flink CDC as a synchronization tool** and add a guide on how to use it with Fluss. This would lay the groundwork for the blog and provide immediate value to early adopters.  StarRocks also has such guide: https://docs.starrocks.io/docs/loading/Flink-connector-starrocks/#synchronize-data-with-flink-cdc-30-with-schema-change-supported

Thanks again for your great work!

PR #1951, opened 2025-11-10
[kv] Introduce a shared memory cache for PrewriteBuffer between different replica

### LINKED ISSUE #1950: Introduce a shared memory cache for PrewriteBuffer between different replica
### Search before asking

### Motivation

When testing the stability of `PrimaryKey tables`, a frequent issue arises where exceptionally high write traffic peaks cause delayed log synchronization. This leads to continuous accumulation in the KV's `PrewriteBuffer`. Since the current `PrewriteBuffer` lacks a back-pressure mechanism, has no memory cap, and utilizes JVM heap memory, it can easily trigger OOM errors, resulting in `TabletServer` crashes.

Therefore, Fluss needs to implement a back-pressure mechanism to prevent this scenario. A promising approach under consideration is to allocate a fixed-size block of off-heap memory shared across different replicas, while also imposing per-replica memory limits.

### Solution

_No response_

### Anything else?

_No response_

### Willingness to contribute

swuferhong:
### Purpose

Linked issue: close #1950 

When testing the stability of `PrimaryKey` tables, a frequent issue arises where exceptionally high write traffic peaks cause delayed log synchronization. This leads to continuous accumulation in the KV's `PrewriteBuffer`. Since the current `PrewriteBuffer` lacks a back-pressure mechanism, has no memory cap, and utilizes JVM heap memory, it can easily trigger OOM errors, resulting in TabletServer crashes.

In this pr, we will introduce an off-heap shared memory cache to share memory between different `Replica`.


fluss-server/src/main/java/org/apache/fluss/server/kv/prewrite/KvPreWriteBufferMemoryPool.java
```diff
@@ -0,0 +1,205 @@
...
+import java.nio.ByteBuffer;
+import java.util.Map;
+import java.util.concurrent.atomic.AtomicLong;
+import java.util.concurrent.locks.Lock;
+import java.util.concurrent.locks.ReentrantLock;
+
+import static org.apache.fluss.utils.concurrent.LockUtils.inLock;
+
+/** a shared off-heap memory pool for {@link KvPreWriteBuffer} of different tableBucket. */
+public class KvPreWriteBufferMemoryPool {
```
platinumhamburg, line 38: I'm curious about why not to reuse the LazyMemorySegmentPool?

swuferhong: `LazyMemorySegmentPool` uses a large memory chunks (page size), for example, the default value of 128 KB. Personally, I don't think it's necessary to set it this large, because a single value of kv record here isn't as big as a recordBatch.

PR #1971, opened 2025-11-12
[docs] add an explanation for setting the configuration of kv.snapshot.interval to 0s

qiuyunbuyu:
While conducting tests following the [official Docker deployment guide](https://fluss.apache.org/docs/install-deploy/deploying-with-docker/), I noticed that all **kv.snapshot.interval** parameters are set to **0S**. I found this configuration quite confusing, as the behavior it implies wasn't immediately clear.

I checked the [configuration documentation](https://fluss.apache.org/docs/maintenance/configuration/#kv), but it didn't provide a clear explanation for a value of **0S**. To understand its meaning, I looked through the code commit history and found the relevant context in [PR #30](https://github.com/apache/fluss/pull/30).

Now that I understand its purpose, I would like to propose that we add specific documentation to clarify the special meaning and intended use-case when this parameter is set to 0S. This would prevent confusion for future users.


website/docs/maintenance/configuration.md
```diff
@@ -132,7 +132,7 @@ during the Fluss cluster working.
...
...                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                               |
```
rionmonster: I'm wondering if this would read slightly more clearly as:

> The interval to perform periodic snapshots for kv data. The default setting is 10 minutes. When set to 0s, snapshotting will be disabled.

qiuyunbuyu:
> Left a single comment/wording nit; otherwise LGTM!

@rionmonster Thank you for the review. I agree with your suggestion. Using "disable" does make the behavior more explicit and aligns with the existing comment in [apache/fluss#30](https://github.com/apache/fluss/pull/30/files). I've applied the change.

PR #1991, opened 2025-11-17
[kv] PrimaryKey table support standby replica to reduce recovery time

### LINKED ISSUE #1990: Primary key table support standby replica
### Search before asking

### Motivation

Primary key table support standby replica. 

During testing, I found that the primary reason for the slow recovery of PrimaryKey tables is the slow download of snapshot files, which accounts for about 90% of the total recovery time, while the apply cdc log phase takes up only a small fraction.

Therefore, could we introduce a standby replica that continuously synchronizes the latest snapshots? During recovery, we could restore directly from this standby replica. 

This approach might even eliminate the need for rebalancing of PrimaryKey table.

### Solution

_No response_

### Anything else?

_No response_

### Willingness to contribute

swuferhong:
### Purpose

Linked issue: close #1990


fluss-server/src/main/java/org/apache/fluss/server/replica/Replica.java
```diff
@@ -545,11 +570,27 @@ private void registerLakeTieringMetrics() {
...
+            checkNotNull(kvSnapshotManager);
+            if (isNowStandby) {
+                kvSnapshotManager.becomeStandby();
+                becomeStandby();
+            } else {
+                // to be new follower.
+                kvSnapshotManager.becomeFollower();
+                if (wasStandby || wasLeader) {
+                    // standby -> leader or leader -> leader
+                    dropKv();
```
platinumhamburg, line 619: When a replica leaves standby (or leader) role, `dropKv()` deletes local KV data. However, the standby download cache inside `KvSnapshotManager` (`downloadedSstFiles`, `downloadedMiscFiles`, `standbySnapshotSize`) is never cleared. If the replica later becomes standby again, `incrementalDownloadSnapshot()` will reuse stale sets and skip downloads for files that no longer exist.

fluss-server/src/main/java/org/apache/fluss/server/replica/ReplicaManager.java
```diff
@@ -758,13 +768,33 @@ public void notifyKvSnapshotOffset(
...
                     // update the snapshot offset.
                     TableBucket tb = notifyKvSnapshotOffsetData.getTableBucket();
-                    LogTablet logTablet = getReplicaOrException(tb).getLogTablet();
-                    logTablet.updateMinRetainOffset(
-                            notifyKvSnapshotOffsetData.getMinRetainOffset());
+                    Replica replica = getReplicaOrException(tb);
+                    if (notifyKvSnapshotOffsetData.getSnapshotId() != null) {
+                        // try to download the snapshot for standby replica.
+                        try {
+                            replica.downloadSnapshot(notifyKvSnapshotOffsetData.getSnapshotId());
```
platinumhamburg: Snapshot download runs under `replicaStateChangeLock`. Long downloads can block leader/follower transitions and other state changes.

fluss-server/src/test/java/org/apache/fluss/server/replica/KvSnapshotITCase.java
platinumhamburg, line 67: - No test verifying standby demotion and re-standby.
- No test verifying failure and recovery paths (download failure, readiness)

fluss-common/src/main/java/org/apache/fluss/metrics/MetricNames.java
```diff
@@ -73,6 +73,7 @@ public class MetricNames {
     public static final String SERVER_LOGICAL_STORAGE_KV_SIZE = "kvSize";
     public static final String SERVER_PHYSICAL_STORAGE_LOCAL_SIZE = "localSize";
     public static final String SERVER_PHYSICAL_STORAGE_REMOTE_LOG_SIZE = "remoteLogSize";
+    public static final String SERVER_PHYSICAL_STORAGE_STANDBY_SIZE = "standbySize";
```
platinumhamburg, line 93: Add a brief comment and doc entry clarifying what `standbySize` represents.

swuferhong:
@platinumhamburg comments addressed.

fluss-server/src/test/java/org/apache/fluss/server/replica/KvSnapshotITCase.java
```diff
@@ -188,6 +190,434 @@ void testKvSnapshotAndDelete() throws Exception {
...
+
+        // wait for standby to eventually download snapshot (may retry on failures)
+        KvSnapshotManager standbySnapshotManager = standbyReplica.getKvSnapshotManager();
+        assertThat(standbySnapshotManager).isNotNull();
+
+        // verify snapshot is eventually downloaded successfully
+        retry(
+                Duration.ofMinutes(1),
+                () -> {
+                    assertThat(standbySnapshotManager.getDownloadedSstFiles()).isNotEmpty();
```
platinumhamburg, line 596: The current test assertion lacks sufficient robustness. We must verify that, following various failure scenarios, the snapshot maintained by the standby node (provided it retains the standby role) is fully valid. Specifically, the snapshot must:
(a) comprise a complete file set—not merely a partial subset;
(b) contain SST files that permit successful database opening;
(c) enable the node to transition seamlessly to the leader role;
(d) support correct recovery after the role transition; and
(e) guarantee zero data loss throughout the process.

fluss-server/src/main/java/org/apache/fluss/server/kv/snapshot/KvSnapshotManager.java
```diff
@@ -0,0 +1,654 @@
...
+        }
+        return latestSnapshot;
+    }
+
+    private void incrementalDownloadSnapshot(CompletedSnapshot completedSnapshot) throws Exception {
+        if (downloadedSstFiles == null || downloadedMiscFiles == null) {
+            // first try to load all ready exists sst files.
+            downloadedSstFiles = new HashSet<>();
+            downloadedMiscFiles = new HashSet<>();
+            loadKvLocalFiles(downloadedSstFiles, downloadedMiscFiles);
```
platinumhamburg, line 246: While reviewing the test code, I’ve developed serious concerns about the current snapshot download implementation. Let me elaborate on these concerns:

- KVSnapshot emphasizes atomicity and completeness: This means that for any given snapshot ID, there must be a deterministic and complete set of files. Partial or incomplete snapshots are strictly disallowed.

- Snapshots are discrete, not incremental: There is no smooth, continuous transition from one snapshot to another. Intermediate, incomplete states may not support consistent startup of RocksDB.

- Log recovery cannot function correctly during an incomplete snapshot transition: The current log recovery mechanism lacks precise mapping of recovery offsets, making it impossible to reliably recover from a partially downloaded snapshot.

- Real-world interruptions are not adequately handled: In practice, downloads can be interrupted at any point. However, neither the current implementation nor the existing tests fully validate how the system recovers from such intermediate, inconsistent states.

Recommendations for Implementation:
- Use a temporary directory for all file downloads and snapshot operations: Avoid performing any writes directly in the actual KvTablet directory. Instead, stage everything in a temporary directory.
- Atomically "commit" only after full consistency is achieved: Only when the temporary directory contains a complete, consistent snapshot—matching the expected file list for the given snapshot ID, and ideally verified with checksums—should it be atomically moved (e.g., via rename) into the final location.

Recommendations for Testing:
- Add test cases that explicitly validate snapshot consistency: Ensure that every snapshot loaded by the system meets the required completeness and integrity criteria.
- Introduce fault injection in tests: Simulate real-world failure scenarios (e.g., abrupt process termination during download) to verify that snapshot integrity is never compromised, even under adverse conditions.

swuferhong:
This pr will be closed because there is already some new pr that splits commits for easier review.

Issue #1731, opened 2025-09-21
[Client] Java Client Add Pojo Support


polyzos:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/fluss/issues) and found nothing similar.

### Description

#1404 introduced a RowConverterUtils to handle the conversion between InternalRow <> POJOs.

Next, we need to use these utilities to allow the users to work with POJOs directly. For example
```
AppendWriter<User> writer = table.newAppend().createWriter(User.class);
User u = new User(1, "John");
writer.append(u)
```

We should handle the `AppendWriter`, `UpsertWriter`, the `LogScanner` and also ensure it can be used by operations like `lookups`, `projections`, and `partial updates`.

### Willingness to contribute

- [x] I'm willing to submit a PR!

gyang94:
Is anyone working on this? If not, I'd like to take it.
Thinking: just leverage the existing `PojoToRowConverter` to convert a pojo object into a row in the writer.

polyzos:
Hi @gyang94 👋 this issue has been addressed here  #1404, but it is for internal use. The next step now is to add this to the java api directly, so users can write POJOS directly, and the work introduced in the PR above can take care of the conversion.

gyang94:
Hi @polyzos got it, thanks for the explanation. Now we are going to add `upsert(T record) and delete(T record)` in `UpsertWriter` interfaces to allow users to input a pojo object directly, instead of manually convert pojo to row.
So what's the next for this issue? Directly implement these functions based on utils in #1404 ,  or design the interfaces and details first as Jark's comment `have another design/FIP to do the final API expose.` ?

polyzos:
The two main steps should be:
1. to make the APIs - TableWriter and related implementations - generic
2. Provide implementations for the related API based on the converter

I'm not sure if we need a FIP here. To me, it seems like there is no design change, but rather it's about making the existing ones more generic. @wuchong WDTY?

PR #1992, opened 2025-11-17
[Client] Add Pojo Support 

polyzos:
This addresses #1731

### DIFF
--- a/fluss-client/src/main/java/org/apache/fluss/client/table/scanner/log/ScanRecords.java
+++ b/fluss-client/src/main/java/org/apache/fluss/client/table/scanner/log/ScanRecords.java
@@ -37,9 +37,7 @@ import java.util.Set;
  */
 @PublicEvolving
 public class ScanRecords implements Iterable<ScanRecord> {
-    public static final ScanRecords empty() {
-        return new ScanRecords(Collections.emptyMap());
-    }
+    public static final ScanRecords EMPTY = new ScanRecords(Collections.emptyMap());
 
     private final Map<TableBucket, List<ScanRecord>> records;
 
--- a/fluss-client/src/main/java/org/apache/fluss/client/table/writer/TableAppend.java
+++ b/fluss-client/src/main/java/org/apache/fluss/client/table/writer/TableAppend.java
@@ -41,7 +41,6 @@ public class TableAppend implements Append {
 
     @Override
     public <T> TypedAppendWriter<T> createTypedWriter(Class<T> pojoClass) {
-        AppendWriterImpl delegate = new AppendWriterImpl(tablePath, tableInfo, writerClient);
-        return new TypedAppendWriterImpl<>(delegate, pojoClass, tableInfo);
+        return new TypedAppendWriterImpl<>(createWriter(), pojoClass, tableInfo);
     }
 }
--- a/fluss-client/src/main/java/org/apache/fluss/client/table/writer/TableUpsert.java
+++ b/fluss-client/src/main/java/org/apache/fluss/client/table/writer/TableUpsert.java
@@ -98,8 +98,6 @@ public class TableUpsert implements Upsert {
 
     @Override
     public <T> TypedUpsertWriter<T> createTypedWriter(Class<T> pojoClass) {
-        UpsertWriterImpl delegate =
-                new UpsertWriterImpl(tablePath, tableInfo, targetColumns, writerClient);
-        return new TypedUpsertWriterImpl<>(delegate, pojoClass, tableInfo, targetColumns);
+        return new TypedUpsertWriterImpl<>(createWriter(), pojoClass, tableInfo, targetColumns);
     }
 }
--- a/fluss-client/src/main/java/org/apache/fluss/client/table/writer/TableWriter.java
+++ b/fluss-client/src/main/java/org/apache/fluss/client/table/writer/TableWriter.java
@@ -28,7 +28,7 @@ import org.apache.fluss.config.ConfigOptions;
  * @since 0.1
  */
 @PublicEvolving
-public interface TableWriter extends AutoCloseable {
+public interface TableWriter {
 
     /**
      * Flush data written that have not yet been sent to the server, forcing the client to send the
@@ -38,9 +38,4 @@ public interface TableWriter extends AutoCloseable {
      * results in an error.
      */
     void flush();
-
-    @Override
-    default void close() throws Exception {
-        // by default do nothing
-    }
 }
--- a/fluss-client/src/main/java/org/apache/fluss/client/table/writer/TypedAppendWriterImpl.java
+++ b/fluss-client/src/main/java/org/apache/fluss/client/table/writer/TypedAppendWriterImpl.java
@@ -30,14 +30,12 @@ import java.util.concurrent.CompletableFuture;
  */
 class TypedAppendWriterImpl<T> implements TypedAppendWriter<T> {
 
-    private final AppendWriterImpl delegate;
-    private final Class<T> pojoClass;
+    private final AppendWriter delegate;
     private final RowType tableSchema;
     private final PojoToRowConverter<T> pojoToRowConverter;
 
-    TypedAppendWriterImpl(AppendWriterImpl delegate, Class<T> pojoClass, TableInfo tableInfo) {
+    TypedAppendWriterImpl(AppendWriter delegate, Class<T> pojoClass, TableInfo tableInfo) {
         this.delegate = delegate;
-        this.pojoClass = pojoClass;
         this.tableSchema = tableInfo.getRowType();
         this.pojoToRowConverter = PojoToRowConverter.of(pojoClass, tableSchema, tableSchema);
     }
@@ -45,11 +43,6 @@ class TypedAppendWriterImpl<T> implements TypedAppendWriter<T> {
     @Override
     public void flush() {
         delegate.flush();
-    }
-
-    @Override
-    public void close() throws Exception {
-        delegate.close();
     }
 
     @Override
--- a/fluss-client/src/main/java/org/apache/fluss/client/table/writer/TypedUpsertWriterImpl.java
+++ b/fluss-client/src/main/java/org/apache/fluss/client/table/writer/TypedUpsertWriterImpl.java
@@ -33,22 +33,10 @@ import java.util.concurrent.CompletableFuture;
  */
 class TypedUpsertWriterImpl<T> implements TypedUpsertWriter<T> {
 
-    @Override
-    public void flush() {
-        delegate.flush();
-    }
-
-    private final UpsertWriterImpl delegate;
-
-    @Override
-    public void close() throws Exception {
-        delegate.close();
-    }
-
-    private final Class<T> pojoClass;
+    private final UpsertWriter delegate;
     private final TableInfo tableInfo;
     private final RowType tableSchema;
-    private final int[] targetColumns; // may be null
+    @Nullable private final int[] targetColumns;
 
     private final RowType pkProjection;
     @Nullable private final RowType targetProjection;
@@ -58,12 +46,8 @@ class TypedUpsertWriterImpl<T> implements TypedUpsertWriter<T> {
     @Nullable private final PojoToRowConverter<T> targetConverter;
 
     TypedUpsertWriterImpl(
-            UpsertWriterImpl delegate,
-            Class<T> pojoClass,
-            TableInfo tableInfo,
-            int[] targetColumns) {
+            UpsertWriter delegate, Class<T> pojoClass, TableInfo tableInfo, int[] targetColumns) {
         this.delegate = delegate;
-        this.pojoClass = pojoClass;
         this.tableInfo = tableInfo;
         this.tableSchema = tableInfo.getRowType();
         this.targetColumns = targetColumns;
@@ -80,6 +64,11 @@ class TypedUpsertWriterImpl<T> implements TypedUpsertWriter<T> {
                 (targetProjection == null)
                         ? null
                         : PojoToRowConverter.of(pojoClass, tableSchema, targetProjection);
+    }
+
+    @Override
+    public void flush() {
+        delegate.flush();
     }
 
     @Override
@@ -124,6 +113,9 @@ class TypedUpsertWriterImpl<T> implements TypedUpsertWriter<T> {
             // set PK fields, others null
             for (String pk : tableInfo.getPhysicalPrimaryKeys()) {
                 int projIndex = projection.getFieldIndex(pk);
+
+                // TODO: this can be optimized by pre-computing
+                // the index mapping in the constructor?
                 int fullIndex = tableSchema.getFieldIndex(pk);
                 full.setField(fullIndex, projected.getField(projIndex));
             }
--- a/fluss-client/src/test/java/org/apache/fluss/client/table/FlussTypedClientITCase.java
+++ b/fluss-client/src/test/java/org/apache/fluss/client/table/FlussTypedClientITCase.java
@@ -49,6 +49,7 @@ import java.time.LocalTime;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
+import java.util.Objects;
 
 import static org.assertj.core.api.Assertions.assertThat;
 
@@ -115,6 +116,86 @@ public class FlussTypedClientITCase extends ClientToServerITCaseBase {
             this.tsNtz = tsNtz;
             this.tsLtz = tsLtz;
         }
+
+        @Override
+        public boolean equals(Object o) {
+            if (this == o) {
+                return true;
+            }
+            if (o == null || getClass() != o.getClass()) {
+                return false;
+            }
+            AllTypesPojo that = (AllTypesPojo) o;
+            return Objects.equals(a, that.a)
+                    && Objects.equals(bool1, that.bool1)
+                    && Objects.equals(tiny, that.tiny)
+                    && Objects.equals(small, that.small)
+                    && Objects.equals(intv, that.intv)
+                    && Objects.equals(big, that.big)
+                    && Objects.equals(flt, that.flt)
+                    && Objects.equals(dbl, that.dbl)
+                    && Objects.equals(ch, that.ch)
+                    && Objects.equals(str, that.str)
+                    && Arrays.equals(bin, that.bin)
+                    && Arrays.equals(bytes, that.bytes)
+                    && Objects.equals(dec, that.dec)
+                    && Objects.equals(dt, that.dt)
+                    && Objects.equals(tm, that.tm)
+                    && Objects.equals(tsNtz, that.tsNtz)
+                    && Objects.equals(tsLtz, that.tsLtz);
+        }
+
+        @Override
+        public int hashCode() {
+            int result =
+                    Objects.hash(
+                            a, bool1, tiny, small, intv, big, flt, dbl, ch, str, dec, dt, tm, tsNtz,
+                            tsLtz);
+            result = 31 * result + Arrays.hashCode(bin);
+            result = 31 * result + Arrays.hashCode(bytes);
+            return result;
+        }
+
+        @Override
+        public String toString() {
+            return "AllTypesPojo{"
+                    + "a="
+                    + a
+                    + ", bool1="
+                    + bool1
+                    + ", tiny="
+                    + tiny
+                    + ", small="
+                    + small
+                    + ", intv="
+                    + intv
+                    + ", big="
+                    + big
+                    + ", flt="
+                    + flt
+                    + ", dbl="
+                    + dbl
+                    + ", ch="
+                    + ch
+                    + ", str='"
+                    + str
+                    + '\''
+                    + ", bin="
+                    + Arrays.toString(bin)
+                    + ", bytes="
+                    + Arrays.toString(bytes)
+                    + ", dec="
+                    + dec
+                    + ", dt="
+                    + dt
+                    + ", tm="
+                    + tm
+                    + ", tsNtz="
+                    + tsNtz
+                    + ", tsLtz="
+                    + tsLtz
+                    + '}';
+        }
     }
 
     /** Minimal POJO representing the primary key for {@link AllTypesPojo}. */
@@ -125,6 +206,28 @@ public class FlussTypedClientITCase extends ClientToServerITCaseBase {
 
         public PLookupKey(Integer a) {
             this.a = a;
+        }
+
+        @Override
+        public boolean equals(Object o) {
+            if (this == o) {
+                return true;
+            }
+            if (o == null || getClass() != o.getClass()) {
+                return false;
+            }
+            PLookupKey that = (PLookupKey) o;
+            return Objects.equals(a, that.a);
+        }
+
+        @Override
+        public int hashCode() {
+            return Objects.hash(a);
+        }
+
+        @Override
+        public String toString() {
+            return "PLookupKey{" + "a=" + a + '}';
         }
     }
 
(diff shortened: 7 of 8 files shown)


polyzos:
FYI @wuchong .. This PR adds support for writing/scanning Pojos directly with the client, while keeping the API as is.
The only difference is that now it's typed.

When I find some time, I want to test the effect in terms of performance, between writing/scanning with Pojos vs InternalRows.

fluss-client/src/main/java/org/apache/fluss/client/converter/PojoType.java
```diff
@@ -73,12 +74,13 @@ static <T> PojoType<T> of(Class<T> pojoClass) {
         for (Map.Entry<String, Field> e : allFields.entrySet()) {
             String name = e.getKey();
             Field field = e.getValue();
+            // Enforce nullable fields: primitives are not allowed in POJO definitions.
             if (field.getType().isPrimitive()) {
                 throw new IllegalArgumentException(
-                        String.format(
-                                "POJO class %s has primitive field '%s' of type %s. Primitive types are not allowed; all fields must be nullable (use wrapper types).",
-                                pojoClass.getName(), name, field.getType().getName()));
+                        "Primitive types are not allowed; all fields must be nullable (use wrapper types).");
```
leekeiabstraction: Curious on why we're removing information that would be useful for debugging from logs? It would be nice to at least retain pojoClass name so that users know which pojo the IAE is thrown for.

fluss-client/src/main/java/org/apache/fluss/client/table/scanner/log/TypedLogScanner.java
```diff
@@ -0,0 +1,102 @@
...
+        for (TableBucket bucket : records.buckets()) {
+            List<ScanRecord<InternalRow>> list = records.records(bucket);
+            List<ScanRecord<T>> converted = new ArrayList<>(list.size());
+            for (ScanRecord<InternalRow> r : list) {
+                InternalRow row = r.getValue();
+                T pojo = converter.fromRow(row);
+                converted.add(
+                        new ScanRecord<>(r.logOffset(), r.timestamp(), r.getChangeType(), pojo));
+            }
+            out.put(bucket, converted);
```
leekeiabstraction: I like the clean delegate approach, however what is the performance implication here? We would be looking at 2x memory usage per poll and 2x conversions when compared against non-POJO use cases?

Should POJO conversion be pushed down to e.g. `CompletedFetch.toScanRecord(LogRecord)`? Alternatively, we should note this for future improvement.

polyzos:
@leekeiabstraction . Indeed, we are looking at roughly 2x performance penalty.
I tested writing/scanning on both PK and Log tables with 10 million records.
I ran 5 iterations for each and calculated the average.

This is basically a trade-off for the users. Using InternalRow/GenericRow directly is way more efficient; however, this might come with some extra complexity and boilerplatete code.

For this reason, I want to give flexibility, probably leave the docs as is, with GenericRow being the go-to approach, but also add a section that Pojos can be used directly and maybe highlight this trade-off.

Moving forward I'm thinking that maybe it makes sense to add some helper classes that also derive the schema for the table from a Pojo.
**Log Table**
Screenshot 2025-11-25 at 3 32 34 PM

**Primary Key Table**
Screenshot 2025-11-25 at 3 51 46 PM

leekeiabstraction:
Thank you @polyzos for addressing the commends and also providing the data from your tests! That gives a clear picture on the performance. Further response questions below:

> This is basically a trade-off for the users. Using InternalRow/GenericRow directly is way more efficient; however, this might come with some extra complexity and boilerplatete code.
> For this reason, I want to give flexibility, probably leave the docs as is, with GenericRow being the go-to approach, but also add a section that Pojos can be used directly and maybe highlight this trade-off.

IMO, this does not need to be a trade-off. I am curious if you have explored implementation complexity or cons around pushing the Pojo conversion down further so that conversion to/from `InternalRow` can be skipped altogether? From a quick look, it does seem like you have most of the interfaces updated to support it.

> Moving forward I'm thinking that maybe it makes sense to add some helper classes that also derive the schema for the table from a Pojo.

Trying to understand, does this relate to performance or a separate thread on further changes that you're planning?

polyzos:
@leekeiabstraction yes, it's a separate thread.
Can you provide some more context in terms of what you mean - push the conversion further down?
On the scan side, I see you mentioned optimization in the `CompletedFetch.toScanRecord(LogRecord)`, that is something indeed that I didn't think of and might result in some further optimization. 
Is there something also on the write side you are thinking of?

leekeiabstraction:
> Can you provide some more context in terms of what you mean - push the conversion further down?
> On the scan side, I see you mentioned optimization in the CompletedFetch.toScanRecord(LogRecord), that is something indeed that I didn't think of and might result in some further optimization.

My aim is so that we can eliminate performance penalty by avoiding performing conversion twice. By "pushing conversion down", I mean you can do something like ~~moving the RowToPojoConverter method calls into LogScannerImpl/CompletedFetch.~~ refactoring so LogScannerImpl so that conversion is done directly to pojo. Strategy pattern could be useful here.  

> Is there something also on the write side you are thinking of?

Currently not, I'm very new to this code base but it's certainly worth exploring especially if we're also converting twice on the write side. Happy to have a look at the write side as well.

fluss-client/src/main/java/org/apache/fluss/client/table/writer/TypedAppendWriter.java
```diff
@@ -0,0 +1,58 @@
...
+    public void flush() {
+        delegate.flush();
+    }
+
+    @Override
+    public CompletableFuture<AppendResult> append(T record) {
+        if (record instanceof InternalRow) {
+            return delegate.append((InternalRow) record);
+        }
+        InternalRow row = pojoToRowConverter.toRow(record);
```
leekeiabstraction: Similar comment here on avoiding converting twice. (POJO -> InternalRow -> IndexedRow).

fluss-client/src/main/java/org/apache/fluss/client/table/writer/TypedUpsertWriter.java
```diff
@@ -0,0 +1,133 @@
...
+                        ? null
+                        : PojoToRowConverter.of(pojoClass, tableSchema, targetProjection);
+    }
+
+    @Override
+    public CompletableFuture<UpsertResult> upsert(T record) {
+        if (record instanceof InternalRow) {
+            return delegate.upsert((InternalRow) record);
+        }
+        InternalRow row = convertPojo(record, /*forDelete=*/ false);
```
leekeiabstraction: Similar comment here on avoiding converting twice. (POJO -> InternalRow -> WriteRecord).

fluss-client/src/main/java/org/apache/fluss/client/table/writer/TypedAppendWriter.java
polyzos: the main conversion here Pojo -> InternalRow.
IndexedRow is of type InternalRow and is used when the user specifies indexed as the log format. 
Arrow is the default one.

fluss-client/src/main/java/org/apache/fluss/client/table/writer/TypedUpsertWriter.java
polyzos: WriteRecord is not a format conversion; it represents operations

fluss-client/src/main/java/org/apache/fluss/client/lookup/TypedLookuper.java
```diff
@@ -0,0 +1,73 @@
...
+        RowType tableSchema = tableInfo.getRowType();
+        RowType keyProjection;
+        if (lookupColumnNames == null) {
+            keyProjection = tableSchema.project(tableInfo.getPrimaryKeys());
+        } else {
+            keyProjection = tableSchema.project(lookupColumnNames);
+        }
+        @SuppressWarnings("unchecked")
+        Class<K> keyClass = (Class<K>) key.getClass();
+        PojoToRowConverter<K> keyConv = PojoToRowConverter.of(keyClass, tableSchema, keyProjection);
```
wuchong: These are very heavy operations and should be invoked on hot-path. The `PojoToRowConverter` should be initialized in the constructor. This can improve the performance a lot.

fluss-client/src/main/java/org/apache/fluss/client/lookup/Lookup.java
```diff
@@ -93,7 +99,11 @@ default Lookup lookupBy(String... lookupColumnNames) {
...
      * changed with ({@link #lookupBy(List)}) method.
      *
+     * <p>The returned lookuper accepts generic keys of type {@code K}. If a key is a POJO, the
+     * client implementation will convert it to an {@code InternalRow} based on the table schema and
+     * the active lookup columns.
+     *
      * @return the lookuper
      */
-    Lookuper createLookuper();
+    <K> Lookuper<K> createLookuper();
```
wuchong: This is a big change for Fluss users, as `Lookuper`, `Appender`, and `Upserter` are fundamental data access interfaces. To maintain backward compatibility, we should **keep these existing interfaces unchanged**.

Instead, we can introduce a new generic interface: `TypedLookuper<T>`. This interface adds a type-safe lookup method:

```java
CompletableFuture<LookupResult> lookup(T lookupKey);
```

We can then provide a new factory method in `Lookup` to create it:

```java
<K> TypedLookuper<T> createTypedLookuper(Class<T> pojoClass);
```

This approach preserves existing APIs while enabling type-safe lookups for users who need them.

fluss-client/src/main/java/org/apache/fluss/client/table/scanner/Scan.java
```diff
@@ -63,7 +64,14 @@ public interface Scan {
...
      */
-    LogScanner createLogScanner();
+    LogScanner<InternalRow> createLogScanner();
+
+    /**
+     * Creates a typed LogScanner to continuously read log data as POJOs of the given class.
+     *
+     * <p>Note: this API doesn't support pre-configured with {@link #limit(int)}.
+     */
+    <T> LogScanner<T> createLogScanner(Class<T> pojoClass);
```
wuchong: Same to `TypedLookuper`, we should keep `LogScanner` unchanged and introduce 

```java
public interface TypedLogScanner<T> {

    @Override
    TypedScanRecords<T> poll(Duration timeout);
}


public class TypedScanRecords<T> extends ScanRecords {
    ...
        @Override
    public Iterator<TypedScanRecord> iterator() {
        ...
    }
}

public class TypedScanRecord<T> extends ScanRecord {
    private final T pojo;
}
```

fluss-client/src/main/java/org/apache/fluss/client/table/writer/Append.java
```diff
@@ -32,6 +32,9 @@ public interface Append {
     // TODO: Add more methods to configure the AppendWriter, such as apply static partitions,
     //  apply overwrites, etc.
 
-    /** Create a new {@link AppendWriter} to write data to a Log Table. */
-    AppendWriter createWriter();
+    /** Create a new {@link AppendWriter} to write data to a Log Table using InternalRow. */
+    AppendWriter<?> createWriter();
+
+    /** Create a new typed {@link AppendWriter} to write POJOs directly. */
+    <T> AppendWriter<T> createWriter(Class<T> pojoClass);
```
wuchong: ditto

fluss-client/src/main/java/org/apache/fluss/client/table/writer/Upsert.java
```diff
@@ -60,8 +57,11 @@ public interface Upsert {
...
-     * Create a new {@link UpsertWriter} with the optional {@link #partialUpdate(String...)}
-     * information to upsert and delete data to a Primary Key Table.
+     * Create a new {@link UpsertWriter} using {@code InternalRow} with the optional {@link
+     * #partialUpdate(String...)} information to upsert and delete data to a Primary Key Table.
      */
-    UpsertWriter createWriter();
+    UpsertWriter<?> createWriter();
+
+    /** Create a new typed {@link UpsertWriter} to write POJOs directly. */
+    <T> UpsertWriter<T> createWriter(Class<T> pojoClass);
```
wuchong: ditto

polyzos:
@wuchong thank you for all your comments. 
I have already introduced typed classes, like 

https://github.com/apache/fluss/pull/1992/changes#diff-463b3a6796c0042a681c569845f0abba145f454e7e40df447326bc212b544304

If you check the test i have also ensured backwards compatibility ... do you mean something different, or am I missing something? 
I have also tested things locally to make sure nothing breaks

wuchong:
@polyzos Yeah, I noticed those typed classes. However, typed classes are internal, only **interfaces are visible to users**.

My suggestion is to introduce **dedicated typed interfaces** (e.g., `interface TypedLookuper<T>`) that are **separate from the existing `Lookuper` interface**, rather than making `Lookuper` itself generic.

While turning `Lookuper` into a generic interface would be *binary-compatible*, it would introduce **type erasure warnings in IDEs** and clutter the public API. More importantly, for use cases involving `InternalRow`, the generic type `T` has **no semantic meaning**, so forcing a type parameter there adds unnecessary noise without benefit.

By keeping `Lookuper` and `TypedLookuper<T>` as distinct interfaces, we maintain clean separation of concerns:  
- `Lookuper` for low-level, type-agnostic access (e.g., `InternalRow`)  
- `TypedLookuper<T>` for high-level, type-safe lookups  

This approach preserves backward compatibility, avoids IDE warnings, and aligns with how users actually interact with the API.

polyzos:
@wuchong I made the required changes. 
Let me know if this approach resonates and works better

wuchong:
Thank you @polyzos , I will take another look.

fluss-client/src/main/java/org/apache/fluss/client/table/writer/TableAppend.java
```diff
@@ -38,4 +38,10 @@ public TableAppend(TablePath tablePath, TableInfo tableInfo, WriterClient writer
     public AppendWriter createWriter() {
         return new AppendWriterImpl(tablePath, tableInfo, writerClient);
     }
+
+    @Override
+    public <T> TypedAppendWriter<T> createTypedWriter(Class<T> pojoClass) {
+        AppendWriterImpl delegate = new AppendWriterImpl(tablePath, tableInfo, writerClient);
```
wuchong: Can simplify to just call `createWriter()`.

fluss-client/src/main/java/org/apache/fluss/client/table/writer/TableUpsert.java
```diff
@@ -95,4 +95,11 @@ public Upsert partialUpdate(String... targetColumnNames) {
     public UpsertWriter createWriter() {
         return new UpsertWriterImpl(tablePath, tableInfo, targetColumns, writerClient);
     }
+
+    @Override
+    public <T> TypedUpsertWriter<T> createTypedWriter(Class<T> pojoClass) {
+        UpsertWriterImpl delegate =
+                new UpsertWriterImpl(tablePath, tableInfo, targetColumns, writerClient);
```
wuchong: Can simplify to just call `createWriter()`.

fluss-client/src/main/java/org/apache/fluss/client/table/writer/TypedUpsertWriterImpl.java
```diff
@@ -0,0 +1,139 @@
...
+
+    @Override
+    public void close() throws Exception {
+        delegate.close();
+    }
+
+    private final Class<T> pojoClass;
+    private final TableInfo tableInfo;
+    private final RowType tableSchema;
+    private final int[] targetColumns; // may be null
```
wuchong: We can add `@Nullable` annotation to indicate it is nullable

fluss-client/src/main/java/org/apache/fluss/client/table/writer/TypedUpsertWriterImpl.java
```diff
@@ -0,0 +1,139 @@
...
+    }
+
+    private final UpsertWriterImpl delegate;
+
+    @Override
+    public void close() throws Exception {
+        delegate.close();
+    }
+
+    private final Class<T> pojoClass;
```
wuchong: Not used, can be removed.

fluss-client/src/test/java/org/apache/fluss/client/table/FlussTypedClientITCase.java
```diff
@@ -0,0 +1,452 @@
...
+
+        try (Table table = conn.getTable(path)) {
+            // write
+            TypedAppendWriter<AllTypesPojo> writer =
+                    table.newAppend().createTypedWriter(AllTypesPojo.class);
+            List<AllTypesPojo> expected = new ArrayList<>();
+            for (int i = 0; i < 5; i++) {
+                AllTypesPojo u = newAllTypesPojo(i);
+                expected.add(u);
+                writer.append(u).get();
```
wuchong:
```suggestion
                writer.append(u);
```

use async flush to reduce test time

fluss-client/src/test/java/org/apache/fluss/client/table/FlussTypedClientITCase.java
```diff
@@ -0,0 +1,452 @@
...
+            List<AllTypesPojo> values = new ArrayList<>();
+            while (values.size() < 4) { // INSERT 1, INSERT 2, UPDATE_BEFORE 1, UPDATE_AFTER 1
+                TypedScanRecords<AllTypesPojo> recs = scanner.poll(Duration.ofSeconds(2));
+                for (TypedScanRecord<AllTypesPojo> r : recs) {
+                    changes.add(r.getChangeType());
+                    values.add(r.getValue());
+                }
+            }
+            assertThat(changes)
+                    .contains(ChangeType.INSERT, ChangeType.UPDATE_BEFORE, ChangeType.UPDATE_AFTER);
```
wuchong:
```suggestion
 assertThat(changes)
                    .containsExactlyInAnyOrder(
                            ChangeType.INSERT,
                            ChangeType.INSERT,
                            ChangeType.UPDATE_BEFORE,
                            ChangeType.UPDATE_AFTER);
```

use `containsExactlyInAnyOrder` to verify all the changes

fluss-client/src/main/java/org/apache/fluss/client/table/writer/TypedAppendWriterImpl.java
```diff
@@ -0,0 +1,63 @@
...
+
+/**
+ * A typed {@link AppendWriter} that converts POJOs to {@link InternalRow} and delegates to the
+ * existing internal-row based writer implementation.
+ */
+class TypedAppendWriterImpl<T> implements TypedAppendWriter<T> {
+
+    private final AppendWriterImpl delegate;
+    private final Class<T> pojoClass;
+    private final RowType tableSchema;
```
wuchong: not used, can be removed.

fluss-client/src/main/java/org/apache/fluss/client/table/writer/TypedUpsertWriterImpl.java
```diff
@@ -0,0 +1,139 @@
...
+        if (projection == tableSchema) {
+            return projected;
+        }
+        // expand projected row to full row if needed
+        GenericRow full = new GenericRow(tableSchema.getFieldCount());
+        if (forDelete) {
+            // set PK fields, others null
+            for (String pk : tableInfo.getPhysicalPrimaryKeys()) {
+                int projIndex = projection.getFieldIndex(pk);
+                int fullIndex = tableSchema.getFieldIndex(pk);
```
wuchong, line 119: Could you please add a `TODO` comment above this code block noting that it can be optimized by pre-computing the index mapping in the constructor?

fluss-client/src/test/java/org/apache/fluss/client/table/FlussTypedClientITCase.java
```diff
@@ -0,0 +1,452 @@
...
+            // primary key lookup using Lookuper API with POJO key
+            TypedLookuper<PLookupKey> lookuper =
+                    table.newLookup().createTypedLookuper(PLookupKey.class);
+            RowType tableSchema = table.getTableInfo().getRowType();
+            RowToPojoConverter<AllTypesPojo> rowConv =
+                    RowToPojoConverter.of(AllTypesPojo.class, tableSchema, tableSchema);
+
+            LookupResult lr = lookuper.lookup(new PLookupKey(1)).get();
+            AllTypesPojo one = rowConv.fromRow(lr.getSingletonRow());
+            assertThat(one.str).isEqualTo("s1");
```
wuchong:
```
assertThat(one).isEqualTo(newAllTypesPojo(1));
```

After adding `equals` and `hashcode` method to the `AllTypesPojo` class, we can simply assert the full record, this can check the full POJO record deserialization.

fluss-client/src/test/java/org/apache/fluss/client/table/FlussTypedClientITCase.java
```diff
@@ -0,0 +1,452 @@
...
+            // verify via lookup and scan using Lookuper + POJO key
+            TypedLookuper<PLookupKey> lookuper =
+                    table.newLookup().createTypedLookuper(PLookupKey.class);
+            RowType tableSchema = table.getTableInfo().getRowType();
+            RowToPojoConverter<AllTypesPojo> rowConv =
+                    RowToPojoConverter.of(AllTypesPojo.class, tableSchema, tableSchema);
+            AllTypesPojo lookedUp =
+                    rowConv.fromRow(lookuper.lookup(new PLookupKey(1)).get().getSingletonRow());
+            assertThat(lookedUp.str).isEqualTo("second");
+            assertThat(lookedUp.dec).isEqualByComparingTo("99.99");
```
wuchong: In order to test the partial update feature, we should assert the other fields are keep unchanged here.

fluss-client/src/test/java/org/apache/fluss/client/table/FlussTypedClientITCase.java
```diff
@@ -0,0 +1,452 @@
...
+
+            TypedLogScanner<AllTypesPojo> scanner =
+                    table.newScan().createTypedLogScanner(AllTypesPojo.class);
+            subscribeFromBeginning(scanner, table);
+            boolean sawUpdateAfter = false;
+            while (!sawUpdateAfter) {
+                TypedScanRecords<AllTypesPojo> recs = scanner.poll(Duration.ofSeconds(2));
+                for (TypedScanRecord<AllTypesPojo> r : recs) {
+                    if (r.getChangeType() == ChangeType.UPDATE_AFTER) {
+                        assertThat(r.getValue().str).isEqualTo("second");
```
wuchong: ditto

fluss-client/src/main/java/org/apache/fluss/client/table/scanner/log/ScanRecords.java
```diff
@@ -37,7 +37,9 @@
  */
 @PublicEvolving
 public class ScanRecords implements Iterable<ScanRecord> {
-    public static final ScanRecords EMPTY = new ScanRecords(Collections.emptyMap());
+    public static final ScanRecords empty() {
+        return new ScanRecords(Collections.emptyMap());
+    }
```
wuchong: Not necessary? I still prefer the preivous implementation because it can avoid some small object overhead (GC).

fluss-client/src/main/java/org/apache/fluss/client/table/writer/TableWriter.java
```diff
@@ -38,4 +38,9 @@ public interface TableWriter {
      * results in an error.
      */
     void flush();
+
+    @Override
+    default void close() throws Exception {
+        // by default do nothing
+    }
```
wuchong: It seems the newly introduced `close()` method doesn’t currently have any meaningful work to perform, as the writers don’t hold any resources at this stage. I suggest holding off on introducing it for now.  

Typically, a `close()` method should either **flush pending records** or **ensure all previously submitted requests have completed**—otherwise, it may mislead users into expecting cleanup or finalization behavior that isn’t actually implemented. This introduces some complex to this PR.

PR #1995, opened 2025-11-18
[lake] Tolerate Paimon lake table existent with different insignificant options

### LINKED ISSUE #1994: [lake] Paimon tables support compatibility when non-critical options are different
### Search before asking

### Motivation

In https://github.com/apache/fluss/pull/1847 Fluss support to tolerate exist Paimon lake table when create Fluss table. But in production environment, datalake manager systems may add some extra options to lake table to help manage. Such as `owner`, `createdAt`, `updatedAt`... 

These options do not affect table compatibility and should be ignored when comparing tables for a match.

### Solution

_No response_

### Anything else?

_No response_

### Willingness to contribute

LiebingYu:
### Purpose

Linked issue: close #1994

### DIFF
--- a/fluss-lake/fluss-lake-paimon/src/main/java/org/apache/fluss/lake/paimon/utils/PaimonTableValidation.java
+++ b/fluss-lake/fluss-lake-paimon/src/main/java/org/apache/fluss/lake/paimon/utils/PaimonTableValidation.java
@@ -1,0 +1,112 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.fluss.lake.paimon.utils;
+
+import org.apache.fluss.exception.TableAlreadyExistException;
+
+import org.apache.paimon.CoreOptions;
+import org.apache.paimon.catalog.Identifier;
+import org.apache.paimon.options.ConfigOption;
+import org.apache.paimon.schema.Schema;
+import org.apache.paimon.table.FileStoreTable;
+
+import java.lang.reflect.Field;
+import java.util.HashMap;
+import java.util.Map;
+
+import static org.apache.fluss.lake.paimon.utils.PaimonConversions.FLUSS_CONF_PREFIX;
+
+/** Utils to verify whether the existing Paimon table is compatible with the table to be created. */
+public class PaimonTableValidation {
+
+    private static final Map<String, ConfigOption<?>> PAIMON_CONFIGS = extractPaimonConfigs();
+
+    public static void validatePaimonSchemaCompatible(
+            Identifier tablePath, Schema existingSchema, Schema newSchema) {
+        // Adjust options for comparison
+        Map<String, String> existingOptions = existingSchema.options();
+        Map<String, String> newOptions = newSchema.options();
+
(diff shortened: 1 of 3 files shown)


fluss-server/src/main/java/org/apache/fluss/server/tablet/TabletService.java
```diff
@@ -282,6 +282,12 @@ public CompletableFuture<LimitScanResponse> limitScan(LimitScanRequest request)
     @Override
     public CompletableFuture<NotifyLeaderAndIsrResponse> notifyLeaderAndIsr(
             NotifyLeaderAndIsrRequest notifyLeaderAndIsrRequest) {
+        // This method should only be called by Coordinator internally
+        if (!currentSession().isInternal()) {
```
luoyuxia: why change this?

fluss-lake/fluss-lake-paimon/src/main/java/org/apache/fluss/lake/paimon/utils/PaimonTableValidation.java
```diff
@@ -0,0 +1,112 @@
...
+import java.util.Map;
+
+import static org.apache.fluss.lake.paimon.utils.PaimonConversions.FLUSS_CONF_PREFIX;
+
+/** Utils to verify whether the existing Paimon table is compatible with the table to be created. */
+public class PaimonTableValidation {
+
+    private static final Map<String, ConfigOption<?>> PAIMON_CONFIGS = extractPaimonConfigs();
+
+    public static void validatePaimonSchemaCapability(
```
luoyuxia:
```suggestion
    public static void validatePaimonSchemaCompatible(
```

fluss-lake/fluss-lake-paimon/src/main/java/org/apache/fluss/lake/paimon/utils/PaimonTableValidation.java
```diff
@@ -0,0 +1,112 @@
...
+            existingOptions.remove(datalakeConfigKey);
+            newOptions.remove(datalakeConfigKey);
+        }
+
+        // remove changeable options
+        removeChangeableOptions(existingOptions);
+        removeChangeableOptions(newOptions);
+
+        // ignore the existing options that are not in new options
+        existingOptions.entrySet().removeIf(entry -> !newOptions.containsKey(entry.getKey()));
```
luoyuxia, line 60: why need this line code? Could you please give me an example?

fluss-lake/fluss-lake-paimon/src/test/java/org/apache/fluss/lake/paimon/LakeEnabledTableCreateITCase.java
```diff
@@ -476,6 +477,54 @@ void testCreateLakeEnableTableWithExistLakeTable() throws Exception {
...
+                .isInstanceOf(LakeTableAlreadyExistException.class)
+                .hasMessageContaining(
+                        "The table `fluss`.`log_table_with_exist_lake_table` already exists in Paimon catalog, "
+                                + "but the table schema is not compatible.");
+
+        // reset fluss.k1 in Paimon
+        SchemaChange schemaChange3 = SchemaChange.setOption("fluss.k1", "v1");
+        paimonCatalog.alterTable(paimonTablePath, Collections.singletonList(schemaChange3), false);
+
+        // add a new Paimon option (not specified in the Fluss table) to Paimon table will be ok
```
luoyuxia, line 504: shouldn't modify a  Paimon option also throw exception? Imagine you modify the partition bucket key.

LiebingYu: These lines only change the `fluss.k1` back to `v1`. Otherwise the following code cannot work.

fluss-lake/fluss-lake-paimon/src/main/java/org/apache/fluss/lake/paimon/utils/PaimonTableValidation.java
LiebingYu: For example, the data lake manager may add a default value for `snapshot.num-retained.min`.
1. user create a Fluss table with lake enable, say `t1`, Fluss will call Paimon catalog to create a Paimon table with `snapshot.num-retained.min = 2`;
2. user drop Fluss table `t1`, the Paimon table should still exist;
3. user create `t1` with same DDL again (no `paimon.snapshot.num-retained.min` option), we can ignore `snapshot.num-retained.min` here, let it keep the default value filled by the first creation.

PR #1996, opened 2025-11-18
mask accessKeyId in logs using MaskedValue

### LINKED ISSUE #1989: [FileSystem] Mask user secret in the S3DelegationTokenProvider
### Search before asking

### Description

Currently, when connecting to filesystems like S3, users add their secrets. However, currently these secrets are exposed in the logs, which might be undesirable for security reasons.
Image

We could implement a simple masking mechanism, i.e if my access key is AS7124FSDFER, only expose some initial letters and mask the others, i.e, AS7*******.

This way, users can both verify the right access key is used, but it's also not exposed as a whole.

### Willingness to contribute

Pikady:
### Purpose

- Linked issue: close #1989

- Purpose: Prevent leaking S3 accessKeyId in logs by introducing a masking utility and applying it to S3 token provider/receiver logging.
 
### Brief change log

- Add org.apache.fluss.utils.MaskedValue for safe logging of sensitive strings in fluss-common.
- Apply MaskedValue.of(...) in S3DelegationTokenProvider and S3DelegationTokenReceiver log statements to mask accessKeyId .
- Add unit tests for the masking utility and the S3 receiver masking path

### Tests

- fluss-common :
  - MaskedValueUtilsTest : verifies prefix visibility, null/empty handling, equals semantics.
- fluss-filesystems/fluss-fs-s3 :
  - S3DelegationTokenReceiverMaskingTest : feeds a fake token to the receiver, asserts masked display of accessKeyId without real S3 connection.


polyzos:
@Pikady, I think that adding a new class and this functionality might be a bit of overkill and create unnecessary overhead. I think before logging, we can just use a function on the string to achieve the masking.
Maybe just extract the first 3 numbers and append a few stars. WDYT?

Pikady:
Makes sense. I'll remove the MaskedValue class and use a simple helper method instead. Will push the changes soon.

PR #2010, opened 2025-11-24
Support Add Column in Fluss.

### LINKED ISSUE #2056: Fluss Support Add column
### Search before asking

### Motivation

Currently, we can not change fluss table's schema, which is not acceptable in production. Thus, I hope fluss can support add column at least. 

### Solution

_No response_

### Anything else?

_No response_

### Willingness to contribute

loserwang1024:
### Purpose

Linked issue: close  #2056


fluss-rpc/src/main/proto/FlussApi.proto
```diff
@@ -915,6 +917,16 @@ message PbAlterConfig {
...
 }
 
+message PbAlterColumn{
+  required int32 alter_type = 1; // ADD Column=0, DROP COLUMN =1, ALTER COLUMN=2
+  optional string modified_column_name = 2;
+  // ColumnPosition
+  optional bytes column_json = 3;
+  optional int32 column_position_type = 4; // LAST=0,FIRST=1,AFTER=3
+  optional string column_after_name = 5;
+}
```
wuchong: This mixes different alter column messages into one structure which looks very complex (don't know which field is required at which type).  

Can we just introduce `PbAddColumn`, `PbRenameColumn`, `PbDropColumn`, etc...? This makes each operation clear. And they can be directly put in `AlterTableRequest` as `repeated` fields.

fluss-rpc/src/main/proto/FlussApi.proto
```diff
@@ -113,6 +113,8 @@ message AlterTableRequest {
   required PbTablePath table_path = 1;
   required bool ignore_if_not_exists = 2;
   repeated PbAlterConfig config_changes = 3;
+  optional int32 schema_id = 4;
```
wuchong: schema id is not necessary, the lastest schema id should be determined by sever.

fluss-server/src/main/java/org/apache/fluss/server/coordinator/MetadataManager.java
```diff
@@ -323,59 +325,86 @@ public long createTable(
...
+            if (!schemaChanges.isEmpty()) {
+                SchemaUpdate schemaUpdate = new SchemaUpdate(tableInfo);
+                for (TableChange.SchemaChange schemaChange : schemaChanges) {
+                    schemaUpdate.applySchemaChange(schemaChange);
+                }
+                newSchema = schemaUpdate.getSchema();
+                if (schemaId == null) {
+                    throw new InvalidTableException("Can't alter table schema without schema id");
+                }
+            }
```
wuchong: Let's have a separate method `addColumn` to handle the schema changes. Mixing them together makes this logic very complex and error-prone.

fluss-client/src/main/java/org/apache/fluss/client/admin/FlussAdmin.java
```diff
@@ -248,19 +249,44 @@ public CompletableFuture<Void> createTable(
...
-                        .collect(Collectors.toList());
-
-        request.addAllConfigChanges(pbFlussTableChanges)
-                .setIgnoreIfNotExists(ignoreIfNotExists)
-                .setTablePath()
-                .setDatabaseName(tablePath.getDatabaseName())
-                .setTableName(tablePath.getTableName());
-        return gateway.alterTable(request).thenApply(r -> null);
+        CompletableFuture<Void> future = new CompletableFuture<>();
+        CompletableFuture.runAsync(
```
wuchong: why run in async mode? We should avoid use `CompletableFuture.runAsync`, because it users a global thread pool which is out of management.

fluss-common/src/main/java/org/apache/fluss/metadata/Schema.java
```diff
@@ -331,19 +411,27 @@ public Schema build() {
...
+        public static final short UNKNOWN_COLUMN_ID = -1;
         private static final long serialVersionUID = 1L;
+        private final short columnId;
         private final String columnName;
         private final DataType dataType;
         private final @Nullable String comment;
 
         public Column(String columnName, DataType dataType) {
-            this(columnName, dataType, null);
+            this(columnName, dataType, null, UNKNOWN_COLUMN_ID);
```
wuchong, line 398: We should be able to get the column id for the previous schema (use the column order). Using -1 as default column is error-prone.

fluss-server/src/main/java/org/apache/fluss/server/zk/ZooKeeperClient.java
```diff
@@ -810,16 +810,18 @@ public void registerPartitionAssignmentAndMetadata(
 
     /** Register schema to ZK metadata and return the schema id. */
     public int registerSchema(TablePath tablePath, Schema schema) throws Exception {
-        int currentSchemaId = getCurrentSchemaId(tablePath);
+        return registerSchema(tablePath, schema, getCurrentSchemaId(tablePath) + 1);
```
wuchong: This is creating a new table, we can directly use schema id = 1 here?

fluss-common/src/main/java/org/apache/fluss/metadata/Schema.java
loserwang1024: Yes, I also want to do it. But as a public api, I cannot remove this constructor ` Column(String columnName, DataType dataType)`. If not set -1, I have no idea how to handle it.

fluss-server/src/main/java/org/apache/fluss/server/zk/ZooKeeperClient.java
loserwang1024: image

sometime schema will be recreated when create table.

fluss-client/src/main/java/org/apache/fluss/client/table/scanner/batch/KvSnapshotBatchScanner.java
```diff
@@ -113,16 +121,6 @@ public KvSnapshotBatchScanner(
...
-    /**
-     * Fetch data from snapshot.
-     *
-     * <p>If the snapshot file reader is not ready in given maximum block time, return an empty
-     * iterator. If the reader is ready, always return the reader if there remains any record in the
-     * reader, otherwise, return null.
-     *
-     * @param timeout The maximum time to block (must not be greater than {@link Long#MAX_VALUE}
-     *     milliseconds)
-     */
```
wuchong, line 125: why removes the javadoc?

fluss-common/src/main/java/org/apache/fluss/metadata/TableChange.java
```diff
@@ -126,4 +213,196 @@ public String toString() {
...
+        private First() {}
+
+        @Override
+        public String toString() {
+            return "FIRST";
+        }
+    }
+
+    /** Column position LAST means the specified column should be the last column. */
+    final class LAST implements ColumnPosition {
```
wuchong: Use  camel-case `First` to keep in alignment with others

fluss-common/src/main/java/org/apache/fluss/metadata/TableChange.java
```diff
@@ -126,4 +213,196 @@ public String toString() {
...
+
+        public String getName() {
+            return name;
+        }
+    }
+
+    /** A table change to modify a column. */
+    class ModifyColumn implements SchemaChange {
+        private final String name;
+        private DataType dataType;
```
wuchong: can be `final`

fluss-common/src/main/java/org/apache/fluss/metadata/TableInfo.java
```diff
@@ -335,6 +335,23 @@ public static TableInfo of(
                 modifiedTime);
     }
 
+    /** Replace a TableInfo with a new SchemaInfo. */
+    public static TableInfo of(TableInfo tableInfo, SchemaInfo schemaInfo) {
```
wuchong: Provide a `public TableInfo withNewSchema(SchemaInfo schemaInfo)` on the `TableInfo` class would be more convenient?

fluss-common/src/main/java/org/apache/fluss/record/LogRecordBatch.java
```diff
@@ -165,6 +166,8 @@ default boolean hasWriterId() {
     /** The read context of a {@link LogRecordBatch} to read records. */
     interface ReadContext {
 
+        boolean isSchemaChange(int schemaId);
```
wuchong: `isSchemaChanged` and add javadocs

fluss-common/src/main/java/org/apache/fluss/record/LogRecordReadContext.java
```diff
@@ -193,17 +218,28 @@ public boolean isProjectionPushDowned() {
...
-                schemaId == this.schemaId,
-                "The schemaId (%s) in the record batch is not the same as the context (%s).",
-                schemaId,
-                this.schemaId);
+        //        checkArgument(
+        //                schemaId == this.schemaId,
+        //                "The schemaId (%s) in the record batch is not the same as the context
+        // (%s).",
+        //                schemaId,
+        //                this.schemaId);
```
wuchong: remove this

fluss-common/src/main/java/org/apache/fluss/row/encode/ValueDecoder.java
```diff
@@ -55,7 +73,7 @@ public static class Value {
         public final short schemaId;
         public final BinaryRow row;
 
-        private Value(short schemaId, BinaryRow row) {
+        public Value(short schemaId, BinaryRow row) {
```
wuchong: only used in `ValueDecoder`, keep `private`.

fluss-common/src/main/java/org/apache/fluss/row/PruneRow.java
```diff
@@ -0,0 +1,84 @@
...
+import java.util.Arrays;
+import java.util.Objects;
+
+import static org.apache.fluss.utils.SchemaUtil.getIndexMapping;
+
+/**
+ * A specific implementation of {@link ProjectedRow} which prunes the columns from the origin schema
+ * to the target schema.
+ */
+public class PruneRow extends ProjectedRow {
```
wuchong: remove this class and move the `isNullAt` logic into `ProjectedRow`.

fluss-client/src/main/java/org/apache/fluss/client/FlussConnection.java
```diff
@@ -104,6 +105,14 @@ public Table getTable(TablePath tablePath) {
         return new FlussTable(this, tablePath, tableInfo);
     }
 
+    @Override
+    public Table getTable(TablePath tablePath, SchemaInfo schemaInfo) {
```
wuchong: As we discussed, we don't need this now, and just using `souceOutputType` in `FlinkSource` is enough.

fluss-common/src/main/java/org/apache/fluss/record/DefaultLogRecordBatch.java
```diff
@@ -217,18 +218,49 @@ public CloseableIterator<LogRecord> records(ReadContext context) {
...
+            @Override
+            public LogRecord next() {
+                LogRecord next = records.next();
+                return new GenericRecord(
+                        next.logOffset(),
+                        next.timestamp(),
+                        next.getChangeType(),
+                        pruneRow.replaceRow(next.getRow()));
+            }
+        };
```
wuchong: We can add two new methods in `org.apache.fluss.record.LogRecordBatch.ReadContext`, and then remove `isSchemaChange`, and `getPruneRow`. 

```java
        /**
         * If the read context defines an output projection (for example, log records may add new
         * columns or reorder columns, but reader need a static schema for the output rows), return
         * a {@link ProjectedRow} that describes the projected output row for the given schemaId.
         * The returned object is used by readers to transform or materialize rows according to the
         * output projection. Returns {@code null} if no output projection is configured.
         *
         * @param schemaId the current row schema id
         * @return a {@link ProjectedRow} describing the output projection, or {@code null} if none
         */
        @Nullable
        ProjectedRow getOutputProjectedRow(int schemaId);
```

And here we can pass the nullable output ProjectedRow into `columnRecordIterator` and `rowRecordIterator` for final projection. And we don't need to wrap an iterator again here.

fluss-common/src/main/java/org/apache/fluss/record/FileLogProjection.java
```diff
@@ -434,14 +398,130 @@ ByteBuffer getLogHeaderBuffer() {
...
+            selectedFieldPositions[i] = position;
+            if (position < prev) {
+                throw new InvalidColumnProjectionException(
+                        "The projection indexes should be in field order, but is "
+                                + Arrays.toString(projectedFields));
+            }
+
+            prev = position;
+        }
+        return selectedFieldPositions;
```
wuchong, line 491: We can remove this for now, as it may affect projection performance, and we don't need the remapping, as we only support add column add end. We can rename the method into `toSelectedFieldPositions()`.

fluss-rpc/src/main/proto/FlussApi.proto
```diff
@@ -915,6 +918,32 @@ message PbAlterConfig {
...
+
+message PbRenameColumn{
+  required string old_column_name = 1;
+  required string new_column_name = 2;
+}
+
+
+message PbModifyColumn{
+  required string column_name = 1;
+  required bytes data_type_json = 2;
```
wuchong: should be `optional`. The following alter column is also valid. 


```
ALTER TABLE prod.db.sample ALTER COLUMN measurement COMMENT 'unit is kilobytes per second';
```

fluss-server/src/main/java/org/apache/fluss/server/coordinator/event/watcher/TableChangeWatcher.java
```diff
@@ -231,4 +242,31 @@ private void processCreatePartition(
...
+        try {
+            int currentSchemaId = zooKeeperClient.getCurrentSchemaId(tablePath);
+            SchemaInfo schemaInfo;
+            if (schemaId != currentSchemaId) {
+                LOG.warn(
+                        "Schema id {} is not equal to current schema id {}. Skipping schema change processing.",
+                        schemaId,
+                        currentSchemaId);
+                return;
+            }
```
wuchong: Why check current schema id? This is a heavy operation. Even if this is an old schema, I think it is still fine to process the schema.

PR #2020, opened 2025-11-26
[server] Disable shutdown hook for log4j to ensure Fluss log in shutdown hook can be collected

### LINKED ISSUE #2019: Collect log in shutdown hook when useing log4j2
### Search before asking

### Description

Log4j2 registers a shutdown hook to perform cleanup operations. However, in the JVM, the execution order of multiple shutdown hooks is not guaranteed. This may cause Log4j2's shutdown hook to run before Fluss's shutdown hook, resulting in logs emitted during Fluss's shutdown phase not being captured.

We can disable Log4j2's shutdown hook and instead shut it down manually using reflection.

### Willingness to contribute

LiebingYu:
### Purpose

Linked issue: close #2019

### DIFF
--- a/fluss-server/src/main/java/org/apache/fluss/server/utils/LogShutdownUtil.java
+++ b/fluss-server/src/main/java/org/apache/fluss/server/utils/LogShutdownUtil.java
@@ -17,8 +17,13 @@
 
 package org.apache.fluss.server.utils;
 
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
 /** Utility class to shut down log system. */
 public class LogShutdownUtil {
+
+    private static final Logger LOG = LoggerFactory.getLogger(LogShutdownUtil.class);
 
     public static void shutdownLogIfPossible() {
         // To avoid binding to a specific logging implementation, we use reflection here.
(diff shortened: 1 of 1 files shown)


fluss-server/src/main/java/org/apache/fluss/server/utils/LogShutdownUtil.java
```diff
@@ -0,0 +1,38 @@
...
+            // To ensure that logs within the JVM shutdown hook can be printed when using Log4j2,
+            // Fluss has disabled Log4j2's shutdown hook; therefore, manual shutdown is required
+            // here.
+            Class<?> logManager = Class.forName("org.apache.logging.log4j.LogManager");
+            logManager.getMethod("shutdown").invoke(null);
+        } catch (ClassNotFoundException e) {
+            System.err.println("Class org.apache.logging.log4j.LogManager not found");
+        } catch (Exception e) {
+            System.err.println(
+                    "Error to invoke shutdown method of org.apache.logging.log4j.LogManager");
```
wuchong: Use Logger to print the error messages, otherwise, if the reflection is failed, we can't find the error log (the `System.err` is not redirect to log files).

PR #2024, opened 2025-11-26
[client]Support initial cluster with retries to avoid initial failed when bootstrap server only contains one available address

### LINKED ISSUE #2014: FlussConnection getTable method fails due to missing retries when fetching metadata
### Search before asking

### Fluss version

0.8.0 (latest release)

### Please describe the bug 🐞

FlussConnection getTable method fails due to missing retries when fetching metadata：

```
java.lang.IllegalArgumentException: table: xxx not found in cluster
	at org.apache.fluss.cluster.Cluster.lambda$getTableOrElseThrow$2(Cluster.java:260)
	at org.apache.fluss.cluster.Cluster.getTableOrElseThrow(Cluster.java:257)
	at org.apache.fluss.client.metadata.MetadataUpdater.getTableInfoOrElseThrow(MetadataUpdater.java:100)
	at org.apache.fluss.client.FlussConnection.getTable(FlussConnection.java:103)
	at org.apache.fluss.flink.sink.writer.FlinkSinkWriter.initialize(FlinkSinkWriter.java:112)
	at org.apache.fluss.flink.sink.writer.AppendSinkWriter.initialize(AppendSinkWriter.java:53)
	at org.apache.fluss.flink.sink.FlinkSink.createWriter(FlinkSink.java:63)
	at org.apache.flink.streaming.runtime.operators.sink.StatelessSinkWriterStateHandler.createWriter(StatelessSinkWriterStateHandler.java:39)
	at org.apache.flink.streaming.runtime.operators.sink.SinkWriterOperator.initializeState(SinkWriterOperator.java:170)
	at org.apache.flink.streaming.api.operators.StreamOperatorStateHandler.initializeOperatorState(StreamOperatorStateHandler.java:122)
	at org.apache.flink.streaming.api.operators.AbstractStreamOperator.initializeState(AbstractStreamOperator.java:285)
	at org.apache.flink.streaming.runtime.tasks.RegularOperatorChain.initializeStateAndOpenOperators(RegularOperatorChain.java:106)
	at org.apache.flink.streaming.runtime.tasks.StreamTask.restoreGates(StreamTask.java:843)
	at org.apache.flink.streaming.runtime.tasks.StreamTaskActionExecutor$1.call(StreamTaskActionExecutor.java:55)
	at org.apache.flink.streaming.runtime.tasks.StreamTask.restoreInternal(StreamTask.java:790)
	at org.apache.flink.streaming.runtime.tasks.StreamTask.restore(StreamTask.java:747)
	at org.apache.flink.runtime.taskmanager.Task.runWithSystemExitMonitoring(Task.java:959)
	at org.apache.flink.runtime.taskmanager.Task.restoreAndInvoke(Task.java:928)
	at org.apache.flink.runtime.taskmanager.Task.doRun(Task.java:751)
	at org.apache.flink.runtime.taskmanager.Task.run(Task.java:567)
	at java.lang.Thread.run(Thread.java:879)
```

### Solution

_No response_

swuferhong:
### Purpose

Linked issue: close #2014 

Support initial cluster with retries to avoid initial failed when bootstrap server only contains one available address.  Currently, retries are only applied to errors on the allowlist: `StaleMetadataException` and `NetworkException`, as these two types of errors may occur during machine upgrades.

### DIFF
--- a/fluss-client/src/main/java/org/apache/fluss/client/metadata/MetadataUpdater.java
+++ b/fluss-client/src/main/java/org/apache/fluss/client/metadata/MetadataUpdater.java
@@ -63,7 +63,7 @@ import static org.apache.fluss.utils.ExceptionUtils.stripExecutionException;
 public class MetadataUpdater {
     private static final Logger LOG = LoggerFactory.getLogger(MetadataUpdater.class);
 
-    private static final int MAX_RETRY_TIMES = 5;
+    private static final int MAX_RETRY_TIMES = 3;
     private static final int RETRY_INTERVAL_MS = 100;
 
     private final RpcClient rpcClient;
@@ -295,16 +295,18 @@ public class MetadataUpdater {
         Cluster cluster = null;
         Exception lastException = null;
         for (InetSocketAddress address : inetSocketAddresses) {
+            ServerNode serverNode = null;
             try {
-                ServerNode serverNode =
+                serverNode =
                         new ServerNode(
                                 -1,
                                 address.getHostString(),
                                 address.getPort(),
                                 ServerType.COORDINATOR);
+                ServerNode finalServerNode = serverNode;
                 AdminReadOnlyGateway adminReadOnlyGateway =
                         GatewayClientProxy.createGatewayProxy(
-                                () -> serverNode, rpcClient, AdminReadOnlyGateway.class);
+                                () -> finalServerNode, rpcClient, AdminReadOnlyGateway.class);
                 if (inetSocketAddresses.size() == 1) {
                     // if there is only one bootstrap server, we can retry to connect to it.
                     cluster =
@@ -315,6 +317,11 @@ public class MetadataUpdater {
                     break;
                 }
             } catch (Exception e) {
+                // We should dis-connected with the bootstrap server id to make sure the next
+                // retry can rebuild the connection.
+                if (serverNode != null) {
+                    rpcClient.disconnect(serverNode.uid());
+                }
                 LOG.error(
                         "Failed to initialize fluss client connection to bootstrap server: {}",
                         address,
@@ -325,9 +332,10 @@ public class MetadataUpdater {
 
         if (cluster == null && lastException != null) {
             String errorMsg =
-                    "Failed to initialize fluss client connection to server because no "
-                            + "bootstrap server is validate. bootstrap servers: "
-                            + inetSocketAddresses;
+                    "Failed to initialize fluss client connection to bootstrap servers: "
+                            + inetSocketAddresses
+                            + ". \nReason: "
+                            + lastException.getMessage();
             LOG.error(errorMsg);
             throw new IllegalStateException(errorMsg, lastException);
         }
@@ -348,18 +356,10 @@ public class MetadataUpdater {
                 return tryToInitializeCluster(gateway);
             } catch (Exception e) {
                 Throwable cause = stripExecutionException(e);
-                if (!(cause instanceof StaleMetadataException
-                        || cause instanceof NetworkException)) {
+                // in case of bootstrap is recovering, we should retry to connect.
+                if (!(cause instanceof StaleMetadataException || cause instanceof NetworkException)
+                        || retryCount >= maxRetryTimes) {
                     throw e;
-                }
-
-                if (retryCount >= maxRetryTimes) {
-                    LOG.warn(
-                            "Max retries ({}) exceeded to connect to {}. ",
-                            maxRetryTimes,
-                            serverNode,
-                            e);
-                    return null;
                 }
 
                 // We should dis-connected with the bootstrap server id to make sure the next
--- a/fluss-client/src/test/java/org/apache/fluss/client/metadata/MetadataUpdaterTest.java
+++ b/fluss-client/src/test/java/org/apache/fluss/client/metadata/MetadataUpdaterTest.java
@@ -36,6 +36,7 @@ import java.util.concurrent.CompletableFuture;
 
 import static org.apache.fluss.server.utils.ServerRpcMessageUtils.buildMetadataResponse;
 import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
 
 /** UT Test for update metadata of {@link MetadataUpdater}. */
 public class MetadataUpdaterTest {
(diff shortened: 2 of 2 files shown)


fluss-client/src/main/java/org/apache/fluss/client/metadata/MetadataUpdater.java
```diff
@@ -293,7 +296,18 @@ private static Cluster initializeCluster(Configuration conf, RpcClient rpcClient
         Exception lastException = null;
         for (InetSocketAddress address : inetSocketAddresses) {
             try {
-                cluster = tryToInitializeCluster(rpcClient, address);
+                ServerNode serverNode =
+                        new ServerNode(
+                                -1,
```
wuchong, line 302: We should `.disconnect()` with the bootstrap server id, otherwise, the second retry will still use the old server host. If the bootstrap server is a SLB, this makes the SLB doesn't work.

fluss-client/src/main/java/org/apache/fluss/client/metadata/MetadataUpdater.java
```diff
@@ -316,14 +330,54 @@ private static Cluster initializeCluster(Configuration conf, RpcClient rpcClient
...
-    private static Cluster tryToInitializeCluster(RpcClient rpcClient, InetSocketAddress address)
+    @VisibleForTesting
+    static @Nullable Cluster tryToInitializeClusterWithRetries(
+            InetSocketAddress address, AdminReadOnlyGateway gateway, int maxRetryTimes)
+            throws Exception {
+        int retryCount = 0;
+        while (retryCount <= maxRetryTimes) {
+            try {
+                return tryToInitializeCluster(gateway);
+            } catch (Exception e) {
```
wuchong, line 357: Instead of retrying only the first bootstrap server, the client should attempt to connect to all provided bootstrap servers before sleeping and retrying. For example, if a user specifies three bootstrap servers and only the first one is unavailable, the client should immediately try the second (or third) server—rather than repeatedly retrying the failed one—improving connection resilience and startup latency.

swuferhong: Agreed. I have change this to two separate code paths: the retry logic is only applied when there is exactly one bootstrap server address.

swuferhong:
@wuchong comments addressed.

fluss-client/src/main/java/org/apache/fluss/client/metadata/MetadataUpdater.java
```diff
@@ -316,14 +335,61 @@ private static Cluster initializeCluster(Configuration conf, RpcClient rpcClient
...
+                    throw e;
+                }
+
+                if (retryCount >= maxRetryTimes) {
+                    LOG.warn(
+                            "Max retries ({}) exceeded to connect to {}. ",
+                            maxRetryTimes,
+                            serverNode,
+                            e);
+                    return null;
```
wuchong: We don't need to logging, we should throw the exception and let the parent method catch it. Otherwise, the root exception is swallowed in the exception message.

fluss-client/src/main/java/org/apache/fluss/client/metadata/MetadataUpdater.java
```diff
@@ -316,14 +335,61 @@ private static Cluster initializeCluster(Configuration conf, RpcClient rpcClient
...
+            try {
+                return tryToInitializeCluster(gateway);
+            } catch (Exception e) {
+                Throwable cause = stripExecutionException(e);
+                if (!(cause instanceof StaleMetadataException
+                        || cause instanceof NetworkException)) {
+                    throw e;
+                }
+
+                if (retryCount >= maxRetryTimes) {
```
wuchong: Add comment above `// in case of bootstrap is recovering, we should retry to connect`

fluss-client/src/main/java/org/apache/fluss/client/metadata/MetadataUpdater.java
```diff
@@ -293,8 +296,24 @@ private static Cluster initializeCluster(Configuration conf, RpcClient rpcClient
...
+                    cluster =
+                            tryToInitializeClusterWithRetries(
+                                    rpcClient, serverNode, adminReadOnlyGateway, MAX_RETRY_TIMES);
+                } else {
+                    cluster = tryToInitializeCluster(adminReadOnlyGateway);
+                    break;
+                }
             } catch (Exception e) {
                 LOG.error(
                         "Failed to initialize fluss client connection to bootstrap server: {}",
```
wuchong: We should improve this exception message, it didn't carry exception message to the top error message which may confuse users. 

```
            String errorMsg =
                    "Failed to initialize fluss client connection to bootstrap servers: "
                            + inetSocketAddresses
                            + ". \nReason: "
                            + lastException.getMessage();
```

fluss-client/src/main/java/org/apache/fluss/client/metadata/MetadataUpdater.java
```diff
@@ -56,12 +57,14 @@
 import java.util.stream.Collectors;
 
 import static org.apache.fluss.client.utils.MetadataUtils.sendMetadataRequestAndRebuildCluster;
+import static org.apache.fluss.utils.ExceptionUtils.stripExecutionException;
 
 /** The updater to initialize and update client metadata. */
 public class MetadataUpdater {
     private static final Logger LOG = LoggerFactory.getLogger(MetadataUpdater.class);
 
     private static final int MAX_RETRY_TIMES = 5;
```
wuchong: Do we need so many retry times? 3 times is enough?

fluss-client/src/main/java/org/apache/fluss/client/metadata/MetadataUpdater.java
```diff
@@ -293,8 +296,24 @@ private static Cluster initializeCluster(Configuration conf, RpcClient rpcClient
...
+                    // if there is only one bootstrap server, we can retry to connect to it.
+                    cluster =
+                            tryToInitializeClusterWithRetries(
+                                    rpcClient, serverNode, adminReadOnlyGateway, MAX_RETRY_TIMES);
+                } else {
+                    cluster = tryToInitializeCluster(adminReadOnlyGateway);
+                    break;
+                }
             } catch (Exception e) {
                 LOG.error(
```
wuchong, line 325: We should `.disconnect()` here as well? for the case of `tryToInitializeCluster` failed

fluss-client/src/main/java/org/apache/fluss/client/metadata/MetadataUpdater.java
```diff
@@ -316,14 +335,61 @@ private static Cluster initializeCluster(Configuration conf, RpcClient rpcClient
         return cluster;
```
wuchong, line 343: We should improve this exception message, it didn't carry exception message to the top error message which may confuse users. 

```
            String errorMsg =
                    "Failed to initialize fluss client connection to bootstrap servers: "
                            + inetSocketAddresses
                            + ". \nReason: "
                            + lastException.getMessage();
```

swuferhong:
@wuchong comments addressed.

PR #2037, opened 2025-11-27
[lake] Use file to store lake snapshot

### LINKED ISSUE #2060: The zk node to store lake snapshot may be large to degrade performance
### Search before asking

### Description

as title, it may be large than 1M easily which is not recomended in zk

### Willingness to contribute

luoyuxia:
### Purpose

Linked issue: close #2060 

### Brief change log
- Change to store lake snapshot info to a file and record the file handle in zk node instead of the full lake snapshot
- Change commit lake snapshot to be asynchronously since it'll have heavy io

### DIFF
--- a/fluss-server/src/main/java/org/apache/fluss/server/zk/data/lake/LakeTable.java
+++ b/fluss-server/src/main/java/org/apache/fluss/server/zk/data/lake/LakeTable.java
@@ -30,6 +30,7 @@ import java.io.ByteArrayOutputStream;
 import java.io.IOException;
 import java.util.Collections;
 import java.util.List;
+import java.util.Objects;
 
 import static org.apache.fluss.metrics.registry.MetricRegistry.LOG;
 import static org.apache.fluss.utils.Preconditions.checkNotNull;
@@ -51,7 +52,7 @@ public class LakeTable {
 
     // Version 2 (current):
     // a list of lake snapshot metadata, record the metadata for different lake snapshots
-    @Nullable private final List<LakeSnapshotMetadata> lakeSnapshotMetadata;
+    @Nullable private final List<LakeSnapshotMetadata> lakeSnapshotMetadatas;
 
     // Version 1 (legacy): the full lake table snapshot info stored in ZK, will be null in version2
     @Nullable private final LakeTableSnapshot lakeTableSnapshot;
@@ -77,30 +78,30 @@ public class LakeTable {
     /**
      * Creates a LakeTable with a list of lake snapshot metadata (version 2 format).
      *
-     * @param lakeSnapshotMetadata the list of lake snapshot metadata
+     * @param lakeSnapshotMetadatas the list of lake snapshot metadata
      */
-    public LakeTable(List<LakeSnapshotMetadata> lakeSnapshotMetadata) {
-        this(null, lakeSnapshotMetadata);
+    public LakeTable(List<LakeSnapshotMetadata> lakeSnapshotMetadatas) {
+        this(null, lakeSnapshotMetadatas);
     }
 
     private LakeTable(
             @Nullable LakeTableSnapshot lakeTableSnapshot,
-            List<LakeSnapshotMetadata> lakeSnapshotMetadata) {
+            List<LakeSnapshotMetadata> lakeSnapshotMetadatas) {
         this.lakeTableSnapshot = lakeTableSnapshot;
-        this.lakeSnapshotMetadata = lakeSnapshotMetadata;
+        this.lakeSnapshotMetadatas = lakeSnapshotMetadatas;
     }
 
     @Nullable
-    public LakeSnapshotMetadata getLakeTableLatestSnapshot() {
-        if (lakeSnapshotMetadata != null && !lakeSnapshotMetadata.isEmpty()) {
-            return lakeSnapshotMetadata.get(0);
+    public LakeSnapshotMetadata getLatestLakeSnapshotMetadata() {
+        if (lakeSnapshotMetadatas != null && !lakeSnapshotMetadatas.isEmpty()) {
+            return lakeSnapshotMetadatas.get(0);
         }
         return null;
     }
 
     @Nullable
-    public List<LakeSnapshotMetadata> getLakeSnapshotMetadata() {
-        return lakeSnapshotMetadata;
+    public List<LakeSnapshotMetadata> getLakeSnapshotMetadatas() {
+        return lakeSnapshotMetadatas;
     }
 
     /**
@@ -116,7 +117,7 @@ public class LakeTable {
             return lakeTableSnapshot;
         }
         FsPath tieredOffsetsFilePath =
-                checkNotNull(getLakeTableLatestSnapshot()).tieredOffsetsFilePath;
+                checkNotNull(getLatestLakeSnapshotMetadata()).tieredOffsetsFilePath;
         FSDataInputStream inputStream =
                 tieredOffsetsFilePath.getFileSystem().open(tieredOffsetsFilePath);
         try (ByteArrayOutputStream outputStream = new ByteArrayOutputStream()) {
@@ -178,5 +179,52 @@ public class LakeTable {
                 LOG.warn("Error deleting filePath at {}", fsPath, e);
             }
         }
+
+        @Override
+        public boolean equals(Object o) {
+            if (this == o) {
+                return true;
+            }
+            if (!(o instanceof LakeSnapshotMetadata)) {
+                return false;
+            }
+            LakeSnapshotMetadata that = (LakeSnapshotMetadata) o;
+            return snapshotId == that.snapshotId
+                    && Objects.equals(tieredOffsetsFilePath, that.tieredOffsetsFilePath)
+                    && Objects.equals(readableOffsetsFilePath, that.readableOffsetsFilePath);
+        }
+
+        @Override
+        public int hashCode() {
+            return Objects.hash(snapshotId, tieredOffsetsFilePath, readableOffsetsFilePath);
+        }
+    }
+
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) {
+            return true;
+        }
+        if (!(o instanceof LakeTable)) {
+            return false;
+        }
+        LakeTable lakeTable = (LakeTable) o;
+        return Objects.equals(lakeSnapshotMetadatas, lakeTable.lakeSnapshotMetadatas)
+                && Objects.equals(lakeTableSnapshot, lakeTable.lakeTableSnapshot);
+    }
+
+    @Override
+    public int hashCode() {
+        return Objects.hash(lakeSnapshotMetadatas, lakeTableSnapshot);
+    }
+
+    @Override
+    public String toString() {
+        return "LakeTable{"
+                + "lakeSnapshotMetadatas="
+                + lakeSnapshotMetadatas
+                + ", lakeTableSnapshot="
+                + lakeTableSnapshot
+                + '}';
     }
 }
--- a/fluss-server/src/main/java/org/apache/fluss/server/zk/data/lake/LakeTableJsonSerde.java
+++ b/fluss-server/src/main/java/org/apache/fluss/server/zk/data/lake/LakeTableJsonSerde.java
@@ -66,7 +66,7 @@ public class LakeTableJsonSerde implements JsonSerializer<LakeTable>, JsonDeseri
 
         generator.writeArrayFieldStart(LAKE_SNAPSHOTS);
         for (LakeTable.LakeSnapshotMetadata lakeSnapshotMetadata :
-                checkNotNull(lakeTable.getLakeSnapshotMetadata())) {
+                checkNotNull(lakeTable.getLakeSnapshotMetadatas())) {
             generator.writeStartObject();
 
             generator.writeNumberField(SNAPSHOT_ID_KEY, lakeSnapshotMetadata.getSnapshotId());
@@ -100,22 +100,24 @@ public class LakeTableJsonSerde implements JsonSerializer<LakeTable>, JsonDeseri
                         "Invalid lake_snapshots field in version 2 format");
             }
 
-            List<LakeTable.LakeSnapshotMetadata> lakeSnapshotMetadata = new ArrayList<>();
+            List<LakeTable.LakeSnapshotMetadata> lakeSnapshotMetadatas = new ArrayList<>();
             Iterator<JsonNode> elements = lakeSnapshotsNode.elements();
             while (elements.hasNext()) {
                 JsonNode snapshotNode = elements.next();
                 long snapshotId = snapshotNode.get(SNAPSHOT_ID_KEY).asLong();
                 String tieredOffsetsPath = snapshotNode.get(TIERED_OFFSETS_KEY).asText();
-                String readableOffsetsPath = snapshotNode.get(READABLE_OFFSETS_KEY).asText();
+                JsonNode readableOffsetsNode = snapshotNode.get(READABLE_OFFSETS_KEY);
+                FsPath readableOffsetsPath =
+                        readableOffsetsNode != null
+                                ? new FsPath(readableOffsetsNode.asText())
+                                : null;
 
                 LakeTable.LakeSnapshotMetadata metadata =
                         new LakeTable.LakeSnapshotMetadata(
-                                snapshotId,
-                                new FsPath(tieredOffsetsPath),
-                                new FsPath(readableOffsetsPath));
-                lakeSnapshotMetadata.add(metadata);
+                                snapshotId, new FsPath(tieredOffsetsPath), readableOffsetsPath);
+                lakeSnapshotMetadatas.add(metadata);
             }
-            return new LakeTable(lakeSnapshotMetadata);
+            return new LakeTable(lakeSnapshotMetadatas);
         } else {
             throw new IllegalArgumentException("Unsupported version: " + version);
         }
--- a/fluss-server/src/main/java/org/apache/fluss/server/zk/data/lake/LakeTableHelper.java
+++ b/fluss-server/src/main/java/org/apache/fluss/server/zk/data/lake/LakeTableHelper.java
@@ -98,7 +98,7 @@ public class LakeTableHelper {
         if (optPreviousLakeTable.isPresent()) {
             // discard previous latest lake snapshot
             LakeTable.LakeSnapshotMetadata previousLakeSnapshotMetadata =
-                    optPreviousLakeTable.get().getLakeTableLatestSnapshot();
+                    optPreviousLakeTable.get().getLatestLakeSnapshotMetadata();
             if (previousLakeSnapshotMetadata != null) {
                 previousLakeSnapshotMetadata.discard();
             }
--- a/fluss-server/src/test/java/org/apache/fluss/server/replica/ReplicaTestBase.java
+++ b/fluss-server/src/test/java/org/apache/fluss/server/replica/ReplicaTestBase.java
@@ -168,7 +168,7 @@ public class ReplicaTestBase {
         conf.setString(ConfigOptions.DATA_DIR, tempDir.getAbsolutePath());
         conf.setString(ConfigOptions.COORDINATOR_HOST, "localhost");
         conf.set(ConfigOptions.REMOTE_DATA_DIR, tempDir.getAbsolutePath() + "/remote_data_dir");
-        conf.set(ConfigOptions.REMOTE_LOG_DATA_TRANSFER_THREAD_NUM, 1);
+        conf.set(ConfigOptions.SERVER_IO_POOL_SIZE, 2);
         // set snapshot interval to 1 seconds for test purpose
         conf.set(ConfigOptions.KV_SNAPSHOT_INTERVAL, Duration.ofSeconds(1));
 
--- a/fluss-server/src/test/java/org/apache/fluss/server/zk/data/lake/LakeTableHelperTest.java
+++ b/fluss-server/src/test/java/org/apache/fluss/server/zk/data/lake/LakeTableHelperTest.java
@@ -140,12 +140,12 @@ class LakeTableHelperTest {
             Optional<LakeTable> optLakeTableAfter = zooKeeperClient.getLakeTable(tableId);
             assertThat(optLakeTableAfter).isPresent();
             LakeTable lakeTableAfter = optLakeTableAfter.get();
-            assertThat(lakeTableAfter.getLakeTableLatestSnapshot())
+            assertThat(lakeTableAfter.getLatestLakeSnapshotMetadata())
                     .isNotNull(); // Version 2 has file path
 
             // Verify: The lake snapshot file exists
             FsPath snapshot2FileHandle =
-                    lakeTableAfter.getLakeTableLatestSnapshot().getReadableOffsetsFilePath();
+                    lakeTableAfter.getLatestLakeSnapshotMetadata().getReadableOffsetsFilePath();
             FileSystem fileSystem = snapshot2FileHandle.getFileSystem();
             assertThat(fileSystem.exists(snapshot2FileHandle)).isTrue();
 
--- a/fluss-server/src/test/java/org/apache/fluss/server/zk/data/lake/LakeTableJsonSerdeTest.java
+++ b/fluss-server/src/test/java/org/apache/fluss/server/zk/data/lake/LakeTableJsonSerdeTest.java
@@ -1,0 +1,77 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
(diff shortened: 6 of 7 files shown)


fluss-server/src/main/java/org/apache/fluss/server/coordinator/CoordinatorEventProcessor.java
```diff
@@ -1146,50 +1172,52 @@ private <T> void processAccessContext(AccessContextEvent<T> event) {
...
+                        for (Map.Entry<Long, LakeTableSnapshot> lakeTableSnapshotEntry :
+                                lakeTableSnapshots.entrySet()) {
+                            Long tableId = lakeTableSnapshotEntry.getKey();
+
+                            PbCommitLakeTableSnapshotRespForTable tableResp =
+                                    response.addTableResp();
+                            tableResp.setTableId(tableId);
+
+                            try {
+                                TablePath tablePath = coordinatorContext.getTablePathById(tableId);
```
wuchong: `coordinatorContext` is not thread-safe, and can't be accessed in other threads. We should get the path before `ioExecutor`.

fluss-server/src/main/java/org/apache/fluss/server/coordinator/CoordinatorEventProcessor.java
```diff
@@ -1146,50 +1172,52 @@ private <T> void processAccessContext(AccessContextEvent<T> event) {
...
+                                if (tablePath == null) {
+                                    throw new RuntimeException(
+                                            String.format(
+                                                    "Failed to find table path for table id: %d",
+                                                    tableId));
+                                }
+                                // this involves IO operation (ZK), so we do it in ioExecutor
```
wuchong, line 1264: Extract the remote file writing  out of the `zooKeeperClient.upsertLakeTableSnapshot` method. This is very confusing that a ZK operation contains a heavy remote file IO and we can't see it from method name and the above comment. The ZK method should only take a remote path, and record it into ZK.

fluss-server/src/main/java/org/apache/fluss/server/zk/data/lake/LakeTableJsonSerde.java
```diff
@@ -0,0 +1,84 @@
...
+        int version = node.get(VERSION_KEY).asInt();
+        if (version == VERSION_1) {
+            // Version 1: ZK node contains full snapshot data, use LakeTableSnapshotJsonSerde
+            LakeTableSnapshot snapshot = LakeTableSnapshotJsonSerde.INSTANCE.deserialize(node);
+            return new LakeTable(snapshot);
+        } else if (version == VERSION_2) {
+            // Version 2: ZK node contains only metadata file path
+            if (!node.has(METADATA_PATH_KEY) || node.get(METADATA_PATH_KEY).isNull()) {
+                throw new IllegalArgumentException(
+                        "Version 2 ZK node must have non-null 'metadata_path' field");
```
wuchong: add the json node string into the exception for easier debugging?

fluss-server/src/test/java/org/apache/fluss/server/zk/ZooKeeperClientTest.java
```diff
@@ -576,4 +584,117 @@ void testZookeeperConfigPath() throws Exception {
...
+            assertThat(mergedSnapshot.getTableId()).isEqualTo(tableId);
+            assertThat(mergedSnapshot.getBucketLogStartOffset()).isEqualTo(bucketLogStartOffset);
+
+            Map<TableBucket, Long> expectedBucketLogEndOffset = new HashMap<>(bucketLogEndOffset);
+            expectedBucketLogEndOffset.putAll(newBucketLogEndOffset);
+            assertThat(mergedSnapshot.getBucketLogEndOffset())
+                    .isEqualTo(expectedBucketLogEndOffset);
+            assertThat(mergedSnapshot.getBucketMaxTimestamp()).isEqualTo(bucketMaxTimestamp);
+            assertThat(mergedSnapshot.getPartitionNameIdByPartitionId())
+                    .isEqualTo(partitionNameIdByPartitionId);
```
wuchong: Add one more snapshot and verify the previous remove snapshot metadata file is successfully deleted.

fluss-common/src/main/java/org/apache/fluss/utils/FlussPaths.java
```diff
@@ -681,6 +683,28 @@ public static FsPath remoteKvSnapshotDir(FsPath remoteKvTabletDir, long snapshot
         return new FsPath(remoteKvTabletDir, REMOTE_KV_SNAPSHOT_DIR_PREFIX + snapshotId);
     }
 
+    /**
+     * Returns the remote path for storing lake snapshot metadata required by Fluss for a table.
+     *
+     * <p>The path contract:
+     *
+     * <pre>
+     * {$remote.data.dir}/lake/{databaseName}/{tableName}-{tableId}/snapshot/{snapshotId}
```
wuchong: Better to use `{snapshotId}.snapshot` as the file name to distinguish the usage of this file, we may have more files under this directory.

fluss-server/src/main/java/org/apache/fluss/server/coordinator/CoordinatorEventProcessor.java
```diff
@@ -1146,50 +1172,52 @@ private <T> void processAccessContext(AccessContextEvent<T> event) {
...
-                                                        tb,
-                                                        lakeTableSnapshot));
-            }
-        }
-        coordinatorRequestBatch.sendNotifyLakeTableOffsetRequest(
-                coordinatorContext.getCoordinatorEpoch());
-        return response;
+                        // send notify lakehouse data request to all replicas via coordinator event
+                        coordinatorEventManager.put(
+                                new NotifyLakeTableOffsetEvent(lakeTableSnapshots));
```
wuchong: This may introduce out-of-order processing of `NotifyLakeTableOffsetEvent`. We may need to add idempotent check for the lake snapshot offsets, like check the input snapshot id is larger than the current `lakeTableSnapshotId`, and then we can safely apply the log offsets.

wuchong (commented):
1. Since the`getLatestLakeSnapshot` RPC adds one more remote storage IO, it may blocks TabletService/CoordinatorService requests queue. I suggested to having an `ioExecutor` in TabletServer as well, and move the IO operation into the `ioExecutor` and return the future. 
2. The current serialized json data of `LakeTableSnapshot` is not compacted. We should design a new serialization format to reduce the size (avoid duplicated information in the json).
3.  I suggest it's best to consider the DV table issue together, it should involve modifying the same `laketable` znode format.

fluss-server/src/main/java/org/apache/fluss/server/coordinator/CoordinatorEventProcessor.java
luoyuxia: But TabletServer will only update snapshot_id/lake_log_end offset,   lake_log_max_timestamp when the snapshot is/lake_log_end offset/lake_log_max_timestamp is larger.  See 
https://github.com/apache/fluss/blob/main/fluss-server/src/main/java/org/apache/fluss/server/log/LogTablet.java#L519
 Is it be enough?

wuchong: I'm not sure about that. I think only update when snapshot id is larger is definitely no risk. I see current approach may risk data consistency problem when there is zombie tiering service.

luoyuxia: Ok.However, Iceberg's snapshot IDs are not strictly incremental, use snapshot won't work for iceberg. Maybe we can use our internal tiering epoch to do the check?

wuchong: @luoyuxia, is the snapshot ID still a `long` value and guaranteed to be unique? Is it the same in Delta Lake?

luoyuxia: > @luoyuxia, is the snapshot ID still a `long` value and guaranteed to be unique? Is it the same in Delta Lake?

Yes, the snapshot ID still a long value and guaranteed to be unique. Delta Lake's snapshot id(called version in delta) is strictly incremental.

wuchong: How iceberg to check which snapshot is newer? Use snapshot timestamp?

luoyuxia: I think snapshot timestamp should work for most case except for the corner case that multiple snapshot commit in same timestamp-ms.
Iceberg  maintains a history with snapshot IDs which helps establish order. In this order, it can compare snapshot although time-consuming.

luoyuxia:
@wuchong Hi, I already addressed your comment.
1. Remove unnecessary information in `LakeTableSnapshot` to reduce store size
2. Store a list of snapshot in zk node, each snapshot records the tiered offset and readable offset.  In this pr, we always set tiered offset to equal to readable offset, and only keep one snapshot since if not consider dv fix, we only need one. But the dv fix issue, I'll extend it to contains multiple  lake snapshot. 
3. For comment https://github.com/apache/fluss/pull/2037#discussion_r2580518738, maybe we can address it in another pr?

fluss-common/src/main/java/org/apache/fluss/utils/FlussPaths.java
```diff
@@ -681,6 +683,45 @@ public static FsPath remoteKvSnapshotDir(FsPath remoteKvTabletDir, long snapshot
...
+    /**
+     * Returns the remote path for storing lake snapshot metadata required by Fluss for a table.
+     *
+     * <p>The path contract:
+     *
+     * <pre>
+     * {$remote.data.dir}/lake/{databaseName}/{tableName}-{tableId}
+     * </pre>
+     */
+    public static FsPath remoteLakeTableSnapshotMetadataDir(
```
wuchong:
```suggestion
    public static FsPath remoteLakeTableSnapshotDir(
```

It doesn't include `metadata`/`manifest` in the path.

fluss-common/src/main/java/org/apache/fluss/utils/FlussPaths.java
```diff
@@ -681,6 +683,45 @@ public static FsPath remoteKvSnapshotDir(FsPath remoteKvTabletDir, long snapshot
...
+     *
+     * <pre>
+     * {$remoteLakeTableSnapshotMetadataDir}/manifest/{uuid}.manifest
+     * </pre>
+     */
+    public static FsPath remoteLakeTableSnapshotManifestPath(
+            String remoteDataDir, TablePath tablePath, long tableId) {
+        return new FsPath(
+                String.format(
+                        "%s/manifest/%s.manifest",
```
wuchong:
```suggestion
                        "%s/metadata/%s.manifest",
```

To keep align with the log manifest path.

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/tiering/committer/FlussTableLakeSnapshot.java
```diff
@@ -54,27 +47,20 @@ public long lakeSnapshotId() {
...
-            TableBucket bucket, String partitionName, long offset, long timestamp) {
-        logEndOffsets.put(Tuple2.of(bucket, partitionName), offset);
-        maxTimestamps.put(Tuple2.of(bucket, partitionName), timestamp);
+    public void addBucketOffset(TableBucket bucket, long offset) {
+        logEndOffsets.put(bucket, offset);
     }
 
-    public long getLogEndOffset(Tuple2<TableBucket, String> bucketPartition) {
-        return logEndOffsets.get(bucketPartition);
+    public void addPartitionBucketOffset(TableBucket bucket, long offset) {
```
wuchong: This method looks the same as `addBucketOffset`? We can remove this method?

fluss-flink/fluss-flink-common/src/test/java/org/apache/fluss/flink/tiering/source/enumerator/TieringSourceEnumeratorTest.java
```diff
@@ -692,22 +679,20 @@ private static CommitLakeTableSnapshotRequest genCommitLakeTableSnapshotRequest(
...
+        for (Map.Entry<Integer, Long> bucketLogEndOffset : bucketLogEndOffsets.entrySet()) {
+            int bucketId = bucketLogEndOffset.getKey();
             TableBucket tb = new TableBucket(tableId, partitionId, bucketId);
             PbLakeTableOffsetForBucket lakeTableOffsetForBucket = reqForTable.addBucketsReq();
             if (tb.getPartitionId() != null) {
                 lakeTableOffsetForBucket.setPartitionId(tb.getPartitionId());
             }
             lakeTableOffsetForBucket.setBucketId(tb.getBucket());
-            lakeTableOffsetForBucket.setLogStartOffset(bucketLogStartOffset.getValue());
             lakeTableOffsetForBucket.setLogEndOffset(bucketLogEndOffsets.get(bucketId));
```
wuchong: `bucketLogEndOffset.getValue()` instead of `bucketLogEndOffsets.get(bucketId)`?

fluss-rpc/src/main/proto/FlussApi.proto
```diff
@@ -462,9 +462,7 @@ message PbLakeTableSnapshotInfo {
 message PbLakeTableOffsetForBucket {
   optional int64 partition_id = 1;
   required int32 bucket_id = 2;
-  optional int64 log_start_offset = 3;
   optional int64 log_end_offset = 4;
-  optional string partition_name = 5;
```
wuchong, line 467: Let's comment them instead of removing, so we know what are the legacy deprecated fields.

fluss-server/src/main/java/org/apache/fluss/server/coordinator/CoordinatorEventProcessor.java
```diff
@@ -1140,6 +1147,33 @@ private void processNotifyKvSnapshotOffsetEvent(NotifyKvSnapshotOffsetEvent even
...
+        Map<Long, LakeTableSnapshot> lakeTableSnapshots = event.getLakeTableSnapshots();
+        Map<TableBucket, Long> tableBucketMaxTieredTimestamps =
+                event.getTableBucketMaxTieredTimestamps();
+        coordinatorRequestBatch.newBatch();
+        for (Map.Entry<Long, LakeTableSnapshot> lakeTableSnapshotEntry :
+                lakeTableSnapshots.entrySet()) {
+            LakeTableSnapshot lakeTableSnapshot = lakeTableSnapshotEntry.getValue();
+            for (Map.Entry<TableBucket, Long> bucketLogEndOffsetEntry :
+                    lakeTableSnapshot.getBucketLogEndOffset().entrySet()) {
+                TableBucket tb = bucketLogEndOffsetEntry.getKey();
```
wuchong: Can simplify into 

```java
for (TableBucket tb : lakeTableSnapshot.getBucketLogEndOffset().keySet()) {
                coordinatorContext
                        .getBucketLeaderAndIsr(tb)
 ...
```

fluss-server/src/main/java/org/apache/fluss/server/coordinator/CoordinatorServer.java
```diff
@@ -407,6 +399,11 @@ CompletableFuture<Void> stopServices() {
                 exception = ExceptionUtils.firstOrSuppressed(t, exception);
             }
 
+            if (ioExecutor != null) {
+                // shutdown io executor
+                ExecutorUtils.gracefulShutdown(5, TimeUnit.SECONDS, ioExecutor);
```
wuchong: We should keep the original try-catch wrapper, and handle potential exceptions?

fluss-server/src/main/java/org/apache/fluss/server/replica/ReplicaManager.java
```diff
@@ -820,18 +820,9 @@ private void updateWithLakeTableSnapshot(Replica replica) throws Exception {
             LakeTableSnapshot lakeTableSnapshot = optLakeTableSnapshot.get();
             long snapshotId = optLakeTableSnapshot.get().getSnapshotId();
             replica.getLogTablet().updateLakeTableSnapshotId(snapshotId);
-
-            lakeTableSnapshot
-                    .getLogStartOffset(tb)
-                    .ifPresent(replica.getLogTablet()::updateLakeLogStartOffset);
-
```
wuchong, line 827: Currently, this method performs heavy I/O operations (e.g., reading from OSS/S3) while holding a lock, which can lead to potential deadlocks or unresponsive behavior.  

From my review, this method is only used to update the lake log offset, specifically for the `fetchFromLake` feature. However, since `fetchFromLake` is planned for removal, this method will become obsolete and can be removed alongside it.  

If my understanding is correct, could you please add a `TODO` comment in the `updateWithLakeTableSnapshot` method indicating that it can be removed when `fetchFromLake` is deprecated?

fluss-common/src/main/java/org/apache/fluss/config/ConfigOptions.java
```diff
@@ -394,6 +394,14 @@ public class ConfigOptions {
                             "The rack for the tabletServer. This will be used in rack aware bucket assignment "
                                     + "for fault tolerance. Examples: `RACK1`, `cn-hangzhou-server10`");
 
+    public static final ConfigOption<Integer> TABLET_SERVER_IO_POOL_SIZE =
+            key("tablet-server.io-pool.size")
```
wuchong: Can we unify the io-pool for coordinator and tablet servrer? We can rename this config to `server.io-pool.size` in this PR. And create follow-up issues:
1. deprecate `kv.snapshot.transfer-thread-num` and `remote.log.data-transfer-thread-num`, and use `server.io-pool.size` instead, may turn it into 10 threads by default. 
2. deprecate `coordinator.io-pool.size` and use `server.io-pool.size` instead.

fluss-server/src/test/java/org/apache/fluss/server/zk/data/LakeTableSnapshotJsonSerdeTest.java
```diff
@@ -33,75 +41,103 @@ class LakeTableSnapshotJsonSerdeTest extends JsonSerdeTestBase<LakeTableSnapshot
...
-                    + "{\"bucket_id\":1,\"log_start_offset\":1,\"log_end_offset\":3,\"max_timestamp\":5}]}",
-            "{\"version\":1,\"snapshot_id\":3,\"table_id\":5,"
-                    + "\"buckets\":[{\"partition_id\":1,\"partition_name\":\"partition1\",\"bucket_id\":1,\"log_start_offset\":1,\"log_end_offset\":3,\"max_timestamp\":5},"
-                    + "{\"partition_id\":2,\"partition_name\":\"partition2\",\"bucket_id\":1,\"log_start_offset\":2,\"log_end_offset\":4,\"max_timestamp\":6}]}"
+            // Test case 1: Empty snapshot
+            "{\"version\":2,\"snapshot_id\":1}",
+            // Test case 2: Non-partition table with consecutive bucket ids [0, 1, 2]
+            "{\"version\":2,\"snapshot_id\":2,\"table_id\":4,\"buckets\":[100,200,300]}",
+            // Test case 3: Non-partition table with missing bucket ids [0, -1, 2, -1, 4]
+            "{\"version\":2,\"snapshot_id\":3,\"table_id\":5,\"buckets\":[100,-1,300,-1,500]}",
```
wuchong: It's complex to deserialize or understand the bucket offset if we use the same property key for partitioned and non-partitioned table. Why not use different property keys for them, like `bucket_offsets`, and `partition_bucket_offsets`?

fluss-server/src/main/java/org/apache/fluss/server/coordinator/RemoteStorageCleaner.java
```diff
@@ -57,10 +61,16 @@ public RemoteStorageCleaner(Configuration configuration, ExecutorService ioExecu
         }
     }
 
-    public void asyncDeleteTableRemoteDir(TablePath tablePath, boolean isKvTable, long tableId) {
+    public void asyncDeleteTableRemoteDir(
+            TablePath tablePath, boolean isKvTable, boolean isLakeEnabled, long tableId) {
         if (isKvTable) {
             asyncDeleteDir(FlussPaths.remoteTableDir(remoteKvDir, tablePath, tableId));
         }
+        if (isLakeEnabled) {
```
wuchong: If a table was enabled datalake, but turned off later, and then the table was deleted. We may leave the lake snapshot metadata file here? Maybe we can simply do the delete for the metadata file, no matter it is datalake enabled or not? (add comment for this if we do such change)

fluss-rpc/src/main/proto/FlussApi.proto
luoyuxia: good suggestion

fluss-server/src/main/java/org/apache/fluss/server/replica/ReplicaManager.java
luoyuxia: Yes, your understanding is correct.

fluss-common/src/main/java/org/apache/fluss/config/ConfigOptions.java
luoyuxia: good suggestion

fluss-server/src/main/java/org/apache/fluss/server/coordinator/CoordinatorServer.java
luoyuxia: Miss it. I think we can follow the reasource release code pattern in here.

fluss-common/src/main/java/org/apache/fluss/config/ConfigOptions.java
```diff
@@ -769,13 +786,19 @@ public class ConfigOptions {
                             "Size of the thread pool used in scheduling tasks to copy segments, "
                                     + "fetch remote log indexes and clean up remote log segments.");
 
+    /**
+     * @deprecated This option is deprecated. Please use {@link ConfigOptions#SERVER_IO_POOL_SIZE}
+     *     instead.
+     */
+    @Deprecated
     public static final ConfigOption<Integer> REMOTE_LOG_DATA_TRANSFER_THREAD_NUM =
```
wuchong, line 794: There is still a usage in the test code. We should update the test code to use the new config.

fluss-server/src/main/java/org/apache/fluss/server/zk/data/lake/LakeTable.java
```diff
@@ -0,0 +1,182 @@
...
+
+    private LakeTable(
+            @Nullable LakeTableSnapshot lakeTableSnapshot,
+            List<LakeSnapshotMetadata> lakeSnapshotMetadata) {
+        this.lakeTableSnapshot = lakeTableSnapshot;
+        this.lakeSnapshotMetadata = lakeSnapshotMetadata;
+    }
+
+    @Nullable
+    public LakeSnapshotMetadata getLakeTableLatestSnapshot() {
```
wuchong: The method name confuses a lot, because it doesn't returns a `LakeTableSnapshot`. Better to name `getLatestLakeSnapshotMetadata`.

fluss-server/src/main/java/org/apache/fluss/server/zk/data/lake/LakeTable.java
```diff
@@ -0,0 +1,182 @@
...
+        if (lakeSnapshotMetadata != null && !lakeSnapshotMetadata.isEmpty()) {
+            return lakeSnapshotMetadata.get(0);
+        }
+        return null;
+    }
+
+    @Nullable
+    public List<LakeSnapshotMetadata> getLakeSnapshotMetadata() {
```
wuchong: It returns a list of `LakeSnapshotMetadata`, better to name `getLakeSnapshotMetadatas`.

fluss-server/src/main/java/org/apache/fluss/server/zk/data/lake/LakeTableJsonSerde.java
```diff
@@ -0,0 +1,123 @@
...
+ *
+ * <ul>
+ *   <li>Version 1 (legacy): ZK node contains full {@link LakeTableSnapshot} data. During
+ *       deserialization, it uses {@link LakeTableSnapshotJsonSerde} to deserialize and wraps the
+ *       result in a {@link LakeTable}.
+ *   <li>Version 2 (current): ZK node contains only the lake table snapshot file paths. The actual
+ *       snapshot data is stored in a remote file pointed by the lake table snapshot file path.
+ * </ul>
+ */
+public class LakeTableJsonSerde implements JsonSerializer<LakeTable>, JsonDeserializer<LakeTable> {
```
wuchong, line 47: I didn't see a `LakeTableJsonSerdeTest` for this class. We should have such test to guaratneee the JSON format is as expected, and also for future compatibility tests.

wuchong (commented):
As we have deprecated some options with new options, could you add some statement for this in the upgrade notes?

luoyuxia:
@wuchong Comments has been addressed.

PR #2042, opened 2025-11-27
[Flink] Support Partial Updates to the Flink Sink

polyzos:


fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/sink/FlussSinkBuilder.java
```diff
@@ -193,4 +219,48 @@ private void validateConfiguration() {
...
+            return null; // full update
+        }
+
+        // Map specified column names to indexes
+        int[] indexes = new int[specifiedColumns.size()];
+        for (int i = 0; i < specifiedColumns.size(); i++) {
+            String col = specifiedColumns.get(i);
+            int idx = allFieldNames.indexOf(col);
+            checkArgument(
+                    idx >= 0, "Column '%s' not found in table schema: %s", col, allFieldNames);
```
leekeiabstraction, line 252: nit: Slightly more context can be provided in the message e.g. "Partial update column '%s' not found in table schema: %s'

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/sink/FlussSinkBuilder.java
```diff
@@ -98,6 +102,23 @@ public FlussSinkBuilder<InputT> setShuffleByBucketId(boolean shuffleByBucketId)
         return this;
     }
 
+    /**
+     * Enable partial update by specifying the column names to update for upsert tables. Primary key
+     * columns must be included in this list.
+     */
+    public FlussSinkBuilder<InputT> setPartialUpdateColumns(List<String> columns) {
+        this.partialUpdateColumns = columns;
```
fresh-borzoni, line 110: should we defensively copy here, otherwise caller has chance to modify passed list?

leekeiabstraction:
IT looks good and run successfully on my local environment. Approving!

MehulBatra:
+1

PR #2043, opened 2025-11-27
[iceberg] support altering table properties

nastra:
### Purpose

This adds support for altering Iceberg table properties

### Tests

added a unit test that sets/removes table properties

### DIFF
--- a/fluss-lake/fluss-lake-iceberg/src/main/java/org/apache/fluss/lake/iceberg/IcebergLakeCatalog.java
+++ b/fluss-lake/fluss-lake-iceberg/src/main/java/org/apache/fluss/lake/iceberg/IcebergLakeCatalog.java
@@ -30,9 +30,11 @@ import org.apache.fluss.metadata.TablePath;
 import org.apache.fluss.utils.IOUtils;
 
 import org.apache.iceberg.PartitionSpec;
+import org.apache.iceberg.RowLevelOperationMode;
 import org.apache.iceberg.Schema;
 import org.apache.iceberg.SortOrder;
 import org.apache.iceberg.Table;
+import org.apache.iceberg.TableProperties;
 import org.apache.iceberg.UpdateProperties;
 import org.apache.iceberg.catalog.Catalog;
 import org.apache.iceberg.catalog.Namespace;
@@ -55,10 +57,17 @@ import java.util.Set;
 import static org.apache.fluss.metadata.TableDescriptor.BUCKET_COLUMN_NAME;
 import static org.apache.fluss.metadata.TableDescriptor.OFFSET_COLUMN_NAME;
 import static org.apache.fluss.metadata.TableDescriptor.TIMESTAMP_COLUMN_NAME;
+import static org.apache.fluss.utils.Preconditions.checkArgument;
 import static org.apache.iceberg.types.Type.TypeID.STRING;
 
 /** An Iceberg implementation of {@link LakeCatalog}. */
 public class IcebergLakeCatalog implements LakeCatalog {
+    @VisibleForTesting
+    static final Set<String> RESERVED_PROPERTIES =
+            Set.of(
+                    TableProperties.MERGE_MODE,
+                    TableProperties.UPDATE_MODE,
+                    TableProperties.DELETE_MODE);
 
     public static final LinkedHashMap<String, Type> SYSTEM_COLUMNS = new LinkedHashMap<>();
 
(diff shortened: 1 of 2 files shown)


fluss-lake/fluss-lake-iceberg/src/main/java/org/apache/fluss/lake/iceberg/IcebergLakeCatalog.java
```diff
@@ -249,6 +271,14 @@ private void setFlussPropertyToIceberg(
         }
     }
 
+    private static String convertFlussPropertyKeyToIceberg(String key) {
```
MehulBatra, line 291: If a user passes "iceberg." or "fluss." (just the prefix with no actual property name)
This will attempt to set/remove a property with an empty key in Iceberg, which could create invalid table metadata
wdyt?

fluss-lake/fluss-lake-iceberg/src/main/java/org/apache/fluss/lake/iceberg/IcebergLakeCatalog.java
```diff
@@ -119,8 +122,27 @@ public void createTable(TablePath tablePath, TableDescriptor tableDescriptor, Co
     @Override
     public void alterTable(TablePath tablePath, List<TableChange> tableChanges, Context context)
```
MehulBatra, line 132: For primary key tables, Fluss automatically sets merge-on-read properties for Iceberg table
eg:  icebergProperties.put("write.delete.mode", "merge-on-read)...etc
We should not allow user to change that for Primary key table I feel it was set purposely  how to merge data and delete files for reading, we can put a check on that.

MehulBatra:
Overall LGTM!
Left small comments, @luoyuxia it would be great if you could take a look too!

fluss-lake/fluss-lake-iceberg/src/main/java/org/apache/fluss/lake/iceberg/IcebergLakeCatalog.java
nastra: You would just end up with an empty table property that wouldn't do anything, so I don't think we need to do any special casing here and one wouldn't have invalid table metadata. Iceberg also doesn't prevent against this case as can be seen in the impl: https://github.com/apache/iceberg/blob/fe23584fc3af9f0ea1371989030b0a99affb233f/core/src/main/java/org/apache/iceberg/PropertiesUpdate.java#L50-L70

fluss-lake/fluss-lake-iceberg/src/main/java/org/apache/fluss/lake/iceberg/IcebergLakeCatalog.java
nastra: that makes sense. I've added some checks and a test to make sure that those reserved properties can't be altered

fluss-lake/fluss-lake-iceberg/src/main/java/org/apache/fluss/lake/iceberg/IcebergLakeCatalog.java
```diff
@@ -119,8 +131,35 @@ public void createTable(TablePath tablePath, TableDescriptor tableDescriptor, Co
...
             throws TableNotExistException {
-        throw new UnsupportedOperationException(
-                "Alter table is not supported for Iceberg at the moment");
+        try {
+            Table table = icebergCatalog.loadTable(toIcebergTableIdentifier(tablePath));
+            UpdateProperties updateProperties = table.updateProperties();
+            for (TableChange tableChange : tableChanges) {
+                if (tableChange instanceof TableChange.SetOption) {
+                    TableChange.SetOption option = (TableChange.SetOption) tableChange;
+                    checkArgument(
```
nastra, line 140: @MehulBatra does it make sense to throw IAE exception here, since the alterTable API doesn't specify what to throw in such a case.

MehulBatra: I believe that should be good, I could see it being used in  as standard utility across

fluss-lake/fluss-lake-iceberg/src/main/java/org/apache/fluss/lake/iceberg/IcebergLakeCatalog.java
MehulBatra: Makes sense, thanks for the reference.

MehulBatra (approved):

PR #2046, opened 2025-11-28
[paimon] Align error msg when altering table properties

nastra:
### Purpose

This aligns the error msg that is thrown when altering table properties of a non-existing table. Previously the error msg contained a paimon-specific identifier and I've changed it to show the `TablePath` instead. I've also added a test

### Tests

Added a unit test to verify the error msg

### DIFF
--- a/fluss-lake/fluss-lake-paimon/src/main/java/org/apache/fluss/lake/paimon/PaimonLakeCatalog.java
+++ b/fluss-lake/fluss-lake-paimon/src/main/java/org/apache/fluss/lake/paimon/PaimonLakeCatalog.java
@@ -84,15 +84,14 @@ public class PaimonLakeCatalog implements LakeCatalog {
     public void createTable(TablePath tablePath, TableDescriptor tableDescriptor, Context context)
             throws TableAlreadyExistException {
         // then, create the table
-        Identifier paimonPath = toPaimon(tablePath);
         Schema paimonSchema = toPaimonSchema(tableDescriptor);
         try {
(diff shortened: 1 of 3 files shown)


LiebingYu:
@nastra Thanks for the PR. LGTM overall, but I think we can also change the msg in `createTable`.
```java
    private void createTable(Identifier tablePath, Schema schema, boolean isCreatingFlussTable)
            throws Catalog.DatabaseNotExistException {
        try {
            // not ignore if table exists
            paimonCatalog.createTable(tablePath, schema, false);
        } catch (Catalog.TableAlreadyExistException e) {
            try {
                Table table = paimonCatalog.getTable(tablePath);
                FileStoreTable fileStoreTable = (FileStoreTable) table;
                validatePaimonSchemaCompatible(
                        tablePath, fileStoreTable.schema().toSchema(), schema);
                // if creating a new fluss table, we should ensure the lake table is empty
                if (isCreatingFlussTable) {
                    checkTableIsEmpty(tablePath, fileStoreTable);
                }
            } catch (Catalog.TableNotExistException tableNotExistException) {
                // shouldn't happen in normal cases
                throw new RuntimeException(
                        String.format(
                                "Failed to create table %s in Paimon. The table already existed "
                                        + "during the initial creation attempt, but subsequently "
                                        + "could not be found when trying to get it. "
                                        + "Please check whether the Paimon table was manually deleted, and try again.",
                                tablePath));
            }
        }
    }
```

nastra:
@LiebingYu I've updated the error msg of `createTable` to include `TablePath` instead of `Identifier`

PR #2058, opened 2025-12-01
[hotfix] Update paimon version to 1.3.1 in fluss quickstart

zhuangchong:
### Purpose

Update paimon version to 1.3.1 in fluss quickstart. Keep consistent with fluss-lake-paimon module paimon version.


zhuangchong:
cc @luoyuxia

Issue #1983, opened 2025-11-14
Lake Tiering Service is not compatible with Flink 2.1 (ClassNotFoundException MultipleParameterTool)


wuchong:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/fluss/issues) and found nothing similar.

### Fluss version

main (development)

### Please describe the bug 🐞

Using lake tiering service of fluss 0.8 with Flink 2.1, throws the following exception:

<img width="887" height="443" alt="Image" src="https://github.com/user-attachments/assets/745b81b6-3eba-4515-8101-c734b01e2cd9" />

```
java.lang.ClassNotFoundException: org.apache.flink.api.java.utils.MultipleParameterTool
        at java.base/java.net.URLClassLoader.findClass(URLClassLoader.java:445) ~[?:?]
        at java.base/java.lang.ClassLoader.loadClass(ClassLoader.java:592) ~[?:?]
```

### Are you willing to submit a PR?

- [ ] I'm willing to submit a PR!

buvb:
@wuchong I can attempt to fix this bug.

luoyuxia:
Flink2.x has rename `org.apache.flink.api.java.utils.MultipleParameterTool` to `org.apache.flink.util.MultipleParameterTool`. Seems we can follow how paimon solve this problem. @buvb Are you interested in this issue? I can assign to you

wuchong:
Yes, we can wrap a adapter for this class.

sd4324530:
I found that Paimon's solution provides the `paimon-flink1-common` module and the `paimon-flink2-common` module, which implement the adapter respectively.  Fluss only has a unified common module. Should it also be split into flink1 and flink2?

And this issue is also the same solution, right? https://github.com/apache/fluss/issues/2032 
@wuchong @luoyuxia

buvb:
@luoyuxia I found that this problem is caused by flink version incompatibility, actually bidirectional incompatibility, I think the best way is to use Adapter to isolate API differences between different flink versions, we can refer to Apache Paimon's approach, you can assign this issue to me

wuchong:
@buvb sorry for the late response. However, I saw @sd4324530 has submitted a PR for this. Thus I assigned this issue to him. 

I copied my comment from the PR to here to continue discussion if needed. 

https://github.com/apache/fluss/pull/2063#issuecomment-3602523204

[...]

PR #2063, opened 2025-12-02
[flink] fix ClassNotFoundException MultipleParameterTool

sd4324530:
### Purpose

Linked issue: close https://github.com/apache/fluss/issues/1983

### Brief change log

Add the `MultipleParameterToolAdapter` in the modules `fluss-flink-common` and `fluss-flink-2.1` and unit tests.

### DIFF
--- a/fluss-flink/fluss-flink-1.19/pom.xml
+++ b/fluss-flink/fluss-flink-1.19/pom.xml
@@ -71,13 +71,6 @@
             <scope>provided</scope>
         </dependency>
 
-        <dependency>
-            <groupId>org.apache.flink</groupId>
-            <artifactId>flink-streaming-java</artifactId>
-            <version>${flink.minor.version}</version>
-            <scope>provided</scope>
-        </dependency>
-
 
         <!-- test dependency -->
         <dependency>


wuchong:
@sd4324530 we already have dedicated flink version modules `fluss-flink-1.18`, `fluss-flink-1.19`, `fluss-flink-1.20`, `fluss-flink-2.1` to handle compatibility, so we don't need to introduce `flink1` and `flink2` modules. There are already some adapter classes in them. You can follow it. 

One more thing is that we should add tests to reproduce the problem. So that we don't encounter it again in the future.

sd4324530:
> @sd4324530 we already have dedicated flink version modules `fluss-flink-1.18`, `fluss-flink-1.19`, `fluss-flink-1.20`, `fluss-flink-2.1` to handle compatibility, so we don't need to introduce `flink1` and `flink2` modules. There are already some adapter classes in them. You can follow it.
> 
> One more thing is that we should add tests to reproduce the problem. So that we don't encounter it again in the future.

@wuchong done, and please take a look.

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/adapter/MultipleParameterToolAdapter.java
```diff
@@ -0,0 +1,40 @@
...
+ * limitations under the License.
+ */
+
+package org.apache.fluss.flink.adapter;
+
+import org.apache.flink.api.java.utils.MultipleParameterTool;
+
+import java.util.Map;
+
+/** A adapter for flink 1.x {@link MultipleParameterTool} . */
```
wuchong:
```suggestion
/**
 * An adapter for Flink {@link MultipleParameterTool} class. The {@link MultipleParameterTool} is
 * moved to a new package since Flink 2.x, so this adapter helps to bridge compatibility for
 * different Flink versions.
 *
 * <p>TODO: remove this class when no longer support all the Flink 1.x series.
 */
```

Add more comment to explain why we need this adapter. Please update all the `MultipleParameterToolAdapter`s.

fluss-flink/fluss-flink-1.18/src/main/java/org/apache/fluss/flink/adapter/MultipleParameterToolAdapter.java
```diff
@@ -0,0 +1,40 @@
...
+ */
+
+package org.apache.fluss.flink.adapter;
+
+import org.apache.flink.api.java.utils.MultipleParameterTool;
+
+import java.util.Map;
+
+/** A adapter for flink 1.x {@link MultipleParameterTool} . */
+public class MultipleParameterToolAdapter {
```
wuchong: We don't need to add the adapter for the 1.18, 1.19, 1.20 modules, because they are the same code with the `fluss-flink-common`, and no need to override the class. We only need to add the adapter for 2.1.

fluss-flink/fluss-flink-1.20/pom.xml
```diff
@@ -71,6 +71,13 @@
             <scope>provided</scope>
         </dependency>
 
+        <dependency>
+            <groupId>org.apache.flink</groupId>
+            <artifactId>flink-streaming-java</artifactId>
+            <version>${flink.minor.version}</version>
+            <scope>provided</scope>
```
wuchong: Why do we need the dependency? `MultipleParameterTool` is not in `flink-streaming-java`, right?

wuchong (commented):
Thanks @sd4324530 , the change looks good to me in general. 

I only left some minor comments.

sd4324530:
@wuchong Thanks. Repaired

wuchong (approved):
I reverted the changes in flink 1.19 pom.xml. If the CI passes, I will merge the PR.

PR #2080, opened 2025-12-02
[metric] Add user level metrics for byteIn and byteOut

### LINKED ISSUE #2059: [metrics] Add metrics to record principal-level byteIn and byteOut
### Search before asking

### Motivation

For Fluss cluster for multi-tenant and users, it is necessary to record "byteIn" and "byteOut" or related metrics not only at table granularity but also at user granularity. 

Assuming such a scenario: a producer write to Fluss log table and many consumers scan the table to get data from their partition respectively. If we only observe the "byteOut" at table granularity, we don't know the proportion "byteOut" of each consumer. Thus hard to locate where a large scan flow from or to calculate respective resource costs of each consumer.

### Solution

Currently We have a "FlussPrincipal" in "Session" like :
```
public class Session implements Serializable {
    private final short apiVersion;
    private final String listenerName;
    private final boolean isInternal;
    private final InetAddress inetAddress;
    **private final FlussPrincipal principal;**
...
}
```

We can add the param "FlussPrincipal" to TabletService::produceLog and TabletService::fetchLog to locate the user who is writing or reading data from fluss. Then we can easily add 2 metrics "bytesInPerUser","bytesOutPerUser" and other metrics if needed.

### Anything else?

_No response_

### Willingness to contribute

zcoo:
### Purpose

Linked issue: close #2059


fluss-server/src/main/java/org/apache/fluss/server/metrics/group/TableMetricGroup.java
```diff
@@ -222,6 +249,48 @@ public Counter failedPrefixLookupRequests() {
...
     }
 
+    // ------------------------------------------------------------------------
+    //  user groups
+    // ------------------------------------------------------------------------
+    private UserMetricGroup getOrCreateUserMetricGroup(String principalName) {
+        return userMetricGroups.computeIfAbsent(
+                principalName,
+                name -> {
+                    return new UserMetricGroup(this, principalName);
```
swuferhong: Replace with lambda expression.

fluss-server/src/main/java/org/apache/fluss/server/metrics/group/TableMetricGroup.java
```diff
@@ -222,6 +249,48 @@ public Counter failedPrefixLookupRequests() {
...
+                });
+    }
+
+    private static class UserMetricGroup extends AbstractMetricGroup {
+        private final String principalName;
+        protected final Counter bytesIn;
+        protected final Counter bytesOut;
+        protected final Counter messagesIn;
```
swuferhong: do we need messagesIn?

fluss-server/src/main/java/org/apache/fluss/server/entity/LogUserContext.java
```diff
@@ -0,0 +1,34 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.fluss.server.entity;
+
+import org.apache.fluss.security.acl.FlussPrincipal;
+
+/** The context information of user who writes or reads table. */
+public class LogUserContext {
```
swuferhong: Why this name called `LogUserContext`. What `Log` means.

fluss-server/src/main/java/org/apache/fluss/server/replica/ReplicaManager.java
```diff
@@ -514,6 +535,16 @@ public void putRecordsToKv(
                 timeoutMs, requiredAcks, entriesPerBucket.size(), kvPutResult, responseCallback);
     }
 
+    public void putRecordsToKv(
```
swuferhong: Don't split this method into two just for testing purposes. If you feel the test changes are extensive, you can extract a common method within the test itself.

fluss-server/src/main/java/org/apache/fluss/server/replica/ReplicaManager.java
```diff
@@ -450,21 +451,30 @@ public void appendRecordsToLog(
...
-                appendToLocalLog(entriesPerBucket, requiredAcks);
+                appendToLocalLog(entriesPerBucket, requiredAcks, userContext);
         LOG.debug("Append records to local log in {} ms", System.currentTimeMillis() - startTime);
 
         // maybe do delay write operation.
         maybeAddDelayedWrite(
                 timeoutMs, requiredAcks, entriesPerBucket.size(), appendResult, responseCallback);
     }
 
+    public void appendRecordsToLog(
```
swuferhong: ditto, do not split.

fluss-server/src/main/java/org/apache/fluss/server/replica/ReplicaManager.java
```diff
@@ -997,8 +1031,8 @@ private Map<TableBucket, PutKvResultForBucket> putToLocalKv(
                 tableMetrics.incKvMessageIn(entry.getValue().getRecordCount());
                 tableMetrics.incKvBytesIn(entry.getValue().sizeInBytes());
                 // metric for cdc log of kv
-                tableMetrics.incLogBytesIn(appendInfo.validBytes());
-                tableMetrics.incLogMessageIn(appendInfo.numMessages());
+                tableMetrics.incLogBytesIn(appendInfo.validBytes(), userContext);
```
swuferhong: I'm not entirely sure whether CDC logs need to be collected. Personally, I think we should collect them—even if we don't yet have a clear pricing model for KV. What do you think?

zcoo: I think so. We'd better collect it for future usage.

zcoo:
@swuferhong  comments addressed. PLAT

fluss-server/src/main/java/org/apache/fluss/server/metrics/group/TableMetricGroup.java
```diff
@@ -43,6 +47,8 @@ public class TableMetricGroup extends AbstractMetricGroup {
 
     private final Map<TableBucket, BucketMetricGroup> buckets = new HashMap<>();
 
+    private final Map<String, UserMetricGroup> userMetricGroups = MapUtils.newConcurrentHashMap();
```
wuchong: I don't see cleanup logic for this. This may lead to memory leak or metrics reporting overhead for idle users. In kafka, there is a `ClientQuotaManager.InactiveSensorExpirationTimeSeconds` to cleanup inactive user metrics. I think we need a similar mechenism.

fluss-server/src/main/java/org/apache/fluss/server/replica/ReplicaManager.java
```diff
@@ -450,14 +451,15 @@ public void appendRecordsToLog(
             int timeoutMs,
             int requiredAcks,
             Map<TableBucket, MemoryLogRecords> entriesPerBucket,
+            UserContext userContext,
```
wuchong: Add `@Nullable` to the parameter if it may pass `null` parameter (even if only tests pass `null`). Same to other methods.

fluss-server/src/main/java/org/apache/fluss/server/metrics/group/TableMetricGroup.java
```diff
@@ -222,6 +242,42 @@ public Counter failedPrefixLookupRequests() {
...
+        }
+
+        @Override
+        protected String getGroupName(CharacterFilter filter) {
+            return "user";
+        }
+
+        @Override
+        protected void putVariables(Map<String, String> variables) {
+            variables.put("name", principalName);
```
wuchong: Use `user` as the variable name would be more intuitive? Name may confuse it's table name.

fluss-server/src/main/java/org/apache/fluss/server/metrics/group/TableMetricGroup.java
```diff
@@ -222,6 +242,42 @@ public Counter failedPrefixLookupRequests() {
         }
     }
 
+    // ------------------------------------------------------------------------
+    //  user groups
+    // ------------------------------------------------------------------------
+    private UserMetricGroup getOrCreateUserMetricGroup(String principalName) {
```
wuchong: IIUC, we only need to collect user-level IO metrics, rather than (user * table)-level. Then I think we can extract this out of `TableMetricGroup` into a global `UserMetrics` (which maintains a map of user to `UserMetricGroup`) in `ReplicaManager`. This will also help us to implement user quota in the future, which has similar implementation in kafka.

fluss-server/src/main/java/org/apache/fluss/server/replica/ReplicaManager.java
```diff
@@ -997,7 +1006,7 @@ private Map<TableBucket, PutKvResultForBucket> putToLocalKv(
                 tableMetrics.incKvMessageIn(entry.getValue().getRecordCount());
                 tableMetrics.incKvBytesIn(entry.getValue().sizeInBytes());
                 // metric for cdc log of kv
-                tableMetrics.incLogBytesIn(appendInfo.validBytes());
+                tableMetrics.incLogBytesIn(appendInfo.validBytes(), userContext);
```
wuchong: I also have concerns about this. The cdc log size is not generated by users, use it as a pricing model may confusing users. Maybe using message num/rate for kv put/lookup is better?

zcoo: Maybe we can consider this later. Currently I think we can just remove cdc log out of user level bytesIn

zcoo:
@wuchong Comment addressed. PLAT~

fluss-common/src/main/java/org/apache/fluss/config/ConfigOptions.java
```diff
@@ -1695,6 +1695,14 @@ public class ConfigOptions {
                                     + "the CoordinatorServer) it is advisable to use a port range "
                                     + "like 9250-9260.");
 
+    public static final ConfigOption<Duration> METRICS_MANAGER_INACTIVE_EXPIRATION_TIME =
+            key("metrics.manager.inactive-expiration-time")
```
wuchong: I think default 1 hour is enough for production use-case. I checked Kafka also not exposing the configuration. Maybe we can remove it.

fluss-common/src/main/java/org/apache/fluss/metrics/MetricNames.java
```diff
@@ -74,6 +74,12 @@ public class MetricNames {
     public static final String SERVER_PHYSICAL_STORAGE_LOCAL_SIZE = "localSize";
     public static final String SERVER_PHYSICAL_STORAGE_REMOTE_LOG_SIZE = "remoteLogSize";
 
+    // --------------------------------------------------------------------------------------------
+    // metrics for user
+    // --------------------------------------------------------------------------------------------
+    public static final String BYTES_IN_COUNT = "bytesInCount";
+    public static final String BYTES_OUT_COUNT = "bytesOutCount";
```
wuchong: `bytesIn` and `bytesOut` already reflect the meaning of the number of bytes.

Issue #2062, opened 2025-12-01
[website] Add redirect for latest stable version docs


leonardBang:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/fluss/issues) and found nothing similar.

### Fluss version

0.8.0 (latest release)

### Please describe the bug 🐞

As Docusaurus will hide version in path for last stable version by default, it lead to   last stable version  links broken， for example: 

https://fluss.apache.org/docs/0.8/table-design/overview/
 

### Are you willing to submit a PR?

- [x] I'm willing to submit a PR!

rionmonster:
@leonardBang 

I'll be happy to take this.

leonardBang:
Thanks for taking this ticket @rionmonster , assigned it to you although I'm trying to fix it but unfortunately I'm not good at FE, lol

leonardBang:
A WIP branch https://github.com/leonardBang/fluss/tree/FLUSS-2062 just for reference

rionmonster:
@leonardBang 

Absolutely. I think the real crux of the issue is that we want to allow explicit navigation to the current version right? If so, [the versioning docs](https://docusaurus.io/docs/versioning#configuring-versioning-behavior) seem to indicate support for a `includeCurrentVersion` configuration that isn't currently set in our config. I also came across the same plugin for supporting the invalid/unsupported redirects as well (which is probably worth doing while we're in here.

I'll dig around a bit. Hopefully it's pretty easy to reproduce locally.

rionmonster:
@leonardBang 

I think I have a fix in the works -- still ironing out a few minor details. I'm wondering if it's worth merging the existing #1961 PR to upgrade Docusaurus to 3.9.2. It's not a huge deal, it'll just require updating that branch as well to include the related plugin updates (if we end up requiring those).

leonardBang:
> Absolutely. I think the real crux of the issue is that we want to allow explicit navigation to the current version right?

No, the `latest` version(i.e 0.8 in our case), fluss has released 0.7 and 0.8 version, and current developing version on main branch is 0.9-SNAPSHOT which is the version responsible to Docusaurus's `current` version. And I've assigned #1961 ticket to you and merged your PR, bump Docusaurus version is welcome.

wuchong:
An idea is to publish twice for the latest version (e.g., current 0.8). But, users still see 2 versions in the droplist which is confusing. 

[...]

PR #2085, opened 2025-12-03
[FLUSS-2062][docs] Added Redirect for Latest Stable Version Docs

rionmonster:
### Purpose

Linked issue: close https://github.com/apache/fluss/issues/2062

Per Issue https://github.com/apache/fluss/issues/2062, this pull request adds redirect support for the latest stable version documentation (e.g. `docs/0.8/engine-flink/getting-started/` → `docs/engine-flink/getting-started/`). This change will ensure that both implicit and explicit references to the latest stable version are resolved correctly, avoiding broken links.

### Brief change log

- Added a dependency on the `@docusaurus/plugin-client-redirects` plugin to support performing client-side redirects.
- Added a new `createRedirects` redirect rule that applies only to the docs area, which will handle ensuring that the existing routes are evaluated and redirect routes are defined for those that are considered "unversioned" (e.g. "next", non-existant versions, etc.)
  - Note: The `createRedirects` function is not at all intuitive and I [highly recommend reviewing over the documentation here](https://docusaurus.io/docs/api/plugins/@docusaurus/plugin-client-redirects#CreateRedirectsFn) to help make sense of things.

### Tests

These changes were tested locally and require an explicit `npm run build` and `npm run serve` as redirect rules are _only_ applied for production builds (so an `npm run start` will _not_ apply them). You can see an example of this running locally and successfully navigating through the available versions _and_ redirecting routes that explicitly use the latest, stable version (0.8) below:

https://github.com/user-attachments/assets/7aafa9c5-aec6-4d60-a3e4-73f6b3b5eb35

### API and Format

N/A

### Documentation

N/A

### DIFF
--- a/website/docusaurus.config.ts
+++ b/website/docusaurus.config.ts
@@ -124,6 +124,32 @@ const config: Config = {
           ],
       },
     ],
+    [
+      '@docusaurus/plugin-client-redirects',
+      {
+          // Create redirects from the available routes that have already been created
+          createRedirects(existingPath) {
+            // Only evaluate paths related to documentation
+            if (!existingPath.startsWith('/docs/')) {
+              return undefined;
+            }
+            
+            // Extract the relative path after /docs/
+            const relativeDocsPath = existingPath.substring(6); 
+            const firstSegment = relativeDocsPath.split('/')[0];
+            
+            // Exclude any known version identifiers aligned with existing routes
+            const existingVersionedRoutes = ['next', ...Object.keys(versionsMap)];
+            if (existingVersionedRoutes.includes(firstSegment)) {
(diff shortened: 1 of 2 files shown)


leonardBang (approved):
@rionmonster Awesome work, I've checked in my local env. +1

rionmonster:
@leonardBang / @wuchong 

Thanks folks! I just went and checked the site and it appears that things are working as expected (confirmed that `https://fluss.apache.org/docs/0.8/engine-flink/delta-joins/#flink-version-support` routes to the implicit `https://fluss.apache.org/docs/engine-flink/delta-joins/#flink-version-support`)!

PR #2087, opened 2025-12-03
[flink] Fix data lake catalog connection resource is not closed in flink catalog

zhuangchong:
### Purpose

data lake catalog connection resource is not closed in flink catalog


zhuangchong:
cc @luoyuxia

PR #2096, opened 2025-12-04
[client] Change the default value of 'client.lookup.max-retries' to Integer.MAX_VALUE

### LINKED ISSUE #2095: The default value of `client.lookup.max-retries` is too small for RetriableException
### Search before asking

### Fluss version

0.8.0 (latest release)

### Please describe the bug 🐞

The default value of `client.lookup.max-retries` is too small for `RetriableException`.  For `Sender`,  the default value of `client.writer.retries` is `Interge.MAX_VALUE`, which means that for a `RetriableException` returned by the server, it's reasonable to wait until the server recovers. I think the `LookupSender` should handle it in the same way as well. 

Currently, we don't yet support retries at the `Flink` level, but this could be considered for future implementation.

### Solution

_No response_

swuferhong:
### Purpose

Linked issue: close #2095 

Change the default value of 'client.lookup.max-retries' to Integer.MAX_VALUE

### DIFF
--- a/fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/catalog/FlinkTableFactory.java
+++ b/fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/catalog/FlinkTableFactory.java
@@ -245,7 +245,8 @@ public class FlinkTableFactory implements DynamicTableSourceFactory, DynamicTabl
 
         // Todo support LookupOptions.MAX_RETRIES. Currently, Fluss doesn't support connector level
         // retry. The option 'client.lookup.max-retries' is only for dealing with the
-        // RetriableException return by server not all exceptions.
+        // RetriableException return by server not all exceptions. Trace by:
+        // https://github.com/apache/fluss/issues/2099
 
         // pass flink io tmp dir to fluss client.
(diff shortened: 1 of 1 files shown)


fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/catalog/FlinkTableFactory.java
```diff
@@ -243,12 +243,9 @@ private static Configuration toFlussClientConfig(
                     }
                 });
 
-        // map flink lookup.max-retries to client.lookup.max-retries
-        if (tableOptions.containsKey(LookupOptions.MAX_RETRIES.key())) {
-            flussConfig.setString(
-                    ConfigOptions.CLIENT_LOOKUP_MAX_RETRIES.key(),
-                    tableOptions.get(LookupOptions.MAX_RETRIES.key()));
-        }
+        // Todo support LookupOptions.MAX_RETRIES. Currently, Fluss doesn't support connector level
```
wuchong, line 246: create an issue for this, and tag as blocker for v0.9

wuchong (approved):

zhuangchong (approved):
+1

PR #2098, opened 2025-12-04
[client] Add unavailable tabletServer set for MetadataUpdater to avoid serverNode info of all tabletServers are invalid

### LINKED ISSUE #2097: SeverNode info of all tabletServers maybe invalid in client metadata when cluster upgrading
### Search before asking

### Fluss version

0.8.0 (latest release)

### Please describe the bug 🐞

There will be one situation that SeverNode info of all tabletServers maybe invalid in client metadata when cluster upgrading, which will cause the `write/read/lookup` operation block forever. Image this case:

There is a `Fluss` cluster with 3 `TabletServers` and a Fluss write job running with high parallelism. The cluster undergoes a rolling upgrade under the following conditions:

1. Pods are not upgraded in-place—their IP addresses change after restart.
2. Fluss networking has no built-in timeout mechanism and relies solely on the Netty client’s connection timeout. If a request is sent to a disconnected server, the client will wait indefinitely for the server’s response (sync acknowledgment) until the Netty timeout of 120 seconds is reached.

Below is the failure scenario:

1) **Initial state**:  
  ts-0 → 192.108.0.1  
  ts-1 → 192.108.0.2  
  ts-2 → 192.108.0.3  

2) **Upgrade starts**: ts-0 becomes unreachable. The client attempts to `updateMetadata` by sending the request to ts-0, fails to connect, and waits 120 seconds.

3) **ts-0 restarts** with a new IP: 192.108.0.4.

4) The client retries `updateMetadata`, again targeting ts-0 (based on stale metadata), and waits another 120 seconds. Meanwhile, ts-1 finishes its upgrade and gets a new IP: 192.108.0.5.

5) Another `updateMetadata` attempt is made—possibly to ts-0 or ts-1 (both now with outdated IPs in the client’s cache)—and the client waits yet another 120 seconds. At this point, ts-2 also completes its upgrade and changes its IP to 192.108.0.6.

6) **After this**, all TabletServer IPs in the client’s metadata cache are stale. No matter which server the client tries to contact, it uses an incorrect IP, causing all subsequent requests to time out. As a result, the job cannot recover automatically.

### Solution

_No response_

swuferhong:
### Purpose

Linked issue: close #2097 

Add unavailable tabletServer set for MetadataUpdater to avoid serverNode info of all tabletServers are invalid. The reason pls see #2097 .

### DIFF
--- a/fluss-client/src/main/java/org/apache/fluss/client/utils/MetadataUtils.java
+++ b/fluss-client/src/main/java/org/apache/fluss/client/utils/MetadataUtils.java
@@ -258,9 +258,11 @@ public class MetadataUtils {
 
     public static @Nullable ServerNode getOneAvailableTabletServerNode(
             Cluster cluster, Set<Integer> unavailableTabletServerIds) {
-        List<ServerNode> aliveTabletServers = cluster.getAliveTabletServerList();
-        aliveTabletServers.removeIf(
-                serverNode -> unavailableTabletServerIds.contains(serverNode.id()));
+        List<ServerNode> aliveTabletServers = new ArrayList<>(cluster.getAliveTabletServerList());
+        if (!unavailableTabletServerIds.isEmpty()) {
+            aliveTabletServers.removeIf(
+                    serverNode -> unavailableTabletServerIds.contains(serverNode.id()));
+        }
 
         if (aliveTabletServers.isEmpty()) {
             return null;
--- a/fluss-common/src/main/java/org/apache/fluss/cluster/Cluster.java
+++ b/fluss-common/src/main/java/org/apache/fluss/cluster/Cluster.java
@@ -155,7 +155,7 @@ public final class Cluster {
     }
 
     public List<ServerNode> getAliveTabletServerList() {
-        return new ArrayList<>(aliveTabletServers);
+        return aliveTabletServers;
     }
 
     /** Get the table path for this table id. */


fluss-common/src/main/java/org/apache/fluss/cluster/Cluster.java
```diff
@@ -155,7 +155,7 @@ public Map<Integer, ServerNode> getAliveTabletServers() {
     }
 
     public List<ServerNode> getAliveTabletServerList() {
-        return aliveTabletServers;
+        return new ArrayList<>(aliveTabletServers);
```
wuchong: Do not copy at here, all the methods in Cluster, doesn't do a copy. Leave the copy out of the method.

fluss-client/src/main/java/org/apache/fluss/client/utils/MetadataUtils.java
```diff
@@ -253,10 +256,14 @@ public NewTableMetadata(
         }
     }
 
-    public static ServerNode getOneAvailableTabletServerNode(Cluster cluster) {
+    public static @Nullable ServerNode getOneAvailableTabletServerNode(
+            Cluster cluster, Set<Integer> unavailableTabletServerIds) {
         List<ServerNode> aliveTabletServers = cluster.getAliveTabletServerList();
+        aliveTabletServers.removeIf(
+                serverNode -> unavailableTabletServerIds.contains(serverNode.id()));
```
wuchong: 1. copy the `cluster.getAliveTabletServerList()` into a new list
2. call `removeIf` only when `unavailableTabletServerIds` is not empty.

PR #2105, opened 2025-12-04
[KV] Support configurable change log image mode for Primary Key Tables

### LINKED ISSUE #2102: [KV] Support Ignoring UPDATE_BEFORE Records in Changelog for Primary Key Tables
### Search before asking

### Motivation

Currently, when updating data in primary key tables, the changelog produces both `UPDATE_BEFORE` (-U) and `UPDATE_AFTER` (+U) records for each update operation. In many scenarios, downstream consumers do not require the previous value information, yet they still incur the cost of storing and transmitting both records.

This feature introduces a new table-level configuration option `table.changelog.ignore-update-before` that allows users to selectively ignore `UPDATE_BEFORE` records, providing the following benefits:

- **Reduced storage costs**: Approximately 50% reduction in changelog storage space for update-heavy workloads
- **Lower transmission overhead**: Decreased CDC data transfer costs between systems
- **Improved write performance**: For full-row update scenarios (using `DefaultRowMerger`), the system can skip the RocksDB lookup for the old value entirely

### 1. New Configuration Option

Add a new configuration option in `ConfigOptions`:

```java
public static final ConfigOption<Boolean> TABLE_CHANGELOG_IGNORE_UPDATE_BEFORE =
        key("table.changelog.ignore-update-before")
                .booleanType()
                .defaultValue(false)
                .withDescription(
                        "Whether to ignore UPDATE_BEFORE records in changelog for the primary key table. "
                                + "When disabled (default), update operations produce both UPDATE_BEFORE and UPDATE_AFTER records. "
                                + "When enabled, update operations only produce UPDATE_AFTER records, "
                                + "which reduces storage and transmission costs but loses the ability to track previous values. "
                                + "This option only affects primary key tables.");
```

Provide accessor method in `TableConfig`:

```java
public boolean isChangelogIgnoreUpdateBefore() {
    return config.get(ConfigOptions.TABLE_CHANGELOG_IGNORE_UPDATE_BEFORE);
}
```

### 2. KvTablet Core Logic

Modify the `KvTablet.putAsLeader()` method to handle the new configuration:

| Scenario | `ignoreUpdateBefore=false` (default) | `ignoreUpdateBefore=true` |
|----------|--------------------------------------|---------------------------|
| Insert | Produce `INSERT` (+I) | Produce `UPDATE_AFTER` (+U)* |
| Full Update (DefaultRowMerger) | Query old value, produce `-U`, `+U` | **Skip old value query**, produce `+U` only |
| Partial Update | Query old value, merge, produce `-U`, `+U` | Query old value, merge, produce `+U` only |
| Delete | Produce `DELETE` (-D) | Produce `DELETE` (-D) (unchanged) |

*Note: When using `DefaultRowMerger` with `ignoreUpdateBefore=true`, inserts are treated as updates for simplicity since the old value lookup is skipped.

### 3. Flink Source Adaptation

Update `FlinkTableSource.getChangelogMode()` to return the appropriate `ChangelogMode` based on the configuration:

- When `ignoreUpdateBefore=true` and `deleteBehavior=ALLOW`:
  - Returns: `INSERT`, `UPDATE_AFTER`, `DELETE`
- When `ignoreUpdateBefore=true` and `deleteBehavior!=ALLOW`:
  - Returns: `INSERT`, `UPDATE_AFTER`
- When `ignoreUpdateBefore=false` (default):
  - Original behavior preserved

### Anything else?

_No response_

### Willingness to contribute

platinumhamburg:
### Purpose

Linked issue: close #2102

### DIFF
--- a/fluss-common/src/main/java/org/apache/fluss/metadata/ChangelogImage.java
+++ b/fluss-common/src/main/java/org/apache/fluss/metadata/ChangelogImage.java
@@ -46,7 +46,7 @@ public enum ChangelogImage {
 
     /** Creates a {@link ChangelogImage} from the given string. */
     public static ChangelogImage fromString(String image) {
-        switch (image.toUpperCase().replace("-", "_")) {
+        switch (image.toUpperCase()) {
             case "FULL":
                 return FULL;
             case "WAL":
--- a/fluss-server/src/test/java/org/apache/fluss/server/kv/KvTabletTest.java
+++ b/fluss-server/src/test/java/org/apache/fluss/server/kv/KvTabletTest.java
@@ -1147,14 +1147,32 @@ class KvTabletTest {
                                 "k1".getBytes(), new Object[] {1, "v1", null}));
         kvTablet.putAsLeader(kvRecordBatch2, new int[] {0, 1});
 
+        endOffset = logTablet.localLogEndOffset();
         // Verify update only produces +U (no -U since using WAL mode)
+        actualLogRecords = readLogRecords(endOffset - 1);
+        expectedLogs =
+                logRecords(
+                        rowType,
+                        endOffset - 1,
+                        Collections.singletonList(ChangeType.UPDATE_AFTER),
+                        Collections.singletonList(new Object[] {1, "v1", null}));
+        checkEqual(actualLogRecords, Collections.singletonList(expectedLogs), rowType);
+
+        // Update with partial columns (column c) - column b value should be retained
+        KvRecordBatch kvRecordBatch3 =
+                kvRecordBatchFactory.ofRecords(
+                        data2kvRecordFactory.ofRecord(
+                                "k1".getBytes(), new Object[] {1, null, "hello"}));
+        kvTablet.putAsLeader(kvRecordBatch3, new int[] {0, 2});
+
+        // Verify update produces +U with column b retained as "v1" and column c set to "hello"
         actualLogRecords = readLogRecords(endOffset);
         expectedLogs =
                 logRecords(
                         rowType,
                         endOffset,
                         Collections.singletonList(ChangeType.UPDATE_AFTER),
-                        Collections.singletonList(new Object[] {1, "v1", null}));
+                        Collections.singletonList(new Object[] {1, "v1", "hello"}));
         checkEqual(actualLogRecords, Collections.singletonList(expectedLogs), rowType);
     }
 
--- a/fluss-flink/fluss-flink-common/src/test/java/org/apache/fluss/flink/sink/FlinkTableSinkITCase.java
+++ b/fluss-flink/fluss-flink-common/src/test/java/org/apache/fluss/flink/sink/FlinkTableSinkITCase.java
@@ -1375,4 +1375,74 @@ abstract class FlinkTableSinkITCase extends AbstractTestBase {
             assertResultsIgnoreOrder(rowIter, expectedRows, true);
         }
     }
+
+    @Test
+    void testWalModeWithDefaultMergeEngineAndAggregation() throws Exception {
+        String tableName = "wal_mode_pk_table";
+        // Create a table with WAL mode and default merge engine
+        tEnv.executeSql(
+                String.format(
+                        "create table %s ("
+                                + " id int not null,"
+                                + " category string,"
+                                + " amount bigint,"
+                                + " primary key (id) not enforced"
+                                + ") with ('table.changelog.image' = 'wal')",
+                        tableName));
+
+        // Insert initial data
+        tEnv.executeSql(
+                        String.format(
+                                "INSERT INTO %s VALUES "
+                                        + "(1, 'A', 100), "
+                                        + "(2, 'B', 200), "
+                                        + "(3, 'A', 150), "
+                                        + "(4, 'B', 250)",
+                                tableName))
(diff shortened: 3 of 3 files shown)


fluss-common/src/main/java/org/apache/fluss/config/ConfigOptions.java
```diff
@@ -1423,6 +1423,17 @@ public class ConfigOptions {
                                     + "The auto increment column can only be used in primary-key table. The data type of the auto increment column must be INT or BIGINT."
                                     + "Currently a table can have only one auto-increment column.");
 
+    public static final ConfigOption<Boolean> TABLE_CHANGELOG_IGNORE_UPDATE_BEFORE =
+            key("table.changelog.ignore-update-before")
```
wuchong: Considering we already have `table.delete.behavior = ignore`, `sink.ignore-delete = true`, I'm afraid the 
`table.changelog.ignore-update-before` may confuse it is a sink operation. 

I found mysql and PG have similar configurations

```
// PG
replica identity = FULL / default / index
// mysql
binlog_row_image = FULL / minimal / nonblob
```

How about introducing a configuration like 

```
table.changelog.image = FULL / NO_UPDATE_BEFORE
```

With default to `FULL`, and we can extend it to support `NO_BLOG`, `NO_UNCHANGED` in the future. 

Besides, we should add documentation to this config that it only emits `+U` for all inserts and updates, because it avoid lookup previous values.

wuchong: @platinumhamburg I’ve reconsidered this again, and I think the name `NO_UPDATE_BEFORE` isn’t quite accurate.  

For delete changelog records, they may only contain the key to be deleted, without the full row data. However, this nuance isn’t reflected in the name `NO_UPDATE_BEFORE`.  

What we really need here is a changelog that functions like a **Write-Ahead Log (WAL)**, primarily for **durability and recovery**, not for CDC-style change data capture.  

Given that, would it be more appropriate to name this option `table.changelog.image=WAL`?

platinumhamburg: > @platinumhamburg I’ve reconsidered this again, and I think the name `NO_UPDATE_BEFORE` isn’t quite accurate.
> 
> For delete changelog records, they may only contain the key to be deleted, without the full row data. However, this nuance isn’t reflected in the name `NO_UPDATE_BEFORE`.
> 
> What we really need here is a changelog that functions like a **Write-Ahead Log (WAL)**, primarily for **durability and recovery**, not for CDC-style change data capture.
> 
> Given that, would it be more appropriate to name this option `table.changelog.image=WAL`?

Sure, I think that’s a much better design.

platinumhamburg: @wuchong I'v updated this PR，thanks for review again.

fluss-common/src/main/java/org/apache/fluss/metadata/ChangelogImage.java
```diff
@@ -0,0 +1,59 @@
...
+     * engine configured) and full row updates (not partial update), an optimization is applied to
+     * skip looking up old values, and in this case INSERT operations are converted to UPDATE_AFTER
+     * events, similar to database WAL (Write-Ahead Log) behavior. This mode reduces storage and
+     * transmission costs but loses the ability to track previous values.
+     */
+    WAL;
+
+    /** Creates a {@link ChangelogImage} from the given string. */
+    public static ChangelogImage fromString(String image) {
+        switch (image.toUpperCase().replace("-", "_")) {
```
wuchong: Why we need to replace `-` with `_`?

fluss-server/src/test/java/org/apache/fluss/server/kv/KvTabletTest.java
```diff
@@ -1044,6 +1044,120 @@ void testAppendDuplicatedKvBatch() throws Exception {
...
+        kvTablet.putAsLeader(kvRecordBatch2, new int[] {0, 1});
+
+        // Verify update only produces +U (no -U since using WAL mode)
+        actualLogRecords = readLogRecords(endOffset);
+        expectedLogs =
+                logRecords(
+                        rowType,
+                        endOffset,
+                        Collections.singletonList(ChangeType.UPDATE_AFTER),
+                        Collections.singletonList(new Object[] {1, "v1", null}));
```
wuchong: It would be better to add a partial-update for column `c` as well, and the result should retain value `v1` for column b. Otherwise, it's hard to verify the behavior of partial update, because currently all the non-updated columns are null.

wuchong:
I appended a commit to add back some code comments of `KvTablet` which I think may have been removed by mistake during refactoring. Please take a look.

fluss-common/src/main/java/org/apache/fluss/metadata/ChangelogImage.java
platinumhamburg: Sorry, this part is legacy code. In the legacy code, some enum names were kept with a “-”. I missed removing it during the refactor; I’ll delete it.

platinumhamburg:
@wuchong I'v updated this PR and resolved all comments, thanks for review again.

PR #2115, opened 2025-12-06
[FLUSS-2113][build] Upgrade Deployed Maven to Latest Stable Release

### LINKED ISSUE #2113: Upgrade Maven to Latest Supported Version
### Search before asking

### Description

While working on another unrelated issue, I encountered the following error when attempting to build Fluss using the existing `./mvnw clean package -DskipTests` command:

```
./mvnw clean package -DskipTests
Error: Failed to validate Maven distribution SHA-256, your Maven distribution might be compromised.
If you updated your Maven version, you need to update the specified distributionSha256Sum property.
```

After digging into it a bit deeper, it seems that the existing version referenced within the `maven-wrapper.properties` file has reached EOF as detailed in [the Maven releases documentation](https://maven.apache.org/docs/history.html#maven-3-8-x-and-before):

> Maven 3.8.9 and before has now reached its end of life and is no longer supported in any way. New plugin releases will require Maven 3.9.0 or later. The following information is made available for reference.

We may want to consider upgrading to the latest 3.9.x release to ensure that a supported version of Maven is being used for the project.

### Willingness to contribute

rionmonster:
### Purpose

Linked issue: close #2113 

Per Issue https://github.com/apache/fluss/issues/2113, this pull request upgrades the existing Maven version referenced within Fluss to the latest stable release. The motivation behind this being that the current version (3.8.6) is no longer supported or maintained.

### Brief change log
Upgraded the Maven wrapper using the built-in utility call to target the latest stable release and generate all of the necessary artifacts:
```
./mvnw -N wrapper:wrapper -Dmaven=3.9.11
```

### Tests
N/A

### API and Format
N/A

### Documentation
N/A

### DIFF
--- a/.mvn/wrapper/maven-wrapper.properties
+++ b/.mvn/wrapper/maven-wrapper.properties
@@ -17,4 +17,4 @@
 wrapperVersion=3.3.4
 distributionType=only-script
 distributionUrl=https://repo.maven.apache.org/maven2/org/apache/maven/apache-maven/3.9.11/apache-maven-3.9.11-bin.zip
-distributionSha512Sum=bcfe4fe305c962ace56ac7b5fc7a08b87d5abd8b7e89027ab251069faebee516b0ded8961445d6d91ec1985dfe30f8153268843c89aa392733d1a3ec956c9978
+distributionSha256Sum=0d7125e8c91097b36edb990ea5934e6c68b4440eef4ea96510a0f6815e7eeadb


rionmonster:
@wuchong 

I was looking into this issue and based on the previously touched files, I figured that you'd be the person to ask. I was running into local issues when working on a separate issue related to Maven and noticed the version we are targeting is no longer supported.

I attempted to upgrade it (along with its references), but noticed that this caused the Maven Enforcer plugin to fail the CI builds for Java 11 since it defaults to Maven 3.8.6, which doesn't fall into the updated enforcement range introduced in this pull request (from `<version>[3.8.6,)</version>` to `<version>[3.9.11,)</version>`). I'm not certain what the best course of action here would be, so I wanted your feedback.

My initial approaches would be:
- **Leverage Maven Wrapper for CI** - Currently the CI builds call Maven directly (e.g. `mvn ...`) however I'm curious if we could update the CI template to use the wrapper as well (uncertain how feasible this is) for consistency across the different builds.
- **Specifying Build Version** - Update the existing Java 11 CI builds to explicitly set the Maven version to the new target of 3.9.11.
- **Relaxing Enforcement** - Just revert the target enforcement version to the previous 3.8.6 to support these older versions. IMO, not ideal since I think we want to get away from those deprecated versions everywhere, but it would support the CI use case.

rionmonster:
Closing based on discussion within #2113 to keep the target version of Fluss in line with that of Flink (3.8.6).

PR #2126, opened 2025-12-08
[server] support database level buckets limit

### LINKED ISSUE #1993: Support maximum number of partitions and buckets at the database level
### Search before asking

### Motivation

During the use of Fluss, I've observed that many users configure an excessively large number of `partitions or buckets number`—even when their actual throughput is low. For log tables, a large number of `partitions/buckets` doesn't consume significant memory. However, for PrimaryKey tables, each bucket corresponds to a separate RocksDB instance, and even when no data is written, each RocksDB instance still consumes approximately 15 MB of memory, placing a heavy burden on the system.

Although Fluss currently supports cluster-wide limits on the maximum number of partitions (max.partition.num) and buckets (max.bucket.num), this approach lacks flexibility and is insufficient for effectively controlling cluster load.

Fluss does not support multi-tenancy. Therefore, we currently treat databases as the tenant boundary in Fluss. We propose introducing per-database limits on the maximum number of partitions and buckets. This would allow us to set a small initial quota during cluster setup and later relax the limits based on individual customer needs—effectively implementing a tenant-level throttling strategy similar to those in systems like Kafka that support true multi-tenancy.

The implementation can be achieved through dynamic configuration parameters （alterClusterConfigs）, allowing cluster administrators to adjust the partition/bucket limits for different databases by updating this dynamic configuration.

### Solution

_No response_

### Anything else?

_No response_

### Willingness to contribute

caozhen1937:
### Purpose

Linked issue: close #1993

### DIFF
--- a/fluss-flink/fluss-flink-common/src/test/java/org/apache/fluss/flink/catalog/FlinkCatalogITCase.java
+++ b/fluss-flink/fluss-flink-common/src/test/java/org/apache/fluss/flink/catalog/FlinkCatalogITCase.java
@@ -17,13 +17,19 @@
 
 package org.apache.fluss.flink.catalog;
 
+import org.apache.fluss.client.Connection;
+import org.apache.fluss.client.ConnectionFactory;
+import org.apache.fluss.client.admin.Admin;
 import org.apache.fluss.cluster.ServerNode;
 import org.apache.fluss.config.ConfigOptions;
 import org.apache.fluss.config.Configuration;
+import org.apache.fluss.config.cluster.AlterConfig;
+import org.apache.fluss.config.cluster.AlterConfigOpType;
 import org.apache.fluss.exception.FlussRuntimeException;
 import org.apache.fluss.exception.InvalidAlterTableException;
 import org.apache.fluss.exception.InvalidConfigException;
 import org.apache.fluss.exception.InvalidTableException;
+import org.apache.fluss.exception.TooManyBucketsException;
 import org.apache.fluss.metadata.DataLakeFormat;
 import org.apache.fluss.metadata.TablePath;
 import org.apache.fluss.server.testutils.FlussClusterExtension;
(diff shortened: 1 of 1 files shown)


fluss-server/src/main/java/org/apache/fluss/server/coordinator/MetadataManager.java
```diff
@@ -279,7 +284,12 @@ public long createTable(
             boolean ignoreIfExists)
             throws TableAlreadyExistException, DatabaseNotExistException {
         // validate table properties before creating table
-        validateTableDescriptor(tableToCreate, maxBucketNum);
+        int dbLevelMaxBucket =
+                DatabaseLimitResolver.resolveMaxBucketForDb(
+                        maxBucketNum,
+                        dynamicConfigManager.describeConfigs(),
+                        tablePath.getDatabaseName());
+        validateTableDescriptor(tableToCreate, maxBucketNum, dbLevelMaxBucket);
```
swuferhong: We only need to pass the value returned by `DatabaseLimitResolver.resolveMaxBucketForDb` into `validateTableDescriptor`, so the separate `maxBucketNum` parameter can be removed. Additionally, could we rename the return value of `DatabaseLimitResolver.resolveMaxBucketForDb` to `maxBucketNum` for clarity?

fluss-server/src/main/java/org/apache/fluss/server/coordinator/MetadataManager.java
```diff
@@ -712,12 +727,26 @@ public void createPartition(
             if (totalBuckets > maxBucketNum) {
                 throw new TooManyBucketsException(
                         String.format(
-                                "Adding partition '%s' would result in %d total buckets for table %s, exceeding the maximum of %d buckets.",
+                                "Adding partition '%s' would result in %d total buckets for table %s, exceeding the cluster-level maximum of %d buckets.",
```
swuferhong: This check can be removed, only retain the check after resolved by `DatabaseLimitResolver`.

fluss-server/src/main/java/org/apache/fluss/server/utils/DatabaseLimitResolver.java
```diff
@@ -0,0 +1,62 @@
...
+import java.util.List;
+import java.util.Objects;
+
+import static org.apache.fluss.server.DynamicServerConfig.DATABASE_LIMITS_PREFIX;
+
+/**
+ * Analyze the database-level quota upper limit, and calculate the effective bucket and partition
+ * upper limit of each database based on cluster configuration and dynamic configuration.
+ */
+public final class DatabaseLimitResolver {
```
swuferhong, line 31: Adding tests to cover this class.

fluss-server/src/main/java/org/apache/fluss/server/utils/TableDescriptorValidation.java
```diff
@@ -163,9 +165,15 @@ private static void checkDistribution(TableDescriptor tableDescriptor, int maxBu
         if (bucketCount > maxBucketNum) {
             throw new TooManyBucketsException(
                     String.format(
-                            "Bucket count %s exceeds the maximum limit %s.",
+                            "Bucket count %s exceeds the cluster-level maximum limit %s.",
                             bucketCount, maxBucketNum));
```
swuferhong: This check can be removed.

fluss-server/src/main/java/org/apache/fluss/server/coordinator/MetadataManager.java
caozhen1937: @swuferhong `maxBucketNum` is still used to represent the maximum limit of the cluster. I will use two variables to distinguish them: `maxBucketNumOfDb` and `maxBucketNumOfCluster`.  How do you think?

swuferhong: > @swuferhong `maxBucketNum` is still used to represent the maximum limit of the cluster. I will use two variables to distinguish them: `maxBucketNumOfDb` and `maxBucketNumOfCluster`. How do you think?

What concerns me most is the implementation. Theoretically, the cluster-level logic should be holistic—when creating a table, it should first check how many buckets currently exist in the cluster and then determine whether the table can be created. However, the current cluster-level implementation actually checks individual tables during creation. 

I think it's reasonable to separate these two aspects, and future expansion maybe consider for the cluster level limit.

swuferhong:
Hi, @caozhen1937  LGTM now. Could you add a test in Flink side call as procedure?

fluss-client/src/test/java/org/apache/fluss/client/admin/FlussAdminITCase.java
```diff
@@ -1403,7 +1403,169 @@ public void testBucketLimitForNonPartitionedTable() throws Exception {
...
+                                Schema.newBuilder()
+                                        .column("id", DataTypes.STRING())
+                                        .column("name", DataTypes.STRING())
+                                        .build())
+                        .distributedBy(20, "id")
+                        .build();
+
+        assertThatThrownBy(() -> admin.createTable(tablePath, tooManyBuckets, false).get())
+                .cause()
+                .isInstanceOf(TooManyBucketsException.class);
```
wuchong, line 1436: assert the message as well, the message should be clear to user the given database has exceed the allowed number of buckets.

fluss-client/src/test/java/org/apache/fluss/client/admin/FlussAdminITCase.java
```diff
@@ -1403,7 +1403,169 @@ public void testBucketLimitForNonPartitionedTable() throws Exception {
...
+        admin.createPartition(tablePath, newPartitionSpec("age", "1"), false).get();
+        admin.createPartition(tablePath, newPartitionSpec("age", "2"), false).get();
+
+        assertThatThrownBy(
+                        () ->
+                                admin.createPartition(
+                                                tablePath, newPartitionSpec("age", "3"), false)
+                                        .get())
+                .cause()
+                .isInstanceOf(TooManyBucketsException.class);
```
wuchong, line 1486: ditto

fluss-server/src/main/java/org/apache/fluss/server/coordinator/MetadataManager.java
```diff
@@ -709,14 +724,19 @@ public void createPartition(
...
+            if (totalBuckets > maxBucketNumOfDb) {
                 throw new TooManyBucketsException(
                         String.format(
-                                "Adding partition '%s' would result in %d total buckets for table %s, exceeding the maximum of %d buckets.",
+                                "Adding partition '%s' would result in %d total buckets for table %s, exceeding the database-level maximum of %d buckets.",
                                 partition.getPartitionName(),
                                 totalBuckets,
                                 tablePath,
-                                maxBucketNum));
+                                maxBucketNumOfDb));
```
wuchong, line 739: It seems we simply drops the support of `maxBucketNumOfCluster` or `maxBucketNum`. This is backward imcompatible, we shouldn't remove them.

swuferhong:
Hi, @wuchong. A small input regarding "Implement a CALL procedure in Flink SQL to dynamically alter cluster-level configs (e.g., database bucket limits)": the PR #2178 will support using a CALL procedure to dynamically set cluster-level configurations. However, I'm still testing the parameters introduced in #2178, and I'll leave a comment on the PR once testing is complete.

Issue #2086, opened 2025-12-03
FlussConnection cann't get table while the request server not ready


swuferhong:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/fluss/issues) and found nothing similar.

### Fluss version

0.8.0 (latest release)

### Please describe the bug 🐞

```
2025-12-03 18:11:40:
java.util.concurrent.CompletionException: org.apache.fluss.exception.NetworkException: org.apache.fluss.exception.DisconnectException: Cannot send request to server xxxx:9123 (id: ts-2, rack: null) because it is disconnected.
	at java.base/java.util.concurrent.CompletableFuture.encodeThrowable(CompletableFuture.java:331)
	at java.base/java.util.concurrent.CompletableFuture.uniApplyNow(CompletableFuture.java:670)
	at java.base/java.util.concurrent.CompletableFuture.uniApplyStage(CompletableFuture.java:658)
	at java.base/java.util.concurrent.CompletableFuture.thenApply(CompletableFuture.java:2100)
	at org.apache.fluss.client.admin.FlussAdmin.getTableInfo(FlussAdmin.java:264)
	at org.apache.fluss.client.FlussConnection.getTable(FlussConnection.java:106)
	at org.apache.fluss.flink.source.reader.FlinkSourceSplitReader.<init>(FlinkSourceSplitReader.java:129)
	at org.apache.fluss.flink.source.reader.FlinkSourceReader.lambda$new$0(FlinkSourceReader.java:69)
	at org.apache.flink.connector.base.source.reader.fetcher.SplitFetcherManager.createSplitFetcher(SplitFetcherManager.java:259)
	at org.apache.flink.connector.base.source.reader.fetcher.SingleThreadFetcherManager.addSplits(SingleThreadFetcherManager.java:148)
	at org.apache.flink.connector.base.source.reader.SourceReaderBase.addSplits(SourceReaderBase.java:341)
	at org.apache.flink.streaming.api.operators.SourceOperator.handleAddSplitsEvent(SourceOperator.java:694)
	at org.apache.flink.streaming.api.operators.SourceOperator.handleOperatorEvent(SourceOperator.java:638)
	at org.apache.flink.streaming.runtime.tasks.OperatorEventDispatcherImpl.dispatchEventToHandlers(OperatorEventDispatcherImpl.java:72)
	at org.apache.flink.streaming.runtime.tasks.RegularOperatorChain.dispatchOperatorEvent(RegularOperatorChain.java:80)
[...]

Issue #483, opened 2025-02-25
Refactor metadata cache and async updating mechanism in client side


wuchong:
### Search before asking

- [x] I searched in the [issues](https://github.com/alibaba/fluss/issues) and found nothing similar.

### Motivation

Currently, the implementation of metadata cache (`Cluster`) and updating mechanism (`MetadataUpdater`) on the client side is not very sophisticated. It lacks the general retry mechanism (#480), the time out for `UpdateMetadata` request is too short to cause job failover (#311), cause job failover when a new partition created (#288).

### Solution

This issue aims to introduce a new design of metadata cache and updating mechanism, to resolve above problems.

### Willingness to contribute

- [x] I'm willing to submit a PR!

swuferhong:
Note: we need to refactor coordinatorService#getTableMetadata() and TabletServerMetadataCache#getPartitionMetadata() in order to avoid get table assignment info from zk while no table assignment in server cache.

wuchong:
For reference: https://github.com/apache/fluss/pull/479 introduced a `AdminDriver` for retry mechanism.

PR #479, opened 2025-02-25
[client] Introduce AdminApiDriver for workflows which lookup broker first then send request.

loserwang1024:
### Purpose

Linked issue: close #288 

Inspired by https://issues.apache.org/jira/browse/KAFKA-12609, which introduce AdminApiDriver do theses things:

- Retry lookup stage on receiving `NOT_LEADER_OR_FOLLOWER` and `LEADER_NOT_AVAILABLE`, whereas in the past we failed the partition directly without retry.

-  retry fulfillment stage on `RetriableException`

### Tests

- AdminApiDriverTest for AdminApiDriver
- com.alibaba.fluss.client.admin.FlussAdminITCase#testListOffsets for Admin#ListOffsets

### API and Format

Some internal interface:

- AdminApiDriver
- AdminApiHandler
- AdminApiFuture
- AdminApiLookupBrokerIdStrategy

PR #2132, opened 2025-12-09
[RPC] Add a retry mechanism for idempotent interface requests

sd4324530:
### Purpose

Linked issue: close https://github.com/apache/fluss/issues/2086

### Brief change log

Add a retry mechanism for idempotent interface requests.

### DIFF
--- a/fluss-client/src/main/java/org/apache/fluss/client/admin/FlussAdmin.java
+++ b/fluss-client/src/main/java/org/apache/fluss/client/admin/FlussAdmin.java
@@ -134,7 +134,7 @@ public class FlussAdmin implements Admin {
                         metadataUpdater::getCoordinatorServer, client, AdminGateway.class);
         this.readOnlyGateway =
                 GatewayClientProxy.createGatewayProxy(
-                        metadataUpdater::getRandomTabletServer, client, AdminGateway.class);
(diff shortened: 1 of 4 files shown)


wuchong:
Thank you for your contribution! However, the retry mechanism we refer to here is not simply a retry of the request itself, but rather a two-stage retry process: the first stage involves locating a node, and the second stage sends the request. In many cases, the issue is that the node is unavailable, so we need to fall back to the first stage and retry by selecting a new node (see https://github.com/apache/fluss/pull/479 for details). This will require a fairly significant refactoring and is already planned as part of that issue (https://github.com/apache/fluss/issues/483). Therefore, I’ll close this pull request for now. Thanks again for your contribution!

sd4324530:
> Thank you for your contribution! However, the retry mechanism we refer to here is not simply a retry of the request itself, but rather a two-stage retry process: the first stage involves locating a node, and the second stage sends the request. In many cases, the issue is that the node is unavailable, so we need to fall back to the first stage and retry by selecting a new node (see #479 for details). This will require a fairly significant refactoring and is already planned as part of that issue (#483). Therefore, I’ll close this pull request for now. Thanks again for your contribution!

Thank you very much for your detailed explanation.

PR #2134, opened 2025-12-09
Support Exactly-Once Semantics for Aggregation Merge Engine

### LINKED ISSUE #2133: Support Exactly-Once Semantics for Aggregation Merge Engine
### Search before asking

### Motivation

When using Aggregation Merge Engine with multiple concurrent Flink jobs writing to the same primary key table, the following issues occur:

1. **Concurrent Conflicts**: Multiple writers updating the same aggregation columns simultaneously cause data inconsistency
2. **No Exactly-Once Guarantee**: Job failover cannot guarantee aggregation accuracy
3. **Duplicate Calculations**: Data replay after job restart leads to duplicate aggregation
4. **State Loss**: Lack of state management to track committed data

These issues make Aggregation Merge Engine unsuitable for production scenarios requiring strict data accuracy.

### Solution

The complete solution has been proposed in the FIP document [FIP-21: Aggregation Merge Engine](https://cwiki.apache.org/confluence/display/FLUSS/FIP-21%3A+Aggregation+Merge+Engine).

### 1. Column Lock Mechanism
- **Client-side**: `ClientColumnLockManager` manages lock acquisition, renewal, and release
- **Server-side**: `ServerColumnLockService` maintains table-level column lock resources
- **Purpose**: Ensures exclusive access to columns, preventing concurrent modification conflicts
- **Granularity**: Table-level locks with column-level control

### 2. State Management
- **WriterState**: Tracks maximum committed offset per TableBucket
- **BucketOffsetTracker**: Client-side offset tracking
- **Integration**: Uses Flink's Operator State for persistence across checkpoints

### 3. Undo Recovery
- **Mechanism**: On failover, undoes uncommitted data written after last checkpoint
- **Implementation**: Reads old values at committed offset and overwrites new values
- **Guarantee**: Ensures data consistency after job restart

### 4. Recovery Mode
- **Purpose**: Special connection mode for undo operations
- **Behavior**: Uses overwrite semantics instead of aggregation during recovery
- **Usage**: Automatically enabled during undo recovery process

### Anything else?

_No response_

### Willingness to contribute

platinumhamburg:
### Purpose

Linked issue: close #2133 2133


fluss-client/src/main/java/org/apache/fluss/client/table/writer/UpsertWriterImpl.java
```diff
@@ -186,6 +310,35 @@ public CompletableFuture<DeleteResult> delete(InternalRow row) {
...
     }
 
+    @Override
+    public Map<TableBucket, Long> bucketsAckStatus() {
+        // Get bucket offset tracker from writer client
+        // The tracker maintains the latest acknowledged offset for each bucket written by this
+        // writer
+        try {
+            Map<TableBucket, Long> allOffsets =
+                    writerClient.getBucketOffsetTracker().getAllOffsets();
```
Jackeyzhe, line 320: I'm not sure the size of `allOffsets`, if it's big, it's better to use a new method like getOffsetsByTableId, it can reduce data size from server.

fluss-client/src/main/java/org/apache/fluss/client/write/ClientColumnLockManager.java
```diff
@@ -0,0 +1,1214 @@
...
+                                                        // Always succeed to allow other releases to
+                                                        // continue
+                                                        return null;
+                                                    }));
+        }
+
+        // Wait for all releases and log summary
+        return chain.thenRun(
+                () -> {
+                    if (!failedServers.isEmpty()) {
```
Jackeyzhe, line 511: Do `failedServers` need to retry util they release success?

fluss-client/src/main/java/org/apache/fluss/client/table/writer/UpsertWriterImpl.java
platinumhamburg: The offset here is tracked by the client based on the response returned for each write; therefore, the client actually retains the offsets of the buckets it has written to.

fluss-client/src/main/java/org/apache/fluss/client/write/ClientColumnLockManager.java
platinumhamburg: Thanks for the review comments. Since the lock has a TTL, we didn’t adopt an aggressive retry strategy; instead, we treated the TTL as a general fallback. However, it is indeed more reasonable to use a more fine-grained retry strategy based on the specific exception.

Also, per the FIP discussion, to reduce complexity, ColumnLock may be temporarily removed from the Phase 1 implementation, so I may trim down this PR accordingly.

Jackeyzhe: Thanks for the response, I'll read the FIP later :)

wuchong:
closing this PR as we will separate it into mulitple PRs

PR #2145, opened 2025-12-11
[Coordinator] Log the reason for failures in bucket or replica state changes

### LINKED ISSUE #2144: [Coordinator] Log the reason for the replica state change failure
### Search before asking

### Motivation

Currently, coordinator server logs replica state change failures using logFailedSateChange(), which records only the event without detailed reason, makes hard to diagnose the online failure. 

### Solution

Add more detailed failure reason.

### Anything else?

_No response_

### Willingness to contribute

platinumhamburg:
### Purpose

Linked issue: close #2144

### DIFF
--- a/fluss-server/src/main/java/org/apache/fluss/server/coordinator/statemachine/ReplicaStateMachine.java
+++ b/fluss-server/src/main/java/org/apache/fluss/server/coordinator/statemachine/ReplicaStateMachine.java
@@ -221,7 +221,8 @@ public class ReplicaStateMachine {
                                 try {
                                     partitionName = getPartitionName(tableBucket);
                                 } catch (PartitionNotExistException e) {
-                                    logFailedSateChange(replica, currentState, targetState, e);
+                                    logFailedSateChange(
+                                            replica, currentState, targetState, e.getMessage());
                                     return;
                                 }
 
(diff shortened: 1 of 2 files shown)


fluss-server/src/main/java/org/apache/fluss/server/coordinator/statemachine/ReplicaStateMachine.java
```diff
@@ -376,12 +380,16 @@ private void logInvalidTransition(
...
+            ReplicaState currState,
+            ReplicaState targetState,
+            Throwable reason) {
         LOG.error(
                 "Fail to change state for table bucket replica {} from {} to {}.",
                 stringifyReplica(replica),
                 currState,
-                targetState);
+                targetState,
+                reason);
```
wuchong, line 392: We previously only logs the exception message, however, the new version prints the full stack of the exception which may make the error message complex and verbose. However still keeping log error message? In this way, we can also simplify the method without creating an Exception object for it.

platinumhamburg: Thanks for your comment, I'll refactor the code.

platinumhamburg: I'v updated this PR.

PR #2154, opened 2025-12-11
[flink] Flink Source need to check tableId on recovery in case that table is recreated.

### LINKED ISSUE #2140: Flink Source need to check tableId on recovery in case that table is removed and created.
### Search before asking

### Fluss version

0.8.0 (latest release)

### Please describe the bug 🐞

Now, if we removed a table  and recreated a same name table, fluss will read new table with old table's split. Thus, If we founded tableId in fluss different with split, just throw exception

### Solution

_No response_

loserwang1024:
### Purpose

Linked issue: close #2140

### DIFF
--- a/fluss-flink/fluss-flink-common/src/test/java/org/apache/fluss/flink/source/reader/FlinkSourceSplitReaderTest.java
+++ b/fluss-flink/fluss-flink-common/src/test/java/org/apache/fluss/flink/source/reader/FlinkSourceSplitReaderTest.java
@@ -174,9 +174,10 @@ class FlinkSourceSplitReaderTest extends FlinkTestBase {
                                                                     null,
                                                                     0)))))
                     .hasMessageContaining(
-                            String.format(
-                                    "the same name table `%s` is dropped and recreated before the job restarted, which maybe cause inconsistency, please restarted the job without savepoint or checkpoint",
-                                    tablePath));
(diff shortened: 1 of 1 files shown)


loserwang1024:
@swuferhong @wuchong , CC

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/source/reader/FlinkSourceSplitReader.java
```diff
@@ -188,15 +188,10 @@ public void handleSplitsChanges(SplitsChange<SourceSplitBase> splitsChanges) {
...
-                tableId = sourceSplitBase.getTableBucket().getTableId();
-            } else {
-                checkArgument(
-                        tableId.equals(sourceSplitBase.getTableBucket().getTableId()),
-                        "table id not equal across splits {}",
-                        splitsChanges.splits());
-            }
+            checkArgument(
+                    tableId.equals(sourceSplitBase.getTableBucket().getTableId()),
+                    "table id not equal across splits {}",
```
wuchong: Make the exception message more clear to users, what table path, what is the different table ids, why this error happens, how to resolve this error.

wuchong (approved):

PR #2155, opened 2025-12-11
[iceberg] Use DataFileSet / DeleteFileSet instead of normal collections

nastra:
### Purpose

For Iceberg V3 with Deletion Vectors we introduced `DataFileSet` & `DeleteFileSet` in Iceberg so that a `DeleteFile` is correctly compared by looking at its `location` / `contentOffset` / `contentSizeInBytes` (https://github.com/apache/iceberg/blob/ec269ee3ec0de4184eb536a6ef4f3523dc91332a/api/src/main/java/org/apache/iceberg/util/DeleteFileSet.java#L100-L102). Using `DataFileSet` prevents duplicate data file instances in collections. Therefore we should switch using to `DataFileSet` / `DeleteFileSet` everywhere to be future-proof.


luoyuxia (approved):
@nastra Thank you very much from the perspective of iceberg. Will never been aware of it without you.

nastra:
thanks for reviewing and merging @luoyuxia

PR #2119, opened 2025-12-07
[kv/auto increment column]: Add an implementation of AutoIncIDBuffer

vamossagar12:
### Purpose

Linked issue: close #2111

### Brief change log

Adding code for AutoIncIDBuffer in fluss-server as described in [FIP-16](https://cwiki.apache.org/confluence/display/FLUSS/FIP-16%3A+Auto-Increment+Column). This just adds changes for the buffer, integration may happen in a subsequent PR.

### Tests

Added some Unit tests to test some of the scenarios.

### API and Format

No.

### Documentation

If is related to [FIP-16](https://cwiki.apache.org/confluence/display/FLUSS/FIP-16%3A+Auto-Increment+Column)

PR #1887, opened 2025-10-29
[kv/auto increment column] add auto increment column in Fluss schema

xx789633:
### Purpose

Linked issue: close #2103

PR #2161, opened 2025-12-12
[kv] Add an implementation of AutoIncIDBuffer

### LINKED ISSUE #2111: [kv/auto increment column] Implement AutoIncIDBuffer on tablet server
### Search before asking

### Motivation

This is a sub-task of Umbrella issue #1886 , aimed at implement AutoIncIDBuffer on tablet server.

### Solution

_No response_

### Anything else?

_No response_

### Willingness to contribute

beryllw:
### Purpose

Linked issue: close #2111


fluss-common/src/main/java/org/apache/fluss/metadata/Schema.java
```diff
@@ -115,6 +115,15 @@ public int[] getPrimaryKeyIndexes() {
                 .orElseGet(() -> new int[0]);
     }
 
+    /** Returns the auto-increment column index, if any, otherwise returns -1. */
```
xx789633: This PR does not need to modify the schema part. Please use the latest schema definition code containing the auto increment column on main branch.

fluss-common/src/test/java/org/apache/fluss/metadata/TableDescriptorTest.java
```diff
@@ -316,4 +316,7 @@ void testPartitionedTable() {
                 .hasMessage(
                         "Bucket key [f0, f3] shouldn't include any column in partition keys [f0].");
     }
+
```
xx789633, line 320: remove this

fluss-server/src/main/java/org/apache/fluss/server/kv/autoinc/SegmentIDGenerator.java
```diff
@@ -0,0 +1,195 @@
...
+        this.columnName = columnName;
+        this.tablePath = tablePath;
+        this.schemaId = schemaId;
+        this.columnIdx = columnIdx;
+        this.fetchIdScheduler = scheduler;
+        this.segments = new AutoIncIdSegment[] {new AutoIncIdSegment(), new AutoIncIdSegment()};
+        this.sequenceIDCounter = sequenceIDCounter;
+    }
+
+    private void prefetchSegmentIds(int segmentIdx) {
```
xx789633: I think we do not need two segments or prefetching optimizations in this version considering we plan to introduce bucket state checkpoint to avoid holes in ID sequence. Just get the next batch of IDs from zk when local one is exhausted.

beryllw: Considering this scenario, the optimization of two segments is indeed not appropriate. May be, we could discard the ID sequence used in the previous snapshot after completing the new snapshot, and retain the IDs associated with the current snapshot as well as unused IDs.

fluss-common/src/main/java/org/apache/fluss/metadata/Schema.java
beryllw: Add this function to quickly retrieve the auto-increment column. It seems that there are no relevant changes in the latest code of the main branch?

vamossagar12:
hi, I had created this Pr: https://github.com/apache/fluss/pull/2119 for this issue. Do we want to prefer this one over that?

xx789633:
Hi @vamossagar12 , we prefer this one to work on since it is more comprehensive.

fluss-server/src/main/java/org/apache/fluss/server/utils/TableDescriptorValidation.java
```diff
@@ -109,6 +111,40 @@ public static void validateTableDescriptor(TableDescriptor tableDescriptor, int
         checkTieredLog(tableConf);
         checkPartition(tableConf, tableDescriptor.getPartitionKeys(), schema);
         checkSystemColumns(schema);
+        checkAutoIncColumns(tableDescriptor, tableConf);
```
xx789633: Remove these checks. We already have auto increment schema checks in schema.normalizeColumns.

fluss-common/src/main/java/org/apache/fluss/utils/json/ColumnJsonSerde.java
```diff
@@ -50,6 +51,10 @@ public void serialize(Schema.Column column, JsonGenerator generator) throws IOEx
             generator.writeStringField(COMMENT, column.getComment().get());
         }
         generator.writeNumberField(ID, column.getColumnId());
+        if (column.isAutoInc()) {
```
xx789633: According to Jark's suggestion in https://github.com/apache/fluss/pull/1887#discussion_r2581883655, we prefer not to introduce any extra metadata information at the column level.

fluss-common/src/main/java/org/apache/fluss/config/ConfigOptions.java
```diff
@@ -1424,6 +1424,20 @@ public class ConfigOptions {
                                     + "The `first_row` merge engine will keep the first row of the same primary key. "
                                     + "The `versioned` merge engine will keep the row with the largest version of the same primary key.");
 
+    public static final ConfigOption<Double> TABLE_AUTO_INC_PREFETCH_LOW_WATER_MARK_RATIO =
```
xx789633: For the first version of the auto increment column, we prefer to discard those prefetching or two-segment optimization techniques.

fluss-common/src/main/java/org/apache/fluss/metadata/Schema.java
```diff
@@ -614,7 +633,7 @@ private static List<Column> normalizeColumns(
                 newColumns.add(
                         new Column(
                                 column.getName(),
-                                column.getDataType().copy(false),
+                                column.getDataType().copy(true),
```
beryllw: Partial update is required, so the data type here needs to be temporarily set to nullable.

fluss-server/src/main/java/org/apache/fluss/server/kv/autoinc/AutoIncProcessor.java
```diff
@@ -0,0 +1,88 @@
...
+
+    static AutoIncProcessor create(
+            TablePath tablePath,
+            int schemaId,
+            Configuration properties,
+            TableConfig tableConf,
+            Schema schema,
+            ZooKeeperClient zkClient) {
+        int[] autoIncColumnIndexes = schema.getAutoIncColumnIndexes();
+        if (autoIncColumnIndexes.length > 1) {
```
xx789633: We already check this in enableAutoIncrement(), do we need to do it again here?

fluss-server/src/main/java/org/apache/fluss/server/zk/data/ZkData.java
```diff
@@ -252,6 +252,41 @@ public static String path() {
...
+    public static final class ColumnZNode {
+        public static String path(TablePath tablePath, int schemaId, int columnIdx) {
+            return ColumnsZNode.path(tablePath, schemaId) + "/" + columnIdx;
+        }
+    }
+
+    /**
+     * The znode for auto increment column. The znode path is:
+     *
+     * <p>/metadata/databases/[databaseName]/tables/[tableName]/schemas/[schemaId]/columns/[columnIdx]/auto_increment_counter
```
xx789633: according to FIP, the znode path should be /metadata/databases/[databaseName]/tables/[tableName]/auto_inc/col_[columnIdx]

fluss-server/src/main/java/org/apache/fluss/server/zk/data/ZkData.java
```diff
@@ -252,6 +252,30 @@ public static String path() {
...
+            return TableZNode.path(tablePath) + "/auto_inc";
+        }
+    }
+
+    /**
+     * The znode for auto increment column. The znode path is:
+     *
+     * <p>/metadata/databases/[databaseName]/tables/[tableName]/auto_inc/col_[columnIdx]
+     */
+    public static final class AutoIncrementColumnZNode {
```
platinumhamburg, line 271: Schema evolution (adding a column) is already supported; It would be better to use column IDs instead of column indices here.

fluss-server/src/main/java/org/apache/fluss/server/zk/ZkSequenceIDCounter.java
```diff
@@ -59,4 +59,19 @@ public long getAndIncrement() throws Exception {
...
+     *
+     * @return The previous sequence ID
+     */
+    @Override
+    public long getAndAdd(Long delta) throws Exception {
+        AtomicValue<Long> incrementValue = sequenceIdCounter.add(delta);
+        if (incrementValue.succeeded()) {
+            return incrementValue.preValue();
+        } else {
+            throw new Exception("Failed to add sequence id counter.");
```
platinumhamburg: Perhaps we could make the exception information clearer—for example, by including the sequenceIDPath.

fluss-server/src/main/java/org/apache/fluss/server/kv/autoinc/AutoIncProcessor.java
```diff
@@ -0,0 +1,83 @@
...
+     * @return the processed value.
+     */
+    BinaryValue processAutoInc(BinaryValue originalValue);
+
+    static AutoIncProcessor create(
+            TablePath tablePath,
+            int schemaId,
+            Configuration properties,
+            TableConfig tableConf,
+            Schema schema,
```
platinumhamburg: Use SchemaGetter instead of Schema.

fluss-server/src/main/java/org/apache/fluss/server/kv/autoinc/AutoIncProcessor.java
```diff
@@ -0,0 +1,83 @@
...
+            ZooKeeperClient zkClient) {
+        int[] autoIncColumnIndexes = schema.getAutoIncColumnIndexes();
```
platinumhamburg: Use column IDs instead of column indices.

fluss-server/src/main/java/org/apache/fluss/server/kv/autoinc/IncIDGenerator.java
```diff
@@ -16,24 +16,15 @@
...
- * @since 0.9
- */
-@Internal
-public interface RowColumnVector extends ColumnVector {
+/** IncIDGenerator is used to generate auto increment column ID. */
+public interface IncIDGenerator {
 
     /**
-     * Return row value.
+     * Returns the next auto increment column ID.
```
platinumhamburg: Consider renaming this method, as "column ID" is misleading and conflicts with the actual ID of a Column object, which can cause confusion. Moreover, while UID is one use case for auto-increment columns, it does not encompass all possible scenarios, so naming it with "ID" is not advisable.

fluss-server/src/main/java/org/apache/fluss/server/kv/autoinc/BoundedSegmentSequenceGenerator.java
```diff
@@ -0,0 +1,108 @@
...
+    private static class AutoIncIdSegment {
+        private long current;
+        private final long end;
+
+        public AutoIncIdSegment(long start, long length) {
+            this.end = start + length;
+            this.current = start;
+        }
+
+        public long remaining() {
```
platinumhamburg: This part is a bit confusing to me—the outer check already handles negative values, so the Math.max here seems like an unnecessary overhead.

fluss-client/src/test/java/org/apache/fluss/client/table/FlussTableITCase.java
```diff
@@ -456,6 +457,47 @@ void testInvalidPrefixLookup() throws Exception {
                                 + "because the lookup columns [b, a] must contain all bucket keys [a, b] in order.");
     }
 
+    @Test
+    void testPutAutoIncColumnAndLookup() throws Exception {
```
platinumhamburg: To achieve more precise test assertions, I recommend explicitly creating a single-bucket table, writing multiple batches of data—some with duplicate primary keys and others with distinct primary keys—and then verifying that the auto-increment column behaves correctly.

fluss-common/src/main/java/org/apache/fluss/config/ConfigOptions.java
```diff
@@ -1424,6 +1424,13 @@ public class ConfigOptions {
                                     + "The `first_row` merge engine will keep the first row of the same primary key. "
                                     + "The `versioned` merge engine will keep the row with the largest version of the same primary key.");
 
+    public static final ConfigOption<Long> TABLE_AUTO_INC_BATCH_SIZE =
+            key("table.auto-inc.batch-size")
+                    .longType()
+                    .defaultValue(100000L)
+                    .withDescription(
+                            "The batch size of auto-increment IDs fetched from the distributed counter each time. This value determines the length of the ID segment cached locally. Default: 100000.");
```
platinumhamburg: New configuration options should be documented, including recommended values and a description of the behavioral impact associated with different settings.

fluss-server/src/main/java/org/apache/fluss/server/kv/KvTablet.java
```diff
@@ -468,8 +474,9 @@ private long applyInsert(
             PaddingRow latestSchemaRow,
             long logOffset)
             throws Exception {
-        walBuilder.append(ChangeType.INSERT, latestSchemaRow.replaceRow(currentValue.row));
-        kvPreWriteBuffer.put(key, currentValue.encodeValue(), logOffset);
+        BinaryValue newValue = autoIncProcessor.processAutoInc(currentValue);
```
platinumhamburg: I've been wondering about the relationship between auto-increment column handling and the Merge Engine. I think they should probably be integrated in the design to avoid redundant encoding of row data, which would otherwise lead to significant performance overhead.

platinumhamburg:
In addition to integration tests, I think we should also add several unit tests in KvTabletTest to cover behavior assertions after a TabletServer restart and after a single segment overflows.

fluss-server/src/main/java/org/apache/fluss/server/kv/KvTablet.java
xx789633: I agree. We had better handle the partial updater and auto increment column at the same time.

beryllw: Currently, auto-increment is only utilized for INSERT operations, which does not overlap with the functionalities of RowMerger and PartialUpdateRowMerger. I agree integrating auto-increment into RowMerger would result in a more streamlined design, and I will try to optimize this part.

beryllw:
> In addition to integration tests, I think we should also add several unit tests in KvTabletTest to cover behavior assertions after a TabletServer restart and after a single segment overflows.

We could create a new issue to address the failover problem, including the design and implementation of its corresponding recovery mechanisms.

fluss-server/src/main/java/org/apache/fluss/server/kv/autoinc/IncIDGenerator.java
```diff
@@ -16,24 +16,17 @@
...
-
-/**
- * Row {@link ColumnVector}.
- *
- * @since 0.9
- */
-@Internal
-public interface RowColumnVector extends ColumnVector {
+/** IncIDGenerator is used to generate auto increment column ID. */
+public interface IncIDGenerator {
```
platinumhamburg: Please rename to SequenceGenerator

fluss-server/src/main/java/org/apache/fluss/server/kv/autoinc/SegmentIncIDGenerator.java
```diff
@@ -0,0 +1,108 @@
...
+import org.apache.fluss.server.SequenceIDCounter;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import javax.annotation.concurrent.NotThreadSafe;
+
+/** Segment ID generator, fetch ID with a batch size. */
+@NotThreadSafe
+public class SegmentIncIDGenerator implements IncIDGenerator {
```
platinumhamburg: ditto

fluss-server/src/main/java/org/apache/fluss/server/kv/autoinc/AutoIncProcessor.java
```diff
@@ -0,0 +1,87 @@
...
+import org.apache.fluss.metadata.Schema;
+import org.apache.fluss.metadata.SchemaGetter;
+import org.apache.fluss.metadata.TablePath;
+import org.apache.fluss.record.BinaryValue;
+import org.apache.fluss.server.zk.ZkSequenceIDCounter;
+import org.apache.fluss.server.zk.ZooKeeperClient;
+import org.apache.fluss.server.zk.data.ZkData;
+
+/** AutoIncProcessor is used to process auto increment column. */
+public interface AutoIncProcessor {
```
platinumhamburg: Perhaps the AutoIncColumnProcessor could be renamed to AutoIncProcessor, and this static factory class could be renamed to AutoIncProcessors.

fluss-server/src/main/java/org/apache/fluss/server/kv/KvTablet.java
platinumhamburg: I agree with this perspective. One point worth noting is that when changelogImage == ChangelogImage.WAL, the applyInsert method might not be invoked. It would be helpful to clarify the intended design and expected behavior in this case.

Additionally, since PartialUpdater is currently used only by DefaultRowMerger and isn’t a general-purpose class, we might defer considering its consolidation for now.

fluss-server/src/main/java/org/apache/fluss/server/kv/autoinc/AutoIncColumnProcessor.java
```diff
@@ -0,0 +1,91 @@
...
+        for (int i = 0; i < fieldDataTypes.length; i++) {
+            flussFieldGetters[i] = InternalRow.createFieldGetter(fieldDataTypes[i], i);
+        }
+        this.rowEncoder = RowEncoder.create(kvFormat, fieldDataTypes);
+        this.schemaId = schemaId;
+    }
+
+    @Nullable
+    @Override
+    public BinaryValue processAutoInc(BinaryValue oldValue) {
```
platinumhamburg: It would be great to add test cases that validate the system’s behavior under schema evolution, particularly when handling rows from different schema versions where the column set and/or column order may differ.

platinumhamburg (changes requested):

fluss-server/src/main/java/org/apache/fluss/server/kv/KvTablet.java
beryllw: > One point worth noting is that when changelogImage == ChangelogImage.WAL, the applyInsert method might not be invoked. It would be helpful to clarify the intended design and expected behavior in this case.

Agree. When `changelogImage == ChangelogImage.WAL`, auto-Increment column aren't supported. We could clarify this in the docs.

platinumhamburg: Here, I believe we should support this. ChangelogImage.WAL is an optimization mode whose core use case is to skip generating UPDATE_BEFORE records when a user's table does not require UPDATE_BEFORE information. This significantly reduces the storage, replication (leader-follower), and read overhead associated with the changelog. Building upon this foundation, skipping the lookup of the old value to further reduce CPU and I/O overhead is merely an incremental optimization. Therefore, under the condition changelogImage == ChangelogImage.WAL, it's not that we cannot support the auto-increment columns; rather, we simply cannot benefit from this incremental optimization and would still need to look up the old value.

beryllw: The `ChangelogImage.WAL` mode is already supported. My concern is that when `changelogImage == ChangelogImage.WAL`, the ChangelogNormalize operator will be inserted, which will introduce additional computational overhead.

platinumhamburg: It's okay; this is Flink's behavior and can be optimized on the Flink side.

fluss-server/src/main/java/org/apache/fluss/server/kv/KvTablet.java
```diff
@@ -466,7 +466,9 @@ private long processUpsert(
         // partial update), we can skip fetching old value for better performance since it
         // always returns new value. In this case, both INSERT and UPDATE will produce
         // UPDATE_AFTER.
```
platinumhamburg: Comment above should be updated. The description of TABLE_CHANGELOG_IMAGE in both ConfigOptions.java and options.md should also be updated.

platinumhamburg (changes requested):

fluss-client/src/test/java/org/apache/fluss/client/table/FlussFailServerTableITCase.java
```diff
@@ -147,6 +153,64 @@ void testLogScan() throws Exception {
...
+
+            for (Object[] record : records) {
+                assertThatRow(lookupRow(lookuper, keyRow.replaceRow(row(record))))
+                        .withSchema(schema.getRowType())
+                        .isEqualTo(row(record));
+            }
+            upsertWriter.upsert(row("e", null, "batch2")).get();
+            // The auto-increment column should start from a new segment for now, and local cached
+            // IDs have been discarded.
+            assertThatRow(lookupRow(lookuper, keyRow.replaceRow(row("e", null, null))))
```
platinumhamburg: This is a good test addition; however, there's no need to create a new test class. The scenario can be simply incorporated into the existing tests in FlussTableITCase.java.

fluss-client/src/test/java/org/apache/fluss/client/table/FlussTableITCase.java
```diff
@@ -454,6 +455,177 @@ void testInvalidPrefixLookup() throws Exception {
...
+        createTable(tablePath, tableDescriptor, true);
+        Table autoIncTable = conn.getTable(tablePath);
+        UpsertWriter upsertWriter =
+                autoIncTable.newUpsert().partialUpdate("col1", "col3").createWriter();
+        Object[][] records = {
+            {"a", 0L, "batch1"},
+            {"b", 1L, "batch1"},
+            {"c", 2L, "batch1"},
+            {"d", 3L, "batch1"},
+            {"e", 4L, "batch1"}
```
platinumhamburg: The tests are generally good, but they can be further improved by abstracting the logic for writing partial updates to a specified schema and performing query assertions into a common utility method to enhance readability—this would significantly simplify the tests. Additionally, the test data should include records with duplicate primary keys to strengthen the effectiveness of behavioral validation.

fluss-client/src/test/java/org/apache/fluss/client/table/FlussTableITCase.java
```diff
@@ -454,6 +455,177 @@ void testInvalidPrefixLookup() throws Exception {
...
+        // schema change case4: read data with new schema.
+        for (Object[] record : recordsWithNewSchema) {
+            assertThatRow(lookupRow(newLookuper, keyRow.replaceRow(row(record))))
+                    .withSchema(newSchema.getRowType())
+                    .isEqualTo(row(record));
+        }
+    }
+
+    @Test
+    void testPutAutoIncColumnAndLookup() throws Exception {
```
platinumhamburg: This test doesn't appear to include any additional behavioral verification; it could potentially be merged with the previous test （testSingleBucketPutAutoIncColumnAndLookup）.

fluss-server/src/main/java/org/apache/fluss/server/kv/autoinc/AutoIncProcessor.java
```diff
@@ -0,0 +1,202 @@
...
+            autoIncUpdater = NO_OP_UPDATER;
+        }
+    }
+
+    // Supports removing or reordering columns; does NOT support adding an auto-increment column to
+    // an existing table.
+    public void configureSchema(int latestSchemaId) {
+        if (latestSchemaId != this.schemaId) {
+            Schema schema = schemaGetter.getSchema(latestSchemaId);
+            int[] autoIncColumnIds = schema.getAutoIncColumnIds();
```
platinumhamburg: From the schema implementation perspective, this is a high-overhead operation and should not be repeatedly computed unnecessarily; at the very least, some form of caching mechanism should be in place.

fluss-server/src/main/java/org/apache/fluss/server/kv/autoinc/AutoIncProcessor.java
```diff
@@ -0,0 +1,202 @@
...
+                    this.autoIncUpdater =
+                            new DefaultAutoIncUpdater(
+                                    kvFormat,
+                                    (short) latestSchemaId,
+                                    schemaGetter.getSchema(latestSchemaId),
+                                    autoIncColumnId,
+                                    sequenceGeneratorMap.get(autoIncColumnId));
+                } else {
+                    throw new IllegalStateException(
+                            "Not supported add auto increment column for a table.");
```
platinumhamburg: I believe intercepting here is unreasonable—the DDL has already succeeded, and the client has even attempted to write data. If this needs to be prohibited, it should be blocked on the DDL side.

fluss-server/src/main/java/org/apache/fluss/server/kv/autoinc/AutoIncProcessor.java
```diff
@@ -0,0 +1,202 @@
...
+                            sequenceGeneratorMap.get(autoIncColumnIds[0]));
+        } else {
+            autoIncUpdater = NO_OP_UPDATER;
+        }
+    }
+
+    // Supports removing or reordering columns; does NOT support adding an auto-increment column to
+    // an existing table.
+    public void configureSchema(int latestSchemaId) {
+        if (latestSchemaId != this.schemaId) {
```
platinumhamburg: The condition here should be inverted with an early return to avoid deep indentation of a large code block, which would otherwise degrade readability.

fluss-server/src/main/java/org/apache/fluss/server/kv/autoinc/AutoIncProcessor.java
```diff
@@ -0,0 +1,202 @@
...
+            int[] autoIncColumnIds = schema.getAutoIncColumnIds();
+            if (autoIncColumnIds.length > 1) {
+                throw new IllegalStateException(
+                        "Only support one auto increment column for a table, but got "
+                                + autoIncColumnIds.length);
+            } else if (autoIncColumnIds.length == 1) {
+                int autoIncColumnId = autoIncColumnIds[0];
+                if (sequenceGeneratorMap.containsKey(autoIncColumnId)) {
+                    this.autoIncUpdater =
+                            new DefaultAutoIncUpdater(
```
platinumhamburg: It seems that different versions of DefaultAutoIncUpdater can be managed using a cache.

fluss-server/src/main/java/org/apache/fluss/server/kv/autoinc/AutoIncProcessor.java
platinumhamburg: Alternatively, the validation/assertion here may be acceptable, but the error message is inappropriate—please verify whether the DDL-side interception is comprehensive.

fluss-server/src/main/java/org/apache/fluss/server/kv/autoinc/AutoIncProcessor.java
```diff
@@ -0,0 +1,202 @@
...
+                if (targetColumnIdx == i) {
+                    rowEncoder.encodeField(i, idGenerator.nextVal());
+                } else {
+                    // use the old row value
+                    if (oldValue == null) {
+                        rowEncoder.encodeField(i, null);
+                    } else {
+                        rowEncoder.encodeField(
+                                i, flussFieldGetters[i].getFieldOrNull(oldValue.row));
+                    }
```
platinumhamburg: Could oldValue here be null? I don't think that's possible—this doesn't appear to be a scenario that RowMerger would encounter.

platinumhamburg: Here, oldValue actually corresponds to newValue in RowMerger.

platinumhamburg (changes requested):
I've left some additional comments.

fluss-client/src/test/java/org/apache/fluss/client/table/FlussTableITCase.java
beryllw: We should also test partitioned and multi-bucket scenarios.

fluss-server/src/main/java/org/apache/fluss/server/kv/autoinc/AutoIncProcessor.java
```diff
@@ -79,60 +90,51 @@ public AutoIncProcessor(
...
-                    throw new IllegalStateException(
-                            "Not supported add auto increment column for a table.");
-                }
-            } else {
-                this.autoIncUpdater = NO_OP_UPDATER;
-                this.sequenceGeneratorMap.clear();
-            }
-            this.schemaId = latestSchemaId;
+        if (sequenceGeneratorMap.isEmpty()) {
+            currentAutoIncUpdater = NO_OP_UPDATER;
```
platinumhamburg: The assignment to currentAutoIncUpdater here will soon be overwritten by the next assignment.

platinumhamburg (changes requested):

fluss-server/src/main/java/org/apache/fluss/server/kv/autoinc/AutoIncProcessor.java
```diff
@@ -0,0 +1,118 @@
...
+    }
+
+    private AutoIncUpdater createAutoIncUpdater(int schemaId) {
+        Schema schema = schemaGetter.getSchema(schemaId);
+        int[] autoIncColumnIds = schema.getAutoIncColumnIds();
+        checkState(
+                autoIncColumnIds.length != sequenceGeneratorMap.size(),
+                String.format(
+                        "Auto-increment column count (%d) does not match sequence generator count (%d). Adding or dropping auto-increment columns is not supported.",
+                        autoIncColumnIds.length, sequenceGeneratorMap.size()));
```
platinumhamburg: checkState supports format strings without requiring String.format()

fluss-server/src/main/java/org/apache/fluss/server/kv/autoinc/AutoIncProcessor.java
```diff
@@ -0,0 +1,118 @@
...
+
+/** AutoIncProcessor is used to process auto increment column. */
+@NotThreadSafe
+public class AutoIncProcessor {
+    private static final AutoIncUpdater NO_OP_UPDATER = new AutoIncUpdater.NoOpUpdater();
+
+    private final SchemaGetter schemaGetter;
+    private final KvFormat kvFormat;
+    private final Cache<Integer, AutoIncUpdater> autoIncUpdaters;
+    private final Map<Integer, SequenceGenerator> sequenceGeneratorMap = new HashMap<>();
```
platinumhamburg: The code in all aspects already enforces that there can be only one auto-increment column, and it cannot be altered or newly added. Therefore, I believe the sequenceGeneratorMap is no longer necessary.

platinumhamburg: Additionally, I think the other parts of this PR are nearly mature.  The main thing left is to clean up AutoIncProcessor. After several refactorings, it still has some confusing or outdated parts—for example, is its name still right? Do its constructor and method names still make sense? In reality, its job has basically become that of an AutoIncUpdaterCache.

beryllw: Thanks a lot for the insightful suggestion during the review!

fluss-server/src/main/java/org/apache/fluss/server/kv/autoinc/AutoIncManager.java
```diff
@@ -0,0 +1,116 @@
...
+import java.time.Duration;
+
+import static org.apache.fluss.utils.Preconditions.checkState;
+
+/**
+ * Manages auto-increment logic for tables, providing schema-specific updaters that handle
+ * auto-increment column assignment during row writes.
+ */
+@NotThreadSafe
+public class AutoIncManager {
```
wuchong: Rename to `AutoIncrementManager` as the class name is not too long, having the complete class name will help developers to understand it, as `inc` has many different meanings.

fluss-server/src/main/java/org/apache/fluss/server/kv/autoinc/AutoIncUpdater.java
```diff
@@ -0,0 +1,46 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.fluss.server.kv.autoinc;
+
+import org.apache.fluss.record.BinaryValue;
+
+/** A updater to auto increment column . */
+public interface AutoIncUpdater {
```
wuchong: ditto. rename to `AutoIncrementUpdater`, and also the sub-classes.

fluss-server/src/main/java/org/apache/fluss/server/kv/autoinc/AutoIncUpdater.java
```diff
@@ -0,0 +1,46 @@
...
+     * Updates the auto-increment column in the given row by replacing its value with a new sequence
+     * number.
+     *
+     * <p>This method may return a new {@link BinaryValue} instance or the same instance if no
+     * update is needed (e.g., in a no-op implementation).
+     *
+     * @param rowValue the input row in binary form, must not be {@code null}
+     * @return a {@link BinaryValue} representing the updated row; never {@code null}
+     */
+    BinaryValue updateAutoInc(BinaryValue rowValue);
```
wuchong: nit: `updateAutoIncColumns` to be more specific.

fluss-common/src/main/java/org/apache/fluss/config/ConfigOptions.java
```diff
@@ -1456,7 +1456,18 @@ public class ConfigOptions {
                                     + "The auto increment column can only be used in primary-key table."
                                     + "With an auto increment column in the table, whenever a new row is inserted into the table, the new row will be assigned with the next available value from the auto-increment sequence."
                                     + "The auto increment column can only be used in primary-key table. The data type of the auto increment column must be INT or BIGINT."
-                                    + "Currently a table can have only one auto-increment column.");
+                                    + "Currently a table can have only one auto-increment column."
+                                    + "Adding an auto increment column to an existing table is not supported.");
+
+    public static final ConfigOption<Long> TABLE_AUTO_INC_BATCH_SIZE =
+            key("table.auto-inc.batch-size")
```
wuchong: Suggest renaming to `table.auto-increment.cache-size`.

- Prefer full, descriptive names when the config key isn’t overly long, this reduces ambiguity and improves readability.
- Both FIP and StarRocks (`auto_increment_cache_size`) use “cache size” instead of “batch size.” This is more accurate, as the setting controls the size of the pre-allocated ID cache, not the batch in which IDs are consumed or flushed.

fluss-client/src/test/java/org/apache/fluss/client/table/FlussTableITCase.java
```diff
@@ -454,6 +455,181 @@ void testInvalidPrefixLookup() throws Exception {
...
+            {"c", null, "batch1"},
+            {"d", null, "batch1"},
+            {"e", null, "batch1"},
+            {"d", null, "batch2"},
+            {"e", null, "batch2"}
+        };
+        partialUpdateRecords(new String[] {"col1", "col3"}, records, autoIncTable);
+
+        Object[][] expectedRecords = {
+            {"a", 0L, "batch1"},
```
wuchong: Since `AUTO_INCREMENT` columns in StarRocks, MySQL, Postgres, they all start value from `1`, rather than `0`. I suggest also following them to start from `1`. We can introduce `table.auto-increment.start-value` in the future to allow users to specifc a different starting value, like zero. 

https://docs.starrocks.io/docs/sql-reference/sql-statements/table_bucket_part_index/auto_increment/#introduction

fluss-server/src/main/java/org/apache/fluss/server/kv/KvManager.java
```diff
@@ -248,6 +249,10 @@ public KvTablet getOrCreateKv(
 
                     File tabletDir = getOrCreateTabletDir(tablePath, tableBucket);
                     RowMerger merger = RowMerger.create(tableConfig, kvFormat, schemaGetter);
+                    AutoIncManager autoIncManager =
+                            new AutoIncManager(
+                                    schemaGetter, tablePath.getTablePath(), conf, zkClient);
```
wuchong: The `conf` is a cluster configuration, not the table config. It's better to pass in the `TableConfig` to avoid mistakes.

fluss-server/src/main/java/org/apache/fluss/server/kv/KvTablet.java
```diff
@@ -450,22 +462,30 @@ private long processUpsert(
...
+                && currentMerger instanceof DefaultRowMerger) {
             return applyUpdate(key, null, currentValue, walBuilder, latestSchemaRow, logOffset);
         }
 
         byte[] oldValueBytes = getFromBufferOrKv(key);
         if (oldValueBytes == null) {
-            return applyInsert(key, currentValue, walBuilder, latestSchemaRow, logOffset);
+            return applyInsert(
+                    key,
+                    currentAutoIncUpdater.updateAutoInc(currentValue),
```
wuchong: I suggest making the updates within the `applyInsert` method so that the logic can be reused wherever `applyInsert` is called.

fluss-server/src/main/java/org/apache/fluss/server/kv/autoinc/PerSchemaAutoIncUpdater.java
```diff
@@ -0,0 +1,92 @@
...
+                            autoIncColumnId, schema.getColumnIds()));
+        }
+        this.rowEncoder = RowEncoder.create(kvFormat, fieldDataTypes);
+        this.fieldDataTypes = fieldDataTypes;
+        this.flussFieldGetters = flussFieldGetters;
+    }
+
+    public BinaryValue updateAutoInc(BinaryValue rowValue) {
+        rowEncoder.startNewRow();
+        for (int i = 0; i < fieldDataTypes.length; i++) {
```
wuchong: nit: just have a `fieldLength` member variable is enough instead of `fieldDataTypes`

fluss-server/src/main/java/org/apache/fluss/server/kv/autoinc/PerSchemaAutoIncUpdater.java
```diff
@@ -0,0 +1,92 @@
...
+            if (targetColumnIdx == i) {
+                rowEncoder.encodeField(i, idGenerator.nextVal());
```
wuchong: We should check int overflow here and throw a `ApiException`. We can setting an upper bounds in `SegmentSequenceGenerator` according the data types. And add test in `SegmentSequenceGeneratorTest`. 


The exception can be `SequenceOverflowException` or `AutoIncrementOverflowException` with messages like 
```
reached maximum value of sequence "<field_name>" (2147483647)
```

fluss-common/src/main/java/org/apache/fluss/metadata/Schema.java
```diff
@@ -128,6 +128,15 @@ public int[] getPrimaryKeyIndexes() {
                 .orElseGet(() -> new int[0]);
     }
 
+    /** Returns the auto-increment columnIds, if any, otherwise returns an empty array. */
+    public int[] getAutoIncColumnIds() {
+        Set<String> autoIncColSet = new HashSet<>(getAutoIncrementColumnNames());
+        return getColumns().stream()
+                .filter(column -> autoIncColSet.contains(column.getName()))
+                .mapToInt(Column::getColumnId)
+                .toArray();
```
wuchong:
```suggestion
        if (autoIncrementColumnNames.isEmpty()) {
            return new int[0];
        } else {
            return getColumns().stream()
                    .filter(column -> autoIncrementColumnNames.contains(column.getName()))
                    .mapToInt(Column::getColumnId)
                    .toArray();
        }
```

Improve the performance

fluss-common/src/main/java/org/apache/fluss/metadata/Schema.java
```diff
@@ -128,6 +128,15 @@ public int[] getPrimaryKeyIndexes() {
                 .orElseGet(() -> new int[0]);
     }
 
+    /** Returns the auto-increment columnIds, if any, otherwise returns an empty array. */
```
wuchong, line 154: We should also update `org.apache.fluss.metadata.Schema.Builder#fromSchema` to consider the autoIncrementFields. And add tests in `TableSchemaTest`

fluss-common/src/main/java/org/apache/fluss/metadata/Schema.java
```diff
@@ -696,9 +710,7 @@ private static List<Column> normalizeColumns(
             }
 
             // primary key and auto increment column should not nullable
-            if ((pkSet.contains(column.getName())
-                            || autoIncrementColumnNames.contains(column.getName()))
```
wuchong, line 752: Why remove this? Auto increment columns must be not nullable.

fluss-server/src/test/java/org/apache/fluss/server/kv/autoinc/SegmentIncIDGeneratorTest.java
```diff
@@ -0,0 +1,149 @@
...
+                                                new TablePath("test_db", "table1"),
+                                                COLUMN_IDX,
+                                                COLUMN_NAME + "_table1",
+                                                new TestingSnapshotIDCounter(snapshotIdGenerator),
+                                                configuration);
+                                for (int j = 0; j < 130; j++) {
+                                    linkedDeque.add(generator.nextVal());
+                                }
+                            })
+                    .start();
```
wuchong: We should `.join()` the threads before assersion? Besides, would be better to also verify the distinct values size should be `130 * 20`

fluss-server/src/main/java/org/apache/fluss/server/kv/autoinc/SegmentSequenceGenerator.java
```diff
@@ -0,0 +1,108 @@
...
+
+    private final long batchSize;
+
+    public SegmentSequenceGenerator(
+            TablePath tablePath,
+            int columnIdx,
+            String columnName,
+            SequenceIDCounter sequenceIDCounter,
+            Configuration properties) {
+        batchSize = properties.getLong(ConfigOptions.TABLE_AUTO_INC_BATCH_SIZE);
```
wuchong:
```suggestion
        this.batchSize = properties.getLong(ConfigOptions.TABLE_AUTO_INC_BATCH_SIZE);
```

fluss-server/src/main/java/org/apache/fluss/server/kv/autoinc/SegmentSequenceGenerator.java
```diff
@@ -0,0 +1,108 @@
...
+                    start,
+                    start + batchSize,
+                    tablePath,
+                    columnIdx,
+                    columnName);
+            segment = new AutoIncIdSegment(start, batchSize);
+        } catch (Exception e) {
+            throw new FlussRuntimeException(
+                    String.format(
+                            "Failed to fetch auto-increment values, table_path=%s, column_idx=%d, column_name=%s.",
```
wuchong: I don't see the needs to log the `column_idx`, seems it's verbose here, maybe we can omit it.

wuchong (commented):
Thanks @beryllw for the great contribution. 

I left some comments. 

Besides, could you also 

1. update the option `table.auto-increment.fields` to `auto-increment.fields` (aligns `bucket.keys`) and move it to `FlinkConnectorOptions` as it is a connector option, as it is not recognized by Fluss server. Naming it `table.*` prefix is a mis-usage. 
2. We should also add it into `FlinkConnectorOptions#ALTER_DISALLOW_OPTIONS`, and add test in `FlinkCatalogITCase#testAlterTable` to verify this config is not allowed to alter and a proper exception and message should be thrown. 
3. Add limitation about "Adding the AUTO INCREMENT columns by using ALTER TABLE is not supported" in https://fluss.apache.org/docs/next/engine-flink/ddl/#add-columns

fluss-common/src/main/java/org/apache/fluss/metadata/Schema.java
beryllw: Tables with auto-increment columns need to be written using partial updates, and partial updates require the auto-increment column to be nullable. 
Alternatively, could we set a default value for the auto-increment column instead?

fluss-common/src/test/java/org/apache/fluss/metadata/TableSchemaTest.java
```diff
@@ -221,6 +222,20 @@ void testSchemaFromSchemaPreservesAggFunction() {
...
+        Schema original =
+                Schema.newBuilder()
+                        .column("id", DataTypes.INT())
+                        .column("value", DataTypes.BIGINT())
+                        .primaryKey("id")
+                        .enableAutoIncrement("value")
+                        .build();
+
+        Schema copied = Schema.newBuilder().fromSchema(original).build();
+        assertThat(copied.getAutoIncrementColumnNames()).isEqualTo(Arrays.asList("value"));
```
wuchong: We can assert the whole schema to make sure all fields are copied.

fluss-server/src/main/java/org/apache/fluss/server/kv/autoinc/BoundedSegmentSequenceGenerator.java
```diff
@@ -0,0 +1,119 @@
...
+        try {
+            long start = sequenceIDCounter.getAndAdd(cacheSize);
+            if (start >= maxAllowedValue) {
+                throw new SequenceOverflowException(
+                        String.format(
+                                "Reached maximum value of sequence \"<%s>\" (%d).",
+                                columnName, maxAllowedValue));
+            }
+
+            long actualEnd = Math.min(start + cacheSize, maxAllowedValue - 1);
```
wuchong: the value of `maxAllowedValue` should be allowed to be generated.

fluss-server/src/main/java/org/apache/fluss/server/kv/autoinc/PerSchemaAutoIncrementUpdater.java
```diff
@@ -0,0 +1,112 @@
...
+                            targetColumnName));
+        }
+        return value;
+    }
+
+    public BinaryValue updateAutoIncrementColumns(BinaryValue rowValue) {
+        rowEncoder.startNewRow();
+        for (int i = 0; i < fieldLength; i++) {
+            if (targetColumnIdx == i) {
+                rowEncoder.encodeField(i, idSupplier.getAsLong());
```
wuchong: there will be cast exception if the column is `INT` type.

fluss-server/src/main/java/org/apache/fluss/server/kv/autoinc/PerSchemaAutoIncrementUpdater.java
```diff
@@ -0,0 +1,112 @@
...
+        this.flussFieldGetters = flussFieldGetters;
+    }
+
+    private long checkedNextInt(long value) {
+        if (value > Integer.MAX_VALUE) {
+            throw new SequenceOverflowException(
+                    String.format(
+                            "Reached maximum value of sequence \"<%s>\" (2147483647).",
+                            targetColumnName));
+        }
```
wuchong: We have check this in `BoundedSegmentSequenceGenerator`, so we don't need to check again here.

beryllw:
@wuchong , thanks for the review and the improvements!

PR #2163, opened 2025-12-12
Prevent Endangered ISR from Losing Leader Permanently

### LINKED ISSUE #2162: Prevent Endangered ISR from Losing Leader Permanently
### Search before asking

### Fluss version

0.8.0 (latest release)

### Please describe the bug 🐞

When the ISR (In-Sync Replicas) shrinks to a single replica, the current implementation will set the leader to NO_LEADER if that replica goes offline. This creates an issue: since ISR expansion requires a leader to process expansion requests, setting the leader—which is the sole member of the ISR—to NO_LEADER is meaningless. As the only replica in the ISR, it must be started in Leader Replica mode to drive the bucket state back to normal state.

### Solution

When ISR contains only one replica (endangered ISR), preserve the current leader even if that replica goes offline temporarily. The key changes:

1. Detect endangered ISR state: `isEndangeredIsr = (leaderAndIsr.isr().size() == 1)`
2. Keep the last replica in ISR to maintain leader election capability
3. Only set leader to `NO_LEADER` when ISR is not endangered
4. Allow the original replica to resume as leader and process ISR expansion requests when it recovers

This ensures the bucket can recover to normal state when replicas come back online, as the preserved leader can handle ISR expansion requests.

platinumhamburg:
### Purpose

Linked issue: close #2162

### DIFF
--- a/fluss-server/src/main/java/org/apache/fluss/server/coordinator/statemachine/ReplicaStateMachine.java
+++ b/fluss-server/src/main/java/org/apache/fluss/server/coordinator/statemachine/ReplicaStateMachine.java
@@ -436,21 +436,32 @@ public class ReplicaStateMachine {
                 // isr doesn't contain the replica, skip
                 continue;
             }
-            int newLeader =
-                    replicaId == leaderAndIsr.leader()
-                            ?
-                            // the leader become offline, set it to no leader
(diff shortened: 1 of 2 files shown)


platinumhamburg:
I will close this issue and PR.

PR #2165, opened 2025-12-12
[common] Fix IndexOutOfBoundsException in ArrowArrayWriter when element count exceeds INITIAL_CAPACITY

### LINKED ISSUE #2164: [Bug] IndexOutOfBoundsException when writing rows with array columns to KV table
### Search before asking

### Fluss version

main (development)

### Minimal reproduce step

Write data containing array columns to a Fluss KV table where:
1. The number of rows is less than 1024 (INITIAL_CAPACITY)
2. The total number of array elements across all rows exceeds 1024

For example: 250 rows with arrays containing 10 integers each (total 2500 elements).

### What do you expect to see?

Data should be written successfully without errors.

### What did you see instead?

```
java.lang.IndexOutOfBoundsException: index: 248, length: 1 (expected: range(0, 248))
    at org.apache.fluss.shaded.arrow.org.apache.arrow.memory.ArrowBuf.checkIndexD(ArrowBuf.java:319)
    at org.apache.fluss.shaded.arrow.org.apache.arrow.memory.ArrowBuf.chk(ArrowBuf.java:306)
    at org.apache.fluss.shaded.arrow.org.apache.arrow.memory.ArrowBuf.getByte(ArrowBuf.java:508)
    at org.apache.fluss.shaded.arrow.org.apache.arrow.vector.BitVectorHelper.setBit(BitVectorHelper.java:82)
    at org.apache.fluss.shaded.arrow.org.apache.arrow.vector.IntVector.set(IntVector.java:160)
    at org.apache.fluss.row.arrow.writers.ArrowIntWriter.doWrite(ArrowIntWriter.java:38)
    at org.apache.fluss.row.arrow.writers.ArrowFieldWriter.write(ArrowFieldWriter.java:59)
    at org.apache.fluss.row.arrow.writers.ArrowArrayWriter.doWrite(ArrowArrayWriter.java:44)
    ...
```

### Root Cause Analysis

In `ArrowWriter.writeRow()`, the `handleSafe` flag is determined by comparing `recordsCount` (number of rows) against `INITIAL_CAPACITY` (1024):

```java
boolean handleSafe = recordsCount >= INITIAL_CAPACITY;
```

When handleSafe = false, Arrow writers use vector.set() which doesn't auto-grow the underlying buffer. When handleSafe = true, they use vector.setSafe() which auto-grows as needed.

The bug is in ArrowArrayWriter: it passes the parent's handleSafe flag (based on row count) to the element writer. However, array element indices are based on the cumulative element count across all arrays, not the row count. This means:

250 rows → handleSafe = false (since 250 < 1024)
But with 10 elements per array, element indices can reach 2500
The element vector was only initialized for 1024 elements
Arrow throws IndexOutOfBoundsException when writing element 1024+

### Solution

_No response_

binary-signal:
## Purpose

Linked issue: close #2164 

Fix `IndexOutOfBoundsException` when writing rows with array columns where the total number of array elements exceeds `INITIAL_CAPACITY` (1024) while the row count stays below it.

### Brief change log

In `ArrowWriter.writeRow()`, the `handleSafe` flag is determined by comparing row count against `INITIAL_CAPACITY`:
```java
boolean handleSafe = recordsCount >= INITIAL_CAPACITY;
```

When `handleSafe = false`, Arrow writers use `vector.set()` which doesn't auto-grow the buffer. The bug is in `ArrowArrayWriter.doWrite()` which passes the parent's `handleSafe` flag to the element writer. However, array element indices grow based on cumulative element count, not row count.

Example: 250 rows with 10-element arrays → row count (250) < 1024 so `handleSafe = false`, but total elements (2500) exceeds the vector's initial capacity, causing `IndexOutOfBoundsException`.

**Fix:**
Always use safe writes (`handleSafe = true`) for array element writers in `ArrowArrayWriter.doWrite()`, since element indices can exceed `INITIAL_CAPACITY` independently of row count.

```java
// Before
elementWriter.write(fieldIndex, array, arrIndex, handleSafe);

// After
elementWriter.write(fieldIndex, array, arrIndex, true);
```

### Tests

- Added `ArrowReaderWriterTest#testArrayWriterWithManyElements`: writes 200 rows with 10-element arrays (2000 total elements), verifying serialization succeeds and data can be read back correctly.

### API and Format

### Documentation

No documentation changes needed. This is a bug fix.
```

### DIFF
--- a/fluss-common/src/main/java/org/apache/fluss/row/arrow/writers/ArrowArrayWriter.java
+++ b/fluss-common/src/main/java/org/apache/fluss/row/arrow/writers/ArrowArrayWriter.java
@@ -20,6 +20,7 @@ package org.apache.fluss.row.arrow.writers;
 
 import org.apache.fluss.row.DataGetters;
 import org.apache.fluss.row.InternalArray;
+import org.apache.fluss.row.arrow.ArrowWriter;
 import org.apache.fluss.shaded.arrow.org.apache.arrow.vector.FieldVector;
 import org.apache.fluss.shaded.arrow.org.apache.arrow.vector.complex.ListVector;
 
@@ -41,11 +42,12 @@ public class ArrowArrayWriter extends ArrowFieldWriter {
         listVector.startNewValue(rowIndex);
         for (int arrIndex = 0; arrIndex < array.size(); arrIndex++) {
             int fieldIndex = offset + arrIndex;
-            // Always use safe writes for array elements because the element index (offset +
-            // arrIndex) can exceed INITIAL_CAPACITY even when the row count doesn't. The parent's
-            // handleSafe is based on row count, but array element indices grow based on the total
-            // number of elements across all arrays, which can be much larger.
-            elementWriter.write(fieldIndex, array, arrIndex, true);
+            // Use element-based index to determine handleSafe, not parent row count.
+            // This fixes issue #2164: when row count < INITIAL_CAPACITY but total
+            // array elements > INITIAL_CAPACITY, we need to use safe mode for elements
+            // beyond the initial capacity.
+            boolean elementHandleSafe = fieldIndex >= ArrowWriter.INITIAL_CAPACITY;
+            elementWriter.write(fieldIndex, array, arrIndex, elementHandleSafe);
         }
         offset += array.size();
         listVector.endValue(rowIndex, array.size());
--- a/fluss-common/src/main/java/org/apache/fluss/row/arrow/ArrowWriter.java
+++ b/fluss-common/src/main/java/org/apache/fluss/row/arrow/ArrowWriter.java
@@ -56,8 +56,11 @@ public class ArrowWriter implements AutoCloseable {
     /**
      * The initial capacity of the vectors which are used to store the rows. The capacity will be
      * expanded automatically if the rows exceed the initial capacity.
+     *
+     * <p>Public for use by nested writers (e.g., ArrowArrayWriter) to determine when to use safe
+     * write mode based on element indices.
      */
-    private static final int INITIAL_CAPACITY = 1024;
+    public static final int INITIAL_CAPACITY = 1024;
 
     /**
      * The buffer usage ratio which is used to determine whether the writer is full. The writer is

### CODE REVIEW

File: fluss-common/src/main/java/org/apache/fluss/row/arrow/writers/ArrowArrayWriter.java [hunk 1]
```diff
@@ -41,7 +41,11 @@ public void doWrite(int rowIndex, DataGetters row, int ordinal, boolean handleSa
         listVector.startNewValue(rowIndex);
         for (int arrIndex = 0; arrIndex < array.size(); arrIndex++) {
             int fieldIndex = offset + arrIndex;
-            elementWriter.write(fieldIndex, array, arrIndex, handleSafe);
+            // Always use safe writes for array elements because the element index (offset +
```
vamossagar12: We should mention in the comments on the class that the handleSafe field is ignored when writing.

vamossagar12 (approved):
LGTM

fluss-common/src/main/java/org/apache/fluss/row/arrow/writers/ArrowArrayWriter.java
```diff
@@ -41,7 +41,11 @@ public void doWrite(int rowIndex, DataGetters row, int ordinal, boolean handleSa
...
+            // arrIndex) can exceed INITIAL_CAPACITY even when the row count doesn't. The parent's
+            // handleSafe is based on row count, but array element indices grow based on the total
+            // number of elements across all arrays, which can be much larger.
+            elementWriter.write(fieldIndex, array, arrIndex, true);
```
platinumhamburg:
```suggestion
            boolean elementHandleSafe = fieldIndex >= ArrowWriter.INITIAL_CAPACITY;
            elementWriter.write(fieldIndex, array, arrIndex, elementHandleSafe);
```

platinumhamburg: Perhaps we could apply a clear conditional judgment here.

platinumhamburg:
+1，just a minor suggestion

wuchong (approved):
I appended a commit from @platinumhamburg #2287 to improve the fix.

platinumhamburg (approved):

PR #2178, opened 2025-12-15
Support shared RocksDB rate limiter in Fluss

### LINKED ISSUE #2177: [KV] Add RocksDB rate limiter support for TabletServer
### Search before asking

### Motivation

RocksDB's flush and compaction operations can generate significant I/O pressure, potentially impacting:
- **System stability**: Uncontrolled I/O may saturate disk bandwidth, affecting other operations
- **Query latency**: High I/O contention can degrade read/write performance
- **Resource predictability**: Difficult to manage multi-tenant TabletServer resource allocation

RocksDB provides a built-in rate limiter mechanism to control flush and compaction write rates, but Fluss currently doesn't expose this capability to users.

### 1. Shared Rate Limiter Architecture
- **Server-level rate limiting**: A single rate limiter shared across all RocksDB instances on each TabletServer
- **Resource isolation**: Prevents any single table from consuming excessive I/O bandwidth
- **Cost efficiency**: Reduces memory overhead compared to per-instance rate limiters

### 2. Configuration Options
**New configuration: `kv.shared-rate-limiter-bytes-per-sec`**
- **Type**: MemorySize
- **Default**: `0b` (disabled)
- **Scope**: Cluster-level, dynamically updatable
- **Example values**: `100MB`, `500MB`, `1GB`

Static configuration in `conf/server.yaml`:
kv.shared-rate-limiter-bytes-per-sec: 200mb### 3. Dynamic Configuration Support
Rate limiter can be updated at runtime without server restart:
- **Persistence**: Changes persisted to ZooKeeper
- **Cluster-wide**: Applies to all TabletServers
- **Zero-downtime**: Thread-safe updates via RocksDB API

### 3. Flink Stored Procedures
SQL-based management interface for operators:

**Set rate limiter:**
-- Named argument (Flink 1.19+)
CALL fluss_catalog.sys.set_shared_rocksdb_rate_limiter(rate_limit => '200MB');

-- Indexed argument
CALL fluss_catalog.sys.set_shared_rocksdb_rate_limiter('500MB');

-- Disable rate limiter
CALL fluss_catalog.sys.set_shared_rocksdb_rate_limiter('0MB');**Query current setting:**
CALL fluss_catalog.sys.get_shared_rocksdb_rate_limiter();

### Anything else?

_No response_

### Willingness to contribute

platinumhamburg:
### Purpose

Linked issue: close #2177

### DIFF
--- a/website/docs/engine-flink/ddl.md
+++ b/website/docs/engine-flink/ddl.md
@@ -27,7 +27,7 @@ The following properties can be set if using the Fluss catalog:
 | bootstrap.servers              | required | (none)    | Comma separated list of Fluss servers.                                                                                                                                                                                 |
 | default-database               | optional | fluss     | The default database to use when switching to this catalog.                                                                                                                                                            |
 | client.security.protocol       | optional | PLAINTEXT | The security protocol used to communicate with brokers. Currently, only `PLAINTEXT` and `SASL` are supported, the configuration value is case insensitive.                                                             |
-| `client.security.{protocol}.*` | optional | (none)    | Client-side configuration properties for a specific authentication protocol. E.g., client.security.sasl.jaas.config. More Details in [authentication](../../security/authentication.md)                                   |
+| `client.security.{protocol}.*` | optional | (none)    | Client-side configuration properties for a specific authentication protocol. E.g., client.security.sasl.jaas.config. More Details in [authentication](../security/authentication.md)                                   |
 | `{lake-format}.*`              | optional | (none)    | Extra properties to be passed to the lake catalog. This is useful for configuring sensitive settings, such as the username and password required for lake catalog authentication. E.g., `paimon.jdbc.password = pass`. |
 
 The following statements assume that the current catalog has been switched to the Fluss catalog using the `USE CATALOG <catalog_name>` statement.
@@ -125,7 +125,7 @@ CREATE TABLE my_part_log_table (
 Fluss partitioned table supports dynamic partition creation, which means you can write data into a partition without pre-creating it.
 You can use the `INSERT INTO` statement to write data into a partitioned table, and Fluss will automatically create the partition if it does not exist.
 See the [Dynamic Partitioning](table-design/data-distribution/partitioning.md#dynamic-partitioning) for more details.
-But you can still use the [Add Partition](engine-flink/ddl/index.md#add-partition) statement to manually add partitions if needed.
+But you can still use the [Add Partition](engine-flink/ddl.md#add-partition) statement to manually add partitions if needed.
 :::
 
 #### Multi-Fields Partitioned Table
--- a/website/docs/maintenance/operations/updating-configs.md
+++ b/website/docs/maintenance/operations/updating-configs.md
@@ -18,7 +18,7 @@ Currently, the supported dynamically updatable server configurations include:
 - `kv.rocksdb.shared-rate-limiter.bytes-per-sec`: Control RocksDB flush and compaction write rate shared across all RocksDB instances on the TabletServer. The rate limiter is always enabled. Set to a lower value (e.g., 100MB) to limit the rate, or a very high value to effectively disable rate limiting.
 
 
-You can update the configuration of a cluster with [Java client](#using-java-client) or [Flink Stored Procedures](../../engine-flink/ddl/procedures.md#cluster-configuration-procedures).
+You can update the configuration of a cluster with [Java client](#using-java-client) or [Flink Procedures](../../engine-flink/procedures.md#cluster-configuration-procedures).
 
 ### Using Java Client
 
@@ -48,11 +48,11 @@ The `AlterConfig` class contains three properties:
 
 ### Using Flink Stored Procedures
 
-For certain configurations, Fluss provides convenient Flink stored procedures that can be called directly from Flink SQL. See [Procedures](engine-flink/ddl/procedures.md#cluster-configuration-procedures) for detailed documentation on using `get_cluster_config` and `set_cluster_config` procedures.
+For certain configurations, Fluss provides convenient Flink stored procedures that can be called directly from Flink SQL. See [Procedures](engine-flink/procedures.md#cluster-configuration-procedures) for detailed documentation on using `get_cluster_config` and `set_cluster_config` procedures.
 
 ## Updating Table Configs
 
-The connector options on a table including [Storage Options](engine-flink/options.md#storage-options) can be updated dynamically by [ALTER TABLE ... SET](engine-flink/ddl/index.md#alter-table) statement. See the example below:
+The connector options on a table including [Storage Options](engine-flink/options.md#storage-options) can be updated dynamically by [ALTER TABLE ... SET](engine-flink/ddl.md#alter-table) statement. See the example below:
 
 ```sql
 -- Enable lakehouse storage for the given table
--- a/website/docs/engine-flink/datastream.mdx
+++ b/website/docs/engine-flink/datastream.mdx
@@ -1,6 +1,6 @@
 ---
 title: "DataStream API"
-sidebar_position: 7
+sidebar_position: 8
 ---
 
 # DataStream API
--- a/website/docs/engine-flink/delta-joins.md
+++ b/website/docs/engine-flink/delta-joins.md
@@ -1,7 +1,7 @@
 ---
 sidebar_label: Delta Joins
 title: Flink Delta Joins
-sidebar_position: 6
+sidebar_position: 7
 ---
 
 # The Delta Join
--- a/website/docs/engine-flink/getting-started.md
+++ b/website/docs/engine-flink/getting-started.md
@@ -22,19 +22,19 @@ For Flink's Table API, Fluss supports the following features:
 
 | Feature Support                                   | Flink | Notes                                  |
 |---------------------------------------------------|-------|----------------------------------------|
-| [SQL Create Catalog](ddl/index.md#create-catalog)       | ✔️    |                                        |
-| [SQL Create Database](ddl/index.md#create-database)     | ✔️    |                                        |
-| [SQL Drop Database](ddl/index.md#drop-database)         | ✔️    |                                        |
-| [SQL Create Table](ddl/index.md#create-table)           | ✔️    |                                        |
-| [SQL Create Table Like](ddl/index.md#create-table-like) | ✔️    |                                        |
-| [SQL Drop Table](ddl/index.md#drop-table)               | ✔️    |                                        |
-| [SQL Create Materialized Table](ddl/index.md#materialized-table) | ✔️    | Continuous refresh mode only |
-| [SQL Alter Materialized Table](ddl/index.md#alter-materialized-table) | ✔️    | Suspend/Resume support |
-| [SQL Drop Materialized Table](ddl/index.md#drop-materialized-table) | ✔️    |                                        |
-| [SQL Show Partitions](ddl/index.md#show-partitions)     | ✔️    |                                        |
-| [SQL Add Partition](ddl/index.md#add-partition)         | ✔️    |                                        |
-| [SQL Drop Partition](ddl/index.md#drop-partition)       | ✔️    |                                        |
-| [Procedures](ddl/index.md#procedures)                   | ✔️    | ACL management and cluster configuration |
+| [SQL Create Catalog](ddl.md#create-catalog)       | ✔️    |                                        |
+| [SQL Create Database](ddl.md#create-database)     | ✔️    |                                        |
+| [SQL Drop Database](ddl.md#drop-database)         | ✔️    |                                        |
+| [SQL Create Table](ddl.md#create-table)           | ✔️    |                                        |
+| [SQL Create Table Like](ddl.md#create-table-like) | ✔️    |                                        |
+| [SQL Drop Table](ddl.md#drop-table)               | ✔️    |                                        |
+| [SQL Create Materialized Table](ddl.md#materialized-table) | ✔️    | Continuous refresh mode only |
+| [SQL Alter Materialized Table](ddl.md#alter-materialized-table) | ✔️    | Suspend/Resume support |
+| [SQL Drop Materialized Table](ddl.md#drop-materialized-table) | ✔️    |                                        |
+| [SQL Show Partitions](ddl.md#show-partitions)     | ✔️    |                                        |
+| [SQL Add Partition](ddl.md#add-partition)         | ✔️    |                                        |
+| [SQL Drop Partition](ddl.md#drop-partition)       | ✔️    |                                        |
+| [Procedures](ddl.md#procedures)                   | ✔️    | ACL management and cluster configuration |
 | [SQL Select](reads.md)                            | ✔️    | Support both streaming and batch mode. |
 | [SQL Limit](reads.md#limit-read)                  | ✔️    | Only for Log Table                     |
 | [SQL Insert Into](writes.md)                      | ✔️    | Support both streaming and batch mode. |
--- a/website/docs/engine-flink/lookups.md
+++ b/website/docs/engine-flink/lookups.md
@@ -1,7 +1,7 @@
 ---
 sidebar_label: Lookups
 title: Flink Lookup Joins
-sidebar_position: 5
+sidebar_position: 6
 ---
 
 # Flink Lookup Joins
--- a/website/docs/engine-flink/options.md
+++ b/website/docs/engine-flink/options.md
@@ -1,6 +1,6 @@
 ---
 title: Connector Options
-sidebar_position: 8
+sidebar_position: 9
 ---
 
 # Connector Options
@@ -57,7 +57,7 @@ Using `ALTER TABLE ... SET` statement to modify the table options. For example, 
 ALTER TABLE log_table SET ('table.datalake.enable' = 'true');
 ```
 
-See more details about [ALTER TABLE ... SET](engine-flink/ddl/index.md#set-properties) and [ALTER TABLE ... RESET](engine-flink/ddl/index.md#reset-properties) documentation.
+See more details about [ALTER TABLE ... SET](engine-flink/ddl.md#set-properties) and [ALTER TABLE ... RESET](engine-flink/ddl.md#reset-properties) documentation.
 
 ## Storage Options
 
--- a/website/docs/engine-flink/procedures.md
+++ b/website/docs/engine-flink/procedures.md
@@ -18,9 +18,9 @@ SHOW PROCEDURES;
 
 ## Access Control Procedures
 
-Fluss provides procedures to manage Access Control Lists (ACLs) for security and authorization.
+Fluss provides procedures to manage Access Control Lists (ACLs) for security and authorization. See the [Security](../security/overview.md) documentation for more details.
 
-## add_acl
+### add_acl
 
 Add an ACL entry to grant permissions to a principal.
 
@@ -69,7 +69,7 @@ CALL sys.add_acl(
 );
 ```
 
-## drop_acl
+### drop_acl
 
 Remove an ACL entry to revoke permissions.
 
@@ -114,7 +114,7 @@ CALL sys.drop_acl(
 );
 ```
 
-## list_acl
+### list_acl
 
 List ACL entries matching the specified filters.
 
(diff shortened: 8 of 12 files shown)


fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/procedure/SetSharedRocksDBRateLimiterProcedure.java
```diff
@@ -0,0 +1,108 @@
...
+                    configEntry = entry;
+                }
+            }
+
+            if (null == configEntry) {
+                throw new IllegalStateException(
+                        "Failed to set shared RocksDB rate limiter: previous config not found, server does not support rate limiter setting.");
+            }
+
+            long oldValue = MemorySize.parseBytes(configEntry.value());
```
loserwang1024: This can move to server, client just need to alter config. All the clients can reuse it.

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/procedure/SetSharedRocksDBRateLimiterProcedure.java
```diff
@@ -0,0 +1,108 @@
...
+ *
+ * <pre>
+ * -- Set rate limiter to 200MB/s
+ * CALL sys.set_shared_rocksdb_rate_limiter('200MB');
+ *
+ * -- Set rate limiter to 1GB/s
+ * CALL sys.set_shared_rocksdb_rate_limiter('1GB');
+ * </pre>
+ */
+public class SetSharedRocksDBRateLimiterProcedure extends ProcedureBase {
```
swuferhong: It can be abstracted into an `AlterConfigProcedure`, which can then be reused for modifying  other cluster parameter.

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/procedure/GetSharedRocksDBRateLimiterProcedure.java
```diff
@@ -0,0 +1,86 @@
...
+/**
+ * Procedure to get current RocksDB rate limiter configuration from cluster config.
+ *
+ * <p>Usage:
+ *
+ * <pre>
+ * CALL sys.get_shared_rocksdb_rate_limiter();
+ * </pre>
+ */
+public class GetSharedRocksDBRateLimiterProcedure extends ProcedureBase {
```
swuferhong: Why need this? In my side, this parameter is only useful to cluster administrators. If it's really necessary to introduce it, I personally think it should also be abstracted into a "describe config" (i.e., show all configs) operation.

swuferhong (commented):
Thanks for your contributions. I left some comments.

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/procedure/SetSharedRocksDBRateLimiterProcedure.java
platinumhamburg: Thanks for comment, I reimplemented a generic procedure, and moved this validation to server side.

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/procedure/SetSharedRocksDBRateLimiterProcedure.java
platinumhamburg: I reimplemented a generic one.

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/procedure/SetClusterConfigProcedure.java
```diff
@@ -0,0 +1,128 @@
...
+ * <p>Usage examples:
+ *
+ * <pre>
+ * -- Set a configuration
+ * CALL sys.set_cluster_config('kv.shared-rate-limiter.bytes-per-sec', '200MB');
+ * CALL sys.set_cluster_config('datalake.format', 'paimon');
+ *
+ * -- Delete a configuration (reset to default)
+ * CALL sys.set_cluster_config('kv.shared-rate-limiter.bytes-per-sec', NULL);
+ * CALL sys.set_cluster_config('kv.shared-rate-limiter.bytes-per-sec', '');
```
swuferhong: Maybe `set_cluster_config`  and `delete_cluster_config` can be divided into different procedures?

fluss-flink/fluss-flink-common/src/test/java/org/apache/fluss/flink/procedure/FlinkProcedureITCase.java
```diff
@@ -86,12 +87,42 @@ void before() throws ExecutionException, InterruptedException {
...
+                        .await();
+                tEnv.executeSql(
+                                String.format(
+                                        "Call %s.sys.set_cluster_config('%s')",
+                                        CATALOG_NAME, ConfigOptions.DATALAKE_FORMAT.key()))
+                        .await();
+            } catch (Exception e) {
+                // Ignore cleanup errors to avoid masking test failures
+            }
+        }
```
swuferhong: Can these be moved into test cases related to set cluster configuration? It feels a bit odd to place them in afterEach？

swuferhong (commented):
LGTM. Left one minor comment.

website/docs/maintenance/operations/updating-configs.md
```diff
@@ -15,9 +15,13 @@ From Fluss version 0.8 onwards, some of the server configs can be updated withou
 Currently, the supported dynamically updatable server configurations include:
 - `datalake.format`: Enable lakehouse storage by specifying the lakehouse format, e.g., `paimon`, `iceberg`.
 - Options with prefix `datalake.${datalake.format}`
+- `kv.shared-rate-limiter-bytes-per-sec`: Control RocksDB flush and compaction write rate shared across all RocksDB instances on the TabletServer.
 
 
-You can update the configuration of a cluster with [Java client](apis/java-client.md).
+You can update the configuration of a cluster with [Java client](apis/java-client.md) or [Flink Stored Procedures](#using-flink-stored-procedures).
```
wuchong:
```suggestion
You can update the configuration of a cluster with [Java client](#using-java-client) or [Flink Stored Procedures](#using-flink-stored-procedures).
```

website/docs/maintenance/operations/updating-configs.md
```diff
@@ -30,13 +34,50 @@ admin.alterClusterConfigs(
 admin.alterClusterConfigs(
         Collections.singletonList(
                 new AlterConfig(DATALAKE_FORMAT.key(), "paimon", AlterConfigOpType.DELETE)));
+
+// Set RocksDB shared rate limiter to 200MB/sec
+admin.alterClusterConfigs(
+        Collections.singletonList(
+                new AlterConfig("kv.shared-rate-limiter-bytes-per-sec", "209715200", AlterConfigOpType.SET)));
```
wuchong:
```suggestion
                new AlterConfig(KV_SHARED_RATE_LIMITER_BYTES_PER_SEC.key(), "200MB", AlterConfigOpType.SET)));
```

website/docs/maintenance/operations/updating-configs.md
```diff
@@ -30,13 +34,50 @@ admin.alterClusterConfigs(
...
+For certain configurations, Fluss provides convenient Flink stored procedures that can be called directly from Flink SQL.
+
+#### Managing RocksDB Rate Limiter
+
+You can dynamically adjust the shared RocksDB rate limiter for all TabletServers using Flink stored procedures:
+
+**Set Shared RocksDB Rate Limiter:**
+```sql title="Flink SQL"
+-- Set rate limiter to 200MB/sec (recommended, use named argument, only supported since Flink 1.19)
+CALL fluss_catalog.sys.set_shared_rocksdb_rate_limiter(rate_limit => '200MB');
```
wuchong: Invoke `USE fluss_catalog` before all the `CALL` statements, this makes the CALL statements is more reusable as the catalog name maybe different for different users.

website/docs/maintenance/operations/updating-configs.md
```diff
@@ -30,13 +34,50 @@ admin.alterClusterConfigs(
...
+
+-- Set rate limiter to 500MB/sec (use indexed argument)
+CALL fluss_catalog.sys.set_shared_rocksdb_rate_limiter('500MB');
```
wuchong: It seems this is not supported yet, should use `set_cluster_config` in documentation?

fluss-server/src/main/java/org/apache/fluss/server/kv/KvTablet.java
```diff
@@ -214,10 +246,13 @@ public static KvTablet create(
                 schemaGetter);
     }
 
-    private static RocksDBKv buildRocksDBKv(Configuration configuration, File kvDir)
+    private static RocksDBKv buildRocksDBKv(
+            Configuration configuration,
+            File kvDir,
+            @Nullable org.rocksdb.RateLimiter sharedRateLimiter)
```
wuchong: Can we replace the qualified class name with `import`?

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/procedure/GetClusterConfigProcedure.java
```diff
@@ -0,0 +1,124 @@
...
+ * <ul>
+ *   <li>A specific configuration key
+ *   <li>All configurations (when key parameter is null or empty)
+ * </ul>
+ *
+ * <p>Usage examples:
+ *
+ * <pre>
+ * -- Get a specific configuration
+ * CALL sys.get_cluster_config('kv.shared-rate-limiter.bytes-per-sec');
```
wuchong: Replace all the `kv.shared-rate-limiter.bytes-per-sec`  in docs/javadocs to `kv.rocksdb.shared-rate-limiter.bytes-per-sec` which is the true config key.

website/docs/maintenance/configuration.md
```diff
@@ -156,6 +156,7 @@ during the Fluss cluster working.
...
...                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                               |
```
wuchong: Update the config key and the default value

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/procedure/GetClusterConfigProcedure.java
```diff
@@ -0,0 +1,124 @@
...
+                for (ConfigEntry entry : configs) {
+                    results.add(
+                            Row.of(
+                                    entry.key(),
+                                    entry.value(),
+                                    entry.source() != null ? entry.source().name() : "UNKNOWN"));
+                }
+
+                if (results.isEmpty()) {
+                    return new Row[] {Row.of("No cluster configurations found", null, null)};
```
wuchong: If there is no configuration, should just return empty as is? Just like select from an empty table? Otherwise, this may be confusing there is a row in the config result: 

config_key | config_value | config_source
-----|-----|-----
No cluster configurations found | |

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/procedure/GetClusterConfigProcedure.java
```diff
@@ -0,0 +1,124 @@
...
+                        break;
+                    }
+                }
+
+                if (results.isEmpty()) {
+                    return new Row[] {
+                        Row.of(
+                                String.format("Configuration key '%s' not found", configKey),
+                                null,
+                                null)
```
wuchong: ditto

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/procedure/GetClusterConfigProcedure.java
```diff
@@ -0,0 +1,124 @@
...
+ * </ul>
+ *
+ * <p>Usage examples:
+ *
+ * <pre>
+ * -- Get a specific configuration
+ * CALL sys.get_cluster_config('kv.shared-rate-limiter.bytes-per-sec');
+ *
+ * -- Get all cluster configurations
+ * CALL sys.get_cluster_config();
```
wuchong, line 51: Could you please add a **“Procedures”** section under the **“DDL”** section in the **“Engine Flink”** documentation to list all supported procedures?

This can be tracked in a separate issue if preferred.  

For the format, I’d recommend following the style used by Iceberg, using **headers and descriptive text** rather than tables. 

Examples:  
- https://iceberg.apache.org/docs/nightly/spark-procedures/#spark-procedures  
- https://paimon.apache.org/docs/master/flink/procedures/

fluss-server/src/main/java/org/apache/fluss/server/kv/KvConfigValidator.java
```diff
@@ -0,0 +1,132 @@
...
+ * Stateless validator for RocksDB shared rate limiter configuration.
+ *
+ * <p>This validator can be registered on CoordinatorServer without requiring a KvManager instance,
+ * enabling early validation of KV configs before they are persisted to ZooKeeper.
+ *
+ * <p>The validator enforces the following rules for RocksDB shared rate limiter:
+ *
+ * <ul>
+ *   <li>Cannot enable rate limiter dynamically if it was not enabled at server startup
+ *   <li>Cannot disable rate limiter dynamically if it was enabled at server startup
```
wuchong: I don’t fully understand the reason behind this limitation. Is there significant overhead in always having a `RateLimiter` enabled in RocksDB, even when configured with a very high throughput limit (e.g., 1 TB/s)?  

If the performance cost is negligible and especially since we’re enabling the rate limiter by default, I’d suggest treating a very high value (or `Long.MAX_VALUE`) as the effective “disabled” state (this also means we don't rely on `0` to disable)

I think switching between enable and disable is a common and useful functionality. Otherwise, it's very confusing the documentation says set to `0` to disable, but it doesn't work.

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/procedure/GetClusterConfigProcedure.java
platinumhamburg: I'll return empty rows directly.

platinumhamburg:
I've updated this PR and resolved all comments above, thanks for review again.

website/docs/engine-flink/ddl.md
```diff
@@ -406,3 +406,252 @@ WITH (
 - Only continuous refresh mode is supported
 - Schema is automatically derived from the query
 - Materialized tables are stored as regular Fluss tables with special metadata
+
+## Procedures
```
wuchong: What I mean is to place the "Procedures" documentation as a separate page under the "DDL" page. Additionally, each procedure method should be formatted as an H3 or H2 header, so it appears in the TOC (table of contents) and can be directly linked via URL.

fluss-server/src/main/java/org/apache/fluss/server/DynamicServerConfig.java
```diff
@@ -115,68 +148,296 @@ boolean isAllowedConfig(String key) {
...
+            } catch (Exception e) {
+                throw new ConfigException(
+                        String.format(
+                                "Cannot parse '%s' as %s for config '%s'",
+                                newValueStr,
+                                configOption.isList()
+                                        ? "List<" + configOption.getClazz().getSimpleName() + ">"
+                                        : configOption.getClazz().getSimpleName(),
+                                configKey),
+                        e);
```
wuchong, line 360: Add the exception message into `ConfigException` top error message, because `ApiException` will truncate the exception stack and will hidden the root cause.

fluss-server/src/main/java/org/apache/fluss/server/DynamicServerConfig.java
```diff
@@ -115,68 +148,296 @@ boolean isAllowedConfig(String key) {
...
-                        break;
-                    }
-                }
+        // Compute effective config changes (merge with initial configs)
+        Map<String, String> effectiveChanges =
+                computeEffectiveChanges(newDynamicConfigs, skipErrorConfig);
+
+        // Early return if no effective changes
+        if (effectiveChanges.isEmpty()) {
+            return;
```
wuchong, line 158: It would be helpful to log an info message when no changes are detected. This aids debugging, for example, when an `ALTER CONFIG` command appears to have no effect.

fluss-server/src/main/java/org/apache/fluss/server/DynamicServerConfig.java
```diff
@@ -115,68 +148,296 @@ boolean isAllowedConfig(String key) {
...
+        // Build new configuration by merging initial + dynamic configs
+        Map<String, String> newConfigMap = buildConfigMap(effectiveChanges);
+        Configuration newConfig = Configuration.fromMap(newConfigMap);
+
+        // Apply changes to all registered ServerReconfigurable instances
+        applyToServerReconfigurables(newConfig, skipErrorConfig);
+
+        // Update internal state
+        updateInternalState(newConfig, newConfigMap, newDynamicConfigs);
+        LOG.info("Dynamic configs changed: {}", newDynamicConfigs);
```
wuchong: I think logging the the chagned configs `effectiveChanges` is more helpful than the full set `newDynamicConfigs`.

fluss-server/src/test/java/org/apache/fluss/server/kv/KvTabletTest.java
```diff
@@ -191,7 +191,8 @@ private KvTablet createKvTablet(
                 rowMerger,
                 DEFAULT_COMPRESSION,
                 schemaGetter,
-                tableConf.getChangelogImage());
+                tableConf.getChangelogImage(),
+                null);
```
wuchong: ditto

fluss-server/src/main/java/org/apache/fluss/server/kv/rocksdb/RocksDBResourceContainer.java
```diff
@@ -73,29 +74,41 @@ public class RocksDBResourceContainer implements AutoCloseable {
...
     public RocksDBResourceContainer(ReadableConfig configuration, @Nullable File instanceBasePath) {
-        this(configuration, instanceBasePath, false);
+        this(configuration, instanceBasePath, false, null);
     }
 
     public RocksDBResourceContainer(
             ReadableConfig configuration,
             @Nullable File instanceBasePath,
             boolean enableStatistics) {
+        this(configuration, instanceBasePath, enableStatistics, null);
```
wuchong: **Nit:** It would be better to use a default `RateLimiter` instance instead of `null`, and add a `checkNotNull` guard for the `sharedRateLimiter` parameter in the constructor.  

Currently, many places still pass `null` for `sharedRateLimiter`. While this is currently used test-only, it risks leaking into production code and could introduce subtle bugs in the future. Using a default rate limiter would make the API safer and more robust.

wuchong (commented):
Thanks @platinumhamburg , this PR is already in a good shape. I only left some minor comments.

fluss-server/src/main/java/org/apache/fluss/server/kv/rocksdb/RocksDBResourceContainer.java
platinumhamburg: Since RocksDBResourceContainer is instance-scoped, a non-shared RateLimiter is meaningless and cannot effectively control service-level traffic. I will remove the overloaded constructor that doesn’t accept a RateLimiter, and instead create a shared RateLimiter at the KvManager layer.

platinumhamburg:
> Thanks @platinumhamburg , this PR is already in a good shape. I only left some minor comments.

Thanks for your detailed review, all comments resolved now.

wuchong (approved):
@platinumhamburg Sorry for the confusion! What I meant by “place the *Procedures* documentation as a separate page under the *DDL* page” was actually to put **“Procedures” at the same navigation level as “DDL”**—not nested inside it—just like Iceberg does: [https://iceberg.apache.org/docs/nightly/spark-procedures/](https://iceberg.apache.org/docs/nightly/spark-procedures/).

I’ve updated the structure accordingly and reverted the changes to `ddl.md`.  
I’ll merge this PR once CI passes.

PR #2189, opened 2025-12-16
[lake] Support schema evolution for lake-enabled tables with AddColum…

### LINKED ISSUE #2128: Support schema evolution for data lake enabled table for paimon
### Search before asking

### Description

we alreay support schema evolution for normal table, but we also support support for data lake enabled table.

Two things to consider:
- we will alter both paimon table and fluss table, what if when success to alter paimon table, but fail to alter fluss table?
- What happens in tiering when we already alter add new column to paimon table, but the records to be tiered to paimon still missing the column?

### Willingness to contribute

buvb:
### Purpose

Linked issue: close #2128

This PR enables schema evolution for datalake-enabled tables, specifically supporting `ADD COLUMN ... LAST` with `NULLABLE` columns. When a user executes `ALTER TABLE ADD COLUMN` on a lake-enabled table, the schema change is first applied to Fluss (source of truth), then synchronized to Paimon.

### Brief change log

- **CoordinatorService**: Pass `LakeCatalog` and `LakeCatalogContext` to [alterTableSchema()]for Paimon synchronization
- **MetadataManager**: Add [syncSchemaChangesToLake()] to sync schema changes to Paimon after Fluss schema update; skip schema registration if schema unchanged (retry idempotency)
- **SchemaUpdate**: Support idempotent [addColumn()] - if column already exists with same type and comment, treat as no-op
- **PaimonLakeCatalog**: Handle `ColumnAlreadyExistException` as idempotent success for retry scenarios
- **PaimonConversions**: Map Fluss `AddColumn` to Paimon `SchemaChange`, inserting new column before system columns
- **FlussRecordAsPaimonRow**: Handle tiering transition period when Fluss record is wider than Paimon schema using `min(internalRow.getFieldCount(), businessFieldCount)`

### Tests

- `LakeEnabledTableCreateITCase#testAlterLakeEnabledTableSchema` - Verify ADD COLUMN syncs to Paimon with correct type and comment
- `FlussRecordAsPaimonRowTest#testFlussRecordWiderThanPaimonSchema` - Verify tiering doesn't crash when Fluss record has more fields than Paimon schema
- `FlussRecordAsPaimonRowTest#testPaimonSchemaWiderThanFlussRecord` - Verify padding NULL for missing fields

### API and Format

### Documentation

No documentation changes required for this MVP. Future documentation may be needed when more schema evolution operations are supported.

### DIFF
--- a/fluss-lake/fluss-lake-paimon/src/main/java/org/apache/fluss/lake/paimon/PaimonLakeCatalog.java
+++ b/fluss-lake/fluss-lake-paimon/src/main/java/org/apache/fluss/lake/paimon/PaimonLakeCatalog.java
@@ -19,6 +19,7 @@ package org.apache.fluss.lake.paimon;
 
 import org.apache.fluss.annotation.VisibleForTesting;
 import org.apache.fluss.config.Configuration;
+import org.apache.fluss.exception.InvalidAlterTableException;
 import org.apache.fluss.exception.TableAlreadyExistException;
 import org.apache.fluss.exception.TableNotExistException;
 import org.apache.fluss.lake.lakestorage.LakeCatalog;
@@ -38,6 +39,8 @@ import org.apache.paimon.table.FileStoreTable;
 import org.apache.paimon.table.Table;
 import org.apache.paimon.types.DataType;
 import org.apache.paimon.types.DataTypes;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import java.util.LinkedHashMap;
 import java.util.List;
@@ -54,6 +57,7 @@ import static org.apache.fluss.metadata.TableDescriptor.TIMESTAMP_COLUMN_NAME;
 /** A Paimon implementation of {@link LakeCatalog}. */
 public class PaimonLakeCatalog implements LakeCatalog {
 
+    private static final Logger LOG = LoggerFactory.getLogger(PaimonLakeCatalog.class);
     public static final LinkedHashMap<String, DataType> SYSTEM_COLUMNS = new LinkedHashMap<>();
 
     static {
@@ -113,14 +117,19 @@ public class PaimonLakeCatalog implements LakeCatalog {
             // Compare current Paimon table schema with expected target schema before altering
             if (shouldAlterTable(tablePath, tableChanges)) {
                 alterTable(tablePath, paimonSchemaChanges);
+            } else {
+                // If schemas already match, treat as idempotent success
+                LOG.info(
+                        "Skipping schema evolution for Paimon table {} because the column(s) to add {} already exist.",
+                        tablePath,
+                        tableChanges);
             }
-            // If schemas already match, treat as idempotent success
         } catch (Catalog.ColumnAlreadyExistException e) {
             // This shouldn't happen if shouldAlterTable works correctly, but keep as safeguard
-            throw new RuntimeException("Unexpected ColumnAlreadyExistException", e);
+            throw new InvalidAlterTableException(e.getMessage());
         } catch (Catalog.ColumnNotExistException e) {
             // This shouldn't happen for AddColumn operations
-            throw new RuntimeException(e);
+            throw new InvalidAlterTableException(e.getMessage());
         }
     }
 
@@ -161,9 +170,9 @@ public class PaimonLakeCatalog implements LakeCatalog {
                                                 .FlussDataTypeToPaimonDataType.INSTANCE);
 
                 if (!field.type().equals(expectedType)) {
-                    throw new IllegalStateException(
+                    throw new InvalidAlterTableException(
                             String.format(
-                                    "Column %s already exists but with different type. "
+                                    "Column '%s' already exists but with different type. "
                                             + "Existing: %s, Expected: %s",
                                     columnName, field.type(), expectedType));
                 }
@@ -176,7 +185,7 @@ public class PaimonLakeCatalog implements LakeCatalog {
                                         && existingComment.equals(expectedComment));
 
                 if (!commentsMatch) {
-                    throw new IllegalStateException(
+                    throw new InvalidAlterTableException(
                             String.format(
                                     "Column %s already exists but with different comment. "
                                             + "Existing: %s, Expected: %s",
@@ -208,7 +217,7 @@ public class PaimonLakeCatalog implements LakeCatalog {
                 }
             } catch (Catalog.TableNotExistException tableNotExistException) {
                 // shouldn't happen in normal cases
-                throw new RuntimeException(
+                throw new InvalidAlterTableException(
                         String.format(
                                 "Failed to create table %s in Paimon. The table already existed "
                                         + "during the initial creation attempt, but subsequently "
--- a/fluss-lake/fluss-lake-paimon/src/test/java/org/apache/fluss/lake/paimon/tiering/FlussRecordAsPaimonRowTest.java
+++ b/fluss-lake/fluss-lake-paimon/src/test/java/org/apache/fluss/lake/paimon/tiering/FlussRecordAsPaimonRowTest.java
@@ -168,66 +168,6 @@ class FlussRecordAsPaimonRowTest {
     }
 
     @Test
-    void testPaimonSchemaWiderThanFlussRecord() {
-        int tableBucket = 0;
-        RowType tableRowType =
-                RowType.of(
-                        new org.apache.paimon.types.BooleanType(),
-                        new org.apache.paimon.types.VarCharType(),
-                        // append three system columns: __bucket, __offset,__timestamp
-                        new org.apache.paimon.types.IntType(),
-                        new org.apache.paimon.types.BigIntType(),
-                        new org.apache.paimon.types.LocalZonedTimestampType(3));
-
-        FlussRecordAsPaimonRow flussRecordAsPaimonRow =
-                new FlussRecordAsPaimonRow(tableBucket, tableRowType);
-
-        long logOffset = 7L;
-        long timeStamp = System.currentTimeMillis();
-        GenericRow genericRow = new GenericRow(1);
-        genericRow.setField(0, true);
-        LogRecord logRecord = new GenericRecord(logOffset, timeStamp, APPEND_ONLY, genericRow);
-        flussRecordAsPaimonRow.setFlussRecord(logRecord);
-
-        assertThat(flussRecordAsPaimonRow.getFieldCount()).isEqualTo(5);
-
-        assertThat(flussRecordAsPaimonRow.getBoolean(0)).isTrue();
-        assertThat(flussRecordAsPaimonRow.isNullAt(1)).isTrue();
-        assertThat(flussRecordAsPaimonRow.getInt(2)).isEqualTo(tableBucket);
-        assertThat(flussRecordAsPaimonRow.getLong(3)).isEqualTo(logOffset);
-        assertThat(flussRecordAsPaimonRow.getLong(4)).isEqualTo(timeStamp);
-    }
-
-    @Test
-    void testFlussRecordWiderThanPaimonSchema() {
-        // With "Lake First" strategy, Fluss record wider than Paimon schema should throw exception
-        int tableBucket = 0;
-        RowType tableRowType =
-                RowType.of(
-                        new org.apache.paimon.types.BooleanType(),
-                        // append three system columns: __bucket, __offset,__timestamp
-                        new org.apache.paimon.types.IntType(),
-                        new org.apache.paimon.types.BigIntType(),
-                        new org.apache.paimon.types.LocalZonedTimestampType(3));
-
-        FlussRecordAsPaimonRow flussRecordAsPaimonRow =
-                new FlussRecordAsPaimonRow(tableBucket, tableRowType);
-
-        long logOffset = 7L;
-        long timeStamp = System.currentTimeMillis();
-        GenericRow genericRow = new GenericRow(2);
-        genericRow.setField(0, true);
-        genericRow.setField(1, BinaryString.fromString("extra"));
-        LogRecord logRecord = new GenericRecord(logOffset, timeStamp, APPEND_ONLY, genericRow);
-
-        // Should throw exception instead of silently truncating data
-        assertThatThrownBy(() -> flussRecordAsPaimonRow.setFlussRecord(logRecord))
-                .isInstanceOf(IllegalStateException.class)
-                .hasMessageContaining(
-                        "Fluss record has 2 fields but Paimon schema only has 1 business fields");
-    }
-
-    @Test
     void testArrayTypeWithIntElements() {
         int tableBucket = 0;
         RowType tableRowType =
--- a/fluss-server/src/main/java/org/apache/fluss/server/coordinator/MetadataManager.java
+++ b/fluss-server/src/main/java/org/apache/fluss/server/coordinator/MetadataManager.java
@@ -343,6 +343,11 @@ public class MetadataManager {
                 // Update Fluss schema (ZK) after Lake sync succeeds
                 if (!newSchema.equals(table.getSchema())) {
                     zookeeperClient.registerSchema(tablePath, newSchema, table.getSchemaId() + 1);
+                } else {
+                    LOG.info(
+                            "Skipping schema evolution for table {} because the column(s) to add {} already exist.",
+                            tablePath,
+                            schemaChanges);
                 }
             }
         } catch (Exception e) {
--- a/fluss-server/src/main/java/org/apache/fluss/server/coordinator/SchemaUpdate.java
+++ b/fluss-server/src/main/java/org/apache/fluss/server/coordinator/SchemaUpdate.java
@@ -88,7 +88,8 @@ public class SchemaUpdate {
         if (existingColumn != null) {
             // Allow idempotent retries: if column name/type/comment match existing, treat as no-op
             if (!existingColumn.getDataType().equals(addColumn.getDataType())
-                    || !Objects.equals(existingColumn.getComment(), addColumn.getComment())) {
+                    || !Objects.equals(
+                            existingColumn.getComment().orElse(null), addColumn.getComment())) {
                 throw new IllegalArgumentException(
                         "Column " + addColumn.getName() + " already exists.");
             }
--- a/fluss-common/src/main/java/org/apache/fluss/metadata/TableChange.java
+++ b/fluss-common/src/main/java/org/apache/fluss/metadata/TableChange.java
@@ -257,6 +257,22 @@ public interface TableChange {
         public ColumnPosition getPosition() {
             return position;
         }
+
+        @Override
+        public String toString() {
+            return "AddColumn{"
+                    + "name='"
+                    + name
+                    + '\''
+                    + ", dataType="
+                    + dataType
+                    + ", comment='"
+                    + comment
+                    + '\''
+                    + ", position="
+                    + position
+                    + '}';
+        }
     }
 
     /** A table change to drop a column. */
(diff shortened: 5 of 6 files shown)


fluss-lake/fluss-lake-paimon/src/test/java/org/apache/fluss/lake/paimon/tiering/FlussRecordAsPaimonRowTest.java
```diff
@@ -163,4 +163,64 @@ void testPrimaryKeyTableRecord() {
...
+
+        assertThat(flussRecordAsPaimonRow.getBoolean(0)).isTrue();
+        assertThat(flussRecordAsPaimonRow.isNullAt(1)).isTrue();
+        assertThat(flussRecordAsPaimonRow.getInt(2)).isEqualTo(tableBucket);
+        assertThat(flussRecordAsPaimonRow.getLong(3)).isEqualTo(logOffset);
+        assertThat(flussRecordAsPaimonRow.getLong(4)).isEqualTo(timeStamp);
+    }
+
+    @Test
+    void testFlussRecordWiderThanPaimonSchema() {
```
loserwang1024, line 583: Please test Fluss Record narrow than Paimon Schema. Currently, each time we read a roundes of a table's splits, the reader will get table info from fluss at first, then get table info from paimon catalog later. If add column between this, the paimon schema is wider than fluss record, we need to put into null.

fluss-lake/fluss-lake-paimon/src/main/java/org/apache/fluss/lake/paimon/tiering/FlussRecordAsPaimonRow.java
```diff
@@ -27,35 +27,38 @@
...
     }
 
     public void setFlussRecord(LogRecord logRecord) {
         this.logRecord = logRecord;
         this.internalRow = logRecord.getRow();
-        this.originRowFieldCount = internalRow.getFieldCount();
-        checkState(
-                originRowFieldCount == tableRowType.getFieldCount() - SYSTEM_COLUMNS.size(),
-                "The paimon table fields count must equals to LogRecord's fields count.");
+        this.originRowFieldCount = Math.min(internalRow.getFieldCount(), businessFieldCount);
```
loserwang1024: If businessFieldCount is smaller than internalRow.getFieldCount(), just throw exception. Because data will be truncated, it is dangerous. To be honest,   businessFieldCount must be not less than  internalRow.getFieldCount() because we only support add column at last now.

The reason is as follows:
 Currently, each time we read a roundes of a table's splits, the reader will get table info from fluss at first, then get table info from paimon catalog later. If add column between this, the paimon schema is wider than fluss record, we need to put into null.

Maybe you can refer what PaddingRow does.(Similiar like it, when write data to fluss, flink data maybe narrower than fluss schema.

fluss-lake/fluss-lake-paimon/src/main/java/org/apache/fluss/lake/paimon/tiering/FlussRecordAsPaimonRow.java
```diff
@@ -27,35 +27,38 @@
...
 import static org.apache.fluss.lake.paimon.utils.PaimonConversions.toRowKind;
-import static org.apache.fluss.utils.Preconditions.checkState;
 
 /** To wrap Fluss {@link LogRecord} as paimon {@link InternalRow}. */
 public class FlussRecordAsPaimonRow extends FlussRowAsPaimonRow {
 
     private final int bucket;
     private LogRecord logRecord;
     private int originRowFieldCount;
+    private final int businessFieldCount;
```
loserwang1024, line 37: maybe dataFieldCount?

loserwang1024 (commented):
Would you like to add test in PaimonTieringITCase or PaimonTieringTest for a e2e test (just like what PaimonTieringITCase#testTieringForAlterTable does). Add coumn before and bewteen tier service.

fluss-server/src/main/java/org/apache/fluss/server/coordinator/MetadataManager.java
```diff
@@ -322,24 +322,26 @@ public long createTable(
...
-                        "Schema evolution is currently not supported for tables with datalake enabled.");
-            }
-
             // validate the table column changes
             if (!schemaChanges.isEmpty()) {
                 Schema newSchema = SchemaUpdate.applySchemaChanges(table, schemaChanges);
-                // update the schema
-                zookeeperClient.registerSchema(tablePath, newSchema, table.getSchemaId() + 1);
+                // update the schema in Fluss (ZK) first - Fluss is the source of truth
+                if (!newSchema.equals(table.getSchema())) {
```
loserwang1024: It better check whether lakehouse can do it before  zookeeperClient.registerSchema.
1.  check that fluss can apply schema change.
SchemaUpdate.applySchemaChanges(table, schemaChanges)
2. Check whether the schema can be apply to lake.
3.  zookeeperClient.registerSchema(tablePath, newSchema, table.getSchemaId() + 1). This can be rarely happen because only zookeeper.
4. syncSchemaChangesToLake.
For example, currently, iceberg catalog is not support add column. If we  zookeeperClient.registerSchema successfully, problem will occur.

Also a problem left is what if  zookeeperClient.registerSchema success but syncSchemaChangesToLake fails? A assumed that:
1. Add column A 
- zookeeperClient.registerSchema success
- syncSchemaChangesToLake fails
2. Add column B
- zookeeperClient.registerSchema success
- syncSchemaChangesToLake success
3. ReAdd column A 
- SchemaUpdate.applySchemaChanges is Idempotent
- syncSchemaChangesToLake success,

Finally, Fluss Table is : column A, column B
Finally, Lake Table is : column B, column A.

I advice each time before zookeeperClient.registerSchema, check whether the column number is same:
1. Add column A 
- zookeeperClient.registerSchema success
- syncSchemaChangesToLake fails
2. Add column B
-  Columns number is not same, just refuse it.


@wuchong , WDYT?

fluss-server/src/main/java/org/apache/fluss/server/coordinator/SchemaUpdate.java
```diff
@@ -83,9 +84,14 @@ public SchemaUpdate applySchemaChange(TableChange columnChange) {
     }
 
     private SchemaUpdate addColumn(TableChange.AddColumn addColumn) {
-        if (existedColumns.containsKey(addColumn.getName())) {
-            throw new IllegalArgumentException(
-                    "Column " + addColumn.getName() + " already exists.");
+        Schema.Column existingColumn = existedColumns.get(addColumn.getName());
```
loserwang1024, line 87: Maybe we can add a comment to display why need it.

fluss-lake/fluss-lake-paimon/src/test/java/org/apache/fluss/lake/paimon/tiering/FlussRecordAsPaimonRowTest.java
buvb: Good suggestion. I'll add a test in `PaimonTieringITCase` to verify ADD COLUMN works correctly during tiering.

fluss-lake/fluss-lake-paimon/src/main/java/org/apache/fluss/lake/paimon/tiering/FlussRecordAsPaimonRow.java
buvb: I prefer `businessFieldCount` because it clearly distinguishes from system columns (`__bucket`, `__offset`, `__timestamp`). `dataFieldCount` might be confused with total field count.

fluss-server/src/main/java/org/apache/fluss/server/coordinator/SchemaUpdate.java
buvb: Will add a comment to explain the idempotent handling in [SchemaUpdate]

fluss-lake/fluss-lake-paimon/src/main/java/org/apache/fluss/lake/paimon/tiering/FlussRecordAsPaimonRow.java
buvb: About [businessFieldCount < internalRow.getFieldCount()]
 I understand your concern, but this scenario CAN happen during the transition period:
 **Timeline:**
1. User executes `ALTER TABLE ADD COLUMN c3`
2. Fluss schema updated (now has c3)
3. New Fluss records are written with c3
4. Tiering starts consuming these new records
5. But [syncSchemaChangesToLake()] hasn't completed yet → Paimon schema still doesn't have c3

In this case, [internalRow.getFieldCount() > businessFieldCount]
If we throw an exception here, tiering will crash during the transition.
Using `Math.min()` to ignore the extra field is safe because:
- The data is not "lost" - it's still in Fluss
- Once Paimon schema syncs, subsequent tiering will include the new column
- This is a temporary transition state, not a permanent data loss

fluss-server/src/main/java/org/apache/fluss/server/coordinator/MetadataManager.java
buvb: ZK success but Lake sync fails
This is already handled by the idempotent design:
- [SchemaUpdate.addColumn()] - treats existing column with same type/comment as no-op
- [PaimonLakeCatalog.alterTable()] - catches `ColumnAlreadyExistException` as success
 
So when user retries, both Fluss and Paimon will handle it gracefully.

loserwang1024: @buvb Yes, If no other add column is happened, it will handle it gracefully. However, as I mention before:

Aassumed that:
1. Add column A
- zookeeperClient.registerSchema success
- syncSchemaChangesToLake fails

2. Add column B
- zookeeperClient.registerSchema success
- syncSchemaChangesToLake success

3. ReAdd column A
- SchemaUpdate.applySchemaChanges is Idempotent
- syncSchemaChangesToLake success.

Finally, Fluss Table is : column A, column B
Finally, Lake Table is : column B, column A.


I have been talked with jark. He advice to add column in lakehouse before fluss. Maybe each time before apply schema change to paimon, we can get the paimon rowType, and check whether is safe.

`RowType rowType = paimonCatalog.getTable(new Identifier(tablePath.getDatabaseName(), tablePath.getTableName())).rowType();`

fluss-lake/fluss-lake-paimon/src/main/java/org/apache/fluss/lake/paimon/tiering/FlussRecordAsPaimonRow.java
loserwang1024: > The data is not "lost" - it's still in Fluss

Yes, but not in paimon. This c3 will be truncated. Tiering won't resent the c3. When union read from paimon, will lacks this column. Error is better than loss data in silence.

I hava been talked with @wuchong , he advice add column in paimon before  fluss( just like what create table does).

loserwang1024:
@luoyuxia @wuchong , Would you like to help a final Review?

buvb:
@luoyuxia @wuchong The main changes are as follows
- Implemented lake-first schema sync, idempotent AddColumn handling, Paimon addColumn mapping, and tiering safeguards (fail when Fluss is wider, pad NULL when Paimon is wider), plus IT/unit coverage.
- Recent fixes: PaimonTieringITCase now writes values for the new column after ADD COLUMN and asserts actual column names; offset check uses a dynamic index to avoid type mismatches after schema changes.
- Testing: In an environment that allows binding ports, please run ./mvnw -pl fluss-lake/fluss-lake-paimon -DskipTests=false test. In restricted environments, ReCreateSameTableAfterTieringTest fails due to port bind being denied; all other tests pass.

fluss-lake/fluss-lake-paimon/src/test/java/org/apache/fluss/lake/paimon/tiering/FlussRecordAsPaimonRowTest.java
```diff
@@ -163,4 +164,64 @@ void testPrimaryKeyTableRecord() {
...
+        LogRecord logRecord = new GenericRecord(logOffset, timeStamp, APPEND_ONLY, genericRow);
+        flussRecordAsPaimonRow.setFlussRecord(logRecord);
+
+        assertThat(flussRecordAsPaimonRow.getFieldCount()).isEqualTo(5);
+
+        assertThat(flussRecordAsPaimonRow.getBoolean(0)).isTrue();
+        assertThat(flussRecordAsPaimonRow.isNullAt(1)).isTrue();
+        assertThat(flussRecordAsPaimonRow.getInt(2)).isEqualTo(tableBucket);
+        assertThat(flussRecordAsPaimonRow.getLong(3)).isEqualTo(logOffset);
+        assertThat(flussRecordAsPaimonRow.getLong(4)).isEqualTo(timeStamp);
```
wuchong: should assert `getTimestamp` because this is a timestamp type.

fluss-lake/fluss-lake-paimon/src/main/java/org/apache/fluss/lake/paimon/PaimonLakeCatalog.java
```diff
@@ -110,8 +110,10 @@ public void alterTable(TablePath tablePath, List<TableChange> tableChanges, Cont
         try {
             List<SchemaChange> paimonSchemaChanges = toPaimonSchemaChanges(tableChanges);
             alterTable(tablePath, paimonSchemaChanges);
-        } catch (Catalog.ColumnAlreadyExistException | Catalog.ColumnNotExistException e) {
-            // shouldn't happen before we support schema change
+        } catch (Catalog.ColumnAlreadyExistException e) {
+            // Column already exists, treat as idempotent success for retry scenarios.
```
wuchong: Given that we may execute multiple `TableChange` operations in a single statement (e.g., adding several columns at once), **blindly ignoring `ColumnAlreadyExistException` could silently skip the addition of some columns**, leading to an incomplete schema update.

A simpler and safer approach, in my view, is to **compare the current Paimon table schema with the expected target schema before performing any `ALTER TABLE`** (like how you did in `org.apache.fluss.server.coordinator.MetadataManager#alterTableSchema`):

- If the schemas **differ**, proceed with the `ALTER TABLE` and **report any errors faithfully to the user** (who can then re-execute if needed).
- If the schemas **already match**, log a clear message (e.g., “Column(s) already exist—skipping ALTER TABLE”) and skip the operation.

This ensures correctness, avoids silent failures, and provides transparent feedback to the user.

fluss-server/src/main/java/org/apache/fluss/server/coordinator/MetadataManager.java
```diff
@@ -322,24 +322,28 @@ public long createTable(
...
-                zookeeperClient.registerSchema(tablePath, newSchema, table.getSchemaId() + 1);
+
+                // Lake First: sync to Lake before updating Fluss schema
+                syncSchemaChangesToLake(
+                        tablePath, table, schemaChanges, lakeCatalog, lakeCatalogContext);
+
+                // Update Fluss schema (ZK) after Lake sync succeeds
+                if (!newSchema.equals(table.getSchema())) {
+                    zookeeperClient.registerSchema(tablePath, newSchema, table.getSchemaId() + 1);
+                }
```
wuchong, line 351: log a clear message (e.g., “Column(s) already exist—skipping ALTER TABLE”) and skip the operation.

fluss-server/src/main/java/org/apache/fluss/server/coordinator/SchemaUpdate.java
```diff
@@ -83,9 +84,15 @@ public SchemaUpdate applySchemaChange(TableChange columnChange) {
...
+        if (existingColumn != null) {
+            // Allow idempotent retries: if column name/type/comment match existing, treat as no-op
+            if (!existingColumn.getDataType().equals(addColumn.getDataType())
+                    || !Objects.equals(existingColumn.getComment(), addColumn.getComment())) {
```
wuchong:
```suggestion
                    || !Objects.equals(existingColumn.getComment().orElse(null), addColumn.getComment())) {
```

The return type of `existingColumn.getComment()` is `Optional<String>` which is not comparable with `addColumn.getComment()`

fluss-lake/fluss-lake-paimon/src/test/java/org/apache/fluss/lake/paimon/tiering/PaimonTieringITCase.java
```diff
@@ -604,6 +605,65 @@ private CloseableIterator<org.apache.paimon.data.InternalRow> getPaimonRowClosea
...
+            List<String> fieldNames = paimonTable.rowType().getFieldNames();
+
+            // Should have: a, b, c3, __bucket, __offset, __timestamp
+            assertThat(fieldNames).contains("a", "b", "c3");
+
+            // 9. Verify all data is present in Paimon (no data loss)
+            List<InternalRow> allRows = new ArrayList<>();
+            allRows.addAll(initialRows);
+            allRows.addAll(newRows);
+            checkDataInPaimonAppendOnlyTable(tablePath, allRows, 0);
```
wuchong: I suggest constructing an **expected list of rows** that includes all six columns (the new column, original user columns, and relevant system columns), and then asserting that this expected list exactly matches the rows read from Paimon. This will ensure comprehensive validation of both schema evolution and data correctness.

fluss-lake/fluss-lake-paimon/src/test/java/org/apache/fluss/lake/paimon/tiering/PaimonTieringITCase.java
```diff
@@ -604,6 +605,65 @@ private CloseableIterator<org.apache.paimon.data.InternalRow> getPaimonRowClosea
...
+            assertReplicaStatus(tableBucket, 6);
+
+            // 8. Verify Paimon table has the new column
+            Identifier tableIdentifier =
+                    Identifier.create(tablePath.getDatabaseName(), tablePath.getTableName());
+            FileStoreTable paimonTable = (FileStoreTable) paimonCatalog.getTable(tableIdentifier);
+            List<String> fieldNames = paimonTable.rowType().getFieldNames();
+
+            // Should have: a, b, c3, __bucket, __offset, __timestamp
+            assertThat(fieldNames).contains("a", "b", "c3");
```
wuchong: assert exactly all the field names and order.

wuchong (approved):
I rebased branch and fixed some problems. Will merge when CI is passed.

wuchong:
@buvb By the way, Please do not use "git merge" to merge branches, otherwise the changes is hard to track. Please use "git rebase" to rebase branches instead. IntelliJ IDEA provide an easy tool to do git rebase, you can find the tool via `VCS -> Git -> Rebase`.

buvb:
@wuchong,

  Thank you so much for reviewing, rebasing the branch, and fixing the remaining issues before merging! I really appreciate your help and patience.

  I apologize for not fully implementing all the review comments in my commits. I encountered some integration test failures on my local environment that prevented me from running the full test suite successfully. That's why some of the improvements (like the complete null-safe comment comparison and the shouldAlterTable() logic) were incomplete in my submission. I should have communicated this issue earlier.

  Thank you for stepping in and completing the implementation.

  Regarding the git workflow: no more git merge, use git rebase instead. This will keep the commit history cleaner and easier to track.


  Best regards,
  buvb

wuchong:
@buvb sorry, I didn't notice that was a working-in-progress commit, so I took over and finish the remaining works. Thank you again for your contribution!

PR #2190, opened 2025-12-17
[common] Introduce MAP type for ARROW, COMPACTED and INDEXED formats

### LINKED ISSUE #1973: Support Map type in log table (Arrow row format)
### Search before asking

### Motivation

Add Map type support for log table scenarios to handle key-value structured data.

### Solution

- Extend Arrow row format to support Map type serialization/deserialization
- Implement Map type encoding/decoding in client SDK
- Add validation for Map type in schema definitions
- Ensure end-to-end data flow works for Map type in log tables

### Anything else?

_No response_

### Willingness to contribute

XuQianJin-Stars:
### Purpose

Linked issue: close #1973

Introduce MAP type support for INDEXED format, achieving feature parity with ARROW and COMPACTED formats for MAP data type handling.

### Brief change log

- Added `readMap()` method to `IndexedRowReader` for MAP type deserialization, following the same pattern as `readArray()`
- Removed `UnsupportedOperationException` for MAP type in `IndexedRowReader.createFieldReader()` 
- Added missing `InternalMap` import to `IndexedRowReader`
- Updated `BinaryWriterTest.testCreateValueSetterForMapThrowsException()` to `testValueSetterWithMapType()` - changed from exception test to positive validation test
- Extended `IndexedRowTest.assertAllTypeEquals()` to include MAP type assertions (validates size, key array, value array, and null handling)
- Extended `IndexedRowReaderTest.assertAllTypeEqualsForReader()` to verify MAP reading

### Tests

**Unit Tests:**
- `IndexedRowTest` (7 tests) - Validates MAP read/write operations with IndexedRow
- `IndexedRowReaderTest` (2 tests) - Validates MAP deserialization with IndexedRowReader  
- `BinaryWriterTest` (18 tests) - Validates MAP type ValueWriter creation and usage
- `ArrowReaderWriterTest` (2 tests) - Confirms ARROW format MAP support continues to work
- `CompactedRowTest` (20 tests) - Confirms COMPACTED format MAP support continues to work

**Test Results:** All 49 tests passed successfully (0 failures, 0 errors, 0 skipped)

**Test Command:**
mvn test -pl fluss-common -Dtest=IndexedRowTest,IndexedRowReaderTest,BinaryWriterTest,ArrowReaderWriterTest,CompactedRowTest -Dcheckstyle.skip=true -Dspotless.check.skip=true### API and Format

**API:** No breaking changes. This is purely additive functionality that removes an `UnsupportedOperationException` and enables existing MAP type infrastructure for INDEXED format.

**Storage Format:** No format changes. The MAP storage format in INDEXED rows was already defined and uses the same `BinaryMap` serialization mechanism as other formats. This change only enables reading MAP data that was already writeable.

### Documentation

No documentation changes required. This change brings INDEXED format to feature parity with ARROW and COMPACTED formats for MAP type support. MAP type usage is already documented in existing Fluss documentation. The change completes internal implementation of existing functionality rather than introducing new user-facing features.

### DIFF
--- a/fluss-common/src/main/java/org/apache/fluss/row/BinaryMap.java
+++ b/fluss-common/src/main/java/org/apache/fluss/row/BinaryMap.java
@@ -108,6 +108,7 @@ public abstract class BinaryMap extends BinarySection implements InternalMap {
         if (this == o) {
             return true;
         }
+        // override equals and only checks the other object is instance of BinaryMap
         if (!(o instanceof BinaryMap)) {
             return false;
         }
--- a/fluss-common/src/main/java/org/apache/fluss/row/InternalArray.java
+++ b/fluss-common/src/main/java/org/apache/fluss/row/InternalArray.java
@@ -150,8 +150,9 @@ public interface InternalArray extends DataGetters {
                                 fieldType.getTypeRoot().toString(), InternalArray.class.getName());
                 throw new IllegalArgumentException(msg);
         }
-        // Always wrap with null check for safety, even for non-nullable types
-        // This is necessary because Arrow format may contain nulls in practice
+        if (!fieldType.isNullable()) {
+            return elementGetter;
+        }
         return (array, pos) -> {
             if (array.isNullAt(pos)) {
                 return null;
--- a/fluss-common/src/main/java/org/apache/fluss/row/columnar/VectorizedColumnBatch.java
+++ b/fluss-common/src/main/java/org/apache/fluss/row/columnar/VectorizedColumnBatch.java
@@ -44,10 +44,6 @@ public class VectorizedColumnBatch {
         return columns.length;
     }
 
-    public ColumnVector getColumn(int colId) {
-        return columns[colId];
-    }
-
     public boolean isNullAt(int rowId, int colId) {
         return columns[colId].isNullAt(rowId);
     }
--- a/fluss-common/src/test/java/org/apache/fluss/row/BinaryWriterTest.java
+++ b/fluss-common/src/test/java/org/apache/fluss/row/BinaryWriterTest.java
@@ -129,7 +129,7 @@ public class BinaryWriterTest {
 
         BinaryWriter.ValueWriter setter =
                 BinaryWriter.createValueWriter(
-                        DataTypes.MAP(DataTypes.INT().copy(false), DataTypes.STRING()),
+                        DataTypes.MAP(DataTypes.INT(), DataTypes.STRING()),
                         BinaryRow.BinaryRowFormat.COMPACTED);
 
         GenericMap map1 =
--- a/fluss-common/src/test/java/org/apache/fluss/row/GenericMapTest.java
+++ b/fluss-common/src/test/java/org/apache/fluss/row/GenericMapTest.java
@@ -369,16 +369,4 @@ public class GenericMapTest {
 
         assertThat(genericMap.hashCode()).isEqualTo(0);
     }
-
-    @Test
-    public void testToString() {
-        Map<String, Integer> javaMap = new HashMap<>();
-        javaMap.put("key1", 1);
-
-        GenericMap genericMap = new GenericMap(javaMap);
-
-        String result = genericMap.toString();
-        assertThat(result).contains("GenericMap");
-        assertThat(result).contains("map=");
-    }
 }
--- a/fluss-common/src/test/java/org/apache/fluss/row/array/CompactedArrayTest.java
+++ b/fluss-common/src/test/java/org/apache/fluss/row/array/CompactedArrayTest.java
@@ -193,12 +193,4 @@ public class CompactedArrayTest {
         assertThat(compactedArray.getString(0)).isEqualTo(BinaryString.fromString("hello"));
         assertThat(compactedArray.getString(1)).isEqualTo(BinaryString.fromString("world"));
     }
-
-    @Test
-    public void testGetRowCachesNestedFields() {
-        RowType rowType = RowType.of(DataTypes.INT(), DataTypes.STRING());
-
-        CompactedArray compactedArray = new CompactedArray(rowType);
-        assertThat(compactedArray).isNotNull();
-    }
 }
--- a/fluss-common/src/test/java/org/apache/fluss/row/arrow/ArrowReaderWriterTest.java
+++ b/fluss-common/src/test/java/org/apache/fluss/row/arrow/ArrowReaderWriterTest.java
@@ -91,7 +91,7 @@ class ArrowReaderWriterTest {
                     DataTypes.ARRAY(DataTypes.INT()),
                     DataTypes.ARRAY(DataTypes.FLOAT().copy(false)), // vector embedding type
                     DataTypes.ARRAY(DataTypes.ARRAY(DataTypes.STRING())), // nested array
-                    DataTypes.MAP(DataTypes.INT().copy(false), DataTypes.STRING()),
+                    DataTypes.MAP(DataTypes.INT(), DataTypes.STRING()),
                     DataTypes.ROW(
                             DataTypes.FIELD("i", DataTypes.INT()),
                             DataTypes.FIELD("r", NESTED_DATA_TYPE),
@@ -350,9 +350,7 @@ class ArrowReaderWriterTest {
         RowType rowType =
                 DataTypes.ROW(
                         DataTypes.FIELD("id", DataTypes.INT()),
-                        DataTypes.FIELD(
-                                "map",
-                                DataTypes.MAP(DataTypes.INT().copy(false), DataTypes.STRING())));
+                        DataTypes.FIELD("map", DataTypes.MAP(DataTypes.INT(), DataTypes.INT())));
 
         try (BufferAllocator allocator = new RootAllocator(Long.MAX_VALUE);
                 VectorSchemaRoot root =
@@ -365,14 +363,14 @@ class ArrowReaderWriterTest {
             // Write 150 rows, each with a 10-entry map.
             // Total entries = 1500, exceeding INITIAL_CAPACITY (1024).
             // But row count (150) < 1024, so handleSafe would be false without the fix.
-            int numRows = 150;
+            int numRows = 200;
             int mapSize = 10;
             for (int i = 0; i < numRows; i++) {
                 Object[] mapEntries = new Object[mapSize * 2];
                 for (int j = 0; j < mapSize; j++) {
                     int key = i * mapSize + j;
                     mapEntries[j * 2] = key;
-                    mapEntries[j * 2 + 1] = fromString("value_" + key);
+                    mapEntries[j * 2 + 1] = key * 2;
                 }
                 GenericMap map = GenericMap.of(mapEntries);
                 writer.writeRow(GenericRow.of(i, map));
@@ -404,8 +402,7 @@ class ArrowReaderWriterTest {
                 for (int j = 0; j < mapSize; j++) {
                     int key = i * mapSize + j;
                     assertThat(row.getMap(1).keyArray().getInt(j)).isEqualTo(key);
-                    assertThat(row.getMap(1).valueArray().getString(j))
-                            .isEqualTo(fromString("value_" + key));
+                    assertThat(row.getMap(1).valueArray().getInt(j)).isEqualTo(key * 2);
                 }
             }
         }
--- a/fluss-common/src/test/java/org/apache/fluss/row/columnar/ColumnarMapTest.java
+++ b/fluss-common/src/test/java/org/apache/fluss/row/columnar/ColumnarMapTest.java
@@ -191,19 +191,6 @@ public class ColumnarMapTest {
     }
 
     @Test
-    public void testEqualsWithSameObject() {
-        Object[] keys = {BinaryString.fromString("key1")};
-        Object[] values = {1};
-
-        InternalArray keyArray = new GenericArray(keys);
-        InternalArray valueArray = new GenericArray(values);
-
-        ColumnarMap map = new ColumnarMap(keyArray, valueArray);
-
-        assertThat(map.equals(map)).isTrue();
-    }
-
-    @Test
     public void testEqualsWithEqualMap() {
         Object[] keys = {BinaryString.fromString("key1"), BinaryString.fromString("key2")};
         Object[] values = {1, 2};
@@ -297,21 +284,6 @@ public class ColumnarMapTest {
         ColumnarMap map2 = new ColumnarMap(keyArray2, valueArray2);
 
         assertThat(map1.hashCode()).isNotEqualTo(map2.hashCode());
-    }
-
-    @Test
-    public void testToString() {
-        Object[] keys = {BinaryString.fromString("k1"), BinaryString.fromString("k2")};
-        Object[] values = {100, 200};
-
-        InternalArray keyArray = new GenericArray(keys);
-        InternalArray valueArray = new GenericArray(values);
-        ColumnarMap map = new ColumnarMap(keyArray, valueArray);
-
-        String result = map.toString();
-        assertThat(result).contains("ColumnarMap");
-        assertThat(result).contains("keyArray");
-        assertThat(result).contains("valueArray");
     }
 
     private static class TestIntColumnVector implements IntColumnVector {
--- a/fluss-common/src/test/java/org/apache/fluss/row/serializer/MapSerializerTest.java
+++ b/fluss-common/src/test/java/org/apache/fluss/row/serializer/MapSerializerTest.java
@@ -30,8 +30,6 @@ import org.apache.fluss.types.DataTypes;
 
 import org.junit.jupiter.api.Test;
 
-import java.util.Map;
-
 import static org.assertj.core.api.Assertions.assertThat;
 
 /** Test for {@link MapSerializer}. */
@@ -193,90 +191,6 @@ public class MapSerializerTest {
         assertThat(binaryMap.size()).isEqualTo(2);
         assertThat(binaryMap.keyArray().getString(0)).isEqualTo(BinaryString.fromString("key1"));
         assertThat(binaryMap.valueArray().getInt(0)).isEqualTo(1);
-    }
-
-    @Test
-    public void testConvertToJavaMap() {
-        GenericMap genericMap =
-                GenericMap.of(
-                        1, BinaryString.fromString("one"),
-                        2, BinaryString.fromString("two"),
-                        3, BinaryString.fromString("three"));
-
-        Map<Object, Object> javaMap =
-                MapSerializer.convertToJavaMap(genericMap, DataTypes.INT(), DataTypes.STRING());
-
-        assertThat(javaMap).isNotNull();
-        assertThat(javaMap.size()).isEqualTo(3);
-        assertThat(javaMap.get(1)).isEqualTo(BinaryString.fromString("one"));
-        assertThat(javaMap.get(2)).isEqualTo(BinaryString.fromString("two"));
-        assertThat(javaMap.get(3)).isEqualTo(BinaryString.fromString("three"));
-    }
-
-    @Test
-    public void testConvertToJavaMapWithEmptyMap() {
-        GenericMap genericMap = GenericMap.of();
-
-        Map<Object, Object> javaMap =
-                MapSerializer.convertToJavaMap(genericMap, DataTypes.INT(), DataTypes.STRING());
-
-        assertThat(javaMap).isNotNull();
-        assertThat(javaMap.size()).isEqualTo(0);
-    }
-
-    @Test
-    public void testConvertToJavaMapWithNullValues() {
-        GenericMap genericMap =
-                GenericMap.of(
-                        1, BinaryString.fromString("one"),
-                        2, null,
-                        3, BinaryString.fromString("three"));
-
-        Map<Object, Object> javaMap =
-                MapSerializer.convertToJavaMap(genericMap, DataTypes.INT(), DataTypes.STRING());
-
-        assertThat(javaMap.size()).isEqualTo(3);
-        assertThat(javaMap.get(1)).isEqualTo(BinaryString.fromString("one"));
-        assertThat(javaMap.get(2)).isNull();
-        assertThat(javaMap.get(3)).isEqualTo(BinaryString.fromString("three"));
-    }
-
-    @Test
-    public void testConvertToJavaMapWithBinaryMap() {
-        MapSerializer serializer =
-                new MapSerializer(
-                        DataTypes.INT(), DataTypes.STRING(), BinaryRow.BinaryRowFormat.COMPACTED);
-
-        GenericMap genericMap =
-                GenericMap.of(
-                        1, BinaryString.fromString("alpha"),
-                        2, BinaryString.fromString("beta"));
-
-        BinaryMap binaryMap = serializer.toBinaryMap(genericMap);
-
-        Map<Object, Object> javaMap =
-                MapSerializer.convertToJavaMap(binaryMap, DataTypes.INT(), DataTypes.STRING());
-
-        assertThat(javaMap.size()).isEqualTo(2);
-        assertThat(javaMap.get(1)).isEqualTo(BinaryString.fromString("alpha"));
-        assertThat(javaMap.get(2)).isEqualTo(BinaryString.fromString("beta"));
-    }
-
-    @Test
-    public void testConvertToJavaMapWithComplexTypes() {
-        GenericMap genericMap =
-                GenericMap.of(
-                        100L, 3.14,
-                        200L, 6.28,
-                        300L, 9.42);
-
-        Map<Object, Object> javaMap =
-                MapSerializer.convertToJavaMap(genericMap, DataTypes.BIGINT(), DataTypes.DOUBLE());
-
-        assertThat(javaMap.size()).isEqualTo(3);
-        assertThat(javaMap.get(100L)).isEqualTo(3.14);
-        assertThat(javaMap.get(200L)).isEqualTo(6.28);
-        assertThat(javaMap.get(300L)).isEqualTo(9.42);
     }
 
     @Test
--- a/fluss-flink/fluss-flink-common/src/test/java/org/apache/fluss/flink/sink/FlinkComplexTypeITCase.java
+++ b/fluss-flink/fluss-flink-common/src/test/java/org/apache/fluss/flink/sink/FlinkComplexTypeITCase.java
@@ -45,9 +45,7 @@ import java.util.List;
 
 import static org.apache.fluss.flink.FlinkConnectorOptions.BOOTSTRAP_SERVERS;
 import static org.apache.fluss.flink.source.testutils.FlinkRowAssertionsUtils.assertResultsIgnoreOrder;
-import static org.apache.fluss.flink.source.testutils.FlinkRowAssertionsUtils.collectRowsWithTimeout;
 import static org.apache.fluss.server.testutils.FlussClusterExtension.BUILTIN_DATABASE;
-import static org.assertj.core.api.Assertions.assertThat;
 import static org.assertj.core.api.Assertions.assertThatThrownBy;
 
 /** Integration tests for Array type support in Flink connector. */
@@ -214,24 +212,23 @@ abstract class FlinkComplexTypeITCase extends AbstractTestBase {
 
         CloseableIterator<Row> rowIter =
                 tEnv.executeSql("select * from complex_log_test").collect();
-        List<String> actual = collectRowsWithTimeout(rowIter, 4, true);
-
-        assertThat(actual).hasSize(4);
-        assertThat(actual.get(0))
-                .contains("+I[1,", "[1, 2, null]", "{1=one, 2=two}", "+I[10, hello]");
-        assertThat(actual.get(1)).contains("+I[2,", "[400, 500]", "{3=three}", "+I[40, world]");
-
-        String row1 = actual.stream().filter(s -> s.startsWith("+I[1,")).findFirst().orElse("");
-        assertThat(row1).contains("10=v1", "20=v2");
-
-        String row2 = actual.stream().filter(s -> s.startsWith("+I[2,")).findFirst().orElse("");
-        assertThat(row2).contains("40=v4");
-
-        String row3 = actual.stream().filter(s -> s.startsWith("+I[3,")).findFirst().orElse("");
-        assertThat(row3).contains("{1=null}", "{k1={10=null}}", "{arr1=null}");
-
-        String row4 = actual.stream().filter(s -> s.startsWith("+I[4,")).findFirst().orElse("");
-        assertThat(row4).contains("{2=two, 3=null}", "{k2=null}", "{arr2=[1, 2], arr3=null}");
+        List<String> expectedRows =
+                Arrays.asList(
+                        "+I[1, [1, 2, null], [100, null, 300], [1.1, null], [2.2, 3.3, null], [a, null, c], [true, null, false], [[1, 2], null, [3]], [[x], [null, y]], [[[1, 2]], [[3, 4, 5]]], "
+                                + "{1=one, 2=two}, {k1={20=v2, 10=v1}}, {arr1=[1, 2, 3]}, {1=+I[100, row1]}, " // map
+                                + "+I[10, hello], +I[20, +I[30, nested], row1], [+I[1, a], +I[2, b]], " // row
+                                + "null, null, null]", // complex nested types
+                        "+I[2, null, [400, 500], [4.4], [5.5], [d, e], [true], [[6, 7, 8]], [[z]], [[[9]]], "
+                                + "{3=three}, {k3={40=v4}}, {arr3=[6]}, {3=+I[300, row3]}, "
+                                + "+I[40, world], +I[50, +I[60, test], row2], [+I[3, c]], "
+                                + "null, null, null]",
+                        "+I[3, null, null, null, null, null, null, null, null, null, "
+                                + "{1=null}, {k1={10=null}}, {arr1=null}, null, "
+                                + "null, null, null, null, null, null]",
+                        "+I[4, null, null, null, null, null, null, null, null, null, "
+                                + "{2=two, 3=null}, {k2=null}, {arr2=[1, 2], arr3=null}, null, "
+                                + "null, null, null, null, null, null]");
+        assertResultsIgnoreOrder(rowIter, expectedRows, true);
     }
 
     @Test
@@ -339,30 +336,17 @@ abstract class FlinkComplexTypeITCase extends AbstractTestBase {
 
         CloseableIterator<Row> rowIter =
                 tEnv.executeSql("select * from complex_log_test").collect();
-        List<String> actual = collectRowsWithTimeout(rowIter, 4, true);
-        assertThat(actual).hasSize(4);
-
-        assertThat(actual.stream().filter(s -> s.startsWith("+I[1, 2024,")).findFirst().orElse(""))
-                .contains(
-                        "[1, 2, null]",
-                        "1=one",
-                        "10=v1",
-                        "+I[10, hello]",
-                        "+I[20, +I[30, nested",
-                        "null, null, null]");
-        assertThat(actual.stream().filter(s -> s.startsWith("+I[2, 2023,")).findFirst().orElse(""))
-                .contains(
-                        "null, [400, 500]",
-                        "3=three",
-                        "20=v2",
-                        "+I[40, world]",
-                        "+I[50, +I[60, test",
-                        "null, null, null]");
-
-        assertThat(actual.stream().filter(s -> s.startsWith("+I[3, 2024,")).findFirst().orElse(""))
-                .contains("{1=null}", "{k1={10=null}}", "{arr1=null}");
-        assertThat(actual.stream().filter(s -> s.startsWith("+I[4, 2023,")).findFirst().orElse(""))
-                .contains("{2=two, 3=null}", "{k2=null}", "{arr2=[1, 2], arr3=null}");
+        List<String> expectedRows =
+                Arrays.asList(
+                        "+I[1, 2024, [1, 2, null], [100, null, 300], [1.1, null], [2.2, 3.3, null], [a, null, c], [true, null, false], [[1, 2], null, [3]], [[x], [null, y]], [[[1, 2]], [[3, 4, 5]]], "
+                                + "{1=one}, {k1={10=v1}}, {arr1=[1, 2]}, +I[10, hello], +I[20, +I[30, nested]], null, null, null]",
+                        "+I[2, 2023, null, [400, 500], [4.4], [5.5], [d, e], [true], [[6, 7, 8]], [[z]], [[[9]]], "
+                                + "{3=three}, {k2={20=v2}}, {arr2=[3, 4, 5]}, +I[40, world], +I[50, +I[60, test]], null, null, null]",
+                        "+I[3, 2024, null, null, null, null, null, null, null, null, null, "
+                                + "{1=null}, {k1={10=null}}, {arr1=null}, null, null, null, null, null]",
+                        "+I[4, 2023, null, null, null, null, null, null, null, null, null, "
+                                + "{2=two, 3=null}, {k2=null}, {arr2=[1, 2], arr3=null}, null, null, null, null, null]");
+        assertResultsIgnoreOrder(rowIter, expectedRows, true);
     }
 
     @Test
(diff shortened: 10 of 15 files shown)


fluss-client/src/test/java/org/apache/fluss/client/table/scanner/log/DefaultCompletedFetchTest.java
```diff
@@ -321,6 +356,18 @@ void testComplexTypeFetch() throws Exception {
...
+            GenericMap simpleMap = (GenericMap) row.getMap(4);
+            assertThat(simpleMap.size()).isGreaterThan(0);
+
+            assertThat(row.getMap(5)).isInstanceOf(GenericMap.class);
+            GenericMap nestedMap = (GenericMap) row.getMap(5);
+            assertThat(nestedMap.size()).isGreaterThan(0);
+
+            assertThat(row.getMap(6)).isInstanceOf(GenericMap.class);
+            GenericMap mapWithArray = (GenericMap) row.getMap(6);
+            assertThat(mapWithArray.size()).isGreaterThan(0);
```
wuchong: Actually, this approach doesn’t verify the actual field values, and it significantly inflates the line count, making the code harder to maintain.  

Moreover, we’ll soon need to add assertions for column projection here as well.

My suggestion: prepare a list of **expected rows** upfront and perform a direct assert equals comparison with the actual result. 

This keeps the test concise, expressive, and much easier to maintain.

fluss-common/src/main/java/org/apache/fluss/row/aligned/AlignedRow.java
```diff
@@ -395,7 +396,13 @@ public InternalArray getArray(int pos) {
                 segments, offset, offsetAndSize, new AlignedArray());
     }
 
-    // TODO: getMap() will be added in Issue #1973
+    @Override
+    public InternalMap getMap(int pos) {
+        assertIndexIsValid(pos);
+        int fieldOffset = getFieldOffset(pos);
+        final long offsetAndSize = BinarySegmentUtils.getLong(segments, fieldOffset);
```
wuchong: keep align with others to use `final long offsetAndSize = segments[0].getLong(fieldOffset)` as we don't support multi-segments yet.

fluss-common/src/main/java/org/apache/fluss/row/arrow/vectors/ArrowMapColumnVector.java
```diff
@@ -0,0 +1,86 @@
...
+        if (!inited) {
+            FieldVector mapVector = ((MapVector) vector).getDataVector();
+            StructVector structVector = (StructVector) mapVector;
+            FieldVector keyVector = structVector.getChildrenFromFields().get(0);
+            FieldVector valueVector = structVector.getChildrenFromFields().get(1);
+            this.keyColumnVector = ArrowUtils.createArrowColumnVector(keyVector, keyType);
+            this.valueColumnVector = ArrowUtils.createArrowColumnVector(valueVector, valueType);
+            inited = true;
+        }
+    }
```
wuchong: initialize in the constructor and remove this method.

fluss-common/src/main/java/org/apache/fluss/row/arrow/vectors/ArrowMapColumnVector.java
```diff
@@ -0,0 +1,86 @@
...
+
+    public ColumnVector getKeyColumnVector() {
+        init();
+        return keyColumnVector;
+    }
+
+    public ColumnVector getValueColumnVector() {
+        init();
+        return valueColumnVector;
+    }
```
wuchong: unsed, remove

fluss-common/src/main/java/org/apache/fluss/row/arrow/vectors/ArrowMapColumnVector.java
```diff
@@ -0,0 +1,86 @@
...
+            this.keyColumnVector = ArrowUtils.createArrowColumnVector(keyVector, keyType);
+            this.valueColumnVector = ArrowUtils.createArrowColumnVector(valueVector, valueType);
+            inited = true;
+        }
+    }
+
+    @Override
+    public InternalMap getMap(int i) {
+        init();
+        MapVector mapVector = (MapVector) vector;
```
wuchong: Direct store `vector` as `MapVector` in `ArrowMapColumnVector` to avoid casting per record.

fluss-common/src/main/java/org/apache/fluss/row/arrow/vectors/ArrowMapColumnVector.java
```diff
@@ -0,0 +1,86 @@
...
+        init();
+        return valueColumnVector;
+    }
+
+    @Override
+    public boolean isNullAt(int i) {
+        return vector.isNull(i);
+    }
+
+    public ColumnVector[] getChildren() {
```
wuchong: unused, remove

fluss-common/src/main/java/org/apache/fluss/row/arrow/writers/ArrowMapWriter.java
```diff
@@ -0,0 +1,61 @@
...
+        InternalArray valueArray = map.valueArray();
+
+        mapVector.startNewValue(rowIndex);
+        for (int i = 0; i < map.size(); i++) {
+            int fieldIndex = offset + i;
+            structVector.setIndexDefined(fieldIndex);
+            keyWriter.write(fieldIndex, keyArray, i, handleSafe);
+            valueWriter.write(fieldIndex, valueArray, i, handleSafe);
+        }
+        offset += map.size();
```
wuchong, line 63: we should override `reset()` method to reset `offset` and `keyWriter` and `valueWriter`. 

I also found that we forgot to add the `reset()` for the newly introduced `ArrowRowWriter`.

fluss-common/src/main/java/org/apache/fluss/row/columnar/ColumnarMap.java
```diff
@@ -0,0 +1,100 @@
...
+        this.valueArray = valueArray;
+    }
+
+    public ColumnarMap(
+            ColumnVector keyColumnVector,
+            ColumnVector valueColumnVector,
+            int offset,
+            int numElements) {
+        this.keyArray = new ColumnarArray(keyColumnVector, offset, numElements);
+        this.valueArray = new ColumnarArray(valueColumnVector, offset, numElements);
```
wuchong: better to invoke `this(..)` to apply the same validation.

```java
this(
                new ColumnarArray(keyColumnVector, offset, numElements),
                new ColumnarArray(valueColumnVector, offset, numElements));
```

fluss-common/src/main/java/org/apache/fluss/row/columnar/VectorizedColumnBatch.java
```diff
@@ -43,6 +43,10 @@ public int getFieldCount() {
         return columns.length;
     }
 
+    public ColumnVector getColumn(int colId) {
```
wuchong: It's very strange to expose the underlying `ColumnVector` directly. Why not introduce a `getMap` on the `VectorizedColumnBatch`, like other complex types `getArray`, `getRow`?

fluss-common/src/main/java/org/apache/fluss/row/columnar/ColumnarRow.java
```diff
@@ -137,7 +138,10 @@ public InternalArray getArray(int pos) {
         return vectorizedColumnBatch.getArray(rowId, pos);
     }
 
-    // TODO: getMap() will be added in Issue #1973
+    @Override
+    public InternalMap getMap(int pos) {
+        return ((MapColumnVector) vectorizedColumnBatch.getColumn(pos)).getMap(rowId);
```
wuchong: Why not implement this like `getArray()` that having a `getMap` method directly on the `VectorizedColumnBatch`?

fluss-common/src/main/java/org/apache/fluss/row/compacted/CompactedRowReader.java
```diff
@@ -325,7 +326,9 @@ static FieldReader createFieldReader(DataType fieldType) {
                 DataType elementType = ((ArrayType) fieldType).getElementType();
                 fieldReader = (reader, pos) -> reader.readArray(elementType);
                 break;
-
+            case MAP:
+                fieldReader = (reader, pos) -> reader.readMap();
```
wuchong: should pass `keyType` and `valueType` into it, otherwise, we can't support `Map<K, Row>` nested type.

fluss-common/src/main/java/org/apache/fluss/row/indexed/IndexedRowReader.java
```diff
@@ -286,15 +295,14 @@ static FieldReader createFieldReader(DataType fieldType) {
                 DataType elementType = ((ArrayType) fieldType).getElementType();
                 fieldReader = (reader, pos) -> reader.readArray(elementType);
                 break;
+            case MAP:
+                fieldReader = (reader, pos) -> reader.readMap();
```
wuchong: ditto

fluss-common/src/main/java/org/apache/fluss/row/BinaryMap.java
```diff
@@ -0,0 +1,166 @@
...
+ * <pre>
+ * [4 byte(keyArray size in bytes)] + [Key BinaryArray] + [Value BinaryArray].
+ * </pre>
+ *
+ * <p>Influenced by Apache Spark UnsafeMapData.
+ *
+ * @since 0.9
+ */
+@PublicEvolving
+public class BinaryMap extends BinarySection implements InternalMap {
```
wuchong: follow `BinaryArray`, mark this as `abstract` class, and have `PrimitiveBinaryMap` subclass.

fluss-common/src/main/java/org/apache/fluss/row/BinaryMap.java
```diff
@@ -0,0 +1,166 @@
...
+
+    /** Creates a {@link BinaryArray} instance for values with the nested data type information. */
+    protected BinaryArray createValueArrayInstance() {
+        return valueArraySupplier.get();
+    }
+
+    /** Creates a nested {@link BinaryMap} with the nested data type information. */
+    protected BinaryMap createNestedMapInstance() {
+        return nestedMapSupplier.get();
+    }
```
wuchong: Follow the pattern used in `BinaryArray` by declaring these as **abstract methods** and implementing them in subclasses.  

This approach eliminates the need for non-serializable fields like `keyArraySupplier`, which become `null` after deserialization and can lead to `NullPointerException` here at runtime.

fluss-common/src/main/java/org/apache/fluss/row/BinaryMap.java
```diff
@@ -0,0 +1,166 @@
...
+        return keyArraySupplier.get();
+    }
+
+    /** Creates a {@link BinaryArray} instance for values with the nested data type information. */
+    protected BinaryArray createValueArrayInstance() {
+        return valueArraySupplier.get();
+    }
+
+    /** Creates a nested {@link BinaryMap} with the nested data type information. */
+    protected BinaryMap createNestedMapInstance() {
```
wuchong: The nested instance has already been handled by `BinaryArray`, so we don't need nested map instance. I mean, we exposed `InternalArray#getArray` and `InternalArray#getRow` interface, so there is nested row instance and nested array instance needs for `BinaryArray`. But we don't expose such APIs for `InternalMap`, so we don't need nested xxx instance. 

Currently, this method is only used by `copy(...)` which actually is a create self map instance, so you should create another method for this purpose, like `createMapInstance`

fluss-common/src/main/java/org/apache/fluss/row/BinarySegmentUtils.java
```diff
@@ -1002,18 +1002,30 @@ public static BinaryArray readBinaryArray(
         return reusedArray;
     }
 
+    public static BinaryMap readBinaryMap(
+            MemorySegment[] segments, int baseOffset, long offsetAndSize) {
```
wuchong: we should do like `readBinaryArray` that pass in a reused `BinaryMap` instead hard-coding `new BinaryMap`.

fluss-common/src/main/java/org/apache/fluss/row/BinarySegmentUtils.java
```diff
@@ -1002,18 +1002,30 @@ public static BinaryArray readBinaryArray(
...
-        throw new UnsupportedOperationException(
-                "Map type is not supported yet. Will be added in Issue #1973.");
+        return readMapData(segments, offset, (int) numBytes);
+    }
+
+    /** Read map data from segments. */
+    public static BinaryMap readMapData(MemorySegment[] segments, int offset, int numBytes) {
+        BinaryMap map = new BinaryMap();
+        map.pointTo(segments, offset, numBytes);
+        return map;
```
wuchong: unused, remove

fluss-common/src/main/java/org/apache/fluss/row/InternalArray.java
```diff
@@ -184,6 +185,29 @@ static ElementGetter createDeepElementGetter(DataType fieldType) {
                             return new GenericArray(objs);
                         };
                 break;
+            case MAP:
+                DataType keyType = ((org.apache.fluss.types.MapType) fieldType).getKeyType();
+                DataType valueType = ((org.apache.fluss.types.MapType) fieldType).getValueType();
```
wuchong: replace qualified name with imports

fluss-common/src/main/java/org/apache/fluss/row/InternalArray.java
```diff
@@ -184,6 +185,29 @@ static ElementGetter createDeepElementGetter(DataType fieldType) {
...
+                            InternalMap inner = array.getMap(pos);
+                            InternalArray keys = inner.keyArray();
+                            InternalArray values = inner.valueArray();
+                            Object[] keyObjs = new Object[keys.size()];
+                            Object[] valueObjs = new Object[values.size()];
+                            for (int i = 0; i < keys.size(); i++) {
+                                keyObjs[i] = keyGetter.getElementOrNull(keys, i);
+                                valueObjs[i] = valueGetter.getElementOrNull(values, i);
+                            }
+                            java.util.Map<Object, Object> map = new java.util.HashMap<>();
```
wuchong: replace qualified name with imports

fluss-common/src/main/java/org/apache/fluss/row/InternalArray.java
```diff
@@ -184,6 +185,29 @@ static ElementGetter createDeepElementGetter(DataType fieldType) {
...
+                elementGetter =
+                        (array, pos) -> {
+                            InternalMap inner = array.getMap(pos);
+                            InternalArray keys = inner.keyArray();
+                            InternalArray values = inner.valueArray();
+                            Object[] keyObjs = new Object[keys.size()];
+                            Object[] valueObjs = new Object[values.size()];
+                            for (int i = 0; i < keys.size(); i++) {
+                                keyObjs[i] = keyGetter.getElementOrNull(keys, i);
+                                valueObjs[i] = valueGetter.getElementOrNull(values, i);
```
wuchong: we don't need the temporary `keyObjs` and `valueObjs`, directly put them into the map, this can reduce small object overhead.

fluss-common/src/main/java/org/apache/fluss/row/InternalRow.java
```diff
@@ -260,6 +263,28 @@ static FieldGetter createDeepFieldGetter(DataType fieldType, int fieldPos) {
...
+                fieldGetter =
+                        row -> {
+                            InternalMap map = row.getMap(fieldPos);
+                            Object[] keys = new Object[map.size()];
+                            Object[] values = new Object[map.size()];
+                            for (int i = 0; i < map.size(); i++) {
+                                keys[i] = keyGetter.getElementOrNull(map.keyArray(), i);
+                                values[i] = valueGetter.getElementOrNull(map.valueArray(), i);
+                            }
+                            java.util.Map<Object, Object> javaMap = new java.util.HashMap<>();
```
wuchong: ditto

fluss-common/src/main/java/org/apache/fluss/utils/json/DataTypeJsonSerde.java
```diff
@@ -285,7 +285,8 @@ private static DataType deserializeMap(JsonNode dataTypeNode) {
         final DataType keyType = DataTypeJsonSerde.INSTANCE.deserialize(keyNode);
         final JsonNode valueNode = dataTypeNode.get(FIELD_NAME_VALUE_TYPE);
         final DataType valueType = DataTypeJsonSerde.INSTANCE.deserialize(valueNode);
-        return new MapType(keyType, valueType);
+        final DataType nonNullableKeyType = keyType.isNullable() ? keyType.copy(false) : keyType;
```
wuchong: do not transform the data type during deserialize, this is very hack. If the key shouldn't be nullable, then this should be guaranteed by `MapType` constructor, not the type serialization/deserialition phase. Serialization/deserialition should just focus on how to read and write types from JSON.

fluss-common/src/main/java/org/apache/fluss/utils/ArrowUtils.java
```diff
@@ -339,6 +342,16 @@ public static ArrowFieldWriter createArrowFieldWriter(FieldVector vector, DataTy
             FieldVector elementFieldVector = ((ListVector) vector).getDataVector();
             return new ArrowArrayWriter(
                     vector, ArrowUtils.createArrowFieldWriter(elementFieldVector, elementType));
+        } else if (vector instanceof MapVector && dataType instanceof MapType) {
+            MapType mapType = (MapType) dataType;
+            MapVector mapVector = (MapVector) vector;
+            StructVector structVector = (StructVector) mapVector.getDataVector();
+            FieldVector keyVector = structVector.getChild("key");
+            FieldVector valueVector = structVector.getChild("value");
```
wuchong: use Arrow's `MapVector.KEY_NAME` and `MapVector.VALUE_NAME`.

fluss-common/src/main/java/org/apache/fluss/row/arrow/vectors/ArrowMapColumnVector.java
```diff
@@ -0,0 +1,86 @@
...
+        this.keyType = keyType;
+        this.valueType = valueType;
+    }
+
+    private void init() {
+        if (!inited) {
+            FieldVector mapVector = ((MapVector) vector).getDataVector();
+            StructVector structVector = (StructVector) mapVector;
+            FieldVector keyVector = structVector.getChildrenFromFields().get(0);
+            FieldVector valueVector = structVector.getChildrenFromFields().get(1);
```
wuchong: Instead of direct access, you should use  
```java
structVector.getChild(MapVector.KEY_NAME)
```  
and  
```java
structVector.getChild(MapVector.VALUE_NAME)
```  
to obtain the `keyVector` and `valueVector`, respectively.

Additionally, to maintain consistency with `ArrowArrayColumnVector` and `ArrowRowColumnVector`, it’s better to move the initialization of `keyColumnVector` and `valueColumnVector` **outside** of `ArrowMapColumnVector`.  

This also allows us to avoid exposing `ArrowUtils.createArrowColumnVector` as `public`, preserving better encapsulation and alignment with existing patterns.

fluss-common/src/main/java/org/apache/fluss/utils/ArrowUtils.java
```diff
@@ -434,9 +450,24 @@ private static Field toArrowField(String fieldName, DataType logicalType) {
...
+                                "Map key type must be non-nullable for Arrow conversion, but got: %s",
+                                keyType.asSummaryString()));
+            }
+            Field keyField = toArrowField("key", keyType);
+            Field valueField = toArrowField("value", mapType.getValueType());
+            FieldType structFieldType = new FieldType(false, ArrowType.Struct.INSTANCE, null);
+            List<Field> structChildren = new ArrayList<>();
+            structChildren.add(keyField);
+            structChildren.add(valueField);
+            Field structField = new Field("entries", structFieldType, structChildren);
```
wuchong: Use `MapVector.KEY_NAME`, `MapVector.VALUE_NAME` and `MapVector.DATA_VECTOR_NAME` to replace the hard code string names ("key", "value", "entires").

fluss-common/src/main/java/org/apache/fluss/utils/ArrowUtils.java
```diff
@@ -434,9 +450,24 @@ private static Field toArrowField(String fieldName, DataType logicalType) {
             for (DataField field : rowType.getFields()) {
                 children.add(toArrowField(field.getName(), field.getType()));
             }
+        } else if (logicalType instanceof MapType) {
+            MapType mapType = (MapType) logicalType;
+            DataType keyType = mapType.getKeyType();
+            if (keyType.isNullable()) {
+                throw new IllegalArgumentException(
```
wuchong: If we enforce non-null keys at this point, I suggest automatically converting the key type to non-nullable in the `MapType` constructor. That means we treat map keys as non-nullable by default, even if the user specifies a nullable type.

Without this conversion, users would be forced to explicitly declare non-nullable keys (e.g., `MAP<INT NOT NULL, STRING>`), which is not user-friendly. 

For example, the following standard-looking DDL would fail:
```sql
CREATE TABLE xxx (
  f1 MAP<INT, STRING>
);
```

By handling the nullability conversion internally, we align with user expectations while maintaining the required constraint that map keys must not be null. This conversion is similar to how we automantically convert data types of primary key fields to not-null.

fluss-common/src/test/java/org/apache/fluss/row/columnar/ColumnarMapTest.java
```diff
@@ -0,0 +1,338 @@
...
+        assertThat(map.valueArray().getDouble(2)).isEqualTo(3.3);
+    }
+
+    private static class TestIntColumnVector implements IntColumnVector {
+        private final int[] values;
+        private final boolean[] nulls;
+
+        TestIntColumnVector(int[] values) {
+            this.values = values;
+            this.nulls = new boolean[values.length];
```
wuchong: as this is a primitive array, all elements should be not-null, rather than null.

fluss-common/src/test/java/org/apache/fluss/row/columnar/ColumnarMapTest.java
```diff
@@ -0,0 +1,338 @@
...
+        }
+    }
+
+    private static class TestLongColumnVector implements LongColumnVector {
+        private final long[] values;
+        private final boolean[] nulls;
+
+        TestLongColumnVector(long[] values) {
+            this.values = values;
+            this.nulls = new boolean[values.length];
```
wuchong: ditto

fluss-common/src/test/java/org/apache/fluss/row/columnar/ColumnarMapTest.java
```diff
@@ -0,0 +1,338 @@
...
+        }
+    }
+
+    private static class TestDoubleColumnVector implements DoubleColumnVector {
+        private final double[] values;
+        private final boolean[] nulls;
+
+        TestDoubleColumnVector(double[] values) {
+            this.values = values;
+            this.nulls = new boolean[values.length];
```
wuchong: ditto

fluss-common/src/test/java/org/apache/fluss/row/GenericMapTest.java
```diff
@@ -0,0 +1,357 @@
...
+        GenericMap genericMap = new GenericMap(javaMap);
+
+        assertThat(genericMap.get("a")).isEqualTo(1);
+        assertThat(genericMap.get("b")).isEqualTo(2);
+        assertThat(genericMap.get("c")).isEqualTo(3);
+        assertThat(genericMap.get("d")).isNull();
+    }
+
+    @Test
+    public void testSize() {
```
wuchong: remove, as this has been tested in other tests

fluss-common/src/test/java/org/apache/fluss/row/GenericMapTest.java
```diff
@@ -0,0 +1,357 @@
...
+        GenericMap genericMap = new GenericMap(javaMap);
+
+        assertThat(genericMap.size()).isEqualTo(3);
+        assertThat(genericMap.get("key1")).isEqualTo(1);
+        assertThat(genericMap.get("key2")).isNull();
+        assertThat(genericMap.get("key3")).isEqualTo(3);
+    }
+
+    @Test
+    public void testEquals() {
```
wuchong: remove all the tests about `equals`, `toString` and `hashcode`. Tests don't need to cover these logics. If we want to improve coverage, they should has already been invoked in other tests and thus has test coverage.

fluss-common/src/test/java/org/apache/fluss/row/columnar/ColumnarMapTest.java
```diff
@@ -0,0 +1,338 @@
...
+        InternalArray valueArray2 = new GenericArray(values2);
+        ColumnarMap map2 = new ColumnarMap(keyArray2, valueArray2);
+
+        assertThat(map1.equals(map1)).isTrue();
+        assertThat(map1.equals(map2)).isTrue();
+        assertThat(map2.equals(map1)).isTrue();
+    }
+
+    @Test
+    public void testNotEquals() {
```
wuchong: remove all the tests about `equals`, `toString` and `hashcode`. Tests don't need to cover these logics. If we want to improve coverage, they should has already been invoked in other tests and thus has test coverage.

fluss-common/src/test/java/org/apache/fluss/row/indexed/IndexedRowReaderTest.java
```diff
@@ -120,8 +120,9 @@ private void assertAllTypeEqualsForReader(IndexedRowReader reader) {
                                 GenericArray.of(fromString("a"), null, fromString("c")),
                                 null,
                                 GenericArray.of(fromString("hello"), fromString("world"))));
+        assertThat(reader.readMap().size()).isEqualTo(3);
```
wuchong: assert the map values, rather than just assert size

fluss-common/src/test/java/org/apache/fluss/row/map/IndexedMapTest.java
```diff
@@ -0,0 +1,323 @@
...
+
+        IndexedMap map = new IndexedMap(keyType, valueType);
+
+        GenericArray keyArray = new GenericArray(new Object[] {1, 2, 3});
+        GenericArray valueArray =
+                new GenericArray(
+                        new Object[] {
+                            org.apache.fluss.row.BinaryString.fromString("a"),
+                            org.apache.fluss.row.BinaryString.fromString("b"),
+                            org.apache.fluss.row.BinaryString.fromString("c")
```
wuchong: replace qualified names with static imporsts

fluss-common/src/test/java/org/apache/fluss/row/map/IndexedMapTest.java
```diff
@@ -0,0 +1,323 @@
...
+import org.apache.fluss.types.DataTypes;
+import org.apache.fluss.types.MapType;
+
+import org.junit.jupiter.api.Test;
+
+import static org.apache.fluss.row.BinaryRow.BinaryRowFormat.INDEXED;
+import static org.assertj.core.api.Assertions.assertThat;
+
+/** Test for {@link IndexedMap}. */
+public class IndexedMapTest {
```
wuchong: Could you please remove `IndexedMapTest` and `CompactedMapTest`?  

These tests provide little to no real value, they don’t verify any behavior or assert expected values; they only check that the constructed map is not `null`. This doesn’t meaningfully validate correctness or prevent regressions.

If the goal is to improve test coverage, those code paths are already covered by other, more substantive tests. And if certain methods truly remain uncovered, we should add **focused, purposeful tests** for those specific cases, rather than relying on AI-generated boilerplate that inflates the test suite without adding confidence in the code.

fluss-common/src/test/java/org/apache/fluss/row/serializer/MapSerializerTest.java
```diff
@@ -0,0 +1,442 @@
...
+import java.util.LinkedHashMap;
+import java.util.Map;
+
+import static org.apache.fluss.row.BinaryRow.BinaryRowFormat.ALIGNED;
+import static org.apache.fluss.row.BinaryRow.BinaryRowFormat.COMPACTED;
+import static org.apache.fluss.row.BinaryRow.BinaryRowFormat.INDEXED;
+import static org.assertj.core.api.Assertions.assertThat;
+
+/** Test for {@link MapSerializer}. */
+public class MapSerializerTest {
```
wuchong, line 36: remove this as well! See the comments on `IndexedMapTest`.

fluss-common/src/test/java/org/apache/fluss/row/BinaryWriterTest.java
```diff
@@ -124,13 +123,27 @@ public void testValueSetterWithDecimalType() {
...
+        GenericMap map2 =
+                GenericMap.of(
+                        3, BinaryString.fromString("three"), 4, BinaryString.fromString("four"));
+
+        setter.writeValue(writer, 0, map1);
+        setter.writeValue(writer, 1, map2);
+        writer.complete();
+
+        assertThat(array.getMap(0).size()).isEqualTo(2);
+        assertThat(array.getMap(1).size()).isEqualTo(2);
```
wuchong, line 146: please assert keys and values, you can assert them by comparing arrays.

fluss-common/src/test/java/org/apache/fluss/row/BinaryWriterTest.java
```diff
@@ -124,13 +123,27 @@ public void testValueSetterWithDecimalType() {
...
-                .isInstanceOf(UnsupportedOperationException.class)
-                .hasMessageContaining("Map type is not supported yet");
+    public void testValueSetterWithMapType() {
+        BinaryArray array = new org.apache.fluss.row.array.PrimitiveBinaryArray();
+        BinaryArrayWriter writer = new BinaryArrayWriter(array, 2, 8);
+
+        BinaryWriter.ValueWriter setter =
+                BinaryWriter.createValueWriter(
+                        DataTypes.MAP(DataTypes.INT().copy(false), DataTypes.STRING()),
+                        BinaryRow.BinaryRowFormat.ALIGNED);
```
wuchong: Use `COMPACTED` format in tests by default as it is more widely used than `ALIGNED` and `INDEXED`.

fluss-common/src/test/java/org/apache/fluss/row/columnar/ColumnarMapTest.java
```diff
@@ -0,0 +1,338 @@
...
+        InternalArray keyArray = new GenericArray(keys);
+        InternalArray valueArray = new GenericArray(values);
+
+        assertThatThrownBy(() -> new ColumnarMap(keyArray, valueArray))
+                .isInstanceOf(IllegalArgumentException.class)
+                .hasMessageContaining("Key array size and value array size must be equal");
+    }
+
+    @Test
+    public void testSize() {
```
wuchong: remove! as this has been tested in other test methods.

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/utils/FlussRowToFlinkRowConverter.java
```diff
@@ -162,8 +165,30 @@ static FlussDeserializationConverter createInternalConverter(DataType flussDataT
...
+                FlussDeserializationConverter keyConverter =
+                        createNullableInternalConverter(mapType.getKeyType());
+                FlussDeserializationConverter valueConverter =
+                        createNullableInternalConverter(mapType.getValueType());
+                return (flussField) -> {
+                    InternalMap flussMap = (InternalMap) flussField;
+                    InternalArray keyArray = flussMap.keyArray();
+                    InternalArray valueArray = flussMap.valueArray();
+                    int size = flussMap.size();
+                    java.util.Map<Object, Object> javaMap = new java.util.HashMap<>();
```
wuchong: ditto

fluss-flink/fluss-flink-common/src/test/java/org/apache/fluss/flink/sink/FlinkComplexTypeITCase.java
```diff
@@ -304,6 +304,108 @@ void testRowTypesInLogTable() throws Exception {
         assertResultsIgnoreOrder(rowIter, expectedRows, true);
     }
 
+    @Test
+    void testMapTypesInLogTable() throws Exception {
```
wuchong: I found that the current tests not cover all the cases, like missing PK table tests for Map and Row type, missing partitioned table test for Row type. 

Could you merge the map type tests and row type tests into the existing `testArrayTypesInPrimaryKeyTable`, `testArrayTypesInPartitionedLogTable`, and `testArrayTypesInLogTable`, and rename them to `testComplexTypesInPrimaryKeyTable`, `testComplexTypesInPartitionedLogTable`, and `testComplexTypesInLogTable`? This can also reduce the test time. 

Besides, could you add more nested type tests? Like 
- `array<map<string, double>>`
- `row<map<int, array<float>>`
- `map<bigint, row<string, array<string>, array<int>>`

fluss-flink/fluss-flink-common/src/test/java/org/apache/fluss/flink/utils/FlussRowToFlinkRowConverterTest.java
```diff
@@ -125,12 +125,16 @@ void testConverter() throws Exception {
             assertThat(stringArray2)
                     .isEqualTo(new BinaryString[] {fromString("hello"), fromString("world")});
 
+            // map
+            assertThat(flinkRow.getMap(22)).isNotNull();
+            assertThat(flinkRow.getMap(22).size()).isEqualTo(3);
```
wuchong: assert map values.

fluss-common/src/main/java/org/apache/fluss/row/arrow/writers/ArrowMapWriter.java
```diff
@@ -0,0 +1,61 @@
...
+        StructVector structVector = (StructVector) mapVector.getDataVector();
+
+        InternalArray keyArray = map.keyArray();
+        InternalArray valueArray = map.valueArray();
+
+        mapVector.startNewValue(rowIndex);
+        for (int i = 0; i < map.size(); i++) {
+            int fieldIndex = offset + i;
+            structVector.setIndexDefined(fieldIndex);
+            keyWriter.write(fieldIndex, keyArray, i, handleSafe);
```
platinumhamburg: Refer to issue: https://github.com/apache/fluss/issues/2164 — handleSafe cannot be directly passed through to lower layers, as it can lead to memory overflow issues.

platinumhamburg (changes requested):

wuchong:
Just a quick note: could you please avoid squashing commits while responding to review comments? Keeping the changes in separate commits makes it much easier to track what has been updated since the last review.

XuQianJin-Stars:
Please do not review this pull request until the CI  tests have passed.

XuQianJin-Stars:
@wuchong Hi, i already updated the pr. Please help review when you got some time.

platinumhamburg:
@XuQianJin-Stars I've noticed that the commit history has been erased again in several recent commits. For a very large PR, this indeed places a heavy burden on reviewers, as their review history and progress are lost, making incremental reviews impossible. Even when a force push is unavoidable due to rebasing, we should still strive to preserve the complete commit history and avoid manually squashing or reordering commits.

XuQianJin-Stars:
> @XuQianJin-Stars I've noticed that the commit history has been erased again in several recent commits. For a very large PR, this indeed places a heavy burden on reviewers, as their review history and progress are lost, making incremental reviews impossible. Even when a force push is unavoidable due to rebasing, we should still strive to preserve the complete commit history and avoid manually squashing or reordering commits.

This merge request was submitted because of file conflicts with the main branch. I just tried using `git reflog` to recover the changes, but there were quite a few modifications, so it's a bit troublesome.

fluss-common/src/test/java/org/apache/fluss/row/array/CompactedArrayTest.java
```diff
@@ -0,0 +1,204 @@
...
+        compactedArray.pointTo(
+                stringArray.getSegments(), stringArray.getOffset(), stringArray.getSizeInBytes());
+
+        assertThat(compactedArray.size()).isEqualTo(2);
+        assertThat(compactedArray.getString(0)).isEqualTo(BinaryString.fromString("hello"));
+        assertThat(compactedArray.getString(1)).isEqualTo(BinaryString.fromString("world"));
+    }
+
+    @Test
+    public void testGetRowCachesNestedFields() {
```
wuchong: I think this doesn't verify the nested fields cache.

fluss-common/src/test/java/org/apache/fluss/row/arrow/ArrowReaderWriterTest.java
```diff
@@ -328,4 +337,77 @@ void testArrayWriterWithManyElements() throws IOException {
...
+     */
+    @Test
+    void testMapWriterWithManyEntries() throws IOException {
+        // Schema with map column
+        RowType rowType =
+                DataTypes.ROW(
+                        DataTypes.FIELD("id", DataTypes.INT()),
+                        DataTypes.FIELD(
+                                "map",
+                                DataTypes.MAP(DataTypes.INT().copy(false), DataTypes.STRING())));
```
wuchong: Currently, this test can't reproduce the exception, we should use INT type to reproduce this problem.

fluss-lake/fluss-lake-iceberg/src/test/java/org/apache/fluss/lake/iceberg/flink/FlinkUnionReadPrimaryKeyTableITCase.java
```diff
@@ -80,7 +80,9 @@ void testUnionReadInStreamMode(Boolean isPartitioned) throws Exception {
         // will read iceberg snapshot, should only +I since no change log
         List<Row> expectedRows = new ArrayList<>();
         if (isPartitioned) {
-            for (String partition : waitUntilPartitions(t1).values()) {
+            List<String> sortedPartitions = new ArrayList<>(waitUntilPartitions(t1).values());
+            sortedPartitions.sort(String::compareTo);
```
wuchong: https://github.com/apache/fluss/pull/2295 should has already fix this ITCase, we don't need this change in this PR?

fluss-lake/fluss-lake-paimon/src/test/java/org/apache/fluss/lake/paimon/flink/FlinkUnionReadPrimaryKeyTableITCase.java
```diff
@@ -108,7 +108,7 @@ void testUnionReadFullType(Boolean isPartitioned) throws Exception {
         // will read paimon snapshot, won't merge log since it's empty
         List<String> resultEmptyLog =
                 toSortedRows(batchTEnv.executeSql("select * from " + tableName));
-        String expetedResultFromPaimon = buildExpectedResult(isPartitioned, partitions, 0, 1);
+        String expetedResultFromPaimon = buildExpectedResult(isPartitioned, 0, 1, t1);
```
wuchong: https://github.com/apache/fluss/pull/2295 should has already fix this ITCase, we don't need this change in this PR?

fluss-common/src/main/java/org/apache/fluss/row/InternalArray.java
```diff
@@ -133,22 +136,22 @@ static ElementGetter createElementGetter(DataType fieldType) {
...
                         String.format(
                                 "type %s not support in %s",
                                 fieldType.getTypeRoot().toString(), InternalArray.class.getName());
                 throw new IllegalArgumentException(msg);
         }
-        if (!fieldType.isNullable()) {
-            return elementGetter;
-        }
+        // Always wrap with null check for safety, even for non-nullable types
+        // This is necessary because Arrow format may contain nulls in practice
```
wuchong: Which case will fail because of this issue? This is very strange to me to remove this, as we don't have any problems before, and you didn't remove the null check branch for `InternalRow` which should also have problem if `InternalArray` has problem.

wuchong (approved):

XuQianJin-Stars:
> Thanks @XuQianJin-Stars , I left some final comments and pushed a commit to resolve them. Please take a look. If there is no other concerns, I will merge it once the CI is passed.

I reviewed the changes, and there are no problems.

PR #2197, opened 2025-12-18
[flink] Union read support read partition data for the partition exists in the lake but not in Fluss

LiebingYu:
### Purpose

Linked issue: close #2194

### DIFF
--- a/fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/source/enumerator/FlinkSourceEnumerator.java
+++ b/fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/source/enumerator/FlinkSourceEnumerator.java
@@ -24,6 +24,8 @@ import org.apache.fluss.client.metadata.KvSnapshots;
 import org.apache.fluss.config.ConfigOptions;
 import org.apache.fluss.config.Configuration;
 import org.apache.fluss.flink.lake.LakeSplitGenerator;
+import org.apache.fluss.flink.lake.split.LakeSnapshotAndFlussLogSplit;
+import org.apache.fluss.flink.lake.split.LakeSnapshotSplit;
 import org.apache.fluss.flink.source.enumerator.initializer.BucketOffsetsRetrieverImpl;
 import org.apache.fluss.flink.source.enumerator.initializer.NoStoppingOffsetsInitializer;
 import org.apache.fluss.flink.source.enumerator.initializer.OffsetsInitializer;
@@ -83,7 +85,8 @@ import static org.apache.fluss.utils.Preconditions.checkState;
  * <p>The enumerator is responsible for:
  *
  * <ul>
- *   <li>Get the all splits(snapshot split + log split) for a table of Fluss to be read.
+ *   <li>Get the all splits(lake split + kv snapshot split + log split) for a table of Fluss to be
+ *       read.
  *   <li>Assign the splits to readers with the guarantee that the splits belong to the same bucket
  *       will be assigned to same reader.
  * </ul>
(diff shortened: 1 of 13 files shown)


fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/source/enumerator/FlinkSourceEnumerator.java
```diff
@@ -83,7 +84,8 @@
  * <p>The enumerator is responsible for:
  *
  * <ul>
- *   <li>Get the all splits(snapshot split + log split) for a table of Fluss to be read.
+ *   <li>Get the all splits(lake split + snapshot split + log split) for a table of Fluss to be
```
luoyuxia:
```suggestion
 *   <li>Get the all splits(lake split + kv snapshot split + log split) for a table of Fluss to be
```

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/source/enumerator/FlinkSourceEnumerator.java
```diff
@@ -664,9 +677,16 @@ private void handlePartitionsRemoved(Collection<Partition> removedPartitionInfo)
...
                 (reader, splits) ->
                         splits.removeIf(
-                                split ->
-                                        removedPartitionsMap.containsKey(
-                                                split.getTableBucket().getPartitionId())));
+                                split -> {
+                                    // Never remove LakeSnapshotSplit, because during union reads,
+                                    // data from the lake partition must still be read even if the
+                                    // partition has already expired in Fluss.
+                                    if (split instanceof LakeSnapshotSplit) {
```
luoyuxia, line 685: What if `LakeSnapshotAndFlussLogSplit`? Such kind of split will also be removed.

fluss-lake/fluss-lake-paimon/src/test/java/org/apache/fluss/lake/paimon/flink/FlinkUnionReadPrimaryKeyTableITCase.java
```diff
@@ -763,6 +767,134 @@ void testUnionReadPrimaryKeyTableFailover(boolean isPartitioned) throws Exceptio
...
+    void testUnionReadPartitionsExistInPaimonButExpiredInFluss() throws Exception {
+        // first of all, start tiering
+        JobClient jobClient = buildTieringJob(execEnv);
+
+        String tableName = "expired_partition_pkTable";
+        TablePath tablePath = TablePath.of(DEFAULT_DB, tableName);
+        Map<TableBucket, Long> bucketLogEndOffset = new HashMap<>();
+        // create table & write initial data
+        long tableId =
+                preparePKTableFullType(tablePath, DEFAULT_BUCKET_NUM, true, bucketLogEndOffset);
```
luoyuxia: I think we can just use `prepareSimplePKTable` to simiplify the test instead of all datatypes.

luoyuxia (commented):
@LiebingYu Hi, thanks for the pr. Left minor comments

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/source/enumerator/FlinkSourceEnumerator.java
LiebingYu: As discuss offline, LakeSnapshotAndFlussLogSplit should not be considered here, as it pertains to how the reader handles partition expiration during the reading process—an issue that should be addressed through other mechanisms, such as a consumer. Introducing it here would add excessive complexity. Furthermore, even if it were included, we still couldn't guarantee complete data retrieval if the partition expires mid-read, because the data in Fluss would still be purged by TTL.

PR #2206, opened 2025-12-19
[kv] include hidden potential internal flush time in KV flush latency

zuston:
### Purpose

This change includes the hidden internal flush time during KV flushing.
`RocksDBWriteBatchWrapper` may trigger a flush implicitly when performing put or delete operations, which caused the previous latency measurement to be incomplete.

### Brief change log

Move startTime to the first line of the method.

### DIFF
--- a/fluss-server/src/main/java/org/apache/fluss/server/kv/KvTablet.java
+++ b/fluss-server/src/main/java/org/apache/fluss/server/kv/KvTablet.java
@@ -105,6 +105,7 @@ public final class KvTablet {
     private final long writeBatchSize;
     private final RocksDBKv rocksDBKv;
     private final KvPreWriteBuffer kvPreWriteBuffer;
+    private final TabletServerMetricGroup serverMetricGroup;
 
     // A lock that guards all modifications to the kv.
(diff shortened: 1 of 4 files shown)


zuston:
> A better approach would be to move the flushLatencyHistogram into RocksDBWriteBatchWrapper and update it directly within RocksDBWriteBatchWrapper#flush. This ensures accurate, per-flush latency tracking.

Make sense. Let me follow your idea. thanks @wuchong

wuchong:
Hi @zuston , Just checking in to see how things are going with this issue. No pressure at all — I'm just wondering if you're still able to work on it. If you're busy or need help, we'd be happy to assist or reassign it. Thanks again for taking it on!

zuston:
> Hi @zuston , Just checking in to see how things are going with this issue. No pressure at all — I'm just wondering if you're still able to work on it. If you're busy or need help, we'd be happy to assist or reassign it. Thanks again for taking it on!

I will continue to followup this and maybe later in this week

wuchong:
Rebased to trigger CI again.

Issue #228, opened 2024-12-19
[Feature] Support spark catalog(base spark 3.3.3)


chunxiaozheng:
### Search before asking

- [X] I searched in the [issues](https://github.com/alibaba/fluss/issues) and found nothing similar.

### Motivation

Support spark catalog(base spark 3.3.3)

### Willingness to contribute

- [X] I'm willing to submit a PR!

Issue #155, opened 2024-12-10
[Umbrella][Feature] Support spark engine


Alibaba-HZY:
### Search before asking

- [X] I searched in the [issues](https://github.com/alibaba/fluss/issues) and found nothing similar.

### Motivation

In roadmap,we plan to support the spark engine.
for kv table spark can support micro batch read, batch read, streaming writer and batch writer.
for log table spark can support micro batch read, streaming writer and batch writer.

According to @wuchong‘s suggestion,i will create some sub-tasks.
#### Task list

- [ ] spark catalog #228
- [ ] streaming read @Alibaba-HZY 
- [ ] batch read 
- [ ] union read 
- [ ] streaming write
- [ ] batch write

### Solution
In task "streaming read" I'll implement it and introduce some base classes
such as 
Introduce flussSparkTable implements org.apache.spark.sql.connector.catalog.Table
Introduce SparkCatalog implements org.apache.spark.sql.connector.catalog.TableCatalog
Introduce SparkInternalRow and SparkTypeUtils

### Willingness to contribute

- [X] I'm willing to submit a PR!

XuQianJin-Stars:
Which versions of Spark are supported?

wuchong:
Thanks @Alibaba-HZY for creating this issue! Could you create some sub-tasks that others can help together? 

Regarding the priorities, because of Fluss is a streaming storage, IMO, the priorities can be `streaming read > batch read > union read > streaming write > batch write`. 

Regarding the versions, let's start to support from Spark 3.5, and extend to Spark 3.4, 3.3 in the future.

Alibaba-HZY:
> Thanks @Alibaba-HZY for creating this issue! Could you create some sub-tasks that others can help together?
> 
> Regarding the priorities, because of Fluss is a streaming storage, IMO, the priorities can be `streaming read > batch read > union read > streaming write > batch write`.
> 
> Regarding the versions, let's start to support from Spark 3.5, and extend to Spark 3.4, 3.3 in the future.

ok done

Alibaba-HZY:
spark need use this api  https://github.com/alibaba/fluss/pull/341 to implement read and write

wuchong:
Phase 1 is Finished in v0.9

PR #1636, opened 2025-09-03
[spark] Support spark catalog and introduce some basic classes 

Alibaba-HZY:
Purpose
Linked issue: open https://github.com/apache/fluss/issues/228
Linked issue: open https://github.com/apache/fluss/issues/155

Support spark catalog and introduce some basic classes to support spark read and write

Brief change log
1.Based on https://github.com/apache/fluss/pull/258 and discussion with @XuQianJin-Stars , I resubmitted [support spakr catalog ] and introduce some basic classes(SparkInternalRow,SparkRow,FlussInputPartition) to support spark read and write.
2.support SupportsPartitionManagement in spark table

Tests
com.alibaba.fluss.connector.spark.catalog.SparkCatalogITCase
com.alibaba.fluss.connector.spark.utils.SparkConversionsTest
com.alibaba.fluss.connector.spark.utils.SparkTypeUtilsTest
com.alibaba.fluss.connector.spark.SparkInternalRowTest

API and Format
no

Documentation
no

PR #2219, opened 2025-12-21
[spark] initiate fluss-spark and introduce spark catalog and table

YannByron:
### Purpose

to introduce spark engine for https://github.com/apache/fluss/issues/155

This is the first pr, which includes:
1) to introduce the basic spark architecture, `fluss-spark-common`, `fluss-spark-ut` and `fluss-spark-3.x` included. 
2) to support spark catalog and table, based on spark 3.5 and 3.4 for now.
3) to support spark CI.

Linked issue: close https://github.com/apache/fluss/issues/228

### DIFF
--- a/fluss-spark/fluss-spark-common/src/main/scala/org/apache/fluss/spark/types/SparkToFlussTypeVisitor.scala
+++ b/fluss-spark/fluss-spark-common/src/main/scala/org/apache/fluss/spark/types/SparkToFlussTypeVisitor.scala
@@ -75,8 +75,8 @@ object SparkToFlussTypeVisitor {
         new DoubleType()
       case dt: org.apache.spark.sql.types.DecimalType =>
         new DecimalType(dt.precision, dt.scale)
-      case _: org.apache.spark.sql.types.BinaryType =>
-        new BinaryType(BinaryType.MAX_LENGTH)
+      case x: org.apache.spark.sql.types.BinaryType =>
+        new BytesType()
       case _: org.apache.spark.sql.types.VarcharType =>
         new StringType()
       case ct: org.apache.spark.sql.types.CharType =>


YannByron:
@wuchong please take a look, thanks.

fluss-spark/pom.xml
```diff
@@ -0,0 +1,340 @@
...
+
+                <plugin>
+                    <groupId>org.scalatest</groupId>
+                    <artifactId>scalatest-maven-plugin</artifactId>
+                    <version>2.1.0</version>
+                    <configuration>
+                        <reportsDirectory>${project.build.directory}/surefire-reports</reportsDirectory>
+                        <junitxml>.</junitxml>
+                        <argLine>-ea -Xmx4g -Xss4m -XX:MaxMetaspaceSize=2g -XX:ReservedCodeCacheSize=128m ${extraJavaTestArgs} -Dio.netty.tryReflectionSetAccessible=true</argLine>
+                        <filereports>PaimonTestSuite.txt</filereports>
```
wuchong: `FlussTestSuite.txt`?

fluss-spark/fluss-spark-common/pom.xml
```diff
@@ -0,0 +1,77 @@
...
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+    <parent>
+        <groupId>org.apache.fluss</groupId>
+        <artifactId>fluss-spark</artifactId>
+        <version>0.9-SNAPSHOT</version>
+    </parent>
+
+    <artifactId>fluss-spark-common</artifactId>
```
wuchong: Should we add a Scala version suffix to the artifact ID (also for `fluss-spark-3.4` and `fluss-spark-3.5` modules)? This would ensure that the published JARs automatically include the Scala version in their artifact names during Maven deployment, following standard Scala cross-build conventions.

pom.xml
```diff
@@ -458,6 +464,20 @@
...
+                <module>fluss-spark/fluss-spark-3.5</module>
+                <module>fluss-spark/fluss-spark-3.4</module>
+            </modules>
+            <activation>
+                <activeByDefault>true</activeByDefault>
+                <property>
+                    <name>spark3</name>
+                </property>
+            </activation>
+        </profile>
```
wuchong: I think we can enable these modues by default?  So that the license checker pipeline can verify these modules as well.

fluss-spark/fluss-spark-common/src/main/scala/org/apache/fluss/spark/FlussCatalog.scala
```diff
@@ -0,0 +1,108 @@
...
+import org.apache.spark.sql.connector.expressions.Transform
+import org.apache.spark.sql.types.StructType
+import org.apache.spark.sql.util.CaseInsensitiveStringMap
+
+import java.util
+import java.util.concurrent.ExecutionException
+
+import scala.collection.JavaConverters._
+
+class FlussCatalog extends TableCatalog with SupportsFlussNamespaces with WithFlussAdmin {
```
wuchong: How about naming it `SparkCatalog`? Since these catalog implementations reside in the Fluss repository alongside those for other engines (such as Flink and Trino), including the engine name in the class name would make it easier to identify and distinguish between them.

fluss-spark/fluss-spark-common/src/main/scala/org/apache/fluss/spark/FlussTable.scala
```diff
@@ -0,0 +1,25 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.fluss.spark
+
+import org.apache.fluss.metadata.TableInfo
+import org.apache.fluss.spark.catalog.{FlussTableInfo, SupportsFlussPartitionManagement}
+
+case class FlussTable(table: TableInfo)
```
wuchong: ditto. `SparkTable`?

fluss-spark/fluss-spark-common/src/main/scala/org/apache/fluss/spark/catalog/FlussTableInfo.scala
```diff
@@ -0,0 +1,42 @@
...
+import org.apache.fluss.metadata.TableInfo
+import org.apache.fluss.spark.SparkConversions
+import org.apache.spark.sql.connector.catalog.{Table, TableCapability}
+import org.apache.spark.sql.types.StructType
+
+import java.util
+
+import scala.collection.JavaConverters._
+
+abstract class FlussTableInfo(tableInfo: TableInfo) extends Table {
```
wuchong: ditto, `AbstractSparkTable`?

fluss-spark/fluss-spark-common/src/main/scala/org/apache/fluss/spark/catalog/SupportsFlussPartitionManagement.scala
```diff
@@ -0,0 +1,53 @@
...
+import org.apache.spark.sql.types.StructType
+
+import java.util
+
+trait SupportsFlussPartitionManagement extends FlussTableInfo with SupportsPartitionManagement {
+
+  override def partitionSchema(): StructType = _partitionSchema
+
+  override def createPartition(ident: InternalRow, properties: util.Map[String, String]): Unit = {
+    throw new UnsupportedOperationException("Creating partition is not supported")
```
wuchong, line 31: Will them be supported in follow-up PRs? It seems the original PR already supports this https://github.com/apache/fluss/pull/1636/files#diff-151d06dcc7dd4e7d0f9aae9be2f042e2588bb6d86820b57ec6fc2f8fef9f62cc

fluss-spark/fluss-spark-common/src/main/scala/org/apache/fluss/spark/types/SparkToFlussTypeVisitor.scala
```diff
@@ -0,0 +1,96 @@
...
+      case _: org.apache.spark.sql.types.LongType =>
+        new BigIntType()
+      case _: org.apache.spark.sql.types.FloatType =>
+        new FloatType()
+      case _: org.apache.spark.sql.types.DoubleType =>
+        new DoubleType()
+      case dt: org.apache.spark.sql.types.DecimalType =>
+        new DecimalType(dt.precision, dt.scale)
+      case _: org.apache.spark.sql.types.BinaryType =>
+        new BinaryType(BinaryType.MAX_LENGTH)
```
wuchong: use `new BytesType`

fluss-spark/fluss-spark-common/src/main/scala/org/apache/fluss/spark/types/SparkToFlussTypeVisitor.scala
```diff
@@ -0,0 +1,96 @@
...
+      case vct: org.apache.spark.sql.types.VarcharType =>
+        new CharType(vct.length)
```
wuchong: use `new StringType()`

fluss-spark/fluss-spark-common/src/main/scala/org/apache/fluss/spark/types/SparkToFlussTypeVisitor.scala
```diff
@@ -0,0 +1,96 @@
...
+      case ct: org.apache.spark.sql.types.CharType =>
+        new CharType(ct.length)
+      case _: org.apache.spark.sql.types.StringType =>
+        new StringType()
+      case _: org.apache.spark.sql.types.DateType =>
+        new DateType()
+      case _: org.apache.spark.sql.types.TimestampType =>
+        new LocalZonedTimestampType()
```
wuchong: Does spark only support precision=6 for timestamp? If yes, how about declaring the precision and add comment about this?

```java
// spark only support 6 digits of precision
return new LocalZonedTimestampType(6);
```

fluss-spark/fluss-spark-common/src/main/scala/org/apache/fluss/spark/types/SparkToFlussTypeVisitor.scala
```diff
@@ -0,0 +1,96 @@
...
+      case _: org.apache.spark.sql.types.TimestampNTZType =>
+        new TimestampType()
```
wuchong: ditto

fluss-spark/fluss-spark-common/src/main/scala/org/apache/fluss/spark/types/SparkToFlussTypeVisitor.scala
```diff
@@ -0,0 +1,96 @@
...
+    }
+    new RowType(flussDataFields.toList.asJava)
+  }
+
+  private def visitMapType(mt: SparkMapType): FlussMapType = {
+    new FlussMapType(visit(mt.keyType), visit(mt.valueType).copy(mt.valueContainsNull))
+  }
+
+  private def visitArrayType(at: SparkArrayType): FlussArrayType = {
+    new FlussArrayType(at.containsNull, visit(at.elementType))
```
wuchong: IIUC, `SparkArrayType.containsNull` means the array element can be null. However, the first boolean parameter of `new FlussArrayType` means the array instance can be null. If we want to set the element to nullable, we should 

```java
new FlussArrayType(visit(at.elementType).copy(at.containsNull))
```

fluss-spark/fluss-spark-common/src/main/scala/org/apache/fluss/spark/types/FlussDataTypeToSparkDataType.scala
```diff
@@ -0,0 +1,113 @@
...
+  override def visit(timestampType: TimestampType): SparkDataType = {
+    SparkDataTypes.TimestampNTZType
+  }
+
+  override def visit(localZonedTimestampType: LocalZonedTimestampType): SparkDataType = {
+    SparkDataTypes.TimestampType
+  }
+
+  override def visit(arrayType: ArrayType): SparkDataType = {
+    SparkDataTypes.createArrayType(arrayType.getElementType.accept(this), arrayType.isNullable)
```
wuchong: ditto, should be 

```scala
SparkDataTypes.createArrayType(elementType.accept(this), elementType.isNullable)
```

fluss-spark/fluss-spark-common/src/main/scala/org/apache/fluss/spark/types/FlussDataTypeToSparkDataType.scala
```diff
@@ -0,0 +1,113 @@
...
+  }
+
+  override def visit(mapType: MapType): SparkDataType = {
+    SparkDataTypes.createMapType(
+      mapType.getKeyType.accept(this),
+      mapType.getValueType.accept(this),
+      mapType.isNullable
```
wuchong: `mapType.getValueType.isNullable`

fluss-spark/fluss-spark-common/src/main/scala/org/apache/fluss/spark/types/FlussDataTypeToSparkDataType.scala
```diff
@@ -0,0 +1,113 @@
...
+ */
+
+package org.apache.fluss.spark.types
+
+import org.apache.fluss.types._
+import org.apache.spark.sql.types.{DataType => SparkDataType, DataTypes => SparkDataTypes}
+
+import scala.collection.JavaConverters._
+
+object FlussDataTypeToSparkDataType extends DataTypeVisitor[SparkDataType] {
```
wuchong: `FlussDataTypeToSparkDataType` -> `FlussToSparkTypeVisitor` to align with `SparkToFlussTypeVisitor`.

fluss-spark/fluss-spark-common/src/main/scala/org/apache/fluss/spark/SparkConversions.scala
```diff
@@ -0,0 +1,99 @@
...
+      tableDescriptorBuilder.distributedBy(bucketNum, bucketKeys: _*)
+    }
+
+    if (caseInsensitiveProps.contains(COMMENT.key)) {
+      tableDescriptorBuilder.comment(caseInsensitiveProps.get(COMMENT.key).get)
+    }
+
+    val (tableProps, customProps) =
+      caseInsensitiveProps.filterNot(SPARK_TABLE_OPTIONS.contains).partition {
+        case (key, _) => FlussConfigUtils.TABLE_OPTIONS.containsKey(key)
```
wuchong: `FlussConfigUtils.TABLE_OPTIONS` is a static set, however, the fluss table created by newer client version may carry additional table options. Therefore, it would be more robust to check whether the config key start with `table.` prefix.

.scalafmt.conf
```diff
@@ -0,0 +1,70 @@
...
+indent.extendSite = 2 # This makes sure extend is not indented as the ctor parameters
+
+# Rewrites
+rewrite.rules = [AvoidInfix, Imports, RedundantBraces, SortModifiers]
+
+# Imports
+rewrite.imports.sort = scalastyle
+rewrite.imports.groups = [
+    ["org.apache.paimon\\..*"],
+    ["org.apache.paimon.shade\\..*"],
```
wuchong: change to fluss package

fluss-spark/fluss-spark-ut/src/test/scala/org/apache/fluss/spark/FlussCatalogTest.scala
```diff
@@ -0,0 +1,138 @@
...
+      admin.getTableInfo(TablePath.of(DEFAULT_DATABASE, "test_pt_tbl")).get()
+    assertThat(testPartitionedTable.getRowType).isEqualTo(
+      RowType
+        .builder()
+        .field("id", DataTypes.INT())
+        .field("name", DataTypes.STRING())
+        .field("pt", DataTypes.STRING())
+        .build())
+    assertThat(testPartitionedTable.getPartitionKeys.get(0)).isEqualTo("pt")
+    assertThat(testPartitionedTable.getCustomProperties.containsKey("key")).isEqualTo(true)
```
wuchong, line 75: What is the `key` to verify?

fluss-spark/fluss-spark-common/pom.xml
YannByron: Sure. I'm just going to do this until scala 2.13 or spark4 is supported. Let me do this in this pr.

fluss-spark/fluss-spark-common/src/main/scala/org/apache/fluss/spark/FlussCatalog.scala
YannByron: ok

fluss-spark/fluss-spark-common/src/main/scala/org/apache/fluss/spark/catalog/SupportsFlussPartitionManagement.scala
YannByron: yes, in follow-up RPs.

fluss-spark/fluss-spark-common/src/main/scala/org/apache/fluss/spark/types/SparkToFlussTypeVisitor.scala
YannByron: ok, spark time timestamp and timestampntz type doesn't support self-specified  precision. let comment about this.

fluss-spark/fluss-spark-common/src/main/scala/org/apache/fluss/spark/types/SparkToFlussTypeVisitor.scala
YannByron: Right. I observe `MapType`, but ignore `ArrayType`.Thanks to correct it.

fluss-spark/fluss-spark-common/src/main/scala/org/apache/fluss/spark/types/FlussDataTypeToSparkDataType.scala
YannByron: OK. And the other sub-classes of `DataTypeVisitor ` can also follow this rule.

wuchong (approved):
LGTM. I rebased the branch to trigger test on latest `main` and appended a commit to mapping Spark BinaryType to Fluss BytesType, because the `BinaryType` in Fluss is a fixed size binary not a `VarbinaryType`.

PR #2223, opened 2025-12-22
[lake] Record a file path storing log offsets in lake snapshot property 

### LINKED ISSUE #2224: Record a file path storing log offsets in lake snapshot property
### Search before asking

### Description

No to store log offsets into lake snapshot directly which will cause snapshot very large.
We should create a file to store log offsets and then record the file path in snapshot path.

### Willingness to contribute

luoyuxia:
### Purpose

Linked issue: close #2224 

### Brief change log
- In `TieringCommitOperator`, first prepare commit log offsets to fluss cluster which will write a file to store the log offsets
- then, record the file path in lake snapshot property, commit the snapshot

### Tests

Existing test

### DIFF
--- a/fluss-rpc/src/main/java/org/apache/fluss/rpc/gateway/CoordinatorGateway.java
+++ b/fluss-rpc/src/main/java/org/apache/fluss/rpc/gateway/CoordinatorGateway.java
@@ -70,7 +70,7 @@ public interface CoordinatorGateway extends RpcGateway, AdminGateway {
 
     /**
      * Prepares lake table snapshots by merging them with existing snapshots and storing them to the
-     * file system.
+     * remote file system.
      *
      * <p>This method is called during the two-phase commit process for lake table snapshots. It
      * performs the following operations for each table in the request:
@@ -81,6 +81,8 @@ public interface CoordinatorGateway extends RpcGateway, AdminGateway {
      *   <li>Stores the merged snapshot to the remote file system. The stored file contains the log
      *       end offset information for each bucket in the table
      *   <li>Returns the file path where the snapshot is stored
+     *   <li>Call {@link #commitLakeTableSnapshot(CommitLakeTableSnapshotRequest)} with the offset
+     *       file path to finalize the snapshot commit to ZooKeeper in the second phase.
      * </ul>
      *
      * @param request the request containing lake table snapshot information for one or more tables
@@ -88,7 +90,7 @@ public interface CoordinatorGateway extends RpcGateway, AdminGateway {
      *     (containing bucket log end offset information) are stored, or error information for
      *     tables that failed to process
      */
-    @RPC(api = ApiKeys.PRE_LAKE_TABLE_SNAPSHOT)
+    @RPC(api = ApiKeys.PREPARE_LAKE_TABLE_SNAPSHOT)
     CompletableFuture<PrepareLakeTableSnapshotResponse> prepareLakeTableSnapshot(
             PrepareLakeTableSnapshotRequest request);
 
--- a/fluss-rpc/src/main/java/org/apache/fluss/rpc/protocol/ApiKeys.java
+++ b/fluss-rpc/src/main/java/org/apache/fluss/rpc/protocol/ApiKeys.java
@@ -80,7 +80,7 @@ public enum ApiKeys {
     REBALANCE(1049, 0, 0, PUBLIC),
     LIST_REBALANCE_PROGRESS(1050, 0, 0, PUBLIC),
     CANCEL_REBALANCE(1051, 0, 0, PUBLIC),
-    PRE_LAKE_TABLE_SNAPSHOT(1052, 0, 0, PRIVATE);
+    PREPARE_LAKE_TABLE_SNAPSHOT(1052, 0, 0, PRIVATE);
 
     private static final Map<Integer, ApiKeys> ID_TO_TYPE =
             Arrays.stream(ApiKeys.values())
--- a/fluss-rpc/src/main/proto/FlussApi.proto
+++ b/fluss-rpc/src/main/proto/FlussApi.proto
@@ -1046,7 +1046,7 @@ message PbPrepareLakeTableRespForTable {
   optional int32 error_code = 1;
   optional string error_message = 2;
   optional int64 table_id = 3;
-  optional string lake_table_bucket_offsets_path = 4;
+  optional string lake_table_offsets_path = 4;
 }
 
 message PbTableOffsets {
--- a/fluss-server/src/main/java/org/apache/fluss/server/coordinator/CoordinatorEventProcessor.java
+++ b/fluss-server/src/main/java/org/apache/fluss/server/coordinator/CoordinatorEventProcessor.java
@@ -1150,12 +1150,15 @@ public class CoordinatorEventProcessor implements EventProcessor {
 
     private void processNotifyLakeTableOffsetEvent(NotifyLakeTableOffsetEvent event) {
         Map<Long, LakeTableSnapshot> lakeTableSnapshots = event.getLakeTableSnapshots();
-        Map<TableBucket, Long> tableBucketMaxTieredTimestamps =
-                event.getTableBucketMaxTieredTimestamps();
+        Map<Long, Map<TableBucket, Long>> tableMaxTieredTimestamps =
+                event.getTableMaxTieredTimestamps();
         coordinatorRequestBatch.newBatch();
         for (Map.Entry<Long, LakeTableSnapshot> lakeTableSnapshotEntry :
                 lakeTableSnapshots.entrySet()) {
             LakeTableSnapshot lakeTableSnapshot = lakeTableSnapshotEntry.getValue();
+            Map<TableBucket, Long> tableBucketMaxTieredTimestamps =
+                    tableMaxTieredTimestamps.getOrDefault(
+                            lakeTableSnapshotEntry.getKey(), Collections.emptyMap());
             for (TableBucket tb : lakeTableSnapshot.getBucketLogEndOffset().keySet()) {
                 coordinatorContext
                         .getBucketLeaderAndIsr(tb)
@@ -1257,6 +1260,7 @@ public class CoordinatorEventProcessor implements EventProcessor {
                     try {
                         CommitLakeTableSnapshotResponse response =
                                 new CommitLakeTableSnapshotResponse();
+                        Set<Long> failedTableIds = new HashSet<>();
                         for (Map.Entry<Long, LakeTableSnapshot> lakeTableSnapshotEntry :
                                 lakeTableSnapshots.entrySet()) {
                             Long tableId = lakeTableSnapshotEntry.getKey();
@@ -1277,18 +1281,24 @@ public class CoordinatorEventProcessor implements EventProcessor {
                                 // this involves IO operation (ZK), so we do it in ioExecutor
                                 lakeTableHelper.registerLakeTableSnapshotV1(
                                         tableId, lakeTableSnapshotEntry.getValue());
-                                // send notify lakehouse data request to all replicas via
-                                // coordinator event
-                                coordinatorEventManager.put(
-                                        new NotifyLakeTableOffsetEvent(
-                                                lakeTableSnapshots,
-                                                commitLakeTableSnapshotData
-                                                        .getTableBucketsMaxTieredTimestamp()));
                             } catch (Exception e) {
+                                failedTableIds.add(tableId);
                                 ApiError error = ApiError.fromThrowable(e);
                                 tableResp.setError(error.error().code(), error.message());
                             }
                         }
+
+                        // remove failed tables
+                        Map<Long, LakeTableSnapshot> commitlakeTableSnapshots =
+                                commitLakeTableSnapshotData.getLakeTableSnapshot();
+                        commitlakeTableSnapshots.keySet().removeAll(failedTableIds);
+                        Map<Long, Map<TableBucket, Long>> tableMaxTieredTimestamps =
+                                commitLakeTableSnapshotData.getTableMaxTieredTimestamps();
+                        tableMaxTieredTimestamps.keySet().removeAll(failedTableIds);
+
+                        coordinatorEventManager.put(
+                                new NotifyLakeTableOffsetEvent(
+                                        commitlakeTableSnapshots, tableMaxTieredTimestamps));
                         callback.complete(response);
                     } catch (Exception e) {
                         callback.completeExceptionally(e);
@@ -1308,6 +1318,7 @@ public class CoordinatorEventProcessor implements EventProcessor {
                     try {
                         CommitLakeTableSnapshotResponse response =
                                 new CommitLakeTableSnapshotResponse();
+                        Set<Long> failedTableIds = new HashSet<>();
                         for (Map.Entry<Long, LakeTable.LakeSnapshotMetadata>
                                 lakeSnapshotMetadataEntry : lakeSnapshotMetadatas.entrySet()) {
                             PbCommitLakeTableSnapshotRespForTable tableResp =
@@ -1317,16 +1328,23 @@ public class CoordinatorEventProcessor implements EventProcessor {
                             try {
                                 lakeTableHelper.registerLakeTableSnapshotV2(
                                         tableId, lakeSnapshotMetadataEntry.getValue());
-                                coordinatorEventManager.put(
-                                        new NotifyLakeTableOffsetEvent(
-                                                commitLakeTableSnapshotData.getLakeTableSnapshot(),
-                                                commitLakeTableSnapshotData
-                                                        .getTableBucketsMaxTieredTimestamp()));
                             } catch (Exception e) {
+                                failedTableIds.add(tableId);
                                 ApiError error = ApiError.fromThrowable(e);
                                 tableResp.setError(error.error().code(), error.message());
                             }
                         }
+                        // remove failed tables
+                        Map<Long, LakeTableSnapshot> lakeTableSnapshots =
+                                commitLakeTableSnapshotData.getLakeTableSnapshot();
+                        lakeTableSnapshots.keySet().removeAll(failedTableIds);
+                        Map<Long, Map<TableBucket, Long>> tableMaxTieredTimestamps =
+                                commitLakeTableSnapshotData.getTableMaxTieredTimestamps();
+                        tableMaxTieredTimestamps.keySet().removeAll(failedTableIds);
+
+                        coordinatorEventManager.put(
+                                new NotifyLakeTableOffsetEvent(
+                                        lakeTableSnapshots, tableMaxTieredTimestamps));
                         callback.complete(response);
                     } catch (Exception e) {
                         callback.completeExceptionally(e);
--- a/fluss-server/src/main/java/org/apache/fluss/server/coordinator/CoordinatorService.java
+++ b/fluss-server/src/main/java/org/apache/fluss/server/coordinator/CoordinatorService.java
@@ -672,7 +672,7 @@ public final class CoordinatorService extends RpcServiceBase implements Coordina
                                         lakeTableHelper.storeLakeTableOffsetsFile(
                                                 tablePath, tableBucketOffsets);
                                 pbPrepareLakeTableRespForTable.setTableId(tableId);
-                                pbPrepareLakeTableRespForTable.setLakeTableBucketOffsetsPath(
+                                pbPrepareLakeTableRespForTable.setLakeTableOffsetsPath(
                                         fsPath.toString());
                             } catch (Exception e) {
                                 Errors error = ApiError.fromThrowable(e).error();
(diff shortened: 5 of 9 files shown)


fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/tiering/committer/TieringCommitOperator.java
```diff
@@ -323,33 +285,45 @@ private void checkFlussNotMissingLakeSnapshot(
...
+            if (trimmedPath.startsWith("{")) {
+                throw new IllegalStateException(
+                        String.format(
+                                "The %s field in snapshot property is a JSON string (tiered by v0.8), "
+                                        + "which is not supported to restore. Snapshot ID: %d, Table: {tablePath=%s, tableId=%d}.",
                                 FLUSS_LAKE_SNAP_BUCKET_OFFSET_PROPERTY,
                                 missingCommittedSnapshot.getLakeSnapshotId(),
                                 tablePath,
                                 tableId));
             }
```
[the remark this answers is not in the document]
luoyuxia: the path shouldn't contains '{'

fluss-server/src/main/java/org/apache/fluss/server/zk/data/lake/LakeTableHelper.java
```diff
@@ -119,24 +176,28 @@ private LakeTableSnapshot mergeLakeTable(
...
+
         try (FSDataOutputStream outputStream =
                 fileSystem.create(
-                        remoteLakeTableSnapshotManifestPath, FileSystem.WriteMode.OVERWRITE)) {
+                        remoteLakeTableSnapshotOffsetPath, FileSystem.WriteMode.OVERWRITE)) {
             outputStream.write(jsonBytes);
         }
-        return remoteLakeTableSnapshotManifestPath;
+        return remoteLakeTableSnapshotOffsetPath;
     }
```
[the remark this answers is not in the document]
luoyuxia: LakeUpsertHelper will clean it up

luoyuxia:
@wuchong Could you please help review this pr? The pr also handle the back compabitlity when use v2 to serialize lake table snapshot

fluss-rpc/src/main/proto/FlussApi.proto
```diff
@@ -449,25 +449,27 @@ message NotifyRemoteLogOffsetsRequest {
 message NotifyRemoteLogOffsetsResponse {
 }
 
-message CommitLakeTableSnapshotRequest {
-  repeated PbLakeTableSnapshotInfo tables_req = 1;
+message PrepareCommitLakeTableSnapshotRequest {
```
wuchong: `PrepareLakeTableSnapshotRequest`

`prepare` and `commit` are 2 different phases.

fluss-rpc/src/main/proto/FlussApi.proto
```diff
@@ -449,25 +449,27 @@ message NotifyRemoteLogOffsetsRequest {
...
+  repeated PbTableBucketOffsets bucket_offsets = 1;
 }
 
-message PbLakeTableSnapshotInfo {
-  optional int64 table_id = 1;
-  required int64 snapshot_id = 2;
-  repeated PbLakeTableOffsetForBucket buckets_req = 3;
+message PrepareCommitLakeTableSnapshotResponse {
+  repeated PbPrepareCommitLakeTableRespForTable prepare_commit_lake_table_resp = 1;
```
wuchong: ditto

fluss-rpc/src/main/proto/FlussApi.proto
```diff
@@ -947,11 +949,53 @@ message PbModifyColumn {
...
+  required int32 bucket_id = 2;
+  // Deprecated: log_start_offset is no longer used. Field number 3 is reserved for protocol compatibility.
+  // optional int64 log_start_offset = 3;
+  optional int64 log_end_offset = 4;
+  // Deprecated: partition_name is no longer used. Field number 5 is reserved for protocol compatibility.
+  // optional string partition_name = 5;
+  optional int64 max_timestamp = 6;
+}
+
+message PbPrepareCommitLakeTableRespForTable {
```
wuchong: add a `table_id` field, as the `PrepareCommitLakeTableSnapshotRequest` request has multiple table ids, we need to distinguish which table is the `PbPrepareCommitLakeTableRespForTable` belong to.

fluss-rpc/src/main/proto/FlussApi.proto
```diff
@@ -947,11 +949,53 @@ message PbModifyColumn {
...
+  optional string lake_table_bucket_offsets_path = 1;
+  optional int32 error_code = 2;
+  optional string error_message = 3;
```
wuchong: Let's follow a standard that `error_code` and `error_message` as the first 2 fields, because we will add more fields at the end.

fluss-rpc/src/main/proto/FlussApi.proto
```diff
@@ -947,11 +949,53 @@ message PbModifyColumn {
...
+}
+
+message PbTableBucketOffsets {
```
wuchong: rename to `PbTableOffsets`? This makes it more clear it is a set of offsets for a table.

fluss-server/src/main/java/org/apache/fluss/server/coordinator/CoordinatorEventProcessor.java
```diff
@@ -1224,6 +1225,48 @@ private <T> void processAccessContext(AccessContextEvent<T> event) {
     private void tryProcessCommitLakeTableSnapshot(
             CommitLakeTableSnapshotEvent commitLakeTableSnapshotEvent,
             CompletableFuture<CommitLakeTableSnapshotResponse> callback) {
+        CommitLakeTableSnapshotData commitLakeTableSnapshotData =
+                commitLakeTableSnapshotEvent.getCommitLakeTableSnapshotData();
+        if (commitLakeTableSnapshotData.getLakeTableSnapshotMetadatas().isEmpty()) {
+            handleCommitLakeTableSnapshotV1(commitLakeTableSnapshotEvent, callback);
+        } else {
+            Map<Long, LakeTable.LakeSnapshotMetadata> lakeSnapshotMetadatas =
```
wuchong: extract the else clode block into a `handleCommitLakeTableSnapshotV2()` method

fluss-server/src/main/java/org/apache/fluss/server/zk/data/lake/LakeTableHelper.java
```diff
@@ -44,99 +46,96 @@ public LakeTableHelper(ZooKeeperClient zkClient, String remoteDataDir) {
...
+        return new TableBucketOffsets(newTableBucketOffsets.getTableId(), bucketLogEndOffset);
     }
 
-    private FsPath storeLakeTableSnapshot(
-            long tableId, TablePath tablePath, LakeTableSnapshot lakeTableSnapshot)
-            throws Exception {
-        // get the remote file path to store the lake table snapshot information
-        FsPath remoteLakeTableSnapshotManifestPath =
-                FlussPaths.remoteLakeTableSnapshotManifestPath(remoteDataDir, tablePath, tableId);
+    public FsPath storeLakeTableBucketOffsets(
```
wuchong: `storeLakeTableOffsetsFile` to make it clear it stores the offsets to remote files.

fluss-server/src/main/java/org/apache/fluss/server/zk/data/lake/LakeTableHelper.java
```diff
@@ -44,99 +46,96 @@ public LakeTableHelper(ZooKeeperClient zkClient, String remoteDataDir) {
...
+        // currently, we keep only one lake snapshot metadata in zk,
+        // todo: in solve paimon dv union read issue #2121, we'll keep multiple lake snapshot
+        // metadata
+        // discard previous lake snapshot metadata
+        if (previousLakeSnapshotMetadatas != null) {
+            previousLakeSnapshotMetadatas.forEach(LakeTable.LakeSnapshotMetadata::discard);
+        }
+    }
+
+    public TableBucketOffsets upsertTableBucketOffsets(
```
wuchong: It sounds like this method will `upsert` bucket offsets into zk (very similar to other methods like `upsertLakeTable`). However, it doesn't.  Would be better to move the logic into `org.apache.fluss.server.RpcServiceBase#prepareCommitLakeTableSnapshot`.

fluss-server/src/main/java/org/apache/fluss/server/zk/data/lake/LakeTableHelper.java
```diff
@@ -44,99 +46,96 @@ public LakeTableHelper(ZooKeeperClient zkClient, String remoteDataDir) {
...
+     * just for back compatibility.
      *
      * @param tableId the table ID
-     * @param tablePath the table path
      * @param lakeTableSnapshot the new snapshot to upsert
      * @throws Exception if the operation fails
      */
-    public void upsertLakeTable(
-            long tableId, TablePath tablePath, LakeTableSnapshot lakeTableSnapshot)
+    public void upsertLakeTable(long tableId, LakeTableSnapshot lakeTableSnapshot)
```
wuchong: `registerLakeTableSnapshotV1()` or `registerLakeTableSnapshotLegacy()` to make it clear it is a legacy operation.

fluss-server/src/main/java/org/apache/fluss/server/zk/data/lake/LakeTableHelper.java
```diff
@@ -44,99 +46,96 @@ public LakeTableHelper(ZooKeeperClient zkClient, String remoteDataDir) {
...
-                        // tieredOffsetsFilePath, but in the future we'll commit a readable offsets
-                        // separately to mark what the readable offsets are for a snapshot since
-                        // in paimon dv table, tiered log end offsets is not same with readable
-                        // offsets
-                        lakeTableSnapshotFsPath);
-
-        // currently, we keep only one lake snapshot metadata in zk,
-        // todo: in solve paimon dv union read issue #2121, we'll keep multiple lake snapshot
-        // metadata
+    public void addLakeTableSnapshotMetadata(
```
wuchong: Currently, the method names in `LakeTableHelper` are quite ambiguous. It’s hard to tell from the name alone what each method actually does (e.g., whether it interacts with ZooKeeper, reads/writes to OSS, or performs only local operations).

To improve clarity and maintainability, I suggest renaming them to be more descriptive of their behavior. For example, rename this method to `registerLakeTableSnapshotV2` or `registerLakeTableSnapshot()`. As `register` is a more common verb to reflec it is a `zkClient` operation.

fluss-server/src/main/java/org/apache/fluss/server/coordinator/CoordinatorEventProcessor.java
```diff
@@ -1224,6 +1225,48 @@ private <T> void processAccessContext(AccessContextEvent<T> event) {
...
+                                } catch (Exception e) {
+                                    ApiError error = ApiError.fromThrowable(e);
+                                    tableResp.setError(error.error().code(), error.message());
+                                }
+                            }
+                            coordinatorEventManager.put(
+                                    new NotifyLakeTableOffsetEvent(
+                                            commitLakeTableSnapshotData.getLakeTableSnapshot(),
+                                            commitLakeTableSnapshotData
+                                                    .getTableBucketsMaxTieredTimestamp()));
```
wuchong: We should move the "putting notify event" into the try block. Otherwise, when the zk register fails, we shouldn't notify the event.

wuchong: We should move the "putting notify event" into the try block. Otherwise, when the zk register fails, we shouldn't notify the event.

fluss-server/src/main/java/org/apache/fluss/server/zk/data/lake/LakeTable.java
```diff
@@ -112,17 +120,19 @@ public List<LakeSnapshotMetadata> getLakeSnapshotMetadatas() {
      *
      * @return the LakeTableSnapshot
      */
-    public LakeTableSnapshot getLatestTableSnapshot() throws Exception {
+    public LakeTableSnapshot getLatestTableSnapshot() throws IOException {
```
wuchong: Consider renaming this method to `getOrReadLatestTableSnapshot` to **explicitly indicate that it may perform I/O operations** (e.g., reading from remote storage) when the snapshot isn't already available in memory. This makes the method’s behavior clearer to callers and improves code readability.

fluss-server/src/main/java/org/apache/fluss/server/zk/data/lake/LakeTableJsonSerde.java
```diff
@@ -61,28 +61,35 @@ public class LakeTableJsonSerde implements JsonSerializer<LakeTable>, JsonDeseri
...
+        if (lakeTable.getLakeSnapshotMetadatas() == null) {
+            // Version 1: ZK node contains full snapshot data, use LakeTableSnapshotJsonSerde
+            LakeTableSnapshotJsonSerde.INSTANCE.serialize(
+                    lakeTable.getLatestTableSnapshot(), generator);
+        } else {
             generator.writeStartObject();
+            generator.writeNumberField(VERSION_KEY, CURRENT_VERSION);
+
+            generator.writeArrayFieldStart(LAKE_SNAPSHOTS);
+            for (LakeTable.LakeSnapshotMetadata lakeSnapshotMetadata :
```
wuchong: It would be better to extract separate `serializeV1()` and `serializeV2()` methods to improve maintainability and readability. The same applies to the `deserialize`.

fluss-server/src/main/java/org/apache/fluss/server/zk/data/lake/LakeTableSnapshotJsonSerde.java
```diff
@@ -18,168 +18,97 @@
...
+ *   <li>snapshot_id: the snapshot ID
+ *   <li>table_id: the table ID (derived from the first bucket)
+ *   <li>buckets: array of bucket objects, each containing bucket_id, optional partition_id, and
+ *       log_end_offset
  * </ul>
+ *
+ * @see LakeTableJsonSerde for the current format (version 2) that uses this serde for legacy
+ *     compatibility
  */
 public class LakeTableSnapshotJsonSerde
```
wuchong: Consider to rename this to `LakeTableSnapshotLegacyJsonSerde` (Currenlty there is too many `LakeXxxSerde`)

fluss-server/src/test/java/org/apache/fluss/server/zk/data/LakeTableSnapshotJsonSerdeTest.java
```diff
@@ -41,104 +34,37 @@ class LakeTableSnapshotJsonSerdeTest extends JsonSerdeTestBase<LakeTableSnapshot
...
+                    + "\"buckets\":[{\"bucket_id\":2,\"log_end_offset\":4},"
+                    + "{\"bucket_id\":1,\"log_end_offset\":3}]}",
+            "{\"version\":1,\"snapshot_id\":3,\"table_id\":5,"
+                    + "\"buckets\":[{\"partition_id\":1,\"bucket_id\":1,\"log_end_offset\":3},"
+                    + "{\"partition_id\":2,\"bucket_id\":1,\"log_end_offset\":4}]}"
         };
     }
-
-    @Test
-    void testBackwardCompatibility() {
```
wuchong, line 115: We still need this for backward compatibility for the `max_timestamp` and `log_start_offset` fields.

fluss-server/src/main/java/org/apache/fluss/server/coordinator/CoordinatorService.java
```diff
@@ -694,6 +708,47 @@ public CompletableFuture<ControlledShutdownResponse> controlledShutdown(
         return response;
     }
 
+    @Override
+    public CompletableFuture<PrepareCommitLakeTableSnapshotResponse> prepareCommitLakeTableSnapshot(
```
wuchong: move this method before `commitLakeTableSnapshot`.

fluss-rpc/src/main/proto/FlussApi.proto
luoyuxia: I was thinking to use array index to distinguish the `PbPrepareCommitLakeTableRespForTable ` belong to. It's a week contract. I'm fine with attaching table id which is a stronger contract, and also align with `CommitLakeTableSnapshotRequest`

luoyuxia:
@wuchong Thanks for review. Comments has been addressed

fluss-server/src/main/java/org/apache/fluss/server/coordinator/CoordinatorEventProcessor.java
```diff
@@ -1262,20 +1275,58 @@ private void tryProcessCommitLakeTableSnapshot(
...
+                            long tableId = lakeSnapshotMetadataEntry.getKey();
+                            tableResp.setTableId(tableId);
+                            try {
+                                lakeTableHelper.registerLakeTableSnapshotV2(
+                                        tableId, lakeSnapshotMetadataEntry.getValue());
+                                coordinatorEventManager.put(
+                                        new NotifyLakeTableOffsetEvent(
+                                                commitLakeTableSnapshotData.getLakeTableSnapshot(),
+                                                commitLakeTableSnapshotData
+                                                        .getTableBucketsMaxTieredTimestamp()));
```
wuchong: Moving the event-publishing logic into the `forEach` loop would cause the event to be sent multiple times, since the event is intended for all table ids, not per individual table.

fluss-server/src/main/java/org/apache/fluss/server/coordinator/CoordinatorEventProcessor.java
```diff
@@ -1262,20 +1275,58 @@ private void tryProcessCommitLakeTableSnapshot(
...
-                                        tableId, tablePath, lakeTableSnapshotEntry.getValue());
+                                lakeTableHelper.registerLakeTableSnapshotV1(
+                                        tableId, lakeTableSnapshotEntry.getValue());
+                                // send notify lakehouse data request to all replicas via
+                                // coordinator event
+                                coordinatorEventManager.put(
+                                        new NotifyLakeTableOffsetEvent(
+                                                lakeTableSnapshots,
+                                                commitLakeTableSnapshotData
+                                                        .getTableBucketsMaxTieredTimestamp()));
```
wuchong: ditto

fluss-rpc/src/main/java/org/apache/fluss/rpc/protocol/ApiKeys.java
```diff
@@ -79,7 +79,8 @@ public enum ApiKeys {
     REMOVE_SERVER_TAG(1048, 0, 0, PUBLIC),
     REBALANCE(1049, 0, 0, PUBLIC),
     LIST_REBALANCE_PROGRESS(1050, 0, 0, PUBLIC),
-    CANCEL_REBALANCE(1051, 0, 0, PUBLIC);
+    CANCEL_REBALANCE(1051, 0, 0, PUBLIC),
+    PRE_LAKE_TABLE_SNAPSHOT(1052, 0, 0, PRIVATE);
```
wuchong: `PRE_LAKE_TABLE_SNAPSHOT` -> `PREPARE_LAKE_TABLE_SNAPSHOT`

While `PRE` might imply "before," it’s ambiguous and non-idiomatic in event or phase naming. `PREPARE` clearly conveys the intent.

fluss-rpc/src/main/proto/FlussApi.proto
```diff
@@ -1012,4 +1014,49 @@ message PbRebalancePlanForBucket {
...
+  // Deprecated: partition_name is no longer used. Field number 5 is reserved for protocol compatibility.
+  // optional string partition_name = 5;
+  optional int64 max_timestamp = 6;
+}
+
+message PbPrepareLakeTableRespForTable {
+  optional int32 error_code = 1;
+  optional string error_message = 2;
+  optional int64 table_id = 3;
+  optional string lake_table_bucket_offsets_path = 4;
```
wuchong: `lake_table_offsets_path` to align with the `TableOffsets` concept.

fluss-rpc/src/main/java/org/apache/fluss/rpc/gateway/CoordinatorGateway.java
```diff
@@ -66,6 +68,30 @@ public interface CoordinatorGateway extends RpcGateway, AdminGateway {
     CompletableFuture<CommitRemoteLogManifestResponse> commitRemoteLogManifest(
             CommitRemoteLogManifestRequest request);
 
+    /**
+     * Prepares lake table snapshots by merging them with existing snapshots and storing them to the
+     * file system.
```
wuchong:
```suggestion
     * Prepares lake table snapshots by merging them with existing snapshots and storing them to the
     * remote file system.
```

fluss-rpc/src/main/java/org/apache/fluss/rpc/gateway/CoordinatorGateway.java
```diff
@@ -66,6 +68,30 @@ public interface CoordinatorGateway extends RpcGateway, AdminGateway {
...
+     *
+     * <p>This method is called during the two-phase commit process for lake table snapshots. It
+     * performs the following operations for each table in the request:
+     *
+     * <ul>
+     *   <li>Merges the new snapshot with the previous latest snapshot (if exists) to ensure
+     *       completeness
+     *   <li>Stores the merged snapshot to the remote file system. The stored file contains the log
+     *       end offset information for each bucket in the table
+     *   <li>Returns the file path where the snapshot is stored
```
wuchong, line 83: Add an item for the second phase. 

```java
<li>Call {@link #commitLakeTableSnapshot(CommitLakeTableSnapshotRequest)} with the offset
     *       file path to finalize the snapshot commit to ZooKeeper in the second phase.
```

wuchong (commented):
Thanks @luoyuxia for the updating. I only left some final comments.

PR #2228, opened 2025-12-23
[lake] Fix zk lake snapshot node compatible issue

luoyuxia:
### Purpose

Linked issue: close #xxx
This fix the compatible issue introduced by #2037
After #2037, when upgrade coordinator server firstly, the coordinator server will write zk node with v2 version, but the tablet server which is still not upgraded can't regconize v2 which casue the exception.

### Brief change log
- introduce a version in commit lake snapshot request, new tiering service should set the version to 2, if the coordinator server received version is v2, use the v2 version to write zk node.

So, the upgrade process can be: 
- upgrade coordinator server 
- upgrade tablet server
- upgrade tiering service


luoyuxia:
@wuchong Could you please help review this pr which to fix the compatible issue for zk lake snapshot node?

luoyuxia:
Will fix in #2223

PR #2229, opened 2025-12-23
[docs] Use the `docker compose ps` to show the running containers

zuston:
### Purpose

Use `docker compose ps` to show only the containers created by the Fluss quick-start, rather than all containers on the machine.


zhuangchong (approved):
+1

wuchong (approved):

Issue #2238, opened 2025-12-24
getRuntimeContext().getAttemptNumber() is not founded in flink 2.1


sd4324530:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/fluss/issues) and found nothing similar.

### Fluss version

0.8.0 (latest release)

### Please describe the bug 🐞

<img width="2418" height="1052" alt="Image" src="https://github.com/user-attachments/assets/26c21d33-16a2-4c11-9c42-492167996ea5" />

`getAttemptNumber()` is not founded in flink 2.1

### Solution

<img width="639" height="285" alt="Image" src="https://github.com/user-attachments/assets/221e055f-d1e1-4696-8eff-bf970a30d0c4" />

The current code is compatible with Flink version 1.18. Newer versions can replace it with: getRuntimeContext().getTaskInfo().getAttemptNumber().
We should need to adapt it for Flink 1.19 and above.

### Are you willing to submit a PR?

- [x] I'm willing to submit a PR!

sd4324530:
@loserwang1024 I will submit my PR here as soon as possible.

loserwang1024:
@sd4324530 Thanks a lot, an adapter for flink common and flink 1.18 is enough.

PR #2241, opened 2025-12-24
[Flink] Add RuntimeContextAdapter

sd4324530:
### Purpose

Linked issue: close https://github.com/apache/fluss/issues/2238

### Brief change log

add RuntimeContextAdapter in module `fluss-flink-1.18` and `fluss-flink-common`

### DIFF
--- a/fluss-flink/fluss-flink-1.19/src/test/java/org/apache/fluss/flink/tiering/committer/Flink119TieringCommitOperatorTest.java
+++ b/fluss-flink/fluss-flink-1.19/src/test/java/org/apache/fluss/flink/tiering/committer/Flink119TieringCommitOperatorTest.java
@@ -17,45 +17,8 @@
 
 package org.apache.fluss.flink.tiering.committer;
 
-import org.apache.fluss.flink.tiering.TestingLakeTieringFactory;
-
-import org.apache.flink.configuration.Configuration;
-import org.apache.flink.runtime.operators.coordination.MockOperatorEventGateway;
-import org.apache.flink.runtime.operators.testutils.DummyEnvironment;
-import org.apache.flink.streaming.api.operators.StreamOperatorParameters;
-import org.apache.flink.streaming.runtime.tasks.SourceOperatorStreamTask;
-import org.apache.flink.streaming.util.MockOutput;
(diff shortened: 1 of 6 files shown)


vamossagar12 (approved):
LGTM

sd4324530:
@wuchong @loserwang1024 ,cc

wuchong (commented):
Thanks @sd4324530 , could you add Unit Tests or Integration Tests to reproduce this issue? So that we won't break this compatibility in the future. 

Others looks good to me.

sd4324530:
@wuchong Fixd. Please take a look.

wuchong:
@sd4324530 thank you. Could you also add this test for flink 1.18 and flink 1.19? This helps verify the added adapter in Flink 1.18 is correct, and not adding adapter in Flink 1.19 is also correct. 

Besides, could you use **rebase** to rebase your branch onto `main` branch instead of using **merge**? It's suggest not using "git merge" to merge branches, otherwise the changes is hard to track. Please use "git rebase" to rebase branches instead. IntelliJ IDEA provide an easy tool to do git rebase, you can find the tool via `VCS -> Git -> Rebase`.

fluss-flink/fluss-flink-1.19/src/test/java/org/apache/fluss/flink/tiering/committer/Flink119TieringCommitOperatorTest.java
```diff
@@ -0,0 +1,61 @@
...
+
+/**
+ * UT for {@link TieringCommitOperator}. Test the compatibility of the `getAttemptNumber` method in
+ * flink 1.19.
+ */
+public class Flink119TieringCommitOperatorTest extends TieringCommitOperatorTest {
+
+    @BeforeEach
+    @Override
+    void beforeEach() throws Exception {
```
wuchong: Do not override parent test method, otherwise, it's hard to maintain. If the Flink classes used in test are also not compatible across different Flink versions, we should do the same `adapter` for them (but leave them in test packages only), you can checkout `ResolvedCatalogMaterializedTableAdapter` for an example.

wuchong (approved):

PR #2255, opened 2025-12-25
Implement basic Aggregate Merge Engine

### LINKED ISSUE #2254: [server] Implement basic Aggregate Merge Engine
### Search before asking

### Description

Subtask of https://github.com/apache/fluss/issues/2133

Implement Aggregate Merge Engine in Fluss Engine with support for primary data types and basic aggregate functions (e.g., SUM, COUNT, MIN, MAX).

### Willingness to contribute

platinumhamburg:
### Purpose

Linked issue: close #2254

### DIFF
--- a/fluss-server/src/test/java/org/apache/fluss/server/kv/rowmerger/RowMergerCreateTest.java
+++ b/fluss-server/src/test/java/org/apache/fluss/server/kv/rowmerger/RowMergerCreateTest.java
@@ -76,9 +76,6 @@ class RowMergerCreateTest {
 
         assertThat(merger).isInstanceOf(AggregateRowMerger.class);
 
-        // Configure target schema before merge
-        merger.configureTargetColumns(null, SCHEMA_ID, schema);
-
         // Verify max aggregation
         BinaryRow row1 = compactedRow(schema.getRowType(), new Object[] {1, 10L});
         BinaryRow row2 = compactedRow(schema.getRowType(), new Object[] {1, 20L});


website/docs/table-design/table-types/pk-table/merge-engines/aggregation.md
```diff
@@ -0,0 +1,534 @@
+---
+sidebar_label: Aggregation
+title: Aggregation Merge Engine
+sidebar_position: 4
```
wuchong:
```suggestion
sidebar_position: 5
```

website/docs/table-design/table-types/pk-table/merge-engines/aggregation.md
```diff
@@ -0,0 +1,534 @@
...
+'table.merge-engine' = 'aggregation'
+```
+
+Then specify the aggregate function for each non-primary key field using connector options:
+
+```
+'fields.<field-name>.agg' = '<function-name>'
+```
+
+**Note**: The recommended way is to use Schema API (see section "API Usage" below). The connector option is provided as an alternative for connector-specific scenarios.
```
wuchong: I think it's hard to tell which way is recommended, as it depends on the client users are working with:  
- For **Java clients**, the `Schema` API is natural.  
- For **compute engines** like Flink or Spark, DDL and connector options are more typical.

Given that most Fluss users interact with it through engines like **Flink** or **Spark**, I suggest **prioritizing Flink SQL examples** in the documentation.

We can use **Tabs** to show both Flink SQL and Java Client examples side by side, with **Flink SQL as the default tab**.  
You can refer to `datastream.mdx` for an example of how to implement tabs in our Markdown documentation.

Besides, we can use `INSERT INTO ... VALUES ...` and `SELECT * FROM ..` for Flink SQL examples to show the input and result data which is easier to demo for users.

website/docs/table-design/table-types/pk-table/merge-engines/aggregation.md
```diff
@@ -0,0 +1,534 @@
...
+    .primaryKey("id")
+    .build();
+
+TableDescriptor.builder()
+    .schema(schema)
+    .property("table.merge-engine", "aggregation")
+    .build();
+
+// Input: (1, 'online', '2024-01-01 10:00:00'), (1, 'offline', '2024-01-01 11:00:00')
+// Result: (1, 'offline', '2024-01-01 11:00:00')
```
wuchong: it's better to insert a `null` for one of the `last_value` column to demostrate the null handling.

website/docs/table-design/table-types/pk-table/merge-engines/aggregation.md
```diff
@@ -0,0 +1,534 @@
...
+Schema schema = Schema.newBuilder()
+    .column("id", DataTypes.BIGINT())
+    .column("tags", DataTypes.STRING(), AggFunction.LISTAGG)
+    .primaryKey("id")
+    .build();
+
+TableDescriptor.builder()
+    .schema(schema)
+    .property("table.merge-engine", "aggregation")
+    .property("table.merge-engine.aggregate.tags.listagg-delimiter", ";")
```
wuchong: This is an issue I didn’t anticipate when designing the Schema API. Since `delimiter` is an **aggregate-specific parameter** (and we may introduce more such parameters in the future), it logically belongs to the **definition of the aggregate function itself**—not as a separate table property.

However, this introduces two concerns:

1. **It breaks our original design principle**: previously, aggregate functions were defined purely in terms of columns. Now, the function definition is split between the schema and table properties, reducing cohesion.
2. **Naming conflict**: the property name is overly long and clashes with the existing Flink connector option `fields.xxx.listagg-delimiter`, which could cause confusion or redundancy.

I see two viable paths forward:

**Option 1: Remove the concept of “aggregate column”**
- The schema would only describe *column names* and *types*.
- All aggregate function metadata (including `delimiter`) would live in **table properties**.
- ✅ *Benefits*: cleaner separation, better alignment across APIs (Java, Flink, Spark, Trino, etc.), and easier extensibility for future engines.

**Option 2: Introduce a richer aggregate function model (similar to `DataTypes`)**  
- Rename `enum AggFunction` → `AggFunctionType`.
- Introduce a class `AggFunction` that can carry parameters (e.g., `delimiter`).
- Add a utility class `AggFunctions` with methods like:
  ```java
  AggFunctions.SUM()
  AggFunctions.LISTAGG(";")
  ```
- Allow inline definition in the schema builder:
  ```java
  .column("tags", DataTypes.STRING(), AggFunctions.LISTAGG(";"))
  ```
- The Flink connector options would still translate into these schema builder calls.
- ✅ *Benefits*: expressive, type-safe, and keeps logic co-located with the column definition.

I’m fine with either approach. Which approach do you prefer?

website/docs/table-design/table-types/pk-table/merge-engines/aggregation.md
```diff
@@ -0,0 +1,534 @@
...
+    .build();
+
+// Input: (1, 'developer'), (1, 'java'), (1, 'flink')
+// Result: (1, 'developer;java;flink')
```
wuchong: It would be better to include an additional `LISTAGG` column that uses the **default behavior** (e.g., comma as delimiter). Since most users will rely on the defaults, demonstrating the expected output in this common case improves usability and sets clear expectations.

website/docs/table-design/table-types/pk-table/merge-engines/aggregation.md
```diff
@@ -0,0 +1,534 @@
...
+
+## Supported Aggregate Functions
+
+Fluss currently supports the following aggregate functions:
+
+### sum
+
+Aggregates values by computing the sum across multiple rows.
+
+- **Supported Data Types**: TINYINT, SMALLINT, INT, BIGINT, FLOAT, DOUBLE, DECIMAL
```
wuchong:
```suggestion
- **Supported Data Types**: `TINYINT`, `SMALLINT`, `INT`, `BIGINT`, `FLOAT`, `DOUBLE`, `DECIMAL`
```

nit: use code format for the data types. Same to others.

website/docs/table-design/table-types/pk-table/merge-engines/aggregation.md
```diff
@@ -0,0 +1,534 @@
...
+
+TableDescriptor tableDescriptor = TableDescriptor.builder()
+    .schema(schema)
+    .property("table.merge-engine", "aggregation")
+    .build();
+
+// Create partial update writer targeting only id, count1, and sum1
+int[] targetColumns = new int[]{0, 1, 3}; // id, count1, sum1
+UpsertWriter partialWriter = table.newUpsert()
+    .withPartialUpdate(targetColumns)
```
wuchong: Use name-based `.partialUpdate("id", "count1", "sum1")` which is more straightforward.

website/docs/table-design/table-types/pk-table/merge-engines/aggregation.md
```diff
@@ -0,0 +1,534 @@
...
+TableDescriptor.builder()
+    .schema(schema)
+    .property("table.merge-engine", "aggregation")
+    .property("table.agg.remove-record-on-delete", "true")  // Default is false
+    .build();
+```
+
+**Configuration options**:
+- **`'table.agg.remove-record-on-delete' = 'false'`** (default): Delete operations will cause an error
+- **`'table.agg.remove-record-on-delete' = 'true'`**: Delete operations will remove the entire record from the table
```
wuchong: This configuration appears to serve the same purpose (and may conflict) with the existing `table.delete.behavior` setting, which defaults to `ALLOW` for regular tables, but is default to `IGNORE` for `FIRST_ROW` and `VERSIONED` merge engines.

I suggest we **reuse** `table.delete.behavior` and simply extend its logic to also default to `IGNORE` for the **aggregation merge engine**, rather than introducing a new, overlapping configuration. This keeps the configuration surface consistent and avoids ambiguity.

website/docs/table-design/table-types/pk-table/merge-engines/aggregation.md
```diff
@@ -0,0 +1,534 @@
...
+:::warning Critical Limitations
+When using the `aggregation` merge engine, be aware of the following critical limitations:
+
+### 1. Exactly-Once Semantics
+
+**Fluss engine does not natively support transactional writes, and therefore does not directly support Exactly-Once semantics at the storage layer.**
+
+Exactly-Once semantics should be achieved through integration with compute engines (e.g., Flink, Spark). For example, after failover, undo operations can be generated for invalid writes to achieve rollback.
+
+For detailed information about Exactly-Once implementation, please refer to: [FIP-21: Aggregation Merge Engine](https://cwiki.apache.org/confluence/display/FLUSS/FIP-21%3A+Aggregation+Merge+Engine)
```
wuchong: From a user’s perspective, the current wording may give the impression that Fluss’s aggregate merge engine **does not support exactly-once semantics at all**. To avoid confusion, we should first clarify this at the beginning of the section:

- **When writing to an aggregate merge engine table using the Flink engine**, Fluss **does provide exactly-once guarantees**. Thanks to Flink’s checkpointing mechanism, in the event of a failure and recovery, the Flink connector automatically performs an **undo operation** to roll back the table state to what it was at the last successful checkpoint. This ensures no over-counting or under-counting: data remains consistent and accurate.

And then to mention the limitation:

- **However, when using the Fluss client API directly** (outside of Flink), **exactly-once is not provided out of the box**. In such cases, users must implement their own recovery logic (similar to what the Flink connector does) by explicitly resetting the table state to a previous version by performing undo operations.

website/docs/table-design/table-types/pk-table/merge-engines/aggregation.md
```diff
@@ -0,0 +1,534 @@
...
+### 2. Delete Operations
+
+By default, delete operations will cause errors:
+- You must set `'table.agg.remove-record-on-delete' = 'true'` if you need to handle delete operations
+- This configuration will remove the entire aggregated record, not reverse individual aggregations
+
+### 3. Data Type Restrictions
+
+Each aggregate function supports specific data types (see function documentation above)
+:::
```
wuchong: I don’t consider this a true limitation—after all, aggregate functions in compute engines (like Flink or Spark) also don’t support all data types universally. I suggest we **remove this item** from the list of limitations.

Separately, we should **explicitly document the limitation around retraction messages** (i.e., handling of `UPDATE_BEFORE` messages). It’s important to clarify which aggregate functions support retraction and which do not, as this directly impacts correctness in streaming workloads.

I think we can also add the **Retraction Handling** to each aggregate function documentation. And have a brief description in the "Limitations" section.

For reference, Paimon provides a clear treatment of this topic here:  

https://paimon.apache.org/docs/master/primary-key-table/merge-engine/aggregation/#retraction

wuchong (commented):
Thanks @platinumhamburg , I didn't complete the review. I gave some feedbacks first after reading the documentation.

website/docs/table-design/table-types/pk-table/merge-engines/aggregation.md
platinumhamburg: I removed the [Data Type Restrictions] section, besides, we are not providing support for retraction semantics in this PR, so for the sake of consistency, the documentation will also temporarily omit detailed descriptions related to retraction. This content will be added in a subsequent PR.

website/docs/table-design/table-types/pk-table/merge-engines/aggregation.md
platinumhamburg: I chose Option 2 and completed the refactoring based on it.

platinumhamburg:
> Thanks @platinumhamburg , I didn't complete the review. I gave some feedbacks first after reading the documentation.

@wuchong I've refactored both the code and documentation based on the feedback. Thanks for review again.

fluss-common/src/main/java/org/apache/fluss/metadata/AggFunctions.java
```diff
@@ -0,0 +1,293 @@
...
+     * <p>Supported data types: STRING, CHAR
+     *
+     * <p>Null handling: Null values are skipped
+     *
+     * @param delimiter the delimiter to use for concatenation
+     * @return a LISTAGG aggregation function with the specified delimiter
+     */
+    public static AggFunction LISTAGG(String delimiter) {
+        Map<String, String> params = new HashMap<>();
+        params.put("delimiter", delimiter);
```
wuchong, line 203: Add `delimiter` as a `public static` constant to enable easy and consistent referencing across the codebase.

fluss-common/src/main/java/org/apache/fluss/metadata/MergeEngineType.java
```diff
@@ -52,9 +52,16 @@ public enum MergeEngineType {
...
 
-    // introduce AGGREGATE merge engine in the future
+    /**
+     * A merge engine that aggregates rows with the same primary key using field-level aggregate
+     * functions. Each non-primary-key field can have its own aggregate function (e.g., sum, max,
+     * min, last_value, etc.). This allows for flexible aggregation semantics at the field level.
+     *
+     * @since 0.9
+     */
+    AGGREGATE;
```
wuchong: `AGGREGATION`

fluss-common/src/test/java/org/apache/fluss/utils/json/SchemaJsonSerdeTest.java
```diff
@@ -131,4 +152,43 @@ protected String[] jsonLackOfColumnId() {
 
         return new String[] {oldSchemaJson1, oldSchemaJson2, oldSchemaJson2, oldSchemaJson3};
     }
+
+    @Test
+    void testSchemaWithAggregationFunctions() {
```
wuchong: Since you have added `SCHEMA_WITH_AGG` into `createObjects()`, this json serialization and deserialization test has already been verified by parent `org.apache.fluss.utils.json.JsonSerdeTestBase#testJsonSerde`. So I think we can remove this test method.

fluss-common/src/test/java/org/apache/fluss/utils/json/SchemaJsonSerdeTest.java
```diff
@@ -131,4 +152,43 @@ protected String[] jsonLackOfColumnId() {
...
+                .isEqualTo(AggFunctions.LAST_VALUE_IGNORE_NULLS());
+        assertThat(deserialized.getAggFunction("product_id"))
+                .isEmpty(); // Primary key has no agg function
+
+        // Test round-trip
+        assertThat(deserialized).isEqualTo(SCHEMA_WITH_AGG);
+    }
+
+    @Test
+    void testSchemaWithoutAggregationFunctions() {
```
wuchong: ditto. we can remove it.

fluss-common/src/test/java/org/apache/fluss/utils/json/SchemaJsonSerdeTest.java
```diff
@@ -89,19 +90,39 @@ public class SchemaJsonSerdeTest extends JsonSerdeTestBase<Schema> {
     static final String SCHEMA_JSON_4 =
             "{\"version\":1,\"columns\":[{\"name\":\"a\",\"data_type\":{\"type\":\"INTEGER\",\"nullable\":false},\"comment\":\"a is first column\",\"id\":0},{\"name\":\"b\",\"data_type\":{\"type\":\"INTEGER\",\"nullable\":false},\"comment\":\"b is second column\",\"id\":1},{\"name\":\"c\",\"data_type\":{\"type\":\"CHAR\",\"nullable\":false,\"length\":10},\"comment\":\"c is third column\",\"id\":2}],\"primary_key\":[\"a\",\"c\"],\"auto_increment_column\":[\"b\"],\"highest_field_id\":2}";
 
+    static final Schema SCHEMA_WITH_AGG =
+            Schema.newBuilder()
+                    .column("product_id", DataTypes.BIGINT().copy(false))
+                    .column("total_sales", DataTypes.BIGINT(), AggFunctions.SUM())
+                    .column("max_price", DataTypes.DECIMAL(10, 2), AggFunctions.MAX())
```
wuchong, line 97: Could you add aggregate functions that has parameters, like `LISTAGG`, to verify the json format of such agg functions.

fluss-common/src/main/java/org/apache/fluss/config/TableConfig.java
```diff
@@ -118,6 +118,28 @@ public Optional<String> getMergeEngineVersionColumn() {
...
+    /**
+     * Gets the listagg delimiter for a specific field in aggregate merge engine.
+     *
+     * <p>This method dynamically constructs the ConfigOption for field-specific delimiter. The
+     * configuration key pattern is: table.merge-engine.aggregate.{fieldName}.listagg-delimiter
+     *
+     * @param fieldName the field name
+     * @return the delimiter string, default to comma if not configured
+     */
+    public String getFieldListaggDelimiter(String fieldName) {
```
wuchong: We can remove this now, as the delimiter parameter is placed in `AggFunction`.

fluss-common/src/test/java/org/apache/fluss/metadata/TableSchemaTest.java
```diff
@@ -121,4 +121,104 @@ void testAutoIncrementColumnSchema() {
                 .isInstanceOf(IllegalStateException.class)
                 .hasMessage("Auto increment column can only be used in primary-key table.");
     }
+
+    @Test
+    void testSchemaBuilderColumnWithAggFunction() {
```
wuchong, line 126: Add LISTAGG with delimiter test

fluss-common/src/main/java/org/apache/fluss/metadata/Schema.java
```diff
@@ -303,7 +322,47 @@ public Builder fromColumns(List<Column> inputColumns) {
...
+        }
+
+        /**
+         * Declares a column with aggregation function that is appended to this schema.
+         *
+         * <p>This method associates an aggregation function with a non-primary key column. It is
+         * only applicable when the table uses aggregation merge engine.
+         *
+         * <p>If aggregation function is not specified for a non-primary key column, it defaults to
+         * {@link AggFunction#LAST_VALUE_IGNORE_NULLS}.
```
wuchong:
```suggestion
         * {@link AggFunctions#LAST_VALUE_IGNORE_NULLS}.
```

fluss-common/src/main/java/org/apache/fluss/metadata/Schema.java
```diff
@@ -303,7 +322,47 @@ public Builder fromColumns(List<Column> inputColumns) {
...
+            checkNotNull(dataType, "Data type must not be null.");
+            checkNotNull(aggFunction, "Aggregation function must not be null.");
+
+            // Validate that this column is not a primary key
+            if (primaryKey != null && primaryKey.columnNames.contains(columnName)) {
+                throw new IllegalArgumentException(
+                        String.format(
+                                "Cannot set aggregation function for primary key column '%s'. "
+                                        + "Primary key columns automatically use 'primary-key' aggregation.",
+                                columnName));
```
wuchong: Could you move the check into `org.apache.fluss.metadata.Schema#normalizeColumns`? That's where we validate the relationship between columns and constraints. Currently, you have to check at multiple places (`column()` and `build()`)

fluss-common/src/main/java/org/apache/fluss/metadata/Schema.java
```diff
@@ -402,6 +461,21 @@ public Schema build() {
...
+                for (Column column : columns) {
+                    if (primaryKey.columnNames.contains(column.getName())
+                            && column.getAggFunction().isPresent()) {
+                        throw new IllegalArgumentException(
+                                String.format(
+                                        "Cannot set aggregation function for primary key column '%s'. "
+                                                + "Primary key columns automatically use 'primary-key' aggregation.",
+                                        column.getName()));
+                    }
+                }
```
wuchong: ditto.

fluss-server/src/main/java/org/apache/fluss/server/kv/rowmerger/aggregate/factory/FieldBoolAndAggFactory.java
```diff
@@ -0,0 +1,50 @@
...
+import org.apache.fluss.server.kv.rowmerger.aggregate.functions.FieldBoolAndAgg;
+import org.apache.fluss.types.BooleanType;
+import org.apache.fluss.types.DataType;
+
+import static org.apache.fluss.utils.Preconditions.checkArgument;
+
+/** Factory for {@link FieldBoolAndAgg}. */
+public class FieldBoolAndAggFactory implements FieldAggregatorFactory {
+
+    public static final String NAME = "bool_and";
```
wuchong: Instead of hardcoding `"bool_and"`, we can use `AggFunctionType.BOOL_AND.getIdentifier()`. This ensures consistency of the identifier across all components.

Additionally, I suggest **using `Enum#name()` directly as the identifier** (e.g., `BOOL_AND.name()`), which eliminates the need to maintain separate identifier fields in `AggFunctionType`. This simplifies the code, reduces redundancy, and guarantees alignment between the enum name and its serialized form.

fluss-common/src/main/java/org/apache/fluss/metadata/AggFunctionType.java
```diff
@@ -0,0 +1,111 @@
...
+        String normalized = name.replace('-', '_').toLowerCase(Locale.ROOT).trim();
+
+        // Try direct match with identifier
+        for (AggFunctionType aggFunc : values()) {
+            if (aggFunc.identifier.equals(normalized)) {
+                return aggFunc;
+            }
+        }
+
+        return null;
```
wuchong: Using `Enum.valueOf` is much more efficient. 

```java
String normalized = name.replace('-', '_').toUpperCase(Locale.ROOT).trim();
return AggFunctionType.valueOf(normalized);
```

fluss-common/src/main/java/org/apache/fluss/metadata/AggFunctionType.java
```diff
@@ -0,0 +1,111 @@
...
+    FIRST_VALUE("first_value"),
+    FIRST_VALUE_IGNORE_NULLS("first_value_ignore_nulls"),
+
+    // String aggregation
+    LISTAGG("listagg"),
+    STRING_AGG("string_agg"), // Alias for LISTAGG - maps to same factory
+
+    // Boolean aggregation
+    BOOL_AND("bool_and"),
+    BOOL_OR("bool_or");
```
wuchong: I think we don't need to maintain additional `identifier` for the agg types, we can directly use `AggFunctionType#name()` as the identifier (which is a upper case).

fluss-server/src/main/java/org/apache/fluss/server/kv/rowmerger/aggregate/factory/FieldAggregatorFactory.java
```diff
@@ -0,0 +1,108 @@
...
+     * @return the factory, or null if not found
+     */
+    static FieldAggregatorFactory getFactory(String identifier) {
+        return FactoryRegistry.INSTANCE.getFactory(identifier);
+    }
+
+    /** Registry for field aggregator factories using Java SPI. */
+    class FactoryRegistry {
+        private static final FactoryRegistry INSTANCE = new FactoryRegistry();
+        private final Map<String, FieldAggregatorFactory> factories;
```
wuchong: We can use `EnumMap` here to better performance, and use the `AggFunctionType` to get factory directly in `create()` method.

```java
private final EnumMap<AggFunctionType, FieldAggregatorFactory> factories =
                new EnumMap<>(AggFunctionType.class);
```

fluss-server/src/main/java/org/apache/fluss/server/kv/rowmerger/aggregate/functions/FieldAggregator.java
```diff
@@ -0,0 +1,68 @@
...
+import java.io.Serializable;
+
+/** Abstract class for aggregating a field of a row. */
+public abstract class FieldAggregator implements Serializable {
+
+    private static final long serialVersionUID = 1L;
+
+    protected final DataType fieldType;
+    protected final DataTypeRoot typeRoot;
+    protected final String name;
```
wuchong: Can use the `AggFunctionType` to replace the `name`.

wuchong: Can use `AggFunctionType` to replace the `name`?

fluss-common/src/main/java/org/apache/fluss/metadata/AggFunctions.java
```diff
@@ -0,0 +1,293 @@
...
+     * <p>Concatenates multiple string values into a single string with a delimiter.
+     *
+     * <p>Supported data types: STRING, CHAR
+     *
+     * <p>Null handling: Null values are skipped
+     *
+     * @return a LISTAGG aggregation function with comma delimiter
+     */
+    public static AggFunction LISTAGG() {
+        return LISTAGG(",");
```
wuchong: nit: Since the runtime implementation already assumes a default delimiter of `,`, we can omit the delimiter parameter for the default `LISTAGG` case. This avoids unnecessary schema overhead while maintaining expected behavior.

fluss-server/src/main/java/org/apache/fluss/server/kv/rowmerger/aggregate/AggregateFieldsProcessor.java
```diff
@@ -0,0 +1,383 @@
...
+     * @return true if at least one non-target column has a non-null value
+     */
+    public static boolean hasNonTargetNonNullField(BinaryRow row, BitSet targetPosBitSet) {
+        for (int pos = 0; pos < row.getFieldCount(); pos++) {
+            if (!targetPosBitSet.get(pos) && !row.isNullAt(pos)) {
+                return true;
+            }
+        }
+        return false;
+    }
```
wuchong, line 384: nit: we can use `targetPosBitSet.nextClearBit(..)` to iterate the non-target fields

```java
for (int i = bs.nextClearBit(0); i <  row.getFieldCount(); i = bs.nextClearBit(i+1)) {
    ...
}
```

fluss-server/src/main/java/org/apache/fluss/server/kv/rowmerger/aggregate/AggregationContext.java
```diff
@@ -0,0 +1,294 @@
...
+            int columnId = schema.getColumns().get(pos).getColumnId();
+            if (targetColumnIdBitSet.get(columnId)) {
+                targetColumnPosBitSet.set(pos);
+            }
+        }
+
+        // Check 1: target columns must contain all primary key columns
+        for (int pkIndex : schema.getPrimaryKeyIndexes()) {
+            if (!targetColumnPosBitSet.get(pkIndex)) {
+                throw new org.apache.fluss.exception.InvalidTargetColumnException(
```
wuchong: replace qualified name with imports

fluss-server/src/main/java/org/apache/fluss/server/kv/rowmerger/aggregate/AggregationContext.java
```diff
@@ -0,0 +1,294 @@
...
+                                "The target write columns must contain the primary key columns %s.",
+                                schema.getColumnNames(schema.getPrimaryKeyIndexes())));
+            }
+        }
+
+        // Check 2: all non-primary key columns must be nullable
+        for (int i = 0; i < fieldDataTypes.length; i++) {
+            if (!primaryKeyColsBitSet.get(i)) {
+                if (!fieldDataTypes[i].isNullable()) {
+                    throw new org.apache.fluss.exception.InvalidTargetColumnException(
```
wuchong: ditto

fluss-server/src/main/java/org/apache/fluss/server/kv/rowmerger/aggregate/AggregationContext.java
```diff
@@ -0,0 +1,294 @@
...
+        }
+
+        // 2. Check Schema for aggregation function (from Column)
+        Optional<AggFunction> schemaAggFunc = schema.getAggFunction(fieldName);
+        if (schemaAggFunc.isPresent()) {
+            return schemaAggFunc.get();
+        }
+
+        // 3. Final fallback - use default aggregation function
+        return AggFunctions.LAST_VALUE_IGNORE_NULLS();
```
wuchong: Can be replaced with a single line:

```java
return schema.getAggFunction(fieldName).orElseGet(AggFunctions::LAST_VALUE_IGNORE_NULLS);
```

fluss-server/src/main/java/org/apache/fluss/server/kv/rowmerger/aggregate/functions/FieldPrimaryKeyAgg.java
```diff
@@ -0,0 +1,40 @@
...
+package org.apache.fluss.server.kv.rowmerger.aggregate.functions;
+
+/* This file is based on source code of Apache Paimon Project (https://paimon.apache.org/), licensed by the Apache
+ * Software Foundation (ASF) under the Apache License, Version 2.0. See the NOTICE file distributed with this work for
+ * additional information regarding copyright ownership. */
+
+import org.apache.fluss.types.DataType;
+
+/** Primary key aggregator - does not aggregate the field, always returns the input field value. */
+public class FieldPrimaryKeyAgg extends FieldAggregator {
```
wuchong: It seems this aggregator is not used, as the primary key fields use `LAST_VALUE` as the agg function. Can we remove this?

fluss-server/src/main/java/org/apache/fluss/server/kv/rowmerger/AggregateRowMerger.java
```diff
@@ -0,0 +1,399 @@
...
+        // This avoids repeated object creation and BitSet construction
+        CacheKey cacheKey = new CacheKey(schemaId, targetColumns);
+        return partialMergerCache.get(
+                cacheKey,
+                k -> {
+                    // Convert target column indices to column IDs
+                    List<Schema.Column> columns = schema.getColumns();
+                    Set<Integer> targetColumnIds = new HashSet<>();
+                    for (int colIdx : targetColumns) {
+                        targetColumnIds.add(columns.get(colIdx).getColumnId());
```
wuchong, line 145: Could you add a TODO on this that the `targetColumns` should already be column ids and not necessary to do the conversion.

cc @loserwang1024 , could you create issue for this? I think we have updated source projection pushdown to use column ids, but forgot to update sink upsert target columns.

fluss-server/src/main/java/org/apache/fluss/server/kv/rowmerger/AggregateRowMerger.java
```diff
@@ -0,0 +1,399 @@
...
+                return false;
+            }
+            if (targetColumns.length != cacheKey.targetColumns.length) {
+                return false;
+            }
+            for (int i = 0; i < targetColumns.length; i++) {
+                if (targetColumns[i] != cacheKey.targetColumns[i]) {
+                    return false;
+                }
+            }
```
wuchong: Simply use `java.util.Arrays#equals(int[], int[])` and will be more efficient.

fluss-server/src/main/java/org/apache/fluss/server/kv/rowmerger/AggregateRowMerger.java
```diff
@@ -0,0 +1,399 @@
...
+        }
+
+        // Use row encoder for aggregation
+        RowEncoder encoder = newContext.getRowEncoder();
+        encoder.startNewRow();
+        AggregateFieldsProcessor.aggregateAllFields(
+                oldValue.row, newValue.row, oldContext, newContext, encoder);
+        BinaryRow mergedRow = encoder.finishRow();
+
+        return new BinaryValue(newValue.schemaId, mergedRow);
```
wuchong: Why not use `targetSchemaId` `RowEncoder` to encode the new row? The `newValue.schemaId` is input data which is defined by client, and is usually outdated.

fluss-server/src/main/java/org/apache/fluss/server/kv/rowmerger/AggregateRowMerger.java
```diff
@@ -0,0 +1,399 @@
...
+            if (oldValue == null || oldValue.row == null) {
+                return newValue;
+            }
+
+            // Get contexts for schema evolution support
+            AggregationContext oldContext = contextCache.getContext(oldValue.schemaId);
+            AggregationContext newContext = contextCache.getContext(newValue.schemaId);
+
+            // Use row encoder for partial update
+            RowEncoder encoder = newContext.getRowEncoder();
```
wuchong: ditto. why not use `RowEncoder` of `targetShemaId`?

fluss-server/src/main/java/org/apache/fluss/server/utils/TableDescriptorValidation.java
```diff
@@ -79,6 +84,25 @@ public static void validateTableDescriptor(TableDescriptor tableDescriptor, int
         // check properties should only contain table.* options,
         // and this cluster know it, and value is valid
         for (String key : tableConf.keySet()) {
+            // Reject old format aggregate function configuration
+            // (table.merge-engine.aggregate.<field-name>)
+            // Allow sub-options (e.g., listagg-delimiter) which are validated when read
+            if (key.startsWith("table.merge-engine.aggregate.")) {
```
wuchong: This can be removed now.

fluss-server/src/test/java/org/apache/fluss/server/kv/rowmerger/aggregate/AggregationContextTest.java
```diff
@@ -0,0 +1,145 @@
...
+    @Test
+    void testPrimaryKeyFieldsNotAggregated() {
+        Schema schema =
+                Schema.newBuilder()
+                        .column("id", DataTypes.INT())
+                        .column("value", DataTypes.BIGINT())
+                        .primaryKey("id")
+                        .build();
+
+        TableConfig tableConfig = new TableConfig(new Configuration());
```
wuchong: nit: can be removed.

fluss-server/src/test/java/org/apache/fluss/server/kv/rowmerger/aggregate/AggregationContextTest.java
```diff
@@ -0,0 +1,145 @@
...
+    void testCompositePrimaryKey() {
+        Schema schema =
+                Schema.newBuilder()
+                        .column("id1", DataTypes.INT())
+                        .column("id2", DataTypes.STRING())
+                        .column("value", DataTypes.BIGINT(), AggFunctions.SUM())
+                        .primaryKey("id1", "id2")
+                        .build();
+
+        TableConfig tableConfig = new TableConfig(new Configuration());
```
wuchong: ditto

fluss-server/src/test/java/org/apache/fluss/server/kv/rowmerger/aggregate/AggregationContextTest.java
```diff
@@ -0,0 +1,145 @@
...
+    void testNoConfigurationUsesLastNonNullValue() {
+        Schema schema =
+                Schema.newBuilder()
+                        .column("id", DataTypes.INT())
+                        .column("value", DataTypes.STRING())
+                        .primaryKey("id")
+                        .build();
+
+        // No configuration at all
+        TableConfig tableConfig = new TableConfig(new Configuration());
```
wuchong: ditto

fluss-server/src/test/java/org/apache/fluss/server/kv/rowmerger/aggregate/AggregationContextTest.java
```diff
@@ -0,0 +1,145 @@
...
+                                DataTypes.STRING(),
+                                AggFunctions.FIRST_VALUE_IGNORE_NULLS())
+                        .column("bool_and_col", DataTypes.BOOLEAN(), AggFunctions.BOOL_AND())
+                        .column("bool_or_col", DataTypes.BOOLEAN(), AggFunctions.BOOL_OR())
+                        .column("listagg_col", DataTypes.STRING(), AggFunctions.LISTAGG())
+                        .column("string_agg_col", DataTypes.STRING(), AggFunctions.STRING_AGG())
+                        .primaryKey("id")
+                        .build();
+
+        TableConfig tableConfig = new TableConfig(new Configuration());
```
wuchong, line 71: ditto

fluss-server/src/test/java/org/apache/fluss/server/kv/rowmerger/aggregate/FieldAggregatorParameterizedTest.java
```diff
@@ -0,0 +1,805 @@
...
+    void testListaggWithCustomDelimiter() {
+        Schema schema =
+                Schema.newBuilder()
+                        .column("id", DataTypes.INT())
+                        .column("value", DataTypes.STRING(), AggFunctions.LISTAGG())
+                        .primaryKey("id")
+                        .build();
+
+        Configuration conf = new Configuration();
+        conf.setString("table.merge-engine.aggregate.value.listagg-delimiter", "|");
```
wuchong: replace to define delimiter in schema

platinumhamburg:
@wuchong Thank you very much for your thorough review and highly valuable feedback, which has greatly improved the code. I’ve incorporated your suggestions through comprehensive adjustments and refactoring, and thanks for review the updated version at your convenience.

fluss-server/src/main/java/org/apache/fluss/server/kv/rowmerger/AggregateRowMerger.java
loserwang1024: I have created a issue for this: https://github.com/apache/fluss/issues/2293

Issue #498, opened 2025-02-26
Support Kerberos Authentication


wuchong:
### Search before asking

- [x] I searched in the [issues](https://github.com/alibaba/fluss/issues) and found nothing similar.

### Motivation

see the discussion: https://github.com/alibaba/fluss/discussions/497

### Willingness to contribute

- [ ] I'm willing to submit a PR!

SML0127:
@wuchong Is this ticket in progress? if not, I'd like to try this. Maybe I should reference flink's [KerberosLoginProvider](https://github.com/apache/flink/blob/master/flink-runtime/src/main/java/org/apache/flink/runtime/security/token/hadoop/KerberosLoginProvider.java) for this

SML0127:
@wuchong Is this ticket in progress? if not, I'd like to try this. Maybe I should reference flink's [KerberosLoginProvider](https://github.com/apache/flink/blob/master/flink-runtime/src/main/java/org/apache/flink/runtime/security/token/hadoop/KerberosLoginProvider.java) for this

wuchong:
Thanks @SML0127 , could you prepare a design doc about how to integration Kerberos to Fluss Authentication framework?

SML0127:
@wuchong 
sorry for the late replay. I will start designing now and share it after done

SML0127:
> Thanks [@SML0127](https://github.com/SML0127) , could you prepare a design doc about how to integration Kerberos to Fluss Authentication framework?

Hi @wuchong
Could I implement this feature based on the design docs (https://docs.google.com/document/d/1Kg1WTWKhR-bsLlYgyIHOIX7NupwAdUkwxj2InhoiHpU/edit?usp=sharing)?

loserwang1024:
Hi, @SML0127 , I am working on sasl/plain now. And pr will be raised in recently two weeks. Maybe you add kerberos authentication later.

SML0127:
Hi @loserwang1024 
Thank you for letting me know. Please mention me when the PR is raised.

loserwang1024:
@SML0127 ，SASL/PLAIN have done, https://github.com/alibaba/fluss/issues/399?issue=alibaba%7Cfluss%7C984

SML0127:
now, start implementation.

wuchong:
Hi @SML0127, thanks for the contribution!

[...]

PR #2263, opened 2025-12-26
[auth] support kerberos authentication

SML0127:
### Purpose

Linked issue: close https://github.com/apache/fluss/issues/498

Support kerberos authentication

### Brief change log

[FIP-7: Support Kerberos vis SASL/GSSAPI](https://cwiki.apache.org/confluence/pages/viewpage.action?pageId=373885589)

### DIFF
--- a/fluss-flink/fluss-flink-common/src/test/java/org/apache/fluss/flink/security/acl/FlinkAuthorizationITCase.java
+++ b/fluss-flink/fluss-flink-common/src/test/java/org/apache/fluss/flink/security/acl/FlinkAuthorizationITCase.java
@@ -17,12 +17,6 @@
 
 package org.apache.fluss.flink.security.acl;
 
-import org.apache.commons.lang3.RandomUtils;
-import org.apache.flink.table.api.EnvironmentSettings;
-import org.apache.flink.table.api.TableEnvironment;
-import org.apache.flink.test.util.AbstractTestBase;
-import org.apache.flink.types.Row;
-import org.apache.flink.util.CollectionUtil;
 import org.apache.fluss.config.ConfigOptions;
 import org.apache.fluss.config.Configuration;
 import org.apache.fluss.config.MemorySize;
@@ -42,6 +36,13 @@ import org.apache.fluss.server.testutils.FlussClusterExtension;
 import org.apache.fluss.utils.FileUtils;
 import org.apache.fluss.utils.ParentResourceBlockingClassLoader;
 import org.apache.fluss.utils.TemporaryClassLoaderContext;
+
+import org.apache.commons.lang3.RandomUtils;
+import org.apache.flink.table.api.EnvironmentSettings;
+import org.apache.flink.table.api.TableEnvironment;
+import org.apache.flink.test.util.AbstractTestBase;
+import org.apache.flink.types.Row;
+import org.apache.flink.util.CollectionUtil;
 import org.apache.hadoop.minikdc.MiniKdc;
 import org.junit.jupiter.api.AfterEach;
 import org.junit.jupiter.api.BeforeAll;


SML0127:
Hi @loserwang1024, Please take a look when you have a time. Thank you

SML0127:
The failure in `testPkCompactedPollFromLatestNoRecords` will be fixed in https://github.com/apache/fluss/pull/2465

SML0127:
Hi @loserwang1024 
Thanks for the feedback. I've added the integration test as requested. Thanks!

SML0127:
Hi @loserwang1024 PTAL.

fluss-common/src/main/java/org/apache/fluss/security/auth/sasl/authenticator/SaslClientAuthenticator.java
```diff
@@ -77,7 +82,9 @@ public String protocol() {
     @Override
     public byte[] authenticate(byte[] data) throws AuthenticationException {
         try {
-            return saslClient.evaluateChallenge(data);
+            return Subject.doAs(
```
loserwang1024, line 91: It's hard to understand why wrapping by `Subject.doAs`. Maybe we can add comment about reason: To bind subject to AccessControlContext so that  can get by Subject.getSubject(AccessController.getContext()) in SaslClient#evaluateChallenge. For example , in kerberos, GssKrb5Client#evaluateChallenge -> GSSContextImpl#initSecContext(java.io.InputStream, java.io.OutputStream)

fluss-common/src/main/java/org/apache/fluss/security/auth/sasl/authenticator/SaslServerAuthenticator.java
```diff
@@ -131,13 +135,16 @@ public void matchProtocol(String protocol) {
...
 
+    @Nullable
     @Override
     public byte[] evaluateResponse(byte[] token) throws AuthenticationException {
         try {
-            return saslServer.evaluateResponse(token);
-        } catch (SaslException e) {
+            return Subject.doAs(
+                    loginManager.subject(),
+                    (PrivilegedExceptionAction<byte[]>) () -> saslServer.evaluateResponse(token));
```
loserwang1024, line 150: add commont as client

fluss-common/src/main/java/org/apache/fluss/security/auth/sasl/gssapi/GssapiServerCallbackHandler.java
```diff
@@ -0,0 +1,126 @@
...
+    /** Checks if the authenticated user has the permission to act as the authorization user. */
+    private boolean isAuthorizedActAs(String authnId, String authzId) {
+        // Default policy: allow the authenticated user to act as themselves.
+        // 1. Exact match
+        if (Objects.equals(authnId, authzId)) {
+            return true;
+        }
+
+        // 2. Check if authnId is "user@REALM" and authzId is "user" or "user@REALM"
+        String shortAuthnId = getShortName(authnId);
```
loserwang1024, line 110: Would you like to tell me the what is authenticated user and authorization user and what's their relation? In your code, it seems that if authnId = user@REALM while authnId = user@NON_REALM, then authnId can act as authzId?

fluss-flink/fluss-flink-common/src/test/java/org/apache/fluss/flink/security/acl/FlinkAuthorizationITCase.java
```diff
@@ -132,6 +141,152 @@ void after() throws ExecutionException, InterruptedException {
...
+
+                String clientJaas =
+                        String.format(
+                                "com.sun.security.auth.module.Krb5LoginModule required "
+                                        + "useKeyTab=true storeKey=true useTicketCache=false "
+                                        + "keyTab=\"%s\" principal=\"%s\";",
+                                keytab.getAbsolutePath(), clientPrincipal);
+
+                // Create a Flink catalog configured with Kerberos credentials.
+                String createCatalogDDL =
```
loserwang1024, line 253: would your like to also test what if the client.security.sasl.jaas.config is not valid for server?  Add we can test with client.security.kerberos.keytab and client.security.kerberos.principal

fluss-common/src/main/java/org/apache/fluss/security/auth/sasl/authenticator/SaslClientAuthenticator.java
SML0127: Thank! I will add some comments as you suggested. 
It's definitely helpful to explain AccessControlContext binding for future readers.

fluss-common/src/main/java/org/apache/fluss/security/auth/sasl/authenticator/SaslServerAuthenticator.java
SML0127: Sure, will add it here as well.

fluss-common/src/main/java/org/apache/fluss/security/auth/sasl/gssapi/GssapiServerCallbackHandler.java
SML0127: Actually, I referenced Kafka's impelementation.

It seems Kafka uses short name mapping strategy by default. (stripping the realm, so non-realm user can act as realm user). I thought following this pattern would be a good practice for compatibility. What do you think about this approach?

fluss-flink/fluss-flink-common/src/test/java/org/apache/fluss/flink/security/acl/FlinkAuthorizationITCase.java
SML0127: Sure, I will add those cases.

loserwang1024:
Look good to me now! @wuchong , WDYT?

SML0127:
Hi @wuchong PTAL

PR #2265, opened 2025-12-27
[FLUSS-2262][lake] Improved Stability for Iceberg Log Table Compaction Test

### LINKED ISSUE #2262: Unstable test IcebergRewriteITCase.testLogTableCompaction
### Search before asking

### Fluss version

0.8.0 (latest release)

### Please describe the bug 🐞

ci link: https://github.com/apache/fluss/actions/runs/20522062759/job/58958890222?pr=2179

### Solution

_No response_

rionmonster:
### Purpose

Linked issue: close #2262  

Per Issue https://github.com/apache/fluss/issues/2262, this pull request addresses a race condition that could sometimes occur and result in the `IcebergRewriteITCase.testLogTableCompaction` test case failing (particularly during CI builds). 

### Brief change log

This change updates the `LakeTableTieringManager` to address potential race conditions surrounding state transitions. In some cases, delayed callbacks or retries could act on tables before their new state was recorded, or stale entries in the pending queue could be handed out even though the table was no longer eligible for tiering. Some additional changes were made to improve table requesting/removal use-cases as well.

### Tests

These changes were initially verified after successfully reproducing the issue (via iterating over the `IcebergRewriteITCase.testLogTableCompaction` test). After the root cause was identified, new tests were introduced to help verify these issues:
- `LakeTableTieringManagerTest.testRequestTableSkipsStaleQueueEntries` - Verifies the changes made to `LakeTableTieringManager.requestTable` to ensure that invalid or stale state transitions would not negatively affect the tiering process.
- `LakeTableTieringManagerTest.testScheduledStateRecordedBeforeTimerCallbackRuns` - Verifies any potential race-conditions related to immediate state changes to ensure no side-effects are created.
- `LakeTableTieringManagerTest.testRemoveLakeTableRemovesAllPendingQueueOccurrences` - Verifies that any potential duplicate entries within the internal `pendingTieringTables` queue are removed during table removal (via `removeLakeTable()`.

### API and Format

N/A

### Documentation

N/A

### Reviewer(s) Requested

@swuferhong (as original reporter), @beryllw (as original author)

### DIFF
--- a/fluss-filesystems/fluss-fs-azure/src/main/resources/META-INF/NOTICE
+++ b/fluss-filesystems/fluss-fs-azure/src/main/resources/META-INF/NOTICE
@@ -32,11 +32,11 @@ This project bundles the following dependencies under the Apache Software Licens
 - org.apache.commons:commons-configuration2:2.8.0
 - org.apache.commons:commons-lang3:3.18.0
 - org.apache.commons:commons-text:1.10.0
-- org.apache.hadoop.thirdparty:hadoop-shaded-guava:1.2.0
+- org.apache.hadoop.thirdparty:hadoop-shaded-guava:1.1.1
 - org.apache.hadoop.thirdparty:hadoop-shaded-protobuf_3_21:1.2.0
 - org.apache.hadoop:hadoop-annotations:3.4.0
 - org.apache.hadoop:hadoop-auth:3.4.0
-- org.apache.hadoop:hadoop-azure:3.4.0
+- org.apache.hadoop:hadoop-azure:3.3.4
 - org.apache.hadoop:hadoop-common:3.4.0
 - org.apache.kerby:kerb-core:2.0.3
 - org.apache.kerby:kerby-asn1:2.0.3
@@ -44,8 +44,10 @@ This project bundles the following dependencies under the Apache Software Licens
 - org.apache.kerby:kerby-util:2.0.3
 - org.apache.httpcomponents:httpclient:4.5.13
 - org.apache.httpcomponents:httpcore:4.4.13
+- org.codehaus.jackson:jackson-core-asl:1.9.13
+- org.codehaus.jackson:jackson-mapper-asl:1.9.13
 - org.codehaus.jettison:jettison:1.5.4
-- org.wildfly.openssl:wildfly-openssl:1.1.3.Final
+- org.wildfly.openssl:wildfly-openssl:1.0.7.Final
 - org.xerial.snappy:snappy-java:1.1.10.4
 
 This project bundles the following dependencies under the BSD license. See bundled license files for details.
@@ -76,5 +78,5 @@ This project bundles the following dependencies under the Eclipse Public License
 - EPL-2.0: https://www.eclipse.org/legal/epl-2.0/
 - Apache-2.0: http://www.apache.org/licenses/LICENSE-2.0.txt
 
-- org.eclipse.jetty:jetty-util-ajax:9.4.53.v20231009
-- org.eclipse.jetty:jetty-util:9.4.53.v20231009
+- org.eclipse.jetty:jetty-util-ajax:9.4.43.v20210629
+- org.eclipse.jetty:jetty-util:9.4.43.v20210629
--- a/fluss-filesystems/fluss-fs-gs/pom.xml
+++ b/fluss-filesystems/fluss-fs-gs/pom.xml
@@ -82,10 +82,6 @@
                     <artifactId>jersey-server</artifactId>
                 </exclusion>
                 <exclusion>
-                    <groupId>com.github.pjfanning</groupId>
-                    <artifactId>jersey-json</artifactId>
-                </exclusion>
-                <exclusion>
                     <groupId>org.apache.avro</groupId>
                     <artifactId>avro</artifactId>
                 </exclusion>
--- a/fluss-filesystems/fluss-fs-gs/src/main/resources/META-INF/NOTICE
+++ b/fluss-filesystems/fluss-fs-gs/src/main/resources/META-INF/NOTICE
@@ -5,7 +5,7 @@ This project bundles the following dependencies under the Apache Software Licens
 - com.fasterxml.jackson.core:jackson-annotations:2.15.3
 - com.fasterxml.jackson.core:jackson-core:2.15.3
 - com.fasterxml.jackson.core:jackson-databind:2.15.3
-- com.fasterxml.woodstox:woodstox-core:5.4.0
+- com.fasterxml.woodstox:woodstox-core:5.3.0
 - com.google.android:annotations:4.1.1.4
 - com.google.api-client:google-api-client-jackson2:2.0.1
 - com.google.api-client:google-api-client:2.2.0
@@ -45,9 +45,8 @@ This project bundles the following dependencies under the Apache Software Licens
 - commons-beanutils:commons-beanutils:1.9.4
 - commons-codec:commons-codec:1.16.0
 - commons-collections:commons-collections:3.2.2
-- commons-io:commons-io:2.14.0
-- commons-logging:commons-logging:1.2
-- io.dropwizard.metrics:metrics-core:3.2.4
+- commons-io:commons-io:2.8.0
+- commons-logging:commons-logging:1.1.3
 - io.grpc:grpc-alts:1.59.1
 - io.grpc:grpc-api:1.59.1
 - io.grpc:grpc-auth:1.59.1
@@ -65,15 +64,6 @@ This project bundles the following dependencies under the Apache Software Licens
 - io.grpc:grpc-stub:1.59.1
 - io.grpc:grpc-util:1.59.1
 - io.grpc:grpc-xds:1.59.1
-- io.netty:netty-buffer:4.1.100.Final
-- io.netty:netty-codec:4.1.100.Final
-- io.netty:netty-common:4.1.100.Final
-- io.netty:netty-handler:4.1.100.Final
-- io.netty:netty-resolver:4.1.100.Final
-- io.netty:netty-transport-classes-epoll:4.1.100.Final
-- io.netty:netty-transport-native-epoll:4.1.100.Final
-- io.netty:netty-transport-native-unix-common:4.1.100.Final
-- io.netty:netty-transport:4.1.100.Final
 - io.opencensus:opencensus-api:0.31.1
 - io.opencensus:opencensus-contrib-exemplar-util:0.31.0
 - io.opencensus:opencensus-contrib-grpc-metrics:0.31.0
@@ -85,23 +75,21 @@ This project bundles the following dependencies under the Apache Software Licens
 - io.opencensus:opencensus-impl-core:0.31.0
 - io.opencensus:opencensus-proto:0.2.0
 - io.perfmark:perfmark-api:0.26.0
-- org.apache.commons:commons-compress:1.24.0
-- org.apache.commons:commons-configuration2:2.8.0
+- org.apache.commons:commons-compress:1.21
+- org.apache.commons:commons-configuration2:2.1.1
 - org.apache.commons:commons-lang3:3.18.0
-- org.apache.commons:commons-text:1.10.0
-- org.apache.hadoop.thirdparty:hadoop-shaded-guava:1.2.0
-- org.apache.hadoop.thirdparty:hadoop-shaded-protobuf_3_21:1.2.0
-- org.apache.hadoop:hadoop-annotations:3.4.0
-- org.apache.hadoop:hadoop-auth:3.4.0
-- org.apache.hadoop:hadoop-common:3.4.0
+- org.apache.commons:commons-text:1.4
+- org.apache.hadoop.thirdparty:hadoop-shaded-guava:1.1.1
+- org.apache.hadoop.thirdparty:hadoop-shaded-protobuf_3_7:1.1.1
+- org.apache.hadoop:hadoop-annotations:3.3.4
+- org.apache.hadoop:hadoop-auth:3.3.4
+- org.apache.hadoop:hadoop-common:3.3.4
 - org.apache.httpcomponents:httpclient:4.5.13
 - org.apache.httpcomponents:httpcore:4.4.13
-- org.apache.kerby:kerb-core:2.0.3
-- org.apache.kerby:kerby-asn1:2.0.3
-- org.apache.kerby:kerby-pkix:2.0.3
-- org.apache.kerby:kerby-util:2.0.3
-- org.bouncycastle:bcprov-jdk15on:1.70
-- org.codehaus.jettison:jettison:1.5.4
+- org.apache.kerby:kerb-core:1.0.1
+- org.apache.kerby:kerby-asn1:1.0.1
+- org.apache.kerby:kerby-pkix:1.0.1
+- org.apache.kerby:kerby-util:1.0.1
 - org.conscrypt:conscrypt-openjdk-uber:2.5.2
 - org.xerial.snappy:snappy-java:1.1.10.4
 
@@ -113,7 +101,7 @@ See bundled license files for details.
 This project bundles the following dependencies under BSD-2 License (https://opensource.org/licenses/BSD-2-Clause).
 See bundled license files for details.
 
-- dnsjava:dnsjava:3.4.0
+- dnsjava:dnsjava:2.1.7
 
 This project bundles the following dependencies under BSD-3 License (https://opensource.org/licenses/BSD-3-Clause).
 See bundled license files for details.
--- a/fluss-filesystems/fluss-fs-hadoop-shaded/src/main/resources/META-INF/NOTICE
+++ b/fluss-filesystems/fluss-fs-hadoop-shaded/src/main/resources/META-INF/NOTICE
@@ -9,40 +9,28 @@ This project bundles the following dependencies under the Apache Software Licens
 - com.fasterxml.jackson.core:jackson-annotations:2.15.3
 - com.fasterxml.jackson.core:jackson-core:2.15.3
 - com.fasterxml.jackson.core:jackson-databind:2.15.3
-- com.fasterxml.woodstox:woodstox-core:5.4.0
+- com.fasterxml.woodstox:woodstox-core:5.3.0
 - com.google.guava:failureaccess:1.0
 - com.google.guava:guava:27.0-jre
 - com.google.guava:listenablefuture:9999.0-empty-to-avoid-conflict-with-guava
 - com.google.j2objc:j2objc-annotations:1.1
 - commons-beanutils:commons-beanutils:1.9.4
 - commons-collections:commons-collections:3.2.2
-- commons-io:commons-io:2.14.0
-- commons-logging:commons-logging:1.2
-- io.dropwizard.metrics:metrics-core:3.2.4
-- io.netty:netty-buffer:4.1.100.Final
-- io.netty:netty-codec:4.1.100.Final
-- io.netty:netty-common:4.1.100.Final
-- io.netty:netty-handler:4.1.100.Final
-- io.netty:netty-resolver:4.1.100.Final
-- io.netty:netty-transport-classes-epoll:4.1.100.Final
-- io.netty:netty-transport-native-epoll:4.1.100.Final
-- io.netty:netty-transport-native-unix-common:4.1.100.Final
-- io.netty:netty-transport:4.1.100.Final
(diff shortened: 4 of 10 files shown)


fluss-lake/fluss-lake-iceberg/src/test/java/org/apache/fluss/lake/iceberg/maintenance/IcebergRewriteITCase.java
```diff
@@ -186,6 +186,9 @@ void testLogTableCompaction() throws Exception {
                             t1, t1Bucket, ++i, true, Collections.singletonList(row(1, "v1"))));
             checkFileStatusInIcebergTable(t1, 3, false);
 
+            // Ensure tiering job has fully processed the previous writes
+            assertReplicaStatus(t1Bucket, i);
```
luoyuxia: I'm curious about it. If the files count already is 3 in iceberg, shouldn't it also mean that the tiering has already tiered all records since we only write 3 records.

luoyuxia (commented):
@rionmonster Thanks for the pr. Only one question

fluss-lake/fluss-lake-iceberg/src/test/java/org/apache/fluss/lake/iceberg/maintenance/IcebergRewriteITCase.java
rionmonster: @luoyuxia 

I agree that it’s a bit odd. The additional replica assertion seems to alleviate any race conditions within this specific test case that we were seeing arise within the failing CI instances (at least based on my local testing to reproduce the issue).

It may simply be a testing artifact as opposed to a legitimate issue, but if it doesn’t seem like a fix, we can explore a few other avenues. It seemed like one of those common race conditions, so leveraging an existing function to help seemed like a decent approach. Happy to explore some additional avenues though, if we feel that we need a bit more exhaustive checks.

luoyuxia: @rionmonster I tried to reporduce it in my local. But fails. I think it'll be better to explore the root cause for I'm afraid of it is caused by another critical issue.

rionmonster: @luoyuxia 

Good to know! It seemed to be pretty consistent regarding pass/fail on my end, but clearly there’s something else at play.

I’ll do some more exploration and see what I find. Thanks for the feedback!

rionmonster: @luoyuxia 

I've been digging into this a bit further and it seems there's some disparity between the actual files being written and retrieving the latest offsets for those (specifically after any types of asynchronous operations, such as compaction, etc.). I think we may need some mechanism to improve the consistency, at least within the bounds of the tests.

I wrote a little monitor to run through several iterations of the tests to see what the state of various bits looked like during each iteration. Here's a sample of those:

**Successful Test (Successfully Performs Compaction as Expected)**
[MONITOR] log_table_33 - TableBucket{tableId=32, bucket=0}
| IcebergFiles | IcebergSnapshotId | LakeSnapshotId | ReplicaLakeSnapId | ReplicaLakeLogEndOff | Timestamp |
|--------------|-------------------|----------------|-------------------|----------------------|-----------|
| 0            | -1                | -1             | -1                | -1                   | 1767113560800 |
| 0            | -1                | -1             | -1                | -1                   | 1767113561046 |
| ...           | ...                | ...              | ...                | ...                   | ...                        |
| 3            | 5182733673261799288 | 5182733673261799288 | 5443797100773076340 | 3                    | 1767113615059 |
| 3            | 5182733673261799288 | 5182733673261799288 | 5443797100773076340 | 3                    | 1767113615307 |
| 2            | 2575057976625237982 | 2575057976625237982 | 5443797100773076340 | 4                    | 1767113615557 |
| 2            | 2575057976625237982 | 2575057976625237982 | 5443797100773076340 | 4                    | 1767113615808 |

**Failing Test (File appeared to never be properly written before expected offset)**
[MONITOR] log_table_34 - TableBucket{tableId=33, bucket=0}
| IcebergFiles | IcebergSnapshotId | LakeSnapshotId | ReplicaLakeSnapId | ReplicaLakeLogEndOff | Timestamp |
|--------------|-------------------|----------------|-------------------|----------------------|-----------|
| 0            | -1                | -1             | -1                | -1                   | 1767113616327 |
...           | ...                | ...              | ...                | ...                   | ...                        | ... |
| 2            | 7273969972093574431 | 7273969972093574431 | 7273969972093574431 | 2                    | 1767113861627 |
| 2            | 7273969972093574431 | 7273969972093574431 | 7273969972093574431 | 2                    | 1767113861882 |
| 2            | 7273969972093574431 | 7273969972093574431 | 7273969972093574431 | 2                    | 1767113862135 |
| 2            | 7273969972093574431 | 7273969972093574431 | 7273969972093574431 | 2                    | 1767113862381 |
| 2            | 7273969972093574431 | 7273969972093574431 | 7273969972093574431 | 2                    | 1767113862633 |

```
[ASSERTION FAILURE] Expected offset 3 but got 2 for bucket TableBucket{tableId=33, bucket=0}
  Replica Lake Snapshot ID: 7273969972093574431
  Current State:
    Iceberg Files: 2
    Iceberg Snapshot ID: 7273969972093574431
    Lake Snapshot ID (from admin): 7273969972093574431
    Replica Lake Snapshot ID: 7273969972093574431
    Replica Lake Log End Offset: 2
```

I'm not sure if this is more of an artifact of the tests themselves or a legitimate issue. Any thoughts? Happy to continue digging. It _feels_ like a race-condition due to the inconsistency, either on writing the files or reading stale offsets from the data lake directly. I'm sure in a real-world environment, this might be tolerable (as I suspect it's just a minor latency spike which would eventually resolve), but in the confines of a test, it's flaky.

rionmonster: @luoyuxia 

I added some additional checks around these to try and better identify the underlying issues. It doesn't look like the files are being written or recognized by Iceberg, which I suspect is simply a race-condition from the additional diagnostics:

```
[ASSERTION FAILURE] Expected offset 3 but got 2 for bucket TableBucket{tableId=21, bucket=0}
  Replica Lake Snapshot ID: 6090342046054561204
  Current State:
    Iceberg Files: 2
    Iceberg Snapshot ID: 6090342046054561204
    Lake Snapshot ID (from admin): 6090342046054561204
    Replica Lake Snapshot ID: 6090342046054561204
    Replica Lake Log End Offset: 2 (expected: 3, diff: 1)
```

Said differently, because the replica offset is off (by one) it means either the tiering job hasn't processed all writes yet (check monitor output above), the offset update is lagging behind file commits (race condition), or the tiering job itself may be stuck or running slowly.

I think we could help resolve this in two ways, which I'm testing through now and will update the PR:
1. Introduce a `waitForTieringToProcess` helper function that would check to verify that the tiering had been executed before continuing to validate the replica statuses.
2. Reduce the test-specific freshness and/or tiering intervals (e.g., `POLL_TIERING_TABLE_INTERVAL`) from their current values of 500ms to 100ms.

Thoughts? I’d probably err on the side of applying both since it’s a flaky test and the more consistency, the better.

rionmonster: @luoyuxia 

I've been tinkering with a few different approaches to handling this and it appears to entirely be related to the asynchronous processes in place (e.g., tiering, datalake freshness, etc.). I tried to avoid introducing any explicit delays such as `Thread.sleep(...)` calls as those generally feel more like addressing an underlying symptom as opposed to the cause.

I just recently ran a test through 500 and 1000 iterations respectively with the `POLL_TIERING_TABLE_INTERVAL` and `TABLE_DATALAKE_FRESHNESS` set to 100ms (from their original 500ms values). This appeared to address the flakiness, although I suppose with all these asynchronous processes, it’s still technically possible for them to fail.

Do this think this would be sufficient or are you aware of any additional configurations that we might be able to use within the context of this test to force flushes to help with consistency in some of these asynchronous processes?

luoyuxia: Thanks @rionmonster for diving into it and sorry for late reply for i have a busy week. I download the log for the fail ci in https://github.com/apache/fluss/actions/runs/20522062759/artifacts/4970768099
And I found the following logs:
```
--------- first tiering
20:09:10,300 [SourceCoordinator-Source: TieringSource] INFO  org.apache.fluss.flink.tiering.source.enumerator.TieringSourceEnumerator [] - Generate Tiering splits for table fluss.log_table.
20:09:10,310 [SourceCoordinator-Source: TieringSource] INFO  org.apache.fluss.flink.tiering.source.enumerator.TieringSourceEnumerator [] - **Generate Tiering 1** splits for table fluss.log_table with cost 10ms.
20:09:10,310 [Source: TieringSource (1/2)#0] INFO  org.apache.flink.connector.base.source.reader.SourceReaderBase [] - Adding split(s) to reader: [TieringLogSplit{tablePath=fluss.log_table, tableBucket=TableBucket{tableId=2, bucket=0}, 

---- second  tiering
org.apache.fluss.flink.tiering.source.enumerator.TieringSourceEnumerator [] - Generate Tiering splits for table fluss.log_table.
20:09:11,303 [SourceCoordinator-Source: TieringSource] INFO  org.apache.fluss.flink.tiering.source.split.TieringSplitGenerator [] - Last committed lake table snapshot info is:LakeSnapshot{snapshotId=3828565410144101634, tableBucketsOffset={TableBucket{tableId=2, bucket=0}=1}}
20:09:11,307 [SourceCoordinator-Source: TieringSource] INFO  org.apache.fluss.flink.tiering.source.enumerator.TieringSourceEnumerator [] - Generate Tiering 1 splits for table fluss.log_table with cost 7ms.
20:09:11,307 [Source: TieringSource (1/2)#0] INFO  org.apache.flink.connector.base.source.reader.SourceReaderBase [] - Adding split(s) to reader: [TieringLogSplit{tablePath=fluss.log_table, tableBucket=TableBucket{tableId=2, bucket=0}, partitionName='null', startingOffset=1, stoppingOffset=2, numberOfSplits=1}]
20:09:11,307 [Source Data Fetcher for Source: TieringSource (1/2)#0] INFO  org.apache.flink.connector.base.source.reader.fetcher.SplitFetcher [] - Starting split fetcher 1
20:09:11,307 [Source Data Fetcher for Source: TieringSource (1/2)#0] INFO  org.apache.fluss.flink.tiering.source.TieringSplitReader     [] - add split tiering-log-split-2-0
20:09:11,310 [Source Data Fetcher for Source: TieringSource (1/2)#0] INFO  org.apache.fluss.flink.tiering.source.TieringSplitReader     [] - Start to tier table fluss.log_table with table id 2.
20:09:11,311 [Source Data Fetcher for Source: TieringSource (1/2)#0] INFO  org.apache.fluss.flink.tiering.source.TieringSplitReader     [] - Subscribe to read log for split tiering-log-split-2-0 from starting offset 1 to end offset 2.

---- third  tiering

org.apache.fluss.flink.tiering.source.enumerator.TieringSourceEnumerator [] - Generate Tiering splits for table fluss.log_table.
20:09:12,303 [SourceCoordinator-Source: TieringSource] INFO  org.apache.fluss.flink.tiering.source.split.TieringSplitGenerator [] - Last committed lake table snapshot info is:LakeSnapshot{snapshotId=7991116837264656570, tableBucketsOffset={TableBucket{tableId=2, bucket=0}=2}}
20:09:12,307 [SourceCoordinator-Source: TieringSource] INFO  org.apache.fluss.flink.tiering.source.enumerator.TieringSourceEnumerator [] - Generate Tiering 1 splits for table fluss.log_table with cost 7ms.
20:09:12,308 [Source: TieringSource (1/2)#0] INFO  org.apache.flink.connector.base.source.reader.SourceReaderBase [] - Adding split(s) to reader: [TieringLogSplit{tablePath=fluss.log_table, tableBucket=TableBucket{tableId=2, bucket=0}, partitionName='null', startingOffset=2, stoppingOffset=3, numberOfSplits=1}]
20:09:12,308 [Source Data Fetcher for Source: TieringSource (1/2)#0] INFO  org.apache.flink.connector.base.source.reader.fetcher.SplitFetcher [] - Starting split fetcher 2
20:09:12,308 [Source Data Fetcher for Source: TieringSource (1/2)#0] INFO  org.apache.fluss.flink.tiering.source.TieringSplitReader     [] - add split tiering-log-split-2-0
20:09:12,311 [Source Data Fetcher for Source: TieringSource (1/2)#0] INFO  org.apache.fluss.flink.tiering.source.TieringSplitReader     [] - Start to tier table fluss.log_table with table id 2.
20:09:12,311 [Source Data Fetcher for Source: TieringSource (1/2)#0] INFO  org.apache.fluss.flink.tiering.source.TieringSplitReader     [] - Subscribe to read log for split tiering-log-split-2-0 from starting offset 2 to end offset 3.


--- then no tiering again, that's where the problem happen

20:09:14,300 [SourceCoordinator-Source: TieringSource-worker-thread-1] INFO  org.apache.fluss.flink.tiering.source.enumerator.TieringSourceEnumerator [] - No available Tiering table found, will poll later.
20:09:14,800 [SourceCoordinator-Source: TieringSource-worker-thread-1] INFO  org.apache.fluss.flink.tiering.source.enumerator.TieringSourceEnumerator [] - No available Tiering table found, will poll later.
....

20:10:19,800 [SourceCoordinator-Source: TieringSource-worker-thread-1] INFO  org.apache.fluss.flink.tiering.source.enumerator.TieringSourceEnumerator [] - No available Tiering table found, will poll later.

**no any tiering happen**
```

I think the keypoint is that why after the third tiering which tiering up to offset 3(3 records), no tiering happen. 
It's expected to happen another tiering since we write the 4th record in 
```
 flussRows.addAll(
                    writeIcebergTableRecords(
                            t1, t1Bucket, ++i, true, Collections.singletonList(row(1, "v1"))));
```

It'll be better that we figure out the reason. Since you can reproduce it in your local, I think you can print more logs  in `LakeTableTieringManager` to find the it doesn't assign the table to tiering service anymore. 

I'm afarid that there is some bug in `LakeTableTieringManager` or the logic of assign the table to tier.

rionmonster: @luoyuxia 

You were right! After adding more logs than I would like to mention and performing some rituals to the debugging gods, I was able to verify the issue, which seemed to be centered around the state transitions themselves.

It appears that during the processes for requesting tables in `LakeTableTieringManager.requestTable()` was relying on the existing pending queue (`pendingTieringTables`) to serve the appropriate table. Because pending entries could be duplicated or stale (e.g., timers firing late, retries, races between state transitions, etc.), it was possible to request a table from the queue that was no longer actually in the "Pending" state. We’d then try to move it to Tiering, fail the state transition, but still hand it out anyway — which could confuse the tiering service and lead to flaky behavior under repetition.

Separately, some state transition side effects (like scheduling delayed tiering) could run asynchronously and request the state of the table before its new state was recorded, causing valid transitions to be rejected if a timer fired early.

I'm wrapping up verification now, but should have an updated PR coming soon.

rionmonster: @luoyuxia 

PR pushed (along with two new related tests). Let me know what you think and thanks again for all of the feedback!

luoyuxia: Cool. I will have a look when i'm free

luoyuxia: >  it was possible to request a table from the queue that was no longer actually in the "Pending" state. 

@rionmonster 
Could you please explain a bit about in which case the table in the queue that was no longer actually in the "Pending" state.? Although it's no in "Pending" state, the tiering service still can requst the table and do the tiering. Also, from the log, I can't see the log like "Fail to change state for table ".
 
To me, it looks like that the pendingTieringTables is empty, but I still can't figure out why

luoyuxia: Also, could you please share me your reproduce branch so that I can reproduce it in my local env.

rionmonster: @luoyuxia 

Sure! Cleaning up some of my notes from the troubleshooting session, but hopefully this workflow makes sense:

1. A table `T` becomes eligible and is assigned normally (Pending to Tiering).
2. Later, a stale duplicate entry for `T` ends up in the pending queue (e.g., due to retries, late timer callbacks, or test churn).
3. Before the fix, `requestTable()` would poll `T` from the pending queue and attempt to assign it again, even though its actual state was no longer Pending.
4. That resulted in an invalid state transition attempt (non-Pending/Tiering to Tiering), but the table was still returned to the tiering service, causing things to become inconsistent between the `pendingTieringTables` and `tieringStates`.

The key difference with this change is that after it `requestTable()` actually verifies the table is in the Pending state before assigning it instead of just relying on its presence in the pending queue.

luoyuxia: @rionmonster 
> Later, a stale duplicate entry for T ends up in the pending queue (e.g., due to retries, late timer callbacks, or test churn).

As for the fail test, which case cause a  stale duplicate T entering into pending queue? Do you have ever find any logs?

> would poll T from the pending queue and attempt to assign it again, even though its actual state was no longer Pending

If it's a invalid state changte, we'll print 
```
Fail to change state for table xxx
```
But I can't see such log in the fail ci 

> but the table was still returned to the tiering service, causing things to become inconsistent between the pendingTieringTables and tieringStates.

To me, it looks me like that the table won't be returned to the tiering service anymore from the log.

So, i'm a little doubt about this case.

 Could you please share me your reproduce branch, I want reproduce it in my local env to find the root cause.

rionmonster: @luoyuxia 

Based on our private conversations exploring this, I'm including the following based on 

Okay — I was able to reproduce the failure successfully with some extended debugging logs added throughout the `LakeTableTieringManager`. I've [created a gist here](https://gist.github.com/rionmonster/6b25bf7def8ac39f2cd6d112ccc02b6c) with the logs.  Reviewing over them, it looks like we have the following chain of events:
- At `411079` we have a series of these "polled tableId=x but tablePath is null (state=null, epoch=null)" type of calls, which repeats 10+ times
- After this burst we can see that the test fails (since rewrite/compaction didn't complete in time)

I think we can interpret this as:
- Tiering service requested work (via `requestTable()`)
- Manager pulled table ids from pendingTieringTables but one of the following was true:
    - `tablePaths` didn't contain the id
    - `tieringStates` didn't contain the id
    - `tableTierEpoch` didn't contain the id
- This means that the table ids were no longer registered (e.g., previous dropped/removed) but were still present in the `pendingTieringTables`
- This causes any `requestTable()` calls to continually drain/recurse/loop over these stale entries which could delay processing of actual pending tables
- This delay could be enough to cause the failed assertion

I think the proposed fixes would help with this, although we may need to consider adjusting the `removeLakeTable()` call as well to ensure we remove the requested table from the pending tables as well, which I'll add to the PR:
```
public void removeLakeTable(long tableId) {
    inLock(lock, () -> {
        // Omitted for brevity

        pendingTieringTables.remove(tableId);
    });
}
```

I've created a branch with these additional logs if you would like to explore it yourself at https://github.com/rionmonster/fluss/tree/for-yuxia-with-logs

rionmonster:
Closing as I believe a separate root cause was found, will issue a separate PR.

PR #2277, opened 2025-12-29
[spark] support batch write

YannByron:
### Purpose

to support spark batch write.

Linked issue: close #xxx

### Tests

org.apache.fluss.spark.row.SparkAsFlussRowTest
org.apache.fluss.spark.row.SparkAsFlussArrayTest
org.apache.fluss.spark.SparkWriteTest

### DIFF
--- a/fluss-spark/fluss-spark-ut/src/test/scala/org/apache/fluss/spark/SparkWriteTest.scala
+++ b/fluss-spark/fluss-spark-ut/src/test/scala/org/apache/fluss/spark/SparkWriteTest.scala
@@ -131,29 +131,18 @@ class SparkWriteTest extends FlussSparkTestBase {
     // update data
     sql(s"""
            |INSERT INTO $DEFAULT_DATABASE.$pkTableName
-           |VALUES (800L, 230L, 603, "addr3")
+           |VALUES (800L, 230L, 603, "addr3"), (900L, 240L, 604, "addr4")
            |""".stripMargin)
 
     val flussRows2 = getRowsWithChangeType(table, Some(logScanner))
     val expectRows2 = Array(
       ("-U", createGenericRow(800L, 23L, 603, BinaryString.fromString("addr3"))),
-      ("+U", createGenericRow(800L, 230L, 603, BinaryString.fromString("addr3")))
-    )
-    assertThat(flussRows2.length).isEqualTo(2)
-    assertThat(flussRows2).containsExactlyElementsOf(expectRows2.toIterable.asJava)
-
-    sql(s"""
-           |INSERT INTO $DEFAULT_DATABASE.$pkTableName
-           |VALUES (900L, 240L, 604, "addr4")
-           |""".stripMargin)
-
-    val flussRows3 = getRowsWithChangeType(table, Some(logScanner))
-    val expectRows3 = Array(
+      ("+U", createGenericRow(800L, 230L, 603, BinaryString.fromString("addr3"))),
       ("-U", createGenericRow(900L, 24L, 604, BinaryString.fromString("addr4"))),
       ("+U", createGenericRow(900L, 240L, 604, BinaryString.fromString("addr4")))
     )
-    assertThat(flussRows3.length).isEqualTo(2)
-    assertThat(flussRows3).containsExactlyElementsOf(expectRows3.toIterable.asJava)
+    assertThat(flussRows2.length).isEqualTo(4)
+    assertThat(flussRows2).containsAll(expectRows2.toIterable.asJava)
   }
 }
 


fluss-spark/fluss-spark-common/src/main/scala/org/apache/fluss/spark/row/SparkAsFlussArray.scala
```diff
@@ -0,0 +1,131 @@
...
+  }
+
+  /**
+   * Returns the timestamp value at the given position.
+   *
+   * <p>The precision is required to determine whether the timestamp value was stored in a compact
+   * representation (see {@link TimestampNtz}).
+   */
+  override def getTimestampNtz(pos: Int, precision: Int): TimestampNtz =
+    TimestampNtz.fromMillis(SparkDateTimeUtils.microsToMillis(arrayData.getLong(pos)))
```
wuchong: I think we can introduce a method `TimestampNtz.fromMicros` like `TimestampLtz.fromEpochMicros` to suppoert convert from micro to `TimestampNtz`. Converting from micro to millis will lose the nano precisions.

fluss-spark/fluss-spark-common/src/main/scala/org/apache/fluss/spark/row/SparkAsFlussRow.scala
```diff
@@ -0,0 +1,129 @@
...
+  }
+
+  /**
+   * Returns the timestamp value at the given position.
+   *
+   * <p>The precision is required to determine whether the timestamp value was stored in a compact
+   * representation (see {@link TimestampNtz}).
+   */
+  override def getTimestampNtz(pos: Int, precision: Int): TimestampNtz =
+    TimestampNtz.fromMillis(SparkDateTimeUtils.microsToMillis(row.getLong(pos)))
```
wuchong: ditto

fluss-spark/fluss-spark-common/src/main/scala/org/apache/fluss/spark/row/SparkAsFlussArray.scala
```diff
@@ -0,0 +1,131 @@
...
+
+package org.apache.fluss.spark.row
+
+import org.apache.fluss.row.{BinaryString, Decimal, InternalArray => FlussInternalArray, InternalRow => FlussInternalRow, TimestampLtz, TimestampNtz}
+
+import org.apache.spark.sql.catalyst.util.{ArrayData => SparkArrayData, SparkDateTimeUtils}
+import org.apache.spark.sql.types.{ArrayType => SparkArrayType, DataType => SparkDataType, StructType}
+
+/** Wraps a Spark [[SparkArrayData]] as a Fluss [[FlussInternalArray]]. */
+class SparkAsFlussArray(arrayData: SparkArrayData, elementType: SparkDataType)
```
wuchong, line 26: I saw `SparkAsFlussRow` extends `Serializable` interface, do we need to make `SparkAsFlussArray` also extend `Serializable`?

fluss-spark/fluss-spark-ut/src/test/scala/org/apache/fluss/spark/SparkWriteTest.scala
```diff
@@ -0,0 +1,271 @@
...
+    val table = loadFlussTable(tablePath)
+    val flussRows = getRowsWithChangeType(table).map(_._2)
+
+    val expectRows = Array(
+      GenericRowBuilder(4)
+        .setField(0, 600L)
+        .setField(1, 21L)
+        .setField(2, 601)
+        .setField(3, BinaryString.fromString("addr1"))
+        .builder(),
```
wuchong: We can replace the builder with `GenericRow.of(800L, 23L, 603, fromString("addr3"))`, which is more concise.

fluss-spark/fluss-spark-ut/src/test/scala/org/apache/fluss/spark/SparkWriteTest.scala
```diff
@@ -0,0 +1,271 @@
...
+          .setField(3, BinaryString.fromString("addr4"))
+          .builder()),
+      (
+        "+U",
+        GenericRowBuilder(4)
+          .setField(0, 800L)
+          .setField(1, 230L)
+          .setField(2, 603)
+          .setField(3, BinaryString.fromString("addr3"))
+          .builder()),
```
wuchong: nit: move this `+U` near after the corresponding `-U` message

fluss-spark/fluss-spark-ut/src/test/scala/org/apache/fluss/spark/SparkWriteTest.scala
```diff
@@ -0,0 +1,271 @@
...
+        "+U",
+        GenericRowBuilder(4)
+          .setField(0, 900L)
+          .setField(1, 240L)
+          .setField(2, 604)
+          .setField(3, BinaryString.fromString("addr4"))
+          .builder())
+    )
+    assertThat(flussRows2.length).isEqualTo(4)
+    assertThat(flussRows2).containsAll(expectRows2.toIterable.asJava)
```
wuchong, line 145: As we set the bucket number to `1`, so the changelog is in order globally, and we can assert `.containsExactlyElementsOf` here to also check the changelog order.

fluss-spark/fluss-spark-common/src/main/scala/org/apache/fluss/spark/write/FlussDataWriter.scala
```diff
@@ -0,0 +1,113 @@
...
+
+  override val writer: AppendWriter = table.newAppend().createWriter()
+
+  override def write(record: InternalRow): Unit = {
+    writer.append(flussRow.replace(record)).whenComplete {
+      (_, exception) =>
+        {
+          if (exception != null) {
+//            logError("Exception occurs while append row to fluss.", exception);
+            throw new RuntimeException("Failed to append record", exception)
```
wuchong: We must not throw exceptions directly in the completion callback, as they won’t propagate to the Spark writer and may be silently ignored.

Instead, we should capture any exception in a volatile field (e.g., `asyncWriterException`) within the Spark writer. Then, we can expose a `checkAsyncException()` method that throws the captured exception if it’s non-null.

This check should be invoked:

- At the beginning of `DataWriter#write`, to catch failures from prior async operations before processing new records, and
- And after `writer.flush()` in `DataWriter#commit`, to ensure any failure during flush or finalization is surfaced during commit.

This pattern ensures async errors are properly reported through Spark’s writer lifecycle. You can take `FlinkSinkWriter` as an example.

fluss-spark/fluss-spark-common/src/main/scala/org/apache/fluss/spark/write/FlussDataWriter.scala
```diff
@@ -0,0 +1,113 @@
...
+
+  override val writer: UpsertWriter = table.newUpsert().createWriter()
+
+  override def write(record: InternalRow): Unit = {
+    writer.upsert(flussRow.replace(record)).whenComplete {
+      (_, exception) =>
+        {
+          if (exception != null) {
+            logError("Exception occurs while upsert row to fluss.", exception);
+            throw new RuntimeException("Failed to upsert record", exception)
```
wuchong: ditto. and we can move the logging to the `checkAsyncException()` method.

fluss-spark/fluss-spark-ut/src/test/scala/org/apache/fluss/spark/SparkWriteTest.scala
YannByron: OK. But Spark will start two tasks to write multiple records, and there is no way to guarantee the order of these change data between records. So to achieve this goal, I only insert one piece of data at a time to validate the order of `-U` and `+U` using `containsExactlyElementsOf`.

wuchong: OK, then the original code makes sense to me.

fluss-common/src/main/java/org/apache/fluss/row/TimestampNtz.java
```diff
@@ -70,6 +71,18 @@ public int getNanoOfMillisecond() {
...
+    /**
+     * Creates an instance of {@link TimestampNtz} from microseconds.
+     *
+     * <p>The nanos-of-millisecond field will be set to zero.
+     *
+     * @param microseconds the number of microseconds since {@code 1970-01-01 00:00:00}; a negative
+     *     number is the number of microseconds before {@code 1970-01-01 00:00:00}
+     */
+    public static TimestampNtz fromMicros(long microseconds) {
+        return new TimestampNtz(Math.floorDiv(microseconds, MICROS_PER_MILLIS), 0);
```
wuchong: Why not convert the microsecond component to nanoseconds? From my perspective, using zero would actually lose precision. The method `org.apache.fluss.row.TimestampLtz#fromEpochMicros` intentionally preserves the full microsecond resolution, maybe we can use the same implementation here?

YannByron: You're right. I will modify this.

fluss-spark/fluss-spark-ut/src/test/scala/org/apache/fluss/spark/SparkWriteTest.scala
YannByron: OK. Let revert the latest change to the original one about this.

Yohahaha:
UT lacks of partition table write.

```
The table is a partitioned table, please use "subscribe(long partitionId, int bucket, long offset)" to subscribe a partitioned bucket instead.
java.lang.IllegalStateException: The table is a partitioned table, please use "subscribe(long partitionId, int bucket, long offset)" to subscribe a partitioned bucket instead.
	at org.apache.fluss.client.table.scanner.log.LogScannerImpl.subscribe(LogScannerImpl.java:176)
	at org.apache.fluss.client.table.scanner.log.LogScanner.subscribeFromBeginning(LogScanner.java:78)
	at org.apache.fluss.spark.FlussSparkTestBase.$anonfun$getRowsWithChangeType$1(FlussSparkTestBase.scala:88)
	at scala.collection.immutable.Range.foreach$mVc$sp(Range.scala:158)
	at org.apache.fluss.spark.FlussSparkTestBase.getRowsWithChangeType(FlussSparkTestBase.scala:88)
```

YannByron:
> UT lacks of partition table write.
> 
> ```
> The table is a partitioned table, please use "subscribe(long partitionId, int bucket, long offset)" to subscribe a partitioned bucket instead.
> java.lang.IllegalStateException: The table is a partitioned table, please use "subscribe(long partitionId, int bucket, long offset)" to subscribe a partitioned bucket instead.
> 	at org.apache.fluss.client.table.scanner.log.LogScannerImpl.subscribe(LogScannerImpl.java:176)
> 	at org.apache.fluss.client.table.scanner.log.LogScanner.subscribeFromBeginning(LogScanner.java:78)
> 	at org.apache.fluss.spark.FlussSparkTestBase.$anonfun$getRowsWithChangeType$1(FlussSparkTestBase.scala:88)
> 	at scala.collection.immutable.Range.foreach$mVc$sp(Range.scala:158)
> 	at org.apache.fluss.spark.FlussSparkTestBase.getRowsWithChangeType(FlussSparkTestBase.scala:88)
> ```

@Yohahaha thanks, I'll check and fix it.

Issue #2270, opened 2025-12-29
Refactor CALL procedure "sys.xxx_cluster_config" to support multiple conifg keys


wuchong:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/fluss/issues) and found nothing similar.

### Motivation

Currently, the `sys.set_cluster_config` procedure is restricted to updating a single configuration key per invocation. This limitation presents two main issues:

1. **Lack of Atomicity**: When multiple interdependent configurations need to be updated (e.g., enabling a feature and setting its threshold), they cannot be applied in a single transaction. If one call fails while the other succeeds, the cluster may be left in an inconsistent state.
2. **Inefficiency**: Updating multiple configurations requires multiple network round-trips and procedure calls.

To address this, we need to extend the capability to allow setting multiple configuration keys within a single, atomic operation.

```sql
-- Current limitation: Single key only
CALL sys.set_cluster_config('kv.rocksdb.shared-rate-limiter.bytes-per-sec', '200MB');

```

### Proposed Solution

We propose introducing a new procedure, `sys.set_cluster_configs` (plural), which accepts variable arguments (varargs). Each argument represents a configuration entry formatted as a key-value string.

**Syntax Example:**

```sql
CALL sys.set_cluster_configs(
  'kv.rocksdb.shared-rate-limiter.bytes-per-sec=200MB', 
  'kv.snapshot.interval=3min'
);

```

#### Implementation Details

The procedure will iterate through the provided arguments and parse each string:

1. **Parsing**: It will locate the first occurrence of the `=` character to separate the `key` from the `value`.
2. **Atomicity**: All parsed configurations will be applied in a single batch. If any configuration is invalid, the entire operation should fail to ensure consistency.

---

### 💡 Alternative Suggestion (Better Design?)

While the `key=value` string parsing works, it can be fragile (e.g., what if the value itself contains an `=`?). A more standard SQL approach is to pass a **Map** if the SQL dialect supports it, or use **variadic key-value pairs**.

**Option B: Variadic Key-Value Pairs (No parsing needed)**
Instead of passing "key=val" strings, you pass them as pairs.

```sql
CALL sys.set_cluster_configs(
  'kv.rocksdb.shared-rate-limiter.bytes-per-sec', '200MB',
  'kv.snapshot.interval', '3min'
);

```

* **Pros:** No need to parse strings; handles special characters in values easily.
* **Cons:** Needs validation to ensure an even number of arguments are passed.

### Other Projects
[...]

PR #2279, opened 2025-12-29
[flink] Refactor CALL procedure "sys.xxx_cluster_config" to support multiple conifg keys

sd4324530:
### Purpose

Linked issue: close https://github.com/apache/fluss/issues/2270

### Brief change log

Refactor CALL procedure "sys.xxx_cluster_config" to support multiple conifg keys

### DIFF
--- a/fluss-flink/fluss-flink-common/src/test/java/org/apache/fluss/flink/procedure/FlinkProcedureITCase.java
+++ b/fluss-flink/fluss-flink-common/src/test/java/org/apache/fluss/flink/procedure/FlinkProcedureITCase.java
@@ -435,7 +435,7 @@ public abstract class FlinkProcedureITCase {
     }
 
     @Test
-    void testSetClusterConfigValidation() throws Exception {
+    void testSetClusterConfigsValidation() throws Exception {
         // Try to set an invalid config (not allowed for dynamic change)
         assertThatThrownBy(
                         () ->
@@ -445,6 +445,7 @@ public abstract class FlinkProcedureITCase {
                                                         CATALOG_NAME))
                                         .await())
                 .rootCause()
+                // TODO: Fix misleading error: non-existent key reported as not allowed.
                 .hasMessageContaining(
                         "The config key invalid.config.key is not allowed to be changed dynamically");
 
@@ -492,7 +493,7 @@ public abstract class FlinkProcedureITCase {
     }
 
     @Test
-    void testResetClusterConfigValidation() throws Exception {
+    void testResetClusterConfigsValidation() throws Exception {
         // Try to reset an invalid config
         assertThatThrownBy(
                         () ->
@@ -502,6 +503,7 @@ public abstract class FlinkProcedureITCase {
                                                         CATALOG_NAME))
                                         .await())
                 .rootCause()
+                // TODO: Fix misleading error: non-existent key reported as not allowed.
                 .hasMessageContaining(
                         "The config key invalid.config.key is not allowed to be changed dynamically");
 


wuchong:
@platinumhamburg could you help to review this?

wuchong:
Besides, the coverage reports error. 

```
Warning:  Rule violated for class org.apache.fluss.flink.procedure.ResetClusterConfigsProcedure: lines covered ratio is 0.66, but expected minimum is 0.70
```

platinumhamburg:
> @platinumhamburg could you help to review this?

Sure, I'll review this PR.

fluss-flink/fluss-flink-common/src/test/java/org/apache/fluss/flink/procedure/FlinkProcedureITCase.java
```diff
@@ -265,12 +266,12 @@ void testDisableAuthorization() throws Exception {
     }
 
     @Test
-    void testGetClusterConfig() throws Exception {
+    void testGetClusterConfigs() throws Exception {
```
platinumhamburg, line 269: Need to enrich this case by covering scenarios, including: 
- retrieval of multiple configurations 
- some configurations do not exist.

fluss-flink/fluss-flink-common/src/test/java/org/apache/fluss/flink/procedure/FlinkProcedureITCase.java
```diff
@@ -305,19 +306,19 @@ void testGetClusterConfig() throws Exception {
...
-                                "Call %s.sys.set_cluster_config('%s')",
+                                "Call %s.sys.reset_cluster_configs('%s')",
                                 CATALOG_NAME,
                                 ConfigOptions.KV_SHARED_RATE_LIMITER_BYTES_PER_SEC.key()))
                 .await();
     }
 
     @Test
-    void testSetClusterConfig() throws Exception {
+    void testSetClusterConfigs() throws Exception {
```
platinumhamburg: Need to enrich the test scenarios covered in this test case, including:
- No parameters passed;
- Empty values passed as parameters;
- Mismatched key-value pairs in the input parameters.

sd4324530:
@platinumhamburg 
fix done, please take a look

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/procedure/SetClusterConfigsProcedure.java
```diff
@@ -45,84 +44,76 @@
...
+ * CALL sys.set_cluster_configs('kv.rocksdb.shared-rate-limiter.bytes-per-sec', '200MB');
+ * CALL sys.set_cluster_configs('datalake.format', 'paimon');
+ *
+ * -- Set multiple configurations at one time
+ * CALL sys.set_cluster_configs('kv.rocksdb.shared-rate-limiter.bytes-per-sec', '200MB','datalake.format', 'paimon');
  *
  * -- Delete a configuration (reset to default)
- * CALL sys.set_cluster_config('kv.rocksdb.shared-rate-limiter.bytes-per-sec', NULL);
- * CALL sys.set_cluster_config('kv.rocksdb.shared-rate-limiter.bytes-per-sec', '');
+ * CALL sys.set_cluster_configs('kv.rocksdb.shared-rate-limiter.bytes-per-sec', '');
```
platinumhamburg: Remove, not supported.

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/procedure/SetClusterConfigsProcedure.java
```diff
@@ -45,84 +44,76 @@
...
             // This will trigger validation on CoordinatorServer before persistence
-            admin.alterClusterConfigs(Collections.singletonList(alterConfig)).get();
+            admin.alterClusterConfigs(configList).get();
 
             return new String[] {
-                String.format(
-                        "Successfully %s configuration '%s'. "
-                                + "The change is persisted in ZooKeeper and applied to all servers.",
-                        operationDesc, configKey)
+                resultMessage + "The change is persisted in ZooKeeper and applied to all servers."
```
platinumhamburg: The current resultMessage aggregates the processing results of all requests into a single line, resulting in poor readability. Since the return value supports returning a String[], please refactor the code to improve readability.

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/procedure/ResetClusterConfigsProcedure.java
```diff
@@ -0,0 +1,108 @@
...
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * Procedure to reset cluster configuration dynamically.
+ *
+ * <p>This procedure allows modifying dynamic cluster configurations. The changes are:
+ *
+ * <ul>
+ *   <li>Validated by the CoordinatorServer before persistence
```
platinumhamburg, line 37: The comment description should be updated. In theory, an operation like "Reset to default value" should always succeed, as the default value should be a valid one.

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/procedure/ResetClusterConfigsProcedure.java
```diff
@@ -0,0 +1,108 @@
...
+ * <pre>
+ * -- reset a configuration
+ * CALL sys.reset_cluster_configs('kv.rocksdb.shared-rate-limiter.bytes-per-sec');
+ *
+ * -- reset multiple configurations at one time
+ * CALL sys.reset_cluster_configs('kv.rocksdb.shared-rate-limiter.bytes-per-sec', 'datalake.format');
+ *
+ * </pre>
+ *
+ * <p><b>Note:</b> Not all configurations support dynamic changes. The server will validate the
```
platinumhamburg: ditto

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/procedure/ResetClusterConfigsProcedure.java
```diff
@@ -0,0 +1,108 @@
...
+                        String.format(
+                                "Successfully %s configuration '%s'. ", operationDesc, configKey));
+            }
+
+            // Call Admin API to modify cluster configuration
+            // This will trigger validation on CoordinatorServer before persistence
+            admin.alterClusterConfigs(configList).get();
+
+            return new String[] {
+                resultMessage + "The change is persisted in ZooKeeper and applied to all servers."
```
platinumhamburg: Ditto, optimize the returned message—do not aggregate responses from all requests into a single line.

fluss-flink/fluss-flink-common/src/test/java/org/apache/fluss/flink/procedure/FlinkProcedureITCase.java
platinumhamburg, line 439: Need to enrich the test scenarios covered in this test case, including:
- No parameters passed;
- Mismatched key-value pairs in the input parameters.

sd4324530:
@platinumhamburg 
I reverted to the first commit and resubmitted, but the `SetClusterConfigProcedure` file is still being deleted. This is strange, and I haven't resolved this issue yet.

platinumhamburg:
> I reverted to the first commit and resubmitted, but the `SetClusterConfigProcedure` file is still being deleted. This is strange, and I haven't resolved this issue yet.

@sd4324530 No worries—this is actually expected behavior. When you revert to an earlier commit and resubmit, Git (and GitHub) determine whether a file is "renamed" or "deleted + added" based on content similarity. If the new version of SetClusterConfigProcedure differs significantly from the original (i.e., less than 50% of the lines are unchanged), Git treats it as a deletion of the old file and creation of a new one, rather than a rename. So even though you reverted and resubmitted, the system still shows it as deleted because the content similarity fell below the threshold. Overall, this aligns with how Git’s rename detection works.

website/docs/engine-flink/procedures.md
```diff
@@ -236,20 +236,51 @@ CALL [catalog_name.]sys.set_cluster_config(config_key => 'STRING')
...
+CALL [catalog_name.]sys.reset_cluster_configs(config_keys => 'key1' [, 'key2', ...])
 ```
 
+**Parameters:**
+
+- `config_keys`(required): The configuration keys to reset.
+
+**Important Notes:**
+
+- Changes are validated before being applied and persisted in ZooKeeper
```
platinumhamburg: The description here is still a bit odd, because the user isn't trying to set a new value—they're simply reverting to the default configuration, which should always be valid.

website/docs/engine-flink/procedures.md
```diff
@@ -236,20 +236,51 @@ CALL [catalog_name.]sys.set_cluster_config(config_key => 'STRING')
...
+- Changes are automatically applied to all servers (Coordinator and TabletServers)
+- Changes survive server restarts
```
platinumhamburg: This short description would be better placed in the set_cluster_configs section.

fluss-flink/fluss-flink-common/src/test/java/org/apache/fluss/flink/procedure/FlinkProcedureITCase.java
```diff
@@ -428,7 +461,61 @@ void testSetClusterConfigValidation() throws Exception {
...
+        assertThatThrownBy(
+                        () ->
+                                tEnv.executeSql(
+                                                String.format(
+                                                        "Call %s.sys.reset_cluster_configs('invalid.config.key')",
+                                                        CATALOG_NAME))
+                                        .await())
+                .rootCause()
+                .hasMessageContaining(
+                        "The config key invalid.config.key is not allowed to be changed dynamically");
```
platinumhamburg, line 508: An invalid (non-existent) key and a key that isn't allowed to be set dynamically are two distinct error cases. The current error message can be misleading. If we prefer not to complicate the validation logic, we could instead use "or" in the error message to cover both scenarios explicitly.

platinumhamburg (changes requested):

fluss-flink/fluss-flink-common/src/test/java/org/apache/fluss/flink/procedure/FlinkProcedureITCase.java
sd4324530: @platinumhamburg This error message is coming from here. I think the reason is the reuse of the `alterConfigs` API. Should we add a `resetConfigs` API to distinguish between these two different operations?
https://github.com/apache/fluss/blob/main/fluss-server/src/main/java/org/apache/fluss/server/DynamicConfigManager.java#L122-L153

fluss-server/src/main/java/org/apache/fluss/server/DynamicConfigManager.java
```diff
@@ -133,7 +133,7 @@ private void prepareIncrementalConfigs(
                     if (!dynamicServerConfig.isAllowedConfig(configPropName)) {
                         throw new ConfigException(
                                 String.format(
-                                        "The config key %s is not allowed to be changed dynamically.",
+                                        "The config key %s is not allowed to be changed dynamically or invalid (non-existent).",
                                         configPropName));
```
platinumhamburg, line 137: @sd4324530 Sorry, you misunderstood my point. What I meant was that, from a testing perspective, the exposed error message doesn't accurately reflect the underlying issue (and indeed, this wasn't introduced by your PR). I don't recommend simply modifying the error message thrown in DynamicConfigManager.java to meet the requirement, as this wouldn't actually address the root problem—that the validation error messages aren't properly categorized. If you don't intend to fix this issue now, I suggest converting the comment into a TODO note explaining the situation and reverting the changes made in DynamicConfigManager.java.

Additionally, I think the PR is overall quite mature and, after addressing these minor adjustments, it should be ready to move into the final review stage.

sd4324530: @platinumhamburg Okay, I think this issue might require a design adjustment to `DynamicConfigManager.alterConfigs`.

PR #2282, opened 2025-12-30
Support rocksdb metrics

### LINKED ISSUE #1320: [KV] Introduce full metrics export for RocksDB in KV Tablet

platinumhamburg:
### Purpose

Linked issue: close #1320

### DIFF
--- a/fluss-server/src/main/java/org/apache/fluss/server/kv/KvManager.java
+++ b/fluss-server/src/main/java/org/apache/fluss/server/kv/KvManager.java
@@ -266,14 +266,6 @@ public final class KvManager extends TabletManagerBase implements ServerReconfig
                                     sharedRocksDBRateLimiter);
                     currentKvs.put(tableBucket, tablet);
 
-                    // Register RocksDB metrics to TableMetricGroup for aggregation
-                    // Note: BucketMetricGroup is already created by Replica when LogTablet is
-                    // created
-                    if (tablet.getRocksDBMetrics() != null) {
-                        serverMetricGroup.registerRocksDBMetrics(
-                                tablePath.getTablePath(), tableBucket, tablet.getRocksDBMetrics());
-                    }
-
                     LOG.info(
                             "Created kv tablet for bucket {} in dir {}.",
                             tableBucket,
@@ -312,9 +304,6 @@ public final class KvManager extends TabletManagerBase implements ServerReconfig
         if (dropKvTablet != null) {
             TablePath tablePath = dropKvTablet.getTablePath();
             try {
-                // Unregister RocksDB metrics from TableMetricGroup
-                serverMetricGroup.unregisterRocksDBMetrics(tablePath, tableBucket);
-
                 dropKvTablet.drop();
                 if (dropKvTablet.getPartitionName() == null) {
                     LOG.info(
@@ -397,13 +386,6 @@ public final class KvManager extends TabletManagerBase implements ServerReconfig
                             currentKvs.get(tableBucket).getKvTabletDir().getAbsolutePath()));
         }
         this.currentKvs.put(tableBucket, kvTablet);
-
-        // Register RocksDB metrics to TableMetricGroup for aggregation
-        // Note: BucketMetricGroup is already created by Replica when LogTablet is created
-        if (kvTablet.getRocksDBMetrics() != null) {
-            serverMetricGroup.registerRocksDBMetrics(
-                    physicalTablePath.getTablePath(), tableBucket, kvTablet.getRocksDBMetrics());
-        }
 
         return kvTablet;
     }
--- a/fluss-server/src/main/java/org/apache/fluss/server/metrics/group/TableMetricGroup.java
+++ b/fluss-server/src/main/java/org/apache/fluss/server/metrics/group/TableMetricGroup.java
@@ -27,12 +27,13 @@ import org.apache.fluss.metrics.NoOpCounter;
 import org.apache.fluss.metrics.ThreadSafeSimpleCounter;
 import org.apache.fluss.metrics.groups.AbstractMetricGroup;
 import org.apache.fluss.metrics.registry.MetricRegistry;
-import org.apache.fluss.server.kv.rocksdb.RocksDBMetrics;
+import org.apache.fluss.server.kv.rocksdb.RocksDBStatistics;
+import org.apache.fluss.utils.MapUtils;
 
 import javax.annotation.Nullable;
 
-import java.util.HashMap;
 import java.util.Map;
+import java.util.stream.Stream;
 
 import static org.apache.fluss.metrics.utils.MetricGroupUtils.makeScope;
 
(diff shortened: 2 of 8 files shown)


fluss-server/src/main/java/org/apache/fluss/server/metrics/group/TableMetricGroup.java
```diff
@@ -236,17 +243,159 @@ public BucketMetricGroup addBucketMetricGroup(
...
+    public void registerRocksDBMetrics(TableBucket tableBucket, RocksDBMetrics rocksDBMetrics) {
+        rocksDBMetricsMap.put(tableBucket, rocksDBMetrics);
+    }
+
+    /**
+     * Unregister RocksDB metrics for a bucket.
+     *
+     * @param tableBucket the table bucket
+     */
+    public void unregisterRocksDBMetrics(TableBucket tableBucket) {
```
swuferhong: This method is never used?

fluss-server/src/main/java/org/apache/fluss/server/kv/KvManager.java
```diff
@@ -304,6 +312,9 @@ public void dropKv(TableBucket tableBucket) {
         if (dropKvTablet != null) {
             TablePath tablePath = dropKvTablet.getTablePath();
             try {
+                // Unregister RocksDB metrics from TableMetricGroup
+                serverMetricGroup.removeTableBucketMetricGroup(tablePath, tableBucket);
```
swuferhong: why not call `serverMetricsGroup.unregisterRocksDBMetrics` as the comment said?

swuferhong (commented):
Thanks @platinumhamburg. LGTM overall, I left one minor comment:

platinumhamburg:
> Thanks @platinumhamburg. LGTM overall, I left one minor comment:

@swuferhong Thank you for the detailed review comments; I've addressed all of them.

swuferhong (approved):
LGTM. Just one minor point to mention: we've already called `removeTableBucketMetricsGroup` in the `stopReplica` method—do we still need to call it in `dropKv`?

platinumhamburg:
> LGTM. Just one minor point to mention: we've already called `removeTableBucketMetricsGroup` in the `stopReplica` method—do we still need to call it in `dropKv`?

@swuferhong You are right, I removed it.

fluss-server/src/main/java/org/apache/fluss/server/metrics/group/TableMetricGroup.java
```diff
@@ -43,6 +44,10 @@ public class TableMetricGroup extends AbstractMetricGroup {
 
     private final Map<TableBucket, BucketMetricGroup> buckets = new HashMap<>();
 
+    // Directly manage RocksDB metrics for aggregation
+    // This is cleaner than passing through BucketMetricGroup
+    private final Map<TableBucket, RocksDBMetrics> rocksDBMetricsMap = new HashMap<>();
```
wuchong: Both this `rocksDBMetricsMap` and `buckets` maps should be `ConcurrentHashMap`, because they may be invoked/iterated in other threads (metric reporter).

fluss-server/src/main/java/org/apache/fluss/server/metrics/group/TableMetricGroup.java
```diff
@@ -236,17 +243,159 @@ public BucketMetricGroup addBucketMetricGroup(
...
+    }
+
+    /**
+     * Register RocksDB metrics for a bucket. This allows table-level aggregation without
+     * registering bucket-level metrics.
+     *
+     * @param tableBucket the table bucket
+     * @param rocksDBMetrics the RocksDB metrics accessor
+     */
+    public void registerRocksDBMetrics(TableBucket tableBucket, RocksDBMetrics rocksDBMetrics) {
```
wuchong: This method is confusing in relation to `registerRocksDBMetrics()`.

Currently:

- `registerRocksDBMetrics()` registers actual Fluss `Metric` instances (i.e., exported to the metrics system),
- while this `registerRocksDBMetrics(TableBucket tableBucket, RocksDBMetrics rocksDBMetrics)` merely adds an internal `RocksDBMetrics` object to this class.

In Fluss, other `XxxMetrics` classes (e.g., `RequestsMetrics`, `UserMetrics`) directly manage and expose true Fluss `Metric`s. In contrast, `RocksDBMetrics` only collects low-level RocksDB statistics and does not register them as metrics by itself.

To clarify the distinction and align naming with responsibility, I propose:

- Rename `RocksDBMetrics` → `RocksDBStatistics`
- Rename this method → `registerRocksDBStatistics()`

This change would improve consistency, reduce confusion, and better reflect the actual role of the class.

fluss-server/src/main/java/org/apache/fluss/server/metrics/group/TableMetricGroup.java
```diff
@@ -236,17 +243,159 @@ public BucketMetricGroup addBucketMetricGroup(
...
+                // Ignore close errors
+            }
+        }
     }
 
     public int bucketGroupsCount() {
         return buckets.size();
     }
 
+    public java.util.Collection<BucketMetricGroup> getBucketMetricGroups() {
```
wuchong, line 253: remove, unused

fluss-server/src/main/java/org/apache/fluss/server/kv/KvManager.java
```diff
@@ -304,6 +312,9 @@ public void dropKv(TableBucket tableBucket) {
         if (dropKvTablet != null) {
             TablePath tablePath = dropKvTablet.getTablePath();
             try {
+                // Unregister RocksDB metrics from TableMetricGroup
+                serverMetricGroup.unregisterRocksDBMetrics(tablePath, tableBucket);
```
wuchong: We already unregister RocksDB metrics in `ReplicaManager#stopReplica`, which invokes `TableMetricGroup#removeBucketMetricGroup(...)` by `TabletServerMetricGroup#removeTableBucketMetricGroup`. That method, in turn, handles the unregistration of the RocksDB metrics.

fluss-server/src/main/java/org/apache/fluss/server/metrics/group/TabletServerMetricGroup.java
```diff
@@ -30,13 +30,19 @@
...
 
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
 import java.util.Map;
 
 /** The metric group for tablet server. */
 public class TabletServerMetricGroup extends AbstractMetricGroup {
 
+    private static final Logger LOG = LoggerFactory.getLogger(TabletServerMetricGroup.class);
```
wuchong: remove, not used

platinumhamburg:
> BucketMetricGroup

@wuchong Thank you for your review comments. I’ll provide a detailed explanation of why I designed it this way. Initially, I decoupled RocksDB metrics from the bucket lifecycle for two main reasons:

We do not intend to collect, register, or report RocksDB monitoring metrics at the bucket granularity. Since BucketMetricGroup is specifically designed to manage bucket-level metrics, from a perspective of functional consistency, it was not appropriate to associate RocksDB metrics with BucketMetricGroup.

In practice, for primary key tables, the lifecycle of a KvTablet does not align with that of a bucket. The BucketMetricGroup is only unregistered when StopReplica is called, whereas the KvTablet must unregister its metric group as soon as it transitions to a follower role. Therefore, the lifecycle of KvTablet-related metrics should be tied to the KvTablet itself, not to the bucket.

platinumhamburg:
@wuchong I'v resolved all review comments, thank for review again.

PR #2305, opened 2026-01-05
[spark] Add spark show columns command ut

Yohahaha:
### Purpose

Linked issue: close #xxx

Add spark show columns command ut, which should throw exception due to V2 datasource was not supportted.


YannByron:
The current integration of Fluss with Spark is just getting started, there will definitely be many features and scenarios that are not supported. I think we can support them directly instead of perceiving it through UT.

PR #2307, opened 2026-01-05
Flink Intergration of Aggregation Merge Engine

### LINKED ISSUE #2244: [flink] Integrate the Aggregate Merge Engine with Apache Flink.
### Search before asking

### Description

Integrate the Aggregate Merge Engine with Apache Flink.

### Willingness to contribute

platinumhamburg:
### Purpose

Linked issue: close #2244

### DIFF
--- a/fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/utils/FlinkAggFunctionParser.java
+++ b/fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/utils/FlinkAggFunctionParser.java
@@ -67,7 +67,8 @@ public class FlinkAggFunctionParser {
             return Optional.empty();
         }
 
-        String funcName = options.getString(columnFuncKey, null);
+        // convert to lower case for consistent parameter option key
+        String funcName = options.getString(columnFuncKey, null).toLowerCase();
         AggFunctionType type = parseAggFunctionType(funcName, columnName);
 
         // Collect column-level parameters: fields.<column>.<function_name>.*
@@ -93,6 +94,7 @@ public class FlinkAggFunctionParser {
             String columnName, AggFunction aggFunction, Map<String, String> options) {
         // Set function name
         String funcKey = AGG_PREFIX + columnName + AGG_SUFFIX;
+        // funcName has already in lower case
         String funcName = aggFunction.getType().toString();
         options.put(funcKey, funcName);
 
--- a/fluss-flink/fluss-flink-common/src/test/java/org/apache/fluss/flink/utils/FlinkAggFunctionParserTest.java
+++ b/fluss-flink/fluss-flink-common/src/test/java/org/apache/fluss/flink/utils/FlinkAggFunctionParserTest.java
@@ -63,6 +63,24 @@ class FlinkAggFunctionParserTest {
         assertThat(result).isPresent();
         assertThat(result.get().getType()).isEqualTo(AggFunctionType.LISTAGG);
         assertThat(result.get().getParameter("delimiter")).contains(";");
+
+        Map<String, String> newOptions = new HashMap<>();
+        FlinkAggFunctionParser.formatAggFunctionToOptions(
+                "tags", AggFunctions.LISTAGG(";"), newOptions);
+        assertThat(newOptions).isEqualTo(options.toMap());
+    }
+
+    @Test
+    void testParseFunctionWithUpperCaseName() {
+        Configuration options = new Configuration();
+        options.setString("fields.tags.agg", "LISTAGG");
+        options.setString("fields.tags.listagg.delimiter", ";");
+
+        Optional<AggFunction> result = FlinkAggFunctionParser.parseAggFunction("tags", options);
+
+        assertThat(result).isPresent();
+        assertThat(result.get().getType()).isEqualTo(AggFunctionType.LISTAGG);
+        assertThat(result.get().getParameter("delimiter")).contains(";");
     }
 
     @Test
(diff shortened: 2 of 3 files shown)


fluss-flink/fluss-flink-common/src/test/java/org/apache/fluss/flink/utils/FlinkAggFunctionParserTest.java
```diff
@@ -0,0 +1,135 @@
...
+
+        assertThatThrownBy(() -> FlinkAggFunctionParser.parseAggFunction("total", options))
+                .isInstanceOf(IllegalArgumentException.class)
+                .hasMessageContaining("Empty aggregation function name");
+    }
+
+    @Test
+    void testRoundTripConversion() {
+        // Test that parse and format are inverse operations
+        java.util.Map<String, String> options = new java.util.HashMap<>();
```
wuchong: replace qualified name with imports

fluss-flink/fluss-flink-common/src/test/java/org/apache/fluss/flink/utils/FlinkConversionsTest.java
```diff
@@ -415,6 +415,59 @@ void testFlinkMaterializedTableConversions() {
...
+                        currentMillis,
+                        currentMillis);
+        CatalogTable convertedFlinkTable = (CatalogTable) FlinkConversions.toFlinkTable(tableInfo);
+
+        // Verify aggregation functions are preserved
+        assertThat(convertedFlinkTable.getOptions())
+                .containsEntry("table.merge-engine", "aggregation")
+                .containsEntry("fields.sum_val.agg", "sum")
+                .containsEntry("fields.tags.agg", "listagg")
+                .containsEntry("fields.tags.agg.params.delimiter", "|");
```
wuchong: Replace to `assertThat(convertedFlinkTable.getOptions()).containsAllEntriesOf(options);` to avoid duplicate declare the map keys and values.

fluss-flink/fluss-flink-common/src/test/java/org/apache/fluss/flink/sink/FlinkTableSinkITCase.java
```diff
@@ -1221,6 +1221,136 @@ void testVersionMergeEngineWithTypeTimestampLTZ9() throws Exception {
...
+                        + "'fields.min_double.agg' = 'min', "
+                        + "'fields.min_string.agg' = 'min', "
+                        + "'fields.first_val.agg' = 'first_value', "
+                        + "'fields.first_val_non_null.agg' = 'first_value_ignore_nulls', "
+                        + "'fields.last_val.agg' = 'last_value', "
+                        + "'fields.last_val_non_null.agg' = 'last_value_ignore_nulls', "
+                        + "'fields.bool_and_val.agg' = 'bool_and', "
+                        + "'fields.bool_or_val.agg' = 'bool_or', "
+                        + "'fields.listagg_val.agg' = 'listagg', "
+                        + "'fields.listagg_val.agg.params.delimiter' = '|')");
```
wuchong: The previous approach `fields.<field-name>.agg.params.<param-name>` has several drawbacks:

Since `fields.<field-name>.agg` is already designated to store the aggregation function name, it cannot simultaneously serve as a parent object for nested parameters. In most configuration systems, such as YAML, JSON, a given key cannot be both a **leaf** (holding a scalar value) and a **branch** (containing child keys) at the same time.

Here are a few alternative proposals which is also shorter:

1. Use `'fields.<field-name>.params.<param>'`, for example:  
   `'fields.tags.params.delimiter' = ','`

2. Or use `'fields.<field-name>.<function-name>.<param>' = 'value'`, for example:  
   `'fields.tags.listagg.delimiter' = ','`

platinumhamburg: I think proposal 2 is a better choice. proposal 1 has an ambiguity: when the parent node of params is a field name, it doesn't explicitly indicate that these are parameters for an aggregation function, which could lead to namespace ambiguity or conflicts. I will refactor based on proposal 2.

platinumhamburg:
Hi @wuchong, thanks for your review comments, I've addressed all of them. Please take another look when you have time.

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/utils/FlinkAggFunctionParser.java
```diff
@@ -88,19 +86,19 @@ public static Optional<AggFunction> parseAggFunction(String columnName, Configur
...
-        options.put(funcKey, aggFunction.getType().toString());
+        String funcName = aggFunction.getType().toString();
+        options.put(funcKey, funcName);
 
-        // Set parameters
+        // Set parameters: fields.<column>.<function_name>.<param>
         for (Map.Entry<String, String> param : aggFunction.getParameters().entrySet()) {
-            String paramKey =
-                    AGG_PREFIX + columnName + AGG_SUFFIX + PARAMS_SUFFIX + "." + param.getKey();
+            String paramKey = AGG_PREFIX + columnName + "." + funcName + "." + param.getKey();
```
wuchong, line 103: The function name value might be in uppercase, so we should convert it to lowercase to ensure the option key is always in lowercase.

wuchong (approved):
Thanks @platinumhamburg . I only appended a commit to fix the only comment with a test for this case and rebased the branch.

wuchong:
I will merge it once the CI is passed.

PR #2320, opened 2026-01-07
[hotfix] use ltz_millis  as paimon system column for timestamp datatype

luoyuxia:
### Purpose

Linked issue: close #xxx
Current, `TIMESTAMP_WITH_LOCAL_TIME_ZONE()` is actaul with presion 6, but fluss's record timestamp is only up to presion 3, so, need to use presion 6

### Brief change log
Change to use `TIMESTAMP_LTZ_MILLIS`


LiebingYu:
@luoyuxia I think it's not a hotfix. For example:
1. user create a lake table A with `__timestamp` of presion 6
2. user disable lake table of A
3. cluster upgrade with `__timestamp` of presion 3
4. user will fail to enable lake table A, since schema not match.

LiebingYu:
+1

PR #2321, opened 2026-01-07
[hotfix] Fix union read can't restore issue

luoyuxia:
### Purpose

Linked issue: close #xxx

### DIFF
--- a/fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/source/state/FlussSourceEnumeratorStateSerializer.java
+++ b/fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/source/state/FlussSourceEnumeratorStateSerializer.java
@@ -23,6 +23,7 @@ import org.apache.fluss.flink.source.split.SourceSplitSerializer;
 import org.apache.fluss.lake.source.LakeSource;
 import org.apache.fluss.lake.source.LakeSplit;
 import org.apache.fluss.metadata.TableBucket;
+import org.apache.fluss.utils.types.Tuple2;
 
 import org.apache.flink.core.io.SimpleVersionedSerializer;
 import org.apache.flink.core.memory.DataInputDeserializer;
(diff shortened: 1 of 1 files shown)


fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/source/state/FlussSourceEnumeratorStateSerializer.java
```diff
@@ -37,17 +38,21 @@
 import java.util.Map;
 import java.util.Set;
 
+import static org.apache.fluss.utils.Preconditions.checkNotNull;
+
```
loserwang1024, line 43: Would like to add what's the different with different version? such as:
image

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/source/state/FlussSourceEnumeratorStateSerializer.java
```diff
@@ -78,24 +100,29 @@ public byte[] serialize(SourceEnumeratorState state) throws IOException {
...
-
         final byte[] result = out.getCopyOfBuffer();
         out.clear();
         return result;
     }
 
     @Override
     public SourceEnumeratorState deserialize(int version, byte[] serialized) throws IOException {
-        if (version != VERSION_0) {
+        if (version != VERSION_0 && version != CURRENT_VERSION) {
```
loserwang1024: > version != VERSION_0 && version != CURRENT_VERSION 

When we support version 3 later, state with version 3 cannot downgrade. We have met this problem in many connectors.  @leonardBang , CC, should we stop higher version's state?

And I also found a problem : we can never downgrade later to version 1. Because in later code will version 1 state: version != VERSION_0 is always true. Maybe we should add this to upgrade document later.

loserwang1024 (commented):
I have left some message.

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/source/state/FlussSourceEnumeratorStateSerializer.java
luoyuxia: 1: so, what's the best pratice in here? Always make the version check pass and make sure version3 compitable with v2? From kafka connector code, seems the kafka connector will still have the problem?

2: what do you mean by saying `we can never downgrade later to version 1`?  Seems it still the same problem to 1?

loserwang1024: I understand it. But in your follower comment "for VERSION_1 and later" maybe inconsistency.

image

Maybe you can use `switch.. case` as kafka does. It's clearer
image

luoyuxia: Agree, that's clearer

luoyuxia:
@loserwang1024 Comments is addressed. PTAL

PR #2322, opened 2026-01-07
[common] Add field_id for Nested Row.

### LINKED ISSUE #2310: Add Nested Row 's field_id
### Search before asking

### Purpose

Linked issue: close #2310 , add field id to nested rows.  

### Brief change log

All fields are assigned unique, sequential IDs in a flattened order, regardless of nesting level. For example:

```text
struct< 
  a: tinyint, 
  b: struct< 
    c: tinyint, 
    d: struct< 
      e: tinyint, 
      f: tinyint
    >, 
    g: string 
  > 
>
```

Then the field Id for each field is:

Field Name | Field ID
-- | --
a | 0
b | 1
b.c | 2
b.d | 3
b.d.e | 4
b.d.f | 5
b.g | 6

### Why Flatten Numerical Order?
* Simplifies ID Management: No need to compute hierarchical offsets (e.g., parent_id * depth + child_index).
* Compatibility: Works seamlessly with flat data structures like Arrow's RecordBatch. In our FileLogProjection, same thing  is also done. Thus later projection push down will be more easier.

image

### API and Format

Add field_id to org.apache.fluss.types.DataField.

### Willingness to contribute

loserwang1024:
### Purpose

Linked issue: close #2310 , add field id to nested rows.  

### Brief change log

All fields are assigned unique, sequential IDs in a flattened order, regardless of nesting level. For example:

```text
struct< 
  a: tinyint, 
  b: struct< 
    c: tinyint, 
    d: struct< 
      e: tinyint, 
      f: tinyint
    >, 
    g: string 
  > 
>
```

Then the field Id for each field is:

Field Name | Field ID
-- | --
a | 0
b | 1
b.c | 2
b.d | 3
b.d.e | 4
b.d.f | 5
b.g | 6

### Why Flatten Numerical Order?
* Simplifies ID Management: No need to compute hierarchical offsets (e.g., parent_id * depth + child_index).
* Compatibility: Works seamlessly with flat data structures like Arrow's RecordBatch. In our FileLogProjection, same thing  is also done. Thus later projection push down will be more easier.

image

### API and Format

Add field_id to org.apache.fluss.types.DataField.

### DIFF
--- a/fluss-common/src/main/java/org/apache/fluss/types/DataTypes.java
+++ b/fluss-common/src/main/java/org/apache/fluss/types/DataTypes.java
@@ -317,17 +317,6 @@ public class DataTypes {
         return new DataField(name, type, fieldId);
     }
 
-    /**
-     * Creates a field definition with field name, data type, and a description.
-     *
-     * @param name the field name
-     * @param type the field data type
-     * @param description the field description
-     * @return a new data field with description but with default field ID (0)
-     * @deprecated Use {@link #FIELD(String, DataType, String, int)} instead to explicitly specify
-     *     field ID. Field ID is required for schema evolution support.
-     */
-    @Deprecated
     public static DataField FIELD(String name, DataType type, String description) {
         return new DataField(name, type, description);
     }
--- a/fluss-common/src/test/java/org/apache/fluss/utils/json/DataTypeJsonSerdeTest.java
+++ b/fluss-common/src/test/java/org/apache/fluss/utils/json/DataTypeJsonSerdeTest.java
@@ -90,7 +90,12 @@ public class DataTypeJsonSerdeTest extends JsonSerdeTestBase<DataType> {
                         new LocalZonedTimestampType(3),
                         new ArrayType(new IntType(false)),
                         new MapType(new BigIntType(false), new IntType(false)),
-                        RowType.of(new BigIntType(), new IntType(false), new StringType()));
+                        new RowType(
+                                true,
+                                Arrays.asList(
+                                        DataTypes.FIELD("f0", new BigIntType(), null),
+                                        DataTypes.FIELD("f1", new IntType(false), null, 1),
+                                        DataTypes.FIELD("f2", new StringType(), null, 2))));
 
         final List<DataType> allTypes = new ArrayList<>();
         // consider nullable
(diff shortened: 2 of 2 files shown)


fluss-flink/fluss-flink-common/src/test/java/org/apache/fluss/flink/sink/FlinkComplexTypeITCase.java
```diff
@@ -592,4 +597,72 @@ void testExceptionsForComplexTypesUsage() {
...
+        // Currently, flink not supported push down nested row projection because
+        // FlinkTableSource.supportsNestedProjection returns false.
+        // Todo: support nested row projection pushdown in
+        // https://github.com/apache/fluss/issues/2311 later.
+        String s = tEnv.explainSql("select id, simple_row.a, nested_row.y.z from row_log_test");
+        assertThat(s)
+                .contains(
+                        "TableSourceScan(table=[[testcatalog, defaultdb, row_log_test, project=[id, simple_row, nested_row]]], fields=[id, simple_row, nested_row])");
+        rowIter =
+                tEnv.executeSql("select id, simple_row.a, nested_row.y.z from row_log_test")
```
wuchong: Can you add a primitive type column `f` at the end of the schema, and `select id, simple_row.a, nested_row.y.z, f from xxx`, this verifies the projection pushdown works if there is complex type in the middle of schema.

fluss-common/src/main/java/org/apache/fluss/utils/json/DataTypeJsonSerde.java
```diff
@@ -291,18 +294,38 @@ private static DataType deserializeMap(JsonNode dataTypeNode) {
...
+                            : null;
+
+            final int fieldId;
+            if (fieldNode.has(FIELD_NAME_FIELD_ID)) {
+                hasFieldWithId = true;
+                fieldId = fieldNode.get(FIELD_NAME_FIELD_ID).asInt();
             } else {
-                fieldDescription = null;
+                hasFieldWithoutId = true;
+                fieldId = autoFieldId++;
```
wuchong: It seems the `autoFieldId` is not used.

fluss-common/src/main/java/org/apache/fluss/utils/json/DataTypeJsonSerde.java
```diff
@@ -291,18 +294,38 @@ private static DataType deserializeMap(JsonNode dataTypeNode) {
...
-            fields.add(new DataField(fieldName, fieldType, fieldDescription));
+
+            fields.add(new DataField(fieldName, fieldType, fieldDescription, fieldId));
         }
+
+        if (hasFieldWithId && hasFieldWithoutId) {
+            throw new TableException(
+                    "Field ID inconsistency detected in row type: "
+                            + "all fields must either have field IDs or none should have field IDs.");
+        }
```
wuchong: I think we can remove this to simplify the logic. We can treat the field id the same as description (a metadata), becuase `RowType` itself doesn't use it. So RowType is hard to know the correctness of the field id. It's the `Schema` use the field id, so it should be the `Schema` to validate it.

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/source/reader/FlinkSourceSplitReader.java
```diff
@@ -575,10 +578,7 @@ private void sanityCheck(RowType flussTableRowType, @Nullable int[] projectedFie
                         ? flussTableRowType.project(projectedFields)
                         // only read the output fields from source
                         : flussTableRowType.project(sourceOutputType.getFieldNames());
-        if (!sourceOutputType.copy(false).equals(tableRowType.copy(false))) {
-            // The default nullability of Flink row type and Fluss row type might be not the same,
-            // thus we need to compare the row type without nullability here.
-
+        if (!compareRowTypesIgnoreFieldId(tableRowType, sourceOutputType)) {
```
wuchong: We don't need this complex logic if we ignore field id in `RowType#equals`.

fluss-common/src/main/java/org/apache/fluss/types/RowType.java
```diff
@@ -172,23 +171,98 @@ public int hashCode() {
         return Objects.hash(super.hashCode(), fields);
     }
 
+    /**
+     * Compares this RowType with another RowType, ignoring field IDs.
+     *
+     * @param other the other RowType to compare with
+     * @return true if the RowTypes are equal ignoring field IDs, false otherwise
+     */
+    public boolean equalsIgnoreFieldId(RowType other) {
```
wuchong: We can provide a `equalsWithFieldId` method for the purpose that compares with field id if we ignore field id in `equals()`. Besides, we should use a DataTypeVisitor to do the equals field id, because there are nested fields.

fluss-common/src/main/java/org/apache/fluss/types/RowType.java
```diff
@@ -172,23 +171,98 @@ public int hashCode() {
...
-        if (!duplicates.isEmpty()) {
+        if (!duplicateNames.isEmpty()) {
+            throw new IllegalArgumentException(
+                    String.format(
+                            "Field names must be unique. Found duplicates: %s", duplicateNames));
+        }
+
+        // Validate and process field IDs
+        final List<Integer> fieldIds =
+                fields.stream().map(DataField::getFieldId).collect(Collectors.toList());
```
wuchong: We can remove the validation, because Schema should check it, not the RowType.

fluss-common/src/main/java/org/apache/fluss/types/RowType.java
```diff
@@ -229,18 +307,20 @@ public static class Builder {
         private final List<DataField> fields = new ArrayList<>();
 
         private final boolean isNullable;
+        private final AtomicInteger fieldId;
```
wuchong: We don't need this because `Schema#column()` will reassign field ids.

fluss-common/src/main/java/org/apache/fluss/metadata/Schema.java
```diff
@@ -728,4 +780,73 @@ public static RowType getKeyRowType(Schema schema, int[] keyIndexes) {
...
+    private static void checkFieldIds(List<Column> columns, int highestFieldId) {
+
+        // Collect all field IDs (including nested fields) for validation
+        List<Integer> allFieldIds = collectAllFieldIds(columns);
+
+        // Validate all field IDs (including nested fields) are unique
+        long uniqueFieldIdsCount = allFieldIds.stream().distinct().count();
+        checkState(
+                uniqueFieldIdsCount == allFieldIds.size(),
+                "All field IDs (including nested fields) must be unique. Found %s unique IDs but expected %s.",
```
wuchong, line 810: nit: Would be better to just print the `columns` which should carries the field id in the result string.

loserwang1024:
@wuchong , I have modified based on comment. Please help check again.Especially, currently, `Schema.Builder#column `、`Schema.Builder#fromFields` and `Schema.Builder#fromRowType` will reassign id while `Schema.Builder#fromColumns ` won't. The later one is used to build from RPC and json value.

wuchong (approved):

PR #2323, opened 2026-01-07
[common]Add Early Validation for Aggregation Function Data Types During Table Creation

### LINKED ISSUE #2302: Add Early Validation for Aggregation Function Data Types During Table Creation
### Search before asking

### Description

Currently, Fluss does **not validate** whether the column data type is compatible with a given aggregation function (e.g., `MAX`, `MIN`) at table creation time. This leads to **runtime failures** or **unexpected behavior** when unsupported types—such as `ARRAY`, `MAP`, or `ROW`—are used with aggregation functions that require orderable types.

To improve developer experience and system robustness, we should **fail fast during DDL validation**.

### ✅ Expected Behavior
- During `CREATE TABLE` (or table descriptor validation), if an aggregation function is specified on a column with an **unsupported data type**, Fluss should **throw a clear, actionable error immediately**.
- The validation should reside in:  
  `org.apache.fluss.server.utils.TableDescriptorValidation#validateAggregationFunctionParameters`

### Solution
To ensure future aggregate functions are easily extensible, we should introduce a standardized way to declare and validate the data types each function supports.

One approach is adding a method such as `validateDataType()` to the `AggFunctionType` interface, similar in spirit to the existing `validateParameter()` that explicitly checks whether a given column type is compatible with the aggregate function. 

### 🧪 Test Coverage
- We can add such tests in `FlussAdminITCase`, like `FlussAdminITCase#testCreateTableWithInvalidProperty`.

### Willingness to contribute

Jackeyzhe:
… Creation

### Purpose

Linked issue: close #2302 

### Brief change log
Add `validateDataType` in `AggFunctionType`, ensure that set correct data type for agg field.

### Tests
Add tests in `FlussAdminITCase`

### DIFF
--- a/fluss-common/src/main/java/org/apache/fluss/metadata/AggFunctionType.java
+++ b/fluss-common/src/main/java/org/apache/fluss/metadata/AggFunctionType.java
@@ -26,6 +26,7 @@ import java.util.Collections;
 import java.util.Locale;
 import java.util.Set;
 
+import static org.apache.fluss.metadata.AggFunctions.PARAM_DELIMITER;
 import static org.apache.fluss.utils.Preconditions.checkArgument;
 
 /**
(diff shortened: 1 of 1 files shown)


Jackeyzhe:
Hi @wuchong could you please help review this PR if you have time?

fluss-common/src/main/java/org/apache/fluss/metadata/AggFunctionType.java
```diff
@@ -111,6 +117,60 @@ public void validateParameter(String parameterName, String parameterValue) {
...
+                };
+            case LAST_VALUE:
+            case LAST_VALUE_IGNORE_NULLS:
+            case FIRST_VALUE:
+            case FIRST_VALUE_IGNORE_NULLS:
+                return DataTypeFamily.values();
+            case LISTAGG:
+            case STRING_AGG:
+                return new DataTypeFamily[] {DataTypeFamily.CHARACTER_STRING};
+            default:
```
wuchong: Listing all function types explicitly is error-prone when adding new aggregate functions, as it’s easy to forget to update the data type validation logic.  

To avoid missing data type checks for new functions, I suggest making **supported parameter types** and **supported data type families** explicit member fields of the `AggFunctionType` class. This way, we can directly retrieve the allowed type families from each function type and perform consistent, centralized validation—reducing maintenance overhead and improving correctness.

fluss-common/src/test/java/org/apache/fluss/metadata/TableDescriptorTest.java
```diff
@@ -344,7 +345,7 @@ void testInvalidSumFunctionWithParameters() {
         Map<String, String> params = new HashMap<>();
         params.put("some_param", "value");
 
-        assertThatThrownBy(() -> AggFunctions.of(AggFunctionType.SUM, params).validate())
+        assertThatThrownBy(() -> AggFunctions.of(AggFunctionType.SUM, params).validateParameters())
```
wuchong, line 349: Add unit tests for the data type validation for agg function columns.

fluss-common/src/main/java/org/apache/fluss/metadata/AggFunction.java
```diff
@@ -104,18 +106,31 @@ public boolean hasParameters() {
...
      *   <li>All parameter values are valid
+     *   <li>The field data type is valid
      * </ul>
      *
-     * @throws IllegalArgumentException if any parameter is invalid
+     * @param fieldType the field data type
+     * @throws IllegalArgumentException if any parameter is invalid or data type is invalid
      */
-    public void validate() {
+    public void validate(DataType fieldType) {
```
wuchong: This method looks like just validating the data types. I think we don't need the combined method. Just separate them into 2 public methods `validateParameters()` and `validateDataType(fieldType)` is more clear.

fluss-client/src/test/java/org/apache/fluss/client/admin/FlussAdminITCase.java
```diff
@@ -1556,4 +1556,102 @@ public void testAddAndRemoveServerTags() throws Exception {
...
+                        .column("and_value", DataTypes.STRING(), AggFunctions.BOOL_AND())
+                        .primaryKey("id")
+                        .build();
+        TableDescriptor t1 =
+                TableDescriptor.builder()
+                        .schema(schema1)
+                        .comment("aggregate merge engine table")
+                        .properties(propertiesAggregate)
+                        .build();
+        assertThatThrownBy(() -> admin.createTable(tablePath, t1, false).get())
```
wuchong, line 1735: The integration test (IT) case is quite heavy. I suggest limiting it to only one or two common error scenarios. The remaining aggregate function validations should be moved to unit tests—such as `TableDescriptorTest`—to keep the IT suite fast and focused.

wuchong (commented):
@Jackeyzhe thanks for the contribution. I left some comments.

fluss-common/src/main/java/org/apache/fluss/metadata/AggFunctionType.java
Jackeyzhe: Thanks for the useful advice! I'm not sure it is better if add `BOOLEAN` in `DataTypeFamily` so that the logic of validate data type families perfor consistent, how do you think?

wuchong: We can use data types directly here, so we don't need to introduce new data tyep family.

wuchong:
Hi @Jackeyzhe , Please do not use "git merge" to merge branches, otherwise the changes is hard to track. Please use "git rebase" to rebase branches instead. IntelliJ IDEA provide an easy tool to do git rebase, you can find the tool via `VCS -> Git -> Rebase`.

Jackeyzhe:
> Hi @Jackeyzhe , Please do not use "git merge" to merge branches, otherwise the changes is hard to track. Please use "git rebase" to rebase branches instead. IntelliJ IDEA provide an easy tool to do git rebase, you can find the tool via `VCS -> Git -> Rebase`.

sorry for that, I'll reset these commit and use git rebase instead

Jackeyzhe:
@wuchong All addressed, PTAL

PR #2326, opened 2026-01-08
[paimon] Fix union read paimon dv table issue

### LINKED ISSUE #2121: UnionRead dv enabled paimon table will loss data
### Search before asking

### Fluss version

0.8.0 (latest release)

### Please describe the bug 🐞

Currently, union read will read the latest commit snapshot by tiering service, but in dv mode, when read the  latest commit snapshot, the data in l0 level won't be read which case data loss.

### Solution

_No response_

luoyuxia:
### Purpose

Linked issue: close #2121 

The pr fixes the union read functionality for Paimon tables with deletion vectors (DV) enabled. 
The fix enables Flink to correctly perform union reads that combine data from both:
- Paimon lake storage (compacted historical data via readable snapshot)
- Fluss log store (incremental changes after the readable snapshot)

### Brief change log
The basic idea is:
1. Determine when a readable snapshot exists by checking compaction status across all buckets
2. Calculate the correct log offsets for each bucket to start reading incremental changes

The full change log is as follows:
- Add `DvTableReadableSnapshotRetriever` to determine readable snapshots for DV tables by analyzing 
  Paimon snapshot compaction history and L0 file status across all buckets/partitions
- Add `PaimonDvTableUtils` with utilities for working with Paimon DV tables, including partition 
  name conversion and bucket analysis
- Add `LakeCommitResult` to represent lake commit results with readable snapshot information so that the fluss cluster can record it
- Extend `Admin` API with `getReadableLakeSnapshot()` method to retrieve readable snapshots
- Update `FlussTableLakeSnapshotCommitter` to commit readable snapshot information along with 
  tiered snapshots
- Update RPC protocol and message handling to support readable snapshot operations
- Update coordinator and ZooKeeper storage to handle readable snapshot metadata

### Tests

- Add `FlinkUnionReadDvTableITCase` with comprehensive integration tests:
  - `testUnionReadDvTable()`: Tests union read on non-partitioned DV table with 3 buckets
  - `testUnionReadDvPartitionedTable()`: Tests union read on partitioned DV table with 2 
    partitions × 2 buckets

- Add `DvTableReadableSnapshotRetrieverTest` with unit tests covering:
  - Readable snapshot calculation for non-partitioned tables
  - Readable snapshot calculation for partitioned tables
  - Edge cases with partial compaction and L0 file handling
 
- Add `PaimonDvTableUtilsTest` for utility function testing

### API and Format

- **New API**: `Admin.getReadableLakeSnapshot(TablePath)` - Returns a readable snapshot if one 
  exists, throws `LakeTableSnapshotNotExistException` otherwise
- **New API**: `Admin.getLakeSnapshotById(TablePath, long)` - Retrieves a specific historical 
  snapshot by ID

- **RPC Changes**: Added new RPC methods for readable snapshot operations in `FlussApi.proto`

### DIFF
--- a/fluss-client/src/main/java/org/apache/fluss/client/admin/Admin.java
+++ b/fluss-client/src/main/java/org/apache/fluss/client/admin/Admin.java
@@ -443,6 +443,13 @@ public interface Admin extends AutoCloseable {
      * <p>This returns the most recent snapshot regardless of its visibility or compaction status.
      * It includes the latest tiered offsets for all buckets.
      *
+     * <p><b>NOTE: This API is not intended for union reads and should be considered internal.</b>
+     * For union read operations (e.g. Flink/Spark reading from Fluss + lake), use {@link
+     * #getReadableLakeSnapshot(TablePath)} instead. Using this method for union reads can lead to
+     * data loss when the latest tiered snapshot is not yet readable (e.g. Paimon DV tables with
+     * un-compacted L0 data). This method remains for internal use cases such as tiering commit and
+     * readable-offset resolution.
+     *
      * <p>Exceptions expected when calling {@code get()} on the returned future:
      *
      * <ul>
--- a/fluss-common/src/main/java/org/apache/fluss/lake/committer/CommitterInitContext.java
+++ b/fluss-common/src/main/java/org/apache/fluss/lake/committer/CommitterInitContext.java
@@ -53,9 +53,10 @@ public interface CommitterInitContext {
     Configuration lakeTieringConfig();
 
     /**
-     * Returns the fluss config.
+     * Returns the Fluss client configuration. This configuration can be used to build a Fluss
+     * client, such as a Connection.
      *
-     * @return the fluss config
+     * @return the Fluss client configuration
      */
-    Configuration flussConfig();
+    Configuration flussClientConfig();
 }
--- a/fluss-common/src/main/java/org/apache/fluss/lake/committer/LakeCommitResult.java
+++ b/fluss-common/src/main/java/org/apache/fluss/lake/committer/LakeCommitResult.java
@@ -41,7 +41,10 @@ import java.util.Objects;
 @PublicEvolving
 public class LakeCommitResult {
 
-    // -1 to enforce to keep all previous snapshots
+    /** Keep only the latest snapshot; discard all previous ones. */
+    public static final Long KEEP_LATEST = null;
+
+    /** Keep all previous snapshots (infinite retention). */
     public static final Long KEEP_ALL_PREVIOUS = -1L;
 
     // The snapshot ID that was just committed
@@ -49,7 +52,8 @@ public class LakeCommitResult {
 
     private final boolean committedIsReadable;
 
-    // The earliest snapshot ID to keep, null means not to keep any previous snapshot
+    // The earliest snapshot ID to keep. KEEP_LATEST (null) means keep only the latest (discard all
+    // previous). KEEP_ALL_PREVIOUS (-1) means keep all previous snapshots (infinite retention).
     @Nullable private final Long earliestSnapshotIDToKeep;
 
     // the readable snapshot, null if
@@ -69,7 +73,7 @@ public class LakeCommitResult {
     }
 
     public static LakeCommitResult committedIsReadable(long committedSnapshotId) {
-        return new LakeCommitResult(committedSnapshotId, true, null, null);
+        return new LakeCommitResult(committedSnapshotId, true, null, KEEP_LATEST);
     }
 
     public static LakeCommitResult unknownReadableSnapshot(long committedSnapshotId) {
--- a/fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/tiering/committer/FlussTableLakeSnapshotCommitter.java
+++ b/fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/tiering/committer/FlussTableLakeSnapshotCommitter.java
@@ -165,10 +165,15 @@ public class FlussTableLakeSnapshotCommitter implements AutoCloseable {
                         prepareLakeSnapshot(
                                 tableId, tablePath, readableSnapshot.getReadableLogEndOffsets());
 
-                // prepare a tiered bucket offset file for the readable snapshot
+                // reuse tiered path when readable snapshot is the committed snapshot
                 String readableSnapshotTieredOffsetsPath =
-                        prepareLakeSnapshot(
-                                tableId, tablePath, readableSnapshot.getTieredLogEndOffsets());
+                        readableSnapshot.getReadableSnapshotId()
+                                        == lakeCommitResult.getCommittedSnapshotId()
+                                ? lakeBucketTieredOffsetsPath
+                                : prepareLakeSnapshot(
+                                        tableId,
+                                        tablePath,
+                                        readableSnapshot.getTieredLogEndOffsets());
                 // commit the readable snapshot
                 commit(
                         tableId,
@@ -278,7 +283,9 @@ public class FlussTableLakeSnapshotCommitter implements AutoCloseable {
      *     stored
      * @param logEndOffsets the log end offsets for each bucket
      * @param logMaxTieredTimestamps the max tiered timestamps for each bucket
-     * @param earliestSnapshotIDToKeep the earliest snapshot id to keep
+     * @param earliestSnapshotIDToKeep the earliest snapshot ID to keep. Null means keep only the
+     *     latest (discard all previous). -1 ({@link LakeCommitResult#KEEP_ALL_PREVIOUS}) means keep
+     *     all previous snapshots (infinite retention).
      * @return the commit request
      */
     private CommitLakeTableSnapshotRequest toCommitLakeTableSnapshotRequest(
--- a/fluss-lake/fluss-lake-paimon/src/main/java/org/apache/fluss/lake/paimon/utils/DvTableReadableSnapshotRetriever.java
+++ b/fluss-lake/fluss-lake-paimon/src/main/java/org/apache/fluss/lake/paimon/utils/DvTableReadableSnapshotRetriever.java
@@ -51,7 +51,6 @@ import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
-import java.util.Objects;
 import java.util.Set;
 import java.util.stream.Collectors;
 
@@ -160,15 +159,18 @@ public class DvTableReadableSnapshotRetriever implements AutoCloseable {
             return null;
         }
 
+        // todo: optimize in #2626, if the compacted snapshot exists in zk, we can
+        // skip the follow check
+
         Map<TableBucket, Long> readableOffsets = new HashMap<>();
 
         FlussTableBucketMapper flussTableBucketMapper = new FlussTableBucketMapper();
 
         // get all the bucket without l0 files and with l0 files
-        Tuple2<Set<PartitionBucket>, Set<PartitionBucket>> bucketsWithoutL0AndWithL0 =
-                getBucketsWithAndWithoutL0AndWithL0(latestCompactedSnapshot);
-        Set<PartitionBucket> bucketsWithoutL0 = bucketsWithoutL0AndWithL0.f0;
-        Set<PartitionBucket> bucketsWithL0 = bucketsWithoutL0AndWithL0.f1;
+        Tuple2<Set<PaimonPartitionBucket>, Set<PaimonPartitionBucket>> bucketsWithoutL0AndWithL0 =
+                getBucketsWithoutL0AndWithL0(latestCompactedSnapshot);
+        Set<PaimonPartitionBucket> bucketsWithoutL0 = bucketsWithoutL0AndWithL0.f0;
+        Set<PaimonPartitionBucket> bucketsWithL0 = bucketsWithoutL0AndWithL0.f1;
 
         // Track the earliest previousAppendSnapshot ID that was accessed
         // This represents the oldest snapshot that might still be needed
@@ -180,13 +182,16 @@ public class DvTableReadableSnapshotRetriever implements AutoCloseable {
             try {
                 latestTieredSnapshot = flussAdmin.getLatestLakeSnapshot(tablePath).get();
             } catch (Exception e) {
-                throw new IOException(
-                        "Failed to read lake snapshot from Fluss server for snapshot "
-                                + latestCompactedSnapshot.id(),
+                LOG.warn(
+                        "Failed to get latest lake snapshot from Fluss server for compacted snapshot {}; "
+                                + "skipping readable snapshot update.",
+                        latestCompactedSnapshot.id(),
                         e);
+                return null;
             }
+            checkTableConsistent(latestTieredSnapshot);
             // for all buckets without l0, we can use the latest tiered offsets
-            for (PartitionBucket bucket : bucketsWithoutL0) {
+            for (PaimonPartitionBucket bucket : bucketsWithoutL0) {
                 TableBucket tableBucket = flussTableBucketMapper.toTableBucket(bucket);
                 if (tableBucket == null) {
                     // can't map such paimon bucket to fluss, just ignore
@@ -200,7 +205,11 @@ public class DvTableReadableSnapshotRetriever implements AutoCloseable {
         // for all buckets with l0, we need to find the latest compacted snapshot which flushed
         // the buckets, the per-bucket offset should be updated to the corresponding compacted
         // snapshot offsets
-        Set<PartitionBucket> allBucketsToAdvance = new HashSet<>(bucketsWithL0);
+        Set<PaimonPartitionBucket> allBucketsToAdvance = new HashSet<>(bucketsWithL0);
+
+        // Cache LakeSnapshot by snapshot ID to avoid repeated getLakeSnapshot RPCs when many
+        // buckets share the same snapshot.
+        Map<Long, LakeSnapshot> lakeSnapshotBySnapshotId = new HashMap<>();
 
         long earliestSnapshotId = checkNotNull(snapshotManager.earliestSnapshotId());
         // From latestCompacted forward traverse compacted snapshots
@@ -217,9 +226,9 @@ public class DvTableReadableSnapshotRetriever implements AutoCloseable {
                 continue;
             }
             // Get buckets flushed by current compacted snapshot
-            Set<PartitionBucket> flushedBuckets = getBucketsWithFlushedL0(currentSnapshot);
+            Set<PaimonPartitionBucket> flushedBuckets = getBucketsWithFlushedL0(currentSnapshot);
             // For each flushed bucket, if offset not set yet, set it
-            for (PartitionBucket partitionBucket : flushedBuckets) {
+            for (PaimonPartitionBucket partitionBucket : flushedBuckets) {
                 TableBucket tb = flussTableBucketMapper.toTableBucket(partitionBucket);
                 if (tb == null) {
                     // can't map such paimon bucket to fluss,just ignore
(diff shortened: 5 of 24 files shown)


luoyuxia:
@wuchong The pr to fix union read paimon dv table is ready. Please help review when you got time

fluss-rpc/src/main/proto/FlussApi.proto
```diff
@@ -378,11 +364,13 @@ message GetKvSnapshotMetadataResponse {
   repeated PbRemotePathAndLocalFile snapshot_files = 2;
 }
 
-message GetLatestLakeSnapshotRequest {
+message GetLakeSnapshotRequest {
```
wuchong, line 404: Why change the RPC method name? Did you verify whether this will break compatibility with old version clients? The request sent by older client may be recognized by the new server?

fluss-server/src/main/java/org/apache/fluss/server/RpcServiceBase.java
```diff
@@ -451,31 +444,48 @@ public CompletableFuture<ListPartitionInfosResponse> listPartitionInfos(
...
-                        } else {
-                            LakeTableSnapshot lakeTableSnapshot = optLakeTableSnapshot.get();
+                        Optional<LakeTableSnapshot> optSnapshot =
+                                readableSnapshot
+                                        ? zkClient.getLatestReadableLakeTableSnapshot(tableId)
+                                        : zkClient.getLakeTableSnapshot(
+                                                tableId,
+                                                request.hasSnapshotId()
+                                                        ? request.getSnapshotId()
+                                                        : null);
```
wuchong, line 485: What is the behavior when both snapshot_id and readable_snapshot are set? I think we don't support that and should throw exceptions?

fluss-server/src/main/java/org/apache/fluss/server/coordinator/event/CommitLakeTableSnapshotEvent.java
```diff
@@ -18,26 +18,26 @@
...
+            CommitLakeTableSnapshotsData commitLakeTableSnapshotsData,
             CompletableFuture<CommitLakeTableSnapshotResponse> respCallback) {
-        this.commitLakeTableSnapshotData = commitLakeTableSnapshotData;
+        this.commitLakeTableSnapshotsData = commitLakeTableSnapshotsData;
         this.respCallback = respCallback;
     }
 
-    public CommitLakeTableSnapshotData getCommitLakeTableSnapshotData() {
-        return commitLakeTableSnapshotData;
+    public CommitLakeTableSnapshotsData getCommitLakeTableSnapshotData() {
```
wuchong: nit: rename the method name to `getCommitLakeTableSnapshotsData`.

fluss-server/src/main/java/org/apache/fluss/server/entity/CommitLakeTableSnapshotsData.java
```diff
@@ -0,0 +1,178 @@
...
+         * servers, enabling tablet servers to report metrics about max tiered timestamps. In the
+         * future, we plan to have the tiering service directly report metrics, and this field will
+         * be removed.
+         */
+        @Nullable private final Map<TableBucket, Long> tableMaxTieredTimestamps;
+
+        // the following field only non-empty since 0.9
+        @Nullable private final LakeTable.LakeSnapshotMetadata lakeSnapshotMetadata;
+
+        // The earliest snapshot ID to keep for DV tables. Null for non-DV tables.
```
wuchong:
```suggestion
        // The earliest snapshot ID to keep for Paimon DV tables. Null for non-Paimon-DV tables.
```

fluss-common/src/main/java/org/apache/fluss/lake/committer/CommitterInitContext.java
```diff
@@ -51,4 +51,11 @@ public interface CommitterInitContext {
      * @return the lake tiering config
      */
     Configuration lakeTieringConfig();
+
+    /**
+     * Returns the fluss config.
+     *
+     * @return the fluss config
+     */
+    Configuration flussConfig();
```
wuchong: Nit: consider renaming this to `flussClientConfig()`.  

The current name is a bit confusing alongside `lakeTieringConfig()`, since both are instances of Fluss’s `Configuration` class—just serving different purposes. Renaming it to `flussClientConfig()` and adding a clarifying comment (e.g., “This configuration can be used to build a Fluss client, such as a `Connection`”) would make the intent much clearer for users.

fluss-server/src/main/java/org/apache/fluss/server/zk/data/lake/LakeTableHelper.java
```diff
@@ -71,28 +76,94 @@ public void registerLakeTableSnapshotV1(long tableId, LakeTableSnapshot lakeTabl
...
-        LakeTable lakeTable = new LakeTable(lakeSnapshotMetadata);
+        registerLakeTableSnapshotV2(tableId, lakeSnapshotMetadata, null);
+    }
+
+    /**
+     * Register a lake table snapshot and clean up old snapshots based on the table type.
+     *
+     * @param tableId the table ID
+     * @param lakeSnapshotMetadata the new snapshot metadata to register
+     * @param earliestSnapshotIDToKeep the earliest snapshot ID to keep. If null, only the latest
```
wuchong, line 87: Add comment to explain what the `-1` means

fluss-lake/fluss-lake-paimon/src/main/java/org/apache/fluss/lake/paimon/utils/DvTableReadableSnapshotRetriever.java
```diff
@@ -0,0 +1,620 @@
...
+                findPreviousSnapshot(tieredSnapshotId, Snapshot.CommitKind.COMPACT);
+        if (latestCompactedSnapshot == null) {
+            // No compacted snapshot found, may happen when no compaction happens or snapshot
+            // expiration, we can't update readable offsets, return null directly
+            LOG.info(
+                    "Can't find latest compacted snapshot before snapshot {}, skip get readable snapshot.",
+                    tieredSnapshotId);
+            return null;
+        }
+
```
wuchong, line 161: Could you add a TODO comment here to optimize performance? Specifically, we can first look up ZooKeeper to check whether `latestCompactedSnapshot.id` has already been registered in the ZK lake node. If it exists, there’s no need to recompute the tiered offsets and readable offsets for this snapshot ID.

In practice, it’s common to have around 10 `APPEND` snapshots following a single `COMPACT` snapshot, so skipping redundant recomputation would significantly improve the efficiency of the tiering service.

fluss-lake/fluss-lake-paimon/src/main/java/org/apache/fluss/lake/paimon/utils/DvTableReadableSnapshotRetriever.java
```diff
@@ -0,0 +1,620 @@
...
+        if (!bucketsWithoutL0.isEmpty()) {
+            // Get latest tiered offsets
+            LakeSnapshot latestTieredSnapshot;
+            try {
+                latestTieredSnapshot = flussAdmin.getLatestLakeSnapshot(tablePath).get();
+            } catch (Exception e) {
+                throw new IOException(
+                        "Failed to read lake snapshot from Fluss server for snapshot "
+                                + latestCompactedSnapshot.id(),
+                        e);
```
wuchong: Maybe just logging the exception and return null to avoid abort the snapshot?

fluss-lake/fluss-lake-paimon/src/main/java/org/apache/fluss/lake/paimon/utils/DvTableReadableSnapshotRetriever.java
```diff
@@ -0,0 +1,620 @@
...
+
+        // Track the earliest previousAppendSnapshot ID that was accessed
+        // This represents the oldest snapshot that might still be needed
+        long earliestSnapshotIdToKeep = LakeCommitResult.KEEP_ALL_PREVIOUS;
+
+        if (!bucketsWithoutL0.isEmpty()) {
+            // Get latest tiered offsets
+            LakeSnapshot latestTieredSnapshot;
+            try {
+                latestTieredSnapshot = flussAdmin.getLatestLakeSnapshot(tablePath).get();
```
wuchong, line 183: nit: should we check the table id consistency? This table may be re-created and have a different table id

fluss-lake/fluss-lake-paimon/src/main/java/org/apache/fluss/lake/paimon/utils/DvTableReadableSnapshotRetriever.java
```diff
@@ -0,0 +1,620 @@
...
+                    if (earliestSnapshotIdToKeep <= 0
+                            || previousAppendSnapshot.id() < earliestSnapshotIdToKeep) {
+                        earliestSnapshotIdToKeep = previousAppendSnapshot.id();
+                    }
+
+                    try {
+                        LakeSnapshot lakeSnapshot =
+                                flussAdmin
+                                        .getLakeSnapshot(tablePath, previousAppendSnapshot.id())
+                                        .get();
```
wuchong: This is invoked for every buckets, if we have a bunch of buckets (like 10K), this will cost a lot of time. I think we can use a simple HashMap to cache the snapshot information, because most of the buckets should request the same snpashot id.

fluss-lake/fluss-lake-paimon/src/main/java/org/apache/fluss/lake/paimon/utils/DvTableReadableSnapshotRetriever.java
```diff
@@ -0,0 +1,620 @@
...
+     * <p>For Paimon DV tables, we check the snapshot's data files to determine which buckets have
+     * no L0 delta files. A bucket has no L0 files if all its data is in base files.
+     *
+     * <p>For partitioned tables, we include partition information in the returned TableBucket
+     * objects.
+     *
+     * @param snapshot the snapshot to check
+     * @return set of TableBucket that have no L0 files
+     */
+    private Tuple2<Set<PartitionBucket>, Set<PartitionBucket>> getBucketsWithAndWithoutL0AndWithL0(
```
wuchong: rename to `getBucketsWithoutL0AndWithL0`

fluss-lake/fluss-lake-paimon/src/main/java/org/apache/fluss/lake/paimon/utils/DvTableReadableSnapshotRetriever.java
```diff
@@ -0,0 +1,620 @@
...
+         * <p>This is the earliest ID among all snapshot that were accessed via {@code
+         * getLakeSnapshot} during the retrieve readable offset. Snapshots before this ID can
+         * potentially be safely deleted.
+         */
+        public long getEarliestSnapshotIdToKeep() {
+            return earliestSnapshotIdToKeep;
+        }
+    }
+
+    private static final class PartitionBucket {
```
wuchong: rename to `PaimonPartitionBucket` and make it public in the module to share with other clases.

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/tiering/committer/FlussTableLakeSnapshotCommitter.java
```diff
@@ -121,21 +126,92 @@ String prepareLakeSnapshot(
         }
     }
 
-    void commit(
+    public void commit(
```
wuchong, line 129: This is not used by production code, can be removed? Because it seems a duplicate code with `org.apache.fluss.flink.tiering.committer.TieringCommitOperator#commitToFluss`

fluss-common/src/main/java/org/apache/fluss/lake/committer/LakeCommitResult.java
```diff
@@ -0,0 +1,183 @@
...
+            @Nullable ReadableSnapshot readableSnapshot,
+            @Nullable Long earliestSnapshotIDToKeep) {
+        this.committedSnapshotId = committedSnapshotId;
+        this.committedIsReadable = committedIsReadable;
+        this.readableSnapshot = readableSnapshot;
+        this.earliestSnapshotIDToKeep = earliestSnapshotIDToKeep;
+    }
+
+    public static LakeCommitResult committedIsReadable(long committedSnapshotId) {
+        return new LakeCommitResult(committedSnapshotId, true, null, null);
```
wuchong: Can we introduce a `public static final Long KEEP_LATEST = null;` and use `KEEP_LATEST` rather than `null` here? Maybe also other places.

fluss-lake/fluss-lake-paimon/src/main/java/org/apache/fluss/lake/paimon/utils/PaimonDvTableUtils.java
```diff
@@ -0,0 +1,193 @@
...
+
+    /**
+     * Get L0 files from a snapshot, grouped by bucket.
+     *
+     * <p>This method uses the scan API to get all L0 files that exist in the snapshot.
+     *
+     * @param snapshot the snapshot to get L0 files from
+     * @return a map from bucket ID to set of L0 file names that exist in the snapshot
+     */
+    private static Map<Tuple2<BinaryRow, Integer>, Set<String>> getL0FilesByBucket(
```
wuchong: Use the `PaimonPartitionBucket` (the `DvTableReadableSnapshotRetriever.PartitionBucket`) to replace the Tuple2 for simplification.

fluss-lake/fluss-lake-paimon/src/main/java/org/apache/fluss/lake/paimon/utils/PaimonDvTableUtils.java
```diff
@@ -0,0 +1,193 @@
...
+        return null;
+    }
+
+    /**
+     * Get deleted L0 files grouped by bucket from a compacted snapshot's delta.
+     *
+     * @param compactedSnapshot the compacted snapshot
+     * @return a map from bucket ID to set of deleted L0 file names
+     */
+    private static Map<Tuple2<BinaryRow, Integer>, Set<String>> getDeletedL0FilesByBucket(
```
wuchong: ditto. use `PaimonPartitionBucket`

fluss-server/src/main/java/org/apache/fluss/server/zk/data/lake/LakeTable.java
```diff
@@ -97,7 +97,19 @@ private LakeTable(
     @Nullable
     public LakeSnapshotMetadata getLatestLakeSnapshotMetadata() {
         if (lakeSnapshotMetadatas != null && !lakeSnapshotMetadatas.isEmpty()) {
-            return lakeSnapshotMetadatas.get(0);
+            return lakeSnapshotMetadatas.get(lakeSnapshotMetadatas.size() - 1);
```
wuchong, line 103: Currently, `org.apache.fluss.flink.tiering.committer.TieringCommitOperator#commitToFluss` may commit both the readable snapshot and the committed snapshot simultaneously, but this operation is **not atomic**. As a result, there’s a risk that users might observe the **readable snapshot as the latest snapshot**, even though it was committed long before the actual latest (committed) snapshot. This leads to incorrect visibility semantics.

To address this, we should sort the `lakeSnapshotMetadatas` list by snapshot timestamp and ensure only the **most recent snapshot** (i.e., the last one after sorting) is treated as the effective committed snapshot for external visibility.

Since this is a rare edge case, we can add a `TODO` with a reference to a tracking issue and defer the fix to a follow-up.

fluss-client/src/main/java/org/apache/fluss/client/admin/Admin.java
```diff
@@ -429,21 +413,67 @@ CompletableFuture<KvSnapshotMetadata> getKvSnapshotMetadata(
...
-     *   <li>{@link LakeTableSnapshotNotExistException} if no any lake snapshot exist.
+     *   <li>{@link TableNotExistException}: If the table does not exist.
+     *   <li>{@link LakeTableSnapshotNotExistException}: If no any snapshots.
      * </ul>
      *
-     * @param tablePath the table path of the table.
+     * @param tablePath The path of the target table.
+     * @return A future returning the latest tiered snapshot.
      */
     CompletableFuture<LakeSnapshot> getLatestLakeSnapshot(TablePath tablePath);
```
wuchong, line 463: Please add a prominent **NOTE** in the Javadoc clearly warning that this API is **not intended for union reads**. Instead, users should use `getReadableLakeSnapshot()`. This method should be considered an internal API now.

fluss-lake/fluss-lake-paimon/src/test/java/org/apache/fluss/lake/paimon/LakeEnabledTableCreateITCase.java
```diff
@@ -994,6 +971,29 @@ void testEnableLakeTableWithLegacySystemTimestampColumn() throws Exception {
...
+                                        .build())
+                        .property(ConfigOptions.TABLE_DATALAKE_ENABLED, true)
+                        .property("paimon.deletion-vectors.enabled", "true")
+                        .partitionedBy("c3")
+                        .build();
+
+        assertThatThrownBy(() -> admin.createTable(tablePath, tableDescriptor, false).get())
+                .rootCause()
+                .hasMessageContaining(
+                        "Only support String type as partitioned key when 'deletion-vectors.enabled' is set to true for paimon, found 'c3' is not String type.");
```
wuchong, line 1017: Could you please add an **upgrade note** highlighting this behavior change?

In the new version, for partitioned tables where the partition key columns are **not of string type**, enabling `table.datalake.enabled` with Paimon format is **no longer allowed** when `paimon.deletion-vectors.enabled` is set to `true`. This combination will result in a configuration error. This will be supported in the near future.

fluss-server/src/main/java/org/apache/fluss/server/RpcServiceBase.java
luoyuxia: good suggestion

fluss-common/src/main/java/org/apache/fluss/lake/committer/LakeCommitResult.java
luoyuxia: good suggestion

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/tiering/committer/FlussTableLakeSnapshotCommitter.java
luoyuxia: Thanks for pointing that. I have make `TieringCommitOperator` to reuse this method.

fluss-lake/fluss-lake-paimon/src/main/java/org/apache/fluss/lake/paimon/utils/DvTableReadableSnapshotRetriever.java
luoyuxia: Thanks for the suggestion. A good point.  I'll optimize it in the follow pull request.

fluss-lake/fluss-lake-paimon/src/main/java/org/apache/fluss/lake/paimon/utils/DvTableReadableSnapshotRetriever.java
luoyuxia: Good suggestion. I think we can do the check it to fast break the time-cosuming loop.

fluss-rpc/src/main/proto/FlussApi.proto
luoyuxia: Just find the rpc method name out of date. Yes, I already verify that request sent by older client can be recognized by the new server.

luoyuxia:
@wuchong Thanks for your review. Comments has been addressed.

PR #2331, opened 2026-01-08
[lake/paimon] Compare paimon schema and Fluss schema before alter table.

### LINKED ISSUE #2330: Compare Paimon schema and Fluss schema when alter table.
### Search before asking

### Fluss version

0.8.0 (latest release)

### Please describe the bug 🐞

When altering a table, compare the Paimon schema with the Fluss schema. If the schemas are not identical, throw an exception. If the Paimon schema is broader than the Fluss schema, verify whether multiple changes are idempotent.

### Solution

_No response_

loserwang1024:
### Purpose

Linked issue: close #2330

### DIFF
--- a/fluss-lake/fluss-lake-paimon/src/main/java/org/apache/fluss/lake/paimon/PaimonLakeCatalog.java
+++ b/fluss-lake/fluss-lake-paimon/src/main/java/org/apache/fluss/lake/paimon/PaimonLakeCatalog.java
@@ -147,7 +147,7 @@ public class PaimonLakeCatalog implements LakeCatalog {
             }
 
             if (!paimonSchemaChanges.isEmpty()) {
-                alterTable(tablePath, paimonSchemaChanges);
+                paimonCatalog.alterTable(toPaimon(tablePath), paimonSchemaChanges, false);
             }
         } catch (Catalog.ColumnAlreadyExistException | Catalog.ColumnNotExistException e) {
             // This shouldn't happen for AddColumn operations
@@ -199,15 +199,6 @@ public class PaimonLakeCatalog implements LakeCatalog {
             paimonCatalog.createDatabase(databaseName, true);
         } catch (Catalog.DatabaseAlreadyExistException e) {
             // do nothing, shouldn't throw since ignoreIfExists
-        }
-    }
-
-    private void alterTable(TablePath tablePath, List<SchemaChange> tableChanges)
-            throws Catalog.ColumnAlreadyExistException, Catalog.ColumnNotExistException {
-        try {
-            paimonCatalog.alterTable(toPaimon(tablePath), tableChanges, false);
-        } catch (Catalog.TableNotExistException e) {
-            throw new TableNotExistException("Table " + tablePath + " does not exist.");
         }
     }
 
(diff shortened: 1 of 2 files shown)


fluss-lake/fluss-lake-paimon/src/main/java/org/apache/fluss/lake/paimon/PaimonLakeCatalog.java
```diff
@@ -133,69 +138,118 @@ public void alterTable(TablePath tablePath, List<TableChange> tableChanges, Cont
...
+            if (flussColumns.size() + tableChanges.size() != paimonFields.size()) {
+                throw new InvalidAlterTableException(errorMsg);
+            }
+
+            for (int i = 0; i < paimonFields.size() - flussColumns.size(); i++) {
+                DataField paimonDataField = paimonFields.get(i + flussColumns.size());
+                TableChange tableChange = tableChanges.get(i);
+                if (!(tableChange instanceof TableChange.AddColumn
+                        && ((TableChange.AddColumn) tableChange).getPosition()
+                                == TableChange.ColumnPosition.last()
```
beryllw: Maybe we could perform this check earlier—it’s not related to the existing Paimon schema. Also, it’s already validated here:
https://github.com/apache/fluss/blob/256bd7f3838a72965720eb2a8c6c7c0575f46b07/fluss-server/src/main/java/org/apache/fluss/server/coordinator/SchemaUpdate.java#L101-L104

fluss-lake/fluss-lake-paimon/src/main/java/org/apache/fluss/lake/paimon/PaimonLakeCatalog.java
```diff
@@ -133,69 +138,118 @@ public void alterTable(TablePath tablePath, List<TableChange> tableChanges, Cont
...
+                        && isColumnAlreadyExists(
+                                paimonDataField, (TableChange.AddColumn) tableChange))) {
+                    throw new InvalidAlterTableException(errorMsg);
```
beryllw: Maybe we could split the checks and provide distinct error messages to help users more easily identify issues related to the Paimon schema.

fluss-lake/fluss-lake-paimon/src/main/java/org/apache/fluss/lake/paimon/PaimonLakeCatalog.java
loserwang1024: @beryllw 

> It’s not related to the existing Paimon schema

If paimon schema is same as fluss schema, all the operations(both add column and set config) can be accepted. Otherwise, only add columns can be accepted to make the schema same.

wuchong:
@loserwang1024 as we discussed offline, we can refactor the implementation to reuse the logic of `PaimonTableValidation#validatePaimonSchemaCompatible` instead of checking fields by fields.

wuchong:
@loserwang1024 the test coverage of `PaimonLakeCatalog` is failed, could you run `PaimonLakeCatalogTest` with coverage and improve the line coverage of `PaimonLakeCatalog` by adding tests `PaimonLakeCatalogTest`? 


```
Warning:  Rule violated for class org.apache.fluss.lake.paimon.PaimonLakeCatalog: lines covered ratio is 0.57, but expected minimum is 0.70
```

image

loserwang1024:
@wuchong some old codes are remained, I have removed them

PR #2347, opened 2026-01-11
[flink] changelog read support for pk table without pushdown optimizations

### LINKED ISSUE #2333: [flink] Basic $changelog read support without pushdown optimizations for primary key table
### Search before asking

### Description

introducing virtual table $changelog for primary key table without pushdown optimization

### Willingness to contribute

MehulBatra:
### Purpose

Linked issue: close #2333

### Brief change log

ChangelogFlinkTableSource.java
  Purpose: Flink table source for $changelog virtual tables
  ────────────────────────────────────────
  File: ChangelogDeserializationSchema.java
  Purpose: Deserializes LogRecord → RowData with metadata
  ────────────────────────────────────────
  File: ChangelogRowConverter.java
  Purpose: Converts records, adds _change_type, _log_offset,
    _commit_timestamp
  ────────────────────────────────────────
  File: RecordToFlinkRowConverter.java
  Purpose: Interface for record converters
  ────────────────────────────────────────
  File: FlinkTableFactory.java
  Change: Routes to ChangelogFlinkTableSource for changelog
    tables
  ────────────────────────────────────────
  File: TableDescriptor.java
  Change: Added CHANGELOG_* column name constants
  ────────────────────────────────────────
  File: TableDescriptorValidation.java
  Change: Validates reserved column names
  Test Coverage
  Test: testChangelogVirtualTableSchema
  Type: Schema
  What it verifies: DESCRIBE shows correct columns
  ────────────────────────────────────────
  File: FlinkCatalog.java
  Change: Detects $changelog suffix, builds virtual table
    schema
  
  - Access via SELECT * FROM table$changelog
  - Metadata columns: _change_type (VARCHAR), _log_offset (BIGINT), _commit_timestamp (TIMESTAMP)
  - All records output as INSERT (change type in column)
  - Supports PK tables with change types: +I, -U, +U, -D

### Tests

  File: ChangelogVirtualTableITCase.java
  Purpose: Integration tests (7 tests)
  ────────────────────────────────────────
  File: ChangelogRowConverterTest.java
  Purpose: Unit tests for converter
  ────────────────────────────────────────

### DIFF
--- a/fluss-flink/fluss-flink-common/src/test/java/org/apache/fluss/flink/source/ChangelogVirtualTableITCase.java
+++ b/fluss-flink/fluss-flink-common/src/test/java/org/apache/fluss/flink/source/ChangelogVirtualTableITCase.java
@@ -403,17 +403,16 @@ abstract class ChangelogVirtualTableITCase extends AbstractTestBase {
 
         // Collect initial inserts
         List<String> results = collectRowsWithTimeout(rowIter, 3, false);
-        List<String> expectedResults =
-                Arrays.asList(
+        assertThat(results)
+                .containsExactlyInAnyOrder(
                         "+I[+I, 1, Item-1, us]", "+I[+I, 2, Item-2, us]", "+I[+I, 3, Item-3, eu]");
-        assertThat(results).isEqualTo(expectedResults);
 
         // Update a record in a specific partition
         CLOCK.advanceTime(Duration.ofMillis(100));
         tEnv.executeSql("INSERT INTO partitioned_test VALUES (1, 'Item-1-Updated', 'us')").await();
         List<String> updateResults = collectRowsWithTimeout(rowIter, 2, false);
-        expectedResults = Arrays.asList("+I[-U, 1, Item-1, us]", "+I[+U, 1, Item-1-Updated, us]");
-        assertThat(updateResults).isEqualTo(expectedResults);
+        assertThat(updateResults)
+                .containsExactly("+I[-U, 1, Item-1, us]", "+I[+U, 1, Item-1-Updated, us]");
 
         rowIter.close();
     }


wuchong:
Great! Thank you @MehulBatra , I will review it.

.gitignore
```diff
@@ -42,3 +42,4 @@ website/versioned_docs
 website/versioned_sidebars
 website/versions.json
 website/pnpm-lock.yaml
+/.claude/
```
wuchong: Should it be `.claude/`? 

nit: And we can move it near the `.vscode/`

fluss-flink/fluss-flink-common/src/test/java/org/apache/fluss/flink/source/ChangelogVirtualTableITCase.java
```diff
@@ -0,0 +1,450 @@
...
+        // Verify metadata columns are listed first
+        // Format: +I[column_name, type, nullable (true/false), ...]
+        assertThat(schemaRows.get(0)).contains("_change_type");
+        assertThat(schemaRows.get(0)).contains("STRING");
+        // Flink DESCRIBE shows nullability as 'false' for NOT NULL columns
+        assertThat(schemaRows.get(0)).contains("false");
+
+        assertThat(schemaRows.get(1)).contains("_log_offset");
+        assertThat(schemaRows.get(1)).contains("BIGINT");
+        assertThat(schemaRows.get(1)).contains("false");
```
wuchong: The `DESCRIBE` result is deterministic, so we should assert the entire output string rather than using `contains`.

Using `contains` makes the test intent unclear to readers and risks missing regressions in other parts of the output that should also be validated.

fluss-flink/fluss-flink-common/src/test/java/org/apache/fluss/flink/source/ChangelogVirtualTableITCase.java
```diff
@@ -0,0 +1,450 @@
...
+
+        // Validate UPDATE_BEFORE record
+        assertThat(updateResults.get(0)).contains("-U");
+        assertThat(updateResults.get(0)).contains("Item-1");
+        assertThat(updateResults.get(0)).contains("100");
+
+        // Validate UPDATE_AFTER record
+        assertThat(updateResults.get(1)).contains("+U");
+        assertThat(updateResults.get(1)).contains("Item-1-Updated");
+        assertThat(updateResults.get(1)).contains("150");
```
wuchong: Simplify to `assertThat(updateResults.get(0)).contains("-U", "Item-1", "100");`

Update others like this.

fluss-flink/fluss-flink-common/src/test/java/org/apache/fluss/flink/source/ChangelogVirtualTableITCase.java
```diff
@@ -0,0 +1,450 @@
...
+
+        // Test delete operation using the proper delete API
+        deleteRows(conn, tablePath, Arrays.asList(row(2, "Item-2", 200L)));
+
+        // Collect delete record
+        List<String> deleteResult = collectRowsWithTimeout(rowIter, 1, true);
+        assertThat(deleteResult).hasSize(1);
+        // Verify the delete record contains the row data (the change type may be -D or -U)
```
wuchong: If this is a delete, the change type must be `-D`, so we should assert the change type here.

fluss-flink/fluss-flink-common/src/test/java/org/apache/fluss/flink/source/ChangelogVirtualTableITCase.java
```diff
@@ -0,0 +1,450 @@
...
+import static org.apache.fluss.flink.FlinkConnectorOptions.BOOTSTRAP_SERVERS;
+import static org.apache.fluss.flink.source.testutils.FlinkRowAssertionsUtils.collectRowsWithTimeout;
+import static org.apache.fluss.flink.utils.FlinkTestBase.writeRows;
+import static org.apache.fluss.server.testutils.FlussClusterExtension.BUILTIN_DATABASE;
+import static org.apache.fluss.testutils.DataTestUtils.row;
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
+
+/** Integration test for $changelog virtual table functionality. */
+abstract class ChangelogVirtualTableITCase extends AbstractTestBase {
```
wuchong, line 60: This is a thorough test. However, we are facing limited testing resource (the Flink test is reaching to the 1 hour limit). And IT cases are most time-consuming. I think many tests in this IT case are doing duplicated things and we can improve this. 

1. `testChangelogVirtualTableWithPrimaryKeyTable` and `testBasicChangelogScanWithMetadataValidation` do the same thing, but `testBasicChangelogScanWithMetadataValidation` verified more things (metadata). So I think we can remove `testChangelogVirtualTableWithPrimaryKeyTable` or merge it into `testBasicChangelogScanWithMetadataValidation`.

2. `testAllChangeTypes()` and `testBasicChangelogScanWithMetadataValidation` both verified all change types. So I think we can change `testAllChangeTypes` to `testProjectionOnChangelogTable` to test projections. We can add one more column to the source table, and select `_change_type` and other 2 columns. 


3. `testChangelogVirtualTableConcurrentChanges` can be removed as it doesn't test concurrent updates. 

4. `testChangelogVirtualTableWithComplexSchema` can be merged into `testChangelogVirtualTableSchemaIntrospection`.

fluss-common/src/main/java/org/apache/fluss/metadata/TableDescriptor.java
```diff
@@ -60,6 +60,11 @@ public final class TableDescriptor implements Serializable {
     public static final String TIMESTAMP_COLUMN_NAME = "__timestamp";
     public static final String BUCKET_COLUMN_NAME = "__bucket";
 
+    // Reserved column names for $changelog virtual table metadata
+    public static final String CHANGELOG_CHANGE_TYPE_COLUMN = "_change_type";
+    public static final String CHANGELOG_LOG_OFFSET_COLUMN = "_log_offset";
+    public static final String CHANGELOG_COMMIT_TIMESTAMP_COLUMN = "_commit_timestamp";
```
wuchong: Both `$changelog` and `$binlog` share the same 3 columns, so naming `CHANGELOG_` prefix doesn't sound good. I think we can remove the `CHANGELOG_` prefix.

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/catalog/FlinkCatalog.java
```diff
@@ -303,6 +305,18 @@ public CatalogBaseTable getTable(ObjectPath objectPath)
             throws TableNotExistException, CatalogException {
         // may be should be as a datalake table
         String tableName = objectPath.getObjectName();
+
+        // Check if this is a virtual table ($changelog or $binlog)
+        if (tableName.endsWith(CHANGELOG_TABLE_SUFFIX)) {
+            return getVirtualChangelogTable(objectPath);
+        } else if (tableName.endsWith(BINLOG_TABLE_SUFFIX)) {
```
wuchong: Paimon also provides a `$binlog` system table. Using `endsWith("$binlog")` can therefore lead to ambiguity or misinterpretation: for example,  the `order$lake$binlog` would be incorrectly recognized as a Fluss binlog table.

A simple fix would be: 

```java
if (tableName.endsWith(CHANGELOG_TABLE_SUFFIX)
                && !tableName.contains(LAKE_TABLE_SPLITTER)) {
```

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/catalog/FlinkCatalog.java
```diff
@@ -862,4 +876,108 @@ private static boolean isPrefixList(List<String> fullList, List<String> prefixLi
...
+    }
+
+    private Schema buildChangelogSchema(Schema originalSchema) {
+        Schema.Builder builder = Schema.newBuilder();
+
+        // Add metadata columns first
+        builder.column("_change_type", org.apache.flink.table.api.DataTypes.STRING().notNull());
+        builder.column("_log_offset", org.apache.flink.table.api.DataTypes.BIGINT().notNull());
+        builder.column(
+                "_commit_timestamp", org.apache.flink.table.api.DataTypes.TIMESTAMP(3).notNull());
```
wuchong: Can replace qualified name with imports?

And as discussed in FIP, this column should be `TIMESTAMP_LTZ` type.

fluss-flink/fluss-flink-common/src/test/java/org/apache/fluss/flink/source/ChangelogVirtualTableITCase.java
```diff
@@ -0,0 +1,450 @@
...
+
+        // Verify we have equal number of -U and +U
+        long updateBeforeCount = results.stream().filter(r -> r.contains("[-U,")).count();
+        long updateAfterCount = results.stream().filter(r -> r.contains("[+U,")).count();
+        assertThat(updateBeforeCount).isEqualTo(5);
+        assertThat(updateAfterCount).isEqualTo(5);
+    }
+
+    @Test
+    public void testChangelogVirtualTableWithComplexSchema() throws Exception {
```
wuchong: rename this to `testDescribeChangelogTable()` and add `SHOW CREATE TABLE` test for this as well.

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/catalog/FlinkCatalog.java
```diff
@@ -862,4 +876,108 @@ private static boolean isPrefixList(List<String> fullList, List<String> prefixLi
...
+
+        // Add all original columns
+        for (Schema.UnresolvedColumn column : originalSchema.getColumns()) {
+            if (column instanceof Schema.UnresolvedPhysicalColumn) {
+                Schema.UnresolvedPhysicalColumn physicalColumn =
+                        (Schema.UnresolvedPhysicalColumn) column;
+                builder.column(physicalColumn.getName(), physicalColumn.getDataType());
```
wuchong: We can simplify the code by calling `builder.fromColumns(originalSchema.getColumns())` to add all original columns at once. This approach is cleaner and more robust, as it avoids the risk of missing new column attributes (e.g., column comments, new column types) when manually copying fields.

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/catalog/FlinkCatalog.java
```diff
@@ -862,4 +876,108 @@ private static boolean isPrefixList(List<String> fullList, List<String> prefixLi
...
+            Schema originalSchema = baseTable.getUnresolvedSchema();
+            Schema changelogSchema = buildChangelogSchema(originalSchema);
+
+            // Copy options from base table
+            Map<String, String> newOptions = new HashMap<>(baseTable.getOptions());
+            newOptions.put(BOOTSTRAP_SERVERS.key(), bootstrapServers);
+            newOptions.putAll(securityConfigs);
+
+            // Create a new CatalogTable with the modified schema
+            return CatalogTable.of(
```
wuchong: The API `CatalogTable.of` is removed in Flink 2.0, so we should use `CatalogTableAdapter.toCatalogTable` to keep compatiblity across different Flink versions. This is also how `FlinkConversions#toFlinkTable` implemented.

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/catalog/FlinkTableFactory.java
```diff
@@ -262,4 +268,72 @@ private LakeTableFactory mayInitLakeTableFactory() {
...
+        int numMetadataColumns = 3;
+        List<RowType.RowField> dataFields =
+                tableOutputType
+                        .getFields()
+                        .subList(numMetadataColumns, tableOutputType.getFieldCount());
+        RowType dataColumnsType = new RowType(new ArrayList<>(dataFields));
+
+        Map<String, String> catalogTableOptions = context.getCatalogTable().getOptions();
+        FactoryUtil.TableFactoryHelper helper = FactoryUtil.createTableFactoryHelper(this, context);
+        final ReadableConfig tableOptions = helper.getOptions();
```
wuchong, line 314: Do the validations for the table option? 

```java
Optional<DataLakeFormat> datalakeFormat = getDatalakeFormat(tableOptions);
List<String> prefixesToSkip = new ArrayList<>(Arrays.asList("table.", "client."));
datalakeFormat.ifPresent(dataLakeFormat -> prefixesToSkip.add(dataLakeFormat + "."));
helper.validateExcept(prefixesToSkip.toArray(new String[0]));
FlinkConnectorOptionsUtils.validateTableSourceOptions(tableOptions);
```

Maybe we should extract them into a shared method.

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/catalog/FlinkTableFactory.java
```diff
@@ -262,4 +268,72 @@ private LakeTableFactory mayInitLakeTableFactory() {
...
+        final FlinkConnectorOptionsUtils.StartupOptions startupOptions =
+                FlinkConnectorOptionsUtils.getStartupOptions(tableOptions, timeZone);
+
+        ResolvedCatalogTable resolvedCatalogTable = context.getCatalogTable();
+
+        // bucket.key is set from the base table's primary key
+        int[] bucketKeyIndexes = getBucketKeyIndexes(tableOptions, dataColumnsType);
+
+        // For changelog virtual tables, if bucket key is present, base table has a PK
+        int[] primaryKeyIndexes = bucketKeyIndexes;
```
wuchong: In fact, users can define a bucket key on any table, regardless of whether it has a primary key. For example:

- A log table may still specify a bucket key.
- A primary key table may use a bucket key that is different from its primary key.
Therefore, treating the bucket key as the primary key is incorrect.

In this context, the primary key is only used by the enumerator to determine whether to fetch snapshot splits (for tables with a primary key) or log-only splits (for log tables).

Since changelog and binlog virtual tables are purely log-based and do not have a primary key, the enumerator should always fetch log splits for them. Hence, we can safely set their primary key to empty here.

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/source/ChangelogFlinkTableSource.java
```diff
@@ -0,0 +1,238 @@
...
+    private final org.apache.flink.table.types.logical.RowType changelogOutputType;
+    private final org.apache.flink.table.types.logical.RowType dataColumnsType;
+    private final int[] primaryKeyIndexes;
+    private final int[] bucketKeyIndexes;
+    private final int[] partitionKeyIndexes;
+    private final boolean streaming;
+    private final FlinkConnectorOptionsUtils.StartupOptions startupOptions;
+    private final long scanPartitionDiscoveryIntervalMs;
+    private final boolean isDataLakeEnabled;
+    @Nullable private final MergeEngineType mergeEngineType;
```
wuchong: not used, can be removed.

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/source/ChangelogFlinkTableSource.java
```diff
@@ -0,0 +1,238 @@
...
+                public boolean isBounded() {
+                    return true;
+                }
+
+                @Override
+                public Source<RowData, ?, ?> createSource() {
+                    if (!isDataLakeEnabled) {
+                        throw new UnsupportedOperationException(
+                                "Changelog virtual table requires data lake to be enabled for batch queries");
+                    }
```
wuchong: We don't need to throw this exception, because even if this table enabled datalake, we don't initialize the lake source in the above `new FlinkSource`. Here we can just simply return the source. 

And please add a batch mode test case for the $changelog table.

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/source/ChangelogFlinkTableSource.java
```diff
@@ -0,0 +1,238 @@
...
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+
+/** A Flink table source for the $changelog virtual table. */
+public class ChangelogFlinkTableSource implements ScanTableSource {
+
+    private static final String CHANGE_TYPE_COLUMN = "_change_type";
+    private static final String LOG_OFFSET_COLUMN = "_log_offset";
+    private static final String COMMIT_TIMESTAMP_COLUMN = "_commit_timestamp";
```
wuchong: not used, please remove

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/source/ChangelogFlinkTableSource.java
```diff
@@ -0,0 +1,238 @@
...
+     * type.
+     */
+    private org.apache.flink.table.types.logical.RowType extractDataColumnsType(
+            org.apache.flink.table.types.logical.RowType changelogType) {
+        List<org.apache.flink.table.types.logical.RowType.RowField> allFields =
+                changelogType.getFields();
+
+        // Skip the first NUM_METADATA_COLUMNS fields (metadata columns)
+        List<org.apache.flink.table.types.logical.RowType.RowField> dataFields =
+                allFields.subList(NUM_METADATA_COLUMNS, allFields.size());
```
wuchong: nit: it will be safer to skip metadata columns by matching the metadata column names.

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/source/ChangelogFlinkTableSource.java
```diff
@@ -0,0 +1,238 @@
...
+                        tableOptions);
+        copy.producedDataType = producedDataType;
+        copy.projectedFields = projectedFields;
+        copy.partitionFilters = partitionFilters;
+        return copy;
+    }
+
+    @Override
+    public String asSummaryString() {
+        return "ChangelogFlinkTableSource";
```
wuchong: As the summary string is used for logging and displaying on UI by Flink. So it's better to name it `FlussChangelogTableSource` to make it more explicit.

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/source/ChangelogFlinkTableSource.java
```diff
@@ -0,0 +1,238 @@
...
+                throw new IllegalArgumentException(
+                        "Unsupported startup mode: " + startupOptions.startupMode);
+        }
+
+        // Create the source with the changelog deserialization schema
+        FlinkSource<RowData> source =
+                new FlinkSource<>(
+                        flussConfig,
+                        tablePath,
+                        hasPrimaryKey(),
```
wuchong: Can pass `true` directly.

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/utils/ChangelogRowConverter.java
```diff
@@ -0,0 +1,125 @@
...
+/**
+ * A converter that transforms Fluss's {@link LogRecord} to Flink's {@link RowData} with additional
+ * metadata columns for the $changelog virtual table.
+ */
+public class ChangelogRowConverter implements RecordToFlinkRowConverter {
+
+    private static final long serialVersionUID = 1L;
+
+    private final FlussRowToFlinkRowConverter baseConverter;
+    private final RowType flussRowType;
```
wuchong: remove, unused.

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/utils/ChangelogRowConverter.java
```diff
@@ -0,0 +1,125 @@
...
+        fields.add(
+                new org.apache.flink.table.types.logical.RowType.RowField(
+                        TableDescriptor.CHANGELOG_CHANGE_TYPE_COLUMN, new VarCharType(false, 2)));
+        fields.add(
+                new org.apache.flink.table.types.logical.RowType.RowField(
+                        TableDescriptor.CHANGELOG_LOG_OFFSET_COLUMN, new BigIntType(false)));
+        fields.add(
+                new org.apache.flink.table.types.logical.RowType.RowField(
+                        TableDescriptor.CHANGELOG_COMMIT_TIMESTAMP_COLUMN,
+                        new TimestampType(false, 3)));
```
wuchong: This column should be `LocalZonedTimestampType`.

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/source/deserializer/ChangelogDeserializationSchema.java
```diff
@@ -0,0 +1,81 @@
...
+import static org.apache.fluss.flink.utils.FlinkConversions.toFlinkRowType;
+
+/**
+ * A deserialization schema that converts {@link LogRecord} objects to Flink's {@link RowData}
+ * format with additional changelog metadata columns.
+ */
+public class ChangelogDeserializationSchema implements FlussDeserializationSchema<RowData> {
+    private static final long serialVersionUID = 1L;
+
+    private final org.apache.flink.table.types.logical.RowType originalTableType;
```
wuchong: not used, remove

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/utils/PlainRowConverter.java
```diff
@@ -0,0 +1,55 @@
...
+
+import static org.apache.fluss.flink.utils.FlinkConversions.toFlinkRowType;
+
+/**
+ * Standard converter that transforms Fluss's {@link LogRecord} to Flink's {@link RowData} without
+ * any additional metadata columns.
+ *
+ * <p>This is the default converter used for regular (non-virtual) tables.
+ */
+public class PlainRowConverter implements RecordToFlinkRowConverter {
```
wuchong: no used, remove

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/utils/RecordToFlinkRowConverter.java
```diff
@@ -0,0 +1,34 @@
...
+package org.apache.fluss.flink.utils;
+
+import org.apache.fluss.record.LogRecord;
+
+import org.apache.flink.table.data.RowData;
+
+import java.io.Serializable;
+
+/** Interface for converting Fluss {@link LogRecord} to Flink {@link RowData}. */
+public interface RecordToFlinkRowConverter extends Serializable {
```
wuchong, line 27: Since we will remove `PlainRowConverter`, there is only one implementation of this interface. Then this interface abstraction looks useless. I think we can remove this interface class. (add `Serializable` for `ChangelogRowConverter` if we decide to remove `RecordToFlinkRowConverter`).

fluss-server/src/main/java/org/apache/fluss/server/utils/TableDescriptorValidation.java
```diff
@@ -66,7 +69,10 @@ public class TableDescriptorValidation {
                             Arrays.asList(
                                     OFFSET_COLUMN_NAME,
                                     TIMESTAMP_COLUMN_NAME,
-                                    BUCKET_COLUMN_NAME)));
+                                    BUCKET_COLUMN_NAME,
+                                    CHANGELOG_CHANGE_TYPE_COLUMN,
+                                    CHANGELOG_LOG_OFFSET_COLUMN,
+                                    CHANGELOG_COMMIT_TIMESTAMP_COLUMN)));
```
wuchong: Add tests for the columns in `org.apache.fluss.client.admin.FlussAdminITCase#testSystemsColumns`

fluss-flink/fluss-flink-1.20/src/test/java/org/apache/fluss/flink/source/Flink120ChangelogVirtualTableITCase.java
```diff
@@ -0,0 +1,21 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.fluss.flink.source;
+
+/** IT case for {@link ChangelogVirtualTableITCase} in Flink 1.20. */
+public class Flink120ChangelogVirtualTableITCase extends ChangelogVirtualTableITCase {}
```
wuchong, line 21: Create ITCases for all Flink versions to test compatibilites?

fluss-flink/fluss-flink-common/src/test/java/org/apache/fluss/flink/source/ChangelogVirtualTableITCase.java
```diff
@@ -0,0 +1,450 @@
...
+        // Validate that we received 2 INSERT records
+        assertThat(results).hasSize(2);
+
+        // Validate metadata columns are present and correctly formatted
+        for (String result : results) {
+            // Parse the row to validate structure
+            String[] parts = result.substring(3, result.length() - 1).split(", ", 6);
+
+            // Validate change type column
+            assertThat(parts[0]).isEqualTo("+I");
```
wuchong: We can set `ManualClock` to the `FlussClusterExtension` to control the commit timestamp by setting the current time `CLOCK.advanceTime(xxxxx, TimeUnit.MILLISECONDS)`, see the usage in `FlinkTableSourceITCase`.

In this way, we can assert the changelog result by comparing the whole row string. (set bucket to 1 to have consistent log_offset numbers)

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/catalog/FlinkCatalog.java
```diff
@@ -862,4 +876,108 @@ private static boolean isPrefixList(List<String> fullList, List<String> prefixLi
         }
         return true;
     }
+
+    /**
+     * Creates a virtual $changelog table by modifying the base table's to include metadata columns.
+     */
+    private CatalogBaseTable getVirtualChangelogTable(ObjectPath objectPath)
```
wuchong, line 888: Please add tests in `FlinkCatalogITCase` to assert the `CatalogTable` got from a $changelog table.

fluss-flink/fluss-flink-common/src/test/java/org/apache/fluss/flink/source/ChangelogVirtualTableITCase.java
wuchong: Besides, please add tests to verify $changelog also works when `scan.startup.mode` option is applied, and also works for partitioned table. Take `org.apache.fluss.flink.source.FlinkTableSourceITCase#testReadLogTableWithDifferentScanStartupMode` as an example.

fluss-flink/fluss-flink-common/src/test/java/org/apache/fluss/flink/source/ChangelogVirtualTableITCase.java
MehulBatra: I agree, but the challenge was that Flink's DESCRIBE output format varies by version, What I read, but I will try for a  workaround!

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/catalog/FlinkCatalog.java
MehulBatra: Good catch, I would have missed it

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/catalog/FlinkCatalog.java
MehulBatra: I am going ahead with TIMESTAMP_LTZ without precision

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/catalog/FlinkTableFactory.java
MehulBatra: Agreed, will create a helper method to share between dynamictablesource and changelogtablesource, later be used by binlog aswell!

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/source/ChangelogFlinkTableSource.java
MehulBatra: I think it should be this way only, because we are also going to add support for reading changelog for log table aswell (+A)

MehulBatra: or maybe i am missing something

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/utils/RecordToFlinkRowConverter.java
MehulBatra: My only concern is we should keep the interface as I will be using that same interface to implement the binlog converter as well, I believe, but yes i will remove PlainRowConverter as it is of no use.

MehulBatra:
> Thanks @MehulBatra for the contribution. I left some comments.

Thank you @wuchong for the extensive feedback, I addressed all of them, please have a look!

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/catalog/FlinkCatalog.java
wuchong: It should be `TIMESTAMP_LTZ(3)` with precision 3. The `TIMESTAMP_LTZ` without precision is default precision 6, however, `_commit_timestamp` is a mill-second which should be precision 3.

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/source/ChangelogFlinkTableSource.java
wuchong: Sorry, I mean pass `false` directly.

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/utils/RecordToFlinkRowConverter.java
wuchong: That would be good reason to introduce the interface. We can introduce that in next PRs if needed.

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/source/ChangelogFlinkTableSource.java
```diff
@@ -0,0 +1,228 @@
...
+public class ChangelogFlinkTableSource implements ScanTableSource {
+
+    private final TablePath tablePath;
+    private final Configuration flussConfig;
+    // The changelog output type (includes metadata columns: _change_type, _log_offset,
+    // _commit_timestamp)
+    private final org.apache.flink.table.types.logical.RowType changelogOutputType;
+    private final org.apache.flink.table.types.logical.RowType dataColumnsType;
+    private final int[] primaryKeyIndexes;
+    private final int[] bucketKeyIndexes;
```
wuchong: They are not necessary here, we can remove them.

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/utils/ChangelogRowConverter.java
```diff
@@ -0,0 +1,120 @@
...
+        fields.add(
+                new org.apache.flink.table.types.logical.RowType.RowField(
+                        TableDescriptor.CHANGE_TYPE_COLUMN, new VarCharType(false, 2)));
+        fields.add(
+                new org.apache.flink.table.types.logical.RowType.RowField(
+                        TableDescriptor.LOG_OFFSET_COLUMN, new BigIntType(false)));
+        fields.add(
+                new org.apache.flink.table.types.logical.RowType.RowField(
+                        TableDescriptor.COMMIT_TIMESTAMP_COLUMN,
+                        new LocalZonedTimestampType(false, 6)));
```
wuchong: The commit timestamp is a mill-second, so we should use precision 3 here.

fluss-flink/fluss-flink-common/src/test/java/org/apache/fluss/flink/source/ChangelogVirtualTableITCase.java
```diff
@@ -0,0 +1,404 @@
...
+        CloseableIterator<Row> rowIterEarliest = tEnv.executeSql(queryEarliest).collect();
+        List<String> earliestResults = collectRowsWithTimeout(rowIterEarliest, 5, true);
+        assertThat(earliestResults).hasSize(5);
+        // All should be INSERT change types
+        for (String result : earliestResults) {
+            assertThat(result).startsWith("+I[+I,");
+        }
+
+        // 2. Test scan.startup.mode='latest' - should only read new records after subscription
+        String optionsLatest = " /*+ OPTIONS('scan.startup.mode' = 'latest') */";
```
wuchong, line 353: We should test timestamp mode to verify the timestamp seeking also works for changelog table.

fluss-flink/fluss-flink-common/src/test/java/org/apache/fluss/flink/source/ChangelogVirtualTableITCase.java
```diff
@@ -0,0 +1,404 @@
...
+        CloseableIterator<Row> rowIter = tEnv.executeSql(query).collect();
+
+        // Collect initial inserts
+        List<String> results = collectRowsWithTimeout(rowIter, 3, false);
+        assertThat(results).hasSize(3);
+
+        // Verify all are INSERT change types
+        for (String result : results) {
+            assertThat(result).startsWith("+I[+I,");
+        }
```
wuchong: We should assert the entire row or string directly rather than using `contains` whenever possible. This makes the assertion more readable and provides a stronger correctness guarantee.

fluss-flink/fluss-flink-common/src/test/java/org/apache/fluss/flink/catalog/FlinkCatalogITCase.java
```diff
@@ -885,6 +885,53 @@ void testCreateCatalogWithLakeProperties() throws Exception {
...
 
+    @Test
+    void testGetChangelogVirtualTable() throws Exception {
+        // Create a primary key table
+        tEnv.executeSql(
+                "CREATE TABLE pk_table_for_changelog ("
+                        + "  id INT NOT NULL,"
+                        + "  name STRING,"
+                        + "  amount BIGINT,"
+                        + "  PRIMARY KEY (id) NOT ENFORCED"
```
wuchong: We should assert partitioned table as well, to make sure the partitioned keys are correctly kept (support partition pushdown in the future)

wuchong:
@MehulBatra Besides, please do not use "git merge" to merge branches, otherwise the changes is hard to track. Please use "git rebase" to rebase branches instead. IntelliJ IDEA provide an easy tool to do git rebase, you can find the tool via `VCS -> Git -> Rebase`.

PR #2350, opened 2026-01-12
[hotfix] Add authentication to the admin APIs that currently lack it

swuferhong:
### Purpose

Linked issue: close #xxx

Add authentication to the admin APIs that currently lack it

### DIFF
--- a/fluss-server/src/main/java/org/apache/fluss/server/coordinator/CoordinatorService.java
+++ b/fluss-server/src/main/java/org/apache/fluss/server/coordinator/CoordinatorService.java
@@ -34,6 +34,8 @@ import org.apache.fluss.exception.LakeTableAlreadyExistException;
 import org.apache.fluss.exception.SecurityDisabledException;
 import org.apache.fluss.exception.TableAlreadyExistException;
 import org.apache.fluss.exception.TableNotPartitionedException;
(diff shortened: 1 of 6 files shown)


fluss-server/src/main/java/org/apache/fluss/server/coordinator/CoordinatorService.java
```diff
@@ -262,13 +289,7 @@ public CompletableFuture<CreateDatabaseResponse> createDatabase(CreateDatabaseRe
...
     @Override
     public CompletableFuture<DropDatabaseResponse> dropDatabase(DropDatabaseRequest request) {
-        if (authorizer != null) {
-            authorizer.authorize(
-                    currentSession(),
-                    OperationType.DROP,
-                    Resource.database(request.getDatabaseName()));
-        }
-
+        authorizeDatabase(OperationType.DROP, request.getDatabaseName());
```
LiebingYu, line 292: I think Drop database is a cluster level operation?

swuferhong: After offline discussion. This is not a bug.

LiebingYu:
+1

PR #2364, opened 2026-01-13
[lake] Make FlussLakeTiering pluggable to customize tiering job const…

### LINKED ISSUE #2281: Make FlussLakeTieringEntrypoint pluggable
### Search before asking

### Description

In our internal user, we will get lake related options from env, but the `FlussLakeTieringEntrypoint` is not pluggable for the lake related options, which force us to copy the whole `FlussLakeTieringEntrypoint`, just to add a piece of code to get lake related options. It's easy to for us forget copy the community fix of `FlussLakeTieringEntrypoint`. I'm look forward a better to avoid copy the whole `FlussLakeTieringEntrypoint`.

### Willingness to contribute

luoyuxia:
…ruct

### Purpose

Linked issue: close #2281 

### Brief change log

Extract the core logic of parse arguments & build tiering job to `FlussLakeTiering`, the external implement can overwrite `FlussLakeTiering`. 

Some thing like 

```
public class FlussAliLakeTieringEntrypoint {

    public static void main(String[] args) throws Exception {
        return new FlussAliLakeTiering()
    }
}

public class FlussAliLakeTiering extends FlussLakeTiering{
 // todo
}
```

### DIFF
--- a/fluss-flink/fluss-flink-tiering/src/main/java/org/apache/fluss/flink/tiering/FlussLakeTiering.java
+++ b/fluss-flink/fluss-flink-tiering/src/main/java/org/apache/fluss/flink/tiering/FlussLakeTiering.java
@@ -27,6 +27,7 @@ import org.apache.flink.core.execution.JobClient;
 import org.apache.flink.streaming.api.environment.StreamExecutionEnvironment;
 
 import java.util.Map;
+import java.util.ServiceLoader;
 
 import static org.apache.flink.runtime.executiongraph.failover.FailoverStrategyFactoryLoader.FULL_RESTART_STRATEGY_NAME;
 import static org.apache.fluss.flink.tiering.source.TieringSourceOptions.DATA_LAKE_CONFIG_PREFIX;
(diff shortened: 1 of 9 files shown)


wuchong:
Thanks @luoyuxia! However, the current approach still feels quite hacky—it relies on class overriding and could lead to various issues, such as compilation failures or runtime incompatibilities.  

If our goal is to make this truly pluggable, we should design a proper plugin mechanism instead—ideally based on a standard pattern like Java SPI.  

For example, if the only environment-specific customization needed is around Flink execution environment and configuration, we could introduce a clean extension point like this:

```java
interface LakeTieringDecoratorPlugin extends Plugin {
    String identifier();
    LakeTieringDecorator createLakeTieringDecorator();
}

interface LakeTieringDecorator {
    // Customize Flink env or configs as needed
    void decorate(
        StreamExecutionEnvironment env,
        Configuration flussConfig,
        Configuration dataLakeConfig,
        Configuration lakeTieringConfig,
        String dataLakeFormat
    );
}
```

This way, different environments (e.g., internal vs. public cloud) can provide their own decorator implementations without touching core code, ensuring better maintainability, testability, and forward compatibility.

Issue #2369, opened 2026-01-14
FirstRowMergeEngine: Read Stuck for Hours Due to Empty Log Entries in Projection Pushdown Queries


loserwang1024:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/fluss/issues) and found nothing similar.

### Fluss version

0.8.0 (latest release)

### Problem

When reading from a FirstRowMergeEngine table, the read operation occasionally gets stuck for several hours. During this period, the fetch delay suddenly increases from around 100ms to several hours. Upon investigating the logs, I observed the following behavior:

The fetch logs continuously print, but no data is actually read:

```java
2026-01-12 20:13:57,856 INFO  org.apache.fluss.client.table.scanner.log.LogFetcher         [] - Preparing fetch request for bucket TableBucket{tableId=17, bucket=1} to server 1
2026-01-12 20:13:57,860 INFO  org.apache.fluss.client.table.scanner.log.LogFetcher         [] - Preparing fetch request for bucket TableBucket{tableId=17, bucket=1} to server 1

.... the repeated log...
2026-01-12 21:51:21,371  INFO  org.apache.fluss.client.table.scanner.log.LogFetcher         [] - Preparing fetch request for bucket TableBucket{tableId=17, bucket=1} to server 1

2026-01-12 21:51:22,390  INFO  org.apache.fluss.client.table.scanner.log.RemoteLogDownloader [] - Successfully downloaded remote log segment file  84cbd058-d9f7-487c-b401-e24c67da3c49_00000000000004368958.log to local cost 8718 ms.
026-01-12 21:51:23,491 [DownloadRemoteLog-[ods.sales_flat_order_item_create]] INFO  org.apache.fluss.client.table.scanner.log.RemoteLogDownloader [] - Consumed and deleted the fetched log segment file 84cbd058-d9f7-487c-b401-e24c67da3c49_00000000000004368958.log for bucket TableBucket{tableId=17, bucket=1}.
```

This indicates that the fetch requests were repeatedly sent for over an hour without returning any data. This is unexpected because the logs do exist.

### Cause
After reviewing the code, I identified the following issue:
[...]

PR #2370, opened 2026-01-14
[server] When applying projection pushdown, return empty records instead of skipping them to ensure offset movement in the client.

loserwang1024:
### Purpose

Linked issue: the detailed analysis is in  #2369.

### Brief change log

Not filter empty log in projection.

### Tests

org.apache.fluss.client.table.FlussTableITCase#testFirstRowMergeEngine

### DIFF
--- a/fluss-client/src/test/java/org/apache/fluss/client/table/FlussTableITCase.java
+++ b/fluss-client/src/test/java/org/apache/fluss/client/table/FlussTableITCase.java
@@ -19,7 +19,6 @@ package org.apache.fluss.client.table;
 
 import org.apache.fluss.client.Connection;
 import org.apache.fluss.client.ConnectionFactory;
-import org.apache.fluss.client.admin.Admin;
 import org.apache.fluss.client.admin.ClientToServerITCaseBase;
 import org.apache.fluss.client.lookup.LookupResult;
 import org.apache.fluss.client.lookup.Lookuper;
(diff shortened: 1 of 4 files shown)


fluss-server/src/test/java/org/apache/fluss/server/kv/KvTabletTest.java
```diff
@@ -1333,6 +1321,28 @@ private MemoryLogRecords logRecords(
                 DEFAULT_COMPRESSION);
     }
 
+    private MemoryLogRecords logRecords(
```
wuchong: not used, remove

fluss-client/src/test/java/org/apache/fluss/client/table/FlussTableITCase.java
```diff
@@ -1160,6 +1164,26 @@ void testInvalidColumnProjection() throws Exception {
...
+                new MemorySize(5 * V0_RECORD_BATCH_HEADER_SIZE));
+        conf.set(
+                ConfigOptions.CLIENT_SCANNER_LOG_FETCH_MAX_BYTES_FOR_BUCKET,
+                new MemorySize(5 * V0_RECORD_BATCH_HEADER_SIZE));
+        final FlussClusterExtension flussClusterExtension =
+                FlussClusterExtension.builder()
+                        .setNumOfTabletServers(3)
+                        .setClusterConf(conf)
+                        .build();
+        flussClusterExtension.start();
```
wuchong: In my local environment, the original test takes about 15 seconds, but it increases to 55 seconds after applying this PR. I think we should optimize it. Here are several optimization opportunities:

1. The test starts a new Fluss cluster in addition to the existing one from the test base class, which adds significant overhead.
2. The issue only occurs when projection pushdown is enabled (`doProjection = true`), so there’s no need to test `doProjection = false`.
3. The table bucket count can be reduced to 1, as it’s sufficient for reproducing the problem.
4. The number of empty record batches can be reduced from 10 to 2, this should still reliably reproduce the issue while avoiding 10 unnecessary RPC round-trips that dominate the runtime.

I suggest introducing a dedicated test class like `CustomFlussClusterITCase` (without extending `ClientToServerITCaseBase`) for tests that require manual cluster management, and adding a focused test method such as `testProjectionPushdownWithEmptyBatches` that incorporates all these optimizations.

loserwang1024:
@wuchong I have modified this pr

Issue #2202, opened 2025-12-18
Refactor the RPC MetadataResponse to introduce invalid partition set


swuferhong:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/fluss/issues) and found nothing similar.

### Description

Currently, when a `MetadataRequest` is sent, if the requested tableBucket contains partitions that do not exist, fluss server will directly throws an RPC-level exception: `PartitionNotExistException`. This makes it impossible for the client to handle the error gracefully. For example, the `LogFetcher` would crash immediately, affecting the reading of other normal tableBuckets.

To address this, it is proposed to return a set of invalid partitions (i.e., all non-existent partitions) instead of throwing an exception. This allows the `LogFetcher` to report these invalid partitions to the user, who can then customize the next fetch logic (e.g., skip or retry specific partitions).

### Willingness to contribute

- [ ] I'm willing to submit a PR!

MilanTyagi2004:
@swuferhong Hi if the issue is not solve i will look into this, if it is possible.

PR #2373, opened 2026-01-14
[Client] Fail Log Scanner when table is dropped during scan

### LINKED ISSUE #2301: Drop a table during log scanner scan records should fail the scanner
### Search before asking

### Motivation

1: start log scanner to scan a table
2: drop table
3: the scanner won't fail and log warning forever:
```
Received unknown table or bucket error in fetch for bucket TableBucket{tableId=29, bucket=0}
```

Actually I expect the log scanner can fail if found the  table not exists any more.

### Solution

_No response_

### Anything else?

_No response_

### Willingness to contribute

Prajwal-banakar:
### Purpose

Linked issue: close #2301

The purpose of this change is to fix a bug where the LogScanner enters an infinite retry loop when a table is dropped during an active scan. Instead of logging "unknown table or bucket" warnings indefinitely, the scanner now correctly identifies the dropped table via a metadata refresh and terminates with a TableNotExistException.

### Brief change log

Metadata Management: Updated MetadataUtils.java to properly handle cache eviction of dropped tables during metadata updates.
Log Fetcher Logic: Modified LogFetcher.java to catch fetch errors and trigger a metadata refresh to verify if the table still exists.
Exception Handling: Implemented logic to distinguish between a truly deleted table and a recreated table (new ID).
If the table is deleted, it throws TableNotExistException to break the retry loop.
If the table is recreated, it updates the internal TableId and continues fetching, maintaining compatibility with Flink failover requirements.

### Tests

Verified the fix with the following tests:

New Integration Test: Added testDropTableWhileScanning in LogScannerITCase.java which simulates a table drop during a scan and asserts that the scanner fails gracefully.
Regression Testing: Verified that LogFetcherITCase.java passes, confirming the scanner correctly identifies table recreation.
Engine Integration: Verified that Flink118TableSourceFailOverITCase.java passes, ensuring the change does not break Flink's ability to recover from table recreation during a job.
Full Module Verification: Ran ./mvnw test -pl fluss-client -o to ensure no regressions were introduced across the 124 existing client tests.

### API and Format

This change does not affect the external public API of Fluss.
This change does not affect the storage format.

### Documentation

This change does not introduce a new feature and does not require a documentation update.

### DIFF
--- a/fluss-client/src/main/java/org/apache/fluss/client/table/scanner/log/LogFetcher.java
+++ b/fluss-client/src/main/java/org/apache/fluss/client/table/scanner/log/LogFetcher.java
@@ -231,7 +231,7 @@ public class LogFetcher implements Closeable {
             }
 
             if (!tableBuckets.isEmpty()) {
-                checkTableId(tableBuckets.get(0).getTableId());
+                checkTableId(tableBuckets);
             }
         } catch (Exception e) {
             // If exception occurs, we suspect table might be gone or recreated.
@@ -239,7 +239,7 @@ public class LogFetcher implements Closeable {
             try {
                 metadataUpdater.updateTableOrPartitionMetadata(tablePath, null);
                 if (!tableBuckets.isEmpty()) {
-                    checkTableId(tableBuckets.get(0).getTableId());
+                    checkTableId(tableBuckets);
                 }
             } catch (Exception ex) {
                 // If checkTableId threw exception, rethrow it.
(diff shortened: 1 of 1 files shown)


luoyuxia (commented):
@swuferhong Could you please help review this?

pbanakar:
Hi @beryllw @swuferhong  would love your review here! Thanks 🙏

beryllw:
> Hi @beryllw @swuferhong would love your review here! Thanks 🙏

Thank you for your contribution to this PR.❤️

The current approach relies on metadata requests to check table and partition states. However, metadata requests are designed with temporary semantics. Table and partition deletion status should be checked directly against ZooKeeper rather than through the cache.For more details, please see https://github.com/apache/fluss/issues/2202.

Thanks for the effort and would be happy to discuss alternative solutions together.

pbanakar:
Hi @beryllw, thank you for the detailed feedback and the pointer to #2202. I understand the concern relying on the metadata cache to detect table deletion is not reliable since the cache has temporary semantics. Could you guide me on whether there is an interim approach that aligns with the direction of #2202, or should I wait for that refactor to land first and then build on top of it? Happy to rework the approach entirely. Thanks!

PR #2377, opened 2026-01-15
[spark] support batch read from fluss cluster

### LINKED ISSUE #2376: [spark] Support Spark Batch Read
### Search before asking

### Motivation

to support batch read.

### Solution

_No response_

### Anything else?

_No response_

### Willingness to contribute

YannByron:
### Purpose

Linked issue: close #2376

### DIFF
--- a/fluss-common/src/main/java/org/apache/fluss/utils/InternalRowUtils.java
+++ b/fluss-common/src/main/java/org/apache/fluss/utils/InternalRowUtils.java
@@ -54,7 +54,7 @@ public class InternalRowUtils {
             GenericRow genericRow = new GenericRow(row.getFieldCount());
             for (int i = 0; i < row.getFieldCount(); i++) {
                 genericRow.setField(
-                        i, copyRow(fieldGetters[i].getFieldOrNull(row), rowType.getTypeAt(i)));
+                        i, copyValue(fieldGetters[i].getFieldOrNull(row), rowType.getTypeAt(i)));
             }
             return genericRow;
         }
@@ -86,7 +86,7 @@ public class InternalRowUtils {
         Object[] newArray = new Object[from.size()];
         for (int i = 0; i < newArray.length; ++i) {
             if (!from.isNullAt(i)) {
-                newArray[i] = copyRow(elementGetter.getElementOrNull(from, i), eleType);
+                newArray[i] = copyValue(elementGetter.getElementOrNull(from, i), eleType);
             } else {
                 newArray[i] = null;
             }
@@ -94,7 +94,7 @@ public class InternalRowUtils {
         return new GenericArray(newArray);
     }
 
-    public static InternalMap copyMap(InternalMap map, DataType keyType, DataType valueType) {
+    private static InternalMap copyMap(InternalMap map, DataType keyType, DataType valueType) {
         if (map instanceof BinaryMap) {
             return ((BinaryMap) map).copy();
         }
@@ -105,13 +105,13 @@ public class InternalRowUtils {
         InternalArray values = map.valueArray();
         for (int i = 0; i < keys.size(); i++) {
             newMap.put(
-                    copyRow(keyGetter.getElementOrNull(keys, i), keyType),
-                    copyRow(valueGetter.getElementOrNull(values, i), valueType));
+                    copyValue(keyGetter.getElementOrNull(keys, i), keyType),
+                    copyValue(valueGetter.getElementOrNull(values, i), valueType));
         }
         return new GenericMap(newMap);
     }
 
-    public static Object copyRow(Object o, DataType type) {
+    private static Object copyValue(Object o, DataType type) {
         if (o instanceof BinaryString) {
             return ((BinaryString) o).copy();
         } else if (o instanceof InternalRow) {
--- a/fluss-spark/fluss-spark-common/src/main/scala/org/apache/fluss/spark/read/FlussBatch.scala
+++ b/fluss-spark/fluss-spark-common/src/main/scala/org/apache/fluss/spark/read/FlussBatch.scala
@@ -19,7 +19,9 @@ package org.apache.fluss.spark.read
 
 import org.apache.fluss.client.{Connection, ConnectionFactory}
 import org.apache.fluss.client.admin.Admin
+import org.apache.fluss.client.initializer.{BucketOffsetsRetrieverImpl, OffsetsInitializer}
 import org.apache.fluss.client.metadata.KvSnapshots
+import org.apache.fluss.client.table.scanner.log.LogScanner
 import org.apache.fluss.config.Configuration
 import org.apache.fluss.metadata.{PartitionInfo, TableBucket, TableInfo, TablePath}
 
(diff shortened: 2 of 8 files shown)


fluss-common/src/main/java/org/apache/fluss/utils/InternalRowUtils.java
```diff
@@ -22,15 +22,117 @@
...
+        InternalArray values = map.valueArray();
+        for (int i = 0; i < keys.size(); i++) {
+            newMap.put(
+                    copyRow(keyGetter.getElementOrNull(keys, i), keyType),
+                    copyRow(valueGetter.getElementOrNull(values, i), valueType));
+        }
+        return new GenericMap(newMap);
+    }
+
+    public static Object copyRow(Object o, DataType type) {
```
wuchong: This method name conflicts with `copyRow(InternalRow row, RowType rowType)` and may be confusing. Can we rename it to `copyValue` to distinguish with `copyRow`?

Besides, since it is only called `InternalRowUtils`, we can change the visibility of this method and `copyArray`, `copyMap` to `private`.

fluss-spark/fluss-spark-common/src/main/scala/org/apache/fluss/spark/read/FlussPartitionReader.scala
```diff
@@ -0,0 +1,71 @@
...
+
+import java.time.Duration
+
+abstract class FlussPartitionReader(tablePath: TablePath, flussConfig: Configuration)
+  extends PartitionReader[InternalRow] {
+
+  protected val POLL_TIMEOUT: Duration = Duration.ofMillis(100)
+  protected lazy val conn: Connection = ConnectionFactory.createConnection(flussConfig)
+  protected lazy val table: Table = conn.getTable(tablePath)
+  protected lazy val tableInfo: TableInfo = conn.getAdmin.getTableInfo(tablePath).get()
```
wuchong: The `tableInfo` can be got from `table`: `table.getTableInfo`.

fluss-spark/fluss-spark-common/src/main/scala/org/apache/fluss/spark/read/FlussBatch.scala
```diff
@@ -0,0 +1,162 @@
...
+            if (snapshotIdOpt.isPresent) {
+              assert(
+                logOffsetOpt.isPresent,
+                "Log offset must be present when snapshot id is present")
+
+              // Create hybrid partition
+              FlussUpsertInputPartition(
+                tableBucket,
+                snapshotIdOpt.getAsLong,
+                logOffsetOpt.getAsLong
```
wuchong: Since this is a batch InputPartition, we should add an end offset to make the log split bounded. The latest end offset can be got from `OffsetsInitializer.latest().getBucketOffsets(..)` method. 

We should:
1. fetch the latest `kvSnapshots`, it is a `map<bucket, snapshot_id&log_start_offset>`.
2. fetch the latest offset from `OffsetsInitializer.latest`, it is a `map<bucket, log_end_offset>`.
3. Join the `kvSnapshots` and `OffsetsInitializer.latest`, to generate a input partition list for each bucket.

fluss-spark/fluss-spark-common/src/main/scala/org/apache/fluss/spark/read/FlussBatch.scala
```diff
@@ -0,0 +1,162 @@
...
+              )
+            } else {
+              // No snapshot yet, only read log from beginning
+              FlussUpsertInputPartition(tableBucket, -1L, 0L)
```
wuchong: We should use `org.apache.fluss.client.table.scanner.log.LogScanner#EARLIEST_OFFSET` instead of `0` to indicate reading log from beginning. Because the `0L` offset maybe TTLed, and be thrown `LogOffsetOutOfRangeException`.

fluss-spark/fluss-spark-common/src/main/scala/org/apache/fluss/spark/read/FlussUpsertPartitionReader.scala
```diff
@@ -0,0 +1,156 @@
...
+
+    // Phase 2: Read incremental log
+    if (logRecords != null && logRecords.hasNext) {
+      val scanRecord = logRecords.next()
+      currentRow = convertToSparkRow(scanRecord)
+      return true
+    }
+
+    // Poll for more log records
+    val scanRecords: ScanRecords = logScanner.poll(POLL_TIMEOUT)
```
wuchong: `logScanner.poll()` is a best-effort API: it may return an empty result due to transient issues (e.g., network glitches) even when unread log records remain on the server. Therefore, we should poll in a loop until we reach the known end offset.

The end offset should be determined at job startup using `OffsetsInitializer.latest().getBucketOffsets(...)`, which gives us the high-watermark for each bucket at the beginning of the batch job.

Since there’s no built-in API to read a bounded log split, we must manually:

- Skip any records with offsets beyond the precomputed end offset, and
- Signal there is no `next` once all buckets have reached their respective end offsets.

fluss-spark/fluss-spark-common/src/main/scala/org/apache/fluss/spark/read/FlussUpsertPartitionReader.scala
```diff
@@ -0,0 +1,156 @@
...
+    // Get records for our bucket
+    val bucketRecords = scanRecords.records(tableBucket)
+    if (bucketRecords.isEmpty) {
+      return false
+    }
+
+    logRecords = bucketRecords.iterator()
+    if (logRecords.hasNext) {
+      val scanRecord = logRecords.next()
+      currentRow = convertToSparkRow(scanRecord)
```
wuchong: The LogRecord is a changelog that contains `-D` (delete) and `-U` (update-before) records. To produce a consistent view, we need to merge these changes with the KV snapshot data in a union-read fashion—just like how we combine data lake snapshots with changelogs.

Fortunately, the KV snapshot scan is already sorted by primary key. We can leverage this by:

1. Materializing the delta changes into a temporary `delta` table;
2. Sorting the `delta` table by primary key using `org.apache.fluss.row.encode.KeyEncoder#of(...)`;
3. Performing a sort-merge between the sorted KV snapshot reader and the sorted delta table reader.

This enables an efficient and correct merge without requiring random lookups or hash-based joins.

fluss-spark/fluss-spark-ut/src/test/scala/org/apache/fluss/spark/SparkReadTest.scala
```diff
@@ -0,0 +1,342 @@
...
+             |""".stripMargin)
+
+      checkAnswer(
+        sql(s"SELECT * FROM $DEFAULT_DATABASE.t ORDER BY orderId"),
+        Row(600L, 21L, 601, "addr1", "2026-01-01") ::
+          Row(700L, 220L, 602, "addr2_updated", "2026-01-01") ::
+          Row(800L, 23L, 603, "addr3", "2026-01-02") ::
+          Row(900L, 240L, 604, "addr4_updated", "2026-01-02") ::
+          Row(1000L, 25L, 605, "addr5", "2026-01-03") ::
+          Row(1100L, 260L, 606, "addr6", "2026-01-03") ::
```
wuchong, line 199: Currently, the test passes even if changelog reading is not properly implemented. This is because the test base uses a very short KV snapshot interval (1 second), so the reader always falls back to the KV snapshot and never actually consumes the changelog.

I think it’s acceptable to keep this as-is for now, since we plan to refactor the changelog merge-read logic in upcoming PRs, as discussed offline. But please create an issue to track this.

fluss-common/src/main/java/org/apache/fluss/utils/InternalRowUtils.java
YannByron: OK. `copyArray` is used outside, others is changed to `private`.

fluss-spark/fluss-spark-common/src/main/scala/org/apache/fluss/spark/read/FlussBatch.scala
YannByron: OK, then we should move `OffsetsInitializer`-related code to `fluss-client` first.

fluss-spark/fluss-spark-ut/src/test/scala/org/apache/fluss/spark/SparkReadTest.scala
YannByron: https://github.com/apache/fluss/issues/2427

fluss-spark/fluss-spark-common/src/main/scala/org/apache/fluss/spark/read/FlussUpsertPartitionReader.scala
YannByron: Please make sure that `BatchScanner.pollBatch()` is also a best-effort API?

fluss-spark/fluss-spark-common/src/main/scala/org/apache/fluss/spark/read/FlussUpsertPartitionReader.scala
YannByron: So, I should fetch and keep all the changes between starting offset and stopping offset since the delta changes is no sorted, then sort then, and execute a sort-merge with kv snapshot, right?
And, if there are too many changes, `spill` is necessary in the future.

wuchong: @YannByron you are right. But Flink connector also didn't implement `spill` logic yet.

fluss-spark/fluss-spark-common/src/main/scala/org/apache/fluss/spark/read/FlussUpsertPartitionReader.scala
wuchong: @YannByron the differnece is that `LogScanner` is unbounded and `BatchScanner ` is bounded. So when `org.apache.fluss.client.table.scanner.log.LogScanner#poll` returns an empty `ScanRecords`, it doesn't mean the source is finished. When the `BatchScanner.pollBatch()` returns `null` iterator, it means it reaches to the batch end.

YannByron: Understand. Thank you.

PR #2383, opened 2026-01-15
[hotfix] Fix union read fail when from timestamp and projection enabled

luoyuxia:
### Purpose

Linked issue: close #xxx
Fix when union read from lake table, and projectinn is enabled, the `__timestamp` in paimon will not match which cause cast exception 

### Brief change log
Use table origin row type instead of projected row type.

### Tests

Adjust the test to use enable projection

### DIFF
--- a/fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/lake/LakeSplitGenerator.java
+++ b/fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/lake/LakeSplitGenerator.java
@@ -119,7 +119,7 @@ public class LakeSplitGenerator {
                     lakeSplits, isLogTable, tableBucketsOffset, partitionNameById);
         } else {
             Map<Integer, List<LakeSplit>> nonPartitionLakeSplits =
-                    lakeSplits.values().iterator().next();
+                    lakeSplits.isEmpty() ? null : lakeSplits.values().iterator().next();
             // non-partitioned table
             return generateNoPartitionedTableSplit(
                     nonPartitionLakeSplits, isLogTable, tableBucketsOffset);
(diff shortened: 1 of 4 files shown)


fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/lake/LakeSplitGenerator.java
```diff
@@ -119,7 +119,7 @@ public List<SourceSplitBase> generateHybridLakeFlussSplits() throws Exception {
                     lakeSplits, isLogTable, tableBucketsOffset, partitionNameById);
         } else {
             Map<Integer, List<LakeSplit>> nonPartitionLakeSplits =
-                    lakeSplits.values().iterator().next();
+                    lakeSplits.isEmpty() ? null : lakeSplits.values().iterator().next();
```
luoyuxia, line 122: when no split is generate, it may be empty. Let's use the safe way.

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/source/FlinkTableSource.java
```diff
@@ -385,12 +384,11 @@ public boolean isBounded() {
...
     }
 
-    private boolean pushTimeStampFilterToLakeSource(
-            LakeSource<?> lakeSource, RowType flussRowType) {
+    private boolean pushTimeStampFilterToLakeSource(LakeSource<?> lakeSource) {
         // will push timestamp to lake
         // we will have three additional system columns, __bucket, __offset, __timestamp
         // in lake, get the  __timestamp index in lake table
-        final int timestampFieldIndex = flussRowType.getFieldCount() + 2;
+        final int timestampFieldIndex = tableOutputType.getFieldCount() + 2;
```
luoyuxia, line 391: `flussRowType` is projected row, we should use origin `tableOutputType`

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/source/enumerator/FlinkSourceEnumerator.java
```diff
@@ -792,6 +782,17 @@ private void assignPendingSplits(Set<Integer> pendingReaders) {
                                 assignedPartitions.put(partitionId, partitionName);
                             }
                         });
+
+                if (pendingHybridLakeFlussSplits != null) {
```
luoyuxia, line 786: move out from the loop to speed it.

beryllw:
LGTM!

PR #2390, opened 2026-01-16
Add roaring bitmap aggregation

### LINKED ISSUE #2389: [Aggregation] Add RoaringBitmap aggregation support
### Search before asking

### Description

- Introduce new aggregation function types (rbm32/rbm64) and Java factory helpers.
- Implement FieldRoaringBitmap32Agg/FieldRoaringBitmap64Agg and register factories via SPI.
- Add RoaringBitmapUtils for reusable serialize/deserialize helpers with bounded buffer reuse.
- Update aggregation documentation with rbm32/rbm64 usage examples.
- Add unit tests for aggregation behavior and AggregationContext mapping.
- Add RoaringBitmap dependency and shade relocation in fluss-server.

### Willingness to contribute

platinumhamburg:
### Purpose

Linked issue: close #2389 2389

### DIFF
--- a/fluss-server/src/main/java/org/apache/fluss/server/utils/RoaringBitmapUtils.java
+++ b/fluss-server/src/main/java/org/apache/fluss/server/utils/RoaringBitmapUtils.java
@@ -21,9 +21,7 @@ package org.apache.fluss.server.utils;
 import org.roaringbitmap.RoaringBitmap;
 import org.roaringbitmap.longlong.Roaring64Bitmap;
 
-import java.io.ByteArrayInputStream;
 import java.io.ByteArrayOutputStream;
-import java.io.DataInputStream;
 import java.io.DataOutputStream;
 import java.io.IOException;
 import java.nio.ByteBuffer;
@@ -71,9 +69,6 @@ public final class RoaringBitmapUtils {
 
     public static void deserializeRoaringBitmap64(Roaring64Bitmap bitmap, byte[] bytes)
             throws IOException {
-        try (ByteArrayInputStream input = new ByteArrayInputStream(bytes);
-                DataInputStream dataInput = new DataInputStream(input)) {
-            bitmap.deserialize(dataInput);
-        }
+        bitmap.deserialize(ByteBuffer.wrap(bytes));
     }
 }


git-issue-roaringbitmap-agg.md
```diff
@@ -0,0 +1,15 @@
+Title: Add RoaringBitmap aggregation support to Fluss merge engine
```
wuchong: remove this file?

fluss-common/pom.xml
```diff
@@ -62,6 +62,12 @@
             <artifactId>fluss-shaded-arrow</artifactId>
         </dependency>
 
+        <dependency>
+            <groupId>org.roaringbitmap</groupId>
+            <artifactId>RoaringBitmap</artifactId>
+            <version>${roaringbitmap.version}</version>
```
wuchong: I noticed that the RoaringBitmap dependency is currently only used in `fluss-server`. To minimize downstream impact, could we move it from `fluss-common` into `fluss-server`?

Placing it in `fluss-common` forces all clients and connectors to inherit this dependency, which is highly likely to cause version conflicts—Paimon uses RoaringBitmap 1.2.1, while Flink uses 1.3.0, for example.

On the other hand, we may eventually need RoaringBitmap in fluss-common once we support local aggregation (e.g., in SQL or client-side pre-aggregation). To prepare for that while avoiding conflicts, I’ve opened an issue in the `fluss-shaded` repository to shade the `org.roaringbitmap` package:
https://github.com/apache/fluss-shaded/issues/4

For now, scoping the dependency to `fluss-server` seems like the safest approach.

fluss-server/src/main/java/org/apache/fluss/server/utils/RoaringBitmapUtils.java
```diff
@@ -0,0 +1,66 @@
...
+    public static void deserializeRoaringBitmap32(RoaringBitmap bitmap, byte[] bytes)
+            throws IOException {
+        bitmap.deserialize(ByteBuffer.wrap(bytes));
+    }
+
+    public static byte[] serializeRoaringBitmap64(Roaring64Bitmap bitmap) throws IOException {
+        bitmap.runOptimize();
+        try (ByteArrayOutputStream output = new ByteArrayOutputStream();
+                DataOutputStream dataOutput = new DataOutputStream(output)) {
+            bitmap.serialize(dataOutput);
```
wuchong, line 65: I’m curious about the choice to use `DataOutputStream` for serializing rbm64, whereas `serializeRoaringBitmap32` uses ByteBuffer.

Is there a specific reason for this difference in serialization strategy? 

The Javadoc of RoaringBitmap#serialize(java.nio.ByteBuffer) says: 

> Serialize this bitmap to a ByteBuffer. This is the preferred method to serialize to a byte array (byte[]) or to a String

pom.xml
```diff
@@ -91,6 +91,7 @@
         <arrow.version>15.0.0</arrow.version>
         <paimon.version>1.3.1</paimon.version>
         <iceberg.version>1.10.0</iceberg.version>
+        <roaringbitmap.version>1.2.1</roaringbitmap.version>
```
wuchong: It seems the latest stable version is 1.3.0, and both Iceberg[1], Flink [2] use that version. Should we upgrade version to that?



[1] https://github.com/apache/iceberg/blob/e93eaab330581999d8fd8e4e6f5f885446cf28c3/gradle/libs.versions.toml#L86
[2] https://cwiki.apache.org/confluence/pages/viewpage.action?pageId=399278679#FLIP556:IntroduceBITMAPDataType-InternalImplementationDetails

wuchong (commented):
Thanks @platinumhamburg , I left some comments.

fluss-server/src/main/java/org/apache/fluss/server/utils/RoaringBitmapUtils.java
platinumhamburg: This is due to API limitations in Roaring64Bitmap:

RoaringBitmap64 returns sizeInBytes as a long, while the maximum allocation size of a ByteBuffer is Integer.MAX_VALUE. This can lead to a dangerous narrowing conversion, so we should avoid this pattern as much as possible.

platinumhamburg:
Thanks @wuchong for the detailed review. I’ve addressed all of your comments—please take another look when you have time.

PR #2409, opened 2026-01-18
[KV] Kvstore snapshot queries

polyzos:


fluss-client/src/main/java/org/apache/fluss/client/table/scanner/batch/KvBatchScanner.java
```diff
@@ -0,0 +1,285 @@
...
+        if (projectedFields != null) {
+            ProjectedRow projectedRow = ProjectedRow.from(projectedFields);
+            projectedRow.replaceRow(newRow);
+            return projectedRow;
+        } else {
+            return newRow;
+        }
+    }
+
+    public void startKeepAlivePeriodically(int keepAliveIntervalMs) {
```
leekeiabstraction: Where do we plan to start to start keepalives? This method seems currently unused.

Separately, I wonder if we're giving clients too much control over holding server resources with keepalives. I can see issue where clients forget to close and server starts leaking iterators. 

Should we simplify and get rid of keepalives and refresh iterator TTL on server side whenever a ScanKvRequest is made? 

It's a two-way door to release this feature without keepalive. We cannot easily remove keepalive mechanism once added as it would break clients who rely on keepalive mechanism.

fluss-client/src/main/java/org/apache/fluss/client/table/Table.java
```diff
@@ -55,6 +56,12 @@ public interface Table extends AutoCloseable {
      */
     Scan newScan();
 
+    /**
+     * Creates a new {@link SnapshotQuery} for this table to configure and execute a snapshot query
+     * to read all current data in a table bucket (requires to be a Primary Key Table).
+     */
+    SnapshotQuery newSnapshotQuery();
```
leekeiabstraction: Nit: SnapshotQuery sounds like querying an existing snapshot. I cannot think of any name though.. maybe just Query?

fluss-server/src/main/java/org/apache/fluss/server/kv/scan/ScannerManager.java
```diff
@@ -0,0 +1,161 @@
...
+        if (context != null) {
+            context.updateLastAccessTime(clock.milliseconds());
+        }
+        return context;
+    }
+
+    public void keepAlive(byte[] scannerId) {
+        ScannerContext context = scanners.get(ByteBuffer.wrap(scannerId));
+        if (context != null) {
+            context.updateLastAccessTime(clock.milliseconds());
```
xx789633: Is this thread safe? What if the context is cleaned right after you retrieve it from the map?

fluss-server/src/main/java/org/apache/fluss/server/kv/scan/ScannerManager.java
```diff
@@ -0,0 +1,161 @@
...
+    public ScannerManager(Configuration conf) {
+        this(conf, SystemClock.getInstance());
+    }
+
+    public ScannerManager(Configuration conf, Clock clock) {
+        this.clock = clock;
+        this.scannerTtlMs = conf.get(ConfigOptions.SERVER_SCANNER_TTL).toMillis();
+        this.cleanupExecutor =
+                Executors.newSingleThreadScheduledExecutor(
+                        new ExecutorThreadFactory("scanner-cleanup-thread"));
```
xx789633: Can we reuse the scheduler defined in TabletServer?

polyzos: absolutely

fluss-server/src/main/java/org/apache/fluss/server/kv/scan/ScannerManager.java
xx789633: I think every scanner should have an access lock protection, just like what kudu does: https://github.com/apache/kudu/blob/master/src/kudu/tserver/scanners.h#L384

PR #2009, opened 2025-11-21
Bump Hadoop to 3.4.0

nastra:
### Purpose

This bumps Hadoop to 3.4.0 and is required before being able to bump Iceberg to the latest version (https://github.com/apache/fluss/pull/2008)

Linked issue: close https://github.com/apache/fluss/issues/1352

PR #2419, opened 2026-01-20
[filesystem] fix the size of fluss-fs-s3 is too big

### LINKED ISSUE #2412: The size of fluss-fs-s3 is too big
### Search before asking

### Description

in 0.8, the jar is only 30Mb, but in main branch, it's about 500MB.  Seem introduced by #2009 , 

### Willingness to contribute

sd4324530:
### Purpose

Linked issue: close https://github.com/apache/fluss/issues/2412

### Brief change log
Because [pr2009](https://github.com/apache/fluss/pull/2009) upgraded Hadoop to version 3.4.0, it became dependent on the   `aws-java-sdk-v2`'s bundle JAR, which is 500+M in size.
This pr downgrades the Hadoop version in the `fluss-filesystems` module to 3.3.4.

### DIFF
--- a/fluss-filesystems/fluss-fs-s3/pom.xml
+++ b/fluss-filesystems/fluss-fs-s3/pom.xml
@@ -75,10 +75,6 @@
                     <artifactId>jersey-server</artifactId>
                 </exclusion>
                 <exclusion>
-                    <groupId>com.github.pjfanning</groupId>
-                    <artifactId>jersey-json</artifactId>
-                </exclusion>
-                <exclusion>
                     <groupId>org.apache.avro</groupId>
                     <artifactId>avro</artifactId>
                 </exclusion>
--- a/fluss-filesystems/fluss-fs-azure/src/main/resources/META-INF/NOTICE
+++ b/fluss-filesystems/fluss-fs-azure/src/main/resources/META-INF/NOTICE
@@ -32,11 +32,11 @@ This project bundles the following dependencies under the Apache Software Licens
 - org.apache.commons:commons-configuration2:2.8.0
 - org.apache.commons:commons-lang3:3.18.0
 - org.apache.commons:commons-text:1.10.0
-- org.apache.hadoop.thirdparty:hadoop-shaded-guava:1.2.0
+- org.apache.hadoop.thirdparty:hadoop-shaded-guava:1.1.1
 - org.apache.hadoop.thirdparty:hadoop-shaded-protobuf_3_21:1.2.0
 - org.apache.hadoop:hadoop-annotations:3.4.0
 - org.apache.hadoop:hadoop-auth:3.4.0
-- org.apache.hadoop:hadoop-azure:3.4.0
+- org.apache.hadoop:hadoop-azure:3.3.4
 - org.apache.hadoop:hadoop-common:3.4.0
 - org.apache.kerby:kerb-core:2.0.3
 - org.apache.kerby:kerby-asn1:2.0.3
@@ -44,8 +44,10 @@ This project bundles the following dependencies under the Apache Software Licens
 - org.apache.kerby:kerby-util:2.0.3
 - org.apache.httpcomponents:httpclient:4.5.13
 - org.apache.httpcomponents:httpcore:4.4.13
+- org.codehaus.jackson:jackson-core-asl:1.9.13
+- org.codehaus.jackson:jackson-mapper-asl:1.9.13
 - org.codehaus.jettison:jettison:1.5.4
-- org.wildfly.openssl:wildfly-openssl:1.1.3.Final
+- org.wildfly.openssl:wildfly-openssl:1.0.7.Final
 - org.xerial.snappy:snappy-java:1.1.10.4
 
 This project bundles the following dependencies under the BSD license. See bundled license files for details.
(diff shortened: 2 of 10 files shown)


fluss-filesystems/fluss-fs-s3/pom.xml
```diff
@@ -233,6 +233,10 @@
                     <groupId>org.slf4j</groupId>
                     <artifactId>slf4j-reload4j</artifactId>
                 </exclusion>
+                <exclusion>
```
luoyuxia: https://hadoop.apache.org/release/3.4.1.html
"We have also introduced a lean tar which is a small tar file that does not contain the AWS SDK because the size of AWS SDK is itself 500 MB. This can ease usage for non AWS users. Even AWS users can add this jar explicitly if desired."

If we use 3.4.1, the fat jar won't be included.

fluss-filesystems/fluss-fs-s3/src/main/java/org/apache/fluss/fs/s3/exception/InvalidCredentialsException.java
```diff
@@ -0,0 +1,32 @@
...
+ */
+
+package org.apache.fluss.fs.s3.exception;
+
+/** Exception thrown when no credentials were found. */
+public class InvalidCredentialsException extends RuntimeException {
+
+    public static final String E_NO_AWS_CREDENTIALS = "No AWS Credentials";
+
+    public InvalidCredentialsException(String credentialProvider) {
```
luoyuxia: I'm curious about the class not found issue. I also check the v0.8 jar for s3
```jar tf fluss-fs-s3-0.8.0-incubating.jar | grep SdkClientException
com/amazonaws/SdkClientException.class
```
The same with the jar build with this pr
```
com/amazonaws/SdkClientException.class
```
So, it looks to me that the same problem will also happen in v0.8, right?

luoyuxia (commented):
@sd4324530 Thanks for the pr. Left minor comments

fluss-filesystems/fluss-fs-s3/src/main/java/org/apache/fluss/fs/s3/exception/InvalidCredentialsException.java
sd4324530: Thank you so much for reminding me. I checked it again carefully.
Fluss v0.8 depends on hadoop version 3.3.4.
I found that the extends chain of `NoAwsCredentialsException` has changed when upgrading from 3.3.4 to 3.4.0:

hadoop 3.3.4:
```mermaid
classDiagram
direction RL
    class AmazonClientException
    class CredentialInitializationException
    class NoAuthWithAWSException
    class NoAwsCredentialsException
    
    AmazonClientException --|> CredentialInitializationException
    CredentialInitializationException --|> NoAuthWithAWSException
    NoAuthWithAWSException --|> NoAwsCredentialsException
```

hadoop 3.4.0:
```mermaid
classDiagram
direction RL
    class SdkClientException
    class CredentialInitializationException
    class NoAuthWithAWSException
    class NoAwsCredentialsException
    
    SdkClientException--|> CredentialInitializationException
    CredentialInitializationException --|> NoAuthWithAWSException
    NoAuthWithAWSException --|> NoAwsCredentialsException
```
In Hadoop 3.3.4, `AmazonClientException` originates from `aws-java-sdk-core-1.12.319.jar`, which is only 1MB in size.
see: https://github.com/apache/fluss/blob/release-0.8/fluss-filesystems/fluss-fs-s3/pom.xml#L188-L192
In Hadoop 3.4.0, `SdkClientException` originates from `bundle-2.23.19.jar`, which is 500+MB in size.

fluss-filesystems/fluss-fs-s3/pom.xml
sd4324530: Thanks, this should be a better solution. I'll test it and give my conclusion later.

sd4324530: @luoyuxia Hadoop 3.4.1 provides a "tar" file instead of the "jar" file we usually use. Simply adjusting the version number in the pom.xml file will not reduce the final package size.
I think we still need to manually exclude the fat JAR.

fluss-filesystems/fluss-fs-s3/src/main/java/org/apache/fluss/fs/s3/exception/InvalidCredentialsException.java
luoyuxia: Thank you for the explantion.

sd4324530:
@luoyuxia 
Replace the full `aws-sdk-v2` dependency in `hadoop-aws` with the minimum required module.
I successfully tested data writing and reading using `minio`.

sd4324530:
@luoyuxia 
I adapted the code in the `fluss-fs-s3` module to AWS Java SDK v2 and removed the v1 version dependency.
I successfully tested data writing and reading using `minio`.
Please take another review if you have time.

luoyuxia (approved):
@sd4324530 Thanks. LGTM!

litiliu:
@luoyuxia  @sd4324530 Why do we chose to degrade the Hadoop version, instead of excluding the whole aws-bundle and include the necessary aws sub-serice?

PR #2430, opened 2026-01-21
[spark] Add sparksql lake paimon table DDL support

Yohahaha:
### Purpose

Linked issue: close #xxx

### Brief change log

Summary:
1. Add paimon lake table DDL support and related UT which learned from `LakeEnabledTableCreateITCase`, when other lake table format were added, modular refactoring will be needed cause `LakeStoragePlugin` was loaded by SPI.
2. Support alter table with only set properties, left TODO for more support.

### Tests

org.apache.fluss.spark.lake.paimon.SparkLakePaimonCatalogTest


Yohahaha:
@YannByron @wuchong

YannByron:
@Yohahaha I highly recommend splitting this PR into two:

- one is to support alter table, which can support the TableChange type as much as possible.
- another is to verify metadata for tables which enable datalake. If possible, try to extend `SparkCatalogTest`.

Yohahaha:
split alter table support to https://github.com/apache/fluss/pull/2437

PR #2431, opened 2026-01-21
[flink] Support custom properties in Flink table factory

### LINKED ISSUE #2425: Flink Connector Fails to Validate Tables with Custom Properties
## Search before asking

## Fluss version
main (development)

### Problem Description

When a Fluss table contains custom properties (e.g., `owner`, `team`, `created_by`), the Flink connector fails with a `ValidationException` during table source/sink creation. This is a common use case in enterprise environments where users add metadata properties to track table ownership and other organizational information.

### Expected Behavior

The Flink connector should accept and ignore custom properties that are stored in Fluss table's `customProperties` field. According to the Fluss design, custom properties are not understood by Fluss core but are stored as part of the table's metadata to provide a mechanism for users to persist user-defined properties.

### Actual Behavior

The Flink connector throws a `ValidationException` with a message like:
```
Unsupported options:

owner
```

This happens because the `FlinkTableFactory.validateExcept()` method only excludes prefixes like `"table."`, `"client."`, and datalake-specific prefixes (e.g., `"paimon."`), but does not account for arbitrary custom properties.

### How to Reproduce

1. Create a Fluss table with a custom property:
```sql
CREATE TABLE my_table (
  id BIGINT,
  name STRING,
  PRIMARY KEY (id) NOT ENFORCED
) WITH (
  'bucket.num' = '10',
  'owner' = 'user@example.com'
);
```

2. Try to read from or write to this table using Flink:
```sql
SELECT * FROM my_table;
```

3. The connector will fail with `ValidationException: Unsupported options: owner`

### Root Cause Analysis

The issue is in the `FlinkTableFactory` class:

**Source Creation**:
```java
List<String> prefixesToSkip = new ArrayList<>(Arrays.asList("table.", "client."));
datalakeFormat.ifPresent(dataLakeFormat -> prefixesToSkip.add(dataLakeFormat + "."));
helper.validateExcept(prefixesToSkip.toArray(new String[0]));
```

**Sink Creation**:
```java
if (datalakeFormat.isPresent()) {
    helper.validateExcept("table.", "client.", datalakeFormat.get() + ".");
} else {
    helper.validateExcept("table.", "client.");
}
```

The validation logic does not exclude custom properties, even though:
1. Fluss server stores custom properties separately from table properties (see `TableDescriptor.getCustomProperties()`)
2. Custom properties are explicitly designed to be "not understood by Fluss" but stored as metadata

### Proposed Fix

We can simply remove the option validation and allow all custom properties to pass through.

LiebingYu:
### Purpose

Linked issue: close #2425

### DIFF
--- a/fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/catalog/FlinkTableFactory.java
+++ b/fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/catalog/FlinkTableFactory.java
@@ -27,7 +27,6 @@ import org.apache.fluss.flink.sink.shuffle.DistributionMode;
 import org.apache.fluss.flink.source.ChangelogFlinkTableSource;
 import org.apache.fluss.flink.source.FlinkTableSource;
 import org.apache.fluss.flink.utils.FlinkConnectorOptionsUtils;
-import org.apache.fluss.metadata.DataLakeFormat;
 import org.apache.fluss.metadata.TablePath;
 
 import org.apache.flink.api.common.RuntimeExecutionMode;
@@ -55,14 +54,12 @@ import java.util.Collections;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
-import java.util.Optional;
 import java.util.Set;
 
 import static org.apache.fluss.config.ConfigOptions.TABLE_DATALAKE_FORMAT;
 import static org.apache.fluss.config.ConfigOptions.TABLE_DELETE_BEHAVIOR;
 import static org.apache.fluss.config.FlussConfigUtils.CLIENT_PREFIX;
 import static org.apache.fluss.flink.catalog.FlinkCatalog.LAKE_TABLE_SPLITTER;
-import static org.apache.fluss.flink.utils.DataLakeUtils.getDatalakeFormat;
 import static org.apache.fluss.flink.utils.FlinkConnectorOptionsUtils.getBucketKeyIndexes;
 import static org.apache.fluss.flink.utils.FlinkConnectorOptionsUtils.getBucketKeys;
 import static org.apache.fluss.flink.utils.FlinkConversions.toFlinkOption;
@@ -98,7 +95,7 @@ public class FlinkTableFactory implements DynamicTableSourceFactory, DynamicTabl
 
         FactoryUtil.TableFactoryHelper helper = FactoryUtil.createTableFactoryHelper(this, context);
         final ReadableConfig tableOptions = helper.getOptions();
-        validateSourceOptions(helper, tableOptions);
+        validateSourceOptions(tableOptions);
 
         boolean isStreamingMode =
                 context.getConfiguration().get(ExecutionOptions.RUNTIME_MODE)
@@ -161,12 +158,6 @@ public class FlinkTableFactory implements DynamicTableSourceFactory, DynamicTabl
     public DynamicTableSink createDynamicTableSink(Context context) {
         FactoryUtil.TableFactoryHelper helper = FactoryUtil.createTableFactoryHelper(this, context);
         final ReadableConfig tableOptions = helper.getOptions();
-        Optional<DataLakeFormat> datalakeFormat = getDatalakeFormat(tableOptions);
-        if (datalakeFormat.isPresent()) {
-            helper.validateExcept("table.", "client.", "fields.", datalakeFormat.get() + ".");
-        } else {
-            helper.validateExcept("table.", "client.", "fields.");
-        }
 
         boolean isStreamingMode =
                 context.getConfiguration().get(ExecutionOptions.RUNTIME_MODE)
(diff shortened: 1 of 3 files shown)


loserwang1024:
@LiebingYu Should we add a specially prefix to custom properties rather than skip all the check? I think it's dangerous for later compability.

Currently org.apache.fluss.flink.utils.FlinkConnectorOptionsUtils#validateTableSourceOptions only check the scan start up. 

why not when catalog read from table, add a prefixes 'customer.' to customers. Then table factory check it and remove the prefixes? @leonardBang @wuchong , CC, WDYT

leonardBang:
> @LiebingYu Should we add a specially prefix to custom properties rather than skip all the check? I think it's dangerous for later compability.

+1 for adding a specific prefix for custom properties instead of removing all existing validation for near all options, it make the code being weak. Imaging kafka and MySQL CDC connector, we also support custom properties with prefix i.e.: 
Kafka：https://nightlies.apache.org/flink/flink-docs-release-2.2/docs/connectors/table/kafka/#properties

wuchong:
I think it’s a good idea to use a dedicated prefix for user-extensible properties to avoid potential key conflicts in the future.

Instead of `customer.*`, consider clearer and more conventional prefixes like `tags.*`, `ext.*`, or `properties.*`, or even begins with `_` prefix such as `_owner=xxx`. These better convey that the keys represent custom, user-defined metadata. The term `customer` is ambiguous and doesn’t clearly signal extensibility. Moreover, from Fluss’s perspective, any property outside the `table.*` namespace is effectively a custom property.

Examples:
```
tags.owner = 'xxx'
ext.owner = 'xxx'
properties.owner = 'xxx'
_owner = 'xxx'
```

wuchong:
Personally, I prefer `tags.*` because it clearly conveys that these are purely custom labels with no semantic impact on connector or storage behavior.

LiebingYu:
> I think it’s a good idea to use a dedicated prefix for user-extensible properties to avoid potential key conflicts in the future.
> 
> Instead of `customer.*`, consider clearer and more conventional prefixes like `tags.*`, `ext.*`, or `properties.*`, or even begins with `_` prefix such as `_owner=xxx`. These better convey that the keys represent custom, user-defined metadata. The term `customer` is ambiguous and doesn’t clearly signal extensibility. Moreover, from Fluss’s perspective, any property outside the `table.*` namespace is effectively a custom property.
> 
> Examples:
> 
> ```
> tags.owner = 'xxx'
> ext.owner = 'xxx'
> properties.owner = 'xxx'
> _owner = 'xxx'
> ```

Hi @wuchong. My question is: do we really need to define a prefix for custom properties? As you said, **“from Fluss’s perspective, any property outside the `table.*` namespace is effectively a custom property.”** When the Flink connector processes properties, it can filter out those custom ones and simply skip validating them(**like the new implementation**). From this perspective, custom properties don’t necessarily need a prefix, right? 

After all, the Fluss server imposes no restrictions on custom properties, but the Flink read/write path introduces an additional constraint. Users can freely specify custom properties when creating a table, but there are restrictions when Flink reads/writes the table. This does not align with the fail-fast principle.

Or, we also need to restrict the prefixes of custom attributes on the Fluss server side?

wuchong:
@LiebingYu thank you, I think you raised a good point that tables created via other platforms (e.g., Spark or the SDK) aren’t subject to this restriction and may include arbitrary options, which could cause Flink reads to fail.

Therefore, I agree with removing the broad validation like `helper.validateExcept` and instead only validating the options explicitly recognized by this connector—for example, using `FlinkConnectorOptionsUtils#validateTableSourceOptions`.

Additionally, I still recommend that platforms use a dedicated prefix such as `tags.*` for custom metadata. This reduces the risk of naming conflicts with other engines and simplifies maintenance by allowing platform-level options to be easily identified and filtered by prefix.

What do you think @leonardBang @loserwang1024 ?

leonardBang:
> @LiebingYu thank you, I think you raised a good point that tables created via other platforms (e.g., Spark or the SDK) aren’t subject to this restriction and may include arbitrary options, which could cause Flink reads to fail.
> 
> Therefore, I agree with removing the broad validation like `helper.validateExcept` and instead only validating the options explicitly recognized by this connector—for example, using `FlinkConnectorOptionsUtils#validateTableSourceOptions`.
> 
> Additionally, I still recommend that platforms use a dedicated prefix such as `tags.*` for custom metadata. This reduces the risk of naming conflicts with other engines and simplifies maintenance by allowing platform-level options to be easily identified and filtered by prefix.
> 
> What do you think @leonardBang @loserwang1024 ?

After discussed with Jark offline, we think Fluss' case is different with Kafka connector as Kakfa Connector does not maintain Table properties but Fluss does. As the custom table properties maybe injected from different computing engine like Flink, Spark and StarRocks when user user these engine to create Fluss table, removing these validation looks necessary for cross different engine user case and backward compatibility for cross different engine versions user case.

LiebingYu:
Thank you all for involving the discussion.

@leonardBang I've update the code as Jark said. Could you please take a look again?

PR #1937, opened 2025-11-05
pass catalog sensitive info when create fluss catalog

xx789633:
### Purpose

Linked issue: close #1903

PR #2432, opened 2026-01-21
[hotfix] Catalog related options should overwrite table lake related options

luoyuxia:
### Purpose

Linked issue: close #xxx
#1937 doesn't use catalog related options to overwrite table options in table witth `$lake` cause catalog related options don't work.

### DIFF
--- a/fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/lake/LakeFlinkCatalog.java
+++ b/fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/lake/LakeFlinkCatalog.java
@@ -30,6 +30,7 @@ import org.apache.paimon.flink.FlinkFileIOLoader;
 import org.apache.paimon.options.Options;
 
 import java.lang.reflect.Method;
+import java.util.HashMap;
 import java.util.Map;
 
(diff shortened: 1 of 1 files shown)


fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/lake/LakeFlinkCatalog.java
```diff
@@ -72,11 +73,10 @@ public Catalog getLakeCatalog(
                                         + "' is set.");
                     }
                     Map<String, String> catalogProperties =
-                            PropertiesUtils.extractAndRemovePrefix(
-                                    lakeCatalogProperties, lakeFormat + ".");
-
+                            new HashMap<>(DataLakeUtils.extractLakeCatalogProperties(tableOptions));
                     catalogProperties.putAll(
```
xx789633, line 78: Maybe an explain comment is preferred here.

luoyuxia: Update

luoyuxia:
@xx789633 Thanks for review. Comments updated

xx789633:
looks good to me

PR #2434, opened 2026-01-22
[FLUSS] Add Producer Offset Snapshot for Exactly-Once semantics

### LINKED ISSUE #2433: [Server] Add Producer Offset Snapshot Registry as Infrastructure for Exactly-Once Semantics
### Search before asking

### Description

### Willingness to contribute

platinumhamburg:
Core changes:
- Add ProducerSnapshotManager for lifecycle management with atomic registration
- Add ProducerSnapshotStore for ZK + remote storage operations
- Add tryRegisterProducerSnapshot in ZooKeeperClient for atomic check-and-create
- Add Admin API: registerProducerOffsets, getProducerOffsets, deleteProducerOffsets
- Add configurable TTL and cleanup interval for producer snapshots

Design highlights:
- Atomic registration via ZK's NodeExistsException handling
- Eventually consistent: ZK as commit point, orphan files cleaned periodically
- UUID-based file naming prevents concurrent upload conflicts

Tests:
- ProducerSnapshotManagerTest: lifecycle, expiration, concurrent atomicity
- ProducerSnapshotJsonSerdeTest: JSON format compatibility

### Purpose

Linked issue: close #2433

### DIFF
--- a/fluss-server/src/main/java/org/apache/fluss/server/coordinator/CoordinatorService.java
+++ b/fluss-server/src/main/java/org/apache/fluss/server/coordinator/CoordinatorService.java
@@ -268,29 +268,40 @@ public final class CoordinatorService extends RpcServiceBase implements Coordina
     @Override
     public void authorizeTable(OperationType operationType, long tableId) {
         if (authorizer != null) {
-            TablePath tablePath;
-            try {
-                // TODO: this will block on the coordinator event thread, consider refactor
-                //  CoordinatorMetadataCache to hold the mapping of table_id to table_path, and then
-                //  we don't need this async request.
-                AccessContextEvent<TablePath> getTablePathEvent =
-                        new AccessContextEvent<>(ctx -> ctx.getTablePathById(tableId));
-                eventManagerSupplier.get().put(getTablePathEvent);
-                tablePath = getTablePathEvent.getResultFuture().get();
-            } catch (Exception e) {
-                throw new UnknownServerException("Failed to get table path by ID " + tableId, e);
-            }
-
-            if (tablePath == null) {
-                throw new UnknownTableOrBucketException(
-                        String.format(
-                                "This server %s does not know this table ID %s. This may happen when the table "
-                                        + "metadata cache in the server is not updated yet.",
-                                name(), tableId));
-            }
-
-            authorizeTable(operationType, tablePath);
+            authorizeTableWithSession(currentSession(), operationType, tableId);
         }
+    }
+
+    /**
+     * Authorize table access with an explicitly provided session.
+     *
+     * <p>This method is used for async operations where the session must be captured before
+     * entering the async block, since currentSession() relies on thread-local storage.
+     */
+    private void authorizeTableWithSession(
+            Session session, OperationType operationType, long tableId) {
+        TablePath tablePath;
+        try {
+            // TODO: this will block on the coordinator event thread, consider refactor
+            //  CoordinatorMetadataCache to hold the mapping of table_id to table_path, and then
+            //  we don't need this async request.
+            AccessContextEvent<TablePath> getTablePathEvent =
+                    new AccessContextEvent<>(ctx -> ctx.getTablePathById(tableId));
+            eventManagerSupplier.get().put(getTablePathEvent);
+            tablePath = getTablePathEvent.getResultFuture().get();
+        } catch (Exception e) {
+            throw new UnknownServerException("Failed to get table path by ID " + tableId, e);
+        }
+
+        if (tablePath == null) {
+            throw new UnknownTableOrBucketException(
+                    String.format(
+                            "This server %s does not know this table ID %s. This may happen when the table "
+                                    + "metadata cache in the server is not updated yet.",
+                            name(), tableId));
+        }
+
+        authorizer.authorize(session, operationType, Resource.table(tablePath));
     }
 
     @Override
@@ -1068,6 +1079,8 @@ public final class CoordinatorService extends RpcServiceBase implements Coordina
     public CompletableFuture<GetProducerOffsetsResponse> getProducerOffsets(
             GetProducerOffsetsRequest request) {
         String producerId = request.getProducerId();
+        // Capture session before entering async block since currentSession() is thread-local
+        Session session = authorizer != null ? currentSession() : null;
 
         return CompletableFuture.supplyAsync(
                 () -> {
@@ -1091,7 +1104,8 @@ public final class CoordinatorService extends RpcServiceBase implements Coordina
                                     .removeIf(
                                             tableId -> {
                                                 try {
-                                                    authorizeTable(OperationType.READ, tableId);
+                                                    authorizeTableWithSession(
+                                                            session, OperationType.READ, tableId);
                                                     return false; // keep this table
                                                 } catch (Exception e) {
                                                     return true; // remove this table
@@ -1123,6 +1137,8 @@ public final class CoordinatorService extends RpcServiceBase implements Coordina
     @Override
     public CompletableFuture<DeleteProducerOffsetsResponse> deleteProducerOffsets(
             DeleteProducerOffsetsRequest request) {
+        // Capture session before entering async block since currentSession() is thread-local
+        Session session = authorizer != null ? currentSession() : null;
         return CompletableFuture.supplyAsync(
                 () -> {
                     try {
@@ -1139,7 +1155,7 @@ public final class CoordinatorService extends RpcServiceBase implements Coordina
                                             .collect(Collectors.toSet());
                             // Check WRITE permission for each table
                             for (Long tableId : tableIds) {
-                                authorizeTable(OperationType.WRITE, tableId);
+                                authorizeTableWithSession(session, OperationType.WRITE, tableId);
                             }
                         }
 


fluss-common/src/main/java/org/apache/fluss/config/ConfigOptions.java
```diff
@@ -354,6 +354,31 @@ public class ConfigOptions {
                                     + "The default value is 10. "
                                     + "This option is deprecated. Please use server.io-pool.size instead.");
 
+    /**
+     * The TTL (time-to-live) for producer offset snapshots. Snapshots older than this TTL will be
+     * automatically cleaned up by the coordinator server.
+     */
+    public static final ConfigOption<Duration> COORDINATOR_PRODUCER_SNAPSHOT_TTL =
+            key("coordinator.producer-snapshot.ttl")
```
wuchong: How about using `coordinator.producer-offsets.ttl` and `coordinator.producer-offsets.cleanup-interval`? These names align more closely with the established concepts of consumer/producer offsets.  

The term "snapshot" can be misleading, as it typically refers to data snapshots, such as KV snapshots or lake snapshots. While we currently store producer offsets in a global snapshot-like structure, the long-term design may allow each producer to persist its offsets independently. Using offset-centric naming now better reflects the semantic intent and keeps the configuration consistent with future extensibility.

fluss-client/src/main/java/org/apache/fluss/client/admin/FlussAdmin.java
```diff
@@ -588,6 +594,83 @@ public CompletableFuture<Void> cancelRebalance(@Nullable String rebalanceId) {
...
+                                long tableId = pbTableOffsets.getTableId();
+                                tableOffsets.put(
+                                        tableId,
+                                        ClientRpcMessageUtils.toTableBucketOffsets(pbTableOffsets));
+                            }
+
+                            long expirationTime =
+                                    response.hasExpirationTime() ? response.getExpirationTime() : 0;
+                            return new ProducerOffsetsResult(
+                                    response.getProducerId(), tableOffsets, expirationTime);
```
wuchong: nit: Move this conversion to `ClientRpcMessageUtils` to improve the readability of `FlussAdmin`.

fluss-client/src/main/java/org/apache/fluss/client/utils/ClientRpcMessageUtils.java
```diff
@@ -542,4 +542,50 @@ public static List<ConfigEntry> toConfigEntries(List<PbDescribeConfig> pbDescrib
...
                 .collect(Collectors.toList());
     }
+
+    /**
+     * Parses a PbTableOffsets into a map of TableBucket to offset.
+     *
+     * @param pbTableOffsets the protobuf table offsets
+     * @return map of TableBucket to offset
+     */
+    public static Map<TableBucket, Long> toTableBucketOffsets(
```
wuchong, line 557: not used, can be removed?

fluss-rpc/src/main/proto/FlussApi.proto
```diff
@@ -1062,4 +1062,53 @@ message PbBucketOffset {
...
   required int32 bucket_id = 2;
   optional int64 log_end_offset = 4;
+}
+
+// ------------------------------------------------------------------------------------------
+// Producer Offset Snapshot Management
+// ------------------------------------------------------------------------------------------
+
+// Register producer offset snapshot request and response
+message RegisterProducerOffsetsRequest {
```
wuchong: Move the `..Request` and `..Response` messages before Inner classes (L622).

fluss-rpc/src/main/proto/FlussApi.proto
```diff
@@ -1062,4 +1062,53 @@ message PbBucketOffset {
...
+}
+
+// Helper message for table offsets in producer snapshot (without table_path)
+message PbProducerTableOffsets {
+  required int64 table_id = 1;
+  repeated PbBucketOffset bucket_offsets = 2;
+}
+
+// Helper message for bucket offsets in producer snapshot
+message PbTableBucketOffset {
```
wuchong: Can we reuse the `PbProducerTableOffsets` in `RegisterProducerOffsetsRequest`? Or vice-versa, reuse `PbTableBucketOffset` in `GetProducerOffsetsResponse`. This can avoid introducing too many PB messages.

fluss-common/src/main/java/org/apache/fluss/rpc/messages/RegisterResult.java
```diff
@@ -0,0 +1,69 @@
...
+/**
+ * Result of producer offset registration.
+ *
+ * <p>This enum indicates whether a producer offset snapshot was newly created or already existed
+ * when calling registerProducerOffsets API.
+ *
+ * @since 0.9
+ */
+@PublicEvolving
+public enum RegisterResult {
```
wuchong, line 31: `RegisterResult` is not an RPC message that is (de)serialized through the netty RPC. I think a better package location for this should be `org.apache.fluss.client.admin` as it is used as a return type of `Admin` method.

fluss-common/src/test/java/org/apache/fluss/utils/RetryUtilsTest.java
```diff
@@ -0,0 +1,280 @@
...
+        // Should fail immediately without retries
+        assertThat(attempts.get()).isEqualTo(1);
+        // Verify interrupt status is preserved
+        assertThat(Thread.currentThread().isInterrupted()).isTrue();
+        // Clear interrupt status for other tests
+        Thread.interrupted();
+    }
+
+    @Test
+    void testInterruptedDuringBackoff() throws Exception {
```
wuchong: This method is unstable. It fails in my local env when repeats 20 times. I think we can remove this test as it is not a critical feature.

fluss-server/src/main/java/org/apache/fluss/server/coordinator/producer/ProducerSnapshotManager.java
```diff
@@ -0,0 +1,573 @@
...
+ *   <li>Periodic cleanup of expired snapshots
+ *   <li>Cleanup of orphan files in remote storage
+ * </ul>
+ *
+ * <p>The manager uses a background scheduler to periodically clean up expired snapshots and orphan
+ * files. The cleanup interval and snapshot TTL are configurable.
+ *
+ * @see ProducerSnapshotStore for low-level storage operations
+ */
+public class ProducerSnapshotManager implements AutoCloseable {
```
wuchong: renaming:

- `ProducerSnapshotManager` -> `ProducerOffsetsManager`
- `ProducerSnapshotStore` -> `ProducerOffsetsStore`
- `ProducerSnapshot` -> `ProducerOffsets`
- `ProducerSnapshotJsonSerde` -> `ProducerOffsetsJsonSerde`

fluss-server/src/main/java/org/apache/fluss/server/coordinator/producer/ProducerSnapshotStore.java
```diff
@@ -0,0 +1,384 @@
...
+    private FsPath writeOffsetsFile(String producerId, long tableId, Map<TableBucket, Long> offsets)
+            throws IOException {
+
+        String fileName = UUID.randomUUID() + ".offsets";
+        FsPath path =
+                new FsPath(
+                        new FsPath(
+                                new FsPath(new FsPath(remoteDataDir, "producers"), producerId),
+                                String.valueOf(tableId)),
+                        fileName);
```
wuchong: We maintain both local and remote paths in a centralized location, `FlussPaths`. This helps ensure consistency, avoids errors from manual path concatenation, and simplifies debugging.  

Could you add this path to `FlussPaths` by introducing a method like `remoteProducerOffsetsPath(remoteDataDir, producerId, tableId, UUID)`, along with Javadoc that clearly explains the structure and meaning of the resulting path string?

fluss-server/src/main/java/org/apache/fluss/server/coordinator/producer/ProducerSnapshotStore.java
```diff
@@ -0,0 +1,384 @@
...
+                INITIAL_BACKOFF_MS,
+                MAX_BACKOFF_MS);
+    }
+
+    private Map<TableBucket, Long> readOffsetsFile(FsPath path) throws IOException {
+        return RetryUtils.executeIOWithRetry(
+                () -> {
+                    FileSystem fs = path.getFileSystem();
+                    if (!fs.exists(path)) {
+                        throw new IOException("Offsets file not found: " + path);
```
wuchong: Throwing `IOException` will lead to retrying. However, I don't think we need retrying if the file not found. Should use a specific exception type to avoid retrying here?

fluss-server/src/main/java/org/apache/fluss/server/coordinator/producer/ProducerSnapshotManager.java
```diff
@@ -0,0 +1,573 @@
...
+        List<String> producerIds = snapshotStore.listProducerIds();
+        int cleanedCount = 0;
+        long currentTime = System.currentTimeMillis();
+
+        for (String producerId : producerIds) {
+            try {
+                Optional<ProducerSnapshot> optSnapshot =
+                        snapshotStore.getSnapshotMetadata(producerId);
+                if (isExpiredSnapshot(optSnapshot, currentTime)) {
+                    snapshotStore.deleteSnapshot(producerId);
```
wuchong: Should use `getSnapshotMetadataWithVersion` and `deleteSnapshotIfVersion` here? Otherwise, this may delete snapshot that updated just now.

fluss-server/src/main/java/org/apache/fluss/server/coordinator/CoordinatorService.java
```diff
@@ -983,4 +1008,127 @@ public FlussPrincipal getFlussPrincipal() {
...
+    // ==================================================================================
+    // Producer Offset Management APIs (for Exactly-Once Semantics)
+    // ==================================================================================
+
+    @Override
+    public CompletableFuture<RegisterProducerOffsetsResponse> registerProducerOffsets(
+            RegisterProducerOffsetsRequest request) {
+        // Authorization: require WRITE permission on cluster
+        if (authorizer != null) {
+            authorizer.authorize(currentSession(), OperationType.WRITE, Resource.cluster());
```
wuchong: We should check authroize for `WRITE` permission on the `tableId` in request.  Only cluster manager has the `WRITE` permission on `cluster`.

fluss-server/src/main/java/org/apache/fluss/server/coordinator/CoordinatorService.java
```diff
@@ -983,4 +1008,127 @@ public FlussPrincipal getFlussPrincipal() {
...
+                },
+                ioExecutor);
+    }
+
+    @Override
+    public CompletableFuture<GetProducerOffsetsResponse> getProducerOffsets(
+            GetProducerOffsetsRequest request) {
+        // Authorization: require READ permission on cluster
+        if (authorizer != null) {
+            authorizer.authorize(currentSession(), OperationType.READ, Resource.cluster());
```
wuchong: ditto. I think we should handle this like `listTables`.

fluss-server/src/main/java/org/apache/fluss/server/coordinator/CoordinatorService.java
```diff
@@ -983,4 +1008,127 @@ public FlussPrincipal getFlussPrincipal() {
...
+                },
+                ioExecutor);
+    }
+
+    @Override
+    public CompletableFuture<DeleteProducerOffsetsResponse> deleteProducerOffsets(
+            DeleteProducerOffsetsRequest request) {
+        // Authorization: require WRITE permission on cluster
+        if (authorizer != null) {
+            authorizer.authorize(currentSession(), OperationType.WRITE, Resource.cluster());
```
wuchong: ditto. We should check whether the user has WIRTE permission on all the tableIds under this producer id.

fluss-server/src/test/java/org/apache/fluss/server/coordinator/producer/ProducerSnapshotManagerTest.java
```diff
@@ -0,0 +1,357 @@
...
+    void testRegisterSnapshotSuccess() throws Exception {
+        String producerId = "test-manager-register";
+        Map<TableBucket, Long> offsets = createTestOffsets();
+
+        boolean created = manager.registerSnapshot(producerId, offsets, null);
+
+        assertThat(created).isTrue();
+        Optional<ProducerSnapshot> snapshot = manager.getSnapshotMetadata(producerId);
+        assertThat(snapshot).isPresent();
+        assertThat(snapshot.get().getExpirationTime()).isGreaterThan(System.currentTimeMillis());
```
wuchong: `isGreaterThanOrEqualTo`?

fluss-server/src/test/java/org/apache/fluss/server/coordinator/producer/ProducerSnapshotStoreTest.java
```diff
@@ -0,0 +1,337 @@
...
+        long expirationTime = System.currentTimeMillis() + 3600000;
+        store.tryStoreSnapshot(producerId, offsets, expirationTime);
+
+        // Read offsets back
+        Map<TableBucket, Long> retrieved = store.readOffsets(producerId);
+        assertThat(retrieved).hasSize(4);
+        assertThat(retrieved.get(new TableBucket(1L, 0))).isEqualTo(100L);
+        assertThat(retrieved.get(new TableBucket(1L, 1))).isEqualTo(200L);
+        assertThat(retrieved.get(new TableBucket(2L, 0))).isEqualTo(300L);
+        assertThat(retrieved.get(new TableBucket(3L, 100L, 0))).isEqualTo(400L);
```
wuchong: Can simply assert by `assertThat(retrieved).isEqualTo(offsets);`?

fluss-server/src/test/java/org/apache/fluss/server/coordinator/producer/ProducerSnapshotStoreTest.java
```diff
@@ -0,0 +1,337 @@
...
+
+        long expirationTime = System.currentTimeMillis() + 3600000;
+        store.tryStoreSnapshot(producerId, offsets, expirationTime);
+
+        Map<TableBucket, Long> retrieved = store.readOffsets(producerId);
+        assertThat(retrieved).hasSize(4);
+        assertThat(retrieved.get(new TableBucket(1L, 0))).isEqualTo(100L);
+        assertThat(retrieved.get(new TableBucket(2L, 10L, 0))).isEqualTo(200L);
+        assertThat(retrieved.get(new TableBucket(2L, 10L, 1))).isEqualTo(201L);
+        assertThat(retrieved.get(new TableBucket(2L, 20L, 0))).isEqualTo(300L);
```
wuchong: ditto

fluss-server/src/main/java/org/apache/fluss/server/coordinator/producer/ProducerSnapshotManager.java
```diff
@@ -0,0 +1,573 @@
...
+     * </ul>
+     *
+     * @param producerId the producer ID (typically Flink job ID)
+     * @param offsets map of TableBucket to offset for all tables
+     * @param ttlMs TTL in milliseconds for the snapshot, or null to use default
+     * @return true if a new snapshot was created (CREATED), false if snapshot already existed
+     *     (ALREADY_EXISTS)
+     * @throws Exception if the operation fails
+     */
+    public boolean registerSnapshot(String producerId, Map<TableBucket, Long> offsets, Long ttlMs)
```
wuchong, line 132: We should validate that `producerId` is a valid string, as it will be used as both a file name and a ZooKeeper node name. An invalid name could lead to serious issues such as file system corruption or ZK operation failures.  

We can reuse the existing utility `org.apache.fluss.metadata.TablePath#detectInvalidName` for this validation.

platinumhamburg:
Thanks @wuchong  for the detailed review. I’ve addressed all the comments above, please take another look when you have time.

fluss-server/src/main/java/org/apache/fluss/server/coordinator/CoordinatorService.java
```diff
@@ -983,4 +1010,140 @@ public FlussPrincipal getFlussPrincipal() {
...
+                                            .collect(Collectors.toSet());
+                            // Check WRITE permission for each table
+                            for (Long tableId : tableIds) {
+                                authorizeTable(OperationType.WRITE, tableId);
+                            }
+                        }
+
+                        producerOffsetsManager.deleteSnapshot(producerId);
+                        return new DeleteProducerOffsetsResponse();
+                    } catch (Exception e) {
```
wuchong, line 1167: We should re-throw all `ApiException`s as-is; otherwise, exceptions such as `AuthorizationException` or `UnknownTableOrBucketException` thrown by `authorizeTable` will not be properly recognized by the client as their original types. 

Please add permissions test in `ProducerOffsetsITCase` to verify the exception type and message for all producer offsets operations.

fluss-server/src/main/java/org/apache/fluss/server/coordinator/producer/ProducerOffsetsManager.java
```diff
@@ -0,0 +1,611 @@
...
+    /**
+     * Validates that a producer ID is valid for use as both a file name and ZK node name.
+     *
+     * @param producerId the producer ID to validate
+     * @throws IllegalArgumentException if producerId is invalid
+     */
+    private void validateProducerId(String producerId) {
+        String invalidReason = TablePath.detectInvalidName(producerId);
+        if (invalidReason != null) {
+            throw new IllegalArgumentException(
```
wuchong: We should use `ApiException` whenever an exception needs to be propagated to the client. Otherwise, it will be wrapped as an `UnknownServerException`, which we should avoid. In this case, maybe we can introduce a `InvalidProducerIdException`.

fluss-common/src/main/java/org/apache/fluss/rpc/messages/RegisterResult.java
wuchong: This class is duplicated with `org.apache.fluss.client.admin.RegisterResult`. We can remove this class, and just use `response.setResult(created ? 0 : 1)` in `CoordinatorService`.

fluss-common/src/test/java/org/apache/fluss/utils/RetryUtilsTest.java
wuchong: It seems this method is still not removed.

fluss-common/src/main/java/org/apache/fluss/rpc/messages/RegisterResult.java
platinumhamburg: Yes—this was my oversight, and I missed the duplicate-delete operation. However, I’m a bit concerned about the new pattern: the API layer defines strongly typed constants, but the lower layer aligns them using magic numbers. This introduces some code smell. Since I haven’t found a better or more elegant approach yet, I’ll update it following the suggested approach for now, and we can explore a better solution later.

platinumhamburg:
@wuchong Thanks for your detailed review and suggestions. I’ve addressed all the comments above—please take a look when you have time.

fluss-common/src/main/java/org/apache/fluss/rpc/messages/RegisterResult.java
wuchong: Yes. We could also consider in the future which package under `fluss-common` would be the most appropriate place for `RegisterResult`. My main concern initially was simply that it doesn't belong in the `*.rpc.message` package.

PR #2436, opened 2026-01-22
[KV] Add Configuration Option to Control RocksDB Statistics Collection

### LINKED ISSUE #2435: [KV] Add Configuration Option to Control RocksDB Statistics Collection
### Search before asking

### Motivation

RocksDB statistics collection has performance overhead. Users should be able to disable it when:
- RocksDB-level metrics are not needed
- Running in resource-constrained environments
- Troubleshooting performance issues

### Solution

**New Configuration**

| Option | Type | Default | Description |
|--------|------|---------|-------------|
| `kv.rocksdb.statistics.enabled` | Boolean | false | Enable RocksDB statistics collection for metrics |

**Files Modified**
- `ConfigOptions.java` - Add new config option
- `KvTablet.java` - Read config to control statistics initialization
- `RocksDBStatistics.java` - Minor optimization: use `getLongProperty()` instead of `getProperty()` to reduce JNI memory overhead
- `configuration.md` - Documentation

### Anything else?

Additional Optimization:

Replaced `RocksDB.getProperty()` with `getLongProperty()` in statistics collection to avoid unnecessary string allocation in JNI layer during high-frequency Prometheus scraping.

### Willingness to contribute

platinumhamburg:
- Use getLongProperty() instead of getProperty() to avoid native string allocation in JNI layer during high-frequency Prometheus metrics scraping
- Add kv.rocksdb.statistics.enabled config option (default: false) to allow disabling RocksDB statistics collection for reduced overhead
- Refactor getPropertyValue to getPropertyLongValue and eliminate code duplication between overloaded methods
- Add configuration documentation for the new option

### Purpose

Linked issue: close #2435 2435


fluss-common/src/main/java/org/apache/fluss/config/ConfigOptions.java
```diff
@@ -1636,6 +1590,19 @@ public class ConfigOptions {
                                     + "The RateLimiter is always enabled. The default value is Long.MAX_VALUE (effectively unlimited). "
                                     + "Set to a lower value (e.g., 100MB) to limit the rate.");
 
+    public static final ConfigOption<Boolean> KV_STATISTICS_ENABLED =
+            key("kv.rocksdb.statistics.enabled")
+                    .booleanType()
+                    .defaultValue(false)
```
Jackeyzhe, line 1596: This option defaults to `false`, while both its description and the user documentation say `true`. More importantly, `KvTablet` previously always enabled RocksDB statistics, so the new default silently disables ticker and histogram collection after an upgrade. Because `RocksDBStatistics` is still registered when its native `Statistics` is null, the corresponding Fluss gauges remain present but continuously report zero, which can silently break monitoring. Could we keep the default as `true` for backward compatibility and let users explicitly opt out? If `false` is intentional, the documentation and description must be aligned and the behavior change should be clearly documented and tested.

platinumhamburg: @Jackeyzhe Thank you for your thorough review and comments. This PR is likely no longer valid — it has drifted away from the context in which it was originally proposed. I'll go ahead and close it. Sorry for any time you wasted on this, but I really appreciate your review feedback.

PR #2456, opened 2026-01-23
[helm] Add `tabletServer.count` configuration option

### LINKED ISSUE #2455: [helm] Add tablet server count to helm charts
### Search before asking

### Description

The Helm chart should enable users set the number of tablet servers as parameter.

### Willingness to contribute

morazow:
### Purpose

Linked issue: close #2455 

### Brief change log

Enable setting the number tablet servers when deploying with helm charts.

### Documentation

Added parameter section in the helm deployment documentation.

### DIFF
--- a/helm/values.yaml
+++ b/helm/values.yaml
@@ -42,16 +42,19 @@ configurationOverrides:
   data.dir: /tmp/fluss/data
   internal.listener.name: INTERNAL
 
-persistence:
-  enabled: false
-  size: 1Gi
-  storageClass:
-
 tablet:
   numberOfReplicas: 3
+  storage:
+    enabled: false
+    size: 1Gi
+    storageClass:
 
 coordinator:
   numberOfReplicas: 1
+  storage:
+    enabled: false
+    size: 1Gi
+    storageClass:
 
 resources: {}
   # We usually recommend not to specify default resources and to leave this as a conscious
--- a/website/docs/install-deploy/deploying-with-helm.md
+++ b/website/docs/install-deploy/deploying-with-helm.md
@@ -141,10 +141,6 @@ The Fluss Helm chart deploys the following Kubernetes resources:
 - **ConfigMap**: Configuration management for `server.yaml` settings
 - **Services**: Headless services providing stable pod DNS names
 
-### Optional Components
-- **PersistentVolumes**: Data persistence when `persistence.enabled=true`
-
-
 ### Step 3: Verify Installation
 
 ```bash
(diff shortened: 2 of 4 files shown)


helm/values.yaml
```diff
@@ -47,6 +47,10 @@ persistence:
   size: 1Gi
   storageClass:
 
+# Tablet server only configuration options
+tabletServer:
```
affo: Just a suggestion, it looks to me that `persistence` is also a tablet server configuration.

Wouldn't it be better to have:

```yaml
tablet:
  numberOfReplicas:
  storage:
    enabled:
    size:
    class:
```

affo: @swuferhong what do you think?

swuferhong: what means 'persistence' for fluss?

morazow: Hello @swuferhong,

In the tablet and coordinator servers we have option to enabled and attach persistent volume (PV), so that the storage is not transient when they are recreated.

For example, for tablet: https://github.com/morazow/fluss/blob/main/helm/templates/sts-tablet.yaml#L99-L109

This could be enabled for data volumes.

I agree with @affo, we should maybe separate these configurations per type since sizes and storage classes could be different for coordinator and tablet.

affo:
This looks like a flakey test: https://github.com/apache/fluss/actions/runs/21281906620/job/61447896522?pr=2456

helm/values.yaml
affo: @swuferhong `persistence` is the current key in `values.yaml`:

```
persistence:
  enabled: false
  size: 1Gi
  storageClass:
```

And it is a top-level key.

What I am suggesting here is to group multiple tablet-relevant configs under `tablet`.
In any case we would need:

```
tablet:
  numberOfReplicas:

coordinator:
  numberOfReplicas: # for HA
```

At this point, we may extend `tablet` to include the current `persistence` values.
Does this make sense to you?

morazow: Updated with `numberOfReplicas` suggestion, let's wait for @swuferhong input for moving persistence to each tablet & coordinator sections 🤝

website/docs/install-deploy/deploying-with-helm.md
```diff
@@ -199,6 +199,12 @@ The following table lists the configurable parameters of the Fluss chart and the
 | `configurationOverrides.data.dir` | Local data directory | `/tmp/fluss/data` |
 | `configurationOverrides.internal.listener.name` | Internal listener name | `INTERNAL` |
 
+### Tablet Server Parameters
+
+| Parameter | Description | Default |
+|-----------|-------------|---------|
+| `tabletServer.count` | Number of TabletServer replicas to deploy | `3` |
```
affo: this is now not coherent with the above 👍 
Let's wait for @swuferhong for his opinion on moving persistence here as well

morazow: Thanks @affo, updated 👍

swuferhong (approved):
LGTM

PR #2464, opened 2026-01-24
[lake] Fix lake encoder causing prefix lookup to miss data when bucket key is a subset of primary key

### LINKED ISSUE #1412: Prefix Lookup Fails in Paimon Lake Format: Key Encoding Mismatch in Fixed-Length Region for Fluss Primary Key Tables
### Search before asking

### Fluss version

0.7.0 (latest release)

### Please describe the bug 🐞

When using the Paimon lake format for Fluss primary key tables, keys are encoded using the Paimon format, which stores length and offset information for variable-length fields. While primary key point queries work correctly, prefix queries fail to return any data. This issue arises because the encoded fixed-length region of the stored key differs from that of the prefix query key.

**Root Cause:**
The PaimonBinaryRowWriter class (see com.alibaba.fluss.row.encode.paimon.PaimonBinaryRowWriter) encodes variable-length fields differently for primary key look up queries and prefix look up queries:

1. Primary Key Point Query:Encodes all key fields (e.g., two string fields)
2. Encodes only the prefix field (e.g., the first string field)

### Solution

Do we need to use Paimon's encoding for storing keys in primary key tables? Is it possible to use Fluss's default KeyEncoder instead?

luoyuxia:
### Purpose

Linked issue: close #1412 

Fix lake encoder causing prefix lookup to miss data when bucket key is a subset of primary key.

When a primary key table has a non-default bucket key (bucket key is a subset of primary key), the previous implementation used lake's encoder (e.g., Paimon) to encode the primary key. However, lake encoders like Paimon store strings longer than 7 characters in a variable-length area, which breaks prefix lookup because the encoded bucket key bytes are no longer a prefix of the encoded primary key bytes.

### Brief change log

This PR:
1. Introduces kv format version attribute for tables to distinguish new tables from legacy ones.
2. Refactors KeyEncoder with separate factory methods:
    - ofPrimaryKeyEncoder: For encoding primary keys. New tables (kv format version 2) with non-default bucket keys use CompactedKeyEncoder to ensure bucket key bytes are always a prefix of primary key bytes, enabling prefix lookup.
    - ofBucketKeyEncoder: For encoding bucket keys used in bucket calculation. Always uses lake's encoder when lake format is present.
3. Maintains backward compatibility: Legacy tables (without kv format version) continue using the original lake encoder to ensure data compatibility.
4. Rejects old clients for new format tables: When old client (API version 0) tries to access new format tables (kv format version 2) with non-default bucket keys, the server returns UNSUPPORTED_VERSION error to prevent data corruption.
5. Performance optimization: When bucket key equals primary key (isDefaultBucketKey = true), prefix lookup is not needed, so we can use lake's encoder for both. This allows reusing the encoded primary key bytes for bucket calculation, saving one encoding operation.

### Tests
- Added testLegacyPKTableUsePaimonEncoder in FlussLakeTableITCase to verify legacy tables continue using Paimon encoder.
- Added testKvFormatV2TableRejectV0Client in ReplicaManagerTest to verify old clients are rejected for new format tables.
- Added testUnionReadNotDefaultBucketKeyTable in FlinkUnionReadPrimaryKeyTableITCase to verify union read works correctly with new encoding.
- Updated existing prefix lookup tests to use longer strings (length > 7) to properly catch encoding issues.

Also verify the back compatibility:
- use client v0.8 to write new table, expected `UnsupportedVersionException` is thrown
- use main branch client to write old table, still write will old encode strategy. Both main branch client and v0.8 client can get the key

### API and Format
Added new config option TABLE_KV_FORMAT_VERSION to track table's kv encoding format version.

### DIFF
--- a/fluss-rpc/src/main/java/org/apache/fluss/rpc/protocol/ApiKeys.java
+++ b/fluss-rpc/src/main/java/org/apache/fluss/rpc/protocol/ApiKeys.java
@@ -44,14 +44,17 @@ public enum ApiKeys {
     UPDATE_METADATA(1013, 0, 0, PRIVATE),
     PRODUCE_LOG(1014, 0, 0, PUBLIC),
     FETCH_LOG(1015, 0, 0, PUBLIC),
+
     // Version 0: Uses lake's encoder for primary key encoding (legacy behavior).
     // Version 1: Uses CompactedKeyEncoder for primary key encoding when bucket key differs from
     //            primary key, enabling prefix lookup support.
     PUT_KV(1016, 0, 1, PUBLIC),
+
     // Version 0: Uses lake's encoder for primary key encoding (legacy behavior).
     // Version 1: Uses CompactedKeyEncoder for primary key encoding when bucket key differs from
     //            primary key, enabling prefix lookup support.
     LOOKUP(1017, 0, 1, PUBLIC),
+
     NOTIFY_LEADER_AND_ISR(1018, 0, 0, PRIVATE),
     STOP_REPLICA(1019, 0, 0, PRIVATE),
     ADJUST_ISR(1020, 0, 0, PRIVATE),
@@ -68,10 +71,12 @@ public enum ApiKeys {
     NOTIFY_LAKE_TABLE_OFFSET(1031, 0, 0, PRIVATE),
     GET_LATEST_LAKE_SNAPSHOT(1032, 0, 0, PUBLIC),
     LIMIT_SCAN(1033, 0, 0, PUBLIC),
+
     // Version 0: Uses lake's encoder for prefix key encoding (legacy behavior).
     // Version 1: Uses CompactedKeyEncoder for prefix key encoding when bucket key differs from
     //            primary key, ensuring encoded bucket key bytes are a prefix of primary key bytes.
     PREFIX_LOOKUP(1034, 0, 1, PUBLIC),
+
     GET_DATABASE_INFO(1035, 0, 0, PUBLIC),
     CREATE_PARTITION(1036, 0, 0, PUBLIC),
     DROP_PARTITION(1037, 0, 0, PUBLIC),
--- a/fluss-server/src/main/java/org/apache/fluss/server/coordinator/CoordinatorService.java
+++ b/fluss-server/src/main/java/org/apache/fluss/server/coordinator/CoordinatorService.java
@@ -28,6 +28,7 @@ import org.apache.fluss.config.cluster.AlterConfig;
 import org.apache.fluss.config.cluster.AlterConfigOpType;
 import org.apache.fluss.exception.ApiException;
 import org.apache.fluss.exception.InvalidAlterTableException;
+import org.apache.fluss.exception.InvalidConfigException;
 import org.apache.fluss.exception.InvalidCoordinatorException;
 import org.apache.fluss.exception.InvalidDatabaseException;
 import org.apache.fluss.exception.InvalidTableException;
@@ -548,7 +549,7 @@ public final class CoordinatorService extends RpcServiceBase implements Coordina
                 newDescriptor = newDescriptor.withProperties(newProperties);
             } else {
                 if (formatVersion > CURRENT_KV_FORMAT_VERSION) {
-                    throw new IllegalArgumentException(
+                    throw new InvalidConfigException(
                             String.format(
                                     "Unsupported kv format version %d. "
                                             + "The maximum supported version is %d.",
--- a/website/docs/maintenance/operations/upgrade-notes-0.8.md
+++ b/website/docs/maintenance/operations/upgrade-notes-0.8.md
@@ -7,6 +7,8 @@ sidebar_class_name: hidden
 
 These upgrade notes discuss important aspects, such as configuration, behavior, or dependencies, that changed between Fluss 0.7 and Fluss 0.8. Please read these notes carefully if you are planning to upgrade your Fluss version to 0.8.
 
+Previous versions can be found in the [archive of upgrade notes](upgrade-notes-archive.md).
+
 ## Deprecation / End of Support
 
 ### Java 8 is Deprecated
--- a/website/docs/maintenance/operations/upgrade-notes-0.9.md
+++ b/website/docs/maintenance/operations/upgrade-notes-0.9.md
@@ -7,6 +7,8 @@ sidebar_position: 4
 
 These upgrade notes discuss important aspects, such as configuration, behavior, or dependencies, that changed between Fluss 0.8 and Fluss 0.9. Please read these notes carefully if you are planning to upgrade your Fluss version to 0.9.
 
+Previous versions can be found in the [archive of upgrade notes](upgrade-notes-archive.md).
+
 ## Adding Columns
 
 Fluss storage format was designed with schema evolution protocols from day one. Therefore, tables created in v0.8 or earlier can immediately benefit from the `ADD COLUMN` feature after upgrading to v0.9 without dropping and re-creating table.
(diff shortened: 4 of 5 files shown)


luoyuxia:
@wuchong Could you please help review when you got some time?

fluss-client/src/test/java/org/apache/fluss/client/table/FlussLakeTableITCase.java
```diff
@@ -192,23 +195,26 @@ void testPrimaryKeyTable(boolean isPartitioned, boolean isDefaultBucketKey) thro
...
+            // init lookup columns
+            List<String> lookUpColumns;
+            Lookuper lookuper;
+            if (isDefaultBucketKey) {
+                lookUpColumns = Arrays.asList("a", "c", "d");
+                lookuper = table.newLookup().createLookuper();
+            } else {
+                lookUpColumns =
+                        isPartitioned ? Arrays.asList("a", "d") : Collections.singletonList("a");
+                lookuper = table.newLookup().lookupBy(lookUpColumns).createLookuper();
```
wuchong, line 207: Currently, in the new version, the `lookupBy` only supports prefix key, but this is not align with the method name.

We should check `lookupColumnNames` is a valid prefix key, then create `PrefixKeyLookup`; if it is a primary key, then create `PrimaryKeyLookup`; otherwise, should throw a exception to let user know the lookup columns are not valid. 

After this change, we don't need this change in the test and thus, provides backward compatiblity.

fluss-common/src/main/java/org/apache/fluss/config/ConfigOptions.java
```diff
@@ -1283,6 +1286,21 @@ public class ConfigOptions {
...
+     * datalake encoding for bucket key (to align with datalake bucket).
+     */
+    public static final ConfigOption<Integer> TABLE_KV_FORMAT_VERSION =
+            key("table.kv.format.version")
+                    .intType()
+                    .noDefaultValue()
+                    .withDescription(
+                            "The version of the kv format for backward compatibility. "
+                                    + "When absent (old tables), primary key encoding may differ from new tables. "
+                                    + "Version 2 indicates new encoding strategy for datalake tables.");
```
wuchong: Add detailed explanation for the config option, explain behavior for different format version. And please update documentation.

fluss-common/src/main/java/org/apache/fluss/row/encode/KeyEncoder.java
```diff
@@ -26,19 +27,87 @@
...
+        }
+        int version = kvFormatVersion.get();
+        if (version == 2) {
+            if (isDefaultBucketKey) {
+                return of(rowType, keyFields, dataLakeFormat);
+            } else {
+                // use CompactedKeyEncoder to support prefix look up
+                return CompactedKeyEncoder.createKeyEncoder(rowType, keyFields);
+            }
+        }
```
wuchong, line 89: We should also fallback to use legacy behavior if format version is `1`, as your config option explains.

fluss-common/src/main/java/org/apache/fluss/config/ConfigOptions.java
```diff
@@ -1283,6 +1286,21 @@ public class ConfigOptions {
...
                                     + "The supported formats are `compacted` and `indexed`.");
 
+    /**
+     * The version of the kv format. This is used for backward compatibility when encoding strategy
+     * changes. Version 1 (absent): Old tables use datalake encoding for both primary key and bucket
+     * key. Version 2: New tables use Fluss encoding for primary key (to support prefix lookup) and
+     * datalake encoding for bucket key (to align with datalake bucket).
+     */
+    public static final ConfigOption<Integer> TABLE_KV_FORMAT_VERSION =
+            key("table.kv.format.version")
```
wuchong: I suggest using `table.kv.format-version` instead of `table.kv.format.version`.

Since `table.kv.format` is already defined as a string value (e.g, `compacted`), using a dot-separator would create a namespace conflict in the config hierarchy. Using a hyphen keeps the parameters flat and ensures backward compatibility with the existing parser.

Besides, Iceberg also uses `format-version` for the version configuration.

fluss-rpc/src/main/java/org/apache/fluss/rpc/protocol/ApiKeys.java
```diff
@@ -44,8 +44,8 @@ public enum ApiKeys {
     UPDATE_METADATA(1013, 0, 0, PRIVATE),
     PRODUCE_LOG(1014, 0, 0, PUBLIC),
     FETCH_LOG(1015, 0, 0, PUBLIC),
-    PUT_KV(1016, 0, 0, PUBLIC),
-    LOOKUP(1017, 0, 0, PUBLIC),
+    PUT_KV(1016, 0, 1, PUBLIC),
+    LOOKUP(1017, 0, 1, PUBLIC),
```
wuchong, line 56: Add comments to explain the version bump and version behaviors.

fluss-server/src/main/java/org/apache/fluss/server/coordinator/CoordinatorService.java
```diff
@@ -492,6 +494,27 @@ private TableDescriptor applySystemDefaults(
...
+        if (newDescriptor.hasPrimaryKey()) {
+            Map<String, String> newProperties = new HashMap<>(newDescriptor.getProperties());
+            String formatVersion = newProperties.get(ConfigOptions.TABLE_KV_FORMAT_VERSION.key());
+            if (formatVersion != null) {
+                throw new IllegalArgumentException(
+                        "Manual configuration of '"
+                                + ConfigOptions.TABLE_KV_FORMAT_VERSION.key()
+                                + "' is not supported. The coordinator chooses the appropriate "
+                                + "KV format version automatically. Remove this property from the "
+                                + "table properties (current value: '"
```
wuchong: Do we really need to be so strict? I think the format version should be user-configurable, similar to how Apache Iceberg format version handles it.  

Right now, we have a critical bug in version 1 that justifies enforcing a specific version, but in the future, version evolution may primarily introduce advanced features rather than fix correctness issues. In such cases, users shouldn’t be forced to adopt a particular version; they should be free to choose based on their needs and risk tolerance.

Relaxing this registration would also help us construct use cases with older format versions, which is beneficial for building compatibility test cases.

fluss-server/src/main/java/org/apache/fluss/server/replica/ReplicaManager.java
```diff
@@ -1746,6 +1768,31 @@ public void shutdown() throws InterruptedException {
...
+                    && kvFormatVersion.get() == KV_FORMAT_VERSION_2
+                    && !tableInfo.isDefaultBucketKey()) {
+                throw new UnsupportedVersionException(
+                        String.format(
+                                "Client API version %d is not supported for table '%s'. "
+                                        + "This table uses new key encoding strategy (kv format version %d). "
+                                        + "Please upgrade your Fluss client to a newer version.",
+                                apiVersion,
+                                tableInfo.getTablePath(),
+                                tableInfo.getTableConfig().getKvFormatVersion().get()));
```
wuchong:
```suggestion
                                kvFormatVersion.get()));
```

fluss-common/src/main/java/org/apache/fluss/row/encode/KeyEncoder.java
```diff
@@ -26,19 +27,87 @@
...
+    }
+
+    /**
+     * Creates a key encoder based on the datalake format.
+     *
+     * @param rowType the row type of the input row
+     * @param keyFields the key fields to encode
+     * @param lakeFormat the datalake format, null means no datalake
      */
     static KeyEncoder of(
```
wuchong, line 120: nit: This is not suggested to be used externally, however, private static method is supported since java9, and we still provides java8 compatibility, we can't use private static here, but could you add an deprecate annotation and comment on this method?

fluss-client/src/test/java/org/apache/fluss/client/table/FlussTableITCase.java
```diff
@@ -305,10 +310,16 @@ void testPutAndPrefixLookup() throws Exception {
...
+
+        // We use strings with length > 7 (e.g., "valuevalue1") to properly test prefix lookup.
+        // Previously, short strings like "value1" hid a bug: Paimon's encoder stores strings
+        // longer than 7 characters in a variable-length area, which breaks prefix lookup since
+        // the encoded bucket key bytes are no longer a prefix of the encoded primary key bytes.
+        // Using longer strings ensures we catch such issues.
+        verifyPutAndLookup(table, new Object[] {1, "a", 1L, "valuevalue1"});
+        verifyPutAndLookup(table, new Object[] {1, "a", 2L, "valuevalue2"});
+        verifyPutAndLookup(table, new Object[] {1, "a", 3L, "valuevalue3"});
+        verifyPutAndLookup(table, new Object[] {2, "a", 4L, "valuevalue4"});
```
wuchong: I tried to run this test in `main` branch, and I can't reproduce the problem.

I think this can't reproduce the problem, because `c` is not a part of prefix key, so the encoding bytes of prefix key is still the same and can fetch result. In order to reproduce the problem, we should make the value length of column `b` larger than `7`.

wuchong (commented):
Thanks @luoyuxia , I left some comments. 

Besides, please add upgrade note for this behavior change and how to address this problem with old format tables.

fluss-client/src/test/java/org/apache/fluss/client/table/FlussTableITCase.java
luoyuxia: Thanks for pointing that.

fluss-server/src/main/java/org/apache/fluss/server/coordinator/CoordinatorService.java
luoyuxia: I agree on the future evolution point

fluss-client/src/test/java/org/apache/fluss/client/table/FlussLakeTableITCase.java
luoyuxia: Agreed. But I found it will cost some effort to do that since it also require to handle the lookUpColumns is with different order which require projection and more tests to cover. I create #2521 to track it and will do it in #2521

wuchong: Thank you @luoyuxia , I think we can skip the support of out-of-order lookup columns, and just support primary key columns in the issue, and support fail-fast for non-primary-key and non-prefix-key cases.

wuchong (approved):
LGTM. I pushed a commit to improve the docs a bit.

luoyuxia:
@wuchong Thanks for the improvement for docs

Issue #2334, opened 2026-01-08
[flink] Basic $changelog read support without pushdown optimizations for log table


MehulBatra:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/fluss/issues) and found nothing similar.

### Description

introducing virtual table $changelog for log table without pushdown optimization

### Willingness to contribute

- [x] I'm willing to submit a PR!

Issue #2466, opened 2026-01-24
[test] testPkCompactedPollFromLatestNoRecords fails with Unknown table or bucket


SML0127:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/fluss/issues) and found nothing similar.

### Fluss version

0.8.0 (latest release)

### Please describe the bug 🐞

The test calls `subscribeFromLatestOffset` immediately before tablet server recognize table/bucket. Resolved with `waitUntilTableRead()` 

- Error Messages
```
Error:  org.apache.fluss.client.table.FlussTableITCase.testPkCompactedPollFromLatestNoRecords  Time elapsed: 2.034 s  <<< ERROR!
java.util.concurrent.ExecutionException: org.apache.fluss.exception.UnknownTableOrBucketException: Unknown table or bucket: TableBucket{tableId=41, bucket=0}
	at java.base/java.util.concurrent.CompletableFuture.reportGet(CompletableFuture.java:395)
	at java.base/java.util.concurrent.CompletableFuture.get(CompletableFuture.java:2005)
	at org.apache.fluss.client.admin.ClientToServerITCaseBase.listOffsets(ClientToServerITCaseBase.java:180)
	at org.apache.fluss.client.admin.ClientToServerITCaseBase.subscribeFromLatestOffset(ClientToServerITCaseBase.java:195)
	at org.apache.fluss.client.table.FlussTableITCase.testPkCompactedPollFromLatestNoRecords(FlussTableITCase.java:1755)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:566)
	at org.junit.platform.commons.util.ReflectionUtils.invokeMethod(ReflectionUtils.java:727)
	at org.junit.jupiter.engine.execution.MethodInvocation.proceed(MethodInvocation.java:60)
	at org.junit.jupiter.engine.execution.InvocationInterceptorChain$ValidatingInvocation.proceed(InvocationInterceptorChain.java:131)
	at org.junit.jupiter.engine.extension.TimeoutExtension.intercept(TimeoutExtension.java:156)
[...]

PR #2467, opened 2026-01-24
[flink] changelog read support for log table without pushdown optimizations

MehulBatra:
### Purpose

Linked issue: close https://github.com/apache/fluss/issues/2334

### Brief change log
$changelog Support for Log Tables
Change Type: Log tables use +A (Append)

### Modified Files
fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/catalog/FlinkCatalog.java

- Removed PK-only validation in getVirtualChangelogTable()

### Tests:

fluss-flink/fluss-flink-common/src/test/java/org/apache/fluss/flink/source/ChangelogVirtualTableITCase.java

- Renamed testChangelogVirtualTableWithNonPrimaryKeyTable → testChangelogVirtualTableWithLogTable
Now tests +A change type for log tables
- testChangelogWithScanStartupMode ->  Fixed race condition by writing 3 records with 1s delays, ensuring at   
At least one is captured after the subscription is ready in latest mode

fluss-flink/fluss-flink-common/src/test/java/org/apache/fluss/flink/catalog/FlinkCatalogITCase.java

- Updated testGetChangelogVirtualTable() to verify log table schema instead of expecting error

### DIFF
--- a/fluss-flink/fluss-flink-common/src/test/java/org/apache/fluss/flink/catalog/FlinkCatalogITCase.java
+++ b/fluss-flink/fluss-flink-common/src/test/java/org/apache/fluss/flink/catalog/FlinkCatalogITCase.java
@@ -920,18 +920,25 @@ abstract class FlinkCatalogITCase {
         // Verify options are inherited from base table
         assertThat(changelogTable.getOptions()).containsEntry("bucket.num", "1");
 
-        // Verify $changelog on non-PK table throws appropriate error
+        // Verify $changelog log tables (append-only with +A change type)
         tEnv.executeSql("CREATE TABLE log_table_for_changelog (id INT, name STRING)");
 
-        assertThatThrownBy(
-                        () ->
-                                catalog.getTable(
-                                        new ObjectPath(
-                                                DEFAULT_DB, "log_table_for_changelog$changelog")))
-                .isInstanceOf(CatalogException.class)
-                .hasRootCauseMessage(
-                        "Virtual $changelog tables are only supported for primary key tables. "
-                                + "Table fluss.log_table_for_changelog does not have a primary key.");
+        CatalogTable logChangelogTable =
+                (CatalogTable)
+                        catalog.getTable(
+                                new ObjectPath(DEFAULT_DB, "log_table_for_changelog$changelog"));
+
+        // Log table changelog should have same metadata columns
+        Schema expectedLogSchema =
+                Schema.newBuilder()
+                        .column("_change_type", DataTypes.STRING().notNull())
+                        .column("_log_offset", DataTypes.BIGINT().notNull())
+                        .column("_commit_timestamp", DataTypes.TIMESTAMP_LTZ(3).notNull())
+                        .column("id", DataTypes.INT())
+                        .column("name", DataTypes.STRING())
+                        .build();
+
+        assertThat(logChangelogTable.getUnresolvedSchema()).isEqualTo(expectedLogSchema);
     }
 
     /**


wuchong:
@MehulBatra could you please fix `FlinkCatalogITCase#testGetChangelogVirtualTable`? 

```
Error:  org.apache.fluss.flink.catalog.Flink120CatalogITCase.testGetChangelogVirtualTable  Time elapsed: 2.058 s  <<< FAILURE!
java.lang.AssertionError: 

Expecting code to raise a throwable.
```

Besides, the unstable `FlussTableITCase.testPkCompactedPollFromLatestNoRecords` has been (maybe) fixed by https://github.com/apache/fluss/issues/2466.

fluss-flink/fluss-flink-common/src/test/java/org/apache/fluss/flink/source/ChangelogVirtualTableITCase.java
```diff
@@ -355,12 +370,19 @@ public void testChangelogWithScanStartupMode() throws Exception {
...
+            Thread.sleep(1000);
+            CLOCK.advanceTime(Duration.ofMillis(100));
+            writeRows(conn, tablePath, Arrays.asList(row(i, "v" + i)), false);
+        }
+
+        // Collect results - we expect at least 1 record (the ones written after subscription)
         List<String> latestResults = collectRowsWithTimeout(rowIterLatest, 1, true);
         assertThat(latestResults).hasSize(1);
-        assertThat(latestResults.get(0)).isEqualTo("+I[+I, 6, v6]");
+        // The record should be one of the records we wrote (v6 to v10)
```
wuchong: Personally, I’m not in favor of adding 3 × 4 seconds (3 more seconds for each test and 4 flink versions matrix) of sleep to tests just to stabilize flaky ones. Simply inserting sleeps doesn’t actually solve the underlying issue—the enumerator might still be delayed due to CPU load during testing.  

The root problem is that we currently lack a clear signal indicating that a `SELECT` query has successfully started. Given this, I suggest removing this test for now, especially since we also don’t cover the `latest` mode in our `scan.startup.mode` configuration tests.  

Alternatively, if we still want to cover this case, we can do like this:

1. Use `collectRowsWithTimeout` to fetch one record with a 5-second timeout.  
2. If no record is received within the timeout, write a new record (with id larger than 5) and repeat step 1 for 10 times.  
3. Once a record is fetched, assert that it matches id larger than 5. 

This strategy avoids excessive waiting in most cases while providing a clearer success condition.

MehulBatra: Sure, I agree with you and like the idea.

wuchong: I removed the latest scan startup test temporarily in https://github.com/apache/fluss/pull/2470, becuase it is quite unstable. We can re-introduce it in this PR with the proposed way above.

MehulBatra:
> @MehulBatra could you please fix `FlinkCatalogITCase#testGetChangelogVirtualTable`?
> 
> ```
> Error:  org.apache.fluss.flink.catalog.Flink120CatalogITCase.testGetChangelogVirtualTable  Time elapsed: 2.058 s  <<< FAILURE!
> java.lang.AssertionError: 
> 
> Expecting code to raise a throwable.
> ```
> 
> Besides, the unstable `FlussTableITCase.testPkCompactedPollFromLatestNoRecords` has been (maybe) fixed by #2466.

I missed it in my latest PR, it was there in the closed one. I will fix it.

fluss-flink/fluss-flink-common/src/test/java/org/apache/fluss/flink/source/ChangelogVirtualTableITCase.java
MehulBatra: ok

MehulBatra: I would create a separate issue to introduce that, I hope it's fine

MehulBatra: #2471

MehulBatra:
Thanks for the review and quick merge @wuchong

PR #2481, opened 2026-01-26
Additional test assertions to ensure parity between Rust client and Java

leekeiabstraction:
### Purpose

We are introducing partitioning utils on Rust client side here: https://github.com/apache/fluss-rust/pull/214

In order to ensure correct partitioning behaviour on Rust side, we add additional assertions in test cases for TimestampLtz/Ntz which Rust client side also mirrors.


leekeiabstraction:
@luoyuxia @fresh-borzoni Appreciate your reviews here

PR #2485, opened 2026-01-26
[kv] Implement insertIfNotExists on tablet server

### LINKED ISSUE #2482: Implement Atomic Lookup-or-Put for KV Tables
### Search before asking

### Motivation

Current KV PUT operations are "blind" overwrites. Streaming scenarios often require "Insert If Not Exists" logic. Performing separate LOOKUP and PUT calls on the client side introduces race conditions, doubles network latency, and complicates auto-increment column handling.
#1886 

### Anything else?

_No response_

### Willingness to contribute

beryllw:
### Purpose

Linked issue: close #2482

### DIFF
--- a/fluss-server/src/main/java/org/apache/fluss/server/kv/partialupdate/PartialUpdater.java
+++ b/fluss-server/src/main/java/org/apache/fluss/server/kv/partialupdate/PartialUpdater.java
@@ -41,6 +41,7 @@ public class PartialUpdater {
 
     private final BitSet partialUpdateCols = new BitSet();
     private final BitSet primaryKeyCols = new BitSet();
+    private final boolean updatePrimaryKeyOnly;
     private final DataType[] fieldDataTypes;
 
     public PartialUpdater(KvFormat kvFormat, short schemaId, Schema schema, int[] targetColumns) {
@@ -60,6 +61,7 @@ public class PartialUpdater {
             flussFieldGetters[i] = InternalRow.createFieldGetter(fieldDataTypes[i], i);
         }
         this.rowEncoder = RowEncoder.create(kvFormat, fieldDataTypes);
+        this.updatePrimaryKeyOnly = partialUpdateCols.equals(primaryKeyCols);
     }
 
     private void sanityCheck(Schema schema, int[] targetColumns) {
@@ -100,6 +102,11 @@ public class PartialUpdater {
      * @return the updated value (schema id + row bytes)
      */
     public BinaryValue updateRow(@Nullable BinaryValue oldValue, BinaryValue partialValue) {
+        if (updatePrimaryKeyOnly && oldValue != null) {
+            // only primary key columns are updated, return the old value directly
+            return oldValue;
+        }
+
         rowEncoder.startNewRow();
         // write each field
         for (int i = 0; i < fieldDataTypes.length; i++) {
--- a/fluss-server/src/main/java/org/apache/fluss/server/replica/ReplicaManager.java
+++ b/fluss-server/src/main/java/org/apache/fluss/server/replica/ReplicaManager.java
@@ -142,6 +142,7 @@ import static org.apache.fluss.config.ConfigOptions.KV_FORMAT_VERSION_2;
 import static org.apache.fluss.server.TabletManagerBase.getTableInfo;
 import static org.apache.fluss.utils.FileUtils.isDirectoryEmpty;
 import static org.apache.fluss.utils.Preconditions.checkArgument;
+import static org.apache.fluss.utils.Preconditions.checkNotNull;
 import static org.apache.fluss.utils.Preconditions.checkState;
 import static org.apache.fluss.utils.concurrent.LockUtils.inLock;
 
@@ -597,13 +598,14 @@ public class ReplicaManager {
 
     /**
      * Collect missing keys from lookup results for insertion, populating both context and batch
-     * maps.
+     * maps. And returns the schema of the table, may return null if there is no missing keys.
      */
-    private void collectMissingKeysForInsert(
+    private Schema collectMissingKeysForInsert(
             Map<TableBucket, List<byte[]>> entriesPerBucket,
             Map<TableBucket, LookupResultForBucket> lookupResults,
             Map<TableBucket, MissingKeysContext> missingKeysContextMap,
             Map<TableBucket, KvRecordBatch> kvRecordBatchMap) {
+        Schema schema = null;
         for (Map.Entry<TableBucket, List<byte[]>> entry : entriesPerBucket.entrySet()) {
             TableBucket tb = entry.getKey();
             LookupResultForBucket lookupResult = lookupResults.get(tb);
@@ -622,13 +624,14 @@ public class ReplicaManager {
                 }
             }
             if (!missingKeys.isEmpty()) {
+                TableInfo tableInfo = getReplicaOrException(tb).getTableInfo();
                 missingKeysContextMap.put(tb, new MissingKeysContext(missingIndexes, missingKeys));
-                kvRecordBatchMap.put(
-                        tb,
-                        KeyRecordBatch.create(
-                                missingKeys, getReplicaOrException(tb).getTableInfo()));
+                kvRecordBatchMap.put(tb, KeyRecordBatch.create(missingKeys, tableInfo));
+                schema = tableInfo.getSchema();
             }
         }
+        // this assumes all table buckets belong to the same table
+        return schema;
     }
 
     /** Lookup a single key value. */
@@ -706,19 +709,18 @@ public class ReplicaManager {
             checkArgument(
                     timeoutMs != null && requiredAcks != null,
                     "timeoutMs and requiredAcks must be set");
+
             Map<TableBucket, MissingKeysContext> entriesPerBucketToInsert = new HashMap<>();
             Map<TableBucket, KvRecordBatch> produceEntryData = new HashMap<>();
-            collectMissingKeysForInsert(
-                    entriesPerBucket,
-                    lookupResultForBucketMap,
-                    entriesPerBucketToInsert,
-                    produceEntryData);
+            Schema schema =
+                    collectMissingKeysForInsert(
+                            entriesPerBucket,
+                            lookupResultForBucketMap,
+                            entriesPerBucketToInsert,
+                            produceEntryData);
 
             if (!produceEntryData.isEmpty()) {
-                // Compute target columns: exclude auto-increment to prevent overwriting
-                TableBucket firstBucket = produceEntryData.keySet().iterator().next();
-                Schema schema = getReplicaOrException(firstBucket).getTableInfo().getSchema();
-
+                checkNotNull(schema, "Schema must be available for insert-if-not-exists");
                 // TODO: Performance optimization: during lookup-with-insert-if-not-exists flow,
                 // the original key bytes are wrapped in KeyRecordBatch, then during putRecordsToKv
                 // they are decoded to rows and immediately re-encoded back to key bytes, causing
--- a/fluss-server/src/test/java/org/apache/fluss/server/kv/KvTabletTest.java
+++ b/fluss-server/src/test/java/org/apache/fluss/server/kv/KvTabletTest.java
@@ -290,17 +290,11 @@ class KvTabletTest {
                                 0,
                                 Arrays.asList(
                                         // -- for batch 1
-                                        ChangeType.INSERT,
-                                        ChangeType.INSERT,
-                                        ChangeType.UPDATE_BEFORE,
-                                        ChangeType.UPDATE_AFTER),
+                                        ChangeType.INSERT, ChangeType.INSERT),
                                 Arrays.asList(
                                         // for k1
                                         new Object[] {1, null, null},
-                                        // for k2: +I
-                                        new Object[] {2, null, null},
-                                        // for k2: -U, +U
-                                        new Object[] {2, null, null},
+                                        // for k2: +I, the second K2 update is skipped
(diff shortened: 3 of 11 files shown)


fluss-common/src/main/java/org/apache/fluss/row/decode/KeyDecoder.java
```diff
@@ -0,0 +1,45 @@
...
+
+import org.apache.fluss.row.InternalRow;
+import org.apache.fluss.types.RowType;
+
+import java.util.List;
+
+/**
+ * Interface for decoding key bytes back to {@link InternalRow}.
+ *
+ * <p>This interface provides functionality to decode compacted key bytes into internal row
```
platinumhamburg: binary key bytes (not just compacted)

fluss-common/src/main/java/org/apache/fluss/row/decode/KeyDecoder.java
```diff
@@ -0,0 +1,45 @@
...
+    /** Decode key bytes to a row containing only key fields, without non-key fields. */
+    InternalRow decodeKey(byte[] keyBytes);
+
+    /**
+     * Create a key decoder to decode the key array bytes of the input row.
+     *
+     * @param rowType the row type of the input row
+     * @param keyFields the key fields to encode
+     */
+    static KeyDecoder of(RowType rowType, List<String> keyFields) {
```
platinumhamburg: The format parameter must be provided; CompactedRow is an optional format, and others are not supported for now.

fluss-common/src/main/java/org/apache/fluss/row/decode/KeyDecoder.java
```diff
@@ -0,0 +1,45 @@
...
+    InternalRow decodeKey(byte[] keyBytes);
+
+    /**
+     * Create a key decoder to decode the key array bytes of the input row.
+     *
+     * @param rowType the row type of the input row
+     * @param keyFields the key fields to decode
+     */
+    static KeyDecoder of(RowType rowType, List<String> keyFields) {
+        return CompactedKeyDecoder.createKeyDecoder(rowType, keyFields);
```
beryllw: Will add KeyDecoders for Paimon and Iceberg later.

fluss-server/src/main/java/org/apache/fluss/server/replica/ReplicaManager.java
```diff
@@ -617,8 +673,80 @@ public void lookups(
...
+        for (Map.Entry<TableBucket, MissingKeysContext> entry :
+                entriesPerBucketToInsert.entrySet()) {
+            TableBucket tb = entry.getKey();
+            if (failedBuckets.contains(tb)) {
+                continue; // Skip buckets that failed during insert
+            }
+            MissingKeysContext missingKeysContext = entry.getValue();
+            Replica replica = getReplicaOrException(tb);
+            TableMetricGroup tableMetrics = replica.tableMetrics();
+            tableMetrics.totalLookupRequests().inc();
```
platinumhamburg: No need to increment here.

fluss-common/src/main/java/org/apache/fluss/record/KeyRecordBatch.java
```diff
@@ -0,0 +1,200 @@
...
+                if (keyIdx < keyFieldIndexes.length && keyFieldIndexes[keyIdx] == i) {
+                    // This is a key field - copy from decoded key row
+                    Object value = keyFieldGetters[keyIdx].getFieldOrNull(keyRow);
+                    rowEncoder.encodeField(i, value);
+                    keyIdx++;
+                } else {
+                    // Non-key field - set to null
+                    rowEncoder.encodeField(i, null);
+                }
+            }
```
platinumhamburg, line 219: The row construction logic incorrectly assumes `keyFieldIndexes` is sorted in ascending order by column position. However, `Schema.getPrimaryKeyIndexes()` returns indexes in PRIMARY KEY declaration order, not column order.

platinumhamburg: Need to add tests for KeyRecordBatch.

fluss-common/src/main/java/org/apache/fluss/record/KeyRecordBatch.java
```diff
@@ -0,0 +1,200 @@
...
+    }
+
+    @Override
+    public int getRecordCount() {
+        return keyBytes.size();
+    }
+
+    @Override
+    public short schemaId() {
+        return 0;
```
platinumhamburg: Is it safe to use constant 0 here?

platinumhamburg:
I’ve thought about it carefully, and I think KeyRecordBatch is too specific—especially since it’s meant for a temporary solution we all agree isn’t final. We could just remove it entirely and use a simpler approach instead. But if we’re short on time, it’s not a big deal—it doesn’t add much code—so we can keep it for now and clean it up later.

fluss-common/src/main/java/org/apache/fluss/row/decode/KeyDecoder.java
```diff
@@ -0,0 +1,80 @@
...
+                "Unsupported kv format version: " + kvFormatVersion);
+    }
+
+    /**
+     * Create a key decoder to decode the key array bytes of the input row.
+     *
+     * @param rowType the row type of the input row
+     * @param keyFields the key fields to decode
+     */
+    static KeyDecoder of(
```
platinumhamburg: This method appears to be dead code now.

fluss-server/src/test/java/org/apache/fluss/server/replica/ReplicaManagerTest.java
```diff
@@ -793,6 +802,123 @@ void testLookup() throws Exception {
...
+        // Scenario 3: Mixed - key100 exists, key300 missing
+        byte[] key300 = keyEncoder.encodeKey(row(new Object[] {300}));
+        List<byte[]> mixed = lookupWithInsert(tb, Arrays.asList(key100, key300)).lookupValues();
+        assertThat(mixed.get(0)).isEqualTo(inserted.get(0)); // existing
+        assertThat(mixed.get(1)).isNotNull(); // newly inserted
+        verifyLookup(tb, key300, mixed.get(1));
+    }
+
+    @Test
+    void testLookupWithInsertIfNotExistsAutoIncrement() throws Exception {
```
wuchong, line 857: Could you add an additional test for **concurrent lookups** with **insert-if-not-exists** and **auto-increment enabled**?

The test should:
- Spawn multiple threads (e.g., 3) that concurrently perform `lookupWithInsert` operations on the **same set of keys** (e.g., 100, 200, 300).
- Verify that all threads receive **consistent lookup results** (i.e., the same inserted values for each key).
- Confirm that exactly **3 changelog entries** are written to the log tablet—one per unique key.

This will validate that concurrent put-and-relookup operations behave correctly under contention, ensuring idempotency and consistency when auto-increment and conditional inserts are enabled.

fluss-common/src/main/java/org/apache/fluss/row/decode/KeyDecoder.java
```diff
@@ -0,0 +1,69 @@
...
+            @Nullable DataLakeFormat lakeFormat,
+            boolean isDefaultBucketKey) {
+        if (kvFormatVersion == 1 || (kvFormatVersion == 2 && isDefaultBucketKey)) {
+            if (lakeFormat == null || lakeFormat == DataLakeFormat.LANCE) {
+                return CompactedKeyDecoder.createKeyDecoder(rowType, keyFields);
+            }
+            if (lakeFormat == DataLakeFormat.PAIMON) {
+                // TODO: Implement key decoding support for Paimon lake format
+                throw new UnsupportedOperationException(
+                        "Paimon lake format does not support key decoding");
```
wuchong: How much effort would it take to support this? I believe it’s essential—since most clusters will enable lakehouse integration, this will become a common and critical path.  

If this requires significant work, please create a **blocker issue** to track it. And please add tests for this case as well.

fluss-common/src/main/java/org/apache/fluss/row/decode/KeyDecoder.java
```diff
@@ -0,0 +1,69 @@
...
+ * Interface for decoding key bytes back to {@link InternalRow}.
+ *
+ * <p>This interface provides functionality to decode binary key bytes into internal row
+ * representation, typically used for primary key decoding in KV tables.
+ */
+public interface KeyDecoder {
+    /** Decode key bytes to a row containing only key fields, without non-key fields. */
+    InternalRow decodeKey(byte[] keyBytes);
+
+    static KeyDecoder ofPrimaryKeyDecoder(
```
wuchong, line 53: Add Javadoc to describe this method's behavior, and include a reference link to `org.apache.fluss.row.encode.KeyEncoder#ofPrimaryKeyEncoder`, as their behaviors are aligned.

fluss-server/src/main/java/org/apache/fluss/server/replica/ReplicaManager.java
```diff
@@ -633,8 +693,84 @@ public void lookups(
...
+            if (!produceEntryData.isEmpty()) {
+                // TODO: Performance optimization: during lookup-with-insert-if-not-exists flow,
+                // the original key bytes are wrapped in KeyRecordBatch, then during putRecordsToKv
+                // they are decoded to rows and immediately re-encoded back to key bytes, causing
+                // redundant encode/decode overhead.
+                putRecordsToKv(
+                        timeoutMs,
+                        requiredAcks,
+                        produceEntryData,
+                        null,
```
wuchong: This is incorrect. We should use **partial updates** to modify only the primary key fields. Otherwise, non-primary-key columns—including auto-increment fields—will be overwritten with `null`. This issue manifests when multiple threads concurrently call `lookupAndInsert` on the same keys.

To reproduce this, I’ve added a test: `testConcurrentLookupWithInsertIfNotExistsAutoIncrement`.

Additionally, we should enforce a validation in `PutRequest`: when performing a put operation, **auto-increment fields must be excluded from the target columns**. This ensures that auto-incremented values are never accidentally overwritten during updates. That means `INSERT INTO t (id, auto_inc) VALUES ...` or `INSERT INTO t VALUES ...` should fail with explicit error message that the auto increment fields can't be updated.

fluss-server/src/main/java/org/apache/fluss/server/replica/ReplicaManager.java
```diff
@@ -633,8 +693,84 @@ public void lookups(
                         tb, new LookupResultForBucket(tb, ApiError.fromThrowable(e)));
             }
         }
+
+        if (insertIfNotExists) {
```
xx789633, line 701: Could you please also add a comment here to indicate that there might be a data race and describe how you solve it? The same key might be inserted by another thread after the lookup check passes.

fluss-server/src/test/java/org/apache/fluss/server/kv/KvTabletTest.java
```diff
@@ -662,6 +663,43 @@ void testAutoIncrementWithMultiThread() throws Exception {
...
+                        "Auto-increment column [uid] at index 1 must not be included in target columns.");
+
+        // no specify target columns, which is also invalid for auto-increment
+        assertThatThrownBy(() -> kvTablet.putAsLeader(kvRecordBatch, null))
+                .isInstanceOf(InvalidTargetColumnException.class)
+                .hasMessageContaining(
+                        "The table contains an auto-increment column [uid], but update target columns are not explicitly specified.");
+
+        // valid case: target columns contain auto-increment column
+        kvTablet.putAsLeader(kvRecordBatch, new int[] {0, 2});
```
platinumhamburg, line 694: Ditto, it appears that the actual behavior of the test code is opposite to what the comments state. Moreover, the test constructs records containing {user_name, uid} (indexes 0, 1), but passes targetColumns = {0, 2}, which refers to {user_name, age}—this is a code smell.

fluss-server/src/test/java/org/apache/fluss/server/kv/KvTabletTest.java
```diff
@@ -662,6 +663,43 @@ void testAutoIncrementWithMultiThread() throws Exception {
...
+                        schema.getRowType().project(Arrays.asList("user_name", "uid")));
+
+        KvRecordBatch kvRecordBatch =
+                kvRecordBatchFactory.ofRecords(
+                        Arrays.asList(
+                                recordFactory.ofRecord("k1".getBytes(), new Object[] {"k1", null}),
+                                recordFactory.ofRecord(
+                                        "k2".getBytes(), new Object[] {"k2", null})));
+        // target columns don't contain auto-increment column
+        assertThatThrownBy(() -> kvTablet.putAsLeader(kvRecordBatch, new int[] {0, 1}))
```
platinumhamburg, line 682: The auto-incrementing index ID is 1, which suggests that the validation logic in the code here is inconsistent with the comments.

fluss-server/src/main/java/org/apache/fluss/server/kv/partialupdate/PartialUpdater.java
```diff
@@ -100,6 +102,11 @@ private void sanityCheck(Schema schema, int[] targetColumns) {
      * @return the updated value (schema id + row bytes)
      */
     public BinaryValue updateRow(@Nullable BinaryValue oldValue, BinaryValue partialValue) {
+        if (updatePrimaryKeyOnly && oldValue != null) {
+            // only primary key columns are updated, return the old value directly
+            return oldValue;
```
platinumhamburg, line 107: @wuchong This optimization appears to have broken the CI tests.

PR #2490, opened 2026-01-27
[hotfix] Fix unstable tiering source enumerator test

luoyuxia:
### Purpose

Linked issue: close #xxx
Fix unstable test
 https://github.com/apache/fluss/actions/runs/21382698242/job/61552666939

### Brief change log
testHandleFailedTieringTableEvent have two tables, table1, table2, whose freshness is 100ms,
it may happen that table1 already add to pending queue, and mark it as fail, the table2 is still not added to the queue.
So, the next request will still get table1 which cause test fails.

I can reproduce it with repeat 100 tests.

I change table freshness to 10ms, then repeat 100 tests. The test won't fail.


beryllw:
LGTM!

PR #2492, opened 2026-01-27
[FLUSS] Add AggMode support for undo recovery in aggregation tables

### LINKED ISSUE #2491: Add MergeMode support for undo recovery in aggregation tables
### Search before asking

### Description

### Willingness to contribute

platinumhamburg:
This commit introduces AggMode (Aggregation Mode) to control how the server handles data aggregation when writing to tables with aggregation merge engine.

Key changes:

1. New AggMode enum with three modes:
   - AGGREGATE (default): Data is aggregated through server-side merge engine
   - OVERWRITE: Bypass merge engine, directly replace values (for undo recovery)
   - LOCAL_AGGREGATE: Reserved for future client-side pre-aggregation

2. Client-side changes:
   - Upsert interface: Added aggregationMode(AggMode) method for fluent API
   - UpsertWriterImpl: Propagates aggMode through WriteRecord
   - KvWriteBatch: Validates aggMode consistency within batch
   - ClientRpcMessageUtils: Validates aggMode consistency across batches
   - WriteRecord: Added aggMode field for upsert/delete operations

3. Server-side changes:
   - KvTablet: Pre-creates overwriteRowMerger for OVERWRITE mode
   - putAsLeader(): Selects appropriate RowMerger based on aggMode
   - Replica/ReplicaManager: Propagates aggMode through call chain
   - TabletService: Extracts aggMode from PutKvRequest

4. Protocol changes:
   - FlussApi.proto: Added optional agg_mode field to PutKvRequest

5. Test coverage:
   - KvTabletAggModeTest: 9 tests covering OVERWRITE mode scenarios
   - KvWriteBatchTest: 3 tests for aggMode consistency validation
   - ClientRpcMessageUtilsTest: 4 tests for multi-batch aggMode validation

This feature enables Flink connector to perform undo recovery by restoring exact historical values during checkpoint failover, bypassing the merge engine.

### Purpose

Linked issue: close #2491 2491

### DIFF
--- a/fluss-server/src/main/java/org/apache/fluss/server/replica/ReplicaManager.java
+++ b/fluss-server/src/main/java/org/apache/fluss/server/replica/ReplicaManager.java
@@ -550,27 +550,6 @@ public class ReplicaManager {
     }
 
     /**
-     * Put kv records to leader replicas of the buckets with default DEFAULT mode.
-     *
-     * <p>This is a convenience method that calls {@link #putRecordsToKv(int, int, Map, int[],
-     * MergeMode, Consumer)} with {@link MergeMode#DEFAULT}.
-     */
-    public void putRecordsToKv(
-            int timeoutMs,
-            int requiredAcks,
-            Map<TableBucket, KvRecordBatch> entriesPerBucket,
-            @Nullable int[] targetColumns,
-            Consumer<List<PutKvResultForBucket>> responseCallback) {
-        putRecordsToKv(
-                timeoutMs,
-                requiredAcks,
-                entriesPerBucket,
-                targetColumns,
-                MergeMode.DEFAULT,
-                responseCallback);
-    }
-
-    /**
      * Put kv records to leader replicas of the buckets, the kv data will write to kv tablet and the
      * response callback need to wait for the cdc log to be replicated to other replicas if needed.
      *
--- a/fluss-server/src/test/java/org/apache/fluss/server/replica/ReplicaManagerTest.java
+++ b/fluss-server/src/test/java/org/apache/fluss/server/replica/ReplicaManagerTest.java
@@ -52,6 +52,7 @@ import org.apache.fluss.rpc.entity.ProduceLogResultForBucket;
 import org.apache.fluss.rpc.entity.PutKvResultForBucket;
 import org.apache.fluss.rpc.protocol.ApiError;
 import org.apache.fluss.rpc.protocol.Errors;
+import org.apache.fluss.rpc.protocol.MergeMode;
 import org.apache.fluss.server.entity.FetchReqInfo;
 import org.apache.fluss.server.entity.NotifyLeaderAndIsrData;
 import org.apache.fluss.server.entity.NotifyLeaderAndIsrResultForBucket;
@@ -481,6 +482,7 @@ class ReplicaManagerTest extends ReplicaTestBase {
                 1,
                 Collections.singletonMap(tb, genKvRecordBatch(DATA_1_WITH_KEY_AND_VALUE)),
                 null,
+                MergeMode.DEFAULT,
                 PUT_KV_VERSION,
                 future::complete);
         assertThat(future.get()).containsOnly(new PutKvResultForBucket(tb, 8));
@@ -494,6 +496,7 @@ class ReplicaManagerTest extends ReplicaTestBase {
                                         Collections.singletonMap(
                                                 tb, genKvRecordBatch(DATA_1_WITH_KEY_AND_VALUE)),
                                         null,
+                                        MergeMode.DEFAULT,
                                         PUT_KV_VERSION,
                                         (result) -> {
                                             // do nothing.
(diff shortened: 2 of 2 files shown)


fluss-client/src/main/java/org/apache/fluss/client/table/writer/TableUpsert.java
```diff
@@ -91,9 +94,15 @@ public Upsert partialUpdate(String... targetColumnNames) {
...
+    @Override
+    public Upsert aggregationMode(AggMode mode) {
+        checkNotNull(mode, "aggregation mode");
+        return new TableUpsert(tablePath, tableInfo, writerClient, this.targetColumns, mode);
+    }
+
     @Override
     public UpsertWriter createWriter() {
-        return new UpsertWriterImpl(tablePath, tableInfo, targetColumns, writerClient);
+        return new UpsertWriterImpl(tablePath, tableInfo, targetColumns, writerClient, aggMode);
```
wuchong: We should add check that if `aggMode` is `OVERWRITE` or `LOCAL_AGGREGATE`, the table should be aggregation merge engine.

fluss-client/src/main/java/org/apache/fluss/client/table/writer/TableUpsert.java
wuchong, line 133: We should add check if the `aggMode` is not the default value, we should throw exception, because the `TypedUpsertWriterImpl` doesn't support the `aggMode` yet.

fluss-client/src/main/java/org/apache/fluss/client/write/WriteBatch.java
```diff
@@ -120,6 +121,15 @@ public abstract boolean tryAppend(WriteRecord writeRecord, WriteCallback callbac
 
     public abstract void abortRecordAppends();
 
+    /**
+     * Get the aggregation mode for this batch (only applicable to KV batches).
+     *
+     * @return the aggregation mode, defaults to AGGREGATE for log batches
+     */
+    public AggMode getAggMode() {
```
wuchong: It's very confusing that `ArrowLogWriteBatch` has a default `AggMode#AGGREGATE`. I checked the code, it seems that providing this method only in `KvWriteBatch` can work.

fluss-common/src/main/java/org/apache/fluss/rpc/protocol/AggMode.java
```diff
@@ -0,0 +1,123 @@
...
+/**
+ * Aggregation mode for write operations.
+ *
+ * <p>This enum controls how the server handles data aggregation when writing to tables with
+ * aggregation merge engine.
+ *
+ * @since 0.9
+ */
+@PublicEvolving
+public enum AggMode {
```
wuchong: I’ve reconsidered the enum API design, and it’s indeed confusing to have an `aggMode` flag on every `PutRequest`, especially since it implies aggregation-specific semantics while can being used across different merge engines. A more general and consistent approach would be to introduce a **`MergeMode`** that applies uniformly to all merge engines.

Here’s the proposed refinement:

1. **Rename `AggMode#AGGREGATE` to `MergeMode#DEFAULT`**  
   This represents the default merge behavior for each engine:  
   - For the *aggregate* merge engine: perform aggregation.  
   - For the *first-row* merge engine: retain the first observed row.  
   - For the *versioned* merge engine: keep the row with the highest version.  

2. **Rename `AggMode#OVERWRITE` to `MergeMode#OVERWRITE`**  
   This mode bypasses all merge logic and directly replaces existing values with new ones—similar to an `OVERWRITE INTO` statement. It’s useful for scenarios like undo/recovery across all merge engines.

3. **Remove `AggMode#LOCAL_AGGREGATE`**  
   This is purely a client-side optimization that performs pre-aggregation before sending data. It actually doesn't change the protocol between client and server. The client still send the result using `MergeMode#DEFAULT`, as the on-wire schema remains unchanged.

This change simplifies the API, improves consistency across merge engines, and clarifies the intent of each mode.

If you agree on the new API, we can rename the enum type in this PR, and postpone the general support MergeMode for all merge eninges in next PR. But if you decide to support it in this PR, please remember to add tests.

platinumhamburg:
> Thanks @platinumhamburg , the code looks in a high quality. I left some comments below.
> 
> Besides, I noticed that there is no any (agg mode or overwrite mode) IT case from client side? Is it planned in next PRs?

Yes, in the next PR.

fluss-common/src/main/java/org/apache/fluss/rpc/protocol/AggMode.java
platinumhamburg: I think this new design makes more sense, and I'll revise it according to this design.

platinumhamburg:
Thanks @wuchong , I've addressed all comments above, please take another look when you have time.

wuchong:
@platinumhamburg LGTM. Could you rebase and resolve the conflicts?

platinumhamburg:
> @platinumhamburg LGTM. Could you rebase and resolve the conflicts?

@wuchong Yes, I've rebased this PR and resolved all conflicts.

wuchong (approved):

PR #2493, opened 2026-01-27
Remove system column paimon v2

luoyuxia:
### Purpose

Linked issue: close #xxx


fluss-lake/fluss-lake-paimon/src/main/java/org/apache/fluss/lake/paimon/source/PaimonRecordReader.java
```diff
@@ -143,14 +143,21 @@ public boolean hasNext() {
...
-                    timestamp,
+                    -1L,
+                    -1L,
                     changeType,
                     projectedRow.replaceRow(paimonRowAsFlussRow.replaceRow(paimonRow)));
         }
     }
+
+    // for legacy table, we will have system column
+    private static boolean hasSystemColumn(RowType paimonRowType) {
```
luoyuxia: I just found in much places, we call similar code to check system, we can extract it to a common method, and all call this method:
```
public final class PaimonUtils {

      private PaimonUtils() {}

      /**
       * Returns whether the physical Paimon table uses the legacy layout.
       *
       * <p>A legacy table is identified by the presence of the {@code __timestamp}
       * system column.
       */
      public static boolean isLegacyTable(RowType tableRowType) {
          return tableRowType.getFieldIndex(TIMESTAMP_COLUMN_NAME) >= 0;
      }
  }
```

luoyuxia: we also need to call `checkSystemColumns(newSchema.getRowType());` in method `validateAlterTableSchema` to prevent users add such column:
```
public static void validateAlterTableSchema(TableInfo table, Schema newSchema) {
      checkSystemColumns(newSchema.getRowType());
      ...
  }
```

fluss-server/src/main/java/org/apache/fluss/server/coordinator/CoordinatorService.java
```diff
@@ -506,9 +467,13 @@ private TableDescriptor applySystemDefaults(
         }
 
         // override the datalake format if the table hasn't set it and the cluster configured
-        if (dataLakeFormat != null
-                && !properties.containsKey(ConfigOptions.TABLE_DATALAKE_FORMAT.key())) {
-            newDescriptor = newDescriptor.withDataLakeFormat(dataLakeFormat);
+        if (dataLakeFormat != null) {
```
luoyuxia: I think we don't need to change it.

Issue #2496, opened 2026-01-27
[FIP-10] Introduce Predicate in Protobuf and RPC Layer


polyzos:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/fluss/issues) and found nothing similar.

### Description

Introduce Predicate in Protobuf and RPC Layer

### Willingness to contribute

- [ ] I'm willing to submit a PR!

PR #2502, opened 2026-01-28
Support Protobuf and RPC for Predicates

polyzos:
This closes [#2496](https://github.com/apache/fluss/issues/2496) and sets the foundation for predicate pushdown 

#### 2. Predicate Serialization Utilities
- **`PbPredicateConverter`**
- **`PredicateMessageUtils`**: A utility class for converting between internal models and RPC messages. It includes entry points for `Filter` and `Predicate` serialization/deserialization, ensuring the server can evaluate predicates sent by clients.
- **`CommonRpcMessageUtils`**: Standardized `DataType` $\leftrightarrow$ `PbDataType` conversions were centralized here to promote reuse across the RPC module.

#### 3. Schema-Aware Filtering and Statistics
- **`Filter` Class**: Introduced a schema-aware `Filter` that wraps predicates with a `schemaId`. This ensures that filtering logic is only applied when the data schema matches the expected filter schema, preventing errors during schema evolution.
- **Statistics Integration**: Added the `LogRecordBatchStatistics` interface and integrated it into the `Filter` logic.


polyzos:
FYI @platinumhamburg  @wuchong

fluss-rpc/src/main/proto/FlussApi.proto
```diff
@@ -1230,4 +1230,91 @@ message PbTableStatsRespForBucket {
...
+  optional float float_value = 5;
+  optional double double_value = 6;
+  optional string string_value = 7;
+  optional bytes binary_value = 8;
+  optional int64 decimal_value = 9;    // Serialized decimal (compact mode)
+  optional bytes decimal_bytes = 10;   // Serialized decimal (non-compact mode)
+  // field numbers 11, 12 reserved for future use
+  optional int64 timestamp_millis_value = 13; // Epoch millis
+  optional int32 timestamp_nano_of_millis_value = 14 [default = 0]; // Nano of millis
+  required bool is_null = 15 [default = false];
```
wuchong: nit: we can move the `is_null` to the field after `type`, because they are both `required` fields. And all the other fields and future fields are literal value fields.

fluss-rpc/src/main/proto/FlussApi.proto
```diff
@@ -1230,4 +1230,91 @@ message PbTableStatsRespForBucket {
...
+  optional int64 bigint_value = 4;
+  optional float float_value = 5;
+  optional double double_value = 6;
+  optional string string_value = 7;
+  optional bytes binary_value = 8;
+  optional int64 decimal_value = 9;    // Serialized decimal (compact mode)
+  optional bytes decimal_bytes = 10;   // Serialized decimal (non-compact mode)
+  // field numbers 11, 12 reserved for future use
+  optional int64 timestamp_millis_value = 13; // Epoch millis
+  optional int32 timestamp_nano_of_millis_value = 14 [default = 0]; // Nano of millis
```
wuchong: Do not use `default`, because proto3 removes the support of "default" keyword.

fluss-rpc/src/main/proto/FlussApi.proto
```diff
@@ -1230,4 +1230,91 @@ message PbTableStatsRespForBucket {
...
+  optional bool boolean_value = 2;
+  optional int32 int_value = 3;
+  optional int64 bigint_value = 4;
+  optional float float_value = 5;
+  optional double double_value = 6;
+  optional string string_value = 7;
+  optional bytes binary_value = 8;
+  optional int64 decimal_value = 9;    // Serialized decimal (compact mode)
+  optional bytes decimal_bytes = 10;   // Serialized decimal (non-compact mode)
+  // field numbers 11, 12 reserved for future use
```
wuchong: Is supporting fields 11 and 12 technically challenging? My understanding is that these represent time-related fields. I recommend enabling support for them directly. If that is not feasible, I suggest relocating `timestamp_millis_value` and `timestamp_nano_of_millis_value` to positions 11 and 12 respectively to maintain schema consistency.

fluss-rpc/src/main/proto/FlussApi.proto
```diff
@@ -1230,4 +1230,91 @@ message PbTableStatsRespForBucket {
...
+  optional PbLeafPredicate leaf = 2;
+  optional PbCompoundPredicate compound = 3;
+}
+
+// Represents a leaf predicate that compares a field with literals
+message PbLeafPredicate {
+  // The function to apply (see PbLeafFunction constants: EQUAL=0 ... NOT_IN=12)
+  required int32 function = 1;
+  // The field reference
+  required PbFieldRef field_ref = 2;
```
wuchong: Could we simplify field references by using the field id (an `int32`) directly? This approach would significantly reduce complexity and serialization overhead. Since field names, data types, and indices can be efficiently derived from the Schema using the schema_id.

fluss-rpc/src/main/proto/FlussApi.proto
```diff
@@ -1230,4 +1230,91 @@ message PbTableStatsRespForBucket {
...
+
+// PbDataTypeRoot constants (using int32 instead of enum for proto3 compatibility)
+// BOOLEAN=0, TINYINT=1, SMALLINT=2, INT=3, BIGINT=4, FLOAT=5, DOUBLE=6,
+// CHAR=7, VARCHAR=8, DECIMAL=9, DATE=10, TIME_WITHOUT_TIME_ZONE=11,
+// TIMESTAMP_WITHOUT_TIME_ZONE=12, TIMESTAMP_WITH_LOCAL_TIME_ZONE=13,
+// BINARY=14, BYTES=15
+
+// Represents a literal value
+message PbLiteralValue {
+  required PbDataType type = 1;
```
wuchong: I think the data type of the literal should be the same with the field reference type, so there is no need to introduce the `PbDataType` for `PbLiteralValue`? 

Besides, introducing a simple int literal_type field would be beneficial to efficiently indicate which specific value field contains the literal, streamlining the extraction logic without iterating all the fields, and making the `PbLiteralValue` self-contained.

fluss-rpc/src/main/java/org/apache/fluss/rpc/util/PredicateMessageUtils.java
```diff
@@ -0,0 +1,564 @@
...
+    //  Proto int32 <-> domain object mapping enums
+    // -------------------------------------------------------------------------
+
+    /** Maps PbPredicate.type int32 values to predicate kinds. */
+    private enum PredicateType {
+        LEAF(0),
+        COMPOUND(1);
+
+        private final int value;
+        private static final Map<Integer, PredicateType> VALUE_MAP = new HashMap<>();
```
wuchong: nit: I think using `PredicateType[]`  array would be more efficient than `HashMap`, because all keys are sequences from 0. This also applies to `LeafFunctionCode#VALUE_MAP`, `CompoundFunctionCode#VALUE_MAP`, `DataTypeRootCode#VALUE_MAP`.

polyzos:
@wuchong comments were addressed.. PTAL

PR #2506, opened 2026-01-28
[helm] Enable SASL authentication configurations

### LINKED ISSUE #2503: [helm] Enable SASL authentication configurations
### Search before asking

### Description

Users should be able to configure the SASL authentication options in the Helm chart values.

### Willingness to contribute

morazow:
### Purpose

Linked issue: close #2503 

### Brief change log

Adds configuration options to Helm charts to enable SASL authentication.

### Documentation

Updated the deploying-with-helm documentation

### DIFF
--- a/helm/templates/_security.tpl
+++ b/helm/templates/_security.tpl
@@ -118,7 +118,43 @@ Usage:
 {{- end -}}
 
 {{/*
-Returns a warning if the internal SASL user is using default credentials.
+Returns the default internal SASL username based on the release name.
+Usage:
+  include "fluss.security.sasl.plain.internal.defaultUsername" .
+*/}}
+{{- define "fluss.security.sasl.plain.internal.defaultUsername" -}}
+{{- printf "fluss-internal-user-%s" .Release.Name -}}
+{{- end -}}
+
+{{/*
+Returns the default internal SASL password based on the release name (sha256 hashed).
+Usage:
+  include "fluss.security.sasl.plain.internal.defaultPassword" .
+*/}}
+{{- define "fluss.security.sasl.plain.internal.defaultPassword" -}}
+{{- printf "fluss-internal-password-%s" .Release.Name | sha256sum -}}
+{{- end -}}
+
+{{/*
+Returns the resolved internal SASL username (user-provided or auto-generated default).
+Usage:
+  include "fluss.security.sasl.plain.internal.username" .
+*/}}
+{{- define "fluss.security.sasl.plain.internal.username" -}}
+{{- .Values.security.internal.sasl.plain.username | default (include "fluss.security.sasl.plain.internal.defaultUsername" .) -}}
+{{- end -}}
+
+{{/*
+Returns the resolved internal SASL password (user-provided or auto-generated default).
+Usage:
+  include "fluss.security.sasl.plain.internal.password" .
+*/}}
+{{- define "fluss.security.sasl.plain.internal.password" -}}
+{{- .Values.security.internal.sasl.plain.password | default (include "fluss.security.sasl.plain.internal.defaultPassword" .) -}}
+{{- end -}}
+
+{{/*
+Returns a warning if the internal SASL user is using auto-generated credentials.
 Usage:
   include "fluss.security.sasl.warnInternalUser" .
 */}}
@@ -126,8 +162,8 @@ Usage:
 {{- if (include "fluss.security.sasl.enabled" .) -}}
   {{- $internalMechanism := include "fluss.security.listener.mechanism" (dict "context" .Values "listener" "internal") -}}
   {{- if eq $internalMechanism "plain" -}}
-    {{- if and (eq .Values.security.internal.sasl.plain.username "fluss-internal-user") (eq .Values.security.internal.sasl.plain.password "fluss-internal-password") -}}
-      {{- print "You are using the DEFAULT SASL credentials for internal communication.\n  It is strongly recommended to override the following values in production:\n    - security.internal.sasl.plain.username\n    - security.internal.sasl.plain.password" -}}
+    {{- if and (not .Values.security.internal.sasl.plain.username) (not .Values.security.internal.sasl.plain.password) -}}
+      {{- print "You are using AUTO-GENERATED SASL credentials for internal communication.\n  It is strongly recommended to set the following values in production:\n    - security.internal.sasl.plain.username\n    - security.internal.sasl.plain.password" -}}
     {{- end -}}
   {{- end -}}
 {{- end -}}
--- a/helm/templates/secret-jaas-config.yaml
+++ b/helm/templates/secret-jaas-config.yaml
@@ -19,6 +19,8 @@
 {{- if (include "fluss.security.sasl.plain.enabled" .) -}}
 {{- $internalMechanism := include "fluss.security.listener.mechanism" (dict "context" .Values "listener" "internal") -}}
 {{- $clientMechanism := include "fluss.security.listener.mechanism" (dict "context" .Values "listener" "client") -}}
+{{- $internalUsername := include "fluss.security.sasl.plain.internal.username" . -}}
+{{- $internalPassword := include "fluss.security.sasl.plain.internal.password" . -}}
 apiVersion: v1
 kind: Secret
 metadata:
@@ -31,12 +33,12 @@ stringData:
 {{- if eq $internalMechanism "plain" }}
     internal.FlussServer {
        org.apache.fluss.security.auth.sasl.plain.PlainLoginModule required
-       user_{{ .Values.security.internal.sasl.plain.username }}="{{ .Values.security.internal.sasl.plain.password }}";
+       user_{{ $internalUsername }}="{{ $internalPassword }}";
     };
     FlussClient {
        org.apache.fluss.security.auth.sasl.plain.PlainLoginModule required
-       username="{{ .Values.security.internal.sasl.plain.username }}"
-       password="{{ .Values.security.internal.sasl.plain.password }}";
+       username="{{ $internalUsername }}"
+       password="{{ $internalPassword }}";
     };
 {{- end }}
 {{- if eq $clientMechanism "plain" }}
--- a/helm/values.yaml
+++ b/helm/values.yaml
@@ -72,8 +72,8 @@ security:
       # "" | plain
       mechanism: ""
       plain:
-        username: "fluss-internal-user"
-        password: "fluss-internal-password"
+        username: ""
+        password: ""
 
 resources: {}
   # We usually recommend not to specify default resources and to leave this as a conscious
--- a/website/docs/install-deploy/deploying-with-helm.md
+++ b/website/docs/install-deploy/deploying-with-helm.md
@@ -190,10 +190,17 @@ The following table lists the configurable parameters of the Fluss chart, and th
 | `security.client.sasl.mechanism` | Client listener SASL mechanism (`""`, `plain`) | `""` |
 | `security.internal.sasl.mechanism` | Internal listener SASL mechanism (`""`, `plain`) | `""` |
 | `security.client.sasl.plain.users` | Client listener username and password pairs for PLAIN | `[]` |
-| `security.internal.sasl.plain.username` | Internal listener PLAIN username | `"fluss-internal-user"` |
-| `security.internal.sasl.plain.password` | Internal listener PLAIN password | `"fluss-internal-password"` |
+| `security.internal.sasl.plain.username` | Internal listener PLAIN username | `""` |
+| `security.internal.sasl.plain.password` | Internal listener PLAIN password | `""` |
 
 Only `plain` mechanism is supported for now. An empty string disables the SASL authentication, and maps to the `PLAINTEXT` protocol.
+
+If the internal SASL username or password is left empty, the chart automatically generates credentials based on the Helm release name:
+
+- Username is set to the `"fluss-internal-user-<release-name>"`
+- Password is set to the SHA-256 hash of `"fluss-internal-password-<release-name>"`
+
+It is recommended to set these explicitly in production.
 
 ### Fluss Configuration Overrides
 
--- a/helm/tests/security_test.yaml
+++ b/helm/tests/security_test.yaml
@@ -78,7 +78,7 @@ suite: plain-internal-defaults
 templates:
   - templates/secret-jaas-config.yaml
 tests:
-  - it: uses default internal credentials when user does not override them
+  - it: uses auto-generated internal credentials when user does not override them
     set:
       security.internal.sasl.mechanism: plain
     asserts:
@@ -86,13 +86,13 @@ tests:
           count: 1
       - matchRegex:
           path: stringData["jaas.conf"]
-          pattern: 'user_fluss-internal-user="fluss-internal-password"'
+          pattern: 'user_fluss-internal-user-RELEASE-NAME="[a-f0-9]{64}"'
       - matchRegex:
           path: stringData["jaas.conf"]
-          pattern: 'username="fluss-internal-user"'
+          pattern: 'username="fluss-internal-user-RELEASE-NAME"'
       - matchRegex:
           path: stringData["jaas.conf"]
-          pattern: 'password="fluss-internal-password"'
+          pattern: 'password="[a-f0-9]{64}"'
 
 ---
 
@@ -244,7 +244,7 @@ tests:
       security.client.sasl.mechanism: bogus
     asserts:
       - failedTemplate:
-          errorMessage: "VALUES VALIDATION:\nsecurity.client.sasl.mechanism must be one of: empty, plain"
+          errorMessage: "VALUES VALIDATION:\nsecurity.client.sasl.mechanism must be empty or: plain"
   - it: accepts empty mechanism as default
     asserts:
       - notFailedTemplate: {}


.github/workflows/helm-chart.yaml
```diff
@@ -0,0 +1,59 @@
...
+      - name: "Checkout code"
+        uses: actions/checkout@v6
+
+      - name: "Setup Helm"
+        uses: azure/setup-helm@v4
+
+      - name: "Lint Helm chart"
+        run: helm lint ./helm
+
+      - name: "Run helm-unittest"
```
affo, line 50: It seems there is an action already for this: https://github.com/marketplace/actions/helm-unit-tests.

You can also avoid the `setup` step using that one, first running the tests, and then lint 🤝

helm/templates/_helpers.tpl
```diff
@@ -63,4 +63,54 @@ Selector labels
...
 app.kubernetes.io/instance: {{ .Release.Name }}
-{{- end }}
\ No newline at end of file
+{{- end }}
+
+{{/*
+Generate JAAS configuration for SASL
+*/}}
+{{- define "fluss.sasl.jaasConfig" -}}
+{{- if .Values.sasl.jaasConfig }}
```
affo: I don't see this value in the values.
I would remove this if and the value, why offering the possibility to hardcode?

helm/templates/_helpers.tpl
```diff
@@ -63,4 +63,54 @@ Selector labels
...
+   user_{{ .username }}="{{ .password }}"
+   {{- end }};
+};
+{{- end }}
+{{- end }}
+
+{{/*
+Return true if SASL is configured in any of the listener protocols
+*/}}
+{{- define "fluss.isSaslEnabled" -}}
```
affo:
```suggestion
{{- define "fluss.sasl.enabled" -}}
```

helm/templates/_helpers.tpl
```diff
@@ -63,4 +63,54 @@ Selector labels
...
+  {{- $id := . -}}
+  {{- $l := index $ctx.Values.listeners $id -}}
+  {{- if regexFind "SASL" (upper $l.protocol) -}}
+    {{- $res = "true" -}}
+  {{- end -}}
+{{- end -}}
+{{- if $res -}}
+{{- true -}}
+{{- end -}}
+{{- end -}}
```
affo:
```suggestion
{{- $enabled := false -}}
{{- range $id, $l := .Values.listeners -}}
  {{- if and (not $enabled) (regexFind "SASL" (upper $l.protocol)) -}}
    {{- $enabled = true -}}
  {{- end -}}
{{- end -}}
{{- $enabled -}}
```

I think this would reach the same goal with less overhead 👍

helm/templates/_helpers.tpl
```diff
@@ -63,4 +63,54 @@ Selector labels
...
+
+{{/*
+Generate ID:SECURITY list for listener protocols
+*/}}
+{{- define "fluss.listeners.protocolMap" -}}
```
affo:
```suggestion
{{- define "fluss.listeners.securityProtocolMap" -}}
```

helm/templates/_helpers.tpl
```diff
@@ -63,4 +63,54 @@ Selector labels
...
+{{- $ctx := . -}}
+{{- $parts := list -}}
+{{- $keys := keys .Values.listeners | sortAlpha -}}
+{{- range $keys }}
+  {{- $id := . -}}
+  {{- $l := index $ctx.Values.listeners $id -}}
+  {{- $parts = append $parts (printf "%s:%s" (upper $id) (upper $l.protocol)) -}}
+{{- end -}}
+{{- join "," $parts -}}
+{{- end }}
```
affo: This looks a bit too much as of now, as you cannot effectively template listeners, but you can for the security protocol map.

I would hardcode this one similarly to listeners.

We can follow up with another PR to template both listeners and protocol map at the same time 🤝

helm/templates/sts-coordinator.yaml
```diff
@@ -77,6 +77,26 @@ spec:
...
+              {{- if eq (len $saslUsers) 0 -}}
+              {{- fail "sasl.users must contain at least one user when SASL is enabled in listeners" -}}
+              {{- end }}
+              {{- $jaasUsers := list -}}
+              {{- range $saslUsers }}
+              {{- $jaasUsers = append $jaasUsers (printf "user_%s=\\\"%s\\\"" .username .password) -}}
+              {{- end }}
+              {{- $clientUser := first $saslUsers -}}
+              echo "security.sasl.enabled.mechanisms: {{ .Values.sasl.mechanism }}" >> $FLUSS_HOME/conf/server.yaml && \
+              echo "security.sasl.plain.jaas.config: org.apache.fluss.security.auth.sasl.plain.PlainLoginModule required {{ join " " $jaasUsers }};" >> $FLUSS_HOME/conf/server.yaml && \
```
affo: I don't understand why having the Jaas secret mounted but also adding this complexity here.
It looks to me as a repetition 🤔

.github/workflows/helm-chart.yaml
```diff
@@ -0,0 +1,59 @@
...
+################################################################################
+
+name: "Helm Chart"
+
+permissions:
+  contents: read
+
+on:
+  pull_request:
+    branches: [main, release-*, ci-*]
```
affo:
```suggestion
```

helm/templates/_helpers.tpl
```diff
@@ -63,4 +63,54 @@ Selector labels
 {{- define "fluss.selectorLabels" -}}
 app.kubernetes.io/name: {{ include "fluss.name" . }}
 app.kubernetes.io/instance: {{ .Release.Name }}
-{{- end }}
\ No newline at end of file
+{{- end }}
+
+{{/*
```
affo, line 68: Better to move this templating part to a separate `_sasl.tpl` file

helm/values.yaml
```diff
@@ -85,3 +94,12 @@ serviceAccount:
...
   annotations: {}
+
+## Fluss SASL configurations for authentication.
+## These are required if SASL listeners are configured.
+sasl:
+  mechanism: PLAIN
+  # List of client users for authentication
+  users:
+    - username: admin
+      password: password
```
affo: maybe good to scope this under a root `security` key.
As after SASL other protocols will probably be supported.

Also, I think the mechanism cannot be set from the outside, as every mechanism has different inputs.

```suggestion
security:
  sasl_plain:
    # List of client users for authentication
    users:
      - username: admin
        password: password
```

helm/templates/_helpers.tpl
```diff
@@ -63,4 +63,54 @@ Selector labels
...
+Generate JAAS configuration for SASL
+*/}}
+{{- define "fluss.sasl.jaasConfig" -}}
+{{- if .Values.sasl.jaasConfig }}
+{{- .Values.sasl.jaasConfig -}}
+{{- else }}
+FlussServer {
```
affo: I like this approach, however, you would need to set

```
export FLUSS_ENV_JAVA_OPTS="-Djava.security.auth.login.config=..."
```

to make this work.

However, there is a caveat, this is a JVM-global setting, hence JAAS would be configured globally.
This would entail also Zookeeper 🤝 

Once adding authentication towards Zookeeper, this should be kept in mind as well.

I also think this can directly be moved to the secret.

helm/templates/secret-sasl.yaml
```diff
@@ -0,0 +1,29 @@
...
+apiVersion: v1
+kind: Secret
+metadata:
+  name: {{ include "fluss.fullname" . }}-sasl-jaas-config
+  labels:
+    {{- include "fluss.labels" . | nindent 4 }}
+type: Opaque
+data:
+  jaas.conf: {{ include "fluss.sasl.jaasConfig" . | b64enc | quote }}
+{{- end -}}
```
affo: I would add this file as generic, JVM-global, JAAS config.

hence, I would rename to secret-jaas-config.yaml.

```suggestion
{{- if (include "fluss.sasl.enabled" .) }}
apiVersion: v1
kind: Secret
metadata:
  name: {{ include "fluss.fullname" . }}-jaas-config
  labels:
    {{- include "fluss.labels" . | nindent 4 }}
type: Opaque
stringData:
  jaas.conf: |
    FlussServer {
       org.apache.fluss.security.auth.sasl.plain.PlainLoginModule required
       {{- range .Values.sasl.users }}
       user_{{ .username }}="{{ .password }}"
       {{- end }};
    };
    
    FlussClient {
      org.apache.fluss.security.auth.sasl.plain.PlainLoginModule required
      username="{{ }}"
      password="{{ }}";
    };
    
    # Space to add Zookeeper authentication below in the future
{{- end -}}
```

.github/workflows/helm-chart.yaml
morazow: Done! Somehow I cannot commit review changes from GitHub UI, addressed on local commits.

morazow:
Hello @affo ,

Thanks for the review! Please have another look 🤝

helm/templates/sts-coordinator.yaml
```diff
@@ -77,6 +77,16 @@ spec:
               echo "" >> $FLUSS_HOME/conf/server.yaml && \
               echo "bind.listeners: ${BIND_LISTENERS}" >> $FLUSS_HOME/conf/server.yaml && \
               echo "advertised.listeners: ${ADVERTISED_LISTENERS}" >> $FLUSS_HOME/conf/server.yaml && \
+              echo "security.protocol.map: INTERNAL:{{ upper .Values.listeners.internal.protocol }},CLIENT:{{ upper .Values.listeners.client.protocol }}" >> $FLUSS_HOME/conf/server.yaml && \
+
+              {{- if (include "fluss.sasl.enabled" .) }}
+              echo "security.sasl.enabled.mechanisms: PLAIN" >> $FLUSS_HOME/conf/server.yaml && \
+              {{ if regexFind "SASL" (upper .Values.listeners.internal.protocol) }}
```
morazow: This is required since the Fluss server itself acts as client to connect to tablets

affo: yeah, the configuration is not required to be here, but configuring SASL and its mechanism is, makes sense 🤝

helm/tests/sasl_test.yaml
```diff
@@ -0,0 +1,203 @@
...
+          path: spec.template.spec.containers[0].command[2]
+          pattern: 'client\.security\.protocol: SASL'
+---
+
+suite: sasl-enabled-statefulset-internal-plaintext
+templates:
+  - templates/sts-coordinator.yaml
+  - templates/sts-tablet.yaml
+tests:
+  - it: does not render internal client sasl config when internal listener is plaintext
```
affo:
```suggestion
  - it: does not render internal client sasl config for coordinator when internal listener is plaintext
```

website/docs/install-deploy/deploying-with-helm.md
```diff
@@ -225,6 +225,11 @@ The following table lists the configurable parameters of the Fluss chart and the
 | `resources.tabletServer.limits.cpu` | CPU limits for tablet servers | Not set |
 | `resources.tabletServer.limits.memory` | Memory limits for tablet servers | Not set |
 
+### SASL Parameters
+
+| Parameter | Description | Default |
+|-----------|-------------|---------|
+| `security.sasl_plain.users` | User list for PLAIN authentication | `[{username: admin, password: password}]` |
```
affo:
```suggestion
| `security.sasl_plain.users` | User list for SASL/PLAIN authentication | `[{username: admin, password: password}]` |
```

helm/values.yaml
```diff
@@ -85,3 +87,12 @@ serviceAccount:
...
   annotations: {}
+
+## Fluss security configurations for authentication.
+## These are required if SASL listeners are configured.
+security:
+  sasl_plain:
+    # List of client users for authentication
+    users:
+      - username: admin
+        password: password
```
xx789633: If the password is not and sasl is enabled, we can generate a random one for the user

helm/templates/secret-jaas-config.yaml
```diff
@@ -0,0 +1,48 @@
...
+  name: {{ include "fluss.fullname" . }}-sasl-jaas-config
+  labels:
+    {{- include "fluss.labels" . | nindent 4 }}
+type: Opaque
+stringData:
+  jaas.conf: |
+    FlussServer {
+       org.apache.fluss.security.auth.sasl.plain.PlainLoginModule required
+       {{- range $saslUsers }}
+       user_{{ .username }}="{{ .password }}"
```
xx789633, line 48: If the password is empty, I think we should either fail or generate a random one for the user. I prefer the latter approach.

helm/templates/sts-coordinator.yaml
```diff
@@ -77,6 +77,16 @@ spec:
               echo "" >> $FLUSS_HOME/conf/server.yaml && \
               echo "bind.listeners: ${BIND_LISTENERS}" >> $FLUSS_HOME/conf/server.yaml && \
               echo "advertised.listeners: ${ADVERTISED_LISTENERS}" >> $FLUSS_HOME/conf/server.yaml && \
+              echo "security.protocol.map: INTERNAL:{{ upper .Values.listeners.internal.protocol }},CLIENT:{{ upper .Values.listeners.client.protocol }}" >> $FLUSS_HOME/conf/server.yaml && \
+
+              {{- if (include "fluss.sasl.enabled" .) }}
+              echo "security.sasl.enabled.mechanisms: PLAIN" >> $FLUSS_HOME/conf/server.yaml && \
```
xx789633: PLAIN is hardcoded. We might support more SASL mechanisms in the future.

helm/templates/sts-tablet.yaml
```diff
@@ -74,6 +74,16 @@ spec:
               echo "tablet-server.id: ${FLUSS_SERVER_ID}" >> $FLUSS_HOME/conf/server.yaml && \
               echo "bind.listeners: ${BIND_LISTENERS}" >> $FLUSS_HOME/conf/server.yaml && \
               echo "advertised.listeners: ${ADVERTISED_LISTENERS}" >> $FLUSS_HOME/conf/server.yaml && \
+              echo "security.protocol.map: INTERNAL:{{ upper .Values.listeners.internal.protocol }},CLIENT:{{ upper .Values.listeners.client.protocol }}" >> $FLUSS_HOME/conf/server.yaml && \
```
xx789633: The SASL logic between coordinator and tablet StatefulSets seems to be duplicated. can we extract them into a template?

helm/tests/sasl_test.yaml
```diff
@@ -0,0 +1,203 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+suite: sasl-disabled-secret
+templates:
+  - templates/secret-jaas-config.yaml
+tests:
```
xx789633: No test for multiple SASL users in JAAS config.

helm/templates/secret-jaas-config.yaml
xx789633, line 48: If the password contains special characters, can this script still handle it?

helm/templates/sts-coordinator.yaml
xx789633, line 102: Values.resources.tabletServer?

helm/values.yaml
```diff
@@ -85,3 +87,12 @@ serviceAccount:
   # Additional annotations to apply to the ServiceAccount.
   # These can be useful, for example, to support integrations like workload identity.
   annotations: {}
+
+## Fluss security configurations for authentication.
+## These are required if SASL listeners are configured.
+security:
+  sasl_plain:
```
xx789633: use camelCase

website/docs/install-deploy/deploying-with-helm.md
```diff
@@ -245,16 +250,47 @@ The chart automatically configures listeners for internal cluster communication
...
+
+```yaml
+security:
+  sasl_plain:
+    users:
+      - username: admin
+        password: password
+```
+
+The `security.sasl_plain.users` field defines the list of usernames and passwords for SASL/PLAIN authentication. When the internal listener protocol uses SASL, the first user in the list is used for internal client authentication. The authentication mechanism is fixed to `PLAIN`.
```
xx789633: There seems to be no support for separate internal and client credentials. Can we provide variables like `auth.sasl.jaas.clientPasswords` and `auth.sasl.jaas.interBrokerPassword` to distinguish the credentials?

xx789633 (commented):
Hi @morazow thanks for the pull request. I left some comments.

helm/templates/secret-jaas-config.yaml
affo: I think @morazow opted for failing in that case, you can see it at line 21:

```
{{- fail "security.sasl_plain.users must contain at least one user when SASL is enabled in listeners" -}}
```

so, he gives for granted that there is at least 1 user

affo: @xx789633 I would favor failing, as otherwise the cluster would be up, SASL enable, and the user would need to explore the config via `kubectl` to find the password to use.
Failing is a way for enforcing that the user is aware of the username/password.

Why would you generate a random one?
I think that would be ok for the internal communication, but not for the user-facing one.

helm/templates/sts-coordinator.yaml
morazow: Good find 👍

helm/values.yaml
morazow: Done!

helm/templates/sts-coordinator.yaml
morazow: For now, let's keep it as `PLAIN`. We can add other mechanisms in the future, once we have TLS support for example.

morazow: It is not clear at the moment how different mechanisms will be used from the users perspective 🤝

helm/templates/secret-jaas-config.yaml
morazow: Done, added tests for it

helm/templates/sts-tablet.yaml
morazow: Let's do it in one of the next PRs, it was suggested before, just to keep clearer for now I kept them. The whole duplicated block could be extracted definitely.

website/docs/install-deploy/deploying-with-helm.md
morazow: I am going to redesign this, it is good suggestion thanks! 

But opt for generating password only for the inter communication, for client or external to fail if password is not set.

helm/tests/sasl_test.yaml
morazow: Fixed 🚀

website/docs/install-deploy/deploying-with-helm.md
morazow: Hey @xx789633, @affo 

I have looked into this. But I think we should address it in the separate PR, because seems it would require change in the core fluss also.

### Context

With multiple separate SASL communications, we would have to prefix JAAS configurations for `FlussServer` and `FlussClient` with, for example, listener name. E.g, `internal.FlussServer { ... }`.

This work at the moment for the server side. 

But client side needs change in the core.

Loading of the client JAAS context doesn't consider the prefixed listener names: 

https://github.com/apache/fluss/blob/main/fluss-common/src/main/java/org/apache/fluss/security/auth/sasl/jaas/JaasContext.java#L163

morazow:
Hello @xx789633, @affo 

Thanks for the suggestions! Please have a look to the PR again.

I have identified two follow-up issues that need to be addressed separately.

- Separate SASL communication for inter tablets and clients
- Special character escaping

I will follow up with issues and PR for each.

### [DONE] Separating SASL Communication

For this to work we would need to prefix the JAAS contents with `listener` name, for example:

```
internal.FlussServer {
    ...
}
```

But this does not work for the client, as [on this line](https://github.com/apache/fluss/blob/main/fluss-common/src/main/java/org/apache/fluss/security/auth/sasl/jaas/JaasContext.java#L163) the client listener name is hard coded as `null`. So we will have to fix the core also and then enable separate SASL for inter tablets and clients in helm charts.

### Special Character for SASL Usernames and Passwords

This is also indeed an issue, which requires core change for SASL client authentication.

Without escaping we would have something like below `jaas.conf` file:

```
root@coordinator-server-0:/opt/fluss# cat /etc/fluss/conf/jaas.conf

FlussServer {
   org.apache.fluss.security.auth.sasl.plain.PlainLoginModule required
   user_admin="pa$$wo\rd!@#%&""
   user_user1="5zFqhXGY0FgXzxUVpzRoA";
};

FlussClient {
   org.apache.fluss.security.auth.sasl.plain.PlainLoginModule required
   username="admin"
   password="pa$$wo\rd!@#%&"";
};
```

This fails on server with configuration error.

It should be correctly escaped as below:

```
FlussServer {
   org.apache.fluss.security.auth.sasl.plain.PlainLoginModule required
   user_admin="pa$$wo\\rd!@#%&\""
   user_user1="5zFqhXGY0FgXzxUVpzRoA";
};

FlussClient {
   org.apache.fluss.security.auth.sasl.plain.PlainLoginModule required
   username="admin"
   password="pa$$wo\\rd!@#%&\"";
};
```

But this again causes issues on client side since the [SaslClientAuthenticator](https://github.com/apache/fluss/blob/main/fluss-common/src/main/java/org/apache/fluss/security/auth/sasl/authenticator/SaslClientAuthenticator.java#L64) does not escape the user provided username and password.

This is the failing test for `SaslAuthenticationITCase`:

```java
@Test
void testSpecialCharactersForPassword() throws Exception {
    final String specialPassword = "pa$$wo\\rd!@#%&\"";
    final Configuration clientConfig = new Configuration();
    clientConfig.setString("client.security.protocol", "sasl");
    clientConfig.setString("client.security.sasl.username", "admin");
    clientConfig.setString("client.security.sasl.password", specialPassword);
    testAuthentication(clientConfig, getDefaultServerConfig());
}
```

Since both of these points require changes to Fluss core packages, let's address them separately.

morazow:
Discussed offline with @affo, we don't have to worry about the prefixing the `FlussClient` block in JAAS configuration file, since there would be only one client for inter tablet communication 🤝

helm/templates/secret-jaas-config.yaml
morazow: Added 👍

helm/values.yaml
morazow: Generating one for the internal communications 🤝

swuferhong:
LGTM. @loserwang1024 Could you take a look at this PR?

website/docs/install-deploy/deploying-with-helm.md
```diff
@@ -245,16 +253,69 @@ The chart automatically configures listeners for internal cluster communication
...
+Set these values for additional configurations:
+
+```yaml
+listeners:
+  internal:
+    protocol: SASL
+    port: 9123
+    security:
+      mechanism: PLAIN
+      users:
```
loserwang1024: This is a bit confusing. Like Kafka, Fluss SASL authentication also does not require each listener to have unique usernames and passwords. Multiple listeners can share the SASL/PLAIN mechanism, but your example seems to suggest that client and internal listeners can enforce account isolation. How does this work?

helm/templates/_sasl.tpl
```diff
@@ -0,0 +1,155 @@
...
+{{- end -}}
+{{- end -}}
+
+{{/*
+Return true if listener protocol is SASL
+Usage: include "fluss.listener.sasl.enabled" (dict "root" . "listener" "internal")
+*/}}
+{{- define "fluss.listener.sasl.enabled" -}}
+{{- $listener := index .root.Values.listeners .listener -}}
+{{- if and $listener $listener.protocol (regexFind "SASL" (upper $listener.protocol)) -}}
```
affo:
```suggestion
{{- if and $listener $listener.protocol (contains "SASL" (upper $listener.protocol)) -}}
```

I think contains would be good as well 🤔

helm/templates/_sasl.tpl
```diff
@@ -0,0 +1,155 @@
...
+# limitations under the License.
+#
+
+{{/*
+Return true if SASL is configured in any of the listener protocols
+*/}}
+{{- define "fluss.sasl.enabled" -}}
+{{- $enabled := false -}}
+{{- range $id, $l := .Values.listeners -}}
+  {{- if and (not $enabled) (regexFind "SASL" (upper $l.protocol)) -}}
```
affo:
```suggestion
  {{- if and (not $enabled) (contains "SASL" (upper $l.protocol)) -}}
```

website/docs/install-deploy/deploying-with-helm.md
morazow: Hello @loserwang1024 👋 ,

Yes, indeed. It works because of the SASL listener prefix, [JaasContext#126](https://github.com/apache/fluss/blob/main/fluss-common/src/main/java/org/apache/fluss/security/auth/sasl/jaas/JaasContext.java#L126).

For example, with both internal and client SASL enabled, the following jaas file will be created:

_$ root@coordinator-server-0:/opt/fluss# cat /etc/fluss/conf/jaas.conf_

```
internal.FlussServer {
   org.apache.fluss.security.auth.sasl.plain.PlainLoginModule required
   "user_internal-hfpjhuc1gtbq"="yOxDGiWv4QD0XILF2KBlXoStM9ITavux";
};

client.FlussServer {
   org.apache.fluss.security.auth.sasl.plain.PlainLoginModule required
   "user_mtoraz"="passWA";
};

FlussClient {
   org.apache.fluss.security.auth.sasl.plain.PlainLoginModule required
   username="internal-hfpjhuc1gtbq"
   password="yOxDGiWv4QD0XILF2KBlXoStM9ITavux";
};

ZookeeperClient {
   org.apache.fluss.shaded.zookeeper3.org.apache.zookeeper.server.auth.DigestLoginModule required
   username="zk-admin"
   password="zk-password";
};
```

As you can see the FlussServers are prefixed for internal and client. The FlussClient doesn't have any prefix, but it is fine since here it acts as client only for the internal communications.

The to run the client jobs:

```
FLUSS_TEST_BOOTSTRAP=coordinator-server-0.coordinator-server-hs.fluss.svc.cluster.local:9124 \
FLUSS_TEST_USERNAME=mtoraz \
FLUSS_TEST_PASSWORD=passWA \
java -jar fluss-smoke-test-sasl.jar test
```

users can use the client username and password.

website/docs/install-deploy/deploying-with-helm.md
```diff
@@ -245,16 +253,69 @@ The chart automatically configures listeners for internal cluster communication
...
+| `SASL`      | Yes            | No                 |
+
+By default, the `PLAINTEXT` protocol is used.
+
+The SASL authentication will be enabled if any of the listener protocols is using `SASL`.
+
+Set these values for additional configurations:
+
+```yaml
+listeners:
```
affo: Before putting the focus on punctual comments, I would stop one second and think about the design of this.

I really like placing the security bit directly into the listener config, e.g.: `listeners.internal.security`.

However, I don't think that this schema:

```yaml
internal:
  protocol: SASL
  security:
    mechanism: PLAIN
     users: []
```

Is actually the best way to proceed, given that different security protocols are going to require different keys to be provided. Just as an example, for `SASL/OAUTHBEARER`, users and password won't fit.

Given that every supported protocol will probably have different inputs, why not:

```yaml
internal:
  security:
    saslPlain:
      ssl: false
      users: []
    saslScram: {}
    saslOAuthBearer: {}
```

So that every protocol can specify the required inputs without further validation?

@morazow @loserwang1024 what do you think?

morazow: Hey @affo ,

Yeah could be simpler, but I think will again complicate in other parts. 

For example, for setting security map protocol or client security mechanism would require complex parsing logic:

```
echo "security.protocol.map: INTERNAL:<?>,CLIENT:<?>" >> $FLUSS_HOME/conf/server.yaml && \
```
Here for the question marks, we directly reference the `Values.listener.<internal|client>.protocol`, no we would need parsing logic to determine the protocol.

But let's think for user friendly and semi future proof design.

helm/templates/_security.tpl
```diff
@@ -0,0 +1,149 @@
...
+{
+  Values: .Values,
+  listener: "client"
+}
+
+Inside the called helper, in this case, 'fluss.security.listener.mechanism', it is accessed as:
+- .Values -> the chart values
+- .listener -> "client"
+
+The reason for this is that include can only pass one argument, so dict is a standard way to pass multiple named inputs.
```
affo: I don't think this explanation is relevant

helm/templates/_security.tpl
```diff
@@ -0,0 +1,149 @@
...
+*/}}
+{{- define "fluss.security.listener.mechanism" -}}
+{{- $listener := index .Values.security .listener | default (dict) -}}
```
affo: I would argue we don't need to pass `Values` in as this function is only used in this chart and does not need to be re-used.

You may access it directly via `$.Values` or `.Values`.

helm/templates/_security.tpl
```diff
@@ -0,0 +1,149 @@
...
+{{- $sasl := $listener.sasl | default (dict) -}}
+{{- $mechanism := lower (default "" $sasl.mechanism) -}}
+{{- if not (has $mechanism (list "none" "plain")) -}}
```
affo: I would also accept an empty mechanism as `none`

helm/templates/_security.tpl
```diff
@@ -0,0 +1,149 @@
...
+Renders security configuration lines that are appended to the server.yaml file.
+Usage:
+  include "fluss.security.renderSecurityOptions" .
+*/}}
+{{- define "fluss.security.renderSecurityOptions" -}}
+{{- $internalProtocol := include "fluss.security.listener.protocol" (dict "Values" .Values "listener" "internal") | trim -}}
+{{- $enabledMechanisms := include "fluss.security.sasl.enabledMechanisms" . | trim }}
+{{- $internalClientMechanism := include "fluss.security.listener.mechanism" (dict "Values" .Values "listener" "internal") | upper }}
+
+{{- if (include "fluss.security.sasl.enabled" .) }}
```
affo: What if we do something similar to metrics here and render to dict to yaml directly in the configmap?
It would seem to me way more readable 🤝

helm/templates/_security.tpl
```diff
@@ -0,0 +1,149 @@
...
+echo "security.sasl.enabled.mechanisms: {{ $enabledMechanisms }}" >> $FLUSS_HOME/conf/server.yaml && \
+{{- if eq $internalProtocol "SASL" }}
+echo "client.security.protocol: SASL" >> $FLUSS_HOME/conf/server.yaml && \
+echo "client.security.sasl.mechanism: {{ $internalClientMechanism }}" >> $FLUSS_HOME/conf/server.yaml && \
+{{- end }}
+
+{{- if (include "fluss.security.sasl.plain.enabled" .) }}
+export FLUSS_ENV_JAVA_OPTS="-Djava.security.auth.login.config=/etc/fluss/conf/jaas.conf ${FLUSS_ENV_JAVA_OPTS}" && \
```
affo: also this one, it would be nice to have it under `env` in the sts templates

helm/templates/secret-jaas-config.yaml
```diff
@@ -0,0 +1,50 @@
...
+apiVersion: v1
+kind: Secret
+metadata:
+  name: {{ include "fluss.fullname" . }}-sasl-jaas-config
+  labels:
+    {{- include "fluss.labels" . | nindent 4 }}
+type: Opaque
+stringData:
+  jaas.conf: |
+{{- if eq (include "fluss.security.listener.mechanism" (dict "Values" .Values "listener" "internal")) "plain" }}
```
affo: Here I would like to have variable definitions at the top, e.g.:

```
{{- $internalMechanism := include "fluss.security.listener.mechanism" (dict "Values" .Values "listener" "internal") | trim -}}
```

helm/templates/secret-jaas-config.yaml
```diff
@@ -0,0 +1,50 @@
...
+{{- end }}
+{{- if eq (include "fluss.security.listener.mechanism" (dict "Values" .Values "listener" "client")) "plain" }}
+    client.FlussServer {
+       org.apache.fluss.security.auth.sasl.plain.PlainLoginModule required
+{{- range .Values.security.client.sasl.plain.users | default (list) }}
+       user_{{ .username }}="{{ .password }}"
+{{- end }};
+    };
+{{- end }}
+{{- if eq (include "fluss.security.listener.mechanism" (dict "Values" .Values "listener" "internal")) "plain" }}
```
affo: Why this block is not rendered together with the server side on top?

helm/templates/sts-coordinator.yaml
```diff
@@ -16,6 +16,9 @@
 # limitations under the License.
 #
 
+{{ include "fluss.security.sasl.validateClientPlainUsers" . }}
```
affo: why is this happening twice?

It was also happening in the secret

helm/templates/sts-coordinator.yaml
```diff
@@ -95,12 +100,17 @@ spec:
             tcpSocket:
               port: {{ .Values.listeners.client.port }}
           resources:
-            {{- toYaml .Values.resources.tabletServer | nindent 12 }}
+            {{- toYaml .Values.resources.coordinatorServer | nindent 12 }}
```
affo: I think this line may be offloaded to a separate PR

helm/templates/sts-coordinator.yaml
```diff
@@ -109,6 +119,11 @@ spec:
         - name: data
           emptyDir: {}
         {{- end }}
+        {{- if (include "fluss.security.sasl.plain.enabled" .) }}
+        - name: sasl-config
+          secret:
+            secretName: {{ include "fluss.fullname" . }}-sasl-jaas-config
```
affo: I suggest to template this and re-use across files.

helm/templates/_security.tpl
```diff
@@ -0,0 +1,149 @@
...
+{{- join "," $mechanisms -}}
+{{- end -}}
+
+{{/*
+Validates that the client PLAIN mechanism block contains the required users.
+
+Usage:
+  include "fluss.security.sasl.validateClientPlainUsers" .
+*/}}
+{{- define "fluss.security.sasl.validateClientPlainUsers" -}}
```
affo, line 104: I would suggest to also check here if the internal username/password did change from the default one.
If not, you can template it under `NOTES.txt` (see https://helm.sh/docs/chart_template_guide/notes_files/) to raise a warning to the user.

In general, I suggest a pattern similar to Bitnami's:

They perform all validation in a single place in _helpers: https://github.com/bitnami/charts/blob/main/bitnami/zookeeper/templates/_helpers.tpl#L294

They fail and output in NOTES.txt: https://github.com/bitnami/charts/blob/main/bitnami/zookeeper/templates/NOTES.txt#L79

In this way you are sure that validation always runs and that the output is entirely visible to the user.

affo (changes requested):
Almost there, now looks way cleaner

helm/templates/_security.tpl
morazow: I think this doesn't work, I remember checking it before:

```
included helper when you pass a custom argument:

  - `.Values`: `.` becomes whatever you passed to include. If you pass (dict "listener" "internal"), then . is that dict, and .Values is nil.
  - `$.Values`: `$` inside a define block is the same as . at the template's top scope, which is still whatever was passed to include. It does NOT refer to the caller's $.

  This is a Helm/Go template limitation: include can only pass one argument, and that argument becomes both . and $ inside the called template.

  So to have both .Values and .listener available, we must either:
  1. (dict "Values" .Values "listener" "internal") — original explicit approach
  2. (merge (dict "listener" "internal") .) — inherits .Values from root context
```

I don't think merge it easier to understand. But I am going to remove the comment, used it to explain this situation.

helm/templates/secret-jaas-config.yaml
morazow: good catch thank you, I think it was copy paste error, fixing now 🤝

helm/templates/sts-coordinator.yaml
```diff
@@ -101,6 +105,7 @@ spec:
               mountPath: /opt/conf
             - name: data
               mountPath: /tmp/fluss/data
+            {{- include "fluss.security.sasl.volumeMountPath" . | nindent 12 }}
```
morazow: Not sure if I like this change, seems it became hard to understand since structure also changed by including the template.

Let me know what you think

helm/templates/_security.tpl
affo: Definitely true! Thanks for the explanation!

However, it is irrelevant to put the explanation in the code as a comment as this is standard Go template 🤝 

I think it would be beneficial not to name `.Values` as `Values` 😅 
That is kind of confusing because it really feels like you are accessing the root values directly in the function and passing an extra argument for the sake of it🤝 

Maybe better to rename to `context`? E.g.:

```
include "fluss.security.listener.mechanism" (dict "context" .Values "listenerName" "client")
```

helm/templates/sts-coordinator.yaml
affo: Ah no, I also don't like it 😅 

That's not what I meant, you could simply template:

```
{{- define "fluss.security.sasl.configName" -}}
{{ include "fluss.fullname" . }}-sasl-jaas-config
{{- end -}}
```

instead of:

```
{{- define "fluss.security.sasl.volumeMountName" -}}
{{- if (include "fluss.security.sasl.plain.enabled" .) }}
- name: sasl-config
  secret:
    secretName: {{ include "fluss.fullname" . }}-sasl-jaas-config
{{- end }}
{{- end -}}
```

It is a common pattern to template the names, so you are sure that using `{{- include "fluss.security.sasl.configName" . -}}` in all the places you reference it you are using the same name, and a change propagates everywhere.

helm/templates/_security.tpl
morazow: Yeah definitely confusing, for it was 😅 That is why I added comment.

Changed now to `context` maybe easier to understand 🤞

helm/templates/sts-coordinator.yaml
morazow: Doing it now 🤝

helm/templates/secret-jaas-config.yaml
```diff
@@ -0,0 +1,49 @@
+{{- if (include "fluss.security.sasl.plain.enabled" .) -}}
```
affo: probably this should go after the license

helm/templates/_security.tpl
```diff
@@ -0,0 +1,131 @@
...
+Allowed mechanism values: 'none', 'plain'
+Usage:
+  include "fluss.security.listener.mechanism" (dict "context" .Values "listener" "client")
+*/}}
+{{- define "fluss.security.listener.mechanism" -}}
+{{- $listener := index .context.security .listener | default (dict) -}}
+{{- $sasl := $listener.sasl | default (dict) -}}
+{{- $mechanism := lower (default "none" $sasl.mechanism) -}}
+{{- if and (ne $mechanism "") (not (has $mechanism (list "none" "plain"))) -}}
+{{- fail (printf "security.%s.sasl.mechanism must be one of: none, plain" .listener) -}}
```
affo: as for metrics, how about an empty mechanism?

affo: this fail can also be part of the validation logic in NOTES.txt, don't you think?

affo: I think all checks would become way more natural and also we should not account for the case of an empty mechanism

helm/templates/NOTES.txt
```diff
@@ -0,0 +1,36 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+CHART NAME: {{ .Chart.Name }}
+CHART VERSION: {{ .Chart.Version }}
+APP VERSION: {{ .Chart.AppVersion }}
+
+{{- if (include "fluss.security.sasl.enabled" .) }}
```
affo: Perfect!

however, I think that Bitnami's pattern is even more solid.

These helper may be under `_security.tpl` names, for example, `fluss.security.sasl.warnInternalUser`. Then you may have something similar to Bitnami's https://github.com/bitnami/charts/blob/main/bitnami/zookeeper/templates/_helpers.tpl#L294:

```yaml
{{/*
Compile all warnings into a single message.
*/}}
{{- define "fluss.security.validateValues" -}}

{{- $errMessages := list -}}
{{- $errmessages := append $errMessages (include "fluss.security.sasl.validateClientPlainUsers" .) -}}

{{- $errMessages := without $errMessages "" -}}
{{- $errMessage := join "\n" $errMessages -}}

{{- $warnMessages := list -}}
{{- $warnmessages := append $warnMessages (include "fluss.security.sasl.warnInternalUser" .) -}}

{{- $warnMessages := without $warnMessages "" -}}
{{- $warnMessage := join "\n" $warnMessages -}}

{{- if $warnMessage -}}
{{-   printf "\nVALUES WARNING:\n%s" $warnMessage -}}
{{- end -}}

{{- if $errMessage -}}
{{-   printf "\nVALUES VALIDATION:\n%s" $errMessage | fail -}}
{{- end -}}

{{- end -}}
```

helm/templates/_security.tpl
affo: It is weird to see the options set here 🤔 

Even if it is a big block, don't you think it would be better to have that directly in the configmap?
For example:

```yaml
server.yaml: |
    {{- range $key, $val := .Values.configurationOverrides }}
    {{ $key }}: {{ tpl (printf "%v" $val) $ }}
    {{- end }}

    ### Security

    {{- $internalProtocol := include "fluss.security.listener.protocol" (dict "context" .Values "listener" "internal") | trim -}}
    {{- $clientProtocol := include "fluss.security.listener.protocol" (dict "context" .Values "listener" "client") | trim -}}
    {{- $enabledMechanisms := include "fluss.security.sasl.enabledMechanisms" . | trim -}}
    {{- $internalMechanism := include "fluss.security.listener.mechanism" (dict "context" .Values "listener" "internal") -}}
    {{- if (include "fluss.security.sasl.enabled" .) }}
    security.protocol.map: INTERNAL:{{ $internalProtocol }},CLIENT:{{ $clientProtocol }}
    security.sasl.enabled.mechanisms: {{ $enabledMechanisms }}
    {{- if ne $internalMechanism "none" }}
    client.security.protocol: SASL
    client.security.sasl.mechanism: {{ upper $internalMechanism }}
    {{- end }}
    {{- end }}
    {{- end -}}
```

I think this would drastically enhance clarity as one clearly sees how the ConfigMap is templated instead of searching around helpers.

helm/templates/_security.tpl
```diff
@@ -0,0 +1,173 @@
...
+  include "fluss.security.sasl.validateMechanisms" .
+*/}}
+{{- define "fluss.security.sasl.validateMechanisms" -}}
+{{- $allowedMechanisms := list "" "plain" -}}
+{{- range $listener := list "internal" "client" -}}
+  {{- $listenerValues := index $.Values.security $listener | default (dict) -}}
+  {{- $sasl := $listenerValues.sasl | default (dict) -}}
+  {{- $mechanism := lower (default "" $sasl.mechanism) -}}
+  {{- if not (has $mechanism $allowedMechanisms) -}}
+    {{- printf "security.%s.sasl.mechanism must be one of: empty, plain" $listener -}}
```
affo:
```suggestion
    {{- printf "security.%s.sasl.mechanism must be empty or: plain" $listener -}}
```

helm/templates/_security.tpl
```diff
@@ -0,0 +1,173 @@
...
+{{/*
+Returns a warning if the internal SASL user is using default credentials.
+Usage:
+  include "fluss.security.sasl.warnInternalUser" .
+*/}}
+{{- define "fluss.security.sasl.warnInternalUser" -}}
+{{- if (include "fluss.security.sasl.enabled" .) -}}
+  {{- $internalMechanism := include "fluss.security.listener.mechanism" (dict "context" .Values "listener" "internal") -}}
+  {{- if eq $internalMechanism "plain" -}}
+    {{- if and (eq .Values.security.internal.sasl.plain.username "fluss-internal-user") (eq .Values.security.internal.sasl.plain.password "fluss-internal-password") -}}
```
affo: Maybe:

```
{{- define "fluss.security.sasl.plain.internal.defaultUsername" -}}
{{- printf "fluss-internal-user-%s" .Release.Name -}}
{{- end -}}

{{- define "fluss.security.sasl.plain.internal.defaultPassword" -}}
{{- printf "fluss-internal-password-%s" .Release.Name | sha256sum -}}
{{- end -}}

{{- define "fluss.security.sasl.plain.internal.username" -}}
{{ .Values.security.internal.sasl.plain.username | default (include "fluss.security.sasl.plain.internal.defaultUsername" .) }}
{{- end -}}

{{- define "fluss.security.sasl.plain.internal.password" -}}
{{ .Values.security.internal.sasl.plain.password | default (include "fluss.security.sasl.plain.internal.defaultPassword" .) }}
{{- end -}}
```

Then you can simplify quite havily the check, as the defaults would be empty (you would also don't need to check for emptyness as if empty we use the default:

```suggestion
  {{- if and
    (not .Values.security.internal.sasl.plain.username)
    (not .Values.security.internal.sasl.plain.password)
  -}}
```

The more I think about this, the more I think we may introduce on top of this using an existing secret, but that should be part of another PR imho 🤝

helm/values.yaml
```diff
@@ -58,6 +58,23 @@ listeners:
...
+      plain:
+        users: []
+
+  internal:
+    sasl:
+      # "" | plain
+      mechanism: ""
+      plain:
+        username: "fluss-internal-user"
+        password: "fluss-internal-password"
```
affo:
```suggestion
        username: ""
        password: ""
```
As per the reasoning above this should now default to empty

morazow: Done 🤝

loserwang1024 (approved):
LGTM

PR #2510, opened 2026-01-29
[Connector] Supports reporting backlog information

### LINKED ISSUE #2509: Support report backlog information to Flink
### Search before asking

### Motivation

In [FLIP-309](https://cwiki.apache.org/confluence/display/FLINK/FLIP-309%3A+Support+using+larger+checkpointing+interval+when+source+is+processing+backlog#FLIP309:Supportusinglargercheckpointingintervalwhensourceisprocessingbacklog-ProposedChanges), Flink introduced the concept of **ProcessingBacklog** to indicate whether a record should be processed in a **low-latency** or **high-throughput** manner. **ProcessingBacklog** can be used to dynamically adjust a job’s checkpoint interval at runtime.

The **Fluss Source** can set **ProcessingBacklog** by marking offsets at the starting position to identify **backlog** data and report this signal to Flink, enabling Flink to perform relevant optimizations based on it.

### Proposed Changes

- **Add backlog boundary tracking in `FlinkSourceEnumerator`**
  - Introduce `backlogOffsets: Map<TableBucket, Long>` to store the backlog boundary offset per bucket.
  - Add/extend `resetBacklog(...)` to initialize splits and persist the backlog boundary offsets.
  - Update `initPartitionedSplits()` to pass `initializeAtBeginning` / backlog boundary offset information into the created splits.
  - Update `handleSourceEvent(...)` to handle a new `BacklogFinishEvent` and update the enumerator’s backlog tracking state.

- **Propagate backlog boundary offset via split metadata**
  - Extend split types to carry backlog boundary offset:
    - `LogSplit`: add `backlogOffset`
    - `HybridSnapshotLogSplit`: add `backlogOffset`

- **Introduce a new source event for backlog completion**
  - Add `BacklogFinishEvent extends SourceEvent` to signal that a specific `TableBucket` has finished consuming backlog (i.e., reader has reached/passed the backlog boundary offset).

- **Report backlog completion from `FlinkSourceSplitReader`**
  - Add new fields:
    - `backlogMarkedOffsets: Map<TableBucket, Long>`: backlog boundary offsets per bucket (from assigned splits).
    - `onlySnapshotBuckets: Set<TableBucket>`: buckets with snapshot-only data (no log backlog to track).
    - `backlogEventSentTbls: Set<TableBucket>`: dedup to ensure `BacklogFinishEvent` is sent once per bucket.
    - `context: SourceReaderContext`: used to send `BacklogFinishEvent` to the coordinator.
  - Update reader logic to detect backlog completion and emit events:
    - `fetch(...)`: hook backlog completion reporting when a bucket completes backlog.
    - `subscribeLog(...)` and `forLogRecords(...)`: check current log offsets against `backlogMarkedOffsets`; once the boundary is reached, send `BacklogFinishEvent` (deduplicated).

### Anything else?

_No response_

### Willingness to contribute

xishuaidelin:
### Purpose

Linked issue: close #2509 

### Brief change log

1. Added support for reporting backlog information in the Flink connector.
2. `FlinkSourceEnumerator` now records the latest offsets of all buckets as backlog boundaries on startup.
3. Introduced `BacklogFinishEvent`: when `FlinkSourceSplitReader` consumes up to the backlog boundary, it sends this event to the Enumerator.
4. `FlinkSourceEnumerator` maintains `hasBacklogTbls` state and sets `isProcessingBacklog` to `false` once all relevant buckets have finished processing backlog data.

### Tests

- **FlinkSourceEnumeratorTest**
  - `testRecordBacklogBoundaryOffsets`: Verifies that the Enumerator correctly records backlog boundary offsets on startup.
  - `testProcessingBacklog`: Verifies logic for handling `BacklogFinishEvent` and updating the backlog status (`isProcessingBacklog`).

- **FlinkSourceSplitReaderTest**
  - `testSendBacklogFinishEventForHybridSnapshotLogSplit`: Verifies that `BacklogFinishEvent` is sent when a `HybridSnapshotLogSplit` finishes the backlog.
  - `testSendBacklogFinishEventForLogSplit`: Verifies that `BacklogFinishEvent` is sent when a `LogSplit` finishes the backlog.

- **SourceSplitSerializerTest**
  - `testHybridSnapshotLogSplitSerdeWithBacklogOffset`: Tests serialization/deserialization for `HybridSnapshotLogSplit` with backlog offsets.
  - `testLogSplitSerdeWithBacklogOffset`: Tests serialization/deserialization for `LogSplit` with backlog offsets.

### DIFF
--- a/fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/source/enumerator/FlinkSourceEnumerator.java
+++ b/fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/source/enumerator/FlinkSourceEnumerator.java
@@ -189,6 +189,7 @@ public class FlinkSourceEnumerator
 
     @Nullable private final LakeSource<LakeSplit> lakeSource;
 
+    // Record the backlog offset of buckets which have backlog data
     private final Map<TableBucket, Long> bucketsWithBacklogOffset;
 
     private final boolean enableBacklogReporting;
@@ -948,9 +949,7 @@ public class FlinkSourceEnumerator
         if (!incrementalAssignment.isEmpty()) {
             LOG.info("Assigning splits to readers {}", incrementalAssignment);
             context.assignSplits(new SplitsAssignment<>(incrementalAssignment));
-            if (enableBacklogReporting) {
-                sendMarkedBacklogOffsetEvents(incrementalAssignment);
-            }
+            sendMarkedBacklogOffsetEvents(incrementalAssignment);
         }
 
         if (noMoreNewSplits) {
@@ -968,6 +967,7 @@ public class FlinkSourceEnumerator
     private void sendMarkedBacklogOffsetEvents(
             Map<Integer, List<SourceSplitBase>> incrementalAssignment) {
         if (bucketsWithBacklogOffset.isEmpty()) {
+            // no recorded buckets if do not report backlog
             return;
         }
         for (Map.Entry<Integer, List<SourceSplitBase>> entry : incrementalAssignment.entrySet()) {
--- a/fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/source/reader/FlinkSourceSplitReader.java
+++ b/fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/source/reader/FlinkSourceSplitReader.java
@@ -114,29 +114,11 @@ public class FlinkSourceSplitReader implements SplitReader<RecordAndPos, SourceS
 
     private final SourceReaderContext context;
     private final Map<TableBucket, Long> markedBacklogOffsets;
-    private final Set<TableBucket> backlogEventSentTbls = new HashSet<>();
     private final Set<String> emptyLogSplits;
     // track split IDs corresponding to removed partitions
     private final Set<String> removedSplits = new HashSet<>();
     // Set to collect table buckets that are unsubscribed.
     private Set<TableBucket> unsubscribedTableBuckets = new HashSet<>();
-
-    public FlinkSourceSplitReader(
-            Configuration flussConf,
-            TablePath tablePath,
-            RowType sourceOutputType,
-            @Nullable int[] projectedFields,
-            FlinkSourceReaderMetrics flinkSourceReaderMetrics,
-            @Nullable LakeSource<LakeSplit> lakeSource) {
-        this(
-                null,
-                flussConf,
-                tablePath,
-                sourceOutputType,
-                projectedFields,
-                flinkSourceReaderMetrics,
-                lakeSource);
-    }
 
     public FlinkSourceSplitReader(
             SourceReaderContext context,
@@ -195,7 +177,7 @@ public class FlinkSourceSplitReader implements SplitReader<RecordAndPos, SourceS
                 // empty log splits mean there is no log data to consume, so the backlog
                 // phase is already complete (including snapshot-only buckets that have a
                 // marked backlog offset).
-                sendFinishedBacklogEvent(new HashSet<>(subscribedBuckets.keySet()));
+                sendFinishedBacklogEvent(new HashSet<>(markedBacklogOffsets.keySet()));
                 return records;
             } else {
                 // if not subscribe any buckets, just return empty records
@@ -454,7 +436,7 @@ public class FlinkSourceSplitReader implements SplitReader<RecordAndPos, SourceS
 
         Map<String, CloseableIterator<RecordAndPos>> splitRecords = new HashMap<>();
         Map<TableBucket, Long> stoppingOffsets = new HashMap<>();
-        Set<TableBucket> backlogFinishedTbl = new HashSet<>();
+        Set<TableBucket> finishedBacklogTbl = new HashSet<>();
         Set<String> finishedSplits = new HashSet<>();
         Map<TableBucket, String> splitIdByTableBucket = new HashMap<>();
         List<TableBucket> tableScanBuckets = new ArrayList<>(scanRecords.buckets().size());
@@ -482,9 +464,9 @@ public class FlinkSourceSplitReader implements SplitReader<RecordAndPos, SourceS
                     stoppingOffsets.put(scanBucket, stoppingOffset);
                     finishedSplits.add(splitId);
                 }
-                if (!backlogEventSentTbls.contains(scanBucket)
+                if (markedBacklogOffsets.containsKey(scanBucket)
                         && lastRecord.logOffset() >= backlogMarkedOffset - 1) {
-                    backlogFinishedTbl.add(scanBucket);
+                    finishedBacklogTbl.add(scanBucket);
                 }
             }
             splitRecords.put(splitId, toRecordAndPos(bucketScanRecords.iterator()));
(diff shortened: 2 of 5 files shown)


fluss-flink/fluss-flink-common/src/test/java/org/apache/fluss/flink/source/enumerator/FlinkSourceEnumeratorTest.java
```diff
@@ -759,6 +760,52 @@ void testPartitionsExpiredInFlussButExistInLake(
...
+            BacklogFinishEvent event1 = new BacklogFinishEvent(bucket1);
+            BacklogFinishEvent event2 = new BacklogFinishEvent(bucket2);
+
+            enumerator.handleSourceEvent(0, event0);
+            enumerator.handleSourceEvent(1, event1);
+            enumerator.handleSourceEvent(2, event2);
+
+            assertThat(context.isProcessingBacklogCalled()).isFalse();
+        }
+    }
```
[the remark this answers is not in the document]
xishuaidelin: Already added the actual data.

xishuaidelin:
Hi @wuchong , could you please take a look at this PR?

wuchong:
I will review it tomorrow. Sorry for the late response.

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/source/event/BacklogFinishEvent.java
```diff
@@ -0,0 +1,65 @@
...
+import org.apache.flink.api.connector.source.SourceEvent;
+
+import java.util.Objects;
+
+/**
+ * Event to signal that the backlog processing for a specific table bucket has been completed. This
+ * event is sent from the source reader to the source enumerator to indicate that all historical
+ * data (backlog) in the specified bucket has been processed.
+ */
+public class BacklogFinishEvent implements SourceEvent {
```
wuchong: nit: Rename to `FinishedBacklogEvent` to align other finished SourceEvent.

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/source/reader/FlinkSourceSplitReader.java
```diff
@@ -122,6 +128,25 @@ public FlinkSourceSplitReader(
             @Nullable int[] projectedFields,
             FlinkSourceReaderMetrics flinkSourceReaderMetrics,
             @Nullable LakeSource<LakeSplit> lakeSource) {
+        this(
+                null,
```
wuchong: Refactor all usages of this constructor to explicitly pass the `SourceReaderContext` instead of `null`. Relying on a null value introduces a significant risk of potential NullPointerExceptions.

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/source/reader/FlinkSourceSplitReader.java
```diff
@@ -109,6 +111,10 @@ public class FlinkSourceSplitReader implements SplitReader<RecordAndPos, SourceS
     private final Map<TableBucket, Long> stoppingOffsets;
     private LakeSplitReaderGenerator lakeSplitReaderGenerator;
 
+    private final SourceReaderContext context;
+    private final Map<TableBucket, Long> backlogMarkedOffsets;
+    private final Set<TableBucket> onlySnapshotBuckets;
+    private final Set<TableBucket> backlogEventSentTbls = new HashSet<>();
```
wuchong: For partitioned tables, frequent creation and removal of partitions mean that caching all `TableBucket` instances without cleanup will lead to memory leaks. Therefore, we should not rely on `backlogEventSentTbls` to determine whether a `TableBucket` needs to report backlog completion. Instead, we should check for the presence of the `TableBucket` in `backlogMarkedOffsets`; if it is absent, the backlog has been processed.

Additionally, maintaining `onlySnapshotBuckets` is unnecessary, as `backlogMarkedOffsets` provides sufficient state. We can trigger the backlog event within `finishCurrentBoundedSplit()` by verifying whether the backlog offset is less than or equal to `HybridSnapshotLogSplit#logStartingOffset`.

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/source/enumerator/FlinkSourceEnumerator.java
```diff
@@ -182,6 +188,8 @@ public class FlinkSourceEnumerator
 
     @Nullable private final LakeSource<LakeSplit> lakeSource;
 
+    private final Map<TableBucket, Long> hasBacklogTbls;
```
wuchong: Please add Javadoc to this field to clarify the concept of backlog and explain its specific purpose. Additionally, I recommend renaming the field to `bucketsWithBacklogOffset` for improved clarity.

xishuaidelin:
> Thank you @xishuaidelin for your contribution. I have left the following comments for your consideration:
> 
> 1. Introduce a flag named `enableBacklogReporting` in the Source to control backlog behavior. This flag should be disabled when processing log tables, operating in batch mode, or when explicitly turned off via configuration (e.g., `scan.backlog.report.enable`).
> 2. Currently, the system sends a backlog event whenever the source starts, restores, or upgrades from a legacy version. To optimize this, please store a boolean flag `isBacklogProcessed` within the `SourceEnumeratorState`. The backlog should only be processed if this flag is false, regardless of whether the source is newly started or restored. For legacy `SourceEnumeratorState` instances, initialize this flag to true to skip backlog processing.
> 3. Do not record the backlog offset within the split itself. Instead, use a separate event to transmit this information.
> 4. Add an Integration Test (IT) case to verify the backlog processing logic. This test should ensure that events are correctly sent to the downstream operator. You can mock an operator to collect these backlog events, similar to the approach used in `org.apache.flink.test.streaming.runtime.RecordAttributesPropagationITCase.TwoInputOperator`.

Hi, @wuchong , 
Thank you for the detailed review and guidance. I have updated the PR to address all your comments:

Backlog Flag: Introduced the enableBacklogReporting flag in FlinkSource. In FlinkTableSource, it is correctly disabled for log tables, batch mode, or when explicitly turned off via the scan.backlog.report.enable configuration. In FlussSourceBuilder, it is disabled for non-PK tables or when the configuration is off.

State Management: Added the isBacklogProcessed boolean to SourceEnumeratorState. The enumerator only initializes backlog processing when both enableBacklogReporting is true and isBacklogProcessed is false. The serializer (version 3) persists this flag; for legacy states (versions 0–2), it is initialized to true to ensure backward compatibility and skip redundant backlog processing on restore.


Offset Handling: Removed the backlog offset recording from the split itself. The enumerator now tracks backlog boundary offsets in a separate bucketsWithBacklogOffset map and transmits them to readers via MarkedBacklogOffsetEvent upon split assignment. Readers report completion back via FinishedBacklogEvent.

Integration Test: Added BacklogProcessingITCase with three test cases: 
1. testBacklogEventsPropagation (verifies isBacklog=true → false transition), 
2. testBacklogReportingDisabled (verifies no backlog events when disabled), 
3. testBacklogStateRestoredFromSavepoint (verifies no re-entry of backlog processing after restore). 


Could you please take another look when you have time?

PR #2515, opened 2026-01-29
[client] List databases return database summary infos.

### LINKED ISSUE #2514: [client] List databases return database summary infos.
### Search before asking

### Motivation

Currently,  list databases only return database names. That means if user want to get the table number and create time for each database, need to:
1.  listTables for each database 
2. getDatabaseInfo for each database. Though detail info is no need.

### Solution

_No response_

### Anything else?

_No response_

### Willingness to contribute

loserwang1024:
### Purpose

Linked issue: close #2514

### DIFF
--- a/fluss-client/src/main/java/org/apache/fluss/client/admin/Admin.java
+++ b/fluss-client/src/main/java/org/apache/fluss/client/admin/Admin.java
@@ -180,7 +180,18 @@ public interface Admin extends AutoCloseable {
     /** List all databases in fluss cluster asynchronously. */
     CompletableFuture<List<String>> listDatabases();
 
-    /** List all databases' summary information in fluss cluster asynchronously. */
+    /**
+     * List all databases' summary information in fluss cluster asynchronously. The difference
+     * between this method and {@link #listDatabases()} is that this method also include some
+     * summaries for the database, like {@link DatabaseSummary#getCreatedTime()} and {@link
+     * DatabaseSummary#getTableCount()}.
+     *
+     * <p>When interacting older version of fluss cluster which does not support this API, it will
+     * fall back to {@link #listDatabases()} with {@code -1} value for {@link
+     * DatabaseSummary#getCreatedTime()} and {@link DatabaseSummary#getTableCount()}.
+     *
+     * @since 0.9
+     */
     CompletableFuture<List<DatabaseSummary>> listDatabaseSummaries();
 
     /**
--- a/fluss-server/src/main/java/org/apache/fluss/server/zk/ZooKeeperClient.java
+++ b/fluss-server/src/main/java/org/apache/fluss/server/zk/ZooKeeperClient.java
@@ -498,15 +498,24 @@ public class ZooKeeperClient implements AutoCloseable {
         List<ZkCheckExistsResponse> statsInBackground =
                 getStatInBackground(path2DatabaseNamesMap.keySet());
         List<DatabaseSummary> databaseSummaries = new ArrayList<>();
-        for (ZkCheckExistsResponse getStatResponse : statsInBackground) {
-            getStatResponse.maybeThrow();
-            // To decrease the cost, use zk node creation time as the database creation
-            // time rather than create_time in node data.
-            databaseSummaries.add(
-                    new DatabaseSummary(
-                            path2DatabaseNamesMap.get(getStatResponse.getPath()),
-                            getStatResponse.getStat().getCtime(),
-                            getStatResponse.getStat().getNumChildren()));
+        for (ZkCheckExistsResponse response : statsInBackground) {
+            Stat stat = response.getStat();
+            if (!response.hasError() && stat != null) {
+                // To decrease the cost, use zk node creation time as the database creation
+                // time rather than create_time in node data.
+                databaseSummaries.add(
+                        new DatabaseSummary(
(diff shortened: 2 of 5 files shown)


fluss-server/src/main/java/org/apache/fluss/server/zk/ZooKeeperClient.java
```diff
@@ -1275,13 +1290,8 @@ public List<String> getChildren(String path) throws Exception {
...
     /** Gets the data and stat of a given zk node path. */
-    public Optional<Stat> getStat(String path) throws Exception {
-        try {
-            Stat stat = zkClient.checkExists().forPath(path);
-            return Optional.ofNullable(stat);
-        } catch (KeeperException.NoNodeException e) {
-            return Optional.empty();
-        }
+    public Stat getStat(String path) throws Exception {
+        return zkClient.checkExists().forPath(path);
```
wuchong: Why make this change? The previous implementation was clearer and safer. The current one doesn’t properly handle the case where the znode does not exist, which can lead to runtime errors or undefined behavior.

fluss-server/src/main/java/org/apache/fluss/server/zk/ZooKeeperClient.java
```diff
@@ -487,6 +488,20 @@ public List<String> listDatabases() throws Exception {
         return getChildren(DatabasesZNode.path());
     }
 
+    public Optional<DatabaseSummary> getDatabaseSummary(String databaseName) throws Exception {
```
wuchong: The current implementation still triggers **N separate I/O operations**—one for each database—which can significantly increase latency. To mitigate this, we can optimize by performing these requests **asynchronously in the background**.

Specifically, you can:

1. Introduce a new method in `ZooKeeperClient`:   
   ```java
   getStatsInBackground(Collection<String> paths)
   ```  
   This method should wrap the existing `handleRequestInBackground` and use a newly defined `ZkGetStatRequest` type to fetch stats for multiple paths concurrently. Take `getDataInBackground` as an example. 

2. Then, add a new high-level method:  
   ```java
   List<DatabaseSummary> listDatabaseSummaries(Collection<String> databaseNames)
   ```  
   which calls `getStatsInBackground` to retrieve all necessary ZooKeeper stats in parallel and constructs the `DatabaseSummary` objects efficiently.

This approach reduces end-to-end latency by eliminating sequential I/O and leveraging concurrent background requests.

fluss-client/src/main/java/org/apache/fluss/client/admin/Admin.java
```diff
@@ -179,6 +180,9 @@ CompletableFuture<Void> dropDatabase(
     /** List all databases in fluss cluster asynchronously. */
     CompletableFuture<List<String>> listDatabases();
 
+    /** List all databases' summary information in fluss cluster asynchronously. */
+    CompletableFuture<List<DatabaseSummary>> listDatabases(ListDatabaseOption option);
```
wuchong: Rename the method to `listDatabaseSummaries()` without any parameters. If user would like only returning database names, they can use `listDatabases()`.

fluss-rpc/src/main/proto/FlussApi.proto
```diff
@@ -93,9 +93,12 @@ message DatabaseExistsResponse {
 
 // list databases request and response
 message ListDatabasesRequest {
+  optional bool database_name_only = 1;
```
wuchong: For legacy clients, this new field will be missing, which typically causes the boolean to default to false. This creates a risk of backward incompatibility: the server might return the full summary instead of just the names, effectively changing the behavior for old clients.

While we could technically work around this by checking `request.hasDatabaseNameOnly() && !request.isDatabaseNameOnly()`, this logic is counter-intuitive and error-prone.

**Suggestion:** A better design would be to use `bool include_summary` instead of `database_name_only`. So, for legacy clients, the default value (`false`) and empty value (unset this field) are consistent and correctly imply 'do not include summary,' automatically preserving the existing behavior without requiring complex null-checks.

fluss-common/src/main/java/org/apache/fluss/metadata/DatabaseSummary.java
```diff
@@ -0,0 +1,108 @@
...
+ * count, and other summary statistics. It is distinct from {@link DatabaseInfo} which contains
+ * complete database metadata including the {@link DatabaseDescriptor}.
+ *
+ * @since 0.6
+ */
+@PublicEvolving
+public class DatabaseSummary {
+    private final String databaseName;
+    private final @Nullable Long createdTime;
+    private final @Nullable Integer tableCount;
```
wuchong: I suggest making these fields **non-nullable** and using primitive types (`long` and `int`) instead of their boxed counterparts.  

If users need database names, they should call `listDatabases()` directly—rather than fetching database summaries and then discarding them.  

This approach makes the API clearer, more predictable, and deterministic in behavior.

fluss-client/src/main/java/org/apache/fluss/client/admin/Admin.java
loserwang1024: With the ListDatabaseOption option, if we want to support additional capabilities later (such as paginated queries), there's no need to add a new method.

loserwang1024:
@wuchong I have modified this pr

PR #2516, opened 2026-01-29
[server] print snapshot size in snapshot finish log

zuston:
### Purpose

Linked issue: close #xxx

Hope to see the snapshot size to measure its upload performance

### DIFF
--- a/fluss-server/src/main/java/org/apache/fluss/server/kv/snapshot/PeriodicSnapshotManager.java
+++ b/fluss-server/src/main/java/org/apache/fluss/server/kv/snapshot/PeriodicSnapshotManager.java
@@ -244,10 +244,11 @@ public class PeriodicSnapshotManager implements Closeable {
                                             snapshotLocation,
                                             snapshotResult);
                                     LOG.info(
-                                            "TableBucket {} snapshot {} finished successfully, size: {}, cost: {} ms.",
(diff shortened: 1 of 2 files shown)


fluss-server/src/main/java/org/apache/fluss/server/kv/snapshot/PeriodicSnapshotManager.java
```diff
@@ -244,9 +244,10 @@ private void asyncSnapshotPhase(
                                             snapshotLocation,
                                             snapshotResult);
                                     LOG.info(
-                                            "TableBucket {} snapshot {} finished successfully, cost {} ms.",
+                                            "TableBucket {} snapshot {} finished successfully, size: {}, cost: {} ms.",
```
wuchong: How about printing incremental size as well? 

```
full size: {}, incremental size: {}
```

wuchong:
@rionmonster, it's recommended to link a dedicated issue for each pull request, except for trivial changes (such as documentation updates, logging adjustments, or typo fixes), which are straightforward and don’t require discussion.

fluss-server/src/main/java/org/apache/fluss/server/kv/snapshot/PeriodicSnapshotManager.java
zuston: done

PR #2522, opened 2026-01-29
docs: updated Flink SQL Client startup and cluster shutdown instructions

niknegi:
### Brief change log
Updated the getting started instruction for Flink sql client and cluster shutdown instructions


MehulBatra:
LGTM! @niknegi Thanks for fixing it.

PR #2525, opened 2026-01-29
[flink] Basic $binlog read support without pushdown optimizations for pk table

### LINKED ISSUE #978: Add binglog system table support for fluss tables
### Search before asking

### Motivation

If you need to audit the changelog of the table, you can use the binlog system table. In the binlog system table, the update before and the update after will be packed in one row. (Similar to what we have in Paimon)

`SELECT * FROM tablename$binlog;`

```

/*
+------------------+----------------------+-----------------------+
|     rowkind      |       column_0       |       column_1        |
+------------------+----------------------+-----------------------+
|        +I        |       [col_0]        |       [col_1]         |
+------------------+----------------------+-----------------------+
|        +U        | [col_0_ub, col_0_ua] | [col_1_ub, col_1_ua]  |
+------------------+----------------------+-----------------------+
|        -D        |       [col_0]        |       [col_1]         |
+------------------+----------------------+-----------------------+
*/

```
### Solution

_No response_

### Anything else?

_No response_

### Willingness to contribute

MehulBatra:
### Purpose

Linked issue: close #978 

### Brief change log

Read support for virtual binlog table
            
Added:                          
  - BinlogRowConverter — the core stateful converter that buffers -U records and  
  merges them with +U into a single row with before/after nested fields           
  - BinlogDeserializationSchema — thin wrapper over the converter                 
  - BinlogFlinkTableSource — Flink ScanTableSource implementation                 
                                                                                  
Modifed:                                                 
  - TableDescriptor — add BEFORE_COLUMN and AFTER_COLUMN constants                
  - FlinkCatalog — wire $binlog detection, add PK-only validation, build nested   
  ROW schema                                                                      
  - FlinkTableFactory — add binlog source creation                                
                                                                                 

  

### Tests

  FlinkCatalogIT - test get virtualbinlog table
  Unit tests — BinlogRowConverterTest.java 
  Integration tests — BinlogVirtualTableITCase.java

### DIFF
--- a/fluss-flink/fluss-flink-common/src/test/java/org/apache/fluss/flink/catalog/FlinkCatalogITCase.java
+++ b/fluss-flink/fluss-flink-common/src/test/java/org/apache/fluss/flink/catalog/FlinkCatalogITCase.java
@@ -953,23 +953,16 @@ abstract class FlinkCatalogITCase {
                 (CatalogTable)
                         catalog.getTable(new ObjectPath(DEFAULT_DB, "pk_table_for_binlog$binlog"));
 
-        // Build expected schema: metadata columns + nested before/after ROW columns
-        org.apache.flink.table.api.DataTypes.Field[] nestedFields =
-                new org.apache.flink.table.api.DataTypes.Field[] {
-                    DataTypes.FIELD("id", DataTypes.INT().notNull()),
-                    DataTypes.FIELD("name", DataTypes.STRING().notNull()),
-                    DataTypes.FIELD("amount", DataTypes.BIGINT())
-                };
-        Schema expectedSchema =
-                Schema.newBuilder()
-                        .column("_change_type", DataTypes.STRING().notNull())
-                        .column("_log_offset", DataTypes.BIGINT().notNull())
-                        .column("_commit_timestamp", DataTypes.TIMESTAMP_LTZ(3).notNull())
-                        .column("before", DataTypes.ROW(nestedFields))
-                        .column("after", DataTypes.ROW(nestedFields))
-                        .build();
-
-        assertThat(binlogTable.getUnresolvedSchema()).isEqualTo(expectedSchema);
+        // use string representation for assertion to simplify the unresolved schema comparison
+        assertThat(binlogTable.getUnresolvedSchema().toString())
+                .isEqualTo(
+                        "(\n"
+                                + "  `_change_type` STRING NOT NULL,\n"
+                                + "  `_log_offset` BIGINT NOT NULL,\n"
+                                + "  `_commit_timestamp` TIMESTAMP_LTZ(3) NOT NULL,\n"
+                                + "  `before` [ROW<id INT NOT NULL, name STRING NOT NULL, amount BIGINT>],\n"
+                                + "  `after` [ROW<id INT NOT NULL, name STRING NOT NULL, amount BIGINT>]\n"
+                                + ")");
 
         // Binlog virtual tables have empty partition keys (columns are nested)
         assertThat(binlogTable.getPartitionKeys()).isEmpty();


fluss-flink/fluss-flink-common/src/test/java/org/apache/fluss/flink/catalog/FlinkCatalogITCase.java
```diff
@@ -932,6 +942,65 @@ void testGetChangelogVirtualTable() throws Exception {
...
+
+        // Verify before and after columns are ROW types containing the original columns
+        String schemaString = binlogSchema.toString();
+        assertThat(schemaString).contains("before");
+        assertThat(schemaString).contains("after");
+        assertThat(schemaString).contains("ROW<");
+        // Verify nested columns exist in the ROW type
+        assertThat(schemaString).contains("id INT NOT NULL");
+        assertThat(schemaString).contains("name STRING NOT NULL");
+        assertThat(schemaString).contains("amount BIGINT");
```
wuchong: Could you follow the approach used in `testGetChangelogVirtualTable` and assert against an expected schema directly? Using `contains` for schema validation is less readable and prone to errors.

fluss-flink/fluss-flink-common/src/test/java/org/apache/fluss/flink/source/BinlogVirtualTableITCase.java
```diff
@@ -0,0 +1,359 @@
...
+        CLOCK.advanceTime(Duration.ofMillis(1000));
+        writeRows(conn, tablePath, Arrays.asList(row(1, "Alice")), false);
+
+        List<String> results = collectRowsWithTimeout(rowIter, 1, true);
+        assertThat(results).hasSize(1);
+
+        // SELECT * returns: _change_type, _log_offset, _commit_timestamp, before, after
+        // before is null for INSERT, after contains the row
+        assertThat(results.get(0)).startsWith("+I[I, 0, ");
+        assertThat(results.get(0)).contains("null, +I[1, Alice]");
```
wuchong: Can't we assert a full row string?

fluss-flink/fluss-flink-common/src/test/java/org/apache/fluss/flink/source/BinlogVirtualTableITCase.java
```diff
@@ -0,0 +1,359 @@
...
+        String queryEarliest =
+                "SELECT _change_type, after.id, after.name FROM startup_binlog_test$binlog"
+                        + optionsEarliest;
+        CloseableIterator<Row> rowIterEarliest = tEnv.executeSql(queryEarliest).collect();
+        List<String> earliestResults = collectRowsWithTimeout(rowIterEarliest, 5, true);
+        assertThat(earliestResults).hasSize(5);
+        // All should be INSERT change types
+        for (String result : earliestResults) {
+            assertThat(result).startsWith("+I[I,");
+        }
```
wuchong: Please assert the full row. If you want to make the result deterministic and ordered. You can change the `TABLE_EXEC_RESOURCE_DEFAULT_PARALLELISM` config to `1` in the `before()` method.

fluss-flink/fluss-flink-common/src/test/java/org/apache/fluss/flink/source/BinlogVirtualTableITCase.java
```diff
@@ -0,0 +1,359 @@
...
+        String optionsTimestamp =
+                " /*+ OPTIONS('scan.startup.mode' = 'timestamp', 'scan.startup.timestamp' = '150') */";
+        String queryTimestamp =
+                "SELECT _change_type, after.id, after.name FROM startup_binlog_test$binlog"
+                        + optionsTimestamp;
+        CloseableIterator<Row> rowIterTimestamp = tEnv.executeSql(queryTimestamp).collect();
+        List<String> timestampResults = collectRowsWithTimeout(rowIterTimestamp, 2, true);
+        assertThat(timestampResults).hasSize(2);
+        // Sort results for deterministic assertion
+        Collections.sort(timestampResults);
```
wuchong: ditto. You can change the `TABLE_EXEC_RESOURCE_DEFAULT_PARALLELISM` config to `1` in the `before()` method.

fluss-flink/fluss-flink-common/src/test/java/org/apache/fluss/flink/utils/BinlogRowConverterTest.java
```diff
@@ -0,0 +1,246 @@
...
+        LogRecord record = createLogRecord(ChangeType.DELETE, 300L, 3000L, 3, "Charlie", 1000L);
+
+        RowData result = converter.convert(record);
+
+        assertThat(result).isNotNull();
+        assertThat(result.getRowKind()).isEqualTo(RowKind.INSERT);
+
+        // Verify metadata columns
+        assertThat(result.getString(0)).isEqualTo(StringData.fromString("D"));
+        assertThat(result.getLong(1)).isEqualTo(300L);
```
wuchong, line 133: assert timestamp column as well?

fluss-flink/fluss-flink-common/src/test/java/org/apache/fluss/flink/utils/BinlogRowConverterTest.java
```diff
@@ -0,0 +1,246 @@
...
+        RowData result = converter.convert(record);
+
+        // Verify row kind is always INSERT for virtual tables
+        assertThat(result).isNotNull();
+        assertThat(result.getRowKind()).isEqualTo(RowKind.INSERT);
+
+        // Verify metadata columns
+        assertThat(result.getString(0)).isEqualTo(StringData.fromString("I"));
+        assertThat(result.getLong(1)).isEqualTo(100L); // log offset
+        assertThat(result.getTimestamp(2, 3)).isNotNull(); // commit timestamp
```
wuchong: We can assert the value of commit timestamp

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/source/BinlogFlinkTableSource.java
```diff
@@ -0,0 +1,183 @@
...
+/** A Flink table source for the $binlog virtual table. */
+public class BinlogFlinkTableSource implements ScanTableSource {
+
+    private final TablePath tablePath;
+    private final Configuration flussConfig;
+    // The binlog output type (includes metadata + nested before/after ROW columns)
+    private final org.apache.flink.table.types.logical.RowType binlogOutputType;
+    // The data columns type extracted from the 'before' nested ROW
+    private final org.apache.flink.table.types.logical.RowType dataColumnsType;
+    private final int[] partitionKeyIndexes;
```
wuchong: We don't need this indexes, we only need a `boolean` flag to indicate it is partitioned or not. This can simplify the code.

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/source/BinlogFlinkTableSource.java
```diff
@@ -0,0 +1,183 @@
...
+                }
+
+                @Override
+                public Source<RowData, ?, ?> createSource() {
+                    return source;
+                }
+            };
+        } else {
+            return SourceProvider.of(source);
+        }
```
wuchong: Simply return `SourceProvider.of(source)` should be fine. It returns `isBounded` as `true` when `streaming` is `false`.
We should also update the `ChangelogFlinkTableSource`.

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/catalog/FlinkCatalog.java
```diff
@@ -960,4 +959,112 @@ private Schema buildChangelogSchema(Schema originalSchema) {
...
+            newOptions.putAll(securityConfigs);
+
+            // Store partition key names for the table source to use.
+            // Since binlog schema has nested columns, we can't use Flink's partition key mechanism.
+            List<String> partitionKeys = baseTable.getPartitionKeys();
+            if (!partitionKeys.isEmpty()) {
+                newOptions.put(
+                        FlinkConnectorOptions.BINLOG_PARTITION_KEYS.key(),
+                        String.join(",", partitionKeys));
+            }
```
wuchong: We don't need extract the partition key indexes, we only need to know whether it is a partitioned table.

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/FlinkConnectorOptions.java
```diff
@@ -245,7 +245,16 @@ public class FlinkConnectorOptions {
                     .withDescription(
                             "The serialized base64 bytes of refresh handler of materialized table.");
 
-    // ------------------------------------------------------------------------------------------
+    /**
+     * Internal option used to pass the base table's partition key column names to the binlog table
+     * source. The value is a comma-separated list of column names.
+     */
+    public static final ConfigOption<String> BINLOG_PARTITION_KEYS =
+            ConfigOptions.key("binlog.partition-keys")
```
wuchong: we don't need this

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/catalog/FlinkTableFactory.java
```diff
@@ -261,11 +276,18 @@ private LakeTableFactory mayInitLakeTableFactory() {
...
      *
+     * @param helper the factory helper for option validation
      * @param tableOptions the table options to validate
      */
-    private static void validateSourceOptions(ReadableConfig tableOptions) {
+    private static void validateSourceOptions(
+            FactoryUtil.TableFactoryHelper helper, ReadableConfig tableOptions) {
+        Optional<DataLakeFormat> datalakeFormat = getDatalakeFormat(tableOptions);
+        List<String> prefixesToSkip =
+                new ArrayList<>(Arrays.asList("table.", "client.", "fields.", "binlog."));
```
wuchong: Please rebase to latest `main` branch, we have removed the `helper.validateExcept` check in connector. See https://github.com/apache/fluss/commit/336216cc91903c4852976340d95dae8d300befa4

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/catalog/FlinkTableFactory.java
```diff
@@ -327,4 +349,67 @@ private DynamicTableSource createChangelogTableSource(
...
+                        .get(FlinkConnectorOptions.BINLOG_PARTITION_KEYS.key());
+        int[] partitionKeyIndexes;
+        if (partitionKeysOption != null && !partitionKeysOption.isEmpty()) {
+            String[] partitionKeyNames = partitionKeysOption.split(",");
+            partitionKeyIndexes = new int[partitionKeyNames.length];
+            for (int i = 0; i < partitionKeyNames.length; i++) {
+                partitionKeyIndexes[i] = dataColumnsType.getFieldIndex(partitionKeyNames[i]);
+            }
+        } else {
+            partitionKeyIndexes = new int[0];
```
wuchong: A boolean flag `isPartitionedTable` is enough.

fluss-common/src/main/java/org/apache/fluss/metadata/TableDescriptor.java
```diff
@@ -65,6 +65,10 @@ public final class TableDescriptor implements Serializable {
     public static final String LOG_OFFSET_COLUMN = "_log_offset";
     public static final String COMMIT_TIMESTAMP_COLUMN = "_commit_timestamp";
 
+    // Reserved column names for $binlog virtual table nested row fields
```
wuchong:
```suggestion
    // column names for $binlog virtual table nested row fields
```

They are not "Reserved" names. "Reserved" means it is not allowed users to use.

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/source/deserializer/BinlogDeserializationSchema.java
```diff
@@ -0,0 +1,85 @@
...
+import static org.apache.fluss.flink.utils.FlinkConversions.toFlinkRowType;
+
+/**
+ * A deserialization schema that converts {@link LogRecord} objects to Flink's {@link RowData}
+ * format with nested before/after row structure for the $binlog virtual table.
+ *
+ * <p>This schema is stateful: it buffers UPDATE_BEFORE (-U) records and returns {@code null} for
+ * them. When the subsequent UPDATE_AFTER (+U) record arrives, it merges both into a single binlog
+ * row. The {@link org.apache.fluss.flink.source.emitter.FlinkRecordEmitter} handles null returns by
+ * skipping emission.
```
wuchong, line 39: I noticed a potential data consistency issue in the current implementation of `FlinkRecordEmitter` regarding how it handles offsets for `$binlog` tables.

**The Problem:**
Currently, `FlinkRecordEmitter` advances the log offset in `LogSplitState` by 1 immediately after a record is processed, regardless of whether the record was fully emitted as a complete transaction (e.g., an `UPDATE` pair).

Specifically, for `$binlog` tables, an `UPDATE` operation is split into two logical events: `UPDATE_BEFORE` and `UPDATE_AFTER`. If a checkpoint occurs exactly between these two events and the job subsequently fails:

1. The `LogSplitState` will have already been updated to `offset + 1`.
2. Upon recovery, the source will skip the original log record and start reading from the next offset.
3. The `UPDATE_AFTER` event is lost, or more critically, `BinlogRowConverter#toBinlogRowData` will throw an `IllegalStateException` because `pendingUpdateBefore` is null when it expects to complete an update sequence.

**Suggested Solution:**
We should modify `FlinkRecordEmitter` to only advance the log offset in the state when `deserializationSchema.deserialize` actually returns a non-null record (signaling that the logical processing for that physical record is complete and emitted).

---

### Revised Code Implementation

I suggest refining the logic to ensure the `emitted` flag correctly dictates the state update. Note that I've added a comment to clarify that "emitted" here implies the full logical unit has been sent downstream.

```java
} else if (splitState.isLogSplitState()) {
    // Attempt to process and emit the record. 
    // For $binlog, this returns true only when a complete row (or the final part of a split) is emitted.
    boolean emitted = processAndEmitRecord(recordAndPosition.record(), sourceOutput);
    
    if (emitted) {
        // Only advance the offset in state if the record was successfully emitted.
        // This ensures that if a crash occurs mid-update (between BEFORE and AFTER),
        // the source will re-read the same log offset upon recovery, 
        // allowing the BinlogDeserializationSchema to correctly reconstruct the state.
        splitState.asLogSplitState().setNextOffset(recordAndPosition.record().logOffset() + 1);
    }
}

wuchong (commented):
Thank @MehulBatra for the great pull request. I left some comments inline. 


Besides, regarding the metadata column `_change_type` in the `$changelog` and `$binlog` virtual tables, the current design uses abbreviations like `+I`, `-U`, `+U`, and `-D`.

I re-considered it again, while these align with Flink's internal `RowKind` representation, I think maybe using full descriptive strings instead would be better, such as:

- `insert` (instead of `+I` and `+A`)
- `update_before` (instead of `-U`)
- `update_after` (instead of `+U`)
- `delete` (instead of `-D`)

And for `$binlog` tables 
- `insert` (instead of `I`)
- `update` (instead of `U`)
- `delete` (instead of `D`)

Reasoning:

- **Improved Readability**: Full words are much more intuitive for SQL users, especially those who may not be familiar with Flink's internal row-kind symbols.

- **Better User Experience**: It makes the audit and incremental ETL queries more self-documenting (e.g., `WHERE _change_type = 'insert'`).

- **Consistency**: This approach is widely adopted by other query engines, like databricks (uses `insert`, `delete`, `update_preimage`), trino (uses `insert` and `delete`), snowflake (uses `INSERT`, `DELETE`)

What do you think about this change? This requires to change the previous `$changelog` table implementation, but I think it is worth to do this before the release. 

[1] databricks: https://docs.databricks.com/aws/en/sql/language-manual/functions/table_changes
[2] trino: https://trino.io/docs/current/connector/iceberg.html#table-changes
[3] snowflake: https://docs.snowflake.com/en/user-guide/streams-intro#label-streams-changes-clause

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/source/deserializer/BinlogDeserializationSchema.java
wuchong: Please also add tests in `FlinkRecordEmitterTest` to verify the state offset in the UPDATE_BEFORE and UPDATE_AFTER case.

MehulBatra: Thank you @wuchong, for the quick feedback. I will go through and address them one by one.

MehulBatra:
> Thank @MehulBatra for the great pull request. I left some comments inline.
> 
> Besides, regarding the metadata column `_change_type` in the `$changelog` and `$binlog` virtual tables, the current design uses abbreviations like `+I`, `-U`, `+U`, and `-D`.
> 
> I re-considered it again, while these align with Flink's internal `RowKind` representation, I think maybe using full descriptive strings instead would be better, such as:
> 
> * `insert` (instead of `+I` and `+A`)
> * `update_before` (instead of `-U`)
> * `update_after` (instead of `+U`)
> * `delete` (instead of `-D`)
> 
> And for `$binlog` tables
> 
> * `insert` (instead of `I`)
> * `update` (instead of `U`)
> * `delete` (instead of `D`)
> 
> Reasoning:
> 
> * **Improved Readability**: Full words are much more intuitive for SQL users, especially those who may not be familiar with Flink's internal row-kind symbols.
> * **Better User Experience**: It makes the audit and incremental ETL queries more self-documenting (e.g., `WHERE _change_type = 'insert'`).
> * **Consistency**: This approach is widely adopted by other query engines, like databricks (uses `insert`, `delete`, `update_preimage`), trino (uses `insert` and `delete`), snowflake (uses `INSERT`, `DELETE`)
> 
> What do you think about this change? This requires to change the previous `$changelog` table implementation, but I think it is worth to do this before the release.
> 
> [1] databricks: https://docs.databricks.com/aws/en/sql/language-manual/functions/table_changes [2] trino: https://trino.io/docs/current/connector/iceberg.html#table-changes [3] snowflake: https://docs.snowflake.com/en/user-guide/streams-intro#label-streams-changes-clause

Agreed. Following industry standards (Databricks, Trino, Snowflake) makes integration and migration much easier for users. The improved readability is worth the change before release.
I'll update the this PR accordingly and create a seperate one for changelog.

fluss-flink/fluss-flink-common/src/test/java/org/apache/fluss/flink/catalog/FlinkCatalogITCase.java
MehulBatra: oh shit, I missed it again. I made sure to cover this in binglogUT, will fix this on Catalog UT aswell

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/source/BinlogFlinkTableSource.java
MehulBatra: I agree it wasn't gonna used ever except to check length > 0, changing it made the code much cleaner, thanks!

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/source/BinlogFlinkTableSource.java
MehulBatra: https://github.com/apache/fluss/issues/2541

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/FlinkConnectorOptions.java
MehulBatra: I renamed and reintroduced as to address BINLOG_IS_PARTITIONED indicate whether the base table is partitioned for $binlog sources.

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/catalog/FlinkTableFactory.java
MehulBatra: I see that after rebasing, the entire logic of prefixesToSkip has been removed from flinkTablefactory, the helper.validateExcept() call was removed, so there's no longer any validation that rejects unknown options or I am missing anything here?

wuchong: Yes, you can checkout the discussion in https://github.com/apache/fluss/pull/2431

fluss-common/src/main/java/org/apache/fluss/metadata/TableDescriptor.java
MehulBatra: But if the user has these column names in its physical table, wouldn't it conflict while writing query for binlog virtual table, hence I intentionally made them reserved?

wuchong: It won't conflict, because the original columns names are nested undert `before` and `after`. If user has a table `(id int, before string)` then the schema of `$binlog` will be `(_change_type string, _log_offset bigint, _commit_timestamp timestamp_ltz(3), before ROW<id int, before string>)`.

You can reference the original `before` column by `select before.before from T`.,

PR #2531, opened 2026-01-30
[paimon] Not bundle paimon in fluss-lake-paimon to avoid class conflict

### LINKED ISSUE #2426: Refactor fluss-lake-paimon packaging to provide thin JARs
### Search before asking

### Description

**Background**
Currently, the fluss-lake-paimon JAR includes a specific version of Paimon within its packaging. While this makes it easy to use out of the box, it lacks flexibility and leads to dependency conflicts in complex environments.

**The Problem**
By embedding Paimon classes directly without shading:

- Class Conflicts: If a user’s environment (e.g., Flink/Spark cluster) already has a different version of Paimon, it results in LinkageError or ClassCastException.

- Version Inflexibility: Users cannot easily swap the Paimon version to a bug-fixed or slightly different version without rebuilding the entire Fluss connector.

**Proposed Solution**
I propose to split the current packaging into two distinct artifacts:
1. fluss-lake-paimon (The "Thin" JAR)

   Behavior: Does not include any Paimon classes. Paimon should be marked as a provided dependency.

   Use Case: Advanced users who want to manage their own Paimon version in their classpath.

   Note: While it is compiled against a specific Paimon version, sticking to Paimon's Public APIs should mitigate most ClassNotFoundException or NoSuchMethodError issues when used with compatible versions.

2. fluss-lake-paimon-bundle (The "Shaded" JAR)

   Behavior: Includes a specific version of Paimon (the one it was compiled against).

   Relocation (Shading): All Paimon classes will be shaded (e.g., relocated to com.fluss.shaded.paimon.*).

   Use Case: Users who want a "plug-and-play" experience without worrying about conflicts with their own application's Paimon version.

### Willingness to contribute

luoyuxia:
### Purpose

Linked issue: close #2426 

### Brief change log
- Not bundle `paimon-bundie` jar in `fluss-lake-paimon`

### Tests

verify it in local env manually.

### DIFF
--- a/website/docs/maintenance/operations/upgrade-notes-0.9.md
+++ b/website/docs/maintenance/operations/upgrade-notes-0.9.md
@@ -62,12 +62,12 @@ Starting from Fluss 0.9, the `fluss-lake-paimon` JAR no longer bundles `paimon-b
 
 🔧 **Action Required**: Manually add a compatible `paimon-bundle` JAR to your Flink `lib` directory.
 
-#### Migration Steps
+### Migration Steps
 
 Choose a compatible version of `paimon-bundle` JAR from the [Apache Paimon Downloads](https://paimon.apache.org/docs/1.3/project/download/) page 
 and place it in `$FLINK_HOME/lib/`.
 
-#### Tested Paimon Versions
+### Tested Paimon Versions
 
 | Use Case        | Required/Tested Versions                          |
 |-----------------|---------------------------------------------------|


fluss-dist/src/main/assemblies/plugins.xml
```diff
@@ -33,10 +33,10 @@
             <unpack>false</unpack>
             <useProjectArtifact>false</useProjectArtifact>
             <useProjectAttachments>false</useProjectAttachments>
-            <useTransitiveDependencies>true</useTransitiveDependencies>
-            <useTransitiveFiltering>true</useTransitiveFiltering>
+            <useTransitiveDependencies>false</useTransitiveDependencies>
             <includes>
                 <include>org.apache.flink:flink-shaded-hadoop-2-uber</include>
+                <include>org.apache.paimon:paimon-bundle</include>
```
luoyuxia, line 39: I'd like to still install the paimon-bundle jar into plugin directory for out of box user experience. The class conflict is not a problem in Fluss coordinator server

website/docs/streaming-lakehouse/integrate-data-lakes/paimon.md
```diff
@@ -97,7 +104,7 @@ For further information, refer to Paimon's [SQL Query documentation](https://pai
 #### Union Read of Data in Fluss and Paimon
 
 ##### Prerequisites
-Download the [fluss-lake-paimon-$FLUSS_VERSION$.jar](https://repo1.maven.org/maven2/org/apache/fluss/fluss-lake-paimon/$FLUSS_VERSION$/fluss-lake-paimon-$FLUSS_VERSION$.jar), and place it into `${FLINK_HOME}/lib`.
+Download the [fluss-lake-paimon-$FLUSS_VERSION$.jar](https://repo1.maven.org/maven2/org/apache/fluss/fluss-lake-paimon/$FLUSS_VERSION$/fluss-lake-paimon-$FLUSS_VERSION$.jar) and [paimon-bundle jar](https://repo.maven.apache.org/maven2/org/apache/paimon/paimon-bundle/1.3.1/paimon-bundle-1.3.1.jar), and place it into `${FLINK_HOME}/lib`.
```
rionmonster: One question — I can see that the Paimon version for the downloads are currently hardcoded to the `1.3.1` release. Would it be useful to store this in a separate variable to better support future changes (e.g. `PAIMON_VERSION`)?

rionmonster (approved):
Left one comment, which isn't at all required. Otherwise, looks good. Approving!

website/docs/streaming-lakehouse/integrate-data-lakes/paimon.md
luoyuxia: good suggestion

luoyuxia:
@wuchong Could you please review it when you got some time?

website/docs/streaming-lakehouse/integrate-data-lakes/paimon.md
wuchong: Good suggestion. What do you think @luoyuxia ?

fluss-dist/src/main/assemblies/plugins.xml
wuchong: Sure.

website/src/plugins/remark-version-replace/index.js
```diff
@@ -53,7 +53,9 @@ const plugin = (options) => {
         const replacements = {
             "$FLUSS_VERSION$": version.fullVersion,
             "$FLUSS_VERSION_SHORT$": version.shortVersion,
-            "$FLUSS_DOCKER_VERSION$": version.dockerVersion
+            "$FLUSS_DOCKER_VERSION$": version.dockerVersion,
+            "$PAIMON_VERSION$": "1.3.1",
+            "$PAIMON_VERSION_SHORT$": "1.3"
```
wuchong: We shouldn’t hard-code the Paimon version in `index.js`, as this file is shared across all versioned documentation sites.  

For example, if we release Fluss 0.9 with Paimon 1.3.1, and later upgrade to Paimon 1.4.1 in a subsequent Fluss version, the shared `index.js` would incorrectly update the **0.9 documentation** to reference Paimon 1.4.1—even though that combination was never tested and may be incompatible.

To ensure version accuracy and isolation, we should move such version-specific variables into `fluss-versions.json`, where they can be properly scoped per Fluss release.

luoyuxia: Thank you for the explantion!

PR #2532, opened 2026-01-30
[spark] Add startup mode for batch read

### LINKED ISSUE #2549: [spark] Add startup mode config
### Search before asking

### Description

subtask of #2527

Add a new option start.up.mode to read different offset from fluss.

### Willingness to contribute

Yohahaha:
### Purpose

Add a new option `start.up.mode` to read different offset from fluss.
This PR only changes batch read related class.

Linked issue: close #2549

### DIFF
--- a/fluss-spark/fluss-spark-common/src/main/scala/org/apache/fluss/spark/read/FlussBatch.scala
+++ b/fluss-spark/fluss-spark-common/src/main/scala/org/apache/fluss/spark/read/FlussBatch.scala
@@ -19,7 +19,7 @@ package org.apache.fluss.spark.read
 
 import org.apache.fluss.client.{Connection, ConnectionFactory}
 import org.apache.fluss.client.admin.Admin
-import org.apache.fluss.client.initializer.{BucketOffsetsRetrieverImpl, OffsetsInitializer}
+import org.apache.fluss.client.initializer.{BucketOffsetsRetrieverImpl, OffsetsInitializer, SnapshotOffsetsInitializer}
 import org.apache.fluss.client.metadata.KvSnapshots
 import org.apache.fluss.client.table.scanner.log.LogScanner
 import org.apache.fluss.config.Configuration
@@ -47,6 +47,10 @@ abstract class FlussBatch(
 
   lazy val partitionInfos: util.List[PartitionInfo] = admin.listPartitionInfos(tablePath).get()
 
+  def startOffsetsInitializer: OffsetsInitializer
+
+  def stoppingOffsetsInitializer: OffsetsInitializer
+
   protected def projection: Array[Int] = {
     val columnNameToIndex = tableInfo.getSchema.getColumnNames.asScala.zipWithIndex.toMap
     readSchema.fields.map {
@@ -72,11 +76,17 @@ class FlussAppendBatch(
     tablePath: TablePath,
     tableInfo: TableInfo,
     readSchema: StructType,
-    startOffsetsInitializer: OffsetsInitializer,
-    stoppingOffsetsInitializer: OffsetsInitializer,
     options: CaseInsensitiveStringMap,
     flussConfig: Configuration)
   extends FlussBatch(tablePath, tableInfo, readSchema, flussConfig) {
+
+  override val startOffsetsInitializer: OffsetsInitializer = {
+    FlussOffsetInitializers.startOffsetsInitializer(options, flussConfig)
+  }
+
+  override val stoppingOffsetsInitializer: OffsetsInitializer = {
+    FlussOffsetInitializers.stoppingOffsetsInitializer(true, options, flussConfig)
+  }
 
   override def planInputPartitions(): Array[InputPartition] = {
     val bucketOffsetsRetrieverImpl = new BucketOffsetsRetrieverImpl(admin, tablePath)
@@ -155,11 +165,21 @@ class FlussUpsertBatch(
     tablePath: TablePath,
     tableInfo: TableInfo,
     readSchema: StructType,
-    startOffsetsInitializer: OffsetsInitializer,
-    stoppingOffsetsInitializer: OffsetsInitializer,
     options: CaseInsensitiveStringMap,
     flussConfig: Configuration)
   extends FlussBatch(tablePath, tableInfo, readSchema, flussConfig) {
+
+  override val startOffsetsInitializer: OffsetsInitializer = {
+    val offsetsInitializer = FlussOffsetInitializers.startOffsetsInitializer(options, flussConfig)
+    if (!offsetsInitializer.isInstanceOf[SnapshotOffsetsInitializer]) {
+      throw new UnsupportedOperationException("Upsert scan only support FULL startup mode.")
+    }
+    offsetsInitializer
+  }
+
+  override val stoppingOffsetsInitializer: OffsetsInitializer = {
+    FlussOffsetInitializers.stoppingOffsetsInitializer(true, options, flussConfig)
+  }
 
   private val bucketOffsetsRetriever = new BucketOffsetsRetrieverImpl(admin, tablePath)
 
--- a/fluss-spark/fluss-spark-common/src/main/scala/org/apache/fluss/spark/read/FlussScan.scala
+++ b/fluss-spark/fluss-spark-common/src/main/scala/org/apache/fluss/spark/read/FlussScan.scala
@@ -17,7 +17,6 @@
 
 package org.apache.fluss.spark.read
 
-import org.apache.fluss.client.initializer.SnapshotOffsetsInitializer
 import org.apache.fluss.config.Configuration
 import org.apache.fluss.metadata.{TableInfo, TablePath}
 import org.apache.fluss.spark.SparkConversions
@@ -47,18 +46,7 @@ case class FlussAppendScan(
   extends FlussScan {
 
   override def toBatch: Batch = {
-    val startOffsetsInitializer =
-      FlussOffsetInitializers.startOffsetsInitializer(options, flussConfig)
-    val stoppingOffsetsInitializer =
-      FlussOffsetInitializers.stoppingOffsetsInitializer(true, options, flussConfig)
-    new FlussAppendBatch(
-      tablePath,
-      tableInfo,
-      readSchema,
-      startOffsetsInitializer,
-      stoppingOffsetsInitializer,
-      options,
-      flussConfig)
+    new FlussAppendBatch(tablePath, tableInfo, readSchema, options, flussConfig)
   }
 }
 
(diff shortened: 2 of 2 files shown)


Yohahaha:
@YannByron @wuchong please help take a look, thank you!

fluss-spark/fluss-spark-common/src/main/scala/org/apache/fluss/spark/SparkConnectorOptions.scala
```diff
@@ -58,4 +58,15 @@ object SparkConnectorOptions {
...
+
+  object StartUpMode extends Enumeration {
+    val FULL, EARLIEST, LATEST, TIMESTAMP = Value
+  }
+
+  val SCAN_START_UP_MODE: ConfigOption[String] =
+    ConfigBuilder
+      .key("scan.startup.mode")
+      .stringType()
+      .defaultValue(StartUpMode.LATEST.toString)
```
wuchong: Should we use the default `FULL` mode to stay aligned with the Flink connector?  
Using `LATEST` by default may result in empty results if the user doesn’t explicitly specify a startup mode for the query.

fluss-spark/fluss-spark-common/src/main/scala/org/apache/fluss/spark/read/FlussAppendPartitionReader.scala
```diff
@@ -36,41 +35,41 @@ class FlussAppendPartitionReader(
...
-      currentRow = convertToSparkRow(scanRecord)
-      return true
-    }
-
-    // Poll for more records
+  private def pollMoreRecords(): Unit = {
     val scanRecords = logScanner.poll(POLL_TIMEOUT)
+    if ((scanRecords == null || scanRecords.isEmpty) && currentOffset < flussPartition.stopOffset) {
+      throw new IllegalStateException(s"No more data from fluss server," +
+        s" but current offset $currentOffset not reach the stop offset ${flussPartition.stopOffset}")
```
wuchong, line 50: `logScanner.poll()` may return empty results when the Fluss server is undergoing recovery, restart, or rebalance. Given that the current `POLL_TIMEOUT` is set to a very short duration (`100ms`), this scenario is highly likely to occur.  

Currently, the source immediately throws an exception if `logScanner.poll()` returns no records, which makes it unstable during Fluss server failover events.

A straightforward fix is to increase the `POLL_TIMEOUT` to **60 seconds**. This means the source will wait up to 60 seconds for data during transient server unavailability. If the Fluss server still hasn’t recovered within that window, we can then throw an exception to alert users.

fluss-spark/fluss-spark-common/src/main/scala/org/apache/fluss/spark/read/FlussAppendPartitionReader.scala
```diff
@@ -36,41 +35,41 @@ class FlussAppendPartitionReader(
...
+    if (!currentRecords.hasNext) {
+      pollMoreRecords()
     }
 
-    currentRecords = bucketRecords.iterator()
+    // If we have records in current batch, return next one
     if (currentRecords.hasNext) {
       val scanRecord = currentRecords.next()
       currentRow = convertToSparkRow(scanRecord)
+      currentOffset += 1
```
wuchong: We can't simply `+1` and should use `currentOffset = scanRecord.logOffset() + 1` instead. Because, there are some record batch increases log offsets without any records in it. Simply +1 will lead to missing some data.

fluss-spark/fluss-spark-common/src/main/scala/org/apache/fluss/spark/read/FlussAppendPartitionReader.scala
```diff
@@ -83,10 +82,13 @@ class FlussAppendPartitionReader(
...
+    logInfo(s"Prepare read table $tablePath partition $partitionId bucket $bucketId" +
+      s" with start offset ${flussPartition.startOffset} stop offset ${flussPartition.stopOffset}")
     if (partitionId != null) {
-      logScanner.subscribeFromBeginning(partitionId, bucketId)
+      logScanner.subscribe(partitionId, bucketId, flussPartition.startOffset)
     } else {
-      logScanner.subscribeFromBeginning(bucketId)
+      logScanner.subscribe(bucketId, flussPartition.startOffset)
     }
+    pollMoreRecords()
```
wuchong, line 95: When both the start and end offsets are set to `LATEST`, the stop offset may be less than or equal to the start offset. In such cases, attempting to poll records will cause `pollMoreRecords()` to throw an exception.  

To avoid this, we should explicitly validate that the start offset is strictly less than the stop offset before initiating polling.

fluss-spark/fluss-spark-common/src/main/scala/org/apache/fluss/spark/read/FlussBatch.scala
```diff
@@ -110,11 +155,12 @@ class FlussUpsertBatch(
     tablePath: TablePath,
     tableInfo: TableInfo,
     readSchema: StructType,
+    startOffsetsInitializer: OffsetsInitializer,
```
wuchong: The `startOffsetsInitializer` is not used in `FlussUpsertBatch`.

fluss-spark/fluss-spark-common/src/main/scala/org/apache/fluss/spark/read/FlussScan.scala
```diff
@@ -60,6 +95,15 @@ case class FlussUpsertScan(
   extends FlussScan {
 
   override def toBatch: Batch = {
-    new FlussUpsertBatch(tablePath, tableInfo, readSchema, options, flussConfig)
+    val startOffsetsInitializer = FlussScan.startOffsetsInitializer(options)
```
wuchong: The `startOffsetsInitializer` is not used `FlussUpsertBatch`. For `FlussUpsertBatch` (primary key tables), we currently only support `FULL` startup mode that reads kv snapshot first, and then swith to the corresponding log offsets. So we should check the startup mode is FULL, otherwise, throw unsupported exception.

fluss-spark/fluss-spark-common/src/main/scala/org/apache/fluss/spark/read/FlussAppendPartitionReader.scala
Yohahaha: poll timeout could be explicit configured now, and default values is 10s align with flink.

fluss-spark/fluss-spark-common/src/main/scala/org/apache/fluss/spark/read/FlussAppendPartitionReader.scala
Yohahaha: add range check, thanks for the advice!

fluss-spark/fluss-spark-common/src/main/scala/org/apache/fluss/spark/SparkFlussConf.scala
```diff
@@ -17,23 +17,37 @@
...
-      .withDescription(READ_OPTIMIZED.doc)
+      .defaultValue(false)
+      .withDescription(
+        "If true, Spark will only read data from data lake snapshot or kv snapshot, not execute merge them with log changes. This is a temporary configuration that will be deprecated when read-optimized table(e.g. `mytbl$ro`) is supported.")
 
+  object StartUpMode extends Enumeration {
+    val FULL, EARLIEST, LATEST, TIMESTAMP = Value
+  }
+
+  val SCAN_START_UP_MODE: ConfigOption[String] =
```
YannByron, line 40: I suggest to place these common options used both spark and flink in `ConfigOptions`, like https://github.com/apache/paimon/blob/a10a44892cd5e9dbac705762ed6774674357692f/paimon-api/src/main/java/org/apache/paimon/CoreOptions.java#L947 in paimon. We can do this in separate pr maybe.

fluss-spark/fluss-spark-common/src/main/scala/org/apache/fluss/spark/read/FlussScan.scala
```diff
@@ -36,6 +37,38 @@ trait FlussScan extends Scan {
   }
 }
 
+object FlussScan {
```
YannByron: Move these to `FlussBatch` or other common place for `MicroBatchStream`. `OffsetsInitializer` affects how to plan (which is controlled in `FlussBatch`), not how to generate `Batch` and `MicroBatchStream`. And keep the original definition of `FlussXXXScan`.

fluss-spark/fluss-spark-common/src/main/scala/org/apache/fluss/spark/SparkFlussConf.scala
```diff
@@ -17,23 +17,37 @@
...
+    ConfigBuilder
+      .key("scan.startup.mode")
+      .stringType()
+      .defaultValue(StartUpMode.FULL.toString)
+      .withDescription("The start up mode when read Fluss table.")
+
+  val LOG_SCANNER_POLL_TIMEOUT: ConfigOption[Duration] =
+    ConfigBuilder
+      .key("log.scanner.poll.timeout")
```
YannByron: I prefer `scan.poll.timeout`.

fluss-spark/fluss-spark-ut/src/test/scala/org/apache/fluss/spark/SparkPrimaryKeyTableReadTest.scala
```diff
@@ -38,6 +38,14 @@ class SparkPrimaryKeyTableReadTest extends FlussSparkTestBase {
     new Configuration()
   }
 
+  override protected def beforeEach(): Unit = {
```
YannByron, line 41: All the configs are default values, this's meaningless,  so I suggest to remove this.

Yohahaha: @YannByron WithFlussAdmin use singleton fluss config that will affect integration test, if we dont reset config here, the fluss config working scope need to rethink.

fluss-spark/fluss-spark-common/src/main/scala/org/apache/fluss/spark/SparkFlussConf.scala
Yohahaha: good advice, let's refactor these common config into fluss-common module in another PR, flink and spark can share it. cc @wuchong

fluss-spark/fluss-spark-common/src/main/scala/org/apache/fluss/spark/read/FlussScan.scala
Yohahaha: OffsetsInitializer here is used to help generate Batch/MicroBatchStream, it helps plan split in each Batch/MicroBatchStream implementation, I dont really get your points, but I will move it to a new object class FlussOffsetInitializers

fluss-spark/fluss-spark-ut/src/test/scala/org/apache/fluss/spark/SparkPrimaryKeyTableReadTest.scala
Yohahaha: UT launched by mvn will fail without it.

Yohahaha:
@YannByron any more comments?

YannByron:
+1. cc @wuchong

fluss-spark/fluss-spark-common/src/main/scala/org/apache/fluss/spark/SparkFlussConf.scala
wuchong: Good point. We can introduce `ConnectorOptions` in `fluss-common` package `org.apache.fluss.config` to share the common options for different connectors.

wuchong (approved):

PR #2537, opened 2026-02-01
[lake/lance] Add Array type support for Lance

### LINKED ISSUE #2186: [lake/lance] Array support for Lance
### Search before asking

### Description

add Array type support for Lance

### Willingness to contribute

XuQianJin-Stars:
### Purpose

Linked issue: close #2186

Add Array type support for Lance storage format to enable storing and retrieving tables with array columns.

### Brief change log

**LanceArrowUtils.java:**
- Added ArrayType import and handling in `toArrowType()` method to convert Fluss ArrayType to Arrow List type
- Enhanced `toArrowField()` method to create children fields for ArrayType, properly handling nested array structures by recursively creating element fields

**ArrowDataConverter.java:**
- Added ListVector import for handling Arrow list vectors
- Implemented `copyListVectorData()` method to handle copying of ListVector data between shaded and non-shaded Arrow implementations
- Enhanced `copyVectorData()` method to detect and delegate ListVector copying to the specialized handler
- Added recursive handling for nested array elements in the data conversion process

**LanceTieringTest.java:**
- Added new test case `testTieringWriteTableWithArrayType()` to verify array type support
- Test creates tables with both STRING array and INT array columns
- Test verifies writing and reading array data through Lance storage
- Added helper methods `genArrayTableRecords()` and `verifyArrayTableRecords()` for array data generation and verification

### Tests

- **Unit Test:** `LanceTieringTest#testTieringWriteTableWithArrayType` - Tests writing and reading tables with array columns (STRING array and INT array) through Lance storage
- **Existing Tests:** All existing Lance tiering tests pass, verifying backward compatibility
- **Test Coverage:** Validates array schema conversion, data writing, committing, and reading with proper array element verification

### API and Format

- **No API changes:** This change only adds support for existing ArrayType in Lance storage implementation
- **No format changes:** Uses standard Arrow List format for array storage, which is compatible with Lance

### Documentation

- **No new feature introduction:** This change adds support for an existing Fluss data type (ArrayType) in the Lance storage backend
- **No documentation update needed:** Array type is already documented in Fluss type system; this implementation extends Lance storage to support it

### DIFF
--- a/fluss-lake/fluss-lake-lance/src/test/java/org/apache/fluss/lake/lance/tiering/LanceTieringITCase.java
+++ b/fluss-lake/fluss-lake-lance/src/test/java/org/apache/fluss/lake/lance/tiering/LanceTieringITCase.java
@@ -24,14 +24,12 @@ import org.apache.fluss.lake.lance.testutils.FlinkLanceTieringTestBase;
 import org.apache.fluss.metadata.TableBucket;
 import org.apache.fluss.metadata.TablePath;
 import org.apache.fluss.row.GenericArray;
-import org.apache.fluss.row.InternalRow;
 import org.apache.fluss.server.zk.data.lake.LakeTable;
 
 import com.lancedb.lance.Dataset;
 import com.lancedb.lance.ReadOptions;
 import com.lancedb.lance.Transaction;
 import org.apache.arrow.memory.RootAllocator;
-import org.apache.arrow.vector.VarCharVector;
 import org.apache.arrow.vector.VectorSchemaRoot;
 import org.apache.arrow.vector.ipc.ArrowReader;
 import org.apache.flink.core.execution.JobClient;
@@ -39,11 +37,8 @@ import org.apache.flink.streaming.api.environment.StreamExecutionEnvironment;
 import org.junit.jupiter.api.BeforeAll;
 import org.junit.jupiter.api.Test;
 
-import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
-import java.util.Iterator;
-import java.util.List;
 import java.util.Map;
 
 import static org.apache.fluss.lake.committer.LakeCommitter.FLUSS_LAKE_SNAP_BUCKET_OFFSET_PROPERTY;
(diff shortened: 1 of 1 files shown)


fluss-lake/fluss-lake-lance/src/test/java/org/apache/fluss/lake/lance/tiering/LanceTieringTest.java
```diff
@@ -322,4 +325,140 @@ private Schema createTable(LanceConfig config) {
 
         return schema;
     }
+
+    @Test
+    void testTieringWriteTableWithArrayType() throws Exception {
```
wuchong: This test looks unusual—it manually instantiates components like `LakeWriter` and `LakeCommitter` using `LanceLakeTieringFactory` to perform tiering. However, this approach is highly error-prone and doesn’t represent the actual end-to-end path that needs to be covered when adding support for a new type like `ARRAY`. Consider this: if we were to add 10 new types, would we really need to write 10 separate, hand-wired integration tests like this? Such tests are not maintainable.

To properly validate `ARRAY` type support, we should perform an end-to-end verification. Specifically, we can simply add a few `ARRAY`-type columns (like `array<string>` and `array<int>`) to the `logTable` created in `org.apache.fluss.lake.lance.tiering.LanceTieringITCase#testTiering`, and then verify that these array fields are correctly written and read in the result assertions.

Additionally, for the Lance format, `ARRAY` types are especially important for **vector embeddings**. Therefore, we should also include a dedicated test case for `ARRAY<FLOAT>` to ensure proper handling of vector data.

fluss-lake/fluss-lake-lance/src/test/java/org/apache/fluss/lake/lance/tiering/LanceTieringITCase.java
```diff
@@ -94,8 +97,46 @@ void testTiering() throws Exception {
...
-        checkDataInLanceAppendOnlyTable(config, flussRows);
+        // check data in lance, including array columns
+        checkDataInLanceAppendOnlyTableWithArrays(config, flussRows);
+        checkSnapshotPropertyInLance(config, Collections.singletonMap(t1Bucket, 30L));
+
+        jobClient.cancel().get();
+    }
+
+    @Test
+    void testTieringWithFloatArrayForVectorEmbeddings() throws Exception {
```
wuchong: Why not merge this test into `testTiering()` by simply adding multiple array-type columns?  
Having a separate test case doesn’t seem to add significant value and unnecessarily increases overall test execution time.

XuQianJin-Stars:
@wuchong @luoyuxia Hi, i already updated the pr. Please help review when you got some time.

fluss-lake/fluss-lake-lance/src/test/java/org/apache/fluss/lake/lance/tiering/LanceTieringITCase.java
```diff
@@ -144,4 +196,61 @@ private void checkDataInLanceAppendOnlyTable(LanceConfig config, List<InternalRo
             assertThat(flussRowIterator.hasNext()).isFalse();
         }
     }
+
+    private void checkDataInLanceWithAllArrayTypes(
```
wuchong: The data assertion code appears quite verbose and hard to maintain. Could we simplify it by using `readerRoot.contentToTSVString()`—similar to how we assert Flink `Row` results? This would make the test more concise and easier to read.

Additionally, it seems you haven’t merged this test into `testTiering()` yet. Is there a specific reason this can’t be done? If so, could you please clarify?

wuchong:
@juice928 stop

PR #2545, opened 2026-02-02
Add undo recovery support for aggregation tables

### LINKED ISSUE #2544: Add undo recovery support for aggregation tables
### Search before asking

### Description

### Willingness to contribute

platinumhamburg:
This PR introduces the undo recovery mechanism for Flink sink writer to handle failure recovery scenarios with aggregation merge engine tables.

Key components:
- ByteArrayWrapper: Utility class for using byte arrays as map keys
- UndoComputer: Computes undo operations by comparing checkpoint state with current log records, supporting both full row and partial update modes
- UndoRecoveryExecutor: Executes undo operations using UpsertWriter
- UndoRecoveryCoordinator: Coordinates the recovery process across buckets, managing log scanning, undo computation, and execution
- BucketRecoveryContext: Holds per-bucket recovery state

The undo recovery works by:
1. Scanning log records from checkpoint offset to current end offset
2. Computing inverse operations for uncommitted records
3. Executing undo operations to restore table state

For partial update mode, INSERT records require full row deletion (not partial column deletion), which is handled by using a separate delete writer with null target columns.

### Purpose

Linked issue: close #2544 2544

### DIFF
--- a/fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/sink/writer/undo/UndoRecoveryExecutor.java
+++ b/fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/sink/writer/undo/UndoRecoveryExecutor.java
@@ -216,7 +216,7 @@ public class UndoRecoveryExecutor {
             // Log progress periodically (at the end of each loop iteration)
             long now = System.currentTimeMillis();
             if (now - lastProgressLogTime >= PROGRESS_LOG_INTERVAL_MS) {
-                long elapsedMs = System.currentTimeMillis() - startTimeMs;
+                long elapsedMs = now - startTimeMs;
                 LOG.info(
                         "Undo recovery waiting for changelog records: {} bucket(s) incomplete, "
                                 + "waited {} minutes so far (max: {} minutes)",


fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/sink/writer/undo/BucketRecoveryContext.java
```diff
@@ -0,0 +1,152 @@
...
+     * <p>Complete means either:
+     *
+     * <ul>
+     *   <li>No recovery is needed (checkpointOffset >= logEndOffset), or
+     *   <li>We have processed all expected records (totalRecordsProcessed >= logEndOffset -
+     *       checkpointOffset)
+     * </ul>
+     *
+     * <p>This implementation assumes that changelog offsets are contiguous (no gaps). The number of
+     * records to process equals logEndOffset - checkpointOffset.
```
wuchong: That’s not accurate. There are certain special log record batches that **advance the log offset even though they contain no  records**. Examples include:
- The first merge engine batch for a key that is **not the first row** ,
- Delete operations targeting **non-existent rows**.

In both cases, the batch is logically empty from a data perspective but still increments the log offset to maintain consistency and ordering guarantees.

See implementation details in:  
https://github.com/apache/fluss/blob/main/fluss-server/src/main/java/org/apache/fluss/server/kv/KvTablet.java#L369-L378

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/sink/writer/undo/BucketRecoveryContext.java
```diff
@@ -0,0 +1,152 @@
...
+     */
+    public boolean isComplete() {
+        // If no recovery is needed, we're already complete
+        if (!needsRecovery()) {
+            return true;
+        }
+        // Number of records to process = logEndOffset - checkpointOffset
+        // (offsets are contiguous, no gaps)
+        long expectedRecords = logEndOffset - checkpointOffset;
+        return totalRecordsProcessed >= expectedRecords;
```
wuchong: Ditto. We shouldn’t rely on the number of records to determine completeness. Instead, we should simply track the processed log offset using `lastProcessedOffset`.

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/sink/writer/undo/UndoComputer.java
```diff
@@ -0,0 +1,210 @@
...
+ *   <li>{@code DELETE} → Re-insert the deleted row (it existed at checkpoint)
+ * </ul>
+ *
+ * <p>For partial update mode, two writers are used:
+ *
+ * <ul>
+ *   <li>{@code partialWriter} - for UPDATE_BEFORE and DELETE undo (restore/re-insert partial
+ *       columns)
+ *   <li>{@code deleteWriter} - for INSERT undo (delete entire row, not just partial columns)
+ * </ul>
```
wuchong: I don’t understand why a full-row delete writer is needed for partial update recovery. Let me illustrate with an example:

1. **Initial state**:  
   - Job-1 upserts into the table on columns `(id, name)` with records `(1, "user1")` and `(2, "user2")`.  
   - Job-2 upserts on columns `(id, price)` with records `(1, 100)` and `(2, 200)`.  
   → The resulting table rows are:  
     `(1, "user1", 100)` and `(2, "user2", 200)`.

2. **After Job-1 fails over and recovers**:  
   To roll back Job-1’s effects, we apply its undo log (i.e., set `name = null` for IDs 1 and 2).  
   The expected result should be:  
   `(1, null, 100)` and `(2, null, 200)` — **not an empty table**.

However, according to the current Javadoc, the outcome would incorrectly be an **empty table**, which is clearly wrong.

Note that **partial deletes already handle row removal correctly**: if all non-primary-key columns become `null`, the row is automatically deleted. Therefore:
- If **only Job-1 existed** (no Job-2), rolling back to a prior checkpoint via partial delete would correctly remove the entire row—**no full-row delete writer is needed**.
- But when **multiple jobs update different column subsets**, we must preserve columns written by other jobs during recovery. A full-row delete would erroneously wipe out those contributions.

Thus, using a full-row delete writer in partial update recovery breaks correctness in multi-writer scenarios. The existing partial delete mechanism, which respects column-level null semantics, is sufficient and more accurate.

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/sink/writer/undo/UndoComputer.java
```diff
@@ -0,0 +1,210 @@
...
+     * <p>Only the first change for each primary key triggers an undo operation.
+     *
+     * @param record the changelog record
+     * @param processedKeys set of already processed primary keys for deduplication
+     * @return CompletableFuture for the async write, or null if skipped
+     */
+    @Nullable
+    public CompletableFuture<?> processRecord(
+            ScanRecord record, Set<ByteArrayWrapper> processedKeys) {
+        byte[] encodedKey = keyEncoder.encodeKey(record.getRow());
```
wuchong, line 88: We should skip key encoding if it is `UPDATE_AFTER` change type to optimize the performance.

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/sink/writer/undo/UndoComputer.java
```diff
@@ -0,0 +1,210 @@
...
+                // Use deleteWriter if available (for partial update mode to delete entire row)
+                if (deleteWriter != null) {
+                    // Flush partial writer before using delete writer to avoid mixing
+                    // records with different target columns in the same batch
+                    if (!lastUsedDeleteWriter) {
+                        writer.flush();
+                        lastUsedDeleteWriter = true;
+                    }
+                    return deleteWriter.delete(row);
+                }
```
wuchong: If we don't need the full row `deleteWriter`, we don't need this complex logic.

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/sink/writer/undo/UndoRecoveryCoordinator.java
```diff
@@ -0,0 +1,259 @@
...
+ *
+ * <ul>
+ *   <li>{@link UndoComputer} - Undo operation computation using {@link KeyEncoder} for primary key
+ *       encoding
+ *   <li>{@link UndoRecoveryExecutor} - Changelog reading and write execution
+ * </ul>
+ *
+ * @see MergeMode#OVERWRITE
+ */
+public class UndoRecoveryCoordinator {
```
wuchong: Coordinator is confusing because there is Coordinator in both Flink and Fluss. I suggest to rename it to `UndoRecoveryManager` or something like that.

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/sink/writer/undo/UndoRecoveryCoordinator.java
```diff
@@ -0,0 +1,259 @@
...
+    public UndoRecoveryCoordinator(Table table, @Nullable int[] targetColumnIndexes) {
+        this.table = table;
+        this.targetColumnIndexes = targetColumnIndexes;
+
+        logPartialUpdateConfig(targetColumnIndexes);
+    }
+
+    private void logPartialUpdateConfig(@Nullable int[] targetColumnIndexes) {
+        if (targetColumnIndexes != null) {
+            LOG.debug(
```
wuchong: I think this is very imporatnt information, and should use `LOG.info`

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/sink/writer/undo/UndoRecoveryCoordinator.java
```diff
@@ -0,0 +1,259 @@
...
+                deleteWriter = table.newUpsert().mergeMode(MergeMode.OVERWRITE).createWriter();
+            }
+
+            // Create UndoComputer with writer for streaming execution
+            Schema schema = table.getTableInfo().getSchema();
+            KeyEncoder keyEncoder =
+                    KeyEncoder.of(
+                            schema.getRowType(),
+                            schema.getPrimaryKey().get().getColumnNames(),
+                            null);
```
wuchong: Use `CompactedKeyEncoder.createKeyEncoder` directly, because `KeyEncoder.of` is
deprecated, and it's hack no passing the lake format parameter. Besides, we should add comment to explain why
we not use the general `KeyEncoder.of` here to avoid confusion.

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/sink/writer/undo/UndoRecoveryManager.java
```diff
@@ -0,0 +1,259 @@
...
+            Schema schema = table.getTableInfo().getSchema();
+            KeyEncoder keyEncoder =
+                    KeyEncoder.of(
+                            schema.getRowType(),
+                            schema.getPrimaryKey().get().getColumnNames(),
+                            null);
+            UndoComputer undoComputer = new UndoComputer(keyEncoder, writer, deleteWriter);
+
+            UndoRecoveryExecutor executor = new UndoRecoveryExecutor(scanner, writer, undoComputer);
+            executor.execute(contexts);
```
wuchong, line 151: nit: add comment to highligh this line  it is a blocking operation which is very heavy and time consuming.

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/sink/writer/undo/UndoRecoveryCoordinator.java
```diff
@@ -0,0 +1,259 @@
...
+
+        List<BucketRecoveryContext> contexts = new ArrayList<>();
+
+        for (Map.Entry<TableBucket, UndoOffsets> entry : bucketUndoOffsets.entrySet()) {
+            TableBucket bucket = entry.getKey();
+            UndoOffsets offsets = entry.getValue();
+
+            BucketRecoveryContext ctx =
+                    new BucketRecoveryContext(bucket, offsets.getCheckpointOffset());
+            ctx.setLogEndOffset(offsets.getLogEndOffset());
```
wuchong: It seems there is no need to **update** LogEndOffset for the context. So we can change it to a `final` field and initialize it in constructor.

fluss-flink/fluss-flink-common/src/test/java/org/apache/fluss/flink/sink/writer/undo/UndoRecoveryCoordinatorITCase.java
```diff
@@ -0,0 +1,977 @@
...
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
+
+/**
+ * Integration tests for {@link UndoRecoveryCoordinator}.
+ *
+ * <p>Tests use complex multi-bucket, multi-key scenarios to ensure only correct implementations
+ * pass. Each test covers multiple aspects to maximize value while minimizing redundancy.
+ */
+public class UndoRecoveryCoordinatorITCase extends FlinkTestBase {
```
wuchong: We don't need to extend `FlinkTestBase`, because there is no Flink cluster needed in this IT case.

fluss-flink/fluss-flink-common/src/test/java/org/apache/fluss/flink/sink/writer/undo/UndoRecoveryManagerITCase.java
```diff
@@ -0,0 +1,977 @@
...
+            for (int i = 0; i < NUM_BUCKETS; i++) {
+                keysAfterCheckpoint.put(i, new HashSet<>());
+            }
+
+            // Phase 2: Operations after checkpoint (batch async)
+            List<CompletableFuture<UpsertResult>> newKeyFutures = new ArrayList<>();
+            for (int id = 200; id < 300; id++) {
+                newKeyFutures.add(writer.upsert(row(id, "new_" + id, id * 10, id)));
+            }
+            for (int i = 0; i < newKeyFutures.size(); i++) {
```
wuchong, line 192: nit: add `writer.flush();` before this line to actively flush data, rather than waiting for async flush.

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/sink/writer/undo/UndoRecoveryExecutor.java
```diff
@@ -0,0 +1,334 @@
...
+
+        // Subscribe and read changelog with streaming execution
+        subscribeAll(toRecover);
+        List<CompletableFuture<?>> allFutures = readChangelogAndExecute(toRecover);
+
+        // Wait for all async writes to complete
+        if (!allFutures.isEmpty()) {
+            CompletableFuture.allOf(allFutures.toArray(new CompletableFuture[0])).get();
+        }
+        writer.flush();
```
wuchong: Move the `writer.flush()` before the future wait. This improves the performance by actively trigger flush, rather than waiting for async flush.

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/sink/writer/undo/UndoRecoveryExecutor.java
```diff
@@ -0,0 +1,334 @@
...
+
+        // Log summary at INFO level for visibility
+        int totalUndoOps = 0;
+        for (BucketRecoveryContext ctx : toRecover) {
+            totalUndoOps += ctx.getProcessedKeys().size();
+        }
+
+        LOG.debug(
```
wuchong: I think this is an important information, change it to `LOG.info`?

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/sink/writer/undo/UndoRecoveryExecutor.java
```diff
@@ -0,0 +1,334 @@
...
+
+        long currentPollTimeoutMs = initialPollTimeoutMs;
+        long totalWaitTimeMs = 0;
+        long lastProgressLogTime = System.currentTimeMillis();
+
+        while (!allComplete(contexts)) {
+            ScanRecords records = scanner.poll(Duration.ofMillis(currentPollTimeoutMs));
+
+            if (records.isEmpty()) {
+                totalWaitTimeMs += currentPollTimeoutMs;
```
wuchong: I think the calculation of `totalWaitTimeMs` is not correct. 

1. The `currentPollTimeoutMs` is not the actual waiting time of the `scanner.poll` method. Because, if there is already data fetched in scanner, it will return directly without waiting. 
2. The `totalWaitTimeMs` is only updated when `records.isEmpty()`. It is possible to wait some time to fetch a record, so when records is not empty, it may cost serveral seconds to wait.

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/sink/writer/undo/UndoRecoveryExecutor.java
```diff
@@ -0,0 +1,334 @@
...
+
+                // Check if we've exceeded the maximum total wait time
+                if (totalWaitTimeMs >= maxTotalWaitTimeMs) {
+                    // Undo recovery failure is fatal - incomplete undo leads to data inconsistency
+                    String incompleteBucketsDetail = getIncompleteBucketsDetail(contexts);
```
wuchong: Just print the the number of buckets in the error message? Otherwise, it maybe a very huge exception message if there are thousands of buckets but very little Flink parallelism.

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/sink/writer/undo/UndoRecoveryExecutor.java
```diff
@@ -0,0 +1,334 @@
...
+                    throw new RuntimeException(
+                            String.format(
+                                    "Undo recovery failed: unable to read all changelog records after "
+                                            + "%.1f minutes of waiting. Incomplete buckets: %s. "
+                                            + "This is a fatal error as incomplete undo recovery "
```
wuchong: Is this really a fatal error and data inconsistency? I think if retry success, it will still recovery the job and data to the checkpointed point?

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/sink/writer/undo/UndoRecoveryExecutor.java
```diff
@@ -0,0 +1,334 @@
...
+                                            + "would lead to data inconsistency.",
+                                    totalWaitTimeMs / 60000.0, incompleteBucketsDetail));
+                }
+
+                // Log progress periodically during long waits
+                long now = System.currentTimeMillis();
+                if (now - lastProgressLogTime >= PROGRESS_LOG_INTERVAL_MS) {
+                    LOG.info(
```
wuchong: This logging is useful; however, it seems to print only when `records` is empty, which makes the logged information highly unpredictable and hard to control.

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/sink/writer/undo/UndoRecoveryExecutor.java
```diff
@@ -0,0 +1,334 @@
...
+                            "Undo recovery waiting for changelog records: {} bucket(s) incomplete, "
+                                    + "waited {} minutes so far (max: {} minutes)",
+                            countIncomplete(contexts),
+                            String.format("%.1f", totalWaitTimeMs / 60000.0),
+                            String.format("%.1f", maxTotalWaitTimeMs / 60000.0));
+                    lastProgressLogTime = now;
+                }
+
+                // Exponential backoff: double the timeout up to maxPollTimeoutMs
+                currentPollTimeoutMs = Math.min(currentPollTimeoutMs * 2, maxPollTimeoutMs);
```
wuchong: I don’t understand why we need to double the poll timeout. It’s safe—and preferable—to use a consistent timeout value (e.g., 10 seconds, as used in `FlinkSourceSplitReader`) throughout. Introducing a doubled or variable timeout adds unnecessary complexity and can make behavior less predictable without clear benefit.

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/sink/writer/undo/UndoRecoveryExecutor.java
```diff
@@ -0,0 +1,334 @@
...
+            // Process records for each bucket, resetting writer state before each bucket
+            for (BucketRecoveryContext ctx : contexts) {
+                if (ctx.isComplete()) {
+                    continue;
+                }
+                List<ScanRecord> bucketRecords = records.records(ctx.getBucket());
+                if (!bucketRecords.isEmpty()) {
+                    undoComputer.resetWriterState();
+                    processRecords(ctx, bucketRecords, allFutures);
+                }
```
wuchong: We should add scanner unsubscribe logic when the context is completed, otherwise, we will continuesly polling data from the completed buckets, and waste a lot of time and resource. 

```
                if (ctx.isComplete()) {
                    // TODO scanner unsubscribe 
                }
```

We may also need to introduce `unsubscribe(int bucket)` method to the `Scanner` interface, but it is not complex.

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/sink/writer/undo/UndoRecoveryExecutor.java
```diff
@@ -0,0 +1,334 @@
...
+        }
+
+        // Log summary
+        for (BucketRecoveryContext ctx : contexts) {
+            LOG.debug(
+                    "Bucket {} read {} records, executed {} undo operations",
+                    ctx.getBucket(),
+                    ctx.getTotalRecordsProcessed(),
+                    ctx.getProcessedKeys().size());
+        }
```
wuchong: Add `if (LOG.isDebugEnabled())` condition for the code block.

wuchong:
Btw, it seems the UNDO recovery for aggregate merge engine table relies on the full changelog of the table. However, `table.changelog.image=WAL` will break this and lead to in-correct results. I think we need to forbid such case, a validation check can be done during table creation and throw exception for tables with agg merge engine and WAL mode. We can do that in another issue. What do you think @platinumhamburg ?

platinumhamburg:
@wuchong Thank you for your detailed review. I’ve addressed all the comments above—please take a look when you have a moment.

platinumhamburg:
> Btw, it seems the UNDO recovery for aggregate merge engine table relies on the full changelog of the table. However, `table.changelog.image=WAL` will break this and lead to in-correct results. I think we need to forbid such case, a validation check can be done during table creation and throw exception for tables with agg merge engine and WAL mode. We can do that in another issue. What do you think @platinumhamburg ?

Yes, this restriction will be addressed in a separate issue [https://github.com/apache/fluss/issues/2608].

fluss-flink/fluss-flink-common/src/test/java/org/apache/fluss/flink/sink/writer/undo/UndoRecoveryCoordinatorITCase.java
```diff
@@ -374,25 +430,11 @@ void testPartialUpdateMultiBucketComprehensiveRecovery() throws Exception {
...
-                verifyFutures.add(
-                        lookuper.lookup(row(keyId))
-                                .thenAccept(
-                                        r ->
-                                                assertThat(r.getSingletonRow())
-                                                        .as(
-                                                                "Zero-offset bucket key %d should be deleted",
-                                                                keyId)
-                                                        .isNull()));
-            }
```
wuchong, line 394: I think we still need to check the `zeroOffsetBucketKeysAfterCheckpoint` related rows after restoring. Could you add this test back?

platinumhamburg: I'll reintroduce this test in #2609.

PR #2268, opened 2025-12-28
[flink] Delta Join additional IT tests and docs improvement

fresh-borzoni:
### Purpose
Linked issue: close #2231

Add comprehensive test coverage for Delta Join feature in Flink 2.2 and improve documentation.
### Brief change log
- Added tests for normal PK tables with `table.delete.behavior=IGNORE` (not just `first_row` merge engine)
- Added tests for joining on bucket key (not just full primary key)
- Added tests verifying LEFT/RIGHT/FULL OUTER joins fail with appropriate validation error
- Added test verifying cascade joins are not supported
- Updated documentation for Flink 2.2 Delta Join with new supported features and limitations

### Tests

- `testDeltaJoinWithPrimaryKeyTableNoDeletes` - normal PK table with `delete.behavior=IGNORE`
- `testDeltaJoinOnBucketKey` - join on bucket key only
- `testDeltaJoinFailsWhenFilterOnNonUpsertKeys` - filter on non-upsert-key columns fails
- `testDeltaJoinOnBucketKey` - join on bucket key only (not full PK)
- `testDeltaJoinFailsWhenSourceHasDelete` - source with DELETE records fails
- `testDeltaJoinFailsWhenJoinKeyNotContainIndex` - join key not containing index fail
- `testDeltaJoinFailsWithLeftJoin` - LEFT JOIN wouldn't be converted to DeltaJoin
- `testDeltaJoinFailsWithRightJoin` - RIGHT JOIN  wouldn't be converted to DeltaJoin
- `testDeltaJoinFailsWithFullOuterJoin` - FULL OUTER JOIN wouldn't be converted to DeltaJoin
- `testDeltaJoinFailsWithCascadeJoin` - cascade join wouldn't be converted to DeltaJoin

### API and Format

No

### Documentation

Yes - updated `docs/engine-flink/delta-joins.md` in Flink 2.2 part.

PR #2546, opened 2026-02-02
[flink] Delta Join additional IT tests and docs improvement

### LINKED ISSUE #2231: Add Flink 2.2 Delta Join tests and documentation
### Search before asking

### Motivation

https://github.com/apache/fluss/pull/2167 supports Flink 2.2 in Fluss and added some tests in `Flink22TableSourceITCase`. However, the tests are very limited:
- use `first_row` fluss table, rather than normal primary key tables (may need to enable `table.delete.behavior=ignore`)
- join on primary keys, rather than bucket key
- innser join, rather than left join or right join

We should also update the documentation with new examples and new limitations. https://fluss.apache.org/docs/next/engine-flink/delta-joins/

### Solution

_No response_

### Anything else?

_No response_

### Willingness to contribute

xuyangzhong:
### Purpose

This pr is based on https://github.com/apache/fluss/pull/2268 and do some refactoring work.

Linked issue: close #2231

### Brief change log

- Added tests for normal PK tables with table.delete.behavior=IGNORE (not just first_row merge engine)
- Added tests for joining on bucket key (not just full primary key)
- Added tests verifying LEFT/RIGHT/FULL OUTER joins fail with appropriate validation error
- Added test verifying cascade joins are not supported
- Updated documentation for Flink 2.2 Delta Join with new supported features and limitations

### Tests

- testDeltaJoinWithPrimaryKeyTableNoDeletes - normal PK table with delete.behavior=IGNORE
- testDeltaJoinOnBucketKey - join on bucket key only
- testDeltaJoinFailsWhenFilterOnNonUpsertKeys - filter on non-upsert-key columns fails
- testDeltaJoinOnBucketKey - join on bucket key only (not full PK)
- testDeltaJoinFailsWhenSourceHasDelete - source with DELETE records fails
- testDeltaJoinFailsWhenJoinKeyNotContainIndex - join key not containing index fail
- testDeltaJoinFailsWithLeftJoin - LEFT JOIN wouldn't be converted to DeltaJoin
- testDeltaJoinFailsWithRightJoin - RIGHT JOIN wouldn't be converted to DeltaJoin
- testDeltaJoinFailsWithFullOuterJoin - FULL OUTER JOIN wouldn't be converted to DeltaJoin
- testDeltaJoinFailsWithCascadeJoin - cascade join wouldn't be converted to DeltaJoin

### API and Format

No

### Documentation

Yes - updated `docs/engine-flink/delta-joins.md` in Flink 2.2 part.

### DIFF
--- a/fluss-flink/fluss-flink-2.2/src/test/java/org/apache/fluss/flink/source/Flink22DeltaJoinITCase.java
+++ b/fluss-flink/fluss-flink-2.2/src/test/java/org/apache/fluss/flink/source/Flink22DeltaJoinITCase.java
@@ -40,6 +40,7 @@ import javax.annotation.Nullable;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.HashMap;
+import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.stream.Collectors;
@@ -103,6 +104,7 @@ public class Flink22DeltaJoinITCase extends FlinkTestBase {
             String columns,
             String primaryKey,
             String bucketKey,
+            @Nullable String partitionKey,
             @Nullable Map<String, String> extraOptions) {
         Map<String, String> withOptions = new HashMap<>();
         if (extraOptions != null) {
@@ -110,16 +112,24 @@ public class Flink22DeltaJoinITCase extends FlinkTestBase {
         }
         withOptions.put("connector", "fluss");
         withOptions.put("bucket.key", bucketKey);
-
-        tEnv.executeSql(
+        StringBuilder ddlBuilder = new StringBuilder();
+        ddlBuilder.append(
                 String.format(
-                        "create table %s ( %s, primary key (%s) NOT ENFORCED ) with ( %s )",
-                        tableName,
-                        columns,
-                        primaryKey,
+                        "create table %s ( %s, primary key (%s) NOT ENFORCED )",
+                        tableName, columns, primaryKey));
+        if (partitionKey != null) {
+            ddlBuilder.append(String.format(" partitioned by (%s)", partitionKey));
+            withOptions.put("table.auto-partition.enabled", "true");
+            withOptions.put("table.auto-partition.time-unit", "year");
+        }
+        ddlBuilder.append(
+                String.format(
+                        " with (%s)",
                         withOptions.entrySet().stream()
                                 .map(e -> String.format("'%s' = '%s'", e.getKey(), e.getValue()))
                                 .collect(Collectors.joining(", "))));
+
+        tEnv.executeSql(ddlBuilder.toString());
     }
 
     /**
@@ -154,12 +164,14 @@ public class Flink22DeltaJoinITCase extends FlinkTestBase {
                 "a1 int, b1 varchar, c1 bigint, d1 int, e1 bigint",
                 "c1, d1",
                 "c1",
+                null,
                 ImmutableMap.of("table.delete.behavior", "IGNORE"));
         createSource(
                 rightTableName,
                 "a2 int, b2 varchar, c2 bigint, d2 int, e2 bigint",
                 "c2, d2",
                 "c2",
+                null,
                 ImmutableMap.of("table.delete.behavior", "IGNORE"));
         createSink(
                 sinkTableName,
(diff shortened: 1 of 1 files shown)


fluss-flink/fluss-flink-2.2/src/test/java/org/apache/fluss/flink/source/Flink22DeltaJoinITCase.java
```diff
@@ -0,0 +1,916 @@
...
+                                + " %s, "
+                                + " primary key (%s) NOT ENFORCED"
+                                + ") with ("
+                                + " 'connector' = 'fluss'"
+                                + ")",
+                        tableName, columns, primaryKey));
+    }
+
+    @Test
+    void testDeltaJoin() throws Exception {
```
wuchong, line 155: I think we should configure all tests to use `table.delete.behavior` and join on the bucket key by default, as this aligns with the recommended approach documented for users implementing delta joins.

Then I suggest to rename this method to `testDeltaJoinOnPrimaryKey` and keep this test as is. And rename `testDeltaJoinOnBucketKey` to `testDeltaJoin`.

fluss-flink/fluss-flink-2.2/src/test/java/org/apache/fluss/flink/source/Flink22DeltaJoinITCase.java
```diff
@@ -0,0 +1,916 @@
...
+                ImmutableMap.of("table.delete.behavior", "IGNORE"));
+
+        String sinkTableName = "sink_table_force_fail";
+        createSink(
+                sinkTableName,
+                "a1 int, b1 varchar, c1 bigint, d1 int, e1 bigint, a2 int, b2 varchar, c2 bigint, d2 int, e2 bigint",
+                "c1, d1, c2, d2");
+
+        // Filter on e1 > e2, where e1 and e2 are NOT part of the upsert key
+        String sql =
```
wuchong, line 450: Add TODO comment above this 

```java
        // TODO we can add a UpsertFilterOperator that can convert the un-match-filter UPSERT record
        //  into DELETE record.
```

fluss-flink/fluss-flink-2.2/src/test/java/org/apache/fluss/flink/source/Flink22DeltaJoinITCase.java
```diff
@@ -0,0 +1,916 @@
...
+
+        String rightTableName = "right_table_materializer";
+        createSource(
+                rightTableName,
+                "a2 int, b2 varchar, c2 bigint, d2 int, e2 bigint",
+                "c2, d2",
+                "c2",
+                ImmutableMap.of("table.delete.behavior", "IGNORE"));
+
+        // Sink PK (a1, a2) doesn't match upstream update key (c1, d1, c2, d2)
```
wuchong, line 1030: Please add TODO on this 

```java
// TODO: this depends on Fluss supports MVCC/point-in-time lookup to support change upsert keys
```

fluss-flink/fluss-flink-2.2/src/test/java/org/apache/fluss/flink/source/Flink22DeltaJoinITCase.java
```diff
@@ -0,0 +1,916 @@
...
+
+        String rightTableName = "right_table_nondeterministic";
+        createSource(
+                rightTableName,
+                "a2 int, c2 bigint, d2 int",
+                "c2, d2",
+                "c2",
+                ImmutableMap.of("table.merge-engine", "first_row"));
+
+        String sinkTableName = "sink_table_nondeterministic";
```
wuchong, line 1069: Please add comment on this 

```java
        // TODO this should be supported in Flink in future for non-deterministic functions before sinking
```

fresh-borzoni:
@xuyangzhong Ack. Since you’re pushing new commits, let's continue with this PR.

fluss-flink/fluss-flink-2.2/src/test/java/org/apache/fluss/flink/source/Flink22DeltaJoinITCase.java
xuyangzhong: I have adapted most of the cases to prefix lookup join

wuchong (approved):

PR #2556, opened 2026-02-02
[FLUSS-2526][server] Fix accidental remote KV snapshot deletion during rebalance

### LINKED ISSUE #2526: Rebalance may cause kv snapshot deleted by accident
### Search before asking

### Fluss version

0.8.0 (latest release)

### Please describe the bug 🐞

Just got a kv snapshot not exists exception after rebanlace  primary key table, still not know why, got the following message:

Image

The current leader for the bucket is 10
I got message in tablet server 10 for rebalance:
```
2026-01-26 17:24:10,579 INFO  org.apache.fluss.server.replica.Replica                      [] - Shrink ISR From [1, 2, 0, 16, 5, 10] to [10] for bucket TableBucket{tableId=7, partitionId=124, bucket=2}. Leader: (high watermark: 7660, end offset: 7660, out of sync replicas: [0, 16, 1, 2, 5])
```

It must be some servers delete the kv snapshot for this bucket, so I search such message globally, I got the following message from the tablet server 1
```
content:2026-01-26 17:24:10,624 INFO org.apache.fluss.server.kv.KvManager                         [] - Delete table's remote bucket snapshot dir of TableBucket{tableId=7, partitionId=124, bucket=2} success.
```

Seem the tablet server 1 mis-delete the kv snapshot for the bucket.

cc @swuferhong 
 

### Solution

_No response_

ZuebeyirEser:
### Purpose

Linked issue: close #2526

This PR addresses the issue where rebalancing a primary key table leads to a `KvSnapshotNotExistException`. The problem was that when a bucket replica was moved during rebalancing, the tablet server being stopped was incorrectly deleting shared remote snapshots from the filesystem, which the new leader still needed to function.

### Brief change log

* Fixed `ReplicaManager#stopReplica` by removing the logic that triggered `kvManager.deleteRemoteKvSnapshot`.
* The previous implementation caused nodes to perform a global deletion of remote snapshots during a local replica cleanup if the `delete` flag was set.
* This ensures that rebalancing or moving replicas only affects local disk state. Remote snapshots are now correctly treated as part of the global table lifecycle rather than being managed by individual nodes during migration.

### Tests

* Added `KvSnapshotDeletionBugReplicationTest` to `fluss-server`.
* The test confirms that when a replica is dropped locally (simulating rebalance behavior), the remote snapshot directory and its metadata remain untouched.
* Verified that this prevents the `KvSnapshotNotExistException` previously seen by leaders after a rebalance.

### API and Format

no

### Documentation

no

### DIFF
--- a/fluss-server/src/test/java/org/apache/fluss/server/kv/KvSnapshotDeletionBugReplicationTest.java
+++ b/fluss-server/src/test/java/org/apache/fluss/server/kv/KvSnapshotDeletionBugReplicationTest.java
@@ -1,3 +1,20 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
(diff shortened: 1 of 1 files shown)


ZuebeyirEser:
@luoyuxia I actually also provided PR for this spesific flaky test as well. Please see: #2524  currently under review by @swuferhong 
FYI

Issue #2536, opened 2026-02-01
[PROPOSAL] Apache Iceberg V3 Support for Fluss


Shekharrajak:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/fluss/issues) and found nothing similar.


### Motivation

## Summary

This proposal introduces Iceberg V3 table format support in Fluss, replacing V2 equality deletes with **deletion vectors** (Roaring bitmaps) for 100-500x storage reduction and 5-10x faster CDC queries.

**Key V3 Features:**
- Deletion Vectors (binary bitmaps in Puffin files)
- New Data Types (Variant, Geometry, Geography, TimestampNano)
- Default Values (schema evolution without backfill)
- Row Lineage (`_row_id`, `_last_updated_sequence_number`)

## Why This Matters

| Metric | V2 (Current) | V3 (Proposed) |
|--------|--------------|---------------|
| Delete storage | 100 bytes/delete | ~1 byte/delete | 
| Query with 10K deletes | 500ms | 50ms | 
| Schema evolution | Full rewrite | Instant (defaults) | 


## Non-Goals (for initial implementation version)

- Variant query pushdown
- Geometry spatial indexing
- Automatic V2→V3 migration
- Cross-table transactions




### Solution


## Implementation Phases

1. Type System: Variant, Geometry, TimestampNano types 
2. Deletion Vectors:  DV writer, committer, merger  
3. Row Lineage:  _row_id assignment and tracking 

Testing & Docs along the way. We can have Github CI workflow setup to have integration testing with various system, query engine and environments.

##  Rough Components:                                                                                                                                                                                
                                                                                                                                                                                             
  1. V3DeltaTaskWriter                                                                                                                                                                       
     - writes data files + builds deletion vectors                                                                                                                                           
     - uses: DeletionVectorWriter                                                                                                                                                            
     - produces: IcebergWriteResultV3                                                                                                                                                        
                                                                                                                                                                                             
  2. DeletionVectorWriter                                                                                                                                                                    
     - serializes Roaring bitmap to Puffin file                                                                                                                                              
     - produces: DeletionVectorMetadata                                                                                                                                                      
                                                                                                                                                                                             
  3. IcebergWriteResultV3                                                                                                                                                                    
     - contains: dataFiles, deletionVectors, firstRowId, addedRows                                                                                                                           
                                                                                                                                                                                             
  4. DeletionVectorMetadata                                                                                                                                                                  
[...]

PR #2561, opened 2026-02-03
feat: Iceberg v3 deletion vectors

Shekharrajak:
### Purpose                                                                                                                                                                                
                                                                                                                                                                                             
  <!-- Linking this pull request to the issue -->                                                                                                                                            
  Linked issue:   https://github.com/apache/fluss/issues/2536                                                                                                                                                              
                                                                                                                                                                                             
  <!-- What is the purpose of the change -->                                                                                                                                                 
  This PR adds foundational support for Iceberg V3 deletion vectors in Fluss. Deletion vectors (DVs) are a V3 feature that provides more efficient row-level deletes by storing deleted row  
  positions in a compact bitmap format (Puffin files) rather than separate position delete files.                                                                                            
                                                                                                                                                                                             
  ### Brief change log                                                                                                                                                                       
                                                                                                                                                                                             
  <!-- Please describe the changes made in this pull request and explain how they address the issue -->                                                                                      
                                                                                                                                                                                             
  - **Add FormatVersionManager**: A utility class for detecting Iceberg table format versions and gating V3-specific features. 
                                            
                                                                                                                                                                                             
  - **Add V3DeltaTaskWriter**: A V3-aware delta writer that uses Iceberg's built-in `BaseDVFileWriter` for position deletes when the table format version supports deletion vectors.         
    - For V3 tables: Uses Puffin-based deletion vectors for position deletes                                                                                                                 
    - For V2 tables: Falls back to traditional position delete files                                                                                                                         
    - Extends `BaseTaskWriter` with DV support detection 

 ### API and Format                                                                                                                                                                         
                                                                                                                                                                                             
  <!-- Does this change affect API or storage format -->                                                                                                                                     
  No breaking API changes. This adds new internal classes:                                                                                                                                   
  - `FormatVersionManager` - New utility class for format version detection                                                                                                                  
  - `V3DeltaTaskWriter` - New writer class                                                                                                    
                                                                                                                                                                                             
  When V3 tables are used, delete files will use Puffin format (`.puffin`) instead of traditional position delete files, which is the standard Iceberg V3 format.                            

  ### Tests  

Unit tests

### DIFF
--- a/fluss-lake/fluss-lake-iceberg/src/test/java/org/apache/fluss/lake/iceberg/tiering/writer/V3DeltaTaskWriterTest.java
+++ b/fluss-lake/fluss-lake-iceberg/src/test/java/org/apache/fluss/lake/iceberg/tiering/writer/V3DeltaTaskWriterTest.java
@@ -239,10 +239,10 @@ class V3DeltaTaskWriterTest {
         Table table = createTableWithFormatVersion(3);
         V3DeltaTaskWriter writer = createWriter(table);
 
-        // Write same key 3 times
-        writer.write(createRecord(table.schema(), 1L, "v1")); // position 0 -> deleted
-        writer.write(createRecord(table.schema(), 1L, "v2")); // position 1 -> deleted
-        writer.write(createRecord(table.schema(), 1L, "v3")); // position 2 -> live
+        // Write same key 3 times - each new write deletes the previous position
+        writer.write(createRecord(table.schema(), 1L, "v1")); // pos 0, no delete yet
+        writer.write(createRecord(table.schema(), 1L, "v2")); // pos 1, deletes pos 0
+        writer.write(createRecord(table.schema(), 1L, "v3")); // pos 2, deletes pos 1 (live)
 
         WriteResult result = writer.complete();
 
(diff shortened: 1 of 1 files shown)


fluss-lake/fluss-lake-iceberg/src/main/java/org/apache/fluss/lake/iceberg/version/FormatVersionManager.java
```diff
@@ -0,0 +1,49 @@
...
+
+    public static final int DEFAULT_FORMAT_VERSION = 2;
+    public static final int V3_FORMAT_VERSION = 3;
+    public static final int MIN_VERSION_FOR_DELETION_VECTORS = 3;
+
+    private FormatVersionManager() {}
+
+    /** Detects format version from table metadata (not properties, as it's reserved). */
+    public static int detectFormatVersion(Table table) {
+        if (table instanceof HasTableOperations) {
```
nastra, line 34: this won't reliably work across all cases unfortunately. However, we have `TableUtil.formatVersion` in Iceberg, which you should be able to use: https://github.com/apache/iceberg/blob/829ae7a11dc1eb62246c801ce1c7e501356c5463/core/src/main/java/org/apache/iceberg/TableUtil.java#L27

MehulBatra: Hi @nastra this PR is drafted/closed so I would suggest not to review, as we will be creating a FIP first and getting approach consensus then moving forward with the development.
Thank you!

PR #2565, opened 2026-02-03
[docs] binlog virtual table 

MehulBatra:
### Brief change log

- Gives info to the user about what a binlog table is and how it can be used & accessed.
- Modified the changetype info for changelog

### DIFF
--- a/website/docs/table-design/virtual-tables.md
+++ b/website/docs/table-design/virtual-tables.md
@@ -91,20 +91,43 @@ SELECT * FROM orders$changelog;
 Output:
 
 ```
-+----+---------------+-------------+---------------------+----------+---------------+---------+
-| op | _change_type  | _log_offset | _commit_timestamp   | order_id | customer_name | amount  |
-+----+---------------+-------------+---------------------+----------+---------------+---------+
-| +I | insert        |           0 | 2024-01-15 10:30:00 |        1 | Rhea          |  100.00 |
-| +I | update_before |           1 | 2024-01-15 10:35:00 |        1 | Rhea          |  100.00 |
-| +I | update_after  |           2 | 2024-01-15 10:35:00 |        1 | Rhea          |  150.00 |
-| +I | delete        |           3 | 2024-01-15 10:40:00 |        1 | Rhea          |  150.00 |
-+----+---------------+-------------+---------------------+----------+---------------+---------+
++---------------+-------------+---------------------+----------+---------------+---------+
+| _change_type  | _log_offset | _commit_timestamp   | order_id | customer_name | amount  |
++---------------+-------------+---------------------+----------+---------------+---------+
+| insert        |           0 | 2024-01-15 10:30:00 |        1 | Rhea          |  100.00 |
+| update_before |           1 | 2024-01-15 10:35:00 |        1 | Rhea          |  100.00 |
+| update_after  |           2 | 2024-01-15 10:35:00 |        1 | Rhea          |  150.00 |
+| delete        |           3 | 2024-01-15 10:40:00 |        1 | Rhea          |  150.00 |
++---------------+-------------+---------------------+----------+---------------+---------+
 ```
 
-:::note
-The `op` column is Flink's row kind indicator. For changelog virtual tables, all rows are emitted as `+I` (insert) to the downstream, while the actual change type is captured in the `_change_type` column.
-:::
 
+### Startup Modes
+
+
+| Mode | Description |
+|------|-------------|
+| `earliest` | Start reading from the beginning of the log |
+| `latest` | Start reading from the current end of the log (only new changes) |
+| `timestamp` | Start reading from a specific timestamp (milliseconds since epoch) |
+
+
+The changelog table supports different startup modes to control where reading begins:
+
+```sql title="Flink SQL"
+-- Read from the beginning (default)
+SELECT * FROM orders$changelog /*+ OPTIONS('scan.startup.mode' = 'earliest') */;
+
+-- Read only new changes from now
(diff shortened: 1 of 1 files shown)


website/docs/table-design/virtual-tables.md
```diff
@@ -55,23 +49,21 @@ For Primary Key Tables, the following change types are supported:
 
 | Change Type | Description |
 |-------------|-------------|
-| `+I` | **Insert** - A new row was inserted |
-| `-U` | **Update Before** - The previous value of an updated row (retraction) |
-| `+U` | **Update After** - The new value of an updated row |
-| `-D` | **Delete** - A row was deleted |
+| `insert` | A new row was inserted |
```
polyzos, line 52: maybe use `op` and `change column` columns? and make them uppercase.
This also helps with consistency  throughout the doc

website/docs/table-design/virtual-tables.md
```diff
@@ -55,23 +49,21 @@ For Primary Key Tables, the following change types are supported:
...
+| `delete` | A row was deleted |
 
 #### Log Tables
 
 For Log Tables (append-only), only one change type is used:
 
 | Change Type | Description |
 |-------------|-------------|
-| `+A` | **Append** - A new row was appended to the log |
+| `insert` | A new row was inserted into the log |
```
polyzos, line 63: ditto

website/docs/table-design/virtual-tables.md
```diff
@@ -99,70 +91,128 @@ SELECT * FROM orders$changelog;
...
 
-| Mode | Description |
-|------|-------------|
-| `earliest` | Start reading from the beginning of the log |
-| `latest` | Start reading from the current end of the log (only new changes) |
-| `timestamp` | Start reading from a specific timestamp (milliseconds since epoch) |
 
 ### Limitations
-- Projection & partition & predicate pushdowns are not supported yet. This will be addressed in future releases.
+- Projection & partition & predicate pushdowns are not supported yet.
```
polyzos: Projections, partition pruning, and predicate pushdowns are not supported yet.

wuchong: We should add the limitation for `$changelog` table as well. Currently, this is under the section of `$binlog` table.

website/docs/table-design/virtual-tables.md
```diff
@@ -99,70 +91,128 @@ SELECT * FROM orders$changelog;
...
+SELECT
+    _change_type,
+    _commit_timestamp,
+    `before`.name AS old_name,
+    `after`.name AS new_name
+FROM users$binlog
+WHERE _change_type = 'update';
 ```
 
 ### Startup Modes
```
wuchong, line 216: We should add "Startup Modes` section for `$changelog` table as well. Currently, this is under the section of `$binlog` table.

website/docs/table-design/virtual-tables.md
```diff
@@ -99,70 +91,128 @@ SELECT * FROM orders$changelog;
...
-The changelog virtual table supports different startup modes to control where reading begins:
+
+| Mode | Description |
+|------|-------------|
+| `earliest` | Start reading from the beginning of the log |
+| `latest` | Start reading from the current end of the log (only new changes) |
+| `timestamp` | Start reading from a specific timestamp (milliseconds since epoch) |
+
+
+The virtual table supports different startup modes to control where reading begins:
```
wuchong: Not all virtual table will support startup modes. We will also include `$lake` (and more in the future) into this page.

website/docs/table-design/virtual-tables.md
```diff
@@ -99,70 +91,128 @@ SELECT * FROM orders$changelog;
...
-| +I | +U           |           2 | 2024-01-15 10:35:00 |        1 | Rhea          |  150.00 |
-| +I | -D           |           3 | 2024-01-15 10:40:00 |        1 | Rhea          |  150.00 |
-+----+--------------+-------------+---------------------+----------+---------------+---------+
++----+---------------+-------------+---------------------+----------+---------------+---------+
+| op | _change_type  | _log_offset | _commit_timestamp   | order_id | customer_name | amount  |
++----+---------------+-------------+---------------------+----------+---------------+---------+
+| +I | insert        |           0 | 2024-01-15 10:30:00 |        1 | Rhea          |  100.00 |
+| +I | update_before |           1 | 2024-01-15 10:35:00 |        1 | Rhea          |  100.00 |
+| +I | update_after  |           2 | 2024-01-15 10:35:00 |        1 | Rhea          |  150.00 |
+| +I | delete        |           3 | 2024-01-15 10:40:00 |        1 | Rhea          |  150.00 |
```
wuchong: Can we remove the `op` column from the output? This is very confusing we carry an additional column `op` in the schema of `$binlog` or `$changelog` table. 

Flink SQL CLI supports options to omit the `op` column 
https://nightlies.apache.org/flink/flink-docs-master/docs/dev/table/sqlclient/#sql-client-result-modes

website/docs/table-design/virtual-tables.md
MehulBatra: Removing the op column and keeping the change_type column values small because it will originally be generated in smallcase, so to make it real and relatble not keeping it uppercase.

website/docs/table-design/virtual-tables.md
MehulBatra: Yes, they aren't. We will work on this in 0.10, currently you can select specific columns via flink sql level but that will not reflect in storage level, hence no projection pushdown aswell.

website/docs/table-design/virtual-tables.md
MehulBatra: Agreed, removed.

PR #2573, opened 2026-02-04
implement lookup insert if not exists on the client side

### LINKED ISSUE #2591: [kv] support lookup insertIfNotExists on Java client
### Search before asking

### Description

Support lookup insertIfNotExists on Java client

### Willingness to contribute

xx789633:
### Purpose

Linked issue: close #2591

### DIFF
--- a/fluss-client/src/main/java/org/apache/fluss/client/lookup/Lookup.java
+++ b/fluss-client/src/main/java/org/apache/fluss/client/lookup/Lookup.java
@@ -95,6 +95,16 @@ public interface Lookup {
     }
 
     /**
+     * Enables insert-if-not-exists behavior for the lookup operation. When enabled, if a lookup
+     * does not find a matching row, a new row will be inserted with the lookup key values. This
+     * feature cannot be used on tables that contain non-nullable columns other than the primary key
+     * or auto-increment columns.
+     *
+     * @return a new Lookup instance with insert-if-not-exists enabled
+     */
+    Lookup enableInsertIfNotExists();
+
+    /**
      * Creates a {@link Lookuper} instance to lookup rows of a primary key table by the specified
      * lookup columns. By default, the lookup columns are the primary key columns, but can be
      * changed with ({@link #lookupBy(List)}) method.
@@ -111,14 +121,4 @@ public interface Lookup {
      * @return the typed lookuper
      */
     <T> TypedLookuper<T> createTypedLookuper(Class<T> pojoClass);
-
-    /**
-     * Enables insert-if-not-exists behavior for the lookup operation. When enabled, if a lookup
-     * does not find a matching row, a new row will be inserted with the lookup key values. This
-     * feature cannot be used on tables that contain non-nullable columns other than the primary key
-     * or auto-increment columns.
-     *
-     * @return a new Lookup instance with insert-if-not-exists enabled
-     */
-    Lookup enableInsertIfNotExists();
 }
--- a/fluss-client/src/main/java/org/apache/fluss/client/lookup/TableLookup.java
+++ b/fluss-client/src/main/java/org/apache/fluss/client/lookup/TableLookup.java
@@ -18,12 +18,15 @@
 package org.apache.fluss.client.lookup;
 
 import org.apache.fluss.client.metadata.MetadataUpdater;
+import org.apache.fluss.metadata.Schema;
 import org.apache.fluss.metadata.SchemaGetter;
 import org.apache.fluss.metadata.TableInfo;
 
 import javax.annotation.Nullable;
 
+import java.util.HashSet;
 import java.util.List;
+import java.util.stream.Collectors;
 
 /** API for configuring and creating {@link Lookuper}. */
 public class TableLookup implements Lookup {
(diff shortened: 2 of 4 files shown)


fluss-client/src/test/java/org/apache/fluss/client/table/FlussTableITCase.java
```diff
@@ -633,6 +633,84 @@ void testPutAutoIncrementColumnAndLookup() throws Exception {
...
+                        .withComment("a is first column")
+                        .column("b", DataTypes.INT())
+                        .withComment("b is second column")
+                        .primaryKey("b")
+                        .column("c", DataTypes.STRING())
+                        .enableAutoIncrement("a")
+                        .build();
+        TableDescriptor tableDescriptor = TableDescriptor.builder().schema(schema).build();
+        createTable(tablePath, tableDescriptor, false);
+        Table invalid_table = conn.getTable(tablePath);
```
beryllw: Error:  src/test/java/org/apache/fluss/client/table/FlussTableITCase.java:[651,15] (naming) LocalVariableName: Name 'invalid_table' must match pattern '^[a-z][a-zA-Z0-9]*$'.

fluss-client/src/main/java/org/apache/fluss/client/utils/ClientRpcMessageUtils.java
beryllw, line 185: May be two functions `makeLookupRequestWithInsertIfNotExists` and `makeLookupRequest` can be merged.

xx789633: comments addressed. please take a look.

fluss-client/src/main/java/org/apache/fluss/client/lookup/Lookup.java
```diff
@@ -111,4 +111,12 @@ default Lookup lookupBy(String... lookupColumnNames) {
...
      */
     <T> TypedLookuper<T> createTypedLookuper(Class<T> pojoClass);
+
+    /**
+     * Enables insert-if-not-exists behavior for the lookup operation. When enabled, if a lookup
+     * does not find a matching row, a new row will be inserted with the lookup key values.
+     *
+     * @return a new Lookup instance with insert-if-not-exists enabled
+     */
+    Lookup enableInsertIfNotExists();
```
beryllw: `Lookup.enableInsertIfNotExists()` doesn't document:
Auto-increment column handling, Nullable column restriction.

fluss-client/src/main/java/org/apache/fluss/client/lookup/TableLookup.java
```diff
@@ -49,23 +51,71 @@ private TableLookup(
...
+                .filter(column -> column.getDataType().isNullable())
+                .filter(column -> !tableInfo.getPrimaryKeys().contains(column.getName()))
+                .anyMatch(
+                        column ->
+                                !tableInfo
+                                        .getSchema()
+                                        .getAutoIncrementColumnNames()
+                                        .contains(column.getName()))) {
+            throw new IllegalArgumentException(
+                    "insertIfNotExists cannot be enabled for tables with nullable columns besides primary key and auto increment columns.");
```
beryllw: InsertIfNotExists requires all non-primary-key columns (except auto-increment) to be NULL.

fluss-client/src/test/java/org/apache/fluss/client/table/FlussTableITCase.java
```diff
@@ -633,6 +633,85 @@ void testPutAutoIncrementColumnAndLookup() throws Exception {
...
+
+        assertThatThrownBy(
+                        () ->
+                                table.newLookup()
+                                        .lookupBy("b")
+                                        .enableInsertIfNotExists()
+                                        .createLookuper())
+                .isInstanceOf(IllegalArgumentException.class)
+                .hasMessageContaining("insertIfNotExists can not be used with prefix lookup");
+        assertThatThrownBy(
```
platinumhamburg: This is a duplicate assertion. Moreover, the test schema contains only one primary key column, 'b', so creating a lookuper with lookupBy("b") should not result in a PrefixKeyLookuper. This issue was not introduced by this PR; rather, the test has exposed an incompleteness in the semantic implementation of TableLookup.

fluss-client/src/main/java/org/apache/fluss/client/lookup/TableLookup.java
```diff
@@ -49,23 +51,71 @@ private TableLookup(
...
         this.schemaGetter = schemaGetter;
         this.metadataUpdater = metadataUpdater;
         this.lookupClient = lookupClient;
         this.lookupColumnNames = lookupColumnNames;
+        this.insertIfNotExists = insertIfNotExists;
+    }
+
+    @Override
+    public Lookup enableInsertIfNotExists() {
+        if (lookupColumnNames != null) {
```
platinumhamburg: This check can be uniformly moved to createLookuper().

fluss-client/src/main/java/org/apache/fluss/client/lookup/TableLookup.java
```diff
@@ -49,23 +51,71 @@ private TableLookup(
...
+            throw new IllegalArgumentException(
+                    "insertIfNotExists cannot be enabled for tables with non-nullable columns besides primary key and auto increment columns.");
+        }
+
+        return new TableLookup(tableInfo, schemaGetter, metadataUpdater, lookupClient, true);
     }
 
     @Override
     public Lookup lookupBy(List<String> lookupColumnNames) {
+        if (insertIfNotExists) {
```
platinumhamburg: Ditto, this check can be uniformly moved to createLookuper().

fluss-client/src/main/java/org/apache/fluss/client/lookup/LookupType.java
```diff
@@ -23,5 +23,6 @@
 @Internal
 public enum LookupType {
     LOOKUP,
+    LOOKUP_WITH_INSERT_IF_NOT_EXISTS,
```
platinumhamburg, line 26: I think there's no need to extend the enum definition here.

platinumhamburg (changes requested):
LGTM, just a few minor issues.

fluss-client/src/main/java/org/apache/fluss/client/lookup/LookupType.java
xx789633: We need to group the different lookups in the lookupQueue into different batches in https://github.com/apache/fluss/blob/main/fluss-client/src/main/java/org/apache/fluss/client/lookup/LookupSender.java#L152 so I introduced this new type

fluss-client/src/main/java/org/apache/fluss/client/lookup/TableLookup.java
xx789633: good suggestion. done.

fluss-client/src/test/java/org/apache/fluss/client/table/FlussTableITCase.java
xx789633: updated.

fluss-client/src/main/java/org/apache/fluss/client/lookup/TableLookup.java
```diff
@@ -35,37 +35,61 @@ public class TableLookup implements Lookup {
...
+    @Override
+    public Lookup enableInsertIfNotExists() {
+        return new TableLookup(tableInfo, schemaGetter, metadataUpdater, lookupClient, null, true);
     }
 
     @Override
     public Lookup lookupBy(List<String> lookupColumnNames) {
         return new TableLookup(
-                tableInfo, schemaGetter, metadataUpdater, lookupClient, lookupColumnNames);
+                tableInfo, schemaGetter, metadataUpdater, lookupClient, lookupColumnNames, false);
```
wuchong: We should copy with the existing `insertIfNotExists`, rather than override it to `false`.

fluss-client/src/main/java/org/apache/fluss/client/lookup/TableLookup.java
```diff
@@ -35,37 +35,61 @@ public class TableLookup implements Lookup {
...
         this.schemaGetter = schemaGetter;
         this.metadataUpdater = metadataUpdater;
         this.lookupClient = lookupClient;
         this.lookupColumnNames = lookupColumnNames;
+        this.insertIfNotExists = insertIfNotExists;
+    }
+
+    @Override
+    public Lookup enableInsertIfNotExists() {
+        return new TableLookup(tableInfo, schemaGetter, metadataUpdater, lookupClient, null, true);
```
wuchong: We should copy with the existing `lookupColumnNames`, rather than override it to `null`.

wuchong (approved):
Thanks @xx789633 , the implemenation looks clean and good to me. I only left 2 minor comments about the copying problem. I have appended a commit to fix it. So I will merge it once CI is passed.

PR #2577, opened 2026-02-05
[hotfix][dist] Don't print jindo log when deploy on console mode

swuferhong:
### Purpose

Linked issue: close #xxx

Don't print jindo log when deploy on console mode.

### DIFF
--- a/fluss-dist/src/main/resources/conf/log4j-console.properties
+++ b/fluss-dist/src/main/resources/conf/log4j-console.properties
@@ -57,3 +57,9 @@ appender.rolling.strategy.max=${env:MAX_LOG_FILE_NUMBER:-10}
 # Suppress the irrelevant (wrong) warnings from the Netty channel handler
 logger.netty.name=org.jboss.netty.channel.DefaultChannelPipeline
 logger.netty.level=OFF
+# don't print jindo log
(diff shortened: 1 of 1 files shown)


fluss-dist/src/main/resources/conf/log4j-console.properties
```diff
@@ -57,3 +57,9 @@ appender.rolling.strategy.max=${env:MAX_LOG_FILE_NUMBER:-10}
 # Suppress the irrelevant (wrong) warnings from the Netty channel handler
 logger.netty.name=org.jboss.netty.channel.DefaultChannelPipeline
 logger.netty.level=OFF
+# don't print jindo log
+logger.jindo_fsStats.name=com.aliyun.jindodata.common.FsStats
+logger.jindo_fsStats.level=OFF
+logger.jindo_jnative.name=com.aliyun.jindodata.jnative.NativeLogger
+logger.jindo_jnative.level=OFF
```
wuchong, line 64: @swuferhong @luoyuxia This looks odd because the open-source Fluss distribution doesn't include any Jindo dependencies, yet the log4j configuration is disabling Jindo's logging. Such changes should be kept internal—for example, in `fluss-ali/fluss-ali-dist`. Moreover, we may need to add logging configurations for other internal dependencies in the future as well.

Issue #2293, opened 2025-12-31
Partial Update use column id rather column position to update data.


loserwang1024:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/fluss/issues) and found nothing similar.

### Description

Currently, we switch from column position to column id  for projection push down, however, in partial Update, we still use column position.

### Willingness to contribute

- [ ] I'm willing to submit a PR!

Priyamanjare54:
I'd like to work on this. Can you assign it to me?

loserwang1024:
@Priyamanjare54 Yes, you can do it. You can refer what scan push-down does now.

Priyamanjare54:
Hi @loserwang1024,
I need to step back from this issue for now due to time constraints. Please feel free to reassign it so it doesn’t block progress.

Thanks for understanding!

loserwang1024:
@Priyamanjare54 ok, we will still use position after I add id to nested rows. And I will work on more compatible way to add a option to control id or position.

fresh-borzoni:
@loserwang1024 I created PR, I hope I got the intention right 🙏

PR #2594, opened 2026-02-07
[BugFix] Fix ArrayIndexOutOfBoundsException in partial updates after ADD COLUMN

### LINKED ISSUE #2988: ArrayIndexOutOfBoundsException in partial updates after ADD COLUMN
### Search before asking

### Fluss version

0.9.0 (latest release)

### Please describe the bug 🐞

After ADD COLUMN, partial updates on existing rows crash with ArrayIndexOutOfBoundsException. ThePartialUpdater is created with the latest schema (4 fields) but reads old rows from RocksDB that were stored with the previous schema (3 fields). At field index 3, GenericRow.fields[3] throws on a 3-element array.

Reproduce: insert a row, ADD COLUMN, partial update the same row - crash.

### Solution

Use column IDs to align rows to the latest schema before the merger sees them

fresh-borzoni:
## Summary
closes #2988
relates to #2293

Partial updates crash with ArrayIndexOutOfBoundsException after ADD COLUMN, the PartialUpdater iterates over the latest schema's field count but old rows in RocksDB have fewer fields. 

The fix re-encodes rows to the latest schema layout using column IDs before the merger sees them, filling new columns with null.

Target column positions and incoming rows are also aligned via column IDs to support future schema operations beyond ADD COLUMN LAST

### DIFF
--- a/fluss-server/src/main/java/org/apache/fluss/server/kv/KvTablet.java
+++ b/fluss-server/src/main/java/org/apache/fluss/server/kv/KvTablet.java
@@ -729,6 +729,7 @@ public final class KvTablet {
      * schema using column IDs. This is needed when a client sends a partial update using an older
      * schema whose positions need to be remapped to the latest schema layout.
      */
+    @VisibleForTesting
     static int[] convertTargetColumns(int[] positions, Schema sourceSchema, Schema targetSchema) {
         List<Integer> sourceColIds = sourceSchema.getColumnIds();
         List<Integer> targetColIds = targetSchema.getColumnIds();


fresh-borzoni:
@loserwang1024 @luoyuxia 
PTAL 🙏

fluss-server/src/main/java/org/apache/fluss/server/kv/KvTablet.java
```diff
@@ -582,6 +618,74 @@ private long applyUpdate(
...
+            BinaryValue value, short latestSchemaId, Schema latestSchema) {
+        if (value.schemaId == latestSchemaId) {
+            return value;
+        }
+
+        Schema sourceSchema = schemaGetter.getSchema(value.schemaId);
+        List<Integer> sourceColIds = sourceSchema.getColumnIds();
+        List<Integer> targetColIds = latestSchema.getColumnIds();
+
+        Map<Integer, Integer> sourceIdToPos = new HashMap<>();
```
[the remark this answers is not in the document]
fresh-borzoni: The allocations only fire during schema transitions and are dominated by the RocksDB point lookup on the same path, not worth the complexity.

fresh-borzoni:
Answered copilot comment

fluss-server/src/main/java/org/apache/fluss/server/kv/KvTablet.java
```diff
@@ -582,6 +618,74 @@ private long applyUpdate(
...
+        for (int i = 0; i < sourceColIds.size(); i++) {
+            sourceIdToPos.put(sourceColIds.get(i), i);
+        }
+
+        InternalRow.FieldGetter[] sourceGetters =
+                InternalRow.createFieldGetters(sourceSchema.getRowType());
+        RowEncoder encoder = RowEncoder.create(kvFormat, latestSchema.getRowType());
```
polyzos: RowEncoder implements AutoCloseable but is never closed. Wrap in try-with-resources to avoid a resource leak on every aligned record.

fluss-server/src/main/java/org/apache/fluss/server/kv/KvTablet.java
```diff
@@ -582,6 +618,74 @@ private long applyUpdate(
...
+        Schema sourceSchema = schemaGetter.getSchema(value.schemaId);
+        List<Integer> sourceColIds = sourceSchema.getColumnIds();
+        List<Integer> targetColIds = latestSchema.getColumnIds();
+
+        Map<Integer, Integer> sourceIdToPos = new HashMap<>();
+        for (int i = 0; i < sourceColIds.size(); i++) {
+            sourceIdToPos.put(sourceColIds.get(i), i);
+        }
+
+        InternalRow.FieldGetter[] sourceGetters =
```
polyzos: createFieldGetters and RowEncoder.create are called once per record, but for a batch, they are constant across all records sharing the same source schema. Maybe we can consider constructing them once per batch?

fresh-borzoni:
@polyzos Thank you for the review
Addressed comments, PTAL 🙏

polyzos:
LGTM.. thank you for your contribution

fluss-server/src/main/java/org/apache/fluss/server/kv/KvTablet.java
```diff
@@ -628,6 +661,102 @@ private long applyUpdate(
...
+                // Column added after the source schema — fill with null.
+                ctx.encoder.encodeField(targetPos, null);
+            } else {
+                ctx.encoder.encodeField(
+                        targetPos, mapping.getters[sourcePos].getFieldOrNull(value.row));
+            }
+        }
+        // copy() is required: the encoder reuses its internal buffer, so the next
+        // startNewRow() would overwrite the row returned here.
+        return new BinaryValue(ctx.latestSchemaId, ctx.encoder.finishRow().copy());
```
wuchong, line 724: After schema evolution, every `oldValue` read from KV store with an older schema goes through `alignToLatestSchema`, which does a full field-by-field re-encode + `copy()`. The previous approach used `PaddingRow` as a zero-copy virtual view. This introduces a significant per-record overhead during the (potentially long) transition period when old-schema data still dominates the KV store.

fluss-server/src/main/java/org/apache/fluss/server/kv/KvTablet.java
```diff
@@ -473,39 +488,50 @@ private void processKvRecords(
                 KvRecordReadContext.createReadContext(kvFormat, schemaGetter);
         ValueDecoder valueDecoder = new ValueDecoder(schemaGetter, kvFormat);
 
-        for (KvRecord kvRecord : kvRecords.records(readContext)) {
-            byte[] keyBytes = BytesUtils.toArray(kvRecord.getKey());
-            KvPreWriteBuffer.Key key = KvPreWriteBuffer.Key.of(keyBytes);
-            BinaryRow row = kvRecord.getRow();
-            BinaryValue currentValue = row == null ? null : new BinaryValue(schemaIdOfNewData, row);
+        try (SchemaAlignmentContext alignmentContext =
+                new SchemaAlignmentContext(latestSchemaId, latestSchema, kvFormat)) {
```
wuchong, line 492: `SchemaAlignmentContext` is created unconditionally on every `processKvRecords` call, even when `schemaIdOfNewData == latestSchemaId` (the common case — no schema evolution). This allocates a `RowEncoder` + `HashMap` on the hot path for every batch. Which will introduce performance regression.

fresh-borzoni:
@wuchong Good judgement, thank you,  I prepared new PR: https://github.com/apache/fluss/pull/3000

PR #2617, opened 2026-02-09
Forbid aggregate merge engine with WAL changelog image

### LINKED ISSUE #2608: Forbid aggregate merge engine tables with WAL changelog image mode
### Search before asking

### Description

The UNDO recovery mechanism for aggregate merge engine tables relies on the full changelog of the table to correctly compute undo operations during failure recovery.

However, when table.changelog.image=WAL is configured, the changelog does not contain full row data, which will cause incorrect UNDO recovery results for aggregate merge engine tables.

We need to add a validation check to prevent this incompatible configuration combination.

### Willingness to contribute

platinumhamburg:
WAL mode lacks UPDATE_BEFORE records needed for UNDO recovery in aggregate tables. Add validation to reject this invalid combination.

### Purpose

Linked issue: close #2608

### DIFF
--- a/fluss-client/src/test/java/org/apache/fluss/client/admin/FlussAdminITCase.java
+++ b/fluss-client/src/test/java/org/apache/fluss/client/admin/FlussAdminITCase.java
@@ -1740,8 +1740,10 @@ class FlussAdminITCase extends ClientToServerITCaseBase {
         assertThatThrownBy(() -> admin.createTable(tablePath1, tableDescriptor1, false).get())
                 .cause()
                 .isInstanceOf(InvalidConfigException.class)
-                .hasMessageContaining(ConfigOptions.TABLE_MERGE_ENGINE.key())
-                .hasMessageContaining(ConfigOptions.TABLE_CHANGELOG_IMAGE.key());
+                .hasMessageContaining(
+                        "Table with 'AGGREGATION' merge engine does not support 'WAL' changelog image mode. "
+                                + "Aggregation merge engine tables require FULL changelog image mode for correct UNDO recovery. "
+                                + "Please set 'table.changelog.image' to 'FULL' or remove the setting.");
 
         // Test 2: Aggregate + FULL should be allowed
         TablePath tablePath2 = TablePath.of("fluss", "test_aggregate_full_changelog_allowed");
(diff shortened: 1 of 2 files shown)


fluss-server/src/main/java/org/apache/fluss/server/utils/TableDescriptorValidation.java
```diff
@@ -302,6 +303,18 @@ private static void checkMergeEngine(
                                     versionColumn.get(), columnType));
                 }
             } else if (mergeEngine == MergeEngineType.AGGREGATION) {
+                // Check aggregate merge engine with WAL changelog image
+                ChangelogImage changelogImage = tableConf.get(ConfigOptions.TABLE_CHANGELOG_IMAGE);
+                if (changelogImage == ChangelogImage.WAL) {
+                    throw new InvalidConfigException(
+                            String.format(
+                                    "Table with '%s' merge engine does not support '%s' changelog image mode. "
```
wuchong: I think we can hardcode the merge engine type and changelog image mode here, rather than print the config key?

fluss-client/src/test/java/org/apache/fluss/client/admin/FlussAdminITCase.java
```diff
@@ -1710,4 +1710,108 @@ void testAlterTableTieredLogLocalSegments() throws Exception {
...
+                TableDescriptor.builder()
+                        .schema(aggregateSchema)
+                        .comment("aggregate merge engine table with WAL changelog")
+                        .properties(properties1)
+                        .build();
+        assertThatThrownBy(() -> admin.createTable(tablePath1, tableDescriptor1, false).get())
+                .cause()
+                .isInstanceOf(InvalidConfigException.class)
+                .hasMessageContaining(ConfigOptions.TABLE_MERGE_ENGINE.key())
+                .hasMessageContaining(ConfigOptions.TABLE_CHANGELOG_IMAGE.key());
```
wuchong, line 1744: Actually, I prefer to assert the **full error message** to ensure that the error we throw is both **user-friendly and correctly formatted**. Otherwise, we’re not fully validating the exception message, which could allow unclear or malformed errors to go unnoticed.

platinumhamburg:
> I pushed a commit to fix the error message. Please take a look.

This refactoring is indeed clearer.

Issue #2619, opened 2026-02-09
Iceberg s3 table tearing bugs


bakjos:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/fluss/issues) and found nothing similar.

### Fluss version

main (development)

### Please describe the bug 🐞

I have been testing Fluss using iceberg as the dalatake with S3 tables and found the following issues:

1. ClassCastException on Select: While running a SELECT count(*) FROM table, I encountered the following exception:

```
java.lang.ClassCastException: class java.time.LocalDate cannot be cast to class java.lang.Integer...
```

2. Performance degradation: Query performance for the data lake table is very slow compared to querying remote storage. It appears that the table metadata is being queried for every split.

3. Configuration issue with TTL: The documentation states that we can use -1 for table.log.ttl, but setting this value throws an InvalidConfigException:

```
org.apache.fluss.exception.InvalidConfigException: Invalid value for config 'table.log.ttl'...
```

### Solution

I've send the PRs to improve/fix the 3 issues

### Are you willing to submit a PR?

- [x] I'm willing to submit a PR!

PR #2620, opened 2026-02-09
[lake/iceberg] Fix local date and time class cast

bakjos:
### Purpose

Fixes ClassCastException on Select: While running a SELECT count(*) FROM table when using iceberg as datalake with s3 tables as the catalog

Linked issue: close https://github.com/apache/fluss/issues/2619


luoyuxia:
@bakjos Thanks for the pr

PR #2622, opened 2026-02-09
[lake/iceberg] Add rest catalog cache

bakjos:
### Purpose

Query performance for the data lake table is very slow compared to querying remote storage. Add per-task lazy caching of Iceberg Catalog and Table inside IcebergLakeSource so that createRecordReader reuses one loadTable for all lake splits in a Flink source task, eliminating O(splits) REST round-trips when using a REST catalog.

Before: N splits → N × (createCatalog + loadTable) → N REST calls per task.
After: N splits → 1 × (createCatalog + loadTable) on first split, then N-1 reuses → 1 REST loadTable per task. With TTL enabled, the cache is refreshed after the TTL period so externally changed table metadata is picked up.

Linked issue: close https://github.com/apache/fluss/issues/2619

### DIFF
--- a/fluss-lake/fluss-lake-iceberg/src/main/java/org/apache/fluss/lake/iceberg/source/IcebergLakeSource.java
+++ b/fluss-lake/fluss-lake-iceberg/src/main/java/org/apache/fluss/lake/iceberg/source/IcebergLakeSource.java
@@ -46,10 +46,27 @@ import static org.apache.fluss.lake.iceberg.utils.IcebergConversions.toIceberg;
 /** Iceberg lake source. */
 public class IcebergLakeSource implements LakeSource<IcebergSplit> {
     private static final long serialVersionUID = 1L;
+
+    /**
+     * Config key for table cache TTL in milliseconds. After this duration, the cached table is
+     * reloaded on next use. Set to 0 to disable TTL (cache never expires). Default: 5 minutes.
+     */
+    public static final String TABLE_CACHE_TTL_MS_KEY = "iceberg.catalog.table-cache-ttl-ms";
+
+    private static final long DEFAULT_TABLE_CACHE_TTL_MS = 5 * 60 * 1000L;
+
     private final Configuration icebergConfig;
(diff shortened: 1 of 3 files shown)


fluss-lake/fluss-lake-iceberg/src/main/java/org/apache/fluss/lake/iceberg/source/IcebergLakeSource.java
```diff
@@ -46,11 +46,28 @@
 /** Iceberg lake source. */
 public class IcebergLakeSource implements LakeSource<IcebergSplit> {
     private static final long serialVersionUID = 1L;
+
+    /**
+     * Config key for table cache TTL in milliseconds. After this duration, the cached table is
+     * reloaded on next use. Set to 0 to disable TTL (cache never expires). Default: 5 minutes.
+     */
+    public static final String TABLE_CACHE_TTL_MS_KEY = "iceberg.catalog.table-cache-ttl-ms";
```
nastra, line 54: I don't think this is actually needed anymore. In Iceberg 1.11.0 this is already solved natively in the REST catalog by having a freshness-aware loading of tables using etags (https://github.com/apache/iceberg/pull/14398). The Iceberg community is planning to release 1.11.0 with that feature within the next few weeks

bakjos: Yes, this issue will be resolved once Apache Iceberg 1.11.0 is released, though we are unsure of the exact timeline. Additionally, are there any plans to update Fluss to use Iceberg 1.11.0?

nastra: once 1.11.0 is released I'll update it in Fluss right away. I'm working in the Iceberg community and we're planning to release 1.11.0 within the next few weeks

PR #2660, opened 2026-02-12
[docs] Update quickstart-flink doc to use kv.snapshot.interval 0s to avoid client obtain token issue

luoyuxia:
### Purpose

Linked issue: close #xxx
To fix client obtain token issue

### DIFF
--- a/website/docs/quickstart/flink.md
+++ b/website/docs/quickstart/flink.md
@@ -161,7 +161,6 @@ volumes:
 The Docker Compose environment consists of the following containers:
 - **RustFS:** an S3-compatible object storage for tiered storage. You can access the RustFS console at http://localhost:9001 with credentials `rustfsadmin/rustfsadmin`. An init container (`rustfs-init`) automatically creates the `fluss` bucket on startup.
 - **Fluss Cluster:** a Fluss `CoordinatorServer`, a Fluss `TabletServer` and a `ZooKeeper` server.
-   - Snapshot interval `kv.snapshot.interval` is configured as 60 seconds. You may want to configure this differently for production systems
    - Credentials are configured directly with `s3.access-key` and `s3.secret-key`. Production systems should use CredentialsProvider chain specific to cloud environments.
 - **Flink Cluster**: a Flink `JobManager`, a Flink `TaskManager`, and a Flink SQL client container to execute queries.
 


website/docs/quickstart/flink.md
```diff
@@ -109,7 +109,7 @@ services:
         s3.access-key: rustfsadmin
         s3.secret-key: rustfsadmin
         s3.path-style-access: true
-        kv.snapshot.interval: 60s
+        kv.snapshot.interval: 0s
```
luoyuxia, line 112: Unfortunately,  we'll still need it. Otherwise, it'll throw exception 
```
org.apache.fluss.exception.SecurityTokenException: Failed to get file access security token: The security token included in the request is invalid. (Service: AWSSecurityTokenService; Status Code: 403; Error Code: InvalidClientTokenId; Request ID: fabc25ae-dffb-43bf-af19-f2896f00d918; Proxy: null)
```
when client try to get sts token from server.
Similar to [#2148 ](https://github.com/apache/fluss/discussions/214)

But to support it, we may need change the code the generate sts token in s3 filesystem.
See:
https://github.com/orgs/rustfs/discussions/1239

website/docs/quickstart/lakehouse.md
```diff
@@ -342,7 +342,6 @@ services:
         datalake.iceberg.s3.path-style-access: true
         datalake.iceberg.client.region: us-east-1
     volumes:
-      - ./lib/fluss-lake-iceberg-$FLUSS_DOCKER_VERSION$.jar:/opt/fluss/plugins/iceberg/fluss-lake-iceberg-$FLUSS_DOCKER_VERSION$.jar
```
luoyuxia, line 345: we don't need to mount the iceberg plugin jar

luoyuxia:
cc @wuchong @leekeiabstraction

website/docs/quickstart/flink.md
wuchong: @luoyuxia Thank you, could you open an issue in fluss to fix the way to generate sts token for s3?

website/docs/quickstart/flink.md
wuchong: Remove this line as well.

website/docs/quickstart/flink.md
luoyuxia: create in https://github.com/apache/fluss/issues/2661

leekeiabstraction:
@luoyuxia did you verify if this is an issue in Flink QuickStart?

I was able to make this work with Flink QuickStart. 

It was only the paimon/iceberg one that was failing

luoyuxia:
@wuchong Thanks for review. Comments addressed.

luoyuxia:
> @luoyuxia did you verify if this is an issue in Flink QuickStart?
> 
> I was able to make this work with Flink QuickStart.
> 
> It was only the paimon/iceberg one that was failing

If just follow the steps of Flink QuickStart. I won't meet this problem. But if I sumibit a sql to query all data of a primary key table, the issue will hapeen

leekeiabstraction:
> If just follow the steps of Flink QuickStart. I won't meet this problem. But if I sumibit a sql to query all data of a primary key table, the issue will hapeen

I see, maybe not as part of this PR, let's also try to add specific SQL query that cause the issue in? So that when we revisit this after fixing, we can naturally validate by running the additional query.

Thank you for catching!

wuchong (approved):

Issue #2661, opened 2026-02-12
Support obtain sts token from rustfs


luoyuxia:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/fluss/issues) and found nothing similar.

### Description

Currently, obtain sts token from rustfs will throw exception which make it's impossible for user to use rustfs/minio.  
See comment in https://github.com/apache/fluss/pull/2660/changes#r2797155884
We may change the api to obtain sts token, which works for s3 and rustfs. 

See:
https://github.com/orgs/rustfs/discussions/1239

### Willingness to contribute

- [ ] I'm willing to submit a PR!

Issue #2662, opened 2026-02-12
[filesystem] Enhance S3 filesystem with independent credentials provider for token delegation


affo:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/fluss/issues) and found nothing similar.

### Description

The token delegation mechanism on the server side is quite "stiff".
It only allows to delegate credentials to clients via AK/SK pair.

Ideally one would like to configure server-side credentials to remote storage and delegated credentials in 2 separate ways, especially for obtaining RW access on the server and RO access on the client side.

Furthermore, one may opt out of the token delegation mechanism and let clients obtain credentials on their own (see https://github.com/apache/fluss/pull/1245)

Details below.

#### How it works now

The Fluss server authenticates to S3 and uses credentials for writing snapshots and log segments.

The Fluss server also serves “delegated” credentials to clients to access S3.

On the server side, the Fluss cluster relies on standard Hadoop FS + AWS SDK authentication, hence it can use any CredentialsProvider supported by the AWS SDK and by Hadoop FS implementing:

```java
package com.amazonaws.auth;
public interface AWSCredentialsProvider {
    AWSCredentials getCredentials();
    void refresh();
}
```

This includes any provider and also the `DefaultAWSCredentialsProviderChain`.

On the client side (credentials delegation), the Fluss cluster relies on the implementation of `org.apache.fluss.fs.Filesystem::obtainSecurityToken`, a method that is invoked via gRPC by the Fluss client.
The client also relies on the `SecurityTokenReceiver` registered on the filesystem scheme for using the credentials used by the server.

In the S3 case, the token provided by the server is hardcoded to use some AK/SK pair (see [S3DelegationTokenProvider.java](https://github.com/apache/fluss/blob/release-0.9/fluss-filesystems/fluss-fs-s3/src/main/java/org/apache/fluss/fs/s3/token/S3DelegationTokenProvider.java) ), which are provided from the configuration.
On the client side, the client configures the [DynamicTemporaryAWSCredentialsProvider](https://github.com/apache/fluss/blob/release-0.9/fluss-filesystems/fluss-fs-s3/src/main/java/org/apache/fluss/fs/s3/token/DynamicTemporaryAWSCredentialsProvider.java) that just uses the credentials fetched by the SecurityTokenReceiver.

More precisely, when instantiating the FileSystem class for the corresponding scheme, the code tests for the presence of `s3.acces-key`-like configuration to assess whether we are on the server-side or on the client-side:

[...]

PR #2674, opened 2026-02-13
[fs] bump Hadoop to 3.4 and migrate to AWS SDK Java v2 

### LINKED ISSUE #1210: [filesystem] Migrate to AWS SDK for Java v2
### Search before asking

### Motivation

Fluss currently uses AWS SDK for Java v1 which is in maintenance mode since 07/2024 and reaches [end-of-support on 31/12/2025](https://aws.amazon.com/blogs/developer/announcing-end-of-support-for-aws-sdk-for-java-v1-x-on-december-31-2025/).

### Solution

Migrate `fluss-fs-s3` to [AWS SDK for Java v2](https://docs.aws.amazon.com/sdk-for-java/latest/developer-guide/migration.html).

### Anything else?

_No response_

### Willingness to contribute

xx789633:
### Purpose

Linked issue: close #1210


fluss-filesystems/fluss-fs-hadoop-shaded/src/main/resources/META-INF/NOTICE
```diff
@@ -9,28 +9,42 @@ This project bundles the following dependencies under the Apache Software Licens
...
+- org.apache.hadoop.thirdparty:hadoop-shaded-guava:1.3.0
+- org.apache.hadoop.thirdparty:hadoop-shaded-protobuf_3_25:1.3.0
+- org.apache.kerby:kerb-core:2.0.3
+- org.apache.kerby:kerb-crypto:2.0.3
+- org.apache.kerby:kerb-util:2.0.3
+- org.apache.kerby:kerby-asn1:2.0.3
+- org.apache.kerby:kerby-config:2.0.3
+- org.apache.kerby:kerby-pkix:2.0.3
+- org.apache.kerby:kerby-util:2.0.3
+- org.codehaus.jettison:jettison:1.5.4
```
wuchong: Are the new dependencies really needed? Especially the netty and kerby dependecies. Netty is a big dependency.

xx789633: it seems those dependencies are always included somewhere (like in https://github.com/apache/fluss/blob/main/fluss-filesystems/fluss-fs-azure/src/main/resources/META-INF/NOTICE)

fluss-filesystems/fluss-fs-s3/src/main/java/org/apache/fluss/fs/s3/token/S3DelegationTokenProvider.java
```diff
@@ -70,31 +70,34 @@ public S3DelegationTokenProvider(String scheme, Configuration conf) {
     public ObtainedSecurityToken obtainSecurityToken() {
         LOG.info("Obtaining session credentials token with access key: {}", accessKey);
 
-        AWSSecurityTokenService stsClient =
-                AWSSecurityTokenServiceClientBuilder.standard()
-                        .withRegion(region)
-                        .withCredentials(
-                                new AWSStaticCredentialsProvider(
-                                        new BasicAWSCredentials(accessKey, secretKey)))
+        StsClient stsClient =
```
luoyuxia, line 73: Seems this will also fix #2661 ? Right?
Could you please help verify it? 
Also check minio can be supported or not.

xx789633: I don't think so. According to the rustfs discussion at https://github.com/orgs/rustfs/discussions/1746, rustfs currently only support STS AssumeRole action and does not yet support GetSessionToken, which is precisely what our code attempts to invoke (https://github.com/apache/fluss/blob/main/fluss-filesystems/fluss-fs-s3/src/main/java/org/apache/fluss/fs/s3/token/S3DelegationTokenProvider.java#L80)

fluss-filesystems/fluss-fs-hadoop-shaded/src/main/resources/META-INF/NOTICE
affo: @xx789633 these new ones are introduced by Hadoop 3.4 right?
If we set them as exclusions, would everything fall apart?

affo (commented):
@xx789633 Thank you so much for this contribution!
I am only adding a comment for what is Hadoop 3.4 introducing in terms of dependencies.
If some may be configured as excluded without impacting functionality, I think we should.

fluss-filesystems/fluss-fs-hadoop-shaded/src/main/resources/META-INF/NOTICE
xx789633: Hi @wuchong @affo, I have excluded netty dependency. 

netty dependency is introduced by Hadoop-common 3.4. I have checked the source code  of hadoop-common 3.4 and found it does not use netty at all. The Hadoop community likely declared these dependencies in hadoop-common to unify Netty version management across all sub-modules (such as hadoop-hdfs, hadoop-yarn, and other server-side components), but hadoop-common itself does not actually use them. By the way, netty in Hadoop is mainly used for IPC/RPC, while Fluss only calls standard filesystem calls.

For kerby, those dependencies were there before this pull request. We just upgrade their versions.

fluss-filesystems/fluss-fs-s3/src/main/resources/META-INF/NOTICE
```diff
@@ -3,57 +3,96 @@ Copyright 2025-2026 The Apache Software Foundation
...
+- software.amazon.awssdk:third-party-jackson-core:2.24.6
+- software.amazon.awssdk:utils:2.24.6
+- software.amazon.eventstream:eventstream:1.0.1
 
 This project bundles the following dependencies under the MIT (https://opensource.org/licenses/MIT)
 See bundled license files for details.
 
 - org.checkerframework:checker-qual:2.5.2
 - org.codehaus.mojo:animal-sniffer-annotations:1.17
+- org.reactivestreams:reactive-streams:1.0.4
```
wuchong, line 87: Which dependency is bringing in this transitive dependency? Can we avoid including it? If it must be included, we are required by ASF policy to add a reference to its license in `META-INF/licenses`.

fluss-filesystems/fluss-fs-s3/src/main/resources/META-INF/NOTICE
```diff
@@ -3,57 +3,96 @@ Copyright 2025-2026 The Apache Software Foundation
...
+- io.netty:netty-codec:4.1.100.Final
+- io.netty:netty-codec-http:4.1.100.Final
+- io.netty:netty-codec-http2:4.1.100.Final
+- io.netty:netty-common:4.1.100.Final
+- io.netty:netty-handler:4.1.100.Final
+- io.netty:netty-resolver:4.1.100.Final
+- io.netty:netty-transport:4.1.100.Final
+- io.netty:netty-transport-classes-epoll:4.1.100.Final
+- io.netty:netty-transport-native-epoll:4.1.100.Final
+- io.netty:netty-transport-native-unix-common:4.1.100.Final
```
wuchong, line 28: Please remove all the netty dependency in the NOTICE files (including NOTICE files in other filesystem modules). 

See the warning log
https://github.com/apache/fluss/actions/runs/22016381283/job/63618691913?pr=2674

fluss-filesystems/fluss-fs-s3/src/main/java/org/apache/fluss/fs/s3/token/S3DelegationTokenProvider.java
wuchong: @luoyuxia , do you have other solutions to resolve the rustfs issue?

fluss-filesystems/fluss-fs-hadoop-shaded/src/main/resources/META-INF/NOTICE
```diff
@@ -9,28 +9,33 @@ This project bundles the following dependencies under the Apache Software Licens
...
 - com.google.j2objc:j2objc-annotations:1.1
 - commons-beanutils:commons-beanutils:1.9.4
 - commons-collections:commons-collections:3.2.2
-- commons-io:commons-io:2.8.0
-- commons-logging:commons-logging:1.1.3
-- org.apache.commons:commons-compress:1.21
-- org.apache.commons:commons-configuration2:2.1.1
+- commons-io:commons-io:2.16.1
+- commons-logging:commons-logging:1.2
+- io.dropwizard.metrics:metrics-core:3.2.4
```
wuchong: `io.dropwizard.metrics:metrics-core` is not used by `hadoop-common` (https://hadoop.apache.org/docs/stable/hadoop-project-dist/hadoop-common/dependency-analysis.html), can you remove it from all the NOTICE files in filesystem modules?

fluss-filesystems/fluss-fs-gs/src/main/resources/META-INF/NOTICE
```diff
@@ -75,21 +85,25 @@ This project bundles the following dependencies under the Apache Software Licens
...
-- org.apache.kerby:kerby-pkix:1.0.1
-- org.apache.kerby:kerby-util:1.0.1
+- org.apache.kerby:kerb-core:2.0.3
+- org.apache.kerby:kerb-crypto:2.0.3
+- org.apache.kerby:kerb-util:2.0.3
+- org.apache.kerby:kerby-asn1:2.0.3
+- org.apache.kerby:kerby-config:2.0.3
+- org.apache.kerby:kerby-pkix:2.0.3
+- org.apache.kerby:kerby-util:2.0.3
+- org.codehaus.jettison:jettison:1.5.4
```
wuchong: `org.codehaus.jettison:jettison` is not used by `hadoop-common` (https://hadoop.apache.org/docs/stable/hadoop-project-dist/hadoop-common/dependency-analysis.html), can you remove it from all the NOTICE files in filesystem modules?

wuchong (commented):
Thanks for the update, @xx789633! I’ve left some comments—please take a look.

Additionally, could you compare the built JAR size of the filesystems module before and after this change? I’m concerned that we might be pulling in too many new dependencies—particularly AWS-related ones—which could significantly increase the size of the bundled JAR.

fluss-filesystems/fluss-fs-hadoop-shaded/src/main/resources/META-INF/NOTICE
affo: @xx789633 well, that is an investigation, thank you 🙏

fluss-filesystems/fluss-fs-gs/src/main/resources/META-INF/NOTICE
xx789633: After investigating the transitive dependency tree, here's where each of the new dependencies comes from:

- netty: new in hadoop 3.4. Hadoop's netty usage is for internal RPC/transport, not needed for filesystem operations.
- io.dropwizard.metrics: new in hadoop 3.4. Hadoop metrics framework, not needed for filesystem operations.
- org.codehaus.jettison: new in hadoop 3.4. Not needed.
- org.reactivestream: from AWS SDK v2 http-client-spi. It is core part of AWS SDK v2's async model, cannot be exclude.

By the way, the size of the fluss-fs-s3 jar is only ~40MB. Not too much difference. @wuchong

fluss-filesystems/fluss-fs-hadoop-shaded/src/main/resources/META-INF/NOTICE
xx789633: maybe we can use localstack to replace rustfs?: https://docs.localstack.cloud/aws/services/sts/#api-coverage it supports GetSessionToken API which is precisely what our code attempts to invoke @luoyuxia @wuchong

xx789633:
I see Flink is also working on a similar issue: https://github.com/apache/flink/pull/27026. Let's wait and see how it progresses.

fluss-filesystems/fluss-fs-s3/src/main/java/org/apache/fluss/fs/s3/token/S3DelegationTokenProvider.java
luoyuxia: Actaully, it does't matter if it don't support GetSessionToken. We can still just use STS tokens just like we did for oss.

affo: @luoyuxia I plan to restructure this part of the code by addressing https://github.com/apache/fluss/issues/2662 anyhow 🤝 
The idea is to allow more flexible `AWSCredentialsProviders` to delegate tokens to clients.
Would probably fix under some peculiar configuration?

litiliu:
@xx789633 ，hi， 
https://github.com/apache/flink/pull/27026 has become active again and was recently updated
to Hadoop 3.4.3 with AWS SDK v2.

This PR has been inactive for several months and now conflicts with
the latest main. Are you still planning to continue updating it?

If not, I would like to open a successor PR based on the latest Fluss
main, while crediting and reusing the relevant work from this PR.

affo:
@litiliu +1 here for the initiative

PR #2690, opened 2026-02-16
[hotfix][flink] Fix COUNT(*) aggregation produces wrong result for legacy clusters

### LINKED ISSUE #2689: [flink] Flink COUNT(*) query on primary key table on old cluster reutrns wrong result
### Search before asking

### Fluss version

main (development)

### Please describe the bug 🐞

Should throw unsupported exception, but returns wrong results. 

### Solution

_No response_

wuchong:
This pull request introduces an important validation to the `countLogTable` method, ensuring that counting rows is not allowed on tables with a primary key. This prevents unsupported operations and guides users to upgrade their Fluss cluster if they need this functionality.

Validation improvements:

* Added a check in `countLogTable` to throw an `IllegalArgumentException` when attempting to count rows on a primary key table, with a message recommending an upgrade to Fluss version ≥ 0.9. (`fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/utils/PushdownUtils.java`)

### Purpose

Linked issue: close #2689


MehulBatra:
LGTM!

Issue #2691, opened 2026-02-16
[helm] Enable pulling from private Docker registry


affo:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/fluss/issues) and found nothing similar.

### Description

As of now `imagePullSecrets` is there in `values.yaml` but not used in templates.  
Furthermore, `image.registry` is not rendered in STS templates.

Fix to actually use both values.

### Willingness to contribute

- [x] I'm willing to submit a PR!

PR #2692, opened 2026-02-16
[helm] Enable pulling from private Docker registry

affo:
### Purpose

Linked issue: close https://github.com/apache/fluss/issues/2691

### Brief change log

 - Uses `images.registry`
 - Renders `image.pullSecrets`

Makes it possible to use an alternate registry and use pull secrets to pull the image if the registry is private.

I favored the approach of adding pull secrets directly to STS and not in the service account for bigger flexibility as the user can still link another service account to pods while pull secrets are attached to pods.

### Tests

No test introduced.
Would be good to add tests once [this PR](https://github.com/apache/fluss/pull/2506) lands (introduces `helm unittest` + github workflow).

### API and Format

NO

### Documentation

No new feature, it is already documented in the README.

### DIFF
--- a/helm/README.md
+++ b/helm/README.md
@@ -98,6 +98,20 @@ Important Fluss options surfaced by the chart:
 - zookeeper.address must point to a reachable ensemble.
 - data.dir defaults to /tmp/fluss/data; use a PVC if persistence.enabled=true.
 
+### Private Docker Registry
+
+If you are pulling the Fluss image from a private Docker registry, you can configure the chart using `image.registry` and `image.pullSecrets`.
+
+#### Image values reference
+
+| Parameter            | Description                              | Default             |
+| -------------------- |------------------------------------------|---------------------|
+| `image.registry`     | Docker image registry                    | `""`                |
+| `image.repository`   | Docker image repository                  | `apache/fluss`      |
+| `image.pullPolicy`   | Image pull policy                        | `IfNotPresent`      |
+| `image.pullSecrets`  | Docker registry secret names as an array | `[]`                |
+
+
 ## Resource management
 
 Set resources with requests/limits as appropriate for production. There are no defaults to make it also run on environments with little resources such as Minikube.


helm/templates/_helpers.tpl
```diff
@@ -63,4 +63,29 @@ Selector labels
...
 app.kubernetes.io/instance: {{ .Release.Name }}
-{{- end }}
\ No newline at end of file
+{{- end }}
+
+{{/*
+Image name
+*/}}
+{{- define "fluss.image" -}}
+{{- $image := printf "%s:%s" .Values.image.repository .Values.image.tag }}
```
xx789633: If a user omits image.tag or sets it to empty, this would produce an invalid image reference like `apache/fluss:`. Maybe we need to add a fallback here.

xx789633: If a user provides a purely numeric tag (e.g. `tag: 1.0`), printf "%s" in Helm may output `apache/fluss:%!s(float64=1)`.

affo: addressed 👍

affo: what a catch! Partially addressed 🤝

affo:
@xx789633 the latest changes now address your doubts 🤝 

`repository` is now required.  
if `tag` is missing, no `:` is added avoiding a malformed image name.

For tag formatting, unfortunately, it is not possible to discriminate `1.0` VS `1` in Go templating, as they would always render to `1` as strings.  However, I added a `toString` that would cover `1.1` for example 🤝 

Wrapping the tag in quotes in YAML is left to the user 👍 
In our default values we offer a tag value wrapped in quotes.

I checked this behavior and it would be the same in Bitnami's charts.
Tested with Zookeeper 🤝

xx789633:
Looks good to me.

swuferhong:
LGTM overall. One more suggestion: Could you add documentation in `helm/README.md` explaining how to use `image.registry` and `image.pullSecrets` when configuring a private Docker registry?

xx789633:
I have appended a commit to document the private docker registry. Merging....

affo:
@xx789633 thanks for promptly acting on this one!
I was too  busy yesterday to promptly address 😓

PR #2699, opened 2026-02-19
  [docs] Reorganize Streaming Lakehouse documentation to higlight deployment and core features 

### LINKED ISSUE #2358: Refactor the structure of the document about integration with  datalake
### Search before asking

### Description

todo

### Willingness to contribute

MehulBatra:
### Purpose

Linked issue: close #2358

### Brief change log                                                                   
                                                                                                                        
  - Add new pages: deploying-streaming-lakehouse, tiering-service, union-read, datalake-catalogs                        
  - Rename integrate-data-lakes/ to datalake-formats/                                                                   
  - Move filesystems/ under tiered-storage/                                                                             
  - Slim down lakehouse-storage.md to focus on configuration                                                            
  - Add "Getting Started" section to overview.md                                                                        
  - Fix auto-compaction and auto-expire-snapshot defaults to false                                                      
  - Update cross-references for new paths


MehulBatra:
@luoyuxia Resolved the conflicts, please have a look!

website/docs/streaming-lakehouse/integrate-data-lakes/_category_.json
```diff
@@ -1,4 +0,0 @@
-{
```
luoyuxia, line 1: This PR renames existing docs paths, but does not preserve redirects for the old URLs.

For example:
- `/docs/maintenance/filesystems/s3`
- `/docs/maintenance/filesystems/hdfs`
- `/docs/streaming-lakehouse/integrate-data-lakes/formats/paimon`
- `/docs/streaming-lakehouse/integrate-data-lakes/catalogs/lakekeeper`

These are published documentation URLs and will become 404 after the rename. The current `createRedirects` logic only creates version aliases for routes that still exist; it does not create redirects for deleted old routes.

Please add explicit redirects for all renamed leaf pages.

website/build_versioned_docs.sh
```diff
@@ -122,6 +122,14 @@ for branch in $branches; do
     # Copy the website/docs directory to the target directory
     $CP_CMD "$version_dir/" || { echo "Failed to copy for branch: $clean_branch_name"; continue; }
     echo "Copied documentation for branch '$clean_branch_name' to '$version_dir'"
+
+    # Fix absolute links in versioned docs (convert /docs/maintenance/filesystems/hdfs.md to relative path)
+    lakehouse_file="$version_dir/maintenance/tiered-storage/lakehouse-storage.md"
+    if [ -f "$lakehouse_file" ]; then
+      sed -i.bak 's|/docs/maintenance/filesystems/hdfs\.md|../../filesystems/hdfs|g' "$lakehouse_file"
```
luoyuxia: The replacement path here looks wrong.

`lakehouse_file` is under `version-$version/maintenance/tiered-storage/lakehouse-storage.md`, so replacing `/docs/maintenance/filesystems/hdfs.md` with `../../filesystems/hdfs` resolves to `version-$version/filesystems/hdfs`.

For release branches, the copied old filesystem docs are still under `version-$version/maintenance/filesystems/hdfs.md`, so this should be `../filesystems/hdfs` instead.

website/docs/install-deploy/deploying-streaming-lakehouse.md
```diff
@@ -0,0 +1,237 @@
...
+
+## Cluster Configuration
+
+You can enable Lakehouse storage through:
+1. **Static configuration**: Configure in `server.yaml` before starting the cluster
+2. **Dynamic configuration**: Enable at runtime using the `set_cluster_configs` procedure
+
+### Method 1: Static Configuration
+
+Configure lakehouse settings in `server.yaml` on all Fluss servers (CoordinatorServer and TabletServer).
```
luoyuxia, line 27: For “Method 1: Static Configuration”, could we add a short explanation of the configuration convention before the Paimon/Iceberg/Lance examples?

Right now users can copy the examples, but it is not clear how to derive the keys when using Hive/REST/Glue/JDBC catalogs.

Suggested points:
- `datalake.enabled: true` explicitly enables lakehouse capability for the cluster.
- `datalake.format` selects the default lake format.
- Format-specific options use the `datalake.<format>.*` prefix.
- Fluss strips the `datalake.<format>.` prefix and passes the remaining keys to the corresponding lake catalog/client. For example, `datalake.paimon.metastore` becomes `metastore`, and `datalake.iceberg.type` becomes `type`.
- Only configure options for the selected `datalake.format`.
- The tiering service also needs the same `datalake.*` options because it is an independent Flink job.

website/docs/install-deploy/deploying-streaming-lakehouse.md
```diff
@@ -0,0 +1,237 @@
...
+);
+```
+
+See [set_cluster_configs](../engine-flink/procedures.md#set_cluster_configs) for more details.
+
+## Adding Required JARs
+
+### Fluss Server JARs
+
+Add JARs to `${FLUSS_HOME}/plugins/<format>/` based on your configuration:
```
luoyuxia, line 106: This dependency table weakens some download/configuration guidance that existed in the old docs.

Suggested changes:
- For `Paimon with S3/OSS`, link to the Paimon download/filesystem JAR docs and mention that users should choose `paimon-s3` / `paimon-oss` matching their Paimon version.
- For `Paimon Hive catalog`, keep the old link to Flink Hive connector bundled Hive jar docs.
- For `Iceberg with S3`, link to Iceberg AWS integration or Maven artifacts, and mention `iceberg-aws` / `iceberg-aws-bundle` should match the Iceberg version.
- For `Hadoop client JARs`, avoid only using the generic name. The old Iceberg docs gave three options: use `HADOOP_CLASSPATH`, download `hadoop-apache-3.3.5-2.jar`, or download a full Hadoop package. Please keep these entry points or link back to the Iceberg Hadoop Dependencies section.
- `For HDFS, see the HDFS setup guide.` feels misplaced here. HDFS is a storage backend dependency, not a lake format dependency. I’d remove it from this table section.

website/docs/streaming-lakehouse/tiering-service.md
```diff
@@ -0,0 +1,79 @@
+---
+title: "Tiering Service"
+sidebar_position: 2
+---
+
+# Tiering Service
+
+The Tiering Service continuously compacts real-time data from Fluss (stored in Arrow format) into data lake formats (Parquet) for cost-efficient long-term storage and analytics.
```
luoyuxia: `data lake formats (Parquet)` is not quite accurate.

Paimon/Iceberg/Lance are lake/table formats. Parquet is only an underlying file format for some cases; Paimon can also use ORC, and Lance should not be described as simply Parquet.

Could we make this more general, for example:

"The Tiering Service continuously compacts real-time data from Fluss into the configured lake format for cost-efficient long-term storage and analytics."

website/docs/streaming-lakehouse/tiering-service.md
```diff
@@ -0,0 +1,79 @@
...
+    /path/to/fluss-flink-tiering-$FLUSS_VERSION$.jar \
+    --fluss.bootstrap.servers localhost:9123 \
+    ...
+```
+
+Multiple tiering service jobs can run simultaneously. They are coordinated by the Fluss cluster to ensure exactly-once semantics and automatic load balancing.
+
+## Monitoring
+
+Key metrics for monitoring the Tiering Service are available through Flink's metrics system. See [Monitoring Metrics](../maintenance/observability/monitor-metrics.md) for details on lakehouse tiering metrics.
```
luoyuxia: The “Monitoring Metrics” link should point directly to the tiering service metrics section.

From this page, the link should be:

`../maintenance/observability/monitor-metrics.md#tiering-service-metrics`

This will route to `/docs/maintenance/observability/monitor-metrics/#tiering-service-metrics`.

website/docs/streaming-lakehouse/tiering-service.md
```diff
@@ -0,0 +1,79 @@
...
+| Log Table (append-only) | TieringLogSplit | Defines starting and stopping offsets for a contiguous range |
+| Primary Key Table | TieringSnapshotSplit | References snapshot ID and log offset for CDC replay |
+
+## Configuration Options
+
+### Tiering Service Options
+
+| Option | Type | Default | Description |
+|--------|------|---------|-------------|
+| `lake.tiering.auto-expire-snapshot` | Boolean | false | Auto-trigger snapshot expiration on commit |
```
luoyuxia, line 73: Could we add a short explanation for how Tiering Service Options are configured?

Right now the table lists `lake.tiering.auto-expire-snapshot`, but users cannot tell where to set it.

The tiering job arguments are split into three groups:
- `--fluss.*`: Fluss client config, for example `--fluss.bootstrap.servers`
- `--datalake.<format>.*`: lake catalog/storage config; Fluss strips the `datalake.<format>.` prefix before passing it to the lake connector
- `--lake.tiering.*`: tiering service job-level config, for example `--lake.tiering.auto-expire-snapshot true`

It would help to add an example:

```shell
${FLINK_HOME}/bin/flink run \
    -Dparallelism.default=3 \
    /path/to/fluss-flink-tiering-$FLUSS_VERSION$.jar \
    --fluss.bootstrap.servers localhost:9123 \
    --datalake.format paimon \
    --datalake.paimon.metastore filesystem \
    --datalake.paimon.warehouse /tmp/paimon \
    --lake.tiering.auto-expire-snapshot true
```

Also worth clarifying:
- `table.datalake.*` is configured per table when creating/altering tables.
- `lake.tiering.*` is configured when starting the tiering service job.

website/docs/streaming-lakehouse/integrate-data-lakes/_category_.json
MehulBatra: Thanks for catching this, you're right that the renamed pages need redirects. I ended up using the existing createRedirects function rather than an explicit redirects: [{from, to}]  array, because the explicit form breaks the build today, and I wanted to explain why.         
                                                                                                
The renamed layout only exists in next (1.0-SNAPSHOT). The unversioned /docs/... URLs still   
resolve to 0.9, which keeps the old structure. So right now an explicit redirect fails on both
ends: its to: target (the new path) doesn't exist yet, so plugin-client-redirects throws      
"these paths are redirected to but do not exist", and its from: path is still a live 0.9 page,
so the plugin would skip it anyway to avoid shadowing a real route.                           
                                                                                                
The old URLs only start 404ing once this restructure becomes the latest release (when 1.0     
replaces 0.9 as the unversioned latest). Old versions themselves (0.6–0.9) keep working and   
need no redirects.                                                                            
                                                                                                
So instead of hardcoding redirects that would either break the build now or need to be        
remembered at 1.0 release time, I added the rename mapping into createRedirects:              
                                                                                                
```
const renameRules = [                                                                         
  { from: '/maintenance/filesystems/', to: '/maintenance/tiered-storage/filesystems/' },      
  { from: '/streaming-lakehouse/integrate-data-lakes/formats/', to:                           
'/streaming-lakehouse/datalake-formats/' },                                                   
  { from: '/streaming-lakehouse/integrate-data-lakes/catalogs/', to:                          
'/streaming-lakehouse/datalake-catalogs/' },                                                  
];                   
```                                                                         
                                                                                                
Because createRedirects keys off routes that actually exist, the target is always valid, and  
each old→new redirect activates exactly when the restructured docs become the latest release —
i.e. precisely when the old URL would otherwise 404. It's build-safe today and covers every   
leaf page under the three moved folders automatically.

website/docs/streaming-lakehouse/tiering-service.md
MehulBatra: I agree, I forgot the orc, lance file format lens, thanks for highlighting.

website/docs/streaming-lakehouse/datalake-formats/iceberg.md
luoyuxia, line 1: Line 265: Could we fix this option name while reorganizing the docs? `table.datalake.auto-maintenance` does not exist in `ConfigOptions`; the available table-level options are `table.datalake.auto-compaction` and `table.datalake.auto-expire-snapshot`. Users copying this example would think maintenance is enabled while the option is not a valid Fluss option. For this example, I think we should either remove the line or change it to `table.datalake.auto-compaction` if compaction is intended.

website/docs/streaming-lakehouse/union-read.md
```diff
@@ -0,0 +1,113 @@
...
+
+# Union Read
+
+Union Read is a core feature of Fluss's Streaming Lakehouse that combines real-time data from Fluss with historical data from the data lake in a single query.
+
+## Overview
+
+For a table with `'table.datalake.enabled' = 'true'`, data exists in two layers:
+- **Fluss (hot data)**: Sub-second fresh data stored in Arrow format
+- **Data Lake (cold data)**: Historical data stored in Parquet format
```
luoyuxia: Could we avoid saying the cold layer is always stored in Parquet format? This page covers Paimon/Iceberg/Lance union read in general: Paimon can be configured with ORC, and Lance is not simply Parquet. A more accurate wording would be something like "Historical data stored in the configured lake format/storage".

website/docs/streaming-lakehouse/datalake-formats/paimon.md
luoyuxia, line 1: Line 185: This page still has a broken relative link to Fluss data types: `table-design/data-types.md` resolves under `streaming-lakehouse/datalake-formats/`. The target is `website/docs/table-design/data-types.md`, so this should be `../../table-design/data-types.md`.

website/docs/streaming-lakehouse/datalake-formats/lance.md
luoyuxia, line 1: Line 128: Same broken relative link here: `table-design/data-types.md` resolves under `streaming-lakehouse/datalake-formats/`, while the actual target is `website/docs/table-design/data-types.md`. Please change it to `../../table-design/data-types.md`.

luoyuxia:
@MehulBatra I left a few more comments.

website/docs/streaming-lakehouse/datalake-formats/iceberg.md
MehulBatra: This is a miss from my end, let me correct it, it was existing, my AI hallucinated and didn't cross verify and fix.

website/docs/streaming-lakehouse/union-read.md
MehulBatra: I just said it to make it sound simple for user who coming across the concept for first time, but your concern is valid, someone with better understand might misunderstand in terms of lance or paimon + orc.

MehulBatra:
> @MehulBatra I left a few more comments.

Thanks for the thoughtful feedback. I'm glad we're simplifying Streamhouse and making the documentation more consistent, crisp, and easy to digest. Really appreciate you taking the time to review it.

luoyuxia:
@MehulBatra I pushed a small docs fix to this PR: in the Paimon integration page, the retention section incorrectly said expired Fluss log data / cleaned-up partitions remain accessible via Iceberg. I changed those references to Paimon, since this section is describing Paimon tiering behavior.

luoyuxia (approved):
@MehulBatra Thanks. LGTM.. Merging...

PR #2703, opened 2026-02-21
Fix the LICENSE and NOTICE files for the binary distribution, including transitive dependencies

jbonofre:
This PR fixes the `LICENSE` and `NOTICE` files for the binary distribution with:
- clearly stating all bundles dependencies/code, including the transitive, shaped in the Fluss artifacts (server and plugins)
- include BSD/MIT license content in line
- update the NOTICE file correctly


LICENSE-bin
```diff
@@ -0,0 +1,1173 @@
...
+|       
+| Third party licenses
+| The OpenPGP library and the MLS library both make use of additional open source code:
+|       
+| openpgp - includes modified BZIP2 library which is licensed under the Apache Software License, Version 2.0.
+| MLS - The MLS Client makes use of io.grpc licensed under Apache Software License, Version 2.0, and com.google.protobuf which is licensed under the 3-Clause BSD License.
+
+------------------------------------------------------------------------------------
+
+This artifacct bundles Aliyun.
```
luoyuxia: nit: seems a typo
`artifacct` -> `artifact`
or 
`artifacct` -> `product`
?

LICENSE-bin
```diff
@@ -0,0 +1,1173 @@
+                                 Apache License
```
luoyuxia, line 1: Just curious — are these LICENSE files in the binary distribution auto-generated or manually maintained? And when Fluss adds new dependencies in the future, do we need to manually include their licenses as well?

luoyuxia: Also, I'm not entirely clear on the distinction between NOTICE and LICENSE — when should something go into the NOTICE file vs. the LICENSE file? Would appreciate some guidance on this so I can better understand the process. 🙏

LICENSE-bin
```diff
@@ -0,0 +1,1173 @@
...
+
+------------------------------------------------------------------------------------
+
+This product bundles Scala.
+Project URL: https://www.scala-lang.org/
+License: Apache License, Version 2.0 - http://www.apache.org/licenses/LICENSE-2.0.txt
+
+------------------------------------------------------------------------------------
+
+This product bundles Apache Hive.
```
luoyuxia, line 922: I'm curious, where does it bundle Hive? Which module or dependency pulls it in?

luoyuxia (commented):
@jbonofre Thanks for the pr. Left some question.
@wuchong Could you please also help review this

LICENSE-bin
jbonofre: It's pretty hard to have all auto-generated (especially when you shade a dependency which has itself a shade dependency). I'm having some scripts to assist me, but I do most the checks "manually".

The `LICENSE` should contain all code and dependencies you bundle in the artifact, with the type of license and copyright. In the case of BSD or MIT license, the license content should be inline our own `LICENSE`.

The `NOTICE` file should be as simple as possible. It should contain only the relevant information of the `NOTICE` file for the Apache license dependencies. For instance, Netty is Apache licensed dependency, so as Netty has a `NOTICE`, the Netty `NOTICE` content should be in our `NOTICE` (the relevant information).
If the Apache licensed dependency doesn't have `NOTICE`, nothing is required.

In the case of ASF dependencies (like Apache Flink), it's not strictly required to include `NOTICE` because it's covered by `This product includes software developed at The Apache Software Foundation (http://www.apache.org/)`.

LICENSE-bin
jbonofre: Hive is shaded in the paimon bundle (in `plugins/paimon`).

LICENSE-bin
luoyuxia: Thanks JB, that's a really thorough and expert explanation! Super helpful, really clears things up for me. 🙏

luoyuxia (approved):
@jbonofre Thank for the pr. LGTM. Merging....

PR #2707, opened 2026-02-22
Fix Lance writer to emit Arrow FixedSizeList for array columns to enable native vector search

### LINKED ISSUE #2706: [lake/lance] Lance writer should emit Arrow FixedSizeList for array columns to enable native vector search
### Search before asking

### Fluss version

main (development)

### Please describe the bug 🐞

When tiering tables with ARRAY<FLOAT> columns to Lance, Fluss always writes them as Arrow's variable-length List<Float32>. This prevents Lance's native vector search (nearest / ANN index) from working, since Lance requires FixedSizeList<Float32>(n) for vector columns.

This seems to be due to LanceArrowUtils.toArrowType(), ArrayType unconditionally maps to ArrowType.List.INSTANCE:

```java
  // fluss-lake/fluss-lake-lance/src/main/java/org/apache/fluss/lake/lance/utils/LanceArrowUtils.java:140
  } else if (dataType instanceof ArrayType) {
      return ArrowType.List.INSTANCE;
  }
```

### Solution

_No response_

leekeiabstraction:
### Purpose

Linked issue: close #2706

Fix Lance writer to emit Arrow FixedSizeList for array columns to enable native vector search.

Querying with pylance native vector search fails because pylance expects embedding column to be of fixedSizeList type instead of variable size list. This PR fixes the conversion from Fluss' arrow data to Lance's arrow data by using FixedSizeList if `<column>.arrow.fixed-size-list.size` is defined, similar to how Spark SQL does it ref: https://lance.org/integrations/spark/operations/ddl/create-table/#creating-large-string-columns

### Tests

Added unit test
- Tested manually that pylance successfully performed a vector search on the tiered data

### DIFF
--- a/fluss-lake/fluss-lake-lance/src/main/java/org/apache/fluss/lake/lance/utils/LanceArrowUtils.java
+++ b/fluss-lake/fluss-lake-lance/src/main/java/org/apache/fluss/lake/lance/utils/LanceArrowUtils.java
@@ -84,16 +84,24 @@ public class LanceArrowUtils {
 
     private static Field toArrowField(
             String fieldName, DataType logicalType, Map<String, String> tableProperties) {
+        checkArgument(
+                !fieldName.contains("."),
+                "Column name '%s' must not contain periods. "
+                        + "Lance does not support field names with periods.",
+                fieldName);
         ArrowType arrowType;
         if (logicalType instanceof ArrayType && tableProperties != null) {
             String sizeStr = tableProperties.get(fieldName + FIXED_SIZE_LIST_SIZE_SUFFIX);
             if (sizeStr != null) {
-                int listSize = -1;
+                int listSize;
                 try {
                     listSize = Integer.parseInt(sizeStr);
-                } catch (NumberFormatException ignored) {
-                    // Not really ignored, IllegalArgumentEx still thrown below.
-                    // This removes duplicate boilerplates for throwing IAE
+                } catch (NumberFormatException e) {
+                    throw new IllegalArgumentException(
+                            String.format(
+                                    "Invalid value '%s' for property '%s', expected a positive integer.",
+                                    sizeStr, fieldName + FIXED_SIZE_LIST_SIZE_SUFFIX),
+                            e);
                 }
 
                 checkArgument(
--- a/fluss-lake/fluss-lake-lance/src/main/java/org/apache/fluss/lake/lance/utils/ArrowDataConverter.java
+++ b/fluss-lake/fluss-lake-lance/src/main/java/org/apache/fluss/lake/lance/utils/ArrowDataConverter.java
@@ -164,26 +164,40 @@ public class ArrowDataConverter {
 
         int valueCount = shadedListVector.getValueCount();
         int expectedListSize = nonShadedFixedSizeListVector.getListSize();
-        int expectedTotalValueCount = valueCount * expectedListSize;
+        int nonNullCount = valueCount - shadedListVector.getNullCount();
+        int expectedTotalChildCount = nonNullCount * expectedListSize;
 
         // Validate that backing data vector element count matches expected fixed-size layout.
-        // If every list has exactly expectedListSize elements, the total must be
-        // valueCount * expectedListSize.
-        int totalValueCount = shadedListVector.getDataVector().getValueCount();
-        if (totalValueCount != expectedTotalValueCount) {
+        int totalChildCount = shadedListVector.getDataVector().getValueCount();
+        if (totalChildCount != expectedTotalChildCount) {
             throw new IllegalArgumentException(
                     String.format(
                             "Total child elements (%d) does not match expected %d for FixedSizeList conversion.",
-                            totalValueCount, expectedTotalValueCount));
+                            totalChildCount, expectedTotalChildCount));
         }
 
-        // Copy the child data vector recursively (e.g., the float values)
+        // Copy child data from the source ListVector to the target FixedSizeListVector.
+        //
+        // In a ListVector, child elements for non-null rows are packed contiguously
+        // (null rows contribute 0 children). In a FixedSizeListVector, child data is
+        // stride-based: row i's data starts at index i * listSize, so null rows still
+        // occupy child slots. When null rows exist, a simple bulk buffer copy won't
+        // work because the layouts differ — we must remap per-row using the source
+        // offset buffer.
         org.apache.fluss.shaded.arrow.org.apache.arrow.vector.FieldVector shadedDataVector =
                 shadedListVector.getDataVector();
         FieldVector nonShadedDataVector = nonShadedFixedSizeListVector.getDataVector();
 
         if (shadedDataVector != null && nonShadedDataVector != null) {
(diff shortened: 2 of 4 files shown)


leekeiabstraction:
@XuQianJin-Stars Appreciate if you can review this PR 🙏

leekeiabstraction:
Addressed Co-pilot comments, also fixed a memory leak that could occur if exception is thrown during conversion from shaded to non-shaded objects.

fluss-lake/fluss-lake-lance/src/main/java/org/apache/fluss/lake/lance/LanceLakeCatalog.java
```diff
@@ -61,7 +61,9 @@ public void createTable(TablePath tablePath, TableDescriptor tableDescriptor, Co
         List<Field> fields = new ArrayList<>();
         // set schema
         fields.addAll(
-                LanceArrowUtils.toArrowSchema(tableDescriptor.getSchema().getRowType())
+                LanceArrowUtils.toArrowSchema(
+                                tableDescriptor.getSchema().getRowType(),
+                                tableDescriptor.getCustomProperties())
```
XuQianJin-Stars, line 66: Need to confirm the return type of `getCustomProperties()`. If it returns an immutable Map or a Properties type rather than `Map<String, String>`, will `tableProperties.get(fieldName + FIXED_SIZE_LIST_SIZE_SUFFIX)` in toArrowField work correctly? Suggest using `Map<String, String>` in the new method signature and ensuring compatibility at the call site.

fluss-lake/fluss-lake-lance/src/main/java/org/apache/fluss/lake/lance/utils/LanceArrowUtils.java
```diff
@@ -47,31 +47,76 @@
...
+import static org.apache.fluss.utils.Preconditions.checkArgument;
+
 /**
  * Utilities for converting Fluss RowType to non-shaded Arrow Schema. This is needed because Lance
  * requires non-shaded Arrow API.
  */
 public class LanceArrowUtils {
 
+    /** Property suffix for configuring a fixed-size list Arrow type on array columns. */
+    public static final String FIXED_SIZE_LIST_SIZE_SUFFIX = ".arrow.fixed-size-list.size";
```
XuQianJin-Stars, line 62: The property key format is <column_name>.arrow.fixed-size-list.size. If the column name itself contains . (e.g., a.b), it leads to parsing ambiguity. For example, a.b.arrow.fixed-size-list.size — it's unclear whether this refers to column a.b or column a with some unrelated suffix.

Suggestion:
1. Explicitly document that column names must not contain `.`; or
2. Validate in toArrowField that the column name does not contain ., and throw a meaningful error message if it does.

fluss-lake/fluss-lake-lance/src/main/java/org/apache/fluss/lake/lance/utils/LanceArrowUtils.java
```diff
@@ -47,31 +47,76 @@
...
+            String fieldName, DataType logicalType, Map<String, String> tableProperties) {
+        ArrowType arrowType;
+        if (logicalType instanceof ArrayType && tableProperties != null) {
+            String sizeStr = tableProperties.get(fieldName + FIXED_SIZE_LIST_SIZE_SUFFIX);
+            if (sizeStr != null) {
+                int listSize = -1;
+                try {
+                    listSize = Integer.parseInt(sizeStr);
+                } catch (NumberFormatException ignored) {
+                    // Not really ignored, IllegalArgumentEx still thrown below.
```
XuQianJin-Stars: Validating for a positive integer is correct. However, if the property value is an empty string "" or whitespace " ", Integer.parseInt will throw a NumberFormatException without a user-friendly error message. Suggest catching NumberFormatException and converting it to an IllegalArgumentException that includes the property key name, making it easier for users to locate the configuration issue.
```
int listSize;
try {
    listSize = Integer.parseInt(sizeStr);
} catch (NumberFormatException e) {
        String.format("Invalid value '%s' for property '%s', expected a positive integer.",
            sizeStr, fieldName + FIXED_SIZE_LIST_SIZE_SUFFIX), e);
}
checkArgument(listSize > 0, "...");
```

fluss-lake/fluss-lake-lance/src/test/java/org/apache/fluss/lake/lance/tiering/LanceTieringTest.java
```diff
@@ -233,6 +246,13 @@ private void verifyLogTableRecords(
                     .isEqualTo(expectRecord.getRow().getString(1).toString());
             assertThat(((VarCharVector) root.getVector(2)).getObject(i).toString())
                     .isEqualTo(expectRecord.getRow().getString(2).toString());
+            // check embedding column
+            java.util.List<?> embeddingValues = embeddingVector.getObject(i);
+            assertThat(embeddingValues).hasSize(EMBEDDING_LIST_SIZE);
+            org.apache.fluss.row.InternalArray expectedArray = expectRecord.getRow().getArray(3);
+            for (int j = 0; j < EMBEDDING_LIST_SIZE; j++) {
+                assertThat((Float) embeddingValues.get(j)).isEqualTo(expectedArray.getFloat(j));
```
XuQianJin-Stars, line 254: Missing null value scenario in test data?

fluss-lake/fluss-lake-lance/src/main/java/org/apache/fluss/lake/lance/LanceLakeCatalog.java
leekeiabstraction: getCustomProperties() returns Map<String, String>

```java
    public Map<String, String> getCustomProperties() {
        return customProperties;
    }
```

fluss-lake/fluss-lake-lance/src/main/java/org/apache/fluss/lake/lance/utils/LanceArrowUtils.java
leekeiabstraction: Good comment, I actually look up Lance doc further and found that they do not support dots in columns. Ref: https://docs.lancedb.com/search/filtering#advanced-sql-filters

Updated to throw exception

fluss-lake/fluss-lake-lance/src/test/java/org/apache/fluss/lake/lance/tiering/LanceTieringTest.java
leekeiabstraction: Great comment. Testing for null actually surfaced a bug.

The fix did take me some time to get to..

leekeiabstraction:
@XuQianJin-Stars Appreciate if you can review again

XuQianJin-Stars:
> @XuQianJin-Stars Appreciate if you can review again

well

XuQianJin-Stars:
LGTM

PR #2732, opened 2026-02-26
[common] A custom rounding policy that reduces Arrow's chunk size fm 16MB to 4MB which same as netty arena.

### LINKED ISSUE #2729: Use default chunk size as 4 MB than than 16Mb in arrow's rolling policy.
### Search before asking

### Description:
Apache Arrow currently calculates chunk sizes based on the assumption that Netty's PooledByteBufAllocatorL uses a **16MB** page size.  If size is blow 16MB,  will use nextPowerOfTwo of requestSize.
```java
  public long getRoundedSize(long requestSize) {
    return requestSize < chunkSize ? CommonUtil.nextPowerOfTwo(requestSize) : requestSize;
  }
```

However, Netty's default maximum buffer size has been **reduced from 16MB to 4MB** in recent versions (see [Netty PR #12108](https://github.com/netty/netty/pull/12108)). This mismatch leads to memory inefficiency in scenarios where chunk sizes fall between 4MB and 16MB.

### Problem Details
For example, if fluss batch size is 4.1MB, then arrow will getRoundedSize as 8MB. 50% is wasted.

### Impact on Flink:
Flink's default off-heap memory is only 128MB, and with multiple slots, this is divided further. It is not only used for Fluss (batch reading, decompression, and Netty network requests) but also for the framework itself or other connectors.

In such constrained environments, Arrow's over-allocation exacerbates memory pressure and reduces throughput.

### Willingness to contribute

loserwang1024:
### Purpose

Linked issue: close #2729

### DIFF
--- a/fluss-common/src/main/java/org/apache/fluss/record/LogRecordReadContext.java
+++ b/fluss-common/src/main/java/org/apache/fluss/record/LogRecordReadContext.java
@@ -25,6 +25,7 @@ import org.apache.fluss.metadata.TableInfo;
 import org.apache.fluss.row.InternalRow;
 import org.apache.fluss.row.InternalRow.FieldGetter;
 import org.apache.fluss.row.ProjectedRow;
+import org.apache.fluss.row.arrow.memory.BufferAllocatorUtil;
 import org.apache.fluss.shaded.arrow.org.apache.arrow.memory.BufferAllocator;
 import org.apache.fluss.shaded.arrow.org.apache.arrow.vector.VectorSchemaRoot;
 import org.apache.fluss.types.DataType;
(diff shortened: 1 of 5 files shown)


loserwang1024:
@wuchong @platinumhamburg @leonardBang , CC

fluss-common/src/main/java/org/apache/fluss/record/LogRecordReadContext.java
```diff
@@ -115,7 +116,11 @@ private static LogRecordReadContext createArrowReadContext(
             boolean projectionPushDowned,
             SchemaGetter schemaGetter) {
         // TODO: use a more reasonable memory limit
-        BufferAllocator allocator = new RootAllocator(Long.MAX_VALUE);
+        BufferAllocator allocator =
+                new RootAllocator(
+                        AllocationListener.NOOP,
+                        Long.MAX_VALUE,
+                        FlussRoundingPolicy.DEFAULT_ROUNDING_POLICY);
```
wuchong: I noticed that `RecordAccumulator` and `KvManager` also use the default `RootAllocator` constructor. Could you verify if these instances require updates as well?

If they do, please refactor the logic into a shared static method to prevent inconsistent initialization. If not, please add a comment explaining why the customized rounding policy is necessary here but not in those other components.

Issue #2673, opened 2026-02-13
[spark] to support map type


YannByron:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/fluss/issues) and found nothing similar.

### Motivation

to support map type

### Willingness to contribute

- [x] I'm willing to submit a PR!

PR #2740, opened 2026-02-27
[spark] support map type

YannByron:
### Purpose

Linked issue: close https://github.com/apache/fluss/issues/2673

### DIFF
--- a/fluss-common/src/main/java/org/apache/fluss/utils/InternalRowUtils.java
+++ b/fluss-common/src/main/java/org/apache/fluss/utils/InternalRowUtils.java
@@ -94,6 +94,12 @@ public class InternalRowUtils {
         return new GenericArray(newArray);
     }
 
+    /**
+     * Creates a copy of the given {@link InternalMap}.
+     *
+     * <p>This method is intended for internal use by the Fluss Spark adapter and is not part of the
+     * stable public API. Its signature and behavior may change without notice across releases.
+     */
     public static InternalMap copyMap(InternalMap map, DataType keyType, DataType valueType) {
         if (map instanceof BinaryMap) {
             return ((BinaryMap) map).copy();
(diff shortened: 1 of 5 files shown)


YannByron:
@wuchong @Yohahaha please take a look.

fluss-spark/fluss-spark-common/src/main/scala/org/apache/fluss/spark/row/DataConverter.scala
```diff
@@ -76,8 +76,7 @@ object DataConverter {
   }
 
   def toSparkMap(flussMap: FlussInternalMap, mapType: FlussMapType): SparkMapData = {
-    // TODO: support map type in fluss-spark
-    throw new UnsupportedOperationException()
+    new FlussAsSparkMap(mapType).replace(flussMap)
```
Yohahaha, line 79: do we need to call `InternalRowUtils.copyMap` ?

YannByron: No. What is needed here is an implementation of `SparkMapData`.

wuchong:
cc @beryllw could you help to review this?

YannByron:
The failed CI is caused by a flaky UT that is fixed in https://github.com/apache/fluss/pull/3045.

fluss-spark/fluss-spark-common/src/main/scala/org/apache/fluss/spark/row/DataConverter.scala
beryllw, line 75: https://github.com/apache/fluss/blob/bb0e12b868b84575b2adc75ba1429a7130488797/fluss-client/src/main/java/org/apache/fluss/client/table/scanner/log/CompletedFetch.java#L95-L106
https://github.com/apache/fluss/blob/b306ed00bcf180ad20adb5b65c9034b43d26d319/fluss-common/src/main/java/org/apache/fluss/row/InternalArray.java#L178-L190

May be we don't need to call `InternalRowUtils.copyArray`?

YannByron: @beryllw Thank you for your comments to make me think further. The main problem here is that `toSparkArray` and `toSparkMap` behave inconsistently on whether or not to `copy`. If we are more rigorous, we should all `copy` to avoid the 'use-after-free' problem. However, since the data objects converted by the 'DataConverter' will only be used in `fluss-spark`, we can control it and consider not copying. If we want to remote the copy in `toSparkArray` method, I think it will be a separate pr.

beryllw:
LGTM

PR #2757, opened 2026-03-01
[server] Add configuration options for multiple remote data directories

### LINKED ISSUE #2753: [server] Add configuration options for multiple remote data directories
## Search before asking

## Motivation

**Use Case:**

Currently, Fluss only supports a single remote data directory via the `remote.data.dir` configuration option. In production environments, users often want to distribute KV snapshot data files and tiered log storage across multiple remote storage paths (e.g., different OSS buckets, different S3 prefixes, or separate remote filesystems) to achieve better I/O throughput, cost optimization, or fault isolation.

**Problem Statement:**

With a single `remote.data.dir`, all remote data is written to one path, which:
- Creates a bottleneck when a single bucket/filesystem cannot handle the throughput
- Prevents cost-tiering strategies (e.g., hot/cold buckets)
- Limits flexibility for multi-tenant or multi-region deployments

**Benefits:**
- Enables users to spread remote data across multiple storage backends for higher aggregate I/O throughput
- Supports weighted distribution to utilize buckets with different capacities proportionally
- Keeps backward compatibility — the existing `remote.data.dir` single-path config continues to work

## Solution

**Proposed Approach:**

Introduce three new configuration options alongside the existing `remote.data.dir`:

| Config Key | Type | Default | Description |
|---|---|---|---|
| `remote.data.dirs` | `List<String>` | (None) | List of remote data directory paths |
| `remote.data.dirs.strategy` | `Enum` | `ROUND_ROBIN` | Directory selection strategy: `ROUND_ROBIN` or `WEIGHTED_ROUND_ROBIN` |
| `remote.data.dirs.weights` | `List<Integer>` | (None) | Weights per directory for `WEIGHTED_ROUND_ROBIN`; size must equal `remote.data.dirs` |

**Example configuration:**

```yaml
# Round-robin across two OSS buckets
remote.data.dirs: oss://bucket1/fluss-remote-data, oss://bucket2/fluss-remote-data
remote.data.dirs.strategy: ROUND_ROBIN

# Weighted round-robin (2:1 ratio)
remote.data.dirs: oss://bucket1/fluss-remote-data, oss://bucket2/fluss-remote-data
remote.data.dirs.strategy: WEIGHTED_ROUND_ROBIN
remote.data.dirs.weights: 2, 1
```

## Anything else?

This change is backward-compatible. If `remote.data.dirs` is not set, the system continues to use the single `remote.data.dir` path as before. The routing/selection logic for writing to one of multiple dirs will be implemented as a follow-up task.

## Willingness to contribute

LiebingYu:
### Purpose

Linked issue: close #2753


website/docs/maintenance/configuration.md
```diff
@@ -36,6 +36,9 @@ during the Fluss cluster working.
...
+| remote.data.dirs                                    | List&lt;String&gt; | (None)                                                                                                                                                                   | The directories used for storing the kv snapshot data files and remote log for log tiered storage in a Fluss supported filesystem. This should be a comma-separated list of remote URIs. If not configured, it defaults to the path specified in `remote.data.dir`. Otherwise, one of the paths from this configuration will be used.                                                                                                                                                                                                                                                                                                                                                                              |
```
wuchong: Explain when (new table/partitions created) and what dir  (by `remote.data.dirs.strategy`) will be used, and the relationship between `remote.data.dirs` and `remote.data.dir` (when to configure which, what behavior when both confiugred).

fluss-common/src/main/java/org/apache/fluss/config/FlussConfigUtils.java
```diff
@@ -77,4 +81,109 @@ static Map<String, ConfigOption<?>> extractConfigOptions(String prefix) {
...
+                                    remoteDataDirs.size(),
+                                    ConfigOptions.REMOTE_DATA_DIRS_WEIGHTS.key(),
+                                    weights.size()));
+                }
+                // Validate all weights are positive
+                for (int i = 0; i < weights.size(); i++) {
+                    if (weights.get(i) < 0) {
+                        throw new IllegalConfigurationException(
+                                String.format(
+                                        "All weights in '%s' must be no less than 0, but found %d at index %d.",
```
wuchong, line 162: The condition `weights.get(i) < 0` should be `weights.get(i) <= 0`, and the error message should be "must be greater than 0".

fluss-common/src/main/java/org/apache/fluss/config/FlussConfigUtils.java
```diff
@@ -77,4 +81,109 @@ static Map<String, ConfigOption<?>> extractConfigOptions(String prefix) {
...
+                                ConfigOptions.REMOTE_DATA_DIRS.key(), i),
+                        e);
+            }
+        }
+
+        // Validate remote.data.dirs.strategy
+        ConfigOptions.RemoteDataDirStrategy remoteDataDirStrategy =
+                conf.get(ConfigOptions.REMOTE_DATA_DIRS_STRATEGY);
+        if (remoteDataDirStrategy == ConfigOptions.RemoteDataDirStrategy.WEIGHTED_ROUND_ROBIN) {
+            List<Integer> weights = conf.get(ConfigOptions.REMOTE_DATA_DIRS_WEIGHTS);
```
wuchong, line 145: We should check the `weights` is not empty when the strategy is `WEIGHTED_ROUND_ROBIN`. Otherwise, the validation will skip the check in the `!remoteDataDirs.isEmpty() && !weights.isEmpty()` code block..

fluss-common/src/main/java/org/apache/fluss/config/FlussConfigUtils.java
```diff
@@ -77,4 +81,109 @@ static Map<String, ConfigOption<?>> extractConfigOptions(String prefix) {
...
+        if (conf.get(ConfigOptions.LOG_SEGMENT_FILE_SIZE).getBytes() > Integer.MAX_VALUE) {
+            throw new IllegalConfigurationException(
+                    String.format(
+                            "Invalid configuration for %s, it must be less than or equal %d bytes.",
+                            ConfigOptions.LOG_SEGMENT_FILE_SIZE.key(), Integer.MAX_VALUE));
+        }
+    }
+
+    /** Validate common server configs. */
+    private static void validServerConfigs(Configuration conf) {
```
wuchong: rename to `validateServerConfigs`

fluss-common/src/main/java/org/apache/fluss/config/FlussConfigUtils.java
```diff
@@ -77,4 +81,109 @@ static Map<String, ConfigOption<?>> extractConfigOptions(String prefix) {
...
+        if (conf.get(ConfigOptions.REMOTE_DATA_DIR) == null) {
+            throw new IllegalConfigurationException(
```
wuchong: We should allow `remote.data.dirs` is set and `remote.data.dir` is empty.

fluss-common/src/main/java/org/apache/fluss/config/FlussConfigUtils.java
```diff
@@ -77,4 +81,109 @@ static Map<String, ConfigOption<?>> extractConfigOptions(String prefix) {
...
+    }
+
+    public static void validateTabletConfigs(Configuration conf) {
+        validServerConfigs(conf);
+
+        Optional<Integer> serverId = conf.getOptional(ConfigOptions.TABLET_SERVER_ID);
+        if (!serverId.isPresent()) {
+            throw new IllegalConfigurationException(
+                    String.format("Configuration %s must be set.", ConfigOptions.TABLET_SERVER_ID));
+        }
```
wuchong: I think all the configuration validation should be shared between Coordinator and TabletServer, except the `TABLET_SERVER_ID`. It is possible for TabletServer to use the `remote.data.dirs` rather than forcing TabletServer always use `remote.data.dir`. Same to other configurations. So could you refactor the config validation to extract the common validation for both Cooridnator and TabletServer?

fluss-common/src/main/java/org/apache/fluss/utils/FlussPaths.java
```diff
@@ -409,6 +409,23 @@ public static UUID uuidFromRemoteIndexCacheFileName(String fileName) {
...
+     * Returns the remote root directory path for storing data files.
+     *
+     * <p>The path contract:
+     *
+     * <pre>
+     * {$remote.data.dir}
+     * </pre>
+     */
+    public static FsPath remoteDataDir(Configuration conf) {
+        return new FsPath(conf.get(ConfigOptions.REMOTE_DATA_DIR));
```
wuchong: I think this method is very confusing, when user set `remote.data.dirs` rather than `remote.data.dir`, this method will fail or return the wrong result. It seems this method is only used by `FlussConfigUtils` for validate the path, could you remove this method, and just `new FsPath` in `FlussConfigUtils`?

wuchong (commented):
Thank @LiebingYu , I left some comments.

website/docs/maintenance/configuration.md
LiebingYu: Updated the description of `remote.data.dir` and `remote.data.dirs`.

fluss-common/src/main/java/org/apache/fluss/config/FlussConfigUtils.java
LiebingYu: I think we need 0. Imagine a scenario where the capacity of a remote storage has reached its limit and we don’t want to transfer any more files to it; in that case, we can set its weight to 0.

fluss-common/src/main/java/org/apache/fluss/config/FlussConfigUtils.java
LiebingYu: As I updated in the document. We should always set `remote.data.dir`, as the recently introduced producer offsets and kv snapshot lease files are not belong to a specific table. 
- kv snapshot lease dir: `{$remote.data.dir}/lease/kv-snapshot/{leaseId}/{tableId}/`
- producer offset dir: `{$remote.data.dir}/producers`

I think we must keep `remote.data.dir` to store producer offsets and kv snapshot lease files for now.

LiebingYu:
@wuchong Thanks for the comments, I've updated, please take a look again.

fluss-common/src/main/java/org/apache/fluss/config/ConfigOptions.java
```diff
@@ -96,8 +96,65 @@ public class ConfigOptions {
                     .stringType()
                     .noDefaultValue()
                     .withDescription(
-                            "The directory used for storing the kv snapshot data files and remote log for log tiered storage "
-                                    + " in a Fluss supported filesystem.");
+                            "The directory in a Fluss supported filesystem for remote data storage. "
+                                    + "This configuration is required. "
```
gyang94: This configuration is required now? suppose a fluss server should be able to run without a remote storage.
OK... see the comments above. 
> As I updated in the document. We should always set remote.data.dir, as the recently introduced producer offsets and kv snapshot lease files are not belong to a specific table.
> kv snapshot lease dir: {$remote.data.dir}/lease/kv-snapshot/{leaseId}/{tableId}/
> producer offset dir: {$remote.data.dir}/producers
> I think we must keep remote.data.dir to store producer offsets and kv snapshot lease files for now.
These features rely on the remote storage. 

My view is: generally we should allow the server starts when no remote storage configured. When users case need these two features, there will be an exception to tell users that they need to setup remote storage before move on.

Or we can have a default value for remote.data.dir: /tmp/fluss/remote-data, to avoid exceptions.

What do you think?

LiebingYu: In fact, Fluss currently relies on remote log tiering to perform local segment expiration. Therefore, `remote.data.dir` must be configured in production environments. 

As for the default value, we have already set it in `server.xml`. I don't think this will hinder users who want to perform a quick local trial.

gyang94: yes, agree with you. Our `server.xml` has a value, but in the ConfigOption definition of remote.data.dir is `noDefaultValue()`. How about set one here?

LiebingYu: I don't think we should set a default value in `ConfigOptions`. `remote.data.dir` is immutable, so if a user forgets to configure it, the system should fail fast rather than running with a default value. Providing a default in `ConfigOptions` might lead users to inadvertently use it, and once they realize it, they would find that `remote.data.dir` cannot be changed.

PR #2767, opened 2026-03-02
[server] Fix LogTieringTask to commit successfully copied segments on partial failure

### LINKED ISSUE #2766: [server] Partial segment copy failure in LogTieringTask will discard all successfully uploaded segments
### Search before asking

### Fluss version

main (development)

### Please describe the bug 🐞

In `LogTieringTask#runOnce()`, the method `copyLogSegmentFilesToRemote` copies candidate log segments to remote storage sequentially. If any single segment copy fails (e.g., due to rate limiting or a transient network error), a `RemoteStorageException` is thrown and propagated to the caller, causing the entire `runOnce() ` method to skip the commit phase. This means all previously successful segment uploads in the same batch are wasted — they are never committed to the remote log manifest and become orphaned files on remote storage. For example, if there are 10 candidate segments and the 6th fails due to rate limiting, the first 5 segments that were already successfully uploaded will not be committed and their work is lost.

### Solution

_No response_

swuferhong:
### Purpose

Linked issue: close #2766

### DIFF
--- a/fluss-server/src/main/java/org/apache/fluss/server/log/remote/LogTieringTask.java
+++ b/fluss-server/src/main/java/org/apache/fluss/server/log/remote/LogTieringTask.java
@@ -229,12 +229,6 @@ public class LogTieringTask implements Runnable {
                     highWatermark);
         }
 
-        // Limit the number of segments to upload per task execution to prevent
-        // overwhelming the remote storage when there is a large backlog.
-        if (candidateLogSegments.size() > maxUploadSegmentsPerTask) {
-            candidateLogSegments = candidateLogSegments.subList(0, maxUploadSegmentsPerTask);
-        }
-
         return candidateLogSegments;
     }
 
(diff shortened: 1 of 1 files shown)


wuchong (approved):

fluss-server/src/main/java/org/apache/fluss/server/log/remote/LogTieringTask.java
```diff
@@ -226,6 +229,12 @@ private List<EnrichedLogSegment> candidateToCopyLogSegments(LogTablet log) {
                     highWatermark);
         }
 
+        // Limit the number of segments to upload per task execution to prevent
+        // overwhelming the remote storage when there is a large backlog.
+        if (candidateLogSegments.size() > maxUploadSegmentsPerTask) {
+            candidateLogSegments = candidateLogSegments.subList(0, maxUploadSegmentsPerTask);
+        }
```
wuchong: I think we should only add `maxUploadSegmentsPerTask` number of segments in the method `candidateLogSegments(..)`, rather than remove segments after `candidateLogSegments(..)`, this avoids unncessary logics.

swuferhong:
@wuchong comments addressed.

wuchong (approved):

PR #2769, opened 2026-03-02
[server] Integrate remote directory selector into table/partition creation

### LINKED ISSUE #2755: [server] Integrate remote directory selector into table/partition creation
## Search before asking

## Description

**Task Overview:**

With the metadata models extended to carry a per-table/partition `remoteDataDir` and the multi-directory configuration options in place, the next step is to wire a directory selection mechanism into the Coordinator so that each newly created table or partition is deterministically assigned a remote data directory at creation time.

This task introduces a pluggable `RemoteDirSelector` interface with `ROUND_ROBIN` and `WEIGHTED_ROUND_ROBIN` implementations, wraps them in a hot-reloadable `RemoteDirDynamicLoader`, and integrates the selector into `CoordinatorService.createTable()` and `createPartition()` flows.

**Purpose:**

- Assign a concrete remote data directory to every table and partition upon creation, based on the configured selection strategy.
- Support dynamic reconfiguration (no restart) when `remote.data.dirs`, `remote.data.dirs.strategy`, or `remote.data.dirs.weights` change at runtime via `DynamicConfigManager`.
- Keep backward compatibility: when `remote.data.dirs` is empty, the selector falls back to the single `remote.data.dir`.

**Scope:**

### 1. RemoteDirSelector Interface & Implementations

- **`RemoteDirSelector`** (new `@ThreadSafe` interface): Defines `String nextDataDir()`. When `remote.data.dirs` is empty, implementations must return `remote.data.dir`.
- **`RoundRobinRemoteDirSelector`** (new class): Uses `AtomicInteger` to cycle through directories. Thread-safe without locking.
- **`WeightedRoundRobinRemoteDirSelector`** (new class): Implements the smooth (interleaved) weighted round-robin algorithm. Maintains `currentWeights[]` array; on each call, adds configured weights then picks the max, subtracts total weight from the winner. Synchronized on a lock object for thread safety.

### 2. RemoteDirDynamicLoader

- **`RemoteDirDynamicLoader`** (new class, implements `ServerReconfigurable`):
  - Holds a volatile `RemoteDirContainer` (inner class) that owns the active `RemoteDirSelector`.
  - `validateReconfigure(Configuration)`: validates that the new config is consistent (strategy + dirs + weights size match).
  - `reconfigure(Configuration)`: detects changes to any of the three config options and atomically replaces `remoteDirContainer` with a new one built from the merged config.
  - Inner class **`RemoteDirContainer`**: factory that reads `remote.data.dirs.strategy` and instantiates the appropriate `RemoteDirSelector`; exposes `nextDataDir()` and `getDefaultRemoteDataDir()`.

### 3. Coordinator Integration

- **`CoordinatorServer`**: Instantiates `RemoteDirDynamicLoader` and registers it with `DynamicConfigManager` alongside `LakeCatalogDynamicLoader`.
- **`CoordinatorService`**:
  - Accepts `RemoteDirDynamicLoader` via constructor injection.
  - In `createTable()`: calls `remoteDirDynamicLoader.getRemoteDataDirContainer().nextDataDir()` and passes the result to `metadataManager.createTable(tablePath, remoteDataDir, tableDescriptor, tableAssignment, ignoreIfExists)`.
  - In `createPartition()`: same selection, passes `remoteDataDir` to `metadataManager.createPartition(...)`.
- **`MetadataManager`**: Add `remoteDataDir` parameter to `createTable()` and `createPartition()`.
- **`AutoPartitionManager`**: Inject `RemoteDirDynamicLoader`; select a `remoteDataDir` for each auto-created partition.
- **`CreatePartitionEvent`**: Add `remoteDataDir` field to carry the selected directory through the event pipeline.
- **`TableChangeWatcher`**: Pass `remoteDataDir` when firing `CreatePartitionEvent`.

## Willingness to contribute

LiebingYu:
### Purpose

Linked issue: close #2755


fluss-server/src/main/java/org/apache/fluss/server/coordinator/remote/RemoteDirDynamicLoader.java
```diff
@@ -0,0 +1,230 @@
...
+
+    @Override
+    public void validate(Configuration newConfig) throws ConfigException {
+        // Get and valid remote data dirs
+        List<String> remoteDataDirs =
+                newConfig
+                        .getOptional(ConfigOptions.REMOTE_DATA_DIRS)
+                        .orElseGet(() -> currentConfiguration.get(ConfigOptions.REMOTE_DATA_DIRS));
+        for (int i = 0; i < remoteDataDirs.size(); i++) {
+            String remoteDataDir = remoteDataDirs.get(i);
```
gyang94: should we validate that the new remoteDataDirs contains all existing remoteDataDirs? Only allow to add new remote dirs, and refuse to remove any exsited remote dir.

fluss-server/src/main/java/org/apache/fluss/server/coordinator/MetadataManager.java
```diff
@@ -692,13 +696,14 @@ public Set<String> getPartitions(TablePath tablePath) {
     public void createPartition(
             TablePath tablePath,
             long tableId,
+            String remoteDataDir,
```
gyang94, line 806: I am thinking about  possible improvements here:
1. 
Will it be better to put the remote dir selection logic inside into the `MetadataManager.createPartition()` and `MetadataManager.createTable()`.

I notice that now there exist some code pattern that (a). select a remoteDataDir value. (b) pass the value into createTable/createPartition functions, when we need to createTable/partition otherwhere.
Put the step(a) inside those functions will make the code more focused and void adding a remoteDataDir parameter in these functions.

Or
2. 
Is is good to add the 'remoteDataDir' field into the `TablePath` class? Make the TablePath object carrys the remote dir info, and pass it through to everywhere. Then we don't need to care about adding an independent remoteDataDir parameter in functions which already has the TablePath parameter.

The implementation now is good to work. Just put my thoughts here. What do you think?

fluss-server/src/main/java/org/apache/fluss/server/coordinator/remote/RemoteDirDynamicLoader.java
LiebingYu: Aggree. I've updated.

fluss-server/src/main/java/org/apache/fluss/server/coordinator/MetadataManager.java
LiebingYu: 1. Put the selection into `MetadataManager` also need to change a lot. By comparison, I think the current approach would actually introduce smaller changes.

2. `TablePath` represents the logical path of a Fluss table, while `remoteDataDir` is the physical storage path in the implementation. I don’t think it’s appropriate to put `remoteDataDir` into `TablePath`.

fluss-server/src/main/java/org/apache/fluss/server/coordinator/remote/RemoteDirDynamicLoader.java
```diff
@@ -0,0 +1,167 @@
...
+    @Override
+    public void reconfigure(Configuration newConfig) throws ConfigException {
+        if (strategyChanged(newConfig)
+                || remoteDataDirsChanged(newConfig)
+                || weightsChanged(newConfig)) {
+            // Create a new container with the merged configuration
+            Configuration mergedConfig = mergeConfigurations(currentConfiguration, newConfig);
+            this.remoteDirSelector = createRemoteDirSelector(mergedConfig);
+            this.currentConfiguration = mergedConfig;
+        }
```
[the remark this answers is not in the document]
LiebingYu: No need to remove.

fluss-server/src/main/java/org/apache/fluss/server/coordinator/remote/RemoteDirDynamicLoader.java
```diff
@@ -0,0 +1,167 @@
...
+     * @param newConfig the new configuration
+     * @param option the configuration option to check
+     * @param <T> the type of the configuration value
+     * @return true if the configuration has changed
+     */
+    private <T> boolean hasConfigChanged(Configuration newConfig, ConfigOption<T> option) {
+        return newConfig
+                .getOptional(option)
+                .map(newValue -> !Objects.equals(newValue, currentConfiguration.get(option)))
+                .orElse(false);
```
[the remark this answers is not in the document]
LiebingYu: No deletions.

fluss-server/src/main/java/org/apache/fluss/server/coordinator/CoordinatorService.java
```diff
@@ -275,6 +278,7 @@ public CoordinatorService(
         this.ioExecutor = ioExecutor;
         this.lakeTableHelper =
                 new LakeTableHelper(zkClient, conf.getString(ConfigOptions.REMOTE_DATA_DIR));
+        this.remoteDirDynamicLoader = remoteDirDynamicLoader;
```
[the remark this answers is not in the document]
LiebingYu: Will resolve in another PR.

fluss-server/src/main/java/org/apache/fluss/server/coordinator/remote/RemoteDirDynamicLoader.java
```diff
@@ -0,0 +1,167 @@
...
+
+    @Override
+    public void validate(Configuration newConfig) throws ConfigException {
+        // Validate new remote data dirs contain all old remote data dirs
+        Optional<List<String>> newRemoteDataDirsOp =
+                newConfig.getOptional(ConfigOptions.REMOTE_DATA_DIRS);
+        if (newRemoteDataDirsOp.isPresent()) {
+            List<String> oldRemoteDataDirs =
+                    currentConfiguration.get(ConfigOptions.REMOTE_DATA_DIRS);
+            Set<String> newRemoteDataDirs = new HashSet<>(newRemoteDataDirsOp.get());
```
wuchong, line 77: We must account for `remote.data.dir` when handling `oldRemoteDataDirs`. When dynamically configuring a cluster to use `remote.data.dirs`, ensure that the original directory specified by `remote.data.dir` is included in `newRemoteDataDirs` to maintain data continuity.

fluss-server/src/test/java/org/apache/fluss/server/coordinator/remote/RemoteDirsITCase.java
```diff
@@ -0,0 +1,278 @@
...
+        }
+
+        // Verify round-robin distribution: each dir should be used at least once
+        Map<String, Integer> dirUsageCount = new HashMap<>();
+        for (String dir : remoteDirsUsed) {
+            dirUsageCount.merge(dir, 1, Integer::sum);
+        }
+
+        // With round-robin, all configured dirs should be used
+        assertThat(dirUsageCount.keySet()).hasSize(REMOTE_DIR_NAMES.size());
```
wuchong, line 133: nit: assert the usage count are all 2?

fluss-server/src/main/java/org/apache/fluss/server/coordinator/remote/RoundRobinRemoteDirSelector.java
```diff
@@ -0,0 +1,55 @@
...
+        this.position = new AtomicInteger(0);
+    }
+
+    @Override
+    public String nextDataDir() {
+        if (remoteDataDirs.isEmpty()) {
+            return remoteDataDir;
+        }
+
+        int index = position.getAndUpdate(i -> (i + 1) % remoteDataDirs.size());
```
wuchong:
```suggestion
        int index = position.getAndIncrement();
        return remoteDataDirs.get(index % remoteDataDirs.size());
```

Consider just increment and mod in one step to make the logic simple and more efficient.

wuchong (commented):
Thanks @LiebingYu, and apologies for the delayed review. The changes look good overall; I have left only a few minor comments.

Additionally, it would be great if we can add IT cases for dynamic configuration to enable more comprehensive end-to-end validation. I recommend adding these tests near `FlussAdminITCase#testDynamicConfigs()`. Specifically, please verify the transition from the old configuration to multiple remote directories using round-robin/weighted strategies, and confirm that `TableInfo#remoteDataDir` correctly distributes table data across the defined remote directories.

LiebingYu:
@wuchong Thanks for the feedback, I've addressed all the comments. PTAL!

wuchong (approved):

wuchong:
Thanks @LiebingYu , the PR LGTM. Could you help to resolve the conflicts?

PR #2780, opened 2026-03-03
[Feature] Support coordinator switch when leader fail

### LINKED ISSUE #2770: [Feature] Support coordinator switch when leader fail
### Search before asking

### Motivation

The main goal is to satisfy Coordinator high availability. 

I split it into 2 issues.

This is the first one, which aim to support coordinator switch when leader fail.

Parent issues : #188 

### Solution

see https://cwiki.apache.org/confluence/display/FLUSS/FIP-9%3A+Support+CoordinatorServer+High+Availability

### Anything else?

_No response_

### Willingness to contribute

zcoo:
### Purpose
This is the part 1 pr for coordinator high availability

Linked issue: close #2770

### Brief change log
see https://cwiki.apache.org/confluence/display/FLUSS/FIP-9%3A+Support+CoordinatorServer+High+Availability


fluss-common/src/main/java/org/apache/fluss/config/ConfigOptions.java
```diff
@@ -341,6 +341,12 @@ public class ConfigOptions {
                                     + " (“50100,50101”), ranges (“50100-50200”) or a combination of both."
                                     + "This option is deprecated. Please use bind.listeners instead, which provides a more flexible configuration for multiple ports");
 
+    public static final ConfigOption<Integer> COORDINATOR_ID =
+            key("coordinator.id")
+                    .intType()
```
swuferhong: As defined in the FIP, the `coordinator.id` should be a UUID instead of an integer. Please modify it.

fluss-server/src/main/java/org/apache/fluss/server/zk/data/ZkData.java
```diff
@@ -305,13 +305,44 @@ public static String path() {
...
+        public static String path() {
+            return "/coordinators/election";
+        }
+    }
+
+    /**
+     * The znode for the active coordinator leader. The znode path is:
      *
-     * <p>Note: introduce standby coordinators in the future for znode "/coordinators/standby/".
+     * <p>/coordinators/leader
```
swuferhong: `/coordinators/leader` or `/coordinators/active`？ The below path is `/coordinators/active`

zcoo: sorry for typo. I keep /coordinators/active as the same with origin path

zcoo:
@swuferhong thanks for your comments, I've solved the problem

fluss-server/src/main/java/org/apache/fluss/server/coordinator/event/watcher/CoordinatorChangeWatcher.java
```diff
@@ -0,0 +1,95 @@
...
+        return "CoordinatorChangeWatcher";
+    }
+
+    protected String getServerIdFromEvent(ChildData data) {
+        try {
+            return ZKPaths.getNodeFromPath(data.getPath());
+        } catch (NumberFormatException e) {
+            throw new FlussRuntimeException(
+                    "Invalid server id in zookeeper path: " + data.getPath(), e);
+        }
```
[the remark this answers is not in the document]
wuchong: +1, we should try catch general `Exception` or remove the retry catch here.

fluss-server/src/main/java/org/apache/fluss/server/coordinator/CoordinatorLeaderElection.java
```diff
@@ -0,0 +1,150 @@
...
+     * @param initLeaderServices the runnable to initialize leader services once elected
+     * @return a CompletableFuture that completes when this server becomes leader
+     */
+    public CompletableFuture<Void> startElectLeaderAsync(Runnable initLeaderServices) {
+        leaderLatch.addListener(
+                new LeaderLatchListener() {
+                    @Override
+                    public void isLeader() {
+                        LOG.info("Coordinator server {} has become the leader.", serverId);
+                        isLeader.set(true);
```
wuchong, line 135: The `isLeader` is never used which I think can be removed.

fluss-server/src/main/java/org/apache/fluss/server/coordinator/CoordinatorLeaderElection.java
```diff
@@ -0,0 +1,150 @@
...
+                new Thread(
+                        () -> {
+                            try {
+                                // todo: Currently, we await the leader latch and do nothing until
+                                // it becomes leader.
+                                // Later we can make it as a hot backup server to continuously
+                                // synchronize metadata from
+                                // Zookeeper, which save time from recovering context
+                                leaderLatch.await();
+                                doInitLeaderServices(initLeaderServices);
```
wuchong: This additional `electionThread` appears redundant. We can invoke `doInitLeaderServices(initLeaderServices)` directly within the `LeaderLatchListener#isLeader` callback, thereby eliminating the overhead of an extra thread. Furthermore, the current logic fails to address how the coordinator should re-enter the leader election state after transitioning from a leader to a follower, such as during network fluctuations or full GC events. By removing the dedicated thread and consolidating the logic within the `LeaderLatchListener`, the implementation becomes significantly simpler. Specifically, we can incorporate the logic to resume leader contention directly in the `notLeader` method, removing the need to manually start and manage a separate `electionThread`.

fluss-server/src/main/java/org/apache/fluss/server/coordinator/CoordinatorLeaderElection.java
```diff
@@ -0,0 +1,150 @@
...
+
+    public boolean isLeader() {
+        return this.isLeader.get();
+    }
+
+    private void relinquishLeadership() {
+        isLeader.set(false);
+        LOG.info("Coordinator server {} has been fenced.", serverId);
+
+        this.close();
```
wuchong: We shouldn't close the election, rather than the coordinator should re-enter election state.

fluss-server/src/main/java/org/apache/fluss/server/coordinator/CoordinatorLeaderElection.java
```diff
@@ -0,0 +1,150 @@
...
+                new LeaderLatchListener() {
+                    @Override
+                    public void isLeader() {
+                        LOG.info("Coordinator server {} has become the leader.", serverId);
+                        isLeader.set(true);
+                    }
+
+                    @Override
+                    public void notLeader() {
+                        relinquishLeadership();
```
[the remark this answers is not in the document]
wuchong: +1. We must ensure the coordinator re-enters the leader election state after transitioning from a leader to a follower. To achieve this, we should stop all leader services by executing the inverse operations of `startCoordinatorLeaderService`.

fluss-server/src/main/java/org/apache/fluss/server/coordinator/CoordinatorServer.java
```diff
@@ -170,10 +176,38 @@ public static void main(String[] args) {
...
+        this.leaderElectionFuture =
+                coordinatorLeaderElection.startElectLeaderAsync(
+                        () -> {
+                            try {
+                                startCoordinatorLeaderService();
+                            } catch (Exception e) {
+                                throw new RuntimeException(e);
+                            }
+                        });
+        return leaderElectionFuture;
```
wuchong: The retrun value is never used.

fluss-server/src/main/java/org/apache/fluss/server/coordinator/CoordinatorServer.java
```diff
@@ -170,10 +176,38 @@ public static void main(String[] args) {
...
+    }
+
+    protected void startCoordinatorLeaderService() throws Exception {
```
wuchong: nit: `startLeaderServices` would be simpler and clearer, and we can introduce `stopLeaderServices` for the inverse operation.

fluss-server/src/test/java/org/apache/fluss/server/coordinator/CoordinatorServerElectionTest.java
```diff
@@ -0,0 +1,153 @@
...
+
+import java.time.Duration;
+import java.util.Arrays;
+import java.util.List;
+import java.util.Objects;
+
+import static org.apache.fluss.testutils.common.CommonTestUtils.waitUntil;
+import static org.assertj.core.api.Assertions.assertThat;
+
+class CoordinatorServerElectionTest {
```
wuchong, line 42: We should add one more test for coordinator becomes from leader to follower, and it can still elect for leader.

website/docs/maintenance/observability/monitor-metrics.md
```diff
@@ -294,12 +294,17 @@ Some metrics might not be exposed when using other JVM implementations (e.g. IBM
   </thead>
   <tbody>
     <tr>
-       <th rowspan="24"><strong>coordinator</strong></th>
-      <td style={{textAlign: 'center', verticalAlign: 'middle' }} rowspan="10">-</td>
+      <th rowspan="25"><strong>coordinator</strong></th>
+      <td style={{textAlign: 'center', verticalAlign: 'middle' }} rowspan="11">-</td>
       <td>activeCoordinatorCount</td>
       <td>The number of active CoordinatorServer in this cluster.</td>
```
wuchong: nit: 


```suggestion
      <td>The number of active CoordinatorServer (only leader) in this cluster.</td>
```

fluss-server/src/main/java/org/apache/fluss/server/zk/ZooKeeperClient.java
```diff
@@ -178,20 +177,35 @@ public Optional<byte[]> getOrEmpty(String path) throws Exception {
     // Coordinator server
     // --------------------------------------------------------------------------------------------
 
-    /** Register a coordinator leader server to ZK. */
+    /** Register a coordinator server to ZK. */
+    public void registerCoordinatorServer(String coordinatorId) throws Exception {
+        String path = ZkData.CoordinatorIdZNode.path(coordinatorId);
+        zkClient.create().creatingParentsIfNeeded().withMode(CreateMode.EPHEMERAL).forPath(path);
```
wuchong: `registerCoordinatorServer()` creates `/coordinators/ids/{serverId}` with no payload, so these length > 0 guards are false for both create and delete notifications. As a result NewCoordinatorEvent and DeadCoordinatorEvent are never emitted, and the new    aliveCoordinatorCount gauge never reflects coordinators joining or leaving after startup. I suggest to serialize the coordinator address in it, so it can have data in it and it would be helpful for debug. Otherwise, we don't know who is the server id (as the server id are randomly generated).

fluss-server/src/main/java/org/apache/fluss/server/coordinator/CoordinatorServer.java
```diff
@@ -170,10 +176,38 @@ public static void main(String[] args) {
...
+        this.coordinatorLeaderElection = new CoordinatorLeaderElection(zkClient, serverId);
+        this.leaderElectionFuture =
+                coordinatorLeaderElection.startElectLeaderAsync(
+                        () -> {
+                            try {
+                                startCoordinatorLeaderService();
+                            } catch (Exception e) {
+                                throw new RuntimeException(e);
+                            }
+                        });
```
wuchong: [P1] The future returned here is only stored; nothing in production chains it to `onFatalError()` or `terminationFuture`. If the election thread later fails (for example `LeaderLatch.start()` fails, or `startCoordinatorLeaderService()` hits a bad bind/configuration during RPC or metadata initialization), `CoordinatorServer.start()` has already succeeded and the process stays alive without a usable coordinator instead of failing fast.

fluss-server/src/main/java/org/apache/fluss/server/coordinator/event/watcher/CoordinatorChangeWatcher.java
```diff
@@ -0,0 +1,95 @@
...
+                        if (newData != null && newData.getData().length > 0) {
+                            String serverId = getServerIdFromEvent(newData);
+                            LOG.info("Received CHILD_ADDED event for server {}.", serverId);
+                            eventManager.put(new NewCoordinatorEvent(serverId));
+                        }
+                        break;
+                    }
+                case NODE_DELETED:
+                    {
+                        if (oldData != null && oldData.getData().length > 0) {
```
wuchong, line 77: [P3] `registerCoordinatorServer()` creates `/coordinators/ids/{serverId}` with no payload, so these `length > 0` guards are false for both create and delete notifications. As a result `NewCoordinatorEvent` and `DeadCoordinatorEvent` are never emitted, and the new `aliveCoordinatorCount` gauge never reflects coordinators joining or leaving after startup.

fluss-server/src/main/java/org/apache/fluss/server/coordinator/event/watcher/CoordinatorChangeWatcher.java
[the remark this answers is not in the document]
zcoo: thanks, there was a copy error here

fluss-server/src/main/java/org/apache/fluss/server/coordinator/CoordinatorLeaderElection.java
zcoo: It is used in CoordinatorService later for judge leader

fluss-server/src/main/java/org/apache/fluss/server/zk/ZooKeeperClient.java
zcoo: ok

fluss-server/src/main/java/org/apache/fluss/server/coordinator/CoordinatorLeaderElection.java
zcoo: In my first design, I didn't create a election thread. However, I encounter a ZK deadlock problem. It happened like this:
Curator event thread 
→ LeaderLatchListener.isLeader()
  → initLeaderServices.run()
    → initCoordinatorLeader()
      → coordinatorEventProcessor.startup()
        → initCoordinatorContext()
          → zooKeeperClient.getSortedTabletServerList() and other ZK request

LeaderLatchListener callback run in Curator inner event process thread. In this thread, execute ZK requests(like getData、getChildren) need the event thread itself to process response, thus cause deadlock. 

So I think creating a new thread to process "initLeaderServices" and "cleanupLeaderServices" is necessary.

zcoo:
@wuchong Thanks for review. Comment addressed. PLAT~

wuchong:
I pushed a commit to improve some code and squash commits. 

1. `@RequireCoordinatorLeader` annotation has no enforcement 
It's purely documentary, requiring developers to manually add `checkLeader()`. This is error-prone (as proven by the `alterDatabase` miss to add the `checkLeader`). I refactored it into check leader in `FlussRequestHandler#processRequest` for all requests to `CoordinatorService`. This also removes the `RequireCoordinatorLeader` annotation class.

2. `leaderReadyFuture` in CoordinatorLeaderElection` is not necessary. 

3. Refactor `CoordinatorServer` to reuse the zk retry loop logic

4. Refactor `CoordinatorLeaderElection` to reuse a cached thread pool, rather than start `Thread` instance. 


I will merge this PR if you don't have any objections.

PR #2781, opened 2026-03-03
[Feature] Coordinator Server Supports coordinator epoch protect

### LINKED ISSUE #2778: [Feature] Support Epoch and ZkEpoch protect for Coordinator leader change
### Search before asking

### Motivation

The main goal is to satisfy Coordinator high availability.

I split it into 2 issues.

This is the second one, which aim to protect Zookeeper request and tablet server request when coordinator leader change.

Parent issues : https://github.com/apache/fluss/issues/188

### Solution

see https://cwiki.apache.org/confluence/display/FLUSS/FIP-9%3A+Support+CoordinatorServer+High+Availability

### Anything else?

_No response_

### Willingness to contribute

zcoo:
### Purpose

Linked issue: close #2778

This is the part 2 pr for coordinator high availability focusing on "coordinator epoch" logic.
Will ready for review when part 1 finish and merge.
### Brief change log
see https://cwiki.apache.org/confluence/display/FLUSS/FIP-9%3A+Support+CoordinatorServer+High+Availability


fluss-server/src/main/java/org/apache/fluss/server/coordinator/CoordinatorServer.java
```diff
@@ -259,6 +288,12 @@ protected void startServices() throws Exception {
                     new AutoPartitionManager(metadataCache, metadataManager, conf);
             autoPartitionManager.start();
 
+            int ioExecutorPoolSize = conf.get(ConfigOptions.COORDINATOR_IO_POOL_SIZE);
+            this.ioExecutor =
+                    Executors.newFixedThreadPool(
+                            ioExecutorPoolSize, new ExecutorThreadFactory("coordinator-io"));
+
```
swuferhong: `ioExecutor` is initialized twice; it was already serialized once previously. This will cause the first thread pool to never be shutdown.

fluss-server/src/main/java/org/apache/fluss/server/coordinator/CoordinatorLeaderElection.java
```diff
@@ -0,0 +1,177 @@
...
+        electionThread.start();
+
+        return leaderReadyFuture;
+    }
+
+    public void doInitLeaderServices(Runnable initLeaderServices) {
+        try {
+            // to avoid split-brain
+            Optional<ZkEpoch> optionalEpoch = zkClient.fenceBecomeCoordinatorLeader(serverId);
+            optionalEpoch.ifPresent(
```
swuferhong: The handling when returning `Optional.empty()` here is problematic? When `BadVersionException` is caught,
    `fenceBecomeCoordinatorLeader` returns `Optional.empty()`, but `doInitLeaderServices` only relinquishes leadership on
    `CoordinatorEpochFencedException`. If the epoch update fails (returning empty), the code continues to execute
    `initLeaderServices.run()`, causing a coordinator that did not successfully acquire the epoch to still run as a leader, which may result in a split-brain.

fluss-server/src/main/java/org/apache/fluss/server/zk/ZooKeeperClient.java
```diff
@@ -178,20 +189,105 @@ public Optional<byte[]> getOrEmpty(String path) throws Exception {
...
+    /** Gets the list of coordinator server Ids. */
+    public int[] getCoordinatorServerList() throws Exception {
+        List<String> coordinatorServers = getChildren(ZkData.CoordinatorIdsZNode.path());
+        return coordinatorServers.stream().mapToInt(Integer::parseInt).toArray();
+    }
+
+    /** Ensure epoch znode exists. */
+    public void ensureEpochZnodeExists() throws Exception {
+        String path = ZkData.CoordinatorEpochZNode.path();
+        if (zkClient.checkExists().forPath(path) == null) {
```
swuferhong, line 329: If two `coordinators` execute almost simultaneously and both pass `checkExists == null`, the first one creates successfully, while the second one will throw `NodeExistsException` which is wrapped as `RuntimeException` causing startup failure. So I think `NodeExistsException` should be ignored here, not thrown `RuntimeException`.

fluss-server/src/main/java/org/apache/fluss/server/coordinator/CoordinatorContext.java
```diff
@@ -67,6 +68,7 @@ public class CoordinatorContext {
     // a success deletion.
     private final Map<TableBucketReplica, Integer> failDeleteNumbers = new HashMap<>();
 
+    private final Set<Integer> liveCoordinatorServers = new HashSet<>();
```
swuferhong: `resetContext()` need to clear `liveCoordinatorServers`?  also need to clear `coordinatorEpochZkVersion`.

fluss-server/src/main/java/org/apache/fluss/server/coordinator/CoordinatorLeaderElection.java
```diff
@@ -0,0 +1,177 @@
...
+                                leaderReadyFuture.completeExceptionally(e);
+                            }
+                        },
+                        "coordinator-leader-election-" + serverId);
+        electionThread.start();
+
+        return leaderReadyFuture;
+    }
+
+    public void doInitLeaderServices(Runnable initLeaderServices) {
```
swuferhong: private ?

fluss-server/src/main/java/org/apache/fluss/server/coordinator/CoordinatorLeaderElection.java
```diff
@@ -0,0 +1,177 @@
...
+            Optional<ZkEpoch> optionalEpoch = zkClient.fenceBecomeCoordinatorLeader(serverId);
+            optionalEpoch.ifPresent(
+                    integer ->
+                            coordinatorContext.setCoordinatorEpochAndZkVersion(
+                                    optionalEpoch.get().getCoordinatorEpoch(),
+                                    optionalEpoch.get().getCoordinatorEpochZkVersion()));
+        } catch (CoordinatorEpochFencedException e) {
+            relinquishLeadership();
+            LOG.warn(
+                    "Coordinator server {} has been fence and not become leader successfully.",
```
swuferhong: `has been fence` -> `has been fenced`.

fluss-common/src/main/java/org/apache/fluss/config/ConfigOptions.java
```diff
@@ -392,6 +392,12 @@ public class ConfigOptions {
                                     + " (“50100,50101”), ranges (“50100-50200”) or a combination of both."
                                     + "This option is deprecated. Please use bind.listeners instead, which provides a more flexible configuration for multiple ports");
 
+    public static final ConfigOption<Integer> COORDINATOR_ID =
```
wuchong: unused, remove

fluss-server/src/main/java/org/apache/fluss/server/coordinator/CoordinatorEventProcessor.java
wuchong, line 1635: this partitioned-table reassignment path still updates ZK without a coordinator epoch / version check, so a fenced coordinator can overwrite `/partitionIds/<partitionId>` after leadership changes.

fluss-server/src/main/java/org/apache/fluss/server/coordinator/CoordinatorLeaderElection.java
```diff
@@ -122,7 +130,25 @@ public void isLeader() {
...
+                                        optionalEpoch.ifPresent(
+                                                integer ->
+                                                        coordinatorContext
+                                                                .setCoordinatorEpochAndZkVersion(
+                                                                        optionalEpoch
+                                                                                .get()
+                                                                                .getCoordinatorEpoch(),
+                                                                        optionalEpoch
+                                                                                .get()
+                                                                                .getCoordinatorEpochZkVersion()));
```
wuchong: The `coordinatorContext` is not thread-safe and must only be updated and accessed within the `CoordinatorEventProcessor`. Since `leaderCallbackExecutor` operates on an asynchronous thread, mutating the context there violates thread safety. 

To resolve this, move the `zkClient.fenceBecomeCoordinatorLeader` call into `org.apache.fluss.server.coordinator.CoordinatorServer#initCoordinatorLeader`, placing it before the synchronized lock block. Similarly, move `registerCoordinatorLeader()` before the synchronized block. Capture the `ZkEpoch` returned by `fenceBecomeCoordinatorLeader` and pass it to the `CoordinatorEventProcessor` constructor to represent the current leader epoch. 

This allows `CoordinatorEventProcessor.startup()` to safely initialize `CoordinatorContext` with the `ZkEpoch`, as the operation remains within the single-threaded execution scope of the processor.

fluss-server/src/main/java/org/apache/fluss/server/coordinator/CoordinatorLeaderElection.java
```diff
@@ -122,7 +130,25 @@ public void isLeader() {
...
                                         initLeaderServices.run();
+                                    } catch (CoordinatorEpochFencedException e) {
+                                        LOG.warn(
+                                                "Coordinator server {} has been fenced and not become leader successfully.",
+                                                serverId);
+                                        throw e;
```
wuchong: When encountering a `FencedException`, we need to transition the current coordinator back to the standby state, for example, by invoking `notLeader()`.

fluss-server/src/main/java/org/apache/fluss/server/coordinator/CoordinatorContext.java
```diff
@@ -709,11 +720,12 @@ private void clearTablesState() {
     public void resetContext() {
         tablesToBeDeleted.clear();
         coordinatorEpoch = 0;
```
wuchong: nit: use `INITIAL_COORDINATOR_EPOCH` like how you updated `coordinatorEpochZkVersion`.

fluss-server/src/main/java/org/apache/fluss/server/coordinator/CoordinatorLeaderElection.java
```diff
@@ -122,7 +130,25 @@ public void isLeader() {
...
+                                                        coordinatorContext
+                                                                .setCoordinatorEpochAndZkVersion(
+                                                                        optionalEpoch
+                                                                                .get()
+                                                                                .getCoordinatorEpoch(),
+                                                                        optionalEpoch
+                                                                                .get()
+                                                                                .getCoordinatorEpochZkVersion()));
                                         initLeaderServices.run();
+                                    } catch (CoordinatorEpochFencedException e) {
```
wuchong, line 131: The catch block below is dead code because `fenceBecomeCoordinatorLeader()` never throws `CoordinatorEpochFencedException` -- it returns `Optional.empty()` on `BadVersionException` instead.

fluss-server/src/test/java/org/apache/fluss/server/testutils/FlussClusterExtension.java
```diff
@@ -200,6 +200,10 @@ public void afterEach(ExtensionContext extensionContext) throws Exception {
             }
         }
         CompletableFuture.allOf(dropFutures.toArray(new CompletableFuture[0])).join();
+
+        for (TabletServer tabletServer : tabletServers.values()) {
+            tabletServer.getReplicaManager().resetCoordinatorEpoch();
```
wuchong: Why is it necessary to manually reset the coordinator epoch? Would tests fail if this step were omitted? Since there are no coordinator restarts before or after each test, the coordinator epoch should remain unchanged. Even in the event of a coordinator restart, the TabletServer should receive the updated coordinator epoch via `UpdateMetadata`, rendering manual epoch resets unnecessary.

fluss-server/src/main/java/org/apache/fluss/server/zk/ZooKeeperClient.java
```diff
@@ -197,6 +209,43 @@ public void registerCoordinatorServer(CoordinatorAddress coordinatorAddress) thr
...
+                    newCoordinatorEpoch);
+
+            // atomically update epoch
+            zkClient.setData()
+                    .withVersion(currentCoordinatorEpochZkVersion)
+                    .forPath(
+                            ZkData.CoordinatorEpochZNode.path(),
+                            ZkData.CoordinatorEpochZNode.encode(newCoordinatorEpoch));
+
+            return Optional.of(getEpoch.nextZkEpoch());
```
wuchong: The `ZkEpoch.nextZkEpoch()` simply increments ZK version by exactly 1. While this is guaranteed by ZK spec for a successful conditional `setData`, it would be more defensive to capture
the `Stat` returned by `setData()`.


```java
Stat stat =
                    zkClient.setData()
                            .withVersion(currentCoordinatorEpochZkVersion)
                            .forPath(
                                    ZkData.CoordinatorEpochZNode.path(),
                                    ZkData.CoordinatorEpochZNode.encode(newCoordinatorEpoch));

            return Optional.of(new ZkEpoch(newCoordinatorEpoch, stat.getVersion()));
```

fluss-server/src/main/java/org/apache/fluss/server/zk/ZooKeeperClient.java
```diff
@@ -346,15 +428,16 @@ public void updatePartitionAssignment(long partitionId, PartitionAssignment part
...
+        deleteRecursiveWithEpochCheck(path, expectedZkVersion, false);
         LOG.info("Deleted table assignment for table id {}.", tableId);
     }
 
-    public void deletePartitionAssignment(long partitionId) throws Exception {
+    public void deletePartitionAssignment(long partitionId, int expectedZkVersion)
+            throws Exception {
         String path = PartitionIdZNode.path(partitionId);
-        zkClient.delete().deletingChildrenIfNeeded().forPath(path);
+        deleteRecursiveWithEpochCheck(path, expectedZkVersion, false);
```
wuchong: `deleteTableAssignment` and `deletePartitionAssignment` don't need epoch check. The only production caller is `MetadataManager.completeDeleteTable/completeDeletePartition`, which already passes `MATCH_ANY_VERSION` — so the epoch check is effectively a no-op today.

More importantly, these are **idempotent cleanup operations** at the end of the DDL delete flow. By the time they're called, all replicas are already in `DeletionSuccessful` state. Even if a stale coordinator executes this cleanup, it won't corrupt any live metadata — the table/partition is already marked for deletion, and the new leader won't recreate the assignment for it.

Since epoch check provides no safety benefit here, I'd suggest removing the `expectedZkVersion` parameter from these two methods and reverting to the original `zkClient.delete().deletingChildrenIfNeeded().forPath(path)`. This avoids the significant performance overhead of `deleteRecursiveWithEpochCheck` (which turns 1 ZK call into O(2N) calls for a path with N children) without sacrificing correctness.

If we can remove epoch check on these 2 methods, we can also remove the `deleteRecursiveWithEpochCheck()` method.

fluss-server/src/main/java/org/apache/fluss/server/zk/ZooKeeperClient.java
```diff
@@ -471,18 +565,21 @@ public void batchUpdateLeaderAndIsr(Map<TableBucket, LeaderAndIsr> leaderAndIsrL
...
         }
         if (!ops.isEmpty()) {
-            zkClient.transaction().forOperations(ops);
+            List<CuratorOp> wrapOps = wrapRequestsWithEpochCheck(ops, expectedZkVersion);
+            zkClient.transaction().forOperations(wrapOps);
         }
     }
 
-    public void deleteLeaderAndIsr(TableBucket tableBucket) throws Exception {
+    public void deleteLeaderAndIsr(TableBucket tableBucket, int expectedZkVersion)
```
wuchong: `deleteLeaderAndIsr` is only used in test, so we don't need epoch check, and we can change the visible modifier to `protected`.

fluss-server/src/main/java/org/apache/fluss/server/zk/ZooKeeperClient.java
```diff
@@ -127,11 +129,19 @@
...
 /**
  * This class includes methods for write/read various metadata (leader address, tablet server
  * registration, table assignment, table, schema) in Zookeeper.
+ *
+ * <p>In some method, 'expectedZkVersion' is used to execute an epoch Zookeeper version check. We
+ * have the following principals to judge if it's necessary to execute epoch Zookeeper version
+ * check. If all condition met, we need to execute epoch Zookeeper version check. 1. The method
+ * create/modify/delete Zk node. 2. It's executed by coordinator server. 3. It is about
+ * metadata(table/partition/leaderAndIsr) rather than server info or ACL info. 4. The Zk node is
+ * persistent rather than ephemeral.
```
wuchong: I think current epoch check is not clear about which type of operations should be added epoch check. I think not all zk nodes need the epoch check, only the assignment and LeaderAndIsr nodes need. So, could you update the principles with following? 


```
Conditions requiring epoch checks (all must be met):
┌─────────────────────────────────────────────────────────┐
│ 1. Invoked by the Coordinator (not the TabletServer)    │
│ 2. Operates on persistent nodes (not ephemeral)         │
│ 3. Constitutes a "control plane" operation:             │
│    partition assignment or LeaderAndIsr election        │
│ 4. Concurrent access to the same path by old and new    │
│    leaders during leader failover                       │
│ 5. No other mechanisms (optimistic locking,             │
│    idempotency, or reloading) provide fallback          │
└─────────────────────────────────────────────────────────┘

In practice, only two types of operations truly require this:
- CRUD for Table/Partition Assignment (assignment decisions).
- CRUD for LeaderAndIsr (leader election results).

These operations are inevitably executed concurrently by the old and new coordinators during failover (as the new leader immediately reassigns partitions), and overwrites cannot be automatically recovered.

All other operations do not require epoch checks because:
- DDL operations are protected against concurrency by client reconnection mechanisms.
- TabletServer operations are unaffected by coordinator failovers.
- ACLs and Configs have their own version control or idempotency guarantees.
- Ephemeral nodes are managed via session lifecycle.
```

fluss-server/src/main/java/org/apache/fluss/server/zk/ZooKeeperClient.java
```diff
@@ -274,13 +352,13 @@ public int[] getSortedTabletServerList() throws Exception {
     // --------------------------------------------------------------------------------------------
 
     /** Register table assignment to ZK. */
-    public void registerTableAssignment(long tableId, TableAssignment tableAssignment)
-            throws Exception {
+    public void registerTableAssignment(
+            long tableId, TableAssignment tableAssignment, int expectedZkVersion) throws Exception {
```
wuchong: This is a special case that we don't need the epoch check, because it is only invoked when table creation and the unique tableId can has natural concurrent guard (table ID conflicts, node existence).

fluss-server/src/main/java/org/apache/fluss/server/zk/ZooKeeperClient.java
wuchong: `updatePartitionAssignment` is not epoch-fenced, A stale coordinator can overwrite partition assignments after leadership changes.

fluss-server/src/test/java/org/apache/fluss/server/zk/ZooKeeperClientTest.java
```diff
@@ -193,23 +197,26 @@ void testLeaderAndIsr() throws Exception {
...
-        LeaderAndIsr leaderAndIsr2 = new LeaderAndIsr(2, 10, Arrays.asList(4, 5, 6), 100, 1000);
+        LeaderAndIsr leaderAndIsr1 = new LeaderAndIsr(1, 10, Arrays.asList(1, 2, 3), 0, 1000);
+        LeaderAndIsr leaderAndIsr2 = new LeaderAndIsr(2, 10, Arrays.asList(4, 5, 6), 0, 1000);
 
-        zookeeperClient.registerLeaderAndIsr(tableBucket1, leaderAndIsr1);
-        zookeeperClient.registerLeaderAndIsr(tableBucket2, leaderAndIsr2);
+        zookeeperClient.registerLeaderAndIsr(
+                tableBucket1, leaderAndIsr1, ZkVersion.MATCH_ANY_VERSION.getVersion());
+        zookeeperClient.registerLeaderAndIsr(
+                tableBucket2, leaderAndIsr2, ZkVersion.MATCH_ANY_VERSION.getVersion());
```
wuchong: It appears that all tests involving ZK epoch versions currently use `MATCH_ANY_VERSION`, which fails to validate the behavior of the production code. I suggest creating the ZK epoch node in `beforeAll()` and using the actual ZK epoch version in the test logic to ensure accurate verification.

fluss-server/src/main/java/org/apache/fluss/server/zk/ZooKeeperClient.java
zcoo: yes thank you for detailed description

zcoo:
@wuchong @swuferhong Thanks for all your comments. Now they are all addressed. PLAT~

zcoo:
@wuchong To improve test coverage for coordinator epoch propagation to TabletServer/Zookeeper after leader switch, I just add 2 test cases in CoordinatorHighAvailabilityITCase:
```
testTabletServerRejectsStaleCoordinatorEpochAfterLeaderSwitch
testZooKeeperRejectsStaleCoordinatorRequestAfterLeaderSwitch
```

wuchong:
I pushed a commit fixes a **race condition in leader election** and **refactors coordinator epoch to be immutable**.

### Key Changes

- **`CoordinatorLeaderElection`**: Moved `isLeader.set(true)` from **before** to **after** `initLeaderServices.run()`, preventing the coordinator from accepting requests before initialization completes. Removed unnecessary `zkClient` and `coordinatorContext` fields.

- **`CoordinatorContext`**: Made `coordinatorEpoch` and `coordinatorEpochZkVersion` **final** fields initialized via `ZkEpoch` in the constructor. Removed the mutable setter. Renamed `getCoordinatorEpochZkVersion()` → `getCoordinatorZkVersion()`.

- **`CoordinatorServer`**: `CoordinatorContext` is now created **locally** in `initCoordinatorLeader()` per election cycle instead of being a long-lived server field. Removed redundant `resetContext()` calls from cleanup/close paths.

- **`CoordinatorEventProcessor`**: Removed `ZkEpoch` constructor parameter (epoch now comes via `CoordinatorContext`). Added `resetContext()` in `close()`.

- **`ReplicaManager`**: Added `getCoordinatorEpoch()` getter and epoch update logging; removed `resetCoordinatorEpoch()`.

- **`ZkEpoch`**: Added `INITIAL_EPOCH` constant for test convenience.

- **`ZooKeeperClient`**: Fixed Javadoc formatting (inline text → `<pre>` and `<ul>` blocks).

- **Tests**: Deleted `TestCoordinatorContext`; updated HA tests to `waitUntil` leader is ready (since `isLeader` is now set after init); replaced manual `UpdateMetadataRequest` workaround with natural epoch propagation wait.
- **ITCase**: Updated `CoordinatorHighAvailabilityITCase#testTabletServerRejectsStaleCoordinatorEpochAfterLeaderSwitch` to update the coordinator epoch of TabletServer by coordinator itself, rather than manually update it in IT case.

Issue #2564, opened 2026-02-03
[website] Remove external resource for Google fonts


wuchong:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/fluss/issues) and found nothing similar.

### Description

> Found 1 external resources: {"ERROR Refused to load the stylesheet 'https://fonts.googleapis.com/css2?family=Inter:wght@400;500;600;700;800&display=swap' because it violates the following Content Security Policy directive: \"style-src 'self' data: blob: 'unsafe-inline' 'unsafe-eval' https://www.apachecon.com/ https://www.communityovercode.org/ https://*.apache.org/ https://apache.org/ https://*.scarf.sh/ https://www.youtube-nocookie.com https://www.youtube.com https://*.algolia.net/ https://*.algolianet.com/ https://*.algolia.io/ https://calendar.google.com\". Note that 'style-src-elem' was not explicitly set, so 'style-src' is used as a fallback."=>1}

>Text of a link expected to match regular expression: Found \d+ external resources
[Websites must not link to externally hosted resources](https://privacy.apache.org/faq/committers.html)

https://whimsy.apache.org/pods/project/fluss complain that we fail at the "Resource" check type. 

### Willingness to contribute

- [ ] I'm willing to submit a PR!

PR #2788, opened 2026-03-03
[blog] Fluss for ai

polyzos:


website/src/css/custom.css
```diff
@@ -21,17 +21,32 @@
  * work well for content-centric websites.
  */
 
+/* Import Inter font from Google Fonts */
+@import url('https://fonts.googleapis.com/css2?family=Inter:wght@400;500;600;700;800&display=swap');
+/* Import 'Roboto' */
+@import url('https://fonts.googleapis.com/css2?family=Roboto:wght@400;500;700&display=swap');
```
wuchong: We shouldn't introduce external resources, see https://github.com/apache/fluss/issues/2564

wuchong (commented):
Btw, please add `<!-- truncate -->` to the blog content to truct content in blog home page.

website/src/css/custom.css
```diff
@@ -219,6 +261,7 @@
             font-size: 14px;
             color: #4c576c;
         }
+        */
```
wuchong: use the same css for the code block of `ol` and `ul`?  Currently, the code format in list is not consistent with other code blocks.

Image

website/src/css/custom.css
wuchong: I like the new fonts. But we can't rely on external resources, maybe we can download the Roboto fonts, and serving them directly from fluss website.

Issue #2790, opened 2026-03-04
[Bug][Lake][Paimon] Fix the situation where resources may not be free up when reading paimon


ThorneANN:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/fluss/issues) and found nothing similar.

### Fluss version

0.9.0 (latest release)

### Please describe the bug 🐞

  Issue:

  In the constructor, after tableRead.createReader() creates a RecordReader and converts it to a CloseableIterator via toCloseableIterator(), if the subsequent
  PaimonRowAsFlussRecordIterator construction throws an exception, the CloseableIterator resource is never closed, causing a resource leak.

```
  org.apache.paimon.reader.RecordReader<InternalRow> recordReader =
          tableRead.createReader(split.dataSplit());
  iterator =
          new PaimonRecordReader.PaimonRowAsFlussRecordIterator(
                  recordReader.toCloseableIterator(), paimonRowType);
```

  Root Cause:

  The toCloseableIterator() call returns a resource that must be closed. If an exception occurs during the PaimonRowAsFlussRecordIterator construction, the constructor fails before
  the iterator is assigned to the field, leaving no way for the caller to close the leaked resource.

### Solution

Extract the CloseableIterator to a local variable and wrap the construction in a try-catch block to ensure proper cleanup on failure:

 ```
 org.apache.paimon.reader.RecordReader<InternalRow> recordReader =
          tableRead.createReader(split.dataSplit());
  org.apache.paimon.utils.CloseableIterator<InternalRow> closeableIterator =
          recordReader.toCloseableIterator();
  try {
      iterator =
              new PaimonRecordReader.PaimonRowAsFlussRecordIterator(
                      closeableIterator, paimonRowType);
  } catch (Throwable t) {
      try {
          closeableIterator.close();
      } catch (Exception closeException) {
          t.addSuppressed(closeException);
      }
      throw t;
  }
```

### Are you willing to submit a PR?

- [x] I'm willing to submit a PR!

PR #2791, opened 2026-03-04
[Lake][Paimon] Fix the situation resources free up when reading paimon

ThorneANN:
### Purpose

Linked issue: https://github.com/apache/fluss/issues/2790

### DIFF
--- a/fluss-lake/fluss-lake-paimon/src/main/java/org/apache/fluss/lake/paimon/source/PaimonRecordReader.java
+++ b/fluss-lake/fluss-lake-paimon/src/main/java/org/apache/fluss/lake/paimon/source/PaimonRecordReader.java
@@ -75,9 +75,20 @@ public class PaimonRecordReader implements RecordReader {
         } else {
             org.apache.paimon.reader.RecordReader<InternalRow> recordReader =
                     tableRead.createReader(split.dataSplit());
-            iterator =
-                    new PaimonRecordReader.PaimonRowAsFlussRecordIterator(
-                            recordReader.toCloseableIterator(), paimonRowType);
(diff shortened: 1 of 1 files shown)


ThorneANN:
@luoyuxia  cc

fluss-lake/fluss-lake-paimon/src/main/java/org/apache/fluss/lake/paimon/source/PaimonRecordReader.java
```diff
@@ -75,9 +75,20 @@ public PaimonRecordReader(
...
-            iterator =
-                    new PaimonRecordReader.PaimonRowAsFlussRecordIterator(
-                            recordReader.toCloseableIterator(), paimonRowType);
+            org.apache.paimon.utils.CloseableIterator<InternalRow> closeableIterator =
+                    recordReader.toCloseableIterator();
+            try {
+                iterator =
+                        new PaimonRecordReader.PaimonRowAsFlussRecordIterator(
+                                closeableIterator, paimonRowType);
+            } catch (Throwable t) {
```
luoyuxia, line 84: Thanks for the pr. Just curious about in which case it'll throw exception in the constuctor

ThorneANN: > Thanks for the pr. Just curious about in which case it'll throw exception in the constuctor

Tks,in the existing code, it is actually unlikely for the constructor to throw an exception, but this fix ensures that the closeableIterator will not be leaked in case the constructor is modified in the future or an unexpected runtime error occurs. It is a correct defensive coding practice.

PR #2799, opened 2026-03-05
[lake/paimon] Support Altering Lake Table Properties

### LINKED ISSUE #2795: [lake] Support Altering Lake Table Properties via ALTER TABLE SET
### Search before asking

### Description

**Task Overview:**

Currently, Fluss does not allow users to alter Lake table properties (i.e., properties prefixed with the datalake format name, such as `paimon.*`) via `ALTER TABLE ... SET (...)` statements. Any attempt to modify such properties results in an `InvalidConfigException`.

**Problem Statement:**

When a user attempts to update Lake-specific table properties, such as partition configuration, the following error is thrown:

```sql
ALTER TABLE my_catalog.my_database.my_lake_table SET (
    'paimon.partition.timestamp-formatter' = 'yyyyMMdd',
    'paimon.partition.timestamp-pattern' = '$ds'
);
```

```
Caused by: org.apache.fluss.exception.InvalidConfigException: 
  Property 'paimon.partition.timestamp-pattern' is not supported to alter which is for datalake table.
```

This blanket restriction prevents users from updating valid and necessary Lake table configurations after table creation, such as:

- `paimon.partition.timestamp-formatter`
- `paimon.partition.timestamp-pattern`

**Purpose:**

Users need the ability to adjust Lake table properties post-creation without having to drop and recreate tables, which is disruptive in production environments.

### Willingness to contribute

LiebingYu:
### Purpose

Linked issue: close #2795

### DIFF
--- a/fluss-lake/fluss-lake-paimon/src/main/java/org/apache/fluss/lake/paimon/utils/PaimonConversions.java
+++ b/fluss-lake/fluss-lake-paimon/src/main/java/org/apache/fluss/lake/paimon/utils/PaimonConversions.java
@@ -122,15 +122,14 @@ public class PaimonConversions {
         for (TableChange tableChange : tableChanges) {
             if (tableChange instanceof TableChange.SetOption) {
                 TableChange.SetOption setOption = (TableChange.SetOption) tableChange;
-                schemaChanges.add(
-                        SchemaChange.setOption(
-                                convertFlussPropertyKeyToPaimon(setOption.getKey()),
-                                setOption.getValue()));
+                String key = convertFlussPropertyKeyToPaimon(setOption.getKey());
+                validateAlterPaimonOptions(key);
+                schemaChanges.add(SchemaChange.setOption(key, setOption.getValue()));
             } else if (tableChange instanceof TableChange.ResetOption) {
                 TableChange.ResetOption resetOption = (TableChange.ResetOption) tableChange;
-                schemaChanges.add(
-                        SchemaChange.removeOption(
-                                convertFlussPropertyKeyToPaimon(resetOption.getKey())));
+                String key = convertFlussPropertyKeyToPaimon(resetOption.getKey());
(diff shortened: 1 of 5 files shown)


LiebingYu:
CC @luoyuxia

fluss-lake/fluss-lake-paimon/src/main/java/org/apache/fluss/lake/paimon/utils/PaimonConversions.java
```diff
@@ -275,6 +274,14 @@ private static void validatePaimonOptions(Map<String, String> properties) {
                 });
     }
 
+    private static void validateAlterPaimonOptions(String key) {
+        if (PAIMON_UNSETTABLE_OPTIONS.contains(key)
+                || CoreOptions.IMMUTABLE_OPTIONS.contains(key)) {
```
luoyuxia, line 279: `bucket.num` is not in `IMMUTABLE_OPTIONS`, but actually, in fluss case, it shouldn't change. Maybe we need to add more to unsupported options.

luoyuxia (commented):
@LiebingYu Thanks for the pr. Left minor comments. PTAL

fluss-lake/fluss-lake-paimon/src/main/java/org/apache/fluss/lake/paimon/utils/PaimonConversions.java
LiebingYu: Currently, Flink doesn't support to alter `bucket.num`.

If we support to alter `bucket.num` in the future, we should change `PAIMON_UNSETTABLE_OPTIONS ` to `PAIMON_UNALTERABLE_OPTIONS` here (because `bucket` exists in `PAIMON_UNSETTABLE_OPTIONS `). But I think we ok to use `PAIMON_UNSETTABLE_OPTIONS ` here for now?

PR #2802, opened 2026-03-06
[spark] refine the format when desc table (#2675)

### LINKED ISSUE #2675: [spark] refine the format when desc table
### Search before asking

### Motivation

Image

### Solution

_No response_

### Anything else?

_No response_

### Willingness to contribute

Akash073-hub:
## What this PR does
Fixes #2675

- Fix `name()` to return table name instead of `tableInfo.toString` which 
  was causing the full TableInfo blob to appear in DESC FORMATTED output
- Clean up `properties()` to only expose user-defined custom properties 
  via `getCustomProperties()`, preventing internal metadata (tableId, 
  schemaId, timestamps etc.) from polluting DESC FORMATTED output
- Remove unused imports (TablePath, FlussConfiguration)

### DIFF
--- a/fluss-spark/fluss-spark-common/src/main/scala/org/apache/fluss/spark/catalog/AbstractSparkTable.scala
+++ b/fluss-spark/fluss-spark-common/src/main/scala/org/apache/fluss/spark/catalog/AbstractSparkTable.scala
@@ -18,8 +18,7 @@
 package org.apache.fluss.spark.catalog
 
 import org.apache.fluss.client.admin.Admin
-import org.apache.fluss.config.{Configuration => FlussConfiguration}
-import org.apache.fluss.metadata.{TableInfo, TablePath}
+import org.apache.fluss.metadata.TableInfo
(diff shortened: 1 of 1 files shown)


Akash073-hub:
Hey @YannByron, I've submitted a PR for this issue!

Here's what I changed:
- Fixed `name()` to return the table name instead of `tableInfo.toString` which was causing the full `TableInfo` blob to appear in `DESC FORMATTED` output
- Cleaned up `properties()` to only expose user-defined custom properties by `getCustomProperties()`, preventing internal metadata (tableId, schemaId, timestamps etc.) from polluting the output
- Removed unused imports

Review once if you need any changes then tell me.Thank you!!

YannByron:
@Akash073-hub thanks for this. Can you add test cases to the code, or add a description in the PR so that we can know the current format clearly?

Akash073-hub:
Hey @YannByron, thanks for the feedback! I'll add test cases to verify the fix. Will update the PR shortly.

PR #2803, opened 2026-03-06
[client] Netty client prefer heap memory to cumulate in decoder.

### LINKED ISSUE #2773: Netty client prefers to use heap memory to decode message to reduce the direct memory use.
### Search before asking

###  Off-Heap Memory Allocation in Flink
Flink's default off-heap memory (taskmanager.memory.task.off-heap.size = 0 + taskmanager.memory.framework.off-heap.size = 128MB) totals 128MB. This is shared between the framework (e.g., Flink's internal operations) and third-party connectors (not only just  Fluss).

From Fluss' perspective, the recommended batch size is 2–5MB (raw data). However, during deserialization and decompression, the memory requirement increases to 10–15MB per batch. If a task has multiple slots, this demand multiplies, leading to severe off-heap memory contention.

**This limited off-heap memory leaves insufficient space for Netty clients**.

### Netty Client Memory Constraints

Flink's default log.replica.fetch.max-bytes is 16MB.A Netty client connecting to 4 servers would require at least 64MB of off-heap memory (16MB × 4).While we could throw an error to force users to increase off-heap memory, this approach is not user-friendly. Instead, we prioritize graceful degradation: if off-heap memory is insufficient, the system should read data more slowly rather than failing abruptly.

###  Proposed Configuration: Heap-First Allocation

I want to propose Heap-First Allocation for netty:
1. Flink's default heap memory is sufficiently large
2.  In NettyClientHandler, the record also will be copied as heap memory. Thus, why not just use PreferHeapByteBufAllocator in netty client(.
Image.

```java
-----------------                     
| network buffer ｜
-----------------
             |
         epoll a piece of network buffer
            |
-----------------                     
|  little direct buffer ｜
-------------------
           | 
       ByteToMessageDecoder.COMPOSITE_CUMULATOR
         | 
--------------------------------------------------                     
|           heap memory( the whole responce body       ｜
----------------------------------------------------




````
### Willingness to contribute

### PR DESCRIPTION
**Pull request by loserwang1024**
### Purpose

Linked issue: close #2773

### DIFF
--- a/fluss-server/src/test/java/org/apache/fluss/server/replica/fetcher/RemoteLeaderEndpointTest.java
+++ b/fluss-server/src/test/java/org/apache/fluss/server/replica/fetcher/RemoteLeaderEndpointTest.java
@@ -145,6 +145,8 @@ public class RemoteLeaderEndpointTest {
     private static Configuration initConfig() {
         Configuration conf = new Configuration();
         conf.setInt(ConfigOptions.DEFAULT_REPLICATION_FACTOR, 3);
+        // Server uses off-heap memory by default
+        conf.setBoolean(ConfigOptions.NETTY_CLIENT_ALLOCATOR_HEAP_BUFFER_FIRST, false);
         return conf;
     }
 }

### CODE REVIEW

File: fluss-common/src/main/java/org/apache/fluss/config/ConfigOptions.java [hunk 1]
```diff
@@ -931,6 +931,15 @@ public class ConfigOptions {
                             "The number of threads that the client uses for sending requests to the "
                                     + "network and receiving responses from network. The default value is 4");
 
+    public static final ConfigOption<Boolean> NETTY_CLIENT_ALLOCATOR_HEAP_BUFFER_FIRST =
```
wuchong, line 944: This config is not used. I found this configuraiton is still useful, I think we should support it.

fluss-rpc/src/main/java/org/apache/fluss/rpc/netty/client/NettyClient.java
```diff
@@ -89,19 +90,19 @@ public NettyClient(Configuration conf, ClientMetricGroup clientMetricGroup) {
...
                 new Bootstrap()
                         .group(eventGroup)
                         .channel(NettyUtils.getClientSocketChannelClass(eventGroup))
-                        .option(ChannelOption.ALLOCATOR, pooledAllocator)
+                        .option(ChannelOption.ALLOCATOR, allocator)
                         .option(ChannelOption.CONNECT_TIMEOUT_MILLIS, connectTimeoutMs)
                         .option(ChannelOption.TCP_NODELAY, true)
                         .option(ChannelOption.SO_KEEPALIVE, true)
-                        .handler(new ClientChannelInitializer(connectionMaxIdle));
+                        .handler(new ClientChannelInitializer(connectionMaxIdle, isInnerClient));
```
wuchong: I think we should get the prefer heap option from `Configuration`, rather than hard code using `isInnerClient` here.

fluss-rpc/src/main/java/org/apache/fluss/rpc/netty/NettyChannelInitializer.java
```diff
@@ -46,15 +49,119 @@ protected void initChannel(SocketChannel ch) throws Exception {
...
+     * A custom {@link ByteToMessageDecoder.Cumulator} that keeps the cumulation buffer on the JVM
+     * heap to avoid off-heap memory pressure from large partial messages.
+     *
+     * <p>On native transports (epoll / kqueue), {@code EpollRecvByteAllocatorHandle.allocate()}
+     * always forces the read buffer to be direct regardless of the channel allocator. The default
+     * {@code MERGE_CUMULATOR}'s fast-path would adopt this direct buffer as the cumulation. For
+     * large partial messages, the cumulation can grow up to the maximum frame size (e.g. 32 MB), so
+     * multiple connections could consume significant off-heap memory.
+     */
+    static final class HeapPreferringCumulator implements ByteToMessageDecoder.Cumulator {
```
wuchong, line 95: Could you add a comment explaining the origin of `HeapPreferringCumulator`, specifically detailing the source class it was derived from and the modifications made? This will help us track changes more effectively.

Besides, please add unit tests for this class. You can let AI to learn from Netty source code to generate UT for this class.

fluss-rpc/src/main/java/org/apache/fluss/rpc/netty/NettyChannelInitializer.java
```diff
@@ -46,15 +49,119 @@ protected void initChannel(SocketChannel ch) throws Exception {
...
         ch.pipeline().addLast("idle", new IdleStateHandler(0, 0, maxIdleTimeSeconds));
     }
+
+    /**
+     * A custom {@link ByteToMessageDecoder.Cumulator} that keeps the cumulation buffer on the JVM
+     * heap to avoid off-heap memory pressure from large partial messages.
+     *
+     * <p>On native transports (epoll / kqueue), {@code EpollRecvByteAllocatorHandle.allocate()}
+     * always forces the read buffer to be direct regardless of the channel allocator. The default
+     * {@code MERGE_CUMULATOR}'s fast-path would adopt this direct buffer as the cumulation. For
```
wuchong: `{@code MERGE_CUMULATOR}` -> `{@link ByteToMessageDecoder#MERGE_CUMULATOR}`.

loserwang1024:
> Thanks @loserwang1024 , It seems you reverted the #2948 for the `RpcClient` changes. However, it seems we don't need to revert this change, because the `preferHeap` should be derived from configuration, rather than hard code to true in client side.

inner client is not needed to preferHeap because off-heap memory is enough now.

fluss-rpc/src/main/java/org/apache/fluss/rpc/netty/NettyChannelInitializer.java
```diff
@@ -46,15 +49,138 @@ protected void initChannel(SocketChannel ch) throws Exception {
...
+                // always copied into a heap cumulation so downstream decoding continues to operate
+                // on heap memory.
+                if (!in.isDirect()) {
+                    cumulation.release();
+                    return in;
+                }
+                // Direct buffer — allocate a heap cumulation pre-sized to the expected frame length
+                // to avoid repeated expansion.
+                cumulation.release();
+                ByteBuf heapCumulation = allocateForFrame(alloc, in);
```
fresh-borzoni: If allocateForFrame throws for whatever reason(OOM f.e), then `in` is never released as finally isn't reached, also if `writeBytes` throws - the symmetric problem with heapCummulation. 
Probably we can use the same pattern as in `expandCummulation`

fluss-client/src/test/java/org/apache/fluss/client/table/scanner/log/DefaultCompletedFetchBufferLifecycleTest.java
```diff
@@ -1,165 +0,0 @@
...
-
-import static org.apache.fluss.record.TestData.DATA1;
-import static org.apache.fluss.record.TestData.DATA1_ROW_TYPE;
-import static org.apache.fluss.record.TestData.DEFAULT_SCHEMA_ID;
-import static org.apache.fluss.record.TestData.TEST_SCHEMA_GETTER;
-import static org.apache.fluss.testutils.DataTestUtils.genMemoryLogRecordsByObject;
-import static org.assertj.core.api.Assertions.assertThat;
-
-/** Tests for {@link DefaultCompletedFetch} zero-copy buffer lifecycle (issue #1184). */
-class DefaultCompletedFetchBufferLifecycleTest {
```
fresh-borzoni, line 42: It looks like this deletion was carried over from rebase/revert innerClient removal, but I think - the test is unrelated to the cumulator change (it tests retain/release in DefaultCompletedFetch.drain(), which is unchanged). Should we keep it?

fluss-server/src/main/java/org/apache/fluss/server/ServerBase.java
```diff
@@ -92,6 +93,17 @@ protected static Configuration loadConfiguration(String[] args, String serverCla
         }
     }
 
+    /**
+     * Applies server-side default values for configuration options that are not explicitly set by
```
fresh-borzoni, line 97: Minor design suggestion: instead of overriding the config default in code, would it be cleaner to have two separate config options following the existing netty.client.* / netty.server.* pattern?

fluss-client/src/test/java/org/apache/fluss/client/table/scanner/log/DefaultCompletedFetchBufferLifecycleTest.java
loserwang1024: my mistake

fluss-rpc/src/main/java/org/apache/fluss/rpc/netty/NettyChannelInitializer.java
loserwang1024: 👌

loserwang1024:
> @loserwang1024 Ty for the nice PR, left minor comments, PTAL

modified it

fluss-server/src/main/java/org/apache/fluss/server/ServerBase.java
loserwang1024: it‘s server side's ClientChannelInitializer(rather than ServerChannelInitializer) .  Thus I still use the netty.client.xxx.

PR #2808, opened 2026-03-06
Fixing the Flaky Test for org.apache.fluss.client.table.scanner.batch.KvSnapshotBatchScannerITCase.testKvSnapshotLease

### LINKED ISSUE #2807: Flaky Test: KvSnapshotBatchScannerITCase.testKvSnapshotLease
### Search before asking

### Fluss version

0.9.0 (latest release)

### Please describe the bug 🐞

Error:  Tests run: 3, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 19.961 s <<< FAILURE! - in org.apache.fluss.client.table.scanner.batch.KvSnapshotBatchScannerITCase
Error:  org.apache.fluss.client.table.scanner.batch.KvSnapshotBatchScannerITCase.testKvSnapshotLease  Time elapsed: 3.16 s  <<< FAILURE!
org.opentest4j.AssertionFailedError: 

expected: [1L, 1L, 1L]
 but was: [1L, 0L, 1L]

The tests that were run was part of 
https://github.com/apache/fluss/actions/runs/22756751008/job/66003123256

###  Root Cause Analysis                                                                                                                                                                                    
                                                                                                                                                                                                         
 Primary Race: Async triggerSnapshot in FlussClusterExtension                                                                                                                                            
                                                                                                                                                                                                         
 In FlussClusterExtension.triggerSnapshot() (line 747-775):                                                                                                                                              
                                                                                                                                                                                                         
 snapshotId = kvSnapshotManager.currentSnapshotId();     // read counter (e.g., 1)                                                                                                                       
 kvSnapshotManager.triggerSnapshot();                     // ASYNC - queues to guardedExecutor                                                                                                           
 nextSnapshotId = kvSnapshotManager.currentSnapshotId(); // read counter again                                                                                                                           
                                                                                                                                                                                                         
 PeriodicSnapshotManager.triggerSnapshot() (line 186) submits work to guardedExecutor.execute(...) which is asynchronous. The actual initSnapshot() call — which increments snapshotIdCounter via        
 getAndIncrement() at KvTabletSnapshotTarget.java:183 — runs on the guardedExecutor thread, NOT the calling thread.                                                                                      
                                                                                                                                                                                                         
 The race:                                                                                                                                                                                               
 1. Test thread reads snapshotId = 1 (counter value after first snapshot)                                                                                                                                
 2. Test thread calls triggerSnapshot() → queues async task                                                                                                                                              
 3. Test thread reads nextSnapshotId = 1 (counter hasn't been incremented yet)                                                                                                                           
 4. nextSnapshotId == snapshotId → method returns null                                                                                                                                                   
 5. triggerAndWaitSnapshots skips this bucket entirely                                                                                                                                                   
                                                                                                                                                                                                         
 The async task eventually runs and creates snapshot 1, but the test has already moved on. When getLatestKvSnapshots() is called, this bucket may still show snapshot 0.                                 
                                                                                                                                                                                                         
 Secondary Race: Log offset not yet flushed to KV                                                                                                                                                        
                                                                                                                                                                                                         
 In KvTabletSnapshotTarget.initSnapshot() (line 170-203), there's a check:                                                                                                                               
 if (logOffset <= logOffsetOfLatestSnapshot) {                                                                                                                                                           
     return Optional.empty();  // skip — no new data                                                                                                                                                     
 }                                                                                                                                                                                                       
                                                                                                                                                                                                         
 Even if the async trigger runs, if the KV tablet hasn't applied the new log entries yet (from putRows), the snapshot is skipped because the flushed log offset hasn't advanced.                         
                                                                                                                                                                                                         
 Flow of the failure                                                                                                                                                                                     
                                                                                                                                                                                                         
 1. triggerAndWaitSnapshots → for bucket 1, triggerSnapshot returns null (race) → bucket 1 is skipped                                                                                                    
 2. Test calls admin.getLatestKvSnapshots() → bucket 1 still at snapshot 0                                                                                                                               
 3. acquireSnapshots stores {bucket0: 1, bucket1: 0, bucket2: 1}                                                                                                                                         
 4. checkKvSnapshotLeaseEquals expects [1L, 1L, 1L] but gets [1L, 0L, 1L]                                                                                                                                
                                                                                                                                                                                                         
 ### Fix                                                                                                                                                                                                     
                                                                                                                                                                                                         
 Modify FlussClusterExtension.triggerSnapshot() to properly handle the async nature. Instead of checking currentSnapshotId() immediately after the async trigger, use a retry/poll approach to wait for  
 the snapshot ID to increment.                                                                                                                                                                           
                                                                                                                                                                                                         
 File to modify                                                                                                                                                                                                                                                                                                                                                                                                   
- fluss-server/src/test/java/org/apache/fluss/server/testutils/FlussClusterExtension.java (lines 747-775)                                                                                               
                                                                                                                                                                                                         

hemanthsavasere:
### Purpose
- Fix a race condition in FlussClusterExtension.triggerSnapshot() that caused testKvSnapshotLease to intermittently fail with expected: [1L, 1L, 1L] but was: [1L, 0L, 1L].

Linked issue: close #2807 

### Brief change log

fluss-server/src/test/java/org/apache/fluss/server/testutils/FlussClusterExtension.java
- Before: Read currentSnapshotId() synchronously right after triggerSnapshot(); if unchanged, returned null and skipped the bucket
- After: Poll currentSnapshotId() in a loop (10ms sleep, 30s timeout) until the ID increments, confirming the async trigger was actually processed; fail with a clear message on timeout
 
### Tests

 - Test: fluss-client/src/test/java/org/apache/fluss/client/table/scanner/batch/KvSnapshotBatchScannerITCase.java
 - Fix target: fluss-server/src/test/java/org/apache/fluss/server/testutils/FlussClusterExtension.java:747-775 
 
### API and Format

- No API changes. Changes only related to tests.

### Documentation

- No new features are introduced

### DIFF
--- a/fluss-server/src/test/java/org/apache/fluss/server/testutils/FlussClusterExtension.java
+++ b/fluss-server/src/test/java/org/apache/fluss/server/testutils/FlussClusterExtension.java
@@ -756,12 +756,12 @@ public final class FlussClusterExtension
                     // Poll until the snapshot ID increments, confirming the async trigger was
                     // processed. triggerSnapshot() submits work to a guardedExecutor
                     // asynchronously, so the counter may not have incremented yet on return.
-                    long deadline = System.currentTimeMillis() + 30_000;
+                    // If the ID does not increment within the timeout, the snapshot was
+                    // legitimately skipped (e.g., no new data since last snapshot).
+                    long deadline = System.currentTimeMillis() + 1_000;
                     while (kvSnapshotManager.currentSnapshotId() <= snapshotId) {
                         if (System.currentTimeMillis() > deadline) {
-                            fail(
-                                    "Timed out waiting for snapshot trigger to be processed for "
-                                            + tableBucket);
+                            return null;
                         }
                         try {
                             Thread.sleep(10);


fluss-server/src/test/java/org/apache/fluss/server/testutils/FlussClusterExtension.java
```diff
@@ -745,33 +745,37 @@ public CompletedSnapshot triggerAndWaitSnapshot(TableBucket tableBucket) {
...
+                            return null;
+                        }
+                        try {
+                            Thread.sleep(10);
+                        } catch (InterruptedException e) {
+                            Thread.currentThread().interrupt();
+                            throw new RuntimeException(e);
+                        }
+                    }
+                    return snapshotId;
```
swuferhong, line 773: can we use `retry()` instead of while loop to assert the snapshot id incremental?

hemanthsavasere: Hi @swuferhong,
As per my analysis 

- Both retry() and waitUntil() throw/fail on timeout. The current while loop returns null on timeout, a fundamentally different semantic. The null return is intentional: the caller triggerAndWaitSnapshot() (line 737) checks for null and calls fail() only if no snapshot was triggered, distinguishing "not triggered" from "legitimately skipped.

- Also, wrapping retry or waitUntil in try-catch to swallow its failure does not serve the its purpose and is arguably less clear than the current while loop.

 A new utility like pollUntil() returning Optional<T> could be added

leekeiabstraction: Hello @hemanthsavasere 

Reviewing this as there are recent failures pertaining this. The root cause looks correct. And I agree with the semantic difference between returning null and throwing exceptions. 

However if we consider the callers of `triggerSnapshot(TableBucket tableBucket)`.

1. `CompletedSnapshot triggerAndWaitSnapshot(TableBucket tableBucket)` which you pointed out. On a null snapshot ID, it fails saying with `No new snapshot triggered for table bucket <>`. `triggerSnapshot(TableBucket tableBucket)` can be be made to use `retry()` with assertion on snapshotID increment + `withFailMessage()` to match the failure message.
2. `void triggerAndWaitSnapshots(Collection<TableBucket> tableBuckets)` currently triggers snapshots and actually drop the `waitUntilSnapshotFinished` for null snapshot ids. I think this means that tests that call it actually may miss actual cases where snapshots were not successfully triggered. Adding assertion through `retry()` will harden our tests.

Let me know what your thoughts are.

leekeiabstraction (changes requested):

Issue #1600, opened 2025-08-27
[Umbrella] FIP-17: Support Full KV Scan for Primary Key Tables


polyzos:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/fluss/issues) and found nothing similar.

### Motivation

Fluss KV currently supports point lookups, but RocksDB allows iterators, range, and prefix scans. 

For small tables (hundreds of rows or a few thousands), that can be used to create dashboards often need to periodically read all keys. 

**Goal:** Assume I have a front-end and I want to create a dashboard to show < 1-2k values.. In this scenario, I could use the fluss client directly to periodically query the whole kv store

### Solution

We can expose functionality to the api to allow a full scan (with snapshot consistency) to simplify such use cases. 

-------------------
An implementation can be found [here](https://github.com/apache/fluss/pull/2809) and for easier reviews it can be broken down to the following individuals PRs:

xx789633:
Hi @polyzos I can take this task. Please assign it to me.

polyzos:
@xx789633 Thank you for your interest. Assigned 👍

xx789633:
Hi @polyzos , I think we can reuse the existing KV snapshots for batch reading (https://github.com/apache/fluss/pull/88) instead of invoking the Rocksdb scan iterator. This reduces the overhead on Rocksdb. What do you think?

xx789633:
Currently, a workaround for full scan can be implemented like this:
```java
            Scan scan = table.newScan().limit(limit).project(projectedFields);
            List<BatchScanner> scanners =
                    tableBuckets.stream()
                            .map(scan::createBatchScanner)
                            .collect(Collectors.toList());
            List<InternalRow> scannedRows = BatchScanUtils.collectLimitedRows(scanners, limit);
```

polyzos:
@xx789633 I have some time and wanted to give this a spin.. If there is no PR, do you mind if I also give this a try?

xx789633:
@polyzos I'm really sorry for the slow progress. I'm currently tied up with some other community tasks.  ​​Please feel free to move forward.

polyzos:
@xx789633 Nothing to do with slow progress and nothing urgent 😁 Based on our offline comms, I realise how busy you are, and because I have some extra time this week, I will be eager to give this a try. 
I will keep you in the loop about any progress 🙏

polyzos:
#1702

polyzos:
[...]

PR #2809, opened 2026-03-07
[KV] Support Full KV Scan 

polyzos:


platinumhamburg:
@polyzos Thanks, Gannis. This is a very powerful and inspiring PR. Bringing it in will significantly strengthen Fluss’s out-of-the-box experience and lay an outstanding foundation for future architectural evolution. However, I have a few small questions. At the moment, it seems there are still three issues:

The close_scanner request can’t actually be sent yet, which means when a scan is terminated early, the server-side scanner can only wait for TTL-based cleanup.
The scanner lifecycle isn’t integrated into the leader migration / replica deletion flow, which may block replica switching and reclamation.
The server has already encoded scan errors into ScanKvResponse, but the client doesn’t handle these errors at all, and may silently treat real failures as “scan completed”.
I also noticed there were already corresponding Copilot review comments on these points, but they were silently closed, so I didn’t add any further review notes. I’d like to hear your thoughts on these issues.

polyzos:
@platinumhamburg thank you for your thoughtful comments, and these are great catches.
I will keep iterating on these.

The points 1 and 3 you mentioned should be fairly easy to address, by adding a future handler for 1 and for 3 dding an error check for `pollBatch()` .

For point 2 `ScannerManager` probably needs to be passed to `ReplicaManager`.

I will make these changes and let me know if you have any other thoughts to improve this

fluss-rpc/src/main/java/org/apache/fluss/rpc/protocol/Errors.java
```diff
@@ -242,10 +245,15 @@ public enum Errors {
     REBALANCE_FAILURE_EXCEPTION(61, "The rebalance task failure.", RebalanceFailureException::new),
     NO_REBALANCE_IN_PROGRESS_EXCEPTION(
             62, "No rebalance task in progress.", NoRebalanceInProgressException::new),
-    INVALID_PRODUCER_ID_EXCEPTION(
-            63,
-            "The client has attempted to perform an operation with an invalid producer ID.",
-            InvalidProducerIdException::new);
+    SCANNER_EXPIRED(
```
platinumhamburg, line 260: The diff removes `INVALID_PRODUCER_ID_EXCEPTION(63, ...)` entirely and replaces it with new scanner error codes starting at 64. Two problems stack up here:

First, `InvalidProducerIdException` is still actively used in `ProducerOffsetsManager` (11+ references). Without the error code mapping, all those exceptions will degrade to `UNKNOWN_SERVER_ERROR` on the wire — clients won't be able to tell what went wrong.

Second, and more critically — current main already has `CONFIG_EXCEPTION(64, ...)`. This PR adds `SCANNER_EXPIRED(64, ...)`. The `Errors` static initializer detects duplicate codes and throws, so TabletServer won't even start.

Looks like the PR was branched before `CONFIG_EXCEPTION` landed on main. A rebase should fix it — restore code 63, and start the new codes from 65:

fluss-client/src/main/java/org/apache/fluss/client/table/scanner/TableScan.java
```diff
@@ -138,6 +145,32 @@ public BatchScanner createBatchScanner(TableBucket tableBucket) {
                 limit);
     }
 
+    @Override
+    public BatchScanner createBatchScanner() {
```
platinumhamburg, line 209: The Javadoc says "only supported for Primary Key Tables", but there's no actual check. If someone calls this on a Log Table, it'll travel all the way to the server's `getLeaderKvTablet()` before failing with `InvalidTableException` — a wasted RPC and a confusing error message.

A quick guard at the top would save everyone some headaches:
```java
if (!tableInfo.hasPrimaryKey()) {
    throw new UnsupportedOperationException(
        "createBatchScanner() without TableBucket is only supported for Primary Key Tables. "
        + "Table: " + tableInfo.getTablePath());
}
```

fluss-client/src/main/java/org/apache/fluss/client/table/scanner/TableScan.java
```diff
@@ -138,6 +145,32 @@ public BatchScanner createBatchScanner(TableBucket tableBucket) {
...
+        int numBuckets = tableInfo.getNumBuckets();
+        List<TableBucket> buckets = new ArrayList<>();
+        if (!tableInfo.isPartitioned()) {
+            for (int b = 0; b < numBuckets; b++) {
+                buckets.add(new TableBucket(tableId, b));
+            }
+        } else {
+            try {
+                List<PartitionInfo> partitions =
+                        conn.getAdmin().listPartitionInfos(tableInfo.getTablePath()).get();
```
platinumhamburg:
```java
List<PartitionInfo> partitions =
    conn.getAdmin().listPartitionInfos(tableInfo.getTablePath()).get();
```

Bare `.get()` with no timeout. If the coordinator is down, this hangs indefinitely. Adding a timeout is a small change that avoids a nasty surprise:
```java
.get(30, TimeUnit.

File: fluss-common/src/main/java/org/apache/fluss/config/ConfigOptions.java [hunk 4]
```diff
@@ -1128,6 +1162,14 @@ public class ConfigOptions {
                                     + "will still be returned to ensure that the fetch can make progress. As such, "
                                     + "this is not a absolute maximum.");
 
+    public static final ConfigOption<MemorySize> CLIENT_SCANNER_KV_FETCH_MAX_BYTES =
```
platinumhamburg, line 1216: Config is defined but never read

fluss-server/src/main/java/org/apache/fluss/server/kv/scan/ScannerContext.java
```diff
@@ -0,0 +1,168 @@
...
+@NotThreadSafe
+public class ScannerContext implements Closeable {
+    private final String scannerId;
+    private final byte[] scannerIdBytes;
+    private final TableBucket tableBucket;
+    private final RocksDBKv rocksDBKv;
+    private final RocksIterator iterator;
+    private final ReadOptions readOptions;
+    private final Snapshot snapshot;
+    private final ResourceGuard.Lease resourceLease;
```
platinumhamburg, line 58: `ScannerContext` holds a `ResourceGuard.Lease`, and `KvTablet.close()` ultimately calls `resourceGuard.close()`, which blocks until all leases are released. If a scanner has a long TTL and the client doesn't close it, `KvTablet.close()` will just hang.

The good news is `makeFollowers` and `stopReplica` both call `closeScannersForBucket` already, so the main paths are covered. But if KvTablet gets closed through some other path (e.g., exception recovery), there's still a risk of getting stuck. Worth adding a cleanup step in `Replica.close()` or `KvTablet.close()` as a safety net.

fluss-server/src/main/java/org/apache/fluss/server/kv/scan/ScannerManager.java
```diff
@@ -0,0 +1,410 @@
...
+     * Removes and closes a known scanner context directly, avoiding a map lookup.
+     *
+     * <p>Uses a conditional remove ({@link java.util.concurrent.ConcurrentHashMap#remove(Object,
+     * Object)}) so that concurrent calls — e.g. from the TTL reaper and a close-scanner RPC
+     * arriving simultaneously — result in exactly one winner closing the context, preventing
+     * double-release of the non-idempotent {@link
+     * org.apache.fluss.server.utils.ResourceGuard.Lease}.
+     */
+    public void removeScanner(ScannerContext context) {
+        if (scanners.remove(context.getId(), context)) {
```
platinumhamburg, line 213: `removeScanner(byte[])` uses `scanners.remove(key)` (unconditional), while `removeScanner(ScannerContext)` uses `scanners.remove(key, value)` (conditional). If the TTL reaper fires at the same time, `decrementCounts` can get called twice for the same scanner, pushing the counter negative. Using conditional removal consistently would close this gap.

fluss-server/src/main/java/org/apache/fluss/server/kv/scan/ScannerContext.java
```diff
@@ -0,0 +1,168 @@
...
+    private long remainingLimit;
+    private int callSeqId = -1;
```
platinumhamburg, line 64: The initial value `-1` works correctly with the client's first continuation sending `0` (`-1 + 1 = 0`), but it's a subtle implicit contract. A one-line comment would save the next person from having to trace through both sides to convince themselves it's right.

platinumhamburg (changes requested):

polyzos:
@platinumhamburg thank you for your detailed review and all the comments.
Currently I will start breaking down this PR as described here
https://github.com/apache/fluss/issues/1600
into smaller ones so its easier to review and merge, and I will handle all your comments in the implementation there.

I will ping you to take a look

PR #2829, opened 2026-03-10
[kv] Support elect standby replicas for primary key table

### LINKED ISSUE #2828: Support elect standby replicas for primary key table
### Search before asking

### Description

Support elect standby replicas for primary key table

### Willingness to contribute

swuferhong:
### Purpose

Linked issue: close #2828

### DIFF
--- a/fluss-server/src/main/java/org/apache/fluss/server/coordinator/statemachine/ReplicaLeaderElection.java
+++ b/fluss-server/src/main/java/org/apache/fluss/server/coordinator/statemachine/ReplicaLeaderElection.java
@@ -21,6 +21,7 @@ import org.apache.fluss.server.coordinator.statemachine.TableBucketStateMachine.
 import org.apache.fluss.server.zk.data.LeaderAndIsr;
 
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Optional;
@@ -41,23 +42,25 @@ public abstract class ReplicaLeaderElection {
          * @param assignments the assignments
          * @param aliveReplicas the alive replicas
          * @param leaderAndIsr the original leaderAndIsr
+         * @param isPrimaryKeyTable whether this table bucket is primary key table
          * @return the election result
          */
         public Optional<ElectionResult> leaderElection(
-                List<Integer> assignments, List<Integer> aliveReplicas, LeaderAndIsr leaderAndIsr) {
-            // currently, we always use the first replica in assignment, which also in aliveReplicas
-            // and
-            // isr as the leader replica.
+                List<Integer> assignments,
+                List<Integer> aliveReplicas,
+                LeaderAndIsr leaderAndIsr,
+                boolean isPrimaryKeyTable) {
             List<Integer> isr = leaderAndIsr.isr();
-            for (int assignment : assignments) {
-                if (aliveReplicas.contains(assignment) && isr.contains(assignment)) {
-                    return Optional.of(
-                            new TableBucketStateMachine.ElectionResult(
-                                    aliveReplicas, leaderAndIsr.newLeaderAndIsr(assignment, isr)));
-                }
-            }
-
-            return Optional.empty();
+            // First we will filter out the assignment list to only contain the alive replicas and
+            // isr.
+            List<Integer> availableReplicas =
+                    assignments.stream()
+                            .filter(
+                                    replica ->
+                                            aliveReplicas.contains(replica)
+                                                    && isr.contains(replica))
+                            .collect(Collectors.toList());
+            return electLeader(availableReplicas, aliveReplicas, leaderAndIsr, isPrimaryKeyTable);
         }
     }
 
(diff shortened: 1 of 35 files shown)


fluss-server/src/main/java/org/apache/fluss/server/coordinator/statemachine/ReplicaLeaderElection.java
```diff
@@ -70,32 +73,66 @@ public static class ControlledShutdownLeaderElection extends ReplicaLeaderElecti
...
+         * @param isPrimaryKeyTable whether this table bucket is primary key table
          * @return the election result
          */
         public Optional<ElectionResult> leaderElection(
                 List<Integer> assignments,
                 List<Integer> aliveReplicas,
                 LeaderAndIsr leaderAndIsr,
-                Set<Integer> shutdownTabletServers) {
+                Set<Integer> shutdownTabletServers,
+                boolean isPrimaryKeyTable) {
```
wuchong, line 85: Can we reuse the `electLeader` static method in this method? It seems most logics are similar.

fluss-server/src/main/java/org/apache/fluss/server/replica/IsrState.java
```diff
@@ -40,22 +40,31 @@ public interface IsrState {
      */
     List<Integer> maximalIsr();
 
+    List<Integer> standbyReplicas();
```
wuchong, line 49: Add javadocs for this method, explain what is standby replicas.

fluss-server/src/main/java/org/apache/fluss/server/zk/data/LeaderAndIsrJsonSerde.java
```diff
@@ -71,6 +80,16 @@ public LeaderAndIsr deserialize(JsonNode node) {
         while (isrNodes.hasNext()) {
             isr.add(isrNodes.next().asInt());
         }
-        return new LeaderAndIsr(leader, leaderEpoch, isr, coordinatorEpoch, bucketEpoch);
+
+        List<Integer> standbyList = new ArrayList<>();
+        if (version > 1) {
```
wuchong: update `VERSION` to `CURRENT_VERSION` and update this to `if (version >= CURRENT_VERSION)`

fluss-server/src/main/java/org/apache/fluss/server/replica/Replica.java
```diff
@@ -576,11 +589,28 @@ private void registerLakeTieringMetrics() {
...
+                isStandbyReplica = true;
+            } else {
+                if (wasStandby || wasLeader) {
+                    // standby -> follower or leader -> follower
+                    if (wasStandby) {
+                        isStandbyReplica = false;
+                    }
+                }
+                // follower -> follower: do nothing.
+            }
```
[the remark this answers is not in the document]
wuchong: big +1, the `wasStandby` and `wasLeader` flag are very confusing here and makes the code hard to understand.

fluss-server/src/main/java/org/apache/fluss/server/replica/Replica.java
```diff
@@ -200,10 +200,12 @@ public final class Replica {
     private final Map<Integer, FollowerReplica> followerReplicasMap =
             MapUtils.newConcurrentHashMap();
 
-    private volatile IsrState isrState = new IsrState.CommittedIsrState(Collections.emptyList());
+    private volatile IsrState isrState =
+            new IsrState.CommittedIsrState(Collections.emptyList(), Collections.emptyList());
     private volatile int leaderEpoch = LeaderAndIsr.INITIAL_LEADER_EPOCH - 1;
     private volatile int bucketEpoch = LeaderAndIsr.INITIAL_BUCKET_EPOCH;
     private volatile int coordinatorEpoch = CoordinatorContext.INITIAL_COORDINATOR_EPOCH;
+    private volatile boolean isStandbyReplica = false;
```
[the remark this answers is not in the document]
wuchong: +1

fluss-server/src/main/java/org/apache/fluss/server/utils/ServerRpcMessageUtils.java
```diff
@@ -1336,12 +1344,18 @@ public static Map<TableBucket, LeaderAndIsr> getAdjustIsrData(AdjustIsrRequest r
                 for (int i = 0; i < reqForBucket.getNewIsrsCount(); i++) {
                     newIsr.add(reqForBucket.getNewIsrAt(i));
                 }
+                List<Integer> standbyReplica = new ArrayList<>();
+                for (int i = 0; i < reqForBucket.getStandbyReplicasCount(); i++) {
+                    standbyReplica.add(reqForBucket.getStandbyReplicaAt(i));
+                }
```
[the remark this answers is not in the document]
wuchong: +1

fluss-server/src/main/java/org/apache/fluss/server/utils/ServerRpcMessageUtils.java
```diff
@@ -1417,6 +1432,10 @@ public static Map<TableBucket, AdjustIsrResultForBucket> getAdjustIsrResponseDat
                 for (int i = 0; i < respForBucket.getIsrsCount(); i++) {
                     isr.add(respForBucket.getIsrAt(i));
                 }
+                List<Integer> standbyReplica = new ArrayList<>();
+                for (int i = 0; i < respForBucket.getStandbyReplicasCount(); i++) {
+                    standbyReplica.add(respForBucket.getStandbyReplicaAt(i));
+                }
```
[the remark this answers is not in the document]
wuchong: +1

fluss-server/src/main/java/org/apache/fluss/server/coordinator/statemachine/TableBucketStateMachine.java
```diff
@@ -677,23 +685,40 @@ public LeaderAndIsr getLeaderAndIsr() {
...
+                standbyReplicas.add(availableReplicas.get(1));
             }
         }
 
-        return Optional.empty();
+        return Optional.of(
+                new ElectionResult(
+                        aliveReplicas,
+                        new LeaderAndIsr(
+                                leader, 0, aliveReplicas, standbyReplicas, coordinatorEpoch, 0)));
```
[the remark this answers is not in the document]
wuchong: +1, looks like we also have problems previously, but happens in rarely cases.

fluss-server/src/main/java/org/apache/fluss/server/zk/data/LeaderAndIsrJsonSerde.java
```diff
@@ -71,6 +80,16 @@ public LeaderAndIsr deserialize(JsonNode node) {
...
+            Iterator<JsonNode> hotStandbyListNodes = node.get(STANDBY_REPLICA).elements();
+            while (hotStandbyListNodes.hasNext()) {
+                standbyList.add(hotStandbyListNodes.next().asInt());
```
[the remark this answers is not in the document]
wuchong: I think this is fine.

fluss-server/src/main/java/org/apache/fluss/server/replica/Replica.java
```diff
@@ -575,11 +589,28 @@ private void registerLakeTieringMetrics() {
...
+                    // standby -> follower or leader -> follower
+                    if (wasStandby) {
+                        isStandbyReplica = false;
+                    }
+                }
+                // follower -> follower: do nothing.
+            }
+
             // it should be from leader to follower, we need to destroy the kv tablet
             dropKv();
```
wuchong, line 613: `dropKv()` is called unconditionally in `onBecomeNewFollower()`, even when the replica is designated as a standby. IIUC, a standby replica should retain the local kv rather than dropped. Is this to be implemented in a follow-up PR, or missed in this PR?

fluss-server/src/main/java/org/apache/fluss/server/coordinator/statemachine/ReplicaLeaderElection.java
```diff
@@ -108,20 +145,109 @@ public ReassignmentLeaderElection(List<Integer> newReplicas) {
...
-                                    liveReplicas, leaderAndIsr.newLeaderAndIsr(assignment, isr)));
-                }
-            }
+            List<Integer> availableReplicas =
+                    newReplicas.stream()
+                            .filter(
+                                    replica ->
+                                            liveReplicas.contains(replica) && isr.contains(replica))
+                            .collect(Collectors.toList());
+            return electLeader(availableReplicas, liveReplicas, leaderAndIsr, isPrimaryKeyTable);
```
wuchong: [P1] When a primary-key bucket already has a live standby, this reassignment path now routes through the generic PK election logic and may promote the current standby instead of honoring `newReplicas[0]` as the rebalance target leader. `GoalOptimizerUtils` encodes the desired leader as the first replica, and `tryToCompleteRebalanceTask()` only completes once the actual leader matches that target, so leader-only rebalances can elect the wrong leader and remain stuck in `REBALANCING` indefinitely. Essentially, the rebalance plan did not account for standby replicas, resulting in a mismatch between the plan and the actual leaders.

fluss-server/src/main/java/org/apache/fluss/server/replica/Replica.java
```diff
@@ -1722,7 +1755,12 @@ private IsrState.PendingExpandIsrState prepareIsrExpand(
 
         LeaderAndIsr newLeaderAndIsr =
                 new LeaderAndIsr(
-                        localTabletServerId, leaderEpoch, isrToSend, coordinatorEpoch, bucketEpoch);
+                        localTabletServerId,
+                        leaderEpoch,
+                        isrToSend,
+                        currentState.standbyReplicas(),
```
wuchong, line 1768: [P2] If a PK bucket currently has no standby and another follower later catches up, `prepareIsrExpand()` sends the enlarged ISR but reuses `currentState.standbyReplicas()` unchanged here. That persists the expanded ISR with an empty standby list, so these buckets never gain a standby on the normal replica-recovery path until some later leader election. If the standby replicas is updated, we should also send a notification to the replica to make it ready for standby.

wuchong (commented):
@swuferhong I left some comments.

fluss-server/src/main/java/org/apache/fluss/server/replica/Replica.java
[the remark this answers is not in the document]
swuferhong: I'd prefer to keep `isNowStandby`, `wasLeader`, and `wasStandby` here. They will be useful for follow-up PRs, because the behavior differs depending on the specific transition — for example, `leader→standby`, `leader→follower`, and `follower→standby` each require different actions: some need to delete data, others need to download data. So we need to distinguish between these cases.

fluss-server/src/main/java/org/apache/fluss/server/replica/Replica.java
swuferhong: This logic will be implements in next pr: https://github.com/apache/fluss/pull/2835

fluss-server/src/main/java/org/apache/fluss/server/replica/Replica.java
swuferhong: This is a bit complex and can be implemented in a separate PR. I'll add a TODO for now.

swuferhong:
@wuchong comments addressed.

wuchong (approved):
LGTM.

wuchong:
@swuferhong I have merged this PR. However, to ensure backward compatibility, we should introduce a table configuration option to explicitly enable standby replicas for legacy tables. For newly created tables, standby replicas can be enabled by default.

swuferhong:
> @swuferhong I have merged this PR. However, to ensure backward compatibility, we should introduce a table configuration option to explicitly enable standby replicas for legacy tables. For newly created tables, standby replicas can be enabled by default.

The linked issue: https://github.com/apache/fluss/issues/3253

PR #2835, opened 2026-03-10
[kv] Support download latest kv snapshot from remote while becoming standby replica

### LINKED ISSUE #2830: Support download latest kv snapshot from remote while becoming standby replica
### Search before asking

### Description

Support download latest kv snapshot from remote while becoming standby replica

### Willingness to contribute

swuferhong:
### Purpose

Linked issue: close #2830


fluss-server/src/main/java/org/apache/fluss/server/kv/snapshot/KvSnapshotManager.java
```diff
@@ -0,0 +1,1080 @@
...
+            LOG.debug(
+                    "No files to move from temp directory {} to final directory {} for bucket {}",
+                    tempDir,
+                    finalDir,
+                    tableBucket);
+            return;
+        }
+
+        int movedCount = 0;
+        for (File file : files) {
```
platinumhamburg, line 942: There are two `checkDownloadNotCancelled(expectedEpoch)` calls before this loop (before deleting obsolete files and before moving), but none **inside** the loop itself. If `awaitStandbyDownload()` times out and the caller bumps `downloadEpoch`, the download thread will not notice until after the entire move loop completes.

In practice this race window is extremely narrow (the move loop is just filesystem metadata ops, milliseconds), but since `checkDownloadNotCancelled` is already used elsewhere as a cooperative cancellation mechanism, adding one inside the loop is zero-cost and closes the theoretical gap:

```java
for (File file : files) {
    checkDownloadNotCancelled(expectedEpoch);
    Path sourcePath = file.toPath();
    Path targetPath = finalDir.resolve(file.getName());
    ...
```

fluss-server/src/test/java/org/apache/fluss/server/kv/snapshot/KvSnapshotManagerTest.java
```diff
@@ -0,0 +1,1092 @@
...
+        snapshotContext =
+                DefaultSnapshotContext.create(
+                        zkClient,
+                        new TestingCompletedKvSnapshotCommitter(),
+                        kvSnapshotResource,
+                        conf);
+        kvSnapshotManager = createSnapshotManager(snapshotContext);
+        startPeriodicUploadSnapshot(NopUploadSnapshotTarget.INSTANCE);
+        // periodic snapshot is disabled when periodicMaterializeDelay is not positive
+        Assertions.assertEquals(0, scheduledExecutorService.getAllScheduledTasks().size());
```
platinumhamburg: Use AssertJ instead.

fluss-server/src/main/java/org/apache/fluss/server/kv/KvManager.java
```diff
@@ -311,39 +311,50 @@ public Optional<KvTablet> getKv(TableBucket tableBucket) {
         return Optional.ofNullable(currentKvs.get(tableBucket));
     }
 
-    public void dropKv(TableBucket tableBucket) {
-        KvTablet dropKvTablet =
+    public void closeOrDropKv(TableBucket tableBucket, boolean needDrop) {
```
platinumhamburg: Using the name closeThenDropKv or simply closeKv might be more accurate, as the current name looks like a choice between two actions.

fluss-server/src/main/java/org/apache/fluss/server/kv/snapshot/KvSnapshotManager.java
```diff
@@ -0,0 +1,1080 @@
...
+    private final AtomicInteger numberOfConsecutiveFailures;
+
+    /** Whether upload snapshot is started. */
+    private volatile boolean isLeader = false;
+
+    /** Whether the replica is a standby replica. */
+    private volatile boolean isStandby = false;
+
+    /** Whether the standby replica is initializing. */
+    private volatile boolean standbyInitializing = false;
```
platinumhamburg: Perhaps it would be simpler to use an enum to represent these orthogonal states?

fluss-server/src/main/java/org/apache/fluss/server/replica/Replica.java
```diff
@@ -684,6 +818,101 @@ public void updateTieredLogLocalSegments(int tieredLogLocalSegments) {
...
+     * download to finish before mutating the same tablet directory. See {@link
+     * #awaitStandbyDownload()}.
+     */
+    private void becomeStandbyAsync() {
+        checkNotNull(snapshotContext);
+        // Wait for any previous standby download to drain before submitting a new one. This
+        // prevents two downloaders from concurrently writing into the same db/ directory when
+        // the replica is repeatedly notified to become standby (e.g. across leader epochs).
+        awaitStandbyDownload();
+        synchronized (this) {
```
platinumhamburg, line 838: `becomeStandbyAsync()` calls `awaitStandbyDownload()` **outside** `synchronized(this)`, then assigns `standbyDownloadFuture` **inside** `synchronized(this)`. Between these two steps, `submitStandbyDownload()` (called from `ReplicaManager.notifyKvSnapshotOffset`, which runs **outside** `replicaStateChangeLock`) can slip in, chain a new download future, and start it on the executor.

This is reachable through a natural protocol sequence during leader failover: the old leader's last committed snapshot triggers `NotifyKvSnapshotOffsetRequest`, and the leader re-election triggers `NotifyLeaderAndIsrRequest` (re-assigning the same standby). Both RPCs arrive at the TabletServer independently and are handled by different RPC threads. `notifyKvSnapshotOffset` releases `replicaStateChangeLock` before calling `submitStandbyDownload`, so it does not block `makeFollower`.

```
Thread C (notifyKvSnapshotOffset, replicaStateChangeLock already released):
│ submitStandbyDownload(X):
│   synchronized(this):                      ← slips into the gap
│     F1 = sDF.handleAsync(downloadSnapshot(X))
│     sDF = F1                               ← F1 starts executing
│
Thread B (makeFollower, holds replicaStateChangeLock + leaderIsrUpdateLock(W)):
│ onBecomeNewFollower():
│   awaitStandbyDownload()       ← may see old sDF (before F1), returns
│   kvSnapshotManager.becomeStandby()   ← bumps epoch to E1
│   becomeStandbyAsync():
│     awaitStandbyDownload()     ← reads sDF; if F1 was created after the read, misses it
│     synchronized(this):
│       sDF = runAsync(downloadLatest) → F3    ← overwrites F1
```

**Result:** F1 is orphaned — no one awaits or tracks it. F1 and F3 execute concurrently, both writing to the same `db/` directory. `downloadEpoch` does not help because both capture the same epoch value (bumped once by `becomeStandby()` before `becomeStandbyAsync()`). `standbyInitializing` is also unreliable since F1 may have passed the check before F3 sets the flag.

**Suggestion:** Make `becomeStandbyAsync()` use the same chaining pattern as `submitStandbyDownload()`, so the new download is serialized after any in-flight one:

```java
private void becomeStandbyAsync() {
    synchronized (this) {
        standbyDownloadFuture = standbyDownloadFuture.handleAsync(
                (v, prevEx) -> {
                    try {
                        checkNotNull(kvSnapshotManager);
                        kvSnapshotManager.downloadLatestSnapshot();
                    } catch (Exception e) {
                        LOG.warn("Failed to download snapshot ...", e);
                    }
                    return null;
                },
                snapshotContext.getAsyncOperationsThreadPool());
    }
}
```

This eliminates the gap between await and assignment, and ensures all downloads — whether from `becomeStandbyAsync` or `submitStandbyDownload` — are serialized through the same future chain.

fluss-server/src/main/java/org/apache/fluss/server/replica/Replica.java
```diff
@@ -711,24 +940,50 @@ private void createKv() {
                         e);
             }
         }
-        // start periodic kv snapshot
-        startPeriodicKvSnapshot(snapshotUsed.orElse(null));
+        // start periodic upload kv snapshot
+        startPeriodicUploadKvSnapshot(snapshotUsed.orElse(null));
```
platinumhamburg: The Snapshot action consists of three sub-actions: checkpoint + upload + commit, so theoretically the method name should probably not be modified.

fluss-server/src/main/java/org/apache/fluss/server/replica/Replica.java
```diff
@@ -746,57 +1001,36 @@ private void mayFlushKv(long newHighWatermark) {
...
         try {
+            optCompletedSnapshot = kvSnapshotManager.downloadLatestSnapshot();
+            long restoreStartOffset = 0;
             Long rowCount;
             AutoIncIDRange autoIncIDRange;
             if (optCompletedSnapshot.isPresent()) {
                 LOG.info(
-                        "Use snapshot {} to restore kv tablet for {} of table {}.",
-                        optCompletedSnapshot.get(),
+                        "Init kv tablet for download latest snapshot {} of {} finish, cost {} ms.",
```
platinumhamburg: The log semantics are a bit strange. A more reasonable description would be Init KvTablet for TableBucket xxx with Snapshot xx.

fluss-server/src/test/java/org/apache/fluss/server/kv/snapshot/KvSnapshotManagerTest.java
```diff
@@ -0,0 +1,1092 @@
...
+    @BeforeAll
+    static void baseBeforeAll() {
+        zkClient =
+                ZOO_KEEPER_EXTENSION_WRAPPER
+                        .getCustomExtension()
+                        .getZooKeeperClient(NOPErrorHandler.INSTANCE);
+    }
+
+    @BeforeEach
+    void before() {
```
platinumhamburg, line 98: The design of before(0 seems unreasonable, as it causes a lot of resource leaks, repeated initialization, and waste.

platinumhamburg (changes requested):
@swuferhong  Thanks for the pr. Just few minor comments. PTAL

platinumhamburg (approved):
LGTM.

PR #2836, opened 2026-03-10
[FLUSS-2686] Add COS filesystem support

### LINKED ISSUE #2686: [File system] Add Tencent Cloud COS integration
### Search before asking

### Motivation

Add support for Tencent Cloud COS (Cloud Object Storage) file system integration to Fluss.

### Solution

Integrate Tencent Cloud COS Hadoop file system adapter class. Configuration options will be passed directly to the COS file system.
### Configuration Example

```properties
remote.data.dir: cosn://bucket/fluss
'fs.cosn.endpoint' = 'ap-guangzhou',
'fs.cosn.userinfo.secretId' = 'your-secret-id',
'fs.cosn.userinfo.secretKey' = 'your-secret-key'
```

## Implementation Plan

1. Create `fluss-fs-cos` module under `fluss-filesystems/`
2. Add Maven dependencies:
   - `org.apache.hadoop:hadoop-cos` 
   - `com.qcloud:cos_api`
3. Implement `CosFileSystemFactory` (similar to `OssFileSystemFactory` and `ObsFileSystemFactory`)
4. Add plugin service loader configuration
5. Implement credential/token provider if needed
6. Add unit tests and integration tests

## Testing

Local testing can be done with environment variables:
- `COSN_SECRET_ID`
- `COSN_SECRET_KEY`
- `COSN_ENDPOINT`
- `COSN_BUCKET`

## References

- Tencent Cloud COS: https://cloud.tencent.com/document/product/436
- Hadoop-COS: https://cloud.tencent.com/document/product/436/6884
- GitHub: https://github.com/tencentyun/hadoop-cos

## Benefits

- Enables Fluss users to use Tencent Cloud COS as remote storage
- Expands cloud provider support for Chinese market
- Consistent with existing file system plugin architecture

### Anything else?

_No response_

### Willingness to contribute

XuQianJin-Stars:
### Purpose

Linked issue: close #2686

This PR adds support for Tencent Cloud Object Storage (COS) as a remote filesystem for Fluss. COS is one of the most widely used cloud storage services in China and this integration enables users to store Fluss remote data on COS, similar to the existing OSS, OBS, S3, and GCS filesystem integrations.

### Brief change log

- **`fluss-fs-cos/pom.xml`**: Add module configuration with dependencies on `hadoop-cos` (3.3.5), `cos_api` (5.6.139), and `fluss-fs-hadoop` for Hadoop filesystem abstraction.
- **`COSFileSystem`**: Extends `HadoopFileSystem` to wrap Hadoop's `CosNFileSystem`, with lazy-initialized security token provider support for obtaining temporary credentials.
- **`COSFileSystemPlugin`**: Implements `FileSystemPlugin` SPI for the `cosn` scheme. Handles Hadoop configuration translation from Fluss config (prefix `fs.cosn.*`), supports three credential modes: static secret key, custom credentials provider, and dynamic security token via `COSSecurityTokenReceiver`.
- **`COSSecurityTokenProvider`**: Generates temporary security tokens using Tencent Cloud STS (Security Token Service) based on existing `secretId`/`secretKey` configuration.
- **`COSSecurityTokenReceiver`**: Implements `SecurityTokenReceiver` SPI to receive and apply security tokens, configuring `DynamicTemporaryCOSCredentialsProvider` as the Hadoop credentials provider.
- **`DynamicTemporaryCOSCredentialsProvider`**: Implements COS SDK's `COSCredentialsProvider` interface, providing `BasicSessionCredentials` from tokens received via `COSSecurityTokenReceiver`.
- **`fluss-filesystems/pom.xml`**: Register `fluss-fs-cos` as a sub-module.

### Tests

- `COSFileSystemBehaviorITCase`: Integration test for basic COS filesystem behavior (create, read, write, delete) using static credentials (`secretId`/`secretKey`).
- `COSWithTokenFileSystemBehaviorITCase`: Integration test for COS filesystem behavior using dynamic security tokens obtained via STS.
- `COSWithTokenFileSystemBehaviorBaseITCase`: Base class for token-based filesystem tests, handling filesystem initialization with both static credentials and security tokens.
- `COSTestCredentials`: Test utility that reads COS credentials and endpoint from environment variables (`COSN_SECRET_ID`, `COSN_SECRET_KEY`, `COSN_ENDPOINT`, `COSN_BUCKET`).

> Note: These are IT (integration tests) that require actual COS credentials and bucket access. They are skipped automatically when credentials are not available via `Assumptions.assumeTrue`.

### API and Format

No existing API or storage format changes. This PR only adds a new filesystem plugin that registers via SPI (`META-INF/services/org.apache.fluss.fs.FileSystemPlugin`).

New configuration keys introduced (all following standard Hadoop COS conventions):
- `fs.cosn.endpoint` — COS region endpoint (e.g., `ap-guangzhou`)
- `fs.cosn.userinfo.secretId` — Tencent Cloud secret ID
- `fs.cosn.userinfo.secretKey` — Tencent Cloud secret key
- `fs.cosn.credentials.provider` — Custom credentials provider class

### Documentation

This PR introduces a new feature. Documentation for COS filesystem configuration should be added to the Fluss documentation site in a follow-up. Usage example:

```yaml
remote.data.dir: cosn://bucket-name/fluss-data
fs.cosn.endpoint: ap-guangzhou
fs.cosn.userinfo.secretId: <your-secret-id>
fs.cosn.userinfo.secretKey: <your-secret-key>

### DIFF
--- a/fluss-filesystems/fluss-fs-cos/pom.xml
+++ b/fluss-filesystems/fluss-fs-cos/pom.xml
@@ -1,0 +1,254 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+ Licensed to the Apache Software Foundation (ASF) under one
+ or more contributor license agreements.  See the NOTICE file
+ distributed with this work for additional information
+ regarding copyright ownership.  The ASF licenses this file
+ to you under the Apache License, Version 2.0 (the
+ "License"); you may not use this file except in compliance
+ with the License.  You may obtain a copy of the License at
+
+      http://www.apache.org/licenses/LICENSE-2.0
+
+ Unless required by applicable law or agreed to in writing, software
+ distributed under the License is distributed on an "AS IS" BASIS,
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+-->
+
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+    <parent>
+        <groupId>org.apache.fluss</groupId>
+        <artifactId>fluss-filesystems</artifactId>
+        <version>0.10-SNAPSHOT</version>
+    </parent>
+
+    <artifactId>fluss-fs-cos</artifactId>
+    <name>Fluss : FileSystems : COS FS</name>
+
+    <properties>
+        <fs.cosn.sdk.version>3.3.5</fs.cosn.sdk.version>
+        <fs.cos.api.version>5.6.139</fs.cos.api.version>
+    </properties>
+
+    <dependencies>
+        <dependency>
+            <groupId>org.apache.fluss</groupId>
+            <artifactId>fluss-common</artifactId>
+            <version>${project.version}</version>
+            <scope>provided</scope>
+        </dependency>
+
+        <dependency>
+            <groupId>org.apache.fluss</groupId>
+            <artifactId>fluss-fs-hadoop-shaded</artifactId>
+            <version>${project.version}</version>
+        </dependency>
+
+        <dependency>
+            <groupId>org.apache.fluss</groupId>
+            <artifactId>fluss-fs-hadoop</artifactId>
+            <version>${project.version}</version>
+        </dependency>
+
+        <dependency>
+            <groupId>org.apache.hadoop</groupId>
+            <artifactId>hadoop-cos</artifactId>
+            <version>${fs.cosn.sdk.version}</version>
+            <exclusions>
+                <exclusion>
+                    <groupId>com.qcloud</groupId>
+                    <artifactId>cos_api</artifactId>
+                </exclusion>
+                <exclusion>
+                    <groupId>org.apache.hadoop</groupId>
+                    <artifactId>hadoop-common</artifactId>
+                </exclusion>
+                <exclusion>
+                    <groupId>ch.qos.reload4j</groupId>
+                    <artifactId>reload4j</artifactId>
+                </exclusion>
+                <exclusion>
+                    <groupId>org.slf4j</groupId>
+                    <artifactId>slf4j-reload4j</artifactId>
+                </exclusion>
+            </exclusions>
+        </dependency>
+
+        <dependency>
+            <groupId>com.qcloud</groupId>
+            <artifactId>cos_api</artifactId>
+            <version>${fs.cos.api.version}</version>
+            <exclusions>
+                <exclusion>
+                    <groupId>javax.xml.bind</groupId>
+                    <artifactId>jaxb-api</artifactId>
+                </exclusion>
+            </exclusions>
+        </dependency>
+
+        <dependency>
+            <!-- Hadoop requires jaxb-api for javax.xml.bind.JAXBException -->
+            <groupId>javax.xml.bind</groupId>
+            <artifactId>jaxb-api</artifactId>
+            <version>${jaxb.api.version}</version>
+            <!-- packaged as an optional dependency that is only accessible on Java 11+ -->
+            <scope>provided</scope>
+        </dependency>
+
+        <dependency>
+            <groupId>org.apache.fluss</groupId>
+            <artifactId>fluss-test-utils</artifactId>
+        </dependency>
+        <!-- for the behavior test suite -->
+        <dependency>
+            <groupId>org.apache.fluss</groupId>
+            <artifactId>fluss-common</artifactId>
+            <version>${project.version}</version>
+            <scope>test</scope>
+            <type>test-jar</type>
+        </dependency>
+    </dependencies>
+
+    <build>
+        <plugins>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-jar-plugin</artifactId>
+                <executions>
+                    <execution>
+                        <goals>
+                            <goal>jar</goal>
+                        </goals>
+                    </execution>
+                </executions>
+                <configuration>
+                    <archive>
+                        <manifestEntries>
+                            <!-- jaxb-api is packaged as an optional dependency that is only accessible on Java 11 -->
+                            <Multi-Release>true</Multi-Release>
+                        </manifestEntries>
+                    </archive>
+                </configuration>
+            </plugin>
+
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-dependency-plugin</artifactId>
+                <executions>
+                    <execution>
+                        <id>copy-javax-jars</id>
+                        <phase>process-resources</phase>
+                        <goals>
+                            <goal>copy</goal>
+                        </goals>
+                    </execution>
+                </executions>
+                <configuration>
+                    <artifactItems>
+                        <artifactItem>
+                            <groupId>javax.xml.bind</groupId>
+                            <artifactId>jaxb-api</artifactId>
+                            <version>${jaxb.api.version}</version>
+                            <type>jar</type>
+                            <overWrite>true</overWrite>
+                        </artifactItem>
+                    </artifactItems>
+                    <outputDirectory>${project.build.directory}/temporary</outputDirectory>
+                </configuration>
+            </plugin>
+
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-antrun-plugin</artifactId>
+                <executions>
+                    <execution>
+                        <id>unpack-javax-libraries</id>
+                        <phase>process-resources</phase>
+                        <goals>
+                            <goal>run</goal>
+                        </goals>
+                        <configuration>
+                            <target>
+                                <echo message="unpacking javax jars"/>
+                                <unzip dest="${project.build.directory}/classes/META-INF/versions/11">
+                                    <fileset dir="${project.build.directory}/temporary">
+                                        <include name="*"/>
+                                    </fileset>
+                                </unzip>
+                            </target>
+                        </configuration>
+                    </execution>
+                </executions>
+            </plugin>
+
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-shade-plugin</artifactId>
+                <executions>
+                    <execution>
+                        <id>shade-fluss</id>
+                        <phase>package</phase>
+                        <goals>
+                            <goal>shade</goal>
+                        </goals>
+                        <configuration>
+                            <artifactSet>
+                                <includes>
+                                    <include>*:*</include>
+                                </includes>
+                                <excludes>
+                                    <exclude>javax.servlet:servlet-api</exclude>
+                                    <exclude>xmlenc:xmlenc</exclude>
+                                </excludes>
+                            </artifactSet>
+                            <filters>
+                                <filter>
+                                    <artifact>*</artifact>
+                                    <excludes>
+                                        <exclude>.gitkeep</exclude>
+                                        <exclude>mime.types</exclude>
+                                        <exclude>mozilla/**</exclude>
+                                        <exclude>LICENSE.txt</exclude>
+                                        <exclude>license/LICENSE*</exclude>
+                                        <exclude>okhttp3/internal/publicsuffix/NOTICE</exclude>
+                                        <exclude>NOTICE</exclude>
+                                    </excludes>
+                                </filter>
+                                <filter>
+                                    <artifact>org.apache.fluss:fluss-fs-hadoop</artifact>
+                                    <excludes>
+                                        <exclude>META-INF/**</exclude>
+                                    </excludes>
+                                </filter>
+                            </filters>
+                            <relocations>
+                                <relocation>
+                                    <pattern>org.apache.commons</pattern>
+                                    <shadedPattern>org.apache.fluss.shaded.org.apache.commons</shadedPattern>
+                                </relocation>
+                            </relocations>
+                        </configuration>
+                    </execution>
+                </executions>
+            </plugin>
+
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-jar-plugin</artifactId>
+                <executions>
+                    <execution>
+                        <goals>
+                            <goal>test-jar</goal>
+                        </goals>
+                    </execution>
+                </executions>
+            </plugin>
+        </plugins>
+    </build>
+
+</project>
--- a/fluss-filesystems/fluss-fs-cos/pom.xml
+++ b/fluss-filesystems/fluss-fs-cos/pom.xml
@@ -121,8 +121,15 @@
                 <artifactId>maven-jar-plugin</artifactId>
                 <executions>
                     <execution>
+                        <id>default-jar</id>
                         <goals>
                             <goal>jar</goal>
+                        </goals>
+                    </execution>
+                    <execution>
+                        <id>test-jar</id>
+                        <goals>
+                            <goal>test-jar</goal>
                         </goals>
                     </execution>
                 </executions>
@@ -237,17 +244,6 @@
                 </executions>
             </plugin>
 
-            <plugin>
-                <groupId>org.apache.maven.plugins</groupId>
-                <artifactId>maven-jar-plugin</artifactId>
-                <executions>
-                    <execution>
-                        <goals>
-                            <goal>test-jar</goal>
-                        </goals>
-                    </execution>
-                </executions>
-            </plugin>
         </plugins>
     </build>
 
--- a/fluss-filesystems/fluss-fs-cos/pom.xml
+++ b/fluss-filesystems/fluss-fs-cos/pom.xml
@@ -24,7 +24,7 @@
     <parent>
         <groupId>org.apache.fluss</groupId>
         <artifactId>fluss-filesystems</artifactId>
-        <version>0.10-SNAPSHOT</version>
+        <version>1.0-SNAPSHOT</version>
     </parent>
 
     <artifactId>fluss-fs-cos</artifactId>
--- a/fluss-filesystems/fluss-fs-cos/pom.xml
+++ b/fluss-filesystems/fluss-fs-cos/pom.xml
@@ -65,6 +65,16 @@
                     <groupId>com.qcloud</groupId>
                     <artifactId>cos_api</artifactId>
                 </exclusion>
+                <!--
+                  Exclude the bundled COS SDK shipped by hadoop-cos to avoid having two
+                  COS SDK variants on the classpath (cos_api-bundle:5.6.69 and the directly
+                  declared cos_api:${fs.cos.api.version}), which produces hundreds of
+                  overlapping com.qcloud.cos.* classes during shading.
+                -->
+                <exclusion>
+                    <groupId>com.qcloud</groupId>
+                    <artifactId>cos_api-bundle</artifactId>
+                </exclusion>
                 <exclusion>
                     <groupId>org.apache.hadoop</groupId>
                     <artifactId>hadoop-common</artifactId>
--- a/fluss-filesystems/fluss-fs-cos/pom.xml
+++ b/fluss-filesystems/fluss-fs-cos/pom.xml
@@ -33,6 +33,7 @@
     <properties>
         <fs.cosn.sdk.version>3.3.5</fs.cosn.sdk.version>
         <fs.cos.api.version>5.6.139</fs.cos.api.version>
+        <fs.cos.sts.sdk.version>3.1.678</fs.cos.sts.sdk.version>
     </properties>
 
     <dependencies>
@@ -83,6 +84,18 @@
             <groupId>com.qcloud</groupId>
             <artifactId>cos_api</artifactId>
             <version>${fs.cos.api.version}</version>
+            <exclusions>
+                <exclusion>
+                    <groupId>javax.xml.bind</groupId>
+                    <artifactId>jaxb-api</artifactId>
+                </exclusion>
+            </exclusions>
+        </dependency>
+
+        <dependency>
+            <groupId>com.tencentcloudapi</groupId>
+            <artifactId>tencentcloud-sdk-java-sts</artifactId>
+            <version>${fs.cos.sts.sdk.version}</version>
             <exclusions>
                 <exclusion>
                     <groupId>javax.xml.bind</groupId>
--- a/fluss-filesystems/fluss-fs-cos/pom.xml
+++ b/fluss-filesystems/fluss-fs-cos/pom.xml
@@ -31,7 +31,7 @@
     <name>Fluss : FileSystems : COS FS</name>
 
     <properties>
-        <fs.cosn.sdk.version>3.3.5</fs.cosn.sdk.version>
+        <fs.hadoop.cos.version>3.3.5</fs.hadoop.cos.version>
         <fs.cos.api.version>5.6.139</fs.cos.api.version>
         <fs.cos.sts.sdk.version>3.1.678</fs.cos.sts.sdk.version>
     </properties>
@@ -59,7 +59,7 @@
         <dependency>
             <groupId>org.apache.hadoop</groupId>
             <artifactId>hadoop-cos</artifactId>
-            <version>${fs.cosn.sdk.version}</version>
+            <version>${fs.hadoop.cos.version}</version>
             <exclusions>
                 <exclusion>
                     <groupId>com.qcloud</groupId>
--- a/fluss-filesystems/fluss-fs-cos/src/main/java/org/apache/fluss/fs/cos/COSFileSystemPlugin.java
+++ b/fluss-filesystems/fluss-fs-cos/src/main/java/org/apache/fluss/fs/cos/COSFileSystemPlugin.java
@@ -1,0 +1,133 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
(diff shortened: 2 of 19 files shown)

### CODE REVIEW

**Comment by XuQianJin-Stars** [comment]
hi @wuchong @leekeiabstraction @luoyuxia Hi, i already updated the pr. Please help review when you got some time.

File: fluss-filesystems/fluss-fs-cos/pom.xml [hunk 1]
```diff
@@ -0,0 +1,250 @@
...
+            <!-- Hadoop requires jaxb-api for javax.xml.bind.JAXBException -->
+            <groupId>javax.xml.bind</groupId>
+            <artifactId>jaxb-api</artifactId>
+            <version>${jaxb.api.version}</version>
+            <!-- packaged as an optional dependency that is only accessible on Java 11+ -->
+            <scope>provided</scope>
+        </dependency>
+
+        <dependency>
+            <groupId>org.apache.fluss</groupId>
```
leekeiabstraction, line 127: Should this be in test scope?

fluss-filesystems/fluss-fs-cos/src/main/java/org/apache/fluss/fs/cos/token/COSSecurityTokenProvider.java
```diff
@@ -0,0 +1,63 @@
...
+
+import java.util.HashMap;
+import java.util.Map;
+
+import static org.apache.fluss.fs.cos.COSFileSystemPlugin.ENDPOINT_KEY;
+
+/** A provider to provide Tencent Cloud COS security token. */
+public class COSSecurityTokenProvider {
+
+    private static final String SECRET_ID = "fs.cosn.userinfo.secretId";
```
leekeiabstraction: This string constant seems to be declared in a few classes. Consider sharing single pub static final?

fluss-filesystems/fluss-fs-cos/src/test/java/org/apache/fluss/fs/cos/COSFileSystemBehaviorITCase.java
```diff
@@ -0,0 +1,63 @@
...
+import org.junit.jupiter.api.AfterAll;
+import org.junit.jupiter.api.BeforeAll;
+
+import java.util.UUID;
+
+/**
+ * An implementation of the {@link FileSystemBehaviorTestSuite} for the COS file system with Hadoop
+ * cos sdk.
+ */
+class COSFileSystemBehaviorITCase extends FileSystemBehaviorTestSuite {
```
leekeiabstraction, line 38: Was the IT test run successful with creds set? Would be good to indicate in pr description

leekeiabstraction (commented):
TY for the PR. Left some comments.

fluss-filesystems/fluss-fs-cos/pom.xml
XuQianJin-Stars: > Should this be in test scope?

The test scope for fluss-test-utils is already defined in the parent pom's dependencyManagement section (line 336 in root pom.xml), so child modules inherit it automatically without needing to declare <scope>test</scope> explicitly. This is consistent with how other filesystem modules (oss, s3, gs, azure, obs, hdfs) declare this dependency.

fluss-filesystems/fluss-fs-cos/src/test/java/org/apache/fluss/fs/cos/COSFileSystemBehaviorITCase.java
XuQianJin-Stars: > Was the IT test run successful with creds set? Would be good to indicate in pr description

Yes, the IT tests have been run successfully with COS credentials configured.

website/docs/_configs/_partial_config.mdx
```diff
@@ -78,7 +78,7 @@
 | `client.lookup.batch-timeout` | `0 s` | Duration | The maximum time to wait for the lookup batch to full, if this timeout is reached, the lookup batch will be closed to send. |
 | `client.lookup.max-retries` | `2147483647` | Integer | Setting a value greater than zero will cause the client to resend any lookup request that fails with a potentially transient error. |
 | `client.scanner.remote-log.prefetch-num` | `4` | Integer | The number of remote log segments to keep in local temp file for LogScanner, which download from remote storage. The default setting is 4. |
-| `client.scanner.io.tmpdir` | `/var/folders/bp/v2l48kz51mx86d743qv0zhzh0000gn/T//fluss` | String | Local directory that is used by client for storing the data files (like kv snapshot, log segment files) to read temporarily |
+| `client.scanner.io.tmpdir` | `/var/folders/7r/lwdsh9ms4gn0fnxs8c6fcfpm0000gn/T//fluss` | String | Local directory that is used by client for storing the data files (like kv snapshot, log segment files) to read temporarily |
```
leekeiabstraction: Looks like unrelated changes crept back into this file

website/docs/_configs/_partial_config.mdx
```diff
@@ -238,9 +238,13 @@
...
+| `remote.data.dirs.strategy` | `ROUND_ROBIN` | RemoteDataDirStrategy | The strategy for selecting the remote data directory from `remote.data.dirs`. The candidate strategies are: [ROUND_ROBIN, WEIGHTED_ROUND_ROBIN], the default strategy is ROUND_ROBIN. ROUND_ROBIN: this strategy employs a round-robin approach to select one from the available remote directories. WEIGHTED_ROUND_ROBIN: this strategy selects one of the available remote directories based on the weights configured in `remote.data.dirs.weights`. |
+| `remote.data.dirs.weights` | `[]` | ArrayList | The weights of the remote data directories. This is a list of weights corresponding to the `remote.data.dirs` in the same order. When `remote.data.dirs.strategy` is set to `WEIGHTED_ROUND_ROBIN`, this must be configured, and its size must be equal to `remote.data.dirs`; otherwise, it will be ignored. |
```
leekeiabstraction: ditto

website/docs/_configs/_partial_config.mdx
```diff
@@ -238,9 +238,13 @@
...
 | `remote.fs.write-buffer-size` | `4 kb` | MemorySize | The default size of the write buffer for writing the local files to remote file systems. |
 | `remote.log.task-interval-duration` | `1 min` | Duration | Interval at which remote log manager runs the scheduled tasks like copy segments, clean up remote log segments, delete local log segments etc. If the value is set to 0, it means that the remote log storage is disabled. |
+| `remote.log.task-max-upload-segments` | `5` | Integer | The maximum number of log segments to upload to remote storage per tiering task execution. This limits the upload batch size to prevent overwhelming the remote storage when there is a large backlog of segments to upload. |
```
leekeiabstraction: ditto

website/docs/_configs/_partial_config.mdx
```diff
@@ -284,7 +288,7 @@
...
 | `table.auto-partition.time-unit` | `DAY` | AutoPartitionTimeUnit | The time granularity for auto created partitions. The default value is `DAY`. Valid values are `HOUR`, `DAY`, `MONTH`, `QUARTER`, `YEAR`. If the value is `HOUR`, the partition format for auto created is yyyyMMddHH. If the value is `DAY`, the partition format for auto created is yyyyMMdd. If the value is `MONTH`, the partition format for auto created is yyyyMM. If the value is `QUARTER`, the partition format for auto created is yyyyQ. If the value is `YEAR`, the partition format for auto created is yyyy. |
-| `table.auto-partition.time-zone` | `Europe/Paris` | String | The time zone for auto partitions, which is by default the same as the system time zone. |
+| `table.auto-partition.time-zone` | `Asia/Shanghai` | String | The time zone for auto partitions, which is by default the same as the system time zone. |
```
leekeiabstraction: ditto

leekeiabstraction (commented):
Looks consistent to other filesystem support. Left comments on unrelated changes being included in this PR (looks like it crept back in), can approve once this is addresed. 

I think we should have someone else who has access to COS to test this as well as I do not have access to COS to manually test. cc @luoyuxia

XuQianJin-Stars:
> Looks consistent to other filesystem support. Left comments on unrelated changes being included in this PR (looks like it crept back in), can approve once this is addresed.
> 
> I think we should have someone else who has access to COS to test this as well as I do not have access to COS to manually test. cc @luoyuxia

hi @leekeiabstraction @luoyuxia Thanks for the review!

I've tested the COS filesystem support with a real COS bucket (region: ap-guangzhou). Here are the test details:

**Test classes executed:**
- `COSFileSystemBehaviorITCase` — Tests COS filesystem with SecretId/SecretKey authentication
- `COSWithTokenFileSystemBehaviorITCase` — Tests COS filesystem with security token authentication

Both classes inherit from `FileSystemBehaviorTestSuite`, which includes the following 19 test cases (per class, 38 total):

| # | Test Method | Description |
|---|------------|-------------|
| 1 | `testPathAndScheme` | Verify URI path and scheme correctness |
| 2 | `testFileExists` | Check existing file detection |
| 3 | `testFileDoesNotExist` | Check non-existing file detection |
| 4 | `testExistingFileDeletion` | Delete an existing file |
| 5 | `testExistingFileRecursiveDeletion` | Recursively delete an existing file |
| 6 | `testNotExistingFileDeletion` | Delete a non-existing file |
| 7 | `testNotExistingFileRecursiveDeletion` | Recursively delete a non-existing file |
| 8 | `testExistingEmptyDirectoryDeletion` | Delete an existing empty directory |
| 9 | `testExistingEmptyDirectoryRecursiveDeletion` | Recursively delete an existing empty directory |
| 10 | `testExistingNonEmptyDirectoryDeletion` | Delete a non-empty directory |
| 11 | `testExistingNonEmptyDirectoryRecursiveDeletion` | Recursively delete a non-empty directory |
| 12 | `testExistingNonEmptyDirectoryWithSubDirRecursiveDeletion` | Recursively delete a non-empty directory with subdirectories |
| 13 | `testMkdirsReturnsTrueWhenCreatingDirectory` | Create directory returns true |
| 14 | `testMkdirsCreatesParentDirectories` | Create parent directories |
| 15 | `testMkdirsReturnsTrueForExistingDirectory` | Mkdirs on existing directory returns true |
| 16 | `testMkdirsFailsForExistingFile` | Mkdirs fails when path is an existing file |
| 17 | `testMkdirsFailsWithExistingParentFile` | Mkdirs fails when parent is a file |
| 18 | `testSimpleFileWriteAndRead` | Write and read file content |
| 19 | `testDirectoryListing` | List directory contents |

**Result: All 38 tests passed ✅** (19 × 2 test classes, 0 failures, 0 errors)

XuQianJin-Stars:
hi @leekeiabstraction @luoyuxia Hi, i already updated the pr. Please help review when you got some time.

fluss-filesystems/fluss-fs-cos/src/main/java/org/apache/fluss/fs/cos/COSFileSystemPlugin.java
```diff
@@ -0,0 +1,134 @@
...
+     * In order to simplify, we make fluss cos configuration keys same with hadoop cos module. So,
+     * we add all configuration key with prefix `fs.cosn` in fluss conf to hadoop conf.
+     */
+    private static final String[] FLUSS_CONFIG_PREFIXES = {"fs.cosn."};
+
+    public static final String SECRET_ID = "fs.cosn.userinfo.secretId";
+    public static final String SECRET_KEY = "fs.cosn.userinfo.secretKey";
+    public static final String CREDENTIALS_PROVIDER = "fs.cosn.credentials.provider";
+
+    public static final String ENDPOINT_KEY = "fs.cosn.bucket.endpoint_suffix";
```
leekeiabstraction, line 53: Just noticed that this is different to what's in the PR description `fs.cosn.endpoint`. Which is the configuration that we are exposing?  `fs.cosn.endpoint` or `fs.cosn.bucket.endpoint_suffix`?

fluss-filesystems/fluss-fs-cos/src/main/java/org/apache/fluss/fs/cos/COSFileSystemPlugin.java
```diff
@@ -0,0 +1,134 @@
...
+
+    /**
+     * In order to simplify, we make fluss cos configuration keys same with hadoop cos module. So,
+     * we add all configuration key with prefix `fs.cosn` in fluss conf to hadoop conf.
+     */
+    private static final String[] FLUSS_CONFIG_PREFIXES = {"fs.cosn."};
+
+    public static final String SECRET_ID = "fs.cosn.userinfo.secretId";
+    public static final String SECRET_KEY = "fs.cosn.userinfo.secretKey";
+    public static final String CREDENTIALS_PROVIDER = "fs.cosn.credentials.provider";
```
leekeiabstraction, line 49: Seems like region is a required property, but we are not defining it here?

Ref: https://hadoop.apache.org/docs/stable/hadoop-cos/cloud-storage/index.html

> fs.cosn.userinfo.region is an required property for Hadoop-COS. The reason is that Hadoop-COS must know the region of the using bucket in order to accurately construct a URL to access it.

leekeiabstraction (commented):
TY for addressing PR comments. Left a couple more questions

fluss-filesystems/fluss-fs-cos/src/main/java/org/apache/fluss/fs/cos/token/COSSecurityTokenProvider.java
```diff
@@ -0,0 +1,62 @@
...
+
+    public COSSecurityTokenProvider(Configuration conf) {
+        endpoint = conf.get(ENDPOINT_KEY);
+        secretId = conf.get(SECRET_ID);
+        secretKey = conf.get(SECRET_KEY);
+    }
+
+    public ObtainedSecurityToken obtainSecurityToken(String scheme) {
+        // For COS, we directly use the configured secret id and secret key as the token.
+        // If STS temporary credentials are needed in the future, this can be extended
```
leekeiabstraction: I think we need to update PR description to be consistent with behaviour here

> COSSecurityTokenProvider: Generates temporary security tokens using Tencent Cloud STS (Security Token Service) based on existing secretId/secretKey configuration.

Please also create issue to track implementation of using STS to generate token/

Actually, why not implement as part of this PR? I'd like to understand the reasoning.

XuQianJin-Stars:
> TY for addressing PR comments. Left a couple more questions

Thanks for the review! I've addressed all three previous comments in the latest push:

1. STS Implementation: `COSSecurityTokenProvider` now calls the Tencent Cloud STS GetFederationToken API to obtain real temporary credentials (tmpSecretId, tmpSecretKey, sessionToken), consistent with the S3/OSS/OBS implementations. Added tencentcloud-sdk-java-sts:3.1.678 dependency.
2. REGION constant: Added public static final String REGION = "fs.cosn.userinfo.region" to `COSFileSystemPlugin`, and it's now used in `COSSecurityTokenProvider` for STS API calls and passed through `additionInfos`.
3. Endpoint key consistency: The code uses fs.cosn.bucket.endpoint_suffix which is the standard Hadoop-COS configuration key. Updated PR description to match.

Also updated the `COSSecurityTokenReceiver` to always use BasicSessionCredentials (since STS always returns temporary credentials with a session token), updated the NOTICE file, and added test coverage for the `REGION` config.
Please let me know about the additional questions — happy to address them!

fluss-filesystems/fluss-fs-cos/src/main/java/org/apache/fluss/fs/cos/COSFileSystemPlugin.java
```diff
@@ -0,0 +1,136 @@
...
+    }
+
+    @Override
+    public FileSystem create(URI fsUri, Configuration flussConfig) throws IOException {
+        org.apache.hadoop.conf.Configuration hadoopConfig = getHadoopConfiguration(flussConfig);
+
+        // set credential provider
+        if (hadoopConfig.get(SECRET_ID) == null) {
+            String credentialsProvider = hadoopConfig.get(CREDENTIALS_PROVIDER);
+            if (credentialsProvider != null) {
```
leekeiabstraction, line 67: Do we need to have IT for this branch?

fluss-filesystems/fluss-fs-cos/pom.xml
```diff
@@ -0,0 +1,263 @@
...
+            <artifactId>fluss-fs-hadoop</artifactId>
+            <version>${project.version}</version>
+        </dependency>
+
+        <dependency>
+            <groupId>org.apache.hadoop</groupId>
+            <artifactId>hadoop-cos</artifactId>
+            <version>${fs.cosn.sdk.version}</version>
```
leekeiabstraction, line 62: This could be better named as hadoop cos version. It is not the sdk version.

fluss-filesystems/fluss-fs-cos/src/main/java/org/apache/fluss/fs/cos/token/COSSecurityTokenProvider.java
```diff
@@ -0,0 +1,128 @@
...
+ */
+public class COSSecurityTokenProvider {
+
+    private static final Logger LOG = LoggerFactory.getLogger(COSSecurityTokenProvider.class);
+
+    /** Default federation token name. */
+    private static final String FEDERATION_TOKEN_NAME = "fluss-cos-federation";
+
+    /** Default duration seconds for temporary credentials, 1800s = 30min. */
+    private static final long DEFAULT_DURATION_SECONDS = 1800L;
```
leekeiabstraction, line 55: Curious on why picking lower default than other implementation like OSS/OBS

leekeiabstraction (commented):
TY for the updates. Added further comments.

fluss-filesystems/fluss-fs-cos/src/main/java/org/apache/fluss/fs/cos/COSFileSystemPlugin.java
XuQianJin-Stars: > Do we need to have IT for this branch?

Good catch — I've added `COSWithCredentialsProviderFileSystemBehaviorITCase` in the latest commit to cover this `CREDENTIALS_PROVIDER` branch.
It reuses `org.apache.hadoop.fs.cosn.auth.EnvironmentVariableCredentialsProvider` shipped with hadoop-cos, which reads `COSN_SECRET_ID / COSN_SECRET_KEY` from the environment — exactly the same variables already required by `COSTestCredentials`. So no extra CI configuration is needed: whenever the existing COS ITs are enabled, this new one runs too. PTAL.

fluss-filesystems/fluss-fs-cos/src/test/java/org/apache/fluss/fs/cos/COSTestCredentials.java
```diff
@@ -0,0 +1,119 @@
...
+
+package org.apache.fluss.fs.cos;
+
+import javax.annotation.Nullable;
+
+import static org.junit.jupiter.api.Assumptions.assumeTrue;
+
+/** Access to credentials to access COS buckets during integration tests. */
+public class COSTestCredentials {
+    @Nullable private static final String ENDPOINT = System.getenv("COSN_ENDPOINT");
```
leekeiabstraction: Should this be endpoint suffix?

leekeiabstraction (commented):
TY for addressing comments. Left one more question on one of the test case. Also, are the CI test failures related to changes within this PR?

simonsssu:
@leekeiabstraction  Hi lee, I have tested this manually.

luoyuxia:
Thanks for adding COS support. I reviewed the implementation and found a few issues that should be addressed before merge.

1. **[major] STS tokens are too broadly scoped**

`COSSecurityTokenProvider` currently uses this default policy:

```java
"action": ["name/cos:*"],
"resource": ["*"]
```

That means the temporary credentials distributed to Fluss clients can access any COS resource allowed by the permanent server-side secret. This is broader than the Fluss remote storage path and could expose non-Fluss data if the permanent credential has wide permissions.

Can we scope the token policy to the configured `remote.data.dir` bucket/prefix, or require an explicit user-provided token policy instead of defaulting to `resource: "*"`?

2. **[major] The shaded plugin currently contains two COS SDK variants**

`hadoop-cos:3.3.5` brings in `com.qcloud:cos_api-bundle:5.6.69`, while this module also directly depends on `com.qcloud:cos_api:5.6.139`. During `package`, Maven Shade reports 902 overlapping `com.qcloud.cos.*` classes/resources between the two jars.

That makes the actual runtime COS SDK version depend on shade ordering and can produce subtle incompatibilities. Please keep only one COS SDK variant, for example by excluding `cos_api-bundle` from `hadoop-cos` if the direct `cos_api` dependency is required, or by dropping the direct `cos_api` dependency if the bundled one is sufficient.

3. **[minor] COS documentation uses the wrong environment provider class/env vars**

The docs show:

```yaml
fs.cosn.credentials.provider: org.apache.hadoop.fs.cosn.auth.EnvironmentVariableCredentialProvider
```

but the actual hadoop-cos class is `org.apache.hadoop.fs.cosn.auth.EnvironmentVariableCredentialsProvider` (plural `Credentials`). Also, the provider reads `COSN_SECRET_ID` and `COSN_SECRET_KEY`, while the docs say `COS_SECRET_ID` and `COS_SECRET_KEY`. Users following the docs would fail authentication.

4. **[minor] Token lifetime in docs does not match the code**

The docs say the temporary credential is valid for 30 minutes, but `COSSecurityTokenProvider.DEFAULT_DURATION_SECONDS` is `3600L` (1 hour). Please align the docs or the code.

5. **[nit] `cos.md` has trailing whitespace**

`git diff --check` reports trailing whitespace in `website/docs/maintenance/filesystems/cos.md` lines 14 and 17.

Validation I ran:

```bash
./mvnw -pl fluss-filesystems/fluss-fs-cos -am -DskipTests compile
./mvnw -pl fluss-filesystems/fluss-fs-cos -am -DskipTests test-compile
./mvnw -pl fluss-filesystems/fluss-fs-cos -am -DskipTests package
git diff --check
```

The Maven commands passed; `package` emitted the duplicate COS SDK warnings mentioned above, and `git diff --check` failed on the trailing whitespace.

XuQianJin-Stars:
> Thanks for adding COS support. I reviewed the implementation and found a few issues that should be addressed before merge.
> 
> 1. **[major] STS tokens are too broadly scoped**
> 
> `COSSecurityTokenProvider` currently uses this default policy:
> 
> ```java
> "action": ["name/cos:*"],
> "resource": ["*"]
> ```
> 
> That means the temporary credentials distributed to Fluss clients can access any COS resource allowed by the permanent server-side secret. This is broader than the Fluss remote storage path and could expose non-Fluss data if the permanent credential has wide permissions.
> 
> Can we scope the token policy to the configured `remote.data.dir` bucket/prefix, or require an explicit user-provided token policy instead of defaulting to `resource: "*"`?
> 
> 2. **[major] The shaded plugin currently contains two COS SDK variants**
> 
> `hadoop-cos:3.3.5` brings in `com.qcloud:cos_api-bundle:5.6.69`, while this module also directly depends on `com.qcloud:cos_api:5.6.139`. During `package`, Maven Shade reports 902 overlapping `com.qcloud.cos.*` classes/resources between the two jars.
> 
> That makes the actual runtime COS SDK version depend on shade ordering and can produce subtle incompatibilities. Please keep only one COS SDK variant, for example by excluding `cos_api-bundle` from `hadoop-cos` if the direct `cos_api` dependency is required, or by dropping the direct `cos_api` dependency if the bundled one is sufficient.
> 
> 3. **[minor] COS documentation uses the wrong environment provider class/env vars**
> 
> The docs show:
> 
> ```yaml
> fs.cosn.credentials.provider: org.apache.hadoop.fs.cosn.auth.EnvironmentVariableCredentialProvider
> ```
> 
> but the actual hadoop-cos class is `org.apache.hadoop.fs.cosn.auth.EnvironmentVariableCredentialsProvider` (plural `Credentials`). Also, the provider reads `COSN_SECRET_ID` and `COSN_SECRET_KEY`, while the docs say `COS_SECRET_ID` and `COS_SECRET_KEY`. Users following the docs would fail authentication.
> 
> 4. **[minor] Token lifetime in docs does not match the code**
> 
> The docs say the temporary credential is valid for 30 minutes, but `COSSecurityTokenProvider.DEFAULT_DURATION_SECONDS` is `3600L` (1 hour). Please align the docs or the code.
> 
> 5. **[nit] `cos.md` has trailing whitespace**
> 
> `git diff --check` reports trailing whitespace in `website/docs/maintenance/filesystems/cos.md` lines 14 and 17.
> 
> Validation I ran:
> 
> ```shell
> ./mvnw -pl fluss-filesystems/fluss-fs-cos -am -DskipTests compile
> ./mvnw -pl fluss-filesystems/fluss-fs-cos -am -DskipTests test-compile
> ./mvnw -pl fluss-filesystems/fluss-fs-cos -am -DskipTests package
> git diff --check
> ```
> 
> The Maven commands passed; `package` emitted the duplicate COS SDK warnings mentioned above, and `git diff --check` failed on the trailing whitespace.

Thanks for the thorough review! The PR has been updated in `80c5435` to address all five points:
1. **STS policy scoping** — `COSSecurityTokenProvider` now derives a default policy from `remote.data.dir`, restricting `name/cos:*` to the target bucket (and prefix). Users can override it via `fs.cosn.security.token.policy`.
2. **Duplicate COS SDK** — excluded `com.qcloud:cos_api-bundle` from `hadoop-cos` so only the directly declared `cos_api` ends up on the classpath. `NOTICE` updated accordingly. The shade duplicate-class warnings are gone.
3. **Docs credential provider** — fixed to `EnvironmentVariableCredentialsProvider` and the correct env vars `COSN_SECRET_ID` / `COSN_SECRET_KEY`.
4. **Token lifetime** — docs now say 1 hour to match `DEFAULT_DURATION_SECONDS = 3600L`.
5. **Trailing whitespace** — cleaned up in `website/docs/maintenance/filesystems/cos.md`; `git diff --check` is clean.
PTAL when you have time.

luoyuxia (approved):
+1

XuQianJin-Stars:
> +1

Thanks for the review! If there are no further concerns, let's go ahead and merge it.

leekeiabstraction (approved):
TY for the PR and addressing all the comments.

PR #2837, opened 2026-03-10
[log] Support dynamic configuration for log.replica.min-in-sync-replicas-number

### LINKED ISSUE #2831: [Feature] Support dynamic configuration for log.replica.min-in-sync-replicas-number
### Search before asking

### Description

Support dynamic configuration for log.replica.min-in-sync-replicas-number

### Willingness to contribute

zcoo:
### Purpose

Linked issue: close #2831

### DIFF
--- a/fluss-server/src/main/java/org/apache/fluss/server/DynamicServerConfig.java
+++ b/fluss-server/src/main/java/org/apache/fluss/server/DynamicServerConfig.java
@@ -43,8 +43,8 @@ import java.util.concurrent.locks.ReentrantReadWriteLock;
 
 import static org.apache.fluss.config.ConfigOptions.DATALAKE_FORMAT;
 import static org.apache.fluss.config.ConfigOptions.KV_SHARED_RATE_LIMITER_BYTES_PER_SEC;
-import static org.apache.fluss.config.ConfigOptions.LOG_REPLICA_MIN_IN_SYNC_REPLICAS_NUMBER;
 import static org.apache.fluss.config.ConfigOptions.KV_SNAPSHOT_INTERVAL;
+import static org.apache.fluss.config.ConfigOptions.LOG_REPLICA_MIN_IN_SYNC_REPLICAS_NUMBER;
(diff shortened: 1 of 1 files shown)


swuferhong:
@zcoo can you resolve the conflicts first?

zcoo:
@swuferhong ok~conflicts addressed

fluss-server/src/main/java/org/apache/fluss/server/replica/ReplicaManager.java
```diff
@@ -326,6 +332,44 @@ public DefaultSnapshotContext getKvSnapshotContext() {
...
+
+    // ============ ServerReconfigurable Implementation ============
+
+    @Override
+    public void validate(Configuration newConfig) throws ConfigException {
+        // Type validation is already handled by DynamicServerConfig.
+        // Here we only do basic sanity checks.
+        int newMinInSyncReplicas =
+                newConfig.get(ConfigOptions.LOG_REPLICA_MIN_IN_SYNC_REPLICAS_NUMBER);
+        if (newMinInSyncReplicas <= 0) {
```
platinumhamburg, line 347: What happens if newMinInSyncReplicas is greater than the replication factor (the number of replicas) configured for the table?

zcoo: I think it will cause problems. For "table.replication.factor" is a table-level config, technically we should make "ConfigOptions.LOG_REPLICA_MIN_IN_SYNC_REPLICAS_NUMBER" a table-level config too. But for simplicity
I just add a check newMinInSyncReplicas to be only 1 or 2, which covers most scenes.

swuferhong: After offline discussions, it was decided that the behavior of dynamic config should be consistent with the behavior of the default cluster configuration at cluster startup, without imposing an upper limit on the configuration.

PR #2842, opened 2026-03-11
[lake/iceberg] only enable column stats when scan filter exists

zuston:
### Purpose

Linked issue: close #xxx


beryllw:
Should we also check if this needs to be fixed?
https://github.com/apache/fluss/blob/dd181eb24ac4e536a50e593992a1d467823a50b9/fluss-lake/fluss-lake-iceberg/src/main/java/org/apache/fluss/lake/iceberg/maintenance/IcebergRewriteDataFiles.java#L91

luoyuxia:
> Should we also check if this needs to be fixed?
> 
> https://github.com/apache/fluss/blob/dd181eb24ac4e536a50e593992a1d467823a50b9/fluss-lake/fluss-lake-iceberg/src/main/java/org/apache/fluss/lake/iceberg/maintenance/IcebergRewriteDataFiles.java#L91

The code must keep includeColumnStats() because  sortFileScanTask() at line 122-138 directly reads f1.file().lowerBounds().get(sortFiledId) to sort files by the __offset column. Removing includeColumnStats() there would cause lowerBounds() to return null, resulting in an NPE

PR #2855, opened 2026-03-12
[kv] Support downloading CDC log segments from remote storage during KV table recovery

### LINKED ISSUE #2853: [Feature] Support downloading CDC log segments from remote storage during KV table recovery
### Search before asking

### Description

Currently, during KV table leader failover in a Fluss cluster, there is a potential scenario where, after a follower downloads the latest KV snapshot and attempts to recover by reading local logs, it discovers that the required
local log segment does not exist.

Although the root cause is currently unknown, Fluss lacks a robust mechanism to ensure recovery can proceed by fetching logs from remote storage when local log segments are missing. This mechanism is critical to guarantee that KV tables can recover successfully in all scenarios, preventing recovery failures due to missing local log segments.

### Willingness to contribute

swuferhong:
### Purpose

Linked issue: close #2853

### DIFF
--- a/fluss-server/src/main/java/org/apache/fluss/server/kv/RemoteLogFetcher.java
+++ b/fluss-server/src/main/java/org/apache/fluss/server/kv/RemoteLogFetcher.java
@@ -81,11 +81,12 @@ public class RemoteLogFetcher implements Closeable {
     private volatile RemoteLogBatchIterator activeIterator;
 
     public RemoteLogFetcher(
-            RemoteLogManager remoteLogManager, TableBucket tableBucket, File dataDir) {
+            RemoteLogManager remoteLogManager, TableBucket tableBucket, File logTabletDir) {
         this(
                 remoteLogManager,
                 tableBucket,
-                dataDir.toPath()
+                logTabletDir
+                        .toPath()
                         .resolve("tmp")
                         .resolve(REMOTE_LOG_RECOVERY_DIR_PREFIX + UUID.randomUUID()));
     }
--- a/fluss-server/src/test/java/org/apache/fluss/server/kv/RemoteLogFetcherTest.java
+++ b/fluss-server/src/test/java/org/apache/fluss/server/kv/RemoteLogFetcherTest.java
@@ -58,8 +58,8 @@ class RemoteLogFetcherTest extends RemoteLogTestBase {
         // Get the end offset of all remote segments to use as localLogStartOffset.
         long remoteEndOffset = segments.get(segments.size() - 1).remoteLogEndOffset();
 
-        File dataDir = logManager.getDataDir();
-        try (RemoteLogFetcher fetcher = new RemoteLogFetcher(remoteLogManager, tb, dataDir)) {
+        File logTabletDir = logTablet.getLogDir();
+        try (RemoteLogFetcher fetcher = new RemoteLogFetcher(remoteLogManager, tb, logTabletDir)) {
             // Fetch all remote log batches from offset 0 to remoteEndOffset.
             Iterable<LogRecordBatch> batches = fetcher.fetch(0, remoteEndOffset);
             List<LogRecordBatch> batchList = new ArrayList<>();
@@ -96,8 +96,8 @@ class RemoteLogFetcherTest extends RemoteLogTestBase {
         long startOffset = 5;
         long remoteEndOffset = segments.get(segments.size() - 1).remoteLogEndOffset();
 
-        File dataDir = logManager.getDataDir();
-        try (RemoteLogFetcher fetcher = new RemoteLogFetcher(remoteLogManager, tb, dataDir)) {
+        File logTabletDir = logTablet.getLogDir();
+        try (RemoteLogFetcher fetcher = new RemoteLogFetcher(remoteLogManager, tb, logTabletDir)) {
             Iterable<LogRecordBatch> batches = fetcher.fetch(startOffset, remoteEndOffset);
             List<LogRecordBatch> batchList = new ArrayList<>();
             for (LogRecordBatch batch : batches) {
@@ -128,8 +128,8 @@ class RemoteLogFetcherTest extends RemoteLogTestBase {
         // Use the start offset of the second segment as localLogStartOffset.
         long localLogStartOffset = segments.get(1).remoteLogStartOffset();
 
-        File dataDir = logManager.getDataDir();
-        try (RemoteLogFetcher fetcher = new RemoteLogFetcher(remoteLogManager, tb, dataDir)) {
+        File logTabletDir = logTablet.getLogDir();
+        try (RemoteLogFetcher fetcher = new RemoteLogFetcher(remoteLogManager, tb, logTabletDir)) {
             Iterable<LogRecordBatch> batches = fetcher.fetch(0, localLogStartOffset);
             List<LogRecordBatch> batchList = new ArrayList<>();
             for (LogRecordBatch batch : batches) {
@@ -151,8 +151,9 @@ class RemoteLogFetcherTest extends RemoteLogTestBase {
         makeLogTableAsLeader(tb, false);
         // Don't upload any segments to remote.
 
-        File dataDir = logManager.getDataDir();
-        try (RemoteLogFetcher fetcher = new RemoteLogFetcher(remoteLogManager, tb, dataDir)) {
+        Replica replica = replicaManager.getReplicaOrException(tb);
+        File logTabletDir = replica.getLogTablet().getLogDir();
+        try (RemoteLogFetcher fetcher = new RemoteLogFetcher(remoteLogManager, tb, logTabletDir)) {
             assertThatThrownBy(() -> fetcher.fetch(0, 100))
                     .isInstanceOf(RemoteStorageException.class)
                     .hasMessageContaining("No remote log segments found");
@@ -161,8 +162,12 @@ class RemoteLogFetcherTest extends RemoteLogTestBase {
 
     @Test
     void testStaleDirectoriesFromUncleanShutdownAreCleanedUp() throws Exception {
-        File dataDir = logManager.getDataDir();
-        Path tmpDir = dataDir.toPath().resolve("tmp");
+        TableBucket tb = new TableBucket(DATA1_TABLE_ID, 0);
+        makeLogTableAsLeader(tb, false);
+        Replica replica = replicaManager.getReplicaOrException(tb);
+        File logTabletDir = replica.getLogTablet().getLogDir();
+
+        Path tmpDir = logTabletDir.toPath().resolve("tmp");
         Files.createDirectories(tmpDir);
 
         // Simulate stale directories left behind by a previous unclean shutdown.
@@ -176,8 +181,7 @@ class RemoteLogFetcherTest extends RemoteLogTestBase {
 
         // Simulate next startup: creating and closing a new RemoteLogFetcher should
         // clean up the entire tmp directory including all stale recovery directories.
-        TableBucket tb = new TableBucket(DATA1_TABLE_ID, 0);
-        RemoteLogFetcher fetcher = new RemoteLogFetcher(remoteLogManager, tb, dataDir);
+        RemoteLogFetcher fetcher = new RemoteLogFetcher(remoteLogManager, tb, logTabletDir);
         fetcher.close();
 
         assertThat(Files.exists(staleDir1)).isFalse();
@@ -199,8 +203,8 @@ class RemoteLogFetcherTest extends RemoteLogTestBase {
         assertThat(segments).isNotEmpty();
         long remoteEndOffset = segments.get(segments.size() - 1).remoteLogEndOffset();
 
-        File dataDir = logManager.getDataDir();
-        RemoteLogFetcher fetcher = new RemoteLogFetcher(remoteLogManager, tb, dataDir);
+        File logTabletDir = logTablet.getLogDir();
+        RemoteLogFetcher fetcher = new RemoteLogFetcher(remoteLogManager, tb, logTabletDir);
         Path tempDir = fetcher.getTempDir();
 
         // Not yet created before fetch.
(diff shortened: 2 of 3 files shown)


fluss-server/src/main/java/org/apache/fluss/server/kv/RemoteLogFetcher.java
```diff
@@ -0,0 +1,322 @@
...
+        LOG.info(
+                "Found {} remote log segments for table bucket {} from offset {} to local log start offset {}",
+                segments.size(),
+                tableBucket,
+                startOffset,
+                localLogStartOffset);
+
+        RemoteLogBatchIterator iterator =
+                new RemoteLogBatchIterator(segments, startOffset, localLogStartOffset);
+        this.activeIterator = iterator;
```
platinumhamburg, line 147: If `fetch()` is called more than once, the previous `RemoteLogBatchIterator` (and its open `FileLogRecords`) gets silently abandoned — file descriptors leak. I know the current call site only calls it once, but since the class implements `Closeable` and manages lifecycle, the contract should be self-consistent.

Simple fix — close the old one first:

```java
public Iterable<LogRecordBatch> fetch(long startOffset, long localLogStartOffset)
        throws Exception {
    // Close any previously active iterator before creating a new one
    RemoteLogBatchIterator prev = this.activeIterator;
    if (prev != null) {
        prev.close();
    }

    List<RemoteLogSegment> segments =
            remoteLogManager.relevantRemoteLogSegments(tableBucket, startOffset);
    // ... rest unchanged
```

fluss-server/src/main/java/org/apache/fluss/server/kv/RemoteLogFetcher.java
```diff
@@ -0,0 +1,322 @@
...
+
+                // move to next segment
+                if (currentSegmentIndex >= segments.size()) {
+                    finished = true;
+                    return;
+                }
+
+                RemoteLogSegment segment = segments.get(currentSegmentIndex++);
+                // skip segments entirely before currentOffset
+                if (segment.remoteLogEndOffset() <= currentOffset) {
```
platinumhamburg: `RemoteLogSegment.remoteLogEndOffset()` is documented as inclusive ("Remote log end offset of this segment (inclusive)"), but the skip check here uses `<=`:

```java
if (segment.remoteLogEndOffset() <= currentOffset) {
    continue;
}
```

When `currentOffset` equals the segment's last record offset, the segment gets incorrectly skipped — that record is lost. During KV recovery this could cause data inconsistency.

Should be strict less-than if the end offset is truly inclusive:

```java
// remoteLogEndOffset() is inclusive — last offset in segment
if (segment.remoteLogEndOffset() < currentOffset) {
    continue;
}
```

Or if it's actually exclusive, a clarifying comment would help since the Javadoc says otherwise.

fluss-server/src/test/java/org/apache/fluss/server/kv/RemoteLogFetcherTest.java
```diff
@@ -0,0 +1,246 @@
...
+import java.nio.file.Path;
+import java.util.ArrayList;
+import java.util.List;
+
+import static org.apache.fluss.record.TestData.DATA1_TABLE_ID;
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
+
+/** Test for {@link RemoteLogFetcher}. */
+class RemoteLogFetcherTest extends RemoteLogTestBase {
```
platinumhamburg, line 41: There's no test for the case where `startOffset == localLogStartOffset`. If the comparison in `advance()` were accidentally changed from `>=` to `>`, the fetcher would return batches beyond the intended range — which during KV recovery means applying duplicate records.

Would be great to add something like:

```java
@Test
void testFetchWithEmptyRange() throws Exception {
    // ... setup ...
    long sameOffset = segments.get(0).remoteLogStartOffset();
    try (RemoteLogFetcher fetcher = new RemoteLogFetcher(remoteLogManager, tb, dataDir)) {
        Iterable<LogRecordBatch> batches = fetcher.fetch(sameOffset, sameOffset);
        List<LogRecordBatch> batchList = new ArrayList<>();
        for (LogRecordBatch batch : batches) {
            batchList.add(batch);
        }
        assertThat(batchList).isEmpty();
    }

File: fluss-server/src/main/java/org/apache/fluss/server/kv/RemoteLogFetcher.java [hunk 4]
```diff
@@ -0,0 +1,322 @@
...
+
+    @Override
+    public void close() {
+        // Close any active iterator to release file handles
+        if (activeIterator != null) {
+            activeIterator.close();
+            activeIterator = null;
+        }
+        LOG.info("Cleaning up remote log recovery temp dir: {}", tempDir);
+        deleteDirectoryQuietly(tempDir.toFile());
```
platinumhamburg: Each `downloadSegment()` writes a file to `tempDir`, and they all stay on disk until `close()` deletes the entire directory. For a recovery with many remote segments, this could eat up disk space unnecessarily. Deleting each segment file after the iterator moves past it would bound disk usage to one segment at a time.

Not blocking — the current approach is simple and correct for bounded recovery scenarios. Just something to keep in mind if segment counts grow large.

fluss-server/src/main/java/org/apache/fluss/server/kv/RemoteLogFetcher.java
```diff
@@ -0,0 +1,322 @@
...
+            if (nextBatch == null && !finished) {
+                advance();
+            }
+            return nextBatch != null;
+        }
+
+        @Override
+        public LogRecordBatch next() {
+            if (!hasNext()) {
+                throw new java.util.NoSuchElementException();
```
platinumhamburg: Nit: `java.util.NoSuchElementException` is used inline instead of being imported. The rest of the codebase consistently imports it.

```java
// Add to imports:
import java.util.NoSuchElementException;

// In next():
throw new NoSuchElementException();
```

fluss-server/src/main/java/org/apache/fluss/server/kv/RemoteLogFetcher.java
swuferhong: Ok.

swuferhong:
@platinumhamburg comments addressed.

LiebingYu:
LGTM.

fluss-server/src/main/java/org/apache/fluss/server/kv/KvRecoverHelper.java
```diff
@@ -87,6 +88,30 @@ public KvRecoverHelper(
...
+                kvTablet,
+                logTablet,
+                recoverPointOffset,
+                recoverPointRowCount,
+                autoIncRange,
+                recoverContext,
+                kvFormat,
+                logFormat,
+                schemaGetter,
+                null);
```
wuchong: Unnecessary new constructor.

fluss-server/src/main/java/org/apache/fluss/server/kv/RemoteLogFetcher.java
```diff
@@ -0,0 +1,333 @@
...
+                    continue;
+                }
+                // skip segments that start at or after localLogStartOffset
+                if (segment.remoteLogStartOffset() >= localLogStartOffset) {
+                    finished = true;
+                    return;
+                }
+
+                try {
+                    File localFile = downloadSegment(segment);
```
wuchong, line 304: Please add a TODO and create an issue to track the potential
optimization of async downloading the next segment while processing the current one.
Otherwise, the recovery process may be significantly slowed down by the
download time of remote segments, especially when there are many segments to
fetch.

fluss-server/src/test/java/org/apache/fluss/server/kv/RemoteLogFetcherTest.java
```diff
@@ -0,0 +1,277 @@
...
+        File dataDir = logManager.getDataDir();
+        try (RemoteLogFetcher fetcher = new RemoteLogFetcher(remoteLogManager, tb, dataDir)) {
+            assertThatThrownBy(() -> fetcher.fetch(0, 100))
+                    .isInstanceOf(RemoteStorageException.class)
+                    .hasMessageContaining("No remote log segments found");
+        }
+    }
+
+    @Test
+    void testTempDirectoryCreatedAndCleanedUp() throws Exception {
```
wuchong: Add a test case simulating an unclean shutdown that leaves temporary directories behind, and ensure they are cleaned up upon the next startup.

fluss-server/src/test/java/org/apache/fluss/server/kv/RemoteLogFetcherTest.java
```diff
@@ -0,0 +1,277 @@
...
+        RemoteLogFetcher fetcher = new RemoteLogFetcher(remoteLogManager, tb, dataDir);
+        Path tempDir = fetcher.getTempDir();
+
+        // Verify temp directory was created.
+        assertThat(Files.exists(tempDir)).isTrue();
+        assertThat(tempDir.getFileName().toString()).startsWith("remote-log-recovery-");
+
+        // Close and verify cleanup.
+        fetcher.close();
+        assertThat(Files.exists(tempDir)).isFalse();
```
wuchong, line 221: Additionally, verify that the temporary parent directory is removed.

fluss-server/src/main/java/org/apache/fluss/server/kv/RemoteLogFetcher.java
```diff
@@ -0,0 +1,333 @@
...
+    /** Tracks the currently active iterator to ensure proper cleanup on close. */
+    private volatile RemoteLogBatchIterator activeIterator;
+
+    public RemoteLogFetcher(
+            RemoteLogManager remoteLogManager, TableBucket tableBucket, File dataDir)
+            throws IOException {
+        this(
+                remoteLogManager,
+                tableBucket,
+                Files.createDirectories(
```
wuchong: The `this()` constructor creates the directory again.

fluss-server/src/main/java/org/apache/fluss/server/kv/RemoteLogFetcher.java
```diff
@@ -0,0 +1,333 @@
...
+                                .resolve(REMOTE_LOG_RECOVERY_DIR_PREFIX + UUID.randomUUID())));
+    }
+
+    @VisibleForTesting
+    RemoteLogFetcher(RemoteLogManager remoteLogManager, TableBucket tableBucket, Path tempDir)
+            throws IOException {
+        this.remoteLogManager = remoteLogManager;
+        this.tableBucket = tableBucket;
+        this.tempDir = tempDir;
+        Files.createDirectories(tempDir);
```
wuchong: I think we can move the tmp dir creation into `fetch()` to lazily crete the directory, so that we can initialize a not-null `RemoteLogFetcher` for `KvRecoverHelper`. Currently, the `remoteLogFetcher` usage in `KvRecoverHelper` is not very clean (what happens when `nextLogOffset < localLogStartOffset` but `remoteLogFetcher` is null? There is no guarantees from the perspective of `KvRecoverHelper`.) This can also simplify the logic of `Replica` a lot.

fluss-server/src/main/java/org/apache/fluss/server/kv/RemoteLogFetcher.java
```diff
@@ -0,0 +1,333 @@
...
+
+    @Override
+    public void close() {
+        // Close any active iterator to release file handles
+        if (activeIterator != null) {
+            activeIterator.close();
+            activeIterator = null;
+        }
+        LOG.info("Cleaning up remote log recovery temp dir: {}", tempDir);
+        deleteDirectoryQuietly(tempDir.toFile());
```
wuchong: I suggest to remove the `tmp` dir as well, there maybe some remaining dirty files if the server is fail during last recovering.

fluss-server/src/main/java/org/apache/fluss/server/kv/KvRecoverHelper.java
```diff
@@ -188,6 +244,41 @@ public void recover() throws Exception {
...
+                                    } else {
+                                        kvBatchWriter.put(resumeRecord.key, resumeRecord.value);
+                                    }
+                                });
+            }
+            return nextFetchOffset;
+        }
+    }
+
     private long readLogRecordsAndApply(
```
wuchong, line 199: I think there is duplicated between `readLogRecordsAndApply` and `readRemoteLogRecordsAndApply`, espcially the `resumeRecordConsumer` logic. I think we can merge them by providing different `batches`. In this way, we can also share the `KvBatchWriter` and `LogRecordReadContext` resources. 


```java
private long readLogRecordsAndApply(
            long startFetchOffset,
            long localLogStartOffset,
            RowCountUpdater rowCountUpdater,
            AutoIncIDRangeUpdater autoIncIdRangeUpdater,
            FetchIsolation fetchIsolation,
            ThrowingConsumer<KeyValueAndLogOffset, Exception> resumeRecordConsumer)
            throws Exception {
        try (LogRecordReadContext readContext = createLogRecordReadContext()) {
            long nextFetchOffset = startFetchOffset;
            while (true) {
                final Iterable<LogRecordBatch> batches;
                if (nextFetchOffset < localLogStartOffset) {
                    batches = remoteLogFetcher.fetch(nextFetchOffset, localLogStartOffset);
                } else {
                    LogRecords logRecords =
                            logTablet
                                    .read(
                                            nextFetchOffset,
                                            recoverContext.maxFetchLogSizeInRecoverKv,
                                            fetchIsolation,
                                            true,
                                            null,
                                            null)
                                    .getRecords();
                    if (logRecords == MemoryLogRecords.EMPTY) {
                        break;
                    }
                    batches = logRecords.batches();
                }

                for (LogRecordBatch logRecordBatch : batches) {
                    nextFetchOffset =
                            applyLogRecordBatch(
                                    logRecordBatch,
                                    readContext,
                                    rowCountUpdater,
                                    autoIncIdRangeUpdater,
                                    resumeRecordConsumer);
                }
            }
            return nextFetchOffset;
        }
    }
```

fluss-server/src/test/java/org/apache/fluss/server/kv/RemoteLogFetcherTest.java
swuferhong: It's not easy to simulate an unclean shutdown in an ITCase, so I only added a unit test to verify that stale directories are cleaned up after recovery.

swuferhong:
@wuchong comments addressed.

PR #2857, opened 2026-03-12
[server] Register ConfigException in Errors to fix misleading UnknownServerException from alterClusterConfigs

### LINKED ISSUE #2856: [server] Admin.alterClusterConfigs() returns UnknownServerException instead of ConfigException on validation failure
### Search before asking

### Description

**Task Overview:**

`Admin.alterClusterConfigs()` returns a misleading `UnknownServerException` wrapping the actual `ConfigException` when a config validation failure occurs on the server side. Callers of this API (e.g., `sys.set_cluster_configs` Flink procedure, Java Admin client) receive an opaque error message like:

```
org.apache.fluss.exception.UnknownServerException: org.apache.fluss.exception.ConfigException: The config key <key> is not allowed to be changed dynamically.
```

instead of a direct, actionable `ConfigException` with the original message.

**Purpose:**

Improve error transparency of `Admin.alterClusterConfigs()` so that callers can clearly identify configuration validation failures without having to unwrap a generic `UnknownServerException`.

**Root Cause:**

`ConfigException` is a subclass of `ApiException` but is **not registered** in `Errors.java` (`CLASS_TO_ERROR` map). Only `InvalidConfigException` (error code 39) is mapped. The call chain is:

1. `DynamicConfigManager.prepareIncrementalConfigs()` throws `ConfigException` (e.g., key not allowed for dynamic change, or invalid value type)
2. `CoordinatorService.alterClusterConfigs()` propagates it via `future.completeExceptionally(e)` inside an `AccessContextEvent`
3. The RPC layer calls `Errors.forException(ConfigException)` — no match found in `CLASS_TO_ERROR` — falls back to `UNKNOWN_SERVER_ERROR`
4. Client receives `UnknownServerException` wrapping the original `ConfigException`

### Willingness to contribute

LiebingYu:
### Purpose

Linked issue: close #2856

### DIFF
--- a/fluss-client/src/test/java/org/apache/fluss/client/admin/FlussAdminITCase.java
+++ b/fluss-client/src/test/java/org/apache/fluss/client/admin/FlussAdminITCase.java
@@ -32,6 +32,7 @@ import org.apache.fluss.config.Configuration;
 import org.apache.fluss.config.cluster.AlterConfig;
 import org.apache.fluss.config.cluster.AlterConfigOpType;
 import org.apache.fluss.config.cluster.ConfigEntry;
+import org.apache.fluss.exception.ConfigException;
 import org.apache.fluss.exception.DatabaseAlreadyExistException;
 import org.apache.fluss.exception.DatabaseNotEmptyException;
 import org.apache.fluss.exception.DatabaseNotExistException;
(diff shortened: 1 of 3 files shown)


LiebingYu:
Hi @loserwang1024, PTAL. Thanks!

PR #2859, opened 2026-03-13
[lake/tiering] fallback to Flink's `io.tmp.dir` when `client.scanner.io.tmpdir` is not set

zuston:
### Purpose

To avoid downloading remote logs into the `/tmp` directory when tiering Flink jobs run on YARN, which could cause uncontrolled disk usage.


zuston:
PTAL @luoyuxia @wuchong

Issue #2865, opened 2026-03-13
[Helm] Restructure helm chart of Fluss


nhuantho:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/fluss/issues) and found nothing similar.

### Motivation

Inspired by the Helm chart structure used in the Apache Airflow project  
(https://github.com/apache/airflow/tree/main/chart), this proposal aims to improve the Fluss Helm chart to make it **cleaner, more structured, and easier to maintain**.

### values.schema.json
Uses the **http://json-schema.org/draft-07/schema** structure.  
This file defines the schema of the Helm chart values and is used to validate `values.yaml`.  
***It can also be used as the source for **auto-generating parameter documentation** for the Helm chart -> Break another issue.***

### values.schema.meta.json
This meta-schema is used to validate `values.schema.json`.  
It ensures that:

- Each parameter includes both **`default`** and **`description`**
- All top-level properties define **`x-docsSection`** to group parameters in the generated documentation

---

## Helm Values Naming Improvements

Based on the architecture concepts described in the Fluss documentation  
(https://fluss.apache.org/docs/concepts/architecture/), some naming updates are proposed to better align with Kubernetes conventions.

### Proposed Changes

| Current Field | Proposed Field | Reason |
|---|---|---|
| `listeners` | `ports` | Represents container or service ports in Kubernetes |
| `storage` | `persistence` | Aligns with standard Helm and Kubernetes terminology for persistent storage |

### New structure value example:
```
coordinatorServer:
  replicas: 1

  livenessProbe:
    failureThreshold: 100
    timeoutSeconds: 1
    initialDelaySeconds: 10
    periodSeconds: 3

  readinessProbe:
    failureThreshold: 100
    timeoutSeconds: 1
    initialDelaySeconds: 10
    periodSeconds: 3

  resources: {}
  #   limits:
  #     cpu: 100m
  #     memory: 128Mi
  #   requests:
  #     cpu: 100m
  #     memory: 128Mi

  persistence:
    enabled: false
    size: 1Gi
    storageClass:

tabletServer:
  replicas: 3

  livenessProbe:
    failureThreshold: 100
    timeoutSeconds: 1
    initialDelaySeconds: 10
    periodSeconds: 3

  readinessProbe:
    failureThreshold: 100
    timeoutSeconds: 1
    initialDelaySeconds: 10
    periodSeconds: 3

  resources: {}
  #   requests:
  #     cpu: 100m
  #     memory: 128Mi
[...]

PR #2472, opened 2026-01-25
[Helm] Chart Component Configuration Isolation

hemanthsavasere:
### Purpose

Linked issue: close #2191

This PR introduces component-specific configuration sections for the Fluss Helm chart, enabling users to independently configure replicas and health probes for coordinator and tablet servers.

###Previously,
1. Hardcoded Replicas: Coordinator (1 replica) and Tablet (3 replicas) counts were hardcoded
2. Hardcoded Health Probes: All liveness and readiness probe settings were hardcoded with identical values for both components
3. Resource Configuration Bug: Coordinator StatefulSet incorrectly referenced `.Values.resources.tabletServer` instead of `.Values.resources.coordinatorServer`

Changes

1. helm/values.yaml
Added two new top-level configuration sections:

```yaml
coordinatorServer:
  replicas: 1
  livenessProbe:
    failureThreshold: 100
    timeoutSeconds: 1
    initialDelaySeconds: 10
    periodSeconds: 3
  readinessProbe:
    failureThreshold: 100
    timeoutSeconds: 1
    initialDelaySeconds: 10
    periodSeconds: 3

tabletServer:
  replicas: 3
  livenessProbe:
    failureThreshold: 100
    timeoutSeconds: 1
    initialDelaySeconds: 10
    periodSeconds: 3
  readinessProbe:
    failureThreshold: 100
    timeoutSeconds: 1
    initialDelaySeconds: 10
    periodSeconds: 3
```

Also updated the `resources` section to use explicit empty defaults instead of being commented out:
```yaml
resources:
  coordinatorServer: {}
  tabletServer: {}
```

2. helm/templates/sts-coordinator.yaml
- Replaced hardcoded `replicas: 1` with `{{ .Values.coordinatorServer.replicas }}`
- Replaced hardcoded liveness probe values with `.Values.coordinatorServer.livenessProbe.*` references
- Replaced hardcoded readiness probe values with `.Values.coordinatorServer.readinessProbe.*` references
- **Fixed bug**: Changed resource reference from `.Values.resources.tabletServer` to `.Values.resources.coordinatorServer`

3. helm/templates/sts-tablet.yaml
- Replaced hardcoded `replicas: 3` with `{{ .Values.tabletServer.replicas }}`
- Replaced hardcoded liveness probe values with `.Values.tabletServer.livenessProbe.*` references
- Replaced hardcoded readiness probe values with `.Values.tabletServer.readinessProbe.*` references

### Tests

Below tests are successful

[...]

PR #2871, opened 2026-03-14
[helm] Restructure helm chart of Fluss

nhuantho:
### Purpose

Linked issue: close https://github.com/apache/fluss/issues/2865

### Brief change log

* Restructured `values.yaml`
  * coordinator -> coordinatorServer (https://fluss.apache.org/docs/concepts/architecture/).
  * tablet -> tabletServer (https://fluss.apache.org/docs/concepts/architecture/).
  * Move resource of coordinatorServer and tabletServer to inside themselves.
  * listeners -> ports.
  * storage -> persistence.
* Created `values.schema.json`:  schema of `values.yaml` and standard to valid `value.yaml`.
* Created `values.schema.schema.json`: schema of `values.schema.json` and standard to valid `values.schema.json`.
* Created `tools/ci/helm_ci/validate_helm.py` to validate helm.
* Add ci to validate helm.
* Add ignore python in `.gitignore`

### DIFF
--- a/tools/ci/helm_ci/requirements.txt
+++ b/tools/ci/helm_ci/requirements.txt
@@ -1,2 +1,26 @@
+#  Licensed to the Apache Software Foundation (ASF) under one
+#  or more contributor license agreements.  See the NOTICE file
+#  distributed with this work for additional information
+#  regarding copyright ownership.  The ASF licenses this file
+#  to you under the Apache License, Version 2.0 (the
+#  "License"); you may not use this file except in compliance
+#  with the License.  You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+#  Unless required by applicable law or agreed to in writing, software
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+#  Unless required by applicable law or agreed to in writing, software
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+
 jsonschema==4.26.0
 PyYAML==6.0.3


nhuantho:
@affo Can you review for me?

nhuantho:
@swuferhong can you review for me?

.github/workflows/helm-chart.yaml
```diff
@@ -44,6 +44,17 @@ jobs:
       - name: "Checkout code"
         uses: actions/checkout@v6
 
+      - name: Set up Python
+        uses: actions/setup-python@v5
+        with:
+          python-version: "3.11"
```
affo, line 50:
```suggestion
          python-version: "3.12"
```

Just better to point to that one if we don't break anything

.github/workflows/helm-chart.yaml
```diff
@@ -44,6 +44,17 @@ jobs:
...
+
+      - name: Install dependencies
+        run: pip install jsonschema==4.26.0 PyYAML==6.0.3
```
affo: I would favor a `requirements.txt` under `tools/ci/helm_ci`.
It would make it easier to validate locally.

The command would become `pip install -r tools/ci/helm_ci/requirements.txt`

.github/workflows/helm-chart.yaml
```diff
@@ -44,6 +44,17 @@ jobs:
...
+
+      - name: Run unstable test reporter
```
affo:
```suggestion
      - name: Validate values
```

tools/ci/helm_ci/validate_helm.py
```diff
@@ -0,0 +1,89 @@
+#  Licensed to the Apache Software Foundation (ASF) under one
```
affo, line 1: Why do we need this script ?

Isn't `helm lint` already picking `values.schema.json` for validation?
Is this only for the meta schema?

affo:
@nhuantho by the way, thanks for the contribution and to light the spotlight on this defect 🤝

affo:
@nhuantho as a side note, one can already structure schema annotations by adding comments prefixed with `--` to pair with `helm-docs` which is a tool we may add for generating the docs https://github.com/norwoodj/helm-docs

tools/ci/helm_ci/validate_helm.py
nhuantho: @affo, from my side, when everyone updated helm. Could they please explain why it needs to be updated and describe the changes in the schema?
For example:
image

nhuantho:
@affo, I will answer some confusion:
* I think it will confuse users who use Helm 
```
# @schema
  # type: object
  # additionalProperties: false
  # @schema
```
* https://github.com/dadav/helm-schema is not fully open source, and it is easy for us to build unit tests for Helm.
* https://github.com/norwoodj/helm-docs is also not fully open source, and I think we can easily build and generate docs.

Focus pr:
* We have the first verified value change to avoid affecting the helm -> https://github.com/apache/fluss/pull/2472
* Suggest defining value with a standard (airflow).
* Build a schema and describe all values of helm, and the first step for auto-generating helm docs.

.github/workflows/helm-chart.yaml
nhuantho: I want to unify the Python version with `.github/workflows/unstable-test-reporter.yaml`

Issue #2182, opened 2025-12-16
[lake/paimon] Support snapshot expiration for lake tables


LiebingYu:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/fluss/issues) and found nothing similar.

### Motivation

Currently, Fluss tiering service won't do snapshot expiration for Paimon lake table, even though `'write-only' = 'false'` is set.

When a Paimon table is set with `'write-only' = 'false'`, it indicates that the engine is expected to perform snapshot expiration. In my opinion, Fluss's tiering service should be consistent with this behavior—if users do not want snapshot expiration, they can set `'write-only' = 'true'`.

### Willingness to contribute

- [x] I'm willing to submit a PR!

PR #2881, opened 2026-03-16
[lake/iceberg] Support snapshot expiration for lake tables

### LINKED ISSUE #2213: [lake/iceberg] Support snapshot expiration for lake tables
### Search before asking

### Motivation

https://github.com/apache/fluss/pull/2184 introduce a new table config `table.datalake.auto-expire-snapshot` and a new tiering service config `lake.tiering.auto-expire-snapshot` to support user to controll the snapshot expiration for lake table. Iceberg should also suport these configs.

### Solution

_No response_

### Anything else?

_No response_

### Willingness to contribute

vamossagar12:
### Purpose

Linked issue: close #2213 

### Brief change log

Similar to https://github.com/apache/fluss/issues/2182, this PR adds support for snapshot expiration for Iceberg tables.

### Tests

fluss-lake/fluss-lake-iceberg/src/test/java/org/apache/fluss/lake/iceberg/tiering/IcebergTieringTest.java

### API and Format

N/A

### Documentation

The configs `table.datalake.auto-expire-snapshot`  and  `tiering service config lake.tiering.auto-expire-snapshot` already exist. This PR makes Iceberg implement it.

Added 2 new configs: `lake.iceberg.expire-snapshot.min-interval-commits` and `lake.iceberg.expire-snapshot.min-interval-ms` to throttle snapshotting.

### DIFF
--- a/fluss-lake/fluss-lake-iceberg/src/main/java/org/apache/fluss/lake/iceberg/tiering/IcebergLakeCommitter.java
+++ b/fluss-lake/fluss-lake-iceberg/src/main/java/org/apache/fluss/lake/iceberg/tiering/IcebergLakeCommitter.java
@@ -17,7 +17,9 @@
 
 package org.apache.fluss.lake.iceberg.tiering;
 
+import org.apache.fluss.config.ConfigOptions;
 import org.apache.fluss.lake.committer.CommittedLakeSnapshot;
+import org.apache.fluss.lake.committer.CommitterInitContext;
 import org.apache.fluss.lake.committer.LakeCommitResult;
 import org.apache.fluss.lake.committer.LakeCommitter;
 import org.apache.fluss.lake.iceberg.maintenance.RewriteDataFileResult;
@@ -28,6 +30,7 @@ import org.apache.iceberg.CatalogUtil;
 import org.apache.iceberg.ContentFile;
 import org.apache.iceberg.DataFile;
 import org.apache.iceberg.DeleteFile;
+import org.apache.iceberg.ExpireSnapshots;
 import org.apache.iceberg.RewriteFiles;
 import org.apache.iceberg.RowDelta;
 import org.apache.iceberg.Snapshot;
@@ -61,12 +64,23 @@ public class IcebergLakeCommitter implements LakeCommitter<IcebergWriteResult, I
 
     private final Catalog icebergCatalog;
     private final Table icebergTable;
+    private final boolean isAutoSnapshotExpiration;
     private static final ThreadLocal<Long> currentCommitSnapshotId = new ThreadLocal<>();
 
-    public IcebergLakeCommitter(IcebergCatalogProvider icebergCatalogProvider, TablePath tablePath)
+    public IcebergLakeCommitter(
+            IcebergCatalogProvider icebergCatalogProvider,
+            CommitterInitContext committerInitContext)
             throws IOException {
         this.icebergCatalog = icebergCatalogProvider.get();
-        this.icebergTable = getTable(tablePath);
+        this.icebergTable = getTable(committerInitContext.tablePath());
+        this.isAutoSnapshotExpiration =
+                committerInitContext
+                                .tableInfo()
+                                .getTableConfig()
+                                .isDataLakeAutoExpireSnapshot()
+                        || committerInitContext
+                                .lakeTieringConfig()
+                                .get(ConfigOptions.LAKE_TIERING_AUTO_EXPIRE_SNAPSHOT);
         // register iceberg listener
         Listeners.register(new IcebergSnapshotCreateListener(), CreateSnapshotEvent.class);
     }
@@ -141,6 +155,12 @@ public class IcebergLakeCommitter implements LakeCommitter<IcebergWriteResult, I
                     snapshotId = rewriteCommitSnapshotId;
                 }
             }
+
+            // Expire old snapshots if auto-expire-snapshot is enabled
+            if (isAutoSnapshotExpiration) {
+                expireSnapshots();
+            }
+
             // Iceberg does not provide cumulative table stats API yet; leave stats as -1 (unknown).
             return LakeCommitResult.committedIsReadable(snapshotId);
         } catch (Exception e) {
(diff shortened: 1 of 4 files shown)


vamossagar12:
@luoyuxia please review this PR. Thanks!

vamossagar12:
@luoyuxia gentle ping!

fluss-lake/fluss-lake-iceberg/src/main/java/org/apache/fluss/lake/iceberg/tiering/IcebergLakeCommitter.java
```diff
@@ -262,6 +279,22 @@ public void close() throws Exception {
...
 
+    /**
+     * Expires old snapshots from the Iceberg table. Uses Iceberg's built-in snapshot expiration
+     * which respects table properties like {@code history.expire.max-snapshot-age-ms} and {@code
+     * history.expire.min-snapshots-to-keep}.
+     */
+    private void expireSnapshots() {
+        try {
+            ExpireSnapshots expireSnapshots =
+                    icebergTable.expireSnapshots().cleanExpiredFiles(true);
```
leekeiabstraction, line 345: Does this expire all non latest snapshots? If so, would this cause recovery failure in the case when tiering service failover and restarts from latest Flink checkpoint and couldn't find a deleted iceberg snapshot?

leekeiabstraction (commented):
TY for the PR, left a comment. PTAL

fluss-lake/fluss-lake-iceberg/src/main/java/org/apache/fluss/lake/iceberg/tiering/IcebergLakeCommitter.java
vamossagar12: You're right — with aggressive expiration settings (e.g., `min-snapshots-to-keep`=1 and `max-snapshot-age-ms`=1ms), the expired snapshot could cause `getMissingLakeSnapshot()` to fail during recovery because `icebergTable.snapshot(expiredId)` returns null.

I've fixed this by making `getMissingLakeSnapshot()` handle expired snapshots gracefully, similar to how Paimon's `SnapshotManager.pickOrLatest()` works. Instead of throwing IllegalStateException, the code now logs a warning and falls through to return the latest Fluss-committed live snapshot for recovery. Since Iceberg never expires the current snapshot, the latest Fluss-committed snapshot is guaranteed to be newer than the expired one.

Added test `testSnapshotExpirationDoesNotBreakRecovery` to validate this scenario. Let me know what you think!

vamossagar12:
One of the failing tests `CoordinatorEventManagerTest.testMetricsUpdatedImmediatelyOnStartup:95 ` seems to be flaky. Related bug: #3188 .  I created #3210 to fix it.

vamossagar12:
@leekeiabstraction please take a look when you get a chance.

polyzos:
@vamossagar12 can you rebase? I think this PR is good, but @MehulBatra  @luoyuxia can you also PTAL?

vamossagar12:
@polyzos thanks, I rebased and pushed. @MehulBatra , @luoyuxia can you also take a look. Thanks!

polyzos:
@vamossagar12 Thanks for this.. I have one concen that i would like your thoughts on as well us from @luoyuxia and @MehulBatra ... Unlike Paimon, where setting `WRITE_ONLY=false` lets expiration run in-band as part of the normal commit, Iceberg's `ExpireSnapshots.commit()` is a **separate metadata commit** that:

- Plans across manifests
- Writes a new `TableMetadata` JSON
- Synchronously deletes files (because `cleanExpiredFiles=true`)

In a streaming tiering scenario, this happens on **every commit** with no throttling. This can materially increase per-commit latency and produce a metadata/delete storm on object stores. Could we:

- Add an opt-in interval (e.g. `lake.iceberg.expire-snapshot.interval` defaulting to every N commits or M minutes), OR
- At minimum, document this characteristic on the config option and recommend conservative `history.expire.max-snapshot-age-ms` / `history.expire.min-snapshots-to-keep` settings?

I don't know whether it's a blocker, but I thought to mention.

vamossagar12:
> @vamossagar12 Thanks for this.. I have one concen that i would like your thoughts on as well us from @luoyuxia and @MehulBatra ... Unlike Paimon, where setting `WRITE_ONLY=false` lets expiration run in-band as part of the normal commit, Iceberg's `ExpireSnapshots.commit()` is a **separate metadata commit** that:
> 
> * Plans across manifests
> * Writes a new `TableMetadata` JSON
> * Synchronously deletes files (because `cleanExpiredFiles=true`)
> 
> In a streaming tiering scenario, this happens on **every commit** with no throttling. This can materially increase per-commit latency and produce a metadata/delete storm on object stores. Could we:
> 
> * Add an opt-in interval (e.g. `lake.iceberg.expire-snapshot.interval` defaulting to every N commits or M minutes), OR
> * At minimum, document this characteristic on the config option and recommend conservative `history.expire.max-snapshot-age-ms` / `history.expire.min-snapshots-to-keep` settings?
> 
> I don't know whether it's a blocker, but I thought to mention.

Thanks @polyzos , it makes sense. Actually I went ahead with option 1 to allow ease of use for the users. I introduced 2  Iceberg-specific config options:

```
lake.iceberg.expire-snapshot.min-interval-commits (default: 10) — minimum commits since last expiration
lake.iceberg.expire-snapshot.min-interval-ms (default: 10min) — minimum time elapsed since last expiration
```

Expiration only fires when both thresholds are satisfied simultaneously. This means:

- In a fast-streaming scenario (e.g. 1 commit/sec), the 10-minute time gate acts as the primary rate-limiter — expiration runs at most once every 10 minutes regardless of commit frequency
- In a slow scenario (e.g. 1 commit/hour), the count gate ensures we don't expire too frequently either

This way the snapshotting can be controlled by the users based on commit counts or time based. If these look ok, we may need to document these (in this PR or a fast follow-up one) Let me now what you think!

vamossagar12:
`org.apache.fluss.client.security.acl.FlussAuthorizationITCase.testDeleteProducerOffsetsAuthorization` is a test which failed but it's known to be flake. I see https://github.com/apache/fluss/pull/2866 to fix it (not merged yet).

vamossagar12:
@polyzos , gentle ping, can you please take a look at this one again when you have some time.. Thanks..

PR #2886, opened 2026-03-17
Add V2 batch format with statistics collection

### LINKED ISSUE #2885: [common] Add V2 batch format with statistics collection for filter pushdown
### Search before asking

### Motivation

Add V2 log record batch format that collects per-column min/max statistics to enable efficient filter│pushdown.

### Solution

V2 batch layout:  [Header (56 bytes)] [Statistics (variable)] [Records (variable)]. See : https://cwiki.apache.org/confluence/display/FLUSS/FIP-10%3A+Support+Log+RecordBatch+Filter+Pushdown

### Anything else?

_No response_

### Willingness to contribute

platinumhamburg:
Introduce V2 batch format that collects min/max statistics for each column to enable efficient filtering.

- Add LogRecordBatchStatistics and related classes for statistics collection
- Add StatisticsConfigUtils for parsing table.statistics.columns configuration
- Extend DefaultLogRecordBatch to support V2 format with statistics
- Place statistics data between header and records with StatisticsLength field
- Add comprehensive tests for statistics collection and parsing

### Purpose

Linked issue: close #2885

### DIFF
--- a/fluss-common/src/main/java/org/apache/fluss/record/DefaultLogRecordBatchStatistics.java
+++ b/fluss-common/src/main/java/org/apache/fluss/record/DefaultLogRecordBatchStatistics.java
@@ -105,6 +105,8 @@ public class DefaultLogRecordBatchStatistics implements LogRecordBatchStatistics
         this.statsNullCounts = nullCounts;
         this.minValuesOffset = minValuesOffset;
         this.maxValuesOffset = maxValuesOffset;
+        checkArgument(minValuesSize > 0, "minValuesSize must be positive");
+        checkArgument(maxValuesSize > 0, "maxValuesSize must be positive");
         this.minValuesSize = minValuesSize;
         this.maxValuesSize = maxValuesSize;
         this.statsIndexMapping = statsIndexMapping;
@@ -117,10 +119,6 @@ public class DefaultLogRecordBatchStatistics implements LogRecordBatchStatistics
 
     @Override
     public InternalRow getMinValues() {
-        if (minValuesSize == 0) {
-            return null;
-        }
-
         // Return cached row if already created
         if (cachedMinRow != null) {
             return cachedMinRow;
@@ -135,10 +133,6 @@ public class DefaultLogRecordBatchStatistics implements LogRecordBatchStatistics
 
     @Override
     public InternalRow getMaxValues() {
-        if (maxValuesSize == 0) {
-            return null;
-        }
-
         // Return cached row if already created
         if (cachedMaxRow != null) {
             return cachedMaxRow;
--- a/fluss-common/src/main/java/org/apache/fluss/record/FileLogProjection.java
+++ b/fluss-common/src/main/java/org/apache/fluss/record/FileLogProjection.java
@@ -21,8 +21,6 @@ import org.apache.fluss.annotation.VisibleForTesting;
 import org.apache.fluss.compression.ArrowCompressionInfo;
 import org.apache.fluss.exception.InvalidColumnProjectionException;
 import org.apache.fluss.metadata.SchemaGetter;
-import org.apache.fluss.record.FileLogInputStream.FileChannelLogRecordBatch;
-import org.apache.fluss.record.bytesview.BytesView;
 import org.apache.fluss.record.bytesview.MultiBytesView;
 import org.apache.fluss.shaded.arrow.com.google.flatbuffers.FlatBufferBuilder;
 import org.apache.fluss.shaded.arrow.org.apache.arrow.flatbuf.Buffer;
@@ -72,7 +70,6 @@ import static org.apache.fluss.record.LogRecordBatchFormat.schemaIdOffset;
 import static org.apache.fluss.record.LogRecordBatchFormat.statisticsLengthOffset;
 import static org.apache.fluss.utils.FileUtils.readFully;
 import static org.apache.fluss.utils.FileUtils.readFullyOrFail;
-import static org.apache.fluss.utils.Preconditions.checkNotNull;
 import static org.apache.fluss.utils.Preconditions.checkState;
 
 /** Column projection util on Arrow format {@link FileLogRecords}. */
@@ -125,124 +122,6 @@ public class FileLogProjection {
         this.schemaGetter = schemaGetter;
         this.compressionInfo = compressionInfo;
         this.selectedFieldPositions = selectedFieldPositions;
-    }
-
-    public BytesView projectRecordBatch(FileChannelLogRecordBatch batch) throws IOException {
-        // Get schema ID from the batch to determine the projection
-        FileChannel channel = batch.fileRecords.channel();
-        int position = batch.position();
-
-        // Read the log header to get schema ID
-        logHeaderBuffer.rewind();
-        readLogHeaderFullyOrFail(channel, logHeaderBuffer, position);
-        logHeaderBuffer.rewind();
-        byte magic = logHeaderBuffer.get(MAGIC_OFFSET);
-        logHeaderBuffer.rewind();
-        short schemaId = logHeaderBuffer.getShort(schemaIdOffset(magic));
-
-        ProjectionInfo currentProjection = getOrCreateProjectionInfo(schemaId);
-        checkNotNull(currentProjection, "There is no projection registered yet.");
-
-        int end = position + batch.sizeInBytes();
-        MultiBytesView.Builder builder = MultiBytesView.builder();
-        int batchSizeInBytes = LOG_OVERHEAD + logHeaderBuffer.getInt(LENGTH_OFFSET);
-        if (position > end - batchSizeInBytes) {
-            // the remaining bytes in the file are not enough to read a full batch
-            return builder.build();
-        }
-
-        // Return null if meets empty batch. The empty batch was generated when build cdc log batch
-        // when there
-        // is no cdc log generated for this kv batch. See the comments about the field
-        // 'lastOffsetDelta' in DefaultLogRecordBatch.
-        int recordBatchHeaderSize = recordBatchHeaderSize(magic);
-        if (batchSizeInBytes == recordBatchHeaderSize) {
-            return builder.build();
-        }
-
-        boolean isAppendOnly =
-                (logHeaderBuffer.get(attributeOffset(magic)) & APPEND_ONLY_FLAG_MASK) > 0;
-
-        // For V1+, skip statistics data between header and records
-        int statisticsLength = 0;
-        if (magic >= LogRecordBatchFormat.LOG_MAGIC_VALUE_V1) {
-            statisticsLength = logHeaderBuffer.getInt(statisticsLengthOffset(magic));
-        }
-        int recordsStartOffset = recordBatchHeaderSize + statisticsLength;
-
-        final int changeTypeBytes;
-        final long arrowHeaderOffset;
-        if (isAppendOnly) {
-            changeTypeBytes = 0;
-            arrowHeaderOffset = position + recordsStartOffset;
-        } else {
-            changeTypeBytes = logHeaderBuffer.getInt(recordsCountOffset(magic));
-            arrowHeaderOffset = position + recordsStartOffset + changeTypeBytes;
-        }
-
-        // read arrow header
-        arrowHeaderBuffer.rewind();
-        readFullyOrFail(channel, arrowHeaderBuffer, arrowHeaderOffset, "arrow header");
-        arrowHeaderBuffer.position(ARROW_IPC_METADATA_SIZE_OFFSET);
-        int arrowMetadataSize = arrowHeaderBuffer.getInt();
-
-        resizeArrowMetadataBuffer(arrowMetadataSize);
-        arrowMetadataBuffer.rewind();
-        readFullyOrFail(
-                channel,
-                arrowMetadataBuffer,
-                arrowHeaderOffset + ARROW_HEADER_SIZE,
-                "arrow metadata");
-
-        arrowMetadataBuffer.rewind();
-        Message metadata = Message.getRootAsMessage(arrowMetadataBuffer);
-        ProjectedArrowBatch projectedArrowBatch =
-                projectArrowBatch(
-                        metadata,
-                        currentProjection.nodesProjection,
-                        currentProjection.buffersProjection,
-                        currentProjection.bufferCount);
-        long arrowBodyLength = projectedArrowBatch.bodyLength();
-
-        int newBatchSizeInBytes =
-                recordBatchHeaderSize
-                        + changeTypeBytes
-                        + currentProjection.arrowMetadataLength
-                        + (int) arrowBodyLength;
-
-        // 3. create new arrow batch metadata which already projected.
-        byte[] headerMetadata =
-                serializeArrowRecordBatchMetadata(
-                        projectedArrowBatch, arrowBodyLength, currentProjection.bodyCompression);
-        checkState(
-                headerMetadata.length == currentProjection.arrowMetadataLength,
-                "Invalid metadata length");
-
-        // 4. update and copy log batch header
-        logHeaderBuffer.position(LENGTH_OFFSET);
-        logHeaderBuffer.putInt(newBatchSizeInBytes - LOG_OVERHEAD);
-
-        // For V1+ format, clear statistics information since projection removes statistics
-        if (statisticsLength > 0) {
-            LogRecordBatchFormat.clearStatisticsFromHeader(logHeaderBuffer, magic);
-        }
-
-        logHeaderBuffer.rewind();
-        // the logHeader can't be reused, as it will be sent to network
-        byte[] logHeader = new byte[recordBatchHeaderSize];
-        logHeaderBuffer.get(logHeader);
-
-        // 5. build log records
-        builder.addBytes(logHeader);
-        if (!isAppendOnly) {
-            builder.addBytes(channel, position + recordsStartOffset, changeTypeBytes);
-        }
-        builder.addBytes(headerMetadata);
-        final long bufferOffset = arrowHeaderOffset + ARROW_HEADER_SIZE + arrowMetadataSize;
-        projectedArrowBatch.buffers.forEach(
-                b -> builder.addBytes(channel, bufferOffset + b.getOffset(), (int) b.getSize()));
-
-        return builder.build();
     }
 
     /**
--- a/fluss-common/src/main/java/org/apache/fluss/record/LogRecordBatchFormat.java
+++ b/fluss-common/src/main/java/org/apache/fluss/record/LogRecordBatchFormat.java
@@ -120,7 +120,7 @@ public class LogRecordBatchFormat {
      * ------------------------------------------
      * </pre>
      *
-     * @since 0.8
+     * @since 1.0
      */
     public static final byte LOG_MAGIC_VALUE_V2 = 2;
 
@@ -195,7 +195,7 @@ public class LogRecordBatchFormat {
      * ------------------------------------------
      * </pre>
      *
-     * @since 0.8
+     * @since 1.0
      */
     public static final byte LOG_MAGIC_VALUE_V1 = 1;
 
--- a/fluss-common/src/main/java/org/apache/fluss/record/LogRecordBatchStatisticsCollector.java
+++ b/fluss-common/src/main/java/org/apache/fluss/record/LogRecordBatchStatisticsCollector.java
@@ -111,7 +111,7 @@ public class LogRecordBatchStatisticsCollector {
      */
     public int estimatedSizeInBytes() {
         return statisticsWriter.estimatedSizeInBytes(
-                GenericRow.of(minValues), GenericRow.of(maxValues), nullCounts);
+                GenericRow.of(minValues), GenericRow.of(maxValues));
     }
 
     /** Reset the collector to collect new statistics. */
--- a/fluss-common/src/main/java/org/apache/fluss/record/LogRecordBatchStatisticsWriter.java
+++ b/fluss-common/src/main/java/org/apache/fluss/record/LogRecordBatchStatisticsWriter.java
@@ -253,11 +253,9 @@ public class LogRecordBatchStatisticsWriter {
      *
      * @param minValues The minimum values as InternalRow, can be null
      * @param maxValues The maximum values as InternalRow, can be null
-     * @param nullCounts The null counts array
      * @return The estimated number of bytes required for serialization
      */
-    public int estimatedSizeInBytes(
-            InternalRow minValues, InternalRow maxValues, Long[] nullCounts) {
+    public int estimatedSizeInBytes(InternalRow minValues, InternalRow maxValues) {
 
         int totalEstimatedBytes = 0;
 
--- a/fluss-common/src/main/java/org/apache/fluss/config/StatisticsColumnsConfig.java
+++ b/fluss-common/src/main/java/org/apache/fluss/config/StatisticsColumnsConfig.java
@@ -34,7 +34,7 @@ import static org.apache.fluss.utils.Preconditions.checkNotNull;
  *   <li>{@link Mode#SPECIFIED}: Collect statistics for specific columns only.
  * </ul>
  *
- * @since 0.6
+ * @since 1.0
  */
 @PublicEvolving
 public class StatisticsColumnsConfig {
--- a/fluss-common/src/main/java/org/apache/fluss/types/DataTypeChecks.java
+++ b/fluss-common/src/main/java/org/apache/fluss/types/DataTypeChecks.java
@@ -63,16 +63,6 @@ public final class DataTypeChecks {
     }
 
     /**
-     * Check if the given data type is a binary type (BINARY or BYTES).
-     *
-     * @param dataType the data type to check
-     * @return true if the data type is binary, false otherwise
-     */
-    public static boolean isBinaryType(DataType dataType) {
-        return dataType.isAnyOf(DataTypeRoot.BINARY, DataTypeRoot.BYTES);
-    }
-
-    /**
      * Check if the given data type is supported for statistics collection. Uses a whitelist
      * approach to ensure only types with meaningful statistical properties are included.
      *
--- a/fluss-common/src/test/java/org/apache/fluss/config/StatisticsColumnsConfigTest.java
+++ b/fluss-common/src/test/java/org/apache/fluss/config/StatisticsColumnsConfigTest.java
@@ -1,0 +1,158 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.apache.fluss.config;
+
+import org.junit.jupiter.api.Test;
+
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.List;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
+
+/** Test for {@link StatisticsColumnsConfig}. */
+class StatisticsColumnsConfigTest {
+
+    @Test
+    void testDisabledMode() {
+        StatisticsColumnsConfig config = StatisticsColumnsConfig.disabled();
+        assertThat(config.getMode()).isEqualTo(StatisticsColumnsConfig.Mode.DISABLED);
+        assertThat(config.getColumns()).isEmpty();
+        assertThat(config.isEnabled()).isFalse();
+    }
+
+    @Test
+    void testAllMode() {
+        StatisticsColumnsConfig config = StatisticsColumnsConfig.all();
+        assertThat(config.getMode()).isEqualTo(StatisticsColumnsConfig.Mode.ALL);
+        assertThat(config.getColumns()).isEmpty();
+        assertThat(config.isEnabled()).isTrue();
+    }
+
+    @Test
+    void testSpecifiedMode() {
+        List<String> columns = Arrays.asList("col1", "col2", "col3");
+        StatisticsColumnsConfig config = StatisticsColumnsConfig.of(columns);
+        assertThat(config.getMode()).isEqualTo(StatisticsColumnsConfig.Mode.SPECIFIED);
+        assertThat(config.getColumns()).containsExactly("col1", "col2", "col3");
+        assertThat(config.isEnabled()).isTrue();
+    }
+
+    @Test
+    void testSpecifiedModeWithEmptyList() {
+        StatisticsColumnsConfig config = StatisticsColumnsConfig.of(Collections.emptyList());
+        assertThat(config.getMode()).isEqualTo(StatisticsColumnsConfig.Mode.SPECIFIED);
+        assertThat(config.getColumns()).isEmpty();
+        assertThat(config.isEnabled()).isTrue();
+    }
+
+    @Test
+    void testSpecifiedModeColumnsAreUnmodifiable() {
+        List<String> columns = Arrays.asList("col1", "col2");
+        StatisticsColumnsConfig config = StatisticsColumnsConfig.of(columns);
+        assertThatThrownBy(() -> config.getColumns().add("col3"))
+                .isInstanceOf(UnsupportedOperationException.class);
+    }
+
+    @Test
+    void testOfWithNullThrows() {
+        assertThatThrownBy(() -> StatisticsColumnsConfig.of(null))
+                .isInstanceOf(NullPointerException.class)
+                .hasMessageContaining("columns must not be null");
+    }
+
+    @Test
+    void testDisabledSingleton() {
+        StatisticsColumnsConfig config1 = StatisticsColumnsConfig.disabled();
+        StatisticsColumnsConfig config2 = StatisticsColumnsConfig.disabled();
+        assertThat(config1).isSameAs(config2);
+    }
+
+    @Test
+    void testAllSingleton() {
+        StatisticsColumnsConfig config1 = StatisticsColumnsConfig.all();
+        StatisticsColumnsConfig config2 = StatisticsColumnsConfig.all();
+        assertThat(config1).isSameAs(config2);
+    }
+
+    @Test
+    void testEqualsAndHashCode() {
+        // same mode and columns
+        StatisticsColumnsConfig specified1 =
+                StatisticsColumnsConfig.of(Arrays.asList("col1", "col2"));
+        StatisticsColumnsConfig specified2 =
+                StatisticsColumnsConfig.of(Arrays.asList("col1", "col2"));
+        assertThat(specified1).isEqualTo(specified2);
+        assertThat(specified1.hashCode()).isEqualTo(specified2.hashCode());
+
+        // same instance
+        assertThat(specified1).isEqualTo(specified1);
+
+        // null and different type
+        assertThat(specified1).isNotEqualTo(null);
+        assertThat(specified1).isNotEqualTo("not a config");
+
+        // different columns
+        StatisticsColumnsConfig specified3 =
+                StatisticsColumnsConfig.of(Arrays.asList("col1", "col3"));
+        assertThat(specified1).isNotEqualTo(specified3);
+
+        // different modes
+        assertThat(StatisticsColumnsConfig.disabled()).isNotEqualTo(StatisticsColumnsConfig.all());
+        assertThat(StatisticsColumnsConfig.disabled()).isNotEqualTo(specified1);
+        assertThat(StatisticsColumnsConfig.all()).isNotEqualTo(specified1);
+
+        // disabled equals disabled
+        assertThat(StatisticsColumnsConfig.disabled())
+                .isEqualTo(StatisticsColumnsConfig.disabled());
+        assertThat(StatisticsColumnsConfig.disabled().hashCode())
+                .isEqualTo(StatisticsColumnsConfig.disabled().hashCode());
+
+        // all equals all
+        assertThat(StatisticsColumnsConfig.all()).isEqualTo(StatisticsColumnsConfig.all());
+        assertThat(StatisticsColumnsConfig.all().hashCode())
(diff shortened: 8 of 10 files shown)


fluss-common/src/main/java/org/apache/fluss/config/TableConfig.java
```diff
@@ -154,4 +157,42 @@ public AutoPartitionStrategy getAutoPartitionStrategy() {
...
+     *     statistics collection is disabled (empty string configuration)
+     */
+    public Optional<List<String>> getStatisticsColumns() {
+        String columnsStr = config.get(ConfigOptions.TABLE_STATISTICS_COLUMNS);
+        if (columnsStr.isEmpty()) {
+            return null; // null means statistics collection is disabled
+        }
+        if ("*".equals(columnsStr)) {
+            return Optional.empty(); // Empty means collect all non-binary columns
+        }
```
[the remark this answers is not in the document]
wuchong: +1 for this. `TableConfig` is a public API, we should improve the semantics to be clear. We can introduce a `StatisticsColumnsConfig` to represent the tri-state type.


```java
public class StatisticsColumnsConfig {

    public enum Mode {
        /** Statistics collection is disabled. */
        DISABLED,
        /** Collect statistics for all non-binary columns. */
        ALL,
        /** Collect statistics only for specified columns. */
        SPECIFIED
    }

    private final Mode mode;
    private final List<String> columns;

    private static final StatisticsColumnsConfig DISABLED_INSTANCE =
            new StatisticsColumnsConfig(Mode.DISABLED, Collections.emptyList());
    private static final StatisticsColumnsConfig ALL_INSTANCE =
            new StatisticsColumnsConfig(Mode.ALL, Collections.emptyList());

    private StatisticsColumnsConfig(Mode mode, List<String> columns) {
        this.mode = mode;
        this.columns = columns;
    }

    public static StatisticsColumnsConfig disabled() {
        return DISABLED_INSTANCE;
    }

    public static StatisticsColumnsConfig all() {
        return ALL_INSTANCE;
    }

    public static StatisticsColumnsConfig of(List<String> columns) {
        return new StatisticsColumnsConfig(Mode.SPECIFIED, columns);
    }

    public Mode getMode() {
        return mode;
    }

    public List<String> getColumns() {
        return columns;
    }
}
```

Then the ` getStatisticsColumns() ` can be changed to 

```java
public StatisticsColumnsConfig getStatisticsColumns() {
    String columnsStr = config.get(ConfigOptions.TABLE_STATISTICS_COLUMNS);
    if (columnsStr.isEmpty()) {
        return StatisticsColumnsConfig.disabled();
    }
    if ("*".equals(columnsStr)) {
        return StatisticsColumnsConfig.all();
    }
    List<String> columns =
            Arrays.stream(columnsStr.split(","))
                    .map(String::trim)
                    .filter(s -> !s.isEmpty())
    return StatisticsColumnsConfig.of(columns);
}
```

And ` isStatisticsEnabled() ` can be simplified to 

```java
public boolean isStatisticsEnabled() {
    return getStatisticsColumns().getMode() != StatisticsColumnsConfig.Mode.DISABLED;
}
````

File: fluss-common/src/main/java/org/apache/fluss/config/TableConfig.java [hunk 2]
```diff
@@ -154,4 +157,42 @@ public AutoPartitionStrategy getAutoPartitionStrategy() {
...
+                        .collect(Collectors.toList());
+        return Optional.of(columns);
+    }
+
+    /**
+     * Checks whether the table is configured to collect statistics for all non-binary columns.
+     *
+     * @return true if configured with "*" (collect all non-binary columns), false otherwise
+     */
+    public boolean isCollectAllNonBinaryColumns() {
```
wuchong: Not used, can be remove?

fluss-common/src/main/java/org/apache/fluss/config/TableConfig.java
```diff
@@ -154,4 +157,42 @@ public AutoPartitionStrategy getAutoPartitionStrategy() {
...
+        }
+        if ("*".equals(columnsStr)) {
+            return Optional.empty(); // Empty means collect all non-binary columns
+        }
+        List<String> columns =
+                Arrays.stream(columnsStr.split(","))
+                        .map(String::trim)
+                        .filter(s -> !s.isEmpty())
+                        .collect(Collectors.toList());
+        return Optional.of(columns);
```
wuchong: nit: Besides, the logic seems is a little duplicated with `StatisticsConfigUtils`. Maybe we can reuse some code.

fluss-common/src/main/java/org/apache/fluss/config/ConfigOptions.java
```diff
@@ -1579,6 +1579,18 @@ public class ConfigOptions {
                                     + "This mode reduces storage and transmission costs but loses the ability to track previous values. "
                                     + "This option only affects primary key tables.");
 
+    public static final ConfigOption<String> TABLE_STATISTICS_COLUMNS =
+            key("table.statistics.columns")
+                    .stringType()
+                    .defaultValue("*")
```
wuchong: We should use `noDefaultValue()` or `''` as default value. Forcing all columns to collect stats is a big overhead and introduces regression to users.

fluss-common/src/main/java/org/apache/fluss/config/StatisticsConfigUtils.java
```diff
@@ -0,0 +1,133 @@
...
+    /**
+     * Validates statistics configuration for a table descriptor.
+     *
+     * @param tableDescriptor the table descriptor to validate
+     * @throws InvalidConfigException if the statistics configuration is invalid
+     */
+    public static void validateStatisticsConfig(TableDescriptor tableDescriptor) {
+        Map<String, String> properties = tableDescriptor.getProperties();
+        String statisticsColumns =
+                properties.getOrDefault(ConfigOptions.TABLE_STATISTICS_COLUMNS.key(), "*");
```
wuchong: Default should be null or empty.

fluss-common/src/main/java/org/apache/fluss/record/LogRecordBatchFormat.java
```diff
@@ -57,6 +59,91 @@ public class LogRecordBatchFormat {
...
+     * +----------------+------------------+------------------+
+     * | BaseOffset     | Length           | Magic            |
+     * | (8 bytes)      | (4 bytes)        | (1 byte)         |
+     * +----------------+------------------+------------------+
+     * | CommitTimestamp                   | LeaderEpoch      |
+     * | (8 bytes)                        | (4 bytes)        |
+     * +----------------------------------+------------------+
+     * | CRC            | SchemaId         | Attributes       |
+     * | (4 bytes)      | (2 bytes)        | (1 byte)         |
+     * |                |                  | [STATISTICS_FLAG]|
```
wuchong: Since `StatisticsLength` is always serialized in the V2 format, the `STATISTICS_FLAG` in the attributes appears redundant. Instead, we can determine the presence of statistics in the record batch by checking whether `StatisticsLength` is zero.

fluss-common/src/main/java/org/apache/fluss/record/MemoryLogRecordsArrowBuilder.java
wuchong: remove the legacy `builder()` that is not used  anymore.

fluss-common/src/main/java/org/apache/fluss/record/MemoryLogRecordsArrowBuilder.java
```diff
@@ -145,13 +193,74 @@ public MultiBytesView build() throws IOException {
...
+                } else {
+                    viewBuilder.addBytes(
+                            segView.getMemorySegment(),
+                            segView.getPosition(),
+                            segView.getBytesLength());
+                }
+            }
+        } else {
+            viewBuilder.addMemorySegmentByteViewList(writtenSegments);
+        }
```
wuchong: This approach introduces significant complexity and reduces efficiency due to repeated statistics memory copying. Our goal is to maintain a contiguous memory segment and enable direct writes to the `pagedOutputView` without copying or clipping.

A straightforward solution is to implement a `serializeToOutputView()` method on `statisticsCollector`, similar to `arrowWriter.serializeToOutputView()`. This allows direct serialization of statistics into the `pagedOutputView`, eliminating the need to allocate and copy temporary memory segments.

Next, serialize the `changeTypeVector` into the `pagedOutputView`. To support this, we must refactor `ChangeTypeVectorWriter` to write change types into a lazily allocated, unpooled list of `MemorySegment` objects instead of writing directly to the `pagedOutputView`. Each `MemorySegment` can be a 1024-byte aligned heap byte array; if the current segment becomes full, a new one should be allocated.

Finally, serialize the Arrow data into the `pagedOutputView`.

fluss-common/src/main/java/org/apache/fluss/record/LogRecordBatchFormat.java
```diff
@@ -185,108 +272,212 @@ public class LogRecordBatchFormat {
...
     public static int crcOffset(byte magic) {
         switch (magic) {
-            case LOG_MAGIC_VALUE_V1:
-                return V1_CRC_OFFSET;
             case LOG_MAGIC_VALUE_V0:
                 return V0_CRC_OFFSET;
+            case LOG_MAGIC_VALUE_V1:
+                return V1_CRC_OFFSET;
+            case LOG_MAGIC_VALUE_V2:
+                return V2_CRC_OFFSET;
```
wuchong, line 293: It is on purpose to put the latest version in the first place for better perf.

fluss-common/src/main/java/org/apache/fluss/record/LogRecordBatchFormat.java
```diff
@@ -185,108 +272,212 @@ public class LogRecordBatchFormat {
...
+    /**
+     * Get the statistics length field offset for the given magic version. Only available for V2 and
+     * later.
+     */
+    public static int statisticsLengthOffset(byte magic) {
+        switch (magic) {
             case LOG_MAGIC_VALUE_V0:
-                return V0_RECORD_BATCH_HEADER_SIZE;
+            case LOG_MAGIC_VALUE_V1:
+                return 0;
```
wuchong: We should throw an explicit exception for V0 and V1, because there is no stats length offset in old format. Returning 0 will lead to hidden problems. Similar to how we handle the newly introduced field `leaderEpochOffset` since V1.

fluss-common/src/main/java/org/apache/fluss/record/LogRecordBatchFormat.java
```diff
@@ -185,108 +272,212 @@ public class LogRecordBatchFormat {
...
+                throw new IllegalArgumentException(
+                        "Statistics not supported in magic version " + magic);
+            case LOG_MAGIC_VALUE_V2:
+                return V2_STATISTICS_DATA_OFFSET;
             default:
                 throw new IllegalArgumentException("Unsupported magic value " + magic);
         }
     }
 
     public static int arrowChangeTypeOffset(byte magic) {
```
wuchong: We should never use this method in code because it returns incorrect offsets for V2 with stats, which poses a significant risk of introducing bugs. We should either replace it with `arrowChangeTypeOffsetWithStats` (potentially renaming it to `arrowChangeTypeOffset` if this method is removed) or refactor the existing code that relies on it.

fluss-common/src/main/java/org/apache/fluss/record/LogRecordBatchStatisticsCollector.java
```diff
@@ -0,0 +1,242 @@
...
+            java.util.Comparator<T> comparator,
+            Function<T, T> valueTransformer) {
+        if (!minSet[statsIndex]) {
+            minValues[statsIndex] = valueTransformer.apply(value);
+            minSet[statsIndex] = true;
+        } else {
+            T currentMin = (T) minValues[statsIndex];
+            if (comparator.compare(value, currentMin) < 0) {
+                minValues[statsIndex] = valueTransformer.apply(value);
+            }
```
wuchong: It seems we can directly check the nullability of the element at `statsIndex` in `minValues` array, rather than relying on an additional `minSet` array? This can reduce the cost of the stats collector.

fluss-common/src/main/java/org/apache/fluss/record/LogRecordBatchStatisticsParser.java
```diff
@@ -0,0 +1,379 @@
...
+
+            // Calculate min values offset
+            int minValuesOffset = currentPos - position;
+
+            if (minValuesSize > 0) {
+                currentPos += minValuesSize;
+                // Advance the inputView position to match currentPos
+                for (int i = 0; i < minValuesSize; i++) {
+                    inputView.readByte();
+                }
```
wuchong: optimization: we can skip reading the actual bytes for min/max values
since we will read them lazily later. We just need to advance the position.
However, `MemorySegmentInputView` does not currently support skipping bytes,
so we can add a `skipBytes()` method on the `MemorySegmentInputView` to advance
the position only.

fluss-common/src/main/java/org/apache/fluss/record/LogRecordBatchStatisticsParser.java
```diff
@@ -0,0 +1,379 @@
...
+     * evaluation of min/max values.
+     *
+     * @param segment The memory segment containing the serialized statistics data
+     * @param position The byte position in the segment where statistics data begins
+     * @param rowType The row type schema used to interpret column types and validate structure
+     * @param schemaId The schema identifier for version compatibility and data interpretation
+     * @return A DefaultLogRecordBatchStatistics object with zero-copy access to the data, or null
+     *     if the data is invalid, corrupted, or has incompatible version
+     */
+    public static DefaultLogRecordBatchStatistics parseStatistics(
```
wuchong, line 91: Add `@Nullable` annotation to the method return type.

fluss-common/src/main/java/org/apache/fluss/record/LogRecordBatchStatisticsParser.java
```diff
@@ -0,0 +1,379 @@
...
+                    rowType,
+                    schemaId,
+                    nullCounts,
+                    minValuesOffset,
+                    maxValuesOffset,
+                    minValuesSize,
+                    maxValuesSize,
+                    statsIndexMapping);
+        } catch (Exception e) {
+            return null;
```
wuchong, line 149: add logging here to log the exception and the fact that parsing failed, so we can debug issues with malformed data

fluss-common/src/main/java/org/apache/fluss/record/LogRecordBatchStatisticsParser.java
```diff
@@ -0,0 +1,379 @@
...
+     * @return A DefaultLogRecordBatchStatistics object with zero-copy access to the data, or null
+     *     if the buffer is null/empty or contains invalid data
+     */
+    public static DefaultLogRecordBatchStatistics parseStatistics(
+            ByteBuffer buffer, RowType rowType, int schemaId) {
+        if (buffer == null || buffer.remaining() == 0) {
+            return null;
+        }
+
+        MemorySegment segment = MemorySegment.wrapOffHeapMemory(buffer);
```
wuchong: Check the buffer is direct memory, before convert it into `wrapOffHeapMemory`. Otherwise, fallback to call `org.apache.fluss.record.LogRecordBatchStatisticsParser#parseStatistics(byte[], org.apache.fluss.types.RowType, int)` automantically.

fluss-common/src/main/java/org/apache/fluss/record/LogRecordBatchStatisticsParser.java
```diff
@@ -0,0 +1,379 @@
...
+     * @return true if the data appears to contain valid statistics format, false otherwise
+     */
+    public static boolean isValidStatistics(
+            MemorySegment segment, int position, int size, RowType rowType) {
+        if (segment == null || size < 3) {
+            return false;
+        }
+
+        try {
+            MemorySegmentInputView inputView = new MemorySegmentInputView(segment, position);
```
wuchong: Maintain field offsets in `LogRecordBatchFormat` or `LogRecordBatchStatistics` to
avoid creating `MemorySegmentInputView` and reading bytes one by one for validation.
This can uniformly to read fields by reading from `MemorySegment` directly with known
offsets, which is more efficient than creating `MemorySegmentInputView` and avoid changing many places when changing the format.

fluss-common/src/main/java/org/apache/fluss/record/LogRecordBatchStatisticsParser.java
```diff
@@ -0,0 +1,379 @@
...
+            for (int i = 0; i < statisticsColumnCount; i++) {
+                inputView.readShort();
+            }
+            currentPos += statisticsColumnCount * 2;
+
+            // Skip null counts (4 bytes per statistics column)
+            int nullCountsSize = statisticsColumnCount * 4;
+            for (int i = 0; i < nullCountsSize; i++) {
+                inputView.readByte();
+            }
```
wuchong: Change to read from `MemorySegment` directly, so we don't need to skipping bytes by reading bytes.

fluss-common/src/main/java/org/apache/fluss/record/LogRecordBatchStatisticsWriter.java
```diff
@@ -0,0 +1,216 @@
...
+        int maxRowBytes = writeRowData(maxValues, outputView);
+        totalBytesWritten += maxRowBytes;
+
+        return totalBytesWritten;
+    }
+
+    private int writeRowData(InternalRow row, OutputView outputView) throws IOException {
+        if (row != null) {
+            // Use AlignedRow.from() to handle conversion uniformly
+            AlignedRow binaryRowData = AlignedRow.from(statsRowType, row);
```
wuchong: Consider to reuse the `AlignedRowWriter` in `LogRecordBatchStatisticsWriter`, so that we don't need to allocate a new byte array for every writes which can reduce a lot of cost.

fluss-common/src/main/java/org/apache/fluss/record/LogRecordBatchStatisticsWriter.java
```diff
@@ -0,0 +1,216 @@
...
+            int rowSize = binaryRowData.getSizeInBytes();
+            outputView.writeInt(rowSize);
+            // Copy data using byte array since copyTo requires byte[] parameter
+            byte[] rowData = new byte[rowSize];
+            binaryRowData.copyTo(rowData, 0);
+            outputView.write(rowData);
```
wuchong: This approach introduces unnecessary data copying. Since the underlying MemorySegment of AlignedRow resides in an on-heap byte array, we can retrieve it directly via `AlignedRow#getSegments()[0]`. After verifying that the segment is indeed on-heap, we can efficiently extract the byte array along with its offset and length without performing additional copies (using `org.apache.fluss.memory.OutputView.write(byte[] b, int offset, int len)`).

fluss-common/src/main/java/org/apache/fluss/record/LogRecordBatchStatisticsWriter.java
```diff
@@ -0,0 +1,216 @@
...
+        }
+
+        // Use AlignedRow.from() to get accurate size estimation
+        // This is still more efficient than writing to a temporary output view
+        // as it only creates the binary representation without doing I/O
+        AlignedRow alignedRow = AlignedRow.from(statsRowType, row);
+        int rowSize = alignedRow.getSizeInBytes();
+        // 4 bytes for size field + actual row data size
+        return 4 + rowSize;
+    }
```
wuchong: This approach incurs significant overhead by performing redundant serialization. In reality, we do not require a precise size prediction; a rough estimate is sufficient. For variable-length types, we can assume a fixed size of 16 bytes, while fixed-length types can be calculated directly. This strategy allows us to cache the row size and eliminate the need for recalculating it on every access.

For reference, Flink employs a similar method to estimate binary row sizes, as seen in `BytesMapTestBase.rowLength()`:

```java
protected int rowLength(RowType tpe) {
    return BinaryRowData.calculateFixPartSizeInBytes(tpe.getFieldCount())
            + BytesHashMap.getVariableLength(tpe.getChildren().toArray(new LogicalType[0]));
}
```

The variable-length estimation logic resides in `AbstractBytesHashMap.getVariableLength(LogicalType[])`:

```java
static int getVariableLength(LogicalType[] types) {
    int length = 0;
    for (LogicalType type : types) {
        if (!BinaryRowData.isInFixedLengthPart(type)) {
            // We currently use a conservative small value assumption
            length += 16;
        }
    }
    return length;
}
```

fluss-common/src/main/java/org/apache/fluss/record/MemoryLogRecordsArrowBuilder.java
```diff
@@ -56,6 +65,7 @@ public class MemoryLogRecordsArrowBuilder implements AutoCloseable {
     private final MemorySegment firstSegment;
     private final AbstractPagedOutputView pagedOutputView;
     private final boolean appendOnly;
+    private final LogRecordBatchStatisticsCollector statisticsCollector;
```
wuchong: Add `@Nullable` annotation for the field and constructor parameter.

fluss-common/src/main/java/org/apache/fluss/record/MemoryLogRecordsArrowBuilder.java
```diff
@@ -244,6 +359,10 @@ public int estimatedSizeInBytes() {
                     arrowChangeTypeOffset(magic)
```
wuchong: We should not invoke `arrowChangeTypeOffset(magic)` in this context because it returns an incorrect offset when the magic number is `2`. Since the sole purpose of this code block is to estimate the row size, we should replace `arrowChangeTypeOffset(magic)` with `recordBatchHeaderSize(magic)` to ensure accurate calculations.

fluss-common/src/main/java/org/apache/fluss/record/MemoryLogRecordsArrowBuilder.java
```diff
@@ -273,8 +392,21 @@ private void writeBatchHeader() throws IOException {
...
+        // write attributes (appendOnly flag and statistics flag)
+        byte attributes = 0;
+        if (appendOnly) {
+            attributes |= 0x01; // set appendOnly flag
+        }
+
+        int statsSize = statisticsBytesLength();
+
+        if (statsSize > 0) {
+            attributes |= STATISTICS_FLAG_MASK;
```
wuchong: we don't need to occupy a flag bit of the attribute, because we already allocated a stats length field for the stats.

fluss-common/src/main/java/org/apache/fluss/record/MemoryLogRecordsArrowBuilder.java
```diff
@@ -287,12 +419,70 @@ private void writeBatchHeader() throws IOException {
...
+     * Compute CRC over the entire batch content starting from schemaId offset. This handles the
+     * case where statistics bytes are in a separate view between header and records.
+     *
+     * <p>CRC coverage: [header from schemaId..end] + [statistics bytes] + [records]
+     *
+     * <p>This must match the read-side validation in {@link
+     * DefaultLogRecordBatch#computeChecksum()} which computes CRC over a contiguous buffer with the
+     * same layout.
+     */
+    private long computeCrc() {
```
wuchong: The previous CRC computation was zero-copy, whereas the new implementation copies the entire batch (typically 1MB to 5MB) into an array. This introduces a significant performance regression. As previously suggested, we should serialize the stats section, change type vector, and Arrow buffers directly into the `pagedOutputView`. This approach eliminates the copying overhead and preserves the original zero-copy CRC logic, greatly simplifying the code.

fluss-common/src/main/java/org/apache/fluss/row/arrow/ArrowWriter.java
```diff
@@ -143,6 +147,10 @@ public int getWriteLimitInBytes() {
         return writeLimitInBytes;
     }
 
+    public RowType getSchema() {
```
wuchong, line 150: not used yet?

fluss-common/src/main/java/org/apache/fluss/config/StatisticsConfigUtils.java
```diff
@@ -0,0 +1,133 @@
...
+            if (!columnTypeMap.containsKey(columnName)) {
+                throw new InvalidConfigException(
+                        String.format(
+                                "Column '%s' specified in statistics collection does not exist in table schema",
+                                columnName));
+            }
+
+            // Check if column type is supported
+            DataType dataType = columnTypeMap.get(columnName);
+            if (DataTypeChecks.isBinaryType(dataType)) {
```
wuchong: Instead of using a blacklist to block unsupported types, I recommend adopting a whitelist approach to explicitly define all supported types. Since the set of data types is likely to expand—for instance, the current implementation overlooks ROW, MAP, and ARRAY types which doesn't support yet, and future plans include introducing BLOB and VARIANT types—relying on a blacklist poses a risk. Contributors adding new types may easily forget to update this specific code block, leading to potential issues.

fluss-common/src/test/java/org/apache/fluss/record/LogRecordBatchStatisticsCollectorTest.java
```diff
@@ -0,0 +1,303 @@
...
+import org.junit.jupiter.api.Test;
+
+import java.io.IOException;
+import java.util.Arrays;
+import java.util.List;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+/** Test for {@link LogRecordBatchStatisticsCollector}. */
+public class LogRecordBatchStatisticsCollectorTest {
```
wuchong, line 39: Please add a test for processing zero rows, and assert the null counts and max/min values.

fluss-common/src/test/java/org/apache/fluss/record/LogRecordBatchStatisticsCollectorTest.java
```diff
@@ -0,0 +1,303 @@
...
+                        nullableRowType,
+                        LogRecordBatchStatisticsTestUtils.createAllColumnsStatsMapping(
+                                nullableRowType));
+
+        List<Object[]> dataWithNulls =
+                Arrays.asList(
+                        new Object[] {1, "a", 10.5},
+                        new Object[] {null, "b", 20.3},
+                        new Object[] {3, null, 15.7},
+                        new Object[] {2, "c", null});
```
wuchong, line 150: I suggest to making the 3rd column all `null`, and assert the max/min values for 3rd column.

fluss-common/src/main/java/org/apache/fluss/record/DefaultLogRecordBatchStatistics.java
```diff
@@ -0,0 +1,472 @@
...
+        Arrays.fill(this.reversedStatsIndexMapping, -1);
+        for (int statsIndex = 0; statsIndex < statsIndexMapping.length; statsIndex++) {
+            this.reversedStatsIndexMapping[statsIndexMapping[statsIndex]] = statsIndex;
+        }
+    }
+
+    @Override
+    public InternalRow getMinValues() {
+        if (minValuesSize <= 0) {
+            return null;
```
wuchong: Under what circumstances could `minValuesSize` be less than zero? The `min` and `max` InternalRow instances themselves should never be null (their individual fields might be null), even if there is zero rows processed. Should we add a validation check in the constructor to ensure `minValuesSize > 0`?

fluss-common/src/main/java/org/apache/fluss/row/aligned/AlignedRow.java
```diff
@@ -476,6 +477,101 @@ public int hashCode() {
...
+                    case TIMESTAMP_WITH_LOCAL_TIME_ZONE:
+                        LocalZonedTimestampType localZonedTimestampType =
+                                (LocalZonedTimestampType) fieldType;
+                        writer.writeTimestampLtz(
+                                i,
+                                row.getTimestampLtz(i, localZonedTimestampType.getPrecision()),
+                                localZonedTimestampType.getPrecision());
+                        break;
+                    default:
+                        // For unsupported types, write as null
```
wuchong: We should throw exception for unsupported types instead of writing null
silently, otherwise, it may cause data loss and hard to debug.

fluss-common/src/main/java/org/apache/fluss/record/DefaultLogRecordBatch.java
```diff
@@ -468,4 +484,71 @@ public void remove() {
...
+            }
+
+            int statisticsLength = segment.getInt(position + statisticsLengthOffset(magic));
+            if (statisticsLength == 0) {
+                return Optional.empty();
+            }
+
+            RowType rowType = context.getRowType(schemaId());
+            if (rowType == null) {
+                return Optional.empty();
```
wuchong, line 521: nit: add some logging here to log the fact that statistics parsing is skipped due
to missing schema information, so that we can monitor if there are any issues
with missing schema information in production.

fluss-common/src/main/java/org/apache/fluss/record/DefaultLogRecordBatch.java
```diff
@@ -468,4 +484,71 @@ public void remove() {
...
+            }
+
+            int statsDataOffset = statisticsDataOffset(magic);
+            LogRecordBatchStatistics statistics =
+                    LogRecordBatchStatisticsParser.parseStatistics(
+                            segment, position + statsDataOffset, rowType, schemaId());
+            return Optional.ofNullable(statistics);
+        } catch (Exception e) {
+            return Optional.empty();
```
wuchong, line 531: nit: add some logging here to log the fact that statistics parsing is skipped due
to missing schema information, so that we can monitor if there are any issues
with missing schema information in production.

fluss-common/src/main/java/org/apache/fluss/record/FileLogInputStream.java
```diff
@@ -124,6 +139,86 @@ public int position() {
             return position;
         }
 
+        public BytesView getBytesView() {
```
wuchong: not used, can be remove?

fluss-common/src/main/java/org/apache/fluss/record/FileLogInputStream.java
```diff
@@ -124,6 +139,86 @@ public int position() {
...
+            return getBytesView(false);
+        }
+
+        /**
+         * Get the BytesView of this record batch.
+         *
+         * @param trimStatistics whether to trim statistics data from the batch
+         * @return the BytesView of this record batch, possibly without statistics
+         */
+        public BytesView getBytesView(boolean trimStatistics) {
```
wuchong: not used, can be remove? If they can be removed, also fullBatch, batchHeader and
cachedHeaderBuffer changes can be reverted.

fluss-common/src/main/java/org/apache/fluss/record/FileLogProjection.java
```diff
@@ -122,6 +127,122 @@ public void setCurrentProjection(
...
+        checkState(
+                headerMetadata.length == currentProjection.arrowMetadataLength,
+                "Invalid metadata length");
+
+        // 4. update and copy log batch header
+        logHeaderBuffer.position(LENGTH_OFFSET);
+        logHeaderBuffer.putInt(newBatchSizeInBytes - LOG_OVERHEAD);
+
+        // For V2 format, clear statistics information since projection removes statistics
+        LogRecordBatchFormat.clearStatisticsFromHeader(logHeaderBuffer, magic);
```
wuchong: We can wrap this in a condition of `if statisticsLength > 0`, to save some CPUs.

fluss-common/src/main/java/org/apache/fluss/record/LogRecordBatchStatisticsWriter.java
```diff
@@ -0,0 +1,216 @@
...
+     * This method provides a more efficient way to calculate statistics size without actually
+     * writing the data to memory.
+     *
+     * @param minValues The minimum values as InternalRow, can be null
+     * @param maxValues The maximum values as InternalRow, can be null
+     * @param nullCounts The null counts array
+     * @return The estimated number of bytes required for serialization
+     */
+    public int estimatedSizeInBytes(
+            InternalRow minValues, InternalRow maxValues, Long[] nullCounts) {
```
wuchong: nit: the parameter `nullCounts` is not used.

fluss-common/src/main/java/org/apache/fluss/record/LogRecordBatchFormat.java
```diff
@@ -57,6 +59,91 @@ public class LogRecordBatchFormat {
...
+     * |  Unused (2-7)   |  Statistics Flag (1) |  AppendOnly Flag (0) |
+     * -----------------------------------------------------------------------
+     * </pre>
+     *
+     * <p>Bit 1 (Statistics Flag): Set to 1 if statistics are present, 0 otherwise Bit 0 (AppendOnly
+     * Flag): Set to 1 if batch is append-only, 0 otherwise
+     *
+     * @since 0.8
+     */
+    public static final byte LOG_MAGIC_VALUE_V2 = 2;
```
wuchong, line 125: Since V1 is never enabled and is blocked by https://github.com/apache/fluss/issues/673. I suggest to update the version to V1 for column stats, and move previous V1 to V2. Otherwise, there is problems when we enable V2 because the leaderEpoch field is empty.

fluss-common/src/main/java/org/apache/fluss/record/MemoryLogRecordsArrowBuilder.java
```diff
@@ -110,7 +142,24 @@ public static MemoryLogRecordsArrowBuilder builder(
...
+    public static MemoryLogRecordsArrowBuilder builder(
+            int schemaId,
+            ArrowWriter arrowWriter,
+            AbstractPagedOutputView outputView,
+            boolean appendOnly,
+            LogRecordBatchStatisticsCollector statisticsCollector) {
+        return new MemoryLogRecordsArrowBuilder(
+                BUILDER_DEFAULT_OFFSET,
+                schemaId,
+                CURRENT_LOG_MAGIC_VALUE,
```
wuchong: 1. mark the `statisticsCollector` parameter as `@Nullable`. 
2. we should use version V0 if statisticsCollector is null, and use V1 if is not null. Otherwise, we always use the V0 (`CURRENT_LOG_MAGIC_VALUE` is V0), and the stats is not serialized.

fluss-client/src/main/java/org/apache/fluss/client/write/ArrowLogWriteBatch.java
```diff
@@ -55,11 +56,13 @@ public ArrowLogWriteBatch(
             int schemaId,
             ArrowWriter arrowWriter,
             AbstractPagedOutputView outputView,
-            long createdMs) {
+            long createdMs,
+            LogRecordBatchStatisticsCollector statisticsCollector) {
```
wuchong: Add `@Nullable` annotation.

fluss-common/src/main/java/org/apache/fluss/record/bytesview/MultiBytesView.java
```diff
@@ -105,7 +117,50 @@ public Builder addMemorySegmentByteViewList(List<MemorySegmentBytesView> bytesVi
...
-        /** Adds a bytes section from a range of {@link FileChannel}. */
+        /**
+         * Adds a {@link BytesView} to this builder. If the bytes view is a {@link
+         * FileRegionBytesView} and is continuous with the last added file region view, they will be
+         * merged to improve file read performance.
+         *
+         * @param bytesView the bytes view to add
+         * @return this builder instance for method chaining
+         */
+        public Builder addBytes(BytesView bytesView) {
```
wuchong: not used, can be remove?

fluss-common/src/main/java/org/apache/fluss/config/TableConfig.java
```diff
@@ -154,4 +157,42 @@ public AutoPartitionStrategy getAutoPartitionStrategy() {
     public long getAutoIncrementCacheSize() {
         return config.get(ConfigOptions.TABLE_AUTO_INCREMENT_CACHE_SIZE);
     }
+
+    /** Gets whether statistics collection is enabled for the table. */
+    public boolean isStatisticsEnabled() {
+        String columnsStr = config.get(ConfigOptions.TABLE_STATISTICS_COLUMNS);
+        return !columnsStr.isEmpty();
```
wuchong: Should be `return columnsStr != null && !columnsStr.isEmpty();` if we making `TABLE_STATISTICS_COLUMNS` defaults to `null`.

fluss-common/src/main/java/org/apache/fluss/config/ConfigOptions.java
```diff
@@ -1579,6 +1579,18 @@ public class ConfigOptions {
...
+                    .withDescription(
+                            "Configures statistics collection for the table. "
+                                    + "Empty string ('') means disable statistics collection completely. "
+                                    + "The value '*' (default) means collect statistics for all non-binary columns. "
+                                    + "Comma-separated list of column names means collect statistics only for the specified columns. "
+                                    + "Binary and bytes columns are not supported for statistics collection. "
+                                    + "Example: 'id,name,timestamp' to collect statistics only for specified columns.");
```
wuchong: Could you also update the documentation in Markdown? Additionally, we need to address compatibility concerns. As I understand it, enabling column statistics requires a higher format version, which mandates that downstream jobs be upgraded to the latest version first (Fluss v1.0 or later), otherwise the consumer jobs will fail when parsing V1/V2 format logs. Please ensure this requirement is clearly stated in the documentation.

platinumhamburg:
Thanks to @wuchong  for the detailed and professional review comments. This PR indeed has many areas for improvement, and I have optimized and fixed it according to the feedback. Please continue reviewing when you have time.

wuchong:
```
Warning:  Rule violated for class org.apache.fluss.row.aligned.AlignedRow: lines covered ratio is 0.68, but expected minimum is 0.70
Warning:  Rule violated for class org.apache.fluss.config.StatisticsColumnsConfig: lines covered ratio is 0.55, but expected minimum is 0.70
```

fluss-common/src/main/java/org/apache/fluss/record/DefaultLogRecordBatchStatistics.java
```diff
@@ -0,0 +1,472 @@
...
+        Arrays.fill(this.reversedStatsIndexMapping, -1);
+        for (int statsIndex = 0; statsIndex < statsIndexMapping.length; statsIndex++) {
+            this.reversedStatsIndexMapping[statsIndexMapping[statsIndex]] = statsIndex;
+        }
+    }
+
+    @Override
+    public InternalRow getMinValues() {
+        if (minValuesSize == 0) {
+            return null;
```
wuchong: Since we always serialize the max and min InternalRow for statistics, a `minValuesSize` of 0 is invalid and should be validated within the constructor. This approach simplifies the usage of `getMinValues` by eliminating the need to handle the nullability of the returned value.

platinumhamburg:
> @platinumhamburg I pushed a commit to address the comment , improve some code, and resolve the code coverage problem. Please take a look .

Thanks a lot! This is a really solid improvement.

PR #2890, opened 2026-03-17
[client] Support specifying rack when adding/removing server tags in Admin API

### LINKED ISSUE #2889: [client] Support specifying rack when adding/removing server tags in Admin API
### Search before asking

### Description

**Task Overview:**

Currently, `Admin.addServerTag()` and `Admin.removeServerTag()` only support specifying tablet servers by their server IDs. This task introduces two new overloaded methods that allow users to target all tablet servers belonging to a specific **rack** in a single call.

**Purpose:**

In production environments, the `rack` attribute of a `ServerNode` typically represents a physical machine (or a physical rack of machines). A common operational scenario is decommissioning a physical host — which may run multiple tablet servers — and marking all of them as `PERMANENT_OFFLINE` or `TEMPORARY_OFFLINE` at once.

Without rack-based targeting, operators must:
1. Call `Admin.getServerNodes()` to retrieve all server nodes.
2. Manually filter the results by the target rack.
3. Extract the server IDs.
4. Call `addServerTag()` / `removeServerTag()` with those IDs.

This is repetitive, error-prone, and adds unnecessary boilerplate in automation scripts. A dedicated rack-based API makes the decommissioning workflow concise and intuitive.

### Willingness to contribute

LiebingYu:
### Purpose

Linked issue: close #2889


LiebingYu:
Hi @swuferhong, PTAL!

fluss-client/src/main/java/org/apache/fluss/client/admin/Admin.java
```diff
@@ -654,6 +661,104 @@ ListOffsetsResult listOffsets(
...
+    default CompletableFuture<Void> addServerTagByRack(List<String> racks, ServerTag serverTag) {
+        checkNotNull(racks, "racks must not be null");
+        checkArgument(!racks.isEmpty(), "racks must not be empty");
+        racks.forEach(
+                r ->
+                        checkArgument(
+                                r != null && !r.isEmpty(),
+                                "rack element must not be null or empty"));
+        Set<String> rackSet = new HashSet<>(racks);
+        return getServerNodes()
```
swuferhong: `getServerNodes()` obtains cluster membership from a random TabletServer's metadata cache, which may lag behind the Coordinator. 

Even with a fresh response, resolving the rack and updating the tags in two separate RPCs leaves a race window between the two operations. Please consider resolving the rack membership and updating the tags in a single Coordinator RPC/event so that the documented “all TabletServers belonging to the specified racks” behavior uses a consistent membership snapshot. 

The same issue applies to `removeServerTagByRack()`.

fluss-client/src/main/java/org/apache/fluss/client/admin/Admin.java
```diff
@@ -654,6 +661,104 @@ ListOffsetsResult listOffsets(
...
+     *       permissions.
+     *   <li>{@link ServerTagAlreadyExistException} If the server tag already exists for any one of
+     *       the matched tabletServers, and the server tag is different from the existing one.
+     * </ul>
+     *
+     * @param racks the list of rack identifiers; all tabletServers whose {@link ServerNode#rack()}
+     *     is contained in this list will be targeted. Must not be null or empty.
+     * @param serverTag the server tag to be added.
+     */
+    default CompletableFuture<Void> addServerTagByRack(List<String> racks, ServerTag serverTag) {
```
swuferhong: Why are these two implementations placed directly in the `Admin` interface?

fluss-client/src/main/java/org/apache/fluss/client/admin/Admin.java
```diff
@@ -654,6 +661,104 @@ ListOffsetsResult listOffsets(
...
+                                            .filter(
+                                                    n ->
+                                                            n.serverType()
+                                                                            == ServerType
+                                                                                    .TABLET_SERVER
+                                                                    && rackSet.contains(n.rack()))
+                                            .map(ServerNode::id)
+                                            .collect(Collectors.toList());
+                            if (serverIds.isEmpty()) {
+                                return CompletableFuture.completedFuture(null);
```
swuferhong: This early return skips both serverTag validation and the ALTER authorization performed by `addServerTag()`. As a result, addServerTagByRack(missingRack, null) succeeds, and a caller without ALTER permission also receives success when no rack matches. The outcome therefore depends on whether the rack currently exists and contradicts the documented exception behavior.

Please validate serverTag before resolving the racks and ensure that authorization is checked even for a no-op. 

The corresponding early return in removeServerTagByRack() has the same problem.

fluss-client/src/main/java/org/apache/fluss/client/admin/Admin.java
```diff
@@ -654,6 +661,104 @@ ListOffsetsResult listOffsets(
...
+    default CompletableFuture<Void> removeServerTagByRack(List<String> racks, ServerTag serverTag) {
+        checkNotNull(racks, "racks must not be null");
+        checkArgument(!racks.isEmpty(), "racks must not be empty");
+        racks.forEach(
+                r ->
+                        checkArgument(
+                                r != null && !r.isEmpty(),
+                                "rack element must not be null or empty"));
+        Set<String> rackSet = new HashSet<>(racks);
+        return getServerNodes()
```
swuferhong: `removeServerTagByRack()` only resolves IDs from `getServerNodes()`, which returns the currently alive TabletServers. Once all servers in a rack have gone offline, this method resolves an empty ID list and returns success without removing their tags.

This is asymmetric with `removeServerTag(List<Integer>, ServerTag)`: the Coordinator-side `processRemoveServerTag()` explicitly tolerates a missing server and still removes its tag. Therefore, a tag applied before decommissioning cannot later be cleaned up by rack after the physical host has gone offline; it remains in ZooKeeper and coordinatorContext until the caller removes it explicitly by server ID.

Please preserve enough rack-to-server information to resolve tagged offline servers. Moving rack resolution to the Coordinator is preferable, but it also requires persisting the rack association—for example, storing the rack together with the server tag or retaining historical server metadata—because the live registration is removed when the server goes offline. If this API is intentionally limited to alive servers, that limitation should at least be stated explicitly in its Javadoc and user documentation.

swuferhong (commented):
Hi, @LiebingYu, thanks for your contributions. I left some comments.

fluss-client/src/main/java/org/apache/fluss/client/admin/Admin.java
LiebingYu: Thanks, this is a valid point. I’m intentionally keeping removeServerTagByRack() as a one-shot convenience selector over the Coordinator’s currently registered TabletServers, rather than introducing persistent rack membership history.

Consequently, it does not remove tags from servers that are already offline or no longer belong to their former rack. Those tags remain persisted by server ID and should be cleared with removeServerTag(serverIds, tag), especially before a server ID is reused.

I have documented this limitation explicitly in the JavaDoc and user documentation. The external cordon prevents new servers from entering the rack during decommissioning, but does not change this removal limitation.

If removing offline servers by rack is required as part of the API contract, I think that should be handled by a separate rack-history/policy design rather than adding that state implicitly in this PR.

platinumhamburg (approved):
LGTM +1.

swuferhong (approved):

PR #2900, opened 2026-03-18
[fluss-client] Support Complex Data Types on the Java API (NestedRow/ROW)

### LINKED ISSUE #2805: Support Complex Data Types on the Java Typed API
### Search before asking

### Description

In Fluss 0.9, we introduced complex data types. We have the java typed api, however we haven't added support for complex data types yet. 

https://fluss.apache.org/docs/apis/java-client/#java-typed-api
https://github.com/apache/fluss/blob/main/fluss-client/src/main/java/org/apache/fluss/client/converter/ConverterCommons.java#L46

### Willingness to contribute

XuQianJin-Stars:
### Purpose

Linked issue: close #2805

This pull request extends the Java Typed API to support the `ROW` data type (nested POJO) in POJO ↔ InternalRow conversion, completing the complex type coverage alongside the already supported `ARRAY` and `MAP` types.

### Brief change log

- **`ConverterCommons.java`**: Added `ROW` type validation in `validateCompatibility()` — ensures the POJO field for a ROW type is not a primitive, array, Collection, or Map. Detailed field-level validation is deferred to the nested converter.
- **`PojoToRowConverter.java`**: Added `case ROW:` in `createFieldConverter()` — recursively creates a nested `PojoToRowConverter` to convert a nested POJO field into a `GenericRow`.
- **`RowToPojoConverter.java`**: Added `case ROW:` in `createRowReader()` — recursively creates a nested `RowToPojoConverter` to convert an `InternalRow` back into a nested POJO.
- **`PojoTypeToFlussTypeConverter.java`**: Added `case ROW:` in `convertElementValue()` — supports serializing nested ROW elements within `ARRAY` and `MAP` types.
- **`FlussArrayToPojoArray.java`**: Added `case ROW:` in `buildElementConverter()` — supports deserializing ROW elements within arrays. Falls back to returning raw `InternalRow` when the target POJO type is `Object.class` (due to Java type erasure in MAP values).
- **`java-client.md`**: Added `ROW → Nested POJO` entry to the type mapping table and a new "Nested POJOs (ROW Type)" documentation section with usage examples and a note about the Java type erasure limitation for `MAP<K, ROW>`.

### Tests

- `NestedRowConverterTest#testSimpleNestedRowRoundTrip` — basic POJO with a nested ROW field, roundtrip serialization/deserialization
- `NestedRowConverterTest#testNullNestedRow` — null nested POJO is correctly handled
- `NestedRowConverterTest#testDeeplyNestedRowRoundTrip` — two-level deep nesting (POJO → ROW → ROW)
- `NestedRowConverterTest#testNestedRowWithArrayFieldRoundTrip` — nested POJO containing an `ARRAY<INT>` field
- `NestedRowConverterTest#testNestedRowWithMapFieldRoundTrip` — nested POJO containing a `MAP<STRING, INT>` field
- `NestedRowConverterTest#testArrayOfNestedRowRoundTrip` — top-level `ARRAY<ROW<...>>` mapped to a POJO array
- `NestedRowConverterTest#testMapWithRowValuesRoundTrip` — `MAP<STRING, ROW<...>>` with type erasure fallback (values deserialized as `InternalRow`)
- `NestedRowConverterTest#testRowFieldWithIncompatibleType` — validates that an array type field is rejected when mapped to ROW
- `NestedRowConverterTest#testRowFieldWithMapType` — validates that a Map type field is rejected when mapped to ROW

### API and Format

No. This change only adds internal converter logic for the existing `ROW` data type. No public API or storage format changes.

### Documentation

Updated `website/docs/apis/java-client.md`:
- Added `ROW | Nested POJO` to the type mapping table
- Added a new "Nested POJOs (ROW Type)" section with code examples showing direct ROW fields, `ARRAY<ROW>`, and `MAP<String, ROW>` usage
- Added a note about Java type erasure limitation for MAP values

### DIFF
--- a/fluss-client/src/main/java/org/apache/fluss/client/converter/PojoArrayToFlussArray.java
+++ b/fluss-client/src/main/java/org/apache/fluss/client/converter/PojoArrayToFlussArray.java
@@ -66,6 +66,25 @@ public class PojoArrayToFlussArray {
                         fieldType, componentClass != null ? componentClass : Object.class);
     }
 
+    /**
+     * Creates an adapter using a pre-built row-element converter, avoiding the cost of calling
+     * {@link #buildRowElementConverter} on every row write. The converter must have been
+     * constructed once at field-converter setup time (e.g. in {@link PojoToRowConverter}).
+     *
+     * <p>Passing {@code null} signals that no ROW-element conversion is needed; the generic {@link
+     * PojoTypeToFlussTypeConverter#convertElementValue} path is used for every element instead.
+     */
+    PojoArrayToFlussArray(
+            Object obj,
+            DataType fieldType,
+            String fieldName,
+            Function<Object, Object> prebuiltRowElementConverter) {
+        this.obj = obj;
+        this.fieldType = fieldType;
+        this.fieldName = fieldName;
+        this.rowElementConverter = prebuiltRowElementConverter;
+    }
+
     public GenericArray convertArray() {
         if (obj == null) {
             return null;
@@ -136,13 +155,13 @@ public class PojoArrayToFlussArray {
      * If the array element type is ROW, compile a {@link PojoToRowConverter} and return a function
      * that applies it to every element. Returns {@code null} for non-ROW element types.
      *
-     * <p>When {@code componentClass} is a concrete POJO class the converter is built eagerly,
-     * which also means every element in the array is converted with a consistent converter
-     * regardless of its runtime sub-type. When {@code componentClass} is {@code Object.class} (the
-     * type is not known statically) the converter is built lazily from the first non-null element's
-     * runtime class and stored in an {@link AtomicReference} so the initialisation is thread-safe.
+     * <p>When {@code componentClass} is a concrete POJO class the converter is built eagerly, which
+     * also means every element in the array is converted with a consistent converter regardless of
+     * its runtime sub-type. When {@code componentClass} is {@code Object.class} (the type is not
+     * known statically) the converter is built lazily from the first non-null element's runtime
+     * class and stored in an {@link AtomicReference} so the initialisation is thread-safe.
      */
-    private static Function<Object, Object> buildRowElementConverter(
+    static Function<Object, Object> buildRowElementConverter(
             DataType fieldType, Class<?> componentClass) {
         if (!(fieldType instanceof ArrayType)) {
             return null;
--- a/fluss-client/src/main/java/org/apache/fluss/client/converter/PojoMapToFlussMap.java
+++ b/fluss-client/src/main/java/org/apache/fluss/client/converter/PojoMapToFlussMap.java
@@ -69,6 +69,28 @@ public class PojoMapToFlussMap {
                         mapType.getValueType(), valueClass != null ? valueClass : Object.class);
     }
 
+    /**
+     * Creates an adapter using pre-built row converters for ROW-typed keys and/or values, avoiding
+     * the cost of calling {@link #buildRowConverter} on every row write. Both converters must have
+     * been constructed once at field-converter setup time (e.g. in {@link PojoToRowConverter}).
+     *
+     * <p>A {@code null} converter for key or value means no ROW conversion is required for that
+     * side; the generic {@link PojoTypeToFlussTypeConverter#convertElementValue} path is used
+     * instead.
+     */
+    PojoMapToFlussMap(
+            Map<?, ?> pojoMap,
+            MapType mapType,
+            String fieldName,
+            Function<Object, Object> prebuiltKeyConverter,
+            Function<Object, Object> prebuiltValueConverter) {
+        this.pojoMap = pojoMap;
+        this.mapType = mapType;
+        this.fieldName = fieldName;
+        this.rowKeyConverter = prebuiltKeyConverter;
+        this.rowValueConverter = prebuiltValueConverter;
+    }
+
     public GenericMap convertMap() {
         if (pojoMap == null) {
             return null;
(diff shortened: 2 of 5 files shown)


XuQianJin-Stars:
hi @polyzos @wuchong Hi, i already updated the pr. Please help review when you got some time.

polyzos:
@XuQianJin-Stars Thanks for the contribution! The feature direction is correct, and the read path implementation is clean. A couple of things to address before this is ready to merge.


**`PojoToRowConverter` re-created per element on the write path**

In `PojoTypeToFlussTypeConverter.java`, the `convertElementValue()` ROW case creates a new `PojoToRowConverter` on every call. Since this method is called per element by both `PojoArrayToFlussArray` and `PojoMapToFlussMap`, that means a full converter — including the `PojoType.of()` reflection scan and field-converter setup — is allocated for every single array element or map entry.

The symmetric read path in `FlussArrayToPojoArray.buildElementConverter()` gets this right: it constructs the `RowToPojoConverter` once and captures it in the `ElementConverter` lambda. The write path should follow the same pattern — handle `ARRAY<ROW>` directly in `PojoArrayToFlussArray` at construction time and pre-compile the element converter there, rather than delegating to `convertElementValue()` for ROW elements.


**Silent type-unsafe round-trip for `Map<K, POJO>` where the value is a ROW type**

The `testMapWithRowValuesRoundTrip` test acknowledges that a user who writes `AddressPojo` values into a `Map<String, AddressPojo>` will read back `InternalRow` objects instead. This is a silent, type-unsafe round-trip failure that would be very surprising to users.

The field's declared generic type (`Map<String, AddressPojo>`) carries the value class at compile time and is recoverable via `Field.getGenericType()` at framework setup time. It would be worth exploring whether `PojoType.Property` can carry that generic type parameter and pass a `Class<V>` hint down to `FlussMapToPojoMap`. If that turns out to be genuinely unresolvable, the framework should throw an `UnsupportedOperationException` at converter construction time with a clear message, rather than silently returning an incompatible type at runtime.


**End-to-end tests belong in `FlussTypedClientITCase`**

Rather than introducing a new `NestedRowConverterTest` class, it would be great to validate ROW type support end-to-end in `FlussTypedClientITCase`, which already covers `ARRAY` and `MAP` complex types with full round-trip assertions against a live cluster. The unit-level converter tests can go into the existing `PojoToRowConverterTest` / `RowToPojoConverterTest` structure, with any new POJOs added to `ConvertersTestFixtures`.

Let me know your thoughts

XuQianJin-Stars:
> @XuQianJin-Stars Thanks for the contribution! The feature direction is correct, and the read path implementation is clean. A couple of things to address before this is ready to merge.
> 
> **`PojoToRowConverter` re-created per element on the write path**
> 
> In `PojoTypeToFlussTypeConverter.java`, the `convertElementValue()` ROW case creates a new `PojoToRowConverter` on every call. Since this method is called per element by both `PojoArrayToFlussArray` and `PojoMapToFlussMap`, that means a full converter — including the `PojoType.of()` reflection scan and field-converter setup — is allocated for every single array element or map entry.
> 
> The symmetric read path in `FlussArrayToPojoArray.buildElementConverter()` gets this right: it constructs the `RowToPojoConverter` once and captures it in the `ElementConverter` lambda. The write path should follow the same pattern — handle `ARRAY<ROW>` directly in `PojoArrayToFlussArray` at construction time and pre-compile the element converter there, rather than delegating to `convertElementValue()` for ROW elements.
> 
> **Silent type-unsafe round-trip for `Map<K, POJO>` where the value is a ROW type**
> 
> The `testMapWithRowValuesRoundTrip` test acknowledges that a user who writes `AddressPojo` values into a `Map<String, AddressPojo>` will read back `InternalRow` objects instead. This is a silent, type-unsafe round-trip failure that would be very surprising to users.
> 
> The field's declared generic type (`Map<String, AddressPojo>`) carries the value class at compile time and is recoverable via `Field.getGenericType()` at framework setup time. It would be worth exploring whether `PojoType.Property` can carry that generic type parameter and pass a `Class<V>` hint down to `FlussMapToPojoMap`. If that turns out to be genuinely unresolvable, the framework should throw an `UnsupportedOperationException` at converter construction time with a clear message, rather than silently returning an incompatible type at runtime.
> 
> **End-to-end tests belong in `FlussTypedClientITCase`**
> 
> Rather than introducing a new `NestedRowConverterTest` class, it would be great to validate ROW type support end-to-end in `FlussTypedClientITCase`, which already covers `ARRAY` and `MAP` complex types with full round-trip assertions against a live cluster. The unit-level converter tests can go into the existing `PojoToRowConverterTest` / `RowToPojoConverterTest` structure, with any new POJOs added to `ConvertersTestFixtures`.
> 
> Let me know your thoughts

Thanks for the detailed review! All three points make sense to me:

1. PojoToRowConverter re-created per element — Agreed, I'll refactor PojoArrayToFlussArray (and PojoMapToFlussMap) to pre-compile the element converter at construction time for ARRAY<ROW> / MAP<K, ROW>, following the same pattern as FlussArrayToPojoArray.buildElementConverter().
2. Silent type-unsafe round-trip for Map<K, POJO> — Good catch. I'll look into extracting the value class from Field.getGenericType() and passing it through PojoType.Property down to FlussMapToPojoMap. If the generic type turns out to be unresolvable, I'll make it fail fast with an UnsupportedOperationException at converter construction time instead of silently returning InternalRow.
3. Test placement — Will do. I'll move the end-to-end round-trip tests into FlussTypedClientITCase, keep the unit-level converter tests in PojoToRowConverterTest / RowToPojoConverterTest, and add the new POJOs to ConvertersTestFixtures. The standalone NestedRowConverterTest will be removed.

XuQianJin-Stars:
hi @polyzos @wuchong Hi, i already updated the pr. Please help review when you got some time.

PR #2907, opened 2026-03-20
[client][fix] Deduplicate insertIfNotExists lookups with same key in same batch to prevent deadlock

### LINKED ISSUE #2874: [test] Unstable test Flink119TableSourceITCase.testLookupInsertIfNotExists
https://github.com/apache/fluss/actions/runs/23095392185/job/67086933773

```
[ERROR] org.apache.fluss.flink.source.Flink119TableSourceITCase.testLookupInsertIfNotExists(boolean)[1]  Time elapsed: 61.483 s  <<< FAILURE!
java.lang.AssertionError: 
```

pbanakar:
### Purpose

Linked issue: close #2874

When `lookup.insert-if-not-exists=true` and `lookup.async=true`, multiple lookups for the same key can land in the same `LookupBatch` within a single drain cycle. The server acquires a row-level latch per key during insert — 
two entries for the same key in one RPC batch cause a deadlock in the server-side handler, leaving `LookupQuery` futures never completed and hanging the `AsyncWaitOperator` indefinitely, resulting in a 60s timeout.

This did not affect `async=false` because sync lookups execute sequentially via `.get()`, so the same key never appears twice in one drain cycle.

### Brief change log

- In `LookupSender.sendLookupRequest`, when `insertIfNotExists=true`, duplicate lookups for the same key bytes within the same batch are detected using `ByteBuffer` key comparison. The duplicate's future is chained to the first lookup's future instead of being added to the RPC batch, ensuring  only one server-side entry is sent per unique key per drain cycle.
### Tests

- `Flink119TableSourceITCase#testLookupInsertIfNotExists` (existing test, previously flaky with `async=true`, now passes consistently with  `-Dsurefire.rerunFailingTestsCount=5`)
### API and Format

### Documentation

No new feature introduced. Bug fix only.

### DIFF
--- a/fluss-client/src/main/java/org/apache/fluss/client/lookup/LookupSender.java
+++ b/fluss-client/src/main/java/org/apache/fluss/client/lookup/LookupSender.java
@@ -44,6 +44,7 @@ import org.slf4j.LoggerFactory;
 
 import javax.annotation.Nullable;
 
+import java.nio.ByteBuffer;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HashMap;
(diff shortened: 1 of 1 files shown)


pbanakar:
@luoyuxia could you please help review this?

PR #2917, opened 2026-03-23
[server] Add retract support for Aggregation Merge Engine

### LINKED ISSUE #2916: [KV] Add retract support for Aggregation Merge Engine
### Search before asking

### Motivation

The Aggregation Merge Engine currently supports upsert and deleteoperations. However, there is no way to **reverse a previously applied aggregation contribution** without deleting the entire row.

Consider an e-commerce scenario: a table aggregates total revenue per shop using `SUM`. When a refund occurs, the system needs to subtract the refunded amount from the running total. Today, this requires the Flink compute layer to maintain full aggregation state — defeating the purpose of pushing aggregation down to the storage engine.

### Solution

Introduce a **retract** operation that computes the inverse of a previous aggregation at the storage layer:
  - `SUM`: subtracts the retracted value
  - `PRODUCT`: divides by the retracted value
  - `LAST_VALUE` / `LAST_VALUE_IGNORE_NULLS`: clears to null Functions that cannot be mathematically reversed (`MAX`, `MIN`, `FIRST_VALUE`, `LISTAGG`, etc.) do not support retract.

### Anything else?

_No response_

### Willingness to contribute

platinumhamburg:
### Purpose

Linked issue: close #2916

### DIFF
--- a/fluss-client/src/main/java/org/apache/fluss/client/metadata/MetadataUpdater.java
+++ b/fluss-client/src/main/java/org/apache/fluss/client/metadata/MetadataUpdater.java
@@ -39,6 +39,7 @@ import org.apache.fluss.rpc.RpcClient;
 import org.apache.fluss.rpc.gateway.AdminReadOnlyGateway;
 import org.apache.fluss.rpc.gateway.CoordinatorGateway;
 import org.apache.fluss.rpc.gateway.TabletServerGateway;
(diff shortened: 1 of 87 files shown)


ThorneANN:
It is a great work

ThorneANN:
hello,is there any progress ?

PR #2920, opened 2026-03-24
[lake/tiering] add table dropped handling and nullable write results

### LINKED ISSUE #2498: TieringService will stuck when drop the table that is tiering
### Search before asking

### Description

1: tiering service start to tiering this table
2: user drop the table
3: the scanner won't fail and log warning forever: 
```
Received unknown table or bucket error in fetch for bucket TableBucket{tableId=29, bucket=0}
```

The coordinator server should tell the tiering service that the table is tiering is dropped via heartbeat so that it can cancle the tiering for this table

### Willingness to contribute

beryllw:
### Purpose

Linked issue: close #2498

This PR improves Lake Tiering's handling of dropped tables and adds cancellation support for tiering operations.

### Brief change log

Dropped Table Handling
- When a table is dropped during active tiering, the TieringSplitReader now gracefully completes all
in-progress splits with empty results instead of failing
- Added handleTableDropped() to mark dropped tables and trigger force completion
- Dropped tables are properly cleaned up from LakeTableTieringManager to prevent resource leaks

Cancellation Support
- Added cancelled flag to TableBucketWriteResult to distinguish between normal and cancelled tiering rounds
- Committers skip commit processing for cancelled results and report back to the coordinator
- Lake writers are closed without completing when tiering is cancelled, discarding uncommitted data


fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/tiering/source/TableBucketWriteResult.java
leekeiabstraction, line 36: As `canceled` field is added, shouldn't the `serialVersionUID` be bumped as well? Similarly the serialiser needs to handle each version of TableBucketWriteResult differently.

leekeiabstraction (commented):
TY for the PR, left a comment.

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/tiering/source/TieringSplitReader.java
```diff
@@ -143,6 +146,13 @@ public RecordsWithSplitIds<TableBucketWriteResult<WriteResult>> fetch() throws I
...
             return records;
         }
+
+        // Check droppedTables BEFORE checkSplitOrStartNext to quickly respond to already-marked
+        // current table
+        if (currentTableId != null && droppedTables.contains(currentTableId)) {
+            return forceCompleteDroppedTable();
+        }
+
         checkSplitOrStartNext();
```
luoyuxia, line 162: if move to a split, and then the split is for the table dropped, will `getOrMoveToTable` cause exception and then fail the job. Maybe we can protect this case .

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/tiering/source/TieringSplitReader.java
```diff
@@ -164,7 +174,26 @@ public RecordsWithSplitIds<TableBucketWriteResult<WriteResult>> fetch() throws I
                 if (reachTieringMaxDurationTables.contains(currentTableId)) {
                     return forceCompleteTieringLogRecords();
                 }
-                ScanRecords scanRecords = currentLogScanner.poll(pollTimeout);
+                ScanRecords scanRecords;
+                try {
+                    scanRecords = currentLogScanner.poll(pollTimeout);
+                } catch (TableNotExistException e) {
```
luoyuxia: IIUC, we don't need to check this exception, just wait it to be forced complete in next fetch.
Also, as disscuss before, poll won't throw `TableNotExistException`.

luoyuxia (commented):
Haven't finish review, but left some comments

fluss-common/src/main/java/org/apache/fluss/lake/writer/LakeWriter.java
```diff
@@ -43,8 +45,9 @@ public interface LakeWriter<WriteResult> extends Closeable {
     /**
      * Completes the writing process and returns the write result.
      *
-     * @return the write result
+     * @return the write result, or null if no data was written (empty write scenario)
```
luoyuxia: I'm curious about when it'll be null(not data written)

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/tiering/source/enumerator/TieringSourceEnumerator.java
```diff
@@ -427,16 +432,76 @@ private void assignSplits() {
...
+     */
+    private void handleTieringTableResponseErrors(LakeTieringHeartbeatResponse heartbeatResponse) {
+        for (PbHeartbeatRespForTable resp : heartbeatResponse.getTieringTableRespsList()) {
+            if (resp.hasError()) {
+                ApiError error = ApiError.fromErrorMessage(resp.getError());
+                Errors errors = error.error();
+                // Check if the error indicates table doesn't exist or tiering epoch is fenced
+                // (which happens when table is dropped and recreated)
+                if (errors == Errors.TABLE_NOT_EXIST
+                        || errors == Errors.UNKNOWN_TABLE_OR_BUCKET_EXCEPTION
```
luoyuxia: not to handle `UNKNOWN_TABLE_OR_BUCKET_EXCEPTION` and `FENCED_TIERING_EPOCH_EXCEPTION`

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/tiering/source/enumerator/TieringSourceEnumerator.java
```diff
@@ -427,16 +432,76 @@ private void assignSplits() {
...
+    }
+
+    /**
+     * Handle a dropped table by marking all related splits to skip, removing from tiering epochs,
+     * and notifying readers.
+     */
+    @VisibleForTesting
+    protected void handleTableDropped(long tableId) {
+        // Remove from tiering table epochs
+        Long tieringEpoch = tieringTableEpochs.remove(tableId);
```
luoyuxia: I feel like it's error-prone in here to remove from tieringTableEpochs, and put `failedTableEpochs`.
Another way is consider table drop as another fail tiering event (send FailedTieringEvent),  so that the whole path is unified.

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/tiering/source/TableBucketWriteResultSerializer.java
```diff
@@ -33,7 +33,10 @@ public class TableBucketWriteResultSerializer<WriteResult>
     private static final ThreadLocal<DataOutputSerializer> SERIALIZER_CACHE =
             ThreadLocal.withInitial(() -> new DataOutputSerializer(64));
 
-    private static final int CURRENT_VERSION = 1;
+    private static final int CURRENT_VERSION = 2;
```
luoyuxia: Acutally, we may not need to bump the version since it's only needed for shuffle... For code clean, we can still keep it as `CURRENT_VERSION = 1`

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/tiering/source/TieringSourceFetcherManager.java
```diff
@@ -55,41 +55,61 @@ public TieringSourceFetcherManager(
...
-                        ((TieringSplitReader<WriteResult>) splitFetcher.getSplitReader())
-                                .handleTableReachTieringMaxDuration(reachTieringDeadlineTable);
+                        action.accept(
+                                (TieringSplitReader<WriteResult>) splitFetcher.getSplitReader());
                         return true;
                     }
 
                     @Override
-                    public void wakeUp() {
-                        // do nothing
```
luoyuxia, line 91: please still keep the comment

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/tiering/source/TieringSplitReader.java
```diff
@@ -570,6 +601,116 @@ public void handleTableReachTieringMaxDuration(long tableId) {
...
+        for (Map.Entry<TableBucket, LakeWriter<WriteResult>> entry : lakeWriters.entrySet()) {
+            try {
+                entry.getValue().close();
+            } catch (Exception e) {
+                LOG.warn("Failed to close orphan lake writer for bucket {}", entry.getKey(), e);
+            }
+        }
+        lakeWriters.clear();
+
+        // Also handle pending snapshot splits for this table
```
luoyuxia: we don't need to handle pending snapshot splits again since it must be in `currentTableSplitsByBucket`. See method 
```
private void addSplitToCurrentTable(TieringSplit split) {
        this.currentTableSplitsByBucket.put(split.getTableBucket(), split);
        if (split.isTieringSnapshotSplit()) {
            this.currentPendingSnapshotSplits.add((TieringSnapshotSplit) split);
        } else if (split.isTieringLogSplit()) {
            subscribeLog((TieringLogSplit) split);
        }
    }
```

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/tiering/source/TieringSplitReader.java
```diff
@@ -570,6 +601,116 @@ public void handleTableReachTieringMaxDuration(long tableId) {
...
+                writeResults.put(bucket, bucketResult);
+                finishedSplitIds.put(bucket, split.splitId());
+                LOG.info(
+                        "Split {} is forced to be finished due to table dropped with empty result.",
+                        split.splitId());
+                splitsIterator.remove();
+            }
+        }
+
+        // Close any remaining lake writers that don't have corresponding splits
```
luoyuxia: do we really need it?
When it happen that we have lake writer but no corresponding splits?

fluss-lake/fluss-lake-paimon/src/main/java/org/apache/fluss/lake/paimon/tiering/PaimonLakeWriter.java
```diff
@@ -84,6 +87,10 @@ public PaimonWriteResult complete() throws IOException {
         } catch (Exception e) {
             throw new IOException("Failed to complete Paimon write.", e);
         }
+        if (commitMessage == null) {
```
luoyuxia: iceberg and lance also need to do same handle?

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/tiering/source/TieringSplitReader.java
```diff
@@ -251,6 +260,25 @@ private void checkSplitOrStartNext() {
...
+        if (droppedTables.contains(pendingTableId)) {
+            TieringSplit firstSplit = pendingSplits.iterator().next();
+            currentTableId = pendingTableId;
+            currentTablePath = firstSplit.getTablePath();
+            currentTableNumberOfSplits = firstSplit.getNumberOfSplits();
+            for (TieringSplit split : pendingSplits) {
+                currentTableSplitsByBucket.put(split.getTableBucket(), split);
+            }
+            LOG.info(
+                    "Skipping RPC for dropped table {} (path: {}), will force complete in next fetch cycle.",
```
luoyuxia: nit:

```suggestion
                    Table {} is already marked dropped; skip opening table {} and force complete in next fetch cycle.
```
`Skipping RPC` looks strange to me

luoyuxia (commented):
@beryllw Thanks for the pr. Left some comments. PTAL

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/tiering/committer/TieringCommitOperator.java
```diff
@@ -151,6 +151,27 @@ public void processElement(StreamRecord<TableBucketWriteResult<WriteResult>> str
...
+                        tableId,
+                        tableBucketWriteResult.tablePath());
+                collectedTableBucketWriteResults.remove(tableId);
+                // Notify the enumerator that this table's tiering is cancelled (dropped)
+                // via FailedTieringEvent with TABLE_DROPPED type.
+                operatorEventGateway.sendEventToCoordinator(
+                        new SourceEventWrapper(
+                                new FailedTieringEvent(
+                                        tableId,
+                                        FailedTieringEvent.FailureType.TABLE_DROPPED,
```
luoyuxia, line 174: isCancelled and TABLE_DROPPED are not the same thing semantically. 
isCancelled describes the state of the write result, while TABLE_DROPPED is a specific cancellation reason.

The current code upgrades any cancelled result into FailedTieringEvent(TABLE_DROPPED), which makes the control-flow reason more specific than the data it is derived from.

A simpler fix here might be to rename cancelled to something reason-specific like tableDropped / isTableDropped().  My main concern is that cancelled is a generic state, while TABLE_DROPPED is a specific cause, so the current mapping feels semantically too broad.

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/tiering/event/FailedTieringEvent.java
```diff
@@ -19,25 +19,48 @@
...
+    public FailureType getFailureType() {
+        return failureType;
+    }
+
+    public String getFailureMessage() {
+        return failureMessage;
+    }
+
+    public boolean isCancelled() {
+        return failureType == FailureType.TABLE_DROPPED;
```
luoyuxia, line 64: dito. The isCancelled and  failureType == TABLE_DROPPED are not the same thing semantically.

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/tiering/source/enumerator/TieringSourceEnumerator.java
```diff
@@ -284,18 +305,23 @@ public void handleSourceEvent(int subtaskId, SourceEvent sourceEvent) {
         if (sourceEvent instanceof FailedTieringEvent) {
             FailedTieringEvent failedEvent = (FailedTieringEvent) sourceEvent;
             long failedTableId = failedEvent.getTableId();
-            Long tieringEpoch = tieringTableEpochs.remove(failedTableId);
             LOG.info(
                     "Tiering table {} is failed, fail reason is {}.",
                     failedTableId,
-                    failedEvent.failReason());
+                    failedEvent.getFailureMessage());
+            Long tieringEpoch = markTableAsFailed(failedTableId);
```
luoyuxia: why extract `markTableAsFailed` into a method, I just feel extract a method is not necessary and make harder to track the code logic

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/tiering/source/enumerator/TieringSourceEnumerator.java
```diff
@@ -284,18 +305,23 @@ public void handleSourceEvent(int subtaskId, SourceEvent sourceEvent) {
...
             if (tieringEpoch == null) {
-                // shouldn't happen, warn it
-                LOG.warn(
-                        "The failed table {} is not in tiering table, won't report it to Fluss to mark as failed.",
-                        failedTableId);
-            } else {
-                failedTableEpochs.put(failedTableId, tieringEpoch);
+                if (!failedEvent.isCancelled()) {
```
luoyuxia: This branch looks over-specialized for logging only. We could simplify it by using a single log path when tieringEpoch == null, instead of distinguishing cancelled vs non-cancelled events here for simplity

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/tiering/source/enumerator/TieringSourceEnumerator.java
```diff
@@ -357,6 +383,16 @@ private void generateAndAssignSplits(
         if (throwable != null) {
             LOG.warn("Failed to request tiering table, will retry later.", throwable);
         }
+
+        // Process dropped table IDs collected during heartbeat processing.
+        if (!pendingDroppedTableIds.isEmpty()) {
+            List<Long> droppedIds = new ArrayList<>(pendingDroppedTableIds);
+            pendingDroppedTableIds.clear();
+            for (long droppedTableId : droppedIds) {
+                handleTableDropped(droppedTableId);
```
luoyuxia: It looks strange to me that `handleTableDropped` is called in `generateAndAssignSplits`. Can we decouple it with method `generateAndAssignSplits`? Once found a tiered table dropped, then handle it with `context.runInCoordinatorThread`

fluss-lake/fluss-lake-paimon/src/main/java/org/apache/fluss/lake/paimon/tiering/RecordWriter.java
```diff
@@ -70,8 +70,18 @@ public RecordWriter(
...
+     * Completes the write process and returns the commit message.
+     *
+     * @return the commit message, or null if no data was written (empty write scenario)
+     */
+    @Nullable
     CommitMessage complete() throws Exception {
         List<CommitMessage> commitMessages = tableWrite.prepareCommit();
+        if (commitMessages.isEmpty()) {
+            // No data was written, return null to indicate empty write
+            return null;
```
luoyuxia: When will it return null? If `CommitMessage` is Nullable, we will need to also mark the `commitMessage` in PaimonWriteResult as Nullable. I think we can revert this change.

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/tiering/committer/TieringCommitOperator.java
```diff
@@ -151,6 +151,27 @@ public void processElement(StreamRecord<TableBucketWriteResult<WriteResult>> str
                 collectTableAllBucketWriteResult(tableId);
 
         if (committableWriteResults != null) {
+            // Check if any result is cancelled (table was dropped)
+            boolean isCancelled =
```
luoyuxia: actually, we will also need to abort the committable to clean the staging data files. Not sure whether the change is big or not. If it's log, maybe you can left a todo

luoyuxia (commented):
@beryllw Thanks for the pr. Left some comments. Also, could you please add a IT to cover it just like TieringITCase#testTieringReachMaxDuration?

binary-signal:
maybe related #3483

PR #2924, opened 2026-03-25
[server] Allow enabling lakehouse on tables created before cluster-level lakehouse is enabled

### LINKED ISSUE #2908: Allow enabling lakehouse on tables created before cluster-level lakehouse is enabled
## Search before asking

## Motivation
Today, enabling lakehouse for an existing table only works reliably if the table was created after the cluster had already enabled datalake support. Here, "enabled datalake support" means the cluster had already configured `datalake.format`; in the current behavior, setting `datalake.format` is treated as enabling datalake support. This causes a compatibility problem for the following user flow:

1. Create a Fluss table when the cluster has not explicitly enabled lakehouse.
2. Later configure the cluster to enable lakehouse.
3. Enable lakehouse for the existing table.

At the moment, step 3 fails for tables created before cluster-level lakehouse was enabled. The root issue is that `datalake.format` currently serves two roles at the same time:

- selecting the lake-format-specific bucketing / key-encoding behavior; and
- indicating that the cluster is ready to create and manage lake tables.

This makes the semantics unclear for new deployments that want to pre-bind the future lake format (for example Paimon, so that bucketing stays consistent) but do not want users to enable lakehouse for tables until the cluster is explicitly switched on.

We need a backward-compatible way to separate "legacy cluster behavior" from "new cluster behavior", while still allowing tables created before `table.datalake.enabled=true` to be enabled later if their bucketing format is already predetermined.

## Solution
Introduce a new cluster config `datalake.enabled` with compatibility semantics:

- `datalake.enabled` is **unset**: treat the cluster as a legacy cluster and keep the current behavior unchanged.
- `datalake.enabled=false`: treat the cluster as a new-style cluster in "pre-bind only" mode.
- `datalake.enabled=true`: treat the cluster as a new-style cluster with lakehouse fully enabled.

For clusters where `datalake.enabled` is explicitly configured (either `true` or `false`):

- require `datalake.format` to be configured;
- automatically persist `table.datalake.format=<cluster datalake.format>` into newly created tables;
- when `datalake.enabled=false`, do **not** allow creating/enabling lake tables yet;
- when `datalake.enabled=true`, allow `ALTER TABLE ... SET ('table.datalake.enabled'='true')` for tables whose `table.datalake.format` already matches the cluster `datalake.format`.

This keeps old clusters fully compatible while enabling the desired flow for new clusters:

1. Create cluster with `datalake.enabled=false` and `datalake.format=paimon`.
2. Create table; Fluss auto-persists `table.datalake.format=paimon`, so writes already follow Paimon bucketing.
3. Later switch cluster to `datalake.enabled=true`.
4. Enable lakehouse for the existing table successfully.

Suggested validation rules:

- If `datalake.enabled` is explicitly set but `datalake.format` is missing, fail fast.
- If a table has no persisted `table.datalake.format`, keep rejecting later lakehouse enablement to avoid bucket inconsistency.
- If a table's `table.datalake.format` differs from the cluster `datalake.format`, reject enablement.
- In new-style clusters, `datalake.format` should be treated as immutable (or at least strongly restricted) once tables have been created with the pre-bound format.

Affected areas likely include:

- cluster config parsing / compatibility checks;
- `CoordinatorService.applySystemDefaults(...)`;
- `LakeCatalogDynamicLoader` load conditions;
- alter-table validation for `table.datalake.enabled`.

## Anything else?
This issue is mainly about compatibility and semantic clarity:

- old clusters should continue to behave exactly as they do today;
- new clusters should be able to pre-bind lake-format bucketing without exposing lakehouse functionality too early;
- users should be able to create a table first, enable cluster lakehouse later, and then enable lakehouse on that table successfully.

## Willingness to contribute

luoyuxia:
### Purpose

Linked issue: close #2908

This change allows lakehouse to be enabled only after the cluster-level lakehouse binding is turned on, while keeping compatibility for clusters and tables created before the new configuration semantics.

### Brief change log

- add cluster config `datalake.enabled` and define `legacy`, `pre-bind`, and `fully-enabled` semantics
- update dynamic lake catalog loading so explicit `datalake.enabled=true` requires `datalake.format`, while `false` skips creating `LakeCatalog` and still tolerates legacy clusters that delete `datalake.format`
- persist `table.datalake.format` when creating tables and validate both `CREATE TABLE` and `ALTER TABLE ... SET ('table.datalake.enabled'='true')` against the cluster datalake state and format
- reject enabling lakehouse for old tables that do not have a persisted table-level datalake format, and reject format mismatches after cluster binding is enabled

### Tests

- `./mvnw -pl fluss-server,fluss-client -am -DskipITs -DfailIfNoTests=false -Dtest=DynamicConfigChangeTest,LakeEnableTableITCase test`

### API and Format

- adds cluster config `datalake.enabled`
- persists `table.datalake.format` for newly created lakehouse tables
- does not change existing storage format, but tightens validation for enabling lakehouse on pre-existing tables

### Documentation

- no

### DIFF
--- a/fluss-server/src/main/java/org/apache/fluss/server/coordinator/MetadataManager.java
+++ b/fluss-server/src/main/java/org/apache/fluss/server/coordinator/MetadataManager.java
@@ -540,10 +540,7 @@ public class MetadataManager {
                 }
 
                 // reuse the same validate logic with the createTable() method
-                TableDescriptorValidation.validateTableDescriptor(
-                        newDescriptor,
-                        maxBucketNum,
-                        lakeCatalogDynamicLoader.getLakeCatalogContainer().getDataLakeFormat());
+                validateTableDescriptor(newDescriptor);
                 // pre alter table properties, e.g. create lake table in lake storage if it's to
                 // enable datalake for the table
                 preAlterTableProperties(


luoyuxia:
@wuchong Could you please help review when you got some time

website/docs/maintenance/operations/upgrade-notes-0.10.md
```diff
@@ -4,3 +4,41 @@ sidebar_position: 4
 ---
 
 # Upgrade Notes from v0.9 to v0.10
+
+## Cluster Configuration Changes
+
+### New `datalake.enabled` Cluster Configuration
+
+Starting in v0.10, Fluss introduces the cluster-level configuration `datalake.enabled` to control whether the cluster is ready to create and manage lakehouse tables.
```
wuchong: We have updated the next version to v1.0, please update here as well. Please also update the title and markdown file name.

fluss-server/src/main/java/org/apache/fluss/server/coordinator/MetadataManager.java
```diff
@@ -368,7 +368,10 @@ public long createTable(
             boolean ignoreIfExists)
             throws TableAlreadyExistException, DatabaseNotExistException {
         // validate table properties before creating table
-        validateTableDescriptor(tableToCreate, maxBucketNum);
+        validateTableDescriptor(
+                tableToCreate,
+                maxBucketNum,
+                lakeCatalogDynamicLoader.getLakeCatalogContainer().getDataLakeFormat());
```
wuchong: [P2] Validate lake-format mismatches before creating the lake table

For `CREATE TABLE` requests where `table.datalake.enabled=true` and `table.datalake.format` differs from the cluster `datalake.format`, `CoordinatorService#createTable()` has already called `lakeCatalog.createTable(...)` before this new validation runs. The request then fails here with `InvalidConfigException`, but the external lake table has already been created while Fluss metadata is not, leaving an orphan that must be cleaned up manually.

Maybe we should move the `validateTableDescriptor` before `metadataManager.createTable` in `CoordinatorService#createTable`?

website/docs/maintenance/operations/upgrade-notes-0.10.md
```diff
@@ -4,3 +4,41 @@ sidebar_position: 4
...
+
+#### Behavior Changes
+
+- If `datalake.enabled` is unset, Fluss preserves the legacy behavior: configuring `datalake.format` alone also enables lakehouse tables.
+- If `datalake.enabled = false`, Fluss pre-binds the lake format for newly created tables but does not allow lakehouse tables yet.
+- If `datalake.enabled = true`, Fluss fully enables lakehouse tables.
+- If `datalake.enabled` is explicitly set to `true`, `datalake.format` must also be configured. When `datalake.enabled = false`, `datalake.format` is optional, but required if you want the pre-bind behavior described above.
```
wuchong: I improved this with AI, to make the meaning of different combination more clear. You can take this as an example:

```
The behavior of Fluss regarding lakehouse table configuration is determined by the combination of `datalake.enabled` and `datalake.format`. The specific rules are as follows:

If `datalake.enabled` is unset, Fluss defaults to legacy behavior. In this state, configuring `datalake.format` alone automatically enables lakehouse tables.

If `datalake.enabled` is set to `false`, lakehouse functionality remains disabled. The `datalake.format` parameter is optional in this scenario. When explicitly configured, it pre-binds the specified lake format to newly created tables, preparing them for future integration without immediately activating lakehouse tables.

If `datalake.enabled` is set to `true`, lakehouse functionality is fully enabled. In this state, `datalake.format` is strictly required and must be provided for the configuration to take effect.
```

website/docs/quickstart/lakehouse.md
```diff
@@ -142,6 +143,8 @@ services:
         s3.path-style-access: true
         s3.assumed.role.arn: arn:aws:iam::000000000000:role/rustfsadmin
         s3.assumed.role.sts.endpoint: http://rustfs:9000
+        kv.snapshot.interval: 0s
```
wuchong: This is intended to be removed?

website/docs/maintenance/operations/updating-configs.md
```diff
@@ -27,13 +28,14 @@ Here is a code snippet to demonstrate how to update the cluster configurations u
 ```java
```
wuchong, line 31: Please help to add a SQL tab for using CALL procedure to update cluster configs: 
https://fluss.apache.org/docs/engine-flink/procedures/#set_cluster_configs

website/docs/maintenance/operations/upgrade-notes-0.10.md
luoyuxia: Thanks. It looks much better.

wuchong (approved):
LGTM.

PR #2926, opened 2026-03-25
[hotfix] Fix Netty ByteBuf leak on follower fetchLog timeout

### LINKED ISSUE #2925: [Server] Netty ByteBuf leak in ReplicaFetcherThread when fetchLog RPC times out
### Search before asking

### Fluss version

0.9.0 (latest release)

### Please describe the bug 🐞

When a follower's ReplicaFetcherThread sends a fetchLog RPC to the leader and the call times out (via CompletableFuture.get(timeout, TimeUnit)), the pooled ByteBuf held by the late-arriving FetchLogResponse is never released. This causes Netty direct memory to grow indefinitely and never self-heal.

Impact:
  - Each timed-out fetchLog RPC leaks one pooled ByteBuf
  - Under normal operation with occasional timeouts: slow leak
  - During leader OOM/restart causing mass timeouts: fast leak
  - Leaked buffers keep Netty PooledByteBufAllocator chunks alive permanently
  - Eventually triggers Old GC pressure when direct memory grows large, leading to a GC death spiral on affected nodes

### Solution

_No response_

platinumhamburg:
When ReplicaFetcherThread's fetchLog RPC times out via .get(timeout), the late-arriving FetchLogResponse's pooled ByteBuf is never released, causing Netty direct memory to grow indefinitely.

Register a thenAccept callback on the CompletableFuture in the catch block to release the ByteBuf when the late response eventually arrives.

### Purpose

Linked issue: close #2925 2925

### DIFF
--- a/fluss-server/src/test/java/org/apache/fluss/server/replica/fetcher/ReplicaFetcherThreadTest.java
+++ b/fluss-server/src/test/java/org/apache/fluss/server/replica/fetcher/ReplicaFetcherThreadTest.java
@@ -427,27 +427,30 @@ public class ReplicaFetcherThreadTest {
             // then the delayed responses arrive after 3s
             timeoutFetcher.start();
 
-            // Wait for at least one timeout + delayed response cycle to complete
-            Thread.sleep(5000);
+            // Wait until at least one delayed response has been allocated
+            retry(
+                    Duration.ofSeconds(10),
+                    () -> assertThat(testingEndpoint.getAllAllocatedByteBufs()).isNotEmpty());
 
             // Shutdown the fetcher to stop new requests
             timeoutFetcher.shutdown();
 
-            // Wait a bit more for any remaining delayed responses to arrive and be cleaned up
-            Thread.sleep(2000);
-
-            // Verify: ALL pooled ByteBufs should have been released by the late-arrival callback
-            java.util.List<ByteBuf> allBufs = testingEndpoint.getAllAllocatedByteBufs();
-            assertThat(allBufs).isNotEmpty();
-            for (int i = 0; i < allBufs.size(); i++) {
-                ByteBuf buf = allBufs.get(i);
-                assertThat(buf.refCnt())
-                        .as(
-                                "Pooled ByteBuf #%d should be released after fetch timeout. "
-                                        + "refCnt > 0 means the buffer leaked.",
-                                i)
-                        .isEqualTo(0);
-            }
+            // Wait until ALL allocated ByteBufs have been released (refCnt == 0).
+            // The thenAccept callback releases them when late responses arrive.
+            retry(
+                    Duration.ofSeconds(15),
+                    () -> {
+                        java.util.List<ByteBuf> allBufs = testingEndpoint.getAllAllocatedByteBufs();
+                        for (int i = 0; i < allBufs.size(); i++) {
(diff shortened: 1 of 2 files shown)


fluss-server/src/test/java/org/apache/fluss/server/replica/fetcher/ReplicaFetcherThreadTest.java
```diff
@@ -374,6 +376,83 @@ tb, genMemoryLogRecordsWithWriterId(DATA1, writerId, 2, 0)),
...
+            timeoutFetcher.start();
+
+            // Wait for at least one timeout + delayed response cycle to complete
+            Thread.sleep(5000);
+
+            // Shutdown the fetcher to stop new requests
+            timeoutFetcher.shutdown();
+
+            // Wait a bit more for any remaining delayed responses to arrive and be cleaned up
+            Thread.sleep(2000);
```
wuchong: On slower CI hosts, or if the fetcher starts one more delayed request just before `shutdown()`, these fixed `Thread.sleep(5000)` / `Thread.sleep(2000)` calls do not guarantee that every scheduled late response has completed before the assertions run. In that case `allBufs` can still include an in-flight buffer with `refCnt() == 1`, so this new regression test will fail nondeterministically even when the production fix is correct.

wuchong:
The CI compile is failed.

fluss-server/src/test/java/org/apache/fluss/server/replica/fetcher/TestingLeaderEndpoint.java
```diff
@@ -51,6 +55,21 @@ public class TestingLeaderEndpoint implements LeaderEndpoint {
...
+    /**
+     * If set, fetchLog will return a future that completes after the specified delay, simulating a
+     * slow leader. The response will carry a pooled ByteBuf to track buffer release.
+     */
+    private volatile ScheduledExecutorService delayExecutor;
+
+    private volatile long delayMs;
+
+    /** The last ByteBuf allocated for a delayed fetch response, for leak detection in tests. */
+    private volatile ByteBuf lastAllocatedByteBuf;
```
wuchong: This is not used, can be removed.

fluss-server/src/test/java/org/apache/fluss/server/replica/fetcher/ReplicaFetcherThreadTest.java
platinumhamburg: Yes, it's a flaky case. Now I replaced Thread.sleep with retry() condition-based waiting — the test now polls until all allocated ByteBufs reach refCnt == 0 (with a 15s timeout), so it won't fail due to timing on slow CI hosts. See the follow-up commit.

fluss-server/src/test/java/org/apache/fluss/server/replica/fetcher/TestingLeaderEndpoint.java
platinumhamburg: Removed.

platinumhamburg:
> The CI compile is failed.

The CI failures are not caused by this PR. Both the CI (Java 8 compile) and Check License jobs fail on LanceTieringTest.java:384 in fluss-lake-lance, which is a pre-existing issue on the main branch

PR #2931, opened 2026-03-25
[flink] Remove redundant `bucket.num` and `bucket.key` from customer properties.

loserwang1024:
### Purpose

Remove redundant `bucket.num` and `bucket.key` from customer properties.


beryllw:
LGTM, with one minor typo.

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/utils/FlinkConversions.java
```diff
@@ -709,6 +709,10 @@ private static Map<String, String> extractCustomProperties(
             Configuration allProperties, Map<String, String> flussTableProperties) {
         Map<String, String> customProperties = new HashMap<>(allProperties.toMap());
         customProperties.keySet().removeAll(flussTableProperties.keySet());
+        // Remove bucket key and bucket num which is stored as table distribution rather than
+        // customer properties.
```
beryllw: It might be better to change "customer properties" to "custom properties".

loserwang1024: done

PR #2933, opened 2026-03-25
[lake/tiering] add total bytes read tracking and metrics for tiering service

### LINKED ISSUE #2932: [lake/tiering] add total bytes read tracking and metrics for tiering service
### Search before asking

### Description

Add total bytes read tracking and metrics for tiering jobs.

### Willingness to contribute

beryllw:
…jobs

### Purpose

Linked issue: close #2932

### DIFF
--- a/fluss-client/src/test/java/org/apache/fluss/client/table/scanner/log/LogFetchCollectorTest.java
+++ b/fluss-client/src/test/java/org/apache/fluss/client/table/scanner/log/LogFetchCollectorTest.java
@@ -22,7 +22,9 @@ import org.apache.fluss.client.metadata.TestingMetadataUpdater;
 import org.apache.fluss.client.table.scanner.ScanRecord;
 import org.apache.fluss.config.Configuration;
 import org.apache.fluss.metadata.TableBucket;
+import org.apache.fluss.record.LogRecordBatch;
 import org.apache.fluss.record.LogRecordReadContext;
+import org.apache.fluss.record.MemoryLogRecords;
 import org.apache.fluss.rpc.entity.FetchLogResultForBucket;
 
 import org.junit.jupiter.api.AfterEach;
@@ -183,13 +185,24 @@ public class LogFetchCollectorTest {
         assertThat(completedFetch1.isConsumed()).isTrue();
         assertThat(completedFetch2.isConsumed()).isTrue();
 
-        // totalBytesRead should be the sum of both completed fetches
+        // Compute the expected per-record size from the batch-level average
+        // (Arrow format records use batch.sizeInBytes() / recordCount as fallback)
+        MemoryLogRecords expectedData = genMemoryLogRecordsByObject(DATA1);
+        int expectedPerRecordSize = 0;
+        int expectedRecordCount = 0;
+        for (LogRecordBatch batch : expectedData.batches()) {
+            expectedPerRecordSize = batch.sizeInBytes() / batch.getRecordCount();
+            expectedRecordCount += batch.getRecordCount();
+        }
+        // Two fetches with the same data
+        long expectedTotal = (long) expectedPerRecordSize * expectedRecordCount * 2;
+
         long totalBytesRead = 0;
         for (ScanRecord record : scanRecords) {
-            assertThat(record.getSizeInBytes()).isGreaterThan(0);
+            assertThat(record.getSizeInBytes()).isEqualTo(expectedPerRecordSize);
             totalBytesRead += record.getSizeInBytes();
         }
-        assertThat(totalBytesRead).isGreaterThan(0);
+        assertThat(totalBytesRead).isEqualTo(expectedTotal);
     }
 
     private DefaultCompletedFetch makeCompletedFetch(
--- a/fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/source/reader/BoundedSplitReader.java
+++ b/fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/source/reader/BoundedSplitReader.java
@@ -21,8 +21,8 @@ import org.apache.fluss.client.table.scanner.ScanRecord;
 import org.apache.fluss.client.table.scanner.batch.BatchScanner;
 import org.apache.fluss.flink.lake.reader.IndexedLakeSplitRecordIterator;
 import org.apache.fluss.flink.source.split.SnapshotSplit;
-import org.apache.fluss.row.BinaryRow;
 import org.apache.fluss.row.InternalRow;
+import org.apache.fluss.row.MemoryAwareGetters;
 import org.apache.fluss.row.ProjectedRow;
 import org.apache.fluss.utils.CloseableIterator;
 
@@ -156,14 +156,13 @@ public class BoundedSplitReader implements AutoCloseable {
         @Override
         public ScanRecord next() {
             InternalRow row = rowIterator.next();
-            int sizeInBytes = -1;
-            if (row instanceof ProjectedRow) {
-                InternalRow underlyingRow = ((ProjectedRow) row).getUnderlyingRow();
-                if (underlyingRow instanceof BinaryRow) {
-                    sizeInBytes = ((BinaryRow) underlyingRow).getSizeInBytes();
-                }
-            } else if (row instanceof BinaryRow) {
-                sizeInBytes = ((BinaryRow) row).getSizeInBytes();
+            // Extract row size when the underlying row supports it (e.g. BinaryRow in
+            // KV-snapshot paths); falls back to -1 for other row types.
+            int sizeInBytes = ScanRecord.UNKNOWN_SIZE_IN_BYTES;
+            InternalRow underlyingRow =
+                    row instanceof ProjectedRow ? ((ProjectedRow) row).getUnderlyingRow() : row;
+            if (underlyingRow instanceof MemoryAwareGetters) {
+                sizeInBytes = ((MemoryAwareGetters) underlyingRow).getSizeInBytes();
             }
             return new ScanRecord(row, sizeInBytes);
         }
--- a/fluss-client/src/main/java/org/apache/fluss/client/table/scanner/ScanRecord.java
+++ b/fluss-client/src/main/java/org/apache/fluss/client/table/scanner/ScanRecord.java
@@ -30,6 +30,9 @@ import java.util.Objects;
 public class ScanRecord implements LogRecord {
     private static final long INVALID = -1L;
 
+    /** Indicates that the size in bytes is unknown for this record. */
+    public static final int UNKNOWN_SIZE_IN_BYTES = -1;
+
     private final long offset;
     private final long timestamp;
     private final ChangeType changeType;
@@ -45,7 +48,7 @@ public class ScanRecord implements LogRecord {
     }
 
     public ScanRecord(long offset, long timestamp, ChangeType changeType, InternalRow row) {
-        this(offset, timestamp, changeType, row, (int) INVALID);
+        this(offset, timestamp, changeType, row, UNKNOWN_SIZE_IN_BYTES);
     }
 
     public ScanRecord(
--- a/fluss-client/src/main/java/org/apache/fluss/client/table/scanner/log/CompletedFetch.java
+++ b/fluss-client/src/main/java/org/apache/fluss/client/table/scanner/log/CompletedFetch.java
@@ -131,7 +131,7 @@ abstract class CompletedFetch {
         if (currentBatch != null && currentBatch.getRecordCount() > 0) {
             return currentBatch.sizeInBytes() / currentBatch.getRecordCount();
         }
-        return -1;
+        return ScanRecord.UNKNOWN_SIZE_IN_BYTES;
     }
 
     boolean isConsumed() {
--- a/fluss-flink/fluss-flink-common/src/test/java/org/apache/fluss/flink/source/reader/BoundedSplitReaderTest.java
+++ b/fluss-flink/fluss-flink-common/src/test/java/org/apache/fluss/flink/source/reader/BoundedSplitReaderTest.java
@@ -19,6 +19,9 @@ package org.apache.fluss.flink.source.reader;
 
 import org.apache.fluss.client.table.scanner.batch.BatchScanner;
 import org.apache.fluss.row.InternalRow;
+import org.apache.fluss.row.ProjectedRow;
+import org.apache.fluss.row.indexed.IndexedRow;
+import org.apache.fluss.types.RowType;
 import org.apache.fluss.utils.CloseableIterator;
 
 import org.junit.jupiter.api.Test;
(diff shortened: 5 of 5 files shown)


fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/tiering/source/TieringSplitReader.java
```diff
@@ -436,6 +443,7 @@ private TableBucketWriteResult<WriteResult> completeLakeWriter(
         WriteResult writeResult = null;
         if (lakeWriter != null) {
             writeResult = lakeWriter.complete();
+            tieringMetrics.recordBytesWritten(lakeWriter.getBytesWritten());
```
leekeiabstraction: LanceLakeWriter currently does not implement getBytesWritten() and therefore default implementation will return -1. Will this corrupt the metric for Lance tiering?

fluss-lake/fluss-lake-paimon/src/main/java/org/apache/fluss/lake/paimon/tiering/PaimonLakeWriter.java
```diff
@@ -84,9 +87,21 @@ public PaimonWriteResult complete() throws IOException {
         } catch (Exception e) {
             throw new IOException("Failed to complete Paimon write.", e);
         }
+        if (commitMessage instanceof CommitMessageImpl) {
+            bytesWritten =
```
luoyuxia: IIUC, we want the bytes of fluss, not the bytes of paimon after written. Since the bytes of paimon will be much less than the bytes of fluss.

luoyuxia: cc @leonardBang

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/tiering/source/TieringSplitReader.java
beryllw: Thanks for the review. Fixed.

fluss-common/src/main/java/org/apache/fluss/lake/writer/LakeWriter.java
```diff
@@ -47,4 +47,13 @@ public interface LakeWriter<WriteResult> extends Closeable {
      * @throws IOException if an I/O error occurs
      */
     WriteResult complete() throws IOException;
+
+    /**
+     * Returns the total bytes written to the lake storage in this write session. Computed from
+     * actual physical file sizes after {@link #complete()} is called. Returns -1 by default for
+     * backward compatibility with third-party implementations.
+     */
+    default long getBytesWritten() {
```
luoyuxia: The method name suggests the raw data size written according to the Javadoc, but it actually returns the physical file size after compression and forma overhead. 
This confuse me.. Maybe we can rename this method for better reflect the actual semantics. For example, getPersistedBytes or something else

fluss-lake/fluss-lake-paimon/src/main/java/org/apache/fluss/lake/paimon/tiering/PaimonLakeWriter.java
leonardBang: +1 to use bytes of fluss record, the file size of Paimon on disk reflects the data after processing during the commit phase, and therefore no longer represents the actual data volume during transmission. The tieringService metrics should capture the total data volume across the entire transfer process.

fluss-common/src/main/java/org/apache/fluss/metrics/MetricNames.java
```diff
@@ -280,4 +280,12 @@ public class MetricNames {
     public static final String NETTY_NUM_ALLOCATIONS_PER_SECONDS = "numAllocationsPerSecond";
     public static final String NETTY_NUM_HUGE_ALLOCATIONS_PER_SECONDS =
             "numHugeAllocationsPerSecond";
+
+    // --------------------------------------------------------------------------------------------
+    // metrics for tiering service
+    // --------------------------------------------------------------------------------------------
+
+    // for lake tiering metrics - operator level
+    public static final String TIERING_SERVICE_WRITTEN_BYTES = "writtenBytes";
```
luoyuxia: This metric is for the lake bytes after written, I'm wondering  what metric name to use for the bytes before write?
Atually, this metrics name make me feel it's for the bytes before write.

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/tiering/source/metrics/TieringMetrics.java
```diff
@@ -0,0 +1,68 @@
...
+                        .addGroup(FLUSS_METRIC_GROUP)
+                        .addGroup(TIERING_SERVICE_GROUP);
+
+        this.writtenBytesCounter =
+                tieringServiceGroup.counter(MetricNames.TIERING_SERVICE_WRITTEN_BYTES);
+        tieringServiceGroup.meter(
+                MetricNames.TIERING_SERVICE_WRITTEN_BYTES_RATE, new MeterView(writtenBytesCounter));
+    }
+
+    /** Records bytes written to the lake. Called once per bucket completion. */
```
leonardBang: Called once per bucket completion.  -> we can clarify more e.g. the calculation time is triggered when commit to lake.

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/tiering/source/metrics/TieringMetrics.java
```diff
@@ -0,0 +1,68 @@
...
+@Internal
+public class TieringMetrics {
+
+    // Metric group names
+    public static final String FLUSS_METRIC_GROUP = "fluss";
+    public static final String TIERING_SERVICE_GROUP = "tieringService";
+
+    private final Counter writtenBytesCounter;
+
+    public TieringMetrics(SourceReaderMetricGroup sourceReaderMetricGroup) {
```
luoyuxia, line 52: so, this metrics is reported to flink metric, right?  Is it same to flink cdc?

website/docs/maintenance/observability/monitor-metrics.md
```diff
@@ -1125,6 +1125,36 @@ How to Use Flink Metrics, you can see [Flink Metrics](https://nightlies.apache.o
...
+      <th class="text-left" style={{width: '225pt'}}>Metrics Name</th>
+      <th class="text-left" style={{width: '165pt'}}>Level</th>
+      <th class="text-left" style={{width: '300pt'}}>Description</th>
+      <th class="text-left" style={{width: '70pt'}}>Type</th>
+    </tr>
+    </thead>
+    <tbody>
+        <tr>
+            <td>writtenBytes</td>
+            <td>Flink Source Operator</td>
```
leonardBang, line 1149: Here look like strange when ehe word `written` appears with `Source Operator`, our Source Operator is a special Operator which is responsible for both read and write, could you add some notes to the  description?

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/tiering/source/metrics/TieringMetrics.java
beryllw: Not quite the same. The tiering job keeps read and write in one operator (no shuffle), while Flink CDC introduces shuffle between the source and sink.

fluss-client/src/main/java/org/apache/fluss/client/table/scanner/log/ScanRecords.java
```diff
@@ -37,12 +37,27 @@
...
         this.records = records;
+        this.totalBytesRead = totalBytesRead;
+    }
+
+    /**
+     * Get the total bytes read from the Fluss log in this batch.
+     *
+     * @return the total bytes read
+     */
+    public long getTotalBytesRead() {
```
luoyuxia:
```suggestion
    public long getTotalBytes() {
```
nit: I think we can remove `read` suffix since `ScanRecords` already means read/scan

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/tiering/source/TieringSplitReader.java
```diff
@@ -502,10 +514,15 @@ private TableBucketWriteResultWithSplitIds forSnapshotSplitRecords(
         LakeWriter<WriteResult> lakeWriter =
                 getOrCreateLakeWriter(
                         bucket, checkNotNull(currentSnapshotSplit).getPartitionName());
+        long bytesRead = 0;
         while (recordIterator.hasNext()) {
             ScanRecord scanRecord = recordIterator.next().record();
             lakeWriter.write(scanRecord);
+            InternalRow row = scanRecord.getRow();
+            // Snapshot path always produces BinaryRow (CompactedRow/IndexedRow).
```
luoyuxia: TieringSplitReader’s snapshot-path byte accounting will throw ClassCastException as soon as snapshot rows are projected or schema-remapped. 

SnapshotFilesReader.next() can legitimately return a ProjectedRow, but TieringSplitReader.forSnapshotSplitRecords() unconditionally casts every InternalRow to BinaryRow. That makes the new metric path unsafe for evolved-schema snapshots or any snapshot scan with projectedFields.

```
SnapshotFilesReader
BinaryValue originValue = valueDecoder.decodeValue(value);
InternalRow originRow = originValue.row;
if (targetSchemaId != originValue.schemaId) {
    int[] indexMapping =
            schemaProjectionCache.computeIfAbsent(
                    originValue.schemaId,
                    sourceSchemaId ->
                            SchemaUtil.getIndexMapping(
                                    schemaGetter.getSchema(sourceSchemaId), targetSchema));
    originRow = ProjectedRow.from(indexMapping).replaceRow(originRow);
}
```

fluss-client/src/main/java/org/apache/fluss/client/table/scanner/log/LogFetchCollector.java
```diff
@@ -135,6 +136,11 @@ public Map<TableBucket, List<ScanRecord>> collectFetch(final LogFetchBuffer logF
 
                         recordsRemaining -= records.size();
                     }
+
+                    // Only count bytes when the fetch is fully consumed
+                    if (nextInLineFetch.isConsumed()) {
```
luoyuxia: IIUC, totalBytesRead here is not an accurate per-ScanRecords measurement. Some ScanRecords may return records with totalBytesRead == 0, while others may accumulate the full size of a completed fetch and look over-counted relative to the records returned in that poll.

fluss-common/src/main/java/org/apache/fluss/row/ProjectedRow.java
```diff
@@ -193,4 +193,11 @@ public static ProjectedRow from(Schema originSchema, Schema expectedSchema) {
         int[] indexMapping = getIndexMapping(originSchema, expectedSchema);
         return new ProjectedRow(indexMapping);
     }
+
+    public int getSizeInBytes() {
```
luoyuxia: It may introduce confluse. it's a projection row, users may expected it represent the bytes after projection. But now, it represents the undering raw row bytes.

fluss-common/src/main/java/org/apache/fluss/record/LogRecord.java
```diff
@@ -62,6 +62,8 @@ public interface LogRecord {
      */
     InternalRow getRow();
 
+    int getSizeInBytes();
```
luoyuxia: is it possible not to expose this method. It's public interface, i'm cautious to expose it because the value is not very precise

fluss-common/src/main/java/org/apache/fluss/row/ProjectedRow.java
```diff
@@ -194,7 +194,14 @@ public static ProjectedRow from(Schema originSchema, Schema expectedSchema) {
...
 
-    public int getSizeInBytes() {
+    /**
+     * Returns the on-wire size in bytes of the underlying raw row (before projection), not the
+     * projected subset. This is used for byte-level metrics tracking where the original record size
+     * is needed.
+     *
+     * @return the size in bytes of the underlying row, or -1 if unknown.
+     */
+    public int getUnderlyingRowSizeInBytes() {
```
luoyuxia: I'd like suggest to rename to getUnderlyingRow() and return row,  
`getUnderlyingRowSizeInBytes` is to specific to me.

fluss-client/src/test/java/org/apache/fluss/client/table/scanner/log/LogFetchCollectorTest.java
```diff
@@ -147,14 +145,51 @@ void testCollectAfterUnassign() throws Exception {
...
+        assertThat(completedFetch1.isConsumed()).isTrue();
+        assertThat(completedFetch2.isConsumed()).isTrue();
+
+        // totalBytesRead should be the sum of both completed fetches
+        long totalBytesRead = 0;
+        for (ScanRecord record : scanRecords) {
+            assertThat(record.getSizeInBytes()).isGreaterThan(0);
+            totalBytesRead += record.getSizeInBytes();
+        }
+        assertThat(totalBytesRead).isGreaterThan(0);
```
leekeiabstraction: We should be able to use a more precise assertion here?

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/source/reader/BoundedSplitReader.java
```diff
@@ -153,7 +155,17 @@ public boolean hasNext() {
...
+            InternalRow row = rowIterator.next();
+            int sizeInBytes = -1;
+            if (row instanceof ProjectedRow) {
+                InternalRow underlyingRow = ((ProjectedRow) row).getUnderlyingRow();
+                if (underlyingRow instanceof BinaryRow) {
+                    sizeInBytes = ((BinaryRow) underlyingRow).getSizeInBytes();
+                }
+            } else if (row instanceof BinaryRow) {
+                sizeInBytes = ((BinaryRow) row).getSizeInBytes();
+            }
```
leekeiabstraction: Do we expect other row types? If not, it might be better to throw exception or log (at debug considering that this is hot path)?

beryllw: Thanks for the comment.
Other row types are possible here since BoundedSplitReader is a shared path — Lake snapshots produce GenericRow and Arrow logs produce ColumnarRow. For Tiering (the consumer of this size info), only KV-snapshot paths are involved, which always yield BinaryRow. The fallback to UNKNOWN_SIZE_IN_BYTES is a safe guard at the shared layer, and the metrics side simply skips it with a > 0 check.

PR #2940, opened 2026-03-26
[fluss-common] Validate partition time when creating partition on auto partition table.

### LINKED ISSUE #2939: Need to validate auto-partition time when creation and write new partition
### Search before asking

### Fluss version

0.9.0 (latest release)

### Please describe the bug 🐞

Currently, when a user manually creates a partition via the Admin API or when the client dynamically creates a partition during a write, there is no validation that the partition time value is:
1. Correctly formatted — matching the pattern expected by the configured TABLE_AUTO_PARTITION_TIME_UNIT (e.g., yyyyMMdd for DAY, yyyy for YEAR).
2. Not out-of-date — i.e., not older than the earliest partition retained by the auto-partition retention policy (TABLE_AUTO_PARTITION_NUM_RETENTION).

This means a user can currently:
* Create a partition like 2024-03-25 on a table with YEAR unit (wrong format), which will never be managed by auto-partition.
* Create a partition that is already beyond the retention window, which will be immediately eligible for deletion.

### Solution

Add a shared utility method PartitionUtils.validateAutoPartitionTime in fluss-common that:
1. Returns early if auto-partition is not enabled for the table.
2. Validates the partition value matches the format string for the configured time unit (e.g., yyyyMMdd for DAY).
3. Validates the partition value is not older than currentTime - numToRetain units, throwing InvalidPartitionException with a descriptive message in both failure cases.

loserwang1024:
### Purpose

Linked issue: close #2939

### DIFF
--- a/fluss-client/src/main/java/org/apache/fluss/client/write/DynamicPartitionCreator.java
+++ b/fluss-client/src/main/java/org/apache/fluss/client/write/DynamicPartitionCreator.java
@@ -21,7 +21,6 @@ import org.apache.fluss.client.admin.Admin;
 import org.apache.fluss.client.metadata.MetadataUpdater;
 import org.apache.fluss.exception.FlussRuntimeException;
 import org.apache.fluss.exception.PartitionNotExistException;
-import org.apache.fluss.metadata.PartitionSpec;
 import org.apache.fluss.metadata.PhysicalTablePath;
 import org.apache.fluss.metadata.ResolvedPartitionSpec;
 import org.apache.fluss.metadata.TablePath;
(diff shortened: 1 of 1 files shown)


loserwang1024:
@platinumhamburg @luoyuxia , CC

fluss-client/src/main/java/org/apache/fluss/client/write/DynamicPartitionCreator.java
```diff
@@ -121,14 +125,19 @@ private boolean forceCheckPartitionExist(PhysicalTablePath physicalTablePath) {
...
+            PhysicalTablePath physicalTablePath,
+            List<String> partitionKeys,
+            AutoPartitionStrategy autoPartitionStrategy) {
         String partitionName = physicalTablePath.getPartitionName();
         TablePath tablePath = physicalTablePath.getTablePath();
         checkArgument(partitionName != null, "Partition name shouldn't be null.");
         ResolvedPartitionSpec resolvedPartitionSpec =
                 ResolvedPartitionSpec.fromPartitionName(partitionKeys, partitionName);
+        PartitionSpec partitionSpec = resolvedPartitionSpec.toPartitionSpec();
+        validateAutoPartitionTime(partitionSpec, partitionKeys, autoPartitionStrategy);
```
platinumhamburg: I think there's a state leak issue in `DynamicPartitionCreator` that could cause confusing behavior.

In `createPartition()`, `validateAutoPartitionTime` runs **after** the path has already been added to `inflightPartitionsToCreate` (line 93). If validation throws `InvalidPartitionException`, the path stays in the set permanently — `admin.createPartition().whenComplete(...)` never fires, so `onPartitionCreationFailed` (which does the `remove()`) never runs either.

What happens next: any subsequent write with the same invalid partition value hits `inflightPartitionsToCreate.add()` → returns `false` → skips creation entirely → tries to send data to a partition that doesn't exist → blows up with a different error. Pretty hard to debug from the user's perspective.

I think the simplest fix is to move the validation before we touch `inflightPartitionsToCreate` at all — e.g. at the top of `checkAndCreatePartitionAsync`, right after the null check:

```java
public void checkAndCreatePartitionAsync(
        PhysicalTablePath physicalTablePath,
        List<String> partitionKeys,
        AutoPartitionStrategy autoPartitionStrategy) {
    String partitionName = physicalTablePath.getPartitionName();
    if (partitionName == null) {
        return;
    }

    // Validate early, before touching any state.
    ResolvedPartitionSpec resolvedPartitionSpec =
            ResolvedPartitionSpec.fromPartitionName(partitionKeys, partitionName);
    validateAutoPartitionTime(
            resolvedPartitionSpec.toPartitionSpec(), partitionKeys, autoPartitionStrategy);

    // ... rest of the existing logic
}
```

This way invalid partitions get rejected immediately without polluting the inflight set, and as a bonus it removes the redundant `ResolvedPartitionSpec` construction in `createPartition`.

loserwang1024: It looks better. Just modified it.

PR #2942, opened 2026-03-26
[server] Fix TabletServer fails to restart after unclean shutdown due to EOFException in log recovery

### LINKED ISSUE #2941: [server] TabletServer fails to restart after unclean shutdown due to EOFException in log recovery
### Search before asking

### Fluss version

main (development)

### Please describe the bug 🐞

After an unclean shutdown (e.g., `SIGKILL`, OOM, power failure), `TabletServer` fails to restart due to a `FlussRuntimeException` wrapping an `EOFException` during log recovery. The server becomes completely unrecoverable without manual intervention.

On unclean shutdown, `recoverSegment()` is only called when `sanityCheck()` throws (i.e., index file is corrupt). When the index file is intact but the `.log` file has a truncated tail — the most common unclean shutdown scenario — `sanityCheck()` passes, `recoverSegment()` is skipped, and a subsequent call to `readNextOffset()` → `maxTimestampSoFar()` → `readLargestTimestamp()` hits the partial batch and crashes.

**Steps to Reproduce:**

1. Start a `TabletServer` and produce data to a table (e.g., `t_geely_2x_freeze_frame_info`, bucket 33).
2. Force-kill the `TabletServer` process (`kill -9`) while writes are in progress, ensuring the active log segment has unflushed data.
3. Restart the `TabletServer`.
4. Observe that the server fails to start with the stack trace below.

**Stack Trace:**

```
org.apache.fluss.exception.FlussRuntimeException: Failed to recovery log
    at org.apache.fluss.server.log.LogManager.loadLogs(LogManager.java:207)
    at org.apache.fluss.server.log.LogManager.startup(LogManager.java:139)
    at org.apache.fluss.server.tablet.TabletServer.startServices(TabletServer.java:228)
    ...
Caused by: org.apache.fluss.exception.FlussRuntimeException: Failed to load record batch at position 495460312
    from FileRecords(size=495460352, file=.../00000000000000000000.log, start=0, end=2147483647)
    at org.apache.fluss.record.FileLogInputStream$FileChannelLogRecordBatch.loadByteBufferWithSize(FileLogInputStream.java:222)
    at org.apache.fluss.record.FileLogInputStream$FileChannelLogRecordBatch.loadBatchHeader(FileLogInputStream.java:211)
    at org.apache.fluss.record.FileLogInputStream$FileChannelLogRecordBatch.commitTimestamp(FileLogInputStream.java:134)
    at org.apache.fluss.record.FileLogRecords.largestTimestampAfter(FileLogRecords.java:386)
    at org.apache.fluss.server.log.LogSegment.readLargestTimestamp(LogSegment.java:644)
    at org.apache.fluss.server.log.LogSegment.readMaxTimestampAndStartOffsetSoFar(LogSegment.java:214)
    at org.apache.fluss.server.log.LogSegment.maxTimestampSoFar(LogSegment.java:200)
    at org.apache.fluss.server.log.LogSegment.recover(LogSegment.java:319)
    at org.apache.fluss.server.log.LogLoader.recoverSegment(LogLoader.java:269)
    at org.apache.fluss.server.log.LogLoader.recoverLog(LogLoader.java:168)
    ...
Caused by: java.io.EOFException: Failed to read `record batch header` from file channel.
    Expected to read 48 bytes, but reached end of file after reading 40 bytes.
    Started read from position 495460312.
    at org.apache.fluss.utils.FileUtils.readFullyOrFail(FileUtils.java:110)
    at org.apache.fluss.utils.FileUtils.loadByteBufferFromFile(FileUtils.java:138)
    ...
```

### Solution

_No response_

LiebingYu:
### Purpose

Linked issue: close #2941

### DIFF
--- a/fluss-server/src/main/java/org/apache/fluss/server/log/LogLoader.java
+++ b/fluss-server/src/main/java/org/apache/fluss/server/log/LogLoader.java
@@ -23,7 +23,6 @@ import org.apache.fluss.exception.InvalidOffsetException;
 import org.apache.fluss.exception.LogSegmentOffsetOverflowException;
 import org.apache.fluss.exception.LogStorageException;
 import org.apache.fluss.metadata.LogFormat;
-import org.apache.fluss.server.exception.CorruptIndexException;
 import org.apache.fluss.utils.FlussPaths;
 import org.apache.fluss.utils.types.Tuple2;
 
@@ -137,6 +136,7 @@ final class LogLoader {
      */
     private Tuple2<Long, Long> recoverLog() throws IOException {
         if (!isCleanShutdown) {
+            long recoverLogStart = System.currentTimeMillis();
             List<LogSegment> unflushed =
                     logSegments.values(recoveryPointCheckpoint, Long.MAX_VALUE);
             int numUnflushed = unflushed.size();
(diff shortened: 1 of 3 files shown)


fluss-server/src/main/java/org/apache/fluss/server/log/LogLoader.java
```diff
@@ -294,6 +289,20 @@ private void loadSegmentFiles() throws IOException {
...
+                        try {
+                            segment.sanityCheck();
+                        } catch (NoSuchFileException e) {
+                            if (isCleanShutdown
+                                    || segment.getBaseOffset() < recoveryPointCheckpoint) {
+                                LOG.error(
+                                        "Could not find offset index file corresponding to log file {} "
+                                                + "for bucket {}, recovering segment and rebuilding index files...",
+                                        logSegments.getTableBucket(),
+                                        segment.getFileLogRecords().file().getAbsoluteFile());
```
leekeiabstraction: Bucket and file path arg seems to be swapped.

fluss-server/src/main/java/org/apache/fluss/server/log/LogLoader.java
```diff
@@ -294,6 +289,20 @@ private void loadSegmentFiles() throws IOException {
...
+
+                        try {
+                            segment.sanityCheck();
+                        } catch (NoSuchFileException e) {
+                            if (isCleanShutdown
+                                    || segment.getBaseOffset() < recoveryPointCheckpoint) {
+                                LOG.error(
+                                        "Could not find offset index file corresponding to log file {} "
+                                                + "for bucket {}, recovering segment and rebuilding index files...",
+                                        logSegments.getTableBucket(),
```
fresh-borzoni: order is reversed

fluss-server/src/main/java/org/apache/fluss/server/log/LogLoader.java
fresh-borzoni: If recovery hits the InvalidOffsetException -> truncateTo(baseOffset) path, logEndOffset can land below recoveryPointCheckpoint, kafka has smth like `Math.min(recoveryPointCheckpoint, logEndOffset)`

fresh-borzoni (commented):
@LiebingYu Thanks for the PR, we bumped into this issue last week, so I studied this code

The unconditional-recoverSegment change is right (matches Kafka), but the fix doesn't actually close this issue in my opinion.                                                                                                  
                                                            
I was able to reproduced the crash by appending 40 zero bytes to the active .log (typical shape after an interrupted write). The crash is inside recover() itself: it asks for max-timestamp before doing anything, the getter scans the whole file to compute it, and the scan walks into the corrupt tail.
                                                                                                                                                                                                                                                     
The possible fix might be small: track max-timestamp in a local variable inside the loop instead of going through the getter. Recovery is iterating every batch anyway.
                                                                                                                                                                                                                                                     
The new test passes against the buggy code: its garbage bytes happen to decode as a huge length, so the iterator rejects the would-be batch before reaching the bug. If we switch the bytes to zeros, then it fails on this branch.

Three smaller things I noticed in the same area, all pre-existing: 
* recover()'s catch is for CorruptRecordException, but it can also fail with CRC check with CorruptMessageException, for example
*  an unrecognised magic byte throws an uncaught IllegalArgumentException from the parser, and we had it during our unclean shutdown when garbage was interpreted this way

wuchong:
Thanks @fresh-borzoni for the review comments. Your finding sounds make sense to me. I think we should address them as well. What do you think @LiebingYu @swuferhong ?

LiebingYu:
@fresh-borzoni Thank you for your insightful discovery.

The new test does exactly bypass the batch parsing, and has been modified to use 40 zero bytes.

On this issue, the core differences between Fluss and Kafka are:
1. When calling `maxTimestampSoFar()`, Fluss also reads the `.log` file to find a more precise `maxTimestampSoFar`, whereas Kafka does not. This causes the exception to occur prematurely. Regarding the fix here, I agree with your approach — since we are already iterating through all batches, there is no need to read the precise `maxTimestampSoFar` upfront before starting. Instead, we can simply update it after reading each batch.

2. Compared to Kafka, Fluss may encounter more types of exceptions during batch validation, including:
+ CorruptRecordException
+ InvalidRecordException
+ CorruptMessageException
+ IllegalArgumentException
+ IndexOutOfBoundsException

All of these have been handled. Could you please help review it again? Feel free to let me know if you have any concerns.

CC @wuchong @swuferhong

wuchong (approved):
LGTM. 

@fresh-borzoni @swuferhong could you take another look?

fresh-borzoni (approved):
Looks good, thank you @LiebingYu

Issue #2946, opened 2026-03-27
Azure FS token delegation does not take effect


affo:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/fluss/issues) and found nothing similar.

### Fluss version

0.9.0 (latest release)

### Please describe the bug 🐞

When performing a KV snapshot read from a client without any Azure-related setting, I get the following trace:

```
org.apache.fluss.exception.FlussRuntimeException: Failed to initialize snapshot files reader.
        at org.apache.fluss.client.table.scanner.batch.KvSnapshotBatchScanner.ensureNoException(KvSnapshotBatchScanner.java:169)
        at org.apache.fluss.client.table.scanner.batch.KvSnapshotBatchScanner.pollBatch(KvSnapshotBatchScanner.java:140)
        at com.ververica.fluss.ReadCmd.readBatchScanner(ReadCmd.java:255)
        at com.ververica.fluss.ReadCmd.readPkFull(ReadCmd.java:173)
        at com.ververica.fluss.ReadCmd.call(ReadCmd.java:63)
        at com.ververica.fluss.ReadCmd.call(ReadCmd.java:29)
        at picocli.CommandLine.executeUserObject(CommandLine.java:2041)
        at picocli.CommandLine.access$1500(CommandLine.java:148)
        at picocli.CommandLine$RunLast.executeUserObjectOfLastSubcommandWithSameParent(CommandLine.java:2461)
        at picocli.CommandLine$RunLast.handle(CommandLine.java:2453)
        at picocli.CommandLine$RunLast.handle(CommandLine.java:2415)
        at picocli.CommandLine$AbstractParseResultHandler.execute(CommandLine.java:2273)
        at picocli.CommandLine$RunLast.execute(CommandLine.java:2417)
        at picocli.CommandLine.execute(CommandLine.java:2170)
        at com.ververica.fluss.TestPkCmd.call(TestPkCmd.java:60)
        at com.ververica.fluss.TestPkCmd.call(TestPkCmd.java:7)
        at picocli.CommandLine.executeUserObject(CommandLine.java:2041)
        at picocli.CommandLine.access$1500(CommandLine.java:148)
        at picocli.CommandLine$RunLast.executeUserObjectOfLastSubcommandWithSameParent(CommandLine.java:2461)
        at picocli.CommandLine$RunLast.handle(CommandLine.java:2453)
        at picocli.CommandLine$RunLast.handle(CommandLine.java:2415)
[...]

PR #2947, opened 2026-03-27
[fs/azure] Set auth.type=Custom for token delegation

affo:
### Purpose

Linked issue: close https://github.com/apache/fluss/issues/2946

Token delegation does not seem to work when the client does not configure any entry of the Azure FS configuration:

1. FlussConnection filters client config:
     FileSystem.initialize(
         Configuration.fromMap(extractPrefix(conf.toMap(), "client.fs.")), null);
     1. extractPrefix filters to client.fs.* keys but does not strip the prefix — so keys remain as client.fs.* in the resulting config.
     2. AzureFileSystemPlugin.getHadoopConfiguration looks for keys with prefix "fs.azure." — but the client config has no such keys (only client.fs.*). The Hadoop config receives no Azure credentials.
     3. setCredentialProvider (AzureFileSystemPlugin.java:64-84): since fs.azure.account.key is absent from the Hadoop config, it falls back to delegation tokens and calls AzureDelegationTokenReceiver.updateHadoopConfig(hadoopConfig).
     4. updateHadoopConfig (AzureDelegationTokenReceiver.java:41-71):
       - Sets fs.azure.account.oauth.provider.type = DynamicTemporaryAzureCredentialsProvider
       - Sets additionInfos entries (just fs.azure.account.oauth2.client.endpoint)
       - DOES NOT set fs.azure.account.auth.type
     5. ABFS driver (AzureBlobFileSystem.initialize → AzureBlobFileSystemStore.<init> → initializeClient): since fs.azure.account.auth.type is not set, defaults to SharedKey auth → calls SimpleKeyProvider.getStorageAccountKey → key not found → "Failure to
     initialize configuration".

### Brief change log

Simply `hadoopConfig.set("fs.azure.account.auth.type", "Custom");` in case no account key is set (presumably, client side).

### Tests

 - Fix existing UT test that was skipping an exception thrown on FS init to actually check the exception is not due to misconfig
 - add check for `"fs.azure.account.auth.type", "Custom"` to existing UTs
 - add IT test to simulate the token delegation path (the previous IT test only checked the "server path" by setting the complete config)

### API and Format

NA

### Documentation

NA

### DIFF
--- a/fluss-filesystems/fluss-fs-azure/src/test/java/org/apache/fluss/fs/azure/AbfsFileSystemDelegationTokenITCase.java
+++ b/fluss-filesystems/fluss-fs-azure/src/test/java/org/apache/fluss/fs/azure/AbfsFileSystemDelegationTokenITCase.java
@@ -44,7 +44,7 @@ import java.util.Map;
  */
 class AbfsFileSystemDelegationTokenITCase extends FileSystemBehaviorTestSuite {
 
-    private static final String ABFS_FS_PATH = "abfs://flus@test.dfs.core.windows.net/test";
+    private static final String ABFS_FS_PATH = "abfs://fluss@test.dfs.core.windows.net/test";
 
     @BeforeAll
     static void setup() throws Exception {
--- a/fluss-filesystems/fluss-fs-azure/src/test/java/org/apache/fluss/fs/azure/AbfsFileSystemBehaviorITCase.java
+++ b/fluss-filesystems/fluss-fs-azure/src/test/java/org/apache/fluss/fs/azure/AbfsFileSystemBehaviorITCase.java
@@ -39,7 +39,7 @@ class AbfsFileSystemBehaviorITCase extends FileSystemBehaviorTestSuite {
 
     private static final String AZURE_ACCOUNT_KEY = "ZmFrZS1rZXkK";
     private static final String ENDPOINT_PREFIX = "http://localhost:";
-    public static final String ABFS_FS_PATH = "abfs://flus@test.dfs.core.windows.net/test";
+    public static final String ABFS_FS_PATH = "abfs://fluss@test.dfs.core.windows.net/test";
 
     private static MockAuthServer mockAuthServer;
 
--- a/fluss-filesystems/fluss-fs-azure/src/test/java/org/apache/fluss/fs/azure/AzureFileSystemPluginTest.java
+++ b/fluss-filesystems/fluss-fs-azure/src/test/java/org/apache/fluss/fs/azure/AzureFileSystemPluginTest.java
@@ -25,6 +25,8 @@ import org.apache.fluss.fs.token.ObtainedSecurityToken;
 
 import org.junit.jupiter.api.Test;
 
+import java.io.PrintWriter;
+import java.io.StringWriter;
 import java.net.URI;
 import java.util.HashMap;
 import java.util.Map;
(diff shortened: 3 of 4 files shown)


fluss-filesystems/fluss-fs-azure/src/test/java/org/apache/fluss/fs/azure/AbfsFileSystemDelegationTokenITCase.java
```diff
@@ -0,0 +1,99 @@
...
+
+/**
+ * Tests that validate the Azure File System Plugin initializes correctly via the delegation token
+ * path — i.e., when no {@code fs.azure.account.key} is configured and the client receives an OAuth
+ * token from the server instead. This is the code path exercised by Fluss clients reading KV
+ * snapshots from remote Azure storage.
+ */
+class AbfsFileSystemDelegationTokenITCase extends FileSystemBehaviorTestSuite {
+
+    private static final String ABFS_FS_PATH = "abfs://flus@test.dfs.core.windows.net/test";
```
polyzos: i assume here you meant to have fluss://?

fluss-filesystems/fluss-fs-azure/src/main/java/org/apache/fluss/fs/azure/token/AzureDelegationTokenReceiver.java
```diff
@@ -56,6 +56,11 @@ public static void updateHadoopConfig(org.apache.hadoop.conf.Configuration hadoo
             LOG.debug("Provider already exists");
         }
 
+        // Tell the ABFS driver to use the custom token provider instead of defaulting to SharedKey.
+        // DynamicTemporaryAzureCredentialsProvider implements CustomTokenProviderAdaptee, which
+        // requires auth.type=Custom to be activated.
+        hadoopConfig.set("fs.azure.account.auth.type", "Custom");
```
polyzos, line 62: `auth.type` is set before the additionInfos null guard

fluss-filesystems/fluss-fs-azure/src/test/java/org/apache/fluss/fs/azure/AbfsFileSystemDelegationTokenITCase.java
affo: it does not matter, but let me fix the typo 🤝 
I am also going to fix the docs, as I copy/pasted the URI from there 🚀

fluss-filesystems/fluss-fs-azure/src/main/java/org/apache/fluss/fs/azure/token/AzureDelegationTokenReceiver.java
affo: do you think this may be a problem?

AFAIK, it should not be, you can see `auth.type` as a setting that goes in pair with `provider.type` above.

If the guard fails everything will fail -> which means we did not yet get any token from the server side -> this problem is definitely out of scope of this PR and potentially impacts any FS implementation in Fluss.

What we are doing here is adding that property to make FS init take into consideration the actual custom Credentials provider that would never be used otherwise.

fluss-filesystems/fluss-fs-azure/src/test/java/org/apache/fluss/fs/azure/AbfsFileSystemDelegationTokenITCase.java
affo: also in another test 😄

affo:
@polyzos I addressed the typo and also made the comments a bit more explicative, fixed also the complex `while` loop to check the stacktrace in the IT test 🤝 

For the `additionInfos`, please check my comment above.

Comment can be squashed without preserving the commit body

PR #2948, opened 2026-03-28
[rpc] Client zero-copy lazy parse ByteBuf to avoid deep memory copy

### LINKED ISSUE #1184: Client lazy parse ByteBuf to avoid deeply memory copy
### Search before asking

### Motivation

Currently, we will do one deeply memory copy when receive `Response` from server, see `NettyClientHandler#channelRead()` for more details:

Image

This behavior would unnecessarily increase heap memory usage and negatively impact read performance. Therefore, if precise manual buffer release can be achieved, we can eliminate memory copying here.

### Solution

_No response_

### Anything else?

_No response_

### Willingness to contribute

fresh-borzoni:
closes #1184  

Fix: each `DefaultCompletedFetc`h retains the parsed `ByteBuf` and releases in `drain()`. The response handler releases the base reference in finally. Netty refcounting keeps the buffer alive exactly until the last bucket's records are consumed.    

Effectively this is similar how it works in Rust SDK with Arc refcounting and Drop automatic release.

Both client paths now use zero-copy. isInnerClient flag removed as it was used only for this.

### DIFF
--- a/fluss-server/src/test/java/org/apache/fluss/server/coordinator/CoordinatorHighAvailabilityITCase.java
+++ b/fluss-server/src/test/java/org/apache/fluss/server/coordinator/CoordinatorHighAvailabilityITCase.java
@@ -91,7 +91,7 @@ class CoordinatorHighAvailabilityITCase {
     @BeforeEach
     void setUp() throws Exception {
         Configuration clientConf = new Configuration();
-        rpcClient = RpcClient.create(clientConf, TestingClientMetricGroup.newInstance(), false);
+        rpcClient = RpcClient.create(clientConf, TestingClientMetricGroup.newInstance());
     }
 
     @AfterEach


fresh-borzoni:
@swuferhong @luoyuxia PTAL 🙏

fluss-client/src/main/java/org/apache/fluss/client/table/scanner/log/LogFetcher.java
```diff
@@ -387,6 +393,11 @@ private synchronized void handleFetchLogResponse(
                             LogRecords logRecords = fetchResultForBucket.recordsOrEmpty();
                             if (!MemoryLogRecords.EMPTY.equals(logRecords)
                                     || fetchResultForBucket.getErrorCode() != Errors.NONE.code()) {
+                                // Retain the parsed buffer so it stays alive while
+                                // this CompletedFetch's records are being consumed.
+                                if (parsedByteBuf != null) {
+                                    parsedByteBuf.retain();
```
polyzos, line 398: retain() is called even for error-only buckets where logRecords is EMPTY... the
buffer is not needed in that case. Only retain when hasRecords is true.

polyzos:
@fresh-borzoni Great contribution to improving the hot read path.
I've added one comment and pushed two small notes for future reference.

fresh-borzoni:
@polyzos Thank you for the review
Addressed comment, PTAL 🙏

polyzos:
@fresh-borzoni can you rebase? and then we can merge

polyzos:
LGTM.. Thank you for your contribution

wuchong:
Thanks @fresh-borzoni @polyzos , I think this is a good improvement for the client. However, I found a potential memory leak in LimitScan and have some concerns on the OOM risk. I created #2998 for further discussion.

PR #2951, opened 2026-03-28
[server] Add server-side filter execution and client-side integration

### LINKED ISSUE #2950: Add server-side record batch filter execution
### Search before asking

### Description

Introduce server-side record batch filtering using batch-level statistics (min/max values, null counts) that are already available in the V1 log batch format. When a client sends a fetch request with a filter predicate, the server evaluates the predicate against each batch's statistics and skips batches that cannot contain matching records.

Key points:

- **Batch-level filtering, not row-level**: the server uses batch statistics to skip entire batches. The client still performs row-level filtering on the returned batches.
- **ARROW format only**: only ARROW log format includes batch-level statistics (V1+ magic). COMPACTED/INDEXED formats fall back to unfiltered reads.
- **Schema evolution safe**: a `PredicateSchemaResolver` adapts the predicate when the batch schema differs from the predicate schema, with safe fallback (include the batch) on any failure.
- **Offset advancement**: when all batches in a fetch are filtered out, the server returns a `filteredEndOffset` so the client can advance past the filtered range without re-fetching.

### Willingness to contribute

platinumhamburg:
### Purpose

Linked issue: close #2950 2950

### DIFF
--- a/fluss-common/src/main/java/org/apache/fluss/record/FileLogProjection.java
+++ b/fluss-common/src/main/java/org/apache/fluss/record/FileLogProjection.java
@@ -158,85 +158,7 @@ public class FileLogProjection {
             return builder.build();
         }
 
-        boolean isAppendOnly =
-                (logHeaderBuffer.get(attributeOffset(magic)) & APPEND_ONLY_FLAG_MASK) > 0;
-
-        // For V1+, skip statistics data between header and records
-        int statisticsLength = 0;
-        if (magic >= LOG_MAGIC_VALUE_V1) {
-            statisticsLength = logHeaderBuffer.getInt(statisticsLengthOffset(magic));
-        }
-        int recordsStartOffset = recordBatchHeaderSize + statisticsLength;
-
-        final int changeTypeBytes;
-        final long arrowHeaderOffset;
-        if (isAppendOnly) {
-            changeTypeBytes = 0;
-            arrowHeaderOffset = position + recordsStartOffset;
-        } else {
-            changeTypeBytes = logHeaderBuffer.getInt(recordsCountOffset(magic));
-            arrowHeaderOffset = position + recordsStartOffset + changeTypeBytes;
-        }
-
-        // read arrow header
-        arrowHeaderBuffer.rewind();
-        readFullyOrFail(channel, arrowHeaderBuffer, arrowHeaderOffset, "arrow header");
-        arrowHeaderBuffer.position(ARROW_IPC_METADATA_SIZE_OFFSET);
-        int arrowMetadataSize = arrowHeaderBuffer.getInt();
-
-        resizeArrowMetadataBuffer(arrowMetadataSize);
-        arrowMetadataBuffer.rewind();
-        readFullyOrFail(
-                channel,
-                arrowMetadataBuffer,
-                arrowHeaderOffset + ARROW_HEADER_SIZE,
-                "arrow metadata");
-
-        arrowMetadataBuffer.rewind();
-        Message metadata = Message.getRootAsMessage(arrowMetadataBuffer);
-        ProjectedArrowBatch projectedArrowBatch =
-                projectArrowBatch(
-                        metadata,
-                        currentProjection.nodesProjection,
-                        currentProjection.buffersProjection,
-                        currentProjection.bufferCount);
-        long arrowBodyLength = projectedArrowBatch.bodyLength();
-
-        int newBatchSizeInBytes =
-                recordBatchHeaderSize
-                        + changeTypeBytes
-                        + currentProjection.arrowMetadataLength
-                        + (int) arrowBodyLength;
-
-        // create new arrow batch metadata which already projected
-        byte[] headerMetadata =
-                serializeArrowRecordBatchMetadata(
-                        projectedArrowBatch, arrowBodyLength, currentProjection.bodyCompression);
-        checkState(
-                headerMetadata.length == currentProjection.arrowMetadataLength,
-                "Invalid metadata length");
-
-        // update and copy log batch header
-        logHeaderBuffer.position(LENGTH_OFFSET);
-        logHeaderBuffer.putInt(newBatchSizeInBytes - LOG_OVERHEAD);
-
-        // For V1+ format, clear statistics information since projection removes statistics
-        LogRecordBatchFormat.clearStatisticsFromHeader(logHeaderBuffer, magic);
-
-        logHeaderBuffer.rewind();
-        byte[] logHeader = new byte[recordBatchHeaderSize];
-        logHeaderBuffer.get(logHeader);
-
-        // build log records
-        builder.addBytes(logHeader);
-        if (!isAppendOnly) {
-            builder.addBytes(channel, position + recordsStartOffset, changeTypeBytes);
-        }
-        builder.addBytes(headerMetadata);
-        final long bufferOffset = arrowHeaderOffset + ARROW_HEADER_SIZE + arrowMetadataSize;
-        projectedArrowBatch.buffers.forEach(
-                b -> builder.addBytes(channel, bufferOffset + b.getOffset(), (int) b.getSize()));
-
+        projectSingleBatch(channel, position, currentProjection, builder, Integer.MAX_VALUE);
         return builder.build();
     }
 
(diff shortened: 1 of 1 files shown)


fluss-client/src/main/java/org/apache/fluss/client/table/scanner/TableScan.java
```diff
@@ -116,7 +131,8 @@ public LogScanner createLogScanner() {
                 conn.getClientMetricGroup(),
                 conn.getOrCreateRemoteFileDownloader(),
                 projectedColumns,
-                schemaGetter);
+                schemaGetter,
+                recordBatchFilter);
```
wuchong, line 135: throw exception in all the other `createXxxScanner` if the `recordBatchFilter` is not null, because they all don't support the filter.

fluss-client/src/main/java/org/apache/fluss/client/table/scanner/Scan.java
```diff
@@ -60,6 +61,19 @@ public interface Scan {
...
+     * Returns a new scan from this that will apply the given predicate filter.
+     *
+     * <p>Note: the filter currently only supports record batch level filtering for log scanners,
+     * not row level filtering. The computing engine still needs to perform secondary filtering on
+     * the results. Batch scanners do not support filter pushdown.
+     *
+     * @param predicate the predicate to apply for record batch level filtering
+     */
+    default Scan filter(@Nullable Predicate predicate) {
+        throw new UnsupportedOperationException("Filter pushdown is not supported by this Scan.");
```
wuchong: We don't need to have default implementation, because it is not a user-defined interface. All the implementation of this interface are provided by Fluss itself, so a simple interface without default implemenation is much cleaner here.

fluss-common/src/test/java/org/apache/fluss/record/FileLogProjectionTest.java
```diff
@@ -524,4 +526,461 @@ private static void assertEquals(List<Object[]> actual, List<Object[]> expected)
...
+                    assertThat(recordCount).isEqualTo(TestData.DATA2.size());
+                }
+            }
+        }
+    }
+
+    @ParameterizedTest
+    @ValueSource(bytes = {LOG_MAGIC_VALUE_V0, LOG_MAGIC_VALUE_V1})
+    void testProjectRecordBatchNoStatisticsClearing(byte recordBatchMagic) throws Exception {
+        // Test that statistics clearing only happens for V2+ versions
```
wuchong: Since we have changed the stats to V1, we should update the tests to reflect that. IIUC, this should test only V0 and we may also need to update other tests.

fluss-server/src/main/java/org/apache/fluss/server/log/LogSegment.java
```diff
@@ -540,19 +647,154 @@ public FetchDataInfo read(
...
+        // Iterate batches from the translated position.
+        // Note: AbstractIterator doesn't implement AutoCloseable, so no explicit cleanup needed.
+        // The iterator's internal state will be garbage collected when this method returns.
+        AbstractIterator<FileChannelLogRecordBatch> iter =
+                fileLogRecords.batchIterator(startPosition, (int) maxPosition);
+
+        MultiBytesView.Builder builder = null;
+        int accumulatedSize = 0;
+        FileChannelLogRecordBatch firstIncludedBatch = null;
+        FileChannelLogRecordBatch lastIncludedBatch = null;
```
wuchong, line 632: lastIncludedBatch is not used, remove

fluss-server/src/main/java/org/apache/fluss/server/log/LogSegment.java
```diff
@@ -540,19 +647,154 @@ public FetchDataInfo read(
...
+                }
+                lastIncludedBatch = batch;
+                if (builder == null) {
+                    builder = MultiBytesView.builder();
+                }
+                builder.addBytes(fileLogRecords.channel(), batch.position(), batchSize);
+                accumulatedSize += batchSize;
+            } else {
+                // With projection: project first, then check size with projected size
+                BytesView projectedBytesView = projection.projectRecordBatch(batch);
```
wuchong, line 699: throw exception if `logFormat != LogFormat.ARROW` like how we do in `readWithoutFilter`.

fluss-server/src/main/java/org/apache/fluss/server/log/PredicateSchemaResolver.java
```diff
@@ -0,0 +1,138 @@
...
+
+    private final Predicate predicate;
+    private final int predicateSchemaId;
+    @Nullable private final Schema predicateSchema;
+    @Nullable private final SchemaGetter schemaGetter;
+
+    private final Map<Integer, Optional<Predicate>> cache = new HashMap<>();
+
+    public PredicateSchemaResolver(
+            Predicate predicate, int predicateSchemaId, @Nullable SchemaGetter schemaGetter) {
```
wuchong: It seems the `schemaGetter` is never null in production code, only tests set it to null. However, this makes the logic (`schemaGetter` null/non-null handling) confusing and complex.  , maybe we can force this field to be not null, and simplify the code.

fluss-server/src/main/java/org/apache/fluss/server/replica/Replica.java
```diff
@@ -1494,13 +1504,91 @@ private LogReadInfo readRecords(FetchParams fetchParams, LogTablet logTablet)
...
+                            predicateResolver);
+        } finally {
+            // Close readContext eagerly — it is only used for statistics extraction during
+            // batch filtering and is NOT referenced by the returned FetchDataInfo records.
+            if (readContext != null) {
+                try {
+                    readContext.close();
+                } catch (Exception e) {
+                    LOG.debug("Failed to close readContext for {}", tableBucket, e);
+                }
```
wuchong: nit: can use `IOUtils.closeQuietly(..)` to close it.

fluss-server/src/main/java/org/apache/fluss/server/replica/Replica.java
```diff
@@ -1494,13 +1504,91 @@ private LogReadInfo readRecords(FetchParams fetchParams, LogTablet logTablet)
...
+                resolvedFilter = null;
+                if (readContext != null) {
+                    try {
+                        readContext.close();
+                    } catch (Exception closeEx) {
+                        LOG.debug("Failed to close readContext for {}", tableBucket, closeEx);
+                    }
+                }
+                readContext = null;
+                predicateResolver = null;
```
wuchong: Would be better to extract them into a separate method to initialize the 3 variables. So if there is any exception, it's easy to fallback without affecting any local variables.

fluss-server/src/main/java/org/apache/fluss/server/log/LogTablet.java
```diff
@@ -395,13 +396,29 @@ public LogAppendInfo appendAsFollower(MemoryLogRecords records) throws Exception
...
     public FetchDataInfo read(
             long readOffset,
             int maxLength,
             FetchIsolation fetchIsolation,
             boolean minOneMessage,
-            @Nullable FileLogProjection projection)
+            @Nullable FileLogProjection projection,
+            @Nullable Predicate recordBatchFilter,
+            @Nullable LogRecordBatch.ReadContext readContext,
+            @Nullable PredicateSchemaResolver predicateResolver)
```
wuchong: Introducing three separate parameters is excessive. It would be better to encapsulate them within a dedicated structure, such as `FilterParams` or `FilterContext`. It will also be easier for us to handling nullability, because the 3 parameters should all be null or all be not-null (IIUC).

fluss-common/src/main/java/org/apache/fluss/record/FileLogRecords.java
```diff
@@ -309,8 +309,27 @@ public FileChannelChunk toChunk() {
...
+    /**
+     * Get an iterator over the record batches in the file, starting at a specific position and
+     * ending at a specific position.
+     *
+     * @param start the start position in the file
+     * @param endPosition the end position in the file, or -1 to use the default end
+     * @return An iterator over batches starting from {@code start}
+     */
+    public AbstractIterator<FileChannelLogRecordBatch> batchIterator(int start, int endPosition)
+            throws IOException {
```
wuchong: The `IOException` is never thrown from the method.

fluss-server/src/main/java/org/apache/fluss/server/log/LogSegment.java
```diff
@@ -540,19 +647,154 @@ public FetchDataInfo read(
...
+            // Apply filter using statistics. On any failure, fall back to including
+            // the batch so a single corrupt batch cannot break the entire fetch.
+            boolean include = true;
+            try {
+                Optional<LogRecordBatchStatistics> statsOpt = batch.getStatistics(readContext);
+                if (statsOpt.isPresent()) {
+                    LogRecordBatchStatistics stats = statsOpt.get();
+                    Predicate effectivePredicate =
+                            predicateResolver != null
+                                    ? predicateResolver.resolve(stats.getSchemaId())
```
wuchong: When `predicateResolver` will be null if this is a filter read? I though the 3 parameter should all be null or all be non-null.

fluss-server/src/main/java/org/apache/fluss/server/utils/ServerRpcMessageUtils.java
```diff
@@ -887,6 +893,22 @@ public static ProduceLogResponse makeProduceLogResponse(
...
     }
 
+    public static @Nullable Map<Long, FilterInfo> getTableFilterInfoMap(FetchLogRequest request) {
+        Map<Long, FilterInfo> result = null;
+        for (PbFetchLogReqForTable tableReq : request.getTablesReqsList()) {
+            if (tableReq.hasFilterPredicate()) {
+                if (result == null) {
+                    result = new HashMap<>();
+                }
+                int schemaId = tableReq.hasFilterSchemaId() ? tableReq.getFilterSchemaId() : -1;
```
wuchong: Is `-1` valid? If not, we should fail-fast by throwing exception.

fluss-rpc/src/main/java/org/apache/fluss/rpc/util/CommonRpcMessageUtils.java
```diff
@@ -202,8 +202,18 @@ public static FetchLogResultForBucket getFetchLogResultForBucket(
                         respForBucket.hasRecords()
                                 ? MemoryLogRecords.pointToByteBuffer(recordsBuffer)
                                 : MemoryLogRecords.EMPTY;
-                fetchLogResultForBucket =
-                        new FetchLogResultForBucket(tb, records, respForBucket.getHighWatermark());
+                if (respForBucket.hasFilteredEndOffset()
+                        && respForBucket.getFilteredEndOffset() >= 0) {
+                    fetchLogResultForBucket =
+                            new FetchLogResultForBucket(
+                                    tb,
```
wuchong, line 209: When `filteredEndOffset` is set, any records in the response are currently silently ignored. Although `records` are typically unset when `filteredEndOffset` is present, a specific edge case leads to inefficiency: if the first batch in a log file matches the filter but subsequent batches do not, the system returns only the next offset of the matched batch without including the `FilteredEndOffset`. Consequently, these already-filtered batches must be re-evaluated in the next request, resulting in unnecessary resource consumption and performance degradation.

fluss-common/src/main/java/org/apache/fluss/record/DefaultLogRecordBatchStatistics.java
```diff
@@ -147,6 +168,9 @@ public InternalRow getMaxValues() {
 
     @Override
     public Long[] getNullCounts() {
```
wuchong, line 149: I think we can optimize `getNullCounts()` return a primitive `int[]`. 

1. `int` count is enough as you only use 4 bytes for each field null count. 
2. `int` rather than boxed type `Integer` to avoid boxing (double size), as you use `-1` as non-exist. 

This can reduce a lot of memory size for `cachedNullCounts`, `statsNullCounts`. Especially for `cachedNullCounts` when total field count is very large (1000+).

fluss-common/src/main/java/org/apache/fluss/record/ArrowNullCountReader.java
```diff
@@ -0,0 +1,167 @@
...
+import java.util.Arrays;
+import java.util.List;
+import java.util.Objects;
+
+/**
+ * Utility for extracting null counts from Arrow RecordBatch FlatBuffer metadata. Used by both
+ * FileChannelLogRecordBatch and DefaultLogRecordBatch to read null counts from Arrow metadata
+ * instead of the statistics binary format (V2).
+ */
+class ArrowNullCountReader {
```
wuchong: Could you revert the commit of `get null count from arrow meta`, this pull request has already been very large. I think we can review this in an separate pull request. Besides, I have some concerns on the cache and deserialization performance.

fluss-common/src/main/java/org/apache/fluss/record/ArrowNullCountReader.java
```diff
@@ -0,0 +1,167 @@
...
+            recordBatch.nodes(node, nodeIndex);
+            nullCounts[i] = node.nullCount();
+        }
+        return nullCounts;
+    }
+
+    /** Cache key combining schemaId and statsIndexMapping. */
+    static final class FieldNodeMappingKey {
+        private final int schemaId;
+        private final int[] statsIndexMapping;
```
wuchong: The cache key lacks `tableId`, it has issues when there are multiple tables.

fluss-client/src/main/java/org/apache/fluss/client/table/scanner/log/LogFetcher.java
```diff
@@ -93,6 +95,9 @@ public class LogFetcher implements Closeable {
     //  bytes from remote file.
     private final LogRecordReadContext remoteReadContext;
     @Nullable private final Projection projection;
+    @Nullable private final Predicate recordBatchFilter;
```
wuchong: unused, remove

fluss-common/src/main/java/org/apache/fluss/record/DefaultLogRecordBatchStatistics.java
wuchong: Will this be addressed in a follow-up PR? Could you create an issue for this?

fluss-common/src/main/java/org/apache/fluss/record/LogRecordReadContext.java
```diff
@@ -247,6 +270,17 @@ public boolean isProjectionPushDowned() {
         return projectionPushDowned;
     }
 
+    /** Get the target schema ID. */
+    public int getTargetSchemaId() {
```
wuchong: unused, can be removed.

fluss-common/src/main/java/org/apache/fluss/record/LogRecordReadContext.java
```diff
@@ -247,6 +270,17 @@ public boolean isProjectionPushDowned() {
...
+        return targetSchemaId;
+    }
+
+    /** Get the schema getter. */
+    @Nullable
```
wuchong: It seems the `schemaGetter` is not-null?

fluss-server/src/main/java/org/apache/fluss/server/log/FilterContext.java
```diff
@@ -0,0 +1,59 @@
...
+    public Predicate getRecordBatchFilter() {
+        return recordBatchFilter;
+    }
+
+    public LogRecordReadContext getReadContext() {
+        return readContext;
+    }
+
+    @Nullable
+    public PredicateSchemaResolver getPredicateResolver() {
```
wuchong, line 61: Why and when the `PredicateSchemaResolver` can be nullable?

fluss-server/src/main/java/org/apache/fluss/server/replica/Replica.java
```diff
@@ -1494,16 +1503,85 @@ private LogReadInfo readRecords(FetchParams fetchParams, LogTablet logTablet)
...
+                            filterSchemaId,
+                            tableBucket);
+                } else {
+                    rowType = filterSchema.getRowType();
+                    schemaIdForContext = filterSchemaId;
+                }
+            } else {
+                rowType = tableInfo.getSchema().getRowType();
+                schemaIdForContext = tableInfo.getSchemaId();
+            }
```
wuchong: I think we don't allow `filterSchemaId` to be `< 0`, using latest schema id is not safe here, maybe throw an exception here?

fluss-server/src/main/java/org/apache/fluss/server/log/LogSegment.java
```diff
@@ -490,11 +503,81 @@ public FetchDataInfo read(
...
+            int maxSize,
+            long maxPosition,
+            boolean minOneMessage,
+            @Nullable FileLogProjection projection,
+            @Nullable Predicate recordBatchFilter,
+            @Nullable LogRecordReadContext readContext)
+            throws IOException {
+        FilterContext filterContext = null;
+        if (recordBatchFilter != null && readContext != null) {
+            filterContext = new FilterContext(recordBatchFilter, readContext, null);
```
wuchong: In this scenario, we should be able to derive the `PredicateSchemaResolver` directly within the `FilterContext`. Since all necessary parameters for the `PredicateSchemaResolver` can be obtained from `recordBatchFilter`, `readContext`, and `filterSchemaId`, moving its construction into `FilterContext` is logical. Relying on a nullable `PredicateSchemaResolver` introduces unnecessary risk and potential errors, as it may lead to unpredictable behavior, such as using an outdated schema to parse statistics.

fluss-common/src/main/java/org/apache/fluss/record/DefaultLogRecordBatchStatistics.java
platinumhamburg: I created a follow-up issue for this issue: https://github.com/apache/fluss/issues/3021

platinumhamburg:
> LGTM.
> 
> I appended a commit to refactor the `FileLogProjection` to share some code. Please take a look.

LGTM.

PR #2953, opened 2026-03-29
[client][backport] Skip blocked bucket instead of stopping drain loop

fresh-borzoni:
When a bucket reaches max in-flight during drain, `break` stops draining all remaining buckets on that node. Since buckets have independent sequence spaces,  change to `continue` so only the blocked bucket is skipped.   
 
 Backport related to this https://github.com/apache/fluss-rust/pull/404#discussion_r2868997578


fresh-borzoni:
@luoyuxia @wuchong @polyzos  PTAL 🙏

polyzos:
LGTM

PR #2972, opened 2026-04-01
[flink] Implement SupportsWatermarkPushDown for FlinkTableSource

### LINKED ISSUE #2971: [flink] FlinkTableSource should implement SupportsWatermarkPushDown
### Search before asking

### Description

Currently, `FlinkTableSource` does not implement the `SupportsWatermarkPushDown` interface. This causes two significant problems when a Fluss table has a `WATERMARK` definition in Flink SQL.

When `SupportsWatermarkPushDown` is **not** implemented, Flink's planner inserts a standalone `WatermarkAssigner` node between the table scan and downstream operators. The Flink optimization rule `PushWatermarkIntoTableSourceScanRule` only fires when the source implements `SupportsWatermarkPushDown`. Without it, the rule never fires and the `WatermarkAssigner` node remains as a separate operator in the query plan.

This breaks partition filter pushdown: Flink's `PushFilterIntoTableSourceScanRule` cannot push filter predicates past the `WatermarkAssigner` node into the source scan. As a result, a query like:

```sql
SELECT a, b, c FROM partitioned_table WHERE c = '2025'
```

produces a plan where the partition filter is **not** pushed into the `TableSourceScan`:

```
-- Without SupportsWatermarkPushDown (current behavior):
Filter(condition=[=(c, '2025')])
  WatermarkAssigner(...)
    TableSourceScan(...)   -- no filter, no partition pruning
```

```
-- With SupportsWatermarkPushDown (expected behavior):
TableSourceScan(
  watermark=[-(ts, 5000:INTERVAL SECOND)],
  watermarkEmitStrategy=[on-periodic],
  filter=[=(c, '2025')]    -- partition pruning applied
)
```

### Willingness to contribute

LiebingYu:
### Purpose

Linked issue: close #2971

### DIFF
--- a/fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/source/FlinkTableSource.java
+++ b/fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/source/FlinkTableSource.java
@@ -48,7 +48,9 @@ import org.apache.fluss.types.DataTypeChecks;
 import org.apache.fluss.types.RowType;
 
 import org.apache.flink.annotation.VisibleForTesting;
+import org.apache.flink.api.common.eventtime.WatermarkStrategy;
 import org.apache.flink.api.common.typeinfo.TypeInformation;
+import org.apache.flink.api.connector.source.Boundedness;
 import org.apache.flink.api.connector.source.Source;
 import org.apache.flink.streaming.api.datastream.DataStream;
 import org.apache.flink.streaming.api.environment.StreamExecutionEnvironment;
@@ -65,6 +67,7 @@ import org.apache.flink.table.connector.source.abilities.SupportsFilterPushDown;
 import org.apache.flink.table.connector.source.abilities.SupportsLimitPushDown;
 import org.apache.flink.table.connector.source.abilities.SupportsProjectionPushDown;
 import org.apache.flink.table.connector.source.abilities.SupportsRowLevelModificationScan;
+import org.apache.flink.table.connector.source.abilities.SupportsWatermarkPushDown;
 import org.apache.flink.table.connector.source.lookup.AsyncLookupFunctionProvider;
 import org.apache.flink.table.connector.source.lookup.LookupFunctionProvider;
 import org.apache.flink.table.connector.source.lookup.PartialCachingAsyncLookupProvider;
@@ -112,7 +115,8 @@ public class FlinkTableSource
                 LookupTableSource,
                 SupportsRowLevelModificationScan,
                 SupportsLimitPushDown,
-                SupportsAggregatePushDown {
+                SupportsAggregatePushDown,
+                SupportsWatermarkPushDown {
 
     public static final Logger LOG = LoggerFactory.getLogger(FlinkTableSource.class);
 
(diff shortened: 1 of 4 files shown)


luoyuxia (commented):
@loserwang1024  Please help review this

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/source/FlinkTableSource.java
```diff
@@ -373,7 +383,25 @@ public boolean isBounded() {
...
+                @Override
+                public DataStream<RowData> produceDataStream(
+                        ProviderContext providerContext, StreamExecutionEnvironment execEnv) {
+                    WatermarkStrategy<RowData> strategy =
+                            watermarkStrategy != null
+                                    ? watermarkStrategy
+                                    : WatermarkStrategy.noWatermarks();
+                    DataStreamSource<RowData> sourceStream =
+                            execEnv.fromSource(source, strategy, "FlussSource-" + tablePath);
+                    providerContext.generateUid(FLUSS_TRANSFORMATION).ifPresent(sourceStream::uid);
```
loserwang1024: I wonder what `providerContext.generateUid(FLUSS_TRANSFORMATION).ifPresent(sourceStream::uid);` is used for?

loserwang1024 (commented):
I have left a mirror comment.

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/source/FlinkTableSource.java
LiebingYu: I think this uid is important for Flink state management. Flink Kafka connector also set this uid: https://github.com/apache/flink-connector-kafka/blob/bd2fd2b2c8a48b7d2a05387fc4fda61e878c902f/flink-connector-kafka/src/main/java/org/apache/flink/streaming/connectors/kafka/table/KafkaDynamicSource.java#L265

loserwang1024: I worry that integrating this PR could break state restoration for prior job. @leonardBang @wuchong , CC, WDYT?

LiebingYu: After some investigation, I think this shouldn't be an issue in most Flink versions, except for version 2.3 and later. 

In Flink `CommonExecTableSourceScan`, `SourceFunctionProvider`([line 160](https://github.com/apache/flink/blob/ef532a23f102bd0562e7a792da38678efae410ee/flink-table/flink-table-planner/src/main/java/org/apache/flink/table/planner/plan/nodes/exec/common/CommonExecTableSourceScan.java#L160)) and `DataStreamScanProvider` ([line 185](https://github.com/apache/flink/blob/ef532a23f102bd0562e7a792da38678efae410ee/flink-table/flink-table-planner/src/main/java/org/apache/flink/table/planner/plan/nodes/exec/common/CommonExecTableSourceScan.java#L185)) will both execute `metadata.fill(sourceTransform);`. For `legacyUidsEnabled() == true` (Before Flink version 2.3), the `sourceTransform` is equivalent.

loserwang1024 (approved):
LGTM

loserwang1024:
@LiebingYu Would you like to rebase before merging it?

LiebingYu:
> @LiebingYu Would you like to rebase before merging it?

Sure, rebase is done.

PR #2973, opened 2026-04-01
[server] Allow enabling lakehouse for log tables without bucket key created before cluster-level lakehouse is enabled 

### LINKED ISSUE #2961: Allow enabling lakehouse for log tables without bucket key created before cluster-level lakehouse is enabled
## Search before asking

## Fluss version
main (development)

## Please describe the bug 🐞
For log tables without a bucket key, enabling lakehouse is currently rejected if the table was created before the cluster-level lakehouse mode was enabled.

The current validation was introduced to keep Fluss data distribution aligned with Paimon. That requirement makes sense for tables with a bucket key, where we want the bucket distribution to stay consistent between Fluss and Paimon.

However, for log tables without a bucket key, this restriction is unnecessarily strict. In this case we do not rely on distribution consistency with Paimon, so these tables should still be allowed to enable lakehouse even if they were created before lakehouse was enabled at the cluster level.

### Reproduction
1. Create a log table without a bucket key while cluster-level lakehouse is disabled.
2. Enable lakehouse at the cluster level.
3. Try to enable lakehouse for the previously created log table.
4. Observe that the operation is rejected.

### Expected behavior
Lakehouse should be allowed for legacy log tables without a bucket key.

### Actual behavior
The current validation rejects enabling lakehouse for these tables.

## Solution
Relax the validation for legacy log tables without a bucket key:

- keep the existing distribution-consistency check for log tables with a bucket key;
- allow enabling lakehouse for log tables without a bucket key, even if the table was created before cluster-level lakehouse was enabled.

luoyuxia:
### Purpose

Linked issue: close #2961

This PR allows enabling datalake for legacy log tables without bucket keys after cluster-level datalake has been enabled.

### Brief change log

- relax alter-table validation so legacy log tables without bucket keys can enable `table.datalake.enabled`
- keep the original restriction for tables that still require distribution alignment with the lake format
- backfill `table.datalake.format` before `preAlterTableProperties(...)` so the altered table descriptor, lake-table creation path, and persisted table metadata stay consistent
- add an IT case covering the legacy log-table-without-bucket-key flow and verifying the backfilled format

### Tests

- `./mvnw -pl fluss-client -am -Dtest=LakeEnableTableITCase -DfailIfNoTests=false clean test`

### API and Format

- no API change
- legacy log tables without bucket keys now persist `table.datalake.format` when datalake is enabled after cluster-level datalake is turned on

### Documentation

- no

### DIFF
--- a/fluss-client/src/test/java/org/apache/fluss/client/table/LakeEnableTableITCase.java
+++ b/fluss-client/src/test/java/org/apache/fluss/client/table/LakeEnableTableITCase.java
@@ -112,6 +112,65 @@ class LakeEnableTableITCase extends ClientToServerITCaseBase {
     }
 
     @Test
+    void testCanEnableDatalakeForLogTableWithoutBucketKeyCreatedBeforeClusterEnabledDatalake()
+            throws Exception {
+        String databaseName = "test_db";
+        String tableName =
+                "test_log_table_without_bucket_key_created_before_cluster_enabled_datalake";
+        TablePath tablePath = TablePath.of(databaseName, tableName);
+
+        admin.alterClusterConfigs(
+                        Collections.singletonList(
+                                new AlterConfig(
+                                        DATALAKE_FORMAT.key(), null, AlterConfigOpType.SET)))
+                .get();
+        assertThat(
+                        FLUSS_CLUSTER_EXTENSION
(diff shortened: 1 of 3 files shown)


beryllw:
LGTM

PR #2976, opened 2026-04-01
Introduce RPC Layer for KV Scan

polyzos:


polyzos:
@wuchong @platinumhamburg @leekeiabstraction PTAL

I had to also fix some tests in the PR, because the CI was quite unstable.. Let me know your thoughts if thats ok

fluss-rpc/src/main/proto/FlussApi.proto
```diff
@@ -300,6 +300,43 @@ message LimitScanResponse{
...
+}
+
+message ScanKvRequest {
+  // Mutually exclusive: either scanner_id (continuation) or bucket_scan_req (new scan).
+  optional bytes scanner_id = 1;
+  optional PbScanReqForBucket bucket_scan_req = 2;
+  // Monotonically increasing sequence number for in-order delivery validation.
+  optional int32 call_seq_id = 3;
+  // Maximum number of bytes of record data to return in this batch.
+  required int32 batch_size_bytes = 4;
```
fresh-borzoni: Wouldn't required blow up serialization in  close-scanner request?

fluss-common/src/main/java/org/apache/fluss/exception/ScannerNotFoundException.java
```diff
@@ -0,0 +1,27 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.fluss.exception;
+
+/** Exception thrown when a scanner is not found. */
+public class ScannerNotFoundException extends ApiException {
```
fresh-borzoni: I don't see it wired in Errors.java, what is the difference between this guy and ScannerNotFound and ScannerExpiredException?

fluss-common/src/main/java/org/apache/fluss/exception/InvalidScanRequestException.java
```diff
@@ -0,0 +1,31 @@
...
+ */
+
+package org.apache.fluss.exception;
+
+/**
+ * Thrown when a ScanKv request is malformed, for example when both {@code scanner_id} and {@code
+ * bucket_scan_req} are set simultaneously, or when neither is set, or when the {@code call_seq_id}
+ * is out of order.
+ */
+public class InvalidScanRequestException extends ApiException {
```
fresh-borzoni, line 28: @PublicEvolving in all error classes

fluss-rpc/src/main/proto/FlussApi.proto
```diff
@@ -300,6 +300,43 @@ message LimitScanResponse{
 }
 
 
+// Full KV scan request and response.
+// A new scan is initiated with bucket_scan_req; subsequent batches use scanner_id.
+message PbScanReqForBucket {
```
fresh-borzoni, line 305: nit: Mb PbBucketScanRequest reads better?

polyzos:
@fresh-borzoni Comments addressed.. for the `PbScanReqForBucket` i would like to keep it as is for now, unless others have objections, so there is consistency with the rest like `PbTableStatsReqForBucket` in terms of `ReqForBucket`

fluss-rpc/src/main/proto/FlussApi.proto
leekeiabstraction: +1

fluss-rpc/src/main/proto/FlussApi.proto
```diff
@@ -300,6 +300,43 @@ message LimitScanResponse{
...
+message ScanKvRequest {
+  // Mutually exclusive: either scanner_id (continuation) or bucket_scan_req (new scan).
+  optional bytes scanner_id = 1;
+  optional PbScanReqForBucket bucket_scan_req = 2;
+  // Monotonically increasing sequence number for in-order delivery validation.
+  optional int32 call_seq_id = 3;
+  // Maximum number of bytes of record data to return in this batch.
+  optional int32 batch_size_bytes = 4;
+  // If true, the server closes the scanner session immediately.
+  optional bool close_scanner = 5;
```
leekeiabstraction, line 322: We can probably split close out as eg closeScanRequest out as a request on its own so that this API is single responsibility.

fluss-server/src/test/java/org/apache/fluss/server/coordinator/TableManagerITCase.java
```diff
@@ -702,35 +702,46 @@ void testSchemaEvolution() throws Exception {
...
-                                new RowType(
-                                        true,
-                                        Arrays.asList(
-                                                DataTypes.FIELD("f0", DataTypes.STRING(), 3),
-                                                DataTypes.FIELD("f1", DataTypes.INT(), 4)))))
-                .isTrue();
+        // check metadata response - retry since coordinator needs time to fully load its state
+        // after restart (waitUntilCoordinatorServerElected only waits for ZK election, not for
+        // the full metadata/schema load to complete)
+        retry(
```
leekeiabstraction: Why are we updating schema evolution test case in this PR?

leekeiabstraction (commented):
TY for the PR. Left some comments

fluss-server/src/test/java/org/apache/fluss/server/coordinator/TableManagerITCase.java
polyzos: @leekeiabstraction see my comment above. I had some weird issues with the CI to make it pass

fluss-rpc/src/main/proto/FlussApi.proto
polyzos: @leekeiabstraction see my comment below for the naming

PR #2987, opened 2026-04-02
[spark][bugfix] Use pk index for comparator align with SortMergeReader

### LINKED ISSUE #2986: [spark][bug] Batch read PK table failed when use random column as primary key
I found a bug while testing reading PK tables, it fails when using the last column as the primary key, current cases in SparkPrimaryKeyTableReadTest all use first column and partition column as primary key.

```scala
test("Spark Read: primary key table with last pk") {
    withTable("t") {
      sql("CREATE TABLE t (id int, name string, pk int, pk2 string) TBLPROPERTIES('primary.key'='pk,pk2')")
      checkAnswer(sql("SELECT * FROM t"), Nil)
      sql("INSERT INTO t VALUES (1, 'a', 10, 'x'), (2, 'b', 20, 'y')")
      checkAnswer(sql("SELECT * FROM t ORDER BY id"), Row(1, "a", 10, "x") :: Row(2, "b", 20, "y") :: Nil)
    }
  }
```

above case will failed with 
```
Job aborted due to stage failure: Task 0 in stage 3.0 failed 1 times, most recent failure: Lost task 0.0 in stage 3.0 (TID 4) (192.168.0.116 executor driver): java.lang.ArrayIndexOutOfBoundsException: Index 2 out of bounds for length 2
	at org.apache.fluss.row.ProjectedRow.getInt(ProjectedRow.java:90)
	at org.apache.fluss.row.InternalRow.lambda$createFieldGetter$ff31e09f$6(InternalRow.java:198)
	at org.apache.fluss.row.encode.CompactedKeyEncoder.encodeKey(CompactedKeyEncoder.java:83)
	at org.apache.fluss.spark.read.FlussUpsertPartitionReader$$anon$1.compare(FlussUpsertPartitionReader.scala:113)
	at org.apache.fluss.spark.read.FlussUpsertPartitionReader$$anon$1.compare(FlussUpsertPartitionReader.scala:111)
	at org.apache.fluss.spark.utils.LogChangesIterator.hasSamePrimaryKey(LogChangesIterator.scala:117)
	at org.apache.fluss.spark.utils.LogChangesIterator.hasNext(LogChangesIterator.scala:85)
	at org.apache.fluss.client.table.scanner.SortMergeReader.readBatch(SortMergeReader.java:90)
	at org.apache.fluss.spark.read.FlussUpsertPartitionReader.initialize(FlussUpsertPartitionReader.scala:217)
	at org.apache.fluss.spark.read.FlussUpsertPartitionReader.<init>(FlussUpsertPartitionReader.scala:86)
	at org.apache.fluss.spark.read.FlussUpsertPartitionReaderFactory.createReader(FlussPartitionReaderFactory.scala:61)
```

_Originally posted by @Yohahaha in https://github.com/apache/fluss/issues/2523#issuecomment-3835480927_

Yohahaha:
### Purpose

Linked issue: close #2986

### Tests

`Spark Read: primary key table with random project`

### DIFF
--- a/fluss-spark/fluss-spark-common/src/main/scala/org/apache/fluss/spark/read/FlussUpsertPartitionReader.scala
+++ b/fluss-spark/fluss-spark-common/src/main/scala/org/apache/fluss/spark/read/FlussUpsertPartitionReader.scala
@@ -27,6 +27,7 @@ import org.apache.fluss.record.LogRecord
 import org.apache.fluss.row.{encode, InternalRow, KeyValueRow}
 import org.apache.fluss.spark.SparkFlussConf
 import org.apache.fluss.spark.utils.LogChangesIterator
+import org.apache.fluss.types.{DataField, RowType}
 import org.apache.fluss.utils.CloseableIterator
 
 import org.apache.spark.internal.Logging
@@ -65,13 +66,12 @@ class FlussUpsertPartitionReader(
     extraProjections ++= projection
     pkIndexes.foreach {
       pkIndex =>
-        projection.find(p => p == pkIndex) match {
-          case Some(index) =>
-            _pkProjection += index
-          case _ =>
+        projection.indexOf(pkIndex) match {
+          case -1 =>
             extraProjections += pkIndex
             _pkProjection += projection.length + i
             i += 1
+          case idx => _pkProjection += idx
         }
     }
     (extraProjections.toArray, _pkProjection.toArray)
@@ -99,9 +99,13 @@ class FlussUpsertPartitionReader(
 
   private def createSortMergeReader(): SortMergeReader = {
     // Create key encoder for primary keys
+    val pkIndexes = tableInfo.getSchema.getPrimaryKeyIndexes
+    val pkFields = new java.util.ArrayList[DataField]()
+    pkIndexes.foreach(i => pkFields.add(rowType.getFields.get(i)))
+    val pkRowType = new RowType(pkFields)
     val keyEncoder =
       encode.KeyEncoder.ofPrimaryKeyEncoder(
-        rowType,
+        pkRowType,
         tableInfo.getPhysicalPrimaryKeys,
         tableInfo.getTableConfig,
         tableInfo.isDefaultBucketKey)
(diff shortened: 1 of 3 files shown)


Yohahaha:
@YannByron @wuchong please help review this pr, thank you!

fluss-spark/fluss-spark-common/src/main/scala/org/apache/fluss/spark/read/FlussUpsertPartitionReader.scala
```diff
@@ -99,9 +99,13 @@ class FlussUpsertPartitionReader(
 
   private def createSortMergeReader(): SortMergeReader = {
     // Create key encoder for primary keys
+    val pkIndexes = tableInfo.getSchema.getPrimaryKeyIndexes
+    val pkFields = new java.util.ArrayList[DataField]()
+    pkIndexes.foreach(i => pkFields.add(rowType.getFields.get(i)))
+    val pkRowType = new RowType(pkFields)
     val keyEncoder =
       encode.KeyEncoder.ofPrimaryKeyEncoder(
-        rowType,
```
YannByron, line 104: I think what is needed here is the row type of input data, not just the row type of primary key.

fluss-spark/fluss-spark-common/src/main/scala/org/apache/fluss/spark/utils/LogChangesIterator.scala
```diff
@@ -48,10 +48,15 @@ case class LogChangesIterator(
...
 ) extends CloseableIterator[KeyValueRow] {
 
+  private val projectRow1 = ProjectedRow.from(pkProjection)
+  private val projectRow2 = ProjectedRow.from(pkProjection)
+
   // Sort the records by primary key and then by offset
   private val sortedLogRecords = logRecords.sortWith {
     case (record1, record2) =>
-      val keyComparison = comparator.compare(record1.getRow, record2.getRow)
+      val keyComparison = comparator.compare(
```
YannByron, line 57: Can you explain why this change is needed?

Yohahaha: SortMergeReader compare row with pk, but here compare with full row, so ut failed when pk col was not at begining.

fluss-spark/fluss-spark-common/src/main/scala/org/apache/fluss/spark/read/FlussUpsertPartitionReader.scala
Yohahaha: actually only pk col is needed, ut already covered this case.

YannByron:
+1.

PR #2995, opened 2026-04-03
[client] Support log scanner scan to arrow record batch

### LINKED ISSUE #2965: support log scanner scan data as arrow reacord batch
### Search before asking

### Description

currently, log scanner only scan data as ScanRecord, it should provide method to scan as ArrowReacordBatch

### Willingness to contribute

luoyuxia:
### Purpose

Linked issue: close #2965

Support `LogScanner` to scan log data as Arrow record batches (`ArrowBatchData`) instead of row-by-row `ScanRecord`, enabling consumers like the tiering service to process columnar Arrow batches directly.

### Brief change log

- Add `LogScannerImpl.pollRecordBatch(Duration)` returning `ArrowScanRecords` for ARROW-format append-only log tables.
- Add `ArrowBatchData` to hold scanned Arrow `VectorSchemaRoot` with log metadata and memory ownership.
- Add `LogRecordBatch.loadArrowBatch(ReadContext)` to load Arrow batch directly from batch memory.
- Introduce `ArrowRecordBatchContext` / `UnshadedArrowBatchAccess` to bridge shaded/unshaded Arrow types with batch-scoped child allocators.
- Add unshaded Arrow compression support and `UnshadedArrowReadUtils` for the unshaded read path with schema evolution support.
- Extract `AbstractLogFetchCollector<T, R>` to share fetch-collection logic between row-based and Arrow batch paths.
- Extract `CompletedFetch.nextFetchedBatch()` / `finishFetchedBatches()` and generalize `LogScannerImpl.doPoll()` for code reuse.

### Tests

- `LogScannerITCase.testPollArrowBatchesWithSchemaEvolution`
- `FileLogInputStreamTest.testLoadArrowBatchFromFileLogInputStream`

### API and Format

- New `@Internal` API: `LogScannerImpl.pollRecordBatch(Duration)`, `ArrowBatchData`, `ArrowScanRecords`.
- No storage format changes.

### Documentation

N/A

### DIFF
--- a/fluss-client/pom.xml
+++ b/fluss-client/pom.xml
@@ -49,6 +49,20 @@
             <version>${project.version}</version>
         </dependency>
 
+        <dependency>
+            <groupId>org.apache.arrow</groupId>
+            <artifactId>arrow-vector</artifactId>
+            <version>${arrow.version}</version>
+            <scope>provided</scope>
+        </dependency>
+
+        <dependency>
+            <groupId>org.apache.arrow</groupId>
+            <artifactId>arrow-memory-netty</artifactId>
+            <version>${arrow.version}</version>
+            <scope>provided</scope>
+        </dependency>
+
         <!-- test dependency -->
         <dependency>
             <groupId>org.apache.fluss</groupId>
--- a/fluss-client/pom.xml
+++ b/fluss-client/pom.xml
@@ -49,6 +49,10 @@
             <version>${project.version}</version>
         </dependency>
 
+        <!-- Unshaded Arrow dependencies for the Arrow scan batch API (ArrowBatchData).
+             Marked as "provided" to avoid Arrow class conflicts with consumers that
+             bring their own Arrow version. Currently only the internal tiering service
+             uses this API and is responsible for including Arrow on its classpath. -->
         <dependency>
             <groupId>org.apache.arrow</groupId>
             <artifactId>arrow-vector</artifactId>
--- a/fluss-client/src/main/java/org/apache/fluss/client/table/scanner/log/AbstractLogFetchCollector.java
+++ b/fluss-client/src/main/java/org/apache/fluss/client/table/scanner/log/AbstractLogFetchCollector.java
@@ -132,6 +132,11 @@ abstract class AbstractLogFetchCollector<T, R> {
             if (fetched.isEmpty()) {
                 throw e;
             }
+        } catch (Exception e) {
+            // Release any off-heap resources (e.g. Arrow buffers) held by
+            // already-fetched records before propagating the unexpected error.
+            closeFetchedRecords(fetched);
+            throw e;
         }
 
         return toResult(fetched);
@@ -241,4 +246,11 @@ abstract class AbstractLogFetchCollector<T, R> {
     protected abstract int recordCount(List<T> fetchedRecords);
 
     protected abstract R toResult(Map<TableBucket, List<T>> fetchedRecords);
+
+    /**
+     * Release resources held by fetched records on failure. The default implementation is a no-op,
+     * suitable for record types that are plain Java objects. Subclasses whose record types hold
+     * off-heap resources (e.g. Arrow buffers) should override this to close them.
+     */
+    protected void closeFetchedRecords(Map<TableBucket, List<T>> fetched) {}
 }
--- a/fluss-client/src/main/java/org/apache/fluss/client/table/scanner/log/AbstractLogFetchCollector.java
+++ b/fluss-client/src/main/java/org/apache/fluss/client/table/scanner/log/AbstractLogFetchCollector.java
@@ -85,7 +85,11 @@ abstract class AbstractLogFetchCollector<T, R> {
 
                     if (!completedFetch.isInitialized()) {
                         try {
-                            logFetchBuffer.setNextInLineFetch(initialize(completedFetch));
+                            CompletedFetch initialized = initialize(completedFetch);
+                            logFetchBuffer.setNextInLineFetch(initialized);
+                            if (initialized == null) {
+                                completedFetch.drain();
+                            }
                         } catch (Exception e) {
                             // Remove a completedFetch upon a parse with exception if
                             // (1) it contains no records, and
--- a/fluss-client/src/main/java/org/apache/fluss/client/table/scanner/log/AbstractLogFetchCollector.java
+++ b/fluss-client/src/main/java/org/apache/fluss/client/table/scanner/log/AbstractLogFetchCollector.java
@@ -1,0 +1,240 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.fluss.client.table.scanner.log;
+
+import org.apache.fluss.annotation.Internal;
+import org.apache.fluss.client.metadata.MetadataUpdater;
+import org.apache.fluss.config.ConfigOptions;
+import org.apache.fluss.config.Configuration;
+import org.apache.fluss.exception.AuthorizationException;
+import org.apache.fluss.exception.FetchException;
+import org.apache.fluss.metadata.TableBucket;
+import org.apache.fluss.metadata.TablePath;
+import org.apache.fluss.rpc.protocol.ApiError;
+import org.apache.fluss.rpc.protocol.Errors;
+
+import org.slf4j.Logger;
+
+import javax.annotation.Nullable;
+import javax.annotation.concurrent.ThreadSafe;
+
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+/** Shared implementation for polling completed fetches into scanner results. */
+@ThreadSafe
+@Internal
+abstract class AbstractLogFetchCollector<T, R> {
+    protected final Logger log;
+    protected final TablePath tablePath;
+    protected final LogScannerStatus logScannerStatus;
+    private final int maxPollRecords;
+    private final MetadataUpdater metadataUpdater;
+
+    protected AbstractLogFetchCollector(
+            Logger log,
+            TablePath tablePath,
+            LogScannerStatus logScannerStatus,
+            Configuration conf,
+            MetadataUpdater metadataUpdater) {
+        this.log = log;
+        this.tablePath = tablePath;
+        this.logScannerStatus = logScannerStatus;
+        this.maxPollRecords = conf.getInt(ConfigOptions.CLIENT_SCANNER_LOG_MAX_POLL_RECORDS);
+        this.metadataUpdater = metadataUpdater;
+    }
+
+    /**
+     * Return the fetched log records, empty the record buffer and update the consumed position.
+     *
+     * <p>NOTE: returning empty records guarantees the consumed position are NOT updated.
+     *
+     * @return The fetched records per partition
+     * @throws FetchException If there is OffsetOutOfRange error in fetchResponse and the
+     *     defaultResetPolicy is NONE
+     */
+    public R collectFetch(final LogFetchBuffer logFetchBuffer) {
+        Map<TableBucket, List<T>> fetched = new HashMap<>();
+        int recordsRemaining = maxPollRecords;
+
+        try {
+            while (recordsRemaining > 0) {
+                CompletedFetch nextInLineFetch = logFetchBuffer.nextInLineFetch();
+                if (nextInLineFetch == null || nextInLineFetch.isConsumed()) {
+                    CompletedFetch completedFetch = logFetchBuffer.peek();
+                    if (completedFetch == null) {
+                        break;
+                    }
+
+                    if (!completedFetch.isInitialized()) {
+                        try {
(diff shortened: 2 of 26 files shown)


fluss-client/pom.xml
```diff
@@ -49,6 +49,20 @@
...
+            <artifactId>arrow-vector</artifactId>
+            <version>${arrow.version}</version>
+            <scope>provided</scope>
+        </dependency>
+
+        <dependency>
+            <groupId>org.apache.arrow</groupId>
+            <artifactId>arrow-memory-netty</artifactId>
+            <version>${arrow.version}</version>
+            <scope>provided</scope>
```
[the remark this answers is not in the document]
luoyuxia: Thanks for pointing this out. This is intentional: we expose unshaded Arrow types in the public API, but avoid pulling Arrow transitively to reduce dependency
  conflicts.

  You are right that this creates a runtime requirement. We’ll document/enforce that applications using the Arrow APIs must include the corresponding Arrow
  dependencies themselves.

fluss-common/src/main/java/org/apache/fluss/record/UnshadedFlussVectorLoader.java
```diff
@@ -0,0 +1,132 @@
...
+
+        try {
+            for (int j = 0; j < bufferLayoutCount; ++j) {
+                ArrowBuf nextBuf = buffers.next();
+                ArrowBuf bufferToAdd =
+                        nextBuf.writerIndex() > 0L
+                                ? codec.decompress(vector.getAllocator(), nextBuf)
+                                : nextBuf;
+                ownBuffers.add(bufferToAdd);
+                if (this.decompressionNeeded) {
```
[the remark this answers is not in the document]
luoyuxia: Thanks for the suggestion. I double-checked this path, and we should keep the current retain() here.

  In this loader flow, decompress(...) consumes/closes the input buffer, so the extra retain() is needed to keep the original ArrowRecordBatch buffer ownership
  balanced until ArrowRecordBatch.close(). Narrowing it to bufferToAdd == nextBuf causes the original batch close path to over-release the buffer and leads to negative
  ref-count in the existing loader tests.

fluss-common/src/main/java/org/apache/fluss/utils/UnshadedArrowReadUtils.java
```diff
@@ -0,0 +1,121 @@
...
+            initFieldVector(targetVector, rowCount);
+            if (columnProjection[i] < 0) {
+                fillNullVector(targetVector, rowCount);
+            } else {
+                FieldVector sourceVector = sourceRoot.getVector(columnProjection[i]);
+                for (int rowId = 0; rowId < rowCount; rowId++) {
+                    targetVector.copyFromSafe(rowId, rowId, sourceVector);
+                }
+                targetVector.setValueCount(rowCount);
+            }
```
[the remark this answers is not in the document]
luoyuxia: Consider it's not a hot path.

luoyuxia:
todo: add slice record batch for fetch offset

luoyuxia:
cc @loserwang1024 @beryllw @wuchong

fluss-client/src/main/java/org/apache/fluss/client/table/scanner/log/LogScannerImpl.java
```diff
@@ -255,12 +256,66 @@ private ScanRecords pollForFetches() {
 
         // send any new fetches (won't resend pending fetches).
         logFetcher.sendFetches();
-
         return logFetcher.collectFetch();
     }
 
+    private ArrowScanRecords pollForRecordBatches() {
```
loserwang1024, line 262: @luoyuxia 
> Add LogScannerImpl.pollRecordBatch(Duration) returning ArrowScanRecords for ARROW-format append-only log tables.

I see it's a internal api only for arrow type. Would you like to make a more abstract public api for `pollRecordBatch`

fluss-common/src/main/java/org/apache/fluss/record/LogRecordReadContext.java
```diff
@@ -359,9 +404,76 @@ public ProjectedRow getOutputProjectedRow(int schemaId) {
 
     public void close() {
         vectorSchemaRootMap.values().forEach(VectorSchemaRoot::close);
+        vectorSchemaRootMap.clear();
         if (bufferAllocator != null) {
             bufferAllocator.close();
         }
+
+        IOUtils.closeQuietly(unshadedBufferAllocator);
```
fresh-borzoni: Batches borrow the scanner's memory, so they must be closed before the scanner. 
Get it backwards and the memory leaks silently, then the batch's close() throws. 
I understand it's for internal use(tiering), but still a bit non-obvious and error-prone.

fresh-borzoni (commented):
@luoyuxia Ty for the PR, LGTM overall, one minor comment
Also shall we file followups based on TODOs?

luoyuxia:
@fresh-borzoni Thanks for your comments. Comments addressed.

fluss-client/src/main/java/org/apache/fluss/client/table/scanner/log/CompletedFetch.java
```diff
@@ -262,33 +263,82 @@ private void fetchRecord(List<ScanRecord> scanRecords) throws Exception {
         cachedRecordException = null;
     }
 
+    /**
+     * The {@link LogRecordBatch batches} are loaded as {@link ArrowBatchData Arrow batches} and
+     * returned.
+     *
+     * @param maxRecords The maximum number of records to return; the actual number returned may be
```
beryllw: > "the actual number returned may be less"

Looking at the loop, batches aren't split — once recordsFetched < maxRecords we pull a full batch and add it whole. So the returned count may be exceed maxRecords. Might be worth updating the Javadoc to reflect this (soft upper bound, not hard).

fluss-common/src/main/java/org/apache/fluss/record/LogRecordReadContext.java
```diff
@@ -268,14 +281,22 @@ public static LogRecordReadContext createCompactedRowReadContext(
...
+                false,
+                schemaGetter);
     }
 
     private LogRecordReadContext(
             LogFormat logFormat,
             RowType targetDataRowType,
             int targetSchemaId,
             BufferAllocator bufferAllocator,
+            int[] selectedFields,
```
beryllw: The new int[] selectedFields constructor parameter doesn't appear to be assigned to any field.Is this an incomplete change, or can the parameter be removed?

fluss-client/src/main/java/org/apache/fluss/client/table/scanner/log/CompletedFetch.java
```diff
@@ -262,33 +263,82 @@ private void fetchRecord(List<ScanRecord> scanRecords) throws Exception {
...
+                        arrowBatchData.close();
+                        continue;
+                    }
+                    arrowBatchData = arrowBatchData.sliceAndTransferOwnership(skipRows);
+                }
+
+                arrowBatches.add(arrowBatchData);
+                recordsRead += arrowBatchData.getRecordCount();
+                recordsFetched += arrowBatchData.getRecordCount();
+                nextFetchOffset = batch.nextLogOffset();
```
beryllw, line 318: Could we verify that a mid-loop failure in fetchArrowBatches doesn't leave nextFetchOffset ahead of records that were closed without being delivered? Looks like batches 1..N-1 may be lost if batch N throws."

fluss-client/src/main/java/org/apache/fluss/client/table/scanner/log/AbstractLogFetchCollector.java
```diff
@@ -0,0 +1,244 @@
...
+                            mergedRecords.addAll(currentRecords);
+                            mergedRecords.addAll(records);
+                            fetched.put(tableBucket, mergedRecords);
+                        }
+
+                        recordsRemaining -= recordCount(records);
+                    }
+                }
+            }
+        } catch (FetchException e) {
```
beryllw, line 132: Since ArrowBatchData owns off-heap memory, would you consider widening the catch (or wrapping in try/finally) so accumulated batches are closed when a non-FetchException propagates?

fluss-common/src/main/java/org/apache/fluss/record/LogRecordReadContext.java
```diff
@@ -359,9 +404,86 @@ public ProjectedRow getOutputProjectedRow(int schemaId) {
...
+            try {
+                unshadedBufferAllocator.close();
+            } catch (Exception e) {
+                throw new RuntimeException(
+                        "Failed to close Arrow buffer allocator. "
+                                + "Arrow batches returned by pollRecordBatch() must be closed "
+                                + "before closing the scanner.",
+                        e);
+            }
+            unshadedBufferAllocator = null;
```
beryllw: `close()` writes unshadedBufferAllocator = null without unshadedArrowResourceLock, racing with `getOrCreateUnshadedBufferAllocator()`.

fluss-client/src/main/java/org/apache/fluss/client/table/scanner/log/ArrowScanRecords.java
```diff
@@ -0,0 +1,107 @@
...
+
+import java.util.Collections;
+import java.util.Iterator;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+
+/** A container that holds the scanned Arrow batches per bucket for a particular table. */
+@Internal
+public class ArrowScanRecords implements Iterable<ArrowBatchData> {
```
fresh-borzoni: ArrowScanRecords is a plain Iterable, but each ArrowBatchData holds off-heap memory the caller must close. If a caller errors mid-iteration (e.g. a lake write fails), then it leaks. 
Everywhere else Fluss seems to return CloseableIterator. 

Could we do the same here so cleanup is automatic?

fresh-borzoni (commented):
@beryllw Nice points 👍 
I looked through, and I have an additional comment, PTAL @luoyuxia

beryllw:
@loserwang1024 cc

fluss-client/src/main/java/org/apache/fluss/client/table/scanner/log/CompletedFetch.java
```diff
@@ -262,33 +263,89 @@ private void fetchRecord(List<ScanRecord> scanRecords) throws Exception {
...
+    /**
+     * The {@link LogRecordBatch batches} are loaded as {@link ArrowBatchData Arrow batches} and
+     * returned.
+     *
+     * @param maxRecords A soft upper bound on the number of records to return. Because batches are
+     *     returned whole (never split), the actual number of records may exceed this value. At
+     *     least one batch is always returned if available, even if it alone exceeds the limit.
+     * @return {@link ArrowBatchData Arrow batches}
+     */
+    public List<ArrowBatchData> fetchArrowBatches(int maxRecords) {
```
loserwang1024: this method no need to be expected as public. pacakge is enough.

fluss-client/src/main/java/org/apache/fluss/client/table/scanner/log/CompletedFetch.java
```diff
@@ -262,33 +263,89 @@ private void fetchRecord(List<ScanRecord> scanRecords) throws Exception {
...
+                    int skipRows = (int) (nextFetchOffset - batchBaseOffset);
+                    if (skipRows >= arrowBatchData.getRecordCount()) {
+                        arrowBatchData.close();
+                        continue;
+                    }
+                    arrowBatchData = arrowBatchData.sliceAndTransferOwnership(skipRows);
+                }
+
+                arrowBatches.add(arrowBatchData);
+                recordsRead += arrowBatchData.getRecordCount();
```
loserwang1024, line 316: recordsRead += arrowBatchData.getRecordCount() - skipRows

fluss-client/src/main/java/org/apache/fluss/client/table/scanner/log/AbstractLogFetchCollector.java
```diff
@@ -0,0 +1,256 @@
...
+                            fetchOffset, tb, error.exception(errorMessage)));
+        } else {
+            throw new FetchException(
+                    String.format(
+                            "Unexpected error code %s while fetching at offset %s from bucket %s: %s",
+                            error, fetchOffset, tb, error.exception(errorMessage)));
+        }
+    }
+
+    protected abstract List<T> fetchRecords(CompletedFetch nextInLineFetch, int maxRecords);
```
loserwang1024: Most code of `org.apache.fluss.client.table.scanner.log.LogFetchCollector#fetchRecords` is same, only `nextInLineFetch.fetchArrowBatches` and `fetchRecords` is different, why not provide a protected class  about it and merge the same logic to AbstractLogFetchCollector.

fresh-borzoni (approved):
@luoyuxia Ty, LGTM 👍

loserwang1024 (commented):
I have left some comment

fluss-client/src/main/java/org/apache/fluss/client/table/scanner/log/CompletedFetch.java
fresh-borzoni: I think - arrowBatchData is the post-slice batch already

luoyuxia: +1

fluss-client/src/main/java/org/apache/fluss/client/table/scanner/log/LogScannerImpl.java
luoyuxia: It's expected to make it an an internal method. I don't want to expose it to public curerntly. Let's just expose it to internal lake tiering module.

fresh-borzoni: Might be a good idea, potentially spark can use this API as well, though it would probably require some changes and projection pushdown support

fresh-borzoni: But I agree, it's a scope creep

luoyuxia:
@loserwang1024 Comments addressed. Please help review again

loserwang1024 (approved):
LGTM

luoyuxia:
@wuchong I'm going to merge this since this only affects internal implementation and does not introduce any public API changes, the risk and impact scope are relatively small. You can still left comment if you find any

PR #2997, opened 2026-04-04
[filesystem][bug] Fix client.fs.* config prefix not stripped before passing to…

### LINKED ISSUE #2996: [filesystem][bug] client.fs.* config silently ignored by all filesystem plugins
### Search before asking

### Fluss version

0.9.0 (latest release)

### Please describe the bug 🐞

Filesystem config passed via `client.fs.*` prefix (introduced in #938) is silently ignored by all filesystem plugins.
Two issues:                                                                                                                  
   
  1. `FlussConnection` uses `extractPrefix()` instead of `extractAndRemovePrefix()`, keeping the `client.fs.` prefix intact. No plugin recognizes keys starting with `client.fs.`.                                                                                                                        
  2. Even after fixing (1), OSS and OBS plugins pass config keys through to Hadoop as-is without rewriting to the Hadoop format. For example, `oss.endpoint` stays `oss.endpoint` but Hadoop expects `fs.oss.endpoint`. S3, Azure, and GS plugins correctly rewrite (e.g., `s3.endpoint` -> `fs.s3a.endpoint`), but OSS and OBS don't.
                                                                                                                                                                                                                                                     
**Example:** User sets `client.fs.s3.path-style-access = true` in catalog config. It's accepted without error but silently ignored. S3 client uses virtual-hosted style, job hangs or fails with DNS errors against S3-compatible stores.          
   


### Solution

`FlussConnection`: use `extractAndRemovePrefix()` to strip the `client.fs.` prefix
Add short prefixes (`oss.`, `obs.`, `azure.`) to plugin prefix lists                                                                                                                                                                             Make OSS and OBS plugins rewrite keys to Hadoop format, consistent with S3/Azure/GS                                                                                                                                                              
              

fresh-borzoni:
## Summary

The `client.fs.*` prefix convention allows Flink clients to pass filesystem-specific config to Fluss. However, the config was silently ignored by all filesystem plugins due to two bugs.                                     
                                                                                                                                                                                                                        
This PR fixes both issues and adds short prefix support (`oss.`, `obs.`, `azure.`) to the remaining plugins, making the config convention work consistently across all filesystems.                                                                
   
Fixes #2996

### DIFF
--- a/fluss-filesystems/fluss-fs-azure/src/main/java/org/apache/fluss/fs/azure/token/AzureDelegationTokenProvider.java
+++ b/fluss-filesystems/fluss-fs-azure/src/main/java/org/apache/fluss/fs/azure/token/AzureDelegationTokenProvider.java
@@ -57,8 +57,8 @@ public class AzureDelegationTokenProvider {
 
         LOG.info("Setting the endpoint key " + endpointKey);
 
-        if (conf.get(endpointKey) != null) {
-            additionInfos.put(endpointKey, conf.get(endpointKey));
+        if (authEndpoint != null) {
+            additionInfos.put(endpointKey, authEndpoint);
         }
     }
 


fresh-borzoni:
@luoyuxia @polyzos PTAL 🙏

polyzos:
LGTM.. thank you for your contribution

fresh-borzoni:
@affo @morazow  Can you pls take a look?

fluss-filesystems/fluss-fs-obs/src/main/java/org/apache/fluss/fs/obs/OBSFileSystemPlugin.java
```diff
@@ -41,7 +41,9 @@ public class OBSFileSystemPlugin implements FileSystemPlugin {
      * In order to simplify, we make fluss obs configuration keys same with hadoop obs module. So,
      * we add all configuration key with prefix `fs.obs` in fluss conf to hadoop conf
```
morazow: Javadoc comment needs to be updated 🤝 

Maybe remove the `fs.obs` and put generic description.

fluss-filesystems/fluss-fs-oss/src/main/java/org/apache/fluss/fs/oss/OSSFileSystemPlugin.java
```diff
@@ -54,7 +54,9 @@ public class OSSFileSystemPlugin implements FileSystemPlugin {
      * In order to simplify, we make fluss oss configuration keys same with hadoop oss module. So,
      * we add all configuration key with prefix `fs.oss` in fluss conf to hadoop conf
```
morazow: Here also, Javadoc update required.

fluss-filesystems/fluss-fs-azure/src/main/java/org/apache/fluss/fs/azure/AzureFileSystemPlugin.java
```diff
@@ -43,7 +43,7 @@
  * (abfs, abfss, wasb, wasbs) based on Azure HDFS support in the hadoop-azure module.
  */
 abstract class AzureFileSystemPlugin implements FileSystemPlugin {
-    private static final String[] FLUSS_CONFIG_PREFIXES = {"fs.azure."};
+    private static final String[] FLUSS_CONFIG_PREFIXES = {"azure.", "fs.azure."};
```
morazow, line 46: I'd suggest to extend the corresponding [AzureFileSystemPluginTest.java#L42](https://github.com/fresh-borzoni/fluss/blob/311304d48871f8709daef81d141eca8f609378f3/fluss-filesystems/fluss-fs-azure/src/test/java/org/apache/fluss/fs/azure/AzureFileSystemPluginTest.java#L42) test with new key.

Same for OBS and OSS changes 👍

fluss-filesystems/fluss-fs-azure/src/main/java/org/apache/fluss/fs/azure/AzureFileSystemPlugin.java
morazow: Is this a bug from before?

With the normalization changes in `getHadoopConfiguration` method, should this be `hadoopConfig` instead of `flussConfig`?

fresh-borzoni: Good catch, ty, fixed this pre-existing bug as well 👍

fluss-filesystems/fluss-fs-azure/src/main/java/org/apache/fluss/fs/azure/AzureFileSystemPlugin.java
fresh-borzoni: I added this test, but there are no OBS/OSS plugin dedicated tests, so probably out of scope here.

fresh-borzoni:
@morazow Ty for the review, addressed, PTAL 🙏

fluss-filesystems/fluss-fs-azure/src/main/java/org/apache/fluss/fs/azure/token/AzureDelegationTokenProvider.java
```diff
@@ -49,15 +46,19 @@ public class AzureDelegationTokenProvider {
...
-        this.clientId = conf.get(CLIENT_ID);
-        this.clientSecret = conf.get(CLIENT_SECRET);
-        this.authEndpoint = conf.get(ENDPOINT_KEY);
+        String clientIdKey = AzureFileSystemOptions.CLIENT_ID.key();
+        String clientSecretKey = AzureFileSystemOptions.CLIENT_SECRET.key();
+        String endpointKey = AzureFileSystemOptions.ENDPOINT_KEY.key();
+
+        this.clientId = conf.get(clientIdKey);
+        this.clientSecret = conf.get(clientSecretKey);
+        this.authEndpoint = conf.get(endpointKey);
```
morazow, line 55: I think we can inline the keys in the `get(...)` calls? Since the keys are not used afterwards.

fresh-borzoni: Well, endpointKey is used later a couple of times in logs and also in additionInfos, so it's bit verbose, but at least consistent and no surprises. The alternative might be more confusing, so I'd prefer this way tbh

I cleaned up `conf.get` for additionInfos though :)

morazow: Ahh yeah, right. Yes, no looks fine from my side 🤝

morazow: I already approved it 👍

affo:
Btw, there is also something similar happening for lake tiering for which you have to specify twice a prefix, e.g.: `datalake.iceberg.iceberg....`, I will try to find a bit of space to address that 😓

PR #3000, opened 2026-04-04
[server][bug] Fix OOB in partial update/delete after ADD COLUMN

fresh-borzoni:
## Summary
closes https://github.com/apache/fluss/issues/2988

When a table has partial updates enabled and ADD COLUMN is performed, reading old rows from RocksDB with fewer fields than the latest schema causes ArrayIndexOutOfBoundsException in PartialUpdater. 

Fix by treating missing fields as null


fresh-borzoni:
@wuchong @polyzos PTAL 🙏

PR #3007, opened 2026-04-07
[client] Implement adaptive fetch rate control for LogScanner

### LINKED ISSUE #3006: [client] Implement adaptive fetch rate control for LogScanner to reduce overhead on partitioned tables
### Search before asking

### Motivation

For partitioned tables (e.g., hourly partitions with 256 buckets, retaining 24 hours of data), a streaming LogScanner subscribes to all partitions, resulting in 256 × 24 = 6144 buckets. However, only the latest partition's 256 buckets actively receive data. The current LogFetcher implementation treats all buckets equally, sending fetch requests to every bucket each round regardless of whether they have data. This causes unnecessary CPU usage, network overhead, and wasted fetch requests on inactive partitions.

### Solution

_No response_

### Anything else?

_No response_

### Willingness to contribute

swuferhong:
**(The sections below can be removed for hotfixes or typos)**
-->

### Purpose

Linked issue: close #3006 

For partitioned tables with many inactive partitions, the LogFetcher
previously sent fetch requests to all subscribed buckets equally,
wasting CPU and network resources on empty partitions.

Introduce BucketFetchRateController that uses exponential backoff to
reduce fetch frequency for buckets that consistently return no data.
Buckets that return data are always fetched at full frequency, and a
single non-empty fetch immediately resets the backoff.

Backoff schedule: 1, 2, 4, 8, 16, 32 rounds (configurable max).

New config options:
- client.scanner.log.adaptive-fetch.enabled (default: true)
- client.scanner.log.adaptive-fetch.max-skip-rounds (default: 32)

### DIFF
--- a/fluss-client/src/main/java/org/apache/fluss/client/table/scanner/log/BucketFetchRateController.java
+++ b/fluss-client/src/main/java/org/apache/fluss/client/table/scanner/log/BucketFetchRateController.java
@@ -1,0 +1,154 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.fluss.client.table.scanner.log;
+
+import org.apache.fluss.annotation.Internal;
+import org.apache.fluss.annotation.VisibleForTesting;
+import org.apache.fluss.metadata.TableBucket;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * Controls the fetch rate for individual buckets based on the amount of data returned in recent
+ * fetches.
+ *
+ * <p>For buckets that consistently return data, they are fetched every round. For buckets that
+ * return no data for consecutive fetches, the fetch frequency is progressively reduced using
+ * exponential backoff. This is particularly useful for partitioned tables where only the latest
+ * partitions have active data, avoiding wasted CPU and network resources on empty partitions.
+ *
+ * <p>The backoff schedule is: after 1 empty fetch, skip 1 round; after 2 empty fetches, skip 2
+ * rounds; after 3, skip 4 rounds; and so on (powers of 2), up to {@code maxSkipRounds}. Any fetch
+ * that returns data immediately resets the backoff to zero.
+ *
+ * <p>This class is NOT thread-safe. Callers must ensure proper synchronization.
+ */
+@Internal
+class BucketFetchRateController {
+
+    private static final Logger LOG = LoggerFactory.getLogger(BucketFetchRateController.class);
+
+    /** Maximum exponent for the exponential backoff (2^5 = 32). */
+    private static final int MAX_BACKOFF_SHIFT = 5;
+
+    private final int maxSkipRounds;
+    private final Map<TableBucket, BucketFetchState> bucketStates;
+
+    BucketFetchRateController(int maxSkipRounds) {
+        this.maxSkipRounds = maxSkipRounds;
+        this.bucketStates = new HashMap<>();
+    }
+
+    /**
+     * Determines whether the given bucket should be included in the current fetch round.
+     *
+     * <p>If the bucket is in a cool down period (i.e., it has been returning empty results), this
+     * method decrements the remaining skip count and returns {@code false}. Otherwise, it returns
+     * {@code true} indicating the bucket should be fetched.
+     *
+     * @param tableBucket the bucket to check
+     * @return {@code true} if the bucket should be fetched in this round
+     */
+    boolean shouldFetch(TableBucket tableBucket) {
+        BucketFetchState state = bucketStates.get(tableBucket);
+        if (state == null) {
+            return true;
+        }
+        if (state.remainingSkipRounds > 0) {
+            state.remainingSkipRounds--;
+            LOG.trace(
+                    "Adaptive fetch: skipping bucket {} (consecutive empty fetches: {}, "
+                            + "remaining skip rounds: {})",
+                    tableBucket,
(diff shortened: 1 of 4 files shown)


fluss-client/src/main/java/org/apache/fluss/client/table/scanner/log/BucketFetchRateController.java
```diff
@@ -0,0 +1,154 @@
...
+ *
+ * <p>This class is NOT thread-safe. Callers must ensure proper synchronization.
+ */
+@Internal
+class BucketFetchRateController {
+
+    private static final Logger LOG = LoggerFactory.getLogger(BucketFetchRateController.class);
+
+    /** Maximum exponent for the exponential backoff (2^5 = 32). */
+    private static final int MAX_BACKOFF_SHIFT = 5;
```
fresh-borzoni, line 51: We also have max_skip_rounds, so with this we have weird situation where user expects config to work, but it's capped at 32.

fluss-client/src/main/java/org/apache/fluss/client/table/scanner/log/BucketFetchRateController.java
```diff
@@ -0,0 +1,154 @@
...
+     * Determines whether the given bucket should be included in the current fetch round.
+     *
+     * <p>If the bucket is in a cool down period (i.e., it has been returning empty results), this
+     * method decrements the remaining skip count and returns {@code false}. Otherwise, it returns
+     * {@code true} indicating the bucket should be fetched.
+     *
+     * @param tableBucket the bucket to check
+     * @return {@code true} if the bucket should be fetched in this round
+     */
+    boolean shouldFetch(TableBucket tableBucket) {
```
fresh-borzoni, line 71: Is it intentional that a single empty fetch arms a skip? 

A streaming scanner caught up to HW alternates "batch / empty at HW" every poll, this throttles the empty half even though the bucket is active. 
Worst-case new-data latency becomes ~(max-skip × poll interval). 

If that's by design, worth a comment in the Javadoc

fluss-client/src/main/java/org/apache/fluss/client/table/scanner/log/BucketFetchRateController.java
```diff
@@ -0,0 +1,154 @@
...
+                    "Adaptive fetch: bucket {} has no data, entering cool down "
+                            + "(consecutive empty fetches: {}, will skip {} rounds)",
+                    tableBucket,
+                    state.consecutiveEmptyFetches,
+                    state.remainingSkipRounds);
+        }
+    }
+
+    /** Removes the tracking state for the given bucket. */
+    void removeBucket(TableBucket tableBucket) {
```
fresh-borzoni, line 126: Do we actually call it? It seems that without it being called we have a leak

fluss-client/src/main/java/org/apache/fluss/client/table/scanner/log/BucketFetchRateController.java
```diff
@@ -0,0 +1,154 @@
...
+        bucketStates.remove(tableBucket);
+    }
+
+    /** Resets all tracking state. */
+    void reset() {
```
fresh-borzoni, line 131: ditto

fluss-client/src/main/java/org/apache/fluss/client/table/scanner/log/LogFetcher.java
```diff
@@ -383,15 +393,17 @@ private synchronized void handleFetchLogResponse(
...
+                            hasData = true;
                             pendRemoteFetches(
                                     fetchResultForBucket.remoteLogFetchInfo(),
                                     fetchOffset,
                                     fetchResultForBucket.getHighWatermark());
                         } else {
                             LogRecords logRecords = fetchResultForBucket.recordsOrEmpty();
-                            boolean hasRecords = !MemoryLogRecords.EMPTY.equals(logRecords);
-                            if (hasRecords) {
+                            hasData = !MemoryLogRecords.EMPTY.equals(logRecords);
```
fresh-borzoni, line 405: After the rebase and #2951 #3032 changes, there is an issue  the new `hasData = !MemoryLogRecords.EMPTY.equals(logRecords)` as it  treats filtered-empty the same as truly empty, so an actively-producing bucket behind a selective WHERE gets throttled.

So better to rebase and reason about it

fluss-client/src/main/java/org/apache/fluss/client/table/scanner/log/LogFetcher.java
```diff
@@ -538,8 +561,20 @@ Map<Integer, FetchLogRequest> prepareFetchLogRequests() {
         }
 
         if (readyForFetchCount == 0) {
+            if (skippedByAdaptiveFetch > 0) {
+                LOG.info(
```
fresh-borzoni, line 565: it might be very noisy, mb debug?

fluss-client/src/main/java/org/apache/fluss/client/table/scanner/log/LogFetcher.java
```diff
@@ -538,8 +561,20 @@ Map<Integer, FetchLogRequest> prepareFetchLogRequests() {
...
+                        "No fetch requests prepared, {} buckets skipped by adaptive fetch",
+                        skippedByAdaptiveFetch);
+            }
             return Collections.emptyMap();
         } else {
+            if (skippedByAdaptiveFetch > 0) {
+                LOG.info(
```
fresh-borzoni, line 572: ditto

fluss-client/src/main/java/org/apache/fluss/client/table/scanner/log/BucketFetchRateController.java
```diff
@@ -0,0 +1,154 @@
...
+                        tableBucket,
+                        state.consecutiveEmptyFetches);
+            }
+            state.consecutiveEmptyFetches = 0;
+            state.remainingSkipRounds = 0;
+        } else {
+            state.consecutiveEmptyFetches++;
+            int shift = Math.min(state.consecutiveEmptyFetches - 1, MAX_BACKOFF_SHIFT);
+            state.remainingSkipRounds = Math.min(1 << shift, maxSkipRounds);
+            LOG.info(
```
fresh-borzoni, line 116: ditto about logging noise

fluss-client/src/main/java/org/apache/fluss/client/table/scanner/log/BucketFetchRateController.java
```diff
@@ -0,0 +1,154 @@
...
+     *
+     * @param tableBucket the bucket that was fetched
+     * @param hasRecords {@code true} if the fetch returned any records
+     */
+    void recordFetchResult(TableBucket tableBucket, boolean hasRecords) {
+        BucketFetchState state =
+                bucketStates.computeIfAbsent(tableBucket, k -> new BucketFetchState());
+        if (hasRecords) {
+            if (state.consecutiveEmptyFetches > 0) {
+                LOG.info(
```
fresh-borzoni, line 104: ditto

PR #3008, opened 2026-04-07
[client] Fix corruption after zero-copy lazy parse ByteBuf

fresh-borzoni:
## Summary

After #2948, INDEXED/COMPACTED rows reference pooled Netty ByteBuf memory. When DefaultCompletedFetch.drain() releases the buffer, BinaryString fields in returned ScanRecords become dangling references — causing corruption like in https://github.com/apache/fluss/actions/runs/24007394419/job/70013234004?pr=3004

Fix: when building field getters for log record projection, defensively copy STRING/CHAR fields via BinaryString.copy() for formats that reference pooled buffers. ARROW is excluded since VectorizedColumnBatch already produces independent      string copies.
The copy flag propagates through nested types (ARRAY, MAP, ROW) so structures like ARRAY<STRING> are also safe.

### DIFF
--- a/fluss-common/src/main/java/org/apache/fluss/row/InternalArray.java
+++ b/fluss-common/src/main/java/org/apache/fluss/row/InternalArray.java
@@ -19,8 +19,6 @@
 package org.apache.fluss.row;
 
 import org.apache.fluss.annotation.PublicEvolving;
-import org.apache.fluss.row.columnar.ColumnarRow;
-import org.apache.fluss.row.columnar.VectorizedColumnBatch;
 import org.apache.fluss.types.ArrayType;
 import org.apache.fluss.types.DataType;
 import org.apache.fluss.types.MapType;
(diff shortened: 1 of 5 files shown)


fresh-borzoni:
@wuchong @loserwang1024 PTAL 🙏 

WDYT in general about this approach?

fluss-common/src/main/java/org/apache/fluss/row/InternalArray.java
```diff
@@ -161,23 +159,26 @@ static ElementGetter createElementGetter(DataType fieldType) {
...
-     * BYTES types, because {@link ColumnarRow} already deep copies the bytes, see {@link
-     * VectorizedColumnBatch#getString(int, int)}. This can be removed once we supports object reuse
-     * for Arrow {@link ColumnarRow}, see {@code CompletedFetch#toScanRecord(LogRecord)}.
-     */
-    static ElementGetter createDeepElementGetter(DataType fieldType) {
+    /** Same as InternalRow.createDeepFieldGetter but for array elements. */
+    static ElementGetter createDeepElementGetter(DataType fieldType, boolean copyStrings) {
         final ElementGetter elementGetter;
         switch (fieldType.getTypeRoot()) {
+            case CHAR:
```
loserwang1024, line 166: would you like to add a test about it?

loserwang1024: @fresh-borzoni , CC

fresh-borzoni: @loserwang1024 Looking into this, thank you 👍

fresh-borzoni:
@loserwang1024 Thank you for the review, added test, PTAL 🙏

fresh-borzoni:
@loserwang1024 let me know what we should do with this

fluss-common/src/main/java/org/apache/fluss/row/InternalRow.java
```diff
@@ -239,22 +237,35 @@ static FieldGetter createFieldGetter(DataType fieldType, int fieldPos) {
...
-     * BYTES types, because {@link ColumnarRow} already deep copies the bytes, see {@link
-     * VectorizedColumnBatch#getString(int, int)}. This can be removed once we supports object reuse
-     * for Arrow {@link ColumnarRow}, see {@code CompletedFetch#toScanRecord(LogRecord)}.
+     * <p>For ARROW, pass copyStrings=false since {@link VectorizedColumnBatch} already copies
+     * strings. For INDEXED and COMPACTED, pass copyStrings=true to copy STRING/CHAR out of the
+     * pooled buffer they reference.
      */
-    static FieldGetter createDeepFieldGetter(DataType fieldType, int fieldPos) {
+    static FieldGetter createDeepFieldGetter(
+            DataType fieldType, int fieldPos, boolean copyStrings) {
```
platinumhamburg, line 249: `copyStrings` feels too specific for a generic deep getter. It gives this method two different lifetime behaviors, requires callers to understand the physical row representation, and becomes incomplete as soon as another type such as BLOB can retain backing memory. This looks like an ownership concern leaking into the field-access interface.

fluss-common/src/test/java/org/apache/fluss/row/InternalRowTest.java
```diff
@@ -53,4 +59,58 @@ void testGetDataClass() {
         assertThat(InternalRow.getDataClass(DataTypes.ROW(new DataField("a", DataTypes.INT()))))
                 .isEqualTo(InternalRow.class);
     }
+
+    /** copyStrings=true copies STRING/CHAR (including nested) so values survive buffer release. */
+    @Test
+    void testDeepFieldGetterCopiesStringsAfterBufferRelease() {
```
platinumhamburg, line 65: Thanks for adding coverage. I do not think this test reproduces the original failure yet: against the pre-fix code it does not compile because the three-argument helper is new, and it never exercises the actual response `ByteBuf` lifecycle. Could we add a regression test through `DefaultCompletedFetch` that fetches INDEXED/COMPACTED records, lets `fetchRecords()` drain the buffer, verifies `refCnt == 0`, and then reads top-level and nested strings from the returned records? That would give us a behavioral RED before the fix.

fluss-common/src/main/java/org/apache/fluss/row/InternalRow.java
```diff
@@ -239,22 +237,35 @@ static FieldGetter createFieldGetter(DataType fieldType, int fieldPos) {
...
         final FieldGetter fieldGetter;
         switch (fieldType.getTypeRoot()) {
+            case CHAR:
```
platinumhamburg, line 252: `IndexedRow#getChar()` already creates a new `byte[]` and returns an independent `BinaryString`, so the additional `.copy()` here copies top-level INDEXED CHAR values a second time. Could we avoid this extra allocation?

platinumhamburg (changes requested):
ignore.

loserwang1024:
@fresh-borzoni it seems no longer need it, just close this?

PR #3030, opened 2026-04-08
[server] Implement JBOD Phase 1 Local Multi-Directory Support

### LINKED ISSUE #145: [Feature] Fluss server supports writing to multiple disks.
### Search before asking

### Motivation

Fluss server supports writing to multiple disks.

### Solution

_No response_

### Anything else?

_No response_

### Willingness to contribute

hanliu0830:
**(The sections below can be removed for hotfixes or typos)**
-->

### Purpose

Linked issue: close #145


fluss-server/src/main/java/org/apache/fluss/server/replica/ReplicaManager.java
```diff
@@ -1559,18 +1574,31 @@ private void startHighWatermarkCheckPointThread() {
...
+                    try {
+                        highWatermarkCheckpoints.get(entry.getKey()).write(entry.getValue());
+                    } catch (Exception e) {
+                        throw new LogStorageException(
+                                "Error while writing to high watermark file", e);
+                    }
+                }
             }
+        } catch (LogStorageException e) {
+            throw e;
```
gyang94, line 1601: This exception has been thrown inside the loop, the outer try-catch could be removed.

fluss-server/src/main/java/org/apache/fluss/server/log/remote/RemoteLogManager.java
```diff
@@ -240,7 +260,10 @@ public long lookupOffsetForTimestamp(TableBucket tableBucket, long timestamp) {
         if (segment == null) {
             return -1L;
         } else {
-            return remoteLogIndexCache.lookupOffsetForTimestamp(segment, timestamp);
+            RemoteLogIndexCache remoteLogIndexCache = remoteLogIndexCacheForBucket(tableBucket);
+            return remoteLogIndexCache == null
+                    ? -1L
```
gyang94, line 265: When remoteLogIndexCache is null, will it be better to throw an exception instead of return -1?
The same returning value -1 will mix up the cases between 1. failing to find the offset index. and 2. indexCache is null.

fluss-server/src/main/java/org/apache/fluss/server/storage/LocalDiskManager.java
```diff
@@ -0,0 +1,744 @@
...
+    /**
+     * Resolves a local path back to the configured data directory that owns it.
+     *
+     * <p>Example: if {@code /data-0} is configured and the path is {@code /data-0/db/tablet-1}, the
+     * returned directory is {@code /data-0}.
+     */
+    public synchronized File resolveDataDir(File path) {
+        Path pathToResolve = path.toPath();
+        for (File dataDir : dataDirs) {
+            if (pathToResolve.startsWith(dataDir.toPath())) {
```
gyang94, line 411: should ensure that the pathToResolve should be an absolute path, since we check with `startsWith` function. `pathToResolve.toAbsolutePath().normalize()`

fluss-server/src/main/java/org/apache/fluss/server/storage/LocalDiskManager.java
```diff
@@ -0,0 +1,744 @@
...
+        }
+
+        private int serverIdAsInt(File dataDir) throws IOException {
+            if (serverId == null || serverId.trim().isEmpty()) {
+                throw new IOException(
+                        "Missing server.id in disk.properties under " + dataDir.getAbsolutePath());
+            }
+
+            try {
+                return Integer.parseInt(serverId.trim());
```
gyang94, line 557: many `trim()` calls in one function. In the constructor the serverId already trimmed.  Or just trim() once in one function.

fluss-server/src/main/java/org/apache/fluss/server/storage/LocalDiskManager.java
```diff
@@ -0,0 +1,744 @@
...
+        private final String serverId;
+
+        private DiskProperties(int version, String diskId, String serverId) {
+            this.version = version;
+            this.diskId = diskId == null ? null : diskId.trim();
+            this.serverId = serverId == null ? null : serverId.trim();
+        }
+
+        private int serverIdAsInt(File dataDir) throws IOException {
+            if (serverId == null || serverId.trim().isEmpty()) {
```
gyang94, line 551: serverId is already trimmed in constructor.

fluss-server/src/main/java/org/apache/fluss/server/storage/LocalDiskManager.java
```diff
@@ -0,0 +1,744 @@
...
+                                + dataDir.getAbsolutePath()
+                                + ": "
+                                + serverId,
+                        e);
+            }
+        }
+
+        private String diskId(File dataDir) throws IOException {
+            if (diskId == null || diskId.trim().isEmpty()) {
+                throw new IOException(
```
gyang94, line 570: same as above. diskId also was trimmed in the constructor.

fluss-server/src/main/java/org/apache/fluss/server/replica/ReplicaManager.java
hanliu0830: > This exception has been thrown inside the loop, the outer try-catch could be removed.

Addressed in 6999b9ec. I removed the redundant outer try-catch since the exception is already handled inside the loop.

fluss-server/src/main/java/org/apache/fluss/server/log/remote/RemoteLogManager.java
hanliu0830: > When remoteLogIndexCache is null, will it be better to throw an exception instead of return -1? The same returning value -1 will mix up the cases between 1. failing to find the offset index. and 2. indexCache is null.

Addressed in 6999b9ec. I changed this branch to throw `NotLeaderOrFollowerException` instead of returning `-1`, so we no longer mix up “index lookup miss” with “failed to resolve the index cache”.

fluss-server/src/main/java/org/apache/fluss/server/storage/LocalDiskManager.java
hanliu0830: > should ensure that the pathToResolve should be an absolute path, since we check with `startsWith` function. `pathToResolve.toAbsolutePath().normalize()`

Thanks. Addressed in 6999b9ec.

fluss-server/src/main/java/org/apache/fluss/server/storage/LocalDiskManager.java
hanliu0830: > many `trim()` calls in one function. In the constructor the serverId already trimmed. Or just trim() once in one function.

Thanks. Addressed in 6999b9ec.

fluss-server/src/main/java/org/apache/fluss/server/storage/LocalDiskManager.java
hanliu0830: > serverId is already trimmed in constructor.

Thanks. Addressed in 6999b9ec.

fluss-server/src/main/java/org/apache/fluss/server/storage/LocalDiskManager.java
hanliu0830: > same as above. diskId also was trimmed in the constructor.

Thanks. Addressed in 6999b9ec.

fluss-server/src/main/java/org/apache/fluss/server/log/LogManager.java
```diff
@@ -141,85 +142,122 @@ public void startup() {
...
+                                loadLogs(dataDir, tabletsToLoad);
+                            } catch (Throwable t) {
+                                throw new FlussRuntimeException(
+                                        "Failed to recover logs from " + dataDir.getAbsolutePath(),
+                                        t);
+                            }
+                        };
+                ExecutorService pool = createThreadPoolByDir("log-recovery-data-dir", dataDir);
+                poolsByDataDir.put(dataDir, pool);
+                jobsForDataDir.add(pool.submit(runnable));
```
wuchong: This code creates a duplicate ExecutorService pool, as `loadLogs(dataDir, tabletsToLoad)` also creates one. We should create only one pool per dataDir, submit load jobs to their respective dataDir pools, and wait for all jobs to complete here. Therefore, we need to refactor `loadLogs(dataDir, tabletsToLoad)` and remove the direct dependency on the `runInThreadPool` method.

fluss-server/src/main/java/org/apache/fluss/server/log/LogManager.java
```diff
@@ -141,85 +142,122 @@ public void startup() {
...
+        recoveryPointCheckpoints = new HashMap<>();
+        for (File dataDir : dataDirs) {
+            recoveryPointCheckpoints.put(
+                    dataDir,
+                    new OffsetCheckpointFile(new File(dataDir, RECOVERY_POINT_CHECKPOINT_FILE)));
+        }
     }
 
     /** Recover and load all logs in the given data directories. */
     private void loadLogs() {
```
wuchong: Rename this method to `loadAllLogs` to distinguish with `loadLogs(..)`

fluss-server/src/main/java/org/apache/fluss/server/log/LogManager.java
```diff
@@ -367,49 +406,64 @@ private LogTablet loadLog(
...
+
+        Map<File, ExecutorService> poolsByDataDir = new LinkedHashMap<>();
+        List<Future<?>> jobsForDataDir = new ArrayList<>();
+        for (Map.Entry<File, List<LogTablet>> entry : logsByDataDir.entrySet()) {
+            File dataDir = entry.getKey();
+            List<LogTablet> logs = entry.getValue();
+            Runnable runnable = () -> shutdown(dataDir, logs);
+            ExecutorService pool = createThreadPoolByDir("log-tablet-closing-data-dir", dataDir);
+            poolsByDataDir.put(dataDir, pool);
+            jobsForDataDir.add(pool.submit(runnable));
```
wuchong: ditto. This creates duplicate ExecutorService pools, we should create only one pool for each data dir to reduce resources.

fluss-server/src/main/java/org/apache/fluss/server/log/LogManager.java
```diff
@@ -367,49 +406,64 @@ private LogTablet loadLog(
...
+        }
+
+        try {
+            if (loadLogsCompletedFlag) {
+                LOG.debug("Writing clean shutdown marker.");
+            }
```
wuchong: What's the purpose to do the logging? I think this `loadLogsCompletedFlag` is used to create the clean shutdown mark in `shutdown(File dataDir, List<LogTablet> logs)` and not necessary to be used here.

fluss-server/src/main/java/org/apache/fluss/server/log/LogManager.java
```diff
@@ -522,23 +577,32 @@ public void run() {
     }
 
     @VisibleForTesting
-    void checkpointRecoveryOffsets() {
-        // Assuming TableBucket and LogTablet are actual types used in your application
-        if (recoveryPointCheckpoint != null) {
-            try {
-                Map<TableBucket, Long> recoveryOffsets = new HashMap<>();
-                for (Map.Entry<TableBucket, LogTablet> entry : currentLogs.entrySet()) {
+    void checkpointRecoveryOffsets(File dataDir) {
```
wuchong, line 573: Could you refactor this method to collect the `LogTablet`s that belong to the `dataDir`, and directly call `checkpointRecoveryOffsets(File dataDir, List<LogTablet> logs)`? This helps us to reuse code.

fluss-server/src/main/java/org/apache/fluss/server/storage/LocalDiskManager.java
```diff
@@ -0,0 +1,744 @@
...
+        return dataDirs;
+    }
+
+    /**
+     * Resolves a local path back to the configured data directory that owns it.
+     *
+     * <p>Example: if {@code /data-0} is configured and the path is {@code /data-0/db/tablet-1}, the
+     * returned directory is {@code /data-0}.
+     */
+    public synchronized File resolveDataDir(File path) {
```
wuchong, line 411: This method is only used to derive the dataDir from `LogTablet`. However, this method appears in many hot path, but this method is heavy (synchronized and path comparision). Can we just have a `dataDir` in the `LogTablet` directly? This can make the code more clean and efficient.

fluss-server/src/main/java/org/apache/fluss/server/TabletManagerBase.java
wuchong, line 148: We should be able to remove this method once we remove the `runInThreadPool(Runnable[] runnableJobs, String poolName)`.

fluss-server/src/main/java/org/apache/fluss/server/log/remote/RemoteLogManager.java
```diff
@@ -104,20 +113,28 @@ public RemoteLogManager(
...
+        this.logManager = logManager;
+        this.localDiskManager = localDiskManager;
+        this.remoteLogIndexCachesByDir = new ConcurrentHashMap<>();
+        for (File dataDir : localDiskManager.dataDirs()) {
+            remoteLogIndexCachesByDir.put(
+                    dataDir,
+                    new RemoteLogIndexCache(
+                            (int)
+                                    conf.get(ConfigOptions.REMOTE_LOG_INDEX_FILE_CACHE_SIZE)
+                                            .getBytes(),
```
wuchong: Move the initialization of `cacheSize` before the foreach block. 

```java
int cacheSize = (int) conf.get(ConfigOptions.REMOTE_LOG_INDEX_FILE_CACHE_SIZE).getBytes();
        for (File dataDir : localDiskManager.dataDirs()) {
            remoteLogIndexCachesByDir.put(
                    dataDir, new RemoteLogIndexCache(cacheSize, remoteLogStorage, dataDir));
        }
```

wuchong (commented):
Thanks @hanliu0830 , I left some comments.

fluss-server/src/main/java/org/apache/fluss/server/storage/LocalDiskManager.java
```diff
@@ -0,0 +1,744 @@
...
+                        dataDir,
+                        onlineDataDirs,
+                        offlineDataDirs,
+                        "Failed to create or validate data directory " + dataDir,
+                        e);
+            }
+        }
+    }
+
+    private void lockDataDirs(Set<File> onlineDataDirs, Set<File> offlineDataDirs) {
```
wuchong, line 305: In a multi-directory deployment where one configured path is already locked by another Fluss process/
    thread, tryLock() returns false here and immediately throws IllegalConfigurationException. Because that
    exception never goes through markDataDirOffline(...), startup aborts even if the other configured data
    directories are healthy, which breaks the new JBOD behavior of dropping bad directories and continuing
    with the rest. Should we treat such case as an offline directory?

wuchong:
### Code review

Found 1 issue:

1. `throw e.getCause()` will throw `NullPointerException` when the caught exception is an `InterruptedException`, since `InterruptedException.getCause()` returns `null` by default. The `InterruptedException` and `ExecutionException` cases should be handled separately — for `InterruptedException`, restore the interrupt flag and rethrow; for `ExecutionException`, unwrap with a null check on `getCause()`.

https://github.com/apache/fluss/blob/9bc453cc9a43706cf81a2d82d461866292fe4cd7/fluss-server/src/main/java/org/apache/fluss/server/log/LogManager.java#L180-L185

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

wuchong (commented):
Besides, The JBOD multi-directory feature has comprehensive unit tests (`LocalDiskManagerTest` with 14 cases, `ReplicaManagerTest` with 3 JBOD-tagged cases, plus adapted `LogManagerTest`/`KvManagerTest`), but it lacks integration tests (ITCase) that validate the end-to-end behavior in a real multi-TabletServer cluster. I'd suggest adding a `JbodMultiDirITCase` under `fluss-server/.../server/replica/` (alongside existing `KvSnapshotITCase`, `ReplicaFetcherITCase`, etc.) using `FlussClusterExtension` with `data.dirs` configured to multiple directories. Key scenarios to cover: (1) creating a table and verifying buckets are distributed across directories, (2) writing and reading data correctly under multi-dir setup, (3) PK table Log/KV co-location across a real cluster. This would require first extending `FlussClusterExtension` to support per-TabletServer `data.dirs` configuration if it doesn't already.

fluss-server/src/main/java/org/apache/fluss/server/log/LogManager.java
hanliu0830: > This code creates a duplicate ExecutorService pool, as `loadLogs(dataDir, tabletsToLoad)` also creates one. We should create only one pool per dataDir, submit load jobs to their respective dataDir pools, and wait for all jobs to complete here. Therefore, we need to refactor `loadLogs(dataDir, tabletsToLoad)` and remove the direct dependency on the `runInThreadPool` method.

Addressed in 7812df7b. I removed the nested per-dir wrapper pool and kept a single real execution pool for each data directory in the recovery path.

fluss-server/src/main/java/org/apache/fluss/server/log/LogManager.java
hanliu0830: > Rename this method to `loadAllLogs` to distinguish with `loadLogs(..)`

Addressed in 7812df7b. I make the responsibility split clearer.

fluss-server/src/main/java/org/apache/fluss/server/log/LogManager.java
hanliu0830: > ditto. This creates duplicate ExecutorService pools, we should create only one pool for each data dir to reduce resources.

Addressed in 7812df7b. The shutdown path had the same nested-pool structure as recovery

fluss-server/src/main/java/org/apache/fluss/server/log/LogManager.java
hanliu0830: > What's the purpose to do the logging? I think this `loadLogsCompletedFlag` is used to create the clean shutdown mark in `shutdown(File dataDir, List<LogTablet> logs)` and not necessary to be used here.

Addressed in e79d0e07. The outer debug log was inherited from the old single-dir flow and became redundant after the per-dir shutdown refactor, so I removed it and kept only the per-dir log at the actual marker write site.

fluss-server/src/main/java/org/apache/fluss/server/TabletManagerBase.java
hanliu0830: > We should be able to remove this method once we remove the `runInThreadPool(Runnable[] runnableJobs, String poolName)`.

Addressed in 7812df7b. Yes, runInThreadPool(...) was no longer needed and I removed it together with the refactor.

hanliu0830:
> ### Code review
> Found 1 issue:
> 
> 1. `throw e.getCause()` will throw `NullPointerException` when the caught exception is an `InterruptedException`, since `InterruptedException.getCause()` returns `null` by default. The `InterruptedException` and `ExecutionException` cases should be handled separately — for `InterruptedException`, restore the interrupt flag and rethrow; for `ExecutionException`, unwrap with a null check on `getCause()`.
> 
> https://github.com/apache/fluss/blob/9bc453cc9a43706cf81a2d82d461866292fe4cd7/fluss-server/src/main/java/org/apache/fluss/server/log/LogManager.java#L180-L185
> 
> 🤖 Generated with [Claude Code](https://claude.ai/code)
> 
> - If this code review was useful, please react with 👍. Otherwise, react with 👎.

Addressed in dc334c31.Yes, I split the handling of `InterruptedException` and `ExecutionException`

fluss-server/src/main/java/org/apache/fluss/server/storage/LocalDiskManager.java
hanliu0830: > This method is only used to derive the dataDir from `LogTablet`. However, this method appears in many hot path, but this method is heavy (synchronized and path comparision). Can we just have a `dataDir` in the `LogTablet` directly? This can make the code more clean and efficient.

Addressed in ae753bed. The current `resolveDataDir(path)` was mostly recovering information that was already known when the `LogTablet` was created. I persisted the owning `dataDir` on `LogTablet`, switched the hot paths to use `logTablet.getDataDir()`, and removed the duplicate `dataDir` state from `Replica` so local directory ownership now has a single true source.

fluss-server/src/main/java/org/apache/fluss/server/log/remote/RemoteLogManager.java
hanliu0830: > Move the initialization of `cacheSize` before the foreach block.
> 
> ```java
> int cacheSize = (int) conf.get(ConfigOptions.REMOTE_LOG_INDEX_FILE_CACHE_SIZE).getBytes();
>         for (File dataDir : localDiskManager.dataDirs()) {
>             remoteLogIndexCachesByDir.put(
>                     dataDir, new RemoteLogIndexCache(cacheSize, remoteLogStorage, dataDir));
>         }
> ```

Addressed in 6dac12a1. I moved the `cacheSize` initialization before the foreach block.

fluss-server/src/main/java/org/apache/fluss/server/storage/LocalDiskManager.java
hanliu0830: > In a multi-directory deployment where one configured path is already locked by another Fluss process/ thread, tryLock() returns false here and immediately throws IllegalConfigurationException. Because that exception never goes through markDataDirOffline(...), startup aborts even if the other configured data directories are healthy, which breaks the new JBOD behavior of dropping bad directories and continuing with the rest. Should we treat such case as an offline directory?

Addressed in 616d917e. After comparing Kafka's handling here, I kept `tryLock() == false` as a fail-fast configuration/deployment error rather than treating it as an offline disk. Kafka only marks log dirs offline for actual I/O failures, while lock conflicts still fail fast, and I added code comments to make that boundary explicit here as well.

wuchong (approved):
I pushed a commit to refactor checkpointRecoveryOffsets(File dataDir) to invoke checkpointRecoveryOffsets(File dataDir, List<LogTablet> logs) which is minor improvements. Others looks good to me. 

I will merge this once the CI is passed. 

The integration IT case `JbodMultiDirITCase` can be added in a follow up PR.

PR #3047, opened 2026-04-09
[server] Manage users for sasl/plain authentication via cluster properties.

loserwang1024:
### Purpose

Linked issue: close #3046

### DIFF
--- a/fluss-common/src/main/java/org/apache/fluss/config/ConfigOptions.java
+++ b/fluss-common/src/main/java/org/apache/fluss/config/ConfigOptions.java
@@ -287,6 +287,16 @@ public class ConfigOptions {
                                     + "and each super user should be specified in the format `principal_type:principal_name`, e.g., `User:admin;User:bob`. "
                                     + "This configuration is critical for defining administrative privileges in the system.");
 
+    public static final ConfigOption<Boolean> SECURITY_ACL_PRINCIPAL_IGNORE_CASE =
+            key("security.acl.principal.ignore-case")
+                    .booleanType()
+                    .defaultValue(false)
+                    .withDescription(
+                            "Whether to perform case-insensitive matching on principal name and type "
+                                    + "during ACL authorization checks. When set to true, principals "
+                                    + "such as 'User:Admin' and 'user:admin' will be treated as the same principal. "
+                                    + "Default is false for strict case-sensitive matching.");
+
     public static final ConfigOption<Integer> MAX_BUCKET_NUM =
             key("max.bucket.num")
                     .intType()
--- a/fluss-common/src/main/java/org/apache/fluss/config/ConfigOptions.java
+++ b/fluss-common/src/main/java/org/apache/fluss/config/ConfigOptions.java
@@ -529,8 +529,8 @@ public class ConfigOptions {
                                     + "Each listener can be associated with a specific authentication protocol. "
                                     + "Listeners not included in the map will use PLAINTEXT by default, which does not require authentication.");
 
-    public static final ConfigOption<List<String>> SERVER_SASL_USERS =
-            key("security.sasl.plain.users")
+    public static final ConfigOption<List<String>> SERVER_SASL_CREDENTIALS =
+            key("security.sasl.plain.credentials")
                     .stringType()
                     .asList()
                     .noDefaultValue()
--- a/fluss-common/src/main/java/org/apache/fluss/config/ConfigOptions.java
+++ b/fluss-common/src/main/java/org/apache/fluss/config/ConfigOptions.java
@@ -529,13 +529,12 @@ public class ConfigOptions {
                                     + "Each listener can be associated with a specific authentication protocol. "
                                     + "Listeners not included in the map will use PLAINTEXT by default, which does not require authentication.");
 
-    public static final ConfigOption<List<String>> SERVER_SASL_CREDENTIALS =
+    public static final ConfigOption<Map<String, String>> SERVER_SASL_CREDENTIALS =
             key("security.sasl.plain.credentials")
-                    .stringType()
-                    .asList()
+                    .mapType()
                     .noDefaultValue()
                     .withDescription(
-                            "List of user credentials for SASL/PLAIN authentication in 'username:password' format. "
+                            "Map of user credentials for SASL/PLAIN authentication in 'username:password' format. "
                                     + "For example: 'admin:admin-secret,bob:bob-secret'. "
                                     + "This is syntactic sugar that auto-generates the JAAS config string.");
 
--- a/fluss-common/src/main/java/org/apache/fluss/config/ConfigOptions.java
+++ b/fluss-common/src/main/java/org/apache/fluss/config/ConfigOptions.java
@@ -519,6 +519,16 @@ public class ConfigOptions {
                                     + "Each listener can be associated with a specific authentication protocol. "
                                     + "Listeners not included in the map will use PLAINTEXT by default, which does not require authentication.");
 
+    public static final ConfigOption<List<String>> SERVER_SASL_USERS =
+            key("security.sasl.plain.users")
+                    .stringType()
+                    .asList()
+                    .noDefaultValue()
+                    .withDescription(
+                            "List of user credentials for SASL/PLAIN authentication in 'username:password' format. "
+                                    + "For example: 'admin:admin-secret,bob:bob-secret'. "
+                                    + "This is syntactic sugar that auto-generates the JAAS config string.");
+
     public static final ConfigOption<Integer> TABLET_SERVER_ID =
             key("tablet-server.id")
                     .intType()
--- a/fluss-common/src/main/java/org/apache/fluss/config/ConfigOptions.java
+++ b/fluss-common/src/main/java/org/apache/fluss/config/ConfigOptions.java
@@ -538,6 +538,17 @@ public class ConfigOptions {
                                     + "For example: 'admin:admin-secret,bob:bob-secret'. "
                                     + "This is syntactic sugar that auto-generates the JAAS config string.");
 
+    public static final ConfigOption<String> SERVER_SASL_PLAIN_JAAS_CONFIG =
+            key("security.sasl.plain.jaas.config")
+                    .stringType()
+                    .noDefaultValue()
+                    .withDescription(
+                            "JAAS configuration string for server-side SASL/PLAIN authentication. "
+                                    + "The value should use PlainLoginModule and define users with "
+                                    + "'user_<username>=\"<password>\"' options. This option is generated "
+                                    + "from 'security.sasl.plain.credentials' when that credential map is set, "
+                                    + "and can also be configured directly for compatibility.");
+
     public static final ConfigOption<Integer> TABLET_SERVER_ID =
             key("tablet-server.id")
                     .intType()
--- a/fluss-common/src/main/java/org/apache/fluss/security/auth/AuthenticationFactory.java
+++ b/fluss-common/src/main/java/org/apache/fluss/security/auth/AuthenticationFactory.java
@@ -90,7 +90,8 @@ public class AuthenticationFactory {
      *     authenticators.
      */
     public static Map<String, Supplier<ServerAuthenticator>> loadServerAuthenticatorSuppliers(
-            Configuration configuration) {
+            Supplier<Configuration> configurationSupplier) {
+        Configuration configuration = configurationSupplier.get();
         PluginManager pluginManager = PluginUtils.createPluginManagerFromRootFolder(configuration);
         Map<String, Supplier<ServerAuthenticator>> serverAuthenticators = new HashMap<>();
         Map<String, String> protocolMap =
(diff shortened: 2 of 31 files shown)


fluss-common/src/main/java/org/apache/fluss/config/ConfigOptions.java
```diff
@@ -519,6 +529,16 @@ public class ConfigOptions {
                                     + "Each listener can be associated with a specific authentication protocol. "
                                     + "Listeners not included in the map will use PLAINTEXT by default, which does not require authentication.");
 
+    public static final ConfigOption<List<String>> SERVER_SASL_USERS =
```
wuchong: Since `org.apache.fluss.server.DynamicConfigManager#describeConfigs()` returns all cluster config values to users with `CLUSTER DESCRIBE`, we should mask sensitive values before creating `ConfigEntrys` there. In particular, `security.sasl.plain.users` should hide the password part, e.g. `root:******`, and `security.sasl.plain.jaas.config` should also be redacted instead of returned verbatim. This keeps the protection centralized for all describe-cluster-config callers.

fluss-rpc/src/main/java/org/apache/fluss/rpc/netty/server/FlussProtocolPlugin.java
```diff
@@ -82,8 +91,73 @@ public RequestHandler<?> createRequestHandler(RpcGatewayService service) {
...
+        }
+        StringBuilder sb =
+                new StringBuilder(
+                        "org.apache.fluss.security.auth.sasl.plain.PlainLoginModule required");
+        for (String entry : users) {
+            int colonIdx = entry.indexOf(':');
+            checkArgument(colonIdx > 0, "Invalid user entry format: '%s'", entry);
+            String username = entry.substring(0, colonIdx);
+            String password = entry.substring(colonIdx + 1);
+            sb.append(String.format(" user_%s=\"%s\"", username, password));
```
wuchong: The generated JAAS string inserts username and password directly inside a quoted option. A password containing JAAS syntax characters such as a double quote is accepted by `validate()`, but produces an invalid or wrong JAAS config for new connections. Please escape these values for `JaasConfig`, or reject unsupported characters during validation.

fluss-common/src/main/java/org/apache/fluss/security/auth/AuthenticationFactory.java
```diff
@@ -90,7 +90,8 @@ public static Supplier<ClientAuthenticator> loadClientAuthenticatorSupplier(
      *     authenticators.
      */
     public static Map<String, Supplier<ServerAuthenticator>> loadServerAuthenticatorSuppliers(
-            Configuration configuration) {
+            Supplier<Configuration> configurationSupplier) {
```
wuchong: I think we don't need to change the `Configuration` to a `Supplier`. The `Configuration` is thread-safe, and we can dynamically update the `Configuration` for the security key, then the next `ServerAuthenticator` will use the new security value.

fluss-server/src/main/java/org/apache/fluss/server/coordinator/CoordinatorServer.java
```diff
@@ -295,6 +296,8 @@ protected void initCoordinatorStandby() throws Exception {
                             serverMetricGroup,
                             RequestsMetrics.createCoordinatorServerRequestMetrics(
                                     serverMetricGroup));
+            // Register FlussProtocolPlugin for dynamic SASL config updates
+            dynamicConfigManager.register(((NettyServer) rpcServer).getFlussProtocolPlugin());
             rpcServer.start();
```
wuchong, line 297: Could we avoid coupling `CoordinatorServer` to the `NettyServer` implementation here? Exposing the RPC server's reconfigurable components through `RpcServer` (for example, `List<ServerReconfigurable>`) would let both CoordinatorServer and TabletServer register them without hard-casting, and would also cover future protocol plugins that implement `ServerReconfigurable`.


Add `getServerReconfigurables` in `RpcServer`
```java
/** Returns server components that should receive dynamic config updates. */
default List<ServerReconfigurable> getServerReconfigurables() {
    return Collections.emptyList();
}
```

fluss-rpc/src/main/java/org/apache/fluss/rpc/netty/server/FlussProtocolPlugin.java
```diff
@@ -21,22 +21,30 @@
...
+import static org.apache.fluss.utils.Preconditions.checkArgument;
 
 /** Build-in protocol plugin for Fluss. */
-public class FlussProtocolPlugin implements NetworkProtocolPlugin {
+public class FlussProtocolPlugin implements NetworkProtocolPlugin, ServerReconfigurable {
     private final ApiManager apiManager;
     private final List<String> listeners;
     private final RequestsMetrics requestsMetrics;
-    private Configuration conf;
+    private volatile Configuration conf;
```
wuchong: We don't need to use `volatile` if we update `conf` directly.

fluss-rpc/src/main/java/org/apache/fluss/rpc/netty/server/FlussProtocolPlugin.java
```diff
@@ -82,8 +91,73 @@ public RequestHandler<?> createRequestHandler(RpcGatewayService service) {
...
+
+    /**
+     * Enriches the given configuration with a generated JAAS config string derived from the
+     * security.sasl.plain.users list (format: 'username:password'). If the list is not present, the
+     * original configuration is returned unchanged.
+     */
+    private static Configuration enrichWithJaasConfig(Configuration config) {
+        List<String> users = config.get(ConfigOptions.SERVER_SASL_USERS);
+        if (users == null) {
+            return config;
```
wuchong: If `users` are empty, it maybe a change from non-empty to empty, so we should also materialize `security.sasl.plain.jaas.config` again here.

fluss-rpc/src/main/java/org/apache/fluss/rpc/netty/server/FlussProtocolPlugin.java
```diff
@@ -82,8 +91,73 @@ public RequestHandler<?> createRequestHandler(RpcGatewayService service) {
...
+        for (String entry : users) {
+            int colonIdx = entry.indexOf(':');
+            checkArgument(colonIdx > 0, "Invalid user entry format: '%s'", entry);
+            String username = entry.substring(0, colonIdx);
+            String password = entry.substring(colonIdx + 1);
+            sb.append(String.format(" user_%s=\"%s\"", username, password));
+        }
+        sb.append(";");
+        Configuration enriched = new Configuration(config);
+        enriched.setString("security.sasl.plain.jaas.config", sb.toString());
```
wuchong: We can update the `conf` member here directly, rather than create a new Configuration.

fluss-rpc/src/main/java/org/apache/fluss/rpc/netty/server/FlussProtocolPlugin.java
```diff
@@ -82,8 +91,73 @@ public RequestHandler<?> createRequestHandler(RpcGatewayService service) {
...
     @VisibleForTesting
     ApiManager getApiManager() {
         return apiManager;
     }
+
+    @VisibleForTesting
+    @Nullable
+    Configuration getConf() {
+        return conf;
+    }
```
wuchong: Not used. Can be removed.

fluss-rpc/src/main/java/org/apache/fluss/rpc/netty/server/FlussProtocolPlugin.java
```diff
@@ -82,8 +91,73 @@ public RequestHandler<?> createRequestHandler(RpcGatewayService service) {
...
+     * original configuration is returned unchanged.
+     */
+    private static Configuration enrichWithJaasConfig(Configuration config) {
+        List<String> users = config.get(ConfigOptions.SERVER_SASL_USERS);
+        if (users == null) {
+            return config;
+        }
+        StringBuilder sb =
+                new StringBuilder(
+                        "org.apache.fluss.security.auth.sasl.plain.PlainLoginModule required");
```
wuchong, line 220: When materializing `SERVER_SASL_USERS` into `security.sasl.plain.jaas.config`, we should also respsect the existing value of `security.sasl.plain.jaas.config`, and override the new userpass in `SERVER_SASL_USERS`.

fluss-server/src/main/java/org/apache/fluss/server/DynamicServerConfig.java
```diff
@@ -108,7 +111,15 @@ class DynamicServerConfig {
...
 
     void register(ServerReconfigurable serverReconfigurable) {
-        serverReconfigures.put(serverReconfigurable.getClass(), serverReconfigurable);
+        inWriteLock(
+                lock,
+                () -> {
+                    if (initialized) {
+                        Configuration configToApply = new Configuration(currentConfig);
+                        applyToServerReconfigurable(serverReconfigurable, configToApply);
+                    }
```
wuchong: I suggest we always call `applyToServerReconfigurable(serverReconfigurable, currentConfig)` before adding to `serverReconfigures`. This approach eliminates the need for `initialized` and `initializeDynamicConfig()`. Additionally, I recommend that callers invoke `startup` before `register` (for both CoordinatorServer and TabletServer). This ensures a safer and more flexible workflow by guaranteeing that the latest dynamic configuration is retrieved before registering ServerReconfigurable.

fluss-flink/fluss-flink-common/src/test/java/org/apache/fluss/flink/procedure/FlinkProcedureITCase.java
```diff
@@ -783,6 +785,126 @@ void testListRebalanceProgress() throws Exception {
...
+        // Verify "bob" can no longer authenticate
+        assertThatThrownBy(() -> tEnv.executeSql(createCatalogDDL).await())
+                .hasMessageContaining("Invalid username or password");
+        // Cleanup: remove bob's ACL
+        tEnv.executeSql(
+                        String.format(
+                                "Call %s.sys.drop_acl('CLUSTER', 'ALLOW', 'User:bob', 'DESCRIBE', '*')",
+                                CATALOG_NAME))
+                .await();
+        // Try to append a user with the same username as the existing "root" user
```
wuchong: Could you add more negative test cases, such as userpass strings containing multiple colons `:`, no colon, or special characters like commas `,` and double quotes `"`? 

Additionally, please test the behavior when attempting to remove a non-existent user. Should we return an exception or error to the user in this scenario?

fluss-common/src/main/java/org/apache/fluss/config/ConfigOptions.java
```diff
@@ -519,6 +529,16 @@ public class ConfigOptions {
...
+            key("security.sasl.plain.users")
+                    .stringType()
+                    .asList()
```
wuchong: Since each entry contains both the username and password, could we use `security.sasl.plain.credentials` for this key? It reads more clearly than `security.sasl.plain.users` and signals that the value contains sensitive credential material, not just user identifiers.

fluss-rpc/src/main/java/org/apache/fluss/rpc/netty/server/FlussProtocolPlugin.java
```diff
@@ -82,8 +91,73 @@ public RequestHandler<?> createRequestHandler(RpcGatewayService service) {
         return new FlussRequestHandler(service);
     }
 
+    // --- ServerReconfigurable ---
+
+    @Override
+    public void validate(Configuration newConfig) throws ConfigException {
```
wuchong, line 123: It would be good to validate the username/password syntax here before generating JAAS config. The username should probably be restricted to a simple safe pattern such as `[A-Za-z0-9._-]+`, since it is used as part of the JAAS option key (`user_${username}`). For passwords, I would avoid limiting them to only letters and digits, but we should at least reject characters that break our `username:password` list format or generated JAAS string, such as commas, quotes, semicolons, backslashes, and control characters. This keeps the dynamic config behavior predictable without silently producing invalid JAAS config.

wuchong (commented):
·org.apache.fluss.server.DynamicConfigManager#describeConfigs#describeConfigs`

loserwang1024:
@wuchong ， I have modified this PR based on your advice. Only one is left for another PR later.

One concern I have is that `security.sasl.plain.credentials` is a bit different from normal sensitive config options. Each list entry stores `username:password` as a single string value, so the config framework cannot naturally distinguish the non-sensitive username part from the sensitive password part.

If we only add a generic sensitive flag, the safest behavior would be to mask the whole config value in `describeClusterConfigs` / `get_cluster_configs`. That is reusable and avoids leaking passwords, but it also means usernames are hidden together with passwords.

If we want to show usernames while hiding passwords, we would need special parsing logic for this specific config key, e.g. rendering `alice:******,bob:******`. I’m not sure that belongs in the generic config display path, because it would make the config framework aware of SASL/PLAIN-specific value syntax and edge cases.

So I’d like to discuss which direction we prefer:
1. Treat the whole `security.sasl.plain.credentials` value as sensitive and mask it entirely.
2. Add special handling for this key to expose usernames but mask passwords.
3. Keep config display generic, and if listing SASL/PLAIN users is needed, add a dedicated user-management/list-users API later.

wuchong:
@loserwang1024 I pushed several commits to enhance the code, please take a look.

wuchong (approved):

PR #3050, opened 2026-04-09
[server][netty] Introduce netty.server.max-request-size to limit the request size

### LINKED ISSUE #3049: Introduce `netty.server.max-request-size` to limit the request size
### Search before asking

### Description

Both Kafka and Pulsar enforce limits on the server side:
[1] In Pulsar, the maxMessageSize parameter can be configured on the server, with a default value of 5 MB (see: https://pulsar.apache.org/docs/next/concepts-messaging/).
[2] In Kafka, the socket.request.max.bytes configuration (often associated with socket.send.buffer.bytes) has a default value of 100 MB (see: https://kafka.apache.org/42/configuration/broker-configs/#brokerconfigs_socket.request.max.bytes).

Similarly, Fluss should introduce a configurable parameter named netty.server.max-request-size, with a default value of 100 MB.

（In our production environment, we have write batch.size values reaching several tens of megabytes; therefore, we require this limit to be set higher.）

### Willingness to contribute

loserwang1024:
### Purpose

Linked issue: close #3049

### DIFF
--- a/fluss-common/src/main/java/org/apache/fluss/config/ConfigOptions.java
+++ b/fluss-common/src/main/java/org/apache/fluss/config/ConfigOptions.java
@@ -916,6 +916,16 @@ public class ConfigOptions {
                     .withDescription(
                             "The number of queued requests allowed for worker threads, before blocking the I/O threads.");
 
+    public static final ConfigOption<MemorySize> NETTY_SERVER_MAX_REQUEST_SIZE =
+            key("netty.server.max-request-size")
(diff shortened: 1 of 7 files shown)


wuchong (commented):
Thanks @loserwang1024 , 

1. Please update the configuration documentation as well. 
2. Please add unit test or IT case to verify invalid request (mocked request with a very large frame length) can be rejected.

wuchong (approved):

PR #3054, opened 2026-04-10
[common] Optimize getNullCounts() to return int[] instead of Long[]

platinumhamburg:
Since null counts are stored as 4-byte integers in the batch statistics binary format, int[] is sufficient and avoids boxing overhead (8 bytes per Long vs 4 bytes per int). Use -1 as the sentinel for "not available" instead of null. This reduces memory usage for cachedNullCounts and statsNullCounts, especially for wide tables with many fields.

Closes #3021

### Purpose

Linked issue: close #3021


wuchong (approved):
LGTM. I pushed a commit to fix the compile problem.

platinumhamburg:
> LGTM. I pushed a commit to fix the compile problem.

Thanks, LGTM.

PR #3057, opened 2026-04-10
[spark] Fix column projection on log/upsert read path

fresh-borzoni:
FlussPartitionReader.rowType uses the full table schema, but scanners return projected rows with shifted ordinals. Type-dependent accessors in FlussAsSparkRow dispatch via rowType.getTypeAt(ordinal), so projected queries on TIMESTAMP/ARRAY/STRUCT columns get a ClassCastException. Override convertToSparkRow in both subclasses to use the projected row type, and stop double-projecting in SortMergeReader.

closes https://github.com/apache/fluss/issues/3029

### DIFF
--- a/fluss-spark/fluss-spark-common/src/main/scala/org/apache/fluss/spark/read/FlussAppendPartitionReader.scala
+++ b/fluss-spark/fluss-spark-common/src/main/scala/org/apache/fluss/spark/read/FlussAppendPartitionReader.scala
@@ -20,6 +20,10 @@ package org.apache.fluss.spark.read
 import org.apache.fluss.client.table.scanner.ScanRecord
 import org.apache.fluss.config.Configuration
 import org.apache.fluss.metadata.{TableBucket, TablePath}
+import org.apache.fluss.row.{InternalRow => FlussInternalRow}
+import org.apache.fluss.spark.row.DataConverter
+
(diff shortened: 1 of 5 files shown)


fresh-borzoni:
@luoyuxia @YannByron @Yohahaha  PTAL 🙏

fluss-spark/fluss-spark-common/src/main/scala/org/apache/fluss/spark/read/FlussAppendPartitionReader.scala
```diff
@@ -29,6 +33,12 @@ class FlussAppendPartitionReader(
     flussConfig: Configuration)
   extends FlussPartitionReader(tablePath, flussConfig) {
 
+  private lazy val projectedRowType = rowType.project(projection)
+
+  override protected def convertToSparkRow(flussRow: FlussInternalRow): InternalRow = {
+    DataConverter.toSparkInternalRow(flussRow, projectedRowType)
+  }
+
```
Yohahaha: could we move these common func to FlussPartitionReader?

Yohahaha (approved):
overall LGTM! left one comments, thank you!

fresh-borzoni:
@Yohahaha Ty for the review, addressed comment, PTAL 🙏

luoyuxia (approved):
+1

PR #3059, opened 2026-04-13
[metrics] Fix pendingRecords metric to return actual row count for primary key tables when lake tiering hasn't started

beryllw:
### Purpose

Linked issue: close #xxx
Fix LAKE_PENDING_RECORDS metric for primary key tables when lake tiering hasn't started yet (getLakeLogEndOffset() < 0). Previously, it returned log size (highWatermark - startOffset) which doesn't reflect actual pending records for PK tables. Now returns materialized row count for PK tables.

### Brief change log

Modified registerLakeTieringMetrics() in Replica.java to use getRowCount() for primary key tables when getLakeLogEndOffset() < 0, while keeping log size calculation for non-PK tables

### DIFF
--- a/fluss-server/src/main/java/org/apache/fluss/server/replica/Replica.java
+++ b/fluss-server/src/main/java/org/apache/fluss/server/replica/Replica.java
@@ -557,9 +557,9 @@ public final class Replica {
                         try {
                             return getRowCount();
                         } catch (InvalidTableException e) {
-                            // WAL mode or v0.9 old table: row count disabled,
-                            // fall back to log-level pending count
-                            return getLogHighWatermark() - getLogStartOffset();
+                            // WAL mode or v0.9 old table with no completed tiering:
+                            // row count disabled, return -1 to indicate unavailable
+                            return -1L;
                         }
                     }
                     return getLogHighWatermark() - lakeLogEndOffset;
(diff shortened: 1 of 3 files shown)


fluss-server/src/main/java/org/apache/fluss/server/replica/Replica.java
```diff
@@ -549,11 +549,13 @@ private void onBecomeNewLeader() {
...
-                MetricNames.LOG_LAKE_PENDING_RECORDS,
-                () ->
-                        getLakeLogEndOffset() < 0L
-                                ? getLogHighWatermark() - getLogStartOffset()
-                                : getLogHighWatermark() - getLakeLogEndOffset());
+                MetricNames.LAKE_PENDING_RECORDS,
+                () -> {
+                    long lakeLogEndOffset = getLakeLogEndOffset();
+                    return lakeLogEndOffset < 0L
+                            ? getRowCount()
```
luoyuxia: what if it's the old table which is created before v0.9?

fluss-server/src/main/java/org/apache/fluss/server/replica/Replica.java
```diff
@@ -549,11 +550,20 @@ private void onBecomeNewLeader() {
...
+                MetricNames.LAKE_PENDING_RECORDS,
+                () -> {
+                    long lakeLogEndOffset = getLakeLogEndOffset();
+                    if (lakeLogEndOffset < 0L) {
+                        try {
+                            return getRowCount();
+                        } catch (InvalidTableException e) {
+                            // WAL mode or v0.9 old table: row count disabled,
+                            // fall back to log-level pending count
+                            return getLogHighWatermark() - getLogStartOffset();
```
luoyuxia: should return -1 ?, otherwise you will never know whether it never tier.
I'm not sure whether auto will require the information about  "whether it never tier".

luoyuxia (approved):
+1

PR #3067, opened 2026-04-13
[filesystem] Support default AWS credential chain for S3 delegation

### LINKED ISSUE #3066: [filesystem] S3 credential delegation requires static keys, blocking IRSA/IAM role usage
### Search before asking

### Motivation

Currently `S3DelegationTokenProvider` hardcodes `AWSStaticCredentialsProvider` when building the STS client, requiring `s3.access-key` / `s3.secret-key` in config. This prevents using the AWS default credential chain (IRSA, instance profiles, environment variables).                                                                                                                                                                                                                            
   
Additionally, `S3FileSystemPlugin` uses the presence of `access.key` to distinguish server from client. Without static keys, the server is misdetected as a client and fails with `NoAwsCredentialsException`.                                     
This blocks running Fluss on EKS with IRSA, which is the standard way to manage credentials on AWS.

### Solution

Fix: fall back to `DefaultAWSCredentialsProviderChain` when no static keys are provided, and require `fs.s3a.assumed.role.arn` in that case (since `GetSessionToken` does not work with temporary credentials). 
Use `roleArn` presence as an additional server-mode indicator in the plugin.  

### Anything else?

Related: #2662, #1245, #2989

### Willingness to contribute

fresh-borzoni:
## Summary

resolves https://github.com/apache/fluss/issues/3066

 Currently the STS client in `S3DelegationTokenProvider` is hardcoded to use `AWSStaticCredentialsProvider`, so credential delegation only works when static access keys are in the config. This prevents using IRSA, instance profiles, or any other credential method from the default AWS chain. Additionally, `S3FileSystemPlugin` uses the presence of `access.key` to distinguish server from client, so a server without static keys is misdetected as a client and crashes with `NoAwsCredentialsException`.
 
 This PR makes the STS client fall back to `DefaultAWSCredentialsProviderChain` when no static keys are configured, and requires `roleArn` in that case since `GetSessionToken` does not work with temporary credentials. The plugin now also treats `roleArn` presence as a server-mode indicator.
                                                                                                                                                                                                                                                     
This is a targeted fix that does not change the delegation protocol or add new config keys. 
The full credential delegation redesign (#2662) can build on top. See also #1245 which attempted a broader approach.                                   
   
Tested E2E with LocalStack: Fluss server with env-var credentials (no static keys in config) + AssumeRole delegation, Flink client reading tiered log segments via delegated credentials.

### DIFF
--- a/fluss-filesystems/fluss-fs-s3/src/main/java/org/apache/fluss/fs/s3/S3FileSystemPlugin.java
+++ b/fluss-filesystems/fluss-fs-s3/src/main/java/org/apache/fluss/fs/s3/S3FileSystemPlugin.java
@@ -45,6 +45,8 @@ public class S3FileSystemPlugin implements FileSystemPlugin {
 
     private static final String ACCESS_KEY_ID = "fs.s3a.access.key";
 
+    private static final String ROLE_ARN_KEY = "fs.s3a.assumed.role.arn";
+
     private static final String[][] MIRRORED_CONFIG_KEYS = {
         {"fs.s3a.access-key", "fs.s3a.access.key"},
         {"fs.s3a.secret-key", "fs.s3a.secret.key"},
@@ -122,7 +124,15 @@ public class S3FileSystemPlugin implements FileSystemPlugin {
     }
 
     private void setCredentialProvider(org.apache.hadoop.conf.Configuration hadoopConfig) {
-        if (hadoopConfig.get(ACCESS_KEY_ID) == null) {
+        boolean hasStaticKeys = hadoopConfig.get(ACCESS_KEY_ID) != null;
+        boolean hasRoleArn = hadoopConfig.get(ROLE_ARN_KEY) != null;
+
+        if (hasStaticKeys || hasRoleArn) {
+            LOG.info(
+                    hasStaticKeys
+                            ? "Using provided static credentials."
+                            : "Using default AWS credential chain with AssumeRole.");
+        } else {
             if (Objects.equals(getScheme(), "s3")) {
                 S3DelegationTokenReceiver.updateHadoopConfig(hadoopConfig);
             } else if (Objects.equals(getScheme(), "s3a")) {
@@ -131,11 +141,8 @@ public class S3FileSystemPlugin implements FileSystemPlugin {
                 throw new IllegalArgumentException("Unsupported scheme: " + getScheme());
             }
             LOG.info(
-                    "{} is not set, using credential provider {}.",
-                    ACCESS_KEY_ID,
+                    "Using credential provider {} for delegated tokens.",
                     hadoopConfig.get(PROVIDER_CONFIG_NAME));
-        } else {
-            LOG.info("{} is set, using provided access key id and secret.", ACCESS_KEY_ID);
         }
     }
 }
--- a/fluss-filesystems/fluss-fs-s3/src/main/java/org/apache/fluss/fs/s3/S3FileSystemPlugin.java
+++ b/fluss-filesystems/fluss-fs-s3/src/main/java/org/apache/fluss/fs/s3/S3FileSystemPlugin.java
@@ -62,16 +62,20 @@ public class S3FileSystemPlugin implements FileSystemPlugin {
 
     @Override
     public FileSystem create(URI fsUri, Configuration flussConfig) throws IOException {
-        org.apache.hadoop.conf.Configuration hadoopConfig =
-                mirrorCertainHadoopConfig(getHadoopConfiguration(flussConfig));
-
-        // set credential provider
-        setCredentialProvider(hadoopConfig);
+        org.apache.hadoop.conf.Configuration hadoopConfig = buildHadoopConfiguration(flussConfig);
 
         // create the Hadoop FileSystem
         org.apache.hadoop.fs.FileSystem fs = new S3AFileSystem();
         fs.initialize(getInitURI(fsUri, hadoopConfig), hadoopConfig);
         return new S3FileSystem(getScheme(), fs, hadoopConfig);
+    }
+
+    @VisibleForTesting
+    org.apache.hadoop.conf.Configuration buildHadoopConfiguration(Configuration flussConfig) {
+        org.apache.hadoop.conf.Configuration hadoopConfig =
+                mirrorCertainHadoopConfig(getHadoopConfiguration(flussConfig));
+        setCredentialProvider(hadoopConfig);
+        return hadoopConfig;
     }
 
     org.apache.hadoop.conf.Configuration getHadoopConfiguration(Configuration flussConfig) {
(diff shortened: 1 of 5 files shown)


fresh-borzoni:
@affo @leekeiabstraction @luoyuxia @michaelkoepf 
PTAL -  targeted fix to enable IRSA support without the full redesign from #2662 🙏

affo:
Thanks @fresh-borzoni!
I am going to have a full review tomorrow morning!

fresh-borzoni:
Addressed copilot's comments

affo (approved):
I really like this one, thank you @fresh-borzoni for the contribution!

I like this cleaner client/server detection separation and I like that you achieved a "partial" separation between server side credentials and delegated credentials, as you can now make the server authenticate using env vars (for example) and delegate via roles.

I think the missing part now would still be to be able to have 2 different roles, one for authenticating on the server side and one for delegating, or any possible collision on the server side. Let me rephrase: you can now split server access wrt delegation only if the credentials provider is "different". Which is already a big step ahead 🎉 

Probably it is even more subtle, as I think you can differentiate only if the server authentication uses a credentials provider that comes sooner in the chain than the one you use for delegating 🤔 Correct me if I am wrong :)

Another missing part is the ability to disable the token delegation as a whole, but that's another story.

The only nit I have in mind: do we need to document this behavior? Probably better to add something to the S3 docs, maybe worth an example.

fluss-filesystems/fluss-fs-s3/src/test/java/org/apache/fluss/fs/s3/token/S3DelegationTokenProviderTest.java
```diff
@@ -0,0 +1,47 @@
...
+package org.apache.fluss.fs.s3.token;
+
+import org.apache.hadoop.conf.Configuration;
+import org.junit.jupiter.api.Test;
+
+import static org.assertj.core.api.Assertions.assertThatCode;
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
+
+/** Tests for {@link S3DelegationTokenProvider} constructor validation. */
+class S3DelegationTokenProviderTest {
```
morazow, line 27: Could you please add test for setting only `fs.s3a.access.` key without `fs.s3a.secret.key` (or vice versa)? This is new logic.

fluss-filesystems/fluss-fs-s3/src/main/java/org/apache/fluss/fs/s3/S3FileSystemPlugin.java
```diff
@@ -121,8 +125,19 @@ private URI getInitURI(URI fsUri, org.apache.hadoop.conf.Configuration hadoopCon
         return fsUri;
     }
 
-    private void setCredentialProvider(org.apache.hadoop.conf.Configuration hadoopConfig) {
-        if (hadoopConfig.get(ACCESS_KEY_ID) == null) {
+    @VisibleForTesting
+    void setCredentialProvider(org.apache.hadoop.conf.Configuration hadoopConfig) {
```
morazow: I think this method was changed to package private so that tests could use it, and validate the hadoop.

Alternatively we could extract a method:

```java
 @VisibleForTesting
 org.apache.hadoop.conf.Configuration buildHadoopConfiguration(Configuration flussConfig) {
     org.apache.hadoop.conf.Configuration hadoopConfig =
             mirrorCertainHadoopConfig(getHadoopConfiguration(flussConfig));
     setCredentialProvider(hadoopConfig);
     return hadoopConfig;
 }

// update create also

 @Override
public FileSystem create(URI fsUri, Configuration flussConfig) throws IOException {
     org.apache.hadoop.conf.Configuration hadoopConfig = buildHadoopConfiguration(flussConfig);
     ...
}

```

Keep the `setCredentialProvider` private again and use extracted method in `create` and update tests with extracted method call.

I find this cleaner since the `mirrorCertainHadoopConfig` and `setCredentialsProvider` methods cannot be called separately.

morazow (changes requested):
Thanks @fresh-borzoni, I have added minor suggestions 🤝

fluss-filesystems/fluss-fs-s3/src/main/java/org/apache/fluss/fs/s3/token/S3DelegationTokenProvider.java
michaelkoepf: consistency: `checkNotNull()` -> `checkArgument()`?

if we now use `checkArgument` to validate `accessKey`, `secretKey` and `roleArn`, we should probably also validate `region` the same way?

fresh-borzoni:
Thanks @affo and @morazow ! Really appreciate the thorough review.
                                                        
@affo  You're right on the credential chain subtlety -  in practice this works cleanly for the IRSA case (the most common deployment pattern on EKS), where the same service account identity handles both data access and the STS AssumeRole call. The two-role separation and delegation disable are good candidates for the full redesign in #2662.
                                                                                                                                                                                                                                                     
Added a docs section for the IRSA/instance profile setup with a config example and IAM requirements. Also addressed @morazow 's feedback - partial credential validation test and a cleaner buildHadoopConfiguration extraction.

fresh-borzoni:
@luoyuxia PTAL as well, pls. 

Mb we wish to backport this to 0.9 as this unblocks proper way to deploy on AWS/EKS. Bc of this a lot of teams wouldn't even consider deploying fluss tbh

fluss-filesystems/fluss-fs-s3/src/main/java/org/apache/fluss/fs/s3/token/S3DelegationTokenProvider.java
```diff
@@ -72,6 +73,16 @@ public S3DelegationTokenProvider(String scheme, Configuration conf) {
         this.secretKey = conf.get(ACCESS_KEY_SECRET);
         this.roleArn = conf.get(ROLE_ARN_KEY);
         this.stsEndpoint = conf.get(STS_ENDPOINT_KEY);
+
+        checkArgument(
+                (accessKey == null) == (secretKey == null),
+                "fs.s3a.access.key and fs.s3a.secret.key must both be set or both be unset.");
```
michaelkoepf: nit: use a more generic name in the error message (cf. region error message)? users might get confused, because the mirroring is an implementation detail and they may have used a different prefix or even different config name. e.g., `access-key` is also supported, iirc.

fluss-filesystems/fluss-fs-s3/src/main/java/org/apache/fluss/fs/s3/token/S3DelegationTokenProvider.java
```diff
@@ -72,6 +73,16 @@ public S3DelegationTokenProvider(String scheme, Configuration conf) {
...
+        if (accessKey == null) {
+            checkArgument(
+                    roleArn != null,
+                    ROLE_ARN_KEY + " must be set when static credentials are not provided.");
```
michaelkoepf: cf. https://github.com/apache/fluss/pull/3067/changes#r3086668521

michaelkoepf (commented):
see comments. i might do another pass over the code, but otherwise i think lgtm

fresh-borzoni:
@michaelkoepf Ty for the review, addressed

morazow (approved):
Thanks, looks fine from my side 🤝

affo:
@fresh-borzoni thanks for adding that note in the documentation, for me it was approved already, we just need a committer to land this, @luoyuxia @polyzos would you mind?

michaelkoepf (approved):

Issue #3070, opened 2026-04-14
[server] Leader-only rebalance tasks should wait for tablet server acknowledgment before proceeding to the next migration


swuferhong:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/fluss/issues) and found nothing similar.

### Motivation

During rebalance, replica migration tasks (e.g., [0,1,2] → [3,4,5]) correctly wait for the task to complete before proceeding to the next migration, which avoids overloading the system. However, leader-only migration tasks (e.g., leader 0 → 1) are completed immediately after triggering the election, without waiting for the tablet server to acknowledge the change. This causes all leader migrations to fire simultaneously, which puts excessive pressure on tablet servers, especially for KV tables that need to rebuild state during leader transitions.

### Willingness to contribute

- [ ] I'm willing to submit a PR!

PR #3071, opened 2026-04-14
[server] Make leader-only rebalance tasks execute sequentially

swuferhong:
### Purpose

Linked issue: #3070 

Previously, leader-only rebalance tasks (where only the leader changes
but the replica set stays the same) were completed immediately after
triggering the leader election, without waiting for the TabletServer
to acknowledge the change. This caused all leader migrations to fire
simultaneously, putting excessive pressure on tablet servers, especially
for KV tables.

This patch removes the immediate `finishRebalanceTask()` call for
leader-only changes and instead completes the task in
`processNotifyLeaderAndIsrResponseReceivedEvent()` when the TabletServer
responds. This ensures leader migrations execute one at a time, matching
the behavior of replica migration tasks.

Also fixes `TestTabletServerGateway` to return per-bucket success results
in `NotifyLeaderAndIsr` responses, which is necessary for the coordinator
to identify acknowledged buckets in the response callback.

### DIFF
--- a/fluss-server/src/main/java/org/apache/fluss/server/coordinator/CoordinatorEventProcessor.java
+++ b/fluss-server/src/main/java/org/apache/fluss/server/coordinator/CoordinatorEventProcessor.java
@@ -980,6 +980,14 @@ public class CoordinatorEventProcessor implements EventProcessor {
             // trigger replicas to offline
             onReplicaBecomeOffline(offlineReplicas);
         }
+
+        // Try to complete rebalance tasks for the buckets in the response.
+        // This is essential for leader-only migrations to ensure they wait for the tablet
+        // server to acknowledge the leader change before proceeding to the next migration.
+        for (NotifyLeaderAndIsrResultForBucket notifyLeaderAndIsrResultForBucket :
+                notifyLeaderAndIsrResultForBuckets) {
+            tryToCompleteRebalanceTask(notifyLeaderAndIsrResultForBucket.getTableBucket());
+        }
     }
 
     private void onReplicaBecomeOffline(Set<TableBucketReplica> offlineReplicas) {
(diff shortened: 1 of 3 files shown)


fluss-server/src/main/java/org/apache/fluss/server/coordinator/CoordinatorEventProcessor.java
```diff
@@ -980,6 +980,14 @@ private void processNotifyLeaderAndIsrResponseReceivedEvent(
             // trigger replicas to offline
             onReplicaBecomeOffline(offlineReplicas);
         }
+
+        // Try to complete rebalance tasks for the buckets in the response.
+        // This is essential for leader-only migrations to ensure they wait for the tablet
+        // server to acknowledge the leader change before proceeding to the next migration.
+        for (NotifyLeaderAndIsrResultForBucket notifyLeaderAndIsrResultForBucket :
+                notifyLeaderAndIsrResultForBuckets) {
+            tryToCompleteRebalanceTask(notifyLeaderAndIsrResultForBucket.getTableBucket());
```
LiebingYu, line 989: After switching to sequential leader transfer one bucket at a time, a new issue arises: if the leader transfer fails for any single bucket, it may block all subsequent leader transfers. However, I understand it is acceptable to not address this issue for now, but users must manually handle such blocking situations. WDYT?

LiebingYu (commented):
@swuferhong Thanks for the PR, only minor comments for me, not block the PR.

wuchong (approved):

fluss-server/src/main/java/org/apache/fluss/server/coordinator/CoordinatorEventProcessor.java
swuferhong: Thinks @LiebingYu. This problem will be trace by this issue: https://github.com/apache/fluss/issues/3096

PR #3088, opened 2026-04-15
[Blog] Column Pruning

polyzos:


polyzos:
@swuferhong @wuchong I took the liberty of porting this to our blog, as it's good and it should make it easier for users to find it under our blog. 
Let me know your thoughts.

swuferhong:
Thanks @polyzos. LGTM.

swuferhong (approved):

website/blog/authors.yml
```diff
@@ -78,4 +78,10 @@ yangwang:
...
   url: https://github.com/platinumhamburg
-  image_url: /img/avatars/yangwang.jpeg
\ No newline at end of file
+  image_url: /img/avatars/yangwang.jpeg
+
+yunhongzheng:
+  name: Yunhong Zheng
+  title: PPMC member of Apache Fluss (Incubating)
+  url: https://github.com/swuferhong
+  image_url: https://github.com/swuferhong.png
```
wuchong: We should use local image rather than external resources, ASF doesn't allow external resources and it will be blocked.

wuchong:
continue in https://github.com/apache/fluss-blog/pull/2

PR #2863, opened 2026-03-13
[Helm] Revert Chart Component Configuration Isolation (#2472)

affo:
This reverts commit d956e75b6f45dcd5264dafd5fd31c4a7a8cd8c04.

### Purpose

Linked issue: close #xxx

Reverts https://github.com/apache/fluss/pull/2472

### Brief change log

The PR landed accidentally, but the review was not addressed: https://github.com/apache/fluss/pull/2472#pullrequestreview-3936502333

PR #3098, opened 2026-04-16
[helm] Support injecting environment variables and external secrets

### LINKED ISSUE #3097: [helm] Support injecting environment variables and external secrets
### Search before asking

### Description

The Fluss Helm chart currently has no way to inject custom environment variables or reference external Kubernetes Secrets/ConfigMaps in the coordinator and tablet server pods. 

This prevents using the External Secrets Operator or any other env-based configuration without modifying the chart templates directly.   

### Willingness to contribute

fresh-borzoni:
## Summary

closes https://github.com/apache/fluss/issues/3097

Enable users to inject custom environment variables into coordinator and tablet server pods via `extraEnv` and `envFrom` fields, supporting integration with the External Secrets Operator and similar tools that provision credentials from external stores

### DIFF
--- a/.github/workflows/helm-chart.yaml
+++ b/.github/workflows/helm-chart.yaml
@@ -48,7 +48,6 @@ jobs:
         run: helm lint ./helm
 
       - name: "Run helm-unittest"
-        uses: d3adb5/helm-unittest-action@v2
-        with:
-          helm-version: latest
-          charts: helm
+        run: |
+          helm plugin install https://github.com/helm-unittest/helm-unittest
+          helm unittest ./helm
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -24,10 +24,12 @@ on:
       - ci-**
     paths-ignore:
       - 'website/**'
+      - 'helm/**'
       - '**/*.md'
   pull_request:
     paths-ignore:
       - 'website/**'
+      - 'helm/**'
       - '**/*.md'
 
 concurrency:
--- a/helm/templates/sts-coordinator.yaml
+++ b/helm/templates/sts-coordinator.yaml
@@ -68,6 +68,13 @@ spec:
             - name: FLUSS_ENV_JAVA_OPTS
               value: "-Djava.security.auth.login.config=/etc/fluss/conf/jaas.conf"
             {{- end }}
+            {{- with .Values.coordinator.extraEnv }}
+            {{- toYaml . | nindent 12 }}
+            {{- end }}
+          {{- with .Values.coordinator.envFrom }}
+          envFrom:
+            {{- toYaml . | nindent 12 }}
+          {{- end }}
           ports:
             - name: internal
               containerPort: {{ .Values.listeners.internal.port }}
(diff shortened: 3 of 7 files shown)


fresh-borzoni:
@affo @morazow @polyzos @dreger1997  PTAL 🙏

affo (approved):
Perfect, thanks for this addon.

I should also have a look to the external secret operator myself, thanks for the pointer 🚀

affo:
@fresh-borzoni also, thanks for the fixes in CI 🤝

.github/workflows/helm-chart.yaml
```diff
@@ -48,7 +48,6 @@ jobs:
         run: helm lint ./helm
 
       - name: "Run helm-unittest"
-        uses: d3adb5/helm-unittest-action@v2
-        with:
-          helm-version: latest
-          charts: helm
+        run: |
```
morazow, line 51: Weren't this working? 🤦

morazow (approved):
Looks good, thanks!

Only comment from my side to understand the change 🤝

.github/workflows/helm-chart.yaml
affo: https://github.com/apache/fluss/actions/runs/24516303125

I think we actually never ran this checks in CI, take for example http://github.com/apache/fluss/pull/2863:

image

@fresh-borzoni fixed CI so that now checks actually run (or maybe a setting was simply fixed? I did not understand what was wrong and why that check was not running), and that did not work:

```
The action d3adb5/helm-unittest-action@v2 is not allowed in apache/fluss because all actions must be from a repository owned by your enterprise, created by GitHub, or match one of the patterns: 1Password/load-secrets-action@13f58eec611f8e5db52ec16247f58c508398f3e6, 1Password/load-secrets-action@8d0d610af187e78a2772c2d18d627f4c52d3fbfb, 1Password/load-secrets-action@92467eb28f72e8255933372f1e0707c567ce2259, 1Password/load-secrets-action@dafbe7cb03502b260e2b2893c753c352eee545bf, AdoptOpenJDK/install-jdk@*, BobAnkh/auto-generate-changelog@*, DavidAnson/markdownlint-cli2-action@07035fd053f7be764496c0f8d8f9f41f98305101, DavidAnson/markdownlint-cli2-action@ce4853d43830c74c1753b39f3cf40f71c2031eb9, EnricoMi/publish-unit-test-result-action@*, JamesIves/github-pages-deploy-action@4a3abc783e1a24aeb44c16e869ad83caf6b4cc23, JamesIves/github-pages-deploy-action@d92aa235d04922e8f08b40ce78cc5442fcfbfa2f, JetBrains/qodana-action@89eb4357efd2b52e639f3216e63edaf33b82622b, Jimver/cuda-toolkit@3d45d157...
```

So, we are not allowed to run that action.

@fresh-borzoni converted this to simply install unittest.

morazow: Ahh got it, thanks for explaining 🤝 

I remember in the Flink repo there were some restrictions on using actions, but good to keep in mind. Maybe ASF have guidelines on external actions.

morazow: Anyway with recent supply chain attacks (e.g, trivy action), this is definitely right change, often it is hard to know what comes with an action.

fresh-borzoni: yes, basically this action isn't on allowlist, so two option - add it explicitly, which requires reaching to ASF support, or do like this.

Kudos to @luoyuxia btw

PR #3102, opened 2026-04-17
[client] release the resource of record accumulator only when sender thread is stopped.

### LINKED ISSUE #3101: java.lang.IllegalArgumentException: Accounted size went negative.
### Search before asking

### Fluss version

0.9.0 (latest release)

### Please describe the bug 🐞

The error is
```java
2026-04-16 22:49:50,872 ERROR org.apache.fluss.client.write.Sender                         [] - Uncaught error in Fluss write sender I/O thread: 
java.lang.IllegalArgumentException: Accounted size went negative.
	at org.apache.fluss.shaded.arrow.org.apache.arrow.util.Preconditions.checkArgument(Preconditions.java:136) ~[blob_p-4eb0b65d03350a588b07f4bbbc14363610125495-2eda0538dee6bd4827a2612b21dd4301:0.10-SNAPSHOT]
	at org.apache.fluss.shaded.arrow.org.apache.arrow.memory.Accountant.releaseBytes(Accountant.java:219) ~[blob_p-4eb0b65d03350a588b07f4bbbc14363610125495-2eda0538dee6bd4827a2612b21dd4301:0.10-SNAPSHOT]
	at org.apache.fluss.shaded.arrow.org.apache.arrow.memory.RootAllocator.releaseBytes(RootAllocator.java:29) ~[blob_p-4eb0b65d03350a588b07f4bbbc14363610125495-2eda0538dee6bd4827a2612b21dd4301:0.10-SNAPSHOT]
	at org.apache.fluss.shaded.arrow.org.apache.arrow.memory.AllocationManager.release(AllocationManager.java:152) ~[blob_p-4eb0b65d03350a588b07f4bbbc14363610125495-2eda0538dee6bd4827a2612b21dd4301:0.10-SNAPSHOT]
	at org.apache.fluss.shaded.arrow.org.apache.arrow.memory.BufferLedger.decrement(BufferLedger.java:157) ~[blob_p-4eb0b65d03350a588b07f4bbbc14363610125495-2eda0538dee6bd4827a2612b21dd4301:0.10-SNAPSHOT]
	at org.apache.fluss.shaded.arrow.org.apache.arrow.memory.BufferLedger.release(BufferLedger.java:124) ~[blob_p-4eb0b65d03350a588b07f4bbbc14363610125495-2eda0538dee6bd4827a2612b21dd4301:0.10-SNAPSHOT]
	at org.apache.fluss.shaded.arrow.org.apache.arrow.memory.BufferLedger.release(BufferLedger.java:104) ~[blob_p-4eb0b65d03350a588b07f4bbbc14363610125495-2eda0538dee6bd4827a2612b21dd4301:0.10-SNAPSHOT]
	at org.apache.fluss.shaded.arrow.org.apache.arrow.vector.BaseValueVector.releaseBuffer(BaseValueVector.java:117) ~[blob_p-4eb0b65d03350a588b07f4bbbc14363610125495-2eda0538dee6bd4827a2612b21dd4301:0.10-SNAPSHOT]
	at org.apache.fluss.shaded.arrow.org.apache.arrow.vector.BaseVariableWidthVector.clear(BaseVariableWidthVector.java:270) ~[blob_p-4eb0b65d03350a588b07f4bbbc14363610125495-2eda0538dee6bd4827a2612b21dd4301:0.10-SNAPSHOT]
	at org.apache.fluss.shaded.arrow.org.apache.arrow.vector.VectorSchemaRoot.clear(VectorSchemaRoot.java:157) ~[blob_p-4eb0b65d03350a588b07f4bbbc14363610125495-2eda0538dee6bd4827a2612b21dd4301:0.10-SNAPSHOT]
	at org.apache.fluss.row.arrow.ArrowWriter.recycle(ArrowWriter.java:308) ~[blob_p-4eb0b65d03350a588b07f4bbbc14363610125495-2eda0538dee6bd4827a2612b21dd4301:0.10-SNAPSHOT]
	at org.apache.fluss.record.MemoryLogRecordsArrowBuilder.build(MemoryLogRecordsArrowBuilder.java:229) ~[blob_p-4eb0b65d03350a588b07f4bbbc14363610125495-2eda0538dee6bd4827a2612b21dd4301:0.10-SNAPSHOT]
	at org.apache.fluss.record.MemoryLogRecordsArrowBuilder.close(MemoryLogRecordsArrowBuilder.java:310) ~[blob_p-4eb0b65d03350a588b07f4bbbc14363610125495-2eda0538dee6bd4827a2612b21dd4301:0.10-SNAPSHOT]
	at org.apache.fluss.client.write.ArrowLogWriteBatch.close(ArrowLogWriteBatch.java:104) ~[blob_p-4eb0b65d03350a588b07f4bbbc14363610125495-2eda0538dee6bd4827a2612b21dd4301:0.10-SNAPSHOT]
	at org.apache.fluss.client.write.RecordAccumulator.drainBatchesForOneNode(RecordAccumulator.java:773) ~[blob_p-4eb0b65d03350a588b07f4bbbc14363610125495-2eda0538dee6bd4827a2612b21dd4301:0.10-SNAPSHOT]
	at org.apache.fluss.client.write.RecordAccumulator.drain(RecordAccumulator.java:288) ~[blob_p-4eb0b65d03350a588b07f4bbbc14363610125495-2eda0538dee6bd4827a2612b21dd4301:0.10-SNAPSHOT]
	at org.apache.fluss.client.write.Sender.sendWriteData(Sender.java:243) ~[blob_p-4eb0b65d03350a588b07f4bbbc14363610125495-2eda0538dee6bd4827a2612b21dd4301:0.10-SNAPSHOT]
	at org.apache.fluss.client.write.Sender.runOnce(Sender.java:194) ~[blob_p-4eb0b65d03350a588b07f4bbbc14363610125495-2eda0538dee6bd4827a2612b21dd4301:0.10-SNAPSHOT]
	at org.apache.fluss.client.write.Sender.run(Sender.java:164) [blob_p-4eb0b65d03350a588b07f4bbbc14363610125495-2eda0538dee6bd4827a2612b21dd4301:0.10-SNAPSHOT]
	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:515) [?:?]
	at java.util.concurrent.FutureTask.run(FutureTask.java:264) [?:?]
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128) [?:?]
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628) [?:?]
```

## Root Cause Analysis

The error `Accounted size went negative` is caused by a **shutdown race condition** in [`WriterClient.close()`](file:///Users/loserwang/workstation/github/fluss/fluss-client/src/main/java/org/apache/fluss/client/write/WriterClient.java#L312-L343).

### The Shutdown Sequence (Buggy)

In [`WriterClient.close()`](file:///Users/loserwang/workstation/github/fluss/fluss-client/src/main/java/org/apache/fluss/client/write/WriterClient.java#L312):

```
Step 1: sender.initiateClose()     ← closes allocator + sets running=false
Step 2: ioThreadPool.shutdown()    ← signals thread pool to stop
Step 3: awaitTermination(timeout)  ← waits for sender thread
Step 4: sender.forceClose()        ← sets forceClose=true
```

The problem: **Step 1 closes the Arrow `BufferAllocator` BEFORE the sender thread has stopped**.

### The Race

1. [`sender.initiateClose()`](file:///Users/loserwang/workstation/github/fluss/fluss-client/src/main/java/org/apache/fluss/client/write/Sender.java#L640-L645) calls [`accumulator.close()`](file:///Users/loserwang/workstation/github/fluss/fluss-client/src/main/java/org/apache/fluss/client/write/RecordAccumulator.java#L964-L974), which:
   - Closes the `arrowWriterPool`
   - Calls `bufferAllocator.releaseBytes(bufferAllocator.getAllocatedMemory())` — resets allocator to 0
   - Calls `bufferAllocator.close()` — closes the allocator
   - Sets `running = false`

2. The sender thread is still executing. It sees `running = false`, exits the main loop, and enters the **shutdown drain loop** at [line 162](file:///Users/loserwang/workstation/github/fluss/fluss-client/src/main/java/org/apache/fluss/client/write/Sender.java#L162-L168):
   ```java
   while (!forceClose && accumulator.hasUnDrained()) {  // forceClose is still FALSE here!
       runOnce();  // tries to drain remaining batches
   }
   ```

3. `runOnce()` → `drain()` → [`drainBatchesForOneNode()`](file:///Users/loserwang/workstation/github/fluss/fluss-client/src/main/java/org/apache/fluss/client/write/RecordAccumulator.java#L777) → `batch.close()` → [`build()`](file:///Users/loserwang/workstation/github/fluss/fluss-common/src/main/java/org/apache/fluss/record/MemoryLogRecordsArrowBuilder.java#L229) → [`ArrowWriter.recycle()`](file:///Users/loserwang/workstation/github/fluss/fluss-common/src/main/java/org/apache/fluss/row/arrow/ArrowWriter.java#L308) → `root.clear()` → tries to release bytes from the **already-closed allocator** (whose accounting is already 0) → **"Accounted size went negative"**

4. The error is caught by the shutdown loop at line 166, logging exactly the message you see:
   ```
   "Uncaught error in Fluss write sender I/O thread: "
   ```

### Fix

The fix is to separate the "stop accepting new appends" phase from the "release resources" phase. `initiateClose()` should only prevent new appends and signal the sender to stop, but NOT destroy the allocator. Resource cleanup should happen AFTER the sender thread has fully exited.

loserwang1024:
### Purpose

Linked issue: close #3101

### DIFF
--- a/fluss-client/src/main/java/org/apache/fluss/client/write/RecordAccumulator.java
+++ b/fluss-client/src/main/java/org/apache/fluss/client/write/RecordAccumulator.java
@@ -59,6 +59,7 @@ import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
 import java.util.concurrent.ConcurrentMap;
+import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicInteger;
 
 import static org.apache.fluss.record.LogRecordBatchFormat.NO_BATCH_SEQUENCE;
@@ -104,7 +105,7 @@ public final class RecordAccumulator {
     private final ChunkedAllocationManager.ChunkedFactory chunkedFactory;
 
     /** Guard to make {@link #destroyResources()} idempotent. */
-    private volatile boolean resourcesDestroyed = false;
+    private final AtomicBoolean resourcesDestroyed = new AtomicBoolean(false);
 
     /** The pool of lazily created arrow {@link ArrowWriter}s for arrow log write batch. */
     private final ArrowWriterPool arrowWriterPool;
@@ -978,15 +979,14 @@ public final class RecordAccumulator {
      *
      * <p>This method is idempotent: subsequent calls after the first are no-ops.
      */
+    @VisibleForTesting
     public void destroyResources() {
-        if (resourcesDestroyed) {
+        if (!resourcesDestroyed.compareAndSet(false, true)) {
             return;
         }
-        resourcesDestroyed = true;
         writerBufferPool.close();
         arrowWriterPool.close();
         bufferAllocator.close();
-        // Release native memory held by the chunked allocation manager factory.
         chunkedFactory.close();
     }
 
--- a/fluss-client/src/main/java/org/apache/fluss/client/write/RecordAccumulator.java
+++ b/fluss-client/src/main/java/org/apache/fluss/client/write/RecordAccumulator.java
@@ -103,6 +103,9 @@ public final class RecordAccumulator {
     /** The chunked allocation manager factory, stored for explicit native memory release. */
     private final ChunkedAllocationManager.ChunkedFactory chunkedFactory;
 
+    /** Guard to make {@link #destroyResources()} idempotent. */
+    private volatile boolean resourcesDestroyed = false;
+
     /** The pool of lazily created arrow {@link ArrowWriter}s for arrow log write batch. */
     private final ArrowWriterPool arrowWriterPool;
 
@@ -960,14 +963,28 @@ public final class RecordAccumulator {
         }
     }
 
-    /** Close this accumulator and force all the record buffers to be drained. */
+    /** Close this accumulator to reject new appends. */
     public void close() {
         closed = true;
+    }
 
+    /**
+     * Destroy all resources held by this accumulator including the Arrow writer pool and buffer
+     * allocator.
+     *
+     * <p>This must only be called after the sender thread has fully exited and no more drain
+     * operations will occur. Otherwise, draining batches may attempt to recycle Arrow writers using
+     * an already-closed allocator, causing "Accounted size went negative" errors.
+     *
+     * <p>This method is idempotent: subsequent calls after the first are no-ops.
+     */
+    public void destroyResources() {
+        if (resourcesDestroyed) {
+            return;
+        }
+        resourcesDestroyed = true;
         writerBufferPool.close();
         arrowWriterPool.close();
-        // Release all the memory segments.
-        bufferAllocator.releaseBytes(bufferAllocator.getAllocatedMemory());
         bufferAllocator.close();
         // Release native memory held by the chunked allocation manager factory.
         chunkedFactory.close();
--- a/fluss-client/src/main/java/org/apache/fluss/client/write/Sender.java
+++ b/fluss-client/src/main/java/org/apache/fluss/client/write/Sender.java
@@ -167,6 +167,8 @@ public class Sender implements Runnable {
             }
         }
 
+        destroyResources();
+
         // TODO if force close failed, add logic to abort incomplete batches.
         LOG.debug("Shutdown of Fluss write sender I/O thread has completed.");
     }
@@ -642,5 +644,19 @@ public class Sender implements Runnable {
         // breaking from the sender loop. Otherwise, we may miss some callbacks when shutting down.
         accumulator.close();
         running = false;
+    }
+
+    /**
+     * Releases all resources held by the accumulator (Arrow writer pool, buffer allocator, etc.).
+     *
+     * <p>In production, this is called at the end of {@link #run()} after both the main loop and
+     * the shutdown drain loop have completed, so no further drain operations will touch the Arrow
+     * allocator.
+     *
+     * <p>This method is idempotent: repeated calls are harmless because the underlying {@link
+     * RecordAccumulator#destroyResources()} tolerates multiple invocations.
+     */
+    void destroyResources() {
+        accumulator.destroyResources();
     }
 }
--- a/fluss-client/src/test/java/org/apache/fluss/client/write/SenderTest.java
+++ b/fluss-client/src/test/java/org/apache/fluss/client/write/SenderTest.java
@@ -43,6 +43,7 @@ import org.apache.fluss.rpc.protocol.Errors;
 import org.apache.fluss.server.tablet.TestTabletServerGateway;
 import org.apache.fluss.utils.clock.SystemClock;
 
+import org.junit.jupiter.api.AfterEach;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
 
@@ -99,6 +100,11 @@ final class SenderTest {
         metadataUpdater = initializeMetadataUpdater();
         writerMetricGroup = TestingWriterMetricGroup.newInstance();
         sender = setupWithIdempotenceState();
+    }
+
+    @AfterEach
+    public void teardown() throws Exception {
+        sender.destroyResources();
     }
 
     @Test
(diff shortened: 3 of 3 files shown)


fluss-client/src/main/java/org/apache/fluss/client/write/RecordAccumulator.java
```diff
@@ -978,4 +986,8 @@ public BucketAndWriteBatches(@Nullable Long partitionId, boolean isPartitionedTa
             this.isPartitionedTable = isPartitionedTable;
         }
     }
+
+    public BufferAllocator getBufferAllocator() {
```
utafrali: `getBufferAllocator()` is not used anywhere in the codebase (neither in production code nor in tests). Either remove it, or if it is intended for future test use, annotate it with `@VisibleForTesting` and move it to match the project convention for test-support methods. As written it leaks an internal implementation detail via the public API with no clear consumer.

fluss-client/src/main/java/org/apache/fluss/client/write/Sender.java
```diff
@@ -643,4 +645,9 @@ public void initiateClose() {
         accumulator.close();
         running = false;
     }
+
```
utafrali, line 648: `destroyResources()` has no Javadoc. It is package-private specifically so tests can call it directly in teardown, which means callers outside `run()` must understand the lifecycle contract. Add a brief Javadoc explaining: (1) under normal operation this is called automatically at the end of `run()`, (2) calling it outside that context (e.g. in tests that only use `runOnce()`) is safe only because no drain loop is active. Without this note, a future test author using the full `run()` on a background thread will silently call `destroyResources()` twice — once from `run()` and once from `@AfterEach teardown()`. Arrow's `BufferAllocator.close()` is not idempotent and will throw on a second call. Consider adding an `AtomicBoolean destroyed` guard inside `destroyResources()` to make the method safe against accidental double invocation.

fluss-client/src/test/java/org/apache/fluss/client/write/SenderTest.java
```diff
@@ -171,6 +177,21 @@ void testInitWriterIdRequest() throws Exception {
...
+        //  Append a record so there is an undrained batch.
+        appendToAccumulator(tb1, row(1, "a"), (tb, leo, e) -> {});
+
+        // Sender begin to sendWriteData.
+        Cluster clusterSnapshot = metadataUpdater.getCluster();
+        RecordAccumulator.ReadyCheckResult readyCheckResult = accumulator.ready(clusterSnapshot);
+
+        // close the accumulator before accumulator drains.
+        sender.initiateClose();
+        accumulator.drain(clusterSnapshot, readyCheckResult.readyNodes, 1);
```
utafrali: The regression test `testCloseSenderBeforeAccumulatorDrain()` carries no assertions; it passes only by not throwing. Add a closing comment or assertion to make the intent explicit, e.g.:
```java
// Before the fix this threw "Accounted size went negative" because the Arrow
// allocator was closed while the drain was still running.
assertThat(accumulator.getBufferAllocator().getAllocatedMemory()).isZero();
```
This also gives the test a meaningful failure message if the invariant breaks again.

utafrali (commented):
The core fix is correct: splitting `close()` (reject appends) from `destroyResources()` (free Arrow resources) and calling the latter only after all drain loops have finished cleanly resolves the 'Accounted size went negative' race. The main issues are an unused public getter that leaks internal state, missing lifecycle documentation on `Sender.destroyResources()` that hides a latent double-close risk, and a regression test with no assertion.

fluss-client/src/main/java/org/apache/fluss/client/write/RecordAccumulator.java
loserwang1024: my mistake, I add this just for my test but forget remove it.

fluss-client/src/main/java/org/apache/fluss/client/write/Sender.java
loserwang1024: done, make sure destroyResources() safe against accidental double invocation

loserwang1024:
@platinumhamburg ,CC

fluss-client/src/main/java/org/apache/fluss/client/write/RecordAccumulator.java
```diff
@@ -960,14 +963,28 @@ public ReadyCheckResult(
...
+     * allocator.
+     *
+     * <p>This must only be called after the sender thread has fully exited and no more drain
+     * operations will occur. Otherwise, draining batches may attempt to recycle Arrow writers using
+     * an already-closed allocator, causing "Accounted size went negative" errors.
+     *
+     * <p>This method is idempotent: subsequent calls after the first are no-ops.
+     */
+    public void destroyResources() {
+        if (resourcesDestroyed) {
```
platinumhamburg: `volatile` guarantees visibility but does **not** guarantee atomicity of the check + set. Two threads can both read `false`, pass the guard, and both enter the destruction path. Arrow's `BufferAllocator.close()` is not idempotent and will throw `IllegalStateException` on a second invocation.

The Javadoc claims "This method is idempotent: subsequent calls after the first are no-ops", but the implementation cannot honor this guarantee under concurrent access.

**Suggested fix:**

```java
private final AtomicBoolean resourcesDestroyed = new AtomicBoolean(false);

public void destroyResources() {
    if (!resourcesDestroyed.compareAndSet(false, true)) {
        return;
    }
    writerBufferPool.close();
    arrowWriterPool.close();
    bufferAllocator.close();
    chunkedFactory.close();
}
```

platinumhamburg (changes requested):

fluss-client/src/main/java/org/apache/fluss/client/write/Sender.java
```diff
@@ -643,4 +645,18 @@ public void initiateClose() {
...
+     * Releases all resources held by the accumulator (Arrow writer pool, buffer allocator, etc.).
+     *
+     * <p>In production, this is called at the end of {@link #run()} after both the main loop and
+     * the shutdown drain loop have completed, so no further drain operations will touch the Arrow
+     * allocator.
+     *
+     * <p>This method is idempotent: repeated calls are harmless because the underlying {@link
+     * RecordAccumulator#destroyResources()} tolerates multiple invocations.
+     */
+    void destroyResources() {
```
platinumhamburg, line 659: Missing `@VisibleForTesting`

platinumhamburg (changes requested):
Thanks @loserwang1024, I left some comments.

platinumhamburg (approved):
LGTM.

PR #3107, opened 2026-04-17
Handle coordinator disconnect in TieringSourceEnumerator

### LINKED ISSUE #3106: TieringSource should fail over when coordinator IP changes and heartbeat retries a stale endpoint
### Search before asking

### Fluss version

0.8.0

### Please describe the bug 🐞

When the coordinator server restarts and its IP changes, the tiering service in `TieringSourceEnumerator` may keep retrying requests against the old coordinator address instead of refreshing metadata and recovering.

In our production setup, the tiering service requests tiering table splits via heartbeat. After the coordinator server restarts with a new IP, the job starts failing with errors like:

```text
org.apache.fluss.exception.NetworkException: org.apache.fluss.exception.DisconnectException: Cannot send request to server 10.x.x.x:9123 (id: cs-0, rack: null) because it is disconnected.
...
TieringSourceEnumerator [] - Failed to request tiering table, will retry later.
org.apache.flink.util.FlinkRuntimeException: Failed to wait heartbeat response due to
```

The relevant stack trace points to:

- `TieringSourceEnumerator$HeartBeatHelper.waitHeartbeatResponse(TieringSourceEnumerator.java:735)`
- `TieringSourceEnumerator.requestTieringTableSplitsViaHeartBeat(TieringSourceEnumerator.java:512)`

Actual behavior:

- the tiering service keeps retrying with the stale coordinator IP;
- the job does not fail over;
- tiering cannot make progress until manual intervention.

Expected behavior:

- the tiering service should not stay stuck retrying a disconnected old coordinator address forever;
- once this kind of disconnect happens, the job should fail over so that the source can restart and recover with fresh metadata;
- longer term, the client-side retry + metadata refresh path introduced by PR #479 should help regular client requests perceive the latest coordinator IP.

One way to reproduce is:

1. Start a Flink job that uses `TieringSource`.
2. Let the job run normally and request tiering splits.
3. Restart the coordinator server so that `cs-0` comes back with a different IP.
4. Wait for the tiering service to send heartbeat-based split requests.
5. Observe that it keeps retrying the old IP and logs `Failed to request tiering table, will retry later.` repeatedly.

This seems specific to the tiering service path because it keeps scheduling retries after `DisconnectException`/`NetworkException`, but it does not recover its coordinator metadata in time.

### Solution

A practical fix for the tiering service would be to treat this heartbeat failure as fatal and trigger Flink job failover instead of retrying forever.

Concretely, when `TieringSourceEnumerator` hits `DisconnectException` / `NetworkException` while waiting for the heartbeat response during `requestTieringTableSplitsViaHeartBeat`, it can throw a non-recoverable exception to fail the source/job. After the job restarts, it can rebuild the client and recover using the latest coordinator metadata.

This issue should be a relatively good starter task because the scope looks contained:

- identify the retry path in `TieringSourceEnumerator`;
- convert the stale-coordinator heartbeat failure into job failover;
- add a regression test that simulates coordinator disconnect / stale metadata for tiering split requests.

utafrali:
### Purpose
Closes #3106

### Brief change log
Coordinator IP changes can leave stale heartbeat connections that just log errors and retry instead of triggering failover. Added exception checks for `NetworkException` and `DisconnectException` - when we hit one during heartbeat, we throw a `FlinkRuntimeException` to force proper failover. Other error types keep the existing log and retry behavior.

### Tests
Added test cases to verify failover behavior when network or disconnect exceptions occur.

### API and Format
N/A

### Documentation
N/A


fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/tiering/source/enumerator/TieringSourceEnumerator.java
```diff
@@ -352,9 +355,17 @@ protected void handleTableTieringReachMaxDuration(
         }
     }
 
-    private void generateAndAssignSplits(
+    @VisibleForTesting
+    void generateAndAssignSplits(
             @Nullable Tuple3<Long, Long, TablePath> tieringTable, Throwable throwable) {
         if (throwable != null) {
```
luoyuxia: maybe we can just throw the exception？
```
if (throwable != null) {
   throw throwable
}
```
?

luoyuxia (commented):
@utafrali Thanks for the pr. Left minor comments. PTAL
Also, please don't forget to rebase main branch

luoyuxia:
@utafrali Seem you clear the file content.

luoyuxia (approved):

PR #2902, opened 2026-03-18
[helm] Refactor validation warning and error checks

morazow:
### Purpose

Linked issue: close #NA

- Refactors the Helm chart validations
- Adds methods to make it easier to add new warning or error validations

### Brief change log

Refactors the helm templates, introduces a new `_validate.tpl` file, that collects all the validation checks with methods `fluss.validate`. This method then called in the `NOTES.txt` file.

### Tests

Only refactor, existing tests pass.

### API and Format

NA

### Documentation

Updated documentation `Developer` section with information on how to structure and extend the validation checks for new features.

PR #3114, opened 2026-04-17
[helm] Add pod annotations, labels, and pod disruption budget support

### LINKED ISSUE #3115: [helm] Add pod annotations, pod labels, and PodDisruptionBudget support
### Search before asking

### Description

The Fluss Helm chart is missing standard Kubernetes pod-level configuration that production deployments rely on. 

There is currently no way to add custom annotations to pods (needed for Prometheus scraping, Vault agent injection, Datadog, etc.), no way to add custom labels (needed for service mesh routing, monitoring selectors), and no PodDisruptionBudget to protect against draining too many replicas during rolling upgrades or node maintenance. 

### Willingness to contribute

fresh-borzoni:
closes https://github.com/apache/fluss/issues/3115

Production Kubernetes deployments commonly need custom pod annotations for integrations like Prometheus scraping or Vault agent injection, custom labels for service mesh routing and monitoring selectors, and PodDisruptionBudgets to prevent simultaneous eviction of tablet server replicas during node maintenance. 

This PR adds all three as standard Helm chart fields for both coordinator and tablet server components, with PDB disabled by default.

### DIFF
--- a/helm/templates/pdb-coordinator.yaml
+++ b/helm/templates/pdb-coordinator.yaml
@@ -1,0 +1,37 @@
+#
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+{{- if .Values.coordinator.podDisruptionBudget.enabled }}
+apiVersion: policy/v1
+kind: PodDisruptionBudget
+metadata:
+  name: coordinator-server
+  labels:
+  {{- include "fluss.labels" . | nindent 4 }}
+spec:
+  {{- with .Values.coordinator.podDisruptionBudget.minAvailable }}
+  minAvailable: {{ . }}
+  {{- end }}
+  {{- with .Values.coordinator.podDisruptionBudget.maxUnavailable }}
+  maxUnavailable: {{ . }}
+  {{- end }}
+  selector:
+    matchLabels:
+      {{- include "fluss.selectorLabels" . | nindent 6 }}
+      app.kubernetes.io/component: coordinator
+{{- end }}
--- a/helm/templates/pdb-coordinator.yaml
+++ b/helm/templates/pdb-coordinator.yaml
@@ -24,11 +24,11 @@ metadata:
   labels:
   {{- include "fluss.labels" . | nindent 4 }}
 spec:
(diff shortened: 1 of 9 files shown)


fresh-borzoni:
@affo @morazow @polyzos PTAL 🙏

helm/templates/pdb-coordinator.yaml
```diff
@@ -0,0 +1,37 @@
...
+metadata:
+  name: coordinator-server
+  labels:
+  {{- include "fluss.labels" . | nindent 4 }}
+spec:
+  {{- with .Values.coordinator.podDisruptionBudget.minAvailable }}
+  minAvailable: {{ . }}
+  {{- end }}
+  {{- with .Values.coordinator.podDisruptionBudget.maxUnavailable }}
+  maxUnavailable: {{ . }}
```
morazow: Reading the docs on the PDB: https://kubernetes.io/docs/tasks/run-application/configure-pdb/#specifying-a-poddisruptionbudget

One can only specify either `minAvailable` or `maxUnavailable`. Could you please add validation for this?

You can check the `NOTES.txt` for example for now, until https://github.com/apache/fluss/pull/2902 is merged 🤝

morazow: Valid also for tablet PDB

morazow: On similar note, I could not find what will happen if user doesn't set any of `minAvailable` or `maxAvailable`, seems undefined.

To be on safe side, let's also add a validation that one of them should be set when PDB is enabled.

helm/tests/pod_metadata_and_pdb_test.yaml
```diff
@@ -0,0 +1,215 @@
...
+suite: coordinator-pdb
+templates:
+  - templates/pdb-coordinator.yaml
+
+tests:
+  - it: should render coordinator PDB with minAvailable
+    set:
+      coordinator.podDisruptionBudget:
+        enabled: true
+        minAvailable: 1
```
morazow, line 200: Let's also add test with `minAvailable` and `maxUnavailable` set to zero 🤝

helm/templates/sts-coordinator.yaml
```diff
@@ -34,6 +34,13 @@ spec:
       labels:
         {{- include "fluss.selectorLabels" . | nindent 8 }}
         app.kubernetes.io/component: coordinator
+        {{- with .Values.coordinator.podLabels }}
+        {{- toYaml . | nindent 8 }}
+        {{- end }}
```
morazow, line 37: Should we put the `podLabels` before the `selectorLabels`?

I wonder what will happen if the `podLabels` overwrite the `selectorLabels`, definitely the matchLabels would be different.

In any case, either again we could validate that podLabels does not contain any selectorLabels, or change the ordering so that selectorLabels overwrite previously set ones (if any).

morazow (changes requested):
Thanks man @fresh-borzoni 🚀 

I have added some comments and requests 🤝

helm/templates/sts-coordinator.yaml
fresh-borzoni: Reordered, good point 👍

fresh-borzoni:
@morazow Ty for the nice review, addressed, PTAL 🙏

morazow (approved):
Looks good 🤝

affo (approved):
Another fundamental one, thanks @fresh-borzoni 🚀

PR #3131, opened 2026-04-19
[server] Resume deletions off startup via scheduled event

### LINKED ISSUE #2891: [server] Throttle replica deletion on Coordinator startup to prevent event queue starvation
### Search before asking

### Description

Currently, when the Coordinator restarts, it calls `resumeDeletions()` synchronously inside `TableManager.startup()`. This causes all pending table/partition deletions accumulated from previous runs to be triggered at once, flooding the `CoordinatorEventManager` queue with a large number of `DeleteReplicaResponseReceivedEvent` events. This blocks the processing of other critical coordinator events such as leader election and ISR adjustments, posing a significant risk during rolling upgrades.

**Root Cause:**

The deletion workflow in Fluss is a two-phase process:

1. **Phase 1** (synchronous): Drop the logical metadata node from ZooKeeper (e.g., `/tables/{tablePath}` for a table, or the partition metadata node for a partition) via `MetadataManager.dropTable()` / `dropPartition()`.
2. **Phase 2** (asynchronous): Send `StopReplica(deleteLocal=true, deleteRemote=true)` to all `TabletServer`s, and only after all responses succeed, delete the assignment node (e.g., `/tabletservers/tables/{tableId}`) via `MetadataManager.completeDeleteTable()` / `completeDeletePartition()`.

The "in-progress" deletion state (`ReplicaDeletionStarted`) lives **only in memory** inside the `CoordinatorContext`. If the Coordinator or a `TabletServer` restarts between Phase 1 and Phase 2, the assignment ZooKeeper node is still present. Upon the next Coordinator startup, `loadTableAssignment()` and `loadPartitionAssignment()` detect these "orphaned" assignment nodes (whose logical metadata has already been removed) and re-enqueue them for deletion via `queueTableDeletion()` / `queuePartitionDeletion()`. Since the in-memory `ReplicaState` has been reset, all of these deletions are immediately eligible and are triggered together in a single `resumeDeletions()` call during startup.

This problem is particularly severe for partitioned tables with auto-partition enabled, where a large number of expired partitions may have been deleted (Phase 1 completed) but their assignment nodes are still present in ZooKeeper.

**Example Flow:**

```
Coordinator startup
  └── TableManager.startup()
        ├── replicaStateMachine.startup()   // in-memory state is brand new
        └── resumeDeletions()               // all N pending deletions triggered at once
              └── onDeleteTable() / onDeletePartition() for each pending item
                    └── replicaStateMachine.handleStateChanges(..., ReplicaDeletionStarted)
                          └── StopReplica(delete=true) sent to TabletServers
                                └── TabletServer responds
                                      └── N * DeleteReplicaResponseReceivedEvent enqueued
                                            └── CoordinatorEventThread blocked
```

**Proposed Solution:**

Replace the direct call to `resumeDeletions()` in `TableManager.startup()` with a scheduled, periodic mechanism that dispatches a `ResumeDeletionEvent` into the `CoordinatorEventManager`. This ensures:

1. **Deletion is decoupled from startup**: The Coordinator becomes available immediately after startup without being blocked by bulk deletions.
2. **Thread safety is preserved**: The deletion logic (`resumeDeletions()`) continues to run exclusively on the `CoordinatorEventThread`, avoiding any concurrency issues with `CoordinatorContext` and `CoordinatorRequestBatch`.
3. **Interleaving with normal events**: Since deletion events are dispatched periodically and processed in the same queue as other coordinator events, they will naturally interleave with leader election, ISR adjustments, and other critical operations.

### Willingness to contribute

wattt3:
Generated-by: [Claude Opus 4.7] following [the guidelines](https://github.com/apache/fluss/blob/main/AGENTS.md)

### Purpose

Linked issue: close #2891

### Brief change log
- `CoordinatorEventProcessor` gains a scheduled executor that fires a `ResumeDeletionEvent`
- Removed the synchronous `resumeDeletions()` call from `TableManager.startup()`

### Tests
- Removed `TableManagerTest.testResumeDeletionAfterRestart`
- Added `CoordinatorEventProcessorTest.testScheduledResumeDeletionResumesPendingDeletion`

### DIFF
--- a/fluss-common/src/main/java/org/apache/fluss/config/ConfigOptions.java
+++ b/fluss-common/src/main/java/org/apache/fluss/config/ConfigOptions.java
@@ -430,6 +430,15 @@ public class ConfigOptions {
                             "The interval for cleaning up expired producer offsets "
                                     + "and orphan files in remote storage. Default is 1 hour.");
 
+    /** The interval at which the coordinator resumes in-progress table and partition deletions. */
+    public static final ConfigOption<Duration> COORDINATOR_RESUME_DELETION_INTERVAL =
+            key("coordinator.resume-deletion.interval")
+                    .durationType()
+                    .defaultValue(Duration.ofSeconds(5))
+                    .withDescription(
+                            "The interval at which the coordinator resumes in-progress table and "
+                                    + "partition deletions via a periodic event. Default is 5 seconds.");
+
     // ------------------------------------------------------------------------
     //  ConfigOptions for Tablet Server
     // ------------------------------------------------------------------------
--- a/fluss-common/src/main/java/org/apache/fluss/config/ConfigOptions.java
+++ b/fluss-common/src/main/java/org/apache/fluss/config/ConfigOptions.java
@@ -452,6 +452,15 @@ public class ConfigOptions {
                             "The interval for cleaning up expired producer offsets "
                                     + "and orphan files in remote storage. Default is 1 hour.");
 
+    /** The interval at which the coordinator resumes in-progress table and partition deletions. */
+    public static final ConfigOption<Duration> COORDINATOR_RESUME_DELETION_INTERVAL =
+            key("coordinator.resume-deletion.interval")
+                    .durationType()
+                    .defaultValue(Duration.ofHours(6))
+                    .withDescription(
+                            "The interval at which the coordinator resumes in-progress table and "
+                                    + "partition deletions via a periodic event. Default is 6 hours.");
+
     // ------------------------------------------------------------------------
     //  ConfigOptions for Tablet Server
     // ------------------------------------------------------------------------
--- a/fluss-server/src/main/java/org/apache/fluss/server/coordinator/CoordinatorEventProcessor.java
+++ b/fluss-server/src/main/java/org/apache/fluss/server/coordinator/CoordinatorEventProcessor.java
@@ -86,6 +86,7 @@ import org.apache.fluss.server.coordinator.event.NotifyLakeTableOffsetEvent;
 import org.apache.fluss.server.coordinator.event.NotifyLeaderAndIsrResponseReceivedEvent;
 import org.apache.fluss.server.coordinator.event.RebalanceEvent;
 import org.apache.fluss.server.coordinator.event.RemoveServerTagEvent;
+import org.apache.fluss.server.coordinator.event.ResumeDeletionEvent;
 import org.apache.fluss.server.coordinator.event.SchemaChangeEvent;
 import org.apache.fluss.server.coordinator.event.TableRegistrationChangeEvent;
 import org.apache.fluss.server.coordinator.event.watcher.CoordinatorChangeWatcher;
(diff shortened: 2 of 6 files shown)


wattt3:
@LiebingYu Hello, could you pleases take a look at this PR when you have time?

fluss-common/src/main/java/org/apache/fluss/config/ConfigOptions.java
```diff
@@ -430,6 +430,15 @@ public class ConfigOptions {
                             "The interval for cleaning up expired producer offsets "
                                     + "and orphan files in remote storage. Default is 1 hour.");
 
+    /** The interval at which the coordinator resumes in-progress table and partition deletions. */
+    public static final ConfigOption<Duration> COORDINATOR_RESUME_DELETION_INTERVAL =
+            key("coordinator.resume-deletion.interval")
+                    .durationType()
+                    .defaultValue(Duration.ofSeconds(5))
```
LiebingYu: 5 seconds is too aggressive. This task is designed to handle exceptional cases — under normal circumstances, deletion will proceed as expected. I think an hour-level interval would be sufficient, for example, 6 hours.

fluss-server/src/main/java/org/apache/fluss/server/coordinator/CoordinatorEventProcessor.java
```diff
@@ -302,9 +314,16 @@ public void startup() {
...
+
+        resumeDeletionScheduledExecutor.scheduleWithFixedDelay(
+                () -> coordinatorEventManager.put(new ResumeDeletionEvent()),
+                resumeDeletionIntervalMs,
+                resumeDeletionIntervalMs,
+                TimeUnit.MILLISECONDS);
     }
 
     public void shutdown() {
+        resumeDeletionScheduledExecutor.shutdownNow();
```
LiebingYu: Need graceful shutdown.

fluss-server/src/main/java/org/apache/fluss/server/coordinator/TableManager.java
```diff
@@ -70,8 +70,6 @@ public void startup() {
         LOG.info("Start up table manager.");
         replicaStateMachine.startup();
         tableBucketStateMachine.startup();
-        // try to resume one deletion after start up
-        resumeDeletions();
```
LiebingYu, line 74: I think we should keep this.

fluss-server/src/test/java/org/apache/fluss/server/coordinator/TableManagerTest.java
```diff
@@ -217,54 +216,6 @@ void testDeleteTable() throws Exception {
         assertThat(coordinatorContext.getAllReplicasForTable(tableId)).isEmpty();
     }
 
-    @Test
-    void testResumeDeletionAfterRestart() throws Exception {
```
LiebingYu, line 221: Keep this too.

LiebingYu (commented):
@wattt3 Thanks! I have some comments.

fluss-server/src/main/java/org/apache/fluss/server/coordinator/CoordinatorEventProcessor.java
wattt3: Changed to use `ExecutorUtils.gracefulShutdown()`

wattt3:
@LiebingYu  Thanks for the review!
I've addressed the comments, PTAL 🙏

PR #2786, opened 2026-03-03
[client] cancel pending log download when closing remote log downloader.

loserwang1024:
### Purpose

Linked issue: close #2783

Issue #187, opened 2024-12-14
[Feature] Arrow Log Supports Compressions 


wuchong:
### Search before asking

- [X] I searched in the [issues](https://github.com/alibaba/fluss/issues) and found nothing similar.

### Motivation

Currently, we don't have any compression for Arrow Log. But Arrow format consumes more disk space than other formats, like Avro, because Arrow is designed for memory layout and aligns on 8 bits. We see Arrow file is 30% larger than Avro file for the same data set in some tests. 

### Solution

Fortunately, Apache Arrow community already introduced buffer compressions: https://arrow.apache.org/docs/format/Columnar.html#compression. This can reduce a lot for the networking cost and disk space, but still have column pruning ability. Arrow already supports LZ4 and ZSTD compression types. 

More benchmark results should be attached for the compression benefits. 

### Anything else?

Currently, Fluss still need to support Java8, and that's why Fluss sticks on arrow-15.0 (the latest is v18.0). Arrow 15.0 Java client already introduced the `CompressionCodec` interfaces for `ArrowStreamWriter`, but doesn't natively support LZ4 and ZSTD compressions (implemented in later versions). But it should work if we learn the implementation by extending our own `CompressionCodec.Factory`: https://github.com/apache/arrow-java/commits/main/compression/src/main/java/org/apache/arrow/compression

### Willingness to contribute

- [ ] I'm willing to submit a PR!

kaori-seasons:
@wuchong Hi, does it mean that condition pushdown is supported in [LogScan](https://github.com/alibaba/fluss/blob/main/fluss-client/src/main/java/com/alibaba/fluss/client/scanner/log/LogScan.java) and the switch for condition pushdown is defined in flussapi.proto?

PR #3132, opened 2026-04-19
[server] Optimize RemoteLogFetcher with async prefetch for recovery

### LINKED ISSUE #3091: [server] RemoteLogFetcher optimize to async downloading
### Search before asking

### Motivation

Currently, `RemoteLogFetcher` fetch remote log sync which will cause the recovery process may be significantly slowed down by the download time of remote segments, especially when there are many segments to fetch. See:https://github.com/apache/fluss/pull/2855#discussion_r3091682902

### Solution

_No response_

### Anything else?

_No response_

### Willingness to contribute

Kaixuan-Duan:
**Purpose**

Linked issue: close #3091
This PR improves KV recovery performance by reducing wait time between remote log segments in `RemoteLogFetcher`.

**Brief change log**

- **Async prefetch window** in RemoteLogFetcher: added a sliding window that downloads the next N remote log segments in the background while the consumer reads the current one, replacing the previous strictly-serial download → open → read → next loop.
- **Bounded sliding-window prefetch** (default depth 4, configurable via `kv.recover.remote-log.prefetch-num`):
uses a `Future[]` ring buffer indexed by `seq % prefetchNum` to cap downloaded-but-not-consumed segments, bounding local-disk usage. A dedicated fixed-thread-pool (default 3, configurable via `kv.recover.remote-log.download-threads`) runs the actual downloads in parallel. As the consumer advances, consumed slots are freed and back-filled, overlapping network I/O with local iteration.
- **Exponential backoff retry**: failed downloads retry up to `5` times with backoff `100ms → 5s` and `0.25` jitter (ExponentialBackoff), preventing thrash on transient remote-storage failures while still surfacing persistent failures to the recovery driver.
- **Failure handling**: when a prefetched future is cancelled (e.g. by `close()`), `fetchSegmentFile()` falls back to a synchronous `downloadSegmentWithRetry` so a single cancelled task does not stall recovery. When a prefetched future fails with `ExecutionException` — meaning the async worker has already exhausted the exponential-backoff retry budget — the failure is propagated directly as `IOException` and the prefetch window is drained. No redundant sync retry round is attempted.

**Tests**

- `./mvnw -pl fluss-server -am clean install -DskipTests`
- `./mvnw -pl fluss-server spotless:apply -Dtest='RemoteLogFetcherTest' -DfailIfNoTests=false test`

**API and Format**

- `kv.recover.remote-log.prefetch-num int` (default 4) : Max remote log segments downloaded but not yet consumed during KV recovery. 1 preserves the legacy one-step-prefetch behavior.
- `kv.recover.remote-log.download-threads` (default 3) : Threads used to download remote log segments during KV recovery. Should be `≤ prefetch-num`.

No breaking changes to user-facing public APIs.

**Documentation**

Added `kv.recover.remote-log.prefetch-num` and `kv.recover.remote-log.download-threads` to `website/docs/maintenance/configuration.md`.


fresh-borzoni (commented):
@Kaixuan-Duan Thanks for the contribution. I will help to review this PR

One process point: this issue was already assigned and I was actively working on it. In that situation, please coordinate on the issue before opening an overlapping PR. Assignment is not exclusive ownership, but it is an important coordination signal, and skipping it usually leads to duplicated effort and fragmented review.

We can evaluate this PR on its merits, but for future cases please check on the issue first.

fluss-server/src/main/java/org/apache/fluss/server/kv/RemoteLogFetcher.java
```diff
@@ -333,6 +356,74 @@ private void advance() {
...
+                    return null;
+                }
+                return segment;
+            }
+            return null;
+        }
+
+        private void cancelPrefetch() {
+            if (nextDownloadedSegmentFuture != null) {
+                nextDownloadedSegmentFuture.cancel(true);
```
fresh-borzoni: cancel(true) on an already-completed future is a no-op and drops the reference to the downloaded File, which then lives in tempDir until fetcher-level close()

fluss-server/src/main/java/org/apache/fluss/server/kv/RemoteLogFetcher.java
```diff
@@ -150,17 +168,21 @@ public Iterable<LogRecordBatch> fetch(long startOffset, long localLogStartOffset
...
-            LOG.info("Cleaning up remote log recovery tmp dir: {}", tmpDir);
-            deleteDirectoryQuietly(tmpDir.toFile());
+        try {
+            // Close any active iterator to release file handles
+            if (activeIterator != null) {
+                activeIterator.close();
+                activeIterator = null;
+            }
+        } finally {
+            downloadExecutor.shutdownNow();
```
fresh-borzoni, line 200: shutdownNow() doesn't wait - if a prefetch is mid-flush, it can write to tempDir after deleteDirectoryQuietly runs. Either downloadExecutor.awaitTermination() with a short timeout before deletion, or make downloadSegment interruption-aware (most S3 SDKs don't honor Thread.isInterrupted() during socket reads, so the interrupt from shutdownNow is effectively decorative)

fluss-server/src/main/java/org/apache/fluss/server/kv/RemoteLogFetcher.java
fresh-borzoni, line 438: If fetch() is called twice, the first Iterable still wraps the now-closed iterator and iterating it re-enters advance() on a closed instance, downloading into the shared tempDir, racing with the new iterator

fluss-server/src/main/java/org/apache/fluss/server/kv/RemoteLogFetcher.java
```diff
@@ -333,6 +356,74 @@ private void advance() {
             }
         }
 
+        private File fetchSegmentFile(RemoteLogSegment segment) throws IOException {
+            if (segment.equals(prefetchedSegment) && nextDownloadedSegmentFuture != null) {
```
fresh-borzoni: This depends on RemoteLogSegment having value-based equals(), or on both references coming from the same segments list (reference equality). Works today, but safer to compare by segment id tbh.

fluss-server/src/main/java/org/apache/fluss/server/kv/RemoteLogFetcher.java
```diff
@@ -333,6 +356,74 @@ private void advance() {
...
+                try {
+                    return nextDownloadedSegmentFuture.get();
+                } catch (InterruptedException e) {
```
fresh-borzoni: Also catch CancellationException - it's unchecked (extends RuntimeException) and CompletableFuture.get() throws it on a cancelled future. Not a live bug in the current state machine (every cancelPrefetch nulls the field) but cheap defense-in-depth, especially given closed is volatile.

fluss-server/src/test/java/org/apache/fluss/server/kv/RemoteLogFetcherTest.java
```diff
@@ -28,10 +28,13 @@
 import org.junit.jupiter.api.Test;
```
fresh-borzoni, line 28: non-blocking: Two of the three new tests inject state via reflection (setPrivateField) instead of exercising a real async prefetch - they cover the branches in fetchSegmentFile, but not close-during-real-in-flight-download or the orphan-file cleanup.

Consider one integration-style test with a real slow/failing download source.

fluss-server/src/main/java/org/apache/fluss/server/kv/RemoteLogFetcher.java
```diff
@@ -333,6 +356,74 @@ private void advance() {
...
+                    throw new IOException(
+                            "Interrupted while waiting for remote log segment download: "
+                                    + segment.remoteLogSegmentId(),
+                            e);
+                } catch (ExecutionException e) {
+                    LOG.warn(
+                            "Prefetched segment {} failed, fallback to sync download.",
+                            segment.remoteLogSegmentId(),
+                            e.getCause());
+                    return downloadSegment(segment);
```
fresh-borzoni: Non-blocking: No retry on transient S3 failure - one flaky segment fails the entire recovery. In fluss-rust  we added exponential backoff (100ms -> 5s with jitter) for this.

fluss-server/src/main/java/org/apache/fluss/server/kv/RemoteLogFetcher.java
```diff
@@ -333,6 +356,74 @@ private void advance() {
...
+                } finally {
+                    nextDownloadedSegmentFuture = null;
+                    prefetchedSegment = null;
+                }
+            }
+            return downloadSegment(segment);
+        }
+
+        private void prefetchNextSegment() {
```
fresh-borzoni: Prefetch depth hardcoded to 1. If S3 p99 download time > consume time for a segment, the downloader sits idle and the optimization is half-realized. On the Rust side ([fluss-rust #187](https://github.com/apache/fluss-rust/pull/187)) we landed on configurable depth with default 4 for exactly this reason. Since it's KV depth = 1 might be fine, but it's still better to configure and reason properly

fresh-borzoni (commented):
Ty, direction is right, I left some cooments, PTAL

fluss-server/src/main/java/org/apache/fluss/server/kv/RemoteLogFetcher.java
Kaixuan-Duan: Thanks — I have updated the logic to handle this explicitly.

fluss-server/src/main/java/org/apache/fluss/server/kv/RemoteLogFetcher.java
Kaixuan-Duan: Thanks for pointing this out. I updated close() to call awaitTermination() after shutdownNow() before deleting the temp directory. If the download executor does not terminate within the timeout, the cleanup is skipped to avoid racing with an in-progress download that may still write into tempDir.

fluss-server/src/main/java/org/apache/fluss/server/kv/RemoteLogFetcher.java
Kaixuan-Duan: I updated the closed iterator to stop immediately by marking it finished, clearing the pending batch, and making `hasNext()` return `false` after close.

fluss-server/src/main/java/org/apache/fluss/server/kv/RemoteLogFetcher.java
Kaixuan-Duan: I changed the prefetch match to compare `remoteLogSegmentId()` instead of relying on `RemoteLogSegment.equals()`.

fluss-server/src/main/java/org/apache/fluss/server/kv/RemoteLogFetcher.java
Kaixuan-Duan: I added this as a defensive guard: if the prefetched future is cancelled and `get()` throws `CancellationException`.

fluss-server/src/main/java/org/apache/fluss/server/kv/RemoteLogFetcher.java
Kaixuan-Duan: Thanks. The retry design in `fluss-rust` is solid, so I followed the same idea here and added retries around synchronous segment downloads with exponential backoff and jitter, so transient remote storage failures are retried before failing the recovery.

Kaixuan-Duan:
@fresh-borzoni Thanks for the review, and sorry for not coordinating on the issue beforehand. I didn’t realize it was already being actively worked on.
The testing code and prefetch depth optimization are still progressing, and I will learn the implementation approach of [fluss-rust #187](https://github.com/apache/fluss-rust/pull/187).

fluss-server/src/test/java/org/apache/fluss/server/kv/RemoteLogFetcherTest.java
Kaixuan-Duan: Ty, Addressed comments. PTAL

fluss-server/src/main/java/org/apache/fluss/server/kv/RemoteLogFetcher.java
Kaixuan-Duan: You're right, I optimized the prefetch depth, PTAL.

Kaixuan-Duan:
@fresh-borzoni Thank you for the review. Addressed, PTAL.

fluss-server/src/main/java/org/apache/fluss/server/kv/RemoteLogFetcher.java
```diff
@@ -333,12 +564,207 @@ private void advance() {
...
+
+                // Non-blocking permit acquisition. If the window is full, back off and let the
+                // consumer release a permit via releasePermitOnce().
+                if (!prefetchPermits.tryAcquire()) {
+                    return;
+                }
+
+                final RemoteLogSegment target = segment;
+                CompletableFuture<File> future;
+                try {
```
fresh-borzoni, line 640: The async download calls downloadSegment(target), while retry only exists in the sync fallback. So a transient remote-storage failure in the prefetch worker isn't retried in the background - instead the consumer later hits the failed future and runs the retry loop synchronously. This pushes retry latency back onto the recovery thread and largely defeats the point of prefetching. The async worker should call downloadSegmentWithRetry(target).

fluss-server/src/main/java/org/apache/fluss/server/kv/RemoteLogFetcher.java
```diff
@@ -333,12 +564,207 @@ private void advance() {
...
+            while ((entry = prefetchQueue.pollFirst()) != null) {
+                CompletableFuture<File> future = entry.future;
+                if (future.isDone()) {
+                    try {
+                        cleanupUnusedPrefetchedFile(future.getNow(null));
+                    } catch (CancellationException | CompletionException ignored) {
+                        // no local file to clean up
+                    }
+                } else {
+                    future.cancel(true);
```
fresh-borzoni: CompletableFuture.cancel(true) flips the future state to CANCELLED but does not propagate an interrupt to the underlying supplyAsync task. The worker keeps running, finishes the download, and writes the file to disk after we thought we cancelled. Even when cancel(true) returns true, the task may still complete. The isDone() check added here only handles the "already-completed when drained" case. 

Switch to executor.submit() and store Future<File> in PrefetchEntry, then FutureTask.cancel(true) actually interrupts the worker

fluss-server/src/test/java/org/apache/fluss/server/kv/RemoteLogFetcherTest.java
```diff
@@ -321,4 +344,475 @@ void testFetchDownloadsFilesToTempDir() throws Exception {
...
+
+            // close() must not block for the full simulated 5s sleep — it should cancel the
+            // in-flight futures and return promptly (well under 5s on any healthy machine).
+            assertThat(closeMs).isLessThan(4_000L);
+
+            // Every in-flight future captured pre-close must end up cancelled or completed
+            // exceptionally (an interrupted Thread.sleep surfaces as InterruptedException).
+            for (CompletableFuture<File> f : inflight) {
+                assertThat(f.isDone()).isTrue();
+                assertThat(f.isCancelled() || f.isCompletedExceptionally()).isTrue();
```
fresh-borzoni: f.isCancelled() is just API state and says nothing about whether the download stopped. After the submit/Future switch, rewrite using smth like PR #2786's BlockingFileDownloader + latch pattern to assert real interruption.

fluss-server/src/main/java/org/apache/fluss/server/kv/RemoteLogFetcher.java
fresh-borzoni, line 68: @NotThreadSafe plus volatile plus async workers is a contradiction. KV recovery is single-threaded per bucket in practice, so we might wish to prefer single-threaded usage here. Also with this hybrid design I believe we have possible race condition in ArrayDeque

WDYT?

fresh-borzoni (commented):
@Kaixuan-Duan Ty for the changes, left some comments, PTAL

fluss-server/src/main/java/org/apache/fluss/server/kv/RemoteLogFetcher.java
```diff
@@ -64,6 +75,14 @@
  * }
  * }</pre>
  *
+ * <p>Segments are prefetched in a sliding window driven by a {@link Semaphore}. The window size
+ * (prefetch depth) and the number of concurrent download threads are configurable via {@code
+ * kv.recover.remote-log.prefetch-num} and {@code kv.recover.remote-log.download-threads}
+ * respectively. When a prefetched segment is consumed, its permit is released and the fetcher
+ * immediately tries to start the next download, giving the window the behavior of Rust's RAII
+ * {@code PrefetchPermit} + recycle-notify pattern in <a
+ * href="https://github.com/apache/fluss-rust/pull/187">fluss-rust PR#187</a>.
```
fresh-borzoni: I wouldn't reference fluss-rust here - cross-repo Javadoc links go stale(especially bc fluss-rust is planned to be ported to this repo), and the description should stand on its own.                                                                                                                           
                                                      
Also: this pattern shines in async Rust where awaiting tasks need explicit notification on permit release (and RAII gives you deterministic permit drop). Java's sync consumer makes the permit-recycle machinery redundant - a Future<File>[] indexed by seq % depth is fine as well.

Not a blocker, just worth considering.

fluss-server/src/main/java/org/apache/fluss/server/kv/RemoteLogFetcher.java
Kaixuan-Duan: The async worker now calls downloadSegmentWithRetry(target), so transient failures are retried in the worker thread instead of falling back to the recovery thread.

fluss-server/src/main/java/org/apache/fluss/server/kv/RemoteLogFetcher.java
Kaixuan-Duan: Replaced supplyAsync with downloadExecutor.submit(Callable) and store the Future<File> in the slot array. cancel(true) on the FutureTask now actually interrupts the worker.

fluss-server/src/test/java/org/apache/fluss/server/kv/RemoteLogFetcherTest.java
Kaixuan-Duan: Done. Added testCloseInterruptsOngoingDownload using PR #2786's barrier+latch pattern. The release latch is never counted down by the test, so the only way finished.countDown() can fire is if Thread.interrupt() reached the worker inside the barrier. assertThat(finished.await(10s)).isTrue() is the actual interruption proof — isCancelled() is kept only as a secondary API-state check.

fluss-server/src/main/java/org/apache/fluss/server/kv/RemoteLogFetcher.java
Kaixuan-Duan: Agreed. Removed volatile/synchronized/Concurrent*/Semaphore/ArrayDeque. The consumer is now strictly single-threaded; the only cross-thread handoff is the per-slot Future, whose visibility is guaranteed by Future.get().

fluss-server/src/main/java/org/apache/fluss/server/kv/RemoteLogFetcher.java
Kaixuan-Duan: You're right. 
Removed the fluss-rust reference and RAII wording from Javadoc. Window is now a fixed Future<File>[] prefetchSlots indexed by seq % prefetchNum.

Kaixuan-Duan:
@fresh-borzoni Thank you for the review. I have already modified it. PTAL

fluss-server/src/test/java/org/apache/fluss/server/kv/RemoteLogFetcherTest.java
```diff
@@ -321,4 +349,519 @@ void testFetchDownloadsFilesToTempDir() throws Exception {
             }
         }
     }
+
+    @Test
+    void testPrefetchFailureFallsBackToSyncDownload() throws Exception {
```
fresh-borzoni: Shall we use fetchLogDataBarrier hook as well here?

fluss-server/src/test/java/org/apache/fluss/server/kv/RemoteLogFetcherTest.java
```diff
@@ -321,4 +349,519 @@ void testFetchDownloadsFilesToTempDir() throws Exception {
...
+            while (iterator.hasNext()) {
+                iterator.next();
+                count++;
+            }
+            assertThat(count).isGreaterThan(1);
+        }
+    }
+
+    @Test
+    void testCloseCancelsPendingPrefetchFuture() throws Exception {
```
fresh-borzoni, line 387: ditto

fluss-server/src/test/java/org/apache/fluss/server/kv/RemoteLogFetcherTest.java
```diff
@@ -321,4 +349,519 @@ void testFetchDownloadsFilesToTempDir() throws Exception {
...
+                assertThat(batch.baseLogOffset()).isGreaterThanOrEqualTo(prevNextOffset);
+                prevNextOffset = batch.nextLogOffset();
+                batchCount++;
+            }
+            assertThat(batchCount).isGreaterThan(0);
+        }
+    }
+
+    @Test
+    void testCloseCancelsAllPendingPrefetchFutures() throws Exception {
```
fresh-borzoni, line 606: ditto

fluss-server/src/main/java/org/apache/fluss/server/kv/RemoteLogFetcher.java
```diff
@@ -333,12 +573,219 @@ private void advance() {
...
+                            segment.remoteLogSegmentId(),
+                            e.getCause());
+                    return downloadSegmentWithRetry(segment);
+                } finally {
+                    // Window slot released — try to submit the next prefetch right away.
+                    fillPrefetchWindow();
+                }
+            }
+
+            // Defensive fallback: this branch is logically unreachable in production.
```
fresh-borzoni: Once remaining tests migrate to the fetchLogDataBarrier pattern, this defensive branch and injectPrefetchEntryForTest can both be deleted:

fluss-server/src/main/java/org/apache/fluss/server/kv/RemoteLogFetcher.java
```diff
@@ -333,12 +573,219 @@ private void advance() {
...
+                    } catch (InterruptedException e) {
+                        // A completed future never actually blocks here, so an interrupt on this
+                        // thread came from elsewhere. Restore the flag and bail out — close()'s
+                        // follow-up shutdownNow()/awaitTermination path will reclaim whatever's
+                        // left in the ring.
+                        Thread.currentThread().interrupt();
+                        return;
+                    }
+                } else {
+                    future.cancel(true);
```
fresh-borzoni: There is a race: future completes between our isDone() check and cancel(true).
Cancel returns false, but we ignore the return - the file's already on disk and we never call get() to retrieve it.

Fix: if cancel returns false, we need to cleanupUnusedPrefetchedFile(future.get()) almost the similar pattern as above

fluss-server/src/main/java/org/apache/fluss/server/kv/RemoteLogFetcher.java
```diff
@@ -189,16 +317,75 @@ private File downloadSegment(RemoteLogSegment segment) throws IOException {
...
+        boolean success = false;
         try (InputStream inputStream = remoteLogStorage.fetchLogData(segment);
                 OutputStream outputStream = Files.newOutputStream(localFile.toPath())) {
             IOUtils.copyBytes(inputStream, outputStream, false);
+            success = true;
         } catch (RemoteStorageException e) {
             throw new IOException(
                     "Failed to download remote log segment: " + segment.remoteLogSegmentId(), e);
+        } finally {
+            if (!success) {
```
fresh-borzoni: let's tighten:  `|| Thread.currentThread().isInterrupted()` to prevent that most network/file reads don't honor interrupts mid-syscall, so worth to check it

fresh-borzoni (commented):
@Kaixuan-Duan Ty for the fixes, overall LGTM, left a couple of comments, PTAL

@swuferhong Can you take a look as well?

Kaixuan-Duan:
@fresh-borzoni Thank you for the review. Addressed, PTAL

fluss-server/src/main/java/org/apache/fluss/server/kv/RemoteLogFetcher.java
Kaixuan-Duan: Done — `added || Thread.currentThread().isInterrupted()` to the `finally`. Drops partial/interrupted temp files that `IOUtils.copyBytes` would otherwise return as `success=true`.

fluss-server/src/main/java/org/apache/fluss/server/kv/RemoteLogFetcher.java
Kaixuan-Duan: Done — extracted a `cleanupCompletedFuture(Future<File>)` helper, used by both the `isDone()` branch and the `!cancel(true)` fallback. No more silent leaks when the task completes between the two checks.

fluss-server/src/test/java/org/apache/fluss/server/kv/RemoteLogFetcherTest.java
Kaixuan-Duan: Fixed, thanks for the feedback.

fluss-server/src/main/java/org/apache/fluss/server/kv/RemoteLogFetcher.java
Kaixuan-Duan: Fixed, thanks for the feedback.

Kaixuan-Duan:
@fresh-borzoni @swuferhong Hi, could you please help review this PR when you have time?
Thanks!

fluss-server/src/main/java/org/apache/fluss/server/kv/RemoteLogFetcher.java
```diff
@@ -333,12 +555,202 @@ private void advance() {
...
+            } catch (IOException cleanupException) {
+                LOG.warn(
+                        "Failed to cleanup unused prefetched segment file {} for table bucket {}.",
+                        prefetchedFile,
+                        tableBucket,
+                        cleanupException);
+            }
+        }
+
         private void closeCurrentFileLogRecords() {
```
fresh-borzoni, line 725: The new config bounds “downloaded but not yet consumed” segments, but consumed segment files are not deleted when we move to the next segment. closeCurrentFileLogRecords() closes FileLogRecords, but the underlying .log file remains in tempDir until fetcher close.

Could we track the currently opened local segment file and delete it after closing FileLogRecords? Something like currentLocalFile, set after fetchSegmentFile(), then delete it in closeCurrentFileLogRecords().

fluss-server/src/main/java/org/apache/fluss/server/kv/RemoteLogFetcher.java
```diff
@@ -136,31 +188,60 @@ public Iterable<LogRecordBatch> fetch(long startOffset, long localLogStartOffset
...
+                        downloadExecutor.awaitTermination(1, java.util.concurrent.TimeUnit.SECONDS);
+            } catch (InterruptedException e) {
+                Thread.currentThread().interrupt();
+                LOG.warn(
+                        "Interrupted while waiting for remote log fetcher download executor termination for table bucket {}.",
+                        tableBucket,
+                        e);
+            }
+
+            if (terminated) {
```
fresh-borzoni: This cleanup path seems risky in two ways.

First, if awaitTermination() times out, we skip cleanup entirely, so one uncooperative download can leave the recovery directory behind.

Second, tempDir.getParent() is the shared tmp/ directory under the log dir, not just this fetcher's recovery directory. That is broader than what this RemoteLogFetcher owns.

Could we delete tempDir itself, not its parent, and attempt that best-effort regardless of whether the executor terminated? If deletion fails because a file is still open, logging that is fine, but I don't think we should skip the cleanup attempt.

fluss-server/src/main/java/org/apache/fluss/server/kv/RemoteLogFetcher.java
```diff
@@ -333,12 +555,202 @@ private void advance() {
...
+         * long as more segments remain.
+         */
+        private File fetchSegmentFile(RemoteLogSegment segment) throws IOException {
+            // Invariant: fillPrefetchWindow() and advance() walk the same pre-filtered
+            // `segments` list in strict order with identical skip rules; consumption is
+            // single-threaded FIFO; prefetchNum >= 1 (Math.max(1, ...) in the ctor). So
+            // the ring head is *always* the segment advance() is asking for.
+            int headSlot = nextConsumeIndex % prefetchNum;
+            RemoteLogSegment headSegment = prefetchSegments[headSlot];
+            assert isSameSegmentId(segment, headSegment)
```
fresh-borzoni: This invariant is important enough that I don't think assert is the right guard. Assertions are normally disabled in production.
Could we make this an explicit runtime check, drain the prefetch window, and throw IOException with the requested/head segment ids?

fluss-server/src/main/java/org/apache/fluss/server/kv/RemoteLogFetcher.java
```diff
@@ -189,16 +294,80 @@ private File downloadSegment(RemoteLogSegment segment) throws IOException {
...
         } catch (RemoteStorageException e) {
             throw new IOException(
                     "Failed to download remote log segment: " + segment.remoteLogSegmentId(), e);
+        } finally {
+            // Most remote/file InputStreams don't honor Thread.interrupt() mid-read: the
+            // call simply returns a normally-completed copy and we'd fall through with
+            // success=true while the worker has been interrupted by close()/cancel(true).
+            // Treat "interrupt observed during the copy" as a failure for cleanup purposes
+            // so we don't leave a stale segment file behind in tempDir.
+            if (!success || Thread.currentThread().isInterrupted()) {
```
fresh-borzoni, line 293: In downloadSegment(), if the copy succeeds but the thread has been interrupted, the finally block deletes localFile but the method can still return that File object.

Usually Future.cancel(true) masks this because the Future is cancelled, but interrupt races can still make this path observable. Could we throw after deleting on the interrupted path rather than returning a file that may no longer exist?

fresh-borzoni (commented):
@Kaixuan-Duan Sorry, I missed this, Ty for the changes, left some comments, PTAL

fresh-borzoni:
@Kaixuan-Duan Can you rebase as well, pls?

fluss-server/src/main/java/org/apache/fluss/server/kv/RemoteLogFetcher.java
```diff
@@ -333,12 +555,202 @@ private void advance() {
...
+                        "Interrupted while waiting for remote log segment download: "
+                                + segment.remoteLogSegmentId(),
+                        e);
+            } catch (CancellationException e) {
+                LOG.warn(
+                        "Prefetched segment {} was cancelled, fallback to sync download.",
+                        segment.remoteLogSegmentId(),
+                        e);
+                return downloadSegmentWithRetry(segment);
+            } catch (ExecutionException e) {
```
fresh-borzoni, line 590: The sync fallback retry in fetchSegmentFile is now a leftover from when retries lived only in the sync path. Could we drop the catch retry call and just propagate the failure? 
The async path already exhausts the retry budget

fluss-server/src/main/java/org/apache/fluss/server/kv/RemoteLogFetcher.java
Kaixuan-Duan: Thanks for the suggestion. Added currentLocalFile field to track the currently opened segment file; closeCurrentFileLogRecords() now deletes it after closing FileLogRecords.

fluss-server/src/main/java/org/apache/fluss/server/kv/RemoteLogFetcher.java
Kaixuan-Duan: Good catch. Removed tempDir.getParent() deletion; now close() always attempts best-effort cleanup of its own UUID subdirectory regardless of executor termination, with a warning log if files are still in use.

fluss-server/src/main/java/org/apache/fluss/server/kv/RemoteLogFetcher.java
Kaixuan-Duan: Agreed. Fixed.

fluss-server/src/main/java/org/apache/fluss/server/kv/RemoteLogFetcher.java
Kaixuan-Duan: Fixed. Added || Thread.currentThread().isInterrupted() to the finally block: if the copy succeeds but the thread was interrupted, the file is deleted and IOException is thrown instead of returning a dangling reference.

fluss-server/src/main/java/org/apache/fluss/server/kv/RemoteLogFetcher.java
Kaixuan-Duan: You're right — the async worker already exhausts retries via downloadSegmentWithRetry(). Removed the redundant sync retry in catch (ExecutionException); now propagates directly as IOException.

Kaixuan-Duan:
@fresh-borzoni Thank you for the review! All 5 comments addressed.
Also rebased onto latest main. PTAL

fresh-borzoni (approved):
@Kaixuan-Duan Ty for the changes, LGTM.
Can you please, update docs as well, as we introduced new config options?

Kaixuan-Duan:
@fresh-borzoni Thanks for the review! I've updated the docs. PTAL

fresh-borzoni:
@Kaixuan-Duan Thank you 👍 

@loserwang1024 @swuferhong Can you take a quick look when you have time?

Kaixuan-Duan:
@loserwang1024 @swuferhong Hi, could you please help view this?

fresh-borzoni:
@Kaixuan-Duan Ty for the PR and sorry for sitting on it.

Happy with this now - I'll leave it open a few days for @swuferhong / @loserwang1024 in case they have thoughts, then merge.

fresh-borzoni:
Merging

PR #3135, opened 2026-04-20
[client] Fix early termination in SortMergeReader on changelog deletes

### LINKED ISSUE #3134: [client] SortMergeReader stops after deleted snapshot rows
### Describe the bug
`SortMergeReader.SnapshotMergedRowIteratorWrapper#hasNext()` can terminate iteration early when a snapshot row is deleted by the changelog.

When `sortMergeWithChangeLog()` returns `SortMergeRows.EMPTY`, `hasNext()` leaves `currentMergedRows` as `null` and immediately returns `false`. The caller then treats the snapshot iterator as exhausted, so all remaining snapshot rows are skipped.

### To Reproduce
A minimal reproduction is:
1. Build a snapshot iterator with ordered primary keys, for example `0..9`.
2. Build a changelog iterator that contains `DELETE` records for keys that exist in the snapshot, for example `2`, `5`, and `8`.
3. Create `SortMergeReader.readBatch()` with the snapshot iterator and the changelog iterator.
4. Iterate through the returned reader.

Expected keys: `0,1,3,4,6,7,9`

Actual behavior before the fix: iteration stops at the first deleted snapshot key and later snapshot rows are never returned.

### Expected behavior
Deleted snapshot rows should be skipped, but iteration should continue until the snapshot iterator is actually exhausted.

### Root cause
The bug is in `SnapshotMergedRowIteratorWrapper#hasNext()` in `fluss-client/src/main/java/org/apache/fluss/client/table/scanner/SortMergeReader.java`.

The method only tries to fetch one `SortMergeRows` result. If that result is empty, it does not advance to the next snapshot row and returns `false` immediately.

### Proposed fix
Change `hasNext()` to keep advancing until it either:
- finds a non-empty merged result and returns the next row; or
- truly exhausts the snapshot iterator.

This means empty merge results produced by changelog deletes are skipped instead of terminating the whole scan.

### Additional context
I prepared a companion fix and regression test in `SortMergeReaderTest` that covers the delete scenario.

luoyuxia:
### Purpose

Linked issue: close #3134

Fix `SortMergeReader` returning early when changelog deletes match snapshot rows during sort-merge reading.

### Brief change log

- update `SnapshotMergedRowIteratorWrapper#hasNext()` to skip empty merge results instead of treating them as end-of-input
- continue advancing until a non-empty merged row is available or the snapshot iterator is actually exhausted
- add a regression test that covers snapshot rows deleted by changelog records and verifies later snapshot rows are still returned

### Tests

- `./mvnw -pl fluss-client spotless:check -DskipTests -Dmaven.test.skip=true`
- `./mvnw -pl fluss-client checkstyle:check -DskipTests -Dmaven.test.skip=true`
- attempted `./mvnw test -pl fluss-client -Dtest=SortMergeReaderTest -Dcheckstyle.skip=true -Dpmd.skip=true -Dspotless.check.skip=true -Dmaven.javadoc.skip=true`, but module test compilation is currently blocked by pre-existing errors in `fluss-client/src/test/java/org/apache/fluss/client/table/LakeEnableTableITCase.java` referencing `DATALAKE_ENABLED`

### API and Format

### Documentation

### DIFF
--- a/fluss-client/src/main/java/org/apache/fluss/client/table/scanner/SortMergeReader.java
+++ b/fluss-client/src/main/java/org/apache/fluss/client/table/scanner/SortMergeReader.java
@@ -321,23 +321,26 @@ public class SortMergeReader {
                 return true;
             }
             try {
-                // if currentMergedRows is null, we need to get the next mergedRows
-                if (currentMergedRows == null) {
-                    SortMergeRows sortMergeRows =
-                            currentLakeSnapshotRecords.hasNext()
-                                    ? currentLakeSnapshotRecords.next()
-                                    : null;
-                    //  next mergedRows is not null and is not empty, set the currentMergedRows
-                    if (sortMergeRows != null && !sortMergeRows.mergedRows.isEmpty()) {
-                        currentMergedRows = sortMergeRows.mergedRows.iterator();
+                while (returnedRow == null) {
+                    if (currentMergedRows == null) {
+                        if (!currentLakeSnapshotRecords.hasNext()) {
+                            return false;
+                        }
+
+                        SortMergeRows sortMergeRows = currentLakeSnapshotRecords.next();
+                        if (!sortMergeRows.mergedRows.isEmpty()) {
+                            currentMergedRows = sortMergeRows.mergedRows.iterator();
+                        }
+                        continue;
+                    }
+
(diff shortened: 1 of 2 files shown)


beryllw:
LGTM!

PR #3140, opened 2026-04-20
[FIP-43][client] Support log scanner and writer for multiple tables

### LINKED ISSUE #3139: Support log scanner for multiple tables.
### Search before asking

### Motivation

Currently, Fluss's LogScanner is designed for single table. However, Kafka's Consumer supports multiple topics. 
I am working at fluss yaml source, which also need this.

### Solution

_No response_

### Anything else?

_No response_

### Willingness to contribute

loserwang1024:
### Purpose

Linked issue: close #3139

### DIFF
--- a/fluss-client/src/main/java/org/apache/fluss/client/table/scanner/MultiTableScan.java
+++ b/fluss-client/src/main/java/org/apache/fluss/client/table/scanner/MultiTableScan.java
@@ -1,0 +1,41 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.fluss.client.table.scanner;
+
+import org.apache.fluss.annotation.PublicEvolving;
+import org.apache.fluss.client.table.scanner.log.MultiTableLogScanner;
+
+/**
+ * Builder for a multi-table scanner.
+ *
+ * <p>Phase 1 surface intentionally exposes no per-table builder configuration: tables are
+ * registered dynamically on the created {@link MultiTableLogScanner} when {@code subscribe(...)} is
+ * first called for a {@code TablePath}. Future phases may add {@code withTable(...)} overloads to
+ * pre-declare tables together with projection / filter pushdown.
+ *
+ * @since 0.7
+ */
+@PublicEvolving
+public interface MultiTableScan {
+
+    /**
+     * Creates a {@link MultiTableLogScanner} to continuously read log data from multiple tables.
+     * Tables are registered dynamically on first {@code subscribe(...)} call.
+     */
+    MultiTableLogScanner createLogScanner();
+}
--- a/fluss-client/src/main/java/org/apache/fluss/client/table/scanner/log/AbstractLogScanner.java
+++ b/fluss-client/src/main/java/org/apache/fluss/client/table/scanner/log/AbstractLogScanner.java
@@ -1,0 +1,207 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.fluss.client.table.scanner.log;
+
+import org.apache.fluss.annotation.Internal;
+import org.apache.fluss.client.metrics.ScannerMetricGroup;
+import org.apache.fluss.exception.WakeupException;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.io.IOException;
+import java.time.Duration;
+import java.util.ConcurrentModificationException;
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.concurrent.atomic.AtomicLong;
+
+/**
+ * Base class that centralizes the shared poll loop, lightweight single-thread guard, wakeup and
+ * close lifecycle for log scanners. Concrete scanners supply the result type {@code R} and an
+ * enrichment step from raw {@link ScanRecords} to that type.
+ *
+ * <p>This class is NOT thread-safe. The lightweight guard only detects accidental concurrent use
+ * and throws {@link ConcurrentModificationException}; it does not serialize callers.
+ */
+@Internal
+abstract class AbstractLogScanner<R> implements AutoCloseable {
+
+    private static final Logger LOG = LoggerFactory.getLogger(AbstractLogScanner.class);
+
+    private static final long NO_CURRENT_THREAD = -1L;
+
+    protected final String name;
+    protected final LogScannerStatus logScannerStatus;
+    protected final LogFetcher logFetcher;
+    protected final ScannerMetricGroup scannerMetricGroup;
+
+    private volatile boolean closed = false;
+
+    // currentThread holds the threadId of the current thread accessing the scanner
+    // and is used to prevent multithreaded access.
+    private final AtomicLong currentThread = new AtomicLong(NO_CURRENT_THREAD);
+    // refCount is used to allow reentrant access by the thread that has acquired currentThread.
+    private final AtomicInteger refCount = new AtomicInteger(0);
+
+    protected AbstractLogScanner(
+            String name,
+            LogScannerStatus logScannerStatus,
+            LogFetcher logFetcher,
+            ScannerMetricGroup scannerMetricGroup) {
+        this.name = name;
+        this.logScannerStatus = logScannerStatus;
+        this.logFetcher = logFetcher;
+        this.scannerMetricGroup = scannerMetricGroup;
+    }
+
+    /**
+     * Template poll loop shared by all concrete scanners.
+     *
+     * <ul>
+     *   <li>If {@link #pollForFetches()} returns a batch with scanner progress, {@link
+     *       #toResult(ScanRecords)} is invoked to produce the concrete result type.
+     *   <li>If the batch has no progress, we wait on {@link LogFetcher#awaitNotEmpty(long)} until
+     *       either data arrives, the timeout elapses, or {@link #wakeup()} is called.
+     *   <li>Any empty/timeout/wakeup path returns {@link #emptyResult()}.
+     * </ul>
+     */
+    public final R poll(Duration timeout) {
+        acquireAndEnsureOpen();
+        try {
+            if (!logScannerStatus.prepareToPoll()) {
+                throw new IllegalStateException(notSubscribedMessage());
+            }
+
+            scannerMetricGroup.recordPollStart(System.currentTimeMillis());
+            long timeoutNanos = timeout.toNanos();
+            long startNanos = System.nanoTime();
+            do {
+                ScanRecords scanRecords = pollForFetches();
+                if (!scanRecords.hasProgress()) {
+                    try {
+                        if (!logFetcher.awaitNotEmpty(startNanos + timeoutNanos)) {
+                            // no data in buffer within the timeout
+                            return emptyResult();
+                        }
+                    } catch (WakeupException e) {
+                        // wakeup() was called, return empty
+                        return emptyResult();
+                    }
+                } else {
+                    // before returning the fetched records, send off the next round of fetches
+                    // to enable pipelining while the user is handling the current batch.
+                    logFetcher.sendFetches();
+                    return toResult(scanRecords);
+                }
+            } while (System.nanoTime() - startNanos < timeoutNanos);
+
+            return emptyResult();
+        } finally {
+            release();
+            scannerMetricGroup.recordPollEnd(System.currentTimeMillis());
+        }
+    }
+
+    private ScanRecords pollForFetches() {
+        ScanRecords scanRecords = logFetcher.collectFetch();
+        if (scanRecords.hasProgress()) {
+            return scanRecords;
+        }
+
+        // send any new fetches (won't resend pending fetches).
+        logFetcher.sendFetches();
+
+        return logFetcher.collectFetch();
+    }
+
+    /** The empty sentinel for the concrete result type. */
+    protected abstract R emptyResult();
+
+    /** Convert a non-empty batch of {@link ScanRecords} into the concrete result type. */
+    protected abstract R toResult(ScanRecords scanRecords);
+
+    /** Error message for {@link IllegalStateException} when no buckets are subscribed. */
+    protected abstract String notSubscribedMessage();
+
+    public final void wakeup() {
+        logFetcher.wakeup();
+    }
+
+    /**
+     * Acquire the light lock and ensure that the scanner hasn't been closed.
+     *
+     * @throws IllegalStateException if the scanner has been closed
+     */
+    protected final void acquireAndEnsureOpen() {
+        acquire();
+        if (closed) {
+            release();
+            throw new IllegalStateException("This scanner has already been closed.");
+        }
+    }
+
+    /**
+     * Acquire the light lock protecting this scanner from multithreaded access. Instead of blocking
+     * when the lock is not available, we throw an exception (multithreaded usage is not supported).
+     *
+     * @throws ConcurrentModificationException if another thread already has the lock
+     */
+    private void acquire() {
+        final Thread thread = Thread.currentThread();
+        final long threadId = thread.getId();
+        if (threadId != currentThread.get()
+                && !currentThread.compareAndSet(NO_CURRENT_THREAD, threadId)) {
+            throw new ConcurrentModificationException(
+                    "Scanner is not safe for multithreaded access. "
+                            + "currentThread(name: "
+                            + thread.getName()
+                            + ", id: "
+                            + threadId
+                            + ")"
+                            + " otherThread(id: "
+                            + currentThread.get()
+                            + ")");
+        }
+        refCount.incrementAndGet();
+    }
+
+    /** Release the light lock protecting the scanner from multithreaded access. */
+    protected final void release() {
+        if (refCount.decrementAndGet() == 0) {
+            currentThread.set(NO_CURRENT_THREAD);
+        }
+    }
+
+    @Override
+    public final void close() {
+        acquire();
+        try {
+            if (!closed) {
+                LOG.trace("Closing log scanner: {}", name);
+                scannerMetricGroup.close();
+                logFetcher.close();
+            }
+            LOG.debug("Log scanner: {} has been closed", name);
+        } catch (IOException e) {
+            throw new RuntimeException("Failed to close log scanner: " + name, e);
+        } finally {
+            closed = true;
+            release();
+        }
+    }
+}
--- a/fluss-client/src/main/java/org/apache/fluss/client/table/scanner/log/LogFetcher.java
+++ b/fluss-client/src/main/java/org/apache/fluss/client/table/scanner/log/LogFetcher.java
@@ -75,7 +75,6 @@ import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
-import java.util.stream.Collectors;
 
 import static org.apache.fluss.rpc.util.CommonRpcMessageUtils.getFetchLogResultForBucket;
 import static org.apache.fluss.utils.Preconditions.checkNotNull;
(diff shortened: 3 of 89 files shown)


wuchong:
Thanks @loserwang1024. Supporting multi-table reads is a valuable enhancement. Currently, we can already read from multiple tables by creating separate `LogScanner` instances via `connection.getTable(tablePath).newScan(..).createLogScanner(..).`

I understand that the primary goal of this PR is to consolidate I/O within the LogFetcher, and maintaining individual LogScanners per table is not a problem. This optimization mirrors our approach on the writer side, where writes for multiple tables are merged into a unified sender instance to aggregate I/O. We can apply a similar strategy here by enabling multi-table support in the LogFetcher and sharing it at the Connection level.

This approach allows us to keep the user-facing API unchanged, leveraging the existing hierarchy of Connection -> Table -> LogScanner without introducing a new multi-table scanner abstraction. It also ensures consistency between reader and writer operations for both single-table and multi-table scenarios. 

What do you think?

loserwang1024:
>  leveraging the existing hierarchy of Connection -> Table -> LogScanner without introducing a new multi-table scanner abstraction. It also ensures consistency between reader and writer operations for both single-table and multi-table scenarios.

this approach introduces two critical issues:
* Blocking Latency vs. Asynchronous Writes: A single reader would need to maintain multiple LogScanner instances and poll them sequentially (e.g., calling scanner.poll(POLL_TIMEOUT) in a loop). If the tables at the beginning of the queue have low data volume, the poll operation will block for the entire timeout period, causing significant latency before subsequent tables are processed.
Note: The reason this pattern works on the writer side is that writes are asynchronous; one write operation does not block others. In contrast, the synchronous nature of reading makes sequential polling inefficient.

* Memory  Overhead: Requiring each reader to manage multiple LogScanner instances creates substantial resource overhead.  Each LogScanner must independently maintain its own memory buffers for parsing Arrow records. Consequently, the aggregate memory footprint becomes significantly larger than necessary compared to a unified approach.

fluss-client/src/main/java/org/apache/fluss/client/table/scanner/log/AbstractLogScanner.java
```diff
@@ -0,0 +1,207 @@
...
+            if (!logScannerStatus.prepareToPoll()) {
+                throw new IllegalStateException(notSubscribedMessage());
+            }
+
+            scannerMetricGroup.recordPollStart(System.currentTimeMillis());
+            long timeoutNanos = timeout.toNanos();
+            long startNanos = System.nanoTime();
+            do {
+                ScanRecords scanRecords = pollForFetches();
+                if (scanRecords.isEmpty()) {
```
fxbing: Should progress-only results be returned here? `ScanRecords` may contain buckets with empty record lists but advanced offsets; `isEmpty()` drops that distinction.

fluss-common/src/main/java/org/apache/fluss/record/LogRecordReadContext.java
```diff
@@ -487,4 +502,45 @@ private static FieldGetter[] buildProjectedFieldGetters(RowType rowType, int[] s
...
+        final RowType dataRowType;
+
+        /** The field indices to select from the data row type. */
+        final int[] selectedFields;
+
+        /** Pre-built field getters corresponding to {@link #selectedFields}. */
+        final FieldGetter[] fieldGetters;
+
+        /** Whether to project records from other schemas to this target schema. */
+        final boolean readAsTargetSchema;
```
fxbing: Is `readAsTargetSchema` still needed? It is written but I could not find a read.

fluss-client/src/main/java/org/apache/fluss/client/table/scanner/log/MultiTableLogScannerImpl.java
```diff
@@ -0,0 +1,356 @@
...
+        this(
+                scannerName,
+                metadataUpdater,
+                conf,
+                remoteFileDownloader,
+                admin,
+                new LogScannerStatus(),
+                // Use a synthetic TablePath for scanner-level metrics scoping.
+                new ScannerMetricGroup(
+                        clientMetricGroup, TablePath.of("multi-table", scannerName)));
```
fxbing, line 109: Should we add per-table metrics for different subscribed tables here?

fluss-client/src/main/java/org/apache/fluss/client/table/scanner/log/MultiTableLogScannerImpl.java
```diff
@@ -0,0 +1,356 @@
...
+     * table (picking up a new tableId after a drop+recreate). Must be called under the
+     * acquire-lock.
+     */
+    private void untrackSubscription(TablePath tablePath, TableBucket tableBucket) {
+        Set<TableBucket> buckets = subscribedBucketsByPath.get(tablePath);
+        if (buckets == null) {
+            return;
+        }
+        buckets.remove(tableBucket);
+        if (buckets.isEmpty()) {
```
fxbing, line 303: Should this also unregister/close the corresponding `LogFetcher` table context when the last bucket is removed?

fluss-client/src/main/java/org/apache/fluss/client/table/scanner/log/MultiTableLogScannerImpl.java
```diff
@@ -0,0 +1,356 @@
...
+    @Override
+    protected String notSubscribedMessage() {
+        return "MultiTableLogScanner is not subscribed to any buckets.";
+    }
+
+    @Override
+    public void subscribe(TablePath tablePath, int bucket, long offset) {
+        acquireAndEnsureOpen();
+        try {
+            TableInfo tableInfo = registerIfAbsent(tablePath);
```
fxbing: Should the partitioned-table check happen before `registerIfAbsent()`? A failed wrong-overload subscribe seems to leave registration state behind.

fluss-common/src/main/java/org/apache/fluss/record/LogRecordBatch.java
```diff
@@ -186,6 +186,9 @@ default ArrowBatchData loadArrowBatch(ReadContext context) {
     /** The read context of a {@link LogRecordBatch} to read records. */
     interface ReadContext {
 
+        /** Gets the table id, or -1 if not available. */
+        long getTableId();
```
fxbing: Is `ReadContext#getTableId()` currently unused?

fxbing (commented):
Thanks for the work on this. I left a few small questions while reading through the scanner/fetcher paths; most are about expected semantics and cleanup boundaries.

fluss-client/src/main/java/org/apache/fluss/client/table/scanner/log/AbstractLogScanner.java
loserwang1024: you're right. when I do this POC, https://github.com/apache/fluss/pull/3242 is not merged.

fluss-client/src/main/java/org/apache/fluss/client/table/scanner/log/MultiTableLogScannerImpl.java
loserwang1024: I want to it in another PR to support bucket level's metric, thus MutilTable and Single table can share same metrics.

fluss-common/src/main/java/org/apache/fluss/record/LogRecordBatch.java
loserwang1024: remove it

fluss-client/src/main/java/org/apache/fluss/client/table/scanner/log/MultiTableLogScannerImpl.java
loserwang1024: done

loserwang1024:
@leonardBang , would you like to help review this?

fluss-client/src/main/java/org/apache/fluss/client/table/writer/MultiTableWriterImpl.java
```diff
@@ -0,0 +1,533 @@
...
+     * </ul>
+     */
+    private TableWriteState resolveStateAndRefreshSchemaIfChanged(
+            TablePath tablePath, int recordSchemaId) {
+        TableWriteState latest = resolveLatestState(tablePath, false);
+        if (recordSchemaId == latest.tableInfo.getSchemaId()) {
+            return latest;
+        }
+
+        writerClient.flush();
```
leonardBang: Would it make sense to enforce the schema boundary in the shared accumulator rather than relying on this flush?

Each `MultiTableWriterImpl` keeps its own schema cache, while writers created from the same connection share the same `WriterClient` and `RecordAccumulator`. Since `WriterClient.flush()` allows other threads to continue sending, one writer may still have a schema-1 log batch buffered when another writer first resolves schema 2 and appends to the same table/bucket. The KV batch checks `schemaId`, but the log batches currently do not, so the resulting batch may carry the old schema id while containing a row encoded with the new schema.

Could `RecordAccumulator` or `WriteBatch.tryAppend()` roll over when the table id, schema id, or write format changes? It may also be helpful to add a test with two writers sharing one connection, leaving the old-schema write outstanding before sending a new-schema record from the second writer.

fluss-client/src/main/java/org/apache/fluss/client/table/writer/MultiTableWriterImpl.java
```diff
@@ -0,0 +1,533 @@
...
+                            + ": record schemaId="
+                            + recordSchemaId
+                            + " is ahead of the current table schemaId="
+                            + latestSchemaId
+                            + " (after metadata refresh).");
+        }
+        if (recordSchemaId == latestSchemaId) {
+            return latest;
+        }
+        // recordSchemaId < latestSchemaId: encode against the historical schema the row was
```
leonardBang: Could we avoid taking the connection-wide flush and metadata-refresh path for every historical-schema record?

`latestSchemaIdByPath` continues to point to the table's latest schema after the historical state has been cached. Therefore, consecutive records carrying the same historical schema id still enter this branch repeatedly, invoking `writerClient.flush()` and forcing a metadata update for each row. This may prevent historical CDC records from batching and may also flush unrelated writers that share the connection.

Would it make sense to track the active write schema separately, or let the shared accumulator roll the batch when the schema changes? A test that sends several historical-schema records without waiting for each future could help verify that the refresh/flush is not repeated per row.

fluss-client/src/main/java/org/apache/fluss/client/table/writer/MultiTableWriterImpl.java
```diff
@@ -0,0 +1,533 @@
...
+                this.rowEncoder = new CompactedRowEncoder(fieldDataTypes);
+            } else {
+                // ARROW format accepts a general InternalRow; no per-row encoder needed.
+                this.rowEncoder = null;
+            }
+        }
+
+        @Override
+        WriteRecord build(ChangeType changeType, InternalRow row) {
+            switch (changeType) {
```
leonardBang: Could we clarify whether the original `ChangeType` is expected to be preserved for log tables?

`INSERT` is accepted here, but the generated `WriteRecord` does not carry the change type, and the row/Arrow log batches later append it as `APPEND_ONLY`. A scan-then-write pipeline may therefore observe a different changelog type from the input. `UPDATE_BEFORE` is also returned as a successful no-op before the target table kind or existence is resolved.

If log tables are intended to preserve changelog semantics, could the change type be carried through to the batch? Otherwise, would it be safer to narrow the API and reject unsupported values explicitly? A round-trip test asserting the scanned `ChangeType`, rather than only row contents, may help document the intended behavior.

fluss-client/src/main/java/org/apache/fluss/client/table/scanner/log/LogFetcher.java
```diff
@@ -161,14 +139,48 @@ public LogFetcher(
...
+                        scannerName, conf, remoteFileDownloader, scannerMetricGroup);
         remoteLogDownloader.start();
+        this.schemaResolution = schemaResolution;
+    }
+
+    /**
+     * Register a new table for multi-table fetching. If the table is already registered, this is a
+     * no-op.
+     *
+     * <p><b>Threading contract:</b> {@code tableReadContexts} is a plain {@link HashMap} and the
```
leonardBang: Would it make sense to include the RPC callback threads in the synchronization/lifecycle contract here?

The scanner entry points are guarded, but `gateway.fetchLog(...).whenComplete(...)` reads `tableReadContexts` asynchronously while `registerTable()` and `unregisterTable()` mutate this plain `HashMap`. `unregisterTable()` may also close a context that is still referenced by an in-flight response, a buffered `CompletedFetch`, or a remote pending fetch. In addition, falling back to another table's read context when the original context is missing could decode the response with an unrelated schema.

Could the request capture a context snapshot or generation and defer closing it until buffered/in-flight references are released? A stale response with no matching context could then be discarded instead of using another table's context.

fluss-client/src/main/java/org/apache/fluss/client/table/scanner/log/LogFetcher.java
```diff
@@ -630,4 +676,49 @@ LogScannerStatus getLogScannerStatus() {
...
+
+        TableReadContext(
+                TableInfo tableInfo,
+                LogRecordReadContext.SchemaResolution schemaResolution,
+                @Nullable Projection projection,
+                @Nullable Predicate recordBatchFilter,
+                SchemaGetter schemaGetter) {
+            this.tablePath = tableInfo.getTablePath();
+            this.isPartitioned = tableInfo.isPartitioned();
+            this.readContext =
```
leonardBang, line 756: Could these read contexts reuse the `ChunkedFactory` owned by `LogFetcher`?

The five-argument `createReadContext()` overload creates a new factory internally, so the factory closed by `LogFetcher.close()` is not the one used by either the local or remote Arrow context. Closing the allocator releases its buffers, but an inaccessible factory may continue retaining active or free native chunks for reuse.

Would it make sense to pass `chunkedFactory` into both calls, as the previous single-table implementation did, or make each context explicitly own and close the factory it creates?

fluss-client/src/main/java/org/apache/fluss/client/table/scanner/log/MultiTableLogScannerImpl.java
```diff
@@ -0,0 +1,381 @@
...
+            TablePath tablePath = pathByTableId.get(tableId);
+            SchemaGetter schemaGetter = schemaGetters.get(tableId);
+            if (tablePath == null || schemaGetter == null) {
+                // Should not happen: only registered tables can produce records.
+                throw new IllegalStateException(
+                        String.format(
+                                "Dropping records for unknown tableId %s in bucket %s",
+                                tableId, bucket));
+            }
+            builder.addConsumedUpToOffset(bucket, scanRecords.consumedUpToOffset(bucket));
```
leonardBang: Could the progress entry also preserve the corresponding table and bucket in the result?

This call records the consumed offset, but the builder creates the `tablePath -> bucket` entry only when at least one materialized record is added below. A progress-only poll can therefore return `hasProgress() == true` and a non-null consumed offset while `tablePaths()` and `buckets(tablePath)` remain empty, which differs from their documented behavior.

Would it make sense for the progress builder method to receive `tablePath` and create an empty bucket entry? The IT helper could then require the expected table and bucket instead of falling back to an assertion on `hasProgress()` after the timeout.

fluss-client/src/main/java/org/apache/fluss/client/table/scanner/log/MultiTableLogScannerImpl.java
```diff
@@ -0,0 +1,381 @@
...
+            TableInfo tableInfo = getTableInfo(tablePath);
+            if (tableInfo.isPartitioned()) {
+                throw new IllegalStateException(
+                        "Table "
+                                + tablePath
+                                + " is a partitioned table, please use "
+                                + "\"subscribe(TablePath, long partitionId, int bucket, long offset)\" "
+                                + "to subscribe a partitioned bucket instead.");
+            }
+            registerIfAbsent(tablePath, tableInfo);
```
leonardBang, line 180: Would it make sense to make this registration transactional with the remaining subscribe steps?

`registerIfAbsent()` has already populated the scanner maps and created the fetcher's read contexts, while the metadata/partition checks and bucket assignment below may still fail. If that happens, no bucket is added to `subscribedBucketsByPath`; a later `unsubscribe()` reaches `untrackSubscription()` with no bucket set and cannot remove the stale registration or close its contexts.

Could the metadata checks run before registration where possible, or could the exception path roll back the maps and fetcher context? A subscription using a nonexistent partition may be a useful regression test.

fluss-client/src/main/java/org/apache/fluss/client/table/scanner/MultiTableScan.java
```diff
@@ -15,42 +15,27 @@
...
-    @Nullable private final int[] projectedFields;
-
-    public LogScan() {
-        this(null);
-    }
-
-    private LogScan(@Nullable int[] projectedFields) {
-        this.projectedFields = projectedFields;
-    }
+public interface MultiTableScan {
```
leonardBang, line 34: Could we keep the existing `org.apache.fluss.client.table.scanner.log.LogScan` class and add `MultiTableScan` as a separate file?

Git represents this change as a rename, but the two classes serve different purposes. Since `LogScan` has been a `@PublicEvolving` API since 0.1 and is still present on `main`, removing it may break existing source users and cause `NoClassDefFoundError` for binaries compiled against it.

Was the removal intentional? If not, retaining `LogScan.java` unchanged while adding this new interface separately would preserve compatibility.

leonardBang (commented):
I did another focused review of the production paths, with particular attention to schema transitions, shared client state, asynchronous fetch lifecycles, and off-heap resource ownership. I left several inline questions describing the current behavior, the scenarios in which it may surface, and possible regression coverage. It would be helpful to clarify or address these cases before merging.

fluss-client/src/main/java/org/apache/fluss/client/table/scanner/MultiTableScan.java
loserwang1024: Yes, I remove it because no place has used it.

fluss-client/src/main/java/org/apache/fluss/client/write/WriteBatch.java
```diff
@@ -243,6 +269,23 @@ boolean isDone() {
...
 
+    abstract boolean tryAppendRecord(WriteRecord writeRecord, WriteCallback callback)
+            throws Exception;
+
+    private boolean canAppend(WriteRecord writeRecord) {
+        if (tableId != writeRecord.getTableInfo().getTableId()
+                || writeFormat != writeRecord.getWriteFormat()) {
+            return false;
+        }
+        if (schemaId != writeRecord.getSchemaId()) {
```
leonardBang: Would it make sense to treat a schema ID mismatch as a batch-compatibility boundary and return `false` here, similar to table ID and write format mismatches? Currently, writing a valid new-schema record while an old-schema batch is buffered throws synchronously, so CDC/schema-evolution callers must explicitly `flush()` at every schema boundary. Returning `false` would allow `RecordAccumulator` to close the old batch and create a schema-specific new batch automatically. It may also be helpful to change `testSchemaSwitchWithoutFlushFails` to verify that both records succeed without an explicit intermediate flush.

fluss-client/src/main/java/org/apache/fluss/client/table/scanner/MultiTableScan.java
leonardBang: Could we keep `org.apache.fluss.client.table.scanner.log.LogScan` and add `MultiTableScan` as a separate class? Although `LogScan` is not currently referenced inside this repository, it has been a `@PublicEvolving` API since 0.1 with public constructors and projection methods. Removing its FQCN may cause existing clients to fail at compile time or with `NoClassDefFoundError`. A deprecation cycle would be safer if the old API is intended to be retired.

fluss-client/src/main/java/org/apache/fluss/client/table/writer/MultiTableWriteImpl.java
```diff
@@ -0,0 +1,60 @@
...
+    @Override
+    public MultiTableWriter createWriter() {
+        // Each writer owns a dedicated WriterClient (buffer pool + sender thread) instead of the
+        // connection-level shared one, so MultiTableWriter#flush() only affects this writer and
+        // schema-id batching of different writers cannot interfere with each other. The client
+        // is closed by MultiTableWriterImpl#close().
+        return new MultiTableWriterImpl(
+                connection.getMetadataUpdater(),
+                connection.getOrCreateAdmin(),
+                new WriterClient(
```
leonardBang, line 54: Could we avoid creating multiple `WriterClient` instances under the same `ClientMetricGroup`, or give each one a unique metric identity? Every client currently creates a `WriterMetricGroup` with the same `client.writer` scope and `client_id`. When two writers coexist on one connection, JMX rejects the duplicate `ObjectName`, while Prometheus writers share the same label set and may overwrite or remove each other's metrics when one writer closes. Reusing the connection-level writer client after fixing batch rollover would avoid this; otherwise, could we introduce a `writer_id` and add a two-writer reporter test?

fluss-client/src/test/java/org/apache/fluss/client/table/scanner/log/LogFetcherFilterITCase.java
```diff
@@ -430,16 +444,18 @@ private LogFetcher createFetcherWithBuckets(Map<TableBucket, Long> scanBuckets)
...
+                new LogFetcher(
+                        "default fetcher",
+                        scannerStatus,
+                        clientConf,
+                        metadataUpdater,
+                        scannerMetricGroup,
+                        new RemoteFileDownloader(1),
+                        LogRecordReadContext.SchemaResolution.TARGET);
+        logFetcher.registerTable(
+                new TableScanSpec(DATA1_TABLE_INFO, null, recordBatchFilter), TEST_SCHEMA_GETTER);
```
leonardBang: Could this helper construct the registered `TableInfo` with the runtime `tableId`, as `setup()` does? The supplied `scanBuckets` use the table created by this test, while `DATA1_TABLE_INFO` carries a different static table ID. As a result, `prepareFetchLogRequests()` cannot find a `TableReadContext` for the scanned buckets and skips the request, causing the two new tests to time out in `hasAvailableFetches()`. Passing the runtime `TableInfo` into this helper should make the registration and bucket IDs consistent.

fluss-client/src/main/java/org/apache/fluss/client/write/WriterClient.java
```diff
@@ -102,7 +110,7 @@ public WriterClient(
             this.maxRequestSize = maxRequestSizeLocal;
             idempotenceManagerLocal = buildIdempotenceManager();
             this.idempotenceManager = idempotenceManagerLocal;
-            this.writerMetricGroup = new WriterMetricGroup(clientMetricGroup);
+            this.writerMetricGroup = writerMetricGroup;
```
leonardBang: Could we assign `this.writerMetricGroup` before calling `buildIdempotenceManager()`? The latter may fail during configuration validation, in which case the field remains null. The catch block then calls `close()`, which unconditionally invokes `writerMetricGroup.close()` and introduces a `NullPointerException` that masks the original exception.

leonardBang (approved):
Thanks @loserwang1024 for the update, the PR looks good to me now, I only left one minor comment for potential NPE. Feel free to merge it once address it.

PR #3148, opened 2026-04-21
[server] Fix coordinator metrics not updated when no events arrive

### LINKED ISSUE #3147: [server] Coordinator metrics not updated after restart when no events arrive
### Search before asking

### Fluss version

main (development)

### Please describe the bug 🐞

After a `CoordinatorServer` restart, once the init context completes, coordinator-level metrics such as `fluss_coordinator_activeTabletServerCount` remain 0 if no client requests or events arrive.

### Solution

The `CoordinatorEventThread.doWork()` has two issues:

1. `lastMetricsUpdateTime` is initialized to `System.currentTimeMillis()`, so the first `doWork()` invocation skips the metrics update (less than 5s elapsed).

2. `queue.take()` blocks indefinitely. If no events are enqueued, the thread never loops back to re-check the metrics update condition.

As a result, metrics are only updated as a side effect of event processing. In an idle coordinator with no incoming requests, the gauges stay at their initial zero values.

swuferhong:
### Purpose

Linked issue: close #3147 

After coordinator restart, if no client requests arrive, the
`CoordinatorEventThread `stays blocked on `queue.take()` and never
updates metrics gauges like activeTabletServerCount.

Fix by replacing `queue.take()` with `queue.poll(timeout)` so the
thread periodically wakes up to refresh metrics, and initializing
`lastMetricsUpdateTime` to 0 to trigger an immediate first update.

### DIFF
--- a/fluss-server/src/main/java/org/apache/fluss/server/coordinator/event/CoordinatorEventManager.java
+++ b/fluss-server/src/main/java/org/apache/fluss/server/coordinator/event/CoordinatorEventManager.java
@@ -33,6 +33,7 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.util.concurrent.LinkedBlockingQueue;
+import java.util.concurrent.TimeUnit;
 import java.util.concurrent.locks.Lock;
 import java.util.concurrent.locks.ReentrantLock;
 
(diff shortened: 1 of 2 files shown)


platinumhamburg (commented):
LGTM! Would you mind adding some tests for this PR?

swuferhong:
@platinumhamburg  comments addressed.

wuchong (approved):

PR #3149, opened 2026-04-21
[client] Ensure UPDATE_BEFORE and UPDATE_AFTER are never split across poll batches

loserwang1024:
### Purpose

Linked issue: close #2880 

### Brief change log

When CompletedFetch.fetchRecords() hits maxRecords, it may cut between a -U (UPDATE_BEFORE) and +U (UPDATE_AFTER) pair. If the next poll returns records from multiple buckets, downstream stateful converters (e.g., BinlogRowConverter) can see an orphaned -U whose pendingUpdateBefore state gets corrupted by interleaved records from other buckets.

Fix: after the fetch loop, if the last record is UPDATE_BEFORE, fetch one additional record to include the matching UPDATE_AFTER in the same batch. Extract the per-record fetch logic into a reusable fetchRecord() method.


loserwang1024:
@luoyuxia @wuchong , CC

luoyuxia (approved):

PR #3151, opened 2026-04-21
[KV] Kvscan Server Side

### LINKED ISSUE #3125: Implement Server-Side KvScan

polyzos:
This PR brings in the server-side implementation for FIP-17, introduced here

https://github.com/apache/fluss/pull/2809


polyzos:
@wuchong @platinumhamburg @leekeiabstraction @fresh-borzoni PTAL

fluss-server/src/main/java/org/apache/fluss/server/replica/Replica.java
```diff
@@ -376,6 +386,11 @@ public Path getTabletParentDir() {
         return kvTablet;
     }
 
+    /** Injects the {@link ScannerManager} so that {@link #dropKv()} can close active scanners. */
+    public void setScannerManager(@Nullable ScannerManager scannerManager) {
```
wuchong: This is a little hack. The `scannerManager` should be immutable when the Replica is created. It should be a part of the constructor.

fluss-server/src/main/java/org/apache/fluss/server/replica/ReplicaManager.java
```diff
@@ -324,6 +329,10 @@ public void startup() {
                 conf.get(ConfigOptions.LOG_REPLICA_MAX_LAG_TIME).toMillis() / 2);
     }
 
+    public void setScannerManager(ScannerManager scannerManager) {
```
wuchong: This is a little hack. The `ScannerManager` should never be null and should be immutable, so it should be set in the constructor.

fluss-server/src/main/java/org/apache/fluss/server/kv/scan/ScannerManager.java
```diff
@@ -0,0 +1,420 @@
...
+ * <p>A background evictor task runs every {@code kv.scanner.expiration-interval} and removes
+ * sessions idle longer than {@code kv.scanner.ttl}. Recently evicted IDs are retained for {@code 2
+ * × ttl} so callers can distinguish "expired" from "never existed."
+ *
+ * <h3>Leadership change</h3>
+ *
+ * {@link #closeScannersForBucket(TableBucket)} must be called when a bucket loses leadership to
+ * release all RocksDB snapshot/iterator resources for that bucket promptly.
+ */
+public class ScannerManager implements AutoCloseableAsync {
```
wuchong, line 64: IIUC, this class is thread-safe, otherwise, it can be shared across worker threads. Please add `@ThreadSafe` annotation to the class.

fluss-server/src/main/java/org/apache/fluss/server/kv/scan/ScannerManager.java
```diff
@@ -0,0 +1,420 @@
...
+     * @param tableBucket the bucket being scanned
+     * @param limit optional row-count limit ({@code null} or ≤ 0 means unlimited)
+     * @return the newly registered {@link ScannerContext}, or {@code null} if the bucket is empty
+     * @throws TooManyScannersException if the per-bucket or per-server limit is exceeded
+     * @throws IOException if the underlying {@link org.apache.fluss.server.utils.ResourceGuard} is
+     *     already closed (the bucket is shutting down)
+     */
+    @Nullable
+    public ScannerContext createScanner(
+            KvTablet kvTablet, TableBucket tableBucket, @Nullable Long limit) throws IOException {
```
wuchong: We can derive the `TableBucket` variable from `kvTablet.getTableBucket()`. Thus, we don't need the `tableBucket` parameter. Otherwise, we have to check the consistence between `kvTablet.getTableBucket()` and `tableBucket`, which is an overhead.

fluss-server/src/main/java/org/apache/fluss/server/kv/scan/ScannerManager.java
```diff
@@ -0,0 +1,420 @@
...
+        AtomicInteger count = perBucketCount.get(tableBucket);
+        return count == null ? 0 : count.get();
+    }
+
+    /**
+     * Closes and removes all active scanner sessions for the given bucket. Must be called when a
+     * bucket loses leadership to prevent stale RocksDB snapshot/iterator leaks.
+     */
+    public void closeScannersForBucket(TableBucket tableBucket) {
+        List<String> toRemove = new ArrayList<>();
```
wuchong: nit: we can track `ScannerContext` rather than id in the `toRemove` List to avoid additional `scanners.get(key)` concurrent map lookup.

fluss-server/src/main/java/org/apache/fluss/server/kv/scan/ScannerManager.java
```diff
@@ -0,0 +1,420 @@
...
+        for (String key : toRemove) {
+            ScannerContext context = scanners.get(key);
+            if (context != null) {
+                LOG.info(
+                        "Closing scanner {} for bucket {} due to leadership change.",
+                        key,
+                        tableBucket);
+                removeScanner(context);
+            }
+        }
```
wuchong, line 223: It's better to cleanup the `perBucketCount` for the TableBucket at the end. Because it's possible there is no scanners for this bucket, but there is remaining `AtomicInteger` in the `perBucketCount`.

fluss-server/src/main/java/org/apache/fluss/server/kv/scan/ScannerManager.java
```diff
@@ -0,0 +1,420 @@
...
+     */
+    private void checkLimits(TableBucket tableBucket) {
+        if (totalScanners.get() >= maxPerServer) {
+            throw new TooManyScannersException(
+                    String.format(
+                            "Cannot create scanner for bucket %s: server-wide limit of %d reached.",
+                            tableBucket, maxPerServer));
+        }
+        AtomicInteger bucketCount =
+                perBucketCount.computeIfAbsent(tableBucket, k -> new AtomicInteger(0));
```
wuchong: It's better to use `perBucketCount.getOrDefault(tableBucket, ZERO).get()` to avoid initializing an `AtomicInteger` for the bucket. `ZERO` can be a `private static final AtomicInteger(0)` in `ScannerManager`.

fluss-server/src/main/java/org/apache/fluss/server/kv/scan/ScannerManager.java
```diff
@@ -0,0 +1,420 @@
...
+    }
+
+    /**
+     * Atomically increments the counters and puts the context in the map. Throws {@link
+     * TooManyScannersException} and rolls back the increments if a concurrent create caused either
+     * limit to be exceeded between the initial check and this call.
+     */
+    private void registerContext(ScannerContext context, TableBucket tableBucket) {
+        AtomicInteger bucketCount =
+                perBucketCount.computeIfAbsent(tableBucket, k -> new AtomicInteger(0));
```
wuchong: Move it after the `newTotal > maxPerServer` check block.

fluss-server/src/test/java/org/apache/fluss/server/kv/scan/ScannerManagerTest.java
```diff
@@ -0,0 +1,361 @@
...
+            byte[] scannerId = context.getScannerId();
+
+            // Advance ManualClock past TTL so the evictor considers the session idle.
+            clock.advanceTime(500, TimeUnit.MILLISECONDS);
+
+            // Wait for the real scheduler to invoke the cleanup task.
+            long deadline = System.currentTimeMillis() + 10_000;
+            while (manager.activeScannerCount() > 0 && System.currentTimeMillis() < deadline) {
+                Thread.sleep(50);
+            }
```
wuchong: Use `retry` for simplification.

```java
retry(
                    Duration.ofSeconds(10),
                    () -> assertThat(manager.activeScannerCount()).isEqualTo(0));
```

fluss-server/src/main/java/org/apache/fluss/server/tablet/TabletService.java
```diff
@@ -435,7 +445,146 @@ public CompletableFuture<NotifyLakeTableOffsetResponse> notifyLakeTableOffset(
...
+            } else {
+                if (!request.hasScannerId()) {
+                    throw new InvalidScanRequestException(
+                            "ScanKvRequest must have either bucket_scan_req (new scan) "
+                                    + "or scanner_id (continuation).");
+                }
+                byte[] scannerId = request.getScannerId();
+                context = scannerManager.getScanner(scannerId);
+                if (context == null) {
+                    if (scannerManager.isRecentlyExpired(scannerId)) {
```
wuchong, line 529: We should check if the request `isCloseScanner`,  and should return a finished response directly here. Otherwise, a close scanner request will receive a `ScannerExpiredException` or `UnknownScannerIdException` exception, since the server has auto-close the scanner when there is no more data.

fluss-server/src/main/java/org/apache/fluss/server/tablet/TabletService.java
```diff
@@ -435,7 +445,146 @@ public CompletableFuture<NotifyLakeTableOffsetResponse> notifyLakeTableOffset(
...
+            if (request.hasCloseScanner() && request.isCloseScanner()) {
+                scannerManager.removeScanner(context);
+                openedContext = null;
+                response.setScannerId(context.getScannerId());
+                response.setHasMoreResults(false);
+                return CompletableFuture.completedFuture(response);
+            }
+
+            // Build the next batch
+            int batchSizeBytes = request.getBatchSizeBytes();
```
wuchong: The batch size bytes field is optional, we should check the existence first.

fluss-server/src/main/java/org/apache/fluss/server/tablet/TabletService.java
```diff
@@ -435,7 +445,146 @@ public CompletableFuture<NotifyLakeTableOffsetResponse> notifyLakeTableOffset(
...
+            if (batchSizeBytes <= 0) {
```
wuchong: We should have a server-side max batch size to protect server OOM, e.g., max 10MB. Otherwise, a bad user can kill the server by setting `Interger.MAX` batch size.

fluss-server/src/main/java/org/apache/fluss/server/tablet/TabletService.java
```diff
@@ -435,7 +445,146 @@ public CompletableFuture<NotifyLakeTableOffsetResponse> notifyLakeTableOffset(
...
+                throw new InvalidScanRequestException("batch_size_bytes must be greater than 0.");
+            }
+            DefaultValueRecordBatch.Builder builder = DefaultValueRecordBatch.builder();
+            long totalBytes = 0L;
+
+            while (context.isValid() && totalBytes < batchSizeBytes) {
+                byte[] value = context.currentValue();
+                builder.append(value);
+                totalBytes += value.length;
```
wuchong: We should use the `builder.sizeInBytes` rather than the totalBytes which is more accurate for the final record batch size.

fluss-server/src/main/java/org/apache/fluss/server/tablet/TabletService.java
```diff
@@ -435,7 +445,146 @@ public CompletableFuture<NotifyLakeTableOffsetResponse> notifyLakeTableOffset(
...
+
+        } catch (Exception e) {
+            // Restore the interrupt flag if a lower-level call wrapped an InterruptedException.
+            // No method in the try block declares `throws InterruptedException` directly, but a
+            // future refactor or a rethrown wrapper should not silently lose the signal.
+            if (e instanceof InterruptedException || e.getCause() instanceof InterruptedException) {
+                Thread.currentThread().interrupt();
+            }
+            response.setErrorCode(Errors.forException(e).code());
+            response.setErrorMessage(e.getMessage() != null ? e.getMessage() : "");
```
wuchong: it's suggested to use

```java
ApiError error = ApiError.fromThrowable(e).
response.setError(error.error().code(), error.message());
```

fluss-server/src/main/java/org/apache/fluss/server/kv/scan/ScannerManager.java
```diff
@@ -0,0 +1,420 @@
...
+     *     already closed (the bucket is shutting down)
+     */
+    @Nullable
+    public ScannerContext createScanner(
+            KvTablet kvTablet, TableBucket tableBucket, @Nullable Long limit) throws IOException {
+        checkLimits(tableBucket);
+
+        String scannerId = generateScannerId();
+        ScannerContext context =
+                kvTablet.openScan(scannerId, limit != null ? limit : -1L, clock.milliseconds());
```
wuchong: Unlike `lookups()`, `prefixLookup()`, and `limitKvScan()`, `replicaManager.getLeaderKvTablet(tableBucket)` returns the raw `KvTablet` after only an unsynchronized `isLeader()` check. If leadership flips after `getLeaderKvTablet()` returns but before
    `createScanner()` registers the session, `makeFollowers()` has already run `closeScannersForBucket()`, so the
    new scanner survives on a follower. Subsequent `scanKv(scanner_id=...)` calls never re-check leadership,
    which lets clients keep reading a stale snapshot from a bucket that no longer has the leader.

A safe way to open scan should be have a `openScan` on the `Replica`, and returns the rocksdb snapshot under a `leaderIsrUpdateLock` read lock, just like other methods (lookups, limitScan...).

fluss-server/src/main/java/org/apache/fluss/server/kv/scan/ScannerContext.java
```diff
@@ -0,0 +1,177 @@
...
+    // in-order check: expectedSeqId = callSeqId + 1 = -1 + 1 = 0.
+    // callSeqId validation is only performed for continuation requests (those carrying a
+    // scanner_id), never for the initial open request (those carrying a bucket_scan_req).
+    private int callSeqId = -1;
+
+    /**
+     * Wall-clock timestamp (ms) of the most recent request that touched this session. Used by
+     * {@link ScannerManager} for TTL-based eviction.
+     */
+    private long lastAccessTime;
```
wuchong: ScannerManager updates lastAccessTime on RPC workers and reads it later from the background TTL evictor, but this state is stored in plain fields. Without `volatile`, the evictor can observe a stale timestamp and expire a scanner that was just used; callSeqId has the same visibility problem if a continuation is handled by a different worker after reconnect/retry. Since ScannerContext is intentionally shared across threads by the new manager, these fields need a happens-before edge.

fluss-server/src/main/java/org/apache/fluss/server/tablet/TabletService.java
```diff
@@ -435,7 +445,146 @@ public CompletableFuture<NotifyLakeTableOffsetResponse> notifyLakeTableOffset(
...
+            // If we made it past createScanner/getScanner but failed to deliver a complete
+            // response, close the session rather than leaking it to TTL. The cursor has
+            // already advanced past rows whose values were never sent; resuming would drop
+            // data. Forcing a restart is the safe option.
+            if (openedContext != null) {
+                scannerManager.removeScanner(openedContext);
+            }
+        }
+
+        return CompletableFuture.completedFuture(response);
```
wuchong, line 645: The `ScanKvResponse` contract in proto file says the initial batch must carry the log offset captured when the RocksDB snapshot was opened, but this implementation only fills `scanner_id`, `has_more_results`, and `records`.

Without that offset, a client cannot do a consistent snapshot-to-log handoff (and the empty-bucket fast path has no other way to return it), so scans can miss or duplicate updates that race with the snapshot.

We should initialize the log_offset into the `ScannerContext` when opening scan on the replica.

wuchong (commented):
Thanks @polyzos , I left some comments.

polyzos:
@wuchong Thanks you for your thorough and detailed review.

I addressed all the comments, rebased on main and added some improvements. Let me know if you think n there is something more that needs to be addressed.

fluss-server/src/main/java/org/apache/fluss/server/kv/scan/ScannerContext.java
```diff
@@ -0,0 +1,227 @@
...
+
+    @Override
+    public void close() {
+        if (closed.compareAndSet(false, true)) {
+            // Fence: closing the iterator while another thread is mid iterator.next() is
+            // undefined at the RocksDB JNI boundary. tryAcquireForUse() re-checks `closed`
+            // after winning its CAS, so any racing waiter releases inUse and lets us through.
+            while (inUse.get()) {
+                Thread.yield();
+            }
```
wuchong: The current implementation uses an unbounded spin loop with Thread.yield(), which poses a significant risk of infinite CPU consumption if the inUse flag is never cleared due to bugs or stuck operations. Furthermore, relying on Thread.yield() is generally discouraged in production environments. Regarding the concern about undefined behavior at the RocksDB JNI boundary, I believe it is unnecessary here. If a scan is active while the corresponding KV tablet is dropped or the scan times out, the appropriate response is to throw an exception to the user rather than waiting indefinitely. Therefore, I recommend replacing the blocking wait with a direct update that sets inUse to false. This approach simplifies the logic and avoids the potential performance pitfalls of busy-waiting.

fluss-server/src/main/java/org/apache/fluss/server/tablet/TabletServer.java
```diff
@@ -291,7 +301,9 @@ protected void startServices() throws Exception {
                             metadataManager,
                             authorizer,
                             dynamicConfigManager,
-                            ioExecutor);
+                            ioExecutor,
+                            scannerManager,
+                            kvScanMaxBatchSizeBytes);
```
wuchong: We should move `kvScanMaxBatchSizeBytes` into `ScannerManager` to reduce the number of parameters in the TabletService constructor. Since this configuration logically belongs to ScannerManager, this change improves code cohesion and simplifies initialization.

wuchong (approved):
LGTM.

I pushed 2 commits to fix the remaining issues. Will merge this once CI is passed.

PR #3152, opened 2026-04-21
[fluss/ai agent] enable ai doc agent on website

MehulBatra:
https://github.com/user-attachments/assets/249e6850-c399-4bdb-b581-666ef739b0c7

### Purpose

Enable Document AI agent to crawl to website,  repo and retreive data and reply based on that to the end user.

### DIFF
--- a/website/docusaurus.config.ts
+++ b/website/docusaurus.config.ts
@@ -20,7 +20,8 @@ import type {Config} from '@docusaurus/types';
 import type * as Preset from '@docusaurus/preset-classic';
 import lightTheme from './src/utils/prismLight';
 import versionReplace from './src/plugins/remark-version-replace/index';
-import { loadVersionData } from './src/utils/versionData';
+import {loadVersionData} from './src/utils/versionData';
+
 const { versionsMap, latestVersion } = loadVersionData();
 
 const config: Config = {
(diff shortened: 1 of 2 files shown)


website/docusaurus.config.ts
```diff
@@ -47,6 +48,26 @@ const config: Config = {
   // Blog content is cloned from a separate repo via setup_blog.sh
   staticDirectories: ['static', 'blog/static'],
 
+  scripts: [
+    {
+      src: 'https://widget.kapa.ai/kapa-widget.bundle.js',
```
polyzos, line 53: @wuchong i think there might be an issue with ASF on this? or is it good?

MehulBatra: I think we already have youtube and google calendar integration, it's the same we did for kappa.
Also apache hudi is already using the agent if that helps.
image

wuchong:
image
It overlaps with the "up" button. Maybe we can put it on the navbar like Doris?

image

website/docusaurus.config.ts
wuchong: Yes, it is blocked by default. And we need to update the `website/.htaccess` to include the related external resource domains. But additional allowances need to be coordinated with privacy team (see previous mailing list https://lists.apache.org/thread/xt0ggzc7f4dj8g91b92w4tk5g8cdslwd -- login via your apache id). 

See Doris: https://github.com/apache/doris-website/blob/asf-site/.htaccess

MehulBatra:
Sure fixing it, with doris I feel one problem, it's very prominent the moment user lands I wanted to make sure the icon is visible.

MehulBatra:
@wuchong @polyzos addressed concerns
image

wuchong:
image

Currently, the "Ask AI" is blocked. We need to request the privacy team for allow the domain. I suggest to revert the commit first, and add it back with the `.htaccess` updates after the privacy team allow. 

image

wuchong:
cc @MehulBatra @polyzos

MehulBatra:
sure will do the needful!

PR #3154, opened 2026-04-21
[helm] scheduling primitives

### LINKED ISSUE #3153: [helm] Add pod scheduling primitives (affinity, nodeSelector, tolerations, topologySpreadConstraints)
### Search before asking

### Description

Production Kubernetes deployments often need to spread tablet server replicas across availability zones and nodes to survive zone/node failures, pin workloads to specific node pools, or allow scheduling on tainted nodes.

Without these primitives exposed in the chart, users have to fork the templates or patch the rendered manifests.

### Willingness to contribute

fresh-borzoni:
closes https://github.com/apache/fluss/issues/3153

Exposes affinity, nodeSelector, tolerations, and topologySpreadConstraints on both coordinator and tablet. Required to spread replicas across zones/nodes under replication factor 3 - without these, a single node failure can take out all tablet replicas for segments not yet tiered.

### DIFF
--- a/helm/values.yaml
+++ b/helm/values.yaml
@@ -54,6 +54,31 @@ tablet:
     enabled: false
     # minAvailable: 1
     # maxUnavailable: 1
+  # Pod scheduling configuration
+  affinity: {}
+  # Example: spread tablet server pods across availability zones and nodes.
+  # If multiple Fluss releases share the cluster, also scope by instance via
+  # `app.kubernetes.io/instance: <release-name>` to avoid cross-release anti-affinity.
+  # affinity:
+  #   podAntiAffinity:
+  #     preferredDuringSchedulingIgnoredDuringExecution:
+  #       - weight: 100
+  #         podAffinityTerm:
+  #           topologyKey: topology.kubernetes.io/zone
+  #           labelSelector:
+  #             matchLabels:
+  #               app.kubernetes.io/name: fluss
+  #               app.kubernetes.io/component: tablet
+  #       - weight: 50
+  #         podAffinityTerm:
+  #           topologyKey: kubernetes.io/hostname
+  #           labelSelector:
+  #             matchLabels:
+  #               app.kubernetes.io/name: fluss
+  #               app.kubernetes.io/component: tablet
+  nodeSelector: {}
+  tolerations: []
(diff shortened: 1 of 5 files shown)


fresh-borzoni:
@morazow @affo @polyzos PTAL 🙏

affo (approved):
Loving where we are going with the Helm charts, thanks @fresh-borzoni for this one, another step towards full flexibility 🚀 

I just have a question, why do you mention "With replication factor 3, a single node failure could take out all replicas simultaneously..."? I mean, this would hold with any replication factor if nodes are co-located, isn't it?

helm/values.yaml
```diff
@@ -54,6 +54,31 @@ tablet:
...
+  # `app.kubernetes.io/instance: <release-name>` to avoid cross-release anti-affinity.
+  # affinity:
+  #   podAntiAffinity:
+  #     preferredDuringSchedulingIgnoredDuringExecution:
+  #       - weight: 100
+  #         podAffinityTerm:
+  #           topologyKey: topology.kubernetes.io/zone
+  #           labelSelector:
+  #             matchLabels:
+  #               app.kubernetes.io/name: fluss
```
morazow: Maybe change this to `<release-name>` as in docs, so that user don't blindly uncomment 🤝 

Also below, lines 77

morazow (approved):
Minor nit from my side in the values file 🤝 

Thanks, shaping nicely 🚀

fresh-borzoni:
> I just have a question, why do you mention "With replication factor 3, a single node failure could take out all replicas simultaneously..."? I mean, this would hold with any replication factor if nodes are co-located, isn't it?

Well, it's just me missing "Even" before with, bc we had exactly this scenario and the first reaction was - but how all 3 simultaneously :)

helm/values.yaml
fresh-borzoni: Changed the example to app.kubernetes.io/instance: <release-name> in both snippets. 

'name: fluss' is the same value across every Fluss release in the cluster, so the anti-affinity would have silently counted pods across releases. Scoping by instance targets only the current release, and the <release-name> placeholder forces the user to substitute before applying.

PR #3169, opened 2026-04-23
[flink] Push down lake filters for non-partitioned scans

### LINKED ISSUE #3168: [flink] Lake filters shoud be pushed down for non-partitioned scans
### Search before asking

### Fluss version

main (development)

### Please describe the bug 🐞

`FlinkTableSource.applyFilters(...)` only forwards converted predicates to `lakeSource.withFilters(...)` inside the `isPartitioned()` branch.

That means a non-partitioned table with data lake enabled never pushes SQL filters down to the lake source, even when the predicates are convertible and the underlying lake format supports filter pushdown. In practice, batch reads in `FULL` mode fall back to Flink-side filtering for the lake portion of the scan instead of letting the lake source prune files or rows earlier.

A minimal reproduction is:

1. Create a non-partitioned Fluss table with data lake enabled.
2. Run a batch query in `FULL` startup mode with filters such as `region = 'HangZhou'` or `value > 1000`.
3. Observe that `FlinkTableSource` never calls `lakeSource.withFilters(...)` for that table, because the pushdown path is guarded by `isPartitioned()`.

Expected behavior:

Convertible filters should be pushed to the lake source whenever `lakeSource != null`, regardless of whether the table is partitioned.

Actual behavior:

Only partitioned tables attempt lake filter pushdown.

### Solution

Move the lake filter pushdown logic out of the partition-only branch so that all lake-backed scans can attempt `lakeSource.withFilters(...)`. Keep the existing fallback behavior when the lake source rejects some predicates.

luoyuxia:
### Purpose

Linked issue: close #3168

Allow `FlinkTableSource` to push filters into `LakeSource` for non-partitioned table scans. Before this change, lake filter pushdown was nested under `isPartitioned()`, so non-partitioned scans in lake-backed `FULL` reads always fell back to Flink-side filtering for historical data.

### Brief change log

- keep lake filter pushdown enabled for partitioned tables and extend the same path to non-partitioned tables
- make lake pushdown only contribute accepted filter metadata so it does not bypass Fluss-side partition pruning or record-batch filtering
- document why `pushdownLakeFilters(...)` updates `acceptedFilters` but intentionally does not rewrite `remainingFilters`
- existing union-read ITs already exercise both partitioned and non-partitioned log tables through `FlinkUnionReadLogTableITCase#testReadLogTableFullType(boolean isPartitioned)` in both Iceberg and Paimon modules

### Tests

- `./mvnw -pl fluss-flink/fluss-flink-common -DskipITs test`
- `./mvnw -pl fluss-flink/fluss-flink-common spotless:check -Dspotless.check.skip=false`
- Existing IT note: `FlinkUnionReadLogTableITCase#testReadLogTableFullType` runs with `isPartitioned = false` and `true`; the partitioned branch additionally asserts partition filter pushdown in the plan

### API and Format

### Documentation

No documentation changes.

### DIFF
--- a/fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/source/FlinkTableSource.java
+++ b/fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/source/FlinkTableSource.java
@@ -572,7 +572,19 @@ public class FlinkTableSource
             acceptedFilters.addAll(recordBatchResult.getAcceptedFilters());
         }
 
-        pushdownLakeFilters(filters, acceptedFilters);
+        if (lakeSource != null) {
+            // Lake pushdown only contributes extra accepted filters.
+            // We intentionally do not mutate remainingFilters here:
+            // 1. partition/record-batch pushdown has already established the Fluss-side scan
+            //    behavior;
+            // 2. FLINK-38635 requires returning the original filters as remaining so Flink keeps
+            //    applying them as a correctness safety net;
+            // 3. LakeSource reports accepted predicates in converted Predicate form rather than the
+            //    original ResolvedExpression instances, so adjusting remainingFilters here would
+            //    add bookkeeping complexity without changing runtime behavior.
+            pushdownLakeFilters(filters, acceptedFilters);
+        }
+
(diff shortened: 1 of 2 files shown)


beryllw:
LGTM！

PR #3171, opened 2026-04-23
[helm] Add existingSecret support for SASL credentials

### LINKED ISSUE #3172: [helm] Add existingSecret support for SASL credentials
### Search before asking

### Description
The Fluss Helm chart currently renders SASL credentials directly from values.yaml into a Helm-managed Secret. This forces operators to either commit plaintext passwords to values.yaml (and the Helm release storage) or fork the chart to integrate with external secret sources like External Secrets Operator.
                                          
The chart should allow each listener to reference a pre-existing Kubernetes Secret for its credentials, keeping plaintext passwords out of the Helm release and out of values.yaml.

### Willingness to contribute

fresh-borzoni:
closes https://github.com/apache/fluss/issues/3172

Allow SASL credentials to come from pre-existing Kubernetes Secrets via existingSecret (internal, zookeeper) and per-user passwordSecretRef (client), so passwords don't need to live in values.yaml or the Helm release. 

On rendering path, an init container reuses the main Fluss image to write jaas.conf from secretKeyRef env at pod startup, the literal-values path is unchanged.

### DIFF
--- a/helm/templates/_jaas.tpl
+++ b/helm/templates/_jaas.tpl
@@ -51,13 +51,22 @@ Usage:
 {{- /* client */ -}}
 {{- if and (include "fluss.security.sasl.plain.enabled" .) (eq $clientMechanism "plain") }}
 {{- range $idx, $user := .Values.security.client.sasl.plain.users | default (list) }}
-- name: {{ include "fluss.security.sasl.plain.client.envVarName" $idx }}
-{{- $userRef := $user.passwordSecretRef | default (dict) }}
-{{- if and $userRef.name $userRef.key }}
+{{- $ref := $user.existingSecret | default (dict) }}
+- name: {{ include "fluss.security.sasl.plain.client.envVarName" (dict "field" "username" "idx" $idx) }}
+{{- if $ref.name }}
   valueFrom:
     secretKeyRef:
-      name: {{ $userRef.name }}
-      key: {{ $userRef.key }}
+      name: {{ $ref.name }}
+      key: {{ $ref.usernameKey | default "username" }}
+{{- else }}
+  value: {{ $user.username | quote }}
+{{- end }}
+- name: {{ include "fluss.security.sasl.plain.client.envVarName" (dict "field" "password" "idx" $idx) }}
+{{- if $ref.name }}
+  valueFrom:
+    secretKeyRef:
+      name: {{ $ref.name }}
+      key: {{ $ref.passwordKey | default "password" }}
 {{- else }}
   value: {{ $user.password | quote }}
 {{- end }}
--- a/helm/templates/_jaas.tpl
+++ b/helm/templates/_jaas.tpl
@@ -1,0 +1,136 @@
+{{/*
+Licensed to the Apache Software Foundation (ASF) under one
+or more contributor license agreements.  See the NOTICE file
+distributed with this work for additional information
+regarding copyright ownership.  The ASF licenses this file
+to you under the Apache License, Version 2.0 (the
+"License"); you may not use this file except in compliance
+with the License.  You may obtain a copy of the License at
+
+     http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/}}
+
+{{/*
+Renders the env entries for the JAAS init container: secretKeyRef for each
+Secret-sourced credential, literal `value:` for each values-sourced one.
+Usage:
+  include "fluss.security.jaas.initContainer.env" .
+*/}}
+{{- define "fluss.security.jaas.initContainer.env" -}}
+{{- $internalMechanism := include "fluss.security.listener.mechanism" (dict "context" .Values "listener" "internal") -}}
+{{- $clientMechanism := include "fluss.security.listener.mechanism" (dict "context" .Values "listener" "client") -}}
+{{- $zkEnabled := include "fluss.security.zookeeper.sasl.enabled" . -}}
+{{- /* internal */ -}}
+{{- if and (include "fluss.security.sasl.plain.enabled" .) (eq $internalMechanism "plain") }}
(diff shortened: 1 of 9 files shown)


fresh-borzoni:
@affo @morazow @polyzos PTAL 🙏

helm/templates/_jaas.tpl
```diff
@@ -0,0 +1,190 @@
...
+Renders the env entries for the JAAS init container: secretKeyRef per
+Secret-sourced credential, literal value per values-sourced credential.
+Usage:
+  include "fluss.security.jaas.initContainer.env" .
+*/}}
+{{- define "fluss.security.jaas.initContainer.env" -}}
+{{- $internalMechanism := include "fluss.security.listener.mechanism" (dict "context" .Values "listener" "internal") -}}
+{{- $clientMechanism := include "fluss.security.listener.mechanism" (dict "context" .Values "listener" "client") -}}
+{{- $zkEnabled := include "fluss.security.zookeeper.sasl.enabled" . -}}
+{{- if and (include "fluss.security.sasl.plain.enabled" .) (eq $internalMechanism "plain") }}
```
affo, line 30: nit: split this block with comment: "internal" for readability

helm/templates/_jaas.tpl
```diff
@@ -0,0 +1,190 @@
...
+{{- if $ref.name }}
+  valueFrom:
+    secretKeyRef:
+      name: {{ $ref.name }}
+      key: {{ $ref.passwordKey | default "password" }}
+{{- else }}
+  value: {{ include "fluss.security.sasl.plain.internal.password" . | quote }}
+{{- end }}
+{{- end }}
+{{- if and (include "fluss.security.sasl.plain.enabled" .) (eq $clientMechanism "plain") }}
```
affo, line 52: nit: split this block with comment: "client" for readability

helm/templates/_jaas.tpl
```diff
@@ -0,0 +1,190 @@
...
+  valueFrom:
+    secretKeyRef:
+      name: {{ $userRef.name }}
+      key: {{ $userRef.key }}
+{{- else }}
+  value: {{ $user.password | quote }}
+{{- end }}
+{{- end }}
+{{- end }}
+{{- if $zkEnabled }}
```
affo, line 76: nit: split this block with comment: "zookeeper" for readability

helm/templates/_jaas.tpl
```diff
@@ -0,0 +1,190 @@
...
+{{/*
+Init container spec that renders jaas.conf at pod startup.
+Usage:
+  include "fluss.security.jaas.initContainer" .
+*/}}
+{{- define "fluss.security.jaas.initContainer" -}}
+{{- $img := .Values.security.jaas.initContainer.image | default (dict) -}}
+- name: render-jaas-config
+  {{- if and $img.repository $img.tag }}
+  image: "{{ $img.repository }}:{{ $img.tag }}"
```
affo: Another minor:
I find this a bit weird that we let the user decide this 🤔 
As the user does not know what the script does, hence it would be difficult to actually pick an image 🤔 

Would it make sense to use `busybox`? I also understand that we can use the Fluss image resulting in 0 overhead given that it is required in any case to run Fluss itself.

Probably just irrelevant.

affo (commented):
Hello @fresh-borzoni and thanks for the contribution!

This is definitely a must-have 🤝 

I do have doubt for the solution implementation, not for the values design that looks perfect 🤝 

This PR is actually duplicating the contents of the JAAS configuration into a "hidden" script in templates, we would end up having the same configuration template in both the configmap (https://github.com/apache/fluss/blob/main/helm/templates/secret-jaas-config.yaml) and in the scripts.

We may tackle this by making both ways of rendering pass through the script.
However, I don't like that much to have that as it is quite hard to maintain and the resulting rendering can only be tested by actually running a Fluss cluster.

The other solution that I have in mind would be: why don't we treat the contents of jaas conf as text file with and load it, and make the init script template it?
Best would be if the text file is actually a gomplate template, so that we can also easily template it locally for verifying the context.
This would need the init container to be one that ships gomplate, see https://docs.gomplate.ca/installing/#use-with-docker.

I see drawbacks in this, as we would require to download that image as init container.

Happy to debate 🤝

fresh-borzoni:
@affo Thank for the nice suggestions, I really liked the idea with envsubst.
Redesigned the approach and removed existing duplication, PTAL 🙏

helm/templates/_jaas.tpl
```diff
@@ -0,0 +1,136 @@
...
+  image: {{ include "fluss.image" . }}
+  imagePullPolicy: {{ .Values.image.pullPolicy }}
+  env:
+    {{- include "fluss.security.jaas.initContainer.env" . | nindent 4 }}
+  command:
+    - /bin/sh
+    - -ec
+    - |
+      umask 077
+      envsubst < /tmpl/jaas.conf > /jaas/jaas.conf
```
affo, line 117: ❤️

website/docs/install-deploy/deploying-with-helm.md
```diff
@@ -210,6 +211,135 @@ It is recommended to set these explicitly in production.
...
+    sasl:
+      mechanism: plain
+      plain:
+        users:
+          - username: alice
+            password: alice-literal-password   # literal — visible in values.yaml
+          - username: bob
+            passwordSecretRef:                 # or resolved at pod startup
+              name: fluss-client-sasl
+              key: bob-password
```
affo: Question: why a separate pattern for this case?

Wouldn't it be more coherent to stick with the `internal` case for consistency?

E.g.:

```yaml
security:
  client:
    sasl:
      mechanism: plain
      plain:
        users:
          - username: alice
            password: alice-literal-password   # literal — visible in values.yaml
          - existingSecret:
               name: fluss-client-sasl-bob   # required
                usernameKey: username       # optional, defaults to "username"
               passwordKey: password       # optional, defaults to "password"
```

fresh-borzoni: leftover from previous design with always literals for ux, fixing 👍

affo: resolved

affo (approved):
@fresh-borzoni thanks for the prompt addressing, now this looks perfect:
 - favor testable and maintainable approach for ConfigMap for JAAS conf by `envsubst` with env vars
 - consistency between internal and client security

PR #3174, opened 2026-04-23
[server] Add TableLifecycleThrottler to throttle drops in coordinator queue

### LINKED ISSUE #3173: [server] Auto partition retention cleanup delayed by DAY partition creation jitter
### Search before asking

### Motivation

When using auto-partitioned tables with DAY time unit, partition cleanup is delayed by up to ~23 hours beyond the expected retention boundary. Expired partitions should be cleaned up based on the actual current time, regardless of the jitter delay. The random delay should only affect partition creation timing.

### Solution

_No response_

### Anything else?

_No response_

### Willingness to contribute

swuferhong:
### Purpose

Linked issue: close #3173 

Cascading `dropTable` / `dropPartition` used to push `DropPartitionEvent` and `DropTableEvent` into the coordinator event queue directly from `TableChangeWatcher` (and via TableManager#resumeDeletions on startup).
A single drop spanning thousands of partitions / stale tables would flood the queue with drop events and starve unrelated coordinator work.

So this pr introduce `ReplicaCleanupManager` uses a FIFO queue with a per-batch bucket budget (coordinator.replica-cleanup.max-buckets-per-batch, default 10000). Drops are admitted as long as the running sum
  of in-flight buckets stays within budget. A starvation guard ensures that when nothing is in-flight, even an over-budget drop is admitted alone. Completion callbacks from TableManager release the budget and trigger the next batch.

 Two submission paths:
 - 1. Watcher path: TableChangeWatcher NODE_DELETED → submitPartitionDrop/submitTableDrop → buffers a DropPartitionEvent/DropTableEvent for the coordinator event queue
 - 2. Startup resume path: TableManager.resumeDeletions() → submitPartitionDropForResume/submitTableDropForResume → executes a Runnable that drives
  onDeletePartition/onDeleteTable directly (bypasses event queue since TableInfo may already be gone)


fluss-server/src/main/java/org/apache/fluss/server/coordinator/AutoPartitionManager.java
```diff
@@ -335,12 +385,46 @@ private void doAutoPartition(Instant now, Set<Long> tableIds, boolean forceDoAut
...
+                                    + "because coordinator event queue size {} >= threshold {}.",
+                            tablePath,
+                            tableId,
+                            qSize,
+                            dropQueueBackpressureThreshold);
+                    skipDrop = true;
+                }
+            }
+
+            if (!skipDrop && (unlimitedBudget || dropBudgetThisRound > 0)) {
```
platinumhamburg: dropBudgetThisRound is always greater than 0.

fluss-server/src/main/java/org/apache/fluss/server/coordinator/AutoPartitionManager.java
```diff
@@ -335,12 +385,46 @@ private void doAutoPartition(Instant now, Set<Long> tableIds, boolean forceDoAut
...
+                // Use the actual current time for drop to ensure timely cleanup.
+                // The random delay is only meant for spreading partition creation load,
+                // not for delaying retention cleanup.
+                int budgetForCall = unlimitedBudget ? Integer.MAX_VALUE : dropBudgetThisRound;
```
platinumhamburg: dropBudgetThisRound appears to be a local variable. Why is it necessary to recompute a new variable based on it?

fluss-server/src/main/java/org/apache/fluss/server/coordinator/AutoPartitionManager.java
```diff
@@ -467,6 +569,9 @@ private void dropPartitions(
         // (a=?,dt=20250506,b=?) (a=?,dt=20250507,b=?) will be retained.
         Iterator<Map.Entry<String, Set<String>>> iterator =
                 currentPartitions.headMap(lastRetainPartitionTime).entrySet().iterator();
+        int dropped = 0;
```
platinumhamburg: Why is this variable needed?

platinumhamburg (changes requested):
It seems that the designed mechanism is somewhat complex and has quite a few configuration variables. Could we simplify the implementation so that doAutoPartition() maintains a PendingDeletePartition set, and an asynchronous deletion thread polls this set, following these rules:

Only delete one partition at a time;
Skip whenever CoordinatorEventQueue is not empty;
The thread can poll the event queue every few seconds to look for a deletion opportunity.

swuferhong:
> It seems that the designed mechanism is somewhat complex and has quite a few configuration variables. Could we simplify the implementation so that doAutoPartition() maintains a PendingDeletePartition set, and an asynchronous deletion thread polls this set, following these rules:
> 
> Only delete one partition at a time; Skip whenever CoordinatorEventQueue is not empty; The thread can poll the event queue every few seconds to look for a deletion opportunity.

Hi @platinumhamburg. The previous design was a bit complex and not general enough (e.g. it didn't really cover user-submitted batch drops of tables or partitions), so the new PR abstracts a common class as a throttling queue for replica drop events. Mind taking another look when you have a moment? Thanks a lot!

platinumhamburg (approved):
LGTM. The tableBucketCount cache might have a leak, but it is bounded, and its memory usage is not high, so it's acceptable.

fluss-server/src/main/java/org/apache/fluss/server/coordinator/CoordinatorEventProcessor.java
```diff
@@ -886,6 +905,11 @@ private void processDropTable(DropTableEvent dropTableEvent) {
 
         // remove table metrics.
         coordinatorMetricGroup.removeTableMetricGroup(dropTableInfo.getTablePath(), tableId);
+
+        // For partitioned tables the table-level assignment is empty, so no
+        // DeleteReplicaResponse will ever arrive to trigger resumeDeletions().
+        // Call it now so the vacuously-complete table is finalized immediately.
+        tableManager.resumeDeletions();
```
luoyuxia: Partition replica deletions will still produce DeleteReplicaResponse events via DropPartitionEvent -> onDeletePartition(); which will then trigger resumeDeletions.

So, do we really need to call `tableManager.resumeDeletions();` in here?

fluss-server/src/main/java/org/apache/fluss/server/coordinator/TableManager.java
```diff
@@ -321,16 +348,12 @@ private void asyncDeletePartitionMetadata(long partitionId) {
     }
 
     private boolean isEligibleForDeletion(long tableId) {
-        // the table is queued for deletion and
```
luoyuxia, line 324: why remove the comments?

fluss-server/src/main/java/org/apache/fluss/server/coordinator/TableManager.java
```diff
@@ -321,16 +348,12 @@ private void asyncDeletePartitionMetadata(long partitionId) {
...
-        // no any replica is in state deletion started
         return coordinatorContext.isTableQueuedForDeletion(tableId)
                 && !coordinatorContext.isAnyReplicaInState(
                         tableId, ReplicaState.ReplicaDeletionStarted);
     }
 
     private boolean isEligibleForDeletion(TablePartition tablePartition) {
-        // the partition is queued for deletion and
```
luoyuxia, line 332: why remove the comments?

fluss-server/src/main/java/org/apache/fluss/server/coordinator/TableManager.java
```diff
@@ -208,45 +224,49 @@ public void resumeDeletions() {
...
         }
     }
 
     private void resumePartitionDeletions() {
         Set<TablePartition> partitionsToDelete =
                 new HashSet<>(coordinatorContext.getPartitionsToBeDeleted());
-        Set<TablePartition> eligiblePartitionDeletion = new HashSet<>();
-
         for (TablePartition partition : partitionsToDelete) {
-            // if all replicas are marked as deleted successfully, then partition deletion is done
```
luoyuxia, line 237: nit: why remove the comments

luoyuxia (commented):
@swuferhong Thanks for the pr.
I feel this design is a bit complex and easy to get wrong. It introduces three new configs, and the semantics of these configs are not very intuitive for users/operators.

Most of the complexity seems to come from the bucket-based budget (`coordinator.replica-cleanup.max-buckets-per-batch`): we need to resolve bucket counts, track in-flight buckets, release budget on completion, handle timeout/abandonment, and make the resume path go through the same accounting.

Do we really need bucket-level throttling here? The main issue we want to solve is that a large cascading drop can enqueue too many DropTable/DropPartition events into the coordinator event processor. A simpler approach could be to limit the number of drop events admitted into the coordinator event processor per batch/drain. That would still address most event-queue flooding cases, while avoiding bucket-count lookup, bucket-budget accounting, timeout configs, and a lot of the lifecycle complexity.

swuferhong:
Hi @luoyuxia, comments addressed. I still exposed the timeout ms parameter as config option, because ITCases share a single coordinator, and some drop events do return slowly during tests, so the timeout interval needs to be reduced for testing purposes.

fluss-server/src/test/java/org/apache/fluss/server/coordinator/event/watcher/TableChangeWatcherTest.java
```diff
@@ -109,7 +111,16 @@ void before() {
...
         eventManager = new TestingEventManager();
-        tableChangeWatcher = new TableChangeWatcher(zookeeperClient, eventManager);
+        // Use a real ReplicaCleanupManager so that drop submissions are forwarded into the
+        // event manager as DropPartitionEvent / DropTableEvent (the existing assertions in this
+        // test verify exactly that).
+        replicaCleanupManager =
+                new ReplicaCleanupManager(
+                        eventManager,
+                        org.apache.fluss.utils.clock.SystemClock.getInstance(),
+                        new org.apache.fluss.config.Configuration());
```
wuchong: static import them and remove unnecessary qualifications.

fluss-server/src/main/java/org/apache/fluss/server/coordinator/TableManager.java
```diff
@@ -66,6 +78,10 @@ public TableManager(
         this.ioExecutor = ioExecutor;
     }
 
+    public void setReplicaCleanupManager(@Nullable ReplicaCleanupManager replicaCleanupManager) {
```
wuchong: Could we wire this through the constructor and make it final instead of mutable/nullable? In production it is set before the event thread starts, but TableManager itself does not enforce that. A later call to setReplicaCleanupManager() would race with resumeDeletions()/completeDelete*() because this field is neither final nor volatile. Constructor injection would keep the existing threading assumptions explicit and avoid an unnecessary data race surface.

fluss-server/src/main/java/org/apache/fluss/server/coordinator/TableLifecycleThrottler.java
```diff
@@ -0,0 +1,566 @@
...
+                            + "any residual replica state will be reconciled on next "
+                            + "coordinator startup.",
+                    currentInflight.drop,
+                    elapsed);
+            currentInflight = null;
+            next = admitNext();
+        } finally {
+            lock.unlock();
+        }
+        executeOutsideLock(next);
```
wuchong: This timeout path can run a resumeAction from the timeout checker thread. For resume drops, execute() calls TableManager.onDeleteTable/onDeletePartition directly, and those paths mutate CoordinatorContext and the state machines, while CoordinatorContext is @NotThreadSafe and normally only touched from the coordinator event thread. Can we route resume actions through the coordinator event thread as well, or otherwise make timeout admission only enqueue work?

fluss-server/src/main/java/org/apache/fluss/server/coordinator/ReplicaCleanupManager.java
```diff
@@ -0,0 +1,566 @@
...
+ * TableManager#onDeleteTable} / {@code TableManager#onDeletePartition} directly (their {@code
+ * TableInfo} is already gone, so a regular {@code DropTableEvent} is not viable).
+ *
+ * <p>Timeout-based abandon: the in-flight drop is timestamped at admission time. A periodic task
+ * checks whether the completion callback has not arrived within the hardcoded timeout (3 minutes)
+ * and abandons the drop with a WARN log. Abandonment only releases the manager's in-memory
+ * tracking; any residual replica state machine entries are reconciled on the next coordinator
+ * startup via {@link TableManager#resumeDeletions()}.
+ */
+public class ReplicaCleanupManager implements AutoCloseable {
```
wuchong: I suggest refactoring the jittered partition creation logic in `AutoPartitionManager` into this manager in the future. Since partition creation and deletion involve similar workloads, they should share the same underlying logic. 

To reflect this broader scope, we could rename the class to `TableLifecycleThrottler` or `PartitionLifecycleThrottler`. This naming better captures the idea that partition creation and deletion are lifecycle events, with table drop naturally included. Such a change would also allow us to incorporate partition creation management more seamlessly in the future.

fluss-server/src/main/java/org/apache/fluss/server/coordinator/CoordinatorEventProcessor.java
```diff
@@ -229,9 +232,14 @@ public CoordinatorEventProcessor(
                         tableBucketStateMachine,
                         new RemoteStorageCleaner(conf, ioExecutor),
                         ioExecutor);
+        this.replicaCleanupManager =
+                new ReplicaCleanupManager(coordinatorEventManager, clock, conf);
+        this.tableManager.setReplicaCleanupManager(replicaCleanupManager);
```
wuchong: We should move the `replicaCleanupManager` into the constructor of `TableManager` to make it not null and final. This also simplify the logic of `TableManager` without dealing with null `replicaCleanupManager`.

fluss-server/src/main/java/org/apache/fluss/server/coordinator/ReplicaCleanupManager.java
```diff
@@ -0,0 +1,566 @@
...
+            LOG.debug(
+                    "Ignoring drop submission ({}) because ReplicaCleanupManager is closed.", drop);
+            return;
+        }
+        // Before start() is called, throttling is inactive: execute immediately.
+        // This covers early watcher callbacks that fire before the manager is activated.
+        if (!started.get()) {
+            drop.execute(eventManager);
+            return;
+        }
```
wuchong: I don't fully understand this. Why we need to execute immediately rather than relying on the throttle approach? Even if the manager has not started yet, I think the functionality still works (i.e., put event to EventManager).

fluss-server/src/main/java/org/apache/fluss/server/coordinator/ReplicaCleanupManager.java
```diff
@@ -0,0 +1,566 @@
...
+            lock.lock();
+            try {
+                currentInflight = null;
+                // Try to admit the next pending drop despite the failure.
+                PendingDrop next = admitNext();
+                if (next != null) {
+                    // Recursive execution outside lock; safe because admitNext() won't
+                    // return non-null again until this next drop completes.
+                    executeOutsideLock(next);
+                }
```
wuchong: Why not execute `next` outside the lock? It's possible the `resumeAction` is time consuming.

fluss-server/src/main/java/org/apache/fluss/server/coordinator/TableManager.java
```diff
@@ -208,45 +224,44 @@ public void resumeDeletions() {
...
+            } else if (isEligibleForDeletion(partition)) {
+                long tableId = partition.getTableId();
+                long partitionId = partition.getPartitionId();
+                if (replicaCleanupManager != null) {
+                    String partitionName = coordinatorContext.getPartitionName(partitionId);
+                    replicaCleanupManager.submitPartitionDropForResume(
+                            tableId,
+                            partitionId,
+                            partitionName,
+                            () -> onDeletePartition(tableId, partitionId));
```
wuchong: Why is `onDeletePartition` invoked after the partition has been fully deleted? As I understand it, when a drop partition event reaches the `CoordinatorEventProcessor`, it already triggers `TableManager.onDeletePartition` in `org.apache.fluss.server.coordinator.CoordinatorEventProcessor.processDropPartition`. This leads to redundant state machine transitions.

If we can avoid calling `onDeletePartition` once the partition deletion is complete, we would no longer need the `resumeAction` in `PendingPartition/TableDrop`. This change would significantly simplify the logic in `ReplicaCleanupManager`.

fluss-server/src/main/java/org/apache/fluss/server/coordinator/TableManager.java
```diff
@@ -51,6 +53,16 @@ public class TableManager {
...
 
+    /**
+     * Optional manager used to throttle the asynchronous replica cleanup that follows {@link
+     * #onDeleteTable(long)} / {@link #onDeletePartition(long, long)}. When set, {@link
+     * #resumeDeletions()} routes eligible drops through the manager (one at a time) instead of
+     * invoking the state-machine transitions directly, and {@link #completeDeleteTable(long)} /
+     * {@link #completeDeletePartition(TablePartition)} notify the manager so the next pending drop
+     * can be admitted.
+     */
+    @Nullable private ReplicaCleanupManager replicaCleanupManager;
```
wuchong: should be not-null and final.

fluss-server/src/main/java/org/apache/fluss/server/coordinator/TableManager.java
```diff
@@ -208,45 +224,44 @@ public void resumeDeletions() {
...
-                eligibleTableDeletion.add(tableId);
-            }
-        }
-        if (!eligibleTableDeletion.isEmpty()) {
-            for (long tableId : eligibleTableDeletion) {
-                onDeleteTable(tableId);
+            } else if (isEligibleForDeletion(tableId)) {
+                if (replicaCleanupManager != null) {
+                    replicaCleanupManager.submitTableDropForResume(
+                            tableId, () -> onDeleteTable(tableId));
```
wuchong: I don't like the design of `resumeAction` in `replicaCleanupManager`, because it can easily break the concurrency safety. As `replicaCleanupManager` is invoked by multiple threads (timeout scheduler, zk watcher, AutoPartitionManager), but the `onDeleteTable` is not thread-safe.

wuchong (commented):
Hi @swuferhong , I left some comments.

fluss-server/src/main/java/org/apache/fluss/server/coordinator/ReplicaCleanupManager.java
swuferhong: Ok. I will rename the class to `TableLifecycleThrottler`, and the work of refactoring the jittered partition creation logic in AutoPartitionManager into this manager will be traced by: https://github.com/apache/fluss/issues/3457

fluss-server/src/main/java/org/apache/fluss/server/coordinator/TableLifecycleThrottler.java
swuferhong: Introduce a new event named `ResumeDropEvent` to make sure the thread safe.

fluss-server/src/main/java/org/apache/fluss/server/coordinator/ReplicaCleanupManager.java
swuferhong: You're right — the `admitNext()` logic works correctly regardless of whether `start()` has been called: if nothing is in-flight, the first drop is admitted immediately. The timeout checker (activated by start()) only provides liveness as a safety net, not correctness.

The early-return was actually counter-productive: if multiple watcher callbacks arrived before start(), they would all bypass throttling and flood the event queue — exactly the scenario the throttler is designed to prevent.

I have Removed the block. All submissions now uniformly go through the lock → deduplicate → admitNext → executeOutsideLock path.

fluss-server/src/main/java/org/apache/fluss/server/coordinator/TableManager.java
swuferhong: Removed the resume action.

fluss-server/src/main/java/org/apache/fluss/server/coordinator/TableManager.java
swuferhong: There are two call sites:
- Startup (CoordinatorEventProcessor.startup() → tableManager.resumeDeletions())
- Runtime (processDeleteReplicaResponseReceived → tableManager.resumeDeletions())

Why the runtime path never triggers a redundant `onDeletePartition`:Inside `resumePartitionDeletions()`, the guard `isEligibleForDeletion` requires:
```
isPartitionQueuedForDeletion(partition)
    && !isAnyReplicaInState(partition, ReplicaDeletionStarted)
```
During normal operation, after `processDropPartition` → `onDeletePartition` fires, all replicas transition to ReplicaDeletionStarted. As deletion responses come back:
- Some replicas move to DeletionSuccessful, but remaining ones are still in DeletionStarted
- So isEligibleForDeletion = false → submitPartitionDropForResume is never called
- Eventually all replicas reach DeletionSuccessful → takes the completeDeletePartition branch directly

In other words, at runtime resumeDeletions() only ever hits the "all done → complete" branch, never the "re-trigger" branch. So, the `isEligibleForDeletion` guard ensures `onDeletePartition` is only re-invoked in the restart recovery scenario (This case is what we want to throtte). During normal runtime, it's a no-op branch that's never reached.

swuferhong:
@wuchong comments addressed.

wuchong (approved):
I appended a commit to improve the code a bit, please take another look.

swuferhong:
LGTM +1

PR #3177, opened 2026-04-23
[server] Remove CDC log metrics from PK table write path to avoid dou…

### LINKED ISSUE #3176: PK table messagesIn/bytesIn metrics double-counted due to CDC log
### Search before asking

### Motivation

For primary key tables, every write to putToLocalKv increments the table-level and server-level `messagesInPerSecond` and `bytesInPerSecond` metrics twice: once for the KV store write, and once for the internal CDC log append.

This causes the reported throughput to be roughly 2x the actual write volume, which is confusing for users monitoring PK table performance — the metric no longer reflects the number of records the user actually wrote.

The CDC log is an internal implementation detail of PK tables (used for changelog replication). Its I/O should not be aggregated into the user-facing table-level metrics, as users have no awareness of or control over it.

### Solution

_No response_

### Anything else?

_No response_

### Willingness to contribute

swuferhong:
…ble counting

### Purpose

Linked issue: close #3176 

For primary key tables, `putToLocalKv()` was incrementing both KV and CDC
log metrics (messagesIn, bytesIn), causing the table-level throughput to
report roughly 2x the actual write volume.

The CDC log is an internal mechanism for changelog replication, and its
metrics should not be counted in the user-facing table-level aggregation,
as this is confusing for users monitoring PK table performance.

### DIFF
--- a/fluss-server/src/main/java/org/apache/fluss/server/replica/ReplicaManager.java
+++ b/fluss-server/src/main/java/org/apache/fluss/server/replica/ReplicaManager.java
@@ -1301,9 +1301,6 @@ public class ReplicaManager implements ServerReconfigurable {
                 // metric for kv
                 tableMetrics.incKvMessageIn(entry.getValue().getRecordCount());
                 tableMetrics.incKvBytesIn(entry.getValue().sizeInBytes());
-                // metric for cdc log of kv
-                tableMetrics.incLogBytesIn(appendInfo.validBytes());
-                tableMetrics.incLogMessageIn(appendInfo.numMessages());
             } catch (Exception e) {
                 if (isUnexpectedException(e)) {
                     LOG.error("Error put records to local kv on replica {}", tb, e);


binary-signal:
The metrics are tracking different things intentionally:                                                                                                                                                                                               lines 1301-1306:                                                                                                                                                                                                                                                              
```java
  // metric for kv                                                                                                                                                                                                                                                                      
  tableMetrics.incKvMessageIn(entry.getValue().getRecordCount());  // KV write volume                                                                                                                                                                                                   
  tableMetrics.incKvBytesIn(entry.getValue().sizeInBytes());        // KV write bytes
  // metric for cdc log of kv                                                                                                                                                                                                                                                           
  tableMetrics.incLogBytesIn(appendInfo.validBytes());              // CDC log volume                                                                                                                                                                                                   
  tableMetrics.incLogMessageIn(appendInfo.numMessages());           // CDC log messages                                                                                                                                                                                                 
```                                                                                                                                                                                                                                                                                        
These are separate metric counters: KvBytesIn vs LogBytesIn, KvMessageIn vs LogMessageIn. They're not double-counting into the same metric. The CDC log is real I/O that actually happens which it gets written, replicated to followers, and consumed by downstream (Flink connectors, lake tiering etc). 

Hiding that from metrics means:                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                
  1. Replication monitoring breaks - followers fetch from the CDC log. If you don't track its volume, you can't reason about replication lag or bandwidth.                                                                                                                              
  2. Storage accounting becomes wrong, the CDC log consumes real disk space and network. Omitting it from metrics hides actual resource usage.
                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                              
                                                                                                                                                                                                                                                                                        
Compare with appendToLocalLog() at line 1254 — for log-only tables, LogBytesIn tracks the user-facing write. For PK tables, KvBytesIn tracks the user-facing write, and LogBytesIn tracks the internal CDC consequence. Both are useful; they just need to be interpreted differently.

PR #3190, opened 2026-04-24
[tiering] Fix Paimon IOManager leak in MergeTreeWriter

luoyuxia:
- [x] Yes (Claude Code - Claude Opus 4.6)

### Purpose

Fix Paimon IOManager resource leak in `MergeTreeWriter` that causes `paimon-io-<UUID>` temp directories to accumulate in `/tmp` and eventually exhaust disk space when tiering service is running.

### Brief change log

The `IOManager` created in `MergeTreeWriter.createTableWrite()` was passed inline to `withIOManager()` without keeping a reference. Since Paimon's `TableWriteImpl` does not own the IOManager lifecycle, the IOManager was never closed when the writer was closed, leaving `paimon-io-<UUID>` directories behind in `/tmp` after each tiering task.

- Store the `IOManager` as a field in `MergeTreeWriter`
- Override `close()` to explicitly close the `IOManager` in a finally block, ensuring temp directories are cleaned up even if `tableWrite.close()` throws

### Tests

- Existing tiering-related integration tests cover the write/close flow
- Manual verification: start tiering service with PK tables and confirm `paimon-io-*` directories in `/tmp` are cleaned up after writer closes

### API and Format

### Documentation

No documentation changes needed.

### DIFF
--- a/fluss-lake/fluss-lake-paimon/src/main/java/org/apache/fluss/lake/paimon/tiering/mergetree/MergeTreeWriter.java
+++ b/fluss-lake/fluss-lake-paimon/src/main/java/org/apache/fluss/lake/paimon/tiering/mergetree/MergeTreeWriter.java
@@ -46,23 +46,42 @@ public class MergeTreeWriter extends RecordWriter<KeyValue> {
 
     private final RowKeyExtractor rowKeyExtractor;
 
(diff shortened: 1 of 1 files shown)


beryllw:
LGTM

PR #3193, opened 2026-04-24
[tiering] Fix Paimon IOManager leak in MergeTreeWriter causing /tmp disk exhaustion

luoyuxia:
The IOManager created in MergeTreeWriter.createTableWrite() was passed inline to withIOManager() without keeping a reference. Since Paimon's TableWriteImpl does not own the IOManager lifecycle, it was never closed, leaving paimon-io-<UUID> directories in /tmp after each tiering task.

This fix stores the IOManager as a field and explicitly closes it in close() via a finally block, ensuring temp directories are cleaned up.

### Purpose

Linked issue: close #xxx


zuston:
We also found this problem due to the default `/tmp` dir, this will make the `/` out of capacity. @luoyuxia I want to respect the option that is exposed to the fluss option.

luoyuxia:
> We also found this problem due to the default `/tmp` dir, this will make the `/` out of capacity. @luoyuxia I want to respect the option that is exposed to the fluss option.

Please feel free to open a pull request

PR #3194, opened 2026-04-24
[server] make RemoteLogManifest sorted with startOffset

zuston:
### Purpose

Without this PR, the segment metadata in the manifest file is not sorted, which makes it difficult to verify multiple offsets one by one.


zuston:
cc @loserwang1024 @luoyuxia A minor change for better readability

fresh-borzoni (approved):
@zuston Ty for the PR, rebased and checked that CI passes, LGTM

PR #3208, opened 2026-04-25
[flink] Fix batch query on empty datalake-enabled table to return 0 rows instead of failing

### LINKED ISSUE #3207: [flink] Batch query on empty datalake-enabled table should return 0 rows instead of failing with UnsupportedOperationException
### Search before asking

### Fluss version

main (development)

### Please describe the bug 🐞

When a Fluss table has `table.datalake.enabled = true` but is empty (no data written yet, so no lake snapshot has been tiered), running a batch query via Flink SQL fails with:

```text
 java.lang.UnsupportedOperationException: Currently, Batch mode can only be supported if one lake snapshot exists for the table.
```

### Solution

The expected behavior is to return an empty result set (0 rows). The Spark connector already handles this correctly by falling back gracefully.

matrixsparse:
### Summary
- When a datalake-enabled table is empty (no lake snapshot), batch query via Flink SQL
  throws UnsupportedOperationException. This PR makes it return empty result
  (0 rows) instead, consistent with Spark connector behavior.

### Root Cause
In `FlinkSourceEnumerator.startInBatchMode()`, `generateHybridLakeFlussSplits()`
returns null when no lake snapshot exists. The stream mode code in the same
commit (79379968) correctly falls back to `initNonPartitionedSplits()`, but the
batch mode path was left to throw an exception.

### Change
- `FlinkSourceEnumerator.java`: Return `Collections.emptyList()` instead of
  throwing `UnsupportedOperationException` when splits is null.

### Verification
- Ran `mvnw test -pl fluss-flink/fluss-flink-common -Dtest="FlinkSourceEnumerator*"`
- All existing tests pass

Fixes #3207

### DIFF
--- a/fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/source/enumerator/FlinkSourceEnumerator.java
+++ b/fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/source/enumerator/FlinkSourceEnumerator.java
@@ -361,9 +361,22 @@ public class FlinkSourceEnumerator
             context.callAsync(
                     () -> {
                         List<SourceSplitBase> splits = generateHybridLakeFlussSplits();
+                        // No lake snapshot exists, fall back to Fluss-only splits
                         if (splits == null) {
-                            throw new UnsupportedOperationException(
-                                    "Currently, Batch mode can only be supported if one lake snapshot exists for the table.");
+                            if (isPartitioned) {
+                                Set<PartitionInfo> partitionInfos = listPartitions();
+                                Collection<Partition> partitions =
+                                        partitionInfos.stream()
+                                                .map(
+                                                        p ->
+                                                                new Partition(
+                                                                        p.getPartitionId(),
+                                                                        p.getPartitionName()))
+                                                .collect(Collectors.toList());
+                                splits = this.initPartitionedSplits(partitions);
+                            } else {
+                                splits = this.initNonPartitionedSplits();
+                            }
                         }
                         return splits;
                     },
--- a/fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/source/enumerator/FlinkSourceEnumerator.java
+++ b/fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/source/enumerator/FlinkSourceEnumerator.java
@@ -363,6 +363,10 @@ public class FlinkSourceEnumerator
                         List<SourceSplitBase> splits = generateHybridLakeFlussSplits();
                         // No lake snapshot exists, fall back to Fluss-only splits
                         if (splits == null) {
+                            LOG.info(
+                                    "No lake snapshot found for table {},"
+                                            + " falling back to Fluss-only splits.",
+                                    tablePath);
                             if (isPartitioned) {
                                 Set<PartitionInfo> partitionInfos = listPartitions();
                                 Collection<Partition> partitions =


binary-signal (approved):
LGMT  +1

matrixsparse:
Thanks for the review!

binary-signal:
@fresh-borzoni @luoyuxia PTAL

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/source/enumerator/FlinkSourceEnumerator.java
```diff
@@ -362,8 +362,7 @@ private void startInBatchMode() {
                     () -> {
                         List<SourceSplitBase> splits = generateHybridLakeFlussSplits();
                         if (splits == null) {
-                            throw new UnsupportedOperationException(
-                                    "Currently, Batch mode can only be supported if one lake snapshot exists for the table.");
+                            return Collections.emptyList();
```
luoyuxia: Why return empty list? Empty list will cause flink return empty result although actaully there are still records in fluss.
One way is that we can still generate only fluss split although lake snapshot doesn't exist. 
Just like spark does in FlussLakeAppendBatch#doPlan

luoyuxia (commented):
@matrixsparse Thanks for the pr. Left one comment. PTAL

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/source/enumerator/FlinkSourceEnumerator.java
matrixsparse: Thanks for the feedback. This has been updated: rather than returning an empty list, the logic now falls back to Fluss-only split generation, consistent with Spark's planFallbackPartitions(). Partition validation is also added for partitioned tables. @luoyuxia PTAL, thanks!

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/source/enumerator/FlinkSourceEnumerator.java
```diff
@@ -361,9 +361,22 @@ private void startInBatchMode() {
             context.callAsync(
                     () -> {
                         List<SourceSplitBase> splits = generateHybridLakeFlussSplits();
+                        // No lake snapshot exists, fall back to Fluss-only splits
                         if (splits == null) {
```
fresh-borzoni, line 365: nit: maybe it's a good idea to add some info log for easier tracing, at least stream-mode has it when it goes through the lake path: https://github.com/apache/fluss/blob/53bdba8535641e3ce9cd430bd09eb93cf9148422/fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/source/enumerator/FlinkSourceEnumerator.java#L315

smth like:
```
LOG.info("No lake snapshot found for table {}, falling back to Fluss-only splits.", tablePath);
```

fresh-borzoni (commented):
@matrixsparse Ty for the PR, LGTM 👍

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/source/enumerator/FlinkSourceEnumerator.java
matrixsparse: Good suggestion! Added the info log for tracing. Thanks @fresh-borzoni!

fresh-borzoni (approved):
Ty, +1

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/source/enumerator/FlinkSourceEnumerator.java
```diff
@@ -361,9 +361,26 @@ private void startInBatchMode() {
...
+                                Set<PartitionInfo> partitionInfos = listPartitions();
+                                Collection<Partition> partitions =
+                                        partitionInfos.stream()
+                                                .map(
+                                                        p ->
+                                                                new Partition(
+                                                                        p.getPartitionId(),
+                                                                        p.getPartitionName()))
+                                                .collect(Collectors.toList());
+                                splits = this.initPartitionedSplits(partitions);
```
luoyuxia, line 380: The issue here is that for primary-key tables, this fallback may generate both kv snapshot splits and log splits. At the moment, the Flink connector does not seem to support merging kv snapshot splits with log splits yet.
In comparison, the Spark connector only generates log splits in this case, with the lake splits left empty, so it can continue to reuse the existing path that already supports merging lake splits with log splits.

@matrixsparse  Would you be open to submitting a follow-up PR to address this? If that is not convenient at the moment, it may be safer to revert this change for now and reopen the issue so we can revisit it with a complete fix later.
cc @fresh-borzoni

matrixsparse: Understood. I will submit a follow-up PR to fix this limitation in the Flink connector. Thanks for pointing this out @luoyuxia

fresh-borzoni: @luoyuxia Good point, thank you, I missed it

PR #3212, opened 2026-04-27
[metrics] Add filter-label-value-characters option for Prometheus reporter

### LINKED ISSUE #3211: Support UTF-8 characters in Prometheus label values
### Search before asking

### Fluss version

0.9.0 (latest release)

### Please describe the bug 🐞

When using Fluss with Prometheus reporter, label values containing non-ASCII characters (e.g. Chinese table names, hyphens) are replaced with underscores because AbstractPrometheusReporter applies CHARACTER_FILTER (regex [^a-zA-Z0-9:_]) to both metric/label names and label values.
However, per the Prometheus data model, label names are restricted to [a-zA-Z0-9_]*, but label values may contain any Unicode characters. The current behavior makes it impossible to preserve hyphens or Chinese characters in label values, which is a valid use case.
This was also reported and fixed in Flink as [FLINK-11174](https://issues.apache.org/jira/browse/FLINK-11174) / [PR #7340](https://github.com/apache/flink/pull/7340).

### Solution

Add a metrics.reporter.prometheus.filter-label-value-characters config option (default true for backward compatibility). When set to false, label values are passed through without character filtering.

beryllw:
### Purpose

Linked issue: close #3211
Label values in Prometheus support UTF-8 characters per the specification. Remove the unnecessary character filtering that replaces non-[a-zA-Z0-9:_] characters with underscores in label values.

### Brief change log

- Remove `filter-label-value-characters` config option
- Add test verifying hyphen is preserved in label values

### DIFF
--- a/fluss-common/src/main/java/org/apache/fluss/config/ConfigOptions.java
+++ b/fluss-common/src/main/java/org/apache/fluss/config/ConfigOptions.java
@@ -1963,6 +1963,18 @@ public class ConfigOptions {
                                     + "the CoordinatorServer) it is advisable to use a port range "
                                     + "like 9250-9260.");
 
+    public static final ConfigOption<Boolean>
+            METRICS_REPORTER_PROMETHEUS_FILTER_LABEL_VALUE_CHARACTERS =
+                    key("metrics.reporter.prometheus.filter-label-value-characters")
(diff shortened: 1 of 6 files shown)


luoyuxia (approved):
+1

PR #3222, opened 2026-04-28
[server] Respect the option of `table.log.ttl` for local segments

zuston:
### Purpose  

Enable TTL-expired local log segments to be rolled (bypass the limitation of num segment retained), while preserving a new empty active segment for future appends.

This PR is to reduce the partitioned log table bucket local storage capacity, especially longterm day level partition log table

### Brief change log  
- Added active-segment TTL roll logic in `LogTablet`.
- Trigger active-segment TTL roll before remote tiering selects upload candidates.
- Added coverage for rolling, uploading, local cleanup, and subsequent appends after TTL.


luoyuxia:
cc @swuferhong

zuston:
could you help review this? @swuferhong

zuston:
gently ping @luoyuxia @swuferhong

fluss-server/src/main/java/org/apache/fluss/server/log/LogTablet.java
```diff
@@ -1196,17 +1259,27 @@ private List<LogSegment> deletableSegments(long endOffset) {
         // readers is in progress.
         List<LogSegment> deletableSegments = new ArrayList<>();
         List<LogSegment> logSegments = localLog.getSegments().values();
-        // ignore the segments configured to be retained
-        for (int i = 0; i < logSegments.size() - tieredLogLocalSegments; i++) {
-            if (logSegments.get(i + 1).getBaseOffset() <= endOffset) {
-                deletableSegments.add(logSegments.get(i));
-            } else {
+        int tierProtectedStartIndex = logSegments.size() - tieredLogLocalSegments;
+        long now = clock.milliseconds();
```
gyang94: We have three different time during `rolling active segment`, `delete local segments` and `delete remote segments`. Will it be better to froze a timestamp at the beginning of LogTieringTask.run() ?
Then we use the same timestamp value in each step of this LogTieringTask round.

zuston: sounds good. let me take a look tomorrow.

fluss-server/src/main/java/org/apache/fluss/server/log/LogTablet.java
```diff
@@ -1196,17 +1259,27 @@ private List<LogSegment> deletableSegments(long endOffset) {
...
         }
         return deletableSegments;
     }
 
+    private boolean isSegmentExpired(long now, LogSegment segment) throws IOException {
+        if (logTtlMs <= 0L) {
+            return false;
+        }
+        return now - segment.maxTimestampSoFar() > logTtlMs;
+    }
```
gyang94, line 1307: could we introduce a new config for rolling active segment? logTTL may be too long for rolling active segment. 
Similar to kafka's `segment.ms` config, generally its value should be less smaller than logTtl.

fluss-common/src/main/java/org/apache/fluss/config/ConfigOptions.java
```diff
@@ -1549,6 +1549,16 @@ public class ConfigOptions {
                                     + "we will retain a log before we will delete old segments to free up "
                                     + "space. If set to -1, the log will not be deleted.");
 
+    public static final ConfigOption<Duration> TABLE_LOG_SEGMENT_ROLL_MS =
+            key("table.log.segment.roll-ms")
```
gyang94, line 1553: It is a duration type config, not good to name it with `ms`. 
Maybe rename it with `roll-threshold` ,`roll-time`, or `active-roll-threshold` ... etc?

fluss-common/src/main/java/org/apache/fluss/config/ConfigOptions.java
```diff
@@ -1549,6 +1549,16 @@ public class ConfigOptions {
...
+                    .durationType()
+                    .defaultValue(Duration.ofMillis(-1))
```
gyang94, line 1555: We should make this config alterable (can be dynamically changed). Otherwise existing tables will always be -1.

fluss-server/src/main/java/org/apache/fluss/server/log/remote/LogTieringTask.java
```diff
@@ -127,14 +127,15 @@ private void runOnce() throws InterruptedException {
             TableMetricGroup metricGroup = replica.tableMetrics();
             maybeUpdateCopiedOffset(logTablet);
 
+            logTablet.rollActiveSegmentIfExpired(currentTimeMs);
```
swuferhong, line 130: I think this is a very dangerous operation.  `rollActiveSegmentIfExpired` is only called by the leader’s `LogTieringTask`, so the new segment boundary is not propagated to followers. For example, the leader may roll and upload `[A, B)`, while a follower continues to keep `[A, C)` as a single segment. If that follower later becomes the leader, `LogSegments.values()` may select and upload the whole [A, C) segment, overlapping the existing remote segment.

`RemoteLogTablet` does not safely support this overlap: both segments remain in the ID/manifest maps, while the start-offset map retains only one UUID. When `[A, B)` expires, removing its start-offset entry may also remove the index for `[A, C)`, and firstKey() may subsequently fail.

Could we either keep segment boundaries consistent across replicas or make the remote metadata overlap/leader-epoch aware?

fluss-server/src/main/java/org/apache/fluss/server/log/LogTablet.java
```diff
@@ -554,11 +597,15 @@ public void updateRemoteLogSize(long remoteLogSize) {
...
+        updateRemoteLogEndOffset(remoteLogEndOffset, clock.milliseconds());
+    }
+
+    public void updateRemoteLogEndOffset(long remoteLogEndOffset, long currentTimeMs) {
         if (remoteLogEndOffset > this.remoteLogEndOffset) {
             this.remoteLogEndOffset = remoteLogEndOffset;
 
             // try to delete these segments already exist in remote storage.
-            deleteSegmentsAlreadyExistsInRemote();
+            deleteSegmentsAlreadyExistsInRemote(currentTimeMs);
```
swuferhong, line 608: Local cleanup is only invoked when `remoteLogEndOffset` increases. During a remote-TTL-only update, however, the manifest end offset normally stays unchanged, decreases, or becomes -1, so this branch is skipped.

Consequently, inactive local segments that were protected by `table.log.tiered.local-segments` when uploaded can remain locally readable indefinitely after their TTL has expired.

The current test uploads another segment after advancing the clock, which increases the remote end offset and masks this path. 

Could cleanup be evaluated on every retention pass, independently of whether the remote end offset advanced? Please also add a test with no newly uploaded segment and assert the local segment count/start offset after TTL expiration.

swuferhong (commented):
Hi, @zuston Thanks for your contribution, this is an very useful feature. I left two questions that I feel need clarification：

fluss-server/src/main/java/org/apache/fluss/server/log/remote/LogTieringTask.java
zuston: Thanks for pointing out this potential issue, @swuferhong.

In Kafka’s tiered storage implementation, overlapping segment ranges are handled at the remote metadata layer using leader epochs. https://github.com/apache/kafka/blob/trunk/storage/src/main/java/org/apache/kafka/server/log/remote/storage/RemoteLogManager.java#L902

For Fluss, I’m also exploring an alternative approach: propagating roll-boundary markers from the leader to followers so that replicas maintain consistent segment boundaries.

What do you think of this approach, @swuferhong?

wuchong:
Continue in https://github.com/apache/fluss/pull/3623

PR #3227, opened 2026-04-28
[spark] Fix empty projection causing Invalid metadata length for COUNT(*)/COUNT(1)

### LINKED ISSUE #2724: [spark] Empty projection in batch read throws Invalid metadata length with COUNT(*)/COUNT(1)
### Search before asking

### Fluss version

main (development)

### Please describe the bug 🐞

When executing aggregate queries like COUNT(*)/COUNT(1) on a Fluss table via Spark SQL, the server throws IllegalStateException: Invalid metadata length at FileLogProjection.project(). This is likely caused by Spark pushing down an empty column projection for count-only queries. The LogFetcher retries indefinitely, making the query hang.

How to Reproduce:
```
CREATE TABLE log_orders (
    order_id BIGINT,
    item_id BIGINT,
    amount INT,
    address STRING
);

INSERT INTO log_orders VALUES
    (1, 101, 10, 'Beijing'),
    (2, 102, 20, 'Shanghai'),
    (3, 103, 30, 'Hangzhou'),
    (4, 104, 40, 'Shenzhen'),
    (5, 105, 50, 'Guangzhou');

SELECT COUNT(*) FROM log_orders;
```
Stack Trace:
```
26/02/25 12:31:07 ERROR LogFetcher: Failed to fetch log from node 0 for bucket TableBucket{tableId=0, bucket=0}
org.apache.fluss.exception.UnknownServerException: The server experienced an unexpected error when processing the request.
26/02/25 12:31:07 WARN LogFetchCollector: Unknown server error while fetching offset -2 for bucket TableBucket{tableId=0, bucket=0}: org.apache.fluss.exception.UnknownServerException: java.lang.IllegalStateException: Invalid metadata length
        at org.apache.fluss.utils.Preconditions.checkState(Preconditions.java:161)
        at org.apache.fluss.record.FileLogProjection.project(FileLogProjection.java:232)
        at org.apache.fluss.server.log.LogSegment.read(LogSegment.java:534)
        at org.apache.fluss.server.log.LocalLog.read(LocalLog.java:409)
        at org.apache.fluss.server.log.LogTablet.read(LogTablet.java:413)
        at org.apache.fluss.server.replica.Replica.readRecords(Replica.java:1498)
        at org.apache.fluss.server.replica.Replica.lambda$fetchRecords$11(Replica.java:1054)
        at org.apache.fluss.utils.concurrent.LockUtils.inLock(LockUtils.java:42)
        at org.apache.fluss.utils.concurrent.LockUtils.inReadLock(LockUtils.java:55)
        at org.apache.fluss.server.replica.Replica.fetchRecords(Replica.java:1050)
        at org.apache.fluss.server.replica.ReplicaManager.readFromLog(ReplicaManager.java:1358)
        at org.apache.fluss.server.replica.ReplicaManager.fetchLogRecords(ReplicaManager.java:562)
        at org.apache.fluss.server.tablet.TabletService.fetchLog(TabletService.java:200)
        at jdk.internal.reflect.GeneratedMethodAccessor4.invoke(Unknown Source)
        at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(Unknown Source)
        at java.base/java.lang.reflect.Method.invoke(Unknown Source)
        at org.apache.fluss.rpc.netty.server.FlussRequestHandler.processRequest(FlussRequestHandler.java:63)
        at org.apache.fluss.rpc.netty.server.FlussRequestHandler.processRequest(FlussRequestHandler.java:34)
        at org.apache.fluss.rpc.netty.server.RequestProcessor.processRequest(RequestProcessor.java:98)
        at org.apache.fluss.rpc.netty.server.RequestProcessor.run(RequestProcessor.java:70)
        at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(Unknown Source)
        at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(Unknown Source)
        at java.base/java.lang.Thread.run(Unknown Source)
```

Possible Cause: Spark optimizes COUNT(*) / COUNT(1) by projecting zero columns. FileLogProjection.project() does not handle this empty projection case, leading to invalid metadata length during log segment reads.

### Solution

_No response_

Kaixuan-Duan:
### Purpose

Linked issue: close #2724 

When Spark pushes down an empty column projection for `COUNT(*)`/`COUNT(1)` queries, the Fluss server fails with `IllegalStateException("Invalid metadata length")` in `FileLogProjection.project()`, causing the client to retry indefinitely and the query to hang.

This PR fixes the issue from two sides:

Server side: reject empty projection early with a clear `InvalidColumnProjectionException` instead of crashing with an internal error.
Spark connector side: fall back to projecting the first column when Spark pushes down an empty projection, so the row count is preserved without fetching unnecessary data.

### Brief change log

- `FileLogProjection#setCurrentProjection`: add a guard that throws InvalidColumnProjectionException when selectedFieldPositions is empty.

- `FileLogProjectionTest`: add testEmptyProjectionRejectsWithClearError to verify the server-side guard.
- `FlussBatch#projection` / `FlussMicroBatchStream#projection`: when readSchema yields an empty projection, fall back to Array(0) (first column).
- `SparkLogTableReadTest`: add `COUNT(*)` and `COUNT(1)` end-to-end tests for log tables.
- `SparkPrimaryKeyTableReadTest`: add `COUNT(*)` end-to-end test for primary key tables.

### Tests
./mvnw -pl fluss-common -DskipTests=false -Dtest=FileLogProjectionTest#testEmptyProjectionRejectsWithClearError test

./mvnw -pl fluss-spark/fluss-spark-ut -am install -DskipTests
./mvnw -pl fluss-spark/fluss-spark-ut -Dsuites='org.apache.fluss.spark.SparkLogTableReadTest' test
./mvnw -pl fluss-spark/fluss-spark-ut -Dsuites='org.apache.fluss.spark.SparkPrimaryKeyTableReadTest' test

### Documentation

### DIFF
--- a/fluss-common/src/main/java/org/apache/fluss/record/FileLogProjection.java
+++ b/fluss-common/src/main/java/org/apache/fluss/record/FileLogProjection.java
@@ -121,6 +121,18 @@ public class FileLogProjection {
             SchemaGetter schemaGetter,
             ArrowCompressionInfo compressionInfo,
             int[] selectedFieldPositions) {
+        // Empty projection is currently not supported on the server side: the Arrow metadata length
+        // estimate disagrees with the serialized length for a zero-field schema, which would
+        // cause the client to retry indefinitely. Fail fast with a clear, non-retriable error
+        // instead.
+        if (selectedFieldPositions != null && selectedFieldPositions.length == 0) {
+            throw new InvalidColumnProjectionException(
+                    "Empty projection is not supported. "
+                            + "Please project at least one column. "
+                            + "For aggregations like COUNT(*) / COUNT(1), "
+                            + "the connector should push down a projection that selects "
+                            + "at least one column.");
+        }
         this.tableId = tableId;
         this.schemaGetter = schemaGetter;
         this.compressionInfo = compressionInfo;
--- a/fluss-spark/fluss-spark-common/src/main/scala/org/apache/fluss/spark/read/FlussBatch.scala
+++ b/fluss-spark/fluss-spark-common/src/main/scala/org/apache/fluss/spark/read/FlussBatch.scala
@@ -26,8 +26,6 @@ import org.apache.spark.sql.connector.read.{Batch, InputPartition, PartitionRead
 import org.apache.spark.sql.types.StructType
 import org.apache.spark.sql.util.CaseInsensitiveStringMap
 
-import scala.collection.JavaConverters._
-
 /**
  * Base class for planner-backed batch scans. The planner (constructed at ScanBuilder-time) opens
  * and releases the Fluss client Connection/Admin itself — scoped to its `plan()` call — and
@@ -47,18 +45,8 @@ abstract class FlussBatch(
     planner: SplitPlanner)
   extends Batch {
 
-  protected def projection: Array[Int] = {
-    val columnNameToIndex = tableInfo.getSchema.getColumnNames.asScala.zipWithIndex.toMap
-    val projected = readSchema.fields.map {
-      field =>
-        columnNameToIndex.getOrElse(
-          field.name,
-          throw new IllegalArgumentException(s"Invalid field name: ${field.name}"))
-    }
-    // The Fluss server does not currently support empty Arrow projections. Fall back to projecting
-    // the first column so the row count is preserved while still avoiding fetching unnecessary columns.
-    if (projected.isEmpty) Array(0) else projected
-  }
+  protected def projection: Array[Int] =
+    FlussScanBuilder.projectionOf(tableInfo, Some(readSchema))
 
   override def planInputPartitions(): Array[InputPartition] = planner.plan()
 }
--- a/fluss-spark/fluss-spark-common/src/main/scala/org/apache/fluss/spark/read/FlussBatch.scala
+++ b/fluss-spark/fluss-spark-common/src/main/scala/org/apache/fluss/spark/read/FlussBatch.scala
@@ -49,12 +49,15 @@ abstract class FlussBatch(
 
   protected def projection: Array[Int] = {
     val columnNameToIndex = tableInfo.getSchema.getColumnNames.asScala.zipWithIndex.toMap
-    readSchema.fields.map {
+    val projected = readSchema.fields.map {
       field =>
         columnNameToIndex.getOrElse(
           field.name,
           throw new IllegalArgumentException(s"Invalid field name: ${field.name}"))
     }
+    // The Fluss server does not currently support empty Arrow projections. Fall back to projecting
+    // the first column so the row count is preserved while still avoiding fetching unnecessary columns.
+    if (projected.isEmpty) Array(0) else projected
   }
 
   override def planInputPartitions(): Array[InputPartition] = planner.plan()
--- a/fluss-spark/fluss-spark-ut/src/test/scala/org/apache/fluss/spark/SparkLogTableReadTest.scala
+++ b/fluss-spark/fluss-spark-ut/src/test/scala/org/apache/fluss/spark/SparkLogTableReadTest.scala
@@ -92,7 +92,7 @@ class SparkLogTableReadTest extends FlussSparkTestBase {
              |(1200L, 270L, 607, NULL)
              |""".stripMargin)
 
-      // COUNT(*) / COUNT(1) — empty projection case (regression test for #2724)
+      // COUNT(*) / COUNT(1) — empty projection case
       checkAnswer(sql(s"SELECT COUNT(*) FROM $DEFAULT_DATABASE.t"), Row(9L) :: Nil)
       checkAnswer(sql(s"SELECT COUNT(1) FROM $DEFAULT_DATABASE.t"), Row(9L) :: Nil)
 
--- a/fluss-spark/fluss-spark-ut/src/test/scala/org/apache/fluss/spark/SparkLogTableReadTest.scala
+++ b/fluss-spark/fluss-spark-ut/src/test/scala/org/apache/fluss/spark/SparkLogTableReadTest.scala
@@ -84,6 +84,30 @@ class SparkLogTableReadTest extends FlussSparkTestBase {
           Row(1000L, 25L) ::
           Row(1100L, 260L) :: Nil
       )
+
+      // Insert one row with a NULL nullable column to make the COUNT(nullable_col)
+      // assertion below actually discriminate between row count and non-null count.
+      sql(s"""
+             |INSERT INTO $DEFAULT_DATABASE.t VALUES
+             |(1200L, 270L, 607, NULL)
+             |""".stripMargin)
+
+      // COUNT(*) / COUNT(1) — empty projection case (regression test for #2724)
+      checkAnswer(sql(s"SELECT COUNT(*) FROM $DEFAULT_DATABASE.t"), Row(9L) :: Nil)
+      checkAnswer(sql(s"SELECT COUNT(1) FROM $DEFAULT_DATABASE.t"), Row(9L) :: Nil)
+
+      // COUNT(*) with filter — verifies the non-empty-projection path is unaffected
+      checkAnswer(
+        sql(s"SELECT COUNT(*) FROM $DEFAULT_DATABASE.t WHERE orderId >= 900"),
+        Row(5L) :: Nil
+      )
+
+      // COUNT on a nullable column — address is nullable STRING and one row has NULL,
+      // so COUNT(address) must be strictly less than COUNT(*).
+      checkAnswer(
+        sql(s"SELECT COUNT(address) FROM $DEFAULT_DATABASE.t"),
+        Row(8L) :: Nil
+      )
     }
   }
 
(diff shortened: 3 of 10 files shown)


fluss-common/src/main/java/org/apache/fluss/record/FileLogProjection.java
```diff
@@ -121,6 +121,19 @@ public void setCurrentProjection(
             SchemaGetter schemaGetter,
             ArrowCompressionInfo compressionInfo,
             int[] selectedFieldPositions) {
+        // Empty projection (selectedFieldPositions.length == 0) is currently not supported on the
```
Yohahaha: It would be good to also verify this behavior and the fix in the Flink connector if needed.

luoyuxia:
@Kaixuan-Duan Hi, seems is it same with #2725. cc @beryllw

fluss-spark/fluss-spark-ut/src/test/scala/org/apache/fluss/spark/SparkLogTableReadTest.scala
```diff
@@ -451,6 +451,20 @@ class SparkLogTableReadTest extends FlussSparkTestBase {
     }
   }
 
+  test("Spark Read: COUNT(*) without filter") {
```
Yohahaha: What happens when there is a filter?

Kaixuan-Duan:
@luoyuxia I’m sorry. It was my fault. I didn't notice that there was already a PR pointing to issue#2724

fluss-common/src/main/java/org/apache/fluss/record/FileLogProjection.java
Kaixuan-Duan: Looked into the Flink connector: it doesn't implement SupportsAggregatePushDown, and Flink's planner doesn't push an empty projection for COUNT(*) either, so this bug is not reproducible there today. The new server-side guard is connector-agnostic, so it still protects Flink/Kafka/direct API users if any caller ever pushes an empty projection in the future.

fluss-spark/fluss-spark-ut/src/test/scala/org/apache/fluss/spark/SparkLogTableReadTest.scala
Kaixuan-Duan: COUNT(*) with filter will not traverse an empty projection path. The original behavior is correct, and this PR does not affect it.

beryllw:
I've closed PR #2725.Please go ahead.
Also, is it possible to use the pre-computed table statistics directly when dealing with PK tables, instead of scanning the data and counting afterward?

fluss-spark/fluss-spark-ut/src/test/scala/org/apache/fluss/spark/SparkLogTableReadTest.scala
beryllw: Perhaps we could add the COUNT(*) test case to the existing test("Spark Read: log table") rather than introducing a new test.

Additionally, we should test COUNT(*) with filters and nullable column types. It would also be worth testing against partitioned tables.

fluss-spark/fluss-spark-ut/src/test/scala/org/apache/fluss/spark/SparkPrimaryKeyTableReadTest.scala
```diff
@@ -376,4 +376,22 @@ class SparkPrimaryKeyTableReadTest extends FlussSparkTestBase {
       assert(numRowsRead == 3L, s"Expected 3 rows read, got $numRowsRead")
     }
   }
+
+  test("Spark Read: primary key table COUNT(*) without filter") {
```
beryllw: ditto

Kaixuan-Duan:
@beryllw Thanks for the suggestion! 
Done. The Spark connector now implements `SupportsPushDownAggregates` and answers `COUNT(*)` / `COUNT(1)` / `COUNT(non-null col)` (no GROUP BY, no DISTINCT) on PK non-lake tables directly via `Admin.getTableStats().getRowCount()` — no data scan. 
Falls back gracefully on `UnsupportedVersionException` for older clusters. PTAL.

Yohahaha:
> I've closed PR #2725.Please go ahead. Also, is it possible to use the pre-computed table statistics directly when dealing with PK tables, instead of scanning the data and counting afterward?

@beryllw curious could we assume table statistics always accurate in fluss?

beryllw:
> @beryllw curious could we assume table statistics always accurate in fluss?

I think we could assume they're always accurate in Fluss — for KV tables it's backed by rowCount, and for log tables it's derived from getHighWatermark() - logStartOffset().

https://github.com/apache/fluss/blob/d4cd1a2771ff254a81877296a5cabfa966d102e2/fluss-server/src/main/java/org/apache/fluss/server/log/LogTablet.java#L248-L250

https://github.com/apache/fluss/blob/c6a1c2dbfe36aa8b503cff04470acf2951ef9e51/fluss-server/src/main/java/org/apache/fluss/server/kv/KvTablet.java#L307-L318

luoyuxia:
@Kaixuan-Duan Thanks for fixing the empty projection issue. My main concern is the newly added Spark `COUNT` pushdown: before merging, I think its behavior should be aligned with the existing Flink count-pushdown behavior.

Flink already has a clear contract in `FlinkTableSource#applyAggregates` / `PushdownUtils.countTable`: only push down a single `COUNT(*)` / `COUNT(1)` / `COUNT(non-null column)` without grouping; do not push down nullable `COUNT(col)`, streaming reads, or lake-enabled reads; and keep the existing row-count-disabled behavior for WAL / legacy PK tables.


I think the empty-projection fix can be kept, but Spark `COUNT` pushdown should either be aligned with Flink behavior, or be split out from this PR.

Kaixuan-Duan:
@luoyuxia Thanks for the suggestion. I've split this PR into a pure bug fix — the aggregation pushdown part has been removed and will be submitted as a follow-up PR.

luoyuxia (approved):
@Kaixuan-Duan Thanks. LGTM. Could you please rebase main branch,
cc @beryllw @Yohahaha @fresh-borzoni Could you please help have a final review

Kaixuan-Duan:
@luoyuxia Thanks，I have rebased the branch onto the main.

Yohahaha (approved):
LGTM!

fluss-spark/fluss-spark-common/src/main/scala/org/apache/fluss/spark/read/FlussBatch.scala
```diff
@@ -49,12 +49,15 @@ abstract class FlussBatch(
...
-    readSchema.fields.map {
+    val projected = readSchema.fields.map {
       field =>
         columnNameToIndex.getOrElse(
           field.name,
           throw new IllegalArgumentException(s"Invalid field name: ${field.name}"))
     }
+    // The Fluss server does not currently support empty Arrow projections. Fall back to projecting
+    // the first column so the row count is preserved while still avoiding fetching unnecessary columns.
+    if (projected.isEmpty) Array(0) else projected
```
fresh-borzoni: projectionOf is the third copy of this name->index resolution (came in with #3559, which pointed only the planners at it), and it still returns empty, which the planners feed to lakeSource.withProject(). 

Shall we put the fallback in projectionOf once and have FlussBatch/FlussMicroBatchStream.projection delegate to it?

fluss-spark/fluss-spark-ut/src/test/scala/org/apache/fluss/spark/SparkPrimaryKeyTableReadTest.scala
```diff
@@ -135,6 +135,32 @@ class SparkPrimaryKeyTableReadTest extends FlussSparkTestBase {
...
       )
+
+      // Insert one row with a NULL nullable column to make the COUNT(nullable_col)
+      // assertion below actually discriminate between row count and non-null count.
+      sql(s"""
+             |INSERT INTO $DEFAULT_DATABASE.t VALUES
+             |(1200L, 270L, 607, NULL)
+             |""".stripMargin)
+
+      // COUNT(*) — empty projection case (regression test for #2724). PK table has 7 rows
```
fresh-borzoni: nit: FlussUpsertPartitionReader always appends pk columns (projectionWithPks), so the PK path never actually sent an empty projection to the server, so  does this test pass without the fix too?

fresh-borzoni (commented):
@Kaixuan-Duan Thank you for the PR, left some minor comments

Also should Projection.of reject empty arrays eagerly? That way it saves direct-API users the RPC round trip, though server guard should stays as the authoritative check still. WDYT?

fluss-spark/fluss-spark-common/src/main/scala/org/apache/fluss/spark/read/FlussBatch.scala
Kaixuan-Duan: Good catch — done. Moved the fallback into FlussScanBuilder.projectionOf and made FlussBatch/FlussMicroBatchStream.projection delegate to it.

fluss-spark/fluss-spark-ut/src/test/scala/org/apache/fluss/spark/SparkPrimaryKeyTableReadTest.scala
Kaixuan-Duan: You're right. Verified locally by disabling the fallback and the test still passed. Removed the redundant PK table tests.

Kaixuan-Duan:
@fresh-borzoni Thanks for the suggestion. Projection.of now rejects empty arrays eagerly. PTAL.

fresh-borzoni (approved):
@Kaixuan-Duan Thank you, LGTM 👍

Issue #3228, opened 2026-04-29
Possible OutOfRangeException when reading the stream log table after resuming from a long interruption


zuston:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/fluss/issues) and found nothing similar.

### Fluss version

0.9.0 (latest release)

### Please describe the bug 🐞

When Flink recovers from a checkpoint after a long interruption, subtasks may throw an OutOfRangeException, as shown in the logs below.

<img width="1919" height="549" alt="Image" src="https://github.com/user-attachments/assets/e85afef6-4681-40f3-99e9-2f5824e43eba" />

I understand this is caused by local tablet logs being offloaded to remote storage, which makes the recovered split offsets stale and leads to an OutOfRange condition.
While this behavior is acceptable for Kafka-style logs, it is not appropriate for data lake–backed log tables. In this case, the job will keep failing indefinitely. We should handle this scenario explicitly to avoid permanent recovery failures.

### Are you willing to submit a PR?

- [x] I'm willing to submit a PR!

zuston:
We see the same failure during long-term backfill, especially when lake tiering is enabled.

### Root cause (two gaps in the current code)

Lake start offset is never updated — it stays at the default Long.MAX_VALUE. As a result, handleFetchOutOfRangeException does not drive canFetchFromLakeLog when the requested offset exists only in the lake tier.

No lake read path on the client — LogFetcher.handleFetchLogResponse cannot fall back to reading from the lake when local log data is missing or out of range.

### Direction

Expose lake metadata to the client so we can implement a proper lake fallback read path.

PR #3229, opened 2026-04-29
[server] Add metrics of log startOffset

zuston:
### Purpose

Linked issue: #3228 .

When the outOfRange exception, I hope I can find the tablet log startOffset from the metric system.

### DIFF
--- a/fluss-server/src/main/java/org/apache/fluss/server/log/remote/RemoteLogTablet.java
+++ b/fluss-server/src/main/java/org/apache/fluss/server/log/remote/RemoteLogTablet.java
@@ -121,6 +121,7 @@ public class RemoteLogTablet {
                     }
                     MetricGroup metricGroup = bucketMetricGroup.addGroup("remoteLog");
                     metricGroup.gauge(MetricNames.LOG_NUM_SEGMENTS, () -> numRemoteLogSegments);
+                    metricGroup.gauge(MetricNames.LOG_START_OFFSET, () -> remoteLogStartOffset);
                     metricGroup.gauge(MetricNames.LOG_END_OFFSET, () -> remoteLogEndOffset);
                     metricGroup.gauge(MetricNames.REMOTE_LOG_SIZE, this::getRemoteSizeInBytes);
                     remoteLogMetrics = metricGroup;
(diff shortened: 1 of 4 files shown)


fluss-server/src/main/java/org/apache/fluss/server/log/remote/RemoteLogTablet.java
```diff
@@ -121,6 +121,7 @@ public void registerMetrics(BucketMetricGroup bucketMetricGroup) {
                     }
                     MetricGroup metricGroup = bucketMetricGroup.addGroup("remoteLog");
                     metricGroup.gauge(MetricNames.LOG_NUM_SEGMENTS, () -> numRemoteLogSegments);
+                    metricGroup.gauge(MetricNames.LOG_START_OFFSET, () -> remoteLogStartOffset);
```
luoyuxia: also update the monitor-metrics doc

luoyuxia: when no remote log, remoteLogStartOffset will be Long.MAX_VALUE, which will be werid, I think maybe we can use -1 for metric report?

luoyuxia (commented):
@zuston Thanks for the pr. Left minor comments. PTAL

fluss-server/src/main/java/org/apache/fluss/server/log/remote/RemoteLogTablet.java
zuston: > when no remote log, remoteLogStartOffset will be Long.MAX_VALUE, which will be werid, I think maybe we can use -1 for metric report?

make sense!

zuston:
> @zuston Thanks for the pr. Left minor comments. PTAL

updated.

fluss-server/src/main/java/org/apache/fluss/server/log/remote/RemoteLogTablet.java
```diff
@@ -121,12 +121,21 @@ public void registerMetrics(BucketMetricGroup bucketMetricGroup) {
                     }
                     MetricGroup metricGroup = bucketMetricGroup.addGroup("remoteLog");
                     metricGroup.gauge(MetricNames.LOG_NUM_SEGMENTS, () -> numRemoteLogSegments);
+                    metricGroup.gauge(MetricNames.LOG_START_OFFSET, this::getRemoteLogStartOffsetForMetric);
```
luoyuxia: nit: how about use lambda to inline it:
```
() -> {
   if (remoteLogStartOffset == INIT_REMOTE_LOG_START_OFFSET) {
            return -1L;
        }
        return remoteLogStartOffset;
}
```

luoyuxia (approved):
+1

PR #3232, opened 2026-04-29
[server] Add request queue size metric per processor

zuston:
### Purpose

Currently, each connection is bound to a dedicated RPC processor. Based on the aggregated RPC queue metrics, we cannot determine whether high latency is caused by a specific processor queue being blocked.
When latency is high, I want to identify whether a particular processor’s queue is congested or stuck.


fresh-borzoni (approved):
@zuston Thank you for the PR, rebased and reran CI, LGTM 👍

zuston:
big thanks @fresh-borzoni

PR #3233, opened 2026-04-29
[server] Support altering table.log.ttl table option

### LINKED ISSUE #3231: Support alter table.log.ttl
### Search before asking

### Motivation

Currently, `table.log.ttl` is not supported to alter. We should support alter it to enable users to adjust the log retention duration after the table is created, otherwise users have to drop and recreate the table to change the log retention, which is inconvenient and unacceptable in production environments.

### Solution

_No response_

### Anything else?

_No response_

### Willingness to contribute

Kaixuan-Duan:
### Purpose

Linked issue: close #3231 

### Brief change log

- Add `ConfigOptions.TABLE_LOG_TTL.key()` to `FlussConfigUtils.ALTERABLE_TABLE_OPTIONS` so the validator no longer rejects ALTER on this option.
- Change `RemoteLogTablet#ttlMs` from `final` to `volatile` and add `getTtlMs()` / `updateTtlMs(long)` so the remote-log expiration check can observe the new value without recreating the tablet.
- Add `Replica#updateLogTtlMs(long)` which looks up the corresponding `RemoteLogTablet` via `RemoteLogManager` and applies the new TTL, guarded by an `IllegalStateException` fallback for the case where the tablet has not been registered yet.
- Extend `ReplicaManager#updateReplicaTableConfig` to collect a `tableIdToLogTtlMs` map from the metadata diff and dispatch `updateLogTtlMs` to each affected replica, mirroring how `table.log.tiered.local-segments` is already handled.
### Tests

./mvnw -pl fluss-client,fluss-server -am test \\
  -Dtest='FlussAdminITCase#testAlterTableLogTtl,'\\
'RemoteLogTabletTest,'\\
'RemoteLogManagerTest,'\\
'RemoteLogTTLTest' \\
  -DfailIfNoTests=false


Kaixuan-Duan:
@luoyuxia I have addressed the feedback. PTAL

luoyuxia:
cc @swuferhong

fluss-server/src/main/java/org/apache/fluss/server/log/remote/RemoteLogManager.java
```diff
@@ -290,6 +290,33 @@ public List<RemoteLogSegment> relevantRemoteLogSegmentsForFetchV0(
...
+     *       initial ttl from {@code Replica.getLogTTLMs()}; whether that reflects the latest ALTER
+     *       depends on the order in which the metadata update arrives at the owning {@code
+     *       ReplicaManager}.
+     * </ul>
+     *
+     * @return the previous ttl in milliseconds, or {@link Optional#empty()} if absent.
+     */
+    public Optional<Long> updateLogTtlMs(TableBucket tableBucket, long newTtlMs) {
+        RemoteLogTablet remoteLogTablet = remoteLogs.get(tableBucket);
+        if (remoteLogTablet == null) {
```
wuchong: When this update reaches a follower, `remoteLogs` has no entry because only leaders register a `RemoteLogTablet`, so the new TTL is discarded here. If that follower later becomes leader, `registerReplica()` rebuilds the tablet from the `Replica`'s creation-time `TableConfig`, which still contains the old TTL. For example, after increasing retention from 1h to 30d, a failover can restore 1h and permanently delete remote segments too early. Please retain the latest TTL in the replica even when no remote tablet exists and use that value when registering a new leader; a follower-ALTER-then-promote test would cover this path.

fluss-server/src/main/java/org/apache/fluss/server/replica/Replica.java
```diff
@@ -702,6 +702,33 @@ public void updateTieredLogLocalSegments(int tieredLogLocalSegments) {
...
+    /**
+     * Update the ttl in milliseconds for remote log segments. This method is called when the table
+     * configuration {@code table.log.ttl} is altered, so that the new ttl takes effect immediately
+     * for the next round of expired-segment evaluation, without requiring a server restart or table
+     * re-creation.
+     *
+     * @param newTtlMs the new ttl in milliseconds; a non-positive value disables expiration
+     */
+    public void updateLogTtlMs(long newTtlMs) {
+        Optional<Long> oldValueOpt = remoteLogManager.updateLogTtlMs(tableBucket, newTtlMs);
```
wuchong: This only updates the remote-log tablet. `LogTablet` keeps its local-retention TTL in the construction-time `final logTtlMs`, and `cleanupExpiredSegments()` continues to read that old value. With remote logging disabled this method returns without updating anything, so increasing `table.log.ttl` from 1h to 30d still lets the local cleaner delete the only copy after 1h. Please update the effective TTL used by `LogTablet` as part of the same metadata change, including when no remote tablet exists, and cover the local-retention path.

fluss-server/src/test/java/org/apache/fluss/server/log/remote/RemoteLogManagerTest.java
```diff
@@ -791,6 +791,43 @@ void testAlterTableTieredLogLocalSegments(boolean partitionedTable) throws Excep
...
+        Replica replica = replicaManager.getReplicaOrException(tb);
+        RemoteLogTablet remoteLog = remoteLogManager.remoteLogTablet(tb);
+
+        // Verify initial ttl matches the configured default table.log.ttl
+        long defaultTtlMs = ConfigOptions.TABLE_LOG_TTL.defaultValue().toMillis();
+        assertThat(remoteLog.getTtlMs()).isEqualTo(defaultTtlMs);
+
+        // 2. Update ttl via Replica.updateLogTtlMs (simulating metadata propagation)
+        long newTtlMs = java.time.Duration.ofDays(1).toMillis();
+        replica.updateLogTtlMs(newTtlMs);
```
wuchong: This test bypasses the behavior introduced by the PR by calling `Replica.updateLogTtlMs()` directly and then only checking the in-memory getter. It would still pass if ALTER metadata never reached the replica, if local cleanup kept the old TTL, or if failover restored the old value. Please exercise the real `admin.alterTable`/metadata-update path and trigger retention, asserting the manifest/segment files and the local-only cleaner behave according to the new TTL; include a follower-to-leader case.

fluss-server/src/main/java/org/apache/fluss/server/log/remote/RemoteLogManager.java
Kaixuan-Duan: Added a volatile latestLogTtlMs field in Replica to cache the latest TTL. updateLogTtlMs() writes to this field before delegating to RemoteLogManager, ensuring the TTL is retained even when a follower has no RemoteLogTablet. getLogTTLMs() now returns this cached value, so registerReplica uses the up-to-date TTL when constructing a new RemoteLogTablet during follower-to-leader promotion.

fluss-server/src/main/java/org/apache/fluss/server/replica/Replica.java
Kaixuan-Duan: Changed LogTablet.logTtlMs from final to volatile and added an updateLogTtlMs() method. In Replica.updateLogTtlMs(), logTablet.updateLogTtlMs() is called before delegating to RemoteLogManager, ensuring the local TTL is updated even when remote logging is disabled. Added a unit test verifying local segment cleanup behavior when remote logging is disabled.

fluss-server/src/test/java/org/apache/fluss/server/log/remote/RemoteLogManagerTest.java
Kaixuan-Duan: Added an integration test that exercises the real ALTER path and metadata propagation, advances the clock to trigger retention, and verifies that remote segments are not deleted and the local cleaner retains segments according to the new TTL. The follower-to-leader scenario is covered via leader election triggered by stopping the leader server.

Kaixuan-Duan:
@wuchong Thanks for the review. I've addressed comments. Could you please take a look?

wuchong:
@Kaixuan-Duan I appended 2 commits to improve the tests and implementations. @gyang94 could you help to review this as well?

gyang94:
LGTM

PR #3236, opened 2026-04-30
[flink] union read from lake with startup timestamp filter

zuston:
### Purpose

Leveraging the union read mechanism, we can achieve low disk overhead while streaming long-term data from the data lake. However, when using the DataStream API as described in the [documentation](https://fluss.apache.org/docs/engine-flink/datastream/)￼, I observed that lake splits are not generated when consuming with a timestamp-based offset.

This PR addresses that limitation by enabling timestamp filter pushdown to the lake layer. Currently, this capability is only supported for log tables. (It is somewhat unintuitive to use timestamp-based consumption for PK tables.)

### Brief change log

1. Enable timestamp filter pushdown in the lake source layer
2. Define clear split boundaries between LakeSplit and LogSplit by using the offset

### DIFF
--- a/fluss-common/src/test/java/org/apache/fluss/lake/source/TestingLakeSource.java
+++ b/fluss-common/src/test/java/org/apache/fluss/lake/source/TestingLakeSource.java
@@ -57,7 +57,7 @@ public class TestingLakeSource implements LakeSource<LakeSplit> {
 
     @Override
     public FilterPushDownResult withFilters(List<Predicate> predicates) {
-        return null;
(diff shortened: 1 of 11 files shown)


binary-signal:
For tables that have a primary key, using a starting offset timestamp is still valid: you identify the lake snapshot with a timestamp ≥ the starting offset, and then continue reading the log table from that point forward.

zuston:
> For tables that have a primary key, using a starting offset timestamp is still valid: you identify the lake snapshot with a timestamp ≥ the starting offset, and then continue reading the log table from that point forward.

make sense. let's support this later

PR #3240, opened 2026-05-02
[spark] partition filter pushdown for non-lake log and kv tables

### LINKED ISSUE #3241: [spark] partition filter pushdown for non-lake log and kv tables
### Search before asking

### Description

When a Spark query on a partitioned non-lake table has a `WHERE` clause referencing partition columns, the connector shall extracts a partition-only sub-predicate and prune the partition list at planning time. 

Partitions that can't match are skipped - no per-partition log or kv-snapshot reads are issued for them.

### Willingness to contribute

fresh-borzoni:
## Summary

closes https://github.com/apache/fluss/issues/3241

Push partition predicates to the planner so partitioned queries skip non-matching partitions instead of issuing per-partition reads. 

Mirrors Flink's `partitionFilters` path. Covers non-lake append and PK tables.

Lake enabled are left as a follow-up.

### DIFF
--- a/fluss-spark/fluss-spark-common/src/main/scala/org/apache/fluss/spark/read/FlussScan.scala
+++ b/fluss-spark/fluss-spark-common/src/main/scala/org/apache/fluss/spark/read/FlussScan.scala
@@ -41,6 +41,8 @@ trait FlussScan extends Scan {
   /** Spark predicates that the scan reports as pushed down (used in [[description]]). */
   def pushedSparkPredicates: Seq[Predicate] = Seq.empty
 
+  def partitionPredicate: Option[FlussPredicate] = None
+
   protected def scanType: String
 
   override def readSchema(): StructType = {
@@ -49,8 +51,13 @@ trait FlussScan extends Scan {
 
   override def description(): String = {
     val base = s"FlussScan: [$tablePath], Type: [$scanType]"
-    if (pushedSparkPredicates.isEmpty) base
-    else s"$base [PushedPredicates: ${pushedSparkPredicates.mkString("[", ", ", "]")}]"
+    val withPushed =
+      if (pushedSparkPredicates.isEmpty) base
+      else s"$base [PushedPredicates: ${pushedSparkPredicates.mkString("[", ", ", "]")}]"
+    partitionPredicate match {
+      case Some(p) => s"$withPushed [PartitionFilter: $p]"
+      case None => withPushed
+    }
   }
 
   override def supportedCustomMetrics(): Array[CustomMetric] =
@@ -63,7 +70,7 @@ case class FlussAppendScan(
     tableInfo: TableInfo,
     requiredSchema: Option[StructType],
     pushedPredicate: Option[FlussPredicate],
-    partitionPredicate: Option[FlussPredicate],
+    override val partitionPredicate: Option[FlussPredicate],
     override val pushedSparkPredicates: Seq[Predicate],
     options: CaseInsensitiveStringMap,
     flussConfig: Configuration)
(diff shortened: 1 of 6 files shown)


fresh-borzoni:
@Yohahaha @YannByron PTAL 🙏

fluss-spark/fluss-spark-common/src/main/scala/org/apache/fluss/spark/read/FlussScan.scala
Yohahaha: append partition filter info

Yohahaha: we can pull out `description` into FlussScan in followup pr.

Yohahaha (approved):
LGTM! only one comments, thank you!

fresh-borzoni:
@Yohahaha Ty for the review, addressed.

@luoyuxia Can you take a quick look, pls?

fresh-borzoni:
Rebased and fixed conflicts

fresh-borzoni:
@luoyuxia Can you take a quick look when you have time?

fluss-spark/fluss-spark-common/src/main/scala/org/apache/fluss/spark/utils/SparkPartitionPredicate.scala
```diff
@@ -0,0 +1,106 @@
...
+object SparkPartitionPredicate {
+
+  def extract(tableInfo: TableInfo, predicates: Seq[Predicate]): Option[FlussPredicate] = {
+    val partitionKeys = tableInfo.getPartitionKeys
+    if (partitionKeys.isEmpty) return None
+
+    val rowType = partitionRowType(tableInfo)
+    val onlyPartitionKeys = new PartitionPredicateVisitor(partitionKeys)
+
+    val converted = predicates.flatMap {
```
luoyuxia, line 40: In this PR, toPartitionRow materializes all partition values as BinaryString, and stringifyLiterals rebuilds every LeafPredicate with DataTypes.STRING. 
After that, predicate evaluation runs through `fluss-common/src/main/java/org/apache/fluss/predicate/LeafPredicate.java` and `fluss-common/src/main/java/org/apache/fluss/predicate/CompareUtils.java`, which means comparisons are performed using the predicate type. 
So for a non-string partition column, range predicates are no longer evaluated with the real type semantics, but with string lexicographic semantics.

A concrete incorrect case is an INT partition column. Suppose partitions include pt2=2 and pt2=10, and Spark pushes pt2 > 2. With correct integer semantics, partition 10 must be kept. With the current implementation, both the row value and the literal become strings, so "10" is compared to "2" lexicographically, and "10" < "2", causing partition 10 to be pruned incorrectly. 
That is a correctness bug, not just a missed optimization, because the partition is skipped before scan.

Seems Flink also have similar issue.

luoyuxia (commented):
@fresh-borzoni Thanks for the pr. Only one question

fluss-spark/fluss-spark-common/src/main/scala/org/apache/fluss/spark/utils/SparkPartitionPredicate.scala
fresh-borzoni: Good observation, I'm thinking to lift this to fluss-common and use PartitionUtils.parseValueOfType in both Spark/Flink.  

Created an issue to fix this: https://github.com/apache/fluss/issues/3292

fresh-borzoni:
@luoyuxia Ty for the review, answered, I'm thinking about fixing the issue separately for both Flink and Spark, are you okay with it?

luoyuxia (approved):

PR #3263, opened 2026-05-07
[client] Refactor RemoteLogDownloader to use chunked file append instead of downloading whole log file

### LINKED ISSUE #3262: [client] Refactor RemoteLogDownloader to use chunked streaming instead of downloading whole segment file
### Search before asking

### Motivation

The current `RemoteLogDownloader` downloads the entire remote log segment file to local disk before any data can be consumed. This has several problems:

- Wasted download bandwidth: If the consumer only needs a portion of a segment (e.g., it unsubscribes, seeks, or closes mid-segment), the remaining bytes have already been downloaded and are simply discarded. For large segments (hundreds of MBs), this wastes significant remote storage bandwidth — which is especially costly for cloud object storage (OSS, S3) where egress is metered.

- High time-to-first-byte latency: The consumer must wait for the full segment to finish downloading before it can start processing any records.

- No flow control at chunk level: The only flow control is the prefetchSemaphore which limits the number of active segments, but does not limit how far ahead data is downloaded within a single segment. This means data continues to be eagerly downloaded even when the consumer is slow, further amplifying unnecessary bandwidth usage.

### Solution

Refactor `RemoteLogDownloader` to read remote log segments in configurable chunks (default 8 MB) instead of downloading the whole file at once:

- Introduce `RemoteSegmentChunkReader` to read fixed-size chunks from the remote FSDataInputStream, aligning on batch boundaries.

- Each chunk is appended to a local temp file and returned as a `FileLogRecords` slice (leveraging OS page cache, no JVM heap copy).

- Add two-level flow control:

1. prefetchSemaphore: limits the number of concurrently active segments (existing, default 4).

1. maxPrefetchChunks: limits the number of unconsumed chunks per segment (new, default 5). The downloader pauses when this limit is reached and resumes only when chunks are consumed.

- Use a continuationQueue to prioritize chunk continuation over new segment downloads.

- Chain chunk futures via nextChunkCallback so that LogFetcher automatically registers `RemotePendingFetch` for each subsequent chunk.

- New config options: `client.scanner.remote-log.chunk-size` (default 8 MB), `client.scanner.remote-log.max-prefetch-chunks` (default 5).

### Anything else?

_No response_

### Willingness to contribute

swuferhong:
### Purpose

Linked issue: close #3262 

Refactor `RemoteLogDownloader` from whole-file download to chunked streaming I/O. This significantly reduces time-to-first-byte latency, saves bandwidth when consumers stop mid-segment, and adds fine-grained flow control at the chunk level.

### DIFF
--- a/fluss-client/src/main/java/org/apache/fluss/client/table/scanner/log/LogFetchBuffer.java
+++ b/fluss-client/src/main/java/org/apache/fluss/client/table/scanner/log/LogFetchBuffer.java
@@ -109,20 +109,31 @@ public class LogFetchBuffer implements AutoCloseable {
                     while (pendings != null && !pendings.isEmpty()) {
                         PendingFetch pendingFetch = pendings.peek();
                         if (pendingFetch.isCompleted()) {
-                            CompletedFetch completedFetch = pendingFetch.toCompletedFetch();
-                            completedFetches.add(completedFetch);
                             pendings.poll();
-                            hasCompleted = true;
+                            try {
+                                CompletedFetch completedFetch = pendingFetch.toCompletedFetch();
+                                completedFetches.add(completedFetch);
+                                hasCompleted = true;
+                            } catch (Throwable t) {
+                                // If toCompletedFetch() fails (e.g. the underlying chunk
+                                // future completed exceptionally), discard this entry so
+                                // the queue is not blocked. The bucket will become fetchable
+                                // again and the server can re-issue a remote fetch.
+                                LOG.warn(
+                                        "Discarding failed pending fetch for bucket {}",
+                                        tableBucket,
+                                        t);
+                            }
                         } else {
                             break;
                         }
                     }
                     if (hasCompleted) {
                         notEmptyCondition.signalAll();
-                        // clear the bucket entry if there is no pending fetches for the bucket.
-                        if (pendings.isEmpty()) {
-                            this.pendingFetches.remove(tableBucket);
-                        }
+                    }
+                    // clear the bucket entry if there are no pending fetches for the bucket.
+                    if (pendings != null && pendings.isEmpty()) {
+                        this.pendingFetches.remove(tableBucket);
                     }
                 });
     }
--- a/fluss-client/src/main/java/org/apache/fluss/client/table/scanner/log/RemoteLogDownloadFuture.java
+++ b/fluss-client/src/main/java/org/apache/fluss/client/table/scanner/log/RemoteLogDownloadFuture.java
@@ -17,40 +17,39 @@
 
 package org.apache.fluss.client.table.scanner.log;
 
-import org.apache.fluss.exception.FlussRuntimeException;
-import org.apache.fluss.record.FileLogRecords;
+import org.apache.fluss.record.LogRecords;
 
-import java.io.File;
-import java.io.IOException;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
 import java.util.concurrent.CompletableFuture;
+import java.util.function.Consumer;
 
-/** Represents the future of a remote log download request. */
+/**
+ * Represents the future of a single chunk read from a remote log segment. Each chunk is delivered
+ * as a {@link LogRecords} via a {@link CompletableFuture}.
+ */
 public class RemoteLogDownloadFuture {
 
-    private final CompletableFuture<File> logFileFuture;
+    private static final Logger LOG = LoggerFactory.getLogger(RemoteLogDownloadFuture.class);
+
+    private final CompletableFuture<LogRecords> chunkFuture;
     private final Runnable recycleCallback;
+    private Consumer<RemoteLogDownloadFuture> nextChunkCallback;
 
     public RemoteLogDownloadFuture(
-            CompletableFuture<File> logFileFuture, Runnable recycleCallback) {
-        this.logFileFuture = logFileFuture;
+            CompletableFuture<LogRecords> chunkFuture, Runnable recycleCallback) {
+        this.chunkFuture = chunkFuture;
         this.recycleCallback = recycleCallback;
     }
 
     public boolean isDone() {
-        return logFileFuture.isDone();
+        return chunkFuture.isDone();
     }
 
-    public FileLogRecords getFileLogRecords(int startPosition) {
-        try {
-            FileLogRecords fileLogRecords = FileLogRecords.open(logFileFuture.join(), false);
-            if (startPosition > 0) {
-                return fileLogRecords.slice(startPosition, Integer.MAX_VALUE);
-            } else {
-                return fileLogRecords;
-            }
-        } catch (IOException e) {
-            throw new FlussRuntimeException(e);
-        }
+    /** Returns the chunk data. Blocks until the chunk is ready. */
+    public LogRecords getLogRecords() {
+        return chunkFuture.join();
     }
 
     public Runnable getRecycleCallback() {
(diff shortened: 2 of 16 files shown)


fluss-client/src/test/java/org/apache/fluss/client/table/scanner/log/RemoteLogDownloaderTest.java
```diff
@@ -338,16 +420,320 @@ private static List<RemoteLogSegment> buildRemoteLogSegmentList(
         return remoteLogSegmentList;
     }
 
+    /**
+     * Tests chunked download with a small chunk size, verifying that: 1. A large segment is split
+     * into multiple chunks. 2. Chunks can be consumed while subsequent chunks are still being
+     * downloaded (边读边下载). 3. Flow control (maxPrefetchChunks) pauses downloading when unconsumed
```
fresh-borzoni, line 426: nit: mix of languages here

fluss-client/src/main/java/org/apache/fluss/client/table/scanner/log/RemoteLogDownloader.java
```diff
@@ -40,187 +41,342 @@
...
     }
 
-    /** Request to fetch remote log segment to local. This method is non-blocking. */
-    public RemoteLogDownloadFuture requestRemoteLog(FsPath logTabletDir, RemoteLogSegment segment) {
-        RemoteLogDownloadRequest request = new RemoteLogDownloadRequest(segment, logTabletDir);
+    /**
+     * Request to read a remote log segment in chunks starting from the given position. This method
+     * is non-blocking and returns a future for the first chunk.
+     */
+    public RemoteLogDownloadFuture requestRemoteLog(
```
fresh-borzoni, line 135: I think we introduced a race:
requestRemoteLog() adds the request to segmentsToFetch and returns. The caller then installs the next-chunk callback on the returned future. But the download thread is already running, so it can poll the request, read chunk 1, and call tryScheduleNextChunk(), which builds chunk 2's future by copying request.downloadFuture.getNextChunkCallback(), still null at that moment. Chunk 2 is read and completed, but nothing is listening,so the bucket silently stops at chunk 1.
Probably we can install the callback before publishing the request

fluss-client/src/main/java/org/apache/fluss/client/table/scanner/log/RemoteLogDownloadFuture.java
```diff
@@ -17,47 +17,69 @@
...
+ * as a {@link LogRecords} via a {@link CompletableFuture}.
+ */
 public class RemoteLogDownloadFuture {
 
-    private final CompletableFuture<File> logFileFuture;
+    private static final Logger LOG = LoggerFactory.getLogger(RemoteLogDownloadFuture.class);
+
+    private final CompletableFuture<LogRecords> chunkFuture;
     private final Runnable recycleCallback;
+    private Consumer<RemoteLogDownloadFuture> nextChunkCallback;
```
fresh-borzoni, line 38: volatile?

fluss-client/src/main/java/org/apache/fluss/client/table/scanner/log/RemoteLogDownloader.java
fresh-borzoni, line 384: A request that's been polled and partially read but is paused is in neither segmentsToFetch nor continuationQueue. 

close() only walks those two queues, so the open FSDataInputStream and the remote-chunk tmp file are left behind.

fluss-common/src/main/java/org/apache/fluss/config/ConfigOptions.java
```diff
@@ -1271,6 +1271,24 @@ public class ConfigOptions {
...
+    public static final ConfigOption<MemorySize> CLIENT_SCANNER_REMOTE_LOG_CHUNK_SIZE =
+            key("client.scanner.remote-log.chunk-size")
+                    .memoryType()
+                    .defaultValue(MemorySize.parse("8mb"))
+                    .withDescription(
+                            "The size of each chunk when downloading remote log segments. "
+                                    + "A larger chunk size reduces the number of remote I/O requests but "
+                                    + "increases memory usage per chunk read. The default setting is 8MB.");
+
+    public static final ConfigOption<Integer> CLIENT_SCANNER_REMOTE_LOG_MAX_PREFETCH_CHUNKS =
```
fresh-borzoni, line 1283: I wonder what will happen if it's 0?

fluss-client/src/main/java/org/apache/fluss/client/table/scanner/log/RemoteLogDownloader.java
```diff
@@ -40,187 +41,342 @@
...
+            }
 
-            // 2. do the actual download work
-            FsPathAndFileName fsPathAndFileName = request.getFsPathAndFileName();
             scannerMetricGroup.remoteFetchRequestCount().inc();
 
-            long startTime = System.currentTimeMillis();
-            // download the remote file to local
-            remoteFileDownloader
-                    .downloadFileAsync(fsPathAndFileName, localLogDir)
```
fresh-borzoni, line 167: The old fetchOnce dispatched the download to RemoteFileDownloader's thread pool (default 3) and returned immediately, so multiple segments could be downloading at once. The new processChunkRead does the fs.open + read inline on the dispatcher, so only one chunk is ever in flight.

this looks like it serializes remote reads, is it intentional or I'm missing smth?

fluss-client/src/main/java/org/apache/fluss/client/table/scanner/log/LogFetchBuffer.java
```diff
@@ -109,20 +109,31 @@ void tryComplete(TableBucket tableBucket) {
...
-                            hasCompleted = true;
+                            try {
+                                CompletedFetch completedFetch = pendingFetch.toCompletedFetch();
+                                completedFetches.add(completedFetch);
+                                hasCompleted = true;
+                            } catch (Throwable t) {
+                                // If toCompletedFetch() fails (e.g. the underlying chunk
+                                // future completed exceptionally), discard this entry so
+                                // the queue is not blocked. The bucket will become fetchable
+                                // again and the server can re-issue a remote fetch.
```
fresh-borzoni, line 121: nit: I'm not sure: what if the issue is permanent? It would be a infitite loop with only warning. Should we have retry with backoff and then surface an error in the scanner?

fluss-client/src/test/java/org/apache/fluss/client/table/scanner/log/RemoteLogDownloaderTest.java
```diff
@@ -282,6 +253,117 @@ void testOrderOfRemoteLogDownloadRequest() {
...
+        // Set prefetchNum = 1 so only one segment is active.
+        conf.set(ConfigOptions.CLIENT_SCANNER_REMOTE_LOG_PREFETCH_NUM, 1);
+        RemoteLogDownloader remoteLogDownloader =
+                new RemoteLogDownloader(DATA1_TABLE_PATH, conf, scannerMetricGroup, 10L);
+        try {
+            remoteLogDownloader.start();
+
+            TableBucket tb = new TableBucket(DATA1_TABLE_ID, 0);
+            // Build a large segment with multiple records so multiple chunks are produced.
+            List<RemoteLogSegment> remoteLogSegments =
```
fresh-borzoni, line 269: It doesn't look big or at least >8mb chunk size: https://github.com/apache/fluss/blob/170e95f9039c54eb7255a84f761b609a77c7e594/fluss-common/src/test/java/org/apache/fluss/record/TestData.java#L44

fresh-borzoni (commented):
@swuferhong Ty for the PR, that's an big improvement!

Now I want to backport it to rust part as well, I left some comments. 
I hope you find them useful, PTAL

PR #3271, opened 2026-05-08
[flink] Fix COUNT(column) aggregate pushdown to reject nullable columns

### LINKED ISSUE #3270: [flink] COUNT(column) aggregate pushdown produces incorrect results for nullable columns
### Search before asking

### Fluss version

0.9.0 (latest release)

#### Problem

The current `applyAggregates` implementation in `FlinkTableSource` incorrectly pushes down `COUNT(column)` queries for nullable columns. When the aggregate function is `CountAggFunction` (which handles both `COUNT(*)` and `COUNT(column)`), it is always treated as equivalent to `COUNT(*)` and pushed down to return the total row count.

However, `COUNT(column)` should exclude NULL values from the count, whereas `COUNT(*)` counts all rows regardless of NULL values. This leads to incorrect query results when the column contains NULL values.

####Expected Behavior

- `COUNT(*)` and `COUNT(1)` should be pushed down (total row count).
- `COUNT(column)` on a **non-nullable** column should be pushed down (equivalent to `COUNT(*)`).
- `COUNT(column)` on a **nullable** column should NOT be pushed down, because the server-side row count includes rows with NULL values in that column.

#### Current Behavior

All forms of COUNT are pushed down regardless of column nullability, producing incorrect results for `COUNT(nullable_column)`.

#### Reproduction

```sql
-- Table with nullable 'address' column containing NULLs
SELECT COUNT(address) FROM my_table;
-- Returns total row count (e.g., 5) instead of non-null count (e.g., 3)
```

### Solution

_No response_

beryllw:
### Purpose

Linked issue: close #3270

Fixes the aggregate pushdown logic to correctly handle COUNT(column) on nullable columns.

### Brief change log

When the aggregate is CountAggFunction with a nullable column argument, the pushdown is rejected so that Flink handles the NULL-excluding count correctly.

### Tests

- FlinkTableSourceBatchITCase#testCountPushDownForPkTable
- FlinkTableSourceBatchITCase#testCountPushDownForLogTable

### DIFF
--- a/fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/source/FlinkTableSource.java
+++ b/fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/source/FlinkTableSource.java
@@ -78,7 +78,6 @@ import org.apache.flink.table.data.RowData;
 import org.apache.flink.table.expressions.AggregateExpression;
 import org.apache.flink.table.expressions.ResolvedExpression;
 import org.apache.flink.table.functions.AsyncLookupFunction;
-import org.apache.flink.table.functions.FunctionDefinition;
 import org.apache.flink.table.functions.LookupFunction;
 import org.apache.flink.table.types.DataType;
 import org.apache.flink.table.types.logical.LogicalType;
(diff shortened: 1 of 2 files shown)


beryllw:
@luoyuxia cc

luoyuxia:
cc @loserwang1024

loserwang1024:
image
@beryllw test fails

beryllw:
> image @beryllw test fails

fixed

loserwang1024:
> When the aggregate is CountAggFunction with a nullable column argument, the pushdown is rejected so that Flink handles the NULL-excluding count correctly.

Is this a bug in flink sql? Maybe we can take a flink jira for it?

beryllw:
> Is this a bug in flink sql? Maybe we can take a flink jira for it?

It's a Flink Fluss SQL Source bug.

loserwang1024 (approved):
LGTM

PR #3272, opened 2026-05-08
[flink] add split assignment mechanism w/o partitionId

### LINKED ISSUE #3269: Optional channel selection for source split in FlinkSourceEnumerator
### Search before asking

### Motivation

We'd better to introduce the similar mechanism (shuffle assign without the partition) like [paimon](https://github.com/apache/paimon/blob/959170d8fe1e871f54d17e4f6a6980eee389cf6e/paimon-flink/paimon-flink-common/src/main/java/org/apache/paimon/flink/source/ContinuousFileSplitEnumerator.java#L336) did

### Solution

_No response_

### Anything else?

_No response_

### Willingness to contribute

zuston:
### Purpose

Linked issue: close #3269 

The motivation of this PR is to align with the sink channel selector. And another point is to make it possible that all bucket data could be routed into the same subtasks if having the partitionIds when `numBucket % concurrency = 0` , that could emlinate the shuffle to improve the performance to reduce backfill time for large-scale data.

### DIFF
--- a/fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/sink/ChannelComputer.java
+++ b/fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/sink/ChannelComputer.java
@@ -59,6 +59,11 @@ public interface ChannelComputer<T> extends Serializable {
         return (startChannel + bucket) % numChannels;
     }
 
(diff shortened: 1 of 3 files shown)


luoyuxia:
cc @loserwang1024

PR #3281, opened 2026-05-08
[catalog] Preserve exception cause when rethrowing catalog exceptions

XuQianJin-Stars:
Several places in the catalog code rethrow caught exceptions by passing only getMessage() to the new exception, which drops the original stack trace and cause. This makes diagnosing failures harder because the root cause is lost.

Preserve the cause by passing the original Throwable as the second argument to the exception constructor in:
  - FlinkCatalog (createTable: lake-table-already-exist and invalid-table branches; alterTable: invalid-table branch)
  - PaimonLakeCatalog (AddColumn ColumnAlready/NotExistException branch)
  - CoordinatorService (UncheckedIOException and generic validate branches when wrapping into InvalidTableException)

No behavior change other than richer exception chaining.

### Purpose

Linked issue: none (minor code cleanup, no GitHub issue filed)

Several places in the catalog code rethrow caught exceptions by passing only
`getMessage()` to the new exception, which drops the original stack trace and
cause. This makes diagnosing failures harder because the root cause is lost
from the logs.

This PR preserves the cause by passing the original `Throwable` as the second
argument to the exception constructor, so that exception chaining is kept
intact and the root cause is visible in logs.

### Brief change log

- `FlinkCatalog#createTable`: preserve cause in the `lake-table-already-exist`
  branch (`new CatalogException(t.getMessage(), t)`) and the `invalid-table`
  branch (`new InvalidTableException(t.getMessage(), t)`).
- `FlinkCatalog#alterTable`: preserve cause in the `invalid-table` branch
  (`new InvalidTableException(t.getMessage(), t)`).
- `PaimonLakeCatalog`: preserve cause for the
  `ColumnAlreadyExistException | ColumnNotExistException` branch in
  `AddColumn` handling
  (`new InvalidAlterTableException(e.getMessage(), e)`).
- `CoordinatorService`: preserve cause when wrapping `UncheckedIOException`
  and generic validation errors into `InvalidTableException`.

### Tests

No new tests are added. This is a trivial rework that only changes the
exception-construction call to additionally pass the original `Throwable`
as the cause. Behavior of existing tests is unaffected.

### API and Format

No API or storage format changes. The exception types thrown remain the
same; only the `cause` field of the thrown exceptions is now populated.

### Documentation

No documentation changes. This is an internal cleanup without any
user-facing behavior change.

### DIFF
--- a/fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/catalog/FlinkCatalog.java
+++ b/fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/catalog/FlinkCatalog.java
@@ -519,9 +519,9 @@ public class FlinkCatalog extends AbstractCatalog {
             } else if (CatalogExceptionUtils.isTableAlreadyExist(t)) {
                 throw new TableAlreadyExistException(getName(), objectPath);
             } else if (CatalogExceptionUtils.isLakeTableAlreadyExist(t)) {
-                throw new CatalogException(t.getMessage());
+                throw new CatalogException(t.getMessage(), t);
             } else if (isTableInvalid(t)) {
-                throw new InvalidTableException(t.getMessage());
+                throw new InvalidTableException(t.getMessage(), t);
             } else {
                 throw new CatalogException(
                         String.format("Failed to create table %s in %s", objectPath, getName()), t);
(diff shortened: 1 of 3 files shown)


luoyuxia (approved):
+1

PR #3291, opened 2026-05-09
[flink] Avoid creating empty lake writer in TieringSplitReader

### LINKED ISSUE #3290: [flink] Avoid creating empty lake writer in TieringSplitReader
### Search before asking

### Fluss version
main (development)

### Please describe the bug 🐞
In `TieringSplitReader.forLogRecords`, the lake writer is created before checking whether the current polled records actually contain any record with `record.logOffset() < stoppingOffset`.

If the polled batch only contains records whose offsets have already reached or passed the split `stoppingOffset`, the split can still be marked finished based on the last record offset, but nothing is written into the lake writer.

This shows up with logical empty batches that still advance offsets, for example with first-row merge engine updates or deleting a non-existent key. In that case `recordWriter.complete()` may fail during tiering commit with:

```
The size of CommitMessage must be 1, but got [].
```

A concrete sequence is:
1. `TieringSplitReader` subscribes a log split with a finite `stoppingOffset`.
2. `forLogRecords` receives `bucketScanRecords` for that bucket, but every record has `logOffset() >= stoppingOffset`.
3. A lake writer has already been created, although no record is written.
4. The reader sees `lastRecord.logOffset() >= stoppingOffset - 1`, finishes the split, and calls `completeLakeWriter()`.
5. The underlying lake writer completes an empty write and fails, for example Paimon throws `The size of CommitMessage must be 1, but got [].`

### Solution
Create the lake writer lazily only when the first record satisfying `record.logOffset() < stoppingOffset` is encountered. If no record is actually written for the batch/split, keep the current `completeLakeWriter()` behavior and return a null write result.

luoyuxia:
### Purpose
Linked issue: close #3290

Fix a tiering log reader bug where `TieringSplitReader` creates a lake writer before it knows whether the current polled batch contains any record below the split `stoppingOffset`.

### Brief change log
- Delay lake writer creation in `TieringSplitReader.forLogRecords` until the first record with `record.logOffset() < stoppingOffset` is encountered.
- Keep the existing split finishing path unchanged so offset advancement without actual writes no longer produces an empty lake writer commit.
- Generative AI disclosure: Codex (GPT-5).

### Tests
- Ran `./mvnw -pl fluss-flink/fluss-flink-common -DskipTests spotless:check checkstyle:check`.
- Did not run unit or integration tests, per request.

### API and Format

### Documentation
No documentation changes.

### DIFF
--- a/fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/tiering/source/TieringSplitReader.java
+++ b/fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/tiering/source/TieringSplitReader.java
@@ -369,12 +369,16 @@ public class TieringSplitReader<WriteResult>
                 continue;
             }
             LOG.info("tiering table bucket stoppingOffset is not empty {}.", bucket);
-            LakeWriter<WriteResult> lakeWriter =
-                    getOrCreateLakeWriter(
-                            bucket, currentTableSplitsByBucket.get(bucket).getPartitionName());
+            LakeWriter<WriteResult> lakeWriter = null;
             for (ScanRecord record : bucketScanRecords) {
                 // if record is less than stopping offset
                 if (record.logOffset() < stoppingOffset) {
+                    if (lakeWriter == null) {
+                        lakeWriter =
+                                getOrCreateLakeWriter(
+                                        bucket,
+                                        currentTableSplitsByBucket.get(bucket).getPartitionName());
+                    }
                     lakeWriter.write(record);
                     if (record.getSizeInBytes() > 0) {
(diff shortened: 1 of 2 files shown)


fresh-borzoni (approved):
+1
nit: just a small regression test will be great, and also I checked quickly SnapshotSplitRecords, it's safe, but relies on BoundedSplitReader logic:
- If a snapshot has zero records, BoundedSplitReader.readBatch() returns null
- if Empty iterator -> writer created, but nothing written. Next call writes records to the same writer, so it's safe

Mb we wish to add the same symmetrical fix there or just leave comment, so that it's less brittle

beryllw:
LGTM!

fresh-borzoni (approved):
Nice, ty 👍

PR #3295, opened 2026-05-09
[Client] Kvscan java client

### LINKED ISSUE #3126: Extend the API to support KvScan

polyzos:
closes #3126

### DIFF
--- a/fluss-client/src/main/java/org/apache/fluss/client/table/scanner/batch/KvBatchScanner.java
+++ b/fluss-client/src/main/java/org/apache/fluss/client/table/scanner/batch/KvBatchScanner.java
@@ -45,7 +45,6 @@ import javax.annotation.Nullable;
 import javax.annotation.concurrent.NotThreadSafe;
 
 import java.io.IOException;
-import java.nio.ByteBuffer;
 import java.time.Duration;
 import java.util.ArrayList;
 import java.util.Collections;
@@ -134,6 +133,10 @@ public final class KvBatchScanner implements BatchScanner {
         } catch (TimeoutException e) {
             return CloseableIterator.emptyIterator();
         } catch (Exception e) {
+            // TODO: we should retry continuation requests on transient networks errors, but
+            //  for simplicity we currently fail the whole scan. Retrying here would require robust
+            //  handling of duplicate continuations (e.g. call_seq_id rollback) to avoid skipping or
+            //  repeating rows.
             terminate();
             throw new IOException(e);
         }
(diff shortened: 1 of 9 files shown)


polyzos:
@wuchong @fresh-borzoni PTAL

wuchong (approved):
I pushed a commit to fix the `call_seq_id` issue. 

> KvBatchScanner.sendContinuation() sends the wrong call_seq_id for the first continuation, which the server will reject with InvalidScanRequestException. Client initializes callSeqId = 0 and pre-increments, so the first continuation carries 1; but the server's ScannerContext initializes callSeqId = -1 (so the first valid client value is 0) and TabletService rejects when requestSeqId !=  context.getCallSeqId() + 1. The new IT tests don't catch this because the scanned data fits in one batch  and no continuation is ever sent; KvBatchScannerTest only asserts the client's locally-emitted sequence against a fake gateway. To fix, either initialize callSeqId = -1 and keep pre-increment, or use post-increment so wire values start at 0.

This can be reproduced by `TableKvScanITCase#snapshotIsolationHidesPostOpenWrites` I updated in the appended commit. 

@polyzos Please take a looks at the changes. If you don't have any problems, I can merge this once the CI is passed.

polyzos:
@wuchong LGTM, thanks for handling this

fluss-client/src/main/java/org/apache/fluss/client/table/scanner/batch/KvBatchScanner.java
```diff
@@ -0,0 +1,349 @@
...
+        if (hasMore) {
+            sendContinuation();
+        } else {
+            drained = true;
+        }
+
+        if (!response.hasRecords()) {
+            return drained ? null : CloseableIterator.emptyIterator();
+        }
+        return CloseableIterator.wrap(parseRecords(response).iterator());
```
fresh-borzoni, line 163: Shall we release underlying ByteBuf here, like we do in other places like LimitBatchScanner?

fluss-client/src/main/java/org/apache/fluss/client/table/scanner/batch/KvBatchScanner.java
```diff
@@ -0,0 +1,349 @@
...
+
+    private CloseableIterator<InternalRow> handleErrorResponse(ScanKvResponse response)
+            throws IOException {
+        Errors error = Errors.forCode(response.getErrorCode());
+        String message = response.hasErrorMessage() ? response.getErrorMessage() : null;
+
+        if (error == Errors.TOO_MANY_SCANNERS
+                && scannerId == null
+                && openRetries < MAX_OPEN_RETRIES) {
+            long delayMs = BASE_RETRY_DELAY_MS * (1L << openRetries);
```
fresh-borzoni, line 214: Wouldn't it conflict with timeout setting for polling? So in theory we can wait more that timeout, do we have any implications with this?

fresh-borzoni (commented):
@polyzos left some comments for followups, not blocking

fresh-borzoni:
I still don't like this RocksDB snapshot pinning, tbh. It feels like it's going to be a problem for wide partitioned tables with many buckets and concurrent users.

Basically it's operational burden, two ways:
  1. Hard to figure out why scans fail. A single scan over >200 buckets per tablet server(easy with partitions) hits TOO_MANY_SCANNERS, the client retries 3× under 700 ms, can't recover (real scans don't drain in sub-second), and dies. The error doesn't say "per-server cap" -  operators see a generic exception and have to dig through server logs. Two concurrent scans interfering with each other is worse: nobody expects read traffic to throttle other read traffic.
  2. Writes degrade because memtables can't be flushed. Each held iterator forces the leader to keep memtables in memory until the scan ends. Memory pressure builds, compaction falls behind, write-path tail latency goes up. Operators see "P99 writes got slow during a batch read" with no causal trail in the logs from symptom back to the scan.

And bumping kv.scanner.max-per-server past 200 doesn't really help - it just trades TOO_MANY_SCANNERS for memory pressure (several GB of pinning per server for many partitions/many buckets tables).

PR #3302, opened 2026-05-12
[Bug] Avoid altering lake catalog for tables that never enabled datalake

### LINKED ISSUE #3297: [Bug] Avoid altering lake catalog for tables that never enabled datalake
### Search before asking

### Fluss version

main (development)

### Please describe the bug 🐞

`MetadataManager#preAlterTableProperties` currently calls `lakeCatalog.alterTable(...)` whenever a lake catalog is configured:

```java
// We should always alter lake table even though datalake is disabled.
// Otherwise, if user alter the fluss table when datalake is disabled, then enable datalake
// again, the lake table will mismatch.
if (lakeCatalog != null) {
    try {
        lakeCatalog.alterTable(tablePath, tableChanges, lakeCatalogContext);
    } catch (TableNotExistException e) {
        // only throw TableNotExistException if datalake is enabled
        if (isDataLakeEnabled(newDescriptor)) {
            throw new FlussRuntimeException(
                    "Lake table doesn't exist for lake-enabled table "
                            + tablePath
                            + ", which shouldn't be happened. Please check if the lake table was deleted manually.",
                    e);
        }
    }
}
```

The intention is reasonable for tables that have been lake-enabled before: if the user disables datalake, alters the Fluss table, and enables datalake again later, the lake table metadata should still stay in sync.

However, the current condition is too broad. For a normal Fluss table that has never been a lake table, any alter-table-properties operation still calls `lakeCatalog.alterTable(...)` when `lakeCatalog != null`. This creates unnecessary overhead and may also cause avoidable interaction with the lake catalog, even though there is no corresponding lake table to update.

One triggering sequence is:

1. Start a Fluss cluster with a lake catalog configured.
2. Create a normal Fluss table without setting `table.datalake.enabled`.
3. Alter unrelated table properties on that normal table.
4. `MetadataManager` still calls `lakeCatalog.alterTable(...)` for the table and suppresses `TableNotExistException` because the new descriptor is not datalake-enabled.

Expected behavior:

For tables that have never opted into datalake, altering Fluss table properties should not call `lakeCatalog.alterTable(...)`.

Actual behavior:

As long as a lake catalog exists, `MetadataManager` attempts to alter the lake table for non-lake tables as well.

### Solution

A simple mitigation is to only alter the lake table when the table descriptor contains the `table.datalake.enabled` key, regardless of whether the value is `true` or `false`.

This keeps the existing behavior for the common case where a table was previously lake-enabled and later disabled, because such tables normally retain `table.datalake.enabled=false`. It avoids calling the lake catalog for tables that never set `table.datalake.enabled`.

This is not a perfect historical signal, because a user could create a table with `table.datalake.enabled=false` from the beginning, but it should address most cases without requiring additional metadata.

MaheshS08:
### Purpose
Linked issue: close #3297

The `preAlterTableProperties` method was calling `lakeCatalog.alterTable()` 
for every table alter operation whenever a lake catalog was configured at the 
cluster level, even for plain Fluss tables that never had `table.datalake.enabled` 
set. This resulted in unnecessary calls to the lake catalog, which would throw 
`TableNotExistException` (silently swallowed) for tables unknown to the lake catalog.

### Brief change log
- In `MetadataManager#preAlterTableProperties`, tightened the guard condition 
  from `if (lakeCatalog != null)` to also check that the current table descriptor 
  contains the `table.datalake.enabled` key (regardless of its value).
- This ensures `lakeCatalog.alterTable()` is only called for tables that have 
  ever opted into datalake, while preserving the existing sync behavior for 
  tables that were previously lake-enabled and later disabled.

### Tests
- No Tests required

### API and Format
No API or storage format changes. This is a behavioral fix in the coordinator's 
metadata management logic only.

### Documentation
No new feature introduced. No documentation changes required.

### DIFF
--- a/fluss-server/src/main/java/org/apache/fluss/server/coordinator/MetadataManager.java
+++ b/fluss-server/src/main/java/org/apache/fluss/server/coordinator/MetadataManager.java
@@ -603,8 +603,12 @@ public class MetadataManager {
         // We should always alter lake table even though datalake is disabled.
         // Otherwise, if user alter the fluss table when datalake is disabled, then enable datalake
         // again, the lake table will mismatch.
-        // Only sync to lake if this table has ever opted into datalake (key present regardless of value).
-        if (lakeCatalog != null && tableDescriptor.getProperties().containsKey(ConfigOptions.TABLE_DATALAKE_ENABLED.key())) {
+        // Only sync to lake if this table has ever opted into datalake (key present regardless of
+        // value).
+        if (lakeCatalog != null
+                && tableDescriptor
+                        .getProperties()
+                        .containsKey(ConfigOptions.TABLE_DATALAKE_ENABLED.key())) {
             try {
                 lakeCatalog.alterTable(tablePath, tableChanges, lakeCatalogContext);
             } catch (TableNotExistException e) {
--- a/fluss-server/src/main/java/org/apache/fluss/server/coordinator/MetadataManager.java
+++ b/fluss-server/src/main/java/org/apache/fluss/server/coordinator/MetadataManager.java
@@ -603,7 +603,7 @@ public class MetadataManager {
         // We should always alter lake table even though datalake is disabled.
         // Otherwise, if user alter the fluss table when datalake is disabled, then enable datalake
         // again, the lake table will mismatch.
-        if (lakeCatalog != null) {
+        if (lakeCatalog != null && tableDescriptor.getProperties().containsKey(ConfigOptions.TABLE_DATALAKE_ENABLED.key())) {
             try {
                 lakeCatalog.alterTable(tablePath, tableChanges, lakeCatalogContext);
             } catch (TableNotExistException e) {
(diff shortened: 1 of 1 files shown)


MaheshS08:
@luoyuxia , could you please review this PR?

luoyuxia:
> @luoyuxia , could you please review this PR?

But ci fails

pbanakar:
Hi @MaheshS08 Looks like the CI failure is from Spotless formatting. You may have missed running `mvn spotless:apply` before pushing the changes.

Also, please go through the [contributing guidelines](https://fluss.apache.org/community/welcome/) once most of these checks are covered there and it helps avoid CI failures

MaheshS08:
@Prajwal-banakar thanks for the head's up. I have re-pushed to the branch with necessary changes.

@luoyuxia , please have a look at this PR, now the checks have been passed.

luoyuxia (approved):
+1

PR #3308, opened 2026-05-13
[client] read fallback to lake when encountering OutOfRange exception

### LINKED ISSUE #3228: Possible OutOfRangeException when reading the stream log table after resuming from a long interruption

zuston:
### Purpose

Linked issue: close #3228 

### Brief change log

1. fix the lake missing startOffset setting
2. server trigger lake read fallback, and then client can read from the lake after specified startOffset


fluss-client/src/main/java/org/apache/fluss/client/table/scanner/log/LogFetcher.java
```diff
@@ -496,6 +514,106 @@ private void pendRemoteFetches(
...
+        }
+    }
+
+    private List<LakeSplit> planLakeSplits(
+            TableBucket tableBucket, LakeLogFetchInfo lakeLogFetchInfo) throws IOException {
+        List<LakeSplit> plannedSplits =
+                checkNotNull(lakeSource).createPlanner(() -> lakeLogFetchInfo.snapshotId()).plan();
+        List<LakeSplit> lakeSplits = new ArrayList<>();
+        for (LakeSplit lakeSplit : plannedSplits) {
+            // todo: partition + bucket filter and offset column filter could be supported in the
```
zuston, line 545: If this PR is OK, I will submit the following PR to optimize the planning

zuston:
@luoyuxia Could you help review this?

Issue #3274, opened 2026-05-08
Introduce HudiBucketingFunction for bucket strategy


fhan688:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/fluss/issues) and found nothing similar.

### Description

Introduce HudiBucketingFunction for bucket strategy

### Willingness to contribute

- [x] I'm willing to submit a PR!

luoyuxia:
close via #3316

PR #3316, opened 2026-05-14
[lake/hudi] Introduce HudiBucketingFunction for bucket strategy

fhan688:
### Purpose

Linked issue: https://github.com/apache/fluss/issues/3274

Introduce Hudi's bucketing strategy into Fluss so that the Fluss server/client can compute the same bucket id as Hudi's BucketIdentifier when tiering data into a Hudi table with bucket index. This is a prerequisite for the upcoming HudiLakeWriter and HudiCompaction PRs, which need to route records to the correct Hudi bucket file.

### Brief change log

fluss-common (production code)

- BucketingFunction.of(...) — add DataLakeFormat.HUDI branch that returns HudiBucketingFunction.

- HudiBucketingFunction — implements BucketingFunction. Decodes a 4-byte big-endian int produced by HudiKeyEncoder and computes (hash & Integer.MAX_VALUE) % numBuckets, matching Hudi's BucketIdentifier.getBucketId(List<String>, int). Includes strict input validation (bucketKey must be exactly 4 bytes, numBuckets must be positive).

- KeyEncoder.createKeyEncoder(...) — add DataLakeFormat.HUDI branch that returns HudiKeyEncoder.

- HudiKeyEncoder — implements KeyEncoder. Computes List<String>.hashCode() inline (h = 31*h + elementStringHash) over the stringified key fields, avoiding intermediate ArrayList/String.valueOf allocations on the hot path. For common numeric types (int, long, byte, short, boolean) the string hash code is computed without materializing the string. Null fields are encoded as "__null__" placeholder (aligned with Hudi's KeyGenUtils.NULL_RECORDKEY_PLACEHOLDER) to avoid collision with the literal string "null".

fluss-lake-hudi (test & build)

- HudiBucketingFunctionTest — 13 test cases covering:

1. Single-field types: INT, BIGINT, STRING, DECIMAL, TIMESTAMP_NTZ

2. Additional types: BOOLEAN, TINYINT, SMALLINT, FLOAT, DATE, TIME, TIMESTAMP_LTZ

3. Composite (multi-field) bucket keys with and without null fields

4. Null field uses placeholder (not literal "null") — regression test

5. Illegal input: bucketKey null / wrong length / numBuckets ≤ 0

6. Boundary: numBuckets=1, Integer.MIN_VALUE hash, negative hash sign-bit handling

7. All tests cross-validate against Hudi's BucketIdentifier.getBucketId(List<String>, int)

- pom.xml — add hudi-flink${flink.major.version}-bundle with <scope>test</scope> so it is only available during unit tests and does not leak into the runtime classpath.

### Tests

HudiBucketingFunctionTest (13 test cases, all passing):

- testIntegerHash / testLongHash / testStringHash / testDecimalHash / testTimestampEncodingHash — original single-field coverage

- testNullFieldUsesPlaceholder / testNullFieldDoesNotCollideWithLiteralNullString — null handling

- testBucketingRejectsInvalidBucketKey / testBucketingRejectsNonPositiveNumBuckets — input validation

- testCompositeBucketKeyMatchesHudiFieldValueRecordKey / testCompositeBucketKeyWithNullFieldUsesPlaceholder — multi-field keys

- testBooleanAndIntegralTypes / testDateAndTimeTypes / testTimestampLtzType — type coverage

- testBucketingNumBucketsBoundaryValues — boundary conditions

### API and Format

No API or storage format changes. This PR only adds new implementations behind existing interfaces (BucketingFunction and KeyEncoder) for a new DataLakeFormat.HUDI enum value that was already defined.

### Documentation

No new user-facing documentation required. This is an internal bucketing strategy implementation.


fhan688:
please help review, thanks!
@XuQianJin-Stars

fluss-common/src/main/java/org/apache/fluss/row/encode/KeyEncoder.java
```diff
@@ -129,6 +130,8 @@ static KeyEncoder of(
             return CompactedKeyEncoder.createKeyEncoder(rowType, keyFields);
         } else if (lakeFormat == DataLakeFormat.ICEBERG) {
             return new IcebergKeyEncoder(rowType, keyFields);
+        } else if (lakeFormat == DataLakeFormat.HUDI) {
+            return new HudiKeyEncoder(rowType, keyFields);
```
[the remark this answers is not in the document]
fhan688: nice review, thanks.

fluss-common/src/main/java/org/apache/fluss/row/encode/hudi/HudiKeyEncoder.java
```diff
@@ -0,0 +1,88 @@
...
+     * is intentionally aligned with Hudi's {@code KeyGenUtils.NULL_RECORDKEY_PLACEHOLDER} so that
+     * the resulting bucket id stays identical to what Hudi would compute on its side.
+     */
+    public static final String NULL_RECORDKEY_PLACEHOLDER = "__null__";
+
+    private final InternalRow.FieldGetter[] fieldGetters;
+
+    public HudiKeyEncoder(RowType rowType, List<String> keys) {
+        // for getting key fields out of fluss internal row
+        fieldGetters = new InternalRow.FieldGetter[keys.size()];
```
XuQianJin-Stars, line 118: The encoded hash here is `values.hashCode()` over a `List<String>` built directly from each key field's `toString()`. However, Hudi's production path goes through `BucketIdentifier#getBucketId(HoodieKey, indexKeyFields, numBuckets)`, which **parses the record-key string by splitting on `:` and `,`**. As soon as a key field's string form contains `:` or `,` (very common for `TIMESTAMP_LTZ`, e.g. `2023-10-25T10:01:13.182Z`, or any user string with a comma), the `List<String>` Hudi reconstructs differs from the one we hash here, and the resulting bucket id will diverge from Hudi's.
Note that `HudiBucketingFunctionTest#testTimestampLtzType` only validates against the `BucketIdentifier.getBucketId(List<String>, int)` overload, which sidesteps this parsing step. Please add an end-to-end test that goes through the `HoodieKey` overload, and either escape `:` / `,` inside `stringifyForRecordKey`, or document the limitation explicitly in the class Javadoc.

fluss-common/src/main/java/org/apache/fluss/row/encode/hudi/HudiKeyEncoder.java
```diff
@@ -0,0 +1,88 @@
...
+    }
+
+    private static String stringifyForRecordKey(Object value) {
+        if (value == null) {
+            return NULL_RECORDKEY_PLACEHOLDER;
+        }
+        if (value instanceof BinaryString) {
+            return value.toString();
+        }
+        return String.valueOf(value);
```
XuQianJin-Stars: `String.valueOf(value)` at line 86 will produce `"[B@xxxx"` for `byte[]`, and reference-style strings for `BinaryArrayData / BinaryMapData / BinaryRowData`. This means a Hudi table with such a column declared as a bucket key would compute non-reproducible bucket ids.
Please reject unsupported types in the constructor by checking `keyDataType.getTypeRoot()` and only allowing primitive/decimal/string/temporal types. Throw `IllegalArgumentException` for the rest, with a clear message.

fluss-common/src/main/java/org/apache/fluss/row/encode/hudi/HudiKeyEncoder.java
fhan688: Thanks — re-read Hudi's `KeyGenUtils.extractRecordKeysByFields` in [Hudi master/release-1.x](https://github.com/apache/hudi). The parser has two modes that we now mirror in `HudiKeyEncoder`:

Single-part recordKey (no `,` and no `:`) → returned verbatim, placeholder NOT round-tripped.
Composite recordKey → split on `:`/`,`, with `"__null__"` → `null` and `"__empty__"` → `""` before `List#hashCode()`.
Two findings while implementing the alignment:

The :-in-value concern in your comment is actually handled correctly by Hudi's look-ahead loop (the `commaPosition < keyValueSep1` retry), so `TIMESTAMP_LTZ` values like `2023-10-25T10:01:13.182Z` are fine — confirmed by a new end-to-end test that goes through `BucketIdentifier.getBucketId(recordKey, "f1,f2", n)`.
The real divergence sources are (a) values containing , (Hudi doesn't escape) and (b) null/empty placeholder round-trip in composite mode. Both are now handled: the encoder produces `List<String>` elements element-wise equal to Hudi's parsed list, and rejects unrepresentable inputs (`,` in value, literal placeholder collision) up front.
Result: bucket id is bit-for-bit equal to Hudi's `BucketIdentifier` across all 21 tests, including the recordKey-string overload that exercises Hudi's production parse path.

XuQianJin-Stars:
+1

luoyuxia (approved):
Thanks @fhan688 @XuQianJin-Stars LGTM

PR #3322, opened 2026-05-15
[flink][spark] Fix partition pruning for non-string partition keys

### LINKED ISSUE #3292: [flink][spark] Partition pruning gives wrong results for non-STRING partition keys with range predicates
### Search before asking

### Description

When pushing down partition filters, both the Spark and Flink connectors stringify everything, predicate literals and partition values alike, before evaluating. That makes equality work fine, but range comparisons fall back to string order.                                             
                                                                                                                                                                                                                                                     
So with an INT partition column and partitions pt=2, pt=10, a query WHERE pt > 2 lexicographically compares "10" < "2" and drops pt=10. Rows go missing. 

 We can build the partition row with typed values using the existing PartitionUtils.parseValueOfType and skip stringification for both Flink/Spark.

### Willingness to contribute

fresh-borzoni:
## Summary

Closes #3292

Partition predicate pushdown stringified literals and partition values before evaluation, so range comparisons fell back to string lexicographic order.
Example: An INT partition column with values 2 and 10 under WHERE pt > 2 lex-compared "10" < "2" and incorrectly dropped partition 10.

Added PartitionUtils.toPartitionRow and PartitionUtils.partitionRowType in fluss-common.


fresh-borzoni:
@Yohahaha @YannByron @luoyuxia PTAL 🙏

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/source/enumerator/FlinkSourceEnumerator.java
```diff
@@ -583,14 +587,10 @@ private List<PartitionInfo> applyPartitionFilter(List<PartitionInfo> partitionIn
...
 
-    private static InternalRow toInternalRow(PartitionInfo partitionInfo) {
-        List<String> partitionValues =
-                partitionInfo.getResolvedPartitionSpec().getPartitionValues();
-        GenericRow genericRow = new GenericRow(partitionValues.size());
-        for (int i = 0; i < partitionValues.size(); i++) {
-            genericRow.setField(i, BinaryString.fromString(partitionValues.get(i)));
-        }
-        return genericRow;
+    private static InternalRow toInternalRow(
```
Yohahaha: nit: we could remove `toInternalRow` and use `PartitionUtils.toPartitionRow` directly

Yohahaha (approved):
parts of Spark LGTM! thank you for the fix!

fresh-borzoni:
@Yohahaha addressed 👍 
@luoyuxia do you mind to take a quick look, pls?

fluss-spark/fluss-spark-common/src/main/scala/org/apache/fluss/spark/utils/SparkPartitionPredicate.scala
```diff
@@ -53,54 +50,17 @@ object SparkPartitionPredicate {
...
       partitionInfos: Seq[PartitionInfo],
       partitionPredicate: Option[FlussPredicate]): Seq[PartitionInfo] =
     partitionPredicate match {
       case None => partitionInfos
-      case Some(predicate) => partitionInfos.filter(p => predicate.test(toPartitionRow(p)))
+      case Some(predicate) =>
+        val rowType = PartitionUtils.partitionRowType(tableInfo)
+        partitionInfos.filter {
+          p =>
+            predicate.test(
```
luoyuxia, line 62: For DATE, that produces Fluss internal Integer day values, and LeafPredicate.get(...) also reads DATE fields as int. 

However, Spark DATE literals are converted to LocalDate in SparkPredicateConverter, because RPC serialization expects that shape. So a query like: 
CREATE TABLE t (id BIGINT, dt DATE) PARTITIONED BY (dt); SELECT * FROM t WHERE dt = DATE '2026-01-02'; can hit LocalDate.compareTo(Integer) during local partition pruning and fail with ClassCastException instead of pruning correctly.

luoyuxia (commented):
@fresh-borzoni Thanks for the pr. Only one comment. PTAL

fluss-spark/fluss-spark-common/src/main/scala/org/apache/fluss/spark/utils/SparkPartitionPredicate.scala
fresh-borzoni: Hmm, checking, interesting finding, thank you!

fluss-common/src/main/java/org/apache/fluss/utils/PartitionUtils.java
```diff
@@ -345,4 +348,28 @@ public static String convertValueOfType(Object value, DataTypeRoot type) {
...
+     */
+    public static GenericRow toPartitionRow(
+            List<String> partitionValues, RowType partitionRowType) {
+        GenericRow row = new GenericRow(partitionValues.size());
+        for (int i = 0; i < partitionValues.size(); i++) {
+            DataTypeRoot type = partitionRowType.getTypeAt(i).getTypeRoot();
+            row.setField(i, parseValueOfType(partitionValues.get(i), type));
+        }
+        return row;
+    }
```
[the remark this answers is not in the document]
fresh-borzoni: it's not a good idea, as it will skip with this "doesn't match" skip fallback

fluss-spark/fluss-spark-common/src/main/scala/org/apache/fluss/spark/utils/SparkPartitionPredicate.scala
fresh-borzoni: Fixed for DATE and TIME additionally, also server pushdown was not applied for them at all and FlussAsSparkRow.getBinary had preexisting bug

fresh-borzoni:
@luoyuxia addressed comments, PTAL 🙏

luoyuxia (approved):
@fresh-borzoni Thanks. LGTM!

PR #3335, opened 2026-05-17
[client] Cache synchronously fetched schemas

### LINKED ISSUE #3336: [client] Avoid repeated synchronous schema fetches in ClientSchemaGetter
### Problem

`ClientSchemaGetter#getSchema` fetches missing schema IDs synchronously through `Admin#getTableSchema`, but the fetched schema is not stored in the local schema cache. If callers request the same schema ID again, the client can repeat the same admin metadata request even though the schema was already fetched successfully.

### Proposed improvement

Cache schemas fetched through the synchronous fallback path in `ClientSchemaGetter#getSchema`. This keeps repeated lookups for the same schema ID local after the first successful fetch while preserving the existing latest-schema update behavior.

### Benefit

This avoids unnecessary admin metadata calls for repeated schema lookups and makes the synchronous fetch path consistent with cache expectations.

### Validation

A focused unit test can verify that two `getSchema(schemaId)` calls for the same initially missing schema ID only trigger one admin fetch and that latest schema metadata is updated when the fetched schema is newer.

Leondon9:
### Purpose

Linked issue: close #3336

`ClientSchemaGetter#getSchema` already looked up missing schema IDs through `Admin#getTableSchema`, but the synchronously fetched schema was not stored in the local `schemasById` cache. As a result, repeated lookups for the same schema ID could issue repeated admin requests.

This PR caches schemas fetched through the synchronous fallback path. It also adds a focused unit test that verifies the first lookup fetches the schema, the second lookup reuses the cache, and the latest schema metadata is advanced when the fetched schema is newer.

### Brief change log

- Store synchronously fetched schemas in `ClientSchemaGetter`'s local schema cache.
- Add `ClientSchemaGetterTest` covering cache reuse and latest-schema update behavior.

### Why this is better

Repeated `getSchema(schemaId)` calls for a schema that was initially missing from the client cache no longer need to call `Admin#getTableSchema` every time. This reduces unnecessary metadata calls while preserving the existing latest-schema update behavior.

### Tests

- `env PATH=/usr/local/bin:/opt/homebrew/bin:/usr/bin:/bin:/usr/sbin ./mvnw -pl fluss-client -am -Dtest=ClientSchemaGetterTest -DfailIfNoTests=false test`
- `JAVA_HOME=$(/usr/libexec/java_home -v 11) ./mvnw clean verify` was attempted locally with Java 11.0.31. It failed in unrelated `fluss-server` integration test `TableManagerITCase.testPartitionedTableManagement`, where the expected date partition key was `20260518` but the actual key was `20260517`. The failure is in server coordinator date partition handling and does not overlap with this client schema cache change.

### API and Format

### Documentation

No documentation changes; this is an internal client metadata cache behavior fix.

### Generative AI disclosure

Yes. This PR was authored with assistance from OpenAI Codex and reviewed by the human contributor before submission.

Generated-by: OpenAI Codex following https://github.com/apache/fluss/blob/main/AGENTS.md

### DIFF
--- a/fluss-client/src/main/java/org/apache/fluss/client/metadata/ClientSchemaGetter.java
+++ b/fluss-client/src/main/java/org/apache/fluss/client/metadata/ClientSchemaGetter.java
@@ -67,6 +67,7 @@ public class ClientSchemaGetter implements SchemaGetter {
             try {
                 SchemaInfo schemaInfo =
                         admin.getTableSchema(tablePath, schemaId).get(1, TimeUnit.MINUTES);
+                schemasById.put(schemaId, schemaInfo.getSchema());
                 if (schemaId > latestSchemaInfo.getSchemaId()) {
                     latestSchemaInfo = schemaInfo;
                 }
--- a/fluss-client/src/test/java/org/apache/fluss/client/metadata/ClientSchemaGetterTest.java
+++ b/fluss-client/src/test/java/org/apache/fluss/client/metadata/ClientSchemaGetterTest.java
@@ -1,0 +1,95 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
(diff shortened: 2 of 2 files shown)


fxbing (approved):
LGTM. The synchronous cache population is consistent with the existing async path and fixes the repeated schema fetch issue. Thanks!

loserwang1024 (approved):
LGTM!

PR #3340, opened 2026-05-18
[server] Add disk-usage write protection to TabletServer

### LINKED ISSUE #3338: [server] Add disk-usage write protection to TabletServer
### Search before asking

### Motivation

When a TabletServer's local data disk is nearly full, continued writes can cause multiple errors, corrupt in-progress segments, or trigger cascading failures across replicas. Currently Fluss has no built-in mechanism to proactively reject writes before the disk runs out of space.

### Solution

Introduce a periodic disk-usage monitoring state machine with hysteresis-based write protection:

- Sampling: A background scheduler thread calls statvfs every `server.data-disk.check-interval` (default 30s) to collect the disk usage ratio. The overhead is negligible (~100μs per call, pure in-memory).
- Multi-disk strategy: When multiple data directories are configured, the collector reports the maximum usage across all distinct FileStores, so that a single nearly-full disk is never masked by other low-usage disks.
- Hysteresis state machine: Writes are locked when any disk's usage reaches server.data-disk.write-limit-ratio (default 0.85) and resume only after usage drops below (ratio - 0.10). The fixed 10% recovery gap prevents rapid lock/unlock oscillation.
- Write interception: Only client-driven writes (appendLog / putKv) are rejected with a retriable DiskWriteLockedException. Follower replication is not blocked to preserve replica consistency.
- Dynamic reconfiguration: `server.data-disk.write-limit-ratio` can be changed at runtime via alterConfigs(). Changes take effect immediately by triggering a synchronous disk check.

### Anything else?

_No response_

### Willingness to contribute

swuferhong:
### Purpose

Linked issue: close #3338 

Introduce a periodic disk-usage monitoring mechanism that proactively rejects client writes when the TabletServer's data disk usage exceeds a configurable high-water-mark ratio, preventing ENOSPC errors and potential data corruption.

Key design decisions:
- Hysteresis state machine with a fixed 10% recovery gap to avoid rapid lock/unlock oscillation (lock at limit, unlock at limit-0.10)
- Max-per-disk strategy: report the highest usage across all distinct FileStores so a single full disk is never masked by other low-usage disks in multi-disk deployments
- Only client-driven writes (appendLog/putKv) are rejected with a retriable DiskWriteLockedException; follower replication is not blocked to preserve replica consistency
- write-limit-ratio supports runtime dynamic reconfiguration via ServerReconfigurable, with an immediate re-check on change
- Setting ratio to 1.0 completely disables the protection

New configuration:
- server.data-disk.write-limit-ratio (default 0.85, dynamic)
- server.data-disk.check-interval (default 30s)

New metrics:
- diskUsageRatio: current disk usage ratio [0.0, 1.0]
- diskWriteLocked: 1 when writes are being rejected, 0 otherwise


zuston (commented):
If exceeding the disk usage ratio threshold (or disk corruption), do we need to make this tablet server as offline or unhealthy status? I think the writer side fencing is not enough, sometimes the disk usage exceeding will not recover automaticlly at the many cases

swuferhong:
> If exceeding the disk usage ratio threshold (or disk corruption), do we need to make this tablet server as offline or unhealthy status? I think the writer side fencing is not enough, sometimes the disk usage exceeding will not recover automaticlly at the many cases

Hi, @zuston. Writer-side fencing is the minimum-sufficient response for a capacity event; promoting it to node-level offline turns a localized capacity problem into a cluster-wide availability incident and triggers cascading failover. Disk corruption is a separate fault domain (IOException-driven Log Directory Failure) and should be addressed in a dedicated PR.

Happy to add a follow-up issue tracking the Log Directory Failure work if that helps.

fluss-server/src/main/java/org/apache/fluss/server/DynamicServerConfig.java
```diff
@@ -64,7 +65,8 @@ class DynamicServerConfig {
                             DATALAKE_FORMAT.key(),
                             LOG_REPLICA_MIN_IN_SYNC_REPLICAS_NUMBER.key(),
                             KV_SHARED_RATE_LIMITER_BYTES_PER_SEC.key(),
-                            KV_SNAPSHOT_INTERVAL.key()));
+                            KV_SNAPSHOT_INTERVAL.key(),
+                            SERVER_DATA_DISK_WRITE_LIMIT_RATIO.key()));
```
wuchong, line 69: This key now passes the coordinator allowlist, but the range check still exists only in `LocalDiskManager.validate()`, which is registered on TabletServer, not CoordinatorServer. Values like `0.0` or `1.5` can therefore be persisted through `AlterConfigs` and only fail later when tablet servers try to apply them. The coordinator path should reject invalid `server.data-disk.write-limit-ratio` updates up front.

I think we should also validate this on the Coordinator via `org.apache.fluss.server.DynamicConfigManager#registerValidator` by extending a `ConfigValidator`. We should also add an IT case for setting valid and invalid `server.data-disk.write-limit-ratio` (maybe near  `FlussAdminITCase#testDynamicConfigs()`).

fluss-server/src/main/java/org/apache/fluss/server/storage/DiskUsageCollector.java
```diff
@@ -0,0 +1,95 @@
...
+                LOG.warn("Failed to get FileStore for data directory {}; skipping.", dir, e);
+                failures++;
+                continue;
+            }
+            if (counted.add(fs)) {
+                long total = fs.getTotalSpace();
+                if (total <= 0L) {
+                    continue;
+                }
+                double ratio = (double) (total - fs.getUsableSpace()) / total;
```
wuchong: `collect()` only treats `Files.getFileStore()` failures as skippable. If `FileStore#getTotalSpace()` or `getUsableSpace()` throws for one data directory, the whole sample aborts instead of skipping just that directory, so `DiskUsageMonitor` can keep a stale lock state even when the other data dirs are still healthy and measurable. This should handle per-filesystem stat failures the same way as lookup failures.

fluss-server/src/main/java/org/apache/fluss/server/storage/LocalDiskManager.java
```diff
@@ -402,6 +463,101 @@ public List<File> dataDirs() {
...
+     * first scheduled tick.
+     */
+    public void startDiskUsageMonitor(Scheduler scheduler) {
+        // initial sample to avoid an open window before the first scheduled run
+        diskUsageMonitor.runOnce();
+        scheduler.schedule(
+                "disk-usage-monitor",
+                diskUsageMonitor::runOnce,
+                diskCheckIntervalMs,
+                diskCheckIntervalMs);
```
wuchong: nit: Setting `delayMs` to 0 can trigger immediate collection, rather than relying on an explicit invocation. This ensures the disk I/O operation executes asynchronously within the scheduler thread, preventing it from blocking the startup process.

wuchong (commented):
Thanks @swuferhong , I only left some minor comments.

swuferhong:
@wuchong  comments addressed.

wuchong (approved):

PR #3350, opened 2026-05-19
[flink] support offset lag

### LINKED ISSUE #3349: add support of offset lag
### Search before asking

### Motivation

Now we already have the timestamp lag in flink connector, but it's necessary to have offset lag especially for the long-term backfill

### Solution

_No response_

### Anything else?

_No response_

### Willingness to contribute

zuston:
### Purpose

Linked issue: close #3349 

Flink currently exposes `currentFetchEventTimeLag`, but it measures event-time lag rather than offset backlog. For Fluss log reading, we need a Kafka-style records lag metric so users can understand how far each subscribed source-reader bucket is behind.

Relying only on timestamp lag can be misleading, especially for backfill jobs that subscribe to multiple time partitions. Older partitions may always appear to have a large event-time lag, even when their records have already been fully consumed.


luoyuxia:
cc @loserwang1024

PR #3355, opened 2026-05-20
[flink] Preserve log-only source restore mode

### LINKED ISSUE #3354: [flink] Avoid initializing lake snapshot splits after log-only source restore
### Search before asking

### Fluss version

main (development)

### Please describe the bug 🐞

When a Flink source starts from a specified log position, such as timestamp/latest/earliest startup instead of full startup, `FlinkTableSource` correctly passes `null` as the `lakeSource` to `FlinkSourceEnumerator`. The source therefore initializes only Fluss log splits and does not initialize lake snapshot splits.

After the job is restored from a checkpoint/savepoint, the recreated source can pass a non-null `lakeSource` to `FlinkSourceEnumerator`. Because the enumerator checkpoint state currently stores `remainingHybridLakeFlussSplits` as `null` in the original log-only startup, the restored enumerator treats the state as if lake splits have not been generated yet. It then initializes lake snapshot splits during restore.

This changes the startup semantics across restore and can duplicate historical data: the original run starts from the specified Fluss log position, but the restored run additionally reads lake snapshot data.

Expected behavior: restoring a source that originally started in log-only mode should preserve that mode and must not initialize lake snapshot splits.

Actual behavior: restore can initialize lake snapshot splits if `lakeSource` becomes non-null, causing duplicated initialization/read of historical lake data.

### Solution

Use the existing `remainingHybridLakeFlussSplits` checkpoint field as a sentinel. For a source enumerator that starts with lake snapshot reading disabled, persist an empty list instead of `null`. On restore, `pendingHybridLakeFlussSplits != null` already means the enumerator should not list/generate lake splits again, so an empty list preserves the original log-only startup semantics without changing the enumerator state schema.

luoyuxia:
### Purpose

Linked issue: close #3354

Preserve the original Fluss-only (non-lake) startup semantics after a Flink source restore. When the source starts from a specified Fluss log position, it should not initialize lake snapshot splits after restore just because `lakeSource` is available again.

### Brief change log

- Persist an empty `remainingHybridLakeFlussSplits` list when the enumerator checkpoints with `lakeSource == null`.
- Allow empty pending hybrid split lists to deserialize without `lakeSource`, while still requiring `lakeSource` for non-empty hybrid split state.
- Document the three-state meaning of `pendingHybridLakeFlussSplits` so `null`, empty, and non-empty states are explicit.
- Add a restore regression test that checkpoints a Fluss-only enumerator, restores it with a non-null `lakeSource` and a registered lake snapshot, and verifies no `LakeSnapshotSplit` is generated.
- Add a serializer round-trip test for empty pending hybrid split state with `lakeSource == null`.

### Tests

- `./mvnw -pl fluss-flink/fluss-flink-common -Dtest=SourceEnumeratorStateSerializerTest,FlinkSourceEnumeratorTest#testRestoreFlussOnlySourceWithLakeSourceDoesNotGenerateLakeSplits test`
- `./mvnw -pl fluss-flink/fluss-flink-common spotless:apply`
- `git diff --check`

### API and Format

No public API, storage format, or checkpoint serializer version change. The fix only changes the value stored in the existing enumerator state field for Fluss-only source checkpoints and permits that empty-state sentinel to deserialize without a lake source.

### Documentation

No new feature or user-facing configuration change.

### Generative AI disclosure

- [x] Yes: OpenAI Codex was used to author this PR following the repository AGENTS.md guidance.

### DIFF
--- a/fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/source/enumerator/FlinkSourceEnumerator.java
+++ b/fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/source/enumerator/FlinkSourceEnumerator.java
@@ -526,9 +526,8 @@ public class FlinkSourceEnumerator
     private void startInStreamModeForNonPartitionedTable() {
         if (lakeSource != null) {
             // Generate lake splits synchronously so that they are available before the
-            // first checkpoint.  This is consistent with the partitioned-table path in
-            // start() and ensures generateHybridLakeFlussSplits() can safely use
-            // checkpointTriggeredBefore to distinguish fresh starts from restores.
+            // first checkpoint. This is consistent with the partitioned-table path in
+            // start().
             List<SourceSplitBase> splits = generateHybridLakeFlussSplits();
             if (splits == null) {
                 // no lake snapshot, fall back to normal Fluss splits
@@ -885,13 +884,6 @@ public class FlinkSourceEnumerator
         // without re-generating.
         if (pendingHybridLakeFlussSplits != null) {
             LOG.info("Still have pending lake fluss splits, shouldn't list splits again.");
-            return pendingHybridLakeFlussSplits;
-        }
-        // Restored from checkpoint but pending lake split is null(e.g. the source was
-        // originally started in Fluss-only mode without lake).  Do not generate lake
-        // splits for this restore; mark as initialized and return empty list.
-        if (checkpointTriggeredBefore) {
-            pendingHybridLakeFlussSplits = Collections.emptyList();
             return pendingHybridLakeFlussSplits;
         }
         try {
--- a/fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/source/enumerator/FlinkSourceEnumerator.java
+++ b/fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/source/enumerator/FlinkSourceEnumerator.java
@@ -140,7 +140,8 @@ public class FlinkSourceEnumerator
      * <p>The field has three states:
      *
      * <ul>
-     *   <li>{@code null}: lake split initialization has not run yet.
+     *   <li>{@code null}: lake split initialization has not run yet, or the source has no lake
+     *       (non-lake table) so initialization will never run.
      *   <li>empty list: lake split initialization has run, or this enumerator was started in
      *       Fluss-only (non-lake) mode and must not initialize lake splits after restore.
      *   <li>non-empty list: lake split initialization has run and these splits still need to be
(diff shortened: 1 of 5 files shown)


luoyuxia:
@loserwang1024 Could you please help review

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/source/enumerator/FlinkSourceEnumerator.java
```diff
@@ -1207,11 +1220,16 @@ public void addReader(int subtaskId) {
 
     @Override
     public SourceEnumeratorState snapshotState(long checkpointId) {
+        List<SourceSplitBase> remainingHybridLakeFlussSplits =
+                // Preserve Fluss-only (non-lake) startup across restore. Otherwise a restored
+                // enumerator with a non-null lakeSource would treat null as "not initialized yet"
+                // and generate lake snapshot splits.
+                lakeSource == null ? Collections.emptyList() : pendingHybridLakeFlussSplits;
```
loserwang1024: I understand what you mean, but the code seems hard to understand without sufficient context. I've also thought about this:

- If the enumerator is created by FlinkSource#createEnumerator, it indicates a stateless restart. Therefore, whether to generate lake splits depends on whether it's a LakeSource.

- If the enumerator is created by FlinkSource#restoreEnumerator, there's no need to generate lake splits again. This is because before the first checkpoint is taken, FlinkSourceEnumerator#start → FlinkSourceEnumerator#generateHybridLakeFlussSplits has already been executed. Thus, upon restoration, the lake splits do not need to be regenerated.

Therefore, even if the job was previously started from a specified timestamp, according to this logic, as long as a checkpoint has been taken, upon stateful restart it will not read the lake splits again.

luoyuxia: I agree the original approach in snapshotState() is hard to understand without sufficient context.

I've reworked the fix to use the checkpointTriggeredBefore flag in generateHybridLakeFlussSplits() instead. While the restore-awareness logic still requires some thought, this approach keeps all the complexity contained within a single method rather than spreading it across snapshotState().

Additionally, I changed startInStreamModeForNonPartitionedTable to call generateHybridLakeFlussSplits() synchronously, consistent with the partitioned-table path in start(). This ensures lake split initialization always completes before any checkpoint can be triggered, which is a prerequisite for the checkpointTriggeredBefore guard to work correctly.

luoyuxia:
@loserwang1024 Comments has been addressed.

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/source/enumerator/FlinkSourceEnumerator.java
```diff
@@ -869,13 +881,19 @@ private List<SourceSplitBase> getLogSplit(
...
+        // Restored from checkpoint with pending lake splits — return them directly
+        // without re-generating.
         if (pendingHybridLakeFlussSplits != null) {
             LOG.info("Still have pending lake fluss splits, shouldn't list splits again.");
             return pendingHybridLakeFlussSplits;
         }
+        // Restored from checkpoint but pending lake split is null(e.g. the source was
+        // originally started in Fluss-only mode without lake).  Do not generate lake
+        // splits for this restore; mark as initialized and return empty list.
+        if (checkpointTriggeredBefore) {
```
fresh-borzoni: I think the checkpointTriggeredBefore guard is solving the restore ambiguity in the wrong layer.

Fresh null means “not initialized yet”.
Restored null should be normalized to “nothing pending”.

That normalization belongs in FlinkSource.restoreEnumerator(), not in generateHybridLakeFlussSplits().
WDYT?

fresh-borzoni (commented):
@luoyuxia Ty for the PR, I agree with the sync change and the serializer reorder, but I have comment about checkpointTriggeredBefore guard.

luoyuxia:
cc @loserwang1024 @loserwang1024 Please help review again.

fresh-borzoni (approved):
@luoyuxia Ty, LGTM 👍

luoyuxia:
@loserwang1024 I'm going to merge this. Feel free to left any comment if you still have

PR #3356, opened 2026-05-21
[flink] Include lake reader in source metrics and correct max timestamp lag semantics

### LINKED ISSUE #3304: [flink] currentFetchEventTimeLag is misleading when subscribing multiple partitions/buckets with uneven lag
### Search before asking

### Fluss version

0.9.0 (latest release)

### Please describe the bug 🐞

When a Flink source subscribes to multiple partition tables (or multiple buckets) whose consuming progress differs significantly, the reported `currentFetchEventTimeLag` metric is much smaller than the actual `currentEmitEventTimeLag`. This makes the metric misleading for monitoring and alerting, because:

- `currentFetchEventTimeLag` stays near 0, suggesting the source has caught up.
- `currentEmitEventTimeLag` reports a large lag (e.g. 2 hours), suggesting records are piling up.
- Yet the Flink job has **no backpressure**, so the gap cannot be explained by downstream slowness.

## Root Cause

In [`FlinkSourceSplitReader#forLogRecords`](file:///Users/loserwang/workstation/github/fluss/fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/source/reader/FlinkSourceSplitReader.java#L426-L494):

```java
maxConsumerRecordTimestampInFetch =
        Math.max(maxConsumerRecordTimestampInFetch, lastRecord.timestamp());
...
flinkSourceReaderMetrics.reportRecordEventTime(
        fetchTimestamp - maxConsumerRecordTimestampInFetch);
```

The aggregated timestamp is the **MAX** across all buckets in the fetch, so the reported lag is effectively the **MIN** lag across buckets.

### Example scenario

- Partition A is 2 hours behind → its last record timestamp is 2 hours old.
- Partition B is reading the latest data → its last record timestamp ≈ `now`.

Every `logScanner.poll(POLL_TIMEOUT)` tends to return partition B's fresh data (since it is always locally ready as a `CompletedFetch`). The `Math.max` across buckets then picks partition B's near-`now` timestamp, so:

```
fetchTimestamp - maxTimestamp ≈ 0   → currentFetchEventTimeLag reports ~0
```

Meanwhile, emit lag is computed **per record** and aggregated as **max**, so partition A's 2-hour-old records correctly inflate `currentEmitEventTimeLag`. This is why the two metrics diverge dramatically even though there is no backpressure.

### Solution

1. **Align fetch lag semantics with emit lag (max-lag across buckets):**
   Change the aggregation in `forLogRecords` from `Math.max` to `Math.min` on record timestamp (equivalent to `max` on lag), so `currentFetchEventTimeLag` reports the worst-case lag across buckets in the fetch.

2. **Expose per-bucket / per-split fetch lag:**
   Add a new gauge `currentFetchEventTimeLag` under the existing per-bucket metric group (next to `currentOffset`):
   - Non-partitioned: `fluss.reader.bucket.{bucket_id}.currentFetchEventTimeLag`
   - Partitioned: `fluss.reader.partition.{partition_id}.bucket.{bucket_id}.currentFetchEventTimeLag`

   This allows users to observe exactly which partition/bucket is lagging, making diagnosis like the scenario above straightforward.

zuston:
### Purpose

Linked issue: close #3304 

### Brief change log

1. Add the lake reader into the source metrics, this is necessary for the long-term backfill when using the fluss multi-layer log table as the source
2. Correct the max timestamp log semantics on subscribing the multi buckets or partitions. that is mentioned in #3304 
3. Make these metrics collecting logic into one place in `FlinkRecordsWithSplitIds`

### DIFF
--- a/fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/lake/reader/IndexedLakeSplitRecordIterator.java
+++ b/fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/lake/reader/IndexedLakeSplitRecordIterator.java
@@ -18,8 +18,8 @@
 
 package org.apache.fluss.flink.lake.reader;
 
+import org.apache.fluss.flink.source.reader.LogRecordRowIterator;
 import org.apache.fluss.record.LogRecord;
-import org.apache.fluss.row.InternalRow;
 import org.apache.fluss.utils.CloseableIterator;
 
 /**
@@ -39,7 +39,7 @@ import org.apache.fluss.utils.CloseableIterator;
  *   <li>Maintains iterator semantics for sequential data access
  * </ul>
  */
-public class IndexedLakeSplitRecordIterator implements CloseableIterator<InternalRow> {
+public class IndexedLakeSplitRecordIterator implements LogRecordRowIterator {
     private final CloseableIterator<LogRecord> logRecordIterators;
     private final int currentLakeSplitIndex;
 
@@ -64,7 +64,7 @@ public class IndexedLakeSplitRecordIterator implements CloseableIterator<Interna
     }
 
     @Override
-    public InternalRow next() {
-        return logRecordIterators.next().getRow();
+    public LogRecord nextLogRecord() {
+        return logRecordIterators.next();
     }
 }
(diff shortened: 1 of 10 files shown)


zuston:
cc @swuferhong @luoyuxia

PR #3358, opened 2026-05-22
[server] Introduce more table level options 

### LINKED ISSUE #3315: Respect the log.* config from the table level options
### Search before asking

### Motivation

I found that `log.segment.file-size` currently only respects server-level configurations and is ignored when defined at the table level. I would like to propose a mechanism that allows the system to prioritize table-specific configurations if they exist, falling back to the global server options only as a default. This would provide better flexibility for handling tables with unique throughput or storage requirements.

### Solution

_No response_

### Anything else?

_No response_

### Willingness to contribute

zuston:
### Purpose

Linked issue: close #3315

When optimizing one PK table performance with RocksDB compression, I found that some table-related options, such as `kv.rocksdb.xxx` and `log.xxx`, are only available at the server level. This makes it difficult to tune options per table, especially when different tables have different workloads. 

Therefore, this PR adds support for respecting these options at the table level.

### Brief Change Log

This PR introduces table-level options while preserving backward compatibility: if a table-level option is not set, the server-level option will be used as the fallback.

To centralize and simplify this fallback logic, this PR introduces `ResolvedTableConfig`.


fluss-server/src/main/java/org/apache/fluss/server/utils/TableDescriptorValidation.java
```diff
@@ -417,6 +418,17 @@ private static void checkTieredLog(Configuration tableConf) {
...
+            throw new InvalidConfigException(
+                    String.format(
+                            "Invalid configuration for %s, it must be less than or equal %d bytes.",
+                            ConfigOptions.LOG_SEGMENT_FILE_SIZE.key(), Integer.MAX_VALUE));
+        }
+    }
+
     private static void checkPartition(
             Configuration tableConf, List<String> partitionKeys, RowType rowType) {
         boolean isPartitioned = !partitionKeys.isEmpty();
```
zuston, line 476: When surfacing this part of the code, could we move the option validation logic into the option definitions themselves? For example:

```java
public static final ConfigOption<Integer> TABLE_TIERED_LOG_LOCAL_SEGMENTS =
            key("table.log.tiered.local-segments")
                    .intType()
                    .defaultValue(2)
                    .valueValidator(v -> v > 0, "must be greater than zero")
                    .withDescription(
                            "The number of log segments to retain in local for each table when log tiered storage is enabled. "
                                    + "It must be greater that 0. The default is 2.");
```

zuston: cc @wuchong

fluss-common/src/main/java/org/apache/fluss/config/FlussConfigUtils.java
```diff
@@ -43,6 +43,8 @@ public class FlussConfigUtils {
 
     static {
         TABLE_OPTIONS = extractConfigOptions("table.");
+        TABLE_OPTIONS.put(
```
lilei1128: Could log.index.file-size also be supported？

zuston: sure. this is the first phase for this PR

zuston:
could you help review this? @luoyuxia

zuston:

fresh-borzoni (commented):
@zuston Thank you for the PR, took a look.

The tests are failing because of the isTableStorageConfig change (startsWith("table.") -> TABLE_OPTIONS.containsKey(key)). 
That drops props like table.datalake.paimon.warehouse, which breaks testAlterTableConfig, testCreateTableWithUnknownOptions, and the lake FlinkUnionReadDvTableITCase.

Could you take a look and fix? Shall we make it more additive?

zuston:
@fresh-borzoni Sorry for the late reply. I’ve updated the description and expanded the scope of the changes. Please take another look when you have time. Thanks!

Issue #3357, opened 2026-05-22
[server] Table deletion stuck permanently when StopReplica request fails


gyang94:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/fluss/issues) and found nothing similar.

### Fluss version

0.9.0 (latest release)

## Summary

When a table is deleted and the `StopReplicaRequest` fails to reach the target TabletServer (e.g., due to network issues or TabletServer being down), the deletion process gets permanently stuck. The replica state remains in `ReplicaDeletionStarted`, blocking any retry. This causes `fluss_coordinator_tableCount` to be higher than the actual tables in ZooKeeper, with no self-healing mechanism.

### The Deletion State Machine

```
OnlineReplica → OfflineReplica → ReplicaDeletionStarted → ReplicaDeletionSuccessful → NonExistentReplica
```

The `completeDeleteTable` (which calls `coordinatorContext.removeTable()` and decreases tableCount) is only invoked when **all** replicas reach `ReplicaDeletionSuccessful`.

### The Deadlock

In `CoordinatorRequestBatch.sendStopRequest()` (line 465-469):

```java
if (throwable != null) {
    // todo: in FLUSS-55886145, we will introduce a sender thread to send the request.
    // in here, we just ignore the error.
    LOG.warn("Failed to send stop replica request to tablet server {}.", serverId, throwable);
    return;  // ← No event produced, no state transition
}
```

When the request **fails to send** (network error, TabletServer down):
1. No `DeleteReplicaResponseReceivedEvent` is produced
2. Replica state stays at `ReplicaDeletionStarted` permanently
3. `isEligibleForDeletion()` checks `!coordinatorContext.isAnyReplicaInState(tableId, ReplicaDeletionStarted)` — returns **false**
4. Even if `resumeDeletions()` is called later, it cannot retry because the eligibility check fails
5. **Permanent deadlock** — no code path can move the replica out of `ReplicaDeletionStarted`

### Contrast with Successful Response Error

When the request is sent successfully but the TabletServer returns an error response, the flow works correctly:
- `DeleteReplicaResponseReceivedEvent` is produced
- `retryDeleteAndSuccessDeleteReplicas` handles the failure
- Replica state transitions to allow retry

The bug is specifically in the **request send failure** path (network-level failure or tablet server down).

### Are you willing to submit a PR?

- [ ] I'm willing to submit a PR!

PR #3359, opened 2026-05-22
[server] Fix table deletion stuck when StopReplicaRequest send fails

lilei1128:
### Purpose

Linked issue: close #xxx

When a StopReplicaRequest send fails with a network-level throwable,
  the coordinator had already transitioned replicas to ReplicaDeletionStarted
  but silently dropped the error. This permanently stuck table deletion because:
  - isEligibleForDeletion() blocks tables with any replica in ReplicaDeletionStarted
  - resumeDeletions() only triggers from processDeleteReplicaResponseReceived,
    which is only reached when a DeleteReplicaResponseReceivedEvent is emitted
  - processDeadTabletServer() skips replicas that are toBeDeleted

### Brief change log

  Fix: when sendStopRequest() fails with a throwable, emit a
  DeleteReplicaResponseReceivedEvent with error results for the
  delete-flagged buckets (delete=true && deleteRemote=true). This feeds
  into the existing retry/give-up logic in processDeleteReplicaResponseReceived,
  which retries up to DELETE_TRY_TIMES (5) and then force-marks replicas as
  ReplicaDeletionSuccessful, allowing deletion to complete.

  Non-deletion stop replicas (delete=false) continue to silently ignore
  send failures as before, since they are best-effort.

### Tests

  Add testDeleteTableWithSendFailure() to verify that deletion completes
  even when all StopReplicaRequest sends fail at the network level (i.e.,
  CompletableFuture.failedFuture, not a response-level error code). The
  existing testDeleteReplicaStateChange() only covered response-level errors.

### DIFF
--- a/fluss-server/src/main/java/org/apache/fluss/server/coordinator/CoordinatorRequestBatch.java
+++ b/fluss-server/src/main/java/org/apache/fluss/server/coordinator/CoordinatorRequestBatch.java
@@ -466,11 +466,27 @@ public class CoordinatorRequestBatch {
                         if (throwable != null) {
                             // todo: in FLUSS-55886145, we will introduce a sender thread to send
                             // the request.
-                            // in here, we just ignore the error.
                             LOG.warn(
                                     "Failed to send stop replica request to tablet server {}.",
                                     serverId,
                                     throwable);
+                            // For delete replicas (delete=true && deleteRemote=true), we must emit
+                            // a failure event so the retry/give-up logic in
+                            // processDeleteReplicaResponseReceived can move the replica out of
+                            // ReplicaDeletionStarted. Without this, replicas get permanently stuck
+                            // in ReplicaDeletionStarted with no code path to recover.
+                            if (!deletedReplicaBuckets.isEmpty()) {
+                                List<DeleteReplicaResultForBucket> failedResults =
+                                        new ArrayList<>();
+                                ApiError apiError = ApiError.fromThrowable(throwable);
+                                for (TableBucket bucket : deletedReplicaBuckets) {
+                                    failedResults.add(
+                                            new DeleteReplicaResultForBucket(
+                                                    bucket, serverId, apiError));
+                                }
+                                eventManager.put(
+                                        new DeleteReplicaResponseReceivedEvent(failedResults));
+                            }
                             return;
                         }
                         // handle the response
--- a/fluss-server/src/main/java/org/apache/fluss/server/log/LogTablet.java
+++ b/fluss-server/src/main/java/org/apache/fluss/server/log/LogTablet.java
@@ -316,8 +316,42 @@ public final class LogTablet {
             Clock clock,
             boolean isCleanShutdown)
             throws Exception {
+        return create(
+                dataDir,
+                tablePath,
+                tabletDir,
+                conf,
+                new Configuration(),
+                serverMetricGroup,
+                recoveryPoint,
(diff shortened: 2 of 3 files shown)


lilei1128:
close #3357

gyang94:
Hi @lilei1128 , thanks for your contribution to fix this issue. Glad to see you are interested in this task.

In your PR, it is a good thinking to reuse the existing retry mechanism to solve this problem. However, the current retry loop doesn't truly solve the problem. The retry loop has no delay between attempts, so it degrades to "fail 5 times in milliseconds → pretend success → data residue on disk."

I'd like to solve this problem by introducing "**per-TS sender**" + "**ReplicaDeletionIneligible state**": 
- Per-TS sender thread gives each TabletServer an independent request queue with backoff-aware retry. A temporarily unreachable TS doesn't burn through retry attempts instantly — the sender sleeps and retries over a meaningful time window.
- ReplicaDeletionIneligible state breaks the current all-or-nothing dilemma. Instead of   force-marking unreachable replicas as successfully deleted, they enter Ineligible —   which (1) unblocks isEligibleForDeletion (since Ineligible ≠ DeletionStarted, resumeDeletions is no longer stuck), and (2) preserves the truth that deletion hasn't happened yet. When the TS comes back online, replicas transition Ineligible → OfflineReplica → ReplicaDeletionStarted and get genuinely deleted. 

If you are interested, welcome to review this [PR](https://github.com/apache/fluss/pull/3391), leave any comments there, or help to push code to fix based on it.

lilei1128:
> Hi @lilei1128 , thanks for your contribution to fix this issue. Glad to see you are interested in this task.
> 
> In your PR, it is a good thinking to reuse the existing retry mechanism to solve this problem. However, the current retry loop doesn't truly solve the problem. The retry loop has no delay between attempts, so it degrades to "fail 5 times in milliseconds → pretend success → data residue on disk."
> 
> I'd like to solve this problem by introducing "**per-TS sender**" + "**ReplicaDeletionIneligible state**":
> 
> * Per-TS sender thread gives each TabletServer an independent request queue with backoff-aware retry. A temporarily unreachable TS doesn't burn through retry attempts instantly — the sender sleeps and retries over a meaningful time window.
> * ReplicaDeletionIneligible state breaks the current all-or-nothing dilemma. Instead of   force-marking unreachable replicas as successfully deleted, they enter Ineligible —   which (1) unblocks isEligibleForDeletion (since Ineligible ≠ DeletionStarted, resumeDeletions is no longer stuck), and (2) preserves the truth that deletion hasn't happened yet. When the TS comes back online, replicas transition Ineligible → OfflineReplica → ReplicaDeletionStarted and get genuinely deleted.
> 
> If you are interested, welcome to review this [PR](https://github.com/apache/fluss/pull/3391), leave any comments there, or help to push code to fix based on it.


Thank you so much for your detailed and insightful review! I really appreciate you taking the time to explain the flaws in the current retry approach and sharing the better design direction.

I fully agree with your points: the existing retry loop without delay can only cause rapid failures and leave data residues on disk, which doesn’t solve the root problem. The solution you proposed — combining **per-TS sender** with backoff-aware retry and **ReplicaDeletionIneligible** state — is elegant and thorough. It not only unblocks the deletion workflow but also keeps the correct state transition logic, which perfectly fixes the all-or-nothing dilemma.

I’m very interested in this design. I will take a close look at your PR.
Thanks again for your guidance!

PR #3361, opened 2026-05-22
[server] Allow table-level overrides for log segment/index/flush options

### LINKED ISSUE #3315: Respect the log.* config from the table level options

lilei1128:
### Purpose

Linked issue: close #3315 

  Add five new table-scoped ConfigOptions (table.log.segment.file-size,
  table.log.index.file-size, table.log.index.interval-size,
  table.log.file-preallocate, table.log.flush.interval-messages) that
  override the corresponding cluster-level log.* settings for a single
  table when explicitly set.  Tables without these properties continue
  to use the server-global values unchanged.

### Brief change log

  - ConfigOptions: add TABLE_LOG_* options with noDefaultValue() so
    absent values fall back to server config automatically
  - FlussConfigUtils: register all five keys in ALTERABLE_TABLE_OPTIONS
    so ALTER TABLE SET takes effect on the next segment roll
  - TableDescriptorValidation: add overflow guard for
    table.log.segment.file-size (mirrors existing server-level check)
  - LogTablet: add buildEffectiveLogConfig() helper and a new create()
    overload that accepts tableProperties; old overload delegates with
    an empty Configuration for full backward compatibility
  - LogManager: thread tableProperties through getOrCreateLog() and
    loadLog() so both the online and recovery paths honour overrides
  - Replica: pass tableInfo.getProperties() into getOrCreateLog()

### Tests

  - LogTabletTest: verify segment roll and auto-flush behaviour with
    table-level overrides vs. server defaults
  - TableDescriptorValidationTest: cover valid values, Integer.MAX_VALUE
    boundary, overflow rejection, and TABLE_OPTIONS registration for
    all five new keys

### DIFF
--- a/fluss-common/src/main/java/org/apache/fluss/config/ConfigOptions.java
+++ b/fluss-common/src/main/java/org/apache/fluss/config/ConfigOptions.java
@@ -1557,6 +1557,56 @@ public class ConfigOptions {
                             "The number of log segments to retain in local for each table when log tiered storage is enabled. "
                                     + "It must be greater that 0. The default is 2.");
 
+    public static final ConfigOption<MemorySize> TABLE_LOG_SEGMENT_FILE_SIZE =
+            key("table.log.segment.file-size")
+                    .memoryType()
+                    .noDefaultValue()
+                    .withDescription(
+                            "The segment file size for the log of this table. "
+                                    + "When set, overrides the cluster-level '"
+                                    + LOG_SEGMENT_FILE_SIZE.key()
+                                    + "' for this table only.");
+
+    public static final ConfigOption<MemorySize> TABLE_LOG_INDEX_FILE_SIZE =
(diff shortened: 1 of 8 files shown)


zuston:
thanks for proposing this. I have submitted the PR #3358  . If you have time, could you help review this?

lilei1128:
> thanks for proposing this. I have submitted the PR #3358 . If you have time, could you help review this?

Thanks for your commit and I will turn off this PR .

Issue #3362, opened 2026-05-22
clearup useless fluss partition info in writer metadata cache


loserwang1024:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/fluss/issues) and found nothing similar.

### Fluss version

0.9.0 (latest release)

### Please describe the bug 🐞

Currently, if a partition written job run for a long time, fluss partition info will increase more and more, and never be cleanup. Most of the partition info is no longer used. This will occupy a lot cpu usage for `Sender#sendWriteData ->Sender#sendWriteData -> writeBatches.entrySet()` and `RecordAccumulator#drain - RecordAccumulator#drainBatchesForOneNode -> Cluster#getAvailableBucketsForPhysicalTablePath` 

Only restart the job, need to clearup:
* useless PhysicalTablePath in org.apache.fluss.client.write.RecordAccumulator#writeBatches
* uneless parittion info in Cluster

### Are you willing to submit a PR?

- [ ] I'm willing to submit a PR!

PR #3367, opened 2026-05-23
[client] Fix unbounded growth of stale partition entries in writer metadata cache

lilei1128:
### Purpose

Linked issue: close #xxx

  When partitions are continuously created and dropped, physical table path
  entries in RecordAccumulator.writeBatches and Cluster.partitionsIdByPath
  were never removed, causing unbounded memory growth and CPU waste from
  iterating over stale entries in ready() and drain() hot loops.

### Brief change log

  Fix 1 (RecordAccumulator): introduce markPathsAsStale() and
  removeStalePathIfEmpty() with a dedicated staleLock so that the
  check-then-remove is atomic with respect to concurrent append() calls.
  New appends to stale paths are rejected immediately.

  Fix 2 (MetadataUtils): during partial metadata updates, remove stale
  partition entries for any table whose partition list was refreshed, so
  that dropped partitions no longer linger in partitionsIdByPath.

  Fix 3 (Sender): call cleanupStaleWriteBatches() on every sendWriteData()
  cycle to mark paths absent from the cluster as stale and remove them once
  their deques are fully drained.

### Tests

  - RecordAccumulatorTest.testMarkAndRemoveStalePathAfterDrain: verifies
    that append() is rejected after markPathsAsStale(), and that
    removeStalePathIfEmpty() succeeds once the deque is drained.
  - RecordAccumulatorTest.testRemoveStalePathReturnsFalseWhenDequeNonEmpty:
    verifies that removeStalePathIfEmpty() returns false and preserves the
    path when pending batches remain.
  - SenderTest.testSenderCleansUpStaleWriteBatchesAfterMetadataUpdate:
    verifies end-to-end that runOnce() removes a stale path from
    writeBatches after the cluster no longer contains it and its deque
    has been drained.

### DIFF
--- a/fluss-client/src/main/java/org/apache/fluss/client/utils/MetadataUtils.java
+++ b/fluss-client/src/main/java/org/apache/fluss/client/utils/MetadataUtils.java
@@ -44,6 +44,7 @@ import javax.annotation.Nullable;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Random;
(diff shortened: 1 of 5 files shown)


lilei1128:
close #3362

PR #3368, opened 2026-05-24
[common] Fix resource leak and silenced exceptions in FileLock

XuQianJin-Stars:
## Purpose

`FileLock#tryLock()` previously caught every `Exception` and returned `false`, which had two issues: (1) genuine I/O failures were indistinguishable from inter-process lock contention, hiding real problems from callers; and (2) when lock acquisition failed *after* `init()` had opened the `FileOutputStream`, that stream was never closed, leaking the underlying file descriptor for the lifetime of the JVM.

`FileLock#unlockAndDestroy()` also chained `unlock()`, channel close and stream close inside a single `try` block; if releasing the lock or closing the channel threw, the `FileOutputStream` was never closed and the lock file might not be deleted.

This PR fixes both bugs and adds a dedicated test class for `FileLock`.

Linked issue: close #xxx

## Brief change log

- `FileLock#tryLock()`
  - Catch `OverlappingFileLockException` explicitly to preserve the documented "return `false` on contention" semantics.
  - Propagate `IOException` / `RuntimeException` so callers can react to real failures instead of silently treating them as contention.
  - Close the `FileOutputStream` opened by `init()` whenever lock acquisition fails, preventing the file-descriptor leak.
- `FileLock#unlockAndDestroy()`
  - Rewritten with nested `try` / `finally` blocks so the channel and the output stream are each closed independently, and the lock file is always deleted at the end — even if releasing the lock or closing the channel throws.
- No behavioral change for the happy path; `NetUtilsTest` (the only production caller) still passes unchanged.

## Tests

Added `FileLockTest` in `fluss-common`, covering:

- happy path: `tryLock()` → `unlock()` → `unlockAndDestroy()` and lock file is removed.
- in-JVM contention: a second `tryLock()` on the same path returns `false` and does **not** throw or leak the stream.
- reuse of the same `FileLock` instance after `unlockAndDestroy()` (re-acquire works).
- `unlockAndDestroy()` is safe to call when `tryLock()` was never invoked.
- `isValid()` returns `false` before any `tryLock()`.
- constructor rejects file names composed entirely of illegal characters.

Existing tests:

- `NetUtilsTest` (the only production user of `FileLock`) continues to pass.

## API and Format

- Public API: **no changes**. Method signatures and contracts of `FileLock` are preserved; only internal exception handling and resource cleanup are corrected.
- Wire / serialization format: not affected.
- Configuration: not affected.

## Documentation

- No user-facing documentation changes required.
- Method-level Javadoc on `tryLock()` already states that `false` means the lock is held elsewhere; the new behavior aligns the implementation with that contract (real I/O errors are no longer hidden behind `false`).

### DIFF
--- a/fluss-common/src/main/java/org/apache/fluss/utils/FileLock.java
+++ b/fluss-common/src/main/java/org/apache/fluss/utils/FileLock.java
@@ -19,6 +19,9 @@ package org.apache.fluss.utils;
 
 import org.apache.fluss.annotation.Internal;
 
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
 import java.io.File;
 import java.io.FileOutputStream;
 import java.io.IOException;
@@ -31,6 +34,8 @@ import static org.apache.fluss.utils.Preconditions.checkNotNull;
 /** A file lock used for avoiding race condition among multiple threads/processes. */
 @Internal
 public class FileLock {
+    private static final Logger LOG = LoggerFactory.getLogger(FileLock.class);
+
     private static final String TEMP_DIR = System.getProperty("java.io.tmpdir");
     private final File file;
     private FileOutputStream outputStream;
@@ -151,8 +156,10 @@ public class FileLock {
         }
         try {
             outputStream.close();
-        } catch (IOException ignored) {
+        } catch (IOException e) {
             // Best-effort cleanup; the original failure will be propagated by the caller.
+            // Log here so users are aware that closing the output stream also failed.
+            LOG.warn("Failed to close output stream for file lock {}", file, e);
         } finally {
             outputStream = null;
         }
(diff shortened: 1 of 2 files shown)


fluss-common/src/main/java/org/apache/fluss/utils/FileLock.java
```diff
@@ -106,21 +122,42 @@ public void unlock() throws IOException {
...
     }
 
+    private void closeOutputStreamQuietly() {
+        if (outputStream == null) {
+            return;
+        }
+        try {
+            outputStream.close();
+        } catch (IOException ignored) {
+            // Best-effort cleanup; the original failure will be propagated by the caller.
```
Jackeyzhe, line 170: nit: Maybe we should add a log here ensure user know the outputStream close fail

XuQianJin-Stars: > nit: Maybe we should add a log here ensure user know the outputStream close fail

done

Jackeyzhe (approved):
I don't have other problem, LGTM

PR #3383, opened 2026-05-26
[Flink] Kvscan flink integration

polyzos:
#3126 extended the java client to support kvscan for the live rocksdb table.

This PR integrates that functionality inside the flink connector

Introduces `KvBatchSplit` and `KvBatchSplitState` as a new split type in the Flink source, enabling the connector to perform bounded full-table scans on primary-key tables via the server-side KV scan API (FIP-17), rather than reading from snapshots.


website/docs/engine-flink/reads.md
```diff
@@ -234,6 +234,51 @@ The server evaluates these predicates against per-batch column statistics and sk
...
 ## Batch Read
 
+### Full Scan of Primary Key Tables
+
+Fluss can perform a bounded full-table scan on a primary-key table directly via the server-side KV scan API.
+
+Enable the feature by setting `client.scanner.kv.server-side.enabled = true` on the table or as a SQL hint:
+
+- This is a **bounded** read. The source finishes once all buckets have been drained and does not continue reading the change-log.
+- On task restart, each bucket is rescanned from scratch. Progress within a scan session is not checkpointed, because an expired or invalidated server-side session cannot be resumed from a mid-point.
```
binary-signal: On task restart, each bucket is scanned again from the beginning. Progress within an active scan session is not checkpointed because expired or invalidated server-side sessions cannot be resumed from an intermediate position.

website/docs/engine-flink/reads.md
```diff
@@ -234,6 +234,51 @@ The server evaluates these predicates against per-batch column statistics and sk
...
+- The feature is disabled by default (`false`). Without it, unbounded (streaming) reads on primary-key tables work as usual; bounded reads require the data-lake integration to be enabled.
```
binary-signal: When disabled, unbounded (streaming) reads on primary-key tables continue to work as usual. Bounded reads require data-lake integration unless server-side KV scanning is enabled.

website/docs/engine-flink/reads.md
```diff
@@ -234,6 +234,51 @@ The server evaluates these predicates against per-batch column statistics and sk
...
+    (3, 'Carol', 'ap-south');
+```
+
+**3. Run a full scan in batch mode:**
+```sql title="Flink SQL"
+SET 'execution.runtime-mode' = 'batch';
+SELECT * FROM pk_table;
+```
+
+You can also enable the feature dynamically without storing it in the table metadata:
```
binary-signal: You can also enable server-side scanning dynamically without storing the option in table metadata:

website/docs/engine-flink/reads.md
```diff
@@ -234,6 +234,51 @@ The server evaluates these predicates against per-batch column statistics and sk
...
+```sql title="Flink SQL"
+SET 'execution.runtime-mode' = 'batch';
+SELECT * FROM pk_table /*+ OPTIONS('client.scanner.kv.server-side.enabled' = 'true') */;
+```
+
 ### Limit Read
 The Fluss source supports limiting reads for both primary-key tables and log tables, making it convenient to preview the latest `N` records in a table.
```
binary-signal, line 327: making it easy to preview the latest N records in a table.

website/docs/engine-flink/reads.md
```diff
@@ -234,6 +234,51 @@ The server evaluates these predicates against per-batch column statistics and sk
 
 ## Batch Read
 
+### Full Scan of Primary Key Tables
+
+Fluss can perform a bounded full-table scan on a primary-key table directly via the server-side KV scan API.
+
+Enable the feature by setting `client.scanner.kv.server-side.enabled = true` on the table or as a SQL hint:
```
binary-signal: Enable this feature by setting `client.scanner.kv.server-side.enabled = true` in the table options or by using a SQL hint.

website/docs/engine-flink/reads.md
```diff
@@ -234,6 +234,51 @@ The server evaluates these predicates against per-batch column statistics and sk
...
+
+- This is a **bounded** read. The source finishes once all buckets have been drained and does not continue reading the change-log.
```
binary-signal: The source finishes after all buckets have been scanned and does not continue consuming the change log.

website/docs/engine-flink/options.md
```diff
@@ -102,6 +102,7 @@ See more details about [ALTER TABLE ... SET](engine-flink/ddl.md#set-properties)
...
 | scan.kv.snapshot.lease.duration               | Duration   | 1day                                            | The time period how long to wait before expiring the kv snapshot lease to avoid kv snapshot blocking to delete.                                                                                                                                                                                                                                                                                                                                                                                                                                    |
+| client.scanner.kv.server-side.enabled         | Boolean    | false                                           | Master switch for using the server-side KV scan (FIP-17) in bounded reads of primary-key tables when no KV snapshot file is available. When false (default), bounded primary-key reads fall back to the prior behavior (log-only when lake is enabled, or fail when lake is disabled). See [Full Scan of Primary Key Tables](engine-flink/reads.md#full-scan-of-primary-key-tables) for details. |
```
binary-signal: Enables server-side KV scanning (FIP-17) for bounded reads on primary-key tables when no KV snapshot file is available. When disabled (default), bounded reads fall back to the previous behavior: read from the log when data-lake integration is enabled, or fail when it is disabled.

polyzos:
@binary-signa Thank you for your comments... Addressed!

binary-signal:
@polyzos I had a look at the PR and can confirm that bounded reads via the Table API work as expected.
That said, there is an issue when performing bounded reads via the DataStream API. It looks like the corresponding logic was missed in FlussSourceBuilder.java. The required changes were already implemented in [flink] Add union read support to datastream (#3432) which allows `setBounded()` to be set in a Fluss Flink Source. The following  guard throws: 


```java
# FlussSourceBuilder.java

if (bounded && !(lakeEnabled && fullStartup)) {
    ...
}
```
The following patch addresses the issue and also adds a test case to cover this scenario.

```
diff --git a/fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/source/FlussSourceBuilder.java b/fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/source/FlussSourceBuilder.java
index 2e80067f..eb95e731 100644
--- a/fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/source/FlussSourceBuilder.java
+++ b/fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/source/FlussSourceBuilder.java
@@ -176,7 +176,10 @@ public class FlussSourceBuilder<OUT> {
      * Builds a bounded source for batch execution. The source reads up to the latest offsets at job
      * startup and then finishes; combined with the default {@link OffsetsInitializer#full()} on a
      * datalake-enabled table this performs a bounded union read of the lake snapshot and the Fluss
-     * log. If not called, the source is unbounded (streaming).
+     * log. Alternatively, a bounded read of a primary-key table is supported via the server-side KV
+     * scan by setting {@code client.scanner.kv.server-side.enabled = true} (see {@link
+     * ConfigOptions#CLIENT_SCANNER_KV_SERVER_SIDE_ENABLED}). If not called, the source is unbounded
+     * (streaming).
      *
      * @return this builder
      */
@@ -353,13 +356,24 @@ public class FlussSourceBuilder<OUT> {
         boolean lakeEnabled = tableInfo.getTableConfig().isDataLakeEnabled();
         boolean fullStartup = offsetsInitializer instanceof SnapshotOffsetsInitializer;
 
-        if (bounded && !(lakeEnabled && fullStartup)) {
+        // bounded read via the server-side KV scan applies to primary-key tables when the master
+        // switch is enabled, mirroring the SQL connector (see FlinkTableSource).
+        boolean kvBatchEnabled = flussConf.get(ConfigOptions.CLIENT_SCANNER_KV_SERVER_SIDE_ENABLED);
+        boolean kvBatchAllowed = hasPrimaryKey && kvBatchEnabled;
+
+        if (bounded && !(lakeEnabled && fullStartup) && !kvBatchAllowed) {
                     String.format(
-                            "Bounded (batch) read requires a datalake-enabled table started in "
-                                    + "full mode (OffsetsInitializer.full()), but table '%s' has "
-                                    + "datalake enabled=%s and full startup mode=%s.",
-                            tablePath, lakeEnabled, fullStartup));
+                            "Bounded (batch) read requires either a datalake-enabled table started "
+                                    + "in full mode (OffsetsInitializer.full()), or a primary-key "
+                                    + "table with '%s' = 'true'. Table '%s' has datalake enabled=%s, "
+                                    + "full startup mode=%s, primary key=%s, server-side KV scan=%s.",
+                            ConfigOptions.CLIENT_SCANNER_KV_SERVER_SIDE_ENABLED.key(),
+                            lakeEnabled,
+                            fullStartup,
+                            hasPrimaryKey,
+                            kvBatchEnabled));
         }
 
         LakeSource<LakeSplit> lakeSource = null;
diff --git a/fluss-flink/fluss-flink-common/src/test/java/org/apache/fluss/flink/source/FlussSourceITCase.java b/fluss-flink/fluss-flink-common/src/test/java/org/apache/fluss/flink/source/FlussSourceITCase.java
index f4b112c4..df211643 100644
--- a/fluss-flink/fluss-flink-common/src/test/java/org/apache/fluss/flink/source/FlussSourceITCase.java
+++ b/fluss-flink/fluss-flink-common/src/test/java/org/apache/fluss/flink/source/FlussSourceITCase.java
@@ -21,6 +21,8 @@ import org.apache.fluss.client.initializer.OffsetsInitializer;
 import org.apache.fluss.client.table.Table;
 import org.apache.fluss.client.table.writer.AppendWriter;
 import org.apache.fluss.client.table.writer.UpsertWriter;
+import org.apache.fluss.config.ConfigOptions;
+import org.apache.fluss.config.Configuration;
 import org.apache.fluss.flink.source.deserializer.RowDataDeserializationSchema;
 import org.apache.fluss.flink.source.testutils.MockDataUtils;
 import org.apache.fluss.flink.source.testutils.Order;
@@ -113,6 +115,37 @@ public class FlussSourceITCase extends FlinkTestBase {
         assertThat(collectedElements).hasSameElementsAs(ORDERS);
     }
 
+    @Test
+    public void testBoundedKvBatchPKSource() throws Exception {
+        createTable(ordersPKTablePath, pkTableDescriptor);
+        writeRowsToTable(ordersPKTablePath);
+
+        // Enable the server-side KV scan so a bounded read is allowed on a primary-key table
+        // without the data-lake integration (mirrors the SQL connector's behavior).
+        Configuration flussConf = new Configuration();
+        flussConf.set(ConfigOptions.CLIENT_SCANNER_KV_SERVER_SIDE_ENABLED, true);
+
+        FlussSource<Order> flussSource =
+                FlussSource.<Order>builder()
+                        .setBootstrapServers(bootstrapServers)
+                        .setDatabase(DEFAULT_DB)
+                        .setTable(pkTableName)
+                        .setStartingOffsets(OffsetsInitializer.full())
+                        .setScanPartitionDiscoveryIntervalMs(1000L)
+                        .setDeserializationSchema(new MockDataUtils.OrderDeserializationSchema())
+                        .setFlussConfig(flussConf)
+                        .setBounded()
+
+        // env.setRuntimeMode(RuntimeExecutionMode.BATCH);
+        DataStreamSource<Order> stream =
+                env.fromSource(flussSource, WatermarkStrategy.noWatermarks(), "Fluss Source");
+
+        List<Order> collectedElements = stream.executeAndCollect(ORDERS.size());
+
+        assertThat(collectedElements).hasSameElementsAs(ORDERS);
+    }
+
     @Test
     public void testTablePKSourceWithProjectionPushdown() throws Exception {
         createTable(ordersPKTablePath, pkTableDescriptor);
```





Quick example to reproduce

create a primary key table with datalake option disabled

```sql
 CREATE DATABASE cdstream_db;

 CREATE TABLE `fluss_catalog`.`cdstream_db`.`timeseries_norm` (
  `streamId` INT NOT NULL,
  `vector` ARRAY<DOUBLE>,
  `firstTimestamp` TIMESTAMP(3),
  `lastTimestamp` TIMESTAMP(3),
  `lastUpdateTime` TIMESTAMP(3),
  CONSTRAINT `PK_streamId` PRIMARY KEY (`streamId`) NOT ENFORCED
)
COMMENT 'This is the normalized timeseries table'
WITH (
  'bootstrap.servers' = 'coordinator-server:9123',
  'bucket.key' = 'streamId',
  'bucket.num' = '2',
  'table.datalake.enabled' = 'false',
  'table.datalake.format' = 'paimon'
)
```

Then try bounded read via DataStream API it will throw
```java
import org.apache.flink.api.common.RuntimeExecutionMode;
import org.apache.flink.api.common.eventtime.WatermarkStrategy;
import org.apache.flink.streaming.api.datastream.DataStream;
import org.apache.flink.streaming.api.environment.StreamExecutionEnvironment;
import org.apache.flink.table.data.RowData;
import org.apache.fluss.config.Configuration;
import org.apache.fluss.flink.source.FlussSource;
import org.apache.fluss.flink.source.deserializer.RowDataDeserializationSchema;

public class FlussDataStreamBoundedReadKvBatch {
    public static void main(String[] args) throws Exception {

        StreamExecutionEnvironment env = StreamExecutionEnvironment.getExecutionEnvironment();


        // env.setRuntimeMode(RuntimeExecutionMode.STREAMING);
        Configuration conf = new Configuration();
        conf.setBoolean("client.scanner.kv.server-side.enabled", true);
        var source = FlussSource.<RowData>builder()
                .setBootstrapServers("localhost:9122")
                .setDatabase("cdstream_db")
                .setTable("timeseries_norm")
                .setDeserializationSchema(new RowDataDeserializationSchema())
                .setBounded()
                .setFlussConfig(conf)
                .build();

        DataStream<RowData> stream = env.fromSource(source, WatermarkStrategy.noWatermarks(), "BoundedFlussKvSource");

        stream.print();
        env.execute("BoundedFlussKvJob");
    }
}
```

polyzos:
@binary-signal yes I need to find some time to rebase, and after we also need to introduce partition pruning and predicate pushdown tot Datastream.

I have tickets to track these

fluss-flink/fluss-flink-common/src/test/java/org/apache/fluss/flink/source/FlinkTableSourceBatchITCase.java
```diff
@@ -250,8 +253,151 @@ void testScanSingleRowFilterException() throws Exception {
...
+                                + " (set '"
+                                + ConfigOptions.TABLE_DATALAKE_ENABLED.key()
+                                + "' = 'true') or server-side KV scan on a"
+                                + " primary-key table (set '"
+                                + ConfigOptions.CLIENT_SCANNER_KV_SERVER_SIDE_ENABLED.key()
+                                + "' = 'true').");
+    }
+
+    @Test
+    void testKvBatchScanOnPkTable() throws Exception {
```
fresh-borzoni: All KV tests write each key exactly once, so a log read would return the same rows and still pass. Shall we add an update + a delete before the scan, so the test can tell KV state from the changelog?

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/source/enumerator/FlinkSourceEnumerator.java
```diff
@@ -583,41 +585,65 @@ private void startInBatchMode() {
...
-                                // Use log-only splits to avoid generating mixed split
-                                // types (HybridSnapshotLogSplit + LogSplit) for
-                                // primary-key tables, which is not supported.
-                                splits =
-                                        this.initLogTablePartitionSplits(
-                                                partitions, startingOffsetsInitializer);
-                            } else {
-                                splits = this.getLogSplit(null, null);
-                            }
+                            splits = generateFlussOnlyBatchSplits(kvBatchEnabled);
```
fresh-borzoni: With lake enabled but no lake snapshot yet, flag off reads the whole log, so updates and deletes come back as extra rows (-U/+U/-D), and SELECT * returns the history, not the table.
Shall we KV-scan PK tables here regardless of the flag or fail? Silently wrong rows as the default seems wrong. WDYT?

fluss-common/src/main/java/org/apache/fluss/config/ConfigOptions.java
```diff
@@ -1418,6 +1418,18 @@ public class ConfigOptions {
...
                                     + "'.");
 
+    public static final ConfigOption<Boolean> CLIENT_SCANNER_KV_SERVER_SIDE_ENABLED =
+            key("client.scanner.kv.server-side.enabled")
+                    .booleanType()
+                    .defaultValue(false)
+                    .withDescription(
+                            "Master switch for using the server-side KV scan in bounded reads "
+                                    + "of primary-key tables. When true, bounded primary-key reads "
+                                    + "scan the current KV state directly via the server-side KV scan. "
```
fresh-borzoni: "scan the current KV state directly" reads as if the flag forces a KV scan, but when a lake snapshot exists the union read still wins and the flag is never consulted and it only really applies when there's no lake data yet. 

Shall we add "has no effect when a lake snapshot already exists" or smth similar? Same- in options.md.

fresh-borzoni (commented):
@polyzos Thank you for the PR, left a couple of comments, PTAL

polyzos:
@fresh-borzoni Thank you for your review and comments.
They all make sense, and I pushed some changes to address them. Let me know your thoughts.

fresh-borzoni (commented):
Thank you @polyzos, looks good overall 👍 

This one's actually still open though: FlinkSourceEnumerator didn't change, so flag-off on a lake-enabled PK table before its first tiering snapshot still log-reads, which returns the changelog, not the table. It only affects the create -> first-tier window (~3 min default, longer if freshness is tuned up), but it's a silent wrong answer there. 
The log can't produce current state for a PK table, so no offset fix helps - could flag-off just fail with a pointer to the option (empty tables still return 0 rows, so #3208's case is kept), and leave the log path for log tables only?

@luoyuxia Can you also advise here?

fresh-borzoni (approved):
@polyzos Thank you for the changes, LGTM 👍 

Now that we have #3677, I made a couple of additions and revised the docs to position each feature properly. Are the changes okay with you?

polyzos:
@fresh-borzoni LGTM, thanks 🙏

PR #3388, opened 2026-05-27
[server]: remove orphan files and directories when tablet starts up

### LINKED ISSUE #3387: [server] TabletServer leaves orphan replica directories after restart during table/partition deletion
### Search before asking

### Fluss version

0.9.0 (latest release)

## Summary

When a TabletServer restarts while one of its replicas is being deleted (the TS was offline when `dropTable` or `dropPartition` was issued), the on-disk replica directories on the restarted TabletServer are never removed. This produces a permanent disk leak — the directories have no in-memory owner and are never reaped.

## Reproduction

1. Start a Fluss cluster with 3 TabletServers.
2. Create a table with `replication-factor = 3`.
3. Stop TabletServer-1.
4. Drop the table (or partition).
5. Restart TabletServer-1.
6. Observe that the data directory on TabletServer-1 (`<dataDir>/<db>/<table>-<tableId>/log-0/`) still exists.

## Root cause

Two independent gaps in the TabletServer's cleanup logic produce the leak, depending on whether the deleted entity is a table or a partition.

### Gap 1: `stopReplica` for `NoneReplica` is a no-op (affects partition deletion)

When a TabletServer restarts, `ReplicaManager.allReplicas` is empty — buckets are only registered when `NotifyLeaderAndIsr` arrives. However, `LogManager.startup()` calls `loadAllLogs()`, which walks the data directories and loads every on-disk log tablet into `LogManager.currentLogs`.

For **partition deletion**, the table metadata still exists in ZK, so `loadAllLogs` succeeds and the orphan log tablet enters `currentLogs`. When the coordinator re-sends `stopReplica(delete=true)` after the TS reconnects, the bucket resolves to `NoneReplica` (not in `allReplicas`), and the current code does nothing:

```java
if (hostedReplica instanceof NoneReplica) {
    // do nothing for this case.
    result.add(new StopReplicaResultForBucket(tb));
}
```

The TS responds "success" without touching the on-disk directory. The log tablet sits in `currentLogs` with no way to be cleaned up.

### Gap 2: `SchemaNotExistException` handler leaves empty parent directories (affects table deletion)

For **table deletion**, `dropTable` removes table metadata (including schemas) from ZK synchronously, before sending `stopReplica`. When the offline TS restarts:

1. `loadAllLogs` → `loadLog` → `getTableInfo` → `SchemaNotExistException` (schema gone from ZK)
2. The existing handler deletes `log-N/` and `kv-N/` tablet directories
3. But **the parent directory** (`<db>/<table>-<tableId>/` or `<db>/<table>-<tableId>/<partition>-p<partitionId>/`) **is left behind empty**

Since the log tablet was already cleaned up at startup, `currentLogs` does not contain the bucket. If a `stopReplica(delete=true)` later arrives, the `NoneReplica` branch finds nothing to sweep — the empty parent directory leaks.

### Solution

_No response_

gyang94:
### Purpose

Linked issue: close #3387

### Brief change log

Two changes, one in each cleanup path:

**1. `ReplicaManager`: handle `NoneReplica` with `deleteLocal=true`**

When the `NoneReplica` branch receives `deleteLocal=true`, look up the bucket in `LogManager.currentLogs`. If present, the log tablet was loaded at startup but never registered in `allReplicas` — it is an orphan. Drop the log via `logManager.dropLog()`, delete the sibling KV tablet directory (via `kvManager.dropKv()` if loaded, otherwise direct `FileUtils.deleteDirectory`), update disk usage accounting, and record the parent directory for empty-dir cleanup.

This handles the **partition deletion** scenario.

**2. `LogManager`: clean up empty parent directories in `SchemaNotExistException` handler**

After the existing handler deletes `log-N/` and `kv-N/`, check whether the parent directory is empty and delete it. For partitioned tables, also check the grandparent (table directory). For non-partitioned tables, the grandparent is the database directory and must NOT be deleted. This is safe under parallel `loadAllLogs` execution — `deleteDirectoryQuietly` tolerates races.

This handles the **table deletion** scenario.

### DIFF
--- a/fluss-server/src/main/java/org/apache/fluss/server/log/LogManager.java
+++ b/fluss-server/src/main/java/org/apache/fluss/server/log/LogManager.java
@@ -23,6 +23,7 @@ import org.apache.fluss.config.Configuration;
 import org.apache.fluss.exception.FlussException;
 import org.apache.fluss.exception.FlussRuntimeException;
 import org.apache.fluss.exception.LogStorageException;
+import org.apache.fluss.exception.PartitionNotExistException;
 import org.apache.fluss.exception.SchemaNotExistException;
 import org.apache.fluss.metadata.LogFormat;
 import org.apache.fluss.metadata.PhysicalTablePath;
@@ -34,6 +35,7 @@ import org.apache.fluss.server.log.checkpoint.OffsetCheckpointFile;
 import org.apache.fluss.server.metrics.group.TabletServerMetricGroup;
 import org.apache.fluss.server.storage.LocalDiskManager;
 import org.apache.fluss.server.zk.ZooKeeperClient;
+import org.apache.fluss.server.zk.data.PartitionRegistration;
 import org.apache.fluss.utils.ExceptionUtils;
 import org.apache.fluss.utils.FileUtils;
 import org.apache.fluss.utils.FlussPaths;
@@ -48,7 +50,9 @@ import javax.annotation.concurrent.ThreadSafe;
 
 import java.io.File;
 import java.io.IOException;
+import java.nio.file.DirectoryNotEmptyException;
 import java.nio.file.Files;
+import java.nio.file.NoSuchFileException;
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.LinkedHashMap;
@@ -386,6 +390,25 @@ public final class LogManager extends TabletManagerBase {
         PhysicalTablePath physicalTablePath = pathAndBucket.f0;
         TablePath tablePath = physicalTablePath.getTablePath();
         TableInfo tableInfo = getTableInfo(zkClient, tablePath);
+
+        // Table schema exists, but the partition may have been dropped (or dropped
+        // and recreated with a new ID) while this TS was offline.
+        // Validate both partition name and partition ID against ZK.
+        String partitionName = physicalTablePath.getPartitionName();
+        if (partitionName != null) {
+            Optional<PartitionRegistration> registration =
+                    zkClient.getPartition(tablePath, partitionName);
+            if (!registration.isPresent()
+                    || registration.get().getPartitionId() != tableBucket.getPartitionId()) {
+                throw new PartitionNotExistException(
+                        String.format(
+                                "Failed to load partition '%s' (partitionId=%d) of table '%s': "
+                                        + "partition not found or partitionId mismatch in "
+                                        + "zookeeper metadata.",
+                                partitionName, tableBucket.getPartitionId(), tablePath));
+            }
+        }
+
         LogTablet logTablet =
                 LogTablet.create(
                         dataDir,
@@ -535,17 +558,19 @@ public final class LogManager extends TabletManagerBase {
                     loadLog(dataDir, tabletDir, cleanShutdown, recoveryPoints, conf, clock);
                 } catch (Exception e) {
                     LOG.error("Fail to loadLog from {}", tabletDir, e);
-                    if (e instanceof SchemaNotExistException) {
+                    if (e instanceof SchemaNotExistException
+                            || e instanceof PartitionNotExistException) {
                         LOG.error(
-                                "schema not exist, table for {} has already been dropped, the residual data will be removed.",
+                                "Table or partition for {} has already been dropped, the residual data will be removed.",
                                 tabletDir,
                                 e);
                         FileUtils.deleteDirectoryQuietly(tabletDir);
 
-                        // Also delete corresponding KV tablet directory if it exists
                         try {
                             Tuple2<PhysicalTablePath, TableBucket> pathAndBucket =
                                     FlussPaths.parseTabletDir(tabletDir);
+
+                            // Also delete corresponding KV tablet directory if it exists
                             File kvTabletDir =
                                     FlussPaths.kvTabletDir(
                                             dataDir, pathAndBucket.f0, pathAndBucket.f1);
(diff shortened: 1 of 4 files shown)


fluss-server/src/main/java/org/apache/fluss/server/log/LogManager.java
```diff
@@ -561,6 +561,34 @@ public void run() {
...
+
+                        // delete empty parent directories. For partitioned tables the
+                        // parent is the partition dir and the grandparent is
+                        // the table dir — both should be removed when empty.
+                        // For non-partitioned tables the parent is the table
+                        // dir (safe to remove) but the grandparent is the
+                        // database dir — must NOT remove. Safe under parallel
+                        // execution: the last job to finish finds the dir
+                        // empty and removes it; deleteDirectoryQuietly
+                        // tolerates races.
```
swuferhong: spotless the docs.

fluss-server/src/main/java/org/apache/fluss/server/log/LogManager.java
```diff
@@ -561,6 +561,34 @@ public void run() {
...
+                        try {
+                            boolean isPartitioned = pathAndBucket.f0.getPartitionName() != null;
+                            File parentDir = tabletDir.getParentFile();
+                            if (parentDir != null && FileUtils.isDirectoryEmpty(parentDir)) {
+                                FileUtils.deleteDirectoryQuietly(parentDir);
```
swuferhong: There's a TOCTOU race: between the `isDirectoryEmpty` check and `deleteDirectoryQuietly`, another thread/loadLog task could create a file. Consider using `Files.delete(path)` which fails on non-empty dirs, making this truly safe under concurrency.

fluss-server/src/main/java/org/apache/fluss/server/log/LogManager.java
```diff
@@ -561,6 +561,34 @@ public void run() {
...
+                                if (isPartitioned) {
+                                    File tableDir = parentDir.getParentFile();
+                                    if (tableDir != null && FileUtils.isDirectoryEmpty(tableDir)) {
```
swuferhong: ditto

fluss-server/src/main/java/org/apache/fluss/server/log/LogManager.java
swuferhong: For the table drop case, the startup cleanup via `SchemaNotExistException` works well.

However, for the partition drop case (table still exists, only partition dropped while TS was offline), the orphan files won't be auto-cleaned. When the TS restarts, loadLog() succeeds because the table schema still exists in ZK, so the log is loaded normally as a NoneReplica. The sweepOrphanTabletDirs mechanism requires a stopReplica(delete=true) to trigger, but the Coordinator won't re-send it — by the time the TS reconnects, the Coordinator has already marked the deletion as "successful" (after retry exhaustion) and removed all tracking metadata for that partition's replicas.

I think you need consider adding a startupcheck for partition existence (similar to the schema check) to cover this gap.

swuferhong (commented):
Hi @gyang94. Thanks for your great work. I left some comments, pls take a look.

The commit msg has a typo: "remote orphan files" — should be "remove orphan files".

fluss-server/src/main/java/org/apache/fluss/server/log/LogManager.java
```diff
@@ -555,11 +572,23 @@ public void run() {
                                         kvTabletDir);
                                 FileUtils.deleteDirectoryQuietly(kvTabletDir);
                             }
-                        } catch (Exception kvDeleteException) {
+
+                            boolean isPartitioned = pathAndBucket.f0.getPartitionName() != null;
+                            File parentDir = tabletDir.getParentFile();
```
swuferhong: recommend named to `partitionDir`.

fluss-server/src/main/java/org/apache/fluss/server/replica/ReplicaManager.java
```diff
@@ -978,7 +980,20 @@ public void stopReplicas(
                         TableBucket tb = data.getTableBucket();
                         HostedReplica hostedReplica = getReplica(tb);
                         if (hostedReplica instanceof NoneReplica) {
-                            // do nothing fort this case.
+                            if (data.isDeleteLocal()) {
+                                try {
+                                    sweepOrphanTabletDirs(tb, deletedTableIds, deletedPartitionIds);
```
swuferhong, line 985: No test to cover this method.  Both new tests drop before the TS restart, so ZK metadata is already absent at startup — this only exercises path A (SchemaNotExistException handler). The newly added ReplicaManager.sweepOrphanTabletDirs (path B) has no test coverage.

Consider adding a case that hits the NoneReplica + deleteLocal branch: restart the TS first (metadata still present → log loaded but no replica), then drop the table/partition.

fluss-server/src/main/java/org/apache/fluss/server/log/LogManager.java
```diff
@@ -386,6 +389,18 @@ private LogTablet loadLog(
         PhysicalTablePath physicalTablePath = pathAndBucket.f0;
         TablePath tablePath = physicalTablePath.getTablePath();
         TableInfo tableInfo = getTableInfo(zkClient, tablePath);
+
+        // Table schema exists, but the partition may have been dropped while
+        // this TS was offline. Check partition registration in ZK.
+        String partitionName = physicalTablePath.getPartitionName();
+        if (partitionName != null && !zkClient.getPartition(tablePath, partitionName).isPresent()) {
```
swuferhong: Validates only partitionName against ZK, not partitionId. 

If a partition is dropped and recreated under the same name (new partitionId), the stale on-disk directory for the old partitionId passes the check and gets loaded as valid data, while its TableBucket still carries the old partitionId. Validating partitionId (e.g. comparing the ZK PartitionRegistration.getPartitionId() with the one parsed from the dir) would be more robust.

swuferhong (commented):
Thanks @gyang94 . LGTM allover, I Left some comments.

swuferhong (approved):
LGTM

PR #3391, opened 2026-05-27
[server] Add per-tablet-server sender retry framework

### LINKED ISSUE #3357: [server] Table deletion stuck permanently when StopReplica request fails

### LINKED ISSUE #4262: [server] Add per-TabletServer sender thread infrastructure
### Search before asking

- [x] I searched existing issues for a dedicated per-TabletServer sender infrastructure task. The related StopReplica bug is tracked separately in #3357.

### Description

Coordinator-to-TabletServer control-plane RPCs currently use direct asynchronous gateway calls without a shared per-TabletServer queue and retry mechanism. Introduce reusable sender infrastructure on the CoordinatorServer so future request integrations can use ordered dispatch, transport-failure retries, and consistent lifecycle management.

Each registered TabletServer should have its own sender thread and in-memory FIFO queue. When an enqueued request encounters a retriable transport failure, that sender retains the current item and retries with backoff; sender threads for other TabletServers can continue independently.

#### Scope

- Provide generic queue items carrying the API key, request-sending function, response callback, coordinator epoch, and enqueue timestamp, without request-specific dispatch logic.
- Manage per-TabletServer queues and sender threads through channel startup, TabletServer registration/removal, and Coordinator shutdown. Make shared gateway access safe across the event and sender threads.
- Retry recognized network/disconnection failures and timeouts with configurable backoff. Complete non-retriable failures through the callback, and propagate fatal errors instead of retrying them. A callback failure must not resend an already completed request.
- Invalidate the old RPC connection before resending after a transport failure, so abandoned in-flight requests do not accumulate. Bound each wait for invalidation completion, retain the current item while invalidation is pending or fails, and resume request sending only after invalidation succeeds.
- Drop queued items with an older coordinator epoch when dequeued, and expose per-TabletServer queue size, queue time, retry count, stale-drop count, and sender liveness metrics.
- Add and document `coordinator.control-request.retry-backoff` (default: `100ms`) and `coordinator.control-request.timeout` (default: `30s`), with a minimum value of `1ms` for both.

#### Scope boundary

This task covers the sender infrastructure only. Existing concrete RPCs, including `StopReplica`, retain their current dispatch and response handling. Migrating requests to the sender framework, changing replica-deletion states, handling deletion recovery after TabletServer reconnects, and resolving #3357 belong to follow-up work.

The queues are in-memory and are cleared when their channels are removed. The framework does not provide durable request delivery or request-specific reconciliation; these semantics must be addressed by each future request integration.

#### Validation

Cover FIFO dispatch and retries, callback/non-retriable failure handling, stale-item drops, sender lifecycle, configuration validation, and connection invalidation failures/timeouts. Include a real RPC regression test verifying that disconnect completes an outstanding client request and a subsequent request can reconnect using the same gateway.

### Related work

- Implementation: #3391 — per-TabletServer sender retry framework.
- Related bug: #3357 — table deletion stuck after a StopReplica failure; resolving that bug is outside this infrastructure task.

### Willingness to contribute

- [x] I'm willing to submit a PR. Implementation is available in #3391.

gyang94:
### Purpose

Closes #4262.

Coordinator-to-TabletServer control-plane RPCs lack a shared per-TabletServer queue and retry mechanism. This PR adds reusable sender infrastructure on the CoordinatorServer: each registered TabletServer has a dedicated sender thread and an in-memory FIFO queue, allowing an enqueued request to retry after transport failures while other TabletServers' sender threads continue independently.

This PR provides the infrastructure for future request integrations. Existing concrete RPCs retain their current dispatch and response handling; migrating them to the sender framework belongs to follow-up PRs.

### Brief change log

- Add generic `QueueItem`, `TabletServerChannelState`, and `ControlRequestSendThread` components, with request-sending functions, typed response callbacks, coordinator epochs, and enqueue timestamps.
- Manage sender startup and shutdown, queue cleanup, and metric registration through the channel lifecycle. Make shared RPC gateway access thread-safe.
- Retry recognized network/disconnection failures and timeouts with backoff while retaining the current item. Complete non-retriable failures through the callback, propagate fatal errors, and prevent callback failures from resending completed requests.
- Invalidate the old connection before resending so abandoned in-flight RPCs are cleaned up. Bound each wait for invalidation completion and retain the current item while invalidation is pending or fails; resend only after invalidation succeeds.
- Drop items from older coordinator epochs when dequeued. Expose per-TabletServer queue size, queue time, retry count, stale-drop count, and sender liveness metrics using the `tablet_server_id` label.
- Add `coordinator.control-request.retry-backoff` (default: `100ms`) and `coordinator.control-request.timeout` (default: `30s`), validating both to be at least `1ms`.

Queues are in-memory and are cleared on channel removal. Request-specific recovery and reconciliation remain the responsibility of future integrations.

### Tests

The following targeted tests passed: `FlussConfigUtilsTest` (7), `ControlRequestSendThreadTest` (12), `RpcGatewayManagerTest` (2), and `CoordinatorChannelManagerTest` (1), for 22 tests total.

Coverage includes FIFO retries, terminal and fatal failures, callback handling, stale-item drops, shutdown/liveness, duration validation, invalidation failures/timeouts, and a real Netty disconnect/reconnect regression test.

Validation with Java 11:

```sh
mvn -o -pl fluss-server -am -Dtest=FlussConfigUtilsTest,ControlRequestSendThreadTest,RpcGatewayManagerTest,CoordinatorChannelManagerTest -Dsurefire.failIfNoSpecifiedTests=false test
mvn -o -pl fluss-server -am -DskipTests package
```

Both commands passed, including Checkstyle, Spotless, and RAT checks. These are targeted tests and a module build, not a full repository test run.

### API and Format

Adds internal sender infrastructure and two Coordinator configuration options. No public client API, wire-protocol, or storage-format changes.

### Documentation

Document the new options in `website/docs/maintenance/configuration.md`.

### Generative AI disclosure

AI-assisted changes using OpenAI Codex.


fluss-server/src/main/java/org/apache/fluss/server/coordinator/channel/QueueItem.java
```diff
@@ -0,0 +1,87 @@
...
+    public int getCoordinatorEpoch() {
+        return coordinatorEpoch;
+    }
+
+    public long getEnqueueTimeMs() {
+        return enqueueTimeMs;
+    }
+
+    @Nullable
+    public Set<TableBucketReplica> getDeletionReplicas() {
```
swuferhong: This method and this field both are not used. I think it need to be removed

fluss-server/src/main/java/org/apache/fluss/server/coordinator/channel/ApiKey.java
```diff
@@ -0,0 +1,30 @@
...
+package org.apache.fluss.server.coordinator.channel;
+
+/**
+ * Enumeration of control-plane RPCs routed through the per-tablet-server sender thread.
+ *
+ * <p>Only {@link #STOP_REPLICA} is supported in this iteration. Future iterations may add other
+ * coordinator → tablet-server RPCs (e.g., {@code NOTIFY_LEADER_AND_ISR}) once they are migrated to
+ * the same sender-thread retry layer.
+ */
+public enum ApiKey {
```
swuferhong: I don't think we need to introduce this new enum. We already have `org.apache.fluss.rpc.protocol.ApiKeys` in fluss-rpc, which is the canonical enum for all wire-protocol APIs and already includes `STOP_REPLICA` — along with all the other control-plane RPCs (`NOTIFY_LEADER_AND_ISR`, `UPDATE_METADATA`, `NOTIFY_REMOTE_LOG_OFFSETS`, etc.).

The new ApiKey only has a single member today, and as we migrate more control-plane RPCs onto the sender thread, its members would just duplicate the ones already in `ApiKeys`. That gives us two parallel "ApiKey" concepts to keep in sync, which seems unnecessary.

There's also no dependency concern: fluss-server already depends on fluss-rpc and uses `ApiKeys `elsewhere.

fluss-common/src/main/java/org/apache/fluss/config/ConfigOptions.java
```diff
@@ -859,6 +859,25 @@ public class ConfigOptions {
                     .withDescription("The amount of time to sleep when fetch bucket error occurs.")
                     .withFallbackKeys("log.replica.fetch-backoff-interval");
 
+    public static final ConfigOption<Duration> COORDINATOR_REQUEST_RETRY_BACKOFF =
```
swuferhong: Why are the two `coordinator-related` config options placed together with the `log-related` ones? Could we move them up to the server module section instead? Also, please add documentation for these options in `configuration.md`.

fluss-common/src/main/java/org/apache/fluss/config/ConfigOptions.java
```diff
@@ -859,6 +859,25 @@ public class ConfigOptions {
...
+            key("coordinator.request-retry.backoff-interval")
```
swuferhong: The two new options use inconsistent key structures:
`coordinator.request-retry.backoff-interval` — uses a request-retry segment
`coordinator.request.timeout` — uses a request segment

These are both about the same thing (control-plane requests from the coordinator), so they should share a common coordinator.request. prefix. 

Right now request-retry introduces a separate sub-namespace that doesn't line up with request.timeout.Suggested rename to keep them under one prefix:
`coordinator.request-retry.backoff-interval`→ `coordinator.request.retry-backoff`
`coordinator.request.timeout` → keep as is

fluss-common/src/main/java/org/apache/fluss/config/ConfigOptions.java
```diff
@@ -859,6 +859,25 @@ public class ConfigOptions {
...
+                    .durationType()
+                    .defaultValue(Duration.ofMillis(100))
+                    .withDescription(
+                            "The backoff duration the coordinator waits before retrying a "
+                                    + "control-plane request to a tablet server after a "
+                                    + "transient RPC-layer failure. Mirrors Kafka's "
```
swuferhong: `Mirrors Kafka's ControllerChannelManager retry backoff (hardcoded 100ms)` -> Suggest removing this part — there's no need to mention it here.

fluss-common/src/main/java/org/apache/fluss/metrics/MetricNames.java
```diff
@@ -45,6 +45,14 @@ public class MetricNames {
     public static final String PARTITION_COUNT = "partitionCount";
     public static final String REPLICAS_TO_DELETE_COUNT = "replicasToDeleteCount";
 
+    // for coordinator sender (per-tablet-server control request sender threads)
+    public static final String SENDER_TOTAL_QUEUE_SIZE = "senderTotalQueueSize";
```
swuferhong: `senderTotalQueueSize` -> this one can be dropped?  It's derivable by aggregating senderQueueSize across tablet servers, so it's not strictly necessary. If you'd like to keep a single "global at-a-glance" metric, leaving it in is fine too — it's a judgment call. But if we're optimizing for a leaner metric set, I'd prefer to drop this one rather than the per-TS gauge, since the per-TS breakdown carries more diagnostic value (it tells you which tablet server is backing up).

fluss-server/src/main/java/org/apache/fluss/server/coordinator/channel/ControlRequestSendThread.java
```diff
@@ -0,0 +1,204 @@
...
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.concurrent.atomic.AtomicReference;
+import java.util.function.IntSupplier;
+import java.util.function.Supplier;
+
+/**
+ * Per-tablet-server sender thread that drains the control-plane request queue and retries on
+ * transient RPC failures. Mirrors Kafka's {@code RequestSendThread}
+ * (ControllerChannelManager.scala:226-300).
```
swuferhong: There's no need to emphasize this here. If you want to indicate that you referenced an Apache open source project, please use a more standard comment format — refer to `LogManager` for an example:

```
/* This file is based on source code of Apache Kafka Project (https://kafka.apache.org/), licensed by the Apache
 * Software Foundation (ASF) under the Apache License, Version 2.0. See the NOTICE file distributed with this work for
 * additional information regarding copyright ownership. */
```

fluss-server/src/main/java/org/apache/fluss/server/coordinator/channel/QueueItem.java
```diff
@@ -0,0 +1,87 @@
...
+
+import javax.annotation.Nullable;
+
+import java.util.Set;
+import java.util.function.BiConsumer;
+
+/**
+ * An immutable item placed on the per-tablet-server sender queue.
+ *
+ * <p>Mirrors Kafka's {@code QueueItem} (ControllerChannelManager.scala:218-219) with two
```
swuferhong: ditto.  If you want to indicate that you referenced an Apache open source project, please use a more standard comment format

fluss-server/src/main/java/org/apache/fluss/server/coordinator/channel/TabletServerChannelState.java
```diff
@@ -0,0 +1,93 @@
...
+import javax.annotation.concurrent.ThreadSafe;
+
+import java.util.concurrent.BlockingQueue;
+
+/**
+ * Per-tablet-server channel state held by {@link
+ * org.apache.fluss.server.coordinator.CoordinatorChannelManager}: the queue, the sender thread, and
+ * the metric handles needed to deregister the per-TS metrics on remove.
+ *
+ * <p>Mirrors Kafka's {@code ControllerBrokerStateInfo} (ControllerChannelManager.scala:768-774).
```
swuferhong: ditto

fluss-server/src/main/java/org/apache/fluss/server/coordinator/channel/TabletServerChannelState.java
```diff
@@ -0,0 +1,93 @@
...
+ * The {@code networkClient} field is intentionally absent — {@code RpcGatewayManager} owns the
+ * gateway exactly the way Kafka's controller owns the {@code NetworkClient}, so we do not duplicate
+ * it on the state. Three Fluss-specific additions surface behavior that has no Kafka analog:
+ *
+ * <ul>
+ *   <li>{@code retryCount} — counts §5.4 retry attempts.
```
swuferhong: What's this tag means: `§5.4` ? Pls check all the code refer to these unknown tag.

fluss-server/src/main/java/org/apache/fluss/server/coordinator/channel/ControlRequestSendThread.java
```diff
@@ -0,0 +1,204 @@
...
+ * Otherwise the item is sent to the tablet server via the gateway, retrying with a configurable
+ * backoff until the send succeeds or the thread is shut down.
+ *
+ * <p>The callback is invoked in its own {@code try/catch} OUTSIDE the retry loop (alignment fix F),
+ * so a buggy response handler cannot re-enter the send retry path and cannot reach the outer
+ * Scenario-E catch.
+ */
+public class ControlRequestSendThread extends ShutdownableThread {
+
+    private static final int HISTOGRAM_WINDOW_SIZE = 100;
```
swuferhong, line 64: This field is never used.

fluss-server/src/main/java/org/apache/fluss/server/coordinator/CoordinatorEventProcessor.java
```diff
@@ -919,43 +959,93 @@ private void processDropPartition(DropPartitionEvent dropPartitionEvent) {
                 dropTableInfo.getTablePath(), tableId, tablePartition.getPartitionId());
     }
 
+    /**
+     * Handles StopReplica deletion responses from the sender thread. Partitions per-bucket results
+     * into succeeded/failed sets, then moves replicas to the appropriate state.
+     *
+     * <p>Mirrors Kafka's {@code processTopicDeletionStopReplicaResponseReceived}
```
swuferhong: Please don't add so many reference comments in the code. Just add the standard Apache attribution comment where needed. Could you check all of them?

fluss-server/src/main/java/org/apache/fluss/server/coordinator/CoordinatorEventProcessor.java
```diff
@@ -919,43 +959,93 @@ private void processDropPartition(DropPartitionEvent dropPartitionEvent) {
...
         }
-        // clear the fail deleted number for the success deleted replicas
-        coordinatorContext.clearFailDeleteNumbers(successDeletedReplicas);
 
-        // pick up the replicas to retry delete and replicas that considered as success delete
-        Tuple2<Set<TableBucketReplica>, Set<TableBucketReplica>>
-                retryDeleteAndSuccessDeleteReplicas =
-                        coordinatorContext.retryDeleteAndSuccessDeleteReplicas(failDeletedReplicas);
+        if (!failed.isEmpty()) {
+            failReplicaDeletion(failed);
```
swuferhong: `failReplicaDeletion` marks the owning table/partition ineligible unconditionally, without checking whether the owning `TabletServer` is still alive. But the only way to reach `failReplicaDeletion` is a response whose bucket result is an error — i.e. the TS actually replied and is alive. (Pure network failures are retried indefinitely by the per-TS sender thread and never surface here.

Meanwhile, the only code path that clears the table/partition ineligible flag is `processNewTabletServer`, which fires on a TS reconnect. A live TS never reconnects, so the flag is never cleared. That produces a permanent hang:

```
stopReplica returns error (TS alive)
   → failReplicaDeletion: replica → Ineligible, table marked ineligible
   → resumeTableDeletions Step 2 (no Started + has Ineligible): Ineligible → Offline
   → resumeTableDeletions Step 3 isEligibleForDeletion = queued && noStarted && !ineligible
                               → table still in ineligible set → false → not deleted
   → replica parks at OfflineReplica; table stays queued + ineligible forever
   → next round: no Ineligible left (Step 2 won't fire), table flag still set (Step 3 still blocks)
   → TS stays alive, never reconnects → flag never cleared → permanent hang
```
This isn't an extreme corner case — it's the ordinary "TS alive but deletion failed" failure path.

Suggested fix:
- In failReplicaDeletion, do not mark the table ineligible for a still-alive TS — just retry with backoff; only mark ineligible when the TS is confirmed offline.

fluss-server/src/main/java/org/apache/fluss/server/coordinator/CoordinatorEventProcessor.java
```diff
@@ -919,43 +959,93 @@ private void processDropPartition(DropPartitionEvent dropPartitionEvent) {
...
+        if (!succeeded.isEmpty()) {
+            completeReplicaDeletion(succeeded);
+        }
+    }
 
-        // transmit to deletion started for retry delete replicas
-        replicaStateMachine.handleStateChanges(
-                retryDeleteAndSuccessDeleteReplicas.f0, ReplicaDeletionStarted);
+    /**
+     * Mirrors Kafka's {@code TopicDeletionManager.failReplicaDeletion}
```
swuferhong: ditto

fluss-server/src/main/java/org/apache/fluss/server/coordinator/CoordinatorEventProcessor.java
```diff
@@ -1116,6 +1206,32 @@ private void processNewTabletServer(NewTabletServerEvent newTabletServerEvent) {
...
+                        .collect(Collectors.toSet());
+        if (!tablesToResume.isEmpty() || !partitionsToResume.isEmpty()) {
+            tablesToResume.forEach(coordinatorContext::removeTableFromIneligibleForDeletion);
+            partitionsToResume.forEach(
+                    coordinatorContext::removePartitionFromIneligibleForDeletion);
+            tableManager.resumeDeletions();
+        }
     }
 
     private void processDeadTabletServer(DeadTabletServerEvent deadTabletServerEvent) {
```
swuferhong, line 1351: Even on the legitimate "TS went offline" path, if that TS's machine is decommissioned and never rejoins under the same id, the table stays ineligible and its deletion hangs forever — there's no timeout, alert, or manual force-clear entry point. Kafka has the same limitation (it also relies on a broker restart), so this is a known constraint rather than a new regression.

- at minimum,  maybe we can add an observability hook so operators can detect a stuck deletion — e.g. a metric for "time a table/partition has been ineligible for deletion," or a periodic WARN log listing tables/partitions that have been ineligible beyond some threshold.

fluss-server/src/main/java/org/apache/fluss/server/coordinator/CoordinatorChannelManager.java
```diff
@@ -191,6 +345,13 @@ protected Optional<TabletServerGateway> getTabletServerGateway(int targetServerI
         return rpcGatewayManager.getRpcGateway(targetServerId);
     }
 
+    @VisibleForTesting
+    Map<Integer, TabletServerChannelState> getChannelStates() {
```
swuferhong: No use. pls remove it.

fluss-server/src/test/java/org/apache/fluss/server/coordinator/channel/ControlRequestSendThreadTest.java
```diff
@@ -0,0 +1,349 @@
...
+                        null));
+
+        assertThat(secondCallbackLatch.await(5, TimeUnit.SECONDS)).isTrue();
+        // gateway invoked exactly twice (once per item, no re-send from callback failure)
+        assertThat(invocationCount.get()).isEqualTo(2);
+        assertThat(getCounter(metricGroup, MetricNames.SENDER_RETRY_COUNT).getCount()).isEqualTo(0);
+        assertThat(getGaugeValue(metricGroup, MetricNames.SENDER_ALIVE)).isEqualTo(1);
+    }
+
+    @SuppressWarnings("unchecked")
```
swuferhong: Redundant suppression, can be remove.

fluss-server/src/main/java/org/apache/fluss/server/coordinator/channel/TabletServerChannelState.java
```diff
@@ -0,0 +1,93 @@
...
+            ControlRequestSendThread sendThread,
+            MetricGroup metricGroup) {
+        this.tabletServerId = tabletServerId;
+        this.serverNode = serverNode;
+        this.queue = queue;
+        this.sendThread = sendThread;
+        this.metricGroup = metricGroup;
+    }
+
+    public int getTabletServerId() {
```
swuferhong: No user. remove it.

fluss-server/src/main/java/org/apache/fluss/server/coordinator/channel/TabletServerChannelState.java
```diff
@@ -0,0 +1,93 @@
...
+        return tabletServerId;
+    }
+
+    public ServerNode getServerNode() {
```
swuferhong: No user. remove it.

fluss-server/src/main/java/org/apache/fluss/server/coordinator/CoordinatorChannelManager.java
```diff
@@ -35,56 +39,145 @@
...
+        ControlRequestSendThread thread = state.getSendThread();
+        if (thread.getState() == Thread.State.NEW) {
+            thread.start();
+        }
     }
 
     public void removeTabletServer(Integer serverId) {
+        synchronized (channelLock) {
+            TabletServerChannelState state = channelStates.remove(serverId);
+            teardownChannelState(serverId, state);
```
swuferhong: Maybe this need to remove outside of the lock. ShutdownableThread.shutdown() blocks until the thread is fully joined, and we call it inside the synchronized (channelLock) block. The same lock guards three hot paths:
- the SENDER_TOTAL_QUEUE_SIZE gauge lambda — read by the metric-reporter thread;
- sendStopBucketReplicaRequest when it looks up the channel state
- getChannelStates()

So for as long as the join takes, metric reporting and control-plane sends to all other tablet servers are blocked on a single TS teardown.

swuferhong (commented):
Hi, @gyang94 thanks for your contributuon, it's an important feature, I left some comments:

fluss-server/src/main/java/org/apache/fluss/server/coordinator/CoordinatorEventProcessor.java
gyang94: if "TS stays alive, never reconnects", but always response with errors, the retry-with-backoff may cost more than "ineligible mark" (comparing with repeated StopReplica requests to a TS, and TS processed it and keeps returning errors), and for permanent errors like disk failure, that retry never terminates.

fluss-server/src/main/java/org/apache/fluss/server/coordinator/channel/ControlRequestSendThread.java
```diff
@@ -141,52 +149,152 @@ public void doWork() throws Exception {
...
                         tabletServerId,
-                        t);
+                        failure);
+                invokeCallback(item, null, failure);
+                return;
+            }
+
+            if (!invalidateConnection(item, failure)) {
+                invokeCallback(item, null, failure);
+                return;
```
swuferhong, line 216: Should we keep this request at the queue head when connection invalidation fails?

At this point, the outcome of the original request is still unknown. In addition, an already-disconnected `ServerConnection` may legitimately return an exceptionally completed close future. Treating that as a terminal failure invokes the callback and advances the FIFO, allowing later requests to overtake an uncertain request.

According to the design, transport failures and timeouts should retain and retry the same queue head while the `TabletServer` channel still exists. Could we retry connection invalidation/reconnection with backoff instead, and only discard the item when the channel is removed?

It would also be helpful to add a test where the connection invalidator fails once and then recovers, verifying that the original request is eventually retried successfully.

fluss-server/src/main/java/org/apache/fluss/server/coordinator/CoordinatorChannelManager.java
```diff
@@ -35,56 +38,133 @@
...
+
+        int id = serverNode.id();
+        synchronized (channelLock) {
+            if (channelStates.containsKey(id)) {
+                return;
+            }
+            BlockingQueue<QueueItem<?>> queue = new LinkedBlockingQueue<>();
+
+            MetricGroup tsGroup =
+                    coordinatorMetricGroup.addGroup("tablet-server-id", String.valueOf(id));
```
swuferhong: metric variable keys are `snake_case` everywhere else in the codebase — `server_id` and `cluster_id` in `CoordinatorMetricGroup`, `event_type` in CoordinatorEventMetricGroup, `tablet_type` in `TableMetricGroup`, `client_id `in `ClientMetricGroup`. `tablet-server-id` is the only kebab-case one and it becomes a user-visible label in the Prometheus reporters, so it is effectively a public name that we cannot rename later without breaking dashboards.

fluss-common/src/main/java/org/apache/fluss/config/ConfigOptions.java
```diff
@@ -576,6 +576,24 @@ public class ConfigOptions {
...
+            key("coordinator.request.retry-backoff")
+                    .durationType()
+                    .defaultValue(Duration.ofMillis(100))
+                    .withDescription(
+                            "The backoff duration the coordinator waits before retrying a "
+                                    + "control-plane request to a tablet server after a "
+                                    + "transient RPC-layer failure.");
+
+    public static final ConfigOption<Duration> COORDINATOR_REQUEST_TIMEOUT =
+            key("coordinator.request.timeout")
```
swuferhong: both options only govern the coordinator's control-plane channel to tablet servers, but `coordinator.request.timeout` reads like a generic timeout for any coordinator request, and it is easy to misread as a timeout for requests the coordinator receives. 


`coordinator.control-request.timeout` / `coordinator.control-request.retry-backoff` may make the blast radius obvious.

fluss-common/src/main/java/org/apache/fluss/config/ConfigOptions.java
```diff
@@ -576,6 +576,24 @@ public class ConfigOptions {
...
+                    .durationType()
```
swuferhong, line 590: add document at `configuration.md`

fluss-server/src/main/java/org/apache/fluss/server/coordinator/channel/ControlRequestSendThread.java
```diff
@@ -0,0 +1,304 @@
...
+                    backoffMs,
+                    failure);
+            backoff();
+        }
+    }
+
+    private boolean invalidateConnection(QueueItem<?> item, Throwable requestFailure)
+            throws InterruptedException {
+        try {
+            connectionInvalidator.get().get();
```
swuferhong: the send attempt itself is properly bounded by `requestTimeoutMs`, but the `connection` invalidation step that runs before every retry is an unbounded `get()`. If the future returned by `rpcClient.disconnect()` does not complete for any reason, the sender thread parks here forever and the queue head is stuck.


Two suggestions:
- Bound the wait, e.g. connectionInvalidator.get().get(requestTimeoutMs, TimeUnit.MILLISECONDS).
- 
- On invalidation failure/timeout, prefer backing off and retrying the same queue head over completing the item with the failure. Right now any invalidation failure calls invokeCallback(item, null, failure) and abandons the request, but the original failure was a transient transport failure, so the request's outcome on the tablet server is unknown and the tablet server is still registered. Dropping it here contradicts the "keep retrying the head while the registration exists" rule that the rest of the loop implements. Failing to close a connection should not turn an unknown-outcome request into a terminal failure for the caller.

```java
 try {
     connectionInvalidator.get().get(requestTimeoutMs, TimeUnit.MILLISECONDS);
     return true;
 } catch (TimeoutException e) {
     // closing the old connection timed out; the request outcome is still unknown,
     // so keep the item at the queue head and retry after the backoff
     log.warn("Timed out invalidating the connection to tabletServer {}", tabletServerId, e);
     return true;
 } catch (ExecutionException | RuntimeException e) {
     ...
 }
```

swuferhong (commented):
Hi, @gyang94  thanks for your contribution. I left some minor comments:

fluss-server/src/main/java/org/apache/fluss/server/coordinator/CoordinatorChannelManager.java
Jackeyzhe, line 230: should the stopReplica request use retry-sender?

gyang94: Eventually yes. I intentionally make this PR to only introduces the generic per-tablet-server sender/retry infrastructure. Migrating StopReplica, or deal with other specific requests, will be handled in  follow-up PRs.

Jackeyzhe: Thanks for clarifying — the scope split makes sense (generic per-tablet-server sender/retry infra first, specific request migration in follow-ups). Looking forward to the StopReplica migration in the follow-up.

leekeiabstraction:
@Jackeyzhe @swuferhong Do you have any further comments on this PR? TY

Jackeyzhe:
@leekeiabstraction Thanks for the ping. I have no further comments on the sender framework itself. The latest review points have been addressed, and the framework-first / request-migration-follow-up split makes sense to me.

One thing before merge: could we align the PR description and issue linkage with the current scope? The body still says `close #3357` and describes the StopReplica retry/resume integration, but at the current head (`060980ed`) `sendStopBucketReplicaRequest` still uses `sendRequest(...)`, while `enqueueRequest(...)` has no production caller. So the failure path reported in #3357 remains unchanged.

Could we keep #3357 open for the StopReplica follow-up and update the description accordingly? After that, I have no further blocker from my side. Happy to help review the follow-up.

gyang94:
> @leekeiabstraction Thanks for the ping. I have no further comments on the sender framework itself. The latest review points have been addressed, and the framework-first / request-migration-follow-up split makes sense to me.
> 
> One thing before merge: could we align the PR description and issue linkage with the current scope? The body still says `close #3357` and describes the StopReplica retry/resume integration, but at the current head (`060980ed`) `sendStopBucketReplicaRequest` still uses `sendRequest(...)`, while `enqueueRequest(...)` has no production caller. So the failure path reported in #3357 remains unchanged.
> 
> Could we keep #3357 open for the StopReplica follow-up and update the description accordingly? After that, I have no further blocker from my side. Happy to help review the follow-up.

Done. Created a new issue only about framework itself, and updated the description of this PR.

Jackeyzhe (approved):

polyzos:
@gyang94 Thanks for this and overall LGTM. One thing I will like to add here, maybe as a follow-up since this PR needs to be merged.

The introduced`LinkedBlockingQueue<>()` is unbounded, and the head item retries indefinitely on transport failure. While a tablet server is down, this means:
- the queue can grow without bound (memory pressure), and
- FIFO ordering lets a single stuck request block every later request to that TS.

Also the stale-epoch drop only fires on dequeue (doWork()), so the in-flight head from an older epoch keeps retrying even after an epoch bump, it's never re-checked once the retry loop has started.

This is ok for this PR since it introduces with no production enqueuer, but i think it should be tracked as a follow-up before any control RPC is migrated onto this path. Options to consider:
- bound the queue with a defined overflow policy,
- re-check the coordinator epoch inside the retry loop (so an epoch bump abandons a doomed in-flight item), and/or
- add a drop/skip policy so one unreachable request can't starve the rest.

Given the priority=blocker label, I think it's best to capture this in a follow-up issue (and ideally referenced in a code comment).

polyzos (approved):

gyang94:
@polyzos thanks. Added a issue ticket here: https://github.com/apache/fluss/issues/4291

Issue #3275, opened 2026-05-08
Introduce Hudi LakeCatalog to create table


fhan688:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/fluss/issues) and found nothing similar.

### Description

Introduce Hudi LakeCatalog to create table

### Willingness to contribute

- [x] I'm willing to submit a PR!

luoyuxia:
Close via #3395

PR #3395, opened 2026-05-28
[lake/hudi] Introduce Hudi LakeCatalog to create table

fhan688:
### Purpose

Linked issue: https://github.com/apache/fluss/issues/3275

This PR introduces the Hudi `LakeCatalog` implementation, enabling Fluss to create tables in Hudi data lake storage. This aligns with the existing Paimon and Iceberg lake catalog support, completing the trio of supported lake formats for table creation.

#### New modules & classes (`fluss-lake/fluss-lake-hudi`):

- `HudiLakeCatalog`: Implements `LakeCatalog` interface, supporting both HMS (Hive Metastore) and DFS (filesystem) catalog modes. Handles table creation with schema compatibility check for crash-recovery idempotency, automatic database creation, and system column (`__bucket, __offset, __timestamp`) appending.

- `FlussDataTypeToHudiDataType`: Implements `DataTypeVisitor` to convert Fluss data types to Flink types (Hudi's type system). Handles `LocalZonedTimestampType` specially: maps to `BIGINT` under HMS mode, `TIMESTAMP_WITH_LOCAL_TIME_ZONE` under DFS mode.

- `HudiConversions`: Core conversion utility. Converts Fluss `TablePath` → Hudi `ObjectPath`, `TableDescriptor` → `ResolvedSchema` / Hudi table properties. Validates `HUDI_UNSETTABLE_OPTIONS` (6 protected options that Fluss auto-manages), checks system column name conflicts, and handles property prefix rewriting (`hudi.xxx` → `xxx`, others → `fluss.xxx`).

- `HudiCatalogUtils`: Factory for creating Hudi `Catalog` instances (`HoodieHiveCatalog` for HMS, `HoodieCatalog for` DFS). Uses copied Configuration to avoid mutating the original.

#### Modifications to existing modules (`fluss-flink/fluss-flink-common`):

- `LakeFlinkCatalog`: Adds HUDI branch in `getLakeCatalog()` with a new `HudiCatalogFactory` inner class that uses reflection to instantiate Hudi catalog (mirroring the Iceberg pattern to avoid compile-time dependency on hudi-flink-bundle).

- `LakeTableFactory`: Adds HUDI branch in `getLakeTableFactory()` with `getHudiFactory()` that reflectively loads `HoodieTableFactory`.

- `HudiLakeStorage`: Replaces the `UnsupportedOperationException` in `createLakeCatalog()` with `new HudiLakeCatalog(hudiConfig)` to wire the SPI path.

Key design decisions:

| Aspect | Decision |
|--------|----------|
| Table type mapping | PK table → MERGE_ON_READ, Log table → COPY_ON_WRITE |
| Index strategy | BUCKET index type, aligned with Fluss's bucketing model |
| Dependency isolation | Hudi bundle loaded via reflection/plugin classloader (no compile-time dependency in fluss-flink-common) |
| Catalog mode | Supports hms (Hive Metastore) and dfs (filesystem) |
| Property rewriting | hudi. prefix stripped; non-hudi properties prefixed with fluss. |
| Idempotency | Schema-compatible duplicate creation is treated as success for crash recovery |

### Tests

`HudiLakeCatalogTest` (14 test cases):

- `testPropertyPrefixRewriting` — verifies hudi.xxx → xxx and non-hudi → fluss.xxx prefix rewriting

- `testCreatePrimaryKeyTable` — PK table (MOR) creation with system columns and primary key

- `testCreateLogTable` — Log table (COW) creation with record key from customProperties

- `testIsHudiSchemaCompatibleWithSameSchema` — compatible schemas return true

- `testIsHudiSchemaCompatibleWithDifferentColumnCount` — different column count returns false

- `testIsHudiSchemaCompatibleWithDifferentColumnName` — different column name returns false

- `testIsHudiSchemaCompatibleWithDifferentColumnType` — different column type returns false

- `testCreateDuplicateTableWithCompatibleSchema` — duplicate creation with compatible schema is idempotent

- t`estCreateDuplicateTableWithIncompatibleSchema` — duplicate creation with incompatible schema throws `TableAlreadyExistException`

- `testUnsettableOptionInPropertiesThrowsException` — protected option in `properties` throws `InvalidConfigException`

- `testUnsettableOptionInCustomPropertiesThrowsException` — protected option in customProperties throws `InvalidConfigException`

- `testNonProtectedHudiOptionPassesValidation` — non-protected option passes validation

- `testSystemColumnBucketConflictThrowsException` — `__bucket` conflict throws `InvalidTableException`

- `testSystemColumnOffsetConflictThrowsException` — `__offset` conflict throws `InvalidTableException`

- `testSystemColumnTimestampConflictThrowsException` — `__timestamp` conflict throws `InvalidTableException`

### API and Format

No API or storage format changes. This PR only adds new implementations for the existing `LakeCatalog` and `LakeStorage SPI` interfaces.

### Documentation

A new feature — Hudi lake catalog support for table creation. Will need documentation updates for the Hudi integration guide.


fhan688:
all tests are passed, please help review, thanks! @XuQianJin-Stars

fluss-lake/fluss-lake-hudi/src/main/java/org/apache/fluss/lake/hudi/HudiLakeCatalog.java
```diff
@@ -0,0 +1,260 @@
...
+        boolean isPkTable = tableDescriptor.getSchema().getPrimaryKeyIndexes().length > 0;
+
+        // Create Hudi catalog table
+        CatalogTable catalogTable =
+                HudiConversions.createHudiCatalogTable(tableDescriptor, isPkTable, catalogMode);
+
+        // Create table in Hudi catalog
+        try {
+            createTable(objectPath, catalogTable, context.isCreatingFlussTable());
+        } catch (DatabaseNotExistException e) {
```
XuQianJin-Stars, line 108: The inner `createTable` throws `org.apache.fluss.exception.TableAlreadyExistException` (unchecked) on schema incompatibility, which is fine. The outer method signature declares `throws org.apache.fluss.exception.TableAlreadyExistException` — looks like it intends to comply with the `LakeCatalog.createTable` contract — but the exception **is not actually thrown as checked via the `throws` clause** because it's a `RuntimeException` subclass. OK, that's not really a problem. However:

**The real bug is in the test `testCreateDuplicateTableWithIncompatibleSchema`**: based on the implementation of `isHudiSchemaCompatible(existingTable, catalogTable)` (with the `ResolvedCatalogBaseTable` vs `Schema.UnresolvedColumn` branches), the `catalogTable` passed into `hudiCatalog.createTable()` is a `CatalogTable.of(...)` (**unresolved**), while `hudiCatalog.getTable()` may return a `ResolvedCatalogBaseTable`. `extractColumns` will go through two different code paths (one uses `column.getDataType()` returning `DataType`, the other uses `UnresolvedColumn` returning `AbstractDataType<?>`). Comparison between `AbstractDataType` and `DataType` via `equals` is not necessarily symmetric — even with identical column names/types it may be judged as "incompatible", and conversely different schemas may fail comparison because the `AbstractDataType` is not fully resolved.

**Suggestion**: First normalize both `existingTable` and `expectedTable` into the same representation (recommended: compare using `org.apache.flink.table.types.logical.LogicalType`, or convert via `getUnresolvedSchema()` to `Schema` then compare on column name + `AbstractDataType`). The dataType comparison should use the `LogicalType.equals` dimension (stripping the conversion class). Paimon's approach with `isPaimonSchemaCompatible` — directly comparing `name+type+nullable` on `Paimon Schema` fields — is more robust.

fluss-lake/fluss-lake-hudi/src/main/java/org/apache/fluss/lake/hudi/HudiLakeCatalog.java
```diff
@@ -0,0 +1,260 @@
...
+                    throw new org.apache.fluss.exception.TableAlreadyExistException(
+                            String.format(
+                                    "The table %s already exists in Hudi catalog, but the table schema is not compatible. "
+                                            + "Please first drop the table in Hudi catalog or use a new table name.",
+                                    tablePath),
+                            e);
+                }
+                // if creating a new fluss table, we should ensure the lake table is empty
+                // TODO: add emptiness check for Hudi table once LakeTieringFactory is implemented
+                if (isCreatingFlussTable) {
```
XuQianJin-Stars, line 151: Compare with `PaimonLakeCatalog.createTable` (line 180) which directly calls `checkTableIsEmpty(...)`. Without the emptiness check, if a user attaches a new Fluss table to a Hudi table that **already contains data**, data corruption/mixing will occur. Even though TieringFactory is not yet ready, at the very least the semantics should be documented (in a comment or in the thrown exception), and consider throwing `TableAlreadyExistException` here as a safety fallback (safer), then loosening it once TieringFactory is ready. **At minimum, do not silently warn and return success.**

fluss-lake/fluss-lake-hudi/src/main/java/org/apache/fluss/lake/hudi/HudiLakeCatalog.java
```diff
@@ -0,0 +1,260 @@
...
+        // Create table in Hudi catalog
+        try {
+            createTable(objectPath, catalogTable, context.isCreatingFlussTable());
+        } catch (DatabaseNotExistException e) {
+            createDatabase(tablePath.getDatabaseName());
+            try {
+                createTable(objectPath, catalogTable, context.isCreatingFlussTable());
+            } catch (DatabaseNotExistException t) {
+                // shouldn't happen in normal cases
+                throw new RuntimeException(
```
XuQianJin-Stars, line 114: This throws a raw `RuntimeException` without passing t as the cause, losing the stack trace. Please change to new `RuntimeException`(msg, t). Similarly at line 144, the inner throw `new RuntimeException`(...) (schema fallback) also drops the cause (`tableNotExistException`) — please add it.

fluss-lake/fluss-lake-hudi/src/main/java/org/apache/fluss/lake/hudi/utils/catalog/HudiCatalogUtils.java
```diff
@@ -0,0 +1,106 @@
...
+
+    public static final String MODE_CONFIG = "mode";
+    public static final String CATALOG_NAME_CONFIG = "name";
+
+    public static final String HUDI_CATALOG_DEFAULT_NAME = "fluss-hudi-catalog";
+    public static final String HIVE_META_STORE_TYPE = "hms";
+    public static final String FILE_SYSTEM_TYPE = "dfs";
+
+    public static Catalog createHudiCatalog(Configuration configuration) {
+        Map<String, String> hudiProps = configuration.toMap();
```
XuQianJin-Stars, line 52: `createHudiCatalog` uses the custom `MODE_CONFIG = "mode"` read in `HudiLakeCatalog`, while `buildHudiCatalog` internally uses `CatalogOptions.MODE.key()` (also `"mode"`) — same string by coincidence, but maintaining two sets of constants invites future drift. Standardize on `CatalogOptions.MODE.key()` as the authoritative key and drop the `MODE_CONFIG` constant to avoid long-term maintenance drift. Same for `CATALOG_NAME_CONFIG = "name"` — recommend using a public Hudi constant.

fluss-lake/fluss-lake-hudi/src/main/java/org/apache/fluss/lake/hudi/FlussDataTypeToHudiDataType.java
```diff
@@ -0,0 +1,180 @@
...
+import org.apache.flink.table.api.DataTypes;
+import org.apache.flink.table.types.DataType;
+
+import static org.apache.fluss.lake.hudi.utils.catalog.HudiCatalogUtils.FILE_SYSTEM_TYPE;
+import static org.apache.fluss.lake.hudi.utils.catalog.HudiCatalogUtils.HIVE_META_STORE_TYPE;
+
+/** Convert from Fluss's data type to Hudi's internal data type. */
+public class FlussDataTypeToHudiDataType implements DataTypeVisitor<DataType> {
+
+    private String catalogMode;
```
XuQianJin-Stars: - The `catalogMode` field is not declared `final`, yet the class exposes a public constructor and provides two **singleton INSTANCEs**. If any caller obtains `DFS_INSTANCE` and accidentally invokes a future mutator, concurrency issues arise. Please change the field to `private final`, and either remove the public constructor or change it to `private` (external code uses the two INSTANCEs uniformly).
- The class-level Javadoc lacks the `@ThreadSafe` annotation; per spec (AGENTS.md §5), multiple threads will share the same INSTANCE, so this should be explicitly marked.

fluss-lake/fluss-lake-hudi/src/main/java/org/apache/fluss/lake/hudi/utils/HudiConversions.java
```diff
@@ -0,0 +1,271 @@
...
+    private static String getRecordKeyField(TableDescriptor tableDescriptor) {
+        String recordKeyOption = HUDI_CONF_PREFIX + FlinkOptions.RECORD_KEY_FIELD.key();
+        String recordKeyField = tableDescriptor.getCustomProperties().get(recordKeyOption);
+        if (recordKeyField == null) {
+            recordKeyField = tableDescriptor.getProperties().get(recordKeyOption);
+        }
+        return recordKeyField;
+    }
+
+    private static void validateHudiOptions(Map<String, String> properties, boolean isPkTable) {
```
XuQianJin-Stars, line 277: Logic: non-PK tables allow users to set `RECORD_KEY_FIELD` (combined with `getRecordKeyField`); PK tables forbid setting it. **However, the test `testUnsettableOptionInCustomPropertiesThrowsException`** expects a PK table passing `hudi.hoodie.datasource.write.recordkey.field` to fail; this rule indeed works at runtime. But the test at line 152 `testCreateLogTable` sets the same key on a **non-PK table** and **expects success** — that looks OK. Readability issue: the two validation scenarios are easy to confuse — **suggest adding a Javadoc on `validateHudiOptions`** explaining that `RECORD_KEY_FIELD` is a legal usage for non-PK tables (i.e., "log table = use the user-specified primary key field as the index field").

More importantly, after the function lets `RECORD_KEY_FIELD` pass, it doesn't validate that "non-PK tables must provide `RECORD_KEY_FIELD`"; that's done in `buildHudiTableProperties` via `IllegalArgumentException("Record key field should be set.")`. **`IllegalArgumentException` does not belong to the Fluss `InvalidConfigException` family** (Section 4); the user sees a raw IAE. This should be changed to `InvalidConfigException` with helpful hints including `tablePath`/the specific key for easier troubleshooting.

fluss-lake/fluss-lake-hudi/src/main/java/org/apache/fluss/lake/hudi/HudiLakeCatalog.java
```diff
@@ -0,0 +1,260 @@
...
+
+        Schema schema = table.getUnresolvedSchema();
+        List<ColumnSignature> columns = new ArrayList<>();
+        for (Schema.UnresolvedColumn column : schema.getColumns()) {
+            columns.add(new ColumnSignature(column.getName(), getDataType(column)));
+        }
+        return columns;
+    }
+
+    private static AbstractDataType<?> getDataType(Schema.UnresolvedColumn column) {
```
XuQianJin-Stars, line 208: `getDataType` does not handle `UnresolvedComputedColumn` and silently returns `null`, `Schema.UnresolvedComputedColumn` has no datatype; returning `null` causes `ColumnSignature.equals` to mistakenly consider different computed columns equal. Hudi tables shouldn't have computed columns in theory, but a safer approach is to throw `IllegalStateException("Unexpected column kind: " + column.getClass())`.

fluss-lake/fluss-lake-hudi/src/main/java/org/apache/fluss/lake/hudi/utils/HudiConversions.java
```diff
@@ -0,0 +1,271 @@
...
+        FlussDataTypeToHudiDataType converter =
+                HIVE_META_STORE_TYPE.equals(catalogMode)
+                        ? FlussDataTypeToHudiDataType.HMS_INSTANCE
+                        : FlussDataTypeToHudiDataType.DFS_INSTANCE;
+
+        List<Column> columns = new ArrayList<>();
+
+        // Add regular columns
+        for (org.apache.fluss.metadata.Schema.Column column :
+                tableDescriptor.getSchema().getColumns()) {
```
XuQianJin-Stars, line 111: `SYSTEM_COLUMNS` (defined at `HudiLakeCatalog.java` L66–72) adds `__bucket / __offset / __timestamp`, but Hudi itself reserves `_hoodie_commit_time` / `_hoodie_record_key` / `_hoodie_partition_path`, etc. In `HudiConversions.createHudiCatalogTable` (L105–119), only collisions against Fluss system columns are checked before writing fluss columns + system columns into the schema. Suggest adding a rejection rule for the `_hoodie_` prefix (or a "reserved column" set parallel to `HUDI_UNSETTABLE_OPTIONS`) with corresponding tests.

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/lake/LakeFlinkCatalog.java
```diff
@@ -86,6 +87,11 @@ public Catalog getLakeCatalog(
                     } else if (lakeFormat == ICEBERG) {
                         catalog = IcebergCatalogFactory.create(catalogName, catalogProperties);
                         this.lakeFormat = ICEBERG;
+                    } else if (lakeFormat == HUDI) {
```
luoyuxia: Could you please remove these changes in `LakeFlinkCatalog` and `LakeTableFactory` in this pr. Let's focus Hudi own catalog instead of flink catalog in this pr.
We do add them back in another pr which requirs a IT to test we can use $lake to read hudi table

fluss-lake/fluss-lake-hudi/src/main/java/org/apache/fluss/lake/hudi/FlussDataTypeToHudiDataType.java
```diff
@@ -0,0 +1,183 @@
...
+
+import org.apache.flink.table.api.DataTypes;
+import org.apache.flink.table.types.DataType;
+
+import javax.annotation.concurrent.ThreadSafe;
+
+import static org.apache.fluss.lake.hudi.utils.catalog.HudiCatalogUtils.FILE_SYSTEM_TYPE;
+import static org.apache.fluss.lake.hudi.utils.catalog.HudiCatalogUtils.HIVE_META_STORE_TYPE;
+
+/** Convert from Fluss's data type to Hudi's internal data type. */
```
luoyuxia: nit: not Hudi's internal data type. It's flink internal data type.
Also add comments why convert to flink internal data type intead of Hudi's internal data type.

fluss-lake/fluss-lake-hudi/src/main/java/org/apache/fluss/lake/hudi/HudiLakeCatalog.java
```diff
@@ -0,0 +1,277 @@
...
+    @Override
+    public void createTable(TablePath tablePath, TableDescriptor tableDescriptor, Context context)
+            throws org.apache.fluss.exception.TableAlreadyExistException {
+        LOG.info("create the lake table for : {} with props: {}", tablePath, tableDescriptor);
+
+        ObjectPath objectPath = HudiConversions.toHudiObjectPath(tablePath);
+
+        boolean isPkTable = tableDescriptor.getSchema().getPrimaryKeyIndexes().length > 0;
+
+        // Create Hudi catalog table
```
luoyuxia: could you please add comment that why convert to flink catalog table and call flink catalog to create hudi table?

fluss-lake/fluss-lake-hudi/src/test/java/org/apache/fluss/lake/hudi/HudiLakeCatalogTest.java
```diff
@@ -0,0 +1,645 @@
...
+        flussHudiLakeCatalog.createTable(tablePath, tableDescriptor, context);
+
+        ObjectPath objectPath = HudiConversions.toHudiObjectPath(tablePath);
+        CatalogBaseTable table = flussHudiLakeCatalog.getHudiCatalog().getTable(objectPath);
+
+        assertThat(table).isNotNull();
+
+        List<String> primaryKeys = new ArrayList<>();
+        primaryKeys.add("id");
+        TableSchema expectHudiSchema =
```
luoyuxia: nit: 'org.apache.flink.table.api.TableSchema' is deprecated

fluss-lake/fluss-lake-hudi/src/test/java/org/apache/fluss/lake/hudi/HudiLakeCatalogTest.java
```diff
@@ -0,0 +1,645 @@
...
+        TablePath tablePath = TablePath.of(database, tableName);
+        TestingLakeCatalogContext context = new TestingLakeCatalogContext();
+        flussHudiLakeCatalog.createTable(tablePath, td, context);
+
+        ObjectPath objectPath = HudiConversions.toHudiObjectPath(tablePath);
+        CatalogBaseTable table = flussHudiLakeCatalog.getHudiCatalog().getTable(objectPath);
+
+        List<String> primaryKeys = new ArrayList<>();
+        primaryKeys.add("id");
+        TableSchema expectHudiSchema =
```
luoyuxia: dito

fluss-lake/fluss-lake-hudi/src/test/java/org/apache/fluss/lake/hudi/HudiLakeCatalogTest.java
```diff
@@ -0,0 +1,645 @@
...
+                .hasMessageContaining("hudi.hoodie.datasource.write.recordkey.field")
+                .hasMessageContaining(tablePath.toString());
+    }
+
+    // ------------------------------------------------------------------
+    // isHudiSchemaCompatible() tests
+    // ------------------------------------------------------------------
+
+    @Test
+    void testIsHudiSchemaCompatibleWithSameSchema() {
```
luoyuxia, line 198: nit: combine  testIsHudiSchemaCompatible releated test into single one test method?

luoyuxia (commented):
@fhan688 Thanks for the pr. LGTM overall. Left minor comment. 
@XuQianJin-Stars Do you want to have another round of review again?

XuQianJin-Stars (approved):
+1

fluss-lake/fluss-lake-hudi/src/main/java/org/apache/fluss/lake/hudi/HudiLakeCatalog.java
luoyuxia: Update: after disucss with author offline, unlike Iceberg and Paimon which provide their own catalog abstractions, Hudi does not have a standalone catalog API. We convert to a Flink catalog table and delegate to the Flink catalog to create the Hudi table.

luoyuxia (approved):
@fhan688 @XuQianJin-Stars Thanks. Merging..

PR #3397, opened 2026-05-28
[spark] Split non-partition and partition predicates from pushPredicates to limit pushdown

Yohahaha:
### Purpose

Linked issue: close #xxx

Fix a bug where LIMIT is never pushed down when a query has partition filters (e.g., WHERE dt = 'x' LIMIT 1). The root cause is that pushPredicates returns all predicates including partition ones, causing Spark to keep a Filter node. Spark's pushDownLimit only invokes pushLimit when there are no filters (PhysicalOperation(_, Nil, ...)), so the Filter node blocks limit pushdown.

### Brief change log

 - `SparkPartitionPredicate.scala`: Refactor extract() to return nonPartitionPredicates and partitionPredicate.
 - `FlussScanBuilder.scala`: Return only non-partition predicates from pushPredicates() across all scan builders (partition filters, ARROW filters, lake filters).

### DIFF
--- a/fluss-spark/fluss-spark-common/src/main/scala/org/apache/fluss/spark/read/FlussScanBuilder.scala
+++ b/fluss-spark/fluss-spark-common/src/main/scala/org/apache/fluss/spark/read/FlussScanBuilder.scala
@@ -63,7 +63,6 @@ trait FlussSupportsPushDownPartitionFilters
   override def pushedPredicates(): Array[Predicate] = acceptedPredicates
 }
 
-/** Adds ARROW server-side log filtering on top of partition pushdown. */
 trait FlussSupportsPushDownV2Filters extends FlussSupportsPushDownPartitionFilters {
 
   protected def convertAndStorePredicates(predicates: Array[Predicate]): Unit = {
@@ -75,8 +74,8 @@ trait FlussSupportsPushDownV2Filters extends FlussSupportsPushDownPartitionFilte
 
   override def pushPredicates(predicates: Array[Predicate]): Array[Predicate] = {
     val nonPartitionPredicates = super.pushPredicates(predicates)
-    // Server-side batch filter only supports ARROW; other log formats reject it.
-    if (tableInfo.getTableConfig.getLogFormat == LogFormat.ARROW) {
+    if (!tableInfo.hasPrimaryKey && tableInfo.getTableConfig.getLogFormat == LogFormat.ARROW) {
+      // Server-side batch filter for log table only supports ARROW; other log formats reject it.
       convertAndStorePredicates(nonPartitionPredicates)
     }
     nonPartitionPredicates
--- a/fluss-spark/fluss-spark-common/src/main/scala/org/apache/fluss/spark/read/FlussScanBuilder.scala
+++ b/fluss-spark/fluss-spark-common/src/main/scala/org/apache/fluss/spark/read/FlussScanBuilder.scala
@@ -156,6 +156,7 @@ class FlussLakeAppendScanBuilder(
       tableInfo,
       requiredSchema,
       pushedPredicate,
+      partitionPredicate,
       acceptedPredicates.toSeq,
       options,
       flussConfig)
@@ -189,6 +190,7 @@ class FlussLakeUpsertScanBuilder(
       tableInfo,
       requiredSchema,
       pushedPredicate,
+      partitionPredicate,
       acceptedPredicates.toSeq,
       options,
       flussConfig)
--- a/fluss-spark/fluss-spark-common/src/main/scala/org/apache/fluss/spark/read/FlussScanBuilder.scala
+++ b/fluss-spark/fluss-spark-common/src/main/scala/org/apache/fluss/spark/read/FlussScanBuilder.scala
@@ -50,6 +50,8 @@ trait FlussSupportsPushDownPartitionFilters
   def tableInfo: TableInfo
 
   protected var partitionPredicate: Option[FlussPredicate] = None
+  protected var pushedPredicate: Option[FlussPredicate] = None
+  protected var acceptedPredicates: Array[Predicate] = Array.empty[Predicate]
 
   override def pushPredicates(predicates: Array[Predicate]): Array[Predicate] = {
     val (nonPartitionPred, partitionPred) =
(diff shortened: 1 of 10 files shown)


Yohahaha:
@fresh-borzoni @luoyuxia @YannByron PTAL!

fluss-spark/fluss-spark-common/src/main/scala/org/apache/fluss/spark/utils/SparkPartitionPredicate.scala
```diff
@@ -28,25 +28,40 @@ import scala.jdk.CollectionConverters._
...
 
     val rowType = PartitionUtils.partitionRowType(tableInfo)
     val onlyPartitionKeys = new PartitionPredicateVisitor(partitionKeys)
 
-    val converted = predicates.flatMap {
+    // Separate predicates: those that can be converted and only touch partition keys
+    val (partitionPredicates, nonPartitionPredicates) = predicates.partition {
+      sparkPredicate =>
+        SparkPredicateConverter
+          .convert(rowType, sparkPredicate)
```
fresh-borzoni: Small thing: we call convert(...) here to classify the predicate, then call it again right below to actually use it. Could we convert once and reuse the result? 

Something like mapping each predicate to its converted form, then partition-ing on isDefined. Only runs at planning time so not a big deal, just avoids doing the same work twice.

fluss-spark/fluss-spark-ut/src/test/scala/org/apache/fluss/spark/SparkPrimaryKeyTableReadTest.scala
```diff
@@ -432,6 +434,28 @@ class SparkPrimaryKeyTableReadTest extends FlussSparkTestBase {
     }
   }
 
+  test("Spark Read: mixed partition and non-partition filter (PK table)") {
```
fresh-borzoni, line 437: Could we add the opposite of the mixed-filter test above? i.e. a partition-only query (WHERE dt = '...') that asserts there's no FilterExec left in the plan. That's really the core thing this PR changes, and right now nothing checks it - easy to accidentally regress later.

fluss-spark/fluss-spark-common/src/main/scala/org/apache/fluss/spark/read/FlussScanBuilder.scala
```diff
@@ -96,8 +98,11 @@ trait FlussLakeSupportsPushDownV2Filters extends FlussSupportsPushDownV2Filters
   def flussConfig: FlussConfiguration
 
   override def pushPredicates(predicates: Array[Predicate]): Array[Predicate] = {
+    val nonPartitionPredicates = super.pushPredicates(predicates)
```
fresh-borzoni, line 97: Here the lake path calls super.pushPredicates, which for ARROW tables runs convertAndStorePredicates and sets pushedPredicate/acceptedPredicates, then we overwrite both a few lines down. 
So that work gets thrown away every time. 

Could we pull the partition-extraction bit into a small helper and call that instead of super, so we skip the ARROW step we don't use?

fluss-spark/fluss-spark-common/src/main/scala/org/apache/fluss/spark/utils/SparkPartitionPredicate.scala
```diff
@@ -63,4 +78,19 @@ object SparkPartitionPredicate {
               PartitionUtils.toPartitionRow(p.getResolvedPartitionSpec.getPartitionValues, rowType))
         }
     }
+
+  /**
+   * Tests whether a partition (described by its ordered partition values) matches the given
+   * predicate. Returns true when no predicate is provided.
+   */
+  def matchesPartition(
```
fresh-borzoni, line 80: If partitionValues ever comes in empty, toPartitionRow gets the wrong number of values and could blow up. 

Should we just guard it (case Some(_) if partitionValues.isEmpty => true)?

fresh-borzoni (commented):
@Yohahaha Ty for the PR and nice catch, LGTM overall.
A couple of minor comments, PTAL

fluss-spark/fluss-spark-common/src/main/scala/org/apache/fluss/spark/read/FlussScanBuilder.scala
Yohahaha: I did a small refactor on filter pushdown. Lake and non-lake paths can now implement
pushdown logic freely. The common partition pushdown logic is in the parent class.

You can see the new interface and class inheritance structure below
```
SupportsPushDownV2Filters (Spark)
│
└── FlussSupportsPushDownPartitionFilters
    │
    ├── FlussSupportsPushDownV2Filters
    │   │
    │   ├── FlussAppendScanBuilder
    │   └── FlussUpsertScanBuilder
    │
    └── FlussLakeSupportsPushDownV2Filters
        │
        ├── FlussLakeAppendScanBuilder
        └── FlussLakeUpsertScanBuilder

```

fluss-spark/fluss-spark-common/src/main/scala/org/apache/fluss/spark/read/FlussScanBuilder.scala
```diff
@@ -163,7 +166,7 @@ class FlussUpsertScanBuilder(
     val tableInfo: TableInfo,
     options: CaseInsensitiveStringMap,
     val flussConfig: FlussConfiguration)
-  extends FlussSupportsPushDownPartitionFilters {
+  extends FlussSupportsPushDownV2Filters {
```
fresh-borzoni, line 168: For PK we read from the KV snapshot and never apply ARROW filters - so convertAndStorePredicates just gets thrown away, and pushedPredicates() reports filters the scan doesn't apply. Results are still fine (Spark re-checks them), just wasted work + misleading.

fresh-borzoni (commented):
@Yohahaha Good refactoring, thank you, only one minor comment

fluss-spark/fluss-spark-common/src/main/scala/org/apache/fluss/spark/read/FlussScanBuilder.scala
Yohahaha: thank you for the input, I just update comments to highlight only log table was supportted.

for pk table, convertAndStorePredicates never be called, pushedPredicates returns empty which means no filter was pushable, I think the behavior is spark expected.

fresh-borzoni (approved):
@Yohahaha Thank you for the changes, LGTM 👍

PR #3319, opened 2026-05-14
[FIP-37] Add bitmap infrastructure: BitmapUtils, RoaringBitmapSerializer, AbstractRbAggFunction

pbanakar:
### Purpose

Linked issue: Part of #3289

This PR adds the foundational infrastructure for FIP-37 RoaringBitmap SQL function implementation. It provides the serialization utilities, custom Flink type serializer, and base aggregate function class that will be used by the bitmap SQL functions (`rb_build_agg`, `rb_or_agg`, `rb_and_agg`, etc.) in subsequent PRs.

### Brief change log

Added the following infrastructure files in `fluss-flink/fluss-flink-common`:

- **BitmapUtils.java**: Utility methods for serializing/deserializing `RoaringBitmap` using the ByteBuffer-based approach, which matches the server-side `RoaringBitmapUtils.serializeRoaringBitmap32` format used by `FieldRoaringBitmap32Agg` for wire compatibility.
- **RoaringBitmapSerializer.java**: Custom Flink `TypeSerializer` for `RoaringBitmap` accumulators to ensure correct checkpoint/savepoint behavior. Without this, Flink falls back to Kryo which is sensitive to internal class layout changes across RoaringBitmap library versions.
- **RoaringBitmapTypeInfo.java**: `TypeInformation` wrapper that provides the custom serializer to Flink's type system.
- **AbstractRbAggFunction.java**: Base class for bitmap aggregate UDFs with `@FunctionHint(accumulator = @DataTypeHint(value = "RAW", bridgedTo = RoaringBitmap.class))` annotation. This tells Flink's Table planner to skip reflection-based POJO field extraction on RoaringBitmap and use the custom `TypeInformation` instead.
- **BitmapUtilsTest.java**: Unit tests covering null handling, empty bitmap, known values round-trip, large cardinality (100K elements), and server serialization compatibility.
- **RoaringBitmapSerializerTest.java**: Unit tests covering `RoaringBitmapSerializer` and `RoaringBitmapTypeInfo` behavior, including serialization round-trip, copy methods, snapshot configuration, and type information checks.
- **pom.xml**: Added `RoaringBitmap` dependency (version 1.3.0 from root pom).

The aggregate functions (`rb_build_agg`, `rb_or_agg`, `rb_and_agg`) and catalog registration will follow in subsequent PRs linked to this issue.

### Tests

Unit tests added and passing:
- `BitmapUtilsTest.testNullInputToBytes()` - null handling.
[...]

Issue #3289, opened 2026-05-09
[FIP-37] Implement RoaringBitmap SQL functions via FlussCatalog


pbanakar:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/fluss/issues) and found nothing similar.

### Motivation

This issue tracks the implementation of FIP-37: https://cwiki.apache.org/confluence/display/FLUSS/FIP-37%3A+Native+RoaringBitmap+Integration+for+Apache+Fluss
Sub-tasks will be tracked as linked PRs.

### Willingness to contribute

- [x] I'm willing to submit a PR!

wuchong:
Thanks @pbanakar for the great work on FIP-37. I think we have implemented all the features and docs for this FIP. And I will close this issue.

wuchong:
However, I found a small bug https://github.com/apache/fluss/issues/3901, could you help to fix it if you are interested?

PR #3398, opened 2026-05-28
[FIP-37] Add bitmap aggregate functions: rb_build_agg, rb_or_agg, rb_and_agg and register via FlussCatalog

pbanakar:
### Purpose

Linked issue: Part of #3289

This PR adds the three Phase 1 bitmap aggregate functions and registers them as built-in catalog functions in FlussCatalog, building on the infrastructure from PR #3319 

After `USE CATALOG fluss_catalog`, users can call these functions directly in Flink SQL without any `CREATE TEMPORARY FUNCTION` statement.

### Brief change log

**New files in `fluss-flink/fluss-flink-common`:**

- `RbBuildAggFunction.java` — `rb_build_agg(INT) -> BYTES`. Aggregates a stream of 32-bit integers into a serialized RoaringBitmap. Extends `AbstractRbAggFunction` from PR #3319.

- `RbOrAggFunction.java` — `rb_or_agg(BYTES) -> BYTES`. Unions multiple serialized RoaringBitmaps via bitwise OR across rows. Extends `AbstractRbAggFunction`.

- `RbAndAggFunction.java` — `rb_and_agg(BYTES) -> BYTES`. Intersects multiple serialized RoaringBitmaps via bitwise AND across rows. Unlike the OR/build variants, AND requires a custom `Accumulator` POJO with an `initialized` flag because an empty bitmap cannot serve as a "not yet initialized" sentinel — once the AND result becomes empty, it must remain empty. The inner `AccumulatorSerializer` follows the same pattern as `RoaringBitmapSerializer`.
-  `RbAggFunctionsTest.java` - Unit tests covering:
 `rb_build_agg`: basic accumulation, null handling, duplicate deduplication, merge, retract throws
`rb_or_agg`: union correctness, null/empty input handling, merge
`rb_and_agg`: intersection correctness, null inputs ignored, empty intersection returns null, merge, reset, Accumulator serializer round-trip
- `RbFunctionsCatalogITCase` with four tests exercising all three functions through the catalog registration path using a real TableEnvironment.

**Modified file: `FlinkCatalog.java`**

Overrides `listFunctions()`, `functionExists()`, and `getFunction()` to serve the three aggregate functions as built-in catalog functions via a static `BUILTIN_BITMAP_FUNCTIONS` map using `CatalogFunctionImpl` from `flink-table-api-java`.

**Catalog registration verified in `FlinkCatalogTest.testBitmapFunctionsRegistered`**

Note: `AccumulatorSerializer` in `RbAndAggFunction` uses `TypeSerializerSingleton`, which is deprecated in Flink 1.20 but is the same approach used in `RoaringBitmapSerializer` (PR #3319, reviewed and approved). This will be revisited when  upgrading to Flink 2.x.

The scalar functions (rb_cardinality, rb_build, rb_contains, rb_to_array, rb_or, rb_and) will follow in the next PR.

### Tests

Unit tests: `RbAggFunctionsTest` — 14 tests, all passing.

Catalog integration: `FlinkCatalogTest.testBitmapFunctionsRegistered` and `testViewsAndFunctions` — verified against a live Fluss cluster.

Verified locally:
- `./mvnw spotless:apply` — 0 violations
- `./mvnw test -pl fluss-flink/fluss-flink-common -Dtest="BitmapUtilsTest,RbAggFunctionsTest,RoaringBitmapSerializerTest,AbstractRbAggFunctionITCase,FlinkCatalogTest"` — BUILD SUCCESS
- `./mvnw verify -pl fluss-flink/fluss-flink-common \-Dit.test=RbFunctionsCatalogITCase \-DfailIfNoTests=false` - BUILD SUCCESS

### API and Format

This change does not affect any storage format or wire protocol. The functions operate on existing `BYTES` columns using the standard RoaringBitmap serialization format already used by the server-side `FieldRoaringBitmap32Agg`.

### Documentation

No user-facing documentation yet. Documentation will be added after the scalar functions are implemented and the full Phase 1 function set is available.

### DIFF
--- a/fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/catalog/FlinkCatalog.java
+++ b/fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/catalog/FlinkCatalog.java
@@ -25,6 +25,9 @@ import org.apache.fluss.config.Configuration;
 import org.apache.fluss.exception.InvalidTableException;
 import org.apache.fluss.flink.FlinkConnectorOptions;
 import org.apache.fluss.flink.adapter.CatalogTableAdapter;
+import org.apache.fluss.flink.functions.bitmap.RbAndAggFunction;
+import org.apache.fluss.flink.functions.bitmap.RbBuildAggFunction;
+import org.apache.fluss.flink.functions.bitmap.RbOrAggFunction;
 import org.apache.fluss.flink.lake.LakeFlinkCatalog;
 import org.apache.fluss.flink.procedure.ProcedureManager;
 import org.apache.fluss.flink.utils.CatalogExceptionUtils;
@@ -85,6 +88,7 @@ import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
+import java.util.Locale;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
@@ -140,9 +144,9 @@ public class FlinkCatalog extends AbstractCatalog {
 
     static {
         Map<String, String> map = new HashMap<>();
-        map.put("rb_build_agg", "org.apache.fluss.flink.functions.bitmap.RbBuildAggFunction");
-        map.put("rb_or_agg", "org.apache.fluss.flink.functions.bitmap.RbOrAggFunction");
-        map.put("rb_and_agg", "org.apache.fluss.flink.functions.bitmap.RbAndAggFunction");
+        map.put("rb_build_agg", RbBuildAggFunction.class.getName());
+        map.put("rb_or_agg", RbOrAggFunction.class.getName());
+        map.put("rb_and_agg", RbAndAggFunction.class.getName());
         BUILTIN_BITMAP_FUNCTIONS = Collections.unmodifiableMap(map);
     }
 
@@ -766,15 +770,14 @@ public class FlinkCatalog extends AbstractCatalog {
     @Override
     public boolean functionExists(ObjectPath objectPath) throws CatalogException {
         return BUILTIN_BITMAP_FUNCTIONS.containsKey(
-                objectPath.getObjectName().toLowerCase(java.util.Locale.ROOT));
+                objectPath.getObjectName().toLowerCase(Locale.ROOT));
     }
 
     @Override
     public CatalogFunction getFunction(ObjectPath functionPath)
             throws FunctionNotExistException, CatalogException {
         String className =
-                BUILTIN_BITMAP_FUNCTIONS.get(
-                        functionPath.getObjectName().toLowerCase(java.util.Locale.ROOT));
+                BUILTIN_BITMAP_FUNCTIONS.get(functionPath.getObjectName().toLowerCase(Locale.ROOT));
         if (className == null) {
             throw new FunctionNotExistException(getName(), functionPath);
         }
--- a/fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/catalog/FlinkCatalog.java
+++ b/fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/catalog/FlinkCatalog.java
@@ -48,11 +48,13 @@ import org.apache.flink.table.catalog.CatalogBaseTable;
 import org.apache.flink.table.catalog.CatalogDatabase;
 import org.apache.flink.table.catalog.CatalogDatabaseImpl;
 import org.apache.flink.table.catalog.CatalogFunction;
+import org.apache.flink.table.catalog.CatalogFunctionImpl;
 import org.apache.flink.table.catalog.CatalogMaterializedTable;
 import org.apache.flink.table.catalog.CatalogPartition;
 import org.apache.flink.table.catalog.CatalogPartitionSpec;
 import org.apache.flink.table.catalog.CatalogTable;
 import org.apache.flink.table.catalog.CatalogView;
+import org.apache.flink.table.catalog.FunctionLanguage;
 import org.apache.flink.table.catalog.ObjectPath;
 import org.apache.flink.table.catalog.ResolvedCatalogBaseTable;
 import org.apache.flink.table.catalog.ResolvedCatalogMaterializedTable;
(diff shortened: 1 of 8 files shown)


pbanakar:
Hi @polyzos @wuchong could you please help review this?

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/catalog/FlinkCatalog.java
```diff
@@ -746,19 +760,24 @@ public void alterPartition(
...
+    public boolean functionExists(ObjectPath objectPath) throws CatalogException {
+        return BUILTIN_BITMAP_FUNCTIONS.containsKey(objectPath.getObjectName().toLowerCase());
     }
 
     @Override
-    public boolean functionExists(ObjectPath objectPath) throws CatalogException {
-        return false;
+    public CatalogFunction getFunction(ObjectPath functionPath)
+            throws FunctionNotExistException, CatalogException {
+        String className = BUILTIN_BITMAP_FUNCTIONS.get(functionPath.getObjectName().toLowerCase());
```
polyzos: uses the default locale.. `maybe use toLowerCase(Locale.ROOT)`?

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/functions/bitmap/RbAndAggFunction.java
```diff
@@ -0,0 +1,305 @@
...
+ *
+ * <p>Note: there is no server-side {@code FieldRoaringBitmapAndAgg} counterpart. This function
+ * executes entirely in Flink. Users should be aware that combining it with {@code
+ * table.merge-engine=aggregation} may produce unexpected results during server-side compaction.
+ *
+ * <p>Returns {@code null} when all inputs are null or the input set is empty.
+ */
+@FunctionHint(
+        accumulator = @DataTypeHint(value = "RAW", bridgedTo = RbAndAggFunction.Accumulator.class))
+public class RbAndAggFunction extends AggregateFunction<byte[], RbAndAggFunction.Accumulator> {
```
polyzos, line 60: RbAndAggFunction has no retract; the other two throw a clear UnsupportedOperationException. In a non-append-only GROUP BY, Flink will fail with a less obvious message. Consider adding an explicit throwing retract for symmetry and clearer errors.

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/catalog/FlinkCatalog.java
```diff
@@ -134,6 +136,18 @@ public class FlinkCatalog extends AbstractCatalog {
     protected Connection connection;
     protected Admin admin;
 
+    private static final Map<String, String> BUILTIN_BITMAP_FUNCTIONS;
```
polyzos, line 143: could be wrapped in `Collections.unmodifiableMap` for safety.

polyzos:
Thanks for the clean, well-documented PR @Prajwal-banakar.
The code is in good shape and the rationale comments make it easy to follow. 

I have a few suggestions before this is ready to merge, mostly one important test gap and one design question.

### 1. Add an end-to-end SQL ITCase (most important)

The current tests call `accumulate()`/`getValue()` directly, which bypasses the Flink planner. The riskiest part of this feature is the `@FunctionHint(accumulator = @DataTypeHint("RAW"))` type inference and the custom accumulator serializer flowing through real codegen, **plus** the new catalog registration resolving the function isn't exercised yet. RAW-accumulator UDFs commonly pass unit tests but break at planner time. Could we add an ITCase (the existing `AbstractRbAggFunctionITCase` from #3319 is a natural home) that runs something like:

```sql
USE CATALOG fluss_catalog;
SELECT rb_build_agg(id), rb_or_agg(b), rb_and_agg(b) FROM t GROUP BY k;
```

through a `TableEnvironment`? That would directly validate the "register via FlussCatalog" claim in the PR title.

### 2. Built-in registrationion
I think hardcoding a `name -> className` map inside the general-purpose `FlinkCatalog` means the functions get reported under *every* database via `listFunctions(dbName)`. Since we only have these functions i think its safe and this list won't keep growing, however I would like more people's thoughts just in case.. @platinumhamburg @wuchong WDYT?

I left a few more minor comments/

Once the end-to-end ITCase is added (1) and we align on the registration approach (2), this should be a quick approve. Thanks again!

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/catalog/FlinkCatalog.java
pbanakar: Done, using Locale.ROOT

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/functions/bitmap/RbAndAggFunction.java
pbanakar: Added explicit retract() throwing UnsupportedOperationException for symmetry

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/catalog/FlinkCatalog.java
pbanakar: Wrapped in Collections.unmodifiableMap

pbanakar:
Hi @polyzos Thanks for review and rememberring this RbFunctionsCatalogITCase as important! initially i plan to add all the phase 1 functions and after that add `RbFunctionsCatalogITCase` howevever, 

1) Added RbFunctionsCatalogITCase with four tests exercising all three functions through the catalog registration path using a real TableEnvironment. Happy to extend this further.

2) On the registration design question: I will wait for @platinumhamburg  and @wuchong  input on this before making changes.

platinumhamburg:
> polyzos

Thanks to @Prajwal-banakar for the contribution and to polyzos for the thorough review. Regarding the built-in registration issue, I agree with polyzos’s view: given that we do not expect a large number of new functions to be added in the foreseeable future, and since these functions do not require authentication or categorized management, the current implementation is acceptable.

> Hi @polyzos Thanks for review and rememberring this RbFunctionsCatalogITCase as important! initially i plan to add all the phase 1 functions and after that add `RbFunctionsCatalogITCase` howevever,
> 
> 1. Added RbFunctionsCatalogITCase with four tests exercising all three functions through the catalog registration path using a real TableEnvironment. Happy to extend this further.
> 2. On the registration design question: I will wait for @platinumhamburg  and @wuchong  input on this before making changes.

Thanks to @Prajwal-banakar for the contribution and to @polyzos for the thorough review. Regarding the built-in registration issue, I agree with @polyzos’s view: given that we do not expect a large number of new functions to be added in the foreseeable future, and since these functions do not require authentication or categorized management, the current implementation is acceptable.

polyzos:
@Prajwal-banakar I pushed some small improvements.. Let me know your thoughts and i can merge once CI passes.
Kudos again and thank you for the great work!

pbanakar:
Thanks @polyzos, it looks good to me and I’m happy with the change. It’s been a great learning experience for me, and I’m really grateful to have your guidance!

PR #3400, opened 2026-05-29
[server] Add Cluster Health API implementation

### LINKED ISSUE #3399: [server] Support Cluster Health API for safe rolling upgrades
### Search before asking

### Motivation

During Kubernetes StatefulSet rolling upgrades, the next `TabletServer` pod should not restart until all replicas from the previously restarted pod have fully recovered (leaders re-elected, ISR restored). Without this, cascading restarts can cause data unavailability or prolonged under-replication.

Currently there is no server-side API to determine whether the cluster has finished recovery. Operators rely on TCP-only readiness probes, which pass as soon as the process binds its port — long before replica recovery completes.

### Solution

  Add a `GetClusterHealth` RPC to the Coordinator that computes cluster health from in-memory
  state (CoordinatorContext). The API returns replica statistics and an overall health status:

  - **GREEN** — all replicas are in-sync and all leaders are active.
  - **YELLOW** — all leaders are active, but some replicas have not yet rejoined ISR.
  - **RED** — one or more leaders have not been confirmed active (election or KV recovery in progress).
  - **UNKNOWN** — health could not be determined.

  A readiness-probe shell script (`readiness-check.sh`) performs a two-step check:
  1. TCP port check (local liveness)
  2. Cluster Health API query (only pass on GREEN)

  This gates StatefulSet rolling upgrades: the next pod only restarts when the cluster is fully healthy.

### Anything else?

_No response_

### Willingness to contribute

swuferhong:
### Purpose

Linked issue: close #3399 

  - Add `GetClusterHealth` RPC to Coordinator that computes cluster health from in-memory state
  - Track inactive leaders in `CoordinatorContext` (marked inactive on `NotifyLeaderAndIsr` send,
    marked active on successful response when responding server is still the leader)
  - Handle send failures in `CoordinatorRequestBatch` by synthesizing error responses to clear
    pending inactive state
  - Add client API `Admin.getClusterHealth()` with `ClusterHealth` / `ClusterHealthStatus` types
  - Add `ClusterHealthReadinessCheck` CLI tool in fluss-dist (exit 0=GREEN, 1=not ready, 2=API unsupported)
  - Add `readiness-check.sh` two-step readiness probe script (TCP + Cluster Health API)
    with first-boot detection and grace period for API-unsupported (mixed-version rolling upgrade)
  - Wire tablet-server readiness probe to `readiness-check.sh` in Helm chart
  - Add documentation for Helm deployment and upgrade guide


fluss-server/src/main/java/org/apache/fluss/server/coordinator/CoordinatorContext.java
```diff
@@ -106,6 +106,13 @@ public class CoordinatorContext {
      */
     private final Map<Integer, Set<TableBucket>> replicasOnOffline = new HashMap<>();
 
+    /**
+     * Tracks leader buckets that are currently inactive — a leader is inactive from the moment we
+     * send NotifyLeaderAndIsr until the target server responds successfully confirming it is the
+     * leader. Also inactive when leader == NO_LEADER.
+     */
+    private final Set<TableBucket> inactiveLeaderBuckets = new HashSet<>();
```
loserwang1024: Naming: inactiveLeaderBuckets → pendingLeaderActivationBuckets

The name over-claims. The set is only populated in sendNotifyLeaderAndIsrRequest — buckets where we've sent a NotifyLeaderAndIsr and await leadership confirmation. Buckets leaderless for other reasons (server offline, rebalance with no new leader yet) aren't in it, though their leader is effectively inactive. So it really means "leader change dispatched, awaiting confirmation" — pendingLeaderActivationBuckets fits. Avoid pendingIsr...: this has nothing to do with ISR (that's a separate dimension driving YELLOW).

Bigger issue: "active leader" is defined twice. computeClusterHealth checks leader != NO_LEADER && liveServers.contains(leader) && !inSet, but the helper only checks the set:

```java
public boolean isLeaderActive(TableBucket bucket) {
    return !inactiveLeaderBuckets.contains(bucket); // ignores liveServers / NO_LEADER
}
```

So for a leader on a dead server, isLeaderActive() says true while computeClusterHealth() says RED — inconsistent, and a trap for future reuse.

Suggestion: rename the set + make isLeaderActive the single source of truth:

```java
public boolean isLeaderActive(TableBucket tb) {
    return getBucketLeaderAndIsr(tb)
        .map(lai -> lai.leader() != LeaderAndIsr.NO_LEADER
                && liveTabletServers.containsKey(lai.leader())
                && !pendingLeaderActivationBuckets.contains(tb))
        .orElse(false);
}
```
Then computeClusterHealth just calls ctx.isLeaderActive(tb).

fluss-server/src/main/java/org/apache/fluss/server/coordinator/CoordinatorService.java
```diff
@@ -291,6 +294,62 @@ public CoordinatorService(
...
+                        && ctx.getLiveTabletServers().containsKey(lai.leader())
+                        && !inactiveLeaders.contains(tb)) {
+                    activeLeaderReplicas++;
+                }
+            }
+        }
+
+        // PbClusterHealthStatus: GREEN=0, YELLOW=1, RED=2, UNKNOWN=3
+        int status;
+        if (numLeaderReplicas == 0) {
```
loserwang1024: This can merge with else {
            status = 0; // GREEN
        }

fluss-server/src/main/java/org/apache/fluss/server/RpcServiceBase.java
```diff
@@ -574,6 +576,18 @@ public CompletableFuture<DescribeClusterConfigsResponse> describeClusterConfigs(
                 new DescribeClusterConfigsResponse().addAllConfigs(toPbConfigEntries(configs)));
     }
 
+    @Override
+    public CompletableFuture<GetClusterHealthResponse> getClusterHealth(
```
loserwang1024: Why not move to CoordinatorGateway? Could tablet server support it?

fluss-server/src/main/java/org/apache/fluss/server/coordinator/CoordinatorEventProcessor.java
```diff
@@ -974,10 +975,17 @@ private void processNotifyLeaderAndIsrResponseReceivedEvent(
...
+            }
+        }
+        for (TableBucket tb : succeededBuckets) {
+            Optional<LeaderAndIsr> laiOpt = coordinatorContext.getBucketLeaderAndIsr(tb);
+            if (laiOpt.isPresent() && laiOpt.get().leader() == serverId) {
+                coordinatorContext.markLeaderActive(tb);
             }
         }
         if (!offlineReplicas.isEmpty()) {
-            // trigger replicas to offline
```
loserwang1024, line 1019: why remove this comment?

fluss-server/src/main/java/org/apache/fluss/server/coordinator/CoordinatorRequestBatch.java
```diff
@@ -422,12 +433,20 @@ private void sendNotifyLeaderAndIsrRequest(int coordinatorEpoch) {
...
-                            // the request, and retry if encounter any error; It may happens that
-                            // the tablet server is offline and will always got error. But,
-                            // coordinator will remove the sender for the tablet server and mark all
-                            // replica in the tablet server as offline. so, in here, if encounter
-                            // any error, we just ignore it.
+                            // Treat all buckets as failed — clears pending state and triggers
+                            // re-election via onReplicaBecomeOffline.
+                            List<NotifyLeaderAndIsrResultForBucket> failedResults =
+                                    new ArrayList<>();
+                            ApiError sendError =
```
loserwang1024: Currently, after this modification, even a network timeout exception, the CoordinatorEventProcessor#processNotifyLeaderAndIsrResponseReceivedEvent will mark the server as offlineReplicas 

won't is any problem?
Image

helm/templates/sts-coordinator.yaml
```diff
@@ -122,7 +124,7 @@ spec:
               echo "bind.listeners: ${BIND_LISTENERS}" >> $FLUSS_HOME/conf/server.yaml && \
               echo "advertised.listeners: ${ADVERTISED_LISTENERS}" >> $FLUSS_HOME/conf/server.yaml && \
 
-              bin/coordinator-server.sh start-foreground
+              exec bin/coordinator-server.sh start-foreground
```
loserwang1024, line 127: why need to change it?

loserwang1024 (commented):
I have left some comment.

helm/templates/sts-coordinator.yaml
swuferhong: Without exec, the Fluss process is not PID 1 in the pod, so graceful shutdown signals won't reach it. I fixed this as well.

fluss-server/src/main/java/org/apache/fluss/server/coordinator/CoordinatorRequestBatch.java
swuferhong: Good catch. I have reverted it — the original "just ignore" approach was intentional (the coordinator detects actual server death via heartbeat timeout). The only thing I kept is clearing the pendingLeaderActivationBuckets on send failure, so the health API doesn't report stale `RED` for buckets that are simply stuck in the pending set. Actual failover is still handled by the existing heartbeat/offline mechanism.

swuferhong:
@loserwang1024 comments addressed. PLTA.

loserwang1024 (approved):
LGTM

fluss-dist/src/main/resources/bin/readiness-check.sh
```diff
@@ -0,0 +1,249 @@
...
+            exit 0
+        fi
+        log_to_main "[readiness-check] API unsupported, waiting (grace period)"
+        exit 1
+        ;;
+    *)
+        log_to_main "[readiness-check] Config error (exit $local_exit), treating as not ready"
+        exit 1
+        ;;
+esac
```
wuchong: Suggest removing the `READINESS_GRACE_SECS` grace-period logic entirely (this `mark_api_unsupported` / `is_grace_period_elapsed` machinery and the `--default 60` plumbing).

The grace period is meant to handle the mixed-version window during a rolling upgrade — "wait a bit, maybe the Coordinator finishes upgrading and starts supporting the API." But it doesn't actually help, and it has a real cost:

- On the **first upgrade from a version that predates this API**, the whole cluster is on the old Coordinator, so the API returns `exit 2` (unsupported) for the *entire* upgrade. Every tablet-server pod is then forced to sit at NotReady for the full `GRACE_SECS` (default 60s) before falling back to TCP-only. Since the StatefulSet rolls serially, that's `60s × pod count` of pure dead time added to the upgrade, with zero protection gained (the API genuinely cannot be available in this window).
- The case the grace period is *trying* to protect — "Coordinator temporarily unreachable / still electing" — should already be `exit 1` (not ready, retry next cycle), not `exit 2`. A clean `UnsupportedVersionException` means the server is up and definitively does not support the API, so there is nothing to wait for: we should TCP-fallback immediately.

So `exit 2` can just latch the fast path right away, same as the config-error `*` branch below it. This removes the grace-period env var, the two marker helpers, and the first-upgrade slowdown.

fluss-rpc/src/main/proto/FlussApi.proto
```diff
@@ -1369,4 +1369,14 @@ message PbLiteralValue {
...
+}
+
+message GetClusterHealthRequest { }
+
+message GetClusterHealthResponse {
+  required int32 num_replicas = 1;
+  required int32 in_sync_replicas = 2;
+  required int32 num_leader_replicas = 3;
+  required int32 active_leader_replicas = 4;
+  required int32 status = 5; // PbClusterHealthStatus: GREEN=0, YELLOW=1, RED=2, UNKNOWN=3
```
wuchong: nit: move the RPC messages to before the line of 
```
// --------------- Inner classes ----------------
```

helm/templates/sts-tablet.yaml
```diff
@@ -128,12 +130,21 @@ spec:
...
+            periodSeconds: {{ .Values.tablet.readinessProbe.periodSeconds | default 5 }}
+            exec:
+              command:
+                - /bin/bash
+                - -c
+                - |
+                  export FLUSS_SERVER_ID=${POD_NAME##*-}
+                  export READINESS_TIMEOUT_MS={{ .Values.tablet.readinessProbe.rpcTimeoutMs | default 5000 }}
+                  export READINESS_TCP_PORT={{ .Values.listeners.client.port }}
+                  export READINESS_BOOTSTRAP_SERVERS="coordinator-server-hs.${POD_NAMESPACE}.svc.cluster.local:{{ .Values.listeners.client.port }}"
```
wuchong: Will the coordinator address change in scenarios such as upgrades, restarts, or failover migrations? If it does, could this prevent the tablet server from connecting to the coordinator during its next restart or upgrade?

wuchong (commented):
The readiness probe currently adds extra dist weight by pulling in `fluss-client-${version}.jar` just to query cluster health. Since this probe only needs one internal Coordinator RPC, it should not depend on the full user-facing client jar.

Suggested lightweight direction:

- Move `ClusterHealthReadinessCheck` into `fluss-server`.
- Run it from the existing `fluss-server-${version}.jar`; then `fluss-dist-${version}.jar` is no longer needed for the probe either.
- Replace `READINESS_BOOTSTRAP_SERVERS` with `COORDINATOR_ADDRESS`, since the probe explicitly connects to a Coordinator address.
- Do not issue a `MetadataRequest` first. The probe already has the Coordinator address and can directly call `getClusterHealth`.
- Remove auth parsing/config from both `readiness-check.sh` and `ClusterHealthReadinessCheck`; the probe can use the internal channel rather than the external client/auth path.
- Remove both `fluss-client-${version}.jar` and `fluss-dist-${version}.jar` from the binary dist assembly for this feature.

This keeps the rolling-upgrade health gate but avoids adding ~66MB of client jar plus the extra dist probe jar to the binary distribution.

helm/templates/sts-tablet.yaml
swuferhong: Switched the readiness probe's bootstrap server to 127.0.0.1 so it no longer depends on the coordinator's address.

fluss-server/src/main/java/org/apache/fluss/server/coordinator/CoordinatorRequestBatch.java
```diff
@@ -409,10 +409,19 @@ private void sendNotifyLeaderAndIsrRequest(int coordinatorEpoch) {
...
             Integer serverId = notifyRequestEntry.getKey();
+            Set<TableBucket> buckets = notifyRequestEntry.getValue().keySet();
             NotifyLeaderAndIsrRequest notifyLeaderAndIsrRequest =
                     makeNotifyLeaderAndIsrRequest(
                             coordinatorEpoch, notifyRequestEntry.getValue().values());
 
+            for (Map.Entry<TableBucket, PbNotifyLeaderAndIsrReqForBucket> entry :
+                    notifyRequestEntry.getValue().entrySet()) {
+                int leader = entry.getValue().getLeader();
+                if (leader == serverId || leader == LeaderAndIsr.NO_LEADER) {
```
wuchong: Why should the NO_LEADER state be included in the pending leader set? Once a NO_LEADER bucket is added to this set, it is never cleared upon receiving the NotifyLeaderAndIsrResponse. Since we already treat NO_LEADER buckets as inactive leaders in org.apache.fluss.server.coordinator.CoordinatorContext#isLeaderActive, there is no need to handle the NotifyLeaderAndIsr path for NO_LEADER buckets.

fluss-server/src/main/java/org/apache/fluss/server/coordinator/CoordinatorContext.java
```diff
@@ -106,6 +106,13 @@ public class CoordinatorContext {
      */
     private final Map<Integer, Set<TableBucket>> replicasOnOffline = new HashMap<>();
 
+    /**
+     * Tracks buckets where a leader change has been dispatched (via NotifyLeaderAndIsr) but not yet
+     * confirmed by the target server. A bucket enters this set when we send the notification and
+     * leaves it when the target server successfully responds confirming it is the leader.
+     */
+    private final Set<TableBucket> pendingLeaderActivationBuckets = new HashSet<>();
```
wuchong, line 114: Suggest adding a gauge metric for `pendingLeaderActivationBuckets`.

This set is the core signal behind the RED status of the health API: `isLeaderActive` returns false for any bucket in it, which makes the health API report RED and blocks the rolling upgrade at the readiness gate. If a leader bucket's activation never gets cleared (e.g. due to an exception or a missed clear), the upgrade hangs at NotReady indefinitely, while operators only see a pod stuck NotReady with no way to tell which bucket is stuck or for how long — they'd have to dig through coordinator logs, which is expensive to diagnose.

Suggest registering a gauge in `CoordinatorMetricGroup` (e.g. `pendingLeaderActivationCount`) backed by the existing read-only accessor `coordinatorContext.getPendingLeaderActivationBuckets().size()`. This mirrors the existing `gauge(MetricNames.BUCKET_COUNT, buckets::size)` in the same file and is consistent in spirit with `offlineBucketCount` / `replicasToDeleteCount`. A value that stays non-zero over time then becomes directly alertable/observable on dashboards, turning "upgrade hangs" from a hard-to-debug mystery into a visible curve.

fluss-dist/pom.xml
```diff
@@ -263,8 +263,24 @@
                             <finalName>${project.artifactId}-${project.version}</finalName>
                             <artifactSet>
                                 <includes>
-                                    <!-- only include log4j-api to package source and javadoc to make sonatype happy -->
-                                    <include>org.apache.logging.log4j:log4j-api</include>
+                                    <!--
+                                        IMPORTANT: do NOT include log4j-api here.
```
wuchong: we should revert this, otherwise, release will fail on sonatype.

fluss-server/src/main/java/org/apache/fluss/server/coordinator/CoordinatorRequestBatch.java
```diff
@@ -428,6 +437,13 @@ private void sendNotifyLeaderAndIsrRequest(int coordinatorEpoch) {
                             // coordinator will remove the sender for the tablet server and mark all
                             // replica in the tablet server as offline. so, in here, if encounter
                             // any error, we just ignore it.
+
+                            // Clear pending state so the health API does not report stale
+                            // RED. The coordinator will detect actual server death via
+                            // heartbeat timeout and trigger re-election separately.
+                            for (TableBucket tb : buckets) {
+                                coordinatorContext.clearPendingLeaderActivation(tb);
```
wuchong: When a NotifyLeaderAndIsr request to a follower fails, buckets still contains the bucket whose pending activation was added for the actual leader. Clearing every bucket here can remove that leader's pending state before the leader has acknowledged the request, allowing getClusterHealth() to report GREEN during a rolling upgrade while the new leader is not confirmed active.

wuchong: Besides, `coordinatorContext.clearPendingLeaderActivation(...)` is called inside the NotifyLeaderAndIsr send-failure callback, which runs on the RPC completion thread (whenComplete is attached with no executor), not the single coordinator event thread. CoordinatorContext is @NotThreadSafe and pendingLeaderActivationBuckets is a plain HashSet that is concurrently read/mutated on the event thread (e.g. addPendingLeaderActivation, isLeaderActive, computeClusterHealth via AccessContextEvent). This is an unsynchronized data race. The success branch right below correctly hands off via eventManager.put(...) instead of touching the context directly — the failure branch should do the same.

swuferhong:
@wuchong comments addressed.

wuchong (approved):

PR #3402, opened 2026-05-29
[flink] Fail fast on batch virtual log table reads

### LINKED ISSUE #3144: Batch reads on ``/`` fail with coordinator NPE instead of a clear unsupported error
### Search before asking

### Fluss version

main (development)

### Please describe the bug 🐞

Batch reads on `$binlog` and `$changelog` virtual tables do not fail fast. Instead, the job is submitted and later fails in the source coordinator with a `NullPointerException`.

This is especially confusing because the failure happens far away from SQL validation and does not explain that virtual log tables are currently stream-oriented.

Reproduction:

1. Create a primary key table with data lake enabled.
2. Run a batch query such as:

```sql
SELECT * FROM `my_table$binlog`;
```

The same problem should also apply to `$changelog`.

Actual behavior:

- The batch job starts.
- `FlinkSourceEnumerator.startInBatchMode()` goes into the hybrid lake split path.
- For virtual tables, the source is created without a `lakeSource`, so `LakeSplitGenerator.generateHybridLakeFlussSplits()` dereferences `lakeSource` and throws `NullPointerException`.

Observed stack trace excerpt:

```text
Caused by: org.apache.flink.util.FlinkRuntimeException: Failed to generate hybrid lake fluss splits
    at org.apache.fluss.flink.source.enumerator.FlinkSourceEnumerator.generateHybridLakeFlussSplits(FlinkSourceEnumerator.java:...)
Caused by: java.lang.NullPointerException
    at org.apache.fluss.flink.lake.LakeSplitGenerator.generateHybridLakeFlussSplits(LakeSplitGenerator.java:118)
```

Relevant code paths:

- `BinlogFlinkTableSource` and `ChangelogFlinkTableSource` create `FlinkSource` as log-only virtual tables.
- `FlinkSourceEnumerator.startInBatchMode()` still routes batch execution to hybrid lake split generation when the base table is lake-enabled.
- `LakeSplitGenerator.generateHybridLakeFlussSplits()` assumes `lakeSource` is non-null.

Expected behavior:

- If batch reads on `$binlog`/`$changelog` are not supported, reject them early with a clear `UnsupportedOperationException` mentioning that these virtual tables only support streaming mode.
- If batch reads are intended to be supported, batch mode should use bounded log-only splits for virtual tables instead of entering the lake hybrid path.

Impact:

- User-visible failure is an internal NPE instead of an actionable error.
- The error is delayed until coordinator split generation, which makes diagnosis unnecessarily difficult.

### Solution

At minimum, add fail-fast validation for `$binlog` and `$changelog` in batch mode so the planner/source creation throws a clear unsupported-mode exception.

A more complete fix would be to make batch virtual-table reads use bounded log-only splits and avoid `LakeSplitGenerator` entirely when `lakeSource == null`.

QuakeWang:
### Purpose

Linked issue: close #3144

Batch reads on `$binlog` and `$changelog` virtual tables could pass source creation and fail later in the source coordinator with an internal lake-source error. These virtual log tables are currently log-only and only support streaming mode.

### Brief change log

  - Reject batch runtime mode in `$binlog` and `$changelog` table source creation.
  - Return a clear `UnsupportedOperationException` before job submission.
  - Add factory and planner-path tests for both virtual table types.

### Tests

  - `mvn -B -ntp -Dskip.on.java8=true -DfailIfNoTests=false -pl fluss-flink/fluss-flink-1.20 -am -Dtest=Flink120TableFactoryTest test`
  - `mvn -B -ntp -Dskip.on.java8=true -DfailIfNoTests=false -pl fluss-flink/fluss-flink-1.20 -am -Dtest=Flink120BinlogVirtualTableITCase,Flink120ChangelogVirtualTableITCase test`

### DIFF
--- a/fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/catalog/FlinkTableFactory.java
+++ b/fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/catalog/FlinkTableFactory.java
@@ -322,6 +322,15 @@ public class FlinkTableFactory implements DynamicTableSourceFactory, DynamicTabl
         FlinkConnectorOptionsUtils.validateTableSourceOptions(tableOptions);
     }
 
+    private static void validateVirtualLogTableRuntimeMode(
+            String virtualTableSuffix, boolean isStreamingMode) {
+        if (!isStreamingMode) {
+            throw new UnsupportedOperationException(
+                    String.format(
+                            "%s virtual tables only support streaming mode.", virtualTableSuffix));
+        }
+    }
+
     /** Creates a ChangelogFlinkTableSource for $changelog virtual tables. */
     private DynamicTableSource createChangelogTableSource(
             Context context, ObjectIdentifier tableIdentifier, String tableName) {
@@ -333,6 +342,7 @@ public class FlinkTableFactory implements DynamicTableSourceFactory, DynamicTabl
         boolean isStreamingMode =
                 context.getConfiguration().get(ExecutionOptions.RUNTIME_MODE)
                         == RuntimeExecutionMode.STREAMING;
+        validateVirtualLogTableRuntimeMode(FlinkCatalog.CHANGELOG_TABLE_SUFFIX, isStreamingMode);
 
         // tableOutputType includes metadata columns: [_change_type, _log_offset, _commit_timestamp,
         // data_cols...]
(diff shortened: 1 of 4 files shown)


luoyuxia (approved):
LGTM @loserwang1024 Could you please have another look

loserwang1024 (approved):
LGTM

luoyuxia:
@QuakeWang The ci fail. You can run mvn spotless:apply to fix it

QuakeWang:
> @QuakeWang The ci fail. You can run mvn spotless:apply to fix it

@luoyuxia done

PR #3414, opened 2026-06-01
[lake] Reset lake filters for each pushdown

QuakeWang:
### Purpose

Lake filter pushdown state was kept on mutable lake source instances. If a later pushdown had no convertible lake predicates, the old predicate could remain active and incorrectly filter subsequent lake reads.

### Brief change log

  - Reset Paimon and Iceberg lake filters on every `withFilters` call.
  - Ensure Flink clears lake filters even when no lake predicates are convertible.
  - Recreate and replay lake pushdown state when copying `FlinkTableSource`, avoiding shared mutable lake source state between copies.
  - Add regression tests for Paimon, Iceberg, and Flink lake source copy isolation.

### Tests

- `./mvnw -pl fluss-flink/fluss-flink-common -am -Dtest='FlinkTableSourceFilterPushDownTest$LakeSourcePushDownTests' -DfailIfNoTests=false -Dsurefire.failIfNoSpecifiedTests=false test`
- `./mvnw -pl fluss-flink/fluss-flink-common,fluss-lake/fluss-lake-paimon,fluss-lake/fluss-lake-iceberg -am -Dtest=FlinkTableSourceFilterPushDownTest,PaimonLakeSourceTest,IcebergLakeSourceTest -DfailIfNoTests=false -Dsurefire.failIfNoSpecifiedTests=false test`

### DIFF
--- a/fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/source/FlinkTableSource.java
+++ b/fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/source/FlinkTableSource.java
@@ -154,6 +154,7 @@ public class FlinkTableSource
 
     // projection push down
     @Nullable private int[] projectedFields;
+    @Nullable private int[][] lakeProjectedFields;
 
     @Nullable private GenericRowData singleRowFilter;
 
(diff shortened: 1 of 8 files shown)


QuakeWang:
@luoyuxia PTAL, Thanks : )

Issue #3314, opened 2026-05-14
[Feature] Support Partition Mark Done for Fluss Tiering (Paimon Lake)


beryllw:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/fluss/issues) and found nothing similar.

### Motivation

For partitioned tables with Paimon lake tiering enabled, downstream batch jobs need a signal that a partition's data is "ready" (i.e., fully tiered). This is the **Mark Done** mechanism — when a partition has been idle (no new data tiered) for a configurable duration, Fluss should execute Paimon's mark-done actions (e.g., write `_SUCCESS` file, notify metastore) so downstream schedulers can safely begin processing.

Currently, Fluss tiering operates at the table level with no partition-level idle tracking or mark-done capability. Paimon already has a complete mark-done action framework (`PartitionMarkDoneAction`), but its trigger mechanism is tightly coupled to Flink checkpoint lifecycle, which doesn't fit Fluss's tiering model.

### Solution

**Summary.** Carry table-level tiering state as a single optional `tiering_state` field in the offset file, transported as one `bytes tiering_state_json` in RPC — following Fluss's existing `schema_json` / `table_json` pattern, so adding new state never changes proto/RPC. Partition mark-done is the first consumer, using a **delete-on-done** model that keeps the state tiny and self-GCing.

#### Storage & RPC

- The offset file (`version=1`) gains an optional top-level `tiering_state` (JSON object) next to `bucket_offsets`. Its content is owned by the lake side (`LakeTieringTableState` in `fluss-common`); the offset serde passes it through **without parsing**.
- RPC: `PbTableOffsets` (in `PrepareLakeTableSnapshotRequest`) and `GetLakeSnapshotResponse` each get one new `optional bytes tiering_state_json` (new field number → wire-compatible). The server moves it through as an opaque blob (persist on PREPARE, read back on GET); the client parses it via `LakeSnapshot.getLakeTieringTableState()`.
- Because it is a single `bytes` field, future state fields need no further proto/RPC change.

#### Partition mark-done (delete-on-done)

- **State**: a table-level `partition_done_initialized` flag + `partition_update_times` of only the not-yet-done partitions; no `doneTime` (read Paimon `_SUCCESS` if needed). Once marked done, a partition is removed.

```json
{
  "version": 1,
  "table_id": 8,
  "partition_offsets": [ { "partition_id": 5, "bucket_offsets": [100, 230] } ],
  "tiering_state": {
    "version": 1,
[...]

PR #3420, opened 2026-06-02
[lake/tiering] Support reporting watermark to Paimon snapshot

### LINKED ISSUE #3419: [tiering] Support reporting watermark to Paimon snapshot
### Search before asking

### Motivation

Currently, Fluss lake tiering can write data into Paimon, but the committed Paimon snapshots do not carry the tiering watermark. This makes it harder to understand the event-time progress of data synchronized from Fluss to the lake table.

Paimon snapshot watermark can be used to better track the synchronization progress of lake tables. By reporting the watermark during tiering commits, users and downstream systems can inspect Paimon snapshots to know how far the lake table has caught up in event time.

### Solution

Propagate watermark information through the lake tiering pipeline and report it when committing to Paimon.

This includes extracting watermarks from Fluss records, carrying them for each split, aggregating them during tiering commit, and writing the final watermark into the committed Paimon snapshot.

### Anything else?

_No response_

### Willingness to contribute

Shawn-Hx:
### Purpose

Linked issue: close #3419 

This PR supports reporting watermarks to Paimon snapshots during lake tiering.

### Brief change log

- Introduced watermark extraction for Fluss rows in the lake tiering pipeline.
- Added watermark propagation through lake write results and tiering commit.
- Aggregated watermarks across tiered buckets before committing to the lake.
- Reported the aggregated watermark to Paimon snapshots when committing.
- Updated serializers and tests to support nullable and negative watermark values.
- Added coverage for watermark extraction, commit aggregation, and Paimon snapshot watermark reporting.

### Tests

Add tests in:
- PaimonWriteResultSerializerTest
- PaimonTieringTest
- TieringCommitOperatorTest
- SimpleWatermarkExtractorTest

### API and Format

This change extends the lake tiering API to carry watermark information through write results and committers.

### Documentation

No user-facing documentation needs to be added.


luoyuxia:
Thanks for working on this.

I have one high-level question about the motivation/use case: if Fluss implements https://github.com/apache/fluss/issues/3314, would that satisfy the requirement this PR is trying to address?

My understanding is that #3314 would provide a partition-level "mark done" / readiness signal for Paimon lake tiering, while this PR reports an event-time watermark into Paimon snapshots. If the main goal is to let downstream batch jobs know when a partition is ready to process, #3314 may be the more direct mechanism and this PR might not be necessary. But if the requirement is specifically to expose event-time progress through Paimon snapshot watermarks, then this PR still seems to cover a different need.

Could you clarify the intended downstream use case here?

Shawn-Hx:
> Thanks for working on this.
> 
> I have one high-level question about the motivation/use case: if Fluss implements #3314, would that satisfy the requirement this PR is trying to address?
> 
> My understanding is that #3314 would provide a partition-level "mark done" / readiness signal for Paimon lake tiering, while this PR reports an event-time watermark into Paimon snapshots. If the main goal is to let downstream batch jobs know when a partition is ready to process, #3314 may be the more direct mechanism and this PR might not be necessary. But if the requirement is specifically to expose event-time progress through Paimon snapshot watermarks, then this PR still seems to cover a different need.
> 
> Could you clarify the intended downstream use case here?

Yes, the main purpose of this PR is to determine whether a partition is done based on the watermark. Paimon’s PartitionMarkDone mechanism supports the watermark-based mode. I’m not sure whether the solution proposed in #3314 takes the watermark into consideration. If we need to determine whether a partition is done based on the watermark, the tiering job must be able to get the watermark. So I think this PR is a prerequisite for #3314.

luoyuxia:
> > Thanks for working on this.
> > I have one high-level question about the motivation/use case: if Fluss implements #3314, would that satisfy the requirement this PR is trying to address?
> > My understanding is that #3314 would provide a partition-level "mark done" / readiness signal for Paimon lake tiering, while this PR reports an event-time watermark into Paimon snapshots. If the main goal is to let downstream batch jobs know when a partition is ready to process, #3314 may be the more direct mechanism and this PR might not be necessary. But if the requirement is specifically to expose event-time progress through Paimon snapshot watermarks, then this PR still seems to cover a different need.
> > Could you clarify the intended downstream use case here?
> 
> Yes, the main purpose of this PR is to determine whether a partition is done based on the watermark. Paimon’s PartitionMarkDone mechanism supports the watermark-based mode. I’m not sure whether the solution proposed in #3314 takes the watermark into consideration. If we need to determine whether a partition is done based on the watermark, the tiering job must be able to get the watermark. So I think this PR is a prerequisite for #3314.

Thanks. The issue is expected to take the watermark into consideration. And I agree on the pr or some similar thing is a prerequisite for #3314. But I think may be we need to limit the scope of the pr to avoid over implementation since seems write watermark is not needed.  @beryllw Could you please have to look to determine which part for watermark is needed in you design for #3314

Shawn-Hx:
Hi @beryllw , could you please take a look? I'm willing to commit if there are any modifications needed.

fluss-flink/fluss-flink-common/src/test/java/org/apache/fluss/flink/tiering/source/watermark/SimpleWatermarkExtractorTest.java
```diff
@@ -0,0 +1,371 @@
...
+        TableInfo tableInfo = createTableInfoWithWatermark("ts", "TIMESTAMP(3)");
+        SimpleWatermarkExtractor extractor = SimpleWatermarkExtractor.create(tableInfo);
+        assertThat(extractor).isNotNull();
+
+        GenericRow row = GenericRow.of(1, TimestampNtz.fromMillis(2000L));
+        assertThat(extractor.currentWatermark(row)).isEqualTo(2000L);
+    }
+
+    @Test
+    void testCreateWithColumnMinusInterval() {
```
beryllw: Nice and thorough coverage here 👍. Just a small non-blocking suggestion: testCreateWithColumnMinusInterval / Decimal / Minute / Hour / Day share the exact same structure and differ only in the expression and expected value. Following the existing testCreateWithTimestampPrecision in this file, they could be folded into a single @ParameterizedTest with @CsvSource — that way adding new cases (e.g. lowercase units or non-backtick columns) is just one more row, and the intent reads more clearly. The Arrow-batch tests are fine to keep separate given the variable rows/nulls. 

```
@ParameterizedTest
@CsvSource({
    // expr, inputMillis, expectedWatermarkMillis
    "`ts`,                           1000,      1000",
    "ts,                             2000,      2000",
    "`ts` - INTERVAL '5' SECOND,     10000,     5000",
    "`ts` - INTERVAL '0.001' SECOND, 1000,      999",
    "`ts` - INTERVAL '2' MINUTE,     200000,    80000",
    "`ts` - INTERVAL '1' HOUR,       7200000,   3600000",
    "`ts` - INTERVAL '1' DAY,        172800000, 86400000",
})
void testNtzWatermarkExtraction(String expr, long inputMillis, long expectedMillis) {
```

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/tiering/committer/TieringCommitOperator.java
```diff
@@ -232,14 +233,35 @@ private CommitResult commitWriteResults(
...
+            Long watermark = null;
             for (TableBucketWriteResult<WriteResult> writeResult : nonEmptyResults) {
                 TableBucket tableBucket = writeResult.tableBucket();
                 logEndOffsets.put(tableBucket, writeResult.logEndOffset());
                 logMaxTieredTimestamps.put(tableBucket, writeResult.maxTimestamp());
+                Long writeResultWatermark = writeResult.writeResult().getWatermark();
+                if (writeResultWatermark == null) {
+                    continue;
+                }
+                watermark =
```
beryllw: nit: Consider adding a brief comment explaining why it's safe for the aggregated watermark to be lower than previously committed — Paimon's FileStoreCommitImpl#tryCommitOnce guarantees snapshot watermark monotonicity via Math.max(current, latestSnapshot.watermark()), so it will never regress. This would save future readers from needing to trace into Paimon source to verify correctness.
https://github.com/apache/paimon/blob/5b797ef63fba5e98bf4fd285fb5a34877c91ee81/paimon-core/src/main/java/org/apache/paimon/operation/FileStoreCommitImpl.java#L1020-L1026

fluss-lake/fluss-lake-paimon/src/main/java/org/apache/fluss/lake/paimon/tiering/PaimonWriteResultSerializer.java
```diff
@@ -38,21 +42,58 @@ public int getVersion() {
...
+
+                    Long watermark = null;
+                    boolean watermarkNonNull = view.readBoolean();
+                    if (watermarkNonNull) {
+                        watermark = view.readLong();
+                    }
+
+                    int len = view.readInt();
+                    byte[] messageBytes = new byte[len];
+                    view.read(messageBytes);
```
beryllw: Might be slightly safer to use view.readFully(messageBytes) here — read(byte[]) doesn't guarantee filling the whole array (and its return value is unused). Not an issue today since messageBytes is the last field, just a nit for consistency with TableBucketWriteResultSerializer.

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/tiering/source/watermark/SimpleWatermarkExtractor.java
```diff
@@ -0,0 +1,218 @@
...
+        Map<String, String> props = tableInfo.getCustomProperties().toMap();
+        RowType rowType = tableInfo.getRowType();
+
+        boolean isWatermarkDefined = false;
+        String rowtimeColumn = null;
+        String watermarkIndex = null;
+
+        for (Map.Entry<String, String> entry : props.entrySet()) {
+            String key = entry.getKey();
+            if (key.startsWith(WATERMARK_PROPERTY_PREFIX)
```
beryllw, line 109:
```
long count = props.keySet().stream()
    .filter(k -> k.startsWith(WATERMARK_PROPERTY_PREFIX) 
              && k.endsWith(WATERMARK_ROWTIME_SUFFIX))
    .count();
if (count > 1) {
    LOG.warn("More than 1 watermark definition...");
    return null;
}
```

beryllw:
Thanks @Shawn-Hx for driving this. LTGM!
From the #3314  (partition mark-done) side, this PR looks quite helpful as a prerequisite.For mark-done, the parts that seem most relevant are the WatermarkExtractor / SimpleWatermarkExtractor and the writer-side propagation, which let the watermark reach the tiering commit side. I plan to build on the watermark extraction/propagation here for the watermark-based mark-done in #3314. 
Thanks again!

beryllw:
One more thought — watermark extraction uses a shared WatermarkExtractor and applies to all lakes, so it feels more like a tiering-layer concern than a per-lake writer one. Since TieringSplitReader already builds the extractor and sees every record, could we compute the watermark there and carry it in TableBucketWriteResult (next to maxTimestamp), instead of each LakeWriter doing it via LakeWriteResult? That would keep the logic in one place and avoid pushing the extends LakeWriteResult bound across the lake API.
@luoyuxia  what do you think about moving watermark computation up to the tiering layer?

Shawn-Hx:
> One more thought — watermark extraction uses a shared WatermarkExtractor and applies to all lakes, so it feels more like a tiering-layer concern than a per-lake writer one. Since TieringSplitReader already builds the extractor and sees every record, could we compute the watermark there and carry it in TableBucketWriteResult (next to maxTimestamp), instead of each LakeWriter doing it via LakeWriteResult? That would keep the logic in one place and avoid pushing the extends LakeWriteResult bound across the lake API. @luoyuxia what do you think about moving watermark computation up to the tiering layer?

Thanks @beryllw for the reply. Calculating watermark in LakeWriter for each lake format involves considering that some lake formats do not support reporting watermark information. In such cases, the unnecessary computation during tiering can be skipped.

Shawn-Hx:
Hi @luoyuxia @beryllw, could you take a look at this PR when you have a moment?

PR #3421, opened 2026-06-03
[fs] Mask S3 delegation token access keys in logs

### LINKED ISSUE #1989: [FileSystem] Mask user secret in the S3DelegationTokenProvider

QuakeWang:
### Purpose

Linked issue: close #1989

The S3 delegation token provider and receiver logged full access key IDs when obtaining or updating temporary session credentials. This exposed credential identifiers in logs, while the log only needs enough information to confirm which key path is being used.

### Brief change log

- Add a package-local S3 token log utility to keep a short access-key prefix and mask the rest.
- Mask access key IDs in S3 delegation token provider and receiver logs.
- Add unit tests for null, short, and normal access key values.

### Tests

- `mvn -pl fluss-filesystems/fluss-fs-s3 -am -Dtest=S3TokenLogUtilsTest -DfailIfNoTests=false clean test`

### DIFF
--- a/fluss-filesystems/fluss-fs-s3/src/main/java/org/apache/fluss/fs/s3/token/S3DelegationTokenProvider.java
+++ b/fluss-filesystems/fluss-fs-s3/src/main/java/org/apache/fluss/fs/s3/token/S3DelegationTokenProvider.java
@@ -107,14 +107,14 @@ public class S3DelegationTokenProvider {
             } else {
                 LOG.info(
                         "Obtaining session credentials via GetSessionToken with access key: {}",
-                        accessKey);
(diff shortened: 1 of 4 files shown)


luoyuxia (approved):
Thanks. LGTM

PR #3424, opened 2026-06-03
[docs] add docs to integrate Fluss + Iceberg via Flink with AWS Glue and Hive

### LINKED ISSUE #2616: [Lake/Iceberg] create doc to integrate fluss + Iceberg via flink with aws glue
### Search before asking

### Description

Guideline or steps for the end user who wants to use fluss + iceberg with the Hive catalog

Image

Structure it like below. Integrate Data Lakes, create a heading that contains all catalog types :

- DataLake Formats/ Integrate Data Lakes/Formats
     -  Iceberg
     -  Paimon
     -  Lance

-  DataLake Catalogs
    - Iceberg  
      -  Glue, 
      -  Hive, etc...

Example you can refer to, while creating doc to follow a structure:
https://hudi.apache.org/docs/catalog_polaris

### Willingness to contribute

qzyu999:
### Purpose

Linked issue: close #2616

This pull request introduces comprehensive integration guides for using AWS Glue and Hive Metastore catalogs when tiering Fluss streaming data to Apache Iceberg. This completes the Iceberg Data Lake Catalogs documentation suite under `docs/streaming-lakehouse/integrate-data-lakes/catalogs/`.

### Brief change log

- **Added AWS Glue Catalog integration guide** (`website/docs/streaming-lakehouse/integrate-data-lakes/catalogs/glue.md`): Documents AWS IAM policy template, required catalog runtime JAR dependencies, `server.yaml` cluster configurations, Flink tiering service commands, and Amazon Athena query verification.
- **Added Hive Metastore integration guide** (`website/docs/streaming-lakehouse/integrate-data-lakes/catalogs/hive.md`): Documents Hive Metastore Thrift connection options, required Hadoop client and Hive runtime classpath dependencies, `HADOOP_CLASSPATH` configuration, Flink tiering commands, and Spark SQL query verification.
- **Updated main Iceberg integration guide** (`website/docs/streaming-lakehouse/integrate-data-lakes/formats/iceberg.md`): Added catalog-specific cross-links for `hive` (linking to Hive Metastore), `glue` (linking to AWS Glue), and `rest` (linking to Lakekeeper).

_Note:_ Changes are based on the existing `lakekeeper.md` as a template, and references were based on existing code and online/offline documentation. The actual AWS Glue/HMS implementations have not yet been tested by the developer.

### Tests

- Built the entire documentation site locally using `npm run build` to verify the page output and ensure there are no broken links (meeting Docusaurus build validation).
- Verified the rendering, layouts, and link references using the local development server at `http://localhost:3000/docs/next/streaming-lakehouse/integrate-data-lakes/formats/iceberg/`.

### API and Format

This is a documentation-only change. It does not affect any public API or storage formats.

### Documentation

This pull request introduces new documentation guides under the `Docusaurus` website subfolder. No changes were made to code-level Javadocs.

- [x] **Generative AI disclosure:**
  - [x] Yes (Antigravity AI Assistant, reviewed by human developer)

### DIFF
--- a/website/docs/streaming-lakehouse/datalake-catalogs/glue.md
+++ b/website/docs/streaming-lakehouse/datalake-catalogs/glue.md
@@ -1,6 +1,6 @@
 ---
 title: AWS Glue
-sidebar_position: 2
+sidebar_position: 3
 ---
 
 # AWS Glue
@@ -15,9 +15,9 @@ This guide explains how to configure Fluss to use AWS Glue as its Iceberg catalo
 
 When Fluss is configured with AWS Glue as its Iceberg catalog:
 
-1. **Data ingestion**: Applications write data to Fluss tables using the Fluss client (Java/Python) or Flink SQL. Fluss stores this data in its log tables.
+1. **Data ingestion**: Applications write data to Fluss tables using the Fluss client (Java/Python) or Flink SQL. Fluss stores this data in its tables (log tables or primary key tables).
 2. **Tiering to Iceberg**: A separate Flink job (the tiering service) periodically reads accumulated data from Fluss, converts it to Parquet format, writes the files to S3, and commits an Iceberg snapshot to the Glue Data Catalog.
-3. **Query via Athena/Spark/Trino**: Any Iceberg-compatible engine can discover and query the tiered tables through AWS Glue  no additional configuration needed.
+3. **Query via Athena/Spark/Trino**: Any Iceberg-compatible engine can discover and query the tiered tables through AWS Glue — no additional configuration needed.
 
 ```mermaid
 flowchart TD
@@ -51,15 +51,18 @@ flowchart TD
     D3 -->|commit snapshot| C1
     C1 -.->|discover tables| C3
     C2 -.->|read data| C3
+    B2 -->|remote log & KV snapshots| C2
 ```
 
-> **Note**: Flink is used here for the tiering service (Fluss  Iceberg/S3/Glue) and optionally for data ingestion via SQL. Applications can also write directly to Fluss using the client library (Java, Python). The tiering service is what bridges Fluss's log tables to the Glue Data Catalog.
+> **Note**: Flink is used here for the tiering service (Fluss → Iceberg/S3/Glue) and optionally for data ingestion via SQL. Applications can also write directly to Fluss using the client library (Java, Python). The tiering service is what bridges Fluss tables to the Glue Data Catalog.
 
 ## Prerequisites
 
 ### Java Version
 
-Fluss 0.9.x requires **Java 17 or later** for the Tablet Server. Java 11 will fail with `NoSuchMethodError: MappedByteBuffer.duplicate()`.
+Fluss requires **Java 11 or later**, with Java 17 recommended (see [Deploying Fluss](../../install-deploy/deploying-local-cluster.md)).
+
+> **NOTE**: The 0.9.1 release binaries do not run on Java 11 — the Tablet Server fails with `NoSuchMethodError: MappedByteBuffer.duplicate()` because the release was compiled on a newer JDK. Use Java 17+ with the 0.9.1 binaries, or build Fluss from source on JDK 11.
 
 ### AWS IAM Permissions
 
@@ -119,32 +122,32 @@ Place the following JARs in the `${FLUSS_HOME}/plugins/iceberg/` directory:
 
 - **Iceberg AWS Bundle**: [iceberg-aws-bundle-1.10.1.jar](https://repo1.maven.org/maven2/org/apache/iceberg/iceberg-aws-bundle/1.10.1/iceberg-aws-bundle-1.10.1.jar)
 - **Iceberg AWS**: [iceberg-aws-1.10.1.jar](https://repo1.maven.org/maven2/org/apache/iceberg/iceberg-aws/1.10.1/iceberg-aws-1.10.1.jar)
-- **Failsafe**: [failsafe-3.3.2.jar](https://repo1.maven.org/maven2/dev/failsafe/failsafe/3.3.2/failsafe-3.3.2.jar)  required by `iceberg-aws` for Glue API retry logic
+- **Failsafe**: [failsafe-3.3.2.jar](https://repo1.maven.org/maven2/dev/failsafe/failsafe/3.3.2/failsafe-3.3.2.jar) — required by `iceberg-aws` for Glue API retry logic
 
 Both Iceberg JARs are required. The bundle provides AWS SDK v2 dependencies, while `iceberg-aws` provides the `GlueCatalog` class that Iceberg loads via reflection. The plugin classloader cannot find `GlueCatalog` from the bundle alone. The `failsafe` library is a transitive dependency of `iceberg-aws` that is not bundled.
 
 > **NOTE**: You need **all three** JARs. Using only the bundle will result in `ClassNotFoundException: org.apache.iceberg.aws.glue.GlueCatalog`. Missing `failsafe` will cause `NoClassDefFoundError: dev/failsafe/FailsafeException`.
 
-Additionally, you need the Fluss S3 filesystem plugin for remote storage access. Place [fluss-fs-s3-$FLUSS_VERSION$.jar]($FLUSS_MAVEN_REPO_URL$/org/apache/fluss/fluss-fs-s3/$FLUSS_VERSION$/fluss-fs-s3-$FLUSS_VERSION$.jar) in `${FLUSS_HOME}/plugins/s3/`. See [S3 Dependencies](/docs/maintenance/filesystems/s3#dependencies) for details.
+Additionally, you need the Fluss S3 filesystem plugin for remote storage access. Place [fluss-fs-s3-$FLUSS_VERSION$.jar]($FLUSS_MAVEN_REPO_URL$/org/apache/fluss/fluss-fs-s3/$FLUSS_VERSION$/fluss-fs-s3-$FLUSS_VERSION$.jar) in `${FLUSS_HOME}/plugins/s3/`. See [S3 Dependencies](../../maintenance/tiered-storage/filesystems/s3.md#dependencies) for details.
 
-> **TIP**: The Fluss binary distribution already includes `fluss-lake-iceberg-$FLUSS_VERSION$.jar` in `plugins/iceberg/`. You do not need to download it separately  only add the three JARs above (iceberg-aws-bundle, iceberg-aws, failsafe).
+> **TIP**: The Fluss binary distribution already includes `fluss-lake-iceberg-$FLUSS_VERSION$.jar` in `plugins/iceberg/`. You do not need to download it separately — only add the three JARs above (iceberg-aws-bundle, iceberg-aws, failsafe).
 
 #### For the Flink Tiering Service
 
 Place the following JARs in `${FLINK_HOME}/lib`:
 
 1. **Iceberg AWS Bundle**: `iceberg-aws-bundle-1.10.1.jar` (same as above)
-2. **Iceberg AWS**: `iceberg-aws-1.10.1.jar` (same as above  provides `GlueCatalog` class)
-3. **Failsafe**: `failsafe-3.3.2.jar` (same as above  Glue retry logic)
+2. **Iceberg AWS**: `iceberg-aws-1.10.1.jar` (same as above — provides `GlueCatalog` class)
+3. **Failsafe**: `failsafe-3.3.2.jar` (same as above — Glue retry logic)
 4. **Fluss Flink Connector**: [fluss-flink-1.20-$FLUSS_VERSION$.jar]($FLUSS_MAVEN_REPO_URL$/org/apache/fluss/fluss-flink-1.20/$FLUSS_VERSION$/fluss-flink-1.20-$FLUSS_VERSION$.jar) (pick the version matching your Flink runtime)
 5. **Fluss Lake Iceberg**: [fluss-lake-iceberg-$FLUSS_VERSION$.jar]($FLUSS_MAVEN_REPO_URL$/org/apache/fluss/fluss-lake-iceberg/$FLUSS_VERSION$/fluss-lake-iceberg-$FLUSS_VERSION$.jar)
-6. **Fluss Flink Tiering**: [fluss-flink-tiering-$FLUSS_VERSION$.jar]($FLUSS_MAVEN_REPO_URL$/org/apache/fluss/fluss-flink-tiering/$FLUSS_VERSION$/fluss-flink-tiering-$FLUSS_VERSION$.jar)  the tiering job JAR itself
+6. **Fluss Flink Tiering**: [fluss-flink-tiering-$FLUSS_VERSION$.jar]($FLUSS_MAVEN_REPO_URL$/org/apache/fluss/fluss-flink-tiering/$FLUSS_VERSION$/fluss-flink-tiering-$FLUSS_VERSION$.jar) — the tiering job JAR itself
 7. **Fluss S3 Filesystem**: [fluss-fs-s3-$FLUSS_VERSION$.jar]($FLUSS_MAVEN_REPO_URL$/org/apache/fluss/fluss-fs-s3/$FLUSS_VERSION$/fluss-fs-s3-$FLUSS_VERSION$.jar) (if S3 is used as Fluss remote storage)
-8. **Hadoop Client**: [hadoop-client-api-3.3.6.jar](https://repo1.maven.org/maven2/org/apache/hadoop/hadoop-client-api/3.3.6/hadoop-client-api-3.3.6.jar) and [hadoop-client-runtime-3.3.6.jar](https://repo1.maven.org/maven2/org/apache/hadoop/hadoop-client-runtime/3.3.6/hadoop-client-runtime-3.3.6.jar)  required by the Iceberg Parquet writer
+8. **Hadoop Client**: [hadoop-client-api-3.3.6.jar](https://repo1.maven.org/maven2/org/apache/hadoop/hadoop-client-api/3.3.6/hadoop-client-api-3.3.6.jar) and [hadoop-client-runtime-3.3.6.jar](https://repo1.maven.org/maven2/org/apache/hadoop/hadoop-client-runtime/3.3.6/hadoop-client-runtime-3.3.6.jar) — required by the Iceberg Parquet writer
 
 > **NOTE**: Despite the Glue catalog itself not requiring Hadoop, the **tiering service** needs Hadoop classes (`org.apache.hadoop.conf.Configuration`) for writing Parquet files via Iceberg. Use the `hadoop-client-api` and `hadoop-client-runtime` JARs (not the full Hadoop distribution or `flink-shaded-hadoop-2-uber`) to avoid classpath conflicts with Flink's bundled Avro version. Using the uber JAR causes `NoSuchMethodError: LogicalTypes.timestampNanos()`.
 
-> **WARNING**: Do **not** add S3 credentials to Flink's configuration file (`config.yaml` or `flink-conf.yaml`) by appending key-value pairs  this can break Flink's memory configuration parser. The credential provider chain configured in Fluss's `server.yaml` handles S3 access for the tiering service automatically.
+> **WARNING**: Do **not** add S3 credentials to Flink's configuration file (`config.yaml` or `flink-conf.yaml`) by appending key-value pairs — this can break Flink's memory configuration parser. The credential provider chain configured in Fluss's `server.yaml` handles S3 access for the tiering service automatically.
 
 ## Configure Fluss with AWS Glue
 
@@ -169,8 +172,8 @@ datalake.iceberg.client.region: us-east-1
 # Use S3FileIO explicitly (Glue catalog defaults to this when warehouse is s3://)
 datalake.iceberg.io-impl: org.apache.iceberg.aws.s3.S3FileIO
 
-# Optional: restrict catalog to a specific Glue database prefix
-datalake.iceberg.glue.catalog-id: <aws-account-id>
+# Optional: use a Glue Data Catalog in another AWS account (defaults to the caller's account)
+datalake.iceberg.glue.id: <aws-account-id>
 ```
 
 #### Authentication
@@ -198,7 +201,7 @@ Example for ECS Fargate:
 fs.s3a.aws.credentials.provider: com.amazonaws.auth.ContainerCredentialsProvider,com.amazonaws.auth.EnvironmentVariableCredentialsProvider,com.amazonaws.auth.InstanceProfileCredentialsProvider
 ```
 
-The credential provider chain handles automatic credential refresh  no static keys or manual token rotation needed.
+The credential provider chain handles automatic credential refresh — no static keys or manual token rotation needed.
 
 > **NOTE**: For non-standard AWS regions (GovCloud, China), you must also set the S3 endpoint explicitly. Hadoop S3A defaults to `s3.amazonaws.com` which does not work for these partitions:
 > ```yaml
@@ -206,7 +209,7 @@ The credential provider chain handles automatic credential refresh  no static ke
 > fs.s3a.endpoint: s3.cn-north-1.amazonaws.com.cn   # China
 > ```
 
-> **NOTE**: You may see a non-fatal warning: `"Session credentials from the configured AWS credentials provider are not supported for Fluss S3 client-token generation."` This is expected  the delegation token subsystem cannot generate tokens from session credentials, but the actual S3 file operations work correctly via the provider chain.
+> **NOTE**: You may see a non-fatal warning: `"Session credentials from the configured AWS credentials provider are not supported for Fluss S3 client-token generation."` This is expected — the delegation token subsystem cannot generate tokens from session credentials, but the actual S3 file operations work correctly via the provider chain.
 
 **Static Credentials (Testing Only)**:
 
@@ -237,11 +240,11 @@ ${FLINK_HOME}/bin/flink run /path/to/fluss-flink-tiering-$FLUSS_VERSION$.jar \
 This section shows the complete setup from a blank Linux machine to data queryable in Athena. Adapt paths and versions for your environment.
 
 ```bash
-#  1. Install prerequisites 
+# — 1. Install prerequisites —
 # Java 17+ required (Amazon Corretto, OpenJDK, etc.)
 java -version  # Must show 17+
 
-#  2. Install ZooKeeper 
+# — 2. Install ZooKeeper —
 wget -q https://archive.apache.org/dist/zookeeper/zookeeper-3.8.4/apache-zookeeper-3.8.4-bin.tar.gz
 tar -xzf apache-zookeeper-3.8.4-bin.tar.gz -C /opt/zookeeper --strip-components=1
 cat > /opt/zookeeper/conf/zoo.cfg <<EOF
@@ -250,11 +253,11 @@ dataDir=/var/lib/zookeeper
 clientPort=2181
 EOF
 
-#  3. Install Fluss 
+# — 3. Install Fluss —
 wget -q https://dlcdn.apache.org/incubator/fluss/fluss-0.9.1-incubating/fluss-0.9.1-incubating-bin.tgz
 tar -xzf fluss-0.9.1-incubating-bin.tgz -C /opt/fluss --strip-components=1
 
-#  4. Add Glue JARs to Fluss plugins/iceberg/ 
+# — 4. Add Glue JARs to Fluss plugins/iceberg/ —
 wget -O /opt/fluss/plugins/iceberg/iceberg-aws-bundle-1.10.1.jar \
   https://repo1.maven.org/maven2/org/apache/iceberg/iceberg-aws-bundle/1.10.1/iceberg-aws-bundle-1.10.1.jar
 wget -O /opt/fluss/plugins/iceberg/iceberg-aws-1.10.1.jar \
@@ -262,7 +265,7 @@ wget -O /opt/fluss/plugins/iceberg/iceberg-aws-1.10.1.jar \
 wget -O /opt/fluss/plugins/iceberg/failsafe-3.3.2.jar \
   https://repo1.maven.org/maven2/dev/failsafe/failsafe/3.3.2/failsafe-3.3.2.jar
 
-#  5. Write server.yaml 
+# — 5. Write server.yaml —
 cat > /opt/fluss/conf/server.yaml <<EOF
 zookeeper.address: localhost:2181
 coordinator.host: localhost
@@ -279,7 +282,7 @@ datalake.iceberg.warehouse: s3://YOUR-BUCKET/iceberg-warehouse
 datalake.iceberg.client.region: YOUR-REGION
 datalake.iceberg.io-impl: org.apache.iceberg.aws.s3.S3FileIO
 
-# S3 credentials  use the provider chain for your environment
+# S3 credentials — use the provider chain for your environment
 # ECS Fargate:
 fs.s3a.aws.credentials.provider: com.amazonaws.auth.ContainerCredentialsProvider,com.amazonaws.auth.EnvironmentVariableCredentialsProvider,com.amazonaws.auth.InstanceProfileCredentialsProvider
 # EKS (IRSA): use WebIdentityTokenCredentialsProvider instead of ContainerCredentialsProvider
@@ -289,7 +292,7 @@ fs.s3a.aws.credentials.provider: com.amazonaws.auth.ContainerCredentialsProvider
 # fs.s3a.endpoint: s3.YOUR-REGION.amazonaws.com
 EOF
 
-#  6. Start services 
+# — 6. Start services —
 /opt/zookeeper/bin/zkServer.sh start
 sleep 3
 /opt/fluss/bin/coordinator-server.sh start
@@ -297,7 +300,7 @@ sleep 5
 /opt/fluss/bin/tablet-server.sh start
 sleep 5
 
-#  7. Install Flink + tiering JARs 
+# — 7. Install Flink + tiering JARs —
 wget -q https://archive.apache.org/dist/flink/flink-1.20.1/flink-1.20.1-bin-scala_2.12.tgz
 tar -xzf flink-1.20.1-bin-scala_2.12.tgz -C /opt/flink --strip-components=1
 
@@ -320,7 +323,7 @@ wget -O /opt/flink/lib/hadoop-client-runtime-3.3.6.jar \
 /opt/flink/bin/start-cluster.sh
 sleep 5
 
-#  8. Create table + insert data (single Flink SQL session) 
+# — 8. Create table + insert data (single Flink SQL session) —
 cat > /tmp/setup.sql <<EOF
 CREATE CATALOG fluss_catalog WITH ('type' = 'fluss', 'bootstrap.servers' = 'localhost:9123');
 USE CATALOG fluss_catalog;
@@ -332,7 +335,7 @@ INSERT INTO test_table VALUES (1, 'hello'), (2, 'world');
 EOF
 /opt/flink/bin/sql-client.sh -f /tmp/setup.sql
 
-#  9. Start tiering service 
+# — 9. Start tiering service —
 /opt/flink/bin/flink run -d /opt/flink/lib/fluss-flink-tiering-0.9.1-incubating.jar \
   --fluss.bootstrap.servers localhost:9123 \
   --datalake.format iceberg \
@@ -341,7 +344,7 @@ EOF
   --datalake.iceberg.client.region YOUR-REGION \
   --datalake.iceberg.io-impl org.apache.iceberg.aws.s3.S3FileIO
 
-#  10. Query in Athena 
+# — 10. Query in Athena —
 # Wait ~30s for tiering to flush, then:
 #   SELECT * FROM my_database.test_table;
 ```
@@ -376,7 +379,7 @@ CREATE TABLE customer_orders (
 
 Fluss will register the database in ZooKeeper and create a corresponding Iceberg table in the Glue Data Catalog (using the same database name). Once data is ingested and the tiering service commits, Parquet files appear in the S3 warehouse path.
 
-> **NOTE**: The database must be created in the Fluss catalog first  Fluss maintains its own database registry in ZooKeeper. The database name in Fluss maps 1:1 to the Glue database name during tiering.
+> **NOTE**: The database must be created in the Fluss catalog first — Fluss maintains its own database registry in ZooKeeper. The database name in Fluss maps 1:1 to the Glue database name during tiering.
 
 ### Query Data with Athena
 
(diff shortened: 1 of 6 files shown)


luoyuxia (commented):
@qzyu999 Thanks for the pr. LGTM overall. Just one question, have you ever verify these catalogs?

qzyu999:
> @qzyu999 Thanks for the pr. LGTM overall. Just one question, have you ever verify these catalogs?

Hi @luoyuxia, as mentioned in the PR description, no it isn't currently:
> Note: Changes are based on the existing lakekeeper.md as a template, and references were based on existing code and online/offline documentation. The actual AWS Glue/HMS implementations have not yet been tested by the developer.

I spoke with @leekeiabstraction to ask about this and he mentioned that I should be getting it tested E2E, as I wasn't sure whether there were other developers who have done these steps themselves. I am currently working on the AWS Glue steps and will update accordingly. Thank you for your patience!

qzyu999:
As of the latest changes, the AWS Glue integration has been tested E2E and the ingest -> tiering into Iceberg works.

luoyuxia (approved):
+1
cc @polyzos Do you want to have a final review?

qzyu999:
Hi @luoyuxia, quick note I am doing an E2E test also for Hive currently, should be done in a day or two. May we please hold the merge until then? Thanks!

luoyuxia:
> Hi @luoyuxia, quick note I am doing an E2E test also for Hive currently, should be done in a day or two. May we please hold the merge until then? Thanks!

Thanks. Sure. Please let me know if you finish it.

qzyu999:
Hi @luoyuxia, both Glue and Hive have been tested E2E and have been demonstrated to work correctly, PTAL.

website/docs/streaming-lakehouse/integrate-data-lakes/catalogs/glue.md
```diff
@@ -0,0 +1,411 @@
...
+
+[AWS Glue](https://aws.amazon.com/glue/) is a serverless data integration service that includes a central metadata repository known as the AWS Glue Data Catalog. The Glue Data Catalog is compatible with Apache Iceberg, making it a convenient option for managing Iceberg table metadata on AWS.
+
+This guide explains how to configure Fluss to use AWS Glue as its Iceberg catalog. For general Iceberg integration details (table mapping, data types, limitations), see [Iceberg](../formats/iceberg.md).
+
+## How It Works
+
+When Fluss is configured with AWS Glue as its Iceberg catalog:
+
+1. **Data ingestion**: Applications write data to Fluss tables using the Fluss client (Java/Python), Flink SQL, or any Kafka-compatible producer. Fluss stores this data in its real-time log (similar to Kafka topics).
```
leekeiabstraction: Hmm, what is meant by Kafka compatible producer here?

Also instead of real time log, let's say log table.

website/docs/streaming-lakehouse/integrate-data-lakes/catalogs/glue.md
```diff
@@ -0,0 +1,411 @@
...
+2. **Tiering to Iceberg**: A separate Flink job (the [tiering service](maintenance/tiered-storage/lakehouse-storage.md#start-the-datalake-tiering-service)) periodically reads accumulated data from Fluss, converts it to Parquet format, writes the files to S3, and commits an Iceberg snapshot to the Glue Data Catalog.
+3. **Query via Athena/Spark/Trino**: Any Iceberg-compatible engine can discover and query the tiered tables through AWS Glue — no additional configuration needed.
```
leekeiabstraction, line 20: Should union read via Flink SQL be mentioned as well?

website/docs/streaming-lakehouse/integrate-data-lakes/catalogs/glue.md
```diff
@@ -0,0 +1,411 @@
...
+
+```
+┌─ Data Sources ─────────┐     ┌─ Fluss Cluster ──────┐     ┌─ AWS ────────────────────┐
```
leekeiabstraction: I think the diagram could be more precise, tiering should point to s3. Coordinator should write snapshot info into aws glue (please verify).

website/docs/streaming-lakehouse/integrate-data-lakes/catalogs/glue.md
```diff
@@ -0,0 +1,167 @@
...
+        "glue:GetTable",
+        "glue:GetTables",
+        "glue:UpdateTable",
+        "glue:DeleteTable"
+      ],
+      "Resource": [
+        "arn:aws:glue:<region>:<account-id>:catalog",
+        "arn:aws:glue:<region>:<account-id>:database/*",
+        "arn:aws:glue:<region>:<account-id>:table/*"
+      ]
```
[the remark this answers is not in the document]
leekeiabstraction: @qzyu999 can you address this comment? Is it valid?

website/docs/streaming-lakehouse/datalake-catalogs/glue.md
```diff
@@ -0,0 +1,411 @@
...
+
+## Prerequisites
+
+### Java Version
+
+Fluss 0.9.x requires **Java 17 or later** for the Tablet Server. Java 11 will fail with `NoSuchMethodError: MappedByteBuffer.duplicate()`.
+
+### AWS IAM Permissions
+
+The processes running Fluss servers and the Flink tiering service must have IAM permissions for both Glue and S3. Below is a minimal IAM policy:
```
leekeiabstraction, line 69: We should add a note that advise user to scope down the policy to least privileged e.g. specific glue database or catalog.

website/docs/streaming-lakehouse/datalake-catalogs/glue.md
```diff
@@ -0,0 +1,411 @@
...
+
+#### For Fluss Servers (Coordinator & Tablet Servers)
+
+Place the following JARs in the `${FLUSS_HOME}/plugins/iceberg/` directory:
+
+- **Iceberg AWS Bundle**: [iceberg-aws-bundle-1.10.1.jar](https://repo1.maven.org/maven2/org/apache/iceberg/iceberg-aws-bundle/1.10.1/iceberg-aws-bundle-1.10.1.jar)
+- **Iceberg AWS**: [iceberg-aws-1.10.1.jar](https://repo1.maven.org/maven2/org/apache/iceberg/iceberg-aws/1.10.1/iceberg-aws-1.10.1.jar)
+- **Failsafe**: [failsafe-3.3.2.jar](https://repo1.maven.org/maven2/dev/failsafe/failsafe/3.3.2/failsafe-3.3.2.jar) — required by `iceberg-aws` for Glue API retry logic
+
+Both Iceberg JARs are required. The bundle provides AWS SDK v2 dependencies, while `iceberg-aws` provides the `GlueCatalog` class that Iceberg loads via reflection. The plugin classloader cannot find `GlueCatalog` from the bundle alone. The `failsafe` library is a transitive dependency of `iceberg-aws` that is not bundled.
```
leekeiabstraction, line 127: Interesting, iceberg aws bundle does not contain GlueCatalog?

website/docs/streaming-lakehouse/integrate-data-lakes/catalogs/glue.md
```diff
@@ -0,0 +1,411 @@
...
+8. **Hadoop Client**: [hadoop-client-api-3.3.6.jar](https://repo1.maven.org/maven2/org/apache/hadoop/hadoop-client-api/3.3.6/hadoop-client-api-3.3.6.jar) and [hadoop-client-runtime-3.3.6.jar](https://repo1.maven.org/maven2/org/apache/hadoop/hadoop-client-runtime/3.3.6/hadoop-client-runtime-3.3.6.jar) — required by the Iceberg Parquet writer
+
+> **NOTE**: Despite the Glue catalog itself not requiring Hadoop, the **tiering service** needs Hadoop classes (`org.apache.hadoop.conf.Configuration`) for writing Parquet files via Iceberg. Use the `hadoop-client-api` and `hadoop-client-runtime` JARs (not the full Hadoop distribution or `flink-shaded-hadoop-2-uber`) to avoid classpath conflicts with Flink's bundled Avro version. Using the uber JAR causes `NoSuchMethodError: LogicalTypes.timestampNanos()`.
+
+> **WARNING**: Do **not** add S3 credentials to `flink-conf.yaml` by appending key-value pairs — this can break Flink's memory configuration parser. Instead, use environment variables (`AWS_ACCESS_KEY_ID`, `AWS_SECRET_ACCESS_KEY`, `AWS_SESSION_TOKEN`, `AWS_REGION`) or a `HADOOP_CONF_DIR` with a `core-site.xml` file.
```
leekeiabstraction: This warning seems out of place. Should it be near where s3 creds are set up?

website/docs/streaming-lakehouse/datalake-catalogs/glue.md
```diff
@@ -0,0 +1,411 @@
...
+7. **Fluss S3 Filesystem**: [fluss-fs-s3-$FLUSS_VERSION$.jar]($FLUSS_MAVEN_REPO_URL$/org/apache/fluss/fluss-fs-s3/$FLUSS_VERSION$/fluss-fs-s3-$FLUSS_VERSION$.jar) (if S3 is used as Fluss remote storage)
+8. **Hadoop Client**: [hadoop-client-api-3.3.6.jar](https://repo1.maven.org/maven2/org/apache/hadoop/hadoop-client-api/3.3.6/hadoop-client-api-3.3.6.jar) and [hadoop-client-runtime-3.3.6.jar](https://repo1.maven.org/maven2/org/apache/hadoop/hadoop-client-runtime/3.3.6/hadoop-client-runtime-3.3.6.jar) — required by the Iceberg Parquet writer
+
+> **NOTE**: Despite the Glue catalog itself not requiring Hadoop, the **tiering service** needs Hadoop classes (`org.apache.hadoop.conf.Configuration`) for writing Parquet files via Iceberg. Use the `hadoop-client-api` and `hadoop-client-runtime` JARs (not the full Hadoop distribution or `flink-shaded-hadoop-2-uber`) to avoid classpath conflicts with Flink's bundled Avro version. Using the uber JAR causes `NoSuchMethodError: LogicalTypes.timestampNanos()`.
```
leekeiabstraction, line 148: I do not see Hadoop client runtime in the list above. 

Also, this note might be unnecessary? It's easier for users to follow a guide with exactly the list of things to use. Warning them about what not to use when it's not listed seems like additional cognitive load.

website/docs/streaming-lakehouse/integrate-data-lakes/catalogs/glue.md
```diff
@@ -0,0 +1,411 @@
...
+
+# Optional: restrict catalog to a specific Glue database prefix
+datalake.iceberg.glue.catalog-id: <aws-account-id>
+```
+
+#### Authentication
+
+**IAM Role (Recommended)**: If running on AWS (EKS, ECS, EC2) with an attached IAM role, no credential configuration is needed for the **Glue catalog connection**. The Iceberg AWS SDK uses the [default credentials provider chain](https://docs.aws.amazon.com/sdk-for-java/latest/developer-guide/credentials-chain.html) automatically.
+
+However, the **Fluss S3 filesystem plugin** (used for `remote.data.dir`) has a known limitation: it uses a custom delegation token mechanism that does not support IMDSv2 or ECS task role credentials. This affects:
```
leekeiabstraction: Are you ok to open an issue on this? Thank you

leekeiabstraction (commented):
Thank you for the PR! It's looking good and thorough. 

I have left some comments. PTAL.

website/docs/streaming-lakehouse/integrate-data-lakes/catalogs/glue.md
fresh-borzoni: Q: can you be more specific about the limitation here? 
I feel it's AI hallucination here

The real constraint is just: you must set either static keys or s3.assumed.role.arn - Fluss won't auto-discover an ambient role with zero config. So the correct Fargate/IMDSv2 guidance is simply s3.assumed.role.arn + s3.region, no static keys.

And workaround advised also feels wrong: it feeds temporary credentials into the one step that demands permanent ones.

PTAL, can you verify this?

website/docs/streaming-lakehouse/integrate-data-lakes/catalogs/hive.md
```diff
@@ -0,0 +1,432 @@
...
+
+```
+┌─ Data Sources ─────────┐     ┌─ Fluss Cluster ──────┐     ┌─ Hadoop / On-Prem ────────┐
+│                         │     │                      │     │                            │
+│  App (Fluss client)   ─┼────▶│  Coordinator         │     │  Hive Metastore            │
+│  Flink CDC job        ─┼────▶│  TabletServer(s)     │     │  (table metadata)          │
+│  Kafka producer       ─┼────▶│  ZooKeeper           │     │                            │
+│                         │     │                      │     │  HDFS / S3 / OSS           │
+└─────────────────────────┘     └──────────┬───────────┘     │  (Parquet data files)      │
+                                           │                  │                            │
```
MehulBatra: Can you create a proper diagram, put an image using draw.io or excali draw.
Thank you!

qzyu999: Hi @MehulBatra, thanks for the suggestion, to keep it programmatic, WDYT of utilizing Mermaid diagrams rather than draw.io or Excalidraw (both which I have used before and think are great)? I believe for Docusaurus there's a first-party plugin for it that can be added.

MehulBatra: The goal is to keep our documentation and diagrams professional and maintain a high-quality standard, rather than looking like an AI-generated draft. Use your judgment here: if a Mermaid diagram is clear, polished, and professional enough, feel free to use it. If a draw.io diagram better matches the style and theme of the website and the existing documentation, that's perfectly fine as well.

The most important thing is consistency and uniformity across the site, so whichever approach you choose should align with the overall look and feel of the documentation.

website/docs/streaming-lakehouse/integrate-data-lakes/catalogs/glue.md
qzyu999: I validated the `glue.md` documentation end-to-end on AWS (ECS Fargate) with Fluss 0.9.1-incubating, Flink 1.20.1, Iceberg 1.10.1, and AWS Glue. This is a summary of findings.

## What Works ✅

The following are **validated and working** on ECS Fargate with the `s3.assumed.role.arn` credential approach:

1. **Fluss cluster starts** — Coordinator + TabletServer + ZooKeeper
2. **S3 filesystem plugin initializes** — "Using default AWS credential chain with AssumeRole" logged
3. **Glue catalog connection** — Iceberg GlueCatalog loads via `S3FileIO`, connects to Glue
4. **Table creation** — Flink SQL `CREATE TABLE ... WITH ('table.datalake.enabled'='true')` succeeds
5. **Table registered in Glue** — confirmed via `aws glue get-table`
6. **Data insertion** — Flink SQL INSERT succeeds
7. **Tiering job starts** — `flink run -d fluss-flink-tiering-0.9.1-incubating.jar` submits and runs (RUNNING state confirmed)
8. **Tiering reads data from TabletServer** — reaches the commit phase

## What Fails ❌

**Tiering snapshot commit fails** with a misleading error:
```
java.io.IOException: Fail to prepare commit table lake snapshot for db.table to Fluss.
Caused by: java.lang.IllegalStateException: Field 'table_id' is not set
```

This is NOT a protobuf serialization issue. It's **two bugs** working together:

## Bug 1: Error masking in `CoordinatorService.prepareLakeTableSnapshot()`

**File:** `fluss-server/.../CoordinatorService.java` (the handler for `PrepareLakeTableSnapshot` RPC)

When `storeLakeTableOffsetsFile()` throws an exception, the catch block:
- ✅ Sets `error_code` and `error_message` in the response
- ❌ Does NOT set `table_id` in the response
- ❌ Does NOT log the exception

```java
// Server side (v0.9.1-incubating, exact released code):
try {
    long tableId = bucketOffsets.getTableId();
    // ... storeLakeTableOffsetsFile() throws here ...
    pbPrepareLakeTableRespForTable.setTableId(tableId);         // ← only on success
    pbPrepareLakeTableRespForTable.setLakeTableOffsetsPath(fsPath.toString());
} catch (Exception e) {
    Errors error = ApiError.fromThrowable(e).error();
    pbPrepareLakeTableRespForTable.setError(error.code(), error.message());
    // ← table_id NOT SET
    // ← exception NOT LOGGED
}
```

## Bug 2: Client reads `table_id` before checking for errors

**File:** `fluss-flink/.../FlussTableLakeSnapshotCommitter.java`

```java
// Client side (v0.9.1-incubating, exact released code):
prepareResp = pbPrepareLakeTableRespForTables.get(0);
checkState(
    prepareResp.getTableId() == tableId,   // ← CRASHES HERE (table_id not set in error response)
    ...);
if (prepareResp.hasErrorCode()) {          // ← NEVER REACHED
    throw ApiError.fromErrorMessage(prepareResp).exception();  // actual error would surface here
}
```

**Result:** The real error from `storeLakeTableOffsetsFile()` is permanently lost. The user sees `"Field 'table_id' is not set"` which provides zero diagnostic value. The tiering service retries indefinitely.

## The Underlying Trigger: `s3.assumed.role.arn` doesn't pass credentials to Hadoop S3A

The reason `storeLakeTableOffsetsFile()` fails is that it writes to the `remote.data.dir` (S3) using Hadoop S3A via the Fluss FileSystem abstraction. Looking at `S3FileSystemPlugin.setCredentialProvider()`:

When `s3.assumed.role.arn` is set (without static keys), the plugin:
- Logs "Using default AWS credential chain with AssumeRole"
- Skips the delegation token credential provider setup
- Does NOT set any explicit Hadoop S3A credentials (`fs.s3a.access.key`, etc.)
- Relies on Hadoop's default credential provider chain

On ECS Fargate, Hadoop's default chain should discover ECS container credentials. **But in practice, the `storeLakeTableOffsetsFile()` write fails** — likely because the bundled Hadoop version doesn't properly support the ECS container credential endpoint, or there's a classpath issue.

**The previous approach (injecting temp credentials from the ECS metadata endpoint as `s3.access.key`/`s3.secret.key`/`s3.session.token`) DID work** because those get mapped directly to `fs.s3a.access.key`/`fs.s3a.secret.key`/`fs.s3a.session.token` — Hadoop S3A uses them for all file operations.

## Summary

| Credential Approach | S3 Init | Table Creation (Glue) | Tiering (S3 write) |
|---|---|---|---|
| `s3.access.key` + `s3.secret.key` + `s3.session.token` (temp creds from ECS metadata) | ✅ | ✅ | ✅ (confirmed in earlier POC) |
| `s3.assumed.role.arn` only | ✅ | ✅ | ❌ (`storeLakeTableOffsetsFile` fails silently) |

## Questions

1. The error handling in `CoordinatorService.prepareLakeTableSnapshot()` swallows exceptions — when `storeLakeTableOffsetsFile()` fails, the exception isn't logged and `table_id` isn't included in the error response. The client then crashes on a missing field before it ever reads the actual error. Can we get the exception logged and `table_id` set in the error path? Without this, any S3/filesystem failure in tiering is completely undebuggable.

2. In `FlussTableLakeSnapshotCommitter.prepareLakeSnapshot()`, the code calls `getTableId()` before checking `hasErrorCode()`. Since the error response doesn't include `table_id` (see above), this always throws `IllegalStateException` and hides the real error message. Flipping the check order would at least surface what's actually failing.

3. This is the main blocker for us: when we set `s3.assumed.role.arn` in `server.yaml`, the S3 filesystem plugin only uses it for delegation token generation — it does NOT configure Hadoop S3A (which actually writes files to S3) to use that role. So `storeLakeTableOffsetsFile()` tries to write to S3 with no credentials. On ECS Fargate/EKS, we'd expect `s3.assumed.role.arn` to mean "use this role for all S3 operations," not just for one internal subsystem. Is there a plan to make this work end-to-end? For now we're falling back to injecting `s3.access.key`/`s3.secret.key`/`s3.session.token` which does work but requires fetching temp creds from the ECS metadata endpoint at startup.

## Environment

- Fluss 0.9.1-incubating (binary from apache.org)
- `fluss-flink-tiering-0.9.1-incubating.jar` (from Maven Central)
- Flink 1.20.1
- Iceberg 1.10.1 + iceberg-aws-bundle + iceberg-aws + failsafe
- ECS Fargate
- Hadoop: `hadoop-client-api-3.3.6` + `hadoop-client-runtime-3.3.6`

fresh-borzoni: Thank you for putting this, we should definitely improve error handling here and `table_id`issue.

The EKS/IRSA "no credential config needed" note doesn't hold for remote.data.dir/tiering as S3A's default chain has no web-identity provider (intentional decision in Hadoop), so IRSA won't resolve there. Needs:s3.aws.credentials.provider: com.amazonaws.auth.WebIdentityTokenCredentialsProvider,com.amazonaws.auth.EnvironmentVariableCredentialsProvider,com.amazonaws.auth.InstanceProfileCredentialsProvider or similar(that's what we have in our cluster on EKS)

Can you try this and share the results?

qzyu999: Hi @fresh-borzoni, thanks for your help with this. I tried adding the credentials provider to `server.yaml`:

```yaml
fs.s3a.aws.credentials.provider: com.amazonaws.auth.ContainerCredentialsProvider,com.amazonaws.auth.EnvironmentVariableCredentialsProvider,com.amazonaws.auth.InstanceProfileCredentialsProvider
```

Confirmed it loads (shows in Coordinator startup log), but tiering still fails with the same `Field 'table_id' is not set` error.

Since the error-masking bug in `CoordinatorService.prepareLakeTableSnapshot()` hides the real exception from `storeLakeTableOffsetsFile()`, I can't see what Hadoop S3A is actually throwing. The Coordinator log has zero errors, the exception is caught, stuffed into the RPC response, and then the client crashes on `getTableId()` before reading it.

My setup is ECS Fargate (not EKS), so I'm using `ContainerCredentialsProvider` instead of `WebIdentityTokenCredentialsProvider`. The container creds endpoint (`http://169.254.170.2`) is reachable, and Iceberg's `S3FileIO` uses it successfully for table creation in Glue and writing metadata to S3.

Question: in your EKS setup, do you set `fs.s3a.aws.credentials.provider` directly in `server.yaml`? Or via a separate `core-site.xml` / `HADOOP_CONF_DIR`? I'm wondering if the Fluss S3 plugin classloader doesn't pass it through to the Hadoop `FileSystem` instance used by `storeLakeTableOffsetsFile()`.

For context, the previous approach (injecting temp creds from the ECS metadata endpoint as `s3.access.key`/`s3.secret.key`/`s3.session.token`) does work, tiering completed successfully. So the issue is specifically with the credential-provider-chain path for `remote.data.dir` writes.

fresh-borzoni: @qzyu999 
Okay, it seems it's hard to debug without fixing the pesky bug.

I'll add the fix, can you build fluss and test on ECS then? I wouldn't expect ECS to have problems since ContainerCredentialsProvide should be in a chain, so it might be something silly, like region not set or similar.

Yes, I set fs.s3a.aws.credentials.provider directly in server.yaml

fresh-borzoni: cc @qzyu999 @luoyuxia @leekeiabstraction 
https://github.com/apache/fluss/pull/3550

qzyu999: Hi @fresh-borzoni, thanks for the fixes and apologies for the delay.

Built from main (includes #3550 + the credentials provider fix) and tested E2E on ECS Fargate (GovCloud us-gov-west-1). Results:

**Auth works ✅** — full pipeline validated with this config in `server.yaml`:

    fs.s3a.aws.credentials.provider: com.amazonaws.auth.ContainerCredentialsProvider,com.amazonaws.auth.EnvironmentVariableCredentialsProvider,com.amazonaws.auth.InstanceProfileCredentialsProvider
    fs.s3a.endpoint: s3.us-gov-west-1.amazonaws.com

Write → Tier → S3 Parquet → Glue commit → Read back → Athena query: all working. You were right, it was "something silly, like region not set", GovCloud needs `fs.s3a.endpoint` explicitly since Hadoop S3A defaults to `s3.amazonaws.com`.

I've updated the `glue.md` auth section in this PR to document `fs.s3a.aws.credentials.provider` as the recommended approach (with a provider chain table for ECS/EKS/EC2). Removed the old temp-key workaround.

**Two unrelated issues found during testing:**

1. **State machine race in `LakeTableTieringManager`**: On first tiering round, the coordinator logs:

       Fail to change state for table 1 from Scheduled to Tiered as it's not a valid state change.
       Fail to change state for table 1 from Scheduled to Scheduled as it's not a valid state change.

   The S3 write succeeds, but the state transition fails. Second round works fine. Seems like tiering completes before the manager transitions Scheduled → Pending → Tiering (small dataset + fast S3). Should I file a separate issue?

2. **Iceberg lake source doesn't support union read for PK tables**: `IcebergLakeSource.createRecordReader()` returns `IcebergRecordReader` which doesn't implement `SortedRecordReader`. Flink SQL SELECT on a PK table throws `UnsupportedOperationException: lake records must instance of sorted view`. Log tables work fine. Paimon and Hudi have sorted reader implementations. Is this a known limitation?

website/docs/streaming-lakehouse/integrate-data-lakes/catalogs/hive.md
qzyu999: Hi @MehulBatra, added some Mermaid diagrams for both Glue and Hive. I searched and no other diagrams exist in the docs currently. Looking at the diagrams myself, I think that they are highly clear and the style fits well into the existing format.

qzyu999:
Hi @fresh-borzoni, should I update the doc to target a future release since the 0.9.1 jar is having issues? What are your thoughts on this issue, given that if they were to follow the doc as-is it would lead to errors for Glue.

MehulBatra (approved):

website/docs/streaming-lakehouse/integrate-data-lakes/catalogs/hive.md
```diff
@@ -0,0 +1,452 @@
...
+        curl -sSfL -o /plugins/iceberg/jackson-databind-2.15.2.jar \
+          https://repo1.maven.org/maven2/com/fasterxml/jackson/core/jackson-databind/2.15.2/jackson-databind-2.15.2.jar
+        curl -sSfL -o /plugins/iceberg/jackson-annotations-2.15.2.jar \
+          https://repo1.maven.org/maven2/com/fasterxml/jackson/core/jackson-annotations/2.15.2/jackson-annotations-2.15.2.jar
+        echo 'Plugins ready:' && ls -la /plugins/iceberg/
+    volumes:
+      - fluss-plugins:/plugins
+
+  coordinator:
+    image: apache/fluss:0.9.0-incubating
```
fresh-borzoni: why 0.9.0?

website/docs/streaming-lakehouse/integrate-data-lakes/catalogs/hive.md
```diff
@@ -0,0 +1,452 @@
...
+      - fluss-plugins:/opt/fluss/plugins
+    healthcheck:
+      test: ["CMD-SHELL", "bash -c 'echo > /dev/tcp/coordinator/9123' 2>/dev/null"]
+      interval: 10s
+      timeout: 5s
+      retries: 12
+      start_period: 30s
+
+  tablet-server:
+    image: apache/fluss:0.9.0-incubating
```
fresh-borzoni: ditto

website/docs/streaming-lakehouse/datalake-catalogs/glue.md
```diff
@@ -0,0 +1,422 @@
...
+SELECT * FROM my_database.customer_orders LIMIT 10;
+```
+
+### Query Data with Flink (Union Read)
+
+```sql title="Flink SQL"
+SET 'execution.runtime-mode' = 'batch';
+
+-- Union read: combines fresh data in Fluss with historical data in Iceberg
+SELECT COUNT(*) FROM customer_orders;
```
fresh-borzoni: I don't think we can do union read on Iceberg PK tables, it's a known limitation.

website/docs/streaming-lakehouse/datalake-catalogs/hive.md
```diff
@@ -0,0 +1,452 @@
...
+-- Query the tiered Iceberg table from Hive Metastore catalog
+SELECT * FROM hive_catalog.my_database.daily_events;
+```
+
+### Query Data with Flink (Union Read)
+
+```sql title="Flink SQL"
+-- Union read: combines fresh data in Fluss with historical data in Iceberg
+SET 'execution.runtime-mode' = 'streaming';
+SELECT * FROM daily_events;
```
fresh-borzoni, line 431: ditto

fresh-borzoni (commented):
@qzyu999 Thank you for checking this, left a couple of comments. I'll check more thoroughly when I'll have a bit more time, but what I've found so far, PTAL.

Nothing is actually broken in 0.9.1 for the path this guide documents:
#3550 only sets table_id and logs the exception on the error branch and it changes whether you can see a failure, not whether tiering works.

Union read on PK tables on Iceberg doesn't really work now.

UPD
finished review

website/docs/streaming-lakehouse/datalake-catalogs/glue.md
fresh-borzoni: @qzyu999 this one as well, it's a valid point

website/docs/streaming-lakehouse/integrate-data-lakes/catalogs/hive.md
```diff
@@ -0,0 +1,452 @@
...
+
+The **Hive Metastore (HMS)** is a central metadata repository commonly used in Apache Hadoop and other big data ecosystems to store schema and metadata information for tables. Apache Iceberg provides native integration with Hive Metastore, storing Iceberg table names and metadata locations directly within HMS.
+
+This guide explains how to configure Fluss to use Hive Metastore as its Iceberg catalog. For general Iceberg integration details (table mapping, data types, limitations), see [Iceberg](../formats/iceberg.md).
+
+## How It Works
+
+When Fluss is configured with Hive Metastore as its Iceberg catalog:
+
+1. **Data ingestion**: Applications write data to Fluss tables using the Fluss client (Java/Python), Flink SQL, or any Kafka-compatible producer. Fluss stores this data in its real-time log.
```
fresh-borzoni: Kafka-compatible producer?

website/docs/streaming-lakehouse/integrate-data-lakes/catalogs/glue.md
fresh-borzoni: @qzyu999 Yes, thank you. For "State machine race" let's file an issue, please. 👍 

I answered about Iceberg PK union read in the review already.

website/docs/streaming-lakehouse/integrate-data-lakes/catalogs/glue.md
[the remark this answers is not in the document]
qzyu999: Hi @leekeiabstraction, the Copilot suggestion about `table/<db>/<table>` is technically more precise, but `table/*` does work as a wildcard matching all tables across all databases in IAM for Glue. I've added a TIP (c0af314f) recommending users scope to specific databases in production: `table/<your-database>/*`.

website/docs/streaming-lakehouse/integrate-data-lakes/catalogs/glue.md
qzyu999: Hi @leekeiabstraction, fixed in 7c56fc47, removed "Kafka-compatible producer" from glue.md. Now says "Fluss client (Java/Python) or Flink SQL". Also, in 9af786ad replaced "real-time log" with "log tables" in both `glue.md` and `hive.md`.

website/docs/streaming-lakehouse/integrate-data-lakes/catalogs/glue.md
qzyu999: Hi @leekeiabstraction, added union read section in the Usage Example (see "Query Data with Flink (Union Read)"). Also added a NOTE clarifying that PK table union reads are not yet supported with Iceberg per @fresh-borzoni's feedback (c0af314f).

website/docs/streaming-lakehouse/integrate-data-lakes/catalogs/glue.md
qzyu999: Hi @leekeiabstraction, fixed in 93986a3, made diagrams more precise. Verified via code: the Coordinator creates the table in Glue (on CREATE TABLE), while the tiering job writes Parquet to S3 AND commits the Iceberg snapshot to Glue (via Iceberg SDK's GlueCatalog). Arrows now reflect this clearly, with dashed lines for read paths.

website/docs/streaming-lakehouse/datalake-catalogs/glue.md
qzyu999: Hi @leekeiabstraction and @fresh-borzoni, fixed in c0af314f, added TIP with scoped ARN examples.

website/docs/streaming-lakehouse/datalake-catalogs/glue.md
qzyu999: Hi @leekeiabstraction, correct, confirmed via E2E testing. The bundle contains the AWS SDK v2 dependencies but the `GlueCatalog` class itself lives in `iceberg-aws`. Both JARs are required. The doc explains this.

website/docs/streaming-lakehouse/integrate-data-lakes/catalogs/glue.md
qzyu999: Hi @leekeiabstraction, moved and rewritten in 334b6b31, the warning now appears contextually in the Flink JARs section rather than the auth section.

website/docs/streaming-lakehouse/datalake-catalogs/glue.md
qzyu999: Hi @leekeiabstraction, fixed, `hadoop-client-api-3.3.6.jar` and `hadoop-client-runtime-3.3.6.jar` are now listed as item 8 in the Flink JARs section. Also simplified the accompanying NOTE per your feedback about reducing cognitive load.

website/docs/streaming-lakehouse/integrate-data-lakes/catalogs/hive.md
qzyu999: Hi @fresh-borzoni, fixed in c0af314f, removed "Kafka-compatible producer" from `hive.md` step 1. Both docs now just say "Fluss client (Java/Python) or Flink SQL".

website/docs/streaming-lakehouse/integrate-data-lakes/catalogs/hive.md
qzyu999: Hi @fresh-borzoni , fixed in c0af314f, replaced hardcoded 0.9.0-incubating with `$FLUSS_VERSION$` placeholder (same pattern used in JARs sections).

website/docs/streaming-lakehouse/integrate-data-lakes/catalogs/hive.md
qzyu999: Hi @fresh-borzoni , same fix as above (c0af314f).

website/docs/streaming-lakehouse/datalake-catalogs/glue.md
qzyu999: Hi @fresh-borzoni, thanks for confirming, fixed in c0af314f, added a NOTE under the Union Read section in both `glue.md` and `hive.md` explaining PK table union reads are not supported with Iceberg, and to query tiered PK data via Athena/Spark instead.

website/docs/streaming-lakehouse/datalake-catalogs/hive.md
qzyu999: Hi @fresh-borzoni, same fix as above (c0af314f).

qzyu999:
Hi @fresh-borzoni, thanks for the clarification, you're right. I re-traced the code path on `v0.9.1-incubating` and confirmed Method C (`fs.s3a.aws.credentials.provider`) does work on 0.9.1:

**How it works on 0.9.1** (`S3FileSystemPlugin.java`):

1. **Line 87-94** (`getHadoopConfiguration`): Iterates all Fluss config keys matching prefixes `"s3.", "s3a.", "fs.s3a."` and copies them to Hadoop config with `fs.s3a.` prefix. So `fs.s3a.aws.credentials.provider` set in `server.yaml` reaches Hadoop config.

2. **Line 132-155** (`setCredentialProvider`): Checks `hasStaticKeys` (line 133-135) and `hasRoleArn` (line 136). When neither is set, falls into the `else` branch (line 143) which calls `S3DelegationTokenReceiver.updateHadoopConfig()`.

3. **`S3DelegationTokenReceiver.updateHadoopConfig()`**: Reads the existing `fs.s3a.aws.credentials.provider` value, and **prepends** `DynamicTemporaryAWSCredentialsProvider` to the chain (doesn't overwrite). Final chain becomes: `DynamicTemporaryAWSCredentialsProvider,ContainerCredentialsProvider,...`

4. At runtime, `DynamicTemporaryAWSCredentialsProvider` tries to generate delegation tokens → fails with the non-fatal warning → Hadoop S3A falls through to `ContainerCredentialsProvider` → works.

**Your bfaaffb8 fix**: Adds an early `return` when `hasCredentialProvider` is true, skipping the delegation token prepend entirely. Result is cleaner (no warning), but the end behavior is the same.

**So the doc and Quick Start script are correct for 0.9.1.** The only user-visible difference is this non-fatal warning on 0.9.x:

> `Session credentials from the configured AWS credentials provider are not supported for Fluss S3 client-token generation.`

I've documented this in the Authentication section NOTE. Should I also add a note in the Authentication section that this warning is expected on 0.9.x and goes away in the next release? Also to what extent do you think I should elaborate on these details within the doc itself?

**Compatibility matrix:**

| Auth Method | Version | ECS Fargate | EKS (IRSA) | EC2 | Notes |
|---|---|---|---|---|---|
| A: Static keys (`s3.access.key`+`s3.secret.key`) | 0.9.x | ✅ | ✅ | ✅ | Keys expire if session tokens |
| A: Static keys | next | ✅ | ✅ | ✅ | Same |
| B: Assumed role (`s3.assumed.role.arn`) | 0.9.x | ❌ | ✅ | ✅ | Relies on Hadoop default chain (no ContainerCredentialsProvider) |
| B: Assumed role | next | ❌ | ✅ | ✅ | Same as 0.9.x. Cannot combine with C (throws error). |
| C: Provider chain (`fs.s3a.aws.credentials.provider`) | 0.9.x | ✅ ⚠️ | ✅ ⚠️ | ✅ ⚠️ | Works; delegation token warning (non-fatal) |
| C: Provider chain | next | ✅ | ✅ | ✅ | Clean — no warning |

⚠️ = Non-fatal warning logged but S3 operations succeed.

Given this, the doc recommends C as the primary approach (works everywhere, all versions) and A as a testing-only fallback. B is not documented since it only works on a subset of environments and cannot be combined with C on the next release.


_**Note: I was only able to test with ECS and am not able to test EKS/EC2. I relied on AI to do a line-by-line code walkthrough to estimate what would be the results for the interaction between Fluss versioning and AWS environment. I've also cleaned up all the other review comments, PTAL and LMK how we can best proceed.**_

website/docs/streaming-lakehouse/integrate-data-lakes/catalogs/glue.md
qzyu999: Hi @fresh-borzoni, thanks for your feedback, filed an issue: here https://github.com/apache/fluss/issues/3672.

fresh-borzoni:
@qzyu999 can you rebase the branch and resolve conflict, please?
I'll try to take a look today and check everything.

regarding to B - it doesn't make sense for ECS, as we are already using assumed role internally there, so it's like use assumed role to generate another assumed role, feels convoluted, but I need to take a look to make sense.

qzyu999:
> @qzyu999 can you rebase the branch and resolve conflict, please? I'll try to take a look today and check everything.
> 
> regarding to B - it doesn't make sense for ECS, as we are already using assumed role internally there, so it's like use assumed role to generate another assumed role, feels convoluted, but I need to take a look to make sense.

Hi @fresh-borzoni, rebased and resolved the merge conflicts, PTAL when available.

fresh-borzoni (approved):
@qzyu999 Thank you, LGTM 👍 
I pushed a commit with small fixes on top, hope you don't mind

Also here is an issue for ECS behaviour you observed: https://github.com/apache/fluss/issues/3761

PR #3427, opened 2026-06-03
[server] Keep flushedLogOffset monotonic to fix row count after failover

### LINKED ISSUE #3386: [test] Unstable test FlussAdminITCase.testGetTableStatsForPrimaryKeyTableWithFailover
https://github.com/apache/fluss/actions/runs/26438789553/job/77827809044

```
[ERROR] org.apache.fluss.client.admin.FlussAdminITCase.testGetTableStatsForPrimaryKeyTableWithFailover  Time elapsed: 10.895 s  <<< FAILURE!
org.opentest4j.AssertionFailedError: 
```

wattt3:
### Purpose

Linked issue: close #3386

`FlussAdminITCase.testGetTableStatsForPrimaryKeyTableWithFailover` is a flaky test(`expected: 170 but was: 160`).

The root cause is `flushedLogOffset = exclusiveUpToLogOffset` in `KvTablet#flush`, which does not keep `flushedLogOffset` monotonically increasing.

When `flush` is called with an `exclusiveUpToLogOffset` lower than the current `flushedLogOffset`, `flushedLogOffset` is rolled back to that lower value, while `rowCount` is not. As a result, the `(flushedLogOffset, rowCount)` pair becomes inconsistent.

During failover, a kv snapshot can capture this inconsistent pair, and recovery then replays the log from the stale offset and double-counts the already-applied deletes, under-reporting the row count

### Brief change log

- `KvTablet#flush` now only advanced `flushedLogOffset` when `exclusiveUpToLogOffset` is greater than current value.

### DIFF
--- a/fluss-server/src/main/java/org/apache/fluss/server/kv/KvTablet.java
+++ b/fluss-server/src/main/java/org/apache/fluss/server/kv/KvTablet.java
@@ -687,7 +687,9 @@ public final class KvTablet {
                     } else {
                         try {
                             int rowCountDiff = kvPreWriteBuffer.flush(exclusiveUpToLogOffset);
-                            flushedLogOffset = exclusiveUpToLogOffset;
+                            if (exclusiveUpToLogOffset > flushedLogOffset) {
+                                flushedLogOffset = exclusiveUpToLogOffset;
+                            }
                             if (rowCount != ROW_COUNT_DISABLED) {
                                 // row count is enabled, we update the row count after flush.
                                 long currentRowCount = rowCount;
--- a/fluss-server/src/test/java/org/apache/fluss/server/kv/KvTabletTest.java
+++ b/fluss-server/src/test/java/org/apache/fluss/server/kv/KvTabletTest.java
@@ -59,6 +59,7 @@ import org.apache.fluss.server.kv.rocksdb.RocksDBStatistics;
 import org.apache.fluss.server.kv.rowmerger.RowMerger;
 import org.apache.fluss.server.kv.scan.OpenScanResult;
 import org.apache.fluss.server.kv.scan.ScannerContext;
+import org.apache.fluss.server.kv.snapshot.TabletState;
 import org.apache.fluss.server.log.FetchIsolation;
 import org.apache.fluss.server.log.LogAppendInfo;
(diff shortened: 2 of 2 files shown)


wattt3:
Hi @swuferhong 
Could you please take a look when you have time?
Thanks!

swuferhong (commented):
LGTM overall, please add a ut test in `KvTabletTest`: write some INSERT/DELETE, call flush(highOffset), then flush(lowOffset), and assert that flushedLogOffset does not regress, rowCount stays unchanged, and `getTabletState()` is self-consistent.

wattt3:
@swuferhong 
Thank you for the review.
I add a unit test. PTAL.

fluss-server/src/main/java/org/apache/fluss/server/kv/KvTablet.java
```diff
@@ -687,7 +687,9 @@ public void flush(long exclusiveUpToLogOffset, FatalErrorHandler fatalErrorHandl
                     } else {
                         try {
                             int rowCountDiff = kvPreWriteBuffer.flush(exclusiveUpToLogOffset);
-                            flushedLogOffset = exclusiveUpToLogOffset;
+                            if (exclusiveUpToLogOffset > flushedLogOffset) {
```
platinumhamburg, line 690: Can we move the guard to before kvPreWriteBuffer.flush()?

platinumhamburg (commented):
LGTM, just one minor question.

fluss-server/src/main/java/org/apache/fluss/server/kv/KvTablet.java
wattt3: @platinumhamburg 
I tried moving the guard before kvPreWriteBuffer.flush(), but it broke some tests that call flush(exclusiveUpToLogOffset = Long.MAX_VALUE).
Once flushedLogOffset reaches MAX, the guard skips the later flushes.
I could adjust those tests to make the move work, but I'd lean toward keeping it as is.
WDYT?

platinumhamburg: @wattt3 Yes, I agree, let's keep it as is.

PR #3429, opened 2026-06-03
[server] Add per-task timeout for RebalanceManager to prevent indefinite blocking

### LINKED ISSUE #3096: [server]  Add timeout mechanism for rebalance tasks to prevent indefinite blocking on missing completion events
### Search before asking

### Motivation

Currently, the rebalance implementation relies on receiving a completion event from the previous task before proceeding to the next one. If the completion event is lost (e.g., due to request loss, network issues, or server crashes), the entire rebalance process gets stuck indefinitely with no way to recover.

### Solution

Introduce a configurable timeout mechanism for rebalance tasks:
1. When a rebalance task is dispatched, start a timeout timer (e.g., configurable via rebalance.task.timeout)
2. If the completion event is not received within the timeout period, automatically treat the task as failed/timed-out and trigger the next task or retry
3. Add appropriate logging to indicate when a timeout occurs vs. a normal completion

This ensures the rebalance process can make forward progress even when individual completion events are lost, 
improving the overall robustness of the rebalance mechanism.

### Anything else?

_No response_

### Willingness to contribute

swuferhong:
### Purpose

Linked issue: close #3096

- Fix a liveness issue where RebalanceManager could block indefinitely if a rebalance bucket's ISR never converges (e.g., target replica is on a dead TabletServer).
- Introduce a per-task timeout (2 minutes) with periodic checks (every 30 seconds). When an in-flight bucket exceeds the timeout, it is treated as COMPLETED, unblocking subsequent buckets and the overall rebalance.
- Add RebalanceTaskTimeoutEvent to the coordinator event system for safe single-threaded state mutation.

### DIFF
--- a/fluss-server/src/main/java/org/apache/fluss/server/coordinator/CoordinatorEventProcessor.java
+++ b/fluss-server/src/main/java/org/apache/fluss/server/coordinator/CoordinatorEventProcessor.java
@@ -85,6 +85,7 @@ import org.apache.fluss.server.coordinator.event.NotifyKvSnapshotOffsetEvent;
 import org.apache.fluss.server.coordinator.event.NotifyLakeTableOffsetEvent;
 import org.apache.fluss.server.coordinator.event.NotifyLeaderAndIsrResponseReceivedEvent;
 import org.apache.fluss.server.coordinator.event.RebalanceEvent;
+import org.apache.fluss.server.coordinator.event.RebalanceTaskTimeoutEvent;
 import org.apache.fluss.server.coordinator.event.RemoveServerTagEvent;
 import org.apache.fluss.server.coordinator.event.SchemaChangeEvent;
 import org.apache.fluss.server.coordinator.event.TableRegistrationChangeEvent;
@@ -122,6 +123,7 @@ import org.apache.fluss.server.zk.data.ZkData.TableIdsZNode;
 import org.apache.fluss.server.zk.data.lake.LakeTableHelper;
 import org.apache.fluss.server.zk.data.lake.LakeTableSnapshot;
 import org.apache.fluss.utils.AutoPartitionStrategy;
+import org.apache.fluss.utils.clock.SystemClock;
 import org.apache.fluss.utils.types.Tuple2;
 
 import org.slf4j.Logger;
@@ -249,7 +251,9 @@ public class CoordinatorEventProcessor implements EventProcessor {
         this.lakeTableTieringManager = lakeTableTieringManager;
         this.coordinatorMetricGroup = coordinatorMetricGroup;
         this.internalListenerName = conf.getString(ConfigOptions.INTERNAL_LISTENER_NAME);
-        this.rebalanceManager = new RebalanceManager(this, zooKeeperClient);
+        this.rebalanceManager =
+                new RebalanceManager(
+                        this, zooKeeperClient, coordinatorEventManager, SystemClock.getInstance());
         this.ioExecutor = ioExecutor;
         this.lakeTableHelper =
                 new LakeTableHelper(zooKeeperClient, conf.getString(ConfigOptions.REMOTE_DATA_DIR));
@@ -302,6 +306,7 @@ public class CoordinatorEventProcessor implements EventProcessor {
 
         // start rebalance manager.
         rebalanceManager.startup();
+        rebalanceManager.start();
     }
 
     public void shutdown() {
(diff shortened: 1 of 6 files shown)


fluss-server/src/main/java/org/apache/fluss/server/coordinator/CoordinatorEventProcessor.java
```diff
@@ -645,6 +650,13 @@ public void process(CoordinatorEvent event) {
             completeFromCallable(
                     cancelRebalanceEvent.getRespCallback(),
                     () -> processCancelRebalance(cancelRebalanceEvent));
+        } else if (event instanceof RebalanceTaskTimeoutEvent) {
+            RebalanceTaskTimeoutEvent timeoutEvent = (RebalanceTaskTimeoutEvent) event;
+            LOG.warn(
+                    "Rebalance task for {} timed out. Treating as completed.",
+                    timeoutEvent.getTableBucket());
+            rebalanceManager.finishRebalanceTask(
+                    timeoutEvent.getTableBucket(), RebalanceStatus.COMPLETED);
```
loserwang1024:
```suggestion
                    timeoutEvent.getTableBucket(), RebalanceStatus.FAILED); or  timeoutEvent.getTableBucket(), RebalanceStatus.TIMEOUT);
```

loserwang1024 (commented):
I left a comment

fluss-server/src/main/java/org/apache/fluss/server/coordinator/rebalance/RebalanceManager.java
```diff
@@ -302,6 +367,8 @@ public RebalanceTask generateRebalanceTask(List<Goal> goalsByPriority) {
     private void processNewRebalanceTask() {
         TableBucket tableBucket = inProgressRebalanceTasksQueue.peek();
         if (tableBucket != null && inProgressRebalanceTasks.containsKey(tableBucket)) {
+            inflightTaskBucket = tableBucket;
+            inflightTaskStartMs = clock.milliseconds();
```
LiebingYu, line 382: Nit: inflightTaskBucket and inflightTaskStartMs rely on a fragile implicit write-ordering invariant

Currently this is safe because every clear path writes inflightTaskStartMs = -1 before clearing inflightTaskBucket, and processNewRebalanceTask() writes inflightTaskBucket before inflightTaskStartMs = now(). Since checkTimeout() reads inflightTaskStartMs first and returns immediately if it is -1, no dangerous interleaving can occur today.

However, this correctness depends on the write order being maintained across four separate call sites, with no compile-time enforcement. If a future change swaps the write order in any of these paths (e.g., writes inflightTaskStartMs before inflightTaskBucket in processNewRebalanceTask), the timeout thread could read a stale bucket or a stale startMs and fire a spurious timeout for the wrong task.

Consider combining both fields into a single immutable holder behind an AtomicReference to make the invariant structurally unbreakable rather than relying on write ordering:
```
private final AtomicReference<InflightTask> inflightTask = new AtomicReference<>();
```

fluss-server/src/main/java/org/apache/fluss/server/coordinator/rebalance/RebalanceManager.java
```diff
@@ -70,8 +78,17 @@
 public class RebalanceManager {
     private static final Logger LOG = LoggerFactory.getLogger(RebalanceManager.class);
 
+    /** Hardcoded timeout for an in-flight rebalance task: 2 minutes. */
+    private static final long REBALANCE_TASK_TIMEOUT_MS = 2 * 60 * 1000L;
```
LiebingYu, line 82: May be we need a new ConfigOption here? I think diferent clusters with different data volume and network confitions may have vastly different rebalance durations.

LiebingYu (commented):
@swuferhong Thanks for the PR. I have some comments.

fluss-server/src/main/java/org/apache/fluss/server/coordinator/rebalance/RebalanceManager.java
swuferhong: Good catch on the ordering fragility. Fixed by applying the volatile publication idiom — `inflightTaskBucket ` now acts as the "gate" variable: written last when setting, cleared first when resetting. The timeout checker reads it first, so JMM happens-before guarantees it always sees a consistent (bucket, startMs) pair. Added comments documenting the contract.

fluss-server/src/main/java/org/apache/fluss/server/coordinator/rebalance/RebalanceManager.java
swuferhong: This parameter will be introduced along with its dynamic update strategy in a follow-up based on requirements.

LiebingYu (approved):

PR #3431, opened 2026-06-04
[server] Fix NPE when processing table registration change

raoluSmile:
## What is the purpose of the change

  This PR fixes a `NullPointerException` in
  `CoordinatorEventProcessor.processTableRegistrationChange`.

  When processing a `TableRegistrationChangeEvent`, the coordinator needs the old
  `TableInfo` to get the current schema info and build the new `TableInfo`.
  However, the old `TableInfo` may be unavailable if the event is stale or the
  coordinator context is incomplete.

  The previous code only checked whether `tableId` was `null`, but
  `CoordinatorContext#getTableIdByPath` returns `TableInfo.UNKNOWN_TABLE_ID` when
  the table path is not found. As a result, an unknown table could still continue
  to `getTableInfoById`, get `null`, and fail at `oldTableInfo.getSchemaInfo()`.

  ## Brief change log

  - Treat `TableInfo.UNKNOWN_TABLE_ID` as an unregistered table when processing
    table registration change events.
  - Skip table registration change events if the old `TableInfo` is not available
    in `CoordinatorContext`.
  - Add regression tests for stale/incomplete coordinator context states.

  ## Tests

  Added two regression tests in `CoordinatorEventProcessorTest`:

  - `testTableRegistrationChangeWithUnknownTable`

    Covers the case where a table registration change event is processed after the
    table path is no longer present in `CoordinatorContext`.

    `CoordinatorContext#getTableIdByPath` returns `TableInfo.UNKNOWN_TABLE_ID`
    for this case. The previous `tableId == null` check did not catch it, so the
    processor continued and eventually dereferenced a null `oldTableInfo`.

  - `testTableRegistrationChangeWithMissingTableInfo`

    Covers the case where the table path to table id mapping exists, but the
    corresponding `TableInfo` is not available in `CoordinatorContext`.

    This represents an incomplete coordinator context state. Since
    `TableRegistration#toTableInfo` needs the old schema info, the event cannot be
    applied safely without the old `TableInfo`.

  Before this fix, both tests failed with an NPE at
  `oldTableInfo.getSchemaInfo()`. After this fix, the stale/incomplete events are
  skipped safely.

  ```bash
  ./mvnw -pl fluss-server -am \
    -Dtest=CoordinatorEventProcessorTest#testTableRegistrationChangeWithUnknownTable+testTableRegistrationChangeWithMissingTableInfo \
    -DfailIfNoTests=false test

  Tests run: 2, Failures: 0, Errors: 0, Skipped: 0
  BUILD SUCCESS


fluss-server/src/main/java/org/apache/fluss/server/coordinator/CoordinatorEventProcessor.java
```diff
@@ -765,6 +765,14 @@ private void processTableRegistrationChange(TableRegistrationChangeEvent event)
         }
 
         TableInfo oldTableInfo = coordinatorContext.getTableInfoById(tableId);
+        if (oldTableInfo == null) {
+            LOG.warn(
+                    "Table {} with id {} is not loaded in coordinator context, "
+                            + "skip processing table registration change.",
+                    tablePath,
+                    tableId);
+            return;
```
loserwang1024, line 925: Even if oldTableInfo is null, we should not return directly, otherwise `coordinatorContext.putTableInfo(newTableInfo);` will not be executed. 

However, I don't think we need this check here. The reason why a NullPointerException was thrown previously was because org.apache.fluss.server.coordinator.CoordinatorContext#getTableIdByPath returned TableInfo.UNKNOWN_TABLE_ID when the entry did not exist. Now that you've added a prior validation check, this issue should no longer occur.

fluss-server/src/main/java/org/apache/fluss/server/coordinator/CoordinatorEventProcessor.java
```diff
@@ -756,7 +756,7 @@ private void processTableRegistrationChange(TableRegistrationChangeEvent event)
 
         // Skip if the table is not yet registered in coordinator context.
         // Should not happen in normal cases.
-        if (tableId == null) {
+        if (tableId == null || tableId == TableInfo.UNKNOWN_TABLE_ID) {
```
loserwang1024: `tableId == TableInfo.UNKNOWN_TABLE_ID` is enough, coordinatorContext.getTableIdByPath never return null

loserwang1024 (commented):
Thanks for your pr, I have left some comment.

fluss-server/src/main/java/org/apache/fluss/server/coordinator/CoordinatorEventProcessor.java
raoluSmile: Thanks for the explanation. I agree that the missing TableInfo case is an inconsistent coordinator context state and should not be silently skipped. I’ll remove the oldTableInfo null check and keep the fix focused on UNKNOWN_TABLE_ID.

fluss-server/src/main/java/org/apache/fluss/server/coordinator/CoordinatorEventProcessor.java
raoluSmile: Thanks for pointing this out. I agree with you.

fluss-server/src/main/java/org/apache/fluss/server/coordinator/CoordinatorEventProcessor.java
```diff
@@ -756,7 +756,7 @@ private void processTableRegistrationChange(TableRegistrationChangeEvent event)
 
         // Skip if the table is not yet registered in coordinator context.
         // Should not happen in normal cases.
-        if (tableId == null) {
+        if (tableId == TableInfo.UNKNOWN_TABLE_ID) {
```
Jackeyzhe, line 910: I think this still does not cover the incomplete coordinator context case described in the PR.
After `getTableIdByPath(tablePath)` returns a valid id, `getTableInfoById(tableId)` may still return null because table path and table info are stored separately in `CoordinatorContext`. In that case, `oldTableInfo.getSchemaInfo()` will still throw an NPE.
Could we restore the `oldTableInfo == null` guard and the corresponding regression test for this case?

raoluSmile: Good point. I checked CoordinatorContext again: table path/id and table info are stored in separate maps, so a valid table id does not guarantee oldTableInfo is available. 
Since TableRegistration#toTableInfo needs the old schema info, the event cannot be applied safely without oldTableInfo. I restored the guard and added the missing-table-info regression test, while keeping the table id check focused on UNKNOWN_TABLE_ID because getTableIdByPath never returns null.

Jackeyzhe (approved):
Thanks for this PR, LGTM

raoluSmile:
please help review, thanks! @loserwang1024

PR #3432, opened 2026-06-04
[flink] Add union read support to datastream

polyzos:
The DataStream `FlussSource/FlussSourceBuilder` never created a `LakeSource`, it always passed `null` to FlinkSource. So DataStream jobs read only Fluss data, even for datalake-enabled tables. Union read worked only via the Flink SQL/Table connector.

This pr addresses this missing piece

### DIFF
--- a/fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/source/FlussSourceBuilder.java
+++ b/fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/source/FlussSourceBuilder.java
@@ -37,6 +37,7 @@ import org.apache.fluss.types.RowType;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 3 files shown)


fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/source/FlussSourceBuilder.java
```diff
@@ -324,6 +347,22 @@ public FlussSource<OUT> build() {
                         ? tableInfo.getRowType().project(projectedFields)
                         : tableInfo.getRowType();
 
+        // union read (lake historical + Fluss) only applies to full startup mode, like the SQL
+        // connector; other startup modes read Fluss only.
+        LakeSource<LakeSplit> lakeSource = null;
+        if (tableInfo.getTableConfig().isDataLakeEnabled()
+                && offsetsInitializer instanceof SnapshotOffsetsInitializer) {
+            lakeSource =
+                    LakeSourceUtils.createLakeSource(tablePath, tableInfo.getProperties().toMap());
```
fresh-borzoni, line 368: What about filters?

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/source/FlussSourceBuilder.java
```diff
@@ -324,6 +347,22 @@ public FlussSource<OUT> build() {
...
+            if (lakeSource != null && projectedFields != null) {
+                int[][] nestedProjectedFields = new int[projectedFields.length][];
+                for (int i = 0; i < projectedFields.length; i++) {
+                    nestedProjectedFields[i] = new int[] {projectedFields[i]};
+                }
+                lakeSource.withProject(nestedProjectedFields);
+            }
+        }
```
fresh-borzoni, line 383: setBounded() only works on a lake table started from full().  Other options seems that the job starts and then crashes later with a message like "lakeSource' is null in batch mode". Both bad cases just mean lakeSource came out null.

Should we check it explicitly and early?

fluss-lake/fluss-lake-iceberg/src/test/java/org/apache/fluss/lake/iceberg/flink/FlinkUnionReadDataStreamITCase.java
```diff
@@ -0,0 +1,460 @@
...
+import static org.apache.fluss.testutils.DataTestUtils.row;
+import static org.apache.fluss.testutils.common.CommonTestUtils.waitUntil;
+import static org.assertj.core.api.Assertions.assertThat;
+
+/**
+ * Integration tests for union read through the DataStream {@link FlussSource}.
+ *
+ * <p>These tests mirror the Flink SQL union-read coverage ({@code FlinkUnionReadLogTableITCase} and
+ * {@code FlinkUnionReadPrimaryKeyTableITCase}) but exercise the programmatic DataStream source.
+ * Each test asserts the three properties that make a union read meaningful:
```
fresh-borzoni, line 73: : ... mm?

fresh-borzoni (commented):
@polyzos Ty for the PR, left some minor comments, PTAL

polyzos:
@fresh-borzoni Thank you for your comments, good catch.. I updated the PR to address them

fresh-borzoni (approved):
@polyzos Ty, LGTM 👍 
Can you please file a followup about partition pruning?

polyzos:
@fresh-borzoni I opened #3452 to cover that

PR #3442, opened 2026-06-06
[flink] Gate single-row lookup pushdown on partition liveness

### LINKED ISSUE #3443: [flink] Single-row lookup pushdown returns empty for lake-only (expired) partitions
### Search before asking

### Fluss version

main (development)

### Please describe the bug 🐞

A full primary-key point query on a lake-enabled partitioned table (FULL) is rewritten to a kv single-row lookup. It reads only Fluss kv, so for a partition expired from Fluss but still in the lake it returns empty, yet the union scan(or a partial-key filter) returns the row. 

The COUNT(*) pushdown already guards this via isDataLakeEnabled and the single-row lookup doesn't.

### Solution

Gate the lookup on partition liveness, else fall back to the union scan.

fresh-borzoni:
closes https://github.com/apache/fluss/issues/3443

A full-PK point query on a lake-enabled partitioned table returns empty when the partition was expired from Fluss but still lives in the lake and the lookup pushdown skips the union scan. 
Gate the lookup on partition liveness, fall back to the scan otherwise.

### DIFF
--- a/fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/source/FlinkTableSource.java
+++ b/fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/source/FlinkTableSource.java
@@ -24,6 +24,7 @@ import org.apache.fluss.config.Configuration;
 import org.apache.fluss.config.StatisticsColumnsConfig;
 import org.apache.fluss.config.TableConfig;
 import org.apache.fluss.flink.FlinkConnectorOptions;
+import org.apache.fluss.flink.row.FlinkAsFlussRow;
 import org.apache.fluss.flink.source.deserializer.RowDataDeserializationSchema;
 import org.apache.fluss.flink.source.lookup.FlinkAsyncLookupFunction;
 import org.apache.fluss.flink.source.lookup.FlinkLookupFunction;
@@ -46,7 +47,6 @@ import org.apache.fluss.predicate.PartitionPredicateVisitor;
 import org.apache.fluss.predicate.Predicate;
 import org.apache.fluss.predicate.PredicateBuilder;
 import org.apache.fluss.predicate.PredicateVisitor;
-import org.apache.fluss.row.GenericRow;
 import org.apache.fluss.types.DataTypeChecks;
 import org.apache.fluss.types.RowType;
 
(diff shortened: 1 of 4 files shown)


fresh-borzoni:
@luoyuxia @loserwang1024 PTAL 🙏

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/source/FlinkTableSource.java
```diff
@@ -889,6 +894,36 @@ private int[] getKeyRowProjection() {
...
+    private boolean lookupCoversAllData(
+            List<ResolvedExpression> filters, Map<Integer, LogicalType> primaryKeyTypes) {
+        if (!isDataLakeEnabled || !isPartitioned()) {
+            return true;
+        }
+        return PushdownUtils.partitionExists(
+                tablePath, flussConfig, resolveLookupPartition(filters, primaryKeyTypes));
+    }
+
+    private PartitionSpec resolveLookupPartition(
```
luoyuxia: Not sure whether it can be simplify to the following:
```
private boolean lookupCoversAllData(GenericRowData lookupRow) {
        if (!isDataLakeEnabled || !isPartitioned()) {
            return true;
        }
        // lookupRow holds all primary key fields (in pk-projection order); partition keys are a
        // subset of the primary key, so the partition can be fully resolved from it.
        RowType flussRowType = FlinkConversions.toFlussRowType(tableOutputType);
        PartitionGetter partitionGetter =
                new PartitionGetter(
                        flussRowType.project(primaryKeyIndexes),
                        flussRowType.project(partitionKeyIndexes).getFieldNames());
        PartitionSpec partitionSpec =
                partitionGetter
                        .getResolvedPartitionSpec(new FlinkAsFlussRow(lookupRow))
                        .toPartitionSpec();
        return PushdownUtils.partitionExists(tablePath, flussConfig, partitionSpec);
    }
```

luoyuxia (commented):
@fresh-borzoni Thanks for the pr. lgtm overall. Only one minor comment
Also, https://cwiki.apache.org/confluence/display/FLUSS/FIP-28%3A+Improve+historical+Partition+Support+for+Fluss+Datalake-Enabled+Tables
will make lookup expire partition from paimon possible. Could you please left a todo when FIP28 is finished, we can then fall into the normal path to lookup instead of full scan?

luoyuxia (approved):
@fresh-borzoni Thanks LGTM! But seems ci fails

fresh-borzoni:
@luoyuxia Thank you, looking into CI

update: fixed the issue, will wait until green

PR #3447, opened 2026-06-07
[client] Enum conversion functionality for typed writer/reader

### LINKED ISSUE #3446: [client] Add enum conversion functionality for Typed Writer/Reader
### Search before asking

### Motivation

when you use typed writer/reader you can't use enum as a field in POJO.

### Solution
Map enum to String

- Adjust validation logic in ConverterCommons
- conversion from pojo to fluss type will be done automatically with toString()
- Add conversion logic from BinaryString to Enum to FlussTypeToPojoTypeConverter 

### Anything else?

_No response_

### Willingness to contribute

VladBanar:
- **Generative AI disclosure:** Indicate whether generative AI tools were used in authoring this PR. If yes, specify the tool below.

Generated-by: [Claude Haiku 4.5] following [the guidelines](https://github.com/apache/fluss/blob/main/AGENTS.md)
Only for tests

### Purpose

Linked issue: close [#3446](https://github.com/apache/fluss/issues/3446)

when you use typed writer/reader you can't use enum as a field in POJO.
### Brief change log

- Adjusted validation logic in ConverterCommons
- Added conversion logic for enum in FlussTypeToPojoTypeConverter:convertTextValue

### Tests

- Added ConverterCommonsTest.java FlussTypeToPojoTypeConverterTest.java

### API and Format
No

### Documentation
No


VladBanar:
@polyzos Please take a look i fixed checks.

VladBanar:
@luoyuxia @MehulBatra @fresh-borzoni  Could you please approve workflow

VladBanar:
@fresh-borzoni Could you please review it? :)

fluss-client/src/main/java/org/apache/fluss/client/converter/FlussTypeToPojoTypeConverter.java
```diff
@@ -74,6 +75,17 @@ static Object convertTextValue(
                         ConverterCommons.charLengthExceptionMessage(fieldName, v.length()));
             }
             return v.charAt(0);
+        } else if (pojoType.isEnum()) {
+            return Arrays.stream(pojoType.getEnumConstants())
```
fresh-borzoni: getEnumConstants() clones the array and scans on every row. Could build a name -> constant map once at converter creation instead.

fluss-client/src/main/java/org/apache/fluss/client/converter/FlussTypeToPojoTypeConverter.java
```diff
@@ -74,6 +75,17 @@ static Object convertTextValue(
...
+                    .filter(e -> e.toString().equals(v.toUpperCase()))
```
fresh-borzoni: Works for uppercase enums, but breaks for any enum whose constants aren't all-uppercase (write/read mismatch)

fluss-client/src/test/java/org/apache/fluss/client/converter/FlussTypeToPojoTypeConverterTest.java
```diff
@@ -0,0 +1,250 @@
+/*
```
fresh-borzoni, line 1: All the enum tests use an uppercase enum, which is the case that works. Could you add a lowercase enum with a full POJO→Row→POJO round-trip? And ideally a List<MyEnum>, so collections hit the same path

fluss-client/src/main/java/org/apache/fluss/client/converter/FlussTypeToPojoTypeConverter.java
```diff
@@ -74,6 +75,17 @@ static Object convertTextValue(
...
+                    .findAny()
+                    .orElseThrow(
+                            () ->
+                                    new IllegalArgumentException(
+                                            String.format(
+                                                    "Could not parse value for enum %s. Expected one of: [%s]",
```
fresh-borzoni: nit: Would it be double brackets here with Arrays.toString?

fresh-borzoni (commented):
@VladBanar Thank you for the PR, left some comments, PTAL

fluss-client/src/main/java/org/apache/fluss/client/converter/FlussTypeToPojoTypeConverter.java
VladBanar: Converter uses static methods ,but i created a map with lazy initialization

fluss-client/src/main/java/org/apache/fluss/client/converter/FlussTypeToPojoTypeConverter.java
VladBanar: fixed

fluss-client/src/main/java/org/apache/fluss/client/converter/FlussTypeToPojoTypeConverter.java
VladBanar: fixed

fluss-client/src/test/java/org/apache/fluss/client/converter/FlussTypeToPojoTypeConverterTest.java
VladBanar: - adjusted test to cover round-trip 
- added lowercase enum
- added test for an array

VladBanar:
@fresh-borzoni Thank you for the review! Addressed the comments PTAL

fresh-borzoni:
@VladBanar Can you please rebase/cherry-pick and resolve conflict?
Also, just noticed: do we really use `mappedName`?

VladBanar:
@fresh-borzoni Yes, you're right, mappedName isn't really needed. I added it so it would be more clear that we have this alias logic. But i can delete it if you think it's not needded.

I'll resolve the conflict as well

VladBanar:
@fresh-borzoni Fixed conflict with merge

fluss-client/src/main/java/org/apache/fluss/client/converter/FlussTypeToPojoTypeConverter.java
```diff
@@ -31,9 +31,30 @@
...
+            new ConcurrentHashMap<>();
+
+    /**
+     * Builds a map of enum name (uppercase) to enum constant for the given enum class. This map is
+     * cached to avoid recreating it for every enum conversion.
+     */
+    private static Map<String, Object> buildEnumConstantsMap(Class<?> enumClass) {
+        Map<String, Object> map = new HashMap<>();
+        for (Object constant : enumClass.getEnumConstants()) {
+            map.put(constant.toString(), constant);
```
fresh-borzoni: Should this key on name() rather than toString()? toString() is overridable and not guaranteed unique, so two constants can collapse the map and a row reads back as the wrong constant

fluss-client/src/main/java/org/apache/fluss/client/converter/FlussTypeToPojoTypeConverter.java
```diff
@@ -31,9 +31,30 @@
...
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
 
 /** Shared utilities for Fluss type and Pojo type. */
 public class FlussTypeToPojoTypeConverter {
+
+    private static final Map<Class<?>, Map<String, Object>> ENUM_CONSTANTS_CACHE =
```
fresh-borzoni: This static cache is keyed by the user's enum Class and never evicted. Under a Flink user-code classloader it pins the class (and its classloader) across job restarts, so metaspace leak.

The rest of this path builds per-field converters once in the closure. 
Could we build the name -> constant map there and capture it, instead of a static map? That was the original idea and it drops the static state entirely.

fluss-client/src/test/java/org/apache/fluss/client/converter/PojoArrayToFlussArrayTest.java
```diff
@@ -242,6 +243,12 @@ public void testArrayWithAllTypes() {
 
         assertThat(resultMap2).containsEntry("test_3", 3);
         assertThat(resultMap2).containsEntry("test_4", 4);
+        // Verify enum array
+        InternalArray enumArray = row.getArray(22);
+        assertThat(enumArray.size()).isEqualTo(3);
+        assertThat(enumArray.getString(0).toString()).isEqualTo("PENDING");
```
fresh-borzoni, line 249: This only checks the write direction. Could we add the read-back (Row -> POJO) for an enum array, and a List<MyEnum> field too? 
The collection read paths (FlussArrayToPojoArray, and map keys/values via FlussMapToPojoMap) aren't exercised for enums yet.

fluss-client/src/test/java/org/apache/fluss/client/converter/ConvertersTestFixtures.java
```diff
@@ -151,6 +155,7 @@ public static TestPojo sample() {
                             put("test_2", 2);
                         }
                     },
+                    StatusEnum.OK,
```
fresh-borzoni, line 171: The round-trip test only ever sees OK. 
Since the lowercase case is the one that was broken, could sample() use error (or add a case that does) so a lowercase enum goes through the full POJO -> Row -> POJO path?

fluss-client/src/main/java/org/apache/fluss/client/converter/ConverterCommons.java
```diff
@@ -148,6 +148,15 @@ static void validateCompatibility(DataType fieldType, PojoType.Property prop) {
             }
             return;
         }
+        if (actual.isEnum()) {
+            if (typeRoot != DataTypeRoot.STRING) {
+                throw new IllegalArgumentException(
+                        String.format(
+                                "Enum field '%s' must be a string type, got %s",
```
fresh-borzoni: Small thing: this fires for CHAR too, but CHAR is a string type (it's in SUPPORTED_TYPES), so the message reads as self-contradictory. 
Probably better wording is "STRING column", wdyt?

fresh-borzoni (commented):
@VladBanar Thank you for the changes, left a couple of comments, PTAL.

Also what is your plan for mappedName, btw?

fluss-client/src/main/java/org/apache/fluss/client/converter/FlussTypeToPojoTypeConverter.java
VladBanar: done

fluss-client/src/main/java/org/apache/fluss/client/converter/FlussTypeToPojoTypeConverter.java
VladBanar: Oh i see what you mean now. Done.

fluss-client/src/main/java/org/apache/fluss/client/converter/ConverterCommons.java
VladBanar: For me it's the same. But changed the wording.

fluss-client/src/test/java/org/apache/fluss/client/converter/ConvertersTestFixtures.java
VladBanar: added another test with lowercase enum

fluss-client/src/test/java/org/apache/fluss/client/converter/PojoArrayToFlussArrayTest.java
VladBanar: added list and map fields to TestPojo for RowToPojoConverterTest:testRoundTripFullSchema

VladBanar:
@fresh-borzoni Thank you for comments. Addressed the comments PTAL

fluss-client/src/main/java/org/apache/fluss/client/converter/ConverterCommons.java
fresh-borzoni: shall we fix write side as well then?

fresh-borzoni (commented):
@VladBanar Thank you for the changes, LGTM, one small issue remains

fluss-client/src/main/java/org/apache/fluss/client/converter/ConverterCommons.java
VladBanar: @fresh-borzoni Really appreciate thorough review. Done!

fluss-client/src/main/java/org/apache/fluss/client/converter/ConverterCommons.java
```diff
@@ -171,13 +180,17 @@ public static String charLengthExceptionMessage(String fieldName, int length) {
...
-        final String s = String.valueOf(v);
+        final String s = objectToString(v);
         if (root == DataTypeRoot.CHAR && s.length() != 1) {
             throw new IllegalArgumentException(charLengthExceptionMessage(fieldName, s.length()));
         }
         return BinaryString.fromString(s);
     }
 
+    private static String objectToString(Object v) {
+        return v != null && v.getClass().isEnum() ? ((Enum<?>) v).name() : String.valueOf(v);
```
fresh-borzoni: v.getClass().isEnum() is false for enum constants with a class body (PLUS { … } -> anonymous subclass). Those would fall back to toString() and desync from the read side again. v instanceof Enum covers all constants and is the idiomatic check.

example I used:
```java
  enum Op {
      PLUS  { public String toString() { return "+"; } },
      MINUS { public String toString() { return "-"; } };
  }
  
```

fresh-borzoni (commented):
@VladBanar Ty, looks fine, just one concern

fluss-client/src/main/java/org/apache/fluss/client/converter/ConverterCommons.java
fresh-borzoni: also instanceof Enum is also null-safe, so it lets the whole helper drop the explicit v != null && guard

VladBanar: @fresh-borzoni  done

fresh-borzoni (approved):
@VladBanar LGTM, I'll merge once CI is green

PR #3450, opened 2026-06-08
[test] Fix ComponentClassLoaderTest tmpdir path handling

### LINKED ISSUE #3437: [UT] ComponentClassLoaderTest fails when java.io.tmpdir contains dots
### Describe the bug

`ComponentClassLoaderTest.testOwnerFirstResourceFoundIgnoresComponent` can fail in CI when `java.io.tmpdir` points to a path that contains dots, for example a temporary directory containing a version-like segment such as `1.0-SNAPSHOT`.

In this case `ComponentClassLoader#getResource(...)` returns `null`, and the test fails because the loaded resource is expected to be the owner class loader resource.

### Reproduction

Run the `fluss-common` test module with a temp directory whose path contains dots, for example:

```bash
mvn -pl fluss-common -Dtest=ComponentClassLoaderTest -Djava.io.tmpdir=/tmp/fluss-1.0-SNAPSHOT test
```

### Root cause

The test uses the `@TempDir` absolute path as both the resource name and the owner-first package/resource prefix.

`ComponentClassLoader` converts package prefixes to resource path prefixes by replacing `.` with `/`. If the temp path contains a segment such as `1.0-SNAPSHOT`, it is converted to `1/0-SNAPSHOT`, so the owner-first resource prefix no longer matches the actual resource name. As a result, the owner-first resource lookup is skipped and `getResource(...)` returns `null`.

### Expected behavior

The test should not depend on the formatting of the machine-specific temporary directory path. It should use a stable package/resource prefix and create an explicit test resource under the temp directory.

### Affected test

`fluss-common/src/test/java/org/apache/fluss/classloading/ComponentClassLoaderTest.java`

litiliu:
## Summary
- Use a stable package/resource prefix in `ComponentClassLoaderTest` instead of the machine-specific temporary directory path.
- Create the test resource explicitly under the JUnit temp directory.
- Avoid failures when `java.io.tmpdir` contains dotted path segments such as `1.0-SNAPSHOT`.

## Root Cause
`ComponentClassLoader` converts package prefixes to resource path prefixes by replacing `.` with `/`. The test previously used the absolute `@TempDir` path as the prefix, so dotted path segments were rewritten and no longer matched the resource name.

Fixes #3437.

## Test Plan
- `mvn -pl fluss-common -Dtest=ComponentClassLoaderTest test`
- `mvn -pl fluss-common test`

AI-assisted changes - reviewed by human developer

### DIFF
--- a/fluss-common/src/test/java/org/apache/fluss/classloading/ComponentClassLoaderTest.java
+++ b/fluss-common/src/test/java/org/apache/fluss/classloading/ComponentClassLoaderTest.java
@@ -25,6 +25,7 @@ import java.io.IOException;
 import java.net.MalformedURLException;
 import java.net.URL;
 import java.net.URLClassLoader;
+import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.util.Collections;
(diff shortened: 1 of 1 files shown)


luoyuxia (approved):
@litiliu Thanks. +1

PR #3456, opened 2026-06-08
[lake/hudi] Introduce Hudi source split planner

fhan688:
### Purpose

  Linked issue: https://github.com/apache/fluss/issues/3276

  This PR introduces the first planner-only Hudi lake source implementation for Fluss.

  Currently, the Hudi lake storage module supports basic Hudi catalog integration, but
  `HudiLakeStorage#createLakeSource` does not return a usable lake source. As a result, Fluss cannot
  plan readable lake splits for data already stored in Hudi.

  This change adds Hudi source split planning based on a completed Hudi instant. Record reading, limit
  pushdown, and Hudi tiering writer support remain explicitly unsupported and can be implemented in
  follow-up PRs.

### Brief change log

  - Add `HudiLakeSource`, `HudiSplit`, `HudiSplitPlanner`, and `HudiSplitSerializer`.
  - Wire `HudiLakeStorage#createLakeSource` to return a Hudi lake source.
  - Add `HudiTableInfo` to resolve Hudi catalog table metadata, meta client, completed timeline,
    filesystem view, table type, partition fields, and bucket-aware metadata.
  - Plan Hudi splits from the requested snapshot instant after validating that the instant exists in
    the completed Hudi timeline.
  - Support split planning for:
    - COW tables through latest base files before or on the requested instant.
    - MOR tables through latest merged file slices before or on the requested instant.
  - Persist Fluss bucket and partition metadata into Hudi table properties for planner-side recovery.
  - Return bucket `-1` for bucket-unaware Fluss log tables.
  - Add and adjust UT coverage for:
    - Hudi lake source planner/serializer wiring.
    - Explicit unsupported limit and record reader behavior.
    - Hudi split serialization version handling.
    - Hudi split planner planning and missing instant behavior.
    - Fluss bucket and partition metadata persisted in Hudi table properties.
    - Partition value extraction from Hudi partition paths.

### Tests

  - `mvn -pl fluss-lake/fluss-lake-hudi -am -DskipITs -Dcheckstyle.skip=true -DfailIfNoTests=false -Dsurefire.failIfNoSpecifiedTests=false
  -Dtest=HudiLakeSourceTest,HudiSplitSerializerTest,HudiTableInfoTest,HudiConversionsTest,HudiSplitPlannerTest test`
  - `mvn -pl fluss-lake/fluss-lake-hudi -am -DskipITs -Dcheckstyle.skip=true -DfailIfNoTests=false -Dsurefire.failIfNoSpecifiedTests=false test`

  The Hudi module test run passed with 54 tests and 0 failures.

### API and Format

  This PR does not change public Fluss APIs or Fluss storage format.

  It adds internal Hudi table properties to preserve Fluss metadata for Hudi source split planning:

  - `fluss.bucket.keys`
  - `fluss.bucket-aware`
  - `fluss.partition.keys`

### Documentation

  No user-facing documentation is added in this PR because this is a planner-only foundation. Hudi
  record reading and Hudi tiering writer support are still not exposed as completed user workflows.


luoyuxia:
@XuQianJin-Stars Could you please help review this one?

fluss-lake/fluss-lake-hudi/src/main/java/org/apache/fluss/lake/hudi/source/HudiSplitPlanner.java
```diff
@@ -0,0 +1,157 @@
...
+                throw new IOException(
+                        String.format(
+                                "Hudi instant time %s does not exist in table %s.",
+                                snapshotTime, tablePath));
+            }
+
+            List<String> partitionPaths =
+                    FSUtils.getAllPartitionPaths(
+                            hudiTableInfo.getEngineContext(), hudiTableInfo.getMetaClient(), false);
+            if (partitionPaths.isEmpty()) {
```
XuQianJin-Stars, line 72: `FSUtils.getAllPartitionPaths(...)` returns an empty list for a **partitioned** table that has no data written yet. With this fallback we then ask `getLatestMergedFileSlicesBeforeOrOn("")`, which is meaningless on a partitioned layout. In addition, `HudiTableInfo#partitionValues("")` returns `emptyList()` for that case, so the resulting `HudiSplit.partition()` has a length different from `partitionFields.size()` and downstream consumers that rely on `partition().size() == partitionColumns.size()` will silently misalign.

**Suggestion:** Only fall back to `""` when the table is **non-partitioned** (`partitionFields.isEmpty()`). For a partitioned table with no discovered partitions, return an empty split list and log at debug.

fluss-lake/fluss-lake-hudi/src/main/java/org/apache/fluss/lake/hudi/source/HudiSplitSerializer.java
```diff
@@ -0,0 +1,51 @@
...
+
+    private static final int CURRENT_VERSION = 1;
+
+    @Override
+    public int getVersion() {
+        return CURRENT_VERSION;
+    }
+
+    @Override
+    public byte[] serialize(HudiSplit hudiSplit) throws IOException {
```
XuQianJin-Stars, line 36: 1. `InstantiationUtils.deserializeObject` returns `Object` with **no type check** — a crafted byte stream can deserialize to arbitrary classes on the classpath. This violates the project security rule "Deserialization: always not trust, use safe load method."
2. Hudi internal types like `FileSlice` / `HoodieBaseFile` do not guarantee a stable `serialVersionUID` across versions, so this is also a forward-compatibility risk.

**Minimum fix:** cast and validate the result, throwing `IOException` on type mismatch.

**Preferred fix:** follow `PaimonSplitSerializer`'s field-by-field serialization — serialize `fileSlice` via Hudi's own Avro/JSON serializer (or its individual fields), and write `bucket` / `partition` explicitly. Don't bind our wire format to Hudi's internal Java serialization layout.

fluss-lake/fluss-lake-hudi/src/main/java/org/apache/fluss/lake/hudi/utils/HudiTableInfo.java
```diff
@@ -0,0 +1,338 @@
...
+        return bucketAware;
+    }
+
+    public List<String> partitionValues(String partitionPath) {
+        return extractPartitionValues(partitionPath, partitionFields);
+    }
+
+    @Override
+    public void close() {
+        fileSystemView.close();
```
XuQianJin-Stars: `close()` should be idempotent and must release **all** resources even if one of them throws. Currently:
1. If `fileSystemView.close()` throws, `closeCatalog(hudiCatalog)` is skipped → catalog leak.
2. Calling `close()` twice will throw on the second invocation.

fluss-lake/fluss-lake-hudi/src/main/java/org/apache/fluss/lake/hudi/source/HudiSplitPlanner.java
```diff
@@ -0,0 +1,157 @@
...
+            for (String partitionPath : partitionPaths) {
+                splits.addAll(planPartition(hudiTableInfo, snapshotTime, partitionPath));
+            }
+            LOG.debug(
+                    "Planned {} Hudi splits for table {} at instant {}.",
+                    splits.size(),
+                    tablePath,
+                    snapshotTime);
+            return splits;
+        } catch (IOException e) {
```
XuQianJin-Stars: The first catch is redundant, and the second one converts every `RuntimeException` (e.g. `IllegalStateException`, `HoodieException`) into an `IOException`, hiding the original type and making operational diagnosis harder.

fluss-lake/fluss-lake-hudi/src/main/java/org/apache/fluss/lake/hudi/source/HudiSplitPlanner.java
```diff
@@ -0,0 +1,157 @@
...
+
+    public HudiSplitPlanner(Configuration hudiConfig, TablePath tablePath, long snapshotId) {
+        this.hudiConfig = hudiConfig;
+        this.tablePath = tablePath;
+        this.snapshotId = snapshotId;
+    }
+
+    @Override
+    public List<HudiSplit> plan() throws IOException {
+        String snapshotTime = String.valueOf(snapshotId);
```
XuQianJin-Stars: Hudi instant time is a string (canonical format is a 17-digit timestamp like `20260608010101000`). Round-tripping it through a `long` and then `String.valueOf` will drop any leading zeros. Today, Hudi writes don't normally produce leading-zero instants, but repaired or imported tables can. Note also that `LakeSource.PlannerContext#snapshotId()` returning `long` is itself a lossy carrier for Hudi.
- Document the constraint that Hudi `snapshotId`s must not have leading zeros.
- Or use `String.format("%017d", snapshotId)` to enforce the 17-digit instant format.
- Add a "leading-zero instant" case to `HudiSplitPlannerTest`.

fluss-lake/fluss-lake-hudi/src/main/java/org/apache/fluss/lake/hudi/source/HudiSplit.java
```diff
@@ -0,0 +1,91 @@
...
+    public List<String> partition() {
+        return partition;
+    }
+
+    public FileSlice getFileSlice() {
+        return fileSlice;
+    }
+
+    @Override
+    public boolean equals(Object o) {
```
XuQianJin-Stars, line 64: `FileSlice#equals/hashCode` only compare `HoodieFileGroupId + baseInstantTime` (log files are not considered). If Hudi changes that contract in a future release, the equality semantics of `HudiSplit` will silently shift.

**Suggestion:** either document this dependency in the class javadoc, or implement equality explicitly using `fileGroupId + baseInstantTime + baseFile.path + logFiles.paths`. OK to defer to a follow-up PR if you prefer.

fluss-lake/fluss-lake-hudi/src/main/java/org/apache/fluss/lake/hudi/utils/HudiTableInfo.java
```diff
@@ -0,0 +1,338 @@
...
+        try {
+            CatalogBaseTable hudiTable = hudiCatalog.getTable(toHudiObjectPath(tablePath));
+            Map<String, String> tableOptions = new HashMap<>(hudiTable.getOptions());
+            tableOptions.putAll(hudiConfig.toMap());
+
+            String basePath = resolveBasePath(tablePath, tableOptions);
+            tableOptions.put(FlinkOptions.PATH.key(), basePath);
+
+            HoodieTableMetaClient metaClient = createMetaClient(basePath, hudiConfig);
+            HoodieTimeline completedTimeline =
```
XuQianJin-Stars, line 113: `filterCompletedAndCompactionInstants()` returns completed commits **plus** REQUESTED / INFLIGHT compaction instants. So:

- The local variable name `completedTimeline` is misleading.
- More importantly, `HudiSplitPlanner#containsInstant(snapshotTime)` can match a **non-completed** compaction instant, and `getLatestMergedFileSlicesBeforeOrOn` at such an instant has undefined semantics.

The PR description says "completed Hudi timeline" but the code does not match that.
- Use `filterCompletedInstants()` if you only want completed commits.
- Or rename the variable to `commitsAndCompactionTimeline` and explicitly handle the inflight-compaction case.

fluss-lake/fluss-lake-hudi/src/main/java/org/apache/fluss/lake/hudi/utils/HudiTableInfo.java
```diff
@@ -0,0 +1,338 @@
...
+                    Collections.unmodifiableMap(new HashMap<>(tableOptions)),
+                    metaClient,
+                    engineContext,
+                    completedTimeline,
+                    fileSystemView,
+                    tableType,
+                    basePath,
+                    Collections.unmodifiableList(new ArrayList<>(partitionFields)),
+                    bucketAware);
+        } catch (Exception e) {
```
XuQianJin-Stars, line 150: If `HoodieTableFileSystemView` has already been constructed and a subsequent step (e.g. `splitCommaSeparated(...)` / `resolveBucketAware(...)`) throws, the FSView is leaked.

**Suggestion:** hold `fsView` in a local and `IOUtils.closeQuietly` it in the catch as well.

fluss-lake/fluss-lake-hudi/src/main/java/org/apache/fluss/lake/hudi/HudiLakeStorage.java
XuQianJin-Stars: `createLakeSource` is implemented in this PR, so calling the storage "a scaffold" is no longer accurate. Please rephrase along the lines of:

> "Hudi lake tiering writer is not implemented yet (tracking <issue/PR link>)."

so users can locate the tracking issue.

fluss-lake/fluss-lake-hudi/src/main/java/org/apache/fluss/lake/hudi/source/HudiSplitPlanner.java
```diff
@@ -0,0 +1,157 @@
...
+                            hudiTableInfo.getEngineContext(), hudiTableInfo.getMetaClient(), false);
+            if (partitionPaths.isEmpty()) {
+                partitionPaths = Collections.singletonList("");
+            }
+
+            List<HudiSplit> splits = new ArrayList<>();
+            for (String partitionPath : partitionPaths) {
+                splits.addAll(planPartition(hudiTableInfo, snapshotTime, partitionPath));
+            }
+            LOG.debug(
```
XuQianJin-Stars: Only logging on the success path makes "why is my query empty?" investigations harder. Please log at `info`/`warn` when `splits.isEmpty()` (especially after a successful `containsInstant` check), since that usually signals a filter or partition mismatch.

fluss-lake/fluss-lake-hudi/src/main/java/org/apache/fluss/lake/hudi/utils/HudiTableInfo.java
```diff
@@ -0,0 +1,338 @@
...
+    }
+
+    @VisibleForTesting
+    public static List<String> extractPartitionValues(
+            String partitionPath, List<String> partitionFields) {
+        if (partitionFields.isEmpty() || partitionPath == null || partitionPath.isEmpty()) {
+            return Collections.emptyList();
+        }
+
+        String[] pathSegments = partitionPath.split("/");
```
XuQianJin-Stars, line 213: The current tests cover "unknown field", "duplicate field" and "mixed style", but not the case where the number of segments equals `partitionFields.size()` yet the field names don't match (e.g. `dt=20260608/dt=20260609`). The implementation correctly throws via the `null` check, but please add a unit test for this edge case.

fluss-lake/fluss-lake-hudi/src/main/java/org/apache/fluss/lake/hudi/utils/HudiConversions.java
XuQianJin-Stars: Trivial fix: `buket` → `bucket`.

fluss-lake/fluss-lake-hudi/src/main/java/org/apache/fluss/lake/hudi/source/HudiSplitPlanner.java
fhan688: > The first catch is redundant, and the second one converts every `RuntimeException` (e.g. `IllegalStateException`, `HoodieException`) into an `IOException`, hiding the original type and making operational diagnosis harder.

Thanks for pointing this out. I removed the redundant catch blocks `in HudiSplitPlanner#plan()`. `IOException` is now propagated directly by the method signature, and unchecked Hudi/runtime exceptions are no longer wrapped into `IOException`, so their original types are preserved for diagnosis.

fluss-lake/fluss-lake-hudi/src/main/java/org/apache/fluss/lake/hudi/source/HudiSplitPlanner.java
fhan688: > Hudi instant time is a string (canonical format is a 17-digit timestamp like `20260608010101000`). Round-tripping it through a `long` and then `String.valueOf` will drop any leading zeros. Today, Hudi writes don't normally produce leading-zero instants, but repaired or imported tables can. Note also that `LakeSource.PlannerContext#snapshotId()` returning `long` is itself a lossy carrier for Hudi.
> 
> * Document the constraint that Hudi `snapshotId`s must not have leading zeros.
> * Or use `String.format("%017d", snapshotId)` to enforce the 17-digit instant format.
> * Add a "leading-zero instant" case to `HudiSplitPlannerTest`.

Good point. Since `LakeSource.PlannerContext#snapshotId()` currently carries the lake snapshot as long, this PR cannot fully preserve arbitrary Hudi instant strings without changing the common source API. As a low-risk fix, I now format the snapshot id as a canonical 17-digit Hudi instant via `String.format(Locale.ROOT, "%017d", snapshotId)`, and added a planner test covering a leading-zero 17-digit instant. If we need to support non-canonical/imported instant strings beyond this, I think the follow-up should extend the lake snapshot carrier to preserve the original string value.

fluss-lake/fluss-lake-hudi/src/main/java/org/apache/fluss/lake/hudi/source/HudiSplit.java
fhan688: > `FileSlice#equals/hashCode` only compare `HoodieFileGroupId + baseInstantTime` (log files are not considered). If Hudi changes that contract in a future release, the equality semantics of `HudiSplit` will silently shift.
> 
> **Suggestion:** either document this dependency in the class javadoc, or implement equality explicitly using `fileGroupId + baseInstantTime + baseFile.path + logFiles.paths`. OK to defer to a follow-up PR if you prefer.

Agreed. I updated `HudiSplit#equals/hashCode` to avoid depending on Hudi's `FileSlice#equals/hashCode` contract. The split now compares stable slice identity fields explicitly: fileGroupId, baseInstantTime, base file path, log file paths, bucket, and partition. I also added a unit test to cover base/log file differences.

fluss-lake/fluss-lake-hudi/src/main/java/org/apache/fluss/lake/hudi/source/HudiSplitPlanner.java
fhan688: > Only logging on the success path makes "why is my query empty?" investigations harder. Please log at `info`/`warn` when `splits.isEmpty()` (especially after a successful `containsInstant` check), since that usually signals a filter or partition mismatch.

Added info logs when planning succeeds but produces no Hudi splits. The partitioned-table/no-discovered-partition detail remains at debug, while the final empty planning result is visible at info for easier investigation

fluss-lake/fluss-lake-hudi/src/main/java/org/apache/fluss/lake/hudi/source/HudiSplitPlanner.java
```diff
@@ -0,0 +1,180 @@
...
+    static String toHudiInstantTime(long snapshotId) {
+        return String.format(Locale.ROOT, "%017d", snapshotId);
+    }
+
+    private List<HudiSplit> planPartition(
+            HudiTableInfo hudiTableInfo, String snapshotTime, String partitionPath)
+            throws IOException {
+        HoodieTableFileSystemView fileSystemView = hudiTableInfo.getFileSystemView();
+        List<HudiSplit> splits = new ArrayList<>();
+        if (hudiTableInfo.getTableType() == HoodieTableType.MERGE_ON_READ) {
```
XuQianJin-Stars, line 116: The MOR branch uses `getLatestMergedFileSlicesBeforeOrOn`, the COW branch uses `getLatestBaseFilesBeforeOrOn`. That's correct for COW. But the planner only dispatches on `tableType == MERGE_ON_READ`; everything else falls into the COW branch.

Question: is **everything that is not exactly `MERGE_ON_READ` guaranteed to be COW** by Hudi's contract? `HoodieTableType` only has these two today, but a defensive `default → throw new IOException("Unsupported Hudi table type: " + ...)` would future-proof this and avoids silently dropping log files if a new type is added. Suggest:

```java
if (tableType == HoodieTableType.MERGE_ON_READ) { ... }
else if (tableType == HoodieTableType.COPY_ON_WRITE) { ... }
else { throw new IOException("Unsupported Hudi table type: " + tableType); }
```

fluss-lake/fluss-lake-hudi/src/main/java/org/apache/fluss/lake/hudi/source/HudiSplitPlanner.java
```diff
@@ -0,0 +1,180 @@
...
+                            "No Hudi partition paths discovered for partitioned table {} at instant {}.",
+                            tablePath,
+                            snapshotTime);
+                    LOG.info(
+                            "Planned no Hudi splits for partitioned table {} at instant {} because no Hudi partition paths were discovered.",
+                            tablePath,
+                            snapshotTime);
+                    return Collections.emptyList();
+                } else {
+                    partitionPaths = Collections.singletonList("");
```
XuQianJin-Stars, line 81: - `L74–L77` logs at DEBUG and `L78–L81 `immediately logs almost the same content at INFO. The DEBUG line is dead weight — remove it, or merge them.

- For a partitioned table where partitions truly haven't been created yet, returning empty splits silently is fine, but consider whether the caller (Flink/Spark) actually expects this vs. an IOException. At minimum, the INFO log should also include the table type so operators can tell "no partitions yet" from a misconfiguration.

XuQianJin-Stars (approved):
+1

luoyuxia (approved):
@fhan688 Thanks for the pr. @XuQianJin-Stars Thanks for reviewing. LGTM!

PR #3458, opened 2026-06-09
[server] Allow altering datalake auto-compaction before lake enabled

wzx140:
## Summary
- Allow table.datalake.auto-compaction to be altered.
- Add coordinator IT coverage for the allowed alter paths

### DIFF
--- a/fluss-server/src/main/java/org/apache/fluss/server/utils/TableDescriptorValidation.java
+++ b/fluss-server/src/main/java/org/apache/fluss/server/utils/TableDescriptorValidation.java
@@ -179,6 +179,15 @@ public class TableDescriptorValidation {
                             ConfigOptions.TABLE_KV_STANDBY_REPLICA_ENABLED.key()));
         }
 
+        if (tableKeysToChange.contains(ConfigOptions.TABLE_DATALAKE_AUTO_COMPACTION.key())
+                && currentConfig.isDataLakeEnabled()) {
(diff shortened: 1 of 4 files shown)


fluss-server/src/main/java/org/apache/fluss/server/utils/TableDescriptorValidation.java
```diff
@@ -179,6 +179,15 @@ public static void validateAlterTableProperties(
                             ConfigOptions.TABLE_KV_STANDBY_REPLICA_ENABLED.key()));
         }
 
+        if (tableKeysToChange.contains(ConfigOptions.TABLE_DATALAKE_AUTO_COMPACTION.key())
+                && currentConfig.isDataLakeEnabled()) {
```
luoyuxia: why reject this when datalake is enabled? IIUC, we can change this no matter datalake enable or not.

luoyuxia (commented):
@wzx140 Thanks for the pr. Left one minor comment. PTAL

fluss-server/src/main/java/org/apache/fluss/server/utils/TableDescriptorValidation.java
wzx140: I don’t see a valid reason to block this when datalake is enabled. I’ll update the logic and remove the restriction.

luoyuxia (approved):
+1

PR #3459, opened 2026-06-09
[client][server] Support adding aggregation columns

wzx140:
## Summary
- Support Java API `TableChange.addColumn` interfaces to pass aggregation functions when altering table schema.
- Carry the aggregation function through ALTER TABLE ADD COLUMN RPC so the server preserves it in the new schema.
- Validate aggregation functions on schema alter for aggregation merge engine tables and reject them on non-aggregation tables.

### DIFF
--- a/fluss-client/src/test/java/org/apache/fluss/client/admin/FlussAdminITCase.java
+++ b/fluss-client/src/test/java/org/apache/fluss/client/admin/FlussAdminITCase.java
@@ -120,6 +120,7 @@ import static org.apache.fluss.metadata.DataLakeFormat.PAIMON;
 import static org.apache.fluss.record.TestData.DATA1_PARTITIONED_TABLE_DESCRIPTOR;
 import static org.apache.fluss.record.TestData.DATA1_SCHEMA;
 import static org.apache.fluss.testutils.DataTestUtils.row;
+import static org.apache.fluss.testutils.InternalRowAssert.assertThatRow;
 import static org.apache.fluss.testutils.common.CommonTestUtils.waitUntil;
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.assertj.core.api.Assertions.assertThatThrownBy;
@@ -600,6 +601,17 @@ class FlussAdminITCase extends ClientToServerITCaseBase {
 
         SchemaInfo schemaInfo = admin.getTableSchema(tablePath).get();
         assertThat(schemaInfo.getSchema().getAggFunction("new_value")).hasValue(AggFunctions.SUM());
+
+        try (Table table = conn.getTable(tablePath)) {
+            UpsertWriter upsertWriter = table.newUpsert().createWriter();
+            upsertWriter.upsert(row(1, 10L, 100L));
+            upsertWriter.upsert(row(1, 20L, 200L));
+            upsertWriter.flush();
+
+            assertThatRow(lookupRow(table.newLookup().createLookuper(), row(1)))
+                    .withSchema(schemaInfo.getSchema().getRowType())
+                    .isEqualTo(row(1, 30L, 300L));
+        }
     }
 
     @Test
@@ -731,6 +743,12 @@ class FlussAdminITCase extends ClientToServerITCaseBase {
         TableInfo tableInfoAggregate = admin.getTableInfo(tablePathAggregate).join();
         assertThat(tableInfoAggregate.getTableConfig().getDeleteBehavior())
                 .hasValue(DeleteBehavior.IGNORE);
+        assertThat(tableInfoAggregate.getSchema().getAggFunction("count"))
+                .hasValue(AggFunctions.SUM());
+        assertThat(tableInfoAggregate.getProperties().toMap())
+                .doesNotContainKey("fields.count.agg");
+        assertThat(tableInfoAggregate.getCustomProperties().toMap())
+                .doesNotContainKey("fields.count.agg");
 
         // Test 2.6: AGGREGATION merge engine with delete behavior explicitly set to ALLOW - should
         // be allowed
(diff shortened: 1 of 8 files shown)


fluss-server/src/main/java/org/apache/fluss/server/utils/TableDescriptorValidation.java
```diff
@@ -323,6 +323,9 @@ private static void checkArrowCompression(Configuration tableConf) {
     private static void checkMergeEngine(
             Configuration tableConf, boolean hasPrimaryKey, Schema schema) {
         MergeEngineType mergeEngine = tableConf.get(ConfigOptions.TABLE_MERGE_ENGINE);
+        if (mergeEngine != MergeEngineType.AGGREGATION) {
+            validateNoAggregationFunctions(schema);
+        }
```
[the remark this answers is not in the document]
wzx140: Even when mergeEngine == null, the schema should not define aggregation functions

fluss-server/src/main/java/org/apache/fluss/server/utils/ServerRpcMessageUtils.java
```diff
@@ -355,17 +359,33 @@ public static List<TableChange> toAddColumns(List<PbAddColumn> addColumns) {
...
     }
 
+    private static AggFunction deserializeAggFunction(byte[] serializedAggFunction) {
+        try {
+            return InstantiationUtils.deserializeObject(
+                    serializedAggFunction, AggFunction.class.getClassLoader());
+        } catch (IOException | ClassNotFoundException e) {
+            throw new FlussRuntimeException("Failed to deserialize aggregation function.", e);
+        }
+    }
```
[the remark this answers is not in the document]
wzx140: Fixed

fluss-server/src/main/java/org/apache/fluss/server/utils/TableDescriptorValidation.java
```diff
@@ -377,6 +380,20 @@ private static void checkMergeEngine(
         }
     }
 
+    /** Validates that the schema doesn't contain any aggregation functions. */
+    public static void validateNoAggregationFunctions(Schema schema) {
```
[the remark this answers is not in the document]
wzx140: Fixed.

wzx140:
@luoyuxia Could you please help review this one?

fluss-common/src/main/java/org/apache/fluss/metadata/TableChange.java
```diff
@@ -269,6 +294,8 @@ public String toString() {
                     + ", comment='"
                     + comment
                     + '\''
+                    + ", aggFunction="
```
luoyuxia: nit: if aggFunction is null, we don't need to print "aggFunction=null"

fluss-server/src/main/java/org/apache/fluss/server/coordinator/MetadataManager.java
```diff
@@ -440,6 +441,7 @@ public void alterTableSchema(
             if (!schemaChanges.isEmpty()) {
                 Schema newSchema =
```
luoyuxia, line 442: I think we should also updat table option to reflect the new added merge coumn. 
Like 
```
CREATE TABLE test_last_value  (
    id BIGINT,
    status STRING,
    last_login TIMESTAMP(3),
    PRIMARY KEY (id) NOT ENFORCED
) WITH (
    'table.merge-engine' = 'aggregation',
    **'fields.status.agg' = 'last_value',**
    **'fields.last_login.agg' = 'last_value'**
);
```

fluss-client/src/test/java/org/apache/fluss/client/admin/FlussAdminITCase.java
```diff
@@ -568,6 +568,62 @@ void testAlterTableColumn() throws Exception {
                 .hasMessageContaining("Column nested_row already exists");
     }
 
+    @Test
+    void testAlterAggregationTableColumnWithAggFunction() throws Exception {
```
luoyuxia, line 573: I think we had better to verify the merge works for new coulmn, may need to write rows to verify it.

luoyuxia (commented):
@wzx140 Thanks for the pr. Left minor comments. PTAL

fluss-server/src/main/java/org/apache/fluss/server/coordinator/MetadataManager.java
wzx140: @luoyuxia  Let me summarize the current situation a bit, since this part is somewhat subtle.

I think `fields.*.agg` should be treated as the Flink SQL option representation of column aggregation functions, rather than as the authoritative storage in Fluss table properties.

For Flink SQL `CREATE TABLE`, these options are parsed and converted into the column-level `AggFunction` in the Fluss schema. For Flink SQL `SHOW CREATE TABLE`, the options can be reconstructed from the column-level aggregation metadata for display.

The actual source of truth in Fluss metadata should remain the schema column `AggFunction`. Persisting the same aggregation information again in table properties would duplicate the metadata in two places and could introduce consistency issues.

This is also consistent with the Java API behavior: when a table is created with column-level `AggFunction`s through the Java API, `TableInfo#getProperties()` / `getCustomProperties()` do not contain `fields.*.agg`; the aggregation metadata is stored on the schema columns.

fluss-common/src/main/java/org/apache/fluss/metadata/TableChange.java
wzx140: done

fluss-client/src/test/java/org/apache/fluss/client/admin/FlussAdminITCase.java
wzx140: done

luoyuxia (approved):
@wzx140 Thanks. LGTM. @platinumhamburg Would you like to take another look? I’m not sure whether adding columns via ALTER TABLE has any corner cases for agg merge engine tables.

platinumhamburg (approved):
LGTM.

PR #3463, opened 2026-06-10
[server][client] Add cooperative KV backpressure based on RocksDB L0

### LINKED ISSUE #3454: Backpressure on KV Operations Instead of Cluster-Wide Stalling
### Search before asking

### Motivation

Primary key table read/write operations involving RocksDB are currently executed on Netty worker threads while holding the bucket lock. When RocksDB slows down — due to disk I/O bottlenecks, compaction pressure, or compute saturation — the bucket lock is held for extended periods. This causes a cascading failure pattern:

1. RocksDB latency increases.
2. Netty worker threads block on the bucket lock waiting for the slow RocksDB operation to complete.
3. RPC requests of **all types** (not just KV writes) pile up in the queue because worker threads are exhausted.
4. The cluster enters a severe "stalling oscillation" — request timeouts cascade, healthy buckets become unreachable, and overall cluster health degrades far beyond the scope of the original slowdown.

The core issue: a single slow bucket's RocksDB degradation propagates into a cluster-wide availability problem.

### Solution

Introduce a **cooperative two-tier backpressure model** between the tablet server and the writer client, driven by the column family's RocksDB L0 file count. The goal is to let the client absorb pressure via per-bucket throttling, and to fail fast at the server before RocksDB's own write stall is allowed to block the RPC handler thread.

**Tier 1 — Piggyback throttle (soft, proactive)**

- On each `PutKv` request the server reads the column family's L0 file count once via `RocksDB#getLongProperty` (no per-call string allocation).
- When L0 reaches `kv.backpressure.l0-slowdown-trigger` (default `12`), the server piggybacks a normalized pressure value `p ∈ [0, 1)` on `PbPutKvRespForBucket`, where `p` linearly maps the gap between the Fluss trigger and RocksDB's own `level0_slowdown_writes_trigger`.
- The client throttles the affected bucket for `client.writer.kv-backpressure.max-throttle-ms * p³` (default ceiling `5000ms`). The cubic mapping gives a smooth ramp-up: small delays at low pressure, sharply growing delays as L0 approaches the storage trigger.

**Tier 2 — Hard rejection (fast-fail, fallback)**

- When L0 reaches RocksDB's `level0_slowdown_writes_trigger`, the server throws `StorageBackpressureException` (a `RetriableException`, Errors code `71`) instead of submitting the write.
- This short-circuits the request so the RPC handler thread is never blocked by RocksDB's internal sleep.
- The client treats this as the maximum throttle window and lets the standard retry path re-enqueue the batch, making the transition between throttle and rejection continuous.

**Why client-cooperative throttling rather than async execution alone**

- Pushing the work onto an internal server-side queue without a feedback signal only relocates the queue — it does not reduce inbound pressure. Cooperative throttling slows producers before the queue forms.
- Storage-side hard rejection still bounds the queue depth and protects the Netty worker pool against the worst case.

### Anything else?

_No response_

### Willingness to contribute

platinumhamburg:
Introduce a two-tier backpressure model to keep RocksDB write stall from blocking RPC handler threads:

- Tier 1 (piggyback): server reports a normalized [0, 1) pressure on `PbPutKvRespForBucket` once L0 reaches `kv.backpressure.l0-slowdown-trigger` (default `8`); client throttles per bucket with `max_throttle * p^2` via `client.writer.kv-backpressure.max-throttle` (default `3s`).
- Tier 2 (hard rejection): once L0 reaches RocksDB's level0_slowdown_writes_trigger, server throws StorageBackpressureException (Errors code 71, RetriableException); client applies the same max-throttle window as backoff before retry.

Add maintenance configuration and observability docs for the new server/client options and metrics.

### Purpose

Linked issue: close #3454 3454


website/docs/maintenance/configuration.md
```diff
@@ -174,6 +174,9 @@ during the Fluss cluster working.
...
+| kv.flush.thread-num                              | Integer    | 10                            | The number of threads that the server uses to asynchronously flush primary-key table writes to RocksDB. The default value is 10. |
+| client.writer.kv-backpressure.max-throttle       | Duration   | 3s                            | The upper bound of the per-bucket throttle delay the client applies in response to KV backpressure signals piggybacked by the server. Throttle delay is computed as: `max_throttle * p^2`, where `p` is the normalized pressure in `[0, 1)` reported by the server. Once the server rejects a write with `StorageBackpressureException`, the client also uses this value as the initial retry backoff so that the transition between throttle and rejection is continuous. The default value is 3 seconds. |
```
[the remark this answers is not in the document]
platinumhamburg: PR description updated.

fluss-common/src/main/java/org/apache/fluss/exception/StorageBackpressureException.java
```diff
@@ -0,0 +1,42 @@
...
+ * under, RocksDB write pressure.
+ *
+ * <p>This is the second tier of the cooperative backpressure model: the first tier is the proactive
+ * client-side throttle (driven by per-bucket pressure piggybacked on responses); once the storage
+ * engine itself is about to enter internal write delay or rejects a no-slowdown write, the server
+ * short-circuits the write and returns this retriable exception so that request processing threads
+ * are not blocked by the storage engine's internal sleep.
+ *
+ * <p>Clients should retry after a backoff equal to the configured throttle ceiling.
+ */
```
luoyuxia, line 35: nit: add since 1.0?

fluss-server/src/main/java/org/apache/fluss/server/kv/prewrite/KvPreWriteBuffer.java
```diff
@@ -194,66 +200,112 @@ public void truncateTo(long targetLogSequenceNumber, TruncateReason truncateReas
...
+                entry.state = EntryState.ACTIVE;
+            }
+        }
+    }
+
+    private static long entryBytes(Key key, Value value) {
+        return key.key.length + (value.value != null ? value.value.length : 0);
+    }
+
+    private static KvEntry previousEntryInBuffer(@Nullable KvEntry entry) {
```
luoyuxia: nit: I'm wondering whether the following is more easy to understand(I did spend some time to understand why the loop is needed):
```
private static @Nullable KvEntry previousEntryInBuffer(@Nullable KvEntry entry) {
      return entry != null && entry.state != EntryState.FLUSHED ? entry : null;
  }
```

fluss-server/src/main/java/org/apache/fluss/server/kv/prewrite/KvPreWriteBuffer.java
luoyuxia: seems getTruncateAsDuplicatedCount and getTruncateAsErrorCount are no longer needed, can we remove them?

fluss-server/src/main/java/org/apache/fluss/server/kv/KvTablet.java
```diff
@@ -667,42 +790,268 @@ private WalBuilder createWalBuilder(int schemaId, RowType rowType) throws Except
...
     }
 
+    /**
+     * Synchronously flushes the pre-write buffer up to the given offset. This method delegates to
+     * {@link #requestFlush(long, FatalErrorHandler)} and blocks until the async flush scheduler
+     * completes the flush. Intended for test use only; production code should use {@link
+     * #requestFlush} which is non-blocking.
+     */
+    @VisibleForTesting
     public void flush(long exclusiveUpToLogOffset, FatalErrorHandler fatalErrorHandler) {
```
luoyuxia: seems this method is only used for testing. Is it possible to move it to test class?

fluss-server/src/main/java/org/apache/fluss/server/kv/KvTablet.java
```diff
@@ -667,42 +790,268 @@ private WalBuilder createWalBuilder(int schemaId, RowType rowType) throws Except
...
+                        return null;
+                    }
+                    if (preparedFlush.isEmpty()) {
+                        flushedLogOffset = targetOffset;
+                    }
+                    return preparedFlush;
+                });
+    }
+
+    private void writePreparedFlush(PreparedFlush preparedFlush) throws Exception {
```
luoyuxia, line 980: writePreparedFlush() writes the prepared entries to RocksDB outside kvLock, while flushedLogOffset is updated later under the lock in completeScheduledFlush(). This allows the following race:

1. A flush for offset N writes its entries to RocksDB.
2. Before completeScheduledFlush() updates flushedLogOffset, the periodic snapshot task acquires kvLock.
 3. The RocksDB checkpoint contains the entries up to N, but TabletState records the old flushedLogOffset.
 4. Recovery restores that checkpoint and replays WAL from the old offset, applying data already contained in the checkpoint again.

This can produce incorrect results for non-idempotent merge operations and make row count or auto-increment state inconsistent.

luoyuxia: With the asynchronous flush path, writePreparedFlush() holds a RocksDB ResourceGuard lease without holding kvLock, and then completeScheduledFlush() needs to acquire kvLock. However, drop() holds kvLock while calling close(), which waits for all RocksDB leases to be released.

This creates the following lock-order inversion:

 1. The flush thread acquires the RocksDB lease and later waits for kvLock in completeScheduledFlush().
 2. The drop thread acquires kvLock and calls rocksDBKv.close().
 3. rocksDBKv.close() waits for the flush thread to release its lease.
 4. Neither thread can proceed.

luoyuxia (commented):
@platinumhamburg Thanks for the pr. I left minor comments. PTAL

fluss-client/src/main/java/org/apache/fluss/client/write/RecordAccumulator.java
```diff
@@ -883,6 +918,84 @@ private boolean shouldStopDrainBatchesForBucket(WriteBatch first, TableBucket ta
...
+        throttleExpiryMs.remove(tableBucket);
+    }
+
+    /**
+     * Apply a hard back-off after the server rejected the write with a {@code
+     * StorageBackpressureException}. This signals that the storage engine has reached its own
+     * slowdown threshold; we stall the bucket for the full {@link #maxThrottleMs} window so the
+     * standard retry path can take effect.
+     */
+    void applyStorageBackpressureBackoff(TableBucket tableBucket) {
```
wuchong: Could we avoid adding this exception-specific method and map a hard rejection to full pressure in `Sender` instead, for example `accumulator.updateThrottle(readyWriteBatch.tableBucket(), 1.0f)`? Both paths install the same full `maxThrottleMs` window, so this method duplicates the state transition and couples `RecordAccumulator` to the exception semantics. `updateThrottle` can document `1.0f` as the internal hard-rejection value (while wire pressure remains below 1) and handle that value with `maxThrottleMs` directly to avoid long-to-float rounding; the dedicated test can then be folded into full-pressure coverage.

fluss-client/src/test/java/org/apache/fluss/client/write/RecordAccumulatorTest.java
```diff
@@ -701,4 +703,96 @@ private int getBatchNumInAccum(RecordAccumulator accum) {
...
+        accum.updateThrottle(tb1, 0.5f);
+
+        // Drain should only produce batches for tb2 (tb1 is throttled)
+        Map<Integer, List<ReadyWriteBatch>> batches =
+                accum.drain(
+                        cluster,
+                        new HashSet<>(Collections.singletonList(node1.id())),
+                        Integer.MAX_VALUE);
+        // tb2 should be in the batch, tb1 should not
+        List<ReadyWriteBatch> node1Batches = batches.get(node1.id());
```
wuchong, line 792: nit: assert tb2 is in the batches

fluss-client/src/test/java/org/apache/fluss/client/write/RecordAccumulatorTest.java
```diff
@@ -701,4 +703,96 @@ private int getBatchNumInAccum(RecordAccumulator accum) {
...
+    @Test
+    void testThrottledBucketSkippedInDrain() throws Exception {
+        IndexedRow row = indexedRow(DATA1_ROW_TYPE, new Object[] {1, "a"});
+        long batchSize = getTestBatchSize(row);
+        RecordAccumulator accum = createTestRecordAccumulator((int) batchSize, Integer.MAX_VALUE);
+        cluster = updateCluster(Arrays.asList(bucket1, bucket2, bucket3));
+
+        // Append records to tb1 and tb2
+        accum.append(createRecord(row), writeCallback, cluster, 0, false);
+        accum.append(createRecord(row), writeCallback, cluster, 1, false);
```
wuchong, line 780: nit: add comment that tb1 and tb2 are both on the same node1, so drain from node1 only return tb2 when tb1 is throttled.

fluss-rpc/src/main/proto/FlussApi.proto
```diff
@@ -940,6 +940,9 @@ message PbPutKvRespForBucket {
   // the log end offset (LEO) of the changelog after this write
   // this is the offset of the next record to be written, used for exactly-once semantics
   optional int64 log_end_offset = 5;
+  // backpressure signal from RocksDB write stall detection:
+  // 0 = normal (no pressure), (0,1) = DELAYED zone (normalized L0 ratio)
+  optional float pressure = 6;
```
wuchong, line 946: nit: add since version

fluss-rpc/src/main/java/org/apache/fluss/rpc/protocol/Errors.java
```diff
@@ -275,7 +276,11 @@ public enum Errors {
     INSUFFICIENT_KV_LEADER_REPLICA_CAPACITY(
             71,
             "The cluster does not have enough KV leader replica capacity.",
-            InsufficientKvLeaderReplicaCapacityException::new);
+            InsufficientKvLeaderReplicaCapacityException::new),
+    STORAGE_BACKPRESSURE_EXCEPTION(
```
wuchong, line 280: This new error code is not backward compatible with clients that only know the existing `Errors` table. A pre-upgrade client maps code 72 to the non-retriable `UNKNOWN_SERVER_ERROR`, so every storage-pressure rejection fails the write instead of backing off and retrying. For long-running clients (for example, Flink jobs), normal server pressure would therefore trigger frequent client-side task/job failovers, which is unacceptable during a rolling upgrade. Please add a server-side compatibility fallback based on the `PUT_KV` request API version: return `STORAGE_BACKPRESSURE_EXCEPTION` only to versions that understand it, and map the same condition for older clients to an existing retriable error such as `KV_STORAGE_EXCEPTION` (`KvStorageException`). Please also add a mixed-version test covering both response codes and the old client's retry behavior.

fluss-server/src/main/java/org/apache/fluss/server/kv/prewrite/KvPreWriteBuffer.java
wuchong: +1, I think this will have better performance as well. I think we don't need to iterate all the linked list, as if the last entry is FLUSHED, then all previous should have been FLUSHED.

fluss-server/src/main/java/org/apache/fluss/server/kv/prewrite/KvPreWriteBuffer.java
```diff
@@ -101,6 +102,9 @@ public class KvPreWriteBuffer implements AutoCloseable {
     // the max LSN in the buffer
     private long maxLogSequenceNumber = -1;
 
+    // Accumulated byte size of entries not yet completed by a flush.
+    private long pendingFlushBytes = 0;
+
     public KvPreWriteBuffer(
             KvBatchWriter kvBatchWriter, TabletServerMetricGroup serverMetricGroup) {
         this.kvBatchWriter = kvBatchWriter;
```
wuchong: This `kvBatchWriter` is not used in `KvPreWriteBuffer`, can be removed.

fluss-server/src/main/java/org/apache/fluss/server/kv/rocksdb/RocksDBKv.java
```diff
@@ -230,4 +318,132 @@ public Cache getBlockCache() {
...
+            lease = rocksDBResourceGuard.acquireResource();
+        } catch (IOException acquireFailed) {
+            // RocksDB is already closed (or being closed); treat as no pressure.
+            return 0L;
+        }
+        try {
+            return readLongProperty(NUM_FILES_AT_LEVEL0);
+        } finally {
+            lease.close();
+        }
```
wuchong: This would leak resources if `rocksDBResourceGuard.acquireResource()` throws non `IOException`. We should merge this 2 try catch blocks, and close the `lease` if it's not null in `finnaly` block.

fluss-server/src/main/java/org/apache/fluss/server/kv/KvTablet.java
```diff
@@ -667,42 +790,268 @@ private WalBuilder createWalBuilder(int schemaId, RowType rowType) throws Except
...
+        boolean flushCompleted = false;
+        try {
+            while (true) {
+                preparedFlush = prepareScheduledFlush();
+                flushCompleted = false;
+                if (preparedFlush == null) {
+                    break;
+                }
+                if (!preparedFlush.isEmpty()) {
+                    writePreparedFlush(preparedFlush);
```
wuchong: **[P0]:** `KvTablet#writePreparedFlush` runs after `KvTablet#prepareScheduledFlush` has released `kvLock`, and `KvTablet#completeScheduledFlush` reacquires it only after all RocksDB writes return. This breaks the snapshot/metadata atomicity promised by `kvLock`: `openScan` or periodic snapshot initialization can acquire the lock while this method is writing, capture RocksDB after one or more `db.write()` calls, but capture the old `flushedLogOffset` and `rowCount`. The persisted snapshot can therefore contain KV state beyond the offset recorded in `TabletState`.

There is also a partial-failure window because `RocksDBWriteBatchWrapper.put/delete` invokes `flushIfNeeded()` every 500 records or when `batchSize` is reached. A single `PreparedFlush` may issue several native writes; if the first two succeed and the third is rejected by `noSlowdown`, `abortScheduledFlush()` marks the entire prepared range ACTIVE even though part of it is already in RocksDB, while `flushedLogOffset` and `rowCount` still describe none of those writes. A snapshot taken before the retry can persist this mismatch.

Please hold the `kvLock` write lock across `prepareScheduledFlush -> writePreparedFlush -> completeScheduledFlush` (or the entire `runScheduledFlush`) so snapshots cannot interleave. In addition, make each successful native write a separately completable prepared sub-batch: after every successful `kvBatchWriter.flush()`, complete exactly those entries in `KvPreWriteBuffer` and advance their `flushedLogOffset`/`rowCount`, so a later scheduled flush skips entries that were already committed.

fluss-server/src/main/java/org/apache/fluss/server/kv/rocksdb/RocksDBWriteBatchWrapper.java
```diff
@@ -147,7 +184,7 @@ private void flushIfNeeded() throws IOException {
     @Override
     public void close() throws IOException {
         try {
-            if (batch.count() != 0) {
+            if (!flushFailed && batch.count() != 0) {
```
wuchong: Add a comment to explain why skip `flush()` when `flushFailed == true` is safe. IIUC, it's safe because the `flush()` has throwed exceptions and the outer of `flush()` will catch the exception to handle how to reprocess the whole write batch.

fluss-server/src/main/java/org/apache/fluss/server/kv/snapshot/SnapshotLocation.java
```diff
@@ -265,16 +263,13 @@ public void close() {
...
-            // check if there was nothing ever written
-            if (outStream == null && pos == 0) {
-                return null;
-            }
-
             synchronized (this) {
                 if (!closed) {
                     try {
+                        if (outStream == null && pos == 0) {
+                            createStream();
```
wuchong: Could you clarify which real RocksDB checkpoint file can be zero-length here? In RocksDB 7.5.3, [`BuildTable`](https://github.com/facebook/rocksdb/blob/v7.5.3/db/builder.cc#L255-L283) abandons an empty builder and removes a zero-sized table file (`Empty SST file not kept`), so a healthy checkpoint should not contain a zero-byte SST. The added test `testUploadEmptyFileCorrectly` also constructs an artificial `empty.log`, not an SST. The old behavior does return `null` for any zero-byte source and later fails the non-null requirement in `KvFileHandleAndLocalPath.of`, so supporting generic zero-length snapshot files can be defensively valid; however, this PR does not demonstrate the production file or scenario that needs it, and the change is unrelated to KV backpressure. Please either document and reproduce the actual zero-length checkpoint file (and describe this generically rather than as an empty SST), or move/remove this unrelated change from this PR.

fluss-server/src/main/java/org/apache/fluss/server/kv/KvFlushScheduler.java
```diff
@@ -0,0 +1,171 @@
...
+        this.flushExecutor = newExecutor(flushThreads, FLUSH_THREAD_POOL_NAME);
+        this.retryTimer = new DefaultTimer("kv-flush-retry");
+        this.retryReaper = new FlushRetryReaper();
+        this.retryReaper.start();
+        LOG.info("Created KV flush scheduler with {} flush threads.", flushThreads);
+    }
+
+    public void enqueue(KvTablet tablet) {
+        FlushTask task = new FlushTask(tablet);
+        synchronized (lifecycleLock) {
```
wuchong: Since `closed` is volatile and both `flushExecutor` and `retryTimer` are thread-safe, we can remove the `synchronized (lifecycleLock)` block to achieve a lock-free design. Introducing a lock solely for coordinating the `closed` state seems excessive.

fluss-server/src/main/java/org/apache/fluss/server/kv/KvFlushScheduler.java
```diff
@@ -0,0 +1,171 @@
...
+        if (flushExecutor != null) {
+            if (isFlushExecutorThread()) {
+                flushExecutor.shutdown();
+            } else {
+                ExecutorUtils.gracefulShutdown(5, TimeUnit.SECONDS, flushExecutor);
+            }
+        }
+        if (retryReaper != null && retryTimer != null) {
+            retryReaper.initiateShutdown();
+            retryTimer.add(
```
wuchong, line 126: nit: Add comment to explain why adding an empty timer task here.

fluss-server/src/main/java/org/apache/fluss/server/kv/KvTablet.java
wuchong: +1 

The current wait condition is also weaker than the method contract: returning when the state becomes `STORAGE_BLOCKED` does not mean that `exclusiveUpToLogOffset` has actually been flushed.

To preserve the existing `Long.MAX_VALUE` behavior without passing a separate `LongSupplier`, `KvTablet` can expose a small package-private testing accessor:

```java
@VisibleForTesting
long localLogEndOffset() {
    return logTablet.localLogEndOffset();
}
```

The blocking logic can then live entirely in a test utility (e.g., `KvTabletTestUtils`) in the same package:

```java
public static void flushAndWait(KvTablet tablet, long exclusiveUpToLogOffset) {
    long targetOffset =
            exclusiveUpToLogOffset == Long.MAX_VALUE
                    ? tablet.localLogEndOffset()
                    : exclusiveUpToLogOffset;

    AtomicReference<Throwable> failure = new AtomicReference<>();
    tablet.requestFlush(targetOffset, failure::set);

    waitUntil(
            () ->
                    failure.get() != null
                            || tablet.getFlushedLogOffset() >= targetOffset,
            Duration.ofSeconds(30),
            String.format(
                    "KV flush did not reach offset %s for %s",
                    targetOffset, tablet.getTableBucket()));

    Throwable cause = failure.get();
    if (cause != null) {
        throw new AssertionError(
                "KV flush failed for " + tablet.getTableBucket(), cause);
    }
}
```

fluss-server/src/main/java/org/apache/fluss/server/kv/KvTablet.java
```diff
@@ -667,42 +790,268 @@ private WalBuilder createWalBuilder(int schemaId, RowType rowType) throws Except
...
+    }
+
+    @GuardedBy("kvLock")
+    private long nextFlushRetryDelayMs() {
+        long delay = MIN_FLUSH_RETRY_DELAY_MS;
+        if (flushRetryAttempts > 0) {
+            int shift = Math.min(flushRetryAttempts, 30);
+            delay = MIN_FLUSH_RETRY_DELAY_MS << shift;
+        }
+        if (flushRetryAttempts < 30) {
```
wuchong: Is a maximum of 5 retries sufficient? Note that `<< 30` results in a very large number, which may cause an overflow.

fluss-server/src/main/java/org/apache/fluss/server/kv/KvTablet.java
```diff
@@ -667,42 +790,268 @@ private WalBuilder createWalBuilder(int schemaId, RowType rowType) throws Except
...
+        }
+        throw new RuntimeException("Flush did not complete within 30 seconds for " + tableBucket);
+    }
+
+    public void requestFlush(long exclusiveUpToLogOffset, FatalErrorHandler fatalErrorHandler) {
+        asyncFatalErrorHandler = fatalErrorHandler;
+        inWriteLock(kvLock, () -> requestFlushInternal(exclusiveUpToLogOffset));
+    }
+
+    public void setFlushCompleteListener(@Nullable Runnable flushCompleteListener) {
```
wuchong, line 841: The `flushCompleteListener` can be set in the constructor. It's not invoked during recovery, so it's safe to initilaized earlier.

fluss-server/src/main/java/org/apache/fluss/server/kv/KvTablet.java
```diff
@@ -894,4 +1256,16 @@ KvPreWriteBuffer getKvPreWriteBuffer() {
     public RocksDBKv getRocksDBKv() {
         return rocksDBKv;
     }
+
+    /** Returns the recent normalized backpressure pressure in {@code [0, 1)}. */
+    public float currentPressure() {
+        return rocksDBKv.currentPressure();
+    }
+
+    private enum FlushState {
```
wuchong: Could we document the allowed `FlushState` transitions as a graph directly in the Java comment near this enum? The state assignments are currently scattered across several methods, making it difficult to understand which source state each operation expects and why the transition is valid. For example:

```java
/**
 * Flush scheduling state for one KV tablet.
 *
 * <p>This state tracks scheduler ownership and retry backoff. Whether more data needs to be
 * flushed is determined separately by {@code requestedFlushOffset} and {@code
 * flushedLogOffset}.
 *
 * <pre>
 * Normal scheduling:
 *
 *   +------+    a new target requires flushing    +--------+
 *   | IDLE | -----------------------------------> | QUEUED |
 *   +------+                                      +--------+
 *                                                     |
 *                                              a worker claims
 *                                               the queued task
 *                                                     |
 *                                                     v
 *                                               +---------+
 *                                               | RUNNING |
 *                                               +---------+
 *                                                     |
 *                                            the target is reached
 *                                            or a fatal error occurs
 *                                                     |
 *                                                     v
 *                                                 +------+
 *                                                 | IDLE |
 *                                                 +------+
 *
 * Backpressure retry:
 *
 *                    storage rejection or
 *                    prepared-entry conflict
 *   +---------+ -------------------------------> +-----------------+
 *   | RUNNING |                                  | STORAGE_BLOCKED |
 *   +---------+                                  +-----------------+
 *                                                        |
 *                                                 the retry timer
 *                                                      fires
 *                                                        |
 *                                                        v
 *                                                   +--------+
 *                                                   | QUEUED |
 *                                                   +--------+
 * </pre>
 *
 * <p>A higher flush target received in {@code QUEUED}, {@code RUNNING}, or {@code
 * STORAGE_BLOCKED} only advances {@code requestedFlushOffset}; it does not change the state or
 * enqueue another task. Closing the tablet sets {@code isClosed} and terminates the state
 * machine. All transitions must happen while holding {@code kvLock}.
 */
```

Please also consider centralizing state changes in a helper such as `transitionFlushState(expected, target, reason)` so that the allowed transitions shown in the graph are enforced by the implementation rather than only documented.

fluss-server/src/main/java/org/apache/fluss/server/replica/Replica.java
```diff
@@ -734,7 +734,29 @@ private void dropKv() {
...
+                                return maybeIncrementLeaderHW(logTablet, clock.milliseconds());
+                            } catch (IOException e) {
+                                fatalErrorHandler.onFatalError(e);
+                                return false;
+                            }
+                        });
+        if (leaderHWIncremented) {
+            tryCompleteDelayedOperations();
+        } else {
+            delayedWriteManager.checkAndComplete(new DelayedTableBucketKey(tableBucket));
```
wuchong: ack=1 shouldn't have delayed operation, so this is also not needed.

fluss-server/src/main/java/org/apache/fluss/server/replica/Replica.java
```diff
@@ -734,7 +734,29 @@ private void dropKv() {
...
+    private void onKvFlushComplete() {
+        boolean leaderHWIncremented =
+                inWriteLock(
+                        leaderIsrUpdateLock,
+                        () -> {
+                            if (!isLeader()) {
+                                return false;
+                            }
+                            try {
+                                return maybeIncrementLeaderHW(logTablet, clock.milliseconds());
```
wuchong, line 750: Because this callback recomputes against the latest replica LEO/ISR after the asynchronous gap, a flush task that successfully materialized target N can observe a newer candidate M > N. `maybeIncrementLeaderHW` then requests another flush and returns false, so the HW is not advanced even to N. Under continuous writes or follower progress, `runScheduledFlush` also keeps draining an ever-increasing `requestedFlushOffset` and invokes this callback only when the run exits, which can leave the HW stale for an unbounded time and delay `acks=-1` writes and fetch visibility. Please capture and bound the target of each scheduled flush run, stop after reaching that target, and on completion advance the HW up to `min(replicationCandidate, completedFlushOffset)` before rescheduling any remaining work. Please add a deterministic test where the target advances during a flush and verify that the HW first reaches the completed target without waiting for the newer one.

fluss-server/src/main/java/org/apache/fluss/server/replica/ReplicaManager.java
```diff
@@ -1358,6 +1360,9 @@ private Map<TableBucket, PutKvResultForBucket> putToLocalKv(
                         tb,
                         appendInfo.firstOffset(),
                         appendInfo.lastOffset());
+                // The pressure field is left at its default (0f) here and filled right before the
+                // response goes out, by either maybeAddDelayedWrite (acks != -1) or
+                // DelayedWrite#onComplete (acks == -1).
                 putResultForBucketMap.put(
                         tb, new PutKvResultForBucket(tb, appendInfo.lastOffset() + 1));
```
wuchong: I think we should fill the pressure value here, rather than in `DelayedWrite`. Because:

1. Here reflects the true pressure after `putKv`. `DelayedWrite` may have more delays and is not accurate. 
2. This can simplify the logic in `DelayedWrite` (the iterate + instance of + set logic)
3. The pressure can be immutable which is initilized during construction.

fluss-server/src/main/java/org/apache/fluss/server/replica/Replica.java
```diff
@@ -1512,6 +1562,30 @@ public Tuple2<Boolean, Errors> checkEnoughReplicasReachOffset(long requiredOffse
         }
     }
 
+    /**
+     * Checks whether the local leader KV view has been flushed to the required offset. Used by
+     * PutKv with {@code acks = 1}: the client should wait for local materialization, but should not
+     * be upgraded to waiting for high watermark replication.
+     */
+    public Tuple2<Boolean, Errors> checkLocalKvFlushedOffset(long requiredOffset) {
```
wuchong: This is not needed for acks = 1.

fluss-server/src/main/java/org/apache/fluss/server/replica/ReplicaManager.java
```diff
@@ -1368,7 +1373,16 @@ private Map<TableBucket, PutKvResultForBucket> putToLocalKv(
                 tableMetrics.incLogBytesIn(appendInfo.validBytes());
                 tableMetrics.incLogMessageIn(appendInfo.numMessages());
             } catch (Exception e) {
-                if (isUnexpectedException(e)) {
+                if (e instanceof StorageBackpressureException) {
```
wuchong, line 1389: **P0:** throw backward-compatible retriable exception for old `apiVersion` client.

fluss-server/src/main/java/org/apache/fluss/server/replica/ReplicaManager.java
```diff
@@ -1699,6 +1717,21 @@ private <T extends WriteResultForBucket> void maybeAddDelayedWrite(
...
+                            try {
+                                ((PutKvResultForBucket) r)
+                                        .setPressure(
+                                                getReplicaOrException(tb)
+                                                        .samplePressureForCompletion());
+                            } catch (Exception ignore) {
+                                // leader moved or replica gone, leave default 0f
+                            }
+                        }
+                    });
```
wuchong: We don't need this complex logic if fill the pressure value at the begginning.

fluss-server/src/main/java/org/apache/fluss/server/replica/ReplicaManager.java
```diff
@@ -1674,7 +1689,9 @@ private <T extends WriteResultForBucket> void maybeAddDelayedWrite(
             int requestBucketSize,
             Map<TableBucket, T> writeResults,
             Consumer<List<T>> responseCallback) {
-        if (delayedWriteRequired(requiredAcks, requestBucketSize, writeResults)) {
+        DelayedWrite.CompletionMode completionMode =
+                delayedWriteCompletionMode(requiredAcks, requestBucketSize, writeResults);
```
wuchong: **P0:** We shouldn't add `DelayedWrite` for ack=1, because ack=1 should respond to the client when the local CDC log is generated (no need to wait high-watermark). So we can remove the `checkLocalKvFlushedOffset()`, the `CompletionMode`, and revert this change.

fluss-server/src/test/java/org/apache/fluss/server/kv/prewrite/KvPreWriteBufferTest.java
```diff
@@ -253,6 +255,110 @@ private static void bufferDelete(
...
+        buffer.truncateTo(3, TruncateReason.ERROR);
+        assertThat(buffer.pendingFlushBytes()).isEqualTo(11);
+
+        // Flush remaining (key2): decreases by 11
+        flushBuffer(buffer, Long.MAX_VALUE);
+        assertThat(buffer.pendingFlushBytes()).isEqualTo(0);
+    }
+
+    @Test
+    void testBufferDoesNotExposeSplitFlushByteCounters() {
```
wuchong: I don't see the necessary of this test method.

fluss-server/src/test/java/org/apache/fluss/server/kv/rocksdb/RocksDBKvTest.java
```diff
@@ -64,4 +70,256 @@ void testRocksDbKv(@TempDir Path tempDir) throws Exception {
...
+                kv.db.flush(flushOpts);
+            }
+        }
+        kv.refreshL0FileCount();
+    }
+
+    private static boolean noSlowdown(RocksDBWriteBatchWrapper writer) throws Exception {
+        Field field = RocksDBWriteBatchWrapper.class.getDeclaredField("options");
+        field.setAccessible(true);
+        return ((WriteOptions) field.get(writer)).noSlowdown();
```
wuchong: Avoid to use reflection if possible. You can set `options` field to package visible, and then directly access it in this test.

fluss-server/src/test/java/org/apache/fluss/server/kv/snapshot/KvSnapshotDataUploaderTest.java
```diff
@@ -104,6 +105,40 @@ void testMultiThreadUploadCorrectly() throws Exception {
         }
     }
 
+    @Test
+    void testUploadEmptyFileCorrectly() throws Exception {
```
wuchong: Is this a valid test? IIUC, there shouldn't be empty sst files for rocksdb snapshot.

fluss-server/src/main/java/org/apache/fluss/server/kv/KvTablet.java
```diff
@@ -212,6 +237,83 @@ public static KvTablet create(
...
+                arrowCompressionInfo,
+                schemaGetter,
+                changelogImage,
+                sharedRateLimiter,
+                new KvFlushScheduler(serverConf),
+                true,
+                autoIncrementManager);
+    }
+
+    public static KvTablet create(
```
wuchong, line 292: We should avoid introducing too many different constructors, otherwise, it's easy to use the wrong one. At lease, please add comment and `@VisibleForTesting` for this method.

fluss-server/src/test/java/org/apache/fluss/server/kv/KvTabletTest.java
```diff
@@ -1496,6 +1541,386 @@ private Value valueOf(BinaryRow row) {
...
+        Field field = KvTablet.class.getDeclaredField("flushState");
+        field.setAccessible(true);
+        for (Object constant : field.getType().getEnumConstants()) {
+            if (constant.toString().equals(stateName)) {
+                field.set(tablet, constant);
+                return;
+            }
+        }
+        throw new IllegalArgumentException("Unknown flush state: " + stateName);
+    }
```
wuchong: Avoid using reflections. They are all in the same package; you can change the members to package visible and add `@VisibleForTesting` annotation.

fluss-server/src/test/java/org/apache/fluss/server/replica/ReplicaManagerTest.java
```diff
@@ -667,6 +667,34 @@ tb, genKvRecordBatch(DATA_1_WITH_KEY_AND_VALUE)),
...
+                Collections.singletonMap(tb, genKvRecordBatch(DATA_1_WITH_KEY_AND_VALUE)),
+                null,
+                MergeMode.DEFAULT,
+                PUT_KV_VERSION,
+                future::complete);
+
+        assertThat(future.get(10, TimeUnit.SECONDS)).containsOnly(new PutKvResultForBucket(tb, 8));
+        Replica replica = replicaManager.getReplicaOrException(tb);
+        assertThat(replica.getLocalLogEndOffset()).isEqualTo(8L);
+        assertThat(replica.getKvTablet().getFlushedLogOffset()).isGreaterThanOrEqualTo(8L);
```
wuchong: Actually, we can't guarantee the kv is flushed when a `ack=1` request is responded.

fluss-server/src/test/java/org/apache/fluss/server/kv/KvTabletTest.java
```diff
@@ -1496,6 +1541,386 @@ private Value valueOf(BinaryRow row) {
...
+                            tmpKvDir,
+                            schemaGetter,
+                            new HashMap<>(),
+                            scheduler);
+
+            setFlushState(kvTablet, "STORAGE_BLOCKED");
+            scheduler.retryLater(kvTablet, 10L);
+
+            kvTablet.close();
+            Thread.sleep(300L);
```
wuchong: Could we make this test deterministic instead of waiting for a fixed 300 ms? Since requestFlushRetry() is package-private and this test is in the same package, we can use ManualKvFlushScheduler, close the tablet, invoke kvTablet.requestFlushRetry() directly to model the timer callback, and then assert that no task was enqueued. This avoids adding 300 ms to every run and removes scheduler-timing flakiness. Retrying the current enqueueCount == 0 assertion would not be sufficient because that negative condition is already true before the delayed callback executes.

wuchong (commented):
Thanks @platinumhamburg , I left some comments below. I annotated the critical problems with **P0** tag. Please take a look.

platinumhamburg:
Thanks @wuchong and @luoyuxia for the detailed feedback. All review comments have been addressed in the latest commits. Please re-review at your convenience.

wuchong (approved):
I pushed a commit to fix a minor problem, please take a look: 

The previous implementation checked the byte limit only against the entries already accumulated in a segment. As a result, adding the current entry could make the segment exceed `maxBytesPerSegment`. The fix evaluates the projected segment size before adding each entry and splits the segment when necessary. A single entry larger than the limit is kept as an oversized singleton to guarantee progress. Unit tests cover projected-size splitting, exact-limit boundaries, oversized entries, combined byte/record limits, and retry after partial flush failure.

platinumhamburg:
> I pushed a commit to fix a minor problem, please take a look:
> 
> The previous implementation checked the byte limit only against the entries already accumulated in a segment. As a result, adding the current entry could make the segment exceed `maxBytesPerSegment`. The fix evaluates the projected segment size before adding each entry and splits the segment when necessary. A single entry larger than the limit is kept as an oversized singleton to guarantee progress. Unit tests cover projected-size splitting, exact-limit boundaries, oversized entries, combined byte/record limits, and retry after partial flush failure.

Thanks, LGTM.

PR #3468, opened 2026-06-10
[client] Propagate write batch failures to flush

raoluSmile:
### Purpose

  When `WriterClient.flush()` waits for pending write batches, it relies on
  `RecordAccumulator.awaitFlushCompletion()`, which waits on each
  `WriteBatch.RequestFuture`.

  Before this change, a write batch could be completed exceptionally through
  `WriteBatch.completeExceptionally(exception)`, but `RequestFuture` only counted
  down its latch and did not keep or rethrow the exception. As a result,
  `flush()` could wait silently without surfacing the underlying write failure,
  and in retry/error paths this could look like a hang to the caller.

  I hit this when TabletServer rejected writes with `DISK_WRITE_LOCKED`.
  The writer kept retrying until request timeout, but the failure was not
  propagated through the flush waiting path, so the test appeared to hang instead
  of failing with the real write error.

  This issue is independent of `DISK_WRITE_LOCKED` itself. Any write batch that
  is completed exceptionally should be observable by callers waiting in `flush()`.

  This is related to, but different from, #1258. #1258 fixed a lost `WriteBatch`
  case where `awaitFlushCompletion()` could wait forever. This PR fixes another
  failure path: a `WriteBatch` is completed exceptionally, but the corresponding
  `RequestFuture.await()` did not propagate the batch exception.

  ### Brief change log

  - Store the batch exception in `WriteBatch.RequestFuture` when a batch completes.
  - Rethrow the stored exception from `RequestFuture.await()`.
  - Let `RecordAccumulator.awaitFlushCompletion()` propagate batch failures.
  - Wrap propagated flush failures in `WriterClient.flush()`.
  - Stop retrying retriable write errors after the request timeout has elapsed.
  - Add a unit test to verify that `awaitFlushCompletion()` observes batch failures.

  ### Tests

  - `./mvnw -pl fluss-client -DfailIfNoTests=false -Dtest=RecordAccumulatorTest#testAwaitFlushCompletionPropagatesBatchFailure test`
  - `./mvnw -pl fluss-client -am -DskipTests compile`

  ### API and Format

  ### Documentation

### DIFF
--- a/fluss-client/src/main/java/org/apache/fluss/client/write/RecordAccumulator.java
+++ b/fluss-client/src/main/java/org/apache/fluss/client/write/RecordAccumulator.java
@@ -375,7 +375,7 @@ public final class RecordAccumulator {
     }
 
     /** Mark all buckets as ready to send and block until to send is complete. */
-    public void awaitFlushCompletion() throws InterruptedException {
+    public void awaitFlushCompletion() throws Exception {
         try {
             // Obtain a copy of all the incomplete write request result(s) at the time of the
             // flush. We must be careful not to hold a reference to the ProduceBatch(s) so that
--- a/fluss-client/src/main/java/org/apache/fluss/client/write/Sender.java
+++ b/fluss-client/src/main/java/org/apache/fluss/client/write/Sender.java
@@ -321,6 +321,7 @@ public class Sender implements Runnable {
         WriteBatch batch = readyWriteBatch.writeBatch();
         return batch.attempts() < retries
                 && !batch.isDone()
+                && batch.waitedTimeMs(System.currentTimeMillis()) < maxRequestTimeoutMs
                 && ((error.exception() instanceof RetriableException)
                         || (idempotenceManager.idempotenceEnabled()
                                 && idempotenceManager.canRetry(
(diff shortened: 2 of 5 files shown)


fluss-client/src/main/java/org/apache/fluss/client/write/RecordAccumulator.java
Jackeyzhe, line 385: If one request future fails, this loop exits immediately and leaves the remaining futures captured by the same flush potentially incomplete. That breaks the documented flush post-condition and can let writer close proceed while another drained/in-flight batch is still pending. Please retain the first failure, wait for every captured future, and propagate the failure afterward; a two-batch regression test would cover this case.

fluss-client/src/main/java/org/apache/fluss/client/write/Sender.java
```diff
@@ -321,6 +321,7 @@ private boolean canRetry(ReadyWriteBatch readyWriteBatch, Errors error) {
         WriteBatch batch = readyWriteBatch.writeBatch();
         return batch.attempts() < retries
                 && !batch.isDone()
+                && batch.waitedTimeMs(System.currentTimeMillis()) < maxRequestTimeoutMs
```
Jackeyzhe, line 324: waitedTimeMs is measured from batch creation and includes batching, metadata lookup, throttling, and queueing, while client.request-timeout is documented as the timeout for one request. This can disable retries for an old queued batch even when its current request fails immediately and client.writer.retries still allows retries. Please model this as a dedicated delivery timeout, or keep this separate retry-policy change out of this PR, and add coverage for the expiration path.

PR #3469, opened 2026-06-10
[lake/tiering] Add pendingRecordLag metric for lake tiering

### LINKED ISSUE #3464: [Feature] Add pendingRecordLag metric for lake tiering
### Search before asking

### Motivation

The current lake tiering bucket metric `timestampLag` is calculated from the timestamp difference between the latest local log record and the latest lake-tiered log record.

This can produce a pulse-like false positive after an idle period:

1. A table writes records and lake tiering catches up.
2. The table stays idle for a while.
3. A new batch is written after the idle period.
4. `timestampLag` immediately becomes large because it includes the idle time between the previous lake-tiered record timestamp and the new local record timestamp.

In this case, lake tiering is not necessarily delayed. The metric is large because new data arrived after a quiet period. This makes `timestampLag` hard to use directly for freshness alerts.

#### Reference: MySQL replication/binlog lag handling

MySQL replication has a similar operational problem: an idle source should not be treated as replication lag. MySQL handles this by making lag depend on replication progress/backlog rather than only comparing the latest source event timestamp with the latest applied event timestamp.

In `SHOW REPLICA STATUS`, MySQL exposes `Seconds_Behind_Source` as the replication delay indicator. Conceptually, it is backlog-aware:

```text
if the replica has caught up and has no unapplied relay log events:
    Seconds_Behind_Source = 0
else:
    Seconds_Behind_Source ~= current_replica_time - timestamp(event currently being applied)
```

The important behavior for this issue is the caught-up case: when there is no unapplied backlog, the lag is `0`. It does not keep increasing just because the source has been idle for a long time.

Reference:

- MySQL `SHOW REPLICA STATUS` and `Seconds_Behind_Source`: https://dev.mysql.com/doc/refman/8.4/en/show-replica-status.html

This suggests that Fluss lake tiering should expose a backlog-aware lag metric. The metric should answer: if there are committed records pending to be tiered to lake, how long has the oldest pending record been waiting? If there is no pending backlog, the lag should be `0`, regardless of how long the table was idle before the next write.

### Solution

Add a new bucket-level lake tiering gauge metric:

```text
pendingRecordsLag
```

Semantics:

```text
if there are no committed records pending lake tiering
   OR estimatedPendingStartTimeMs < 0:
    pendingRecordsLag = 0
else:
    pendingRecordsLag = max(0, currentTimeMillis - estimatedPendingStartTimeMs)
```

The `estimatedPendingStartTimeMs < 0` guard ensures the `-1` (no-pending) sentinel never flows into `now - (-1)`, which would otherwise emit an epoch-scale bogus value if `hasPendingRecords` and the state were ever transiently inconsistent.

A record is "pending" when:

```text
firstPendingOffset = lakeLogEndOffset >= 0
    ? max(lakeLogEndOffset, localLogStartOffset)
    : localLogStartOffset

hasPendingRecords = firstPendingOffset < highWatermark
```

#### Approach: maintain an estimated pending start time (no IO)

Rather than reading the record at `firstPendingOffset` to get its timestamp (which requires local, and possibly remote, IO on every metric scrape), each lake-enabled bucket maintains a single lightweight in-memory state:

```text
estimatedPendingStartTimeMs
```

State update rules:

```text
pendingRecords 0 -> >0:
    estimatedPendingStartTimeMs = now

pendingRecords >0 -> 0:
    estimatedPendingStartTimeMs = -1

lake tiering progress updated and still has pending records:
    timestampLag = max(0, logMaxTimestamp - lakeMaxTimestamp)
    estimatedPendingStartTimeMs =
        max(estimatedPendingStartTimeMs, now - timestampLag)
```

The first rule (`0 -> >0` sets `estimatedPendingStartTimeMs = now`) solves the idle-then-write pulse: a freshly-written record starts its lag from `0` instead of inheriting an old `lakeMaxTimestamp`.

The progress-update rule uses `now - timestampLag` not as an exact reconstruction of the first pending timestamp, but as a progress signal to push an outdated pending start forward, so the estimate does not stay stuck near already-tiered data.

#### When each update happens

The three rules hang off **two real events**: a high-watermark advance (new committed records become visible) and a lake tiering progress update (`lakeLogEndOffset` / `lakeMaxTimestamp` advances). `hasPendingRecords` everywhere uses the same `firstPendingOffset < highWatermark` as the metric.

```text
Rule 1 (0 -> >0, set now):
    event:  high watermark advances
    where:  on the high-watermark update path (LogTablet.updateHighWatermarkMetadata)
    when:   right after HWM advances, if hasPendingRecords flips false -> true
    note:   must be detected eagerly on this path, not lazily at scrape/progress time,
            otherwise `now` lands later than when the record actually became pending,
            causing systematic underestimation.

Rule 2 (>0 -> 0, set -1):
    event:  lake tiering progress advances
    where:  on the progress update path (ReplicaManager.notifyLakeTableOffset ->
            LogTablet.updateLakeLogEndOffset)
    when:   after lakeLogEndOffset advances, if firstPendingOffset >= highWatermark
            (backlog drained)

Rule 3 (still pending, correct forward):
    event:  lake tiering progress advances (same path as Rule 2)
    when:   after the progress update, if hasPendingRecords is still true, apply
            estimatedPendingStartTimeMs = max(estimatedPendingStartTimeMs, now - timestampLag)
```

Rules 2 and 3 are mutually exclusive on the same progress-update path: after applying the update, go to Rule 2 if there is no pending backlog, otherwise Rule 3.

This state is kept only in runtime memory; it is not persisted/recovered. If lost, the next `0 -> >0` transition or progress update rebuilds the estimate.

#### Quantitative error analysis

Define:

```text
now                          = current progress-update time
lakeMaxTimestamp             = max timestamp already tiered to lake
logMaxTimestamp              = max timestamp of committed local log records
firstPendingTimestamp        = true timestamp of the first pending record
truePendingLag               = now - firstPendingTimestamp        (ground truth)
estimationError              = pendingRecordsLag - truePendingLag
```

After a progress update, the corrected estimate is bounded by `timestampLag`:

```text
candidateStartTimeMs = now - timestampLag
estimatedPendingStartTimeMs >= candidateStartTimeMs
=> pendingRecordsLag = now - estimatedPendingStartTimeMs
                     <= now - candidateStartTimeMs
                     = timestampLag

  pendingRecordsLag <= timestampLag
```

`estimatedPendingStartTimeMs >= candidateStartTimeMs` is not an extra assumption; it follows directly from the correction rule `estimatedPendingStartTimeMs = max(estimatedPendingStartTimeMs, candidateStartTimeMs)`, since `max(x, candidateStartTimeMs) >= candidateStartTimeMs`.

This bound holds **at the moment of the progress update** — `now` and `candidateStartTimeMs` use the same update's `now`. Between two progress updates `estimatedPendingStartTimeMs` is fixed while wall-clock `now` keeps growing, so `pendingRecordsLag` rises above that update's `timestampLag`. That is the intended behavior where lag keeps growing while tiering is stalled, and does not contradict the bound (which is about the update instant, not arbitrary times).

**Overestimate bound.** Starting from `pendingRecordsLag <= timestampLag`, and substituting `timestampLag = logMaxTimestamp - lakeMaxTimestamp` and `truePendingLag = now - firstPendingTimestamp`:

```text
estimationError
    = pendingRecordsLag - truePendingLag
    <= timestampLag - truePendingLag
    = (logMaxTimestamp - lakeMaxTimestamp) - (now - firstPendingTimestamp)
    = (logMaxTimestamp - lakeMaxTimestamp) - now + firstPendingTimestamp
    = (firstPendingTimestamp - lakeMaxTimestamp) - (now - logMaxTimestamp)

since (now - logMaxTimestamp) >= 0:

    estimationError <= firstPendingTimestamp - lakeMaxTimestamp
```

The last regrouping just moves `firstPendingTimestamp` next to `lakeMaxTimestamp` and `logMaxTimestamp` next to `now`, turning the expression into a boundary-gap term `(firstPendingTimestamp - lakeMaxTimestamp)` minus an idle term `(now - logMaxTimestamp)`.

So a positive error is bounded by the timestamp gap at the `stoppingOffset` boundary — the gap between the last tiered record and the first pending record. Under continuous high-frequency writes this gap is typically milliseconds-to-seconds, far smaller than `freshness`.

Example (boundary write gap -> overestimate):

```text
lakeMaxTimestamp = 10:30, firstPendingTimestamp = 10:40, logMaxTimestamp = 10:45, now = 10:50
timestampLag         = 15min
candidateStartTimeMs = 10:35
pendingRecordsLag    = 15min
truePendingLag       = 10min
estimationError      = +5min   (the 10:30-10:40 write gap)
```

A positive `estimationError` therefore comes mainly from the gap between `lakeMaxTimestamp` and the true `firstPendingTimestamp`.

#### Underestimate analysis

The underestimate direction is defined as:

```text
underEstimate = truePendingLag - pendingRecordsLag
```

This direction needs more attention than overestimation. The root cause is that the correction subtracts a timestamp-domain `timestampLag` from the wall-clock `now`:

```text
candidateStartTimeMs = now - timestampLag
                     = now - (logMaxTimestamp - lakeMaxTimestamp)
                     = lakeMaxTimestamp + (now - logMaxTimestamp)
```

When the correction term is selected, `pendingRecordsLag = timestampLag = logMaxTimestamp - lakeMaxTimestamp`, so:

```text
underEstimate
    = truePendingLag - pendingRecordsLag
    = (now - firstPendingTimestamp) - (logMaxTimestamp - lakeMaxTimestamp)
    = now - firstPendingTimestamp - logMaxTimestamp + lakeMaxTimestamp
    = (now - logMaxTimestamp) - (firstPendingTimestamp - lakeMaxTimestamp)
```

If the pending boundary is continuous, i.e. `firstPendingTimestamp ~= lakeMaxTimestamp`, then:

```text
underEstimate ~= now - logMaxTimestamp
```

So underestimation is essentially "how long since the last write". With continuous writes `now ~= logMaxTimestamp` and underestimation is near 0; after writes stop, `now - logMaxTimestamp` keeps growing and the metric can systematically underestimate the true pending time. Low-frequency writing only shows this problem when the gap between two writes is large enough that `logMaxTimestamp` clearly lags `now`.

Example:

```text
10:00-10:30 continuous write A; estimatedPendingStartTimeMs = 10:00
10:30       writes stop, logMaxTimestamp = 10:30
11:00       lakeMaxTimestamp only reaches 10:15; 10:15-10:30 still pending

timestampLag                = 10:30 - 10:15 = 15min
candidateStartTimeMs        = 11:00 - 15min = 10:45
estimatedPendingStartTimeMs = max(10:00, 10:45) = 10:45
pendingRecordsLag           = 11:00 - 10:45 = 15min
```

But the true first pending is around 10:15:

```text
truePendingLag = 11:00 - 10:15 = 45min
underEstimate  = 45min - 15min = 30min   (= now - logMaxTimestamp = 11:00 - 10:30)
```

Here the metric reports 15min while the true pending time is 45min — a systematic 30min underestimate.

#### Error bound

Both directions reduce to the same signed expression (right after a progress update):

```text
estimationError = pendingRecordsLag - truePendingLag
                = (firstPendingTimestamp - lakeMaxTimestamp) - (now - logMaxTimestamp)
```

So the error is just two terms:

```text
overestimate  <= firstPendingTimestamp - lakeMaxTimestamp   (write gap at the tiering boundary)
underestimate <= now - logMaxTimestamp                       (time since the last write)
```

In the vast majority of cases a lake-enabled table is written continuously, and under continuous writing both terms are tiny — the boundary gap is the inter-record write interval (ms-s), and `now ~= logMaxTimestamp` — so the error is essentially 0. The error only grows in the corner cases: a write gap right at the boundary (overestimate) or writes going silent while a backlog is still draining (underestimate, ~= how long since the last write).

Both directions are transient, driven by the same mechanism: a progress update fires at least once per `freshness` and advances `lakeMaxTimestamp` / `lakeLogEndOffset`, which simultaneously closes the overestimate gap and drains the backlog behind the underestimate.

**Overestimate** has a hard per-update bound: since `truePendingLag >= 0` and `pendingRecordsLag <= timestampLag`,

```text
overestimate = estimationError <= pendingRecordsLag <= timestampLag
```

a high estimate can never exceed the reported value, which is itself capped at `timestampLag` (`~ freshness` in steady state). It also decays over time: the boundary hole that inflates `firstPendingTimestamp - lakeMaxTimestamp` is consumed as soon as a progress update tiers past it and advances `lakeMaxTimestamp` to the post-hole timestamp — within ~1-2 `freshness`, since a round always commits within `freshness`. After that the gap is back to the normal inter-record interval.

**Underestimate** at any instant is `now - logMaxTimestamp` (idle since the last write), which grows only while `pendingRecords > 0`. It does not grow forever either: after writes stop no new backlog is added, and since a tiering round commits within `freshness` with its `stoppingOffset` eventually covering the final high watermark, `pendingRecords` reaches 0 within ~1-2 `freshness` (when tiering keeps up), forcing the error back to 0. So the underestimate is a transient bounded by the drain duration of the write-stop backlog — `freshness`-level in steady state. If tiering is genuinely behind, draining takes longer, but then `timestampLag` is large, so the metric still reports an elevated value rather than silently sitting near 0.

| direction | upper bound | how it self-corrects |
|-----------|-------------|----------------------|
| overestimate | `<= timestampLag` (`~ freshness` in steady state) | `lakeMaxTimestamp` advances past the hole within ~1-2 `freshness` |
| underestimate | `<= drain duration of write-stop backlog` (`~ freshness` in steady state) | `pendingRecords -> 0` within ~1-2 `freshness` |

#### Tradeoff

The metric trades a small boundary-gap-level error (and a possible underestimate when writes go silent with backlog undrained) for eliminating the minute-scale idle-then-write pulse of `timestampLag`, with no metric-scrape IO and no complex in-memory window state.

The existing `timestampLag` metric is kept unchanged for compatibility and can serve as an auxiliary diagnostic.

### Anything else?

This avoids alert spikes caused by idle periods followed by new writes. The metric measures the age of the actual lake-tiering backlog instead of the timestamp distance between the latest local record and latest lake-tiered record.

### Willingness to contribute

beryllw:
### Purpose

Linked issue: close #3464

### DIFF
--- a/fluss-server/src/main/java/org/apache/fluss/server/log/LogTablet.java
+++ b/fluss-server/src/main/java/org/apache/fluss/server/log/LogTablet.java
@@ -281,13 +281,21 @@ public final class LogTablet {
     }
 
     /**
+     * Returns the timestamp lag between the latest local log record and the latest tiered lake log
+     * record.
+     */
+    public long getTimestampLag() {
+        return lakeMaxTimestamp < 0L ? -1L : Math.max(0L, localMaxTimestamp() - lakeMaxTimestamp);
+    }
+
+    /**
      * Returns the backlog-aware lag of committed records pending lake tiering in milliseconds.
      *
      * <p>Returns 0 when there are no committed records pending lake tiering or when the pending
      * start time estimate is not initialized yet.
      */
     public long getPendingRecordsLag(long currentTimeMs) {
-        if (estimatedPendingStartTimeMs < 0L) {
+        if (estimatedPendingStartTimeMs < 0L || !hasPendingLakeBacklog()) {
             return 0L;
         }
         return Math.max(0L, currentTimeMs - estimatedPendingStartTimeMs);
@@ -650,10 +658,10 @@ public final class LogTablet {
             clearPendingStartTime();
             return;
         }
-        if (lakeMaxTimestamp < 0L) {
+        long timestampLag = getTimestampLag();
+        if (timestampLag < 0L) {
             return;
         }
-        long timestampLag = Math.max(0L, localMaxTimestamp() - lakeMaxTimestamp);
         long candidatePendingStartTimeMs = Math.max(0L, clock.milliseconds() - timestampLag);
         advancePendingStartTime(candidatePendingStartTimeMs);
     }
--- a/fluss-server/src/main/java/org/apache/fluss/server/log/LogTablet.java
+++ b/fluss-server/src/main/java/org/apache/fluss/server/log/LogTablet.java
@@ -132,6 +132,8 @@ public final class LogTablet {
     private volatile long lakeLogStartOffset = Long.MAX_VALUE;
     private volatile long lakeLogEndOffset = -1L;
     private volatile long lakeMaxTimestamp = -1;
+    // Best-effort estimate under concurrent high-watermark and lake-progress updates.
+    // Metric reads are allowed to observe transient intermediate states.
     private volatile long estimatedPendingStartTimeMs = -1L;
 
     private LogTablet(
@@ -278,13 +280,20 @@ public final class LogTablet {
         return lakeMaxTimestamp;
     }
 
+    /**
+     * Returns the backlog-aware lag of committed records pending lake tiering in milliseconds.
+     *
+     * <p>Returns 0 when there are no committed records pending lake tiering or when the pending
+     * start time estimate is not initialized yet.
+     */
     public long getPendingRecordsLag(long currentTimeMs) {
-        if (!hasPendingRecordsForLakeTiering() || estimatedPendingStartTimeMs < 0L) {
+        if (estimatedPendingStartTimeMs < 0L) {
             return 0L;
         }
         return Math.max(0L, currentTimeMs - estimatedPendingStartTimeMs);
     }
 
+    /** Returns the estimated start time of the current pending lake-tiering backlog for tests. */
     @VisibleForTesting
     public long getEstimatedPendingStartTimeMs() {
         return estimatedPendingStartTimeMs;
@@ -495,7 +504,7 @@ public final class LogTablet {
         if (newHighWatermark.getMessageOffset() < 0) {
             throw new IllegalArgumentException("High watermark offset should be non-negative");
         }
-        boolean hadPendingRecords = hasPendingRecordsForLakeTiering();
+        boolean hadPendingRecords = hasPendingLakeBacklog();
         synchronized (lock) {
             if (newHighWatermark.getMessageOffset() < highWatermarkMetadata.getMessageOffset()) {
                 LOG.warn(
@@ -507,7 +516,7 @@ public final class LogTablet {
             highWatermarkMetadata = newHighWatermark;
             // TODO log offset listener to update log offset.
         }
-        maybeUpdatePendingStartTimeAfterHighWatermarkAdvance(hadPendingRecords);
+        onHighWatermarkAdvanced(hadPendingRecords);
         LOG.trace(
                 "Setting high watermark {} for bucket {}",
                 newHighWatermark,
@@ -614,8 +623,8 @@ public final class LogTablet {
     public void updateLakeLogEndOffset(long lakeLogEndOffset) {
         if (lakeLogEndOffset > this.lakeLogEndOffset) {
             this.lakeLogEndOffset = lakeLogEndOffset;
-            if (!hasPendingRecordsForLakeTiering()) {
-                estimatedPendingStartTimeMs = -1L;
+            if (!hasPendingLakeBacklog()) {
+                clearPendingStartTime();
             }
         }
     }
@@ -623,22 +632,22 @@ public final class LogTablet {
     public void updateLakeMaxTimestamp(long lakeMaxTimestamp) {
         if (lakeMaxTimestamp > this.lakeMaxTimestamp) {
             this.lakeMaxTimestamp = lakeMaxTimestamp;
-            maybeCorrectPendingStartTimeAfterLakeProgress();
+            onLakeProgressUpdated();
         }
     }
 
-    private void maybeUpdatePendingStartTimeAfterHighWatermarkAdvance(boolean hadPendingRecords) {
-        boolean hasPendingRecords = hasPendingRecordsForLakeTiering();
+    private void onHighWatermarkAdvanced(boolean hadPendingRecords) {
+        boolean hasPendingRecords = hasPendingLakeBacklog();
         if (!hasPendingRecords) {
-            estimatedPendingStartTimeMs = -1L;
+            clearPendingStartTime();
         } else if (!hadPendingRecords) {
-            estimatedPendingStartTimeMs = clock.milliseconds();
+            markPendingStartTimeNow();
         }
     }
 
-    private void maybeCorrectPendingStartTimeAfterLakeProgress() {
-        if (!hasPendingRecordsForLakeTiering()) {
-            estimatedPendingStartTimeMs = -1L;
+    private void onLakeProgressUpdated() {
+        if (!hasPendingLakeBacklog()) {
+            clearPendingStartTime();
             return;
         }
         if (lakeMaxTimestamp < 0L) {
@@ -646,16 +655,28 @@ public final class LogTablet {
         }
         long timestampLag = Math.max(0L, localMaxTimestamp() - lakeMaxTimestamp);
         long candidatePendingStartTimeMs = Math.max(0L, clock.milliseconds() - timestampLag);
-        estimatedPendingStartTimeMs =
-                Math.max(estimatedPendingStartTimeMs, candidatePendingStartTimeMs);
+        advancePendingStartTime(candidatePendingStartTimeMs);
     }
 
-    private boolean hasPendingRecordsForLakeTiering() {
+    private boolean hasPendingLakeBacklog() {
         long firstPendingOffset =
                 lakeLogEndOffset >= 0L
                         ? Math.max(lakeLogEndOffset, localLogStartOffset())
                         : localLogStartOffset();
         return firstPendingOffset < getHighWatermark();
+    }
+
+    private void clearPendingStartTime() {
+        estimatedPendingStartTimeMs = -1L;
+    }
+
+    private void markPendingStartTimeNow() {
+        estimatedPendingStartTimeMs = clock.milliseconds();
+    }
+
+    private void advancePendingStartTime(long candidatePendingStartTimeMs) {
+        estimatedPendingStartTimeMs =
+                Math.max(estimatedPendingStartTimeMs, candidatePendingStartTimeMs);
     }
 
     public void loadWriterSnapshot(long lastOffset) throws IOException {
--- a/fluss-server/src/main/java/org/apache/fluss/server/log/LogTablet.java
+++ b/fluss-server/src/main/java/org/apache/fluss/server/log/LogTablet.java
@@ -285,7 +285,7 @@ public final class LogTablet {
      * record.
      */
     public long getTimestampLag() {
-        return lakeMaxTimestamp < 0L ? -1L : Math.max(0L, localMaxTimestamp() - lakeMaxTimestamp);
+        return lakeMaxTimestamp < 0L ? -1L : localMaxTimestamp() - lakeMaxTimestamp;
     }
 
     /**
@@ -662,7 +662,8 @@ public final class LogTablet {
         if (timestampLag < 0L) {
             return;
         }
-        long candidatePendingStartTimeMs = Math.max(0L, clock.milliseconds() - timestampLag);
+        long candidatePendingStartTimeMs =
+                Math.max(0L, clock.milliseconds() - Math.max(0L, timestampLag));
         advancePendingStartTime(candidatePendingStartTimeMs);
     }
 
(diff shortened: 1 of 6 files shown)


fluss-server/src/main/java/org/apache/fluss/server/log/LogTablet.java
```diff
@@ -599,15 +631,63 @@ public void updateLakeLogStartOffset(long lakeHouseLogStartOffset) {
...
     }
 
     public void updateLakeMaxTimestamp(long lakeMaxTimestamp) {
         if (lakeMaxTimestamp > this.lakeMaxTimestamp) {
             this.lakeMaxTimestamp = lakeMaxTimestamp;
+            onLakeProgressUpdated();
         }
     }
 
+    private void onHighWatermarkAdvanced(boolean hadPendingRecords) {
```
luoyuxia: is there any diffeernce between  `hadPendingRecords` and `hasPendingRecords`

fluss-server/src/main/java/org/apache/fluss/server/log/LogTablet.java
```diff
@@ -599,15 +631,63 @@ public void updateLakeLogStartOffset(long lakeHouseLogStartOffset) {
...
+        long timestampLag = getTimestampLag();
+        if (timestampLag < 0L) {
+            return;
+        }
+        long candidatePendingStartTimeMs =
+                Math.max(0L, clock.milliseconds() - Math.max(0L, timestampLag));
+        advancePendingStartTime(candidatePendingStartTimeMs);
+    }
+
+    private boolean hasPendingLakeBacklog() {
```
luoyuxia: hasPendingLakeBacklog -> hasPendingLakeTieringRecords

fluss-server/src/main/java/org/apache/fluss/server/log/LogTablet.java
```diff
@@ -599,15 +631,63 @@ public void updateLakeLogStartOffset(long lakeHouseLogStartOffset) {
...
+        long firstPendingOffset =
```
luoyuxia: can be simplified to 
```
Math.max(lakeLogEndOffset, localLogStartOffset())  < getHighWatermark();
```
?

fluss-server/src/main/java/org/apache/fluss/server/log/LogTablet.java
```diff
@@ -599,15 +631,63 @@ public void updateLakeLogStartOffset(long lakeHouseLogStartOffset) {
...
 
     public void updateLakeMaxTimestamp(long lakeMaxTimestamp) {
         if (lakeMaxTimestamp > this.lakeMaxTimestamp) {
             this.lakeMaxTimestamp = lakeMaxTimestamp;
+            onLakeProgressUpdated();
         }
     }
 
+    private void onHighWatermarkAdvanced(boolean hadPendingRecords) {
+        boolean hasPendingRecords = hasPendingLakeBacklog();
```
luoyuxia: nit: we can skip handle pendingRecords for non-tablelake table

fluss-server/src/main/java/org/apache/fluss/server/log/LogTablet.java
```diff
@@ -599,15 +631,63 @@ public void updateLakeLogStartOffset(long lakeHouseLogStartOffset) {
...
+    private void onLakeProgressUpdated() {
+        if (!hasPendingLakeBacklog()) {
+            clearPendingStartTime();
+            return;
+        }
+        long timestampLag = getTimestampLag();
+        if (timestampLag < 0L) {
+            return;
+        }
+        long candidatePendingStartTimeMs =
```
luoyuxia: nit: Result of 'max' is the same as the second argument making the call meaningless

fluss-server/src/main/java/org/apache/fluss/server/log/LogTablet.java
luoyuxia: actually, I do spend some time to fingure out it. I'm wondering whether the following is better:
```
private void onHighWatermarkUpdated(long previousHighWatermark, long currentHighWatermark) {
      long firstPendingOffset = Math.max(lakeLogEndOffset, localLogStartOffset());

      boolean hadPendingBeforeUpdate = firstPendingOffset < previousHighWatermark;
      boolean hasPendingAfterUpdate = firstPendingOffset < currentHighWatermark;

      if (!hasPendingAfterUpdate) {
          clearPendingStartTime();
      } else if (!hadPendingBeforeUpdate) {
          markPendingStartTimeNow();
      }
  }
```

website/docs/maintenance/observability/monitor-metrics.md
```diff
@@ -890,6 +890,11 @@ Some metrics might not be exposed when using other JVM implementations (e.g. IBM
       <td>The timestamp lag between the latest log record and the latest tiered lake log record for this table bucket, in milliseconds.</td>
       <td>Gauge</td>
     </tr>
+    <tr>
+      <td>pendingRecordsLag</td>
+      <td>The backlog-aware lag for committed records pending lake tiering for this table bucket, in milliseconds. It is 0 when there are no committed records pending lake tiering.</td>
```
luoyuxia: actually, for me, the backlog-aware is not intuitive.. Some thing like this:
```
The elapsed time, in milliseconds, since the oldest committed record in this table bucket became pending for lake tiering. Returns 0 when no committed records are pending lake tiering.
```
?

luoyuxia (commented):
@beryllw Thanks for the pr. Left some comments. PTAL

luoyuxia (approved):
@beryllw Thanks. LGTM

PR #3471, opened 2026-06-10
[spark] use CamelCase in config names

### LINKED ISSUE #3462: [spark] use CamelCase in spark configs
### Search before asking

### Description

as per title

### Willingness to contribute

fresh-borzoni:
closes https://github.com/apache/fluss/issues/3462

making SparkConfig consistent


fresh-borzoni:
@Yohahaha @YannByron @luoyuxia PTAL 🙏

fluss-spark/fluss-spark-common/src/main/scala/org/apache/fluss/spark/SparkFlussConf.scala
```diff
@@ -39,21 +39,21 @@ object SparkFlussConf {
...
     ConfigBuilder
-      .key("scan.startup.mode")
+      .key("scanStartupMode")
       .stringType()
       .defaultValue(StartUpMode.FULL.toString)
       .withDescription("The start up mode when read Fluss table.")
 
   val SCAN_POLL_TIMEOUT: ConfigOption[Duration] =
     ConfigBuilder
-      .key("scan.poll.timeout")
```
Yohahaha, line 49: scan.pollTimeout

Yohahaha (commented):
Sorry for the late reply. 

I think we should keep the current dot-style configuration. For longer configuration names, use camelCase. for `scan.maxRecordsPerPartition`, scan represents the configuration module, and maxRecordsPerPartition represents a specific setting within the scan module. This naming convention is also used in Spark, for example `spark.sql.optimizer.xxx`

Jackeyzhe:
One small question: the deprecation path — is there a plan to keep the old snake_case names working with
a `@Deprecated(since="...")` annotation and a log warning when read, or do you plan a hard 0.x → 0.y cut-off? If the latter, please call out the migration note in the PR description.

Issue #1953, opened 2025-11-10
support tiering for custom database mapping


luoyuxia:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/fluss/issues) and found nothing similar.

### Description

as title

### Willingness to contribute

- [ ] I'm willing to submit a PR!

SML0127:
If no one is working on this, could you assign it to me?

Issue #1952, opened 2025-11-10
Support custom database mapping in coordinator server


luoyuxia:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/fluss/issues) and found nothing similar.

### Description

as title

### Willingness to contribute

- [ ] I'm willing to submit a PR!

LiebingYu:
@luoyuxia I can take this. Let's briefly align on the solution first. My idea is to add a `lakeDatabase` field to `DatabaseRegistration`, so that when a user creates a Fluss database, they can specify the corresponding data lake database themselves.
```java
public class DatabaseDescriptor {

    private final String lakeDatabase;
}
```

Another choise is add this `lakeDatabase` to `customProperties`, but this solution need a lot of specific processing when alter database properties. What your opinion?

luoyuxia:
> [@luoyuxia](https://github.com/luoyuxia) I can take this. Let's briefly align on the solution first. My idea is to add a `lakeDatabase` field to `DatabaseRegistration`, so that when a user creates a Fluss database, they can specify the corresponding data lake database themselves.
> 
> public class DatabaseDescriptor {
> 
>     private final String lakeDatabase;
> }
> Another choise is add this `lakeDatabase` to `customProperties`, but this solution need a lot of specific processing when alter database properties. What your opinion?

I think it's reasable to set in database level since it can simplify user's operation. And we can still allow users to specific the lake database per-table in the future if we found some time users want to map different table in same fluss database to different lake database.

One question is that what will happen when user alter database to set another lakeDatabase? Or, if users already have some tables with lake enabled in this database, but some day it want to alter the database to set lakeDatabase. What will happen for the existing tables with lake enabled?

LiebingYu:
> > [@luoyuxia](https://github.com/luoyuxia) I can take this. Let's briefly align on the solution first. My idea is to add a `lakeDatabase` field to `DatabaseRegistration`, so that when a user creates a Fluss database, they can specify the corresponding data lake database themselves.
> > public class DatabaseDescriptor {
> > ```
> > private final String lakeDatabase;
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > }
> > Another choise is add this `lakeDatabase` to `customProperties`, but this solution need a lot of specific processing when alter database properties. What your opinion?
> 
[...]

PR #3476, opened 2026-06-12
[lake/paimon] Support custom Paimon lake table path

wzx140:
## Summary
- Support mapping a Fluss table to a Paimon lake table with a custom database and/or table name through `table.datalake.database-name` and `table.datalake.table-name`.
- Fall back to the corresponding Fluss database or table name when either mapping option is omitted.
- Use the resolved lake table path for lake table creation and alteration, Paimon tiering writes and commits, Flink `$lake` and metadata-table reads, and Flink union reads.
- Validate custom lake database and table names, and restrict custom lake table path mapping to Paimon.
- Allow path-changing mapping alterations only when datalake is disabled and no tiering progress exists. Reject such alterations after datalake is enabled or tiering progress has been recorded.
- Keep Spark log-only fallback working before a lake snapshot exists, but reject actual Spark lake reads for tables with custom lake path mapping in this phase.
- Document the new mapping options, defaults, and alteration restrictions.

## Test Coverage
- `FlinkCatalogTest#testGetLakeTableWithCustomLakePath`: verifies Flink catalog resolves `$lake` and metadata-table reads to the mapped Paimon database and table.
- `FlinkUnionReadPrimaryKeyTableITCase#testUnionReadWithCustomLakeTablePath`: verifies Flink reads from the mapped Paimon table and merges lake data with the Fluss tail.
- `PaimonTieringITCase#testTieringWithCustomLakeTablePath`: verifies Paimon tiering writes to the mapped table and records the expected snapshot offsets.
- `LakeEnabledTableCreateITCase#testAlterLakePathOptionsValidation`: verifies mapping options can be changed while datalake is disabled and no tiering progress exists, are used when enabling datalake, and cannot change the resolved lake path after datalake is enabled or tiering progress exists.
- `LakeEnabledTableCreateITCase#testAlterLakePathOptionsOnlyForPaimon`: verifies custom lake table path mapping is rejected for non-Paimon lake formats.
- `TableDescriptorValidationTest`: verifies custom lake database and table names are validated and valid mapped names are accepted.
- `SparkLakePaimonLogTableReadTest`: verifies Spark can fall back to log-only reads before a lake snapshot exists and rejects actual lake reads when custom lake path mapping is configured.


fluss-spark/fluss-spark-common/src/main/scala/org/apache/fluss/spark/SparkTable.scala
```diff
@@ -68,6 +68,10 @@ class SparkTable(
         flussConfig.get(SparkFlussConf.SCAN_START_UP_MODE))
       .toUpperCase
     val isFullMode = startupMode == SparkFlussConf.StartUpMode.FULL.toString
+    if (isDataLakeEnabled && LakeTableUtil.hasCustomLakePath(tableInfo.getProperties)) {
+      throw new UnsupportedOperationException(
+        "Custom lake table path is not supported in Spark connector yet.")
+    }
```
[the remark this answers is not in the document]
wzx140: Good point. I narrowed this check to Spark lake-read paths only.

For log/append tables, Spark now rejects custom lake table path only when `datalake.enabled=true`, scan mode is `FULL`, and custom mapping is configured. For primary-key tables, it still rejects when custom mapping is configured because the current Spark scan path uses lake upsert scan.

I also updated the error message to: `Custom lake table path is not supported for Spark lake reads yet.`

fluss-server/src/main/java/org/apache/fluss/server/zk/data/lake/LakeTableHelper.java
```diff
@@ -79,6 +79,24 @@ public void registerLakeTableSnapshotV2(
...
+        zkClient.deleteLakeTable(tableId);
+        LakeTable lakeTable = optLakeTable.get();
+        List<LakeTable.LakeSnapshotMetadata> lakeSnapshotMetadatas =
+                lakeTable.getLakeSnapshotMetadatas();
+        if (lakeSnapshotMetadatas != null) {
+            for (LakeTable.LakeSnapshotMetadata lakeSnapshotMetadata : lakeSnapshotMetadatas) {
+                lakeSnapshotMetadata.discard();
+            }
+        }
+    }
```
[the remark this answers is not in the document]
wzx140: Fixed. `clearLakeTableProgress` now discards the snapshot metadata first and deletes the ZK lake progress node after cleanup succeeds, so an IO failure during discard will not lose the progress reference.

fluss-server/src/main/java/org/apache/fluss/server/zk/ZooKeeperClient.java
```diff
@@ -1365,6 +1365,16 @@ public Optional<LakeTable> getLakeTable(long tableId) throws Exception {
...
 
+    /** Deletes the {@link LakeTable} for the given table ID if it exists. */
+    public void deleteLakeTable(long tableId) throws Exception {
+        String zkPath = LakeTableZNode.path(tableId);
+        try {
+            zkClient.delete().forPath(zkPath);
+        } catch (KeeperException.NoNodeException ignored) {
+            // Ignore if the lake table progress has not been committed yet.
+        }
+    }
```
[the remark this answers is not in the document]
wzx140: Fixed. `deleteLakeTable` now uses `deletingChildrenIfNeeded().forPath(...)` and still ignores `NoNodeException`, so the cleanup is robust if the lake-table znode gains children later.

fluss-lake/fluss-lake-paimon/src/test/java/org/apache/fluss/lake/paimon/LakeEnabledTableCreateWithHiveCatalogITCase.java
```diff
@@ -0,0 +1,198 @@
...
+    public static final FlussClusterExtension FLUSS_CLUSTER_EXTENSION =
+            FlussClusterExtension.builder()
+                    .setNumOfTabletServers(3)
+                    .setClusterConf(initConfig())
+                    .build();
+
+    private static final String DATABASE = "fluss";
+    private static final int BUCKET_NUM = 3;
+
+    private static Catalog paimonCatalog;
```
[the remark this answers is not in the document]
wzx140: Fixed. Added `@AfterAll` cleanup to close `paimonCatalog` when it is non-null and reset it to null.

fluss-lake/fluss-lake-paimon/src/test/java/org/apache/fluss/lake/paimon/LakeEnabledTableCreateWithHiveCatalogITCase.java
```diff
@@ -0,0 +1,198 @@
...
+        conf.setString("datalake.paimon.warehouse", warehousePath);
+        conf.setString(
+                "datalake.paimon.javax.jdo.option.ConnectionDriverName",
+                "org.apache.derby.jdbc.EmbeddedDriver");
+        conf.setString("datalake.paimon.datanucleus.schema.autoCreateAll", "true");
+        conf.setString("datalake.paimon.hive.metastore.schema.verification", "false");
+
+        paimonCatalog =
+                CatalogFactory.createCatalog(
+                        CatalogContext.create(Options.fromMap(extractLakeProperties(conf))));
```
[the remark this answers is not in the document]
wzx140: Fixed by the same `@AfterAll` cleanup added for the static `paimonCatalog`.

pom.xml
```diff
@@ -105,6 +105,12 @@
         <paimon.version>1.3.1</paimon.version>
         <iceberg.version>1.10.1</iceberg.version>
         <hudi.version>1.1.0</hudi.version>
+        <hive.version>2.3.10</hive.version>
+        <datanucleus-api-jdo.version>4.2.4</datanucleus-api-jdo.version>
+        <datanucleus-core.version>4.1.17</datanucleus-core.version>
+        <datanucleus-rdbms.version>4.1.19</datanucleus-rdbms.version>
+        <javax-jdo.version>3.2.0-m3</javax-jdo.version>
+        <derby.version>10.14.2.0</derby.version>
```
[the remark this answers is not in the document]
wzx140: I kept this version for now because it matches the Hive 2.3.x / DataNucleus 4.x test dependency set used by the Hive metastore path in this IT. This dependency is test-scoped only. I can switch it if we confirm a compatible final release works with the embedded Hive metastore setup.

wzx140:
@luoyuxia  Could you please help review this one?

luoyuxia:
@wzx140 Thanks for the work!

#1953 only asks for custom **database** mapping. This PR goes quite a bit beyond that scope — on top of database mapping it also adds table-name mapping, custom `paimon.path`, Spark-read rejection, tiering-progress clearing, and a new Hive-metastore ITCase (pulling in `hive-exec` / `datanucleus` / `derby` test deps), ending up at ~900 lines across 26 files.

Could we start with **just the database-name mapping** that #1953 needs and confirm whether that already satisfies the use case? If so, the rest (table-name mapping, custom `paimon.path`, …) can come as separate follow-ups once we agree they're actually needed. That keeps this PR small, scoped to the issue, and free of the heavy Hive test dependencies.

wzx140:
@luoyuxia I was looking at both #1953 and #1952 together when working on this.

The database mapping part is already in progress, but internally the per-table mapping is actually the higher-priority requirement for us.

To keep the scope manageable, I’m happy to remove the custom paimon.path, Hive ITCase / test dependency changes, and other unrelated pieces from this PR. Then we can focus the review on the mapping functionality first and iterate on the rest separately if needed.

Does that sound reasonable to you?

luoyuxia:
> @luoyuxia I was looking at both #1953 and #1952 together when working on this.
> 
> The database mapping part is already in progress, but internally the per-table mapping is actually the higher-priority requirement for us.
> 
> To keep the scope manageable, I’m happy to remove the custom paimon.path, Hive ITCase / test dependency changes, and other unrelated pieces from this PR. Then we can focus the review on the mapping functionality first and iterate on the rest separately if needed.
> 
> Does that sound reasonable to you?

Thanks. It sounds reasonable to me

wzx140:
@luoyuxia I’ve removed the unrelated changes. The PR is now around 600 lines across 22 files, with roughly half of the diff consisting of new unit tests, so the implementation itself is significantly smaller than before.

If you have time, I’d appreciate another review.

fluss-server/src/main/java/org/apache/fluss/server/coordinator/MetadataManager.java
```diff
@@ -577,6 +580,29 @@ public void alterTableProperties(
...
+                LakeTableUtil.getLakeTablePath(
+                        tableInfo.getTablePath(), newDescriptor.getProperties());
+        if (currentLakeTablePath.equals(newLakeTablePath)) {
+            return;
+        }
+
+        boolean lakeTablePathChangeForbidden =
+                tableInfo.getTableConfig().isDataLakeEnabled()
+                        || zookeeperClient.getLakeTable(tableInfo.getTableId()).isPresent();
+        if (lakeTablePathChangeForbidden) {
```
luoyuxia: Would it meet your requirements if these options could only be set when creating the table and could not be modified afterward?

Supporting path changes would introduce additional complexity around existing lake tables, tiering progress, and data migration

fluss-common/src/main/java/org/apache/fluss/config/ConfigOptions.java
```diff
@@ -1813,6 +1813,20 @@ public class ConfigOptions {
                                     + "The `table.datalake.format` can be pre-defined before enabling `table.datalake.enabled`. This allows the data lake feature to be dynamically enabled on the table without requiring table recreation. "
                                     + "If `table.datalake.format` is not explicitly set during table creation, the table will default to the format specified by the `datalake.format` configuration in the Fluss cluster.");
 
+    public static final ConfigOption<String> TABLE_DATALAKE_DATABASE_NAME =
```
luoyuxia, line 1948: also add the option to document

fluss-server/src/main/java/org/apache/fluss/server/coordinator/MetadataManager.java
wzx140: @luoyuxia  Yes, that would meet my requirements.

The current expectation is that once the Paimon table is created, these name mapping options become immutable. Allowing them to be modified afterward would require handling much more complex scenarios, including existing lake tables, tiering progress, and potential data migration, which we’d prefer to avoid.

fluss-common/src/main/java/org/apache/fluss/config/ConfigOptions.java
wzx140: done

luoyuxia, line 1948: @wuchong Could you please take a look at the public API design here?

This PR introduces two table options, table.datalake.database-name and table.datalake.table-name, to map a Fluss table to a Paimon table with a different database or table name.

wuchong: Thanks @luoyuxia and @wzx140 , the new options `table.datalake.database-name` and `table.datalake.table-name,` look good to me.

fluss-server/src/main/java/org/apache/fluss/server/coordinator/MetadataManager.java
```diff
@@ -577,6 +580,29 @@ public void alterTableProperties(
         }
     }
 
+    private void validateAlterLakeTablePath(TableInfo tableInfo, TableDescriptor newDescriptor)
```
luoyuxia: Based on the discussion above, my understanding is that these two mapping options should be create-time only. However, the current implementation still allows changing them when datalake is disabled and no tiering progress exists, and also allows option changes when the resolved path remains unchanged.

Could we remove these options from ALTERABLE_TABLE_OPTIONS and drop validateAlterLakeTablePath() entirely? That would match the agreed semantics and avoid the additional lifecycle, migration, and in-flight tiering complexity.

fluss-spark/fluss-spark-common/src/main/scala/org/apache/fluss/spark/read/lake/FlussLakeUtils.scala
```diff
@@ -52,6 +52,10 @@ object FlussLakeUtils extends Logging {
       catalogProperties: util.Map[String, String],
       tableProperties: util.Map[String, String],
       tablePath: TablePath): LakeSource[LakeSplit] = {
+    if (LakeTableUtil.hasCustomLakePath(tableProperties)) {
+      throw new UnsupportedOperationException(
```
luoyuxia, line 59: please add todo comment and create a issue for it.

fluss-lake/fluss-lake-paimon/src/test/java/org/apache/fluss/lake/paimon/tiering/PaimonTieringITCase.java
```diff
@@ -188,6 +188,32 @@ void testTiering() throws Exception {
         }
     }
 
+    @Test
+    void testTieringWithCustomLakeTablePath() throws Exception {
```
luoyuxia: This seems redundant with `testUnionReadWithCustomLakeTablePath`, which already starts the tiering job and verifies that data is written to and read from the mapped Paimon table. The additional snapshot-offset assertion here covers generic tiering behavior that is already exercised by `testTiering`. Could we remove this test?

fluss-lake/fluss-lake-paimon/src/test/java/org/apache/fluss/lake/paimon/LakeEnabledTableCreateITCase.java
```diff
@@ -1055,6 +1059,234 @@ void testEnableLakeTableAfterAlterTableProperties() throws Exception {
         admin.alterTable(tablePath, Collections.singletonList(enableLake), false).get();
     }
 
+    @Test
+    void testAlterLakePathOptionsValidation() throws Exception {
```
luoyuxia: If these mapping options are create-time only as discussed above, this test is exercising a state machine that we should not support: altering the path while lake is disabled, allowing changes when the resolved path is unchanged, and gating changes on tiering progress. Could we replace it with a small test asserting that SET/RESET of both mapping options is always rejected after table creation?

fluss-lake/fluss-lake-paimon/src/test/java/org/apache/fluss/lake/paimon/LakeEnabledTableCreateITCase.java
```diff
@@ -1055,6 +1059,234 @@ void testEnableLakeTableAfterAlterTableProperties() throws Exception {
...
+                                        .get())
+                .cause()
+                .isInstanceOf(InvalidAlterTableException.class)
+                .hasMessageContaining(
+                        "cannot be altered when 'table.datalake.enabled' is true or the table has tiering progress")
+                .hasMessageContaining(ConfigOptions.TABLE_DATALAKE_TABLE_NAME.key());
+    }
+
+    @Test
+    void testAlterLakePathOptionsOnlyForPaimon() throws Exception {
```
luoyuxia: This constraint belongs to create-time descriptor validation rather than ALTER behavior. Could we move this coverage to `TableDescriptorValidationTest` by validating a descriptor with a custom lake path against `DataLakeFormat.ICEBERG`, and remove this IT test?

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/catalog/FlinkCatalog.java
```diff
@@ -469,23 +471,19 @@ public CatalogBaseTable getTable(ObjectPath objectPath)
     }
 
     protected CatalogBaseTable getLakeTable(
-            String databaseName,
-            String tableName,
+            TablePath lakeTablePath,
+            String requestedTableName,
```
luoyuxia: I find `requestedTableName` difficult to follow because it represents the original Flink catalog object name (for example, `fluss_table$lake$snapshots`), while this method ultimately accesses a resolved lake object such as `custom_lake_table$snapshots`.

Could we resolve the lake object name at the call site and let `getLakeTable()` only load the resolved object? For example:

```java
String lakeObjectName =
        LakeTableUtil.getLakeTableName(
                lakeTablePath.getTableName(), tableName, LAKE_TABLE_SPLITTER);

return getLakeTable(
        lakeTablePath.getDatabaseName(),
        lakeObjectName,
        tableInfo.getProperties(),
        getLakeCatalogProperties());
```

Then the method can accept `lakeDatabaseName` and `lakeObjectName` directly. This keeps the `$lake`/metadata-table translation at the request-handling layer and avoids passing the ambiguous `requestedTableName` through another layer.

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/lake/LakeTableFactory.java
```diff
@@ -26,23 +28,30 @@
...
     private final LakeFlinkCatalog lakeFlinkCatalog;
 
     public LakeTableFactory(LakeFlinkCatalog lakeFlinkCatalog) {
         this.lakeFlinkCatalog = lakeFlinkCatalog;
     }
 
     public DynamicTableSource createDynamicTableSource(
-            DynamicTableFactory.Context context, String tableName) {
-        ObjectIdentifier originIdentifier = context.getObjectIdentifier();
+            DynamicTableFactory.Context context, String requestedTableName) {
```
luoyuxia: I’m a bit confused by the rename from `tableName` to `requestedTableName`; the new name is not very intuitive to me.

As I understand it, this value is the original `tableName` after `$lake` has been removed:

```text
fluss_table$lake$snapshots
    -> fluss_table$snapshots
```

It is neither the complete name requested by the user nor the actual lake table name after applying the custom mapping. Could we keep the original `tableName`, or use a more precise name such as `logicalLakeObjectName`?

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/lake/LakeTableFactory.java
```diff
@@ -26,23 +28,30 @@
 import org.apache.flink.table.factories.FactoryUtil;
 
 import java.util.Collections;
+import java.util.Map;
 
 /** A factory to create {@link DynamicTableSource} for lake table. */
 public class LakeTableFactory {
+
+    private static final String FLUSS_CONF_PREFIX = "fluss.";
```
luoyuxia: This dependency on the persisted `fluss.*` options feels implicit and fragile. `LakeTableFactory` assumes that every lake format persists the original Fluss options using this exact prefix. If we later stop persisting these options, filter them out, or change the prefix, custom lake path resolution will stop working and may silently fall back to the original Fluss path.

Could we resolve the final lake identifier once in `FlinkCatalog`, where `TableInfo#getLakeTablePath()` and the metadata-table suffix are already available, and attach it as transient metadata to the returned `CatalogTable` copy?

For example:

```java
Map<String, String> options = new HashMap<>(lakeTable.getOptions());
options.put(RESOLVED_LAKE_DATABASE, lakeTablePath.getDatabaseName());
options.put(RESOLVED_LAKE_OBJECT, lakeObjectName);
return lakeTable.copy(options);
```

`LakeTableFactory` can then consume and remove these values before delegating to the Paimon/Iceberg factory:

```java
Map<String, String> options =
        new HashMap<>(context.getCatalogTable().getOptions());

String lakeDatabase = checkNotNull(options.remove(RESOLVED_LAKE_DATABASE));
String lakeObject = checkNotNull(options.remove(RESOLVED_LAKE_OBJECT));

ObjectIdentifier lakeIdentifier =
        ObjectIdentifier.of(
                context.getObjectIdentifier().getCatalogName(),
                lakeDatabase,
                lakeObject);
```

These metadata values would exist only on the in-memory table returned to the Flink planner and would be removed before creating the delegated factory context. This avoids another metadata RPC while removing the dependency on the persisted `fluss.*` properties.

fluss-common/src/main/java/org/apache/fluss/metadata/LakeTableUtil.java
```diff
@@ -0,0 +1,101 @@
...
+import java.util.Optional;
+
+/** Utility methods for resolving external lake table metadata. */
+@Internal
+public final class LakeTableUtil {
+
+    private LakeTableUtil() {}
+
+    /** Returns the table path used to access the external datalake table. */
+    public static TablePath getLakeTablePath(
```
luoyuxia: Could we simplify the path API to expose a single resolver that accepts `ReadableConfig`?

```java
public static TablePath resolveLakeTablePath(
        TablePath flussTablePath, ReadableConfig tableConfig)
```

Both `Configuration` and the table properties held by `TableInfo` can use this API, so the `Map`/`Configuration` overloads and the private optional getters are not needed.

I also find `hasCustomLakePath()` ambiguous: it currently checks whether either option is present, not whether the resolved lake path actually differs from the Fluss path. An explicit identity mapping would therefore be reported as custom. Callers can express the intended semantics directly: use `Configuration#contains(...)` for option-presence validation, and compare the resolved path with the Fluss path when checking for an actual path change.

I tried this approach, including the affected call sites and a focused test, in [ac6cd522f](https://github.com/luoyuxia/fluss/commit/ac6cd522f1c161e802a354c1b15f2875796eb1d1) for reference.

luoyuxia (commented):
@wzx140 I left some comments. PTAL.

fluss-server/src/main/java/org/apache/fluss/server/coordinator/MetadataManager.java
wzx140: @luoyuxia The mapping options may not be suitable as strictly create-time-only because a Fluss table can be created without datalake enabled. Users should be able to decide the target Paimon database and table when enabling datalake through ALTER TABLE.

I kept the options alterable with the current lifecycle restrictions: changing the resolved lake path is rejected once datalake is enabled or tiering progress exists. The tiering-progress check also covers tables where datalake was previously enabled and later disabled, preventing the mapping from being changed after data has already been tiered and preserving tiered-data integrity.

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/lake/LakeTableFactory.java
wzx140: @luoyuxia Removed requestedTableName from LakeTableFactory. The factory no longer needs to interpret the logical Flink table name or remove $lake itself.

It now receives the already-resolved lake identifier through the CatalogTable metadata prepared by FlinkCatalog RESOLVED_LAKE_DATABASE/RESOLVED_LAKE_OBJECT.

luoyuxia:
@wzx140 I append two commits to refine the code. PTAL

luoyuxia (approved):

luoyuxia (approved):

Issue #203, opened 2024-12-16
[Feature] Upgrade programing language to Java 17


wuchong:
### Search before asking

- [X] I searched in the [issues](https://github.com/alibaba/fluss/issues) and found nothing similar.

### Motivation

Currently, the default programming language of Fluss is still Java 8. We would like to upgrade the programming language used in this project to Java 17. This upgrade will provide several benefits, including using the latest version of spotless, Google Java Format, and dependencies, and using new language features, and long-term support. 

However, we can't directly upgrade to Java 17, because most Flink users are still using Java 8 (Flink 2.0 turns to default Java 11), this requires Fluss release client/connector jars in Java 8 or Java 11 to be compatible with Flink. 

### Solution

Therefore, this is going to be a long work, and I would break it down into several phases:

## Phase 1
Upgrade the local development environment to Java 17, but still compile and release the project in Java 8. 

- [ ] CI build test on Java 8 with spotless skipped and compile checks on Java 17
- [ ] #114  (default local language to Java 17)

## Phase 2
Release for Java 8 & 11 & 17. 

- [ ] CI build test on Java 8 & 11 & 17 
- [ ] Release binary distribution for only Java 17
- [ ] Release maven artifacts for both Java 8 and Java 11 and Java 17

## Phase 3
Drop the legacy Java version support for Flink. Upgrade the language level to Java 17. 

- [ ] Remove CI build for Java 8 and Java 11

### Willingness to contribute

- [ ] I'm willing to submit a PR!

psxjoy:
I think we should expedite the deprecation of JDK 8, given that Checkstyle only supports it up to version 9.6, with later versions requiring JDK 11 as the minimum（linked #220 ）.
Moreover, certain features are exclusively supported in JDK 11 and later.

我认为我们应该加快 JDK 8 的弃用，因为 Checkstyle 仅支持到 9.6 版本，而更高版本至少需要 JDK 11。
而且目前有些功能只有在jdk11中才有对应的特性。

simonsssu:
Just one comment for phase-1,  some new features like Pattern Matching for switch in JDK17 may be involved in local dev environment while compile and release will fail with jdk8.

Samrat002:
I'm curious why we shouldn't upgrade to Java 21 or higher?
If anyone had examined the effort estimate more closely previously, I'm not sure, looking for more context

litiliu:
[...]

PR #114, opened 2024-12-04
Allow project to build with Java 17 and 21

geoand:
The formatting changes and a result of the new version of Google Java Format being used, which in turn
is needed for spotless to build on Java 17 and 21.

This change makes no guarantee that project actually runs on 17 or 21, so it's only a very small first step

PR #3477, opened 2026-06-12
[build] CI build test on Java 17 and 21 (#203 Phase 3)

vasiliy-mikhailov:
## Purpose

Phase 3 of #203: CI build-test coverage for Java 17 and Java 21 (default and release stay Java 11; no `target.java.version` is changed). As @wuchong noted when closing #114, building on 17/21 largely works already — this PR adds the missing CI coverage from the Phase-3 checklist, plus the two test-level fixes that running the suite on 17/21 surfaces.

## Changes

- `pom.xml`: add `--add-exports=java.rmi/sun.rmi.registry=ALL-UNNAMED` to the existing `extraJavaTestArgs` block — `JMXServer$JmxRegistry` extends the JDK-internal `sun.rmi.registry.RegistryImpl`, which JDK 17+ strong encapsulation rejects (`IllegalAccessError: superclass access check failed`; JDK 8/11 only warn). Without it the whole `fluss-server` suite errors on 17/21. Harmless on 8/11 thanks to the existing `-XX:+IgnoreUnrecognizedVMOptions`.
- `fluss-lake/fluss-lake-hudi/pom.xml`: add `org.apache.hadoop:hadoop-client-runtime` at **test** scope — Hudi's Flink bundle references hadoop-client-api shaded classes (`org.apache.hadoop.shaded.com.ctc.wstx.*`); without the runtime jar `HudiLakeCatalogTest` can fail with `NoClassDefFoundError` depending on which StAX provider wins. Verified green on 11 and 21 with the dependency.
- `ci.yaml`: add `compile-on-jdk17` / `compile-on-jdk21` (compile-only, mirroring `compile-on-jdk8`) so every PR proves the build on 17/21 cheaply.
- `nightly.yaml`: add full `build-on-jdk17` / `build-on-jdk21` test runs via `ci-template`.

## Spotless

The 17/21 jobs pass `-Dspotless.check.skip=true`, following the precedent #204 established for the Java 8 leg: google-java-format 1.15.0 cannot run on JDK 21 (`JCTree$JCImport.getQualifiedIdentifier` `NoSuchMethodError`), and gjf versions that support 21 require a Java 17 runtime — no gjf version spans 11 and 21, so the format check stays on the default-JDK jobs until Phase 4.

## A linkage hazard worth a maintainer decision (not changed here)

The build uses `maven.compiler.source`/`target`, not `release`. Compiling on JDK 17/21 therefore links against newer JDK APIs even with target 11 — we reproduced jars built on 21 crashing on a Java 11 runtime with `NoSuchMethodError: java.nio.MappedByteBuffer.duplicate()` (covariant override added in JDK 13). If artifacts may ever be built on 17/21, `maven.compiler.release=11` is the fix; we left it out because it interacts with the `java8` profile (javac 8 has no `--release`) and deserves your preferred treatment.

## Verification (local, Temurin 11/17/21)

- `mvn -T 1C clean install -DskipTests`: green on 11, 17 (both pristine) and 21 (with the spotless skip).
- Full-reactor `mvn test` on 17 and 21: identical outcomes to the Java 11 baseline run, with `fluss-server` (1048 tests) green on all three after the `--add-exports` fix, and `HudiLakeCatalogTest` green on 11/21 after the dependency fix.
- Pre-existing, JDK-independent: `fluss-fs-hdfs` fails test discovery identically on 11/17/21 under a plain `mvn test` (it needs the staged `-Ptest-*` invocation); `fluss-spark-ut`'s 2-minute lake-sync waits time out under our shared-load environment — the added nightly jobs on clean runners are the right adjudicator for it.

### DIFF
--- a/fluss-docgen/website/docs/_configs/_partial_config.mdx
+++ b/fluss-docgen/website/docs/_configs/_partial_config.mdx
@@ -1,0 +1,359 @@
+{/* This file is auto-generated. Do not edit directly. */}
+
+## Acl Configurations
+
+| Key | Default | Type | Description |
+| :--- | :--- | :--- | :--- |
+| `acl.notification.expiration-time` | `15 min` | Duration | The duration for which ACL notifications are valid before they expire. This configuration determines the time window during which an ACL notification is considered active. After this duration, the notification will no longer be valid and will be discarded. The default value is 15 minutes. This setting is important to ensure that ACL changes are propagated in a timely manner and do not remain active longer than necessary. |
+
+## Advertised Configurations
+
+| Key | Default | Type | Description |
+| :--- | :--- | :--- | :--- |
+| `advertised.listeners` | `none` | String | The externally advertised address and port for client connections. Required in distributed environments when the bind address is not publicly reachable. Format matches `bind.listeners` (listener_name://host:port). Defaults to the value of `bind.listeners` if not explicitly configured. |
+
+## Allow Configurations
+
+| Key | Default | Type | Description |
+| :--- | :--- | :--- | :--- |
+| `allow.create.log.tables` | `true` | Boolean | Whether to allow creation of log tables. When set to false, attempts to create log tables (tables without primary key) will be rejected. The default value is true. |
+| `allow.create.kv.tables` | `true` | Boolean | Whether to allow creation of kv tables (primary key tables). When set to false, attempts to create kv tables (tables with primary key) will be rejected. The default value is true. |
+
+## Authorizer Configurations
+
+| Key | Default | Type | Description |
+| :--- | :--- | :--- | :--- |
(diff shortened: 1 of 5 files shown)


fluss-docgen/website/docs/_configs/_partial_config.mdx
```diff
@@ -0,0 +1,359 @@
+{/* This file is auto-generated. Do not edit directly. */}
```
pbanakar, line 1: Hi @vasiliy-mikhailov, I noticed this file is auto-generated during the build process and is typically related to documentation generation. It doesn't seem directly related to the Java 17/21 CI changes in this PR.

Could you clarify whether this change is intentional or if it was generated as part of a local build?

PR #3482, opened 2026-06-12
[server] Fix tableCount metric after dropping partitioned tables

### LINKED ISSUE #3481: [server] tableCount metric stays inflated after dropping partitioned tables
### Search before asking

### Fluss version

0.9.0 (latest release)

## Description

After a partitioned table is dropped, the coordinator's `tableCount`
gauge often does not decrement to the correct value. The dropped table
lingers in `CoordinatorContext.tablesToBeDeleted` (and therefore in
`tablePathById`) until some unrelated replica-deletion response happens
to invoke `TableManager.resumeDeletions()`. If no other table is being
dropped, the entry can stay there indefinitely.

## Reproduction

1. Create 5 tables, 2 of which are partitioned.
2. Drop both partitioned tables. → `tableCount` reads `4` instead of `3`.
3. Drop one of the remaining non-partitioned tables. → `tableCount`
   suddenly drops to `2` (because that drop's stop-replica response
   "piggy-backs" `resumeDeletions()` and finally evicts the partitioned
   tables together).

A pure case can be reproduced by dropping a partitioned table that has
**zero partitions**: the entry never leaves `tablesToBeDeleted`.

## Root cause

`CoordinatorEventProcessor.processDropTable` calls
`TableManager.onDeleteTable()`, which is a no-op for partitioned tables
(their buckets live under `partitionAssignments`, not
`tableAssignments`). Because partitioned tables also have no
table-level replicas, `getAllReplicasForTable(tableId)` returns empty
and `areAllReplicasInState(.., ReplicaDeletionSuccessful)` is
**vacuously true** — i.e. the table is *ready* to be cleaned up the
moment it enters `tablesToBeDeleted`. The bug is purely that nothing
triggers `resumeDeletions()` for the partitioned-table path.

### Solution

_No response_

swuferhong:
### Purpose

Linked issue: close #3481 

processDropTable's `onDeleteTable()` is a no-op for partitioned tables
(their buckets live under partitionAssignments, not tableAssignments)
and they have no table-level replicas, so areAllReplicasInState is
vacuously true the moment the table enters tablesToBeDeleted. Without
an explicit trigger, resumeDeletions() runs only when some unrelated
stop-replica response piggy-backs it -- which can be never if no other
drop is in flight. The dropped table then lingers in tablesToBeDeleted
and tablePathById, inflating the tableCount gauge until something
else gets dropped.

### DIFF
--- a/fluss-server/src/main/java/org/apache/fluss/server/coordinator/event/CoordinatorEventManager.java
+++ b/fluss-server/src/main/java/org/apache/fluss/server/coordinator/event/CoordinatorEventManager.java
@@ -32,6 +32,7 @@ import org.apache.fluss.utils.concurrent.ShutdownableThread;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import java.util.Set;
 import java.util.concurrent.LinkedBlockingQueue;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.locks.Lock;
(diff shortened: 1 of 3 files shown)


fluss-server/src/main/java/org/apache/fluss/server/coordinator/event/CoordinatorEventManager.java
```diff
@@ -119,7 +120,19 @@ private void updateMetricsViaAccessContext() {
...
+                            // from tablePathById). We dedup by tableId rather than subtracting
+                            // sizes so the result is robust even if tablesToBeDeleted ever
+                            // drifts out of sync with tablePathById -- it can never go negative
+                            // and only counts ids that are truly still in tablePathById.
+                            Set<Long> tablesToBeDeleted = context.getTablesToBeDeleted();
+                            int tableCount = 0;
+                            for (Long tableId : context.allTables().keySet()) {
+                                if (!tablesToBeDeleted.contains(tableId)) {
+                                    tableCount++;
+                                }
```
wuchong: We should reverse the lookup by iterating over `tablesToBeDeleted` and checking for existence in `context.allTables()`. If a match is found, decrement `tableSize`. This approach significantly improves performance.

```java
Set<Long> allTables = context.allTables().keySet();
int tableCount = allTables.size();
for (Long toDelete : context.getTablesToBeDeleted()) {
    if (allTables.contains(toDelete)) {
        tableCount--;
    }
}
```

wuchong (approved):
I pushed a commit to fix this comment. Please take another look.

PR #3486, opened 2026-06-15
[security] Redact sensitive config values in startup logs

### LINKED ISSUE #3485: [security] Redact sensitive configuration values in startup logs
### Search before asking

### Fluss version

main branch

### Please describe the bug

`GlobalConfiguration` logs every loaded configuration property during startup. This can leak credentials when sensitive configuration values are provided as plain strings.

For example, when an S3 access key is configured:

```yaml
fs.s3a.access.key: <access-key>
```

the startup log currently includes the raw credential value:

```text
Loading configuration property: fs.s3a.access.key=<access-key>
```

This is a security issue because credentials can be exposed to log aggregation systems and users who only have log access.

The current masking logic only hides values that are already represented as `Password` objects. However, configuration loaded from YAML is added through `setString`, so arbitrary sensitive keys such as `fs.s3a.access.key` are logged as plain strings.

### Expected behavior

Sensitive configuration values should be redacted when configuration properties are logged.

The redacted value should always be the fixed string:

```text
******
```

For example:

```text
Loading configuration property: fs.s3a.access.key=******
```

The matching should be case-insensitive. At minimum, `fs.s3a.access.key` and keys containing or matching the following sensitive patterns should be redacted, following Flink's sensitive key handling:

- `password`
- `secret`
- `fs.azure.account.key`
- `apikey`
- `api-key`
- `auth-params`
- `service-key`
- `token`
- `basic-auth`
- `jaas.config`
- `http-headers`

### Possible solution

Add a centralized helper for detecting sensitive configuration keys and use it in `GlobalConfiguration.logConfiguration`.

The existing `Password.HIDDEN_CONTENT` constant already uses the desired fixed redaction value (`******`) and can be reused.

litiliu:
## Summary
- Redact sensitive configuration values when `GlobalConfiguration` logs startup properties.
- Match sensitive keys case-insensitively using Flink-style substring matching.
- Cover Fluss access-key variants such as `fs.s3a.access.key`, `s3.access-key`, and `fs.oss.accessKeyId`.

Fixes #3485

## Test Plan
- `mvn -pl fluss-common -Dtest=ConfigurationTest,GlobalConfigurationTest test`

### DIFF
--- a/fluss-common/src/main/java/org/apache/fluss/config/ConfigurationUtils.java
+++ b/fluss-common/src/main/java/org/apache/fluss/config/ConfigurationUtils.java
@@ -17,6 +17,7 @@
 
 package org.apache.fluss.config;
 
+import org.apache.fluss.utils.CollectionUtils;
 import org.apache.fluss.utils.TimeUtils;
 
 import javax.annotation.Nonnull;
@@ -38,6 +39,37 @@ import static org.apache.fluss.config.StructuredOptionsSplitter.escapeWithSingle
 
 /** Utility class for {@link Configuration} related helper functions. */
 public class ConfigurationUtils {
+
+    private static final String[] SENSITIVE_KEY_PARTS = {
+        "password",
+        "secret",
+        "fs.azure.account.key",
+        "apikey",
+        "api-key",
+        "auth-params",
+        "service-key",
+        "token",
+        "basic-auth",
+        "jaas.config",
+        "http-headers",
+        "private.key",
+        "private-key",
+        "access.key",
+        "access-key",
+        "access_key",
+        "accesskey"
+    };
+
+    // Exact allowlist for non-secret options that match broad sensitive key parts.
+    private static final Set<String> NON_SENSITIVE_KEYS =
+            Arrays.stream(
+                            new String[] {
+                                ConfigOptions.FILESYSTEM_SECURITY_TOKEN_RENEWAL_RETRY_BACKOFF.key(),
+                                ConfigOptions.FILESYSTEM_SECURITY_TOKEN_RENEWAL_TIME_RATIO.key()
+                            })
+                    .map(key -> key.toLowerCase(Locale.ROOT))
+                    .collect(Collectors.toSet());
+
     // --------------------------------------------------------------------------------------------
     //  Type conversion
     // --------------------------------------------------------------------------------------------
@@ -79,6 +111,47 @@ public class ConfigurationUtils {
         }
 
         throw new IllegalArgumentException("Unsupported type: " + clazz);
+    }
+
+    /**
+     * Returns a log-safe representation for a configuration value.
+     *
+     * @param key configuration key
+     * @param value configuration value
+     * @return {@link Password#HIDDEN_CONTENT} for sensitive values, otherwise the original value
+     */
+    public static Object hideSensitiveValue(String key, Object value) {
+        if (value == null) {
+            return null;
+        }
+        return value instanceof Password || isSensitiveKey(key) ? Password.HIDDEN_CONTENT : value;
+    }
+
+    /**
+     * Returns a copy of the given configuration map with sensitive values hidden.
+     *
+     * @param values configuration key/value pairs
+     * @return a new map containing log-safe values
+     */
+    public static Map<String, String> hideSensitiveValues(Map<String, String> values) {
+        Map<String, String> hiddenValues =
+                CollectionUtils.newHashMapWithExpectedSize(values.size());
+        values.forEach(
+                (key, value) -> hiddenValues.put(key, (String) hideSensitiveValue(key, value)));
+        return hiddenValues;
+    }
+
+    static boolean isSensitiveKey(String key) {
+        String lowerKey = key.toLowerCase(Locale.ROOT);
+        if (NON_SENSITIVE_KEYS.contains(lowerKey)) {
+            return false;
+        }
+        for (String sensitiveKeyPart : SENSITIVE_KEY_PARTS) {
+            if (lowerKey.contains(sensitiveKeyPart)) {
+                return true;
+            }
+        }
+        return false;
     }
 
     @SuppressWarnings("unchecked")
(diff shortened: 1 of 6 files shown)


litiliu:
@Prajwal-banakar PTAL

pbanakar:
HI @litiliu Thanks for the fix, this looks good to me and i've one small question, the implementation uses substring matching (contains) for all sensitive key parts. Is this intentionally aligned with Flink's behavior? I'm asking because patterns such as token and secret may also match non-sensitive configuration keys.

litiliu:
> HI @litiliu Thanks for the fix, this looks good to me and i've one small question, the implementation uses substring matching (contains) for all sensitive key parts. Is this intentionally aligned with Flink's behavior? I'm asking because patterns such as token and secret may also match non-sensitive configuration keys.

@Prajwal-banakar Yes, this is intentional and aligned with Flink's existing behavior. The implementation follows the same semantics as `GlobalConfiguration.isSensitive(...)`, which lowercases the key and checks whether it contains any of the configured sensitive key parts.

This can lead to conservative masking for keys containing terms like `token` or `secret`, even if they are not actually sensitive. I think that is acceptable here because it is safer to over-mask than to accidentally expose credentials, and it keeps the behavior consistent with the rest of Flink's configuration/logging handling. If we want stricter matching in the future, it should probably be changed centrally in Flink's sensitive-key detection logic rather than only at this call site.

litiliu:
@luoyuxia  please help review

litiliu:
@fresh-borzoni  PTAL

fluss-common/src/main/java/org/apache/fluss/config/ConfigurationUtils.java
```diff
@@ -38,6 +38,25 @@
 
 /** Utility class for {@link Configuration} related helper functions. */
 public class ConfigurationUtils {
+
+    private static final String[] SENSITIVE_KEY_PARTS = {
```
fresh-borzoni, line 43: fs.gs.auth.service.account.private.key (GCS inline auth, forwarded via fs.gs.*) matches none of these. 
Shall we add private.key / private-key?

fluss-common/src/main/java/org/apache/fluss/config/GlobalConfiguration.java
```diff
@@ -126,7 +126,7 @@ private static void logConfiguration(String prefix, Configuration config) {
                                 "{} configuration property: {}={}",
                                 prefix,
                                 key,
-                                value instanceof Password ? Password.HIDDEN_CONTENT : value));
+                                ConfigurationUtils.hideSensitiveValue(key, value)));
```
fresh-borzoni, line 129: Only the startup path is redacted. datalake.* is a dynamic-reconfigurable prefix and carries secret-key/access-key, so it looks the same key you mask here logs in cleartext when changed at runtime. 

Shall we pull hideSensitiveValue into a shared formatter and reuse it there?

fluss-common/src/main/java/org/apache/fluss/config/ConfigurationUtils.java
fresh-borzoni, line 43: Also #3526 adds a second S3-cred list that already diverges, is it worth unifying?

fluss-common/src/main/java/org/apache/fluss/config/ConfigurationUtils.java
```diff
@@ -38,6 +38,25 @@
...
+        "password",
+        "secret",
+        "fs.azure.account.key",
+        "apikey",
+        "api-key",
+        "auth-params",
+        "service-key",
+        "token",
```
fresh-borzoni, line 51: token also masks client.filesystem.security.token.renewal.backoff/time-ratio - non-secrets you'd want visible when debugging cred refresh, WDYT?

fresh-borzoni (commented):
@litiliu Thank you for the PR, left some comments, PTAL

fluss-common/src/main/java/org/apache/fluss/config/ConfigurationUtils.java
litiliu: Good point. I agree it is worth unifying the S3 credential detection, but I would prefer to handle that in #3526 or a focused follow-up.

The matching in this PR is intentionally generic and conservative for log redaction, following Flink-style sensitive-key substring matching. The #3526 logic is S3-specific and should probably use a more precise matcher for supported S3 config prefixes/suffixes, such as `s3.`, `s3a.`, `fs.s3a.` plus credential-bearing suffixes.

I will keep this PR scoped to the generic redaction path and avoid folding a filesystem config refactor into it. We can extract a shared S3 credential matcher separately so #3526 and the S3 filesystem code do not drift.
What do you think？

litiliu:
@fresh-borzoni Thanks for the careful review. I pushed updates for the actionable comments:

- Added `private.key` / `private-key` to the sensitive-key matching and covered the GCS private key cases in tests.
- Reused `ConfigurationUtils.hideSensitiveValue(s)` for dynamic reconfiguration logs, so `datalake.*` credential changes are redacted as well.
- Added an exact non-sensitive allowlist for the token-renewal configs using existing `ConfigOptions`, normalized case-insensitively, with lowercase/uppercase test coverage.

For the S3 credential matcher unification with #3526, I’d prefer to keep that as a focused follow-up since that path is S3-specific while this PR keeps the generic log-redaction behavior.

fresh-borzoni (commented):
@litiliu Thank you for the changes, LGTM overall, one comment:
ZooKeeperClient.upsertEntityConfigs seems to also log secrets by the glance, shall we fix it as well, WDYT?

litiliu:
> @litiliu Thank you for the changes, LGTM overall, one comment: ZooKeeperClient.upsertEntityConfigs seems to also log secrets by the glance, shall we fix it as well, WDYT?

Good point, Thanks you,  Already fixed, PTAL when you have time again.

fresh-borzoni (approved):
@litiliu Thank you, LGTM 👍

PR #3491, opened 2026-06-16
[test] Fix disk usage collector test

hanliu0830:
### Purpose

Linked issue: close #xxx

Fix a flaky assertion in `DiskUsageCollectorTest#testMultipleDirsOnSameFileStoreDeduplicated`.

The test creates two collectors over the same underlying `FileStore` and expects the collected disk usage to match. Since the two collectors sample usage sequentially, the values can differ by a tiny floating-point amount and make the test flaky.

### Brief change log

- Use AssertJ `isCloseTo(..., offset(1e-6))` instead of exact equality for disk usage comparison.
- Keep the test semantics unchanged: multiple directories on the same `FileStore` should still be deduplicated.


swuferhong (approved):
LGTM

PR #3494, opened 2026-06-18
[FLUSS-3483][coordinator] Reject lake snapshot commits for dropped tables

### LINKED ISSUE #3483: commitLakeTableSnapshot writes stale snapshot path after dropTable, leaving the tiering pipeline permanently stuck
### Search before asking

### Fluss version

main (development)

### Please describe the bug 🐞

When a table is dropped while a lake-tiering commit is in flight (or queued on the tiering client), the coordinator can:
1. Resurrect the dropped tableId's /laketable znode at /fluss/tabletservers/tables/<oldId>/laketable, because commitLakeTableSnapshot does no validity check against the table's lifecycle state.

2. Leak the old tableId into the new snapshot pointer: after the table is recreated with a new tableId, /fluss/tabletservers/tables/<newId>/laketable can contain a tiered_offsets URI pointing at the old <old-table>-<oldId>/metadata/<uuid>.offsets. Since the cleanup pipeline removed that file (or the file was never re-stored after the recreate), every subsequent getLakeSnapshot for the new tableId fails with FileNotFoundException, and the TieringSourceEnumerator retries forever.

The bug is silent: writes and reads of the live table continue to work, but the lake-tiering service is permanently broken for that table until the cluster state is manually repaired.

```text
2026-06-13 19:53:37,373 WARN  org.apache.fluss.flink.tiering.source.enumerator.TieringSourceEnumerator [] - Fail to generate Tiering splits for table cdstream_db.timeseries.
org.apache.flink.util.FlinkRuntimeException: Failed to get table snapshot for table cdstream_db.timeseries
	at org.apache.fluss.flink.tiering.source.split.TieringSplitGenerator.generateTableSplits(TieringSplitGenerator.java:78) ~[fluss-flink-2.2-1.0-SNAPSHOT.jar:1.0-SNAPSHOT]
	at org.apache.fluss.flink.tiering.source.enumerator.TieringSourceEnumerator.generateTieringSplits(TieringSourceEnumerator.java:453) ~[fluss-flink-2.2-1.0-SNAPSHOT.jar:1.0-SNAPSHOT]
	at org.apache.fluss.flink.tiering.source.enumerator.TieringSourceEnumerator.generateAndAssignSplits(TieringSourceEnumerator.java:363) ~[fluss-flink-2.2-1.0-SNAPSHOT.jar:1.0-SNAPSHOT]
	at org.apache.flink.runtime.source.coordinator.ExecutorNotifier.lambda$notifyReadyAsync$3(ExecutorNotifier.java:131) ~[flink-dist-2.2.1.jar:2.2.1]
	at org.apache.flink.util.ThrowableCatchingRunnable.run(ThrowableCatchingRunnable.java:40) ~[flink-dist-2.2.1.jar:2.2.1]
	at org.apache.flink.util.MdcUtils.lambda$wrapRunnable$1(MdcUtils.java:70) ~[flink-dist-2.2.1.jar:2.2.1]
	at java.base/java.util.concurrent.Executors$RunnableAdapter.call(Unknown Source) ~[?:?]
	at java.base/java.util.concurrent.FutureTask.run(Unknown Source) ~[?:?]
	at java.base/java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.run(Unknown Source) ~[?:?]
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(Unknown Source) ~[?:?]
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(Unknown Source) ~[?:?]
	at java.base/java.lang.Thread.run(Unknown Source) [?:?]
Caused by: java.util.concurrent.ExecutionException: org.apache.fluss.exception.UnknownServerException: org.apache.fluss.exception.FlussRuntimeException: Failed to get lake table snapshot for table: cdstream_db.timeseries, table id: 62
	at org.apache.fluss.server.RpcServiceBase.lambda$getLakeSnapshot$2(RpcServiceBase.java:541)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(Unknown Source)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(Unknown Source)
	at java.base/java.lang.Thread.run(Unknown Source)
Caused by: java.io.FileNotFoundException: No such file or directory: s3://fluss/remote-data/lake/cdstream_db/timeseries-58/metadata/cf19d6cc-4a98-4ce0-b518-67dd182fbbe6.offsets
	at org.apache.hadoop.fs.s3a.S3AFileSystem.s3GetFileStatus(S3AFileSystem.java:3866)
	at org.apache.hadoop.fs.s3a.S3AFileSystem.innerGetFileStatus(S3AFileSystem.java:3688)
	at org.apache.hadoop.fs.s3a.S3AFileSystem.extractOrFetchSimpleFileStatus(S3AFileSystem.java:5401)
	at org.apache.hadoop.fs.s3a.S3AFileSystem.open(S3AFileSystem.java:1465)
	at org.apache.hadoop.fs.s3a.S3AFileSystem.open(S3AFileSystem.java:1441)
	at org.apache.hadoop.fs.FileSystem.open(FileSystem.java:976)
	at org.apache.fluss.fs.hdfs.HadoopFileSystem.open(HadoopFileSystem.java:72)
	at org.apache.fluss.fs.hdfs.HadoopFileSystem.open(HadoopFileSystem.java:39)
	at org.apache.fluss.fs.PluginFileSystemWrapper$ClassLoaderFixingFileSystem.open(PluginFileSystemWrapper.java:95)
	at org.apache.fluss.server.zk.data.lake.LakeTable.toLakeTableSnapshot(LakeTable.java:203)
	at org.apache.fluss.server.zk.data.lake.LakeTable.toLakeTableSnapshot(LakeTable.java:198)
	at org.apache.fluss.server.zk.data.lake.LakeTable.getOrReadLatestTableSnapshot(LakeTable.java:142)
	at org.apache.fluss.server.zk.ZooKeeperClient.getLakeTableSnapshot(ZooKeeperClient.java:1383)
	at org.apache.fluss.server.RpcServiceBase.lambda$getLakeSnapshot$2(RpcServiceBase.java:513)
	... 3 more

	at java.base/java.util.concurrent.CompletableFuture.reportGet(Unknown Source) ~[?:?]
	at java.base/java.util.concurrent.CompletableFuture.get(Unknown Source) ~[?:?]
	at org.apache.fluss.flink.tiering.source.split.TieringSplitGenerator.generateTableSplits(TieringSplitGenerator.java:67) ~[fluss-flink-2.2-1.0-SNAPSHOT.jar:1.0-SNAPSHOT]
	... 11 more
Caused by: org.apache.fluss.exception.UnknownServerException: org.apache.fluss.exception.FlussRuntimeException: Failed to get lake table snapshot for table: cdstream_db.timeseries, table id: 62
	at org.apache.fluss.server.RpcServiceBase.lambda$getLakeSnapshot$2(RpcServiceBase.java:541)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(Unknown Source)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(Unknown Source)
	at java.base/java.lang.Thread.run(Unknown Source)
Caused by: java.io.FileNotFoundException: No such file or directory: s3://fluss/remote-data/lake/cdstream_db/timeseries-58/metadata/cf19d6cc-4a98-4ce0-b518-67dd182fbbe6.offsets
	at org.apache.hadoop.fs.s3a.S3AFileSystem.s3GetFileStatus(S3AFileSystem.java:3866)
	at org.apache.hadoop.fs.s3a.S3AFileSystem.innerGetFileStatus(S3AFileSystem.java:3688)
	at org.apache.hadoop.fs.s3a.S3AFileSystem.extractOrFetchSimpleFileStatus(S3AFileSystem.java:5401)
	at org.apache.hadoop.fs.s3a.S3AFileSystem.open(S3AFileSystem.java:1465)
	at org.apache.hadoop.fs.s3a.S3AFileSystem.open(S3AFileSystem.java:1441)
	at org.apache.hadoop.fs.FileSystem.open(FileSystem.java:976)
	at org.apache.fluss.fs.hdfs.HadoopFileSystem.open(HadoopFileSystem.java:72)
	at org.apache.fluss.fs.hdfs.HadoopFileSystem.open(HadoopFileSystem.java:39)
	at org.apache.fluss.fs.PluginFileSystemWrapper$ClassLoaderFixingFileSystem.open(PluginFileSystemWrapper.java:95)
	at org.apache.fluss.server.zk.data.lake.LakeTable.toLakeTableSnapshot(LakeTable.java:203)
	at org.apache.fluss.server.zk.data.lake.LakeTable.toLakeTableSnapshot(LakeTable.java:198)
	at org.apache.fluss.server.zk.data.lake.LakeTable.getOrReadLatestTableSnapshot(LakeTable.java:142)
	at org.apache.fluss.server.zk.ZooKeeperClient.getLakeTableSnapshot(ZooKeeperClient.java:1383)
	at org.apache.fluss.server.RpcServiceBase.lambda$getLakeSnapshot$2(RpcServiceBase.java:513)
	... 3 more
```

## How to reproduce

1. Start Fluss with `datalake.format=paimon`, `remote.data.dir=s3://…` and a primary-key table
   `db.t` with `table.datalake.enabled=true`.
2. Write enough data to trigger at least one successful tier-snapshot for `db.t`. Confirm
   `s3://…/lake/db/t-<id1>/metadata/<uuid>.offsets` exists and
   `/fluss/tabletservers/tables/<id1>/laketable` contains a `lake_snapshots` entry pointing at it.
3. `dropTable db.t` and **immediately** `createTable db.t` again. (In our case this is the
   default behaviour of a data-load utility that calls `dropTable + createTable` per run.)
4. Observe the post-drop ZK state:
   - `/fluss/tabletservers/tables/<id1>` still exists with `[laketable]` as its only child
     (the rest of the znode tree was cleaned, but the `/laketable` child was re-created by a
     late `commit`).
   - `/fluss/tabletservers/tables/<id2>/laketable` contains `tiered_offsets` referencing
     `lake/db/t-<id1>/metadata/<uuid>.offsets` (an old-tableId path).
5. Observe the `TieringSource` enumerator:
   ```
   WARN  TieringSourceEnumerator - Fail to generate Tiering splits for table db.t.
   Caused by: FlussRuntimeException: Failed to get lake table snapshot for table: db.t,
       table id: <id2>
   Caused by: FileNotFoundException:
       s3://…/lake/db/t-<id1>/metadata/<uuid>.offsets
   ```
   `currentFailedTableEpochs: {<id2>=N}` grows unbounded.

## Observations

ZK state in a running cluster after dev-loop drop+recreate cycles:

```text
/fluss/tabletservers/tables                       [58, 62, 63, 64, 65]
/fluss/tabletservers/tables/58                    [laketable]    ← orphan
/fluss/tabletservers/tables/62                    [buckets, laketable]
/fluss/tabletservers/tables/62/laketable          (see below)
```

`/fluss/tabletservers/tables/62/laketable` content:

```json
{
  "version": 2,
  "lake_snapshots": [
    {
      "snapshot_id": 1,
      "tiered_offsets": "s3://fluss/remote-data/lake/cdstream_db/timeseries-58/metadata/cf19d6cc-4a98-4ce0-b518-67dd182fbbe6.offsets"
    }
  ]
}
```

S3:
- `lake/cdstream_db/timeseries-58/` directory key remains; the offsets file is gone.
- `lake/cdstream_db/timeseries-62/` does not exist.
- Sibling tables `timeseries_norm-63/`, `correlations-64/` are healthy (file present, ids match).

`timeseries_norm` and `correlations` are uninvolved because their previous incarnations
didn't complete a tier-snapshot before being dropped; only the table whose previous incarnation
had a successful tier-snapshot is caught in the race.

## Root cause analysis

Drop flow (`CoordinatorEventProcessor.processDropTable`,
`fluss-server/src/main/java/org/apache/fluss/server/coordinator/CoordinatorEventProcessor.java:861`):

```java
private void processDropTable(DropTableEvent dropTableEvent) {
    ...
    coordinatorContext.queueTableDeletion(...);
    tableManager.onDeleteTable(tableId);
    if (dropTableEvent.isAutoPartitionTable()) { ... }
    if (dropTableEvent.isDataLakeEnabled()) {
        lakeTableTieringManager.removeLakeTable(tableId);     // ← in-memory only
    }
    ...
}
```

`LakeTableTieringManager.removeLakeTable`
(`fluss-server/src/main/java/org/apache/fluss/server/coordinator/LakeTableTieringManager.java:324`)
is purely in-memory:

```java
public void removeLakeTable(long tableId) {
    inWriteLock(lock, () -> {
        tablePaths.remove(tableId);
        tableLakeFreshness.remove(tableId);
        lastTieringResult.remove(tableId);
        ...
    });
}
```

The actual ZK cleanup of `/fluss/tabletservers/tables/<id>` (which includes any `/laketable`
child) is `deleteTableAssignment(tableId)`, called later from `completeDeleteTable`
(`fluss-server/src/main/java/org/apache/fluss/server/coordinator/MetadataManager.java:336+`).

Meanwhile, `commitLakeTableSnapshot`
(`fluss-server/src/main/java/org/apache/fluss/server/coordinator/CoordinatorEventProcessor.java:2033`)
calls `LakeTableHelper.registerLakeTableSnapshotV2(tableId, lakeSnapshotMetadata, ...)`:

```java
zkClient.upsertLakeTable(tableId, lakeTable, optPreviousTable.isPresent());
```

This unconditionally upserts. There is no check that:
- the tableId is still present in `coordinatorContext`,
- the table hasn't been queued for deletion,
- the `lakeSnapshotMetadata.tieredOffsetsFilePath`'s `-<id>` segment matches the commit's
  `tableId` argument.

So any commit landing in the window between `processDropTable` and the eventual recursive
delete of `/fluss/tabletservers/tables/<oldId>` will:
- recreate the `/laketable` child under the old id (since `upsertLakeTable` will create parent
  znodes as needed for `LakeTableZNode.path(tableId)` writes), or
- write the stale path into the new id's `/laketable` if the commit's `tableId` is the new id
  but the metadata it carries was prepared under the old id.

## Suggested fixes

Any one of these would close the race, ideally all three:

1. **Reject commits for non-existent tableIds.**
   In `CoordinatorEventProcessor.processCommitLakeTableSnapshotEvent`, before calling
   `registerLakeTableSnapshotV2`, look up `coordinatorContext.getTableInfoById(tableId)` and
   refuse the commit (return an `ApiError`) if the table has been dropped / queued for deletion.

2. **Validate path-tableId consistency.**
   In `LakeTableHelper.registerLakeTableSnapshotV2` (or earlier at the RPC boundary), parse the
   incoming `tieredOffsetsFilePath` and assert that the directory segment matches
   `FlussPaths.remoteLakeTableSnapshotOffsetPath(remoteDataDir, tablePath, tableId)`'s
   expected layout. Reject mismatches with a clear error.

3. **Wipe the lake znode synchronously in `processDropTable`.**
   In `processDropTable`, before issuing the asynchronous tablet-bucket cleanup, delete
   `LakeTableZNode.path(tableId)` synchronously (and the parent
   `/fluss/tabletservers/tables/<id>` if safe to do at that point). That removes the resurrection
   target and makes any late commit fail cleanly.

(1) alone is enough to close the user-visible failure mode. (2) defends in depth against any
client that constructs a malformed commit. (3) is the cleanest fix but interacts with the
existing two-phase delete pipeline.

## Workaround for affected clusters

Stop the tiering job, then via `zkCli.sh`:

```
delete /fluss/tabletservers/tables/<newId>/laketable
delete /fluss/tabletservers/tables/<oldId>/laketable
delete /fluss/tabletservers/tables/<oldId>
```

Restart the tiering job. It will re-prepare a fresh offsets file for the new tableId.

## Relevant source pointers

- `fluss-server/src/main/java/org/apache/fluss/server/coordinator/CoordinatorEventProcessor.java:861-889` — `processDropTable`
- `fluss-server/src/main/java/org/apache/fluss/server/coordinator/CoordinatorEventProcessor.java:2000-2052` — `commitLakeTableSnapshot` event handler
- `fluss-server/src/main/java/org/apache/fluss/server/coordinator/CoordinatorService.java:830-900` — `prepareLakeTable` / `commitLakeTableSnapshot` RPCs
- `fluss-server/src/main/java/org/apache/fluss/server/coordinator/MetadataManager.java:336-360` — `dropTable` / `completeDeleteTable`
- `fluss-server/src/main/java/org/apache/fluss/server/coordinator/LakeTableTieringManager.java:324-345` — `removeLakeTable`
- `fluss-server/src/main/java/org/apache/fluss/server/zk/data/lake/LakeTableHelper.java:92-125` — `registerLakeTableSnapshotV2`
- `fluss-server/src/main/java/org/apache/fluss/server/zk/data/lake/LakeTableHelper.java:183-203` — `storeLakeTableOffsetsFile`
- `fluss-server/src/main/java/org/apache/fluss/server/zk/ZooKeeperClient.java:489-493` — `deleteTableAssignment`
- `fluss-server/src/main/java/org/apache/fluss/server/zk/data/ZkData.java:438-444` — `TableIdsZNode.path()`
- `fluss-server/src/main/java/org/apache/fluss/server/zk/data/ZkData.java:665-678` — `LakeTableZNode`

### Solution

_No response_

Jackeyzhe:
### Purpose

Linked issue: close #3483 

This PR fixes a race where a stale V2 lake snapshot commit can be accepted after a table is dropped and recreated. In that case, the recreated table may get a `/laketable` znode pointing to an offsets metadata file generated for the old table id, causing lake tiering to repeatedly fail when reading the stale offsets path.

### Brief change log
- Reject V2 lake snapshot commits when the target table id no longer exists or is queued for deletion.
- Validate the V2 lake snapshot offsets metadata before writing it to ZK:
  - the offsets file must be readable;
  - the file-level `tableId` must match the commit table id;
  - every `TableBucket` in the offsets file must also belong to the commit table id.
- Validate both tiered offsets and readable offsets metadata paths when they differ.
- Add regression coverage for:
  - non-existent table id commits;
  - table queued for deletion;
  - recreated table id with stale offsets metadata from the old table id;
  - RPC-level V2 commit after table drop.

### Tests
``` bash
mvn -pl fluss-server -DskipITs -Dcheckstyle.skip -Drat.skip -Dspotless.check.skip=true -Dtest=CoordinatorEventProcessorTest#testCommitLakeTableSnapshotV2RejectsMismatchedOffsetsFileTableId+testCommitLakeTableSnapshotV2RejectsNonExistentTable+testCommitLakeTableSnapshotV2RejectsQueuedForDeletionTable test
```

``` bash
mvn -pl fluss-server -DskipITs -Dcheckstyle.skip -Drat.skip -Dspotless.check.skip=true -Dtest=CommitLakeTableSnapshotITCase#testCommitLakeTableSnapshotV2RejectedAfterDropTable test
```

### API and Format
no

### Documentation
no

### DIFF
--- a/fluss-server/src/main/java/org/apache/fluss/server/coordinator/CoordinatorEventProcessor.java
+++ b/fluss-server/src/main/java/org/apache/fluss/server/coordinator/CoordinatorEventProcessor.java
@@ -2278,6 +2278,20 @@ public class CoordinatorEventProcessor implements EventProcessor {
             CompletableFuture<CommitLakeTableSnapshotResponse> callback) {
         CommitLakeTableSnapshotsData commitLakeTableSnapshotsData =
                 commitLakeTableSnapshotEvent.getCommitLakeTableSnapshotsData();
+        Map<Long, CommitLakeTableSnapshotsData.CommitLakeTableSnapshot>
+                commitLakeTableSnapshotByTableId =
+                        commitLakeTableSnapshotsData.getCommitLakeTableSnapshotByTableId();
+        Map<Long, TablePath> tablePathById = new HashMap<>();
+        Set<Long> toBeDeletedTableIds = new HashSet<>();
+        for (Long tableId : commitLakeTableSnapshotByTableId.keySet()) {
+            TablePath tablePath = coordinatorContext.getTablePathById(tableId);
+            if (tablePath != null) {
+                tablePathById.put(tableId, tablePath);
+            }
+            if (coordinatorContext.isTableQueuedForDeletion(tableId)) {
+                toBeDeletedTableIds.add(tableId);
+            }
+        }
         ioExecutor.execute(
                 () -> {
                     try {
@@ -2285,10 +2299,7 @@ public class CoordinatorEventProcessor implements EventProcessor {
                                 new CommitLakeTableSnapshotResponse();
                         Set<Long> failedTableIds = new HashSet<>();
                         for (Map.Entry<Long, CommitLakeTableSnapshotsData.CommitLakeTableSnapshot>
-                                entry :
-                                        commitLakeTableSnapshotsData
-                                                .getCommitLakeTableSnapshotByTableId()
-                                                .entrySet()) {
+                                entry : commitLakeTableSnapshotByTableId.entrySet()) {
                             PbCommitLakeTableSnapshotRespForTable tableResp =
                                     response.addTableResp();
                             long tableId = entry.getKey();
@@ -2300,6 +2311,7 @@ public class CoordinatorEventProcessor implements EventProcessor {
                                     throw new FlussRuntimeException(
                                             "Lake snapshot metadata is null for table " + tableId);
                                 }
+                                ensureTableNotDeleted(tablePathById, toBeDeletedTableIds, tableId);
                                 lakeTableHelper.registerLakeTableSnapshotV2(
                                         tableId,
                                         snapshot.getLakeSnapshotMetadata(),
@@ -2319,6 +2331,18 @@ public class CoordinatorEventProcessor implements EventProcessor {
                         callback.completeExceptionally(e);
                     }
                 });
+    }
+
+    private void ensureTableNotDeleted(
+            Map<Long, TablePath> tablePathById, Set<Long> toBeDeletedTableIds, long tableId) {
+        if (!tablePathById.containsKey(tableId)
+                || tablePathById.get(tableId) == null
+                || toBeDeletedTableIds.contains(tableId)) {
+            throw new TableNotExistException(
+                    "Table "
+                            + tableId
+                            + " not found or queued for deletion in coordinator context.");
+        }
     }
 
     private ControlledShutdownResponse tryProcessControlledShutdown(
--- a/fluss-server/src/main/java/org/apache/fluss/server/coordinator/CoordinatorEventProcessor.java
+++ b/fluss-server/src/main/java/org/apache/fluss/server/coordinator/CoordinatorEventProcessor.java
@@ -2281,15 +2281,11 @@ public class CoordinatorEventProcessor implements EventProcessor {
         Map<Long, CommitLakeTableSnapshotsData.CommitLakeTableSnapshot>
                 commitLakeTableSnapshotByTableId =
                         commitLakeTableSnapshotsData.getCommitLakeTableSnapshotByTableId();
-        Map<Long, TablePath> tablePathById = new HashMap<>();
-        Set<Long> toBeDeletedTableIds = new HashSet<>();
+        Set<Long> unavailableTableIds = new HashSet<>();
         for (Long tableId : commitLakeTableSnapshotByTableId.keySet()) {
-            TablePath tablePath = coordinatorContext.getTablePathById(tableId);
-            if (tablePath != null) {
-                tablePathById.put(tableId, tablePath);
-            }
-            if (coordinatorContext.isTableQueuedForDeletion(tableId)) {
-                toBeDeletedTableIds.add(tableId);
+            if (coordinatorContext.getTablePathById(tableId) == null
+                    || coordinatorContext.isTableQueuedForDeletion(tableId)) {
+                unavailableTableIds.add(tableId);
             }
         }
         ioExecutor.execute(
@@ -2311,7 +2307,12 @@ public class CoordinatorEventProcessor implements EventProcessor {
                                     throw new FlussRuntimeException(
                                             "Lake snapshot metadata is null for table " + tableId);
                                 }
-                                ensureTableNotDeleted(tablePathById, toBeDeletedTableIds, tableId);
+                                if (unavailableTableIds.contains(tableId)) {
+                                    throw new TableNotExistException(
+                                            "Table "
+                                                    + tableId
+                                                    + " not found or queued for deletion in coordinator context.");
+                                }
                                 lakeTableHelper.registerLakeTableSnapshotV2(
                                         tableId,
                                         snapshot.getLakeSnapshotMetadata(),
@@ -2331,18 +2332,6 @@ public class CoordinatorEventProcessor implements EventProcessor {
                         callback.completeExceptionally(e);
                     }
                 });
-    }
-
-    private void ensureTableNotDeleted(
-            Map<Long, TablePath> tablePathById, Set<Long> toBeDeletedTableIds, long tableId) {
-        if (!tablePathById.containsKey(tableId)
-                || tablePathById.get(tableId) == null
-                || toBeDeletedTableIds.contains(tableId)) {
-            throw new TableNotExistException(
-                    "Table "
-                            + tableId
-                            + " not found or queued for deletion in coordinator context.");
-        }
     }
 
     private ControlledShutdownResponse tryProcessControlledShutdown(
--- a/fluss-server/src/main/java/org/apache/fluss/server/coordinator/CoordinatorEventProcessor.java
+++ b/fluss-server/src/main/java/org/apache/fluss/server/coordinator/CoordinatorEventProcessor.java
@@ -2307,15 +2307,11 @@ public class CoordinatorEventProcessor implements EventProcessor {
         Map<Long, CommitLakeTableSnapshotsData.CommitLakeTableSnapshot>
                 commitLakeTableSnapshotByTableId =
                         commitLakeTableSnapshotsData.getCommitLakeTableSnapshotByTableId();
-        Map<Long, TablePath> tablePathById = new HashMap<>();
-        Set<Long> toBeDeletedTableIds = new HashSet<>();
+        Set<Long> unavailableTableIds = new HashSet<>();
         for (Long tableId : commitLakeTableSnapshotByTableId.keySet()) {
-            TablePath tablePath = coordinatorContext.getTablePathById(tableId);
-            if (tablePath != null) {
-                tablePathById.put(tableId, tablePath);
-            }
-            if (coordinatorContext.isTableQueuedForDeletion(tableId)) {
-                toBeDeletedTableIds.add(tableId);
+            if (coordinatorContext.getTablePathById(tableId) == null
+                    || coordinatorContext.isTableQueuedForDeletion(tableId)) {
+                unavailableTableIds.add(tableId);
             }
         }
         ioExecutor.execute(
@@ -2337,7 +2333,12 @@ public class CoordinatorEventProcessor implements EventProcessor {
                                     throw new FlussRuntimeException(
                                             "Lake snapshot metadata is null for table " + tableId);
                                 }
-                                ensureTableNotDeleted(tablePathById, toBeDeletedTableIds, tableId);
+                                if (unavailableTableIds.contains(tableId)) {
+                                    throw new TableNotExistException(
+                                            "Table "
+                                                    + tableId
+                                                    + " not found or queued for deletion in coordinator context.");
+                                }
                                 lakeTableHelper.registerLakeTableSnapshotV2(
                                         tableId,
                                         snapshot.getLakeSnapshotMetadata(),
@@ -2357,18 +2358,6 @@ public class CoordinatorEventProcessor implements EventProcessor {
                         callback.completeExceptionally(e);
                     }
                 });
-    }
-
-    private void ensureTableNotDeleted(
-            Map<Long, TablePath> tablePathById, Set<Long> toBeDeletedTableIds, long tableId) {
-        if (!tablePathById.containsKey(tableId)
-                || tablePathById.get(tableId) == null
-                || toBeDeletedTableIds.contains(tableId)) {
-            throw new TableNotExistException(
-                    "Table "
-                            + tableId
-                            + " not found or queued for deletion in coordinator context.");
-        }
     }
 
     private ControlledShutdownResponse tryProcessControlledShutdown(
--- a/fluss-server/src/main/java/org/apache/fluss/server/coordinator/CoordinatorEventProcessor.java
+++ b/fluss-server/src/main/java/org/apache/fluss/server/coordinator/CoordinatorEventProcessor.java
@@ -2304,6 +2304,20 @@ public class CoordinatorEventProcessor implements EventProcessor {
             CompletableFuture<CommitLakeTableSnapshotResponse> callback) {
         CommitLakeTableSnapshotsData commitLakeTableSnapshotsData =
                 commitLakeTableSnapshotEvent.getCommitLakeTableSnapshotsData();
+        Map<Long, CommitLakeTableSnapshotsData.CommitLakeTableSnapshot>
+                commitLakeTableSnapshotByTableId =
+                        commitLakeTableSnapshotsData.getCommitLakeTableSnapshotByTableId();
+        Map<Long, TablePath> tablePathById = new HashMap<>();
+        Set<Long> toBeDeletedTableIds = new HashSet<>();
+        for (Long tableId : commitLakeTableSnapshotByTableId.keySet()) {
+            TablePath tablePath = coordinatorContext.getTablePathById(tableId);
+            if (tablePath != null) {
+                tablePathById.put(tableId, tablePath);
+            }
+            if (coordinatorContext.isTableQueuedForDeletion(tableId)) {
+                toBeDeletedTableIds.add(tableId);
+            }
+        }
         ioExecutor.execute(
                 () -> {
                     try {
@@ -2311,10 +2325,7 @@ public class CoordinatorEventProcessor implements EventProcessor {
                                 new CommitLakeTableSnapshotResponse();
                         Set<Long> failedTableIds = new HashSet<>();
                         for (Map.Entry<Long, CommitLakeTableSnapshotsData.CommitLakeTableSnapshot>
-                                entry :
-                                        commitLakeTableSnapshotsData
-                                                .getCommitLakeTableSnapshotByTableId()
-                                                .entrySet()) {
+                                entry : commitLakeTableSnapshotByTableId.entrySet()) {
                             PbCommitLakeTableSnapshotRespForTable tableResp =
                                     response.addTableResp();
                             long tableId = entry.getKey();
@@ -2326,6 +2337,7 @@ public class CoordinatorEventProcessor implements EventProcessor {
                                     throw new FlussRuntimeException(
                                             "Lake snapshot metadata is null for table " + tableId);
                                 }
+                                ensureTableNotDeleted(tablePathById, toBeDeletedTableIds, tableId);
                                 lakeTableHelper.registerLakeTableSnapshotV2(
                                         tableId,
                                         snapshot.getLakeSnapshotMetadata(),
@@ -2345,6 +2357,18 @@ public class CoordinatorEventProcessor implements EventProcessor {
                         callback.completeExceptionally(e);
                     }
                 });
+    }
+
+    private void ensureTableNotDeleted(
+            Map<Long, TablePath> tablePathById, Set<Long> toBeDeletedTableIds, long tableId) {
+        if (!tablePathById.containsKey(tableId)
+                || tablePathById.get(tableId) == null
+                || toBeDeletedTableIds.contains(tableId)) {
+            throw new TableNotExistException(
+                    "Table "
+                            + tableId
+                            + " not found or queued for deletion in coordinator context.");
+        }
     }
 
     private ControlledShutdownResponse tryProcessControlledShutdown(
--- a/fluss-server/src/main/java/org/apache/fluss/server/zk/ZooKeeperClient.java
+++ b/fluss-server/src/main/java/org/apache/fluss/server/zk/ZooKeeperClient.java
@@ -1415,7 +1415,7 @@ public class ZooKeeperClient implements AutoCloseable {
         if (isUpdate) {
             zkClient.setData().forPath(zkPath, zkData);
         } else {
-            zkClient.create().forPath(zkPath, zkData);
+            zkClient.create().creatingParentsIfNeeded().forPath(zkPath, zkData);
         }
     }
 
--- a/fluss-server/src/main/java/org/apache/fluss/server/zk/ZooKeeperClient.java
+++ b/fluss-server/src/main/java/org/apache/fluss/server/zk/ZooKeeperClient.java
@@ -1415,7 +1415,7 @@ public class ZooKeeperClient implements AutoCloseable {
         if (isUpdate) {
             zkClient.setData().forPath(zkPath, zkData);
         } else {
-            zkClient.create().creatingParentsIfNeeded().forPath(zkPath, zkData);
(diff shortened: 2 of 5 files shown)


fluss-server/src/test/java/org/apache/fluss/server/coordinator/CoordinatorEventProcessorTest.java
```diff
@@ -1585,6 +1565,108 @@ void testLeaderOnlyRebalanceExecutesSequentially() throws Exception {
         verifyIsr(tb2, 1, Arrays.asList(0, 1, 2));
     }
 
+    @Test
+    void testCommitLakeTableSnapshotV2RejectsNonExistentTable() throws Exception {
```
luoyuxia: nit: I think we can remove the ut since testCommitLakeTableSnapshotV2RejectedAfterDropTable alreay cover it. I want to keep this test class short if possible

fluss-server/src/main/java/org/apache/fluss/server/coordinator/CoordinatorEventProcessor.java
```diff
@@ -2145,6 +2135,16 @@ private void handleCommitLakeTableSnapshotV2(
                 });
     }
 
+    private void ensureTableCanAcceptLakeSnapshot(long tableId) {
```
luoyuxia: nit: rename to ensureTableNotDeleted? I can image maybe other event maywell need this method.

luoyuxia (commented):
@Jackeyzhe Thanks for the pr. LGTM overall. Left minor comments. PTAL

luoyuxia:
@Jackeyzhe Thanks for working on this.

I wonder if we can make this fix simpler and close the race at the ZK write boundary instead. The current approach adds coordinator-side lifecycle checks before submitting the actual ZK write to `ioExecutor`, but that check can become stale before `registerLakeTableSnapshotV2` runs. For example, the commit event may pass the context check while the table still exists, then `dropTable` can delete `/tabletservers/tables/<tableId>`, and the later IO task can still recreate `/tabletservers/tables/<tableId>/laketable` because `upsertLakeTable` uses `creatingParentsIfNeeded()`.

Would it be enough to change the create branch in `ZooKeeperClient.upsertLakeTable` to avoid recreating the table parent?

```java
// Do not recreate /tabletservers/tables/<tableId> for a late commit.
zkClient.create().forPath(zkPath, zkData);
```

Then a late commit after `deleteTableAssignment(tableId)` would fail with `NoNodeException` instead of resurrecting the old tableId znode. We can convert that to `TableNotExistException` for the commit response. The coordinator-side queued/deleted check can still be kept as an early rejection, but the ZK write itself should be what prevents the resurrection race.

Jackeyzhe:
> @Jackeyzhe Thanks for working on this.
> 
> I wonder if we can make this fix simpler and close the race at the ZK write boundary instead. The current approach adds coordinator-side lifecycle checks before submitting the actual ZK write to `ioExecutor`, but that check can become stale before `registerLakeTableSnapshotV2` runs. For example, the commit event may pass the context check while the table still exists, then `dropTable` can delete `/tabletservers/tables/<tableId>`, and the later IO task can still recreate `/tabletservers/tables/<tableId>/laketable` because `upsertLakeTable` uses `creatingParentsIfNeeded()`.
> 
> Would it be enough to change the create branch in `ZooKeeperClient.upsertLakeTable` to avoid recreating the table parent?
> 
> ```java
> // Do not recreate /tabletservers/tables/<tableId> for a late commit.
> zkClient.create().forPath(zkPath, zkData);
> ```
> 
> Then a late commit after `deleteTableAssignment(tableId)` would fail with `NoNodeException` instead of resurrecting the old tableId znode. We can convert that to `TableNotExistException` for the commit response. The coordinator-side queued/deleted check can still be kept as an early rejection, but the ZK write itself should be what prevents the resurrection race.

Thanks for pointing this out. I agree that the coordinator-side check is only an early rejection and does not fully close the race before the actual ZK write.

I'll update the fix to make the ZK write boundary authoritative: for the create branch in ZooKeeperClient.upsertLakeTable, avoid creating missing parents and use create().forPath(...) directly. Then a late commit after the table assignment znode has been deleted will fail with NoNodeException instead of recreating /tabletservers/tables/<tableId>. I'll convert that failure to TableNotExistException for the commit response.

I'll keep the coordinator-side queued/deleted check as an early rejection, but rely on the ZK create semantics to prevent resurrection.

luoyuxia:
> > @Jackeyzhe Thanks for working on this.
> > I wonder if we can make this fix simpler and close the race at the ZK write boundary instead. The current approach adds coordinator-side lifecycle checks before submitting the actual ZK write to `ioExecutor`, but that check can become stale before `registerLakeTableSnapshotV2` runs. For example, the commit event may pass the context check while the table still exists, then `dropTable` can delete `/tabletservers/tables/<tableId>`, and the later IO task can still recreate `/tabletservers/tables/<tableId>/laketable` because `upsertLakeTable` uses `creatingParentsIfNeeded()`.
> > Would it be enough to change the create branch in `ZooKeeperClient.upsertLakeTable` to avoid recreating the table parent?
> > ```java
> > // Do not recreate /tabletservers/tables/<tableId> for a late commit.
> > zkClient.create().forPath(zkPath, zkData);
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Then a late commit after `deleteTableAssignment(tableId)` would fail with `NoNodeException` instead of resurrecting the old tableId znode. We can convert that to `TableNotExistException` for the commit response. The coordinator-side queued/deleted check can still be kept as an early rejection, but the ZK write itself should be what prevents the resurrection race.
> 
> Thanks for pointing this out. I agree that the coordinator-side check is only an early rejection and does not fully close the race before the actual ZK write.
> 
> I'll update the fix to make the ZK write boundary authoritative: for the create branch in ZooKeeperClient.upsertLakeTable, avoid creating missing parents and use create().forPath(...) directly. Then a late commit after the table assignment znode has been deleted will fail with NoNodeException instead of recreating /tabletservers/tables/. I'll convert that failure to TableNotExistException for the commit response.
> 
> I'll keep the coordinator-side queued/deleted check as an early rejection, but rely on the ZK create semantics to prevent resurrection.

Can we remove the coordinator-side queued/deleted check to keep code simple?

Jackeyzhe:
> > > @Jackeyzhe Thanks for working on this.
> > > I wonder if we can make this fix simpler and close the race at the ZK write boundary instead. The current approach adds coordinator-side lifecycle checks before submitting the actual ZK write to `ioExecutor`, but that check can become stale before `registerLakeTableSnapshotV2` runs. For example, the commit event may pass the context check while the table still exists, then `dropTable` can delete `/tabletservers/tables/<tableId>`, and the later IO task can still recreate `/tabletservers/tables/<tableId>/laketable` because `upsertLakeTable` uses `creatingParentsIfNeeded()`.
> > > Would it be enough to change the create branch in `ZooKeeperClient.upsertLakeTable` to avoid recreating the table parent?
> > > ```java
> > > // Do not recreate /tabletservers/tables/<tableId> for a late commit.
> > > zkClient.create().forPath(zkPath, zkData);
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > Then a late commit after `deleteTableAssignment(tableId)` would fail with `NoNodeException` instead of resurrecting the old tableId znode. We can convert that to `TableNotExistException` for the commit response. The coordinator-side queued/deleted check can still be kept as an early rejection, but the ZK write itself should be what prevents the resurrection race.
> > 
> > 
> > Thanks for pointing this out. I agree that the coordinator-side check is only an early rejection and does not fully close the race before the actual ZK write.
> > I'll update the fix to make the ZK write boundary authoritative: for the create branch in ZooKeeperClient.upsertLakeTable, avoid creating missing parents and use create().forPath(...) directly. Then a late commit after the table assignment znode has been deleted will fail with NoNodeException instead of recreating /tabletservers/tables/. I'll convert that failure to TableNotExistException for the commit response.
> > I'll keep the coordinator-side queued/deleted check as an early rejection, but rely on the ZK create semantics to prevent resurrection.
> 
> Can we remove the coordinator-side queued/deleted check to keep code simple?

sure, I'll remove it

Jackeyzhe:
Hi @luoyuxia,

The new CI run (28710725908) surfaced another NoNodeException failure (`/tabletservers/tables/<id>/laketable`) for partitioned tables — traced to `MetadataManager.createTable` skipping `registerTableAssignment` when `tableAssignment == null`.

On second thought, I'm going to put the coordinator-side pre-check back in. Without
it, this PR can't ship without bundling a MetadataManager change (And I'm not sure if this is reasonable). 
With it, this PR stays focused on #3483 and the partitioned-table parent-znode issue can be tracked separately. I'll restore `ensureTableNotDeleted` in `handleCommitLakeTableSnapshotV2`, keep the new V2 catch in `LakeTableHelper.registerLakeTableSnapshotV2` (so the ZK boundary still rejects late commits), and re-run CI.

Thanks!

luoyuxia:
> Hi @luoyuxia,
> 
> The new CI run (28710725908) surfaced another NoNodeException failure (`/tabletservers/tables/<id>/laketable`) for partitioned tables — traced to `MetadataManager.createTable` skipping `registerTableAssignment` when `tableAssignment == null`.
> 
> On second thought, I'm going to put the coordinator-side pre-check back in. Without it, this PR can't ship without bundling a MetadataManager change (And I'm not sure if this is reasonable). With it, this PR stays focused on #3483 and the partitioned-table parent-znode issue can be tracked separately. I'll restore `ensureTableNotDeleted` in `handleCommitLakeTableSnapshotV2`, keep the new V2 catch in `LakeTableHelper.registerLakeTableSnapshotV2` (so the ZK boundary still rejects late commits), and re-run CI.
> 
> Thanks!

Thanks for your patient. Will have another review

luoyuxia:
@Jackeyzhe I simplified the design a bit after rebasing the branch onto the latest main:

- collect unavailable table IDs directly, instead of keeping a table-path map and a separate queued-for-deletion set;
- inline the one-use validation and remove the helper;
- remove the two coordinator unit tests, while keeping the RPC-level drop-table integration test.

The retained CommitLakeTableSnapshotITCase#testCommitLakeTableSnapshotV2RejectedAfterDropTable passes locally.

Jackeyzhe:
Hi @luoyuxia , thanks for the simplification and the rebase!

I've reviewed the changes and they look good:

1. **Inline check** — collecting `unavailableTableIds` upfront and checking inline in the IO loop is cleaner than the separate helper. Agree with this approach.
2. **IT test only** — `testCommitLakeTableSnapshotV2RejectedAfterDropTable` covers the end-to-end RPC → coordinator → ZK flow well. Dropping the unit tests in favor of the integration test makes sense here.
3. One thing I noticed: `upsertLakeTable` still uses `creatingParentsIfNeeded()`, so the ZK-level defense against late-commit znode resurrection we discussed earlier is no longer in place. The coordinator-side check covers the common case (table already dropped before the commit event is processed), but the narrow window between collecting `unavailableTableIds` and the IO task execution remains. Should we track that as a follow-up issue, or is this considered acceptable for now given the partitioned-table complexity the ZK boundary fix introduced?

CI is all green. Happy to help if there's anything else needed before this can move forward.

Issue #3490, opened 2026-06-16
[lake] Union read on partitioned table returns duplicated rows when partition column is non-String type


bryndenZh:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/fluss/issues) and found nothing similar.

### Fluss version

main

### Please describe the bug 🐞

For a partitioned table with `table.datalake.enabled = true` and a **non-`STRING` partition column** (`DATE`, `INT`, `TIMESTAMP`, etc.), after data is tiered to the lake the default union read returns each row twice (once from the lake, once from the Fluss log), so the row count doubles. Affects both log tables and primary-key tables.

#### Reproduce

```sql
CREATE TABLE log_part_date (id BIGINT, dt DATE) PARTITIONED BY (dt)
WITH ('table.datalake.enabled' = 'true', 'table.datalake.freshness' = '10s');
INSERT INTO log_part_date VALUES (1, DATE '2024-03-01');
-- after tiering commits a snapshot, in batch mode:
SELECT COUNT(*) FROM log_part_date;        -- expected 1, actual 2
SELECT COUNT(*) FROM log_part_date$lake;   -- 1
```

#### Root cause

[`PaimonSplit.partition()`](https://github.com/apache/fluss/blob/main/fluss-lake/fluss-lake-paimon/src/main/java/org/apache/fluss/lake/paimon/source/PaimonSplit.java#L54-L67) reads every partition field via `BinaryRow.getString(i)` regardless of logical type, producing garbage for non-`STRING` columns. The lake-side partition name then never matches the Fluss-side name in `LakeSplitGenerator.generatePartitionTableSplit`, so the same partition is emitted as both a lake split and a Fluss-log split.

### Are you willing to submit a PR?

- [x] I'm willing to submit a PR!

luoyuxia:
@bryndenZh Thanks for raising it.  Assigned to you

PR #3500, opened 2026-06-19
[lake] Fix union read duplicating rows for non-String partition columns

bryndenZh:
## Purpose

Related issue: #3490

For a partitioned table with `table.datalake.enabled=true` and a non-`STRING`
partition column (`DATE`, `INT`, `TIMESTAMP`, ...), union read returns each tiered
row twice.

`PaimonSplit#partition()` read every field via `BinaryRow.getString` regardless of
type, so for non-`STRING` columns the lake-side partition name was garbage and never
matched the Fluss-side name in `LakeSplitGenerator` — the partition was then emitted
as both a lake split and a Fluss-log split.

## Brief change log

- Add `PaimonPartitionUtils#partitionValues` to render partition values type-aware,
  matching the Fluss-side format (`PartitionUtils#convertValueOfType`).

## Tests

- `PaimonPartitionUtilsTest`: every partition type's output equals
  `PartitionUtils#convertValueOfType` for the same value; plus string and
  multi-column cases.
- `PaimonSplitTest#testPaimonSplitWithDatePartition`: a DATE-partitioned table
  yields `["2024-03-01"]` end-to-end through the planner (garbage before).
- `PaimonSplitSerializerTest`: partition values round-trip through serialization.

## API and Format

`PaimonSplit`'s serialized form now includes the partition values (appended after
the existing fields).

## Documentation

No user-facing documentation change.


luoyuxia:
@bryndenZh Thanks for the pr.

Nice catch on the root cause. One thing worth tightening before merge:

**`PaimonPartitionUtils#toFlussPartitionString` is a second copy of `PartitionUtils#convertValueOfType`.** The two `switch` statements now have to stay in lockstep forever — if a new partition type is added on the Fluss side and only `convertValueOfType` is updated, the lake-side and Fluss-side names silently diverge again, which is exactly the class of bug this PR fixes. The test that asserts "the two produce the same string" is really a symptom of maintaining a mirror.

We can delegate formatting to the single authoritative `convertValueOfType`, and reuse the existing `PaimonRowAsFlussRow` adapter + Fluss's `InternalRow.createFieldGetter`, so there's no per-type logic here at all:

```java
public static List<String> toFlussPartitionValues(
        BinaryRow partition, org.apache.fluss.types.RowType flussPartitionType) {
    PaimonRowAsFlussRow flussRow = new PaimonRowAsFlussRow().replaceRow(partition);
    List<String> values = new ArrayList<>(partition.getFieldCount());
    for (int i = 0; i < partition.getFieldCount(); i++) {
        DataType flussType = flussPartitionType.getTypeAt(i);
        Object value = InternalRow.createFieldGetter(flussType, i).getFieldOrNull(flussRow);
        values.add(PartitionUtils.convertValueOfType(value, flussType.getTypeRoot()));
    }
    return values;
}
```

This needs a Paimon→Fluss type converter, which doesn't exist yet (we only have the reverse `FlussDataTypeToPaimonDataType`). Adding the symmetric `PaimonDataTypeToFlussDataType` as a `DataTypeDefaultVisitor` is small, mechanical, and reusable across the whole `fluss-lake-paimon` module. Watch three mappings: Paimon `VARCHAR → STRING`, `VARBINARY → BYTES`, `LocalZonedTimestamp → TIMESTAMP_LTZ`.

A few smaller points:

- **Put these in `PaimonConversions`** (the module's single Paimon↔Fluss entry point) as `toFlussRowType` / `toFlussPartitionValues`, instead of a new `PaimonPartitionUtils` class.
- **`DvTableReadableSnapshotRetriever#getPartitionNameFromBinaryRow` has the same `getString`-only bug** — point it at the same helper so both paths are fixed together.
- **Perf: hoist the type resolution out of the loop.** The Paimon→Fluss conversion is identical for every split / partition bucket. Resolve the Fluss row type once and reuse it per split:
  ```java
  RowType flussPartitionType = toFlussRowType(fileStoreTable.schema().logicalPartitionType());
  for (Split split : scan.plan().splits()) { ... toFlussPartitionValues(dataSplit.partition(), flussPartitionType) ... }
  ```
- **(Separate issue) `PaimonSplitSerializer` appends the partition fields but keeps `VERSION_1`.** An old `VERSION_1` checkpoint would hit EOF on restore. Consider bumping to `VERSION_2` with a back-compat read path.

luoyuxia:
@bryndenZh Hi, any progress on this pr?

bryndenZh:
> @bryndenZh Thanks for the pr.
> 
> Nice catch on the root cause. One thing worth tightening before merge:
> 
> **`PaimonPartitionUtils#toFlussPartitionString` is a second copy of `PartitionUtils#convertValueOfType`.** The two `switch` statements now have to stay in lockstep forever — if a new partition type is added on the Fluss side and only `convertValueOfType` is updated, the lake-side and Fluss-side names silently diverge again, which is exactly the class of bug this PR fixes. The test that asserts "the two produce the same string" is really a symptom of maintaining a mirror.
> 
> We can delegate formatting to the single authoritative `convertValueOfType`, and reuse the existing `PaimonRowAsFlussRow` adapter + Fluss's `InternalRow.createFieldGetter`, so there's no per-type logic here at all:
> 
> ```java
> public static List<String> toFlussPartitionValues(
>         BinaryRow partition, org.apache.fluss.types.RowType flussPartitionType) {
>     PaimonRowAsFlussRow flussRow = new PaimonRowAsFlussRow().replaceRow(partition);
>     List<String> values = new ArrayList<>(partition.getFieldCount());
>     for (int i = 0; i < partition.getFieldCount(); i++) {
>         DataType flussType = flussPartitionType.getTypeAt(i);
>         Object value = InternalRow.createFieldGetter(flussType, i).getFieldOrNull(flussRow);
>         values.add(PartitionUtils.convertValueOfType(value, flussType.getTypeRoot()));
>     }
>     return values;
> }
> ```
> 
> This needs a Paimon→Fluss type converter, which doesn't exist yet (we only have the reverse `FlussDataTypeToPaimonDataType`). Adding the symmetric `PaimonDataTypeToFlussDataType` as a `DataTypeDefaultVisitor` is small, mechanical, and reusable across the whole `fluss-lake-paimon` module. Watch three mappings: Paimon `VARCHAR → STRING`, `VARBINARY → BYTES`, `LocalZonedTimestamp → TIMESTAMP_LTZ`.
> 
> A few smaller points:
> 
> * **Put these in `PaimonConversions`** (the module's single Paimon↔Fluss entry point) as `toFlussRowType` / `toFlussPartitionValues`, instead of a new `PaimonPartitionUtils` class.
> * **`DvTableReadableSnapshotRetriever#getPartitionNameFromBinaryRow` has the same `getString`-only bug** — point it at the same helper so both paths are fixed together.
> * **Perf: hoist the type resolution out of the loop.** The Paimon→Fluss conversion is identical for every split / partition bucket. Resolve the Fluss row type once and reuse it per split:
>   ```java
>   RowType flussPartitionType = toFlussRowType(fileStoreTable.schema().logicalPartitionType());
>   for (Split split : scan.plan().splits()) { ... toFlussPartitionValues(dataSplit.partition(), flussPartitionType) ... }
>   ```
> * **(Separate issue) `PaimonSplitSerializer` appends the partition fields but keeps `VERSION_1`.** An old `VERSION_1` checkpoint would hit EOF on restore. Consider bumping to `VERSION_2` with a back-compat read path.

 Thanks for the suggestions. I’ve updated the PR accordingly. PTAL.

fluss-lake/fluss-lake-paimon/src/main/java/org/apache/fluss/lake/paimon/source/PaimonSplitSerializer.java
```diff
@@ -55,11 +65,20 @@ public PaimonSplit deserialize(int version, byte[] serialized) throws IOExceptio
...
+            DataInputStream dis = new DataInputStream(in);
+            boolean isBucketUnAware = dis.readBoolean();
             if (version == VERSION_1) {
-                DataInputStream dis = new DataInputStream(in);
-                boolean isBucketUnAware = dis.readBoolean();
-                return new PaimonSplit(dataSplit, isBucketUnAware);
+                // VERSION_1 payloads do not contain partition values. Keep them readable, but
+                // don't try to reconstruct partition names from DataSplit because non-string
+                // partition columns require type information that is not present in the payload.
+                return new PaimonSplit(dataSplit, isBucketUnAware, Collections.emptyList());
```
luoyuxia: The PR returns new PaimonSplit(dataSplit, isBucketUnAware, Collections.emptyList()) for old VERSION_1 payloads. However, old payloads still contain DataSplit.partition(). For the previously supported string partition case, the old PaimonSplit.partition() recovered the value via BinaryRow.getString(i). After this change, deserializing the same v1 bytes changes ["A"] into [].

This breaks the expected compatibility contract of SimpleVersionedSerializer. It is reasonable that non-string v1 payloads cannot be reconstructed type-safely, but old string partition splits should not lose their existing behavior.

Suggested fix: in the VERSION_1 branch, preserve the old logic by reading dataSplit.partition() with getString(i) when fields exist. Add a v1 compatibility test that serializes old-format bytes and verifies a string partition still round-trips.

luoyuxia (commented):
@bryndenZh Thanks for update. LGTM overall. Only one minor comment

luoyuxia:
@bryndenZh Hi, I append a small commit. I'm going to merge it. Feel free to leave any comment if you have.

bryndenZh:
@luoyuxia Thanks for the fix. Looks good to me.

Issue #3278, opened 2026-05-08
Hudi source sortedReader and PredicateConverter


fhan688:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/fluss/issues) and found nothing similar.

### Description

Hudi source sortedReader and PredicateConverter

### Willingness to contribute

- [x] I'm willing to submit a PR!

luoyuxia:
Close via #3501

PR #3501, opened 2026-06-20
[lake/hudi] Support predicate pushdown

fhan688:
### Purpose

  Linked issue: https://github.com/apache/fluss/issues/3278

  This PR adds predicate pushdown support for the Hudi lake source on top of the existing Hudi split planner and source reader work.

  The sorted reader part is intentionally not included because the current Hudi reader path does not guarantee that emitted records are ordered by primary key. Declaring `SortedRecordReader`
  without that guarantee could break batch union read correctness. Hudi primary-key batch union read can stay unsupported for now, consistent with the current Iceberg behavior.

### Brief change log

  - Add `FlussToHudiExpressionPredicateConverter` to convert Fluss predicates to Hudi `ExpressionPredicates`.
    - Supports `=`, `!=`, `<`, `<=`, `>`, `>=`, `IN`, `NOT IN`, `AND`, and `OR`.
    - Keeps unsupported predicates as remaining filters.
    - Handles Hudi metadata-column offset when mapping Fluss field indexes to Hudi schema fields.
    - Uses the actual Avro field position when creating Flink `FieldReferenceExpression`.
    - Builds binary `AND`/`OR` trees to match Hudi's predicate evaluation behavior.

  - Wire predicate pushdown into `HudiLakeSource`.
    - Converts accepted filters in `withFilters`.
    - Stores converted Hudi predicates and passes them to `HudiRecordReader`.
    - Avoids loading Hudi schema for empty filter lists.

  - Add unit tests for:
    - Hudi predicate conversion.
    - Hudi source empty-filter behavior.

### Tests

  - `mvn -pl fluss-lake/fluss-lake-hudi -Dcheckstyle.skip=true spotless:apply`
  - `mvn -pl fluss-lake/fluss-lake-hudi -am clean test -DskipITs -Dcheckstyle.skip=true -DfailIfNoTests=false`
  - `git diff --check`
  - 
### API and Format

### Documentation

  No documentation change. This is an internal Hudi lake source implementation improvement.


fluss-lake/fluss-lake-hudi/src/main/java/org/apache/fluss/lake/hudi/source/HudiSortedRecordReader.java
```diff
@@ -0,0 +1,232 @@
...
+import javax.annotation.Nullable;
+
+import java.io.IOException;
+import java.io.Serializable;
+import java.util.ArrayList;
+import java.util.Comparator;
+import java.util.List;
+
+/** Hudi record reader that exposes primary-key ordering for Fluss union read. */
+public class HudiSortedRecordReader implements SortedRecordReader {
```
luoyuxia: Does the underlying hudi record reader ensure the records emit is ordered? If not, we can just remove HudiSortedRecordReader like iceberg. The SortedRecordReader only needed for  batch union read primary key table. We can leave it unsupported. That's fine, we also don't support for icebrg.

fhan688: > Does the underlying hudi record reader ensure the records emit is ordered? If not, we can just remove HudiSortedRecordReader like iceberg. The SortedRecordReader only needed for batch union read primary key table. We can leave it unsupported. That's fine, we also don't support for icebrg.

Thanks for pointing this out. You are right that the current Hudi reader path does not guarantee records are emitted in primary-key order.

  `SortedRecordReader` requires `read()` to return records ordered by `order()`, and `SortMergeReader` relies on each lake iterator being sorted. Since `HudiRecordReader` currently delegates to
  Hudi file-slice readers directly, we cannot safely claim that contract here.

  I removed `HudiSortedRecordReader` and its tests in the latest update. This PR now only keeps Hudi predicate pushdown support. Hudi primary-key batch union read can stay unsupported for now,
  consistent with the current Iceberg behavior.

fluss-lake/fluss-lake-hudi/src/main/java/org/apache/fluss/lake/hudi/source/HudiLakeSource.java
luoyuxia, line 99: do we need to pass `predicates` to planner to reduce the plan splits?

luoyuxia (commented):
@fhan688 Thanks for the pr. LGTM overall. Just one question

fluss-lake/fluss-lake-hudi/src/main/java/org/apache/fluss/lake/hudi/source/HudiLakeSource.java
fhan688: > do we need to pass `predicates` to planner to reduce the plan splits?

 Good point. Passing predicates to the planner would help reduce planned splits, similar to Iceberg/Paimon.

  For this PR, the predicates are applied in `HudiRecordReader`, so the result correctness is preserved, but split planning still scans all partition/file slices. Hudi's current planner path uses
  `HoodieTableFileSystemView` directly, and the converted `ExpressionPredicates` are mainly reader/parquet predicates rather than a planner-level scan filter API.

  I think planner-level pruning, especially partition pruning, is a valid follow-up optimization. If you prefer, I can keep this PR focused on reader-level predicate pushdown and leave Hudi split pruning to a separate PR.

luoyuxia: Thanks for the explanation

Issue #3279, opened 2026-05-08
Introduce Tag field in TieringSplit


fhan688:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/fluss/issues) and found nothing similar.

### Description

Introduce Tag field in TieringSplit

### Willingness to contribute

- [x] I'm willing to submit a PR!

luoyuxia:
Closed via #3503

PR #3503, opened 2026-06-21
[lake/hudi] Introduce tag field for tiering splits

fhan688:
### Purpose

  Linked issue: https://github.com/apache/fluss/issues/3279

  This PR introduces split tags for the tiering source so lake writers can coordinate all splits that belong to the same tiering round. Each generated tiering round has one first-split tag and a shared round tag for the  remaining splits.

### Brief change log

  - Add `WriterInitContext#tag()` as a default method.
  - Add immutable tag support to `TieringSplit`, `TieringLogSplit`, and `TieringSnapshotSplit`.
  - Populate tiering-round tags when generating splits.
  - Preserve split tags through copy, split state conversion, and split serialization.
  - Pass split tags from `TieringSplitReader` to `TieringWriterInitContext`.
  - Add tests for tag serialization, state conversion, and tag assignment.

### Tests

  - `git diff --check`
  - `mvn -pl fluss-flink/fluss-flink-common -am -DskipITs -Dcheckstyle.skip=true -DfailIfNoTests=false -Dtest=TieringSplitSerializerTest,TieringSplitStateTest test`

### API and Format

  - Adds a source-compatible default method `WriterInitContext#tag()`.
  - Updates internal `TieringSplitSerializer` version to include split tags.
  - Does not change Fluss table data format or lake storage format.

### Documentation

  No user-facing documentation changes.

### DIFF
--- a/fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/tiering/source/split/TieringSplitSerializer.java
+++ b/fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/tiering/source/split/TieringSplitSerializer.java
@@ -39,7 +39,7 @@ public class TieringSplitSerializer implements SimpleVersionedSerializer<Tiering
 
     public static final TieringSplitSerializer INSTANCE = new TieringSplitSerializer();
 
-    private static final int VERSION_1 = 1;
+    private static final int VERSION_0 = 0;
 
     private static final ThreadLocal<DataOutputSerializer> SERIALIZER_CACHE =
             ThreadLocal.withInitial(() -> new DataOutputSerializer(64));
@@ -47,7 +47,7 @@ public class TieringSplitSerializer implements SimpleVersionedSerializer<Tiering
     private static final byte TIERING_SNAPSHOT_SPLIT_FLAG = 1;
     private static final byte TIERING_LOG_SPLIT_FLAG = 2;
 
-    private static final int CURRENT_VERSION = VERSION_1;
+    private static final int CURRENT_VERSION = VERSION_0;
 
     @Override
     public int getVersion() {
@@ -86,8 +86,9 @@ public class TieringSplitSerializer implements SimpleVersionedSerializer<Tiering
         out.writeInt(split.getNumberOfSplits());
         // write skipCurrentRound
         out.writeBoolean(split.shouldSkipCurrentRound());
-        // write split tag
-        out.writeUTF(split.getTag());
+        // write split metadata for this tiering round
+        out.writeInt(split.getSplitIndex());
+        out.writeLong(split.getTieringRoundTimestamp());
         if (split.isTieringSnapshotSplit()) {
             // Snapshot split
             TieringSnapshotSplit tieringSnapshotSplit = split.asTieringSnapshotSplit();
@@ -111,7 +112,7 @@ public class TieringSplitSerializer implements SimpleVersionedSerializer<Tiering
 
     @Override
     public TieringSplit deserialize(int version, byte[] serialized) throws IOException {
-        if (version != VERSION_1) {
+        if (version != VERSION_0) {
             throw new IOException("Unknown version " + version);
         }
         final DataInputDeserializer in = new DataInputDeserializer(serialized);
@@ -141,7 +142,8 @@ public class TieringSplitSerializer implements SimpleVersionedSerializer<Tiering
         // deserialize number of splits
         int numberOfSplits = in.readInt();
         boolean skipCurrentRound = in.readBoolean();
-        String tag = in.readUTF();
+        int splitIndex = in.readInt();
+        long tieringRoundTimestamp = in.readLong();
 
         if (splitKind == TIERING_SNAPSHOT_SPLIT_FLAG) {
             // deserialize snapshot id
@@ -156,7 +158,8 @@ public class TieringSplitSerializer implements SimpleVersionedSerializer<Tiering
                     logOffsetOfSnapshot,
                     numberOfSplits,
                     skipCurrentRound,
-                    tag);
+                    splitIndex,
+                    tieringRoundTimestamp);
         } else {
             // deserialize starting offset
             long startingOffset = in.readLong();
@@ -170,7 +173,8 @@ public class TieringSplitSerializer implements SimpleVersionedSerializer<Tiering
                     stoppingOffset,
                     numberOfSplits,
                     skipCurrentRound,
-                    tag);
+                    splitIndex,
+                    tieringRoundTimestamp);
         }
     }
 }
--- a/fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/tiering/source/split/TieringSplitSerializer.java
+++ b/fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/tiering/source/split/TieringSplitSerializer.java
@@ -39,7 +39,7 @@ public class TieringSplitSerializer implements SimpleVersionedSerializer<Tiering
 
     public static final TieringSplitSerializer INSTANCE = new TieringSplitSerializer();
 
-    private static final int VERSION_0 = 0;
+    private static final int VERSION_1 = 1;
 
     private static final ThreadLocal<DataOutputSerializer> SERIALIZER_CACHE =
             ThreadLocal.withInitial(() -> new DataOutputSerializer(64));
@@ -47,7 +47,7 @@ public class TieringSplitSerializer implements SimpleVersionedSerializer<Tiering
     private static final byte TIERING_SNAPSHOT_SPLIT_FLAG = 1;
     private static final byte TIERING_LOG_SPLIT_FLAG = 2;
 
-    private static final int CURRENT_VERSION = VERSION_0;
+    private static final int CURRENT_VERSION = VERSION_1;
 
     @Override
     public int getVersion() {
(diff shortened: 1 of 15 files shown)


fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/tiering/source/split/TieringSplitSerializer.java
```diff
@@ -39,15 +39,15 @@ public class TieringSplitSerializer implements SimpleVersionedSerializer<Tiering
 
     public static final TieringSplitSerializer INSTANCE = new TieringSplitSerializer();
 
-    private static final int VERSION_0 = 0;
+    private static final int VERSION_1 = 1;
```
luoyuxia: we can avoid bump the version number since the tiering job is statelsss and the serializer is just for network transmission.

fluss-test-coverage/src/test/java/org/apache/fluss/lake/writer/WriterInitContextCoverageTest.java
```diff
@@ -0,0 +1,61 @@
...
+import org.apache.fluss.metadata.TablePath;
+
+import org.junit.jupiter.api.Test;
+
+import javax.annotation.Nullable;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+/** Tests for {@link WriterInitContext}. */
+class WriterInitContextCoverageTest {
```
luoyuxia: seems the test doesn't have much meaning, we can remove this test.

luoyuxia:
@fhan688 Thanks for the PR! I spent quite a while trying to understand the purpose of the `tag` field, and I have some concerns about this approach.

The current `tag` feels a bit hacky and hard to follow. It's a free-form `String` that overloads two meanings via the `"firstSplit-"` prefix plus a timestamp, and the only way to interpret it is by string-parsing the prefix. I can imagine the Hudi writer will then have to parse this `tag` string as well to figure out the first split, which ends up being fairly hard-coded.

I'm not very familiar with Hudi, so please correct me if I'm wrong. But based on FIP-24, my understanding of what we actually need is:

> the Hudi writer needs to know which split is the first split, so that only the `LakeWriter` handling the first split generates the global `instantTime`, and the other writers retrieve it via `CkpMetadata`.

If that's the case, instead of encoding this into a string `tag`, what if we introduce a typed field `splitIndex` in `TieringSplit` representing the index of the current split within this tiering round? It pairs naturally with the existing `numberOfSplits` (`splitIndex` / `numberOfSplits`, similar to Flink's `getIndexOfThisSubtask()` / `getNumberOfParallelSubtasks()`), so it reads much less weird. Then the coordinator is simply the writer whose split has `splitIndex == 0`, and that writer generates the global `instantTime`.

This avoids the string parsing / hard-coding, makes the split immutable (no `setTag()`), and the "exactly one first split" property becomes structural (contiguous indices ⇒ exactly one `0`).

luoyuxia (commented):
@fhan688 Thanks for the pr.

fhan688:
> @fhan688 Thanks for the PR! I spent quite a while trying to understand the purpose of the `tag` field, and I have some concerns about this approach.
> 
> The current `tag` feels a bit hacky and hard to follow. It's a free-form `String` that overloads two meanings via the `"firstSplit-"` prefix plus a timestamp, and the only way to interpret it is by string-parsing the prefix. I can imagine the Hudi writer will then have to parse this `tag` string as well to figure out the first split, which ends up being fairly hard-coded.
> 
> I'm not very familiar with Hudi, so please correct me if I'm wrong. But based on FIP-24, my understanding of what we actually need is:
> 
> > the Hudi writer needs to know which split is the first split, so that only the `LakeWriter` handling the first split generates the global `instantTime`, and the other writers retrieve it via `CkpMetadata`.
> 
> If that's the case, instead of encoding this into a string `tag`, what if we introduce a typed field `splitIndex` in `TieringSplit` representing the index of the current split within this tiering round? It pairs naturally with the existing `numberOfSplits` (`splitIndex` / `numberOfSplits`, similar to Flink's `getIndexOfThisSubtask()` / `getNumberOfParallelSubtasks()`), so it reads much less weird. Then the coordinator is simply the writer whose split has `splitIndex == 0`, and that writer generates the global `instantTime`.
> 
> This avoids the string parsing / hard-coding, makes the split immutable (no `setTag()`), and the "exactly one first split" property becomes structural (contiguous indices ⇒ exactly one `0`).

Thanks for the detailed suggestion. I agree that overloading `tag` with both the first-split marker and a timestamp makes the protocol hard to follow.

  I double-checked the Hudi tiering writer side. Besides identifying the writer that creates the global Hudi instant, the writer also needs a round-level timestamp to avoid consuming a stale pending instant from
  `CkpMetadata`. In the current Hudi writer flow, non-first writers wait for `CkpMetadata.lastPendingInstant()`, and compare it with the split-generated timestamp to make sure the pending instant belongs to the current
  tiering round. So `splitIndex` alone can cover the first-writer election, but it cannot replace the round timestamp semantics.

  I’ll update the design to make these two meanings explicit instead of encoding them into a free-form string:

  - introduce a typed `splitIndex` in `TieringSplit`, where `splitIndex == 0` identifies the writer responsible for creating the Hudi instant;
  - introduce an explicit round timestamp field, for example `tieringRoundTimestamp`, to let Hudi writers validate the pending instant from `CkpMetadata`;
  - remove the `"firstSplit-"` string parsing and avoid mutable `setTag()`.

  This should keep the first-split property structural while preserving the timestamp requirement needed by the Hudi writer.

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/tiering/source/split/TieringSplitSerializer.java
fhan688: > we can avoid bump the version number since the tiering job is statelsss and the serializer is just for network transmission.

Thanks for the explanation. That makes sense.

  Since the tiering source is stateless and this serializer is only used for split transmission, I’ll avoid bumping the serializer version here and keep it as version 0 while carrying the new fields in the current
  serialization format.

fluss-test-coverage/src/test/java/org/apache/fluss/lake/writer/WriterInitContextCoverageTest.java
fhan688: > seems the test doesn't have much meaning, we can remove this test.

Thanks for pointing this out. I agree this standalone test is not very meaningful from a behavior perspective.

  I added it only because the new default method in `WriterInitContext` triggered the JaCoCo per-class coverage check. I’ll remove this coverage-only test and move the coverage to a more natural place, or avoid introducing
  a default method that needs such artificial coverage if the API is updated based on the split metadata redesign.

luoyuxia:
> > @fhan688 Thanks for the PR! I spent quite a while trying to understand the purpose of the `tag` field, and I have some concerns about this approach.
> > The current `tag` feels a bit hacky and hard to follow. It's a free-form `String` that overloads two meanings via the `"firstSplit-"` prefix plus a timestamp, and the only way to interpret it is by string-parsing the prefix. I can imagine the Hudi writer will then have to parse this `tag` string as well to figure out the first split, which ends up being fairly hard-coded.
> > I'm not very familiar with Hudi, so please correct me if I'm wrong. But based on FIP-24, my understanding of what we actually need is:
> > > the Hudi writer needs to know which split is the first split, so that only the `LakeWriter` handling the first split generates the global `instantTime`, and the other writers retrieve it via `CkpMetadata`.
> > 
> > 
> > If that's the case, instead of encoding this into a string `tag`, what if we introduce a typed field `splitIndex` in `TieringSplit` representing the index of the current split within this tiering round? It pairs naturally with the existing `numberOfSplits` (`splitIndex` / `numberOfSplits`, similar to Flink's `getIndexOfThisSubtask()` / `getNumberOfParallelSubtasks()`), so it reads much less weird. Then the coordinator is simply the writer whose split has `splitIndex == 0`, and that writer generates the global `instantTime`.
> > This avoids the string parsing / hard-coding, makes the split immutable (no `setTag()`), and the "exactly one first split" property becomes structural (contiguous indices ⇒ exactly one `0`).
> 
> Thanks for the detailed suggestion. I agree that overloading `tag` with both the first-split marker and a timestamp makes the protocol hard to follow.
> 
> I double-checked the Hudi tiering writer side. Besides identifying the writer that creates the global Hudi instant, the writer also needs a round-level timestamp to avoid consuming a stale pending instant from `CkpMetadata`. In the current Hudi writer flow, non-first writers wait for `CkpMetadata.lastPendingInstant()`, and compare it with the split-generated timestamp to make sure the pending instant belongs to the current tiering round. So `splitIndex` alone can cover the first-writer election, but it cannot replace the round timestamp semantics.
> 
> I’ll update the design to make these two meanings explicit instead of encoding them into a free-form string:
> 
> * introduce a typed `splitIndex` in `TieringSplit`, where `splitIndex == 0` identifies the writer responsible for creating the Hudi instant;
> * introduce an explicit round timestamp field, for example `tieringRoundTimestamp`, to let Hudi writers validate the pending instant from `CkpMetadata`;
> * remove the `"firstSplit-"` string parsing and avoid mutable `setTag()`.
> 
> This should keep the first-split property structural while preserving the timestamp requirement needed by the Hudi writer.

Thanks for digging into the Hudi writer side — that makes sense

fluss-test-coverage/src/test/java/org/apache/fluss/lake/writer/WriterInitContextCoverageTest.java
luoyuxia: we can just ignore the test coverage for WriterInitContext.

luoyuxia (approved):
@fhan688 Thanks. +1

PR #3505, opened 2026-06-21
[flink] Preserve offsets for empty tiering buckets

### LINKED ISSUE #3413: [flink] Preserve tiering offsets for empty-batch buckets
### Search before asking

### Fluss version

main (development)

### Please describe the bug 🐞

Related to #2371 and the fix being discussed in #3242.

When tiering a table where some buckets produce lake write results and other buckets only advance through empty materialized batches, the empty-batch buckets can lose their advanced log offsets during commit.

Triggering sequence:

1. A tiering round finishes multiple bucket splits for the same table.
2. At least one bucket writes real records to the lake and returns a non-null `TableBucketWriteResult.writeResult()`.
3. At least one other bucket reaches its stopping offset through empty batches, for example duplicate upserts under the `FIRST_ROW` merge engine, and returns a `TableBucketWriteResult` with a valid `logEndOffset` but `writeResult() == null`.
4. `TieringCommitOperator` filters the completed bucket results to non-empty write results before collecting `logEndOffsets`.
5. The lake commit proceeds because some buckets are non-empty, but the offset-only bucket is absent from the `logEndOffsets` committed to Fluss.

Expected behavior: when a commit contains any lake data, Fluss should commit log offset metadata for all completed buckets that have a known `logEndOffset`, including buckets that only advanced through empty batches.

Actual behavior: offset-only buckets are filtered out together with their null lake write result, so their committed lake offset may not advance even though the split finished.

### Solution

Keep using only non-null lake write results for the lake committer, but collect `logEndOffsets` and max tiered timestamps from all completed `TableBucketWriteResult`s whose `logEndOffset` is known.

This intentionally does not add support for a fully offset-only table commit where every bucket has `writeResult() == null`; it only fixes the mixed case where the same commit already contains lake data from other buckets.

hutiefang76:
## What is changed

Fixes #3413.

This change keeps lake writes limited to non-empty bucket write results, but collects committed Fluss log offsets and max tiered timestamps from all completed bucket write results in a mixed commit round. That preserves offsets for buckets that only advanced through an empty materialized batch while another bucket produced lake data in the same commit.

The existing empty-write-result test now asserts that the empty bucket offset is included in the committed lake snapshot metadata.

## Why

Before this change, `TieringCommitOperator` built the `logEndOffsets` map from only non-empty write results. In a mixed commit, empty-batch buckets with `writeResult() == null` were filtered out, so their completed log offsets were not committed even though the commit round had lake data to publish.

## Tests

- Red/green verified `TieringCommitOperatorTest#testCommitMeetsEmptyWriteResult`; before the production change it failed because bucket 0 offset was missing from the committed snapshot.
- `mvn -s <empty-settings> -pl fluss-flink/fluss-flink-common -am -Dtest=TieringCommitOperatorTest#testCommitMeetsEmptyWriteResult -Dsurefire.failIfNoSpecifiedTests=false test`
- `mvn -s <empty-settings> -pl fluss-flink/fluss-flink-common -am -Dtest=TieringCommitOperatorTest -Dsurefire.failIfNoSpecifiedTests=false test`
- `git diff --check`

I used an empty temporary Maven settings file locally to avoid a private mirror configured in my user Maven settings.

## AI assistance

This PR was prepared with assistance from OpenAI Codex. I reviewed the change and verified the tests above.

### DIFF
--- a/fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/tiering/committer/TieringCommitOperator.java
+++ b/fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/tiering/committer/TieringCommitOperator.java
@@ -232,7 +232,7 @@ public class TieringCommitOperator<WriteResult, Committable>
 
             Map<TableBucket, Long> logEndOffsets = new HashMap<>();
             Map<TableBucket, Long> logMaxTieredTimestamps = new HashMap<>();
-            for (TableBucketWriteResult<WriteResult> writeResult : nonEmptyResults) {
+            for (TableBucketWriteResult<WriteResult> writeResult : committableWriteResults) {
                 TableBucket tableBucket = writeResult.tableBucket();
                 logEndOffsets.put(tableBucket, writeResult.logEndOffset());
                 logMaxTieredTimestamps.put(tableBucket, writeResult.maxTimestamp());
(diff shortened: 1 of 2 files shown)


hutiefang76:
Closing as superseded by #3893, which landed the same offset-preservation fix with broader handling for all-empty and unknown-offset cases. Thanks to the maintainers for carrying the issue through with the more complete implementation.

PR #3506, opened 2026-06-22
[flink] Refactor tiering split assertions to ignore round metadata

luoyuxia:
### Purpose

Linked issue: close #xxx

`#3503` added `splitIndex` and `tieringRoundTimestamp` to `TieringSplit`'s `equals`/`hashCode`. This broke `TieringSourceEnumeratorTest`: the expected splits constructed in the tests carry `UNKNOWN` round metadata, while the actual assigned splits carry generation-time values, so `containsExactlyInAnyOrderElementsOf` no longer matches.

This PR reworks the test assertions so that the round metadata is validated on its own terms instead of being forced into the value-equality comparison, replacing the previous `assertTieringRoundMetadata`/`clearTieringSplitMetadata` helpers that mutated the actual splits in place.

### Brief change log

- Add `assertValidTieringRound` to validate the per-round metadata of the assigned splits (split indexes cover `0..n-1`, exactly one first split, every split reports the round size, all splits share the same positive tiering round timestamp).
- Add `assertTieringSplitsMatch`, which validates the round metadata and then normalizes `splitIndex`/`tieringRoundTimestamp` away (via `copy`) before comparing the remaining fields with the regular `equals`; apply it consistently across all assignment assertions.
- Remove the now-unused `splitIndex`/`tieringRoundTimestamp` validation from `TieringSplit`.
- Trim the redundant identity/default-metadata cases in `TieringSplitSerializerTest`.

### Tests

- `TieringSourceEnumeratorTest` (9 cases) and `TieringSplitSerializerTest` (10 cases) pass: `mvn test -pl fluss-flink/fluss-flink-common -Dtest=TieringSourceEnumeratorTest,TieringSplitSerializerTest`.

### API and Format

No. Test-only change plus removal of internal validation helpers; no public API or storage format change.

### Documentation

No.

### DIFF
--- a/fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/tiering/source/split/TieringSplit.java
+++ b/fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/tiering/source/split/TieringSplit.java
@@ -87,8 +87,6 @@ public abstract class TieringSplit implements SourceSplit {
         }
         this.numberOfSplits = numberOfSplits;
         this.skipCurrentRound = skipCurrentRound;
-        validateSplitIndex(numberOfSplits, splitIndex);
-        validateTieringRoundTimestamp(tieringRoundTimestamp);
         this.splitIndex = splitIndex;
         this.tieringRoundTimestamp = tieringRoundTimestamp;
     }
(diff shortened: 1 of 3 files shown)


fresh-borzoni (approved):
LGTM 👍

Issue #3282, opened 2026-05-08
Introduce Hudi LakeCommitter


fhan688:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/fluss/issues) and found nothing similar.

### Description

Introduce Hudi LakeCommitter

### Willingness to contribute

- [x] I'm willing to submit a PR!

luoyuxia:
Closed via #3509

PR #3509, opened 2026-06-22
[lake/hudi] Introduce Hudi LakeCommitter

fhan688:
### Purpose

  Linked issue: https://github.com/apache/fluss/issues/3282

  This PR completes the Hudi lake tiering write path by introducing the Hudi `LakeCommitter`.
  Together with the existing Hudi source, split planner/reader, writer, and write result serializer changes, this enables Fluss tiering service to write records into Hudi and commit Hudi instants as readable lake snapshots.

### Brief change log

  - Add `HudiCommittable` and `HudiCommittableSerializer` for transferring Hudi write statuses from lake writers to the committer.
  - Add `HudiLakeCommitter` to:
    - aggregate `HudiWriteResult`s into a committable;
    - commit one Hudi instant with Fluss snapshot metadata;
    - mark committed/aborted instants in `CkpMetadata`;
    - discover missing Hudi snapshots committed by Fluss tiering.
  - Wire Hudi committer creation through `HudiLakeTieringFactory`.
  - Enable `HudiLakeStorage#createLakeTieringFactory`.
  - Add the required `hadoop-mapreduce-client-core` provided dependency for Hudi Parquet writing.
  - Add tests for committable serialization, lake storage factory wiring, and a local Hudi write-commit-recovery flow.

### Tests

  - `mvn -pl fluss-lake/fluss-lake-hudi -am -DskipITs -Dcheckstyle.skip=true -DfailIfNoTests=false test`

### API and Format

  - Adds a Hudi-specific lake tiering committable type and serializer version `1`.
  - Enables Hudi `LakeTieringFactory` creation from `HudiLakeStorage`.
  - No Fluss table storage format change.
  - Hudi commits produced by Fluss tiering include extra metadata such as `commit-user=__fluss_lake_tiering` and Fluss lake snapshot properties.

### Documentation

  No documentation changes in this PR.


fluss-lake/fluss-lake-hudi/src/main/java/org/apache/fluss/lake/hudi/tiering/HudiLakeCommitter.java
```diff
@@ -0,0 +1,231 @@
...
+    }
+
+    private static boolean isCommittedBy(
+            HoodieTimeline timeline, HoodieInstant instant, String commitUser) {
+        try {
+            HoodieCommitMetadata metadata = timeline.readCommitMetadata(instant);
+            Map<String, String> extraMetadata = metadata.getExtraMetadata();
+            return extraMetadata != null && commitUser.equals(extraMetadata.get(COMMITTER_USER));
+        } catch (Exception e) {
+            LOG.debug("Failed to read Hudi commit metadata for instant {}.", instant, e);
```
luoyuxia: This swallows all exceptions and treats them as "not committed by Fluss". Since getMissingLakeSnapshot sits on the duplicate-commit protection path, a transient failure while reading commit metadata (IO hiccup, temporarily unreadable metadata file, etc.) would cause an instant that was actually committed by Fluss to be classified as non-Fluss. As a result getMissingLakeSnapshot could return an older snapshot — or null — and we'd lose the guard against re-committing already-tiered data.

Suggestion, just throw exception?
```
try {
          HoodieCommitMetadata metadata = timeline.readCommitMetadata(instant);
          Map<String, String> extraMetadata = metadata.getExtraMetadata();
          // genuinely no commit-user -> not committed by Fluss
          return extraMetadata != null && commitUser.equals(extraMetadata.get(COMMITTER_USER));
      } catch (Exception e) {
          // a read failure must not be silently treated as "not committed by Fluss",
          // otherwise we may miss an already-tiered snapshot and re-commit duplicated data
          throw new IOException(
                  "Failed to read Hudi commit metadata for instant " + instant + ".", e);
      }
```

luoyuxia (commented):
@fhan688 Thanks for the pr. Lgtm overall. Just one comment

fluss-lake/fluss-lake-hudi/src/main/java/org/apache/fluss/lake/hudi/tiering/HudiLakeCommitter.java
```diff
@@ -0,0 +1,238 @@
...
+            throws IOException {
+        HoodieTimeline latestLakeTimeline =
+                getCompletedTimelineCommittedBy(FLUSS_LAKE_TIERING_COMMIT_USER);
+        Optional<HoodieInstant> latestLakeInstant =
+                latestLakeTimeline.getReverseOrderedInstantsByCompletionTime().findFirst();
+        if (!latestLakeInstant.isPresent()) {
+            return null;
+        }
+
+        long latestLakeSnapshotId = Long.parseLong(latestLakeInstant.get().requestedTime());
```
XuQianJin-Stars: `getMissingLakeSnapshot` uses `requestedTime()` which is not guaranteed to align with the value returned in `commit()`

```java
// commit
return LakeCommitResult.committedIsReadable(Long.parseLong(instant));
```

```java
// getMissingLakeSnapshot
Optional<HoodieInstant> latestLakeInstant =
        latestLakeTimeline.getReverseOrderedInstantsByCompletionTime().findFirst();
...
long latestLakeSnapshotId = Long.parseLong(latestLakeInstant.get().requestedTime());
```

The snapshot id returned in `commit()` is the Hudi `instantTime` produced by the writer (via `writeClient.startCommit()`). In `getMissingLakeSnapshot()` you order by **completion time** but read `requestedTime()`. When two commits complete out of order (C1 requested first, completed last; C2 requested last, completed first), the "latest by completion time" is C2 — but its `requestedTime` is **not** monotonic with the snapshot id returned by `commit()`. As a result the `latestLakeSnapshotId <= latestLakeSnapshotIdOfFluss` check can miss a real missing snapshot.

**Fix:** order by `requestedTime()` (i.e. `getReverseOrderedInstants()`), or compare on the same axis as `commit()`'s return value.

fluss-lake/fluss-lake-hudi/src/main/java/org/apache/fluss/lake/hudi/tiering/HudiLakeCommitter.java
```diff
@@ -0,0 +1,238 @@
...
+        String instant = entry.getKey();
+        List<WriteStatus> statuses = entry.getValue();
+
+        Map<String, String> commitMetadata = new HashMap<>(snapshotProperties);
+        commitMetadata.put(COMMITTER_USER, FLUSS_LAKE_TIERING_COMMIT_USER);
+
+        try {
+            validateWriteStatuses(instant, statuses);
+            if (writeClient.getHeartbeatClient().isHeartbeatExpired(instant)) {
+                writeClient.getHeartbeatClient().start(instant);
```
XuQianJin-Stars: restarting an expired heartbeat right before commit is unsafe

```java
if (writeClient.getHeartbeatClient().isHeartbeatExpired(instant)) {
    writeClient.getHeartbeatClient().start(instant);
}
```

In Hudi an expired heartbeat means the instant is considered abandoned by other writers; restarting it and immediately calling `commit()` is undefined in Hudi 1.x and on some versions Hudi will silently roll the instant back during commit. If you only want to tolerate a short heartbeat stall, use `HoodieHeartbeatClient#extendHeartbeat`; otherwise this fallback should be removed and the caller should retry the whole instant from scratch.

fluss-lake/fluss-lake-hudi/src/main/java/org/apache/fluss/lake/hudi/tiering/HudiLakeCommitter.java
```diff
@@ -0,0 +1,238 @@
...
+                    commitMetadata);
+            boolean committed = writeClient.commit(instant, statuses, Option.of(commitMetadata));
+            if (!committed) {
+                ckpMetadata.abortInstant(instant);
+                throw new IOException("Failed to commit Hudi instant " + instant + ".");
+            }
+
+            ckpMetadata.commitInstant(instant);
+            LOG.info("Committed Hudi instant {} successfully.", instant);
+            return LakeCommitResult.committedIsReadable(Long.parseLong(instant));
```
XuQianJin-Stars: failed commit only aborts ckpMetadata, no `rollback`, and parses instant as `long` without protection

```java
boolean committed = writeClient.commit(instant, statuses, Option.of(commitMetadata));
if (!committed) {
    ckpMetadata.abortInstant(instant);
    throw new IOException("Failed to commit Hudi instant " + instant + ".");
}
return LakeCommitResult.committedIsReadable(Long.parseLong(instant));
```

Two issues:
- When `commit` returns `false`, Hudi has already produced inflight files. Calling only `ckpMetadata.abortInstant(instant)` leaves the timeline / inflight / requested metadata behind. This is inconsistent with `abort()`  which performs both `writeClient.rollback(instant)` and `ckpMetadata.abortInstant(instant)`. Reuse that path or at least call `writeClient.rollback(instant)` here.
- `Long.parseLong(instant)` can throw `NumberFormatException` for non-numeric Hudi instant types (e.g. compaction instants `.compaction` suffixed). Wrap in `IOException` or assert numeric in `validateWriteStatuses`.

fluss-lake/fluss-lake-hudi/src/main/java/org/apache/fluss/lake/hudi/tiering/HudiLakeCommitter.java
```diff
@@ -0,0 +1,238 @@
...
+    public void abort(HudiCommittable committable) throws IOException {
+        Set<String> instants = new LinkedHashSet<>(committable.getWriteStatuses().keySet());
+        instants.addAll(committable.getCompactionWriteStatuses().keySet());
+        for (String instant : instants) {
+            try {
+                writeClient.rollback(instant);
+                ckpMetadata.abortInstant(instant);
+                LOG.info("Aborted Hudi instant {}.", instant);
+            } catch (Exception e) {
+                throw new IOException("Failed to abort Hudi instant " + instant + ".", e);
```
XuQianJin-Stars: A single failing instant prevents later instants from being rolled back. Collect exceptions (`use Throwable#addSuppressed)` and throw a combined IOException after the loop. Cleanup of "good" instants should never depend on the success of "bad" ones.

fluss-lake/fluss-lake-hudi/src/main/java/org/apache/fluss/lake/hudi/tiering/HudiLakeCommitter.java
```diff
@@ -0,0 +1,238 @@
...
+        if (extraMetadata == null) {
+            throw new IOException("Failed to load committed Hudi instant extra metadata.");
+        }
+        return new CommittedLakeSnapshot(latestLakeSnapshotId, extraMetadata);
+    }
+
+    @Override
+    public void close() throws Exception {
+        IOUtils.closeQuietly(ckpMetadata, "hudi checkpoint metadata");
+        IOUtils.closeQuietly(hudiTableInfo, "hudi table info");
```
XuQianJin-Stars, line 207: does not close the write client，`HoodieFlinkWriteClient` is `AutoCloseable` and owns a heartbeat thread pool, executor, embedded timeline server and FS view. If `hudiTableInfo.close()` does not transitively close `writeClient`, this is a real resource leak (visible as orphan threads in long-running TM JVMs). Please either close `writeClient` explicitly here, or document the ownership chain in `HudiWriteTableInfo`.

fluss-lake/fluss-lake-hudi/src/main/java/org/apache/fluss/lake/hudi/tiering/HudiCommittableSerializer.java
```diff
@@ -0,0 +1,106 @@
...
+    public int getVersion() {
+        return CURRENT_VERSION;
+    }
+
+    @Override
+    public byte[] serialize(HudiCommittable hudiCommittable) throws IOException {
+        byte[] writeStatusesBytes =
+                InstantiationUtils.serializeObject(hudiCommittable.getWriteStatuses());
+        byte[] compactionWriteStatusesBytes =
+                InstantiationUtils.serializeObject(hudiCommittable.getCompactionWriteStatuses());
```
XuQianJin-Stars: serializing `WriteStatus` via JDK Java serialization is not safe across Hudi versions
`WriteStatus` and its transitive types (`HoodieRecordLocation`, `HoodieWriteStat`, …) are `Serializable` but Hudi does not contract `serialVersionUID` stability across versions. After even a Hudi patch upgrade, Fluss checkpoints recovered from the old format may fail with `InvalidClassException`. Compare with `IcebergLakeCommitter` / `PaimonLakeCommitter` which serialize the **fields they actually need** rather than Hudi/Paimon/Iceberg internal classes.

**Recommendation:** extract the minimal fields needed for `commit()` (fileId, partitionPath, totalErrorRecords, the bytes of `HoodieWriteStat` written by Hudi's own JSON `HoodieCommitMetadata`, etc.) and serialize them explicitly. This also keeps the committable size predictable.

fluss-common/src/main/java/org/apache/fluss/row/decode/KeyDecoder.java
```diff
@@ -60,6 +60,9 @@ static KeyDecoder ofPrimaryKeyDecoder(
             if (lakeFormat == null || lakeFormat == DataLakeFormat.LANCE) {
                 return CompactedKeyDecoder.createKeyDecoder(rowType, keyFields);
             }
+            if (lakeFormat == DataLakeFormat.HUDI) {
+                return CompactedKeyDecoder.createKeyDecoder(rowType, keyFields);
```
XuQianJin-Stars: asymmetric placement of the HUDI branch hurts readability and is bug-prone

Encoder:
```java
// KeyEncoder.java L85-L88 (short-circuits at the top)
if (dataLakeFormat == DataLakeFormat.HUDI) {
}
```

Decoder:
```java
// KeyDecoder.java L60-L65 (nested inside the v1 / v2-defaultBucketKey branch)
if (lakeFormat == null || lakeFormat == DataLakeFormat.LANCE) {
    return CompactedKeyDecoder.createKeyDecoder(rowType, keyFields);
}
if (lakeFormat == DataLakeFormat.HUDI) {
    return CompactedKeyDecoder.createKeyDecoder(rowType, keyFields);
}
```

On the only combination that matters (`kvFormatVersion == 1 && !isDefaultBucketKey && lakeFormat == HUDI`) both sides happen to agree today, but the asymmetry will make future changes (e.g. introducing a HudiKeyDecoder for lossy encoding) silently diverge encoder and decoder. Please:
- Either move the HUDI short-circuit to the **same** position on both sides.
- Or remove the short-circuit on `KeyEncoder` and let the explicit decision tree handle it.

Add a comment on both sides explaining "Hudi's lake-format encoder is lossy, therefore the PK encoder must always use `CompactedKeyEncoder` regardless of `isDefaultBucketKey`".

Also: L60 and L63 in `KeyDecoder` have identical bodies — collapse them:

```java
if (lakeFormat == null
        || lakeFormat == DataLakeFormat.LANCE
        || lakeFormat == DataLakeFormat.HUDI) {
    return CompactedKeyDecoder.createKeyDecoder(rowType, keyFields);
}
```

fluss-lake/fluss-lake-hudi/src/main/java/org/apache/fluss/lake/hudi/tiering/HudiCommittable.java
```diff
@@ -0,0 +1,136 @@
...
+            @Nullable Map<String, List<WriteStatus>> statusesByInstant) {
+        if (statusesByInstant == null || statusesByInstant.isEmpty()) {
+            return Collections.emptyMap();
+        }
+
+        Map<String, List<WriteStatus>> copiedStatuses = new HashMap<>();
+        for (Map.Entry<String, List<WriteStatus>> entry : statusesByInstant.entrySet()) {
+            copiedStatuses.put(
+                    entry.getKey(),
+                    Collections.unmodifiableList(new ArrayList<>(entry.getValue())));
```
XuQianJin-Stars: `WriteStatus` lists in `Hudi` can be large (one entry per file group). Copying them twice (once in the Builder, once here) doubles peak heap during checkpoint. Since the Builder is private and not exposed, wrap with `Collections.unmodifiableMap(...)` without the deep copy.

fluss-lake/fluss-lake-hudi/src/main/java/org/apache/fluss/lake/hudi/tiering/HudiLakeCommitter.java
```diff
@@ -0,0 +1,238 @@
...
+            return null;
+        }
+
+        long latestLakeSnapshotId = Long.parseLong(latestLakeInstant.get().requestedTime());
+        if (latestLakeSnapshotIdOfFluss != null
+                && latestLakeSnapshotId <= latestLakeSnapshotIdOfFluss) {
+            return null;
+        }
+
+        HoodieCommitMetadata metadata =
```
XuQianJin-Stars, line 192: Add a null check on metadata itself (Hudi 1.x can return a sentinel empty `HoodieCommitMetadata` on some failure paths, but a future version may not), or wrap the call in `try/catch `with `IOException`.

XuQianJin-Stars (approved):
+1

fluss-client/src/main/java/org/apache/fluss/client/lookup/PrimaryKeyLookuper.java
```diff
@@ -81,7 +81,7 @@ public PrimaryKeyLookuper(
                         tableInfo.getTableConfig(),
                         tableInfo.isDefaultBucketKey());
         this.bucketKeyEncoder =
-                tableInfo.isDefaultBucketKey()
+                tableInfo.isDefaultBucketKey() && lakeFormat != DataLakeFormat.HUDI
```
luoyuxia: I think we should avoid putting the Hudi-specific `lakeFormat != DataLakeFormat.HUDI` check in the callers.

The reason for the special case makes sense: for Hudi, the primary key encoder must be lossless (`CompactedKeyEncoder`), while the bucket key encoder still needs to follow Hudi's hash-based bucket encoding. So even when `isDefaultBucketKey` is true, the bucket encoder cannot safely reuse the primary key encoder.

But this is really a `KeyEncoder` capability/strategy decision, not something `PrimaryKeyLookuper` or `UpsertWriterImpl` should know about. Could we move this into the `KeyEncoder` factory instead? For example:

  ```java
  KeyEncoder.ofBucketKeyEncoder(
          rowType,
          bucketKeys,
          tableConfig,
          isDefaultBucketKey,
          primaryKeyEncoder); and let KeyEncoder internally decide whether the primary key encoder can be reused for bucket routing. That would keep the Hudi exception in one place and avoid future callers accidentally missing the same condition.

luoyuxia (commented):
@fhan688 Thanks. Left one comment again. PTAL

fluss-lake/fluss-lake-hudi/src/main/java/org/apache/fluss/lake/hudi/tiering/HudiLakeCommitter.java
fhan688: > does not close the write client，`HoodieFlinkWriteClient` is `AutoCloseable` and owns a heartbeat thread pool, executor, embedded timeline server and FS view. If `hudiTableInfo.close()` does not transitively close `writeClient`, this is a real resource leak (visible as orphan threads in long-running TM JVMs). Please either close `writeClient` explicitly here, or document the ownership chain in `HudiWriteTableInfo`.

Thanks for pointing this out. I checked the close path and the Hudi write client is already closed through the existing committer close chain.

  `HudiLakeCommitter.close()` closes `hudiTableInfo`, and `HudiWriteTableInfo.close()` delegates to
  `IOUtils.closeQuietly(writeClient::close, "hudi write client")`. So we don't need to call `writeClient.close()` again directly in `HudiLakeCommitter`; that would just duplicate the existing close path.

luoyuxia (approved):
@fhan688 @XuQianJin-Stars Thanks all. +1

PR #3511, opened 2026-06-23
[server] Add negative cache for non-existent partition IDs in metadata requests

### LINKED ISSUE #3510: [server] Repeated metadata requests for deleted partitions cause unnecessary ZooKeeper pressure
### Search before asking

### Description

When partitions are deleted (e.g., during hourly partition rotation), clients that still hold
stale partition IDs continue to send `MetadataRequest` to the server. Each request triggers a
ZooKeeper query to resolve the partition ID, only to discover the partition no longer exists.

Under high-frequency partition rotation scenarios, this causes significant ZooKeeper pressure
from redundant lookups for the same non-existent partition IDs.

### Willingness to contribute

swuferhong:
### Purpose

Linked issue: close #3510 

During hourly partition rotation, clients holding stale partition IDs repeatedly trigger
ZooKeeper lookups that always return "not found". This adds unnecessary pressure on ZK.
The negative cache eliminates these redundant queries by remembering the "not found" result
for a configurable TTL period.

### DIFF
--- a/fluss-server/src/main/java/org/apache/fluss/server/RpcServiceBase.java
+++ b/fluss-server/src/main/java/org/apache/fluss/server/RpcServiceBase.java
@@ -626,6 +626,7 @@ public abstract class RpcServiceBase extends RpcGatewayService implements AdminR
             Optional<PhysicalTablePath> physicalTablePath =
                     metadataProvider.getPhysicalTablePathFromCache(partitionId);
             if (physicalTablePath.isPresent()) {
+                partitionNegativeCache.markExistent(partitionId);
                 partitionPaths.add(physicalTablePath.get());
             } else if (partitionNegativeCache.isKnownNonExistent(partitionId)) {
                 // Fast-path only after the positive metadata cache misses. A stale negative-cache
@@ -649,6 +650,7 @@ public abstract class RpcServiceBase extends RpcGatewayService implements AdminR
             }
             for (long partitionId : partitionIdsNotExistsInCache) {
                 if (partitionIdAndPaths.containsKey(partitionId)) {
+                    partitionNegativeCache.markExistent(partitionId);
                     partitionPaths.add(partitionIdAndPaths.get(partitionId));
                 } else {
                     // Only cache when the authoritative partition assignment is also gone. A miss
--- a/fluss-server/src/main/java/org/apache/fluss/server/RpcServiceBase.java
+++ b/fluss-server/src/main/java/org/apache/fluss/server/RpcServiceBase.java
@@ -652,8 +652,12 @@ public abstract class RpcServiceBase extends RpcGatewayService implements AdminR
                 if (partitionIdAndPaths.containsKey(partitionId)) {
                     partitionPaths.add(partitionIdAndPaths.get(partitionId));
                 } else {
-                    // Mark this partition ID in the negative cache to avoid future ZK queries
-                    partitionNegativeCache.markNonExistent(partitionId);
+                    // Only cache when the authoritative partition assignment is also gone. A miss
+                    // from the scoped table-path lookup may simply mean that the request omitted
+                    // the owning table or the session is not authorized for it.
+                    if (isPartitionAssignmentMissingFromZk(partitionId)) {
+                        partitionNegativeCache.markNonExistent(partitionId);
+                    }
                     throw new PartitionNotExistException(
                             String.format(
                                     "The partition id '%d' does not exist or you don't have permission to access it.",
@@ -693,5 +697,16 @@ public abstract class RpcServiceBase extends RpcGatewayService implements AdminR
                 new HashSet<>(metadataCache.getAllAliveTabletServers(listenerName).values());
         return buildMetadataResponse(
                 coordinatorServer, aliveTabletServers, tablesMetadata, partitionsMetadata);
+    }
+
+    private boolean isPartitionAssignmentMissingFromZk(long partitionId) {
+        try {
+            return !zkClient.getPartitionAssignment(partitionId).isPresent();
+        } catch (Exception e) {
+            throw new FlussRuntimeException(
+                    String.format(
+                            "Failed to get partition assignment for partition %d.", partitionId),
+                    e);
+        }
     }
 }
--- a/fluss-server/src/main/java/org/apache/fluss/server/RpcServiceBase.java
+++ b/fluss-server/src/main/java/org/apache/fluss/server/RpcServiceBase.java
@@ -17,6 +17,7 @@
 
 package org.apache.fluss.server;
 
+import org.apache.fluss.annotation.VisibleForTesting;
 import org.apache.fluss.cluster.ServerNode;
 import org.apache.fluss.cluster.ServerType;
 import org.apache.fluss.config.cluster.ConfigEntry;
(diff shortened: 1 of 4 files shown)


fluss-server/src/main/java/org/apache/fluss/server/metadata/PartitionNegativeCache.java
```diff
@@ -0,0 +1,141 @@
...
+    /** Default TTL for negative cache entries: 10 minutes of no access. */
+    private static final Duration DEFAULT_TTL = Duration.ofMinutes(10);
+
+    /**
+     * Cleanup interval: only run eviction check when at least this many milliseconds have passed
+     * since the last cleanup.
+     */
+    private static final long CLEANUP_INTERVAL_MS = 60_000;
+
+    private final ConcurrentHashMap<Long, Long> cache;
```
loserwang1024: Why not useUse Guava Cache but  implementation by yourself? 

1. **Unreliable expiration cleanup** — maybeCleanup() is only called inside markNonExistent(). If no new partitions are marked as non-existent for a long time (in practice, markNonExistent is only triggered after a ZK query confirms non-existence), expired entries will linger in memory indefinitely. While isKnownNonExistent() lazily removes individual expired entries, it never performs a full sweep.
2. **The project already has shaded Guava, so you can use it directly**:
```java
import org.apache.fluss.shaded.guava32.com.google.common.cache.Cache;
import org.apache.fluss.shaded.guava32.com.google.common.cache.CacheBuilder;

private final Cache<Long, Boolean> negativeCache = CacheBuilder.newBuilder()
    .expireAfterAccess(10, TimeUnit.MINUTES)
    .maximumSize(10000)  // prevent OOM in extreme cases
    .build();
```
3. Advantages of Guava Cache:
* Automatic eviction: access-time-based TTL without manual maintenance
* Bounded size: maximumSize prevents unbounded growth (the current implementation has no size limit)
* Battle-tested thread safety: no need to roll your own CAS logic

@wuchong ,WDYT?

fluss-server/src/main/java/org/apache/fluss/server/RpcServiceBase.java
```diff
@@ -614,6 +623,15 @@ protected MetadataResponse processMetadataRequest(
         long[] partitionIds = request.getPartitionsIds();
         List<Long> partitionIdsNotExistsInCache = new ArrayList<>();
         for (long partitionId : partitionIds) {
+            // Fast-path: throw immediately for partition IDs known to not exist,
```
loserwang1024: I  think negative cache check should NOT be placed before metadata cache lookup.

Problem scenario — a race condition exists:
1. A new partition is being created; its ID has been assigned.
2. A client sends a metadata request, but the server's metadata cache hasn't synced yet, and ZK may not have the entry written yet either.
3. Server queries ZK, finds nothing → calls markNonExistent(partitionId).
4. Partition creation completes; metadata cache is updated via ZK watch.
5. Subsequent requests are blocked by the negative cache and can never retrieve the now-existing partition.

Suggested fix: Move the negative cache check to after the metadata cache lookup but before the ZK query:

loserwang1024 (commented):
I have left some comment

fluss-server/src/main/java/org/apache/fluss/server/metadata/PartitionNegativeCache.java
swuferhong: I have changed it to guava cache

swuferhong:
@loserwang1024 comments addressed.

swuferhong:
Hi, @loserwang1024 Could you take another look？

fluss-server/src/main/java/org/apache/fluss/server/RpcServiceBase.java
```diff
@@ -634,6 +651,12 @@ protected MetadataResponse processMetadataRequest(
                 if (partitionIdAndPaths.containsKey(partitionId)) {
                     partitionPaths.add(partitionIdAndPaths.get(partitionId));
                 } else {
+                    // Only cache when the authoritative partition assignment is also gone. A miss
+                    // from the scoped table-path lookup may simply mean that the request omitted
+                    // the owning table or the session is not authorized for it.
+                    if (isPartitionAssignmentMissingFromZk(partitionId)) {
```
loserwang1024, line 659: Look good to me, only one mirror problem: whether partitionId in partitionIdsNotExistsInCache is included in partitionNegativeCache? If not, this is no need. Maybe this can move blew `partitionIdAndPaths = zkClient.getPartitionIdAndPaths(authorizedTables)` for each existed partitionIdAndPaths.

swuferhong: Thanks for the comment. `partitionIdsNotExistsInCache` indeed does not include IDs already hit by the negative cache.
The extra `isPartitionAssignmentMissingFromZk(partitionId)` check is still needed before markNonExistent, because `getPartitionIdAndPaths(authorizedTables)` is scoped by the requested/authorized table paths. A miss there may mean the partition belongs to another table or is not visible in this request scope, rather than globally non-existent.
I addressed the mirror case by invalidating stale negative-cache entries when we confirm the partition exists:
metadata cache hit: call `partitionNegativeCache.markExistent(partitionId)`
ZK partitionIdAndPaths hit: call `partitionNegativeCache.markExistent(partitionId)`
This keeps the negative cache effective for real non-existent partitions while avoiding stale negative entries blocking existing partitions.

loserwang1024 (approved):
LGTM! @wuchong , WDYT?

fluss-server/src/main/java/org/apache/fluss/server/RpcServiceBase.java
```diff
@@ -164,9 +167,15 @@ public RpcServiceBase(
         this.metadataManager = metadataManager;
         this.authorizer = authorizer;
         this.dynamicConfigManager = dynamicConfigManager;
+        this.partitionNegativeCache = new PartitionNegativeCache();
         this.ioExecutor = ioExecutor;
```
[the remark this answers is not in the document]
swuferhong: This option feels a bit too detailed to expose as a public server configuration at this stage, and the current defaults should be sufficient for the intended use case. We can revisit making it configurable later if we see real operational needs around different partition rotation or request patterns.

fluss-server/src/main/java/org/apache/fluss/server/metadata/PartitionNegativeCache.java
```diff
@@ -0,0 +1,127 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.fluss.server.metadata;
+
+import org.apache.fluss.annotation.VisibleForTesting;
+import org.apache.fluss.shaded.guava32.com.google.common.base.Ticker;
+import org.apache.fluss.shaded.guava32.com.google.common.cache.Cache;
```
wuchong: Use `com.github.benmanes.caffeine.cache.Cache` instead, which has better performance. Server-side consistently uses caffeine cache, e.g., `PartialUpdateCache`.

swuferhong:
@wuchong copliot's comments addressed. PTAL, thanks.

wuchong (commented):
I only have a minor comment.

swuferhong:
@wuchong comments addressed.

wuchong (approved):

Issue #3283, opened 2026-05-08
Introduce Hudi Compaction


fhan688:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/fluss/issues) and found nothing similar.

### Description

Introduce Hudi Compaction

### Willingness to contribute

- [x] I'm willing to submit a PR!

luoyuxia:
closed via #3518

PR #3518, opened 2026-06-24
[lake/hudi] Introduce Hudi compaction in tiering service

fhan688:
### Purpose

  Linked issue: https://github.com/apache/fluss/issues/3283

  This PR completes the Hudi MOR tiering path by supporting Hudi compaction in the lake tiering service.

  Before this change, Hudi tiering could write and commit data files/log files, but compaction write stats were not supported by the committer. As a result, `table.datalake.auto-compaction` could not form a complete schedule -> execute -> commit lifecycle for Hudi MOR tables.

### Brief change log

  - Add `HudiCompactionService` to coordinate Hudi compaction scheduling, inflight transition, bucket/partition-scoped execution,
  commit, and rollback.
  - Extend `HudiLakeWriter` to asynchronously execute inflight compaction for MOR tables when data lake auto compaction is enabled.
  - Extend `HudiLakeCommitter` to commit compaction results after normal Hudi commit, schedule the next compaction round, and rollback compaction instants on abort.
  - Update Hudi source-side timeline handling to include completed compaction instants, so committed compaction snapshots can be discovered and read.
  - Skip empty MOR file slices during split planning/verification to handle pending or compacted file groups safely.
  - Extend Hudi tiering tests to cover the full compaction lifecycle across tiering rounds.

### Tests

  - `mvn -pl fluss-lake/fluss-lake-hudi -Dcheckstyle.skip=true spotless:apply`
  - `git diff --check`
  - `mvn -pl fluss-lake/fluss-lake-hudi -am -DskipITs=false -Dcheckstyle.skip=true -DfailIfNoTests=false
  -Dtest=HudiTieringTest,HudiTieringITCase test`

### API and Format

  No public API changes.

  No Fluss storage format changes.

  This PR uses the existing data lake auto-compaction configuration path and Hudi MOR compaction timeline semantics.

### Documentation

  This introduces functional support for Hudi MOR auto compaction in tiering service. No new configuration option is added.

### DIFF
--- a/fluss-lake/fluss-lake-hudi/src/main/java/org/apache/fluss/lake/hudi/source/HudiSplitPlanner.java
+++ b/fluss-lake/fluss-lake-hudi/src/main/java/org/apache/fluss/lake/hudi/source/HudiSplitPlanner.java
@@ -119,6 +119,9 @@ public class HudiSplitPlanner implements Planner<HudiSplit> {
                             .getLatestMergedFileSlicesBeforeOrOn(partitionPath, snapshotTime)
                             .collect(Collectors.toList());
             for (FileSlice fileSlice : fileSlices) {
+                if (fileSlice.isEmpty()) {
+                    continue;
+                }
                 splits.add(toHudiSplit(hudiTableInfo, partitionPath, fileSlice));
             }
             return splits;
--- a/fluss-lake/fluss-lake-hudi/src/main/java/org/apache/fluss/lake/hudi/source/HudiSplitPlanner.java
+++ b/fluss-lake/fluss-lake-hudi/src/main/java/org/apache/fluss/lake/hudi/source/HudiSplitPlanner.java
@@ -119,6 +119,8 @@ public class HudiSplitPlanner implements Planner<HudiSplit> {
                             .getLatestMergedFileSlicesBeforeOrOn(partitionPath, snapshotTime)
                             .collect(Collectors.toList());
             for (FileSlice fileSlice : fileSlices) {
+                // Pending MOR compaction plans can expose empty file slices; they are not readable
+                // at the requested completed instant.
                 if (fileSlice.isEmpty()) {
                     continue;
                 }
--- a/fluss-lake/fluss-lake-hudi/src/main/java/org/apache/fluss/lake/hudi/tiering/HudiCompactionService.java
+++ b/fluss-lake/fluss-lake-hudi/src/main/java/org/apache/fluss/lake/hudi/tiering/HudiCompactionService.java
@@ -274,9 +274,10 @@ public class HudiCompactionService {
         }
 
         String latestInstant = null;
-        for (Map.Entry<String, HudiWriteStats> entry : compactionWriteStats.entrySet()) {
-            String compactionInstant = entry.getKey();
-            HudiWriteStats writeStats = entry.getValue();
+        List<String> compactionInstants = new ArrayList<>(compactionWriteStats.keySet());
+        Collections.sort(compactionInstants);
+        for (String compactionInstant : compactionInstants) {
+            HudiWriteStats writeStats = compactionWriteStats.get(compactionInstant);
 
             if (writeStats.getTotalErrorRecords() > 0 && !conf.get(FlinkOptions.IGNORE_FAILED)) {
                 LOG.warn(
--- a/fluss-lake/fluss-lake-hudi/src/main/java/org/apache/fluss/lake/hudi/tiering/HudiCompactionService.java
+++ b/fluss-lake/fluss-lake-hudi/src/main/java/org/apache/fluss/lake/hudi/tiering/HudiCompactionService.java
@@ -81,7 +81,16 @@ public class HudiCompactionService {
 
     private InternalSchemaManager internalSchemaManager;
 
-    public HudiCompactionService(
+    public static HudiCompactionService forScheduler(HudiWriteTableInfo hudiTableInfo) {
+        return new HudiCompactionService(hudiTableInfo, null, null);
+    }
+
+    public static HudiCompactionService forExecutor(
+            HudiWriteTableInfo hudiTableInfo, TableBucket tableBucket, @Nullable String partition) {
+        return new HudiCompactionService(hudiTableInfo, tableBucket, partition);
+    }
+
+    private HudiCompactionService(
             HudiWriteTableInfo hudiTableInfo,
             @Nullable TableBucket tableBucket,
             @Nullable String partition) {
@@ -94,7 +103,7 @@ public class HudiCompactionService {
         this.partition = partition;
     }
 
-    public boolean scheduleCompaction() {
+    public boolean scheduleCompaction() throws IOException {
         LOG.info("Scheduling Hudi compaction for table {}.", hudiTableInfo.getTablePath());
         metaClient.reloadActiveTimeline();
         try {
@@ -107,16 +116,20 @@ public class HudiCompactionService {
             }
             return scheduled;
         } catch (Exception e) {
-            LOG.warn(
-                    "Failed to schedule Hudi compaction for table {}.",
-                    hudiTableInfo.getTablePath(),
+            throw new IOException(
+                    "Failed to schedule Hudi compaction for table "
+                            + hudiTableInfo.getTablePath()
+                            + ".",
                     e);
-            return false;
         }
     }
 
     public void markSelectedCompactionsInflight() {
         List<String> compactionInstantTimes = getSelectedCompactionInstantTimes();
+        if (compactionInstantTimes.isEmpty()) {
+            return;
+        }
+        compactionInstantTimes = validateSelectedCompactionInstantTimes(compactionInstantTimes);
         if (compactionInstantTimes.isEmpty()) {
             return;
         }
@@ -141,6 +154,8 @@ public class HudiCompactionService {
             HoodieInstant requestedInstant =
                     instantGenerator.getCompactionRequestedInstant(timestamp);
             if (pendingCompactionTimeline.containsInstant(requestedInstant)) {
+                // Each Fluss tiering table has a single lake committer, so requested->inflight
+                // transitions are serialized by the committer task for this table.
                 table.getActiveTimeline().transitionCompactionRequestedToInflight(requestedInstant);
                 LOG.info("Marked Hudi compaction instant {} as inflight.", timestamp);
             }
@@ -219,26 +234,28 @@ public class HudiCompactionService {
         for (Pair<String, HoodieCompactionPlan> planPair : compactionPlans) {
             String instantTime = planPair.getLeft();
             HoodieCompactionPlan compactionPlan = planPair.getRight();
-            for (HoodieCompactionOperation operation : compactionPlan.getOperations()) {
-                if (!belongsToCurrentWriter(operation, hudiPartitionPath, currentBucket)) {
-                    continue;
-                }
+            metaClient.reload();
+            try {
+                CompactionUtil.setAvroSchema(writeConfig, metaClient);
+                internalSchemaManager = null;
+                HoodieReaderContext<?> readerContext = createReaderContext();
+                for (HoodieCompactionOperation operation : compactionPlan.getOperations()) {
+                    if (!belongsToCurrentWriter(operation, hudiPartitionPath, currentBucket)) {
+                        continue;
+                    }
 
-                HoodieFlinkMergeOnReadTableCompactor<?> compactor =
-                        new HoodieFlinkMergeOnReadTableCompactor<>();
-                CompactionOperation compactionOperation =
-                        CompactionOperation.convertFromAvroRecordInstance(operation);
+                    HoodieFlinkMergeOnReadTableCompactor<?> compactor =
+                            new HoodieFlinkMergeOnReadTableCompactor<>();
+                    CompactionOperation compactionOperation =
+                            CompactionOperation.convertFromAvroRecordInstance(operation);
 
-                metaClient.reload();
-                try {
-                    CompactionUtil.setAvroSchema(writeConfig, metaClient);
                     List<WriteStatus> writeStatuses =
                             compactor.compact(
                                     writeConfig,
                                     compactionOperation,
                                     instantTime,
                                     table.getTaskContextSupplier(),
-                                    createReaderContext(true),
+                                    readerContext,
                                     table);
                     writeStatusesByInstant
                             .computeIfAbsent(instantTime, ignored -> new ArrayList<>())
@@ -250,16 +267,11 @@ public class HudiCompactionService {
                             hudiPartitionPath,
                             currentBucket.getBucket(),
                             instantTime);
-                } catch (Exception e) {
-                    throw new IOException(
-                            String.format(
-                                    "Failed to execute Hudi compaction for table %s, partition %s, bucket %s, instant %s.",
-                                    hudiTableInfo.getTablePath(),
-                                    hudiPartitionPath,
-                                    currentBucket.getBucket(),
-                                    instantTime),
-                            e);
                 }
+            } catch (Exception e) {
+                writeStatusesByInstant.remove(instantTime);
+                throw rollbackCompactionAfterFailure(
+                        instantTime, hudiPartitionPath, currentBucket, e);
             }
         }
         return writeStatusesByInstant;
@@ -273,9 +285,9 @@ public class HudiCompactionService {
             return null;
         }
 
-        String latestInstant = null;
         List<String> compactionInstants = new ArrayList<>(compactionWriteStats.keySet());
         Collections.sort(compactionInstants);
+        String latestInstant = compactionInstants.get(compactionInstants.size() - 1);
         for (String compactionInstant : compactionInstants) {
             HudiWriteStats writeStats = compactionWriteStats.get(compactionInstant);
 
@@ -295,9 +307,6 @@ public class HudiCompactionService {
             }
 
             doCommitCompaction(compactionInstant, writeStats, snapshotProperties);
-            if (latestInstant == null || compactionInstant.compareTo(latestInstant) > 0) {
-                latestInstant = compactionInstant;
-            }
             LOG.info("Committed Hudi compaction instant {}.", compactionInstant);
         }
         return latestInstant;
@@ -322,10 +331,51 @@ public class HudiCompactionService {
                 .collect(Collectors.toList());
     }
 
-    private HoodieReaderContext<?> createReaderContext(boolean needReloadMetaClient) {
+    private List<String> validateSelectedCompactionInstantTimes(
+            List<String> compactionInstantTimes) {
+        HoodieTimeline pendingCompactionTimeline =
+                table.getActiveTimeline().filterPendingCompactionTimeline();
+        InstantGenerator instantGenerator = table.getInstantGenerator();
+        List<String> validCompactionInstantTimes = new ArrayList<>();
+        for (String timestamp : compactionInstantTimes) {
+            HoodieCompactionPlan compactionPlan = loadCompactionPlan(timestamp);
+            if (isValidCompactionPlan(compactionPlan)) {
+                validCompactionInstantTimes.add(timestamp);
+                continue;
+            }
+
+            HoodieInstant inflightInstant =
+                    instantGenerator.getCompactionInflightInstant(timestamp);
+            if (pendingCompactionTimeline.containsInstant(inflightInstant)) {
+                LOG.warn(
+                        "Rollback invalid inflight Hudi compaction instant {} for table {}.",
+                        timestamp,
+                        hudiTableInfo.getTablePath());
+                table.rollbackInflightCompaction(
+                        inflightInstant, writeClient.getTransactionManager());
+                metaClient.reloadActiveTimeline();
+            } else {
+                LOG.warn(
+                        "Skip invalid Hudi compaction plan {} for table {}.",
+                        timestamp,
+                        hudiTableInfo.getTablePath());
+            }
+        }
+        return validCompactionInstantTimes;
+    }
+
+    private HoodieCompactionPlan loadCompactionPlan(String timestamp) {
+        try {
+            return CompactionUtils.getCompactionPlan(metaClient, timestamp);
+        } catch (Exception e) {
+            throw new HoodieException("Failed to get Hudi compaction plan " + timestamp, e);
+        }
+    }
+
+    private HoodieReaderContext<?> createReaderContext() {
         Supplier<InternalSchemaManager> internalSchemaManagerSupplier =
                 () -> {
-                    if (internalSchemaManager == null || needReloadMetaClient) {
+                    if (internalSchemaManager == null) {
                         internalSchemaManager =
                                 InternalSchemaManager.get(metaClient.getStorageConf(), metaClient);
                     }
@@ -338,6 +388,33 @@ public class HudiCompactionService {
                 Collections.emptyList(),
                 metaClient.getTableConfig(),
                 Option.empty());
+    }
+
+    private IOException rollbackCompactionAfterFailure(
+            String instantTime,
+            String hudiPartitionPath,
+            TableBucket currentBucket,
+            Exception cause) {
+        IOException failure =
+                new IOException(
+                        String.format(
+                                "Failed to execute Hudi compaction for table %s, partition %s, bucket %s, instant %s.",
+                                hudiTableInfo.getTablePath(),
+                                hudiPartitionPath,
+                                currentBucket.getBucket(),
+                                instantTime),
+                        cause);
+        try {
+            CompactionUtil.rollbackCompaction(
+                    table, instantTime, writeClient.getTransactionManager());
+            metaClient.reloadActiveTimeline();
+        } catch (Exception rollbackFailure) {
+            failure.addSuppressed(
+                    new IOException(
+                            "Failed to rollback Hudi compaction instant " + instantTime + ".",
+                            rollbackFailure));
+        }
+        return failure;
     }
 
     private void doCommitCompaction(
@@ -382,11 +459,12 @@ public class HudiCompactionService {
         try {
             return BucketIdentifier.bucketIdFromFileId(fileId) == currentBucket.getBucket();
         } catch (RuntimeException e) {
-            return fileId.contains(BucketIdentifier.bucketIdStr(currentBucket.getBucket()));
+            LOG.warn("Failed to parse Hudi bucket id from file id {}.", fileId, e);
+            return false;
         }
     }
 
-    private String toHudiPartitionPath(@Nullable String partitionName) {
+    private String toHudiPartitionPath(@Nullable String partitionName) throws IOException {
         if (partitionName == null || partitionName.isEmpty()) {
             return "";
         }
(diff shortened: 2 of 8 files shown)


luoyuxia (commented):
@XuQianJin-Stars Could you please help review this pr?

fluss-lake/fluss-lake-hudi/src/main/java/org/apache/fluss/lake/hudi/source/HudiSplitPlanner.java
```diff
@@ -119,6 +119,9 @@ private List<HudiSplit> planPartition(
                             .getLatestMergedFileSlicesBeforeOrOn(partitionPath, snapshotTime)
                             .collect(Collectors.toList());
             for (FileSlice fileSlice : fileSlices) {
+                if (fileSlice.isEmpty()) {
```
XuQianJin-Stars, line 124: Skipping empty file slices is correct for MOR, but please add a short comment explaining why we can encounter an empty slice here (e.g. a pending compaction has scheduled a new file group but no base file has been written yet). Without that context the next reader will likely wonder if this is hiding a bug.

fluss-lake/fluss-lake-hudi/src/main/java/org/apache/fluss/lake/hudi/tiering/HudiCompactionService.java
```diff
@@ -0,0 +1,465 @@
...
+                                                timestamp,
+                                                CompactionUtils.getCompactionPlan(
+                                                        metaClient, timestamp));
+                                    } catch (Exception e) {
+                                        throw new HoodieException(
+                                                "Failed to get Hudi compaction plan " + timestamp,
+                                                e);
+                                    }
+                                })
+                        .filter(pair -> isValidCompactionPlan(pair.getRight()))
```
XuQianJin-Stars, line 212: Plans that fail `isValidCompactionPlan` are filtered out here, but at this point `markSelectedCompactionsInflight` has already transitioned the corresponding requested instants to inflight. Nothing in the codebase rolls those inflight instants back, which leaves them as orphan inflight compactions on the timeline forever and will block future compaction scheduling on the same file groups.

Either (a) validate the plan before marking it inflight, or (b) explicitly rollback invalid plans here via `table.rollbackInflightCompaction(...)` before filtering them out.

fluss-lake/fluss-lake-hudi/src/main/java/org/apache/fluss/lake/hudi/tiering/HudiCompactionService.java
```diff
@@ -0,0 +1,465 @@
...
+                if (!belongsToCurrentWriter(operation, hudiPartitionPath, currentBucket)) {
+                    continue;
+                }
+
+                HoodieFlinkMergeOnReadTableCompactor<?> compactor =
+                        new HoodieFlinkMergeOnReadTableCompactor<>();
+                CompactionOperation compactionOperation =
+                        CompactionOperation.convertFromAvroRecordInstance(operation);
+
+                metaClient.reload();
```
XuQianJin-Stars: `metaClient.reload()` (and `setAvroSchema` on the next line) are called once per compaction operation. A single plan typically contains many operations across the same file groups, so reloading the metaClient inside the inner loop is wasted work. Move `reload()` and `setAvroSchema()` out to the per-plan loop (around line 219), they only need to run once per instant.

fluss-lake/fluss-lake-hudi/src/main/java/org/apache/fluss/lake/hudi/tiering/HudiCompactionService.java
```diff
@@ -0,0 +1,465 @@
...
+                    List<WriteStatus> writeStatuses =
+                            compactor.compact(
+                                    writeConfig,
+                                    compactionOperation,
+                                    instantTime,
+                                    table.getTaskContextSupplier(),
+                                    createReaderContext(true),
+                                    table);
+                    writeStatusesByInstant
+                            .computeIfAbsent(instantTime, ignored -> new ArrayList<>())
```
XuQianJin-Stars, line 261: If `compactor.compact(...)` fails partway through, the writeStatuses already added to `writeStatusesByInstant` for prior operations of the same instant are silently dropped (we throw and the map never reaches `commitCompaction`), but the inflight instant on the timeline is **not** rolled back here. That can leave half-written log/base files on storage and an inflight instant that will be re-attempted later with stale data.

Please either roll back the inflight instant in a `catch`/`finally` (via `CompactionUtil.rollbackCompaction(table, instantTime, txnManager)`) before rethrowing, or document very clearly that recovery is delegated to the next round (and add an integration test that exercises it).

fluss-lake/fluss-lake-hudi/src/main/java/org/apache/fluss/lake/hudi/tiering/HudiCompactionService.java
```diff
@@ -0,0 +1,465 @@
...
+            return null;
+        }
+
+        String latestInstant = null;
+        List<String> compactionInstants = new ArrayList<>(compactionWriteStats.keySet());
+        Collections.sort(compactionInstants);
+        for (String compactionInstant : compactionInstants) {
+            HudiWriteStats writeStats = compactionWriteStats.get(compactionInstant);
+
+            if (writeStats.getTotalErrorRecords() > 0 && !conf.get(FlinkOptions.IGNORE_FAILED)) {
```
XuQianJin-Stars, line 294: `throw new IOException(...)` here will reach `HudiLakeCommitter#commitCompactionAndSchedule` (line 228) which only logs a WARN and continues. The error-record check therefore has no real effect on the main commit. Either:
- propagate the failure all the way up so the data commit also fails (preferred), or
- remove this `throw` and only roll back, since the current behaviour is misleading.
Otherwise users running with `IGNORE_FAILED=false` will silently get partial compaction results.

XuQianJin-Stars (commented):
Thanks for adding MOR auto-compaction support. Overall the architecture (async execute on writer + schedule/commit on committer) is reasonable

fluss-lake/fluss-lake-hudi/src/main/java/org/apache/fluss/lake/hudi/tiering/HudiCompactionService.java
```diff
@@ -0,0 +1,465 @@
...
+    private static boolean belongsToCurrentBucket(
+            HoodieCompactionOperation operation, TableBucket currentBucket) {
+        String fileId = operation.getFileId();
+        if (fileId == null || fileId.isEmpty()) {
+            return false;
+        }
+        try {
+            return BucketIdentifier.bucketIdFromFileId(fileId) == currentBucket.getBucket();
+        } catch (RuntimeException e) {
+            return fileId.contains(BucketIdentifier.bucketIdStr(currentBucket.getBucket()));
```
XuQianJin-Stars: The fallback `fileId.contains(BucketIdentifier.bucketIdStr(currentBucket.getBucket()))` is unsafe. For example, if `currentBucket=1` then `bucketIdStr` is `"00000001"`, which will also match a file id like `"00010001-..."` belonging to bucket 65537. Recommend dropping the fallback entirely: if `bucketIdFromFileId` throws, log a WARN with the fileId and return `false`. A wrong-bucket false positive is much worse than a missed compaction.

fluss-lake/fluss-lake-hudi/src/main/java/org/apache/fluss/lake/hudi/tiering/HudiCompactionService.java
```diff
@@ -0,0 +1,465 @@
...
+
+        List<WriteStatus> writeStatuses = new ArrayList<>(stats.size());
+        for (HoodieWriteStat stat : stats) {
+            WriteStatus writeStatus = new WriteStatus();
+            writeStatus.setStat(stat);
+            writeStatus.setFileId(stat.getFileId());
+            writeStatus.setPartitionPath(stat.getPartitionPath());
+            writeStatuses.add(writeStatus);
+        }
+        writeStatuses.get(0).setTotalErrorRecords(writeStats.getTotalErrorRecords());
```
XuQianJin-Stars: Stuffing the entire compaction's `totalErrorRecords` into the first `WriteStatus` of the list is suspicious. Downstream `CompactHelpers.createCompactionMetadata` aggregates per-status counts; doing it this way may double-count or miscount when multiple statuses are present. Please verify against Hudi's metadata aggregation logic and either set the count per-status proportionally or split it as Hudi expects.

fluss-lake/fluss-lake-hudi/src/main/java/org/apache/fluss/lake/hudi/tiering/HudiCompactionService.java
```diff
@@ -0,0 +1,465 @@
...
+        compactionConfig.archiveMinCommits = config.get(FlinkOptions.ARCHIVE_MIN_COMMITS);
+        compactionConfig.cleanPolicy = config.get(FlinkOptions.CLEAN_POLICY);
+        compactionConfig.cleanRetainCommits = config.get(FlinkOptions.CLEAN_RETAIN_COMMITS);
+        compactionConfig.cleanRetainHours = config.get(FlinkOptions.CLEAN_RETAIN_HOURS);
+        compactionConfig.cleanRetainFileVersions =
+                config.get(FlinkOptions.CLEAN_RETAIN_FILE_VERSIONS);
+        compactionConfig.compactionDeltaCommits = config.get(FlinkOptions.COMPACTION_DELTA_COMMITS);
+        compactionConfig.compactionDeltaSeconds = config.get(FlinkOptions.COMPACTION_DELTA_SECONDS);
+        compactionConfig.compactionMaxMemory = config.get(FlinkOptions.COMPACTION_MAX_MEMORY);
+        compactionConfig.compactionTargetIo = config.get(FlinkOptions.COMPACTION_TARGET_IO);
```
XuQianJin-Stars, line 542: Manually copying ~15 fields from `FlinkOptions` into `FlinkCompactionConfig` is error-prone and silently breaks every time we bump the Hudi version (newly added options will simply be ignored). Please check whether Hudi exposes a helper such as `FlinkCompactionConfig.fromConf(Configuration)` and use it; if not, at minimum add a `// TODO keep in sync with FlinkCompactionConfig` comment and a unit test that asserts every public field of `FlinkCompactionConfig` has been touched.

fluss-lake/fluss-lake-hudi/src/main/java/org/apache/fluss/lake/hudi/tiering/HudiLakeCommitter.java
```diff
@@ -67,6 +68,7 @@ public HudiLakeCommitter(
         this.hudiTableInfo = HudiWriteTableInfo.create(hudiCatalogProvider, tablePath);
         this.writeClient = hudiTableInfo.getWriteClient();
         this.ckpMetadata = ckpMetadataProvider.get(tablePath, hudiTableInfo);
+        this.compactionService = new HudiCompactionService(hudiTableInfo, null, null);
```
XuQianJin-Stars: Constructing `HudiCompactionService` here with `tableBucket=null` and `partition=null` couples the commit-side and writer-side roles into one class. The constructor of `HudiCompactionService` does not validate that callers using `executeCompaction(...)` actually have a non-null `tableBucket`, so the failure surfaces only at runtime via `checkNotNull` in `executeCompaction`. Consider splitting into `HudiCompactionScheduler` (used by committer) and `HudiCompactionExecutor` (used by writer) to make the misuse impossible.

fluss-lake/fluss-lake-hudi/src/main/java/org/apache/fluss/lake/hudi/tiering/HudiLakeCommitter.java
```diff
@@ -129,7 +129,11 @@ public LakeCommitResult commit(
 
             ckpMetadata.commitInstant(instant);
             LOG.info("Committed Hudi instant {} successfully.", instant);
-            return LakeCommitResult.committedIsReadable(parseSnapshotId(instant));
+            String latestCommittedInstant =
+                    commitCompactionAndSchedule(
+                            committable.getCompactionWriteStats(), commitMetadata);
+            return LakeCommitResult.committedIsReadable(
+                    parseSnapshotId(getLatestCommittedInstant(instant, latestCommittedInstant)));
```
XuQianJin-Stars: The Fluss snapshot id is now `max(dataInstant, compactionInstant)`. Compaction instants are minted *after* the data instant (see `commitCompactionAndSchedule`), so this id will normally be the compaction instant. That value is later parsed back with `parseSnapshotId` and used by Fluss to track the lake snapshot mapping (see line 190 `getMissingLakeSnapshot`). The next tiering round will read the data instant from the timeline and compare against the previously-stored compaction instant, breaking the idempotency check.

The Fluss snapshot id should remain the data commit instant; compaction commits should be transparent to the Fluss layer. If there is a reason this needs to be `max(...)` please add a comment explaining how `getMissingLakeSnapshot` still works correctly.

fluss-lake/fluss-lake-hudi/src/main/java/org/apache/fluss/lake/hudi/tiering/HudiLakeCommitter.java
```diff
@@ -155,22 +157,24 @@ public void abort(HudiCommittable committable) throws IOException {
                                         "Failed to abort Hudi instant " + instant + ".", e));
             }
         }
+        for (String instant : committable.getCompactionWriteStats().keySet()) {
+            try {
+                abortCompactionInstant(instant);
```
XuQianJin-Stars: Ordering: data instants are aborted first and compaction instants second. If a compaction was scheduled against base files produced by the data instant being aborted, rolling back data first leaves the inflight compaction referencing a missing base file, and the subsequent rollback may fail or corrupt the timeline. Suggest aborting compaction instants first, then data instants, and add a unit test that exercises both maps populated.

fluss-lake/fluss-lake-hudi/src/main/java/org/apache/fluss/lake/hudi/tiering/HudiLakeCommitter.java
```diff
@@ -207,6 +211,53 @@ public void close() throws Exception {
...
+        }
+
+        String latestCommittedInstant = null;
+        try {
+            latestCommittedInstant =
+                    compactionService.commitCompaction(compactionWriteStats, commitMetadata);
+            if (latestCommittedInstant != null) {
+                LOG.info("Committed Hudi compaction instant {}.", latestCommittedInstant);
+            }
+        } catch (Exception e) {
```
XuQianJin-Stars: All exceptions from `commitCompaction` are reduced to a WARN. Combined with the WARN at line 240 for `scheduleCompaction`, repeated compaction failures are entirely invisible from the outside. Please either:
- propagate, or
- expose a numeric metric (`compactionCommitFailures`, `compactionScheduleFailures`) on `LakeCommitter` so operators can alert on them.
A log-only failure surface for a long-running tiering service will be very hard to operate.

fluss-lake/fluss-lake-hudi/src/main/java/org/apache/fluss/lake/hudi/tiering/HudiLakeCommitter.java
```diff
@@ -207,6 +211,53 @@ public void close() throws Exception {
...
+                    "Failed to schedule Hudi compaction for table {}.",
+                    hudiTableInfo.getTablePath(),
+                    e);
+        }
+        return latestCommittedInstant;
+    }
+
+    private boolean isAutoCompactionEnabled() {
+        return hudiTableInfo.getTableType() == HoodieTableType.MERGE_ON_READ
+                && hudiTableInfo.getFlinkConfig().get(FlinkOptions.COMPACTION_SCHEDULE_ENABLED);
```
XuQianJin-Stars, line 248: This uses Hudi's own `FlinkOptions.COMPACTION_SCHEDULE_ENABLED` whereas `HudiLakeWriter#shouldRunCompaction` (line 174) uses Fluss-side `tableInfo.getTableConfig().isDataLakeAutoCompaction()`. The two switches must agree, otherwise the writer can run compaction execution and produce write stats that the committer then refuses to commit (or vice versa).

Please unify on a single source of truth. Easiest path: in `HudiWriteTableInfo.create`, set `COMPACTION_SCHEDULE_ENABLED` from `tableConfig.isDataLakeAutoCompaction()`; or in this method also check the Fluss flag.

fluss-lake/fluss-lake-hudi/src/main/java/org/apache/fluss/lake/hudi/tiering/HudiLakeCommitter.java
```diff
@@ -207,6 +211,53 @@ public void close() throws Exception {
...
+    }
+
+    static String getLatestCommittedInstant(
+            String dataCommitInstant, @Nullable String compactionInstant) {
+        if (compactionInstant == null || dataCommitInstant.compareTo(compactionInstant) >= 0) {
+            return dataCommitInstant;
```
XuQianJin-Stars: Two equal instants both go to `dataCommitInstant` because of `>=`. That is correct, but please add a unit test for the equal-timestamps edge case so the choice does not regress (instant timestamps in Hudi have millisecond precision and collisions, while rare, are possible).

fluss-lake/fluss-lake-hudi/src/main/java/org/apache/fluss/lake/hudi/tiering/HudiLakeCommitter.java
```diff
@@ -250,6 +301,15 @@ private void abortInstant(String instant) throws IOException {
         }
     }
 
+    private void abortCompactionInstant(String instant) throws IOException {
+        try {
+            CompactionUtil.rollbackCompaction(
```
XuQianJin-Stars, line 297: `writeClient.getHoodieTable()` is cached against the last refresh; if `abort` is invoked before any `reloadActiveTimeline()` since the inflight compaction was created, the rollback target may be stale. Add `hudiTableInfo.getMetaClient().reloadActiveTimeline()` immediately before this call (similar to what `getCompletedTimelineCommittedBy` does).

fluss-lake/fluss-lake-hudi/src/main/java/org/apache/fluss/lake/hudi/tiering/HudiLakeWriter.java
```diff
@@ -75,6 +86,16 @@ public HudiLakeWriter(
         }
 
         this.recordWriter = new HudiRecordWriter(writerInitContext, hudiTableInfo, ckpMetadata);
+        if (shouldRunCompaction(writerInitContext)) {
+            this.compactionExecutor =
+                    Executors.newSingleThreadExecutor(
+                            new ExecutorThreadFactory(
+                                    "hudi-compact-" + writerInitContext.tableBucket()));
+            this.compactionFuture = executeCompactionAsync(hudiCatalogProvider, writerInitContext);
```
XuQianJin-Stars: Starting an asynchronous compaction task inside the constructor is risky: if any line below this throws (e.g. the LOG.info call OOMs), the future is already running and references `this`/`hudiCatalogProvider` even though the object was never published to the caller. That is the classic 'this escape during construction' pattern.

Move `executeCompactionAsync(...)` to be the very last statement of the constructor, after the LOG.info, or trigger the async task lazily on the first `complete()` invocation.

fluss-lake/fluss-lake-hudi/src/main/java/org/apache/fluss/lake/hudi/tiering/HudiLakeWriter.java
```diff
@@ -91,22 +112,98 @@ public void write(LogRecord record) throws IOException {
     public HudiWriteResult complete() throws IOException {
         try {
             Map<String, List<WriteStatus>> writeStatuses = recordWriter.complete();
-            return HudiWriteResult.fromWriteStatuses(writeStatuses, Collections.emptyMap());
+            Map<String, List<WriteStatus>> compactionWriteStatuses = Collections.emptyMap();
+            if (compactionFuture != null) {
+                compactionWriteStatuses = compactionFuture.get();
```
XuQianJin-Stars: `compactionFuture.get()` has no timeout. If the compaction stalls (lock waits, slow IO, classloader leak, etc.) the entire tiering job is blocked indefinitely. Please add a configurable timeout, and on timeout `cancel(true)` the future and surface the failure (or fall back to an empty result with a metric increment).

Additionally, the surrounding `catch (Exception e)` at line 120 will catch `InterruptedException` without restoring the interrupt flag. If `get()` is interrupted, please call `Thread.currentThread().interrupt()` before wrapping the cause in `IOException`.

fluss-lake/fluss-lake-hudi/src/main/java/org/apache/fluss/lake/hudi/tiering/HudiLakeWriter.java
```diff
@@ -91,22 +112,98 @@ public void write(LogRecord record) throws IOException {
...
+
+    @Nullable
+    private IOException closeCompactionExecutor() {
+        try {
+            if (compactionFuture != null && !compactionFuture.isDone()) {
+                compactionFuture.cancel(true);
+            }
+            if (compactionExecutor != null) {
+                compactionExecutor.shutdown();
+                if (!compactionExecutor.awaitTermination(30, TimeUnit.SECONDS)) {
```
XuQianJin-Stars: Hardcoded 30s await for executor termination. This should be configurable; a long-running compaction may not stop in 30s and we currently only emit a WARN, leaking the executor. Suggest reading from `tableConfig` (e.g. `lake.hudi.tiering.compaction.shutdown-timeout`) with a sensible default, and bubbling up a failure if the executor cannot be shut down in time.

fluss-lake/fluss-lake-hudi/src/main/java/org/apache/fluss/lake/hudi/utils/HudiTableInfo.java
```diff
@@ -111,7 +111,9 @@ public static HudiTableInfo create(TablePath tablePath, Configuration hudiConfig
             HoodieTableFileSystemView fileSystemView = null;
             HoodieTableMetaClient metaClient = createMetaClient(basePath, hudiConfig);
             HoodieTimeline completedTimeline =
-                    metaClient.getCommitsTimeline().filterCompletedInstants();
+                    metaClient
+                            .getCommitsAndCompactionTimeline()
+                            .filterCompletedAndCompactionInstants();
```
XuQianJin-Stars: The semantics of `completedTimeline` have been silently changed from 'completed commits' to 'completed commits + any compaction state'. `HudiSplitPlanner.java:62` uses `getCompletedTimeline().containsInstant(snapshotTime)` to decide whether a snapshot is readable. With the new definition, an inflight or requested compaction instant will be reported as readable, which can lead a reader to plan splits over partial state.

Please keep `completedTimeline` strictly completed-only and introduce a separate accessor (e.g. `getFileSystemViewTimeline()`) for the file-system view that needs the broader timeline. The two consumers have different correctness requirements.

fluss-lake/fluss-lake-hudi/src/test/java/org/apache/fluss/lake/hudi/testutils/FlinkHudiTieringTestBase.java
```diff
@@ -187,6 +189,7 @@ protected long createPkTable(
 
         if (enableAutoCompaction) {
             tableBuilder.property(ConfigOptions.TABLE_DATALAKE_AUTO_COMPACTION.key(), "true");
```
XuQianJin-Stars, line 196: `"hudi."` is repeated in several places in this PR. Please extract a constant (the same prefix is implicitly defined elsewhere when Fluss forwards table options into Hudi config; reusing the existing constant would prevent drift).

fluss-lake/fluss-lake-hudi/src/test/java/org/apache/fluss/lake/hudi/testutils/FlinkHudiTieringTestBase.java
```diff
@@ -456,6 +460,30 @@ protected void checkFlussOffsetsInSnapshot(
...
+                                        .getCommitsAndCompactionTimeline()
+                                        .filterCompletedInstants();
+                        assertThat(
+                                        timeline.getInstantsAsStream()
+                                                .anyMatch(
+                                                        instant ->
+                                                                isCompactionInstant(
+                                                                        timeline, instant)))
+                                .isTrue();
+                    }
```
XuQianJin-Stars: A 1-minute retry timeout on `checkHudiCompactionCommitted` is generous; please reduce to 30s or expose as a constant. Also, `isCompactionInstant` reads commit metadata and may throw `IOException`; if it does, the current implementation propagates a `RuntimeException` and fails the retry. For a polling helper it would be more lenient to `return false` on read errors and let the next poll succeed.

fluss-lake/fluss-lake-hudi/src/test/java/org/apache/fluss/lake/hudi/testutils/FlinkHudiTieringTestBase.java
```diff
@@ -521,6 +549,31 @@ private List<String> collectHudiRows(
...
+        if (hudiTableInfo.getTableType() == HoodieTableType.MERGE_ON_READ) {
+            return hudiTableInfo
+                    .getFileSystemView()
+                    .getLatestMergedFileSlicesBeforeOrOn(partition, latestInstantTime)
+                    .collect(Collectors.toList());
+        }
+        return hudiTableInfo
+                .getFileSystemView()
+                .getLatestFileSlicesBeforeOrOn(partition, latestInstantTime, true)
+                .collect(Collectors.toList());
```
XuQianJin-Stars, line 557: The MOR vs COW branches use different APIs (`getLatestMergedFileSlicesBeforeOrOn` vs `getLatestFileSlicesBeforeOrOn(... , true)`) - please add a one-line comment explaining why `includeFileSliceBefore=true` is needed for COW so future readers do not flip it back.

fluss-lake/fluss-lake-hudi/src/test/java/org/apache/fluss/lake/hudi/tiering/HudiTieringITCase.java
```diff
@@ -97,7 +97,7 @@ protected static void beforeAll() {
     @Test
```
XuQianJin-Stars, line 97: Flipping `enableAutoCompaction` from `false` to `true` for the existing PK-tiering test removes coverage of the non-compaction path. Please keep the original test (auto-compaction off) and add a new test method for auto-compaction on; otherwise we lose a regression baseline.

luoyuxia:
@XuQianJin-Stars Do you still have comments?

XuQianJin-Stars:
> @XuQianJin-Stars Do you still have comments?

nothing

XuQianJin-Stars (approved):
+1

luoyuxia (approved):
@fhan688 @XuQianJin-Stars Thanks all. Merging..

PR #3396, opened 2026-05-28
[lake] Fix DvTableReadableSnapshotRetriever scanning wrong snapshot

luoyuxia:
## Summary
- `getBucketsWithoutL0AndWithL0()` passed `SCAN_SNAPSHOT_ID` and `BATCH_SCAN_MODE` via table options to `store().newScan()`, but these options are only consumed by table-level scans (`DataTableBatchScan`), not by store-level scans (`AbstractFileStoreScan`). This means the scan always hit the latest snapshot instead of the specified one.
- Fix by using the direct `.withSnapshot(snapshot)` API on `FileStoreScan`, consistent with `getBucketsWithFlushedL0()` in the same file.
- Remove unused `CoreOptions` import.

## Test Plan
- Existing tests cover the affected code paths (`DvTableReadableSnapshotRetrieverTest`, `FlinkUnionReadDvTableITCase`)
- The behavioral change is that the scan now correctly targets the specified snapshot instead of always scanning the latest one

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #3519, opened 2026-06-24
[lake][paimon] Fix DV readable snapshot and offset selection

luoyuxia:
### Purpose

Linked issue: #2121 (follow-up; already closed). Related PR: #3396.

This PR fixes **Bug 1 (scanning the wrong snapshot)**, including keeping its offsets consistent with the selected COMPACT. **Bug 2 is intentionally deferred.** The file-store scan ignores the table-level snapshot options, so it currently classifies buckets using the latest snapshot rather than the intended COMPACT.

After recovery, Fluss may also have registered a newer APPEND without registering the preceding COMPACT. For example, s1 APPEND contains bucket0 up to offset 3; s2 compacts it; recovered s3 APPEND advances bucket0 to 6 and introduces bucket1. Retrieving s2 after another APPEND must return and register `{bucket0: 3}`, without taking offset 6 or bucket1 from s3.

**Bug 2 (premature deletion of historical APPEND offsets) is deferred.** If a bucket receives new L0 files after its previous flush's APPEND metadata has been deleted, readable-snapshot advancement can pause while tiering continues. It can recover after the affected bucket's new L0 is flushed and all required snapshot offsets are available. APPENDs alone, or compaction of other buckets, do not guarantee recovery; a rarely flushed bucket can remain stalled for a long time. The regression test covers a pause followed by recovery after flushing the affected bucket.

Retaining historical APPENDs for buckets without L0 would prevent the premature deletion in Bug 2. A cold bucket or historical partition can consequently hold the table-wide retention boundary far back while its flush history remains available. Fluss retains every snapshot metadata entry from that boundary onward, so this can retain substantially more history than the individual required snapshots. We are deferring that retention change until we have an approach that also avoids excessive retention for cold buckets and historical partitions. During a Bug 2 stall, the existing `KEEP_ALL_PREVIOUS` fallback can also accumulate metadata; this PR does not resolve that tradeoff.

### Brief change log

- Select the COMPACT explicitly with `FileStoreScan.withSnapshot(snapshot)`.
- Read no-L0 offsets from the APPEND immediately preceding that COMPACT; preserve offsets for buckets whose files were removed, and resolve buckets with L0 from their flush history.
- Register the readable snapshot's complete offset maps without merging offsets from a newer APPEND. Ordinary tiered APPEND commits continue merging incremental bucket updates.
- Update the existing scenario to assert Bug 2's pause and recovery, and add a regression for snapshot/offset pairing after APPEND recovery.

### Tests

Passed on JDK 11 (16 tests, no failures or skips):

- `DvTableReadableSnapshotRetrieverTest` (3 tests): incremental offsets, Bug 2 pause/recovery, partitioned tables, and recovered APPEND offset pairing.
- `FlussTableLakeSnapshotCommitterTest` (5 tests).
- `TieringCommitOperatorTest` (6 tests).
- `FlinkUnionReadDvTableITCase` (2 tests): non-partitioned and partitioned union reads.

```sh
./mvnw -pl fluss-lake/fluss-lake-paimon -am spotless:apply test \
  -Dtest=DvTableReadableSnapshotRetrieverTest,FlussTableLakeSnapshotCommitterTest,TieringCommitOperatorTest,FlinkUnionReadDvTableITCase \
  -Dsurefire.failIfNoSpecifiedTests=false
./mvnw spotless:check validate
```

Checkstyle, Spotless, and license validation passed. The full repository `clean verify` suite was not run.

### API and Format

No public API or storage format changes. Uses the existing `ignorePreviousTableOffsets` request field for complete readable-snapshot offset maps.

### Documentation

No new feature. Updated code comments to describe snapshot-specific offsets and the existing retention limitation.

### DIFF
--- a/fluss-lake/fluss-lake-paimon/src/main/java/org/apache/fluss/lake/paimon/utils/DvTableReadableSnapshotRetriever.java
+++ b/fluss-lake/fluss-lake-paimon/src/main/java/org/apache/fluss/lake/paimon/utils/DvTableReadableSnapshotRetriever.java
@@ -32,7 +32,6 @@ import org.apache.fluss.metadata.TablePath;
 import org.apache.fluss.utils.ExceptionUtils;
 import org.apache.fluss.utils.types.Tuple2;
 
-import org.apache.paimon.CoreOptions;
 import org.apache.paimon.Snapshot;
 import org.apache.paimon.data.BinaryRow;
 import org.apache.paimon.data.BinaryString;
@@ -495,16 +494,10 @@ public class DvTableReadableSnapshotRetriever implements AutoCloseable {
         Set<PaimonPartitionBucket> bucketsWithoutL0 = new HashSet<>();
         Set<PaimonPartitionBucket> bucketsWithL0 = new HashSet<>();
 
-        // Scan the snapshot to get all splits including level0
-        Map<String, String> scanOptions = new HashMap<>();
-        scanOptions.put(CoreOptions.SCAN_SNAPSHOT_ID.key(), String.valueOf(snapshot.id()));
-        // hacky: set batch scan mode to compact to make sure we can get l0 level files
-        scanOptions.put(
-                CoreOptions.BATCH_SCAN_MODE.key(), CoreOptions.BatchScanMode.COMPACT.getValue());
-
+        // Scan the snapshot to get all data files including L0 level files
         Map<BinaryRow, Map<Integer, List<ManifestEntry>>> manifestsByBucket =
                 FileStoreScan.Plan.groupByPartFiles(
-                        fileStoreTable.copy(scanOptions).store().newScan().plan().files());
+                        fileStoreTable.store().newScan().withSnapshot(snapshot).plan().files());
 
         for (Map.Entry<BinaryRow, Map<Integer, List<ManifestEntry>>> manifestsByBucketEntry :
                 manifestsByBucket.entrySet()) {
--- a/fluss-lake/fluss-lake-paimon/src/main/java/org/apache/fluss/lake/paimon/utils/DvTableReadableSnapshotRetriever.java
+++ b/fluss-lake/fluss-lake-paimon/src/main/java/org/apache/fluss/lake/paimon/utils/DvTableReadableSnapshotRetriever.java
@@ -255,19 +255,19 @@ public class DvTableReadableSnapshotRetriever implements AutoCloseable {
             return null;
         }
 
-        // We keep snapshots because for a compacted snapshot, if a bucket has L0, we find the
-        // snapshot that exactly holds those L0, then use that snapshot's previous APPEND's tiered
-        // offset as the readable offset (that offset is safe to read). When the current compacted
-        // snapshot has no L0 in any bucket, we do not traverse; for any later compact we would
-        // traverse to (going backwards in time), if some bucket has L0, the snapshot that exactly
-        // holds that L0 must be after the current compacted snapshot on the timeline. So that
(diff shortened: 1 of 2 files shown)


fluss-lake/fluss-lake-paimon/src/main/java/org/apache/fluss/lake/paimon/utils/DvTableReadableSnapshotRetriever.java
```diff
@@ -426,6 +416,112 @@ public ReadableSnapshotResult getReadableSnapshotAndOffsets(long tieredSnapshotI
...
+                }
+            }
+        }
+
+        if (!bucketsToAnchor.isEmpty()) {
+            // some bucket's most recent flush was not found within the retained snapshots
+            allAnchorsResolved = false;
+        }
+
+        return allAnchorsResolved ? earliestSnapshotIdToKeep : LakeCommitResult.KEEP_ALL_PREVIOUS;
```
fresh-borzoni: If one bucket's flush history has aged out, this returns KEEP_ALL_PREVIOUS and throws away the bound the other buckets already gave us - so retention never shrinks while that bucket stays cold. 

The unresolvable anchor is by definition older than earliestSnapshotId (already gone from Paimon), so keep-all can't recover it anyway, flooring at earliestSnapshotId is just as safe and lets retention track Paimon's expiry. 

WDYT?

fluss-lake/fluss-lake-paimon/src/main/java/org/apache/fluss/lake/paimon/utils/DvTableReadableSnapshotRetriever.java
```diff
@@ -426,6 +416,112 @@ public ReadableSnapshotResult getReadableSnapshotAndOffsets(long tieredSnapshotI
...
+        for (long currentSnapshotId = latestCompactedSnapshotId;
+                currentSnapshotId >= earliestSnapshotId && !bucketsToAnchor.isEmpty();
+                currentSnapshotId--) {
+            Snapshot currentSnapshot = snapshotManager.tryGetSnapshot(currentSnapshotId);
+            if (currentSnapshot == null
+                    || currentSnapshot.commitKind() != Snapshot.CommitKind.COMPACT) {
+                continue;
+            }
+            // The first flush encountered going backwards is the bucket's most recent flush.
+            for (PaimonPartitionBucket partitionBucket : getBucketsWithFlushedL0(currentSnapshot)) {
```
fresh-borzoni: This re-walks the same snapshots as the with-L0 loop above, so getBucketsWithFlushedL0 scans manifests twice per compacted snapshot. Could it fold into that loop: one pass, offsets for the with-L0 buckets, anchor-only for the rest?

fresh-borzoni (commented):
@luoyuxia Thank you, looks good, left minor comments, PTAL

luoyuxia:
@fresh-borzoni Thanks for the review! I've narrowed this PR to Bug 1: scanning the intended COMPACT snapshot and keeping its offsets consistent.

I've deferred Bug 2 and removed the extra retention traversal. Retaining historical APPENDs for every no-L0 bucket can pin the table-wide retention boundary for cold buckets or historical partitions, causing excessive snapshot metadata retention.

Bug 2 can recover after the affected bucket’s new L0 files are flushed, provided the required snapshot offsets are available. The updated tests cover both the pause and recovery; all 16 related tests pass.

Could you take another look at the updated scope?

fresh-borzoni (approved):
@luoyuxia Thank you, LGTM 👍 
The narrowed scope makes sense to me, so let's defer and solve the more general retention problem separately.

PR #3521, opened 2026-06-25
[flink] add support for flink 2.3

### LINKED ISSUE #3520: [flink] support flink 2.3
### Search before asking

### Motivation

flink 2.3 release : https://flink.apache.org/2026/06/25/apache-flink-2.3.0-release-announcement/

### Solution

_No response_

### Anything else?

_No response_

### Willingness to contribute

sd4324530:
### Purpose

Linked issue: close #3520

This PR adds Flink 2.3 engine support to Apache Fluss by introducing a new `fluss-flink-2.3` module. To accommodate the API changes in `CatalogMaterializedTable` / `IntervalFreshness` introduced in Flink 2.3, corresponding version adapters are introduced in the common module so that cross-version connector code can compile against multiple Flink majors.

### Brief change log

**1. New `fluss-flink-2.3` module**

- Added `fluss-flink-2.3` as a sub-module in `fluss-flink/pom.xml` to provide connector support for Flink 2.3-based deployments.
- The module mirrors the structure of `fluss-flink-2.2` and ships with complete source and test directories. Version-specific adapters (e.g., `MultipleParameterToolAdapter`, `SchemaAdapter`, `SinkAdapter`, `TypeInformationAdapter`) follow the existing pattern and are provided alongside the module.

**2. New version adapters in the common module**

To handle the API changes of `CatalogMaterializedTable` / `IntervalFreshness` in Flink 2.3, the following adapters are introduced under `fluss-flink-common/src/main/java/org/apache/fluss/flink/adapter/`:

- `CatalogMaterializedTableAdapter`: wraps `CatalogMaterializedTable.Builder` to abstract away differences introduced in Flink 2.3 (such as the new `originalQuery` / `expandedQuery` fields), so the shared common code does not depend on a specific Flink version.
- `IntervalFreshnessAdapter`: a new adapter for `IntervalFreshness` and its inner `TimeUnit` enum. In Flink 2.3 the `IntervalFreshness.TimeUnit` type has been reworked (moved/repackaged), so this adapter provides a unified way to parse and serialize time units — converting between `String` names and the version-specific `TimeUnit` enum (via the `TimeUnitAdapter` wrapper). The common module no longer needs to depend on a Flink-version-specific `IntervalFreshness.TimeUnit` class, and the existing `MATERIALIZED_TABLE_INTERVAL_FRESHNESS_TIME_UNIT` config can keep its `stringType()` form unchanged.

  > ⚠️ **Note on the diff:** the deletion of `ResolvedCatalogMaterializedTableAdapter` and the addition of `IntervalFreshnessAdapter` may show up in the GitHub PR diff as a **rename** (51% similarity). This is a false positive: their Apache 2.0 license headers happen to push the file-level similarity above Git's 50% rename-detection threshold, but the two classes have **independent responsibilities and share no business logic**. The change is in reality a pure add + delete:
  >
  > - **Removed** (`fluss-flink-common/src/test/java/.../adapter/ResolvedCatalogMaterializedTableAdapter.java`): the old test-only static helper that faked a 2-arg `ResolvedCatalogMaterializedTable` constructor call as a workaround for https://issues.apache.org/jira/browse/FLINK-38532. It is **not** removed because FLINK-38532 has been fixed — rather, the test code has been refactored to use a template-method pattern (see point **5** below), so the workaround is no longer referenced anywhere and the helper becomes dead code.
  > - **Added** (`fluss-flink-common/src/main/java/.../adapter/IntervalFreshnessAdapter.java`): a fresh public adapter for `IntervalFreshness.TimeUnit` parsing/serialization, unrelated to the removed helper.
  >
  > Reviewers can confirm this by running `git show 21c65a6c --no-renames --name-status`, which reveals the real `A` + `D` pair.

**3. Configuration and serialization compatibility**

- `FlinkConnectorOptions.MATERIALIZED_TABLE_INTERVAL_FRESHNESS_TIME_UNIT` is changed from `enumType(IntervalFreshness.TimeUnit.class)` to `stringType()`, avoiding a hard dependency on a Flink-version-specific enum class in the common module; concrete enum parsing is delegated to `IntervalFreshnessAdapter`.
- `FlinkConversions` is updated to use the new adapters for materialization-table serialization/deserialization, ensuring consistent semantics across versions.

**4. Back-port to fluss-flink-2.2**

`CatalogMaterializedTableAdapter` is also added to `fluss-flink-2.2`, and `Flink22CatalogTest` is updated accordingly, so that the 2.2 module continues to share the same common code after the new adapter is introduced.

> Same `git rename` false positive applies here: the deletion of `fluss-flink-2.2/src/test/.../ResolvedCatalogMaterializedTableAdapter.java` and the addition of `fluss-flink-2.2/src/main/.../CatalogMaterializedTableAdapter.java` are also displayed as a 50%-similarity rename for the same license-header reason. They are independent changes.

**5. Test refactor: template-method pattern for version-specific constructors**

The old `ResolvedCatalogMaterializedTableAdapter.create()` helper masked the Flink-version-specific `ResolvedCatalogMaterializedTable` constructor signature with a 2-arg fake. To support Flink 2.3's new 5-arg constructor (which adds `StartMode`), the catalog test hierarchy is refactored to a **template-method** pattern:

- The parent `FlinkCatalogTest` (in `fluss-flink-common`) now exposes a `protected` `createResolvedCatalogMaterializedTable(...)` method with a default 2-arg-constructor implementation.
- `Flink22CatalogTest` overrides it to use the 4-arg constructor (`origin`, `resolvedSchema`, `refreshMode`, `intervalFreshness`).
- `Flink23CatalogTest` overrides it to use the 5-arg constructor (additionally passing `StartMode.of(StartMode.StartModeKind.FROM_BEGINNING)`).

This lets each Flink version exercise its native constructor signature, removes the need for the static helper, and makes the test code self-documenting about which Flink version it targets. Note that the parent `FlinkCatalogTest` no longer imports `ResolvedCatalogMaterializedTableAdapter`.

**6. Test coverage**

- A complete ITCase suite (`Flink23*ITCase`) is added under the `fluss-flink-2.3` module, covering catalog, metrics, procedure, authorization, sink, source (including binlog/changelog virtual tables, delta join, failover), and tiering.
- `Flink23MultipleParameterToolTest` is added to validate `MultipleParameterToolAdapter` behavior.
- `FlinkCatalogTest` and `Flink22CatalogTest` are updated for the adapter-related cases.

**7. Test compatibility fix: Flink 2.3 `ON CONFLICT` validation vs. Delta Join tests**

While porting the Delta Join ITCases to Flink 2.3, an unexpected upstream planner behavior change was discovered. This PR works around it in the test code only; the broader impact on Fluss end-users upgrading to Flink 2.3 is left for community discussion.

- **What changed in Flink 2.3.** Flink 2.3 introduces a new planner option `ExecutionConfigOptions.TABLE_EXEC_SINK_REQUIRE_ON_CONFLICT` (`table.exec.sink.require-on-conflict`, default **`true`**). Inside `FlinkChangelogModeInferenceProgram.SatisfyUpdateKindTraitVisitor.analyzeUpsertMaterializeStrategy`, Flink now throws `ValidationException("The query has an upsert key that differs from the primary key of the sink table ...")` whenever:
  - `TABLE_EXEC_SINK_UPSERT_MATERIALIZE = AUTO` (default), **and**
  - the sink table's primary key does not contain the query's upsert keys, **and**
  - no `ON CONFLICT` clause is supplied.

  In Flink 2.2 this validation did not exist.

- **Why this affects Fluss's Delta Join tests.** Delta Join semantics define the upsert key from the join condition, not from the sink's primary key. In several Flink23DeltaJoinITCase scenarios (e.g., `testDeltaJoinWithJoinKeyExceedsPrimaryKey` with join condition `c1=c2 AND d1=d2 AND e1=e2` into a sink whose PK is `(c1, d1)`), the upsert key `(c1, d1, e1)` legitimately exceeds the sink PK. Under Flink 2.2 these tests asserted that `StreamPhysicalDeltaJoinForceValidator` throws `The current sql doesn't support to do delta join optimization`. Under Flink 2.3 the new `ON CONFLICT` validator fires **before** the Delta Join validator is reached, so the original error message is never produced and the assertions fail.

- **Why this PR only touches the tests.** The new Flink validation is the **correct** behavior in the general case — silently allowing mismatched upsert keys leads to non-deterministic results at the sink, which is exactly what Flink 2.3 is trying to prevent. Disabling it in the connector would silently regress that protection for all Fluss users. Whether/how Fluss should expose this knob (e.g., as a Fluss-level connector option, or by injecting a `ConflictStrategy` from `FlinkTableSink` when the underlying Fluss table has a `last_row` merge engine) is a product-level decision that should be discussed with the community and is out of scope for this PR.

  This is the minimal, scoped workaround. `Flink22DeltaJoinITCase` is untouched (the option did not exist in 2.2).

**8. Test compatibility fix: Flink 2.3 `ON CONFLICT` validation vs. Table Sink partial-upsert tests**

A second, distinct impact of the same Flink 2.3 validation was uncovered while running `Flink23TableSinkITCase` on CI: the Fluss **partial-upsert** test path is also blocked. Same planner option, same root cause, but a different surface — addressed with the same minimal-scoped pattern.

- **The fix.** `Flink23TableSinkITCase` was previously an empty subclass of `FlinkTableSinkITCase`. This PR turns it into a proper subclass with a single `@BeforeEach` that disables `TABLE_EXEC_SINK_REQUIRE_ON_CONFLICT` on the streaming `TableConfig`, mirroring the precedent set in Section 7.

- **Why this PR only touches the tests.** Same reasoning as Section 7: Flink 2.3's validation is the correct general-case behavior, and silently turning it off at the connector level would regress the protection for real users. The proper follow-up is for Fluss to participate in the new model — most naturally by having `FlinkTableSink.applyOperations(...)` (or `getSinkRuntimeProvider`) inject a `ConflictStrategy` (e.g., `DEDUPLICATE`, semantically equivalent to Fluss's current first-write-wins plus target-column overwrite) when the sink is a Fluss PK table, so partial upserts remain expressible from SQL. That is a product/API change and is explicitly out of scope for this engine-port PR.

  `Flink22TableSinkITCase` is untouched (the option did not exist in 2.2).

### Tests

- Unit tests:
  - `Flink23MultipleParameterToolTest`
  - `Flink23CatalogTest`
  - `Flink23TieringCommitOperatorTest`
  - `FlinkCatalogTest`, `Flink22CatalogTest` (adapter-related cases updated; `Flink22/23CatalogTest` now exercise their native `ResolvedCatalogMaterializedTable` constructors via the new template-method override)
- Integration tests (ITCases, all under the `fluss-flink-2.3` module):
  - `Flink23CatalogITCase`, `Flink23MaterializedTableITCase`
  - `Flink23MetricsITCase`
  - `Flink23ProcedureITCase`
  - `Flink23AuthorizationITCase`
  - `Flink23ComplexTypeITCase`, **`Flink23TableSinkITCase` (now works around Flink 2.3's new `ON CONFLICT` planner validation for partial upserts, see Section 8)**, `Flink23UndoRecoveryITCase`
  - `Flink23BinlogVirtualTableITCase`, `Flink23ChangelogVirtualTableITCase`, **`Flink23DeltaJoinITCase` (now works around Flink 2.3's new `ON CONFLICT` planner validation, see Section 7)**, `Flink23TableSourceBatchITCase`, `Flink23TableSourceFailOverITCase`, `Flink23TableSourceITCase`
  - `Flink23TieringITCase`

### API and Format

No breaking changes to the public API.

### Generative AI disclosure
 - [x] Yes — Claude Code(minimax-m3)

### DIFF
--- a/fluss-flink/fluss-flink-2.3/pom.xml
+++ b/fluss-flink/fluss-flink-2.3/pom.xml
@@ -1,0 +1,268 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+ Licensed to the Apache Software Foundation (ASF) under one
+ or more contributor license agreements.  See the NOTICE file
+ distributed with this work for additional information
+ regarding copyright ownership.  The ASF licenses this file
+ to you under the Apache License, Version 2.0 (the
+ "License"); you may not use this file except in compliance
+ with the License.  You may obtain a copy of the License at
+
+      http://www.apache.org/licenses/LICENSE-2.0
+
+ Unless required by applicable law or agreed to in writing, software
+ distributed under the License is distributed on an "AS IS" BASIS,
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+-->
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+    <parent>
+        <groupId>org.apache.fluss</groupId>
+        <artifactId>fluss-flink</artifactId>
+        <version>1.0-SNAPSHOT</version>
+    </parent>
+
+    <artifactId>fluss-flink-2.3</artifactId>
+    <name>Apache Fluss (Incubating) : Engine Flink : 2.3 </name>
+    <properties>
+        <flink.major.version>2.3</flink.major.version>
+        <flink.minor.version>2.3.0</flink.minor.version>
+    </properties>
+
+    <dependencies>
+        <!-- Fluss dependency -->
+        <dependency>
+            <groupId>org.apache.fluss</groupId>
+            <artifactId>fluss-client</artifactId>
+            <version>${project.version}</version>
+        </dependency>
+
+        <dependency>
+            <groupId>org.apache.fluss</groupId>
+            <artifactId>fluss-flink-common</artifactId>
+            <version>${project.version}</version>
+            <exclusions>
+                <exclusion>
+                    <groupId>*</groupId>
+                    <artifactId>*</artifactId>
+                </exclusion>
+            </exclusions>
+        </dependency>
+
+        <!-- Flink dependency -->
+        <dependency>
+            <groupId>org.apache.flink</groupId>
+            <artifactId>flink-core</artifactId>
+            <version>${flink.minor.version}</version>
+            <scope>provided</scope>
+        </dependency>
+
+        <dependency>
+            <groupId>org.apache.flink</groupId>
+            <artifactId>flink-table-common</artifactId>
+            <version>${flink.minor.version}</version>
+            <scope>provided</scope>
+        </dependency>
+
+        <dependency>
+            <groupId>org.apache.flink</groupId>
+            <artifactId>flink-streaming-java</artifactId>
+            <version>${flink.minor.version}</version>
+            <scope>provided</scope>
+        </dependency>
+
+        <dependency>
+            <groupId>org.apache.flink</groupId>
+            <artifactId>flink-runtime</artifactId>
+            <version>${flink.minor.version}</version>
+            <scope>provided</scope>
+        </dependency>
+
+        <!-- test dependency -->
+        <dependency>
+            <groupId>org.apache.fluss</groupId>
+            <artifactId>fluss-flink-common</artifactId>
+            <version>${project.version}</version>
+            <scope>test</scope>
+            <type>test-jar</type>
+        </dependency>
+
+        <dependency>
+            <groupId>org.apache.fluss</groupId>
+            <artifactId>fluss-server</artifactId>
+            <version>${project.version}</version>
+            <scope>test</scope>
+        </dependency>
+
+        <dependency>
+            <groupId>org.apache.flink</groupId>
+            <artifactId>flink-table-test-utils</artifactId>
+            <version>${flink.minor.version}</version>
+            <scope>test</scope>
+        </dependency>
+
+        <dependency>
+            <groupId>org.apache.flink</groupId>
+            <artifactId>flink-connector-base</artifactId>
+            <version>${flink.minor.version}</version>
+            <scope>test</scope>
+        </dependency>
+
+        <dependency>
+            <groupId>org.apache.fluss</groupId>
+            <artifactId>fluss-server</artifactId>
+            <version>${project.version}</version>
+            <scope>test</scope>
+            <type>test-jar</type>
+        </dependency>
+
+        <dependency>
+            <groupId>org.apache.fluss</groupId>
+            <artifactId>fluss-rpc</artifactId>
+            <scope>test</scope>
+            <type>test-jar</type>
+            <version>${project.version}</version>
+        </dependency>
+
+        <!-- for curator TestingServer  -->
+        <dependency>
+            <groupId>org.apache.curator</groupId>
+            <artifactId>curator-test</artifactId>
+            <version>${curator.version}</version>
+            <scope>test</scope>
+        </dependency>
+
+        <dependency>
+            <groupId>org.apache.flink</groupId>
+            <artifactId>flink-table-common</artifactId>
+            <version>${flink.minor.version}</version>
+            <scope>test</scope>
+            <type>test-jar</type>
+        </dependency>
+
+        <dependency>
+            <groupId>org.apache.flink</groupId>
+            <artifactId>flink-connector-test-utils</artifactId>
+            <version>${flink.minor.version}</version>
+            <scope>test</scope>
+        </dependency>
+
+
+        <dependency>
+            <groupId>org.apache.fluss</groupId>
+            <artifactId>fluss-test-utils</artifactId>
+        </dependency>
+
+        <dependency>
+            <groupId>org.apache.fluss</groupId>
+            <artifactId>fluss-common</artifactId>
+            <version>${project.version}</version>
+            <scope>test</scope>
+            <type>test-jar</type>
+        </dependency>
+
+        <dependency>
+            <groupId>org.apache.flink</groupId>
+            <artifactId>flink-clients</artifactId>
+            <version>${flink.minor.version}</version>
+            <scope>test</scope>
+        </dependency>
+
+        <dependency>
+            <groupId>org.apache.flink</groupId>
+            <artifactId>flink-sql-gateway</artifactId>
+            <version>${flink.minor.version}</version>
+            <scope>test</scope>
+        </dependency>
+
+        <dependency>
+            <groupId>org.apache.flink</groupId>
+            <artifactId>flink-sql-gateway</artifactId>
+            <version>${flink.minor.version}</version>
+            <type>test-jar</type>
+            <scope>test</scope>
+        </dependency>
+    </dependencies>
+
+    <build>
+        <plugins>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-compiler-plugin</artifactId>
+                <configuration>
+                    <!-- compilation of main sources -->
+                    <skipMain>${skip.on.java8}</skipMain>
+                    <!-- compilation of test sources -->
+                    <skip>${skip.on.java8}</skip>
+                </configuration>
+            </plugin>
+
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-surefire-plugin</artifactId>
+                <version>3.0.0-M5</version>
+                <executions>
+                    <!-- Test end with ITCase is e2e test in this module   -->
+                    <execution>
+                        <id>integration-tests</id>
+                        <phase>integration-test</phase>
+                        <inherited>false</inherited>
+                        <goals>
+                            <goal>test</goal>
+                        </goals>
+                        <configuration>
+                            <skip>${skip.on.java8}</skip>
+                            <includes>
+                                <include>**/*ITCase.*</include>
+                            </includes>
+                            <!-- e2e test with flink/zookeeper cluster, we set forkCount=1 -->
+                            <forkCount>1</forkCount>
+                        </configuration>
+                    </execution>
+                    <!-- others unit tests -->
+                    <execution>
+                        <id>default-test</id>
+                        <phase>test</phase>
+                        <inherited>false</inherited>
+                        <goals>
+                            <goal>test</goal>
+                        </goals>
+                        <configuration>
+                            <skip>${skip.on.java8}</skip>
+                            <excludes>
+                                <exclude>**/*ITCase.*</exclude>
+                            </excludes>
+                        </configuration>
+                    </execution>
+                </executions>
+            </plugin>
+
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-shade-plugin</artifactId>
+                <executions>
+                    <execution>
+                        <id>shade-fluss</id>
+                        <phase>package</phase>
+                        <goals>
+                            <goal>shade</goal>
+                        </goals>
+                        <configuration>
+                            <artifactSet>
+                                <includes combine.children="append">
+                                    <include>org.apache.fluss:fluss-flink-common</include>
+                                    <include>org.apache.fluss:fluss-client</include>
+                                </includes>
+                            </artifactSet>
+                        </configuration>
+                    </execution>
+                </executions>
+            </plugin>
+        </plugins>
+    </build>
+
+</project>
--- a/fluss-flink/fluss-flink-2.3/pom.xml
+++ b/fluss-flink/fluss-flink-2.3/pom.xml
@@ -1,0 +1,294 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+ Licensed to the Apache Software Foundation (ASF) under one
+ or more contributor license agreements.  See the NOTICE file
+ distributed with this work for additional information
+ regarding copyright ownership.  The ASF licenses this file
+ to you under the Apache License, Version 2.0 (the
+ "License"); you may not use this file except in compliance
+ with the License.  You may obtain a copy of the License at
+
+      http://www.apache.org/licenses/LICENSE-2.0
+
+ Unless required by applicable law or agreed to in writing, software
+ distributed under the License is distributed on an "AS IS" BASIS,
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+-->
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+    <parent>
+        <groupId>org.apache.fluss</groupId>
+        <artifactId>fluss-flink</artifactId>
+        <version>1.0-SNAPSHOT</version>
+    </parent>
+
+    <artifactId>fluss-flink-2.3</artifactId>
+    <name>Apache Fluss : Engine Flink : 2.3 </name>
+    <properties>
+        <flink.major.version>2.3</flink.major.version>
+        <flink.minor.version>2.3.0</flink.minor.version>
+    </properties>
+
+    <dependencies>
+        <!-- Fluss dependency -->
+        <dependency>
+            <groupId>org.apache.fluss</groupId>
+            <artifactId>fluss-client</artifactId>
+            <version>${project.version}</version>
+        </dependency>
+
+        <dependency>
+            <groupId>org.apache.fluss</groupId>
+            <artifactId>fluss-flink-common</artifactId>
+            <version>${project.version}</version>
+            <exclusions>
+                <exclusion>
+                    <groupId>*</groupId>
+                    <artifactId>*</artifactId>
+                </exclusion>
+            </exclusions>
+        </dependency>
+
+        <!-- Flink dependency -->
+        <dependency>
+            <groupId>org.apache.flink</groupId>
+            <artifactId>flink-core</artifactId>
+            <version>${flink.minor.version}</version>
+            <scope>provided</scope>
+        </dependency>
+
+        <dependency>
+            <groupId>org.apache.flink</groupId>
+            <artifactId>flink-table-common</artifactId>
+            <version>${flink.minor.version}</version>
+            <scope>provided</scope>
+        </dependency>
+
+        <dependency>
+            <groupId>org.apache.flink</groupId>
+            <artifactId>flink-streaming-java</artifactId>
+            <version>${flink.minor.version}</version>
+            <scope>provided</scope>
+        </dependency>
+
+        <dependency>
+            <groupId>org.apache.flink</groupId>
+            <artifactId>flink-runtime</artifactId>
+            <version>${flink.minor.version}</version>
+            <scope>provided</scope>
+        </dependency>
+
+        <!-- test dependency -->
+        <dependency>
+            <groupId>org.apache.fluss</groupId>
+            <artifactId>fluss-flink-common</artifactId>
+            <version>${project.version}</version>
+            <scope>test</scope>
+            <type>test-jar</type>
+        </dependency>
+
+        <dependency>
+            <groupId>org.roaringbitmap</groupId>
+            <artifactId>RoaringBitmap</artifactId>
+            <version>${roaringbitmap.version}</version>
+        </dependency>
+
+        <dependency>
+            <groupId>org.apache.fluss</groupId>
+            <artifactId>fluss-server</artifactId>
+            <version>${project.version}</version>
+            <scope>test</scope>
+        </dependency>
+
+        <dependency>
+            <groupId>org.apache.flink</groupId>
+            <artifactId>flink-table-test-utils</artifactId>
+            <version>${flink.minor.version}</version>
+            <scope>test</scope>
+        </dependency>
+
+        <dependency>
+            <groupId>org.apache.flink</groupId>
+            <artifactId>flink-connector-base</artifactId>
+            <version>${flink.minor.version}</version>
+            <scope>test</scope>
+        </dependency>
+
+        <dependency>
+            <groupId>org.apache.fluss</groupId>
+            <artifactId>fluss-server</artifactId>
+            <version>${project.version}</version>
+            <scope>test</scope>
+            <type>test-jar</type>
+        </dependency>
+
+        <dependency>
+            <groupId>org.apache.fluss</groupId>
+            <artifactId>fluss-rpc</artifactId>
+            <scope>test</scope>
+            <type>test-jar</type>
+            <version>${project.version}</version>
+        </dependency>
+
+        <!-- for curator TestingServer  -->
+        <dependency>
+            <groupId>org.apache.curator</groupId>
+            <artifactId>curator-test</artifactId>
+            <version>${curator.version}</version>
+            <scope>test</scope>
+        </dependency>
+
+        <dependency>
+            <groupId>org.apache.flink</groupId>
+            <artifactId>flink-table-common</artifactId>
+            <version>${flink.minor.version}</version>
+            <scope>test</scope>
+            <type>test-jar</type>
+        </dependency>
+
+        <dependency>
+            <groupId>org.apache.flink</groupId>
+            <artifactId>flink-connector-test-utils</artifactId>
+            <version>${flink.minor.version}</version>
+            <scope>test</scope>
+        </dependency>
+
+
+        <dependency>
+            <groupId>org.apache.fluss</groupId>
+            <artifactId>fluss-test-utils</artifactId>
+        </dependency>
+
+        <dependency>
+            <groupId>org.apache.fluss</groupId>
+            <artifactId>fluss-common</artifactId>
+            <version>${project.version}</version>
+            <scope>test</scope>
+            <type>test-jar</type>
+        </dependency>
+
+        <dependency>
+            <groupId>org.apache.flink</groupId>
+            <artifactId>flink-clients</artifactId>
+            <version>${flink.minor.version}</version>
+            <scope>test</scope>
+        </dependency>
+
+        <dependency>
+            <groupId>org.apache.flink</groupId>
+            <artifactId>flink-sql-gateway</artifactId>
+            <version>${flink.minor.version}</version>
+            <scope>test</scope>
+        </dependency>
+
+        <dependency>
+            <groupId>org.apache.flink</groupId>
+            <artifactId>flink-sql-gateway</artifactId>
+            <version>${flink.minor.version}</version>
+            <type>test-jar</type>
+            <scope>test</scope>
+        </dependency>
+    </dependencies>
+
+    <build>
+        <plugins>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-compiler-plugin</artifactId>
+                <configuration>
+                    <!-- compilation of main sources -->
+                    <skipMain>${skip.on.java8}</skipMain>
+                    <!-- compilation of test sources -->
+                    <skip>${skip.on.java8}</skip>
+                </configuration>
+            </plugin>
+
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-surefire-plugin</artifactId>
+                <version>3.0.0-M5</version>
+                <executions>
+                    <!-- Test end with ITCase is e2e test in this module   -->
+                    <execution>
+                        <id>integration-tests</id>
+                        <phase>integration-test</phase>
+                        <inherited>false</inherited>
+                        <goals>
+                            <goal>test</goal>
+                        </goals>
+                        <configuration>
+                            <skip>${skip.on.java8}</skip>
+                            <includes>
+                                <include>**/*ITCase.*</include>
+                            </includes>
+                            <!-- e2e test with flink/zookeeper cluster, we set forkCount=1 -->
+                            <forkCount>1</forkCount>
+                        </configuration>
+                    </execution>
+                    <!-- others unit tests -->
+                    <execution>
+                        <id>default-test</id>
+                        <phase>test</phase>
+                        <inherited>false</inherited>
+                        <goals>
+                            <goal>test</goal>
+                        </goals>
+                        <configuration>
+                            <skip>${skip.on.java8}</skip>
+                            <excludes>
+                                <exclude>**/*ITCase.*</exclude>
+                            </excludes>
+                        </configuration>
+                    </execution>
+                </executions>
+            </plugin>
+
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-shade-plugin</artifactId>
+                <executions>
+                    <execution>
+                        <id>shade-fluss</id>
+                        <phase>package</phase>
+                        <goals>
+                            <goal>shade</goal>
+                        </goals>
+                        <configuration>
+                            <artifactSet>
+                                <includes combine.children="append">
+                                    <include>org.apache.fluss:fluss-flink-common</include>
+                                    <include>org.apache.fluss:fluss-client</include>
+                                    <include>org.roaringbitmap:RoaringBitmap</include>
+                                </includes>
+                            </artifactSet>
+                            <relocations combine.children="append">
+                                <relocation>
+                                    <pattern>org.roaringbitmap</pattern>
+                                    <shadedPattern>org.apache.fluss.shaded.org.roaringbitmap</shadedPattern>
+                                </relocation>
+                            </relocations>
+                            <filters combine.children="append">
+                                <filter>
+                                    <artifact>org.roaringbitmap:RoaringBitmap</artifact>
+                                    <excludes>
+                                        <exclude>META-INF/versions/**</exclude>
+                                    </excludes>
+                                </filter>
+                            </filters>
+                            <transformers combine.children="append">
+                                <transformer implementation="org.apache.maven.plugins.shade.resource.ManifestResourceTransformer">
+                                    <mainClass>org.apache.fluss.flink.action.FlussActionEntrypoint</mainClass>
+                                </transformer>
+                            </transformers>
+                        </configuration>
+                    </execution>
+                </executions>
+            </plugin>
+        </plugins>
+    </build>
+
+</project>
--- a/fluss-flink/fluss-flink-2.3/src/main/java/org/apache/fluss/flink/adapter/SinkAdapter.java
+++ b/fluss-flink/fluss-flink-2.3/src/main/java/org/apache/fluss/flink/adapter/SinkAdapter.java
@@ -1,0 +1,43 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.fluss.flink.adapter;
+
+import org.apache.flink.api.common.operators.MailboxExecutor;
+import org.apache.flink.api.connector.sink2.Sink;
+import org.apache.flink.api.connector.sink2.SinkWriter;
+import org.apache.flink.api.connector.sink2.WriterInitContext;
+import org.apache.flink.metrics.groups.SinkWriterMetricGroup;
+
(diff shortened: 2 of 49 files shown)


polyzos:
Hi @sd4324530 and thank you for all the great work 🙏 Overall the PR LGTM 👍 However there is one thing I would like to clarify.

Flink 2.3's new `table.exec.sink.require-on-conflict=true` throws a **ValidationException** for partial upserts and for delta joins whose upsert key exceeds the sink PK. The PR keeps its tests green by disabling that option in `@BeforeEach (Flink23TableSinkITCase, Flink23DeltaJoinITCase)`. That means on Flink 2.3, real users doing partial-upsert or delta-join, which are big Fluss features, will hit this exception out of the box unless they manually flip the flag. 

How do you think we should best handle this? Let me know your thoughts

sd4324530:
@polyzos 
Thank you very much for your review.
Regarding the configuration option you mentioned, I don't have a very in-depth understanding of this mechanism; I've mainly relied on reading the community documentation.
Community description of this configuration: "When a query produces an updating table with an upsert key that differs from the sink table's primary key, multiple records with different upsert keys may map to the same primary key. The ON CONFLICT clause specifies how to resolve these primary key conflicts at the sink." [1]

I think this issue has always existed and isn't a new problem introduced with Flink 2.3. If the Fluss cluster's inherent mechanisms already prevent this problem, I think we should be able to disable this configuration by default.

What do you think?

[1] https://nightlies.apache.org/flink/flink-docs-release-2.3/docs/sql/reference/dml/insert/#on-conflict-clause

polyzos:
@sd4324530 agreed 😄 @fresh-borzoni any thoughts since you are a heavy user? 😄

fresh-borzoni:
@polyzos @sd4324530 I'll take a look today, it's in my list to review 👍

fluss-flink/fluss-flink-2.3/src/main/java/org/apache/fluss/flink/adapter/SinkAdapter.java
```diff
@@ -0,0 +1,43 @@
...
+ * Flink sink adapter which hide the different version of createWriter method.
+ *
+ * <p>TODO: remove this class when no longer support all the Flink 1.x series.
+ */
+public abstract class SinkAdapter<InputT> implements Sink<InputT> {
+
+    @Override
+    public SinkWriter<InputT> createWriter(WriterInitContext writerInitContext) throws IOException {
+        return createWriter(
+                writerInitContext.getMailboxExecutor(), writerInitContext.metricGroup());
```
leonardBang: Would it make sense to keep the three-argument `createWriter` hook here and pass `writerInitContext.getTaskInfo().getIndexOfThisSubtask()`, as the Flink 2.2 adapter does? `FlinkSink` in the common module still overrides `createWriter(MailboxExecutor, SinkWriterMetricGroup, int)`, while this adapter now invokes and declares only the two-argument method. When these classes are packaged together in the Flink 2.3 artifact, creating a writer can therefore fail with an `AbstractMethodError`.

Could we align the adapter signature with the common implementation? It may also be helpful to add a regression test that creates the sink writer through the final Flink 2.3 shaded artifact.

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/utils/FlinkConversions.java
leonardBang, line 603: Would it make sense to persist the original and expanded queries separately here? In Flink 2.3, `getDefinitionQuery()` represents the expanded query, while the restore path currently uses this single stored value for `definitionQuery`, `originalQuery`, and `expandedQuery`. If the original user SQL differs from the expanded query, a catalog round trip will therefore lose the original SQL.

Could we introduce dedicated properties for the original and expanded queries, while retaining the existing definition-query property as a compatibility fallback? It may also be helpful to add a round-trip test where the original and expanded queries are intentionally different.

leonardBang (commented):
Thanks @sd4324530 for the contribution, I left two comments

fluss-flink/fluss-flink-2.3/src/main/java/org/apache/fluss/flink/adapter/SinkAdapter.java
sd4324530: This change originates from this commit:https://github.com/apache/fluss/commit/afc34bbd83c816813a937f6c883f0f004c2f59ec
I'll fix it.

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/utils/FlinkConversions.java
sd4324530: fix:https://github.com/apache/fluss/pull/3521/changes/a0a29b08dfea3c245eade7911fcff06184c5f6f2

fluss-flink/fluss-flink-2.3/src/main/java/org/apache/fluss/flink/adapter/SinkAdapter.java
sd4324530: fix:https://github.com/apache/fluss/pull/3521/changes/1c5bb53d8c408f5380357b89d10de4bd1b0a401e

fightBoxing:
## 🐛 Bug fix: `AbstractMethodError` in `SinkAdapter`

Hi @sd4324530 (and reviewers), thanks for the great work on Flink 2.3 support!

While testing this PR against a Fluss (K8s) + Flink 2.3 standalone cluster, every Flink job that writes to a Fluss table fails at runtime with:

```
java.lang.AbstractMethodError: Receiver class org.apache.fluss.flink.sink.FlinkSink
does not define or inherit an implementation of the resolved method
'createWriter(MailboxExecutor, SinkWriterMetricGroup)' of abstract class
org.apache.fluss.flink.adapter.SinkAdapter.
    at org.apache.fluss.flink.adapter.SinkAdapter.createWriter(SinkAdapter.java:37)
    at org.apache.flink.streaming.runtime.operators.sink.StatelessSinkWriterStateHandler.createWriter(...)
    ...
```

### Root cause

The abstract `createWriter` in `fluss-flink-2.3/SinkAdapter` has signature:

```java
protected abstract SinkWriter<InputT> createWriter(
        MailboxExecutor mailboxExecutor, SinkWriterMetricGroup metricGroup);   // 2 args
```

but `fluss-flink-common/FlinkSink` overrides it with **3 args**:

```java
protected SinkWriter<InputT> createWriter(
        MailboxExecutor mailboxExecutor, SinkWriterMetricGroup metricGroup, int subtaskIndex);
```

Because the signatures don't match, `FlinkSink` never truly implements the abstract method → `AbstractMethodError` at runtime.

For reference, `fluss-flink-2.2/SinkAdapter` already has the 3-arg signature, so this looks like a copy-paste omission in the 2.3 adapter.

### Fix

Align `SinkAdapter` in `fluss-flink-2.3` with `fluss-flink-2.2` by propagating `subtaskIndex` from `WriterInitContext.getTaskInfo().getIndexOfThisSubtask()`.

```java
@Override
public SinkWriter<InputT> createWriter(WriterInitContext writerInitContext) throws IOException {
    return createWriter(
            writerInitContext.getMailboxExecutor(),
            writerInitContext.metricGroup(),
            writerInitContext.getTaskInfo().getIndexOfThisSubtask());
}

protected abstract SinkWriter<InputT> createWriter(
        MailboxExecutor mailboxExecutor,
        SinkWriterMetricGroup metricGroup,
        int subtaskIndex);
```

### Verification

With this fix applied on top of the current PR head (`24a219484`), I ran three end-to-end SQL demos on Fluss (K8s, 1 coordinator + 3 tablet-servers) + Flink 2.3 standalone:

| Demo | Description | Result |
|------|-------------|--------|
| 1 | Lookup Join: `datagen` events LEFT JOIN Fluss pk-table dim | ✅ 20/20 rows correctly enriched |
| 2 | Delta Join: two Fluss pk-tables INNER JOIN | ✅ Planner emits `DeltaJoin` operator; 5/5 rows correct, no join state |
| 3 | Cascaded Delta Join: three-level INNER JOIN | ✅ Planner emits **two** `DeltaJoin` operators; 5/5 rows correct |

Job graph excerpt (Demo 3):

```
Source: FlussSource-sales -> DropUpdateBefore
Source: FlussSource-sale_items -> DropUpdateBefore
Source: FlussSource-sale_item_addons -> DropUpdateBefore
DeltaJoin[7] -> Calc[8]
DeltaJoin[13] -> Calc[14] -> ConstraintEnforcer[15]
Sink(sale_item_addons_enriched): Writer
```

### Patch

- **Branch**: [`fightBoxing/fluss:fix-sink-adapter-flink-2.3`](https://github.com/fightBoxing/fluss/tree/fix-sink-adapter-flink-2.3)
- **Commit**: `79ee7752e`
- Single-file change, only 6 insertions / 2 deletions.

Happy to open this as a follow-up PR against your `support-flink-2.3` branch if you'd like — just let me know your preference. 🙏

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/utils/FlinkConversions.java
```diff
@@ -643,6 +658,8 @@ private static CatalogMaterializedTable toFlinkMaterializedTable(
             Map<String, String> options) {
         // Validate required materialized table options first
         String definitionQuery = options.get(MATERIALIZED_TABLE_DEFINITION_QUERY.key());
+        String originalQuery = options.get(MATERIALIZED_TABLE_ORIGINAL_QUERY.key());
```
fresh-borzoni, line 661: Tables created before this PR only have `definition-query`, so both of these come back null and Flink 2.3 rejects them with checkNotNull therefore reading any existing materialized table fails with `NullPointerException: Original query must not be null.` 
Should we fall back to definitionQuery when the new keys are absent?

fluss-flink/fluss-flink-common/src/test/java/org/apache/fluss/flink/utils/FlinkConversionsTest.java
```diff
@@ -529,6 +537,142 @@ void testAggregationFunctionRoundTrip() {
...
+     * expanded-query from their own materialized-table option keys. Prior to adapting to Flink 2.3
+     * the deserialization re-used {@code definitionQuery} for both the original and expanded query.
+     * The test asserts the three queries are read independently and stripped from the resulting
+     * flink table's {@code getOptions()} (consumed by the materialized-table prefix).
+     *
+     * <p>This test runs against the {@code fluss-flink-common} test classpath, where the {@code
+     * CatalogMaterializedTableAdapter} for Flink 1.20/2.2 treats original/expanded as no-ops, so
+     * the assertion deliberately targets only what the common adapter exposes: the consumed keys
+     * are absent from {@code getOptions()} and the {@code definitionQuery} matches the input.
+     * Verifying distinct original/expanded values survive a roundtrip is left to the Flink 2.3
```
fresh-borzoni: This says the roundtrip is covered in the 2.3 module, but I couldn't find that test there.
Could we add it, plus the case where neither key is present?

fluss-flink/fluss-flink-2.3/src/test/java/org/apache/fluss/flink/source/Flink23DeltaJoinITCase.java
```diff
@@ -0,0 +1,1092 @@
...
+                String.format(
+                        "INSERT INTO %s SELECT a1, c1, a2, c2, a3, c3 "
+                                + "FROM %s "
+                                + "INNER JOIN %s ON c1 = c2 AND d1 = d2 "
+                                + "INNER JOIN %s ON c1 = c3 AND d1 = d3",
+                        sinkTableName, table1, table2, table3);
+
+        assertThatThrownBy(() -> tEnv.explainSql(sql))
+                .isInstanceOf(ValidationException.class)
+                .hasMessageContaining("doesn't support to do delta join optimization");
```
fresh-borzoni: Cascaded delta join works on 2.3. The same 3-way join with the sink PK aligned to the join key plans as nested DeltaJoin for me. 
This test still throws only because its sink PK (c1, c2, c3) forces upsertMaterialize=[true]. 
Should we invert it?

website/docs/engine-flink/getting-started.md
```diff
@@ -36,7 +37,7 @@ Verify downloaded JARs using the [verification instructions](/downloads#verifyin
 ## Supported Flink Versions
 | Fluss Connector Versions | Supported Flink Versions |
 |--------------------------|--------------------------| 
-| $FLUSS_VERSION_SHORT$    | 1.18, 1.19, 1.20         |
+| $FLUSS_VERSION_SHORT$    | 1.18, 1.19, 1.20, 2.2, 2.3 |
```
fresh-borzoni, line 40: TY, 2.2 was missing here. `delta-joins.mdx` still lists cascade joins as a 2.2 limitation and multi-way joins under "Future Plan", worth a Flink 2.3 section. 
And fluss-flink-common/README.md's version table doesn't list 2.3.
WDYT?

fluss-flink/fluss-flink-2.3/src/test/java/org/apache/fluss/flink/catalog/Flink23MaterializedTableITCase.java
```diff
@@ -0,0 +1,70 @@
...
+     * partial-update handling in {@code FlinkTableSink#getSinkRuntimeProvider} runs, leaving the
+     * refresh job in FAILED state. {@code waitUntilAllTasksAreRunning} (called from {@link
+     * MaterializedTableITCase#testCreateMaterializedTableInContinuousMode()}) then hangs until the
+     * JUnit timeout because Flink 2.3's {@code CommonTestUtils.waitUntilCondition(supplier)} has no
+     * overall timeout — it only retries with a 100 ms sleep between polls.
+     *
+     * <p>Restore the pre-2.3 behaviour by disabling the option in the SQL Gateway session so the
+     * refresh job can be scheduled and reach RUNNING.
+     */
+    @BeforeAll
```
fresh-borzoni: I removed this whole @BeforeAll and reverted the protected widening in the common class and 2/2 pass on 2.3. It also hides the parent's static setUp rather than extending it, so it will drift. 

Can we drop it, or was there a failure that prompted it?

fluss-flink/fluss-flink-2.3/src/test/java/org/apache/fluss/flink/source/Flink23DeltaJoinITCase.java
```diff
@@ -0,0 +1,1092 @@
...
+                .set(
+                        OptimizerConfigOptions.TABLE_OPTIMIZER_DELTA_JOIN_STRATEGY,
+                        OptimizerConfigOptions.DeltaJoinStrategy.FORCE);
+        // Flink 2.3 introduces ExecutionConfigOptions.TABLE_EXEC_SINK_REQUIRE_ON_CONFLICT
+        // (default true), which makes FlinkChangelogModeInferenceProgram throw a
+        // "upsert key differs from primary key" ValidationException before the
+        // StreamPhysicalDeltaJoinForceValidator runs. Disable it here so the existing
+        // delta-join "doesn't support to do delta join optimization" error remains
+        // reachable from these tests.
+        tEnv.getConfig().set(ExecutionConfigOptions.TABLE_EXEC_SINK_REQUIRE_ON_CONFLICT, false);
```
fresh-borzoni, line 96: Only the negative tests need this as they assert on an error message that Flink 2.3 changed. 
The positive ones pass without it. Turning the validation off for the whole class to keep those assertions green also hides it from every other test here. 
Could we update the assertions instead?

fresh-borzoni (commented):
@sd4324530 Thank you for the PR, looks good overall, left a couple of comments, PTAL

fresh-borzoni:
> @sd4324530 agreed 😄 @fresh-borzoni any thoughts since you are a heavy user? 😄
@polyzos it's real and fairly common, but it isn't a Fluss thing.

Partial upsert and delta join both plan fine on 2.3, I checked. What throws is re-keying
and regular joins, where the sink PK doesn't cover the query's upsert key.

That's FLIP-558 at the planner level though as it applies to every upsert sink on 2.3, not just Fluss. Our sink is a plain upsert sink and there's no connector hook to opt out, so there's nothing for us to fix. 
Users should add `ON CONFLICT DO DEDUPLICATE` or set `table.exec.sink.require-on-conflict=false`.

So I'd just document it in the docs.

fresh-borzoni:
If I understand this correctly, the only difference between 2.2 module is the materialized table adapters, the rest is identical
Mb it would be a good followup to factor out the generic part? 

cc @loserwang1024 as I saw there was some discussion about 2.1 and 2.2 and it ended up with dropping 2.1, so there might be some context that I miss :)

sd4324530:
> ## 🐛 Bug fix: `AbstractMethodError` in `SinkAdapter`
> Hi @sd4324530 (and reviewers), thanks for the great work on Flink 2.3 support!
> 
> While testing this PR against a Fluss (K8s) + Flink 2.3 standalone cluster, every Flink job that writes to a Fluss table fails at runtime with:
> 
> ```
> java.lang.AbstractMethodError: Receiver class org.apache.fluss.flink.sink.FlinkSink
> does not define or inherit an implementation of the resolved method
> 'createWriter(MailboxExecutor, SinkWriterMetricGroup)' of abstract class
> org.apache.fluss.flink.adapter.SinkAdapter.
>     at org.apache.fluss.flink.adapter.SinkAdapter.createWriter(SinkAdapter.java:37)
>     at org.apache.flink.streaming.runtime.operators.sink.StatelessSinkWriterStateHandler.createWriter(...)
>     ...
> ```
> 
> ### Root cause
> The abstract `createWriter` in `fluss-flink-2.3/SinkAdapter` has signature:
> 
> ```java
> protected abstract SinkWriter<InputT> createWriter(
>         MailboxExecutor mailboxExecutor, SinkWriterMetricGroup metricGroup);   // 2 args
> ```
> 
> but `fluss-flink-common/FlinkSink` overrides it with **3 args**:
> 
> ```java
> protected SinkWriter<InputT> createWriter(
>         MailboxExecutor mailboxExecutor, SinkWriterMetricGroup metricGroup, int subtaskIndex);
> ```
> 
> Because the signatures don't match, `FlinkSink` never truly implements the abstract method → `AbstractMethodError` at runtime.
> 
> For reference, `fluss-flink-2.2/SinkAdapter` already has the 3-arg signature, so this looks like a copy-paste omission in the 2.3 adapter.
> 
> ### Fix
> Align `SinkAdapter` in `fluss-flink-2.3` with `fluss-flink-2.2` by propagating `subtaskIndex` from `WriterInitContext.getTaskInfo().getIndexOfThisSubtask()`.
> 
> ```java
> @Override
> public SinkWriter<InputT> createWriter(WriterInitContext writerInitContext) throws IOException {
>     return createWriter(
>             writerInitContext.getMailboxExecutor(),
>             writerInitContext.metricGroup(),
>             writerInitContext.getTaskInfo().getIndexOfThisSubtask());
> }
> 
> protected abstract SinkWriter<InputT> createWriter(
>         MailboxExecutor mailboxExecutor,
>         SinkWriterMetricGroup metricGroup,
>         int subtaskIndex);
> ```
> 
> ### Verification
> With this fix applied on top of the current PR head (`24a219484`), I ran three end-to-end SQL demos on Fluss (K8s, 1 coordinator + 3 tablet-servers) + Flink 2.3 standalone:
> 
> Demo	Description	Result
> 1	Lookup Join: `datagen` events LEFT JOIN Fluss pk-table dim	✅ 20/20 rows correctly enriched
> 2	Delta Join: two Fluss pk-tables INNER JOIN	✅ Planner emits `DeltaJoin` operator; 5/5 rows correct, no join state
> 3	Cascaded Delta Join: three-level INNER JOIN	✅ Planner emits **two** `DeltaJoin` operators; 5/5 rows correct
> Job graph excerpt (Demo 3):
> 
> ```
> Source: FlussSource-sales -> DropUpdateBefore
> Source: FlussSource-sale_items -> DropUpdateBefore
> Source: FlussSource-sale_item_addons -> DropUpdateBefore
> DeltaJoin[7] -> Calc[8]
> DeltaJoin[13] -> Calc[14] -> ConstraintEnforcer[15]
> Sink(sale_item_addons_enriched): Writer
> ```
> 
> ### Patch
> * **Branch**: [`fightBoxing/fluss:fix-sink-adapter-flink-2.3`](https://github.com/fightBoxing/fluss/tree/fix-sink-adapter-flink-2.3)
> * **Commit**: `79ee7752e`
> * Single-file change, only 6 insertions / 2 deletions.
> 
> Happy to open this as a follow-up PR against your `support-flink-2.3` branch if you'd like — just let me know your preference. 🙏

@fightBoxing Thank you for your review!
This issue has been fixed in https://github.com/apache/fluss/commit/1c5bb53d8c408f5380357b89d10de4bd1b0a401e
Please verify it again.

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/utils/FlinkConversions.java
sd4324530: Hi @fresh-borzoni , I've read the design document for this feature([FLIP-546](https://cwiki.apache.org/confluence/spaces/FLINK/pages/383191507/FLIP-546+Introduce+CREATE+OR+ALTER+for+Materialized+Tables#FLIP546%3AIntroduceCREATEORALTERforMaterializedTables-5.Compatibility%2CDeprecation%2CandMigrationPlan)), and it should be compatible with existing materialized tables?

fresh-borzoni: @sd4324530 
The case that's missing is upgrade, not new tables.
Someone creates a materialized table on Fluss today. Later they move to Flink 2.3 with this connector. That table was written before original-query and expanded-query existed, so it only has definition-query and every read of it now fails with NullPointerException: Original query must not be null.

Tables created after this PR are fine, since we write all three. It's the ones already out there that break. Try it :)

website/docs/engine-flink/getting-started.md
sd4324530: done:https://github.com/apache/fluss/pull/3521/changes/491ae5b9f944e7b5ebf44680de86f400971057cc

fluss-flink/fluss-flink-2.3/src/test/java/org/apache/fluss/flink/catalog/Flink23MaterializedTableITCase.java
sd4324530: done:https://github.com/apache/fluss/pull/3521/changes/e7bb9b13fb8bc180a262b7a50b1985c4de7d4f68

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/utils/FlinkConversions.java
sd4324530: done:https://github.com/apache/fluss/pull/3521/changes/3fa056c90cb4484ac5009228c9e4a3ee857c5c06

fluss-flink/fluss-flink-2.3/src/test/java/org/apache/fluss/flink/source/Flink23DeltaJoinITCase.java
sd4324530: done:https://github.com/apache/fluss/pull/3521/changes/7b9f78944436a3122354bf013c33efeebd19d80a

fluss-flink/fluss-flink-common/src/test/java/org/apache/fluss/flink/utils/FlinkConversionsTest.java
fresh-borzoni: TY for the legacy case. The other half of @leonardBang's ask: a roundtrip where original and
expanded differ is still only in the common test, where the adapter no-ops both. 
Worth adding to `Flink23CatalogTest` too?

fluss-flink/fluss-flink-2.3/src/test/java/org/apache/fluss/flink/catalog/Flink23MaterializedTableITCase.java
fresh-borzoni: TY. The protected widening on FILE_CATALOG_STORE  and  defaultSessionEnvironment in the
common class is unused now, worth reverting too?

fluss-flink/fluss-flink-2.3/src/test/java/org/apache/fluss/flink/sink/Flink23TableSinkITCase.java
```diff
@@ -0,0 +1,39 @@
...
+    /**
+     * Flink 2.3 introduces {@link ExecutionConfigOptions#TABLE_EXEC_SINK_REQUIRE_ON_CONFLICT}
+     * (default {@code true}). In {@code FlinkChangelogModeInferenceProgram} this triggers a "upsert
+     * key differs from primary key" ValidationException for partial upserts such as {@code INSERT
+     * INTO pk_table(a, b) VALUES ...} where the query carries no upsert key. The parent class
+     * {@link FlinkTableSinkITCase#testPartialUpsert()} (and its siblings) rely on Fluss's own
+     * partial-update handling at the sink layer, so disable the option here to keep those tests
+     * reachable.
+     */
+    @BeforeEach
```
fresh-borzoni: This one is still class-wide. 
Could it get the same treatment as the delta join class? Of the 49 tests here only testPartialUpsert and testIgnoreDelete[isPrimaryKeyTable=true] need it.

The javadoc is also slightly off: `INSERT INTO pk_table(a, b) VALUES ...` plans fine on 2.3, it's the fromChangelogStream phase that throws, because that source has no declared key.

fluss-flink/fluss-flink-2.3/src/test/java/org/apache/fluss/flink/source/Flink23DeltaJoinITCase.java
fresh-borzoni: I checked the three claims added to delta-joins.mdx against 2.3. 
With the sink PK covering the join's upsert key and require-on-conflict at its default, cascaded delta join, RAND() in the projection before the sink, and a lookup join after the delta join all plan as DeltaJoin.

So the docs are right and two tests are stale: this one and testDeltaJoinFailsWithNonDeterministicFunctions. Both pass only because their sink PK doesn't
cover the query's upsert key: (c1) vs (c1, d1) in the NDF one, so they'd throw with or without the condition they name. Keying those sinks to cover the upsert key makes them isolate what they claim, and then they don't need disableSinkRequireOnConflict()  either.

Alos worth adding a lookup-join-after-delta-join test too, since that claim has no coverage so far.

fresh-borzoni (commented):
@sd4324530 Thank you, only minor comments left, PTAL 👍

fluss-flink/fluss-flink-2.3/src/test/java/org/apache/fluss/flink/sink/Flink23TableSinkITCase.java
sd4324530: Done:https://github.com/apache/fluss/pull/3521/changes/0efb5c37d1b38a774d74c527f3eafd34721faac6

fluss-flink/fluss-flink-2.3/src/test/java/org/apache/fluss/flink/catalog/Flink23MaterializedTableITCase.java
sd4324530: Thanks.
Done:https://github.com/apache/fluss/pull/3521/changes/c22d01c7f2d415cecd05adbe44ca9ef45249198c

sd4324530:
@fresh-borzoni 
Hi, I'm having trouble with the "lookup-join-after-delta-join" test I added.I hit a planner-level limitation in Flink 2.3. Could you please take a look? The code was written with AI assistance.

https://github.com/apache/fluss/pull/3521/changes/0af874bca914b099a567bf74e5ecfc49b6c39237

fluss-flink/fluss-flink-2.3/src/test/java/org/apache/fluss/flink/source/Flink23DeltaJoinITCase.java
```diff
@@ -0,0 +1,1100 @@
...
+import static org.apache.fluss.flink.FlinkConnectorOptions.BOOTSTRAP_SERVERS;
+import static org.apache.fluss.flink.source.testutils.FlinkRowAssertionsUtils.assertResultsIgnoreOrder;
+import static org.apache.fluss.server.testutils.FlussClusterExtension.BUILTIN_DATABASE;
+import static org.apache.fluss.testutils.DataTestUtils.row;
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
+
+/** IT case for Delta Join optimization in Flink 2.3. */
+@MultiVersionTest
+public class Flink23DeltaJoinITCase extends FlinkTestBase {
```
sd4324530, line 58: @fresh-borzoni 
Hi, I noticed that most of the tests in this class are consistent with Flink 2.2.
Is it possible to remove test methods that are completely identical to those in Flink 2.2, keeping only the parts that need to be tested independently in Flink 2.3, such as: `testDeltaJoinWithJoinKeyExceedsPrimaryKey`, `testDeltaJoinFailsWhenJoinKeyNotContainIndex`, `testDeltaJoinFailsWithOuterJoin`, and `testDeltaJoinFailsWithSinkMaterializer`?
This speeds up testing and reduces redundant testing, as the @MultiVersionTest annotation aims to achieve.

fluss-flink/fluss-flink-common/src/test/java/org/apache/fluss/flink/utils/FlinkConversionsTest.java
sd4324530: Done: https://github.com/apache/fluss/pull/3521/changes/863c337ec97736016738ec0723a66be78322285e

fresh-borzoni (approved):
@sd4324530 Thank you, I hope you don't mind my help here

On the lookup join: it's not a planner limitation. A prefix join is 1:N, so its output is
both sides' PKs  and the sink keyed `(c1, d1)` didn't cover it, leaving no upsert key, hence the `ON CONFLICT DO DEDUPLICATE`, hence a SinkUpsertMaterializer, hence no delta join. Your query without the lookup join gives `DeltaJoin=false`. Keying the sink `(c1, d1, c2, d2)` like `testDeltaJoin` does makes your original join work with FORCE and no `ON CONFLICT`.

The commit also adds back cascade and non-deterministic-function tests as positive ones (both work on 2.3), reads back the materialized table roundtrip, and adds @leonardBang's SinkAdapter regression test.

PTAL

@leonardBang @loserwang1024 @polyzos Please take a look, if you have other comments or spot anything, appreciate it 🙏

fluss-flink/fluss-flink-2.3/pom.xml
```diff
@@ -0,0 +1,294 @@
...
+        <groupId>org.apache.fluss</groupId>
+        <artifactId>fluss-flink</artifactId>
+        <version>1.0-SNAPSHOT</version>
+    </parent>
+
+    <artifactId>fluss-flink-2.3</artifactId>
+    <name>Apache Fluss : Engine Flink : 2.3 </name>
+    <properties>
+        <flink.major.version>2.3</flink.major.version>
+        <flink.minor.version>2.3.0</flink.minor.version>
```
leonardBang, line 33: The officially published Flink 2.3 Maven artifacts target Java 17 bytecode, while the Fluss CI reactor and documented build environment still use JDK 11. Consequently, the standard build will fail with `class file has wrong version 61.0, should be 55.0`. Although Flink source provides a Java 11 target profile, there is no separate official Java 11-compatible Maven artifact. Could we build this module with a JDK 17 toolchain/job while keeping the generated Fluss classes Java 11-compatible, and document the additional build requirement?

leonardBang (commented):
Thanks @sd4324530 for the update and @fresh-borzoni for the polish, the PR change looks good to me, I only left one comment about the JDK version.

fluss-flink/fluss-flink-2.3/pom.xml
leonardBang: -rw-r--r--   1 bangjiangxu  staff  1888920 Aug 25 20:25 flink-core-2.3.0.jar
drwxr-xr-x   3 bangjiangxu  staff       96 Jun 14 22:43 org
(base) bangjiangxu test~  $file org/apache/flink/configuration/YamlParserUtils.class
org/apache/flink/configuration/YamlParserUtils.class: compiled Java class data, version 55.0 (Java SE 11)

I double-checked the jar version is JDK11, and JDK 17 is the recommend runtime version from flink community docs. Thus current JDK 11 CI is okay.

leonardBang (approved):
+1 from my side.

fluss-flink/fluss-flink-2.3/pom.xml
fresh-borzoni: @leonardBang Are you sure this applies? Flink 2.3.0's published artifacts are Java 11 bytecode.

CI kinda is a check now:  Java 11 / flink2 includes fluss-flink-2.3 and has passed on every head of this PR. Version 61 class files aren't readable by JDK 11's javac at all, so it would have failed on the first run rather than getting this far.

Do you have a specific artifact in mind? Happy to check it if so.I

I understand it'll matter once Flink stops targeting Java 11 in its artifacts, but that hasn't happened yet, so feels a bit premature, wdyt?

leonardBang: @fresh-borzoni Sorry for the misunderstand, I've updated in above comment and resolved this comment.

sd4324530:
> @sd4324530 Thank you, I hope you don't mind my help here
> 
> On the lookup join: it's not a planner limitation. A prefix join is 1:N, so its output is both sides' PKs and the sink keyed `(c1, d1)` didn't cover it, leaving no upsert key, hence the `ON CONFLICT DO DEDUPLICATE`, hence a SinkUpsertMaterializer, hence no delta join. Your query without the lookup join gives `DeltaJoin=false`. Keying the sink `(c1, d1, c2, d2)` like `testDeltaJoin` does makes your original join work with FORCE and no `ON CONFLICT`.
> 
> The commit also adds back cascade and non-deterministic-function tests as positive ones (both work on 2.3), reads back the materialized table roundtrip, and adds @leonardBang's SinkAdapter regression test.
> 
> PTAL
> 
> @leonardBang @loserwang1024 @polyzos Please take a look, if you have other comments or spot anything, appreciate it 🙏

@fresh-borzoni Thank you very much for your help.

PR #3526, opened 2026-06-25
[client] Reject client-side S3 credential options

### LINKED ISSUE #3524: [client][filesystem] Reject client-side S3 credential options
### Search before asking

### Fluss version

1.0-snapshot

### Please describe the bug

PR #938 allows Fluss clients to forward `client.fs.*` options to the client-side filesystem initialization.

For S3, users should not configure credential options through `client.fs.*`. Credentials, AssumeRole, and STS token generation are expected to be configured on Fluss servers. Fluss clients should use server-issued temporary filesystem tokens through `DynamicTemporaryAWSCredentialsProvider`.

If a user mistakenly configures credential-bearing options such as:

```text
client.fs.s3a.access.key
client.fs.s3a.secret.key
client.fs.s3a.session.token
client.fs.s3a.assumed.role.*
client.fs.s3a.aws.credentials.provider
```

`FlussConnection` strips the `client.fs.` prefix and passes them to `FileSystem.initialize`. The S3 filesystem plugin then sees static credentials or AssumeRole settings and skips the delegated-token setup path, so the client-side S3 filesystem is not configured with `DynamicTemporaryAWSCredentialsProvider`.

This is a user misconfiguration, but failing fast would make the problem clearer and avoid silently changing the credential provider path.

Expected behavior:

- Reject S3 credential-bearing options under `client.fs.*`.
- Continue to allow non-credential runtime options such as:

```text
client.fs.s3a.endpoint
client.fs.s3a.region
client.fs.s3a.path-style-access
client.fs.s3a.path.style.access
```

### Solution

Add validation before forwarding `client.fs.*` options to `FileSystem.initialize`, and fail fast for S3 credential-bearing keys.

litiliu:
### Purpose

Linked issue: close #3524

This PR adds fail-fast validation for S3 credential-bearing options under `client.fs.*`.

Users should not configure S3 credentials through `client.fs.*`. If they mistakenly set options such as `client.fs.s3a.access.key`, `client.fs.s3a.secret.key`, `client.fs.s3a.session.token`, `client.fs.s3a.assumed.role.*`, or `client.fs.s3a.aws.credentials.provider`, those options are stripped to S3 filesystem keys during client-side filesystem initialization. The S3 plugin can then skip the delegated-token setup path, so the client filesystem is not configured with `DynamicTemporaryAWSCredentialsProvider`.

Failing fast makes this misconfiguration explicit. Non-credential runtime options, such as endpoint, region, and path-style access, are still allowed.

### Brief change log

- Validate `client.fs.*` before forwarding it to `FileSystem.initialize`.
- Reject S3 credential-bearing options under `client.fs.*`.
- Keep non-credential S3 filesystem options forwarded as before.
- Add unit tests for allowed and rejected options.

### Tests

- `mvn -pl fluss-client -Dtest=FlussConnectionTest test` passed before rebasing the branch.
- After rebasing onto `apache/main`, the same command failed during local `fluss-client` compilation due to unrelated generated RPC/client source mismatches in the local checkout, before running this test.

### API and Format

### Documentation

No documentation changes.

### Generative AI disclosure

- [x] Yes (OpenAI Codex)

### DIFF
--- a/fluss-client/src/main/java/org/apache/fluss/client/FlussConnection.java
+++ b/fluss-client/src/main/java/org/apache/fluss/client/FlussConnection.java
@@ -17,6 +17,7 @@
 
 package org.apache.fluss.client;
 
+import org.apache.fluss.annotation.VisibleForTesting;
 import org.apache.fluss.client.admin.Admin;
 import org.apache.fluss.client.admin.FlussAdmin;
 import org.apache.fluss.client.lookup.LookupClient;
@@ -33,7 +34,9 @@ import org.apache.fluss.cluster.ServerNode;
 import org.apache.fluss.config.ConfigOptions;
 import org.apache.fluss.config.Configuration;
 import org.apache.fluss.exception.FlussRuntimeException;
+import org.apache.fluss.exception.IllegalConfigurationException;
 import org.apache.fluss.fs.FileSystem;
+import org.apache.fluss.fs.S3FileSystemConfigUtils;
 import org.apache.fluss.metadata.TablePath;
 import org.apache.fluss.metrics.registry.MetricRegistry;
 import org.apache.fluss.rpc.GatewayClientProxy;
@@ -45,6 +48,7 @@ import java.time.Duration;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
+import java.util.Map;
 
 import static org.apache.fluss.client.utils.MetadataUtils.getOneAvailableTabletServerNode;
 import static org.apache.fluss.config.FlussConfigUtils.CLIENT_PREFIX;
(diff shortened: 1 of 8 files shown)


litiliu:
Closing this PR after re-evaluating the scenario.

The issue this PR handles is mainly a user misconfiguration case: users should not configure S3 credentials such as AK/SK, session token, AssumeRole, or credential provider settings through `client.fs.*`. If they do, the client-side filesystem can choose a different credential provider path, but that is caused by the invalid configuration itself rather than a normal Fluss client flow.

Given that, adding extra validation here is not clearly necessary. We can revisit if this becomes a real user-facing problem.

PR #3529, opened 2026-06-25
[client] Avoid per-record AutoPartitionStrategy construction in WriterClient.doSend in log tables

### LINKED ISSUE #3527: [client] WriterClient.doSend rebuilds AutoPartitionStrategy per record on non partitioned tables
### Search before asking

### Fluss version

main (development)

### Please describe the bug

A Flink job sinking to a non-partitioned Fluss append-only log table
backpressures the upstream pipeline. Profiling the sink subtask shows that
roughly half of its CPU is spent on auto-partition strategy construction for a
table that isn't partitioned in the first place.

#### Table

```sql
CREATE TABLE `fluss_catalog`.`cdstream_db`.`cluster_combinations`
  (`combinationId`       VARCHAR(2147483647),
   `leftClusters`        ARRAY<INT>,
   `rightClusters`       ARRAY<INT>,
   `level`               INT,
   `size`                BIGINT,
   `decisive`            BOOLEAN,
   `positive`            BOOLEAN,
   `bounded`             BOOLEAN,
   `singleton`           BOOLEAN,
   `parentId`            VARCHAR(2147483647),
   `boundingTimestamp`   TIMESTAMP(3),
   `discoveredTimestamp` TIMESTAMP(3),
   `lowerBound` DOUBLE,
   `upperBound` DOUBLE) WITH (
                          'bucket.num' = '64',
                          'table.log.format' = 'arrow',
                          'table.log.ttl' = '1h',
                          'table.replication.factor' = '1',
                          'table.datalake.enabled' = 'false'
                          );
```

#### Writer sink flamegraph

Image

| Code path                                                                       | Samples |    % |
|---------------------------------------------------------------------------------|--------:|-----:|
| `java.util.TimeZone.getTimeZone(String)`                                        |     331 | 41.4 |
| `AutoPartitionStrategy.from` + `TableConfig.getAutoPartitionStrategy` around it |      71 |  8.9 |
| Arrow row encoding                                                              |     136 | 17.0 |
| User serializer                                                                 |      93 | 11.6 |
| ZSTD compression in batch close                                                 |      49 |  6.1 |
| Everything else (sink, bucket assignment, RPC, IO)                              |     <50 |   <6 |

About 50% of the writer thread's CPU is in `AutoPartitionStrategy.from(...)`,
dominated by the JDK's `TimeZone.getTimeZone(String)`. That JDK method walks a
global static registry under a synchronized block and clones the returned object
every call, so it is genuinely slow to invoke at high frequency.

#### Call chain

```
WriterClient.doSend                                                          (line 182)
  dynamicPartitionCreator.checkAndCreatePartitionAsync(
      physicalTablePath,
      tableInfo.getPartitionKeys(),
      tableInfo.getTableConfig().getAutoPartitionStrategy())                 // per record
    TableConfig.getAutoPartitionStrategy                                     (line 161)
      return AutoPartitionStrategy.from(config);                             // new object each call
        AutoPartitionStrategy.from(Configuration)                            (line 57)
          TimeZone.getTimeZone(                                              (line 64)
              conf.getString(TABLE_AUTO_PARTITION_TIMEZONE))                 // 41% of CPU
```

#### Source

`fluss-client/src/main/java/org/apache/fluss/client/write/WriterClient.java:179-182` —
called once per record:

```java
dynamicPartitionCreator.checkAndCreatePartitionAsync(
        physicalTablePath,
        tableInfo.getPartitionKeys(),
        tableInfo.

getTableConfig().

getAutoPartitionStrategy());
```

`fluss-common/src/main/java/org/apache/fluss/config/TableConfig.java:159-162` —
uncached, builds a fresh strategy every call:

```java
/** Gets the auto partition strategy of the table. */
public AutoPartitionStrategy getAutoPartitionStrategy() {
    return AutoPartitionStrategy.from(config);
}
```

`fluss-common/src/main/java/org/apache/fluss/utils/AutoPartitionStrategy.java:57-65`:

```java
public static AutoPartitionStrategy from(Configuration conf) {
    return new AutoPartitionStrategy(
            conf.getBoolean(ConfigOptions.TABLE_AUTO_PARTITION_ENABLED),
            conf.getString(ConfigOptions.TABLE_AUTO_PARTITION_KEY),
            conf.get(ConfigOptions.TABLE_AUTO_PARTITION_TIME_UNIT),
            conf.getInt(ConfigOptions.TABLE_AUTO_PARTITION_NUM_PRECREATE),
            conf.getInt(ConfigOptions.TABLE_AUTO_PARTITION_NUM_RETENTION),
            TimeZone.getTimeZone(conf.getString(ConfigOptions.TABLE_AUTO_PARTITION_TIMEZONE)));
}
```

The callee, `DynamicPartitionCreator.checkAndCreatePartitionAsync` (
`fluss-client/src/main/java/org/apache/fluss/client/write/DynamicPartitionCreator.java:68-113`),
discards the `autoPartitionStrategy` argument for non-partitioned tables and
only ever reads it on first write to a partition for partitioned tables:

```java
public void checkAndCreatePartitionAsync(
        PhysicalTablePath physicalTablePath,
        List<String> partitionKeys,
        AutoPartitionStrategy autoPartitionStrategy) {
    String partitionName = physicalTablePath.getPartitionName();
    if (partitionName == null) {
        // no need to check and create partition
        return;                                                  // argument never read
    }
    Optional<Long> partitionIdOpt
            = metadataUpdater.getPartitionId(physicalTablePath);
    if (!partitionIdOpt.isPresent()) {
        if (inflightPartitionsToCreate.contains(physicalTablePath)) {
            ...
        } else if (forceCheckPartitionExist(physicalTablePath)) {
            ...
        } else {
            ResolvedPartitionSpec resolvedPartitionSpec =
                    ResolvedPartitionSpec.fromPartitionName(partitionKeys, partitionName);
            validateAutoPartitionTime(
                    resolvedPartitionSpec.toPartitionSpec(),
                    partitionKeys,
                    autoPartitionStrategy);                      // the only reader
            ...
        }
    }
}
```

So in steady state:

- Non-partitioned table: strategy used 0 times out of N records.
- Partitioned table: strategy used at most P times out of N records, where P is
  the number of distinct partitions ever written. For long-running streaming
  jobs P/N is negligible.

The strategy is still constructed N times out of N because Java evaluates method
arguments before the call. Pure waste for non-partitioned tables, and roughly
N/P times too often for partitioned ones. Fixing it does not change behaviour;
it only saves CPU.

#### When this regressed

The `TimeZone.getTimeZone(String)` call has lived inside
`AutoPartitionStrategy.from(...)` since the initial commit (`a1280c68`,
2024-11-25). At that point nothing called it on the hot path.

The regression came in with #1002, "Support dynamically create partition when
writing" (`abb3b86a`, 2025-06-05). That PR added the
`dynamicPartitionCreator.checkAndCreatePartitionAsync(..., tableInfo.getTableConfig().getAutoPartitionStrategy())`
call inside `WriterClient.doSend` and so promoted `getAutoPartitionStrategy()`
from a metadata-path call into an O(records) hot-path call.

#### How to reproduce

Any Flink job that writes at a non-trivial record rate (we saw it at ~2.5M
records/s, 16-way parallel writer) into a Fluss table without auto partition enabled. The flamegraph above came from async-profiler attached
to the Flink TaskManager, filtered to the writer subtask thread.

binary-signal:
### Purpose

Linked issue: close #3527

This PR removes the per-record cost which improves write throughput for fluss log tables.

### Brief change log

Two patches:

1. Skip the dynamic-partition check when the table has no partition keys (
   `fluss-client/.../WriterClient.java`):

   ```java
   if (!tableInfo.getPartitionKeys().isEmpty()) {
       dynamicPartitionCreator.checkAndCreatePartitionAsync(
               physicalTablePath,
               tableInfo.getPartitionKeys(),
               tableInfo.getTableConfig().getAutoPartitionStrategy());
   }
   ```

   `DynamicPartitionCreator.checkAndCreatePartitionAsync` already returns
   immediately when `partitionName == null`. This just hoists the same check up
   to the caller so the expensive argument never has to be evaluated. No
   behaviour change.

2. Memoise the strategy on `TableConfig` (
   `fluss-common/.../config/TableConfig.java`):

   ```java
   public class TableConfig {
       private final Configuration config;
       private volatile AutoPartitionStrategy autoPartitionStrategy;

       public AutoPartitionStrategy getAutoPartitionStrategy() {
           AutoPartitionStrategy s = autoPartitionStrategy;
           if (s == null) {
               s = AutoPartitionStrategy.from(config);
               autoPartitionStrategy = s; // benign race: same value
           }
           return s;
       }
       ...
   }
   ```

   `TableConfig` is immutable after construction. `AutoPartitionStrategy` is
   too with all fields `final`. A volatile field is enough.

### Tests

No new unit tests. The change is observable only as a perf improvement.

#### Fluss Flink Sink / Taskmanager

| Metric                                         |    Before |     After |  Delta |
|------------------------------------------------|----------:|----------:|-------:|
| Writer records/s, avg (sum across 16 subtasks) | 2,526,516 | 2,826,183 | +11.9% |
| Writer records/s, peak                         | 3,342,630 | 4,421,649 | +32.3% |
| Writer records/s, p95                          | 3,082,199 | 4,198,872 | +36.2% |
| TaskManager JVM CPU load, avg                  |     51.5% |     44.1% |  -7.4% |
| JVM GC time, ms per second, avg                |     1,433 |       999 | -30.3% |
| Writer task busy time, ms/s, avg               |     642.5 |     449.7 | -30.0% |
| Writer task idle time, ms/s, avg               |     357.5 |     550.3 | +53.9% |
| Records per batch, avg                         |    40,971 |    53,531 | +30.7% |
| Bytes per batch, avg                           | 1,267,516 | 1,645,327 | +29.8% |

#### Per-record CPU cost

TaskManager level: `cpu_load / throughput`

- Before: 0.515 / 2,526,516 = 204 ns of CPU per record
- After: 0.441 / 2,826,183 = 156 ns of CPU per record
- Down 23.5%

Writer-task level: `busy_ms_per_s / throughput`

- Before: 642.5 / 2,526,516 = 254 ns per record
- After: 449.7 / 2,826,183 = 159 ns per record
- Down 37.4%

The 37% drop at the writer-task level matches before flamegraph attributed
roughly half the writer CPU to
`TimeZone.getTimeZone` plus the `AutoPartitionStrategy.from` construction around
it. Removing both shaves about that much off the per-record cost. The remaining
work (Arrow encoding, ZSTD, serialisation) didn't change.

#### What got faster

Throughput is up at every percentile. Peak and p95 moved more than the mean,
which fits the picture of the writer no longer being the bottleneck.

CPU is down even though throughput is up. TM doing more records with less
work.

TM GC time dropped. The fix removes the per-record allocation of
`AutoPartitionStrategy` and the throwaway `Configuration` view, which is the
kind of short-lived garbage that drives young-gen churn.

Batches got about 30% bigger by both record count and bytes. With the writer
faster, the batch window naturally accumulates more records before the timeout
fires, so each RPC carries more payload.

#### Fluss Server-side

| Metric                                                            |     Before |      After |                                         Delta |
|-------------------------------------------------------------------|-----------:|-----------:|----------------------------------------------:|
| `fluss_tabletserver_table_messagesInPerSecond`, avg               |  2,526,438 |  2,816,269 |                                        +11.5% |
| `fluss_tabletserver_table_messagesInPerSecond`, peak              |  3,346,781 |  4,464,978 |                                        +33.4% |
| `fluss_tabletserver_table_messagesInPerSecond`, p95               |  3,077,221 |  4,198,341 |                                        +36.4% |
| `fluss_tabletserver_table_bytesInPerSecond`, avg                  |  77.9 MB/s |  86.6 MB/s |                                        +11.2% |
| `fluss_tabletserver_table_bytesInPerSecond`, peak                 | 103.2 MB/s | 138.1 MB/s |                                        +33.8% |
| `fluss_tabletserver_table_bytesOutPerSecond`, avg                 |  99.7 MB/s |  90.0 MB/s |                        -9.7% (see note below) |
| `fluss_tabletserver_table_totalProduceLogRequestsPerSecond`, avg  |      149.1 |      186.7 |                                        +25.2% |
| `fluss_tabletserver_table_totalProduceLogRequestsPerSecond`, peak |      427.5 |      373.2 | -12.7% (fewer peak RPCs at higher throughput) |
| `fluss_tabletserver_table_failedProduceLogRequestsPerSecond`      |          0 |          0 |                                               |

**Flink Task - Fluss Sink Flamegraph Before**
Screenshot 2026-06-25 at 14 21 58

**Flink Task - Fluss Sink Flamegraph After**
No trace of getTimezone
Screenshot 2026-06-24 at 23 24 43

**Flink** 
Note: on left hand side is before and on right hand side is after

Screenshot 2026-06-25 at 16 18 32

Screenshot 2026-06-25 at 16 19 05

Screenshot 2026-06-25 at 16 19 32

Screenshot 2026-06-25 at 16 19 51

Screenshot 2026-06-25 at 16 20 34

Screenshot 2026-06-25 at 16 20 50

**Fluss**
Note: on left hand side is before and on right hand side is after

Screenshot 2026-06-25 at 16 21 49

Screenshot 2026-06-25 at 16 23 46

Screenshot 2026-06-25 at 16 25 11

Screenshot 2026-06-25 at 16 27 04

Screenshot 2026-06-25 at 16 28 22

### API and Format

### Documentation

None needed. No new feature, no user-visible behaviour change.

### DIFF
--- a/fluss-client/src/main/java/org/apache/fluss/client/write/WriterClient.java
+++ b/fluss-client/src/main/java/org/apache/fluss/client/write/WriterClient.java
@@ -176,10 +176,15 @@ public class WriterClient {
 
             TableInfo tableInfo = record.getTableInfo();
             PhysicalTablePath physicalTablePath = record.getPhysicalTablePath();
-            dynamicPartitionCreator.checkAndCreatePartitionAsync(
-                    physicalTablePath,
-                    tableInfo.getPartitionKeys(),
-                    tableInfo.getTableConfig().getAutoPartitionStrategy());
+            // Skip on non-partitioned tables: the callee returns immediately when
+            // partitionName is null, but the expensive AutoPartitionStrategy argument
+            // would still be evaluated per record without this guard.
+            if (!tableInfo.getPartitionKeys().isEmpty()) {
+                dynamicPartitionCreator.checkAndCreatePartitionAsync(
+                        physicalTablePath,
+                        tableInfo.getPartitionKeys(),
+                        tableInfo.getTableConfig().getAutoPartitionStrategy());
+            }
 
             // maybe create bucket assigner.
             Cluster cluster = metadataUpdater.getCluster();
--- a/fluss-client/src/main/java/org/apache/fluss/client/write/WriterClient.java
+++ b/fluss-client/src/main/java/org/apache/fluss/client/write/WriterClient.java
@@ -179,7 +179,7 @@ public class WriterClient {
             // Skip on non-partitioned tables: the callee returns immediately when
             // partitionName is null, but the expensive AutoPartitionStrategy argument
             // would still be evaluated per record without this guard.
-            if (!tableInfo.getPartitionKeys().isEmpty()) {
+            if (tableInfo.isPartitioned()) {
                 dynamicPartitionCreator.checkAndCreatePartitionAsync(
                         physicalTablePath,
                         tableInfo.getPartitionKeys(),
--- a/fluss-client/src/main/java/org/apache/fluss/client/write/WriterClient.java
+++ b/fluss-client/src/main/java/org/apache/fluss/client/write/WriterClient.java
@@ -176,14 +176,9 @@ public class WriterClient {
 
             TableInfo tableInfo = record.getTableInfo();
             PhysicalTablePath physicalTablePath = record.getPhysicalTablePath();
-            // Skip on non-partitioned tables: the callee returns immediately when
-            // partitionName is null, but the expensive AutoPartitionStrategy argument
-            // would still be evaluated per record without this guard.
+            // Skip the call entirely on non-partitioned tables; there is no partition to create.
             if (tableInfo.isPartitioned()) {
-                dynamicPartitionCreator.checkAndCreatePartitionAsync(
-                        physicalTablePath,
-                        tableInfo.getPartitionKeys(),
-                        tableInfo.getTableConfig().getAutoPartitionStrategy());
+                dynamicPartitionCreator.checkAndCreatePartitionAsync(physicalTablePath, tableInfo);
             }
 
             // maybe create bucket assigner.
--- a/fluss-common/src/main/java/org/apache/fluss/config/TableConfig.java
+++ b/fluss-common/src/main/java/org/apache/fluss/config/TableConfig.java
@@ -44,6 +44,11 @@ public class TableConfig {
     // the table properties configuration
     private final Configuration config;
 
+    // Cached, lazily-built strategy. Safe because TableConfig is immutable after construction
+    // and AutoPartitionStrategy itself is fully immutable. A benign race on first call
+    // produces the same value, so a plain volatile is sufficient.
+    private volatile AutoPartitionStrategy autoPartitionStrategy;
+
     /**
      * Creates a new table config.
      *
@@ -158,7 +163,12 @@ public class TableConfig {
 
     /** Gets the auto partition strategy of the table. */
     public AutoPartitionStrategy getAutoPartitionStrategy() {
-        return AutoPartitionStrategy.from(config);
+        AutoPartitionStrategy s = autoPartitionStrategy;
+        if (s == null) {
+            s = AutoPartitionStrategy.from(config);
+            autoPartitionStrategy = s;
+        }
+        return s;
     }
 
     /** Gets the number of auto-increment IDs cached per segment. */
--- a/fluss-common/src/main/java/org/apache/fluss/config/TableConfig.java
+++ b/fluss-common/src/main/java/org/apache/fluss/config/TableConfig.java
@@ -161,7 +161,12 @@ public class TableConfig {
         return ArrowCompressionInfo.fromConf(config);
     }
 
-    /** Gets the auto partition strategy of the table. */
+    /**
+     * Gets the auto partition strategy of the table.
+     *
+     * <p>The result is cached on first access; this assumes the underlying config is not mutated
+     * after this {@code TableConfig} is constructed.
+     */
     public AutoPartitionStrategy getAutoPartitionStrategy() {
         AutoPartitionStrategy s = autoPartitionStrategy;
         if (s == null) {
--- a/fluss-common/src/main/java/org/apache/fluss/config/TableConfig.java
+++ b/fluss-common/src/main/java/org/apache/fluss/config/TableConfig.java
@@ -44,11 +44,6 @@ public class TableConfig {
     // the table properties configuration
     private final Configuration config;
 
-    // Cached, lazily-built strategy. Safe because TableConfig is immutable after construction
-    // and AutoPartitionStrategy itself is fully immutable. A benign race on first call
-    // produces the same value, so a plain volatile is sufficient.
-    private volatile AutoPartitionStrategy autoPartitionStrategy;
-
     /**
      * Creates a new table config.
      *
@@ -161,19 +156,9 @@ public class TableConfig {
         return ArrowCompressionInfo.fromConf(config);
     }
 
-    /**
-     * Gets the auto partition strategy of the table.
-     *
-     * <p>The result is cached on first access; this assumes the underlying config is not mutated
-     * after this {@code TableConfig} is constructed.
-     */
+    /** Gets the auto partition strategy of the table. */
     public AutoPartitionStrategy getAutoPartitionStrategy() {
-        AutoPartitionStrategy s = autoPartitionStrategy;
-        if (s == null) {
-            s = AutoPartitionStrategy.from(config);
-            autoPartitionStrategy = s;
-        }
-        return s;
+        return AutoPartitionStrategy.from(config);
     }
 
     /** Gets the number of auto-increment IDs cached per segment. */
--- a/fluss-client/src/main/java/org/apache/fluss/client/write/DynamicPartitionCreator.java
+++ b/fluss-client/src/main/java/org/apache/fluss/client/write/DynamicPartitionCreator.java
@@ -23,6 +23,7 @@ import org.apache.fluss.exception.FlussRuntimeException;
 import org.apache.fluss.exception.PartitionNotExistException;
 import org.apache.fluss.metadata.PhysicalTablePath;
 import org.apache.fluss.metadata.ResolvedPartitionSpec;
+import org.apache.fluss.metadata.TableInfo;
 import org.apache.fluss.metadata.TablePath;
 import org.apache.fluss.utils.AutoPartitionStrategy;
 import org.apache.fluss.utils.ExceptionUtils;
@@ -66,9 +67,7 @@ public class DynamicPartitionCreator {
     }
 
     public void checkAndCreatePartitionAsync(
-            PhysicalTablePath physicalTablePath,
-            List<String> partitionKeys,
-            AutoPartitionStrategy autoPartitionStrategy) {
+            PhysicalTablePath physicalTablePath, TableInfo tableInfo) {
         String partitionName = physicalTablePath.getPartitionName();
         if (partitionName == null) {
             // no need to check and create partition
@@ -87,7 +86,11 @@ public class DynamicPartitionCreator {
                 // if the partition exists, we should skip creating it.
                 LOG.debug("Partition {} already exists, skipping.", physicalTablePath);
             } else {
-                // Validate early, before touching any state.
+                // Validate early, before touching any state. The strategy is only resolved here,
+                // on the partition-creation path, not on the common "already exists" path.
+                List<String> partitionKeys = tableInfo.getPartitionKeys();
+                AutoPartitionStrategy autoPartitionStrategy =
+                        tableInfo.getTableConfig().getAutoPartitionStrategy();
                 ResolvedPartitionSpec resolvedPartitionSpec =
                         ResolvedPartitionSpec.fromPartitionName(partitionKeys, partitionName);
                 validateAutoPartitionTime(
--- a/fluss-common/src/test/java/org/apache/fluss/config/TableConfigTest.java
+++ b/fluss-common/src/test/java/org/apache/fluss/config/TableConfigTest.java
@@ -44,13 +44,4 @@ class TableConfigTest {
         TableConfig tableConfig3 = new TableConfig(conf);
         assertThat(tableConfig3.getDeleteBehavior()).hasValue(DeleteBehavior.IGNORE);
     }
-
-    @Test
-    void testAutoPartitionStrategyIsCached() {
-        TableConfig tableConfig = new TableConfig(new Configuration());
-
-        // the strategy is memoized, so repeated calls return the same instance
-        assertThat(tableConfig.getAutoPartitionStrategy())
-                .isSameAs(tableConfig.getAutoPartitionStrategy());
-    }
 }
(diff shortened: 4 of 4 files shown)


fluss-client/src/main/java/org/apache/fluss/client/write/WriterClient.java
```diff
@@ -176,10 +176,15 @@ private void doSend(WriteRecord record, WriteCallback callback) {
...
             TableInfo tableInfo = record.getTableInfo();
             PhysicalTablePath physicalTablePath = record.getPhysicalTablePath();
-            dynamicPartitionCreator.checkAndCreatePartitionAsync(
-                    physicalTablePath,
-                    tableInfo.getPartitionKeys(),
-                    tableInfo.getTableConfig().getAutoPartitionStrategy());
+            // Skip on non-partitioned tables: the callee returns immediately when
+            // partitionName is null, but the expensive AutoPartitionStrategy argument
+            // would still be evaluated per record without this guard.
+            if (!tableInfo.getPartitionKeys().isEmpty()) {
```
morazow: Maybe use tableInfo.isPartitioned here?

fluss-common/src/main/java/org/apache/fluss/config/TableConfig.java
```diff
@@ -158,7 +163,12 @@ public ArrowCompressionInfo getArrowCompressionInfo() {
 
     /** Gets the auto partition strategy of the table. */
     public AutoPartitionStrategy getAutoPartitionStrategy() {
```
morazow, line 160: Should we add test for this asserting returns same instance across calls?

binary-signal:
PTAL @fresh-borzoni && @luoyuxia xD

fluss-common/src/main/java/org/apache/fluss/config/TableConfig.java
```diff
@@ -158,7 +163,12 @@ public ArrowCompressionInfo getArrowCompressionInfo() {
 
     /** Gets the auto partition strategy of the table. */
```
morazow: We should also extend the javadoc here:

```
 * <p>The result is cached on first access; this assumes the underlying config is not
 * mutated after this {@code TableConfig} is constructed.
```

morazow (changes requested):
Thanks @binary-signal great improvements! 🚀 

I have added couple suggestions.

Another question: Do you have benchmarks / flamegraphs for autoPartitionStrategy caching only? E.g, with partitioned table.

fluss-common/src/main/java/org/apache/fluss/config/TableConfig.java
binary-signal: test added

fluss-common/src/main/java/org/apache/fluss/config/TableConfig.java
binary-signal: comment addressed

fluss-client/src/main/java/org/apache/fluss/client/write/WriterClient.java
binary-signal: comment was addressed

binary-signal:
> Thanks @binary-signal great improvements! 🚀
> 
> I have added couple suggestions.
> 
> Another question: Do you have benchmarks / flamegraphs for autoPartitionStrategy caching only? E.g, with partitioned table.


Good question  and no, i don't have benchmarks / flamegraphs  for autoPartitionStrategy caching.

The benchmark was run on a non-partitioned log table. On that path the new `isPartitioned()` guard short-circuits before `getAutoPartitionStrategy()` is ever called, so the memoisation (patch 2) contributes ~nothing~ to those results, the entire gain there comes from patch 1 skipping the per-record work.

The caching only matters for partitioned tables. There the guard passes, and since Java evaluates the argument eagerly, `getAutoPartitionStrategy()` runs on every record (`WriterClient.java:182`):

```java
private void doSend(WriteRecord record, WriteCallback callback) {
    ...
    ...
    if (tableInfo.isPartitioned()) {
        dynamicPartitionCreator.checkAndCreatePartitionAsync(
                physicalTablePath,
                tableInfo.getPartitionKeys(),
                tableInfo.getTableConfig().getAutoPartitionStrategy()); // arg evaluated per record
    }
    ...

```

The cache turns that per-record call into a single volatile read (`TableConfig.java:165`):

```java
public AutoPartitionStrategy getAutoPartitionStrategy() {
    AutoPartitionStrategy s = autoPartitionStrategy; // volatile read on the hot path
    if (s == null) {                                 // construct once
        s = AutoPartitionStrategy.from(config);
        autoPartitionStrategy = s;
    }
    return s;
}
```

Without it, each record paid a fresh `AutoPartitionStrategy.from(config)`, which includes the `TimeZone.getTimeZone()` lookup that showed up in the flamegraph (`AutoPartitionStrategy.java:57`):

```java
public static AutoPartitionStrategy from(Configuration conf) {
    return new AutoPartitionStrategy(
            conf.getBoolean(ConfigOptions.TABLE_AUTO_PARTITION_ENABLED),
            conf.getString(ConfigOptions.TABLE_AUTO_PARTITION_KEY),
            conf.get(ConfigOptions.TABLE_AUTO_PARTITION_TIME_UNIT),
            conf.getInt(ConfigOptions.TABLE_AUTO_PARTITION_NUM_PRECREATE),
            conf.getInt(ConfigOptions.TABLE_AUTO_PARTITION_NUM_RETENTION),
            TimeZone.getTimeZone(conf.getString(ConfigOptions.TABLE_AUTO_PARTITION_TIMEZONE))); // costly lookup
}
```

So I'd expect a similar flamegraph delta to the non-partitioned case  `getTimeZone()` + `AutoPartitionStrategy.from()` disappearing from the writer hot path  but I haven't measured it yet.

morazow (approved):
Thanks for explanation @binary-signal 🙏 

Looks good from my side. Let's wait for PMC / contributors to review and merge 🤝

fluss-client/src/main/java/org/apache/fluss/client/write/WriterClient.java
```diff
@@ -176,10 +176,15 @@ private void doSend(WriteRecord record, WriteCallback callback) {
...
-                    tableInfo.getPartitionKeys(),
-                    tableInfo.getTableConfig().getAutoPartitionStrategy());
+            // Skip on non-partitioned tables: the callee returns immediately when
+            // partitionName is null, but the expensive AutoPartitionStrategy argument
+            // would still be evaluated per record without this guard.
+            if (tableInfo.isPartitioned()) {
+                dynamicPartitionCreator.checkAndCreatePartitionAsync(
+                        physicalTablePath,
+                        tableInfo.getPartitionKeys(),
+                        tableInfo.getTableConfig().getAutoPartitionStrategy());
```
fresh-borzoni: I wonder if instead of caching we can do a bit different. 
The strategy is only actually used when a partition gets created, on the common "already exists" path it's built and thrown away. 

 Instead of caching it, you could pass tableInfo into checkAndCreatePartitionAsync and resolve getAutoPartitionStrategy() inside the create branch.? Then the cached field + its thread-safety reasoning aren't needed at all

WDYT?

fresh-borzoni (commented):
@binary-signal Thank you for the PR, left suggestion, PTAL

fluss-client/src/main/java/org/apache/fluss/client/write/WriterClient.java
binary-signal: That makes sense, I will provide a revised version later today or tomorrow

binary-signal: comment addessed

binary-signal:
@fresh-borzoni PTAL 🙏

fresh-borzoni (approved):
@binary-signal LGTM, thank you

PR #3535, opened 2026-06-29
[lake/iceberg] Fix integer overflow when converting Iceberg LocalTime to Fluss millis-of-day

Guosmilesmile:
### Purpose

Linked issue: close #xxx

Fix an integer-overflow bug in the fluss-iceberg lake source path when reading Iceberg TIME columns (and ARRAY<TIME>).
The previous implementation in IcebergRecordAsFlussRow#getInt and IcebergArrayAsFlussArray#getInt was:

```
return (int) ((LocalTime) value).toNanoOfDay() / 1_000_000;
```
Due to Java operator precedence the (int) cast is applied before the division, casting the long returned by toNanoOfDay() (up to 86_399_999_999_999) to int, which silently overflows for any LocalTime greater than ~00:00:02.148. As a result, e.g. 09:00:00 is decoded as -1859 instead of 32_400_000

https://github.com/apache/fluss/blob/486fe0a8ae8c0677c468b94dfdfec39535453f36/fluss-common/src/main/java/org/apache/fluss/predicate/PredicateBuilder.java#L303

In PredicateBuilder handling here is correct, so we should adjusted the error part to use the corresponding method.

### Brief change log

Fix IcebergRecordAsFlussRow#getInt to perform the division on long and only cast the final value:

```

  return (int) (((LocalTime) value).toNanoOfDay() / 1_000_000L);
  
```

### Tests

This change is verified by the new unit tests:
* testGetIntWithLocalDateAndLocalTime — sanity test for normal values on both row and array paths.
* testGetIntWithLocalTimeNearEndOfDay — asserts toNanoOfDay() > Integer.MAX_VALUE before the conversion, then asserts that getInt returns the expected 86_399_999, proving the overflow is fixed.

### API and Format

### Documentation

### DIFF
--- a/fluss-lake/fluss-lake-iceberg/src/main/java/org/apache/fluss/lake/iceberg/source/IcebergArrayAsFlussArray.java
+++ b/fluss-lake/fluss-lake-iceberg/src/main/java/org/apache/fluss/lake/iceberg/source/IcebergArrayAsFlussArray.java
@@ -79,7 +79,7 @@ public class IcebergArrayAsFlussArray implements InternalArray {
             return (int) ((LocalDate) value).toEpochDay();
         }
         if (value instanceof LocalTime) {
-            return (int) ((LocalTime) value).toNanoOfDay() / 1_000_000;
+            return (int) (((LocalTime) value).toNanoOfDay() / 1_000_000);
         }
         return (Integer) value;
     }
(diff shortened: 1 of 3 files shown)


Guosmilesmile:
Hey @fresh-borzoni @luoyuxia , if it's convenient , could you help trigger the CI pipeline and review the corresponding code? 

Thanks!

luoyuxia (approved):

Guosmilesmile:
Thanks @luoyuxia for reviewing and merging !

PR #3539, opened 2026-06-29
[server] Reject KV leader promotion under disk write lock

### LINKED ISSUE #3538: Primary-key table leader migration can amplify disk pressure under high disk usage
### Search before asking

### Fluss version

0.9.0 (latest release)

### Please describe the bug 🐞

When `TabletServers` are close to full disk usage, a `TabletServer` shutdown can trigger leader migration. For log tables, leader migration usually does not need to download much new data, so the existing data disk write protection limits the impact.

For primary-key tables, however, making a replica leader may initialize KV state by downloading remote KV snapshots. If this happens while the target `TabletServers` is already above the data disk write-limit ratio, the snapshot download can fill the disk and make the cluster harder to recover.

This can turn a single `TabletServer` shutdown under high disk usage into broader disk pressure across the cluster, because newly elected primary-key table leaders may consume additional local disk space during recovery.

### Solution

_No response_

swuferhong:
### Purpose

Linked issue: close #3538 

When a `TabletServer` is under data disk write protection, making a
primary-key table replica leader may download KV snapshots and consume
more local disk space. This can fill the disk during leader migration
and amplify failures when the cluster is already under high disk usage.

Reject new KV leader promotion with `DISK_WRITE_LOCKED` when the local
tablet server is write-locked. Log table leader promotion and idempotent
leader notifications are not affected.


fluss-server/src/main/java/org/apache/fluss/server/coordinator/CoordinatorEventProcessor.java
```diff
@@ -262,6 +269,18 @@ public CoordinatorEventProcessor(
...
+                conf.get(ConfigOptions.COORDINATOR_OFFLINE_LEADER_RETRY_INTERVAL).toMillis();
+        if (offlineLeaderRetryIntervalMs <= 0) {
+            throw new IllegalArgumentException(
+                    String.format(
+                            "%s must be positive, but was %d ms.",
+                            ConfigOptions.COORDINATOR_OFFLINE_LEADER_RETRY_INTERVAL.key(),
+                            offlineLeaderRetryIntervalMs));
+        }
+        this.offlineLeaderRetryScheduler =
+                Executors.newSingleThreadScheduledExecutor(
```
wuchong: Reuse a shared scheduler for lightweight coordinator tasks

This change adds another component-owned `ScheduledExecutorService` just to periodically enqueue a lightweight coordinator event. Threads are a precious resource, and we should be deliberate before introducing one more scheduler per feature. The coordinator side already has several similar lightweight schedulers (`TableLifecycleThrottler`, `AutoPartitionManager`, `RebalanceManager`, `LakeTableTieringManager`, `ProducerOffsetsManager`, `KvSnapshotLeaseManager`), which makes it easy for future changes to keep adding dedicated threads unchecked.

Could we follow the TabletServer pattern instead: let the server own a shared `FlussScheduler`/background scheduler and pass it to coordinator components that only need lightweight delayed or periodic tasks? At minimum, this new retry logic should reuse an existing shared scheduler rather than creating its own single-thread executor.

fluss-server/src/main/java/org/apache/fluss/server/coordinator/CoordinatorEventProcessor.java
```diff
@@ -587,6 +609,25 @@ private void onShutdown() {
         tabletServerChangeWatcher.stop();
     }
```
wuchong, line 611: Make offline-leader retry scheduling lazy instead of polling forever

The current implementation starts a fixed-delay task at coordinator startup and enqueues `RetryOfflineLeaderEvent` every interval, even when there are no offline replicas to retry. In the common healthy case this means the coordinator wakes up forever, enters the event queue, scans an empty set, and returns. This is lightweight per run, but it is still another always-on background loop, and it becomes a pattern that is easy to copy for future recovery paths.

Could we make the retry scheduler event-driven/lazy instead? For example:

1. Do not start the offline-leader retry task during `CoordinatorEventProcessor.startup()`.
2. When `onReplicaBecomeOffline(...)` adds an offline marker, check whether there are retryable offline replicas on live tablet servers.
3. If there is no existing scheduled retry, schedule a one-shot retry after `coordinator.offline-leader.retry-interval`.
4. When the scheduled task fires, only enqueue one `RetryOfflineLeaderEvent`; keep all `CoordinatorContext` reads/writes inside the coordinator event thread.
5. In `processRetryOfflineLeader()`, clear the scheduled-future marker before processing.
6. If the retry still fails, the later `NotifyLeaderAndIsrResponseReceivedEvent` will mark the replica offline again and schedule the next one-shot retry.
7. If the retry succeeds, no replica is marked offline again, so no further retry is scheduled.

This keeps the same recovery behavior while avoiding permanent empty polling in healthy clusters.

wuchong (commented):
I left some minor comments.

fluss-server/src/main/java/org/apache/fluss/server/coordinator/CoordinatorEventProcessor.java
swuferhong: Updated. This retry logic now reuses the shared coordinator scheduler, and I added TODOs for the existing coordinator components that still own dedicated lightweight schedulers so they can be migrated in follow-up PRs.

wuchong (approved):
LGTM

PR #3540, opened 2026-06-29
[fs] Support AWS S3 credentials provider mode

### LINKED ISSUE #3493: [feature] Support AWS S3 credentials provider mode to refresh credentials without restarting servers
### Search before asking

### Motivation

Some Fluss deployments access S3 with long-term IAM user AK/SK stored in the AWS shared credentials file (`~/.aws/credentials`) instead of putting `fs.s3a.access.key` / `fs.s3a.secret.key` in `server.yaml`. Those AK/SK are rotated periodically for security compliance, and today the only way to pick up rotated credentials is to restart the affected Fluss server processes. Restarting tablet servers can make ongoing writes unstable and temporarily affect cluster availability.

Fluss should support the standard Hadoop S3A `fs.s3a.aws.credentials.provider` configuration (e.g. `com.amazonaws.auth.profile.ProfileCredentialsProvider`) so credentials can refresh without restarting servers.

### Solution

If `fs.s3a.aws.credentials.provider` is explicitly configured, treat it as **server-side provider mode** unless it is rejected together with `fs.s3a.assumed.role.arn`: let S3A use the provider for server operations, and have `S3DelegationTokenProvider` use the same provider for STS `GetSessionToken`. The explicit provider list is authoritative; when static AK/SK are also configured, the provider wins. Do not inject `DynamicTemporaryAWSCredentialsProvider`. Keep returning the same `ObtainedSecurityToken`, so clients need no changes. Rely on the AWS SDK provider's own refresh (no Fluss-managed reload scheduler).

**Goals**

- Support an explicitly configured `fs.s3a.aws.credentials.provider`, especially `ProfileCredentialsProvider`.
- Allow rotating AK/SK in `~/.aws/credentials` without restarting servers, applied to both `remote.data.dir` access and `S3DelegationTokenProvider`'s STS calls.
- Preserve existing static AK/SK, AssumeRole, and client delegated-token behavior when no server-side provider is explicitly configured.
- Never log or leak secret values.

**Non-goals**

- No Fluss-managed credential-file reload scheduler; no hot-reload of static `fs.s3a.access.key` / `fs.s3a.secret.key`; no storing AK/SK in ZooKeeper or dynamic config.
- No custom provider + AssumeRole source credentials in the first version.
- No session/temporary credentials as input for client-token generation.
- No rebuilding the Fluss `FileSystem` cache at runtime; in-flight streams may not switch credentials mid-stream.

**Current behavior**

`S3FileSystemPlugin` treats the filesystem as server mode only when static AK/SK or `fs.s3a.assumed.role.arn` is set; otherwise it injects `DynamicTemporaryAWSCredentialsProvider` (a client delegated-token helper), blocking normal AWS SDK provider use. `S3DelegationTokenProvider` captures static AK/SK in its constructor and the snapshot never refreshes.

**Credential mode resolution**

`S3FileSystemPlugin` selects the mode with an explicit priority matrix (top to bottom):

| # | Static AK/SK | roleArn | Explicit provider | Mode |
|---|---|---|---|---|
| 1 | any | set | set | **rejected** (provider + AssumeRole is a non-goal in v1) |
| 2 | any | unset | set | **server-side provider mode (new; explicit provider wins over static AK/SK)** |
| 3 | set | unset | unset | static AK/SK (unchanged) |
| 4 | set | set | unset | existing static AK/SK + AssumeRole behavior (unchanged) |
| 5 | unset | set | unset | AssumeRole (unchanged) |
| 6 | unset | unset | unset | delegated-token client path (unchanged) |

parse `fs.s3a.aws.credentials.provider` as a class list (split on `,`, trim, exact fully-qualified-name compare) — **not** substring `contains`. The matrix applies to both `s3://` and `s3a://`.

`S3FileSystemPlugin` determines explicit provider mode only from Fluss configuration keys (`s3.aws.credentials.provider`, `s3a.aws.credentials.provider`, or `fs.s3a.aws.credentials.provider`). It then writes an internal Hadoop configuration marker so `S3DelegationTokenProvider` can distinguish an explicit Fluss provider from values that may come from Hadoop default resources. The marker is not a user-facing option.

**`S3DelegationTokenProvider` lifecycle**

Today AK/SK are frozen in the constructor and the provider instance is cached for the process lifetime — the root cause that blocks refresh. The new design splits lifetimes:

- Build the credential provider / Hadoop `AWSCredentialProviderList` **once** per instance and reuse it (preserves the provider's own refresh throttling; avoids re-reading the profile each call).
- On **each** `obtainSecurityToken()`: call `getCredentials()` again, reject `AWSSessionCredentials` before any STS call, and use the freshly resolved long-term credentials for that STS request. Never cache AK/SK in the constructor.
- The STS client may still be created per call and `shutdown()` in `finally`.

**Provider instantiation**

Reuse Hadoop S3A's own loader `S3AUtils.createAWSCredentialProviderSet(null, conf)` so the token path and server-side S3A select providers with the same Hadoop configuration and constructor support. Do not hand-write a parallel loader (it would drift). Only provider class names already accepted by S3A are supported; no Fluss wrapper provider in v1.

The token path does not require `S3FileSystem` to store or pass the filesystem URI. The S3A loader is invoked with a `null` URI and the Hadoop configuration.

**Region**

`fs.s3a.region` is a **Fluss-specific** requirement of the token path: the STS client reads region only from `fs.s3a.region` (not S3A's `fs.s3a.endpoint.region`, not the profile/environment). Provider mode does not relax it; a missing `fs.s3a.region` fails fast. Future option: fall back to `fs.s3a.endpoint.region` — deferred.

**Backward compatibility**

Default behavior, static AK/SK without an explicit provider, AssumeRole without an explicit provider, and the client-side `DynamicTemporaryAWSCredentialsProvider` path are all unchanged. Explicitly configuring a server-side provider is the only new behavior; if static AK/SK are also configured, the explicit provider wins. Clients are unaffected.

### Anything else?

`RpcServiceBase` caches the obtained token for 60s. Operators must keep old AK/SK valid for a grace period covering the provider refresh delay + 60s cache + STS token validity + longest expected in-flight S3 operation. This design intentionally avoids a Fluss-managed reload mechanism and uses the AWS SDK provider as the refresh boundary, keeping the change localized to `fluss-fs-s3`.

### Willingness to contribute

- [x] I am willing to submit a PR.

litiliu:
### Purpose

Closes #3493.

This PR adds a server-side AWS S3 credentials provider mode for the S3 filesystem. When `fs.s3a.aws.credentials.provider` is explicitly configured in Fluss configuration, Fluss treats it as the authoritative server-side credential source instead of injecting the client delegated-token provider.

This is intended for deployments that use standard AWS SDK/Hadoop S3A providers such as `com.amazonaws.auth.profile.ProfileCredentialsProvider`, so rotated long-term credentials can be picked up by the provider without restarting Fluss servers.

The full motivation and design discussion are in #3493. This PR description keeps the reviewer-facing summary and the final implemented behavior.

### Brief change log

- Detect explicitly configured S3 credential providers from Fluss config keys:
  - `s3.aws.credentials.provider`
  - `s3a.aws.credentials.provider`
  - `fs.s3a.aws.credentials.provider`
- Store an internal Hadoop configuration marker so `S3DelegationTokenProvider` can distinguish an explicit Fluss provider from Hadoop default resources.
- In explicit provider mode:
  - do not inject `DynamicTemporaryAWSCredentialsProvider`;
  - reject provider mode combined with `fs.s3a.assumed.role.arn`;
  - make the explicit provider win over static AK/SK if both are configured;
  - reuse Hadoop S3A's provider loader for the token path.
- Resolve credentials from the configured provider on each STS token request instead of freezing AK/SK in the token provider constructor.
- Reject session credentials before calling STS, so temporary credentials are not used as input for client-token generation.
- Preserve existing static AK/SK, AssumeRole, and delegated-token client behavior when no explicit server-side provider is configured.

Credential mode resolution:

| Static AK/SK | AssumeRole ARN | Explicit provider | Result |
|---|---|---|---|
| any | set | set | rejected |
| any | unset | set | server-side provider mode; provider wins |
| set | unset | unset | static AK/SK |
| set | set | unset | existing static AK/SK + AssumeRole behavior |
| unset | set | unset | AssumeRole |
| unset | unset | unset | delegated-token client path |

### Tests

- `mvn -pl fluss-filesystems/fluss-fs-s3 test -Dtest=S3FileSystemPluginTest,S3DelegationTokenProviderTest`

Added/updated coverage for:

- explicit provider mode does not inject `DynamicTemporaryAWSCredentialsProvider`;
- static AK/SK and AssumeRole modes keep existing behavior;
- explicit provider wins over static AK/SK;
- explicit provider + AssumeRole is rejected;
- explicitly configuring `DynamicTemporaryAWSCredentialsProvider` for server mode is rejected;
- configured provider credentials are resolved for each token request;
- session credentials are rejected before STS.

### API and Format

The PR adds an internal Hadoop configuration marker under `fluss.fs.s3.aws.credentials.provider.explicitly.configured`. It is not a user-facing option; it only carries whether the provider was explicitly configured through Fluss config.

### Documentation

No separate documentation update in this PR. The user-facing behavior and operational motivation are described in #3493.

### Generative AI disclosure

- [x] Yes: OpenAI Codex was used to help implement and review this change, following the repository `AGENTS.md` guidance.

### DIFF
--- a/fluss-filesystems/fluss-fs-s3/src/main/java/org/apache/fluss/fs/s3/S3FileSystemPlugin.java
+++ b/fluss-filesystems/fluss-fs-s3/src/main/java/org/apache/fluss/fs/s3/S3FileSystemPlugin.java
@@ -23,7 +23,9 @@ import org.apache.fluss.config.Configuration;
 import org.apache.fluss.fs.FileSystem;
 import org.apache.fluss.fs.FileSystemPlugin;
 import org.apache.fluss.fs.s3.token.S3ADelegationTokenReceiver;
+import org.apache.fluss.fs.s3.token.S3DelegationTokenProvider;
 import org.apache.fluss.fs.s3.token.S3DelegationTokenReceiver;
+import org.apache.fluss.utils.StringUtils;
 
 import org.apache.hadoop.fs.s3a.S3AFileSystem;
 import org.slf4j.Logger;
@@ -31,6 +33,7 @@ import org.slf4j.LoggerFactory;
 
 import java.io.IOException;
 import java.net.URI;
+import java.util.Map;
 import java.util.Objects;
 
 import static org.apache.fluss.fs.s3.token.S3DelegationTokenReceiver.PROVIDER_CONFIG_NAME;
@@ -41,6 +44,10 @@ public class S3FileSystemPlugin implements FileSystemPlugin {
     private static final Logger LOG = LoggerFactory.getLogger(S3FileSystemPlugin.class);
 
     private static final String[] FLUSS_CONFIG_PREFIXES = {"s3.", "s3a.", "fs.s3a."};
+
+    private static final String[] CREDENTIAL_PROVIDER_CONFIG_KEYS = {
+        "s3.aws.credentials.provider", "s3a.aws.credentials.provider", PROVIDER_CONFIG_NAME
+    };
 
     private static final String HADOOP_CONFIG_PREFIX = "fs.s3a.";
 
@@ -74,6 +81,12 @@ public class S3FileSystemPlugin implements FileSystemPlugin {
     org.apache.hadoop.conf.Configuration buildHadoopConfiguration(Configuration flussConfig) {
         org.apache.hadoop.conf.Configuration hadoopConfig =
                 mirrorCertainHadoopConfig(getHadoopConfiguration(flussConfig));
+        boolean hasCredentialProvider = hasConfiguredCredentialProvider(flussConfig);
+        // Preserve whether the provider came from Fluss config. Token providers should not infer
+        // explicit server-side provider mode from Hadoop default resources.
+        hadoopConfig.setBoolean(
+                S3DelegationTokenProvider.CREDENTIAL_PROVIDER_EXPLICITLY_CONFIGURED,
+                hasCredentialProvider);
         setCredentialProvider(hadoopConfig);
         return hadoopConfig;
     }
@@ -130,27 +143,55 @@ public class S3FileSystemPlugin implements FileSystemPlugin {
     }
 
     private void setCredentialProvider(org.apache.hadoop.conf.Configuration hadoopConfig) {
+        boolean hasCredentialProvider =
+                hadoopConfig.getBoolean(
+                        S3DelegationTokenProvider.CREDENTIAL_PROVIDER_EXPLICITLY_CONFIGURED, false);
         boolean hasStaticKeys =
                 hadoopConfig.get(ACCESS_KEY_ID) != null
                         && hadoopConfig.get(ACCESS_KEY_SECRET) != null;
         boolean hasRoleArn = hadoopConfig.get(ROLE_ARN_KEY) != null;
+
+        if (hasCredentialProvider) {
+            if (hasRoleArn) {
+                throw new IllegalArgumentException(
+                        "AssumeRole and a custom AWS credentials provider cannot be configured together.");
+            }
+            LOG.info(
+                    "Using configured AWS credential provider(s) for server-side S3 access: {}",
+                    hadoopConfig.get(PROVIDER_CONFIG_NAME));
+            return;
+        }
 
         if (hasStaticKeys || hasRoleArn) {
             LOG.info(
                     hasStaticKeys
                             ? "Using provided static credentials."
                             : "Using default AWS credential chain with AssumeRole.");
-        } else {
-            if (Objects.equals(getScheme(), "s3")) {
-                S3DelegationTokenReceiver.updateHadoopConfig(hadoopConfig);
-            } else if (Objects.equals(getScheme(), "s3a")) {
-                S3ADelegationTokenReceiver.updateHadoopConfig(hadoopConfig);
-            } else {
-                throw new IllegalArgumentException("Unsupported scheme: " + getScheme());
-            }
-            LOG.info(
-                    "Using credential provider {} for delegated tokens.",
-                    hadoopConfig.get(PROVIDER_CONFIG_NAME));
+            return;
         }
+
+        if (Objects.equals(getScheme(), "s3")) {
+            S3DelegationTokenReceiver.updateHadoopConfig(hadoopConfig);
+        } else if (Objects.equals(getScheme(), "s3a")) {
+            S3ADelegationTokenReceiver.updateHadoopConfig(hadoopConfig);
+        } else {
+            throw new IllegalArgumentException("Unsupported scheme: " + getScheme());
+        }
+        LOG.info(
+                "Using credential provider {} for delegated tokens.",
+                hadoopConfig.get(PROVIDER_CONFIG_NAME));
+    }
+
+    private boolean hasConfiguredCredentialProvider(Configuration flussConfig) {
+        if (flussConfig == null) {
+            return false;
+        }
+        Map<String, String> configMap = flussConfig.toMap();
+        for (String key : CREDENTIAL_PROVIDER_CONFIG_KEYS) {
+            if (!StringUtils.isNullOrWhitespaceOnly(configMap.get(key))) {
+                return true;
+            }
+        }
+        return false;
     }
 }
--- a/fluss-filesystems/fluss-fs-s3/src/main/java/org/apache/fluss/fs/s3/token/S3DelegationTokenProvider.java
+++ b/fluss-filesystems/fluss-fs-s3/src/main/java/org/apache/fluss/fs/s3/token/S3DelegationTokenProvider.java
@@ -19,7 +19,11 @@ package org.apache.fluss.fs.s3.token;
 
 import org.apache.fluss.fs.token.CredentialsJsonSerde;
 import org.apache.fluss.fs.token.ObtainedSecurityToken;
+import org.apache.fluss.utils.StringUtils;
 
+import com.amazonaws.auth.AWSCredentials;
+import com.amazonaws.auth.AWSCredentialsProvider;
+import com.amazonaws.auth.AWSSessionCredentials;
 import com.amazonaws.auth.AWSStaticCredentialsProvider;
 import com.amazonaws.auth.BasicAWSCredentials;
 import com.amazonaws.client.builder.AwsClientBuilder;
@@ -30,11 +34,14 @@ import com.amazonaws.services.securitytoken.model.AssumeRoleResult;
 import com.amazonaws.services.securitytoken.model.Credentials;
 import com.amazonaws.services.securitytoken.model.GetSessionTokenResult;
 import org.apache.hadoop.conf.Configuration;
+import org.apache.hadoop.fs.s3a.AWSCredentialProviderList;
+import org.apache.hadoop.fs.s3a.S3AUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import javax.annotation.Nullable;
 
+import java.io.IOException;
 import java.util.Arrays;
 import java.util.HashMap;
 import java.util.Map;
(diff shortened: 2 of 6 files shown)


litiliu:
@fresh-borzoni could you please help review this when you have time? Thanks!

fluss-filesystems/fluss-fs-s3/src/main/java/org/apache/fluss/fs/s3/token/S3DelegationTokenProvider.java
```diff
@@ -63,21 +73,31 @@ public class S3DelegationTokenProvider {
...
 
         checkArgument(
                 (accessKey == null) == (secretKey == null),
                 "S3 access key and secret key must both be set or both be unset.");
-        if (accessKey == null) {
+        if (hasCredentialProvider && roleArn != null) {
+            throw new IllegalArgumentException(
+                    "AssumeRole and a custom AWS credentials provider cannot be configured together.");
+        }
+        this.credentialProviderList =
```
fresh-borzoni, line 106: The description says configuring DynamicTemporaryAWSCredentialsProvider for server mode is "rejected", but there's no guard, so it just gets instantiated here and throws NoAwsCredentialsException at the first token request. Can you clarify?

fluss-filesystems/fluss-fs-s3/src/main/java/org/apache/fluss/fs/s3/token/S3DelegationTokenProvider.java
```diff
@@ -147,6 +168,33 @@ private AWSSecurityTokenService buildStsClient() {
         return builder.build();
     }
 
+    @Nullable
+    AWSCredentialsProvider createStsCredentialsProvider() {
+        if (credentialProviderList != null) {
+            AWSCredentials credentials = credentialProviderList.getCredentials();
+            checkArgument(
```
fresh-borzoni, line 183: This is effectively the "long-term creds only" gate, and it's the thing people will trip on  instance profiles/IRSA roles all return session creds and land here at token time (lazily). 

Given s3.md already has an IRSA section, can we add a short note there for this new mode: long-term creds only, not compatible with AssumeRole? Otherwise it's a bit hidden for operational usage

fluss-filesystems/fluss-fs-s3/src/main/java/org/apache/fluss/fs/s3/token/S3DelegationTokenProvider.java
```diff
@@ -29,12 +32,16 @@
 import com.amazonaws.services.securitytoken.model.AssumeRoleResult;
 import com.amazonaws.services.securitytoken.model.Credentials;
 import com.amazonaws.services.securitytoken.model.GetSessionTokenResult;
+import org.apache.commons.lang3.StringUtils;
```
fresh-borzoni: commons-lang3 is only transitive here, shall we use org.apache.commons.lang3.StringUtils?

fluss-filesystems/fluss-fs-s3/src/main/java/org/apache/fluss/fs/s3/S3FileSystemPlugin.java
```diff
@@ -23,14 +23,17 @@
 import org.apache.fluss.fs.FileSystem;
 import org.apache.fluss.fs.FileSystemPlugin;
 import org.apache.fluss.fs.s3.token.S3ADelegationTokenReceiver;
+import org.apache.fluss.fs.s3.token.S3DelegationTokenProvider;
 import org.apache.fluss.fs.s3.token.S3DelegationTokenReceiver;
 
+import org.apache.commons.lang3.StringUtils;
```
fresh-borzoni: ditto

fluss-filesystems/fluss-fs-s3/src/main/java/org/apache/fluss/fs/s3/token/S3DelegationTokenProvider.java
```diff
@@ -63,21 +73,31 @@ public class S3DelegationTokenProvider {
...
         this.stsEndpoint = conf.get(STS_ENDPOINT_KEY);
+        boolean hasCredentialProvider =
+                conf.getBoolean(CREDENTIAL_PROVIDER_EXPLICITLY_CONFIGURED, false)
+                        && StringUtils.isNotBlank(conf.getTrimmed(AWS_CREDENTIALS_PROVIDER));
 
         checkArgument(
                 (accessKey == null) == (secretKey == null),
                 "S3 access key and secret key must both be set or both be unset.");
-        if (accessKey == null) {
+        if (hasCredentialProvider && roleArn != null) {
```
fresh-borzoni, line 95: Since we're adding all these checks in this PR -  this one fails fast, but the session/empty-cred cases only fail on the first token request. 

Could we check those at construction too, so they're consistent? Not blocking though

fresh-borzoni (commented):
@litiliu Thank you for the PR, looks good overall, couple of minor comments, PTAL

fluss-filesystems/fluss-fs-s3/src/main/java/org/apache/fluss/fs/s3/token/S3DelegationTokenProvider.java
litiliu: Good catch. DynamicTemporaryAWSCredentialsProvider is an internal provider used for client-side delegated credentials and is not intended to be configured manually as a server-side credential source. However, since the configuration accepts arbitrary provider class names, I've added an explicit check to reject it before constructing the S3A provider list, along with a regression test. It now fails with a clear configuration error instead of failing on the first credential request. Thanks!

fluss-filesystems/fluss-fs-s3/src/main/java/org/apache/fluss/fs/s3/token/S3DelegationTokenProvider.java
litiliu: Thanks for the suggestion. S3DelegationTokenProvider itself is lazily constructed during the first obtainSecurityToken() call, so resolving credentials in the constructor would not move the failure to server startup; it would only perform the validation earlier within the same request. Since the credentials are already validated before calling STS and this is non-blocking, I'd prefer to keep the resolution lazy.

fluss-filesystems/fluss-fs-s3/src/main/java/org/apache/fluss/fs/s3/token/S3DelegationTokenProvider.java
litiliu: Added a note to the IRSA/instance profile section clarifying that the explicit server-side provider mode only supports long-term credentials and cannot be combined with s3.assumed.role.arn. IRSA and instance profile providers should continue to use the default AWS credential chain. Thanks!

fluss-filesystems/fluss-fs-s3/src/main/java/org/apache/fluss/fs/s3/token/S3DelegationTokenProvider.java
litiliu: Good point. I replaced org.apache.commons.lang3.StringUtils with Fluss's org.apache.fluss.utils.StringUtils, so we no longer rely on the transitive commons-lang3 dependency.

fluss-filesystems/fluss-fs-s3/src/main/java/org/apache/fluss/fs/s3/S3FileSystemPlugin.java
litiliu: Updated this one as well to use Fluss's StringUtils.

litiliu:
@fresh-borzoni Thanks for the review! I've addressed the comments, added the S3 credential-provider documentation, and rebased the branch onto the latest main. PTAL when you have time.

fresh-borzoni (approved):
@litiliu Thank you, I've added small change in the docs for the use-case you added(otherwise it's only the note, so it's a bit weird), hope it's alright, merging.

PR #3528, opened 2026-06-25
[lake/hudi] Add Hudi Flink union read log table IT

fhan688:
### Purpose

  Linked issue: https://github.com/apache/fluss/issues/3284

  This PR adds integration tests for Hudi log table union read based on the Flink union read test base introduced in #3522.

  The tests verify that Fluss can read tiered Hudi snapshot data together with newly appended Fluss log data for datalake-enabled  log tables.

### Brief change log

  - Add `FlinkUnionReadLogTableITCase` for Hudi log table union read.
  - Verify batch full-type reads for both partitioned and non-partitioned log tables.
  - Verify union read after the tiering job is canceled and new Fluss log records are appended.
  - Verify projection pushdown and partition filter pushdown.
  - Verify streaming union read with partition discovery.
  - Verify savepoint-based failover and restore for a streaming insert job reading from a Hudi-backed log table.
  - Reuse the Hudi Flink union read test base from #3522 and enable checkpointing for the tiering job.

### Tests

  - `mvn -pl fluss-lake/fluss-lake-hudi -Dcheckstyle.skip=true spotless:apply`
  - `mvn -pl fluss-client,fluss-lake/fluss-lake-hudi -am -DskipITs -Dcheckstyle.skip=true -DfailIfNoTests=false test-compile`
  - `mvn -pl fluss-lake/fluss-lake-hudi -am -DskipITs=false -Dcheckstyle.skip=true -DfailIfNoTests=false
  -Dtest=FlinkUnionReadLogTableITCase test`

### API and Format

  No API or storage format changes.

### Documentation

  No documentation changes. This PR only adds integration test coverage.

PR #3522, opened 2026-06-25
[lake/hudi] Add Flink union read test base for Hudi

fhan688:
### Purpose

  Linked issue: N/A

  This PR adds the shared Flink union read test base for Hudi, preparing follow-up IT cases for validating Hudi tiering union reads from lake snapshots and Fluss logs.

### Brief change log

  - Add `FlinkUnionReadTestBase` for Hudi union read tests.
  - Initialize batch and streaming `StreamTableEnvironment` with the Fluss catalog.
  - Support rebuilding the streaming table environment from a savepoint path.
  - Add reusable Hudi tiering test helpers for waiting bucket sync and asserting replica lake offsets.

### Tests

  - `spotless:apply`
  - `mvn -pl fluss-lake/fluss-lake-hudi -am -DskipITs -Dcheckstyle.skip=true -DfailIfNoTests=false test-compile`
  - `git diff --check`

### API and Format

  No public API or storage format changes. This is test-only infrastructure.

### Documentation

  No documentation changes are needed.

Issue #3284, opened 2026-05-08
Flink Union Read Test of Hudi


fhan688:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/fluss/issues) and found nothing similar.

### Description

Flink Union Read Test of Hudi

### Willingness to contribute

- [x] I'm willing to submit a PR!

PR #3541, opened 2026-06-29
[lake/hudi] Add Hudi Flink union read primary key table IT

fhan688:
Linked issue: https://github.com/apache/fluss/issues/3284

Extend the Hudi lake integration so Flink union read works for primary key tables. Building on the union-read test base added in #3522 and the log-table IT in #3528, this PR adds the missing piece for PK tables: a sorted record reader for Hudi merge-on-read splits (so remote Hudi rows can be merged with Fluss log records by primary key) plus end-to-end IT coverage.

### Brief change log

- `fluss-lake/fluss-lake-hudi/src/main/java/org/apache/fluss/lake/hudi/source/HudiSortedRecordReader.java` — new `SortedRecordReader` implementation for Hudi. Resolves record-key columns from `FlinkOptions.RECORD_KEY_FIELD`, drains the underlying `HudiRecordReader`, sorts in memory by the projected key positions, and exposes `order()` so the union-read merger can interleave remote Hudi rows with Fluss log records. Includes a `canSortByRecordKey(...)` predicate that rejects projections dropping any record-key column.
- `fluss-lake/fluss-lake-hudi/src/main/java/org/apache/fluss/lake/hudi/source/HudiLakeSource.java` — in `createRecordReader`, route to `HudiSortedRecordReader` when the Hudi table is `MERGE_ON_READ` and the projection retains every record-key field; otherwise keep the existing `HudiRecordReader` path. Mirrors the contract already used by `PaimonLakeSource` / `PaimonSortedRecordReader`.

### Tests

- New IT `fluss-lake/fluss-lake-hudi/src/test/java/org/apache/fluss/lake/hudi/flink/FlinkUnionReadPrimaryKeyTableITCase.java` (extends `FlinkUnionReadTestBase`):
  - `testUnionReadFullType` — batch union read over a full-type PK schema, projection (`select c3, c4`, `select c3`) and predicate pushdown (`c4 = 30`); validates state both before and after upsert.
  - `testUnionReadInStreamMode` — streaming changelog correctness on partitioned and non-partitioned PK tables, including `-U`/`+U` emission after upsert.
  - `testUnionReadPrimaryKeyTableFailover` — `INSERT ... SELECT` pipeline stopped with a canonical savepoint and resumed; verifies changelog continuity across restart.
- New UT `fluss-lake/fluss-lake-hudi/src/test/java/org/apache/fluss/lake/hudi/source/HudiSortedRecordReaderTest.java` — covers `canSortByRecordKey` projection rules for single-column and composite primary keys.

### API and Format

No public API or storage format changes. `HudiSortedRecordReader` reuses the existing `org.apache.fluss.lake.source.SortedRecordReader` contract defined in `fluss-common`.

### Documentation

No new user-facing feature; no documentation changes required.

### DIFF
--- a/fluss-lake/fluss-lake-hudi/src/main/java/org/apache/fluss/lake/hudi/source/HudiLakeSource.java
+++ b/fluss-lake/fluss-lake-hudi/src/main/java/org/apache/fluss/lake/hudi/source/HudiLakeSource.java
@@ -27,6 +27,7 @@ import org.apache.fluss.lake.source.RecordReader;
 import org.apache.fluss.metadata.TablePath;
 import org.apache.fluss.predicate.Predicate;
 
+import org.apache.hudi.common.model.HoodieTableType;
 import org.apache.hudi.org.apache.avro.Schema;
 import org.apache.hudi.source.ExpressionPredicates;
 import org.apache.hudi.util.StreamerUtil;
@@ -102,6 +103,10 @@ public class HudiLakeSource implements LakeSource<HudiSplit> {
     @Override
     public RecordReader createRecordReader(ReaderContext<HudiSplit> context) throws IOException {
         try {
+            if (shouldUseSortedRecordReader()) {
+                return new HudiSortedRecordReader(
+                        hudiConfig, tablePath, context.lakeSplit(), project, predicates);
+            }
             return new HudiRecordReader(
                     hudiConfig, tablePath, context.lakeSplit(), project, predicates);
         } catch (Exception e) {
@@ -119,6 +124,13 @@ public class HudiLakeSource implements LakeSource<HudiSplit> {
             return StreamerUtil.getTableAvroSchema(hudiTableInfo.getMetaClient(), true);
         } catch (Exception e) {
             throw new RuntimeException("Fail to get Hudi schema for " + tablePath + ".", e);
+        }
+    }
+
+    private boolean shouldUseSortedRecordReader() throws IOException {
+        try (HudiTableInfo hudiTableInfo = HudiTableInfo.create(tablePath, hudiConfig)) {
+            return hudiTableInfo.getTableType() == HoodieTableType.MERGE_ON_READ
+                    && HudiSortedRecordReader.canSortByRecordKey(hudiTableInfo, project);
         }
     }
 }
--- a/fluss-lake/fluss-lake-hudi/src/main/java/org/apache/fluss/lake/hudi/source/HudiLakeSource.java
+++ b/fluss-lake/fluss-lake-hudi/src/main/java/org/apache/fluss/lake/hudi/source/HudiLakeSource.java
@@ -103,7 +103,14 @@ public class HudiLakeSource implements LakeSource<HudiSplit> {
     @Override
     public RecordReader createRecordReader(ReaderContext<HudiSplit> context) throws IOException {
         try {
-            if (shouldUseSortedRecordReader()) {
+            if (context.requireSortedRecords()) {
+                if (!shouldUseSortedRecordReader()) {
+                    throw new UnsupportedOperationException(
+                            "Hudi lake source can not provide sorted records for "
+                                    + tablePath
+                                    + ". Sorted records are only supported for MERGE_ON_READ tables"
+                                    + " whose projection contains all Hudi record key fields.");
+                }
                 return new HudiSortedRecordReader(
                         hudiConfig, tablePath, context.lakeSplit(), project, predicates);
             }
--- a/fluss-lake/fluss-lake-hudi/src/main/java/org/apache/fluss/lake/hudi/source/HudiSortedRecordReader.java
+++ b/fluss-lake/fluss-lake-hudi/src/main/java/org/apache/fluss/lake/hudi/source/HudiSortedRecordReader.java
@@ -1,0 +1,433 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.fluss.lake.hudi.source;
+
+import org.apache.fluss.config.Configuration;
+import org.apache.fluss.lake.hudi.utils.HudiTableInfo;
+import org.apache.fluss.lake.source.RecordReader;
+import org.apache.fluss.lake.source.SortedRecordReader;
+import org.apache.fluss.metadata.TablePath;
+import org.apache.fluss.record.GenericRecord;
+import org.apache.fluss.record.LogRecord;
+import org.apache.fluss.row.BinaryString;
+import org.apache.fluss.row.Decimal;
+import org.apache.fluss.row.GenericRow;
+import org.apache.fluss.row.InternalRow;
+import org.apache.fluss.types.DataTypeRoot;
+import org.apache.fluss.utils.CloseableIterator;
+import org.apache.fluss.utils.InternalRowUtils;
+
+import org.apache.flink.table.api.Schema;
+import org.apache.flink.table.types.AbstractDataType;
+import org.apache.flink.table.types.DataType;
+import org.apache.flink.table.types.logical.DecimalType;
+import org.apache.flink.table.types.logical.LocalZonedTimestampType;
+import org.apache.flink.table.types.logical.LogicalType;
+import org.apache.flink.table.types.logical.TimestampType;
+import org.apache.hudi.configuration.FlinkOptions;
+import org.apache.hudi.source.ExpressionPredicates;
+
+import javax.annotation.Nullable;
+
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.Comparator;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.stream.Collectors;
+
+import static org.apache.fluss.lake.hudi.HudiLakeCatalog.SYSTEM_COLUMNS;
+
+/** Sorted Hudi record reader for primary key table union read. */
+public class HudiSortedRecordReader implements SortedRecordReader {
+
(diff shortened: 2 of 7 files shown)


fhan688:
please help review, thanks! @XuQianJin-Stars

fluss-lake/fluss-lake-hudi/src/main/java/org/apache/fluss/lake/hudi/source/HudiSortedRecordReader.java
```diff
@@ -0,0 +1,433 @@
...
+        CloseableIterator<LogRecord> iterator = delegate.read();
+        List<LogRecord> records = new ArrayList<>();
+        try {
+            while (iterator.hasNext()) {
+                records.add(copyRecord(iterator.next(), sortOrder.producedTypes));
+            }
+        } finally {
+            iterator.close();
+        }
+        records.sort(
```
XuQianJin-Stars, line 101: The `read() `method loads all records into memory before sorting. This may cause `OOM` for large-scale datasets. Suggestion: Consider using external sort or limiting the number of records read in a single batch.

fluss-lake/fluss-lake-hudi/src/main/java/org/apache/fluss/lake/hudi/source/HudiSortedRecordReader.java
```diff
@@ -0,0 +1,433 @@
...
+            default:
+                throw new IllegalArgumentException(
+                        "Unsupported Hudi primary key type: " + logicalType);
+        }
+    }
+
+    private static int compareValues(Object value1, Object value2, LogicalType logicalType) {
+        switch (logicalType.getTypeRoot()) {
+            case BOOLEAN:
+                return Boolean.compare((boolean) value1, (boolean) value2);
```
XuQianJin-Stars, line 368: `NullPointerException` Risk: The compareValues method directly unboxes `BOOLEAN` type without handling `null` values. Suggestion: Check for `null` values before comparison, or ensure the caller does not pass `null`.

fluss-lake/fluss-lake-hudi/src/main/java/org/apache/fluss/lake/hudi/source/HudiSortedRecordReader.java
```diff
@@ -0,0 +1,433 @@
...
+            }
+        }
+        return 0;
+    }
+
+    private static Object getField(InternalRow row, int pos, DataType dataType) {
+        LogicalType logicalType = dataType.getLogicalType();
+        switch (logicalType.getTypeRoot()) {
+            case BOOLEAN:
+                return row.getBoolean(pos);
```
XuQianJin-Stars, line 329: `getField` Method Does Not Handle `Null` Values: Directly calls `row.getBoolean(pos)` etc., but if the field is `null` in the record, it will return default values instead of `null`.

fluss-lake/fluss-lake-hudi/src/main/java/org/apache/fluss/lake/hudi/source/HudiSortedRecordReader.java
```diff
@@ -0,0 +1,433 @@
...
+    @Override
+    public CloseableIterator<LogRecord> read() throws IOException {
+        if (delegate == null) {
+            return CloseableIterator.wrap(Collections.emptyIterator());
+        }
+        CloseableIterator<LogRecord> iterator = delegate.read();
+        List<LogRecord> records = new ArrayList<>();
+        try {
+            while (iterator.hasNext()) {
+                records.add(copyRecord(iterator.next(), sortOrder.producedTypes));
```
XuQianJin-Stars, line 96: Unnecessary Deep Copy: Deep copies all records before sorting, increasing memory and CPU overhead. Suggestion: If `HudiRecordReader` returns `LogRecord` that are already independent (not sharing internal buffers), this copy can be omitted.

fluss-lake/fluss-lake-hudi/src/main/java/org/apache/fluss/lake/hudi/source/HudiSortedRecordReader.java
```diff
@@ -0,0 +1,433 @@
...
+        }
+
+        Map<String, DataType> dataTypesByName = new HashMap<>();
+        Map<String, Integer> positionsByName = new HashMap<>();
+        List<DataType> userDataTypes = new ArrayList<>();
+        int position = 0;
+        for (Schema.UnresolvedColumn column :
+                hudiTableInfo.getHudiTable().getUnresolvedSchema().getColumns()) {
+            if (SYSTEM_COLUMNS.containsKey(column.getName())) {
+                continue;
```
XuQianJin-Stars, line 171: Position Calculation Issue in resolveRecordKeyInfo: When skipping `SYSTEM_COLUMNS`, the position calculation may be inconsistent with the actual schema position.

fluss-lake/fluss-lake-hudi/src/main/java/org/apache/fluss/lake/hudi/source/HudiLakeSource.java
```diff
@@ -27,6 +27,7 @@
 import org.apache.fluss.metadata.TablePath;
 import org.apache.fluss.predicate.Predicate;
 
+import org.apache.hudi.common.model.HoodieTableType;
 import org.apache.hudi.org.apache.avro.Schema;
 import org.apache.hudi.source.ExpressionPredicates;
```
XuQianJin-Stars, line 32: Unused Import: The constructor receives a predicates parameter but it is not used.

fluss-lake/fluss-lake-hudi/src/test/java/org/apache/fluss/lake/hudi/flink/FlinkUnionReadPrimaryKeyTableITCase.java
```diff
@@ -0,0 +1,596 @@
...
+
+    @TempDir public static File savepointDir;
+
+    @BeforeAll
+    protected static void beforeAll() {
+        FlinkUnionReadTestBase.beforeAll();
+    }
+
+    @ParameterizedTest
+    @ValueSource(booleans = {false})
```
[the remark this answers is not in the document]
fhan688: Fixed. I added the partitioned case to testUnionReadFullType and verified both parameterized runs pass.

fluss-lake/fluss-lake-hudi/src/main/java/org/apache/fluss/lake/hudi/source/HudiSortedRecordReader.java
fhan688: Good point. The current implementation sorts records within one Hudi split to satisfy the SortedRecordReader contract. Introducing spill/external sort would be a larger follow-up improvement, so I’d keep this PR focused on correctness for now.

fluss-lake/fluss-lake-hudi/src/main/java/org/apache/fluss/lake/hudi/source/HudiSortedRecordReader.java
fhan688: Fixed. I added an explicit non-null state check before reading key fields, so compareValues will not receive null key values.

fluss-lake/fluss-lake-hudi/src/main/java/org/apache/fluss/lake/hudi/source/HudiSortedRecordReader.java
fhan688: Fixed at the compareRows entry point. Since Hudi record keys map to Fluss primary key fields, null key values are invalid and now fail fast with a clear state check.

fluss-lake/fluss-lake-hudi/src/main/java/org/apache/fluss/lake/hudi/source/HudiSortedRecordReader.java
fhan688: The copy is still required here. HudiRecordReader reuses the projected row wrapper, while this reader buffers records before sorting. I added a comment to make this ownership requirement explicit.

fluss-lake/fluss-lake-hudi/src/main/java/org/apache/fluss/lake/hudi/source/HudiSortedRecordReader.java
fhan688: The behavior is intentional because project indexes are based on Fluss user fields, while Hudi system columns are appended internally and stripped from produced records. I renamed the variables to userFieldPosition/keyUserFieldPositions to make this clearer.

fluss-lake/fluss-lake-hudi/src/main/java/org/apache/fluss/lake/hudi/source/HudiLakeSource.java
fhan688: This is already used: predicates is passed into HudiRecordReader when creating the delegate reader, so I kept it unchanged.

XuQianJin-Stars (approved):
+1

fluss-lake/fluss-lake-hudi/src/main/java/org/apache/fluss/lake/hudi/source/HudiSortedRecordReader.java
```diff
@@ -0,0 +1,445 @@
...
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.stream.Collectors;
+
+import static org.apache.fluss.lake.hudi.HudiLakeCatalog.SYSTEM_COLUMNS;
+import static org.apache.fluss.utils.Preconditions.checkState;
+
+/** Sorted Hudi record reader for primary key table union read. */
+public class HudiSortedRecordReader implements SortedRecordReader {
```
luoyuxia, line 61: do we real need to implement SortedRecordReader for hudi?
To me, the cost is high since we will need to sort, which is easy to oom. It's only need for batch reading, and not required for streaming reading.

luoyuxia (commented):
@fhan688 Thanks for the pr. I only have one question. PTAL

fluss-lake/fluss-lake-hudi/src/main/java/org/apache/fluss/lake/hudi/source/HudiSortedRecordReader.java
fhan688: Thanks for raising this. I agree that the sorted reader should only be required for batch primary-key union
  read, because the current batch LakeSnapshotAndLogSplitScanner relies on SortedRecordReader for sort-merge with Fluss changelog records.

  For streaming read, sorting is not required. The current HudiLakeSource selection is too broad because
  createRecordReader cannot distinguish whether the caller needs sorted records, so streaming paths may also pay the sorting cost. I will refine the reader selection so HudiSortedRecordReader is only used when the
  batch union-read path requires sorted lake records, and keep normal HudiRecordReader for streaming reads.

luoyuxia (approved):

PR #3542, opened 2026-06-29
[lake/iceberg] Fix listener leak in IcebergLakeCommitter

Guosmilesmile:
### Purpose

Because Iceberg's `Listeners` registry is:
- **JVM-global** (a `static` `Map<Class, Queue<Listener>>`),
- **append-only** (no `deregister` / `unregister` API),

every newly created `IcebergLakeCommitter` permanently appended one more
listener to the global queue. In a long-running tiering job, the queue keeps
growing and every `CreateSnapshotEvent` is dispatched to all of them,
causing:

1. **Memory leak** – listener instances are never garbage-collected as long
   as the `Listeners` class is loaded.
2. **CPU overhead** – each Iceberg commit triggers N listener notifications,
   where N grows over time.
3. **Correctness risk for shared catalogs** – when multiple `IcebergLakeCommitter` instances live in the same JVM (e.g. multiple tiering tasks), all of their listeners write to the same `ThreadLocal<Long> currentCommitSnapshotId`. The last writer wins, but the duplicated work is still wasted.

### Brief change log

- Move listener registration from the constructor to a `static {}` block in
  `IcebergLakeCommitter`, so the listener is registered exactly **once per
  ClassLoader** (effectively once per JVM).
- Keep the `ThreadLocal#remove()` call after each commit so no thread-local
  entry is leaked across commits.
- Add `IcebergLakeCommitterListenerLeakTest` which:
  - Reflectively reads the private `Listeners.LISTENERS` registry.
  - Creates multiple `IcebergLakeCommitter` instances.
  - Asserts that the number of `IcebergSnapshotCreateListener` instances in the registry stays at exactly **1**, regardless of how many committers are constructed.

### Tests

The behavior is already coverd by existing test.

### API and Format

### Documentation


fluss-lake/fluss-lake-iceberg/src/test/java/org/apache/fluss/lake/iceberg/tiering/IcebergLakeCommitterListenerLeakTest.java
```diff
@@ -0,0 +1,142 @@
...
+import java.io.File;
+import java.lang.reflect.Field;
+import java.util.LinkedList;
+import java.util.Map;
+import java.util.Queue;
+
+import static org.apache.fluss.lake.iceberg.utils.IcebergConversions.toIceberg;
+import static org.assertj.core.api.Assertions.assertThat;
+
+class IcebergLakeCommitterListenerLeakTest {
```
luoyuxia: Can we just remove this test since it's a trival change and the behavior is already coverd by existing test

luoyuxia (commented):
@Guosmilesmile Thanks for the pr. lgtm overall. I only left one comment

fluss-lake/fluss-lake-iceberg/src/test/java/org/apache/fluss/lake/iceberg/tiering/IcebergLakeCommitterListenerLeakTest.java
Guosmilesmile: Sure, remove IcebergLakeCommitterListenerLeakTest.

luoyuxia (approved):
Thanks. LGTM

Guosmilesmile:
Thanks @luoyuxia for reviewing  !

PR #3547, opened 2026-06-29
[server] Periodically checkpoint log recovery points

### LINKED ISSUE #3546: [server] Unclean shutdown recovery scans flushed log segments without periodic recovery point checkpoints
### Search before asking

### Fluss version

main (development)

### Please describe the bug 🐞

After an unclean `TabletServer` shutdown, log recovery reads
`recovery-point-offset-checkpoint` to decide where segment recovery should
start. Today this checkpoint is mainly written during clean shutdown or
truncate paths, so after an unclean shutdown it can be stale or missing.
When that happens, recovery may scan many already-flushed segments across
all replicas, which can significantly slow TabletServer startup.

### Solution

_No response_

swuferhong:
### Purpose

Linked issue: close #3546 

Add a recovery point checkpoint task in LogManager. The task runs every 60 seconds by default and persists the latest flushed recovery point for each log directory, so unclean startup can avoid recovering already-flushed log segments.


fluss-server/src/main/java/org/apache/fluss/server/log/LogManager.java
```diff
@@ -616,6 +632,13 @@ void checkpointRecoveryOffsets(File dataDir) {
                         .collect(Collectors.toList()));
     }
 
+    @VisibleForTesting
```
LiebingYu: nit: This function is not for testing

LiebingYu (approved):
@swuferhong LGTM overall.

wuchong (approved):

PR #3548, opened 2026-06-30
[flink]Newly discovered partition read from earliest rather than scan.startup.mode

### LINKED ISSUE #3543: [flink] scan.startup.mode=latest can skip rows written to a new partition before partition discovery
### Search before asking

### Fluss version

0.9.0 (latest release)

### Please describe the bug 🐞

For a running **partitioned Fluss source** with:

  - `scan.startup.mode = 'latest'`
  - `scan.partition.discovery.interval > 0`

 rows can be silently skipped if they are written to a **new partition before that partition is discovered** by the source enumerator.

#### Expected behavior
`scan.startup.mode=latest` should skip data that already existed **when the job started**, but it should not skip rows written to a partition that is created **after**  the job is already running.

#### Actual behavior
For a newly created partition:
1. the partition is created and starts receiving rows
2. the running source does not see it until the next discovery cycle
3. when the partition is discovered, it is initialized from `latest`
4. rows already present in that partition are skipped forever

#### Reproduction
We could reproduced this with a new ITCase:
- start a partitioned source with `scan.startup.mode='latest'`
- wait for the initial discovery cycle
- add a new partition
- immediately write an early batch into that partition
- wait for the next discovery cycle
- write a late batch into the same partition

Observed result:
- only the late batch is consumed
- the early batch is skipped

#### Code path
The issue comes from reusing the same offsets initializer for both:
- partitions known at job startup
- partitions discovered later

Relevant code:
- `fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/source/FlinkTableSource.java`
- `fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/source/enumerator/FlinkSourceEnumerator.java`

`scan.startup.mode='latest'` is mapped to `OffsetsInitializer.latest()`, and newly discovered partitions are initialized through the same `startingOffsetsInitializer`,  so they also start from the partition's latest offset at discovery time.

#### Impact

This can cause silent data loss for partitioned Flink jobs reading Fluss tables with `scan.startup.mode=latest`, especially when partitions are created dynamically and the first rows arrive before the next discovery tick.

### Solution

Introduce a separate startup policy for **newly discovered partitions** instead of always reusing `scan.startup.mode`, or default `scan.new-partition.startup.mode=earliest` for **newly discovered partitions**.

loserwang1024:
### Purpose

Linked issue: close https://github.com/apache/fluss/issues/3543

### DIFF
--- a/fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/source/enumerator/FlinkSourceEnumerator.java
+++ b/fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/source/enumerator/FlinkSourceEnumerator.java
@@ -155,6 +155,22 @@ public class FlinkSourceEnumerator
     private final OffsetsInitializer startingOffsetsInitializer;
     private final OffsetsInitializer stoppingOffsetsInitializer;
 
+    /**
+     * The offsets initializer used for partitions discovered after the initial startup. Following
+     * FLIP-288 semantics, newly discovered partitions always start from earliest to prevent data
+     * loss.
+     */
+    private final OffsetsInitializer newDiscoveryOffsetsInitializer;
+
+    /**
+     * Splits whose starting offsets have been initialized but that have not yet been assigned to
+     * any reader. This map is persisted in checkpoint state (via {@link #snapshotState}) so that on
+     * failover restore these splits are directly placed into {@link #pendingSplitAssignment}
+     * without re-initialization, preserving the original offset strategy determined at first
+     * discovery time (FLIP-288).
+     */
+    private final Map<TableBucket, SourceSplitBase> unassignedSplits;
+
     private final LeaseContext leaseContext;
 
     /** checkpointId -> tableBuckets who finished consume kv snapshots. */
@@ -169,6 +185,14 @@ public class FlinkSourceEnumerator
     // This flag will be marked as true if periodically partition discovery is disabled AND the
     // split initializing has finished.
     private boolean noMoreNewSplits = false;
+
+    /**
+     * Whether the initial partition discovery has been completed. Following FLIP-288, this flag
+     * alone determines offset strategy: partitions discovered before this flag is set to {@code
+     * true} use the user-configured {@link #startingOffsetsInitializer}, while partitions
+     * discovered after use {@link #newDiscoveryOffsetsInitializer} (earliest) to prevent data loss.
+     */
+    private boolean initialDiscoveryFinished;
 
     private boolean lakeEnabled = false;
 
@@ -260,7 +284,9 @@ public class FlinkSourceEnumerator
                 partitionFilters,
                 lakeSource,
                 leaseContext,
-                checkpointTriggeredBefore);
+                checkpointTriggeredBefore,
+                false,
+                Collections.emptyList());
     }
 
     public FlinkSourceEnumerator(
@@ -278,7 +304,9 @@ public class FlinkSourceEnumerator
             @Nullable Predicate partitionFilters,
             @Nullable LakeSource<LakeSplit> lakeSource,
             LeaseContext leaseContext,
-            boolean checkpointTriggeredBefore) {
+            boolean checkpointTriggeredBefore,
+            boolean initialDiscoveryFinished,
+            List<SourceSplitBase> unassignedSplits) {
         this(
                 tablePath,
                 flussConf,
@@ -295,7 +323,9 @@ public class FlinkSourceEnumerator
                 partitionFilters,
                 lakeSource,
                 leaseContext,
-                checkpointTriggeredBefore);
+                checkpointTriggeredBefore,
+                initialDiscoveryFinished,
+                unassignedSplits);
     }
 
     public FlinkSourceEnumerator(
@@ -314,7 +344,9 @@ public class FlinkSourceEnumerator
             @Nullable Predicate partitionFilters,
             @Nullable LakeSource<LakeSplit> lakeSource,
             LeaseContext leaseContext,
-            boolean checkpointTriggeredBefore) {
+            boolean checkpointTriggeredBefore,
+            boolean initialDiscoveryFinished,
+            Collection<SourceSplitBase> unassignedSplits) {
         this(
                 tablePath,
                 flussConf,
@@ -332,7 +364,9 @@ public class FlinkSourceEnumerator
                 lakeSource,
                 new WorkerExecutor(context),
                 leaseContext,
-                checkpointTriggeredBefore);
+                checkpointTriggeredBefore,
+                initialDiscoveryFinished,
+                unassignedSplits);
     }
 
     FlinkSourceEnumerator(
@@ -369,7 +403,9 @@ public class FlinkSourceEnumerator
                 lakeSource,
                 workerExecutor,
                 leaseContext,
-                checkpointTriggeredBefore);
+                checkpointTriggeredBefore,
+                false,
+                Collections.emptyList());
     }
 
     FlinkSourceEnumerator(
@@ -389,7 +425,9 @@ public class FlinkSourceEnumerator
             @Nullable LakeSource<LakeSplit> lakeSource,
             WorkerExecutor workerExecutor,
             LeaseContext leaseContext,
-            boolean checkpointTriggeredBefore) {
+            boolean checkpointTriggeredBefore,
+            boolean initialDiscoveryFinished,
+            Collection<SourceSplitBase> unassignedSplits) {
         checkArgument(
                 splitPerAssignmentBatchSize > 0,
                 "Split assignment batch size must be positive, but was %s.",
@@ -407,6 +445,7 @@ public class FlinkSourceEnumerator
                         ? null
                         : new LinkedList<>(pendingHybridLakeFlussSplits);
         this.startingOffsetsInitializer = startingOffsetsInitializer;
+        this.newDiscoveryOffsetsInitializer = OffsetsInitializer.earliest();
         this.scanPartitionDiscoveryIntervalMs = scanPartitionDiscoveryIntervalMs;
         this.streaming = streaming;
         this.partitionFilters = partitionFilters;
@@ -417,6 +456,8 @@ public class FlinkSourceEnumerator
         this.leaseContext = leaseContext;
         this.checkpointTriggeredBefore = checkpointTriggeredBefore;
         this.splitPerAssignmentBatchSize = splitPerAssignmentBatchSize;
+        this.initialDiscoveryFinished = initialDiscoveryFinished;
+        this.unassignedSplits = indexByPartition(unassignedSplits);
     }
 
     @Override
@@ -432,6 +473,13 @@ public class FlinkSourceEnumerator
             throw new FlinkRuntimeException(
                     String.format("Failed to get table info for %s", tablePath),
                     ExceptionUtils.stripCompletionException(e));
+        }
+
+        // Find splits where the start offset has been initialized but not yet assigned to readers.
+        // These splits must not be reinitialized to keep offsets consistent with first discovery.
+        final Collection<SourceSplitBase> preinitializedSplits = unassignedSplits.values();
+        if (!unassignedSplits.isEmpty()) {
+            addSplitToPendingAssignments(preinitializedSplits);
         }
 
         if (isPartitioned) {
@@ -509,7 +557,9 @@ public class FlinkSourceEnumerator
                                 // Use log-only splits to avoid generating mixed split
                                 // types (HybridSnapshotLogSplit + LogSplit) for
                                 // primary-key tables, which is not supported.
-                                splits = this.initLogTablePartitionSplits(partitions);
+                                splits =
+                                        this.initLogTablePartitionSplits(
+                                                partitions, startingOffsetsInitializer);
                             } else {
                                 splits = this.getLogSplit(null, null);
                             }
@@ -526,6 +576,14 @@ public class FlinkSourceEnumerator
     }
 
     private void startInStreamModeForNonPartitionedTable() {
+        // If we have restored unassigned splits from checkpoint state, skip re-initialization.
+        // These splits already have their offsets resolved and will be assigned to readers
+        // when they register (via addReader -> assignPendingSplits).
+        if (!pendingSplitAssignment.isEmpty()) {
+            noMoreNewSplits = true;
+            return;
+        }
+
         if (lakeSource != null) {
             // Generate lake splits synchronously so that they are available before the
             // first checkpoint. This is consistent with the partitioned-table path in
@@ -620,8 +678,16 @@ public class FlinkSourceEnumerator
                 tablePath,
                 partitionInfos.size());
 
-        final PartitionChange partitionChange = getPartitionChange(partitionInfos);
+        final PartitionChange partitionChange =
+                getPartitionChange(partitionInfos, !initialDiscoveryFinished);
+
         if (partitionChange.isEmpty()) {
+            // No partition changes found. For the empty-table case (no initial partitions
+            // to track), mark initial discovery as finished immediately since there are
+            // no splits that need to be persisted in state first.
+            if (!initialDiscoveryFinished) {
+                initialDiscoveryFinished = true;
+            }
             LOG.debug("No partition changes detected for table {}", tablePath);
             return;
         }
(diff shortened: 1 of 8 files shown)


loserwang1024:
@beryllw @leonardBang @wuchong @swuferhong  , CC

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/source/enumerator/FlinkSourceEnumerator.java
```diff
@@ -978,6 +1100,12 @@ private void handleSplitsAdd(List<SourceSplitBase> splits, Throwable t) {
                         t);
             }
         }
+
+        initialDiscoveryFinished = true;
+        for (SourceSplitBase split : splits) {
+            unassignedSplits.put(split.getTableBucket(), split);
```
leonardBang: This stores unassigned splits by TableBucket, but TableBucket is not unique for all split types. LakeSnapshotSplit uses splitIndex to distinguish multiple lake splits in the same bucket, so checkpointing only unassignedSplits.values() can drop all but one split (or fail on restore with duplicate keys). Could we persist these by splitId or as a list instead of de-duplicating by TableBucket?

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/source/state/SourceEnumeratorState.java
```diff
@@ -44,15 +46,62 @@ public class SourceEnumeratorState {
...
             Set<TableBucket> assignedBuckets,
             Map<Long, String> assignedPartitions,
             @Nullable List<SourceSplitBase> remainingHybridLakeFlussSplits,
             String leaseId) {
+        this(
+                assignedBuckets,
+                assignedPartitions,
+                remainingHybridLakeFlussSplits,
+                leaseId,
+                false,
```
leonardBang: This default is used when restoring old V2 enumerator state, but it leaves initialDiscoveryFinished=false. After an upgrade, partitions created after that old checkpoint can be classified as initial partitions and use the user startup mode (for example latest), which reintroduces the data-loss case this patch is trying to avoid. Could we default old V2 state to true here, or otherwise make the migration preserve post-initial discovery semantics?

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/source/enumerator/FlinkSourceEnumerator.java
loserwang1024: I have refactor the unassignedSplits and add more in comment.
image

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/source/state/SourceEnumeratorState.java
loserwang1024: If false is changed to true, the initial split—which is not yet fully assigned—will be treated as "earliest", violating the intended semantics. In extreme cases, a checkpoint might be taken before any splits are distributed at all, causing the entire job to restart from the earliest offset upon recovery.

Setting it to false ensures semantic consistency: although there may be a brief period at startup during which data is read from the earliest offset, this behavior aligns with the original design, and normal processing resumes shortly afterward.

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/source/enumerator/FlinkSourceEnumerator.java
```diff
@@ -155,6 +155,45 @@ public class FlinkSourceEnumerator
     private final OffsetsInitializer startingOffsetsInitializer;
     private final OffsetsInitializer stoppingOffsetsInitializer;
 
+    /**
+     * The offsets initializer used for partitions discovered after the initial startup. Following
+     * FLIP-288 semantics, newly discovered partitions always start from earliest to prevent data
```
leonardBang: Could you also post FLIP-288 url here to help reader to understand the background?

loserwang1024:
@leonardBang @wuchong @swuferhong @luoyuxia , CC, please help review

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/source/enumerator/FlinkSourceEnumerator.java
leonardBang, line 1106: `handlePartitionsRemoved()` removes dropped-partition splits only from `pendingSplitAssignment`, but those same splits are also added to `unassignedSplits` in `handleSplitsAdd()` and checkpointed in `snapshotState()`.

If a partition is dropped after split initialization but before reader registration, the pending split is removed but remains in `unassignedSplits`; after failover restore, it can be assigned again for a deleted partition.

Could we reuse the same removal predicate for both `pendingSplitAssignment` and `unassignedSplits.removeIf(...)`, and add a snapshot/restore regression test for this path?

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/source/state/FlussSourceEnumeratorStateSerializer.java
```diff
@@ -229,13 +242,69 @@ private SourceEnumeratorState deserializeV2(byte[] serialized) throws IOExceptio
 
         // deserialize lease context
         LeaseContext leaseContext = deserializeLeaseId(in);
+        // V2 does not have initialDiscoveryFinished flag; default to true (safe choice
+        // to prevent data loss: treat all newly discovered partitions as post-initial).
```
leonardBang, line 262: The comment says V2 restore defaults `initialDiscoveryFinished` to `true`, but the implementation calls the 4-arg `SourceEnumeratorState` constructor, whose default is `false`.

This means that after upgrading from an old V2 checkpoint, the first discovered partitions after restore are still treated as initial partitions and use the user-configured `scan.startup.mode`; with `latest`, data written to partitions created after the checkpoint may still be skipped.

Could we either explicitly pass `true` here, or update the comment/tests to document that this compatibility behavior is intentional? From the data-loss-prevention perspective, defaulting to `true` is safer.

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/source/state/FlussSourceEnumeratorStateSerializer.java
```diff
@@ -70,11 +71,13 @@ public class FlussSourceEnumeratorStateSerializer
     private static final int VERSION_0 = 0;
     private static final int VERSION_1 = 1;
     private static final int VERSION_2 = 2;
+    private static final int VERSION_3 = 3;
+    private static final int VERSION_4 = 4;
```
leonardBang: Do we need both `VERSION_3` and `VERSION_4` in this PR? Since neither format has been released yet, it seems enough to introduce a single new serializer version on top of V2, containing both `initialDiscoveryFinished` and `unassignedSplits`.

fluss-flink/fluss-flink-common/src/test/java/org/apache/fluss/flink/source/state/SourceEnumeratorStateSerializerTest.java
```diff
@@ -191,4 +191,104 @@ void testEmptyPendingSplitsCheckpointSerdeWithoutLakeSource() throws Exception {
...
+
+        SourceEnumeratorState deserialized = serializer.deserialize(2, v2Bytes);
+
+        assertThat(deserialized.getAssignedBuckets()).isEqualTo(assignedBuckets);
+        assertThat(deserialized.getAssignedPartitions()).isEqualTo(assignedPartitions);
+        assertThat(deserialized.getLeaseId()).isEqualTo("v2LeaseId");
+        // V2 uses the 4-param constructor which defaults initialDiscoveryFinished=false.
+        // On restore, the first discovery round will treat all discovered partitions
+        // as initial partitions (safe behavior).
+        assertThat(deserialized.isInitialDiscoveryFinished()).isFalse();
```
leonardBang: This V2 compatibility test does not actually produce V2 bytes: the current serializer is V4, but the test only truncates one byte from the current payload.

It passes because the V2 deserializer ignores trailing bytes, not because real V2 compatibility is verified. Could we add a `serializeV2` test helper or hand-build a real V2 payload?

leonardBang (commented):
Thanks @loserwang1024 for the contribution, I left some comments

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/source/state/FlussSourceEnumeratorStateSerializer.java
loserwang1024: my mistake.

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/source/state/FlussSourceEnumeratorStateSerializer.java
loserwang1024: done, either explicitly pass true now.

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/source/enumerator/FlinkSourceEnumerator.java
```diff
@@ -155,6 +155,46 @@ public class FlinkSourceEnumerator
     private final OffsetsInitializer startingOffsetsInitializer;
     private final OffsetsInitializer stoppingOffsetsInitializer;
 
+    /**
+     * The offsets initializer used for partitions discovered after the initial startup. Following
+     * FLIP-288(<a
+     * href="https://cwiki.apache.org/confluence/spaces/FLINK/pages/350784113/FLIP-529+Connections+in+Flink+SQL+and+TableAPI">...</a>)
```
leonardBang: ？

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/source/state/FlussSourceEnumeratorStateSerializer.java
leonardBang: plz also update the java doc.

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/source/state/FlussSourceEnumeratorStateSerializer.java
```diff
@@ -99,6 +101,12 @@ public byte[] serialize(SourceEnumeratorState state) throws IOException {
         // write lease context
         serializeLeaseId(out, state);
 
+        // write initial discovery finished flag (VERSION_3+)
+        out.writeBoolean(state.isInitialDiscoveryFinished());
+
+        // write unassigned splits (VERSION_4)
```
leonardBang: update `VERSION_3+` `version_4`？

leonardBang (commented):
Thanks @loserwang1024 for the update, I only left some doc comments

leonardBang (approved):
+1 from my side

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/source/enumerator/FlinkSourceEnumerator.java
```diff
@@ -978,6 +1170,26 @@ private void handleSplitsAdd(List<SourceSplitBase> splits, Throwable t) {
                         t);
             }
         }
+
+        initialDiscoveryFinished = true;
+        if (pendingHybridLakeFlussSplits != null) {
+            // removed from the pendingHybridLakeFlussSplits since this split already be moved to
+            // unassignedSplits
+            pendingHybridLakeFlussSplits.removeAll(splits);
```
wuchong, line 1178: When restoring pending lake splits from a checkpoint, the splits are inadvertently discarded. Specifically, `generateHybridLakeFlussSplits()` returns the field itself when `pendingHybridLakeFlussSplits` already exists (https://github.com/apache/fluss/blob/26644daacd4f9a371e946cce451f130fd043e8e7/fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/source/enumerator/FlinkSourceEnumerator.java#L1061). Subsequently, `handleSplitsAdd()` executes `pendingHybridLakeFlussSplits.removeAll(splits)`. If both references point to the same list, this operation clears it entirely. As a result, the subsequent `unassignedSplits.addAll(splits)` adds nothing, and the restored lake splits are never reassigned to readers.

Suggestion: return a copy for https://github.com/apache/fluss/blob/26644daacd4f9a371e946cce451f130fd043e8e7/fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/source/enumerator/FlinkSourceEnumerator.java#L1061. Or, do a deep copy at the beginning of `org.apache.fluss.flink.source.enumerator.FlinkSourceEnumerator#handleSplitsAdd` for `splits`

fluss-flink/fluss-flink-common/src/test/java/org/apache/fluss/flink/source/state/SourceEnumeratorStateSerializerTest.java
```diff
@@ -144,6 +144,32 @@ void testV0Compatibility() throws Exception {
...
+
+        // Use serializeV2 to produce a real V2 payload
+        byte[] v2Bytes = serializer.serializeV2(originalState);
+
+        SourceEnumeratorState deserialized = serializer.deserialize(2, v2Bytes);
+        assertThat(deserialized.getAssignedBuckets()).isEqualTo(assignedBuckets);
+        assertThat(deserialized.getAssignedPartitions()).isEqualTo(assignedPartitions);
+        assertThat(deserialized.getLeaseId()).isEqualTo("v2LeaseId");
+        assertThat(deserialized.isInitialDiscoveryFinished()).isTrue();
+        assertThat(deserialized.getUnassignedSplits()).isEmpty();
```
wuchong, line 170: Add tests to verify the serde of unassigned splits.

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/source/enumerator/FlinkSourceEnumerator.java
```diff
@@ -407,6 +469,7 @@ public FlinkSourceEnumerator(
                         ? null
                         : new LinkedList<>(pendingHybridLakeFlussSplits);
         this.startingOffsetsInitializer = startingOffsetsInitializer;
+        this.newDiscoveryOffsetsInitializer = OffsetsInitializer.earliest();
```
wuchong, line 472: Currently, this is hard-coded. What if `startingOffsetsInitializer` is timestamp-based, such as a future timestamp? In that case, should we use the timestamp from `startingOffsetsInitializer` instead?

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/source/enumerator/FlinkSourceEnumerator.java
```diff
@@ -526,6 +621,19 @@ private void startInBatchMode() {
...
+        // If we have restored unassigned splits from checkpoint state, skip re-initialization.
+        // These splits already have their offsets resolved and will be assigned to readers
+        // when they register (via addReader -> assignPendingSplits).
+        if (!pendingSplitAssignment.isEmpty()) {
+            LOG.info(
+                    "Skipping split re-initialization for non-partitioned table {}: "
+                            + "{} splits already restored from checkpoint state.",
+                    tablePath,
+                    pendingSplitAssignment.values().stream().mapToInt(List::size).sum());
+            noMoreNewSplits = true;
```
wuchong: Why set `noMoreNewSplits` to true here? Do you mean to set `initialDiscoveryFinished` to true here?

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/source/enumerator/FlinkSourceEnumerator.java
loserwang1024: when timestamp is large than current timestamp, server will throw exception:
```java
if (fetchTimestamp > currentTimestamp) {
            throw new InvalidTimestampException(
                    String.format(
                            "Get offset error for table bucket %s, "
                                    + "the fetch timestamp %s is larger than the current timestamp %s",
                            tableBucket, fetchTimestamp, currentTimestamp));
        }
```

PR #3550, opened 2026-06-30
[server] Set table_id on prepareLakeTableSnapshot error response

### LINKED ISSUE #3551: [server] prepareLakeTableSnapshot masks the real error on failure
### Search before asking

### Fluss version

main (development)

### Please describe the bug 🐞

When CoordinatorService.prepareLakeTableSnapshot fails, it sets the error code on the response but not table_id, and doesn't log the exception. 

The tiering committer reads table_id before the error code, so it dies with IllegalStateException: Field 'table_id' is not set, hiding the real failure.

### Solution

_No response_

fresh-borzoni:
closes https://github.com/apache/fluss/issues/3551

When anything in the per-table loop of `CoordinatorService.prepareLakeTableSnapshot` fails, the catch set `error_code`/`error_message` but **not** `table_id`, and didn't log the exception. 
The tiering committer reads `table_id` before checking the error, so it crashed with Field 'table_id' is not set` and
masks the real failure and no trace in logs

This sets `table_id` on the error response and logs the exception, so the real error surfaces
both to the client and in the coordinator logs.


luoyuxia (approved):
@fresh-borzoni Thanks. LGTM

PR #3556, opened 2026-07-02
[server] Require new leader ack for leader-only rebalance

### LINKED ISSUE #3555: [server] Leader-only rebalance may finish before new leader acknowledges NotifyLeaderAndIsr
### Search before asking

### Problem

For leader-only rebalance, the coordinator is expected to execute leader migrations sequentially: the next task should not start until the current target leader has applied `NotifyLeaderAndIsr`.

However, the current completion check can finish a leader-only rebalance task too early. `NotifyLeaderAndIsr` is sent to multiple replicas, and the coordinator may complete the task after receiving a successful response from a replica that is not the new leader, as long as the ZooKeeper leader has already changed to the target leader.

This means the rebalance task can be marked completed before the new leader has acknowledged the request. For KV tables, this is risky because the leader-side KV state may not be ready yet.

### Expected Behavior

For leader-only rebalance, a task should only be completed when:

- the `NotifyLeaderAndIsr` result for the bucket succeeded
- the response comes from the target new leader
- the current leader is the target new leader

In other words, completion should require `result.succeeded()` and `responseServerId == newLeader`.

### Root Cause

The leader-only rebalance completion path only checked whether the current leader matched the target leader. It did not consider which tablet server returned the `NotifyLeaderAndIsr` response, nor whether the bucket-level result succeeded.

### Willingness to contribute

swuferhong:
### Purpose

Linked issue: close #3555 

Leader-only rebalance previously completed once the current leader matched
the target leader after any `NotifyLeaderAndIsr` response for the bucket. Since
`NotifyLeaderAndIsr` is sent to multiple replicas, this could complete the task
before the new leader acknowledged the request.

Pass the bucket result and response server id into the completion check, and
only complete leader-only rebalance when the result succeeded and the response
came from the target new leader. Keep LeaderAndIsr change handling from
directly completing leader-only rebalance tasks.

### DIFF
--- a/fluss-server/src/main/java/org/apache/fluss/server/coordinator/CoordinatorEventProcessor.java
+++ b/fluss-server/src/main/java/org/apache/fluss/server/coordinator/CoordinatorEventProcessor.java
@@ -1612,13 +1612,13 @@ public class CoordinatorEventProcessor implements EventProcessor {
                             planForBucket.getOriginReplicas(), planForBucket.getNewReplicas());
             try {
                 if (planForBucket.isLeaderChanged() && !reassignment.isBeingReassigned()) {
+                    if (!isSuccessfulLeaderOnlyRebalanceResponseFromNewLeader(
+                            notifyLeaderAndIsrResultForBucket, responseServerId, planForBucket)) {
+                        return;
+                    }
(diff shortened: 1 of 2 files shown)


swuferhong:
@wuchong copilot's comments addressed. PTAL, thanks.

wuchong (approved):

PR #3558, opened 2026-07-02
[server] Auto partition day format supporting yyyy-MM-dd

### LINKED ISSUE #3557: [server] Support auto-partition time format like "yyyy-MM-dd"
### Search before asking

## Motivation

Today, auto-partitioned tables with `table.auto-partition.time-unit=DAY` always use `yyyyMMdd` as the partition value format.

This is limiting in cases where users want DAY partitions to be organized in a more readable and conventional date form such as `yyyy-MM-dd`. It also aligns better with typed `DATE` partition values, which are naturally represented in dashed form.

### Solution

Introduce a new table option:
 - `table.auto-partition.time-format`

Supported values:
  - `yyyyMMdd` (default, backward compatible)
  - `yyyy-MM-dd`

Behavior:
  - this option only applies when `table.auto-partition.time-unit=DAY`
  - when set to `yyyy-MM-dd`, all DAY auto-partition behavior
should use dashed format consistently:
    - auto pre-created partitions
    - manual partition creation validation
    - writer-side dynamic partition creation
    - retention / expiration logic

### Anything else?

_No response_

### Willingness to contribute

gyang94:
### Purpose

Linked issue: close #3557 

### Brief change log

 Introduce a new table option:

  - `table.auto-partition.day-format`

  Supported values:

  - `yyyyMMdd` (default, backward compatible)
  - `yyyy-MM-dd` (new)


fluss-server/src/main/java/org/apache/fluss/server/utils/TableDescriptorValidation.java
```diff
@@ -518,6 +554,25 @@ private static void checkPartition(
...
+                    DataType autoPartitionDataType =
+                            rowType.getTypeAt(rowType.getFieldIndex(autoPartitionKey));
+                    if (autoPartitionDataType.getTypeRoot() == DataTypeRoot.DATE
+                            && autoPartition.dayFormat() != AutoPartitionDayFormat.YYYY_MM_DD) {
+                        throw new InvalidTableException(
+                                String.format(
+                                        "Table property '%s' must be '%s' when auto partition key '%s' has DATE type.",
+                                        ConfigOptions.TABLE_AUTO_PARTITION_DAY_FORMAT.key(),
+                                        AutoPartitionDayFormat.YYYY_MM_DD.pattern(),
+                                        autoPartitionKey));
```
wuchong: Why don't we allow other date formats for date-type partitions?

fluss-server/src/main/java/org/apache/fluss/server/utils/TableDescriptorValidation.java
```diff
@@ -458,9 +472,31 @@ private static void checkTieredLog(Configuration tableConf) {
...
+            Configuration tableConf,
+            List<String> partitionKeys,
+            RowType rowType,
+            boolean enforceDateDayFormatCompatibility) {
         boolean isPartitioned = !partitionKeys.isEmpty();
         AutoPartitionStrategy autoPartition = AutoPartitionStrategy.from(tableConf);
+        boolean hasExplicitDayFormat =
+                tableConf.contains(ConfigOptions.TABLE_AUTO_PARTITION_DAY_FORMAT);
+
+        if (hasExplicitDayFormat && !autoPartition.isAutoPartitionEnabled()) {
```
wuchong: **Allow the day format to be retained when auto partitioning is disabled**

For a table configured with `table.auto-partition.day-format=yyyy-MM-dd`, executing `ALTER TABLE ... SET table.auto-partition.enabled=false` keeps the persisted day-format property, so this check throws `InvalidTableException`. Since the day-format option itself cannot be altered, users have no valid way to disable auto partitioning.

Please allow the format to be retained when auto partitioning is disabled, so it can still be used to automatically validate the partition-name format when partitions are created.

fluss-common/src/main/java/org/apache/fluss/config/ConfigOptions.java
```diff
@@ -1604,14 +1583,24 @@ public class ConfigOptions {
...
+                                    + "auto created is yyyyMMdd by default. "
                                     + "If the value is `MONTH`, the partition format for "
                                     + "auto created is yyyyMM. "
                                     + "If the value is `QUARTER`, the partition format for "
                                     + "auto created is yyyyQ. "
                                     + "If the value is `YEAR`, the partition format for "
                                     + "auto created is yyyy.");
 
+    public static final ConfigOption<String> TABLE_AUTO_PARTITION_DAY_FORMAT =
+            key("table.auto-partition.day-format")
```
wuchong: Would it make sense to generalize this option to `table.auto-partition.time-format` instead of introducing a DAY-specific public key? The formatting path already handles `HOUR`, `DAY`, `MONTH`, `QUARTER`, and `YEAR`. Once `day-format` becomes public, supporting formats such as `yyyy-MM` or `yyyy-MM-dd-HH` later would require additional per-unit options or deprecating this key.

If we generalize it, however, I don't think we should accept arbitrary `DateTimeFormatter` patterns without validation. Partition retention currently relies on the lexicographical ordering of formatted partition values through `compareTo()` and `TreeMap.headMap()`, and the format must preserve the configured granularity. For example, an HOUR format without `HH` would cause multiple hours to map to the same partition, while `MM-yyyy` or non-zero-padded fields would not sort chronologically.

Could we introduce `table.auto-partition.time-format`, use the existing unit-specific formats when the option is absent for backward compatibility, and validate supported formats per time unit so that they are fixed-width, ordered from the most significant field to the least significant field, and include the required time-unit field?

Example: https://help.aliyun.com/zh/hologres/developer-reference/dynamic-partition-management?spm=a2c4g.11186623.help-menu-113622.d_5_0_5_2_3.613f2519QaKdYr

fluss-server/src/main/java/org/apache/fluss/server/utils/TableDescriptorValidation.java
```diff
@@ -88,6 +90,14 @@ public static void validateTableDescriptor(
...
             int maxBucketNum,
             @Nullable DataLakeFormat clusterDataLakeFormat) {
+        validateTableDescriptor(tableDescriptor, maxBucketNum, clusterDataLakeFormat, true);
+    }
+
+    public static void validateTableDescriptor(
+            TableDescriptor tableDescriptor,
+            int maxBucketNum,
+            @Nullable DataLakeFormat clusterDataLakeFormat,
+            boolean enforceDateDayFormatCompatibility) {
```
wuchong: Why introducing this flag? It seems this variable is always `true`?

wuchong:
Btw, please also update the auto partition documentation (https://fluss.apache.org/docs/table-design/data-distribution/partitioning/#auto-partitioning) and the configuration page (https://fluss.apache.org/docs/engine-flink/options/#storage-options).

fluss-server/src/main/java/org/apache/fluss/server/utils/TableDescriptorValidation.java
gyang94: When we convert a DATE type value to a partition name( in 'string' type), we have a hidden assumption that it is in the 'yyyy-MM-dd' fomat. See `PartitionNameConverters.dayToString()` and `PartitionNameConverters.parseDayString()`.

```
public static String dayToString(int days) {
    OffsetDateTime day = EPOCH.plusDays(days);
    return String.format(
            Locale.ROOT,
            "%04d-%02d-%02d",
            day.getYear(),
            day.getMonth().getValue(),
            day.getDayOfMonth());
}

public static int parseDayString(String dayStr) {
    String[] parts = dayStr.split("-");
    int year = Integer.parseInt(parts[0]);
    int month = Integer.parseInt(parts[1]);
    int day = Integer.parseInt(parts[2]);
    LocalDateTime date = LocalDateTime.of(year, month, day, 0, 0);
    long epochDay = date.toLocalDate().toEpochDay();
    return (int) epochDay;
}
```
Thus if our auto partition is not in `yyyy-MM-dd`, the partition name of a row data will not match the auto partition name.

fluss-common/src/main/java/org/apache/fluss/utils/PartitionUtils.java
```diff
@@ -171,39 +183,225 @@ public static ResolvedPartitionSpec generateAutoPartition(
...
+        validateTimeFields(timeFormat, timeUnit);
+    }
+
+    private static void validateTimeFields(String timeFormat, AutoPartitionTimeUnit timeUnit) {
+        List<TimeFormatField> expectedFields = getExpectedTimeFormatFields(timeUnit);
+        List<TimeFormatField> actualFields = new ArrayList<>();
+        boolean inLiteral = false;
+        for (int index = 0; index < timeFormat.length(); ) {
+            char patternChar = timeFormat.charAt(index);
+            if (patternChar == '\'') {
```
swuferhong, line 301: The custom time format should also be validated against Fluss partition-name restrictions.

For example, `yyyy/MM/dd` currently passes `validateTimeFormat()` and generates a value such as `2026/07/16`, but `/` is not allowed in Fluss partition values (the file system restrictions). As a result, the table can be created successfully, but automatic partition creation will fail later.

Formats producing overlong values or values with reserved prefixes may have the same problem. Could we format a representative timestamp here and validate the result using the existing partition-name validation rules?

website/docs/table-design/data-distribution/partitioning.md
```diff
@@ -74,7 +75,17 @@ The time unit for the automatic partition table `auto-partition.time-unit` can t
...
 | YEAR      | yyyy             | 2024       |
-	
+
+The default format can be overridden with `table.auto-partition.time-format`. Custom formats must satisfy all of the following rules:
+
+- Time fields are ordered from the largest to the smallest unit.
+- Every field that affects the configured partition granularity is present.
+- Numeric fields use a fixed width: four digits for year, two for month, day, and hour, and one for quarter.
+
+For example, `yyyy-MM-dd` and `yyyy/MM/dd` are valid DAY formats. `MM-yyyy` has the wrong field order, `yyyy-MM` omits the day, and `yyyy-M-dd` uses a variable-width month, so they are rejected. Literal text can be quoted, as in `yyyy-'Q'Q` for QUARTER. When the auto-partition key has DATE type, the time unit must be DAY and the format must be `yyyy-MM-dd`.
```
swuferhong: `yyyy/MM/dd` is a invalid format?

fluss-common/src/main/java/org/apache/fluss/utils/PartitionUtils.java
```diff
@@ -171,39 +183,225 @@ public static ResolvedPartitionSpec generateAutoPartition(
...
+            throw new IllegalArgumentException(
+                    String.format("Invalid date-time format pattern '%s'.", timeFormat), e);
+        }
+        validateTimeFields(timeFormat, timeUnit);
+    }
+
+    private static void validateTimeFields(String timeFormat, AutoPartitionTimeUnit timeUnit) {
+        List<TimeFormatField> expectedFields = getExpectedTimeFormatFields(timeUnit);
+        List<TimeFormatField> actualFields = new ArrayList<>();
+        boolean inLiteral = false;
```
swuferhong, line 298: Could we reject optional sections (`[` and `]`) in custom time formats?

For example, `yyyy[-MM[-dd]]` currently passes the field validation because it contains year, month, and day fields. However, `DateTimeFormatter` also accepts `2026` and `2026-07` with this pattern.

This means a DAY-partitioned table may accept partition names that do not identify a specific day. It also breaks the fixed-width assumption used by partition retention ordering. Add a test to verify it.

swuferhong (commented):
Hi, @gyang94, thanks for your contributions. I left some comments:

wuchong (approved):
LGTM.

PR #3591, opened 2026-07-06
[flink] Clean KV files and buckets with confirmed empty references

### LINKED ISSUE #3489: Support safe cleanup of orphaned KV shared SST files
### Search before asking

### Description

Shared SST files are not owned by a single snapshot directory. They may be reused by multiple retained snapshots and still-in-use snapshots. A client-side cleanup action cannot safely decide that a shared SST is orphaned by only checking currently active snapshot IDs or currently visible shared file handles.

The main correctness challenge is concurrency with snapshot lifecycle:

- a snapshot may have uploaded shared SST files but not committed yet
- a snapshot commit result may be temporarily uncertain
- active refs fetched by the cleanup action are only a point-in-time view
- object-store listing and deletion are not protected by the snapshot lifecycle lock

Deleting a shared SST from the cleanup action can therefore race with an in-flight snapshot and cause a later committed snapshot to reference a deleted SST file.

Shared SST cleanup should be implemented with server-side snapshot lifecycle coordination, or with an equivalent protocol that can protect in-flight snapshots.

### Willingness to contribute

platinumhamburg:
### Purpose

Closes #3489. Clean orphan KV files and log buckets while retaining files referenced by active metadata.

### Changes

- Resolve shared SST references from snapshot metadata remote paths, including UUID object names.
- Recognize uploaded UUID files in private snapshot directories.
- Scan log buckets with no committed remote manifest and KV buckets with a confirmed empty active snapshot set.
- Require complete active references; refresh the active snapshot view once when metadata disappears and retain files if metadata or the refresh cannot be resolved.
- Coordinate targets that disappear during enumeration and retain files with unavailable modification times or at the cutoff boundary.

### API and Format

Changes are confined to the Flink action module. No public API, snapshot JSON, wire format, or action option changes.

### Validation

92 orphan focused tests and 19 Flink 1.20 Action integration tests passed with no failures or skips. JDK 11 reactor package, Spotless and Checkstyle passed.


leekeiabstraction:
@platinumhamburg Do you want to include this in 1.0 release?

PR #3594, opened 2026-07-06
[test] Raise disk write-limit ratio in FlussClusterExtension to avoid test flakiness

### LINKED ISSUE #3593: Raise disk write-limit ratio in FlussClusterExtension to avoid test flakiness
### Search before asking

### Motivation

The tablet server rejects writes once data-disk usage exceeds
server.data-disk.write-limit-ratio (default 0.85). On CI machines or
local dev boxes whose disk usage is already above 85%, tests that write
data through FlussClusterExtension fail with disk-write-limited errors
that are unrelated to the code under test.

### Solution

_No response_

### Anything else?

_No response_

### Willingness to contribute

xuyangzhong:
### Purpose

Linked issue: close #3593

The tablet server rejects writes once the data-disk usage exceeds
`server.data-disk.write-limit-ratio` (default `0.85`), and only resumes after
usage drops below `ratio - 0.10`. `FlussClusterExtension` inherits this default,
so tests that write data through it fail with disk-write-limited errors on CI
runners or local machines whose disk usage is already above 85%. These failures
are unrelated to the code under test and make the suite flaky and
environment-dependent.

This change makes the shared test cluster ignore disk-usage protection so tests
no longer depend on the host's free disk space.

### Brief change log

- Set `server.data-disk.write-limit-ratio` to `0.99` in `FlussClusterExtension`
  so the disk-usage write protection effectively stays out of the way during
  tests, while still guarding against a truly full disk.
- Fix the import ordering of `org.apache.curator.test.TestingServer`.

### Tests

No new tests. This only adjusts test infrastructure. Existing tests based on
`FlussClusterExtension` continue to pass and are no longer affected by the host
disk-usage limit.

### API and Format

No. This change is limited to test utilities and does not affect any public API
or storage format.

### Documentation

No. This is a test-only change and introduces no new feature.

### DIFF
--- a/fluss-server/src/test/java/org/apache/fluss/server/testutils/FlussClusterExtension.java
+++ b/fluss-server/src/test/java/org/apache/fluss/server/testutils/FlussClusterExtension.java
@@ -1006,6 +1006,8 @@ public final class FlussClusterExtension
             clusterConf.set(
                     ConfigOptions.COORDINATOR_LIFECYCLE_THROTTLER_TIMEOUT_CHECK_INTERVAL,
                     Duration.ofSeconds(1));
+            // Set a high data disk write limit ratio to avoid disk write limit when testing
+            clusterConf.set(ConfigOptions.SERVER_DATA_DISK_WRITE_LIMIT_RATIO, 0.99);
         }
 
         /** Sets the number of tablet servers. */


swuferhong (approved):
LGTM+1

PR #3597, opened 2026-07-06
[server] Limit max bucket num per table partition

swuferhong:
### Purpose

Linked issue: close #xxx

Make `max.bucket.num` apply to non-partitioned tables and each partition
of partitioned tables, instead of the total bucket count of a
partitioned table. Set the default bucket limit to 4096 while keeping
`max.partition.num` at 2000.


fresh-borzoni (approved):
@swuferhong LGTM 👍 
I'll rebase and merge once CI/CD passes (with elixir fixes)

PR #3598, opened 2026-07-06
[ci] Cache only elixir deps, not _build, to fix stale NIF load

### LINKED ISSUE #3599: [ci] Elixir integration flaky: fluss_nif.so fails to load
### Search before asking

### Description

Failed to load NIF library: '.../fluss_nif.so: cannot open shared object file: No such file or directory' (UndefinedFunctionError) function Fluss.Native.table_descriptor_new/2 is undefined

It's not deterministic as the same code passes on one run and fails on another (e.g. green at review, red later on an unrelated PR).

The cause is caching: the job caches Mix's _build, and when that cache is restored, mix compile thinks the app is already built and skips the rustler step, so fluss_nif.so never gets (re)placed, and the tests can't load Fluss.Native. The heavy native build is already covered by the Rust target/ cache, so caching _build on top of it just causes this.

### Willingness to contribute

fresh-borzoni:
closes https://github.com/apache/fluss/issues/3599

Fixes the recurring `elixir-integration` CI failure where the Rust NIF fails to load:

```
Failed to load NIF library: '.../fluss_nif.so: cannot open shared object file: No such file or directory'
(UndefinedFunctionError) function Fluss.Native.table_descriptor_new/2 is undefined (module Fluss.Native is not available)
```

### DIFF
--- a/.github/workflows/client-integration.yml
+++ b/.github/workflows/client-integration.yml
@@ -84,6 +84,8 @@ jobs:
           fi
           echo "Changed files:"; echo "$changed"
           has() { echo "$changed" | grep -qE "$1"; }
+          # a change to this workflow itself should exercise every suite
+          has '^\.github/workflows/client-integration\.yml' && all=true || true
           protocol=false; core=false; py=false; cpp=false; ex=false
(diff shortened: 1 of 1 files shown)


fresh-borzoni:
@swuferhong this should help you, PTAL

nicolazar (approved):

PR #3601, opened 2026-07-06
[lake]Tolerate Iceberg lake table existent if the schema and properties matches

Guosmilesmile:
### Purpose
Since we have support this feature in paimon and hudi, I think we should support it in iceberg too.

### DIFF
--- a/fluss-lake/fluss-lake-iceberg/src/main/java/org/apache/fluss/lake/iceberg/IcebergLakeCatalog.java
+++ b/fluss-lake/fluss-lake-iceberg/src/main/java/org/apache/fluss/lake/iceberg/IcebergLakeCatalog.java
@@ -48,11 +48,14 @@ import org.apache.iceberg.exceptions.AlreadyExistsException;
 import org.apache.iceberg.exceptions.NoSuchNamespaceException;
 import org.apache.iceberg.exceptions.NoSuchTableException;
 import org.apache.iceberg.types.Type;
+import org.apache.iceberg.types.TypeUtil;
+import org.apache.iceberg.types.Types;
 
 import javax.annotation.Nullable;
 
 import java.util.ArrayList;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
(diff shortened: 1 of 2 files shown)


luoyuxia (dismissed):
@Guosmilesmile Thanks. Could you please help rebase main branch

fluss-lake/fluss-lake-iceberg/src/main/java/org/apache/fluss/lake/iceberg/IcebergLakeCatalog.java
```diff
@@ -293,14 +310,189 @@ private TableIdentifier toIcebergTableIdentifier(TablePath tablePath) {
...
+            Types.NestedField existing = existingFields.get(i);
+            Types.NestedField expected = newFields.get(i);
+            if (!existing.name().equals(expected.name())
+                    || !existing.type().equals(expected.type())
+                    || existing.isOptional() != expected.isOptional()) {
+                return false;
+            }
+        }
+
+        return true;
```
luoyuxia: Blocking: This compatibility check ignores Schema.identifierFieldIds(). A pre-created Iceberg table with the same columns, partition spec, sort order, and properties but no identifier fields therefore passes this check. IcebergLakeWriter#createRecordWriter then sees an empty identifier set and selects AppendOnlyTaskWriter, so UPDATE_BEFORE and DELETE records are appended instead of emitted as equality deletes, breaking primary-key semantics. Please compare identifier fields by resolved column names (raw IDs can differ between schemas) and add coverage for missing or incorrect identifiers.

Guosmilesmile: Good Catch. I will add  compare identifier fields.

Guosmilesmile:
CI error since lake Test meet 60min timeout.

Guosmilesmile:
Raise a pr to split old flink module to single Test module。https://github.com/apache/fluss/pull/3692

luoyuxia (approved):
tt

PR #3603, opened 2026-07-07
[cpp] Encode CreateTable column types losslessly

fresh-borzoni:
Migrated from flus-rust repo
closes https://github.com/apache/fluss-rust/issues/637

Creating a table with a nested MAP/ROW column went through an Arrow schema, which affected timestamp/time precision, dropped column comments, and turned CHAR(n) into STRING. So it's better to fix it and use native converter and not Arrow schema's one.

Now the column types go to the server natively and are stored exactly as declared.

### DIFF
--- a/fluss-rust/bindings/cpp/include/fluss.hpp
+++ b/fluss-rust/bindings/cpp/include/fluss.hpp
@@ -304,7 +304,7 @@ class DataType {
     static DataType String() { return DataType(TypeId::String); }
     static DataType Bytes() { return DataType(TypeId::Bytes); }
     static DataType Date() { return DataType(TypeId::Date); }
-    static DataType Time() { return DataType(TypeId::Time); }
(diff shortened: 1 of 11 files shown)


fresh-borzoni:
@charlesdong1991 Thank you for the review in fluss-rust, let's continue here then
Addressing comments, PTAL

> does it hold true for Time? i think precision will change still for time?

Good catch — it was actually lossy, not just docs. Fixed: Time now carries precision end-to-end, so TIME(p) round-trips like TIMESTAMP(p). Added a test covering it.
> so we will drop description? i guess it is intended?

Yes, it's better to leave it as follow-up(a lot of plumbing), also python also doesn't propagate for nested fields in ROW. Created an issue: https://github.com/apache/fluss/issues/3604
> Nit: since the targets are precision, comment and char, can we add an integration test that create a table with those type, and then assert get-table-info returns exactly those values?

Added a test,  good call 👍

charlesdong1991 (approved):
Nice!! 👍

PR #3606, opened 2026-07-07
[client] Add log read preference for remote-first fetch

### LINKED ISSUE #3607: Introduce log read preference for remote-first tiering reads
### Search before asking

### Motivation

The motivation of this PR is to reduce TabletServer RPC and local disk pressure when fetching high-throughput logs.

Based on observations from production workloads, when ingestion throughput reaches around 20 GB/s, log fetch traffic from the tiering service can reach about 30 GB/s. This puts significant additional pressure on TabletServers, especially on RPC handling and local disk reads.

### Solution

_No response_

### Anything else?

_No response_

### Willingness to contribute

zuston:
### Purpose

The motivation of this PR is to reduce TabletServer RPC and local disk pressure when fetching high-throughput logs.

Based on observations from production workloads, when ingestion throughput reaches around 20 GB/s, log fetch traffic from the tiering service can reach about 30 GB/s. This puts significant additional pressure on TabletServers, especially on RPC handling and local disk reads.

link the issue of #3607 

### Brief change log

To address this, this PR makes it possible for tiering reads to fetch log data directly from remote storage when the requested offsets are already covered by remote logs.

### DIFF
--- a/fluss-common/src/main/java/org/apache/fluss/rpc/protocol/FetchLogReadPreference.java
+++ b/fluss-common/src/main/java/org/apache/fluss/rpc/protocol/FetchLogReadPreference.java
@@ -17,14 +17,13 @@
 
 package org.apache.fluss.rpc.protocol;
 
-import org.apache.fluss.annotation.Internal;
+import org.apache.fluss.annotation.PublicEvolving;
 
 /** Read preference for fetch log requests. */
-@Internal
+@PublicEvolving
 public enum FetchLogReadPreference {
     LOCAL_FIRST(0, "local-first"),
     REMOTE_FIRST(1, "remote-first");
-
     private final int value;
     private final String name;
 
--- a/fluss-common/src/main/java/org/apache/fluss/rpc/protocol/FetchLogReadPreference.java
+++ b/fluss-common/src/main/java/org/apache/fluss/rpc/protocol/FetchLogReadPreference.java
@@ -1,0 +1,53 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.fluss.rpc.protocol;
+
+import org.apache.fluss.annotation.Internal;
+
+/** Read preference for fetch log requests. */
+@Internal
+public enum FetchLogReadPreference {
+    LOCAL_FIRST(0, "local-first"),
+    REMOTE_FIRST(1, "remote-first");
+
+    private final int value;
+    private final String name;
+
+    FetchLogReadPreference(int value, String name) {
+        this.value = value;
+        this.name = name;
+    }
+
+    public int value() {
+        return value;
+    }
+
+    public static FetchLogReadPreference from(int value) {
+        for (FetchLogReadPreference preference : values()) {
+            if (preference.value == value) {
+                return preference;
+            }
+        }
+        return LOCAL_FIRST;
+    }
+
+    @Override
+    public String toString() {
+        return name;
+    }
+}
--- a/fluss-flink/fluss-flink-common/src/test/java/org/apache/fluss/flink/tiering/TieringITCase.java
+++ b/fluss-flink/fluss-flink-common/src/test/java/org/apache/fluss/flink/tiering/TieringITCase.java
@@ -34,15 +34,11 @@ import org.apache.fluss.types.DataTypes;
 import org.apache.fluss.utils.ExceptionUtils;
 
 import org.apache.flink.api.common.RuntimeExecutionMode;
-import org.apache.flink.core.execution.JobClient;
 import org.apache.flink.streaming.api.environment.StreamExecutionEnvironment;
 import org.junit.jupiter.api.AfterAll;
 import org.junit.jupiter.api.BeforeAll;
 import org.junit.jupiter.api.BeforeEach;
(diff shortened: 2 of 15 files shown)


zuston:
If you have time, could you help review this? @fresh-borzoni

fluss-common/src/main/java/org/apache/fluss/rpc/protocol/FetchLogReadPreference.java
```diff
@@ -0,0 +1,53 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.fluss.rpc.protocol;
+
+import org.apache.fluss.annotation.Internal;
+
+/** Read preference for fetch log requests. */
+@Internal
```
fresh-borzoni: Why @Internal really?

fluss-flink/fluss-flink-common/src/test/java/org/apache/fluss/flink/tiering/TieringITCase.java
```diff
@@ -110,7 +120,86 @@ void testTieringReachMaxDuration() throws Exception {
             tieredRecords = countTieredRecords(pkTableLakeSnapshot);
             assertThat(tieredRecords).isLessThan(recordCount);
         } finally {
-            jobClient.cancel();
+            jobClient.cancel().get();
+        }
+    }
+
+    @Test
+    @Order(2)
```
fresh-borzoni: Why does this need a fixed method order + cancel().get()? 
The new test creates its own table, so it should be independent. Do I miss smth here?

fluss-server/src/main/java/org/apache/fluss/server/replica/ReplicaManager.java
```diff
@@ -1529,6 +1545,37 @@ public Map<TableBucket, LogReadResult> readFromLog(
...
+            RemoteLogFetchInfo remoteLogFetchInfo =
+                    fetchLogFromRemote(replica, normalizedFetchOffset);
+            if (remoteLogFetchInfo != null) {
+                return new FetchLogResultForBucket(
+                        tb, remoteLogFetchInfo, replica.getLogHighWatermark());
+            }
+            return new FetchLogResultForBucket(
+                    tb,
+                    ApiError.fromThrowable(
+                            new LogOffsetOutOfRangeException(
```
fresh-borzoni: Should a remote miss here return null (fall through to local) instead of erroring? If the offset is in the local+remote overlap, local can still serve it and a genuinely-unavailable offset would still error via the normal out-of-range path.  It's minor and seems unreachable in practice, but still a bit confusing

fresh-borzoni (commented):
@zuston Thank you for the PR, left some comments, PTAL

fluss-common/src/main/java/org/apache/fluss/rpc/protocol/FetchLogReadPreference.java
zuston: removed

fluss-server/src/main/java/org/apache/fluss/server/replica/ReplicaManager.java
zuston: updated.

fluss-flink/fluss-flink-common/src/test/java/org/apache/fluss/flink/tiering/TieringITCase.java
zuston: One tip that the tiering job will share the multi table tiering tasks, to ensure the test case stability, we have to ensure only one tiering job with remote_first option.

zuston: I have updated this part code.

zuston:
thanks for your review @fresh-borzoni . Updated

fluss-server/src/main/java/org/apache/fluss/server/replica/ReplicaManager.java
```diff
@@ -1529,6 +1545,34 @@ public Map<TableBucket, LogReadResult> readFromLog(
...
+                    fetchLogFromRemote(replica, normalizedFetchOffset);
+            if (remoteLogFetchInfo != null) {
+                return new FetchLogResultForBucket(
+                        tb, remoteLogFetchInfo, replica.getLogHighWatermark());
+            }
+            // Remote log is expected to cover the offset, but segments/manifest may not be ready
+            // yet.
+            // REMOTE_FIRST is a preference, so fall back to local reads.
+            return null;
+        } catch (Exception e) {
```
fresh-borzoni, line 1580: This runs before the leader check in fetchRecords(), a demoted ex-leader only rejects stale clients because remoteLogTablet(tb) happens to throw and the catch-all eats it. Shall we gate on replica.isLeader() explicitly? 

Also, swallowing every exception with no log, so how would we ever notice remote storage misbehaving?

fresh-borzoni (commented):
@zuston Thank you, one comment, PTAL

fluss-server/src/main/java/org/apache/fluss/server/replica/ReplicaManager.java
zuston: nice catch. I have updated

fresh-borzoni (approved):
@zuston Thank you, LGTM 👍 , merging once CI passes

fresh-borzoni:
@zuston Appologies, I haven't noticed merge conflict - can you, please, resolve it?

zuston:
> @zuston Appologies, I haven't noticed merge conflict - can you, please, resolve it?

sure! rebased

fresh-borzoni:
@zuston Thank you
It seems CI failures aren't your server change, it's testTieringReachMaxDuration losing its timing race now that your new test runs first in the same class (failed 3x on this PR across 1.18/2.0/2.2, never on main).

That test asserts "tiering is still incomplete when I look", which only holds if you look early enough, so any warm-up shift from a sibling test breaks it.

Shall we move your new test into its own ITCase class? The old test's raciness is worth fixing too (wait for full tiering, then assert it took >1 snapshot - no race), but that's a separate issue, not this PR. You can choose whatever you feel works the best for you

zuston:
> Shall we move your new test into its own ITCase class? The old test's raciness is worth fixing too (wait for full tiering, then assert it took >1 snapshot - no race), but that's a separate issue, not this PR. You can choose whatever you feel works the best for you

I have fixed the prev tests directly. Anyway the failure is caused by this PR

fresh-borzoni:
@zuston Thank you, appreciate it 👍

zuston:
thanks for your patient and thorough review! @fresh-borzoni

PR #3608, opened 2026-07-07
[tiering] fix flink tiering duplicate write result

### LINKED ISSUE #3562: Fast fail due to the duplicate write results for tiering committer
### Search before asking

### Fluss version

0.9.0 (latest release)

### Please describe the bug 🐞

The bug is caused by overlapping tiering rounds for the same table.
When an old round is force-finished due to max duration, its results may still be pending in the committer. Meanwhile, the enumerator can start and assign a new round for the same table. Since TableBucketWriteResult does not carry an epoch/round id, the committer groups results only by table and bucket.
So results from the old round and the new round can be mixed together. If both contain the same bucket, the committer treats them as duplicate write results and fails.

Image

Image

### Solution

_No response_

zuston:
### Purpose

Linked issue: close #3562 

This PR fixes duplicate tiering write results caused by slow split generation.

When split generation takes a long time, the source enumerator may hold a requested table epoch without reporting it as an active tiering epoch. If the tiering trigger interval is short, the coordinator can assign the same table again with a new epoch before the previous round finishes generating and processing splits. The committer then may receive write results from two rounds of the same table and detect duplicate bucket results.


luoyuxia (approved):
@zuston Thanks. LGTM!

PR #3612, opened 2026-07-08
[server] Add configurable disk write-limit recover gap

### LINKED ISSUE #3611: [server] Support dynamic configuration for disk write-limit recover gap
### Search before asking

### Motivation

`DiskUsageMonitor` currently uses a hard-coded recover gap of `0.10` when
deciding when to recover from data disk write lock. This means writes only resume
after disk usage drops below `server.data-disk.write-limit-ratio - 0.10`, which
can be difficult to reach in practice.

We should make this recover gap configurable and dynamically updatable.

### Solution

_No response_

### Anything else?

_No response_

### Willingness to contribute

swuferhong:
### Purpose

Linked issue: close #3611 

Parameterize the `DiskUsageMonitor` recover gap with `server.data-disk.write-limit-recover-gap` and set its default value to `0.05`.

The recover gap now supports dynamic updates together with `server.data-disk.write-limit-ratio`. Startup and dynamic config validation now `enforce 0 < recover-gap < write-limit-ratio <= 1.0`.


fluss-common/src/main/java/org/apache/fluss/config/ConfigOptions.java
```diff
@@ -385,10 +385,24 @@ public class ConfigOptions {
...
+                                    + "Writes resume after the usage drops below "
+                                    + "(server.data-disk.write-limit-ratio - "
+                                    + "server.data-disk.write-limit-recover-gap). "
                                     + "Set to 1.0 to disable the disk-usage protection entirely. "
-                                    + "The valid range is (0.1, 1.0].");
+                                    + "The valid range is "
+                                    + "(server.data-disk.write-limit-recover-gap, 1.0].");
+
+    public static final ConfigOption<Double> SERVER_DATA_DISK_WRITE_LIMIT_RECOVER_GAP =
+            key("server.data-disk.write-limit-recover-gap")
```
wuchong: Could we consider renaming these configuration keys and expressing both values as absolute watermarks?

Currently, we have:

- `server.data-disk.write-limit-ratio`
- `server.data-disk.write-limit-recover-gap`

Combining an absolute ratio with a relative gap requires users to calculate the actual recovery threshold themselves (`limit-ratio - recover-gap`). It also makes validation and dynamic updates harder to reason about.

I suggest using two mirrored thresholds that map directly to the write-protection state transitions:

- **`server.data-disk.write-limit-ratio`** — for example, `0.85`, at which write protection is activated.
- **`server.data-disk.write-recover-ratio`** — for example, `0.80`, at which write protection is deactivated.

The relationship then becomes straightforward:

```text
0 < write-recover-ratio < write-limit-ratio <= 1.0
```

wuchong (commented):
Thanks @swuferhong , I only have one comment.

swuferhong:
@wuchong comments addressed. PTAL, thanks.

wuchong (approved):
LGTM.

PR #3614, opened 2026-07-08
[server] Add resource-based KV leader replica capacity control

### LINKED ISSUE #3613: [server] Add resource-based KV leader replica capacity control
### Search before asking

### Motivation

Fluss currently relies mainly on `max.bucket.num` and `max.partition.num` to limit table metadata size. These limits do not directly protect the cluster from creating too many KV leader replicas, which can consume resident memory on `TabletServers` and increase OOM risk.

We need a coordinator-side capacity guard that can estimate the total KV leader replica capacity from TabletServer memory and reject new KV table/partition creation when the cluster no longer has enough capacity.

### Solution

_No response_

### Anything else?

_No response_

### Willingness to contribute

swuferhong:
### Purpose

Linked issue: close #3613 

Add `tabletServer` resource reporting and a coordinator-side `KvLeaderReplicaCapacityManager` to protect clusters from creating too many KV leader replicas.

`TabletServers` now report optional CPU and memory information during registration. The coordinator derives `kvLeaderReplicaCapacity` from live `tabletServer` memory and `kv.leader-replica.memory-reserved`, tracks the current KV leader replica count in memory, rebuilds it on coordinator leadership startup, and checks/reserves capacity when creating KV tables or partitions.

The capacity limit is disabled unless more than half of live tablet servers report memory. Unknown-memory `TabletServers`use the average known memory when the limit is enabled.

Also add dynamic config support, metrics, RPC error mapping, serde compatibility, auto-partition accounting, and tests for resource probing, capacity calculation, manual create/drop, and auto partition create/drop.


fluss-server/src/main/java/org/apache/fluss/server/coordinator/CoordinatorService.java
```diff
@@ -447,6 +450,9 @@ public CompletableFuture<DropDatabaseResponse> dropDatabase(DropDatabaseRequest
         DropDatabaseResponse response = new DropDatabaseResponse();
         metadataManager.dropDatabase(
                 request.getDatabaseName(), request.isIgnoreIfNotExists(), request.isCascade());
+        if (request.isCascade()) {
```
beryllw: Thanks for the change! One concern about rebuildCurrentCount(metadataManager) in dropDatabase: it rescans all databases/tables, but admin operations run concurrently on the coordinator. If another request drops a database mid-scan, listTables() throws DatabaseNotExistException, which propagates out and fails this dropDatabase request — even with ignoreIfNotExists=true.

This is what fails the Rust IT test_error_database_already_exist in CI: its drop_database(..., cascade=true) gets a DatabaseNotExist error for a different database (test_create_database) that a parallel test was dropping at the same time.

fluss-server/src/main/java/org/apache/fluss/server/coordinator/KvLeaderReplicaCapacityManager.java
```diff
@@ -0,0 +1,221 @@
...
+    }
+
+    @Override
+    public void validate(Configuration newConfig) throws ConfigException {
+        long memoryReservedBytes = getLeaderReplicaMemoryReservedBytes(newConfig);
+        if (memoryReservedBytes <= 0) {
+            throw new ConfigException(
+                    ConfigOptions.KV_LEADER_REPLICA_MEMORY_RESERVED.key()
+                            + " must be greater than 0.");
+        }
```
wuchong: Is it possible (how to) to disable the leader-replica memory reserve?

fluss-server/src/main/java/org/apache/fluss/server/coordinator/AutoPartitionManager.java
wuchong: Restrict the number of constructor. Esp. for the constructor only used for tests, mark them package-visible and `@VisibleForTesting`.

fluss-server/src/main/java/org/apache/fluss/server/coordinator/CoordinatorServer.java
```diff
@@ -240,20 +244,28 @@ protected void initCoordinatorStandby() throws Exception {
...
             this.remoteDirDynamicLoader = new RemoteDirDynamicLoader(conf);
+            this.metadataCache = new CoordinatorMetadataCache();
+            this.kvLeaderReplicaCapacityManager =
+                    new KvLeaderReplicaCapacityManager(conf, metadataCache);
+            serverMetricGroup.gauge(
+                    MetricNames.KV_LEADER_REPLICA_COUNT,
+                    kvLeaderReplicaCapacityManager::getKvLeaderReplicaCount);
+            serverMetricGroup.gauge(
+                    MetricNames.KV_LEADER_REPLICA_CAPACITY,
+                    kvLeaderReplicaCapacityManager::getKvLeaderReplicaCapacity);
```
wuchong: Move the metric register into `KvLeaderReplicaCapacityManager`.

fluss-server/src/main/java/org/apache/fluss/server/coordinator/KvLeaderReplicaCapacityManager.java
```diff
@@ -0,0 +1,221 @@
...
+        if (knownMemoryTabletServerCount * 2 <= liveTabletServerCount) {
+            return CAPACITY_LIMIT_DISABLED;
+        }
+
+        long averageKnownMemoryBytes = knownMemoryBytes / knownMemoryTabletServerCount;
+        long totalMemoryBytes =
+                knownMemoryBytes
+                        + averageKnownMemoryBytes
+                                * (liveTabletServerCount - knownMemoryTabletServerCount);
+        return totalMemoryBytes / leaderReplicaMemoryReservedBytes;
```
wuchong: Add debug log for the capacity calculation details.

fluss-common/src/main/java/org/apache/fluss/config/ConfigOptions.java
```diff
@@ -547,6 +555,22 @@ public class ConfigOptions {
                             "The rack for the tabletServer. This will be used in rack aware bucket assignment "
                                     + "for fault tolerance. Examples: `RACK1`, `cn-hangzhou-server10`");
 
+    public static final ConfigOption<Double> TABLET_SERVER_RESOURCE_CPU_CORES =
+            key("tablet-server.resource.cpu-cores")
```
wuchong: **Clarify that TabletServer resource options are advertised capacity, not enforced limits**

The current keys, `tablet-server.resource.cpu-cores` and `tablet-server.resource.memory-size`, sound like they configure or enforce the resources available to the TabletServer process, such as a CPU quota, cgroup memory limit, or JVM heap size. However, these values only override the resource capacity advertised to the CoordinatorServer. CPU is currently used for resource reporting, while memory is also used to estimate the cluster-level KV leader replica capacity.

Could we rename them to make the advertised-only semantics explicit?

- `tablet-server.advertised-resource.cpu-cores`
- `tablet-server.advertised-resource.memory-size`

The Java constants could be renamed accordingly:

- `TABLET_SERVER_ADVERTISED_RESOURCE_CPU_CORES`
- `TABLET_SERVER_ADVERTISED_RESOURCE_MEMORY_SIZE`

The descriptions should also explicitly state that these options do not enforce resource limits. For example:

```java
"The CPU capacity, in cores, that this tablet server advertises "
        + "to the CoordinatorServer for resource reporting. "
        + "This option does not limit CPU usage or configure a cgroup CPU quota. "
        + "If not configured, the tablet server detects the value from "
        + "cgroup CPU quota or the JVM runtime."
```

```java
"The memory capacity that this tablet server advertises "
        + "to the CoordinatorServer for resource reporting and "
        + "cluster-level KV leader replica capacity estimation. "
        + "This option does not configure JVM heap size, reserve memory, "
        + "or enforce a process or container memory limit. "
        + "It represents total usable capacity, not current memory usage "
        + "or free memory. If not configured, the tablet server detects "
        + "the value from cgroup or operating system information."
```

Using `advertised-resource` makes it clear that these values are informational inputs for reporting and capacity planning, rather than resource limits enforced by the TabletServer.

fluss-server/src/main/java/org/apache/fluss/server/coordinator/KvLeaderReplicaCapacityManager.java
```diff
@@ -0,0 +1,221 @@
...
+    private final CoordinatorMetadataCache metadataCache;
+
+    private long leaderReplicaMemoryReservedBytes;
+    private long currentKvLeaderReplicaCount;
+
+    public KvLeaderReplicaCapacityManager(
+            Configuration conf, CoordinatorMetadataCache metadataCache) {
+        this.metadataCache = checkNotNull(metadataCache, "metadataCache should not be null.");
+        this.leaderReplicaMemoryReservedBytes = getLeaderReplicaMemoryReservedBytes(conf);
+    }
```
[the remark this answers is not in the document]
wuchong: `kv.leader-replica.memory-reserved` currently accepts `0b` from the static configuration, but the capacity calculation divides the reported TabletServer memory by this value and can therefore throw `ArithmeticException`. Dynamic validation is also inconsistent because it rejects zero.

I think we should allow defining `0` as the explicit switch for disabling memory-based KV leader replica capacity control. Map it to `CAPACITY_LIMIT_DISABLED` before performing the division, allow it in dynamic validation, and document the behavior in the config description. Positive values should enable capacity estimation as usual, while negative values remain invalid.


The improved option description:

```java
"The estimated memory consumption of each KV leader replica, "
                                + "used by the CoordinatorServer to calculate the cluster-level "
                                + "KV leader replica capacity. This value does not reserve or "
                                + "enforce memory on the tablet server. A value of 0 disables "
                                + "memory-based KV leader replica capacity control.");
```

wuchong (commented):
Thanks @swuferhong , I commented the major concern here:

**Avoid maintaining an exactly-once KV leader count outside `CoordinatorContext`**

`currentKvLeaderReplicaCount` creates a second mutable representation of coordinator state. Keeping it accurate requires reservation, rollback and release logic across every table/partition CRUD path, while failover rebuild performs an expensive and non-atomic metadata scan.

Please maintain an event-thread-owned `Set<TableBucket>` in `CoordinatorContext` instead. Add observed KV buckets on create, conservatively treat unknown orphan assignments as KV during failover, and remove them idempotently from `onDeleteTableBucket` as `ResumeDropEvent`s are admitted by the lifecycle throttler. `KvLeaderReplicaCapacityManager` should only keep a volatile snapshot derived from `kvBuckets.size()` and perform a best-effort `observed + requested > capacity` check, without reservation, rollback, decrease, or full rebuild logic.

wuchong:
Btw, Could we rename `KvLeaderReplicaCapacityManager` to `ReplicaCapacityController`?

`Replica` provides a better component boundary because capacity control may later cover log replicas and KV hot replicas in addition to KV leader replicas. `Controller` is also more accurate than `Manager`: this component does not manage replica lifecycles; it makes creation admission decisions based on observed replica counts and cluster resource capacity.

However, the current methods, metrics, and configuration should remain explicitly scoped to KV leaders, since that is the only replica type currently supported.

swuferhong:
Hi, @wuchong comments addressed. PTAL, thanks.

fluss-common/src/main/java/org/apache/fluss/config/ConfigOptions.java
```diff
@@ -577,6 +588,30 @@ public class ConfigOptions {
                             "The rack for the tabletServer. This will be used in rack aware bucket assignment "
                                     + "for fault tolerance. Examples: `RACK1`, `cn-hangzhou-server10`");
 
+    public static final ConfigOption<Double> TABLET_SERVER_ADVERTISED_RESOURCE_CPU_CORES =
+            key("tablet-server.advertised-resource.cpu-cores")
+                    .doubleType()
+                    .noDefaultValue()
+                    .withDeprecatedKeys("tablet-server.resource.cpu-cores")
```
wuchong: We don't need to keep compatibility with `tablet-server.resource.cpu-cores`, as it is never exposed to users.

fluss-common/src/main/java/org/apache/fluss/config/ConfigOptions.java
```diff
@@ -577,6 +588,30 @@ public class ConfigOptions {
...
+                            "The CPU capacity, in cores, that this tablet server advertises to the "
+                                    + "CoordinatorServer for resource reporting. This option does not limit "
+                                    + "CPU usage or configure a cgroup CPU quota. If not configured, the tablet "
+                                    + "server detects the value from cgroup CPU quota or the JVM runtime.");
+
+    public static final ConfigOption<MemorySize> TABLET_SERVER_ADVERTISED_RESOURCE_MEMORY_SIZE =
+            key("tablet-server.advertised-resource.memory-size")
+                    .memoryType()
+                    .noDefaultValue()
+                    .withDeprecatedKeys("tablet-server.resource.memory-size")
```
wuchong: ditto

fluss-server/src/main/java/org/apache/fluss/server/metadata/CoordinatorMetadataCache.java
```diff
@@ -94,6 +94,19 @@ public Set<TabletServerInfo> getAliveTabletServerInfos() {
         return Collections.unmodifiableSet(tabletServerInfos);
     }
 
+    public Set<ServerInfo> getLiveTabletServerInfos() {
+        Map<Integer, ServerInfo> aliveTabletServers = metadataSnapshot.aliveTabletServers;
```
wuchong: Get `metadataSnapshot` into a local variable first, then access it via the local variable, otherwise, there is data consistency problem.

fluss-server/src/test/java/org/apache/fluss/server/tablet/TabletServerResourceProbeTest.java
```diff
@@ -0,0 +1,148 @@
...
+    @Test
+    void testProbeFromDeprecatedResourceConfig() {
+        Configuration conf = new Configuration();
+        conf.setString("tablet-server.resource.cpu-cores", "4.0");
+        conf.setString("tablet-server.resource.memory-size", "2kb");
+
+        TabletServerResource resource = new TabletServerResourceProbe(conf, tempDir).probe();
+
+        assertThat(resource.getCpuCores()).isEqualTo(4.0);
+        assertThat(resource.getMemoryBytes()).isEqualTo(2048);
```
wuchong: We don't need to keep compatible with them as they are never exposed to users.

fluss-server/src/main/java/org/apache/fluss/server/coordinator/CoordinatorEventProcessor.java
```diff
@@ -597,6 +604,14 @@ private void loadAssignment(
         }
     }
 
+    @VisibleForTesting
+    void trackKvBucketsForLoadedAssignment(long tableId, Set<TableBucket> tableBuckets) {
+        TableInfo tableInfo = coordinatorContext.getTableInfoById(tableId);
+        if (tableInfo == null || tableInfo.hasPrimaryKey()) {
```
wuchong, line 617: Add comment to explain why tableInfo not found is treated as kv buckets.

fluss-server/src/main/java/org/apache/fluss/server/coordinator/CoordinatorService.java
```diff
@@ -715,40 +734,50 @@ public CompletableFuture<CreatePartitionResponse> createPartition(
...
+                tableInfo.getTableConfig().getAutoPartitionStrategy());
 
         ResolvedPartitionSpec partitionToCreate =
-                ResolvedPartitionSpec.fromPartitionSpec(table.partitionKeys, partitionSpec);
+                ResolvedPartitionSpec.fromPartitionSpec(
+                        tableInfo.getPartitionKeys(), partitionSpec);
+        if (request.isIgnoreIfNotExists()
+                && metadataManager
+                        .getPartitions(tablePath)
+                        .contains(partitionToCreate.getPartitionName())) {
```
wuchong: Use `org.apache.fluss.server.coordinator.MetadataManager.getOptionalPartitionRegistration` to check partition exists which is more cheap?

fluss-server/src/main/java/org/apache/fluss/server/coordinator/ReplicaCapacityController.java
```diff
@@ -0,0 +1,197 @@
...
+
+    private static final Logger LOG = LoggerFactory.getLogger(ReplicaCapacityController.class);
+
+    /** Negative capacity means the automatic capacity limit is disabled. */
+    public static final long CAPACITY_LIMIT_DISABLED = -1L;
+
+    private final Object lock = new Object();
+    private final CoordinatorMetadataCache metadataCache;
+
+    private long kvLeaderReplicaMemoryReservedBytes;
```
wuchong: can be volatile, then we don't need to use lock synchronized

fluss-server/src/main/java/org/apache/fluss/server/coordinator/ReplicaCapacityController.java
```diff
@@ -0,0 +1,197 @@
...
+
+    private void registerMetrics(CoordinatorMetricGroup coordinatorMetricGroup) {
+        coordinatorMetricGroup.gauge(
+                MetricNames.KV_LEADER_REPLICA_COUNT, this::getKvLeaderReplicaCount);
+        coordinatorMetricGroup.gauge(
+                MetricNames.KV_LEADER_REPLICA_CAPACITY, this::getKvLeaderReplicaCapacity);
+    }
+
+    private long getKvLeaderReplicaCapacityLocked() {
+        if (kvLeaderReplicaMemoryReservedBytes == 0) {
```
wuchong: Should copy to a local variable first, then access via the local variable. Otherwise, there maybe in-consistency in this method.

fluss-server/src/main/java/org/apache/fluss/server/coordinator/CoordinatorEventProcessor.java
```diff
@@ -821,6 +842,7 @@ private void processCreateTable(CreateTableEvent createTableEvent) {
         coordinatorContext.putTableInfo(tableInfo);
         TableAssignment tableAssignment = createTableEvent.getTableAssignment();
         tableManager.onCreateNewTable(tablePath, tableInfo.getTableId(), tableAssignment);
+        updateObservedKvLeaderReplicaCount();
```
wuchong: **[P2] Centralize the observed KV bucket count update in `TableManager`**

Could we move `updateObservedKvLeaderReplicaCount()` into `TableManager` and invoke it immediately after `CoordinatorContext#addKvBuckets` / `removeKvBuckets`?

Currently, every `CoordinatorEventProcessor` caller must remember to refresh the derived count after calling `TableManager#onCreateNewTable`, `onCreateNewPartition`, `onDeleteTable`, or `onDeletePartition`. Missing any current or future call site would leave the admission controller with a stale count.

For example:

```java
private void updateObservedKvLeaderReplicaCount() {
    replicaCapacityController.updateObservedKvLeaderReplicaCount(
            coordinatorContext.getKvBucketCount());
}

private void onCreateNewTableBucket(long tableId, Set<TableBucket> tableBuckets) {
    TableInfo tableInfo = coordinatorContext.getTableInfoById(tableId);
    if (tableInfo != null && tableInfo.hasPrimaryKey()) {
        coordinatorContext.addKvBuckets(tableBuckets);
        updateObservedKvLeaderReplicaCount();
    }

    // Continue with state-machine transitions...
}

private void onDeleteTableBucket(
        Set<TableBucket> tableBuckets, Set<TableBucketReplica> allReplicas) {
    // Remove unconditionally because orphan assignments may also have been tracked as KV buckets.
    coordinatorContext.removeKvBuckets(tableBuckets);
    updateObservedKvLeaderReplicaCount();

    // Continue with state-machine transitions...
}
```

Keeping the source-of-truth mutation and derived-count publication together also ensures that the count remains consistent if a later state-machine transition throws.

`CoordinatorEventProcessor` would then only need explicit synchronization at the two lifecycle boundaries that bypass `TableManager`:

- after failover initialization finishes loading all assignments;
- after shutdown resets `CoordinatorContext`.

This would remove the duplicated updates from the normal create, delete, and `ResumeDropEvent` paths and make future lifecycle changes much less error-prone.

fluss-common/src/main/java/org/apache/fluss/config/ConfigOptions.java
```diff
@@ -319,6 +319,17 @@ public class ConfigOptions {
                                     + "This default is capped to reduce the risk that an assignment znode exceeds "
                                     + "ZooKeeper's packet size limit.");
 
+    public static final ConfigOption<MemorySize> KV_LEADER_REPLICA_MEMORY_RESERVED =
+            key("kv.leader-replica.memory-reserved")
+                    .memoryType()
+                    .defaultValue(MemorySize.parse("8mb"))
```
wuchong: I'm thinking that, should we disable it by default? I'm worried there are some bad cases in different environment. And it can be a advanced feature that can be enabled by users. 

Btw, we should add documentation about this feature (when and how to enable/disable/configure it).  Can be a follow-up issue/PR.

swuferhong: Ok, I aggre with you, and I will introduce documents in next pr.

fluss-server/src/main/java/org/apache/fluss/server/coordinator/CoordinatorEventProcessor.java
```diff
@@ -319,20 +323,6 @@ public void startup() {
...
-        // can process the LeaderRequests that are generated by replicaStateMachine.startup() and
-        // partitionStateMachine.startup().
-        // update coordinator metadata cache when CoordinatorServer start.
-        HashSet<ServerInfo> tabletServerInfoList =
-                new HashSet<>(coordinatorContext.getLiveTabletServers().values());
-        serverMetadataCache.updateMetadata(
-                coordinatorContext.getCoordinatorServerInfo(),
-                tabletServerInfoList,
-                coordinatorContext.getServerTags());
-        updateTabletServerMetadataCacheWhenStartup(tabletServerInfoList);
```
wuchong, line 334: Is the failed case related to this change?

By the way, I believe it is acceptable if the auto-partition manager starts before the metadata cache update, as this does not cause server failure. The task will be rescheduled at the next interval, and the partition will eventually be created. On the other hand, we should add a delay (e.g., `1min`) to the schedule in the auto-partition manager instead of using `0`. Starting immediately is risky because the server is typically busy during startup.

swuferhong: This should have been caused by the state before commit 004e514b, where `metadataCache` was initialized twice, so the `metadataCache` instance passed as a constructor argument to `ReplicaCapacityController` got overwritten.

Adding the delay may cause many ITCases that rely on auto partition to fail. We'd need to expose the delay as a config parameter so it can be configured in tests. Could we do this in a follow-up PR?

swuferhong:
Hi, @wuchong  comments addressed and ci passed.

wuchong (approved):
LGTM.

PR #3617, opened 2026-07-08
[flink] respect flink io tmp multi dirs to fix disk exhaustion

### LINKED ISSUE #3616: Incorrect scanner default log dirs on Yarn
### Search before asking

### Fluss version

0.9.0 (latest release)

### Please describe the bug 🐞

For the logic in `getClientScannerIoTmpDir` , when the job runs on Hadoop YARN, Flink’s `TMP_DIRS` may contain multiple directories. In the current codebase, this value is treated as a single path, which can cause data to be written to an incorrect location and lead to uncontrolled disk usage.

### Solution

_No response_

zuston:
### Purpose

Linked issue: close #3616 

### Brief change log

respect  flink io tmp multi dirs to fix disk exhaustion


fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/utils/FlinkConnectorOptionsUtils.java
```diff
@@ -201,13 +202,38 @@ public static long parseTimestamp(String timestampStr, String optionKey, ZoneId
...
+            org.apache.flink.configuration.Configuration flinkConfig,
+            int taskIndex) {
+        return flussConf
+                .getOptional(CLIENT_SCANNER_IO_TMP_DIR)
+                .orElseGet(
+                        () -> {
+                            String[] flinkTmpDirs = getFlinkIoTmpDirs(flinkConfig);
+                            int idx = taskIndex % flinkTmpDirs.length;
+                            return new File(flinkTmpDirs[idx], "/fluss").getAbsolutePath();
+                        });
```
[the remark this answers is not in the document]
wuchong: +1

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/utils/FlinkConnectorOptionsUtils.java
```diff
@@ -201,13 +202,38 @@ public static long parseTimestamp(String timestampStr, String optionKey, ZoneId
...
+            String[] paths = splitPaths(flinkConfig.get(CoreOptions.TMP_DIRS));
+            if (paths.length > 0) {
+                return paths;
             }
         }
-        return flussConf.getString(CLIENT_SCANNER_IO_TMP_DIR);
+        return new String[] {System.getProperty("java.io.tmpdir")};
+    }
+
+    private static String[] splitPaths(@Nonnull String separatedPaths) {
```
beryllw: Flink already provides a built-in helper for splitting multiple directories here — ConfigurationUtils.splitPaths (and the wrapper parseTempDirectories). It keeps the separator handling consistent with Flink internals, so we could probably just reuse it and avoid maintaining a separate splitting logic. What do you think?

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/utils/FlinkConnectorOptionsUtils.java
```diff
@@ -201,13 +202,38 @@ public static long parseTimestamp(String timestampStr, String optionKey, ZoneId
...
+                .getOptional(CLIENT_SCANNER_IO_TMP_DIR)
+                .orElseGet(
+                        () -> {
+                            String[] flinkTmpDirs = getFlinkIoTmpDirs(flinkConfig);
+                            int idx = taskIndex % flinkTmpDirs.length;
+                            return new File(flinkTmpDirs[idx], "/fluss").getAbsolutePath();
+                        });
+    }
+
+    private static String[] getFlinkIoTmpDirs(
```
wuchong: Use Flink's `org.apache.flink.configuration.ConfigurationUtils#parseTempDirectories`?

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/utils/FlinkConnectorOptionsUtils.java
wuchong: +1

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/source/FlinkSource.java
```diff
@@ -290,7 +290,8 @@ public SourceReader<OUT, SourceSplitBase> createReader(SourceReaderContext conte
 
         flussConf.set(
```
beryllw: Just a personal question, might be missing something here — since flussConf is a member field of the source, calling set seems to mutate it in place, and this config eventually gets passed down to the Connection and shared by the downstream scanners. So I'm wondering whether multiple readers reusing the same instance could end up sharing the same config. Would it be safer to copy it first before setting？
Curious how you see it @luoyuxia ?

zuston:
updated. @wuchong @beryllw

zuston:
CI failure looks unrelated with this PR. gentle ping @wuchong @beryllw

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/source/FlinkSource.java
zuston: aha. this logic is aligned with the previous impl. I think it's OK to share

zuston:
@fresh-borzoni If you have time, could you help review this?

luoyuxia:
@zuston Thanks for working on this. I found one issue with the current directory-selection strategy:

`taskIndex` is the operator global subtask index, while `io.tmp.dirs` contains local directories on each TaskManager. Flink does not guarantee that the subtasks placed on one TaskManager have different values modulo the number of local directories. For example, with two tmp dirs, subtasks 0 and 2 can be scheduled on the same TaskManager; both select directory 0, leaving directory 1 unused. Multiple Fluss sources or jobs can produce the same deterministic hotspot. Therefore, this may still exhaust one disk—the problem this PR is intended to fix.

Could you please use a TaskManager/JVM-local round-robin allocator, or Flink provided `ConfigurationUtils.getRandomTempDirectory()`? At minimum, the local disk index should not be derived from the global subtask index. Please also add coverage for multiple readers sharing a TaskManager.

Also, please rebase this PR onto the latest `main` branch. Thanks!

zuston:
gently ping @luoyuxia

PR #3621, opened 2026-07-08
[kv] Add row TTL support

### LINKED ISSUE #3620: [KV] Support row-level TTL for primary-key tables
### Search before asking

### Motivation

Primary-key tables often keep the latest value for each key, but many workloads only need those values within a bounded lifetime. Without table-level row TTL, users must either issue explicit deletes, maintain cleanup jobs, or tolerate unbounded KV state growth.

This is inefficient for common use cases such as session state, device heartbeat state, deduplicated event state, and temporary lookup dimensions. Fluss should provide a native row-level TTL mechanism so stale KV rows can be cleaned up automatically with clear, best-effort semantics.

### Solution

Introduce row-level TTL for primary-key tables through `table.kv.row.ttl`. The `table.kv.row.ttl` namespace explicitly scopes row-level expiration to KV state and distinguishes it from log or partition retention. Row TTL is disabled by default. When enabled, expired rows are removed asynchronously during RocksDB compaction. Cleanup is best-effort: a row is not guaranteed to become inaccessible immediately after its expiration time.

Fluss owns the versioned KV value layout used by compaction policies. The layout carries a generic value tag, which row TTL uses as its TTL reference timestamp. Writers, readers, recovery paths, and compaction filters interpret values through the same layout contract. This avoids relying on a RocksDB-specific opaque value format and allows different compaction policies and TTL time semantics to share one storage representation.

#### Time semantics

When only `table.kv.row.ttl` is configured, TTL uses processing time. The lifetime of a row starts from the timestamp associated with its latest write.

Users may configure `table.kv.row.ttl.time-column` to use event time. The referenced column may use:

- `BIGINT`, interpreted as epoch milliseconds; or
- `TIMESTAMP_LTZ`.

A null event-time value means that the row does not expire. The time column is persisted by stable column identity, so renaming it does not change the TTL association.

Updates, including partial updates, recompute the TTL value tag from the resulting complete row. Recovery preserves the original TTL semantics instead of extending row lifetime from the recovery time.

`table.kv.row.ttl.changelog-mode` defines the CDC behavior of expiration. Only `none` is supported in this version, so compaction-based expiration does not emit delete records.

#### Compatibility

TTL tables use a versioned Fluss KV value layout containing the additional value tag. All encoding and decoding of this layout is centralized in the value-layout abstraction.

Clients using raw KV operations must advertise a protocol version that understands this layout before accessing a TTL table. Older clients are rejected instead of being allowed to decode or write the table incorrectly.

The KV format version and stable identity of an event-time column are persisted in table metadata so that servers, clients, snapshots, and recovery paths derive consistent behavior.

#### Limitations

- TTL cleanup is asynchronous and best-effort.
- Compaction-based expiration does not produce delete records in CDC.
- Altering row TTL or its time column is not supported in this change.
- Dropping a column referenced by row TTL is not supported.
- Exact row-count pushdown is not available for TTL tables because compaction removes rows outside the normal write path.

### Anything else?

_No response_

### Willingness to contribute

platinumhamburg:
### Purpose

Linked issue: close #3620

#### Motivation

Primary key tables retain their latest row state indefinitely. Applications that only need recent state must currently clean it up externally.

This change introduces opt-in row-level TTL for primary key tables. Row TTL is intended for cases that require per-row expiration and can accept weaker changelog semantics. Time-partitioned expiration remains the recommended approach when applicable because dropping complete partitions preserves changelog completeness.

#### Solution

Row TTL is enabled with `table.kv.ttl` and is disabled when the option is absent.

```sql
CREATE TABLE pk_table (
    id BIGINT,
    event_time TIMESTAMP_LTZ(3),
    name STRING,
    PRIMARY KEY (id) NOT ENFORCED
) WITH (
    'table.kv.ttl' = '7 d',
    'table.kv.ttl.time-column' = 'event_time'
);

### CODE REVIEW

**Comment by hiSandog** [comment]
Row TTL touches encoding, snapshot reads, lookups, and Flink conversion, so the intended expiration boundary would be helpful to state explicitly in the PR description. Is a row considered expired when `currentTime == expirationTime`, and is the decision based on a server-supplied clock or the reading client’s clock? A focused test around the exact boundary, including a snapshot created before expiry but read after expiry, would make the behavior easier to preserve across scanners.

**Comment by platinumhamburg** [comment]
> Row TTL touches encoding, snapshot reads, lookups, and Flink conversion, so the intended expiration boundary would be helpful to state explicitly in the PR description. Is a row considered expired when `currentTime == expirationTime`, and is the decision based on a server-supplied clock or the reading client’s clock? A focused test around the exact boundary, including a snapshot created before expiry but read after expiry, would make the behavior easier to preserve across scanners.

Hi @hiSandog , thanks for your suggestion, the documentation changes in the PR explicitly define TTL as a best-effort mechanism rather than providing precise expiration guarantees. There is no commitment to immediate expiration once the time threshold is crossed (i.e., we will not perform additional row-level visibility filtering strictly for TTL semantics, as this would generally exceed typical user requirements and incur significant performance overhead). Furthermore, the expiration decision is explicitly based on the server clock. For scenarios involving event-time TTL, the TTL configuration must clearly specify either precise timezone semantics or an exact timestamp.

File: fluss-server/src/main/java/org/apache/fluss/server/utils/TableDescriptorValidation.java [hunk 1]
```diff
@@ -242,6 +252,154 @@ private static void checkSystemColumns(RowType schema) {
...
+
+        validateRowTTLDuration(rowTTL.get());
+
+        if (timeColumn.isPresent()) {
+            Schema.Column column = getRowTTLTimeColumn(schema, timeColumn.get());
+            validateRowTTLTimeColumnType(column.getDataType());
+            if (timeColumnId.isPresent()) {
+                validateRowTTLTimeColumnId(timeColumnId.get(), column.getColumnId());
+            }
+        }
```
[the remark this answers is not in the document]
platinumhamburg: Issue updated.

website/docs/table-design/data-distribution/ttl.md
```diff
@@ -8,4 +8,46 @@ sidebar_position: 3
...
+) WITH (
+    'bucket.num' = '4',
+    'table.row.ttl' = '7 d',
+    'table.row.ttl.time-column' = 'event_time'
+);
+```
+
+The event-time column must be `BIGINT` epoch milliseconds or `TIMESTAMP_LTZ`. Rows with null event-time values do not expire through row TTL.
+
+Row TTL must be configured when creating the primary key table. Changing or disabling `table.row.ttl`, or changing `table.row.ttl.time-column`, with `ALTER TABLE ... SET` or `ALTER TABLE ... RESET` is not supported in this version.
```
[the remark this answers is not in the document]
platinumhamburg: Issue updated.

website/docs/engine-flink/ddl.md
```diff
@@ -282,7 +282,7 @@ ALTER TABLE my_table MODIFY WATERMARK FOR ts AS ts - INTERVAL '20' SECOND;
 ### SET properties
 The SET statement allows users to configure one or more connector options including the [Storage Options](engine-flink/options.md#storage-options) for a specified table. If a particular option is already configured on the table, it will be overridden with the new value.
 
-When using SET to modify [Storage Options](engine-flink/options.md#storage-options), the Fluss cluster will dynamically apply the changes to the table. This can be useful for modifying table behavior without needing to recreate the table.
+When using SET to modify [Storage Options](engine-flink/options.md#storage-options), the Fluss cluster will apply the changes according to each option's runtime contract. This can be useful for modifying table behavior without needing to recreate the table.
```
wuchong: I feel that the original phrasing more clearly conveys that parameter updates are dynamically applied **to the table.**

website/docs/engine-flink/ddl.md
```diff
@@ -308,6 +308,7 @@ ALTER TABLE my_table SET ('table.log.tiered.local-segments' = '5');
 **Limits**
 - If lakehouse storage (`table.datalake.enabled`) is already enabled for a table, options with lakehouse format prefixes (e.g., `paimon.*`) cannot be modified again.
 - After changing `table.datalake.historical-partition.enabled`, restart existing lookup jobs that need to look up historical partition data so that their clients load the updated table configuration.
+- `table.kv.row.ttl` cannot be modified with `ALTER TABLE ... SET` in this version. Configure row-level TTL when creating the primary-key table.
```
wuchong: I don't think we need to explain each parameter that does not support dynamic updates, as there are too many. By default, parameters do not support hot updates, except for those listed in the **Supported Options to modify** section above. Therefore, we only need to maintain that support list.

The **Limits** section provides additional explanations for `table.datalake.enabled` and `table.datalake.historical-partition.enabled` because, although these parameters support hot updates, they are subject to certain restrictions.

website/docs/engine-flink/ddl.md
```diff
@@ -319,6 +320,8 @@ The following example illustrates reset the `table.datalake.enabled` option to i
 ALTER TABLE my_table RESET ('table.datalake.enabled');
 ```
 
+`table.kv.row.ttl` cannot be reset to disable row-level TTL. To change or disable row-level TTL, create a new table with the desired TTL configuration and migrate the data.
```
wuchong: ditto

fluss-common/src/main/java/org/apache/fluss/config/ConfigOptions.java
```diff
@@ -1764,6 +1769,33 @@ public class ConfigOptions {
...
+                                    + "If not set, row-level TTL is disabled. "
+                                    + "Expired rows may remain visible until RocksDB compaction removes them.");
+
+    public static final ConfigOption<RowTtlChangelogMode> TABLE_KV_ROW_TTL_CHANGELOG_MODE =
+            key("table.kv.row.ttl.changelog-mode")
+                    .enumType(RowTtlChangelogMode.class)
+                    .defaultValue(RowTtlChangelogMode.NONE)
+                    .withDescription(
+                            "The changelog mode for row-level TTL cleanup. "
+                                    + "Only 'none' is supported in this version, which means TTL cleanup does not emit delete records.");
```
wuchong: I think we can simplify this to `table.kv.ttl.changelog=true/false`, as there is no changelog mode options for it (only emitting delete event, **or not**).

fluss-common/src/main/java/org/apache/fluss/config/ConfigOptions.java
```diff
@@ -1764,6 +1769,33 @@ public class ConfigOptions {
                                     + "Tables created before this option was introduced are treated as disabled. "
                                     + "Can be dynamically enabled via ALTER TABLE.");
 
+    public static final ConfigOption<Duration> TABLE_KV_ROW_TTL =
+            key("table.kv.row.ttl")
+                    .durationType()
+                    .noDefaultValue()
+                    .withDescription(
+                            "The best-effort row-level TTL for primary key tables. "
```
wuchong: Could we clarify that this option controls only the TTL of primary key data? The changelog retention of a primary key table is still controlled separately by `table.log.ttl`.

Suggested wording:

> The best-effort row-level TTL for primary key data. The changelog retention of a primary key table is controlled separately by `table.log.ttl`.

website/docs/table-design/data-distribution/ttl.md
```diff
@@ -8,6 +8,48 @@ sidebar_position: 3
 Fluss supports TTL for data by setting the TTL attribute for tables with `'table.log.ttl' = '<duration>'` (default is 7 days). Fluss can periodically and automatically check for and clean up expired data in the table.
 
 For log tables, this attribute indicates the expiration time of the log table data.
-For primary key tables, this attribute indicates the expiration time of the changelog and does not represent the expiration time of the primary key table data. If you also want the data in the primary key table to expire automatically, please use [auto partitioning](partitioning.md#auto-partitioning).
```
wuchong, line 11: I think it's still useful to retain the partition TTL document here.

fluss-common/src/main/java/org/apache/fluss/record/DefaultValueRecord.java
```diff
@@ -21,18 +21,21 @@
...
  * A value record is a tuple consisting of a value row and a schema id for the row.
  *
  * <p>The schema is as follows:
  *
  * <ul>
  *   <li>Length => int32
- *   <li>SchemaId => int16
- *   <li>Value => {@link BinaryRow}
+ *   <li>RawValue => schema id, optional internal fields, and {@link BinaryRow} as defined by {@link
+ *       KvValueLayout}
```
wuchong: **[P0]**: I don't suggest to change the wire protocol here, since 
1. the internal timestamp is not necessary for clients, 
2. it increases the network overhead 
3. it increase the compatibility compelxity of the clients. 

I think it's easy to transport only the `schema_id + value` even if we store additional system columns in RocksDB.

fluss-common/src/main/java/org/apache/fluss/record/ValueRecordBatch.java
```diff
@@ -71,5 +72,8 @@ interface ReadContext {
          * @param schemaId the schema of the kv records
          */
         RowDecoder getRowDecoder(int schemaId);
+
+        /** Gets the KV value layout used to locate row bytes in this value record batch. */
+        KvValueLayout getKvValueLayout();
```
wuchong: We don't need this if we don't change the kv record wire protocol.

fluss-common/src/main/java/org/apache/fluss/row/decode/FixedSchemaDecoder.java
```diff
@@ -48,24 +49,49 @@ public class FixedSchemaDecoder {
...
 
+    /** The raw value layout used to locate the row payload. */
+    private final KvValueLayout kvValueLayout;
+
     public FixedSchemaDecoder(KvFormat kvFormat, Schema sourceSchema, Schema targetSchema) {
+        this(
+                kvFormat,
+                sourceSchema,
+                targetSchema,
+                KvValueLayout.forKvFormatVersion(KV_FORMAT_VERSION_2));
```
wuchong: We don't need this as well, if we don't change the `ValueRecord` wire protocol.

fluss-server/src/main/java/org/apache/fluss/server/utils/RowTtlUtils.java
```diff
@@ -0,0 +1,52 @@
...
+    public static long validateAndCeilTtlDurationToMillis(Duration ttl) {
+        checkNotNull(ttl, "ttl must not be null.");
+        checkArgument(
+                ttl.compareTo(Duration.ZERO) > 0,
+                "'%s' must be positive.",
+                ConfigOptions.TABLE_KV_ROW_TTL.key());
+
+        try {
+            return Math.addExact(
+                    Math.multiplyExact(ttl.getSeconds(), 1000L),
```
wuchong: This conversion seems more complex than necessary. `Duration.toMillis()` already performs exact overflow checks and throws `ArithmeticException`; for a best-effort TTL, truncating sub-millisecond precision should be sufficient. Introducing a dedicated ceil-and-overflow helper also creates a different `Duration` conversion convention from the rest of the codebase, which may lead to unnecessary refactoring or inconsistent handling elsewhere. Could we keep the positive-value validation and simply use `ttl.toMillis()` here (wrapping its exception only if we need a friendlier configuration error)?

fluss-common/src/main/java/org/apache/fluss/row/encode/KvValueLayout.java
```diff
@@ -0,0 +1,183 @@
...
+    private static final int VALUE_TAG_OFFSET = SCHEMA_ID_OFFSET + SCHEMA_ID_LENGTH;
+    private static final int VALUE_TAG_LENGTH = 8;
+    private static final int NO_OFFSET = -1;
+    private static final KvValueLayout KV_FORMAT_VERSION_1_LAYOUT =
+            new KvValueLayout(
+                    HeaderKind.SCHEMA_ONLY, SCHEMA_ID_OFFSET, SCHEMA_ID_LENGTH, NO_OFFSET);
+    private static final KvValueLayout KV_FORMAT_VERSION_2_LAYOUT =
+            new KvValueLayout(
+                    HeaderKind.SCHEMA_ONLY, SCHEMA_ID_OFFSET, SCHEMA_ID_LENGTH, NO_OFFSET);
+    private static final KvValueLayout KV_FORMAT_VERSION_3_LAYOUT =
```
wuchong: Please add javadoc on each format version layout, like how we did in `org.apache.fluss.record.LogRecordBatchFormat#LOG_MAGIC_VALUE_V2`. We should also explain the byte order used for the elements, since there are different byte order in them.

fluss-common/src/main/java/org/apache/fluss/row/encode/KvValueLayout.java
```diff
@@ -0,0 +1,183 @@
...
+                || absoluteValueTagOffset > value.length - VALUE_TAG_LENGTH) {
+            throw new IndexOutOfBoundsException(
+                    "Cannot write value tag at offset "
+                            + absoluteValueTagOffset
+                            + " to value of length "
+                            + value.length
+                            + ".");
+        }
+        long bigEndianValueTag =
+                MemorySegment.LITTLE_ENDIAN ? Long.reverseBytes(valueTag) : valueTag;
```
wuchong, line 216: Could we add a comment here explaining why the timestamp is encoded in big-endian rather than following the little-endian encoding used by the schema ID and BinaryRow? FRocksDB’s native C++ `DeserializeTimestamp` reads these eight bytes most-significant-byte first, so as long as we use this compaction filter, the timestamp has to follow its big-endian contract. This mixed-endian layout is quite non-obvious and otherwise easy to “simplify” incorrectly later.

fluss-common/src/main/java/org/apache/fluss/row/encode/KvValueLayout.java
```diff
@@ -0,0 +1,183 @@
...
+            new KvValueLayout(
+                    HeaderKind.SCHEMA_WITH_VALUE_TAG,
+                    SCHEMA_ID_OFFSET,
+                    SCHEMA_ID_LENGTH + VALUE_TAG_LENGTH,
+                    VALUE_TAG_OFFSET);
+
+    private final HeaderKind headerKind;
+    private final int schemaIdOffset;
+    private final int rowPayloadOffset;
+    private final int valueTagOffset;
```
wuchong, line 54: What is the value tag mean? Why not just using `timestamp` to reference this which is very straightforward.

fluss-rpc/src/main/java/org/apache/fluss/rpc/protocol/ApiKeys.java
```diff
@@ -101,12 +106,12 @@ public enum ApiKeys {
...
     GET_PRODUCER_OFFSETS(1054, 0, 0, PUBLIC),
     DELETE_PRODUCER_OFFSETS(1055, 0, 0, PUBLIC),
-    ACQUIRE_KV_SNAPSHOT_LEASE(1056, 0, 0, PUBLIC),
+    ACQUIRE_KV_SNAPSHOT_LEASE(1056, 0, 1, PUBLIC),
     RELEASE_KV_SNAPSHOT_LEASE(1057, 0, 0, PUBLIC),
     DROP_KV_SNAPSHOT_LEASE(1058, 0, 0, PUBLIC),
     GET_TABLE_STATS(1059, 0, 0, PUBLIC),
     ALTER_DATABASE(1060, 0, 0, PUBLIC),
-    SCAN_KV(1061, 0, 0, PUBLIC),
+    SCAN_KV(1061, 0, 1, PUBLIC),
```
wuchong: I think all the APIKEYS don't need to bump version if we keep the `ValueRecord` wire protocol unchanged.

fluss-flink/fluss-flink-common/src/test/java/org/apache/fluss/flink/source/FlinkTableSourceBatchITCase.java
```diff
@@ -658,16 +657,40 @@ void testCountPushDownWithWALMode() throws Exception {
...
+                                        + "  'table.kv.row.ttl' = '1 h',"
+                                        + "  'table.kv.format-version' = '3')",
+                                tableName))
+                .await();
+
+        String query = String.format("SELECT COUNT(*) FROM %s", tableName);
+        assertThatThrownBy(() -> tEnv.explainSql(query))
+                .hasMessageContaining(
+                        String.format(
+                                "Row count is disabled for this table '%s.test_count_table_with_row_ttl' because row TTL cleanup does not maintain exact row count.",
```
wuchong: This case is failed.

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/source/FlinkTableSource.java
```diff
@@ -742,15 +742,16 @@ public boolean applyAggregates(
...
         if (streaming
                 || aggregateExpressions.size() != 1
                 || groupingSets.size() > 1
                 || (groupingSets.size() == 1 && groupingSets.get(0).length > 0)
                 // The count pushdown feature is not supported when the data lake is enabled.
                 // Otherwise, it'll cause miss count data in lake. But In the future, we can push
                 // down count into lake.
-                || isDataLakeEnabled) {
+                || isDataLakeEnabled
+                || !canPushDownRowCount()) {
```
wuchong: We shouldn't add this filter, otherwise, `org.apache.fluss.flink.source.FlinkTableSourceBatchITCase#testCountPushDownWithRowTTL` is failed.

fluss-server/src/main/java/org/apache/fluss/server/coordinator/CoordinatorService.java
```diff
@@ -812,20 +824,23 @@ private TableDescriptor applySystemDefaults(
...
             Map<String, String> newProperties = new HashMap<>(newDescriptor.getProperties());
-            Integer formatVersion =
-                    Configuration.fromMap(newProperties).get(ConfigOptions.TABLE_KV_FORMAT_VERSION);
+            Configuration newTableConf = Configuration.fromMap(newProperties);
+            Integer formatVersion = newTableConf.get(ConfigOptions.TABLE_KV_FORMAT_VERSION);
             if (formatVersion == null) {
-                // set current kv format version for default
+                int defaultKvFormatVersion =
+                        newTableConf.getOptional(ConfigOptions.TABLE_KV_ROW_TTL).isPresent()
+                                ? KV_FORMAT_VERSION_3
```
wuchong: Selecting V3 here breaks normal row-TTL primary-key tables: both `KeyEncoder.ofPrimaryKeyEncoder` and `KeyDecoder.ofPrimaryKeyDecoder` only accept KV format versions 1 and 2, and throw `UnsupportedOperationException` for version 3. The encoder is reached when creating an upsert writer or lookuper, and also during tablet recovery. Since V3 changes only the value layout, please make the key encoder and decoder handle V3 like V2, with corresponding coverage, before selecting it as the default here.

fluss-common/src/main/java/org/apache/fluss/config/TableConfig.java
```diff
@@ -41,6 +42,11 @@
 @PublicEvolving
 public class TableConfig {
 
+    /**
+     * Internal KV table property that binds the row TTL time column to a stable schema column id.
+     */
+    public static final String KV_ROW_TTL_TIME_COLUMN_ID_KEY = "table.kv.row.ttl.time-column-id";
```
wuchong: Do we need to persist this derived column ID as a hidden table property? Other column-dependent features, such as auto-increment columns, resolve the column ID from `Schema` at runtime instead of maintaining a second property. Rename, drop, and modify-column operations are not currently supported, so the configured column name remains stable under the schema evolution we support today. This key creates a second source of truth and already requires special-case handling in table validation. Could we resolve the TTL column from the latest schema for now, and address stable column references consistently for all column-dependent features in a separate issue when rename/drop support is introduced?

fluss-rpc/src/main/java/org/apache/fluss/rpc/protocol/ApiKeys.java
```diff
@@ -50,21 +50,23 @@ public enum ApiKeys {
...
 
     NOTIFY_LEADER_AND_ISR(1018, 0, 0, PRIVATE),
     STOP_REPLICA(1019, 0, 0, PRIVATE),
     ADJUST_ISR(1020, 0, 0, PRIVATE),
     LIST_OFFSETS(1021, 0, 0, PUBLIC),
     COMMIT_KV_SNAPSHOT(1022, 0, 0, PRIVATE),
-    GET_LATEST_KV_SNAPSHOTS(1023, 0, 0, PUBLIC),
-    GET_KV_SNAPSHOT_METADATA(1024, 0, 0, PUBLIC),
+    GET_LATEST_KV_SNAPSHOTS(1023, 0, 1, PUBLIC),
+    GET_KV_SNAPSHOT_METADATA(1024, 0, 1, PUBLIC),
```
wuchong, line 68: IIUC, currently, old clients don't validate the KV format version when parsing RocksDB data? In such cases, this `GET_KV_SNAPSHOT_METADATA` serves as our final line of defense. This is the only RPC I can see the reason to bump the version.

fluss-server/src/main/java/org/apache/fluss/server/coordinator/CoordinatorService.java
```diff
@@ -1351,7 +1387,8 @@ public CompletableFuture<AcquireKvSnapshotLeaseResponse> acquireKvSnapshotLease(
                 authorizeTable(OperationType.READ, tableId);
             }
 
-            validateKvTable(tableId);
+            TableInfo tableInfo = validateKvTable(tableId);
+            validateClientVersionForPkTable(ApiKeys.ACQUIRE_KV_SNAPSHOT_LEASE, tableInfo);
```
wuchong: Adding validation in `GET_KV_SNAPSHOT_METADATA` is enough. We don't need to check every kv operations, kv lease doesn't mean it is reading the kv snapshot.

fluss-server/src/main/java/org/apache/fluss/server/RpcServiceBase.java
```diff
@@ -382,6 +386,7 @@ public CompletableFuture<GetLatestKvSnapshotsResponse> getLatestKvSnapshots(
                             + tablePath
                             + "' is a partitioned table, but partition name is not provided.");
         }
+        validateClientVersionForPkTable(ApiKeys.GET_LATEST_KV_SNAPSHOTS, tableInfo);
```
wuchong: Adding validation in `GET_KV_SNAPSHOT_METADATA` is enough. We don't need to check every kv operations.

fluss-server/src/main/java/org/apache/fluss/server/replica/ReplicaManager.java
```diff
@@ -2439,29 +2443,68 @@ public void shutdown() throws InterruptedException {
         checkpointHighWatermarks();
     }
 
-    private void validateClientVersionForPkTable(int apiVersion, TableInfo tableInfo) {
-        if (apiVersion > 0) {
+    /**
+     * Returns whether this API version is new enough to access any primary-key table without
+     * reading table-specific compatibility metadata.
+     */
+    public static boolean canSkipClientVersionValidation(ApiKeys apiKey, int apiVersion) {
```
wuchong: Move them into dedicated util classes, this validation doesn't belong to `ReplicaManager`. `ReplicaManager` is already very big.

fluss-server/src/main/java/org/apache/fluss/server/kv/KvManager.java
```diff
@@ -460,6 +511,10 @@ public KvTablet loadKv(
         return kvTablet;
     }
 
+    private static int getKvFormatVersion(TableConfig tableConfig) {
+        return tableConfig.getKvFormatVersion().orElse(ConfigOptions.KV_FORMAT_VERSION_2);
```
wuchong: default should be V1, tables created before V2 introduced has no default value.

fluss-common/src/main/java/org/apache/fluss/config/ConfigOptions.java
```diff
@@ -1764,6 +1769,33 @@ public class ConfigOptions {
                                     + "Tables created before this option was introduced are treated as disabled. "
                                     + "Can be dynamically enabled via ALTER TABLE.");
 
+    public static final ConfigOption<Duration> TABLE_KV_ROW_TTL =
+            key("table.kv.row.ttl")
```
wuchong: Rename to `table.kv.ttl` to align with `table.log.ttl`, and also all the methods `..RowTTL()` to `..KvTTL()`. All the `TTL`, are row-level-based, the other one is `partition expiration` which is partition-level-based.

website/docs/table-design/data-distribution/ttl.md
```diff
@@ -8,6 +8,48 @@ sidebar_position: 3
...
+    name STRING,
+    PRIMARY KEY (id) NOT ENFORCED
+) WITH (
+    'bucket.num' = '4',
+    'table.kv.row.ttl' = '7 d',
+    'table.kv.row.ttl.time-column' = 'event_time'
+);
+```
+
+The event-time column must be `BIGINT` epoch milliseconds or `TIMESTAMP_LTZ`. Rows with null event-time values do not expire through row TTL.
```
wuchong: Why `TIMESTAMP` type is not supported?

wuchong (commented):
Hi @platinumhamburg , thanks for the great pull request. 

<html><head></head><body><p>I have a major compatibility concern with reusing <code dir="ltr">table.kv.format-version=3</code> for the new RocksDB value layout. This option currently represents the <strong>key encoding version</strong>: it is consumed by <code dir="ltr">KeyEncoder</code> and <code dir="ltr">KeyDecoder</code>, and existing clients only understand versions 1 and 2. Assigning version 3 to a new table would therefore break existing clients even though the key encoding itself has not changed.</p><p>I think we need to distinguish the following three independent compatibility dimensions:</p><div><div><div>

Dimension | Suggested name | Scope | Persisted table property?
-- | -- | -- | --
Key encoding | Keep `table.kv.format-version` in this PR; consider `table.kv.key-encoding-version` in a separate change | Encoding of keys used by clients, RPC requests, and RocksDB | Yes; persisted and immutable
RPC value encoding | Conceptually `table.kv.value-encoding-version` | Wire format used by Lookup/PutKV and other KV RPCs | No; negotiated through the RPC API version
RocksDB value layout | `table.kv.value-layout-version` | Physical layout of values stored in RocksDB | Yes, but preferably as a system-managed internal property

</div></div></div><p>These versions solve different compatibility problems:</p><ul><li>The <strong>key encoding version</strong> is an end-to-end table property because clients must encode keys in exactly the same way as the server and RocksDB. It must remain stable for the lifetime of the table.</li><li>The <strong>RPC value encoding version</strong> is a wire-protocol concern. It should be negotiated through the RPC API version rather than persisted in table metadata. Clients should continue to receive <code dir="ltr">[schema_id][value_bytes]</code> without needing to understand the server-side timestamp layout.</li><li>The <strong>RocksDB value layout version</strong> describes the server-side physical representation, such as version 1 being <code dir="ltr">[schema_id][value_bytes]</code> and version 2 being <code dir="ltr">[timestamp][schema_id][value_bytes]</code>. It needs to be persisted so the server can correctly interpret existing data, but application clients should normally not depend on it.</li></ul><p>For this PR, I suggest keeping <code dir="ltr">table.kv.format-version</code> unchanged with versions 1 and 2, and introducing only <code dir="ltr">table.kv.value-layout-version</code>. Existing tables without the new property would default to value layout version 1, while new TTL-enabled tables could use:</p><pre dir="ltr"><code>table.kv.format-version = 2
table.kv.value-layout-version = 2</code></pre><p>Before returning values through KV RPCs, the server should remove the storage-only timestamp and expose the existing <code dir="ltr">[schema_id][value_bytes]</code> representation. This keeps the RocksDB layout internal, avoids unnecessarily coupling RPC clients to the storage format, and preserves compatibility with existing lookup/write clients.</p></body></html>

**This also lays the foundation for dynamically enabling KV TTL, meaning that all tables will default to `value-layout-version` 2, regardless of whether KV TTL is enabled. However, this can be deferred to future work.**

fluss-common/src/main/java/org/apache/fluss/row/decode/FixedSchemaDecoder.java
platinumhamburg: The wire format remains unchanged. The remaining change only replaces a detached schema-id offset with KvValueLayout.PLAIN, so all value decoding constants stay centralized in KvValueLayout. Its decoding behavior is still the original plain RPC value layout.

fluss-common/src/main/java/org/apache/fluss/record/DefaultValueRecord.java
platinumhamburg: Addressed. The KV RPC value format remains [little-endian schema id][BinaryRow]. LONG_TAG is only a RocksDB storage layout. Lookup, prefix lookup, and scan responses use KvValueLayout offsets to exclude the internal tag when constructing the RPC payload, so the tag is neither exposed to clients nor transmitted over the wire. Snapshot reads are the exception because the client opens RocksDB directly, and that path is protected by GET_KV_SNAPSHOT_METADATA v1.

fluss-rpc/src/main/java/org/apache/fluss/rpc/protocol/ApiKeys.java
platinumhamburg: Addressed. The row-TTL-specific API version bumps were removed. Only GET_KV_SNAPSHOT_METADATA uses v1 as the compatibility boundary for clients that directly open RocksDB snapshots. PUT_KV v2 is an independent change already present on main for storage-backpressure error compatibility, not a row-TTL protocol change.

fluss-server/src/main/java/org/apache/fluss/server/coordinator/CoordinatorService.java
platinumhamburg: Addressed by separating the KV key format from the physical value layout. Row TTL no longer introduces KV format V3. table.kv.format-version continues to control key encoding, while the internal table.kv.value-layout-version selects PLAIN or LONG_TAG for RocksDB values.

fluss-rpc/src/main/java/org/apache/fluss/rpc/protocol/ApiKeys.java
platinumhamburg: Agreed. GET_KV_SNAPSHOT_METADATA is now the only row-TTL-specific RPC version boundary. A v0 client is rejected only when it attempts to read a tagged snapshot; ordinary RPC lookup and scan payloads retain their existing wire format.

fluss-server/src/main/java/org/apache/fluss/server/kv/KvManager.java
platinumhamburg: Addressed. Value-layout selection no longer falls back through the KV format version. Tables without the internal value-layout property resolve to KvValueLayout.PLAIN, preserving the physical layout of existing tables.

fluss-server/src/main/java/org/apache/fluss/server/replica/ReplicaManager.java
platinumhamburg: Addressed. The row-TTL-specific client-version validation was removed from this KV operation because the RPC value format remains unchanged.

fluss-server/src/main/java/org/apache/fluss/server/RpcServiceBase.java
platinumhamburg: Addressed. The row-TTL-specific client-version validation was removed from this KV operation because the RPC value format remains unchanged.

fluss-server/src/main/java/org/apache/fluss/server/coordinator/CoordinatorService.java
platinumhamburg: Addressed. The row-TTL-specific validation was removed. The remaining key-format validation is pre-existing behavior and is unrelated to row TTL.

fluss-common/src/main/java/org/apache/fluss/row/encode/KvValueLayout.java
platinumhamburg: Addressed. The value-layout version contract and byte order are now documented in the Javadocs.

fluss-common/src/main/java/org/apache/fluss/record/ValueRecordBatch.java
platinumhamburg: Addressed

website/docs/engine-flink/ddl.md
platinumhamburg: Addressed. Restored the original wording to state that the Fluss cluster dynamically applies the changes to the table.

website/docs/engine-flink/ddl.md
platinumhamburg: Addressed. Removed the row-TTL-specific restrictions from the SET and RESET documentation. The supported-options list remains the source of truth for options that support dynamic updates.

fluss-common/src/main/java/org/apache/fluss/config/ConfigOptions.java
platinumhamburg: Addressed. The option description now clarifies that table.kv.row.ttl controls primary-key data only, while changelog retention is controlled separately by table.log.ttl. The Flink connector options documentation has been updated consistently.

website/docs/table-design/data-distribution/ttl.md
platinumhamburg: Addressed. The TTL overview now retains an entry for auto-partition expiration and links to the existing Auto Partitioning documentation.

fluss-common/src/main/java/org/apache/fluss/row/encode/KvValueLayout.java
platinumhamburg: Addressed. Added Javadoc explaining that the long tag uses big-endian encoding to match FRocksDB's native DeserializeTimestamp contract, which reads the timestamp most-significant byte first.

fluss-flink/fluss-flink-common/src/test/java/org/apache/fluss/flink/source/FlinkTableSourceBatchITCase.java
platinumhamburg: Addressed. The test now verifies that COUNT(*) on a row-TTL table falls back to a normal Flink aggregate and table scan instead of failing during planning. It also executes the query and verifies the result.

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/source/FlinkTableSource.java
platinumhamburg: The filter is intentional because TTL compaction does not maintain an exact stored row count. The failing test was asserting the wrong behavior and has been updated to verify fallback without aggregate pushdown. Removing the filter would allow Flink to use an inaccurate row count for TTL tables.

fluss-common/src/main/java/org/apache/fluss/config/ConfigOptions.java
platinumhamburg: Addressed by removing the changelog-mode option entirely. TTL cleanup currently has one fixed behavior: it does not emit delete events. We can introduce table.kv.ttl.changelog when delete changelog support is implemented.

fluss-server/src/main/java/org/apache/fluss/server/utils/RowTtlUtils.java
platinumhamburg: Addressed. TTL conversion now uses Duration.toMillis(), requires the converted duration to be at least 1 ms, and wraps overflow with a configuration-friendly error. Sub-millisecond precision is truncated consistently with the standard Duration conversion.

fluss-common/src/main/java/org/apache/fluss/row/encode/KvValueLayout.java
platinumhamburg: 64-bit value because its meaning is defined by the compaction strategy rather than the physical layout. Row TTL currently stores a timestamp, while a future secondary-index partition TTL implementation may store a partition ID in the same field. Naming it timestamp would incorrectly couple the reusable value layout to the current TTL use case.

fluss-common/src/main/java/org/apache/fluss/config/TableConfig.java
platinumhamburg: Addressed. The hidden TTL time-column ID property and its special-case validation have been removed. The configured column name is resolved from the latest schema when the tablet opens. Rename, drop, and modify-column operations remain unsupported, so a general stable-column-reference solution can be introduced consistently when those operations are supported.

fluss-common/src/main/java/org/apache/fluss/config/ConfigOptions.java
platinumhamburg: Addressed. The options are now table.kv.ttl and table.kv.ttl.time-column, aligned with table.log.ttl. The corresponding TableConfig methods have also been renamed to getKvTTL() and getKvTTLTimeColumn().

website/docs/table-design/data-distribution/ttl.md
platinumhamburg: Addressed. Event-time TTL now supports BIGINT, TIMESTAMP, and TIMESTAMP_LTZ. TIMESTAMP is interpreted using the TabletServer system time zone, and the documentation states that all TabletServers must use the same system time zone. The behavior is covered by a non-UTC test.

platinumhamburg:
@wuchong Thanks for your detailed review feedback,  The current implementation responds to the P0 compatibility concern with the following model.

### Compatibility model

The implementation now separates the three compatibility dimensions discussed in the review:

1. **KV key encoding**

   `table.kv.format-version` continues to describe key encoding only. Row TTL does not introduce a new KV format version, so existing key encoders and decoders remain unchanged.

2. **RocksDB value layout**

   `table.kv.value-layout-version` is a system-managed internal table property describing the physical RocksDB value layout:

   - `PLAIN` (version 1): `[little-endian schema_id][BinaryRow]`
   - `TAGGED` (version 2): `[big-endian 64-bit tag][little-endian schema_id][BinaryRow]`

   Tables without this property default to `PLAIN`, preserving compatibility with existing tables. New non-TTL tables also use `PLAIN`; only tables that require the tag use `TAGGED`.

3. **KV RPC value encoding**

   The KV RPC wire format remains unchanged as `[schema_id][BinaryRow]`. The internal tag is neither exposed to RPC clients nor transmitted over the network.

   Lookup and prefix-lookup responses pass a slice of the RocksDB value directly to the protobuf serializer, excluding the storage-only prefix without allocating a temporary stripped byte array. Limit scan and scan responses perform the same projection while constructing the existing value-record representation.

Consequently, regular PUT, lookup, prefix lookup, limit scan, and scan APIs did not receive row-TTL-specific version bumps. Existing clients can continue to access ordinary tables and use the unchanged RPC paths against TTL tables.

The exception is direct KV snapshot reading, because that client opens RocksDB data and therefore must understand the physical value layout. `GET_KV_SNAPSHOT_METADATA` v1 is the only row-TTL-specific compatibility boundary. A v0 client is rejected only when it attempts to read a tagged snapshot; it can still read snapshots using the plain layout.

All physical value encoding and decoding offsets are centralized in `KvValueLayout`. The storage layout is not inferred from the key format or duplicated in individual read/write paths.

### Intentional differences from the suggestions

There are a few places where the implementation does not follow the suggestions literally:

1. **The field remains a generic value tag rather than being named `timestamp`.**

   Row TTL currently interprets the tag as a timestamp, but the physical layout does not own that semantic. Other compaction strategies may use the same layout for different metadata; for example, partition TTL for a secondary index may encode a partition ID. Keeping the layout generic avoids coupling a reusable storage contract to row TTL.

2. **Non-TTL tables continue to use `PLAIN` instead of preemptively using `TAGGED`.**

   Dynamic TTL enablement is not supported in this change. Adding an unused eight-byte tag to every ordinary KV value would impose permanent storage and encoding overhead without providing a current capability. A future design for dynamically enabling TTL can define the required layout transition explicitly.

3. **The Flink row-count pushdown guard is retained.**

   TTL compaction removes entries outside the normal KV write path and therefore does not maintain an exact stored row count. Allowing count pushdown would return incorrect results. The connector declines the pushdown for TTL tables, allowing Flink to execute a normal aggregate over the table scan instead of failing or returning an inaccurate count.

4. **No changelog boolean is introduced yet.**

   TTL cleanup currently has only one supported behavior: it does not emit delete records. An option whose only meaningful value is `false` would not provide a real choice. A changelog option can be introduced together with delete-event support when that behavior exists.

The remaining review suggestions were applied, including using `Duration.toMillis()`, resolving the event-time column from the schema, supporting `TIMESTAMP`, documenting byte order, retaining partition-expiration documentation, and limiting the RPC compatibility check to tagged snapshot reads. PTAL when you have time.

fluss-server/src/main/java/org/apache/fluss/server/coordinator/CoordinatorService.java
```diff
@@ -812,23 +825,22 @@ private TableDescriptor applySystemDefaults(
...
                         ConfigOptions.TABLE_KV_FORMAT_VERSION.key(),
                         String.valueOf(CURRENT_KV_FORMAT_VERSION));
-            } else {
-                if (formatVersion > CURRENT_KV_FORMAT_VERSION) {
-                    throw new InvalidConfigException(
-                            String.format(
-                                    "Unsupported kv format version %d. "
-                                            + "The maximum supported version is %d.",
-                                    formatVersion, CURRENT_KV_FORMAT_VERSION));
-                }
```
wuchong, line 829: It's still meaningful to keep this validation?

fluss-server/src/main/java/org/apache/fluss/server/log/LogTablet.java
```diff
@@ -895,11 +911,21 @@ private LogAppendInfo append(MemoryLogRecords records, boolean appendAsLeader)
...
+                        && assignedCommitTimestamp < localMaxTimestamp) {
+                    throw new IllegalArgumentException(
+                            String.format(
+                                    "Assigned commit timestamp %s is older than local maximum %s.",
+                                    assignedCommitTimestamp, localMaxTimestamp));
+                }
+                long commitTimestamp =
+                        assignedCommitTimestamp != null
+                                ? assignedCommitTimestamp
+                                : Math.max(localMaxTimestamp, clock.milliseconds());
```
wuchong: Is it strictly necessary to guarantee that the log commitTimestamp is exactly equal to the KV timestamp? This requirement appears to significantly increase complexity and introduce additional lock overhead. From my perspective, this level of strictness seems unnecessary.

fluss-server/src/main/java/org/apache/fluss/server/kv/KvTablet.java
```diff
@@ -567,8 +708,12 @@ public LogAppendInfo putAsLeader(
                     PaddingRow latestSchemaRow = new PaddingRow(latestRowType.getFieldCount());
                     // get offset to track the offset corresponded to the kv record
                     long logEndOffsetOfPrevBatch = logTablet.localLogEndOffset();
+                    long commitTimestamp = logTablet.nextCommitTimestamp();
```
wuchong: Is it strictly necessary to guarantee that the log commitTimestamp is exactly equal to the KV timestamp? This requirement appears to significantly increase complexity and introduce additional lock overhead. From my perspective, this level of strictness seems unnecessary.

Can we just get the current timestamp from `clock` here?

fluss-common/src/main/java/org/apache/fluss/config/TableConfig.java
```diff
@@ -77,6 +78,12 @@ public Optional<Integer> getKvFormatVersion() {
         return config.getOptional(ConfigOptions.TABLE_KV_FORMAT_VERSION);
     }
 
+    /** Gets the persisted KV value layout version. */
```
wuchong: Add more description to explain the difference bettween this and `getKvFormatVersion`.

fluss-server/src/main/java/org/apache/fluss/server/utils/TableDescriptorValidation.java
```diff
@@ -129,6 +139,12 @@ public static void validateTableDescriptor(
         checkDeleteBehavior(tableConf, hasPrimaryKey);
         checkTieredLog(tableConf);
         checkHistoricalPartition(tableDescriptor, tableConf);
+        checkKvTTL(tableConf, schema, hasPrimaryKey);
+        checkKvFormatVersion(tableConf);
+        checkKvValueLayout(
+                tableConf,
+                hasPrimaryKey,
+                tableConf.getOptional(ConfigOptions.TABLE_KV_TTL).isPresent());
```
wuchong: nit: move this into `checkKvValueLayout` which can be derived from `tableConf`.

website/docs/table-design/data-distribution/ttl.md
```diff
@@ -11,3 +11,51 @@ For log tables, this attribute indicates the expiration time of the log table da
...
+);
+```
+
+The event-time column must be `BIGINT` epoch milliseconds, `TIMESTAMP`, or `TIMESTAMP_LTZ`. `TIMESTAMP` values are interpreted in the TabletServer's system time zone, so all TabletServers must use the same system time zone. Rows with null event-time values do not expire through row TTL.
+
+Row TTL must be configured when creating the primary key table. Changing or disabling `table.kv.ttl`, or changing `table.kv.ttl.time-column`, with `ALTER TABLE ... SET` or `ALTER TABLE ... RESET` is not supported in this version.
+
+## Rolling Upgrades
+
+Existing tables remain usable during a rolling upgrade. Before creating a row-TTL table, upgrade every server so that all servers understand the tagged value layout used for row TTL. After a row-TTL table exists, downgrading to a version that does not understand that layout is not supported.
```
wuchong:
```suggestion
Before creating a row-TTL primary key table, upgrade every server to Fluss 1.0+ so that all servers understand the tagged value layout used for row TTL. After a row-TTL table exists, downgrading to a version < 1.0 that does not understand that layout is not supported.
```

The rolling upgrades section should be a sub-section of the KV TTL, so I removed the section title.

fluss-common/src/main/java/org/apache/fluss/config/InternalTableOptions.java
```diff
@@ -0,0 +1,48 @@
...
+
+package org.apache.fluss.config;
+
+import org.apache.fluss.annotation.Internal;
+
+import java.util.Optional;
+
+/** Internal table options persisted and managed by Fluss. */
+@Internal
+public final class InternalTableOptions {
```
wuchong: Could we treat `table.kv.value-layout-version` as a regular table config, similar to `table.kv.format-version`, instead of introducing the `InternalTableOptions` concept and the related special-case validation/filtering paths? This value is still persisted in `TableDescriptor.properties` and exposed through `getTableInfo`, so the current abstraction does not actually make it internal. The coordinator can overwrite and validate the value and reject ALTERs just as it does today. If we later need truly hidden system metadata, that should be designed separately and stored outside user-visible properties, for example as a dedicated field in the ZK table registration.

fluss-common/src/main/java/org/apache/fluss/config/InternalTableOptions.java
```diff
@@ -0,0 +1,48 @@
...
+
+    public static final ConfigOption<Integer> KV_VALUE_LAYOUT_VERSION =
+            ConfigBuilder.key("table.kv.value-layout-version")
```
wuchong: I do not think we need to introduce the concept of an internal option at this stage, as these settings are not truly internal. They remain visible to users through SHOW CREATE TABLE, TableInfo, and TableConfig. If we genuinely intend to support internal configurations, that should be addressed in a separate field rather than within `properties`. For now, I suggest removing the special handling logic for internal options and treating them consistently with standard options like `table.kv.format-version`.

wuchong (commented):
Thanks @platinumhamburg , I left some new comments.

platinumhamburg:
> Thanks @platinumhamburg , I left some new comments.

@wuchong Thanks, all comments addressed; please take another look when you have time.

fluss-rpc/src/main/java/org/apache/fluss/rpc/entity/LookupResultForBucket.java
```diff
@@ -18,40 +18,74 @@
...
 import java.util.List;
 
+import static org.apache.fluss.utils.Preconditions.checkArgument;
+
 /** Result of {@link LookupRequest} for each table bucket. */
 public class LookupResultForBucket extends ResultForBucket {
 
     private final List<byte[]> values;
+    @Nullable private final KvValueLayout kvValueLayout;
+    @Nullable private final List<KvValueLayout> kvValueLayouts;
```
wuchong: `LookupResultForBucket` currently carries both a shared `KvValueLayout` and a per-value `List<KvValueLayout>`. This makes the relationship between `values`, `kvValueLayout`, and `kvValueLayouts` difficult to understand and leaks RocksDB storage-layout details into upper layers.

I think raw values should be converted as soon as they leave RocksDB. For example, `KvTablet#multiGet`—and similarly `multiGetFromBufferOrKv`, `prefixLookup`, and `limitScan`—can use the `KvValueLayout` already owned by the tablet to return RPC-ready `ByteArraySlice` values.

From `KvTablet` upward, callers would only pass RPC-ready values. Then `LookupResultForBucket` and `PrefixLookupResultForBucket` would no longer need to carry layout information, and `ReplicaManager` would not need to reconstruct `KvValueLayout` from `TableConfig`. This establishes a clear boundary for raw storage values, prevents accidental misuse of plain `byte[]`, preserves zero-copy RPC serialization, and removes a significant amount of conversion plumbing.

wuchong:
I squashed and appended a commit to fix the above issue and improved some code. Please take a look. @platinumhamburg

platinumhamburg:
> I squashed and appended a commit to fix the above issue and improved some code. Please take a look. @platinumhamburg

LGTM. The storage boundary is much clearer after this refactoring.

PR #3623, opened 2026-07-08
[server] local segment ttl

### LINKED ISSUE #3728: [server] Support TTL-based cleanup for local tiered log segments
### Motivation

`table.log.ttl` defines how long log data should be retained, but inactive local segments kept by tiered storage are reclaimed primarily according to `table.log.tiered.local-segments` and remote-upload events. For tables that retain several local segments, historical or idle partitions can therefore keep data locally after its TTL has expired, causing unnecessary disk usage.

Local TTL enforcement also needs a time-driven trigger: when a table is idle, no later log or remote-manifest event may occur to initiate cleanup.

### Proposed changes

- Add a positive server option, `log.retention.check-interval`, to control how often local retention is checked.
- Periodically scan loaded log tablets and remove the contiguous prefix of inactive local segments whose maximum timestamp has exceeded `table.log.ttl`.
- Keep the active segment and respect the existing high-watermark, remote-log-end-offset, and minimum-retained-offset safety bounds.
- Preserve the existing `table.log.tiered.local-segments` cleanup behavior for uploaded segments.

### Acceptance criteria

- Expired inactive local segments are removed for both partitioned and non-partitioned tables.
- Cleanup can run without a new upload or remote-manifest update.
- The active segment is never removed by TTL cleanup.
- The retention-check interval must be greater than zero.
- Tables with log TTL disabled are unaffected.

### Related pull request

Implemented by #3623.

gyang94:
### Purpose

Closes #3728.

Enforce table.log.ttl for inactive local log segments retained by tiered storage, including idle tables where no new upload or remote-manifest event occurs.

### Brief change log

- Add log.retention.check-interval with a default of 5 minutes and require it to be greater than zero.
- Schedule periodic TTL cleanup across loaded log tablets.
- Delete only the contiguous prefix of expired inactive segments while preserving the active segment and existing safety offsets.
- Preserve the table.log.tiered.local-segments cleanup behavior for uploaded segments.

### Tests

- Add TieredLocalSegmentTTLTest coverage for partitioned and non-partitioned tables, missing remote-log-end offsets, and cleanup bounds.
- Add configuration validation coverage for the retention-check interval.

### API and Format

Adds a server configuration option. No public API or storage format changes.

### Documentation

No user-facing documentation was added.


fluss-server/src/main/java/org/apache/fluss/server/log/LogTablet.java
```diff
@@ -556,10 +594,12 @@ public void updateRemoteLogSize(long remoteLogSize) {
...
-
-            // try to delete these segments already exist in remote storage.
-            deleteSegmentsAlreadyExistsInRemote();
         }
+
+        // Try to delete segments that have already existed in remote storage on every retention
+        // pass. A remote TTL cleanup may leave the manifest end offset unchanged or decrease it,
+        // but locally retained inactive segments may have expired in the meantime.
+        deleteSegmentsAlreadyExistsInRemote();
     }
```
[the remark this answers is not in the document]
gyang94: It is an existing edge case which may not trigger `deleteSegmentsAlreadyExistsInRemote` (no matter newly added TTL or existing local-segment file counts). 
I'd suggest to handle this in another issue.

swuferhong:
Hi, @gyang94 I don't think local TTL cleanup should be coupled with remote log manifest updates. Local TTL is part of the local log's retention semantics, so it would be better to run it as a periodic task in `LogManager`, scanning all `LogTablets` independently of whether the remote manifest has changed.
The remote log should only provide the safety watermark indicating which segments have been uploaded successfully. Local deletion should still be bounded by the highest uploaded offset, the high watermark, and the minimum retain offset. same comments: https://github.com/apache/fluss/pull/3222#discussion_r3557551287

fluss-flink/fluss-flink-common/src/test/java/org/apache/fluss/flink/source/FlinkTableSourceBatchITCase.java
```diff
@@ -59,11 +59,14 @@
 abstract class FlinkTableSourceBatchITCase extends FlinkTestBase {
 
     static final String CATALOG_NAME = "testcatalog";
-    static final String DEFAULT_DB = "defaultdb";
     protected StreamTableEnvironment tEnv;
+    private String databaseName;
+    private boolean databaseCreated;
```
swuferhong, line 64: Why this class need modify?

fluss-server/src/main/java/org/apache/fluss/server/log/LogTablet.java
```diff
@@ -1196,17 +1231,29 @@ private List<LogSegment> deletableSegments(long endOffset) {
...
+        int tierProtectedStartIndex = logSegments.size() - tieredLogLocalSegments;
+        long now = clock.milliseconds();
+
+        for (int i = 0; i < logSegments.size() - 1; i++) {
+            if (logSegments.get(i + 1).getBaseOffset() > endOffset) {
                 break;
             }
+            if (i < tierProtectedStartIndex
+                    || isSegmentExpired(now, logSegments.get(i), logTtlMs)) {
+                deletableSegments.add(logSegments.get(i));
```
swuferhong: In the retention window, break as soon as the first non-deletable segment is encountered, so that only a single contiguous prefix is deleted. If timestamps are monotonic, this changes nothing.

fluss-server/src/main/java/org/apache/fluss/server/log/LogTablet.java
```diff
@@ -556,9 +594,6 @@ public void updateRemoteLogSize(long remoteLogSize) {
     public void updateRemoteLogEndOffset(long remoteLogEndOffset) {
         if (remoteLogEndOffset > this.remoteLogEndOffset) {
             this.remoteLogEndOffset = remoteLogEndOffset;
-
-            // try to delete these segments already exist in remote storage.
-            deleteSegmentsAlreadyExistsInRemote();
```
swuferhong, line 594: Why can this inline cleanup be removed? If we remove it, local segments won't be reclaimed even after they've already been moved to remote — they can only be cleaned up by TTL. That makes our "retain 2 local segments" (tieredLogLocalSegments) mechanism useless, and local disk will grow unbounded.

swuferhong (commented):
Thanks for your contribution. I left some comments.

swuferhong (approved):
LGTM +1. Hi @wuchong could you take a look?

zuston (commented):
Thanks for working on this. After reviewing the changes, I have a couple of questions:

1. Without this PR, if we set table.log.tiered.local-segments = 1, would that achieve the same goal as this PR?
2. It looks like this PR does not roll the active segment for historical/inactive partitions.

swuferhong:
> Thanks for working on this. After reviewing the changes, I have a couple of questions:
> 
> 1. Without this PR, if we set table.log.tiered.local-segments = 1, would that achieve the same goal as this PR?
> 2. It looks like this PR does not roll the active segment for historical/inactive partitions.

Hi, @zuston. Based on the offline discussion with @gyang94 , implementing TTL for the active segment in the short term is fairly tricky and requires a detailed design—but that doesn't mean it shouldn't be done. You can continue working on this in a follow-up PR.

Regarding whether it can be set to `table.log.tiered.local-segments = 1`: these are two completely different concepts. One is about how much local data to reserve as a local read cache, and the other is about how long data should be retained before TTL kicks in. Consider the following scenario: the user has an hourly-partitioned table where partitions are retained for 24 hours, the log TTL is set to 1 hour, and the number of local segments is 5 (to ensure that the most recent hour of appended data stays on the server). In this case, the historical 23 hours would incur a large amount of redundant storage—even though the user has configured the log TTL to 1 hour.

zuston:
thanks for your quick reply. @swuferhong 

> the user has an hourly-partitioned table where partitions are retained for 24 hours, the log TTL is set to 1 hour, and the number of local segments is 5 (to ensure that the most recent hour of appended data stays on the server). In this case, the historical 23 hours would incur a large amount of redundant storage—even though the user has configured the log TTL to 1 hour.

make sense if setting the large number of local segments reservation.

zuston (approved):

swuferhong:
> thanks for your quick reply. @swuferhong
> 
> > the user has an hourly-partitioned table where partitions are retained for 24 hours, the log TTL is set to 1 hour, and the number of local segments is 5 (to ensure that the most recent hour of appended data stays on the server). In this case, the historical 23 hours would incur a large amount of redundant storage—even though the user has configured the log TTL to 1 hour.
> 
> make sense if setting the large number of local segments reservation.

Because Fluss remote currently lacks column pruning capability, many users will still choose to configure `table.log.tiered.local-segments` to a large value. Regarding TTL for active segments, could you and @gyang94  come up with a proposal docs first?

zuston:
> Regarding TTL for active segments, could you and @gyang94 come up with a proposal docs first?

OK. I also want to cut off the storage cost for the long-term historical partition.

fluss-common/src/main/java/org/apache/fluss/config/ConfigOptions.java
```diff
@@ -852,6 +852,15 @@ public class ConfigOptions {
                                     + "flush which acts as the log recovery point. The default "
                                     + "setting is 60 seconds.");
 
+    public static final ConfigOption<Duration> LOG_RETENTION_CHECK_INTERVAL =
+            key("log.retention-check-interval")
```
wuchong: Could we rename this option to `log.retention.check-interval` and require it to be strictly positive?

Using `log.retention` as the common prefix provides a consistent namespace for future retention-related options, such as `log.retention.bytes` and `log.retention.idle-timeout`.

I also don't think `0` should disable this task. The periodic check is the only time-driven cleanup trigger for idle tables, so disabling it means `table.log.ttl` can no longer be enforced reliably when no further log or manifest events occur. Per-table retention can already be disabled with `table.log.ttl = -1`, and Kafka's analogous `log.retention.check.interval.ms` only accepts positive values. Unless we have a concrete use case for globally disabling retention checks, please validate this option as `> 0` and remove the `0`-disables semantics.

fluss-server/src/main/java/org/apache/fluss/server/log/LogTablet.java
```diff
@@ -556,10 +594,10 @@ public void updateRemoteLogSize(long remoteLogSize) {
...
         }
+
+        // try to delete these segments already exist in remote storage.
+        deleteSegmentsAlreadyExistsInRemote();
```
wuchong: We don't need to move this out of the if block?

wuchong (approved):

PR #3627, opened 2026-07-09
[client] Fix data hole in primary key table batch read

### LINKED ISSUE #3626: [Bug] Spark batch read on primary key table returns a data hole (missing newest rows)
### Search before asking

### Fluss version

0.9.0 (latest release)

### Please describe the bug 🐞

Batch-reading a primary key table from Spark can silently drop the newest rows, producing a "data hole". For a table continuously written with monotonically increasing primary keys, this query:
```
SELECT COUNT(*) AS total_count,
       MAX(seq_id) AS max_seq,
       MIN(seq_id) AS min_seq
FROM fluss_catalog.fluss.fluss_fault_test_pk;
```
returns a `total_count` smaller than `max_seq - min_seq + 1` — a whole batch of the latest records is missing.

This is a plain primary key table; datalake tiering is NOT enabled. Reading the same data from Flink (with a filter + limit) returns the rows correctly, so the data does exist in Fluss — only the Spark batch read loses it.

#### How to reproduce
1. Create a primary key table (no datalake).
2. Write a batch of rows (e.g. seq_id 1..100) and let a KV snapshot be taken.
3. Append a new batch whose keys are all greater than the snapshot max key
   (e.g. seq_id 101..200); these rows stay in the log tail.
4. Run the aggregation query above from Spark.

Expected: `[200, 200, 1]`
Actual:   `[100, 100, 1]`  ← rows 101..200 are missing.

#### Root cause

Reading a primary key table merges the KV snapshot (sorted by pk) with the log tail (change log) via `SortMergeReader`. The merge was driven solely by the snapshot iterator: once all snapshot rows were consumed it stopped, dropping the trailing change log records whose keys are greater than the max snapshot key (they were put back during sort-merge to wait for a larger snapshot row that never comes).

Spark's non-lake batch reader `FlussUpsertPartitionReader` calls `readBatch()` exactly once, so it hits this directly. The lake scanner calls `readBatch()` repeatedly and happened to emit the tail on a later call, which is why datalake-enabled tables and Flink are not affected.

### Solution

_No response_

beryllw:
### Purpose

Linked issue: close #3626

### Brief change log

- Fix silent data loss when batch-reading a primary key table from Spark: change log records whose keys are greater than the max snapshot key were dropped, so `COUNT(*) < max_seq - min_seq + 1`.
- Root cause is in `SortMergeReader`: the merge was driven by the snapshot iterator and stopped once the snapshot was exhausted, discarding trailing change log records. It surfaces on the non-lake Spark batch read path (`FlussUpsertPartitionReader`), which calls `readBatch()` only once.

Rewrite `SortMergeReader` as a single two-way merge iterator over the snapshot stream and the change log stream (peek both heads, advance the smaller key; for an equal key the change log overrides the snapshot; a delete removes the row).The merged output is now identical regardless of how many times `readBatch()` is called — Spark calls it once, the lake scanner calls it repeatedly — which removes the structural cause of the dropped tail and also unifies projection handling. The public API is unchanged, so no caller needs to be modified.

### DIFF
--- a/fluss-client/src/main/java/org/apache/fluss/client/table/scanner/SortMergeReader.java
+++ b/fluss-client/src/main/java/org/apache/fluss/client/table/scanner/SortMergeReader.java
@@ -26,30 +26,32 @@ import org.apache.fluss.utils.CloseableIterator;
 
 import javax.annotation.Nullable;
 
-import java.util.ArrayList;
 import java.util.Collections;
 import java.util.Comparator;
 import java.util.Iterator;
 import java.util.List;
 import java.util.NoSuchElementException;
 import java.util.PriorityQueue;
-import java.util.function.Function;
 
 /**
- * A sort merge reader to merge snapshot record and change log. Both of them should have the same
- * primary key encoding when comparing the keys.
+ * A sort merge reader that merges snapshot records and change log records into a single stream
+ * sorted by primary key. Both inputs must share the same primary key encoding.
+ *
+ * <p>The merge is driven by a single {@link MergeIterator} (a two-way merge). For the same key the
+ * change log wins over the snapshot: an update overrides the value, a delete removes the row. Using
+ * one iterator keeps the result identical regardless of how many times {@link #readBatch()} is
+ * called, so trailing change log records (keys greater than the max snapshot key) are never
+ * dropped.
  */
 public class SortMergeReader {
 
     private final ProjectedRow snapshotProjectedPkRow;
     private final CloseableIterator<LogRecord> snapshotRecordIterator;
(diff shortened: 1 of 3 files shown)


wuchong:
cc @Yohahaha could you help to review this?

Yohahaha (approved):
thank you for the fix!

the SortMergeReader looks more robust to me after extract two iters into one merge iter.

wuchong (approved):
LGTM

I think the Copilots' review comment is not critical, because existing callers either fully consume the iterator before closing it or close the scanner when terminating early.

luoyuxia (approved):
+1

PR #3629, opened 2026-07-10
[common] Fix equals/hashCode contract violation in Schema.PrimaryKey

### LINKED ISSUE #3628: [common] Schema.PrimaryKey violates the equals/hashCode contract
### Search before asking

### Fluss version

0.9.0 (latest release)

## Summary

`Schema.PrimaryKey.equals()` compares only the primary-key column names, but `hashCode()` folds in `super.hashCode()`:

```java
// fluss-common/src/main/java/org/apache/fluss/metadata/Schema.java
    ...
    return Objects.equals(columnNames, that.columnNames);   // value-based
}

public int hashCode() {
    return Objects.hash(super.hashCode(), columnNames);      // includes Object.hashCode()
}
```

`PrimaryKey` extends only `Object`, so `super.hashCode()` is the per-instance `Object.hashCode()` value. Two `PrimaryKey` instances with the same columns are therefore `equals()` but return different hash codes, breaking the general contract of `Object.hashCode()`:

`Schema.hashCode()` includes the primary key, so the violation propagates to `Schema`, and to the public `TableDescriptor`, whose `equals`/`hashCode` include the schema. `Schema` is `@PublicEvolving` and `PrimaryKey` is `@PublicStable`, so any caller that puts a schema or table descriptor into a `HashSet`/`HashMap` is affected. This is easy to hit after a schema round-trips through its JSON form (`Schema#toJsonBytes` / `fromJsonBytes`, which is how Fluss persists and reloads a schema, e.g. via `SchemaZNode`): the reloaded schema is `equals()` to the in-memory one but is not found in a hash-based collection.

## How to reproduce

```java
Schema schema = Schema.newBuilder()
        .column("id", DataTypes.INT()).column("name", DataTypes.STRING())
        .primaryKey("id").build();

Schema reloaded = Schema.fromJsonBytes(schema.toJsonBytes());

schema.equals(reloaded);                    // true
schema.hashCode() == reloaded.hashCode();   // false  <- contract violation

Set<Schema> set = new HashSet<>();
set.add(schema);
set.contains(reloaded);                     // false  <- equal schema not found
```

Standalone runnable reproduction against the released `0.9.1-incubating` artifact (no cluster, no Fluss build needed): https://github.com/NestDream/fluss-pk-hashcode-demo. Run `./run-before.sh`; the core check is [SchemaHashCodeDemo.java#L30-L37](https://github.com/NestDream/fluss-pk-hashcode-demo/blob/main/src/main/java/io/github/nestdream/flussdemo/SchemaHashCodeDemo.java#L30-L37), with captured before/after logs under [`evidence/`](https://github.com/NestDream/fluss-pk-hashcode-demo/tree/main/evidence).

## Root Cause

The sibling types in the same file, `Schema.hashCode()` and `Column.hashCode()`, hash only their value fields with no identity term; `PrimaryKey` is the only one that folds in `super.hashCode()`, which points to a slip rather than a design choice. (`constraintName` is intentionally excluded from `equals()`: it is a derived, non-persisted name that is regenerated on deserialize by `SchemaJsonSerde`. So the correct alignment is to hash the same field `equals()` compares.)

### Solution

Compute `hashCode()` over the same field `equals()` compares:

```java
public int hashCode() {
    return Objects.hash(columnNames);
}
```

I have the one-line fix ready with a unit test (JSON round-trip) and an end-to-end ITCase (create table via the Admin API, read the schema back, use it as a hash key), both failing before the fix and passing after.

NestDream:
### Purpose

Linked issue: close #3628

`Schema.PrimaryKey.hashCode()` folds in `super.hashCode()` (the `Object` per-instance hash) while `equals()` compares only `columnNames`, so two value-equal primary keys return different hash codes. This violates the `Object.equals`/`hashCode` contract and propagates through `Schema.hashCode()` and the public `TableDescriptor`, so an equal schema/table descriptor can be missing from a `HashSet`/`HashMap` (e.g. after a `toJsonBytes`/`fromJsonBytes` round-trip). Standalone reproduction against the released 0.9.1-incubating artifact: https://github.com/NestDream/fluss-pk-hashcode-demo

### Brief change log

- `Schema.PrimaryKey.hashCode()`: `Objects.hash(super.hashCode(), columnNames)` → `Objects.hash(columnNames)`, matching the field `equals()` compares and the value-based hashing already used by `Schema` and `Column`.

### Tests

- `TableSchemaTest#testPrimaryKeySchemaSurvivesJsonRoundTripAsHashKey` (unit): round-trips a primary-key schema via `toJsonBytes`/`fromJsonBytes` and asserts the reloaded schema hashes equally and is found in a `HashSet<Schema>`; also asserts two primary keys with equal columns but different constraint names are equal and hash-equal. Fails before the fix, passes after.
- `FlussAdminITCase#testPrimaryKeySchemaReadBackIsUsableAsHashKey` (end-to-end): creates a primary-key table via `Admin.createTable`, reads the schema back via `Admin.getTableSchema`, and asserts it is usable as a hash key. Fails before the fix, passes after.
- Existing `TableSchemaTest`, `TableDescriptorTest`, `SchemaJsonSerdeTest` continue to pass (`mvn verify -pl fluss-common`: 1684 tests, 0 failures).

### API and Format

No public API or storage format change. `hashCode()` is never persisted or transmitted (`SchemaJsonSerde` writes explicit fields; on-the-wire schema identity is the separate `schemaId`).

### Documentation

Assisted by: Claude Code (Fable 5) following [the guidelines](https://github.com/apache/fluss/blob/main/AGENTS.md)

### DIFF
--- a/fluss-client/src/test/java/org/apache/fluss/client/admin/FlussAdminITCase.java
+++ b/fluss-client/src/test/java/org/apache/fluss/client/admin/FlussAdminITCase.java
@@ -102,10 +102,12 @@ import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 import java.util.OptionalLong;
+import java.util.Set;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ExecutionException;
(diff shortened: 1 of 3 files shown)


NestDream:
Hi @fresh-borzoni @wuchong, could you take a look at this PR? It fixes an equals/hashCode contract violation in `Schema.PrimaryKey` (#3628) And it is an one-line fix plus a regression test. CI also needs first-time-contributor approval to run. Thanks!

Jackeyzhe:
Thanks for this PR, LGTM

PR #3630, opened 2026-07-10
[lookup] Support historical partition lookup for lake tables

### LINKED ISSUE #3632: [FIP-28] Support point lookup for historical partitions
### Search before asking

### Description

Parent issue: #3631

Support point lookup for historical partitions of datalake-enabled primary key tables. When a lookup targets an expired Fluss partition whose data has been tiered to lake storage, the client/server path should resolve the historical partition and read the matching row from the lake table instead of failing with partition-not-found behavior.

The implementation should cover:

- Client lookup request plumbing for carrying the original partition name when the active Fluss partition no longer exists.
- Server-side routing that distinguishes normal lookup from historical partition lookup.
- A lake lookup SPI and Paimon implementation for point lookup by primary key.
- Schema evolution behavior when old lake data is returned through the current Fluss table schema.
- Flow control for historical lake lookup so it cannot exhaust normal lookup resources.
- Unit and integration coverage for normal lookup, historical lookup, mixed routing, Paimon lookup, and schema evolution.

Related PR: #3630

### Willingness to contribute

luoyuxia:
### Purpose

Linked issue: close #3632

Part of #3631.

This PR adds historical partition lookup support for datalake-enabled primary key tables. It lets clients resolve an expired Fluss partition to the corresponding lake partition and serve point lookups from lake storage when the partition is no longer available in Fluss.

### Brief change log

- Add historical partition metadata utilities and client-side lookup plumbing for requests carrying partition names.
- Add lake lookup SPI plus the Paimon implementation, including key conversion, row encoding, schema evolution padding, and lookup cache temp context.
- Route historical lookup requests through TabletServer/ReplicaManager to lake lookup instead of normal KV lookup.
- Add dedicated historical lookup admission control with `netty.server.max-queued-historical-requests` and a retriable throttle error.
- Add client-side backoff for historical lookup throttling without blocking normal lookup dispatch.
- Keep the Rust lookup client compatible with the updated lookup RPC schema.

### Tests

- `./mvnw test -pl fluss-rpc -am -Dtest=ApiErrorTest -DfailIfNoTests=false`
- `./mvnw test -pl fluss-client -am -Dtest=LookupSenderTest -DfailIfNoTests=false`
- `./mvnw test -pl fluss-server -am -Dtest=HistoricalLakeLookupManagerTest -DfailIfNoTests=false`
- `./mvnw test -pl fluss-server -am -Dtest=ReplicaFetcherThreadTest -DfailIfNoTests=false`
- `./mvnw spotless:check -pl fluss-common,fluss-rpc,fluss-client,fluss-server`
- `cargo fmt --all -- --check` from `fluss-rust/`
- `cargo check -p fluss-rs` from `fluss-rust/`

Full affected-module `verify` was started locally, but the environment hit unrelated local-state issues before the replica fetcher test was made independent from host disk usage.

### API and Format

- Adds `LakeTableLookuper`/lookup context APIs for lake lookup implementations.
- Adds `partition_name` to lookup RPC bucket requests for historical lookup routing.
- Adds `HISTORICAL_LOOKUP_THROTTLED` RPC error mapping.
- Adds `netty.server.max-queued-historical-requests` server config.

### Documentation

N/A. The FIP-28 design documents are intentionally not included in this PR diff.

### Generative AI disclosure


zuston:
Thanks for proposing this. @luoyuxia  I’m also exploring this area with Fluss + Paimon.

Regarding the proposed client fallback to server-side historical partition lookup, I’m a bit concerned that this could be risky. The lake lookup operation can be much slower than a local RocksDB lookup, especially when Paimon is backed by HDFS with poor performance. If this lookup is executed on the tablet server side, it may slow down Netty worker threads and affect the latency of other RPC requests.

For the case without historical data ingestion, I think this could potentially be handled by direct client-side lake lookup instead. On the Fluss side, we could consider introducing a Flink hybrid lake lookup async function for this purpose.
For the case with historical data ingestion, client-side lookup could still be used as a fallback when the historical partition + key is not found on the tablet server side.

Overall, I would prefer using client-side PK lookup where possible, to avoid overloading tablet servers during historical data ingestion bursts.

luoyuxia:
> Thanks for proposing this. @luoyuxia I’m also exploring this area with Fluss + Paimon.
> 
> Regarding the proposed client fallback to server-side historical partition lookup, I’m a bit concerned that this could be risky. The lake lookup operation can be much slower than a local RocksDB lookup, especially when Paimon is backed by HDFS with poor performance. If this lookup is executed on the tablet server side, it may slow down Netty worker threads and affect the latency of other RPC requests.
> 
> For the case without historical data ingestion, I think this could potentially be handled by direct client-side lake lookup instead. On the Fluss side, we could consider introducing a Flink hybrid lake lookup async function for this purpose. For the case with historical data ingestion, client-side lookup could still be used as a fallback when the historical partition + key is not found on the tablet server side.
> 
> Overall, I would prefer using client-side PK lookup where possible, to avoid overloading tablet servers during historical data ingestion bursts.

Thanks for sharing your thoughts. First, I’d like to clarify that the expensive Paimon lookup does not run on the Netty worker thread. It is handed off to the TabletServer IO executor through CompletableFuture.supplyAsync. Its concurrency is bounded by a semaphore, and excess requests receive a retriable throttling error with client-side backoff.

I agree that historical lookup still consumes additional TabletServer CPU, memory, network IO, and local disk for the Paimon lookup cache. This is a resource trade-off, even though it does not block the Netty event loop.

An important motivation for server-side historical lookup is historical data ingestion. We want to support writing into historical partitions while still producing the correct changelog. The server-side write path therefore needs to retrieve the previous value from historical storage before applying a write. Some server-accessible historical lookup infrastructure is required independently of the client fallback in this PR; the client lookup path reuses the same infrastructure.

Server-side lookup also keeps the generic Fluss lookup API independent of a particular lake implementation. Clients do not need to manage Paimon/Hadoop dependencies, lake credentials, catalog configuration, key encoding, snapshot selection, and fallback semantics.

A Flink hybrid lake lookup async function would be useful for Flink workloads, but Fluss also has Java, Rust, CPP, Python clients. Requiring every client to implement its own lake lookup and fallback logic would duplicate behavior and would be particularly difficult for Rust and Python clients when the lake implementation is JVM-based.

I think the two approaches can coexist, similar to Paimon’s local lookup and remote query service models. A remote lookup service can provide generic lookup capability for historical ingestion and all Fluss clients. The current PR runs this capability on TabletServers, but it could potentially be moved to a dedicated service later for better resource isolation. Client-side local lookup can then be introduced as an optional optimization where practical, with Flink async lookup as one possible first integration.

For this PR, my preference is to keep the bounded asynchronous server-side path because it also supports historical ingestion semantics, and explore client-side lookup as an optional strategy in a follow-up.

zuston:
> I think the two approaches can coexist, similar to Paimon’s local lookup and remote query service models.

+1. 

> For this PR, my preference is to keep the bounded asynchronous server-side path because it also supports historical ingestion semantics, and explore client-side lookup as an optional strategy in a follow-up.

After this PR is merged, let me reuse the lake lookup interface to implement the client side fallback mechanism.

fluss-lake/fluss-lake-paimon/src/main/java/org/apache/fluss/lake/paimon/lookup/PaimonLakeTableLookuper.java
```diff
@@ -0,0 +1,359 @@
...
+    private org.apache.paimon.data.BinaryRow wrapLookupKey(byte[] key) {
+        org.apache.paimon.data.BinaryRow keyRow =
+                new org.apache.paimon.data.BinaryRow(primaryKeyFieldCount);
+        keyRow.pointTo(MemorySegment.wrap(key), 0, key.length);
+        return keyRow;
+    }
+
+    private void initializeFilesIfNeeded(org.apache.paimon.data.BinaryRow partition, int bucketId) {
+        PaimonPartitionBucket partitionBucket = new PaimonPartitionBucket(partition, bucketId);
+        if (initializedBuckets.contains(partitionBucket)) {
```
fresh-borzoni, line 303: Looks like we never refresh the file set after the first scan. If tiering commits more data into this bucket later, lookups will miss it and the expiry never fires on a hot partition. 

Should we check the snapshot id and rescan when it changes?

fluss-lake/fluss-lake-paimon/src/main/java/org/apache/fluss/lake/paimon/lookup/PaimonLakeTableLookuper.java
```diff
@@ -0,0 +1,359 @@
...
+    }
+
+    private void ensureValueEncoder(short schemaId, RowType valueRowType) {
+        if (hasCachedValueEncoder && cachedValueSchemaId == schemaId) {
+            return;
+        }
+
+        IOUtils.closeQuietly(cachedValueRowEncoder, "Fluss value row encoder");
+        DataType[] fieldTypes = valueRowType.getChildren().toArray(new DataType[0]);
+        cachedValueRowEncoder = new CompactedRowEncoder(fieldTypes);
```
fresh-borzoni: nit: The client decodes lookup values with the table's kv format, but here we always encode compacted. 
What happens for a table with table.kv.format = indexed?

fluss-lake/fluss-lake-paimon/src/main/java/org/apache/fluss/lake/paimon/lookup/PaimonLakeTableLookuper.java
```diff
@@ -0,0 +1,359 @@
...
+    public PaimonLakeTableLookuper(
+            Configuration paimonConfig, TablePath tablePath, String ioTmpDir) {
+        this.paimonConfig = checkNotNull(paimonConfig, "paimonConfig must not be null.");
+        this.tablePath = checkNotNull(tablePath, "tablePath must not be null.");
+        this.ioTmpDir = checkNotNull(ioTmpDir, "ioTmpDir must not be null.");
+        this.initializedBuckets = new HashSet<>();
+    }
+
+    @Override
+    public synchronized @Nullable byte[] lookup(byte[] key, LookupContext context)
```
fresh-borzoni, line 134: This synchronizes all lake lookups of a table on one lock, and each waiting lookup holds an ioExecutor thread. 
That pool is size 10? by default and also does snapshot/remote-log transfers. Should historical lookups use their own pool?

fluss-lake/fluss-lake-paimon/src/main/java/org/apache/fluss/lake/paimon/lookup/PaimonLakeTableLookuper.java
```diff
@@ -0,0 +1,359 @@
...
+ * lookup I/O failure closes the cached query state, reopens the table from the latest snapshot, and
+ * retries once.
+ *
+ * <p>Lookup and close operations are synchronized because they share mutable Paimon query, local
+ * cache, and value-encoding state.
+ */
+public class PaimonLakeTableLookuper implements LakeTableLookuper {
+
+    // Hard-code a conservative 1GB default for now to avoid unbounded lookup cache growth.
+    private static final String DEFAULT_LOOKUP_CACHE_MAX_DISK_SIZE = "1gb";
```
fresh-borzoni: nit: this is per table, so 20 lake tables can use 20gb of tmp disk. Do we need a global cap?

fluss-server/src/main/java/org/apache/fluss/server/replica/HistoricalLakeLookupManager.java
```diff
@@ -0,0 +1,346 @@
...
+                    String.format(
+                            "Invalid original partition name %s for historical lookup on table %s.",
+                            originalPartitionName, tablePath));
+        }
+
+        LakeTableLookuper.LookupContext lookupContext =
+                new LakeTableLookuper.LookupContext(
+                        originalPartitionSpec,
+                        tableBucket.getBucket(),
+                        (short) tableInfo.getSchemaInfo().getSchemaId(),
```
fresh-borzoni: Replica.tableInfo is frozen when the replica is created. After adding a column, won't this keep encoding with the old schema and return null for the new column? 
Maybe use replica.getSchemaGetter() instead?

fluss-rpc/src/main/proto/FlussApi.proto
```diff
@@ -946,6 +946,8 @@ message PbLookupReqForBucket {
   optional int64 partition_id = 1;
   required int32 bucket_id = 2;
   repeated bytes keys = 3;
+  // The original partition name for historical lookup. It is unset for normal lookup.
+  optional string original_partition_name = 4;
```
fresh-borzoni, line 955: An old tablet server ignores this field and looks up the empty historical rocksdb and returns null instead of an error. During rolling upgrades that's silent misses. Gate on api version instead?

fluss-server/src/main/java/org/apache/fluss/server/coordinator/CoordinatorService.java
```diff
@@ -723,16 +725,36 @@ public CompletableFuture<CreatePartitionResponse> createPartition(
 
         // first, validate the partition spec, and get resolved partition spec.
         PartitionSpec partitionSpec = getPartitionSpec(request.getPartitionSpec());
-        validatePartitionSpec(tablePath, table.partitionKeys, partitionSpec, true);
+        ResolvedPartitionSpec partitionToCreate;
+        if (isHistoricalPartitionCreate(table, partitionSpec)) {
+            // Historical system partitions are lookup metadata, so creating one requires the same
+            // permission as reading the table instead of writing table data.
+            authorizeTable(OperationType.READ, tablePath);
```
fresh-borzoni: Why READ? A read-only client can now create partitions and for multi-key tables the prefix values come from the lookup key, so bad keys create real partitions until max.partition.num is full and auto-partitioning breaks.

fluss-client/src/main/java/org/apache/fluss/client/lookup/LookupSender.java
fresh-borzoni, line 339: Q: Historical lookups share this semaphore with normal ones, and lake lookups are slow, they can hold all the permits and block normal lookups. The FIP splits this into two semaphores (client.lookup.historical-inflight-ratio, historical capped at ~10%). Was dropping that intentional?

fluss-client/src/main/java/org/apache/fluss/client/lookup/PrimaryKeyLookuper.java
```diff
@@ -115,19 +125,89 @@ public CompletableFuture<LookupResult> lookup(InternalRow lookupKey) {
...
+
+    /** Falls back to historical lookup if the lookup key belongs to a historical partition. */
+    private CompletableFuture<LookupResult> mayFallbackToHistoricalLookup(
+            int bucketId, byte[] keyBytes, InternalRow lookupKey) {
+        String originalPartitionName = partitionGetter.getPartition(lookupKey);
+        if (!isHistoricalLookupCandidatePartition(
+                tableInfo, originalPartitionName, Instant.now())) {
+            return CompletableFuture.completedFuture(new LookupResult(Collections.emptyList()));
+        }
+        metadataUpdater.invalidPhysicalTableBucketAndPartitionMeta(
```
fresh-borzoni, line 158: This happens on every lookup to an expired partition, right? Each one does a failed metadata RPC plus a full Cluster copy here. 

Can we cache the "this partition is expired" decision?

fluss-client/src/main/java/org/apache/fluss/client/lookup/HistoricalPartitionResolver.java
```diff
@@ -0,0 +1,186 @@
...
+                return result;
+            }
+        } catch (Throwable t) {
+            result.completeExceptionally(t);
+            return result;
+        }
+
+        // If the historical system partition still does not exist, create it idempotently and
+        // refresh metadata again so the caller can route the lookup request by partition id.
+        admin.createPartition(
```
fresh-borzoni: Creating a partition from the lookup path feels backwards: a read allocates real replicas and rocksdb that may never see a write. Could we route lookups without the partition existing, and leave creation to the write path?

fluss-server/src/main/java/org/apache/fluss/server/replica/ReplicaManager.java
```diff
@@ -978,6 +1124,7 @@ public void stopReplicas(
 
                     for (StopReplicaData data : stopReplicaDataList) {
                         TableBucket tb = data.getTableBucket();
+                        historicalLakeLookupManager.invalidateTableLookuper(tb.getTableId());
```
fresh-borzoni: minor: This closes the lookuper on the calling thread, and close() waits for the lock held by an in-flight lookup. So stopping a replica can block on a slow S3 read. Mb move the close to the ioExecutor?

fluss-server/src/main/java/org/apache/fluss/server/coordinator/AutoPartitionManager.java
```diff
@@ -525,8 +526,15 @@ private void dropPartitions(
                 dropIterator = entry.getValue().iterator();
             }
 
+            boolean shouldRemoveEntry = true;
             while (dropIterator.hasNext()) {
                 String partitionName = dropIterator.next();
+                if (isHistoricalPartitionName(
```
fresh-borzoni: Related to dropping here, drop doesn't wait for tiering to catch up. If a partition is dropped before its tail is tiered, historical lookups quietly serve older values. Should drop wait for the tiering offset, like local log deletion waits for remote copy?

fluss-server/src/main/java/org/apache/fluss/server/replica/ReplicaManager.java
```diff
@@ -774,6 +871,49 @@ public void lookups(
         lookups(false, null, null, entriesPerBucket, apiVersion, responseCallback);
     }
 
+    /** Lookup historical lake data for requests that carry original partition names. */
+    public void historicalLookups(
```
fresh-borzoni, line 812: This goes straight to lake, but the FIP says buffer -> historical rocksdb -> lake. How will the write-path PR plug that in here?

fresh-borzoni (commented):
@luoyuxia Thank you, massive feature, left some comments and questions, PTAL

fluss-client/src/main/java/org/apache/fluss/client/lookup/PrimaryKeyLookuper.java
```diff
@@ -115,19 +125,89 @@ public CompletableFuture<LookupResult> lookup(InternalRow lookupKey) {
...
         int bucketId = bucketingFunction.bucketing(bkBytes, numBuckets);
         TableBucket tableBucket = new TableBucket(tableInfo.getTableId(), partitionId, bucketId);
+        return lookupBucket(tableBucket, pkBytes, insertIfNotExists, null, lookupKey);
+    }
+
+    /** Falls back to historical lookup if the lookup key belongs to a historical partition. */
+    private CompletableFuture<LookupResult> mayFallbackToHistoricalLookup(
+            int bucketId, byte[] keyBytes, InternalRow lookupKey) {
+        String originalPartitionName = partitionGetter.getPartition(lookupKey);
+        if (!isHistoricalLookupCandidatePartition(
```
platinumhamburg: `tableInfo` is fixed when the `Lookuper` is created, but `table.auto-partition.num-retention` can be changed while the lookuper is running.

Suppose a client creates a lookuper with retention set to 7. On January 10, retention is changed to 1, so the Coordinator removes partition `20240105`. The existing lookuper still checks `20240105` with retention 7 and returns an empty result, while a client using the updated table info sends the same lookup through the historical path.

Can we use the current table configuration here, or let the Coordinator make this decision? The current IT changes retention before creating the lookuper, so it does not cover this case.

fluss-client/src/main/java/org/apache/fluss/client/lookup/PrimaryKeyLookuper.java
```diff
@@ -115,19 +125,89 @@ public CompletableFuture<LookupResult> lookup(InternalRow lookupKey) {
...
+                tableInfo, originalPartitionName, Instant.now())) {
```
platinumhamburg: The Coordinator removes expired partitions using the CoordinatorServer clock, but this code calculates the retention boundary again using the client clock.

For an hourly partitioned table with retention set to 1, the Coordinator may be just past 10:00 and remove partition `2024010108`, while a client is still just before 10:00. The client refreshes metadata and sees that the partition is gone, but this check still treats it as retained and returns an empty result. A few seconds later, the same lookup starts using the historical path without any data or metadata change.

Can this decision use the Coordinator's time instead of `Instant.now()` on each client? A test with the Coordinator and client on opposite sides of an hour boundary would cover this case.

platinumhamburg (changes requested):

fluss-client/src/main/java/org/apache/fluss/client/lookup/LookupSender.java
luoyuxia: historical lookup is expected to be quite little, after offine discussion, we tend to keep the code simple to remove client.lookup.historical-inflight-ratio in current pr

fluss-lake/fluss-lake-paimon/src/main/java/org/apache/fluss/lake/paimon/lookup/PaimonLakeTableLookuper.java
luoyuxia: Historical lookup is triggered only after the original Fluss partition has expired and been dropped.
Since this pr does not support writes to expired partitions, so no new rows are expected and initializing the file set once is sufficient.
I also add a todo to explan it.

fluss-lake/fluss-lake-paimon/src/main/java/org/apache/fluss/lake/paimon/lookup/PaimonLakeTableLookuper.java
luoyuxia: Good point. 
For the initial implementation, I’d prefer to keep using the shared executor. Historical lookup is optional, and introducing a dedicated pool and another configuration option would add resources and configuration surface for clusters that may never use it. We expect the initial lookup traffic to be limited and can tune the existing I/O pool for deployments that need it.
I suggest revisiting a dedicated, possibly lazily created executor if real workloads show contention

fluss-rpc/src/main/proto/FlussApi.proto
luoyuxia: I agree that this can still return a silent miss while the cluster is running mixed TabletServer versions. However, before this change, lookup against an expired or otherwise non-existent partition already returned null. 
Before this pr, it will check partition exists or not, if not, it'll return null then

fluss-server/src/main/java/org/apache/fluss/server/replica/ReplicaManager.java
luoyuxia: This PR only introduces the lake lookup path. The follow-up PR that adds write support will also implement the historical RocksDB lookup and wire it in here, so the final lookup order will be buffer → historical RocksDB → lake, as described in the FIP.

fluss-client/src/main/java/org/apache/fluss/client/lookup/HistoricalPartitionResolver.java
luoyuxia: I agree that creating real replicas(no rocksdb for read now) from a read path is not ideal. However, the historical system partition is currently also the routing and ownership anchor: the client needs its partition ID to resolve a bucket leader, and the server uses the corresponding replica to validate the request and obtain table/schema metadata.

Routing historical lookups without creating the partition would require a separate virtual-routing path: selecting a TabletServer without partition leadership, resolving table and schema metadata without a replica, handling retries, and gating rolling upgrades. That is broader than removing the createPartition call.

fluss-server/src/main/java/org/apache/fluss/server/coordinator/AutoPartitionManager.java
luoyuxia: Agreed, this is a real correctness issue. Auto-partition cleanup is currently driven by retention time and does not coordinate with tiering progress. Historical lookup exposes the problem as silently stale results, but it does not introduce the underlying lifecycle gap. 

More generally, the same behavior already affects active consumers: for example, a partition may be deleted while a Flink job is still consuming it, causing the remaining data to become unavailable.

Fixing this requires additional lifecycle coordination rather than a local change here. We need to define and persist a deletion barrier, verify that every bucket has been fully tiered, handle stalled or failed tiering, and recover the pending deletion state after restarts.

This PR only changes historical lookup routing and preserves the historical system partition during retention cleanup. I’ll create a separate issue to track safe partition expiration and link it here.

fluss-client/src/main/java/org/apache/fluss/client/lookup/PrimaryKeyLookuper.java
luoyuxia: This is a valid clock-skew edge case.

In a normally time-synchronized deployment, triggering it requires the partition cleanup to complete during the short interval after the Coordinator crosses a partition boundary but before the client does. A lookup must also target the newly expired time slice during that same interval. The negative decision is not cached, so subsequent lookups start using the historical path as soon as the client crosses the boundary.

Making the decision fully authoritative would require moving it to the Coordinator or exposing Coordinator-derived expiration state through metadata/RPC, which is a broader cross-layer change.

Given the narrow and transient failure window, I’d prefer to track this in a follow-up issue and keep the current PR scoped.

luoyuxia:
@fresh-borzoni @platinumhamburg Thanks for your review. I already address your comments

fluss-lake/fluss-lake-paimon/src/main/java/org/apache/fluss/lake/paimon/lookup/PaimonLakeTableLookuper.java
```diff
@@ -0,0 +1,375 @@
...
+    private org.apache.paimon.data.BinaryRow convertPartition(
+            ResolvedPartitionSpec partitionSpec, RowType valueRowType) {
+        return toPaimonPartition(
+                partitionSpec,
+                valueRowType,
+                fileStoreTable().schema().logicalRowType(),
+                partitionKeyExtractor()::partition);
+    }
+
+    private org.apache.paimon.data.BinaryRow wrapLookupKey(byte[] key) {
```
beryllw: Should we double-check the key encoding for historical lookup on kv-format-v2 tables? For a non-default bucket key, `PrimaryKeyLookuper` sends a Fluss `CompactedKeyEncoder` key, but `PaimonLakeTableLookuper.wrapLookupKey` treats the bytes as a Paimon `BinaryRow` — so it might be worth confirming these two encodings line up (the current ITCase uses a default bucket key, so it wouldn't cover this path).

fluss-lake/fluss-lake-paimon/src/test/java/org/apache/fluss/lake/paimon/lookup/HistoricalPartitionLookupITCase.java
```diff
@@ -0,0 +1,218 @@
...
+    protected FlussClusterExtension getFlussClusterExtension() {
+        return FLUSS_CLUSTER_EXTENSION;
+    }
+
+    private static Schema partitionedPkSchema() {
+        return Schema.newBuilder()
+                .column("id", DataTypes.INT())
+                .column("dt", DataTypes.STRING())
+                .column("name", DataTypes.STRING())
+                .primaryKey("id", "dt")
```
beryllw: Could we add a test with a non-default bucket key (bucket key != primary key)? The current ITCase uses a default bucket key, so it only covers the Paimon-encoder path — a non-default bucket key would exercise the CompactedKeyEncoder path, where the historical-lookup key encoding may not match what Paimon expects (this affects point lookups too, since the encoder is table-level).

fluss-server/src/main/java/org/apache/fluss/server/utils/ServerRpcMessageUtils.java
```diff
@@ -1138,6 +1139,41 @@ public static Map<TableBucket, List<byte[]>> toLookupData(LookupRequest lookupRe
...
+                            tableId,
+                            lookupReqForBucket.hasPartitionId()
+                                    ? lookupReqForBucket.getPartitionId()
+                                    : null,
+                            lookupReqForBucket.getBucketId());
+            List<byte[]> keys = new ArrayList<>(lookupReqForBucket.getKeysCount());
+            for (int i = 0; i < lookupReqForBucket.getKeysCount(); i++) {
+                keys.add(lookupReqForBucket.getKeyAt(i));
+            }
+            lookupEntryData.put(
```
beryllw: nit: in toHistoricalLookupData, two buckets with the same TableBucket but different originalPartitionName silently overwrite each other. Might be safer to fail fast on a duplicate here — the Java client already splits these, so this would just guard against other clients.

platinumhamburg (approved):
LGTM.

fresh-borzoni (approved):
@luoyuxia Thx for the answers and fixes, LGTM

luoyuxia:
@wuchong Could you please help have a final review?

beryllw:
LGTM!

fluss-common/src/main/java/org/apache/fluss/config/ConfigOptions.java
```diff
@@ -157,6 +157,13 @@ public class ConfigOptions {
                     .withDescription(
                             "The default size of the write buffer for writing the local files to remote file systems.");
 
+    public static final ConfigOption<String> IO_TMP_DIR =
+            key("io.tmpdir")
```
wuchong: `IO_TMP_DIR` is only consumed by the server-side historical lake lookup path, while Fluss already scopes comparable options under `server.*` (for example, `server.io-pool.size`) and uses `client.scanner.io.tmpdir` for the client-side temporary directory. Exposing this new server-only setting as the generic `io.tmpdir` makes the public configuration namespace inconsistent and can cause it to be mistaken for a process-wide option. Maybe `server.io.tmpdir` is better here.

fluss-client/src/main/java/org/apache/fluss/client/lookup/LookupClient.java
```diff
@@ -100,7 +101,18 @@ public CompletableFuture<byte[]> lookup(
             TableBucket tableBucket,
             byte[] keyBytes,
             boolean insertIfNotExists) {
-        LookupQuery lookup = new LookupQuery(tablePath, tableBucket, keyBytes, insertIfNotExists);
```
wuchong, line 103: This method is unused and can be removed. We should keep the interface simple to prevent misuse, as excessive method overloading increases complexity and the risk of errors.

fluss-client/src/main/java/org/apache/fluss/client/lookup/LookupQuery.java
```diff
@@ -39,7 +41,16 @@ public class LookupQuery extends AbstractLookupQuery<byte[]> {
 
     LookupQuery(
```
wuchong: This constructor can be removed.

fluss-client/src/main/java/org/apache/fluss/client/lookup/LookupQuery.java
wuchong: Add `@VisibleForTesting` annotatiton.

fluss-server/src/main/java/org/apache/fluss/server/coordinator/CoordinatorService.java
```diff
@@ -742,17 +744,38 @@ public CompletableFuture<CreatePartitionResponse> createPartition(
...
-        // second, check whether the partition is out-of-date.
-        validateAutoPartitionTime(
-                partitionSpec,
-                tableInfo.getPartitionKeys(),
-                tableInfo.getTableConfig().getAutoPartitionStrategy());
+        ResolvedPartitionSpec partitionToCreate;
+        if (isHistoricalPartitionCreate(tableInfo, partitionSpec)) {
+            // Historical system partitions are lookup metadata, so creating one requires the same
+            // permission as reading the table instead of writing table data.
+            authorizeTable(OperationType.READ, tablePath);
```
wuchong: Routing historical creation through this public `createPartition` RPC under `READ` authorization gives a read-only principal a DDL capability: a client can create persistent historical system partitions without the normal `WRITE` authorization, potentially consuming partition and replica quota. Please move ownership of this lifecycle to the coordinator instead. Introduce a dynamically alterable table-level historical-partition option, defaulting to `false`; when enabled, the coordinator/auto-partition manager should create and retain the required `__historical__` partition(s), and when disabled it should remove them. The lookup client can then assume that the historical routing partition already exists and only refresh/resolve metadata, which closes the READ-to-DDL privilege leak and simplifies the client path.

fluss-server/src/main/java/org/apache/fluss/server/coordinator/AutoPartitionManager.java
```diff
@@ -541,13 +542,14 @@ private void dropPartitions(
                 currentPartitions.headMap(lastRetainPartitionTime).entrySet().iterator();
         while (iterator.hasNext()) {
             Map.Entry<String, Set<String>> entry = iterator.next();
-            dropPartitions(tablePath, partitionKeys, iterator, entry);
+            dropPartitions(tablePath, partitionKeys, autoPartitionStrategy, iterator, entry);
```
wuchong: If we introduce historical partition enable option on table, we can simple check whether the table enabled this option, and check the time partition value (`entry.key()`) is equal to `__historical__`, then we can skip to drop this partition. This can avoid to change the logic of `dropPartitions`.

fluss-client/src/main/java/org/apache/fluss/client/lookup/PrimaryKeyLookuper.java
```diff
@@ -115,19 +146,111 @@ public CompletableFuture<LookupResult> lookup(InternalRow lookupKey) {
...
+        return lookupBucket(tableBucket, pkBytes, insertIfNotExists, null, lookupKey);
+    }
+
+    /** Falls back to historical lookup if the lookup key belongs to a historical partition. */
+    private CompletableFuture<LookupResult> mayFallbackToHistoricalLookup(
+            int bucketId, byte[] keyBytes, String originalPartitionName, InternalRow lookupKey) {
+        // The lookuper keeps tableInfo from creation time, but retention can be changed later.
+        // Fetch the latest table metadata before classifying a missing partition as historical and
+        // verify that the table path still refers to the same table.
+        return admin.getTableInfo(tableInfo.getTablePath())
```
wuchong: We can fail fast if the table does not enable historical partitions, eliminating the need to fetch table metadata again. Instead, we can directly send the historical lookup request to the server, which will verify whether the partition has expired. This approach is more reliable as it avoids potential clock skew issues between the client and the server.

fluss-server/src/main/java/org/apache/fluss/server/replica/HistoricalLakeLookupManager.java
```diff
@@ -0,0 +1,403 @@
...
+                                            "Historical lookup is throttled for "
+                                                    + tableBucket
+                                                    + "."))));
+        }
+
+        CompletableFuture<LookupResultForBucket> future;
+        try {
+            future =
+                    CompletableFuture.supplyAsync(
+                            () -> lookupInternal(lookupData, tableInfo, schemaInfo), ioExecutor);
```
wuchong: `ioExecutor` is the shared TabletServer I/O pool (10 threads by default), while `PaimonLakeTableLookuper.lookup` synchronizes the entire table lookuper. For a hot table, one worker performs the lake I/O and the remaining accepted lookups can occupy other shared workers waiting on the same monitor; with the historical admission limit defaulting to 50, this can exhaust the pool and starve snapshot transfer, remote-log work, and other server I/O. Please isolate historical lookup on a bounded dedicated executor or reserve capacity in the shared pool. 

Longer term, an event-pool model with table-ID affinity could route all lookup events for the same table to one worker thread. This preserves per-table serialization without wasting multiple threads on synchronized waits.

fluss-common/src/main/java/org/apache/fluss/lake/lakestorage/LakeTableLookuper.java
```diff
@@ -0,0 +1,92 @@
...
+/**
+ * A table-level point lookuper for data stored in lake storage.
+ *
+ * @since 1.0
+ */
+@PublicEvolving
+public interface LakeTableLookuper extends AutoCloseable {
+
+    /**
+     * Looks up one key from the lake table.
```
wuchong, line 37: We should add description that the implementation should be thread-safe.

fluss-client/src/main/java/org/apache/fluss/client/lookup/LookupSender.java
```diff
@@ -236,7 +254,47 @@ private void sendLookupRequest(
...
+
+    /**
+     * Packs lookup batches into RPC request groups so each request has one lookup kind and response
+     * dispatch can still use {@link TableBucket} as the lookup key.
+     *
+     * <p>Normal and historical lookups are sent in separate RPCs. Historical lookups can have
+     * multiple batches with the same target {@link TableBucket} but different original partition
+     * names. The request carries that original partition name, but the response does not. If those
+     * batches were sent in one RPC, two response buckets would have the same {@link TableBucket}
+     * and the client could not map each response back to the right batch.
```
wuchong: Can we add `original_partition_name` to the `PbLookupRespForBucket`? So that we can recognize the response buckets for each original partition. This can simplify much the client logic and improve the throughput.

fluss-client/src/main/java/org/apache/fluss/client/lookup/LookupSender.java
```diff
@@ -445,6 +504,18 @@ private void reEnqueueLookup(AbstractLookupQuery<?> lookup) {
         lookupQueue.reEnqueue(lookup);
     }
 
+    private long prepareRetry(AbstractLookupQuery<?> lookup, Exception exception) {
```
wuchong, line 509: Add a TODO above this, this will introduce out-of-order for the lookup queries. We will see whether we need to handle this in the future.

fluss-server/src/main/java/org/apache/fluss/server/replica/ReplicaManager.java
```diff
@@ -775,6 +873,52 @@ public void lookups(
...
+            responseCallback.accept(Collections.emptyMap());
+            return;
+        }
+
+        Map<TableBucket, LookupResultForBucket> result = new ConcurrentHashMap<>();
+        AtomicInteger remainingLookups = new AtomicInteger(lookupData.size());
+        for (LookupDataForBucket data : lookupData) {
+            CompletableFuture<LookupResultForBucket> lookupFuture;
+            try {
+                Replica replica = getReplicaOrException(data.tableBucket());
```
wuchong, line 826: Should throw an exception if the replica is not a priamry key table.

fluss-server/src/main/java/org/apache/fluss/server/replica/ReplicaManager.java
```diff
@@ -237,6 +246,47 @@ public ReplicaManager(
...
+                serverMetricGroup,
+                userMetrics,
+                scannerManager,
+                clock,
+                ioExecutor,
+                localDiskManager,
+                null);
+    }
+
+    public ReplicaManager(
```
wuchong, line 241: Please avoid introducing too many constructors, especially public constructors. Otherwise, it's easy to use the wrong constructor.

fluss-client/src/main/java/org/apache/fluss/client/lookup/PrimaryKeyLookuper.java
```diff
@@ -53,6 +61,16 @@ class PrimaryKeyLookuper extends AbstractLookuper implements Lookuper {
...
+    private final Admin admin;
+    private final HistoricalPartitionResolver historicalPartitionResolver;
+
+    /**
+     * Partition names already confirmed as expired and eligible for historical lookup.
+     *
+     * <p>Expired partitions cannot be recreated or written in the current lifecycle, so this
+     * decision remains valid for the lifetime of the lookuper.
+     */
+    private final Set<String> confirmedHistoricalPartitions;
```
wuchong, line 73: Add a TODO for this, in the future, we can introduce a dedicate `list_historical_partitions` or `list_all_partitions`, to initialize the set at the beginning, to avoid too many exception which slows down the bootstrap perfromance.

fluss-client/src/main/java/org/apache/fluss/client/lookup/LookupSender.java
wuchong: @luoyuxia This doesn't match the FIP-28 design, could you add a TODO comment to explain this? I'm afraid we may need to introduce this configuration sooner or later.

fluss-server/src/main/java/org/apache/fluss/server/utils/ServerRpcMessageUtils.java
```diff
@@ -1138,6 +1139,50 @@ public static Map<TableBucket, List<byte[]>> toLookupData(LookupRequest lookupRe
...
+                            tb,
+                            keys,
+                            lookupReqForBucket.hasOriginalPartitionName()
+                                    ? lookupReqForBucket.getOriginalPartitionName()
+                                    : null);
+            LookupDataForBucket previous = lookupEntryData.putIfAbsent(tb, lookupData);
+            if (previous != null) {
+                throw new IllegalArgumentException(
+                        String.format(
+                                "Duplicate historical lookup requests for table bucket %s with "
```
wuchong: We should allow different original partition for the same table bucket on a single RPC, this is very important for the performance on production.

fluss-server/src/main/java/org/apache/fluss/server/tablet/TabletService.java
```diff
@@ -292,30 +296,59 @@ public CompletableFuture<PutKvResponse> putKv(PutKvRequest request) {
...
                     currentSession().getApiVersion(),
                     value -> response.complete(makeLookupResponse(value, errorResponseMap)));
         } else {
-            Map<TableBucket, List<byte[]>> interesting =
-                    authorizeRequestData(
-                            READ, lookupData, errorResponseMap, LookupResultForBucket::new);
-            if (interesting.isEmpty()) {
-                return CompletableFuture.completedFuture(makeLookupResponse(errorResponseMap));
+            boolean historicalLookupRequest = hasHistoricalLookup(request);
+            if (historicalLookupRequest) {
```
wuchong, line 326: IIUC, we don't allow hybrid normal lookup and historical lookup in a single RPC, currently, it is guaranteed by client. But server should add this hard check, otherwise, normal lookup will also be routed to historical lookup and returns wrong result.

fluss-server/src/main/java/org/apache/fluss/server/replica/ReplicaManager.java
```diff
@@ -979,6 +1129,9 @@ public void stopReplicas(
 
                     for (StopReplicaData data : stopReplicaDataList) {
                         TableBucket tb = data.getTableBucket();
+                        long tableId = tb.getTableId();
+                        ioExecutor.execute(
+                                () -> historicalLakeLookupManager.invalidateTableLookuper(tableId));
```
wuchong: `StopReplica` does not imply that the table was dropped: it is also sent for replica-offline transitions, reassignments, and expired-partition deletion. This code invalidates the table-scoped lookuper once per bucket, so a normal rebalance or a multi-bucket partition expiration can repeatedly discard Paimon's catalog, `LocalTableQuery`, `IOManager`, and local lookup cache, forcing later historical lookups through expensive cold initialization. A newly expired partition does not need table-wide invalidation because the lookuper initializes each `(partition, bucket)` lazily, and schema changes are already handled by the `schemaId` check.

Please remove this invalidation from `stopReplicas` and trigger it only when `UpdateMetadata` applies an actual table-deletion tombstone, deduplicated by table ID, for example:

```java
Set<Long> deletedTableIds = metadataCache.updateClusterMetadata(clusterMetadata);
updateReplicaTableConfig(clusterMetadata);

if (!deletedTableIds.isEmpty()) {
    ioExecutor.execute(
            () ->
                    deletedTableIds.forEach(
                            historicalLakeLookupManager::invalidateTableLookuper));
}
```

Filtering on `deleteRemote` is not sufficient because expired-partition deletion also sets it.

fluss-lake/fluss-lake-paimon/src/main/java/org/apache/fluss/lake/paimon/lookup/PaimonLakeTableLookuper.java
```diff
@@ -0,0 +1,421 @@
...
+        }
+
+        org.apache.paimon.data.BinaryRow keyRow =
+                new org.apache.paimon.data.BinaryRow(primaryKeyFieldCount);
+        keyRow.pointTo(MemorySegment.wrap(paimonKey), 0, paimonKey.length);
+        return keyRow;
+    }
+
+    private void initializeFilesIfNeeded(org.apache.paimon.data.BinaryRow partition, int bucketId) {
+        PaimonPartitionBucket partitionBucket = new PaimonPartitionBucket(partition, bucketId);
```
wuchong, line 302: `RowPartitionKeyExtractor` uses a generated projection whose default is `reusedOutRecord=true`, so `convertPartition` returns the same mutable `BinaryRow` instance on every lookup. This line retains that row in `initializedBuckets`, and `refreshFiles` also retains it as a key in `LocalTableQuery.tableView`; converting the next partition mutates keys that are already inside hash-based collections. In particular, two partitions with the same `BinaryRow.hashCode()` make the second partition appear initialized and reuse the first partition file view, which can return a row from the wrong lake partition or produce a false miss. Please copy the projected row before it escapes (for example, `partitionKeyExtractor().partition(...).copy()`) and add a two-partition hash-collision regression test.

wuchong (commented):
Thanks @luoyuxia , besides the inline comment. I want to discuss the historical partition layout design. Note: we can support single partition key in this PR, and introduce multiple-partition-keys in follow-ups, so this discussion can be decided later. 

### `__historical__` partition per table vs. `__historical__` partition per static prefix partition

Since historical reads and writes do not rely on Fluss partition pruning, I do not see a strong benefit in preserving the static partition prefix and creating partitions such as `us$__historical__` and `eu$__historical__`.

The original partition identity is already preserved separately: lookup requests carry the original partition name, while historical writes can retain it in the row or request metadata. Therefore, the historical Fluss partition is primarily a routing and buffering mechanism rather than a representation of the original lake partition.

I would prefer one fixed `__historical__` partition per table, regardless of whether the table has single-level or multi-level partitioning. This makes routing deterministic, keeps metadata and replica usage bounded, and avoids creating and managing one permanent historical partition for every static prefix. It also significantly simplifies historical partition discovery and lifecycle management.

The main trade-off is that historical traffic shares one set of bucket assignments, but the table’s existing bucket distribution and historical-path flow control should still provide parallelism. But this can be addressed by supporting a different bucket number for historical partition in the future. Another problem is the current `ResolvedPartitionSpec` requires the same number of partition-keys/values, but the single `__historical__` may need a special logic for this.

luoyuxia:
@wuchong Comments has been addressed. PTAL

fluss-server/src/main/java/org/apache/fluss/server/replica/HistoricalLakeLookupManager.java
```diff
@@ -0,0 +1,472 @@
...
+            LookupDataForBucket lookupData, TableInfo tableInfo, SchemaInfo schemaInfo) {
+        CompletableFuture<LookupResultForBucket> future =
+                CompletableFuture.supplyAsync(
+                        () -> lookupInternal(lookupData, tableInfo, schemaInfo),
+                        getOrCreateHistoricalPartitionExecutor());
+        pendingLookups.add(future);
+        return future;
+    }
+
+    private synchronized ExecutorService getOrCreateHistoricalPartitionExecutor() {
```
wuchong: We don't need to lazily initialize `ThreadPoolExecutor`, because the `ThreadPoolExecutor` itself doens't have much overhead, all the threads are started lazily. This can simplify and remove the lock on each submit path.

fluss-server/src/main/java/org/apache/fluss/server/replica/HistoricalLakeLookupManager.java
```diff
@@ -0,0 +1,472 @@
...
+        if (historicalPartitionExecutor == null) {
+            int maxThreadPoolSize =
+                    conf.get(ConfigOptions.SERVER_HISTORICAL_PARTITION_THREAD_POOL_MAX_SIZE);
+            ThreadPoolExecutor executor =
+                    new ThreadPoolExecutor(
+                            maxThreadPoolSize,
+                            maxThreadPoolSize,
+                            HISTORICAL_PARTITION_THREAD_KEEP_ALIVE.toMillis(),
+                            TimeUnit.MILLISECONDS,
+                            new ArrayBlockingQueue<>(maxQueuedHistoricalRequests),
```
wuchong: Since we have a outside `lookupPermits` to constraints the number of queued requests, I think we don't need to limit the queue size of the thread pool?

fluss-server/src/main/java/org/apache/fluss/server/replica/HistoricalLakeLookupManager.java
```diff
@@ -0,0 +1,472 @@
...
+ * access. Caffeine expiration is scheduled on the shared TabletServer scheduler, allowing idle
+ * resources to be released even if no subsequent lookup accesses the cache.
+ */
+class HistoricalLakeLookupManager implements AutoCloseable {
+
+    private static final String PAIMON_LOOKUP_DIR_NAME = "paimon-lookup";
+    private static final String LOOKUPER_CACHE_EXPIRATION_TASK_NAME =
+            "historical-lookuper-cache-expiration";
+    private static final Duration LOOKUPER_CACHE_EXPIRATION = Duration.ofHours(3);
+    private static final Duration HISTORICAL_PARTITION_THREAD_KEEP_ALIVE = Duration.ofSeconds(60);
```
wuchong: Can be a little longer, such as 10min, to allow the flink job restart period.

fluss-server/src/main/java/org/apache/fluss/server/utils/TableDescriptorValidation.java
```diff
@@ -169,6 +170,26 @@ private static void checkTableLakeFormatMatchesCluster(
...
+                || tableDescriptor.getPartitionKeys().size() != 1
+                || !tableConf.get(ConfigOptions.TABLE_AUTO_PARTITION_ENABLED)
+                || !tableConf.get(ConfigOptions.TABLE_DATALAKE_ENABLED)
+                || tableConf.getOptional(ConfigOptions.TABLE_DATALAKE_FORMAT).orElse(null)
+                        != DataLakeFormat.PAIMON) {
+            throw new InvalidConfigException(
+                    String.format(
+                            "'%s' can only be enabled for auto-partitioned Paimon primary key "
+                                    + "tables with a single partition key.",
+                            ConfigOptions.TABLE_DATALAKE_HISTORICAL_PARTITION_ENABLED.key()));
```
wuchong: Could we make this validation error more actionable by listing every unmet prerequisite? The current generic message does not tell users whether `table.auto-partition.enabled` or `table.datalake.enabled` is disabled, `table.datalake.format` is missing or not `paimon`, or the table lacks a primary key or exactly one partition key. Reporting all unmet requirements would let users correct the table definition in one pass.

fluss-client/src/main/java/org/apache/fluss/client/lookup/PrimaryKeyLookuper.java
```diff
@@ -115,19 +139,107 @@ public CompletableFuture<LookupResult> lookup(InternalRow lookupKey) {
...
+                                // Propagate its failure instead of falling back again.
+                                if (!(error instanceof PartitionNotExistException)
+                                        || originalPartitionName != null) {
+                                    lookupFuture.completeExceptionally(error);
+                                    return;
+                                }
+
+                                // The cached normal partition was deleted. Re-evaluate the routing
+                                // using the lookup key.
+                                String partitionName = partitionGetter.getPartition(lookupKey);
```
wuchong: Capture the original partition name before entering the asynchronous callback instead of reading `lookupKey` here. `FlinkAsyncLookupFunction` reuses a single `FlinkAsFlussRow` and mutates it with `replace()` for each invocation, so while request A is waiting for its RPC, request B can replace the backing row. If A then gets `PartitionNotExistException`, this callback combines A’s `keyBytes` with B’s partition name and may query the wrong historical partition or return a false miss. Please pass the immutable partition name computed during initial routing (we should use an explicit boolean flag to indicate historical lookup rather than relying on the nullable `original_partition_name` ) through the callback, and cover this interleaving with a deterministic test.

fluss-lake/fluss-lake-paimon/src/main/java/org/apache/fluss/lake/paimon/lookup/PaimonLakeTableLookuper.java
```diff
@@ -0,0 +1,418 @@
...
+            throw new IllegalStateException("Paimon lake table lookuper has been closed.");
+        }
+    }
+
+    private void ensureInitialized(RowType valueRowType) throws Exception {
+        if (localTableQuery != null) {
+            return;
+        }
+
+        catalog =
```
wuchong: Initialize these resources transactionally. If any step after assigning `catalog` fails, such as loading the table, validating the primary key, creating the IO manager/query, or initializing the key converter, this object retains partially initialized fields. A later lookup can overwrite the previous `Catalog` or `IOManager` without closing it, causing unbounded resource leaks under repeated failures; if `localTableQuery` was already assigned, the next lookup skips initialization with the remaining fields incomplete. Please build the resources in local variables and publish them only after full success, or close and reset every field on failure, and add a failure-injection test.

fluss-server/src/main/java/org/apache/fluss/server/coordinator/AutoPartitionManager.java
```diff
@@ -230,6 +244,62 @@ public void handleAutoPartitionStrategyChange(
...
+        inLock(
+                lock,
+                () -> {
+                    long tableId = tableInfo.getTableId();
+                    TreeMap<String, Set<String>> currentPartitions =
+                            checkNotNull(
+                                    partitionsByTable.get(tableId),
+                                    "Auto partition state does not exist for table " + tableId);
+                    if (!currentPartitions.containsKey(HISTORICAL_PARTITION_VALUE)) {
+                        createPartition(
```
wuchong, line 259: Please make this system-partition lifecycle self-healing instead of a one-shot best-effort transition. `createPartition` catches capacity and other failures without propagating them, so once the table option is persisted as enabled, a transient failure can leave `__historical__` absent with no online retry and historical lookups keep failing until a coordinator restart or another option change. The disable path has the symmetric problem: `MetadataManager.dropPartition` logs and swallows ZooKeeper deletion failures, after which this manager removes the in-memory entry and reports success. 

Could we move historical system partition creation into the synchronous `CoordinatorService` create/alter flow? Validate the additional KV leader capacity first, then create the table and `__historical__` before completing the RPC. Any capacity or metadata failure should be propagated to the caller instead of being swallowed. If the table was already persisted when partition creation failed, make that partial state explicit and tell the user to either drop and recreate the table, or disable and re-enable the historical-partition option after fixing the underlying issue. Startup reconciliation can remain as crash-recovery fallback, but a handled failure must never be reported as a successful create/enable operation.

fluss-server/src/main/java/org/apache/fluss/server/replica/HistoricalLakeLookupManager.java
```diff
@@ -0,0 +1,472 @@
...
+        pendingLookups.add(future);
+        return future;
+    }
+
+    private synchronized ExecutorService getOrCreateHistoricalPartitionExecutor() {
+        if (closed) {
+            throw new IllegalStateException("Historical lake lookup manager has been closed.");
+        }
+        if (historicalPartitionExecutor == null) {
+            int maxThreadPoolSize =
```
wuchong: Validate `maxThreadPoolSize` as greater than zero before constructing the executor. The new integer option accepts zero or negative values, so the server can start successfully and the first historical lookup then throws `IllegalArgumentException` from `ThreadPoolExecutor`, leaving the feature unavailable with a delayed and non-actionable failure. Please fail fast during manager or configuration initialization, similar to `maxQueuedHistoricalRequests`, and add boundary tests.

wuchong (commented):
Thanks @luoyuxia , I left some minor comments.

luoyuxia:
@jark
Thanks for reviewing again. Comments has been addressed.

wuchong (approved):
LGTM.

PR #3635, opened 2026-07-11
[lake] Support partition mark-done for lakehouse tiering (processing-time)

### LINKED ISSUE #3314: [Feature] Support Partition Mark Done for Fluss Tiering (Paimon Lake)

beryllw:
…time)

### Purpose

Linked issue: close #3314

### DIFF
--- a/fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/tiering/committer/TieringCommitOperator.java
+++ b/fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/tiering/committer/TieringCommitOperator.java
@@ -21,6 +21,7 @@ import org.apache.fluss.client.Connection;
 import org.apache.fluss.client.ConnectionFactory;
 import org.apache.fluss.client.admin.Admin;
 import org.apache.fluss.client.metadata.LakeSnapshot;
+import org.apache.fluss.config.ConfigOptions;
 import org.apache.fluss.config.Configuration;
 import org.apache.fluss.exception.LakeTableSnapshotNotExistException;
 import org.apache.fluss.flink.tiering.event.FailedTieringEvent;
@@ -30,9 +31,11 @@ import org.apache.fluss.flink.tiering.source.TieringSource;
 import org.apache.fluss.lake.committer.CommittedLakeSnapshot;
 import org.apache.fluss.lake.committer.LakeCommitResult;
 import org.apache.fluss.lake.committer.LakeCommitter;
+import org.apache.fluss.lake.committer.PartitionMarkDoneMaintainer;
 import org.apache.fluss.lake.committer.TieringStats;
 import org.apache.fluss.lake.writer.LakeTieringFactory;
 import org.apache.fluss.lake.writer.LakeWriter;
+import org.apache.fluss.lake.writer.PartitionMarkDoneEnabler;
 import org.apache.fluss.metadata.TableBucket;
 import org.apache.fluss.metadata.TableInfo;
 import org.apache.fluss.metadata.TablePath;
@@ -219,6 +222,8 @@ public class TieringCommitOperator<WriteResult, Committable>
                     "Commit tiering write results is empty for table {}, table path {}",
                     tableId,
                     tablePath);
+            // still run partition mark-done maintenance for the empty round
+            maybeCommitMarkDoneMaintenance(tableId, tablePath);
             return new CommitResult(null, null);
         }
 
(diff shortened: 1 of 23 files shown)


leonardBang:
Could we validate the complete mark-done action configuration before returning `true` in `PaimonPartitionMarkDone#isPartitionMarkDoneEnabled`? For example, `partition.mark-done-action=custom` without `partition.mark-done-action.custom.class` passes this check. A regression assertion expecting this method to return `false` currently fails because it returns `true`; `PaimonLakeCommitter` later catches the action-construction failure and sets `partitionMarkDone` to `null`. The enumerator can therefore keep emitting synthetic empty rounds for a fully caught-up table while every committer performs no maintenance, and the next poll repeats the same sequence. Could this check reuse the action validation, or could table validation fail clearly instead? A regression test covering the repeated empty-round scheduling would also help.

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/tiering/committer/TieringCommitOperator.java
```diff
@@ -290,6 +295,66 @@ private CommitResult commitWriteResults(
...
+                                            flussTableLakeSnapshotCommitter.prepareLakeSnapshot(
+                                                    tableId, tablePath, Collections.emptyMap()));
+            if (maintenanceSnapshot != null) {
+                flussTableLakeSnapshotCommitter.commit(
+                        tableId,
+                        maintenanceSnapshot.getLakeSnapshotId(),
+                        maintenanceSnapshot
+                                .getSnapshotProperties()
+                                .get(FLUSS_LAKE_SNAP_BUCKET_OFFSET_PROPERTY),
+                        null,
```
leonardBang, line 349: Could we preserve readability information when committing the maintenance snapshot here? This call always passes `null` for `readableLakeBucketTieredOffsetsPath` and `KEEP_ALL_PREVIOUS`. For a non-DV Paimon table, the maintenance snapshot only changes properties and remains readable with the merged offsets, but Fluss will continue reporting the previous readable snapshot. With automatic Paimon snapshot expiration and staggered mark-done rounds, that previous snapshot can be expired while Fluss still points to it, causing union reads to request a missing snapshot. Could the maintainer return a `LakeCommitResult` (or equivalent readability metadata), so non-DV maintenance snapshots are registered as readable while DV tables retain their existing compaction semantics? A regression test with `snapshot.num-retained.max=2`, two maintenance commits, and a real union read would help.

fluss-lake/fluss-lake-paimon/src/main/java/org/apache/fluss/lake/paimon/tiering/PaimonPartitionMarkDone.java
```diff
@@ -0,0 +1,512 @@
...
+        if (!"process-time".equalsIgnoreCase(markDoneMode)) {
+            LOG.warn(
+                    "Option {} is set to {} for table {} but only the process-time mode is "
+                            + "supported, partition mark-done is disabled.",
+                    PARTITION_MARK_DONE_MODE.key(),
+                    markDoneMode,
+                    tableInfo.getTablePath());
+            return false;
+        }
+        return true;
```
leonardBang, line 236: Could we validate the complete mark-done action configuration before returning `true` here? For example, `partition.mark-done-action=custom` without `partition.mark-done-action.custom.class` passes this check. A regression assertion expecting this method to return `false` currently fails because it returns `true`; `PaimonLakeCommitter` later catches the action-construction failure and sets `partitionMarkDone` to `null`. The enumerator can therefore keep emitting synthetic empty rounds for a fully caught-up table while every committer performs no maintenance, and the next poll repeats the same sequence. Could this check reuse the action validation, or could table validation fail clearly instead? A regression test covering the repeated empty-round scheduling would also help.

fluss-lake/fluss-lake-paimon/src/main/java/org/apache/fluss/lake/paimon/tiering/MarkDoneStateJsonSerde.java
```diff
@@ -0,0 +1,96 @@
...
+    @Override
+    public void serialize(MarkDoneState state, JsonGenerator generator) throws IOException {
+        generator.writeStartObject();
+        generator.writeBooleanField(INITIALIZED_FIELD, state.isInitialized());
+        generator.writeObjectFieldStart(PENDING_FIELD);
+        for (Map.Entry<String, Long> entry : state.getPendingPartitions().entrySet()) {
+            generator.writeNumberField(entry.getKey(), entry.getValue());
+        }
+        generator.writeEndObject();
+        generator.writeEndObject();
```
leonardBang, line 57: Would it make sense to preserve unknown fields during serialization, or add explicit version gating? Ignoring unknown fields makes deserialization tolerant, but once the pending state changes this serializer rewrites only `initialized` and `pending`. I tested a state containing `"futureState":{"epoch":7}` and then added a pending partition; the resulting JSON was `{"initialized":true,"pending":{"p1":100,"p2":200}}`, so the future field was removed. During a rolling downgrade, an older tiering service could therefore erase state introduced by a newer version. Could we retain the original unknown `JsonNode` fields during read-modify-write, or reject unsupported newer state versions? It would be useful to test preservation across an actual state transition rather than only successful deserialization.

fluss-lake/fluss-lake-paimon/src/main/java/org/apache/fluss/lake/paimon/tiering/PaimonLakeCommitter.java
```diff
@@ -111,59 +139,184 @@ public LakeCommitResult commit(
...
 
         try {
-            tableCommit = fileStoreTable.newCommit(commitUser);
-            // don't skip empty commits: tiering relies on empty snapshots to persist bucket
-            // offsets when only empty WAL batches were consumed
-            tableCommit.ignoreEmptyCommit(false);
-            tableCommit.commit(manifestCommittable);
+            if (partitionMarkDone != null) {
+                runPartitionMarkDone(manifestCommittable);
+            }
```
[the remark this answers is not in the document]
beryllw: Mark-done actions, including custom actions, are required to be idempotent and may be retried after failures, so we will retain the current retry semantics without introducing a more complex commit protocol.

PR #3636, opened 2026-07-12
[elixir] Support complex types (Array/Map/Row) and nullability

nicolazar:
### Purpose

This PR adds nested types and per-node nullability to the Elixir bindings, in both schema definitions and row values.

Linked issue: close #xxx

### Brief change log

- DataType gains recursive Array/Map/Row variants with recursive converters.
- Nullability uses an additive `{:not_null, dt}` wrapper (types nullable by default, composable at any depth). The read path now preserves the flag, so primary-key columns read back as `{:not_null, ...}`.
- Row value shapes: array -> list; map -> %{k => v}; row -> map keyed by field-name atom.

### Tests

New unit tests (schema builder) and integration round-trips (schema and values); updated existing primary-key read assertions.

### DIFF
--- a/fluss-rust/bindings/elixir/test/integration/log_table_test.exs
+++ b/fluss-rust/bindings/elixir/test/integration/log_table_test.exs
@@ -210,6 +210,98 @@ defmodule Fluss.Integration.LogTableTest do
 
       cleanup_table(admin, table_name)
     end
+
+    test "round-trips a deeply-nested array<map<string, row>>", %{conn: conn, admin: admin} do
+      table_name = "ex_test_deep_nested_#{:rand.uniform(100_000)}"
+      cleanup_table(admin, table_name)
+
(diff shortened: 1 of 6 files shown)


fluss-rust/bindings/elixir/test/integration/log_table_test.exs
```diff
@@ -166,6 +166,50 @@ defmodule Fluss.Integration.LogTableTest do
 
       cleanup_table(admin, table_name)
     end
+
+    test "round-trips array, map, and row values", %{conn: conn, admin: admin} do
```
fresh-borzoni, line 170: Can we add a deeply-nested value (array<map<string,row>> or smth) and a nulls-at-depth case (null complex column + null array element/map value/row field)?

fresh-borzoni (approved):
@nicolazar Very nice!
A little bit more tests and good to merge.

fresh-borzoni (approved):

PR #3639, opened 2026-07-13
[tiering][iceberg] Validate partition spec before lake tiering

### LINKED ISSUE #3638: [tiering][iceberg] Validate partition spec before lake tiering
### Search before asking

### Fluss version

main (development)

### Problem

The Iceberg lake tiering writer does not validate that the current Iceberg partition spec is still compatible with the partitioning definition used by Fluss.

An external Iceberg client can evolve the partition spec after lake tiering has been enabled. A newly created or restarted tiering writer may then load the evolved Iceberg spec but continue using partition values calculated according to the Fluss table's original bucket definition.

This problem already applies to the existing explicit bucket partitioning and does not depend on implicit/hidden partition support.

### Reproduction example

1. Create a Fluss primary-key table distributed by `id` into 8 buckets and enable Iceberg lake tiering.
2. Wait for the Iceberg table to be created and verify that its current partition spec contains `bucket[8](id)`.
3. Stop or pause the tiering job so that its lake writer will be recreated later.
4. Using an external Iceberg client, evolve the Iceberg table's current partition spec from `bucket[8](id)` to `bucket[16](id)`. The Fluss table remains configured with 8 buckets.
5. Restart the tiering job, causing a new Iceberg lake writer to load the current `bucket[16](id)` spec.
6. Write a record whose `bucket[8](id)` value differs from its `bucket[16](id)` value, and wait for it to be tiered.
7. Inspect the committed data file's partition metadata or query the table with a predicate on that `id`.

### Current behavior

The writer creates a `PartitionKey` using the current Iceberg spec, but directly sets the Fluss table bucket ID into its bucket field. It does not apply the current Iceberg bucket transform to the original `id` value.

For example, if `bucket[8](id) = 3` while `bucket[16](id) = 11`, the writer records `3` as the value of the `bucket[16](id)` partition field. Because both values are integers, this mismatch may not cause a type or commit error.

The file can therefore be committed with incorrect partition metadata. Iceberg readers may project the predicate using `bucket[16](id)` and prune that file, causing the record to be missing from filtered queries even though the data file was committed successfully.

### Expected behavior

When creating the Iceberg lake writer, Fluss should verify that the current Iceberg partition spec is semantically equivalent to the partitioning definition expected from the Fluss table. If it is incompatible, writer initialization should fail with a clear error before any incorrectly partitioned file can be committed.

### Solution

Add an Iceberg partition-spec compatibility validator to the lake writer.

The validation should compare the expected Fluss-to-Iceberg mapping with the current Iceberg spec, including:

- source columns;
- transform types;
- transform parameters such as the bucket count;
- partition field order where it affects the produced partition row.

Run the validation when the Iceberg writer is initialized. If table metadata can be refreshed while the writer remains active, cache the validated Iceberg spec ID and validate again when the current spec ID changes.

Reject incompatible specs with a non-retriable, descriptive exception that includes both the expected and current partition definitions.

litiliu:
### Purpose

Linked issue: close #3638

Iceberg lake tiering constructs partition keys from the current Iceberg partition spec, but fills them with partition and bucket values produced according to the Fluss table metadata. If an external client evolves the Iceberg partition spec, a newly created writer can silently attach incorrect partition metadata to data files.

This change validates the current Iceberg partition spec before the tiering writer is created and fails fast when the source columns, field order, transforms, or transform parameters are incompatible with the Fluss table.

### Brief change log

- Extract the Fluss-to-Iceberg partition-spec construction into a shared internal utility used by both automatic Iceberg table creation and writer validation.
- Add a partition-spec validator that compares the expected and current Iceberg specs by field order, source field ID, and transform, including parameters such as bucket count.
- Run the validation before creating the Iceberg task writer and throw a descriptive, non-retriable `InvalidTableException` on mismatch.
- Add regression coverage for an externally evolved bucket transform and align the existing tiering test metadata with the tested table definitions.

### Tests

- `mvn -pl fluss-lake/fluss-lake-iceberg verify`
  - 108 unit tests passed.
  - 28 integration tests passed.

### API and Format

No public API or storage-format changes. The shared partition-spec utility is marked `@Internal`, and the existing Iceberg table mapping remains unchanged.

### Documentation

No documentation changes. This is a correctness fix with no new user-facing configuration.

### Generative AI disclosure

### DIFF
--- a/fluss-lake/fluss-lake-iceberg/src/main/java/org/apache/fluss/lake/iceberg/tiering/IcebergLakeWriter.java
+++ b/fluss-lake/fluss-lake-iceberg/src/main/java/org/apache/fluss/lake/iceberg/tiering/IcebergLakeWriter.java
@@ -68,6 +68,7 @@ public class IcebergLakeWriter implements LakeWriter<IcebergWriteResult> {
             throws IOException {
         this.icebergCatalog = icebergCatalogProvider.get();
         this.icebergTable = getTable(writerInitContext.tablePath());
+        IcebergPartitionSpecValidator.validate(icebergTable, writerInitContext.tableInfo());
 
         // Create a record writer
         this.recordWriter = createRecordWriter(writerInitContext);
--- a/fluss-lake/fluss-lake-iceberg/src/main/java/org/apache/fluss/lake/iceberg/tiering/IcebergPartitionSpecValidator.java
+++ b/fluss-lake/fluss-lake-iceberg/src/main/java/org/apache/fluss/lake/iceberg/tiering/IcebergPartitionSpecValidator.java
@@ -1,0 +1,81 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.fluss.lake.iceberg.tiering;
+
+import org.apache.fluss.exception.InvalidTableException;
+import org.apache.fluss.lake.iceberg.IcebergSchemaUtils;
+import org.apache.fluss.lake.iceberg.utils.IcebergPartitionSpecUtils;
+import org.apache.fluss.metadata.TableDescriptor;
+import org.apache.fluss.metadata.TableInfo;
+
+import org.apache.iceberg.PartitionField;
+import org.apache.iceberg.PartitionSpec;
+import org.apache.iceberg.Schema;
+import org.apache.iceberg.Table;
+
+import java.util.List;
+
+/** Validates that an Iceberg partition spec is compatible with Fluss lake tiering. */
+final class IcebergPartitionSpecValidator {
+
+    private IcebergPartitionSpecValidator() {}
+
+    static void validate(Table icebergTable, TableInfo tableInfo) {
+        TableDescriptor tableDescriptor = tableInfo.toTableDescriptor();
+        Schema icebergSchema = icebergTable.schema();
+        Schema expectedSchema =
+                IcebergSchemaUtils.createIcebergSchema(tableDescriptor, tableInfo.hasPrimaryKey());
+        if (!IcebergSchemaUtils.compatibleWith(icebergSchema, expectedSchema)) {
+            throw new InvalidTableException(
+                    String.format(
+                            "Iceberg schema is incompatible with Fluss table %s. "
+                                    + "Expected Iceberg schema: %s, but the current Iceberg schema is: %s.",
+                            tableInfo.getTablePath(), expectedSchema, icebergSchema));
+        }
+
+        PartitionSpec icebergSpec = icebergTable.spec();
+        PartitionSpec expectedSpec =
+                IcebergPartitionSpecUtils.createPartitionSpec(tableDescriptor, icebergSchema);
+
+        if (!compatibleWith(icebergSpec, expectedSpec)) {
+            throw new InvalidTableException(
+                    String.format(
+                            "Iceberg partition spec is incompatible with Fluss table %s. "
+                                    + "Expected Iceberg partition spec: %s, but the current Iceberg partition spec is: %s.",
+                            tableInfo.getTablePath(), expectedSpec, icebergSpec));
+        }
+    }
+
+    private static boolean compatibleWith(PartitionSpec currentSpec, PartitionSpec expectedSpec) {
+        List<PartitionField> currentFields = currentSpec.fields();
+        List<PartitionField> expectedFields = expectedSpec.fields();
+        if (currentFields.size() != expectedFields.size()) {
+            return false;
+        }
+
+        for (int i = 0; i < currentFields.size(); i++) {
+            PartitionField currentField = currentFields.get(i);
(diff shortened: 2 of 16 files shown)


litiliu:
@fresh-borzoni  Would you mind taking a look at this PR when you have time? Thanks.

fluss-lake/fluss-lake-iceberg/src/main/java/org/apache/fluss/lake/iceberg/utils/IcebergPartitionSpecUtils.java
```diff
@@ -0,0 +1,104 @@
...
+        }
+
+        if (bucketKeys.isEmpty() && isPrimaryKeyTable) {
+            throw new IllegalArgumentException(
+                    "Bucket key must be set for primary key Iceberg tables");
+        }
+
+        PartitionSpec.Builder builder = PartitionSpec.builderFor(icebergSchema);
+        for (String partitionKey : partitionKeys) {
+            if (!icebergSchema.findType(partitionKey).typeId().equals(STRING)) {
```
fresh-borzoni, line 75: What happens if the column was renamed in iceberg instead of the spec changing?

fluss-lake/fluss-lake-iceberg/src/main/java/org/apache/fluss/lake/iceberg/tiering/IcebergPartitionSpecValidator.java
```diff
@@ -0,0 +1,45 @@
...
+final class IcebergPartitionSpecValidator {
+
+    private IcebergPartitionSpecValidator() {}
+
+    static void validate(Table icebergTable, TableInfo tableInfo) {
+        PartitionSpec icebergSpec = icebergTable.spec();
+        PartitionSpec expectedSpec =
+                IcebergPartitionSpecUtils.createPartitionSpec(tableInfo, icebergTable.schema());
+
+        if (!icebergSpec.compatibleWith(expectedSpec)) {
```
fresh-borzoni: nit: compatibleWith also compares partition field names. Two concerns:
-  simple updateSpec().renameField() blocks tiering, even though the write path never looks at field names
- names are persisted in existing tables, so if createPartitionSpec ever changes its naming, every existing table fails validation after an upgrade
Should we compare position + sourceId + transform instead?

fluss-lake/fluss-lake-iceberg/src/main/java/org/apache/fluss/lake/iceberg/tiering/IcebergLakeWriter.java
```diff
@@ -68,6 +68,7 @@ public IcebergLakeWriter(
             throws IOException {
         this.icebergCatalog = icebergCatalogProvider.get();
         this.icebergTable = getTable(writerInitContext.tablePath());
+        IcebergPartitionSpecValidator.validate(icebergTable, writerInitContext.tableInfo());
```
fresh-borzoni, line 71: Should an incompatible table fail the whole tiering job (that's what happens when we throw here), or just that table?

fresh-borzoni (commented):
@litiliu Thank you for the PR, left some questions, PTAL

fluss-lake/fluss-lake-iceberg/src/main/java/org/apache/fluss/lake/iceberg/tiering/IcebergPartitionSpecValidator.java
litiliu: Agreed. I changed the partition-spec comparison to compare the fields in order by `sourceId` and transform only, intentionally ignoring partition field names. Source column names, types, nullability, and order are still validated separately by the schema compatibility check. I also added `testAllowDifferentPartitionFieldNames` to verify that renaming only the partition fields does not block tiering.

fluss-lake/fluss-lake-iceberg/src/main/java/org/apache/fluss/lake/iceberg/utils/IcebergPartitionSpecUtils.java
litiliu: If a source column is renamed in Iceberg, its schema no longer matches the Fluss schema because column names, types, nullability, and order are compared. The table will therefore fail validation with an `InvalidTableException`. I also added `testRejectRenamedIcebergColumn` to cover this case.

fluss-lake/fluss-lake-iceberg/src/main/java/org/apache/fluss/lake/iceberg/tiering/IcebergLakeWriter.java
litiliu: It should fail only the incompatible table, not the whole tiering job. I moved the validation out of `IcebergLakeWriter` and into `TieringSourceEnumerator` before split generation. A validation exception is caught by the existing per-table failure path, no splits are assigned, and the failed table epoch is recorded for the next heartbeat so the job can continue processing other tables. I added `testTableValidationFailureDoesNotAssignSplits` to cover the validation-failure path.

fresh-borzoni: @litiliu 
Realised my own comment created a hole here: now nothing checks the spec between the enumerator validation and the actual write. If someone evolves the spec in that window, we're silently writing wrong partition metadata again, the exact thing this PR fixes.

Shall we just keep the validate here too? Table's already loaded, it's one line. And it can't crash-loop like before as old drift gets quarantined by the enumerator now, so this would only ever fire on a mid-round change. 

Worst case one restart, next round the table is quarantined anyway. This seems like a correct tradeoff.

fresh-borzoni (commented):
@litiliu Thank you, looks good, let's close one hole left and good to merge

fluss-lake/fluss-lake-iceberg/src/main/java/org/apache/fluss/lake/iceberg/tiering/IcebergLakeWriter.java
litiliu: @fresh-borzoni  Good idea, it's very considerate to verify the schema both on the enumerator and writer side. I have fixed it.

fresh-borzoni (approved):
@litiliu Thank you, LGTM 👍

PR #3641, opened 2026-07-13
[flink] Fix TableNotExistException when listing partitions on  changelog / binlog virtual tables

### LINKED ISSUE #3640: [flink] TableNotExistException when listing partitions on $changelog/$binlog virtual tables
### Search before asking

### Fluss version

main (development)

### Please describe the bug 🐞

When querying a `$changelog` or `$binlog` virtual table on a partitioned primary-key table, the Flink optimizer calls `FlinkCatalog.listPartitions()` on the virtual table name (e.g., `realtime$changelog`). This throws `TableNotExistException` because the virtual table suffix is not stripped before calling `admin.listPartitionInfos()`.

The same issue exists in `tableExists()`, `createPartition()`, and `dropPartition()` — all of them pass the full virtual table name (including `$changelog`/`$binlog` suffix) directly to the admin API, which queries the Fluss server for a physical table that doesn't exist.

`getTable()` correctly handles this by dispatching to `getVirtualChangelogTable()`/`getVirtualBinlogTable()`, but the partition/existence methods do not have this handling.

**Stack trace:**

Image

1. Create a partitioned primary-key table in Fluss
2. Query the $changelog virtual table via Flink SQL (e.g., SELECT * FROM table$changelog)
3. The Flink optimizer's statistics recomputation triggers listPartitions on the virtual table name, causing the exception

### Solution

Add a toPhysicalTablePath() helper method to FlinkCatalog that strips $changelog, $binlog, and $lake suffixes from the table name. Update tableExists(), listPartitions(), createPartition(), and dropPartition() to use this helper instead of toTablePath().

naivedogger:
### Purpose

Linked issue: close #3640 

When the Flink optimizer calls `listPartitions()` on a `$changelog` or `$binlog` virtual table, it throws `TableNotExistException` because the virtual table suffix is not stripped before querying the Fluss server.

`getTable()` correctly dispatches virtual table names to `getVirtualChangelogTable()`/`getVirtualBinlogTable()`, but `tableExists()`, `listPartitions()`, `createPartition()`, and `dropPartition()` all pass the unstripped virtual table name directly to the admin API, which fails because no physical table named `xxx$changelog` exists.

### Brief change log

- Added `toPhysicalTablePath()` private helper method to `FlinkCatalog` that strips `$changelog`, `$binlog`, and `$lake` suffixes from the table name, returning the base physical table path
- Updated `tableExists()` to use `toPhysicalTablePath()` — virtual tables now correctly report existence based on the base table
- Updated `listPartitions(ObjectPath, CatalogPartitionSpec)` to use `toPhysicalTablePath()` — virtual tables now list partitions of the base table
- Updated `createPartition()` and `dropPartition()` to use `toPhysicalTablePath()` — partition management through virtual table names delegates to the base table

### Tests

- Added `FlinkCatalogTest#testVirtualTablePartitionsAndExistence` which verifies:
  - `tableExists()` returns `true` for `$changelog` and `$binlog` virtual table names
  - `listPartitions()` returns the same partitions as the base table for both virtual table suffixes
  - `listPartitions()` with a `CatalogPartitionSpec` works correctly on virtual table names
  - `createPartition()` and `dropPartition()` work through virtual table names
- Existing `FlinkCatalogTest#testOperatePartitions` continues to pass (no regression)

### API and Format

No API or storage format changes. The fix only affects internal catalog method behavior for virtual table names.

### Documentation

No documentation changes needed — this is a bug fix that makes virtual table behavior consistent with `getTable()`.

### DIFF
--- a/fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/catalog/FlinkCatalog.java
+++ b/fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/catalog/FlinkCatalog.java
@@ -705,7 +705,7 @@ public class FlinkCatalog extends AbstractCatalog {
                     CatalogException {
         // $changelog / $binlog are read-only virtual tables; reject partition mutations.
         if (isVirtualTable(objectPath)) {
-            throw new CatalogException(
+            throw new UnsupportedOperationException(
                     String.format(
                             "Cannot create partition on read-only virtual table %s in %s. "
                                     + "Please create the partition on the base table instead.",
@@ -759,7 +759,7 @@ public class FlinkCatalog extends AbstractCatalog {
             throws PartitionNotExistException, CatalogException {
         // $changelog / $binlog are read-only virtual tables; reject partition mutations.
         if (isVirtualTable(objectPath)) {
-            throw new CatalogException(
+            throw new UnsupportedOperationException(
                     String.format(
                             "Cannot drop partition on read-only virtual table %s in %s. "
                                     + "Please drop the partition on the base table instead.",
--- a/fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/catalog/FlinkCatalog.java
+++ b/fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/catalog/FlinkCatalog.java
@@ -490,7 +490,8 @@ public class FlinkCatalog extends AbstractCatalog {
 
     @Override
     public boolean tableExists(ObjectPath objectPath) throws CatalogException {
-        TablePath tablePath = toTablePath(objectPath);
+        // For virtual tables ($changelog, $binlog), check if the base physical table exists
+        TablePath tablePath = toPhysicalTablePath(objectPath);
         try {
             return admin.tableExists(tablePath).get();
         } catch (Exception e) {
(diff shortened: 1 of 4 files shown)


MehulBatra:
Thanks for tackling #3640. The root cause is correct, and the `listPartitions()` fix is the right approach.
A couple of comments:
>
> * `listPartitions()` stripping the virtual suffix is exactly what we want. A partitioned `$changelog`/`$binlog` read should resolve partitions from the base table.
> * `tableExists()` returning `true` for virtual tables also makes sense since they are queryable.
>
> One thing I think should change is `createPartition()` and `dropPartition()`. `$changelog` and `$binlog` are read-only virtual tables, so partition operations should be rejected with a clear error instead of being forwarded to the base table. Because of that, the corresponding tests should expect these operations to fail rather than succeed.
>
> One small nit: I would also keep the `$lake` handling out of `toPhysicalTablePath()`. This PR is about `$changelog` and `$binlog`, and adding `$lake` changes the behavior of `tableExists()` while introducing unrelated scope. It would be better to keep the helper limited to the virtual tables this PR is fixing.

 Overall, the read path changes look good to me. I just think we should reject mutating DDL on virtual tables and keep the `$lake` changes for a separate PR.

MehulBatra (commented):
Left one small comment

naivedogger:
Thanks @MehulBatra for your review! I have pushed an update to address the comments:

- Reject createPartition / dropPartition on read-only $changelog / $binlog virtual tables with a clear error (read path unchanged).
- Reverted the out-of-scope $lake handling in toPhysicalTablePath().
- Updated tests to expect these operations to fail.

PTAL, thanks!

loserwang1024 (approved):
LGTM

PR #3643, opened 2026-07-14
[rust] Hash-distribute log-table appends by declared bucket key

### LINKED ISSUE #3644: [rust] Log-table appends ignore the declared bucket key
### Search before asking

### Description

On a log table with a declared bucket key (`distributed_by(n, [keys])`), the Rust client doesn't hash appends by that key and every row in a flush lands in one bucket via the sticky assigner, regardless of the key. 
The KV/upsert path and the Java client both hash by the bucket key, so bucket-key co-location is silently broken,
and the same table written by the Rust vs Java client gets bucketed differently.

### Willingness to contribute

fresh-borzoni:
closes https://github.com/apache/fluss/issues/3644

Log tables with a declared bucket key weren't hashing appends by it and every row in a flush landed in one bucket (sticky), unlike the KV/upsert path and the Java client, so bucket-key co-location was silently broken.

Now the append path hashes by the bucket key like Java: `append(row)` hashes each row, and `append_arrow_batch` splits a batch across buckets (single-bucket batches send as-is, mixed batches fan out per bucket). 

Tables without a bucket key are unchanged.

### DIFF
--- a/fluss-rust/crates/fluss/src/client/table/append.rs
+++ b/fluss-rust/crates/fluss/src/client/table/append.rs
@@ -15,13 +15,18 @@
 // specific language governing permissions and limitations
 // under the License.
 
+use crate::bucketing::BucketingFunction;
 use crate::client::table::partition_getter::{PartitionGetter, get_physical_path};
 use crate::client::{WriteRecord, WriteResultFuture, WriterClient};
-use crate::error::Error::IllegalArgument;
+use crate::error::Error::{IllegalArgument, UnexpectedError};
 use crate::error::Result;
 use crate::metadata::{PhysicalTablePath, TableInfo, TablePath};
+use crate::row::encode::{KeyEncoder, KeyEncoderFactory};
 use crate::row::{ColumnarRow, InternalRow};
-use arrow::array::RecordBatch;
+use arrow::array::{RecordBatch, UInt32Array};
+use bytes::Bytes;
+use parking_lot::Mutex;
+use std::collections::HashMap;
 use std::sync::Arc;
 
 pub struct TableAppend {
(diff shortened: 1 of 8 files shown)


fresh-borzoni:
@charlesdong1991 @leekeiabstraction PTAL 🙏

fluss-rust/crates/fluss/src/client/table/append.rs
```diff
@@ -141,12 +186,61 @@ impl AppendWriter {
...
+        }
+
+        // Group rows by bucket, keeping one key per bucket (it hashes back there).
+        let num_buckets = self.table_info.get_num_buckets();
+        let batch_arc = Arc::new(batch.clone());
+        let row_type = Arc::new(self.table_info.row_type.clone());
+        let mut groups: HashMap<i32, (Vec<u32>, Bytes)> = HashMap::new();
+        for i in 0..batch.num_rows() {
+            let row = ColumnarRow::new(Arc::clone(&batch_arc), Arc::clone(&row_type), i, None)?;
+            let (bucket, key) = router.bucket_of(&row, num_buckets)?;
```
charlesdong1991: Nit: it might not be very efficient when row numbers are large given this will do full typed-batch builds on each iteration, because it will also downcast each column? maybe can utilize set_row_id in ColumnarRow and move the cursor instead?

fluss-rust/crates/fluss/src/client/table/append.rs
```diff
@@ -141,12 +186,61 @@ impl AppendWriter {
             Arc::new(PhysicalTablePath::of(Arc::clone(&self.table_path)))
         };
 
+        let Some(router) = self.bucket_router.as_ref() else {
+            return self.send_arrow_batch(batch, physical_table_path, None);
+        };
+        if batch.num_rows() == 0 {
+            return self.send_arrow_batch(batch, physical_table_path, None);
```
charlesdong1991: so empty batch will result in sticky assigner, but if the first batch is empty, will this mistake all following keyed appends? 

maybe we can add a test to see how it behaves?

charlesdong1991 (commented):
very nice! I left two nit comments

fluss-rust/crates/fluss/src/client/table/append.rs
fresh-borzoni: Good catch, it's actually a bug, thank you for spotting, fixed

fluss-rust/crates/fluss/src/client/table/append.rs
fresh-borzoni: Good idea 👍

fresh-borzoni:
@charlesdong1991 Thank you for the review, addressed, PTAL 🙏

charlesdong1991 (approved):
LGTM!

PR #3645, opened 2026-07-14
[server] Coordinator standby tracks dynamic config changes to avoid s…

### LINKED ISSUE #3625: [server] Coordinator standby may use stale dynamic configs after failover
### Search before asking

### Fluss version

0.9.0 (latest release)

## Description

Coordinator-side dynamic configs may become stale on an already-running standby `CoordinatorServer`. The coordinator `DynamicConfigManager` currently fetches dynamic configs from ZooKeeper during startup, but coordinator instances ignore later config change notifications. As a result, a standby coordinator does not continuously track the latest dynamic configs while it remains standby.

When that standby later becomes the active coordinator, leader-only components are initialized and registered against the local `DynamicConfigManager`. If the local `DynamicServerConfig.currentConfig` is stale, those components may be configured with old values even though ZooKeeper already contains newer dynamic configs.

This is not specific to one config option. It can affect coordinator-side dynamic configs in general. 

## Impact

After coordinator failover, the new active coordinator may not respect the latest dynamic configs stored in ZooKeeper.

## Root Cause

Coordinator `DynamicConfigManager` uses `isCoordinator = true`, and the config change notification handler returns early for coordinator instances. Therefore, standby coordinators do not apply dynamic config changes after their initial startup fetch.

Leader-only components, such as `RebalanceManager`, are registered when the coordinator becomes leader. Their initial configuration comes from the local `DynamicServerConfig.currentConfig`, which may be stale on a long-running standby.

## Proposed Solution

Refresh dynamic configs from ZooKeeper when a standby CoordinatorServer becomes leader, before registering leader-only components.

Specifically:

1. Add a method to `DynamicConfigManager`, for example:

   ```java
   public void refreshFromZooKeeper() throws Exception {
       Map<String, String> entityConfigs = zooKeeperClient.fetchEntityConfig();
       dynamicServerConfig.updateDynamicConfig(entityConfigs, true);
   }

### Solution

_No response_

fhan688:
### Purpose

Linked issue: close #3625

On a `CoordinatorServer`, `DynamicConfigManager` fetched dynamic configs from ZooKeeper only once at startup and then ignored every later config-change notification: the notification handler returned early whenever `isCoordinator == true`. As a result, a long-running **standby** coordinator's local config drifted away from ZooKeeper.

When such a standby was later promoted to leader, its components were configured from that stale local snapshot even though ZooKeeper already held newer values. The most impactful case is SASL/PLAIN multi-user credentials (`security.sasl.plain.credentials`, hot-reloaded by `FlussProtocolPlugin` and registered on both server roles): after failover, the new leader could authenticate clients against a frozen credential set — a dynamically added user would be rejected, and a dynamically **removed** user would still be accepted.

This PR makes standby coordinators continuously apply dynamic config-change notifications, exactly like tablet servers do, so a promoted leader always reflects the latest config.

### Brief change log

- `DynamicConfigManager`: remove the `isCoordinator` early-return in the change-notification handler, and drop the now-dead `isCoordinator` field and constructor parameter. Both server roles now apply notifications.
- Fix a concurrency issue that this change surfaces on the leader, which now both edits configs (via `alterConfigs`) and listens for notifications. Previously the notification path fetched from ZooKeeper **outside** the write lock, and `alterConfigs` applied-then-persisted without holding the lock across both steps, so a stale fetch could be applied on top of a newer, already-committed change. `DynamicServerConfig` now exposes:
  - `refreshDynamicConfig(configSupplier, skipErrorConfig)` — fetch and apply atomically under the write lock (used by `startup()` and the notification handler);
  - `updateAndPersistDynamicConfig(configs, persistAction)` — apply locally and persist to ZooKeeper atomically under the write lock (used by `alterConfigs`).
- Update `CoordinatorServer` and `TabletServer` to the two-argument `DynamicConfigManager` constructor.

### Tests

- `DynamicConfigChangeTest#testTracksConfigChangeNotificationsFromOtherWriter` (new): a manager applies a config change written to ZooKeeper by another writer — the core behavior the fix enables. Verified it fails without the fix (`expected 3 but was 0`).
- `CoordinatorHighAvailabilityITCase#testStandbyTracksDynamicConfigAndNewLeaderUsesItAfterFailover` (new): alter a dynamic config on the leader, assert the standby tracks it, kill the leader's ZK session to trigger failover, and assert the promoted leader still uses the up-to-date value.
  Verified it fails without the fix ("Standby coordinator did not apply dynamic config change while standby").
- Updated the existing `DynamicConfigManager` construction call sites in `DynamicConfigChangeTest` and made each test close its manager after use (previously leaked notification watchers, now material because coordinators react to notifications).
- Full runs pass: `DynamicConfigChangeTest` (26), `CoordinatorHighAvailabilityITCase` (6), and the server startup/election suites (`CoordinatorServerTest`, `TabletServerTest`, `CoordinatorServerElectionTest`).

### API and Format

No public API or storage-format changes. `DynamicConfigManager`'s constructor signature changed, but it is an internal server class. No RPC message or ZooKeeper znode-format changes.

### Documentation

No. This is a bug fix with no new user-facing feature or configuration.

### DIFF
--- a/fluss-server/src/main/java/org/apache/fluss/server/DynamicConfigManager.java
+++ b/fluss-server/src/main/java/org/apache/fluss/server/DynamicConfigManager.java
@@ -40,7 +40,15 @@ import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 
-/** Manager for dynamic configurations. */
+/**
+ * Manager for dynamic configurations.
+ *
+ * <p>Used by both the CoordinatorServer and the TabletServer. Every instance, regardless of server
+ * role or leadership status, continuously applies dynamic config-change notifications from
+ * ZooKeeper. In particular a standby CoordinatorServer keeps tracking changes so that, once it is
+ * promoted to leader, its components (e.g. SASL credentials) already reflect the latest config
+ * rather than a stale snapshot taken at startup.
+ */
 public class DynamicConfigManager {
     private static final Logger LOG = LoggerFactory.getLogger(DynamicConfigManager.class);
     private static final long CHANGE_NOTIFICATION_EXPIRATION_MS = 15 * 60 * 1000L;
@@ -48,13 +56,10 @@ public class DynamicConfigManager {
     private final DynamicServerConfig dynamicServerConfig;
     private final ZooKeeperClient zooKeeperClient;
     private final ZkNodeChangeNotificationWatcher configChangeListener;
-    private final boolean isCoordinator;
 
-    public DynamicConfigManager(
-            ZooKeeperClient zooKeeperClient, Configuration configuration, boolean isCoordinator) {
+    public DynamicConfigManager(ZooKeeperClient zooKeeperClient, Configuration configuration) {
         this.dynamicServerConfig = new DynamicServerConfig(configuration);
         this.zooKeeperClient = zooKeeperClient;
-        this.isCoordinator = isCoordinator;
         this.configChangeListener =
                 new ZkNodeChangeNotificationWatcher(
                         zooKeeperClient,
@@ -68,8 +73,7 @@ public class DynamicConfigManager {
     public void startup() throws Exception {
         try {
             configChangeListener.start();
-            Map<String, String> entityConfigs = zooKeeperClient.fetchEntityConfig();
-            dynamicServerConfig.updateDynamicConfig(entityConfigs, true);
+            dynamicServerConfig.refreshDynamicConfig(zooKeeperClient::fetchEntityConfig, true);
         } catch (Exception e) {
             LOG.error("Failed to update dynamic configs from zookeeper", e);
         }
@@ -312,10 +316,10 @@ public class DynamicConfigManager {
 
     @VisibleForTesting
     protected void alterServerConfigs(Map<String, String> configsProps) throws Exception {
-        dynamicServerConfig.updateDynamicConfig(configsProps, false);
-
-        // Apply to zookeeper only after verification.
-        zooKeeperClient.upsertServerEntityConfig(configsProps);
+        // Apply locally and persist to zookeeper atomically, so a concurrent change notification
+        // (this server may also be listening) cannot interleave and revert the change.
+        dynamicServerConfig.updateAndPersistDynamicConfig(
+                configsProps, () -> zooKeeperClient.upsertServerEntityConfig(configsProps));
     }
 
     private static void validateListOrMapType(String configKey) {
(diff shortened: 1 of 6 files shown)


fluss-server/src/main/java/org/apache/fluss/server/DynamicConfigManager.java
```diff
@@ -40,21 +40,26 @@
...
     private static final Logger LOG = LoggerFactory.getLogger(DynamicConfigManager.class);
     private static final long CHANGE_NOTIFICATION_EXPIRATION_MS = 15 * 60 * 1000L;
 
     private final DynamicServerConfig dynamicServerConfig;
     private final ZooKeeperClient zooKeeperClient;
     private final ZkNodeChangeNotificationWatcher configChangeListener;
-    private final boolean isCoordinator;
 
-    public DynamicConfigManager(
-            ZooKeeperClient zooKeeperClient, Configuration configuration, boolean isCoordinator) {
```
loserwang1024, line 54: As discussed, please only allow the follower coordinator to subscribe to ZooKeeper notifications. The leader already maintains the latest configuration changes, so having it subscribe here may cause version rollback issues (e.g., A → B → A). When a follower transitions to become the leader, please disable the subscription at that point.

fhan688: > As discussed, please only allow the follower coordinator to subscribe to ZooKeeper notifications. The leader already maintains the latest configuration changes, so having it subscribe here may cause version rollback issues (e.g., A → B → A). When a follower transitions to become the leader, please disable the subscription at that point.

Thanks @loserwang1024, you're right — done. I've reworked it so the leader no longer consumes notifications:

- Added a `listeningEnabled` gate to `DynamicConfigManager`. The notification handler no-ops when it's off, so a paused instance won't react to ZooKeeper changes.
- `pauseListening()` is called in `CoordinatorServer#initCoordinatorLeader` (on promotion) and `resumeListening()` in `cleanupCoordinatorLeader` (on demotion). A standby/tablet server keeps listening as before.
- On demotion, `resumeListening()` also re-fetches once from ZooKeeper: while the server was leader the watcher kept advancing its last-processed sequence id but skipped the bodies, so past notifications won't replay — the explicit re-sync picks up whatever the new leader changed meanwhile.
- Reverted the write-lock refactor from the previous revision (`refreshDynamicConfig`/`updateAndPersistDynamicConfig`) back to the original `updateDynamicConfig`. With the leader no longer listening, a single instance only ever has one writer, so that locking is no longer needed.

I could reproduce the A → B → A rollback you described in a unit test (`SET null` → `DELETE` restoring the default was rolled back to `null` by the manager's own notification), which is now covered by `testPausedManagerIgnoresNotificationsAndResumeReSyncs`. The failover IT is extended to check the promoted leader keeps the value it tracked as standby. PTAL.

loserwang1024 (approved):
LGTM

PR #3647, opened 2026-07-14
[server] Offload NotifyLeaderAndIsr to dedicated single-thread executor

### LINKED ISSUE #3646: Offload NotifyLeaderAndIsr from RPC worker threads
### Search before asking

### Background

`NotifyLeaderAndIsr` is currently processed on the RPC worker thread. The underlying `becomeLeaderOrFollower` operation may perform relatively expensive replica state transitions, such as initializing local replicas and updating replica fetchers. A slow transition can therefore occupy an RPC worker thread and delay unrelated RPC requests.

### Solution

Introduce a dedicated single-thread executor for `NotifyLeaderAndIsr` operations.
The executor is created and managed by `TabletServer` , injected into `TabletService` , and gracefully shut down before replica-related resources are closed. `TabletService` parses the request on the RPC worker thread and submits the replica state transition to the dedicated executor.
A single thread is sufficient because replica state changes are already serialized by the replica state change lock. It also preserves the execution order of `NotifyLeaderAndIsr` requests while preventing slow transitions from blocking other RPCs.
The RPC semantics remain unchanged: the response future is completed only after `becomeLeaderOrFollower` finishes.

### Solution

_No response_

### Anything else?

_No response_

### Willingness to contribute

zuston:
### Purpose

Linked issue: close #3646

`NotifyLeaderAndIsr` is currently processed on the RPC worker thread. The underlying `becomeLeaderOrFollower` operation may perform relatively expensive replica state transitions, such as initializing local replicas and updating replica fetchers. A slow transition can therefore occupy an RPC worker thread and delay unrelated RPC requests.

### Brief change log

Introduce a dedicated single-thread executor for `NotifyLeaderAndIsr` operations.
The executor is created and managed by `TabletServer` , injected into `TabletService` , and gracefully shut down before replica-related resources are closed. `TabletService` parses the request on the RPC worker thread and submits the replica state transition to the dedicated executor.

A single thread is sufficient because replica state changes are already serialized by the replica state change lock. It also preserves the execution order of `NotifyLeaderAndIsr` requests while preventing slow transitions from blocking other RPCs.
The RPC semantics remain unchanged: the response future is completed only after `becomeLeaderOrFollower` finishes.

### DIFF
--- a/fluss-server/src/main/java/org/apache/fluss/server/tablet/TabletService.java
+++ b/fluss-server/src/main/java/org/apache/fluss/server/tablet/TabletService.java
@@ -154,6 +154,7 @@ public final class TabletService extends RpcServiceBase implements TabletServerG
     private final ScannerManager scannerManager;
     private final CoordinatorGateway coordinatorGateway;
     private final String interListenerName;
+    private final ExecutorService replicaStateChangeExecutor;
 
     public TabletService(
             int serverId,
@@ -165,6 +166,7 @@ public final class TabletService extends RpcServiceBase implements TabletServerG
             @Nullable Authorizer authorizer,
             DynamicConfigManager dynamicConfigManager,
             ExecutorService ioExecutor,
+            ExecutorService replicaStateChangeExecutor,
             ScannerManager scannerManager,
             CoordinatorGateway coordinatorGateway,
             String interListenerName) {
@@ -184,6 +186,7 @@ public final class TabletService extends RpcServiceBase implements TabletServerG
         this.scannerManager = scannerManager;
         this.coordinatorGateway = coordinatorGateway;
         this.interListenerName = interListenerName;
+        this.replicaStateChangeExecutor = replicaStateChangeExecutor;
     }
 
     @Override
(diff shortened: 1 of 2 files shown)


zuston:
Could you help review this? @fresh-borzoni

fluss-server/src/main/java/org/apache/fluss/server/tablet/TabletService.java
```diff
@@ -359,12 +362,26 @@ public CompletableFuture<GetTableStatsResponse> getTableStats(GetTableStatsReque
...
         CompletableFuture<NotifyLeaderAndIsrResponse> response = new CompletableFuture<>();
+        int coordinatorEpoch = notifyLeaderAndIsrRequest.getCoordinatorEpoch();
         List<NotifyLeaderAndIsrData> notifyLeaderAndIsrRequestData =
                 getNotifyLeaderAndIsrRequestData(notifyLeaderAndIsrRequest);
-        replicaManager.becomeLeaderOrFollower(
-                notifyLeaderAndIsrRequest.getCoordinatorEpoch(),
-                notifyLeaderAndIsrRequestData,
-                result -> response.complete(makeNotifyLeaderAndIsrResponse(result)));
+        try {
+            replicaStateChangeExecutor.execute(
```
fresh-borzoni: Should stopReplicas go through this executor too? 
It takes the same replicaStateChangeLock but still runs on the RPC worker, so a stop can now overtake queued notifies, which the per-connection ordering used to prevent.

Same-executor for both would fix the ordering, and also keep stops from blocking an RPC worker on the lock while a slow transition runs. 
WDYT?

fresh-borzoni (commented):
@zuston Thank you, makes sense overall,  left a suggestion, PTAL

fluss-server/src/main/java/org/apache/fluss/server/tablet/TabletService.java
zuston: nice suggestion. let me add the remaining rpcs into this dedicated executor.

zuston:
updated. @fresh-borzoni

fresh-borzoni (approved):
@zuston Thank you, LGTM 👍

PR #3649, opened 2026-07-14
[server] Wait for coordinator leader cleanup during shutdown

### LINKED ISSUE #3648: [Bug] Coordinator leader cleanup may be interrupted during shutdown
### Search before asking

### Fluss version

0.9.0 (latest release)

### Please describe the bug 🐞

`CoordinatorLeaderElection.close()` closes the `LeaderLatch` and immediately shuts down the leader callback executor
```java
leaderLatch.close();
leaderCallbackExecutor.shutdownNow();
```

Closing the LeaderLatch triggers the notLeader() callback, which submits leader cleanup asynchronously to leaderCallbackExecutor.
There is a race where shutdownNow() interrupts or cancels the cleanup task before it finishes. As a result, cleanupCoordinatorLeader() may not complete and the coordinator leader registration may remain in ZooKeeper during shutdown.
This can cause a newly started CoordinatorServer to repeatedly fail registration:
```java
Node for coordinator server already exists in Zookeeper.
Retrying register after 3000 ms....
```

It can also cause coordinator restart tests to time out while waiting for the old ZooKeeper node to disappear.

### Solution

CoordinatorLeaderElection.close() should ensure that leader cleanup finishes before shutting down the callback executor.
The cleanup must still execute exactly once when LeaderLatch.close() and the explicit shutdown path race with each other.

loserwang1024:
### Purpose

Linked issue: close https://github.com/apache/fluss/issues/3648

### DIFF
--- a/fluss-server/src/main/java/org/apache/fluss/server/coordinator/CoordinatorLeaderElection.java
+++ b/fluss-server/src/main/java/org/apache/fluss/server/coordinator/CoordinatorLeaderElection.java
@@ -64,6 +64,8 @@ public class CoordinatorLeaderElection implements AutoCloseable {
     private final AtomicReference<CompletableFuture<Void>> pendingCleanup =
             new AtomicReference<>(CompletableFuture.completedFuture(null));
 
+    private volatile Consumer<Throwable> cleanupLeaderServices;
+
     public CoordinatorLeaderElection(ZooKeeperClient zkClient, String serverId) {
         this.serverId = serverId;
         this.leaderLatch =
@@ -76,7 +78,7 @@ public class CoordinatorLeaderElection implements AutoCloseable {
                         r -> {
                             Thread t = new Thread(r, "coordinator-leader-callback-" + serverId);
                             // Daemon threads ensure the JVM can exit even if close() is not
-                            // called. Orderly shutdown is handled by close() -> shutdownNow().
+                            // called. Orderly shutdown is handled by close().
                             t.setDaemon(true);
                             return t;
                         });
@@ -93,6 +95,7 @@ public class CoordinatorLeaderElection implements AutoCloseable {
      */
     public void startElectLeaderAsync(
             Runnable initLeaderServices, Consumer<Throwable> cleanupLeaderServices) {
+        this.cleanupLeaderServices = cleanupLeaderServices;
         leaderLatch.addListener(
                 new LeaderLatchListener() {
                     @Override
@@ -145,32 +148,7 @@ public class CoordinatorLeaderElection implements AutoCloseable {
 
                     @Override
                     public void notLeader() {
-                        if (isLeader.compareAndSet(true, false)) {
-                            LOG.warn(
-                                    "Coordinator server {} has lost the leadership, cleaning up leader services.",
-                                    serverId);
-                            // Submit cleanup to leaderCallbackExecutor. The cached thread
-                            // pool can spawn a new thread even if init is still running.
-                            // The cleanup completion is tracked via pendingCleanup so
-                            // that subsequent init waits for it.
-                            CompletableFuture<Void> cleanupFuture = new CompletableFuture<>();
-                            pendingCleanup.set(cleanupFuture);
-                            leaderCallbackExecutor.execute(
-                                    () -> {
-                                        try {
-                                            if (cleanupLeaderServices != null) {
-                                                cleanupLeaderServices.accept(null);
-                                            }
-                                        } catch (Exception e) {
-                                            LOG.error(
-                                                    "Failed to cleanup leader services for server {}",
-                                                    serverId,
-                                                    e);
-                                        } finally {
-                                            cleanupFuture.complete(null);
-                                        }
-                                    });
-                        }
+                        triggerCleanupLeaderServices();
                     }
                 });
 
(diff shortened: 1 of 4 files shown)


fluss-server/src/main/java/org/apache/fluss/server/coordinator/CoordinatorLeaderElection.java
```diff
@@ -194,10 +172,51 @@ public void close() {
...
+            leaderCallbackExecutor.shutdownNow();
+        }
     }
 
     public boolean isLeader() {
         return this.isLeader.get();
     }
+
+    private void triggerCleanupLeaderServices() {
+        if (isLeader.compareAndSet(true, false)) {
```
swuferhong: `isLeader` is only set to true after `initLeaderServices.run()` completes. If `close()` is called before initialization completes, this problem also will be happen.  The implementation should explicitly track `initialization/closing` state instead of using `isLeader` as both readiness state and cleanup ownership state?

swuferhong (commented):
LGTM overall. I left one comment.

fresh-borzoni (commented):
@loserwang1024 @swuferhong Thank you, left a suggestion, take a look if I'm missing smth, please.

Good catch @swuferhong 👍 
Probably we can go even further:
There's also a case beyond close(): leadership flap while init is still running. notLeader's CAS sees false, cleanup is never queued, and the re-election starts a second init concurrently with the first, so pendingCleanup doesn't cover the exact case it was built for.
I checked why cleanup got its own thread in #2780 (and reasoning "avoid blocking init tasks" in the git history): both callbacks are synchronized in CoordinatorServer, so they already run one at a time and the extra thread never bought real parallelism. The only unlocked part is init's fence/register, and curator's bounded retries mean it can't hang forever.

So, combining: single-thread executor + a small state field written only from it STANDBY / INITIALIZING / LEADER / CLOSED. 
Cleanup task cleans if INITIALIZING or LEADER (fenced init stops being special), isLeader() stays state == LEADER, all checks happen inside the tasks in event order, and pendingCleanup + the 60s wait get deleted. 

It's the same shape as Flink's DefaultLeaderElectionService - single-thread leadershipOperationExecutor + explicit running state. 
WDYT?

loserwang1024:
@fresh-borzoni It's exactly what I want to do! @fresh-borzoni @swuferhong , CC

fluss-server/src/main/java/org/apache/fluss/server/coordinator/CoordinatorLeaderElection.java
```diff
@@ -186,18 +140,138 @@ public void notLeader() {
...
+            initializationFailure = e;
+        } catch (Exception e) {
+            LOG.error("Failed to initialize leader services for server {}", serverId, e);
+            initializationFailure = e;
+        }
+
+        if (initializationFailure == null && !closing.get()) {
+            state = State.LEADER;
+        } else {
+            cleanupLeaderServices(initializationFailure);
```
fresh-borzoni, line 234: If init throws we clean up and go STANDBY but still hold the latch, so nothing re-elects and we're left with a leader that isn't serving. 
Looks pre-existing, maybe a follow-up, but could we at least add a test for the cleanup path here? wdyt?

fluss-server/src/main/java/org/apache/fluss/server/coordinator/CoordinatorLeaderElection.java
```diff
@@ -186,18 +140,138 @@ public void notLeader() {
...
+        leaderCallbackExecutor.shutdown();
+        awaitCallbackExecutorTermination();
     }
 
     public boolean isLeader() {
-        return this.isLeader.get();
+        return !closing.get() && state == State.LEADER;
+    }
+
+    private void submitLeadershipEvent(boolean leader, Runnable initLeaderServices) {
```
fresh-borzoni: nit: initLeaderServices is passed to the standby branch too but only leader uses it — drop it?

fluss-server/src/main/java/org/apache/fluss/server/coordinator/CoordinatorLeaderElection.java
```diff
@@ -186,18 +140,138 @@ public void notLeader() {
...
+                                cleanupLeaderServices(null);
+                            }
+                        } finally {
+                            closeFuture.complete(null);
+                        }
+                    });
         }
 
-        leaderCallbackExecutor.shutdownNow();
+        closeFuture.join();
```
fresh-borzoni: Think my earlier "can't hang" only covered init - cleanup takes lock and shuts down the event processor / channel manager, none of which are curator-bounded, so close() can now block the shutdown thread until it's SIGKILL'd. 

The callback threads are daemon and old close() didn't wait at all, so maybe just bound the wait (timeout on join) and move on? shutdownNow() won't help if it's stuck on lock anyway. wdyt?

btw there are two waits here - join() and awaitCallbackExecutorTermination(). Both need bounding, else the hang just moves to the second one.

fresh-borzoni (commented):
@loserwang1024 Thank you, looks good, only a few comments, PTAL

fluss-server/src/main/java/org/apache/fluss/server/coordinator/CoordinatorLeaderElection.java
loserwang1024: done

fluss-server/src/main/java/org/apache/fluss/server/coordinator/CoordinatorLeaderElection.java
loserwang1024: You are right. This behavior comes from the existing strategy.

I also considered releasing the current leadership when leader initialization fails, for example by closing the current LeaderLatch. However, that alone would be problematic for a single-coordinator deployment, because there may be no other candidate to take over. To handle this correctly, we likely need to release the current election leadership and then rejoin the election, probably with some backoff to avoid a tight retry loop.

That recovery behavior is outside the scope of this PR, so I think we should handle it in a separate follow-up. For this PR, I added a regression/reproduction test to document the current behavior:
`org.apache.fluss.server.coordinator.CoordinatorLeaderElectionTest#testInitializationFailureKeepsLeaderLatchAndBlocksReElection`.

fluss-server/src/main/java/org/apache/fluss/server/coordinator/CoordinatorLeaderElection.java
loserwang1024: done

fresh-borzoni (approved):
@loserwang1024 Thank you, LGTM 👍

PR #3651, opened 2026-07-14
[kv] Close KV tablets concurrently during shutdown

### LINKED ISSUE #3650: Improve TabletServer shutdown by closing KV tablets concurrently
### Search before asking

### Description

`KvManager.shutdown()` currently closes all KV tablets sequentially. Each `KvTablet.close()` waits for its write lock and closes the corresponding RocksDB instance before the next tablet can start closing.

When a TabletServer contains many KV tablets, the shutdown latency grows linearly with the number of tablets. This may cause the TabletServer to exceed the process termination grace period and eventually result in an unclean shutdown.

### Observed behavior

The issue was observed in a cluster containing approximately 32,000 buckets （no data）.

In one TabletServer shutdown:

- `KvManager` started shutting down at `13:20:42.235`.
- `KvManager` completed shutting down at `13:21:11.705`.
- The total `KvManager` shutdown time was approximately **29,470 ms**.

In another test with KV data (3GB per bucket), the intervals between consecutive close-start logs can be used to approximate the preceding tablet's close time:

- Minimum: approximately **10 ms**
- Maximum: approximately **953 ms**
- Average: approximately **236 ms**

### Solution

_No response_

### Anything else?

_No response_

### Willingness to contribute

swuferhong:
### Purpose

Linked issue: close #3650 

KvManager previously closed KV tablets sequentially, causing shutdown latency to grow linearly with the number of tablets.

### DIFF
--- a/fluss-server/src/main/java/org/apache/fluss/server/ServerBase.java
+++ b/fluss-server/src/main/java/org/apache/fluss/server/ServerBase.java
@@ -179,22 +179,42 @@ public abstract class ServerBase implements AutoCloseableAsync, FatalErrorHandle
                 "Fatal error occurred while running the {}. Shutting it down...",
                 getServerName(),
                 exception);
-        if (ExceptionUtils.isJvmFatalError(exception)) {
-            System.exit(-1);
+        if (ExceptionUtils.isJvmFatalOrOutOfMemoryError(exception)) {
+            exitJvm();
         } else if (fatalErrorShutdownTriggered.compareAndSet(false, true)) {
-            FutureUtils.orTimeout(
-                    getTerminationFuture(),
-                    FATAL_ERROR_SHUTDOWN_TIMEOUT_MS,
-                    TimeUnit.MILLISECONDS,
-                    String.format(
-                            "Waiting for %s shutting down timed out after %s ms.",
-                            getServerName(), FATAL_ERROR_SHUTDOWN_TIMEOUT_MS));
-
-            Thread shutdownThread =
-                    new ExecutorThreadFactory(getServerName() + "-fatal-error-shutdown")
-                            .newThread(() -> closeAsync(Result.FAILURE));
-            shutdownThread.start();
+            try {
+                FutureUtils.orTimeout(
+                        getTerminationFuture(),
+                        FATAL_ERROR_SHUTDOWN_TIMEOUT_MS,
+                        TimeUnit.MILLISECONDS,
+                        String.format(
+                                "Waiting for %s shutting down timed out after %s ms.",
+                                getServerName(), FATAL_ERROR_SHUTDOWN_TIMEOUT_MS));
+                startFatalErrorShutdownThread();
+            } catch (Throwable shutdownFailure) {
+                try {
+                    LOG.error(
+                            "Failed to start fatal-error shutdown for {}. Terminating the process.",
+                            getServerName(),
+                            shutdownFailure);
+                } finally {
+                    exitJvm();
+                }
+            }
         }
+    }
+
+    @VisibleForTesting
+    void startFatalErrorShutdownThread() {
+        Thread shutdownThread =
+                new ExecutorThreadFactory(getServerName() + "-fatal-error-shutdown")
+                        .newThread(() -> closeAsync(Result.FAILURE));
+        shutdownThread.start();
+    }
+
+    @VisibleForTesting
+    void exitJvm() {
+        System.exit(-1);
     }
 
     @Override
--- a/fluss-server/src/main/java/org/apache/fluss/server/ServerBase.java
+++ b/fluss-server/src/main/java/org/apache/fluss/server/ServerBase.java
@@ -36,6 +36,7 @@ import org.apache.fluss.server.utils.ShutdownHookUtil;
 import org.apache.fluss.server.utils.SignalHandler;
 import org.apache.fluss.utils.AutoCloseableAsync;
 import org.apache.fluss.utils.ExceptionUtils;
+import org.apache.fluss.utils.concurrent.ExecutorThreadFactory;
 import org.apache.fluss.utils.concurrent.FutureUtils;
 
 import org.slf4j.Logger;
@@ -49,6 +50,7 @@ import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
+import java.util.concurrent.atomic.AtomicBoolean;
 
 import static org.apache.fluss.server.utils.LogShutdownUtil.shutdownLogIfPossible;
 
@@ -77,11 +79,13 @@ public abstract class ServerBase implements AutoCloseableAsync, FatalErrorHandle
     protected FileSystem remoteFileSystem;
     protected PluginManager pluginManager;
 
+    private final AtomicBoolean fatalErrorShutdownTriggered = new AtomicBoolean();
+
+    private Thread shutDownHook;
+
     protected ServerBase(Configuration conf) {
         this.conf = conf;
     }
-
-    private Thread shutDownHook;
 
     protected static Configuration loadConfiguration(String[] args, String serverClassName) {
         try {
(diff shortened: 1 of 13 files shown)


fluss-server/src/main/java/org/apache/fluss/server/log/LogManager.java
```diff
@@ -473,69 +475,55 @@ public void shutdown() {
...
-                            throw new FlussRuntimeException(e);
-                        }
-                    };
-            jobsForTabletDir.add(pool.submit(runnable));
+    private void closeLogTablet(LogTablet logTablet) {
+        try {
+            logTablet.flush(true);
+            logTablet.close();
+        } catch (Exception e) {
+            LOG.warn("Exception while closing log tablet {}.", logTablet.getTableBucket(), e);
```
fresh-borzoni, line 508: If flush fails here, we still write the clean-shutdown marker below  and next startup skips log recovery because of it. 
Shouldn't a failed flush block the marker for that dir? 

The helper already fails the future when the action throws, so it's probably this: don't catch here, catch at the join and skip the marker, wdyt?

fresh-borzoni (commented):
@swuferhong Thank you, looks good overall, left one question, PTAL

fluss-server/src/main/java/org/apache/fluss/server/log/LogManager.java
swuferhong: Thanks @fresh-borzoni, good catch. The old implementation also wrote the clean-shutdown marker after catching an ExecutionException, so the issue predates this PR. However, this refactor currently swallows the exception even earlier. I’ll preserve the failure on the per-directory future and only write the marker when all close tasks for that directory complete successfully.

fresh-borzoni (approved):
@swuferhong Thanks, LGTM 👍

fluss-server/src/main/java/org/apache/fluss/server/log/LogManager.java
```diff
@@ -473,70 +475,63 @@ public void shutdown() {
...
-    private void waitForShutdownLogsInDir(LogShutdownTask shutdownTask) {
-        for (Future<?> future : shutdownTask.jobs) {
-            try {
-                future.get();
-            } catch (InterruptedException e) {
-                LOG.warn("Interrupted while shutting down LogManager.");
-            } catch (ExecutionException e) {
-                LOG.warn("There was an error in one of the threads during LogManager shutdown", e);
-            }
+    private void waitForShutdownLogsInDir(
```
platinumhamburg, line 513: Could we preserve each per-tablet close failure here? `CompletableFuture.allOf()` completes exceptionally if any child fails, but `join()` exposes only one exceptional cause. Before this change, `waitForShutdownLogsInDir` iterated over every `Future` and logged each `ExecutionException`.

If multiple `LogTablet`s in the same data directory fail to close, the new code logs only one failure and loses the other affected `TableBucket`s, which is an observability regression during shutdown. The close tasks still all run and the clean-shutdown marker decision remains correct, so this is non-blocking.

Logging failures with the `TableBucket` inside each tablet's close action, or retaining the child futures and aggregating all causes after `allOf`, would preserve the previous diagnostics.

fluss-server/src/main/java/org/apache/fluss/server/kv/KvManager.java
```diff
@@ -206,13 +206,7 @@ public void shutdown() {
...
         isShutdown = true;
         List<KvTablet> kvs = new ArrayList<>(currentKvs.values());
-        for (KvTablet kvTablet : kvs) {
-            try {
-                kvTablet.close();
-            } catch (Exception e) {
-                LOG.warn("Exception while closing kv tablet {}.", kvTablet.getTableBucket(), e);
-            }
-        }
+        closeTabletsConcurrently(kvs, "kv-tablet-closing", this::closeKvTablet).join();
```
platinumhamburg, line 209: Could we confirm that all users of the KV tablets have been quiesced before starting this parallel close?

`TabletServer.stopServices()` starts `rpcServer.closeAsync()` but does not await it before calling `kvManager.shutdown()`. `TabletService.shutdown()` is currently a no-op, and `ReplicaManager.shutdown()`—which stops snapshot, fetcher, and delayed-operation activity—runs only after the KV and Log managers have already shut down.

This race already existed with serial close, but this change can move up to `recoveryThreads` tablets into the closing state at once, potentially increasing the number of queued or in-flight operations that observe a closed RocksDB.

If pending operations are allowed to fail during server shutdown, could we document or test that contract? If graceful draining is intended, the RPC and Replica users should be awaited or otherwise quiesced before this call. An integration test with an in-flight put, fetch, or snapshot during TabletServer shutdown would make the intended behavior explicit.

fluss-server/src/main/java/org/apache/fluss/server/kv/KvTablet.java
```diff
@@ -670,37 +670,47 @@ private WalBuilder createWalBuilder(int schemaId, RowType rowType) throws Except
...
-                                    new KvStorageException("Failed to flush kv pre-write buffer."));
-                        }
-                    }
-                });
+                        });
+
+        if (flushError != null) {
+            // The fatal error handler may synchronously shut down the server, which closes this
+            // tablet from another thread. Invoke it after releasing kvLock to avoid deadlock.
+            fatalErrorHandler.onFatalError(
```
fresh-borzoni: This fixes kvLock, but we still call onFatalError under leaderIsrUpdateLock (this flush runs inside it, appendRecordsToLeader/putRecordsToLeader too). Should onFatalError just run shutdown on its own thread so no caller lock matters?

platinumhamburg (changes requested):
LGTM overall, just two minor comments.

fluss-server/src/main/java/org/apache/fluss/server/kv/KvTablet.java
swuferhong: Good idea!

fresh-borzoni (approved):
@swuferhong Thank you, LGTM
If @platinumhamburg doesn't have any further comments, feel free to merge 👍

platinumhamburg (approved):
LGTM.

fluss-server/src/main/java/org/apache/fluss/server/tablet/TabletServer.java
```diff
@@ -465,21 +478,8 @@ CompletableFuture<Void> stopServices() {
...
-
-                if (logManager != null) {
-                    logManager.shutdown();
-                }
-
-                if (replicaManager != null) {
-                    replicaManager.shutdown();
-                }
+                shutdownReplicaManager();
+                shutdownTabletManagers();
```
platinumhamburg, line 495: These calls are in the same `try` block, so any exception from `shutdownReplicaManager()` skips `shutdownTabletManagers()` and every cleanup step after it.

One concrete path is `checkpointHighWatermarks()`: a checkpoint write failure in one data directory is thrown after replica activity has already been stopped, but it still prevents KV and log tablets in every directory from being closed.

This deterministically turns the TabletServer shutdown into an unclean shutdown. Local logs are not force-flushed, recovery checkpoints and clean-shutdown markers are not written, and even healthy data directories must go through unclean recovery on the next startup.

Please preserve the shutdown order but handle these steps separately: record the ReplicaManager failure, continue shutting down the tablet managers and remaining resources, and report the aggregated failure afterward. Please also add a regression test where `shutdownReplicaManager()` fails and verify that `shutdownTabletManagers()` still runs.

fluss-server/src/main/java/org/apache/fluss/server/ServerBase.java
```diff
@@ -177,15 +181,19 @@ public void onFatalError(Throwable exception) {
...
                     FATAL_ERROR_SHUTDOWN_TIMEOUT_MS,
                     TimeUnit.MILLISECONDS,
                     String.format(
                             "Waiting for %s shutting down timed out after %s ms.",
                             getServerName(), FATAL_ERROR_SHUTDOWN_TIMEOUT_MS));
+
+            Thread shutdownThread =
+                    new ExecutorThreadFactory(getServerName() + "-fatal-error-shutdown")
+                            .newThread(() -> closeAsync(Result.FAILURE));
+            shutdownThread.start();
```
platinumhamburg: `isJvmFatalError()` does not include `OutOfMemoryError`, so an OOM—including “unable to create new native thread”—takes this branch.

The one-shot shutdown flag is set before the new thread is created. If `newThread()` or `start()` throws, shutdown never starts, and later fatal-error callbacks are ignored because the flag is already set.

Please treat OOM as immediately fatal by using `isJvmFatalOrOutOfMemoryError()`, or catch thread creation/start failures and terminate the process as a fallback. Please also add a test covering this failure path.

platinumhamburg (changes requested):
LGTM, just two minor issues.

swuferhong:
@platinumhamburg  comments addressed.

platinumhamburg (approved):
LGTM.

wuchong (approved):

PR #3654, opened 2026-07-14
[server] Batch LeaderAndIsr updates for controlled shutdown

### LINKED ISSUE #3653: [server] Speed up controlled shutdown leader migration with batch ZooKeeper operations
### Search before asking

### Motivation

During controlled shutdown, the coordinator needs to migrate bucket leaders away from the shutting-down tablet server.

Currently this path processes leader buckets one by one. Each bucket does synchronous ZooKeeper read/update of LeaderAndIsr, which becomes expensive when a tablet server owns many leader buckets. In one test with `10,131` leader buckets, controlled shutdown leader state changes took about `113s`, while dispatching the batched tablet server requests only took `22ms`. The main cost was ZooKeeper reads and updates.

We can improve this by batching LeaderAndIsr reads and updates for the controlled shutdown path, while preserving the existing leader election semantics. The optimized path should be limited to controlled shutdown, and should fall back to the existing individual update path when batch reads are incomplete or batch updates fail.

### Solution

_No response_

### Anything else?

_No response_

### Willingness to contribute

swuferhong:
### Purpose

Linked issue: close #3653 

Optimize controlled shutdown leader migration by batching LeaderAndIsr reads and updates in ZooKeeper. The new path is only used for controlled shutdown leader election and falls back to individual updates when batch reads are incomplete or batch updates fail.

This reduces the coordinator-side cost when a shutting-down tablet server owns many leader buckets, while preserving the existing election semantics.

### DIFF
--- a/fluss-server/src/main/java/org/apache/fluss/server/coordinator/statemachine/TableBucketStateMachine.java
+++ b/fluss-server/src/main/java/org/apache/fluss/server/coordinator/statemachine/TableBucketStateMachine.java
@@ -206,7 +206,7 @@ public class TableBucketStateMachine {
                             BucketState.OnlineBucket,
                             String.format(
                                     "Can't find partition name for partition: %s.",
-                                    tableBucket.getBucket()));
+                                    tableBucket.getPartitionId()));
                     continue;
                 }
             }
--- a/fluss-server/src/main/java/org/apache/fluss/server/coordinator/statemachine/TableBucketStateMachine.java
+++ b/fluss-server/src/main/java/org/apache/fluss/server/coordinator/statemachine/TableBucketStateMachine.java
@@ -383,7 +383,7 @@ public class TableBucketStateMachine {
                                 targetState,
                                 String.format(
                                         "Can't find partition name for partition: %s.",
-                                        tableBucket.getBucket()));
+                                        tableBucket.getPartitionId()));
                         return;
                     }
                 }
@@ -516,7 +516,7 @@ public class TableBucketStateMachine {
                             BucketState.OnlineBucket,
                             String.format(
                                     "Can't find partition name for partition: %s.",
-                                    tableBucket.getBucket()));
+                                    tableBucket.getPartitionId()));
                     continue;
                 }
             }
--- a/fluss-server/src/main/java/org/apache/fluss/server/coordinator/statemachine/TableBucketStateMachine.java
+++ b/fluss-server/src/main/java/org/apache/fluss/server/coordinator/statemachine/TableBucketStateMachine.java
@@ -31,6 +31,8 @@ import org.apache.fluss.server.entity.RegisterTableBucketLeadAndIsrInfo;
 import org.apache.fluss.server.zk.ZooKeeperClient;
 import org.apache.fluss.server.zk.data.LeaderAndIsr;
 import org.apache.fluss.shaded.guava32.com.google.common.collect.Sets;
+import org.apache.fluss.shaded.zookeeper3.org.apache.zookeeper.KeeperException;
+import org.apache.fluss.utils.ExceptionUtils;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -38,8 +40,10 @@ import org.slf4j.LoggerFactory;
 import javax.annotation.Nullable;
 
 import java.util.ArrayList;
+import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
+import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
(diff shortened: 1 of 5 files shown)


fluss-server/src/main/java/org/apache/fluss/server/coordinator/statemachine/TableBucketStateMachine.java
```diff
@@ -145,6 +155,165 @@ public void handleStateChange(
...
+        for (TableBucket tableBucket : tableBuckets) {
+            coordinatorContext.putBucketStateIfNotExists(
+                    tableBucket, BucketState.NonExistentBucket);
+            if (!checkValidTableBucketStateChange(tableBucket, BucketState.OnlineBucket)) {
+                continue;
+            }
+
+            BucketState currentState = coordinatorContext.getBucketState(tableBucket);
+            if (currentState == BucketState.NewBucket) {
+                bucketsToRetryIndividually.add(tableBucket);
```
LiebingYu, line 195: nit: The current controlled shutdown tests in `TableBucketStateMachineTest` only exercise the batch path with buckets already in `OnlineBucket` state. Consider adding a test with a mixed set — some buckets in `OnlineBucket` and some in `NewBucket`.

LiebingYu (commented):
@swuferhong Thansk for the PR. Only minor comments.

LiebingYu (approved):

fluss-server/src/main/java/org/apache/fluss/server/coordinator/statemachine/TableBucketStateMachine.java
```diff
@@ -145,6 +155,165 @@ public void handleStateChange(
...
+
+    private Set<TableBucket> batchUpdateLeaderAndIsrWithFallback(
+            Map<TableBucket, LeaderAndIsr> leaderAndIsrs) {
+        Set<TableBucket> updatedBuckets = new HashSet<>();
+        if (leaderAndIsrs.isEmpty()) {
+            return updatedBuckets;
+        }
+
+        try {
+            zooKeeperClient.batchUpdateLeaderAndIsr(
```
loserwang1024, line 275: batchUpdateLeaderAndIsr() splits the input into multiple ZooKeeper transactions of at most MAX_BATCH_SIZE, so an exception may be thrown after some earlier transactions have already committed. The current fallback retries every bucket individually, including buckets from transactions that were already confirmed successful.  
Rewriting them should still succeed because the version check is against the coordinator epoch znode, which is not changed by LeaderAndIsr updates. Therefore, this appears correct but results in redundant ZooKeeper writes. Would it be worth propagating the set of definitely committed buckets from batchUpdateLeaderAndIsr() so that only failed or uncertain transactions need to be retried?

fluss-server/src/main/java/org/apache/fluss/server/coordinator/statemachine/TableBucketStateMachine.java
```diff
@@ -145,6 +155,165 @@ public void handleStateChange(
...
+                electionResults.size(),
+                System.currentTimeMillis() - electionStartTimeMs,
+                bucketsToRetryIndividually.size());
+
+        Map<TableBucket, LeaderAndIsr> leaderAndIsrsToUpdate = new HashMap<>();
+        electionResults.forEach(
+                (tableBucket, electionResult) ->
+                        leaderAndIsrsToUpdate.put(tableBucket, electionResult.leaderAndIsr));
+        Set<TableBucket> updatedBuckets =
+                batchUpdateLeaderAndIsrWithFallback(leaderAndIsrsToUpdate);
```
loserwang1024, line 245: If BadVersionException occurs, need to stop here directly,  on need to bucketsToRetryIndividually later.

loserwang1024 (commented):
I left some minor comment

fluss-server/src/main/java/org/apache/fluss/server/coordinator/statemachine/TableBucketStateMachine.java
swuferhong: I think these retries write the same LeaderAndIsr values and validate against the coordinator epoch znode rather than the LeaderAndIsr node version, so they are idempotent and do not cause a correctness issue. The downside is limited to redundant ZooKeeper writes on the exceptional path.

Propagating the definitely committed buckets would require changing the ZooKeeper client API or introducing a partial-progress exception/result type. Given the additional complexity and the goal of keeping this PR minimally invasive, `I suggest leaving it unchanged for now and considering it as a separate follow-up optimization`.

fluss-server/src/main/java/org/apache/fluss/server/coordinator/statemachine/TableBucketStateMachine.java
swuferhong: A simple return may be unsafe because earlier `ZooKeeper` transactions or individual retries may have already succeeded. Returning immediately would skip updating the coordinator cache for those buckets. Since subsequent fenced writes will fail safely, I prefer keeping the current behavior.

fresh-borzoni (approved):
@swuferhong Thank you, LGTM 👍 

UPD:
a couple of inconsistencies, listed them, but other than this - all good.

fluss-server/src/main/java/org/apache/fluss/server/coordinator/statemachine/TableBucketStateMachine.java
fresh-borzoni: Shall we fix this as well?

fresh-borzoni: ditto

PR #3658, opened 2026-07-15
[common] Add config providers to resolve secrets in configuration

### LINKED ISSUE #3659: [common] Support config providers to resolve secrets in configuration
### Search before asking

### Description

Secrets in server.yaml (catalog OAuth creds, SASL passwords, object-store keys) can only be written as literals today. One secret makes the whole file sensitive: it can't be checked in or shipped as a plain K8s ConfigMap, and deployments end up rendering the entire config through init containers (envsubst) just to inject a single value.

Kafka-style config providers looks nice here. A value can be an indirection marker resolved at load time:
```
config.providers: directory
config.providers.directory.param.allowed.paths: /etc/fluss/secrets
datalake.paimon.s3.secret-key: ${directory:/etc/fluss/secrets:s3-secret-key}
```

- ConfigProvider SPI + built-ins: directory (mounted K8s Secret volume), env (secretKeyRef), file (properties)
- opt-in via config.providers, fail-fast on unknown providers/markers, mandatory allowed.paths guard for file-reading providers

### Willingness to contribute

fresh-borzoni:
closes https://github.com/apache/fluss/issues/3659
Today a secret in server.yaml (catalog OAuth creds, SASL passwords) can only be a literal and so the whole file becomes sensitive, and on K8s you end up rendering it through an init container just to inject one value.

This adds a Kafka-style ConfigProvider SPI: a value can be a marker that Fluss dereferences at load time, so the config file only ever holds pointers.
```
config.providers: directory
config.providers.directory.param.allowed.paths: /etc/fluss/secrets
datalake.paimon.s3.secret-key: ${directory:/etc/fluss/secrets:s3-secret-key}
```

### DIFF
--- a/fluss-common/src/main/java/org/apache/fluss/config/provider/AllowedPaths.java
+++ b/fluss-common/src/main/java/org/apache/fluss/config/provider/AllowedPaths.java
@@ -1,0 +1,97 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.fluss.config.provider;
+
+import javax.annotation.Nullable;
+
+import java.io.IOException;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * The mandatory {@code allowed.paths} parameter of file-reading {@link ConfigProvider}s: a
+ * comma-separated list of roots the provider may read from.
+ */
+final class AllowedPaths {
+
+    static final String PARAM = "allowed.paths";
+
+    private final String identifier;
+    private final List<Path> roots;
+
+    private AllowedPaths(String identifier, List<Path> roots) {
+        this.identifier = identifier;
+        this.roots = roots;
+    }
+
+    static AllowedPaths parse(String identifier, @Nullable String value) {
+        List<Path> roots = new ArrayList<>();
+        if (value != null) {
+            for (String root : value.split(",")) {
+                String trimmed = root.trim();
+                if (!trimmed.isEmpty()) {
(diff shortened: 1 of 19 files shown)


fresh-borzoni:
@morazow @affo  What do you think about it?

fluss-common/src/main/java/org/apache/fluss/config/provider/AllowedPaths.java
```diff
@@ -0,0 +1,97 @@
...
+                        + " of the '"
+                        + identifier
+                        + "' config provider.");
+    }
+
+    private static Path canonicalize(Path path) {
+        try {
+            return Paths.get(path.toFile().getCanonicalPath());
+        } catch (IOException e) {
+            return path.toAbsolutePath().normalize();
```
morazow, line 94: Should we fall through here? Throw with error.

```suggestion
 throw new IllegalArgumentException("Cannot canonicalize path '" + path + "' for allowed-path check.", e);
```

fluss-common/src/test/java/org/apache/fluss/config/provider/ConfigProvidersTest.java
```diff
@@ -0,0 +1,262 @@
...
+        Configuration missing = new Configuration();
+        missing.setString(ConfigOptions.CONFIG_PROVIDERS.key(), "env");
+        missing.setString("x", "${env:FLUSS_DEFINITELY_UNSET_VAR_XYZ}");
+        assertThatThrownBy(() -> ConfigProviders.resolve(missing))
+                .isInstanceOf(IllegalConfigurationException.class)
+                .hasMessageContaining("FLUSS_DEFINITELY_UNSET_VAR_XYZ");
+    }
+
+    @Test
+    void testDoubleDollarEscapesLiteralMarker() {
```
morazow, line 156: I think we should have similar test, without config providers defined. It should fail because `$${...}` will not be escaped.

Or in general run the escape processing always even if the config providers are not defined.

fluss-common/src/main/java/org/apache/fluss/config/provider/ConfigProviders.java
```diff
@@ -0,0 +1,211 @@
...
+                                        + "': "
+                                        + e.getMessage(),
+                                e);
+                    }
+                    providers.put(provider.identifier(), provider);
+                }
+            }
+
+            for (String name : requested) {
+                if (!providers.containsKey(name)) {
```
morazow, line 174: I wonder if we should fail for duplicate keys. E.g, for cases if 3rd party jar provider with similar name is included.

morazow (changes requested):
Thanks @fresh-borzoni very good proposal! I have added some suggestions.

In general, this does not apply to the client side yet, right?

fresh-borzoni:
Thanks for the review @morazow, addressed, PTAL
It does apply to the client side: ConnectionFactory resolves markers in both overloads, which also covers the Flink connector:  FlinkCatalog/sources/sinks all connect through it.

Not covered: dynamic cluster configs (markers rejected with an explicit error, so providers are startup-only) and non-Java clients. This is a good follow-ups.

morazow (approved):
Thanks, good from my side 🤝

PR #3671, opened 2026-07-16
[server] skip adding fetchers for followers without a leader

zuston:
### Purpose

Skip adding fetchers for followers without a leader. 
Without this PR, the lots of logs about the `ServerNode -1 is not available in metadata cache` will be shown in the log file


zuston:
minor fix. Could you help take a look? @fresh-borzoni

fresh-borzoni (approved):
@zuston Thank you, makes sense, LGTM 👍

PR #3675, opened 2026-07-17
[flink] Support custom lookup shuffle by bucket for Flink 2.x

### LINKED ISSUE #3670: [flink] Support custom lookup shuffle for Flink lookup join.
### Search before asking

### Motivation

Fluss's Flink lookup join doesn't custom-shuffle the probe stream, so probe rows are distributed to lookup subtasks arbitrarily. This hurts lookup.cache (partial) hit rate and enlarges RPC fan-out, since each subtask may hit any bucket / any TabletServer.
Flink 2.0 added SupportsLookupCustomShuffle ([FLIP-462](https://cwiki.apache.org/confluence/spaces/FLINK/pages/309496789/FLIP-462+Support+Custom+Data+Distribution+for+Input+Stream+of+Lookup+Join)), which lets a source define the probe-stream distribution. We can shuffle by bucket key so each subtask handles a stable set of buckets → better cache locality and lower RPC fan-out.

### Solution

- Implement SupportsLookupCustomShuffle and provide an InputDataPartitioner that maps the lookup key to a Fluss bucket using the existing bucketing logic (consistent with the write path / PrimaryKeyLookuper).
- Only enable when the lookup key covers the bucket key (PK lookup / prefix lookup on bucket key); otherwise fall back to default distribution.
- Note: the interface is Flink 2.0+, so this targets fluss-flink-2.2 only (FlinkTableSource lives in fluss-flink-common which must stay Flink 1.18 compatible).

### Anything else?

_No response_

### Willingness to contribute

naivedogger:
### Purpose

Closes #3670.

Implement Flink's `SupportsLookupCustomShuffle` for Fluss lookup joins on Flink 2.2. With the standard `LOOKUP('shuffle' = 'true')` hint, the probe stream is repartitioned using Fluss bucket routing to improve lookup-cache locality and limit bucket fan-out. Lookup joins without the hint keep their existing distribution.

### Brief change log

- Add a same-FQN `SupportsLookupCustomShuffleAdapter`: the common interface stays compatible with Flink 1.x, while the Flink 2.2 interface extends `SupportsLookupCustomShuffle`. `FlinkTableSource` implements the adapter directly.
- Add `FlussLookupInputPartitioner`, using normalized lookup keys and the same bucket-key encoding and `BucketingFunction` as Fluss clients, including lake-format-specific bucketing.
- Support full primary-key and prefix lookups. Existing `LookupNormalizer` validation ensures that valid normalized lookup keys contain all bucket keys.
- Apply the same routing strategy to partitioned and non-partitioned tables:
  - If the bucket count is a multiple of lookup parallelism, use `bucketId % parallelism`.
  - If parallelism is a larger multiple of the bucket count, assign each bucket an equal-size, disjoint round-robin subtask subset and hash the normalized lookup key within it.
  - Otherwise, use weighted logical slots derived from the GCD of the two counts. This keeps routing deterministic and bucket fan-out bounded, with balanced expected load when bucket traffic and within-bucket key hashes are uniform.
- Include partition keys in the normalized lookup key. A `(partition, bucket)` tablet may be accessed by multiple subtasks in the low-bucket and weighted-slot cases; specialized tablet-affinity optimization is outside this PR.
- Use bucket-key presence to determine shuffle eligibility. An empty partitioner leaves Flink's default distribution unchanged. The resolved bucket count is required when constructing a bucket partitioner, and the constructor checks that it is positive.
- Fluss Catalog supplies the resolved `bucket.num`, including when omitted from the DDL. Raw connector DDL outside the Fluss Catalog is currently unsupported and fails at connector discovery, so this PR does not add a fallback for that path.

### Tests

- `FlussLookupInputPartitionerTest`: 16 passing cases covering frozen bucket vectors across lake formats, composite and reordered keys, prefix lookups, null keys, determinism, partitioned keys, divisible mappings, and weighted routing for both `3 buckets / 10 subtasks` and `10 buckets / 3 subtasks`.
- `Flink22TableSourceITCase#testLookupCustomShuffle`: 4 passing MiniCluster cases with `bucket.num = {1, 3, 8, 10}` and parallelism 4. A downstream forward-connected operator records the runtime subtask, which is checked against the partitioner along with lookup results, repeated-key stability, bucket affinity where applicable, and use of all subtasks.
- `Flink120TableFactoryTest` and `Flink22TableFactoryTest`: 7 passing cases each, including bucket-key eligibility.
- Re-ran the two existing Catalog tests for unsupported temporary and non-Fluss-Catalog tables. Both passed. Separate local lookup SQL planning checks confirmed discovery failure with and without `bucket.num` and the shuffle hint; no new repository tests were added for this unsupported path.
- Targeted tests, Spotless, and Checkstyle passed for the affected modules. The full repository test suite was not rerun locally.

### API and Format

No Fluss public API, storage format, or wire format changes. The Flink 2.2 connector adds the standard `SupportsLookupCustomShuffle` source ability.

### Documentation

Document the shuffle hint, routing strategy, and Catalog-provided bucket count. Remove the unsupported raw-DDL guidance.


zuston (commented):
thanks for your work! @naivedogger . I have a minor question about this. 

>  which improves lookup cache locality

Could you help describe more about this? I haven't seen the similar lookup cache related code.

naivedogger:
Thanks @zuston for your comment!

For some context: I'm trying to help enhance Fluss's Flink connector, and this feature is similar to Paimon's [Large Scale Lookup (Fixed Bucket)](https://paimon.apache.org/docs/master/flink/sql-lookup/#large-scale-lookup-fixed-bucket) — it only kicks in on Flink 2.0+ and when the join keys contain all bucket keys.

**What shuffle lookup buys us:** rows of the same bucket are routed to the same lookup subtask, so each subtask only deals with its own buckets. This way, when partial caching is enabled, each subtask only needs to cache its own slice of the table, giving a potentially better cache hit ratio and less duplicated cached data across subtasks.

Here the cache refers to [Flink's built-in lookup caching](https://nightlies.apache.org/flink/flink-docs-stable/docs/sql/reference/queries/hints/), not a Fluss-specific one — that's probably why it was hard to spot. `FULL` caching isn't supported yet; I'm planning to add it as a follow-up.

zuston:
Thanks for your detailed explanation, @naivedogger. Much appreciated. 🙏

fluss-flink/fluss-flink-2.2/src/main/java/org/apache/fluss/flink/adapter/LookupShuffleSourceAdapter.java
```diff
@@ -0,0 +1,96 @@
...
+ *       partitioner.
+ * </ul>
+ *
+ * <p>The unwrapped (hash-fallback) branch is a defensive safety net rather than a reachable state
+ * for a Fluss table: a Fluss table can only be created through the {@code FlussCatalog}, which
+ * always defaults {@code bucket.num} and derives the bucket key from the primary key, while Flink
+ * rejects creating a Fluss source as a temporary/{@code connector}-based table. It is therefore
+ * covered by adapter unit tests rather than an end-to-end planner test.
+ */
+public class LookupShuffleSourceAdapter {
```
loserwang1024: The current `LookupShuffleSourceAdapter` seems more complicated than necessary. It performs capability gating in the factory, wraps `FlinkTableSource` with an additional subtype, and requires extra copy/state-preservation logic.

Could we follow the existing version-specific superclass adapter pattern instead? For example, let `FlinkTableSource` extend a same-FQN `SupportsLookupCustomShuffleAdapter`.

*  The Flink 2.2 implementation can implement `SupportsLookupCustomShuffle` and provide `getPartitioner()`
```java

public interface SupportsLookupCustomShuffleAdapter extends SupportsLookupCustomShuffle {
    @Override
    default Optional<InputDataPartitioner> getPartitioner() {
        Optional<InputDataPartitionerAdapter> partitionerAdapter = getPartitionerAdapter();
        Optional<InputDataPartitioner> inputDataPartitioner = partitionerAdapter.isPresent() ? Optional.of(partitionerAdapter.get()) : Optional.empty();
        return inputDataPartitioner;
        
    }

    Optional<InputDataPartitionerAdapter> getPartitionerAdapter();

    interface InputDataPartitionerAdapter extends  InputDataPartitioner{}
}

```
* the lower-version implementation :
```java
public interface SupportsLookupCustomShuffleAdapter  {


    Optional<InputDataPartitionerAdapter> getPartitionerAdapter();

    interface InputDataPartitionerAdapter   {
        int partition(RowData joinKeys, int numPartitions);
        default boolean isDeterministic() {
            return true;
        }
    }
}
```

fluss-flink/fluss-flink-2.2/src/test/java/org/apache/fluss/flink/source/FlinkLookupShuffleTableSourceTest.java
```diff
@@ -0,0 +1,247 @@
...
+        RowData second =
+                GenericRowData.of(
+                        2, StringData.fromString("cn"), StringData.fromString("2026-08-20"));
+
+        // The golden bucket vectors pin both ids to Iceberg bucket 4.
+        assertThat(partitioner.partition(second, 4)).isEqualTo(partitioner.partition(first, 4));
+    }
+
+    @Test
+    void testGetPartitionerPresentForPrefixLookup() {
```
loserwang1024: I don't like this test.
1. ItCase need to be add to flink-common module.
2.  An IT case test means end to end, please add  flink sql tests for ITCase(Adding test case in FlinkTableSourceITCase  is enough, referring org.apache.fluss.flink.source.FlinkTableSourceITCase#testLookup1PkTable)
```sql
SELECT /*+ LOOKUP('table'='Customers', 'shuffle'='true') */ o.order_id, o.total, c.country, c.zip
FROM Orders AS o
JOIN Customers FOR SYSTEM_TIME AS OF o.proc_time AS c
ON o.customer_id = [c.id](http://c.id/);
```

3. To verify shuffle is enbaled, check the result of tevn.explainSql("xxx")

loserwang1024 (commented):
I have left some message.

fluss-flink/fluss-flink-2.2/src/main/java/org/apache/fluss/flink/adapter/LookupShuffleSourceAdapter.java
naivedogger: Thanks for the suggestion. The previous implementation incorrectly treated custom shuffle as a capability that needed factory-level gating and fallback handling, which introduced an extra table source wrapper and copy/state-preservation logic. I also overcomplicated the case where the bucket count is smaller than the lookup parallelism.

I have changed it to follow the same-FQN adapter pattern: `FlinkTableSource` directly implements `SupportsLookupCustomShuffleAdapter`, the Flink 2.2 adapter extends `SupportsLookupCustomShuffle`, and the factory returns `FlinkTableSource` directly.

For non-partitioned tables, the revised routing is:
- If `numBuckets >= numPartitions`, route by `bucketId % numPartitions`.
- If `numBuckets < numPartitions`, assign each bucket a disjoint round-robin subset of subtasks and hash the normalized lookup key within that subset. This keeps the same key on a stable subtask while allowing all lookup subtasks to be utilized instead of leaving the extra subtasks idle.

fluss-flink/fluss-flink-2.2/src/test/java/org/apache/fluss/flink/source/FlinkLookupShuffleTableSourceTest.java
naivedogger: Thanks, agreed. The previous tests were mostly Flink 2.2-specific implementation tests and did not sufficiently verify the real SQL runtime behavior.
I removed the redundant tests and moved the IT coverage into FlinkTableSourceITCase. The test now runs a real SQL lookup join on a Flink MiniCluster, checks explainSql, and verifies the actual lookup subtask routing for both bucket.num < parallelism and bucket.num >= parallelism. The Flink 2.2 subclass only enables this test capability.

naivedogger:
Thanks @loserwang1024 for the detailed review! I have pushed a commit addressing your comments. PTAL when convenient!

naivedogger:
Hi @loserwang1024, I have one more question about partitioned tables and would appreciate your opinion.

I checked Paimon's current implementation and aligned this PR with it. Paimon's `BucketIdExtractor` derives only the bucket id, and partitioned and non-partitioned tables share the same `BucketShuffleStrategy`. The revised Fluss implementation now behaves the same way:

- when `bucket.num >= lookup parallelism`, use `bucketId % parallelism`;
- when `bucket.num < lookup parallelism`, assign each bucket a round-robin subtask subset and hash the complete normalized lookup key within that subset. For partitioned tables, that complete key includes the partition key.

This keeps the same lookup key stable and uses all lookup subtasks. The trade-off is that one `(partition, bucket)` tablet may be accessed by multiple subtasks when the bucket count is small. Before this PR, Flink's regular hash shuffle already had this property, so it is not a correctness regression; a more specialized partition/tablet-affinity optimization could be handled separately later.

Do you think keeping partitioned and non-partitioned tables consistent in this PR is reasonable, or do you see a need to preserve `(partition, bucket)` affinity here?

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/source/FlinkTableSource.java
```diff
@@ -532,11 +549,41 @@ public LookupRuntimeProvider getLookupRuntimeProvider(LookupContext context) {
         }
     }
 
+    @Nullable
+    private InputDataPartitionerAdapter createLookupInputPartitioner(
+            LookupNormalizer lookupNormalizer) {
+        if (numBuckets == null) {
```
loserwang1024: Could we use the presence of bucket keys, rather than numBuckets == null, to determine whether bucket shuffle should be applied?

At this point, every Fluss table already has a resolved bucket count, and `FlinkConversions#toFlinkTable `always exposes it through `bucket.num`. Bucket keys are what distinguish hash distribution from random/sticky distribution.

Therefore, numBuckets should be modeled as a non-null int and used only to calculate the bucket ID. The bucket-shuffle eligibility check should be based on whether bucket keys are present.

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/source/lookup/FlussLookupInputPartitioner.java
```diff
@@ -0,0 +1,140 @@
...
+            LookupNormalizer normalizer,
+            RowType keyFlinkRowType,
+            List<String> bucketKeyNames,
+            @Nullable DataLakeFormat lakeFormat,
+            int numBuckets) {
+        this.normalizer = normalizer;
+        this.keyFlinkRowType = keyFlinkRowType;
+        this.bucketKeyNames = bucketKeyNames;
+        this.lakeFormat = lakeFormat;
+        checkArgument(numBuckets > 0, "numBuckets must be positive, but was %s.", numBuckets);
```
loserwang1024, line 97: add same check as org.apache.fluss.flink.source.lookup.LookupNormalizer#validateAndCreateLookupNormalizer.

loserwang1024:
> Hi @loserwang1024, I have one more question about partitioned tables and would appreciate your opinion.
> 
> I checked Paimon's current implementation and aligned this PR with it. Paimon's `BucketIdExtractor` derives only the bucket id, and partitioned and non-partitioned tables share the same `BucketShuffleStrategy`. The revised Fluss implementation now behaves the same way:
> 
> * when `bucket.num >= lookup parallelism`, use `bucketId % parallelism`;
> * when `bucket.num < lookup parallelism`, assign each bucket a round-robin subtask subset and hash the complete normalized lookup key within that subset. For partitioned tables, that complete key includes the partition key.
> 
> This keeps the same lookup key stable and uses all lookup subtasks. The trade-off is that one `(partition, bucket)` tablet may be accessed by multiple subtasks when the bucket count is small. Before this PR, Flink's regular hash shuffle already had this property, so it is not a correctness regression; a more specialized partition/tablet-affinity optimization could be handled separately later.
> 
> Do you think keeping partitioned and non-partitioned tables consistent in this PR is reasonable, or do you see a need to preserve `(partition, bucket)` affinity here?

The current assignment makes every downstream subtask reachable, but does not utilize them evenly when the number of subtasks is not divisible by the number of buckets. For example, with 3 buckets and 10 subtasks, the buckets are assigned 4, 3, and 3 candidate subtasks respectively. Assuming similar traffic across buckets, subtasks assigned to the first bucket receive only one quarter of that bucket’s traffic, while the others receive one third.

Consider using a weighted logical-slot assignment similar to WeightedBucketIdAssignment. For example, 3 buckets and 10 subtasks can be represented by 30 logical slots, with 10 slots per bucket and 3 slots per physical subtask. A deterministic hash of the normalized lookup key can then select a logical slot within its bucket. This would preserve deterministic routing and bounded bucket fan-out while distributing the expected load evenly across all downstream subtasks.

```mermaid
flowchart TB
    A["Assumption: 3 buckets have equal traffic<br/>10 downstream sink subtasks"]

    subgraph CURRENT["Current algorithm — disjoint candidate sets"]
        direction TB
        C0["Bucket 0<br/>1/3 traffic"] --> C0S["Sinks 0, 3, 6, 9<br/>each receives 1/12 = 8.3%"]
        C1["Bucket 1<br/>1/3 traffic"] --> C1S["Sinks 1, 4, 7<br/>each receives 1/9 = 11.1%"]
        C2["Bucket 2<br/>1/3 traffic"] --> C2S["Sinks 2, 5, 8<br/>each receives 1/9 = 11.1%"]
        CI["Result: all sinks are reachable,<br/>but expected load is 8.3% or 11.1%"]
    end

    subgraph WEIGHTED["Proposed algorithm — 30 logical slots"]
        direction TB
        L["LCM(3 buckets, 10 sinks) = 30 slots<br/>10 slots per bucket · 3 slots per sink"]

        W0["Bucket 0 · slots 0–9"] --> W0S["Sink 0 ×3 · Sink 1 ×3<br/>Sink 2 ×3 · Sink 3 ×1"]
        W1["Bucket 1 · slots 10–19"] --> W1S["Sink 3 ×2 · Sink 4 ×3<br/>Sink 5 ×3 · Sink 6 ×2"]
        W2["Bucket 2 · slots 20–29"] --> W2S["Sink 6 ×1 · Sink 7 ×3<br/>Sink 8 ×3 · Sink 9 ×3"]

        WI["Each sink owns exactly 3 slots<br/>Expected load = 3/30 = 10%"]
    end

    A --> CURRENT
    A --> WEIGHTED
```

loserwang1024 (commented):
some advice

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/source/FlinkTableSource.java
naivedogger: Agreed. The eligibility check is now based on whether bucket keys are present. Without bucket keys, the source returns an empty partitioner and Flink uses its default distribution. bucket.num is only read as a local primitive int when creating the bucket partitioner, so no factory or constructor changes are needed.

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/source/lookup/FlussLookupInputPartitioner.java
naivedogger: Added. The partitioner now validates that bucket keys are non-empty and are a subset of the normalized lookup-key fields, consistent with LookupNormalizer.

naivedogger:
Thanks @loserwang1024 for the great suggestions! I have addressed the comments and simplified the implementation:
- Bucket custom shuffle is enabled based on the presence of bucket keys; tables without bucket keys fall back to Flink's default distribution.
- Bucket keys are validated against the normalized lookup-key fields.
- For shuffle routing, existing mappings are preserved when the bucket and subtask counts evenly divide each other. Otherwise, weighted logical slots derived from their GCD are used to balance the expected subtask load while keeping routing deterministic and bucket fan-out bounded. The same strategy applies to partitioned and non-partitioned tables.
- Added unit coverage for both `3 buckets / 10 subtasks` and `10 buckets / 3 subtasks`, and MiniCluster IT coverage for `bucket.num = {1, 3, 8, 10}`. All related tests pass.

loserwang1024 (approved):
LGTM ， @leonardBang , would you like to give a final check.

leonardBang:
Thanks for the thorough work here — the routing algorithm (divisible / round-robin / LCM weighted-slot branches), the frozen golden vectors across lake formats, and the MiniCluster ITCase that compares the real runtime subtask against the production partitioner are all solid. One thing I'd like you to double-check before merging.

**`createLookupInputPartitioner` runs for every lookup join and hard-fails when `bucket.num` can't be resolved, even when the `shuffle=true` hint is absent.**

In `FlinkTableSource#getLookupRuntimeProvider`, `this.lookupInputPartitioner = createLookupInputPartitioner(lookupNormalizer)` is invoked unconditionally. As soon as `bucketKeyIndexes.length > 0`, it calls `checkNotNull(tableOptions.get(BUCKET_NUMBER), ...)`, and the `FlussLookupInputPartitioner` constructor additionally does `checkArgument(bucketKeys ⊆ lookupKeys)`.

For tables resolved through the Fluss Catalog this is safe, since `bucket.num` is always injected (`FlinkConversions` puts `tableInfo.getNumBuckets()` into the options). But for a Fluss table declared via raw DDL (`connector = 'fluss'` in a non-Fluss catalog) that sets `bucket.key` without `bucket.num`, `tableOptions` won't contain `BUCKET_NUMBER`, so `checkNotNull` throws and the lookup job fails at plan/build time — even though the user never requested custom shuffle. This seems to diverge from the stated intent of "fail explicitly **when custom shuffle is requested**": today it effectively fails for *any* lookup on a bucket-keyed table whose `bucket.num` isn't statically present.

@naivedogger Could you confirm whether that DDL path is actually reachable/supported? If it is, I'd suggest making it fail-soft — return `null` (i.e. `Optional.empty()`) when `bucket.num` is missing or the bucket keys aren't a subset of the lookup keys — so the optimization degrades to Flink's default distribution instead of regressing an otherwise-working lookup join. Alternatively, gate the construction on `LookupContext#preferCustomShuffle()` (available on Flink 2.2) and keep the fail-fast only for the case where the shuffle was actually requested.

Note this is about robustness, not join correctness: if a DDL-declared `bucket.num` differs from the real server-side bucket count, only the locality optimization degrades — each subtask's lookuper still routes to the correct bucket.

naivedogger:
Thanks @leonardBang for pointing this out. I confirmed that the raw DDL path you mentioned is reachable and updated the implementation accordingly. When `bucket.num` is missing or non-positive, the source now returns an empty partitioner, allowing regular lookup joins to continue and custom shuffle to fall back to Flink’s default distribution.
I kept the bucket-key/lookup-key validation unchanged because valid full-PK and prefix lookups are already normalized to include all bucket keys. I also added regression coverage for the fallback behavior on Flink 1.20 and 2.2 and clarified it in the documentation.

naivedogger:
Thanks @leonardBang, and sorry for the confusion in my earlier reply. Raw connector DDL outside the Fluss Catalog currently fails at connector discovery, before lookup setup; I verified this with lookup SQL planning checks both with and without `bucket.num` and the shuffle hint. The supported Fluss Catalog path supplies the resolved bucket count even when it is omitted from the DDL. I've therefore restored the required bucket-count check and removed the fallback test and documentation for the unsupported path in 3438ceceb. 
For the bucket-key subset concern, the existing `LookupNormalizer` rejects lookups missing required bucket keys before the partitioner is constructed. This validation predates this PR, so keeping the partitioner's check does not reject an otherwise-supported lookup. I also re-ran the existing missing-key SQL tests and checked full-PK and prefix normalization with projection. Could you take another look?

leonardBang:
Thanks @naivedogger for the update, LGTM. wait the CI green...

leonardBang (approved):

Issue #3656, opened 2026-07-15
Support batch read for PK tables based on SST files


loserwang1024:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/fluss/issues) and found nothing similar.

### Description

The parent issue proposes an SST-file-based batch read approach for PK tables. This issue implements the core `KvSnapshotAndLogBatchScanner`, following the pattern of [`LakeSnapshotAndLogSplitScanner`](https://github.com/apache/fluss/blob/main/fluss-client/src/main/java/org/apache/fluss/client/table/scanner/batch/LakeSnapshotAndLogSplitScanner.java) but replacing the lakehouse snapshot data source with KV snapshot SST files.

### Willingness to contribute

- [x] I'm willing to submit a PR!

PR #3677, opened 2026-07-17
[flink] Flink batch read based on SST merge.

loserwang1024:
### Purpose

Linked issue: close http://github.com/apache/fluss/issues/3656

### DIFF
--- a/fluss-client/src/main/java/org/apache/fluss/client/table/scanner/batch/KvSnapshotAndLogBatchScanner.java
+++ b/fluss-client/src/main/java/org/apache/fluss/client/table/scanner/batch/KvSnapshotAndLogBatchScanner.java
@@ -1,0 +1,325 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.fluss.client.table.scanner.batch;
+
+import org.apache.fluss.annotation.Internal;
+import org.apache.fluss.client.table.Table;
+import org.apache.fluss.client.table.scanner.ScanRecord;
+import org.apache.fluss.client.table.scanner.SortMergeReader;
+import org.apache.fluss.client.table.scanner.log.LogScanner;
+import org.apache.fluss.client.table.scanner.log.ScanRecords;
+import org.apache.fluss.memory.MemorySegment;
+import org.apache.fluss.metadata.Schema;
+import org.apache.fluss.metadata.TableBucket;
+import org.apache.fluss.metadata.TableInfo;
+import org.apache.fluss.record.ChangeType;
+import org.apache.fluss.record.LogRecord;
+import org.apache.fluss.row.InternalRow;
+import org.apache.fluss.row.KeyValueRow;
+import org.apache.fluss.row.encode.KeyEncoder;
+import org.apache.fluss.types.RowType;
+import org.apache.fluss.utils.CloseableIterator;
+import org.apache.fluss.utils.IOUtils;
+
+import javax.annotation.Nullable;
+
+import java.io.IOException;
+import java.time.Duration;
+import java.util.Arrays;
+import java.util.Comparator;
+import java.util.List;
+import java.util.Map;
+import java.util.TreeMap;
+import java.util.stream.Collectors;
+import java.util.stream.IntStream;
+
+import static org.apache.fluss.utils.Preconditions.checkArgument;
+
+/**
+ * A bounded scanner that merges an immutable KV snapshot with a fixed primary-key changelog range.
+ */
+@Internal
+public class KvSnapshotAndLogBatchScanner implements BatchScanner {
+
+    private final TableBucket tableBucket;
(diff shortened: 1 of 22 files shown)


loserwang1024:
@fresh-borzoni @leonardBang @beryllw , CC

fluss-client/src/main/java/org/apache/fluss/client/table/scanner/batch/KvSnapshotAndLogBatchScanner.java
```diff
@@ -0,0 +1,325 @@
...
+            return;
+        }
+        closed = true;
+        IOUtils.closeQuietly(sortMergeIterator);
+        IOUtils.closeQuietly(snapshotRecordIterator);
+        IOUtils.closeQuietly(snapshotScanner);
+        IOUtils.closeQuietly(logScanner);
+    }
+
+    private static ProjectionPlan createProjectionPlan(
```
naivedogger, line 209: nit: this is logically identical to [LakeSnapshotAndLogSplitScanner#getNeedProjectFields](https://github.com/apache/fluss/blob/1496616c5212d590dd20f87e92c43d86b07751ca/fluss-client/src/main/java/org/apache/fluss/client/table/scanner/batch/LakeSnapshotAndLogSplitScanner.java#L105-L142)(pk-column completion + adjustProjectedFields mapping, just packaged differently). Maybe worth extracting into a shared helper in this package so future fixes don't need to be applied twice?

fluss-client/src/main/java/org/apache/fluss/client/table/scanner/batch/KvSnapshotAndLogBatchScanner.java
```diff
@@ -0,0 +1,325 @@
...
+            return true;
+        }
+
+        if (snapshotScanner == null) {
+            snapshotRecordIterator = CloseableIterator.emptyIterator();
+            snapshotIteratorInitialized = true;
+            return true;
+        }
+
+        CloseableIterator<InternalRow> snapshotRows = snapshotScanner.pollBatch(timeout);
```
naivedogger: `initializeSnapshotIterator` calls `snapshotScanner.pollBatch()` only once and treats the returned iterator as the *complete* snapshot stream. This assumes a single `pollBatch` returns the whole snapshot, which holds for `KvSnapshotBatchScanner` today but violates the poll-until-null contract of `BatchScanner` — a chunked implementation would silently lose data. Could we make `SnapshotRecordIterator` lazily poll until `pollBatch` returns null? If the underlying scanner ever becomes chunked (like `KvBatchScanner`), this would silently drop snapshot data.

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/source/split/HybridSnapshotLogSplit.java
```diff
@@ -22,6 +22,9 @@
 import javax.annotation.Nullable;
 
 import java.util.Objects;
+import java.util.Optional;
+
+import static org.apache.fluss.utils.Preconditions.checkArgument;
 
 /**
  * The hybrid split for first reading the snapshot files and then switch to read the cdc log from a
```
naivedogger: nit: the class javadoc still only describes the streaming two-phase behavior ("first reading the snapshot files and then switch to read the cdc log"). Since this split now also models bounded batch reads (merged snapshot + bounded log, possibly with no snapshot at all via `NO_SNAPSHOT_ID`), could we update the javadoc and add a doc for the `NO_SNAPSHOT_ID` semantics?

naivedogger (commented):
Thanks @loserwang1024 for the PR! Overall looks good to me — left a few comments, PTAL.

naivedogger (approved):
LGTM

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/source/enumerator/FlinkSourceEnumerator.java
```diff
@@ -589,37 +592,29 @@ private void startInBatchMode() {
...
                         }
                         return splits;
                     },
                     this::handleSplitsAdd);
         } else {
-            throw new UnsupportedOperationException(
-                    String.format(
-                            "Batch only supports when table option '%s' is set to true.",
-                            ConfigOptions.TABLE_DATALAKE_ENABLED));
+            context.callAsync(flussOnlyBatchSplitGenerator::generate, this::handleSplitsAdd);
```
leonardBang, line 601: Would it make sense to exclude buckets that already have restored unassigned splits from batch split generation?

 `start()` first restores `unassignedSplits` into `pendingSplitAssignment`, but then unconditionally calls `startInBatchMode()`. The generator's `tableBucketSkipper` only checks `assignedTableBuckets`. Therefore, if split generation completes before the restored split is assigned to a reader, the bucket is not skipped and a second split is generated for the same bucket. Both splits are subsequently added and assigned, which may produce duplicate rows after recovery.

 I reproduced this by restoring one unassigned bounded PK split, running the one-time generator before registering readers, and observed two splits for the same bucket.

 Could we include the pending/unassigned buckets in the skipper? It would also be helpful to add a recovery test that checkpoints a bounded split before assignment and verifies that exactly one split is assigned per bucket.

leonardBang (commented):
Thanks @loserwang1024 for the contribution and @naivedogger for the review, the PR is in a good scope, I only left one comment

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/source/enumerator/FlinkSourceEnumerator.java
loserwang1024: >Would it make sense to exclude buckets that already have restored unassigned splits from batch split generation

why need to considered  restored unassigned splits in batch read?

leonardBang: Correction to my previous comment: After re-checking Flink's execution-mode in AI-generated test, the RuntimeExecutionMode.BATCH is missed, therefore, the restore scenario I described is not reachable through the normal batch SQL path covered by this PR.

leonardBang (approved):
+1 from my side.

PR #3683, opened 2026-07-17
[tiering] Use exponential delay restart strategy for Flink tiering job

### LINKED ISSUE #3681: [tiering] Flink tiering job fails permanently after transient errors
### Search before asking

### Fluss version

0.9.0 (latest release)

### Please describe the bug 🐞

The Fluss Lake Tiering job does not rely on Flink checkpoints so checkpointing is disabled by default.
However, when checkpointing is disabled and no restart strategy is explicitly configured, Flink uses the no-restart strategy by default.

Currently, FlussLakeTiering configures the full-restart failover strategy but does not configure a job restart strategy. When a transient exception occurs during tiering, the Flink job transitions directly to the FAILED state and is not restarted. As a result, the tiering service stops permanently and requires manual job resubmission.

The existing failover integration test enables checkpointing and therefore does not reproduce the actual production configuration where checkpointing is disabled, leaving this issue uncovered.

### Solution

_No response_

Shawn-Hx:
Configure the stateless tiering job to retry indefinitely when checkpointing is disabled, preventing transient failures from terminating the job permanently.

Move the failover integration test to fluss-flink-tiering so that it covers the production job configuration.

### Purpose

Linked issue: close #3681 

### Tests

`TieringFailoverITCase`
  - Verifies that checkpointing is disabled.
  - Injects two consecutive lake write failures.
  - Verifies that the job recovers and continues tiering data correctly.

### API and Format

### Documentation

### DIFF
--- a/fluss-flink/fluss-flink-tiering/pom.xml
+++ b/fluss-flink/fluss-flink-tiering/pom.xml
@@ -117,6 +117,11 @@
             <scope>test</scope>
         </dependency>
 
+        <!--
+         Required at test runtime because TieringSource loads FlinkConnectorOptionsUtils, which
+         references ValidationException from flink-table-common. fluss-flink-common declares this
+         dependency as provided, so it is not transitively available to this module's tests.
+        -->
         <dependency>
             <groupId>org.apache.flink</groupId>
             <artifactId>flink-table-common</artifactId>
--- a/fluss-flink/fluss-flink-tiering/pom.xml
+++ b/fluss-flink/fluss-flink-tiering/pom.xml
@@ -57,6 +57,72 @@
             <version>${flink.minor.version}</version>
             <scope>provided</scope>
         </dependency>
+
+        <!-- test dependencies -->
+        <dependency>
+            <groupId>org.apache.fluss</groupId>
+            <artifactId>fluss-flink-common</artifactId>
+            <version>${project.version}</version>
+            <scope>test</scope>
+            <type>test-jar</type>
+        </dependency>
+
+        <dependency>
+            <groupId>org.apache.fluss</groupId>
+            <artifactId>fluss-server</artifactId>
+            <version>${project.version}</version>
+            <scope>test</scope>
+        </dependency>
+
+        <dependency>
+            <groupId>org.apache.fluss</groupId>
+            <artifactId>fluss-server</artifactId>
+            <version>${project.version}</version>
+            <scope>test</scope>
+            <type>test-jar</type>
+        </dependency>
+
+        <dependency>
+            <groupId>org.apache.fluss</groupId>
+            <artifactId>fluss-common</artifactId>
+            <version>${project.version}</version>
+            <scope>test</scope>
+            <type>test-jar</type>
+        </dependency>
+
+        <dependency>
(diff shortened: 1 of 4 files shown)


fluss-flink/fluss-flink-tiering/src/main/java/org/apache/fluss/flink/tiering/FlussLakeTiering.java
```diff
@@ -96,6 +97,13 @@ public FlussLakeTiering(String[] args) {
         org.apache.flink.configuration.Configuration flinkConfig =
                 new org.apache.flink.configuration.Configuration();
         flinkConfig.set(JobManagerOptions.EXECUTION_FAILOVER_STRATEGY, FULL_RESTART_STRATEGY_NAME);
+        // Configure restart strategy: the tiering job is completely stateless (offsets are
+        // persisted in Fluss Server's LakeSnapshot, not in Flink checkpoint state), so it is
+        // safe to restart indefinitely on failure. Without this, Flink defaults to "no-restart"
+        // when checkpoint is not enabled, causing the job to fail permanently on any exception.
+        flinkConfig.set(RestartStrategyOptions.RESTART_STRATEGY, "fixed-delay");
```
luoyuxia: Before this change, the failover test enabled checkpointing and therefore inherited Flink 1.20’s default exponential-delay restart policy (1s initial backoff, 1.5 multiplier, and 1min maximum backoff). Selecting fixed-delay here changes the behavior to retry every 1s indefinitely. During a persistent lake/storage failure, this can continuously hit the external system and flood logs instead of backing off. Is this fixed retry cadence intentional? If the goal is to retain the retry behavior previously exercised by the checkpoint-enabled test while enabling restarts without checkpoints, could we explicitly select `exponential-delay` instead (whose default attempt count is effectively infinite), or otherwise document and explicitly configure the fixed-delay policy?

fluss-flink/fluss-flink-tiering/pom.xml
```diff
@@ -57,6 +57,72 @@
...
+        </dependency>
+
+        <dependency>
+            <groupId>org.apache.flink</groupId>
+            <artifactId>flink-connector-base</artifactId>
+            <version>${flink.minor.version}</version>
+            <scope>test</scope>
+        </dependency>
+
+        <dependency>
```
luoyuxia, line 125: Is this dependency needed for the moved failover IT? I could not find any `org.apache.flink.table` usage in `TieringFailoverITCase`, `FlinkTieringTestBase`, or the tiering execution path exercised by this test. The other newly added dependencies map to concrete test classes or runtime requirements, while this one appears unnecessary. Could we remove it to keep the test classpath minimal, or document which class requires it?

fluss-flink/fluss-flink-tiering/src/main/java/org/apache/fluss/flink/tiering/FlussLakeTiering.java
Shawn-Hx: You are right. exponential-delay is more appropriate. Fixed.

luoyuxia (approved):
@Shawn-Hx Thanks. LGTM

PR #3684, opened 2026-07-17
[hotfix][test] Fix DynamicConfigManager constructor invocation

fresh-borzoni:
#3658 crossed with #3645, which removed the isCoordinator arg from DynamicConfigManager — two tests added in #3658 still call the old 3-arg constructor, breaking test compile on main.


morazow (approved):
Thanks 🚀

PR #3685, opened 2026-07-17
[common][helm] helm support for config providers

### LINKED ISSUE #3686: [helm] Declarative secrets support for config provider markers
### Search before asking

### Description

#3658 added config providers, so a server.yaml value can be a ${directory:...} marker instead of a literal secret. But on Kubernetes, actually using that with our chart is still manual: you have to wire extraVolumes, mount the Secret, and hand-write the config.providers lines, including allowed.paths, which is easy to get wrong and is the security guard.

The chart should make this declarative: say which Secrets you want mounted (or injected as env), and get the provider config generated for you, with allowed.paths and the env allowlist derived from what you declared, so the guards are correct by construction and the rendered ConfigMap never contains secret material.

### Willingness to contribute

fresh-borzoni:
closes https://github.com/apache/fluss/issues/3686

On the helm side, secrets become declarative:
```
 secrets:
   mounts:
     - name: paimon-creds
       secretName: fluss-paimon-creds

 configurationOverrides:
   datalake.paimon.s3.secret-key: ${directory:/etc/fluss/secrets/paimon-creds:secret-key}
```

The chart generates the config.providers config with allowed.paths/env allowlist derived from the declared entries, so the rendered server.yaml stays a plain ConfigMap with no secret material. 

Plus some follow-ups from the #3658 review threads:
- allowed.paths canonicalization now fails closed
- duplicate provider identifiers on the classpath fail fast
- test + docs pinning the gated $${ escape semantics

### DIFF
--- a/helm/templates/_secrets.tpl
+++ b/helm/templates/_secrets.tpl
@@ -1,0 +1,133 @@
+{{/*
+Licensed to the Apache Software Foundation (ASF) under one
+or more contributor license agreements.  See the NOTICE file
+distributed with this work for additional information
+regarding copyright ownership.  The ASF licenses this file
+to you under the Apache License, Version 2.0 (the
+"License"); you may not use this file except in compliance
+with the License.  You may obtain a copy of the License at
+
+     http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/}}
+
+{{/*
+Whether any secret mounts or env entries are configured.
+Usage:
+  include "fluss.secrets.enabled" .
+*/}}
+{{- define "fluss.secrets.enabled" -}}
+{{- if or .Values.secrets.mounts .Values.secrets.env -}}
+true
+{{- end -}}
+{{- end -}}
+
+{{/*
+Base directory under which each secret mount gets its own subdirectory.
+*/}}
+{{- define "fluss.secrets.basePath" -}}
+/etc/fluss/secrets
+{{- end -}}
+
+{{/*
+Pod volumes for the configured secret mounts.
+Usage:
+  include "fluss.secrets.volumes" .
+*/}}
+{{- define "fluss.secrets.volumes" -}}
+{{- range .Values.secrets.mounts }}
+- name: secret-{{ .name }}
+  secret:
+    secretName: {{ .secretName }}
+{{- end }}
+{{- end -}}
+
+{{/*
+Read-only container volumeMounts, one subdirectory per secret mount.
+Usage:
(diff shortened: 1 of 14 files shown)


fresh-borzoni:
@morazow @affo A bit more helm conveniences :) PTAL 🙏

website/docs/install-deploy/deploying-with-helm.md
```diff
@@ -383,6 +383,45 @@ The same pattern works with Sealed Secrets, HashiCorp Vault Agent Injector (prod
...
+[config provider marker](../security/secrets.md) instead of a literal, so the rendered
+`server.yaml` ConfigMap never contains secret material. The chart generates the matching
+`config.providers` settings, restricted to the declared mounts and env names.
+
+```bash
+kubectl create secret generic fluss-paimon-creds \
+  --from-literal=access-key=... --from-literal=secret-key=...
+```
+
+```yaml
```
affo, line 404: can we also add an example for `env`?

helm/templates/_secrets.tpl
```diff
@@ -0,0 +1,133 @@
...
+{{- if or .Values.secrets.mounts .Values.secrets.env -}}
+true
+{{- end -}}
+{{- end -}}
+
+{{/*
+Base directory under which each secret mount gets its own subdirectory.
+*/}}
+{{- define "fluss.secrets.basePath" -}}
+/etc/fluss/secrets
```
affo: do you think it is better to make this overridable by the user in case?

helm/values.yaml
```diff
@@ -146,6 +146,18 @@ listeners:
   client:
     port: 9124
 
+# Secrets referenced from configurationOverrides via config provider markers:
+# mounts are read with ${directory:/etc/fluss/secrets/<name>:<key>}, env entries
+# with ${env:<NAME>}. The chart generates the matching config.providers lines.
+secrets:
```
affo, line 152: I think we can add a `basePath` here, so that the user can override, and it defaults to `etc/fluss/secrets`

helm/values.yaml
```diff
@@ -146,6 +146,18 @@ listeners:
...
+  # - name: paimon-creds
```
affo, line 154: do you think in this case it would be better to rename `name` to something similar to `subDir`?

I think this would be more explicit for the user to understand that to use it it should be `${directory:<basePath>/<SubDir>:key}`.

Just a nit, `name` still fits well, it took me some time to understand :)

affo (commented):
@fresh-borzoni thank you so much for this contrib!

Most of my comments are nits, the contribution looks already very good

helm/values.yaml
fresh-borzoni: Kept name as it also names the volume (secret-<name>), where subDir would read oddly, and it keeps the mounts/env lists consistent with the usual k8s list shape. Added comment to help reasoning.

fresh-borzoni:
@affo Thank you for the review, addressed, PTAL 🙏

affo (approved):
Perfect, thank you Anton!

PR #3687, opened 2026-07-17
[helm] Support statefulset and service annotations

### LINKED ISSUE #3688: [helm] StatefulSets and main services should support annotations
### Search before asking

### Description

The chart is inconsistent about annotation hooks: the metrics services take them (metrics.prometheus.service.annotations), pods take them (podAnnotations), but the statefulset metadata and the main headless services take none at all.

Workload and service annotations are where a lot of standard tooling hooks in - Reloader for restart-on-secret-rotation, Argo CD sync-waves, external-dns, org-mandated ownership annotations enforced by policy engines. Today the only way to attach any of these is patching the rendered output out-of-band, which defeats values.yaml as the single source of truth.

Add coordinator.annotations / tablet.annotations for the statefulsets and the same under service: for the headless services, matching the existing podAnnotations style.

### Willingness to contribute

fresh-borzoni:
closes https://github.com/apache/fluss/issues/3688

The metrics services already accept annotations, but the statefulsets and the main headless services don't. This adds coordinator.annotations / tablet.annotations for the statefulset metadata, and coordinator.service.annotations / tablet.service.annotations for the services, following the existing podAnnotations style.

Workload-level annotations are where tooling like Reloader (secret.reloader.stakater.com/reload) or Argo CD sync-waves hook in, without them you're stuck patching the chart output out-of-band.

Pure passthrough, no change to rendered output unless the values are set.

### DIFF
--- a/helm/templates/sts-coordinator.yaml
+++ b/helm/templates/sts-coordinator.yaml
@@ -22,6 +22,10 @@ metadata:
   name: coordinator-server
   labels:
   {{- include "fluss.labels" . | nindent 4 }}
+  {{- with .Values.coordinator.annotations }}
+  annotations:
+    {{- toYaml . | nindent 4 }}
+  {{- end }}
 spec:
   serviceName: coordinator-server-hs
   replicas: {{ .Values.coordinator.numberOfReplicas }}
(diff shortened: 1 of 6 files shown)


fresh-borzoni:
@affo @morazow PTAL 🙏

morazow (approved):
Looks good 🤝

affo:
Looks wonderful! Thanks for introducing this new level of flexibility.

It will be probably renamed to `extraAnnotations` in the future, but for now we don't have "base annotations", so it makes sense that it is just `annotations`.

@fresh-borzoni I don't see any update in the website documentation. Would you be willing to take care to add this in a separate PR?

PR #3699, opened 2026-07-20
[CVE][filesystem] Align Hadoop dependencies to 3.4.3

### LINKED ISSUE #3682: [CVE/filesystem] Align Hadoop dependencies to 3.4.3
### Search before asking

### Motivation

Fluss currently uses different Hadoop versions:

- fluss.hadoop.version: 3.4.0
- fs.hadoopshaded.version: 3.3.4

This issue proposes aligning the Hadoop dependencies to 3.4.3.

### Solution

- Upgrade fluss.hadoop.version from 3.4.0 to 3.4.3
- Upgrade fs.hadoopshaded.version from 3.3.4 to 3.4.3
- Migrate fluss-fs-s3 to AWS SDK v2
- Exclude software.amazon.awssdk:bundle and include required SDK modules
- Review filesystem JAR size and NOTICE/LICENSE files
- Verify S3, HDFS, OSS, GS and Azure filesystem plugins

## Related

- #1210
- #1352
- #2511
- #2674
- apache/flink#27026

### Willingness to contribute

litiliu:
### Purpose

Linked issue: close #3682

Fluss currently uses inconsistent Hadoop versions (`fluss.hadoop.version=3.4.0`, `fs.hadoopshaded.version=3.3.4`). This PR aligns all filesystem plugins to Hadoop 3.4.3 and migrates `fluss-fs-s3` from the end-of-life AWS SDK for Java v1 to the modular AWS SDK v2, while preserving authentication, STS, KMS and S3 transfer-manager support.

### Brief change log

- Bump `fluss.hadoop.version` 3.4.0 → 3.4.3 and `fs.hadoopshaded.version` 3.3.4 → 3.4.3.
- Migrate `fluss-fs-s3` to AWS SDK v2 (`software.amazon.awssdk`): STS / KMS / auth / s3 / s3-transfer-manager as individual modules; exclude `software.amazon.awssdk:bundle`; drop the unused DynamoDB dependency.
- Rewrite `S3DelegationTokenProvider` and `DynamicTemporaryAWSCredentialsProvider` against SDK v2 APIs; remove the copied `XmlResponsesSaxParser` and its checkstyle/coverage suppressions.
- Exclude netty / dropwizard-metrics / jettison that are not used by `hadoop-common`; refresh `NOTICE`/`LICENSE` for all filesystem modules (including COS); add the reactive-streams license.

#### Artifact size review

| Artifact | Before | After | Change |
|---|---|---|---|
| Hadoop shaded | 22.20 MB | 24.13 MB | +1.94 MB (+8.72%) |
| HDFS | 34.66 MB | 35.29 MB | +0.63 MB (+1.82%) |
| S3 | 31.39 MB | 44.52 MB | +13.14 MB (+41.86%) |
| OSS | 38.35 MB | 28.40 MB | −9.95 MB (−25.95%) |
| COS | 41.63 MB | 39.84 MB | −1.79 MB (−4.30%) |
| GS | 65.78 MB | 67.71 MB | +1.94 MB (+2.94%) |
| Azure | 38.35 MB | 30.80 MB | −7.54 MB (−19.68%) |
| **Total** | **272.35 MB** | **270.70 MB** | **−1.65 MB (−0.61%)** |

The S3 increase is primarily the AWS SDK V2 S3 model and its asynchronous Netty transport. The `s3-transfer-manager` classes themselves account for only about 0.20 MB compressed, but Hadoop 3.4.3 S3A invokes `S3TransferManager.copy()` on copy/rename paths, so its module and transport cannot be removed safely. The resulting S3 artifact contains STS, KMS, Transfer Manager, and the async transport, and contains neither DynamoDB nor the complete AWS SDK bundle.

### Tests

- Updated `S3DelegationTokenProviderTest` for the SDK v2 credential APIs.
- Existing filesystem IT cases (S3 / OSS / COS / GS / Azure / HDFS).
- `mvn clean verify`, spotless and RAT.

### API and Format

No public API or storage-format change. Note: `s3.aws.credentials.provider` (`fs.s3a.aws.credentials.provider`) now requires an AWS SDK v2 `software.amazon.awssdk.auth.credentials.AwsCredentialsProvider` implementation instead of an SDK v1 provider (documented).

### Documentation

Updated the S3 and HDFS filesystem docs for AWS SDK v2 / Hadoop 3.4.3.

### Generative AI disclosure

Generated-by: OpenAI Codex following the guidelines (https://github.com/apache/fluss/blob/main/AGENTS.md)


litiliu:
@luoyuxia  @fresh-borzoni  PTAL

litiliu:
@leekeiabstraction  Could you please take a look?

fluss-filesystems/fluss-fs-azure/src/test/java/org/apache/fluss/fs/azure/AbfsFileSystemBehaviorITCase.java
```diff
@@ -53,6 +54,7 @@ static void setup() {
                 CONFIG_PREFIX + ".oauth2.client.endpoint",
                 ENDPOINT_PREFIX + mockAuthServer.getPort());
         configuration.setString(CONFIG_PREFIX + ".key", AZURE_ACCOUNT_KEY);
+        configuration.setBoolean(HNS_ENABLED, false);
```
leekeiabstraction, line 57: Can you elaborate why this change is necessary?

leekeiabstraction (commented):
TY for the PR. Did a quick eyeball, left a question. Will follow up with more thorough review once addressed.

fluss-filesystems/fluss-fs-azure/src/test/java/org/apache/fluss/fs/azure/AbfsFileSystemBehaviorITCase.java
litiliu: These tests use the fake Azure endpoint test.dfs.core.windows.net. With Hadoop 3.4.3, if fs.azure.account.hns.enabled is not configured, ABFS performs a root ACL request during filesystem initialization to detect the account type. Since fsPath.getFileSystem() initializes ABFS before the test replaces its backing filesystem with MemoryFileSystem, this request causes unnecessary DNS/network retries and make CI time out.

litiliu: @leekeiabstraction the question has been addressed, please reiew when you have time.

fluss-filesystems/fluss-fs-s3/pom.xml
```diff
@@ -218,8 +255,8 @@
             <version>${fs.hadoopshaded.version}</version>
             <exclusions>
                 <exclusion>
-                    <groupId>com.amazonaws</groupId>
-                    <artifactId>aws-java-sdk-bundle</artifactId>
+                    <groupId>software.amazon.awssdk</groupId>
+                    <artifactId>bundle</artifactId>
```
leekeiabstraction, line 259: I think the glue guideline will have to be updated and re-tested as well as it currently refers to `com.amazonaws` packages.

fluss-filesystems/fluss-fs-s3/src/main/java/org/apache/fluss/fs/s3/token/S3DelegationTokenProvider.java
```diff
@@ -120,84 +123,78 @@ public S3DelegationTokenProvider(String scheme, Configuration conf) throws IOExc
...
-            builder.withCredentials(stsCredentialsProvider);
+            builder.credentialsProvider(stsCredentialsProvider);
         }
 
         if (stsEndpoint != null) {
-            builder.withEndpointConfiguration(
-                    new AwsClientBuilder.EndpointConfiguration(stsEndpoint, region));
-        } else {
-            builder.withRegion(region);
+            builder.endpointOverride(URI.create(stsEndpoint));
```
leekeiabstraction: URI would return null scheme/protocol here if stsEndpoint string does not have it. I believe V1 SDK prepends with "https://". V2 SDK does not normalise because the method argument type is URI. I believe without defined scheme/protocol, .getScheme() will return null. Will this cause an NPE somewhere later?

website/docs/maintenance/tiered-storage/filesystems/s3.md
```diff
@@ -107,17 +109,17 @@ The server authenticates using its own credentials (static keys, IRSA, instance
...
 remote.data.dir: s3://<your-bucket>/path/to/remote/storage
 s3.region: <your-s3-region>
-s3.aws.credentials.provider: com.amazonaws.auth.profile.ProfileCredentialsProvider
+s3.aws.credentials.provider: software.amazon.awssdk.auth.credentials.ProfileCredentialsProvider
 ```
 
 | Configuration | Description |
 |---|---|
-| `s3.aws.credentials.provider` | Fully qualified class name of an AWS SDK / Hadoop S3A credentials provider. It is used both for the server's own S3 access and for resolving the credentials used to mint temporary client credentials via `GetSessionToken`. Takes precedence over `s3.access-key`/`s3.secret-key` when both are configured. |
+| `s3.aws.credentials.provider` | Fully qualified class name of an AWS SDK V2 / Hadoop S3A credentials provider implementing `AwsCredentialsProvider`. It is used both for the server's own S3 access and for resolving the credentials used to mint temporary client credentials via `GetSessionToken`. Takes precedence over `s3.access-key`/`s3.secret-key` when both are configured. |
```
leekeiabstraction, line 132: This is breaking change that requires users to update their configuration. Should upgrade note be update as well?

> website/docs/maintenance/operations/upgrade-notes-1.0.md

fluss-filesystems/fluss-fs-s3/src/main/resources/META-INF/NOTICE
```diff
@@ -1,59 +1,111 @@
...
+- software.amazon.awssdk:retries-spi:2.44.4
+- software.amazon.awssdk:s3:2.44.4
+- software.amazon.awssdk:s3-transfer-manager:2.44.4
+- software.amazon.awssdk:sdk-core:2.44.4
+- software.amazon.awssdk:sts:2.44.4
+- software.amazon.awssdk:third-party-jackson-core:2.44.4
+- software.amazon.awssdk:utils:2.44.4
+- software.amazon.awssdk:utils-lite:2.44.4
+- software.amazon.eventstream:eventstream:1.0.1
+- software.amazon.s3.analyticsaccelerator:analyticsaccelerator-s3:1.3.1
```
leekeiabstraction, line 97: This uber jar relocated third party library into shaded output. Its dependencies caffeine, failsafe and lib thrift do not appear in this NOTICE file. Do we need to add those entries in?

leekeiabstraction (commented):
TY for the PR, left further comments. PTAL 🙏

fluss-filesystems/fluss-fs-s3/src/main/resources/META-INF/NOTICE
```diff
@@ -1,59 +1,111 @@
...
-- org.apache.hadoop:hadoop-annotations:3.3.4
-- org.apache.hadoop:hadoop-auth:3.3.4
-- org.apache.hadoop:hadoop-aws:3.3.4
-- org.apache.hadoop:hadoop-common:3.3.4
-- org.apache.hadoop.thirdparty:hadoop-shaded-guava:1.1.1
-- org.apache.hadoop.thirdparty:hadoop-shaded-protobuf_3_7:1.1.1
+- org.apache.commons:commons-text:1.14.0
+- org.apache.hadoop:hadoop-annotations:3.4.3
+- org.apache.hadoop:hadoop-auth:3.4.3
+- org.apache.hadoop:hadoop-aws:3.4.3
```
leekeiabstraction, line 45: `hadoop-aws` changes default input stream to S3 analytics accelerator, this has impact on resource usage as by default the accelerator has memory capacity of 2 GiB. I think we need to do something like what Flink side is doing, overriding the default to classic mode instead. 

ref: https://github.com/apache/flink/pull/27026/changes

fluss-filesystems/fluss-fs-s3/pom.xml
litiliu: Nice catch. I updated the Glue guide to use the AWS SDK v2 credential provider classes and added coverage to verify that the documented ECS, EKS, and EC2 provider chains can be loaded.

fluss-filesystems/fluss-fs-s3/src/main/java/org/apache/fluss/fs/s3/token/S3DelegationTokenProvider.java
litiliu: Nice catch. A scheme-less endpoint does fail AWS SDK v2 validation. The code now preserves the SDK v1 behavior by defaulting to HTTPS when the scheme is absent, with tests for scheme-less, HTTP, and HTTPS endpoints.

website/docs/maintenance/tiered-storage/filesystems/s3.md
litiliu: Nice catch. I added an upgrade note covering the AWS SDK v1-to-v2 credential provider migration, including the affected configuration keys, built-in provider mappings, and the custom provider interface change.

fluss-filesystems/fluss-fs-s3/src/main/resources/META-INF/NOTICE
litiliu: Nice catch. I verified the shaded JAR contents and upstream dependency metadata, then added Caffeine 2.9.3, Failsafe 3.3.2, Parquet Format 2.10.0, and libthrift 0.19.0 to the bundled dependency inventory.

fluss-filesystems/fluss-fs-s3/src/main/resources/META-INF/NOTICE
litiliu: Nice catch. Fluss now defaults to the classic input stream only when no stream type is configured and the legacy prefetch option is not enabled. Explicit analytics and prefetch configurations are preserved, with tests covering all three cases.

litiliu:
@leekeiabstraction Nice catches. I have addressed the review comments in 445fc7d and resolved the corresponding threads. Please continue the review when you have time. Thanks!

leekeiabstraction:
Thank you @litiliu for the updates. Can you confirm if the steps in the glue doc has been tested?

@qzyu999 Can you review the glue doc changes here as you have context? Much appreciated.

litiliu:
> Thank you @litiliu for the updates. Can you confirm if the steps in the glue doc has been tested?
> 
> @qzyu999 Can you review the glue doc changes here as you have context? Much appreciated.

@leekeiabstraction Thanks for checking. To be precise, I have not re-run the complete Glue walkthrough end to end on ECS, EKS, or EC2. I verified that the documented AWS SDK v2 provider classes exist and implement the interface expected by Hadoop S3A, but I did not test live credential retrieval against the AWS runtime metadata endpoints.

After another look, I think the guide should avoid maintaining separate provider lists for ECS, EKS, and EC2 altogether. AWS SDK v2 already provides [`DefaultCredentialsProvider`](https://docs.aws.amazon.com/java/api/latest/software/amazon/awssdk/auth/credentials/DefaultCredentialsProvider.html), which owns the environment-detection logic for environment variables, EKS web identity, ECS task roles, EC2 instance profiles, and shared AWS profiles.
I therefore propose replacing the environment-specific table with:
`fs.s3a.aws.credentials.provider: software.amazon.awssdk.auth.credentials.DefaultCredentialsProvider`
This keeps the AWS SDK as the single source of truth and prevents the documentation from drifting when its provider chain changes. Users can still configure a specific provider when they intentionally need to restrict the credential source.
What do you think,  @qzyu999 ?

qzyu999:
Hi @litiliu and @leekeiabstraction, the SDK v2 class name swap for the existing providers LGTM, just a mechanical migration.

The change for `DefaultCredentialsProvider` I haven't tested, but IIUC it seems that it would be set in the `server.yaml` which ideally is sufficient for majority of use-cases. It would also still provide room for users to do custom overrides in case they have some outlier setup that has issues with `DefaultCredentialsProvider`, or they want to maybe avoid some slower cold starts with it as I believe it will try multiple methods with timeout until one works.

SGTM as long as the user can still choose their own path in case `DefaultCredentialsProvider` isn't suitable.

However, until it has been tested E2E it's hard to say for sure what should go in the docs regarding this issue. Perhaps we can go that route while adding some note that it's not tested in all possible use-cases? That might not be necessary either.

litiliu:
@qzyu999 Thanks for confirming the AWS SDK v2 class-name swap. I agree that switching the guide to DefaultCredentialsProvider would require additional end-to-end validation. To keep this PR focused on the mechanical v1-to-v2 migration, I won’t include that additional change in this PR and will retain the current provider mappings. @leekeiabstraction

leekeiabstraction:
Sounds good to me @litiliu, can either you or @qzyu999 raise an issue to follow up with the DefaultCredentialsProvider + manual e2e test/verification? Thank you.

litiliu:
@leekeiabstraction, Created #3916 to track the DefaultCredentialsProvider evaluation and manual E2E verification. If you have no further concerns, would you mind approving this PR when you have time? Thanks!

fluss-filesystems/fluss-fs-s3/src/test/java/org/apache/fluss/fs/s3/SeaweedFsS3FileSystemBehaviorITCase.java
```diff
@@ -0,0 +1,148 @@
...
+import org.testcontainers.junit.jupiter.Testcontainers;
+
+import java.nio.charset.StandardCharsets;
+import java.time.Duration;
+import java.util.Arrays;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+/** Tests the Hadoop S3A and AWS SDK v2 path against a hermetic S3-compatible backend. */
+@Testcontainers(disabledWithoutDocker = true)
```
leekeiabstraction, line 41: Why do we set `disabledWithoutDocker`? Doing so may cause the test be skipped instead of failed if CI / local build does not have docker running

leekeiabstraction (approved):

leekeiabstraction:
Approved, TY for your contribution!

PR #3702, opened 2026-07-20
[server] Fix spurious ISR shrink after leader election

### LINKED ISSUE #3698: [server] NotifyLeaderAndIsr may trigger a spurious ISR shrink
### Search before asking

### Fluss version

main (development)

### Please describe the bug 🐞

A newly elected leader may remove healthy followers from ISR as soon as its first write advances the leader LEO, before those followers send their next fetch. This includes the initial leader activation of a newly created partition. A repeated `NotifyLeaderAndIsr` request with the same leader epoch may cause the same problem.

This can appear as an `isrShrinksPerSecond` spike when traffic starts on an otherwise healthy table, followed by an ISR expansion after the next follower fetch. This was observed when an existing primary-key table created a new partition and immediately received `PutKv` traffic.

#### How to reproduce

1. Create a new partition and make TabletServer 1 the leader of one of its buckets with replicas and ISR `[1, 2, 3]`.
2. Issue a write that advances the leader LEO before followers 2 and 3 send their first fetch to the leader.
3. Trigger the ISR expiration check.

Expected: the ISR remains `[1, 2, 3]` during the normal `log.replica.max-lag-time` grace period.

Actual: the ISR shrinks to `[1]`. Followers 2 and 3 can rejoin after their next fetch.

#### Root cause

There are two related follower lifecycle problems:

1. When a local replica is not already the leader, its `followerReplicasMap` may be empty. This happens both during the initial leader activation of a newly created partition, where [`ReplicaManager.maybeCreateReplica()`](https://github.com/apache/fluss/blob/d7057f52f5ef9a9e61d0b0718f23734eff9f2302/fluss-server/src/main/java/org/apache/fluss/server/replica/ReplicaManager.java#L2066-L2115) creates a fresh `Replica`, and during a follower-to-leader transition, where [`Replica.makeFollower()`](https://github.com/apache/fluss/blob/d7057f52f5ef9a9e61d0b0718f23734eff9f2302/fluss-server/src/main/java/org/apache/fluss/server/replica/Replica.java#L478-L532) has cleared the map. [`Replica.makeLeader()`](https://github.com/apache/fluss/blob/d7057f52f5ef9a9e61d0b0718f23734eff9f2302/fluss-server/src/main/java/org/apache/fluss/server/replica/Replica.java#L419-L469) then creates new `FollowerReplica` objects. These objects start with an unknown LEO and `lastCaughtUpTimeMs = 0`. Because `resetFollowerReplicaState()` is not called, followers in the committed ISR do not receive the normal leader-election lag grace period.
2. When an existing leader receives another `NotifyLeaderAndIsr` request, [`updateAssignmentAndIsr()`](https://github.com/apache/fluss/blob/d7057f52f5ef9a9e61d0b0718f23734eff9f2302/fluss-server/src/main/java/org/apache/fluss/server/replica/Replica.java#L1219-L1247) unconditionally replaces every `FollowerReplica`. This discards the follower LEO and caught-up timestamps, even when the leader epoch and replica assignment have not changed.

[`FollowerReplica.resetFollowerReplicaState()`](https://github.com/apache/fluss/blob/d7057f52f5ef9a9e61d0b0718f23734eff9f2302/fluss-server/src/main/java/org/apache/fluss/server/replica/FollowerReplica.java#L98-L128) already defines the intended initialization for leader election and same-broker re-election, but it currently has no production caller.

### Solution

Follow the replica lifecycle used by Kafka:

- Keep existing `FollowerReplica` objects for unchanged assignments, for example with `computeIfAbsent`, so repeated notifications do not discard their fetch state.
- When the leader epoch advances, reset each follower through the existing `resetFollowerReplicaState(...)` method. Followers in the committed ISR should receive the normal lag grace period.
- Add regression tests for an immediate first write after initial leader activation, a repeated notification with the same leader epoch, and a same-broker leader re-election.

Kafka similarly [resets remote replica state when the leader epoch changes](https://github.com/apache/kafka/blob/56455183deefcd50f6b76f5dea5e84dd884274cb/core/src/main/scala/kafka/cluster/Partition.scala#L633-L660) and [retains existing replica objects with `computeIfAbsent`](https://github.com/apache/kafka/blob/56455183deefcd50f6b76f5dea5e84dd884274cb/core/src/main/scala/kafka/cluster/Partition.scala#L843-L849).

fxbing:
### Purpose

Linked issue: close #3698

A newly elected leader may remove healthy followers from ISR immediately after the first write because their newly created `FollowerReplica` state does not receive the normal lag grace period.

A repeated `NotifyLeaderAndIsr` request can also replace existing `FollowerReplica` objects and discard their fetch offsets and caught-up timestamps, causing the same spurious ISR shrink.

### Brief change log

- Preserve existing `FollowerReplica` objects for unchanged assignments.
- Reset follower state when the leader epoch advances:
  - initialize ISR followers with the normal lag grace period for a new leader;
  - preserve follower LEO for a same-broker leader re-election.
- Keep follower state unchanged for repeated notifications with the same leader epoch.
- Add regression tests for the initial leader election, repeated notification, and same-broker re-election cases.

### Tests

- `AdjustIsrTest#testShrinkIsr`
- `AdjustIsrTest#testDoNotShrinkIsrAfterRepeatedNotifyLeaderAndIsr`
- `AdjustIsrTest#testPreserveFollowerStateOnLeaderReelection`
- `AdjustIsrTest`, `FollowerReplicaTest`, and `ReplicaTest`: 25 tests passed.
- Checkstyle and Spotless checks passed.

### API and Format

No public API, configuration, metric, RPC, or storage-format changes.

### Documentation

No documentation changes. This is an internal replica lifecycle bug fix.

### Generative AI disclosure

### DIFF
--- a/fluss-server/src/main/java/org/apache/fluss/server/replica/Replica.java
+++ b/fluss-server/src/main/java/org/apache/fluss/server/replica/Replica.java
@@ -437,9 +437,21 @@ public final class Replica {
 
                             int requestLeaderEpoch = data.getLeaderEpoch();
                             if (requestLeaderEpoch > leaderEpoch) {
+                                boolean isNewLeader = !isLeader();
                                 leaderEpoch = requestLeaderEpoch;
                                 onBecomeNewLeader();
                                 leaderReplicaIdOpt.set(localTabletServerId);
+                                currentTimeMs = clock.milliseconds();
+                                long leaderEndOffset = logTablet.localLogEndOffset();
+                                for (FollowerReplica followerReplica :
+                                        followerReplicasMap.values()) {
+                                    followerReplica.resetFollowerReplicaState(
+                                            currentTimeMs,
+                                            leaderEndOffset,
+                                            isNewLeader,
+                                            data.getIsr()
+                                                    .contains(followerReplica.getFollowerId()));
+                                }
                                 LOG.info(
                                         "TabletServer {} becomes leader for bucket {}",
                                         localTabletServerId,
@@ -1233,7 +1245,8 @@ public final class Replica {
             // Due to code paths accessing followerReplicasMap without a lock, first add the new
             // replicas and then remove the old ones.
             for (Integer replica : followers) {
-                followerReplicasMap.put(replica, new FollowerReplica(replica, tableBucket));
+                followerReplicasMap.computeIfAbsent(
+                        replica, id -> new FollowerReplica(id, tableBucket));
             }
             for (Integer replica : removedReplicas) {
                 followerReplicasMap.remove(replica);
--- a/fluss-server/src/main/java/org/apache/fluss/server/replica/Replica.java
+++ b/fluss-server/src/main/java/org/apache/fluss/server/replica/Replica.java
@@ -441,6 +441,8 @@ public final class Replica {
                                 leaderEpoch = requestLeaderEpoch;
                                 onBecomeNewLeader();
                                 leaderReplicaIdOpt.set(localTabletServerId);
+                                // onBecomeNewLeader may recover a KV snapshot, so start the ISR lag
+                                // grace period after it completes.
                                 currentTimeMs = clock.milliseconds();
                                 long leaderEndOffset = logTablet.localLogEndOffset();
                                 for (FollowerReplica followerReplica :
--- a/fluss-server/src/test/java/org/apache/fluss/server/replica/AdjustIsrTest.java
+++ b/fluss-server/src/test/java/org/apache/fluss/server/replica/AdjustIsrTest.java
@@ -24,7 +24,9 @@ import org.apache.fluss.exception.IneligibleReplicaException;
 import org.apache.fluss.metadata.TableBucket;
 import org.apache.fluss.rpc.entity.ProduceLogResultForBucket;
 import org.apache.fluss.server.entity.FetchReqInfo;
+import org.apache.fluss.server.entity.NotifyLeaderAndIsrData;
 import org.apache.fluss.server.log.FetchParams;
+import org.apache.fluss.server.zk.data.LeaderAndIsr;
 
 import org.junit.jupiter.api.Test;
 
@@ -105,7 +107,7 @@ public class AdjustIsrTest extends ReplicaTestBase {
     }
 
     @Test
-    void testShrinkIsr() {
+    void testShrinkIsr() throws Exception {
         // replica set is 1,2,3 , isr set is 1,2,3.
         TableBucket tb = new TableBucket(DATA1_TABLE_ID, 1);
         makeLogTableAsLeader(tb, Arrays.asList(1, 2, 3), Arrays.asList(1, 2, 3), false);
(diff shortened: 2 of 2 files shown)


fxbing:
@swuferhong PTAL

fluss-server/src/main/java/org/apache/fluss/server/replica/Replica.java
```diff
@@ -437,9 +437,21 @@ public void makeLeader(NotifyLeaderAndIsrData data) throws IOException {
 
                             int requestLeaderEpoch = data.getLeaderEpoch();
                             if (requestLeaderEpoch > leaderEpoch) {
+                                boolean isNewLeader = !isLeader();
                                 leaderEpoch = requestLeaderEpoch;
                                 onBecomeNewLeader();
                                 leaderReplicaIdOpt.set(localTabletServerId);
+                                currentTimeMs = clock.milliseconds();
```
swuferhong, line 446: Why need to reset again? This looks unnecessary.

fluss-server/src/test/java/org/apache/fluss/server/replica/AdjustIsrTest.java
```diff
@@ -105,20 +107,70 @@ tb, new FetchReqInfo(tb.getTableId(), 10L, Integer.MAX_VALUE)),
...
+        fetchFromFollower(tb, 2, leaderEndOffset);
+        fetchFromFollower(tb, 3, leaderEndOffset);
+
+        int newBucketEpoch = replica.getBucketEpoch() + 1;
+        notifyLeaderAndIsr(replica, replicas, replica.getLeaderEpoch(), newBucketEpoch);
+        assertThat(replica.getBucketEpoch()).isEqualTo(newBucketEpoch);
+
+        manualClock.advanceTime(
+                conf.get(ConfigOptions.LOG_REPLICA_MAX_LAG_TIME).toMillis() + 1,
+                TimeUnit.MILLISECONDS);
```
swuferhong: the followers have already fetched up to the leader LEO, so isCaughtUp passes via the first condition (leaderEndOffset == getLogEndOffset()) and short-circuits. That means the tests never really verify that lastCaughtUpTimeMs was reset/preserved correctly.

Could we append one more batch after the notification, so that the leader LEO is ahead of the follower LEO? Then the assertion can only pass through currentTimeMs - lastCaughtUpTimeMs <= maxLag, which distinguishes "follower LEO was preserved" from "follower got the grace period".

swuferhong (commented):
LGTM overall. Left some mirror comments:

fluss-server/src/main/java/org/apache/fluss/server/replica/Replica.java
fxbing: This only refreshes currentTimeMs; it does not reset the follower state twice. onBecomeNewLeader() may perform time-consuming KV recovery, including downloading and restoring a snapshot. We re-read the clock afterward so the ISR lag grace period starts after leader initialization completes instead of being consumed during recovery. I've added a comment to clarify this.

fluss-server/src/test/java/org/apache/fluss/server/replica/AdjustIsrTest.java
fxbing: Good point. I appended another batch after the repeated notification so the leader LEO is ahead of the follower LEO. I removed the max-lag time advance because this assertion now specifically verifies the preserved `lastCaughtUpTimeMs` within the grace period; advancing beyond `maxLag` should correctly trigger an ISR shrink. The expiration behavior remains covered by `testShrinkIsr`.

swuferhong (approved):
LGTM

PR #3711, opened 2026-07-21
[rust][client] Support comma-separated bootstrap servers

### LINKED ISSUE #3710: [rust][client] Support comma-separated bootstrap servers
### Fluss version

0.9.0 (latest release)

### Search before asking

- [x] I searched in the [issues](https://github.com/apache/fluss-rust/issues) and found nothing similar.

### Please describe the bug 

https://github.com/apache/fluss-rust/issues/585

The Rust client currently supports only one bootstrap endpoint, even though the public configuration is named `bootstrap_servers` / `bootstrap.servers`.

A comma-separated value such as:

```rust
let mut config = Config::default();
config.bootstrap_servers =
    "127.0.0.1:9123,127.0.0.1:9124".to_string();

let conn = FlussConnection::create(config).await?;
```

is passed as a single string through `Metadata::new` and `Metadata::init_cluster`.

`Metadata::parse_bootstrap` calls `to_socket_addrs()` on the complete string, causing the comma-separated value to be interpreted as one socket address and fail to resolve.

### Expected behavior

The Rust client should:
- Parse bootstrap_servers as a comma-separated list of host:port entries.
- Trim whitespace around each entry.
- Reject empty entries.
- Try the configured endpoints in deterministic order.
- Continue to the next endpoint if an earlier endpoint cannot be resolved or connected.
- Preserve the configured list when reinit_cluster reinitializes metadata.
- Preserve existing behavior for a single bootstrap endpoint.

This would align the Rust client with the Java client, where bootstrap.servers is already defined as a list of bootstrap endpoints.

### Impact
Users of the Rust client cannot currently configure multiple bootstrap nodes for high availability or failover.
This also affects clients built on the Rust implementation, including the Python, C++, and Elixir bindings.
If the only configured bootstrap endpoint is unavailable, these clients cannot initialize metadata even when other cluster endpoints are healthy.

### Solution

_No response_

slfan1989:
### Purpose

Linked issue: close #3710

### Brief change log

Support comma-separated bootstrap_servers in the Rust client so users can configure multiple bootstrap endpoints for high availability.

#### Brief change log
- Parse bootstrap_servers as a comma-separated list of bootstrap endpoints.
- Trim whitespace around each endpoint and reject empty entries.
- Try bootstrap servers in order during metadata initialization until one succeeds.
- Preserve existing single-bootstrap-server behavior.

### Tests

Added unit tests.

### API and Format

### Documentation

No documentation changes.

### DIFF
--- a/fluss-rust/crates/fluss/src/client/metadata.rs
+++ b/fluss-rust/crates/fluss/src/client/metadata.rs
@@ -22,7 +22,7 @@ use crate::metadata::{PhysicalTablePath, TableBucket, TablePath};
 use crate::proto::MetadataResponse;
 use crate::rpc::message::UpdateMetadataRequest;
 use crate::rpc::{RpcClient, ServerConnection};
-use log::info;
+use log::{info, warn};
 use parking_lot::RwLock;
 use std::collections::HashSet;
 use std::net::{SocketAddr, ToSocketAddrs};
@@ -94,11 +94,7 @@ impl Metadata {
         for bootstrap in bootstrap_servers.split(',') {
             let bootstrap = bootstrap.trim();
             if bootstrap.is_empty() {
-                return Err(Error::IllegalArgument {
-                    message: format!(
-                        "Invalid bootstrap servers '{bootstrap_servers}': empty bootstrap server"
-                    ),
-                });
+                continue;
             }
             bootstraps.push(BootstrapServer {
                 raw: bootstrap.to_string(),
@@ -117,13 +113,20 @@ impl Metadata {
 
     async fn init_cluster(bootstrap_servers: &str, connections: Arc<RpcClient>) -> Result<Cluster> {
         let bootstraps = Self::parse_bootstrap_servers(bootstrap_servers)?;
+        let bootstrap_count = bootstraps.len();
         let mut errors = Vec::new();
         let mut last_connection_error = None;
 
-        for bootstrap in bootstraps {
+        for (index, bootstrap) in bootstraps.into_iter().enumerate() {
             let socket_address = match bootstrap.address {
                 Ok(socket_address) => socket_address,
                 Err(err) => {
+                    if index + 1 < bootstrap_count {
+                        warn!(
+                            "Failed to resolve bootstrap server '{}', trying the next server: {err}",
+                            bootstrap.raw
+                        );
+                    }
                     errors.push(format!("{}: {err}", bootstrap.raw));
                     continue;
                 }
@@ -138,7 +141,13 @@ impl Metadata {
             match Self::fetch_cluster_from_bootstrap(&server_node, connections.clone()).await {
                 Ok(cluster) => return Ok(cluster),
                 Err(err) => {
-                    errors.push(format!("{socket_address}: {err}"));
+                    if index + 1 < bootstrap_count {
+                        warn!(
+                            "Failed to initialize cluster from bootstrap server '{}', trying the next server: {err}",
+                            bootstrap.raw
+                        );
+                    }
+                    errors.push(format!("{}: {err}", bootstrap.raw));
                     last_connection_error = Some(err);
                 }
             }
@@ -157,6 +166,10 @@ impl Metadata {
         last_connection_error: Option<Error>,
     ) -> Error {
         if let Some(error) = last_connection_error {
+            warn!(
+                "Unable to initialize cluster from bootstrap servers '{bootstrap_servers}': {}",
+                errors.join("; ")
+            );
             return error;
         }
 
@@ -508,13 +521,17 @@ mod tests {
     }
 
     #[test]
-    fn parse_bootstrap_rejects_empty_comma_separated_entries() {
-        let err = Metadata::parse_bootstrap_servers("127.0.0.1:8080, , localhost:9090")
-            .expect_err("empty bootstrap entries should be rejected");
+    fn parse_bootstrap_ignores_empty_comma_separated_entries() {
+        let bootstraps =
+            Metadata::parse_bootstrap_servers("127.0.0.1:8080, , localhost:9090,").unwrap();
 
-        assert!(
-            err.to_string().contains("empty bootstrap server"),
-            "unexpected error: {err}"
-        );
+        assert_eq!(bootstraps.len(), 2);
+        assert_eq!(bootstraps[0].raw, "127.0.0.1:8080");
+        assert_eq!(bootstraps[1].raw, "localhost:9090");
+    }
+
+    #[test]
+    fn parse_bootstrap_rejects_config_without_servers() {
+        assert!(Metadata::parse_bootstrap_servers(" , ").is_err());
     }
 }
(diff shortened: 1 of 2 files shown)


fluss-rust/crates/fluss/src/client/metadata.rs
```diff
@@ -77,20 +83,95 @@ impl Metadata {
...
+                -1,
+                socket_address.ip().to_string(),
+                socket_address.port() as u32,
+                ServerType::Unknown,
+            );
+
+            match Self::fetch_cluster_from_bootstrap(&server_node, connections.clone()).await {
+                Ok(cluster) => return Ok(cluster),
+                Err(err) => {
+                    errors.push(format!("{socket_address}: {err}"));
```
fresh-borzoni: Can we add a warn! when a server fails and we move to the next one? Right now failover is invisible — nothing tells the operator that server 1 is down.

fluss-rust/crates/fluss/src/client/metadata.rs
```diff
@@ -77,20 +83,95 @@ impl Metadata {
...
+            last_connection_error,
+        ))
+    }
+
+    fn bootstrap_initialization_error(
+        bootstrap_servers: &str,
+        errors: Vec<String>,
+        last_connection_error: Option<Error>,
+    ) -> Error {
+        if let Some(error) = last_connection_error {
```
fresh-borzoni, line 168: If two servers fail with different errors, the user only sees the last one. E.g. auth error on A, timeout on B -> they see only the timeout. Can we log all of errors before returning?

fluss-rust/crates/fluss/src/client/metadata.rs
```diff
@@ -77,20 +83,95 @@ impl Metadata {
...
-            -1,
-            socket_address.ip().to_string(),
-            socket_address.port() as u32,
-            ServerType::Unknown,
+    fn parse_bootstrap_servers(bootstrap_servers: &str) -> Result<Vec<BootstrapServer>> {
+        let mut bootstraps = Vec::new();
+
+        for bootstrap in bootstrap_servers.split(',') {
+            let bootstrap = bootstrap.trim();
+            if bootstrap.is_empty() {
```
fresh-borzoni, line 96: "host:9123," (trailing comma) fails here, but works in the Java client as it skips empty entries. 
Should we do the same?

fluss-rust/crates/fluss/src/client/metadata.rs
```diff
@@ -77,20 +83,95 @@ impl Metadata {
...
+                }
+            };
+            let server_node = ServerNode::new(
+                -1,
+                socket_address.ip().to_string(),
+                socket_address.port() as u32,
+                ServerType::Unknown,
+            );
+
+            match Self::fetch_cluster_from_bootstrap(&server_node, connections.clone()).await {
```
fresh-borzoni, line 141: There's no test for the failover itself like server A fails, server B succeeds. Can we add one?

fluss-rust/crates/fluss/src/client/metadata.rs
```diff
@@ -77,20 +83,95 @@ impl Metadata {
...
+                    ),
+                });
+            }
+            bootstraps.push(BootstrapServer {
+                raw: bootstrap.to_string(),
+                address: Self::parse_bootstrap(bootstrap),
+            });
+        }
+
+        if bootstraps.is_empty() {
```
fresh-borzoni, line 105: nit: Dead code as split(',') never returns zero entries, so "" hits the empty-entry error above first.

fluss-rust/crates/fluss/src/client/metadata.rs
fresh-borzoni: nit: Error messages mix labels: parse errors show what the user configured (bootstrap.raw), connection errors show the resolved IP. Shall we use raw in both?

fresh-borzoni (commented):
@slfan1989 Thank you for the PR, left some comments PTAL

One follow-up:Java retries a single bootstrap server 3x with backoff on retriable errors (tryToInitializeClusterWithRetries, the "bootstrap is recovering" case). The Rust client still has zero retries. Pre-existing, out of scope here, but worth an issue. 
WDYT?

fluss-rust/crates/fluss/src/client/metadata.rs
slfan1989: Thanks for the suggestion! I added `warn!` logs before falling back to the next bootstrap server. The log includes the configured endpoint and failure reason, and is emitted only when another server is available.

fluss-rust/crates/fluss/src/client/metadata.rs
slfan1989: I added a summary warning containing all bootstrap server failures before returning. The last original error is still returned to preserve its type and error code for the bindings.

fluss-rust/crates/fluss/src/client/metadata.rs
slfan1989: Thanks for pointing out the difference! I aligned the Rust client with the Java client by skipping empty comma-separated entries.

Configurations such as `host:9123,` and `host1:9123,,host2:9123` are now accepted, while configurations containing no valid servers, such as `""`, `","`, or `" , "`, are still rejected.

I also added unit tests for both cases.

fluss-rust/crates/fluss/src/client/metadata.rs
slfan1989: I added an integration test that uses an unreachable endpoint as the first bootstrap server and the running test cluster as the second. It verifies that the client falls back to the second endpoint and initializes successfully.

fluss-rust/crates/fluss/src/client/metadata.rs
slfan1989: This check is now reachable after empty bootstrap entries are filtered out. It rejects configurations containing no actual bootstrap servers, such as an empty string or " , ".  I also added a unit test covering this case.

fluss-rust/crates/fluss/src/client/metadata.rs
slfan1989: Updated to use `bootstrap.raw` consistently for connection errors, so the originally configured address is preserved.

slfan1989:
> @slfan1989 Thank you for the PR, left some comments PTAL
> 
> One follow-up:Java retries a single bootstrap server 3x with backoff on retriable errors (tryToInitializeClusterWithRetries, the "bootstrap is recovering" case). The Rust client still has zero retries. Pre-existing, out of scope here, but worth an issue. WDYT?

@fresh-borzoni Thank you for the review and for pointing this out! I agree that retrying a single bootstrap server on retriable errors would improve consistency with the Java client and better handle temporarily recovering servers. Since it is pre-existing and outside the scope of this PR, I’ll create a separate issue to track it.

fresh-borzoni (approved):
@slfan1989 Thank you, LGTM 👍

slfan1989:
> @slfan1989 Thank you, LGTM 👍

@fresh-borzoni Thank you for the review and helpful suggestions! I really appreciate it.

PR #3694, opened 2026-07-19
[rust][tests] Fix flaky test_poll_batches by using retry loop for projection test

LiebingYu:
### Purpose

Linked issue: close #3693

PR #3712, opened 2026-07-21
[rust] Fix lost wakeup in LogFetchBuffer::await_not_empty

### LINKED ISSUE #3713: [rust] Lost wakeup in LogFetchBuffer::await_not_empty causes empty poll
### Search before asking

### Fluss version

0.9.0 (latest release)

### Please describe the bug 🐞

`await_not_empty` builds its `notified()` future after the `is_empty()` check, and the buffer signals via `notify_waiters()`, so a fetch landing in that window is lost and `poll()` blocks to its timeout, returning empty with data available. 
Root cause of the flaky `test_poll_batches` (#3694).

### Solution

arm the waiter with `enable()` before the check (tokio's documented `notified()`/`enable()` pattern).

fresh-borzoni:
closes https://github.com/apache/fluss/issues/3713

Arm the waiter with `enable()` before the emptiness check so a `notify_waiters()` racing the check isn't lost. 
Root cause behind #3694's flaky poll. 

Also drops the unused `DEFAULT_POLL_INTERVAL`.

### DIFF
--- a/fluss-rust/crates/fluss/src/client/table/log_fetch_buffer.rs
+++ b/fluss-rust/crates/fluss/src/client/table/log_fetch_buffer.rs
@@ -139,35 +139,34 @@ impl LogFetchBuffer {
     pub async fn await_not_empty(&self, timeout: Duration) -> Result<bool> {
         let deadline = Instant::now() + timeout;
 
+        // Arm before checking: notify_waiters() stores no permit and only wakes already-armed
+        // waiters, so arming late would drop a notification that races the emptiness check.
+        let notified = self.not_empty_notify.notified();
+        tokio::pin!(notified);
+
         loop {
-            // Check if buffer is not empty
+            notified.as_mut().enable();
+
             if !self.is_empty() {
                 return Ok(true);
             }
 
-            // Check if woken up
             if self.woken_up.swap(false, Ordering::Acquire) {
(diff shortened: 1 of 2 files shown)


fresh-borzoni:
@charlesdong1991 @LiebingYu PTAL 🙏

fluss-rust/crates/fluss/src/client/table/log_fetch_buffer.rs
```diff
@@ -1064,6 +1063,29 @@ mod tests {
         assert!(completed.take_error().is_some());
     }
 
+    #[tokio::test]
+    async fn await_not_empty_returns_true_when_data_arrives_during_wait() {
```
charlesdong1991, line 1067: will this pass on the old code too?

charlesdong1991 (approved):
lgtm

fluss-rust/crates/fluss/src/client/table/log_fetch_buffer.rs
fresh-borzoni: Yeah, it's not really a regression test. It'd pass on the old code too, since the 20ms delay means we're already parked before the notify fires. It's just the first test of the success path returning true when data arrives mid-wait. 
Before this we only covered the error exits. So fixing this gap just bc at the same place :)

PR #3715, opened 2026-07-21
[test] Tolerate removed tablet servers FlussClusterExtension helpers

### LINKED ISSUE #3716: [test] Tolerate removed tablet servers in FlussClusterExtension wait helpers
### Search before asking

### Description

Follow-up to #3667: the same NPE race exists at three more getTabletServerById call sites in FlussClusterExtension. During a failover the lookup can return null (leader temporarily NO_LEADER, or the server already removed), and since retry(...) only retries AssertionError, the NPE fails the test instead of letting the wait continue.

### Willingness to contribute

fresh-borzoni:
closes https://github.com/apache/fluss/issues/3716

Follow-up to #3667, which fixed a NullPointerException in FlussClusterExtension#getReplica when a tablet server is removed while a failover is in progress. The same race exists at three more call sites.


fresh-borzoni:
@swuferhong PTAL 🙏

PR #3734, opened 2026-07-22
[server] prevent remote log cache size overflow

zuston:
### Purpose

prevent remote log cache size overflow, especially when setting this cache size > 2GiB.


zuston:
minor fix @fresh-borzoni

fresh-borzoni (approved):
@zuston Thank you, LGTM 👍

PR #3736, opened 2026-07-22
Attached Blog Posts to Docs

polyzos:
A while back we generated a few blog posts to go deep into fluss storage hierarchy and lake tiering.
However these blog posts sit on the blog and might be harder to find, or lost as time goes by.

This PR links these deep dive blog posts to the documentation pages, so people can easily find more information just by navigating the docs .


fresh-borzoni (approved):
@polyzos Thank you, merging

PR #3737, opened 2026-07-22
[server] release remote log index cache entries on deletion

zuston:
### Purpose

Explicitly release remote log index cache entries on deletion.

### DIFF
--- a/fluss-server/src/main/java/org/apache/fluss/server/log/remote/LogTieringTask.java
+++ b/fluss-server/src/main/java/org/apache/fluss/server/log/remote/LogTieringTask.java
@@ -158,16 +158,16 @@ public class LogTieringTask implements Runnable {
 
                 if (success) {
                     if (!expiredRemoteLogSegments.isEmpty()) {
-                        // Release the mmap-backed local indexes before deleting the remote files.
-                        remoteLogIndexCache.removeAll(
-                                expiredRemoteLogSegments.stream()
-                                        .map(RemoteLogSegment::remoteLogSegmentId)
-                                        .collect(Collectors.toList()));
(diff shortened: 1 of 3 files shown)


zuston:
ptal @fresh-borzoni

fluss-server/src/main/java/org/apache/fluss/server/log/remote/LogTieringTask.java
```diff
@@ -154,6 +158,11 @@ private void runOnce() throws InterruptedException {
 
                 if (success) {
                     if (!expiredRemoteLogSegments.isEmpty()) {
+                        // Release the mmap-backed local indexes before deleting the remote files.
+                        remoteLogIndexCache.removeAll(
```
fresh-borzoni: Q: shall we removeAll after deleteRemoteLogSegmentFiles instead of before?
In the gap between manifest commit and remote deletion, an in-flight reader can miss the cache and re-download the index (remote files still exist at that point), re-inserting an entry that nothing will ever remove again, essentially the exact leak this PR fixes. 

The read-refcount TODO would fix the reader-error half of this race, but not the leak half and the reorder(the thing that I propose here) composes with it if it ever lands: drain -> delete files -> removeAll.

WDYT?

fresh-borzoni (commented):
@zuston Thank you, this is a very good addition 👍 Left suggestion, PTAL

fluss-server/src/main/java/org/apache/fluss/server/log/remote/LogTieringTask.java
zuston: nice catch. this is a potential race condition.

fresh-borzoni (approved):
@zuston Thank you, LGTM 👍

PR #3738, opened 2026-07-22
[filesystem] avoid zero-skip loop in HadoopDataInputStream

### LINKED ISSUE #3718: [filesystem] HadoopDataInputStream.seek loops forever past EOF
### Search before asking

I searched titles, bodies, and comments for `HadoopDataInputStream`, `skipFully`, `seek past EOF`, `HDFS seek hang`, `Hadoop skip EOF`, and related filesystem terms. I also inspected the open Hadoop dependency-alignment PR #3699; it does not touch this code path.

### Fluss version

- `main (development)` at `31621117db9b21e8e00a6e716b3a13b7b66e18be`
- The same implementation is present in `v0.9.1-incubating`, `v0.9.0-incubating`, and `v0.8.0-incubating`

### Please describe the bug 🐞

`HadoopDataInputStream.seek(long)` optimizes small forward seeks by calling `skipFully(delta)`. If the requested position is past EOF, the underlying `InputStream.skip(long)` returns `0`. The loop in `skipFully` then makes no progress and spins forever:

https://github.com/apache/fluss/blob/31621117db9b21e8e00a6e716b3a13b7b66e18be/fluss-filesystems/fluss-fs-hadoop/src/main/java/org/apache/fluss/fs/hdfs/HadoopDataInputStream.java#L132-L135

This contradicts the public `FSDataInputStream.seek` contract, which says callers cannot seek past the end of the stream and reports seek errors through `IOException`:

https://github.com/apache/fluss/blob/31621117db9b21e8e00a6e716b3a13b7b66e18be/fluss-common/src/main/java/org/apache/fluss/fs/FSDataInputStream.java#L36-L43

Java's `InputStream.skip` contract permits returning fewer bytes than requested, including `0`.

#### Reproduction

The existing `SeekableByteArrayInputStream` test fixture already returns `0` from `skip` at EOF. Adding this focused case to `HadoopDataInputStreamTest` reproduces the hang:

```java
@Test
void testSeekPastEndOfStream() {
    FSDataInputStream input =
            new FSDataInputStream(new SeekableByteArrayInputStream(new byte[1]));
    HadoopDataInputStream stream = new HadoopDataInputStream(input);

    assertThatThrownBy(() -> stream.seek(2)).isInstanceOf(EOFException.class);
}
```

Run it with:

```bash
./mvnw -pl fluss-filesystems/fluss-fs-hadoop -am \
  -Dtest=HadoopDataInputStreamTest \
  -Dsurefire.failIfNoSpecifiedTests=false test
```

I also reproduced this twice against Hadoop's real local filesystem implementation using a one-byte file and `seek(2)`. Both runs timed out after two seconds:

```text
before seek target=2 length=1
TIMEOUT after 2s (seek did not return)
```

The nearby control `seek(1)` returned normally with position `1`.

#### Expected behavior

The call should promptly throw `IOException`/`EOFException` because the requested position is past EOF.

#### Actual behavior

`seek(2)` never returns. The thread remains in the `skipFully` loop.

No special configuration or external service is required.

### Solution

Delegate `skipFully` to Hadoop's existing `org.apache.hadoop.io.IOUtils.skipFully(InputStream, long)`, or implement the same zero-progress handling: when `skip` returns `0`, read one byte and throw `EOFException` if EOF has been reached.

A regression test can be added to the existing `HadoopDataInputStreamTest`. This should require no new dependency and no API or storage-format change.

wolfkill:
Fixes #3718.

## Root Cause

`HadoopDataInputStream.seek()` optimizes small forward seeks by skipping instead of issuing a Hadoop `seek()`.

The old local `skipFully()` loop assumed every `FSDataInputStream.skip(bytes)` call makes forward progress. Some streams return `0` at EOF or when no bytes can currently be skipped. In that case the remaining byte count never decreases and the loop can spin forever.

## Changes

- Delegate small forward skips to Hadoop's `org.apache.hadoop.io.IOUtils.skipFully(InputStream, long)`.
- Remove the local `skipFully` implementation.
- Add a regression test using a seekable stream that returns one zero-byte skip and then EOF on read, verifying `EOFException` and that zero-progress skip is not retried.

## Validation

- `JAVA_HOME=/Users/cc/Library/Java/JavaVirtualMachines/corretto-18.0.2/Contents/Home mvn -pl fluss-filesystems/fluss-fs-hadoop -am -Dtest=HadoopDataInputStreamTest#testSeekPastEndOfStreamThrowsEOFException -DfailIfNoTests=false test`
- `JAVA_HOME=/Users/cc/Library/Java/JavaVirtualMachines/corretto-18.0.2/Contents/Home mvn -pl fluss-filesystems/fluss-fs-hadoop -am -Dtest=HadoopDataInputStreamTest -DfailIfNoTests=false test`
- `JAVA_HOME=/Users/cc/Library/Java/JavaVirtualMachines/corretto-18.0.2/Contents/Home mvn -pl fluss-filesystems/fluss-fs-hadoop -am -DfailIfNoTests=false test`
- `git diff --check upstream/main...HEAD`

### DIFF
--- a/fluss-filesystems/fluss-fs-hadoop/src/main/java/org/apache/fluss/fs/hdfs/HadoopDataInputStream.java
+++ b/fluss-filesystems/fluss-fs-hadoop/src/main/java/org/apache/fluss/fs/hdfs/HadoopDataInputStream.java
@@ -19,6 +19,8 @@ package org.apache.fluss.fs.hdfs;
 
 import org.apache.fluss.fs.FSDataInputStream;
 
+import org.apache.hadoop.io.IOUtils;
+
 import javax.annotation.Nonnull;
 
 import java.io.IOException;
@@ -72,7 +74,7 @@ public final class HadoopDataInputStream extends FSDataInputStream {
 
         if (delta > 0L && delta <= MIN_SKIP_BYTES) {
             // Instead of a small forward seek, we skip over the gap
-            skipFully(delta);
+            IOUtils.skipFully(fsDataInputStream, delta);
         } else if (delta != 0L) {
             // For larger gaps and backward seeks, we do a real seek
             forceSeek(seekPos);
@@ -121,17 +123,5 @@ public final class HadoopDataInputStream extends FSDataInputStream {
      */
     public void forceSeek(long seekPos) throws IOException {
         fsDataInputStream.seek(seekPos);
-    }
-
-    /**
-     * Skips over a given amount of bytes in the stream.
-     *
-     * @param bytes the number of bytes to skip.
-     * @throws IOException
-     */
-    public void skipFully(long bytes) throws IOException {
-        while (bytes > 0) {
-            bytes -= fsDataInputStream.skip(bytes);
-        }
     }
 }
--- a/fluss-filesystems/fluss-fs-hadoop/src/test/java/org/apache/fluss/fs/hdfs/HadoopDataInputStreamTest.java
+++ b/fluss-filesystems/fluss-fs-hadoop/src/test/java/org/apache/fluss/fs/hdfs/HadoopDataInputStreamTest.java
@@ -22,6 +22,7 @@ import org.apache.hadoop.fs.PositionedReadable;
 import org.apache.hadoop.fs.Seekable;
 import org.junit.jupiter.api.Test;
 
+import java.io.EOFException;
 import java.io.IOException;
 import java.io.InputStream;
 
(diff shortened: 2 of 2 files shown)


litiliu:
would it be better to use the org.apache.hadoop.io.IOUtils.skipFully?

wolfkill:
Updated this to use `org.apache.hadoop.io.IOUtils.skipFully(fsDataInputStream, delta)`.

I checked the Hadoop 3.4.0 implementation before changing it: when `skip()` returns `0`, `IOUtils.skipFully` attempts a single-byte `read()` and throws `EOFException` if that returns `-1`, so it avoids the zero-skip loop and gives the expected prompt EOF signal for seek-past-end.

The regression test now verifies that path by asserting `EOFException`, one `skip()` call, and one EOF `read()` call. I also rebased the branch onto latest `apache/main`; current branch head is `e324d87e`, and compare reports `behind_by=0`, `ahead_by=1`.

Local verification:
- `JAVA_HOME=/Users/cc/Library/Java/JavaVirtualMachines/corretto-18.0.2/Contents/Home mvn -pl fluss-filesystems/fluss-fs-hadoop -am -Dtest=HadoopDataInputStreamTest#testSeekPastEndOfStreamThrowsEOFException -DfailIfNoTests=false test`
- `JAVA_HOME=/Users/cc/Library/Java/JavaVirtualMachines/corretto-18.0.2/Contents/Home mvn -pl fluss-filesystems/fluss-fs-hadoop -am -Dtest=HadoopDataInputStreamTest -DfailIfNoTests=false test`
- `JAVA_HOME=/Users/cc/Library/Java/JavaVirtualMachines/corretto-18.0.2/Contents/Home mvn -pl fluss-filesystems/fluss-fs-hadoop -am -DfailIfNoTests=false test`
- `git diff --check upstream/main...HEAD`

litiliu:
looks good to me

PR #3743, opened 2026-07-22
[server] Support DescribeTabletServers API for safe scaling and upgrades

### LINKED ISSUE #3570: [server] Support DescribeTabletServers API for safe scaling and rolling upgrades
### Search before asking

### Motivation

When scaling a Fluss cluster, a `TabletServer` must not be removed while it still hosts replicas. Since terminating a non-empty server causes under-replication or data unavailability.

Operational tooling (e.g, Fluss Kubernetes Operator, see the FIP discussion thread: https://lists.apache.org/thread/p5sldj4v79t1p6yw5dr52b2j3579df5r) needs to verify that a target server is empty before allowing its termination, and to report per-server tablet load in cluster status.

Today this information is coordinator-internal only: 

- `CoordinatorContext.replicasOnTabletServer()` knows the answer, but no client-visible API exposes it.
- `Admin.getServerNodes()` lists tablet servers without their replica load, and `Admin.getClusterHealth()` (#3399) returns only cluster-wide aggregates.

An external controller therefore cannot distinguish an evacuated server from one still hosting replicas.

### Proposed Solution

Add a read-only `DescribeTabletServers` RPC to the Coordinator, following the exact pattern of `GetClusterHealth` (#3399 / PR #3400):

- New proto request/response messages.
- New method on `AdminReadOnlyGateway`, handled by `CoordinatorService`.
- A coordinator handler that tallies replicas per server from in-memory `CoordinatorContext` state.
- A new `Admin.describeTabletServers()` (`@PublicEvolving`) returning, per `TabletServer`, the same four counters getClusterHealth() reports cluster-wide, scoped to that server: numReplicas, inSyncReplicas, numLeaderReplicas, activeLeaderReplicas. These back the operator's `serverGreen(serverId)` predicate (rolling-upgrade gate), the scale-in safety gate (numReplicas == 0), and per-server tablet load in cluster status.

This is a small, low-risk read API: computed from existing in-memory coordinator state, no changes to the network protocol beyond the new RPC, no log/KV format changes, and no new write paths.

This is the only server-side dependency of the Kubernetes Operator FIP's v1alpha1 milestone.

### Willingness to contribute

morazow:
## Summary

- Adds a read-only `DescribeTabletServers` RPC to the Coordinator, following the pattern of `GetClusterHealth` (#3399 / #3400).
- New `Admin.describeTabletServers()` (`@PublicEvolving`) returns, per tablet server, the same four counters `getClusterHealth()` reports cluster-wide, scoped to that server: `numReplicas`, `inSyncReplicas`, `numLeaderReplicas`, `activeLeaderReplicas`.
- Counters are computed from in-memory `CoordinatorContext` state on the coordinator event thread (via `AccessContextEvent`); tablet servers forward the request to the coordinator over the internal listener.
- Live and shutting-down servers are always reported (even with zero replicas), and dead servers still holding assigned replicas remain visible, so a scale-in gate cannot treat them as drained.
- This backs the Kubernetes Operator FIP's `serverGreen(serverId)` rolling-upgrade gate, the scale-in safety gate (`numReplicas == 0`), and per-server tablet load reporting.

Fixes #3570

## Test Plan

- New unit test `DescribeTabletServersTest` (fluss-server) covering: empty cluster, full ISR, out-of-ISR replicas, inactive leaders, leaderless buckets, evacuated servers, dead-but-assigned servers, partitioned tables, and reconciliation of per-server sums with `computeClusterHealth`.
- New ITCase `FlussAdminITCase#testDescribeTabletServersDuringRollingUpgrade`: healthy cluster is green and sums match cluster health; a stopped server stays reported, non-empty and non-green; after restart all servers become green again.
- `mvn spotless:check` clean on fluss-rpc, fluss-server, fluss-client, fluss-flink-common.

### DIFF
--- a/fluss-client/src/test/java/org/apache/fluss/client/admin/FlussAdminITCase.java
+++ b/fluss-client/src/test/java/org/apache/fluss/client/admin/FlussAdminITCase.java
@@ -2870,7 +2870,16 @@ class FlussAdminITCase extends ClientToServerITCaseBase {
         waitAllReplicasReady(tableId, 3);
 
         // Phase 1: Cluster is healthy - every live server is reported, hosts replicas of the
-        // created table (replication factor 3 on 3 servers) and is green.
+        // created table (replication factor 3 on 3 servers) and is green. The cluster is shared
+        // with other tests, so wait until residue from them (e.g. a recovering ISR) has settled
+        // before taking the snapshot asserted below.
+        waitUntil(
+                () ->
+                        admin.describeTabletServers().get().stream()
+                                .allMatch(FlussAdminITCase::isServerGreen),
+                Duration.ofMinutes(1),
+                "All tablet servers should be green before the rolling upgrade starts");
+
         List<TabletServerDescription> servers = admin.describeTabletServers().get();
         assertThat(servers).extracting(TabletServerDescription::getServerId).contains(0, 1, 2);
         for (TabletServerDescription server : servers) {
--- a/fluss-client/src/test/java/org/apache/fluss/client/admin/FlussAdminITCase.java
+++ b/fluss-client/src/test/java/org/apache/fluss/client/admin/FlussAdminITCase.java
@@ -2860,4 +2860,85 @@ class FlussAdminITCase extends ClientToServerITCaseBase {
         assertThat(afterRecovery.getNumLeaderReplicas())
                 .isEqualTo(afterRecovery.getActiveLeaderReplicas());
     }
+
+    @Test
+    void testDescribeTabletServersDuringRollingUpgrade() throws Exception {
+        TablePath tablePath = TablePath.of("test_db", "describe_tablet_servers_table");
+        TableDescriptor tableDescriptor =
+                TableDescriptor.builder().schema(DEFAULT_SCHEMA).distributedBy(3, "id").build();
+        long tableId = createTable(tablePath, tableDescriptor, true);
+        waitAllReplicasReady(tableId, 3);
+
+        // Phase 1: Cluster is healthy - every live server is reported, hosts replicas of the
+        // created table (replication factor 3 on 3 servers) and is green.
+        List<TabletServerDescription> servers = admin.describeTabletServers().get();
+        assertThat(servers).extracting(TabletServerDescription::getServerId).contains(0, 1, 2);
+        for (TabletServerDescription server : servers) {
+            assertThat(server.getNumReplicas()).isGreaterThan(0);
+            assertThat(isServerGreen(server)).isTrue();
+        }
+
+        // The per-server counters must sum up to the cluster-wide health counters.
+        ClusterHealth health = admin.getClusterHealth().get();
+        assertThat(servers.stream().mapToInt(TabletServerDescription::getNumReplicas).sum())
+                .isEqualTo(health.getNumReplicas());
+        assertThat(servers.stream().mapToInt(TabletServerDescription::getInSyncReplicas).sum())
+                .isEqualTo(health.getInSyncReplicas());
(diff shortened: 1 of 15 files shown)


affo (approved):
Matches the expectations for unlocking a detailed description of the tablet server and assessing a server health on a per-server basis.

Thanks for this contribution!

affo (commented):
@morazow and the community.
I also noticed that there is no way as of now to list the tags of servers, I do wonder if would be the right place to do that in your opinion :)

morazow:
@affo good idea! So you think it makes sense as follow up PR or fine to add here?

affo:
@morazow I would directly include that in this PR, as semantically it would make sense as we "describe" tablet servers. But I would also like to have an additional feedback on this one as well 🤝

fluss-client/src/test/java/org/apache/fluss/client/admin/FlussAdminITCase.java
```diff
@@ -2860,4 +2868,94 @@ void testClusterHealthDuringRollingUpgrade() throws Exception {
...
+                "All tablet servers should be green before the rolling upgrade starts");
+
+        List<TabletServerDescription> servers = admin.describeTabletServers().get();
+        assertThat(servers).extracting(TabletServerDescription::getServerId).contains(0, 1, 2);
+        for (TabletServerDescription server : servers) {
+            assertThat(server.getNumReplicas()).isGreaterThan(0);
+            assertThat(isServerGreen(server)).isTrue();
+        }
+
+        // The per-server counters must sum up to the cluster-wide health counters.
```
fresh-borzoni, line 2898: A leaderless bucket is counted by cluster health but by no server here, so these sums can differ while every server still looks green. The waitUntil above doesn't rule that out. testPerServerSumsReconcileWithClusterHealth already covers this. 
Do we need it?

fresh-borzoni (commented):
@morazow Thank you for the PR!
Reading this alongside #1410 and #4061, the split looks like: 
[#1410](https://github.com/apache/fluss/pull/1410) DescribeCluster for topology + ServerTag, getClusterHealth for aggregates, this one for per-server counters
https://github.com/apache/fluss/pull/4061 the cache both health APIs read from.

If it holds, two things follow here: the tag commit is redundant with #1410, and both PRs currently claim ApiKey 1065 so one needs renumbering.

I'd hold the API-shape questions until that settles rather than have you redo work.

let me know when it settles, so I can help with API-shape

affo:
@fresh-borzoni at issue #1389 we conducted quite a big discussion also on the performance of this call given that it runs in the coordinator loop itself.

This led me to shape in draft a PR for an health cache that can be used by both cluster health and per-tablet health: https://github.com/apache/fluss/pull/4061.

The cache uses a new mechanism thought for the performance updates, a new CoalescingRefreshCache class that embeds the logic for no-read-lock, but write lock, and enhances the logic to allow for different dirty states to optimize heavy updates on the coordinator side.

The same mechanism can be re-used across caches as well.

What would you guys think (@fresh-borzoni , @morazow )?

In case that one seems like a good approach we can:
 - rebase this one and wire the cache mechanism
 - on the second thing for server tags, yeah we already agreed on that issue that that should stay in `getServerNodes`, my only point was not to perform a direct ZK blocking call but to re-use the cache.

----

I re-read your comment above and realized you are aware of #4061 🤝 😅 , I misread at the beginning.
If you may also give feedback to that one now that you have the whole context it would be amazing 🚀

morazow:
Thanks @affo , @fresh-borzoni for review!

It would make sense to wait for #4061 and rebase to use the new cache mechanism.

I'll also drop the serverTags commit from this PR 🤝

PR #3751, opened 2026-07-23
[server] add support of `KvPreWriteBuffer` backpressure

### LINKED ISSUE #3532: [server] add the metric of KV preWriteBuffer
### Search before asking

### Motivation

Under high write throughput on a single tablet server, if replication falls behind, the KV preWriterBuffer may be unable to release buffered data because the high watermark has not advanced, eventually leading to OOM on the tablet server.

for the above case, we should do the following ways to eliminate risk
1. introduce the metric to observe
2. introduce the backpressure mechanism to limit client writing speed.

### Solution

_No response_

### Anything else?

_No response_

### Willingness to contribute

zuston:
### Purpose

Linked issue: close #3532

Without a backpressure mechanism for `KvPreWriteBuffer` , the tablet server is at risk of OOM, particularly when ingestion throughput is high but replication is slow. A similar issue can occur during recovery: if the follower is not yet ready while the leader continues accepting client writes, buffered data may accumulate rapidly, causing the leader tablet server to run out of memory and exit.

### Brief change log

1. Introduce a global `KvPreWriteBuffer` memory pool for each tablet server. When memory usage reaches the configured high watermark, incoming client writes are temporarily rejected and retried until usage falls below the low watermark.

### DIFF
--- a/fluss-server/src/main/java/org/apache/fluss/server/kv/prewrite/KvMemoryConsumer.java
+++ b/fluss-server/src/main/java/org/apache/fluss/server/kv/prewrite/KvMemoryConsumer.java
@@ -1,0 +1,82 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.fluss.server.kv.prewrite;
+
+import org.apache.fluss.annotation.Internal;
+
+import javax.annotation.concurrent.NotThreadSafe;
+
+import static org.apache.fluss.utils.Preconditions.checkArgument;
+import static org.apache.fluss.utils.Preconditions.checkNotNull;
+import static org.apache.fluss.utils.Preconditions.checkState;
+
+/** A per-buffer consumer of the TabletServer-wide KV pre-write-buffer memory guard. */
+@Internal
+@NotThreadSafe
+abstract class KvMemoryConsumer {
+
+    private final KvPreWriteBufferMemoryManager memoryManager;
+    private long acquiredBytes;
+
+    protected KvMemoryConsumer(KvPreWriteBufferMemoryManager memoryManager) {
+        this.memoryManager = checkNotNull(memoryManager, "Memory manager must not be null.");
+    }
+
+    protected final boolean tryAcquireMemory(long bytes) {
+        checkArgument(bytes >= 0, "The number of bytes to acquire must not be negative.");
+        if (!memoryManager.isEnabled()) {
+            return true;
+        }
+        if (!memoryManager.tryReserve(bytes)) {
+            return false;
+        }
+        acquiredBytes += bytes;
+        return true;
+    }
+
+    protected final void freeMemory(long bytes) {
+        checkArgument(bytes >= 0, "The number of bytes to free must not be negative.");
+        if (!memoryManager.isEnabled()) {
+            return;
+        }
+        checkState(
+                bytes <= acquiredBytes,
+                "Cannot free %s bytes when this consumer has only acquired %s bytes.",
+                bytes,
+                acquiredBytes);
+        if (bytes > 0) {
+            memoryManager.release(bytes);
+            acquiredBytes -= bytes;
+        }
+    }
+
+    protected final void freeAllMemory() {
+        if (acquiredBytes > 0) {
+            memoryManager.release(acquiredBytes);
+            acquiredBytes = 0;
+        }
+    }
+
+    protected final boolean isMemoryGuardEnabled() {
+        return memoryManager.isEnabled();
+    }
+
+    protected final KvPreWriteBufferMemoryManager memoryManager() {
+        return memoryManager;
+    }
+}
--- a/fluss-server/src/main/java/org/apache/fluss/server/kv/prewrite/KvPreWriteBuffer.java
+++ b/fluss-server/src/main/java/org/apache/fluss/server/kv/prewrite/KvPreWriteBuffer.java
@@ -18,6 +18,8 @@
 package org.apache.fluss.server.kv.prewrite;
 
 import org.apache.fluss.annotation.VisibleForTesting;
+import org.apache.fluss.exception.RecordTooLargeException;
+import org.apache.fluss.exception.StorageBackpressureException;
 import org.apache.fluss.memory.MemorySegment;
 import org.apache.fluss.metrics.Counter;
 import org.apache.fluss.record.ChangeType;
@@ -87,7 +89,11 @@ import static org.apache.fluss.utils.UnsafeUtils.BYTE_ARRAY_BASE_OFFSET;
  * head to tail, it will stop flush.
  */
 @NotThreadSafe
-public class KvPreWriteBuffer {
+public class KvPreWriteBuffer extends KvMemoryConsumer implements AutoCloseable {
+
+    // A conservative estimate of heap retained by a KvEntry, its wrappers, the linked-list node,
+    // the hash-map entry, and associated object/array headers. Payload bytes are added separately.
+    private static final long ENTRY_OVERHEAD_BYTES = 128L;
 
     // a mapping from the key to the kv-entry
     private final Map<Key, KvEntry> kvEntryMap = new HashMap<>();
@@ -105,7 +111,16 @@ public class KvPreWriteBuffer {
     // Accumulated byte size of entries not yet completed by a flush.
     private long pendingFlushBytes = 0;
 
+    /** Creates a pre-write buffer with the defensive memory guard disabled. */
     public KvPreWriteBuffer(TabletServerMetricGroup serverMetricGroup) {
+        this(serverMetricGroup, KvPreWriteBufferMemoryManager.disabled());
+    }
+
+    /** Creates a pre-write buffer backed by the TabletServer-wide memory guard. */
+    public KvPreWriteBuffer(
+            TabletServerMetricGroup serverMetricGroup,
+            KvPreWriteBufferMemoryManager memoryManager) {
+        super(memoryManager);
         truncateAsDuplicatedCount = serverMetricGroup.kvTruncateAsDuplicatedCount();
         truncateAsErrorCount = serverMetricGroup.kvTruncateAsErrorCount();
     }
@@ -146,6 +161,8 @@ public class KvPreWriteBuffer {
                             + ", but the new log sequence number is "
                             + lsn);
         }
+
+        reserveMemory(key, value);
 
         // create the kv entry with previous pointer if exists, and put the new entry to the map
         KvEntry kvEntry =
@@ -205,6 +222,7 @@ public class KvPreWriteBuffer {
                                 + targetLogSequenceNumber);
             }
             pendingFlushBytes -= entryBytes(entry.getKey(), entry.getValue());
+            releaseMemory(entry);
             boolean removed = kvEntryMap.remove(entry.getKey(), entry);
             // if the latest entry is removed, we need to rollback the previous entry to the map
             if (removed) {
@@ -260,6 +278,7 @@ public class KvPreWriteBuffer {
             }
             entry.state = EntryState.FLUSHED;
             pendingFlushBytes -= entryBytes(entry.getKey(), entry.getValue());
+            releaseMemory(entry);
             kvEntryMap.remove(entry.getKey(), entry);
         }
         if (allKvEntries.isEmpty()) {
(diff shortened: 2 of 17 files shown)


zuston:
ptal @fresh-borzoni

beryllw:
@platinumhamburg Could you take a look when you get a chance?

luoyuxia:
I fell like #3463 can solve this backpressure problem

fluss-common/src/main/java/org/apache/fluss/exception/InsufficientKvPreWriteBufferException.java
```diff
@@ -0,0 +1,43 @@
...
+
+package org.apache.fluss.exception;
+
+import org.apache.fluss.annotation.PublicEvolving;
+
+/**
+ * The tablet server rejected a KV write because its global pre-write buffer high watermark was
+ * reached.
+ *
+ * <p>The client may retry the write after a backoff. The rejected write has not been appended to
```
wuchong: Thanks for working on this. We also have [#3463](https://github.com/apache/fluss/pull/3463), which implements cooperative KV backpressure based on RocksDB L0 metrics. Could you try it and see whether it also addresses the problem described here?

In general, backpressure needs client-side cooperation—for example, retry with exponential backoff—rather than only returning a retriable error from the server. #3463 introduces the server-client throttling/backoff mechanism, so it may be better to let #3463 land first and then build this change on top of it.

My understanding is that the main difference between the two PRs is the source of the pressure signal: #3463 derives pressure from the RocksDB L0 file count, while this PR derives it from `KvPreWriteBuffer` memory usage to prevent OOM. If so, could this PR reuse the mechanism introduced by #3463 and add pre-write-buffer memory pressure as another backpressure source?

zuston: Thanks, agreed. I reviewed #3463, and I agree that its cooperative server-client throttling should be the primary backpressure mechanism.

This PR is intended as a defensive, last-resort safeguard rather than a replacement for #3463. While #3463 derives pressure per KvTablet from RocksDB L0/flush conditions, this PR limits the TabletServer-wide aggregate heap retained by all KvPreWriteBuffers. It protects the server from OOM in exceptional cases—for example, when replication stalls, clients do not react quickly enough, or memory accumulates across many tablets.

Now that #3463 has landed, I’ll rebase this PR and reuse its cooperative path: expose pre-write-buffer memory usage as an additional pressure source, combine it with the existing RocksDB pressure, and use StorageBackpressureException for the final hard-rejection guard. The global high watermark will therefore act only as the last line of defense, while normal traffic regulation remains handled by #3463’s client-side throttling mechanism.

zuston: One concern is that enforcing a global memory limit could introduce a performance regression. Could each bucket acquire memory in chunks or batches instead of requesting it for every entry? WDYT @wuchong

zuston: After reviewing the entire putAsLeader pipeline, I think it may be worth extending this PR to make its on-heap memory usage bounded and more predictable by allocating memory through a segment pool similar to `LazyMemorySegmentPool` .

zuston:
Could you help review this? @wuchong @platinumhamburg

zuston:
@fresh-borzoni Could you help review this if you have time?

fluss-server/src/main/java/org/apache/fluss/server/kv/prewrite/KvPreWriteBuffer.java
```diff
@@ -260,6 +278,7 @@ public int completeFlush(PreparedFlush preparedFlush) {
             }
             entry.state = EntryState.FLUSHED;
             pendingFlushBytes -= entryBytes(entry.getKey(), entry.getValue());
+            releaseMemory(entry);
```
platinumhamburg, line 281: Calling releaseMemory(entry) here does not necessarily mean that the entry is no longer retained by the buffer.

For repeated updates to the same key, a newer KvEntry may still reference this flushed entry through previousEntry. In that case, manager.usedBytes decreases while the old entry, key, value, and payload remain reachable. The guard may therefore admit more writes than the buffer's actual retained memory allows, undermining its primary OOM-protection goal.

Could we make the reservation lifetime follow ownership instead: release the reservation only after the entry has been detached from every persistent buffer structure? Making the same-key version chain capable of unlinking flushed versions is one possible solution, but the exact implementation does not need to be prescribed here.

The same ownership rule should also apply to close(): it currently releases all reservations without first clearing the map, list, and version links.

fluss-server/src/main/java/org/apache/fluss/server/kv/prewrite/KvMemoryConsumer.java
```diff
@@ -0,0 +1,82 @@
...
+import javax.annotation.concurrent.NotThreadSafe;
+
+import static org.apache.fluss.utils.Preconditions.checkArgument;
+import static org.apache.fluss.utils.Preconditions.checkNotNull;
+import static org.apache.fluss.utils.Preconditions.checkState;
+
+/** A per-buffer consumer of the TabletServer-wide KV pre-write-buffer memory guard. */
+@Internal
+@NotThreadSafe
+abstract class KvMemoryConsumer {
```
platinumhamburg, line 31: I have a concern about the KvMemoryConsumer abstraction. KvPreWriteBuffer is currently its only subclass, and KvPreWriteBuffer is the only component that knows when an entry enters or leaves the map, the LSN-ordered list, and the same-key version chain.

Keeping acquiredBytes in the superclass separates object ownership from the reservation lifecycle. It also makes it easier to release memory based on a call site rather than on whether the entry has actually been detached.

If there is no concrete second consumer yet, would it be simpler for KvPreWriteBuffer to compose KvPreWriteBufferMemoryManager directly and maintain its own local reserved bytes? A shared abstraction could be extracted later when an actual reuse requirement appears.

fluss-server/src/main/java/org/apache/fluss/server/kv/prewrite/KvPreWriteBuffer.java
```diff
@@ -289,10 +308,55 @@ public void abortAllPrepared() {
...
+        }
+        long memorySize = estimatedRetainedBytes(key, value);
+        if (memorySize > memoryManager().highWatermarkBytes()) {
+            throw new RecordTooLargeException(
+                    String.format(
+                            "The KV entry requires %s bytes, which exceeds the global KV "
+                                    + "pre-write-buffer high watermark of %s bytes.",
+                            memorySize, memoryManager().highWatermarkBytes()));
+        }
+        if (!tryAcquireMemory(memorySize)) {
```
platinumhamburg, line 342: This hard admission failure also applies to recovery. KvRecoverHelper rebuilds the uncommitted HW-to-LEO suffix through putToPreWriteBuffer(), so the StorageBackpressureException thrown here can abort tablet recovery.

A rejected client write can back off and wait for normal flush progress. Recovery does not have the same progress mechanism, so retrying with the same suffix and the same memory budget may repeatedly produce the same failure.

Recovery still needs the pre-write buffer under the current semantics, so simply bypassing the buffer is not the answer. Could we define a separate, bounded recovery admission strategy and explain how it guarantees progress when the HW-to-LEO suffix does not fit into the budget? A recovery test with a suffix larger than the budget would also be important.

fluss-server/src/main/java/org/apache/fluss/server/kv/prewrite/KvPreWriteBufferMemoryManager.java
```diff
@@ -0,0 +1,234 @@
...
+            long usedBytes = usedBytes(currentState);
+            if (isUnderPressure(currentState)) {
+                return false;
+            }
+            long remainingBytes = highWatermarkBytes - usedBytes;
+            if (requiredBytes > remainingBytes) {
+                // Do not latch pressure at or below the low watermark. Otherwise, a single large
+                // reservation could block all writes even though no release is required to cross
+                // the resume threshold.
+                long nextState =
```
platinumhamburg, line 113: Is it necessary for a failed reservation to change the global pressure latch?

For example, with low=80, high=100, used=81, and required=20, this request is rejected and actual usage remains 81. However, the manager becomes latched under pressure, so even a subsequent 1-byte request is rejected until existing usage falls to 80.

This allows one relatively large request to block smaller writes across all tablets while actual usage is still below the high watermark. I would expect this request to be rejected independently while global pressure continues to reflect actual reserved usage. If this behavior is intentional, could we document the rationale and the intended admission contract?

fluss-server/src/test/java/org/apache/fluss/server/kv/prewrite/KvPreWriteBufferTest.java
```diff
@@ -454,6 +456,93 @@ void testPendingFlushBytesTracking() {
...
+                .isInstanceOf(StorageBackpressureException.class);
+
+        assertThat(buffer.getMaxLSN()).isEqualTo(1);
+        assertThat(buffer.getAllKvEntries()).hasSize(1);
+        assertThat(buffer.getKvEntryMap()).hasSize(1);
+        assertThat(manager.usedBytes()).isEqualTo(138);
+    }
+
+    @Test
+    void testMemoryReleasedOnlyAfterFlushCompletion() {
```
platinumhamburg, line 512: This test uses two different keys, so it only verifies that the manager's counter changes. It cannot detect the case where a newer version of the same key still retains flushed predecessors through previousEntry.

Could we add a same-key, multi-version partial-flush scenario—for example, append e1, e2, and e3 for the same key, flush only e1/e2, and verify that:

1. The surviving entry no longer references the flushed history.
2. Reservations cover only entries still owned by the buffer.
3. Truncating the remaining version leaves the map, list, and accounting empty.

Checking usedBytes alone is insufficient here because the implementation and the test can share the same incorrect release assumption.

fluss-server/src/test/java/org/apache/fluss/server/kv/KvTabletTest.java
```diff
@@ -1593,6 +1617,38 @@ void testPutAsLeaderRejectsCurrentBatchThatExceedsFlushBudget() throws Exception
         assertThat(logTablet.localLogEndOffset()).isEqualTo(0L);
     }
 
+    @Test
+    void testPreWriteBufferMemoryPressureAndCloseRelease() throws Exception {
```
platinumhamburg, line 1621: Calling memoryManager.tryReserve() directly injects pressure into the test, so this primarily verifies the wiring of KvTablet.currentPressure() rather than production write behavior.

Could we generate pressure entirely through the real path instead:

putAsLeader()
  -> KvPreWriteBuffer reservation
  -> admission rejection at the limit
  -> complete rollback of the current batch
  -> scheduled flush releases the reservation
  -> a subsequent putAsLeader() succeeds

The test should also verify that the rejected batch leaves no residual changes in the WAL offset, buffer contents, or usedBytes. Using a controllable scheduler is appropriate, but directly manipulating the manager should not replace the real admission path.

platinumhamburg (changes requested):
Overall, using a TabletServer-wide memory budget shared by all KV pre-write buffers looks like a reasonable direction. The role of KvPreWriteBufferMemoryManager as the global accounting and pressure-reporting point also makes sense.

Before going further into implementation details, however, I think this PR should first tighten two core contracts:

1. The reservation lifetime must match the lifetime of entries actually retained by KvPreWriteBuffer.
2. Recovery needs a separate, bounded admission strategy that guarantees progress rather than directly reusing the hard-rejection path for client writes.

The current implementation can reduce accounted usage while older versions of the same key are still reachable, and it can repeatedly fail while recovering the same HW-to-LEO suffix. I suggest addressing these ownership and progress properties, together with behavior-level tests, before proceeding with a more detailed review.

Also, KvMemoryConsumer moves the per-buffer reservation state into an abstract superclass, while KvPreWriteBuffer is the only component that knows when entries are attached and detached. Composing the manager and local reservation state directly in KvPreWriteBuffer would make the ownership boundary clearer than inheritance here.

platinumhamburg:
Hi @zuston , do you still plan to continue working on this PR?
After giving this some more thought, I don't think we should couple the backpressure signal with the memory soft limit here. We could simplify this PR by implementing only a soft limit that rejects new writes when the limit is reached, without introducing a backpressure mechanism.
In my view, the actual backpressure signal should be based on overall server memory pressure, because the memory footprint of a write is not limited to the prewrite buffer—it is also affected by other memory consumers along the write path.
If you don't plan to continue with this PR, please let me know. We can close it and move the soft-limit/rejection part forward separately.

zuston:
@platinumhamburg thanks for your notice and suggestions. 

This limitation is necessary for our case, so I will follow your suggestions  to work continuely. 

> We could simplify this PR by implementing only a soft limit that rejects new writes when the limit is reached, without introducing a backpressure mechanism

Sounds good, this will be more simple. 

> In my view, the actual backpressure signal should be based on overall server memory pressure, because the memory footprint of a write is not limited to the prewrite buffer—it is also affected by other memory consumers along the write path.

How to monitor the memory usage? Just using the java on-heap memory usage?

platinumhamburg:
> @platinumhamburg thanks for your notice and suggestions.
> 
> This limitation is necessary for our case, so I will follow your suggestions to work continuely.
> 
> > We could simplify this PR by implementing only a soft limit that rejects new writes when the limit is reached, without introducing a backpressure mechanism
> 
> Sounds good, this will be more simple.
> 
> > In my view, the actual backpressure signal should be based on overall server memory pressure, because the memory footprint of a write is not limited to the prewrite buffer—it is also affected by other memory consumers along the write path.
> 
> How to monitor the memory usage? Just using the java on-heap memory usage?

I think we should consider the full memory footprint — including heap, non-heap, and even any memory that could potentially cause OOM — as a unified pressure signal.

zuston:
> I think we should consider the full memory footprint — including heap, non-heap, and even any memory that could potentially cause OOM — as a unified pressure signal.

This is somewhat challenging because memory usage can be unpredictable, making it difficult to set an appropriate threshold.

For example, enabling historical partition lookup can cause memory consumption to increase rapidly, while imposing an on-heap memory limit may negatively affect the upper streaming pipeline.

platinumhamburg:
> > I think we should consider the full memory footprint — including heap, non-heap, and even any memory that could potentially cause OOM — as a unified pressure signal.
> 
> This is somewhat challenging because memory usage can be unpredictable, making it difficult to set an appropriate threshold.
> 
> For example, enabling historical partition lookup can cause memory consumption to increase rapidly, while imposing an on-heap memory limit may negatively affect the upper streaming pipeline.

I don’t think these are contradictory. The primary goal of memory backpressure is to move the system away from the risk of OOM. If historical lookups have already pushed memory usage to a critical level, blocking upstream traffic is expected behavior—it prevents the system from entering an even more dangerous state.
At the same time, we also need to prevent historical lookups from consuming an unbounded amount of memory at the expense of other parts of the system.
Ultimately, I believe we will have two parallel mechanisms: a memory-watermark-based backpressure system that prevents the system from reaching dangerous memory levels, and a ratio-based memory allocation and accounting system that distributes memory appropriately among subsystems while usage remains within safe limits. @wuchong 
What do you think?

PR #3752, opened 2026-07-23
[server] fix zk notification purge race

### LINKED ISSUE #3753: [Bug] Concurrent ZK notification cleanup throws NoSuchElementException
### Search before asking

### Fluss version

0.9.0 (latest release)

#### Description

In a multi-server cluster, updating dynamic configs may produce the following error after the configs have been applied successfully:

```text
ERROR ZkNodeChangeNotificationWatcher - Error while purging obsolete notification change for path = /config/changes/config_change_0000000007
java.util.NoSuchElementException: No value present
    at java.util.Optional.get(Optional.java:143)
    at org.apache.fluss.server.authorizer.ZkNodeChangeNotificationWatcher.purgeObsoleteNotifications(ZkNodeChangeNotificationWatcher.java:142)
```

Image

#### Root cause
Each watcher reads a snapshot of all children under /config/changes and independently purges expired nodes. After one watcher deletes a node, another watcher may still have it in its children snapshot.
getStat() then returns Optional.empty(), but the result is accessed directly with Optional.get():
```
Stat state = zooKeeperClient.getStat(notificationNode).get();
```

### Solution

Do empty check before get the state.

gyang94:
### Purpose

Linked issue: close #3753 

### Brief change log

empty check for zk stats

### DIFF
--- a/fluss-server/src/main/java/org/apache/fluss/server/authorizer/ZkNodeChangeNotificationWatcher.java
+++ b/fluss-server/src/main/java/org/apache/fluss/server/authorizer/ZkNodeChangeNotificationWatcher.java
@@ -139,8 +139,14 @@ public class ZkNodeChangeNotificationWatcher {
         for (String notification : sortedNotifications) {
             String notificationNode = seqNodeRoot + "/" + notification;
             try {
-                Stat state = zooKeeperClient.getStat(notificationNode).get();
-                if (now - state.getCtime() >= changeExpirationMs) {
+                Optional<Stat> state = zooKeeperClient.getStat(notificationNode);
+                if (!state.isPresent()) {
+                    LOG.debug(
+                            "Notification {} has already been purged by another watcher",
+                            notificationNode);
+                    continue;
+                }
+                if (now - state.get().getCtime() >= changeExpirationMs) {
                     LOG.debug("Purging change notification {}", notificationNode);
                     zooKeeperClient.deletePath(notificationNode);
                 }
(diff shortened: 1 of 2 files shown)


fresh-borzoni (commented):
@gyang94 Thank you, fix is sound, some ideas about the test below, PTAL

The test passes even with the fix reverted because our test config has the root logger OFF, so the appender never gets the ERROR and noneMatch on an empty list is always true.
So a couple of suggestions:
1. mockito isn't really needed: the handler runs before the purge and they share the children list, so a handler that deletes the notification node gives the purge a stale list, the exact race from the issue.
a sketch:
```
 TestingNotificationHandler handler =
         new TestingNotificationHandler() {
             @Override
             public void processNotification(byte[] notification) {
                 super.processNotification(notification);
                 deleteAllNotificationNodes(); // purge now sees a stale list
             }
         };
```
 2. with root OFF the appender only works through its own LoggerConfig, smth like this:
```java
LoggerContext ctx = (LoggerContext) LogManager.getContext(false);
Configuration cfg = ctx.getConfiguration();
LoggerConfig lc = new LoggerConfig(ZkNodeChangeNotificationWatcher.class.getName(), Level.DEBUG, false);
lc.addAppender(appender, Level.DEBUG, null);
cfg.addLogger(lc.getName(), lc);
ctx.updateLoggers();
```
and assert the debug message is present rather than the error absent, so it can't go vacuous again.

fresh-borzoni (approved):
@gyang94 Thank you, LGTM 👍

PR #3756, opened 2026-07-23
[rpc] Bound length-delimited field length against readable bytes

### LINKED ISSUE #3757: [rpc] RPC decoder allocates before validating field length
### Search before asking

### Description

The generated protobuf field decoders read a length-delimited field's length prefix and allocate for it before checking that many bytes are actually present. 

So a small request with a large length prefix makes the server allocate a big array on the Netty event-loop
thread before it reads.

### Willingness to contribute

fresh-borzoni:
closes https://github.com/apache/fluss/issues/3757
Generated Protobuf decoders allocate `new byte[len]` for a length-delimited field before checking the declared length fits the bytes present, so a bogus length prefix results in  a large allocation. 

Added `ProtoCodecUtils.readBytesLen`, which rejects a length that is negative or larger than the buffer's readable bytes, and route the bytes, string, repeated-bytes and repeated-string decoders through it.

### DIFF
--- a/fluss-common/src/main/java/org/apache/fluss/utils/ProtoCodecUtils.java
+++ b/fluss-common/src/main/java/org/apache/fluss/utils/ProtoCodecUtils.java
@@ -178,6 +178,20 @@ public final class ProtoCodecUtils {
         return result;
     }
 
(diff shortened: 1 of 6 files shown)


wuchong (approved):
Thanks @fresh-borzoni , LGMT.

PR #3771, opened 2026-07-26
[client][rust] Align fixed-schema scans by column ID

### LINKED ISSUE #3770: [Improvement][rust] Align fixed-schema log reads by column ID
### Search before asking

### Motivation

The Rust fixed-schema log scanner currently aligns historical Arrow batches with the scanner schema by Arrow field name.

This differs from the KV `FixedSchemaDecoder`, which uses stable Fluss column IDs through `index_mapping`.

Name-based alignment is not robust for schema evolution:

- A column with the same ID but a different name may be treated as missing.
- A newly added column that reuses an old column name but has a different ID
  may incorrectly read the old column value.

Although the Fluss server does not currently support `RENAME COLUMN`, the Rust fixed-schema decoding path should consistently use column IDs as the stable column identity.

### Solution

Use Fluss column IDs when aligning historical Arrow batches with the fixed target schema:

- Preserve values when the source and target columns have the same ID.
- Fill the target column with NULL when its ID does not exist in the source
  schema.
- Do not match columns that only have the same name but different IDs.
- Apply the same behavior to local and remote log reads.
- Reuse the existing `index_mapping` logic used by `FixedSchemaDecoder`.

### Anything else?

This improvement does not change the public API or storage format.

### Willingness to contribute

slfan1989:
### Purpose

Linked issue: close #3770

The Rust fixed-schema log scanner previously aligned historical Arrow batches with the scanner schema by field name.

This behavior was inconsistent with the KV `FixedSchemaDecoder`, which uses stable Fluss column IDs. Name-based alignment may lose values when a column keeps the same ID but changes its name, or incorrectly reuse values when
different column IDs have the same name.

This PR aligns fixed-schema log reads by column ID, consistent with `FixedSchemaDecoder`.

### Brief change log

- Capture the fixed target Fluss schema when creating a fixed-schema log scanner.
- Build source-to-target mappings with the existing `index_mapping` utility,  using stable column IDs.
- Pass the mapping to Arrow `ReadContext` for both local and remote log reads.
- Preserve values when source and target columns have the same ID.
- Fill target columns with NULL when their IDs do not exist in the source schema.
- Prevent columns with the same name but different IDs from being incorrectly matched.
- Add regression tests for local and remote fixed-schema decoding.

### Tests

Added Rust unit tests.

### API and Format

### Documentation

### DIFF
--- a/fluss-rust/crates/fluss/src/client/table/log_fetch_buffer.rs
+++ b/fluss-rust/crates/fluss/src/client/table/log_fetch_buffer.rs
@@ -1380,8 +1380,15 @@ mod tests {
         row.set_field(0, 1_i32);
         row.set_field(1, "alice");
 
-        let batch = fetch_fixed_schema_batch(source_table_info, &target_table_info, &row, false)?;
-        assert_eq!(batch.column(1).null_count(), 1);
+        for is_remote in [false, true] {
+            let batch = fetch_fixed_schema_batch(
+                Arc::clone(&source_table_info),
+                &target_table_info,
+                &row,
+                is_remote,
+            )?;
+            assert_eq!(batch.column(1).null_count(), 1);
+        }
         Ok(())
     }
 
--- a/fluss-rust/crates/fluss/src/client/table/log_fetch_buffer.rs
+++ b/fluss-rust/crates/fluss/src/client/table/log_fetch_buffer.rs
@@ -984,13 +984,17 @@ mod tests {
         ArrowCompressionInfo, ArrowCompressionRatioEstimator, ArrowCompressionType,
         DEFAULT_NON_ZSTD_COMPRESSION_LEVEL,
     };
-    use crate::metadata::{DataField, DataTypes, PhysicalTablePath, RowType, TablePath};
+    use crate::metadata::{
+        Column, DataField, DataTypes, PhysicalTablePath, RowType, Schema, TableDescriptor,
+        TableInfo, TablePath,
+    };
     use crate::record::{
         APPEND_ONLY_FLAG_MASK, ATTRIBUTES_OFFSET, LENGTH_LENGTH, LENGTH_OFFSET, LOG_OVERHEAD,
         MemoryLogRecordsArrowBuilder, RECORDS_OFFSET, ReadContext, to_arrow_schema,
     };
     use crate::row::GenericRow;
     use crate::test_utils::{build_table_info, build_table_info_with_columns};
+    use arrow::array::{Array, StringArray};
     use std::sync::Arc;
 
     fn expect_data<T>(result: FetchResult<T>) -> T {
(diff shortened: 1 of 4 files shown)


slfan1989:
While reviewing the Rust Arrow fixed-schema read path, I noticed that historical data is currently aligned by field name rather than by stable column ID. Under typical schema-evolution semantics, this could prevent old data from being matched correctly after a column rename, so I initially considered it a bug.

However, after local testing and further investigation, I confirmed that Fluss currently supports only a limited set of schema-evolution operations, such as adding nullable columns. Although the client exposes a `RENAME COLUMN` interface, the corresponding functionality has not yet been implemented on the server. Therefore, this issue cannot currently be reproduced through the complete client-server workflow.

Nevertheless, the Arrow fixed-schema path is inconsistent with the KV `FixedSchemaDecoder` in how fields are aligned. To provide consistent semantics across both read paths and establish a reliable foundation for supporting additional schema-evolution operations in the future, I submit a PR that changes Arrow fixed-schema alignment from field names to stable column IDs.

fluss-rust/crates/fluss/src/client/table/scanner.rs
```diff
@@ -1234,16 +1234,17 @@ impl LogFetcher {
...
-            .with_fixed_schema(fixed_schema),
-        );
+        let mut resolver = ReadContextResolver::new(
+            initial_schema_id,
+            read_context,
+            remote_read_context,
+            projected_fields,
+        )
+        .with_schema_getter(Arc::clone(&schema_getter));
+        if fixed_schema {
```
fresh-borzoni, line 1244: with_fixed_schema runs even when projected_fields is Some, so fixed_target pairs a full-schema mapping with the projected arrow schema and row type from the initial context. Dead today since register_schema returns early under projection, but the two halves disagree and the only thing that would catch it is a per-batch Schema alignment has N indexes for K target fields. 

Mb skip the capture when projection is active? WDYT

fluss-rust/crates/fluss/src/record/arrow.rs
```diff
@@ -1707,33 +1717,53 @@ impl ReadContext {
 fn align_record_batch_to_schema(
     record_batch: RecordBatch,
     target_schema: SchemaRef,
+    schema_alignment: &[i32],
 ) -> Result<RecordBatch> {
-    if record_batch.schema_ref() == &target_schema {
-        return Ok(record_batch);
+    if schema_alignment.len() != target_schema.fields().len() {
```
fresh-borzoni: nit: This length check (and the usize::try_from + bounds check below) validates a per-context invariant on every batch. 
Should we assert it once in with_target_schema_alignment, next to the existing debug_assert!(self.projection.is_none()), and leave the it clean here?

fluss-rust/crates/fluss/src/client/table/log_fetch_buffer.rs
```diff
@@ -1222,6 +1315,76 @@ mod tests {
...
+            vec![
+                Column::new("id", DataTypes::int()).with_id(0),
+                Column::new("name", DataTypes::string()).with_id(2),
+            ],
+        )?;
+        let mut row = GenericRow::new(2);
+        row.set_field(0, 1_i32);
+        row.set_field(1, "alice");
+
+        let batch = fetch_fixed_schema_batch(source_table_info, &target_table_info, &row, false)?;
```
fresh-borzoni: nit: ..._preserves_renamed_column_by_id loops [false, true], this one is local-only. Same loop here?

fresh-borzoni (commented):
@slfan1989 Thank you, good catch, left a couple of comments

Since this is about making the arrow and KV read paths agree on how columns are identified - there's a third read path that doesn't handle schema_id at all: batch_scanner.rs:192 decodes every log batch with the current table schema, while decode_kv_batch right below builds a decoder per schema id. Doesn't a limit scan over pre-ADD-COLUMN batches fail there?

Can you check if it's a bug, and if it's what it looks like - file an issue/PR?

fluss-rust/crates/fluss/src/client/table/scanner.rs
slfan1989: Thanks for pointing this out. Since projection pins the resolver to the initial context, the fixed target is not used in this case. I updated with_fixed_schema to skip fixed_target capture when projection is active and added a regression test to cover this behavior.

fluss-rust/crates/fluss/src/record/arrow.rs
slfan1989: Thanks for the suggestion. I moved the alignment length, source-index bounds, and type invariant checks into `with_target_schema_alignment`. The per-batch alignment path now only applies the prevalidated mapping, while `RecordBatch::try_new` remains the final schema validation. I also added tests for invalid alignment lengths and source indexes.

fluss-rust/crates/fluss/src/client/table/log_fetch_buffer.rs
slfan1989: Updated this test to loop over `[false, true]`, so it now covers both local and remote read paths, consistent with `...preserves_renamed_column_by_id`.

slfan1989:
> @slfan1989 Thank you, good catch, left a couple of comments
> 
> Since this is about making the arrow and KV read paths agree on how columns are identified - there's a third read path that doesn't handle schema_id at all: batch_scanner.rs:192 decodes every log batch with the current table schema, while decode_kv_batch right below builds a decoder per schema id. Doesn't a limit scan over pre-ADD-COLUMN batches fail there?
> 
> Can you check if it's a bug, and if it's what it looks like - file an issue/PR?

@fresh-borzoni Thanks for pointing this out. I confirmed that this is indeed a bug. I reproduced it with an Arrow log batch written before an `ADD COLUMN`. 

The limit-scan log path decodes every batch using the current table schema, which fails when reading the older batch because the newly added field is absent.

The log batch already carries its `schema_id`, so this path should resolve the source schema per batch and align it with the current target schema, consistent with the regular log-fetch and KV paths. 

I’ll track this in a separate issue(#3808) and follow-up PR, and link them here once created.

fresh-borzoni (approved):
@slfan1989 Thank you, LGTM 👍

slfan1989:
> @slfan1989 Thank you, LGTM 👍

@fresh-borzoni Thanks for the review and the helpful comments! I really appreciate it.

PR #3777, opened 2026-07-26
[FIP-37] Add bitmap Phase 2 functions and register via FlussCatalog

pbanakar:
### Purpose

Linked issue: Part of #3289

This PR completes the Phase 2 implementation of FIP-37 by adding the three remaining RoaringBitmap functions: one aggregate function (rb_xor_agg) and two scalar functions (rb_xor, rb_andnot). All three are registered as built-in catalog functions in FlussCatalog, building on the infrastructure from PR #3319 and the Phase 1 functions from PRs #3398 and #3492.

After USE CATALOG fluss_catalog, users can call these functions directly in Flink SQL without any CREATE TEMPORARY FUNCTION statement.

### Brief change log

New files in fluss-flink/fluss-flink-common:

RbXorAggFunction.java — rb_xor_agg(BYTES) -> BYTES. Aggregates multiple serialized RoaringBitmaps via bitwise XOR across rows. Returns elements that appear in an odd number of input bitmaps — useful for change detection and symmetric difference analysis. Extends AbstractRbAggFunction.

RbXorFunction.java — rb_xor(BYTES, BYTES) -> BYTES. Returns the symmetric difference of two bitmaps: elements present in exactly one of the two inputs. Returns null if either argument is null.

RbAndNotFunction.java — rb_andnot(BYTES, BYTES) -> BYTES. Returns elements present in the left bitmap but not in the right bitmap. Useful for exclusion analysis such as "users who visited page A but not page B." Returns null if either argument is null.

Modified files:

FlinkCatalog.java — registers rb_xor_agg, rb_xor, and rb_andnot in the BUILTIN_BITMAP_FUNCTIONS map.

RbAggFunctionsTest.java — adds unit tests for RbXorAggFunction covering basic XOR correctness, null input handling, empty accumulator, merge, and retract throws.

RbScalarFunctionsTest.java — adds unit tests for RbXorFunction and RbAndNotFunction covering correctness, null semantics, disjoint sets, and edge cases.

FlinkCatalogTest.java — updates testViewsAndFunctions and testBitmapFunctionsRegistered to include the three new functions.

RbFunctionsCatalogITCase.java — adds end-to-end integration tests for all three functions through the catalog registration path using a real TableEnvironment.

### Tests

**Verified locally:**

./mvnw spotless:apply — 0 violations
./mvnw test -pl fluss-flink/fluss-flink-common -Dtest="RbAggFunctionsTest,RbScalarFunctionsTest,FlinkCatalogTest" — BUILD SUCCESS
./mvnw verify -pl fluss-flink/fluss-flink-common -Dit.test=RbFunctionsCatalogITCase -DfailIfNoTests=false — BUILD SUCCESS

Note: ./mvnw verify -Dit.test=RbFunctionsCatalogITCase is blocked locally by a pre-existing TieringSourceEnumeratorTest classloading failure unrelated to this PR (upstream dependency TieringTableValidator missing from local build). All unit tests pass. The ITCase will be verified by CI.

### API and Format

This change is purely additive and does not affect any storage formats or wire protocols. The functions operate on BYTES columns containing standard RoaringBitmap serialized data.

### Documentation

User-facing documentation for all Phase 1 and Phase 2 functions will be added in a follow-up PR to website/docs/table-design/merge-engines/aggregation.md.


pbanakar:
Hi @platinumhamburg  @polyzos could you please help retriger the CI and Review this

polyzos:
@Prajwal-banakar great work 👍 
One thing that needs to be addressed, though `rb_xor` and `rb_andnot` look correct and consistent with the Phase 1 functions. But `rb_xor_agg` has a correctness bug.

The problem is `merge()`. `RbXorAggFunction` inherits it from `AbstractRbAggFunction`, which ORs the partials together. That's fine for `rb_or_agg`, but for XOR the partial accumulators have to be combined with XOR, same reason `RbAndAggFunction` overrides `merge()` with AND. Right now under two-phase / local-global aggregation the merge unions the partials instead of computing the symmetric difference.

The fix should be small overriding `merge()` to XOR the partials:

    @Override
    public void merge(RoaringBitmap acc, Iterable<RoaringBitmap> it) {
        for (RoaringBitmap other : it) {
            if (other != null) {
                acc.xor(other);
            }
        }
    }


A couple of things on the tests while you're in there. `testXorAggMerge` currently asserts the OR result (`{1,2,3} OR {3,4,5}` → cardinality 5) — that's baking in the bug; it should expect the XOR `{1,2,4,5}` (cardinality 4). 
It's worth adding a case where two partials each hold `{1}`, and the merged result cancels to an empty result, since that's the exact scenario that was broken. All the XOR tests only use two bitmaps, so nothing actually exercises the "odd number of inputs" contract — an accumulate test over three-plus bitmaps would cover it. And to catch this class of bug going forward, add an ITCase that runs streaming with `mini-batch.size=1` + two-phase, so the merge path is real.

NIT: `rb_xor` has no empty-result test; the agg's Javadoc should mention it returns null when the result fully cancels, which differs from the scalar; and if there's no server-side XOR-agg counterpart, mirror the `merge-engine=aggregation` caveat from `RbAndAggFunction`.

pbanakar:
Hi @polyzos, addressed all the review comments:

- Overrode merge() in RbXorAggFunction to use xor() instead of inheriting or() from AbstractRbAggFunction
- Fixed testXorAggMerge to assert the correct XOR result {1,2,4,5} (cardinality 4) instead of the OR result
- Added testXorAggMergeCancelsToEmpty — two partials each holding {1} cancel to empty, getValue returns null
- Added testXorAggThreeInputs — exercises the "odd number of inputs" contract: {1,2} XOR {2,3} XOR {3,4} = {1,4}
- Added testXorIdenticalSetsProducesEmpty for scalar rb_xor — identical inputs produce empty bitmap (not null)

why the CI is not running from yesterday🤔All 87 unit tests pass locally. PTAL!

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/functions/bitmap/RbXorAggFunction.java
```diff
@@ -0,0 +1,73 @@
...
+ * <p>Aggregates multiple serialized {@link RoaringBitmap} values using bitwise XOR across rows.
+ * Returns elements that appear in an odd number of input bitmaps — useful for change detection and
+ * symmetric difference analysis.
+ *
+ * <p>Note: there is no server-side {@code FieldRoaringBitmapXorAgg} counterpart. This function
+ * executes entirely in Flink. Users should be aware that combining it with {@code
+ * table.merge-engine=aggregation} may produce unexpected results during server-side compaction.
+ *
+ * <p>Null and empty inputs are ignored. Returns {@code null} when all inputs are null or the result
+ * fully cancels to an empty bitmap.
```
wuchong: Returning `null` when XOR cancels to an empty bitmap conflates a valid aggregate result with the no-input/all-null case. For example, aggregating two identical non-null bitmaps makes `rb_cardinality(rb_xor_agg(...))` return SQL `NULL` instead of `0`, so downstream SQL cannot distinguish cancellation from an all-null group. Please track whether any non-null bitmap was accumulated (including through `merge`) and serialize an empty bitmap when input was seen.

The test can reproduce this problem.

```java
class RbFunctionsCatalogITCase {

    @Test
    void testRbXorAggIdenticalBitmapsReturnsEmptyBitmap() throws Exception {
        byte[] bitmap = BitmapUtils.toBytes(RoaringBitmap.bitmapOf(1, 2, 3));

        Table source =
                tEnv.fromValues(
                        DataTypes.ROW(
                                DataTypes.FIELD("k", DataTypes.INT()),
                                DataTypes.FIELD("bmap", DataTypes.BYTES())),
                        Row.of(1, bitmap),
                        Row.of(1, bitmap));
        tEnv.createTemporaryView("identical_bitmaps", source);

        TableResult result =
                tEnv.executeSql(
                        "SELECT rb_cardinality(rb_xor_agg(bmap)) "
                                + "FROM identical_bitmaps GROUP BY k");
        List<Row> rows = CollectionUtil.iteratorToList(result.collect());

        assertThat(rows).hasSize(1);
        assertThat(rows.get(0).getField(0)).isEqualTo(0L);
    }
}
```

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/functions/bitmap/RbXorAggFunction.java
```diff
@@ -0,0 +1,73 @@
...
+        acc.xor(BitmapUtils.fromBytes(bitmapBytes));
+    }
+
+    /**
+     * Retraction is not supported for bitmap XOR aggregation.
+     *
+     * @throws UnsupportedOperationException always
+     */
+    public void retract(RoaringBitmap acc, @Nullable byte[] bitmapBytes) {
+        throw new UnsupportedOperationException(
```
wuchong: Because this `retract` method is present, Flink can select it for updating inputs and call it for `UPDATE_BEFORE`/`DELETE` records; the unconditional exception then turns a normal changelog into a runtime job failure. XOR is self-inverse, so retraction can use the same XOR operation and null/empty handling as `accumulate` (or the function must be rejected before execution on non-append inputs). Flink also implemented this method: https://github.com/apache/flink/blob/3c8740f3f24aaa407112b5b576e5bfbf9cf2d24a/flink-table/flink-table-runtime/src/main/java/org/apache/flink/table/runtime/functions/aggregate/AbstractBitmapXorWithRetractAggFunction.java

wuchong (commented):
Thanks @pbanakar , I left some comments.

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/functions/bitmap/RbXorAggFunction.java
pbanakar: Updated RbXorAggFunction to distinguish no non-null input from a valid empty result after XOR cancellation, including merge handling. Added a test verifying identical bitmaps return cardinality 0 rather than NULL.

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/functions/bitmap/RbXorAggFunction.java
pbanakar: Implemented retract support for XOR using the same XOR operation, Thanks for the file link, it helped!

pbanakar:
Hi @wuchong @platinumhamburg adressed both comments, ready for another look! PTAL

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/functions/bitmap/RbXorAggFunction.java
```diff
@@ -0,0 +1,312 @@
...
+
+    // -------------------------------------------------------------------------
+    // Accumulator
+    // -------------------------------------------------------------------------
+
+    /** Mutable accumulator that tracks initialization state for XOR aggregation. */
+    public static final class Accumulator {
+
+        /** True after the first non-null input has been accumulated. */
+        public boolean initialized = false;
```
wuchong: Please align this accumulator with Flink's retractable bitmap-XOR implementation: [AbstractBitmapXorWithRetractAggFunction](https://github.com/apache/flink/blob/3c8740f3f24aaa407112b5b576e5bfbf9cf2d24a/flink-table/flink-table-runtime/src/main/java/org/apache/flink/table/runtime/functions/aggregate/AbstractBitmapXorWithRetractAggFunction.java).

A boolean `initialized` cannot represent the state after the last non-null input is retracted: after `accumulate(x)` followed by `retract(x)`, it remains `true` and `getValue()` returns an empty bitmap, while the correct result is `NULL`. Please replace it with a `long nonNullCount`, increment/decrement it in `accumulate`/`retract`, sum it in `merge`, reset it to zero, and return `NULL` only when the count is zero. This also preserves the distinct case where non-null inputs remain but XOR cancellation produces a valid empty bitmap.

Please also declare the accumulator itself as a `STRUCTURED` type through `getAccumulatorDataType()` instead of declaring it as `RAW` and implementing custom `TypeInformation`/`TypeSerializer`. A structured accumulator exposes the count and bitmap fields to Flink's state type system, should be more efficient than serializing the whole accumulator as an opaque RAW value, and removes the extra TypeInfo/serializer maintenance.

pbanakar: Hi @wuchong, re-examined the getAccumulatorDataType() approach — DataTypes.STRUCTURED in Flink 1.20.3 (the version Fluss currently targets) requires DataTypes.Field[] but DataTypes.FIELD() returns UnresolvedField, causing a compile error. getAccumulatorDataType() is not in AggregateFunction in Flink 1.20 either. I've kept the custom TypeInformation/TypeSerializer pattern already established by RbAndAggFunction in this codebase, i think this is the correct approach for Flink 1.20. Happy to migrate to STRUCTURED when Fluss upgrades Flink or  do you have any  suggestions LMK,

The core semantic fix is addressed: boolean initialized → long nonNullCount, with increment in accumulate, decrement in retract, sum in merge, and null returned only when nonNullCount <= 0. Added full serializer coverage tests including the zero-count path.

platinumhamburg (approved):
LGTM.

wuchong (approved):
LGTM

PR #3778, opened 2026-07-27
[server] Support independent local log TTL

### LINKED ISSUE #3748: [Feature] Support independent TTL configuration for local log segments
### Search before asking

### Motivation

Currently, `table.log.ttl` is used for both:

- Retaining log segments in remote storage.
- Retaining local log segments when tiered storage is enabled.

This couples the local and remote retention periods. In practice, users may want to retain data in remote storage for a long time while deleting already-uploaded local segments earlier to reduce local disk usage.

There is also an ambiguity around `remoteLogEndOffset = -1`. It may mean either:

1. No log segment has ever been copied to remote storage.
2. Log segments were copied previously, but all remote segments have since expired and been deleted.

Local segment cleanup must distinguish these cases because a local segment should only be deleted after it has been successfully copied to remote storage.

### Solution

Introduce a new table configuration:

```text
table.log.local-ttl
```

This configuration controls how long local log segments are retained after they have been copied to remote storage.

The expected behavior is:

- `table.log.ttl` continues to control remote log retention.
- `table.log.local-ttl` controls local log segment retention.
- If `table.log.local-ttl` is not configured, it defaults to `table.log.ttl` for backward compatibility.
- `table.log.local-ttl` must be less than or equal to `table.log.ttl`.
- Local segments that have not been copied to remote storage must not be deleted, even if their local TTL has expired.

#### Remote Offset Tracking

Track two different remote offsets:

- The current remote log end offset derived from the active remote log manifest.
- The highest offset that has ever been successfully copied to remote storage.

The highest copied offset should be persisted and remain monotonic even after all corresponding remote segments expire.

This allows local cleanup to determine whether a segment was previously uploaded without relying on the ambiguous `remoteLogEndOffset = -1` value.

After a TabletServer or replica restart, the highest copied offset should be restored from persistent metadata.

### Anything else?

_No response_

### Willingness to contribute

gyang94:
### Purpose

Closes #3748.

`table.log.ttl` currently controls both remote retention and local segment retention. This couples the two lifecycles and prevents copied local segments from being removed earlier to reduce local disk usage.

### Brief change log

- Add `table.log.local-ttl` with a default value of one day.
- Use the option for local segment expiration while `table.log.ttl` continues to control remote retention.
- Bound local cleanup by the existing `highestCopiedEndOffset`, retaining segments that have not been copied to remote storage.
- Validate positive local TTL values and require an explicitly configured local TTL to be no greater than a positive remote TTL.
- Support dynamically setting and resetting the option through `ALTER TABLE` and propagate changes to online replicas.
- Continue using `table.log.ttl` for local cleanup when remote log tiering is disabled.

### Tests

- `mvn -o test -pl fluss-server -am -Dtest=TieredLocalSegmentTTLTest -Dsurefire.failIfNoSpecifiedTests=false`
- `mvn -o test -pl fluss-client -am -Dtest=FlussAdminITCase#testAlterTableLocalLogTtl -Dsurefire.failIfNoSpecifiedTests=false`
- `TableConfigTest`, `LocalSegmentTTLTest`, and `FlinkConversionsTest`
- Spotless, Checkstyle, RAT, and `git diff --check`

### API and Format

This adds the `table.log.local-ttl` table configuration option. It does not change the remote manifest, RPC, or storage format.

### Documentation

The Flink options, tiered-storage, and TTL documentation are updated for the new option.

### Generative AI disclosure

- [x] Yes — Codex was used to assist with implementation, tests, and documentation; the changes were reviewed by the author.


website/docs/maintenance/tiered-storage/remote-storage.md
```diff
@@ -35,7 +35,10 @@ Below is the list for all configurations to control the log segments tiered beha
 
 When local log segments are copied to remote storage, the local log segments will be deleted to reduce local disk cost.
 But sometimes, we want to keep the several latest log segments retain in local, although they have been coped to remote storage for better read performance.
-You can control how many log segments to retain in local by setting the configuration `table.log.tiered.local-segments`(default is 2) per table.
+You can control local retention per table with `table.log.tiered.local-segments` (default is 2)
+and `table.log.local-ttl` (default is 1 day). A configured local TTL must be greater than zero
+and, when `table.log.ttl` is positive, less than or equal to `table.log.ttl`. A local segment is
+deleted only after it has been copied to remote storage.
```
wuchong, line 42: "A local segment" --> "An expired local segment"

fluss-client/src/test/java/org/apache/fluss/client/admin/FlussAdminITCase.java
```diff
@@ -928,6 +951,24 @@ void testCreateTableWithInvalidProperty() {
...
+    @Test
+    void testCreateTableWithLocalTtlAndInfiniteRemoteTtl() throws Exception {
+        TablePath tablePath =
+                TablePath.of(
+                        DEFAULT_TABLE_PATH.getDatabaseName(),
+                        "test_local_ttl_with_infinite_remote_ttl");
+        TableDescriptor tableDescriptor =
+                TableDescriptor.builder()
+                        .schema(DEFAULT_SCHEMA)
+                        .property(ConfigOptions.TABLE_LOG_TTL.key(), "0ms")
```
wuchong, line 963: If `0ms` is not allowed for `TABLE_LOG_LOCaAL_TTL`, why is it permitted for `TABLE_LOG_TTL`?  I think we should allow `0` to disable TTL, just like `table.log.ttl` and align with `table.log.ttl`.

fluss-server/src/test/java/org/apache/fluss/server/log/remote/TieredLocalSegmentTTLTest.java
```diff
@@ -94,34 +100,24 @@ void testInactiveTieredLocalSegmentRemovedAfterTtl(boolean partitionTable) throw
 
     @ParameterizedTest
     @ValueSource(booleans = {true, false})
-    void testExpiredActiveSegmentWaitsForHighWatermarkWithoutRemoteLogEndOffset(
```
wuchong, line 97: Why this test is removed? I think this is meaningful to test expire active segments.

fluss-common/src/main/java/org/apache/fluss/config/ConfigOptions.java
```diff
@@ -1835,6 +1835,16 @@ public class ConfigOptions {
                             "The number of log segments to retain in local for each table when log tiered storage is enabled. "
                                     + "It must be greater that 0. The default is 2.");
 
+    public static final ConfigOption<Duration> TABLE_LOG_LOCAL_TTL =
+            key("table.log.local-ttl")
+                    .durationType()
+                    .defaultValue(Duration.ofDays(1))
```
wuchong: Please leave `table.log.local-ttl` unset by default and fall back to `table.log.ttl`, as Kafka does with `local.retention.ms`. With a fixed one-day default, existing tables can silently change their local retention after an upgrade, causing abrupt changes in disk usage and read performance. A table configured only with `table.log.ttl=1h` also bypasses this invariant check because validation uses `getOptional(TABLE_LOG_LOCAL_TTL)`, while runtime cleanup still uses the one-day default, so its effective local TTL exceeds the remote TTL. This violates the `local TTL <= remote TTL` constraint and the backward-compatibility requirement in [Issue #3748](https://github.com/apache/fluss/issues/3748). Please keep the new option without a default and make `TableConfig#getLocalLogTTLMs()` use it only when explicitly configured; otherwise, return `table.log.ttl`. Resetting the option should restore the same fallback.

fluss-common/src/main/java/org/apache/fluss/config/ConfigOptions.java
```diff
@@ -1835,6 +1835,16 @@ public class ConfigOptions {
...
+                    .withDescription(
+                            "The time to retain local log segments after they have been copied to remote storage. "
```
wuchong: Improved a bit for the descriptions 

```
public static final ConfigOption<Duration> TABLE_LOG_TTL =
        key("table.log.ttl")
                .durationType()
                .defaultValue(Duration.ofDays(7))
                .withDescription(
                        "The time to live for log segments. The configuration controls the "
                                + "maximum time log segments are retained before they become "
                                + "eligible for deletion. When remote log tiering is enabled, "
                                + "this value controls the retention of remote log segments. "
                                + "Setting the value to -1 disables TTL-based deletion. "
                                + "The default value is 7 days.");

public static final ConfigOption<Duration> TABLE_LOG_LOCAL_TTL =
        key("table.log.local-ttl")
                .durationType()
                .noDefaultValue()
                .withDescription(
                        "The time to live for local log segments. The configuration controls the "
                                + "maximum time local log segments are retained before they become "
                                + "eligible for deletion. When remote log tiering is enabled, an "
                                + "expired local segment is deleted only after it has been copied "
                                + "to remote storage. Setting the value to -1 disables TTL-based "
                                + "deletion. If not configured, the value inherits `table.log.ttl`. "
                                + "When both values are positive, it must be less than or equal to "
                                + "`table.log.ttl`.");
```

fluss-server/src/test/java/org/apache/fluss/server/log/remote/TieredLocalSegmentTTLTest.java
```diff
@@ -94,34 +100,24 @@ void testInactiveTieredLocalSegmentRemovedAfterTtl(boolean partitionTable) throw
...
-        logManager.reconfigure(conf);
         makeLogTableAsLeader(tb, partitionTable);
         LogTablet logTablet = replicaManager.getReplicaOrException(tb).getLogTablet();
 
         addMultiSegmentsToLogTablet(logTablet, 5);
         assertThat(remoteLogManager.remoteLogTablet(tb).getRemoteLogEndOffset()).isEmpty();
 
-        manualClock.advanceTime(Duration.ofHours(2));
-        logTablet.updateHighWatermark(logTablet.localLogEndOffset() - 1L);
+        manualClock.advanceTime(Duration.ofMinutes(90));
```
wuchong: We should advance the time for > 2 hours, to validate the log can be expired by the high-watermark.

fluss-server/src/main/java/org/apache/fluss/server/log/LogManager.java
```diff
@@ -502,6 +507,12 @@ private LogTablet loadLog(
         return logTablet;
     }
 
+    private long effectiveLocalLogTtlMs(long logTtlMs, long localLogTtlMs) {
```
wuchong: Could we move the effective local-TTL resolution into `LogTablet` and make every construction and hot-update path pass both `logTtlMs` and `localLogTtlMs` explicitly? The background is that these two values have different configuration meanings: `table.log.ttl` remains authoritative for local cleanup when remote tiering is disabled, while `table.log.local-ttl` controls local cleanup when tiering is enabled and falls back to `table.log.ttl` when it is not configured. Therefore the runtime value used by local cleanup is always `remoteLogEnabled ? localLogTtlMs : logTtlMs`.

The current patch resolves that runtime value in `LogManager` for normal creation and disk recovery, but the same policy is duplicated in `ReplicaManager#updateReplicaTableConfig()` for hot updates, after which `Replica#updateLocalLogTtlMs()` and `LogTablet#updateLocalLogTtlMs()` receive only the already-resolved value. The testing `LogTablet#create()` overload also passes `getLocalLogTTLMs()` directly. This makes the effective-value policy distributed across several call sites: a new construction or update path can bypass it, and the creation and hot-update implementations can drift independently. It also leaves `LogTablet.localLogTtlMs` ambiguous because the field actually stores the runtime effective TTL rather than necessarily the configured local TTL.

I suggest using the following ownership model:

1. Keep `TABLE_LOG_LOCAL_TTL` without a fixed default and let `TableConfig#getLocalLogTTLMs()` implement only the configuration fallback to `table.log.ttl`.
2. Remove `LogManager#effectiveLocalLogTtlMs()`. Every `LogTablet` constructor/factory call, including the normal creation path, disk recovery path, and testing overload, should pass both `logTtlMs` and `localLogTtlMs` unchanged. `LogManager#getOrCreateLog()` can continue returning an existing entry directly; no additional reconciliation is needed there.
3. Let `LogTablet` keep `remoteLogEnabled` and a single `volatile long effectiveLocalLogTtlMs`. Its constructor should initialize that field through one private `effectiveLocalLogTtlMs(logTtlMs, localLogTtlMs)` helper, and all expiration checks should use this field.
4. Replace `updateLocalLogTtlMs(long)` with `updateLogTtls(long logTtlMs, long localLogTtlMs)`, and have it call the same helper before storing the new effective value. This guarantees that construction and hot update use exactly the same rule.
5. For hot updates, let `ReplicaManager#updateReplicaTableConfig()` store `TableConfig` directly in a `Map<Long, TableConfig>`, then use its existing single pass over online replicas to call `replica.updateLogTtls(tableConfig.getLogTTLMs(), tableConfig.getLocalLogTTLMs())`. `Replica` only needs to forward both values to `LogTablet` and optionally log the old/new effective value. There is no need to delegate updates to `LogManager` or scan `currentLogs`.

This keeps the update complexity at `O(number of tables + number of replicas)`, gives `LogTablet` sole ownership of the runtime policy it executes, and ensures that creation, recovery, SET, and RESET all derive `effectiveLocalLogTtlMs` through the same implementation.

fluss-common/src/main/java/org/apache/fluss/config/FlussConfigUtils.java
```diff
@@ -52,6 +52,7 @@ public class FlussConfigUtils {
                         ConfigOptions.TABLE_DATALAKE_FRESHNESS.key(),
                         ConfigOptions.TABLE_DATALAKE_AUTO_COMPACTION.key(),
                         ConfigOptions.TABLE_TIERED_LOG_LOCAL_SEGMENTS.key(),
+                        ConfigOptions.TABLE_LOG_LOCAL_TTL.key(),
                         ConfigOptions.TABLE_AUTO_PARTITION_ENABLED.key(),
```
wuchong, line 56: `table.log.ttl` is the more commonly used retention option and remains authoritative when remote tiering is disabled, as well as the fallback when `table.log.local-ttl` is unset. This PR only makes `table.log.local-ttl` hot-updatable, so let's keep the current scope and add hot-update support for setting and resetting `table.log.ttl`, including propagation to online replicas, in the follow-up tracked by [#3231](https://github.com/apache/fluss/issues/3231).

wuchong (approved):
LGTM.

PR #3781, opened 2026-07-27
[server] feat: committed isr boundary

### LINKED ISSUE #3780: [server] Prevent uncommitted ISR changes from being installed in CoordinatorContext
### Search before asking

### Motivation

When processing an `AdjustIsr` request, the Coordinator first attempts to persist all proposed leader and ISR changes through a batch ZooKeeper update. If the batch update fails, it retries each bucket individually.

Currently, a failed individual update is recorded as an error but then also receives an unconditional success result. In addition, all proposed changes are installed in `CoordinatorContext` and used for rebalance completion, regardless of whether their ZooKeeper writes succeeded.

This can cause:

- The same bucket to receive both error and success results.
- Failed ISR updates to be installed in Coordinator memory.
- `CoordinatorContext` to diverge from ZooKeeper.
- Rebalance completion to observe uncommitted ISR state.
- Future metadata propagation to publish non-authoritative ISR state.

### Solution

Explicitly separate proposed ISR changes from durably committed changes:

- Build and validate all proposed changes first.
- If the batch ZooKeeper update succeeds, treat all proposed changes as committed.
- If the batch update fails, retry each bucket individually and mark only successful writes as committed.
- Return exactly one success or error result for every requested bucket.
- Install only committed changes into `CoordinatorContext`.
- Run rebalance completion only for committed buckets.
- Keep validation failures out of ZooKeeper persistence attempts.

Add deterministic failure-injection tests covering batch success, fallback success, mixed individual outcomes, complete individual failure, and validation failures mixed with persistence outcomes.

### Anything else?

_No response_

### Willingness to contribute

gyang94:
### Purpose

Linked issue: close #3780

### Brief change log

separate proposed ISR changes from durably committed changes

### DIFF
--- a/fluss-server/src/main/java/org/apache/fluss/server/coordinator/CoordinatorEventProcessor.java
+++ b/fluss-server/src/main/java/org/apache/fluss/server/coordinator/CoordinatorEventProcessor.java
@@ -1925,7 +1925,8 @@ public class CoordinatorEventProcessor implements EventProcessor {
         // TODO verify leader epoch.
 
         List<AdjustIsrResultForBucket> result = new ArrayList<>();
-        Map<TableBucket, LeaderAndIsr> newLeaderAndIsrList = new HashMap<>();
+        Map<TableBucket, LeaderAndIsr> proposedChanges = new HashMap<>();
+        Map<TableBucket, LeaderAndIsr> committedChanges = new HashMap<>();
         for (Map.Entry<TableBucket, LeaderAndIsr> entry : leaderAndIsrList.entrySet()) {
             TableBucket tableBucket = entry.getKey();
             LeaderAndIsr tryAdjustLeaderAndIsr = entry.getValue();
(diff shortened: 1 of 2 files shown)


fluss-server/src/main/java/org/apache/fluss/server/coordinator/CoordinatorEventProcessor.java
```diff
@@ -1959,41 +1960,42 @@ private List<AdjustIsrResultForBucket> tryProcessAdjustIsr(
...
+            proposedChanges.put(tableBucket, newLeaderAndIsr);
         }
 
         try {
             zooKeeperClient.batchUpdateLeaderAndIsr(
-                    newLeaderAndIsrList, coordinatorContext.getCoordinatorZkVersion());
-            newLeaderAndIsrList.forEach(
+                    proposedChanges, coordinatorContext.getCoordinatorZkVersion());
+            committedChanges.putAll(proposedChanges);
+            committedChanges.forEach(
```
morazow, line 1970: nit: maybe we can unify these two lines? And move `committedChanges` initialisation closer, above the try.

morazow (approved):
Added minor nit 🤝

PR #3785, opened 2026-07-27
[server][rpc] Return ISR and bucket epoch in bucket metadata

### LINKED ISSUE #3782: [rpc] Add authoritative ISR to Fluss metadata and RPC
### Search before asking

### Motivation

`LeaderAndIsr` is the authoritative source of a bucket’s leader, leader epoch, ISR, and bucket epoch. However, the current `BucketMetadata` and `PbBucketMetadata` only carry:

- Bucket ID
- Leader ID
- Leader epoch
- Replica assignment

They do not carry the authoritative ISR or the generation of the complete leader/ISR state. The RPC schema also contains an existing `TODO` for adding ISR information.

Without these fields, metadata consumers cannot distinguish between assigned replicas and replicas that are currently in the ISR. They also cannot determine whether two metadata records represent different generations of the leader/ISR state.

The metadata contract should additionally distinguish between:

- An authoritative leader/ISR state.
- A known absence of leader/ISR state.
- Legacy metadata that does not carry authoritative leader/ISR information.

### Solution

Extend the Fluss metadata model and RPC contract to carry authoritative leader/ISR state:

- Add `isr` and nullable `bucketEpoch` fields to `BucketMetadata`.
- Use `bucketEpoch = -1` to represent a known absence of leader/ISR state.
- Use `bucketEpoch = null` to represent legacy metadata without authoritative leader/ISR information.
- Defensively copy and expose immutable replica and ISR lists.
- Include ISR and bucket epoch in `equals`, `hashCode`, and `toString`.
- Add the following backward-compatible fields to `PbBucketMetadata`:
  - `optional int32 bucket_epoch = 5`
  - `repeated int32 isr_id = 6 [packed = true]`
- Update RPC serialization and deserialization to preserve ISR, bucket epoch, and legacy field-presence semantics.
- Populate leader, leader epoch, ISR, and bucket epoch from the same `LeaderAndIsr` instance in both Coordinator and ZooKeeper metadata paths.
- Propagate the authoritative state through table, partition, and incremental metadata updates.
- Add coverage for model immutability, RPC round trips, legacy payloads, authoritative empty ISR, missing leader/ISR state, and Coordinator/ZooKeeper metadata construction.

### Anything else?

_No response_

### Willingness to contribute

gyang94:
### Purpose

Linked issue: close #3782

### DIFF
--- a/fluss-rpc/src/main/proto/FlussApi.proto
+++ b/fluss-rpc/src/main/proto/FlussApi.proto
@@ -869,8 +869,10 @@ message PbBucketMetadata {
   // optional as some time the leader may not elected yet
   optional int32 leader_id = 2;
   repeated int32 replica_id = 3 [packed = true];
-  // TODO: Add isr here.
   optional int32 leader_epoch = 4;
+  // Generation of the complete leader/ISR state. Absence indicates legacy metadata.
+  optional int32 bucket_epoch = 5;
+  repeated int32 isr_id = 6 [packed = true];
 }
 
 message PbProduceLogReqForBucket {
--- a/fluss-rpc/src/main/proto/FlussApi.proto
+++ b/fluss-rpc/src/main/proto/FlussApi.proto
@@ -882,7 +882,7 @@ message PbBucketMetadata {
   // Generation of the complete leader/ISR state.
   // Absence indicates legacy metadata; -1 indicates no leader/ISR state exists.
   optional int32 bucket_epoch = 5;
-  repeated int32 isr_id = 6 [packed = true];
+  repeated int32 isr = 6 [packed = true];
 }
 
 message PbProduceLogReqForBucket {
--- a/fluss-rpc/src/main/proto/FlussApi.proto
+++ b/fluss-rpc/src/main/proto/FlussApi.proto
@@ -879,7 +879,8 @@ message PbBucketMetadata {
   optional int32 leader_id = 2;
   repeated int32 replica_id = 3 [packed = true];
   optional int32 leader_epoch = 4;
-  // Generation of the complete leader/ISR state. Absence indicates legacy metadata.
+  // Generation of the complete leader/ISR state.
+  // Absence indicates legacy metadata; -1 indicates no leader/ISR state exists.
   optional int32 bucket_epoch = 5;
   repeated int32 isr_id = 6 [packed = true];
 }
--- a/fluss-rpc/src/main/proto/FlussApi.proto
+++ b/fluss-rpc/src/main/proto/FlussApi.proto
@@ -878,8 +878,10 @@ message PbBucketMetadata {
   // optional as some time the leader may not elected yet
   optional int32 leader_id = 2;
   repeated int32 replica_id = 3 [packed = true];
-  // TODO: Add isr here.
   optional int32 leader_epoch = 4;
+  // Generation of the complete leader/ISR state. Absence indicates legacy metadata.
+  optional int32 bucket_epoch = 5;
+  repeated int32 isr_id = 6 [packed = true];
 }
 
 message PbProduceLogReqForBucket {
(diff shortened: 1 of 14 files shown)


fluss-rpc/src/main/proto/FlussApi.proto
```diff
@@ -870,8 +869,10 @@ message PbBucketMetadata {
   // optional as some time the leader may not elected yet
   optional int32 leader_id = 2;
   repeated int32 replica_id = 3 [packed = true];
-  // TODO: Add isr here.
   optional int32 leader_epoch = 4;
+  // Generation of the complete leader/ISR state. Absence indicates legacy metadata.
```
morazow: Should we add comment for `-1` case? Since `-1` has different meaning.

Or remove comment all together.

morazow:
```suggestion
  // Generation of the complete leader/ISR state. Absence indicates legacy metadata, -1 means the sender has no leader/ISR state for this bucket yet.
```

fluss-server/src/main/java/org/apache/fluss/server/utils/ServerRpcMessageUtils.java
```diff
@@ -648,6 +648,15 @@ private static List<PbBucketMetadata> toPbBucketMetadata(
...
             }
 
+            Integer bucketEpoch = bucketMetadata.getBucketEpoch();
+            if (bucketEpoch != null) {
+                pbBucketMetadata.setBucketEpoch(bucketEpoch);
+            }
+
+            for (Integer isr : bucketMetadata.getIsr()) {
+                pbBucketMetadata.addIsrId(isr);
+            }
```
morazow:
```suggestion
            for (Integer isrId : bucketMetadata.getIsr()) {
                pbBucketMetadata.addIsrId(isrId);
            }
```

morazow (approved):
Hello @gyang94 , 

Thanks for the PR 🚀 Overall looks good, I left minor nits.

Please also don't forget to adjust the PR title:

```
[component] Brief description

e.g.

[server][rpc] Return ISR and bucket epoch in bucket metadata
```

affo:
Hello @gyang94! I performed a high-level comparison with this PR and #4209.

This PR and #4029 are complementary halves: this one builds the internal plumbing (authoritative ISR + `bucket_epoch` through the metadata path), #4029 builds the public admin door for the same bucket-level state. They overlap in `BucketMetadata.java`, `ZooKeeperClient.java`, and the proto — suggest this PR lands first, since it defines the authoritative shape (#4029's `BucketMetadata` change is a subset of this one), and #4029 rebases to reuse it.

One design question worth a sentence in the description: `PbBucketMetadata` is embedded in both `UpdateMetadataRequest` (coordinator → tablet servers, where ISR/epoch are correctness state — fencing stale updates) and the client-facing `MetadataResponse` (where clients only need the leader for routing). Is exposing ISR/epoch to every client on every metadata fetch intentional, or a consequence of the shared message shape? Kafka carries ISR in client metadata too, so it's defensible — but that's the accretion path that eventually pushed Kafka to move describe-grade state into a dedicated RPC (which is exactly what #4029 adds), so recording the intent either way would help future reviewers.

fluss-rpc/src/main/proto/FlussApi.proto
```diff
@@ -870,8 +869,11 @@ message PbBucketMetadata {
   // optional as some time the leader may not elected yet
   optional int32 leader_id = 2;
   repeated int32 replica_id = 3 [packed = true];
-  // TODO: Add isr here.
   optional int32 leader_epoch = 4;
+  // Generation of the complete leader/ISR state.
+  // Absence indicates legacy metadata; -1 indicates no leader/ISR state exists.
+  optional int32 bucket_epoch = 5;
+  repeated int32 isr_id = 6 [packed = true];
```
swuferhong: Nit: would `isr` be a more appropriate field name than `isr_id` here?

ISR already represents the collection of in-sync replicas, while `isr_id` may sound like the ID of a separate ISR entity. Using `isr` would also align better with the public Java API `BucketInfo#getIsr()`.

```protobuf
repeated int32 isr = 7 [packed = true];
```

swuferhong (commented):
LGTM +1, left a small comment:

swuferhong (approved):

wuchong (approved):

PR #3790, opened 2026-07-28
[server] fix KV snapshot restoring fallback

### LINKED ISSUE #3740: PK table bucket can't rebuild from the missing snapshot in the remote storage
### Search before asking

### Fluss version

0.9.0 (latest release)

### Please describe the bug 🐞

After a rolling update of the cluster, I found that one bucket of a PK table could not initialize RocksDB from the snapshot and WAL. The relevant failure logs are shown below:

```
Caused by: org.apache.hadoop.ipc.RemoteException: File does not exist: /rbf/data/fluss/kv/fluss_lake/opal_state_v1_fg_2456-149/100/snap-408/420d54b8-9af9-4047-85df-747876f4a750
        at org.apache.hadoop.hdfs.server.namenode.INodeFile.valueOf(INodeFile.java:86)
        at org.apache.hadoop.hdfs.server.namenode.INodeFile.valueOf(INodeFile.java:76)
        at org.apache.hadoop.hdfs.server.namenode.FSDirStatAndListingOp.getBlockLocations(FSDirStatAndListingOp.java:156)
        at org.apache.hadoop.hdfs.server.namenode.FSNamesystem.getBlockLocations(FSNamesystem.java:2063)
        at org.apache.hadoop.hdfs.server.namenode.NameNodeRpcServer.getBlockLocations(NameNodeRpcServer.java:796)
        at org.apache.hadoop.hdfs.protocolPB.ClientNamenodeProtocolServerSideTranslatorPB.getBlockLocations(ClientNamenodeProtocolServerSideTranslatorPB.java:458)
        at org.apache.hadoop.ipc.ProtobufRpcEngine$Server$ProtoBufRpcInvoker.call(ProtobufRpcEngine.java:539)
        at org.apache.hadoop.ipc.RPC$Server.call(RPC.java:1110)
        at org.apache.hadoop.ipc.Server$RpcCall.run(Server.java:1050)
        at org.apache.hadoop.ipc.Server$RpcCall.run(Server.java:973)
        at javax.security.auth.Subject.doAs(Subject.java:422)
        at org.apache.hadoop.security.UserGroupInformation.doAs(UserGroupInformation.java:1918)
        at org.apache.hadoop.ipc.Server$Handler.run(Server.java:2982)

        at org.apache.hadoop.ipc.Client.getRpcResponse(Client.java:1569) ~[hadoop-common-3.2.2.jar:?]
        at org.apache.hadoop.ipc.Client.call(Client.java:1515) ~[hadoop-common-3.2.2.jar:?]
        at org.apache.hadoop.ipc.Client.call(Client.java:1412) ~[hadoop-common-3.2.2.jar:?]
        at org.apache.hadoop.ipc.ProtobufRpcEngine$Invoker.invoke(ProtobufRpcEngine.java:244) ~[hadoop-common-3.2.2.jar:?]
        at org.apache.hadoop.ipc.ProtobufRpcEngine$Invoker.invoke(ProtobufRpcEngine.java:129) ~[hadoop-common-3.2.2.jar:?]
        at jdk.proxy2.$Proxy27.getBlockLocations(Unknown Source) ~[?:?]
        at org.apache.hadoop.hdfs.protocolPB.ClientNamenodeProtocolTranslatorPB.getBlockLocations(ClientNamenodeProtocolTranslatorPB.java:333) ~[hadoop-hdfs-client-3.2.2.jar:?]
        at jdk.internal.reflect.GeneratedMethodAccessor11.invoke(Unknown Source) ~[?:?]
        at jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43) ~[?:?]
        at java.lang.reflect.Method.invoke(Method.java:569) ~[?:?]
        at org.apache.hadoop.io.retry.RetryInvocationHandler.invokeMethod(RetryInvocationHandler.java:426) ~[hadoop-common-3.2.2.jar:?]
        at org.apache.hadoop.io.retry.RetryInvocationHandler$Call.invokeMethod(RetryInvocationHandler.java:169) ~[hadoop-common-3.2.2.jar:?]
```

But the _metadata file exists. the content as follows:

```
{
    "version": 1,
    "table_id": 149,
    "bucket_id": 100,
    "snapshot_id": 408,
    "snapshot_location": "rbf/data/fluss/kv/fluss_lake/opal_state_v1_fg_2456-149/100/snap-408",
    "kv_snapshot_handle": {
        "shared_file_handles": [
            {
                "kv_file_handle": {
                    "path": "rbf/data/fluss/kv/fluss_lake/opal_state_v1_fg_2456-149/100/shared/aaf668b2-847d-4ea4-b798-62c5c1bcfeab",
                    "size": 19301626
                },
                "local_path": "000755.sst"
            },
            {
                "kv_file_handle": {
                    "path": "bf/data/fluss/kv/fluss_lake/opal_state_v1_fg_2456-149/100/shared/11051cc5-5cb8-4962-a642-45cda6cafef6",
                    "size": 27621774
                },
                "local_path": "000733.sst"
            },
            {
                "kv_file_handle": {
                    "path": "rbf/data/fluss/kv/fluss_lake/opal_state_v1_fg_2456-149/100/shared/5bfa680a-652e-4621-9088-f22edc6b0cac",
                    "size": 67651196
                },
                "local_path": "000704.sst"
            },
            {
                "kv_file_handle": {
                    "path": "bf/data/fluss/kv/fluss_lake/opal_state_v1_fg_2456-149/100/shared/04871351-8c5b-415f-b84c-3a4ae89e60c6",
                    "size": 67647338
                },
                "local_path": "000705.sst"
            },
            {
                "kv_file_handle": {
                    "path": "rbf/data/fluss/kv/fluss_lake/opal_state_v1_fg_2456-149/100/shared/986074a4-4511-4994-97cf-265234902006",
                    "size": 59108533
                },
                "local_path": "000706.sst"
            },
            {
                "kv_file_handle": {
                    "path": "rbf/data/fluss/kv/fluss_lake/opal_state_v1_fg_2456-149/100/shared/fa74dc40-9010-4952-afc4-9f7332355906",
                    "size": 52334
                },
                "local_path": "000775.sst"
            },
            {
                "kv_file_handle": {
                    "path": "rbf/data/fluss/kv/fluss_lake/opal_state_v1_fg_2456-149/100/shared/b971948c-7878-4b38-bfe0-daf1d82c70fa",
                    "size": 10772241
                },
                "local_path": "000774.sst"
            }
        ],
        "private_file_handles": [
            {
                "kv_file_handle": {
                    "path": "rbf/data/fluss/kv/fluss_lake/opal_state_v1_fg_2456-149/100/snap-408/1929184d-aa34-46b0-b914-8f70080d4c87",
                    "size": 2377
                },
                "local_path": "MANIFEST-000766"
            },
            {
                "kv_file_handle": {
                    "path": "rbf/data/fluss/kv/fluss_lake/opal_state_v1_fg_2456-149/100/snap-408/dde56efa-6b62-43e4-8c85-294df4739868",
                    "size": 6411
                },
                "local_path": "OPTIONS-000769"
            },
            {
                "kv_file_handle": {
                    "path": "rbf/data/fluss/kv/fluss_lake/opal_state_v1_fg_2456-149/100/snap-408/420d54b8-9af9-4047-85df-747876f4a750",
                    "size": 16
                },
                "local_path": "CURRENT"
            }
        ],
        "snapshot_incremental_size": 10833379
    },
    "log_offset": 6782001
}
```

This bug looks similar to #1304. However, because the “file not found” error message is different, the KV initialization process cannot fall back to the previous snapshot, even though that snapshot appears to be valid.

### Solution

_No response_

zuston:
### Purpose

Fix KV tablet recovery when the latest remote snapshot is broken or missing. The recovery process now correctly detects missing snapshot files reported by different filesystem implementations, removes the broken snapshot metadata, and safely falls back to the previous available snapshot.

Close #3740

### Brief change log

- Detect missing snapshot data through the exception cause chain, including Hadoop's `File does not exist` error message.
- Distinguish newly created and restored snapshot handles so discarding a restored broken snapshot does not delete shared files referenced by older snapshots.
- Add regression coverage for broken snapshot cleanup and the `latest snapshot → previous snapshot` fallback flow.

### DIFF
--- a/fluss-server/src/main/java/org/apache/fluss/server/kv/snapshot/CompletedSnapshot.java
+++ b/fluss-server/src/main/java/org/apache/fluss/server/kv/snapshot/CompletedSnapshot.java
@@ -27,7 +27,9 @@ import org.apache.fluss.utils.concurrent.FutureUtils;
 import javax.annotation.Nullable;
 import javax.annotation.concurrent.NotThreadSafe;
 
+import java.io.FileNotFoundException;
 import java.io.IOException;
+import java.nio.file.NoSuchFileException;
 import java.util.Arrays;
 import java.util.List;
 import java.util.Objects;
@@ -204,6 +206,9 @@ public class CompletedSnapshot {
     public static boolean isSnapshotDataNotExists(Throwable throwable) {
         Throwable cause = throwable;
         while (cause != null) {
+            if (cause instanceof FileNotFoundException || cause instanceof NoSuchFileException) {
+                return true;
+            }
             String message = cause.getMessage();
             if (message != null
                     && (message.contains(SNAPSHOT_DATA_NOT_EXISTS_ERROR_MESSAGE)
@@ -213,6 +218,27 @@ public class CompletedSnapshot {
             cause = cause.getCause();
         }
         return false;
+    }
+
+    /**
+     * Returns whether the throwable indicates missing snapshot data and the given path is confirmed
+     * to be absent from remote storage.
+     *
+     * <p>If checking the path fails, this method conservatively returns {@code false} so callers do
+     * not perform destructive cleanup. The verification failure is attached to the original
+     * throwable as a suppressed exception for diagnostics.
+     */
+    public static boolean isSnapshotDataNotExists(Throwable throwable, FsPath dataPath) {
+        if (!isSnapshotDataNotExists(throwable)) {
+            return false;
+        }
+
+        try {
+            return !dataPath.getFileSystem().exists(dataPath);
+        } catch (Exception verificationException) {
+            throwable.addSuppressed(verificationException);
+            return false;
+        }
     }
 
     private void disposeMetadata() throws IOException {
--- a/fluss-server/src/main/java/org/apache/fluss/server/kv/snapshot/CompletedSnapshot.java
+++ b/fluss-server/src/main/java/org/apache/fluss/server/kv/snapshot/CompletedSnapshot.java
@@ -88,6 +88,9 @@ public class CompletedSnapshot {
 
     public static final String SNAPSHOT_DATA_NOT_EXISTS_ERROR_MESSAGE = "No such file or directory";
 
+    private static final String HADOOP_SNAPSHOT_DATA_NOT_EXISTS_ERROR_MESSAGE =
+            "File does not exist";
+
     public CompletedSnapshot(
             TableBucket tableBucket,
             long snapshotID,
@@ -163,9 +166,8 @@ public class CompletedSnapshot {
     }
 
     public CompletableFuture<Void> discardAsync(Executor ioExecutor) {
-        // it'll discard the snapshot files for kv, it'll always discard
-        // the private files; for shared files, only if they're not be registered in
-        // SharedKvRegistry, can the files be deleted.
+        // Always discard private files. Shared files are discarded directly only when the handle
+        // represents a newly created snapshot that has not been registered.
         CompletableFuture<Void> discardKvFuture =
                 FutureUtils.runAsync(kvSnapshotHandle::discard, ioExecutor);
 
@@ -193,6 +195,24 @@ public class CompletedSnapshot {
 
     public static FsPath getMetadataFilePath(FsPath snapshotLocation) {
         return new FsPath(snapshotLocation, SNAPSHOT_METADATA_FILE_NAME);
+    }
+
+    /**
+     * Returns whether the throwable or any of its causes indicates that snapshot data no longer
+     * exists in remote storage.
+     */
+    public static boolean isSnapshotDataNotExists(Throwable throwable) {
+        Throwable cause = throwable;
+        while (cause != null) {
+            String message = cause.getMessage();
+            if (message != null
+                    && (message.contains(SNAPSHOT_DATA_NOT_EXISTS_ERROR_MESSAGE)
+                            || message.contains(HADOOP_SNAPSHOT_DATA_NOT_EXISTS_ERROR_MESSAGE))) {
+                return true;
+            }
+            cause = cause.getCause();
+        }
+        return false;
     }
 
     private void disposeMetadata() throws IOException {
--- a/fluss-server/src/main/java/org/apache/fluss/server/kv/snapshot/KvSnapshotHandle.java
+++ b/fluss-server/src/main/java/org/apache/fluss/server/kv/snapshot/KvSnapshotHandle.java
@@ -27,6 +27,7 @@ import java.util.Objects;
 import java.util.stream.Collectors;
 
 import static org.apache.fluss.utils.Preconditions.checkNotNull;
+import static org.apache.fluss.utils.Preconditions.checkState;
 
 /**
  * A handle to the snapshot of a kv tablet. It contains the share file handles and the private file
@@ -50,6 +51,11 @@ public class KvSnapshotHandle {
      */
     private boolean ownsSharedFiles;
 
+    /**
+     * The registry where shared files were registered, or null if registration has not happened.
+     */
+    private transient SharedKvFileRegistry sharedKvFileRegistry;
+
     KvSnapshotHandle(
             List<KvFileHandleAndLocalPath> sharedFileHandles,
             List<KvFileHandleAndLocalPath> privateFileHandles,
@@ -69,7 +75,13 @@ public class KvSnapshotHandle {
         return new KvSnapshotHandle(sharedFileHandles, privateFileHandles, incrementalSize, true);
     }
 
-    /** Restores a handle that only references already-persisted shared files. */
+    /**
+     * Restores a handle that only references already-persisted shared files.
+     *
+     * <p>A restored handle never directly deletes shared files because it cannot prove that no
+     * other snapshot references them. Shared files are reclaimed only through registry-based
+     * cleanup; retaining an untracked file is preferred over deleting a file that is still in use.
+     */
     static KvSnapshotHandle restore(
             List<KvFileHandleAndLocalPath> sharedFileHandles,
             List<KvFileHandleAndLocalPath> privateFileHandles,
@@ -133,7 +145,11 @@ public class KvSnapshotHandle {
     }
 
     public void registerKvFileHandles(SharedKvFileRegistry registry, long snapshotID) {
-        checkNotNull(registry);
+        checkState(
+                sharedKvFileRegistry != registry,
+                "The kv file handle has already registered its shared kv files to the given registry.");
+
+        sharedKvFileRegistry = checkNotNull(registry);
         ownsSharedFiles = false;
 
         for (KvFileHandleAndLocalPath handleAndLocalPath : sharedFileHandles) {
--- a/fluss-server/src/main/java/org/apache/fluss/server/kv/snapshot/KvSnapshotHandle.java
+++ b/fluss-server/src/main/java/org/apache/fluss/server/kv/snapshot/KvSnapshotHandle.java
@@ -50,7 +50,7 @@ public class KvSnapshotHandle {
      */
     private boolean ownsSharedFiles;
 
-    private KvSnapshotHandle(
+    KvSnapshotHandle(
             List<KvFileHandleAndLocalPath> sharedFileHandles,
             List<KvFileHandleAndLocalPath> privateFileHandles,
             long incrementalSize,
--- a/fluss-server/src/main/java/org/apache/fluss/server/kv/snapshot/KvSnapshotHandle.java
+++ b/fluss-server/src/main/java/org/apache/fluss/server/kv/snapshot/KvSnapshotHandle.java
@@ -27,7 +27,6 @@ import java.util.Objects;
 import java.util.stream.Collectors;
 
 import static org.apache.fluss.utils.Preconditions.checkNotNull;
-import static org.apache.fluss.utils.Preconditions.checkState;
 
 /**
  * A handle to the snapshot of a kv tablet. It contains the share file handles and the private file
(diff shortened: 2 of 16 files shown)


zuston:
PTAL @fresh-borzoni @wuchong

fluss-server/src/main/java/org/apache/fluss/server/kv/snapshot/CompletedSnapshot.java
```diff
@@ -195,6 +197,24 @@ public static FsPath getMetadataFilePath(FsPath snapshotLocation) {
         return new FsPath(snapshotLocation, SNAPSHOT_METADATA_FILE_NAME);
     }
 
+    /**
+     * Returns whether the throwable or any of its causes indicates that snapshot data no longer
+     * exists in remote storage.
+     */
+    public static boolean isSnapshotDataNotExists(Throwable throwable) {
```
fresh-borzoni, line 206: This deletes the znode and the files, so a wrong true here is unrecoverable. A standby during failover also says File does not exist while the file is fine, should we confirm with fileSystem.exists() before treating the snapshot as broken?

And should we match on FileNotFoundException / NoSuchFileException too? Otherwise S3, OSS and GCS each need another string added here later.

fluss-server/src/main/java/org/apache/fluss/server/kv/snapshot/KvSnapshotHandle.java
```diff
@@ -110,7 +120,7 @@ public void discard() {
             LOG.warn("Could not properly discard misc file states.", e);
         }
 
-        if (!isRegistered) {
+        if (ownsSharedFiles) {
```
fresh-borzoni, line 136: discard() no longer touches the shared files, which is right, but it's not obvious from here who does. 
What happens to the ones this snapshot alone referenced once its znode is gone?

fluss-server/src/main/java/org/apache/fluss/server/kv/snapshot/KvSnapshotHandle.java
```diff
@@ -123,11 +133,8 @@ public void discard() {
     }
 
     public void registerKvFileHandles(SharedKvFileRegistry registry, long snapshotID) {
-        checkState(
```
fresh-borzoni, line 126: Any reason to drop the checkState completely? A double registration goes silent now.

fluss-server/src/test/java/org/apache/fluss/server/replica/ReplicaTest.java
```diff
@@ -618,13 +624,24 @@ void testBrokenSnapshotRecovery(@TempDir File snapshotKvTabletDir) throws Except
...
+            FsPath path = new FsPath(sharedFilePath);
+            assertThat(path.getFileSystem().exists(path)).isTrue();
+        }
 
-        // ZK metadata should still show snapshot2 as latest (file corruption hasn't been detected
-        // yet)
+        // Simulate snapshot corruption by deleting only one private file while its metadata and
+        // shared files remain intact.
+        assertThat(snapshot2.getKvSnapshotHandle().getPrivateFileHandles()).isNotEmpty();
+        snapshot2.getKvSnapshotHandle().getPrivateFileHandles().get(0).getKvFileHandle().discard();
```
fresh-borzoni, line 745: Deleting a local file gives No such file or directory, so the Hadoop message from the issue isn't exercised on the replica path at all. 
Could we add a small unit test on isSnapshotDataNotExists with a nested File does not exist chain?

fluss-server/src/test/java/org/apache/fluss/server/replica/ReplicaTest.java
```diff
@@ -634,8 +651,31 @@ void testBrokenSnapshotRecovery(@TempDir File snapshotKvTabletDir) throws Except
         // create a new replica with the same snapshot context
         // During initialization, it will try to use snapshot2 but find it broken,
         // then handle the broken snapshot and fall back to snapshot1
+        List<Long> attemptedSnapshotIds = new ArrayList<>();
         testKvSnapshotContext =
-                new TestSnapshotContext(snapshotKvTabletDir.getPath(), kvSnapshotStore);
+                new TestSnapshotContext(snapshotKvTabletDir.getPath(), kvSnapshotStore) {
```
fresh-borzoni, line 757: The round-trip is the important bit as production always hands out a snapshot deserialized from ZK, and the base context handing out the live object is what hid this. 
Shall we move it into TestSnapshotContext instead of an anonymous subclass?

fresh-borzoni (commented):
@zuston Thank you, a good catch 👍  LGTM overall, couple of comments

fluss-server/src/main/java/org/apache/fluss/server/kv/snapshot/CompletedSnapshot.java
zuston: > This deletes the znode and the files, so a wrong true here is unrecoverable. A standby during failover also says File does not exist while the file is fine, should we confirm with fileSystem.exists() before treating the snapshot as broken?

good suggestion. 

> And should we match on FileNotFoundException / NoSuchFileException too? Otherwise S3, OSS and GCS each need another string added here later.

The type checks cover structured exceptions, while Hadoop may wrap the remote failure as a generic IOException and preserve only the “File does not exist” message. Therefore, the message matching is still required as a compatibility fallback.

fluss-server/src/main/java/org/apache/fluss/server/kv/snapshot/KvSnapshotHandle.java
zuston: this part is also missing in the current codebase. maybe we need introduce another GC mechanism to remove these reference=0 files.

fluss-server/src/main/java/org/apache/fluss/server/kv/snapshot/KvSnapshotHandle.java
zuston: undo this part

fluss-server/src/test/java/org/apache/fluss/server/replica/ReplicaTest.java
zuston: added in the `CompletedSnapshotTest`

fluss-server/src/test/java/org/apache/fluss/server/replica/ReplicaTest.java
zuston: Moved the JSON round-trip into the base `TestSnapshotContext` . The anonymous provider now delegates to the base implementation and only records the attempted snapshot IDs for verifying the fallback order.

fluss-server/src/main/java/org/apache/fluss/server/replica/Replica.java
```diff
@@ -854,7 +854,7 @@ private void downloadKvSnapshots(CompletedSnapshot completedSnapshot, Path kvTab
         try {
             kvSnapshotDataDownloader.transferAllDataToDirectory(downloadSpec, closeableRegistry);
         } catch (Exception e) {
-            if (e.getMessage().contains(CompletedSnapshot.SNAPSHOT_DATA_NOT_EXISTS_ERROR_MESSAGE)) {
+            if (CompletedSnapshot.isSnapshotDataNotExists(e)) {
```
fresh-borzoni: This still takes the one-arg overload, and it's the branch that removes the znode and discards the files.

I tried it with snapshot 0 fully intact on disk and the downloader failing once with `IOException("Fail to download kv snapshot.", new FileNotFoundException("File does not exist: <sst>"))`: `handleSnapshotBroken(0)` fires and the store goes empty. Checking the download spec's handles before that call keeps the snapshot, and `testBrokenSnapshotRecovery` stays green because the private file it deletes really is gone. Shall we?

fluss-server/src/main/java/org/apache/fluss/server/kv/snapshot/KvSnapshotHandle.java
```diff
@@ -46,22 +46,47 @@ public class KvSnapshotHandle {
...
+            List<KvFileHandleAndLocalPath> privateFileHandles,
+            long incrementalSize) {
+        return new KvSnapshotHandle(sharedFileHandles, privateFileHandles, incrementalSize, true);
+    }
+
+    /**
+     * Restores a handle that only references already-persisted shared files.
+     *
+     * <p>A restored handle never directly deletes shared files because it cannot prove that no
+     * other snapshot references them. Shared files are reclaimed only through registry-based
```
fresh-borzoni: Nothing registers a discarded snapshot, so the shared files only it referenced are never reclaimed by the registry either, they stay until the table dir is dropped. Leaking beats deleting a file snapshot1 still needs, so the trade is right, I'd just rather the comment said that instead. Plus a TODO and an issue for the reference=0 GC you mentioned?

fluss-server/src/test/java/org/apache/fluss/server/replica/ReplicaTest.java
```diff
@@ -647,19 +685,23 @@ void testBrokenSnapshotRecovery(@TempDir File snapshotKvTabletDir) throws Except
...
+                                Tuple2.of("k2", new Object[] {2, "b"}),
                                 Tuple2.of("k3", new Object[] {4, "d"})));
         verifyGetKeyValues(kvTablet, expectedKeyValues);
 
-        // Verify the core functionality: KvTablet successfully initialized despite broken snapshot
-        // The key test is that the system can handle broken snapshots and recover correctly
-
-        // Verify that we successfully simulated the broken snapshot condition
         File metadataFile = new File(snapshot2.getMetadataFilePath().getPath());
         assertThat(metadataFile.exists()).isFalse();
```
fresh-borzoni: This one can't fail. `_METADATA` is written by `CompletedSnapshotStore` on the coordinator and `TestingCompletedKvSnapshotCommitter` doesn't, so the file never exists here, even for a healthy snapshot. The store assertion above already covers the cleanup, drop this one?

fresh-borzoni (commented):
@zuston Thanks, looks better, a couple of things left, PTAL

fluss-server/src/main/java/org/apache/fluss/server/kv/snapshot/KvSnapshotHandle.java
zuston: this is tracked by https://github.com/apache/fluss/issues/3489

zuston: I have updated the comment

fluss-server/src/test/java/org/apache/fluss/server/replica/ReplicaTest.java
zuston: nice catch. removed

fresh-borzoni (approved):
@zuston Thank you, LGTM 👍

PR #3807, opened 2026-07-29
[kv] Avoid unnecessary RocksDB flushes during Tablet Server shutdown

### LINKED ISSUE #3806: [kv] Avoid unnecessary RocksDB flushes during Tablet Server shutdown
### Search before asking

### Motivation

Closing a RocksDB-backed KV tablet currently uses RocksDB's default close behavior, which flushes dirty memtables before closing the database.

During Tablet Server shutdown, this happens for every local KV replica. When a Tablet Server hosts many replicas, these flushes can produce substantial disk I/O and make shutdown take a long time.

The flushed SST files do not establish a committed Fluss recovery point:

- If a completed snapshot exists, replica initialization recreates the local KV directory, downloads the snapshot, and replays the changelog after the snapshot offset.
- If no completed snapshot exists, recovery rebuilds the KV state by replaying the changelog from offset 0.
- A close-time RocksDB flush does not commit the corresponding Fluss recovery metadata, such as the snapshot offset and other tablet state.

Therefore, flushing dirty memtables during Tablet Server shutdown provides little recovery benefit while adding shutdown latency proportional to the number of KV replicas and their dirty memtable volume.

However, flush-on-close cannot be disabled globally. Existing direct `KvManager.shutdown()` and local-reopen behavior relies on RocksDB preserving the local state.

### Solution

Introduce an explicit KV close mode to distinguish the two lifecycle requirements:

- `PRESERVE_LOCAL_STATE`: retain the existing RocksDB flush-on-close behavior.
- `DISCARD_UNPERSISTED_STATE`: allow unpersisted memtable state to be discarded because recovery will rebuild it from a completed snapshot and the changelog.

Use the modes as follows:

- Keep the no-argument `RocksDBKv.close()`, `KvTablet.close()`, and `KvManager.shutdown()` methods in `PRESERVE_LOCAL_STATE` mode for compatibility.
- Use `DISCARD_UNPERSISTED_STATE` when shutting down the KV manager as part of Tablet Server shutdown.
- Use `DISCARD_UNPERSISTED_STATE` when dropping a tablet because its local directory is deleted afterward.
- Immediately before closing RocksDB in discard mode, dynamically set `avoid_flush_during_shutdown=true`.
- Keep the global RocksDB DB options unchanged.
- Ensure all RocksDB native resources are still released if setting the dynamic option fails.

This keeps existing local-reopen semantics while avoiding shutdown-time flushes only where the local unpersisted state is not a recovery source.

### Anything else?

_No response_

### Willingness to contribute

platinumhamburg:
### Purpose

Linked issue: close #3806 3806

### DIFF
--- a/fluss-server/src/main/java/org/apache/fluss/server/kv/KvCloseMode.java
+++ b/fluss-server/src/main/java/org/apache/fluss/server/kv/KvCloseMode.java
@@ -1,0 +1,27 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.fluss.server.kv;
+
+/** Controls whether closing a KV tablet must persist its current local RocksDB state. */
+public enum KvCloseMode {
+    /** Preserve the existing local-reopen behavior by allowing RocksDB to flush on close. */
+    PRESERVE_LOCAL_STATE,
+
+    /** Skip flushing data that recovery will rebuild from a snapshot and changelog. */
+    DISCARD_UNPERSISTED_STATE
+}
--- a/fluss-server/src/main/java/org/apache/fluss/server/kv/KvTablet.java
+++ b/fluss-server/src/main/java/org/apache/fluss/server/kv/KvTablet.java
@@ -1288,7 +1288,15 @@ public final class KvTablet {
     }
 
     public void close() throws Exception {
-        LOG.info("close kv tablet {} for table {}.", tableBucket, physicalPath);
+        close(KvCloseMode.PRESERVE_LOCAL_STATE);
+    }
+
+    public void close(KvCloseMode closeMode) throws Exception {
+        LOG.info(
+                "Close kv tablet {} for table {} with mode {}.",
+                tableBucket,
+                physicalPath,
+                closeMode);
         boolean shouldClose =
                 inWriteLock(
                         kvLock,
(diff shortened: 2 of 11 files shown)


fluss-server/src/main/java/org/apache/fluss/server/kv/KvTablet.java
```diff
@@ -1318,7 +1326,7 @@ public void drop() throws Exception {
                 kvLock,
                 () -> {
                     // first close the kv.
-                    close();
+                    close(KvCloseMode.DISCARD_UNPERSISTED_STATE);
```
luoyuxia, line 1329: Could we still delete the KV directory when setting `avoid_flush_during_shutdown` fails?

`RocksDBKv.close(DISCARD_UNPERSISTED_STATE)` closes the DB, column-family handles, and options in its `finally`, but then rethrows the captured `RocksDBException`. That exception exits this lambda before `FileUtils.deleteDirectory(...)` runs. `KvManager.dropKv()` has already removed the tablet from `currentKvs`, so a retry through the manager cannot clean up the untracked directory.

Please either treat failure to apply this shutdown optimization as a warning once the fallback close has completed, or ensure the directory is deleted after confirming the native DB is closed while preserving the original exception. A regression test covering `KvTablet.drop()` with an injected option-setting failure would protect this path.

fluss-server/src/main/java/org/apache/fluss/server/kv/KvCloseMode.java
```diff
@@ -0,0 +1,27 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.fluss.server.kv;
+
+/** Defines whether closing a KV tablet must persist its current local RocksDB state. */
+public enum KvCloseMode {
```
luoyuxia, line 21: Do we really need two close modes here?

From the current production call sites, both TabletServer shutdown and tablet drop can discard unflushed memtable state: recovery rebuilds the KV from snapshots/changelog, or the local directory is deleted. `PRESERVE_LOCAL_STATE` seems to exist only to retain the behavior of directly reopening the same local RocksDB directory after `KvManager.shutdown()`.

Is that direct local-reopen behavior still a contract we need? If not, could we simplify this change by always setting `avoid_flush_during_shutdown=true` before closing RocksDB and remove `KvCloseMode` and its propagation? Since RocksDB WAL is disabled, this would intentionally make close terminal: callers must recover through Fluss instead of relying on unflushed local RocksDB state.

luoyuxia (commented):
@platinumhamburg Thanks for the comment. PTAL

fluss-server/src/main/java/org/apache/fluss/server/kv/KvCloseMode.java
platinumhamburg: Yes, we do need the two close modes here. In the future, we plan to support dynamically altering options for primary key tables, which requires KvTablet to support dynamic reload — and in that mode, we expect to retain the flush capability.

platinumhamburg:
@luoyuxia Thank you for your comments; I have addressed them. Please take a look when you have time.

luoyuxia (approved):
+1

PR #3810, opened 2026-07-30
[tiering] fail tiering job when readers exceed max duration

zuston:
### Purpose

When a table reaches its max tiering duration (datalake.freshness), the tiering enumerator already force-completes in-flight work by notifying source readers via TieringReachMaxDurationEvent. In some cases, however, readers remain stuck and never acknowledge completion (FinishedTieringEvent), leaving the tiering job hanging indefinitely.

This change adds a fast-fail safeguard in TieringSourceEnumerator: after max duration is reached, if the table is still active after another maxTieringDurationMs window, the coordinator fails the Flink tiering job with a FlinkRuntimeException. This allows Flink to restart/recover the stuck source readers instead of leaving the job running forever.


zuston:
cc @luoyuxia

zuston:
gently ping @luoyuxia

PR #3811, opened 2026-07-30
[server] batch fetch latest bucket snapshots to reduce latency

zuston:
### Purpose

For primary-key tables, tiering split generation fetches the latest KV snapshot for every bucket in each active partition. The previous implementation performed two sequential ZooKeeper reads per bucket: listing all snapshot IDs and then reading the latest snapshot data.
With 13 partitions and 800 buckets, this resulted in up to 20,800 sequential ZooKeeper reads, causing split generation to take around 450–490 seconds. This delay also extended the actual duration of each tiering round.

### Brief change log

Change snapshot loading to two pipelined phases:
1. Fetch snapshot children for all buckets concurrently.
2. Determine the latest snapshot ID for each bucket in memory.
3. Fetch the corresponding snapshot data concurrently.

The implementation reuses the existing background read utilities and respects the configured maximum inflight requests. Missing or concurrently deleted snapshot nodes are handled gracefully, while the existing result semantics and upper-layer interfaces remain unchanged.


zuston:
@luoyuxia @loserwang1024 could you help review this?

luoyuxia (approved):

Issue #3797, opened 2026-07-28
[TLS] Wire TLS encryption into the client Netty pipeline


affo:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/fluss/issues) and found nothing similar.

### Description

Part of #3786 ([FIP-29: (m)TLS Support](https://cwiki.apache.org/confluence/spaces/FLUSS/pages/406620533/FIP-29+m+TLS+Support)).

**Wave 2** — depends on: SSL config & context foundations (#3796), server pipeline integration (#3792).

- `NettyClient` — builds a client `SslContext` once when `client.security.ssl.enabled=true`.
- `ServerConnection.establishConnection` — adds `SslHandler` at the pipeline head, configures hostname verification (`client.security.ssl.endpoint.identification.algorithm`, default `https`) using the target host/port.
- `TlsDataPathITCase` — end-to-end client↔server data-path test (append/scan) over TLS, verifying the zero-copy fallback path from #3792.

**Depends on:** SSL config & context foundations (#3796), server pipeline integration (#3792).

_Estimated diff size: ~330 lines._

### Willingness to contribute

- [x] I'm willing to submit a PR!

Issue #3792, opened 2026-07-28
[TLS] Wire TLS encryption into the server Netty pipeline


affo:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/fluss/issues) and found nothing similar.

### Description

Part of #3786 ([FIP-29: (m)TLS Support](https://cwiki.apache.org/confluence/spaces/FLUSS/pages/406620533/FIP-29+m+TLS+Support)).

**Wave 1** — depends on: SSL config & context foundations (#3796).

Wires `SslHandler` into the server-side pipeline, enabling real encrypted connections for the first time — encryption only, no client-certificate requirement yet (`ClientAuth.NONE` on every TLS listener).

- `ServerChannelInitializer` — `pipeline.addFirst("ssl", ...)` for listeners in `security.ssl.enabled.listeners`, before the frame decoder.
- `NettyServer` / `FlussProtocolPlugin` — thread `Configuration` + listener name into `createChannelHandler`; minimal fail-fast validation (a TLS-enabled listener requires a keystore).
- `TlsFileRegionHandler` — copies file-backed log-fetch records into JVM memory before encryption, since `sendfile` and `SslHandler` are mutually exclusive.
- `TlsClusterITCase` — end-to-end cluster test with an encrypted listener.

**Depends on:** SSL config & context foundations (#3796).
**Depended on by:** client pipeline integration (#3797), internal listener migration (#3798), certificate hot-reload (#3799), mTLS authentication (#3793), Flink connector test (#3794), Spark connector test (#3800).

_Estimated diff size: ~415 lines._

_(Note: this replaces this ticket's original broader "server + client" scope, now split further into dedicated tickets.)_

### Willingness to contribute

- [x] I'm willing to submit a PR!

PR #3813, opened 2026-07-30
[FIP-29] Add SSL config parsing and SslContext/SslHandler factory

### LINKED ISSUE #3796: [TLS] Add SSL config parsing and SslContext/SslHandler factory
### Search before asking

### Description

Part of #3786 ([FIP-29: (m)TLS Support](https://cwiki.apache.org/confluence/spaces/FLUSS/pages/406620533/FIP-29+m+TLS+Support)).

**Wave 0** — no dependencies, can start immediately.

Add the foundational, currently-unused building blocks that later tickets wire into the Netty pipeline — no runtime behavior change on their own.

- `SslConfig` — parses keystore/truststore paths, passwords, type (JKS/PKCS12), enabled protocols, cipher suites from `Configuration`.
- `SslContextFactory` — builds a Netty `SslContext`/`SslHandler` from an `SslConfig`, for both server and client sides.
- New config options: `security.ssl.enabled.listeners`, `security.ssl.enabled.protocols`, `security.ssl.cipher.suites`, `security.ssl.keystore.*`, `security.ssl.key.password`, `security.ssl.truststore.*`, and the `client.security.ssl.*` mirror.
- Unit tests covering keystore/truststore loading, protocol/cipher restriction, and hostname-verification setup.

**Depends on:** none.
**Depended on by:** server pipeline integration (#3792), client pipeline integration (#3797), certificate hot-reload (#3799).

_Estimated diff size: ~925 lines._

### Willingness to contribute

MicheleGuerriero:
### Purpose

Linked issue: close #3796

First building block of FIP-29 (mTLS support for Fluss RPC). This adds
the configuration surface and a factory for building Netty SSL
contexts/handlers, but does not wire them into any pipeline yet — no
runtime behavior changes as a result of this PR.

### Brief change log

- `ConfigOptions`: add server-side `security.ssl.*` options (enabled
  listeners, protocols, cipher suites, keystore/truststore paths and
  passwords, reload interval) and client-side `client.security.ssl.*`
  options (mirrors the server options, plus endpoint identification
  algorithm for hostname verification).
- `SslConfig`: new immutable holder that parses and validates the
  above options from a `Configuration`, with `fromServerConfig`/
  `fromClientConfig` factory methods. Fails fast with a clear message
  if a server enables TLS without configuring a keystore.
- `SslContextFactory`: new factory building Netty `SslContext`/
  `SslHandler` instances from an `SslConfig`, using the JDK SSL
  provider. Supports per-listener client-certificate requirement
  (`setNeedClientAuth`, for mTLS listeners) on the server side, and
  SNI + hostname verification on the client side.

Later tickets (#3792 server pipeline, #3797 client pipeline) will wire
`SslContextFactory` into the actual Netty channel pipelines.

### Tests

Added `SslContextFactoryTest`, backed by `TestSslUtils` (generates a
self-signed certificate and on-the-fly JKS keystore/truststore files,
no committed key material):

- server/client `SslContext` creation
- enabled-protocol and cipher-suite filtering
- key-password fallback to keystore password
- client endpoint-identification-algorithm toggling (hostname
  verification on/off)
- server `needClientAuth` toggling for mTLS
- an end-to-end embedded-channel TLS handshake between a server and
  client `SslHandler`, asserting the negotiated session uses a real
  TLS protocol/cipher (not plaintext)
- fail-fast validation when no keystore is configured

Verified locally on `fluss-common` and `fluss-rpc`: `mvn test` (all
green, `SslContextFactoryTest` 9/9), `spotless:check`,
`checkstyle:check`, Apache RAT license check, and a JDK 8 build
(`-Pjava8`) confirming Java 8 source compatibility.

### API and Format

Adds new `ConfigOption`s (`security.ssl.*`, `client.security.ssl.*`)
and two new `@Internal` classes in `fluss-rpc`
(`org.apache.fluss.rpc.netty.ssl`). No changes to existing public
APIs, RPC wire format, or storage format. The new classes are
currently unused by any production code path.

### Documentation

No user-facing behavior yet, so no documentation changes in this PR.
Configuration docs for `security.ssl.*` / `client.security.ssl.*`
will be added once these options take effect (server/client pipeline
wiring in the following tickets).

- [x] Yes (Claude Code, Claude Opus 5)

### DIFF
--- a/fluss-common/src/main/java/org/apache/fluss/config/ConfigOptions.java
+++ b/fluss-common/src/main/java/org/apache/fluss/config/ConfigOptions.java
@@ -701,15 +701,6 @@ public class ConfigOptions {
                             "The format of the server truststore file. Supported values are `JKS` "
                                     + "and `PKCS12`. The default is `JKS`.");
 
-    public static final ConfigOption<Duration> SERVER_SSL_RELOAD_INTERVAL =
-            key("security.ssl.reload.interval")
-                    .durationType()
-                    .defaultValue(Duration.ofMinutes(5))
-                    .withDescription(
-                            "How often the server checks the keystore/truststore files for changes and "
-                                    + "rebuilds the SSL context (certificate hot-reload). Set to 0 to "
-                                    + "disable periodic reloading.");
-
     public static final ConfigOption<Integer> TABLET_SERVER_ID =
             key("tablet-server.id")
                     .intType()
@@ -1630,15 +1621,6 @@ public class ConfigOptions {
                                     + "server hostname against the server certificate. The default "
                                     + "`https` enables hostname verification; an empty string disables "
                                     + "it (not recommended in production).");
-
-    public static final ConfigOption<Duration> CLIENT_SSL_RELOAD_INTERVAL =
-            key("client.security.ssl.reload.interval")
-                    .durationType()
-                    .defaultValue(Duration.ofMinutes(5))
-                    .withDescription(
-                            "How often the client checks the keystore/truststore files for changes and "
-                                    + "rebuilds the SSL context (certificate hot-reload). Set to 0 to "
-                                    + "disable periodic reloading.");
 
     public static final ConfigOption<MemorySize> CLIENT_SCANNER_LOG_FETCH_MAX_BYTES =
             key("client.scanner.log.fetch.max-bytes")
--- a/fluss-common/src/main/java/org/apache/fluss/config/ConfigOptions.java
+++ b/fluss-common/src/main/java/org/apache/fluss/config/ConfigOptions.java
@@ -609,6 +609,107 @@ public class ConfigOptions {
                                     + "from 'security.sasl.plain.credentials' when that credential map is set, "
                                     + "and can also be configured directly for compatibility.");
 
+    // ------------------------------------------------------------------------
+    //  Server-side TLS (transport encryption + mTLS) options
+    // ------------------------------------------------------------------------
+
+    public static final ConfigOption<List<String>> SERVER_SSL_ENABLED_LISTENERS =
+            key("security.ssl.enabled.listeners")
+                    .stringType()
+                    .asList()
+                    .noDefaultValue()
+                    .withDescription(
+                            "A comma-separated list of listener names for which TLS transport "
+                                    + "encryption is enabled, e.g. `CLIENT,INTERNAL`. A keystore "
+                                    + "(`security.ssl.keystore.path`) must be configured when any "
+                                    + "listener is listed here. Listeners not listed accept plaintext "
+                                    + "connections. TLS is orthogonal to the authentication protocol "
+                                    + "configured via `security.protocol.map`.");
+
+    public static final ConfigOption<List<String>> SERVER_SSL_ENABLED_PROTOCOLS =
+            key("security.ssl.enabled.protocols")
+                    .stringType()
+                    .asList()
+                    .defaultValues("TLSv1.2", "TLSv1.3")
+                    .withDescription(
+                            "The list of TLS protocols enabled for incoming connections. "
+                                    + "`TLSv1.2` is kept in the default list for compatibility with "
+                                    + "JDK 8 builds older than 8u261 (which do not support TLS 1.3).");
+
+    public static final ConfigOption<List<String>> SERVER_SSL_CIPHER_SUITES =
+            key("security.ssl.cipher.suites")
+                    .stringType()
+                    .asList()
+                    .noDefaultValue()
+                    .withDescription(
+                            "A comma-separated list of cipher suites enabled for TLS connections. "
+                                    + "If not set, the JDK/provider defaults for the negotiated "
+                                    + "protocol are used.");
+
+    public static final ConfigOption<String> SERVER_SSL_KEYSTORE_PATH =
+            key("security.ssl.keystore.path")
+                    .stringType()
+                    .noDefaultValue()
+                    .withDescription(
+                            "The location of the keystore file holding the server certificate and "
+                                    + "private key. Required when any listener is listed in "
+                                    + "`security.ssl.enabled.listeners`.");
+
+    public static final ConfigOption<Password> SERVER_SSL_KEYSTORE_PASSWORD =
+            key("security.ssl.keystore.password")
+                    .passwordType()
+                    .noDefaultValue()
+                    .withDescription("The password to access the server keystore.");
+
+    public static final ConfigOption<String> SERVER_SSL_KEYSTORE_TYPE =
+            key("security.ssl.keystore.type")
+                    .stringType()
+                    .defaultValue("JKS")
+                    .withDescription(
+                            "The format of the server keystore file. Supported values are `JKS` and "
+                                    + "`PKCS12`. The default is `JKS`.");
+
+    public static final ConfigOption<Password> SERVER_SSL_KEY_PASSWORD =
+            key("security.ssl.key.password")
+                    .passwordType()
+                    .noDefaultValue()
+                    .withDescription(
+                            "The password of the private key in the server keystore. If not set, the "
+                                    + "keystore password (`security.ssl.keystore.password`) is used.");
+
+    public static final ConfigOption<String> SERVER_SSL_TRUSTSTORE_PATH =
+            key("security.ssl.truststore.path")
+                    .stringType()
+                    .noDefaultValue()
+                    .withDescription(
+                            "The location of the truststore file holding the certificates trusted "
+                                    + "by the server. Required for `mTLS` listeners (a listener whose "
+                                    + "`security.protocol.map` entry is `mTLS`), to validate client "
+                                    + "certificates.");
+
+    public static final ConfigOption<Password> SERVER_SSL_TRUSTSTORE_PASSWORD =
+            key("security.ssl.truststore.password")
+                    .passwordType()
+                    .noDefaultValue()
+                    .withDescription("The password to access the server truststore.");
+
+    public static final ConfigOption<String> SERVER_SSL_TRUSTSTORE_TYPE =
+            key("security.ssl.truststore.type")
+                    .stringType()
+                    .defaultValue("JKS")
+                    .withDescription(
+                            "The format of the server truststore file. Supported values are `JKS` "
+                                    + "and `PKCS12`. The default is `JKS`.");
+
+    public static final ConfigOption<Duration> SERVER_SSL_RELOAD_INTERVAL =
+            key("security.ssl.reload.interval")
+                    .durationType()
+                    .defaultValue(Duration.ofMinutes(5))
+                    .withDescription(
+                            "How often the server checks the keystore/truststore files for changes and "
+                                    + "rebuilds the SSL context (certificate hot-reload). Set to 0 to "
+                                    + "disable periodic reloading.");
+
     public static final ConfigOption<Integer> TABLET_SERVER_ID =
             key("tablet-server.id")
                     .intType()
@@ -1432,6 +1533,112 @@ public class ConfigOptions {
                                     + "This is useful for standalone tools or scripts that run outside "
                                     + "the server JVM but still need to load authentication plugins "
                                     + "shipped in plugins/.");
+
+    // ------------------------------------------------------------------------
+    //  Client-side TLS (transport encryption + mTLS) options
+    // ------------------------------------------------------------------------
+
+    public static final ConfigOption<Boolean> CLIENT_SSL_ENABLED =
+            key("client.security.ssl.enabled")
+                    .booleanType()
+                    .defaultValue(false)
+                    .withDescription(
+                            "Whether the client establishes TLS-encrypted connections to the server. "
+                                    + "Must match the security posture of the listener the client "
+                                    + "connects to (see `security.ssl.enabled.listeners` on the server).");
+
+    public static final ConfigOption<List<String>> CLIENT_SSL_ENABLED_PROTOCOLS =
+            key("client.security.ssl.enabled.protocols")
+                    .stringType()
+                    .asList()
+                    .defaultValues("TLSv1.2", "TLSv1.3")
+                    .withDescription(
+                            "The list of TLS protocols enabled for client connections. `TLSv1.2` is "
+                                    + "kept in the default list for compatibility with JDK 8 builds "
+                                    + "older than 8u261.");
+
+    public static final ConfigOption<List<String>> CLIENT_SSL_CIPHER_SUITES =
+            key("client.security.ssl.cipher.suites")
+                    .stringType()
+                    .asList()
+                    .noDefaultValue()
+                    .withDescription(
+                            "A comma-separated list of cipher suites enabled for client TLS "
+                                    + "connections. If not set, JDK/provider defaults are used.");
+
+    public static final ConfigOption<String> CLIENT_SSL_TRUSTSTORE_PATH =
+            key("client.security.ssl.truststore.path")
+                    .stringType()
+                    .noDefaultValue()
+                    .withDescription(
+                            "The location of the truststore file holding the certificates the client "
+                                    + "uses to verify the server certificate. If unset, the JVM "
+                                    + "default trust store is used.");
+
+    public static final ConfigOption<Password> CLIENT_SSL_TRUSTSTORE_PASSWORD =
+            key("client.security.ssl.truststore.password")
+                    .passwordType()
+                    .noDefaultValue()
+                    .withDescription("The password to access the client truststore.");
+
+    public static final ConfigOption<String> CLIENT_SSL_TRUSTSTORE_TYPE =
+            key("client.security.ssl.truststore.type")
+                    .stringType()
+                    .defaultValue("JKS")
+                    .withDescription(
+                            "The format of the client truststore file. Supported values are `JKS` "
+                                    + "and `PKCS12`. The default is `JKS`.");
+
+    public static final ConfigOption<String> CLIENT_SSL_KEYSTORE_PATH =
+            key("client.security.ssl.keystore.path")
+                    .stringType()
+                    .noDefaultValue()
+                    .withDescription(
+                            "The location of the keystore file holding the client certificate and "
+                                    + "private key. Only needed for mutual TLS, where the client "
+                                    + "presents a certificate to the server.");
+
+    public static final ConfigOption<Password> CLIENT_SSL_KEYSTORE_PASSWORD =
+            key("client.security.ssl.keystore.password")
+                    .passwordType()
+                    .noDefaultValue()
+                    .withDescription("The password to access the client keystore.");
+
+    public static final ConfigOption<String> CLIENT_SSL_KEYSTORE_TYPE =
+            key("client.security.ssl.keystore.type")
+                    .stringType()
+                    .defaultValue("JKS")
+                    .withDescription(
+                            "The format of the client keystore file. Supported values are `JKS` and "
+                                    + "`PKCS12`. The default is `JKS`.");
+
+    public static final ConfigOption<Password> CLIENT_SSL_KEY_PASSWORD =
+            key("client.security.ssl.key.password")
+                    .passwordType()
+                    .noDefaultValue()
+                    .withDescription(
+                            "The password of the private key in the client keystore. If not set, the "
+                                    + "keystore password (`client.security.ssl.keystore.password`) is "
+                                    + "used.");
+
+    public static final ConfigOption<String> CLIENT_SSL_ENDPOINT_IDENTIFICATION_ALGORITHM =
+            key("client.security.ssl.endpoint.identification.algorithm")
+                    .stringType()
+                    .defaultValue("https")
+                    .withDescription(
+                            "The endpoint identification algorithm used by the client to verify the "
+                                    + "server hostname against the server certificate. The default "
+                                    + "`https` enables hostname verification; an empty string disables "
+                                    + "it (not recommended in production).");
+
+    public static final ConfigOption<Duration> CLIENT_SSL_RELOAD_INTERVAL =
+            key("client.security.ssl.reload.interval")
+                    .durationType()
+                    .defaultValue(Duration.ofMinutes(5))
+                    .withDescription(
+                            "How often the client checks the keystore/truststore files for changes and "
+                                    + "rebuilds the SSL context (certificate hot-reload). Set to 0 to "
+                                    + "disable periodic reloading.");
 
     public static final ConfigOption<MemorySize> CLIENT_SCANNER_LOG_FETCH_MAX_BYTES =
             key("client.scanner.log.fetch.max-bytes")
--- a/fluss-rpc/src/main/java/org/apache/fluss/rpc/netty/ssl/SslConfig.java
+++ b/fluss-rpc/src/main/java/org/apache/fluss/rpc/netty/ssl/SslConfig.java
@@ -27,6 +27,7 @@ import javax.annotation.Nullable;
 import java.time.Duration;
 import java.util.Collections;
 import java.util.List;
+import java.util.Optional;
 
 import static org.apache.fluss.utils.Preconditions.checkArgument;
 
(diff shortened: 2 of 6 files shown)


fluss-common/src/main/java/org/apache/fluss/config/ConfigOptions.java
```diff
@@ -609,6 +609,107 @@ public class ConfigOptions {
                                     + "from 'security.sasl.plain.credentials' when that credential map is set, "
                                     + "and can also be configured directly for compatibility.");
 
+    // ------------------------------------------------------------------------
+    //  Server-side TLS (transport encryption + mTLS) options
+    // ------------------------------------------------------------------------
+
+    public static final ConfigOption<List<String>> SERVER_SSL_ENABLED_LISTENERS =
```
affo, line 616: I wonder if there is a way to mark these options as "experimental" for the time being 🤔 

My concern is that these options are now exposed to the outer world without anything wired, and this may compromise the release (say that we don't finalize the feature and it is half baked).

I would need the feedback of a committer here, or a PMC even :) @fresh-borzoni would you feel like it?

fluss-rpc/src/main/java/org/apache/fluss/rpc/netty/ssl/SslConfig.java
```diff
@@ -0,0 +1,192 @@
...
+        this.keyPassword = keyPassword;
+        this.truststorePath = truststorePath;
+        this.truststorePassword = truststorePassword;
+        this.truststoreType = truststoreType;
+        this.endpointIdentificationAlgorithm = endpointIdentificationAlgorithm;
+        this.reloadInterval = reloadInterval;
+    }
+
+    /** Build and validate the server-side TLS configuration. */
+    public static SslConfig fromServerConfig(Configuration conf) {
```
affo: I do understand that this contribution does not switch yet on `SERVER_SSL_ENABLED_LISTENERS` to enforce the validation that remains implicit, but this also makes sense as we don't have yet the wiring and those methods stay not invoked for now.

In the future one should extract the SSL config by checking that.

But that would turn to be cumbersome as part of the logic of checking whether SSL is enabled would fall out of the config class itself.

I wonder if it would be a neater approach to embed right now all this knowledge into this class, may return an `Optional` to embed the semantics of the fact that SSL may be there or not.

affo (commented):
The work looks very good and aligned with the FIP 1:1, thank you! 🤝 

I only have a couple doubts marked as a comments 🤝

fluss-rpc/src/main/java/org/apache/fluss/rpc/netty/ssl/SslConfig.java
MicheleGuerriero: Sounds good to me. And since nothing consumes these factory methods yet, this PR is the cheapest moment to fix this. I have changed it in [c31c730](https://github.com/apache/fluss/pull/3813/changes/c31c730a98213827474131769c9f3b3899e5c010).                                                                                                                                                                

One nuance: Optional can only encode the "is TLS configured at all" question. On the server side enablement is per-listener, so the pipeline wiring will still need the listener set for per-listener handler installation. To keep that knowledge adjacent to the config class, I have also exposed `enabledListeners()` on the server-side `SslConfig`, so callers read the listener set from the parsed config object instead of going back to the raw `Configuration`.

affo (commented):
Amazing job, I sent en email at `dev@` to see whether it would be best to merge this contributions on a separate branch and merge after 1.0 feature freeze, waiting for consensus there on how to operate

fluss-rpc/src/main/java/org/apache/fluss/rpc/netty/ssl/SslContextFactory.java
```diff
@@ -0,0 +1,195 @@
...
+                            config.keystorePassword(),
+                            config.keyPassword());
+            SslContextBuilder builder =
+                    SslContextBuilder.forServer(kmf)
+                            .sslProvider(SslProvider.JDK)
+                            .protocols(config.enabledProtocols());
+            if (!config.cipherSuites().isEmpty()) {
+                builder.ciphers(config.cipherSuites());
+            }
+            if (config.truststorePath() != null) {
```
fresh-borzoni, line 80: When no truststore is configured this skips trustManager(), so the context falls back to the JVM default truststore and setNeedClientAuth(true) at line 141 then accepts any client cert chaining to a public CA in cacerts.
FIP-29 wants an mTLS listener without security.ssl.truststore.path to refuse to start. 
Should SslConfig read security.protocol.map so it can be enforced?

fluss-common/src/main/java/org/apache/fluss/config/ConfigOptions.java
```diff
@@ -609,6 +609,107 @@ public class ConfigOptions {
...
+
+    public static final ConfigOption<String> SERVER_SSL_TRUSTSTORE_TYPE =
+            key("security.ssl.truststore.type")
+                    .stringType()
+                    .defaultValue("JKS")
+                    .withDescription(
+                            "The format of the server truststore file. Supported values are `JKS` "
+                                    + "and `PKCS12`. The default is `JKS`.");
+
+    public static final ConfigOption<Duration> SERVER_SSL_RELOAD_INTERVAL =
```
fresh-borzoni: This and client.security.ssl.reload.interval describe hot-reload that isn't implemented yet, and docgen reflects over every option in this class, so they land in the published config table as soon as it's regenerated. 
Can the reload options come with the reloader instead?

fluss-rpc/src/main/java/org/apache/fluss/rpc/netty/ssl/SslConfig.java
```diff
@@ -0,0 +1,231 @@
...
+     */
+    public static Optional<SslConfig> fromServerConfig(Configuration conf) {
+        List<String> enabledListeners =
+                orEmpty(conf.get(ConfigOptions.SERVER_SSL_ENABLED_LISTENERS));
+        if (enabledListeners.isEmpty()) {
+            return Optional.empty();
+        }
+
+        String keystorePath = conf.getString(ConfigOptions.SERVER_SSL_KEYSTORE_PATH);
+        checkArgument(
```
fresh-borzoni: nit: Should this be IllegalConfigurationException? That's what LocalDiskManager throws for bad server config?

fluss-rpc/src/test/java/org/apache/fluss/rpc/netty/ssl/SslContextFactoryTest.java
```diff
@@ -0,0 +1,241 @@
...
+
+        SslHandler noVerifyHandler =
+                SslContextFactory.createClientSslHandler(
+                        sslContext, ByteBufAllocator.DEFAULT, "localhost", 9123, "");
+        assertThat(noVerifyHandler.engine().getSSLParameters().getEndpointIdentificationAlgorithm())
+                .isNull();
+    }
+
+    @Test
+    void testServerSslHandlerClientAuthRequirement() {
```
fresh-borzoni, line 133: This checks the flag but not that a client without a cert gets rejected. testServerAndClientNegotiateTls already pumps a handshake, can you add one with requireClientAuth=true and no client keystore? 

PKCS12 is advertised in the keystore.type descriptions and untested too.

fluss-rpc/src/test/java/org/apache/fluss/rpc/netty/ssl/TestSslUtils.java
```diff
@@ -0,0 +1,107 @@
...
+    public static final String PASSWORD = "test-password";
+
+    /** The listener name {@link #setServerSslConfig} enables TLS for. */
+    public static final String TLS_LISTENER = "CLIENT";
+
+    private TestSslUtils() {}
+
+    /** Generate a self-signed certificate whose subject CN is {@code fqdn}. */
+    public static SelfSignedCertificate generateCertificate(String fqdn) throws Exception {
+        return new SelfSignedCertificate(fqdn);
```
fresh-borzoni, line 49: SelfSignedCertificate needs BouncyCastle or the sun.security.x509 internals, so all 12 tests error on JDK 17 and pass on 11. CI only runs tests on 11 so it stays green. --add opens=java.base/sun.security.x509=ALL-UNNAMED in extraJavaTestArgs makes them pass on 17, or a test scoped bcpkix(I guess this is better since it will work on jdk 21 as well)

Can we do one of these?

fluss-common/src/main/java/org/apache/fluss/config/ConfigOptions.java
fresh-borzoni: @affo Let's just move with feature branch until the feature is complete and then include it into the appropriate release, it's the least obtrusive approach

fluss-rpc/src/main/java/org/apache/fluss/rpc/netty/ssl/SslContextFactory.java
```diff
@@ -0,0 +1,195 @@
...
+                builder.trustManager(
+                        trustManagerFactory(
+                                config.truststorePath(),
+                                config.truststoreType(),
+                                config.truststorePassword()));
+            }
+            return builder.build();
```
fresh-borzoni, line 87: The engine is created per connection in createServerSslHandler, and that's the only place these protocol and cipher names get validated. So a typo in security.ssl.enabled.protocols or security.ssl.cipher.suites starts the server cleanly and then throws IllegalArgumentException: Unsupported protocol, while building the handler for every TLS connection, without naming the option that's wrong. 
An empty enabled.protocols is worse as the engine comes up with no protocols enabled and every handshake just fails.

fresh-borzoni (commented):
@MicheleGuerriero Thank you for the PR, I left some comments, PTAL

fluss-common/src/main/java/org/apache/fluss/config/ConfigOptions.java
MicheleGuerriero: Removed both in `5b5b115`. `security.ssl.reload.interval` and `client.security.ssl.reload.interval` are gone, along with the `SslConfig.reloadInterval` field they fed, and they'll come back in the PR that actually adds the reloader.

fluss-rpc/src/main/java/org/apache/fluss/rpc/netty/ssl/SslConfig.java
MicheleGuerriero: Fixed in `f53277a`

fluss-rpc/src/main/java/org/apache/fluss/rpc/netty/ssl/SslContextFactory.java
MicheleGuerriero: Confirmed all three cases against the JDK provider before fixing, and it's exactly as you describe — the context builds fine in every one of them.
    
Fixed in `f3cd33e`: `fromServerConfig`/`fromClientConfig` now reject unsupported protocol and cipher suite names and an empty protocol list, throwing `IllegalConfigurationException` that names the exact option (server or client key, since the validator takes the ConfigOption rather than just the values), the offending entries, and the supported set. Supported names come from one throwaway SSLEngine off the default SSLContext.

fluss-rpc/src/test/java/org/apache/fluss/rpc/netty/ssl/SslContextFactoryTest.java
MicheleGuerriero: Both added in `172dfd1`. I extracted the handshake pump from `testServerAndClientNegotiateTls` as you suggested, and built three tests on it: an mTLS server rejecting a client that presents no certificate (Empty client certificate chain), the same server accepting one that presents a trusted certificate, and a full mutual handshake with PKCS12 keystores and truststores on both sides.

fluss-rpc/src/test/java/org/apache/fluss/rpc/netty/ssl/TestSslUtils.java
MicheleGuerriero: Went with the test-scoped bcpkix as you suggested, fixed in `d4b39c6`. Netty's SelfSignedCertificate prefers BouncyCastle when it's on the classpath, so the dependency alone is enough and TestSslUtils didn't need rewriting.

fluss-rpc/src/main/java/org/apache/fluss/rpc/netty/ssl/SslContextFactory.java
MicheleGuerriero: `SslConfig` now reads `security.protocol.map`. Fixed in 54bf270.  `fromServerConfig` intersects the protocol map with `security.ssl.enabled.listeners`, collects the listeners whose protocol is `mTLS` (matched with `equalsIgnoreCase`, same as `AuthenticationFactory` matches authentication protocol names), and refuses the configuration when any of them has no `security.ssl.truststore.path` — naming the listener and saying why, rather than letting the context inherit `cacerts`. It's exposed as `requiresClientAuth(listener)` / `clientAuthListeners()`, so the pipeline wiring reads the requirement off the parsed config instead of the raw `Configuration`.

fluss-rpc/src/main/java/org/apache/fluss/rpc/netty/ssl/SslContextFactory.java
```diff
@@ -0,0 +1,197 @@
...
+        } catch (Exception e) {
+            throw new FlussRuntimeException("Failed to build the server SSL context.", e);
```
polyzos, line 97: make the error messages more descriptive, so the users can understand why things fail.. As it is right now, it would be hard for a user to identify the reason of failure - for example, paths, passwords etc.

fluss-rpc/src/main/java/org/apache/fluss/rpc/netty/ssl/SslContextFactory.java
```diff
@@ -0,0 +1,197 @@
...
+                builder.keyManager(
+                        keyManagerFactory(
+                                config.keystorePath(),
+                                config.keystoreType(),
+                                config.keystorePassword(),
+                                config.keyPassword()));
+            }
+            return builder.build();
+        } catch (Exception e) {
+            throw new FlussRuntimeException("Failed to build the client SSL context.", e);
```
polyzos, line 139: same

fluss-rpc/src/main/java/org/apache/fluss/rpc/netty/ssl/SslContextFactory.java
```diff
@@ -0,0 +1,197 @@
...
+        KeyStore trustStore = loadKeyStore(path, type, storePassword);
+        TrustManagerFactory tmf =
+                TrustManagerFactory.getInstance(TrustManagerFactory.getDefaultAlgorithm());
+        tmf.init(trustStore);
+        return tmf;
+    }
+
+    private static KeyStore loadKeyStore(String path, String type, String password)
+            throws Exception {
+        KeyStore keyStore = KeyStore.getInstance(type);
```
polyzos: `type` isn't validated at parse time, a typo starts the server and fails late.

fluss-rpc/src/main/java/org/apache/fluss/rpc/netty/ssl/SslConfig.java
```diff
@@ -0,0 +1,374 @@
...
+        if (keystorePath == null) {
+            throw new IllegalConfigurationException(
+                    "'%s' must be configured when any listener enables TLS via '%s'.",
+                    ConfigOptions.SERVER_SSL_KEYSTORE_PATH.key(),
+                    ConfigOptions.SERVER_SSL_ENABLED_LISTENERS.key());
+        }
+
+        Map<String, String> protocolMap = conf.get(ConfigOptions.SERVER_SECURITY_PROTOCOL_MAP);
+        Set<String> clientAuthListeners =
+                enabledListeners.stream()
```
polyzos, line 160: `protocolMap.get(listener)` is exact-match while the value is `equalsIgnoreCase`. 
With enabled.listeners: client + protocol.map: CLIENT:mTLS misses; we have one-way TLS, no truststore, silent cacerts fallback. We can document exact-match requirement or normalize.

fluss-rpc/src/test/java/org/apache/fluss/rpc/netty/ssl/TestSslUtils.java
```diff
@@ -0,0 +1,138 @@
...
+
+/**
+ * Test helpers for generating self-signed certificates and JKS keystores/truststores on the fly, so
+ * TLS tests do not need committed key material. Backed by Netty's {@link SelfSignedCertificate},
+ * which generates the certificate with BouncyCastle when it is on the classpath — hence the
+ * test-scoped {@code bcpkix-jdk18on} dependency. Its fallback generator reaches into {@code
+ * sun.security.x509}, which needs {@code --add-opens} from JDK 9 on and stops working altogether
+ * from JDK 18 on.
+ */
+public class TestSslUtils {
```
polyzos, line 38: The current test only asserts the param is set. Add an end-to-end handshake with "https" against a cert whose CN ≠ host and assert rejection.

And maybe it's good to add a handshake where the client truststore does not contain the server cert and

polyzos:
@MicheleGuerriero thank you for your contribution.. Overall LGTM 👍  I left some comments, and the CI needs fixing, it seems

fluss-rpc/src/main/java/org/apache/fluss/rpc/netty/ssl/SslConfig.java
MicheleGuerriero: Documented rather than normalized — ed758ad.

The asymmetry mirrors how the server resolves each half at runtime: `FlussProtocolPlugin` picks a listener's authenticator with a plain `suppliers.get(listenerName)` on `security.protocol.map`, while `AuthenticationFactory` matches a plugin to a protocol name with `equalsIgnoreCase`. If I matched listener names loosely here, `SslConfig` would classify a listener as mTLS and demand a truststore for it, while the server still authenticates it as PLAINTEXT, a disagreement between the two layers instead of one wrong but consistent outcome.

fluss-rpc/src/main/java/org/apache/fluss/rpc/netty/ssl/SslContextFactory.java
MicheleGuerriero: Fixed in bdcb58d. `SslConfig` now validates the store types at parse time and throws `IllegalConfigurationException` naming the exact option (server or client), the bad value, and the types the JVM does support.

fluss-rpc/src/main/java/org/apache/fluss/rpc/netty/ssl/SslContextFactory.java
MicheleGuerriero: Fixed in f87dd52. Each store now carries the options that configure it, so the failures an operator can actually cause name both the file and the option.

Same on the client side, naming the `client.security.ssl.*` options. The generic "Failed to build the SSL context" wrapper is now only around the context build itself, so it no longer swallows these.

fluss-rpc/src/test/java/org/apache/fluss/rpc/netty/ssl/TestSslUtils.java
MicheleGuerriero: Both added in 0f3a107. The handshake helper now takes the host the client dials and the endpoint identification algorithm, and three tests run on it: a certificate issued for another host is rejected with `https` set, the same handshake succeeds when the certificate matches the host and a server certificate outside the client truststore is rejected.

MicheleGuerriero:
Thanks for the review @polyzos! CI should be fixed by 1cb84fb . Could you approve the workflow runs when you get a chance?
I've replied on the individual threads for the rest of your comments.

polyzos:
@fresh-borzoni do you have any further comments here? If not, we can merge. @MicheleGuerriero, maybe it's also worth rebasing first.

fluss-rpc/src/main/java/org/apache/fluss/rpc/netty/ssl/SslConfig.java
```diff
@@ -0,0 +1,420 @@
...
+
+        // The listener name is looked up exactly and the protocol name compared ignoring case,
+        // because that is how each is resolved at runtime: FlussProtocolPlugin selects a listener's
+        // authenticator with a plain map lookup on security.protocol.map, while
+        // AuthenticationFactory
+        // matches a plugin to a protocol name with equalsIgnoreCase. Matching listener names
+        // loosely
+        // here would classify a listener as mTLS that the server then authenticates as PLAINTEXT.
+        Map<String, String> protocolMap = conf.get(ConfigOptions.SERVER_SECURITY_PROTOCOL_MAP);
+        Set<String> clientAuthListeners =
```
fresh-borzoni: The FIP validation table says an mTLS listener that isn't in security.ssl.enabled.listeners should refuse to start. 

Here it's accepted, and testMtlsListenerWithoutTlsNeedsNoTruststore asserts exactly that. It can't happen yet since there's no mTLS plugin, but this class already reads security.protocol.map, so the check fits here. Can we add it and flip that test?

fluss-common/src/main/java/org/apache/fluss/config/ConfigOptions.java
```diff
@@ -609,6 +609,102 @@ public class ConfigOptions {
                                     + "from 'security.sasl.plain.credentials' when that credential map is set, "
                                     + "and can also be configured directly for compatibility.");
 
+    // ------------------------------------------------------------------------
+    //  Server-side TLS (transport encryption + mTLS) options
+    // ------------------------------------------------------------------------
+
+    public static final ConfigOption<List<String>> SERVER_SSL_ENABLED_LISTENERS =
```
fresh-borzoni: Agree we shouldn't match loosely. But a typo here means the listener silently runs plaintext, and with SASL that means credentials in clear. 
Can we reject names that aren't in bind.listeners? Fine to do it in #3792 if that's easier. security.protocol.map has the same hole today - a typo'd key falls back to PLAINTEXT.

fluss-rpc/src/main/java/org/apache/fluss/rpc/netty/ssl/SslConfig.java
```diff
@@ -0,0 +1,420 @@
...
+                        truststoreType,
+                        ""));
+    }
+
+    /**
+     * Build and validate the client-side TLS configuration, or {@link Optional#empty()} when TLS is
+     * disabled (i.e. {@code client.security.ssl.enabled} is false).
+     */
+    public static Optional<SslConfig> fromClientConfig(Configuration conf) {
+        if (!conf.get(ConfigOptions.CLIENT_SSL_ENABLED)) {
```
fresh-borzoni: Same for the client rows of the FIP table: client.security.protocol=mTLS without client.security.ssl.enabled, or without a keystore, should fail. Now it's accepted.

fluss-rpc/src/main/java/org/apache/fluss/rpc/netty/ssl/SslContextFactory.java
```diff
@@ -0,0 +1,320 @@
...
+     * Build a server {@link SslContext} from the {@code security.ssl.*} configuration, or {@link
+     * Optional#empty()} when TLS is not enabled for any listener.
+     */
+    public static Optional<SslContext> createServerSslContext(Configuration conf) {
+        return SslConfig.fromServerConfig(conf).map(SslContextFactory::createServerSslContext);
+    }
+
+    /** Build a server {@link SslContext} from a parsed {@link SslConfig}. */
+    public static SslContext createServerSslContext(SslConfig config) {
+        KeyManagerFactory kmf =
```
fresh-borzoni: If the keystore has no private key (e.g. someone points it at the truststore), the context builds fine and every handshake fails with No available authentication scheme. 
Can we check for a key entry when loading it?

fluss-rpc/src/main/java/org/apache/fluss/rpc/netty/ssl/SslConfig.java
```diff
@@ -0,0 +1,420 @@
...
+                        enabledProtocols,
+                        cipherSuites,
+                        keystorePath,
+                        conf.getString(ConfigOptions.CLIENT_SSL_KEYSTORE_PASSWORD),
+                        keystoreType,
+                        conf.getString(ConfigOptions.CLIENT_SSL_KEY_PASSWORD),
+                        truststorePath,
+                        conf.getString(ConfigOptions.CLIENT_SSL_TRUSTSTORE_PASSWORD),
+                        truststoreType,
+                        conf.getString(
```
fresh-borzoni: nit: htps passes parsing and then fails every handshake. Shall we validate it like the protocols: https/ldaps ignoring case, or empty?

fresh-borzoni (commented):
@MicheleGuerriero Thank you, took a look, some comments, PTAL

PR #3816, opened 2026-07-31
[client] Prevent EOF during concurrent remote log downloads

### LINKED ISSUE #3815: [client] Concurrent remote log downloads can cause Unexpected EOF
### Search before asking

### Fluss version

main (development)

### Please describe the bug 🐞

When multiple scanners read the same remote log segment, duplicate downloads may target the same local file.

`LogFetcher` may reuse a `fetchableBuckets()` snapshot captured before an asynchronous metadata update, causing an already buffered bucket to be fetched again. `RemoteFileDownloader` then writes directly to the final path, truncating the file while another scanner is reading it.

The reader may fail with `CorruptMessageException: Unexpected EOF`.

Expected: an open reader can finish reading the downloaded segment.

Actual: a duplicate download may truncate the segment being read.

### Solution

- Recheck `fetchableBuckets()` under the `LogFetcher` lock before creating fetch requests.
- Download to a unique temporary file and atomically move the completed file to the final path.
- Add regression tests for both races.

This does not change any public API or storage format.

fxbing:
### Purpose

Linked issue: close #3815

`LogFetcher` could reuse a fetchable-bucket snapshot captured before a metadata update. If a response populated the buffer during that update, the same bucket could be fetched again. A duplicate remote-log download then truncated the local segment while an existing reader was still using it, causing `Unexpected EOF`.

### Brief change log

- Recompute fetchable buckets under the `LogFetcher` lock after metadata updates.
- Download each remote segment to a unique same-directory temporary file, then atomically publish the completed file.
- Clean up temporary files when a download or publish fails.

### Tests

- `LogFetcherTest` and `RemoteLogDownloaderTest`: 13 tests passed.
- `./mvnw -o -pl fluss-client validate`

### API and Format

### Documentation

### Generative AI disclosure

Yes. OpenAI Codex was used to assist with implementation and review.

### DIFF
--- a/fluss-client/src/main/java/org/apache/fluss/client/table/scanner/RemoteFileDownloader.java
+++ b/fluss-client/src/main/java/org/apache/fluss/client/table/scanner/RemoteFileDownloader.java
@@ -24,6 +24,7 @@ import org.apache.fluss.fs.FsPathAndFileName;
 import org.apache.fluss.fs.utils.FileDownloadSpec;
 import org.apache.fluss.fs.utils.FileDownloadUtils;
 import org.apache.fluss.utils.CloseableRegistry;
+import org.apache.fluss.utils.FileUtils;
 import org.apache.fluss.utils.IOUtils;
 import org.apache.fluss.utils.concurrent.ExecutorThreadFactory;
 
@@ -32,7 +33,6 @@ import java.io.IOException;
 import java.io.OutputStream;
 import java.nio.file.Files;
 import java.nio.file.Path;
-import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
 import java.util.concurrent.CompletableFuture;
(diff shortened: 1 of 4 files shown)


fluss-client/src/main/java/org/apache/fluss/client/table/scanner/log/LogFetcher.java
```diff
@@ -212,9 +212,10 @@ public void sendFetches() {
         List<TableBucket> fetchable = fetchableBuckets();
```
loserwang1024:
```suggestion

  checkAndUpdateMetadata(fetchableBuckets());

```

fluss-client/src/main/java/org/apache/fluss/client/table/scanner/RemoteFileDownloader.java
```diff
@@ -88,21 +88,31 @@ public CompletableFuture<Long> downloadFileAsync(
...
 
-            return IOUtils.copyBytes(inputStream, outputStream, false);
+            FileUtils.atomicMoveWithFallback(temporaryFile, targetFilePath, false);
+            return downloadBytes;
         } catch (Exception ex) {
-            throw new IOException(ex);
-        } finally {
-            closeableRegistry.forEach(IOUtils::closeQuietly);
+            IOException failure = new IOException(ex);
+            if (temporaryFile != null) {
```
loserwang1024, line 109: The temporaryFile won't be clear if throw OutOfMemoryError. Thus here need to modified as throw or, close in final scope(same as before)

fluss-client/src/main/java/org/apache/fluss/client/table/scanner/RemoteFileDownloader.java
fresh-borzoni, line 123: transferAllToDirectory right below still writes straight to the target via FileDownloadUtils. Should that get the same treatment? Odd to have both in one class.

fluss-client/src/test/java/org/apache/fluss/client/table/scanner/log/RemoteLogDownloaderTest.java
```diff
@@ -148,6 +156,136 @@ void testPrefetchNum() throws Exception {
         }
     }
 
+    @Test
+    void testDuplicateRemoteLogDownloadDoesNotBreakOpenReader() throws Exception {
+        CountDownLatch duplicateDownloadCopyStarted = new CountDownLatch(1);
+        CountDownLatch continueDuplicateDownload = new CountDownLatch(1);
+
+        class BlockingRemoteFileDownloader extends RemoteFileDownloader {
```
fresh-borzoni: nit: helper class inside the test method, rest of the file has them at the bottom.

fluss-client/src/main/java/org/apache/fluss/client/table/scanner/RemoteFileDownloader.java
```diff
@@ -88,21 +88,28 @@ public CompletableFuture<Long> downloadFileAsync(
...
-            return IOUtils.copyBytes(inputStream, outputStream, false);
+            FileUtils.atomicMoveWithFallback(temporaryFile, targetFilePath, false);
+            return downloadBytes;
         } catch (Exception ex) {
             throw new IOException(ex);
         } finally {
-            closeableRegistry.forEach(IOUtils::closeQuietly);
+            if (temporaryFile != null) {
+                Path fileToDelete = temporaryFile;
+                IOUtils.closeQuietly(() -> Files.deleteIfExists(fileToDelete));
```
fresh-borzoni: This closeQuietly overload swallows everything silently. Before, a failed delete showed up as a suppressed exception. Use closeQuietly(closeable, name) so a leaked temp file isn't invisible?

fresh-borzoni (commented):
@fxbing Thank you, left a couple of comments, PTAL

fluss-client/src/main/java/org/apache/fluss/client/table/scanner/RemoteFileDownloader.java
fxbing: Thanks for pointing this out. transferAllToDirectory() is only used for KV snapshot downloads. Each scanner downloads into its own UUID-scoped directory, and SnapshotFilesReader is created only after all files have finished downloading. Therefore, concurrent scanners do not share local target paths, and no reader can observe a file while it is being written.
The remote log path is different: duplicate fetches can write to the same canonical segment path while an existing scanner is already reading it. That is the race addressed by the temporary-file and atomic-move change.

fresh-borzoni: makes sense, thank you

fresh-borzoni (approved):
@fxbing Thank you, LGTM 👍

PR #3817, opened 2026-07-31
[flink][client] Close writer immediately on sink failover

### LINKED ISSUE #3809: [flink] Close writer immediately on sink failover
### Search before asking

### Fluss version

0.9.0 (latest release)

### Please describe the bug 🐞

Currently, the Fluss Flink sink closes the client through `Connection.close()`, which closes the underlying `WriterClient` with `Duration.ofMillis(Long.MAX_VALUE)`:

- `fluss-client/src/main/java/org/apache/fluss/client/FlussConnection.java`
- `writerClient.close(Duration.ofMillis(Long.MAX_VALUE))`

This means sink close may wait indefinitely for pending writer requests to finish. During Flink task failover or cancellation, if the sender still has records that cannot be sent downstream, the task may be blocked in close and failover cannot proceed promptly.

This behavior is unreasonable for Flink failover/cancel paths. The Kafka sink handles this differently: `FlinkKafkaInternalProducer#close()` closes immediately by calling `super.close(Duration.ZERO)`.
```java
@Override
public void close() {
    if (!closed) {
        LOG.debug("Closing immediately {}", this);
        super.close(Duration.ZERO);
        closed = true;
    }
}
```

### Solution

Fluss should provide a timeout-based close API that allows callers to choose between graceful close and immediate/fast close.For the Flink sink, close during task failover/cancel should use immediate close semantics, similar to Kafka's FlinkKafkaInternalProducer#close().

naivedogger:
### Purpose

Linked issue: close #3809

The Flink sink closed the client via `Connection.close()`, which closes the underlying `WriterClient` with an unbounded timeout. With pending records that cannot be sent (e.g. cluster unavailable during failover, with infinite writer retries by default), the sender's drain loop never exits and the task blocks in close, preventing failover from proceeding.

### Brief change log

- `Connection`: add a timeout-based `close(Duration)` API (default method, delegating to `close()`); zero/negative timeout means immediate close, abandoning unsent and in-flight requests.
- `FlussConnection`: implement `close(Duration)`; the no-arg `close()` keeps the previous graceful semantics (unbounded timeout).
- `WriterClient#close(Duration)`: a non-positive timeout now force-closes the sender up front instead of initiating a graceful close; when a graceful close times out, force-close the sender before interrupting the io thread, and bound the final wait (5s safety net) instead of reusing the caller timeout.
- `FlinkSinkWriter#close`: close the connection with `Duration.ZERO`. Data correctness is unaffected: records are flushed on checkpoints, and on failover/cancellation un-flushed records are replayed from the last checkpoint. This mirrors Kafka's `FlinkKafkaInternalProducer#close()`.

### Tests

- `FlinkSinkWriterTest#testCloseReturnsPromptlyWhenFlussUnavailable`: reproduces the failover scenario (cluster down + pending record + default infinite retries) and asserts close returns promptly. Verified it hangs and fails without the fix.
- Existing `FlinkSinkWriterTest` tests still pass, including `testCloseExceptionWhenFlussUnavailable` (close still surfaces async write errors).

### API and Format

Adds `Connection#close(Duration)` as a default method on the `@PublicEvolving` interface; existing implementations and callers are unaffected. No storage format changes.

### Documentation

No documentation changes.

### DIFF
--- a/fluss-client/src/main/java/org/apache/fluss/client/write/WriterClient.java
+++ b/fluss-client/src/main/java/org/apache/fluss/client/write/WriterClient.java
@@ -332,47 +332,40 @@ public class WriterClient {
         writerMetricGroup.close();
 
         if (sender != null) {
-            if (timeoutMs > 0) {
-                // graceful close: stop accepting new records, but keep the sender running to
-                // send out the pending records.
-                sender.initiateClose();
-            } else {
-                // immediate close: abandon all unsent and in-flight requests right away so
-                // that close never blocks (e.g. during Flink task failover/cancellation).
-                sender.forceClose();
-            }
+            sender.initiateClose();
         }
 
         if (ioThreadPool != null) {
             ioThreadPool.shutdown();
 
-            try {
-                if (!ioThreadPool.awaitTermination(timeoutMs, TimeUnit.MILLISECONDS)) {
-                    // the graceful close didn't finish within the timeout (or an immediate
-                    // close was requested): abandon the pending requests so that the sender
-                    // thread can exit its drain loop, and interrupt it to break any blocking
-                    // waits.
-                    if (sender != null) {
-                        sender.forceClose();
+            if (timeoutMs > 0) {
+                try {
+                    if (!ioThreadPool.awaitTermination(timeoutMs, TimeUnit.MILLISECONDS)) {
+                        LOG.warn("Writer graceful close timed out after {} ms.", timeoutMs);
                     }
-                    ioThreadPool.shutdownNow();
-
-                    if (!ioThreadPool.awaitTermination(
-                            FORCE_CLOSE_TERMINATION_TIMEOUT_MS, TimeUnit.MILLISECONDS)) {
-                        LOG.error("Failed to shutdown writer.");
-                    }
+                } catch (InterruptedException e) {
+                    LOG.error("Interrupted while waiting for writer sender thread.", e);
+                    Thread.currentThread().interrupt();
                 }
-            } catch (InterruptedException e) {
-                if (sender != null) {
-                    sender.forceClose();
-                }
-                ioThreadPool.shutdownNow();
-                Thread.currentThread().interrupt();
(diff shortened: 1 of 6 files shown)


fluss-client/src/main/java/org/apache/fluss/client/write/WriterClient.java
```diff
@@ -319,26 +326,46 @@ private Sender newSender(short acks, int retries) {
...
     public void close(Duration timeout) {
-        LOG.info("Closing writer.");
+        long timeoutMs = timeout.toMillis();
+        LOG.info("Closing writer with timeout {} ms.", timeoutMs);
 
         writerMetricGroup.close();
 
         if (sender != null) {
-            sender.initiateClose();
+            if (timeoutMs > 0) {
```
loserwang1024: Is it needed?if timeout <= 0, oThreadPool.awaitTermination will return false.

Image

fluss-client/src/main/java/org/apache/fluss/client/write/WriterClient.java
```diff
@@ -319,26 +326,46 @@ private Sender newSender(short acks, int retries) {
...
             ioThreadPool.shutdown();
 
             try {
-                if (!ioThreadPool.awaitTermination(timeout.toMillis(), TimeUnit.MILLISECONDS)) {
+                if (!ioThreadPool.awaitTermination(timeoutMs, TimeUnit.MILLISECONDS)) {
+                    // the graceful close didn't finish within the timeout (or an immediate
+                    // close was requested): abandon the pending requests so that the sender
+                    // thread can exit its drain loop, and interrupt it to break any blocking
+                    // waits.
+                    if (sender != null) {
```
loserwang1024: Please implement it referring  kafka.

Image

loserwang1024 (commented):
I left some comment

loserwang1024 (approved):
LGTM!

PR #3821, opened 2026-08-01
[rust] Fix Arrow limit scan across schema evolution

### LINKED ISSUE #3808: [rust] Limit scan fails to read Arrow log batches written before ADD COLUMN
### Search before asking

### Fluss version

0.9.0 (latest release)

### Please describe the bug 🐞

The Rust `LimitBatchScanner` log path does not handle schema evolution when reading Arrow log batches.

`decode_log_batch` creates a single `ArrowReadContext` using the current table schema and uses it to decode every returned `LogRecordBatch`. However, each log batch carries its own `schema_id` and may have been written using an older schema.

As a result, after an `ADD COLUMN`, a limit scan that includes batches written before the schema change fails because the decoder expects the newly added Arrow field to exist in the old batch.

### Steps to reproduce

1. Create an Arrow log table with the following schema:

```text
   id INT,
   name STRING
```

2. Append and flush one or more records.
3. Add a new column:
```
ALTER TABLE ... ADD COLUMN age BIGINT;
```

4. Create a limit scanner using the latest table metadata.
5. Scan a bucket containing records written before the schema change.

A minimal unit-level reproduction is to:
1. Build an Arrow log batch using the old two-column schema.
2. Call `decode_log_batch` with a `TableInfo` containing the new three-column schema.

### Solution

_No response_

slfan1989:
### Purpose

Linked issue: close #3808

The Rust `LimitBatchScanner` previously decoded every Arrow log batch using the current table schema, without considering the `schema_id` stored in each batch.

As a result, a limit scan could fail when reading batches written before an `ADD COLUMN`, because the decoder expected fields that were not present in the older Arrow batch.

This change makes the Arrow limit-scan path decode each batch using its write-time schema and align it with the current table schema, consistent with the regular log-fetch and KV read paths.

### Brief change log

- Build a fixed-schema `ReadContextResolver` for Arrow log limit scans.
- Resolve the source schema from each `LogRecordBatch` using its `schema_id`.
- Lazily fetch and cache historical schemas when they are not already available.
- Align decoded batches with the current target schema by stable column ID.
- Concatenate batches only after schema alignment.
- Apply column projection after batches have been aligned and concatenated.
- Add unit and integration coverage for batches written before and after `ADD COLUMN`.

### Tests

Add Some Test.

### API and Format

No.

### Documentation

No.

### DIFF
--- a/fluss-rust/crates/fluss/src/client/table/batch_scanner.rs
+++ b/fluss-rust/crates/fluss/src/client/table/batch_scanner.rs
@@ -18,12 +18,14 @@
 //! Bounded batch scanner backed by a single `LimitScanRequest`, polled with
 //! `next_batch` until it returns `None` (like `RecordBatchLogReader`).
 //!
-//! The KV branch decodes a [`ValueRecordBatch`], decoding each record against
-//! its own schema id via [`FixedSchemaDecoder`] so older records are projected
-//! onto the current schema (the same path as lookup).
+//! Both log and KV batches are decoded against their write-time schema id and
+//! aligned to the current table schema. The KV branch uses [`FixedSchemaDecoder`]
+//! (the same path as lookup), while the log branch resolves an Arrow read
+//! context per batch.
 
 use crate::client::ClientSchemaGetter;
 use crate::client::metadata::Metadata;
+use crate::client::table::read_context_resolver::ReadContextResolver;
 use crate::error::{ApiError, Error, FlussError, Result};
 use crate::metadata::{KvFormat, RowType, Schema, TableBucket, TableInfo};
 use crate::proto::ErrorResponse;
@@ -158,7 +160,9 @@ async fn run_limit_scan(pending: &PendingScan, bucket: &TableBucket) -> Result<S
     // Choose the payload format from table metadata, not the response's advisory
     // `is_log_table` flag.
     let (batch, base_offset) = if !pending.table_info.has_primary_key() {
-        decode_log_batch(&pending.table_info, projected, raw, limit)?
+        let resolver = create_log_read_context_resolver(&pending.table_info)?
+            .with_schema_getter(Arc::clone(&pending.schema_getter));
+        decode_log_batch(&pending.table_info, &resolver, projected, raw, limit).await?
     } else {
         // KV (primary-key) limit scan: no log offset, so base_offset is 0.
         let batch = decode_kv_batch(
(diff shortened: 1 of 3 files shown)


fresh-borzoni (commented):
@slfan1989 Thank you for the PR, but looks like CI/CD is failing on new tests

slfan1989:
> @slfan1989 Thank you for the PR, but looks like CI/CD is failing on new tests

@fresh-borzoni > Thank you for pointing this out. I’ve fixed the timing-dependent test assertion. Could you please review the PR again?

fluss-rust/crates/fluss/src/client/table/batch_scanner.rs
```diff
@@ -175,21 +179,18 @@ async fn run_limit_scan(pending: &PendingScan, bucket: &TableBucket) -> Result<S
...
-fn decode_log_batch(
+async fn decode_log_batch(
     table_info: &TableInfo,
+    resolver: &ReadContextResolver,
     projected_fields: Option<&[usize]>,
     raw: Vec<u8>,
     limit: usize,
 ) -> Result<(RecordBatch, i64)> {
-    let row_type = Arc::new(table_info.get_row_type().clone());
     let full_schema = to_arrow_schema(table_info.get_row_type())?;
```
fresh-borzoni: Why build the arrow schema twice, here and :248? Can we build it once?

fluss-rust/crates/fluss/src/client/table/batch_scanner.rs
```diff
@@ -434,14 +474,33 @@ mod tests {
         )
     }
 
+    fn build_table_info_at_schema(
```
fresh-borzoni: Could build_log_batch take the schema id instead? Then build_table_info_at_schema goes away and arg 3 stops meaning buckets in one helper and schema id in the other.

fluss-rust/crates/fluss/src/client/table/batch_scanner.rs
```diff
@@ -629,6 +721,71 @@ mod tests {
...
+        current_row.set_field(0, 2_i32);
+        current_row.set_field(1, "bob");
+        current_row.set_field(2, 30_i64);
+        raw.extend(build_log_batch(&target_table_info, &[current_row]));
+        let resolver = create_log_read_context_resolver(&target_table_info).expect("resolver");
+        resolver
+            .register_schema(1, source_table_info.get_schema())
+            .expect("register source schema");
+
+        let (batch, _) = decode_log_batch(&target_table_info, &resolver, None, raw, usize::MAX)
```
fresh-borzoni: Can you add the projection case here, projecting age (index 2), the column the old batch doesn't have, so it comes back as nulls? And one with limit = 2 across both batches, where the last 2 rows survive and the base offset moves by 1. Just so a future refactor can't quietly break the order.

fluss-rust/crates/fluss/src/client/table/batch_scanner.rs
```diff
@@ -229,6 +243,31 @@ fn decode_log_batch(
...
+        i16::try_from(table_info.get_schema_id()).map_err(|_| Error::UnexpectedError {
+            message: format!(
+                "Schema id {} does not fit in 16 bits — wire format violated",
+                table_info.get_schema_id()
+            ),
+            source: None,
+        })?;
+    Ok(
+        ReadContextResolver::new(schema_id, local_context, remote_context, None)
+            .with_fixed_schema(table_info.get_schema()),
```
fresh-borzoni, line 279: Why unconditional? LogFetcher::new only does this if fixed_schema. 
Should it error instead of ignoring the flag?

fresh-borzoni (commented):
@slfan1989 Thank you for the PR, left a couple of comments, PTAL

fluss-rust/crates/fluss/src/client/table/batch_scanner.rs
slfan1989: Thank you for pointing this out. The target Arrow schema is now built once and shared between the read-context resolver and the log batch decoder.

fluss-rust/crates/fluss/src/client/table/batch_scanner.rs
slfan1989: Thank you for the suggestion. I updated `build_log_batch` to take the schema ID explicitly and removed `build_table_info_at_schema`, so the third argument of `build_table_info_with_columns` consistently represents the bucket count.

fluss-rust/crates/fluss/src/client/table/batch_scanner.rs
slfan1989: Thank you for the suggestion. I added coverage for projecting the added `age` column from old batches and applying a limit across schema versions, including the resulting row order and base offset.

fluss-rust/crates/fluss/src/client/table/batch_scanner.rs
slfan1989: Thank you for pointing this out. Log-table `LimitBatchScanner` now returns an error for `with_fixed_schema(false)` instead of silently ignoring the flag. I also documented this limitation and added a regression test.

fresh-borzoni (approved):
@slfan1989 Thank you, LGTM 👍

PR #3822, opened 2026-08-02
[server] Support RemoteLogManifest to handle overlap RemoteLogSegments

### LINKED ISSUE #3835: [server] Support overlapping remote log segments via extending manifest to unblock active segment rolling
### Search before asking

### Motivation

Fluss tiers sealed log segments to remote storage per bucket. Replica leaders can roll segments at different offsets, so their physical segment boundaries are not guaranteed to match. After a leader switch, the new leader may need to upload a segment whose physical range overlaps data already present in remote storage:

```text
remote already has:      A physical=[0, 10)
new leader wants to add: B physical=[5, 20)
```

Rejecting `B` prevents tiering from advancing beyond offset 10, while treating both physical files as fully readable creates overlapping logical coverage. This becomes a routine blocker when active segments can be rolled on demand.

The remote manifest therefore needs to distinguish the physical range of each stored object from the logical range exposed by the current manifest.

### Expected behavior

Physical remote segments may overlap, but the logical view referenced by a manifest must remain ordered, non-overlapping, and continuous.

For the example above, the manifest should expose:

```text
A physical=[0, 10), logical=[0, 5)
B physical=[5, 20), logical=[5, 20)
```

Tiering progress must also remain independent from the current logical end. Replacing or expiring logical ranges must not cause already copied offsets to be uploaded again.

### Resolution

- Extend `RemoteLogSegment` with logical start/end offsets while retaining its physical range.
- Merge newly copied segments by clipping or replacing older logical ranges; reject additions that would create a gap.
- Persist `highestCopiedEndOffset` separately from the logical remote-log end and propagate it to replicas.
- Make logical segment lookup, timestamp lookup, and KV recovery respect the logical bounds.
- Preserve FetchLog v0 compatibility by stopping each response before a physical overlap and resuming from the client's advanced offset.
- Delete physical objects superseded by the new logical view only after the new manifest is committed.

The existing manifest version is retained. New manifest fields are optional: old manifests default logical ranges to their physical ranges and copy progress to the remote-log end. The internal commit/notification RPC fields use the same backward-compatible fallback.

Implemented by #3822.

### Anything else?

_No response_

### Willingness to contribute

gyang94:
### Purpose

Remote log segments created by different replica leaders can have different physical boundaries. After a leader switch, a newly sealed local segment may overlap data that is already stored remotely. The previous tiering flow assumed physically non-overlapping segments, so it could not continue copying in this case.

This change allows physical remote log objects to overlap while exposing a single continuous, non-overlapping logical log view.

Closes #3835.

### Brief change log

- Add logical start/end offsets to `RemoteLogSegment` while retaining physical offsets for remote object and index access.
- Merge copied segments into the existing manifest by clipping or replacing older logical ranges, and reject additions that introduce a gap.
- Track `highestCopiedEndOffset` independently from the current logical remote-log end so expiration or replacement does not lose tiering progress.
- Make offset lookup, timestamp lookup, and KV remote-log recovery honor logical segment bounds.
- Keep FetchLog v0 compatible by returning the maximal physically contiguous prefix before an overlap; the client resumes from its advanced offset in the next fetch.
- Remove superseded physical segment objects only after the updated manifest is committed.

### Tests

- Added manifest serde compatibility and overlap merge tests, including clipping, replacement, multiple additions, gaps, and retained copy progress.
- Added remote tablet tests for logical lookup and FetchLog v0 overlap handling.
- Added remote KV fetcher coverage for segments produced with different replica boundaries.
- Added internal RPC compatibility tests for `highestCopiedEndOffset`.

### API and Format

- No public client API changes.
- The remote manifest remains at version 1. Logical bounds and `highest_copied_end_offset` are optional fields and are omitted when they equal their legacy defaults.
- Existing manifests default logical bounds to physical bounds and copy progress to the remote-log end offset.
- The internal commit/notification RPCs add optional `highest_copied_end_offset` fields with the same legacy fallback.

### Documentation

No user-facing configuration or API changes.


fluss-server/src/test/java/org/apache/fluss/server/log/remote/RemoteLogManifestCommitterTest.java
```diff
@@ -0,0 +1,310 @@
...
+import java.util.concurrent.CompletableFuture;
+
+import static org.apache.fluss.record.TestData.DATA1_PHYSICAL_TABLE_PATH;
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.times;
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.when;
```
swuferhong: Do not use Mockito in tests. Instead, test by constructing instances of the relevant classes.

fluss-server/src/main/java/org/apache/fluss/server/coordinator/CoordinatorEventProcessor.java
```diff
@@ -389,13 +395,23 @@ private void initCoordinatorContext() throws Exception {
                 zooKeeperClient.getTabletServers(currentServers);
```
swuferhong, line 389: This check reads the ephemeral registration of every assigned replica. As a result, an offline follower （maybe server is restart） is treated as not supporting Manifest V2 and blocks all manifest commits from the leader.
Maybe we can check only alive assigned replicas through `coordinatorContext.getLiveTabletServers()`.

fluss-common/src/main/java/org/apache/fluss/remote/RemoteLogManifestReplacementPlanner.java
```diff
@@ -0,0 +1,241 @@
...
+        return new Result(planType, resultManifest, segmentsToUnreference);
+    }
+
+    /**
+     * Moves a continuous expired logical prefix to unreferenced metadata without deleting objects.
+     *
+     * <p>The last active reference is retained because the current V2 handle format represents only
+     * non-empty ranges. This is conservative: expiration may be delayed, but readable data is never
+     * removed early.
+     */
```
swuferhong: The expiration logic always retains the final manifest reference？ For an idle table, active-segment rolling may upload one final segment with no subsequent segment, causing expired remote data to be retained indefinitely.
Please support publishing an empty Manifest V2, for example through an empty-handle, so TTL can remove the last remote segment.

fluss-server/src/main/java/org/apache/fluss/server/metadata/ServerInfo.java
```diff
@@ -45,6 +50,7 @@ public class ServerInfo {
     private final Map<String, Endpoint> endpointMap;
     private final ServerType serverType;
     private final TabletServerResource resource;
+    private final Set<String> capabilities;
```
swuferhong: Manifest V2 should be enabled through capability negotiation rather than assuming that either `Coordinator` or `TabletServer` is upgraded first.

`TabletServers` should advertise their supported manifest version, and `Coordinator` should expose their V2 capability as well. The Coordinator should select the effective manifest version based on the capabilities of the Coordinator cluster and alive assigned TabletServers, then communicate that version to the leader through an existing response or metadata RPC.

A `TabletServer` must write V2 only after receiving an explicit version=2 selection. A missing version field should default to V1 for backward compatibility. 

This ensures:

- New TabletServers continue writing V1 when connected to an old Coordinator.

- New Coordinators continue accepting V1 while old TabletServers remain.

- V2 is activated only after all relevant components support it.

- Once a V2 handle is published, the bucket must not fall back to V1.

swuferhong: Why use `String` not use `enum`?

fluss-flink/fluss-flink-common/src/test/java/org/apache/fluss/flink/action/orphan/OrphanFilesCleanITCase.java
```diff
@@ -194,16 +194,16 @@ void mixedOrphanAndActiveFilesInSameBucket() throws Exception {
                         + "\"remote_log_segments\":["
```
swuferhong, line 194: `ActiveRefsFetcher.parseLogSegmentRelativePaths()` only includes segments from `manifest.getRemoteLogSegmentList().` Manifest V2 segments recorded in `getUnreferencedRemoteLogSegments()` are therefore classified as orphans and may be deleted before the configured V2 GC grace period expires.

fluss-server/src/main/java/org/apache/fluss/server/log/remote/DefaultRemoteLogStorage.java
```diff
@@ -236,6 +238,82 @@ public void deleteRemoteLogManifestSnapshot(FsPath remoteLogManifestPath)
...
+                new FsPath(
+                        FlussPaths.remoteLogTabletDir(remoteLogDir, physicalTablePath, tableBucket),
+                        REMOTE_LOG_METADATA_DIR_NAME);
+        try {
+            if (!fileSystem.exists(metadataDir)) {
+                return new ArrayList<>();
+            }
+            List<RemoteLogStorageObject> objects = new ArrayList<>();
+            for (FileStatus status : fileSystem.listStatus(metadataDir)) {
+                if (!status.isDir() && status.getPath().getName().endsWith(".manifest")) {
```
swuferhong: Use `FlussPaths.REMOTE_LOG_MANIFEST_FILE_SUFFIX`

fluss-common/src/main/java/org/apache/fluss/remote/UnreferencedRemoteLogSegment.java
```diff
@@ -0,0 +1,111 @@
...
+
+/** Metadata retained for a remote segment that no longer participates in logical reads. */
+@Internal
+public final class UnreferencedRemoteLogSegment {
+
+    /** The reason why a physical remote segment stopped participating in logical reads. */
+    public enum Reason {
+        REPLACED,
+        EXPIRED,
+        DELETE_STARTED
```
swuferhong: no use for this value `DELETE_STARTED`

fluss-server/src/main/java/org/apache/fluss/server/log/remote/RemoteLogManifestUpdatePlan.java
```diff
@@ -0,0 +1,127 @@
...
+
+    public List<RemoteLogSegment> segmentsToCopy() {
+        return segmentsToCopy;
+    }
+
+    public List<RemoteLogSegment> segmentsToUnreference() {
+        return segmentsToUnreference;
+    }
+
+    public List<PlanType> planTypes() {
```
swuferhong: no use

fluss-server/src/main/java/org/apache/fluss/server/log/remote/RemoteLogManifestUpdatePlan.java
```diff
@@ -0,0 +1,127 @@
...
+
+    public RemoteLogManifest resultManifest() {
+        return resultManifest;
+    }
+
+    public List<RemoteLogSegment> segmentsToCopy() {
+        return segmentsToCopy;
+    }
+
+    public List<RemoteLogSegment> segmentsToUnreference() {
```
swuferhong: no use

fluss-server/src/main/java/org/apache/fluss/server/log/remote/LogTieringTask.java
```diff
@@ -201,6 +231,437 @@ private void runOnce() throws InterruptedException {
...
+                appliedPlanTypes.add(candidatePlan.planType);
+                successfulNextCopyOffset = candidatePlan.enrichedLogSegment.nextSegmentOffset;
+                continue;
+            }
+            if (!copyPlannedSegment(logTablet, candidatePlan, metricGroup)) {
+                break;
+            }
+            copiedSegments.add(candidatePlan.remoteLogSegment);
+            RemoteLogManifestReplacementPlanner.Result replayed =
+                    RemoteLogManifestReplacementPlanner.plan(
```
swuferhong: `planCandidates()` already invokes the replacement planner for every candidate and advances its working manifest, but it retains only the first planType. After copying, this loop invokes the same planner again, discards the second planType, and uses only its manifest result.

Please combine planning, copying, and applying into a single loop, or retain the complete planning result in `CandidatePlan`, to avoid duplicate logic and the risk of the two planning paths diverging.

swuferhong (commented):
I left some comments:

fluss-common/src/main/java/org/apache/fluss/config/ConfigOptions.java
```diff
@@ -1091,6 +1091,33 @@ public class ConfigOptions {
...
+    public static final ConfigOption<Boolean> REMOTE_LOG_MANIFEST_V2_WRITER_ENABLED =
+            key("remote.log.manifest-v2-writer-enabled")
+                    .booleanType()
+                    .defaultValue(false)
+                    .withDescription(
+                            "Enables CAS publishing of Manifest V2 snapshots and overlap-aware "
+                                    + "remote segment replacement. Keep disabled until every "
+                                    + "remote-log reader supports manifest version dispatch.");
+
+    public static final ConfigOption<Boolean> REMOTE_LOG_MANIFEST_V2_GC_ENABLED =
```
swuferhong: Why do we need this option? Is there really a case where V2 is enabled but GC is not? I'd suggest removing it and letting the V2 enable flag control this.

fluss-server/src/main/java/org/apache/fluss/server/log/remote/RemoteLogTablet.java
```diff
@@ -132,6 +131,14 @@ public void registerMetrics(BucketMetricGroup bucketMetricGroup) {
                             });
                     metricGroup.gauge(MetricNames.LOG_END_OFFSET, () -> remoteLogEndOffset);
                     metricGroup.gauge(MetricNames.REMOTE_LOG_SIZE, this::getRemoteSizeInBytes);
+                    metricGroup.gauge(
+                            MetricNames.REMOTE_UNREFERENCED_BYTES,
+                            this::getUnreferencedSizeInBytes);
```
swuferhong: I'd suggest not introducing too many table-bucket-level metrics — could these be aggregated to the table level instead? Or added later on demand. We did a dedicated cleanup of bucket-level metrics before and kept only the truly important ones: too many bucket-level metrics will seriously overwhelm the collection agent.

fluss-common/src/main/java/org/apache/fluss/remote/RemoteLogManifestJsonSerde.java
```diff
@@ -73,27 +78,55 @@ public void serialize(RemoteLogManifest manifest, JsonGenerator generator) throw
...
     public RemoteLogManifest deserialize(JsonNode node) {
-        String databaseName = node.get(DATABASE_NAME_FIELD).asText();
-        String tableName = node.get(TABLE_NAME_FIELD).asText();
+        int version = required(node, VERSION_KEY).asInt();
+        if (version != RemoteLogManifest.VERSION_1 && version != RemoteLogManifest.VERSION_2) {
+            throw new IllegalArgumentException(
+                    "Unsupported remote log manifest version: " + version);
+        }
+        if (version == RemoteLogManifest.VERSION_1) {
+            rejectField(node, GENERATION_FIELD, version);
```
swuferhong: This check feels pretty useless. For V1, isn't it enough to just read the corresponding field? Why do we need this check at all?

fluss-common/src/main/java/org/apache/fluss/remote/RemoteLogManifest.java
```diff
@@ -20,47 +20,253 @@
...
+            endOffset = Math.max(endOffset, reference.logicalEndOffset());
+        }
+        return endOffset;
+    }
+
     public RemoteLogManifest trimAndMerge(
             List<RemoteLogSegment> deletedSegments, List<RemoteLogSegment> addedSegments) {
+        if (version != VERSION_1) {
+            throw new IllegalStateException("trimAndMerge only supports V1 manifests");
+        }
```
swuferhong: This path is unreachable, so adding this check doesn't add much value.

fluss-common/src/main/java/org/apache/fluss/remote/RemoteLogManifest.java
```diff
@@ -20,47 +20,253 @@
...
+        return references;
+    }
+
+    private List<RemoteLogSegmentReference> buildV2ActiveReferences(
+            @Nullable Long remoteLogStartOffset) {
+        List<RemoteLogSegmentReference> references = new ArrayList<>(remoteLogSegmentList.size());
+        for (int i = 0; i < remoteLogSegmentList.size(); i++) {
+            RemoteLogSegment segment = remoteLogSegmentList.get(i);
+            long logicalStartOffset =
+                    i == 0 ? remoteLogStartOffset.longValue() : segment.remoteLogStartOffset();
```
swuferhong: Is this method parameter actually nullable or non-null? If it's nullable, there's an NPE risk here.

fluss-server/src/main/java/org/apache/fluss/server/log/remote/LogTieringTask.java
```diff
@@ -201,6 +231,437 @@ private void runOnce() throws InterruptedException {
         }
     }
 
+    private void runOnceV2(
+            LogTablet logTablet, TableMetricGroup metricGroup, boolean allowConflictReplan)
+            throws Exception {
+        RemoteLogManifest baseManifest = remoteLog.currentManifest();
+        VersionedRemoteLogManifestHandle baseHandle = remoteLog.currentHandle();
+        if (baseHandle != null) {
+            RemoteLogManager.validateManifestHandle(baseHandle.handle(), baseManifest);
```
swuferhong: There are a lot of validations like this throughout the PR — are they all necessary? This kind of logic actually hurts code readability. Could you go through them and sort out which checks are genuinely required on this path and which can be removed?

fluss-common/src/main/java/org/apache/fluss/remote/RemoteLogManifestJsonSerde.java
```diff
@@ -43,22 +42,28 @@ public class RemoteLogManifestJsonSerde
...
     private static final String START_OFFSET_FIELD = "start_offset";
     private static final String END_OFFSET_FIELD = "end_offset";
     private static final String MAX_TIMESTAMP_FIELD = "max_timestamp";
     private static final String SEGMENT_SIZE_IN_BYTES_FIELD = "size_in_bytes";
-    private static final int SNAPSHOT_VERSION = 1;
 
     @Override
     public void serialize(RemoteLogManifest manifest, JsonGenerator generator) throws IOException {
         generator.writeStartObject();
+        generator.writeNumberField(VERSION_KEY, manifest.getVersion());
```
swuferhong: The V1 manifest serde does not validate the version and reads `remote_log_segments` as complete physical ranges. A V2 manifest may contain overlapping or start-clipped segments, so an old `TabletServer` can load it as V1 and expose ranges that are logically hidden by V2.

The same issue exists for the V1 handle reader, which ignores the handle version. Please ensure that V2 activation is persisted and that any V1 `TabletServer` is fenced from rejoining replica assignment or leadership after activation. A transient check of currently alive servers is insufficient unless re-registration also enforces this fence.

fluss-common/src/main/java/org/apache/fluss/remote/RemoteLogManifestReplacementPlanner.java
```diff
@@ -0,0 +1,241 @@
...
+        }
+        activeSegments.add(candidate);
+
+        List<UnreferencedRemoteLogSegment> unreferencedSegments =
+                new ArrayList<>(manifest.getUnreferencedRemoteLogSegments());
+        for (RemoteLogSegment segment : segmentsToUnreference) {
+            unreferencedSegments.add(
+                    new UnreferencedRemoteLogSegment(
+                            segment,
+                            unreferencedAtMs,
```
swuferhong: `unreferencedAtMs` is assigned while planning, before segment upload and manifest CAS. Upload and commit latency therefore consume part of the configured grace period, this may allow deletion earlier than one full grace period after the old manifest stopped being authoritative.

Please use a post-publication/two-phase GC transition: first publish the segment as GC-ineligible unreferenced metadata, then record its deletion eligibility after the authoritative transition has been observed.

fluss-common/src/main/java/org/apache/fluss/remote/RemoteLogManifestReplacementPlanner.java
```diff
@@ -0,0 +1,241 @@
...
+        long remoteStartOffset = manifest.getRemoteLogStartOffset();
+        long remoteEndOffset = manifest.getRemoteLogEndOffset();
+        long candidateStartOffset = candidate.remoteLogStartOffset();
+        long candidateEndOffset = candidate.remoteLogEndOffset();
+
+        if (candidateEndOffset <= remoteEndOffset) {
+            return new Result(PlanType.ALREADY_COVERED, manifest, Collections.emptyList());
+        }
+        if (candidateStartOffset > remoteEndOffset) {
+            return new Result(PlanType.GAP, manifest, Collections.emptyList());
```
swuferhong: Returning the unchanged manifest for GAP leaves `nextCopyOffset` unchanged. The next task interval selects the same local candidate and reaches the same GAP again, so remote tiering can never make forward progress after local retention has removed the missing range.

fluss-common/src/main/java/org/apache/fluss/remote/RemoteLogManifestReplacementPlanner.java
```diff
@@ -0,0 +1,241 @@
...
+                "Candidate physical table path does not match manifest");
+        checkArgument(
+                candidate.tableBucket().equals(manifest.getTableBucket()),
+                "Candidate table bucket does not match manifest");
+        ensureNewSegmentId(manifest, candidate.remoteLogSegmentId());
+
+        if (manifest.getRemoteLogSegmentList().isEmpty()) {
+            RemoteLogManifest resultManifest =
+                    RemoteLogManifest.createV2(
+                            manifest.getGeneration() + 1,
```
swuferhong: Why we need change this in plan stage? LogTieringTask's `resultManifest = withGeneration(replayed.resultManifest(), targetGeneration);` will change it again.

fluss-common/src/main/java/org/apache/fluss/remote/RemoteLogManifestReplacementPlanner.java
```diff
@@ -0,0 +1,241 @@
...
+                    new UnreferencedRemoteLogSegment(
+                            segment,
+                            unreferencedAtMs,
+                            UnreferencedRemoteLogSegment.Reason.REPLACED,
+                            candidate.remoteLogSegmentId()));
+        }
+
+        RemoteLogManifest resultManifest =
+                RemoteLogManifest.createV2(
+                        manifest.getGeneration() + 1,
```
swuferhong: ditto, maybe this no need to change.

fluss-common/src/main/java/org/apache/fluss/remote/RemoteLogManifestReplacementPlanner.java
```diff
@@ -0,0 +1,241 @@
...
+
+        public PlanType planType() {
+            return planType;
+        }
+
+        public RemoteLogManifest resultManifest() {
+            return resultManifest;
+        }
+
+        public List<RemoteLogSegment> segmentsToUnreference() {
```
swuferhong: no caller.

fluss-common/src/main/java/org/apache/fluss/remote/RemoteLogManifestReplacementPlanner.java
```diff
@@ -0,0 +1,241 @@
...
+import static org.apache.fluss.utils.Preconditions.checkArgument;
+import static org.apache.fluss.utils.Preconditions.checkNotNull;
+
+/** Builds an immutable V2 manifest update for one completely copied candidate segment. */
+@Internal
+public final class RemoteLogManifestReplacementPlanner {
+
+    /** Classification of a candidate physical segment against the current logical remote range. */
+    public enum PlanType {
+        INITIAL_COPY,
```
swuferhong: only test call this type.

fluss-common/src/main/java/org/apache/fluss/remote/RemoteLogManifestReplacementPlanner.java
```diff
@@ -0,0 +1,241 @@
...
+        ALREADY_COVERED,
+        GAP,
+        APPEND,
```
swuferhong: ditto, only test call this.

fluss-common/src/main/java/org/apache/fluss/remote/RemoteLogManifestReplacementPlanner.java
```diff
@@ -0,0 +1,241 @@
...
+        REPLACE_AND_CLIP,
```
swuferhong: ditto

fluss-common/src/main/java/org/apache/fluss/remote/RemoteLogManifestReplacementPlanner.java
```diff
@@ -0,0 +1,241 @@
...
+        return RemoteLogManifest.createV2(
+                manifest.getGeneration(),
+                manifest.getPhysicalTablePath(),
+                manifest.getTableBucket(),
+                activeSegments,
+                references.get(expireCount).logicalStartOffset(),
+                unreferencedSegments);
+    }
+
+    private static void ensureNewSegmentId(RemoteLogManifest manifest, UUID candidateId) {
```
swuferhong: This looks like another redundant check — RemoteLogManifest.createV2() already validates this. Suggest going through the PR and removing the redundant validations.

fluss-rpc/src/main/proto/FlussApi.proto
```diff
@@ -565,10 +565,21 @@ message CommitRemoteLogManifestRequest {
   required int64 remote_log_end_offset = 6;
   required int32 coordinator_epoch = 7;
   required int32 bucket_leader_epoch = 8;
+  // Manifest V2 CAS fields. All are omitted by the legacy V1 commit path.
+  optional int32 manifest_format_version = 9;
+  // 0 = ABSENT, 1 = PRESENT.
+  optional int32 expected_handle_state = 10;
+  optional string expected_manifest_path = 11;
```
swuferhong: For a ZK CAS, `expected_zk_version` is all you need — `setData().withVersion(...)` already fails on any version mismatch, and any concurrent write bumps the version. Checking path/generation on top of that is redundant: if zkVersion matches it's the same handle (so they match too); if it doesn't, the CAS fails anyway.

Keeping them also costs an extra `getVersionedRemoteLogManifestHandle` read before every CAS. Dropping them means two fewer wire fields, no pre-read, and we can delete the matching PRESENT-branch checks in `ServerRpcMessageUtils`. Maybe we can just keep `expected_zk_version` and `new_manifest_generation`.

fluss-common/src/main/java/org/apache/fluss/remote/RemoteLogManifestV2Migration.java
```diff
@@ -0,0 +1,121 @@
...
+        List<RemoteLogSegment> activeSegments = new ArrayList<>();
+        for (RemoteLogSegment segment : sameStartWinners) {
+            if (activeSegments.isEmpty()) {
+                activeSegments.add(segment);
+                continue;
+            }
+
+            RemoteLogSegment previous = activeSegments.get(activeSegments.size() - 1);
+            if (segment.remoteLogStartOffset() > previous.remoteLogEndOffset()) {
+                throw new IllegalArgumentException(
```
swuferhong: Throwing here leaves the authoritative V1 manifest unchanged, and the periodic task retries the same migration indefinitely. V1 gaps can legitimately exist after remote tiering was disabled, local retention advanced, or disaster recovery.

Please return a structured GAPPED migration result. If local log or lake coverage can prove recovery is safe, rebuild from the current candidate as an INITIAL_COPY and move the old references to unreferenced metadata.

fluss-common/src/main/java/org/apache/fluss/config/ConfigOptions.java
```diff
@@ -1091,6 +1091,33 @@ public class ConfigOptions {
                                     + "prevent overwhelming the remote storage when there is a large "
                                     + "backlog of segments to upload.");
 
+    public static final ConfigOption<Boolean> REMOTE_LOG_MANIFEST_V2_WRITER_ENABLED =
+            key("remote.log.manifest-v2-writer-enabled")
```
swuferhong: This option doesn't seem meaningful. Upgrading to the new version should just enable it by default. What we actually need is for V1 and V2 to be compatible regardless of whether the coordinator or the tablet server is upgraded first. For example: if the coordinator is upgraded first, buckets whose leader has been upgraded use V2, while those not yet upgraded keep using V1; if the tablet server is upgraded first, it can only use V1 until the coordinator upgrade completes and the metadata contains V2 info, after which the leader can commit V2.

fluss-server/src/main/java/org/apache/fluss/server/log/remote/LogTieringTask.java
```diff
@@ -387,8 +849,11 @@ public boolean tryToCommitRemoteLogManifest(
                     return false;
                 } else {
                     // commit succeed.
-                    // TODO: commit with version to avoid the manifest has been updated
-                    remoteLogTablet.addAndDeleteLogSegments(newAddedSegments, expiredSegments);
+                    if (newRemoteLogManifest.getVersion() == RemoteLogManifest.VERSION_1) {
```
swuferhong: useless if branch? this method only call by v1

fluss-server/src/main/java/org/apache/fluss/server/log/remote/LogTieringTask.java
```diff
@@ -416,7 +881,12 @@ public boolean tryToCommitRemoteLogManifest(
         return false;
     }
 
-    private boolean commitRemoteLogManifest(CommitRemoteLogManifestData data) throws Exception {
+    private boolean commitRemoteLogManifest(
+            CommitRemoteLogManifestData data, RemoteLogManifest manifest) throws Exception {
+        if (data.isV2CasCommit()) {
```
swuferhong: can v2 call this method?

fluss-common/src/main/java/org/apache/fluss/remote/RemoteLogManifestV2Migration.java
```diff
@@ -0,0 +1,183 @@
...
+            long startOffset = sortedSegments.get(index).remoteLogStartOffset();
+            while (groupEnd < sortedSegments.size()
+                    && sortedSegments.get(groupEnd).remoteLogStartOffset() == startOffset) {
+                groupEnd++;
+            }
+
+            RemoteLogSegment winner = sortedSegments.get(groupEnd - 1);
+            if (groupEnd - index > 1
+                    && sortedSegments.get(groupEnd - 2).remoteLogEndOffset()
+                            == winner.remoteLogEndOffset()) {
```
swuferhong: V1 permits multiple segment objects with the same [start, end) range, and the existing reader already resolves identical start offsets deterministically by keeping the later manifest entry. Throwing here makes dynamic V2 activation fail forever for any bucket containing such a valid V1 history.

The sort is stable, so migration can preserve the V1 manifest order as the tie-breaker: keep the last same-range entry active and move the earlier duplicates to unreferencedSegments. 

Please cover this with both a migration unit test and a LogTieringTask V1-to-V2 activation test.

fluss-server/src/main/java/org/apache/fluss/server/log/remote/LogTieringTask.java
```diff
@@ -113,95 +163,543 @@ public void run() {
...
+                createV2CommitData(baseHandle, resultManifest, manifestPath);
+        RemoteLogManifestCommitResult commitResult;
+        try {
+            commitResult =
+                    remoteLogManifestCommitter.commitAndApply(
+                            commitData, resultManifest, remoteLog);
+        } catch (Exception unknownResult) {
+            // The result is UNKNOWN or was superseded before local apply. Objects must remain
+            // untouched until a later authoritative reconciliation proves they are unreferenced.
+            throw unknownResult;
```
swuferhong: catch and throw? why need try catch.

fluss-server/src/main/java/org/apache/fluss/server/log/remote/LogTieringTask.java
```diff
@@ -113,95 +163,543 @@ public void run() {
...
+            applyPublishedOffsets(logTablet, resultManifest);
+            nextCopyOffset = successfulNextCopyOffset;
+            return;
+        }
+
+        cleanupRejectedPlan(copiedSegments, manifestPath, metricGroup, logTablet);
+        if (commitResult == RemoteLogManifestCommitResult.CONFLICT && allowConflictReplan) {
+            nextCopyOffset = findNextCopyOffset(logTablet);
+            runOnceV2(logTablet, metricGroup, false);
+        }
```
swuferhong: `INVALID_MANIFEST` needs to fall back to the V1 writer, otherwise we get an upload → reject → delete churn loop during upgrades.

Right now only `CONFLICT` gets special handling; `INVALID_MANIFEST` and `FENCED` both fall through to `cleanupRejectedPlan`, which deletes the segments we just uploaded:

But `INVALID_MANIFEST` is exactly what the coordinator returns when its own writer gate is still closed. So:

Scenario: a future release flips the default to true, and an operator upgrades the tablet servers first while the coordinator is still on the old version (gate closed).
→ TS sends a V2 commit → coordinator replies INVALID_MANIFEST → cleanupRejectedPlan deletes the freshly uploaded segments → next tiering interval repeats the whole thing. For the entire upgrade window we burn remote bandwidth plus PUT/DELETE request cost and tiering makes zero progress.

Suggestion: treat `INVALID_MANIFEST` as "the coordinator is not ready for V2 yet" rather than as a hard rejection — for this round, fall back to `runOnceV1` for the bucket, and set a process-local flag so the next round probes V2 before committing to it. That makes the config-gated rollout genuinely order-independent: neither upgrade order causes churn, and we don't have to document a required upgrade sequence.

Please also add test coverage for the `INVALID_MANIFEST` and `FENCED` paths — neither is exercised today.

fluss-server/src/main/java/org/apache/fluss/server/entity/CommitRemoteLogManifestData.java
```diff
@@ -44,19 +48,112 @@ public class CommitRemoteLogManifestData {
...
+            int coordinatorEpoch,
+            int bucketLeaderEpoch) {
+        return new CommitRemoteLogManifestData(
+                tableBucket,
+                remoteLogManifestPath,
+                remoteLogStartOffset,
+                remoteLogEndOffset,
+                coordinatorEpoch,
+                bucketLeaderEpoch,
+                2,
```
swuferhong: use `RemoteLogManifest.VERSION_2` ?

fluss-server/src/main/java/org/apache/fluss/server/entity/CommitRemoteLogManifestData.java
```diff
@@ -44,19 +48,112 @@ public class CommitRemoteLogManifestData {
...
```
swuferhong: ditto

fluss-server/src/main/java/org/apache/fluss/server/entity/CommitRemoteLogManifestData.java
```diff
@@ -44,19 +48,112 @@ public class CommitRemoteLogManifestData {
...
         this.remoteLogStartOffset = remoteLogStartOffset;
         this.remoteLogEndOffset = remoteLogEndOffset;
         this.coordinatorEpoch = coordinatorEpoch;
         this.bucketLeaderEpoch = bucketLeaderEpoch;
+        this.manifestFormatVersion = manifestFormatVersion;
+        this.expectedZkVersion = expectedZkVersion;
+        this.newManifestGeneration = newManifestGeneration;
+    }
+
+    public static CommitRemoteLogManifestData v2Absent(
```
swuferhong: recommand rename to `v2CreateIfAbsent()`

fluss-server/src/main/java/org/apache/fluss/server/entity/CommitRemoteLogManifestData.java
```diff
@@ -44,19 +48,112 @@ public class CommitRemoteLogManifestData {
...
+                remoteLogStartOffset,
+                remoteLogEndOffset,
+                coordinatorEpoch,
+                bucketLeaderEpoch,
+                2,
+                null,
+                newManifestGeneration);
+    }
+
+    public static CommitRemoteLogManifestData v2Present(
```
swuferhong: recommad rename to `v2CompareAndSet()`

fluss-server/src/main/java/org/apache/fluss/server/coordinator/CoordinatorEventProcessor.java
```diff
@@ -213,6 +217,40 @@ public CoordinatorEventProcessor(
...
+                conf,
+                ioExecutor,
+                metadataManager,
+                kvSnapshotLeaseManager,
+                scheduler,
+                clock,
+                new RemoteManifestV2WriterGate(conf));
+    }
+
+    public CoordinatorEventProcessor(
```
swuferhong, line 201: Could we avoid adding a new constructor here?

fluss-server/src/test/java/org/apache/fluss/server/coordinator/CoordinatorEventProcessorTest.java
```diff
@@ -1145,6 +1149,10 @@ void testStartupResumesDropPartitionThroughCleanupManager() throws Exception {
 
     @Test
     void testNotifyOffsetsWithShrinkISR(@TempDir Path tempDir) throws Exception {
+        eventProcessor.shutdown();
+        manifestV2WriterEnabledForTest = true;
+        eventProcessor = buildCoordinatorEventProcessor();
+        eventProcessor.startup();
```
swuferhong: Why was this seemingly unrelated test modified?

fluss-server/src/main/java/org/apache/fluss/server/log/remote/RemoteLogManager.java
```diff
@@ -95,6 +100,27 @@ public RemoteLogManager(
...
+                zkClient,
+                coordinatorGateway,
+                localDiskManager,
+                logManager,
+                clock,
+                ioExecutor,
+                () -> conf.getBoolean(ConfigOptions.REMOTE_LOG_MANIFEST_V2_WRITER_ENABLED));
+    }
+
+    public RemoteLogManager(
```
swuferhong, line 89: public LogTieringTask(Could we avoid adding a new constructor here?

fluss-server/src/main/java/org/apache/fluss/server/log/remote/RemoteLogManager.java
```diff
@@ -119,6 +146,30 @@ public RemoteLogManager(
...
+                localDiskManager,
+                logManager,
+                remoteLogStorage,
+                scheduledExecutor,
+                clock,
+                () -> conf.getBoolean(ConfigOptions.REMOTE_LOG_MANIFEST_V2_WRITER_ENABLED));
+    }
+
+    @VisibleForTesting
+    public RemoteLogManager(
```
swuferhong, line 112: ditto. Could we avoid adding a new constructor here?

fluss-server/src/main/java/org/apache/fluss/server/log/remote/LogTieringTask.java
```diff
@@ -75,17 +93,49 @@ public LogTieringTask(
...
+                remoteLogIndexCache,
+                coordinatorGateway,
+                zooKeeperClient,
+                clock,
+                maxUploadSegmentsPerTask,
+                () -> manifestV2WriterEnabled,
+                manifestV2GcGracePeriodMs);
+    }
+
+    public LogTieringTask(
```
swuferhong, line 73: Could we avoid adding a new constructor here?

fluss-server/src/main/java/org/apache/fluss/server/log/remote/LogTieringTask.java
```diff
@@ -113,95 +163,543 @@ public void run() {
...
+                }
+            }
+        }
+    }
+
+    private void runOnceV2(
+            LogTablet logTablet, TableMetricGroup metricGroup, boolean allowConflictReplan)
+            throws Exception {
+        RemoteLogManifest baseManifest = remoteLog.currentManifest();
+        VersionedRemoteLogManifestHandle baseHandle = remoteLog.currentHandle();
```
swuferhong: `baseManifest` and `baseHandle` are read through two separate lock acquisitions. If an old task overlaps with a newly started task, remoteLog may be updated between these reads, producing an inconsistent pair—for example, an old manifest with a newer ZK handle/version. The subsequent CAS could then succeed using the newer ZK version while publishing a manifest derived from stale state, potentially overwriting a more recent update.

Please expose an atomic snapshot from `RemoteLogTablet` that contains both the manifest and its versioned handle (like `ManifestSnapshot`), captured under the same read lock, and use that snapshot as the base for planning and CAS.

fluss-server/src/main/java/org/apache/fluss/server/log/remote/LogTieringTask.java
```diff
@@ -113,95 +163,543 @@ public void run() {
...
+                        .orElse(null);
+        LogSegmentFiles logSegmentFiles =
+                new LogSegmentFiles(
+                        segment.getFileLogRecords().file().toPath(),
+                        toPathIfExists(segment.offsetIndex().file()),
+                        toPathIfExists(segment.timeIndex().file()),
+                        writerIdSnapshotFile == null ? null : writerIdSnapshotFile.toPath());
+        try {
+            remoteLogStorage.copyLogSegmentFiles(remoteLogSegment, logSegmentFiles);
+        } catch (RemoteStorageException e) {
```
swuferhong: Interruption may not stop the task from publishing a manifest. 

`DefaultRemoteLogStorage` wraps `InterruptedException` in `RemoteStorageException`, while `copyPlannedSegment()` treats it as an ordinary copy failure and returns false. 

The task may then continue to write and commit a manifest containing an already-copied prefix, migration, expiration, or GC changes. In addition, `RemoteLogManifestCommitter.commit()` catches `InterruptedException` through its broad Exception handler and continues reconciliation/retries.

fluss-server/src/main/java/org/apache/fluss/server/log/remote/LogTieringTask.java
```diff
@@ -113,95 +163,543 @@ public void run() {
...
+    }
+
+    private void runOnceV2(
+            LogTablet logTablet, TableMetricGroup metricGroup, boolean allowConflictReplan)
+            throws Exception {
+        RemoteLogManifest baseManifest = remoteLog.currentManifest();
+        VersionedRemoteLogManifestHandle baseHandle = remoteLog.currentHandle();
+
+        long now = clock.milliseconds();
+        sweepOrphanObjects(baseManifest, baseHandle, now, metricGroup);
```
swuferhong: `orphan sweep` currently runs on every per-bucket tiering execution. With the default one-minute task interval, each bucket performs full listings of segment objects and manifest snapshots, potentially followed by one ZK read per deletion candidate and an unbounded number of deletions. 

At scale, this can generate substantial load on `ZooKeeper`, occupy the `remote-log scheduler threads`, and delay normal tiering and TTL processing.

Please move orphan collection out of the per-minute tiering hot path and run it as a lower-frequency, independently scheduled GC process. The sweep should also be bounded, for example by limiting the number of objects scanned/deleted per run and supporting incremental progress with a cursor. ZK handle validation should preferably be performed per batch rather than once per object.

swuferhong (commented):
Thanks @gyang94 for such a big feature. I left some comments.

fluss-common/src/main/java/org/apache/fluss/config/ConfigOptions.java
```diff
@@ -1106,6 +1106,25 @@ public class ConfigOptions {
...
+                    .booleanType()
+                    .defaultValue(false)
+                    .withDescription(
+                            "Dynamically and irreversibly enables CAS publishing of Manifest V2 "
+                                    + "snapshots and overlap-aware remote segment replacement. "
+                                    + "Keep disabled during rolling upgrades and enable only after "
+                                    + "all servers support manifest version dispatch.");
+
+    public static final ConfigOption<Duration> REMOTE_LOG_MANIFEST_V2_GC_GRACE_PERIOD =
+            key("remote.log.manifest-v2-gc-grace-period")
```
wuchong: Could we rename this option to `remote.log.delete-delay` and the constant to `REMOTE_LOG_DELETE_DELAY`? `GC` is implementation jargon that many operators may not recognize, while `manifest-v2` exposes an internal format version in a public configuration key and will age poorly as the format evolves. This option delays physical deletion of unreferenced segments, obsolete manifest snapshots, and orphan objects, so `delete-delay` states the user-visible behavior more directly and remains version-neutral.

fluss-server/src/main/java/org/apache/fluss/server/log/LogTablet.java
```diff
@@ -627,8 +630,13 @@ public void updateRemoteLogSize(long remoteLogSize) {
     public void updateRemoteLogEndOffset(long remoteLogEndOffset) {
         if (remoteLogEndOffset > this.remoteLogEndOffset) {
```
wuchong: We should allow `remoteLogEndOffset` to be updated back to `-1`, the same to `remoteLogStartOffset`. Otherwise, there is no difference between `remoteLogEndOffset` and `highestCopiedEndOffset` here.

fluss-server/src/main/java/org/apache/fluss/server/log/remote/LogTieringTask.java
```diff
@@ -113,95 +141,468 @@ public void run() {
...
+        List<UnreferencedRemoteLogSegment> retained = new ArrayList<>();
+        List<UUID> deletedSegmentIds = new ArrayList<>();
+        for (UnreferencedRemoteLogSegment unreferenced :
+                manifest.getUnreferencedRemoteLogSegments()) {
+            if (!gracePeriodElapsed(unreferenced.unreferencedAtMs(), now)) {
+                retained.add(unreferenced);
+                continue;
+            }
+            try {
+                remoteLogStorage.deleteLogSegmentFiles(unreferenced.remoteLogSegment());
```
wuchong: We shouldn't delete remote log segment before committing manifest, because commit may fail. In V1, we only delete expired segements after committing.

fluss-server/src/main/java/org/apache/fluss/server/log/remote/LogTieringTask.java
```diff
@@ -113,95 +141,468 @@ public void run() {
...
+                metricGroup.remoteLogDeleteRequests().inc();
+                deletedSegmentIds.add(unreferenced.remoteLogSegment().remoteLogSegmentId());
+            } catch (Exception e) {
+                retained.add(unreferenced);
+                metricGroup.remoteLogDeleteErrors().inc();
+                metricGroup.remoteGcFailures().inc();
```
wuchong: We don't need to introduce another metric for failed delete remote log segment. This should already covered by `metricGroup.remoteLogDeleteErrors()`.

fluss-server/src/main/java/org/apache/fluss/server/log/remote/RemoteLogManager.java
```diff
@@ -258,13 +276,22 @@ public long lookupOffsetForTimestamp(TableBucket tableBucket, long timestamp) {
...
-        RemoteLogSegment segment = remoteLogTablet.findSegmentByTimestamp(timestamp);
-        if (segment == null) {
-            return -1L;
-        } else {
-            return remoteLogIndexCacheForBucket(tableBucket)
-                    .lookupOffsetForTimestamp(segment, timestamp);
+        for (RemoteLogSegmentReference reference :
+                remoteLogTablet.allRemoteLogSegmentReferences()) {
+            RemoteLogSegment segment = reference.remoteLogSegment();
+            // A clipped segment's physical max timestamp is only a conservative candidate: it
```
wuchong: For large tables with tens of thousands of segments per bucket, this approach may lead to performance regression. I suggest using `remoteLogTablet.findSegmentByTimestamp` to optimize this method (note that `findSegmentByTimestamp` is not currently used in the PR). However, we must first correct its implementation to ensure semantic accuracy, as the presence of a logical end offset means the maximum timestamp may no longer be reliable. Please refer to my comments for further details.


**Updates to `remoteLogTablet.findSegmentByTimestamp`:** 
Considering we already have logical start and logical end offset in  `RemoteLogSegment`. We should retrieve the segment with the smallest logical start offset. If the logical end offset differs from the physical end offset (we can have a `isEndOffsetClipped` method on `RemoteLogSegment`), it indicates that the segment has been truncated, and the max timestamp may be inaccurate. In this case, also fetch the segment with the next timestamp and return both to the caller, allowing the caller to determine the correct segment. This should be implemented as a loop that continues until a non-truncated segment is found. Consequently, the return type should be changed to a List, sorted by logical start offset.

fluss-server/src/main/java/org/apache/fluss/server/zk/data/RemoteLogManifestHandle.java
```diff
@@ -19,18 +19,112 @@
...
 public class RemoteLogManifestHandle {
+    public static final int VERSION_1 = 1;
+    public static final int VERSION_2 = 2;
+
+    private final int version;
     private final FsPath remoteLogManifestPath;
+    private final @Nullable Long manifestGeneration;
+    private final @Nullable Long remoteLogStartOffset;
     private final long remoteLogEndOffset;
+    private final long highestCopiedEndOffset;
```
wuchong: I think we don't need to update `RemoteLogManifestHandle`. All the members are not necessary, they are only used to validate the equality with the content in manifest file. We can simplify just trust the content of the manifest file, and use `highestCopiedEndOffset` in there.

fluss-common/src/main/java/org/apache/fluss/remote/UnreferencedRemoteLogSegment.java
```diff
@@ -0,0 +1,117 @@
...
+    /** The reason why a physical remote segment stopped participating in logical reads. */
+    public enum Reason {
+        REPLACED,
+        EXPIRED
+    }
+
+    private final RemoteLogSegment remoteLogSegment;
+    private final long unreferencedAtMs;
+    private final Reason reason;
+    private final @Nullable UUID replacementSegmentId;
```
wuchong: This is only used for REPLACED segments, and is null for EXPIRED segments. It is not used in production code, we can remove it for now, and introduce when needed.

fluss-common/src/main/java/org/apache/fluss/remote/RemoteLogManifest.java
```diff
@@ -20,43 +20,273 @@
...
+ * snapshot of a remote log tablet.
+ *
+ * <p>Every segment uses a half-open physical offset range {@code [startOffset, endOffset)}.
  */
-public class RemoteLogManifest {
+public final class RemoteLogManifest {
+    public static final int VERSION_1 = 1;
+    public static final int VERSION_2 = 2;
+
+    private final int version;
```
wuchong: we don't need `version` if we still use V1, and all other new fields are optional.

fluss-common/src/main/java/org/apache/fluss/remote/RemoteLogManifest.java
```diff
@@ -20,43 +20,273 @@
...
+    private final long generation;
```
wuchong: we don't need `generation` as this is not the case we need to resolve in this PR.

fluss-common/src/main/java/org/apache/fluss/remote/RemoteLogManifest.java
```diff
@@ -20,43 +20,273 @@
...
     private final PhysicalTablePath physicalTablePath;
     private final TableBucket tableBucket;
     private final List<RemoteLogSegment> remoteLogSegmentList;
+    private final @Nullable Long persistedRemoteLogStartOffset;
+    private final long highestCopiedEndOffset;
+    private final List<UnreferencedRemoteLogSegment> unreferencedRemoteLogSegments;
+    private final List<RemoteLogSegmentReference> activeReferences;
```
wuchong: we don't need `activeReferences` if `RemoteLogSegment` itself contains logical start/end offset.

fluss-common/src/main/java/org/apache/fluss/remote/RemoteLogManifest.java
```diff
@@ -20,43 +20,273 @@
...
+    private final long logicalRemoteLogStartOffset;
+    private final long logicalRemoteLogEndOffset;
```
wuchong: We don't need (pre-compute) `logicalRemoteLogStartOffset`, and `logicalRemoteLogEndOffset` as they can be calculated by the logical start/end offset from first/last segemnt with `O(1)` complexity.

fluss-common/src/main/java/org/apache/fluss/remote/RemoteLogManifest.java
```diff
@@ -20,43 +20,273 @@
...
+public final class RemoteLogManifest {
+    public static final int VERSION_1 = 1;
+    public static final int VERSION_2 = 2;
+
+    private final int version;
+    private final long generation;
     private final PhysicalTablePath physicalTablePath;
     private final TableBucket tableBucket;
     private final List<RemoteLogSegment> remoteLogSegmentList;
+    private final @Nullable Long persistedRemoteLogStartOffset;
```
wuchong: I think we also don't need this. As we have the logical start offset in `remoteLogSegmentList`.

wuchong (commented):
Hi @gyang94 , I have a high-level concern about the current design. The implementation has become very complex and introduces a substantial amount of code that does not appear to be strictly necessary for the core purpose of this PR. In particular, it adds parallel Manifest V1/V2 paths, migration logic, a dedicated writer gate, a separate logical-reference model, and a replacement planner. Some of these paths duplicate existing manifest update and read logic.

I strongly suggest simplifying the design before merging this PR. Otherwise, future changes to offset semantics, fetching, expiration, GC, and compatibility will need to be kept consistent across multiple representations and code paths, resulting in a high long-term maintenance cost. Changes that are useful but unrelated to supporting overlapping segments should preferably be moved to separate follow-up PRs.

More specifically, I suggest the following:

1. **Do not introduce `remote.log.manifest-v2-writer-enabled`.**

   For rolling-upgrade compatibility, I think we should expose a switch for rolling the active segment instead and keep it disabled by default. During the rolling upgrade, active-segment rolling remains disabled, so no overlapping segments can be produced.

   We should also avoid bumping the manifest version. The existing manifest format can be extended with optional fields such as `highestCopiedEndOffset`. While active-segment rolling is disabled, `highestCopiedEndOffset` is always equal to `remoteEndOffset`, so old servers can continue reading manifests written by new servers while safely ignoring the additional field. Once every server has been upgraded, active-segment rolling can be enabled.

   This approach should allow us to remove a large amount of Manifest V2-specific code, including the writer gate, the V1/V2 dispatch paths, and `RemoteLogManifestV2Migration`.

2. **Make `RemoteLogSegment` self-contained by storing its logical range directly.**

   I suggest adding `logicalStartOffset` and `logicalEndOffset` to `RemoteLogSegment` itself. Separating logical bounds into `RemoteLogSegmentReference` means that any caller holding only a `RemoteLogSegment` can accidentally use the physical offsets, which is especially error-prone when a segment has been logically clipped because of an overlap.

   Keeping both the physical and logical ranges in the segment object would make its semantics explicit at every call site. It should also allow us to remove `RemoteLogSegmentReference` and the additional V2-specific read paths introduced around it, such as `RemoteLogTablet#relevantRemoteLogSegmentReferences`.

3. **Reuse `RemoteLogManifest#trimAndMerge` instead of introducing `RemoteLogManifestReplacementPlanner`.**

   I do not think we need a separate planner and another manifest-update state machine. We should extend `RemoteLogManifest#trimAndMerge(deletedSegments, addedSegments)` so that it can trim and merge overlapping segments correctly.

   Rolling the active segment can introduce overlaps, but it should not introduce gaps or skipped segments. Therefore, the additional gap/skip planning states do not seem necessary for the problem this PR is trying to solve. Handling overlap directly in `trimAndMerge` would keep manifest mutation in one place and avoid maintaining duplicate update rules.

<br>

The core design principle should be to avoid introducing a separate V2 path and instead adapt the existing path in place. This PR should focus exclusively on supporting overlapping segments. It should deliberately leave other pre-existing problems—such as log gaps, skipped logs, manifest generation, ZooKeeper CAS, and `VersionedRemoteLogManifestHandle`—out of scope. Those concerns can be addressed independently in follow-up PRs.

wuchong (commented):
Thanks @gyang94 for the updating. I think the overall code is very good now. I have some minor improvements and will open a follow-up PR for this. I will merge this PR for now.

wuchong (approved):
LGTM.

PR #3823, opened 2026-08-02
[docs] Add RoaringBitmap SQL functions documentation

pbanakar:
### Purpose

Linked issue: Part of #3289

Adds a dedicated SQL Functions reference page for the 12 RoaringBitmap SQL functions introduced by FIP-37 (PRs #3319, #3398, #3492, #3777). These are Flink-side functions registered by FlussCatalog and are distinct from the storage-level rbm32/rbm64 aggregators.
Documenting them on the Aggregation Merge Engine page blurred the ownership boundary, so this PR moves the content to a dedicated page.

### Brief change log

Added `website/docs/engine-flink/sql-functions.md`:
- New page titled "SQL Functions" placed after Delta Joins in the Flink engine sidebar
- Opens with a "How to Use" section explaining catalog setup and how to enable the functions
- Documents all 12 functions (4 aggregate, 8 scalar) with signature, null semantics table,
  and a self-contained deterministic VALUES-based example showing expected output
- Notes where functions have no server-side counterpart (rb_and_agg, rb_xor_agg)

Modified `website/docs/table-design/merge-engines/aggregation.md`:
- Removed the RoaringBitmap SQL Functions section
- Added a short cross-reference tip block next to rbm32/rbm64 pointing to the new page

### Tests

Documentation-only change. Verified locally by running the website dev server.
SQL examples verified against implemented functions in PRs #3319, #3398, #3492, and #3777.

### API and Format

No API or storage format changes. Documentation only.

### Documentation

Yes — this PR adds the user-facing function reference for FIP-37, completing the documentation commitment made in PRs #3319, #3398, #3492, and #3777.


pbanakar:
@polyzos @platinumhamburg Could you please take a look at this PR?

website/docs/table-design/merge-engines/aggregation.md
```diff
@@ -970,6 +970,272 @@ TableDescriptor.builder()
 </TabItem>
 </Tabs>
 
+## RoaringBitmap SQL Functions
```
wuchong: These are Flink-side SQL functions registered by the Fluss catalog, not storage-level aggregate functions executed by the Fluss merge engine, so documenting them in the Aggregation Merge Engine page blurs the ownership and execution boundary. Could we move this section to a dedicated `website/docs/engine-flink/sql-functions.md` page titled **SQL Functions**, place it after **Delta Joins** in the Engine Flink sidebar, and leave only a short cross-link next to `rbm32`/`rbm64` here?

For each function, please make the example self-contained and show the expected output. Prefer deterministic `VALUES` inputs so readers can immediately connect the inputs, output, and documented behavior. Also, please rename **End-to-End Example** to **How to Use** and move to the beginning of the page, and use that section to explain how to enable and use these Fluss-provided Flink functions, followed by the broader usage example.

website/docs/table-design/merge-engines/aggregation.md
```diff
@@ -970,6 +970,272 @@ TableDescriptor.builder()
...
+FROM uv_agg
+WHERE ymd = '20260101';
+```
+
+#### rb_build
+
+Constructs a serialized `RoaringBitmap` from an array of integers within a single row.
+
+- **Signature**: `rb_build(values ARRAY<INT>) -> BYTES`
+- **Null Handling**: Null elements in the array are ignored; returns `NULL` if all elements are null
```
wuchong: `rb_build` does not return `NULL` when all array elements are null. The implementation returns a serialized empty bitmap for any non-null array, and the [existing test](https://github.com/apache/fluss/blob/590845ea165863d0d59d7c9fe1ba4f44727b132c/fluss-flink/fluss-flink-common/src/test/java/org/apache/fluss/flink/functions/bitmap/RbScalarFunctionsTest.java#L90-L96) explicitly verifies that an all-null array produces a non-null empty bitmap. Therefore, `rb_build(ARRAY[CAST(NULL AS INT)]) IS NULL` is actually false, and `rb_cardinality(...)` returns 0. Please update the documentation to state that only a null array returns `NULL`; empty arrays and all-null arrays return an empty bitmap.

wuchong (commented):
Thanks @pbanakar , I left some comments.

website/docs/table-design/merge-engines/aggregation.md
pbanakar: Fixed rb_build null semantics: a NULL array returns NULL; an empty array or all-null array returns an empty bitmap

website/docs/table-design/merge-engines/aggregation.md
pbanakar: - Moved the entire section to a new dedicated page website/docs/engine-flink/sql-functions.md, placed after Delta Joins in the Flink sidebar
- Added a short cross-link in aggregation.md next to rbm32/rbm64 pointing to the new page
- Rewrote "End-to-End Example" as "How to Use" and moved it to the top of the page to explain catalog setup first
- All examples now use deterministic VALUES inputs and show explicit expected outputs

pbanakar:
Thanks @wuchong for review, this looks more structured and well organized than before, PTAL another look when you have some time thanks!

website/docs/engine-flink/sql-functions.md
```diff
@@ -0,0 +1,333 @@
+---
+id: sql-functions
+title: SQL Functions
+sidebar_label: SQL Functions
+sidebar_position: 7
```
wuchong: Both SQL Functions and Delta Joins currently use `sidebar_position: 7`, so this page is not explicitly placed after Delta Joins as stated in the PR description. Please set this value to `8`.

website/docs/engine-flink/sql-functions.md
```diff
@@ -0,0 +1,333 @@
...
+SELECT rb_cardinality(rb_build(ARRAY[1, 2, 3, 2]));
+-- Output: 3
+```
+
+All functions below operate on `BYTES` columns containing standard 32-bit RoaringBitmap
+serialized data — the same wire format used by the `rbm32` storage-level aggregator.
+
+---
+
+## Aggregate Functions
```
wuchong: Could we place Scalar Functions before Aggregate Functions? `rb_build`, `rb_cardinality`, and the other scalar functions are the foundational operations for creating, inspecting, and manipulating a bitmap, and the introductory example already uses them. Presenting single-row bitmap operations before cross-row aggregation avoids forward references and gives the page a more natural progression. Within the scalar section, please introduce `rb_build` before `rb_cardinality`, since `rb_build` is the primitive used throughout the examples.

website/docs/engine-flink/sql-functions.md
```diff
@@ -0,0 +1,333 @@
...
+
+Aggregate functions reduce multiple rows into a single bitmap result.
+
+### rb_build_agg
+
+Builds a serialized `RoaringBitmap` from a column of `INT` values across rows.
+
+**Signature:** `rb_build_agg(value INT) → BYTES`
+
+| Input | Output |
```
wuchong: The null-handling tables take up a large portion of the page and feel repetitive, while the same rules would be easier to scan as short prose. Please use the earlier compact format and add a `**Null Handling:** ...` item alongside `**Signature:**` for each function instead of using a table.

website/docs/table-design/merge-engines/aggregation.md
```diff
@@ -970,6 +970,12 @@ TableDescriptor.builder()
 </TabItem>
 </Tabs>
 
+:::tip
+To query and manipulate bitmap columns from Flink SQL, use the built-in RoaringBitmap SQL
+functions (`rb_build_agg`, `rb_or_agg`, `rb_cardinality`, etc.) provided by FlussCatalog.
```
wuchong, line 973: Could we also update the `rbm32` Flink SQL example above to construct bitmap values with `rb_build(ARRAY[...])` instead of opaque `X3A30...` literals? This makes the example easier to understand and lets users modify and rerun it without manually producing serialized bytes. Likewise, please select `rb_to_array(visit_bitmap)` in the result query so the union is shown in a human-readable form, and remove or update the note saying that values must be pre-serialized on the client side.

website/docs/engine-flink/sql-functions.md
```diff
@@ -0,0 +1,333 @@
...
+|---|---|
+| NULL rows | ignored |
+| all NULL | `NULL` |
+
+```sql
+-- Input: two bitmaps {1,2} and {2,3}
+SELECT rb_cardinality(rb_or_agg(bmap)) AS result
+FROM (
+    VALUES
+        (X'3A30000000000000020000000102'),  -- bitmap {1, 2}
```
wuchong: Please build these inputs with `rb_build` rather than hard-coded serialized bytes, for example by placing arrays in `VALUES` and applying `rb_build` before `rb_or_agg`. This keeps the example self-explanatory and avoids opaque binary literals whose contents readers cannot verify or safely modify.

pbanakar:
Hi @wuchong addressed all comments, ready for another look, Thanks 🙏

wuchong (approved):
Thanks @pbanakar , I added 2 commits to improve the SQL examples. LGTM.

PR #3825, opened 2026-08-03
[lake/paimon] Make PaimonSplit state survive Paimon class relocation

### LINKED ISSUE #3824: [lake] Restoring PaimonSplit state fails with ClassNotFoundException when Paimon classes are relocated
### Search before asking

### Fluss version

0.9.0 (latest release)

### Please describe the bug 🐞

`PaimonSplitSerializer` persists `DataSplit` into Flink checkpoint/savepoint state via Java serialization, which embeds the class name `org.apache.paimon.table.source.DataSplit` into the state bytes.

Distributions that shade the Fluss connector and relocate Paimon classes (e.g. to `x.y.shaded.org.apache.paimon.*`) cannot restore state written by a non-relocated build: `java.io.IOException: Failed to deserialize PaimonSplit Caused by: java.lang.ClassNotFoundException: org.apache.paimon.table.source.DataSplit`

### Steps to reproduce

1. Run a union read job (Flink source reading a datalake-enabled table) with a non-relocated Fluss connector, stop with a savepoint.
2. Restore the job from that savepoint with a connector build that relocates Paimon classes.
3. Restore fails with the exception above.

### Root cause

Java serialization writes the fully-qualified class name into the bytes, so the state format breaks whenever Paimon classes are renamed by shading. Both the source operator state and the enumerator checkpoint are affected, since they share the same serializer.

Note: Iceberg and Lance split serializers already use hand-written binary and are not affected; Paimon is the only exception.

### Affected versions

All versions where `PaimonSplitSerializer` uses Java serialization (V1/V2 format).

### Solution

_No response_

beryllw:
### Purpose

Linked issue: close #3824

### DIFF
--- a/fluss-lake/fluss-lake-paimon/src/main/java/org/apache/fluss/lake/paimon/source/PaimonSplitSerializer.java
+++ b/fluss-lake/fluss-lake-paimon/src/main/java/org/apache/fluss/lake/paimon/source/PaimonSplitSerializer.java
@@ -24,6 +24,7 @@ import org.apache.fluss.lake.serializer.SimpleVersionedSerializer;
 import org.apache.paimon.data.BinaryRow;
 import org.apache.paimon.io.DataInputViewStreamWrapper;
 import org.apache.paimon.io.DataOutputViewStreamWrapper;
+import org.apache.paimon.table.FallbackReadFileStoreTable;
 import org.apache.paimon.table.source.DataSplit;
 import org.apache.paimon.utils.InstantiationUtil;
 import org.slf4j.Logger;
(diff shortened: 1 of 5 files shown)


fluss-lake/fluss-lake-paimon/src/main/java/org/apache/fluss/lake/paimon/source/PaimonSplitSerializer.java
```diff
@@ -62,26 +76,52 @@ public byte[] serialize(PaimonSplit paimonSplit) throws IOException {
...
+                return deserializeV3(serialized);
+            default:
+                throw new IOException("Unsupported PaimonSplit serialization version: " + version);
+        }
+    }
+
+    private PaimonSplit deserializeV3(byte[] serialized) throws IOException {
+        ByteArrayInputStream in = new ByteArrayInputStream(serialized);
+        DataInputViewStreamWrapper view = new DataInputViewStreamWrapper(in);
+        DataSplit dataSplit = DataSplit.deserialize(view);
```
luoyuxia: [Major] `serialize()` dispatches virtually, but VERSION_3 always deserializes with the static base `DataSplit.deserialize()`. This breaks legitimate `DataSplit` subclasses. For example, in Paimon 1.3.1, [`FallbackDataSplit`](https://github.com/apache/paimon/blob/release-1.3.1/paimon-core/src/main/java/org/apache/paimon/table/FallbackReadFileStoreTable.java#L246-L289) appends an `isFallback` flag after the base payload and is emitted when a fallback branch is scanned. This line consumes only the base payload, so the next read treats `isFallback` as `isBucketUnAware`, the partition-size read becomes byte-shifted, and the restored split also loses its subtype. This can corrupt the restored fields, fail deserialization, or route a fallback split to the main branch.

luoyuxia (approved):

PR #3831, opened 2026-08-03
[server] Roll expired active segments for remote retention

### LINKED ISSUE #3830: [server] Active log segment is not rolled after table.log.ttl expires
### Search before asking

### Fluss version

main (development)

### Please describe the bug 🐞

When remote log is enabled and no new records are appended after the active log segment exceeds `table.log.ttl`, the periodic local retention cleanup does not roll the active segment.

#### Steps to reproduce

1. Enable remote log with `remote.log.task-interval-duration > 0`.
2. Configure a short `table.log.ttl`.
3. Append records to create a non-empty active segment.
4. Stop writing and wait until the active segment exceeds the configured TTL.
5. Trigger the periodic local log retention cleanup.

#### Actual behavior

The active segment remains active and is not uploaded to remote storage. It is therefore not eligible for the subsequent local segment cleanup.

The current implementation of `LogTablet.deletableExpiredSegments()` only scans inactive segments and excludes the active segment.

#### Expected behavior

The cleanup should roll a non-empty expired active segment when:

- remote log is enabled;
- the active segment is expired by `table.log.ttl`;
- `highWatermark >= logEndOffset`.

The rolled segment should not be deleted in the same cleanup pass. After it is uploaded to remote storage, a later cleanup pass should remove it according to the existing remote boundary and retention rules.

This behavior is similar to [KAFKA-15047](https://issues.apache.org/jira/browse/KAFKA-15047).

### Solution

- Extend the existing contiguous TTL scan to inspect the active segment.
- Roll the active segment only after preceding inactive segments satisfy the existing TTL and offset boundaries.
- Keep the active segment out of the current deletion list.
- Preserve the existing `remoteLogEndOffset`, `minRetainOffset`, high-watermark, and empty-active-segment behavior.
- Add regression coverage in `TieredLocalSegmentTTLTest` for:
  - active segment roll after TTL expiration;
  - high watermark below LEO;
  - remote end offset boundaries;
  - remote log disabled;
  - empty active segment.

This change does not modify any public API, RPC protocol, or storage format.

Potential leader/follower segment boundary differences during timestamp roll are out of scope and will be handled separately.

fxbing:
### Purpose

Linked issue: close #3830

Local TTL cleanup currently scans only inactive log segments. When a table becomes idle after its active segment exceeds `table.log.ttl`, that segment remains active and cannot be uploaded to remote storage or removed by subsequent local retention.

This change rolls a non-empty expired active segment during periodic local retention once the high watermark reaches the log end offset. The rolled segment is not deleted in the same cleanup pass; it can first be uploaded and then removed by a later cleanup.

Leader/follower segment boundary alignment during timestamp-based roll is intentionally out of scope.

### Brief change log

- Cache whether remote log is enabled from `remote.log.task-interval-duration`.
- Extend the contiguous TTL scan to inspect the active segment while preserving the existing `remoteLogEndOffset` and `minRetainOffset` boundaries for inactive segments.
- Roll the active segment only when remote log is enabled, the segment is non-empty and expired, and `highWatermark >= logEndOffset`.
- Do not require `remoteLogEndOffset` to cover the active segment before rolling, because rolling is required before the segment can be uploaded.
- Use the existing `LogTablet.roll(Optional.empty())` path and narrow its exception declaration to `IOException`.
- Extend `TieredLocalSegmentTTLTest` for partitioned and non-partitioned tables, covering:
  - upload and cleanup lifecycle;
  - high-watermark gating;
  - remote offset boundaries;
  - disabled remote log;
  - disabled table TTL;
  - empty active segments.

### Tests

- `./mvnw -pl fluss-server -DskipITs -Dcheckstyle.skip -Drat.skip -Dspotless.check.skip=true -Dtest=TieredLocalSegmentTTLTest test`
  - 10 tests passed.
- `./mvnw -pl fluss-server spotless:check`

### API and Format

No public API, RPC protocol, configuration, or storage format changes.

### Documentation

### Generative AI disclosure

Yes. OpenAI Codex was used to assist with implementation and review.

### DIFF
--- a/fluss-server/src/main/java/org/apache/fluss/server/log/LogTablet.java
+++ b/fluss-server/src/main/java/org/apache/fluss/server/log/LogTablet.java
@@ -109,7 +109,6 @@ public final class LogTablet {
     private final Clock clock;
     private final boolean isChangeLog;
     private final long logTtlMs;
-    private final boolean remoteLogEnabled;
 
     @GuardedBy("lock")
     private volatile LogOffsetMetadata highWatermarkMetadata;
@@ -161,8 +160,6 @@ public final class LogTablet {
         this.writerStateManager = writerStateManager;
         this.highWatermarkMetadata = new LogOffsetMetadata(0L);
         this.logTtlMs = logTtlMs;
-        this.remoteLogEnabled =
-                conf.get(ConfigOptions.REMOTE_LOG_TASK_INTERVAL_DURATION).toMillis() > 0L;
 
         this.scheduler = scheduler;
         // scheduler the writer expiration interval check.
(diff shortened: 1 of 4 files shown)


fluss-server/src/main/java/org/apache/fluss/server/log/LogTablet.java
```diff
@@ -1359,13 +1362,28 @@ private List<LogSegment> deletableExpiredSegments(long endOffset) throws IOExcep
...
+            boolean active = i == logSegments.size() - 1;
+            if ((!active && logSegments.get(i + 1).getBaseOffset() > endOffset)
+                    || !isSegmentExpired(now, segment, logTtlMs)) {
                 break;
             }
-            deletableSegments.add(logSegments.get(i));
+
+            if (active) {
+                shouldRoll =
+                        remoteLogEnabled
```
gyang94: Should rolling active segments depend on `remoteLogEnabled`?  Rolling active local segments is independent with remote copy.

fluss-server/src/main/java/org/apache/fluss/server/log/LogTablet.java
```diff
@@ -1350,7 +1353,7 @@ private List<LogSegment> deletableRemoteSegments(long endOffset) {
         return deletableSegments;
     }
 
-    /** Returns the contiguous prefix of inactive segments that has expired. */
+    /** Returns the contiguous prefix of expired segments and rolls an expired active segment. */
     private List<LogSegment> deletableExpiredSegments(long endOffset) throws IOException {
```
gyang94, line 1362: Now in this `deletableExpiredSegments` function, we have new 'rolling' operations. The responsibility is not clear.
Maybe we can add a new `maybeRollExpiredActiveSegment()` to do it.

gyang94 (approved):

wuchong (approved):
LGTM.

Issue #3844, opened 2026-08-04
[rust] Filter pushdown implementation


leekeiabstraction:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/fluss/issues) and found nothing similar.

### Motivation

Implement filter pushdown on log table

### Willingness to contribute

- [ ] I'm willing to submit a PR!

PR #3847, opened 2026-08-04
[rust] Add predicates as part of filter pushdown

### LINKED ISSUE #3846: [rust] predicate/proto wiring
### Search before asking

### Description

Update Proto for predicate wire format

### Willingness to contribute

leekeiabstraction:
## Summary
- Closes #3846.
- Adds `fluss::predicate`: an unresolved predicate tree (`Predicate::Leaf`/`Compound`) and a `col()` builder, the first piece of server-side filter pushdown for Rust log scans.
- Leaves reference columns by name and hold raw literals; binding to schema field ids and column types happens later, when the predicate is attached to a scan.
- `and`/`or` flatten into n-ary nodes to match the wire representation. The protocol has no negation node, so negation is expressed with `ne`/`not_in`.
- No behaviour change yet — nothing constructs or sends a predicate until the encoder lands.

## Test Plan
- `cargo test -p fluss-rs --lib` — 592 passed, including 6 new tests covering leaf construction, literal conversion, combinator flattening and the protocol wire codes.
- `cargo fmt` / `cargo clippy -p fluss-rs --lib` clean, apart from expected `dead_code` on the wire-code helpers the encoder will consume.

### DIFF
--- a/fluss-rust/crates/fluss/src/predicate/mod.rs
+++ b/fluss-rust/crates/fluss/src/predicate/mod.rs
@@ -1,0 +1,481 @@
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+//! Predicates for server-side filter pushdown on log scans.
+//!
+//! ```rust
+//! use fluss::predicate::col;
+//!
+//! let p = col("age").gt(30i64).and(col("name").starts_with("A"));
+//! ```
+//!
+//! The server prunes whole Arrow batches by their statistics, so a filtered scan
+//! returns a superset of the matching records and callers must still filter
+//! exactly. The protocol has no negation node; negate with [`ColumnRef::ne`] and
+//! [`ColumnRef::not_in`].
+//!
+//! Setting a predicate on a scan is only supported on log scans over tables
+//! with the ARROW log format.
+
+use crate::row::{Decimal, TimestampLtz, TimestampNtz};
+
+/// Comparison applied by a [`Predicate::Leaf`] to one column.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+pub enum LeafFunction {
+    Equal,
+    NotEqual,
+    LessThan,
+    LessOrEqual,
+    GreaterThan,
+    GreaterOrEqual,
+    IsNull,
+    IsNotNull,
+    StartsWith,
+    Contains,
+    EndsWith,
+    In,
+    NotIn,
+}
+
+impl LeafFunction {
+    /// Wire code for `PbLeafPredicate.function`.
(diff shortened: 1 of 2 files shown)


leekeiabstraction:
@charlesdong1991 @fresh-borzoni Appreciate a review here please 🙏

fluss-rust/crates/fluss/src/predicate/mod.rs
```diff
@@ -0,0 +1,468 @@
...
+
+//! Predicates for server-side filter pushdown on log scans.
+//!
+//! ```rust
+//! use fluss::predicate::col;
+//!
+//! let p = col("age").gt(30i64).and(col("name").starts_with("A"));
+//! ```
+//!
+//! The server prunes whole Arrow batches by their statistics, so a filtered scan
```
fresh-borzoni, line 26: On the Java side, TableScan throws when the log format is not Arrow, and throws again if a filter is set on a batch scanner. The same code written against this API would silently do nothing instead, which is a hard thing to discover. 
Should we mirror those checks when the predicate is attached to a scan?

fluss-rust/crates/fluss/src/predicate/mod.rs
```diff
@@ -0,0 +1,468 @@
...
+            Some(col("a").eq(1i32))
+        );
+
+        let folded = Predicate::and_all(vec![col("a").eq(1i32), col("b").eq(2i32)]).unwrap();
+        assert_eq!(folded, col("a").eq(1i32).and(col("b").eq(2i32)));
+    }
+
+    #[test]
+    fn wire_codes_match_the_protocol() {
+        assert_eq!(LeafFunction::Equal.code(), 0);
```
fresh-borzoni, line 465: why do we test only 5 out if 13?

fresh-borzoni (commented):
@leekeiabstraction Thank you, left one question, PTAL

fluss-rust/crates/fluss/src/predicate/mod.rs
leekeiabstraction: Agreed, I've added an additional doc line 

```
Setting a predicate on a scan is only supported on log scans over tables with the ARROW log format.
```

The behaviour to reject for batch scan or non-arrow log will be implemented in PR for #3844

fluss-rust/crates/fluss/src/predicate/mod.rs
leekeiabstraction: Nice catch, addressed.

leekeiabstraction:
@fresh-borzoni TY for the review. Addressed your comments, PTAL 🙏

fluss-rust/crates/fluss/src/predicate/mod.rs
```diff
@@ -0,0 +1,481 @@
...
+        self.leaf(LeafFunction::IsNull, vec![])
+    }
+
+    /// `column IS NOT NULL`.
+    pub fn is_not_null(self) -> Predicate {
+        self.leaf(LeafFunction::IsNotNull, vec![])
+    }
+
+    /// `column IN (values...)`. An empty `values` matches nothing.
+    pub fn is_in<V: Into<Literal>>(self, values: impl IntoIterator<Item = V>) -> Predicate {
```
charlesdong1991, line 302: when i check PredicateBuilder, it emits a real in leaf when >20 literals, otherwise or on equals due to performance reason, maybe nice to add a note not to revisit later

fluss-rust/crates/fluss/src/predicate/mod.rs
```diff
@@ -0,0 +1,481 @@
...
+    fn leaf(field: &str, function: LeafFunction, literals: Vec<Literal>) -> Predicate {
+        Predicate::Leaf {
+            field: field.to_string(),
+            function,
+            literals,
+        }
+    }
+
+    #[test]
+    fn builds_leaf_predicates() {
```
charlesdong1991, line 358: we likely have no need to test all functions since it's quite straightforward, but fyi here

charlesdong1991 (approved):
looks good to me!

fresh-borzoni (approved):
@leekeiabstraction LGTM, thank you

PR #3857, opened 2026-08-05
[security] Support impersonation via SASL/PLAIN authorization id

### LINKED ISSUE #3837: Support impersonation via SASL/PLAIN authorization id
### Search before asking

### Motivation

Middle-tier services (gateways, web consoles, query services) often connect to Fluss with their own service credentials but need to run operations on behalf of end users, so that ACLs are evaluated against the end user rather than the service account. This "impersonation" (a.k.a. doAs / proxy user) capability is standard in comparable systems: Hadoop proxy user, Hive doAs, Pulsar originalPrincipal, Trino impersonation.

The SASL/PLAIN protocol ([RFC 4616](https://datatracker.ietf.org/doc/html/rfc4616)) already carries this semantic natively via the optional authorization id:
```
message = [authzid] UTF8NUL authcid UTF8NUL passwd
```
Fluss currently parses the authorization id in `PlainSaslServer.evaluateResponse()` but unconditionally rejects any value different from the authenticated username, and the client hard-codes it to `null` in `SaslServerFactory.createSaslClient()`. The rest of the plumbing is already in place: `SaslServerAuthenticator.createPrincipal()` builds the principal from `saslServer.getAuthorizationID()`.

### Solution

Enable the RFC 4616 authorization id end-to-end, gated by an explicit server-side allowlist. No wire protocol change, no new SPI.

**Client:** new config option `client.security.sasl.authorization-id` (String, no default). When set, it is passed to `Sasl.createSaslClient(...)` and the JDK `PlainClient` encodes it per RFC 4616. When unset, the bytes on the wire are identical to today.

**Server:** allow `authzid != authcid` only if granted via a JAAS option on `PlainLoginModule`, consistent with the existing `user_<name>` style:

```yaml
security.sasl.plain.jaas.config: |
  org.apache.fluss.security.auth.sasl.plain.PlainLoginModule required
    user_admin="admin-pass"
    user_gateway="gw-pass"
    user_alice="alice-pass"
    impersonate_admin="*"                # admin may impersonate any user
    impersonate_gateway="alice,bob";     # gateway may only impersonate alice/bob
```
* If the check passes, the effective principal becomes the authorization id, and an INFO log records the impersonation for auditing. All downstream ACL checks apply to the effective principal, matching the existing `createPrincipal()` behavior.

* If `impersonate_<authcid>` is not configured, the request is rejected exactly as today — fully backward compatible, and impersonation cannot be enabled from the client side alone.

* The impersonator must still present its own valid password; the authorization id does not weaken authentication. Docs will keep the existing recommendation to use TLS with SASL/PLAIN.

Out of scope (possible follow-ups, would need a FIP): exposing original/effective principal on `Session`, an `IMPERSONATE` ACL operation type, and impersonation for future non-PLAIN mechanisms.

### Anything else?

_No response_

### Willingness to contribute

beryllw:
### Purpose

Linked issue: close #3837


fluss-common/src/test/java/org/apache/fluss/security/auth/sasl/plain/PlainSaslServerTest.java
litiliu, line 1: is it possible to add e2eIT？

fluss-common/src/main/java/org/apache/fluss/security/auth/sasl/plain/PlainImpersonationCallback.java
```diff
@@ -0,0 +1,60 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
```
litiliu: JDk already provided the [javax.security.sasl.AuthorizeCallback](https://docs.oracle.com/en/java/javase/11/docs/api/java.security.sasl/javax/security/sasl/AuthorizeCallback.html)，is it possible to reuse this class?

beryllw:
Thanks a lot for the review, @litiliu!
@loserwang1024 could you take a look when you get a chance?

litiliu (commented):
One compatibility concern with #3047: `FlussProtocolPlugin` materializes `security.sasl.plain.credentials` into `security.sasl.plain.jaas.config` by parsing and regenerating only `user_*` options. This PR introduces `impersonate_*` as a non-credential JAAS option, so it will be silently dropped whenever the credentials map is configured or dynamically updated.

Is impersonation intentionally unsupported together with `security.sasl.plain.credentials`? If not, the JAAS regeneration should preserve `impersonate_*`, with tests covering both startup and dynamic credentials updates.

beryllw:
I'll open a follow-up PR to support dynamically altering the impersonation grants themselves.

fluss-common/src/main/java/org/apache/fluss/security/auth/sasl/plain/PlainServerCallbackHandler.java
```diff
@@ -67,6 +68,16 @@
...
     private static final String JAAS_USER_PREFIX = "user_";
     private static final String JAAS_IMPERSONATE_PREFIX = "impersonate_";
+
+    /**
+     * The JAAS option prefixes this handler consumes, and the registry the protocol plugin must
+     * preserve when it regenerates the JAAS config from {@code security.sasl.plain.credentials}.
+     * Register a new prefix here when this handler starts consuming a new option; a unit test
+     * asserts the plugin's extraction covers every prefix listed here.
+     */
+    public static final List<String> KNOWN_OPTION_PREFIXES =
```
litiliu: Would it be simpler and more compatible to preserve all options from the original PlainLoginModule entry instead of maintaining a prefix allowlist? Since security.sasl.plain.credentials is syntactic sugar, materializing it should ideally overlay the corresponding user_* options without dropping unrelated JAAS options. We could parse the single entry through JaasConfig, copy AppConfigurationEntry#getOptions(), overlay the credentials, and serialize it safely. This would also avoid exposing and synchronizing KNOWN_OPTION_PREFIXES.

fluss-common/src/main/java/org/apache/fluss/security/auth/sasl/authenticator/SaslClientAuthenticator.java
```diff
@@ -147,7 +150,8 @@ public void initialize(AuthenticateContext context) throws AuthenticationExcepti
         }
 
         try {
-            saslClient = createSaslClient(mechanism, hostAddress, pros, loginManager);
+            saslClient =
```
loserwang1024, line 159: 1. why don't use an seperate authothenticator but use sasl/plain.
2. It seems that any user can react as CLIENT_SASL_AUTHORIZATION_ID? I don't think it is a right choice.

PR #3860, opened 2026-08-05
[server] Make timed-out rebalance tasks recoverable

### LINKED ISSUE #3859: [Improvement] Make timed-out rebalance tasks recoverable
### Search before asking

### Motivation

A bucket-level rebalance task currently has a timeout introduced by #3096. When the timeout is reached, the task may be treated as finished even though the replica migration is still in an intermediate state.

For example, a migration from:

```text
[0, 1, 2] -> [1, 0, 3]
```

may time out after reaching:

```text
[0, 1, 2, 3]
```

If the task is removed from tracking at this point, late AdjustIsr or NotifyLeaderAndIsr responses are ignored and the bucket may remain in an intermediate assignment indefinitely. Temporary ZooKeeper failures, RPC timeouts, network fluctuations, coordinator restarts, or TabletServer restarts can all expose this problem.

### Proposed solution

Make timeout a non-final execution state and keep the migration recoverable:

- Remove `TIMEOUT` from the final rebalance statuses.
- Release the normal execution slot after timeout so unrelated pending tasks can continue.
- Continue tracking and periodically reconciling the timed-out task.
- Retry the current migration phase idempotently from persisted assignment and LeaderAndIsr state.
- Avoid increasing the bucket epoch during repeated Phase A retries.
- Fence timeout, reconciliation, and NotifyLeaderAndIsr events by rebalance ID, bucket, attempt ID, coordinator epoch, leader epoch, and bucket epoch.
- Require the target assignment, leader, and ISR to reach the expected state before completion.
- Recover unfinished tasks after coordinator restart or failover.
- Persist cancellation intent before changing local task state, cancel tasks that have not started, and safely drain admitted tasks.
- Do not classify an origin assignment as clean if its ISR still contains replicas introduced by the migration.
- Retry persistence of the final status if the ZooKeeper update fails.
- Treat a bucket as completed if its table or partition is deleted during rebalance.

### Acceptance criteria

- A timed-out task remains visible as non-final and can complete after a late valid response.
- The next pending task can start without losing tracking of the timed-out attempt.
- Repeated Phase A reconciliation does not continuously increase the bucket epoch.
- Stale responses from an older attempt or epoch cannot advance or complete the current task.
- Coordinator restart resumes persisted Phase A, Phase B, cancellation, and finalization states idempotently.
- A dirty origin state such as origin assignment plus an extra ISR replica is not marked canceled.
- A temporarily unavailable target TabletServer can resume the migration after returning with the same server ID.
- Tests cover timeout, late completion, retry, fencing, restart, cancellation, deletion, and TabletServer failure/recovery.

### Non-goals

This task does not cover replacement of a TabletServer that remains offline for a long time. It also does not introduce configurable rebalance concurrency; that work is tracked independently by #3536 and can build on this recovery state machine.

### Willingness to contribute

swuferhong:
### Purpose

Linked issue: close #3859 

Treat rebalance timeout as a non-final state and continue tracking timed-out bucket migrations without blocking subsequent tasks.

Fence stale timeout and NotifyLeaderAndIsr events by rebalance attempt and request epochs. Reconcile migrations idempotently from persisted assignment and LeaderAndIsr state, and recover unfinished tasks after coordinator restart.

Persist cancellation intent before changing local state, drain admitted tasks safely, retry final-state persistence, and treat deleted buckets as completed.

### DIFF
--- a/fluss-server/src/main/java/org/apache/fluss/server/coordinator/rebalance/RebalanceManager.java
+++ b/fluss-server/src/main/java/org/apache/fluss/server/coordinator/rebalance/RebalanceManager.java
@@ -24,11 +24,14 @@ import org.apache.fluss.cluster.rebalance.RebalanceResultForBucket;
 import org.apache.fluss.cluster.rebalance.RebalanceStatus;
 import org.apache.fluss.cluster.rebalance.ServerTag;
 import org.apache.fluss.exception.NoRebalanceInProgressException;
+import org.apache.fluss.exception.RebalanceFailureException;
 import org.apache.fluss.metadata.TableBucket;
 import org.apache.fluss.server.coordinator.CoordinatorContext;
-import org.apache.fluss.server.coordinator.CoordinatorEventProcessor;
 import org.apache.fluss.server.coordinator.event.EventManager;
+import org.apache.fluss.server.coordinator.event.FinalizeRebalanceEvent;
 import org.apache.fluss.server.coordinator.event.RebalanceTaskTimeoutEvent;
+import org.apache.fluss.server.coordinator.event.ReconcileRebalanceTaskEvent;
+import org.apache.fluss.server.coordinator.event.RecoverRebalanceEvent;
 import org.apache.fluss.server.coordinator.rebalance.goal.Goal;
 import org.apache.fluss.server.coordinator.rebalance.goal.GoalOptimizer;
 import org.apache.fluss.server.coordinator.rebalance.model.ClusterModel;
@@ -64,16 +67,17 @@ import java.util.concurrent.TimeUnit;
 
 import static org.apache.fluss.cluster.rebalance.RebalanceStatus.CANCELED;
 import static org.apache.fluss.cluster.rebalance.RebalanceStatus.COMPLETED;
+import static org.apache.fluss.cluster.rebalance.RebalanceStatus.FAILED;
 import static org.apache.fluss.cluster.rebalance.RebalanceStatus.FINAL_STATUSES;
 import static org.apache.fluss.cluster.rebalance.RebalanceStatus.NOT_STARTED;
 import static org.apache.fluss.cluster.rebalance.RebalanceStatus.REBALANCING;
+import static org.apache.fluss.cluster.rebalance.RebalanceStatus.TIMEOUT;
 import static org.apache.fluss.utils.Preconditions.checkArgument;
-import static org.apache.fluss.utils.Preconditions.checkNotNull;
 
 /**
  * A rebalance manager to generate rebalance plan, and execution rebalance plan.
  *
- * <p>This manager can only be used in {@link CoordinatorEventProcessor} as a single threaded model.
+ * <p>This manager can only be used in the coordinator event loop as a single threaded model.
  */
 public class RebalanceManager {
     private static final Logger LOG = LoggerFactory.getLogger(RebalanceManager.class);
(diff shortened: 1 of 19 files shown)


fluss-server/src/main/java/org/apache/fluss/server/coordinator/rebalance/RebalanceManager.java
```diff
@@ -84,50 +94,72 @@ public class RebalanceManager {
...
     private static final long TIMEOUT_CHECK_INTERVAL_MS = 30 * 1000L;
 
+    /** Hardcoded upper bound for the exponential reconciliation backoff: 5 minutes. */
+    private static final long MAX_RECONCILE_BACKOFF_MS = 5 * 60 * 1000L;
+
+    /**
+     * Hardcoded time after which a timed-out task is given up on while it cannot make progress at
+     * all, because a target replica is not hosted by a live tablet server: 30 minutes.
+     */
+    private static final long TARGET_UNAVAILABLE_TIMEOUT_MS = 30 * 60 * 1000L;
```
platinumhamburg, line 104: Non-blocking: these constants define externally observable operational behavior: how long an unavailable target blocks a rebalance, how long a live-but-stalled migration may remain non-final, and how many timed-out migrations continue consuming coordinator and ZooKeeper resources.

A single hardcoded policy may not fit clusters with different sizes and recovery objectives. Please consider exposing these three values as validated coordinator configuration options while retaining the current values as defaults. The timeout-check interval and reconciliation backoff can remain internal implementation details.

fluss-server/src/test/java/org/apache/fluss/server/coordinator/rebalance/RebalanceManagerTest.java
```diff
@@ -230,106 +169,404 @@ void testTimeoutEnqueuesEvent() throws Exception {
...
+        recoveringManager.startup();
+
+        assertThat(recoveringManager.hasInProgressRebalance()).isTrue();
+        assertThat(recoveringManager.getRebalanceId()).isNull();
+        assertThat(recordingEventManager.events).hasSize(1);
+        RecoverRebalanceEvent recoveryEvent =
+                (RecoverRebalanceEvent) recordingEventManager.events.get(0);
+        assertThat(recoveryEvent.getRebalanceTask()).isEqualTo(storedTask);
+
+        recoveringManager.recoverRebalance(recoveryEvent.getRebalanceTask());
```
platinumhamburg, line 515: Non-blocking test suggestion: this verifies that startup emits a recovery event and that invoking recoverRebalance dispatches the persisted plan, but it does not exercise the end-to-end restart path through a rebuilt CoordinatorEventProcessor.

It would be useful to add a test that stops a processor while a bucket is in a Phase A or Phase B intermediate state, creates another processor against the same ZooKeeper state, and verifies that recovery runs automatically and converges to a persisted final state. The test could also verify that a delayed response from the old coordinator cannot advance the recovered attempt.

fluss-server/src/test/java/org/apache/fluss/server/coordinator/rebalance/RebalanceManagerTest.java
```diff
@@ -230,106 +169,404 @@ void testTimeoutEnqueuesEvent() throws Exception {
...
+                        clock,
+                        new RecordingEventManager(),
+                        new TestingRebalanceExecutor(coordinatorContext));
+
+        TableBucket tableBucket = new TableBucket(1L, 0);
+        manager.registerRebalance("keep-retrying-test", plans(tableBucket), NOT_STARTED);
+        RebalanceExecutionKey attempt = manager.getExecutionKey(tableBucket);
+        assertThat(manager.timeoutRebalanceTask(attempt)).isTrue();
 
+        clock.advanceTime(Duration.ofMinutes(31));
```
platinumhamburg, line 430: Non-blocking test suggestion: advancing the clock by 31 minutes only verifies that live targets bypass the 30-minute unavailable-target deadline. The separate 24-hour NO_PROGRESS_TIMEOUT_MS branch remains untested.

Please consider advancing the manual clock to immediately before and after that deadline while keeping the observable bucket state unchanged. The test should verify that reconciliation first continues and then transitions to FAILED, releases hasInProgressRebalance(), and no longer prevents a subsequent rebalance.

platinumhamburg (approved):
LGTM +1.

I have left three non-blocking suggestions concerning operational configurability and regression coverage.

LiebingYu (approved):

PR #3863, opened 2026-08-05
[flink] Support batch delete and update in flink sink

### LINKED ISSUE #3724: [flink] Support batch DELETE and Update for primary-key tables
### Search before asking

### Motivation

Currently, Fluss Flink Connector only supports DELETE pushdown when the DELETE condition contains full primary-key equality predicates, for example:
```sql
DELETE FROM orders WHERE id = 1001;
```
For batch DELETE statements with non-primary-key filters, partial primary-key filters, partition filters, or range conditions, Fluss does not support the row-level delete path yet. This issue aims to support batch DELETE for Fluss primary-key tables.

Users can run batch maintenance operations on Fluss tables, such as:

```sql
DELETE FROM orders WHERE dt = '2026-07-21';

DELETE FROM orders WHERE status = 'expired';

DELETE FROM orders WHERE update_time < TIMESTAMP '2026-07-01 00:00:00';
```

### Solution

_No response_

### Anything else?

_No response_

### Willingness to contribute

loserwang1024:
### Purpose

Linked issue: close #3724

### DIFF
--- a/fluss-flink/fluss-flink-common/src/test/java/org/apache/fluss/flink/sink/FlinkTableSinkITCase.java
+++ b/fluss-flink/fluss-flink-common/src/test/java/org/apache/fluss/flink/sink/FlinkTableSinkITCase.java
@@ -915,6 +915,64 @@ abstract class FlinkTableSinkITCase extends AbstractTestBase {
     }
 
     @Test
+    void testBatchDeleteStmtOnPkTable() throws Exception {
+        String tableName = "pk_table_batch_delete_test";
+        tBatchEnv.executeSql(
+                String.format(
+                        "create table %s ("
+                                + " a int not null,"
+                                + " b bigint, "
+                                + " c string,"
+                                + " primary key (a) not enforced"
+                                + ")",
+                        tableName));
+        List<String> insertValues =
+                Arrays.asList(
+                        "(1, 3501, 'Beijing')",
+                        "(2, 3502, 'Shanghai')",
+                        "(3, 3503, 'Berlin')",
+                        "(4, 3504, 'Seattle')",
+                        "(5, 3505, 'Boston')",
+                        "(6, 3506, 'London')");
+        tBatchEnv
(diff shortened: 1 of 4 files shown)


fluss-flink/fluss-flink-common/src/test/java/org/apache/fluss/flink/sink/FlinkTableSinkITCase.java
```diff
@@ -1102,14 +1201,20 @@ void testUnsupportedDeleteAndUpdateStmtOnPartialPK() {
...
+        rowIter = tBatchEnv.executeSql(String.format("select * from %s", t2)).collect();
+        assertResultsIgnoreOrder(
+                rowIter,
+                Arrays.asList(
+                        "+I[1, 4004, " + partition1 + "]",
+                        "+I[1, 4004, " + partition2 + "]",
+                        "+I[2, 2001, " + partition2 + "]"),
+                true);
+
+        tBatchEnv.executeSql("DELETE FROM " + t2 + " WHERE a = 1").await();
```
naivedogger, line 1214: This test uses `WHERE a = 1` on a table partitioned by `c`, so partition pruning is not covered. Since the issue explicitly mentions partition-filter deletes like `DELETE FROM orders WHERE dt = '2026-07-21'`, would it be possible to also add a partition-only filter case (e.g., `WHERE c = '<partition>'`) to cover that path?

naivedogger (approved):
Thanks @loserwang1024 for the pr! Overall LGTM, just left one minor suggestion.

fluss-flink/fluss-flink-common/src/test/java/org/apache/fluss/flink/sink/FlinkTableSinkITCase.java
loserwang1024: done it.

PR #3864, opened 2026-08-05
[server] Manage historical lookup local cache capacity

### LINKED ISSUE #3862: [FIP-28] Manage historical lookup local cache capacity
### Search before asking\n\n- [x] I searched in the [issues](https://github.com/apache/fluss/issues) and found nothing similar.\n\n### Description\n\nParent issue: #3631\n\nHistorical partition lookups materialize local Paimon lookup files on TabletServers. Without server-level cache management, cached lookupers from many tables can consume unbounded local disk and configuration changes can leave stale lookup state behind.\n\nAdd bounded local cache management for historical lookup with the following behavior:\n\n- Provide a server-wide historical lookup cache capacity and a per-table cache size.\n- Reserve configured capacity when admitting a table lookuper and evict inactive cached lookupers in least-recently-used order when capacity is needed.\n- Reject invalid per-table cache sizes and throttle admission when a new lookuper cannot fit.\n- Use stable per-table lookup cache directories and clean stale cache directories during startup.\n- Recreate cached lookupers when relevant table or lake configuration changes.\n- Allow the server-wide capacity to be updated dynamically through the server configuration procedure.\n- Expose focused metrics for cached table count, capacity evictions, historical lookup traffic, in-flight requests, and Paimon lookup file materialization.\n\n### Willingness to contribute\n\n- [x] I'm willing to submit a PR!

luoyuxia:
### Purpose

Linked issue: closes #3862

Historical partition lookups materialize local Paimon lookup files on TabletServers. This change bounds their configured disk capacity across tables and keeps cached lookupers consistent when table or lake configuration changes.

### Brief change log

- Add server-wide and per-table historical lookup cache size options.
- Reserve capacity per cached table lookuper and use LRU eviction when admission needs space.
- Use stable per-table cache directories and clean stale directories during startup.
- Recreate cached lookupers after relevant table or lake configuration changes.
- Support dynamic server cache-capacity updates through the cluster configuration procedure.
- Add focused cache and historical lookup metrics.

### Tests

- Common configuration tests: 8 passed.
- Server cache budget, lifecycle, validation, and metric tests: 23 passed.
- Paimon lake lookuper tests: 9 passed.
- Flink 1.20 procedure integration tests: 21 passed.

### API and Format

Adds server and table configuration options and extends the evolving lake lookup context with a metric recorder. No RPC wire-format or storage-format changes.

### Documentation

Configuration descriptions are included. Cache lifecycle and capacity semantics are covered by the historical lookup local cache management design document.

### Generative AI disclosure

- [x] Codex (GPT-5), following the repository AGENTS.md guidelines.


fluss-server/src/main/java/org/apache/fluss/server/replica/HistoricalLakeLookupManager.java
```diff
@@ -259,39 +312,71 @@ void invalidateTableLookuper(long tableId) {
...
+    }
+
+    int numInflightRequests() {
+        return maxQueuedHistoricalRequests - lookupPermits.availablePermits();
+    }
+
+    void reconfigure(Configuration newConf) {
+        checkNotNull(newConf, "newConf must not be null.");
+        boolean lakeConfigChanged;
+        synchronized (this) {
```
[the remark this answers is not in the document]
luoyuxia: Thanks. The publication order is intentional rather than requiring an atomic pair. Both fields are volatile: reconfigure publishes `conf` before `lakeConfigVersion`, while lookup reads the version before `conf`. Therefore, if a lookup observes the new version, the volatile happens-before relationship guarantees that it also observes the corresponding configuration.

The only possible mixed observation is old version + new configuration. That is safe: an existing old lookuper may finish the lookup as a pre-update request, while a newly created lookuper uses the new configuration but carries the old version and is retired by `invalidateAll()` or by the next lookup through the version mismatch. The unsafe new version + old configuration combination cannot occur, so an atomic holder is not required here.

fluss-server/src/main/java/org/apache/fluss/server/replica/HistoricalLakeLookupManager.java
```diff
@@ -259,39 +312,71 @@ void invalidateTableLookuper(long tableId) {
...
+            }
+
+            lakeConfigChanged = hasLakeConfigChanged(conf, newConf);
+            // Publish the configuration before its version. A lookup that observes the new version
+            // must also observe the matching configuration snapshot.
+            conf = newConf;
+            if (lakeConfigChanged) {
+                lakeConfigVersion++;
+            }
+        }
```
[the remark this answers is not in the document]
luoyuxia: Thanks. The publication order is intentional rather than requiring an atomic pair. Both fields are volatile: reconfigure publishes `conf` before `lakeConfigVersion`, while lookup reads the version before `conf`. Therefore, if a lookup observes the new version, the volatile happens-before relationship guarantees that it also observes the corresponding configuration.

The only possible mixed observation is old version + new configuration. That is safe: an existing old lookuper may finish the lookup as a pre-update request, while a newly created lookuper uses the new configuration but carries the old version and is retired by `invalidateAll()` or by the next lookup through the version mismatch. The unsafe new version + old configuration combination cannot occur, so an atomic holder is not required here.

fluss-lake/fluss-lake-paimon/src/main/java/org/apache/fluss/lake/paimon/lookup/PaimonLakeTableLookuper.java
```diff
@@ -106,6 +101,7 @@ public class PaimonLakeTableLookuper implements LakeTableLookuper {
     private @Nullable LocalTableQuery localTableQuery;
     private @Nullable RowPartitionKeyExtractor partitionKeyExtractor;
     private int primaryKeyFieldCount;
+    private long lookupFileMaterializationCount;
```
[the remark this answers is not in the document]
luoyuxia: `PaimonLakeTableLookuper.lookup()` is synchronized, so lookups on the same cached lookuper are serialized. The counter snapshot, the Paimon lookup, and the final comparison all execute while holding the lookuper monitor. `TrackingIOManager.createChannel(prefix)` is invoked synchronously from Paimon 1.3.1 through the `LocalTableQuery` / `LookupLevels.createLookupFile` call path, so the increment also occurs within that same synchronized lookup invocation.

Therefore, concurrent server requests cannot overlap these counter deltas or cause cross-request attribution. `close()` is synchronized for the same shared-state reason, so an atomic counter is not needed here.

fluss-server/src/main/java/org/apache/fluss/server/replica/HistoricalLakeLookupManager.java
```diff
@@ -311,8 +396,253 @@ private LookupResultForBucket lookupInternal(
...
+                        clusterConf);
+        // Replace the reservation atomically: the old and replacement cache sizes never count
+        // against the global budget at the same time.
+        Reservation reservation =
+                budgetManager.tryReplace(currentLookuper.reservation, cacheSizeBytes);
+        if (reservation == null) {
+            // The candidate was never published, while the current lookuper remains usable.
+            closeLookuper(lookuper, tableLookupDir);
+            return null;
+        }
```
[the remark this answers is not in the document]
luoyuxia: The create-before-replace ordering is intentional. `tryReplace()` mutates the budget state rather than performing a read-only capacity check. Reserving first would retire the current reservation before the replacement lookuper exists, and a later creation failure would require a dedicated rollback operation that remains correct across concurrent limit changes and removal callbacks.

The current Paimon lookuper is lazy: construction only stores configuration. The catalog, IOManager, local directories, and lookup files are initialized on the first `lookup()`, after `tryReplace()` succeeds and the candidate is published. Therefore, failed replacement admission does not materialize unreserved local cache data. We accept the small amount of temporary wrapper/plugin construction so creation failures leave the existing lookuper and reservation unchanged.

fluss-common/src/main/java/org/apache/fluss/config/ConfigOptions.java
```diff
@@ -428,6 +428,25 @@ public class ConfigOptions {
                             "The maximum number of threads used for historical partition operations, such as lake lookups and writes. "
                                     + "Threads are started lazily and released after the keep-alive timeout when idle.");
 
+    public static final ConfigOption<MemorySize>
+            SERVER_HISTORICAL_PARTITION_LOOKUP_CACHE_MAX_DISK_SIZE =
+                    key("server.historical-partition.lookup-cache.max-disk-size")
+                            .memoryType()
+                            .defaultValue(MemorySize.parse("80gb"))
```
platinumhamburg: Since we can't assume what hardware the service will run on, we use percentage-based settings for disk capacity protection. Could this parameter be changed to a percentage as well?

fluss-common/src/main/java/org/apache/fluss/config/ConfigOptions.java
```diff
@@ -1850,6 +1869,15 @@ public class ConfigOptions {
                                     + "to look up historical partition data so that their clients load the "
                                     + "updated table configuration.");
 
+    public static final ConfigOption<MemorySize>
+            TABLE_DATALAKE_HISTORICAL_PARTITION_LOOKUP_CACHE_MAX_DISK_SIZE =
+                    key("table.datalake.historical-partition.lookup-cache.max-disk-size")
+                            .memoryType()
+                            .defaultValue(MemorySize.parse("8gb"))
```
platinumhamburg: Ditto

luoyuxia:
@platinumhamburg Thanks for review. Comments has been addressed.

fluss-rpc/src/main/java/org/apache/fluss/rpc/netty/server/NettyServerHandler.java
```diff
@@ -298,13 +300,16 @@ private void updateRequestMetrics(FlussRequest request, long requestEndTimeMs) {
...
     private Optional<RequestsMetrics.Metrics> getMetrics(FlussRequest request) {
         boolean isFromFollower = false;
+        boolean isHistoricalLookup = false;
         ApiMessage requestMessage = request.getMessage();
         if (request.getApiKey() == ApiKeys.FETCH_LOG.id) {
             // for fetch, we need to identify it's from client or follower
             FetchLogRequest fetchLogRequest = (FetchLogRequest) requestMessage;
             isFromFollower = fetchLogRequest.getFollowerServerId() >= 0;
+        } else if (request.getApiKey() == ApiKeys.LOOKUP.id) {
+            isHistoricalLookup = hasHistoricalLookup((LookupRequest) requestMessage);
```
platinumhamburg: Adding a loop computation for every request here seems a bit inefficient. It shouldn't cause any significant performance regression currently, but it might be worth adding a TODO to revisit this later.

platinumhamburg (approved):
LGTM.

wuchong:
There are "merge" commits, please remove the "merge" commit from the PR's history.

luoyuxia:
> There are "merge" commits, please remove the "merge" commit from the PR's history.

merge commit has been removed

fluss-server/src/main/java/org/apache/fluss/server/replica/HistoricalLakeLookupManager.java
```diff
@@ -311,8 +434,253 @@ private LookupResultForBucket lookupInternal(
...
+        return new CachedLakeTableLookuper(
```
wuchong: The cache-miss and replacement branches duplicate the table-directory resolution, lookuper construction, and the eight-argument `CachedLakeTableLookuper` construction. Could we extract these mechanical creation/wrapping steps into helpers while keeping the different reservation ordering in each branch? A miss must reserve before creating resources, whereas replacement must create the candidate before `tryReplace`, so the transaction-specific cleanup should remain local. Centralizing the shared construction would reduce the risk that a future constructor or configuration change updates only one path.

fluss-server/src/main/java/org/apache/fluss/server/replica/HistoricalLakeLookupManager.java
```diff
@@ -82,51 +91,58 @@
...
+    private boolean historicalLookupCacheRootDirCreated;
+
+    private volatile boolean started;
 
     HistoricalLakeLookupManager(
             Configuration conf,
             @Nullable PluginManager pluginManager,
-            int serverId,
+            File dataDir,
+            long lookupVolumeBytes,
```
wuchong: Rename to `dataDirVolumeBytes`. IIUC, this is not the lookupDir volume size, but the whole dataDir volume size.

fluss-server/src/main/java/org/apache/fluss/server/replica/HistoricalLakeLookupManager.java
```diff
@@ -208,7 +264,15 @@ CompletableFuture<LookupResultForBucket> lookup(
 
         CompletableFuture<LookupResultForBucket> future;
         try {
-            future = submitLookup(lookupData, tableInfo, schemaInfo);
+            future =
+                    submitLookup(
+                            lookupData,
+                            tableInfo,
+                            schemaInfo,
+                            tableConfig,
```
wuchong: `tableConfig` can be accessed from `tableInfo`, we don't need additional parameter.

fluss-common/src/main/java/org/apache/fluss/config/ConfigOptions.java
```diff
@@ -1858,6 +1876,15 @@ public class ConfigOptions {
                                     + "to look up historical partition data so that their clients load the "
                                     + "updated table configuration.");
 
+    public static final ConfigOption<Double>
+            TABLE_DATALAKE_HISTORICAL_PARTITION_LOOKUP_CACHE_MAX_DISK_RATIO =
+                    key("table.datalake.historical-partition.lookup-cache.max-disk-ratio")
```
wuchong: I do not think we should expose `table.datalake.historical-partition.lookup-cache.max-disk-ratio` as a table option. It is difficult for users to configure meaningfully because the ratio is based on a TabletServer data volume, while each server owns only a changing subset of the table buckets as the cluster scales. It is also too coarse as an upfront reservation: for example, a 1 MB table still reserves 1% of a 1 TB volume, and the default 10% global / 1% per-table settings admit only ten tables, so twenty active tables can churn at table granularity despite using little physical space. Removing the public option would also eliminate the fragile `metadataManager.updateHistoricalLookupCacheMaxRatio` dynamic-config path and table-descriptor validation. I would prefer a single cluster-level disk-cache budget, with tables competing through LRU/TTL so hot tables can use more of the shared capacity and operators can expand the global ratio dynamically. If changing the allocator is too large for this PR, please keep the per-table ratio as an internal default for now rather than exposing a public table property that will be difficult to remove later.

fluss-server/src/main/java/org/apache/fluss/server/replica/HistoricalLookupCacheBudgetManager.java
```diff
@@ -0,0 +1,175 @@
...
+ * <ul>
+ *   <li>Each table ID has at most one current reservation.
+ *   <li>{@code reservedBytes} is the sum of the reservations in {@code reservationsByTableId}.
+ *   <li>{@code reservedBytes >= 0}. After a dynamic limit reduction, existing reservations may
+ *       exceed {@code maxBytes}; the caller may evict cached lookupers before retrying the next
+ *       admission.
+ * </ul>
+ */
+@ThreadSafe
+final class HistoricalLookupCacheBudgetManager {
```
wuchong: Move to `org.apache.fluss.server.replica.historical` package with `HistoricalLakeLookupManager` together.

fluss-server/src/main/java/org/apache/fluss/server/replica/Replica.java
```diff
@@ -186,7 +186,8 @@ public final class Replica {
 
     private final SchemaGetter schemaGetter;
     private final TableInfo tableInfo;
-    private final TableConfig tableConfig;
+    // Metadata updates replace this snapshot after applying configuration-specific side effects.
+    private volatile TableConfig tableConfig;
```
wuchong: Should we hot-update the `tableInfo`? Otherwise, the `tableConfig` within `tableInfo` may lag behind `this.tableConfig`, which can easily lead to bugs. If we do this, we won't need to maintain `this.tableConfig` separately and can simply use `tableInfo.getTableConfig()`.

But if we remove the per-table level ratio, then we don't need this.

fluss-server/src/main/java/org/apache/fluss/server/replica/ReplicaManager.java
```diff
@@ -359,13 +360,26 @@ public ReplicaManager(
         this.ioExecutor = ioExecutor;
         this.minInSyncReplicas = conf.get(ConfigOptions.LOG_REPLICA_MIN_IN_SYNC_REPLICAS_NUMBER);
         this.scannerManager = checkNotNull(scannerManager, "scannerManager");
+        // Historical lookup cache capacity currently uses only the first data volume.
+        File historicalLookupDataDir = localDiskManager.dataDirs().get(0);
+        long historicalLookupVolumeBytes =
+                Files.getFileStore(historicalLookupDataDir.toPath()).getTotalSpace();
```
wuchong: Should rename to `dataDir` and `dataDirVolumeBytes`?

fluss-server/src/main/java/org/apache/fluss/server/replica/ReplicaManager.java
```diff
@@ -359,13 +360,26 @@ public ReplicaManager(
...
         this.historicalLakeLookupManager =
-                new HistoricalLakeLookupManager(conf, pluginManager, serverId, scheduler);
+                new HistoricalLakeLookupManager(
+                        conf,
+                        pluginManager,
+                        historicalLookupDataDir,
+                        historicalLookupVolumeBytes,
+                        scheduler);
+        serverMetricGroup.registerHistoricalPartitionInflightRequests(
+                "lookup", historicalLakeLookupManager::numInflightRequests);
```
wuchong: move into registerMetrics()

fluss-server/src/main/java/org/apache/fluss/server/replica/ReplicaManager.java
```diff
@@ -847,18 +889,27 @@ public void historicalLookups(
                                         data.originalPartitionName(),
                                         ApiError.fromThrowable(e)));
             }
+            TableMetricGroup historicalLookupMetrics = tableMetrics;
```
wuchong: It seems unnecessary to rename it. After all, this isn't a `HistoricalLookupMetricGroup`, so is there any issue with using `tableMetrics` directly?

fluss-rpc/src/main/java/org/apache/fluss/rpc/netty/server/RequestsMetrics.java
```diff
@@ -59,6 +61,9 @@ private RequestsMetrics(MetricGroup serverMetricsGroup, Collection<ApiKeys> apiK
                 addMetrics(serverMetricsGroup, toRequestName(apiKey, false));
             }
         }
+        if (apiKeys.contains(ApiKeys.LOOKUP)) {
+            addMetrics(serverMetricsGroup, HISTORICAL_LOOKUP_METRICS_KEY);
+        }
```
wuchong: Unify this into `toRequestName()` to avoid special handling, similar to how we handle `isFromFollower`. The `isHistorical` flag can also be passed as a parameter. We may have `historicalPut` in the future.

fluss-server/src/main/java/org/apache/fluss/server/utils/ServerRpcMessageUtils.java
```diff
@@ -1776,6 +1781,10 @@ public static ListPartitionInfosResponse toListPartitionInfosResponse(
         ListPartitionInfosResponse listPartitionsResponse = new ListPartitionInfosResponse();
         for (Map.Entry<String, PartitionRegistration> partitionRegistration :
                 partitionRegistrations.entrySet()) {
+            // TODO: Return the actual lake partitions instead of the internal historical partition.
+            if (HISTORICAL_PARTITION_VALUE.equals(partitionRegistration.getKey())) {
```
wuchong: It is recommended to directly remove `HISTORICAL_PARTITION_VALUE` from `partitionRegistrations` within `org.apache.fluss.server.RpcServiceBase.listPartitionInfos`. This keeps the business logic at the upper layer and ensures that `ServerRpcMessageUtils` remains a pure conversion utility without embedding business logic.

fluss-server/src/main/java/org/apache/fluss/server/TabletManagerBase.java
```diff
@@ -115,6 +116,9 @@ protected List<File> listTabletsToLoad(File dataDir) {
         // Get all database directory.
         File[] dbDirs = FileUtils.listDirectories(dataDir);
         for (File dbDir : dbDirs) {
+            if (dbDir.getName().equals(HISTORICAL_LOOKUP_CACHE_DIR_NAME)) {
```
wuchong: Let's filter `REMOTE_LOG_INDEX_LOCAL_CACHE` as well here?

fluss-server/src/main/java/org/apache/fluss/server/metrics/group/TabletServerMetricGroup.java
```diff
@@ -239,6 +240,20 @@ public Counter failedIsrUpdates() {
...
+     *
+     * @param operation historical partition operation
+     * @param inflightRequests gauge for accepted requests that have not completed
+     */
+    public void registerHistoricalPartitionInflightRequests(
+            String operation, Gauge<Integer> inflightRequests) {
+        HistoricalPartitionOperationMetricGroup operationMetricGroup =
+                new HistoricalPartitionOperationMetricGroup(registry, this, operation);
+        operationMetricGroup.gauge(
+                MetricNames.HISTORICAL_PARTITION_INFLIGHT_REQUESTS, inflightRequests);
```
wuchong: We do not recommend registering user-defined `MetricGroup` instances, as this requires manual lifecycle management, including explicit calls to `close()`.

In this case, I believe registering a standard metric for the number of in-flight requests is sufficient (without the `operation` label). Future in-flight write requests can also be included in this metric, as they all rely on lake lookups.

fluss-server/src/main/java/org/apache/fluss/server/metrics/group/TableMetricGroup.java
```diff
@@ -577,6 +614,95 @@ protected String getGroupName(CharacterFilter filter) {
         }
     }
 
+    /**
+     * Collects historical lookup metrics in the KV table scope and separates lake lookup metrics by
+     * whether the lookup materialized local files.
+     */
+    private static class HistoricalLookupMetricGroup extends AbstractMetricGroup {
```
wuchong: same here.

fluss-server/src/main/java/org/apache/fluss/server/replica/HistoricalLakeLookupManager.java
```diff
@@ -82,51 +91,58 @@
...
+    @GuardedBy("this")
+    private boolean historicalLookupCacheRootDirCreated;
+
+    private volatile boolean started;
 
     HistoricalLakeLookupManager(
             Configuration conf,
             @Nullable PluginManager pluginManager,
-            int serverId,
+            File dataDir,
```
wuchong, line 157: **P0**: Passing only `dataDir` and `lookupVolumeBytes` leaves this manager disconnected from the disk-write protection maintained by `LocalDiskManager`. When the data volume crosses the write-limit ratio, append/put requests are stopped, but historical lookups can still create and populate local lookup-cache files, increasing pressure on an already full disk and potentially exhausting it. Please inject `LocalDiskManager` here and make cache I/O participate in disk protection: stop admitting or materializing new cache files while the disk is write-locked and/or invalidate cached lookupers to release their files until usage recovers.

fluss-common/src/main/java/org/apache/fluss/metrics/MetricNames.java
```diff
@@ -102,6 +104,12 @@ public class MetricNames {
     public static final String DISK_USAGE_RATIO = "diskUsageRatio";
     public static final String DISK_WRITE_LOCKED = "diskWriteLocked";
 
+    // for historical lookup cache
+    public static final String HISTORICAL_LOOKUP_CACHED_TABLE_COUNT =
```
wuchong: Could we standardize the historical lookup metrics by introducing `historical` as a dedicated metric group at both the TabletServer and table scopes?

The current names encode the domain inconsistently: `historicalPartition*`, `historicalLookup*`, `totalHistorical*`, and unprefixed `lake*`. This makes metric discovery and dashboard queries unnecessarily inconsistent. I suggest exposing the following Prometheus surface (excluding the common TabletServer labels):

| Scope | Prometheus metric name | Additional labels |
|---|---|---|
| TabletServer | `fluss_tabletserver_historical_inflightRequests` | - |
| TabletServer | `fluss_tabletserver_historical_lookupCacheDiskSize` |  New added |
| TabletServer | `fluss_tabletserver_historical_lookupCacheTableCount` | - |
| TabletServer | `fluss_tabletserver_historical_lookupCacheCapacityEvictions` | - |
| Table | `fluss_tabletserver_table_historical_totalLookupRequestsPerSecond` | `database`, `table`, `tablet_type="kv"` |
| Table | `fluss_tabletserver_table_historical_failedLookupRequestsPerSecond` | same as above |
| Table | `fluss_tabletserver_table_historical_lakeLookupsPerSecond` | same as above, plus `lookup_file_materialization="true|false"` |
| Table | `fluss_tabletserver_table_historical_lakeLookupTimeMs` | same as above, plus `lookup_file_materialization` and `quantile` |

Implementation-wise, the server-level metrics could be registered under `serverMetricGroup.addGroup("historical")`. At the table scope, `HistoricalLookupMetricGroup#getGroupName()` could return `"historical"`, while the individual metric names are shortened to `totalLookupRequestsPerSecond`, `lakeLookupTimeMs`, etc. This keeps the hierarchy in the metric scope instead of repeating it inconsistently in each metric name.

Since lookup is currently the only historical operation, `inflightRequests` does not need an additional `operation="lookup"` label. Future in-flight historical operations can be included in this aggregate metric.

Please also add `lookupCacheDiskSize` as a gauge reporting the current on-disk cache footprint in bytes. With the existing TabletServer labels, this allows operators to monitor cache disk usage and its distribution across the cluster.

fluss-server/src/main/java/org/apache/fluss/server/metrics/group/TableMetricGroup.java
```diff
@@ -577,6 +614,95 @@ protected String getGroupName(CharacterFilter filter) {
...
+        }
+
+        private void recordLookup(long lookupTimeNanos) {
+            lakeLookups.inc();
+            lakeLookupTimeMs.update(TimeUnit.NANOSECONDS.toMillis(lookupTimeNanos));
+        }
+
+        @Override
+        protected void putVariables(Map<String, String> variables) {
+            variables.put("lookup_file_materialization", String.valueOf(lookupFileMaterialization));
```
wuchong: Could we rename the `lookup_file_materialization` label to `lookup_file_downloaded="true|false"`?
“Materialization” is not very intuitive operationally and can easily be interpreted as loading data into memory.

fluss-server/src/main/java/org/apache/fluss/server/metrics/group/TabletServerMetricGroup.java
luoyuxia: I was thinking to use operation label for future write metric, but I agree that we can unify it.

fluss-server/src/main/java/org/apache/fluss/server/replica/HistoricalLookupCacheBudgetManager.java
luoyuxia: Remove HistoricalLookupCacheBudgetManager, let's keep HistoricalLakeLookupManager still in `org.apache.fluss.server.replica` package.

fluss-server/src/main/java/org/apache/fluss/server/metrics/group/TableMetricGroup.java
luoyuxia: we may still need a new MetricGroup for this since we need to register metric with same name but different label.

luoyuxia:
@wuchong Thanks for your review. Comments has been addressed

fluss-server/src/main/java/org/apache/fluss/server/replica/HistoricalLakeLookupManager.java
```diff
@@ -311,8 +524,28 @@ private LookupResultForBucket lookupInternal(
...
     }
 
+    private void ensureDiskWritable() {
+        try {
+            diskWriteGuard.run();
+        } catch (DiskWriteLockedException e) {
+            // Release local cache files after the disk is write-locked. Idle lookupers close
+            // immediately, while active lookupers close after their final lookup releases them.
+            // Run pending cache maintenance now so removal callbacks are processed promptly.
+            lakeTableLookupers.invalidateAll();
```
wuchong: I believe we should avoid invalidating and clearing the cache here, as this would significantly impact online historical lookups. We should aim to maintain a certain level of cache stability. We can remove the `invlidateAll()` and `cleanup()` in this PR. 

In the future, we could introduce a `min-disk-ratio` (default=0.01) to manage disk usage. When disk usage is high, we can clear the cache down to the `min-disk-ratio` to free up space for write operations. When disk usage is low, we can allow the cache to expand up to a `max-disk-ratio`. (Add this as TODO comment as a future work).

fluss-server/src/main/java/org/apache/fluss/server/replica/HistoricalLakeLookupManager.java
```diff
@@ -190,9 +254,55 @@ private static com.github.benmanes.caffeine.cache.Scheduler createCacheScheduler
...
     CompletableFuture<LookupResultForBucket> lookup(
-            LookupDataForBucket lookupData, TableInfo tableInfo, SchemaInfo schemaInfo) {
+            LookupDataForBucket lookupData,
+            TableInfo tableInfo,
+            SchemaInfo schemaInfo,
+            LakeTableLookuper.LookupMetricRecorder lookupMetricRecorder) {
+        checkState(started, "Historical lake lookup manager has not been started.");
         TableBucket tableBucket = lookupData.tableBucket();
+        try {
+            ensureDiskWritable();
```
wuchong: We should not call `ensureDiskWritable()` at the entry point of the lookup, as a lookup operation does not necessarily involve writes. Instead, we should check `ensureDiskWritable()` within `org.apache.fluss.lake.paimon.lookup.PaimonLakeTableLookuper.TrackingIOManager#createChannel(java.lang.String)`. If a `DiskWriteLockedException` occurs, it should be wrapped in an `IOException` and propagated upward. Then, in the exception handling block of `org.apache.fluss.lake.lakestorage.LakeTableLookuper#lookup`, we can unwrap it back to a `DiskWriteLockedException` (a retriable exception) before returning it to the client.

This approach ensures that pure lookup operations are not blocked; blocking only occurs when downloading cache files.

fluss-server/src/main/java/org/apache/fluss/server/replica/HistoricalLakeLookupManager.java
```diff
@@ -160,17 +213,28 @@ class HistoricalLakeLookupManager implements AutoCloseable {
...
-                                        cachedLookuper.invalidate();
+                                        if (cause == RemovalCause.SIZE) {
+                                            capacityEvictions.inc();
+                                            LOG.info(
+                                                    "Evicted historical lookup cache for table {} (table ID {}) because the cache retains at most {} tables.",
+                                                    cachedLookuper.tablePath,
+                                                    cachedLookuper.tableId,
+                                                    MAX_CACHED_TABLES);
+                                        }
+                                        onLookuperRemoved(cachedLookuper);
```
wuchong: Could we simplify this listener to a single `onLookuperRemoved(cachedLookuper, cause)` call and move the null check, size-eviction accounting/logging, and invalidation into that method? That would keep the removal policy in one place and leave the Caffeine builder easier to scan.


```java
    Caffeine.newBuilder()
      .removalListener(this::onLookuperRemoved);

    private void onLookuperRemoved(
            Long ignored, @Nullable CachedLakeTableLookuper cachedLookuper, RemovalCause cause) {
        if (cachedLookuper == null) {
            return;
        }
        if (cause == RemovalCause.SIZE) {
            capacityEvictions.inc();
            LOG.info(
                    "Evicted historical lookup cache for table {} (table ID {}) because the cache retains at most {} tables.",
                    cachedLookuper.tablePath,
                    cachedLookuper.tableId,
                    MAX_CACHED_TABLES);
        }
        cachedLookuper.invalidate();
    }
```

fluss-server/src/main/java/org/apache/fluss/server/replica/HistoricalLakeLookupManager.java
```diff
@@ -369,37 +607,94 @@ LakeTableLookuper createLakeTableLookuper(
...
+                || !Objects.equals(
+                        extractLakeProperties(currentConf), extractLakeProperties(newConf));
     }
 
-    private static String preparePaimonLookupTempDir(Configuration conf, int serverId) {
-        File paimonLookupTempDir =
-                new File(
-                        new File(conf.get(ConfigOptions.SERVER_IO_TMP_DIR), PAIMON_LOOKUP_DIR_NAME),
-                        String.valueOf(serverId));
+    private synchronized File getOrCreateHistoricalLookupCacheRootDir() {
```
wuchong: Create the `historicalLookupCacheRootDir` directly during startup to avoid acquiring a lock on every lookup. It is just a directory and does not consume significant space.

fluss-server/src/main/java/org/apache/fluss/server/replica/ReplicaManager.java
```diff
@@ -359,13 +360,25 @@ public ReplicaManager(
...
+                        dataDir,
+                        dataDirVolumeBytes,
+                        scheduler);
 
         registerMetrics();
     }
 
     public void startup() {
+        historicalLakeLookupManager.startup();
+        historicalLakeLookupManager.startLookupCacheDiskSizeMonitor(scheduler);
```
wuchong: nit: Can we merge this 2 methods? `historicalLakeLookupManager.startup(scheduler)`

wuchong (commented):
@luoyuxia , I only left some minor comments.

luoyuxia:
@wuchong Thanks for helpful review. Comments has been addressed.

wuchong (approved):

Issue #3495, opened 2026-06-18
[Feature] Support SASL/OAUTHBEARER authentication for Fluss RPC


litiliu:
## Search before asking

- [x] I searched in the issues and found nothing similar.

## Motivation

Fluss currently supports pluggable authentication and built-in SASL/PLAIN
authentication. In many production environments, service identity is managed through
OAuth2/OIDC providers and clients authenticate with bearer tokens, usually JWTs.

It would be useful for Fluss to support the standard SASL/OAUTHBEARER mechanism
in the native Fluss RPC authentication layer. This would allow clients to
authenticate with OAuth bearer tokens, and allow Fluss servers to validate those
tokens and map validated claims to `FlussPrincipal` for the existing ACL authorization flow.

This feature can be designed to work together with [FIP-29: (m)TLS Support](https://cwiki.apache.org/confluence/display/FLUSS/FIP-29%3A+%28m%29TLS+Support). FIP-29 keeps transport security and authentication orthogonal: TLS protects
the transport, while `client.security.protocol` and `security.protocol.map` select the authentication protocol. Following that model, OAuth bearer token
authentication should be introduced as a SASL mechanism.

## Solution

Add support for `OAUTHBEARER` as a SASL mechanism:

```properties
client.security.protocol=SASL
client.security.sasl.mechanism=OAUTHBEARER
```

Server-side example:

```yaml
security.protocol.map: CLIENT:SASL
security.sasl.enabled.mechanisms: OAUTHBEARER
security.sasl.oauthbearer.jwks.endpoint.url: https://issuer.example.com/.well-known/jwks.json
security.sasl.oauthbearer.expected.issuer: https://issuer.example.com
security.sasl.oauthbearer.expected.audience: fluss
security.sasl.oauthbearer.principal.claim.name: sub
```

Proposed implementation direction:

- Refactor the current SASL implementation so SASL mechanisms are extensible
  instead of hard-coding only `PLAIN`.
- Preserve existing SASL/PLAIN behavior and compatibility.
- Add an `OAUTHBEARER` mechanism for the native Fluss RPC authentication layer.
- Add client-side token provider support.
- Add server-side JWT/JWKS token validation.
- Validate token signature, issuer, audience, expiration, not-before, and
  configurable clock skew.
- Map a configurable token claim to `FlussPrincipal`.
- Reuse the existing ACL authorization flow after authentication succeeds.
- Add tests for successful authentication, invalid token, expired token, wrong
[...]

PR #3865, opened 2026-08-05
[security] Make SASL mechanisms extensible

litiliu:
### Purpose

Related issue: #3495

This PR prepares the Fluss RPC SASL authentication layer for adding built-in mechanisms such as OAUTHBEARER. The existing implementation constructs PLAIN authenticators directly, which makes adding another mechanism require changing the PLAIN implementation path.

This is intentionally a foundation-only change. It does not add OAuth token acquisition, JWT validation, JWKS handling, group principals, or ACL changes.

### Brief change log

- Add a stateless `SaslAuthenticatorFactory` that selects a built-in SASL mechanism and creates one connection-local authenticator.
- Extract the existing PLAIN server implementation into `PlainSaslServerAuthenticator`.
- Keep `SaslServerAuthenticator` as the connection-local dispatcher that validates the requested server mechanism and delegates authentication.
- Preserve the existing PLAIN configuration, JAAS fallback, RPC handshake, and authentication behavior.
- Reject an unsupported client mechanism when the client authenticator is created instead of during connection initialization.
- Add focused coverage for mechanism selection, case normalization, independent authenticator instances, and unsupported mechanisms.

### Tests

- `./mvnw -pl fluss-common -DskipITs -Dtest=SaslAuthenticatorFactoryTest,PlainSaslServerTest clean test`
  - 7 tests passed.
  - Checkstyle and Spotless passed.
- `git diff --check`

### API and Format

No public API, RPC wire format, or storage format changes.

The new factory is internal. No new configuration options are introduced.

### Documentation

No documentation changes are required because this PR does not expose a new authentication mechanism or configuration.

### DIFF
--- a/fluss-common/src/main/java/org/apache/fluss/security/auth/sasl/SaslAuthenticatorFactory.java
+++ b/fluss-common/src/main/java/org/apache/fluss/security/auth/sasl/SaslAuthenticatorFactory.java
@@ -1,0 +1,80 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.fluss.security.auth.sasl;
+
+import org.apache.fluss.annotation.Internal;
+import org.apache.fluss.config.ConfigOptions;
+import org.apache.fluss.config.Configuration;
+import org.apache.fluss.exception.AuthenticationException;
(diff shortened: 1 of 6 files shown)


litiliu:
@loserwang1024  hongshun, PTAL

litiliu:
@affo PTAL, Thanks!

litiliu:
@polyzos  PTAL, Thanks!

fluss-common/src/main/java/org/apache/fluss/security/auth/sasl/SaslAuthenticatorFactory.java
```diff
@@ -0,0 +1,80 @@
...
+import org.apache.fluss.security.auth.sasl.authenticator.PlainSaslServerAuthenticator;
+import org.apache.fluss.security.auth.sasl.authenticator.SaslClientAuthenticator;
+import org.apache.fluss.security.auth.sasl.plain.PlainSaslServer;
+
+import java.util.Locale;
+
+/** Factory for creating connection-local SASL authenticators. */
+@Internal
+public final class SaslAuthenticatorFactory {
+    private SaslAuthenticatorFactory() {}
```
affo, line 51: I see here that adding a new mechanism would imply editing 3 `switch` statements.
Can we make it better and store a map `mechanism -> (clientAuth, serverAuth)`?
The result would be that one would imply add an entry in the map and the implementation of the methods below would simply test the maps and would not need changes.

affo (commented):
Amazing job! Left a comment just for cleaning up the factory.

affo (approved):
Thank you!

litiliu:
@fresh-borzoni Would you mind taking a look at this PR, when you have time, Thanks!

PR #3870, opened 2026-08-05
[filesystem] Add JuiceFS filesystem support

### LINKED ISSUE #3866: [filesystem] Add JuiceFS filesystem support
### Search before asking

### Motivation

[JuiceFS](https://github.com/juicedata/juicefs) is a high-performance distributed file system designed for cloud-native scenarios, which can be connected to various object storage services at the underlying layer. Some enterprises adopt JuiceFS as a unified access layer for various object storage services. It ships with a Hadoop-compatible Java SDK, making it well-suited to serve as remote storage for snapshots of primary-key tables and tiered log segments of log tables within Fluss. 

### Solution

Add a new filesystem plugin module `fluss-fs-juicefs` under `fluss-filesystems/`, registered under the `jfs://` URI scheme, bridging Fluss to the JuiceFS Hadoop SDK.

* **Module**: `fluss-filesystems/fluss-fs-juicefs`                                                                                                                                                                                                       
* **Scheme**: `jfs://`                                                                                                                                                                                                                                   
* **Underlying SDK**: `io.juicefs:juicefs-hadoop`                                                                                                                                                      
* **Distribution**: Not bundled in the default binary distribution; installed manually into `${FLUSS_HOME}/plugins/juicefs/`, consistent with the COS / OBS / Azure plugins.

## Design / Implementation Overview

The plugin is intentionally thin and reuses `HadoopFileSystem` from `fluss-fs-hadoop`, mirroring the structure of the existing OSS / COS plugins.   

1. **`JuiceFsPlugin implements FileSystemPlugin`**
* `getScheme()` returns `"jfs"`.
* `create(URI, Configuration)` builds a Hadoop `Configuration` from Fluss config, applies JuiceFS defaults, and instantiates the JuiceFS Hadoop `FileSystem` via reflection (`FileSystem.newInstance`) so that Fluss has no hard compile-time dependency on `io.juicefs.*`.
* Forwards any Fluss configuration entry whose key starts with `fs.jfs.` or `juicefs.` to the Hadoop `Configuration`.
* Injects sensible defaults when the user does not set them:
  * `fs.jfs.impl = io.juicefs.JuiceFileSystem`
  * `fs.jfs.impl.disable.cache = false`

2. **`JuiceFsFileSystem extends HadoopFileSystem`**
* Wraps the JuiceFS Hadoop `FileSystem`.
* Overrides `obtainSecurityToken()` to return an **empty placeholder token**, because JuiceFS handles authentication locally (see below).

3. **SPI registration** via `META-INF/services/org.apache.fluss.fs.FileSystemPlugin`.

4. **Shading / packaging** consistent with the other filesystem plugins (depends on `fluss-fs-hadoop-shaded`, `fluss-fs-hadoop`, and the `juicefs-hadoop` SDK).

5. **Tests**: A `JuiceFsPluginTest` covering scheme registration, prefix-based config forwarding (`fs.jfs.*`, `juicefs.*`, and unrelated keys), default-value injection (and non-override of user values), and null-config safety.

### Configuration Example

Minimum required configuration in `server.yaml`:
```
# The dir that used to be as the remote storage of Fluss                                                                                                                                                                                                 
remote.data.dir: jfs://<your-volume-name>/path/to/remote/storage                                                                                                                                                                                         
                                                                                                                                                                                                                                                              
# JuiceFS metadata engine address of the pre-created volume, e.g. redis://<host>:<port>/<db>                                                                                                                                                                                                                        
juicefs.meta: <your-meta-url> 
```

Any other `fs.jfs.*` / `juicefs.*` keys documented in the [JuiceFS Hadoop Java SDK docs](https://juicefs.com/docs/community/hadoop_java_sdk) can be added to `server.yaml` and will be transparently forwarded, e.g.:
```                                                                                                                                                                                                                                                 
juicefs.cache-dir: /data*/jfscache                                                                                                                                                                                                                       
juicefs.cache-size: 1024                                                                                                                                                                                                                                 
juicefs.access-log: /tmp/juicefs.access.log                                                                                                                                                                                                              
```

### Authentication Model

Unlike the OSS / S3 / COS plugins, Fluss does **not** perform any STS or delegation-token exchange for JuiceFS:
* The JuiceFS client on each node authenticates directly against its metadata engine and backing object storage, using credentials embedded when the JuiceFS volume was formatted (or supplied via `juicefs.access-key` / `juicefs.secret-key` overrides where applicable). 
* **Deployment implication**: every Fluss process that accesses remote storage (CoordinatorServer, TabletServer, and every client) must have (a) the plugin installed, (b) the same `juicefs.*` configuration, and (c) network reachability to the JuiceFS metadata engine and object storage.

### References

* JuiceFS Hadoop Java SDK: https://juicefs.com/docs/community/hadoop_java_sdk
* JuiceFS GitHub: https://github.com/juicedata/juicefs

### Willingness to contribute

weizuo93:
### Purpose

Linked issue: close #3866

This PR adds support for [JuiceFS](https://juicefs.com/docs/community/introduction/) — a widely-adopted open-source distributed file system that is built on top of object storage and exposes a Hadoop-compatible Java SDK. With this change, users can point `remote.data.dir` at a `jfs://` volume and store Fluss snapshots (Primary-Key Tables) and tiered log segments (Log Tables) on their existing JuiceFS deployments

### Brief change log

* Add a new Maven module **`fluss-filesystems/fluss-fs-juicefs`**, registered as a child of `fluss-filesystems`.
* Implement `JuiceFsPlugin implements FileSystemPlugin`:
  * Registers the `jfs://` URI scheme.
  * Loads `io.juicefs.JuiceFileSystem` reflectively via `org.apache.hadoop.fs.FileSystem.newInstance`, so Fluss has no hard compile-time dependency on `io.juicefs.*` classes. 
  * Forwards any Fluss configuration entry whose key starts with `fs.jfs.` or `juicefs.` to the underlying Hadoop `Configuration`.
  * Injects sensible defaults when not set by the user:
    * `fs.jfs.impl = io.juicefs.JuiceFileSystem`
    * `fs.jfs.impl.disable.cache = false`
* Implement `JuiceFsFileSystem extends HadoopFileSystem`:
  * Wraps the JuiceFS Hadoop `FileSystem`.
  * Returns an **empty placeholder security token** from `obtainSecurityToken()`, because JuiceFS handles authentication locally on each node against its metadata engine and backing object storage — no STS / delegation-token exchange is required.
* SPI registration via `META-INF/services/org.apache.fluss.fs.FileSystemPlugin`.
* Shading / packaging consistent with sibling filesystem plugins; depends on `fluss-fs-hadoop-shaded`, `fluss-fs-hadoop`, and `io.juicefs:juicefs-hadoop:1.4.0`.
* Documentation:
  * New page `website/docs/maintenance/tiered-storage/filesystems/juicefs.md`.
  * Add JuiceFS to the `overview.md` list of pluggable filesystems.

### Tests

Unit tests added in `JuiceFsPluginTest`:
* `testScheme` — verifies the plugin registers the `jfs` scheme.
* `testHadoopConfigPropagation` — verifies `fs.jfs.*` and `juicefs.*` entries are forwarded to the Hadoop `Configuration`, and that unrelated keys are **not** forwarded.
* `testApplyJuiceFsDefaultsInjectsImpl` — verifies default injection of `fs.jfs.impl` and `fs.jfs.impl.disable.cache`.
* `testApplyJuiceFsDefaultsDoesNotOverrideUserValues` — verifies user-supplied values take precedence over the injected defaults.
* `testGetHadoopConfigurationWithNullFlussConfig` — null-safety.

Run locally with:
```
mvn test -Dtest=JuiceFsPluginTest -DfailIfNoTests=false -pl fluss-filesystems/fluss-fs-juicefs -am
```

### API and Format

No changes to public APIs, wire format, or on-disk / storage format. The change is purely additive — a new implementation of the existing `FileSystemPlugin` SPI.

### Documentation

* New page: [`website/docs/maintenance/tiered-storage/filesystems/juicefs.md`](website/docs/maintenance/tiered-storage/filesystems/juicefs.md) — installation, prerequisite (JuiceFS volume creation), required / optional `server.yaml` configuration, authentication model, and pointers to the [JuiceFS Hadoop Java SDK documentation](https://juicefs.com/docs/community/hadoop_java_sdk) for advanced tuning.
* Updated `website/docs/maintenance/tiered-storage/filesystems/overview.md` to include JuiceFS in the intro list and add a bullet under the "Pluggable File Systems" section.

### DIFF
--- a/fluss-filesystems/fluss-fs-juicefs/pom.xml
+++ b/fluss-filesystems/fluss-fs-juicefs/pom.xml
@@ -1,0 +1,193 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+ Licensed to the Apache Software Foundation (ASF) under one
+ or more contributor license agreements.  See the NOTICE file
+ distributed with this work for additional information
+ regarding copyright ownership.  The ASF licenses this file
+ to you under the Apache License, Version 2.0 (the
+ "License"); you may not use this file except in compliance
+ with the License.  You may obtain a copy of the License at
+
+      http://www.apache.org/licenses/LICENSE-2.0
+
+ Unless required by applicable law or agreed to in writing, software
+ distributed under the License is distributed on an "AS IS" BASIS,
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+-->
+
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+    <parent>
+        <groupId>org.apache.fluss</groupId>
+        <artifactId>fluss-filesystems</artifactId>
+        <version>1.0-SNAPSHOT</version>
+    </parent>
+
+    <artifactId>fluss-fs-juicefs</artifactId>
+    <name>Fluss : FileSystems : JuiceFS</name>
+
+    <properties>
+        <fs.juicefs.sdk.version>1.4.0</fs.juicefs.sdk.version>
+    </properties>
+
+    <dependencies>
+        <dependency>
+            <groupId>org.apache.fluss</groupId>
+            <artifactId>fluss-common</artifactId>
+            <version>${project.version}</version>
+            <scope>provided</scope>
+        </dependency>
+
+        <dependency>
+            <groupId>org.apache.fluss</groupId>
+            <artifactId>fluss-fs-hadoop-shaded</artifactId>
+            <version>${project.version}</version>
+        </dependency>
+
+        <dependency>
+            <groupId>org.apache.fluss</groupId>
+            <artifactId>fluss-fs-hadoop</artifactId>
+            <version>${project.version}</version>
+        </dependency>
+
+        <dependency>
+            <groupId>io.juicefs</groupId>
+            <artifactId>juicefs-hadoop</artifactId>
+            <version>${fs.juicefs.sdk.version}</version>
+            <exclusions>
+                <!-- provided by fluss-fs-hadoop-shaded -->
+                <exclusion>
+                    <groupId>org.apache.hadoop</groupId>
+                    <artifactId>hadoop-common</artifactId>
+                </exclusion>
+                <exclusion>
+                    <groupId>org.apache.hadoop</groupId>
+                    <artifactId>hadoop-client</artifactId>
+                </exclusion>
+                <!-- avoid logger conflicts -->
+                <exclusion>
+                    <groupId>ch.qos.reload4j</groupId>
+                    <artifactId>reload4j</artifactId>
+                </exclusion>
+                <exclusion>
+                    <groupId>org.slf4j</groupId>
+                    <artifactId>slf4j-reload4j</artifactId>
+                </exclusion>
+                <exclusion>
+                    <groupId>org.slf4j</groupId>
+                    <artifactId>slf4j-log4j12</artifactId>
+                </exclusion>
+                <exclusion>
+                    <groupId>log4j</groupId>
+                    <artifactId>log4j</artifactId>
+                </exclusion>
+                <exclusion>
+                    <groupId>commons-logging</groupId>
+                    <artifactId>commons-logging</artifactId>
+                </exclusion>
+            </exclusions>
+        </dependency>
+
+        <!-- test dependencies -->
+        <dependency>
+            <groupId>org.apache.fluss</groupId>
+            <artifactId>fluss-test-utils</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.fluss</groupId>
+            <artifactId>fluss-common</artifactId>
+            <version>${project.version}</version>
+            <scope>test</scope>
+            <type>test-jar</type>
+        </dependency>
+    </dependencies>
+
+    <build>
+        <plugins>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-shade-plugin</artifactId>
+                <executions>
+                    <execution>
+                        <id>shade-fluss</id>
+                        <phase>package</phase>
+                        <goals>
+                            <goal>shade</goal>
+                        </goals>
+                        <configuration>
+                            <artifactSet>
+                                <includes>
+                                    <include>*:*</include>
+                                </includes>
+                                <excludes>
+                                    <exclude>javax.servlet:servlet-api</exclude>
+                                    <exclude>xmlenc:xmlenc</exclude>
+                                </excludes>
+                            </artifactSet>
+                            <filters>
+                                <filter>
+                                    <artifact>*</artifact>
+                                    <excludes>
+                                        <exclude>.gitkeep</exclude>
+                                        <exclude>mime.types</exclude>
+                                        <exclude>mozilla/**</exclude>
+                                        <exclude>LICENSE.txt</exclude>
+                                        <exclude>license/LICENSE*</exclude>
+                                        <exclude>NOTICE</exclude>
+                                    </excludes>
+                                </filter>
+                                <filter>
+                                    <artifact>org.apache.fluss:fluss-fs-hadoop</artifact>
+                                    <excludes>
+                                        <exclude>META-INF/**</exclude>
+                                    </excludes>
+                                </filter>
+                                <filter>
+                                    <!--
+                                      juicefs-hadoop is an uber jar that bundles Sun Jersey 1.x
+                                      (CDDL/GPLv2 dual-licensed) and its metadata files. Fluss
+                                      does not use Jersey; strip these classes and resources to
+                                      keep the shaded artifact free of GPL-flagged content and
+                                      avoid conflicting META-INF headers from the uber jar.
+                                    -->
+                                    <artifact>io.juicefs:juicefs-hadoop</artifact>
+                                    <excludes>
+                                        <exclude>com/sun/jersey/**</exclude>
+                                        <exclude>META-INF/maven/com.sun.jersey/**</exclude>
+                                        <exclude>META-INF/taglib.tld</exclude>
+                                        <exclude>META-INF/LICENSE</exclude>
+                                        <exclude>META-INF/NOTICE.txt</exclude>
+                                        <exclude>META-INF/DEPENDENCIES</exclude>
+                                    </excludes>
+                                </filter>
+                            </filters>
+                            <relocations>
+                                <relocation>
+                                    <pattern>org.apache.commons</pattern>
+                                    <shadedPattern>org.apache.fluss.shaded.org.apache.commons</shadedPattern>
+                                </relocation>
+                            </relocations>
+                        </configuration>
+                    </execution>
+                </executions>
+            </plugin>
+
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-jar-plugin</artifactId>
+                <executions>
+                    <execution>
+                        <goals>
+                            <goal>test-jar</goal>
+                        </goals>
+                    </execution>
+                </executions>
+            </plugin>
+        </plugins>
+    </build>
+
+</project>
--- a/fluss-filesystems/fluss-fs-juicefs/src/main/java/org/apache/fluss/fs/juicefs/JuiceFsFileSystem.java
+++ b/fluss-filesystems/fluss-fs-juicefs/src/main/java/org/apache/fluss/fs/juicefs/JuiceFsFileSystem.java
@@ -1,0 +1,50 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.fluss.fs.juicefs;
+
+import org.apache.fluss.fs.hdfs.HadoopFileSystem;
+import org.apache.fluss.fs.token.ObtainedSecurityToken;
+
+import org.apache.hadoop.fs.FileSystem;
+
+import java.util.Collections;
+
+/**
+ * A {@link org.apache.fluss.fs.FileSystem} for JuiceFS that wraps a {@link HadoopFileSystem}.
+ *
+ * <p>Unlike the OSS / S3 plugins, JuiceFS does not require Fluss to obtain or distribute a
+ * delegation token: the JuiceFS client itself authenticates against the meta server using
+ * locally-configured credentials (e.g. {@code juicefs.access-key} / {@code juicefs.secret-key}, or
+ * implicit IAM). We therefore return an empty placeholder token from {@link
+ * #obtainSecurityToken()}.
+ */
+class JuiceFsFileSystem extends HadoopFileSystem {
+
+    private static final ObtainedSecurityToken EMPTY_TOKEN =
+            new ObtainedSecurityToken(
+                    JuiceFsPlugin.SCHEME, new byte[0], null, Collections.emptyMap());
+
+    JuiceFsFileSystem(FileSystem hadoopFileSystem) {
+        super(hadoopFileSystem);
+    }
+
+    @Override
+    public ObtainedSecurityToken obtainSecurityToken() {
+        return EMPTY_TOKEN;
+    }
+}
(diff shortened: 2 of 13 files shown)


fluss-filesystems/fluss-fs-juicefs/src/main/java/org/apache/fluss/fs/juicefs/JuiceFsPlugin.java
```diff
@@ -0,0 +1,151 @@
...
+                if (key.startsWith(prefix)) {
+                    String value =
+                            flussConfig.getString(
+                                    ConfigBuilder.key(key).stringType().noDefaultValue(), null);
+                    conf.set(key, value);
+
+                    LOG.debug(
+                            "Adding Fluss config entry for {} as {} to Hadoop config",
+                            key,
+                            conf.get(key));
```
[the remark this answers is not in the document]
weizuo93: Done

website/docs/maintenance/tiered-storage/filesystems/juicefs.md
```diff
@@ -0,0 +1,92 @@
...
+
+Unlike the OSS / S3 / COS plugins, Fluss does **not** perform any STS or delegation-token exchange for JuiceFS. The JuiceFS client authenticates locally on each node:
+
+- Against the **metadata engine** using the URL configured via `juicefs.meta`.
+- Against the backing **object storage** using the credentials that were provided when the volume was formatted (or via optional `juicefs.access-key` / `juicefs.secret-key` overrides, if applicable).
+
+The Fluss server returns an empty placeholder security token to clients when they request access to the remote storage. This means every process that needs to read remote data (CoordinatorServer, TabletServer, and every client) must:
+
+1. Have direct network access to the JuiceFS metadata engine and object storage.
+2. Have the JuiceFS plugin installed and the same `juicefs.*` configuration available.
```
luoyuxia: Could we document the client-side setup separately here? `FlussConnection` only forwards options with the `client.fs.*` prefix and calls `FileSystem.initialize(..., null)`, so it neither forwards bare `juicefs.*` options nor loads plugins from the plugin directory. A client that reads tiered files needs the JuiceFS plugin JAR on its application classpath and options such as `client.fs.juicefs.meta`; following only the current `server.yaml` and `${FLUSS_HOME}/plugins/juicefs/` instructions can result in an unsupported `jfs` scheme or a missing metadata URL.

fluss-filesystems/fluss-fs-juicefs/src/main/java/org/apache/fluss/fs/juicefs/JuiceFsPlugin.java
```diff
@@ -0,0 +1,172 @@
...
+            for (String prefix : FLUSS_CONFIG_PREFIXES) {
+                if (key.startsWith(prefix)) {
+                    String value =
+                            flussConfig.getString(
+                                    ConfigBuilder.key(key).stringType().noDefaultValue(), null);
+                    conf.set(key, value);
+                    LOG.debug(
+                            "Adding Fluss config entry for {} as {} to Hadoop config",
+                            key,
+                            isSensitive(key) ? HIDDEN_CONTENT : conf.get(key));
```
luoyuxia: Could we avoid logging configuration values entirely? The current key-based mask still exposes credential-bearing values whose keys do not contain `password` or `secret`: for example, `juicefs.meta` may be `redis://user:password@host/...`, and keys such as `juicefs.token`, `juicefs.access-key`, and `juicefs.push-auth` are also not masked. Logging only the forwarded key should provide enough debugging information without risking credential leakage.

fluss-filesystems/fluss-fs-juicefs/src/main/java/org/apache/fluss/fs/juicefs/JuiceFsFileSystem.java
```diff
@@ -0,0 +1,50 @@
...
+            new ObtainedSecurityToken(
+                    JuiceFsPlugin.SCHEME, new byte[0], null, Collections.emptyMap());
+
+    JuiceFsFileSystem(FileSystem hadoopFileSystem) {
+        super(hadoopFileSystem);
+    }
+
+    @Override
+    public ObtainedSecurityToken obtainSecurityToken() {
+        return EMPTY_TOKEN;
```
luoyuxia, line 48: `obtainSecurityToken()` returns a non-null token with scheme `jfs`, so `DefaultSecurityTokenManager` always forwards it to `SecurityTokenReceiverRepository`. This module does not register a `jfs` receiver, causing every client to log `no receiver found` and schedule another token refresh after the retry backoff. Could we add a no-op receiver and SPI entry, similar to HDFS, or explicitly skip receiver notification for placeholder tokens? A small regression test for this path would also help.

luoyuxia (commented):
@weizuo93 Thanks for the PR! I left a few minor comments. PTAL.

weizuo93:
@luoyuxia  All comments have been addressed. PTAL again.

website/docs/maintenance/tiered-storage/filesystems/juicefs.md
weizuo93: Done

fluss-filesystems/fluss-fs-juicefs/src/main/java/org/apache/fluss/fs/juicefs/JuiceFsPlugin.java
weizuo93: Done

fluss-filesystems/fluss-fs-juicefs/src/main/java/org/apache/fluss/fs/juicefs/JuiceFsFileSystem.java
weizuo93: OK

fluss-filesystems/fluss-fs-juicefs/pom.xml
```diff
@@ -0,0 +1,193 @@
...
+                                </filter>
+                                <filter>
+                                    <!--
+                                      juicefs-hadoop is an uber jar that bundles Sun Jersey 1.x
+                                      (CDDL/GPLv2 dual-licensed) and its metadata files. Fluss
+                                      does not use Jersey; strip these classes and resources to
+                                      keep the shaded artifact free of GPL-flagged content and
+                                      avoid conflicting META-INF headers from the uber jar.
+                                    -->
+                                    <artifact>io.juicefs:juicefs-hadoop</artifact>
```
luoyuxia: **Blocking:** `juicefs-hadoop:1.4.0` is already an uber JAR. The published artifact contains `io/juicefs/shaded/org/json/**`, and its embedded Maven metadata identifies the code as `org.json:json:20180813`, whose JSON License contains the “Good, not Evil” restriction. [JuiceFS v1.4.0 explicitly depends on and relocates that version](https://github.com/juicedata/juicefs/blob/v1.4.0/sdk/java/pom.xml#L77-L78). The [ASF third-party license policy](https://www.apache.org/legal/resolved.html#category-x) classifies the pre-20220924 JSON License as Category X and prohibits distributing it in ASF products.

This filter removes Jersey only, so the shaded Fluss plugin will still redistribute those JSON classes. RAT/license CI does not catch this because the upstream uber JAR publishes a dependency-reduced POM. Please use or rebuild a compliant SDK and re-audit the embedded dependencies/NOTICE before publishing this module. Simply excluding `io/juicefs/shaded/org/json/**` is not sufficient because `JuiceFileSystemImpl` uses `JSONObject` at runtime.

luoyuxia, line 157: **Blocking:** `juicefs-hadoop:1.4.0` is already an uber JAR. The published artifact contains `io/juicefs/shaded/org/json/**`, and its embedded Maven metadata identifies the code as `org.json:json:20180813`, whose JSON License contains the “Good, not Evil” restriction. [JuiceFS v1.4.0 explicitly depends on and relocates that version](https://github.com/juicedata/juicefs/blob/v1.4.0/sdk/java/pom.xml#L77-L78). The [ASF third-party license policy](https://www.apache.org/legal/resolved.html#category-x) classifies the pre-20220924 JSON License as Category X and prohibits distributing it in ASF products.

This filter removes Jersey only, so the shaded Fluss plugin will still redistribute those JSON classes. RAT/license CI does not catch this because the upstream uber JAR publishes a dependency-reduced POM. Please use or rebuild a compliant SDK and re-audit the embedded dependencies/NOTICE before publishing this module. Simply excluding `io/juicefs/shaded/org/json/**` is not sufficient because `JuiceFileSystemImpl` uses `JSONObject` at runtime.

website/docs/maintenance/tiered-storage/filesystems/juicefs.md
```diff
@@ -0,0 +1,151 @@
...
+
+[JuiceFS](https://juicefs.com) is a distributed POSIX-compatible file system built on top of object storage and a separate metadata engine (Redis, TiKV, MySQL, PostgreSQL, etc.). It ships with a Hadoop-compatible Java SDK, which Fluss uses to store snapshots for Primary-Key Tables and tiered log segments for Log Tables on a JuiceFS volume.
+
+## Install JuiceFS Plugin on Server Nodes
+
+JuiceFS support is not included in the default Fluss distribution. To enable JuiceFS support on the server side (CoordinatorServer and every TabletServer), you need to manually install the filesystem plugin into Fluss.
+
+1. **Prepare the plugin JAR**:
+
+   - Download the `fluss-fs-juicefs-$FLUSS_VERSION$.jar` from the [Maven Repository](https://repo1.maven.org/maven2/org/apache/fluss/fluss-fs-juicefs/$FLUSS_VERSION$/fluss-fs-juicefs-$FLUSS_VERSION$.jar).
```
luoyuxia, line 16: **Documentation:** The published plugin currently supports Linux x86_64 only. `juicefs-hadoop:1.4.0` bundles only `libjfs-amd64.so.gz`; on Linux ARM64, macOS, and Windows, `JuiceFileSystemImpl` selects different native resource names that are not present in the JAR, so initialization fails at runtime.

Please explicitly document the Linux x86_64 support restriction here (and in the client-side setup if appropriate) so users on other platforms do not install an unusable plugin.

website/docs/maintenance/tiered-storage/filesystems/overview.md
```diff
@@ -42,4 +42,6 @@ The Fluss project supports the following file systems:
 
 - **[Tencent Cloud COS](cos.md)** is supported by `fluss-fs-cos` and registered under the `cosn://` URI scheme. Please make sure to [manually install the COS plugin](cos.md#install-cos-plugin-manually).
 
+- **[JuiceFS](juicefs.md)** is supported by `fluss-fs-juicefs` and registered under the `jfs://` URI scheme. Please make sure to [manually install the JuiceFS plugin](juicefs.md#install-juicefs-plugin-manually).
```
luoyuxia, line 45: **Minor:** This link points to `#install-juicefs-plugin-manually`, but the target heading is `Install JuiceFS Plugin on Server Nodes`, whose generated anchor is `#install-juicefs-plugin-on-server-nodes`.

The current fragment does not navigate to the installation section. Please update the link to use the generated anchor.

website/docs/maintenance/tiered-storage/filesystems/juicefs.md
```diff
@@ -0,0 +1,151 @@
...
+sidebar_position: 8
+---
+
+# JuiceFS
+
+[JuiceFS](https://juicefs.com) is a distributed POSIX-compatible file system built on top of object storage and a separate metadata engine (Redis, TiKV, MySQL, PostgreSQL, etc.). It ships with a Hadoop-compatible Java SDK, which Fluss uses to store snapshots for Primary-Key Tables and tiered log segments for Log Tables on a JuiceFS volume.
+
+## Install JuiceFS Plugin on Server Nodes
+
+JuiceFS support is not included in the default Fluss distribution. To enable JuiceFS support on the server side (CoordinatorServer and every TabletServer), you need to manually install the filesystem plugin into Fluss.
```
luoyuxia, line 12: **Question (Java 17):** The [JuiceFS Hadoop SDK requirements](https://juicefs.com/docs/community/hadoop_java_sdk/#5-java-runtime-version) state that higher Java runtimes such as Java 17 need `--add-exports=java.base/sun.nio.ch=ALL-UNNAMED`. Fluss recommends Java 17+, but this guide does not mention this option for the server or client-side processes.

Could you confirm whether the packaged `juicefs-hadoop:1.4.0` has been verified to work on Java 17 without this option? If it is required, please document where to configure it for CoordinatorServer/TabletServer, Flink JobManager/TaskManager, and Spark driver/executor. If it is not required for this integration, confirmation or test evidence would be sufficient.

website/docs/maintenance/tiered-storage/filesystems/juicefs.md
```diff
@@ -0,0 +1,151 @@
...
+
+**Flink SQL / Table API.** Connector options declared in `WITH (...)` are propagated into the underlying client `Configuration`, so JuiceFS options can be set per-table:
+
+```sql
+CREATE TABLE fluss_table (
+  ...
+) WITH (
+  'connector' = 'fluss',
+  'bootstrap.servers' = '<coordinator-host>:<port>',
+  'client.fs.juicefs.meta' = 'redis://<host>:<port>/<db>',
```
luoyuxia, line 125: [MAJOR] Please avoid persisting JuiceFS credentials in table options.

`client.fs.juicefs.meta` may contain metadata-store credentials, and the section above also recommends `client.fs.juicefs.access-key` / `client.fs.juicefs.secret-key`. When a persistent table is created through `FlinkCatalog`, these options are copied into `TableDescriptor.customProperties`, stored in Fluss metadata, and exposed again when the table is read. This is unsafe and also difficult to operate: credentials are duplicated per table, and rotation requires updating every table definition.

OSS/S3 keep the filesystem initialization credentials on the server side and use the existing filesystem security-token channel to provide client-side authentication before the actual filesystem instance is created. Could JuiceFS follow the same model? Ideally, the server obtains and returns a JuiceFS delegation token. If JuiceFS cannot issue one, a simpler fallback would be to return the minimal client filesystem bootstrap/authentication information from the server through `ObtainedSecurityToken` and install it in `JuiceFsSecurityTokenReceiver`, rather than persisting it in the table definition.

luoyuxia (commented):
@weizuo93 Thanks for the pr. I left some comments again. PTAL

PR #3873, opened 2026-08-05
[flink] Complete bounded log splits after filtered batches

### LINKED ISSUE #3872: [flink] Batch Log Table query does not finish when column statistics filter out all records
Thank you very much for your feedback!

## Search before asking

## Fluss version

`main (development)`

### What happened

When a Flink Batch job reads a Fluss Log Table with column statistics enabled, and the query predicate filters out all records in a bucket, the query returns no records but the Flink job remains in the `RUNNING` state.

The bounded batch read does not complete.

### What I expected

The query should return an empty result, and the Flink Batch job should finish normally.

### Steps to reproduce

Start Fluss and Flink, then run the following scenario:

1. Configure Flink to use batch mode with low parallelism:

   ```sql
   SET 'execution.runtime-mode' = 'batch';
   SET 'parallelism.default' = '1';
   ```

2. Create a Log Table with two buckets and statistics enabled for `amount`:

   ```sql
   CREATE TABLE stats_log (
       order_id BIGINT,
       amount INT
   ) WITH (
       'bucket.num' = '2',
       'table.statistics.columns' = 'amount'
   );
   ```

3. Write two records with positive values and ensure that the records are distributed across the two buckets, for example:

   ```text
   amount = 10
   amount = 20
   ```

4. Run a query that matches no records:

   ```sql
   SELECT *
   FROM stats_log
   WHERE amount < 0;
   ```

5. Observe the result and the Flink job state.

6. Remove the `table.statistics.columns` configuration and repeat the same query with the same data as a control test.

### Additional context

With `table.statistics.columns = 'amount'` enabled, the query returns no records, but the Flink job remains in the `RUNNING` state. Without this configuration, the same query returns `Empty set` and finishes normally.

Column statistics allow Fluss to skip batches that cannot match the predicate, returning no records while still advancing the consumed offset to the end. The Flink source reader checks split completion only from the last materialized record and ignores this progress-only result, so the split is never completed.

## Solution

Use the consumed offset progress to determine whether a bounded split has reached its stopping offset, including when the scan returns no materialized records.

Gezi-lzq:
### Purpose

Linked issue: close #3872

Flink bounded Log Table batch scans can fail to finish when statistics filtering materializes no records even though the scanner has consumed the split through its stopping offset.

### Brief change log

- Use scanner consumed progress to mark bounded log splits finished when a poll contains no materialized records.
- Keep the existing last-record offset fallback for results without consumed progress.
- Add a single-bucket statistics-filtered batch regression test.

### Tests

- `FlinkSourceSplitReaderTest`: 7/7 passed.
- `Flink120TableSourceBatchITCase#testFilteredLogTableBatchScanCompletesWhenNoRecordsMatch`: 1/1 passed.
- Spotless and Checkstyle checks passed for the affected modules.

### API and Format

### Documentation

No documentation changes required.

### Generative AI disclosure

- [x] Yes (OpenAI Codex)

### DIFF
--- a/fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/source/reader/FlinkSourceSplitReader.java
+++ b/fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/source/reader/FlinkSourceSplitReader.java
@@ -468,6 +468,11 @@ public class FlinkSourceSplitReader implements SplitReader<RecordAndPos, SourceS
             splitIdByTableBucket.put(scanBucket, splitId);
             tableScanBuckets.add(scanBucket);
             List<ScanRecord> bucketScanRecords = scanRecords.records(scanBucket);
+            Long consumedUpToOffset = scanRecords.consumedUpToOffset(scanBucket);
+            if (consumedUpToOffset != null && consumedUpToOffset >= stoppingOffset) {
+                stoppingOffsets.put(scanBucket, stoppingOffset);
+                finishedSplits.add(splitId);
+            }
             if (!bucketScanRecords.isEmpty()) {
                 final ScanRecord lastRecord = bucketScanRecords.get(bucketScanRecords.size() - 1);
                 // We keep the maximum message timestamp in the fetch for calculating lags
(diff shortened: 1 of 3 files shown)


Gezi-lzq:
Hi @loserwang1024, when you have time, could you please take a look at this PR? Thanks!

Gezi-lzq:
Fixed by PR #4007, which has now been merged. This PR can be closed.

PR #3882, opened 2026-08-06
[server] remove multi-key partitions from cache correctly

zuston:
### Purpose

Fix incorrect Auto Partition cache removal for tables with multiple partition keys.

### Brief change log

- Fixed partition removal from the nested cache structure.
- Removed empty time-based cache entries.
- Added a regression test for multi-key partition removal.
- Resolved the test file merge conflict while preserving existing coverage.

### DIFF
--- a/fluss-server/src/main/java/org/apache/fluss/server/coordinator/AutoPartitionManager.java
+++ b/fluss-server/src/main/java/org/apache/fluss/server/coordinator/AutoPartitionManager.java
@@ -354,21 +354,10 @@ public class AutoPartitionManager implements AutoCloseable {
                 lock,
                 () -> {
                     if (autoPartitionTables.containsKey(tableId)) {
-                        TableInfo tableInfo = autoPartitionTables.get(tableId);
-                        TreeMap<String, Set<String>> partitionMap = partitionsByTable.get(tableId);
-                        if (tableInfo.getPartitionKeys().size() > 1) {
-                            String autoPartitionValue =
-                                    extractAutoPartitionValue(tableInfo, partitionName);
-                            Set<String> partitionSet = partitionMap.get(autoPartitionValue);
-                            if (partitionSet != null) {
-                                partitionSet.remove(partitionName);
-                                if (partitionSet.isEmpty()) {
-                                    partitionMap.remove(autoPartitionValue);
-                                }
-                            }
-                        } else {
-                            partitionMap.remove(partitionName);
-                        }
+                        removePartitionFromPartitionsByTable(
+                                autoPartitionTables.get(tableId),
+                                partitionsByTable.get(tableId),
+                                partitionName);
                     }
                 });
     }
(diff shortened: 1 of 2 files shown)


fluss-server/src/main/java/org/apache/fluss/server/coordinator/AutoPartitionManager.java
```diff
@@ -673,6 +687,30 @@ protected Integer getAutoCreateDayDelayMinutes(long tableId) {
...
+                    if (partitionMap == null) {
+                        return null;
+                    }
+
+                    Map<String, Set<String>> snapshot = new TreeMap<>();
+                    partitionMap.forEach(
+                            (partitionTime, partitionSet) ->
+                                    snapshot.put(
+                                            partitionTime,
+                                            partitionSet == null
```
Jackeyzhe, line 714: Nice fix, nit: could you add a test covering `partitionSet == null` branch?

zuston:
cc @swuferhong

fluss-server/src/main/java/org/apache/fluss/server/coordinator/AutoPartitionManager.java
zuston: make sense

fluss-server/src/main/java/org/apache/fluss/server/coordinator/AutoPartitionManager.java
```diff
@@ -354,7 +354,21 @@ public void removePartition(long tableId, String partitionName) {
                 lock,
                 () -> {
                     if (autoPartitionTables.containsKey(tableId)) {
```
fresh-borzoni, line 356: addPartition just above calls addPartitionToPartitionsByTable. This does the inverse, but inline, so the "one partition key vs many" check now lives in three places: here, extractAutoPartitionValue, and addPartitionToPartitionsByTable. 
Can we add a removePartitionFromPartitionsByTable helper next to the add one so the two sides stay in sync?

fluss-server/src/test/java/org/apache/fluss/server/coordinator/AutoPartitionManagerTest.java
```diff
@@ -677,6 +677,41 @@ void testDayPartitionDropShouldNotBeDelayedByJitter() throws Exception {
         assertThat(partitions.keySet()).contains("20250420", "20250421", "20250422");
     }
 
+    @Test
+    void testRemovePartitionFromMultiplePartitionKeysTable() throws Exception {
```
fresh-borzoni: This table uses num-retention = 2, and with retention on the stale entry gets cleaned up by the next retention pass regardless. The case where it's never cleaned is num-retention = -1, where dropPartitions returns before touching the map. Could we cover that one too?

fresh-borzoni (commented):
@zuston Thank you for the PR, left a couple of comments, PTAL

fluss-server/src/main/java/org/apache/fluss/server/coordinator/AutoPartitionManager.java
zuston: nice suggestion!

fluss-server/src/test/java/org/apache/fluss/server/coordinator/AutoPartitionManagerTest.java
fresh-borzoni: @zuston Sorry, take back my -1 ask. It's not something a test can check and as written it does nothing anyway, since the test never triggers the scheduler. 
Let's put it back to @Test, please

fresh-borzoni (commented):
@zuston Thank you, LGTM 👍 

cc @swuferhong

fluss-server/src/test/java/org/apache/fluss/server/coordinator/AutoPartitionManagerTest.java
zuston: +1. removed

fresh-borzoni (approved):
@zuston Thank you, LGTM 👍

fresh-borzoni:
@zuston Thanks for the reminder, merging

PR #3883, opened 2026-08-06
[spark] Support time-range incremental batch reads

### LINKED ISSUE #3842: [Spark] Support time-range (incremental) batch reads for log and primary key tables
### Search before asking

### Motivation

Downstream pipelines need to **incrementally** build dimension tables and fact tables from data landing in Fluss, on a periodic schedule. A concrete case: an hourly Spark batch job that reads **only the rows written to Fluss in the past hour** and feeds that delta into the downstream computation — instead of rescanning the whole table every cycle.

Today the Spark connector's batch read always returns the **full** table, so there is no way to fetch a time-bounded increment. This forces full recomputation each cycle and makes incremental downstream processing impractical (slow and costly).

This issue proposes a **time-range batch read**: given a `[t1, t2)` window (e.g. `[now-1h, now)`), return the data written to Fluss within that window — for log tables, primary key tables, and lake-enabled tables — so downstream dimension/fact tables can be computed incrementally.

### Solution

Add timestamp-based bounds to the Spark batch scan, named consistently with the Flink connector:

- **Start:** `scan.startup.mode=timestamp` + `scan.startup.timestamp` (inclusive).
- **End:** `scan.bounded.mode=timestamp` + `scan.bounded.timestamp` (exclusive; defaults to the latest committed data at planning time).
- **Window semantics:** left-closed / right-open `[t1, t2)` on the record commit timestamp.
- **Output:** log tables return the raw records in the window; primary key tables return the keys inserted/updated in the window, folded to their latest value (deleted keys excluded).
- **Retention (Phase 1):** reads only data still retained in Fluss (bounded by `table.log.ttl`); lake-enabled tables read from Fluss only. If the start precedes retained data, the read **fails fast** by default; `scan.startup.timestamp.out-of-range=adjust` clamps to the earliest retained data instead.
- Default behavior is unchanged (`scan.startup.mode=full`).

### Anything else?

Follow-up (**Phase 2**, out of scope here): for lake-enabled primary key tables, combine the lake changelog with the Fluss changelog to extend the readable window beyond `table.log.ttl`.

### Willingness to contribute

Yohahaha:
## Summary

Support time-range incremental batch reads in the Spark connector, so pipelines can read only the data written within a `[start, end)` window (start inclusive, end exclusive) — the building block for an hourly/daily incremental job.

**New TVF** `fluss_incremental_between_timestamp(table, start[, end])`, registered via `FlussSparkSessionExtensions`:

```sql
-- the past hour, computed in SQL
SELECT * FROM fluss_incremental_between_timestamp(
  'log_table',
  date_format(now() - INTERVAL 1 HOUR, 'yyyy-MM-dd HH:mm:ss'));
```

The table argument accepts `table`, `database.table` or `catalog.database.table`. The bound arguments accept a string (epoch millis or `yyyy-MM-dd HH:mm:ss`), an integral epoch-millis value, a `DATE` (start of that day) or a `TIMESTAMP`/`TIMESTAMP_NTZ`, all in the Spark session time zone, and may come from any constant expression (column references are rejected). **Both bounds are resolved during analysis** — omitting `end` fills in Spark's `current_timestamp()` — so the window is pinned by the statement: rows committed while the query is planned stay out of it, and re-executing the same relation reads the same window.

**New options** backing the TVF, also usable from the DataFrame API. They are per-query read options and deliberately **not** read from session configuration, so a window can never leak into later reads; they are batch-only and ignored by streaming:

| Option | Default | Meaning |
|--------|---------|---------|
| `scan.incremental.start.timestamp` | (none) | Enables the incremental read; inclusive lower bound. |
| `scan.incremental.end.timestamp` | (none) | Exclusive upper bound; unset means "up to the latest committed data". |

**Read semantics per table type:**
- Log table: the records appended within the window.
- Primary key table: the keys inserted/updated in the window, folded to their latest value as of the window end (changelog range only, no kv snapshot; keys only deleted in the window are excluded).
- Lake-enabled table: same as above, but always read from Fluss only — an incremental read never unions the lake snapshot.

**Exactness of the window.** The start bound positions the scan through the server's timestamp-to-offset lookup, but the scan stops at the latest offset and **both** bounds are then applied to each record's commit timestamp by the reader (`FlussTimeRange`, carried on the input partition), terminating the partition at the first record at or after `end`. That keeps `[start, end)` exact even for data already tiered to remote storage, where a timestamp lookup is only as accurate as the sparse server-side time index — and it avoids sending a driver-computed "now" to the server, which would be rejected as a future timestamp under clock skew.

**Retention.** A window reaching further back than what Fluss still retains (`table.log.ttl`) is not an error: the dropped part simply yields fewer rows, or none. The result is always a genuine subset of the requested window, so a caller never sees foreign data — there is no separate out-of-range mode to configure, and no extra `listOffsets(EARLIEST)` round trip at planning time.

**Invalid windows fail fast** instead of silently changing semantics: a blank or unparseable timestamp, an end without a start, and (in the TVF, at analysis time) a start that is not strictly before the end.

**Also:**
- `scan.startup.mode` is documented as affecting streaming reads only; plain batch reads remain full-table, so default behavior is unchanged.
- Fixes an unrelated leak where `SET spark.sql.fluss.*` values were merged into the catalog-wide configuration shared by every table, instead of a per-scan copy.
- The scan description now reports the resolved window, e.g. `FlussScan: [fluss.t], Type: [Append] [TimeRange: [1767225600000, 1767312000000)]`.
- Docs: `website/docs/engine-spark/reads.md` (new "Time-Range Batch Read" section) and `options.md`.

Fixes #3842

## Test Plan
- [x] Unit tests `FlussOffsetInitializersTest`: option gating, blank/malformed values, end-without-start, and the time range handed to the reader.
- [x] Integration tests `SparkTimeRangeTvfTest` (partitioned tables by default): log-table window incl. the pinned upper bound, primary key changelog folding, non-partitioned log/pk tables, equivalence of all timestamp argument forms, invalid usage, and that bounds are never taken from session configuration.
- [x] Integration tests `SparkLakeTimeRangeReadTest`: log and pk tables skip the lake/kv snapshot.
- [x] `mvn spotless:check` passes on the affected modules.


Yohahaha:
@YannByron @fresh-borzoni @luoyuxia @beryllw PTAL, this is an actual customer requirement.

fluss-spark/fluss-spark-common/src/main/scala/org/apache/fluss/spark/read/FlussOffsetInitializers.scala
```diff
@@ -18,39 +18,179 @@
...
+
+  /**
+   * Whether a start timestamp preceding the earliest retained data fails fast (default) instead of
+   * being clamped to that offset. Controlled by `scan.incremental.timestamp.out-of-range`.
+   */
+  def failOnTimestampOutOfRange(options: CaseInsensitiveStringMap): Boolean = {
+    val mode =
+      incrementalOption(
+        options,
+        SparkFlussConf.SCAN_INCREMENTAL_TIMESTAMP_OUT_OF_RANGE).get.toUpperCase
```
YannByron: **Must fix:** this can break plain (non-incremental) batch reads, and the failure modes are unfriendly.

Two problems in this method:

1. `.get` on the `Option`. `incrementalOption` filters blank values, so `scan.incremental.timestamp.out-of-range=""` yields `None` and this throws `NoSuchElementException: None.get`.
2. `Enumeration.withName` throws `NoSuchElementException: No value found for 'WARN'` on any invalid value — not an `IllegalArgumentException`, and it never tells the user which values are legal.

What amplifies both is the call site: `failOnOutOfRange` is a strict `val` in the planner body (`SplitPlanner.scala:284` for append, `:658` for upsert), so it is evaluated on **every batch read**, including non-incremental ones. A typo in this incremental-only option therefore fails an ordinary full-table read with a bare `NoSuchElementException`, which is quite hard to trace back to the option that caused it.

Suggestion: make the call sites `lazy val` so it is only evaluated in incremental mode, use `getOrElse(option.defaultValue())` instead of `.get`, and throw an `IllegalArgumentException` naming the supported values — the same style `startOffsetsInitializer` already uses just below (L106-L113), which reads well.

fluss-spark/fluss-spark-common/src/main/scala/org/apache/fluss/spark/read/SplitPlanner.scala
```diff
@@ -657,6 +742,72 @@ class UpsertPlanner(
...
+    if (failOnOutOfRange) {
+      checkTimeRangeWithinRetention(
+        partitionName,
+        buckets,
+        startBucketOffsets.asScala,
+        bucketOffsetsRetriever)
+    }
+
+    val tableId = tableInfo.getTableId
+    buckets.map {
```
YannByron: **Must fix:** missing empty-range guard here, asymmetric with the append path.

The append planner gained exactly this guard in the same PR (L365-L371):

```scala
if (startOffset >= stopOffset) {
  // Empty range (e.g. a time-range window with no data, or an empty bucket): emit no
  // partition so the append reader is not handed an invalid [start, start) range.
  Seq.empty[InputPartition]
}
```

but here every bucket emits a partition unconditionally. Unlike `createUpsertPartitions`, which starts from the `EARLIEST_OFFSET` sentinel and so can never hit `start == stop`, this path resolves a concrete start offset — so an empty window in a bucket/partition is routinely reachable.

To be clear, it does not throw: `FlussUpsertPartitionReader`'s `logScanFinished` (`logStartingOffset >= logStoppingOffset`) short-circuits the poll loop, and `snapshotId == -1` skips the snapshot, so the task cleanly yields 0 rows. But it is not free either — `FlussPartitionReader.rowType` is a strict `val`, so merely constructing the reader forces `conn` -> `table` -> `getTableInfo`. Every empty partition therefore costs a Spark task, a Fluss `Connection` and an RPC. On a partitioned PK table that is `numPartitions x numBuckets` empty tasks even when the window only touched one partition.

For reference, `createLakeUpsertPartition` also guards (`if (!needLogSplit && !needLakeSplit) return None`), so this is the only one of the four planner paths that emits unconditionally. Mirroring the append guard would make them consistent and remove the reliance on the reader's short-circuit as the sole safety net.

Note also that the existing "empty window returns no rows" test builds a *log* table, so it exercises the append path that already has the guard — the upsert empty-range case looks uncovered.

fluss-spark/fluss-spark-common/src/main/scala/org/apache/fluss/spark/read/SplitPlanner.scala
```diff
@@ -657,6 +742,72 @@ class UpsertPlanner(
...
+      bucketId =>
+        val tableBucket = partitionId match {
+          case Some(pid) => new TableBucket(tableId, pid, bucketId)
+          case None => new TableBucket(tableId, bucketId)
+        }
+        FlussUpsertInputPartition(
+          tableBucket,
+          -1L,
```
YannByron: **Must fix:** `snapshotId = -1` combined with `read.optimized=true` silently returns zero rows.

`FlussUpsertPartitionReader` gates both of its inputs:

```scala
val logIterator = if (logScanFinished || readOptimized) {   // L194
  CloseableIterator.emptyIterator[KeyValueRow]()
} else {
  createLogChangesIterator()
}

val snapshotIterators = if (snapshotId == -1) { null } else { createSnapshotIterator() }  // L200
```

An incremental read always passes `-1` here — correctly, that is the established "no kv snapshot" sentinel. But when `read.optimized` is on, the log side is emptied as well, so `SortMergeReader` receives two empty inputs and the query returns **an empty result with no error and no warning**.

`readOptimized` comes from `flussConfig` (`FlussUpsertPartitionReader.scala:57`), i.e. the session-level `spark.sql.fluss.read.optimized`. A deployment that enables it globally for performance would get silently empty results from every incremental TVF query — arguably the worst failure mode for an incremental pipeline, since it is indistinguishable from "no data was written in this window".

`read.optimized` has no meaning for an incremental read (there is no snapshot to read optimized). Suggest either ignoring it on this path (always read the log) or failing fast with an explicit message. Either way the combination deserves a test; it is currently uncovered.

Nit while here: `-1` is now a magic number in three places (`SplitPlanner.scala:738`, this line, and `FlussUpsertPartitionReader.scala:200`). The Flink side has a named constant for it (`HybridSnapshotLogSplit.NO_SNAPSHOT_ID`); a `NO_SNAPSHOT_ID` on the `FlussUpsertInputPartition` companion object would be a cheap cleanup.

YannByron (commented):
Thanks for the PR — the feature direction makes sense and the docs are unusually thorough. I traced the PK changelog-fold semantics (`snapshotId = -1` -> reader skips the snapshot -> `SortMergeReader` drops delete rows) down through `FlussUpsertPartitionReader` and `SortMergeReader`, and it does behave as described.

I have three must-fix findings before this lands, all on the incremental upsert path plus the option parsing. Details are inline; summary:

1. `read.optimized=true` combined with an incremental PK read silently returns zero rows, with no error. This is the one I would call a genuine bug.
2. `failOnTimestampOutOfRange` uses `.get` and `Enumeration.withName`, and is evaluated eagerly for *every* batch read — so a bad value of an incremental-only option breaks plain full-table reads with a bare `NoSuchElementException`.
3. `createIncrementalUpsertPartitions` is missing the empty-range guard that the append path gained in this same PR, so empty buckets each still spin up a Spark task and a Fluss connection.

I also collected some non-blocking notes (short-circuiting the lake-snapshot probe in incremental mode, end-side handling versus the server's strict `ts > now` rejection in `Replica#getOffsetByTimestamp`, the `scan.incremental.*` naming versus Fluss's existing `scan.startup.*` vocabulary, and the `Thread.sleep`-based test timing). I left those out to keep this review focused — happy to post them separately if useful.

fluss-spark/fluss-spark-common/src/main/scala/org/apache/fluss/spark/catalyst/plans/logical/FlussTableValuedFunctions.scala
```diff
@@ -0,0 +1,225 @@
...
+            e
+          )
+      }
+    if (value == null) {
+      throw new IllegalArgumentException(s"Timestamp arguments of $fnName must not be null.")
+    }
+    evaluable.dataType match {
+      case StringType => value.toString
+      case ShortType | IntegerType | LongType => value.toString
+      case TimestampType | TimestampNTZType => (value.asInstanceOf[Long] / 1000L).toString
```
beryllw: TimestampNTZType stores wall-clock micros, so / 1000L pins it to UTC, while the string form is parsed in the session time zone — the same literal can resolve to a window shifted by the zone offset. Could we convert NTZ via the session time zone (matching the string path and the Flink connector's parseTimestamp convention), or reject it?

fluss-spark/fluss-spark-common/src/main/scala/org/apache/fluss/spark/read/FlussOffsetInitializers.scala
```diff
@@ -18,39 +18,188 @@
...
+  }
+
+  /**
+   * Reads a `scan.incremental.*` option from the scan options, falling back to its default. A blank
+   * value counts as unset, so a whitespace-only start timestamp never enables an incremental read.
+   */
+  private def incrementalOption(
+      options: CaseInsensitiveStringMap,
+      option: ConfigOption[String]): Option[String] =
+    Option(options.getOrDefault(option.key(), option.defaultValue())).filter(_.trim.nonEmpty)
```
beryllw, line 175: Since incrementalOption filters blank values, a blank start makes isIncrementalRead return false, so both planners quietly fall back to the plain batch branch — start = beginning of the table, stop = latest. The query succeeds, just reading the entire table instead of the requested window.

fluss-spark/fluss-spark-common/src/main/scala/org/apache/fluss/spark/catalyst/plans/logical/FlussTableValuedFunctions.scala
Yohahaha: Fixed in c1fc1935: NTZ arguments are now interpreted in the session time zone (`ntzMicrosToEpochMillis`), matching the string form and the Flink `parseTimestamp` convention. Added a regression test asserting an NTZ literal and the identical datetime string resolve to the same window under a non-UTC session time zone.

fluss-spark/fluss-spark-common/src/main/scala/org/apache/fluss/spark/read/FlussOffsetInitializers.scala
Yohahaha: Fixed in c1fc1935: an explicitly blank start now fails fast with a clear error instead of quietly falling back to a full-table read. While at it, two more windows now fail at planning time: an end timestamp set without a start, and start >= end when the end is an explicit timestamp. Unit and e2e tests added.

fluss-spark/fluss-spark-common/src/main/scala/org/apache/fluss/spark/catalyst/plans/logical/FlussTableValuedFunctions.scala
```diff
@@ -0,0 +1,250 @@
...
+              "unix_timestamp() are supported, references to table columns are not.",
+            e
+          )
+      }
+    if (value == null) {
+      throw new IllegalArgumentException(s"Timestamp arguments of $fnName must not be null.")
+    }
+    evaluable.dataType match {
+      case StringType => value.toString
+      case ShortType | IntegerType | LongType => value.toString
```
fresh-borzoni, line 153: fluss_incremental_between_timestamp('t', current_date() - INTERVAL 1 DAY, current_date()) fails with Unsupported timestamp argument type DateType. Can we accept DATE too?

fluss-spark/fluss-spark-common/src/main/scala/org/apache/fluss/spark/read/FlussOffsetInitializers.scala
```diff
@@ -18,39 +18,230 @@
...
+          s"earliest retained offset $earliestOffset. The requested time range exceeds Fluss " +
+          s"retention (table.log.ttl); narrow the time range or increase table.log.ttl.")
+    }
+  }
+
+  /**
+   * Whether a start timestamp preceding the earliest retained data fails fast (default) instead of
+   * being clamped to that offset. Controlled by `scan.incremental.timestamp.out-of-range`.
+   */
+  def failOnTimestampOutOfRange(options: CaseInsensitiveStringMap): Boolean = {
```
fresh-borzoni: This is options-only, and the TVF only writes start and end, so adjust is DataFrame-only, but the docs list it as supported.  

Fall back to session config?

fresh-borzoni (commented):
@Yohahaha Thank you for the PR, left some commments, PTAL

fluss-spark/fluss-spark-common/src/main/scala/org/apache/fluss/spark/catalyst/plans/logical/FlussTableValuedFunctions.scala
Yohahaha: good idea, already add DATE support.

fluss-spark/fluss-spark-common/src/main/scala/org/apache/fluss/spark/read/FlussOffsetInitializers.scala
Yohahaha: nice catch! I re-design the behavior of new TVF, only start and end is needed now.

Yohahaha:
@luoyuxia @beryllw @fresh-borzoni PTAL!

fluss-spark/fluss-spark-common/src/main/scala/org/apache/fluss/spark/read/SplitPlanner.scala
```diff
@@ -657,6 +709,77 @@ class UpsertPlanner(
...
+    val tableId = tableInfo.getTableId
+    buckets.flatMap {
+      bucketId =>
+        val tableBucket = partitionId match {
+          case Some(pid) => new TableBucket(tableId, pid, bucketId)
+          case None => new TableBucket(tableId, bucketId)
+        }
+        val startOffset = Long2long(startBucketOffsets.get(Integer.valueOf(bucketId)))
+        val stopOffset = Long2long(stoppingBucketOffsets.get(Integer.valueOf(bucketId)))
+        if (startOffset >= stopOffset) {
```
fresh-borzoni, line 781: This can't fire anymore stopOffset is always latest, so an empty window still gets a partition. 

If there's data after the window the start offset lands past the end, the reader keeps nothing, and LogChangesIterator throws NoSuchElementException on .head.

fresh-borzoni (commented):
@Yohahaha Thank you, left one comment, PTAL

fluss-spark/fluss-spark-common/src/main/scala/org/apache/fluss/spark/read/SplitPlanner.scala
Yohahaha: thank you for catching this real bug! fixed with return empty iter when log records is empty.

fluss-spark/fluss-spark-common/src/main/scala/org/apache/fluss/spark/read/FlussOffsetInitializers.scala
```diff
@@ -18,39 +18,159 @@
...
+        }
+      }
+      // A batch read stops at the latest committed data, an incremental one included: its end bound
+      // is applied by the reader on the record commit timestamp (see [[incrementalTimeRange]]).
       OffsetsInitializer.latest()
+    }
+  }
+
+  /** The `[start, end)` window of an incremental read, empty for a plain batch read. */
+  def incrementalTimeRange(options: CaseInsensitiveStringMap): Option[FlussTimeRange] = {
```
fresh-borzoni, line 115: The option description says a window whose start isn't strictly before its end fails fast, but requireValidWindow now only runs in the TVF, so spark.read.option(start > end) just returns no rows. It used to run in the planner and covered both. incrementalTimeRange parses both bounds and both planners call it, could the check go there? WDYT?

website/docs/engine-spark/reads.md
```diff
@@ -254,6 +254,92 @@ INSERT INTO fluss_order_with_lake VALUES
...
+  .option("scan.incremental.end.timestamp", "2026-01-02 00:00:00")
+  .table("fluss_catalog.fluss.log_table")
+```
+
+:::note
+The `scan.incremental.*` options are per-query read options only. Unlike the options in [Options](options.md), they are **not** picked up from session configuration (`SET spark.sql.fluss.scan.incremental...` has no effect), which keeps a time window from silently applying to later reads in the same session.
+:::
+
+:::warning Retention boundary
+A time-range read returns the data Fluss still retains, which is bounded by `table.log.ttl` (default 7 days). A window reaching further back is **not** an error: the part that has already been dropped simply yields fewer rows, or none at all, so the result is always a subset of the requested window. Increase `table.log.ttl` if you need to read further back. Data available only in tiered lake storage is not read by this mode.
```
fresh-borzoni, line 334: Q: Spark's own Kafka source fails by default when retention has dropped the requested offsets, with false as the explicit opt-out. 

example - job that drifts past table.log.ttl (paused 10 days, ttl 7) reads a subset and reports success, and the operator is totally unaware. Agreed the earliest-offset RPC cost more than it bought but would a flag on the ListOffsets response be a cheap way to keep the signal?

fresh-borzoni (commented):
@Yohahaha Thank you, LGTM overall, only minor comments left

fluss-spark/fluss-spark-common/src/main/scala/org/apache/fluss/spark/read/FlussOffsetInitializers.scala
Yohahaha: thanks for feedback, move valid check to common path

website/docs/engine-spark/reads.md
Yohahaha: table.log.ttl is already available to the planner via tableInfo.getTableConfig.getLogTTLMs, so the planner can compare the requested window start against ttl and log a warning — zero RPC, zero protocol change, Spark-only. It over-reports (TTL is a lower bound on retention), but for a warning that's the safer direction. I'll add that here.

we can add config to throw when found retention in following pr if needed.

fresh-borzoni (commented):
@Yohahaha Thank you, LGTM 👍 

@YannByron @beryllw cc 
Do you have any other questions from your side?

fresh-borzoni (approved):

Yohahaha:
@wuchong @luoyuxia @beryllw If there are no more issues, can we merge this PR?

PR #3884, opened 2026-08-06
[FLUSS-3553][build] Fix shaded Jackson classes leaking into uber-jars

### LINKED ISSUE #3553: [fluss-client][ fluss-flink-*] Uber-jars  leak  com.fasterxml.jackson.core  alongside the shaded copy
### Search before asking

### Fluss version

main (development)

### Please describe the bug 🐞

The published fluss-client and fluss-flink-* uber-jars each ship jackson-core twice:

  1. correctly relocated under org/apache/fluss/shaded/jackson2/com/fasterxml/jackson/core/…  and
  2. an unrelocated copy at the root com/fasterxml/jackson/core/… 

The unrelocated copy is an old jackson-core (2.15.x). Because it sits com.fasterxml.jackson.core package, it shadows the jackson-core a downstream application puts on its  classpath. Any app using a newer jackson (e.g. 2.18.x) then breaks at runtime  in our case YAML config loading:

```
Exception in thread "main" java.lang.NoSuchMethodError: 'com.fasterxml.jackson.core.JsonToken com.fasterxml.jackson.dataformat.yaml.YAMLParser._updateToken(com.fasterxml.jackson.core.JsonToken)'
	at com.fasterxml.jackson.dataformat.yaml.YAMLParser.nextToken(YAMLParser.java:532)
	at com.fasterxml.jackson.databind.ObjectMapper._readTreeAndClose(ObjectMapper.java:4854)
	at com.fasterxml.jackson.databind.ObjectMapper.readTree(ObjectMapper.java:3199)
```
 
_updateToken was added in jackson-core 2.16, but the leaked 2.15.x jackson-core wins the class-load race, so jackson-dataformat-yaml 2.18.x fails.

Same result for fluss-flink-2.2 , fluss-fs-s3, which ships only the unrelocated jackson-core 2.15.3. fluss-common is clean, so the unrelocated copy enters during the client/flink uber-jar assembly, a transitive com.fasterxml.jackson.core:jackson-core gets bundled without passing through the com.fasterxml.jackson → org.apache.fluss.shaded.jackson2.com.fasterxml.jackson relocation (the relocated set comes from the pre-shaded fluss-shaded-jackson artifact).

Expected
The uber-jars should expose no classes under the root com/fasterxml/jackson/**. Either relocate every jackson class (add/extend the com.fasterxml.jackson relocation rule so it also cover the transitively-bundled jackson-core), or exclude the transitive plain jackson-core from the shade artifactSet so only the pre-shaded fluss-shaded-jackson classes remain.

 
  Environment

  - Fluss: 0.9.0-incubating / 0.9.1-incubating (also reproduced on 1.0-SNAPSHOT)
  - Leaked jackson-core: 2.15.x; app jackson: 2.18.2
  - Flink 2.2, Java 21

### Solution

_No response_

Jackeyzhe:
## Summary

`fluss-shaded-jackson` contains unshaded Multi-Release JAR (MRJ) entries under
`META-INF/versions/*/com/fasterxml/jackson/core/io/doubleparser/` (and sibling
parser dirs). The Maven Shade Plugin only relocates base-path classes, not MRJ
entries. These leak into uber-jars (`fluss-client`, `fluss-flink-*`) and can
shadow downstream apps' jackson-core 2.16+, causing `NoSuchMethodError:
JsonToken._updateToken`.

## Root Cause

The shade plugin's relocation (`<relocation>`) only transforms classes at their
base path (`com/fasterxml/jackson/...`). Multi-Release JAR entries stored under
`META-INF/versions/<java-version>/com/fasterxml/jackson/...` are **not** relocated
and remain at their original package path. When these entries are packaged into
uber-jars, they can win the class-loading race against the consumer's own
jackson-core dependency.

## Fix

Add `<exclude>META-INF/versions/**/com/fasterxml/**</exclude>` to the root
`pom.xml` global shade filter (`<artifact>*</artifact>`). This cascades to ALL
modules via `combine.children="append"`.

The base `FastDoubleParser` implementation (at the relocated path) is sufficient
for all Java versions; the MRJ entries are only performance optimizations for
specific JDK versions.

## Test Plan

- Before fix: 5 uber-jars (fluss-client + fluss-flink-1.18/1.19/1.20/2.2/tiering)
  contain unshaded jackson MRJ entries
- After fix: 0 unshaded jackson entries in all 5 uber-jars
- `./mvnw -pl fluss-common test` passes (267 tests, 0 failures)
- `./mvnw spotless:check` passes
- `git diff --check` clean

## Files Changed

- `pom.xml` (+11 lines: 1 exclude rule in global shade filter)

Closes #3553

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -913,6 +913,17 @@
                                     'META-INF/versions/11/META-INF/maven/'.
                                      -->
                                 <exclude>**/META-INF/maven/?*/?*/**</exclude>
+                                <!-- Exclude unshaded Multi-Release Jar (MRJ) entries from
+                                     fluss-shaded-jackson. The shade relocation only handles
+                                     base-path classes; MRJ entries under META-INF/versions/*/
+                                     remain at their original com.fasterxml.jackson path and
+                                     leak into the uber-jar, shadowing downstream apps' own
+                                     jackson-core (e.g. NoSuchMethodError: JsonToken._updateToken
+                                     added in 2.16 but leaked 2.15.x wins class-load race).
+                                     The base FastDoubleParser implementation (at the relocated
+                                     path) is sufficient for all Java versions; the MRJ entries
+                                     are only performance optimizations. -->
+                                <exclude>META-INF/versions/**/com/fasterxml/**</exclude>
                             </excludes>
                         </filter>
                     </filters>


Jackeyzhe:
Gentle ping for review when anyone has a moment. All CI checks are green, and the fix is a single 11-line exclusion rule in the root pom.xml global shade filter.

Quick recap: fluss-shaded-jackson ships unshaded Multi-Release JAR entries (META-INF/versions/**/com/fasterxml/**) that the Maven Shade Plugin doesn't relocate. These leak into every uber-jar (fluss-client, fluss-flink-*) and can shadow downstream jackson-core 2.16+, causing NoSuchMethodError: JsonToken._updateToken at runtime.

The fix adds one <exclude> to the global <artifact>*</artifact> filter, which cascades to all modules via combine.children="append". Verified: 0 unshaded jackson entries in all 5 previously-affected uber-jars; fluss-common tests pass (267, 0 failures).

Would appreciate it if @wuchong, @fresh-borzoni, or @Yohahaha could take a look when convenient. Happy to make adjustments if anything needs changing. Thanks!

Yohahaha (approved):
thank you for the fix, verified locally, LGTM!

Jackeyzhe:
Thanks @Yohahaha for the review and for verifying locally — glad the root-pom MRJ  
exclusion fix works for you!

For context, this same shading-bug class showed up again in `fluss-fs-s3` (PR #4073,  
still under review) and I just filed #4116 to track the rest of the `fluss-filesystems`  
family (oss / gs / azure / obs / cos / hdfs) — they all still ship unshaded jackson +  
guava today. Happy to help drive those follow-ups.

PR #3885, opened 2026-08-06
[rust][client][server][gateway] Relax partial update nullability rule

### LINKED ISSUE #3849: [rust] Partial update wrongly rejects NOT NULL columns that are explicitly listed in target columns
### Search before asking

### Fluss version

main (development)

### Please describe the bug 🐞

In the Rust client, the partial-update sanity check in UpsertWriterFactory (fluss-rust/crates/fluss/src/client/table/upsert.rs) requires every non-primary-key, non-auto-increment column to be nullable — including columns that are explicitly listed in the target columns and therefore written on every request:
https://github.com/apache/fluss/blob/4f89e3e8b1bec158aa4a12bc8dbbbb695e3822b4/fluss-rust/crates/fluss/src/client/table/upsert.rs#L276-L289
The nullability rule only makes sense for omitted columns (the server has to fill them). A column that is present in the target columns is always provided by the writer, so requiring it to be nullable is unnecessary and rejects valid usage.

### Solution

Apply the nullability check only to omitted columns:
```
if !target_column_set[i] && !pk_column_set[i] && !auto_increment_column_set[i] { ... }
```
and adjust the error message to say "omitted column". Add unit tests covering: (a) a NOT NULL column listed in target columns is accepted, (b) an omitted NOT NULL column is still rejected.

gstamatakis95:
## Summary

Partial update validation required every non-primary-key column of a primary key table to be nullable, including
columns explicitly listed in the target columns. A listed column is supplied by the writer on every request, so the
requirement rejected valid schemas. The loop's own comment described the intended behavior ("check the columns not in
targetColumns"), but the loop never consulted the target column set.

The check is duplicated across four layers that validate independently, and all four carried the defect:

| Layer | Location |
|---|---|
| Rust client | `fluss-rust/.../client/table/upsert.rs`, `UpsertWriterFactory::sanity_check` |
| Java client | `fluss-client/.../writer/UpsertWriterImpl.java`, `sanityCheck` |
| Server | `fluss-server/.../kv/partialupdate/PartialUpdater.java`, `sanityCheck` |
| REST gateway | `fluss-gateway/src/protocol/rest/records.rs`, `sparse_targets` |

The reported symptom cannot be resolved in the Rust client alone. A client-only fix permits writer construction and
defers the same rejection to the server at first write, so all four layers change together.

## Resulting validation rules

| Column | Requirement | Rationale |
|---|---|---|
| Omitted from target columns | Must be nullable | `updateRow` writes null into it when the row does not yet exist |
| Listed in target columns | May be `NOT NULL` | The writer supplies a value on every request |
| Auto increment | Must be nullable | Always omitted, and assigned its value only after the merge |
| Non-primary-key target column, under partial delete | Must be nullable | `deleteRow` sets it to null unless the whole row is removed |

## Divergence from the fix proposed in the issue

The issue proposes retaining the auto-increment exemption
(`!target_column_set[i] && !pk_column_set[i] && !auto_increment_column_set[i]`) and this PR removes it. Removing the
exemption is the correct direction because of ordering on the write path. `PartialUpdater.updateRow` null-fills
omitted columns before `AutoIncrementUpdater` assigns a value (`KvWriteProcessor.processUpsert` to `applyInsert` to
`updateAutoIncrementColumns`). Relaxing the server instead would allow a `NOT NULL` auto increment column to reach
`BinaryWriter.createNotNullValueWriter` holding null, producing an NPE surfaced to the client as
`UnknownServerException`.

## Runtime guards

The previous rule guaranteed that a row reaching `PartialUpdater` cannot carry null in a non-nullable slot. Relaxing
it needs guards, because the decode path is not defensive: `InternalRow.createFieldGetter` and `CompactedRowReader`
omit the `isNullAt` branch for non-nullable types, and CompactedRow is a sequential variable-length encoding, so a
null bit in a non-nullable slot corrupts every later field (IndexedRow, whose reader honors the null bit per column,
corrupts only that column).

- `updateRow` rejects a null value supplied for a `NOT NULL` target column, primary key columns included. The check
  runs before any field getter, because the first getter deserializes the whole row with the non-null-checking
  readers and would fail first with an error dependent on the bytes that follow. `isNullAt` reads only the null-bit
  header, so the check itself never deserializes.
- `deleteRow` rejects partial delete when a non-primary-key target column is `NOT NULL`. The guard is placed after
  the `isFieldsNull` short circuit, so whole-row removal, which nulls nothing, remains legal. The guard is server
  side only, since legality depends on the stored row, and a client-side equivalent would be stricter and would make
  a server-supported operation unreachable.
- The Java client runs the same null check per row before encoding, with the server's message, since the encoder
  would otherwise fail with a bare `NullPointerException` before the request is sent.
- The gateway requires a non-null value for every `NOT NULL` target column in each upsert entry of a sparse batch,
  using the decoder's existing per-row required-column enforcement. Delete entries carry only primary key values, so
  the server judges them against the stored row.

One path stays outside the server guards: target columns covering every schema column short circuit in
`DefaultRowMerger.configureTargetColumns` and never build a `PartialUpdater`. The exposure equals a plain full-row
upsert, the Java client guard covers the case client side, and the encoders reject a null before anything is stored.

## Merge engines

The `first_row` and `versioned` mergers reject partial update outright, and the `aggregation` merger does not fill
omitted columns with null on the first write, so aggregation tables keep requiring every column except the primary key to
be nullable. The Java client now checks all three at writer creation with the server's wording. Before, `first_row`
and `versioned` failed at the first write as `UnknownServerException`, and aggregation was rejected at creation with
the generic message.
`MergeMode.OVERWRITE` is exempt, because `KvWriteProcessor` merges an overwrite with the default merger rather than
the configured merge engine.

The Rust client reads no merge engine configuration today. The table config is available at writer creation, but the
guard is not part of this PR, to keep the Rust diff small. A Rust user on such a table still sees the rejection at
the first write. The Rust client also has no per-row check for a null in a `NOT NULL` target column. Such a write fails
in the encoder with a type error before it is sent, so nothing bad reaches the server. The gateway performs no merge
engine check either, so a REST partial update on such a table is rejected by the backend writer rather than at
preflight.

## Error message

Changed from:

```
Partial Update requires all columns except primary key to be nullable, but column %s is NOT NULL.
```

to:

```
Partial Update requires all columns omitted from the target columns to be nullable, but omitted column %s is NOT NULL.
```

A `NOT NULL` auto increment column gets a dedicated message, because the user cannot follow the generic advice of
listing the column, targeting an auto increment column is itself rejected:

```
Partial Update requires the auto increment column %s to be nullable, since it is always omitted from the target columns and assigned by the server.
```

The gateway keeps its own message style and now reports the omitted column the same way, including the dedicated
auto increment message.

## Flink

A Flink sink writing a partial update (a column subset in the SQL INSERT, or partial update columns on
`FlussSinkBuilder`) with a `NOT NULL` target column initializes and writes, and the first retract record is rejected
by the server's delete guard, which is not retriable. Insert-only pipelines are unaffected. The sink does not check this at plan time, so the rejection surfaces when the first delete arrives.

## Compatibility

No wire format, storage format, or public API signature changes. The change relaxes validation, so previously
accepted writes remain accepted. Three cases move their failure from the first write to writer creation: partial
update on `first_row` and `versioned` tables, and a `NOT NULL` auto increment column. The rejection message for
aggregation tables changes from `Partial Update requires...` to `Partial aggregate requires...`.

Fixes #3849

### DIFF
--- a/fluss-client/src/main/java/org/apache/fluss/client/table/writer/UpsertWriterImpl.java
+++ b/fluss-client/src/main/java/org/apache/fluss/client/table/writer/UpsertWriterImpl.java
@@ -126,7 +126,6 @@ class UpsertWriterImpl extends AbstractTableWriter implements UpsertWriter {
             targetColumnsSet.set(targetColumnIndex);
         }
 
-        BitSet pkColumnSet = new BitSet();
         // check the target columns contains the primary key
         for (String key : primaryKeys) {
             int pkIndex = rowType.getFieldIndex(key);
@@ -136,7 +135,6 @@ class UpsertWriterImpl extends AbstractTableWriter implements UpsertWriter {
                                 "The target write columns %s must contain the primary key columns %s.",
                                 rowType.project(targetColumns).getFieldNames(), primaryKeys));
             }
-            pkColumnSet.set(pkIndex);
         }
 
         BitSet autoIncrementColumnSet = new BitSet();
@@ -152,17 +150,21 @@ class UpsertWriterImpl extends AbstractTableWriter implements UpsertWriter {
             autoIncrementColumnSet.set(autoIncrementColumnIndex);
         }
 
-        // check the columns not in targetColumns should be nullable
+        // an omitted column is written as null, so it must be nullable. auto increment columns
+        // are always omitted and only get their value on the server.
         for (int i = 0; i < rowType.getFieldCount(); i++) {
-            // column not in primary key and not in auto increment column
-            if (!pkColumnSet.get(i) && !autoIncrementColumnSet.get(i)) {
-                // the column should be nullable
-                if (!rowType.getTypeAt(i).isNullable()) {
+            if (!targetColumnsSet.get(i) && !rowType.getTypeAt(i).isNullable()) {
+                String columnName = rowType.getFieldNames().get(i);
+                if (autoIncrementColumnSet.get(i)) {
                     throw new IllegalArgumentException(
                             String.format(
-                                    "Partial Update requires all columns except primary key to be nullable, but column %s is NOT NULL.",
-                                    rowType.getFieldNames().get(i)));
+                                    "Partial Update requires the auto increment column %s to be nullable, since it is always omitted from the target columns and assigned by the server.",
+                                    columnName));
                 }
+                throw new IllegalArgumentException(
+                        String.format(
+                                "Partial Update requires all columns omitted from the target columns to be nullable, but omitted column %s is NOT NULL.",
+                                columnName));
             }
         }
     }
--- a/fluss-client/src/main/java/org/apache/fluss/client/table/writer/UpsertWriterImpl.java
+++ b/fluss-client/src/main/java/org/apache/fluss/client/table/writer/UpsertWriterImpl.java
@@ -155,13 +155,9 @@ class UpsertWriterImpl extends AbstractTableWriter implements UpsertWriter {
             autoIncrementColumnSet.set(autoIncrementColumnIndex);
         }
 
-        // The aggregation merge engine returns the new row unchanged on the first write instead of
-        // null filling the omitted columns, so it keeps requiring every column except the primary
-        // key to be nullable. The server enforces the same rule when it builds the merger, and
-        // checking it here keeps the rejection at writer creation instead of at the first write.
-        // OVERWRITE is exempt, since it bypasses the configured merge engine on the server and
-        // merges with the default merger, which null fills the omitted columns like any other
-        // partial update.
+        // the aggregation merge engine does not null fill omitted columns on the first write, so
+        // it keeps requiring every column except the primary key to be nullable, like the server.
+        // OVERWRITE is exempt, since the server merges it with the default merger instead.
         if (mergeEngineType == MergeEngineType.AGGREGATION && mergeMode != MergeMode.OVERWRITE) {
             for (int i = 0; i < rowType.getFieldCount(); i++) {
                 if (!primaryKeys.contains(rowType.getFieldNames().get(i))
--- a/fluss-client/src/main/java/org/apache/fluss/client/table/writer/UpsertWriterImpl.java
+++ b/fluss-client/src/main/java/org/apache/fluss/client/table/writer/UpsertWriterImpl.java
@@ -219,6 +219,21 @@ class UpsertWriterImpl extends AbstractTableWriter implements UpsertWriter {
                                 columnName));
             }
         }
+
+        // the auto increment column is always set, so a partial delete could never collapse
+        // the row and would always fail on a NOT NULL non-primary-key target column.
+        if (!autoIncrementColumnSet.isEmpty()) {
+            for (int i = 0; i < rowType.getFieldCount(); i++) {
+                if (targetColumnsSet.get(i)
+                        && !primaryKeys.contains(rowType.getFieldNames().get(i))
+                        && !rowType.getTypeAt(i).isNullable()) {
+                    throw new IllegalArgumentException(
+                            String.format(
+                                    "Partial Update on a table with an auto increment column requires all target columns except the primary key to be nullable, but target column %s is NOT NULL, since the auto increment column is always set and a partial delete could therefore never succeed.",
+                                    rowType.getFieldNames().get(i)));
+                }
+            }
+        }
     }
 
     /**
--- a/fluss-client/src/main/java/org/apache/fluss/client/table/writer/UpsertWriterImpl.java
+++ b/fluss-client/src/main/java/org/apache/fluss/client/table/writer/UpsertWriterImpl.java
@@ -274,9 +274,9 @@ class UpsertWriterImpl extends AbstractTableWriter implements UpsertWriter {
 
     /**
      * Rejects a null in a NOT NULL target column. Runs before any field getter or encoding, which
-     * would fail with a bare NullPointerException. Slightly stricter than the server when the
-     * target columns cover every schema column, since the server then skips the PartialUpdater, but
-     * such a row is not reachable through the public API.
+     * would fail with a bare NullPointerException. When the target columns cover every schema
+     * column the server skips the PartialUpdater and this check is the only guard, and the encoders
+     * would reject the null anyway, so it never changes which rows are accepted.
      */
     private void checkNotNullTargetColumns(InternalRow row) {
         for (int index : notNullTargetColumns) {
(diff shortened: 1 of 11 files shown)


fluss-client/src/test/java/org/apache/fluss/client/table/FlussTableITCase.java
```diff
@@ -959,6 +963,53 @@ void testPartialPutAndDelete() throws Exception {
...
+                    .isEqualTo(compactedRow(schema.getRowType(), new Object[] {1, "bbb", 100L}));
+
+            // updating the same target columns again keeps column b untouched
+            upsertWriter.upsert(row(1, null, 200L)).get();
+            assertThat(lookupRow(lookuper, rowKey))
+                    .isEqualTo(compactedRow(schema.getRowType(), new Object[] {1, "bbb", 200L}));
+
+            // a null in the NOT NULL target column c is rejected before anything is sent
+            assertThatThrownBy(() -> upsertWriter.upsert(row(1, null, null)))
+                    .isInstanceOf(IllegalArgumentException.class)
```
fresh-borzoni, line 998: Can you also assert the row is unchanged and a following upsert still lands?

fluss-server/src/main/java/org/apache/fluss/server/kv/partialupdate/PartialUpdater.java
```diff
@@ -134,11 +178,20 @@ public BinaryValue updateRow(@Nullable BinaryValue oldValue, BinaryValue partial
...
      * @return the value after partial deleted
+     * @throws InvalidTargetColumnException if a non-primary-key target column is NOT NULL, since it
+     *     would have to be set to null
      */
     public @Nullable BinaryValue deleteRow(BinaryValue value) {
         if (isFieldsNull(value.row, partialUpdateCols)) {
+            // the whole row is removed, so no column is set to null
             return null;
         } else {
+            if (!notNullNonPkTargetCols.isEmpty()) {
```
fresh-borzoni, line 198: On an auto increment table the isFieldsNull escape above can never fire - the server always fills the auto increment column, so it's never null. I tried to run it and upserts work, every delete throws. 
On main the schema was rejected outright. Should the check move up to sanityCheck?

website/docs/table-design/table-types/pk-table.md
```diff
@@ -74,6 +74,20 @@ follows:
...
 | 2 | 3.0 | t2 |
 
+### Nullability requirements
+
+A column omitted from the target columns (the columns being written) is set to `null` when the row does not exist yet,
+so every omitted column must be nullable. A target column is always supplied by the writer, so it may be declared
+`NOT NULL`. An auto increment column is always omitted from the target columns and only receives its value on the
+server, so it must be nullable as well.
+
+Partial delete has a stricter requirement. It sets the target columns other than the primary key to `null`, so it is
```
fresh-borzoni, line 84: table.delete.behavior=ignore|disable skips this path entirely, worth mentioning as the workaround. 
Also a Flink sink on this schema fails on the first delete and keeps restarting, since sink.ignore-delete defaults to false. 
Should we say deletes aren't available by default here?

fluss-client/src/main/java/org/apache/fluss/client/table/writer/UpsertWriterImpl.java
fresh-borzoni, line 165: FIRST_ROW and VERSIONED don't support partial update at all, and the old rule was catching some of those early by accident. Now they fail at first write with UnknownServerException. 
Should we reject them here too, like FlinkTableSink does?

Also Rust has no merge engines yet, so this only applies to Java clients. Shall we file an issue for the Rust side?

fluss-client/src/main/java/org/apache/fluss/client/table/writer/UpsertWriterImpl.java
```diff
@@ -152,19 +174,51 @@ private static void sanityCheck(
...
-            // column not in primary key and not in auto increment column
-            if (!pkColumnSet.get(i) && !autoIncrementColumnSet.get(i)) {
-                // the column should be nullable
-                if (!rowType.getTypeAt(i).isNullable()) {
+        // the first_row and versioned mergers reject partial update outright on the server, at
+        // the first write, so fail at writer creation instead. OVERWRITE is exempt, since the
+        // server merges it with the default merger.
+        if (mergeMode != MergeMode.OVERWRITE) {
+            if (mergeEngineType == MergeEngineType.FIRST_ROW) {
+                throw new IllegalArgumentException(
```
fresh-borzoni, line 182: This used to be accepted here and rejected by the server later. Can we add a test?

fresh-borzoni (commented):
@gstamatakis95 Thank you for the PR, left some comments, PTAL

fresh-borzoni:
@gstamatakis95  Some of my line refs drifted, sorry -  the test request is about the auto increment branch at line 210, and the assertion request is about the partial delete at line 1006. Also I missed that you've already added the first_row/versioned rejection, so please ignore that half.

fresh-borzoni:
@beryllw Can you also take a look, pls? Mb we wish to include it since it includes fixes for gateway

website/docs/table-design/table-types/pk-table.md
gstamatakis95: Also the answer to deletes is Yes. See also updated docs.

fluss-server/src/main/java/org/apache/fluss/server/kv/partialupdate/PartialUpdater.java
fresh-borzoni: Thanks, the auto increment case looks handled.

Correction on my earlier message: I said sanityCheck, but I meant just above the isFieldsNull check here, in sanityCheck it would reject the writer and take the upsert relaxation with it.

The same thing happens without auto increment. On (a pk, b STRING NOT NULL, c, d), a writer targeting a,b can delete fine while c and d are still null, the row just gets removed outright. Once anything writes c, that same delete throws instead, and keeps throwing.

Moving the guard above isFieldsNull covers both, and notNullNonPkTargetCols is empty for every schema that exists on main, so nothing changes for current tables. The check at :121 could then go, which would also let an upsert-only auto increment table keep a NOT NULL column.

beryllw:
@platinumhamburg cc

beryllw:
I agree that relaxing this restriction is reasonable in principle. However, it also adds considerable complexity across clients and the server, particularly around partial deletes, merge engines, and auto-increment columns. Since we haven’t seen a concrete community use case for this yet, perhaps we could keep the existing restriction for now and revisit it when there is clearer demand.

PR #3886, opened 2026-08-07
[lake/tiering][follow] Allow overriding tiering remote-first read preference

beryllw:
This is a follow-up of #3606, which hardcodes the remote-first read preference in TieringSource#createReader so that an explicitly configured client.scanner.log.read-preference is silently overwritten.

Apply the remote-first default in LakeTieringJobBuilder#build only when the option is not explicitly configured, so users can fall back to local-first as an escape hatch. The tiering default behavior and the default of regular Flink source jobs are not changed.

### Purpose

Linked issue: close #xxx


beryllw:
@luoyuxia @zuston cc

luoyuxia (approved):

zuston (approved):
nice catch.

PR #3887, opened 2026-08-07
[flink] add support of hybrid lookup with lake

zuston:
### Purpose

When Fluss is used as the hot layer and the lake as the cold layer for primary-key tables, the Flink connector previously had no way to look up data across both layers. This PR adds a hybrid lookup mechanism for auto-partitioned tables that determines whether a lookup key belongs to a live Fluss partition or should fall back to the lake.

### Brief change log

When `lookup.lake-fallback.enabled=true`, lookup tries Fluss first and falls back to an async lake point lookup if the partition is missing. This is built on a shared `LookupRuntime` abstraction (`FlussLookupRuntime` + `LakeLookupRuntime`) and a new `HybridLakeAsyncLookupFunction`.
Currently supports Paimon, async lookup, auto-partitioned tables, and full primary-key lookup only.


wuchong:
I think https://github.com/apache/fluss/pull/3630 already can cover this feature...

PR #3889, opened 2026-08-07
[server] Support group-based ACL authorization

### LINKED ISSUE #3829: [security] Support group-based ACL authorization for authenticated identities
### Search before asking

### Motivation

Fluss ACL entries can represent principals such as `User`, `Group`, and `Role`, but the current authentication and authorization path only carries and evaluates one `FlussPrincipal`.

For example, SASL/PLAIN always creates one `User` principal from the authorization ID. `Session` stores only that principal, and `DefaultAuthorizer` evaluates ACLs only against `session.getPrincipal()`.

This prevents an authenticated identity from carrying group membership such as:

```text
User:alice
Group:data-engineers
Group:fluss-readers
```

This is particularly relevant to [#3495](https://github.com/apache/fluss/issues/3495), where a validated JWT can provide the user identity through `sub` and group membership through a `groups` claim. Mapping only `sub` to a single `FlussPrincipal` cannot support group-based ACL policies.

This feature should provide authentication-mechanism-independent support for additional principals. It should be usable by OAUTHBEARER, Kerberos, mTLS, LDAP, and custom authentication plugins rather than being implemented only inside the OAuth mechanism.

Related design: [FIP-2: Fluss Authentication and Authorization](https://cwiki.apache.org/confluence/display/FLUSS/FIP-2%3A+Fluss+Authentication+and+Authorization).

### Solution

Extend the authenticated session identity from one primary principal to:

- One primary principal, preserving the existing `Session#getPrincipal()` behavior.
- Zero or more additional principals, such as `Group` or `Role`.
- An immutable, validated, and deduplicated collection of all authenticated principals.

Preserve source compatibility for existing authentication plugins, for example through a default server-authenticator method that returns no additional principals.

Update the default authorizer so an operation is allowed when the existing resource, operation, and host rules match an `ALLOW` ACL for any authenticated principal. Existing SASL/PLAIN behavior must remain unchanged because it produces only the primary `User` principal.

OAUTHBEARER support from #3495 can then map:

```text
sub -> primary User principal
groups[] -> additional Group principals
```

The implementation should also define which identity is used for logs and user metrics, avoid logging JWTs or credentials, and bound the number and size of additional principals.

Add tests covering:

- Existing single-user SASL/PLAIN authorization behavior.
- Authorization granted through a Group ACL.
- Multiple groups where only one matches.
- No matching user or group ACL.
- Duplicate and empty group values.
- Preservation of resource, operation, and host matching.
- Backward compatibility for authentication plugins that return only one principal.

### Anything else?

This proposal adds group-based ACL authorization. It does not introduce a complete RBAC management system with role creation, user-to-role assignment, role hierarchies, or role lifecycle APIs. Those capabilities can be designed separately if needed.

### Willingness to contribute

litiliu:
### Purpose

Linked issue: close #3829

Allow authentication mechanisms to attach group and role principals to an authenticated session so ACL authorization can evaluate both the primary and additional principals.

### Brief change log

- Allow authenticators to optionally provide additional principals; existing implementations continue to return none by default.
- Propagate additional principals through `FlussRequest` and `Session`.
- Evaluate regular ACLs against the primary and additional principals while keeping superuser checks limited to the primary principal.
- Update the security documentation for primary and additional principal semantics.

### Tests

- `AuthenticationTest`
- `DefaultAuthorizerTest`

### API and Format

This adds a default method to `ServerAuthenticator`. It does not change the RPC wire format or storage format.

### Documentation

Updated the authorization and security overview documentation.

### DIFF
--- a/fluss-common/src/main/java/org/apache/fluss/security/auth/ServerAuthenticator.java
+++ b/fluss-common/src/main/java/org/apache/fluss/security/auth/ServerAuthenticator.java
@@ -23,6 +23,8 @@ import org.apache.fluss.security.acl.FlussPrincipal;
 
 import java.io.Closeable;
 import java.io.IOException;
+import java.util.Collection;
+import java.util.Collections;
 
 /**
  * Authenticator for server side.
@@ -95,6 +97,11 @@ public interface ServerAuthenticator extends Closeable {
      * complete).
      */
     FlussPrincipal createPrincipal();
+
+    /** Creates additional principals associated with the authenticated identity. */
+    default Collection<FlussPrincipal> createAdditionalPrincipals() {
+        return Collections.emptyList();
+    }
 
     /** Close the authenticator. */
     default void close() throws IOException {}
--- a/fluss-common/src/main/java/org/apache/fluss/security/auth/ServerAuthenticator.java
+++ b/fluss-common/src/main/java/org/apache/fluss/security/auth/ServerAuthenticator.java
@@ -23,6 +23,8 @@ import org.apache.fluss.security.acl.FlussPrincipal;
 
 import java.io.Closeable;
 import java.io.IOException;
+import java.util.Collections;
+import java.util.List;
 
 /**
  * Authenticator for server side.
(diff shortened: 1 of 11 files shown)


litiliu:
@loserwang1024  Please help review this PR when you have time, Thanks in advance!

litiliu:
@affo  Please take a look when you have time, Thanks!

litiliu:
@polyzos  PTAL, Thanks!

affo (commented):
Hello and thanks for the amazing contribution!
While the contribution holds valuable, this looks to me as a divergence from the original https://cwiki.apache.org/confluence/spaces/FLUSS/pages/372214350/FIP-2+Fluss+Authentication+and+Authorization in that it replaces a single-principal model with a multi-principal model. It also touches a `PublicEvolving` interface.
I would favor an explicit FIP round, or even something lighter, but disclosed to the entire community in order to change the model.

For example, why having the concept of "additional principals" and not change `FlussPrincipal createPrincipal();` to `Iterable<FlussPrincipal> createPrincipals();`?

Maybe I am just being picky, I suggest to involve into this review a committer/PMC with a solid grasp of this part of the codebase 🤝

PR #3899, opened 2026-08-08
[rust] Fix log append rejecting NOT NULL columns

### LINKED ISSUE #3897: [rust] Log append reject NOT NULL columns
### Search before asking

### Fluss version

main (development)

### Please describe the bug 🐞

Appending to a log table with a NOT NULL column failed on the first append (declared as non-nullable but contains null values), regardless of the data even it had no nulls

### Solution

i think fluss-rust estimates Arrow overhead using a throwaway one-row all-null probe batch, but validated it against the real (non-nullable) schema, so any NOT NULL column made the probe fail before user data was touched. 

Can build the probe from a nullability-relaxed schema copy instead; nullability is schema-header metadata that's excluded from the measurement, so the estimate is unchanged

charlesdong1991:
### Purpose

Linked issue: close #3897

### Brief change log

- Build the probe from a nullability-relaxed schema clone (nullability is schema-message metadata, excluded from the measured overhead). 
- Adds a regression test.

### Tests

- Adds a regression test.

### DIFF
--- a/fluss-rust/crates/fluss/src/record/arrow.rs
+++ b/fluss-rust/crates/fluss/src/record/arrow.rs
@@ -590,15 +590,15 @@ fn estimate_arrow_ipc_overhead(
 ) -> Result<usize> {
     use arrow::array::new_null_array;
 
-    // Create a 1-row batch of nulls. Null arrays have minimal, predictable
-    // data: no validity bitmap, no variable-length data, just fixed-width
-    // zero buffers. This lets us compute raw data size exactly.
-    let null_arrays: Vec<ArrayRef> = schema
-        .fields()
-        .iter()
-        .map(|field| new_null_array(field.data_type(), 1))
-        .collect();
-    let batch = RecordBatch::try_new(schema.clone(), null_arrays)?;
+    let fields = schema.fields();
+    let mut probe_fields = Vec::with_capacity(fields.len());
+    let mut null_arrays: Vec<ArrayRef> = Vec::with_capacity(fields.len());
+    for f in fields {
+        null_arrays.push(new_null_array(f.data_type(), 1));
+        probe_fields.push(f.as_ref().clone().with_nullable(true));
+    }
+    let probe_schema: SchemaRef = Arc::new(arrow_schema::Schema::new(probe_fields));
+    let batch = RecordBatch::try_new(probe_schema.clone(), null_arrays)?;
 
     // Sum the raw buffer sizes — this is what buffer_size() would report.
     let raw_data: usize = batch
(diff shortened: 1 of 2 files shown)


charlesdong1991:
cc @fresh-borzoni @leekeiabstraction would be great if can take a look! 🙏

fluss-rust/crates/fluss/src/record/arrow.rs
```diff
@@ -1902,6 +1902,83 @@ mod tests {
     use crate::row::{DataGetters, GenericRow};
     use crate::test_utils::build_table_info;
 
+    #[test]
+    fn nonnullable_append_builder_roundtrips_for_both_append_modes() {
```
leekeiabstraction, line 1906: Good catch on the bug, are you OK to also look into adding to existing integration test for log tables? Would be great if the tests there exercise some non_nullable columns as well

leekeiabstraction (commented):
+1 and TY for the PR. Left a comment, PTAL

fluss-rust/crates/fluss/src/record/arrow.rs
```diff
@@ -590,15 +590,15 @@ fn estimate_arrow_ipc_overhead(
...
-    // Create a 1-row batch of nulls. Null arrays have minimal, predictable
-    // data: no validity bitmap, no variable-length data, just fixed-width
-    // zero buffers. This lets us compute raw data size exactly.
-    let null_arrays: Vec<ArrayRef> = schema
-        .fields()
-        .iter()
-        .map(|field| new_null_array(field.data_type(), 1))
-        .collect();
-    let batch = RecordBatch::try_new(schema.clone(), null_arrays)?;
+    let fields = schema.fields();
```
leekeiabstraction, line 593: Fix looks right to me. Worth noting the the fix unlocks the prebuilt arrow-batch path, which does not validate the caller's batch schema against the table's so nulls can still get written into a NOT NULL column and then fail to decode later. 

Are you OK to fix that part in this PR too or follow up with new issue/PR?

Much appreciated!

fluss-rust/crates/fluss/src/record/arrow.rs
charlesdong1991: sounds good, added! @leekeiabstraction

fluss-rust/crates/fluss/src/record/arrow.rs
charlesdong1991: Yep, i am aware of that, prefer to handle it in a follow-up, and also got a PoC running on my side so i can fully exercise that path before opening the next PR. 

will file an issue to track if you are okay with that! @leekeiabstraction

charlesdong1991:
Thanks! addressed your comment, PTAL @leekeiabstraction

fluss-rust/crates/fluss/src/record/arrow.rs
leekeiabstraction: Please do file the issue to track it. Thank you @charlesdong1991 !

leekeiabstraction (approved):
LGTM.

PR #3906, opened 2026-08-09
[flink] Validate database existence when resolving catalog functions in FlinkCatalog

### LINKED ISSUE #3901: [Bug] FlinkCatalog resolves built-in functions under nonexistent databases
### Search before asking

- [x] I searched the existing issues and found nothing similar.

### Fluss version

main (development)

### Please describe the bug 🐞

`FlinkCatalog` currently ignores the database component when resolving the built-in bitmap functions:

- `listFunctions(String dbName)` always returns `BUILTIN_BITMAP_FUNCTIONS`.
- `functionExists(ObjectPath)` only checks the function name.
- `getFunction(ObjectPath)` only checks the function name.

As a result, a fully qualified reference can resolve even when its database does not exist:

```sql
SELECT fluss_catalog.nonexistent.rb_build(ARRAY[1, 2]);
```

This is inconsistent with the database-scoped semantics of Flink catalog functions. It also makes it impossible to guarantee that the database component of `<catalog>.<database>.<function>` identifies an existing database.

The expected behavior is:

- `listFunctions(nonexistentDatabase)` throws `DatabaseNotExistException`.
- `functionExists(nonexistentDatabase.function)` returns `false`.
- `getFunction(nonexistentDatabase.function)` throws `FunctionNotExistException`.
- Built-in functions continue to resolve from every existing database so that unqualified references work with the current Fluss catalog and database.

### Solution

Validate the database before listing or resolving built-in functions in `FlinkCatalog`, and add unit/integration coverage for:

1. Listing and resolving built-in functions from an existing database.
2. Resolving a fully qualified built-in function from a nonexistent database.
3. Calling a built-in function by its unqualified name after switching to a Fluss catalog/database.

pbanakar:
### Purpose

Linked issue: close #3901

`FlinkCatalog#listFunctions`, `#functionExists`, and `#getFunction` resolved the built-in RoaringBitmap SQL functions (introduced in FIP-37) purely by function name, ignoring the database component of the request entirely. As a result, a fully qualified reference to a built-in function resolved successfully even when its database did not exist, e.g.:

```sql
SELECT fluss_catalog.nonexistent_db.rb_build(ARRAY[1, 2]);
```

This broke the database-scoped semantics Flink catalogs are expected to guarantee, and made it impossible to rely on the database component of `<catalog>.<database>.<function>` identifying an existing database.

### Brief change log

- `FlinkCatalog.java`:
  - `listFunctions(String dbName)` now throws `DatabaseNotExistException` if `dbName` does not exist, before returning the built-in bitmap function names.
  - `functionExists(ObjectPath)` now returns `false` if the object path's database does not exist, before checking the built-in function map.
  - `getFunction(ObjectPath)` now throws `FunctionNotExistException` if the object path's database does not exist, before resolving the function class.
  - All three reuse the existing `databaseExists(String)` (already backed by  `admin.databaseExists()`), so no new dependency or RPC path was introduced.
  - Built-in functions still resolve unqualified from any existing database — only the database-existence gate is new; no scoping to a specific database was added.

### Tests

- Verified locally:
  - `./mvnw test -pl fluss-flink/fluss-flink-common -Dtest="FlinkCatalogTest"` — BUILD SUCCESS
  - `./mvnw verify -pl fluss-flink/fluss-flink-common -Dit.test="Flink118CatalogITCase,Flink119CatalogITCase,Flink120CatalogITCase,Flink22CatalogITCase,RbFunctionsCatalogITCase"` — BUILD SUCCESS

### API and Format

no new public API is introduced.

### Documentation

No user-facing documentation changes needed

### DIFF
--- a/fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/catalog/FlinkCatalog.java
+++ b/fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/catalog/FlinkCatalog.java
@@ -802,11 +802,17 @@ public class FlinkCatalog extends AbstractCatalog {
     @Override
     public List<String> listFunctions(String dbName)
             throws DatabaseNotExistException, CatalogException {
+        if (!databaseExists(dbName)) {
+            throw new DatabaseNotExistException(getName(), dbName);
+        }
         return new ArrayList<>(BUILTIN_BITMAP_FUNCTIONS.keySet());
     }
 
     @Override
     public boolean functionExists(ObjectPath objectPath) throws CatalogException {
+        if (!databaseExists(objectPath.getDatabaseName())) {
+            return false;
+        }
         return BUILTIN_BITMAP_FUNCTIONS.containsKey(
                 objectPath.getObjectName().toLowerCase(Locale.ROOT));
     }
@@ -814,6 +820,9 @@ public class FlinkCatalog extends AbstractCatalog {
     @Override
     public CatalogFunction getFunction(ObjectPath functionPath)
             throws FunctionNotExistException, CatalogException {
+        if (!databaseExists(functionPath.getDatabaseName())) {
+            throw new FunctionNotExistException(getName(), functionPath);
+        }
         String className =
                 BUILTIN_BITMAP_FUNCTIONS.get(functionPath.getObjectName().toLowerCase(Locale.ROOT));
         if (className == null) {
--- a/fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/catalog/FlinkCatalog.java
+++ b/fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/catalog/FlinkCatalog.java
@@ -811,7 +811,10 @@ public class FlinkCatalog extends AbstractCatalog {
     @Override
     public boolean functionExists(ObjectPath objectPath) throws CatalogException {
         if (!databaseExists(objectPath.getDatabaseName())) {
-            return false;
+            throw new CatalogException(
+                    String.format(
+                            "Database %s does not exist in catalog %s.",
+                            objectPath.getDatabaseName(), getName()));
         }
         return BUILTIN_BITMAP_FUNCTIONS.containsKey(
                 objectPath.getObjectName().toLowerCase(Locale.ROOT));
@@ -821,7 +824,10 @@ public class FlinkCatalog extends AbstractCatalog {
     public CatalogFunction getFunction(ObjectPath functionPath)
             throws FunctionNotExistException, CatalogException {
         if (!databaseExists(functionPath.getDatabaseName())) {
-            throw new FunctionNotExistException(getName(), functionPath);
+            throw new CatalogException(
+                    String.format(
+                            "Database %s does not exist in catalog %s.",
+                            functionPath.getDatabaseName(), getName()));
         }
         String className =
                 BUILTIN_BITMAP_FUNCTIONS.get(functionPath.getObjectName().toLowerCase(Locale.ROOT));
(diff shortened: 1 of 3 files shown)


fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/catalog/FlinkCatalog.java
```diff
@@ -802,18 +802,27 @@ public void alterPartition(
...
+        }
         return BUILTIN_BITMAP_FUNCTIONS.containsKey(
                 objectPath.getObjectName().toLowerCase(Locale.ROOT));
     }
 
     @Override
     public CatalogFunction getFunction(ObjectPath functionPath)
             throws FunctionNotExistException, CatalogException {
+        if (!databaseExists(functionPath.getDatabaseName())) {
+            throw new FunctionNotExistException(getName(), functionPath);
```
wuchong, line 824: Masking a missing database as a missing function can be misleading to users. I believe we should explicitly inform the user that the database does not exist. We can throw a `CatalogException` while wrapping it with a clear "database not found" message.

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/catalog/FlinkCatalog.java
```diff
@@ -802,18 +802,27 @@ public void alterPartition(
...
+        if (!databaseExists(dbName)) {
+            throw new DatabaseNotExistException(getName(), dbName);
+        }
         return new ArrayList<>(BUILTIN_BITMAP_FUNCTIONS.keySet());
     }
 
     @Override
     public boolean functionExists(ObjectPath objectPath) throws CatalogException {
+        if (!databaseExists(objectPath.getDatabaseName())) {
+            return false;
```
wuchong, line 814: Same here, we should throw `CatalogException` with database not found message rather than telling users the function not exist.

wuchong (commented):
Thanks @pbanakar , I left some minor comments.

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/catalog/FlinkCatalog.java
pbanakar: Updated getFunction to throw CatalogException with an explicit "Database X does not exist in catalog Y" message when the database is not found,

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/catalog/FlinkCatalog.java
pbanakar: Same, updated this and also the unit test assertions accordingly

pbanakar:
Thanks @wuchong! Quick heads up,  throwing CatalogException in functionExists broke FlinkProcedureITCase#testCallNonExistProcedure in CI. Turns out Flink internally uses functionExists() as a boolean probe during function/procedure resolution, so throwing there short-circuits the lookup chain before it reaches Calcite's validator users get "Database ... does not exist" instead of the expected "No match found for function signature ...".

Can we revert just functionExists back to returning false and keep getFunction throwing CatalogException as you suggested, Does that split sound right to you, or would you prefer a different approach for functionExists?

wuchong:
Thank you @pbanakar , I think you are right. I'm fine to revert the changes on `functionExists`.

pbanakar:
thanks jark!, reverted it and i'll ping here once again once the ci passes

pbanakar:
Hi @wuchong , Hit the same issue with getFunction as functionExists CatalogException here also breaks Flink's internal probe chain, need to revert getFunction also, you suggestions valuable here!  if you know of a cleaner way to handel this instead of reverting? , happy to dig further, Please let me know!

wuchong:
Thank you @pbanakar , let's revert both of them.

wuchong (approved):

PR #3910, opened 2026-08-09
[flink] Track consumed progress during undo recovery

### LINKED ISSUE #3909: [Bug] Undo recovery may time out when scanning empty log batches
### Search before asking

### Fluss version

0.9.0 (latest release)

### Please describe the bug 🐞

Undo recovery currently determines whether a bucket has reached its recovery target from the offset of the last materialized `ScanRecord`.

This does not work when the recovery range contains empty or fully filtered log batches. Such batches advance the scanner's consumed offset but do not produce any `ScanRecord`. As a result, the recorded offset may remain behind the target log end offset even though the scanner has already consumed the entire recovery range.

The affected bucket is therefore never considered complete, and undo recovery eventually fails with a timeout.

The same completion logic can also process records at or beyond the captured recovery target when a poll returns data crossing that boundary.

### Solution

Track recovery completion using the consumed offset reported by each poll instead of the offset of the last materialized record.

Undo recovery now completes a bucket after the scanner has consumed up to the captured target log end offset, including ranges containing empty or filtered batches. Records at or beyond that target are excluded from undo processing.

platinumhamburg:
### Purpose

Linked issue: close #3909 3909

### DIFF
--- a/fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/sink/undo/BucketRecoveryContext.java
+++ b/fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/sink/undo/BucketRecoveryContext.java
@@ -43,7 +43,7 @@ public class BucketRecoveryContext {
     private final long logEndOffset;
 
     private final Set<ByteArrayWrapper> processedKeys;
-    private long lastProcessedOffset;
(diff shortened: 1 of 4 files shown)


luoyuxia (approved):
LGTM!

PR #3915, opened 2026-08-10
[flink] Route undo offsets by sink subtask

### LINKED ISSUE #3914: [Bug] Undo recovery offset reports may be routed across sink subtasks
### Search before asking

### Fluss version

0.9.0 (latest release)

### Please describe the bug 🐞

The Flink sink uses a `ProducerOffsetReporter` to forward completed write offsets from each `UpsertSinkWriter` to its upstream `UndoRecoveryOperator`.

Flink creates a separate deserialized factory instance for each runtime subtask. However, these instances originate from the same serialized factory and therefore retain the same holder ID.

When multiple sink subtasks run in the same TaskManager JVM, they also share the same static delegate registry. Registrations under the shared holder ID may then overwrite one another. Depending on task placement and the timing of operator registration and the reporter's first delegate lookup, an offset report may be routed to an `UndoRecoveryOperator` belonging to another sink subtask.

A delayed close from an earlier execution attempt may also remove the registry entry installed by its replacement because registry removal is not ownership-aware.

The problem does not require rescaling, although failover or rescaling can make it more likely to surface by changing subtask placement and lifecycle ordering. It does not necessarily affect the table result during uninterrupted execution, but it can make the checkpointed Undo Recovery offsets unreliable and affect a later recovery.

### Solution

Bind each producer offset reporter to a reporter group ID and the runtime sink subtask index. Independently deserialized factory instances preserve the group ID, while each runtime writer and operator derive the same subtask-specific reporter key.

Remove a reporter mapping only when the closing operator still owns it, preventing a delayed close from unregistering a replacement operator. Fail fast when an offset cannot be delivered instead of silently dropping it.

- [*] I'm willing to submit a PR!

platinumhamburg:
### Purpose

Linked issue: close #3914 3914

### DIFF
--- a/fluss-flink/fluss-flink-1.18/src/main/java/org/apache/fluss/flink/adapter/SinkAdapter.java
+++ b/fluss-flink/fluss-flink-1.18/src/main/java/org/apache/fluss/flink/adapter/SinkAdapter.java
@@ -33,9 +33,12 @@ public abstract class SinkAdapter<InputT> implements Sink<InputT> {
 
     @Override
     public SinkWriter<InputT> createWriter(InitContext initContext) throws IOException {
-        return createWriter(initContext.getMailboxExecutor(), initContext.metricGroup());
+        return createWriter(
+                initContext.getMailboxExecutor(),
+                initContext.metricGroup(),
+                initContext.getSubtaskId());
     }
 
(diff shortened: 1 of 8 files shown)


luoyuxia (approved):
+1

PR #3920, opened 2026-08-10
[server] Fix ISR shrink race during leader transition

### LINKED ISSUE #3927: maybeShrinkIsr can submit an ISR excluding the leader during a follower-to-leader transition
### Search before asking

### Fluss version

main (development)

### Description

A race condition in `Replica#maybeShrinkIsr` can produce an invalid `AdjustISR` request when a replica is promoted from follower to leader concurrently with the periodic ISR shrink task.

The request may contain a stale or empty ISR that does not include the current leader. The Coordinator previously accepted this update, leaving the bucket in an invalid state where `leader ∉ ISR`.

### Symptoms

This issue may occur during a rolling restart or leader rebalance. A characteristic log is:

```text
Shrink ISR From [] to [] for bucket <bucket>. Leader: (..., out of sync replicas: [5])
```

Other indicators include:

- The bucket metadata contains a leader that is absent from the ISR, for example `leader=<id>, isr=[]`.
- The number of in-sync replicas remains lower than the total number of replicas.
- TabletServer readiness or a rolling restart may remain blocked.
- Recovery may require another leader transition or recreation of the affected leader replica.

### Root cause

Previously, `maybeShrinkIsr` read `isrState` before acquiring `leaderIsrUpdateLock`, but checked the replica role and follower lag after acquiring the lock.

The following interleaving was therefore possible:

1. `maybeShrinkIsr` snapshots the follower state, whose local ISR is empty.
2. Before the shrink check runs, `makeLeader` promotes the replica and installs the new leader state and ISR.
3. `maybeShrinkIsr` acquires the lock and now observes that the replica is the leader with an out-of-sync follower.
4. It passes the leader-only shrink check, but constructs the new ISR from the stale follower snapshot.
5. An invalid `AdjustISR` request whose ISR excludes the leader is submitted and accepted by the Coordinator.

The race mixes the new leader role with an ISR snapshot from the previous follower state.

### Expected behavior

The leader role, committed ISR, and follower lag used for an ISR shrink must come from a consistent state protected by `leaderIsrUpdateLock`.

The Coordinator must never accept an ISR update that does not contain its leader.

### Solution

Read and validate `isrState` under `leaderIsrUpdateLock`. Before preparing the shrink, acquire the write lock, re-read the committed ISR, and recompute the out-of-sync replicas.

As a safety net, reject `AdjustISR` requests at the Coordinator when the proposed ISR does not contain the proposed leader.

Fixed by https://github.com/apache/fluss/pull/3920.

hanliu0830:
Re-read ISR state and recompute follower lag under the leader/ISR lock before preparing a shrink, preventing stale follower state from crossing a leader transition.

Reject AdjustISR proposals whose leader is absent from the new ISR as a coordinator-side safety net.

### Purpose

Linked issue: close #3927

maybeShrinkIsr previously read isrState before acquiring the leader/ISR lock. If the replica became leader between that read and the shrink check, it could combine the new leader role with stale follower ISR state, potentially producing an invalid ISR that excluded the leader. This change keeps the state read and shrink calculation under the lock, with Coordinator-side validation as a safety net.

### DIFF
--- a/fluss-server/src/main/java/org/apache/fluss/server/coordinator/CoordinatorEventProcessor.java
+++ b/fluss-server/src/main/java/org/apache/fluss/server/coordinator/CoordinatorEventProcessor.java
@@ -2035,13 +2035,27 @@ public class CoordinatorEventProcessor implements EventProcessor {
                 throw new InvalidUpdateVersionException(
                         "The request bucket epoch in adjust isr request is lower than current bucket epoch in coordinator.");
             } else {
-                if (!newLeaderAndIsr.isr().contains(newLeaderAndIsr.leader())) {
+                if (newLeaderAndIsr.leader() != currentLeaderAndIsr.leader()) {
+                    String errorMsg =
+                            String.format(
+                                    "Rejecting adjustIsr request for table bucket %s because request leader %s "
+                                            + "does not match current leader %s",
+                                    tableBucket,
+                                    newLeaderAndIsr.leader(),
+                                    currentLeaderAndIsr.leader());
+                    LOG.error(errorMsg);
+                    throw new FencedLeaderEpochException(errorMsg);
+                }
+
+                if (!newLeaderAndIsr.isr().contains(currentLeaderAndIsr.leader())) {
                     String errorMsg =
                             String.format(
                                     "Rejecting adjustIsr request for table bucket %s because leader %s "
                                             + "is not in the new ISR %s",
-                                    tableBucket, newLeaderAndIsr.leader(), newLeaderAndIsr.isr());
-                    LOG.info(errorMsg);
+                                    tableBucket,
+                                    currentLeaderAndIsr.leader(),
+                                    newLeaderAndIsr.isr());
+                    LOG.error(errorMsg);
                     throw new IneligibleReplicaException(errorMsg);
                 }
 
--- a/fluss-server/src/main/java/org/apache/fluss/server/coordinator/CoordinatorEventProcessor.java
+++ b/fluss-server/src/main/java/org/apache/fluss/server/coordinator/CoordinatorEventProcessor.java
@@ -2035,6 +2035,16 @@ public class CoordinatorEventProcessor implements EventProcessor {
                 throw new InvalidUpdateVersionException(
                         "The request bucket epoch in adjust isr request is lower than current bucket epoch in coordinator.");
             } else {
+                if (!newLeaderAndIsr.isr().contains(newLeaderAndIsr.leader())) {
+                    String errorMsg =
+                            String.format(
+                                    "Rejecting adjustIsr request for table bucket %s because leader %s "
+                                            + "is not in the new ISR %s",
+                                    tableBucket, newLeaderAndIsr.leader(), newLeaderAndIsr.isr());
+                    LOG.info(errorMsg);
+                    throw new IneligibleReplicaException(errorMsg);
+                }
+
                 // Check if the new ISR are all ineligible replicas (doesn't contain any shutting
                 // down tabletServers).
                 Set<Integer> ineligibleReplicas = new HashSet<>(newLeaderAndIsr.isr());
--- a/fluss-server/src/main/java/org/apache/fluss/server/replica/Replica.java
+++ b/fluss-server/src/main/java/org/apache/fluss/server/replica/Replica.java
@@ -1865,25 +1865,25 @@ public final class Replica {
      * <p>This function can be triggered when a replica's LEO has incremented.
      */
     private void maybeExpandISr(FollowerReplica followerReplica) {
-        IsrState currentIsrState = isrState;
         boolean needsIsrUpdate =
-                !currentIsrState.isInflight()
-                        && inReadLock(leaderIsrUpdateLock, () -> needsExpandIsr(followerReplica));
+                inReadLock(
+                        leaderIsrUpdateLock,
+                        () -> !isrState.isInflight() && needsExpandIsr(followerReplica));
 
         if (needsIsrUpdate) {
             Optional<IsrState.PendingExpandIsrState> adjustIsrUpdateOpt =
                     inWriteLock(
                             leaderIsrUpdateLock,
                             () -> {
+                                IsrState currentIsrState = isrState;
                                 // check if this replica needs to be added to the ISR.
-                                if (currentIsrState instanceof IsrState.CommittedIsrState) {
-                                    if (needsExpandIsr(followerReplica)) {
-                                        return Optional.of(
-                                                prepareIsrExpand(
-                                                        (IsrState.CommittedIsrState)
-                                                                currentIsrState,
-                                                        followerReplica.getFollowerId()));
-                                    }
+                                if (isLeader()
+                                        && currentIsrState instanceof IsrState.CommittedIsrState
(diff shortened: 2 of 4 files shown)


fluss-server/src/main/java/org/apache/fluss/server/coordinator/CoordinatorEventProcessor.java
```diff
@@ -2035,6 +2035,16 @@ private void validateLeaderAndIsr(TableBucket tableBucket, LeaderAndIsr newLeade
                 throw new InvalidUpdateVersionException(
                         "The request bucket epoch in adjust isr request is lower than current bucket epoch in coordinator.");
             } else {
+                if (!newLeaderAndIsr.isr().contains(newLeaderAndIsr.leader())) {
```
swuferhong: The leader that actually gets persisted is `currentLeaderAndIsr.leader()`, not `newLeaderAndIsr.leader()`. Validating one field and persisting a different one leaves a gap in what is meant to be a safety net:

```java
else if (!newLeaderAndIsr.isr().contains(currentLeaderAndIsr.leader()))
```

fluss-server/src/main/java/org/apache/fluss/server/coordinator/CoordinatorEventProcessor.java
```diff
@@ -2035,6 +2035,16 @@ private void validateLeaderAndIsr(TableBucket tableBucket, LeaderAndIsr newLeade
...
+                    String errorMsg =
+                            String.format(
+                                    "Rejecting adjustIsr request for table bucket %s because leader %s "
+                                            + "is not in the new ISR %s",
+                                    tableBucket, newLeaderAndIsr.leader(), newLeaderAndIsr.isr());
+                    LOG.info(errorMsg);
```
swuferhong: recommand to use `LOG.error`.  `LOG.info` is too quiet for this

fluss-server/src/main/java/org/apache/fluss/server/replica/Replica.java
```diff
@@ -1897,28 +1897,27 @@ private void maybeExpandISr(FollowerReplica followerReplica) {
     }
 
     void maybeShrinkIsr() {
```
swuferhong, line 1898: `maybeExpandISr` still has the same stale-read bug. Maybe we can fix it in the same pr.

fluss-server/src/test/java/org/apache/fluss/server/replica/AdjustIsrTest.java
```diff
@@ -126,6 +129,72 @@ void testShrinkIsr() throws Exception {
...
+            retry(
+                    Duration.ofSeconds(10),
+                    () -> assertThat(leaderIsrUpdateLock.getQueueLength()).isEqualTo(3));
+        } finally {
+            leaderIsrUpdateLock.writeLock().unlock();
+        }
+
+        makeLeaderFuture.get(10, TimeUnit.SECONDS);
+        advanceClockFuture.get(10, TimeUnit.SECONDS);
+        shrinkIsrFuture.get(10, TimeUnit.SECONDS);
```
swuferhong: This submits three tasks that all block on a lock to `ForkJoinPool.commonPool()`. `CompletableFuture.runAsync` only uses the common pool when `getCommonPoolParallelism() > 1`, and FJP does not compensate for plain lock() blocking (only for ManagedBlocker/join). On a 3-core machine parallelism is 2, so the third task never starts and retry(() -> getQueueLength() == 3) times out — a flaky failure that depends purely on the CI runner's core count. getQueueLength() is also documented as an estimate, which makes it a fragile synchronisation barrier.

Please also use an explicit executor rather than the common pool, so the test cannot be affected by unrelated common-pool usage in the same JVM.

fluss-server/src/test/java/org/apache/fluss/server/coordinator/CoordinatorEventProcessorTest.java
```diff
@@ -1271,6 +1271,41 @@ void testProcessAdjustIsr() throws Exception {
         assertThat(resultForBucketMap.keySet())
                 .containsAnyElementsOf(bucketLeaderAndIsrMap.keySet());
         assertThat(resultForBucketMap.values()).allMatch(AdjustIsrResultForBucket::succeeded);
+
```
swuferhong, line 1297: Appending the negative case to testProcessAdjustIsr couples "normal adjust succeeds" with "illegal adjust is rejected" in one test, which makes failures harder to localise. Suggest extracting it into its own test, e.g. `testProcessAdjustIsrRejectsIsrWithoutLeader`.

swuferhong (commented):
Hi, @hanliu0830. Thanks for the good catch for this error. I left some comments:

fluss-server/src/main/java/org/apache/fluss/server/coordinator/CoordinatorEventProcessor.java
hanliu0830: Good catch. Fixed in 33a1b0514: the safety-net check now verifies that `newLeaderAndIsr.isr()` contains `currentLeaderAndIsr.leader()`, which is the leader persisted by the Coordinator. I also added a separate check that the requested leader matches the current leader, so a stale or non-leader request is fenced.

fluss-server/src/main/java/org/apache/fluss/server/coordinator/CoordinatorEventProcessor.java
hanliu0830: Updated in 33a1b0514. Both illegal AdjustIsr paths now log at ERROR before rejecting the request.

fluss-server/src/main/java/org/apache/fluss/server/replica/Replica.java
hanliu0830: Fixed in 33a1b0514. `maybeExpandISr` now reads and revalidates `isrState` under `leaderIsrUpdateLock`, and prepares the expansion from that locked `CommittedIsrState`. `prepareIsrExpand` also builds the new ISR from the same `currentState`.

fluss-server/src/test/java/org/apache/fluss/server/replica/AdjustIsrTest.java
hanliu0830: Updated in 33a1b0514. The race test now uses a dedicated fixed thread pool with three threads and waits on `hasQueuedThread(thread)` for each task instead of relying on the common pool or `getQueueLength()`.

fluss-server/src/test/java/org/apache/fluss/server/coordinator/CoordinatorEventProcessorTest.java
hanliu0830: Done in 33a1b0514. The invalid request cases are now separate tests: `testProcessAdjustIsrRejectsRequestFromNonLeader` and `testProcessAdjustIsrRejectsIsrWithoutCurrentLeader`, while the happy-path test remains focused on successful adjustment.

swuferhong (approved):
LGTM +1

wuchong (approved):
LGTM. I appended a minor commit to avoid using Java reflection.

PR #3922, opened 2026-08-10
[server] Ignore lake snapshot refresh failures during leader election

### LINKED ISSUE #3919: [server] Prevent lake snapshot refresh failures from blocking leader election
## Search before asking

## Fluss version

main (development)

## Please describe the bug 🐞

When a data-lake-enabled replica becomes leader, `ReplicaManager.makeLeaders` calls `updateWithLakeTableSnapshot` after `replica.makeLeader`. The method reads the latest table-level lake offsets file referenced by ZooKeeper.

A normal lake snapshot commit updates the ZooKeeper pointer and then deletes offsets files for discarded snapshots. A tablet server can already have read the previous ZooKeeper value when that previous offsets file is deleted. Opening the stale path then fails with `FileNotFoundException` or an object-store `404 NoSuchKey`.

The exception currently propagates to `makeLeaders`, so an optional lake snapshot refresh causes the whole leader transition to be reported as failed:

```text
ERROR ReplicaManager - Error make replica TableBucket{...} to leader
java.io.IOException: Read from .../metadata/<uuid>.offsets
with error message: 404 Not Found
<Code>NoSuchKey</Code>
```

A production occurrence showed a Paimon commit finishing about one second before the leader transition failed, which is consistent with the old offsets file being discarded concurrently.

Expected behavior: failure to initialize the cached lake snapshot ID and lake log end offset must not fail leader election. The regular lake offset notification path can update these values after the next successful synchronization.

Actual behavior: the replica has already executed `makeLeader`, but the subsequent lake snapshot read failure makes the leader request return an error.

Reproduction outline:

1. Register a data-lake-enabled table and a lake snapshot metadata entry that references an offsets file.
2. Delete that offsets file, simulating concurrent cleanup of a discarded snapshot.
3. Make one of the table replicas leader.
4. Observe that the leader transition returns an error because `updateWithLakeTableSnapshot` propagates the read failure.

## Solution

Make `updateWithLakeTableSnapshot` best-effort. Catch failures while reading or applying the lake snapshot, log a warning with the table bucket, and allow `makeLeaders` to continue starting leader services. Add a regression test verifying that a missing offsets file does not fail leader election.

luoyuxia:
### Purpose

Linked issue: close #3919

A lake table commit may clean an old offsets file while a TabletServer is still reading snapshot metadata that references it. The resulting read failure currently rejects the replica leader transition. Lake snapshot initialization is best-effort state that will be refreshed by the next synchronization, so it should not block leader election.

### Brief change log

- Catch and log failures from `updateWithLakeTableSnapshot` instead of propagating them through `makeLeaders`.
- Add a regression test with lake snapshot metadata that references a missing offsets file and verify that the replica still becomes leader.

### Tests

- `./mvnw -o -pl fluss-server -DskipITs -Dfast -Dtest=ReplicaManagerTest#testLakeSnapshotReadFailureDoesNotFailLeaderTransition test` (passed; also ran Checkstyle and Spotless with 0 violations)
- `./mvnw -o -pl fluss-server -DskipITs -Dfast -Dtest=ReplicaManagerTest test` (attempted; unrelated tests were blocked because the local disk usage was about 90%, above the configured 85% write-lock threshold)

### API and Format

### Documentation

No documentation changes; this is a bug fix.

### DIFF
--- a/fluss-server/src/main/java/org/apache/fluss/server/replica/ReplicaManager.java
+++ b/fluss-server/src/main/java/org/apache/fluss/server/replica/ReplicaManager.java
@@ -1286,17 +1286,21 @@ public class ReplicaManager implements ServerReconfigurable {
     }
 
     // NOTE: This method can be removed when fetchFromLake is deprecated
-    private void updateWithLakeTableSnapshot(Replica replica) throws Exception {
+    private void updateWithLakeTableSnapshot(Replica replica) {
         TableBucket tb = replica.getTableBucket();
-        Optional<LakeTableSnapshot> optLakeTableSnapshot =
-                zkClient.getLakeTableSnapshot(replica.getTableBucket().getTableId(), null);
-        if (optLakeTableSnapshot.isPresent()) {
-            LakeTableSnapshot lakeTableSnapshot = optLakeTableSnapshot.get();
-            long snapshotId = optLakeTableSnapshot.get().getSnapshotId();
-            replica.getLogTablet().updateLakeTableSnapshotId(snapshotId);
-            lakeTableSnapshot
-                    .getLogEndOffset(tb)
-                    .ifPresent(replica.getLogTablet()::updateLakeLogEndOffset);
+        try {
+            Optional<LakeTableSnapshot> optLakeTableSnapshot =
+                    zkClient.getLakeTableSnapshot(replica.getTableBucket().getTableId(), null);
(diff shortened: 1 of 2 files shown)


beryllw:
LGTM!

swuferhong (approved):
LGTM +1

PR #3930, opened 2026-08-10
[flink] Spill cdc log to rocksDB for pk batch read in case of OOM.

### LINKED ISSUE #3657: Optimize log data buffering using RocksDB for deduplication and sorting
### Search before asking

### Motivation

In both the `KvSnapshotAndLogBatchScanner` proposed in Sub-Issue 1 and the existing [`LakeSnapshotAndLogSplitScanner`](https://github.com/apache/fluss/blob/main/fluss-client/src/main/java/org/apache/fluss/client/table/scanner/batch/LakeSnapshotAndLogSplitScanner.java), incremental log records are buffered in an in-memory `TreeMap<InternalRow, KeyValueRow>`:

```java
// LakeSnapshotAndLogSplitScanner.java, line 67
private Map<InternalRow, KeyValueRow> logRows;
// ...
logRows = new TreeMap<>(rowComparator);
```

**Problem:** The volume of logs between two KV snapshots can be very large. With the default `kv.snapshot.interval` of 10 minutes, a high-throughput workload (e.g., 100 MB/s) can produce tens of GB of log data within a single interval. Buffering all of this in an in-memory `TreeMap` leads to OOM.

### Solution

Replace the in-memory `TreeMap` with a local RocksDB instance to buffer, deduplicate, and sort log records:

1. **Write:** Each fetched log record is written to a local temporary RocksDB as `key → value`
2. **Deduplication:** Subsequent writes for the same key automatically overwrite prior records (RocksDB's `put` semantics provide natural deduplication)
3. **Deletion:** DELETE and UPDATE_BEFORE records are written as **tombstones** to preserve the delete semantics and prevent data loss
4. **Sorting:** RocksDB stores keys in sorted order; reading via `RocksIterator` yields records sorted by primary key
5. **Merge:** The RocksDB log iterator is passed to `SortMergeReader`, which performs a two-way merge with the snapshot's `SnapshotFilesReader`

```
┌──────────────────────────────────────────────────────────┐
│                  Client Side                             │
│                                                          │
│  ┌─────────────┐     ┌──────────────┐                   │
│  │ KV Snapshot │     │  LogScanner  │                   │
│  │  SST Files  │     │ (incremental)│                   │
│  │  (download) │     │   (fetch)    │                   │
│  └──────┬──────┘     └──────┬───────┘                   │
│         │                   │                            │
│         ▼                   ▼                            │
│  ┌─────────────┐     ┌──────────────┐                   │
│  │SnapshotFiles│     │ Log Buffer   │                   │
│  │   Reader    │     │ (RocksDB)    │                   │
│  │ (RocksDB)   │     │ (dedup+sort) │                   │
│  └──────┬──────┘     └──────┬───────┘                   │
│         │                   │                            │
│         ▼                   ▼                            │
│  ┌──────────────────────────────────┐                   │
│  │       SortMergeReader            │                   │
│  │  (snapshot + log merge by PK)    │                   │
│  └──────────────┬───────────────────┘                   │
│                 │                                        │
│                 ▼                                        │
│         Sorted InternalRow stream                        │
└──────────────────────────────────────────────────────────┘
```

### Anything else?

_No response_

### Willingness to contribute

loserwang1024:
### Purpose

Linked issue: close https://github.com/apache/fluss/issues/3657

### DIFF
--- a/fluss-client/src/main/java/org/apache/fluss/client/table/scanner/batch/KvSnapshotAndLogBatchScanner.java
+++ b/fluss-client/src/main/java/org/apache/fluss/client/table/scanner/batch/KvSnapshotAndLogBatchScanner.java
@@ -23,15 +23,12 @@ import org.apache.fluss.client.table.Table;
 import org.apache.fluss.client.table.scanner.ScanRecord;
 import org.apache.fluss.client.table.scanner.SortMergeReader;
 import org.apache.fluss.client.table.scanner.log.LogScanner;
-import org.apache.fluss.client.table.scanner.log.ScanRecords;
 import org.apache.fluss.memory.MemorySegment;
 import org.apache.fluss.metadata.Schema;
 import org.apache.fluss.metadata.TableBucket;
 import org.apache.fluss.metadata.TableInfo;
-import org.apache.fluss.record.ChangeType;
 import org.apache.fluss.record.LogRecord;
 import org.apache.fluss.row.InternalRow;
-import org.apache.fluss.row.KeyValueRow;
 import org.apache.fluss.row.encode.KeyEncoder;
 import org.apache.fluss.types.RowType;
 import org.apache.fluss.utils.CloseableIterator;
@@ -43,10 +40,9 @@ import java.io.IOException;
 import java.io.UncheckedIOException;
 import java.time.Duration;
 import java.util.Comparator;
-import java.util.Map;
 import java.util.NoSuchElementException;
-import java.util.TreeMap;
 
+import static org.apache.fluss.client.table.scanner.batch.SortedLogRows.DEFAULT_SPILL_THRESHOLD;
 import static org.apache.fluss.utils.Preconditions.checkArgument;
 
 /**
@@ -55,17 +51,14 @@ import static org.apache.fluss.utils.Preconditions.checkArgument;
 @Internal
 public class KvSnapshotAndLogBatchScanner implements BatchScanner {
 
-    private final TableBucket tableBucket;
-    private final long logStoppingOffset;
     private final int[] keyIndexesInScanRow;
     @Nullable private final int[] adjustProjectedFields;
     private final Comparator<InternalRow> primaryKeyComparator;
-    private final Map<InternalRow, KeyValueRow> logRows;
+    @Nullable private final SortedLogRows logRows;
 
     @Nullable private final BatchScanner snapshotScanner;
-    @Nullable private final LogScanner logScanner;
 
-    private boolean logScanFinished;
+    private boolean logRowsLoaded;
     private boolean finished;
     private boolean closed;
 
@@ -79,18 +72,37 @@ public class KvSnapshotAndLogBatchScanner implements BatchScanner {
             long snapshotId,
             long logStartingOffset,
             long logStoppingOffset,
-            @Nullable int[] projectedFields) {
+            @Nullable int[] projectedFields,
+            String scannerTmpDir) {
+        this(
+                table,
+                tableBucket,
+                snapshotId,
+                logStartingOffset,
+                logStoppingOffset,
+                projectedFields,
+                scannerTmpDir,
+                DEFAULT_SPILL_THRESHOLD);
+    }
+
+    @VisibleForTesting
+    KvSnapshotAndLogBatchScanner(
+            Table table,
+            TableBucket tableBucket,
+            long snapshotId,
+            long logStartingOffset,
+            long logStoppingOffset,
+            @Nullable int[] projectedFields,
+            String scannerTmpDir,
+            int spillThreshold) {
         checkArgument(
                 table.getTableInfo().hasPrimaryKey(),
                 "KvSnapshotAndLogBatchScanner only supports primary-key tables.");
-        this.tableBucket = tableBucket;
-        this.logStoppingOffset = logStoppingOffset;
 
         ProjectionPlan projectionPlan = createProjectionPlan(table.getTableInfo(), projectedFields);
         this.keyIndexesInScanRow = projectionPlan.keyIndexesInScanRow;
         this.adjustProjectedFields = projectionPlan.adjustProjectedFields;
         this.primaryKeyComparator = createPrimaryKeyComparator(table.getTableInfo());
-        this.logRows = new TreeMap<>(primaryKeyComparator);
 
         this.snapshotScanner =
                 snapshotId >= 0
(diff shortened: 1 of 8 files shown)


loserwang1024:
@naivedogger @beryllw @leonardBang , CC

fluss-client/src/main/java/org/apache/fluss/client/table/scanner/batch/SortedLogRows.java
```diff
@@ -0,0 +1,391 @@
...
+                changeType == ChangeType.DELETE || changeType == ChangeType.UPDATE_BEFORE;
+        put(scanRecord.getRow(), isDelete);
+    }
+
+    private void put(InternalRow row, boolean isDelete) throws IOException {
+        checkNotClosed();
+
+        if (rocksDBHandle == null) {
+            putToMemory(row, isDelete);
+            if (memoryRows.size() > spillThreshold) {
```
naivedogger, line 170: Could we use a serialized-byte threshold instead of the number of distinct keys here? The goal is to prevent OOM, but memoryRows.size() does not provide an upper bound on memory usage because row sizes can vary significantly. For example, 8192 rows of 32 KiB already retain about 256 MiB of row data, excluding TreeMap and object overhead, so the process may OOM before spilling. It would be safer to track the retained serialized bytes and spill based on a configurable byte-size limit. Also, could you clarify how the default value of 8192 was determined?

fluss-client/src/main/java/org/apache/fluss/client/table/scanner/batch/KvSnapshotAndLogBatchScanner.java
```diff
@@ -203,7 +203,7 @@ public void close() throws IOException {
         IOUtils.closeQuietly(sortMergeIterator);
         IOUtils.closeQuietly(snapshotRecordIterator);
         IOUtils.closeQuietly(snapshotScanner);
-        IOUtils.closeQuietly(logScanner);
+        IOUtils.closeQuietly(logRows);
```
naivedogger, line 207: Closing `logRows` here does not appear to close the `RocksDBLogRowsIterator` that has already been passed to `SortMergeReader`.

The same path appears to exist in `LakeSnapshotAndLogSplitScanner`. Would it make sense for `SortMergeReader` to implement `Closeable`, own its input iterators, and be explicitly closed by the scanners before `logRows`?

fluss-client/src/main/java/org/apache/fluss/client/table/scanner/batch/LakeSnapshotAndLogSplitScanner.java
```diff
@@ -176,38 +182,32 @@ public boolean requireSortedRecords() {
...
-        try {
-            if (logScanner != null) {
-                logScanner.close();
-            }
-            if (lakeRecordIterators != null) {
-                for (CloseableIterator<LogRecord> iterator : lakeRecordIterators) {
-                    iterator.close();
-                }
+        IOUtils.closeQuietly(currentSortMergeIterator);
+        IOUtils.closeQuietly(logRows);
```
naivedogger: Ditto

naivedogger (commented):
Thanks @loserwang1024 for this PR! Left some comments, PTAL.

beryllw:
This workload is essentially append-once and scan-once, requiring only a single ordered traversal. Could you clarify why RocksDB was chosen over a lighter spillable external sort?

fluss-client/src/main/java/org/apache/fluss/client/table/scanner/batch/SortedLogRows.java
```diff
@@ -0,0 +1,391 @@
...
+        }
+
+        @Override
+        public String name() {
+            return "fluss-sorted-log-rows-comparator";
+        }
+
+        @Override
+        public int compare(ByteBuffer left, ByteBuffer right) {
+            return rowComparator.compare(
```
beryllw: RocksDB may invoke custom comparators concurrently, while the KV comparator ultimately shares a KeyEncoder backed by a mutable writer. Could we use order-preserving encoded keys with RocksDB’s bytewise comparator, or otherwise ensure that the entire comparator chain is thread-safe?

beryllw:
@luoyuxia cc

fluss-client/src/main/java/org/apache/fluss/client/table/scanner/batch/SortedLogRows.java
loserwang1024: It is difficult to accurately estimate memory usage from the in-memory objects. We could use the size of the value after converting it to a `byte[]`, but that would require eagerly serializing the value before spilling, and the resulting size would still not accurately represent the actual heap usage.

For now, we prefer not to expose the spill threshold as a configuration option because this is still an internal implementation detail, and we do not want to introduce a parameter tied to the current implementation prematurely. We can expose an appropriate configuration later if a concrete need arises.

The current threshold of 8192 was chosen with reference to the maximum limit-query value of 1024, with additional headroom. Its original purpose was also to prevent an unexpectedly large amount of data from accumulating in memory.

fluss-client/src/main/java/org/apache/fluss/client/table/scanner/batch/KvSnapshotAndLogBatchScanner.java
loserwang1024: > Closing logRows here does not appear to close the RocksDBLogRowsIterator that has already been passed to SortMergeReader.

RocksDBLogRowsIterator is passed to `org.apache.fluss.client.table.scanner.SortMergeReader.MergeIterator` and will be closed here.

fluss-client/src/main/java/org/apache/fluss/client/table/scanner/batch/SortedLogRows.java
loserwang1024: Thanks for raising this. RocksDB was chosen primarily to simplify the implementation, rather than because this workload inherently requires a general-purpose KV store. KV snapshot reading already uses RocksDB, so we can reuse the existing RocksDB dependency and lifecycle utilities instead of introducing and maintaining a separate spillable external-sort implementation. `SortedLogRows` is also an internal implementation detail, so we can replace it with a dedicated spill mechanism later if benchmarks show that RocksDB becomes a bottleneck.

That said, your concern about the JNI comparator overhead is valid. I plan to limit the RocksDB-backed implementation to `KvSnapshotAndLogBatchScanner` and stop using it in `LakeSnapshotAndLogSplitScanner`, for two reasons:

1. The default KV snapshot interval is 10 minutes, while the default lake freshness is 3 minutes. Therefore, the changelog range accumulated between KV snapshots is typically larger, making the KV snapshot path more susceptible to OOM.

2. The KV snapshot primary-key order is already defined by an unsigned bytewise comparison of the encoded primary-key bytes, which is consistent with RocksDB’s default bytewise comparator. We can therefore use the encoded primary-key bytes directly as the RocksDB key and remove the custom Java comparator entirely, avoiding the native-to-Java comparator callbacks. In contrast, a lake snapshot’s ordering is provided by its lake-specific `SortedRecordReader` and is not necessarily consistent with RocksDB’s bytewise ordering.

This keeps the spill optimization focused on the path with the greater OOM risk while also simplifying the RocksDB key and comparator implementation.

fluss-client/src/main/java/org/apache/fluss/client/table/scanner/batch/SortedLogRows.java
leonardBang: Thanks for the discussion. I think the requirement here can be stated more precisely: the current count-based threshold does not bound retained heap, so a valid wide-row workload can exhaust heap before spill; this does not mean every deployment will necessarily OOM.

I ran a small-heap reproduction against the current revision using the actual `SortedLogRows.load()` path: `-Xmx128m`, 64 KiB values, 100 distinct keys per poll, and the production threshold of 8192. The JVM ran out of heap after approximately 1500 generated records while `isSpilled()` was still false. The direct stack was `IndexedRow.copy -> SortedLogRows.copyRow -> putToMemory -> load`.

Since `putToMemory()` already materializes a copied `BinaryRow`, would it make sense to track a retained-byte estimate such as `encodedKey.length + copiedRow.getSizeInBytes()`, subtract the previous value when replacing the same key, and trigger spill based on that byte limit? It may also help to release entries incrementally during migration and add a small-heap/large-row regression test.

fluss-client/src/main/java/org/apache/fluss/client/table/scanner/batch/KvSnapshotAndLogBatchScanner.java
leonardBang: The non-empty successful path does close the changelog iterator through the returned `MergeIterator`, but there still seems to be an ownership gap when `readBatch()` produces no output.

For example, in a supported batch path where the spilled changelog consists entirely of tombstones, or every snapshot row is deleted, `MergeIterator.hasNext()` drains the inputs and returns false. `SortMergeReader.readBatch()` therefore returns null, `sortMergeIterator` remains null, and scanner close proceeds to close `logRows`/RocksDB without explicitly closing `RocksDBLogRowsIterator`. The same gap exists if the first `hasNext()` throws. I reproduced the zero-output case with a tracking iterator: `readBatch()` returned null while the changelog iterator's `close()` had not been called.

Would it make sense for `SortMergeReader` to implement `Closeable` and own both input iterators, with the scanner closing it before `logRows`? A regression test covering all-tombstone output and the first-`hasNext()` exception path would help verify the ownership.

PR #3934, opened 2026-08-10
[rust] add support of iceberg format key encoder

zuston:
### Purpose

add support of iceberg format key encoder

### DIFF
--- a/fluss-rust/crates/fluss/src/row/encode/iceberg_key_encoder.rs
+++ b/fluss-rust/crates/fluss/src/row/encode/iceberg_key_encoder.rs
@@ -1,0 +1,210 @@
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+use bytes::Bytes;
+
+use crate::error::Error::IllegalArgument;
+use crate::error::Result;
+use crate::metadata::RowType;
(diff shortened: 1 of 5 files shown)


zuston:

fluss-rust/crates/fluss/src/row/encode/mod.rs
```diff
@@ -111,9 +113,9 @@ impl KeyEncoderFactory {
             Some(DataLakeFormat::Lance) => Ok(Box::new(CompactedKeyEncoder::create_key_encoder(
                 row_type, key_fields,
             )?)),
-            Some(DataLakeFormat::Iceberg) => Err(Error::UnsupportedOperation {
-                message: "KeyEncoder for Iceberg format is not yet implemented".to_string(),
-            }),
+            Some(DataLakeFormat::Iceberg) => {
```
fresh-borzoni, line 116: It seems append-only Iceberg tables bucket through here too.  Can you add an append IT then?

fluss-rust/crates/fluss/src/row/encode/iceberg_key_encoder.rs
```diff
@@ -0,0 +1,210 @@
...
+        let row_type = RowType::with_data_types_and_field_names(vec![DataTypes::int()], vec!["id"]);
+        let error = match IcebergKeyEncoder::new(&row_type, &["missing".to_string()]) {
+            Ok(_) => panic!("expected IllegalArgument"),
+            Err(error) => error,
+        };
+        assert!(error.to_string().contains("not found in input row type"));
+    }
+
+    #[test]
+    fn encodes_values_with_iceberg_binary_layout() {
```
fresh-borzoni: These check the bytes against values copied from our Java code, not from Iceberg — and the IT can't help since write and read share the encoder.

Could you assert Iceberg's Appendix B hash vectors instead?  Run them through IcebergBucketingFunction with num_buckets = i32::MAX so the modulo cancels and you get the raw hash back. Negative ones need hash & i32::MAX accordingly.

Let me know if you need more clarifications.

fresh-borzoni (commented):
@zuston Thank you for the PR, LGTM over, some minor comments and suggestions, PTAL

fluss-rust/crates/fluss/src/row/encode/mod.rs
zuston: added

fluss-rust/crates/fluss/src/row/encode/iceberg_key_encoder.rs
zuston: really insightful comment! updated

fresh-borzoni (approved):
@zuston Thank you, LGTM 👍

PR #3935, opened 2026-08-10
[rust] ignore missing partitions during scanner metadata updates

zuston:
### Purpose

ignore missing partitions during scanner metadata updates

### DIFF
--- a/fluss-rust/crates/fluss/src/client/table/scanner.rs
+++ b/fluss-rust/crates/fluss/src/client/table/scanner.rs
@@ -1599,14 +1599,33 @@ impl LogFetcher {
 
                     let error = FlussError::for_code(error_code);
                     if Self::should_invalidate_table_meta(error) {
-                        // TODO: Consider triggering table meta invalidation from sender/lookup paths.
                         let table_id = table_bucket.table_id();
                         let cluster = metadata.get_cluster();
                         if let Some(table_path) = cluster.get_table_path_by_id(table_id) {
-                            let physical_tables = HashSet::from([PhysicalTablePath::of(Arc::new(
-                                table_path.clone(),
-                            ))]);
-                            metadata.invalidate_physical_table_meta(&physical_tables);
+                            let physical_table_path = match table_bucket.partition_id() {
+                                Some(partition_id) => {
+                                    match cluster.get_partition_name(partition_id) {
+                                        Some(partition_name) => {
+                                            Some(PhysicalTablePath::of_partitioned(
+                                                Arc::new(table_path.clone()),
+                                                Some(partition_name.clone()),
+                                            ))
+                                        }
+                                        None => {
+                                            warn!(
+                                                "Partition id {partition_id} is missing from partition_name_by_id while invalidating metadata for table {table_path}"
+                                            );
+                                            None
+                                        }
+                                    }
+                                }
+                                None => Some(PhysicalTablePath::of(Arc::new(table_path.clone()))),
+                            };
+                            if let Some(physical_table_path) = physical_table_path {
+                                metadata.invalidate_physical_table_meta(&HashSet::from([
+                                    physical_table_path,
+                                ]));
+                            }
                         } else {
                             warn!(
                                 "Table id {table_id} is missing from table_path_by_id while invalidating table metadata"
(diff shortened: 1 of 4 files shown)


fluss-rust/crates/fluss/src/client/table/scanner.rs
```diff
@@ -1415,15 +1415,21 @@ impl LogFetcher {
             Ok(())
         };
 
-        // TODO: Handle PartitionNotExist error like java side
-        update_result.or_else(|e| {
-            if let Error::RpcError { source, .. } = &e
+        update_result.or_else(|error| {
+            if matches!(error.api_error(), Some(FlussError::PartitionNotExists)) {
```
fresh-borzoni, line 1429: I think Java gets away with ignoring this because it already rejected bad partition ids much earlier, at subscribe time — LogScannerImpl puts the partition id on the wire so the server throws right there. 

We never added that check on the Rust side, so the first poll() has been the only place a user finds out. This change takes that away too, and then a wrong partition id just means poll() returns empty forever with nothing logged above trace!

Should `subscribe_partition`  and `do_subscribe_buckets ` pass the partition ids to the metadata update, the way Java's checkAndUpdatePartitionMetadata does?

fluss-rust/crates/fluss/src/client/table/scanner.rs
fresh-borzoni, line 1416: Separate thing I noticed while reading: we batch every leaderless bucket into one request, and the server rejects the whole request as soon as it hits a missing id, so nothing gets refreshed at all. So if P1 gets dropped and later P2's leader moves, P2 is stuck behind P1's dead id and never recovers? I think it's the same for Java part, mb worth a follow-up?

fluss-rust/crates/fluss/src/client/table/scanner.rs
fresh-borzoni, line 1429: Any chance of a test? The PR should add an IT that drops one of two subscribed partitions and asserts the survivor keeps draining

fresh-borzoni (commented):
@zuston Thank you for the PR, left some comments, PTAL

fluss-rust/crates/fluss/src/client/table/scanner.rs
zuston: Agreed, this looks like a separate pre-existing issue shared with Java

fluss-rust/crates/fluss/src/client/table/scanner.rs
zuston: > Should subscribe_partition and do_subscribe_buckets  pass the partition ids to the metadata update, the way Java's checkAndUpdatePartitionMetadata does?

good idea. updated

zuston: added.

fluss-rust/crates/fluss/src/client/table/scanner.rs
fresh-borzoni: I studied this more, Java invalidates the partition path with exact-key matching and we invalidate the table path and filter on get_table_path(), so one dead partition drops every partition's leaders. The next refresh then carries the dropped id alongside the live ones and the server rejects the whole request, and with the error swallowed the scanner just returns empty forever

Can we scope the invalidation to the partition here? Then the existing test can append to the survivor after the drop and actually assert progress. Needs both sides though: pass the partition name into the path at scanner.rs:1606, and stop invalidate_physical_table_meta collapsing it back to the table path at cluster.rs:111.

fresh-borzoni (commented):
@zuston Thank you, left one additional comment, PTAL

fluss-rust/crates/fluss/src/client/table/scanner.rs
zuston: Nice catch. Ideally, we could also take the error type into account when determining the invalidation scope. That said, using only the partition name for fine-grained invalidation should not be a problem.

fresh-borzoni (approved):
@zuston Thank you, LGTM, now it matches Java's behaviour, so it's fine to merge

But as I already noticed -  dropped partition stays in the subscription forever, so every poll() re-asks the server for the dead id and swallows the error.
Could we file a follow-up, please?

Simplest fix is probably to remember the dead ids and stop putting them in the refresh batch, that also stops one dead id from dragging a live partition's refresh down with it.  WDYT?

zuston:
> @zuston Thank you, LGTM, now it matches Java's behaviour, so it's fine to merge
> 
> But as I already noticed - dropped partition stays in the subscription forever, so every poll() re-asks the server for the dead id and swallows the error. Could we file a follow-up, please?
> 
> Simplest fix is probably to remember the dead ids and stop putting them in the refresh batch, that also stops one dead id from dragging a live partition's refresh down with it. WDYT?

sure. let me fix this.

PR #3937, opened 2026-08-10
[server] Refine remote log manifest overlap handling

wuchong:
### Purpose

Follow up on the remote log overlap manifest handling introduced by #3822 and resolve the remaining implementation follow-ups.

### Brief change log

- Keep manifest start/end getters aligned with physical remote segment offsets.
- Make immutable remote manifest snapshot reads lock-free.
- Remove the obsolete incremental RemoteLogTablet mutation helper.
- Clarify timestamp lookup behavior and update overlap regression coverage.

### Tests

- `./mvnw -pl fluss-common -DskipITs -Dfast -Dtest=RemoteLogManifestOverlapTest test` (11 tests)
- `./mvnw -pl fluss-server -am -DskipITs -Dfast -Dtest=RemoteLogTabletTest,RemoteLogTabletOverlapTest,DefaultRemoteLogStorageTest -Dsurefire.failIfNoSpecifiedTests=false test` (22 tests)
- Checkstyle, Spotless, and RAT checks passed in the reactor build.

### API and Format

- Remote manifest start/end getters now consistently report physical remote offsets.
- No storage format changes.

### Documentation

No user-facing documentation changes are required.

### Generative AI disclosure

- [x] Yes — Codex (gpt-5.6-sol), reviewed by the human developer.

### DIFF
--- a/fluss-common/src/main/java/org/apache/fluss/remote/RemoteLogManifest.java
+++ b/fluss-common/src/main/java/org/apache/fluss/remote/RemoteLogManifest.java
@@ -145,8 +145,8 @@ public class RemoteLogManifest {
     public long getRemoteLogStartOffset() {
         long startOffset = Long.MAX_VALUE;
         for (RemoteLogSegment remoteLogSegment : remoteLogSegmentList) {
-            if (remoteLogSegment.logicalStartOffset() < startOffset) {
-                startOffset = remoteLogSegment.logicalStartOffset();
+            if (remoteLogSegment.remoteLogStartOffset() < startOffset) {
+                startOffset = remoteLogSegment.remoteLogStartOffset();
             }
         }
         return startOffset;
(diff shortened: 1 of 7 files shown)


wuchong:
cc @gyang94

fluss-common/src/main/java/org/apache/fluss/remote/RemoteLogManifest.java
```diff
@@ -145,8 +145,8 @@ public RemoteLogManifest trimAndMerge(
     public long getRemoteLogStartOffset() {
         long startOffset = Long.MAX_VALUE;
         for (RemoteLogSegment remoteLogSegment : remoteLogSegmentList) {
-            if (remoteLogSegment.logicalStartOffset() < startOffset) {
-                startOffset = remoteLogSegment.logicalStartOffset();
+            if (remoteLogSegment.remoteLogStartOffset() < startOffset) {
```
gyang94, line 148: Sometimes the physical range is not equal to logical range. Use physical range may expose unavailable data to uers.

Assume:
1. There is a remote manifest contains A:[0,10), B:[10,20)
2. A expired, the manifest visible range is [10,20)
3. leader switch, new physical segment is [5,25)
4. After merge, there is only one segment should be with physical range [5,25), but logical range should be [10,25).

The hidden prefix [5,10) should not be readable again.

wuchong: Thank you @gyang94 , I think we can create a follow-up pull request to address this.

PR #3940, opened 2026-08-10
[server] Support hot update for active segment retention roll

### LINKED ISSUE #3939: Support hot update for active segment retention roll
### Search before asking

### Fluss version

main (development)

### Please describe the bug 🐞

#3831 introduced `log.retention.roll-active-segment.enabled` to control whether TTL cleanup rolls an expired active log segment. However, the option is read into a per-`LogTablet` boolean when the tablet is created and is not registered as a dynamically reconfigurable server option. Changing the cluster configuration at runtime therefore does not affect existing tablets.

The implementation also calls `roll()` from `LogTablet#deletableExpiredSegments`. This method is used as a finder for deletable segments and is expected to only compute and return a result. Rolling the active segment mutates log state, so the side effect breaks the method semantics and makes the cleanup flow harder to reason about.

Expected behavior:

- Updating `log.retention.roll-active-segment.enabled` through dynamic server configuration immediately affects existing and newly created tablets.
- Methods that find deletable segments remain side-effect free; active-segment rolling is performed by an explicit cleanup action.

### Solution

- Register `log.retention.roll-active-segment.enabled` as a dynamic server option.
- Let `LogManager` own a shared `AtomicBoolean`, implement `ServerReconfigurable`, and pass the shared value to every `LogTablet`.
- Separate tiered-storage cleanup from TTL cleanup, move active-segment rolling into the TTL cleanup action, and keep `deletableExpiredSegments` read-only.
- Add dynamic-configuration and retention regression tests.

wuchong:
### Purpose

Linked issue: close #3939

#3831 introduced `log.retention.roll-active-segment.enabled`, but the value is cached in each `LogTablet` at construction time and is not wired into dynamic server reconfiguration. Existing tablets therefore do not observe runtime updates.

The TTL cleanup path also invokes `roll()` from `deletableExpiredSegments`. A method used as a deletable-segment finder should only compute a result; mutating log state there obscures the cleanup flow and breaks its read-only semantics.

This change makes the option dynamically reconfigurable and moves active-segment rolling into an explicit TTL cleanup action.

### Brief change log

- Register `log.retention.roll-active-segment.enabled` as a dynamic server option.
- Make `LogManager` implement `ServerReconfigurable` and own a shared `AtomicBoolean` observed by existing and newly created `LogTablet` instances.
- Separate tiered-storage and TTL cleanup actions so each path owns its deletion reason and behavior.
- Keep `deletableExpiredSegments` side-effect free and perform active-segment rolling in `cleanupExpiredSegments`.
- Add dynamic-configuration and retention regression coverage.

### Tests

- `./mvnw -pl fluss-server -am -DskipITs -Dcheckstyle.skip -Drat.skip -Dspotless.check.skip=true -Dtest=DynamicConfigChangeTest,LogManagerTest,LogTabletTest,LocalSegmentTTLTest,TieredLocalSegmentTTLTest -Dsurefire.failIfNoSpecifiedTests=false test`
  - 80 tests passed.
- `./mvnw -pl fluss-server spotless:check`

### API and Format

No public API, RPC protocol, or storage format changes. The existing server configuration option now supports dynamic updates.

### Documentation

### DIFF
--- a/fluss-server/src/main/java/org/apache/fluss/server/DynamicServerConfig.java
+++ b/fluss-server/src/main/java/org/apache/fluss/server/DynamicServerConfig.java
@@ -50,6 +50,7 @@ import static org.apache.fluss.config.ConfigOptions.KV_LEADER_REPLICA_MEMORY_RES
 import static org.apache.fluss.config.ConfigOptions.KV_SHARED_RATE_LIMITER_BYTES_PER_SEC;
 import static org.apache.fluss.config.ConfigOptions.KV_SNAPSHOT_INTERVAL;
 import static org.apache.fluss.config.ConfigOptions.LOG_REPLICA_MIN_IN_SYNC_REPLICAS_NUMBER;
+import static org.apache.fluss.config.ConfigOptions.LOG_RETENTION_ROLL_ACTIVE_SEGMENT_ENABLED;
 import static org.apache.fluss.config.ConfigOptions.REMOTE_DATA_DIRS;
 import static org.apache.fluss.config.ConfigOptions.REMOTE_DATA_DIRS_STRATEGY;
 import static org.apache.fluss.config.ConfigOptions.REMOTE_DATA_DIRS_WEIGHTS;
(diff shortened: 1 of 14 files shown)


wuchong:
cc @gyang94 @fxbing

fxbing (approved):
LGTM

gyang94 (approved):
LGTM

PR #3942, opened 2026-08-11
[server] Clarify remote log start offset semantics

### LINKED ISSUE #3943: Remote log start offset may expose an expired physical prefix
### Search before asking

### Fluss version

main (development)

### Please describe the bug 🐞

After #3937, the remote log start offset derived from a manifest can use the physical start offset of the underlying remote object instead of the start of the manifest-visible logical range.

When an overlapping replacement object starts before the current logical range, propagating its physical start offset can make the tablet report an already expired prefix as readable.

### How to reproduce

Start with a manifest whose visible range begins at offset `10`, then merge a replacement remote object whose physical range is `[5, 25)`. The replacement is retained with logical range `[10, 25)`, but using the physical object start reports remote log start offset `5`.

### Expected behavior

The manifest and tablet should report logical start offset `10`. Code that accesses the underlying remote object should explicitly use physical start offset `5`.

A fix is proposed in #3942. The fix only changes start-offset semantics and naming; remote log end-offset behavior remains unchanged.

### Solution

Keep manifest-visible start offsets logical, and rename the underlying remote object start-offset API to `physicalStartOffset`. Propagate explicit logical start-offset names through the manifest, coordinator, and tablet call chain.

gyang94:
### Purpose

Close #3943.

Keep the remote log start offset reported by the manifest and tablet aligned with the manifest-visible logical range. A replacement object may physically start before the current visible range; for example, a physical object covering `[5, 25)` can be visible only from offset `10`. Reporting physical offset `5` would incorrectly expose an expired prefix.

### Brief change log

- return the logical start offset from `RemoteLogManifest`
- rename physical segment start-offset APIs to `physicalStartOffset`
- rename propagated start-offset fields and variables so logical and physical semantics are explicit
- leave all remote log end-offset APIs and semantics unchanged
- add regression coverage for a physical start offset preceding the manifest-visible logical start

### Tests

- `mvn -o -pl fluss-common -DskipITs -Dfast -Dtest=RemoteLogManifestOverlapTest test`
- `mvn -o -pl fluss-server -am -DskipITs -Dfast -Dtest=RemoteLogTabletOverlapTest,RemoteLogTabletTest,RemoteLogManagerTest -Dsurefire.failIfNoSpecifiedTests=false test`
- `mvn -o -pl fluss-client,fluss-server,fluss-flink/fluss-flink-common -am -DskipTests -Dfast package`

### API and Format

This renames internal Java APIs for remote segment physical start offsets. It does not change the RPC wire format, manifest storage format, or remote log end-offset semantics.

### Documentation

No user-facing documentation changes are required.

### Generative AI disclosure

- [x] Yes — Codex (GPT-5)

### DIFF
--- a/fluss-client/src/main/java/org/apache/fluss/client/table/scanner/log/LogFetcher.java
+++ b/fluss-client/src/main/java/org/apache/fluss/client/table/scanner/log/LogFetcher.java
@@ -541,7 +541,7 @@ public class LogFetcher implements Closeable {
             RemoteLogSegment segment = remoteLogSegments.get(i);
             if (i > 0) {
                 posInLogSegment = 0;
-                fetchOffset = segment.remoteLogStartOffset();
+                fetchOffset = segment.physicalStartOffset();
             }
             RemoteLogDownloadFuture downloadFuture =
                     remoteLogDownloader.requestRemoteLog(remoteLogTabletDir, segment);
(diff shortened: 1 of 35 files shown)


wuchong:
Continue with https://github.com/apache/fluss/pull/3956

PR #3947, opened 2026-08-11
[flink] Fix BETWEEN literal extraction in PredicateConverter

### LINKED ISSUE #4000: [flink] BETWEEN predicate fails at runtime after pushdown: literal extraction missing in PredicateConverter
### Search before asking

### Fluss version

0.9.0 (latest release)

### Please describe the bug 🐞

`PredicateConverter` mishandles the Flink `BETWEEN` expression when converting to a Fluss `Predicate`: the lower/upper bounds are passed through as raw `ValueLiteralExpression` AST nodes instead of being extracted to real Java literals:

```java
// before the fix
return builder.between(builder.indexOf(fieldRefExpr.getName()), children.get(1), children.get(2));
This is inconsistent with how EQUALS / IN / LIKE are handled (they all call extractLiteral(...)). The consequence:

toString()-based assertions still pass, because ValueLiteralExpression.toString() prints something like "10", so the converted predicate looks correct (And([GreaterOrEqual(long1, 10), LessOrEqual(long1, 20)])).
But at runtime, predicate.test(row) fails with Unsupported type: BIGINT in CompareUtils.compareLiteral, because the predicate holds Flink AST nodes rather than typed literal values.

### Solution

Extract both bounds with `extractLiteral(fieldRefExpr.getOutputDataType(), ...)` before calling `builder.between(...)`, consistent with the other comparison branches (`EQUALS` / `IN` / `LIKE`). Also add a row-evaluation regression test (`testBetweenEvaluatesAgainstRow`) that calls `predicate.test(GenericRow.of(...))` — the kind of test that would have caught this bug, since the previous parameterized test only compared `toString()` output.


### PR DESCRIPTION
**Pull request by WenDing-Y**
## Summary

Fix `PredicateConverter` handling of Flink `BETWEEN` so lower/upper bounds are real Java literals instead of unextracted `ValueLiteralExpression` nodes.

fix https://github.com/apache/fluss/issues/4000

Previously the converter did:

```java
builder.between(idx, children.get(1), children.get(2));
```

which stored Flink AST nodes as predicate literals. That made `toString()`-based assertions pass (because `ValueLiteralExpression.toString()` looks like `"10"`), while `predicate.test(row)` failed at runtime (`Unsupported type: BIGINT` in `CompareUtils.compareLiteral`).

This change extracts both bounds via `extractLiteral(...)`, consistent with `EQUALS` / `IN` / `LIKE`, and adds a row-evaluation test that would have caught the bug.

## Changes

- **`PredicateConverter`**: extract `BETWEEN` lower/upper bounds with `extractLiteral` before calling `builder.between`.
- **`PredicateConverterTest`**:
  - add `testBetweenEvaluatesAgainstRow` (`predicate.test(GenericRow...)`)
  - align parameterized expected bounds to `10L` / `20L` (BIGINT field + literal cast)

## Why the old test passed

The parameterized case only compared:

```text
converter.visit(BETWEEN).toString()
  ==
BUILDER.between(...).toString()
```

Both printed as something like:

```text
And([GreaterOrEqual(long1, 10), LessOrEqual(long1, 20)])
```

even though one side held `ValueLiteralExpression` objects and the other held numeric literals. String equality hid the type mismatch.

## Test plan

- [x] `./mvnw -pl fluss-flink/fluss-flink-common -am test -Dtest=PredicateConverterTest`
- [x] Confirm `testBetweenEvaluatesAgainstRow` fails on the unfixed code and passes after the fix

### DIFF
--- a/fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/utils/PredicateConverter.java
+++ b/fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/utils/PredicateConverter.java
@@ -130,8 +130,12 @@ public class PredicateConverter implements ExpressionVisitor<Predicate> {
         } else if (func == BuiltInFunctionDefinitions.BETWEEN) {
             FieldReferenceExpression fieldRefExpr =
                     extractFieldReference(children.get(0)).orElseThrow(UnsupportedExpression::new);
+            Object lowerBound =
+                    extractLiteral(fieldRefExpr.getOutputDataType(), children.get(1));
+            Object upperBound =
+                    extractLiteral(fieldRefExpr.getOutputDataType(), children.get(2));
             return builder.between(
-                    builder.indexOf(fieldRefExpr.getName()), children.get(1), children.get(2));
+                    builder.indexOf(fieldRefExpr.getName()), lowerBound, upperBound);
         } else if (func == BuiltInFunctionDefinitions.LIKE) {
             FieldReferenceExpression fieldRefExpr =
                     extractFieldReference(children.get(0)).orElseThrow(UnsupportedExpression::new);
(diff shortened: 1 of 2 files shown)


WenDing-Y:
@wuchong

WenDing-Y:
@fresh-borzoni

PR #3950, opened 2026-08-11
[rust] Validate append_arrow_batch nullability before enqueue

### LINKED ISSUE #3949: [rust] append_arrow_batch accepts nulls in NOT NULL columns and fails later on scan
### Search before asking

### Fluss version

main (development)

### Please describe the bug 🐞

`append_arrow_batch` can accept a caller-provided Arrow `RecordBatch` where a NOT NULL Fluss column contains null values. The write is accepted in some paths and the failure appears later during decode/scan.

The client should reject invalid batches at append time, before enqueue/split/write, with a clear `IllegalArgument` error.

### Solution

Add schema + nullability validation for caller-supplied Arrow batches at append entry, and apply the same validation in prebuilt builder path so all ingest entry points enforce identical invariant

charlesdong1991:
### Purpose

Linked issue: close #3949 

### Brief change log

`append_arrow_batch` can accept a caller-provided Arrow `RecordBatch` where a NOT NULL Fluss column contains null values. The write is accepted in some paths and the failure appears later during decode/scan.

### Tests
- add unit tests and integration tests

### DIFF
--- a/fluss-rust/crates/fluss/src/record/arrow.rs
+++ b/fluss-rust/crates/fluss/src/record/arrow.rs
@@ -25,7 +25,7 @@ use crate::record::ScanRecord;
 use crate::row::column_vector::TypedBatch;
 use crate::row::column_writer::{ColumnWriter, round_up_to_8};
 use crate::row::{ColumnarRow, InternalRow};
-use arrow::array::{ArrayBuilder, ArrayRef, new_null_array};
+use arrow::array::{Array, ArrayBuilder, ArrayRef, new_null_array};
 use arrow::{
     array::RecordBatch,
     buffer::Buffer,
@@ -124,10 +124,20 @@ pub trait ArrowRecordBatchInnerBuilder: Send {
     fn estimated_size_in_bytes(&self) -> usize;
 }
 
-#[derive(Default)]
-pub struct PrebuiltRecordBatchBuilder {
+pub(crate) struct PrebuiltRecordBatchBuilder {
+    row_type: RowType,
     arrow_record_batch: Option<Arc<RecordBatch>>,
     records_count: i32,
+}
+
+impl PrebuiltRecordBatchBuilder {
+    fn new(row_type: RowType) -> Self {
+        Self {
+            row_type,
+            arrow_record_batch: None,
+            records_count: 0,
+        }
+    }
 }
 
 impl ArrowRecordBatchInnerBuilder for PrebuiltRecordBatchBuilder {
@@ -144,6 +154,7 @@ impl ArrowRecordBatchInnerBuilder for PrebuiltRecordBatchBuilder {
         if self.arrow_record_batch.is_some() {
             return Ok(false);
         }
+        validate_append_record_batch(record_batch.as_ref(), &self.row_type)?;
         self.records_count = record_batch.num_rows() as i32;
         self.arrow_record_batch = Some(record_batch);
         Ok(true)
@@ -291,7 +302,7 @@ impl MemoryLogRecordsArrowBuilder {
         } = config;
         let arrow_batch_builder: Box<dyn ArrowRecordBatchInnerBuilder> = {
             if to_append_record_batch {
-                Box::new(PrebuiltRecordBatchBuilder::default())
+                Box::new(PrebuiltRecordBatchBuilder::new(row_type.clone()))
             } else {
                 Box::new(RowAppendRecordBatchBuilder::new(&row_type)?)
             }
@@ -674,6 +685,11 @@ fn parse_ipc_message(
     let body_buffer = Buffer::from(body_data);
 
     Ok((batch_metadata, body_buffer, message.version()))
+}
+
+pub(crate) fn validate_append_record_batch(batch: &RecordBatch, row_type: &RowType) -> Result<()> {
+    let typed = TypedBatch::build(batch, row_type)?;
+    typed.check_not_null(row_type)
 }
 
 pub fn to_arrow_schema(fluss_schema: &RowType) -> Result<SchemaRef> {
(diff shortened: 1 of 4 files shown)


fluss-rust/crates/fluss/src/record/arrow.rs
```diff
@@ -207,12 +207,22 @@ pub trait ArrowRecordBatchInnerBuilder: Send {
     fn estimated_size_in_bytes(&self) -> usize;
 }
 
-#[derive(Default)]
-pub struct PrebuiltRecordBatchBuilder {
+pub(crate) struct PrebuiltRecordBatchBuilder {
```
charlesdong1991, line 127: IMHO, i don't think this should be a public api in the first place

charlesdong1991:
cc @leekeiabstraction @fresh-borzoni PTAL 

thanks in advance!! 🙏

fluss-rust/crates/fluss/src/record/arrow.rs
```diff
@@ -1979,6 +2012,102 @@ mod tests {
...
+            vec![Arc::new(arrow::array::Int64Array::from(vec![1_i64, 2_i64]))
+                as arrow::array::ArrayRef],
+        )
+        .expect("nullable arrow batch without nulls");
+        validate_append_record_batch(&ok_batch, &row_type)
+            .expect("nullable metadata alone must not reject a null-free batch");
+    }
+
+    #[test]
+    fn prebuilt_builder_rejects_nulls_in_not_null_column() {
```
leekeiabstraction, line 1498: Can you also add test cases for nested types?

leekeiabstraction (commented):
Thank you for the PR. Left a comment

fluss-rust/crates/fluss/src/record/arrow.rs
charlesdong1991: i thought i left a comment that it doesn't support nested type because IIUC, java side doesn't work with nested types either based on code 🤔 i didn't have time to do an end-to-end run on Java yesterday because our PoC was for python... 

i will check today and confirm if that is the case, and create an issue for Java if so...

charlesdong1991: Hi @leekeiabstraction sorry for late reply, was focused on something else... i ran Java, and my previous comment was wrong, and it does support nested types...

so i added tests too, and also tested it on our PoC, and seems work as expected.

PTAL 🙏

charlesdong1991:
gentle ping, would be great to get a reivew 🙏 @leekeiabstraction @fresh-borzoni

fluss-rust/crates/fluss/src/record/arrow.rs
```diff
@@ -1179,6 +1190,26 @@ fn parse_ipc_message(
     Ok((batch_metadata, body_buffer, message.version()))
 }
 
+pub(crate) fn validate_append_record_batch(batch: &RecordBatch, row_type: &RowType) -> Result<()> {
+    TypedBatch::build(batch, row_type)?;
+
+    for (i, field) in row_type.fields().iter().enumerate() {
```
fresh-borzoni: what about NOT NULL inside a container? 
ARRAY<INT NOT NULL> holding [1, null, 3] - the list itself isn't null, so null_count() is 0 and it passes. scan then fails with `non-nullable child of type Int32 contains nulls not present in parent List(non-null Int32)`. Also the same for ROW<seq INT NOT NULL>

fluss-rust/crates/fluss/tests/integration/log_table.rs
```diff
@@ -88,15 +126,23 @@ mod table_test {
             .append_arrow_batch(batch1)
             .expect("Failed to append batch with mixed nullability");
 
-        let batch2 = record_batch!(
-            ("c1", Int32, [4, 5, 6]),
-            ("c2", Utf8, ["a4", "a5", "a6"]),
-            ("c3", Int64, [40, 50, 60])
+        let batch2_schema = std::sync::Arc::new(arrow::datatypes::Schema::new(vec![
```
fresh-borzoni: nit: Arc and the arrays are imported a dozen lines up

fluss-rust/crates/fluss/tests/integration/log_table.rs
```diff
@@ -78,6 +78,44 @@ mod table_test {
             .create_writer()
             .expect("Failed to create writer");
 
+        {
+            use arrow::array::{Int32Array, Int64Array, StringArray};
```
fresh-borzoni: nit: move these up with the module imports?

fluss-rust/crates/fluss/src/record/arrow.rs
```diff
@@ -1979,6 +2010,305 @@ mod tests {
...
+        Arc::new(ListArray::new(
+            Arc::new(Field::new("item", inner.data_type().clone(), true)),
+            OffsetBuffer::new(outer_offsets.into()),
+            Arc::new(inner),
+            outer_validity.map(NullBuffer::from),
+        ))
+    }
+
+    #[test]
+    fn validate_append_record_batch_rejects_nulls_in_not_null_nested_columns() {
```
fresh-borzoni: these all null the container, not the element, same null_count() line as the scalar test above.

fluss-rust/crates/fluss/src/record/arrow.rs
```diff
@@ -1979,6 +2010,305 @@ mod tests {
...
+                    vec![0, 2, 4],
+                    vec![0, 1, 2],
+                    None,
+                ),
+            ),
+        );
+    }
+
+    #[test]
+    fn validate_append_record_batch_allows_null_elements_in_not_null_array() {
```
fresh-borzoni, line 1702: element type is nullable here, so this is just asserting arrow's default

fluss-rust/crates/fluss/src/record/arrow.rs
```diff
@@ -227,6 +237,7 @@ impl ArrowRecordBatchInnerBuilder for PrebuiltRecordBatchBuilder {
         if self.arrow_record_batch.is_some() {
             return Ok(false);
         }
+        validate_append_record_batch(record_batch.as_ref(), &self.row_type)?;
```
fresh-borzoni, line 157: isn't this already checked? 
append_arrow_batch validates the whole batch at append.rs:178 before splitting it.

fresh-borzoni (commented):
@charlesdong1991 Thank you for the PR, left a couple of comments, PTAL

fluss-rust/crates/fluss/src/record/arrow.rs
charlesdong1991: ohh, good catch, didn't consider this scenario @fresh-borzoni thanks!

updated

fluss-rust/crates/fluss/src/record/arrow.rs
charlesdong1991: yeah, it is checked at append.rs:178, but it's on one caller, and append_batch is for every prebuilt batch passes through... 

with another thought, i think it's better to keep the check here, and drop the check at 178 actually, so one validation nobody can bypass... WDYT? @fresh-borzoni

fluss-rust/crates/fluss/src/record/arrow.rs
charlesdong1991: this is for different purpose actually as the accept side of element nullability branch... so the recursive walk reaches the element, sees a null, and must not reject because element type is nullable, so it will guard against rejecting, not arrow... 

but if you feel it's not needed, i can also drop! let me know!

fresh-borzoni: fair, it guards the accept branch now. my comment predates the recursion, so all good 👍

fluss-rust/crates/fluss/src/row/column_vector.rs
```diff
@@ -224,6 +248,41 @@ impl TypedColumn {
         }
     }
 
+    fn check_not_null(
```
fresh-borzoni, line 277: does this handle sliced batches? 
values() returns the whole child array, so I think a slice would still see the trimmed rows' nulls.

fluss-rust/crates/fluss/src/record/arrow.rs
fresh-borzoni: tried dropping 178 and the poison batch enqueues 4 of its 6 rows before the error comes back, so it needs to stay.

but can this one ever fire? for_append_record_batch is only called from append_arrow_batch, and the sub-batches come from take_rows on a batch 178 already cleared.

fluss-rust/crates/fluss/tests/integration/log_table.rs
```diff
@@ -78,6 +92,36 @@ mod table_test {
...
+                        Some(10),
+                        Some(20),
+                        None,
+                        Some(40),
+                        Some(50),
+                        Some(60),
+                    ])),
+                ],
+            )
+            .expect("poison batch");
```
fresh-borzoni, line 118: Could this assert the offsets are still 0 instead of inferring from the record count below? 
A partial enqueue would then show up directly, rather than depending on what the poll happens to return first.
Also groups is a HashMap, so whether the clean bucket gets sent before the poisoned one errors depends on iteration order -  when I dropped 178 line check and run this a few times - it passes on some runs.

fresh-borzoni (commented):
@charlesdong1991 Thank you for the PR, great work overall 👍  Left some minor comments, PTAL

fluss-rust/crates/fluss/src/record/arrow.rs
charlesdong1991: yeah... i think my demo setup messed up that it didn't get the changes in the run 🤔 now, re-running failed... i kept both @fresh-borzoni 

the builder check inside append_batch is the write boundary for future callers, it does not fire on todays append_arrow_batch path after 178. so i think they are not duplicates, there is a send() between them

fluss-rust/crates/fluss/src/row/column_vector.rs
charlesdong1991: you are right... i updated, and also extended the walk as well since notice a bit more gaps on other types

fluss-rust/crates/fluss/src/row/column_vector.rs
```diff
@@ -34,6 +34,62 @@ use arrow::array::{
...
+    container: &dyn Array,
+    grandparent: Option<&dyn Array>,
+    fluss_type: &DataType,
+    path: String,
+) -> Result<()> {
+    let logical_len = container.len();
+    let ancestor_all_valid = grandparent.is_none_or(|p| p.null_count() == 0);
+    if container.null_count() == 0 && ancestor_all_valid {
+        let live = offset_window_slice(child, offsets, logical_len);
+        return check_child_not_null(live.as_ref(), fluss_type, &path);
```
fresh-borzoni: could the TypedColumn::build be hoisted out of the loop? 
it rebuilds per row, and one null container is enough

fresh-borzoni (commented):
@charlesdong1991 Thank you, LGTM 👍  One comment, PTAL

fluss-rust/crates/fluss/src/row/column_vector.rs
charlesdong1991: per-row rebuild is only on the mixed null path (some list/map rows null), for all-valid path, it builds once.

just to clarify, when it comes to hoisting, do you mean "build once on values() / the offset window, check with parent=None"? if that its the case, it would count leftover child nulls under null containers, and will cause the slice bug

Or you meant to reuse the boxed element/key/value TypedColumn that build() already created (check_not_null currently ignores them with _) and walk live offset ranges on that column? if that is the case, i can work on it as ROW already does that with inner @fresh-borzoni

charlesdong1991:
Hi @fresh-borzoni I presume you meant the second option. 

ARRAY/MAP now reuse the boxed typed child and only walk live offset ranges, so sliced values and leftovers under null containers stay excluded. I also preserved all ancestor masks for nested ROW and added regression tests.

PTAL 🙏 thanks again!

fresh-borzoni (approved):
@charlesdong1991 Thank you, I pushed some fixes on top, hope it's okay with you:

Turns out the masking was doing too much. a null list entry doesn't actually hide the nulls sitting in its range, arrow checks them regardless, so those batches write fine and then blow up on read. same story when a null ROW is further up the tree.

So I dropped the skipping and left just the offset window. Side effect is that it answers your hoisting question as the per-record loop was only there for the skipping, so it's gone entirely. ARRAY<ROW<INT NOT NULL>> with one null container went 1830µs -> 3µs on 20k rows in my tests on my machine.

For the tests I switched to writing each batch out and reading it back and asserting the two agree, rather than writing down accept/reject by hand.

PR #3952, opened 2026-08-12
[common] Fix equals/hashCode contract in value and kv record batches

### LINKED ISSUE #3951: [common] DefaultValueRecordBatch and DefaultKvRecordBatch override equals() without hashCode()
### Search before asking

### Fluss version

main (development)

## Summary

`DefaultValueRecordBatch` and `DefaultKvRecordBatch` both override `equals()` with value
semantics but do not override `hashCode()`. Both classes extend `Object` (they only
`implements` their respective interfaces) and neither `equals()` delegates to
`super.equals()`, so `hashCode()` is the identity hash from `Object`.

Two batches with identical bytes are therefore `equals()` but return different hash codes,
breaking the general contract of `Object.hashCode()`:

`DefaultValueRecordBatch` — value-based `equals()` at
[`DefaultValueRecordBatch.java:122`](https://github.com/apache/fluss/blob/main/fluss-common/src/main/java/org/apache/fluss/record/DefaultValueRecordBatch.java#L122),
no `hashCode()` anywhere in the class:

```java
@Override
    ...
    DefaultValueRecordBatch that = (DefaultValueRecordBatch) o;
    int sizeInBytes = sizeInBytes();
    return sizeInBytes == that.sizeInBytes()
            && segment.equalTo(that.segment, position, that.position, sizeInBytes);
}
```

`DefaultKvRecordBatch` — the same `equals()` shape at
[`DefaultKvRecordBatch.java:209`](https://github.com/apache/fluss/blob/main/fluss-common/src/main/java/org/apache/fluss/record/DefaultKvRecordBatch.java#L209),
also with no `hashCode()`.

## Why this looks like an oversight rather than a deliberate choice

The third batch implementation in the same package, `DefaultLogRecordBatch`, has the
identical `equals()` shape *and* overrides `hashCode()` consistently with it
([`DefaultLogRecordBatch.java:316`](https://github.com/apache/fluss/blob/main/fluss-common/src/main/java/org/apache/fluss/record/DefaultLogRecordBatch.java#L316)):

```java
@Override
public int hashCode() {
    return MurmurHashUtils.hashBytes(segment, position, sizeInBytes());
}
```

All three classes are memory-segment-backed record batches in
`org.apache.fluss.record`, all three compare the same byte range in `equals()`, and only
one of the three hashes it. The Checkstyle `EqualsHashCode` module is not enabled in
`tools/maven/checkstyle.xml`, which is why CI does not flag this today.

## Impact

Any `HashSet`/`HashMap` keyed on these batches would treat byte-identical batches as
distinct keys, so `contains()`/`get()` would miss. Nothing does that today: the only
collections involved are `Map<TableBucket, KvRecordBatch>` (e.g.
`ServerRpcMessageUtils#getPutKvData`, `ReplicaManager#putRecordsToKv`), where the batch is
the *value* type and its `hashCode()` is never consulted. So this is a latent correctness
trap for future callers rather than an active failure — but changing it now is free,
whereas finding it later means debugging a silent hash miss.

There is also a smaller, immediate diagnostic cost. Neither class overrides `toString()`,
so `Object.toString()` renders them as `getClass().getName() + "@" +
Integer.toHexString(hashCode())`. With an identity hash, the hex suffix carries no
information about content: two *equal* batches print different suffixes. This is visible in
the assertion failure in #3877, where the message is

```
expected: org.apache.fluss.record.DefaultValueRecordBatch@60eebc2e
 but was: org.apache.fluss.record.DefaultValueRecordBatch@1408822
```

Those two hex values do not tell a reader whether the contents actually differ. Once
`hashCode()` is content-derived, differing suffixes do correspond to differing bytes.

To be explicit: this is **not** a fix for the flakiness in #3877 — the reason the two
batches genuinely differ there is a separate question, and this change does not address it.

### Solution

Override `hashCode()` in both classes, mirroring `DefaultLogRecordBatch` exactly so the
three sibling implementations stay consistent:

```java
@Override
public int hashCode() {
    return MurmurHashUtils.hashBytes(segment, position, sizeInBytes());
}
```

`equals()` compares the byte range `[position, position + sizeInBytes())`, and
`MurmurHashUtils.hashBytes(segment, position, sizeInBytes())` hashes exactly that range, so
the two are consistent by construction.

For scope: I swept `src/main/java` across the repo for classes that override
`equals(Object)` without `hashCode()`. Besides these two there are five more where the
superclass is `Object` and so the contract is genuinely violated —
`TieringSourceEnumeratorState`, `RemoteLogManifestHandle`, and the three nested
`IsrState` implementations. A further group (`WriteResultForBucket` and subclasses, the
`ModifyColumn*` classes) inherits a `hashCode()` from a base class that hashes a subset of
the fields compared by `equals()`; that is legal, only a weaker hash, so I have left it
alone. I would rather keep this issue to the two record-batch classes so the PR stays to a
single module and a single concern, and follow up on the other five separately.

vbhanuchander-lang:
### Purpose

Linked issue: close #3951

`DefaultValueRecordBatch` and `DefaultKvRecordBatch` override `equals()` with value
semantics but never override `hashCode()`. Both classes only `implements` their interface,
so their superclass is `Object`, and neither `equals()` delegates to `super.equals()` —
`hashCode()` is therefore the identity hash. Two byte-identical batches are `equals()` but
return different hash codes, which violates the general contract of `Object.hashCode()`.

The third batch implementation in the same package, `DefaultLogRecordBatch`, has the
identical `equals()` shape and *does* hash the same byte range
([`DefaultLogRecordBatch.java:316`](https://github.com/apache/fluss/blob/main/fluss-common/src/main/java/org/apache/fluss/record/DefaultLogRecordBatch.java#L316)),
which is why this reads as an oversight rather than a deliberate choice. The Checkstyle
`EqualsHashCode` module is not enabled in `tools/maven/checkstyle.xml`, so CI does not flag
it today.

### Brief change log

- `DefaultValueRecordBatch`: add `hashCode()`, mirroring `DefaultLogRecordBatch` exactly.
- `DefaultKvRecordBatch`: same.

Both use `MurmurHashUtils.hashBytes(segment, position, sizeInBytes())`. `equals()` compares
the byte range `[position, position + sizeInBytes())` via
`segment.equalTo(that.segment, position, that.position, sizeInBytes)`, and `hashBytes`
hashes exactly that range, so `equals()` and `hashCode()` are consistent by construction.

### Tests

New in `DefaultValueRecordBatchTest` (new file) and `DefaultKvRecordBatchTest`:

| Test | Asserts |
| --- | --- |
| `testEqualsAndHashCode` | two independently built, byte-identical batches are equal **and** share a hash code (fails before this change) |
| `testHashCodeDiffersForDifferentContents` | batches with different contents are unequal and hash differently |
| `testEmptyBatchesAreEqualAndShareHashCode` | (value batch only) the empty-batch case, which is the shape asserted in `ReplicaManagerTest#testLimitScanPrimaryKeyTable` |

Verified locally with `mvn clean verify -pl fluss-common`: 1748 + 186 tests pass, Checkstyle
reports 0 violations, `spotless:check` and `apache-rat:check` pass.

### API and Format

No API or storage-format change. `hashCode()` is additive.

The only behavioural change is for a `HashSet`/`HashMap` **keyed** on one of these batches,
which would previously have used the identity hash. Nothing does that today — the batches
appear only as map *value* types (e.g. `Map<TableBucket, KvRecordBatch>` in
`ServerRpcMessageUtils#getPutKvData` and `ReplicaManager#putRecordsToKv`), where
`hashCode()` is never consulted.

### Documentation

No new feature, no documentation change.

### Note on #3877

I found this while diagnosing the flaky `ReplicaManagerTest#testLimitScanPrimaryKeyTable`
in #3877. This PR does **not** fix that flakiness — why the two batches genuinely differ
there is a separate question. It does make the failure message more useful: neither class
overrides `toString()`, so `Object.toString()` renders the identity hash, and the
`...@60eebc2e` vs `...@1408822` in that CI log cannot tell a reader whether the contents
actually differ (two *equal* batches would also print different suffixes). With a
content-derived `hashCode()`, differing suffixes do mean differing bytes.

### Generative AI disclosure

- [x] Yes — Claude Code (Opus 5). All changes reviewed by me before submitting.

### DIFF
--- a/fluss-common/src/main/java/org/apache/fluss/record/DefaultKvRecordBatch.java
+++ b/fluss-common/src/main/java/org/apache/fluss/record/DefaultKvRecordBatch.java
@@ -23,6 +23,7 @@ import org.apache.fluss.exception.CorruptMessageException;
 import org.apache.fluss.memory.MemorySegment;
 import org.apache.fluss.record.bytesview.BytesView;
 import org.apache.fluss.utils.CloseableIterator;
+import org.apache.fluss.utils.MurmurHashUtils;
 import org.apache.fluss.utils.crc.Crc32C;
 
 import java.nio.ByteBuffer;
@@ -217,6 +218,11 @@ public class DefaultKvRecordBatch implements KvRecordBatch {
         int sizeInBytes = sizeInBytes();
         return sizeInBytes == that.sizeInBytes()
                 && segment.equalTo(that.segment, position, that.position, sizeInBytes);
+    }
+
+    @Override
+    public int hashCode() {
+        return MurmurHashUtils.hashBytes(segment, position, sizeInBytes());
     }
 
     abstract class KvRecordIterator implements CloseableIterator<KvRecord> {
--- a/fluss-common/src/main/java/org/apache/fluss/record/DefaultValueRecordBatch.java
+++ b/fluss-common/src/main/java/org/apache/fluss/record/DefaultValueRecordBatch.java
@@ -21,6 +21,7 @@ import org.apache.fluss.memory.MemorySegment;
 import org.apache.fluss.memory.MemorySegmentOutputView;
 import org.apache.fluss.row.BinaryRow;
 import org.apache.fluss.utils.CloseableIterator;
+import org.apache.fluss.utils.MurmurHashUtils;
 
 import java.io.IOException;
 import java.nio.ByteBuffer;
@@ -130,6 +131,11 @@ public class DefaultValueRecordBatch implements ValueRecordBatch {
         int sizeInBytes = sizeInBytes();
         return sizeInBytes == that.sizeInBytes()
                 && segment.equalTo(that.segment, position, that.position, sizeInBytes);
+    }
+
+    @Override
+    public int hashCode() {
+        return MurmurHashUtils.hashBytes(segment, position, sizeInBytes());
     }
 
     // ------------------------------------------------------------------------------------------
--- a/fluss-common/src/test/java/org/apache/fluss/record/DefaultKvRecordBatchTest.java
+++ b/fluss-common/src/test/java/org/apache/fluss/record/DefaultKvRecordBatchTest.java
@@ -27,7 +27,9 @@ import org.junit.jupiter.api.Test;
 import java.util.ArrayList;
 import java.util.List;
 
+import static org.apache.fluss.record.TestData.DATA1_ROW_TYPE;
 import static org.apache.fluss.record.TestData.DATA1_SCHEMA;
+import static org.apache.fluss.testutils.DataTestUtils.compactedRow;
 import static org.assertj.core.api.Assertions.assertThat;
 
 /** Test for {@link DefaultKvRecordBatch}. */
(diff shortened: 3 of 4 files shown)


vbhanuchander-lang:
Could a committer approve the pending workflow runs? All three (`CI`, `Check License`,
`Client Integration`) are at `action_required`, so nothing has executed on this branch yet.
Locally `mvn clean verify -pl fluss-common` is green — 1748 + 186 tests, 0 Checkstyle
violations — and `spotless:check` and `apache-rat:check` pass.

cc @loserwang1024 @luoyuxia — you are the most recent contributors to these files. The change is
four lines of production code: `DefaultValueRecordBatch` and `DefaultKvRecordBatch` override
`equals()` with value semantics but never override `hashCode()`, so byte-identical batches are
equal while hashing differently. `DefaultLogRecordBatch` in the same package already has the
identical `equals()` shape plus the matching `hashCode()`, and I mirrored it exactly rather than
inventing anything.

Details, including why nothing currently depends on the old identity hash, are in #3951.

wuchong (approved):

PR #3953, opened 2026-08-12
[flink] Preserve complete undo recovery baselines

### LINKED ISSUE #3838: [Flink] Flink aggregation-table sink may lose the complete Undo Recovery baseline
### Search before asking

### Fluss version

0.9.0 (latest release)

### Please describe the bug 🐞

The Undo Recovery operator previously persisted only buckets that required Undo Recovery or received new writes. Buckets whose offsets remained unchanged could therefore disappear from subsequent checkpoints.

After another restore, these missing buckets were treated as having offset 0. This could trigger unnecessary full-history Undo Recovery and significantly increase recovery time for aggregation tables.

Legacy writer state cannot indicate whether its bucket offsets form a complete baseline. Restoring such state as complete would therefore be unsafe.

### Solution

Introduce a versioned writer state format that explicitly represents a complete Undo Recovery baseline and carries the associated table identity.

During recovery:

- enumerate all live buckets assigned to the subtask;
- preserve the complete non-zero baseline independently from the buckets requiring Undo;
- treat missing buckets as an explicit zero baseline only for complete V2 state;
- reject legacy V1 state because its completeness cannot be determined safely;
- persist one complete V2 state during every checkpoint, including an empty baseline.

platinumhamburg:
### Purpose

Linked issue: close #3838

### DIFF
--- a/fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/sink/state/WriterState.java
+++ b/fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/sink/state/WriterState.java
@@ -32,12 +32,30 @@ import java.util.Objects;
  * writer is responsible for. During failover recovery, these offsets are used to generate undo logs
  * and rollback to the checkpoint state.
  *
+ * <p>V2 state is a complete but sparse recovery baseline. It stores only positive offsets; a live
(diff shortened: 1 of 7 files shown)


wuchong (approved):
LGTM.

PR #3956, opened 2026-08-12
[server] Fix remote log start offset of RemoteLogManifest may expose an expired physical prefix

### LINKED ISSUE #3943: Remote log start offset may expose an expired physical prefix

gyang94:
### Purpose

Remote log manifests can expose a logical range narrower than the physical ranges of their persisted segments. The aggregate manifest offsets are propagated as the remotely readable range, so they must be derived from logical boundaries.

### Brief change log

- Make `RemoteLogManifest#getRemoteLogStartOffset` return the minimum logical start offset.
- Make `RemoteLogManifest#getRemoteLogEndOffset` return the maximum logical end offset.
- Preserve the existing Java API and RPC field names.
- Add coverage for manifests with clipped physical prefixes and suffixes.

### Tests

- `mvn -o -pl fluss-common -Dtest=RemoteLogManifestOverlapTest test`
- Checkstyle and Spotless passed as part of the Maven lifecycle.

### API and Format

No API, RPC field, or storage-format changes.

### Documentation

No documentation changes.

### Generative AI disclosure

- [x] Yes — Codex (gpt-5)

### DIFF
--- a/fluss-common/src/main/java/org/apache/fluss/remote/RemoteLogManifest.java
+++ b/fluss-common/src/main/java/org/apache/fluss/remote/RemoteLogManifest.java
@@ -145,8 +145,8 @@ public class RemoteLogManifest {
     public long getRemoteLogStartOffset() {
         long startOffset = Long.MAX_VALUE;
         for (RemoteLogSegment remoteLogSegment : remoteLogSegmentList) {
-            if (remoteLogSegment.remoteLogStartOffset() < startOffset) {
-                startOffset = remoteLogSegment.remoteLogStartOffset();
+            if (remoteLogSegment.logicalStartOffset() < startOffset) {
+                startOffset = remoteLogSegment.logicalStartOffset();
             }
         }
         return startOffset;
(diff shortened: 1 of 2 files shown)


wuchong (approved):
LGTM.

Issue #3957, opened 2026-08-12
[Umbrella][FIP-49] Implement the Fluss Gateway REST API


beryllw:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/fluss/issues) and found nothing similar.

### Motivation

FIP-49 has been accepted to add a stateless, multi-cluster REST Gateway for Apache Fluss. The Gateway exposes metadata discovery, DDL, writes, primary-key lookups, and prefix lookups to HTTP clients while preserving Fluss authentication and authorization semantics.

- FIP: https://cwiki.apache.org/confluence/spaces/FLUSS/pages/444334561/FIP-49+Fluss+Gateway+REST+API
- Discussion: https://lists.apache.org/thread/s5yjysvoj8o6nxdj49m99hg8yn51tgjw
- Vote: https://lists.apache.org/thread/t5jkcvo8b07oxjywbckotbsxl8fhcd76
- Vote result: https://lists.apache.org/thread/f4217jbyy1dv139ml9t5ybwkhsymy2xm

A proof of concept implements most of the proposal, but it is too large to review as one pull request. This umbrella tracks capability-sized issues and review units that can be tested and merged independently.

### Sub-tasks

The work is split along three parallel lines so each unit can be reviewed and merged independently: capability units run against the `GatewayBackend` port with the testing backend and do not wait for native connectivity; native-connectivity units are config-gated and land in parallel; delivery items trail the rest.

**Foundation**

- [x] #4002
- [x] #3958
- [x] #3969

**Capability line**

- [x] #4067
- [x] #4082
- [x] #4083
- [x] #4084

**Delivery**

- [x] #4085
- [x] #4160

**Future Work**
- [ ] #4081
- [ ] [FIP-49][gateway] Add batched primary-key and prefix lookups with admission limits
- [ ] [FIP-49][gateway] Wrap up: complete the Prometheus metric families with a checklist regression, add user documentation, validate the OpenAPI specification, and execute the dockerized end-to-end quickstart
- [ ] [FIP-49][gateway] And Propagate HTTP principals to Fluss with bounded act-as connection pools

### Willingness to contribute

- [x] I'm willing to submit a PR!

beryllw:
@leonardBang @loserwang1024 cc

PR #3963, opened 2026-08-12
[gateway] Add the Gateway runtime, common HTTP layer, lifecycle, and test infrastructure

### LINKED ISSUE #3958: [FIP-49][gateway] Add the Gateway runtime, common HTTP layer, lifecycle, and test infrastructure
### Search before asking

### Description

[FIP-49](https://cwiki.apache.org/confluence/spaces/FLUSS/pages/444334561/FIP-49+Fluss+Gateway+REST+API) introduces a standalone REST Gateway for Fluss. This task adds the shared runtime and HTTP foundation used by the subsequent Gateway APIs:

- add `fluss-gateway` as an independent Cargo workspace with library and executable entry points;
- add the REST and metrics listeners;
- add `GET /health`, the health endpoint FIP-49 defines;
- add common HTTP handling for request IDs, error responses, unknown routes, request-body limits, and request deadlines;
- add OpenAPI 3.1 generation and checked-in document drift checking for the routes and schemas introduced by this task;
- add process lifecycle management, including task supervision, startup and runtime failure propagation, `SIGTERM`, request draining, task shutdown, and process exit codes;
- add Gateway CI checks, MSRV verification, and Linux binary compilation; 
- add shared test infrastructure for unit tests, real-listener tests, process lifecycle tests, and fixed-version Fluss cluster E2E tests.

### Willingness to contribute

beryllw:
### Purpose

Linked issue: close #3958

First PR of FIP-49: the Gateway runtime and HTTP foundation. Authentication, the Fluss backend and the data
APIs come in follow-up issues.

Distilled from the FIP-49 PoC; the foundation was written by @gstamatakis95 and is credited via `Co-authored-by`.

### Brief change log

- `fluss-gateway/` as an independent Cargo workspace, library plus executable.
- `gateway.yaml` configuration: flat dotted keys, env and CLI overrides, validation, stable exit codes.
- Common HTTP layer: one error envelope, request IDs, 404/405/413/504.
- `GET /health` returning `{status, uptime_ms}`.
- Lifecycle: REST and Prometheus listeners, task supervision, SIGTERM draining.
- Metrics: the FIP-49 families that need no backend.
- OpenAPI 3.1 generated from the router and checked in, with CI drift and lint gates.
- CI: build, test, clippy, fmt, rustdoc, dependency licences, MSRV.

### Tests

66 tests: unit tests, the HTTP contract over a real listener, and the compiled binary's startup, health, SIGTERM and exit codes. A `--features integration_tests` suite drives the gateway against a dockerized Fluss cluster on demand; it is not a pull-request gate yet.

### API and Format

New endpoints `GET /health` and `GET /v1/openapi.json`. No change to existing APIs or storage formats.

### Documentation

Module-level rustdoc, enforced by `cargo doc -D warnings`.


beryllw:
@fresh-borzoni  This is the first PR for FIP-49 — could you help review it when you have time? Thanks a lot! 🙏

beryllw:
@luoyuxia @loserwang1024 @naivedogger cc

beryllw:
Split out the module scaffolding (workspace setup, toolchain/lint/license configs, placeholder targets, and the dedicated gateway CI gates) into #4002 for a fast first merge. This PR keeps all of the gateway runtime, HTTP layer, lifecycle, and test suites, and will be rebased once #4002 lands.

fluss-gateway/src/lifecycle.rs
```diff
@@ -0,0 +1,663 @@
...
+        })
+}
+
+/// Serves one Axum listener until process cancellation starts graceful drain.
+async fn serve(
+    listener: tokio::net::TcpListener,
+    router: Router,
+    shutdown: CancellationToken,
+) -> Result<(), String> {
+    let server = axum::serve(listener, router).with_graceful_shutdown(async move {
```
fresh-borzoni: A client can open a connection, send half a request head, and we hold it forever. No header-read timeout here, and the request-timeout middleware never sees these because it only runs after hyper has a complete head. There's no auth on the listener either, so anyone who can reach the port can tie up sockets and tasks until we run out of them. 
Mb set one on the hyper builder?

fluss-gateway/src/config.rs
```diff
@@ -0,0 +1,1102 @@
...
+                problems.push(format!(
+                    "{} must be 1-128 ASCII letters, digits, dots, underscores, or hyphens",
+                    INSTANCE_ID_KEY
+                ));
+            }
+        }
+        // Port 0 asks the OS for a free port, so two ephemeral listeners never collide even though the
+        // configured addresses are equal.
+        if server.metrics.enabled
+            && server.metrics.bind_address == rest_address
```
fresh-borzoni: You can put both listeners on the same port and this won't notice, e.g  0.0.0.0:8080 for REST, 127.0.0.1:8080 for metrics. The bad part is that it starts up clean and the process looks healthy while nothing can use it, combine it with readiness checks and K8s and we are asking for the problems

fluss-gateway/src/protocol/rest/mod.rs
```diff
@@ -0,0 +1,699 @@
...
+
+    let limits = move |mut request: Request, next: Next| async move {
+        let request_id = request
+            .extensions()
+            .get::<RequestId>()
+            .cloned()
+            .unwrap_or_default();
+        request
+            .extensions_mut()
+            .insert(deadline_from_now(request_timeout));
```
fresh-borzoni, line 342: This deadline and the timeout below expire together, so a handler that answers at the deadline has its response thrown away and the caller gets a bare 504. Should be smth like a grace to the middleware timeout, or just subtract it here?

fluss-gateway/src/protocol/rest/health.rs
```diff
@@ -0,0 +1,134 @@
...
+    tag = "health",
+    responses(
+        (status = 200, description = "Gateway liveness and uptime", body = HealthResponse),
+        (status = 405, description = "Wrong method for this route", body = ErrorEnvelope),
+    )
+)]
+pub(crate) async fn health(State(state): State<RestState>) -> Response {
+    // The response type is the documented schema, so the payload cannot drift from the contract.
+    json_response(&HealthResponse {
+        status: "ok",
```
fresh-borzoni, line 62: /health says ok while draining and before we start serving, the two windows where every other route returns 503, so LB keeps sending traffic we reject. I'm confused here, can you clarify?

fresh-borzoni (commented):
@beryllw Thank you for the PR! 
I looked quickly with a limited time I had, left some quick comments, PTAL

fresh-borzoni (commented):
@beryllw Thank you, but the thing  I'd look at again is dropping axum::serve to get at the builder.

auto first checks whether the connection opens with the HTTP/2 preface, and that read has no timer on it, so a socket that sends literally nothing is still parked (half a request head does get closed now). Pulling in server-auto also flips on hyper/http2, so anything that does send that preface now gets an HTTP/2 connection, and if it then just sits there hyper's graceful shutdown waits on it, my SIGTERM sat for the full drain and exited 1. Then the accept arm returns Err, so one EMFILE (tried it with ulimit -n 40) and run() takes the whole process down, where axum just sleeps a second and carries on.

Both the parked socket and the HTTP/2 surface can go away with hyper::server::conn::http1::Builder instead of auto, the timer starts at setup there and there's no HTTP/2 to serve. Needs the non-upgradeable Connection, graceful.watch won't take the other one. For accept, could we copy what axum does in handle_accept_error (serve/listener.rs), skip the connection-level errors and sleep a second on the rest instead of returning?

WDYT?

beryllw:
Hey @leekeiabstraction — would you be interested in helping review this PR (or any of the follow-ups in #3957)? It's part of the FIP-49 gateway implementation, which we're targeting for 1.0. The timeline is a bit tight, so any early feedback would really help. Thanks! 🙏

fluss-gateway/src/lifecycle.rs
```diff
@@ -0,0 +1,706 @@
...
+        builder
+            .timer(TokioTimer::new())
+            .header_read_timeout(header_read_timeout);
+        let service = TowerToHyperService::new(router.clone());
+        // The non-upgradeable connection is what `graceful.watch` accepts; the gateway has no
+        // upgrade-based protocol.
+        let connection = graceful.watch(builder.serve_connection(TokioIo::new(socket), service));
+        tokio::spawn(async move {
+            if let Err(error) = connection.await {
+                log::debug!("connection ended with an error: {error}");
```
fresh-borzoni: Everything that ends a connection lands here at debug, including a client that tried HTTP/2, so ingress configured for h2c upstream would look like a dead gateway from both sides. Shall we provide a clue for the users in the logs, mb warn for protocol errors?

fluss-gateway/Cargo.toml
```diff
@@ -42,7 +46,46 @@ path = "src/main.rs"
...
-# The runtime (configuration, lifecycle, HTTP layer, and tests) arrives with
-# the Gateway foundation change (FIP-49) and brings its dependencies with it.
+[dependencies]
+axum = { version = "0.8", default-features = false, features = ["http1", "matched-path", "tokio"] }
+# The listeners follow axum's official serve-with-hyper example: axum::serve does not expose the
+# hyper builder that the header read timeout lives on. The plain http1 builder (not auto) has no
+# HTTP/2 preface sniffing, so the header timer covers a connection from its first byte.
+hyper = { version = "1", default-features = false, features = ["http1", "server"] }
+# GracefulShutdown and the tower-to-hyper service adapter for the serve loop above.
+hyper-util = { version = "0.1", default-features = false, features = ["server-graceful", "service", "tokio"] }
```
fresh-borzoni: nit: graceful.watch needs hyper-util's http1 feature (the GracefulConnection impl for http1::Connection is behind it) and we're not asking for it, it only builds because axum's http1 turns it on. 

Worth doing it explicitly, otherwise the day that changes the natural fix is to put server-auto back, and HTTP/2 comes back with it.

fluss-gateway/src/lifecycle.rs
```diff
@@ -0,0 +1,706 @@
...
+) -> Result<(), String> {
+    let graceful = hyper_util::server::graceful::GracefulShutdown::new();
+    loop {
+        let socket = tokio::select! {
+            biased;
+            _ = shutdown.cancelled() => break,
+            accepted = listener.accept() => match accepted {
+                Ok((socket, _remote)) => socket,
+                Err(error) => {
+                    handle_accept_error(error).await;
```
fresh-borzoni: This await is outside the select!, so a SIGTERM arriving during the sleep has to wait it out. axum keeps the sleep inside Listener::accept and the serve loop selects that against the shutdown signal, so there it gets cancelled. Shall we move it into the select?

fresh-borzoni (approved):
@beryllw Thank you, I think it's good overall, left some minors, but I approve. Let me know when you wish to merge it.

fluss-gateway/src/lifecycle.rs
```diff
@@ -0,0 +1,712 @@
...
+        let mut builder = http1::Builder::new();
+        // The header read timeout runs on hyper's background timer, so one must be installed first.
+        builder
+            .timer(TokioTimer::new())
+            .header_read_timeout(header_read_timeout);
+        let service = TowerToHyperService::new(router.clone());
+        // The non-upgradeable connection is what `graceful.watch` accepts; the gateway has no
+        // upgrade-based protocol.
+        let connection = graceful.watch(builder.serve_connection(TokioIo::new(socket), service));
+        tokio::spawn(async move {
```
luoyuxia: The `JoinHandle` returned here is discarded, so this connection task is not owned by the process-level `JoinSet`. When `drain_tasks()` reaches its deadline and calls `abort_all()`, it can abort only the listener task running `serve()`; connection tasks spawned here may keep running independently. As a result, `RunningGateway::shutdown()` can return while a timed-out handler is still executing in the same runtime. Once write APIs are added, work with backend side effects could continue after shutdown has reported completion.

Could we keep lifecycle ownership of these connection tasks and abort + join them when the drain deadline expires? A regression test with a handler that outlives the drain timeout could also assert that the handler is dropped before `shutdown()` returns.

.github/workflows/gateway-ci.yml
```diff
@@ -71,10 +69,18 @@ jobs:
           RUST_LOG: DEBUG
           RUST_BACKTRACE: full
 
+      - name: OpenAPI drift check
+        run: |
+          cargo test --lib protocol::rest::openapi::tests::export_checked_in_document -- --ignored --exact
```
luoyuxia: The Cargo validation commands in this workflow currently run without `--locked`. If `Cargo.toml` is changed without the corresponding `Cargo.lock` update, Cargo may rewrite the lockfile in the CI workspace and still pass, and the workflow does not check that diff. Similarly, the checked-in `DEPENDENCIES.rust.tsv` has a manual generation recipe but no drift check.

Could we run the relevant validation commands with `--locked` and add an inventory regeneration plus `git diff --exit-code`, similar to the OpenAPI drift check? That would ensure the committed lockfile and license inventory describe the dependency graph CI actually validated.

luoyuxia (commented):
The overall foundation is thoughtfully designed, with strong coverage around configuration, the HTTP error contract, OpenAPI generation, and lifecycle behavior. I left two suggestions inline.

luoyuxia (approved):

PR #3970, opened 2026-08-12
[gateway] Validate configuration and redact secrets

### LINKED ISSUE #3969: [FIP-49][gateway] Validate Gateway configuration, restrict Fluss client options, and redact secrets
### Search before asking

### Description

[FIP-49](https://cwiki.apache.org/confluence/spaces/FLUSS/pages/444334561/FIP-49+Fluss+Gateway+REST+API) requires deterministic Gateway configuration and must prevent native-client options from bypassing Gateway-owned guarantees or exposing credentials. This task:

- adds flat dotted YAML configuration with documented file, environment-variable, and CLI precedence;
- adds typed server, security, request-limit, and per-cluster connection settings;
- validates supported `client.*` options, value ranges, conversions, and cross-field constraints before listeners bind or Fluss connections are created;
- prevents client options from disabling Gateway-owned acknowledgement, idempotence, and retry behavior;
- rejects a statically configured Fluss authorization identity because user mode supplies it per request;
- defines `connection.service.account` and `connection.service.secret` as the canonical service credential settings while retaining deterministic legacy last-wins overrides without rejecting them in this task;
- emits one redacted deprecation warning per affected cluster and redacts tokens, users, passwords, and service secrets from debug output, errors, and logs; and
- adds unit, property, and process tests for precedence, invalid configuration, client-option restrictions, warning behavior, and secret redaction.

Part of FIP-49.

### Willingness to contribute

beryllw:
### Purpose

Linked issue: close #3969

Adds the typed, validated and redacted Gateway configuration on top of #3963.

### Brief change log

- Typed per-cluster connection, security and request-limit settings, with `gateway.clusters` declaring which clusters may be configured.
- User identity mode requires service credentials over `client.security.protocol: sasl`.
- `connection.service.account` / `.secret` become canonical; the legacy SASL keys keep last-wins precedence and warn once per cluster.
- Credentials become a `Secret` newtype, redacted in logs, errors and the startup dump.

### Documentation

Module rustdoc only; user documentation lands with the delivery task in #3957.

### DIFF
--- a/.github/workflows/gateway-ci.yml
+++ b/.github/workflows/gateway-ci.yml
@@ -25,6 +25,7 @@ on:
       - main
     paths:
       - 'fluss-gateway/**'
+      - 'fluss-rust/Cargo.toml'
       - 'fluss-rust/crates/fluss/**'
       - 'fluss-rpc/src/main/proto/**'
       - '.github/workflows/gateway-ci.yml'
@@ -33,6 +34,7 @@ on:
       - main
     paths:
       - 'fluss-gateway/**'
+      - 'fluss-rust/Cargo.toml'
       - 'fluss-rust/crates/fluss/**'
       - 'fluss-rpc/src/main/proto/**'
       - '.github/workflows/gateway-ci.yml'
--- a/.github/workflows/gateway-ci.yml
+++ b/.github/workflows/gateway-ci.yml
@@ -15,7 +15,7 @@
 # specific language governing permissions and limitations
 # under the License.
 
-# Gateway-only CI: gateway changes never build the fluss-rust workspace and vice versa.
+# Gateway CI also follows the native Rust client sources used by its configuration adapter.
 # `uses:` step inputs are relative to the repository root.
 name: Gateway CI
 
@@ -25,12 +25,16 @@ on:
       - main
     paths:
       - 'fluss-gateway/**'
+      - 'fluss-rust/crates/fluss/**'
+      - 'fluss-rpc/src/main/proto/**'
       - '.github/workflows/gateway-ci.yml'
   pull_request:
     branches:
       - main
     paths:
       - 'fluss-gateway/**'
+      - 'fluss-rust/crates/fluss/**'
+      - 'fluss-rpc/src/main/proto/**'
       - '.github/workflows/gateway-ci.yml'
   workflow_dispatch:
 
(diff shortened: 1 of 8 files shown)


beryllw:
@fresh-borzoni @luoyuxia cc

.github/workflows/gateway-ci.yml
```diff
@@ -25,12 +25,16 @@ on:
       - main
     paths:
       - 'fluss-gateway/**'
+      - 'fluss-rust/crates/fluss/**'
```
fresh-borzoni, line 29: fluss-rs takes arrow, tokio and others as workspace = true, so what the gateway compiles is pinned in fluss-rust/Cargo.toml, which isn't in this list. Shall we add it?

fluss-gateway/src/config.rs
fresh-borzoni: KiB is fine, it's KB that worries me. Fluss's MemorySize lowercases the unit and treats kb as 1024, so 64KB is 64000 bytes here and 65536 in server.yaml. Shall we match java or nb add more descriptive error?
Also 1d and 1 s parse in TimeUtils but not here.

fluss-gateway/src/config.rs
```diff
@@ -315,18 +584,238 @@ impl RestServerConfig {
...
+            native.security_protocol = "sasl".to_string();
+            native.security_sasl_mechanism = "PLAIN".to_string();
+            native.security_sasl_username = account.clone();
+            native.security_sasl_password = secret.expose().to_string();
+        }
+        native
+    }
+}
+
+impl fmt::Debug for ClusterConfig {
```
fresh-borzoni: Secret already redacts itself in Debug, so #[derive(Debug)] prints the same thing

fluss-gateway/src/config.rs
fresh-borzoni, line 1006: We warn when gateway.rest.listen is not loopback, but not when the metrics listener is not loopback. The FIP default is 0.0.0.0:9095, so the metrics port can be open to the network with no warning.

Same for trusted-header auth: the principal comes from the x-forwarded-user header, so on a non-loopback listener any client can set it and become any user.

Should we warn for both here?

fluss-gateway/src/config.rs
```diff
@@ -355,27 +844,163 @@ impl ShutdownConfig {
...
+                    "cluster ID {id:?} must be at most 63 characters, start with a lowercase letter, and contain only lowercase letters, digits, or underscores"
+                ));
+            }
+            if !cluster
+                .bootstrap_servers
+                .split(',')
+                .any(|server| !server.trim().is_empty())
+            {
+                problems.push(format!(
+                    "{} must configure at least one server",
```
fresh-borzoni: This accepts anything non-empty, so host, host:99999 and http://h:9123 all load. The bad part is the process then starts clean and /ready says 200 while every request to that cluster fails. 
Shall we parse each entry as host:port here?

fresh-borzoni (commented):
@beryllw Thank you for the PR, left some comments , PTAL

fresh-borzoni (approved):
@beryllw Thank you, LGTM 👍 
I'll merge once CI passes.

Issue #3845, opened 2026-08-04
[rust] Filter pushdown integration test


leekeiabstraction:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/fluss/issues) and found nothing similar.

### Description

IT for filter pushdown

### Willingness to contribute

- [ ] I'm willing to submit a PR!

leekeiabstraction:
Closed by #4109

PR #3971, opened 2026-08-12
[rust] Wire the scan filter through to the log fetcher

### LINKED ISSUE #3844: [rust] Filter pushdown implementation

leekeiabstraction:
## Summary
- Closes #3844.
- Adds `TableScan::filter`, which encodes the predicate against the full row type and sends it with the schema id on every fetch request.
- A response now advances the bucket offset past `filtered_end_offset`, whether the server emptied the fetch or only pruned its tail, and the proto comment claiming those two fields are mutually exclusive is corrected to match the server.

## Test Plan
- `cargo test -p fluss-rs --lib`: 628 passed, covering the request fields, both filtered response shapes, and the `EARLIEST_OFFSET` sentinel that -1 must not outrank.
- Full integration suite passes (65), though the test cluster runs 0.9.1, which predates server-side filtering, so pruning is not yet exercised end to end.

### DIFF
--- a/fluss-rust/crates/fluss/src/client/table/log_fetch_buffer.rs
+++ b/fluss-rust/crates/fluss/src/client/table/log_fetch_buffer.rs
@@ -354,6 +354,9 @@ impl PendingFetch for CompletedPendingFetch {
     }
 }
 
+/// Sentinel for a fetch whose response carried no `filtered_end_offset`.
+pub(crate) const NO_FILTERED_END_OFFSET: i64 = -1;
+
 /// Default implementation of CompletedFetch for in-memory log records
 /// Used for local fetches from tablet server
 pub struct DefaultCompletedFetch {
@@ -365,6 +368,9 @@ pub struct DefaultCompletedFetch {
     resolver: Arc<ReadContextResolver>,
     is_remote: bool,
     next_fetch_offset: i64,
+    /// Offset the server scanned up to while pruning batches by their
+    /// statistics, or [`NO_FILTERED_END_OFFSET`] when it sent no filtered range.
+    filtered_end_offset: i64,
     high_watermark: i64,
     size_in_bytes: usize,
     consumed: bool,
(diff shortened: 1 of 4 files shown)


leekeiabstraction:
@platinumhamburg @fresh-borzoni @charlesdong1991 Appreciate a review here please 🙏

fluss-rust/crates/fluss/src/client/table/log_fetch_buffer.rs
```diff
@@ -1237,6 +1273,151 @@ mod tests {
         Ok(())
     }
 
+    /// An empty fetch carrying only a filtered range, as the server sends when
+    /// it prunes every batch it scanned.
+    fn filtered_empty_fetch(
```
leekeiabstraction, line 1278: Fluss core ITs currently test against 0.9.1-incubating. To unblock integration test, we'd need to wait for 1.0 RC.

fresh-borzoni: The 0.9.1 limitation is only local. CI builds the server from this tree and runs the Rust ITs against it (FLUSS_VERSION=dev in client-integration.yml), so you can write it now.

One ask for that test though: assert that something actually got pruned. The server swallows any predicate error and quietly reads unfiltered, so checking rows alone would pass even if none of this works.

fresh-borzoni (commented):
@leekeiabstraction Thank you for the PR, one clarification about tests, PTAL

fluss-rust/crates/fluss/src/client/table/log_fetch_buffer.rs
leekeiabstraction: Thanks @fresh-borzoni , I noticed that I missed batch statistics part completely. Without the batch statistics calculation, encoding and wiring, Rust client written batches wouldn't be able to be filtered on server side. 

I've raised #4008, #4009 and #4010 to address the batch statistics implementations.

As for integration testing this PR, I'm thinking that I can circle back to it with issue #3845 when the above are all addressed to unblock the merging of this and keep things tightly scoped and manage-able for reviews. WDYT?

fresh-borzoni (approved):
@leekeiabstraction Thank you for the answer, makes sense, LGTM then 👍

PR #3974, opened 2026-08-13
[server] Fix equals/hashCode contract in ISR states and remote log manifest handle

### LINKED ISSUE #3973: [server] Four classes override equals() without hashCode()
### Search before asking

### Fluss version

main (development)

## Summary

Follow-up to #3951, which covers the same defect in the record-batch classes. Four more classes
in `fluss-server` override `equals()` with value semantics but do not override `hashCode()`. In
each case the superclass is `Object` and `equals()` does not delegate to `super.equals()`, so
`hashCode()` is the identity hash and byte-for-byte equal instances hash differently, breaking
the general contract of `Object.hashCode()`:

The four:

| Class | `equals()` compares |
| --- | --- |
| `IsrState.CommittedIsrState` ([IsrState.java:85](https://github.com/apache/fluss/blob/main/fluss-server/src/main/java/org/apache/fluss/server/replica/IsrState.java#L85)) | `isr` |
| `IsrState.PendingExpandIsrState` ([IsrState.java:171](https://github.com/apache/fluss/blob/main/fluss-server/src/main/java/org/apache/fluss/server/replica/IsrState.java#L171)) | `newInSyncReplicaId`, `sentLeaderAndIsr`, `lastCommittedState` |
| `IsrState.PendingShrinkIsrState` ([IsrState.java:244](https://github.com/apache/fluss/blob/main/fluss-server/src/main/java/org/apache/fluss/server/replica/IsrState.java#L244)) | `outOfSyncReplicaIds`, `sentLeaderAndIsr`, `lastCommittedState` |
| `RemoteLogManifestHandle` ([RemoteLogManifestHandle.java:49](https://github.com/apache/fluss/blob/main/fluss-server/src/main/java/org/apache/fluss/server/zk/data/RemoteLogManifestHandle.java#L49)) | `remoteLogManifestPath`, `remoteLogEndOffset` |

## Why these four belong together

`PendingExpandIsrState.equals()` and `PendingShrinkIsrState.equals()` both compare
`lastCommittedState`, which is a `CommittedIsrState`. A correct `hashCode()` for either pending
state therefore has to fold in `CommittedIsrState.hashCode()` — so fixing the pending states
without also fixing `CommittedIsrState` would still leave the composed hash identity-derived and
inconsistent. They have to be done as one change.

`LeaderAndIsr`, the other field these two compare, already implements both `equals()` and
`hashCode()` consistently, so composing it is safe.

## Impact

Any `HashSet`/`HashMap` keyed on these types would fail to find an equal instance. I checked and
nothing keys them today, so this is a latent trap rather than an active failure. The ISR states
in particular are compared frequently in the replica code, and equality already works — it is
only hashing that is broken, which is exactly the kind of thing that bites silently when someone
later reaches for a `Set`.

The Checkstyle `EqualsHashCode` module is not enabled in `tools/maven/checkstyle.xml`, which is
why CI does not flag any of this.

### Solution

Add `hashCode()` to each class over exactly the fields its own `equals()` already compares,
using `Objects.hash(...)` to match the existing idiom in `LeaderAndIsr` in the same module.

One thing I deliberately did **not** change, and would like a maintainer's read on:
`CommittedIsrState` holds both `isr` and `standbyReplicas`, but its `equals()` compares only
`isr`. I kept `hashCode()` consistent with the `equals()` that is there today — `Objects.hash(isr)` —
rather than widening `equals()`, since changing equality semantics for ISR state is a behavioural
change that deserves its own discussion. If the omission of `standbyReplicas` from `equals()` is
itself unintended, I am happy to open a separate issue for it.

## Scope

Two further cases came out of the same sweep and are **not** included, for reasons worth stating:

- `TieringSourceEnumeratorState` (`fluss-flink-common`) also has no `hashCode()`, but its problem
  is larger than a missing hash: `equals()` is implemented as
  `this.toString().equals(that.toString())` against any non-null argument, so it is asymmetric and
  will accept unrelated types. Adding a `hashCode()` alone would only half-fix it, so it needs its
  own issue.
- `WriteResultForBucket` and its subclasses, and the `ModifyColumn*` classes in the Flink 1.18/1.19
  modules, inherit a `hashCode()` from a base class that hashes a subset of the fields their
  `equals()` compares. That is legal — merely a weaker hash — so I have left them alone.

vbhanuchander-lang:
### Purpose

Linked issue: close #3973

Follow-up to #3951 / #3952, which fix the same defect in the record-batch classes.
`CommittedIsrState`, `PendingExpandIsrState`, `PendingShrinkIsrState` and
`RemoteLogManifestHandle` override `equals()` with value semantics but never override
`hashCode()`. Each extends `Object` and none delegates to `super.equals()`, so `hashCode()` is
the identity hash: equal instances hash differently, violating the general contract of
`Object.hashCode()`.

The four belong in one change. `PendingExpandIsrState.equals()` and
`PendingShrinkIsrState.equals()` both compare `lastCommittedState`, a `CommittedIsrState`, so a
correct hash for either pending state has to fold in `CommittedIsrState.hashCode()` — fixing the
pending states without `CommittedIsrState` would leave the composed hash identity-derived.
`LeaderAndIsr`, the other field they compare, already implements both methods consistently.

### Brief change log

- `IsrState.CommittedIsrState`: `hashCode()` over `isr`.
- `IsrState.PendingExpandIsrState`: over `newInSyncReplicaId`, `sentLeaderAndIsr`, `lastCommittedState`.
- `IsrState.PendingShrinkIsrState`: over `outOfSyncReplicaIds`, `sentLeaderAndIsr`, `lastCommittedState`.
- `RemoteLogManifestHandle`: over `remoteLogManifestPath`, `remoteLogEndOffset`.

Each covers exactly the fields its own `equals()` already compares, using `Objects.hash(...)` to
match the existing idiom in `LeaderAndIsr` in the same module.

`CommittedIsrState` holds `standbyReplicas` as well as `isr`, but its `equals()` compares only
`isr`, so `hashCode()` hashes only `isr`. I kept it consistent with the `equals()` that exists
today rather than widening equality, since changing ISR-state equality is a behavioural change
that deserves its own discussion — flagged in #3973 in case that omission is itself unintended.

### Tests

New `IsrStateTest` (7 tests) and `RemoteLogManifestHandleTest` (4 tests): for each class, two
independently constructed equal instances must share a hash code, unequal instances must hash
differently, and equal instances must deduplicate in a `HashSet`.

**6 of the 11 fail without the production change** and all 11 pass with it. The other 5 are the
"different contents hash differently" cases, which identity hashing also satisfies; they are there
to guard against an over-broad hash later.

Verified locally: `mvn test -pl fluss-server -Dtest='IsrStateTest,RemoteLogManifestHandleTest'`
passes, `checkstyle:check` reports 0 violations, and `spotless:apply` leaves the tree clean.

### API and Format

No API or storage-format change; `hashCode()` is additive. The only behavioural difference would
be for a `HashSet`/`HashMap` keyed on these types, and nothing keys them today.

### Documentation

No new feature, no documentation change.

### Scope

Two cases from the same sweep are deliberately excluded, and I would rather say why than leave
them looking overlooked:

- `TieringSourceEnumeratorState` (`fluss-flink-common`) has no `hashCode()` either, but its
  `equals()` is `this.toString().equals(that.toString())` against any non-null argument — it is
  asymmetric and accepts unrelated types. A `hashCode()` alone would half-fix it, so it needs its
  own issue.
- `WriteResultForBucket` and subclasses, and the `ModifyColumn*` classes in the Flink 1.18/1.19
  modules, inherit a `hashCode()` from a base that hashes a subset of the `equals()` fields. That
  is legal, only a weaker hash, so they are untouched.

### Generative AI disclosure

- [x] Yes — Claude Code (Opus 5). All changes reviewed by me before submitting.

### DIFF
--- a/fluss-server/src/main/java/org/apache/fluss/server/replica/IsrState.java
+++ b/fluss-server/src/main/java/org/apache/fluss/server/replica/IsrState.java
@@ -22,6 +22,7 @@ import org.apache.fluss.server.zk.data.LeaderAndIsr;
 
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Objects;
 
 /* This file is based on source code of Apache Kafka Project (https://kafka.apache.org/), licensed by the Apache
  * Software Foundation (ASF) under the Apache License, Version 2.0. See the NOTICE file distributed with this work for
@@ -91,6 +92,11 @@ public interface IsrState {
             }
             CommittedIsrState that = (CommittedIsrState) o;
             return isr.equals(that.isr);
+        }
+
+        @Override
+        public int hashCode() {
+            return Objects.hash(isr);
         }
 
         @Override
@@ -182,6 +188,11 @@ public interface IsrState {
         }
 
         @Override
+        public int hashCode() {
+            return Objects.hash(newInSyncReplicaId, sentLeaderAndIsr, lastCommittedState);
+        }
+
+        @Override
         public String toString() {
             return "PendingExpandIsrState{"
                     + "newInSyncReplicaId="
@@ -252,6 +263,11 @@ public interface IsrState {
             return outOfSyncReplicaIds.equals(that.outOfSyncReplicaIds)
                     && sentLeaderAndIsr.equals(that.sentLeaderAndIsr)
                     && lastCommittedState.equals(that.lastCommittedState);
+        }
+
+        @Override
+        public int hashCode() {
+            return Objects.hash(outOfSyncReplicaIds, sentLeaderAndIsr, lastCommittedState);
         }
 
         @Override
--- a/fluss-server/src/main/java/org/apache/fluss/server/zk/data/RemoteLogManifestHandle.java
+++ b/fluss-server/src/main/java/org/apache/fluss/server/zk/data/RemoteLogManifestHandle.java
@@ -19,6 +19,8 @@ package org.apache.fluss.server.zk.data;
 
 import org.apache.fluss.fs.FsPath;
 
+import java.util.Objects;
+
 /**
  * The remote log manifest handle of a table bucket stored in {@link ZkData.BucketRemoteLogsZNode}.
  *
@@ -56,6 +58,11 @@ public class RemoteLogManifestHandle {
         RemoteLogManifestHandle that = (RemoteLogManifestHandle) o;
         return remoteLogManifestPath.equals(that.remoteLogManifestPath)
                 && remoteLogEndOffset == that.remoteLogEndOffset;
+    }
+
+    @Override
+    public int hashCode() {
+        return Objects.hash(remoteLogManifestPath, remoteLogEndOffset);
     }
 
     @Override
--- a/fluss-server/src/test/java/org/apache/fluss/server/replica/IsrStateTest.java
+++ b/fluss-server/src/test/java/org/apache/fluss/server/replica/IsrStateTest.java
@@ -1,0 +1,119 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.fluss.server.replica;
+
+import org.apache.fluss.server.replica.IsrState.CommittedIsrState;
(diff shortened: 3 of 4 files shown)


vbhanuchander-lang:
Could a committer approve the pending workflow runs here? All three (`CI`, `Check License`,
`Client Integration`) are still at `action_required`, so nothing has executed on this branch and
there are zero completed check runs.

This is the sibling of #3952, which was merged this morning (eaeae17) — same contract, same pattern,
`fluss-server` instead of `fluss-common`.

Verified locally on a clean build of `fluss-server`:

- `IsrStateTest` 7 tests, `RemoteLogManifestHandleTest` 4 tests — 11 passing, 0 failures
- 0 Checkstyle violations, `spotless:check` clean

vbhanuchander-lang:
@wuchong @platinumhamburg could one of you take a look when you have a moment? You have both merged most of the recent `fluss-server/` changes, so tagging you rather than leaving this unrouted.

`LeaderAndIsr` and the remote log manifest classes override `equals` without `hashCode` (and in one case compare only a subset of fields), so instances behave incorrectly in hash-based collections — the kind of thing that silently misbehaves in ISR bookkeeping rather than failing loudly.

Small and self-contained, with tests. This has been open since mid-August; #3952 merged from the same area if that helps for context.

platinumhamburg (approved):
@vbhanuchander-lang Thanks for your contribution. LGTM +1. Wait for the CI passed.

PR #3977, opened 2026-08-13
[server] add support of rebuilding KV from local snapshot

zuston:
### Purpose

This PR enables KV tablets to be rebuilt from valid local snapshots, speeding up recovery—especially when the tablet server restarts without a leader change.

### Brief change log

1. reserve the local latest snapshot to the local KV tablet main dir
2. recover KV tablet from the latest snapshot if it's valid


fluss-server/src/main/java/org/apache/fluss/server/kv/snapshot/RocksIncrementalSnapshot.java
```diff
@@ -323,16 +327,9 @@ protected NativeRocksDBSnapshotResources(
 
         @Override
         public void release() {
```
fresh-borzoni, line 342: release() is empty now and SnapshotRunner is its only caller. Remove it?

fluss-server/src/test/java/org/apache/fluss/server/replica/ReplicaTest.java
```diff
@@ -631,22 +634,42 @@ void testKvReplicaSnapshot(@TempDir File snapshotKvTabletDir) throws Exception {
...
         kvSnapshotStore = testKvSnapshotContext.testKvSnapshotStore;
-        makeKvReplicaAsFollower(kvReplica, 1);
-
-        // check the kv tablet should be null since it has become follower
-        assertThat(kvReplica.getKvTablet()).isNull();
 
-        // make as leader again, should restore from snapshot
+        // Recover as leader after the in-place restart, without an intervening role transition.
         makeKvReplicaAsLeader(kvReplica, 2);
+        assertThat(downloadedRemoteSnapshot).isFalse();
```
fresh-borzoni, line 673: Can we get the other half: invalid checkpoint, assert downloadedRemoteSnapshot is true?

fluss-server/src/main/java/org/apache/fluss/server/replica/Replica.java
```diff
@@ -847,12 +833,29 @@ private Optional<CompletedSnapshot> initKvTablet() {
...
+                            System.currentTimeMillis() - start);
+                } else {
+                    tabletDir =
+                            kvManager.deleteAndCreateTabletDir(
+                                    logTablet.getDataDir(), physicalPath, tableBucket);
+                    downloadKvSnapshots(completedSnapshot, tabletDir.toPath());
+                }
 
                 // as we have downloaded kv files into the tablet dir, now, we can load it
                 kvTablet = kvManager.loadKv(tabletDir, schemaGetter, this::onKvFlushComplete);
```
fresh-borzoni: When the local copy won't open, loadKv throws and we never reach downloadKvSnapshots, so the broken-snapshot handling from #1482 never runs, and the retries at 743 just repeat. Therefore replica ends up stuck with nothing clearing the directory.

Shall we drop the local checkpoint and fall through to the download instead? Not via handleSnapshotBroken though, the remote snapshot is fine here.

fluss-server/src/main/java/org/apache/fluss/server/kv/snapshot/RocksIncrementalSnapshot.java
```diff
@@ -113,12 +113,16 @@ public void notifySnapshotComplete(long completedSnapshotId) {
             uploadedSstFiles.keySet().removeIf(snapshotId -> snapshotId < completedSnapshotId);
             lastCompletedSnapshotId = completedSnapshotId;
         }
+        // The local checkpoint is useful after an in-place restart. Keep the committed snapshot and
+        // remove older or uncommitted checkpoints only after the remote commit succeeds.
+        LocalKvSnapshotUtils.retainOnly(instanceBasePath, completedSnapshotId);
```
fresh-borzoni: Should this be behind a flag? Local recovery is normally opt-in, like Flink has state.backend.local-recovery and also #2179 shipped snapshot retention with a TTL and metrics. 
This turns retention on for every KV bucket with no way off, so feels off.

fresh-borzoni (commented):
@zuston Thank you for the PR, left a couple of comments and questions, PTAL

fluss-server/src/main/java/org/apache/fluss/server/replica/Replica.java
fresh-borzoni, line 674: This deletes the whole tablet dir for standbys too, so the retained checkpoint only survives while the replica stays leader. A standby that gets promoted still downloads from remote.

Is keeping it for standbys in scope, or deliberately out? Looks like what the TODO in processRetryOfflineLeader is anticipating.

zuston: Good point. This PR only optimizes in-place restarts of the leader TabletServer. Standby promotion still recovers from remote storage; maintaining local snapshots on standbys requires a separate lifecycle and is out of scope for this PR. We can track it separately, that is another bigger feature to speed up the KV leader transition.

fluss-server/src/main/java/org/apache/fluss/server/kv/snapshot/RocksIncrementalSnapshot.java
zuston: I have introduced the dedicated option for this feature, so let's reserve this

fluss-common/src/main/java/org/apache/fluss/config/ConfigOptions.java
```diff
@@ -2054,6 +2054,14 @@ public class ConfigOptions {
                             "The interval to perform periodic snapshot for kv data. "
                                     + "The default setting is 10 minutes.");
 
+    public static final ConfigOption<Boolean> KV_SNAPSHOT_LOCAL_RECOVERY_ENABLED =
+            key("kv.snapshot.local-recovery.enabled")
```
fresh-borzoni, line 2058: This one's static, while the other KV knobs can be changed at runtime (kv.snapshot.interval, kv.leader-replica.memory-reserved, the disk ratios). Since it costs disk, someone watching DISK_USAGE_RATIO go up can only switch it off by restarting.

Could we do what the interval does? PeriodicSnapshotManager takes a LongSupplier for it and the key is in ALLOWED_CONFIG_KEYS.

Though switching it off wouldn't free anything by itself. release() would stop keeping new checkpoints, but snap-N is still there until the next remote restore, so notifySnapshotComplete would need to clear the snap-* dirs when it's off.

fresh-borzoni (approved):
@zuston Thank you, LGTM 👍 One non-blocking comment

cc @swuferhong

fluss-common/src/main/java/org/apache/fluss/config/ConfigOptions.java
zuston: > This one's static, while the other KV knobs can be changed at runtime (kv.snapshot.interval, kv.leader-replica.memory-reserved, the disk ratios). Since it costs disk, someone watching DISK_USAGE_RATIO go up can only switch it off by restarting.

Yes, this could be updated dynamiclly to follow up in the next PR. One tip that this feature won't cost extra disk capacity because the snapshot is just a hard link referenced by the rocksdb files.

zuston:
rebased @fresh-borzoni

fluss-server/src/main/java/org/apache/fluss/server/kv/KvManager.java
```diff
@@ -370,18 +372,20 @@ private long getSharedWriteBufferUsage() {
...
+     * <p>When local recovery is disabled, delete directories left by a previous TabletServer
+     * process before accepting assignments. Scan the disk rather than registered tablets so that
+     * directories without logs or table metadata are also removed. Symbolic links below a data
+     * directory are skipped, and cleanup I/O failures are logged without preventing startup.
+     * Directories are atomically renamed before deletion so partially deleted state cannot be
      * reopened as a live KV tablet.
      */
     public void startup() {
-        cleanupStaleKvDirectories();
+        if (!conf.get(ConfigOptions.KV_SNAPSHOT_LOCAL_RECOVERY_ENABLED)) {
```
fresh-borzoni: With the flag on this skips the #4401 cleanup, so #4400 comes back.

Controlled shutdown keeps the old leader's KV dir, since CoordinatorRequestBatch's addNotifyLeaderRequestForTabletServers skips shutting-down servers. The server restarts as a follower, and Replica.dropKv() is a no-op because kvTablet == null. So db/ + snap-N stay on every bucket it now follows, and they survive a table drop.

Reproduced this behaviour in ReplicaTest: with the flag on, db/ + snap-0 survive restart -> follower -> Replica.delete(). With the flag off, the startup cleanup removes them.

Could we keep the dirs at startup, but delete a leftover dir when the replica becomes follower with no open tablet? And how would dirs of buckets that are never assigned back get reclaimed?

@swuferhong WDYT?

fluss-common/src/main/java/org/apache/fluss/config/ConfigOptions.java
fresh-borzoni: Not quite. snap-N hard-links the SSTs at snapshot time. Compaction later removes them from db/, but snap-N keeps them alive until N+1 completes.

Could we at least mention this in the option description?

fluss-server/src/main/java/org/apache/fluss/server/replica/Replica.java
```diff
@@ -891,15 +878,7 @@ private Optional<CompletedSnapshot> initKvTablet() {
...
-                // always create a new dir for the kv tablet
-                File tabletDir =
-                        kvManager.createTabletDir(
-                                logTablet.getDataDir(), physicalPath, tableBucket);
-                // down the snapshot to target tablet dir
-                downloadKvSnapshots(completedSnapshot, tabletDir.toPath());
-
-                // as we have downloaded kv files into the tablet dir, now, we can load it
-                kvTablet = kvManager.loadKv(tabletDir, schemaGetter, this::onKvFlushComplete);
+                kvTablet = restoreKvTablet(completedSnapshot);
```
fresh-borzoni: Not specifically coming from this PR, but noticed while reading and reproduced on main: if recoverKvTablet fails once after a snapshot restore, all 5 initKvTablet attempts fail. Attempt 1's RocksDB is still open, so each retry re-downloads the snapshot and fails with RocksDBException: lock hold by current process .../db/LOCK. The no-snapshot branch already reuses the open tablet. Shall we file an issue for this since we are here?

fluss-server/src/main/java/org/apache/fluss/server/replica/Replica.java
```diff
@@ -973,6 +953,59 @@ private Optional<CompletedSnapshot> initKvTablet() {
...
+                        kvManager.loadKv(
+                                optionalTabletDir.get(), schemaGetter, this::onKvFlushComplete);
+                LOG.info(
+                        "Rebuilt kv tablet for {} of table {} from retained local snapshot {} that costs {} ms.",
+                        tableBucket,
+                        physicalPath,
+                        completedSnapshot.getSnapshotID(),
+                        System.currentTimeMillis() - start);
+                return restoredKvTablet;
+            } catch (KvBuildingException localRecoveryException) {
```
fresh-borzoni: nit: a counter for local hits vs fallbacks would show whether this actually kicks in. #2179 shipped metrics with snapshot retention, is it applicable here, wdyt?

fluss-server/src/main/java/org/apache/fluss/server/kv/snapshot/LocalKvSnapshotUtils.java
```diff
@@ -0,0 +1,208 @@
...
+                                + "Falling back to remote snapshot recovery.",
+                        snapshotDirectory);
+            }
+            return false;
+        }
+
+        Path restoreDirectory = kvTabletDir.toPath().resolve(RESTORE_DIRECTORY_PREFIX + snapshotId);
+        Path activeDbDirectory = RocksDBKvBuilder.getInstanceRocksDBPath(kvTabletDir).toPath();
+        try {
+            FileUtils.deleteDirectory(restoreDirectory.toFile());
```
fresh-borzoni: nit: a leftover .db-restore-<id> gets removed only by a later restore with the same id. 
Could retainOnly also sweep .db-restore-*?

fluss-server/src/main/java/org/apache/fluss/server/replica/Replica.java
```diff
@@ -973,6 +953,59 @@ private Optional<CompletedSnapshot> initKvTablet() {
         return optCompletedSnapshot;
     }
 
+    private KvTablet restoreKvTablet(CompletedSnapshot completedSnapshot) throws Exception {
+        checkNotNull(kvManager);
+        long start = System.currentTimeMillis();
```
fresh-borzoni: nit: clock.milliseconds() like initKvTablet above? Is there any specific reason to use different one?

fluss-server/src/main/java/org/apache/fluss/server/replica/Replica.java
fresh-borzoni: Which setup is this aimed at?

With RF≥2, leadership moves to another replica on controlled shutdown or crash, so this server restarts as a follower. The local copy helps only if leadership returns before the new leader's next snapshot. After that the snapshot id moves on and we download anyway.

So it mostly helps RF=1 or ISR={leader}. Shall we say that in the option description?

fluss-server/src/main/java/org/apache/fluss/server/kv/snapshot/LocalKvSnapshotUtils.java
```diff
@@ -0,0 +1,208 @@
...
+                        || relativePath.getNameCount() != 1
+                        || !expectedFile.getKey().equals(relativePath.toString())
+                        || expectedFile.getValue() < 0) {
+                    return false;
+                }
+
+                Path localFile = snapshotDirectory.resolve(relativePath).normalize();
+                if (!localFile.getParent().equals(snapshotDirectory.normalize())
+                        || !Files.isRegularFile(localFile, LinkOption.NOFOLLOW_LINKS)
+                        || Files.size(localFile) != expectedFile.getValue()) {
```
fresh-borzoni: Validation checks only name and size, so a corrupted SST of the same size passes. RocksDB verifies block checksums only on read, so open succeeds and the fallback never runs.

On read, RocksDBKv.multiGet uses multiGetAsList, which returns null for a key whose read fails. So corrupted keys look missing instead of failing.
So should we verify SST checksums before accepting the local copy?

fresh-borzoni (commented):
@zuston Thank you, left some comments, PTAL

PR #3981, opened 2026-08-13
[flink] Fix ClassNotFoundException for RoaringBitmap in shaded Flink connector JAR

### LINKED ISSUE #3980: [flink] Fix ClassNotFoundException for RoaringBitmap in shaded Flink connector JAR
### Search before asking

### Fluss version

main (development)

### Please describe the bug 🐞

ClassNotFoundException: org.roaringbitmap.RoaringBitmap when using FlussCatalog bitmap functions (rb_build_agg, rb_cardinality, rb_or_agg) introduced in FIP-37.

Steps to reproduce:
1. Build apache/fluss-quickstart-flink from main branch
2. Start environment with docker compose
3. Enter Flink SQL Client
4. CREATE CATALOG fluss_catalog WITH ('type' = 'fluss', 'bootstrap.servers' = 'coordinator-server:9123');
5. USE CATALOG fluss_catalog;
6. Run any SQL using rb_build_agg() or rb_cardinality()

Error:
java.lang.ClassNotFoundException: org.roaringbitmap.RoaringBitmap

Root cause confirmed:
RoaringBitmap is declared as a compile-scope dependency in fluss-flink-common/pom.xml but is NOT listed in the maven-shade-plugin artifactSet in fluss-flink-1.20/pom.xml.

The shade plugin only includes:
- org.apache.fluss:fluss-flink-common
- org.apache.fluss:fluss-client

So org.roaringbitmap:RoaringBitmap is dropped from the final JAR.

Confirmed by: jar tf fluss-flink-1.20-1.0-SNAPSHOT.jar | grep -i roaring→ returns only Fluss wrapper classes, no org/roaringbitmap/* classes

### Solution

Add org.roaringbitmap:RoaringBitmap to the artifactSet includes in the maven-shade-plugin configuration in fluss-flink-1.20/pom.xml (and fluss-flink-1.18, fluss-flink-1.19, fluss-flink-2.2):

<include>org.roaringbitmap:RoaringBitmap</include>

pbanakar:
### Purpose
Linked issue: close #3980 

Fix `ClassNotFoundException: org.roaringbitmap.RoaringBitmap` at runtime when using FlussCatalog bitmap functions introduced in FIP-37 (`rb_build_agg`, `rb_cardinality`, `rb_or_agg`).

### Brief change log

**Root cause:** `RoaringBitmap` is declared as `compile` scope in `fluss-flink-common/pom.xml`, but Maven's dependency mediation resolves it as `test` scope in the version-specific Flink connector modules. This happens because `fluss-flink-common:test-jar` is a `test`-scoped dependency in those modules, and its transitive `RoaringBitmap` dependency overrides the `compile` scope from the main artifact. As a result, the `maven-shade-plugin` excludes `org/roaringbitmap/*` classes from the final shaded connector JAR.

Confirmed by: ./mvnw dependency:tree -pl fluss-flink/fluss-flink-1.20 | grep roaring 
→ org.roaringbitmap:RoaringBitmap:jar:1.3.0:test ← wrong scope

jar tf fluss-flink-1.20-*.jar | grep "org/roaringbitmap" → (empty) ← classes missing from JAR

**Fix:** Add an explicit `compile`-scope `RoaringBitmap` dependency in `fluss-flink-1.18`, `fluss-flink-1.19`, `fluss-flink-1.20`, and `fluss-flink-2.2` `pom.xml` files to override the `test` scope resolution.

After fix: jar tf fluss-flink-1.20-*.jar | grep "org/roaringbitmap" → org/roaringbitmap/RoaringBitmap.class (and all other classes present)

### Tests

Verified end-to-end locally on Flink 1.20:
- Built `apache/fluss-quickstart-flink` from main branch with fix applied
- Ran full user profile quickstart using `rb_build_agg` and `rb_cardinality`
- Confirmed correct results — `unique_visitor_count` and `total_clicks` accumulating in real time with no errors

### API and Format

No API or storage format changes. This is a packaging fix only — 4 `pom.xml` files modified.

### Documentation

No documentation changes. This fix unblocks the Real-Time User Profile quickstart tutorial (PR #2669 ) which demonstrates these functions.

### DIFF
--- a/fluss-flink/fluss-flink-1.18/pom.xml
+++ b/fluss-flink/fluss-flink-1.18/pom.xml
@@ -88,6 +88,12 @@
         </dependency>
 
         <dependency>
+            <groupId>org.roaringbitmap</groupId>
+            <artifactId>RoaringBitmap</artifactId>
+            <version>${roaringbitmap.version}</version>
+        </dependency>
+
+        <dependency>
             <groupId>org.apache.fluss</groupId>
             <artifactId>fluss-server</artifactId>
             <version>${project.version}</version>
@@ -217,6 +223,7 @@
                                 <includes combine.children="append">
                                     <include>org.apache.fluss:fluss-flink-common</include>
                                     <include>org.apache.fluss:fluss-client</include>
+                                    <include>org.roaringbitmap:RoaringBitmap</include>
                                 </includes>
                             </artifactSet>
                             <filters combine.children="append">
--- a/fluss-flink/fluss-flink-1.18/pom.xml
+++ b/fluss-flink/fluss-flink-1.18/pom.xml
@@ -234,6 +234,12 @@
                             </relocations>
                             <filters combine.children="append">
                                 <filter>
+                                    <artifact>org.roaringbitmap:RoaringBitmap</artifact>
+                                    <excludes>
+                                        <exclude>META-INF/versions/**</exclude>
+                                    </excludes>
+                                </filter>
+                                <filter>
                                     <artifact>org.apache.fluss:fluss-flink-common</artifact>
                                     <excludes>
                                         <exclude>org/apache/fluss/flink/action/**</exclude>
--- a/fluss-flink/fluss-flink-1.18/pom.xml
+++ b/fluss-flink/fluss-flink-1.18/pom.xml
@@ -226,6 +226,12 @@
                                     <include>org.roaringbitmap:RoaringBitmap</include>
                                 </includes>
                             </artifactSet>
+                            <relocations combine.children="append">
+                                <relocation>
+                                    <pattern>org.roaringbitmap</pattern>
+                                    <shadedPattern>org.apache.fluss.shaded.org.roaringbitmap</shadedPattern>
+                                </relocation>
+                            </relocations>
                             <filters combine.children="append">
                                 <filter>
                                     <artifact>org.apache.fluss:fluss-flink-common</artifact>
--- a/fluss-flink/fluss-flink-1.18/src/main/resources/META-INF/NOTICE
+++ b/fluss-flink/fluss-flink-1.18/src/main/resources/META-INF/NOTICE
@@ -1,0 +1,9 @@
+fluss-flink-connector
+Copyright 2024-2026 The Apache Software Foundation
+
+This project includes software developed at
+The Apache Software Foundation (http://www.apache.org/).
+
+This project bundles the following dependencies under the Apache Software License 2.0 (http://www.apache.org/licenses/LICENSE-2.0.txt)
+
+- org.roaringbitmap:RoaringBitmap:1.3.0
--- a/fluss-flink/fluss-flink-1.19/pom.xml
+++ b/fluss-flink/fluss-flink-1.19/pom.xml
@@ -226,6 +226,14 @@
                                     <shadedPattern>org.apache.fluss.shaded.org.roaringbitmap</shadedPattern>
                                 </relocation>
                             </relocations>
+                            <filters combine.children="append">
+                                <filter>
+                                    <artifact>org.roaringbitmap:RoaringBitmap</artifact>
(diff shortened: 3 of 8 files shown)


platinumhamburg:
@pbanakar Thanks for investigating and fixing this runtime packaging issue.

At first glance, adding a third-party dependency to the shade configuration of all four version-specific connector modules looks unusual, since these modules previously only assembled Fluss artifacts. After reviewing the dependency and packaging structure again, however, I think handling RoaringBitmap at the final connector level is reasonable in this case.

The version-specific modules exclude all transitive dependencies of `fluss-flink-common`, so RoaringBitmap must be declared explicitly before it can be included in the final connector JAR. Moving the shade step into `fluss-flink-common` would instead change that library's published artifact and create inconsistencies between its main JAR, test JAR, sources, and published POM. Introducing a separate shaded module also seems unnecessarily heavy for this fix.

There are two remaining issues:

1. Please relocate `org.roaringbitmap` to `org.apache.fluss.shaded.org.roaringbitmap`, consistent with the server-side packaging, to avoid potential dependency conflicts.

2. Please add the corresponding NOTICE file to each version-specific connector module. This should also resolve the current License Check failures.

Since this is the first time these version-specific connector modules explicitly include a third-party artifact in their own shade configuration, it would be helpful to get another opinion on this dependency boundary. @polyzos  @wuchong , could you please take a look as well?

pbanakar:
Hi @platinumhamburg, Thank you for your support, addressed both issues, PTAL!
- Added relocation org.roaringbitmap → org.apache.fluss.shaded.org.roaringbitmap consistent with fluss-server packaging
- Added NOTICE files to all 4 Flink connector modules listing RoaringBitmap:1.3.0

Jackeyzhe:
Thanks for the fix — verified the approach works: RoaringBitmap 1.3.0 has no transitive dependencies (the old `org.roaringbitmap:shims` split was dropped in 1.x), so a single `<include>` covers everything, and the base-path classes are correctly relocated.

One packaging nuance worth noting (hit the same class of issue in #3884): RoaringBitmap 1.3.0 ships as a multi-release JAR — it contains `META-INF/versions/11/org/roaringbitmap/ArraysShim.class` alongside the base `org/roaringbitmap/ArraysShim.class`. The shade plugin's relocation doesn't apply to `META-INF/versions/**` entries, so the versions/11 copy will land in the connector uber-jars **unshaded**, under the original `org.roaringbitmap` package.

This is functionally harmless — all references are relocated to the shaded path, and the base-path `ArraysShim` (the pre-Java-11 fallback) is included and relocated, so runtime behavior is correct. But for packaging hygiene, it may be worth excluding the MRJ copies so no unshaded third-party classes remain in the jar, e.g. with a filter on the artifact:

```xml
<filter>
    <artifact>org.roaringbitmap:RoaringBitmap</artifact>
    <excludes>
        <exclude>META-INF/versions/**</exclude>
    </excludes>
</filter>
```

(Excluding them is safe: the base-path class is the compatible fallback; only the Java-11-optimized variant is dropped.)

Aside from that, +1 from me — and I also agree with @platinumhamburg's take on the dependency boundary: shading at the connector level keeps `fluss-flink-common`'s published artifacts unchanged, which seems like the right call here.

pbanakar:
Hi @Jackeyzhe, thank you for the detailed review and for catching the MRJ packaging nuance! Added the filter to exclude META-INF/versions/** from the RoaringBitmap artifact across all 4 connector modules. Verified that no unshaded MRJ entries remain in the shaded JAR. PTAL!

wuchong (approved):
LGTM. 

As a future improvement, we could consider publishing RoaringBitmap through `apache/fluss-shaded`, similar to Arrow and Guava. This would allow modules in the main Fluss repository to depend on a pre-shaded artifact without managing the shading and relocation logic themselves, while also ensuring consistent packaging across modules.

Issue #3905, opened 2026-08-09
[FIP-27] Document clean lake schemas and rolling upgrade requirements


luoyuxia:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/fluss/issues) and found nothing similar.

### Description

Document the user-visible schema behavior and rolling-upgrade requirements introduced by [FIP-27](https://cwiki.apache.org/confluence/display/FLUSS/FIP-27%3A+Remove+Mandatory+System+Columns+From+Fluss+Lake+Tables).

Parent issue: #2411

#### Scope

- Document that newly created lake tables contain only user-defined columns by default.
- Document that existing legacy lake tables are not migrated and remain supported.
- Explain how legacy and clean physical layouts are detected.
- Document the required upgrade order: lake-reading Flink connectors and lake storage plugins, tiering service, and then the Fluss cluster.
- Document that old Flink connectors using `FULL` startup mode cannot read newly created clean lake tables.
- Document rollback limitations after a clean lake table has been created.
- Clarify that system columns remain disabled by default and that any future opt-in behavior is outside FIP-27.
- Add release notes or upgrade notes where appropriate.

#### Acceptance criteria

- The lake table documentation describes both clean and legacy layouts.
- The compatibility matrix and upgrade order are available in user-facing documentation.
- The documentation is consistent with the behavior delivered by the Paimon, Iceberg, and Hudi implementation issues.

### Willingness to contribute

- [ ] I'm willing to submit a PR!

Issue #2411, opened 2026-01-19
Remove Mandatory System Columns From Fluss Lake Tables


luoyuxia:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/fluss/issues) and found nothing similar.

### Description

See FIP https://cwiki.apache.org/confluence/display/FLUSS/FIP-27%3A+Remove+Mandatory+System+Columns+From+Fluss+Lake+Tables

Although not vote yet, I'd like to create the issue firstly.

### Willingness to contribute

- [ ] I'm willing to submit a PR!

hemanthsavasere:
Hi @luoyuxia 
I would like to work on this issue. Can you please assign this to me. Thanks

luoyuxia:
cc @fhan688

fhan688:
> cc [@fhan688](https://github.com/fhan688)

Thanks!  I'll work on this.

PR #3982, opened 2026-08-13
[lake/paimon] Support clean and legacy Paimon lake table schemas

### LINKED ISSUE #3902: [FIP-27] Support clean and legacy Paimon lake table schemas
### Search before asking

### Description

Implement the Paimon part of [FIP-27](https://cwiki.apache.org/confluence/display/FLUSS/FIP-27%3A+Remove+Mandatory+System+Columns+From+Fluss+Lake+Tables).

Parent issue: #2411

New Paimon lake tables should contain only user-defined columns. Existing tables containing `__bucket`, `__offset`, and `__timestamp` must remain readable and writable.

#### Scope

- Create new Paimon tables without the mandatory system columns.
- Detect whether an existing Paimon table uses the legacy or clean layout.
- Accept legacy tables when data lake tiering is re-enabled.
- Keep schema evolution working for both layouts. New business columns should remain before the system columns in a legacy table and should be appended normally in a clean table.
- Make both row and Arrow-batch tiering writers emit system columns only for legacy tables.
- Make lake readers, sorted readers, projections, and historical lookup work with both layouts.
- Reject partial or type-incompatible legacy system-column layouts with a clear error.

#### Acceptance criteria

- A newly created Paimon lake table exposes only user-defined columns.
- Legacy tables continue tiering and can be read with Flink `FULL` startup mode without migration.
- Disabling and re-enabling tiering preserves the existing physical layout.
- Tests cover clean and legacy layouts, writer paths, reader projections, schema evolution, and re-enabling tiering.
- Tests cover log and primary-key tables and partitioned and non-partitioned tables where supported.

### Willingness to contribute

fhan688:
### Purpose

Linked issue: close #3902

Sub-task of the FIP-27 umbrella (#2411): *Remove Mandatory System Columns From Fluss Lake Tables*.

Today every Paimon lake table Fluss creates is forced to carry three mandatory system columns (`__bucket`, `__offset`, `__timestamp`) as its last physical columns. They pollute the schema users see from Paimon and other engines, and impose a system-metadata-based physical layout.

This PR implements the **Paimon** part of FIP-27: newly created Paimon lake tables use a **clean** physical schema containing only user-defined columns, while existing **legacy** tables that still carry the three system columns remain fully readable and writable **without any schema migration**. Both layouts are supported across create, tiering writers, readers, projections, schema evolution, and re-enabling tiering.

### Brief change log

- **Layout detection (single source of truth):** add `PaimonSystemColumns` with a `LakeLayout` enum and `detectLayout(RowType)`. Detection is purely schema-based — no new table property or metadata, so existing tables are never migrated:
  - no system columns → `CLEAN`;
  - all three present, in canonical order, as the trailing columns with compatible types → `LEGACY`;
  - partial / out-of-order / type-incompatible system columns → rejected with a clear `InvalidTableException`.
- **Create:** `PaimonConversions.toPaimonSchema` no longer appends the system columns; new tables are clean. The user-column name-conflict check against system names is kept.
- **Schema evolution:** `toPaimonSchemaChanges` now takes the target layout — for a legacy table a new business column is still inserted **before** the first system column; for a clean table it is appended normally.
- **Compatibility / re-enable tiering:** `PaimonTableValidation.isPaimonSchemaCompatible` detects the existing layout and, for a legacy table, enriches the freshly generated clean schema with the trailing system columns before comparison, so disabling and re-enabling tiering preserves the existing physical layout. The `__timestamp` precision-6→3 relaxation is guarded so it only applies to legacy tables (a clean table has no `__timestamp`).
- **Row tiering writer:** `FlussRecordAsPaimonRow` emits the three system values only for legacy tables; for clean tables the business-field count equals the full row and no system fields are written. Layout is threaded through `PaimonLakeWriter` → `RecordWriter` → `AppendOnlyWriter` /`MergeTreeWriter`.
- **Arrow-batch tiering writer:** for clean tables `AppendOnlyArrowBatchHelper` writes the original `VectorSchemaRoot` directly; for legacy tables it keeps enriching the batch with the `__bucket`/`__offset`/`__timestamp` vectors.
- **Reader / projection:** `PaimonRecordReader` projects and reads `__offset`/`__timestamp` only for legacy tables. For clean tables it emits a sentinel `-1` log offset / timestamp, consistent with the existing `LakeRecordRecordEmitter` contract (`logOffset() >= 0` marks the incremental phase) and the existing `UNKNOWN_OFFSET = -1` convention.
- **Row adapter:** `PaimonRowAsFlussRow` no longer assumes a fixed number of trailing system columns globally; the trailing-system-column count is explicit, which also fixes latent miscounts for nested/projected rows (e.g. the lookup path, where Paimon already projects system columns away).

### Tests

- Adapted `FlussRecordAsPaimonRowTest` to the layout-aware writer (existing cases are legacy-layout).
- **TODO before merge — new clean-layout coverage:** a newly created table exposes only user columns; legacy tables keep tiering and are readable with Flink `FULL` startup; disable+re-enable preserves layout; row and Arrow-batch writer paths; reader projections; schema evolution; and rejection of a partial/type-incompatible legacy layout. Cases should span log and primary-key tables, and partitioned and non-partitioned tables.
- **Build note:** `mvn clean verify` has **not** been run locally (JDK 11 build environment not available on my machine; only JDK 8). Verified signature consistency, imports, and static review; full compile + IT run is pending on a JDK 11 environment.

### API and Format

No public API change. This changes the **physical schema of newly created Paimon lake tables** (clean layout by default). Existing tables are not migrated and keep their current physical format. Compatibility / rolling-upgrade requirements are covered by the umbrella #2411 and documented in #3905:

- New lake storage plugins and Flink connectors continue reading legacy tables.
- New tiering services keep writing the legacy layout when the target table has the system columns.
- Old tiering services must not process newly created clean tables; old Flink connectors using
  `FULL` startup mode must not read newly created clean tables.
- Safe upgrade order: lake-reading Flink connectors and lake storage plugins → tiering service →
  Fluss cluster.

### Documentation

Feature behavior (clean vs. legacy layouts, detection, and the rolling-upgrade/compatibility matrix) is documented separately under #3905. No standalone doc change in this PR.


fluss-lake/fluss-lake-paimon/src/main/java/org/apache/fluss/lake/paimon/utils/PaimonSystemColumns.java
```diff
@@ -0,0 +1,174 @@
...
+ *       mandatory Fluss system columns {@code __bucket}, {@code __offset}, {@code __timestamp} as
+ *       its last three physical columns.
+ * </ul>
+ *
+ * <p>Detection is based purely on the physical Paimon schema, so no extra metadata or table
+ * property is needed and existing tables are never migrated. A table that carries only some of the
+ * system columns, or carries them with an unexpected type, is neither a clean nor a valid legacy
+ * table and is rejected with a clear error.
+ */
+public class PaimonSystemColumns {
```
luoyuxia: I feel like it's a little complex to introduce a new class for it. I think we can just use a method like I did in https://github.com/apache/fluss/pull/2493/changes#diff-e125d6b59322af48ed676cbd230c662ea8e002be31fedfce8e0a66873bdfc23aR156

fluss-lake/fluss-lake-paimon/src/main/java/org/apache/fluss/lake/paimon/utils/PaimonConversions.java
```diff
@@ -203,14 +205,27 @@ public static List<SchemaChange> toPaimonSchemaChanges(List<TableChange> tableCh
...
 
-                String firstSystemColumnName = SYSTEM_COLUMNS.keySet().iterator().next();
-                schemaChanges.add(
-                        SchemaChange.addColumn(
-                                addColumn.getName(),
-                                paimonDataType,
-                                addColumn.getComment(),
-                                SchemaChange.Move.before(
-                                        addColumn.getName(), firstSystemColumnName)));
+                if (lakeLayout == LakeLayout.LEGACY) {
```
luoyuxia: also add a IT to verify that we can still add column for table with legacy system column

fluss-lake/fluss-lake-paimon/src/main/java/org/apache/fluss/lake/paimon/source/PaimonRecordReader.java
```diff
@@ -46,6 +48,14 @@
 /** Record reader for paimon table. */
 public class PaimonRecordReader implements RecordReader {
```
luoyuxia, line 47: I feel like the changes is a little of complex, you can refer to https://github.com/apache/fluss/pull/2493/changes#diff-e125d6b59322af48ed676cbd230c662ea8e002be31fedfce8e0a66873bdfc23a to simplify the code

fluss-lake/fluss-lake-paimon/src/test/java/org/apache/fluss/lake/paimon/LakeEnabledTableCreateITCase.java
luoyuxia, line 1: Also add a test to verify the tiering & union read still work for legacy table. See https://github.com/apache/fluss/pull/2493/changes#diff-9b69c1d14758f0ca8bc71ae7eb8186be3741545178e86226d312c76264f6a0caR418

luoyuxia: Also add a test in this IT to cover the case:
- a table with legacy system column
- disable datalake
- enable datalake again

luoyuxia (commented):
@fhan688 Thanks for the pr. Left minor comments. You can take https://github.com/apache/fluss/pull/2493 as a reference to simplify the code

fluss-lake/fluss-lake-paimon/src/main/java/org/apache/fluss/lake/paimon/source/PaimonRecordReader.java
```diff
@@ -117,11 +141,25 @@ public PaimonRowAsFlussRecordIterator(
                 org.apache.paimon.utils.CloseableIterator<InternalRow> paimonRowIterator,
                 RowType paimonRowType) {
             this.paimonRowIterator = paimonRowIterator;
-            this.logOffsetColIndex = paimonRowType.getFieldIndex(OFFSET_COLUMN_NAME);
-            this.timestampColIndex = paimonRowType.getFieldIndex(TIMESTAMP_COLUMN_NAME);
 
-            int[] project = IntStream.range(0, paimonRowType.getFieldCount() - 2).toArray();
-            projectedRow = ProjectedRow.from(project);
+            int fieldCount = paimonRowType.getFieldCount();
+            if (!hasSystemColumn(paimonRowType)) {
```
luoyuxia: for empty projection, paimonRowType will be empty, hasSystemColumn will throw IndexOutOfBoundsException, we may need to pass `hasSystemColumn` to `PaimonRowAsFlussRecordIterator`

fluss-lake/fluss-lake-paimon/src/main/java/org/apache/fluss/lake/paimon/source/PaimonRecordReader.java
```diff
@@ -117,11 +141,25 @@ public PaimonRowAsFlussRecordIterator(
...
+                // No system columns are read; all projected fields are business fields, and the
+                // log offset / timestamp are not available from the lake table.
+                this.logOffsetColIndex = -1;
```
luoyuxia: I feel like it's a little of complex. I can we can just do like https://github.com/apache/fluss/pull/2493/changes#diff-e125d6b59322af48ed676cbd230c662ea8e002be31fedfce8e0a66873bdfc23aL106 did, we don't need to care about logoffset and timestmap for legecay table.

fluss-lake/fluss-lake-paimon/src/main/java/org/apache/fluss/lake/paimon/utils/PaimonConversions.java
```diff
@@ -264,10 +284,11 @@ public static Schema toPaimonSchema(TableDescriptor tableDescriptor) {
                     column.getComment().orElse(null));
         }
 
-        // add system metadata columns to schema
-        for (Map.Entry<String, DataType> systemColumn : SYSTEM_COLUMNS.entrySet()) {
-            schemaBuilder.column(systemColumn.getKey(), systemColumn.getValue());
-        }
+        // FIP-27: newly created lake tables use a clean physical schema containing only
```
luoyuxia: nit: removee this comments?

fluss-lake/fluss-lake-paimon/src/main/java/org/apache/fluss/lake/paimon/PaimonLakeCatalog.java
```diff
@@ -68,8 +68,9 @@ public class PaimonLakeCatalog implements LakeCatalog {
         // We need __bucket system column to filter out the given bucket
         // for paimon bucket-unaware append only table.
         // It's not required for paimon bucket-aware table like primary key table
-        // and bucket-aware append only table, but we always add the system column
-        // for consistent behavior
+        // and bucket-aware append only table, but legacy tables always carry the system column
+        // for consistent behavior. Under FIP-27 these columns are no longer added to newly created
+        // (clean) tables; they only remain on legacy tables created before FIP-27.
         SYSTEM_COLUMNS.put(BUCKET_COLUMN_NAME, DataTypes.INT());
```
luoyuxia: rename it to LEGACY_SYSTEM_COLUMNS?

fluss-lake/fluss-lake-paimon/src/main/java/org/apache/fluss/lake/paimon/source/PaimonRecordReader.java
```diff
@@ -117,11 +141,25 @@ public PaimonRowAsFlussRecordIterator(
...
+                this.timestampColIndex = -1;
+                projectedRow = ProjectedRow.from(IntStream.range(0, fieldCount).toArray());
+            } else {
+                // Legacy layout: applyProject appended exactly __offset and __timestamp (not
+                // __bucket) as the last two projected fields, so the business fields are all fields
+                // except those trailing two.
+                this.logOffsetColIndex = paimonRowType.getFieldIndex(OFFSET_COLUMN_NAME);
+                this.timestampColIndex = paimonRowType.getFieldIndex(TIMESTAMP_COLUMN_NAME);
+                int[] project = IntStream.range(0, fieldCount - 2).toArray();
```
luoyuxia: note it should be `IntStream.range(0, fieldCount - 3)`, the previous code is wrong.

fluss-lake/fluss-lake-paimon/src/test/java/org/apache/fluss/lake/paimon/tiering/PaimonTieringITCase.java
```diff
@@ -188,6 +189,37 @@ void testTiering() throws Exception {
         }
     }
 
+    @Test
+    void testTieringForLegacyTable() throws Exception {
```
luoyuxia, line 193: I think we can skip this test since testUnionReadLegacyTable already verify tiering works

fluss-lake/fluss-lake-paimon/src/main/java/org/apache/fluss/lake/paimon/source/PaimonRecordReader.java
```diff
@@ -102,6 +117,15 @@ private ReadBuilder applyProject(
         return readBuilder.withProjection(paimonProject);
     }
 
+    /** A legacy table carries the three system columns, ending with {@code __timestamp}. */
+    private static boolean hasSystemColumn(RowType paimonRowType) {
```
luoyuxia: I just found in much places, we call similar code to check system, we can extract it to a common method, and all call this method:
```
public final class PaimonUtils {

      private PaimonUtils() {}

      /**
       * Returns whether the physical Paimon table uses the legacy layout.
       *
       * <p>A legacy table is identified by the presence of the {@code __timestamp}
       * system column.
       */
      public static boolean isLegacyTable(RowType tableRowType) {
          return tableRowType.getFieldIndex(TIMESTAMP_COLUMN_NAME) >= 0;
      }
  }
```

luoyuxia: we also need to call `checkSystemColumns(newSchema.getRowType());` in method `validateAlterTableSchema` to prevent users add such column:
```
public static void validateAlterTableSchema(TableInfo table, Schema newSchema) {
      checkSystemColumns(newSchema.getRowType());
      ...
  }
```

luoyuxia (commented):
@fhan688 Thanks for update. I left minor comments. PTAL. The spark ci fails

fluss-lake/fluss-lake-paimon/src/test/java/org/apache/fluss/lake/paimon/tiering/PaimonTieringITCase.java
luoyuxia: sorry i miss that, we need add it back to make test coverage happy. I'll apend a commit to add it back

luoyuxia:
@fhan688 Hi, I append commits to add test back. PTAL

luoyuxia (approved):

fhan688:
> @fhan688 Hi, I append commits to add test back. PTAL

thanks! LGTM

PR #3985, opened 2026-08-13
[server] Add listRebalances API and expose rebalance timestamps

### LINKED ISSUE #3965: Add completion timestamp and historical listing to RebalanceProgress
### Search before asking

### Motivation

`Admin#listRebalanceProgress` returns `RebalanceProgress`, which carries `rebalanceId`, `status`, a `progress` fraction, and a per-bucket map — but no timestamp for when a rebalance started or finished. In addition, calling `listRebalanceProgress(null)` only ever returns the currently in-progress rebalance task; there's no way to retrieve information about rebalances that have already completed.

This makes it impossible to build any operator-facing tooling that answers "when did the cluster last rebalance?" or "how long did the last rebalance take?" — useful for health dashboards, alerting, and audit trails. Today that information is only recoverable by tailing server logs.

### Solution

- Add `startedAt`/`completedAt` (or equivalent) timestamp fields to `RebalanceProgress`.
- Retain a bounded history of recently-completed rebalance tasks (server-side or ZooKeeper-backed) so that `listRebalanceProgress` (or a new method) can return past tasks, not just the current in-progress one — including their final status (COMPLETED/FAILED/CANCELLED) and timestamps.
- A retention window/limit (e.g. keep the last N completed tasks) would bound the added storage/memory cost.

### Anything else?

_No response_

### Willingness to contribute

morazow:
## Summary

Rebalance operations were observable only while in flight: `listRebalanceProgress(id)`
served the current rebalance, and nothing recorded when one started or finished. This PR
adds the history and the timestamps, then surfaces them through the SQL procedure.

**Storage and the new listing API**

- `RebalanceTask` carries `startedAtMs`/`completedAtMs` (JSON serde v2; v1 znodes read
  back as `-1`, meaning unset).
- On completion or cancellation the final task is also written to a bounded ZooKeeper
  history at `/cluster/rebalance_history/<rebalanceId>` (last 10 kept). A failed history
  write is logged and never fails the transition; ~~corrupt entries are skipped, not
  fatal~~ *(review update)* corrupt entries are skipped on read and deleted at the next
  retention trim.
- New `Admin#listRebalances()` (`ApiKeys.LIST_REBALANCES`) returns a `RebalanceInfo`
  summary - id, status, timestamps - for the current rebalance plus the retained history,
  newest first. Per-bucket detail stays on `listRebalanceProgress(id)`.
- *(review update)* `listRebalanceProgress(id)` now also serves a **historical** rebalance
  id from the ZooKeeper history, with full per-bucket detail and the task's final status.
  An id that is neither current nor retained still throws
  `NoRebalanceInProgressException`, with a message pointing at `listRebalances()`.

**Timestamps on the existing progress RPC**

- `RebalanceProgress` gains `startedAtMs()`/`completedAtMs()`, transported by two new
  optional fields on `ListRebalanceProgressResponse` (`-1` maps to absent).
- `RebalanceProgressJsonSerializer` emits both when set.

**SQL surface**

- `sys.list_rebalance` gains `started_at`/`completed_at` `TIMESTAMP_LTZ(3)` columns.
- ~~Called without a rebalance id it now returns one row per known rebalance (current plus
  retained history, newest first) instead of only the current one. Historical rows carry
  null progress and plan detail.~~
- *(review update)* `sys.list_rebalance` keeps its original single-row semantics; a new
  `sys.list_rebalances` procedure lists summaries (id, status, `started_at`,
  `completed_at`) of all known rebalances in a single RPC, mirroring the Admin API
  one-to-one.
- Docs updated: `engine-flink/procedures.md`, `maintenance/operations/rebalance.md`.

Closes #3965.

## Notes for reviewers

**Where the ZooKeeper read happens.** `listRebalances` needs both in-memory state and
ZooKeeper history. `CoordinatorEventProcessor#processListRebalances` snapshots the current
rebalance on the coordinator event thread, where it mutates, then reads the history on the
`ioExecutor` so the event loop never blocks on ZooKeeper.
`RebalanceManager#listRebalances(RebalanceInfo)` takes that snapshot as a parameter and
touches nothing else, which is what makes it safe off the event thread.

~~**The no-id procedure path issues two RPCs** - `listRebalanceProgress(null)` for the
current entry's detail plus `listRebalances()` for the summaries, stitched client-side -
so a rebalance finishing between them can render one transiently stale row, corrected on
the next call. The alternative is carrying progress and plan detail for the current entry
inside `ListRebalancesResponse`; that was left out to keep the new response summary-only.
Happy to switch to the single-snapshot shape if you prefer it.~~

*(resolved in review)* The procedure split removed the two-RPC stitch entirely:
`sys.list_rebalances` is backed by the single `listRebalances()` RPC, so the transiently
stale row can no longer occur.

**Three commits.** The first adds the storage groundwork and the listing API, the second
the timestamp exposure and the SQL surface, the third addresses review comments (procedure
split, historical-id drill-down, corrupt-entry cleanup). Each compiles and passes its
tests on its own, so the PR is fine to rebase-merge or squash.

## Test Plan

- Unit: `RebalanceTaskJsonSerdeTest` (serde v2 round-trip, v1 document reads back as `-1`),
  `ZooKeeperClientTest` (history retention, idempotent re-register, corrupt-entry
  ~~skip~~ deletion during the trim, sibling-znode isolation), `RebalanceManagerTest`
  (timestamp stamping, failover restore with and without timestamps, no re-completion of
  an already-final restored task, newest-first ordering, empty history, historical-id
  drill-down and unknown-id error), `CoordinatorEventProcessorTest` (the response
  callback completes when the rebalance manager is closed), `ClientRpcMessageUtilsTest`
  (`-1` <-> absent mapping), `RebalanceProgressJsonSerializerTest`.
- Integration: `RebalanceITCase` (end-to-end `listRebalances` and progress timestamps on a
  live cluster), `FlussAuthorizationITCase#testListRebalances` (`DESCRIBE` on
  `Resource.cluster()`, deny and allow), `FlinkProcedureITCase` (both procedures,
  timestamp columns pinned to `Instant` with ordering).


morazow:
**Scope note for reviewers:** per-bucket detail for historical rebalances

This PR keeps `listRebalances()` deliberately summary-only (id, status, timestamps). The per-bucket plan of a retained historical rebalance is already persisted, each history znode stores the full `RebalanceTask`, plan included.

But `listRebalanceProgress(id)` still throws `NoRebalanceInProgressException` for non-current ids, so that plan is not yet reachable through the API.

It is easier to enable this API: `listRebalanceProgress(id)` fall back to the ZooKeeper history when the id is not current, and build the RebalanceProgress from the stored plan. No proto or serde change is needed; the existing response message already carries per-bucket plans. 

Keeping it out of this PR keeps the review surface small and leaves one design point open for discussion:

- History stores plans plus the overall final status, not per-bucket outcomes (those live only in coordinator memory). A historical RebalanceProgress would stamp every bucket with the overall status — exact for COMPLETED, approximate for FAILED/CANCELED. If per-bucket outcomes matter, the follow-up would instead need a serde bump to store RebalanceResultForBucket per bucket. Opinions welcome.

The follow-up also changes one contract: `NoRebalanceInProgressException` would no longer be thrown for ids still in Zookeeper, which I read as the intent of #3965.

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/procedure/ListRebalanceProcessProcedure.java
```diff
@@ -60,23 +64,54 @@ public class ListRebalanceProcessProcedure extends ProcedureBase {
             },
             output =
                     @DataTypeHint(
-                            "ROW<rebalance_id STRING, rebalance_status STRING, rebalance_progress STRING, rebalance_plan STRING>"))
+                            "ROW<rebalance_id STRING, rebalance_status STRING, rebalance_progress STRING, rebalance_plan STRING, started_at TIMESTAMP_LTZ(3), completed_at TIMESTAMP_LTZ(3)>"))
     public Row[] call(ProcedureContext context, @Nullable String rebalanceId) throws Exception {
-        Optional<RebalanceProgress> progressOpt = admin.listRebalanceProgress(rebalanceId).get();
+        if (rebalanceId != null) {
```
affo: **Suggestion: add `sys.list_rebalances` instead of overloading `sys.list_rebalance()`.**

The Admin API cleanly separates `listRebalanceProgress` (detail, one rebalance) from `listRebalances` (summaries, all known) — nice split. But the procedure layer multiplexes both behind `sys.list_rebalance`, which has a few downsides:

- A singular-named procedure returning one row per rebalance is counterintuitive.
- The no-arg call changes from "at most one row" to "one row per known rebalance" — a silent behavior change for anything parsing its output.
- It forces the two-RPC stitch you mention in the PR description, with the transiently-stale-row window; a single `listRebalances` RPC behind a dedicated procedure wouldn't have it.
- Historical rows carry null `rebalance_progress`/`rebalance_plan` columns that a summary-only schema wouldn't need.

Keeping `sys.list_rebalance` with its current semantics (plus the new timestamp columns) and adding `sys.list_rebalances(rebalance_id STRING, rebalance_status STRING, started_at TIMESTAMP_LTZ(3), completed_at TIMESTAMP_LTZ(3))` would mirror the Admin API one-to-one and sidestep all four points.

The resulting shape maps cleanly to a standard list/get pair:

| REST analogy | Admin API | SQL procedure | Returns |
|---|---|---|---|
| `GET /rebalances` | `listRebalances()` | `sys.list_rebalances()` *(new)* | Summaries of all known rebalances — the ongoing one first, then the retained history (id, status, `started_at`, `completed_at`) |
| `GET /rebalances/{id}` | `listRebalanceProgress(id)` | `sys.list_rebalance('<id>')` | Full detail for one rebalance: status, progress, per-bucket plan, timestamps |
| `GET /rebalances/current` | `listRebalanceProgress(null)` | `sys.list_rebalance()` | Full detail of the ongoing rebalance (today's no-arg behavior, unchanged) |

Every id the list returns should then be a valid argument to the get — see my other comment about historical ids.

website/docs/engine-flink/procedures.md
```diff
@@ -537,7 +539,7 @@ CALL [catalog_name.]sys.cancel_rebalance(
 
 **Parameters:**
 
-- `rebalanceId` (optional): The rebalance ID to cancel. If omitted, cancels the most recent rebalance operation. The rebalance ID is returned when calling the `rebalance` procedure.
+- `rebalanceId` (optional): The rebalance ID to cancel. If omitted, cancels the most recent rebalance operation. The ID to pass here is the value returned by the `rebalance` procedure.
```
affo, line 570: `TIMESTAMP_LTZ` is the right type here 👍 — one doc note worth adding: the rendered value depends on the session's `table.local-time-zone` (defaulting to the client JVM's zone), so two operators can see different strings for the same row. Suggest appending to the `started_at`/`completed_at` descriptions: "rendered in the session time zone; set `table.local-time-zone` to control it."

fluss-flink/fluss-flink-common/src/test/java/org/apache/fluss/flink/procedure/FlinkProcedureITCase.java
```diff
@@ -807,13 +807,29 @@ void testListRebalanceProgress() throws Exception {
...
+            List<Row> listResult = CollectionUtil.iteratorToList(rows);
+            assertThat(listResult).isNotEmpty();
+            Row row = listResult.get(0);
+            assertThat(row.getArity()).isEqualTo(6);
+            assertThat(row.getField(0)).isEqualTo(progress.rebalanceId());
+            assertThat(row.getField(1)).isEqualTo(RebalanceStatus.COMPLETED.toString());
+            assertThat((String) row.getField(3)).startsWith("{\"rebalance_id\":");
+            assertThat(row.getField(4)).isNotNull();
+            assertThat(row.getField(5)).isNotNull();
+        }
```
affo, line 834: The new assertions only check `isNotNull()`, which would pass even if the column came back as the wrong conversion class or a nonsense value. Consider pinning them down:

```java
assertThat(row.getField(4)).isInstanceOf(Instant.class);
assertThat((Instant) row.getField(5)).isAfterOrEqualTo((Instant) row.getField(4));
```

That also guards the `-1` ↔ proto-absent mapping against future regressions.

website/docs/maintenance/operations/rebalance.md
```diff
@@ -97,11 +98,16 @@ if (progress.isPresent()) {
     // Check if rebalance is complete
     if (p.status() == RebalanceStatus.COMPLETED) {
         System.out.println("Rebalance completed successfully!");
+        System.out.println("Took " + (p.completedAtMs() - p.startedAtMs()) + " ms");
```
affo: Small caveat on this example: both stamps come from coordinator wall clocks, and after a failover `started_at` and `completed_at` can come from different machines, so the subtraction is skew-sensitive (theoretically even negative). Fine to keep the example — maybe just add "(approximate; timestamps are wall-clock and may span a coordinator failover)".

fluss-server/src/main/java/org/apache/fluss/server/coordinator/rebalance/RebalanceManager.java
affo, line 298: Now that historical rebalance ids are user-visible in the listing, users will copy one and call `sys.list_rebalance('<historical-id>')` / `Admin#listRebalanceProgress(id)` — which throws `NoRebalanceInProgressException` for any id except the current one (pre-existing behavior, but the listing makes it much easier to hit; and if the procedure is split as suggested in my other comment, the list-then-drill-down flow makes it the natural next call). Two options, either fine:

- Serve a summary-only row from the ZK history for a historical id — this also gives the drill-down a proper landing place, or
- Keep the exception but extend the message to point at the listing, e.g. "… is not the current rebalance task; finished rebalances are listed by `Admin#listRebalances()`".

fluss-server/src/main/java/org/apache/fluss/server/zk/ZooKeeperClient.java
```diff
@@ -1677,6 +1680,70 @@ public void deleteRebalanceTask() throws Exception {
         deletePath(RebalanceZNode.path());
     }
 
+    /**
+     * Writes a finished rebalance task to the bounded ZooKeeper-backed history, then trims the
+     * oldest entries beyond {@code retentionCount} (ordered by {@link
+     * RebalanceTask#getCompletedAtMs()}).
+     */
+    public void registerRebalanceHistory(RebalanceTask rebalanceTask, int retentionCount)
```
affo, line 1688: A corrupt history znode currently lives forever: `getRebalanceHistory` skips it (good), but the retention trim only deletes entries present in the decoded listing, so an undecodable child is never cleaned up — the javadoc even documents this. Since the id is known at that point, a cheap fix is to delete undecodable entries during the trim instead of only skipping them.

fluss-server/src/main/java/org/apache/fluss/server/coordinator/rebalance/RebalanceManager.java
```diff
@@ -84,6 +87,9 @@ public class RebalanceManager {
     /** Hardcoded interval for the periodic timeout check: 30 seconds. */
     private static final long TIMEOUT_CHECK_INTERVAL_MS = 30 * 1000L;
 
+    /** Hardcoded bound on the number of completed rebalance tasks retained in ZK history. */
+    private static final int HISTORY_RETENTION_COUNT = 10;
```
affo, line 91: Is hardcoding the retention at 10 intentional for the first cut, or would a `ConfigOption` (e.g. `coordinator.rebalance.history-retention-count`) make sense already? No strong opinion — just worth deciding now, since changing the default later is easier than adding the option after operators start relying on the behavior.

affo (commented):
Thanks, this is a well-structured change — the bottom-up layering is clean, the failover restore paths and the final-status-before-history ordering are handled carefully, and the test coverage matches the claims in the description. I went through it layer by layer and found no blocking issues.

One design suggestion (splitting the SQL procedure instead of overloading it) and a handful of smaller points inline. The procedure split is the only one I'd push on before merge, since it also removes the two-RPC staleness window flagged in the description.

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/procedure/ListRebalanceProcessProcedure.java
morazow: addressed: split the procedure: new sys.list_rebalances lists summaries in one RPC; sys.list_rebalance keeps single-row semantics

https://github.com/apache/fluss/pull/3985/commits/2901d9a1e5a65cb43bb8b26ae396fd5b8ddcb1a2

fluss-flink/fluss-flink-common/src/test/java/org/apache/fluss/flink/procedure/FlinkProcedureITCase.java
morazow: addressed: pinned assertions to Instant with completed >= started ordering in both procedure test blocks

https://github.com/apache/fluss/pull/3985/commits/2901d9a1e5a65cb43bb8b26ae396fd5b8ddcb1a2

fluss-server/src/main/java/org/apache/fluss/server/coordinator/rebalance/RebalanceManager.java
morazow: addressed: intentional for the first cut, consistent with the neighboring hardcoded bounds; config options can be added compatibly later (default unchanged)

https://github.com/apache/fluss/pull/3985/commits/2901d9a1e5a65cb43bb8b26ae396fd5b8ddcb1a2

fluss-server/src/main/java/org/apache/fluss/server/coordinator/rebalance/RebalanceManager.java
morazow: addressed: historical ids now serve full detail (plan, status, timestamps) from ZK history; unknown ids throw with a message pointing at Admin#listRebalances()

https://github.com/apache/fluss/pull/3985/commits/2901d9a1e5a65cb43bb8b26ae396fd5b8ddcb1a2

fluss-server/src/main/java/org/apache/fluss/server/zk/ZooKeeperClient.java
morazow: addressed: undecodable (and unknown-status) history entries are deleted during the retention trim; getRebalanceHistory reuses the new single-entry helper

https://github.com/apache/fluss/pull/3985/commits/2901d9a1e5a65cb43bb8b26ae396fd5b8ddcb1a2

website/docs/engine-flink/procedures.md
morazow: addressed: added the session-time-zone note to started_at/completed_at in both list_rebalance and list_rebalances sections

https://github.com/apache/fluss/pull/3985/commits/2901d9a1e5a65cb43bb8b26ae396fd5b8ddcb1a2

website/docs/maintenance/operations/rebalance.md
morazow: addressed: added the wall-clock/failover caveat into the example's printed output

https://github.com/apache/fluss/pull/3985/commits/2901d9a1e5a65cb43bb8b26ae396fd5b8ddcb1a2

morazow:
addressed: procedure split applied (sys.list_rebalances added, staleness window gone) and all inline comments addressed in the linked commit

https://github.com/apache/fluss/pull/3985/commits/2901d9a1e5a65cb43bb8b26ae396fd5b8ddcb1a2

fluss-server/src/main/java/org/apache/fluss/server/coordinator/rebalance/RebalanceManager.java
morazow: I kept this hardcoded since the neighboring `TIMEOUT_CHECK_INTERVAL_MS` etc are also hardcoded. We can introduce config with default for all 3, what do you think? I can prepare follow up PR.

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/procedure/ListRebalanceProcessProcedure.java
morazow: Yes, this was bothering me, addressed now 🤝

fluss-flink/fluss-flink-common/src/test/java/org/apache/fluss/flink/procedure/FlinkProcedureITCase.java
morazow: Done 👍

affo (approved):
Thanks for addressing! Now looks good to me!

PR #3988, opened 2026-08-14
[c++] Add bounded RecordBatch log reader APIs

### LINKED ISSUE #3987: [c++] Support bounded RecordBatch log reads
### Search before asking

### Description

A related issue exists in the former standalone Rust repository: apache/fluss-rust#546.

### Description

The Rust client already provides `RecordBatchLogReader` for consuming logs up to fixed stopping offsets, but this capability is not exposed through the C++ binding.

The binding should support bounded log reads where callers can:
- assign multiple table buckets to one reader;
- read fixed per-bucket `[starting_offset, stopping_offset)` ranges;
- optionally specify a timestamp range and resolve it per bucket;
- consume one Arrow RecordBatch at a time;
- distinguish a polling timeout from completion;
- stop after every assigned bucket reaches its stopping offset.

### Willingness to contribute

naivedogger:
### Purpose

Linked issue: close #3987

Expose bounded `RecordBatchLogReader` APIs through the C++ binding.

### API example

```cpp
fluss::RecordBatchLogReader reader;
check("create_bounded_reader",
      table.NewScan().CreateRecordBatchLogReader(
          admin,
          assigned_buckets,
          fluss::TimestampRange{
              starting_timestamp_ms,
              stopping_timestamp_ms,
          },
          reader));

while (true) {
    fluss::RecordBatchReadResult result;
    check("next_batch", reader.NextBatch(1000, result));

    if (result.status == fluss::BoundedReadStatus::TimedOut) {
        continue;
    }
    if (result.status == fluss::BoundedReadStatus::Finished) {
        break;
    }

    process(result.batch->GetArrowRecordBatch());
}
```

The API also supports:

- explicit starting and stopping offsets for each bucket;
- starting offsets configured through scanner subscriptions;
- stopping at explicit or latest offsets;
- partitioned and non-partitioned tables.

`TimedOut` allows callers to check cancellation or deadlines before retrying. `Finished` means all assigned buckets have reached their stopping offsets.

### Tests

- `cargo test -p fluss-rs --lib client::table::reader`
- `cargo test -p fluss-cpp --lib`
- `cargo clippy -p fluss-rs -p fluss-cpp --lib -- -D warnings`
- `cargo fmt --check`
- CMake C++ build
- Docker-backed C++ integration tests for explicit offsets, latest offsets,
  timestamp ranges, and partitioned tables

### API and Format

This PR adds backward-compatible public C++ APIs.

### Documentation

Updated the C++ README, API reference, log table guide, and example program.

### DIFF
--- a/fluss-rust/bindings/cpp/include/fluss.hpp
+++ b/fluss-rust/bindings/cpp/include/fluss.hpp
@@ -1930,7 +1930,8 @@ class RecordBatchLogReader {
 
     bool Available() const;
 
-    /// Waits up to timeout_ms for the next batch.
+    /// Waits up to timeout_ms for the next batch. A non-positive timeout_ms
+    /// polls without waiting.
     /// Read `out.status` only when the returned `Result` is `Ok()`.
     /// TimedOut leaves the reader valid for a later retry; Finished means every
     /// subscribed bucket reached its stopping offset.
@@ -1944,7 +1945,8 @@ class RecordBatchLogReader {
     /// retriable REQUEST_TIME_OUT; `out` may then contain a partial set of
     /// complete batches. Only an `Ok()` result means the bounded result is
     /// complete. The timeout is checked between complete Arrow batches and never
-    /// splits a batch already being returned.
+    /// splits a batch already being returned. A non-positive timeout_ms returns
+    /// REQUEST_TIME_OUT without attempting a read.
     /// The reader remains valid after timeout, but retrying is an explicit caller
     /// policy rather than part of this operation; callers should not retry
     /// indefinitely.
--- a/fluss-rust/bindings/cpp/include/fluss.hpp
+++ b/fluss-rust/bindings/cpp/include/fluss.hpp
@@ -1701,9 +1701,6 @@ class TableScan {
 
     std::vector<size_t> ResolveNameProjection() const;
     Result DoCreateScanner(LogScanner& out, bool is_record_batch);
-    Result ResolveTimestampRanges(Admin& admin, const std::vector<TableBucket>& buckets,
-                                  const TimestampRange& range,
-                                  std::vector<RecordBatchLogReadRange>& out) const;
 
     ffi::Table* table_{nullptr};
     std::vector<size_t> projection_;
@@ -1874,6 +1871,18 @@ class LogScanner {
     Result CreateRecordBatchLogReaderUntilOffsets(const std::vector<ReaderStopOffset>& offsets,
                                                   RecordBatchLogReader& out);
 
+    /// Creates a bounded reader from per-bucket offset ranges, subscribing
+    /// every non-empty range. Range validation happens in the SDK.
+    Result CreateRecordBatchLogReaderFromRanges(const std::vector<RecordBatchLogReadRange>& ranges,
+                                                RecordBatchLogReader& out);
+
+    /// Creates a bounded reader for a timestamp range over the given buckets.
+    /// The SDK resolves both timestamps to per-bucket offsets.
+    Result CreateRecordBatchLogReaderBetweenTimestamps(const Admin& admin,
+                                                       const std::vector<TableBucket>& buckets,
+                                                       const TimestampRange& range,
+                                                       RecordBatchLogReader& out);
+
     ffi::LogScanner* scanner_{nullptr};
 };
 
@@ -1912,6 +1921,19 @@ class RecordBatchLogScanner {
    private:
     friend class TableScan;
     explicit RecordBatchLogScanner(ffi::LogScanner* scanner) noexcept;
+
+    /// Transfers this scanner into a reader bounded by per-bucket offset ranges,
+    /// subscribing every non-empty range. The scanner becomes unavailable on
+    /// success.
+    Result CreateRecordBatchLogReaderFromRanges(const std::vector<RecordBatchLogReadRange>& ranges,
+                                                RecordBatchLogReader& out) &&;
+
+    /// Transfers this scanner into a reader bounded by a timestamp range over
+    /// the given buckets. The scanner becomes unavailable on success.
+    Result CreateRecordBatchLogReaderBetweenTimestamps(const Admin& admin,
+                                                       const std::vector<TableBucket>& buckets,
+                                                       const TimestampRange& range,
+                                                       RecordBatchLogReader& out) &&;
 
     LogScanner scanner_;
 };
--- a/fluss-rust/bindings/cpp/include/fluss.hpp
+++ b/fluss-rust/bindings/cpp/include/fluss.hpp
@@ -1936,16 +1936,18 @@ class RecordBatchLogReader {
     /// subscribed bucket reached its stopping offset.
     Result NextBatch(int64_t timeout_ms, RecordBatchReadResult& out);
 
-    /// Drains remaining batches until every stopping offset is reached or the
-    /// total budget of timeout_ms elapses. Batches are *appended* to `out`, so a
-    /// timed-out call can be resumed by calling again with the same `out`.
-    /// This operation is not atomic: on timeout, `out` may contain a partial set
-    /// of complete batches. Only an `Ok()` result means every stopping offset has
-    /// been reached. The timeout never splits a returned Arrow batch; the reader
-    /// stays valid and a retry continues after the batches already appended.
-    /// The timeout result is a retriable REQUEST_TIME_OUT. `timeout_ms` bounds
-    /// one invocation only; callers that retry must enforce their own overall
-    /// deadline or cancellation condition to avoid retrying indefinitely.
+    /// Drains remaining batches using timeout_ms as the total execution budget
+    /// for this invocation. Callers should normally pass the query's remaining
+    /// execution time and invoke this method once.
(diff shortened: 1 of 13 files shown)


naivedogger:
@fresh-borzoni @charlesdong1991 @leekeiabstraction Appreciate a review here when you have time 🙏

fluss-rust/bindings/cpp/src/table.cpp
```diff
@@ -1857,6 +2088,203 @@ Result LogScanner::PollRecordBatch(int64_t timeout_ms, ArrowRecordBatches& out)
...
+    if (!Available()) {
+        return utils::make_client_error("RecordBatchLogReader not available");
+    }
+
+    out.status = BoundedReadStatus::TimedOut;
+    out.batch.reset();
+    auto ffi_result = reader_->record_batch_log_reader_next_batch(timeout_ms);
+    auto result = utils::from_ffi_result(ffi_result.result);
+    if (!result.Ok()) {
+        return result;
```
leekeiabstraction, line 2148: When FFI call fails, we return with `out.status` still set to `TimedOut`, this will cause caller following documentation to retry forever on unretriable errors.

fluss-rust/bindings/cpp/include/fluss.hpp
```diff
@@ -1801,12 +1854,96 @@ class LogScanner {
...
+
+    bool Available() const;
+
+    /// Waits up to timeout_ms for the next batch.
+    /// TimedOut leaves the reader valid for a later retry; Finished means every
+    /// subscribed bucket reached its stopping offset.
+    Result NextBatch(int64_t timeout_ms, RecordBatchReadResult& out);
+
+    /// Drains all remaining batches until every stopping offset is reached.
+    Result CollectAllBatches(ArrowRecordBatches& out);
```
leekeiabstraction: This doesn't have a timeout arg, what happens if it is an error such as a tablet server outage? Will this hang caller forever?

leekeiabstraction (commented):
TY for the PR, added a couple of comments. PTAL.

fluss-rust/bindings/cpp/src/table.cpp
naivedogger: Fixed. If the FFI call fails, `NextBatch()` now leaves `out` as `Finished` with no batch instead of `TimedOut`. This prevents callers from mistaking an actual error for a timeout and retrying forever. Callers should still check the returned Result before using out.

fluss-rust/bindings/cpp/include/fluss.hpp
naivedogger: Fixed. `CollectAllBatches(timeout_ms, out)` now uses `timeout_ms` as the time limit for the entire collection. If the timeout expires before all stopping offsets are reached, it stops and returns `REQUEST_TIME_OUT`; any complete batches already collected remain in `out`. Only `Ok()` means the result is complete. The documentation, examples, and tests have been updated accordingly.

naivedogger:
Thanks @leekeiabstraction for the thorough review! Both comments have been addressed and the updates have been pushed. PTAL when you have a chance.

fresh-borzoni (commented):
@naivedogger Thank you for the PR, I have a structural comment:
A few pieces that are not C++ specific ended up on the C++ side: CollectAllBatches duplicates collect_all_batches() from reader.rs, ResolveTimestampRanges  repeats the lookup already in query_partitioned_offsets, and range validation is split between table.cpp and validate_stopping_offsets. 

Python only wires until_latest today, so it cannot reuse any of it. 
Should these move into reader.rs so the C++ side stays wiring only and other bindings can reuse the SDK?

naivedogger:
@fresh-borzoni Thanks for your comment! You're right that none of those three belong on the C++ side. I pushed a commit moving the drain loop, the timestamp resolution and the range validation into the Rust core, so the C++ side is just wiring now and other bindings can reuse them. PTAL when you have time!

fluss-rust/crates/fluss/src/client/table/reader.rs
```diff
@@ -323,6 +542,184 @@ impl arrow::record_batch::RecordBatchReader for SyncRecordBatchLogReader {
     }
 }
 
+/// Validate that every bucket in a bounded read belongs to the scanned table
+/// and appears exactly once.
+fn validate_read_buckets<'a>(
```
fresh-borzoni, line 626: I think this doesn't check that the bucket exists: a range on a bucket past num_buckets constructs fine and then collect_all_batches() hangs, while the timestamp path fails right away. 
Should we validate the bucket id here?

fluss-rust/crates/fluss/src/client/table/reader.rs
```diff
@@ -323,6 +542,184 @@ impl arrow::record_batch::RecordBatchReader for SyncRecordBatchLogReader {
...
+            if let Some(partition_id) = bucket.partition_id() {
+                bucket_ids_by_partition
+                    .entry(partition_id)
+                    .or_default()
+                    .push(bucket.bucket_id());
+            }
+        }
+
+        let mut resolved: HashMap<TableBucket, i64> = HashMap::new();
+        for (partition_id, bucket_ids) in bucket_ids_by_partition {
```
fresh-borzoni: This resolves one partition per round trip, and new_between_timestamps calls it twice. 
Mb a try_join_all here?

fluss-rust/crates/fluss/src/client/table/reader.rs
```diff
@@ -146,11 +197,155 @@ impl RecordBatchLogReader {
...
+                }
+                None => {
+                    bucket_offsets.insert(range.bucket.bucket_id(), range.starting_offset);
+                }
+            }
+            stopping_offsets.insert(range.bucket, range.stopping_offset);
+        }
+
+        if !bucket_offsets.is_empty() {
+            scanner.subscribe_buckets(&bucket_offsets).await?;
```
fresh-borzoni: new_until_latest takes the reader guard before touching subscriptions, here we subscribe first. 
Should the guard move to the top?

fresh-borzoni (commented):
@naivedogger Thank you for the PR, left a couple of minor comments, PTAL

Also new_between_timestamps and the partitioned branch of OffsetResolver are only tested from C++,  probably we need to test it in rust SDK. WDYT?

fluss-rust/crates/fluss/src/client/table/reader.rs
naivedogger: Good catch. I added bucket ID validation against [0, num_buckets) in the Rust SDK before creating any subscriptions, together with unit coverage for out-of-range IDs.

fluss-rust/crates/fluss/src/client/table/reader.rs
naivedogger: Done. OffsetResolver::resolve() now runs independent partition offset lookups concurrently with try_join_all. The starting and stopping timestamps are still resolved sequentially; the concurrency is across partitions within each resolution.

fluss-rust/crates/fluss/src/client/table/reader.rs
naivedogger: Yes, I moved it to the top. new_from_ranges() now acquires the existing active-reader guard before inspecting or changing subscriptions and uses internal reader-only subscription helpers while holding it. The existing flag is wrapped in an RAII guard so errors or cancellation during async construction release it automatically.

naivedogger:
Thanks @fresh-borzoni, all comments are addressed: bucket validation, concurrent per-partition offset lookups, guard-first reader construction with safe cleanup, and Rust coverage for partitioned timestamp reads. I also fixed some edge-cases in a separate commit. PTAL!

fluss-rust/crates/fluss/src/client/table/reader.rs
```diff
@@ -94,25 +170,32 @@ impl RecordBatchLogReader {
...
-        };
+        validate_read_buckets(
+            scanner.table_id(),
+            scanner.is_partitioned(),
+            scanner.num_buckets(),
+            subscribed.iter().map(|(bucket, _)| bucket),
+        )?;
+
+        let stopping_offsets = query_latest_offsets(admin, &scanner, &subscribed).await?;
+        unsubscribe_completed_buckets(
```
fresh-borzoni, line 190: I think this consumes the caller's subscriptions. Subscribing one bucket of an empty table and then constructing a reader leaves the scanner with none, so a second reader fails with "No buckets subscribed". 

Also the same at line 235 for new_until_offsets. 
Is it Intended?

fresh-borzoni (commented):
@naivedogger Thank you, LGTM overall, one minor comment, PTAL

fluss-rust/crates/fluss/src/client/table/reader.rs
naivedogger: Yes, this is intended. A reader consumes the scanner’s subscriptions, completed buckets are unsubscribed immediately, and any remaining subscriptions are cleaned up when the reader is dropped. Therefore, a shared scanner handle must subscribe again before creating another reader. The empty-table case is simply an already-complete subscription.

fresh-borzoni (approved):
@naivedogger Thank you, LGTM 👍

PR #3994, opened 2026-08-14
[server] Eliminate global lock for KV tablet creation/deletion/load

zuston:
### Purpose

This PR replaces the global lock for KV tablet creation, deletion, and loading with bucket-level locking. Operations on the same bucket remain serialized, while operations on different buckets can proceed concurrently. This is a prerequisite for parallelizing `makeLeaderOrFollower`.


zuston:

fluss-server/src/test/java/org/apache/fluss/server/kv/KvManagerTest.java
```diff
@@ -355,6 +356,27 @@ void testDropKv(String partitionName) throws Exception {
         assertThat(kvManager.getKv(tableBucket1)).isPresent();
     }
 
+    @Test
+    void testKvRunConcurrentlyForDifferentBuckets() throws Exception {
```
fresh-borzoni, line 363: The risky case is dropKv racing getOrCreateKv on the same bucket, where the lock is swapped mid-flight. Can we add that?

fluss-server/src/main/java/org/apache/fluss/server/kv/KvManager.java
```diff
@@ -356,9 +357,20 @@ public Optional<KvTablet> getKv(TableBucket tableBucket) {
...
     public void dropKv(TableBucket tableBucket) {
-        KvTablet dropKvTablet =
-                inLock(tabletCreationOrDeletionLock, () -> currentKvs.remove(tableBucket));
+        inKvLock(
+                tableBucket,
+                () -> {
+                    try {
+                        doDropKv(tableBucket);
+                    } finally {
+                        kvLocks.remove(tableBucket);
```
fresh-borzoni: This runs inside synchronized (lock), so once the entry is gone another thread can create a fresh lock and enter before this one leaves. It only works because it's the last statement for now.
Can it move out of the action rather than depend on that?

fresh-borzoni (commented):
@zuston Thank you for the PR, left some comments and questions, PTAL

Am I right that this doesn't change any timing yet? getOrCreateKv, loadKv and dropKv are all reached under replicaStateChangeLock, via becomeLeaderOrFollower or stopReplicas, so the new locks are taken but never contended.

And LogManager.logCreationOrDeletionLock is still a single global lock, on the same path via getOrCreateLog. Parallelising makeLeaderOrFollower would serialise there instead. Is the log side a follow-up?

zuston:
> Am I right that this doesn't change any timing yet? getOrCreateKv, loadKv and dropKv are all reached under replicaStateChangeLock, via becomeLeaderOrFollower or stopReplicas, so the new locks are taken but never contended.

Yes, these operations are still serialized by `replicaStateChangeLock` , so this change does not introduce parallel execution yet. Within KvManager, operations on different buckets are independent, so bucket-level locking is sufficient to preserve same-bucket serialization without a global KV lock.

> And LogManager.logCreationOrDeletionLock is still a single global lock, on the same path via getOrCreateLog. Parallelising makeLeaderOrFollower would serialise there instead. Is the log side a follow-up?

yes. this could be improved in another PR.

fresh-borzoni (commented):
@zuston Thank you, LGTM 👍 

cc @swuferhong

fresh-borzoni (approved):

zuston:
rebased. @fresh-borzoni

fresh-borzoni (approved):
@zuston Thank you, LGTM 👍 Added the test on top to guarantee we don't have memory leak, will merge once CI passes

PR #3998, opened 2026-08-14
[log] Recover segments containing a single empty batch

### LINKED ISSUE #3997: [Bug] Log segment recovery fails when the segment contains a single empty batch
### Search before asking

### Fluss version

0.9.0 (latest release)

### Please describe the bug 🐞

A TabletServer may fail to recover a log segment that contains exactly one valid empty V0 record batch.

The problem occurs only during recovery. Normal append and follower replication do not use `LogSegment.readNextOffset()` to discover the segment end. When the segment is reopened, recovery calls this method to derive the next offset from the last persisted batch.

A valid empty V0 batch contains no records, so its encoded size is exactly `V0_RECORD_BATCH_HEADER_SIZE`. However, `LogSegment.readWithoutFilter()` currently treats `adjustedMaxSize <= V0_RECORD_BATCH_HEADER_SIZE` as an empty read. It therefore rejects the valid equality case and returns `MemoryLogRecords.EMPTY`.

`LogSegment.readNextOffset()` then receives no batch, and `Iterables.getLast()` throws `NoSuchElementException`, preventing the replica from being loaded.

### Solution

Change the boundary condition from `adjustedMaxSize <= V0_RECORD_BATCH_HEADER_SIZE` to `adjustedMaxSize < V0_RECORD_BATCH_HEADER_SIZE`.

The fix is valid because `V0_RECORD_BATCH_HEADER_SIZE` is the exact minimum size of a structurally valid batch. A smaller buffer cannot contain a complete batch and is still rejected, while an equal-sized buffer may contain a valid empty batch and must be parsed. Existing length and checksum validation continue to reject malformed data.

Add a regression test with a segment containing one empty indexed batch and verify that `readNextOffset()` returns `baseOffset + 1`. This fixes recovery without changing the log format, replication protocol, or public API.

platinumhamburg:
### Purpose

Linked issue: close #3997

### DIFF
--- a/fluss-server/src/main/java/org/apache/fluss/server/log/LogSegment.java
+++ b/fluss-server/src/main/java/org/apache/fluss/server/log/LogSegment.java
@@ -567,7 +567,7 @@ public final class LogSegment {
         int adjustedMaxSize =
                 minOneMessage ? Math.max(maxSize, startOffsetAndSize.getSize()) : maxSize;
         // use V0 size as the lower bound, since V1 header size is large than V0
-        if (adjustedMaxSize <= V0_RECORD_BATCH_HEADER_SIZE) {
+        if (adjustedMaxSize < V0_RECORD_BATCH_HEADER_SIZE) {
             return new FetchDataInfo(offsetMetadata, MemoryLogRecords.EMPTY);
         }
         if (projection == null) {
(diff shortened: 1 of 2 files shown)


swuferhong (approved):
LGTM +1  waiting for ci pass.

PR #3999, opened 2026-08-14
[server] Parallelize replica leader transitions

zuston:
### Purpose

This PR reduces the latency of replica leader transitions by processing independent buckets concurrently within a `NotifyLeaderAndIsr` request.
Previously, `ReplicaManager.makeLeaders()` transitioned replicas sequentially. This increased request latency when multiple replicas became leaders together, especially for primary-key tables where each transition may initialize RocksDB by restoring a remote snapshot and replaying logs.

Snapshot files within a single bucket are already downloaded concurrently through `ioExecutor`. The main benefit of this change is parallelizing the bucket-level transition: independent buckets can overlap snapshot download, RocksDB initialization and replay, snapshot metadata reads, and remote-log setup.

### Brief change log

Introduce the `tablet-server.replica-transition-thread-num` option to control concurrency, with a default value of 10.

In this first phase, `makeFollowers` remains on the existing state-change executor and is not moved into the background transition pool.

### DIFF
--- a/fluss-common/src/main/java/org/apache/fluss/config/ConfigOptions.java
+++ b/fluss-common/src/main/java/org/apache/fluss/config/ConfigOptions.java
@@ -649,6 +649,14 @@ public class ConfigOptions {
                             "The rack for the tabletServer. This will be used in rack aware bucket assignment "
                                     + "for fault tolerance. Examples: `RACK1`, `cn-hangzhou-server10`");
 
+    public static final ConfigOption<Integer> TABLET_SERVER_REPLICA_TRANSITION_THREAD_NUM =
+            key("tablet-server.replica-transition-thread-num")
+                    .intType()
+                    .defaultValue(10)
+                    .withDescription(
+                            "The maximum number of replica role transitions that can run "
+                                    + "concurrently in a TabletServer.");
+
     public static final ConfigOption<Double> TABLET_SERVER_ADVERTISED_RESOURCE_CPU_CORES =
             key("tablet-server.advertised-resource.cpu-cores")
                     .doubleType()
--- a/fluss-common/src/main/java/org/apache/fluss/config/ConfigOptions.java
+++ b/fluss-common/src/main/java/org/apache/fluss/config/ConfigOptions.java
@@ -405,7 +405,7 @@ public class ConfigOptions {
     public static final ConfigOption<Integer> SERVER_IO_POOL_SIZE =
             key("server.io-pool.size")
                     .intType()
-                    .defaultValue(10)
+                    .defaultValue(2)
                     .withDescription(
                             "The size of the IO thread pool to run blocking operations for both coordinator and tablet servers. "
                                     + "This includes discard unnecessary snapshot files, transfer kv snapshot files, "
--- a/fluss-server/src/main/java/org/apache/fluss/server/replica/ReplicaManager.java
+++ b/fluss-server/src/main/java/org/apache/fluss/server/replica/ReplicaManager.java
@@ -123,9 +123,11 @@ import org.apache.fluss.server.zk.ZooKeeperClient;
 import org.apache.fluss.server.zk.data.LeaderAndIsr;
 import org.apache.fluss.server.zk.data.lake.LakeTableSnapshot;
 import org.apache.fluss.utils.ByteArraySlice;
+import org.apache.fluss.utils.ExecutorUtils;
 import org.apache.fluss.utils.FileUtils;
 import org.apache.fluss.utils.FlussPaths;
 import org.apache.fluss.utils.clock.Clock;
+import org.apache.fluss.utils.concurrent.ExecutorThreadFactory;
 import org.apache.fluss.utils.concurrent.Scheduler;
 
 import org.slf4j.Logger;
@@ -152,6 +154,8 @@ import java.util.Set;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.locks.Lock;
@@ -196,6 +200,7 @@ public class ReplicaManager implements ServerReconfigurable {
 
     private final TabletServerMetadataCache metadataCache;
     private final ExecutorService ioExecutor;
+    private final ExecutorService replicaTransitionExecutor;
     private final ProjectionPushdownCache projectionsCache = new ProjectionPushdownCache();
     private final Lock replicaStateChangeLock = new ReentrantLock();
 
@@ -361,6 +366,10 @@ public class ReplicaManager implements ServerReconfigurable {
         this.userMetrics = userMetrics;
         this.clock = clock;
         this.ioExecutor = ioExecutor;
+        this.replicaTransitionExecutor =
+                Executors.newFixedThreadPool(
+                        conf.get(ConfigOptions.TABLET_SERVER_REPLICA_TRANSITION_THREAD_NUM),
+                        new ExecutorThreadFactory("tablet-server-replica-transition-" + serverId));
         this.minInSyncReplicas = conf.get(ConfigOptions.LOG_REPLICA_MIN_IN_SYNC_REPLICAS_NUMBER);
         this.scannerManager = checkNotNull(scannerManager, "scannerManager");
         // Historical lookup cache capacity currently uses only the first data volume.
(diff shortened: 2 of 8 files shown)


zuston:

fluss-common/src/main/java/org/apache/fluss/config/ConfigOptions.java
```diff
@@ -649,6 +649,14 @@ public class ConfigOptions {
                             "The rack for the tabletServer. This will be used in rack aware bucket assignment "
                                     + "for fault tolerance. Examples: `RACK1`, `cn-hangzhou-server10`");
 
+    public static final ConfigOption<Integer> TABLET_SERVER_REPLICA_TRANSITION_THREAD_NUM =
+            key("tablet-server.replica-transition-thread-num")
+                    .intType()
+                    .defaultValue(10)
```
fresh-borzoni, line 655: The description credits restoring RocksDB from remote snapshots, but initKvTablet already fans that download out over ioExecutor one task per file. Is the win here mainly overlapping one bucket's RocksDB replay with another bucket's download, plus the ZK and manifest reads?

fluss-common/src/main/java/org/apache/fluss/config/ConfigOptions.java
```diff
@@ -649,6 +649,14 @@ public class ConfigOptions {
                             "The rack for the tabletServer. This will be used in rack aware bucket assignment "
                                     + "for fault tolerance. Examples: `RACK1`, `cn-hangzhou-server10`");
 
+    public static final ConfigOption<Integer> TABLET_SERVER_REPLICA_TRANSITION_THREAD_NUM =
+            key("tablet-server.replica-transition-thread-num")
```
fresh-borzoni: remote.log.data-transfer-thread-num and kv.snapshot.transfer-thread-num were both deprecated in favour of server.io-pool.size. 
Should this be derived from that as well, rather than a new option?

fluss-server/src/test/java/org/apache/fluss/server/replica/ReplicaManagerTest.java
```diff
@@ -1796,6 +1799,63 @@ void becomeLeaderOrFollower() throws Exception {
...
+                                        INITIAL_BUCKET_EPOCH))),
+                future::complete);
+
+        assertThat(future.get())
+                .containsExactlyInAnyOrder(
+                        new NotifyLeaderAndIsrResultForBucket(firstBucket),
+                        new NotifyLeaderAndIsrResultForBucket(secondBucket));
+        assertThat(spyingReplicaManager.getReplicaOrException(firstBucket).isLeader()).isTrue();
+        assertThat(spyingReplicaManager.getReplicaOrException(secondBucket).isLeader()).isTrue();
+    }
```
fresh-borzoni, line 1890: Could we also cover one bucket failing while the others succeed? That is the part this refactor could break, and a stale bucket epoch is enough to make Replica.makeLeader throw for just that bucket.

fluss-server/src/test/java/org/apache/fluss/server/replica/ReplicaManagerTest.java
```diff
@@ -1796,6 +1799,63 @@ void becomeLeaderOrFollower() throws Exception {
...
+    @Test
+    void testMakeLeadersInParallel() throws Exception {
+        ReplicaManager spyingReplicaManager = spy(replicaManager);
+        replicaManager = spyingReplicaManager;
+
+        CountDownLatch leadersStarted = new CountDownLatch(2);
+        doAnswer(
+                        invocation -> {
+                            leadersStarted.countDown();
+                            if (!leadersStarted.await(10, TimeUnit.SECONDS)) {
```
fresh-borzoni: This assumes the default pool size and with tablet-server.replica-transition-thread-num=1 the test waits the full 10s before failing. Could we set the option in getServerConf() and shorten the timeout?

fluss-server/src/test/java/org/apache/fluss/server/replica/ReplicaManagerTest.java
```diff
@@ -1796,6 +1799,63 @@ void becomeLeaderOrFollower() throws Exception {
                 replicaManager.getReplicaOrException(tb), true, 1, INITIAL_BUCKET_EPOCH);
     }
 
+    @Test
+    void testMakeLeadersInParallel() throws Exception {
+        ReplicaManager spyingReplicaManager = spy(replicaManager);
+        replicaManager = spyingReplicaManager;
```
fresh-borzoni: Is this assignment needed? spyingReplicaManager is used everywhere below.

fluss-server/src/main/java/org/apache/fluss/server/replica/ReplicaManager.java
```diff
@@ -1389,30 +1398,50 @@ private void makeLeaders(
...
+                                () -> makeLeader(replica, data), replicaTransitionExecutor));
             } catch (Exception e) {
                 LOG.error("Error make replica {} to leader", tb, e);
                 result.put(
                         tb, new NotifyLeaderAndIsrResultForBucket(tb, ApiError.fromThrowable(e)));
             }
         }
+
+        for (CompletableFuture<NotifyLeaderAndIsrResultForBucket> future : makeLeaderFutures) {
+            NotifyLeaderAndIsrResultForBucket leaderResult = future.join();
```
fresh-borzoni, line 1418: TabletManagerBase.closeTabletsConcurrently uses allOf so every future completes before it returns. Joining in a loop means an escaping Error leaves the remaining transitions running with replicaStateChangeLock already released. 

Could we follow the same pattern?

fluss-server/src/main/java/org/apache/fluss/server/replica/ReplicaManager.java
```diff
@@ -2549,6 +2578,7 @@ public Replica getReplica() {
     public static final class OfflineReplica implements HostedReplica {}
 
     public void shutdown() throws InterruptedException {
+        ExecutorUtils.gracefulShutdown(5, TimeUnit.SECONDS, replicaTransitionExecutor);
```
fresh-borzoni: shutdownNow() on replicaStateChangeExecutor used to interrupt the thread doing the transition. It now interrupts a thread waiting in join(), so the workers are only interrupted here, after scheduler.shutdown() and scannerManager.close(). 
Could we create the pool in TabletServer like ioExecutor and drain it at TabletServer:483, so gracefulShutdown reaches the worker threads again?

fresh-borzoni (commented):
@zuston Thank you for the PR, left some comments and questions, PTAL

fluss-common/src/main/java/org/apache/fluss/config/ConfigOptions.java
zuston: good point! I have updated the PR description.

fluss-server/src/test/java/org/apache/fluss/server/replica/ReplicaManagerTest.java
zuston: added.

fluss-server/src/main/java/org/apache/fluss/server/tablet/TabletServer.java
```diff
@@ -486,6 +495,14 @@ CompletableFuture<Void> stopServices() {
                 exception = ExceptionUtils.firstOrSuppressed(t, exception);
             }
 
+            try {
+                if (replicaTransitionExecutor != null) {
+                    ExecutorUtils.gracefulShutdown(5, TimeUnit.SECONDS, replicaTransitionExecutor);
```
fresh-borzoni: drain the transition pool before replicaStateChangeExecutor, since allOf().join() ignores the interrupt and the workers are what need signalling.

fresh-borzoni (commented):
@zuston Thank you, LGTM overall, one comment, PTAL

fresh-borzoni (commented):
@zuston Thank you, LGTM 👍 

cc @swuferhong

fresh-borzoni (approved):

fluss-common/src/main/java/org/apache/fluss/config/ConfigOptions.java
swuferhong, line 655: The configured value only bounds the number of bucket transitions; it does not bound the actual recovery concurrency.

Each KV leader recovery creates a private `RemoteLogFetcher` executor, whose default size is `3`. With the proposed default of `10` transitions, a mass leader election may therefore create up to `30` remote-log download threads, in addition to the `10` transition threads, shared snapshot-transfer work, and RocksDB background initialization/compaction threads.

This burst happens precisely during failover or rebalancing, when the server is already under pressure, and may amplify CPU, native-memory, file-descriptor, disk-I/O, and remote-storage load. Ten concurrent RocksDB initializations are also potentially expensive even when remote-log replay is not required.

Could we either:
- use a substantially more conservative default, such as 1 or 2;
- share and globally bound the remote-log recovery executor instead of creating one per bucket; or
provide a multi-bucket recovery benchmark/stress test demonstrating that the default of 10 is safe?

I think this should be evaluated before making 10 the production default. WDYT @platinumhamburg

fluss-server/src/main/java/org/apache/fluss/server/tablet/TabletServer.java
```diff
@@ -478,6 +487,14 @@ CompletableFuture<Void> stopServices() {
...
```
swuferhong, line 492: `ExecutorUtils.gracefulShutdown()` eventually calls `shutdownNow()`, but queued `CompletableFuture$AsyncSupply` tasks returned by `shutdownNow()` are only discarded; their corresponding futures are not automatically completed or cancelled.

If the pool is saturated, a running transition exceeds the five-second timeout, and other transitions remain queued, `makeLeaders()` can stay blocked forever in `CompletableFuture.allOf(...).join()`. Shutting down `replicaStateChangeExecutor` afterwards cannot release that thread because `join()` is not interruptible. 

Meanwhile,` ReplicaManager` and tablet resources continue shutting down and may race with transition workers that are still running.

Could we explicitly track and cancel the transition futures during forced shutdown, use an interruptible wait, and ensure every submitted future reaches a terminal state? Please also add a deterministic shutdown test with a one-thread transition pool, one blocked transition, and a second queued transition.

fluss-server/src/test/java/org/apache/fluss/server/replica/ReplicaManagerTest.java
```diff
@@ -1796,6 +1799,96 @@ void becomeLeaderOrFollower() throws Exception {
...
+        doAnswer(
+                        invocation -> {
+                            leadersStarted.countDown();
+                            if (!leadersStarted.await(10, TimeUnit.SECONDS)) {
+                                throw new AssertionError(
+                                        "Leader transitions did not run in parallel");
+                            }
+                            return invocation.callRealMethod();
+                        })
+                .when(spyingReplicaManager)
```
swuferhong, line 1835: Could we avoid `Mockito` for test?

fluss-server/src/main/java/org/apache/fluss/server/replica/ReplicaManager.java
```diff
@@ -1389,30 +1394,54 @@ private void makeLeaders(
                         .map(NotifyLeaderAndIsrData::getTableBucket)
                         .collect(Collectors.toSet()));
 
+        List<CompletableFuture<NotifyLeaderAndIsrResultForBucket>> makeLeaderFutures =
+                new ArrayList<>(replicasToBeLeader.size());
         for (NotifyLeaderAndIsrData data : replicasToBeLeader) {
```
swuferhong, line 1399: Can we guarantee or validate that `replicasToBeLeader `contains each `TableBucket` only once?

The normal coordinator path currently builds the request from a map, but the RPC representation is a repeated list and the `TabletServer` does not enforce uniqueness. Duplicate entries would schedule concurrent transitions for the same Replica. 

Replica.makeLeader() is protected by its per-replica lock, but `registerReplica()` and the lake-snapshot refresh run before that lock, so their side effects can still race.

Please either reject duplicate buckets, deduplicate the request before submission, or serialize transition tasks by `TableBucket`.

swuferhong (commented):
Hi, @zuston, thanks for your contributions, I left some comments:

fluss-common/src/main/java/org/apache/fluss/config/ConfigOptions.java
zuston: Makes sense. For now, we could use a more conservative default. How about setting it to 2?

fluss-server/src/main/java/org/apache/fluss/server/replica/ReplicaManager.java
zuston: Good point, this is necessary to ensure the unique before entering the parallelize transitions

PR #4001, opened 2026-08-14
[server] Support primary-key writes to historical partitions

### LINKED ISSUE #3633: [FIP-28] Support writing to historical partitions
### Search before asking

### Description

Parent issue: #3631

Support writing records that target historical partitions for datalake-enabled tables. After point lookup support is available, the remaining FIP-28 work is to define and implement the write path for records whose partition no longer exists as an active Fluss partition but still exists in lake storage.

The implementation should cover:

- The expected semantics for writes targeting historical partitions, including whether the write should update lake storage directly, recreate Fluss-side metadata, or be rejected under specific conditions.
- Metadata refresh/lifecycle behavior needed once historical partition writes are supported.
- Integration with partition resolution in writers.
- Lake-format-specific implementation, starting with Paimon.
- Tests for accepted writes, rejected writes, schema evolution, and lookup-after-write behavior.

This issue intentionally remains separate from point lookup because write support has different lifecycle, consistency, and metadata refresh requirements.

### Willingness to contribute

luoyuxia:
Generated-by: Codex (GPT-5) following [the guidelines](https://github.com/apache/fluss/blob/main/AGENTS.md)

### Purpose

Linked issue: close #3633

Allow tablet servers to process primary-key writes that target the shared historical partition. The RPC carries the original partition name so the server can preserve partition identity while multiple original partitions share one physical bucket.

This PR provides the server-side write path only. Client-side routing is deferred to a follow-up PR. Historical KV recovery, snapshots, and cleanup are also left as explicit follow-up work.

### Brief change log

- Extend the PutKv bucket RPC with optional original-partition context and preserve it during server decoding.
- Dispatch historical writes to the bounded historical executor while keeping normal writes on the existing path.
- Reuse the existing KvTablet merge, WAL, backpressure, segmented-flush, and retry behavior.
- Namespace local historical keys by original partition and primary key, fall back to lake data on a local miss, and retain delete tombstones locally.
- Include the local historical overlay in historical lookups.

### Tests

- `./mvnw clean install -DskipTests -DskipITs -pl fluss-protogen,fluss-rpc`
- `./mvnw -pl fluss-server -DskipITs -Dtest=HistoricalPkWriteProcessorTest,HistoricalKvKeyEncoderTest,ServerRpcMessageUtilsTest test`
- Result: 6 tests passed; Checkstyle reported 0 violations and Spotless passed.

### API and Format

- Extends the PutKv RPC bucket message with the optional `original_partition_name` field.
- Introduces a local historical KV key format: length-prefixed UTF-8 original partition name followed by the physical primary key.
- Does not change the public Java client API or enable client-side historical routing.

### Documentation

The change is covered by FIP-28 and issue #3633. User-facing documentation is deferred until the client write path is enabled.


luoyuxia:

fluss-server/src/main/java/org/apache/fluss/server/replica/Replica.java
```diff
@@ -733,6 +763,40 @@ private void createKv() {
         startPeriodicKvSnapshot(snapshotUsed.orElse(null));
     }
 
+    private void createHistoricalKv() {
+        checkNotNull(kvManager);
+        try {
+            // Historical buckets are a local overlay over lake data. Start with an empty tablet
```
wuchong: This initializes an empty historical overlay and starts serving it before the retained WAL is replayed. After a restart or leader change, a delete can miss both the empty local state and a lagging lake snapshot, be acknowledged as a no-op, and allow a previously written row to reappear permanently. Please replay the retained WAL from the last durably tiered offset—or keep the replica unavailable—before accepting historical reads or writes.

fluss-server/src/main/java/org/apache/fluss/server/replica/Replica.java
```diff
@@ -733,6 +763,40 @@ private void createKv() {
...
+                            arrowCompressionInfo,
+                            this::onKvFlushComplete);
+            if (kvTablet.getRocksDBStatistics() != null) {
+                bucketMetricGroup.registerRocksDBStatistics(kvTablet.getRocksDBStatistics());
+            }
+            // TODO: Recover historical KV state from retained WAL after a restart or leadership
+            // change. WAL replay must rebuild the composite key from the original partition name
+            // and primary key because one physical bucket can contain writes for multiple
+            // partitions. Historical KV tablets deliberately do not use KV snapshots.
+            // TODO: Clean up historical KV state after the corresponding WAL is fully tiered to
```
wuchong, line 757: The local historical overlay is never reclaimed after its WAL is durably tiered. Every distinct historical key and tombstone therefore remains in RocksDB for the lifetime of the replica; sustained historical updates can grow the shared data disk without bound and eventually trigger write protection for unrelated tablets. 

Is this planed to expire RocksDB records by the row TTL when https://github.com/apache/fluss/pull/3621 is merged?

fluss-server/src/main/java/org/apache/fluss/server/replica/Replica.java
```diff
@@ -1188,6 +1260,73 @@ public LogAppendInfo putRecordsToLeader(
...
+                    }
+                    if (!isHistoricalPartition()) {
+                        throw new InvalidPartitionException(
+                                "Historical write request must target a historical partition.");
+                    }
+
+                    validateInSyncReplicaSize(requiredAcks);
+                    KvTablet kv = this.kvTablet;
+                    checkNotNull(kv, "KvTablet for the historical replica shouldn't be null.");
+                    // TODO: Move fallback lake lookup outside leaderIsrUpdateLock and kvLock
```
wuchong: Because putHistoricalRecordsToLeader holds leaderIsrUpdateLock across kv.putHistoricalAsLeader, and that call performs synchronous fallback lake/file I/O after acquiring kvLock, a slow or stuck cache miss can block leader transitions indefinitely. Please perform the fallback outside these locks and revalidate the leader epoch immediately before applying the write so that an old leader cannot commit.

fluss-server/src/main/java/org/apache/fluss/server/kv/KvWriteProcessor.java
```diff
@@ -0,0 +1,446 @@
...
+            walBuilder.append(ChangeType.UPDATE_AFTER, latestSchemaRow.replaceRow(newValue.row));
+            stateAccessor.update(key, newValue.encodeValue(), logOffset + 1);
+            return logOffset + 2;
+        }
+    }
+
+    private byte[] getFromState(
+            KvPreWriteBuffer.Key encodedPrimaryKey, KvStateAccessor stateAccessor)
+            throws Exception {
+        return stateAccessor.lookup(encodedPrimaryKey).value();
```
wuchong: Would a lazy-miss, unlock-and-retry flow work here? Keep the normal attempt entirely local so WAL-mode full overwrites and prewrite/RocksDB hits never query the lake. If the historical accessor reaches a genuine local miss, surface an internal NEEDS_LAKE_LOOKUP outcome and let the existing failure path roll back any staged prewrite mutations. Capture the leader epoch, leave both kvLock and leaderIsrUpdateLock, resolve only the missing keys from the lake asynchronously, and memoize both FOUND and ABSENT for this logical request. Then reacquire the locks and retry only if the captured epoch is still current; on retry, consult local state first and use the memoized lake result only while the key remains absent. The admission and ordering token should remain held across the asynchronous gap so later writes for the same historical key cannot overtake it. This preserves the fast paths, avoids lake I/O under the locks, and fences stale old-leader commits.

fluss-server/src/main/java/org/apache/fluss/server/replica/historical/HistoricalPartitionManager.java
```diff
@@ -0,0 +1,326 @@
...
+            PutKvDataForBucket putData,
+            @Nullable int[] targetColumns,
+            MergeMode mergeMode,
+            int requiredAcks) {
+        try {
+            String originalPartitionName =
+                    checkNotNull(
+                            putData.originalPartitionName(),
+                            "originalPartitionName must not be null");
+            HistoricalWriteKey orderingKey =
```
wuchong: Using the original partition name as part of the ordering key allows writes for two original partitions mapped to the same physical historical TableBucket to execute concurrently. The WAL and idempotent writer sequence are scoped to the physical TableBucket, so a later batch can append first and make the earlier batch fail with OutOfOrderSequenceException. Please serialize historical writes by TableBucket only; the original partition name should remain part of the state-key namespace, but not the executor ordering key.

fluss-server/src/main/java/org/apache/fluss/server/utils/ServerRpcMessageUtils.java
```diff
@@ -1101,22 +1101,49 @@ public static FetchLogResponse makeFetchLogResponse(
...
+            TableBucket tableBucket =
                     new TableBucket(
                             tableId,
                             putKvReqForBucket.hasPartitionId()
                                     ? putKvReqForBucket.getPartitionId()
                                     : null,
                             putKvReqForBucket.getBucketId());
-            produceEntryData.put(tb, kvRecords);
+            PutKvDataForBucket previous =
+                    putKvData.putIfAbsent(
```
wuchong: Different original partitions can legitimately map to the same physical historical TableBucket, but this map is keyed only by TableBucket. A per-node request containing both entries therefore treats the second one as a duplicate and rejects a valid batch. Please preserve the compound identity of (TableBucket, originalPartitionName) in the request and response bookkeeping, while retaining duplicate rejection only for identical compound keys.

fluss-rpc/src/main/proto/FlussApi.proto
```diff
@@ -933,6 +933,8 @@ message PbPutKvReqForBucket {
   optional int64 partition_id = 1;
   required int32 bucket_id = 2;
   required bytes records = 3;
+  // The original partition name for historical PK writes. It is unset for normal writes.
+  optional string original_partition_name = 4;
```
wuchong, line 943: Please first bump the maximum PUT_KV API version when introducing original_partition_name. The client-side feature gate can be worked out in a follow-up: a client that negotiated an older PUT_KV version should explicitly reject requests carrying this field instead of sending them. The version bump is the prerequisite that makes this capability detectable; without it, an old protobuf server silently ignores the unknown field and can treat the historical request as a normal write to the shared physical bucket.

fluss-server/src/main/java/org/apache/fluss/server/replica/historical/HistoricalLakeLookupManager.java
```diff
@@ -659,6 +467,84 @@ private static void deleteTableLookupDirIfEmpty(File tableLookupDir) {
         }
     }
 
+    @Nullable
+    byte[] lookupValue(
```
wuchong: Can we reuse `lookup()` method or merge them? They looks very similar.

fluss-server/src/main/java/org/apache/fluss/server/replica/historical/HistoricalPartitionManager.java
```diff
@@ -0,0 +1,326 @@
...
+                        }
+                    },
+                    () ->
+                            new PutKvResultForBucket(
+                                    putData.tableBucket(),
+                                    ApiError.fromThrowable(
+                                            new HistoricalPartitionThrottledException(
+                                                    "Historical write is throttled for "
+                                                            + putData.tableBucket()
+                                                            + "."))));
```
wuchong: nit: add original partition name information in the exception message?

fluss-server/src/main/java/org/apache/fluss/server/replica/historical/HistoricalPartitionManager.java
```diff
@@ -0,0 +1,326 @@
...
+                    () ->
+                            new LookupResultForBucket(
+                                    tableBucket,
+                                    null,
+                                    lookupData.originalPartitionName(),
+                                    ApiError.fromThrowable(
+                                            new HistoricalPartitionThrottledException(
+                                                    "Historical lookup is throttled for "
+                                                            + tableBucket
+                                                            + "."))));
```
wuchong: nit: add original partition name information in the exception message?

fluss-server/src/main/java/org/apache/fluss/server/replica/historical/HistoricalPartitionManager.java
```diff
@@ -0,0 +1,326 @@
...
+                    new LookupResultForBucket(
+                            tableBucket,
+                            null,
+                            lookupData.originalPartitionName(),
+                            ApiError.fromThrowable(e)));
+        }
+    }
+
+    /** Writes records to the local overlay of a historical partition. */
+    public CompletableFuture<PutKvResultForBucket> put(
```
wuchong: nit: rename to `putKv`

fluss-server/src/main/java/org/apache/fluss/server/replica/historical/HistoricalPartitionTaskExecutor.java
```diff
@@ -0,0 +1,209 @@
...
+
+    /** Creates a historical-partition task executor from the server configuration. */
+    public HistoricalPartitionTaskExecutor(Configuration conf) {
+        this(conf, null);
+    }
+
+    /** Creates a historical-partition task executor backed by the supplied executor. */
+    @VisibleForTesting
+    public HistoricalPartitionTaskExecutor(
+            Configuration conf, ExecutorService historicalPartitionExecutor) {
```
wuchong:
```suggestion
            Configuration conf, @Nullable ExecutorService historicalPartitionExecutor) {
```

fluss-server/src/main/java/org/apache/fluss/server/kv/KvTablet.java
```diff
@@ -1124,13 +849,13 @@ private void notifyFlushComplete() {
     /** put key,value,logOffset into pre-write buffer directly. */
     void putToPreWriteBuffer(
```
wuchong, line 899: Only non-historical can invoke this method, should we add check 才能调用，是否需要 check `historicalPartition` flag？

fluss-server/src/main/java/org/apache/fluss/server/kv/KvTablet.java
```diff
@@ -1148,13 +873,9 @@ public Executor getGuardedExecutor() {
         return runnable -> inWriteLock(kvLock, runnable::run);
     }
 
-    // get from kv pre-write buffer first, if can't find, get from rocksdb
+    // Get from the state pre-write buffer first, then fall back to the underlying storage.
     private byte[] getFromBufferOrKv(KvPreWriteBuffer.Key key) throws IOException {
```
wuchong: This method is only invoked by `multiGetFromBufferOrKv` now, can be inlined into `multiGetFromBufferOrKv`.

fluss-server/src/main/java/org/apache/fluss/server/kv/KvTablet.java
wuchong, line 942: Add `checkState(!historicalPartition)`

fluss-server/src/main/java/org/apache/fluss/server/replica/Replica.java
```diff
@@ -272,6 +278,10 @@ public Replica(
                         tableInfo.getSchema());
         this.tableInfo = tableInfo;
         this.tableConfig = tableInfo.getTableConfig();
+        String partitionName = physicalPath.getPartitionName();
+        this.historicalPartition =
+                tableInfo.getPartitionKeys().size() == 1
+                        && HISTORICAL_PARTITION_VALUE.equals(partitionName);
```
wuchong: `HISTORICAL_PARTITION_VALUE.equals(partitionName)` is enough.

fluss-server/src/main/java/org/apache/fluss/server/replica/Replica.java
```diff
@@ -316,7 +326,7 @@ public long logicalStorageLogSize() {
     }
 
     public long logicalStorageKvSize() {
-        if (isLeader() && isKvTable()) {
+        if (isLeader() && isKvTable() && !isHistoricalPartition()) {
```
wuchong: I think we should account for the logical KV size of historical partitions. Since the RocksDB storage for historical partitions can still be substantial, ignoring this part may lead to inaccurate estimates of cluster disk usage.

fluss-server/src/main/java/org/apache/fluss/server/replica/ReplicaManager.java
```diff
@@ -996,9 +1099,7 @@ public void lookups(
     }
 
     private boolean isHistoricalPartitionReplica(Replica replica) {
-        String partitionName = replica.getPhysicalTablePath().getPartitionName();
-        return replica.getTableInfo().getPartitionKeys().size() == 1
-                && HISTORICAL_PARTITION_VALUE.equals(partitionName);
+        return replica.isHistoricalPartition();
```
wuchong: This invocation can be inlined, no need such a method.

fluss-server/src/main/java/org/apache/fluss/server/replica/ReplicaManager.java
```diff
@@ -1858,7 +1959,7 @@ private boolean isUnexpectedException(Exception e) {
                 || e instanceof StorageBackpressureException);
     }
 
-    private boolean isUnexpectedHistoricalLookupException(Exception e) {
+    private boolean isUnexpectedHistoricalPartitionException(Exception e) {
```
wuchong: nit: can be simplified to isUnexpectedHistoricalException

fluss-server/src/main/java/org/apache/fluss/server/kv/KvTablet.java
```diff
@@ -135,36 +119,24 @@ public final class KvTablet {
...
 
     private final LogTablet logTablet;
-    private final ArrowWriterProvider arrowWriterProvider;
-    private final MemorySegmentPool memorySegmentPool;
 
     private final File kvTabletDir;
     private final long writeBatchSize;
     private final RocksDBKv rocksDBKv;
     private final KvPreWriteBuffer kvPreWriteBuffer;
+    private final LocalKvStateAccessor localKvStateAccessor;
```
wuchong: The abstraction of `KvStateAccessor` is excellent as it hides the differences between local and historical access. However, hardcoding `LocalKvStateAccessor` here breaks this abstraction and makes the code fragile. For instance, a historical partition tablet might mistakenly use this local accessor (e.g., via `encodeKey()`), leading to errors. I suggest generalizing this to `KvStateAccessor` and determining whether it is a local or historical accessor in the constructor parameters, thereby avoiding hardcoding within the logic.

Given that we cannot perform lake lookups within `org.apache.fluss.server.kv.KvStateAccessor#lookup` (as it holds the `kvLock`), the only aspect that may need abstraction and generalization is `encodeKey`. Could we consider passing the partition name as a nullable parameter? This would allow us to use a unified `KvStateAccessor` within `KvTablet`.

fluss-server/src/main/java/org/apache/fluss/server/kv/KvTablet.java
wuchong: We should disable `rowCount` for historical partition before we support it. Add a TODO to support it in the future.

fluss-server/src/main/java/org/apache/fluss/server/kv/historical/HistoricalKvStateAccessor.java
```diff
@@ -0,0 +1,111 @@
...
+ * namespaces every physical primary key with {@code originalPartitionName} before delegating to the
+ * local accessor. For merge reads, the local state is authoritative when it contains either a value
+ * or a tombstone; the external fallback is consulted only when the key is absent locally.
+ */
+@Internal
+public final class HistoricalKvStateAccessor implements KvStateAccessor {
+
+    private final KvStateAccessor localAccessor;
+    private final String originalPartitionName;
+    private final @Nullable HistoricalValueLookup fallbackLookup;
```
wuchong: Why it is nullable? I didn't see any `null` value for this parameter.

fluss-server/src/main/java/org/apache/fluss/server/replica/Replica.java
luoyuxia: I left todo in here to left it to another pr to keep this pr samll. But I think we can do it in this pr to make it focus.

fluss-server/src/main/java/org/apache/fluss/server/replica/Replica.java
luoyuxia: I left todo to do clean in follow pr.  The quick way is to drop the kv directly after idle for a while of time / the size larger that a threshold. 

The plan is to integrate cleanup of the local historical overlay with row TTL. After looked into the pr,  IIUC, this can build on #3621 by using the durable-tiering watermark as an additional guard for TTL-based compaction: a historical record should be removed only after it has expired and its corresponding WAL has been durably tiered.

fluss-rpc/src/main/proto/FlussApi.proto
luoyuxia: Okay. I originally intended to introduce PUT_KV v3 in the follow-up client-side PR together with the client feature gate. But I agree that bumping the maximum version in this PR is necessary to make the capability detectable, so I’ll add it here. The client-side rejection for negotiated versions below v3 will follow in the next PR.

luoyuxia:
@wuchong Thanks for your review. Comments has been addressed. Can you please help review again?

fluss-server/src/main/java/org/apache/fluss/server/kv/KvRecoverHelper.java
```diff
@@ -331,13 +353,57 @@ private void initSchema(int schemaId) throws Exception {
...
+                        : null;
         rowEncoder = RowEncoder.create(kvFormat, dataTypes);
         currentFieldGetters = new InternalRow.FieldGetter[currentRowType.getFieldCount()];
         for (int i = 0; i < currentRowType.getFieldCount(); i++) {
             currentFieldGetters[i] = InternalRow.createFieldGetter(currentRowType.getTypeAt(i), i);
         }
     }
 
+    /** Extracts the original partition name from a historical WAL row. */
+    private static final class HistoricalPartitionNameExtractor {
```
wuchong: This duplicates `org.apache.fluss.client.table.getter.PartitionGetter`: both resolve partition fields from a `RowType`, convert their values with `PartitionUtils.convertValueOfType`, and build a `ResolvedPartitionSpec`. Could we move or extract `PartitionGetter` into `fluss-common` under a neutral package and reuse it here? `fluss-server` should not depend on `fluss-client`, but all of `PartitionGetter`'s current dependencies already live in `fluss-common`. Please keep the recovery-specific non-empty partition-key check at this call site, since the existing getter permits an empty key list. If moving the existing public class is a compatibility concern, we can retain a delegating wrapper in its current client package.

fluss-server/src/main/java/org/apache/fluss/server/kv/KvTablet.java
```diff
@@ -514,6 +497,72 @@ public LogAppendInfo putAsLeader(KvRecordBatch kvRecords, @Nullable int[] target
...
+     * method releases the tablet lock.
+     */
+    public List<byte[]> findKeysRequiringLakeLookup(
+            KvRecordBatch kvRecords,
+            @Nullable int[] targetColumns,
+            MergeMode mergeMode,
+            String originalPartitionName)
+            throws Exception {
+        checkState(historicalPartition, "%s is not a historical KV tablet", tableBucket);
+        return inWriteLock(
```
wuchong: Why using write lock here? IIUC, this is read-only operation.

fluss-server/src/main/java/org/apache/fluss/server/kv/KvWriteProcessor.java
```diff
@@ -0,0 +1,556 @@
...
+            return logOffset + 1;
+        } else {
+            walBuilder.append(ChangeType.UPDATE_BEFORE, latestSchemaRow.replaceRow(oldValue.row));
+            walBuilder.append(ChangeType.UPDATE_AFTER, latestSchemaRow.replaceRow(newValue.row));
+            stateAccessor.update(key, newValue.encodeValue(), logOffset + 1);
+            return logOffset + 2;
+        }
+    }
+
+    private byte[] getFromState(
```
wuchong, line 480: Please add javadoc on this method to explain the diffe`rence between `encodedPrimaryKey`, `primaryKey`. I think the `encodedPrimaryKey` is not accurate now, better to rename it.

wuchong (approved):

luoyuxia:
Let me merge this since I still have one pull request based on it.

PR #4006, opened 2026-08-15
[client] Expose Arrow IPC batches with optional Java adapter

### LINKED ISSUE #4005: [client] Expose Arrow changelog batches for primary-key tables
### Motivation

Fluss already stores ARROW primary-key log batches as Arrow IPC rows plus a per-row change-type vector. The Java client currently exposes row-oriented polling publicly, while record-batch polling is internal and rejects primary-key tables.

StreamFusion needs a Flink source path that can retain Arrow buffers from the Fluss Java client instead of first materializing every live-log record as RowData. Other vectorized consumers have the same need.

### Proposed scope

- Expose record-batch polling on the public `LogScanner` API for ARROW log tables.
- Preserve the stored per-row `ChangeType` values for primary-key changelogs, including `UPDATE_BEFORE`, `UPDATE_AFTER`, and `DELETE`.
- Keep the change-type vector aligned when batches are sliced or truncated at subscribed offsets.
- Preserve explicit Arrow ownership and close semantics.
- Keep existing scanner implementations compatible by providing a default unsupported implementation.

Snapshot reads can continue through the normal row conversion path initially; this issue targets the live log path where Fluss already has the desired Arrow representation.

A caller-provided Arrow allocator would also be useful for consumers such as StreamFusion, but allocator ownership is a broader API/lifecycle decision and can be handled in a follow-up.

jordepic:
## Summary

Expose Arrow log polling for both primary-key and append-only tables using the ARROW log format. Primary-key batches include their stored per-row changelog types. Non-empty top-level projections use the existing tablet-server pruning path for local reads.

The core client returns `ArrowIpcBatch`: owned, immutable heap-backed IPC schema and record-batch messages plus offsets, timestamps, change types, and the metadata needed for column mapping and partial-batch reads. These batches remain valid after subsequent polls or scanner closure, and require no unshaded Arrow Java dependency or explicit close.

The optional `fluss-client-arrow` artifact provides `ArrowBatchReader.read(batch, allocator)` and closeable `ArrowBatch` results exposing `VectorSchemaRoot`. It brings Arrow Java 15.0.0 dependencies and applies schema evolution, column ordering, and row slicing. The caller owns the allocator; closing the scanner cannot invalidate decoded batches. Existing Paimon tiering consumes IPC batches through this adapter, which is included alongside the Paimon plugin in the distribution.

Empty log batches are skipped before reading an IPC payload and still advance consumed offsets. Arrow poll results distinguish row data from offset-only progress. The first row or batch poll fixes a scanner's polling mode.

Document the dependency, supported version, allocator ownership, and the distinction between physical IPC data and logical scan results. This covers log scanning, not primary-key snapshot scans; it does not change the stored log format.

## Test Plan

- Full `verify` for `fluss-common`, `fluss-client`, and `fluss-client-arrow`: 2,505 tests, zero failures, two existing skips.
- All 26 cases in `TieringSplitReaderTest` and `PaimonTieringTest` passed.
- Clean install of the adapter and Paimon modules with their reactor dependencies, followed by 21 focused regression tests.
- Repository-wide Spotless check and affected-module Checkstyle passed; new API and adapter sources also compile with Java 8's API surface (`--release 8`).
- Standalone consumers verified both the core client without unshaded Arrow and the adapter's transitive Maven dependencies.
- Standard Arrow Java IPC reader interoperability covered NONE, LZ4, and Zstd compression.

### DIFF
--- a/fluss-client/src/main/java/org/apache/fluss/client/table/scanner/log/LogScanner.java
+++ b/fluss-client/src/main/java/org/apache/fluss/client/table/scanner/log/LogScanner.java
@@ -57,6 +57,22 @@ public interface LogScanner extends AutoCloseable {
     ScanRecords poll(Duration timeout);
 
     /**
+     * Polls Arrow record batches from the tablet server without materializing individual rows.
+     *
+     * <p>This method is supported for tables whose log format is {@code ARROW}. Each returned batch
+     * owns its Arrow memory and, for primary-key tables, carries the stored per-row changelog
+     * types. Callers must close the returned {@link ArrowScanRecords} after processing it.
+     *
+     * @param timeout the timeout to poll
+     * @return the Arrow batches returned by this poll
+     * @throws UnsupportedOperationException if the table does not use the {@code ARROW} log format
+     * @throws IllegalStateException if the scanner is not subscribed to any buckets
+     */
+    default ArrowScanRecords pollRecordBatch(Duration timeout) {
+        throw new UnsupportedOperationException("Arrow record batch polling is not supported");
+    }
+
+    /**
      * Subscribe to the given table bucket in given offset dynamically. If the table bucket is
      * already subscribed, the offset will be updated.
      *
--- a/fluss-client/src/main/java/org/apache/fluss/client/table/scanner/log/LogScanner.java
+++ b/fluss-client/src/main/java/org/apache/fluss/client/table/scanner/log/LogScanner.java
@@ -66,10 +66,12 @@ public interface LogScanner extends AutoCloseable {
      * @param timeout the timeout to poll
      * @return the Arrow batches returned by this poll
      * @throws UnsupportedOperationException if the table does not use the {@code ARROW} log format
+     *     or the scanner implementation does not support Arrow record batch polling
      * @throws IllegalStateException if the scanner is not subscribed to any buckets
      */
     default ArrowScanRecords pollRecordBatch(Duration timeout) {
-        throw new UnsupportedOperationException("Arrow record batch polling is not supported");
+        throw new UnsupportedOperationException(
+                "This LogScanner implementation does not support Arrow record batch polling");
     }
 
     /**
--- a/fluss-client/src/test/java/org/apache/fluss/client/table/scanner/log/LogScannerITCase.java
+++ b/fluss-client/src/test/java/org/apache/fluss/client/table/scanner/log/LogScannerITCase.java
@@ -684,15 +684,15 @@ public class LogScannerITCase extends ClientToServerITCaseBase {
                         IntVector keys = (IntVector) b.getVectorSchemaRoot().getVector(0);
                         VarCharVector values = (VarCharVector) b.getVectorSchemaRoot().getVector(1);
                         for (int rowId = 0; rowId < b.getRecordCount(); rowId++) {
-                            int offset = (int) b.getBaseLogOffset() + rowId;
+                            long offset = b.getBaseLogOffset() + rowId;
                             assertThat(offset).isEqualTo(count);
                             assertThat(b.getChangeType(rowId))
-                                    .isEqualTo(expectedChangeTypes[offset]);
+                                    .isEqualTo(expectedChangeTypes[count]);
                             assertThat(changeTypes.get(rowId))
-                                    .isEqualTo(expectedChangeTypes[offset].toByteValue());
-                            assertThat(keys.get(rowId)).isEqualTo(expectedKeys[offset]);
+                                    .isEqualTo(expectedChangeTypes[count].toByteValue());
+                            assertThat(keys.get(rowId)).isEqualTo(expectedKeys[count]);
                             assertThat(values.getObject(rowId).toString())
-                                    .isEqualTo(expectedValues[offset]);
+                                    .isEqualTo(expectedValues[count]);
                             count++;
                         }
                     }
@@ -718,12 +718,12 @@ public class LogScannerITCase extends ClientToServerITCaseBase {
                         assertThat(b.getVectorSchemaRoot().getFieldVectors()).hasSize(1);
                         VarCharVector values = (VarCharVector) b.getVectorSchemaRoot().getVector(0);
                         for (int rowId = 0; rowId < b.getRecordCount(); rowId++) {
-                            int offset = (int) b.getBaseLogOffset() + rowId;
+                            long offset = b.getBaseLogOffset() + rowId;
                             assertThat(offset).isEqualTo(count);
                             assertThat(b.getChangeType(rowId))
-                                    .isEqualTo(expectedChangeTypes[offset]);
+                                    .isEqualTo(expectedChangeTypes[count]);
                             assertThat(values.getObject(rowId).toString())
-                                    .isEqualTo(expectedValues[offset]);
+                                    .isEqualTo(expectedValues[count]);
                             count++;
                         }
                     }
--- a/fluss-client/src/test/java/org/apache/fluss/client/table/scanner/log/LogScannerITCase.java
+++ b/fluss-client/src/test/java/org/apache/fluss/client/table/scanner/log/LogScannerITCase.java
@@ -43,6 +43,7 @@ import org.junit.jupiter.api.Test;
 import org.junit.jupiter.params.ParameterizedTest;
 import org.junit.jupiter.params.provider.ValueSource;
 
+import java.nio.ByteBuffer;
 import java.time.Duration;
 import java.util.ArrayList;
 import java.util.Collections;
@@ -507,6 +508,65 @@ public class LogScannerITCase extends ClientToServerITCaseBase {
     }
 
     @Test
+    void testPollArrowBatchesWithPrimaryKeyChangelog() throws Exception {
+        TablePath tablePath = TablePath.of("test_db_1", "test_arrow_batches_with_changelog");
+        TableDescriptor tableDescriptor =
+                TableDescriptor.builder()
+                        .schema(
+                                Schema.newBuilder()
+                                        .column("a", DataTypes.INT())
+                                        .column("b", DataTypes.STRING())
+                                        .primaryKey("a")
+                                        .build())
+                        .distributedBy(1)
+                        .logFormat(LogFormat.ARROW)
+                        .build();
+        createTable(tablePath, tableDescriptor, false);
+
+        try (Table table = conn.getTable(tablePath)) {
+            UpsertWriter writer = table.newUpsert().createWriter();
+            writer.upsert(row(1, "old")).get();
+            writer.flush();
+            writer.upsert(row(1, "new")).get();
+            writer.flush();
+            writer.upsert(row(2, "deleted")).get();
+            writer.flush();
+            writer.delete(row(2, "deleted")).get();
+            writer.flush();
+
+            ChangeType[] expectedChangeTypes = {
+                ChangeType.INSERT,
+                ChangeType.UPDATE_BEFORE,
+                ChangeType.UPDATE_AFTER,
+                ChangeType.INSERT,
+                ChangeType.DELETE
+            };
+            int[] expectedKeys = {1, 1, 1, 2, 2};
+            String[] expectedValues = {"old", "old", "new", "deleted", "deleted"};
+
+            try (LogScanner scanner = table.newScan().createLogScanner()) {
+                scanner.subscribeFromBeginning(0);
+                pollAndVerifyChangelogArrowBatches(
+                        scanner, expectedChangeTypes, expectedKeys, expectedValues, 0);
+            }
+
+            try (LogScanner scanner = table.newScan().project(new int[] {1}).createLogScanner()) {
+                scanner.subscribeFromBeginning(0);
+                pollAndVerifyProjectedChangelogArrowBatches(
+                        scanner, expectedChangeTypes, expectedValues);
+            }
+
+            // Offset 2 starts in the middle of the update batch and verifies that slicing keeps
+            // the change-type vector aligned with the Arrow rows.
+            try (LogScanner scanner = table.newScan().createLogScanner()) {
+                scanner.subscribe(0, 2L);
+                pollAndVerifyChangelogArrowBatches(
+                        scanner, expectedChangeTypes, expectedKeys, expectedValues, 2);
+            }
+        }
+    }
+
+    @Test
     void testPollArrowBatchesWithSchemaEvolution() throws Exception {
         TablePath tablePath = TablePath.of("test_db_1", "test_arrow_batches_with_schema_evolution");
         TableDescriptor tableDescriptor =
@@ -548,7 +608,7 @@ public class LogScannerITCase extends ClientToServerITCaseBase {
 
             int totalRecords = 6;
             // subscribe from beginning and verify all 6 records
-            try (LogScannerImpl scanner = (LogScannerImpl) table.newScan().createLogScanner()) {
+            try (LogScanner scanner = table.newScan().createLogScanner()) {
                 scanner.subscribeFromBeginning(0);
                 pollAndVerifyArrowBatches(scanner, totalRecords, 0);
             }
@@ -556,7 +616,7 @@ public class LogScannerITCase extends ClientToServerITCaseBase {
             // subscribe from the middle of the first batch (offset 1)
             // to ensure records before the subscribe offset are not returned
             int subscribeOffset = 1;
-            try (LogScannerImpl scanner2 = (LogScannerImpl) table.newScan().createLogScanner()) {
+            try (LogScanner scanner2 = table.newScan().createLogScanner()) {
                 scanner2.subscribe(0, subscribeOffset);
                 pollAndVerifyArrowBatches(
                         scanner2, totalRecords - subscribeOffset, subscribeOffset);
@@ -565,7 +625,7 @@ public class LogScannerITCase extends ClientToServerITCaseBase {
     }
 
     private void pollAndVerifyArrowBatches(
-            LogScannerImpl scanner, int expectedRecords, int minExpectedOffset) {
+            LogScanner scanner, int expectedRecords, int minExpectedOffset) {
         int count = 0;
         long deadline = System.nanoTime() + Duration.ofSeconds(30).toNanos();
         while (count < expectedRecords) {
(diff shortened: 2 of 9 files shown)


woodsbrandon044-hash (approved):

wuchong:
cc @luoyuxia, do you think we need anything else if we want to expose `ArrowBatch` as a public API in `LogScanner`?

woodsbrandon044-hash:
> cc @luoyuxia, do you think we need anything else if we want to expose `ArrowBatch` as a public API in `LogScanner`?

Who you talking about

fluss-common/src/main/java/org/apache/fluss/record/ArrowBatchData.java
```diff
@@ -128,8 +193,11 @@ public ArrowBatchData truncateAndTransferOwnership(int rowCount) {
                 rowCount,
                 getRecordCount());
         VectorSchemaRoot slicedRoot = vectorSchemaRoot.slice(0, rowCount);
+        byte[] slicedChangeTypes =
+                changeTypes == null ? null : Arrays.copyOfRange(changeTypes, 0, rowCount);
         close();
-        return new ArrowBatchData(slicedRoot, baseLogOffset, timestamp, schemaId);
+        return new ArrowBatchData(
+                slicedRoot, baseLogOffset, timestamp, schemaId, slicedChangeTypes);
```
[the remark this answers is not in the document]
woodsbrandon044-hash: Hope your blessed today

fluss-client/src/test/java/org/apache/fluss/client/table/scanner/log/LogScannerITCase.java
```diff
@@ -506,6 +507,65 @@ void testSubscribeOutOfRangeLog() throws Exception {
...
+        try (Table table = conn.getTable(tablePath)) {
+            UpsertWriter writer = table.newUpsert().createWriter();
+            writer.upsert(row(1, "old")).get();
+            writer.flush();
+            writer.upsert(row(1, "new")).get();
+            writer.flush();
+            writer.upsert(row(2, "deleted")).get();
+            writer.flush();
+            writer.delete(row(2, "deleted")).get();
+            writer.flush();
```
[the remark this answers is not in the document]
woodsbrandon044-hash: > `UpsertWriter` is created but never closed. If it holds network resources/buffers, this can leak across tests and cause flaky behavior. Wrap it in try-with-resources (or ensure a `close()` in a `finally`) to release resources deterministically.

Information

fluss-common/src/main/java/org/apache/fluss/record/ArrowBatchData.java
```diff
@@ -105,9 +165,14 @@ skipRows < getRecordCount(),
...
         VectorSchemaRoot slicedRoot = vectorSchemaRoot.slice(skipRows, remainingRows);
+        byte[] slicedChangeTypes =
+                changeTypes == null
+                        ? null
+                        : Arrays.copyOfRange(changeTypes, skipRows, changeTypes.length);
         // release original vector buffers; sliced vectors hold independent copies
         close();
-        return new ArrowBatchData(slicedRoot, baseLogOffset + skipRows, timestamp, schemaId);
+        return new ArrowBatchData(
+                slicedRoot, baseLogOffset + skipRows, timestamp, schemaId, slicedChangeTypes);
```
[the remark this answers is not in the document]
jordepic: Addressed in a7e9371ac. Fluss uses Arrow 15, where VectorSchemaRoot.slice calls TransferPair.splitAndTransfer for each vector, so the sliced root owns valid vector buffers after the original root is closed. I clarified the ownership comment to describe the transfer instead of calling it a copy.

fluss-client/src/main/java/org/apache/fluss/client/table/scanner/log/LogScanner.java
```diff
@@ -56,6 +56,22 @@ public interface LogScanner extends AutoCloseable {
...
+     * owns its Arrow memory and, for primary-key tables, carries the stored per-row changelog
+     * types. Callers must close the returned {@link ArrowScanRecords} after processing it.
+     *
+     * @param timeout the timeout to poll
+     * @return the Arrow batches returned by this poll
+     * @throws UnsupportedOperationException if the table does not use the {@code ARROW} log format
+     * @throws IllegalStateException if the scanner is not subscribed to any buckets
+     */
+    default ArrowScanRecords pollRecordBatch(Duration timeout) {
+        throw new UnsupportedOperationException("Arrow record batch polling is not supported");
```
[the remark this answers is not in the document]
jordepic: Resolved in a7e9371ac. The default message now says that the specific LogScanner implementation does not support Arrow polling, and the Javadoc distinguishes that case from a non-ARROW table.

fluss-client/src/test/java/org/apache/fluss/client/table/scanner/log/LogScannerITCase.java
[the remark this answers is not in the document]
jordepic: No code change is applicable here: UpsertWriter extends TableWriter, which is not AutoCloseable and exposes no close method. Its WriterClient is shared and owned by FlussConnection, which closes it with the connection; the test already flushes deterministically.

fluss-client/src/test/java/org/apache/fluss/client/table/scanner/log/LogScannerITCase.java
```diff
@@ -601,6 +661,78 @@ private void pollAndVerifyArrowBatches(
...
+                        for (int rowId = 0; rowId < b.getRecordCount(); rowId++) {
+                            int offset = (int) b.getBaseLogOffset() + rowId;
+                            assertThat(offset).isEqualTo(count);
+                            assertThat(b.getChangeType(rowId))
+                                    .isEqualTo(expectedChangeTypes[offset]);
+                            assertThat(changeTypes.get(rowId))
+                                    .isEqualTo(expectedChangeTypes[offset].toByteValue());
+                            assertThat(keys.get(rowId)).isEqualTo(expectedKeys[offset]);
+                            assertThat(values.getObject(rowId).toString())
+                                    .isEqualTo(expectedValues[offset]);
```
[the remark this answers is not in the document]
jordepic: Resolved in a7e9371ac. The test now keeps the computed log offset as a long and uses the already bounded int count for array indexing. I made the same cleanup in the projected changelog helper.

fluss-common/src/main/java/org/apache/fluss/record/ArrowBatchData.java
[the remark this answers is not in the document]
jordepic: Addressed in a7e9371ac. Fluss uses Arrow 15, where VectorSchemaRoot.slice calls TransferPair.splitAndTransfer for each vector, so the truncated root remains valid after closing the original. The ownership wording is now explicit.

fluss-common/src/main/java/org/apache/fluss/record/DefaultLogRecordBatch.java
```diff
@@ -277,17 +277,23 @@ public ArrowBatchData loadArrowBatch(ReadContext context) {
...
             int recordsDataOffset = recordsDataOffset();
             int arrowOffset = position + recordsDataOffset;
-            int arrowLength = sizeInBytes() - recordsDataOffset;
+            byte[] changeTypes = null;
+            if (!isAppendOnly()) {
+                changeTypes = new byte[getRecordCount()];
+                segment.get(arrowOffset, changeTypes);
+                arrowOffset += changeTypes.length;
+            }
+            int arrowLength = position + sizeInBytes() - arrowOffset;
```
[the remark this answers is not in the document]
jordepic: Resolved in a7e9371ac. arrowLength is now computed entirely from batch-relative values: sizeInBytes, recordsDataOffset, and changeTypesLength. Only arrowOffset adds the absolute segment position.

jordepic:
Resolved the Copilot review comments in a7e9371ac:\n\n- Clarified Arrow 15 slice ownership: VectorSchemaRoot.slice already uses per-vector splitAndTransfer, so the sliced root remains valid after closing the original.\n- Reworked Arrow payload length calculation to use batch-relative values.\n- Clarified the default unsupported LogScanner implementation error and Javadoc.\n- Kept changelog test offsets as long values without narrowing.\n- Confirmed UpsertWriter is not AutoCloseable; its shared WriterClient is owned and closed by FlussConnection.\n\nValidation:\n- FileLogInputStreamTest: 9 passed\n- LogScannerITCase: 13 passed\n- Spotless, Checkstyle, RAT, and affected reactor compilation passed\n\nAll six review threads have been replied to and resolved.

jordepic:
@loserwang1024 and @luoyuxia would you mind taking a look here? Thank you!

polyzos:
I pushed a small commit to fix a Javadoc. LGTM 👍
I just want your thoughts on changing the API to public evolving before merging @luoyuxia

polyzos (approved):

luoyuxia:
@polyzos Thanks for take me into consideration. Will have a look in the few days

fluss-common/src/main/java/org/apache/fluss/record/LogRecordReadContext.java
```diff
@@ -453,7 +453,7 @@ public ArrowBatchData createArrowBatchData(
                 readRoot = null;
             }
             ArrowBatchData arrowBatchData =
-                    new ArrowBatchData(outputRoot, baseLogOffset, timestamp, schemaId);
+                    new ArrowBatchData(outputRoot, baseLogOffset, timestamp, schemaId, changeTypes);
```
luoyuxia: With the projection restriction removed from `pollRecordBatch()`, this path only applies the schema-evolution mapping and never applies `target.selectedFields`. This leaves two projection cases incorrect:

- **Local fetch:** For a table `[a, b]`, `project(new int[] {1, 0})` should return `[b, a]`. The fetch request uses `getProjectionInOrder()`, so the server returns `[a, b]`; the Arrow result keeps that order instead of applying the reordering indices.
- **Remote log fetch:** For `project(new int[] {1})`, `remoteReadContext` disables server-side projection and retains the full row type. Since the Arrow path does not apply the client-side selection either, it returns `[a, b]` instead of `[b]`.

The row-based path applies these indices through `getSelectedFieldGetters()`, so `poll()` and `pollRecordBatch()` can return different column selections/orders for the same projected scan. With same-typed columns, the local case can silently swap values; the remote case can change the output schema when reading moves between remote and local logs.

fluss-client/src/main/java/org/apache/fluss/client/table/scanner/log/LogScanner.java
```diff
@@ -56,6 +56,24 @@ public interface LogScanner extends AutoCloseable {
...
+     * owns its Arrow memory and, for primary-key tables, carries the stored per-row changelog
+     * types. Callers must close the returned {@link ArrowScanRecords} after processing it.
+     *
+     * @param timeout the timeout to poll
+     * @return the Arrow batches returned by this poll
+     * @throws UnsupportedOperationException if the table does not use the {@code ARROW} log format
+     *     or the scanner implementation does not support Arrow record batch polling
+     * @throws IllegalStateException if the scanner is not subscribed to any buckets
+     */
+    default ArrowScanRecords pollRecordBatch(Duration timeout) {
```
luoyuxia, line 88: Having both `poll()` and `pollRecordBatch()` on the same scanner feels a bit easy to misuse. For example, if `poll()` stops partway through a batch, switching to `pollRecordBatch()` currently skips the unread rows in that batch.

Would it be worth following `TypedLogScanner` and introducing a separate `ArrowLogScanner`, so the polling mode is fixed when the scanner is created? TBH, I'm not sure which API design is better, but it seems worth considering before making this public.

fluss-common/src/main/java/org/apache/fluss/record/LogRecordReadContext.java
jordepic: Fixed in [be0ff9e34](https://github.com/jordepic/fluss/commit/be0ff9e34ac71cf49a323744680e9a53e2a6dc48). The Arrow output mapping now composes `target.selectedFields` with the schema-evolution mapping. Local fetches restore the requested column order after server-side pruning; remote fetches apply the requested selection and order in the client, including nulls for columns absent from older schemas.

Added regression coverage for reordered same-typed columns on both primary-key and append-only tables, plus remote subsets/reordering with schema evolution and mid-batch offsets. The focused Arrow scanner and remote-fetch tests pass on Java 11.

Validation: 86 targeted tests, Spotless, and Checkstyle passed. A broader dependency `verify` run hit two storage-metric assertions in `ReplicaTest`; those two tests pass on the untouched PR head, and all 15 `ReplicaTest` cases passed on rerun with this change. The full dependency verification did not complete.

fluss-client/src/main/java/org/apache/fluss/client/table/scanner/log/LogScanner.java
jordepic: The unread-row loss is fixed in [be0ff9e34](https://github.com/jordepic/fluss/commit/be0ff9e34ac71cf49a323744680e9a53e2a6dc48). I kept the current API for this PR and made the polling modes mutually exclusive per scanner: the first poll selects the mode, including an empty poll, and switching between `poll()` and `pollRecordBatch()` throws `IllegalStateException` before consuming buffered records. Both methods now document that changing modes requires a new scanner.

The regression tests cover both switching directions, empty first polls, and continuing row polling after a rejected switch partway through a batch. They pass on Java 11. This addresses the data-loss case while leaving a separate `ArrowLogScanner` as an API design option.

The broader verification results and rerun details are in the [projection reply](https://github.com/apache/fluss/pull/4006#discussion_r4028158314).

fluss-client/src/main/java/org/apache/fluss/client/table/scanner/log/LogScanner.java
```diff
@@ -49,13 +49,44 @@ public interface LogScanner extends AutoCloseable {
...
+     * order. Columns are pruned on the tablet server for local log reads and in the client for
+     * remote log reads.
+     *
+     * <p>The first poll fixes the polling mode for this scanner, even if it returns no data. Calls
+     * to this method cannot be mixed with {@link #poll(Duration)}. Create a new scanner to change
+     * the polling mode.
+     *
+     * @param timeout the timeout to poll
+     * @return the Arrow batches returned by this poll
+     * @throws UnsupportedOperationException if the table does not use the {@code ARROW} log format
```
fresh-borzoni, line 83: A project that depends on fluss-client alone can't compile batch.getVectorSchemaRoot() — the jar ships shaded Arrow only, nothing under org.apache.arrow. As far as I can tell the only working caller today is tiering to Paimon, which gets Arrow from paimon-arrow and probes for it with Class.forName first.

How is a user meant to get these classes? If they have to add arrow-vector and arrow-memory-netty themselves, should that be in the docs with a version, or would a separate fluss-client-arrow artifact that depends on them for real be cleaner?

fluss-common/src/main/java/org/apache/fluss/record/DefaultLogRecordBatch.java
```diff
@@ -277,17 +277,24 @@ public ArrowBatchData loadArrowBatch(ReadContext context) {
...
+            boolean appendOnly = isAppendOnly();
+            int changeTypesLength = appendOnly ? 0 : getRecordCount();
+            int arrowOffset = position + recordsDataOffset + changeTypesLength;
+            byte[] changeTypes = null;
+            if (!appendOnly) {
+                changeTypes = new byte[changeTypesLength];
+                segment.get(position + recordsDataOffset, changeTypes);
+            }
+            int arrowLength = sizeInBytes() - recordsDataOffset - changeTypesLength;
             batchAccess.loadArrowBatch(segment, arrowOffset, arrowLength);
```
fresh-borzoni: Deleting a key that isn't there writes a header with no Arrow payload, so this line asks Arrow to decode 0 bytes and it throws "Unexpected end of input when reading a RecordBatch". records() skips those batches with an early return on getRecordCount() == 0.

Should fetchArrowBatches() do its count check before loadArrowBatch() rather than after, and advance nextFetchOffset there?

fluss-client/src/main/java/org/apache/fluss/client/table/scanner/log/LogScanner.java
```diff
@@ -49,13 +49,44 @@ public interface LogScanner extends AutoCloseable {
...
+     *     read from, or Arrow record batch polling has already been selected.
      */
     ScanRecords poll(Duration timeout);
 
+    /**
+     * Polls Arrow record batches from the tablet server without materializing individual rows.
+     *
+     * <p>This method is supported for tables whose log format is {@code ARROW}. Each returned batch
+     * owns its Arrow memory and, for primary-key tables, carries the stored per-row changelog
+     * types. Callers must close the returned {@link ArrowScanRecords} after processing it.
```
fresh-borzoni: The "close batches before the scanner" rule is clear and the error message says exactly that. The awkward part is what happens next: Arrow marks the allocator closed before throwing, so closing the batch you were just told to close throws Attempting operation on allocator when allocator is closed and the release never runs.

 Could close() leave the allocator alone while batches are still outstanding, so the caller can act on the error?

fresh-borzoni (commented):
@jordepic Thank you for the PR, left a couple of questions/comments, PTAL

fluss-client/src/main/java/org/apache/fluss/client/table/scanner/log/LogScanner.java
luoyuxia: cc @loserwang1024 Please help review the public api change

fluss-client/src/main/java/org/apache/fluss/client/table/scanner/log/LogScanner.java
jordepic: Updated in [a0fa582d5](https://github.com/apache/fluss/commit/a0fa582d53b2c00511377372bb3260ad01c31f27).

Reworked this boundary: `fluss-client` now returns `ArrowIpcBatch`, containing owned IPC schema and record-batch bytes plus the Fluss metadata needed for changelog types, column mapping and row slicing. Both `fluss-common` and `fluss-client` no longer depend on unshaded Arrow Java.

The new optional `fluss-client-arrow` artifact declares Arrow Java 15.0.0 dependencies and exposes `ArrowBatchReader.read(batch, allocator)` for callers who want `VectorSchemaRoot`. It applies the projection/evolution/slicing metadata and leaves allocator ownership with the caller. The Java API documentation now shows the dependency and usage. Existing Paimon tiering is migrated to the adapter.

Added interoperability coverage using the standard Arrow Java stream reader for uncompressed, LZ4, and Zstd batches. Primary-key and append-only projection tests remain covered in the adapter integration suite.

fluss-common/src/main/java/org/apache/fluss/record/DefaultLogRecordBatch.java
jordepic: Updated in [a0fa582d5](https://github.com/apache/fluss/commit/a0fa582d53b2c00511377372bb3260ad01c31f27).

Fixed: the record-count check now runs before reading the IPC payload, and empty log batches advance `nextFetchOffset`. Arrow polling also returns offset-only progress without treating it as row data.

Added a regression test that deletes a nonexistent primary key, verifies that polling returns no rows while advancing the consumed offset, then writes an insert and verifies scanning continues at the correct offset.

fluss-client/src/main/java/org/apache/fluss/client/table/scanner/log/LogScanner.java
jordepic: Updated in [a0fa582d5](https://github.com/apache/fluss/commit/a0fa582d53b2c00511377372bb3260ad01c31f27).

Fixed by removing the scanner-owned unshaded allocator. Core polling now returns immutable heap-backed IPC batches that require no close and remain valid after scanner closure. The optional Arrow Java adapter decodes with a caller-supplied allocator; closing an `ArrowBatch` closes only its vectors.

Added a regression test that closes the scanner with decoded batches still outstanding, reads those vectors successfully, decodes the retained IPC again after scanner closure, and verifies that closing batches releases their allocations while leaving the caller's allocator usable.

jordepic:
@fresh-borzoni Addressed your feedback in [a0fa582d5](https://github.com/apache/fluss/commit/a0fa582d53b2c00511377372bb3260ad01c31f27):

- Core polling returns owned Arrow IPC bytes, with no unshaded Arrow dependency.
- Optional `fluss-client-arrow` decodes into `VectorSchemaRoot` using a caller-owned allocator; batches survive scanner closure.
- Empty delete batches skip decoding and still advance offsets.

Added documentation and regression/interoperability tests. Both primary-key and append-only Arrow logs use the existing `FetchLog` RPC, retaining local server-side projection and remote client-side projection.

PR #4007, opened 2026-08-16
[flink] Fix batch log scan with empty buckets

### LINKED ISSUE #3972: [flink] Batch Log Table scan fails when some buckets are empty
### Search before asking

### Fluss version

main (development)

##### What happened

When a Flink Batch job reads a multi-bucket Fluss Log Table, the job may fail if some buckets are empty and the source reader runs with low parallelism.

The failure is:

```text
java.lang.IllegalStateException:
Have records for a split that was not registered
```

The same query can complete successfully when:

- `bucket.num = 1`;
- all buckets contain data;
- Flink parallelism is increased.

##### What I expected

The batch scan should return all records from non-empty buckets and finish normally, regardless of whether some buckets are empty.

##### Steps to reproduce

```sql
SET 'execution.runtime-mode' = 'batch';
SET 'parallelism.default' = '1';

CREATE TABLE repro_log (
    id BIGINT,
    payload STRING
) WITH (
    'bucket.num' = '2'
);

INSERT INTO repro_log VALUES
    (1, 'alpha'),
    (2, 'beta');

SELECT *
FROM repro_log;
```

In the reproducing environment, the inserted records are assigned to one bucket while the other bucket remains empty.

##### Actual behavior

The batch job fails with:

```text
java.lang.IllegalStateException:
Have records for a split that was not registered
```

#### Suspected cause

`FlinkSourceSplitReader` multiplexes multiple Fluss buckets through one Fluss `LogScanner`.

When one bounded split reaches its stopping offset, the reader reports the split through `finishedSplits()`. Flink then removes the split from its own source-reader state. However, the corresponding bucket may still remain subscribed in the Fluss scanner, and prefetched results may still be returned with the completed split ID.

This can cause Flink to receive records for a split that has already been unregistered.

The issue is related to #3872, where scanner progress can advance to the stopping offset without producing materialized records.

#### Suggested direction

When a bounded log split is completed:

1. determine completion using `consumedUpToOffset >= stoppingOffset`;
2. remove the bucket from the connector-side split mapping;
3. unsubscribe the bucket from the Fluss scanner;
4. discard pending and prefetched results for that bucket;
5. only then expose the split as finished to Flink.

Gezi-lzq:
### Purpose

Linked issue: close #3972

When a Flink batch job scans a multi-bucket Fluss Log Table with low parallelism, an empty bucket can remain active after its bounded split has finished. Subsequent fetch results may then contain a split ID that Flink has already unregistered.

This change ensures that bounded log splits are completed based on scanner progress and that finished buckets are retired before the next fetch.

### Brief change log

- Complete bounded log splits when `consumedUpToOffset` reaches the stopping offset.
- Handle empty bounded splits starting from `EARLIEST_OFFSET`.
- Remove finished buckets from the Flink source reader mapping and unsubscribe them from the log scanner.
- Add regression coverage for empty buckets and empty bounded splits.

### Tests

- `FlinkSourceSplitReaderTest`
- `Flink120TableSourceBatchITCase`
- `./mvnw -pl fluss-flink/fluss-flink-common -DskipTests verify`

### API and Format

### Documentation

No documentation changes.

### Generative AI Disclosure

- [x] Yes — OpenAI Codex

### DIFF
--- a/fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/source/reader/FlinkSourceSplitReader.java
+++ b/fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/source/reader/FlinkSourceSplitReader.java
@@ -178,9 +178,7 @@ public class FlinkSourceSplitReader implements SplitReader<RecordAndPos, SourceS
                     return FlinkRecordsWithSplitIds.emptyRecords();
                 }
                 ScanRecords scanRecords = logScanner.poll(POLL_TIMEOUT);
-                FlinkRecordsWithSplitIds records = forLogRecords(scanRecords);
-                removeFinishedSplits(records.finishedSplits());
-                return records;
+                return forLogRecords(scanRecords);
             }
         }
     }
@@ -518,6 +516,8 @@ public class FlinkSourceSplitReader implements SplitReader<RecordAndPos, SourceS
                 new FlinkRecordsWithSplitIds(
                         splitRecords, splitIterator, tableScanBuckets.iterator(), finishedSplits);
         stoppingOffsets.forEach(recordsWithSplitIds::setTableBucketStoppingOffset);
+        // Build the current result before retiring finished buckets.
+        removeFinishedSplits(finishedSplits);
         return recordsWithSplitIds;
     }
 
--- a/fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/source/reader/FlinkSourceSplitReader.java
+++ b/fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/source/reader/FlinkSourceSplitReader.java
@@ -468,6 +468,11 @@ public class FlinkSourceSplitReader implements SplitReader<RecordAndPos, SourceS
             splitIdByTableBucket.put(scanBucket, splitId);
             tableScanBuckets.add(scanBucket);
             List<ScanRecord> bucketScanRecords = scanRecords.records(scanBucket);
+            Long consumedUpToOffset = scanRecords.consumedUpToOffset(scanBucket);
+            if (consumedUpToOffset != null && consumedUpToOffset >= stoppingOffset) {
+                stoppingOffsets.put(scanBucket, stoppingOffset);
+                finishedSplits.add(splitId);
+            }
             if (!bucketScanRecords.isEmpty()) {
                 final ScanRecord lastRecord = bucketScanRecords.get(bucketScanRecords.size() - 1);
                 // We keep the maximum message timestamp in the fetch for calculating lags
--- a/fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/source/reader/FlinkSourceSplitReader.java
+++ b/fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/source/reader/FlinkSourceSplitReader.java
@@ -178,7 +178,9 @@ public class FlinkSourceSplitReader implements SplitReader<RecordAndPos, SourceS
                     return FlinkRecordsWithSplitIds.emptyRecords();
                 }
                 ScanRecords scanRecords = logScanner.poll(POLL_TIMEOUT);
-                return forLogRecords(scanRecords);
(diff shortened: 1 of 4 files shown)


Gezi-lzq:
Hi @loserwang1024 and @wuchong, when you have time, could you please help review this PR?

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/source/reader/FlinkSourceSplitReader.java
```diff
@@ -258,7 +260,7 @@ private void subscribeLog(SourceSplitBase split, long startingOffset) {
             Optional<Long> stoppingOffsetOpt = logSplit.getStoppingOffset();
             if (stoppingOffsetOpt.isPresent()) {
                 Long stoppingOffset = stoppingOffsetOpt.get();
-                if (startingOffset >= stoppingOffset) {
+                if (startingOffset >= stoppingOffset || stoppingOffset == 0) {
```
loserwang1024, line 263: In which situation,  stoppingOffset == 0?

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/source/reader/FlinkSourceSplitReader.java
```diff
@@ -468,6 +470,11 @@ private FlinkRecordsWithSplitIds forLogRecords(ScanRecords scanRecords) {
             splitIdByTableBucket.put(scanBucket, splitId);
             tableScanBuckets.add(scanBucket);
             List<ScanRecord> bucketScanRecords = scanRecords.records(scanBucket);
+            Long consumedUpToOffset = scanRecords.consumedUpToOffset(scanBucket);
+            if (consumedUpToOffset != null && consumedUpToOffset >= stoppingOffset) {
```
loserwang1024, line 474: Maybe we can unsubscribe each finished `scanBucket` directly in `forLogRecords` here, avoiding duplicate  remove logic in two places(removeFinishedSplits).

loserwang1024 (commented):
I have left some tiny comment

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/source/reader/FlinkSourceSplitReader.java
Gezi-lzq: This covers a newly created table or partition with no data, where the latest offset is `0`. Since `EARLIEST_OFFSET` is the sentinel value `-2`, `startingOffset >= stoppingOffset` evaluates to `false` even though there are no records to read. Therefore, we need to handle `stoppingOffset == 0` explicitly.

fluss-flink/fluss-flink-common/src/main/java/org/apache/fluss/flink/source/reader/FlinkSourceSplitReader.java
Gezi-lzq: Thanks for the suggestion. I kept the unsubscribe logic in `fetch()` because `forLogRecords()` is currently responsible for converting `ScanRecords` into `FlinkRecordsWithSplitIds`. Moving it there would introduce scanner-state side effects and mix result construction with lifecycle cleanup. There is no duplicate removal implementation here; `removeFinishedSplits()` remains the single cleanup helper.

Gezi-lzq:
Thanks @loserwang1024 for taking the time to review this PR.

This PR incorporates the fix from PR #3873 for issue #3872, using `consumedUpToOffset >= stoppingOffset` to complete bounded splits when filtering produces no materialized records but scanner progress has reached the end, preventing batch jobs from hanging. It further fixes issue #3972, where an empty bucket in a multi-bucket Log Table could cause batch scans to fail under low parallelism.

loserwang1024 (approved):
LGTM

Gezi-lzq:
Thanks @loserwang1024 for the review and support! This is my first PR contributed to the Fluss community, and I’m happy to be part of it. 😊

PR #4011, opened 2026-08-16
[rust] Add V1 batch statistics collection

### LINKED ISSUE #4008: [rust] Batch statistics implementation
### Search before asking

### Description

[rust] Batch statistics implementation

### Willingness to contribute

leekeiabstraction:
## Summary
- Closes #4008.
- Adds the pieces a writer needs to produce V1 log record batch statistics: `AlignedRowWriter` for the min and max rows, `record::statistics` to reduce an Arrow batch to per-column bounds and null counts, and the `table.statistics.columns` plumbing that selects the columns.
- Statistics are derived from the finished Arrow batch rather than row by row as Java does, so the row-append and `append_arrow_batch` paths are covered by one implementation.
- Nothing emits them yet, so batches stay in the V0 format and behaviour is unchanged.

## Test Plan
- `cargo test -p fluss-rs --lib`: 646 passed, including 16 tests asserting `AlignedRowWriter` output byte by byte against Java's layout and 5 covering the statistics block.
- `cargo fmt --all --check` and `cargo clippy -p fluss-rs --all-targets` clean.

### DIFF
--- a/fluss-rust/crates/fluss/src/metadata/table.rs
+++ b/fluss-rust/crates/fluss/src/metadata/table.rs
@@ -1399,8 +1399,8 @@ impl TableInfo {
     ///
     /// Empty when the table has not enabled statistics. `*` keeps only the
     /// columns whose type supports statistics, while an explicit list is taken
-    /// as given so that naming an unsupported column fails at write time as it
-    /// does in Java.
+    /// as given: the server already rejects an unsupported type when the table
+    /// is created or altered, so the client trusts it as Java's does.
     ///
     /// # Errors
     /// Returns an error if a named column is absent from the table schema.
--- a/fluss-rust/crates/fluss/src/metadata/table.rs
+++ b/fluss-rust/crates/fluss/src/metadata/table.rs
@@ -1049,6 +1049,38 @@ impl Display for PhysicalTablePath {
     }
 }
 
+/// Resolves `table.statistics.columns` against `row_type`, keeping any failure
+/// as a message so that it can be surfaced later without breaking construction.
+fn resolve_stats_index_mapping(
+    table_config: &TableConfig,
+    row_type: &RowType,
+) -> std::result::Result<Vec<usize>, String> {
+    let names = match table_config.get_statistics_columns() {
+        StatisticsColumns::Disabled => return Ok(Vec::new()),
+        StatisticsColumns::All => {
+            return Ok(row_type
+                .fields()
+                .iter()
+                .enumerate()
+                .filter(|(_, field)| is_supported_statistics_type(field.data_type()))
+                .map(|(index, _)| index)
+                .collect());
+        }
+        StatisticsColumns::Specified(names) => names,
+    };
+
+    names
+        .iter()
+        .map(|name| {
+            row_type
+                .fields()
+                .iter()
+                .position(|field| field.name() == name)
+                .ok_or_else(|| format!("Statistics column '{name}' not found in table schema"))
+        })
+        .collect()
+}
+
 #[derive(Debug, Clone)]
 pub struct TableInfo {
     pub table_path: TablePath,
@@ -1067,6 +1099,9 @@ pub struct TableInfo {
     pub comment: Option<String>,
     pub created_time: i64,
     pub modified_time: i64,
+    /// Resolved once at construction. The failure is held rather than raised so
+    /// that a malformed property only breaks writers, not metadata loading.
+    stats_index_mapping: std::result::Result<Vec<usize>, String>,
 }
 
 impl TableInfo {
@@ -1302,6 +1337,7 @@ impl TableInfo {
         let physical_primary_keys =
             Self::generate_physical_primary_key(&primary_keys, &partition_keys);
         let table_config = TableConfig::from_properties(properties.clone());
+        let stats_index_mapping = resolve_stats_index_mapping(&table_config, &row_type);
 
         TableInfo {
             table_path,
@@ -1320,6 +1356,7 @@ impl TableInfo {
             comment,
             created_time,
             modified_time,
+            stats_index_mapping,
         }
     }
 
@@ -1404,34 +1441,13 @@ impl TableInfo {
     ///
     /// # Errors
     /// Returns an error if a named column is absent from the table schema.
-    pub fn get_stats_index_mapping(&self) -> Result<Vec<usize>> {
-        let names = match self.table_config.get_statistics_columns() {
-            StatisticsColumns::Disabled => return Ok(Vec::new()),
-            StatisticsColumns::All => {
-                return Ok(self
-                    .row_type
-                    .fields()
-                    .iter()
-                    .enumerate()
-                    .filter(|(_, field)| is_supported_statistics_type(field.data_type()))
-                    .map(|(index, _)| index)
-                    .collect());
-            }
-            StatisticsColumns::Specified(names) => names,
-        };
-
-        names
-            .iter()
-            .map(|name| {
-                self.row_type
-                    .fields()
-                    .iter()
-                    .position(|field| field.name() == name)
-                    .ok_or_else(|| Error::IllegalArgument {
-                        message: format!("Statistics column '{name}' not found in table schema"),
-                    })
-            })
-            .collect()
+    pub fn get_stats_index_mapping(&self) -> Result<&[usize]> {
+        match &self.stats_index_mapping {
+            Ok(mapping) => Ok(mapping),
+            Err(message) => Err(IllegalArgument {
+                message: message.clone(),
+            }),
+        }
     }
 
     pub fn get_table_config(&self) -> &TableConfig {
--- a/fluss-rust/crates/fluss/src/metadata/table.rs
+++ b/fluss-rust/crates/fluss/src/metadata/table.rs
@@ -22,6 +22,7 @@ use crate::metadata::DataLakeFormat;
 use crate::metadata::datatype::{
     DataField, DataType, RowType, UNASSIGNED_FIELD_ID, reassign_field_ids,
 };
+use crate::record::is_supported_statistics_type;
 use crate::{BucketId, PartitionId, SnapshotId, TableId};
 use core::fmt;
 use serde::{Deserialize, Serialize};
@@ -32,6 +33,10 @@ use strum_macros::EnumString;
 
 /// Sentinel for a column whose stable id has not yet been assigned.
 pub const UNKNOWN_COLUMN_ID: i32 = -1;
+
+/// Table property selecting the columns that written batches collect statistics
+/// for, either `*` or a comma-separated list.
+pub const TABLE_STATISTICS_COLUMNS: &str = "table.statistics.columns";
 
 #[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq)]
 pub struct Column {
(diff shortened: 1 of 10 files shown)


leekeiabstraction:
@fresh-borzoni @charlesdong1991 Appreciate review here 🙏

fluss-rust/crates/fluss/src/record/statistics.rs
```diff
@@ -0,0 +1,732 @@
...
+    mapping: &[usize],
+) -> Result<Option<Vec<u8>>> {
+    if mapping.is_empty() || batch.num_rows() == 0 {
+        return Ok(None);
+    }
+
+    let mut null_counts = Vec::with_capacity(mapping.len());
+    let mut bounds = Vec::with_capacity(mapping.len());
+    for &column_index in mapping {
+        let column = batch.column(column_index);
```
fresh-borzoni, line 141: mapping is checked against the table schema but not against the batch, and batch.column(i) panics out of range.

fluss-rust/crates/fluss/src/metadata/table.rs
```diff
@@ -1355,6 +1395,45 @@ impl TableInfo {
...
+    /// Column indices, in order, that written batches collect statistics for.
+    ///
+    /// Empty when the table has not enabled statistics. `*` keeps only the
+    /// columns whose type supports statistics, while an explicit list is taken
+    /// as given so that naming an unsupported column fails at write time as it
+    /// does in Java.
+    ///
+    /// # Errors
+    /// Returns an error if a named column is absent from the table schema.
+    pub fn get_stats_index_mapping(&self) -> Result<Vec<usize>> {
```
fresh-borzoni: Java rejects unsupported statistics columns at table creation, not at write time. Why do we prefer to differ here?

fluss-rust/crates/fluss/src/row/aligned/aligned_row_writer.rs
```diff
@@ -0,0 +1,608 @@
...
+        let pos = self.current_pos;
+        let slice = &bytes[..length.min(bytes.len())];
+        self.write_bytes_internal(pos, slice);
+        self.current_pos = pos + 1;
+    }
+
+    fn write_decimal(&mut self, value: &Decimal, precision: u32) {
+        let pos = self.current_pos;
+        if Decimal::is_compact_precision(precision) {
+            let unscaled = value.to_unscaled_long().unwrap_or(0);
```
fresh-borzoni: unwrap_or(0) turns a decimal that doesn't fit in i64 into a silent 0, which then ships as a min or max. Java throws here. Shall we propagate it or there is a reason for clamping?

fresh-borzoni (commented):
@leekeiabstraction Thank you for the PR, left a couple of comments, PTAL

fluss-rust/crates/fluss/src/record/statistics.rs
leekeiabstraction: Well spotted, added mapped index range checks. Note that check against table schema is done in `column_bounds`

fluss-rust/crates/fluss/src/metadata/table.rs
leekeiabstraction: Well spotted, I've made an error in the documentation. The implementation is already mirroring Java's `TableInfo.getStatsIndexMapping()` which is called at write time.

fluss-rust/crates/fluss/src/row/aligned/aligned_row_writer.rs
leekeiabstraction: Well spotted, I've updated to mirror Java side client. Using panic instead of Result::Error due to existing signature of writers. Changing the signature or the reducing writer visibility to pub(crate) probably requires some discussion and is out of scope of this PR.

leekeiabstraction:
@fresh-borzoni TY for your review, I've addressed your comments and also push a separate fix to a bug I found. PTAL.

fluss-rust/crates/fluss/src/row/aligned/aligned_row_writer.rs
```diff
@@ -0,0 +1,623 @@
...
+    /// always 8 bytes and the low half would otherwise go to waste.
+    fn set_offset_and_size(&mut self, pos: usize, offset: usize, size: u64) {
+        let packed = ((offset as i64) << 32) | (size as i64);
+        let field_offset = self.field_offset(pos);
+        self.put_long_le(field_offset, packed);
+    }
+
+    /// Inlines a payload of at most 7 bytes into the field slot, with
+    /// `len | 0x80` in the slot's high byte as Java's `writeBytesToFixLenPart`.
+    fn write_bytes_to_fix_len_part(&mut self, pos: usize, bytes: &[u8]) {
```
naivedogger: The low-level encoding logic here substantially overlaps with `FlussArrayWriter` in `binary_array.rs`. Both implement the same wire-format rules for inline bytes, aligned variable-length storage, offset/size packing, and non-compact Decimal/Timestamp encoding, but through separate code paths.

On the Java side, the common encoding logic is centralized in `AbstractBinaryWriter` and reused by both `AlignedRowWriter` and `BinaryArrayWriter`, while writer-specific field addressing remains separate.

Would it make sense to extract a small internal encoding/buffer helper in Rust so that `AlignedRowWriter` and `FlussArrayWriter` can share these rules without requiring the same API or trait? This would reduce duplication and help prevent format divergence.

naivedogger (approved):
Thanks @leekeiabstraction for working on this! Overall, LGTM. I left one minor comment, PTAL.

fluss-rust/crates/fluss/src/row/aligned/aligned_row_writer.rs
fresh-borzoni: +1, and there is a third copy: PaimonBinaryRowWriter implements the same rules again

fluss-rust/crates/fluss/src/row/aligned/aligned_row_writer.rs
fresh-borzoni: The module row::aligned is introduced in this PR, so pub(crate) mod aligned would not change any writer signature and would keep the ARRAY and MAP panics out of the public API.

ValueWriter::write_value routes ARRAY and MAP into any BinaryWriter, and Java's AlignedRow supports both through AlignedArray and AlignedMap, so a caller using this writer outside statistics would hit an abort rather than an error. With the module made internal, the only new warning is that size_in_bytes has no caller, which is already true today.

Would it make sense to keep the module internal until the writers can return Result? The visibility can be widened later without a compatibility concern, while the reverse is not true.

fluss-rust/crates/fluss/src/record/statistics.rs
```diff
@@ -0,0 +1,908 @@
...
+                }
+                _ => Ok(None),
+            }
+        }
+        DataType::Decimal(decimal_type) => {
+            let array = column
+                .as_any()
+                .downcast_ref::<PrimitiveArray<Decimal128Type>>()
+                .ok_or_else(|| unexpected_array(column, data_type))?;
+            let (precision, scale) = (decimal_type.precision(), decimal_type.scale());
```
fresh-borzoni, line 324: DECIMAL takes the scale from the type, not the array, so a Decimal128(10,2) batch on a DECIMAL(10,3) column gives 1.234 instead of 12.34. column_vector.rs uses Decimal::from_arrow_decimal128 for this. Should the source scale come from column.data_type() here too?

fluss-rust/crates/fluss/src/record/statistics.rs
```diff
@@ -0,0 +1,908 @@
...
+                column.as_any().downcast_ref::<Time64MicrosecondArray>(),
+                1_000,
+            ),
+            ArrowType::Time64(_) => time64_as_millis(
+                column.as_any().downcast_ref::<Time64NanosecondArray>(),
+                1_000_000,
+            ),
+            _ => Err(unexpected_array(column, data_type)),
+        },
+        DataType::String(_) | DataType::Char(_) => {
```
fresh-borzoni, line 301: Java's collector has no CHAR branch so it ships a null bound, while we ship a real one. A CHAR predicate would prune our batches and never Java's. 
Should we file an issue on the Java side?

fluss-rust/crates/fluss/src/metadata/table.rs
```diff
@@ -1355,6 +1395,45 @@ impl TableInfo {
...
+    /// Column indices, in order, that written batches collect statistics for.
+    ///
+    /// Empty when the table has not enabled statistics. `*` keeps only the
+    /// columns whose type supports statistics, while an explicit list is taken
+    /// as given: the server already rejects an unsupported type when the table
+    /// is created or altered, so the client trusts it as Java's does.
+    ///
+    /// # Errors
+    /// Returns an error if a named column is absent from the table schema.
+    pub fn get_stats_index_mapping(&self) -> Result<Vec<usize>> {
```
fresh-borzoni: Java caches this because it's on the per-batch path, here every call re-splits the property string. Should we cache it on TableInfo?

fluss-rust/crates/fluss/src/record/statistics.rs
```diff
@@ -0,0 +1,908 @@
...
+            ),
+        });
+    }
+
+    let mut null_counts = Vec::with_capacity(mapping.len());
+    let mut bounds = Vec::with_capacity(mapping.len());
+    for &column_index in mapping {
+        let column = batch.column(column_index);
+        null_counts.push(column.null_count() as i32);
+        bounds.push(column_bounds(
```
fresh-borzoni, line 143: Java never fails a write over statistics, MemoryLogRecordsArrowBuilder catches and writes length 0, while here any column_bounds error reaches the caller. 
Is it intended?

fresh-borzoni (commented):
@leekeiabstraction Thank you, left some additional comments and question, PTAL

fluss-rust/crates/fluss/src/record/statistics.rs
leekeiabstraction: Raised issue https://github.com/apache/fluss/issues/4066

fluss-rust/crates/fluss/src/record/statistics.rs
leekeiabstraction: Well caught, updated.

fluss-rust/crates/fluss/src/record/statistics.rs
leekeiabstraction: Yes, the thinking is to leave the logic to handle Err and skip writing statistics in rust side's `MemoryLogRecordsArrowBuilder`

fluss-rust/crates/fluss/src/metadata/table.rs
leekeiabstraction: Caching on per-batch path actually need more machinery on Rust side than I'd like. Updated to cache on TableInfo instantiation.

fluss-rust/crates/fluss/src/row/aligned/aligned_row_writer.rs
leekeiabstraction: Thank you both, I've refactored it a bit.  There could be more refactoring to reduce duplication and risk of drift, but let's address that outside of this PR so that it's manage-able and easier to review.

fluss-rust/crates/fluss/src/record/statistics.rs
leekeiabstraction: * In #4010

leekeiabstraction:
@fresh-borzoni Thank you for the detailed review. Updated the PR, PTAL 🙏

fresh-borzoni (approved):
@leekeiabstraction Thank you, LGTM 👍

